PR #1909, opened 2026-05-29
fix(cpp): handle variadic pack dependent lookup

azizur100389:
## Summary
Fixes #1894.

Adds conservative C++ handling for variadic packs and dependent-name lookup:

- suppresses concrete EXTENDS capture for pack-expanded bases like struct Mix : B...
- treats class templates with pack-expanded bases as having an unknown dependent base set, so unqualified member calls do not bind to unrelated class-owned methods
- preserves current-instantiation member lookup when the member is declared on the current class
- keeps unknown-specialization member types unresolved
- adds exact C++ fixture assertions for variadic pack arity, fold expressions, pack-base suppression, current-instantiation positives, and unknown-specialization negatives

## Validation
- npx prettier --check on touched TypeScript files
- npx tsc --noEmit
- REGISTRY_PRIMARY_CPP=1 focused C++ block: 7 passed
- REGISTRY_PRIMARY_CPP=0 focused C++ block: 6 passed, 260 skipped
- REGISTRY_PRIMARY_CPP=1 full C++ integration: 266 passed
- REGISTRY_PRIMARY_CPP=0 full C++ integration: 203 passed, 63 skipped
- first npm test run: one unrelated timing-sensitive phase-timer bound failed; direct rerun of test/unit/phase-timer.test.ts passed 6/6
- second npm test run: 394 files passed, 9857 tests passed, 50 skipped
- node dist/cli/index.js detect-changes: low risk, 0 affected processes

## Notes
Legacy DAG cannot represent the new pack-base suppression and still over-emits Mix::run -> B::inherited, so the new negative assertion is documented in the existing C++ legacy parity expected-failure set.

Issue #1894, opened 2026-05-29
cpp templates: variadic packs & dependent-name resolution


magyargergo:
**Phase 1 · Size L · Part of #1890**

Consolidates audit items **B5** (variadic packs / folds / pack-bases) + **B7** (current-instantiation dependent-name resolution). Builds on the completed scope-resolution baseline in #1564.

## Symptom
```cpp
template<class... Ts> void log(Ts... xs){ (consume(xs), ...); } // fold over consume → no edge; pack arity wrong
template<class... B> struct Mix : B... {};                       // pack-expansion base not captured
template<class T> struct W { typename T::value_type v; void f(){ v.use(); } }; // dependent member type → no edge
```

## Anchor on main
- Packs: `captures.ts:535-563` records variadic param names but `constraint-extractor.ts:283-303` increments one `argIdx` per param with no pack semantics; arity (`computeCppDeclarationArity`, used `captures.ts:100-114`) has no variadic marker like the `'params'` flag the narrower checks at `overload-narrowing.ts:117-120`; `iterBaseClasses` (`captures.ts:567-581`) skips `parameter_pack_expansion` bases; `query.ts` has no `fold_expression` capture.
- Dependent names: two-phase support covers only *bases* (`two-phase-lookup.ts:268-279`, `captures.ts:594-636`); `typename T::value_type` member types aren't inferred (`interpret.ts`/`captures.ts` stop at `type_identifier`/`template_type`); current-instantiation vs unknown-specialization (`[temp.dep.type]`) not distinguished.

## ISO reference
`[temp.variadic]`, `[expr.prim.fold]`, `[temp.deduct.call]` p1; `[temp.dep]`, `[temp.dep.type]`, `[temp.res]`.

## Conservative option
Mark pack-bearing templates as open-ended arity (honored by the narrower). Capture `fold_expression`; emit a single edge only for an unambiguous single-callee fold. Treat `Base...` pack bases as dependent (suppress unqualified binding). Distinguish **current-instantiation** members (resolvable within the enclosing template's own class/bases) and resolve those; keep `T::member` (unknown specialization) unresolved; recognize `typename`/`.template` tokens to avoid mis-parse.

## Full-ISO option
Element-wise pack deduction & expansion (needs Engine 3 / L6); full current-instantiation member lookup per `[temp.dep.type]`.

## Definition of Done
[...]

PR #2247, opened 2026-06-18
fix(cpp-hooks): handle pack-base comments and missing hook overrides

azizur100389:
## Summary

Detect C++ pack-expanded bases when tree-sitter surfaces a comment between the base type and the ellipsis marker.

## Motivation / context

Follow-up to #1909: Gergo's review noted that `B /*pack*/ ...` was not treated as a pack-expanded base because comment nodes stopped the sibling scan before `...`.

## Areas touched

## Scope & constraints

**In scope**

- Skip C++ comment siblings while looking for the pack-expansion ellipsis after a base specifier.
- Extend the existing variadic dependent-resolution fixture to cover `B /*pack*/ ...`.
- Rebaseline the C++ scope-capture fingerprint with a note for the intentional fixture drift.

- No broader C++ two-phase lookup refactor.
- No changes to non-C++ language behavior.

## Implementation notes

The scanner still stops at commas, access specifiers, and other named nodes; it only treats comments as trivia between the base and `...`.

## Testing & verification

- [x] `cd gitnexus && npx vitest run test/integration/resolvers/cpp.test.ts -t "variadic packs"`
- [x] `cd gitnexus && node --import tsx bench/scope-capture/measure.mjs --check`

## Risk & rollout

Low risk. GitNexus staged-change analysis reported one touched symbol, `isFollowedByPackExpansion`, with low risk and no affected processes. The scope-capture check passed for all languages; C++ scaling remained linear at `1.086 < 1.5`.

## Checklist

### DIFF
--- a/gitnexus/hooks/claude/hook-db-lock-probe.cjs
+++ b/gitnexus/hooks/claude/hook-db-lock-probe.cjs
@@ -99,6 +99,17 @@ function resolveHookBinary(tool) {
   return tool;
 }
 
+function hasMissingHookBinaryOverride(tool) {
+  const envKey = tool === 'lsof' ? 'GITNEXUS_HOOK_LSOF_PATH' : 'GITNEXUS_HOOK_PS_PATH';
+  const fromEnv = process.env[envKey];
+  if (!fromEnv || !String(fromEnv).trim()) return false;
+  try {
+    return !fs.existsSync(String(fromEnv).trim());
+  } catch {
+    return true;
+  }
+}
+
 // Sentinel:
 //   undefined = not resolved yet (resolve lazily, on first lsof/ps fallback)
 //   string    = self-tested coreutils timeout/gtimeout path (use as wrapper)
@@ -597,6 +608,9 @@ function linuxProcScanFindGitNexusServer(dbPathAbs, myPid) {
 
 function unixLsofPsFindGitNexusServer(dbPathAbs, myPid) {
   const guard = resolveUnixGuardTimeout();
+  // An explicit missing override models ENOENT and must fail open instead of
+  // falling through to a host binary with different process-table visibility.
+  if (hasMissingHookBinaryOverride('lsof')) return false;
   const lsofPath = resolveHookBinary('lsof');
   // The spawnSync timeouts below (lsof 1000ms / ps 500ms) are deliberately
   // SHORTER than the wrapper budgets (2s / 1s): on the supervised path Node's
@@ -629,9 +643,12 @@ function unixLsofPsFindGitNexusServer(dbPathAbs, myPid) {
   if (guard && (lsof.status === 124 || lsof.status === 137)) return true;
 
   const pids = (lsof.stdout || '').split(/\s+/).filter(Boolean);
+  const psMissing = hasMissingHookBinaryOverride('ps');
   const psPath = resolveHookBinary('ps');
   for (const pid of pids) {
     if (Number(pid) === myPid) continue;
+    // Missing ps means we cannot verify that this pid is a GitNexus server.
+    if (psMissing) continue;
     const [psCmd, psArgs] = guard
       ? [guard, ['-k', '1', '1', psPath, '-p', pid, '-o', 'command=']]
       : [psPath, ['-p', pid, '-o', 'command=']];
--- a/gitnexus/test/fixtures/lang-resolution/cpp-variadic-dependent-resolution/main.cpp
+++ b/gitnexus/test/fixtures/lang-resolution/cpp-variadic-dependent-resolution/main.cpp
@@ -24,7 +24,7 @@ void foldAmbiguous(Ts... xs) {
 }
 
 template <class... B>
-struct Mix : B... {
+struct Mix : B /*pack*/ ... {
   void run() {
     inherited();
     helper();
--- a/gitnexus/test/fixtures/lang-resolution/cpp-variadic-dependent-resolution/main.cpp
+++ b/gitnexus/test/fixtures/lang-resolution/cpp-variadic-dependent-resolution/main.cpp
@@ -24,6 +24,13 @@ void foldAmbiguous(Ts... xs) {
 }
 
 template <class... B>
+struct PlainMix : B... {
+  void plainRun() {
+    inherited();
+  }
+};
+
+template <class... B>
 struct Mix : B /*pack*/ ... {
   void run() {
     inherited();
--- a/gitnexus-claude-plugin/hooks/hook-db-lock-probe.cjs
+++ b/gitnexus-claude-plugin/hooks/hook-db-lock-probe.cjs
@@ -99,6 +99,17 @@ function resolveHookBinary(tool) {
   return tool;
 }
 
+function hasMissingHookBinaryOverride(tool) {
+  const envKey = tool === 'lsof' ? 'GITNEXUS_HOOK_LSOF_PATH' : 'GITNEXUS_HOOK_PS_PATH';
+  const fromEnv = process.env[envKey];
+  if (!fromEnv || !String(fromEnv).trim()) return false;
+  try {
+    return !fs.existsSync(String(fromEnv).trim());
+  } catch {
+    return true;
+  }
+}
+
 // Sentinel:
 //   undefined = not resolved yet (resolve lazily, on first lsof/ps fallback)
 //   string    = self-tested coreutils timeout/gtimeout path (use as wrapper)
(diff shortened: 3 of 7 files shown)


gitnexus/hooks/claude/hook-db-lock-probe.cjs
```diff
@@ -597,6 +608,9 @@ function linuxProcScanFindGitNexusServer(dbPathAbs, myPid) {
 
 function unixLsofPsFindGitNexusServer(dbPathAbs, myPid) {
   const guard = resolveUnixGuardTimeout();
+  // An explicit missing override models ENOENT and must fail open instead of
+  // falling through to a host binary with different process-table visibility.
+  if (hasMissingHookBinaryOverride('lsof')) return false;
```
magyargergo, line 613: **[P2] This new guard isn't regression-guarded by a test.** `hasMissingHookBinaryOverride` is new in this PR, and the existing tests that drive the missing-override path — `hooks.test.ts:2031` (`ENOENT lsof → augment still runs`) and `:2473` (`ps ENOENT → augment runs`) — already pass on `main` **without** this code: when the override path is missing, `resolveHookBinary` (line 81) falls through to host `/usr/bin/lsof` / `/bin/ps`, which finds no holder for the empty temp DB, so augment runs either way. So deleting this `return false` (and the `psMissing` skip just below) leaves both tests green on any host with a real lsof/ps — nothing locks in the new behavior.

Suggested: assert the host binary is **not** consulted — add a `lsofMarkerPath` to `createHookToolDir` and assert it is absent after the run; for ps, make the lsof-reported PID look like a GitNexus server via `psOutputByPid` so the test fails without the `psMissing` skip. Applies to both `hook-db-lock-probe.cjs` copies. *[code-read]*

gitnexus/test/fixtures/lang-resolution/cpp-variadic-dependent-resolution/main.cpp
```diff
@@ -24,7 +24,7 @@ void foldAmbiguous(Ts... xs) {
 }
 
 template <class... B>
-struct Mix : B... {
+struct Mix : B /*pack*/ ... {
```
magyargergo, line 34: **[P3, minor] This fixture edit replaces the plain `B...` case rather than adding to it.** This line was `struct Mix : B...`; the comment variant is now the only pack-base form the `#1894` tests exercise, so the comment-free path through `isFollowedByPackExpansion` loses its dedicated assertion. Low risk — the pack path is still covered and CI is green across the whole describe block — but consider adding a second base/struct for the plain `B...` form (and optionally a `// line` comment variant) so both paths stay guarded against future grammar/loop changes. *[code-read]*

magyargergo (commented):
## Tri-review — `fix(cpp): detect pack-expanded bases across comments`

**Methods (3):** GitNexus swarm (risk, test/CI) · Compound-Engineering personas (correctness, adversarial, maintainability, testing) · **Codex**. Engine breakdown: **6 Claude lanes + Codex (live — the only non-Claude engine).** Codex+Claude agreement is weighted as strong corroboration; Claude-only agreement is "consistent across personas," not independent confirmation.

**Verdict:** the C++ fix is **correct, minimal, and properly tested — ship-worthy.** Two non-blocking concerns: (1) the PR bundles an **unrelated hooks change** under a cpp-scoped title; (2) that hooks change is **not regression-guarded** by any test that would fail without it. Neither is a correctness defect.

### ✅ What's solid (validated, not merely unchallenged)
- **The C++ fix is correct.** `isFollowedByPackExpansion` now skips `comment` siblings *before* the `isNamed → return false` fallback, and the ordering is load-bearing: tree-sitter-cpp emits `comment` as a **named** node, so without the skip a comment between a base and `...` wrongly returned `false`. Correctness and Codex independently traced it; the adversarial lane exercised the real tree-sitter-cpp@0.23.2 grammar over 20+ base-clause shapes (multiple comments, `//` vs `/* */`, after `access_specifier`, before/after `,`, trailing) — all correct. The change can only convert a premature `false` into "keep scanning"; it can never invert a `true` (the `...` and `,`/`access_specifier` guards run first).
- **The fix is genuinely covered.** The fixture now reads `struct Mix : B /*pack*/ ...` and the existing pack-base tests in `cpp.test.ts` (the `#1894` describe block — e.g. *"does not emit a concrete EXTENDS edge for a pack-expanded base"*) exercise it; without the captures.ts change they fail. I reproduced the scope-capture bench locally (`measure.mjs --check` → `PASS (14 languages)`, exit 0): the cpp fingerprint matches the committed `baselines.json` value (`d12d9f3e…`) and scaling stays linear (< 1.5). The `tests / benchmarks` CI job is green too — the baseline update is validated, not hand-edited. *[reproduced]*
- **The hooks change is a production no-op.** `hasMissingHookBinaryOverride` (new in this PR) returns `false` whenever the env var is unset, so both new guards are inert in real deployments — `GITNEXUS_HOOK_LSOF_PATH` / `GITNEXUS_HOOK_PS_PATH` are never set in production. The affected path (`unixLsofPsFindGitNexusServer`) is macOS/BSD-only; Linux uses a separate procfs scan. The adversarial and correctness lanes both **refuted** the data-corruption hypothesis: no production code, settings, installer, CLI, or CI sets those vars to a missing path; the lbug's own lock still applies; and the new fail-open matches the file's documented Unix contract.

### ⚠️ Inline findings (2)
1. **[P2] Scope / branch hygiene — the hooks change is unrelated to the cpp title.** Single commit `9740863b`, titled `fix(cpp):`, bundles the one-line C++ parser fix with a ~29-line DB-lock-probe change across **both** `hook-db-lock-probe.cjs` copies (`gitnexus/hooks/claude/` + `gitnexus-claude-plugin/hooks/`). Risk, maintainability, and Codex all flag it independently: the `fix(cpp)` scope omits a behavioral hook change, which muddies `detect_changes` / bisect / release-notes — gates this repo relies on. **Recommend splitting** into `fix(cpp):` + `fix(hooks):` (or at minimum retitling). *(See inline note on the hook change.)*
2. **[P2] The new hook guards are not regression-guarded.** Codex + the testing lane + a coordinator trace agree the existing missing-override tests pass on `main` *without* this change. (Details inline at `hook-db-lock-probe.cjs:613`.)

### Lower-priority (non-blocking)
- **Fixture coverage trade-off:** the change *replaces* the plain `struct Mix : B...` case rather than adding to it, so the comment-free path loses its dedicated assertion; only a `/* */` comment is exercised (not `//`, though both are the same `comment` node type). Minor — see inline at `main.cpp:27`.
- **`.trim()` inconsistency (benign, Claude-only):** `hasMissingHookBinaryOverride` calls `fs.existsSync(String(fromEnv).trim())` while `resolveHookBinary` (line 84) uses the un-trimmed value, so a whitespace-padded *valid* path makes the two disagree. Correctness confirmed the consequence is benign (the guard merely declines to fire and falls back to prior host-binary behavior — no wrong verdict). Not Codex-corroborated.
- **`EPERM` treated as `ENOENT`:** the `catch { return true }` makes a permission error on the override path fail-open rather than distinguishing it (Codex). Test-only path, very low priority.
- **Copy duplication** of `hook-db-lock-probe.cjs` is a pre-existing smell; this PR correctly keeps both copies **byte-identical** (credit). No third-copy parity gap — `antigravity` ships a different hook file, and there are only two copies of this probe in the repo.

### Refuted / validated (a feature, not a gap)
- **Data-corruption from fail-open:** adversarial could construct no production trigger (env vars have no production setter; macOS-only path; lbug lock still applies). **Refuted.**
- **C++ edge cases** (multi-comment, line comment, comment around `,`/`access_specifier`, trailing): traced and refuted across correctness + adversarial + Codex.

### CI
Green except **Vercel** (`Authorization required to deploy` — infra/auth, not a code gate). All `quality` / `tests` / `CodeQL` / `gitleaks` / tree-sitter-ABI / cross-platform jobs pass. `mergeStateStatus: UNSTABLE` is solely the Vercel failure → mergeable on the code gates.

---
*Automated multi-tool digest (GitNexus swarm + Compound-Engineering + Codex), critic-gated. Verify findings before acting.*

azizur100389:
Thanks for the tri-review. I addressed the two inline findings in 8ddb4e82:\n\n- Added regression coverage for the missing hook override paths. The macOS/BSD lsof/ps tests now install a self-testing timeout wrapper with markers, so they fail if a missing GITNEXUS_HOOK_LSOF_PATH or GITNEXUS_HOOK_PS_PATH falls through to a guard-wrapped fallback binary. The ps case also reports a GitNexus-looking process line, so the missing-ps skip is load-bearing.\n- Restored separate comment-free C++ pack-base coverage by adding PlainMix : B... alongside Mix : B /*pack*/ ... and asserting both the EXTENDS suppression and inherited-call suppression.\n\nI also retitled the PR to include the hook scope, and rebaselined the C++ scope-capture fingerprint after the fixture-only drift. Local targeted checks passed: hooks.test.ts, cpp variadic packs, tsc --noEmit, and scope-capture --check.

PR #2057, opened 2026-06-05
fix(java): dedupe inherited RequestMapping prefixes

azizur100389:
## Summary
- dedupe inherited Spring route prefixes when a controller and implemented interface use the same `@RequestMapping` prefix
- keep distinct prefixes compositional, including shared leading segments such as `/open` + `/open/ai/query`
- add Java HTTP extractor regressions for the duplicate-prefix case and the conservative non-dedupe case

## Validation
- `npx vitest run test/unit/group/http-route-extractor.test.ts --maxWorkers=1` (89 passed)
- `npx tsc --noEmit`
- `npx prettier --check src/core/group/extractors/http-patterns/java.ts test/unit/group/http-route-extractor.test.ts`
- `npm test` was run locally: 10,050 passed, 61 skipped, 6 failed. The failures were outside this Java HTTP extractor change; rerunning the failed files with `--maxWorkers=1` reduced them to the pre-existing/unrelated literal-collectors C# assertion and Swift timing tripwire.

Fixes #2028

PR #2254, opened 2026-06-19
feat(group): Kotlin Spring HTTP consumer extraction + provider parity with Java

glier:
## Summary

Brings the **Kotlin** group/contract HTTP extractor up to parity with the Java plugin for inter-service contract detection, and **unifies the shared (language-agnostic) consumer logic** into a new module so the two plugins cannot drift apart.

Before this PR a Kotlin+Spring estate surfaced HTTP **providers** but almost no **consumers** and no cross-repo links — the entire inter-service call graph was invisible. Root cause: in tree-sitter-kotlin an `interface` is a `class_declaration`, so the Kotlin plugin (a) mis-classified `@FeignClient` interface methods as providers, (b) had no Spring HTTP Interface / `@RequestLine` support, and (c) emitted phantom providers for plain route interfaces.

## What changed

**Consumers (new for Kotlin):**
- `@FeignClient` interface `@(Get|…)Mapping` methods are now emitted as OpenFeign **consumers** (a remote call), not providers.
- Spring 6 HTTP Interface `@(Get|…)Exchange` (with optional class-level `@HttpExchange(url)` prefix) — added for **both Java and Kotlin**.
- Native OpenFeign `@RequestLine`, gated to interfaces (Feign proxies are interfaces only), mirroring `java.ts`'s `findEnclosingInterface` check.

**Providers (Kotlin parity with `scanSpringProject`):**
- A `@(Get|…)Mapping` on a non-Feign interface is treated as a route *contract*, not a served route — it is skipped in `scan()` so the implementing `@RestController` is the sole provider (Java drops these implicitly via `interface_declaration`).
- `scanProject` inherits interface routes onto the implementing class, gated on the class being `@RestController`/`@Controller` (`kotlinClassIsController` handles both the attached `modifiers` shape and the detached leading-arg-form `prefix_expression` shape), so non-controller implementers don't emit phantom providers.

**Shared module (`spring-consumer-shared.ts`):**
- Language-agnostic primitives (`REST_TEMPLATE_`/`WEB_CLIENT_`/`EXCHANGE` verb maps, `joinPath`, `joinInheritedSpringPath`, `parseRequestLine`, framework + confidence constants) now live in one place; `java.ts` and `kotlin.ts` both import them so a polyglot repo emits byte-identical contract IDs.

**Array-of-paths (both languages):**
- Spring route attributes are `String[]`; a multi-element array registers the route under **every** element. The class/Feign/HttpExchange prefix maps now accumulate all elements (were last-write-wins) and emission cross-products prefixes × method paths, so `@RequestMapping(["/a","/b"]) + @GetMapping(["/x","/y"])` yields all four contract IDs. Array form is matched via a predicate-free alternation over Kotlin `collection_literal` / Java `element_value_array_initializer`.

## Tests

Comprehensive Java + Kotlin cases: consumer-vs-provider classification, `@*Exchange`, `@RequestLine` (interface-only + plain-interface + anti-overreach on a class), interface-based controller inheritance, non-controller negative case, detached `@RestController`, single- and multi-element array paths (method-level and class-prefix cross-product, plus a `#2057`-style prefix-dedup parity case). `test/unit/group` is green; `tsc` / ESLint / Prettier clean.

Validated end-to-end on a private multi-service Kotlin+Spring estate: consumer extraction (`@FeignClient` + `@*Exchange`) and provider detection now produce correct cross-repo links, with no provider/consumer mis-classification.

## Known limitations (out of scope for this PR)

- **Cross-link matching is path-only** — `matching.ts` ignores `@FeignClient(name = …)`, so two services that expose the same path can be linked by path collision. Disambiguating by the Feign service name is a separate follow-up.
- **Kotlin `fun interface`** (functional/SAM interfaces) parse to a grammar `ERROR` node in the vendored tree-sitter-kotlin, so routes declared on a `fun interface` are not extracted — needs a grammar upgrade or ERROR-recovery heuristic.
- Gateway path rewrites (a consumer calling a gateway-prefixed path that the provider serves un-prefixed) are not reconciled — that's a routing-topology concern, not a static-extraction one.

## Checklist

- [x] Tests pass for the touched package (`gitnexus`)
- [x] Conventional-commit PR title
- [x] No secrets, tokens, or machine-specific paths
- [x] Pre-commit hook clean (lint-staged + typecheck)
- [x] Change confined to a single domain (`core/group/extractors/http-patterns` + its tests)


gitnexus/src/core/group/extractors/http-patterns/kotlin.ts
```diff
@@ -547,8 +929,68 @@ function buildKotlinPlugin(language: unknown): HttpLanguagePlugin {
...
+        const valueNode = match.captures.value;
+        const nameNode = match.captures.method_name;
+        const methodNode = match.captures.method;
+        if (!valueNode || !methodNode) continue;
+        const raw = unquoteLiteral(valueNode.text);
+        const parsed = raw !== null ? parseRequestLine(raw) : null;
+        if (!parsed) continue;
+        const enclosingClass = findEnclosingClass(methodNode);
+        if (!enclosingClass || !isKotlinInterface(enclosingClass)) continue;
+        const prefixes = feignPrefixByClassId.get(enclosingClass.id) ?? [''];
```
magyargergo: **[P2 · parity bug · reproduced by Codex + correctness + adversarial]** A `@RequestLine` method on an interface that has a class-level `@RequestMapping("/x")` but no `@FeignClient(path=…)` drops the `/x` prefix here, while `java.ts` applies it (via its `interfaceRequestMappingPrefixes` fallback into `feignPrefixByInterfaceId`) → Java emits `consumer GET /x/ping`, Kotlin emits `consumer GET /ping`, breaking the byte-identical-contract goal. It's also internally inconsistent: the `@GetMapping`-in-`@FeignClient` path at line 815 already has the `?? prefixByClassId.get(enclosingClass.id)` fallback that this line omits.

Fix (mirror line 815):
```ts
const prefixes = feignPrefixByClassId.get(enclosingClass.id) ??
  prefixByClassId.get(enclosingClass.id) ?? [''];
```
_[reproduced ×3] + [code-read]_

gitnexus/src/core/group/extractors/http-patterns/kotlin.ts
```diff
@@ -400,49 +401,401 @@ function buildKotlinPlugin(language: unknown): HttpLanguagePlugin {
...
+      {
+        meta: {},
+        query: `
+          (function_declaration
+            (modifiers
+              (annotation
+                (constructor_invocation
+                  (user_type (type_identifier) @ann (#eq? @ann "RequestLine"))
+                  (value_arguments
+                    (value_argument . (string_literal) @value)))))
```
magyargergo, line 668: **[P2 · coverage gap · reproduced by adversarial, flagged by testing lane]** `SPRING_REQUEST_LINE_PATTERNS` matches only the positional `@RequestLine("GET /x")` form (the leading `.` anchor requires the string literal to be the first argument). The named-arg form `@RequestLine(value = "GET /x")` — which `java.ts` accepts (its key check allows `value`) — is silently dropped, so Kotlin emits **zero** consumers where Java emits one.

Add a second pattern for the named form:
```
(value_argument (simple_identifier) @key (#eq? @key "value") (string_literal) @value)
```
_[reproduced] + [code-read]_

magyargergo (commented):
## PR #2254 Tri-Review — Kotlin Spring HTTP consumer extraction + Java parity

**Methods (7):** GitNexus swarm (risk-architect, test/CI) + Compound-Engineering personas (correctness, adversarial, maintainability, testing) — **all Claude** — plus **Codex**, the one independent engine (ran live and reproduced the headline). Breakdown: **6 Claude lanes + 1 Codex**, so Claude-lane agreement is "consistent across personas," and **Codex+Claude agreement (the headline) is the strong signal** — not 7 independent confirmations. (The GitNexus test/CI lane ended mid-investigation; its domain was covered by the testing + risk lanes. Implementation-session intent was recovered via claude-mem history.)

**What's solid (validated — credit to the author):**
- **The Java refactor is provably faithful.** Every constant moved into `spring-consumer-shared.ts` (`FEIGN_CONFIDENCE=0.7`, `REQUEST_LINE_CONFIDENCE=0.75`, `OPENFEIGN_FRAMEWORK='openfeign'`, …) byte-matches the pre-PR inlined literal it replaced — no existing Java contract is re-keyed or re-scored. `classPrefix:string → classPrefixes:string[]` is output-identical for the single-prefix case (the only case that existed) and correctly fixes the previously-truncated multi-element `@RequestMapping({"/a","/b"})`.
- **Refuted suspicions (probed, found sound):** `@*Exchange` enclosing-scope resolution is symmetric Java↔Kotlin; the shared `findEnclosingClass` is byte-identical to the removed local helper; `isKotlinInterface` is correct; the `#2057` double-prefix dedup and prefix×route cross-product hold parity; `REQUEST_LINE_VERB_RE` is not ReDoS-vulnerable (anchored both ends, bounded input); the `scanProject` addition is compatible with the orchestrator's optional-guard.
- **The abstraction boundary is the right call** — share pure value-maps/helpers, keep per-language AST walkers.

**① Headline (inline) — P2, reproduced by Codex + correctness + adversarial.** `kotlin.ts:978`: a `@RequestLine` method on an interface with a class-level `@RequestMapping("/x")` (and no `@FeignClient(path=…)`) **drops the `/x` prefix in Kotlin, while Java applies it** → Java `GET /x/ping` vs Kotlin `GET /ping`, defeating the byte-identical-contract goal. It's also internally inconsistent: the `@GetMapping`-in-Feign path at `kotlin.ts:815` already carries the `?? prefixByClassId.get(id)` fallback that line 978 omits. One-line fix.

**② Second inline — P2, reproduced by adversarial + flagged by testing lane.** `kotlin.ts:530` (`SPRING_REQUEST_LINE_PATTERNS`) matches only the positional `@RequestLine("GET /x")`; the named-arg `@RequestLine(value = "GET /x")` form — which `java.ts` accepts — is silently dropped, so Kotlin emits zero consumers where Java emits one.

**Lower-priority / single-lane / contested (verify before acting):**
- **`arrayOf("/a")` annotation arrays** (P3, reproduced by adversarial + correctness): the Kotlin queries DO handle the array-literal `["/a","/b"]` (`collection_literal`) but NOT the `arrayOf("/a","/b")` call form (a `call_expression`), so `@RequestMapping(arrayOf("/api"))` emits no prefix and `@GetMapping(arrayOf("/x"))` emits no route. Java has no `arrayOf`, so it's a Kotlin-only coverage hole, not a same-syntax divergence. Add an `arrayOf` branch, or land an explicit documented limitation + anti-overreach test.
- **WebClient long-form asymmetry** (P3, adversarial reproduced): Kotlin detects `webClient.method(HttpMethod.X).uri()` consumers that `java.ts` *deliberately defers* — Kotlin over-detects relative to Java. Decide the parity direction and document it.
- **Fully-qualified inline annotations** (P2 but niche, adversarial only): `kotlinAnnotationName`/`collectKotlinSupertypes` take the *first* `type_identifier` segment, so an inline `@org.springframework…RestController` / FQN supertype may resolve to `org`/`com` and be missed, while the route queries match FQNs that `java.ts`'s `scoped_identifier` shape rejects — divergence both ways. Rare (almost all code imports), and the exact grammar AST shape wasn't independently corroborated — recommend the author confirm the `user_type` shape and use the trailing segment.
- **`prefix_expression` branch — engines disagree:** Codex calls `kotlinClassIsController`'s `prefix_expression` branch *dead code with an incorrect grammar comment* (says `@RestController("bean")` parses under `modifiers`); adversarial says it can cause a rare formatting-sensitive class-prefix loss; correctness refuted it for realistic (package-bearing) code. Low-risk — author should confirm the AST and either remove the dead branch + fix the comment, or add the missing detached-prefix scan.

**Maintainability (non-blocking):** good direction, but two drift channels the PR set out to close remain — `pushKotlinPrefix` duplicates `java.ts`'s `pushPrefix` (identical 4-line body; both belong in the shared module), and `scanKotlinProject` ≈ `scanSpringProject` are ~70-line near-duplicate algorithms with no shared type. `kotlin.ts` now crosses 1005 lines; one stale comment ("…the keys of `WEB_CLIENT_SHORT_TO_HTTP` above" — it's now imported from the shared module, not above).

**Test gaps (parity claim not fully proven):** Kotlin twins are missing for several Java-tested consumer scenarios — `@RequestLine` query-strip, mixed `@RequestLine`+`@GetMapping`, malformed-value rejection, `@FeignClient(path)`-wins-when-`@RequestMapping`-first, and `@FeignClient(path)`+`@RequestLine` prefix join. Kotlin provider tests never assert `confidence` (only contractId/framework), so a wrong-confidence regression would pass. A single table-driven harness feeding matched `.java`/`.kt` pairs through both plugins and diffing the emitted contract sets would have caught both inline findings directly.

**CI:** required checks green — format, lint, typecheck, CodeQL, gitleaks, tree-sitter ABI (all 3 OS), and the macOS full test run (which **exercises the Kotlin tests — they run and pass, not silently skipping**). `tests/ubuntu/coverage` + Windows + `Build & Push` still pending; **Vercel fails on deploy-auth only** (infra, not code) — the sole reason `mergeStateStatus` is UNSTABLE.

**Net:** none of these is a regression — Kotlin had no `@RequestLine`/`@*Exchange`/Feign-consumer support before this PR, so the gaps are incompleteness in net-new functionality, not breakage. Reasonable to merge with the two inline fixes applied (or tracked as fast-follows).

*Automated multi-tool digest (GitNexus swarm + CE personas + Codex). Verify before acting on any single item.*

PR #2227, opened 2026-06-16
feat(impact): opt-in PDG-backed impact mode - statement + inter-procedural slicing, resolved-callee-id soundness, mutation-oracle validated

magyargergo:
## Summary

This PR adds an opt-in **`mode:'pdg'`** to the `impact` tool that answers a question the call graph cannot: **"which statements does changing the code at line N affect?"** It computes a forward **program slice** over a persisted **Program Dependence Graph** — data dependence (`REACHING_DEF`) and control dependence (`CDG`) over BasicBlocks — and follows that slice **across function boundaries**. The default `mode:'callgraph'` is unchanged and byte-identical; PDG mode is additive and opt-in.

## How it fits together

There are two phases. `gitnexus analyze --pdg` builds and persists the dependence substrate once; `impact(mode:'pdg', line:N)` reads it to compute a slice on demand.

```mermaid
flowchart LR
  subgraph ingest["analyze --pdg  ·  build the substrate (once)"]
    direction TB
    SRC["source files"] --> CFG["per-function CFG"]
    CFG --> RD["REACHING_DEF<br/>data dependence"]
    CFG --> CDG["CDG<br/>control dependence"]
    CFG --> CID["BasicBlock.calleeIds<br/>resolved callee symbol ids"]
    CFG --> CSUM["CALL_SUMMARY<br/>formal → return summaries"]
  end
  RD & CDG & CID & CSUM --> G[("persisted graph<br/>(LadybugDB)")]
  subgraph readq["impact(mode:'pdg', line:N)  ·  query (read)"]
    direction TB
    SEED["seed block at line N"] --> INTRA["intra-procedural slice<br/>REACHING_DEF + CDG"]
    INTRA --> INTER["inter-procedural descent (calleeIds)<br/>+ return-value ascent (CALL_SUMMARY)"]
    INTER --> RES["affectedStatements<br/>+ symbol reach"]
  end
  G --> SEED
```

### What `analyze --pdg` persists

- **`REACHING_DEF`** — statement-granular data dependence; each edge carries the full def→use line chain in its `reason`, so the slice can recover the interior statements of coalesced straight-line blocks.
- **`CDG`** — control dependence (post-dominator + Ferrante), language-agnostic.
- **`BasicBlock.calleeIds`** — the *resolved callee symbol ids* of each block's call sites (not leaf names), harvested across all 12 languages.
[...]

PR #2258, opened 2026-06-20
fix(impact-pdg): make the Impact PDG Mutation Report workflow pass (3 latent oracle bugs)

magyargergo:
The nightly **Impact PDG Mutation Report** workflow ([run 27878048944](https://github.com/abhigyanpatwari/GitNexus/actions/runs/27878048944)) failed. It was the **first real execution** of the `--mutation` oracle added in #2227 (prior runs only did vitest + actionlint), so three independent latent bugs surfaced in sequence — each fixed and re-verified by dispatching the workflow on this branch.

| # | Symptom | Root cause | Fix |
|---|---------|-----------|-----|
| 1 | `ERR_MODULE_NOT_FOUND: src/cli/lazy-action.js` (~2s in) | The harness shells `analyze` out as `node --import tsx src/cli/index.ts …`. CI's runner Node **22.22.3** enables native TypeScript type-stripping by default (graduated in 22.18); it handles the `.ts` entry instead of tsx, and native stripping does **not** remap `./lazy-action.js`→`lazy-action.ts` the way tsx does. | Run the analyze child from the built **`dist/cli/index.js`** (plain JS — no tsx/strip-types; parse workers also load from `dist/`), which the workflow already builds; fall back to tsx's own CLI over `src/` for build-free local runs. |
| 2 | `Cannot find package '@babel/traverse'` (~45s in) | `mutation-oracle.mjs` imports `@babel/parser`, `@babel/traverse`, `@babel/generator`, `@babel/types`, none declared. Two are hoisted into the gitnexus tree transitively; `@babel/traverse` and `@babel/generator` only exist at the monorepo root, so a fresh `npm ci` in `gitnexus/` can't resolve them. | Declare all four as devDependencies. |
| 3 | Recall gate: `min realized recall 0.000 < floor 0.5` (oracle ran fully) | The gate scored every check with a numeric `recall`, including the **upstream** fixture `intra-control-loop`. The oracle is a *forward* value-diff (mutate the criterion line, observe downstream value changes), so its behavioral AIS can never intersect a *reverse* (upstream) PDG slice — recall is 0 by construction. `measure.mjs` already marks these `recallGated: false`. | The standalone `gate-mutation-recall.mjs` now honors `recallGated`, so the floor applies only to the downstream cases a forward oracle can fairly validate. |

## Verification

Reproduced #1 and #2 against a real **Node v22.22.3** binary; validated each fix locally and then by dispatching the actual workflow on this branch. Final run is **green**:

```
✓ Run PDG impact mutation oracle (~280s)
✓ Gate on mutation recall regression
  - scored cases: 11 of 15
  - min realized recall: 1.000 (floor 0.5)
  - mean realized recall: 1.000
  - cases below floor: 0
```

Workers run green on 22.22.3 in CI (independently confirmed via the `tests / benchmarks` job of run `27874383902`). No production code changed — only the bench harness, its declared dev deps, and the gate script.

### DIFF
--- a/gitnexus/bench/impact-pdg/README.md
+++ b/gitnexus/bench/impact-pdg/README.md
@@ -227,10 +227,14 @@ analyze via a temp `GITNEXUS_HOME`, mock-free**. Per fixture:
    first, keeping the source tree clean).
 2. **Shell out** to the real CLI as a child process — child-process isolation
    sidesteps `process.exit`; real `saveMeta` + `registerRepo` land in the temp
-   home; parse workers spawn from `dist/` (so the harness needs a built `dist/`):
+   home. The harness prefers the built `dist/` CLI (plain JS, no tsx; the parse
+   workers it spawns also load from `dist/`), so it needs a built `dist/`; it
+   falls back to tsx's own CLI over `src/` for build-free local runs. (`node
+   --import tsx src/cli/index.ts` is avoided: Node ≥22.18 native type-stripping
+   breaks the `.ts` entry's `./lazy-action.js`→`.ts` import resolution.)
 
    ```
-   node --import tsx src/cli/index.ts analyze <fixtureCopy> --pdg --skip-git --index-only
+   node dist/cli/index.js analyze <fixtureCopy> --pdg --skip-git --index-only
    ```
 3. `new LocalBackend(); await init()` resolves the fixture via the **real**
    registry (the parent process sets `GITNEXUS_HOME` too, so `init()` reads the
--- a/gitnexus/bench/impact-pdg/gate-mutation-recall.mjs
+++ b/gitnexus/bench/impact-pdg/gate-mutation-recall.mjs
@@ -17,7 +17,14 @@ const floor = Number(process.env.MUTATION_RECALL_FLOOR ?? '0.5');
 
 const report = JSON.parse(fs.readFileSync(reportPath, 'utf8'));
 const checks = Array.isArray(report?.mutation?.checks) ? report.mutation.checks : [];
-const scored = checks.filter((c) => typeof c.recall === 'number');
+// Gate only the checks the oracle marked recall-gated. measure.mjs sets
+// `recallGated: false` for cases a forward value-diff oracle cannot fairly
+// score against the PDG slice: UPSTREAM fixtures (the oracle runs in its native
+// downstream sense, so its behavioral AIS can never intersect a reverse slice —
+// recall is 0 by construction) and id-discrimination corroboration fixtures.
+// Those still carry a numeric `recall` for the report, so the legacy
+// `typeof c.recall === 'number'` filter wrongly tripped the floor on them.
+const scored = checks.filter((c) => c.recallGated === true && typeof c.recall === 'number');
 const recalls = scored.map((c) => c.recall);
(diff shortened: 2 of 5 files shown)


magyargergo (commented):
## Tri-review digest — PR #2258

**Verdict:** production-ready (small CI/tooling PR; the two findings below are already addressed in `00e52fa7`).

**Methods:** 5 reviewer lanes (risk, test/CI, correctness, adversarial, maintainability) + Codex. **Engine breakdown:** all 5 lanes are Claude personas — their agreement is *consistent across personas*, **not** independent confirmation. **Codex evaporated** (returned no structured findings), so treat this as a single-engine, multi-persona review — not three independent confirmations.

**Scope:** 4 files in the original PR, no production code — `measure.mjs` (dist-preferred analyze child), `gate-mutation-recall.mjs` (recall-gate filter), `package.json`+lock (+4 `@babel` devDeps). Full diff read against base `78b4077`.

### What's solid (validated, not merely unobjected)
The three original fixes are correct and CI-confirmed green (run [27879420266](https://github.com/abhigyanpatwari/GitNexus/actions/runs/27879420266), Node 22.22.3): the dist-preferred child clears the `lazy-action.js` startup crash; the four `@babel` devDeps match the oracle's imports exactly and are lockfile-pinned + integrity-verified (caret supply-chain risk **refuted**); the `recallGated` gate filter correctly excludes the upstream + id-corroboration fixtures that a forward value-diff oracle cannot fairly score.

### Findings — both already addressed in `00e52fa7`
1. **Vacuous-green gate (latent).** `gate-mutation-recall.mjs` carries a *pre-existing* null-exit path — `if (min !== null && min < floor)`, so an empty `scored` set (`min === null`) passes the gate. The new `recallGated === true` filter is strictly narrower than the prior `typeof recall === 'number'`, so it **exposed** that latent path in more inputs (a degenerate corpus, or a harvest that silently emptied every behavioral AIS — adversarial sketched a Babel-drift chain). Not triggerable on today's 15-fixture corpus (~12 are recall-gated). Flagged consistently across the Claude personas (correlated, not independent). **Fixed:** the gate now fails when checks exist but none are recall-gated; a genuinely empty (0-check) report still passes. *Blocks merge: no (resolved).*
2. **README doc drift.** `README.md:233` was not in the original PR diff but was made stale by it — it still documented the replaced `node --import tsx src/cli/index.ts` invocation. **Fixed** in `00e52fa7` (README now in the diff), updated to the dist-preferred form. *Blocks merge: no (resolved).*

### Refuted / not blocking
- **Stale-`dist/` measurement** and **eager `tsx/cli` resolve throwing**: real but confined to local dev (the workflow builds `dist/` fresh + full `npm ci`); not a CI/prod hazard.
- **`@babel` caret range**: refuted — lockfile-pinned exact + integrity, installed via `npm ci`, first-party packages.

### CI & coverage
The pre-hardening commit ran green (run 27879420266); the hardening commit re-runs CI now. Testing gap (noted by 3 lanes): no unit test for `gate-mutation-recall.mjs` — a fixture-driven test of the `recallGated` filter + the new empty-`scored` guard would lock the contract (follow-up, not blocking).

_Automated multi-tool digest — 5 Claude personas; Codex unavailable, so treat as single-engine. Verify before acting._

Issue #2203, opened 2026-06-15
perf(ingestion): large-repo analyze wall time dominated by graph-DB emit, not PDG


magyargergo:
## Summary

Large-repo `analyze` wall time is dominated by the parse + graph-DB-write pipeline, not the PDG layer. Investigate emit/persistence throughput.

## Background

PDG scale validation (PR #2197) timing breakdown:

| target | files | total wall | scope-resolution | emit | **pdg solve** |
|---|---|---|---|---|---|
| postgres | 2,398 | ~39 min | 20.6 s | 12.4 s | 4.46 s |
| linux `fs/` | 2,176 | ~65 min | 179 s | 171 s | **7.16 s** |

The analyze wall is ~1–1.6 s/file at scale. The PDG algorithms are a tiny fraction (kernel `fs/`: 7.2 s out of a 65-min wall). The time goes to parsing and **graph-DB persistence** of ~1M-edge graphs (LadybugDB writes), plus scope-resolution.

This is a pre-existing pipeline-throughput property — **not** a regression from the PDG work — but it dominates the at-scale experience.

## Proposed work

- Profile the emit/persistence path on the postgres / kernel-`fs/` benchmarks; investigate batched/bulk inserts, fewer write round-trips, or parallelized emit.

## Acceptance

- Measurable files/sec improvement on the postgres and kernel-`fs/` benchmarks with byte-identical graph content.

Surfaced by #2195 scale validation. Refs #2195.

Issue #1169, opened 2026-04-28
analyze exits 0 with empty .gitnexus/lbug.wal on Windows (Python repo, no segfault) — also reproduces on 1.4.1 (pre-LadybugDB)


DzikPasnik:
Likely related to #1162 and #1154 but distinct enough to file separately:
this case is **Python** (not C++) and the process **exits cleanly with
code 0** (no segfault, no stderr) yet never finalizes the database.

## Environment
- OS: Windows 11 Pro (10.0.26200)
- Shell: Git Bash (MSYS2)
- Node.js: v24
- GitNexus: 1.6.3 (and `rc` = 1.6.4-rc.21 — same behavior)

## Repro (minimal — 2 files, ~200 bytes total)
```bash
mkdir gn-test && cd gn-test && git init -q
git config user.email t@t && git config user.name t

cat > foo.py <<'PY'
def hello(name: str) -> str:
    return f"hi {name}"
class Greeter:
    def greet(self, n): return hello(n)
PY

cat > bar.py <<'PY'
from foo import Greeter
def main(): Greeter().greet("world")
if __name__ == "__main__": main()
PY

git add -A && git commit -qm init

gitnexus analyze . --skip-agents-md
# stdout: just the "GitNexus Analyzer" banner. exit 0.

ls .gitnexus/
# lbug       (4096 B)
# lbug.wal   (~25 KB)   ← never flushed/committed

gitnexus list      # → \"No indexed repositories found.\"
gitnexus status    # → \"Repository not indexed.\"
```

## Also reproduces on a real ~430-file Python project
After excluding \`.venv/\` (≈10.6k vendored .py files) via \`.gitnexusignore\`,
the symptom is identical: 23-line warning log (\`scope extraction failed\`
for empty \`__init__.py\` files plus \"Invalid argument\" for ~10 large
sources), banner-only stdout, exit 0, **no \`.gitnexus/\` materialized**.

The \"Invalid argument\" warnings during scope extraction are *not* the
root cause — adding those files to \`.gitnexusignore\` does not change the
outcome. The persistence/finalize step never runs.

## Also tested 1.4.1 (pre-LadybugDB migration, #275)
Same symptom: \`gitnexus@1.4.1 analyze\` (via isolated npm cache) returns
exit 0 with no \`.gitnexus/\` directory created at all. So this is **not a
fresh regression introduced by the LadybugDB v0.15 migration** — the
finalize path appears broken on Windows across both DB engines (old
[...]

PR #2264, opened 2026-06-21
fix(lbug): stop --pdg analyze double-free (skip LadybugDB close-destructor crash) + harden connection serialization

magyargergo:
## Problem

`gitnexus analyze --pdg` aborts at the end ("Saving metadata..." ~98%) with:

```
double free or corruption (out)
[1]  ... IOT instruction  gitnexus analyze --pdg
```

This is native heap corruption (glibc SIGABRT) — pure JS can't double-free.

## Root cause

LadybugDB is single-writer and its `Connection` is **not safe for concurrent query execution**. Two things issue queries on the same module-singleton `conn` during analyze:

- the **WAL-checkpoint driver** (`wal-checkpoint-driver.ts`) on a 5s `setInterval` → `conn.query('CHECKPOINT')`, and
- the analyze **pipeline** holding that same `conn` for a long COPY.

`--pdg` adds the `BasicBlock` + `REACHING_DEF / CDG / POST_DOMINATE / TAINTED / CALL_SUMMARY / TAINT_PATH` tables, so a COPY outlasts the 5s tick. The checkpoint then runs concurrently with the in-flight COPY on one connection → two libuv workers mutate shared native state → heap smash (caught at a later `free()`, hence "at 98%"). Without `--pdg` the writes finish inside the tick window, so the overlap rarely fires.

## Fix

- New `src/core/lbug/conn-lock.ts` — `withConnLock`, a promise-chain mutex.
- Every singleton-`conn` helper now runs its full query **+ result drain** inside the lock: `queryAndDrain` (when `targetConn === conn`), `executePrepared`, `executeWithReusedStatement`, `flushWAL`, `tryFlushWAL`, `getLbugStats`, `deleteAllInterprocTaintPaths`, `deleteAllCallSummaries`.
- `if (inflight) return` reentrancy guard in the driver `tick()` so overdue ticks can't stack checkpoints.
- `streamQuery` is intentionally **not** wrapped (read path; its per-row callback can re-enter the adapter, and the driver is inactive there).

## Reproduction & validation

- **Repro:** concurrent `conn.query(bigWrite)` ∥ `conn.query('CHECKPOINT')` on one raw `@ladybugdb/core` connection → native crash; the *identical* workload serialized → stable.
- **Fix validated end-to-end:** drove the same overlap through the real locked adapter (init + real WAL driver + write ∥ 2×CHECKPOINT) — clean.
- New unit tests: `conn-lock.test.ts` (no overlap / FIFO / throw-releases) and `wal-checkpoint-driver-reentrancy.test.ts` (mock + fake timers). Existing `lbug-load-overlap*` + `lbug-adapter-wal-schema` + `analyze-wal-checkpoint-failure` still pass.

> Note: `@ladybugdb/core` 0.17.1 also segfaults on its own "Maximum database size reached" (16 GiB) error path — a separate latent native bug, orthogonal to this crash (worth a follow-up issue).


gitnexus/src/core/lbug/lbug-adapter.ts
```diff
@@ -2036,44 +2080,49 @@ export const deleteAllCommunitiesAndProcesses = async (): Promise<{
...
-    // A missing table on a freshly-initialized DB is the benign, expected case
-    // (the count query above is what throws) — stay silent. Any OTHER failure
-    // (lock, disk, native error) would leave stale TAINT_PATH rows that the
-    // subsequent re-extract then DUPLICATES (CodeRelation has no PK), so it
-    // must ABORT the writeback (#2084 review P2-5): re-throw so the caller's
-    // crash-recovery dirty flag forces a clean full rebuild on the next run,
-    // rather than silently writing duplicate cross-function findings.
-    const msg = err instanceof Error ? err.message : String(err);
-    if (/no table|not exist|not found|does not exist|Table .* does not exist/i.test(msg)) {
+  // count + DELETE run as one critical section on the singleton connection so a
```
magyargergo, line 2170: **[P1] Incomplete fix — incremental `--pdg` still races.** You wrapped this (`deleteAllInterprocTaintPaths`) and `deleteAllCallSummaries`, but three structurally-identical singleton-`conn` sites that run inside the same WAL-driver window (`run-analyze.ts:1000`→`1551`) were left unwrapped, so they can still run concurrently with the driver `CHECKPOINT` on the one connection — the same double-free class this PR fixes, on the incremental path:

- `deleteAllCommunitiesAndProcesses` (`lbug-adapter.ts:2050`/`2055`) — count **+ `DETACH DELETE`** (a write), called at `run-analyze.ts:1126`. Its neighbor comment even says *"exactly like Community/Process."*
- `queryImporters` (`lbug-adapter.ts:2017`) — raw `conn.query`, BFS loop at `run-analyze.ts:1067`.
- `deleteNodesForFile` count (`lbug-adapter.ts:1952`) — raw `targetConn!.query` while its sibling DELETE at `:1960` is locked.

Fix: route these through `withConnLock` (or a locked row-returning helper). Corroborated by Codex + correctness + adversarial; re-verified by reading the call sites. `[code-read]`

magyargergo (commented):
## Tri-review: PR #2264 — lbug connection serialization

**Methods (3) / engines (2).** GitNexus swarm (risk, test/CI) + Compound-Engineering personas (correctness, adversarial, performance, maintainability) + **Codex**. Six Claude reviewer lanes were dispatched — the `gitnexus-risk-architect` lane ended mid-investigation with no structured findings — plus **Codex, the only non-Claude engine**. So Claude-lane agreement is "consistent across personas," not independent confirmation; the strong signal is **Codex + a Claude lane agreeing**, which is what the headline below has.

**Validated — actively checked, not merely unchallenged:**
- **Mutex (`conn-lock.ts`) is correct** — FIFO chain, release-on-throw via `finally`, returns the wrapped result (Codex + correctness + the new unit tests).
- **No deadlock** — adversarial + Codex traced every wrapped helper; all are leaf-level, none calls another wrapped helper, so the non-reentrancy invariant holds for the current code.
- **The #2203 node-COPY ‖ rel-emit overlap is preserved** — rel-emit writes CSVs and never touches `conn`, so wrapping `queryAndDrain` does not serialize it away (performance lane).
- **`streamQuery` exemption is safe today** — its only callers are the read-only API graph-stream routes; no analyze/driver-active path reaches it (Codex + adversarial).
- **`queryAndDrain`'s `targetConn === conn` guard** correctly locks the singleton and skips temp connections; lock overhead is negligible.

**P1 — the fix is incomplete on the incremental `--pdg` path.** Three singleton-`conn` query sites run inside the WAL-checkpoint-driver window (`run-analyze.ts:1000`→`1551`) but bypass `withConnLock`, so they can still execute concurrently with the driver's `CHECKPOINT` on the one connection — the *same* double-free class this PR fixes, just on the incremental (not full-rebuild) path. **Corroborated by Codex + correctness + adversarial, and re-verified by reading the call sites:**
- `deleteAllCommunitiesAndProcesses` — raw `conn.query` count **and `DETACH DELETE`** (`lbug-adapter.ts:2050`/`2055`), called at `run-analyze.ts:1126`. It's the structural twin of `deleteAllInterprocTaintPaths` / `deleteAllCallSummaries`, which this PR *did* wrap — the neighbor comment even says *"exactly like Community/Process."* **Most dangerous (a write).**
- `queryImporters` — raw `conn.query` (`lbug-adapter.ts:2017`), called in a BFS loop at `run-analyze.ts:1067`.
- `deleteNodesForFile` count — raw `targetConn!.query` (`lbug-adapter.ts:1952`) while the sibling `DETACH DELETE` at `:1960` *is* locked via `queryAndDrain`; called (singleton) at `run-analyze.ts:1115`.

  Fix: route these through `withConnLock` (or a `queryAndDrain`-style locked, row-returning helper).

**P2 — undrained DELETE result** (Codex). The wrapped `deleteAllInterprocTaintPaths` / `deleteAllCallSummaries` (and `deleteAllCommunitiesAndProcesses`) `await conn.query(… DELETE …)` without draining/closing the returned `QueryResult` (the count result is closed; the delete result is dropped). Pre-existing pattern — worth draining for consistency with the PR's own "query + drain inside the lock" contract.

**P2/P3 — hardening (Claude lanes only; optional but valuable):**
- The non-reentrancy invariant is comment-only — a future wrapped→wrapped call hangs silently. A one-boolean re-entry assertion in `withConnLock` turns that into a test-visible throw.
- A single private `connQuery` / `queryRows` choke-point (all singleton `conn.query` routed through it) would have **structurally prevented** the P1 trio, versus per-site wrapping that's easy to miss — the strongest long-term fix.
- `__resetConnLockForTests` naming diverges from the repo's `_initLockPathForTest` test-seam convention (single underscore).

**Tests.** The new unit tests deterministically prove serialization (`maxActive === 1`, FIFO, release-on-throw) and the reentrancy guard, and land in the CI `default` vitest project. Gap (corroborates the headline): nothing exercises the **incremental writeback path** with the driver live — exactly where the three unwrapped sites hide.

**CI:** pending at review time (branch just pushed).

**Recommendation.** Correct mutex and a genuine fix for the full-rebuild crash, but **incomplete for its own stated goal** — the incremental `--pdg` path keeps the same race at three sites. Address the P1 trio (ideally via the choke-point refactor) before merge.

*Automated multi-tool digest (GitNexus swarm + CE personas + Codex — one independent engine). Verify against the code before acting.*

gitnexus/src/core/lbug/lbug-adapter.ts
magyargergo: Fixed — all three now route through `withConnLock` (with a routing assertion each in the new `test/integration/lbug-conn-serialization.test.ts`, a call-through `withConnLock` spy):

- `deleteAllCommunitiesAndProcesses` (count + DETACH DELETE) → 77505991
- `queryImporters` → 071b1201
- `deleteNodesForFile` count, gated on `targetConn === conn` so per-query/temp connections stay lock-free → f5fb6fc3

magyargergo:
## Tri-review findings addressed (6 commits, one per fix)

| Finding | Fix | Commit |
|---|---|---|
| **P1** — `deleteAllCommunitiesAndProcesses` raw count + DETACH DELETE on the singleton during the driver window | wrap in `withConnLock` | 77505991 |
| **P1** — `queryImporters` raw read in the importer BFS loop | wrap in `withConnLock` | 071b1201 |
| **P1** — `deleteNodesForFile` count unlocked while its sibling DELETE was locked | gate the count on `targetConn === conn` | f5fb6fc3 |
| **P2** — `deleteAll*` dropped the DELETE `QueryResult` (only the count was closed) | drain/close via `closeQueryResults` | ed720685 |
| **Hardening** — non-reentrancy invariant was comment-only (silent deadlock risk) | `AsyncLocalStorage` re-entry guard (throws, not hangs) | f0cb3685 |
| **Nit** — test-seam naming | `__resetConnLockForTests` → `_resetConnLockForTests` | 837a8528 |

**Re-entry guard note:** a boolean flag (as originally suggested) would false-fire on legitimately-queued top-level callers — they also run while the lock is held. `AsyncLocalStorage` marks the running `fn`'s async context, so only a *true* nested (wrapped→wrapped) call throws. Covered by tests: sequential and concurrent top-level callers do **not** throw.

**Deferred (#6):** the single `connQuery`/`queryRows` choke-point that would structurally prevent the P1 class is left as a follow-up refactor PR — kept here to one-commit-per-fix with bounded blast radius.

**Tests:** new `test/integration/lbug-conn-serialization.test.ts` (routing + drain, in the `lbug-db` project) and re-entry-guard cases in `test/unit/conn-lock.test.ts`. `tsc` clean; 53 unit + 4 integration tests green locally.

🤖 follow-up to the automated tri-review

magyargergo:
## Root cause corrected — the real `--pdg` crash is a LadybugDB close-destructor double-free

After building and running a real `analyze --pdg` locally (the e2e repro the reviews flagged as missing) and capturing a **gdb backtrace**, the actual crash is:

```
"double free or corruption (out)"  →  abort()
  lbug::main::ClientContext::~ClientContext()      ← LadybugDB engine
  lbug::main::Connection::~Connection()
  NodeConnection::Close(...)                        ← conn.close() (safeClose)
```

It's a **double-free inside LadybugDB's own `ClientContext` destructor during `conn.close()`** — and it **reproduces with the WAL driver OFF and with serial load**, so it is *not* the checkpoint/COPY concurrency the earlier commits on this branch serialized. That serialization is still valid hardening (concurrent queries on a single-writer connection are UB), but it is **orthogonal** to this crash. `@ladybugdb/core` 0.17.1 is the latest stable; this is an upstream engine bug.

It fires during teardown **after** a fully-written, checkpointed index (meta/registry/AI-context and the WAL CHECKPOINT all complete first), so the data is intact — the process just aborts on the way out. A JS `try/catch` can't catch a glibc `abort()`.

## The actual fix (commit a84645a4)

`closeLbug({ skipNativeClose })` CHECKPOINTs for durability (`flushWAL`) then **skips `conn.close()`/`db.close()`**, leaving the handles referenced so no GC finalizer re-runs the destructor. The CLI `analyze` command opts in via `skipNativeCloseOnExit` on all three paths that `process.exit` (success / error / SIGINT); long-lived callers (MCP server, tests) keep the real close. Mirrors this repo's existing native-cleanup precedent — the pool adapter's fire-and-forget close (`pool-adapter.ts`) and the ONNX teardown.

## Validation (end-to-end, real CLI)

- `analyze --pdg --force` → **exit 0**, `Repository indexed successfully`, 193,876 nodes / 417,247 edges (was: `free(): invalid pointer`, exit 134).
- Re-open with `analyze --pdg` (no `--force`) → reads clean, `Incremental: changed=0 … 1637 unchanged file rows preserved` → **CHECKPOINT-only persistence is durable** without `db.close()`.
- Regression test: `closeLbug({ skipNativeClose })` checkpoints, leaves the connection ready + queryable.

Follow-up: file the `ClientContext::~ClientContext` double-free upstream with LadybugDB. The branch's connection-serialization commits remain as hardening.

🤖 follow-up after local gdb investigation

magyargergo:
Thanks — all three addressed (separate commits):

| Finding | Fix | Commit |
|---|---|---|
| **P1-1** CI red: `closeLbug` comment contained literal `conn.close()`/`db.close()`, tripping the structural guard in `lbug-checkpoint.test.ts` | reworded the comment (code already delegates close to `safeClose`; guard intent preserved) | `8589feb5` |
| **P1-2** CLI error path could hang — `skipNativeClose` on the error path left live native handles while the CLI soft-returns (`process.exitCode = 1`) | error path now does a **real close**; only the success path (which guarantees a following `process.exit`) skips. A late-error close may abort but terminates — it can't hang | `f3b80962` |
| **P2-3** forked analyze worker did a real close before force-exit | worker now passes `skipNativeCloseOnExit: true` (it `process.exit(0)`s after) so the destructor can't abort it before it sends `complete` | `8a6799c5` |

Re-validated locally: `tsc` clean; `lbug-checkpoint` + `conn-lock` + `wal-checkpoint-driver-reentrancy` (20) and `lbug-conn-serialization` (5) green; `analyze --pdg --force` exits 0 (193,879 nodes). The invariant is now explicit in `closeLbug`: **skip the native close only when the caller guarantees a `process.exit` after** (CLI success + worker); every other path closes for real.

🤖 follow-up addressing review P1-1 / P1-2 / P2-3

gitnexus/src/cli/analyze.ts
```diff
@@ -1273,6 +1276,12 @@ const analyzeCommandImpl = async (
         // Extra fetch-wrapper names from `.gitnexusrc` (#1589/#1852 residual);
         // forwarded to the routes phase consumer scan.
         fetchWrappers: options.fetchWrappers,
+        // The CLI always process.exit()s after this returns (success path at the
+        // end of analyzeCommandImpl, error/interrupt paths via process.exit too),
+        // so the finalize close skips the native conn/db close — it can double-free
+        // in LadybugDB's ClientContext destructor after --pdg writes (#2264). The
+        // CHECKPOINT keeps the index durable; process exit reclaims the handles.
+        skipNativeCloseOnExit: true,
```
magyargergo, line 1298: **[P1] Full-analysis success path can HANG (regression).** Setting `skipNativeCloseOnExit: true` makes `runFullAnalysis` finalize via `closeLbug({ skipNativeClose })` — native handles left **open**, relying on a guaranteed `process.exit`. But on the full-analysis success path the CLI then calls `await assertAnalysisFinalized(repoPath)` (`analyze.ts:1360`, *not* try/caught), which throws in a documented state (meta/registry missing, see `1353-1359`). That hits the outer catch (~`1464`) → `process.exitCode = 1; return` (soft, no `process.exit`) → the open native handles wedge the event loop → **hang instead of exit 1**.

Pre-PR the real close released the handles, so this is a regression — the same class as the error-path hang already fixed, but the success-then-post-work window. The `alreadyUpToDate` path is safe (never opens the DB).

Fix: force `process.exit(1)` in the outer catch when a skip-close happened (or real-close there before returning). Verified by code-read. `[code-read]`

magyargergo (commented):
## Tri-review (re-run): PR #2264 — lbug close-crash fix + connection serialization

**Methods / engines.** 5 Claude reviewer lanes (correctness, adversarial, production-risk, maintainability, test/CI) + coordinator code-read verification. **Codex (the independent engine) was dispatched but returned an async/background run with no structured findings in time — so this is effectively a Claude-lanes review; the headline below is graded on the coordinator's own code-read of the path, not on cross-engine agreement.** The prior tri-review (commit `82428228`, pre-fix) is superseded — its findings are addressed.

### 🔴 P1 — CLI full-analysis success path can HANG (regression)
On a real `analyze` (not the up-to-date fast path), `runFullAnalysis` finalizes via `closeLbug({ skipNativeClose })` (`src/core/run-analyze.ts:1564`) — by design it CHECKPOINTs and leaves the native handles **open**, relying on a guaranteed `process.exit`. The CLI then runs `await assertAnalysisFinalized(repoPath)` at `src/cli/analyze.ts:1360`, which is **not** in a local try/catch and **does throw** in a documented, observed state (meta.json / registry missing — see the comment at `analyze.ts:1353-1359`, *"on Windows it has been observed to return successfully with neither artifact present"*). The throw hits the outer catch (~`analyze.ts:1464`): `process.exitCode = 1; return` — a **soft return with no `process.exit`**. With the native handles still open, Node's event loop never drains → the process **hangs** instead of exiting 1.

This is a **regression**: pre-PR the finalize did a real `closeLbug()` that released the handles, so the same throw soft-returned cleanly. It is the **same hang class as the already-fixed error-path P1** — the success-then-post-work window was missed. The `alreadyUpToDate` fast path is **safe** (it never opens the DB — comment at `1335-1336`).

Corroboration: correctness (P1) + adversarial (P1), **verified by coordinator code-read** `[code-read]`. Trigger: a full `analyze` where the run succeeds (skip-closes) but post-finalize verification throws (or any throw from post-work before `process.exit(0)` at `1631`).

**Fix options:** force `process.exit(1)` in the outer catch when a skip-close occurred; or do a real `closeLbug()` in the catch before the soft return; or don't pass `skipNativeCloseOnExit` until after the throwing post-work.

### ✅ Validated (actively checked — credit)
- **The error-path hang from the prior review is correctly fixed** — `run-analyze.ts:1592` does a REAL close on error, so a soft-returning caller still releases handles.
- **`withConnLock` mutex is correct** — FIFO, release-on-throw, AsyncLocalStorage re-entry guard; coverage of singleton-`conn` write paths is comprehensive.
- **CHECKPOINT-only durability holds** — re-open after skip-close reads clean (coordinator-verified end-to-end).
- **No leak in the long-lived MCP/server path** — `skipNativeCloseOnExit` is set by exactly two exit-guaranteed callers (CLI success, forked worker); MCP analyzes via a forked child (adversarial refuted the leak).
- conn-lock re-entry guard, the #2203 node-COPY‖rel-emit overlap, and the `streamQuery` exemption are sound for current code.

### Lower-priority (body, non-blocking)
- **P2** `loadCachedEmbeddings` (`lbug-adapter.ts:1673/1685/1693`) issues raw `conn.query` outside the lock — safe only because it runs *before* the WAL driver starts; a sequencing invariant not enforced by code (risk lane).
- **P2** the error-path real close after large `--pdg` writes could itself abort (SIGABRT) → degrades a late error from an actionable exit-1 message to a raw native abort (data still durable). Accepted tradeoff; a guarded checkpoint-then-skip there would preserve the message.
- **P2** maintainability: the `skipNativeClose` / `skipNativeCloseOnExit` invariant ("only when a `process.exit` follows") is enforced by **convention** across 3 call sites, not by type — exactly the gap the P1 exploits. The structural test greps source/comments for `conn.close()`/`db.close()` literals (this PR had to reword a comment to pass) — brittle; an ESLint `no-restricted-syntax` rule would be sturdier.
- **P3** rel COPY passes `conn` (not the captured `writeConn`) at `lbug-adapter.ts:1139`; worker IPC `send()` throwing on a closed channel could skip the exit (narrow); SIGINT during a long COPY now pauses until the lock frees.

### Test gaps
No test exercises the success-then-finalize-failure hang (the P1), the `skipNativeCloseOnExit` threading (CLI/worker set it; error path must NOT), or asserts the error-path real-close. The native crash itself isn't directly testable (would kill the vitest worker) — acceptable; the fix is validated indirectly + end-to-end manually.

### CI
At review time: **0 failures** (18 pass, ~7 pending).

*Automated multi-tool digest (5 Claude lanes + coordinator code-read; Codex async-pending, not an independent confirmation). Verify against the code before acting.*

gitnexus/src/cli/analyze.ts
magyargergo: Fixed — adopted the unified **"both paths skip-close + the CLI guarantees the exit"** design (root invariant: *skip-close requires a guaranteed `process.exit`*):

- **`8130e625`** — the outer `analyzeCommand` wrapper now runs `if (isLbugReady()) process.exit(...)` *after* its try/finally, so **every** soft error-return (the `assertAnalysisFinalized` throw at `analyze.ts:1360` plus the other soft-returns in the outer catch) force-exits while LadybugDB handles are still open. `isLbugReady()` = `conn !== null && db !== null`, so it's true only after a skip-close left handles open — and false for early errors *and* in the `analyze-*` unit tests that mock `runFullAnalysis` (the DB is never opened), so their soft-return path is preserved.
- **`047530d0`** — the error path now also skip-closes (CHECKPOINT-then-skip) instead of a real `closeLbug()`, so the outer catch can't itself hit the `ClientContext` destructor double-free after `--pdg` writes. This also resolves the P2 "error-path real close could abort" from the body.

New regression test `test/unit/analyze-finalize-failure-exits.test.ts` asserts the skip-closed path **terminates** (force-exits) on a finalize failure rather than hanging, and that it does **not** force-exit when handles are closed (soft-return preserved).

This supersedes the previously-validated real-close-on-error — rationale in the summary comment below.

magyargergo:
## Round-3 fixes for the tri-review (review 4539386053) — pushed (`8a6799c5..e56ddd10`)

Each finding addressed as its own commit (planned with `ce-plan`, executed with `ce-work`):

| Finding | Commit(s) | What changed |
|---|---|---|
| 🔴 P1 — success-path hang | `8130e625` + `047530d0` | unified skip-close + CLI-guaranteed exit (details in the inline reply) |
| P2 — error-path real close could abort | `047530d0` | error path now CHECKPOINT-then-skip (no real close after `--pdg` writes) → the actionable exit-1 message survives instead of degrading to a SIGABRT. This is the "guarded checkpoint-then-skip" suggested in the body |
| P2 — `loadCachedEmbeddings` raw `conn.query` | `b96a55fb` | the 3 reads now run inside `withConnLock` (capturing `const c = conn`) — no longer leans on the unenforced "runs before the WAL driver starts" sequencing |
| P2 — brittle structural test (greps comments) | `c2275c81` | the test strips comments before grepping the adapter source, so rewording a comment can no longer trip it (the cited root cause). I kept the structural guard rather than swapping to an ESLint `no-restricted-syntax` rule — happy to add that as a follow-up if you'd prefer the sturdier form |
| P3 — rel COPY uses `conn` not `writeConn` | `17b0b93e` | passes the already-captured `writeConn`, matching the node COPY |
| P3 — worker IPC `send()` could skip the exit | `caa8543b` → `e56ddd10` | the worker reports every error to the parent over IPC and schedules `process.exit` in a `finally`, so a `send()` throw on a closed channel can't skip the exit — and nothing is swallowed (SIGTERM cleanup failures are reported too, not `catch {}`-dropped) |
| P3 — SIGINT pauses behind a long COPY | `1c7767f9` | the SIGINT cleanup CHECKPOINT is bounded by a `Promise.race` timeout so a single Ctrl-C stays responsive |
| Test gap — success-then-finalize hang | `8130e625` | new `test/unit/analyze-finalize-failure-exits.test.ts` |

**Design note (P1 + error-path P2).** I adopted the recommended unified design — *both* the success and error paths skip-close, and the CLI guarantees the exit (`if (isLbugReady()) process.exit(...)`). This **supersedes the previously-validated real-close-on-error**: that real close was itself capable of the `ClientContext` destructor double-free after large `--pdg` writes (the P2 above), so replacing it with CHECKPOINT-then-skip + a forced exit fixes the hang *and* removes the abort — with no unit-test breakage, since the `analyze-*` tests mock `runFullAnalysis` so `isLbugReady()` stays false and their soft-return is preserved.

**Verification.** `tsc` clean; the affected vitest files pass (default + `lbug-db` projects, run in isolation per the pre-existing `lbug-core-adapter` flake); and a local `analyze --pdg --force` on this repo after `npm run build` now completes (exit 0, ~194k nodes) where it previously SIGABRT'd at "Saving metadata…".

I'll file an upstream LadybugDB bug separately for the `ClientContext` destructor double-free on `conn.close()` (gdb-confirmed) that this PR works around.

gitnexus/src/core/run-analyze.ts
```diff
@@ -768,7 +778,21 @@ export async function runFullAnalysis(
...
+      // through to the pipeline (which registers it, honoring the flag) instead
+      // of early-returning an unregistered repo the flag could never heal.
+      // For a PLAIN analyze we deliberately do NOT self-heal: an up-to-date but
+      // unregistered repo early-returns here and the CLI's assertAnalysisFinalized
+      // surfaces it as a hard failure (#1169) rather than silently registering a
+      // possibly half-finalized index. `isRepoRegistered` is only read on the
+      // opt-in branch so the common fast path keeps its single-stat cost.
+      const healUnregistered =
+        options.allowDuplicateName === true && !(await isRepoRegistered(repoPath));
+      if (!dirty && !healUnregistered) {
```
magyargergo, line 796: **[P2] Up-to-date-but-unregistered repo is reported `complete` on the server/worker path.** `[code-read]` (Codex AREA G + coordinator-verified)

This `if (!dirty && !healUnregistered)` returns `alreadyUpToDate: true` without registering for a plain analyze (`healUnregistered` is true only with `--allow-duplicate-name`). The CLI catches the unregistered state via `assertAnalysisFinalized` (analyze.ts:1317/1378), but the forked worker (`src/server/analyze-worker.ts:128`) sends `{type:'complete'}` **without** that check — so a server/web analyze of a half-finalized repo (meta.json present, registry entry missing — e.g. after a prior collision-aborted run or a wiped registry) is reported successful while the repo stays unregistered/invisible to `list_repos`.

The comment just above asserts "the CLI's `assertAnalysisFinalized` surfaces it as a hard failure (#1169)" — true only on the CLI; the worker path has no equivalent.

*Fix:* call `assertAnalysisFinalized(repoPath)` in `analyze-worker.ts` before sending `complete`, or move the unregistered-fast-path invariant into `runFullAnalysis` so both callers share it.

Trigger: a repo with up-to-date `meta.json` but no registry entry, analyzed via the server worker (default `allowDuplicateName:false`).

gitnexus/src/server/analyze-worker.ts
```diff
@@ -44,27 +44,55 @@ export interface ErrorMessage {
...
+// SKIPS the native close. A real conn.close()/db.close() here can double-free in
+// LadybugDB's ClientContext destructor after --pdg writes, AND it would block
+// behind the in-flight COPY's connection lock — so a single cancel could abort
+// the worker or hang until the COPY releases. Bound the CHECKPOINT with a short
+// timeout; process.exit reclaims the native handles regardless. A CHECKPOINT
+// failure is reported to the parent over IPC, not swallowed; the exit lives in
+// `finally` so it always fires.
+const SIGTERM_CLEANUP_TIMEOUT_MS = 2000;
+process.on('SIGTERM', () => {
   send({ type: 'error', message: 'Analysis cancelled (worker received SIGTERM)' });
```
magyargergo: **[P3] SIGTERM cancellation doesn't coordinate with completion.** `[code-read]` (Codex AREA E + ce-adversarial corroborated)

This leading `send({type:'error', ...})` runs unconditionally and the handler doesn't coordinate with the message handler that sends `{type:'complete'}` (line 128). A cancel near the finish line races: `complete` can still win (a cancelled job reported complete), or this `error` can land after a `complete` and flip the job to `failed`. The parent's `message` handler (`analyze-launch.ts:83-108`) has no terminal-state guard — unlike its sibling `exit` handler (line 121) — and `updateJob` (`analyze-job.ts`) does `Object.assign(job, update)` before its `isTerminal` check (which only stamps `completedAt`), so a post-terminal message overwrites the final status.

*Fix:* set a `terminating` flag here that the message handler checks before sending `complete`, or have the parent ignore IPC messages once the job is terminal.

(Status-reporting race only — not lock corruption: `cancelJob` doesn't release the repo lock and the parent's `Set`-based release is idempotent.)

magyargergo (commented):
## Tri-review (full re-review) — PR #2264: LadybugDB `--pdg` close-crash fix + connection serialization

> Supersedes the two earlier tri-reviews on this PR (4539255441, 4539386053) — those were COMMENT reviews and can't be API-dismissed; this is the current authoritative pass over the full diff against `main`.

**Methods / engines.** 5 Claude reviewer lanes (correctness, adversarial, performance, maintainability, testing) + **Codex (independent engine, live this run)** + coordinator code-read verification. Two GitNexus lanes (risk, test/CI) ended mid-investigation — their domains are covered by the adversarial/correctness and testing lanes. The **strong** signals below are *Codex + a Claude lane agreeing*; everything else is coordinator-verified by re-reading the code (`[code-read]` — worktrees can't `npm run build`, so nothing is `[reproduced]` here except CI, which is green).

**Credit — the core is sound.** Codex (AREAs A–D, F) and the correctness + adversarial lanes independently **refuted** the scariest failure modes: the `targetConn === conn` lock gate does not misfire (`conn` is reassigned only at open/close, never mid-load, and `writeConn` is captured once); the `AsyncLocalStorage` re-entry guard has no production nesting path (and it *throws*, never silently no-ops, if one is added — the delete-helpers take sequential, not nested, locks); no `skipNativeClose` path returns to a long-lived process without a guaranteed `process.exit`; the WAL-driver `stop()`/`inflight` interaction is race-free; and CHECKPOINT-then-skip-close is durable (no pending COPY/txn/adapter buffer a successful CHECKPOINT misses). **Performance:** the `withConnLock` serialization adds negligible overhead because the analyze workload is already serial on the event loop — the "don't degrade performance" constraint holds. The `conn-lock.test.ts` and `wal-checkpoint-driver-reentrancy.test.ts` suites are genuinely strong (FIFO, release-on-throw, re-entry, fake-timer inflight coverage).

### 🟠 P2 — the server/worker path can report `complete` for an up-to-date-but-unregistered repo  *(inline)*
`run-analyze.ts:795` returns `alreadyUpToDate` **without registering** for a plain analyze (`healUnregistered` is only true with `--allow-duplicate-name`). The CLI catches this via `assertAnalysisFinalized` (analyze.ts:1317/1378), but `analyze-worker.ts` sends `{type:'complete'}` (line 128) with **no** such check — so a server/web analyze of a half-finalized repo (meta.json written, registry entry missing — e.g. after a prior collision-aborted run, or a wiped registry) is reported **successful** while the repo stays unregistered/invisible to `list_repos`. The PR's own `run-analyze.ts` comment asserts `assertAnalysisFinalized` surfaces this as a hard failure (#1169) — but that holds only on the CLI. Surfaced by **Codex** (AREA G); coordinator-verified `[code-read]`. Pre-existing, but directly adjacent to this PR (it codifies the invariant the worker path lacks). *Fix:* have the worker call `assertAnalysisFinalized(repoPath)` before sending `complete`, or move the unregistered-fast-path invariant into `runFullAnalysis` itself.

### 🟡 P3 — SIGTERM cancellation doesn't coordinate with completion  *(inline)*
The worker `SIGTERM` handler (analyze-worker.ts:87) unconditionally sends `{type:'error','Analysis cancelled'}` and doesn't coordinate with the message handler that sends `complete` (line 128). Near the finish line this races: a cancel can still let `complete` win (a **cancelled job reported complete**), or the leading `error` can land after a `complete` and flip the job to `failed`. The parent's `message` handler (analyze-launch.ts:83-108) has **no terminal-state guard** — unlike its sibling `exit` handler (line 121) which checks `status === 'complete' || 'failed'` — and `updateJob` (analyze-job.ts) runs `Object.assign(job, update)` *before* its `isTerminal` check (which only stamps `completedAt`, it does not block a status overwrite). So a post-terminal IPC message overwrites the final status. Flagged independently by **Codex (AREA E) and the adversarial lane**; coordinator-verified `[code-read]`. Pre-existing in shape, but the PR rewrote this handler (#2264). *Fix:* set a `terminating` flag the message handler checks, or have the parent ignore messages once the job is terminal. *(Note: this is a status-reporting race, not lock corruption — `cancelJob` doesn't release the repo lock, and the parent's `Set`-based release is idempotent.)*

### Lower-priority (body, non-blocking)
- **Test gaps.** The lock-routing suite (`lbug-conn-serialization.test.ts`) asserts 4 helpers but **not** `deleteAllInterprocTaintPaths` / `deleteAllCallSummaries`, which are also `withConnLock`-wrapped (lbug-adapter.ts:2130 / 2189) and reachable in the driver window — reverting either wrapper wouldn't be caught (the file header claims "every singleton-conn helper"). `analyze-finalize-failure-exits.test.ts` asserts `process.exit(1)` but not that a pre-set `process.exitCode` is *forwarded* (the production code is `process.exit(process.exitCode ?? 1)`). No test for the **negative** gate (a temp-conn `deleteNodesForFile` must NOT lock) or the SIGINT/SIGTERM bounded `Promise.race`.
- **Maintainability (design, not bugs).** The `skipNativeClose` "only safe before a guaranteed `process.exit`" invariant is enforced by **convention** across ~4 call sites, not by type — a dedicated `closeLbugBeforeExit()` (no flag) would make misuse unrepresentable. The `targetConn === conn` object-identity gate is duplicated (queryAndDrain + deleteNodesForFile); the re-entry rule lives only in prose + a runtime throw. The SIGINT (analyze.ts) and SIGTERM (analyze-worker.ts) bounded-cleanup `Promise.race` blocks are near-identical and could share a helper.
- **Comment-only invariants (3 lanes).** `streamQuery` (deliberately unwrapped) and the node-COPY ‖ rel-emit overlap are safe **today** only because the server forks analyze workers and the rel-emit pass never touches `conn`; a future *in-process* analyze inside the API server would re-open the corruption window with no runtime guard. Worth a guard or an invariant test, not a blocker.
- `closeLbug({skipNativeClose})` skips `finalizeLbugSidecarsAfterClose` (the real `safeClose` runs it). Codex rated durability **sound**, so this is low-risk WAL/sidecar housekeeping the next run reconciles — a one-line note (or running it on the skip path) would close the gap.

### CI
All green at review time (28 pass, 1 skip). The `--pdg` crash fix and a green ubuntu-coverage lane were verified end-to-end by the coordinator earlier in the session.

*Automated multi-tool digest (5 Claude lanes + live Codex + coordinator code-read). No P0/P1 — the headline items are a P2 and a P3, both pre-existing-but-PR-adjacent. Verify against the code before acting.*

koriyoshi2041:
Did an independent pass on the latest head (`bb0c592`). The important invariant I wanted to check was the skip-close path: `analyzeCommand()` now force-exits if a post-finalize soft-return leaves LadybugDB ready, and the forked worker/SIGTERM paths also schedule exit in `finally`, so I don't see the earlier hang class remaining.

Local validation:
- `cd gitnexus && npm test -- --run test/unit/conn-lock.test.ts test/unit/wal-checkpoint-driver-reentrancy.test.ts test/unit/analyze-finalize-failure-exits.test.ts test/unit/lbug-checkpoint.test.ts test/integration/lbug-conn-serialization.test.ts` -> 28 passed
- `cd gitnexus-shared && npm run build` -> passed
- `cd gitnexus && npx tsc --noEmit` -> passed

Only tiny non-blocking thing I noticed: the comment inside `closeLbug({ skipNativeClose })` still says the runFullAnalysis error path intentionally does not pass this, but the current design does pass it from CLI/worker callers and relies on the caller-side forced exit. The behavior looks consistent; just a stale comment if you touch this again.

gitnexus/src/core/run-analyze.ts
magyargergo: Fixed in **`1565e14f`** — the forked worker now calls `assertAnalysisFinalized(repoPath)` before sending `complete`, mirroring the CLI, so a half-finalized (indexed-but-unregistered) repo is reported as `{type:'error'}` instead of a false success. The run→finalize→report contract was extracted into a side-effect-free `analyze-worker-core` seam (the entry module's top-level `process.on` handlers make it untestable directly) and unit-tested with injected fakes (`analyze-worker-core.test.ts`). Kept it in the worker rather than moving the throw into `runFullAnalysis` — that would change the `alreadyUpToDate` contract for all callers; `assertAnalysisFinalized` is already the shared single-source check.

gitnexus/src/server/analyze-worker.ts
magyargergo: Fixed in **`7e1e5828`** (worker) + **`717359c7`** (parent, defense-in-depth):
- Worker: a single terminal-outcome claim (`createTerminalClaim`) shared by the message handler and the SIGTERM handler — whoever claims first reports its `complete`/`error`, the other skips its terminal send (single-threaded JS makes the check-and-set atomic). Cleanup + exit still run regardless.
- Parent: the launcher message handler now returns early once the job is terminal (mirroring its sibling `exit` handler), and `analyze-job.ts` `updateJob` refuses transitions out of a terminal state — so a late IPC message can't re-release the repo lock or flip the reported status.

Both verified by unit tests (`analyze-worker-core.test.ts` claim-coordination, `analyze-job.test.ts` terminal-immutability).

magyargergo:
## All tri-review (4539713542) findings fixed — pushed (`bb0c5920..a11c4c95`)

Planned with `ce-plan` (deepened), executed with `ce-work` — **each finding its own commit**, all green (tsc clean; 149 unit tests across the touched files + 9 lbug-db integration tests in isolation).

| Finding | Commit(s) | What changed |
|---|---|---|
| 🟠 **P2** — worker reports `complete` for an unregistered repo | `1565e14f` | worker calls `assertAnalysisFinalized` before `complete` (via a testable `analyze-worker-core` seam) |
| 🟡 **P3** — SIGTERM cancellation vs completion race | `7e1e5828` + `717359c7` | shared `createTerminalClaim` slot (worker) + parent-side terminal-immutability guard (`analyze-launch` early-return, `updateJob` refuses post-terminal transitions) |
| Test gap — 2 delete helpers' lock routing | `1a7aefdc` | assert `deleteAllInterprocTaintPaths`/`deleteAllCallSummaries` route through `withConnLock` |
| Test gap — negative gate | `cdc3f934` | assert temp-conn `deleteNodesForFile` does **not** lock |
| Test gap — exitCode forwarding | `8ee17efc` | assert the force-exit forwards `process.exitCode` (not a hardcoded 1) |
| Maintainability — `skipNativeClose` convention | `4702bee8` | dedicated `closeLbugBeforeExit()` replaces the boolean flag; foot-gun removed |
| Maintainability — `targetConn === conn` dup | `2e535383` | `isSharedSingletonConn(c)` predicate |
| Maintainability — SIGINT/SIGTERM dup | `aa7003da` | shared `boundedCheckpointBeforeExit` helper |
| Maintainability — registry case-fold dup | `ed57b3b0` | `registryPathEquals(a, b)` (6 sites consolidated) |
| Hardening — `streamQuery` comment-only invariant | `4449bb4a` | runtime guard: driver-active flag → `streamQuery` throws if it would race a CHECKPOINT |
| Hardening — skip-close skips sidecar finalize | `a11c4c95` | documented the deliberate deferral (running it with handles open risks a Windows file-lock; next run's `preflightLbugSidecars` reconciles) |

**Notes:** `717359c7` deliberately touches the parent-side `analyze-launch.ts` / `analyze-job.ts` (outside the original PR diff) as the robust complement to the worker fix — kept as its own clearly-scoped commit. The worker-test seam was necessary because `analyze-worker.ts` is a `fork()` entry module (top-level `process.on` handlers) and can't be imported directly in a unit test.

Refuted items from the review (lock gate, re-entry guard, skip-close-without-exit, WAL-driver stop/inflight, CHECKPOINT durability) were left as-is — they were validated, not defects.

PR #2234, opened 2026-06-17
feat(routes): persist HTTP method on Route nodes (#2138 part 1/2)

henry201605:
## Summary

**This is part 1 of a two-PR plan for issue #2138** (*Optimize HttpRouteExtractor: skip source-scan for languages with full ingestion Route-node coverage*).

- **Part 1 (this PR) — data completeness:** persist the HTTP verb on `Route` graph nodes so the contract extractor can read it from the graph instead of re-parsing handler source.
- **Part 2 (follow-up PR) — the actual optimization:** link `HANDLES_ROUTE` to the handler *symbol* (not just the file) so `extractProvidersGraph` no longer needs `getDetections()` at all; only then can fully graph-covered files skip the source-scan **and its parse**, which is what #2138 ultimately asks for. That PR will ship a parse/scan call-count benchmark proving the reduction.

### Why split it this way

The earlier attempt at #2138 (#2167) skipped source-scan emission at the merge layer and was correctly closed because the win was ~zero: every handler file is already parsed up front (`collectProjectDetections`) **and** re-parsed by the graph path's `getDetections()` backfill. The parse is the cost, and it cannot be removed until the graph path stops depending on source scanning. That dependency has two halves — **method** and **handler name**. This PR removes the *method* half; Part 2 removes the *handler-name* half. Part 1 is deliberately **not** framed as a perf PR.

## What this PR changes

The ingestion routes phase already computes each route's verb (`ExtractedRoute.httpMethod` for Spring/Laravel framework routes, `ExtractedDecoratorRoute.httpMethod` for decorator routes) but dropped it when creating the `Route` node. For `framework-route` sources the edge `reason` is undecodable by `methodFromRouteReason`, so the graph path previously had to re-scan source to recover the verb.

- **routes phase:** carry `httpMethod` into `RouteEntry` and persist it as `Route.method`. Filesystem-derived routes (Next.js/Expo/PHP file routes) have no structural verb and stay method-less.
- **HttpRouteExtractor:** `HANDLES_ROUTE` query returns `route.method`; `extractProvidersGraph` prefers it and falls back to the edge reason when absent — **fail-open and fully backward compatible** with indexes built before this change.

## Validation

- `npx tsc --noEmit` — no new errors introduced (pre-existing unrelated errors aside).
- New unit tests in `test/unit/group/http-route-graph-method.test.ts`: graph-method precedence over an undecodable reason, multi-verb handler disambiguation via the persisted verb, case normalization, and old-index fallback.
[...]

PR #2167, opened 2026-06-11
perf(group): skip redundant HTTP provider fallback

azizur100389:
## Summary
- mark Java, Python, and PHP HTTP plugins as having complete ingestion Route coverage
- skip source-provider emission only when graph contracts resolve and cover every cached source provider key
- preserve source fallback for empty, unresolved, and mixed graph/source coverage
- add a regression for a graph-covered file that still contains a source-only route

## Validation
- \
px tsc --noEmit\`n- \
px vitest run test/unit/group/http-route-extractor.test.ts --maxWorkers=2\ — 91 passed
- full suite: 10,922 passed; the sole failing literal-collector assertion is reproducible on unchanged main

Fixes #2138

PR #2078, opened 2026-06-08
feat(ingestion): Java Spring route annotation → Route node extraction

henry201605:
## Summary

Adds ingestion-layer support for Java Spring annotation-based routing, enabling `route_map`, `shape_check`, and `api_impact` MCP tools to work on Java Spring projects.

**Before this PR:** Route node count for all Java projects = 0. The group contract layer (`http-patterns/java.ts`) supported Spring annotations for cross-repo HTTP contract matching, but the ingestion pipeline did not generate graph Route nodes from them.

**After this PR:** Spring `@GetMapping`, `@PostMapping`, `@PutMapping`, `@DeleteMapping`, `@PatchMapping` annotations on methods generate Route nodes. Class-level `@RequestMapping` is correctly joined as a URL prefix, with per-class resolution for multiple controllers in one file.

## Implementation

Route extraction is wired through the `LanguageProvider` contract rather than language-specific branches in the shared parse worker:

### `route-extractors/spring.ts` (new)
The ingestion-layer Spring extractor. Uses a single predicate-free tree-sitter query to capture all route annotations, then a two-phase loop: pass 1 collects class-level `@RequestMapping` prefixes keyed by `class_declaration` node id; pass 2 resolves each method route's prefix via `findEnclosingClass`. Handles positional and named (`path=` / `value=`) arguments, and multiple classes per file with independent prefixes.

### `language-provider.ts` + `languages/java.ts`
Added an optional `extractDecoratorRoutes` hook to `LanguageProviderConfig`. The Java provider registers `extractSpringRoutes`. The parse worker calls `provider.extractDecoratorRoutes?.()` generically — no language names in the shared dispatch path.

### `route-extractors/spring-shared.ts` (new)
Shared primitives used by both the ingestion extractor (`spring.ts`) and the group extractor (`http-patterns/java.ts`) so the two can't drift apart: `METHOD_ANNOTATION_TO_HTTP`, `findEnclosingClass`, `isRouteMemberKey`, and a safe `unquoteSpringLiteral`. Lives in `ingestion/` (lower layer); group imports from it, matching the existing `group → ingestion` direction.

## Tests

- **`test/unit/java-spring-route-ingestion.test.ts`** (9 cases): prefix joining, bare routes, all five HTTP verbs, `@PatchMapping`, named args, multi-class isolation, `produces`/`consumes` filtering, line numbers.
[...]

PR #2265, opened 2026-06-21
perf(group/http): skip source parse for graph-covered route files (#2138 Part 2)

### LINKED ISSUE #2138: Optimize HttpRouteExtractor: skip source-scan for languages with full ingestion Route-node coverage
## Context

Split out from #2078 (Java Spring route ingestion).

`HttpRouteExtractor.extract()` always runs two paths and merges them:
- `extractProvidersGraph` — reads Route nodes from the indexed graph
- `extractProvidersSourceScan` — re-scans every source file with the language plugin

The source-scan path exists to capture routes in languages/files **not** covered by graph edges (see the comment at `http-route-extractor.ts:~285`).

## Problem

Now that ingestion emits Route nodes for Java Spring (#2078), the graph path already finds those routes, so the source scan re-discovers the same routes and `mergeGraphAndSourceContracts` dedupes them. The per-file tree-sitter parse is cached, but the plugin's annotation-matching loop still runs twice per Java file — pure redundant work that can never add information the graph path doesn't already have.

This is wasteful, not incorrect — #2078 is correct as-is. But it's worth removing.

## Proposed direction

When ingestion reliably emits Route nodes for a language, `group sync` could skip (or narrow) the source scan for files already fully covered by the graph — e.g. pass the graph coverage set into `extractProvidersSourceScan` and skip files whose graph Route nodes all have resolved handlers.

## Scope / caveats

- Affects any language with full ingestion Route coverage — **Java, Python/FastAPI, PHP/Laravel** — not just Java.
- Must handle **partial graph coverage** (a file with some graph routes but also source-only routes) without dropping the source-only ones.
- Changing the graph/source-scan fallback is larger than #2078's scope, hence this separate issue.

henry201605:
## What & why

Part 2 of #2138 (Part 1 = #2234, persisting `Route.method`, merged).

`HttpRouteExtractor` parses every source file with tree-sitter up front, even
when the indexed graph already holds the file's routes. Now that ingestion
emits `Route` nodes with a resolved handler symbol (U0–U3a in this branch),
the graph is authoritative for a fully-covered file's **providers** — so for
those files we can skip the source scan *and* the parse. This is the
measurable parse reduction #2167 was asked for and couldn't demonstrate (it
skipped only the merge-layer emission, which saved ~0 because the parse still
ran).

## How

- `HttpLanguagePlugin` gains `routeCoverage?: 'complete' | 'partial'`
  (default `'partial'`). Marked `'complete'` for Java/Python/PHP, whose
  provider routes are fully emitted as `Route` nodes.
- `extract()` runs the graph provider pass first to build a `coveredFiles`
  set (every `HANDLES_ROUTE` row for the file resolved a handler symbol +
  language is `'complete'`), then excludes those files from
  `collectProjectDetections` and both source scans.
- **Fail-open per file**: any unresolved row, a `'partial'` language, or a
  graph-less run leaves the file in the scan set — no route is dropped.

## Consumer safety (the subtle part)

`routeCoverage: 'complete'` asserts **provider** Route-node completeness only.
But the same languages' `scan()` also detects **consumers** (RestTemplate /
WebClient / OkHttp / Feign / Spring HTTP Interface `@*Exchange`, Guzzle /
`Http::`, `requests` / `httpx`), and ingestion's `FETCHES` edges are JS/TS-only
— the graph can't back up server-side consumers. A provider-covered controller
that also calls out would otherwise lose its consumer contract.

So skipping is gated by a cheap, **parse-free** text check:
`HttpLanguagePlugin.hasConsumerSignals(content)` returns `false` only when the
raw source provably has no outbound-HTTP call this plugin detects. A file stays
covered (parse-skipped) only when this is `false`; a positive signal, a missing
hook, or an unreadable file keeps it in the scan set. The orchestrator names no
languages — token knowledge lives in the plugins.

Net: pure-provider controllers skip the parse (the win); controllers that also
call out are still parsed (no consumer loss).

## Testing

- New `route-parse-skip` integration test spies the real `parseSourceSafe` to
  **count parses** over a temp repo of Spring controllers + a mock DB:
  baseline (every file parsed), fully-covered (**0 parses**), mixed
  (unresolved file falls back, resolved stays skipped), provider+consumer (a
  covered controller calling RestTemplate is parsed; its consumer contract
  survives).
- `tsc --noEmit` clean; group/route/http-patterns/contract/csv/schema suites
  green (955 tests); `emit-persistence --check` PASS (no CSV column change).

## Notes

- Rebased on current `main`, including #2254 (Kotlin consumer extraction +
  Java parity). Java's `hasConsumerSignals` covers the new Spring HTTP
  Interface `@*Exchange` consumer idiom that #2254 added.
- **Kotlin is intentionally left `'partial'`** (never parse-skipped, so its
  consumers are always scanned — safe). Marking Kotlin `'complete'` for the
  same parse-skip win is a natural follow-up, pending confirmation that
  ingestion emits `Route` nodes for Kotlin Spring providers.
- The `hasConsumerSignals` regexes are deliberately broad (over-matching only
  costs a parse, never drops data). Happy to tighten if you'd prefer.


gitnexus/src/core/group/extractors/http-patterns/java.ts
```diff
@@ -619,6 +619,21 @@ function scanSpringProject(files: readonly HttpScanInput[]): HttpFileDetections[
 export const JAVA_HTTP_PLUGIN: HttpLanguagePlugin = {
   name: 'java-http',
   language: Java,
+  // Spring @(Get|Post|...)Mapping providers are emitted as Route nodes by the
+  // ingestion routes phase (#2078), so the graph is authoritative for Java
+  // providers — see HttpRouteExtractor's parse-skip (#2138 Part 2).
+  routeCoverage: 'complete',
```
magyargergo: **[P1 — merge-blocking] `routeCoverage: 'complete'` over-claims for Java → parse-skip silently drops provider routes.**

This asserts ingestion emits a Route node for *every* provider the group `scan()` detects, but the graph set is a strict subset. Three skip-induced data-loss cases:
1. **Array-form** `@GetMapping({"/a","/b"})` — the group query (`java.ts:124`) matches `element_value_array_initializer`; ingestion's `spring.ts` route queries (`:51,59,64,72`) match `(string_literal)` only.
2. **Interface-inherited** Spring routes — `scanSpringInheritanceProject` (`spring-consumer-shared.ts:197-268`) composes them, but ingestion only walks class-direct annotations.
3. **Same-URL multi-verb** (GET+POST `/orders`) — the Route node is keyed by URL only, so the verbs collapse to one row.

Trigger: a pure-provider `@RestController` with one simple resolvable `@GetMapping("/x")` **plus** any of the above and no consumer calls — the file is marked covered (all rows resolved) and removed from the scan set, so the group-only routes vanish. Existing passing group tests prove these shapes are expected (`http-route-extractor.test.ts:1850` array, `:863` inherited).

Fix: keep Java `routeCoverage: 'partial'` until ingestion matches the group scan (array-form + interface inheritance + per-verb Route identity), **or** run `scanProject` over the full file set and only suppress per-file provider *emission* for covered files. Add regression tests for all three skip cases.

_Cross-engine: Codex swarm + Codex ce + Codex adversarial + Claude ce-adversarial/correctness. Coordinator-verified the array-form code asymmetry directly. [code-read]_

gitnexus/src/core/ingestion/call-processor.ts
```diff
@@ -243,6 +245,68 @@ export const processRoutesFromExtracted = async (
...
+    const methodId = model.symbols.lookupExactAll(controllerDef.filePath, route.methodName)[0]
+      ?.nodeId;
+    put(route.routePath, route.prefix ?? null, methodId);
+  }
+
+  // Decorator routes (Spring / FastAPI / generic) — the decorated handler in
+  // the route's own file.
+  for (const dr of decoratorRoutes) {
+    if (!dr.handlerName) continue;
+    const handlerId = model.symbols.lookupExactAll(dr.filePath, dr.handlerName)[0]?.nodeId;
```
magyargergo: **[P2] `handlerSymbolId` can resolve to the wrong handler.**

`lookupExactAll(dr.filePath, dr.handlerName)[0]` (and the Laravel path at `:294`) takes the *first* def of a same-named symbol — an overloaded/same-named handler resolves to an arbitrary, possibly wrong UID, even though `spring.ts` captured the exact decorated node and then discarded that identity.

Compounding it: `resolveRouteHandlerSymbols.put` is first-writer-per-URL but *skips* unresolvable URLs, while the routes phase `addRoute` is first-writer-wins for the Route node. On a same-URL collision where the node-winning route is unresolvable but a later same-URL route resolves, the winning node (`routes.ts:288`) gets stamped with the *other* route's handler.

Fail-open prevents outright route loss, so this is P2 — but impact analysis / contract handlers then point at the wrong symbol. Fix: disambiguate by the captured handler node's line/position (or skip to fail open) instead of `[0]`.

_Cross-engine: Codex swarm + Codex ce + Claude ce-correctness/adversarial. [code-read]_

magyargergo (commented):
# PR #2265 review — `perf(group/http): skip source parse for graph-covered route files (#2138 Part 2)`

**Automated two-engine tri-review digest.** Verdict: **not production-ready as written** — one merge-blocking P1 data-loss class. The architecture is sound and fail-open; the blocker is a single over-claimed invariant (`routeCoverage: 'complete'` for Java) that the optimization trusts but ingestion does not actually satisfy.

## Engines & how to read this

- **Engine A — Claude.** GitNexus PR swarm (pr-facts, security-boundary complete; risk-architect ended mid-investigation, its domain fully covered by the ce lanes) + ce-code-review personas (correctness, adversarial, api-contract, testing, maintainability, performance, project-standards — all complete).
- **Engine B — Codex (gpt-5.5 / xhigh).** All three legs ran: swarm Solo mode (**B1, complete** — HIGH risk, "not production-ready") + ce-code-review (**B2, complete**, verdict "Not ready", 6 findings; its `correctness`/`adversarial`/`project-standards` ce sub-lanes were degraded by stream disconnects, so only the `main` + `api-contract` + `testing` sub-lanes emitted) + native **adversarial-review (B3, complete** — verdict "needs-attention / No ship", 2 high findings).

This is **two engines**, not five independent reviews. The strongest signals below are where a Codex lane and a Claude lane independently reached the same finding. Where a lane's severity disagrees with the coordinator's own re-read of the code, this digest follows the coordinator calibration and says why.

**CI:** MERGEABLE / **UNSTABLE** — the only red is Vercel deploy-auth (team-authorization infra), the same pattern as merged #2254 / #2138-Part-1. All 28 substantive checks pass (coverage 13221/13280, macOS/Windows, CodeQL, gitleaks, typecheck/lint/format). **Vercel is not a code blocker.** Status is UNSTABLE because of infra, not code.

## What's solid (credit where due)

- **Fail-open architecture is genuinely robust.** Every degraded path (URL-key mismatch, unresolved handler, unreadable file, missing plugin, old-schema query) defaults to "run the source scan" — so the bugs below are *silent under-emission on the optimized path*, never crashes. The headline P1 is the one place that property is breached.
- **Security surface refuted across the board** (security-boundary lane, code-tested):
  - **ReDoS** on all three `hasConsumerSignals` regexes (java/php/python): refuted — all linear, no nested quantifiers, sub-ms on 100k-char adversarial inputs.
  - **Path traversal** via `readSafe(repoPath, f)`: refuted — `path.resolve` + `path.relative` + `..`/absolute rejection, null fail-open; `f` originates from graph rows, not user input.
  - **Cypher injection**: refuted — `handlerSymbolId` is never interpolated; `CONTAINS_QUERY` uses bound `$fileId`; `HANDLES_ROUTE_QUERY` is a constant string.
  - **Unicode/bidi**: clean (`git diff --check` clean, no bidi controls, only em-dash/arrow in prose).
- **`@PatchMapping` is already fixed** (parse-worker.ts:1096, landed in #2078) — a stale prior-context lead, correctly retired by the facts lane.
- **Kotlin correctly left `'partial'`** (no `routeCoverage` key → never skipped → safe). Java `\bwebClient\b` lowercase is intentional (scan only detects the lowercase receiver). PHP and Java consumer regexes are co-extensive with their scans (adversarial lane refuted both). PHP ingestion (Laravel) is a *superset* of the group PHP scan → no PHP provider over-claim.
- **Schema/CSV/COPY column parity is correct** (8 columns aligned across schema.ts / csv-generator.ts / lbug-adapter.ts; `handlerSymbolId` read only via the explicit projection, so the M0/M1 `getNodeQuery` exhaustiveness hazard does not apply). `bench/baselines.json` fingerprint change is the expected consequence of the new column. CHANGELOG correctly untouched. No `any`/`as any` introduced. RFC#909 language-naming rule respected (`resolveRouteHandlerSymbols` keys on `controllerName`/`handlerName`, not language names).
- **Performance:** net positive, no regressions (all perf findings NEGLIGIBLE/N-A; the per-row CONTAINS_QUERY is pre-existing, not worsened).

---

## P1 (MERGE-BLOCKING) — Java `routeCoverage: 'complete'` over-claims; parse-skip silently drops provider routes

**File:** `gitnexus/src/core/group/extractors/http-patterns/java.ts:625` (`routeCoverage: 'complete',` — added in this diff)
**Cross-engine:** **STRONG.** Codex swarm (B1) Finding 1 + Codex ce (B2) findings #1 & #2 + Claude ce-adversarial F1/F2 (probe-confirmed) + Claude ce-correctness (verb-loss). Coordinator independently verified the array-form code asymmetry by reading `spring.ts` and `java.ts`. **[code-read]** (worktrees can't build; adversarial lane ran real tree-sitter probes on both producers).

**Root cause.** `routeCoverage: 'complete'` asserts that ingestion emits a graph Route node for **every** provider the group `scan()` detects. For Java this is **false**: the graph provider set is a strict *subset* of the group-scan provider set, and this PR removes the source scan that used to recover the difference. A file is parse-skipped iff (1) its plugin declares `'complete'`, (2) every `HANDLES_ROUTE` row for it resolved a `handlerSymbolId`, and (3) `hasConsumerSignals` returns false. When a file has *one* simple resolvable mapping plus a group-only route shape, condition (2) is satisfied by the simple route alone → file covered → skipped → the group-only routes vanish.

**Three sub-cases, same root cause:**

1. **Array-form `@GetMapping({"/a","/b"})`** *(confidence 100, probe-confirmed).* Ingestion's `ROUTE_ANNOTATION_QUERY` (`spring.ts:51,59,64,72`) matches only `(string_literal) @value` — it has **no** `element_value_array_initializer` branch. The group query (`java.ts:124`) matches `[(string_literal) @value (element_value_array_initializer (string_literal) @value)]`. Existing passing group test: `http-route-extractor.test.ts:1850`. → `/a`, `/b` get no graph Route node; a co-located simple route covers the file; file skipped; `/a`,`/b` are never emitted.
2. **Interface-inherited Spring routes** *(confidence 90).* The group `scanSpringInheritanceProject` (`spring-consumer-shared.ts:197-268`) composes interface `@RequestMapping`s onto implementing controllers; ingestion `spring.ts` only walks class-direct annotations. `collectProjectDetections` runs `scanProject` over `scanFiles` *minus* the covered controller, so the inheritance pass never receives the controller and can't emit the inherited provider. Existing passing group test: `http-route-extractor.test.ts:863`.
3. **Same-URL multi-verb (GET + POST `/orders`)** *(confidence 100, Codex ce #2 + Codex adversarial B3 finding 1).* The Route node id is `generateId('Route', routeURL)` — keyed by URL only; method is a property. GET+POST collapse to one node / one `HANDLES_ROUTE` row. The file reads "all resolved" on that single row → covered → skipped. Pre-PR the source scan emitted both verbs and the merge recovered the dropped one; with the scan skipped, the non-surviving verb's provider contract is lost. (Not Java-specific — it's the URL-keyed Route identity, but Java is the live trigger.) **Both Codex legs (ce + adversarial) independently surfaced this one.**

**Concrete trigger (sub-case 1, cleanest).** A pure-provider `@RestController` with `@GetMapping("/covered")` + `@GetMapping({"/a","/b"})` and no consumer (RestTemplate/WebClient) calls: `/covered` resolves a handler → file all-resolved + Java `'complete'` → covered; `hasConsumerSignals` false → file removed from `scanFiles` → `/a`,`/b` never emitted.

**Fix (both engines converge).** Keep Java at `routeCoverage: 'partial'` until ingestion provider extraction matches the group scan (array-form branch in `ROUTE_ANNOTATION_QUERY` + interface inheritance + per-verb Route identity), **OR** run `scanProject` over the FULL file set and only suppress per-file provider *emission* for covered files (so cross-file inheritance metadata is retained). Add regression tests for the array-form, interface-inherited, and multi-verb skip cases.

---

## P2 — `handlerSymbolId` mis-attribution to the wrong handler (fail-open keeps it from being data loss)

**File:** `gitnexus/src/core/ingestion/call-processor.ts:303` (`lookupExactAll(...)[0]`) and `:294`; stamp at `gitnexus/src/core/ingestion/pipeline-phases/routes.ts:288`.
**Cross-engine:** **STRONG.** Codex swarm (B1) Finding 2 + Codex ce (B2) #6 + Claude ce-correctness F1/F2 + Claude ce-adversarial F4. **[code-read].** Coordinator-confirmed.

Two distinct mechanisms produce a wrong-but-plausible handler UID:

- **First-writer asymmetry.** The routes phase `addRoute` (`routes.ts:241-247`) is first-writer-wins per URL: the first route registered for a URL wins the Route node, later same-URL routes are counted as duplicates. `resolveRouteHandlerSymbols.put` (`call-processor.ts:276-307`) is also first-writer per URL but *skips* unresolvable URLs and iterates in an independent order (Laravel routes before decorator routes). On a same-URL collision where the node-winning route is unresolvable (e.g. ambiguous controller short-name) but a later same-URL route resolves, `routeHandlerSymbols[url]` holds the *later* route's handler, which is then stamped onto the winning route's node. Also reachable cross-source (a filesystem/Next route, registered first, winning a URL a framework route also normalizes to — the resolver never sees filesystem routes).
- **`lookupExactAll(file, name)[0]`** takes the first def of a same-named symbol → an overloaded/same-named handler resolves to an arbitrary (possibly wrong) UID. `spring.ts` captures the exact decorated node but resolution discards that identity and re-looks-up by name only.

**Effect.** A wrong-but-plausible `handlerSymbolId` flows into the emitted contract and impact analysis points at the wrong handler. Fail-open (URL-key mismatch → no stamp) prevents outright route loss, so this is P2, not P1.

**Fix.** Have `put` reserve the URL slot even when unresolved (mirror `addRoute`'s first-writer-wins exactly), and key on the winning routes-phase entry identity; when `lookupExactAll` returns >1, disambiguate by the captured handler node's line/position or skip resolution (fail-open) rather than guessing `[0]`.

## P2-LATENT (dormant) — Python `hasConsumerSignals` is not a superset of Python `scan()`

**File:** `gitnexus/src/core/group/extractors/http-patterns/python.ts:929`.
**Cross-engine:** Codex swarm F3 (rated P3) + Codex ce #3 (rated P1) + Claude ce-adversarial F3 (P2, flags dormancy) + facts lane. **[code-read].** **Codex rated this P1 but both Codex legs missed that it is currently unreachable** — coordinator calibrates to **P2-latent**.

`hasConsumerSignals` (`/\brequests\s*\.|\bhttpx\b/`) only matches `requests`/`httpx`, but `scan()` also emits `python-http-wrapper` consumers for `uri=`/`url=` keyword/variable calls (`python.ts:1095-1150`) on *any* object. Per the `types.ts:111-116` contract, a `'complete'` plugin's `hasConsumerSignals` must cover every consumer shape `scan()` emits — it does not.

**Why dormant (and why not a blocker):** for a Python file to be skipped it must first resolve a `handlerSymbolId`, but FastAPI decorator routes set **no** `handlerName` (generic worker path, `parse-worker.ts:1450-1463`) and Django sets `methodName: null` (`django.ts:181`). So **no Python file is parse-skipped today** — the Python optimization is effectively a no-op now. The gap **arms** the moment a follow-up gives FastAPI routes a `handlerName` (a natural Part-3 change). Fix now (broaden the regex, or derive it from the scan's identifier set) or document the precondition. (Same class, pre-existing & not this PR: the Python consumer regex also misses aiohttp/urllib.)

---

## Lower-priority / release notes

- **Existing indexes need a re-analyze (NOT merge-blocking, NOT #2265-unique).** The new `Route.handlerSymbolId` column + 8-col COPY are applied via `CREATE NODE TABLE`, which swallows "already exists", so an existing `.lbug` DB keeps its old 7-col Route table. The read path is fail-open: `HANDLES_ROUTE_QUERY` against the old schema returns NULL → `?? ''` → all files "not resolved" → full scan; only graph-only providers (Django) transiently degrade until re-analyze. **However**, a re-analyze that re-COPYs into the *old* table will hit a column-not-found COPY error (api-contract F1) — users must `clean` + analyze (or `--force`) to pick up `handlerSymbolId`.
  - **Both Codex legs (ce #5 + adversarial B3 finding 2) + Claude ce-api-contract rate the missing `INCREMENTAL_SCHEMA_VERSION` bump (`repo-manager.ts:248`, still 4) as high/P1; this digest demotes it — but it deserves the maintainer's explicit attention.** The premise is verified: an upgraded repo stays incremental-eligible against the old Route table, and an incremental writeback that re-COPYs Route could hit the missing column. The reason it is *not* treated as a #2265-introduced blocker: #2234 — which added `Route.method` one Part earlier via the *identical* schema-only pattern — also did **not** bump it and shipped two days ago, so this is a pre-existing pattern spanning both Parts, and the lbug DB is a derived cache (the project's model is "re-analyze on schema change"). Because I can't reproduce the incremental-COPY failure in a non-building worktree, I won't post it as a maintainer-signed assertion. **Recommended action:** confirm the incremental Route writeback is actually safe across a column addition; if not, bump `INCREMENTAL_SCHEMA_VERSION` to cover *both* new Route columns (`method` + `handlerSymbolId`) and add an old-schema upgrade regression test.
- **Maintainability (minor, P2/P3):** `extractProvidersGraph` has dual responsibility (returns contracts *and* mutates a `coveredFiles` out-param Set) with a duplicated `CONTAINS_QUERY` across fast/legacy paths; the `routeCoverage:'complete'` ⟹ `hasConsumerSignals` invariant is unenforced in the types (a silent data-loss trap for future plugin authors — could be a discriminated union); dead alias `const pathNorm = pathNormEarly` (~http-route-extractor.ts:480); positional `row[N]`/`hit[N]` fallbacks on `Record<string,unknown>`.
- **Test gaps (none would catch the P1):**
  - `resolveRouteHandlerSymbols` (new exported fn, ~5 branches) has **zero** direct unit tests.
  - No `handlerSymbolId` CSV→COPY→query roundtrip test (`Route.method` has `route-method-roundtrip.test.ts`; this column has no analogue).
  - Consumer-safety gate fail-open branches (null content, missing plugin, absent `hasConsumerSignals`) untested.
  - **No regression test for any of the three P1 data-loss cases** (array-form / interface-inherited / multi-verb parse-skip).
  - Positive note: `http-route-graph-method.test.ts` is a real test, not a worker-dependent silent no-op (confirmed by ce-testing + project-standards).

## Validated / refuted (verification credit)

| Claim | Outcome | Lane |
|---|---|---|
| ReDoS on the 3 `hasConsumerSignals` regexes | **Refuted** (linear, sub-ms on 100k inputs) | security-boundary |
| Path traversal via `readSafe` | **Refuted** (resolve/relative/`..` guard, null fail-open) | security-boundary |
| Cypher injection via `handlerSymbolId` | **Refuted** (never interpolated; `$fileId` bound) | security-boundary |
| Unicode / bidi controls | **Clean** | security-boundary |
| `@PatchMapping` dropped | **Already fixed** (#2078, parse-worker.ts:1096) | facts |
| PHP / Java consumer regexes miss a scan shape | **Refuted** (co-extensive) | ce-adversarial |
| PHP provider over-claim | **Refuted** (Laravel ingestion ⊇ group scan) | ce-adversarial |
| Kotlin skip path | **Refuted** (defaults `'partial'`, never skipped) | ce-adversarial |
| Old/mixed DB schema corruption | **Fail-open** (try/catch → `[]` → full scan) | ce-adversarial, api-contract |
| `bench/baselines.json` fingerprint change | **Expected** (new CSV column) | performance |
| CHANGELOG touched / `any` introduced / RFC#909 broken | **None** | project-standards |

---

*This is an automated two-engine digest (Claude: GitNexus swarm + ce-code-review; Codex gpt-5.5/xhigh: swarm + ce-code-review + adversarial-review). Verify the P1 fix direction and the inline anchors against the diff before acting; the recommendation is to hold merge until Java is either downgraded to `'partial'` or ingestion is made provider-complete.*

magyargergo:
@henry201605 what do you think about this https://github.com/abhigyanpatwari/GitNexus/pull/2269?

henry201605:
Thanks for the detailed review — the P1 is correct. I confirmed all three sub-cases against the code:

- **array-form**: `spring.ts` route queries match `(string_literal)` only; the group query (`java.ts:124`) also matches `element_value_array_initializer`.
- **interface-inherited**: `scanSpringInheritanceProject` composes them; ingestion walks class-direct annotations only.
- **same-URL multi-verb**: the Route node is URL-keyed, so `addRoute` drops the second verb as a duplicate.

So `routeCoverage: 'complete'` for Java over-claims — agreed it's merge-blocking. Plan for this PR:

- **Java → `'partial'`** (your first option). Removes the data-loss path; Java is always source-scanned until ingestion matches the group scan.
- **Python → `'partial'`** as well. It's a no-op today anyway (FastAPI/Django set no `handlerName`, so no Python file is ever skipped), and leaving it `'complete'` is a latent trap the moment a `handlerName` lands. I'll also widen the Python `hasConsumerSignals` regex to a true superset of `scan()` (the `uri=`/`url=` wrapper, aiohttp, urllib) so the precondition holds when it does get enabled.
- **PHP stays `'complete'`** — your adversarial lane confirmed Laravel ingestion ⊇ the group scan, and it's the one language the skip actually engages for. If you'd rather be conservative about the URL-keyed multi-verb identity for PHP too, I'll downgrade it — your call.
- **P2**: disambiguate `handlerSymbolId` by the captured node's position instead of `[0]`, and make `put` reserve the URL slot like `addRoute` so a later same-URL route can't stamp the winning node; fail open when it can't disambiguate.
- Regression tests for the array-form / interface-inherited / multi-verb skip cases, plus direct tests for `resolveRouteHandlerSymbols`.

Net: the PR keeps the foundation (resolved handler symbols persisted on Route nodes + the parse-skip mechanism) and the PHP win, with no data loss. I'd like to do the ingestion-completeness work as a **separate follow-up** rather than grow this PR — array-form branch in `ROUTE_ANNOTATION_QUERY`, interface-inherited provider emission, and per-verb Route identity.

On #2269 — the cross-repo `trace` stitching is where the handler-symbol resolution in this PR pays off concretely. Today HTTP provider contracts carry `symbolUid: ''`, so the bridge anchors by file; the graph fast-path in this PR stamps the ingestion-resolved handler UID onto a covered provider's `symbolUid`, so the bridge can anchor on the symbol instead of falling back to the file. That benefit is independent of the parse-skip — it lives in the graph provider path, so downgrading Java to `'partial'` keeps it.

The follow-up (ingestion provider-completeness) reads naturally as "give the currently-empty cases a real `symbolUid`": array-form and interface-inherited routes get a Route node at all, and per-verb Route identity gives the second verb of a same-URL pair its own resolved provider instead of an empty-uid source-scan one. None of that needs a change on the trace side — `contractId` already carries the method — it just widens how often your bridge anchors on a symbol rather than a file. If you'd rather I sequence that follow-up to land before leaning on it in trace, happy to.

magyargergo:
Please continue working on this one 🙏 I only shared my work with you because I thought you might be interested.

henry201605:
Thanks again for the thorough review — really helpful in pinning these down. Pushed the fixes, rebased on top of your #2268.

**P1 — Java `routeCoverage` → `'partial'`.** Java is now always source-scanned, so the three group-only shapes (array-form `@GetMapping({...})`, interface-inherited routes, same-URL multi-verb) can't be dropped. Python → `'partial'` too (it was a no-op anyway — no FastAPI/Django route resolves a `handlerName`, so nothing was ever skipped — and `'complete'` was only a latent trap). PHP stays `'complete'` (the one language the skip engages for, and your adversarial lane already confirmed Laravel ingestion ⊇ the group scan).

Added the regression guards you asked for (array-form / interface-inherited / multi-verb): each asserts the group-only route survives. I verified they actually bite — flipping Java back to `'complete'` turns all three red.

**P2 — `resolveRouteHandlerSymbols`.** Two changes: (1) the `claim` helper (was `put`) now reserves a URL slot on first encounter even when unresolved, so within the framework + decorator routes the resolver sees, a later same-URL route can't override the first one's slot. (The cross-source case you noted — a filesystem/Next route winning a URL a framework route also normalizes to — is unchanged: the resolver never receives filesystem routes, so that path stays fail-open.) (2) on an ambiguous same-name lookup I went with the **fail-open** option you offered rather than line/position disambiguation — `SymbolDefinition` doesn't carry a node position at this layer, and threading one through the symbol table felt out of scope for this PR. Ambiguous or zero matches → no `handlerSymbolId` → source-scan fallback (never a wrong handler). Added direct unit tests (the function had none): unique-resolve, ambiguous/unknown fail-open, the same-URL reservation case, and first-writer-wins.

Also widened the Python `hasConsumerSignals` to a true superset of its `scan()` (uri=/url= wrapper, aiohttp, urllib); Java's gate already covers the consumer set #2268 expanded (same receivers, and the `@*Exchange` token is in there). Added a unit test per language, and a `Route.handlerSymbolId` CSV→COPY→query roundtrip.

Ingestion provider-completeness (what later lets Java flip back to `'complete'`) stays the separate follow-up I mentioned. Happy to adjust any of this if you'd prefer a different cut — thanks!

Issue #2204, opened 2026-06-15
feat(taint): register source/sink models for more languages


magyargergo:
## Summary

The taint layer runs for every `--pdg` language but only **JS/TS** have registered source/sink models. Add per-language taint source/sink specs so cross-language security flows light up.

## Background

Cross-language PDG validation (PR #2197): the taint solver ran on all 12 languages, but `TAINTED` findings were emitted only where a model exists —

- `express` (JS): 34 `TAINTED` findings
- every other language (python, java, go, c#, php, ruby, rust, swift, kotlin, dart, c/c++): **0** findings, `taint=0ms` — no model registered.

The engine (per-function summaries, interprocedural fixpoint, `TAINTED`/`TAINT_PATH` emit, `explain`) is implemented and language-agnostic — it just needs per-language source/sink/sanitizer specs (see `getSourceSinkConfig` / `registerBuiltinTaintModels`).

## Proposed work

- Register taint models for additional ecosystems: HTTP request params (sources), SQL / command-exec / file-path / template sinks, and per-ecosystem sanitizers. Start with python + java (highest-value security surfaces).

## Acceptance

- python + java models registered with unit fixtures showing real source→sink flows.
- Non-zero `TAINTED` findings on representative real repos (e.g. a Flask app, a Spring app) in the validation harness.

Surfaced by #2195 cross-language validation. Refs #2195.

azizur100389:
I’d like to take a conservative first slice here: Python taint models first, focused on common request-parameter sources and high-confidence SQL/command/path sinks, with fixtures proving real source-to-sink flows and avoiding broad sanitizer guesses. If that lands cleanly, Java can follow as a separate PR.

magyargergo:
> I’d like to take a conservative first slice here: Python taint models first, focused on common request-parameter sources and high-confidence SQL/command/path sinks, with fixtures proving real source-to-sink flows and avoiding broad sanitizer guesses. If that lands cleanly, Java can follow as a separate PR.

Can you please create a subticket for this? 🙏

azizur100389:
Created the focused subticket here: #2252. I’ll target the Python-only slice there and leave Java / broader language coverage under the parent follow-up scope.

azizur100389:
Created the focused Java taint subticket here: #2261. I’ll target the conservative Java call-result source and high-confidence sink slice there, following the merged Python work from #2253.

PR #2267, opened 2026-06-21
feat(taint): add conservative Java source/sink model

### LINKED ISSUE #2261: feat(taint): Java source/sink model slice
## Summary

Add a conservative first Java taint-model slice for #2204, following the merged Python work in #2253.

This slice should focus on Java call-result taint sources and high-confidence sinks that the current CFG/taint substrate can prove without guessing.

## Scope

- Register a Java source/sink model through the existing built-in taint model registry.
- Support servlet-style request call-result sources when the receiver and method shape are conservative enough to avoid broad false positives, for example conventional `request`/`req` receivers calling methods such as `getParameter` or `getHeader`.
- Taint only the recorded `resultDefs` of matching source calls.
- Add high-confidence Java sinks only when the current harvested site/provenance data can distinguish them from unrelated same-named methods.
- Preserve existing JS/TS and Python model behavior.

## Out of scope for this slice

- Spring controller-parameter annotation inference unless the existing substrate can support it soundly.
- Broad matching of every method named `execute`, `query`, `exec`, `read`, or `write`.
- Sanitizers. Register none unless semantics and provenance are unambiguous.
- Unsupported, ambiguous, shadowed, call-rooted, or unproven forms should degrade to no match.

## Acceptance

- Positive fixture for Java request call-result source.
- Positive source-to-sink flow.
- Source result assigned then propagated.
- Direct nested source-to-sink expression if supported by harvested data.
- Negative fixture for unrelated same-named receiver/method.
- Negative fixture for unassigned call-result.
- Negative fixture for sink with no tainted dangerous argument.
- Registry idempotence and Java inclusion coverage.
- Existing Python and TS/JS model behavior remains unchanged.
- Model-version invalidation coverage.

Refs #2204.

azizur100389:
## Summary

- add a generic call-result taint-source variant that seeds only proven `resultDefs`
- register a conservative Java model for servlet `request`/`req` `getParameter` and `getHeader` results
- model only static-import-proven `java.nio.file.Files.readString` as the initial Java path sink
- preserve legacy member-read finding and `TAINTED` edge identities while adding deterministic call-result identities
- cover intra-procedural propagation, function-summary harvesting, emission, registry/version behavior, and negative provenance cases

## Conservative boundaries

- `Paths.get` and `Path.of` are path constructors, not sinks
- JDBC, broad `execute*`, `Runtime.exec`, and `ProcessBuilder` matching remain out of scope because current site data cannot prove their receiver types safely
- unassigned call results are not seeded
- unsupported or same-named calls without exact static-import provenance do not match
- no Java sanitizers are registered

## Validation

- `npx tsc --noEmit`
- affected suite: 16 files, 477 tests passed
- taint suite: 14 files, 237 tests passed
- Java resolver integration: 186 tests passed
- five subprocess-heavy CLI files rerun alone: 57 tests passed on both feature and clean current `upstream/main`
- `git diff --check`
- GitNexus `detect_changes`: LOW risk, 0 affected processes
- Solo production-readiness review: production-ready with minor follow-ups; no production-readiness issues found against the current DoD bar

Full `npm test` was run on Windows: 10,515 passed, 121 skipped, and 11 failed across eight unrelated environment/performance-sensitive files. The three persistent failures reproduce on a clean archive of current `upstream/main` (`git-utils` trailing-space path, literal collector, Swift scaling). The remaining five files pass standalone on both clean main and this branch and fail only under full-suite process/resource contention. No taint or Java resolver test failed.

Closes #2261
Refs #2204

### DIFF
--- a/gitnexus/src/core/ingestion/taint/java-model.ts
+++ b/gitnexus/src/core/ingestion/taint/java-model.ts
@@ -1,0 +1,24 @@
+/**
+ * Built-in Java taint model (#2261 first slice).
+ *
+ * This deliberately starts small. Servlet request input is modeled only when a
+ * conventional request receiver's call result is assigned to a binding. Sinks
+ * are limited to static-import-proven JDK filesystem operations that current
+ * harvested call-site/import data can identify without broad same-name matching.
+ * No sanitizers are registered in this slice.
+ */
+
+import type { SourceSinkSanitizerSpec } from './source-sink-config.js';
+
+export const JAVA_TAINT_MODEL: SourceSinkSanitizerSpec = {
+  sources: [
+    {
+      type: 'call-result',
+      kind: 'remote-input',
+      receivers: ['request', 'req'],
+      methods: ['getParameter', 'getHeader'],
+    },
+  ],
+  sinks: [{ name: 'readString', kind: 'path-traversal', args: [0], module: 'java.nio.file.Files' }],
+  sanitizers: [],
+};
--- a/gitnexus-shared/src/scope-resolution/types.ts
+++ b/gitnexus-shared/src/scope-resolution/types.ts
@@ -105,6 +105,13 @@ export type ParsedImport =
       readonly localName: string;
       readonly importedName: string;
       readonly targetRaw: string;
+      /**
+       * Set by providers when `targetRaw` already names the imported symbol
+       * rather than only its containing module. Consumers that compose
+       * `<local>.<member>` paths can then use `targetRaw.<member>` instead of
+       * duplicating `importedName`.
+       */
+      readonly targetIncludesImportedName?: boolean;
     }
   /**
    * Per-name import with rename.
@@ -119,6 +126,8 @@ export type ParsedImport =
       readonly importedName: string;
       readonly alias: string;
       readonly targetRaw: string;
+      /** See the same field on the `named` variant. */
+      readonly targetIncludesImportedName?: boolean;
     }
   /**
    * Qualified module handle, with or without rename. `importedName` is the
--- a/gitnexus/src/core/ingestion/languages/java/interpret.ts
+++ b/gitnexus/src/core/ingestion/languages/java/interpret.ts
@@ -30,6 +30,7 @@ export function interpretJavaImport(captures: CaptureMatch): ParsedImport | null
         localName: nameCap?.text ?? simpleName,
         importedName: simpleName,
         targetRaw: sourceCap.text,
+        targetIncludesImportedName: true,
       };
     }
     case 'wildcard': {
--- a/gitnexus/src/core/ingestion/taint/emit.ts
+++ b/gitnexus/src/core/ingestion/taint/emit.ts
@@ -232,7 +232,6 @@ export function emitFileTaint(
       const b = bindings[idx];
       return b === undefined ? `#${idx}` : bindingKey(b);
     };
-
     // SANITIZES — one edge per kill, REGARDLESS of findings (kills can and do
     // exist with zero findings: a fully-sanitized flow IS the kill evidence).
     for (const kill of flows.kills) {
@@ -263,13 +262,27 @@ export function emitFileTaint(
       // `exec(req.body, req.query)`'s two findings; `property` is free-text
       // (string-literal subscripts) and rides LAST so it cannot collide into
       // another component.
-      const id = generateId(
-        'TAINTED',
-        `${fnAnchor}:${finding.sinkKind}:` +
-          `${pointKey(source.point)}.${source.siteIndex}:${bKey(source.objectBindingIdx)}:` +
-          `${pointKey(sink.point)}.${sink.siteIndex}.${sink.argIndex}:${bKey(sink.bindingIdx)}:` +
-          `${sink.entryName}:${source.property}`,
-      );
+      const id =
+        source.type === 'member-read'
+          ? generateId(
+              'TAINTED',
+              `${fnAnchor}:${finding.sinkKind}:` +
+                `${pointKey(source.point)}.${source.siteIndex}:${bKey(source.objectBindingIdx)}:` +
+                `${pointKey(sink.point)}.${sink.siteIndex}.${sink.argIndex}:${bKey(
+                  sink.bindingIdx,
+                )}:` +
+                `${sink.entryName}:${source.property}`,
+            )
+          : generateId(
+              'TAINTED',
+              `${fnAnchor}:${finding.sinkKind}:` +
+                `${pointKey(source.point)}.${source.siteIndex}:call-result:` +
+                `${bKey(source.resultBindingIdx)}:${source.calleeName}:` +
+                `${pointKey(sink.point)}.${sink.siteIndex}.${sink.argIndex}:${bKey(
+                  sink.bindingIdx,
+                )}:` +
+                `${sink.entryName}`,
+            );
       if (seenEdgeIds.has(id)) continue;
       seenEdgeIds.add(id);
       // `kind` rides the reason's `;<kind>` header — the only persisted
--- a/gitnexus/src/core/ingestion/taint/match.ts
+++ b/gitnexus/src/core/ingestion/taint/match.ts
@@ -76,8 +76,10 @@
 import type { ParsedImport } from 'gitnexus-shared';
 import type { FunctionCfg, SiteRecord } from '../cfg/types.js';
 import type {
+  TaintCallResultSourceEntry,
   SourceSinkSanitizerSpec,
   TaintMemberSourceEntry,
+  TaintSourceEntry,
   TaintSanitizerEntry,
   TaintSinkEntry,
 } from './source-sink-config.js';
@@ -99,10 +101,23 @@ export type TaintImportIndex = ReadonlyMap<string, TaintImportBinding>;
 
 /** A member-read site matched as a taint source. */
 export interface MatchedSourceRead {
+  readonly type: 'member-read';
   /** Index into the owning statement's `sites` array. */
   readonly siteIndex: number;
   readonly entry: TaintMemberSourceEntry;
 }
+
+/** A call-result source matched on a call site with direct result definitions. */
+export interface MatchedSourceCall {
+  readonly type: 'call-result';
+  /** Index into the owning statement's `sites` array. */
+  readonly siteIndex: number;
+  readonly entry: TaintCallResultSourceEntry;
+  /** Bindings directly defined by this call result. Never empty. */
+  readonly resultDefs: readonly number[];
+}
+
+export type MatchedSource = MatchedSourceRead | MatchedSourceCall;
 
 /** A call/new site matched as a sink. */
 export interface MatchedSinkCall {
@@ -141,7 +156,7 @@ export interface StatementMatches {
   readonly blockIndex: number;
   readonly statementIndex: number;
   readonly line: number;
-  readonly sources: readonly MatchedSourceRead[];
+  readonly sources: readonly MatchedSource[];
   readonly sinks: readonly MatchedSinkCall[];
   readonly sanitizers: readonly MatchedSanitizerCall[];
 }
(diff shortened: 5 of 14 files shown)


koriyoshi2041:
Pulled this branch locally and did a focused pass on the Java taint slice. The conservative boundaries look consistent with the PR description: call-result sources only seed recorded `resultDefs`, `readString` only matches the static-import-proven `java.nio.file.Files` sink, and `Path.of` / `Paths.get` stay out of sink matching.

Local checks from `gitnexus/`:

```
npm test -- --run test/unit/taint/java-model-match.test.ts test/unit/taint/model-match.test.ts test/unit/taint/summary-harvest.test.ts test/unit/taint/taint-emit.test.ts
# 4 files, 81 tests passed

npm test -- --run test/unit/taint
# 14 files, 237 tests passed

npx tsc --noEmit
# passed
```

I do not see a code-level blocker in this slice. The remaining unstable signal looks like the external Vercel authorization status rather than the taint/test surface.

gitnexus/src/core/ingestion/taint/java-model.ts
```diff
@@ -0,0 +1,24 @@
...
+export const JAVA_TAINT_MODEL: SourceSinkSanitizerSpec = {
+  sources: [
+    {
+      type: 'call-result',
+      kind: 'remote-input',
+      receivers: ['request', 'req'],
+      methods: ['getParameter', 'getHeader'],
+    },
+  ],
+  sinks: [{ name: 'readString', kind: 'path-traversal', args: [0], module: 'java.nio.file.Files' }],
```
magyargergo, line 22: **[P2 coverage limitation — reproduced, cross-engine]** This `Files.readString` / `java.nio.file.Files` sink fires only for `import static java.nio.file.Files.readString;`. The idiomatic `import java.nio.file.Files; Files.readString(...)` canonicalizes (match.ts:267 builds `[imp.module, imp.member, ...rest]`) to `java.nio.file.Files.Files.readString` (doubled `Files` segment), which never equals the exact string the sink check requires at match.ts:315 — so the **dominant real-world Java import form is silently missed** (probe: `{sinks:0, findings:0}`). Only the rarer static-import form fires.

Not hard-blocking (opt-in `--pdg`, scoped a *conservative first slice*, fails-safe / under-reports), but it is the **#1 issue**: a user scanning a genuinely vulnerable servlet written with the normal import would see zero findings and could read that as a clean bill of health. Recommend either teaching Java class-imported-static-receiver canonicalization to drop the doubled type segment, or filing a tracking issue + adding a documenting negative test so the limitation is visible.

_Cross-engine: Claude ce-adversarial F1 (conf 95) + Codex-adversarial `[high]` + Codex-ce residual risk #1._ [reproduced]

magyargergo (commented):
**Two-engine tri-review.** Engine A (Claude): swarm [partial — risk-architect + test-ci-verifier ended mid-investigation, domains covered by ce] + ce (correctness, adversarial, testing, maintainability, standards, api-contract). Engine B (Codex gpt-5.5/xhigh): ce [live — "Ready to merge", zero findings, 3 residual risks] + adversarial [live — "needs-attention", one [high] finding] + swarm [turn FAILED mid-stream after `detect_changes`, partial → not a finding source]. Cross-engine corroboration IS available (B-adversarial corroborates the headline finding).

## What's solid (credited across both engines)

- **No correctness defect found by either engine.** `tsc --noEmit` clean; the full taint suite (16 files / 243–244 tests) runs green, executed independently by ce-correctness, ce-adversarial, and Codex-ce (the latter ran the 4 touched files: 81 tests pass). Codex-ce verdict: "Ready to merge."
- **Existing JS/TS/Python member-read taint is preserved** [reproduced]. The new `type:'member-read' | 'call-result'` union is a no-op for member-read-only models: `isCallResultSource` returns false for entries without `type`, the call-result branches are never taken, and the member-source-gating in summary-harvest is logically identical to the prior unconditional path. Confirmed by ce-adversarial F6, ce-correctness FOCUS#4, the A-swarm risk fragment, and Codex.
- **Source-type identity keys don't collide** [reproduced]. Member-read embeds a `bindingKey` (`name@module` / `name:line:col`, always `@`/`:`-suffixed) where call-result embeds the literal `call-result` token + `calleeName`; `findingKey` puts a numeric `objectBindingIdx` where call-result puts the literal `'call-result'`. Byte-collision is impossible. ce-adversarial F5 + ce-correctness FOCUS#1 both refuted the briefing's collision concern; verified against the two new emit-id assertions in taint-emit.test.ts.
- **RFC#909 clean** [code-read, confidence 100]. grep over all five shared `core/ingestion/taint/*` files (match, propagate, emit, source-sink-config, summary-harvest) finds zero language names / Java literals / method names — all confined to `java-model.ts`, registered via the language-neutral registry. Exactly the union-variant extension the design anticipated.
- **New test parses REAL Java end-to-end** through the full CFG pipeline (not synthetic fact injection), with a solid negative-test complement (wrong receiver, no static import, unrelated import, untainted input, unassigned result).
- **Opt-in, fail-safe design.** Runs only under `--pdg` → zero default-run blast radius. The taint model fails SAFE (under-reports). CHANGELOG untouched; conventional-commit title; correct `<language>-model.ts` file placement.

## Headline finding (P2, cross-engine corroborated) — Java path-traversal sink misses the dominant regular-import idiom

**Both engines independently traced this.** For the idiomatic Java form:
```java
import java.nio.file.Files;
import java.nio.file.Paths;
String p = request.getParameter("path");
Files.readString(Paths.get(p));   // classic path-traversal
```
the import canonicalization at `match.ts:267` builds `[imp.module, imp.member, ...rest]` = `java.nio.file.Files` (module) + `Files` (member) + `readString`, yielding `java.nio.file.Files.Files.readString` (doubled `Files` segment). The sink check at `match.ts:315` requires the exact `java.nio.file.Files.readString`, so it never fires. The sink matches **only** the rarer static-import form `import static java.nio.file.Files.readString;`. Probe result for the regular-import form: `{ sinks: 0, findings: 0 }`. [reproduced by Claude ce-adversarial F1 (conf 95); reproduced/traced by Codex-adversarial [high] at match.ts:261-315; listed as residual risk #1 by Codex-ce]

**Calibrated position: strong P2 coverage limitation / recommended follow-up — NOT hard-merge-blocking.** Codex-adversarial rates this [high] "do not ship"; Claude ce-adversarial rates it P2 "intentional/documented conservatism." Reconciling: the feature is opt-in (`--pdg` only), the scope is explicitly "conservative first slice" and documented in the `java-model.ts:1-9` docstring as "static-import-proven", and it fails SAFE (under-reports, never a false clean). That removes the merge blocker. **But** the regular-import form is the overwhelmingly common real-world Java idiom, so the coverage gap is severe in practice: a user scanning a genuinely vulnerable servlet would see zero findings and could mistake that for a clean bill of health. This is the #1 finding. Recommendation (collaborative, not blocking): either teach Java callee canonicalization to drop the doubled type segment for a class-imported static-method receiver (`java.nio.file.Files.readString`), OR, at minimum, file an explicit follow-up issue + a documenting negative test, so the limitation isn't mistaken for working Java path-traversal detection. The UX should frame findings as best-effort/incomplete.

## Other findings (none P0/P1)

- **Optional-vs-required `type` discriminant inconsistency** (P2, ce-maintainability M3, ce-api-contract AC-3) [code-read]. `TaintMemberSourceEntry.type` is optional (`type?: 'member-read'`, source-sink-config.ts:103) while `TaintCallResultSourceEntry.type` is required and the occurrence types (`MatchedSource`, `TaintSourceOccurrence`) treat `type` as a required discriminant. The discriminant logic is sound today (it gates on presence of `'call-result'`, not absence of `'member-read'`), but the config-level union isn't exhaustively checkable. Suggest making `type` required on member-read entries.
- **Near-duplicate id/key construction** (P2, ce-maintainability M1/M2) [code-read]. The two `generateId('TAINTED', …)` branches in emit.ts:265-285 and the two `findingKey` branches in propagate.ts:531-555 share a fully copy-pasted outer frame and differ only in the discriminator segment; the schemas must be hand-kept in sync. Suggest a shared `sourceFragment(source)` helper.
- **Guarded widening cast** (P2/P3, ce-maintainability M4, ce-correctness FOCUS#3, prior-context) [code-read]. `site.resultDefs as readonly number[]` (match.ts:383) is guarded by a non-empty-length check so it is provably sound and is not `as any`; a narrowed `const resultDefs = site.resultDefs; if (resultDefs && …)` avoids the cast per the project's strict-typing preference. Style nit, not a defect.
- **Stale doc-comments** (P3, ce-maintainability M5/M6) [code-read confirmed]. typescript-model.ts:116 still says "Register the built-in models for TypeScript, JavaScript, and Python" while the body registers Java at :122. `BaseSourceOccurrence`'s JSDoc (propagate.ts ~159-163) still describes "the matched member-read occurrence" though it is now the common base of both variants.
- **Field-qualified servlet request missed** (P2 security FN, ce-adversarial F2) [reproduced]. `this.request.getParameter(...)` produces a 3-segment path and is rejected by the `resolved.path.length === 2` gate; only a direct local/param named `request`/`req` matches. Common servlet idiom; documented conservatism direction.
- **Name-based source FP** (P3, ce-adversarial F3) [reproduced] — **refuted as a blocker**. A local object named `req`/`request` with a `getParameter` method is matched with no servlet-type check. False-positive is the safe direction for a detector and is documented; surfaced for a future type-constraint follow-up only.
- **`calleeName` `?? methods.join('|')` fallback** (informational, ce-adversarial residual, ce-correctness FOCUS#2) [code-read] — dead defensive code: the propagate-time site array is the same one the matcher indexed, so `srcSite.callee` is always defined for a matched call-result source. Harmless; could be dropped for clarity.

## Refuted suspicions (validation is a feature)

The briefing's three headline regression concerns were all reproduced and **refuted** by both engines:
- **Interprocedural call-result seed drop** — REFUTED [reproduced, ce-adversarial F4 + ce-correctness FOCUS#4]. The `sourceReturn`-only-member-read gating doesn't lose interproc taint: call-result seeds (`seedId:-1`) propagate through the worklist and still populate `sourceToReturn`/`sourceToCallArg`/`callResults`. Pinned by the new summary-harvest.test.ts call-arg test.
- **Edge-id / finding-identity collision** — REFUTED [reproduced] (see "What's solid").
- **JS/TS/Python regression** — REFUTED [reproduced] (see "What's solid").
- **Taint surviving a `Paths.get`/reassignment hop** — confirmed CORRECT [reproduced, ce-adversarial F7]: the statement-level precision floor taints the def, and `get` is not a sanitizer, so the flow survives (1 finding).

## Test gaps (recommended, not blocking) — all [code-read]

- **`getHeader` test asserts only finding count**, not `source.type`/`sinkKind`/binding (ce-testing F3). Weaker than the parallel `getParameter` test.
- **`req` short-form receiver is untested** (ce-testing F4) — all nine tests use `request`; a narrowing of the model to `['request']` would pass silently.
- **Call-result `source→return` interproc path has no committed unit test** (ce-testing F5, ce-correctness, ce-adversarial); the new test covers `sourceToCallArg` but not `sourceToReturn`. Pin the behavior either way.
- **No negative/documenting test for the regular-import sink miss** (ce-adversarial testing-gaps) — would make the F1 limitation visible and guard the FP-suppression invariant.
- **If-branching test-style nits** (project no-if-in-vitests preference, ce-testing F1/F2; flagged out-of-scope-as-standards by ce-standards): `if (decoded.ok)` at taint-emit.test.ts:141-144 (vacuous-pass risk — inner asserts dead if `ok` is true with an unexpected shape; prefer `assert(decoded.ok)`); `if (sinkSite === undefined) throw` at java-model-match.test.ts:102/176 (narrowing throws, lower risk). These are user-MEMORY preferences, not project-standards violations.

## CI status

Green except Vercel (auth-required infra failure — ignorable, not a code signal). `tests / ubuntu / coverage` passed; the real-parse Java static-import sink flow is validated end-to-end. `tsc --noEmit` clean (run by multiple lanes).

---
*Automated tri-review digest (Claude ce+swarm[partial] + Codex gpt-5.5 ce+adversarial[swarm failed mid-stream]); cross-engine confidence-weighted. Verify findings against the diff before acting.*

azizur100389:
﻿Addressed in `7f766882`:

- Fixed regular Java import canonicalization for `Files.readString(...)` without adding a sink alias.
- Added coverage for regular/static imports, unrelated imports, receiver shadowing, `req`, cross-language behavior, and source-to-return propagation.
- Applied the requested type narrowing and documentation cleanups.

Validation completed: focused taint and Java tests, typecheck, lint, formatting, coverage/full CI tests, Windows/macOS platform tests, ABI checks, and benchmarks pass.

PR #1888, opened 2026-05-29
feat(java): expand static HTTP consumer extraction coverage

ShiningXu:
## Summary

Refs #1873 for the Java-only static cases covered here.

- support fully-qualified Spring and OpenFeign annotations
- preserve `@RequestMapping` prefixes on non-Feign Spring controller interfaces
- extract static `URI.create(...)` and literal `UriComponentsBuilder` paths in RestTemplate calls, including `fromUriString(...)`, `fromHttpUrl(...)`, and query-only intermediate calls
- extract WebClient `method(HttpMethod.X).uri(...)` long-form calls, aligned with Kotlin PR #1884 boundaries: no double emit, no variable-bound verb guessing, and only GET/POST/PUT/DELETE/PATCH long-form verbs
- infer OkHttp verbs from sibling builder calls such as `.get()`, `.head()`, `.post(...)`, `.delete()`, and `.method("PATCH", ...)`
- cover Java HttpClient helper methods including `HEAD()` plus `.method("PATCH", ...)` and default GET builders

## Deferred

- dynamic variable URLs, constants, string concatenation, `@Value`, helper/wrapper methods, and cross-statement data-flow remain deferred
- Kotlin changes are intentionally out of scope

## Validation

- `npx vitest run test/unit/group/http-route-extractor.test.ts`
- `npx vitest run test/unit/group`
- `npx tsc --noEmit`
- `npx prettier --check src/core/group/extractors/http-patterns/java.ts test/unit/group/http-route-extractor.test.ts`
- `npx eslint src/core/group/extractors/http-patterns/java.ts test/unit/group/http-route-extractor.test.ts` (warnings only from existing non-null assertions in the test file)
- `git diff --check`
- `./gitnexus/dist/cli/index.js detect-changes --scope unstaged --repo GitNexus`

PR #2268, opened 2026-06-21
feat(group): expand Java and Kotlin HTTP consumer extraction (re #1888)

magyargergo:
## Summary

This carries on the #1888 effort to pull HTTP client calls out of source code so they can be matched against the routes that serve them. It began as a port of #1888 onto current main, since the original branch was written against the old `java.ts` layout and could not be merged cleanly, and it grew from there as review feedback came in. Everything lives in the group HTTP plugin for Java and Kotlin plus its unit tests, and no existing contract changes shape.

## What it does now

Java RestTemplate calls that build their path instead of passing a plain string now resolve, as long as the path is still knowable from the source. That covers `URI.create("/x")` and `UriComponentsBuilder` chains, and it follows the same query stripping and slash rules Spring itself applies, so the path we record is the one the code actually requests.

OkHttp and Java HttpClient requests now report their real verb rather than always saying GET. The verb sits on a sibling call in the builder chain, so the code walks the chain to find it: a literal `.method("X")`, one of the verb helpers like `.post()` or `.delete()`, or the plain `.build()` that genuinely defaults to GET. If the verb is a variable we cannot read from the source, we skip that call instead of guessing.

The builder matching also stopped caring about call order. A header or timeout set before the request is built no longer makes us miss it, and a builder call that comes before `.url()` or `.uri()` is handled too. Before emitting anything we confirm the chain really is a `Request.Builder` or an `HttpRequest.newBuilder`, so an unrelated `.url()` elsewhere does not sneak in. The `HttpRequest.newBuilder(uri)` constructor form is picked up as well.

Fully qualified Spring route annotations, the `@org.springframework...GetMapping` form, now match too, which is something the Kotlin side already did.

Kotlin OkHttp finally infers the verb the same way Java does. It used to emit GET for everything, so the same code in a `.java` file and a `.kt` file produced different results. They agree now, and there is a small harness that runs matched Java and Kotlin sources through both plugins and checks the output is the same on both.

## What is left out on purpose

A couple of things from the original #1888 are not here by design. The WebClient long form already landed on main through #2254, and the old single file provider emit for plain interfaces would now double count, since main resolves the interface to controller relationship across files.

## Testing

The whole group test suite passes, types check clean, and there are targeted tests for every case above, including rows in the Java and Kotlin parity harness that assert both languages produce identical contracts.

The history is a run of small focused commits rather than one big drop, because it grew over a few review passes.

### DIFF
--- a/gitnexus/src/core/group/extractors/http-patterns/java.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/java.ts
@@ -199,7 +199,7 @@ const REST_TEMPLATE_PATTERNS = compilePatterns({
         (method_invocation
           object: (identifier) @obj (#eq? @obj "restTemplate")
           name: (identifier) @method
-          arguments: (argument_list . (string_literal) @path))
+          arguments: (argument_list . (_) @path))
       `,
     },
   ],
@@ -216,7 +216,7 @@ const REST_TEMPLATE_EXCHANGE_PATTERNS = compilePatterns({
           object: (identifier) @obj (#eq? @obj "restTemplate")
           name: (identifier) @method (#eq? @method "exchange")
           arguments: (argument_list
-            . (string_literal) @path
+            . (_) @path
             (field_access
               object: (identifier) @httpMethodCls (#eq? @httpMethodCls "HttpMethod")
               field: (identifier) @http_method)))
@@ -379,6 +379,49 @@ function hasAnnotation(node: Parser.SyntaxNode, names: string | readonly string[
     stack.push(...cur.namedChildren);
   }
   return false;
+}
+
+// ─── Statically-resolvable consumer path helpers ──────────────────────
+// RestTemplate calls increasingly pass a non-literal path argument that is
+// still statically derivable — `URI.create("/x")` or a `UriComponentsBuilder`
+// fluent chain. These helpers resolve those shapes to a literal path; a
+// genuinely dynamic argument (a variable, a non-`URI`/`UriComponentsBuilder`
+// call) resolves to null and the call site is skipped.
+
+function methodInvocationName(node: Parser.SyntaxNode): string | null {
+  return node.type === 'method_invocation' ? (node.childForFieldName('name')?.text ?? null) : null;
+}
+
+function methodInvocationObject(node: Parser.SyntaxNode): Parser.SyntaxNode | null {
+  return node.type === 'method_invocation' ? node.childForFieldName('object') : null;
+}
+
+function methodInvocationArguments(node: Parser.SyntaxNode): Parser.SyntaxNode[] {
+  const argsNode = node.type === 'method_invocation' ? node.childForFieldName('arguments') : null;
+  return argsNode?.namedChildren ?? [];
+}
+
+function firstLiteralArgument(node: Parser.SyntaxNode): string | null {
+  const first = methodInvocationArguments(node)[0];
+  return first?.type === 'string_literal' ? unquoteLiteral(first.text) : null;
+}
+
+/** Resolve a `URI.create("/path")` call to its literal path; null otherwise. */
+function extractUriCreatePath(node: Parser.SyntaxNode): string | null {
+  if (node.type !== 'method_invocation') return null;
+  if (methodInvocationObject(node)?.text !== 'URI' || methodInvocationName(node) !== 'create')
+    return null;
+  return firstLiteralArgument(node);
+}
+
+/**
+ * Resolve a statically-derivable path argument to a literal path: a bare
+ * string literal, or a `URI.create("/x")` call. (U2 extends this with
+ * `UriComponentsBuilder` fluent chains.) Genuinely dynamic arguments → null.
+ */
+function extractStaticPathExpression(node: Parser.SyntaxNode): string | null {
+  if (node.type === 'string_literal') return unquoteLiteral(node.text);
+  return extractUriCreatePath(node);
 }
 
 interface MethodRouteAnnotation {
@@ -727,7 +770,7 @@ export const JAVA_HTTP_PLUGIN: HttpLanguagePlugin = {
       if (!methodNode || !pathNode) continue;
       const httpMethod = REST_TEMPLATE_TO_HTTP[methodNode.text];
       if (!httpMethod) continue;
-      const path = unquoteLiteral(pathNode.text);
+      const path = extractStaticPathExpression(pathNode);
       if (path === null) continue;
       out.push({
         role: 'consumer',
@@ -743,7 +786,7 @@ export const JAVA_HTTP_PLUGIN: HttpLanguagePlugin = {
       const httpMethodNode = match.captures.http_method;
       const pathNode = match.captures.path;
       if (!httpMethodNode || !pathNode) continue;
-      const path = unquoteLiteral(pathNode.text);
+      const path = extractStaticPathExpression(pathNode);
       if (path === null) continue;
       out.push({
         role: 'consumer',
--- a/gitnexus/src/core/group/extractors/http-patterns/java.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/java.ts
@@ -845,9 +845,10 @@ export const JAVA_HTTP_PLUGIN: HttpLanguagePlugin = {
       const path = unquoteLiteral(pathNode.text);
       if (path === null) continue;
       const method = inferOkHttpMethod(callNode);
-      // An explicit `.method(verb, …)` with a non-literal verb is unresolvable —
-      // emit nothing rather than a wrong GET.
-      if (method === null) continue;
+      // An explicit `.method(verb, …)` with a non-literal or empty verb is
+      // unresolvable (null) — emit nothing rather than a wrong GET or an empty
+      // `http::::/path` contract.
+      if (!method) continue;
       out.push({
         role: 'consumer',
         framework: 'okhttp',
@@ -874,7 +875,7 @@ export const JAVA_HTTP_PLUGIN: HttpLanguagePlugin = {
       const path = unquoteLiteral(pathNode.text);
       if (path === null) continue;
       const method = inferHttpClientMethod(callNode);
-      if (method === null) continue;
+      if (!method) continue;
       out.push({
         role: 'consumer',
         framework: 'java-http-client',
--- a/gitnexus/src/core/group/extractors/http-patterns/java.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/java.ts
@@ -503,7 +503,13 @@ function appendPath(base: string, subPath: string): string {
  * and the `query*` family pass through (query attributes do not change the
  * path). Any non-literal segment or unknown call → null.
  */
-function extractUriComponentsBuilderPath(node: Parser.SyntaxNode): string | null {
+// A UriComponentsBuilder chain deeper than this is not realistic source; cap the
+// recursion so a pathological / machine-generated chain returns null instead of
+// overflowing the stack (mirrors the project's other AST-depth guards).
+const MAX_BUILDER_DEPTH = 100;
+
+function extractUriComponentsBuilderPath(node: Parser.SyntaxNode, depth = 0): string | null {
+  if (depth > MAX_BUILDER_DEPTH) return null;
   if (node.type !== 'method_invocation') return null;
   const name = methodInvocationName(node);
   const objectNode = methodInvocationObject(node);
@@ -521,12 +527,12 @@ function extractUriComponentsBuilderPath(node: Parser.SyntaxNode): string | null
   }
   if (!objectNode) return null;
   if (name === 'path') {
-    const base = extractUriComponentsBuilderPath(objectNode);
+    const base = extractUriComponentsBuilderPath(objectNode, depth + 1);
     const subPath = firstLiteralArgument(node);
     return base !== null && subPath !== null ? appendPath(base, subPath) : null;
   }
   if (name === 'pathSegment') {
-    const base = extractUriComponentsBuilderPath(objectNode);
+    const base = extractUriComponentsBuilderPath(objectNode, depth + 1);
     if (base === null) return null;
     const args = methodInvocationArguments(node);
     const segments = args
@@ -547,7 +553,7 @@ function extractUriComponentsBuilderPath(node: Parser.SyntaxNode): string | null
     name === 'replaceQueryParam' ||
     name === 'replaceQueryParams'
   )
-    return extractUriComponentsBuilderPath(objectNode);
+    return extractUriComponentsBuilderPath(objectNode, depth + 1);
   return null;
 }
 
--- a/gitnexus/src/core/group/extractors/http-patterns/java.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/java.ts
@@ -31,6 +31,9 @@ import {
   extractStaticPathExpression,
   inferOkHttpMethod,
   inferHttpClientMethod,
+  okHttpUrlRootsAtBuilder,
+  httpClientUriRootsAtNewBuilder,
+  httpClientChainHasUriCall,
 } from './java-static-path.js';
 import type {
   HttpDetection,
@@ -278,20 +281,14 @@ const WEB_CLIENT_LONG_FORM_PATTERNS = compilePatterns({
   ],
 } satisfies LanguagePatterns<Record<string, never>>);
 
-// ─── Consumer: OkHttp `new Request.Builder().url("path")` ─────────────
-// Note: `Request.Builder` is a `scoped_type_identifier` whose text includes
-// the dot, so `#eq?` against the literal string matches cleanly (no need
-// to escape a regex dot). The verb is recovered by `inferOkHttpMethod`
-// (java-static-path.ts) walking UP from the matched `.url(...)` call.
-//
-// PRE-`.url()` LIMITATION (pre-existing, tracked): the query anchors `.url()`'s
-// object on the `Request.Builder` object-creation DIRECTLY, so a builder call
-// BEFORE `.url()` — `new Request.Builder().addHeader(...).url("/x")` — is not
-// matched. This is the OkHttp dual of the Java-HttpClient pre-`.uri()` gap
-// (see JAVA_HTTP_CLIENT_PATTERNS): both concern calls *before* the path call,
-// distinct from the post-path intervening calls (`.header()`/`.timeout()`) the
-// verb-walk already handles. Relaxing either anchor risks over-matching
-// `.url()`/`.uri()` on unrelated objects, so the fix is deliberately deferred.
+// ─── Consumer: OkHttp `Request.Builder()…url("path")` ─────────────────
+// Match a bare `.url("literal")` call on ANY receiver, capturing it as `@call`.
+// `okHttpUrlRootsAtBuilder` (JS) then verifies the receiver chain bottoms out on
+// `new Request.Builder()` — re-imposing the framework anchor while allowing
+// builder calls BEFORE `.url()` (`new Request.Builder().addHeader(...).url("/x")`)
+// that the old object-direct query dropped, and rejecting a `.url(...)` on an
+// unrelated object. The verb is recovered by `inferOkHttpMethod` scanning the
+// whole chain (java-static-path.ts).
 const OK_HTTP_PATTERNS = compilePatterns({
   name: 'java-okhttp',
   language: Java,
@@ -300,8 +297,6 @@ const OK_HTTP_PATTERNS = compilePatterns({
       meta: {},
       query: `
         (method_invocation
-          object: (object_creation_expression
-            type: (scoped_type_identifier) @type (#eq? @type "Request.Builder"))
           name: (identifier) @method (#eq? @method "url")
           arguments: (argument_list . (string_literal) @path)) @call
       `,
@@ -309,18 +304,15 @@ const OK_HTTP_PATTERNS = compilePatterns({
   ],
 } satisfies LanguagePatterns<Record<string, never>>);
 
-// Anchor on the `HttpRequest.newBuilder().uri(URI.create("..."))` core only,
-// capturing the `.uri(...)` call as `@call`. The verb (a `.GET()/.POST()/.PUT()/`
+// Match a bare `.uri(URI.create("literal"))` call on ANY receiver, capturing it
+// as `@call`. `httpClientUriRootsAtNewBuilder` (JS) verifies the chain includes
+// `HttpRequest.newBuilder()` — allowing calls BEFORE `.uri()` (`.version(v)`
+// `.uri(...)`) that the old object-direct query dropped, and rejecting a `.uri(...)`
+// on an unrelated object (e.g. WebClient). The verb (a `.GET()/.POST()/.PUT()/`
 // `.DELETE()/.HEAD()` helper, a `.method("VERB", body)` literal, or the bare-build
-// default) lives on a sibling call further UP the chain and is recovered by
-// `inferHttpClientMethod` walking up from `@call` — exactly like OkHttp's
-// `@call` + `inferOkHttpMethod`. One `.uri(...)` per chain = one match; the walk
-// is transparent to intervening `.header()/.timeout()/.version()` calls AFTER
-// `.uri()`, so a header/timeout hop before the terminal no longer drops the
-// contract. (A call BEFORE `.uri()` — `.version(V).uri(...)` — is outside this
-// root and stays unmatched, a separate pre-existing limitation; see OK_HTTP
-// PATTERNS comment for the OkHttp dual.) Matching `.uri(...)` regardless of a
-// trailing `.build()` mirrors the accepted OkHttp over-match posture.
+// default) is recovered by `inferHttpClientMethod` scanning the whole chain.
+// Matching `.uri(...)` regardless of a trailing `.build()` mirrors the accepted
+// OkHttp over-match posture.
 const JAVA_HTTP_CLIENT_PATTERNS = compilePatterns({
   name: 'java-http-client',
   language: Java,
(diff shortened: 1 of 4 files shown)


gitnexus/test/unit/group/http-route-extractor.test.ts
```diff
@@ -2678,26 +2905,27 @@ class OkClient(private val client: OkHttpClient) {
...
-        // intentionally does not walk the chain to recover the verb —
-        // it emits `method: 'GET'` for every match, mirroring the Java
-        // plugin's `OK_HTTP_PATTERNS` (java.ts).
+        // DOCUMENTED ASYMMETRY (not "Java parity" anymore): OkHttp encodes the
+        // HTTP verb on a sibling call (`.post(body)` / `.delete()` / ...), not on
+        // `.url(...)`. The Java plugin now WALKS the builder chain to recover that
+        // verb (`inferOkHttpMethod` in java.ts). The Kotlin plugin
+        // (`kotlin.ts:OK_HTTP_PATTERNS`) does NOT yet — it still emits
+        // `method: 'GET'` for every match. So a polyglot repo currently emits
+        // `http::POST::/api/users` from a `.kt` source's sibling `.post()` and
```
magyargergo: **[P3] Reversed asymmetry comment.** This says `.kt`→POST / `.java`→GET, but it's the opposite: Java now infers the verb (`inferOkHttpMethod`), so `.java` emits POST, while Kotlin still defaults, so `.kt` emits GET. It also contradicts the test's own assertion ~6 lines below (`Kotlin GET-default → http::GET::/api/users`). Swap the two languages in these two lines.

`[code-read]` — also independently flagged by Codex (the one non-Claude engine).

gitnexus/test/unit/group/http-route-extractor.test.ts
```diff
@@ -1771,6 +1817,145 @@ class ApiClient {
...
+          c.role === 'consumer' &&
+          c.meta.framework === 'okhttp' &&
+          c.symbolRef.filePath.endsWith('OkHttpVerbs.java'),
+      );
+
+      // Exact set-equality pin: explicit verbs resolve; `.build()` with no verb
+      // (orders/4) AND the variable-bound `.method(verb, body)` (orders/5) both
+      // default to GET — no over-emit, no verb invented from the `verb` variable.
+      // Numeric segments normalize to {param}, collapsing same-verb paths.
+      expect(new Set(okhttp.map((c) => c.contractId))).toEqual(
```
magyargergo, line 2094: **[P2] This set-equality can't distinguish default-GET from explicit-GET.** `orders/0` (`.get()`), `orders/4` (bare `.build()` default-GET), and `orders/5` (variable-bound `.method()`) all normalize to `http::GET::/api/orders/{param}` and collapse into a single Set slot (further deduped by `dedupeContracts`). If the bare-`.build()` or variable-`.method()` branch stopped emitting, this test would still pass — two `inferOkHttpMethod` branches are unverified. Give those two cases distinct URL paths (e.g. `/api/bare-build`, `/api/dyn-verb`) and assert each `contractId` individually.

`[code-read]` — Codex + the CE testing lane both flagged this independently.

gitnexus/src/core/group/extractors/http-patterns/java.ts
```diff
@@ -381,6 +449,133 @@ function hasAnnotation(node: Parser.SyntaxNode, names: string | readonly string[
...
+ * variable-bound-verb limitation).
+ */
+function inferOkHttpMethod(urlCall: Parser.SyntaxNode): string {
+  let cur: Parser.SyntaxNode = urlCall;
+  let parent = cur.parent;
+  while (parent?.type === 'method_invocation' && methodInvocationObject(parent)?.id === cur.id) {
+    const name = methodInvocationName(parent);
+    if (name && ['get', 'head', 'post', 'put', 'delete', 'patch'].includes(name))
+      return name.toUpperCase();
+    if (name === 'method') return firstLiteralArgument(parent)?.toUpperCase() ?? 'GET';
```
magyargergo: **[P2] Variable-bound `.method(verb, body)` emits a GET contract.** When the verb is a non-literal, `firstLiteralArgument(parent)` returns null and this falls back to `GET`, asserting a GET consumer that may actually be POST/PUT/etc. Not a regression (main hardcoded GET for every OkHttp call), but it diverges from the WebClient long-form path, which *skips* a variable-bound verb rather than guessing. Consider skipping emission for the non-literal `.method()` case for consistency.

`[code-read]` — Codex rated this P1; the Claude correctness/adversarial lanes considered GET an intended fallback, so flagging for a deliberate decision rather than as a hard blocker.

gitnexus/src/core/group/extractors/http-patterns/java.ts
```diff
@@ -381,6 +449,133 @@ function hasAnnotation(node: Parser.SyntaxNode, names: string | readonly string[
...
+ */
+function extractUriComponentsBuilderPath(node: Parser.SyntaxNode): string | null {
+  if (node.type !== 'method_invocation') return null;
+  const name = methodInvocationName(node);
+  const objectNode = methodInvocationObject(node);
+  if (
+    (name === 'fromPath' || name === 'fromUriString' || name === 'fromHttpUrl') &&
+    objectNode?.text === 'UriComponentsBuilder'
+  )
+    return firstLiteralArgument(node);
```
magyargergo: **[P3] A query string in a `UriComponentsBuilder` seed produces a wrong path.** The seed returns its literal verbatim, so `fromUriString("/base?x=1").path("/sub")` → `appendPath("/base?x=1", "/sub")` = `/base?x=1/sub`, which `normalizeHttpPath` then truncates at `?` to `/base` (intended: `/base/sub`). Narrow — a query baked into the seed literal *before* a later `.path()` — and not covered by tests. Strip the query from the seed literal here (or split it off before appending).

`[reproduced]` — confirmed via a standalone node repro of `appendPath` + `normalizeHttpPath`.

magyargergo (commented):
**Tri-review of PR #2268** (re-implements #1888's Java HTTP consumer extraction onto current `main`).

**Methods — three, but independence is asymmetric.** GitNexus swarm (risk, test/CI) + Compound-Engineering personas (correctness, adversarial, maintainability, testing) + **Codex** = **6 Claude-based lanes + 1 live Codex run**. The 6 Claude lanes read the same diff under different prompts, so their agreement is "consistent across personas," not independent confirmation. **Codex is the only different engine** — Codex+Claude agreement is the strong signal here. _Disclosure: this is a self-review (reviewer = PR author)._

### What's solid (validated)
- **Correctness lane found no real bug** after live tree-sitter-java AST probes — the new helpers terminate and are null-safe on the shapes probed.
- The **FQN widening is snapshot-clean**: only the FQN limitation test changed behavior; all other Java contracts are byte-identical. It stays predicate-free (avoids the 0.21.x predicate-bucket hazard).
- The three Java-`HttpClient` pattern families are **AST-verified disjoint** — no double-emit.
- All 5 capabilities have targeted tests; **CI is fully green** (31 checks pass, 3 skipped as expected for a non-web PR; none failing).
- **No P0/P1 regression found** — everything below is edge-case / quality.

### Findings (inline)
1. **[P3 · Codex + author] Reversed asymmetry comment** — `test.ts:2916-2917` says `.kt`→POST / `.java`→GET; it's the opposite (Java now infers the verb → POST; Kotlin still defaults → GET), and it contradicts the test's own assertion ~6 lines below.
2. **[P2 · Codex + CE-testing] OkHttp verb test masks two branches** — the set-equality assertion collapses explicit `.get()`, the bare-`.build()` default-GET, and the variable-bound `.method()` cases into one `http::GET::/api/orders/{param}` slot; the default-GET and variable-verb branches could silently stop emitting and the test would still pass.
3. **[P2 · Codex] Variable-bound `.method(verb, body)` → GET contract** — `inferOkHttpMethod` falls back to GET for a non-literal verb, asserting a possibly-wrong GET consumer. Not a regression (main hardcoded GET for all OkHttp), but inconsistent with WebClient long-form, which *skips* a variable-bound verb. (Codex rated P1; the Claude correctness/adversarial lanes considered GET an intended fallback — framed P2 since it's pre-existing behavior with the path still correct.)
4. **[P3 · CE-adversarial, reproduced] Query-in-seed wrong path** — `UriComponentsBuilder.fromUriString("/base?x=1").path("/sub")` resolves to `/base` (the sub-path is glued after the query, then truncated at `?`) instead of `/base/sub`. Narrow and uncovered.

### Lower-priority (not inline)
- **[P3] Stale `kotlin.ts` source comment** (`kotlin.ts:440-443`, not in this diff) — still claims OkHttp GET-default "mirrors `java.ts:OK_HTTP_PATTERNS` … same trade-off Java has accepted," which this PR makes false. Update it alongside the test comment so the Java/Kotlin parity asymmetry is documented honestly on both sides.
- **[P3] No recursion depth guard** in `extractUriComponentsBuilderPath` (2 Claude lanes; Codex refuted) — latent stack overflow only on pathological 100s-deep builder chains; the repo has a `withNesting` precedent.
- **[accepted · 3 lanes]** FQN simple-name collision (`@com.evil.GetMapping` → false route) is the same trade-off the simple-name path + `hasAnnotation` already make, and is the intended Kotlin parity — not a new defect class.
- **[note]** `java.ts` grew 860→1096 lines; the static-path helpers could later move to their own file.

### Refuted (validation is a feature)
RestTemplate `(_) @path` over-match (the `extractStaticPathExpression` null-gate handles non-static args); `HttpClient` default-GET matching OkHttp/StringBuilder `.build()`; FQN duplicate matches; recursion non-termination on normal input.

---
_Automated multi-tool digest (6 Claude lanes + 1 Codex), critic-gated. No P0/P1/merge-blocker found; the P2/P3 items are quality/edge-case. Verify before acting._

koriyoshi2041:
I did a focused local pass on this branch. The existing route extractor suite is green for me:

```text
cd gitnexus
npm test -- --run test/unit/group/http-route-extractor.test.ts
# 1 file passed, 143 tests passed
```

I also independently checked the UriComponentsBuilder query-seed edge noted above by adding a temporary assertion for:

```java
UriComponentsBuilder.fromUriString("/base?x=1").path("/query-sub").build().toUriString()
```

Expected `http::GET::/base/query-sub`, but the focused test failed with that contract missing. So the P3 is real, but it looks narrow: the current happy path for `fromUriString("/base").path("/sub").queryParam(...)` still passes, and CI is otherwise clean.

magyargergo (commented):
# Tri-review digest — PR #2268 "feat(group): expand Java HTTP consumer extraction"

Repo `abhigyanpatwari/GitNexus` · diff base `dbd4e1c9` → head `a74461ea` · 4 files (`java.ts` +215, new `java-static-path.ts` +123, `kotlin.ts` comment-only, `http-route-extractor.test.ts` +358).

## 1. Engines & method

- **Claude** ran **swarm + ce**. Swarm: security-boundary (COMPLETE/clean), risk-architect (PARTIAL — captured conclusion), test-ci (ended mid-investigation; domain cross-covered by ce-testing + ce-correctness which actually ran the suite). CE personas: correctness, adversarial, api-contract, testing, maintainability, project-standards — all live.
- **Codex / GPT-5.5** ran **swarm + ce + adversarial**. **Swarm (B1) EVAPORATED** (stream `response.failed` mid-turn — no synthesized findings; treat as unavailable). **ce (B2) LIVE** (completed despite Azure 404 / reconnect turbulence; some validator subagents 404'd, main thread manually validated from code). **adversarial (B3) LIVE** (verdict needs-attention, one [medium] finding).
- **Cross-engine corroboration is available** for the headline (Claude + Codex B2 + Codex B3 independently reached it).
- **Self-review disclosure:** the reviewer (magyargergo) is the PR author (head repo is the `abhigyanpatwari` fork). Findings are kept evidence-grounded; treat the digest as input, not an external sign-off.

## 2. What's solid / validated (credit)

A large block of suspected defects was **constructed as concrete snippets and refuted** across both engines — these are validation wins, not silence:

- **Recursion overflow** in `extractUriComponentsBuilderPath` — REFUTED. `MAX_BUILDER_DEPTH=100`, all three recursive branches pass `depth+1`, overflow returns clean `null` (no partial/misleading path). [reproduced, both engines]
- **`inferOkHttpMethod` non-termination / verb-from-unrelated-chain** — REFUTED. Up-walk is gated by node-identity receiver check `methodInvocationObject(parent)?.id === cur.id`; monotonic toward root; acyclic AST. [reproduced]
- **ReDoS** — REFUTED. All `#match?` / `WEB_CLIENT_LONG_VERB_RE` regexes are anchored fixed-string alternations, no nested quantifiers.
- **Java-HttpClient three-family double-emit** — REFUTED. Families are disjoint (terminal call's `object:` differs); each input produces exactly one detection. [reproduced]
- **RestTemplate dynamic-arg over-match** (the `(_) @path` widening) — REFUTED. `extractStaticPathExpression` nulls every genuinely-dynamic arg (`URI.create(var)`, `"/a"+var`, `String.format(...)`, bare identifier). [reproduced]
- **Query-in-`.path()` corruption / wrong truncation** — REFUTED. Seed `split('?')[0]` plus the downstream global query-strip cover both positions → correct path. [reproduced]
- **FQN double-route / crash / `@com.evil.GetMapping` widening** — REFUTED. `name:` is a field constraint so a deep FQN yields exactly 1 match; `simpleName` reduces correctly; **predicate-free node-type alternation `[(identifier)(scoped_identifier)]` — does NOT reintroduce the tree-sitter 0.21 predicate-bucket hazard**. [reproduced]
- **Path-traversal** — REFUTED. Extracted strings are HTTP-path metadata, never used in fs/exec.

Clean lanes:
- **Security boundary: clean.** No secrets/injection/unsafe ops; no Docker/workflow/YAML changes; hidden-Unicode hygiene clean (`git diff --check` clean, no bidi controls).
- **API contract: no breaking change.** Confidence values consistent (okhttp 0.7, java-http-client 0.65, spring-rest-template 0.7 — no drift); framework tags unchanged; verb casing always uppercase-or-null; **FQN widening is byte-identical for existing contracts** (simple-name annotations map to themselves).
- **Project standards: clean.** No `any`/`as any`/non-null assertions; no `if`-branching in new tests; ESM `.js` extensions present; CHANGELOG untouched; shared `core/ingestion/` not modified; language logic correctly placed in the `http-patterns/` extractor layer (not shared core).
- **Accepted/documented trade-offs (NOT defects):** the FQN simple-name collision (`@com.evil.GetMapping`, commit `334f4cd5`) and the Java↔Kotlin OkHttp verb asymmetry (Java now infers verbs, Kotlin still defaults GET — `kotlin.ts:433-450`) are both documented, intended follow-ups.

## 3. HEADLINE — Java HttpClient intervening-call under-extraction (STRONG, cross-engine)

**Corroboration: Codex B2 (ce) + Codex B3 (adversarial) + Claude ce-correctness F1 + Claude ce-adversarial residual** — the strongest cross-engine finding in this review.

- **File:** `gitnexus/src/core/group/extractors/http-patterns/java.ts`, the three new HttpClient pattern families (`JAVA_HTTP_CLIENT_PATTERNS` @303, `JAVA_HTTP_CLIENT_METHOD_PATTERNS` @331, `JAVA_HTTP_CLIENT_DEFAULT_GET_PATTERNS` @361).
- **Severity:** **P2** · **Trigger:** any `HttpRequest.newBuilder().uri(URI.create("/p"))` chain with an intervening builder call before the terminal — e.g. `.uri(URI.create("/api/users/1")).header("Accept","application/json").build()` or `.uri(...).timeout(t).method("PATCH", body).build()`.
- **Mechanism [code-read, verified in worktree]:** all three queries require the terminal call (`@http_method` verb helper / `@method` / `@build`) to be the `name:` of a `method_invocation` whose `object:` is the `.uri(URI.create(...))` call **directly** (java.ts:316–322, 344–351, 374–381). With an intervening `.header(...)`, `.build()`'s `object:` is the `.header()` call, not `.uri()` → **no family matches → the consumer contract is silently dropped** (under-extraction; never a wrong contract).
- **Scope (important — position matters):** the headline rigidity is **post-`.uri()`** and affects the **three Java HttpClient families** (propagated by this PR's two new families). OkHttp behaves differently and must not be conflated:
  - **(a) Post-`.url()` is fine for OkHttp.** `inferOkHttpMethod` (java.ts:472–485) walks UP the chain transparently — its `while` loop advances past any unrecognized intervening call (`.addHeader()`, `.tag()`), so `.url("/x").addHeader(...).post(body).build()` still resolves POST. The headline (post-`.uri()`) pattern does **not** affect OkHttp verb/path.
  - **(b) OkHttp has a SEPARATE, PRE-EXISTING rigidity — *before* `.url()`.** ce-correctness F1 + ce-adversarial both reproduced that `new Request.Builder().header("a","b").url("/x")…` yields **0 matches** — a call before `.url()` breaks the `object: (object_creation_expression "Request.Builder")` constraint. This is a real but **pre-existing** gap (the OkHttp path query was already `object:`-direct before this PR), **not introduced here**, and at a *different* chain position than the headline. So ce-correctness was **not** over-stating — it correctly flagged a distinct OkHttp gap; my earlier "refuted/over-statement" framing was wrong and is corrected here.
- **Not a regression:** the pre-existing verb-helper family already required the direct shape; this PR propagates the same rigidity to the two new families. Postable because headers/timeouts are near-universal in real `HttpClient` code, so the realized miss-rate is non-trivial.
- **Fix (follow-up, not a merge blocker):** mirror `inferOkHttpMethod` — anchor on the `.uri(URI.create(...))` call (or `.build()`), walk outward through neutral builder calls (`header`/`timeout`/`version`), record a literal verb if present, default GET only at build. Add regression tests with a header/timeout hop before `.build()`/`.method(...)`. At minimum, document the direct-object limitation inline (parity with how the Kotlin OkHttp verb-walk limitation is documented).
- **Inline:** YES — anchor on the `JAVA_HTTP_CLIENT_DEFAULT_GET_PATTERNS` family const (java.ts:361, added line) or the "Default GET form" comment block.

## 4. Inline-candidate findings

| # | Finding | File:line (RIGHT) | Sev | Tag | Inline? |
|---|---------|-------------------|-----|-----|---------|
| H | HttpClient intervening-call under-extraction (headline §3) | java.ts:361 | P2 | code-read | **inline** |
| F2 | Empty-verb `.method("", body)` → malformed `http::::/path` | java.ts:967 | P3 | reproduced | inline (low-key) or body |

**F2 — empty-string verb literal emits a malformed contract.** Corroboration: **Claude ce-correctness F2 + Claude ce-adversarial F1** (single-engine — Claude only — but **reproduced** end-to-end).
- **Trigger:** `.method("", body)` (OkHttp loop java.ts:925, and HttpClient `.method` loop java.ts:961–975).
- **Mechanism [reproduced in node]:** `unquoteLiteral('""')` returns `""` (not `null`); `"" ?? null` keeps `""` (empty string is not nullish) and `""?.toUpperCase()` → `""`. The guards `if (method === null)` (line 925) and `if (httpMethod === null || path === null)` (line 967) do **not** catch `""`, so a detection with `method: ''` is pushed → contractId `http::::/path`.
- **Severity:** **P3** — real but **extremely narrow**: no real OkHttp/HttpClient code writes an empty verb literal (rejected at runtime). Flagged for completeness.
- **Fix:** guard for falsiness — `if (!httpMethod) continue;` (and `inferOkHttpMethod` return `null` for a falsy verb).
- **Inline:** optional. Given the narrowness, a low-key inline on java.ts:967 (the `httpMethod === null || path === null` guard, an added line) is reasonable; equally defensible as a body note. **Recommend body** to keep the inline set focused on the headline.

## 5. Lower-priority / body items + test gaps

**B1 — `.path()` slash-join vs Spring semantics (Codex B2 ONLY — a Claude blind spot; credit Codex).**
- **File:** `java-static-path.ts:45` (`appendPath`) used by both `.path()` (line 87) and `.pathSegment()` (line 90).
- `appendPath` always inserts exactly one slash. Real Spring `UriComponentsBuilder.path("users")` **raw-appends** (`/api` → `/apiusers`), while `.pathSegment("users")` slash-joins (`/api` → `/api/users`). So `fromPath("/api").path("users")` yields `/api/users` from the extractor but `/apiusers` in real Spring. [code-read, verified]
- **Severity: P3, body (not inline)** — single-engine, **only triggers on a no-leading-slash `.path()` arg** (all the PR's tests use leading slashes → untested + uncommon), and the extractor's slash-joined output is arguably the *more useful* route anyway. Debatable-whether-defect. Credit Codex for the cross-engine-blind-spot catch.

**B2 — Maintainability (cross-engine: Claude ce-maintainability + Codex B2 maintainability subagent). P2 quality, body.**
- `java.ts` is **1015 lines** [verified `wc -l`] — 15 over the "<~1000" goal its own extracted module's comment states (`java-static-path.ts:10`). Note: NOT a written standards rule (no file-length cap in AGENTS/CLAUDE/GUARDRAILS/CONTRIBUTING), so project-standards correctly files it as informational; maintainability files it P2.
- Three near-identical HttpClient emission loops (java.ts:945/961/978) each repeat `confidence: 0.65`; query duplication is domain-justified (tree-sitter template strings).
- Three single-consumer exports in `java-static-path.ts` (`methodInvocationName`/`methodInvocationObject`/`firstLiteralArgument`).
- **Clean co-fix:** move `inferOkHttpMethod` into `java-static-path.ts` (its only deps already live there) → drops java.ts to ~988 (meets the goal) and lets the three helpers de-export. Author already did one extraction pass, so this is structural/quality, body not inline.
- Minor (P3): `@http_method` capture name is overloaded (identifier in one pattern, string_literal in another) — rename to `@verb_string` in the `.method` family for clarity.

**Test gaps (P3, body):**
- No test exercises an **intervening builder call** (`.header()`/`.timeout()`) between `.uri()` and the terminal — the headline silent-drop is unpinned. Adding one would pin it as the accepted limitation.
- No **count guard** on the query-seed-strip test (java-route test ~1910–1940): two `toBeDefined` asserts, no `toHaveLength(2)` / `toBeUndefined('/base')` — a hypothetical double-emit would pass (not reachable today).
- No `exchange()` + `UriComponentsBuilder` builder-arg end-to-end test (covered indirectly via `getForObject` through the same `extractStaticPathExpression`).
- No `.method("")` empty-verb test (F2) and no `.method("CUSTOM_VERB", body)` pass-through test.
- Residual (pre-existing, not introduced): protocol-relative host `fromUriString("//host/api")` leaks the host as the first path segment (`normalizeConsumerPath` only strips `^https?://`).
- Residual (pre-existing, not introduced): **OkHttp pre-`.url()` rigidity** — a builder call before `.url()` (`new Request.Builder().header("a","b").url("/x")…`) yields 0 matches because the OkHttp path query requires `.url()`'s `object:` to be the `Request.Builder` `object_creation_expression` directly. Reproduced by ce-correctness F1 + ce-adversarial. Separate from the headline (different chain position) and not introduced by this PR; a fair follow-up alongside the HttpClient walk.

## 6. CI status

All required checks **GREEN** at review (self-review — reviewer == author). typecheck / lint / format pass; unit suite runs and passes on ubuntu-coverage, macos-latest, windows-latest, and both packaged-install-smoke jobs; CodeQL pass; gitleaks pass; benchmarks pass. `e2e (chromium)` skipped (no web-module changes — expected). Claude ce-correctness independently ran the isolated `http-route-extractor.test.ts` → 145/145 green.

## 7. Verdict & closing

**Verdict: production-ready with minor follow-ups.** **Merge status:** clean — `MERGEABLE`, CI green, no P0/P1, no security/API-contract/type-safety issue; the open items are non-blocking quality/completeness gaps. **Blocks merge: no.**

The headline (HttpClient intervening-call under-extraction, P2) is a genuine under-extraction gap but not a regression and not a blocker; the maintainability P2 (1015-line / loop duplication) and the `.path()` slash-join and empty-verb P3s are best tracked as follow-ups. 

This is an automated multi-tool digest — **Claude** ran swarm + ce; **Codex/GPT-5.5** ran swarm (evaporated mid-run), ce, and adversarial. The one cross-engine finding (headline) had genuine Codex+Claude agreement; everything else is single-engine and labeled as such. Self-review (reviewer == author). Verify each finding against the code before acting.

gitnexus/src/core/group/extractors/http-patterns/java.ts
```diff
@@ -315,7 +319,66 @@ const JAVA_HTTP_CLIENT_PATTERNS = compilePatterns({
...
+      `,
+    },
+  ],
+} satisfies LanguagePatterns<Record<string, never>>);
+
+// Default GET form: `HttpRequest.newBuilder().uri(URI.create("...")).build()` with
+// no verb helper between `.uri(...)` and `.build()`. The `.build()` object must be
+// the `.uri(...)` call DIRECTLY, so a chain carrying a verb helper or `.method(...)`
+// does not also match here.
+const JAVA_HTTP_CLIENT_DEFAULT_GET_PATTERNS = compilePatterns({
```
magyargergo: **[P2 · code-read · Codex + Claude cross-engine] Java HttpClient intervening-call under-extraction.**

All three new Java HttpClient pattern families require the terminal call (`@http_method` verb helper / `@method` / `@build`) to be invoked **directly** on the `.uri(URI.create(...))` node — as this family's own comment states, "the `.build()` object must be the `.uri(...)` call DIRECTLY". A realistic chain with an intervening builder call —

```java
HttpRequest.newBuilder().uri(URI.create("/api/users/1")).header("Accept","application/json").build()
// or: .uri(...).timeout(t).method("PATCH", body).build()
```

— makes the terminal's `object:` the `.header()`/`.timeout()` call, so **no family matches and the consumer contract is silently dropped** (under-extraction only — never a wrong contract). Headers/timeouts are near-universal in real `HttpClient` code, so the realized miss-rate is non-trivial.

**Scope:** HttpClient-specific (post-`.uri()`). OkHttp's `inferOkHttpMethod` walks the chain transparently for calls *after* `.url()`, so it is unaffected by this shape. (OkHttp has a *separate*, pre-existing rigidity for a call *before* `.url()` — `new Request.Builder().header(...).url("/x")` → 0 matches — out of scope here.)

**Not a regression:** the pre-existing verb-helper family already required the direct shape; this PR propagates that rigidity to the two new families. Non-blocking, but since this PR's goal is to *expand* HttpClient extraction, the header/timeout gap is worth closing.

**Follow-up:** mirror `inferOkHttpMethod` — anchor on `.uri(...)`/`.build()` and walk outward through neutral builder calls (`header`/`timeout`/`version`), recording a literal verb if present and defaulting GET only at build — plus a regression test with a header/timeout hop. At minimum, document the direct-object limitation inline (parity with how the Kotlin OkHttp verb-walk limitation is documented).

_Corroborated by: Codex adversarial (B3) + Codex ce (B2) + Claude ce-correctness + Claude ce-adversarial._

PR #2269, opened 2026-06-22
feat(group): cross-repo call trace using PDG

magyargergo:
## Summary

Extends the `trace` MCP tool to follow a call path **across repository boundaries** within a group. When `from` and `to` resolve to different member repos, the trace stitches the two repo-local segments together over a single `ContractLink` bridge boundary (an HTTP consumer→provider link, joined on `symbolUid`), reported as a `CONTRACT_LINK` hop. With `pdg: true` (experimental, opt-in), each boundary-adjacent segment is annotated with its intra-procedural REACHING_DEF data-flow.

Before this, `trace` stopped at the repo edge. The headline of the follow-up work in this PR is that it now **actually resolves end-to-end**: HTTP contracts carry a real `symbolUid` (they used to be empty), so the bridge join connects — which also makes `impact @group` fan out across the boundary (it was 0 cross-repo hits before).

```mermaid
graph LR
  subgraph FE["frontend repo"]
    A["caller()"] -->|CALLS| B["fetchUsers()"]
  end
  subgraph BR["group bridge.lbug"]
    C["Contract (consumer)<br/>http::GET::/api/users"] -->|ContractLink| D["Contract (provider)<br/>http::GET::/api/users"]
  end
  subgraph BE["backend repo"]
    E["listUsers()"] -->|CALLS| F["queryUsers()"]
  end
  B -. symbolUid .-> C
  D -. symbolUid .-> E
```

`from --(local trace)--> consumer --(ContractLink)--> provider --(local trace)--> to`

## Request flow: symbol-to-symbol trace

```mermaid
sequenceDiagram
  actor U as Caller
  participant T as trace tool
  participant G as runGroupTrace
  participant M as members (per-repo lbug)
  participant BR as bridge.lbug
  U->>T: trace @group from=fetchUsers to=listUsers (pdg)
  T->>G: groupTrace(...)
  G->>M: resolveSymbol(from) across ALL members
  G->>M: resolveSymbol(to) across ALL members
  Note over G: from in frontend, to in backend → cross-repo
  G->>BR: CY_CROSSINGS_BETWEEN(frontend, backend)
  BR-->>G: ContractLinks ordered by confidence (LIMIT-bounded)
  loop each crossing (memoized per endpoint uid)
    G->>M: trace(from → consumerUid)   [segment A]
    G->>M: trace(providerUid → to)     [segment B]
  end
  opt pdg:true
    G->>M: REACHING_DEF walk on boundary-adjacent segments
  end
  G-->>U: ok: hops + CONTRACT_LINK hop + dataFlow + notes
```

## Request flow: destination trace (reaching anonymous handlers)

An anonymous handler (`router.get('/x', (req,res) => ...)`) has no symbol node, so it can never be named as `to`. Omitting `to` follows the consumer's outgoing HTTP call to wherever it lands, reported by route + file even when the handler is unnameable. Ambiguity is reported, never silently resolved.

```mermaid
sequenceDiagram
  actor U as Caller
  participant G as runGroupTrace
  participant BR as bridge.lbug
  U->>G: trace @group from=fetchUsers   (no `to`)
  G->>BR: CY_CROSSINGS_FROM(frontend)
  BR-->>G: outgoing ContractLinks
  loop each crossing
    alt consumer has resolved uid (PRECISE)
      G->>G: trace(from → consumerUid) connects?
    else empty uid (FILE-level)
      G->>G: does `from` live in the consumer file?
    end
  end
  alt one distinct endpoint
    G-->>U: ok → endpoint route + file (anonymous note)
  else multiple endpoints
    G-->>U: ambiguous (role: to, candidate routes)
  else none
    G-->>U: not_found
  end
```

## How a contract gets its `symbolUid`

The root-cause fix. HTTP contracts hardcoded `symbolUid:''`, so neither trace nor impact could traverse the bridge. Now each detection resolves to a real symbol, falling through on a miss:

```mermaid
flowchart TD
  D["HTTP detection (consumer / provider)"] --> Q1{"Route node has<br/>handlerSymbolId?<br/>(2265 provider path)"}
  Q1 -->|yes, provider fast path| U1["symbolUid = handlerSymbolId<br/>(authoritative, no source parse)"]
  Q1 -->|no| Q2{"named handler?"}
  Q2 -->|yes| U2["resolveSymbolByName<br/>over File-[DEFINES]->symbol"]
  Q2 -->|no| Q3{"call-site line set?"}
  Q3 -->|yes| U3["resolveContainingSymbol<br/>innermost span containing the call"]
  Q3 -->|no| U4["empty uid"]
  U4 --> F["cross-trace file-level boundary fallback<br/>(flagged in the notes channel)"]
```

Resolution matches symbols by `filePath` across `Function`/`Method`/`CodeElement`, so it reaches methods nested in classes (Java/Kotlin), not just top-level functions. The call-site line is wired through all bundled language plugins (Node/TS, Python, Go, PHP, Kotlin, Java).

## What changed

**Cross-repo trace engine** — `GroupService.groupTrace` + the pure `runGroupTrace` stitching engine in `cross-trace.ts`; a strictly-typed discriminated `GroupTraceResult` (`ok | not_found | ambiguous | error`), per-hop `repo` tags, typed `crossings[]`, centralized degraded-state notes. The `trace` tool routes to `groupTrace` on `@group`; single-repo `trace` is byte-identical (regression-tested). `pdg:true` adds an anchored, bind-param-only REACHING_DEF walk; absent a `--pdg` layer it degrades to call-level hops with a note.

**symbolUid resolution (root cause)** — populate the contract `symbolUid` via containment (the function the call lives in) + named-handler resolution, over the correct `File-[DEFINES]->symbol` edge. This is what makes the bridge join connect for HTTP, and it fixes the identical `impact @group` gap. Found en route: the pre-existing graph resolver queried `CONTAINS` (the File→Folder edge), so it never resolved a uid.

**Destination trace** — omit `to` on an `@group` trace to follow a consumer to its HTTP endpoint, including fully anonymous handlers (reported by route + file:line). Precise hits beat the weak file-level tier; multiple endpoints → `ambiguous` with candidates rather than a silent pick.

**Bridge reopen fix** — `closeBridgeDb` issued `CHECKPOINT` on *every* handle including read-only ones; a CHECKPOINT on a read-only connection leaves a WAL/shadow lock artifact that made the next in-process read-only open of the same `bridge.lbug` fail — breaking *repeated* `@group` trace/impact in a long-lived MCP server. Fix: a `BridgeHandle._readOnly` flag and skip CHECKPOINT for it. Linux/macOS in-process reopen now works; **Windows in-process reopen remains a documented limitation** (an attempt to close it with a post-close handle-release probe was reverted after Windows CI proved it ineffective — the reopen tests stay Windows-skipped).

**Bounded crossing fan-out** — `CY_CROSSINGS_BETWEEN` / `CY_CROSSINGS_FROM` use `ORDER BY confidence DESC` + a `LIMIT` (strongest crossings kept; truncation surfaced in `notes[]`, never silent), and each repo-local segment is memoized by endpoint uid so the cost is per distinct endpoint, not per crossing.

## Review and hardening

- **Two-engine tri-review** (Claude swarm + ce, Codex GPT-5.5 swarm + ce + adversarial). Every finding fixed: a P1 destination wrong-endpoint (file-fallback made `trace(from→from)` trivially succeed → highest-confidence sibling won) now reports `ambiguous`; name resolution no longer blocked when a provider has no line; `resolveContainingSymbol` probes the base-correct `line-1` first; anonymous Express handlers emit `name:null` so they cannot borrow an unrelated `handler` function; the first-symbol-in-file guess and the dead `CONTAINS`/`pickSymbolUid` fallback were removed. Cypher injection was refuted by both engines (all bind-params; the only interpolation is an integer-clamped `LIMIT`).
- **Contributor review** (@koriyoshi2041): degraded-member notes were dropped on a *successful* group trace — `ok` only means "unique among queryable members." Fixed so same-repo and cross-repo successes carry the degraded note; added their two suggested regression guards (destination success + degraded member; multiple precise destination hits → ambiguous).

## Kept current with main (#2265)

Merged `main`, which landed #2265 ("skip source parse for graph-covered route files", #2138 Part 2) in the same subsystem. Combined both: `extractProvidersGraph` carries the new `coveredFiles` parse-skip optimization AND the symbolUid resolver; the `handlerSymbolId` fast path is authoritative for providers (its display-name lookup uses the correct `CONTAINING_QUERY`, not the File→Folder `CONTAINS`), with the containment resolver as the legacy fallback.

## Security

Bind-param-only throughout. The bridge queries use `$fromRepo`/`$toRepo`/`$filePath`; the PDG anchor `idPrefix` is a bound DB-derived param; the only interpolation is an integer-clamped `LIMIT`. Group name is regex-validated.

## Test plan

- Unit + integration: cross-trace stitching, destination trace (named / anonymous / ambiguous / degraded), symbolUid containment resolution, the bridge reopen regression, and #2265's route-handler-symbol roundtrip — green.
- `bench/cross-repo-trace/verify.mjs`: a self-contained harness that drives the real `analyze --pdg → syncGroup → trace/impact` pipeline across 4 scenarios (named precise, anonymous destination, impact fan-out, multi-language Python + file-fallback). Expected verdict **10/10**.

<sub>This is the bounded, experimental-mode increment of the cross-repo PDG impact work deferred in the unified-PDG plan: cross-repo transport is `ContractLink`, PDG is enrichment only (never crosses the boundary), single cross-hop clamped at `MAX_SUPPORTED_CROSS_DEPTH = 1`. Full cross-program (SDG-like) data flow stays deferred.</sub>


magyargergo (commented):
## Tri-review digest — cross-repo call trace using PDG

**Engines & lanes (two engines, same scaffolding — agreement across them is the strong signal):**
- **Claude** — GitNexus PR swarm (risk-architect, test-ci-verifier, security-boundary) + ce-code-review personas (correctness, reliability, maintainability, adversarial, api-contract).
- **Codex (GPT-5.5 / xhigh)** — `$gitnexus-pr-review` (Solo) + `ce-code-review` (report-only) + `adversarial-review`. **All three Codex legs returned real findings** — genuine both-engine review.
- Reviewed at base `723d9e0b`; the live head has since moved past it (fixes below).

> Author note: findings were triaged and the merge-blocking + cross-engine-corroborated ones fixed in follow-up commits on this branch; this digest records what the review found and what was addressed.

### 🔴 Headline — Codex caught a live CI failure the Claude lanes missed
The Codex swarm leg fetched the live check state and found **`tests / windows-latest (platform-sensitive)` = FAILURE** (4 `bridge-db` reopen tests). Root cause: the branch had flipped the bridge close-then-reopen tests to run on Windows, but the writable-close→read-open handoff does not release the OS file handle there. **Fixed** (`096edb84`): reverted the Windows test-flip and removed the ineffective close-side probe; kept the load-bearing read-only `CHECKPOINT`-skip (which fixes the Linux/macOS in-process reopen — the reproduced bug).

### ✅ Solid, cross-engine-validated
- **No P0/P1.**
- **Cypher injection refuted four ways** (security-boundary + ce-adversarial + risk-architect + codex): all of `from`/`to`/`repo`/`crossDepth`/`limit` flow through bind params; the only interpolated `LIMIT` is integer-clamped/constant; `idPrefix` is a bound DB-derived value.
- **Read-only `CHECKPOINT`-skip is correct** — writable handles still checkpoint (durability preserved); confirmed by four lanes.
- Hop/edge alignment (`edges == hops-1`, `CONTRACT_LINK` between segments) and the `segA`/`segB` memoization keys are correct (correctness lane); `resolveBridgeNeighbors` extraction is byte-identical (risk-architect).
- **Refuted & dropped:** the "new close-side probe vs existing open-side retry — redundant/deadlock?" suspicion — ce-reliability verified it's strictly additive, no race.

### ✅ Resolved past the review base
| # | Severity | Finding | Lanes | Fix |
|---|---|---|---|---|
| 1 | P1 | Windows CI failing (bridge reopen on Windows) | codex-swarm | `096edb84` |
| 2 | P2 | Member-resolution errors masked as `not_found` | codex-swarm+adversarial, ce-reliability | `38e4ccee` — `degradedMembers` note |
| 3 | P2 | Crossing cap can hide the only path; no `truncated` flag | codex-adversarial, ce-adversarial, ce-api-contract | `38e4ccee` — `truncated:true` on capped `not_found` |
| 4 | P2 | `crossDepth` schema `maximum:10` vs clamp-to-1 | ce-api-contract | `38e4ccee` — schema `maximum:1` |
| 5 | P2 | ARCHITECTURE.md: `@group/memberPath` advertised as scoping `trace` (it's advisory) | codex-ce | `956598c8` — doc clarified |

### 🟡 Open follow-ups (P2/P3 — non-blocking)
- **[P2] maintainability** — `_pdgFlowsForGroupImpl` (`local-backend.ts`) duplicates the `resolveBlockAnchor` span-anchor clause (incl. the `+1` offset). Extract a shared `resolveSpanAnchor` helper.
- **[P3] maintainability** — the `method==='trace'` group dispatch forwards 11 params by hand; a future param can be silently dropped. Forward `{ name, ...rest }`.
- **[P3] api-contract** — the group `status:'ok'` shape adds `group`/`crossings`/`notes`/per-hop `repo` vs single-repo `trace`; additive (MCP-tolerant) but a strict CLI formatter could diverge. Document.
- **[P3] adversarial** — PDG enrichment widens to a whole-file `id`-prefix match when a symbol has no valid line span; mirrors existing `resolveBlockAnchor` fallback, bounded by `LIMIT`. Treat absent-span as cannot-anchor.
- **[P3] docs** — ARCHITECTURE.md links `docs/plans/2026-06-18-002-…-plan.md`, which is gitignored (`docs/`) so not resolvable from the repo. Acknowledged: it's a rationale pointer; the deferral content is in the PR body.
- **[P3] tests** — pre-existing `as any` (`calltool-dispatch.test.ts:629`) and a teardown `if` (`cross-trace-e2e.test.ts`) — minor convention nits.

### Classifications
- **Merge state:** `mergeable` (CI re-confirmed on the latest head — see below).
- **Branch hygiene:** `clean feature/fix PR` — single coherent feature + its review/robustness follow-ups; no unrelated churn; CHANGELOG untouched.
- **Verdict:** `production-ready with minor follow-ups` — no P0/P1; Cypher-injection-safe; the merge-blocking Windows failure and all cross-engine-corroborated P2s are fixed on this branch; the remaining items are P2/P3 maintainability/contract/doc polish.

**CI:** All green on `956598c8` — **28 passed, 1 skipped, 0 failed**, `MERGEABLE` / `CLEAN`. The previously-failing `tests / windows-latest (platform-sensitive)` is now **SUCCESS** (macOS + CI Gate green too).

_Automated multi-tool digest (Claude + Codex/GPT-5.5). Verify before acting._

koriyoshi2041:
I did a focused local pass on the new group trace surface. The existing group trace/dispatch slice is green for me:

```text
cd gitnexus
npm test -- --run test/unit/group/cross-trace.test.ts test/unit/calltool-dispatch.test.ts -t "group-mode trace|runGroupTrace"
# 2 files passed; 13 tests passed, 173 skipped by filter
```

One edge I think is worth carrying into the follow-up list: degraded member resolution is only surfaced on `not_found`, not on successful `ok` resolution. I added a local probe where `reg-fe` resolves `checkout -> callUsers` and traces successfully, while `reg-be` throws from `resolveSymbol`. The result was still:

```text
{ status: "ok", notes: [] }
```

Since group trace resolves names across all members, that `ok` is really "unique among queried members" rather than authoritative uniqueness for the whole group. If the unavailable member also contains `checkout` or `callUsers`, the user should probably see a degraded-member note, or be nudged to rerun with `from_uid`/`to_uid`. Small fix direction: carry `fromRes.degraded`/`toRes.degraded` through successful same-repo and cross-repo results too, not just the `not_found` branches.

I removed the probe after validating it; local worktree is clean.

magyargergo:
Thanks for taking the time to dig into this. You are right, and it is a great catch.

I reproduced it locally with the setup you described. The success path in runGroupTrace was seeding an empty notes array before handing off to the same repo and cross repo stitchers, so the degraded information from resolveSymbol was getting dropped on any ok result. Your framing is exactly right: an ok here really only means unique among the members we were able to query, so if the member that threw also held one of the endpoints, that result is not authoritative.

I pushed a fix in 014dfe5d. It seeds the dispatch notes with the degraded members from both the from and to resolution, so a successful same repo or cross repo trace now carries the same "could not be queried" note that the not_found branches already had. I also added a regression test that mirrors your probe: the frontend member resolves the trace while the backend member throws, and the ok result now includes the degraded note naming app/backend.

Worth noting that with the symbolUid work in this PR the common case is a clean single member resolution, but the degraded note is still the right behavior any time a member cannot be queried, so your point holds either way.

While I have you: do you have any recommendations for what else is worth testing on this surface, ideally with the expected output you would want to see? I want to make sure the degraded and ambiguous destination paths behave the way you would expect, so concrete cases would be really helpful.

koriyoshi2041:
Thanks for the quick fix — I pulled the latest head (`2aa16a2`) and reran the focused slice:

```text
cd gitnexus
npm test -- --run test/unit/group/cross-trace.test.ts test/unit/calltool-dispatch.test.ts -t "group-mode trace|runGroupTrace"
# 2 files passed; 18 tests passed, 174 skipped by filter
```

The new degraded-success regression covers the edge I had in mind. The other cases I would still pin are both around destination trace:

1. `destination` success with a degraded member. Example: `reg-fe` resolves `from` and follows an outgoing HTTP link to an anonymous backend handler, while `reg-be` throws during `resolveSymbol`. Expected: `status: "ok"`, normal anonymous-handler endpoint/crossing, and the `could not be queried` note. That keeps the no-`to` path aligned with explicit `to` traces: success is still only authoritative among queryable members.

2. multiple precise destination hits. Example: one `from` reaches two consumer symbols with non-empty `consumerUid`, each linked to a different provider route. Expected: `status: "ambiguous"`, `role: "to"`, and candidates for both route ids. The current file-level ambiguous case is good; this would separately pin the stronger precise tier so a future change does not silently pick the highest-confidence destination.

I would treat both as regression guards, not blockers. The latest shape looks much safer than the first version I tested.

magyargergo:
Both of these are great, thanks for the concrete cases with expected output. I added them as regression guards in 6149c2e7.

1. destination success with a degraded member: reg fe resolves from and follows the link to an anonymous backend handler while reg be throws during resolveSymbol; the result is status ok with the anonymous endpoint and crossing, plus the could not be queried note, so the no to path stays aligned with explicit to traces.
2. multiple precise destination hits: dispatch reaches two consumers with resolved uids, each linked to a different provider route; the result is status ambiguous, role to, with candidates for both route ids. It sits alongside the existing file level ambiguous test and pins the stronger precise tier so a future change cannot silently pick the highest confidence destination.

Both passed against current behavior on the first run. Full group suite is green. Appreciate you turning the review into concrete guards.

PR #2271, opened 2026-06-22
fix(lang-kotlin): support `fun interface` extraction via tree-sitter-kotlin re-vendor

### LINKED ISSUE #87: Functional interface parsing has ERROR
There is an error parsing functional interfaces.

For example (from kotlinlang.org):

```kotlin
fun interface IntPredicate {
   fun accept(i: Int): Boolean
}
```

The node which should represent the body/`accept` method is an error, rather than correctly parsing out the abstract function.

glier:
## Problem

The vendored `tree-sitter-kotlin@0.3.8` (fwcd) parses a Kotlin **functional (SAM) interface** — `fun interface Foo { ... }` — as an `ERROR` node and drops the declaration **and** its abstract method. So `fun interface` types are never extracted (no `Interface` node, no method).

Repro (old grammar):

```
fun interface Clicker { fun onClick(): Boolean }
```
→ `(source_file (ERROR "fun" (user_type (type_identifier)) (simple_identifier)) (lambda_literal ...))`

The fix exists upstream: **[fwcd/tree-sitter-kotlin#169](https://github.com/fwcd/tree-sitter-kotlin/pull/169)** (closes [#87](https://github.com/fwcd/tree-sitter-kotlin/issues/87)), merged to `main` 2025-04-25 — but it is **not in any npm release** (latest tag is still `0.3.8`, from 2024-08-03; `main` is the unreleased `0.4.0`).

## Change

Re-vendor the grammar from the unreleased fwcd `main` commit [`c8ac3d26`](https://github.com/fwcd/tree-sitter-kotlin/commit/c8ac3d2627240160b999a2c100de3babbdb8f419):

- Refresh `src/{parser.c,scanner.c,node-types.json,tree_sitter/*.h}` and `bindings/node/index.js`; bump the vendor `version` `0.3.8 → 0.4.0`; record the pinned SHA + rationale in `_vendoredBy` and the vendor `README.md`.
- Switch the prebuild workflow's kotlin registry `kind: 'npm' → 'vendored'` — the fix is unreleased on npm, so prebuilds must build from the vendored C source (like swift/dart/proto).
- Add a `hold` to `.github/vendored-grammars.json` so the weekly auto-update monitor does **not** revert the pin: `isNewer` is strict-inequality, so `0.3.8 !== 0.4.0` would otherwise re-vendor the broken npm `0.3.8` over `0.4.0`.
- Add 3 regression tests + a fixture asserting fun interfaces extract as `Interface` nodes with their methods, and that plain-interface heritage still resolves.

**No `KOTLIN_QUERIES` change needed:** the new grammar models `fun interface Foo` as a `class_declaration` with an `"interface"` keyword child (plus an extra `"fun"` modifier child), which the existing interface rule `(class_declaration "interface" (type_identifier) @name)` already matches.

## Verification

Full Kotlin suite green against the new grammar (built locally for darwin-arm64):

- `test/unit/{kotlin-scope-captures,kotlin-static-marker}.test.ts`, `test/unit/cfg/kotlin-visitor.test.ts`, `test/integration/resolvers/{kotlin,kotlin-coverage}.test.ts` → **300 + 233 pass** (incl. the 3 new fun-interface cases).
- `test/unit/{vendored-grammars,grammar-update-monitor}.test.ts` → **28 pass**.
- Confirmed every node type referenced by `KOTLIN_QUERIES` is present in the new `node-types.json` (no silent node-type drift from the 0.3.8→main jump).

## ⚠️ Prebuilds / CI sequencing (please read before merge)

`prebuilds/` are **intentionally not in this PR** — only `darwin-arm64` is buildable locally; the other 5 platforms require the cross-build CI.

The version bump (`0.3.8 → 0.4.0`) auto-triggers **`build-tree-sitter-prebuilds`**, which regenerates all 6 platform binaries from the vendored source. Until those binaries land, `node-gyp-build` keeps loading the committed `0.3.8` prebuild, so:

- **CI for the new kotlin tests will be RED** on this PR, and
- the grammar change is **inert at runtime** until prebuilds are regenerated.

Because this is a **fork PR**, the prebuild workflow's `aggregate` (auto-PR) job is skipped — it will only upload artifacts. **Recommended:** a maintainer runs `build-tree-sitter-prebuilds` (workflow_dispatch, `grammars=kotlin`) to regenerate + commit the 6 binaries + `SHA256SUMS`, and that lands first or together with this PR. (`SHA256SUMS` is only written by that workflow, never verified at install/runtime.)

### DIFF
--- a/.github/scripts/test_check_tree_sitter_upgrade_readiness.py
+++ b/.github/scripts/test_check_tree_sitter_upgrade_readiness.py
@@ -61,9 +61,10 @@ def _physical_vendor_grammars() -> set[str]:
 def _render_report() -> tuple[str, int]:
     """Run main() with network mocked to mirror PRODUCTION; return (md, exit_code).
 
-    - npm grammars resolve to a permissive "Ready" peer dep, so the ONLY blocker
-      left is the held vendored tree-sitter-c — letting us assert the hold is
-      load-bearing (exit code stays non-zero because of it).
+    - npm grammars resolve to a permissive "Ready" peer dep, so the only blockers
+      left are the held vendored grammars (tree-sitter-c, tree-sitter-kotlin) plus
+      the intentionally-pinned tree-sitter-cpp — letting us assert holds are
+      load-bearing (exit code stays non-zero because of them).
     - npm_view_json records its calls so we can prove vendored grammars are never
       npm-queried.
     - fetch_text mirrors the real workflow: upstream parser.c resolves to a real
@@ -363,13 +364,15 @@ class ReportRendering(TestCase):
         # Counts are derived from _render_report()'s mock corpus (all npm peer
         # deps mocked permissive): of the 10 npm-installed grammars, 9 render
         # Ready and 1 — tree-sitter-cpp — is the intentional pin (#1242), so it is
-        # not counted ready. The 2 blockers are that same pinned tree-sitter-cpp
-        # plus the vendored, ABI-held tree-sitter-c (the only out-of-range
-        # vendored grammar). If a grammar is added/removed or a pin/hold changes,
+        # not counted ready. The 3 blockers are that same pinned tree-sitter-cpp
+        # plus two held vendored grammars: ABI-held tree-sitter-c (#1242/#858) and
+        # tree-sitter-kotlin (pinned to an unreleased fwcd main commit for `fun
+        # interface` support — ABI 14 is in range, but a hold counts as a blocker
+        # until it is lifted). If a grammar is added/removed or a pin/hold changes,
         # update _render_report()'s mock AND these expected counts together; a
         # mismatch here means the report prose drifted, not the regex.
         self.assertEqual(ready.groups(), ("9", "10"))
-        self.assertEqual(blockers.group(1), "2")
+        self.assertEqual(blockers.group(1), "3")
 
     def _matrix_row(self, name: str) -> str:
         for line in self.report.splitlines():
--- a/.github/vendored-grammars.json
+++ b/.github/vendored-grammars.json
@@ -12,7 +12,8 @@
     },
     "kotlin": {
       "name": "tree-sitter-kotlin",
-      "upstream": { "npm": "tree-sitter-kotlin" }
+      "upstream": { "npm": "tree-sitter-kotlin" },
+      "hold": "pinned to unreleased fwcd main commit c8ac3d26 for `fun interface` support (fwcd/tree-sitter-kotlin#169, closes #87) — npm latest (0.3.8) lacks the fix, so the monitor must NOT auto-revert (isNewer is strict-inequality: 0.3.8 != 0.4.0). Drop this hold and bump when upstream cuts a release that includes the fix"
     },
     "dart": {
       "name": "tree-sitter-dart",
--- a/.github/workflows/build-tree-sitter-prebuilds.yml
+++ b/.github/workflows/build-tree-sitter-prebuilds.yml
@@ -29,12 +29,17 @@ name: Build tree-sitter prebuilds
 # incl. macOS + arm64). It is DELIBERATELY NOT wired into normal PR/push CI. It
 # runs only:
 #   1. on manual dispatch (workflow_dispatch); or
-#   2. when a covered grammar's recorded version actually CHANGES — the `guard`
-#      job is the real gate (it diffs the recorded version vs the PR base); the
-#      `paths:` filter below only makes ordinary code PRs cost ZERO matrix time.
-# Net effect: an ordinary code PR triggers nothing; bumping one grammar costs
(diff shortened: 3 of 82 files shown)


magyargergo:
Please double check it with https://github.com/abhigyanpatwari/GitNexus/issues/2194. We need to make sure this fits into the matrix. For now, we only support tree-sitter on 0.21.x and you need to compile it locally on this tree-sitter version and double check if it can generate the grammar properly.

magyargergo (changes requested):
Please check my comment above.

glier:
Thanks @magyargergo — I checked against #2194, and on the ABI/matrix side this already fits, so I want to confirm that and then ask you about the right way to get CI green, since the grammar-bump + prebuild flow is your pipeline.

**ABI / #2194 matrix — fits without a runtime bump:**
- The vendored `src/parser.c` is `#define LANGUAGE_VERSION 14`, i.e. **ABI 14**, inside the supported `tree-sitter@0.21.x` range (13–14). It's also within the 0.25 target range (13–15), so it won't block that upgrade.
- #2194 already tracks this exact commit: *`tree-sitter-kotlin … upstream fwcd/tree-sitter-kotlin@c8ac3d262724 ABI 14 (in target range)`* — the pin here is that same commit.

**Compiled + verified locally on `tree-sitter@0.21.1`** (the pinned runtime):
- Built the committed `parser.c`/`scanner.c` against `tree-sitter@0.21.1`; `Parser.setLanguage()` accepts it (the ABI gate throws on >14 — it doesn't), and `fun interface IntPredicate { fun accept(i: Int): Boolean }` parses with no ERROR.
- `python3 .github/scripts/check-tree-sitter-upgrade-readiness.py --assert-current` → green: `tree-sitter-kotlin: vendored ABI 14 in range [vendored, held]`.
- Full Kotlin suite green (300 unit/cfg/resolver + 233 integration incl. the new cases). We vendor upstream's committed ABI-14 `parser.c` directly rather than regenerating, so no newer tree-sitter CLI is in play.

**The red CI — and a process question for you:**
The only failing checks are the 3 new `fun interface` tests. They fail because the committed `prebuilds/` are still the 0.3.8 binaries, and `node-gyp-build` prefers a committed prebuild over a source build — so CI loads the old grammar until the prebuilds are regenerated from the new source. The version bump triggers `build-tree-sitter-prebuilds`, but on a fork PR its `aggregate` (auto-PR) job is skipped, so it only uploads artifacts and can't commit the binaries here.

I'm not sure of the intended sequencing for a vendored-grammar bump in this repo — could you point me the right way? A couple of options I see:
1. A maintainer dispatches `build-tree-sitter-prebuilds` (`grammars=kotlin`) so the 6 rebuilt binaries + `SHA256SUMS` land (in this branch or a companion PR merged first/together), after which the tests go green; or
2. I pull the 3 `fun interface` tests out into a follow-up PR that merges right after the prebuilds, so this PR is green on its own.

Happy to do whichever fits your workflow — or follow a different process if there's an established one I've missed.

magyargergo:
I saw one ci job is red: https://github.com/abhigyanpatwari/GitNexus/actions/runs/27964323565/job/82753978209#step:4:1

---

I'll also check this locally if i can build it on tree-sitter 0.22.1

glier:
Thanks for the link — didn't spot that failing job at first. Tracked it down: the `hold` I added for kotlin makes the readiness report count it as a blocker (it treats any held vendored grammar that way), so the hardcoded blocker count in the readiness unit test was off by one. Fixed in d4066203 — bumped it 2→3. ABI's still 14, so it builds fine on 0.21.x/0.22.1, the hold's just what flags it.

Only thing left red is the 3 fun-interface tests — still the prebuild question from above.

magyargergo:
We also need to regenerate the fingerprints for kotling benchmarks. https://github.com/abhigyanpatwari/GitNexus/actions/runs/27979931111/job/82807100858?pr=2271

glier:
Already did that in 6105cf7f — but I baselined the kotlin scope-capture fingerprint to the *new* grammar (`4900431791…`), so like the fun-interface tests it only goes green once the regenerated prebuilds land. Right now CI loads the committed 0.3.8 binary, computes the old fingerprint (`b4406b7c…`, 2 fewer capture groups — the missing fun-interface ones) and stays red.

I can flip the baseline to the current old-grammar value so the bench goes green now, but the fun-interface tests need the new binary regardless — so it really all hinges on regenerating the kotlin prebuilds. Want me to handle that somehow, or will you kick `build-tree-sitter-prebuilds`?

magyargergo:
Ok, so the plan is to include the new binary artifacts here.

magyargergo (approved):

magyargergo:
Thanks for this contribution! @glier

Issue #2280, opened 2026-06-23
Spring route extraction: ingestion/group parity gaps blocking Java parse-skip (#2138 follow-up)


henry201605:
## Background

#2265 (#2138 Part 2) added a parse-skip optimisation: the group HTTP scan skips re-parsing route files already covered by the graph. This requires the ingestion-layer route extractor to recognise the same routes the group layer does, otherwise covered-but-unrecognised routes are dropped.

During #2265 review, Java was set to `routeCoverage: 'partial'` (conservative — no skip) because three Spring forms are recognised by the group extractor (`group/extractors/http-patterns/java.ts`) but **not** by the ingestion extractor (`route-extractors/spring.ts`). Per #2078, the two extractors stay separate by design but must not silently drift.

This issue tracks closing those three gaps. Only once all three are closed can Java be flipped back to `'complete'` (re-enabling parse-skip and letting the currently-empty `Route.handlerSymbolId` values resolve for Java).

## Subtasks

- [x] #2287 — Method-level array form (`@GetMapping({"/a", "/b"})`). Done in #2281.
- [x] #2288 — Interface-inherited mappings (group cross-joins via `scanSpringInheritanceProject`; ingestion is per-file with no inheritance pass).
- [x] #2289 — Multi-verb Route identity (`Route` ids are URL-only, so same-URL different-verb routes collapse). Touches `routes.ts`, `call-processor.ts`, `processes.ts` and the cross-repo bridge (#2269); identity convention to be agreed first.

## Notes

- Class-level `@RequestMapping` array prefixes are also unmatched (rare).
- Each gap is independent and can land in its own PR; they share only the goal of restoring `'complete'`.

henry201605:
@magyargergo thanks for merging this! Quick follow-up on the tracking issue: #2281 closed gap (1) — method-level array-form mappings — but #2280 was opened to track three gaps, and (2) and (3) are still open:

- (2) interface-inherited Spring mappings (ingestion is per-file, no inheritance pass)
- (3) multi-verb Route identity (Route ids are URL-only, so same-URL GET+POST collapses into one node)

Both are still needed before Java's `routeCoverage` can flip to `'complete'`. (3) in particular touches the Route id convention across `routes.ts`, `call-processor.ts`, `processes.ts` and the cross-repo bridge, so I'd want to align with you on the identity scheme before starting it.

[...]

Issue #909, opened 2026-04-18
Meta: Pure Scope-Based Resolution in the Semantic Model


magyargergo:
**RFC:** [🧭 RFC: Pure Scope-Based Resolution in the Semantic Model](https://www.notion.so/346dc50b6ed281cfaacbe480bf231d50)
**Predecessor:** [🏗️ RFC: From fuzzy lookups to a semantic model](https://www.notion.so/33adc50b6ed2819eaff5f522f278196d)
**Complements:** [🗺️ Ingestion Roadmap — Phase 3 + Phase 4](https://www.notion.so/344dc50b6ed281619ab5f9527ee789b3)
**Triggering feedback:** #902 (closed; work absorbed into #912)

---

## What this is

Replace today's call-resolution DAG and `TieredCandidates` API with a pure scope-based resolution layer on the SemanticModel. Resolver consults **three scope-aware per-kind entry points** — `ClassRegistry.lookup(name, scope)`, `MethodRegistry.lookup(name, scope)`, `FieldRegistry.lookup(name, scope)` — backed by a scope tree with declarations, imports, and type bindings. Confidence and provenance become first-class per resolution. No AST reads at resolution time. Details in the Notion RFC.

## Rollout — 5 rings, shadow-first

```
Ring 1 (types) → Ring 2 (plumbing: shared + package-local)
             → Ring 3 (per-language migrations, parallel)
             → Ring 4 (DAG retirement, gated on all production langs flipped)
             → Ring 5 (perf + docs)
```

Per-language flip threshold: **≥ 99% fixture parity, ≥ 98% corpus parity** vs legacy DAG under shadow mode. See RFC §6 for the full delivery plan.

### Ring 1 — Types & contracts
- [x] #910 — Scope-resolution types + constants in `gitnexus-shared`
- [x] #911 — LanguageProvider hook interface extensions

### Ring 2 — SHARED (pure logic, to `gitnexus-shared`)
- [x] #912 — `Scope` / `ScopeTree` / `PositionIndex` (generalizes closed #902)
- [x] #913 — `DefIndex` / `ModuleScopeIndex` / `QualifiedNameIndex`
- [x] #914 — `MethodDispatchIndex` build function over `HeritageMap`
- [x] #915 — SCC-aware finalize algorithm (bounded fixpoint)
- [x] #916 — `resolveTypeRef` (strict helper)
[...]

PR #2281, opened 2026-06-23
fix(ingestion/routes): recognise Spring method-level array-form route mappings

### LINKED ISSUE #2287: Spring route extraction: ingestion does not recognise method-level array-form mappings (#2280)
Subtask of #2280. ✅ Done in #2281.

## Gap (resolved)

The ingestion-layer Spring extractor only matched the bare single-string form of method-level mappings, so the array form — `@GetMapping({"/a", "/b"})` / `@PostMapping(path = {"/x", "/y"})` — produced no Route node, while the group layer already recognised it. This was the root of the #2265 array-form gap.

## Resolution (#2281)

- `route-extractors/spring.ts` now matches the array form (positional and `path =`/`value =` named-arg), emitting one Route per path.
- Added a parity test (`spring-route-parity.test.ts`) running one shared fixture through both `extractSpringRoutes` and `JAVA_HTTP_PLUGIN.scan`.
- Method-level array routes nested under a class-level *array-form* `@RequestMapping` are deliberately suppressed (rather than emitted with a dropped prefix); full class-array cross-product support is out of scope here.

Closed as completed.

henry201605:
## What

The ingestion-layer Spring route extractor (`route-extractors/spring.ts`) now recognises the **array form** of method-level mappings, e.g. `@GetMapping({"/a", "/b"})` and `@PostMapping(path = {"/x", "/y"})`, emitting one `Route` per path. Previously only the bare single-string form was matched, so array-form endpoints were silently missing from the graph.

This mirrors the group-layer extractor (`group/extractors/http-patterns/java.ts`), which already handled the array form. The divergence between the two Spring extractors was the root cause of the array-form gap noted during #2265 review.

## Why

Per the maintainer request in #2078, the two Spring extraction paths (ingestion `spring.ts` and group `java.ts`) stay **separate** by design, but must not silently drift on which routes they recognise. When they disagree, the graph under-covers what the group scan sees.

## How

- `spring.ts`: the method-level branches of `ROUTE_ANNOTATION_QUERY` now accept `[(string_literal) (element_value_array_initializer (string_literal))]` for both positional and `path =`/`value =` named-arg forms. A multi-element array yields one tree-sitter match per element, so the existing Phase 2 loop emits one route per path with no special-casing.
- New parity test `test/unit/group/spring-route-parity.test.ts`: runs one shared Java fixture through both `extractSpringRoutes` and `JAVA_HTTP_PLUGIN.scan` and asserts the provider `(method, path)` sets are identical — so the two extractors can't drift apart again.

## Scope

This is the first of the three ingestion/group parity gaps tracked in #2280. `routeCoverage` for Java stays `'partial'` (unchanged) — the flip to `'complete'` waits until all three gaps are closed. Class-level `@RequestMapping` array prefixes remain unmatched (rare).

Part of #2280.

## Testing

- `npm test` (group + route suites green; parity test added)
- `npx tsc --noEmit` clean
- prettier / eslint clean

### DIFF
--- a/gitnexus/test/unit/group/spring-route-parity.test.ts
+++ b/gitnexus/test/unit/group/spring-route-parity.test.ts
@@ -13,8 +13,17 @@
  * otherwise the graph under-covers what the group scan sees, which is exactly
  * the divergence behind the #2265 array-form gap (the group query matched
  * `@GetMapping({"/a","/b"})`, ingestion's didn't). This test runs one shared
- * fixture through both and asserts the provider sets are identical, so the two
- * can't silently drift again.
+ * fixture through both and asserts the provider sets are identical for the
+ * shapes ingestion claims to cover: bare, named-arg, and array-form method
+ * routes under a *scalar* (or absent) class prefix.
+ *
+ * Known, deliberate divergence (NOT covered by the equality assertions): a
+ * method-level array route nested under a class-level *array-form*
+ * @RequestMapping. Ingestion suppresses it (it can't resolve which of several
+ * class prefixes to apply, and a dropped-prefix route is a wrong signal), while
+ * the group layer emits the full cross-product. The last test pins this so the
+ * suppression can't silently regress into emitting wrong routes; full
+ * class-array support is tracked in #2280.
  */
 import { describe, it, expect } from 'vitest';
 import Parser from 'tree-sitter';
@@ -93,5 +102,98 @@ public class PlainController {
 }
 `;
     expect(ingestionProviders(src)).toEqual(groupProviders(src));
+  });
+
+  it('agree on named-arg ARRAY forms (path = {...} / value = {...})', () => {
+    // Guards the spring.ts named-array query branch specifically: positional
+    // arrays alone would not exercise it.
+    const src = `package com.example;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+@RequestMapping("/api")
+public class NamedArrayController {
+  @PutMapping(value = {"/update", "/modify"}) public Object upd() { return null; }
+  @DeleteMapping(path = {"/x", "/y"}) public Object del() { return null; }
+}
+`;
+    const group = groupProviders(src);
+    expect(group).toEqual(
+      new Set(['PUT /api/update', 'PUT /api/modify', 'DELETE /api/x', 'DELETE /api/y']),
+    );
+    expect(ingestionProviders(src)).toEqual(group);
+  });
+
+  it('do not leak non-route arrays (consumes/produces) as routes — array analogue', () => {
+    // The scalar `produces` anti-regression already exists in the route tests;
+    // this is its array form. `consumes`/`produces` arrays must never surface as
+    // provider routes; only the path value does.
+    const src = `package com.example;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+public class ContentTypeController {
+  @GetMapping(value = "/v", consumes = {"application/json", "application/xml"}, produces = {"application/json"})
+  public Object v() { return null; }
+  @PostMapping(consumes = {"application/json"})
+  public Object noPath() { return null; }
+}
+`;
+    const ingestion = ingestionProviders(src);
+    const group = groupProviders(src);
+    // Only the explicit path leaks through; the consumes/produces arrays do not,
+    // and the path-less @PostMapping contributes nothing.
+    expect(group).toEqual(new Set(['GET /v']));
+    expect(ingestion).toEqual(group);
+
+    // Pure-consumes controller (no path anywhere) → EMPTY provider set on both
+    // sides: a consumes/produces array must never be misread as a route path.
+    const consumesOnly = `package com.example;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+public class ConsumesOnlyController {
+  @PostMapping(consumes = {"application/json", "application/xml"})
+  public Object a() { return null; }
+  @PutMapping(produces = {"application/json"})
+  public Object b() { return null; }
+}
+`;
+    expect(groupProviders(consumesOnly)).toEqual(new Set());
+    expect(ingestionProviders(consumesOnly)).toEqual(new Set());
+  });
+
+  it('pins the deliberate class-array divergence: ingestion suppresses, group emits cross-product (#2280)', () => {
+    // A method-level array under a class-level ARRAY-form @RequestMapping. There
+    // is no single class prefix to apply, so ingestion suppresses the route
+    // rather than emit it unprefixed (a wrong signal). The group layer emits the
+    // full cross-product. This is a KNOWN gap (#2280), pinned here so the
+    // suppression can't silently regress into emitting wrong unprefixed routes.
+    const src = `package com.example;
+import org.springframework.web.bind.annotation.*;
+
+@RestController
+@RequestMapping({"/base/one", "/base/two"})
+public class MultiPrefixController {
+  @GetMapping({"/primary", "/alias"}) public Object x() { return null; }
+}
+`;
+    const ingestion = ingestionProviders(src);
+    const group = groupProviders(src);
+
+    // group: correct cross-product of the two class prefixes × two method paths.
+    expect(group).toEqual(
+      new Set([
+        'GET /base/one/primary',
+        'GET /base/two/primary',
+        'GET /base/one/alias',
+        'GET /base/two/alias',
+      ]),
+    );
+    // ingestion: suppressed — emits NO route for the array method (never a wrong
+    // unprefixed `GET /primary` / `GET /alias`).
+    expect(ingestion).toEqual(new Set());
+    // And the divergence is asymmetric-by-design: ingestion ⊊ group here.
+    expect(ingestion).not.toEqual(group);
   });
 });
--- a/gitnexus/test/unit/group/spring-route-parity.test.ts
+++ b/gitnexus/test/unit/group/spring-route-parity.test.ts
@@ -1,0 +1,97 @@
+/**
+ * Parity test for the two Spring route extractors (#2078 maintainer request,
+ * #2138 follow-up).
+ *
+ * GitNexus parses Spring `@(Get|Post|...)Mapping` annotations in TWO places:
+ *   - ingestion `route-extractors/spring.ts` → `extractSpringRoutes` (produces
+ *     graph `Route` nodes)
+ *   - group `http-patterns/java.ts` → `JAVA_HTTP_PLUGIN.scan` (produces
+ *     cross-repo HTTP contracts)
+ *
+ * They serve different layers and stay separate, but they MUST agree on the
+ * set of provider (method, path) routes they recognise for the same source —
+ * otherwise the graph under-covers what the group scan sees, which is exactly
+ * the divergence behind the #2265 array-form gap (the group query matched
+ * `@GetMapping({"/a","/b"})`, ingestion's didn't). This test runs one shared
+ * fixture through both and asserts the provider sets are identical, so the two
+ * can't silently drift again.
+ */
+import { describe, it, expect } from 'vitest';
+import Parser from 'tree-sitter';
+import Java from 'tree-sitter-java';
+import { extractSpringRoutes } from '../../../src/core/ingestion/route-extractors/spring.js';
+import { JAVA_HTTP_PLUGIN } from '../../../src/core/group/extractors/http-patterns/java.js';
+
+function parse(src: string): Parser.Tree {
+  const p = new Parser();
+  p.setLanguage(Java);
+  return p.parse(src);
+}
+
+/** Canonical `METHOD /a/b` form so prefix-join / slash / case differences wash out. */
+function canon(method: string, ...segments: string[]): string {
+  const path = `/${segments.join('/').split('/').filter(Boolean).join('/')}`;
+  return `${method.toUpperCase()} ${path.toLowerCase()}`;
+}
+
+/** ingestion side: join the class prefix + method path the way the routes phase does. */
+function ingestionProviders(src: string): Set<string> {
+  return new Set(
+    extractSpringRoutes(parse(src), 'X.java').map((r) =>
+      canon(r.httpMethod, r.prefix ?? '', r.routePath),
+    ),
+  );
+}
+
+/** group side: provider detections (path already prefix-joined by the plugin). */
+function groupProviders(src: string): Set<string> {
+  return new Set(
+    JAVA_HTTP_PLUGIN.scan(parse(src))
+      .filter((d) => d.role === 'provider')
+      .map((d) => canon(d.method, d.path)),
+  );
+}
+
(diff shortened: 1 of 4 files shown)


gitnexus/test/unit/group/spring-route-parity.test.ts
```diff
@@ -0,0 +1,97 @@
...
+ *   - group `http-patterns/java.ts` → `JAVA_HTTP_PLUGIN.scan` (produces
+ *     cross-repo HTTP contracts)
+ *
+ * They serve different layers and stay separate, but they MUST agree on the
+ * set of provider (method, path) routes they recognise for the same source —
+ * otherwise the graph under-covers what the group scan sees, which is exactly
+ * the divergence behind the #2265 array-form gap (the group query matched
+ * `@GetMapping({"/a","/b"})`, ingestion's didn't). This test runs one shared
+ * fixture through both and asserts the provider sets are identical, so the two
+ * can't silently drift again.
```
magyargergo: **P2 [reproduced]** — Class-array-prefix wrong-route (real location: spring.ts:56-68, unchanged class branches + prefixByClassId: Map<number,string>). The widened METHOD branches now match `@GetMapping({...})`, but the class branches still match only a single string_literal. Trigger reproduced on PR head: `@RequestMapping({"/base/one","/base/two"}) class` + `@GetMapping({"/primary","/alias"})` → ingestion=[GET /primary, GET /alias] (WRONG — prefix dropped) vs group=[GET /base/one/primary, GET /base/two/primary, GET /base/one/alias, GET /base/two/alias]. On base this method emitted nothing, so the PR turns a missing route into a wrong unprefixed Route node (GET /primary) in place of the real GET /base/one/primary; the group HTTP cross-repo matcher then derives a provider contract from that wrong path and can mis-match a consumer or miss the real endpoint (the server serves /base/one/primary, NOT /primary) — a false signal, not merely a missing one. Severity P2 not blocker: class-level array-form @RequestMapping is a rare idiom, AND the same wrong-unprefixed-route already happens for scalar method paths under a class array prefix pre-PR — this PR extends a pre-existing latent bug rather than introducing a new class. Fix options: (1) minimal-safe — suppress method-array routes when the enclosing class has an array-form @RequestMapping; (2) complete — promote prefixByClassId to Map<number,string[]> and mirror the array alternation onto the class branches (group already does this via prefixByTypeId); (3) at minimum — narrow this test's 'can't silently drift again' claim and add a documented it.todo/xfail class-array fixture. Maintainer's call.

gitnexus/test/unit/group/spring-route-parity.test.ts
```diff
@@ -0,0 +1,97 @@
...
+function groupProviders(src: string): Set<string> {
+  return new Set(
+    JAVA_HTTP_PLUGIN.scan(parse(src))
+      .filter((d) => d.role === 'provider')
+      .map((d) => canon(d.method, d.path)),
+  );
+}
+
+describe('Spring route extractor parity — ingestion spring.ts vs group java.ts', () => {
+  it('agree on bare, named-arg, and array-form method routes under a class prefix', () => {
```
magyargergo, line 65: **P2 [code-read]** — Named-arg ARRAY form is untested. The new branch at spring.ts:83-84 accepts `@PutMapping(value={...})` / `path={...}`, but this fixture only covers positional arrays plus SCALAR named args (path="/make", value="/update"). A regression in the named-array branch would pass the suite. (Coordinator probe confirms the behavior currently WORKS — `@PostMapping(value={"/a","/b"})` → POST /a, POST /b — so this is an unguarded-coverage gap, not a live bug.) Suggested: add `@PutMapping(value={"/update","/modify"})` + `@DeleteMapping(path={"/x","/y"})` parity cases asserting both extractors emit all paths, plus a `consumes={...}` case asserting an empty provider set (the array analogue of the existing scalar produces anti-regression). Cross-engine: Codex B2 #1 (conf 100) + Claude testing.

magyargergo (commented):
## Tri-review digest — PR #2281 `fix(ingestion/routes): recognise Spring method-level array-form route mappings`

**Verdict: APPROVE WITH COMMENTS** — net-positive fix (closes the #2265 method-level array-form gap; correct, false-positive-safe, mirrors group `java.ts`). One P2 edge-case (class-array-prefix wrong-route) is worth addressing but is *not* a hard blocker — maintainer's call. Two smaller P2/P3 hygiene items (named-array test gap, stale comments).

---

### Engine breakdown
- **Engine A — Claude.** swarm: `risk-architect`, `test-ci-verifier`; ce personas: `correctness`, `testing`, `maintainability`, `project-standards`, `adversarial`. All seven lanes LIVE.
- **Engine B — Codex (gpt-5.5 / xhigh).** All three legs **LIVE**: `swarm` (B1) — `production-ready with minor follow-ups`, incl. a GitNexus impact pass (1 changed symbol `ROUTE_ANNOTATION_QUERY`, 0 upstream dependents, `extractSpringRoutes` 1 test caller, **LOW** risk); `ce` (B2) — `Ready with fixes`; `adversarial` (B3) — `needs-attention` / "no-ship".

Cross-engine corroboration is strong on every postable finding: each was reached by at least one Codex lane **and** independently by a Claude lane, and the coordinator reproduced the headline on the PR head. **Note the engine split on the headline's severity** (below): only the *adversarial* lanes (Codex B3 + Claude adversarial) escalated the class-array case to a blocker; the Codex swarm (B1), Codex ce (B2), and the Claude risk/correctness lanes all classified it as **pre-existing and documented / out-of-scope** — which is exactly why it lands at P2-not-blocker.

---

### Credit — what's solid (validated, lead with this)
The core change is **correct and safe**; multiple lanes mechanically verified it (not just code-read):

- **Widening is correct.** The method-level query now matches `@GetMapping({"/a","/b"})` (positional + named `path=`/`value=`). Tree-sitter yields **one match per array element** (each carrying exactly one `@value`), so the existing Phase-2 last-wins `caps[name]=node` loop emits one route per path with no special-casing — it can *never* collapse a multi-element array. `[reproduced]` Correctness probed raw `Query.matches()`: 3-element array → 3 matches, 1 value-capture each; no path-dropping (correctness + adversarial + Codex B2/B3).
- **False-positive-SAFE.** `consumes`/`produces`/`headers`/`params`/`name` arrays never leak as routes — positional first-arg is always a path; named non-route arrays are rejected by `isRouteMemberKey` (spring-shared.ts:43-46). `[reproduced]` `@GetMapping(value="/v", consumes={...}, produces={...})` → only `GET /v`; `consumes`-only → `[]` (correctness + adversarial + risk + Codex).
- **Graceful & robust.** Empty `{}` → no route/no crash; trailing comma, mixed string/non-string elements, nested arrays, concat exprs, unicode paths, 50-element arrays — all behave and match group output (adversarial probes).
- **Mirrors group `java.ts` method branches exactly** (java.ts:130/143/156/165) — genuine parity for the covered shapes; PR #2254 already proved this semantic for the group layer.
- **Project-standards CLEAN** (project-standards lane, all rules quotable + verified): no `CHANGELOG` edit, no `any`/`as any`, no `if` in the test, RFC #909 satisfied (route logic in `route-extractors/spring.ts`, no language-named branches in shared pipeline), NodeNext `.js` import specifiers, conventional-commit title, portable.
- **Test runs REAL prod code**, not a worker no-op. `[reproduced]` test-ci-verifier ran it locally: **2/2 pass, 2.1s**, in-process tree-sitter, no `distWorkerExists()` gate — the #2078 unit-tier no-op gotcha does NOT apply. Strong `toEqual` set-equality with a hardcoded 5-element expected set guards against both-empty agreement.

---

### Headline finding — P2: class-array prefix produces wrong **unprefixed** routes
**`[reproduced]`** · escalated by **Codex B3 (adversarial)** + Claude adversarial; the divergence itself was also noted (as pre-existing/documented) by Claude correctness, testing, maintainability, risk and Codex B1/B2.

**Trigger:** a controller with a class-level *array-form* `@RequestMapping` plus a method-level array mapping. The PR widened the **method** query branches but left the **class** branches matching only a single `(string_literal)` (spring.ts:56-68, unchanged; `prefixByClassId: Map<number,string>`).

**Observed (coordinator probe on PR head):**
```java
@RequestMapping({"/base/one","/base/two"})
public class A { @GetMapping({"/primary","/alias"}) public Object x(){return null;} }
```
- **ingestion** → `["GET /primary","GET /alias"]`  ← WRONG: prefix dropped
- **group**     → `["GET /base/one/primary","GET /base/two/primary","GET /base/one/alias","GET /base/two/alias"]`  ← correct cross-product

**Why it matters (and why the risk lane's framing was wrong):** On BASE the method-array branch did not exist, so this method emitted *nothing*. The PR converts "missing route" into "wrong route." The Claude risk lane initially argued "no active harm — a real subpath the server serves directly"; **that is incorrect** and was refuted by the coordinator + Codex B3: the server serves `/base/one/primary`, NOT `/primary`. The unprefixed `/primary` is a **wrong `Route` node** emitted into the graph in place of the real `/base/one/primary`; the group HTTP cross-repo matcher then derives a provider contract from that wrong path, so it can mis-match a consumer or miss the real endpoint — a *false* signal, not merely a missing one. (The wrong `Route` node is directly reproduced; the cross-repo-match consequence is the reasoned downstream effect.) Codex B3 rated it `[medium]` / "no-ship."

**Honest severity calibration → P2, not a hard blocker:**
- (a) class-level *array-form* `@RequestMapping` is a **rare** idiom;
- (b) the same wrong-unprefixed-route already occurs for **scalar** method paths under a class array prefix **pre-PR** (`@GetMapping("/x")` → `GET /x`), so this PR **extends a pre-existing latent bug**, it does not introduce a new bug class;
- (c) the common case — method array under a scalar / no class prefix — is correct and a clear net improvement.

**Suggested fixes (offer all; maintainer's call):**
1. *Minimal-safe (Codex B3):* suppress method-array routes when the enclosing class has an array-form `@RequestMapping` — never emit wrong data.
2. *Complete:* add class-level array support + cross-product — change `prefixByClassId` to `Map<number,string[]>` and mirror the array alternation onto the class branches (the group side already does this via `prefixByTypeId: Map<number,string[]>`).
3. *At minimum:* narrow the test's "can't silently drift again" claim and add a documented `it.todo`/xfail fixture pinning the class-array divergence so it's machine-visible.

Inline anchor: new test file / the in-diff method branches. The class branches (real location spring.ts:56-68) are unchanged → not directly inline-able; referenced in this body.

---

### Other findings

**P2 — named-arg ARRAY form is untested.** `[code-read]` · cross-engine: **Codex B2 #1 (conf 100)** + Claude testing.
The new branch at spring.ts:83-84 accepts `@PutMapping(value={"/a","/b"})` / `path={"/a","/b"}`, but the parity fixture only covers positional arrays + SCALAR named args. A regression in the named-array branch would pass the suite. NOTE: coordinator probe confirms the behavior currently **works** (`@PostMapping(value={"/a","/b"})` → `POST /a`,`POST /b`) — this is an **unguarded-coverage gap, not a live bug**.
*Suggested:* add `@PutMapping(value={"/update","/modify"})` + `@DeleteMapping(path={"/x","/y"})` parity cases; add a `consumes={...}` false-positive case asserting an empty provider set (the array analogue of the existing scalar `produces` anti-regression).
Inline anchor: new test file (in-diff).

**P3 — stale comments now mislead.** `[code-read]` · cross-engine: **Codex B2 #2 + Codex B3** + Claude risk.
`java.ts:681` `routeCoverage` comment still lists "(1) array-form `@GetMapping({...})`" as a reason ingestion under-covers — now stale for method-level. The same stale claim sits in `test/integration/route-parse-skip.test.ts:176`. **Both files are UNCHANGED by the diff → cannot inline → body note only.** Keep `routeCoverage:'partial'` (the interface-inheritance + same-URL multi-verb reasons still hold); just drop the method-level array item.

---

### Lower-priority + refuted + test-gap notes (body only)
- **"can't silently drift again" overstated** (all Claude lanes + Codex B2/B3): the parity harness covers only method-level forms under a string class prefix. Narrow the claim or add xfail fixtures.
- **Query-fragment duplication** across `spring.ts` and `java.ts` with no shared constant (Claude maintainability flagged P2 → treat as **P3 advisory**): a future edit to one query arm must be mirrored by hand. Optional: export a shared `SPRING_METHOD_ROUTE_VALUE_PATTERN` from `spring-shared.ts`.
- **`canon()` `.toLowerCase()`** masks case-sensitivity divergence (Claude testing + Codex B2): Spring paths are case-sensitive; lowercasing both sides could hide a real divergence. Fixtures are all lowercase so no current test is wrong. P3.
- **Empty-string array element** `{""}` → `""` (ingestion) vs `"/"` (group), masked by `canon()` (Claude adversarial). Rare/degenerate idiom; P3.
- **Pre-existing orthogonal divergences the test also fails to catch** — interface-method routes (ingestion over-emits + drops the interface prefix; `findEnclosingClass` only matches `class_declaration`) and FQN/`scoped_identifier` annotation names (ingestion under-emits vs group). These are **pre-existing, not regressions, on untouched code** — cited only as supporting evidence that the parity claim is broader than the actual coverage, **not** as separate diff findings.

**Refuted suspicions (credit):** false-positive route explosion (REFUTED — guarded by `isRouteMemberKey`); path-dropping via last-wins collapse (REFUTED — one match per element); class-level annotation misfire (REFUTED — new branches are `method_declaration`-only); malformed-query module-load failure (REFUTED — CI green); crashes on malformed/edge inputs (REFUTED — all graceful).

---

### CI / merge status
- **All green except Vercel** ("Authorization required to deploy") — known deploy-auth, **not a code gate**. CodeQL, `quality/format` (root prettier — new test file is root-clean), lint, typecheck, multi-platform tests all PASS.
- **`mergeStateStatus = BEHIND` main** (mergeable = MERGEABLE). Minor hygiene: **update-from-main before merge** if branch protection requires a current base.

---

*Automated multi-tool digest — Claude (swarm + ce personas) and Codex gpt-5.5/xhigh (swarm + ce + adversarial), all legs live. Verify findings against the diff before acting.*

henry201605:
Thanks for the thorough tri-review — the cross-engine breakdown was really helpful. Addressed the three actionable items:

- **P2 class-array wrong-route** → went with option 1 (suppress) + option 3 (pin). The class branches now match the array form only to *detect* it; a method-level array route under a class-level array `@RequestMapping` is suppressed rather than emitted with a dropped prefix, so ingestion stays a strict subset of the group scan (never a wrong unprefixed Route). A new parity case pins this: group emits the cross-product, ingestion emits ∅. Scalar method paths under an array class prefix are unchanged (pre-existing, out of scope). Full class-array cross-product support is left to #2280.
- **P2 named-array coverage** → added `value={...}` / `path={...}` parity cases, a `consumes`/`produces` array false-positive case, and a dedicated empty-provider-set assertion for a path-less consumes/produces controller.
- **P3 stale comments** → updated the `routeCoverage` comment in `java.ts` and the `route-parse-skip.test.ts` note; also narrowed the parity test's "can't drift" claim.

`routeCoverage` stays `'partial'`. I left the P3 advisories (shared query constant, `canon()` lowercasing, empty-string element) as-is for now — happy to fold any of them in if you'd prefer. Rebased onto the latest branch head; tsc / lint / prettier clean, group + route suites green.

PR #2282, opened 2026-06-23
feat(group): resolve inline HTTP provider handlers via call-site line (#2276)

### LINKED ISSUE #2276: HTTP provider symbolUid: wire call-site line for inline handlers in Python/Go/PHP/Kotlin/Java
## Limitation

#2269 wires the call-site `line` through all six HTTP plugins for **consumer** detections (and Node/TS providers, including express inline arrows). Named providers in every language resolve by handler name (or #2265's `handlerSymbolId`). But **inline/closure provider handlers in non-Node languages** do not set `line`, so they cannot resolve via containment and fall back to file-level. Common cases:

```php
Route::get('/x', function () { ... });        // Laravel closure
```
```go
http.HandleFunc("/x", func(w http.ResponseWriter, r *http.Request) { ... })  // Go anonymous handler
```

## Suggested direction

Set `line: <node>.startPosition.row + 1` on the **provider** detection emit sites in the python / go / php / kotlin / java plugins (mirroring the consumer wiring already in place), so an inline provider resolves to its containing/arrow symbol via `resolveContainingSymbol`. Mechanical, plugin-local change.

## Pointers

- `gitnexus/src/core/group/extractors/http-patterns/{python,go,php,kotlin,java}.ts` — provider `out.push({...})` sites
- `gitnexus/src/core/group/extractors/http-route-extractor.ts` — `resolveContainingSymbol` (consumes `line`)

Good first issue — follow-up from #2269.

magyargergo:
## What & why

Closes #2276 (follow-up to #2269). #2269 wired the call-site `line` through every HTTP plugin's **consumer** detections so a consumer contract resolves to the function containing the `fetch`/client call. Provider detections were left behind: an **inline / closure provider handler** in the non-Node plugins (Go `http.HandleFunc("/x", func(){…})`, Laravel `Route::get('/x', function(){…})`) carried no `line`, so it could not resolve by line-span containment and degraded to a file-level `symbolUid` — leaving cross-repo trace/impact unable to attribute the route to a real symbol.

This PR wires `line` onto the provider emit sites across the five non-Node plugins so inline providers resolve to their containing / closure symbol, matching the consumer behavior and the Node/Express provider behavior already in place.

## The load-bearing detail

The resolver (`resolveDetectionSymbol` in `http-route-extractor.ts`) short-circuits: a provider with a **truthy `name`** is resolved **by name only** and never falls through to line-span containment (the guard that prevents a route attaching to its registrar). `line` is consulted only for consumers and for providers whose `name` is falsy. So "just set `line`" is necessary but **not sufficient** for the headline cases — inline handlers must emit **both** `name: null` **and** `line`.

## Per-language changes

| Plugin | Change |
|--------|--------|
| **Go** (`go.ts`) | Widened the HandleFunc + framework-route handler capture from `(identifier)` to `[(identifier) (func_literal)]`; an inline func literal now emits `name: null` + `line: (handlerNode ?? pathNode).startPosition.row + 1`. Named identifier handlers keep resolving by name. |
| **PHP** (`php.ts`) | The Laravel route query optionally captures the closure handler `(argument [(anonymous_function) (arrow_function)] @closure)?`; a closure emits `name: null` + line and resolves to its containing symbol (service-provider `boot()`, controller method). Named-controller routes are unchanged. |
| **Python** (`python.ts`) | FastAPI `@app`/`@router` providers (already `name: null`) gain `line`. Best-effort: FastAPI routes are graph-authoritative and the function span starts at `def`, so this lands the single-decorator source-scan fallback. Flask `add_url_rule` already carried `line`. |
| **Kotlin / Java** (`kotlin.ts`, `java.ts`) | Spring `@*Mapping` named-method providers gain `line` for parity with the consumer emits and a future inline DSL. Inert for current resolution (verified): a named Spring method resolves by name and never reaches containment. |

## Testing

`gitnexus/test/unit/group/http-route-extractor.test.ts` — 9 new tests via the existing mock-`dbExecutor` containment harness:
- Go inline `http.HandleFunc` + gin closure resolve to the containing function; named handler still resolves by name.
- Laravel closure (`function(){}` and `fn() =>`) nested in a method resolves to that method; named-controller route unchanged.
- FastAPI `@app` decorated handler resolves to the decorated function.
- Documented-limitation guards: a **file-scope** Laravel closure and a **multi-decorator** FastAPI handler both degrade to file-level (no mis-attribution) rather than resolving.

Full file: 182 tests pass. `tsc` clean. Broader group/route suites (consumer signals, multi-verb, graph-method, cross-trace, Spring parity): green.

## Scope boundaries / documented limitations

- **File-scope Laravel closures** (the common `routes/web.php` pattern) have no enclosing function, and PHP closures are not indexed as symbols, so they remain file-level. Method-nested closures resolve. Closure indexing is a separate ingestion concern.
- **Multi-decorator FastAPI** handlers degrade to file-level (the detection line sits above the `def`-span). FastAPI routes are graph-authoritative regardless.
- Named-controller Laravel routes (`[Controller::class, 'method']`) keep the `'route'` label — resolving them to a real method is graph-backed and out of scope here.

## Residual Review Findings

From the multi-engine `ce-code-review` pass (7 reviewers, 0 P0/P1):

- **[P3 · pre-existing · go.ts:34]** Go multi-argument routes with middleware before the handler (`r.GET("/x", mw, func(){})`) emit two provider detections — one for the middleware identifier and one for the inline closure — because the handler capture is unanchored. This is a pre-existing over-match (it already affected `r.GET("/x", mw, namedHandler)`); the func-literal widening compounds the inline case. **Deferred** per the plan's scope boundary (out of scope; pre-existing). Fix path: anchor `@handler` to the trailing argument.
- **[P3 · advisory · python.ts]** A FastAPI decorator nested inside an enclosing function resolves to the enclosing function rather than the inner handler. This is strictly better than file-level, matches the issue's "containing symbol" intent, and FastAPI is graph-authoritative anyway.
- **[P2 · advisory · http-route-extractor.test.ts]** The test file is ~7.2k lines (pre-existing); the inline-provider tests are a future extraction candidate.

### DIFF
--- a/gitnexus/src/core/group/extractors/http-patterns/go.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/go.ts
@@ -31,7 +31,7 @@ const FRAMEWORK_ROUTE_PATTERNS = compilePatterns({
             field: (field_identifier) @http_method (#match? @http_method "^(GET|POST|PUT|DELETE|PATCH)$"))
           arguments: (argument_list
             (interpreted_string_literal) @path
-            (identifier) @handler))
+            [(identifier) (func_literal)] @handler))
       `,
     },
   ],
@@ -51,7 +51,7 @@ const HANDLE_FUNC_PATTERNS = compilePatterns({
             field: (field_identifier) @fn (#eq? @fn "HandleFunc"))
           arguments: (argument_list
             (interpreted_string_literal) @path
-            (identifier) @handler))
+            [(identifier) (func_literal)] @handler))
       `,
     },
   ],
@@ -138,12 +138,18 @@ export const GO_HTTP_PLUGIN: HttpLanguagePlugin = {
       if (!methodNode || !pathNode) continue;
       const path = unquoteLiteral(pathNode.text);
       if (path === null) continue;
+      // An inline `func(){…}` handler has no name → emit `name: null` and a
+      // `line` so it resolves to its containing/closure symbol by line-span
+      // containment (like a consumer). A named identifier handler keeps its
+      // name and resolves by name; `line` is harmless there.
+      const isInlineHandler = handlerNode?.type === 'func_literal';
       out.push({
         role: 'provider',
         framework: 'go-framework',
         method: methodNode.text.toUpperCase(),
         path,
-        name: handlerNode?.text ?? null,
+        name: isInlineHandler ? null : (handlerNode?.text ?? null),
+        line: (handlerNode ?? pathNode).startPosition.row + 1,
         confidence: 0.8,
       });
     }
@@ -155,12 +161,16 @@ export const GO_HTTP_PLUGIN: HttpLanguagePlugin = {
       if (!pathNode) continue;
       const path = unquoteLiteral(pathNode.text);
       if (path === null) continue;
+      // Inline `func(){…}` handler → resolve by containment (see go-framework
+      // note above); a named handler resolves by name.
+      const isInlineHandler = handlerNode?.type === 'func_literal';
       out.push({
         role: 'provider',
         framework: 'go-stdlib',
         method: 'GET',
         path,
-        name: handlerNode?.text ?? null,
+        name: isInlineHandler ? null : (handlerNode?.text ?? null),
+        line: (handlerNode ?? pathNode).startPosition.row + 1,
         confidence: 0.8,
       });
     }
--- a/gitnexus/src/core/group/extractors/http-patterns/go.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/go.ts
@@ -17,8 +17,11 @@ import type { HttpDetection, HttpLanguagePlugin } from './types.js';
 
 // ─── Provider: framework routing ──────────────────────────────────────
 // Matches `\w+\.GET(...)` etc. (gin, echo, chi all share this shape).
-// Captures the HTTP method (field name), path literal, and handler
-// identifier passed as the second argument.
+// Captures the HTTP method (field name), path literal, and the handler —
+// anchored to the LAST argument (`@handler .`) so a variadic middleware
+// chain (`r.GET("/x", mw, handler)`, gin/echo/chi style) binds the real
+// handler, not a middleware identifier (which would otherwise over-match
+// and attach the route to the wrong symbol — see #2276 review).
 const FRAMEWORK_ROUTE_PATTERNS = compilePatterns({
   name: 'go-framework-route',
   language: Go,
@@ -31,7 +34,8 @@ const FRAMEWORK_ROUTE_PATTERNS = compilePatterns({
             field: (field_identifier) @http_method (#match? @http_method "^(GET|POST|PUT|DELETE|PATCH)$"))
           arguments: (argument_list
             (interpreted_string_literal) @path
-            [(identifier) (func_literal)] @handler))
+            [(identifier) (func_literal)] @handler
+            .))
       `,
     },
   ],
@@ -51,7 +55,8 @@ const HANDLE_FUNC_PATTERNS = compilePatterns({
             field: (field_identifier) @fn (#eq? @fn "HandleFunc"))
           arguments: (argument_list
             (interpreted_string_literal) @path
-            [(identifier) (func_literal)] @handler))
+            [(identifier) (func_literal)] @handler
+            .))
       `,
     },
   ],
--- a/gitnexus/src/core/group/extractors/http-patterns/php.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/php.ts
@@ -37,7 +37,9 @@ const LARAVEL_ROUTE_SPEC: PatternSpec<Record<string, never>> = {
     (scoped_call_expression
       scope: (name) @scope (#eq? @scope "Route")
       name: (name) @method (#match? @method "^(get|post|put|delete|patch)$")
-      arguments: (arguments . (argument (string) @path)))
+      arguments: (arguments
+        . (argument (string) @path)
+        (argument [(anonymous_function) (arrow_function)] @closure)?))
   `,
 };
 
@@ -150,12 +152,22 @@ export const PHP_HTTP_PLUGIN: HttpLanguagePlugin = {
       if (!methodNode || !pathNode) continue;
       const path = phpStringText(pathNode);
       if (path === null) continue;
+      // A closure handler (`Route::get('/x', function(){…})` / `fn() => …`) has
+      // no name → emit `name: null` + the registration line so it resolves to
+      // its containing symbol (e.g. a service-provider `boot()` or controller
+      // method) by line-span containment. A named-controller route keeps the
+      // `'route'` label — resolving its array/string handler to a real method is
+      // a separate, graph-backed concern. NOTE: a closure at FILE scope
+      // (routes/web.php) has no enclosing function and PHP closures are not yet
+      // indexed as symbols, so it still degrades to file-level (see #2276).
+      const closureNode = match.captures.closure;
       out.push({
         role: 'provider',
         framework: 'laravel',
         method: methodNode.text.toUpperCase(),
         path,
-        name: 'route',
+        name: closureNode ? null : 'route',
+        line: (closureNode ?? pathNode).startPosition.row + 1,
         confidence: 0.8,
       });
     }
--- a/gitnexus/src/core/group/extractors/http-patterns/python.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/python.ts
@@ -1057,6 +1057,12 @@ export const PYTHON_HTTP_PLUGIN: HttpLanguagePlugin = {
         method: httpMethod,
         path,
         name: null,
+        // The decorated handler has no captured name → resolve by line-span
+        // containment. Best-effort fallback: FastAPI routes are graph-backed
+        // (ingestion decorator routes) and the function span starts at `def`
+        // (decorators excluded), so this lands the single-decorator case and
+        // degrades to file-level for multi-decorator stacks.
+        line: pathNode.startPosition.row + 1,
         confidence: 0.8,
       });
     }
@@ -1101,6 +1107,8 @@ export const PYTHON_HTTP_PLUGIN: HttpLanguagePlugin = {
           method: httpMethod,
           path: p,
           name: null,
+          // Best-effort containment fallback — see the @app provider note above.
+          line: pathNode.startPosition.row + 1,
           confidence: 0.8,
         });
       }
--- a/gitnexus/src/core/group/cross-trace.ts
+++ b/gitnexus/src/core/group/cross-trace.ts
@@ -1036,12 +1036,13 @@ const FILE_BASENAME_RE =
 
 /**
  * A provider endpoint's display label. A resolved handler has a real function
- * name; the source-scan fallbacks leave a generic token (`'handler'`/`'fetch'`)
- * or a file basename. Those are treated as anonymous and shown as
+ * name; the source-scan fallbacks leave a generic token (`'handler'`/`'fetch'`,
+ * or `'route'` for an unresolved named-controller / closure Laravel route) or a
+ * file basename. Those are treated as anonymous and shown as
  * `<METHOD /path handler>` so the endpoint is still identifiable by route. When
  * the bridge row carries a resolved `providerUid`, the name IS a real symbol —
- * the `'handler'`/`'fetch'` sentinel check is suppressed so a function genuinely
- * named `handler` is not mislabeled anonymous.
+ * the sentinel check is suppressed so a function genuinely named `handler` (or,
+ * hypothetically, `route`) is not mislabeled anonymous.
  */
 function providerLabel(
   providerName: string,
@@ -1052,7 +1053,8 @@ function providerLabel(
   const generic =
     providerName === '' ||
     FILE_BASENAME_RE.test(providerName) ||
-    (!resolved && (providerName === 'handler' || providerName === 'fetch'));
+    (!resolved &&
+      (providerName === 'handler' || providerName === 'fetch' || providerName === 'route'));
   return generic
     ? { label: `<${contractId} handler>`, anon: true }
     : { label: providerName, anon: false };
--- a/gitnexus/src/core/group/extractors/http-patterns/java.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/java.ts
@@ -746,6 +746,11 @@ export const JAVA_HTTP_PLUGIN: HttpLanguagePlugin = {
           method: route.httpMethod,
           path: joinPath(prefix, route.rawPath),
           name: route.methodName,
+          // Spring providers are named controller methods resolved BY NAME;
+          // `line` is wired for parity with the consumer emits (and a future
+          // inline DSL) and is inert for current resolution — a named provider
+          // never falls through to line-span containment.
+          line: route.methodNode.startPosition.row + 1,
           confidence: 0.8,
         });
       }
--- a/gitnexus/src/core/group/extractors/http-patterns/java.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/java.ts
@@ -746,11 +746,13 @@ export const JAVA_HTTP_PLUGIN: HttpLanguagePlugin = {
           method: route.httpMethod,
           path: joinPath(prefix, route.rawPath),
           name: route.methodName,
-          // Spring providers are named controller methods resolved BY NAME;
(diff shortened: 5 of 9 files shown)


koriyoshi2041:
Pulled latest head `cd969d98` and did a focused pass on the inline-provider line wiring.

Local checks from `gitnexus/`:

```text
npm test -- --run test/unit/group/http-route-extractor.test.ts
# 1 file passed; 182 tests passed

npm test -- --run test/unit/group/http-route-extractor.test.ts test/unit/group/http-consumer-signals.test.ts test/unit/group/http-route-graph-method.test.ts test/unit/group/http-route-multi-verb.test.ts test/unit/group/cross-trace.test.ts test/unit/calltool-dispatch.test.ts
# 6 files passed; 414 tests passed

npx tsc --noEmit
# passed after rebuilding the local gitnexus-shared package
```

I also checked the main attribution boundaries: named Go handlers still resolve by name instead of the registrar, Laravel named-controller routes stay at the previous file-level behavior, and the documented file-scope Laravel / multi-decorator FastAPI cases degrade instead of attaching to the wrong symbol. I do not see a blocker in this slice.

magyargergo:
Thanks for the focused pass, @koriyoshi2041 — appreciate you independently reproducing the boundaries. Each one you verified is now pinned by an automated test so it can't silently regress:

- **Named Go handler → resolves by name, not the registrar** — `keeps Go NAMED HandleFunc handler resolving by name, not containment` covered the `HandleFunc` path. Since the widening also touched the **framework-route** query, I just added `keeps a NAMED gin framework-route handler resolving by name, not its registrar` (commit `434e8722`): the mock indexes both `listOrders` and an enclosing `registerRoutes` whose span covers the `r.GET` line, asserting resolution lands on `listOrders` — proving the newly-emitted `line` never diverts a named provider to its registrar.
- **Laravel named-controller → prior file-level behavior** — `leaves a Laravel NAMED-controller route at the prior behavior (no closure path)`.
- **File-scope Laravel closure → degrades** — `leaves a FILE-scope Laravel closure at file-level (no enclosing symbol)`.
- **Multi-decorator FastAPI → degrades** — `leaves a multi-decorator FastAPI handler at file-level (line above def-span)`.

Full file is now 183 tests passing; `tsc` clean. The one known follow-up — the pre-existing Go multi-arg-middleware over-match (`r.GET(path, mw, func(){})` emitting two detections) — remains tracked in the PR description's Residual Review Findings and is out of scope for this slice.

gitnexus/src/core/group/extractors/http-patterns/go.ts
```diff
@@ -31,7 +31,7 @@ const FRAMEWORK_ROUTE_PATTERNS = compilePatterns({
             field: (field_identifier) @http_method (#match? @http_method "^(GET|POST|PUT|DELETE|PATCH)$"))
           arguments: (argument_list
             (interpreted_string_literal) @path
-            (identifier) @handler))
+            [(identifier) (func_literal)] @handler))
```
magyargergo: **[P2 · advisory · cross-engine: Claude swarm/ce + Codex ce + Codex adversarial · reproduced]**

The widened `@handler` capture is **unanchored** after `@path`, so `r.GET("/x", mw, func(){})` yields two provider detections (`mw` *and* the closure). `dedupeContracts` keys on `…|symbolRef.name` (http-route-extractor.ts:917) so both survive, then `mergeGraphAndSourceContracts` keys on `contractId` alone and keeps the **first** (`mw`) — http-route-extractor.ts:929-936. Net: the route attaches to the middleware, the **same wrong attribution as pre-PR**, so the new inline-containment benefit is neutralized for this multi-arg shape (one rare ordering — closure-first, named-last — is a genuinely net-new mis-attribution to the registrar). Same pattern at go.ts:54 (HandleFunc).

Pre-existing and deferred per the PR's scope boundary — **not merge-blocking** — but no test pins the multi-arg shape.

**Fix:** anchor `@handler` to the last argument (or post-filter to the final handler node in `scan`), and add a regression test asserting exactly one provider for `r.GET("/x", mw, func(){})`.

magyargergo (commented):
**Verdict: clean, low-risk, well-tested PR — no merge blockers. Findings are advisory (P2/P3) only; the load-bearing resolver invariant is correctly upheld and CI is fully green.**

This PR wires a call-site `line` onto inline/closure HTTP **provider** handler emits across the five non-Node language plugins (go/java/kotlin/php/python.ts) so inline handlers resolve to their containing/closure symbol by line-span containment instead of degrading to file-level. +2 test files (unit +429, new real-LadybugDB integration +114). Closes #2276; follow-up to #2269/#2265.

## How this review was produced (engine breakdown)
- **Claude** ran a swarm (risk-architect, test-ci-verifier, security-boundary) **+** ce personas (correctness, adversarial, api-contract, testing, maintainability, project-standards). All lanes completed.
- **Codex (GPT-5.5 / xhigh)** ran **ce** and **adversarial** live and completed both. The Codex **swarm** leg (B1) evaporated mid-run on a stream disconnect and produced **no** structured findings — it is treated as unavailable, not as a corroboration source. So "cross-engine" below means a Claude lane **and** Codex ce or Codex adversarial agreeing.

## What's solid (credit where due)
- **The load-bearing invariant is correctly handled.** `resolveDetectionSymbol` (http-route-extractor.ts:499–529) short-circuits any provider with a **truthy `name`** to name-only resolution and `return null` on miss — it never falls through to `resolveContainingSymbol`. Only `name == null` providers reach containment. Every inline plugin emit correctly pairs `name:null` **with** `line`; every named (Spring) emit keeps a truthy name. Independently re-read and confirmed across all lanes. **[code-read]**
- **Strong tests, including a real pipeline E2E.** New `http-inline-handler-symbol-roundtrip.test.ts` drives the **real** ingestion pipeline + LadybugDB and asserts an inline Go `HandleFunc` resolves to `main` via `source_scan_resolved`. It is NOT on the sequential-exclusion list, so it runs by default in Vitest + the CI coverage job (no skip guard). Unit tests use exact `toMatchObject` assertions for symbolUid/symbolName/strategy and include regression guards for the named-handler and documented-limitation paths.
- **Documented limitations are guarded.** File-scope Laravel closures and multi-decorator FastAPI handlers degrade to file-level (`symbolUid === ''`) with tests pinning that true-negative.
- **Clean security / standards / typing.** No trust-boundary, network, auth, CI/Docker/YAML, or Cypher-construction change (security lane: no blockers). `git diff --check` and bidi scans clean; new temp dirs use `fs.mkdtemp(os.tmpdir()…)` (CodeQL-insecure-temp gate satisfied). No `any`/`as any`, no `if`-guarded vacuous assertions, CHANGELOG untouched, route logic stays in per-language `http-patterns/*.ts` (RFC #909 shared-core rule honored). `HttpDetection.line` is `line?: number` — additive and type-safe.

## Headline adjudication — Go middleware over-match (the lanes disagreed; resolved here)
For `r.GET("/x", mw, func(){})` the widened **unanchored** capture `[(identifier)(func_literal)] @handler` (go.ts:34/54) yields TWO matches: `mw` (identifier, truthy name → resolves by name / file-level) and the closure (`name:null` → containment to the enclosing fn). The lanes split on what happens next; the true mechanism, traced through the code: `dedupeContracts` keys on `contractId|filePath|name` (http-route-extractor.ts:917) so **both survive** that step → `mergeGraphAndSourceContracts` then keys on **`contractId` alone** (lines 929–936, first-writer-wins) and **collapses to the FIRST emission (`mw`)**. Net result: the route is attributed to the middleware `mw` — the **same wrong attribution that already existed pre-PR** (the old `(identifier)`-only query matched only `mw` for this shape). So the widening's containment benefit is **neutralized** for this multi-arg shape; it is **pre-existing, compounded-but-not-worsened, deferred in the PR body, and untested**. This reconciles ce-correctness ("dedup discards → benign"), ce-api-contract/Codex-B3 ("name in dedup key → two survive"), and ce-adversarial (the genuinely net-new sub-case is only the unusual *closure-first, named-last* ordering `r.GET(p, func(){}, realHandler)`, which now mis-attributes to the registrar — an edge ordering, not the common middleware-first one). **P3 / advisory, not merge-blocking.** Verified against http-route-extractor.ts:913–937. **[code-read]**

## Findings (advisory only — none merge-blocking)

| # | Sev | File:line | Corroboration | Tag | Problem → Fix |
|---|-----|-----------|---------------|-----|---------------|
| 1 | P2 | go.ts:34/54 + http-route-extractor.ts:929–936 | **Cross-engine**: Claude (risk-architect, ce-correctness, ce-adversarial, ce-api-contract AC-2, ce-testing T2) + Codex ce (P2) + Codex adversarial (P2) | [reproduced] | Unanchored `@handler` over-matches middleware in `r.GET(path, mw, fn)`; contractId-only merge collapses to `mw` (pre-existing, neutralizes the new containment benefit; one rare ordering is net-new mis-attribution). No test pins multi-arg Go. → Anchor `@handler` to the last argument (or post-filter to the final handler in `scan`); add a regression test asserting exactly one provider for `r.GET("/x", mw, func(){})`. |
| 2 | P2 | python.ts:1111 | **Cross-engine**: Claude ce-testing (T1, conf 95) + Codex ce (P2) | [code-read] | `line` was added to BOTH the FastAPI `@app` (~1065) **and** `@router`/APIRouter (~1111) provider emits, but the two new #2276 tests (test:893, :968) cover only `@app`. Existing `@router` tests (test:6695/6725/6753) call `extract(null,…)`, so containment / `source_scan_resolved` never runs for `@router`. The `@router` containment path ships untested. → Add two `@router.<verb>` tests with a mock CONTAINING_QUERY: one nested-in-function asserting `source_scan_resolved`, one module-scope asserting `symbolUid === ''`. |
| 3 | P3 | php.ts:169 → http-route-extractor.ts:783 (+ cross-trace.ts:1053–1055) | **Cross-engine**: Claude (ce-api-contract AC-1, ce-correctness F2, ce-maintainability M3, ce-testing T5, risk-architect F2) + Codex ce (rated P1, but downgraded here) | [reproduced] | An unresolved file-scope Laravel **closure** now emits `name:null`, so the `?? 'handler'` fallback flips the persisted `symbolName` from `'route'` → `'handler'`. `providerLabel` (cross-trace.ts:1055) suppresses `'handler'`/`'fetch'` as anonymous but **not** `'route'`, so an unresolved Laravel closure now shows `<contractId handler>` in `group_trace`/`group_cross_impact` MCP output instead of `route`. **No production code keys on the literal `'route'`** (grepped all of `group/` + `ingestion/`); bridge node identity excludes symbolName and is fully rebuilt each sync, so no cross-run corruption. Cosmetic display delta. → Either add `'route'` to the `providerLabel` generic-sentinel list (it was always a synthetic placeholder, never a real symbol name), or pin `symbolName` in the closure/named-controller tests. |
| 4 | P3 | java.ts:748 / kotlin.ts:1021 | Single-engine (Claude: ce-maintainability M1, ce-testing T3, risk-architect suspicion) | [code-read] | Spring emits pass `name: route.methodName` (java, typed `string|null`) / `name: nameNode?.text ?? null` (kotlin) with `line` now set. The "inert" comments depend on `name` always being truthy — an invariant the **grammar enforces** (the TSQ requires the method-name node) but the **type does not**. A hypothetical null-name match would reach containment, though it would land on the method's own span (still the correct symbol, not a wrong one). No negative regression test. → Narrow the type to `string` or guard `if (!route.methodName) continue;`; optionally add a cross-file comment pointing at the http-route-extractor.ts:499 invariant. Very low risk. |

## Refuted suspicions (validation is a feature)
- **Truthy-name → containment bypass — REFUTED** (Claude ce-adversarial, ce-correctness; Codex ce). The single call site of `resolveContainingSymbol` is gated by the 499–523 invariant; no plugin sets a truthy name and bypasses it. **[code-read]**
- **Go "duplicate contract in output" — REFUTED.** The contractId-only merge collapses to one (mis-attributed) contract, not a duplicate (this is the mechanism behind Finding 1, not a separate output-duplication bug).
- **PHP named-controller mis-fire to `name:null` — REFUTED** (ce-adversarial, Codex adversarial). The `@closure` capture is anchored to the 2nd argument; `[Ctrl::class,'m']` and `'Ctrl@index'` both yield one detection with `name:'route'`, no closure captured. **[reproduced by lanes on real grammar]**
- **Off-by-one / anchor lands inside a WRONG symbol — REFUTED.** The Go/PHP anchor is the registration/closure-start line, always inside the registrar's span; Python single-decorator resolves via the resolver's defensive `pick(line)` second probe. Worst case is documented under-resolution to file-level, never wrong-symbol. **[reproduced]**
- **Java/Kotlin `line` is a live risk — REFUTED as a risk** (see Finding 4): inert today, and benign even in the impossible null-name case.
- **Vacuous limitation tests — REFUTED.** The file-scope-closure / multi-decorator tests assert `symbolUid === ''` with the provider defined, exercising the real detection path (a mis-attribution would make `symbolUid` truthy and fail the assertion).

## CI status (re-checked at synthesis time)
**Fully green.** All checks pass: Build & Push gitnexus (12m38s), tests/ubuntu coverage (15m41s), macOS + Windows platform-sensitive, tree-sitter ABI (all 3 OS), typecheck/lint/format, CodeQL, gitleaks, packaged-install smoke (ubuntu/windows). `e2e (chromium)` shows `skipping` (web module unchanged — conditional skip, not a failure). The "Build/tests pending at review time" caveat from the lanes has resolved to pass.

## Coverage note
Two real test gaps worth closing in a follow-up, both already named above: (1) the FastAPI **`@router`** containment path (Finding 2) — the one in-diff line that ships with zero containment coverage; (2) the Go **multi-arg middleware** shape (Finding 1) — the exact surface the query widening changed. Neither blocks merge: `@router` FastAPI routes are graph-backed (this source-scan path is a fallback), and the Go multi-arg case is no worse than pre-PR.

---
*Automated multi-tool digest (Claude swarm + ce + Codex ce/adversarial; Codex swarm leg evaporated). Verify findings against the code before acting — all are advisory; none block merge.*

gitnexus/src/core/group/extractors/http-patterns/python.ts
```diff
@@ -1101,6 +1107,8 @@ export const PYTHON_HTTP_PLUGIN: HttpLanguagePlugin = {
           method: httpMethod,
           path: p,
           name: null,
+          // Best-effort containment fallback — see the @app provider note above.
+          line: pathNode.startPosition.row + 1,
```
magyargergo, line 1111: **[P2 · advisory · cross-engine: Claude ce-testing + Codex ce · code-read]**

`line` was added to the FastAPI **`@router`/APIRouter** provider emit here, but no test exercises its containment path: the two new #2276 FastAPI tests cover only `@app.*`, and the existing `@router` tests call `extract(null, …)` (no `dbExecutor`), so `resolveDetectionSymbol` / `source_scan_resolved` never runs for `@router`. This is the one in-diff line that ships with zero containment coverage.

Not blocking — `@router` routes are graph-backed and this source-scan path is a fallback.

**Fix:** add two `@router.<verb>` tests with a mock `CONTAINING_QUERY` — one nested-in-function asserting `source_scan_resolved`, one module-scope asserting `symbolUid === ''`.

gitnexus/src/core/group/extractors/http-patterns/php.ts
```diff
@@ -150,12 +152,22 @@ export const PHP_HTTP_PLUGIN: HttpLanguagePlugin = {
...
+      // (routes/web.php) has no enclosing function and PHP closures are not yet
+      // indexed as symbols, so it still degrades to file-level (see #2276).
+      const closureNode = match.captures.closure;
       out.push({
         role: 'provider',
         framework: 'laravel',
         method: methodNode.text.toUpperCase(),
         path,
-        name: 'route',
+        name: closureNode ? null : 'route',
```
magyargergo, line 169: **[P3 · advisory · cross-engine: Claude ce/risk + Codex ce · reproduced]**

Flipping an unresolved file-scope Laravel closure to `name: null` makes the `?? 'handler'` fallback (http-route-extractor.ts:783) change the persisted `symbolName` from `'route'` → `'handler'`. `providerLabel` (cross-trace.ts:1053-1055) suppresses `'handler'`/`'fetch'` as anonymous but **not** `'route'`, so these routes now render as `<contractId handler>` in `group_trace` / `group_cross_impact` MCP output instead of `route`.

Cosmetic: no production code keys on the literal `'route'`, and bridge node identity excludes `symbolName` (rebuilt each sync), so no cross-run corruption. P3.

**Fix:** add `'route'` to the `providerLabel` generic-sentinel set (it was always a synthetic placeholder, never a resolved symbol name), or pin `symbolName` in the closure / named-controller tests.

Issue #1489, opened 2026-05-10
gitnexus_query crashes via MCP with "bm25Results is not iterable" — FTS5 extension unavailable in MCP process


MiaomiaoShi1004:
## Summary

`gitnexus_query` crashes when invoked through the MCP server with `Error: bm25Results is not iterable`. The same query, run through the CLI on the same machine and same index, returns results successfully. This points to FTS5 being unavailable specifically in the MCP server's SQLite context.

The hook log emitted before each MCP tool call confirms it:

```
GitNexus: FTS extension unavailable; continuing without FTS features. load-only policy: extension not pre-installed
```

So the warning is correctly surfaced, but the downstream code in `query` doesn't guard against the FTS-disabled path — `bm25Results` ends up `undefined` and the result-fusion step tries to iterate it.

## Environment

- gitnexus: **1.6.4** (also reproduced on 1.6.2 before upgrading)
- Node: v25.9.0
- macOS, arm64

## Reproduction

1. `npx gitnexus analyze --force` (succeeds, all symbols indexed)
2. From a Claude Code session with the gitnexus MCP server connected, invoke `mcp__gitnexus__query` with any query string and an explicit `repo` parameter.
3. Crash: `Error: bm25Results is not iterable`

## Comparison: CLI works, MCP doesn't

Same machine, same install, same index:

```bash
$ npx gitnexus query "any query string" --repo my-repo
# returns JSON with definitions[] populated, processes[] empty — no crash
```

vs.

```
mcp__gitnexus__query({ query: "any query string", repo: "my-repo" })
# Error: bm25Results is not iterable
```

## What still works through MCP

These all work fine (no FTS dependency):
- `gitnexus_context`
- `gitnexus_impact`
- Resource reads: `gitnexus://repo/{name}/clusters`, `/processes`, `/process/{name}`, `/context`

So the bug is isolated to the BM25 hybrid-ranking path inside `query`.

## Likely fix

Either:
1. Guard the result-fusion code so a missing `bm25Results` falls back to semantic-only / definitions-only ranking instead of crashing, or
2. Make FTS5 a hard requirement and fail loudly at MCP server startup with a clear message, instead of warning and then crashing on first query.

Option 1 is preferable since FTS5 may be genuinely unavailable in some sandboxed contexts (the "load-only policy: extension not pre-installed" warning in the MCP process suggests this is what's happening).

Happy to test a patch.

magyargergo:
[...]

PR #2283, opened 2026-06-23
fix(mcp): tolerate adapter-materialized line:0 in impact callgraph mode (#2279)

### LINKED ISSUE #2279: impact callgraph mode fails when agent appends omitted optional line as 0
## Area

## Summary

The `impact` MCP tool fails in callgraph mode when an agent/tool adapter materializes the omitted optional `line` parameter as `line: 0`.

## Context

The latest version exposes `impact` with two modes:

- `mode: "callgraph"`
- `mode: "pdg"`

`line` is only meaningful for PDG statement anchoring. For callgraph impact, the agent does not intentionally provide `line`; however, the tool call payload ends up including `line: 0`.

From local source inspection:

- The `impact` schema defines `line` as optional.
- `line` has `minimum: 1`.
- The schema does not define `default: 0`.
- The CLI forwards `line` only when explicitly supplied.
- The backend rejects any provided `line` unless `mode: "pdg"`.

This suggests the bug is likely at the MCP client / agent adapter boundary: omitted optional numeric fields are being serialized as `0` instead of being omitted.

## Expected behavior

Calling callgraph impact without a line should work.

Expected payload:

```json
{
  "target": "SomeSymbol",
  "direction": "upstream",
  "mode": "callgraph"
}
```

The call should route to the normal callgraph/BFS impact path.

## Actual behavior

The agent/tool adapter sends:

```json
{
  "target": "SomeSymbol",
  "direction": "upstream",
  "mode": "callgraph",
  "line": 0
}
```

The call fails because:

- `line: 0` violates the schema minimum of `1`.
- If it reaches the backend, callgraph mode rejects any provided `line` because `line` is PDG-only.

## Steps to reproduce

1. Invoke the `impact` MCP tool through an agent/tool adapter.
2. Use callgraph mode without intentionally passing `line`:

   ```json
   {
     "target": "SomeSymbol",
     "direction": "upstream",
     "mode": "callgraph"
   }
   ```

3. Observe that the actual serialized tool arguments include:

   ```json
   {
     "line": 0
   }
   ```

4. The impact call fails before callgraph traversal runs.

## Logs / screenshots

Observed invalid payload shape:

```json
{
  "target": "SomeSymbol",
  "direction": "upstream",
  "mode": "callgraph",
  "line": 0
}
```

Representative failure cause:

```text
line must be >= 1
```

or backend-level rejection equivalent to:

```text
Parameter 'line' is only supported with mode:'pdg'.
```

## Notes for maintainers

This does not appear to be caused by GitNexus setting a default of `0`; local source inspection shows `line` is optional and has no `default`.

Possible fixes:

- Preferred: fix the MCP client / agent adapter to omit unset optional numeric fields.
- Defensive GitNexus option: split callgraph and PDG impact into separate tools, or use a conditional schema where `line` is only valid for `mode: "pdg"`.
- Preserve current validation that `mode: "pdg"` with `line: 0` is invalid.

magyargergo:
## Summary

Fixes #2279. The `impact` MCP tool failed in `mode:"callgraph"` when an MCP
client/agent adapter serialized the *omitted* optional `line` parameter as
`line: 0` instead of dropping it. GitNexus never defaults `line` to `0`; the
spurious value is introduced at the client boundary and broke the call at two
points:

1. **Client-side** — strict clients that validate the advertised schema rejected
   the request because `line` declared `minimum: 1`, so it never reached GitNexus.
2. **Backend-side** — `_impactImpl` rejected it because `line` is a PDG-only
   statement anchor and is forbidden on the callgraph path.

## What changed

- **Backend tolerance** (`local-backend.ts`): `_impactImpl` now treats a literal
  `line: 0` as *omitted* on the callgraph path (`effectiveLine`) and runs the
  normal symbol→symbol BFS. The coercion is deliberately narrow — only the
  literal `0`, only when `mode !== 'pdg'`.
- **Advertised schema** (`tools.ts`): `impact` `line.minimum` relaxed `1 → 0`,
  with a description note that `0` (or omission) means "no statement anchor" and
  `mode:'pdg'` still requires a positive line. The advertised schema is advisory;
  the backend remains the real gate.

Invariants preserved: `pdg` + `line:0` is still rejected (there is no 1-based
source line `0` to anchor on); a genuine *positive* `line` on callgraph still
errors; negative/fractional lines still error.

### Incidental root-cause fix: log severity for swallowed query degradations

While running the test suite, an ERROR-level log (`level:50`) surfaced during a
*passing* test that intentionally injects a slice-callees query failure. Root
cause: `logQueryError` — the shared handler for **swallowed, gracefully-degraded**
best-effort query failures (~24 call sites, each catches and continues with a safe
fallback) — logged all of them at `logger.error`, the same severity as fatal
failures. That is a false alarm that also drowns genuine errors.

Fixed the severity to match reality:
- benign missing optional table/label/column (e.g. a repo without
  processes/communities, or a pre-v3 PDG index lacking the `calleeIds` column —
  which fails on *every* pdg-downstream impact for such an index) → `debug`;
- any other swallowed failure → `warn`;
- `error` is reserved for failures that actually abort an operation.

Also fixed the sibling bm25/FTS fallback, which logged its swallowed
"FTS indexes may not exist" degradation at `error` while its own import-failure
sibling already used `warn`.

## Testing

- Regression matrix for #2279 (callgraph `line:0` runs the BFS and is
  byte-identical to omitting `line`; positive line on callgraph still errors;
  `pdg` + `line:0` still errors; advertised schema advertises `minimum:0` + the
  `0/omitted = no anchor` description).
- The slice-callees degradation test now asserts `warn` (level 40), not `error`;
  a new test covers the benign-missing-table → `debug` branch.
- bm25/FTS warn severity pinned in the existing #1489 test.
- Full unit suite green: 432 files, 8982 tests passing.

## Residual Review Findings

The production-readiness review surfaced one **pre-existing, out-of-scope** finding,
recorded here for durability:

- **[P2 · pre-existing · not addressed]** `gitnexus/src/mcp/local/local-backend.ts`
  (`rename:apply-edit` catch, ~L4171): a swallowed `fs.writeFile` failure is logged
  and the rename still returns `{ status: 'success', applied: true }`, so a
  *partial* rename is reported as complete. This predates this PR and is unrelated
  to #2279. The correct fix is in the rename **return shape** (e.g. `status:'partial'`
  + `write_errors[]`), not the log level — so it was intentionally left out of this
  PR's logging-severity change (which keeps the failure visible at `warn`).
  Recommend tracking as its own issue.

### DIFF
--- a/gitnexus/src/mcp/local/local-backend.ts
+++ b/gitnexus/src/mcp/local/local-backend.ts
@@ -266,10 +266,31 @@ export const IMPACT_RELATION_CONFIDENCE: Readonly<Record<string, number>> = {
 const confidenceForRelType = (relType: string | undefined): number =>
   IMPACT_RELATION_CONFIDENCE[relType ?? ''] ?? 0.5;
 
-/** Structured error logging for query failures — replaces empty catch blocks */
+/**
+ * Structured logging for *swallowed* query failures — replaces empty catch
+ * blocks. Every caller of this helper catches the failure and degrades to a
+ * safe fallback (the operation still returns a result), so these are NOT
+ * operation-level errors and must not be logged at `error`:
+ *
+ *  - A benign missing optional table/label/column — a repo analyzed without
+ *    processes/communities, or a pre-v3 PDG index lacking the `calleeIds`
+ *    column — is a normal configuration, not a failure. Logged at `debug`
+ *    (suppressed at the default `info` level; surfaced only when troubleshooting).
+ *  - Any other swallowed failure is an unexpected-but-handled degradation:
+ *    logged at `warn` so it stays observable without raising a false `error`
+ *    alarm that would drown genuine, operation-aborting failures.
+ *
+ * `error` is intentionally NOT used here — it is reserved for failures that
+ * actually abort an operation, which log directly rather than through this
+ * best-effort-degradation helper.
+ */
 function logQueryError(context: string, err: unknown): void {
   const msg = err instanceof Error ? err.message : String(err);
-  logger.error({ context, err: msg }, 'GitNexus query failed');
+  if (isBenignMissingTableError(err)) {
+    logger.debug({ context, err: msg }, 'GitNexus query skipped (missing optional data)');
+    return;
+  }
+  logger.warn({ context, err: msg }, 'GitNexus query failed (degraded)');
 }
 
 /**
@@ -1955,9 +1976,14 @@ export class LocalBackend {
     try {
       ftsResponse = await searchFTSFromLbug(query, limit, repo.lbugPath);
     } catch (err: any) {
-      logger.error(
+      // Swallowed, gracefully-degraded failure: the search falls back to
+      // semantic-only (a valid result), and the most common cause is simply an
+      // un-indexed FTS extension — a normal configuration, not an operation
+      // error. Logged at warn (matching the sibling import-failure fallback
+      // above), never error, so it does not raise a false alarm.
+      logger.warn(
         { err: err.message },
-        'GitNexus: BM25/FTS search failed (FTS indexes may not exist) -',
+        'GitNexus: BM25/FTS search failed (FTS indexes may not exist) — falling back to semantic-only',
       );
       return { results: [], ftsUsed: false };
     }
--- a/gitnexus/src/mcp/local/local-backend.ts
+++ b/gitnexus/src/mcp/local/local-backend.ts
@@ -268,9 +268,10 @@ const confidenceForRelType = (relType: string | undefined): number =>
 
 /**
  * Structured logging for *swallowed* query failures — replaces empty catch
- * blocks. Every caller of this helper catches the failure and degrades to a
- * safe fallback (the operation still returns a result), so these are NOT
- * operation-level errors and must not be logged at `error`:
+ * blocks. The level reflects telemetry severity, NOT a promise about the
+ * caller: most callers catch the failure and degrade to a genuinely safe
+ * fallback (a usable result, usually with a caller-visible `partial`/`ftsUsed`
+ * flag), so these are not operation-level errors and must not log at `error`:
  *
  *  - A benign missing optional table/label/column — a repo analyzed without
  *    processes/communities, or a pre-v3 PDG index lacking the `calleeIds`
@@ -283,6 +284,13 @@ const confidenceForRelType = (relType: string | undefined): number =>
  * `error` is intentionally NOT used here — it is reserved for failures that
  * actually abort an operation, which log directly rather than through this
  * best-effort-degradation helper.
+ *
+ * Contract for callers (#2283 review): only route a failure here when the
+ * caller ALSO surfaces the degradation in its result (a `partial` flag,
+ * `failed_files`, `traversalComplete:false`, …). A mutating or safety-critical
+ * path that would otherwise report success/clean (e.g. `rename` apply, the
+ * `detect_changes` safety gate) MUST set that result-level signal — `warn`
+ * alone is not a substitute for an honest result.
  */
 function logQueryError(context: string, err: unknown): void {
   const msg = err instanceof Error ? err.message : String(err);
@@ -303,7 +311,12 @@ function logQueryError(context: string, err: unknown): void {
  */
 function isBenignMissingTableError(err: unknown): boolean {
   const msg = err instanceof Error ? err.message : String(err ?? '');
-  return /does not exist|no such (table|label|rel)|unknown (table|label)|not (defined|found)/i.test(
+  // The `not (defined|found)` arm is scoped to a schema object (table/label/
+  // rel/column/property), mirroring lbug-adapter's isMissingColumnError
+  // (`/(table|column|property).*not found/i`): an unscoped "not found" matched
+  // operation failures like `rg: not found` (ripgrep absent) or `Symbol not
+  // found`, which this helper would then silently demote to `debug` (#2283).
+  return /does not exist|no such (table|label|rel)|unknown (table|label)|(table|label|rel|column|property)[^\n]*\bnot (defined|found)\b/i.test(
     msg,
   );
 }
@@ -3851,6 +3864,9 @@ export class LocalBackend {
 
     // Map diff hunks to indexed symbols via range overlap
     const changedSymbols: any[] = [];
+    // Set if a swallowed graph query fails below — surfaces `partial:true` so a
+    // degraded run cannot report a false-clean `risk_level:'low'` (#2283).
+    let queryDegraded = false;
     for (const fileDiff of fileDiffs) {
       if (fileDiff.hunks.length === 0) continue;
 
@@ -3898,6 +3914,12 @@ export class LocalBackend {
         }
       } catch (e) {
         logQueryError('detect-changes:file-symbols', e);
+        // The symbol query failed: changedSymbols stays empty and the result
+        // would otherwise look like a clean no-op (`changed_count:0`,
+        // `risk_level:'low'`). detect_changes is the pre-commit safety gate, so
+        // flag the result `partial` rather than let a swallowed failure
+        // masquerade as "nothing changed" (#2283).
+        queryDegraded = true;
       }
     }
 
@@ -3936,6 +3958,7 @@ export class LocalBackend {
         }
       } catch (e) {
         logQueryError('detect-changes:process-lookup', e);
+        queryDegraded = true;
       }
     }
 
@@ -3958,6 +3981,9 @@ export class LocalBackend {
       },
       changed_symbols: changedSymbols,
       affected_processes: Array.from(affectedProcesses.values()),
+      // A swallowed query failure makes the counts/risk above incomplete — tell
+      // the caller so the safety gate isn't trusted as a clean result (#2283).
+      ...(queryDegraded && { partial: true }),
     };
   }
 
@@ -4158,6 +4184,7 @@ export class LocalBackend {
     const allChanges = Array.from(changes.values());
     const totalEdits = allChanges.reduce((sum, c) => sum + c.edits.length, 0);
 
+    const failedFiles: string[] = [];
     if (!dry_run) {
       // Apply edits to files
       for (const change of allChanges) {
@@ -4168,13 +4195,17 @@ export class LocalBackend {
           content = content.replace(regex, new_name);
           await fs.writeFile(fullPath, content, 'utf-8');
         } catch (e) {
+          // A swallowed write failure must not be reported as a full success
+          // (#2283): record the file so the result can degrade to 'partial'
+          // with the unwritten files listed, rather than masquerading as done.
           logQueryError('rename:apply-edit', e);
+          failedFiles.push(change.file_path);
         }
       }
     }
 
     return {
-      status: 'success',
+      status: failedFiles.length > 0 ? 'partial' : 'success',
       old_name: oldName,
       new_name,
       files_affected: allChanges.length,
@@ -4183,6 +4214,7 @@ export class LocalBackend {
       text_search_edits: astSearchEdits,
       changes: allChanges,
       applied: !dry_run,
+      ...(failedFiles.length > 0 && { failed_files: failedFiles }),
     };
   }
 
@@ -4878,7 +4910,11 @@ export class LocalBackend {
         symType,
         direction,
         maxDepth,
-        line: params.line,
+        // Use the normalized line, not raw params.line, so the gate and the
+        // engine share one source of truth (#2283). Identity in pdg mode today
+        // — effectiveLine === params.line when mode === 'pdg' — but this stays
+        // correct if the normalization ever stops being an identity here.
+        line: effectiveLine,
         limit: Number.isFinite(params.limit) ? params.limit : 100,
         // KTD2 extraction-seam discipline: hand the engine its DB dependency
         // explicitly rather than `this.`-binding it. LocalBackend owns repo
--- a/gitnexus/src/mcp/local/local-backend.ts
+++ b/gitnexus/src/mcp/local/local-backend.ts
@@ -4490,9 +4490,21 @@ export class LocalBackend {
     }
     const mode = modeResult.mode;
 
+    // #2279: some MCP client/agent adapters serialize an *omitted* optional
+    // numeric field as `0` rather than dropping it, so callgraph calls arrive
+    // carrying a spurious `line: 0`. `line` is meaningless on the callgraph path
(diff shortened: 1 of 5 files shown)


gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -266,10 +266,31 @@ export const IMPACT_RELATION_CONFIDENCE: Readonly<Record<string, number>> = {
...
+ * best-effort-degradation helper.
+ */
 function logQueryError(context: string, err: unknown): void {
   const msg = err instanceof Error ? err.message : String(err);
-  logger.error({ context, err: msg }, 'GitNexus query failed');
+  if (isBenignMissingTableError(err)) {
+    logger.debug({ context, err: msg }, 'GitNexus query skipped (missing optional data)');
+    return;
+  }
+  logger.warn({ context, err: msg }, 'GitNexus query failed (degraded)');
```
magyargergo, line 301: **[P2 · advisory · code-read]** The `error → {debug | warn}` downgrade here is applied through the **shared** `logQueryError` helper, which serves ~23 call-sites of **mixed contract**. It's correct for the best-effort enrichment callers (they return a usable result, usually with a caller-visible `partial:true` / `ftsUsed:false` flag), but it lowers telemetry for callers that do **not** degrade safely — and the new docstring above (line 271) over-claims that *"every caller … degrades to a safe fallback (the operation still returns a result)."* False for at least:

• `rename:apply-edit` (~4171, mutating path): a swallowed `fs.writeFile` failure still returns `status:'success'` / `applied: !dry_run` with no `failed_files` — a partial rename reported as success, now at warn/debug instead of error.
• `detect-changes:file-symbols` / `:process-lookup` (~3900 / ~3938): a query failure returns `changed_count:0`, `risk_level:'low'` with **no** partial flag — a real changeset looks like a clean low-risk no-op (this is the pre-commit safety gate). `impact:depth-traversal` (5396), by contrast, self-mitigates via `partial:true` at line 5735.
• `isBenignMissingTableError` (304-309) is **context-blind** text matching: a required-table `"Table X does not exist"`, or `"rg: not found"` at `rename:ripgrep` (4154), now demotes to `debug` — invisible at the default `info` level.

**Blocks merge: No.** The swallow-and-succeed in rename/detect-changes is **pre-existing** — this PR only lowers the log level (an observability regression), it does not introduce the silent success. Fix options: give `logQueryError` a severity/contract param, or scope the downgrade to enrichment callers while keeping `error`/returning `partial` for the mutating + safety-critical callers; narrow the docstring claim; add pin tests for rename-write-failure and detect_changes-query-failure.

_Cross-engine corroboration: Codex/GPT-5.5 (swarm-risk P2, ce P2, adversarial [high]) + Claude (reliability, adversarial, risk, maintainability)._

gitnexus/src/mcp/tools.ts
```diff
@@ -469,9 +469,14 @@ SERVICE: optional monorepo path prefix (case-sensitive path segments). When "rep
...
-          minimum: 1,
+          // `minimum: 0` (not 1) so strict client/agent adapters that materialize
+          // an omitted optional numeric field as `0` do not reject the request
+          // before sending (#2279). A positive line is still required for a real
+          // pdg anchor — the backend enforces that — but `0`/omitted means "no
+          // statement anchor" and is tolerated on the callgraph path.
+          minimum: 0,
           description:
-            "1-based source line — PDG statement anchor (mode:'pdg'). Seeds affectedStatements on the statement at this line; inter-procedural symbols are still returned in interproceduralByDepth/pdgInterprocedural and the compatibility byDepth bucket.",
+            "1-based source line — PDG statement anchor (mode:'pdg'). Seeds affectedStatements on the statement at this line; inter-procedural symbols are still returned in interproceduralByDepth/pdgInterprocedural and the compatibility byDepth bucket. 0 (or omitted) means no statement anchor; mode:'pdg' requires a positive line.",
```
magyargergo: **[P3 · code-read]** This new `line` description ends *"…0 (or omitted) means no statement anchor; mode:'pdg' requires a positive line."* — which **contradicts** the top-level `impact` description at line 418: *"Without 'line', pdg returns whole-symbol inter-procedural reach plus local whole-symbol PDG diagnostics."* A `pdg` call **without** a line is a valid (degraded whole-symbol) call — `pdg-impact.ts:1795-1798` takes the whole-symbol-seed path, it does not error. "requires a positive line" misleads an agent: it may avoid valid no-line pdg calls, or synthesize `line:0` to satisfy the apparent requirement (which then hits the hard rejection).

Suggest scoping the clause to the callgraph path, e.g. *"omit line for whole-symbol pdg, or pass a positive 1-based line for a statement-anchored slice; literal 0 is tolerated only as an omitted-line compatibility sentinel on the callgraph path and is rejected for mode:'pdg'."*

**Blocks merge: No** (doc-accuracy only). _Cross-engine: Codex ce (P3) + Claude api-contract._

magyargergo (commented):
## Tri-engine review — PR #2283 · `fix(mcp): tolerate adapter-materialized line:0 in impact callgraph mode (#2279)`

**Engines:** Claude (swarm: risk, test/CI, security, hygiene + ce personas: correctness, testing, maintainability, project-standards, adversarial, api-contract, reliability) and Codex/GPT-5.5-xhigh (swarm + ce + adversarial — all three legs live with real findings). Treat this as **two independent engines, not five separate confirmations**: agreement within one engine is consistency; agreement *across* Claude and Codex is the strong signal.

### Verdict: production-ready with minor follow-ups — no blocking issues; recommend merge.

This is a correct, well-tested, low-risk fix (posted as a collaborative `COMMENT`, not a formal approval). The core `line:0` normalization is sound across every cell of the truth table, was attacked from multiple angles on both engines and **held**, CI is green (`mergeStateStatus: CLEAN`), and branch hygiene is clean (`clean feature/fix PR`; merge state `mergeable`). The items below are **advisory / follow-up, not merge blockers** — the one substantive theme (a shared-helper log-level downgrade) is an *observability* regression layered on *pre-existing* behavior, plus a self-contradicting doc string.

---

### What's solid (validated, both engines)

- **The `line:0` fix is correct.** `effectiveLine = mode !== 'pdg' && params.line === 0 ? undefined : params.line` is read only by the two validation gates. On the callgraph path a spurious `0` becomes `undefined` → BFS runs identically to an omitted line; on `pdg` the ternary is an identity, so `line:0` still hits the positive-integer gate and is rejected before traversal. The PR's `expect(callgraphZero).toEqual(omitted)` byte-identity test pins this. pdg-impact.ts:1798 adds defense-in-depth (`line >= 1`), so no BasicBlock can anchor at line 0 even hypothetically. Truth table (callgraph+0→BFS; callgraph+positive/neg/frac→error; pdg+0/neg/frac→error; pdg+positive→proceeds) holds in every cell, including `-0`/`0.0`.
- **Schema `minimum: 1 → 0` is additive and backward-compatible** — strictly relaxes client-side validation; the backend still rejects `pdg+0` and negatives. Credit the back-compat care.
- **No stdout pollution.** All new `logger.debug`/`logger.warn` calls route through pino to stderr (fd 2); no `console.*` introduced; MCP stdio transport unaffected.
- **Hygiene clean.** 4 linear commits off base, no merge-from-main, CHANGELOG untouched (correct per repo policy), conventional `fix(mcp):` titles.

---

### Headline finding — shared-helper log downgrade reduces telemetry for non-degrading callers + over-claiming docstring  ·  P2, advisory  ·  [INLINE → local-backend.ts:293]

The `error → {debug if isBenignMissingTableError else warn}` rework is applied through the **shared** `logQueryError` helper (local-backend.ts:287-293, in diff), used by ~23 call-sites of **mixed contract**. It is directionally correct for the ~20 best-effort enrichment callers, which return a usable result and usually surface a caller-visible flag (`partial:true` / `ftsUsed:false` / `enrichmentDegraded`). But it lowers telemetry for callers that **do not** degrade safely, and the new docstring over-claims: *"Every caller of this helper catches the failure and degrades to a safe fallback (the operation still returns a result)."* That is false for at least three sites (all confirmed by code-read in the worktree):

- **`rename:apply-edit`** (local-backend.ts:~4171, NOT in diff — mutating path, `dry_run:false`): an `fs.writeFile` failure is swallowed and the function still returns `status:'success'` / `applied: !dry_run` with no `failed_files`. A partially-applied rename reports as full success — now at `warn` (or `debug` if the message matches the benign regex) instead of `error`. **Blocks merge: No** (pre-existing; PR changes log level only).
- **`detect-changes:file-symbols` / `:process-lookup`** (local-backend.ts:~3900 / ~3938, NOT in diff): a query failure leaves `changed_count:0`, `risk_level:'low'` with **no partial flag** — a real changeset reports as a clean low-risk no-op. This is the pre-commit safety gate that CLAUDE.md tells agents to trust (`MUST run detect_changes() before committing`). Was `error`, now `warn`/`debug`.
- **`isBenignMissingTableError`** (regex `does not exist|no such (table|label|rel)|unknown (table|label)|not (defined|found)`, local-backend.ts:304-309, NOT in diff) is **context-blind text matching**. A required-table `"Table X does not exist"` on `impact:depth-traversal` (5396) or `"rg: not found"` at `rename:ripgrep` (4154) now demotes to `debug` (invisible at the default `info` level). The PR newly applies this demotion uniformly across all callers.

**Honest framing (preserve this):** the swallow-and-succeed behavior in rename/detect-changes is **pre-existing** — this PR changes only the **log level** (reducing visibility); it does **not** introduce the silent success. So this is an *observability regression + an over-claiming docstring*, **not** a hard merge blocker.

**Fix options (don't overstate):** give `logQueryError` a severity/contract parameter, OR scope the downgrade to enrichment callers while keeping `error`/returning `partial` for the mutating + safety-critical callers; narrow the docstring's "every caller degrades to a safe fallback" claim; add pin tests for rename-write-failure and detect_changes-query-failure so they cannot return clean success / `risk:'low'`. (`impact:depth-traversal` is the one required-table caller that is self-mitigating — it always emits `partial:true` regardless of log level, so observability drops but the correctness signal survives.)

**Corroboration:** Codex swarm-risk (P2), Codex ce (P2 conf100), Codex adversarial ([high], "needs-attention / no ship") + Claude reliability (F1/F2/F3 P2-P3), Claude adversarial (F1/F2 P3), Claude risk-architect (P2/P3), Claude maintainability (M-3, residual). Cross-engine, multiple lanes — strong on existence, but severity capped at P2/advisory because the result bug is pre-existing.

---

### Secondary finding — description self-contradiction misleads agents  ·  P3  ·  [INLINE → tools.ts:479]

The new `line` schema description (tools.ts:479, in diff) ends: *"…0 (or omitted) means no statement anchor; mode:'pdg' requires a positive line."* This **contradicts** the top-level `impact` description at tools.ts:418: *"Without 'line', pdg returns whole-symbol inter-procedural reach plus local whole-symbol PDG diagnostics."* A `pdg` call **without** a line is a valid (degraded whole-symbol) call — pdg-impact.ts:1795-1798 takes the whole-symbol-seed path, it does **not** error. "requires a positive line" misleads an agent: it may avoid valid no-line pdg calls, or synthesize `line:0` to satisfy the apparent requirement → which then hits the hard rejection. Clean to fix in-PR: scope the clause to the callgraph path, e.g. *"omit line for whole-symbol pdg, or pass a positive 1-based line for a statement-anchored slice; literal 0 is tolerated only as an omitted-line compatibility sentinel on the callgraph path and is rejected for mode:'pdg'."* **Blocks merge: No** (doc-accuracy only).

**Corroboration:** Codex ce (P3 conf100) + Claude ce-api-contract (AC-1/AC-2). Cross-engine.

---

### Tertiary items (body only — judge, none blocking)

- **Test gaps (Claude ce-testing, Claude test-ci, Claude maintainability TG):** (a) the benign-debug-branch test is an *absence* assertion — it cannot distinguish "logged at debug" from "not logged at all", so it passes vacuously if the `logQueryError` call were deleted. Tighten by capturing at `debug` and asserting `level === 10`, or spy on `logger.debug`. (b) No explicit test that `line:-1` on **callgraph** still errors (the narrow-coercion contract for negatives is verified only by inspection; the positive-line test pins the same path). Low value but cheap. P3 each.
- **Typing casts (Claude project-standards PS-1/PS-2):** 3 new `(backend as any)` private-method spies + `mode as 'callgraph'|undefined` in the new tests. The user's strict-typing rule forbids *new* `as any`, but these continue the file's pervasive pre-existing private-method-spy idiom (no typed Vitest alternative for private members), and lint/typecheck passed. The `mode as …` cast is the one avoidable addition (type the `it.each` tuple instead). The `if` inside `mockImplementation` is mock dispatch, **not** an assertion guard — not a no-if-branching violation. Advisory, not blocking.
- **Maintainability (Claude M-1):** `effectiveLine` vs `params.line` is a two-name trap — the PDG dispatch at local-backend.ts:4881 still reads `params.line` directly. Identity today, but a future non-identity normalization could silently diverge. Consider threading `effectiveLine` through, or renaming it `callgraphEffectiveLine`. P3.
- **CLI/doc staleness (Claude AC-3):** CLI `--line` help (cli/index.ts:348) doesn't mention the 0-toleration convention. Cosmetic, info-level.

---

### Refuted suspicions — validation is a feature (both engines)

- **`line:0` leaking into BFS output (callgraph vs omitted differ)** — REFUTED. `params.line` never reaches `_runImpactBFS`; byte-identity test holds. (Claude correctness, Claude adversarial Attack 1, Codex.)
- **`line:0` anchoring a PDG slice at line 0** — REFUTED. `pdg+0` is rejected by the positive-integer gate before traversal; pdg-impact.ts:1798 is a second guard. (Claude adversarial Attack 3, Claude security, Codex.)
- **Alternate "omitted" encodings (`"0"`, `null`, `-0`, `0.0`) regressing** — REFUTED. `-0`/`0.0` behave as `0`; `"0"`/`null` were already rejected pre-PR and are schema-illegal for an integer field. No regression. (Claude adversarial Attack 2.)
- **Cypher injection via `params.line`** — REFUTED. `callTool` does no schema coercion; `params.line` is numeric-validated before any query use; no string-injection surface. (Claude ce-adversarial, Codex adversarial, Claude security.)
- **stdout / JSON-RPC pollution from the new log calls** — REFUTED. Logger is stderr-bound. (Claude correctness, Claude adversarial, Claude project-standards PS-5.)

---

### CI status

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. CodeQL, gitleaks, quality/format, and tests on macOS / Ubuntu / Windows all pass. No workflow files changed.

---

*This digest is an automated multi-tool synthesis (Claude swarm+ce and Codex/GPT-5.5 swarm+ce+adversarial), merged with cross-engine confidence weighting and gated by a synthesis critic. Verify findings against the diff before acting.*

---

PR #2286, opened 2026-06-23
feat(ingestion): make doc comments searchable across all languages

### LINKED ISSUE #2270: Java/Kotlin method and class Javadoc not searchable via MCP query
# Java/Kotlin method and class Javadoc not searchable via MCP `query`

  ### Summary

  For Python and PHP repositories, MCP `query({query: "..."})` can find methods
  and classes when the only signal is the docstring / docblock. For Java (and
  other JVM / "leading-comment" languages), the equivalent Javadoc text is
  absent from the embedding, so the same query returns nothing — even though
  the Javadoc exists in source and the graph node has the right `id`.

  This is a **consistency gap**, not a feature scope question: the
  `descriptionExtractor` hook, the `description` column, and the
  embedding-time concatenation of `description` are all already wired and
  working for Python and PHP. Java just never registers the hook.

  ### Reproduction

  Minimal fixture (`Probe.java`):

  ```java
  package demo;

  public class Probe {
      /**
       * Computes the running balance across all user accounts.
       * Uses lazy evaluation and caches intermediate results in Redis.
       * @param userId the unique user identifier
       * @return the computed balance as BigDecimal
       * @deprecated since 2.0, use computeBalanceV2
       */
      public java.math.BigDecimal computeBalance(Long userId) {
          return null;
      }
  }
  ```

  Steps:

  1. Index a Java repo containing a method whose Javadoc contains a unique
     marker word, e.g. `@deprecated since 2.0, use computeBalanceV2`.
  2. Call MCP `query({query: "computeBalanceV2"})`.
  3. **Expected**: the method `Probe.computeBalance` is in the result set.
  4. **Actual**: the method is not returned by semantic search.

  Equivalent query in a Python repo (`@deprecated since 2.0, use
  computeBalanceV2` placed inside a docstring) returns the function
  correctly, because `descriptionExtractor` is registered for Python.

  ### Root cause

  Two independent omissions:

  #### 1. Java provider never registers `descriptionExtractor`

  `gitnexus/src/core/ingestion/languages/python.ts:134` and
  `gitnexus/src/core/ingestion/languages/php.ts:296` both set
  `descriptionExtractor` on the provider. A repo-wide search for
  `descriptionExtractor` shows only those two registrations plus the
  type definition in `language-provider.ts:239`:

  ```
  $ grep -rn "descriptionExtractor" gitnexus/src/core/ingestion/
  gitnexus/src/core/ingestion/language-provider.ts:239:  readonly descriptionExtractor?: (
  gitnexus/src/core/ingestion/languages/python.ts:134:  descriptionExtractor: pythonDescriptionExtractor,
  gitnexus/src/core/ingestion/languages/php.ts:296:    descriptionExtractor: phpDescriptionExtractor,
  gitnexus/src/core/ingestion/workers/parse-worker.ts:2144:      const description = provider.descriptionExtractor?.(nodeLabel, nodeName, captureMap);
  ```

  Java, Kotlin, TypeScript, JavaScript, Go, Rust, C#, C, C++, Ruby, Swift,
  Dart, Cobol, and Vue all skip it. Java is the most visible gap because of
  the Javadoc convention.

  #### 2. The graph `content` column window only captures the closing `*/`

  `gitnexus/src/core/lbug/csv-generator.ts:165-176` builds each symbol's
  `content` from the source file:

  ```ts
  const startLine = node.properties.startLine;       // 1-based
  const endLine = node.properties.endLine;
  const start = Math.max(0, startLine - 2);
  const end = Math.min(lines.length - 1, endLine + 2);
  const snippet = lines.slice(start, end + 1).join('\n');
  ```

  For the fixture above, with `startLine = 11` (the `public ...` line),
  `start = 9`, so the snippet starts at 1-based line 10 — the `@deprecated`
  line. The rest of the Javadoc body is dropped:

  ```
  * @deprecated since 2.0, use computeBalanceV2
  */
  public java.math.BigDecimal computeBalance(Long userId) {
      return null;
  }
  ```

  `description` is empty for Java because of issue 1, so this is the entire
  context the embedder ever sees. The phrases "Computes the running
  balance…", "caches intermediate results in Redis", and
  "@param userId the unique user identifier" are nowhere in the embedding.

  #### 3. The pipeline is otherwise ready

  The plumbing is intact — Java just isn't filling the existing slot:

  - `gitnexus-shared/src/graph/types.ts:66` — `description?: string` on
    every node
  - `gitnexus/src/core/lbug/schema.ts` — `description STRING` column on
    every symbol table
  - `gitnexus/src/core/ingestion/workers/parse-worker.ts:2144, 2280` —
    reads `provider.descriptionExtractor`, writes it onto the symbol
  - `gitnexus/src/core/lbug/csv-generator.ts:487, 523, 541, 574, 595, 611`
    — writes `description` to CSV, `COPY`d into LadybugDB
  - `gitnexus/src/core/embeddings/embedding-pipeline.ts:128, 138, 147` —
    `RETURN n.description AS description`
  - `gitnexus/src/core/embeddings/text-generator.ts:87-94` —
    `description` prepended to the embedding text

  ### Proposed fix

  **Small (recommended for a focused PR):** register
  `descriptionExtractor` for Java and Kotlin, following the
  `pythonDescriptionExtractor` pattern in
  `gitnexus/src/core/ingestion/languages/python.ts:75-101`. The
  implementation walks backwards from each `method_declaration`,
  `constructor_declaration`, `class_declaration`, and
  `interface_declaration` over preceding siblings, picks up the nearest
  `block_comment` that begins with `/**`, strips the leading `*` markers,
  joins lines with spaces, and caps the result at
  `DEFAULT_EMBEDDING_CONFIG.maxDescriptionLength` (150 chars, see
  `gitnexus/src/core/embeddings/types.ts:243`). Add a sibling helper for
  KDoc at the same time since Kotlin has the same convention.

  The Java capture filter at
  `gitnexus/src/core/ingestion/languages/java/captures.ts:183-190`
  already explicitly excludes `block_comment` / `line_comment` from
  argument lists (per #1920). The new extractor does not need any new
  tree-sitter capture — it walks `definitionNode.previousSibling` like
  the existing Ruby `attr_accessor` extractor at
  `gitnexus/src/core/ingestion/call-routing.ts:100-117` already does.

  **Medium:** additionally make the `startLine - 2` window in
  `csv-generator.ts:170` configurable per language via a new
  `commentWindowBefore: number` field on `LanguageProvider`, defaulting
  to `2` so existing behavior is preserved. Set ~20 for Java / Kotlin /
  TS / JS / Go / Rust / C# / C++. This helps languages where the leading
  comment block is long.

  **Large:** both of the above, plus extending `descriptionExtractor`
  coverage to all 14 other providers. Out of scope for a single PR.

  ### Acceptance test

  Add a Vitest fixture under `gitnexus/test/unit/` that:

  1. Parses the `Probe.java` above with the Java language provider.
  2. Asserts `descriptionExtractor('Method', 'computeBalance', captureMap)`
     returns a string containing the substring `"Computes the running
     balance"` (or a stable substring like `"balance across all accounts"`).
  3. Asserts the same for `'Class'` / `'Probe'`, picking up the class-level
     Javadoc if present.

  For the medium fix, add a second assertion that
  `csv-generator.ts:extractContent` returns content containing
  `"Computes the running balance"` when called with the class's
  `startLine` / `endLine`.

  ### Notes

  - Existing tests for `descriptionExtractor` in Python / PHP should be
    re-run unchanged to confirm no regression.
  - The `description` field is already populated for community /
    process nodes via `cluster-enricher.ts` and the wiki generator; this
    change does not touch those paths.
  - The fix is scoped to ingestion only — no lbug schema migration, no
    embedding table rebuild, no MCP server change.

magyargergo:
## Summary

Methods, classes, and other top-level symbols are now findable by MCP `query`
when the only signal is their doc comment — for **every** documentable language,
not just Python and PHP. A Javadoc/KDoc/JSDoc/Doxygen/godoc/RDoc comment now lands
in the symbol's `description`, which the embedding metadata header already
concatenates, so a semantic query for a phrase that appears only in the docs
returns the documented symbol.

Before, a query for `"computeBalanceV2"` (a term living only in a Java
`@deprecated` Javadoc) returned nothing, because Java — and Go, Rust, C#, C/C++,
Kotlin, Swift, Dart, Ruby, TypeScript/JavaScript — never registered the
`descriptionExtractor` hook. Only Python (docstrings) and PHP (Eloquent
metadata) did. This closes that consistency gap.

Closes #2270.

## What changed

- A shared, **language-neutral** `extractLeadingDocComment` helper plus a
  `createLeadingDocDescriptionExtractor` factory in
  `core/ingestion/utils/ast-helpers.ts`. It reads the doc text off a definition
  node's preceding sibling, matching on the comment's text prefix rather than a
  grammar node type (the comment node is named differently per grammar:
  `block_comment`, `multiline_comment`, `comment`, `line_comment`), so it stays
  out of the no-language-names rule for shared ingestion code (AGENTS.md / RFC #909).
- Each provider registers the hook in one line. PHP keeps its richer Eloquent
  metadata and falls through to the docblock only when there's none. Python
  keeps its docstring extractor.

| Doc-comment style | Languages |
|---|---|
| Block (`/** … */`, `/*! … */`) | Java, Kotlin, TypeScript, JavaScript, C, C++, Swift |
| Triple-slash / bang-slash line runs (`///`, `//!`) | Rust, C#, Dart, Swift, C++ |
| godoc (`//` runs) | Go |
| RDoc/YARD (`#` runs) | Ruby |
| Existing (unchanged) | Python (docstrings), PHP (Eloquent, now + docblock fallback) |

Line-comment runs enforce row-adjacency (godoc/rustdoc semantics): a run stops at
a blank line, so an unrelated earlier comment block, a license header, or a Ruby
shebang + `# frozen_string_literal:` magic comment is not absorbed into the first
declaration's description.

This is purely additive — no schema, embedding-config, or `description`-column
changes. The newly-populated descriptions flow through the existing
`query` / `context` / `cypher` MCP tools with no new agent-facing surface.

## Scope

In scope: method/constructor- and type-level (class/interface/enum/record/struct/
object) doc comments. Out of scope (intentional): field/property/variable/const
doc comments, and the issue's secondary "content window" item (`extractContent`'s
`startLine - 2` lookback in `csv-generator.ts`) — that is language-agnostic and
framed as optional in the issue; the `description` path resolves the reported
search failure without it.

## Testing

- New unit tests: shared-helper edge cases (no comment, non-doc block, line
  comment, empty `/** */` and degenerate `/**/`, gutter stripping, multi-line
  `///` runs, `//!` inner doc, `/*!` Doxygen block), Java + Kotlin provider
  behavior, a cross-language registration matrix (14 languages) and a behavior
  probe per comment family (TS/Go/Rust/C#/Ruby/C++/PHP), line-comment adjacency
  (Go/Ruby blank-line non-attachment + two-block adjacency), and PHP
  Eloquent-wins-over-docblock ordering. 47 assertions, all green.
- `tsc --noEmit` clean; eslint 0 errors; prettier clean.

## Residual Review Findings

From the code-review pass on this branch (all safe/deterministic fixes were
applied in `fix(review): apply autofix feedback`; the items below were deferred):

- **P2 · maintainability · `gitnexus/src/core/ingestion/utils/ast-helpers.ts`** —
  the file is ~1140 lines; the doc-comment helpers (`normalizeBlockDocComment`,
  `extractLeadingDocComment`, `DOC_BEARING_LABELS`,
  `createLeadingDocDescriptionExtractor`) could move to a dedicated
  `utils/doc-comment-extractor.ts`. Deferred — structural refactor with import
  churn across 12+ files; out of scope for this PR.

Advisory (no action needed): a file-top `/** … */` license block directly above
the first declaration with no intervening import/package sibling would be taken as
that declaration's description (block adjacency is intentionally not enforced —
some grammars fold the trailing newline into the comment node, making a strict
row check unreliable); C# `///` XML tags are retained verbatim; Dart/Swift are
covered by the registration matrix only (their native grammars can fail to load
in some CI environments) and exercise the identical shared factory via the other
languages.

---

Claude Code


koriyoshi2041 (commented):
Checked this locally on `6ab8ef5`.

What I verified:
- `git diff --check upstream/main...HEAD` passes.
- `npm test -- --run test/unit/leading-doc-comment.test.ts test/unit/leading-doc-description-all-languages.test.ts test/unit/java-description-extractor.test.ts test/unit/kotlin-description-extractor.test.ts` passes: 4 files / 47 tests.
- `cd gitnexus-shared && npm run build -- --pretty false` passes.
- `cd gitnexus && npx tsc --noEmit --pretty false` passes.
- Scoped eslint over the touched ingestion files reports no errors.

The provider-hook shape also looks consistent with the repo rule that shared ingestion helpers stay language-neutral. I did not find a blocker in this pass.

gitnexus/src/core/ingestion/utils/ast-helpers.ts
```diff
@@ -926,6 +926,138 @@ export function findChild(node: SyntaxNode, type: string): SyntaxNode | null {
...
+ *
+ * Normalization mirrors Python docstring handling: strip the comment delimiters
+ * / per-line markers, then collapse whitespace to single spaces so tag content
+ * (`@param`, `@deprecated since 2.0, use computeBalanceV2`) survives.
+ */
+export function extractLeadingDocComment(
+  node: SyntaxNode,
+  lineCommentPrefixes: readonly string[] = DEFAULT_LINE_DOC_PREFIXES,
+): string | undefined {
+  const prev = node.previousNamedSibling;
```
magyargergo: **[P1]** Exported TS/JS decls drop their JSDoc: extractLeadingDocComment reads node.previousNamedSibling, but the TS query captures the INNER function_declaration/class_declaration (tree-sitter-queries.ts:18,27). For `export function`/`export class`/`export default`, the JSDoc precedes the wrapping export_statement, so previousNamedSibling is null and the doc is dropped. Coordinator-REPRODUCED against the real grammar. GitNexus's own JSDoc-param code already handles this exact case (languages/javascript/captures.ts:349-353 walks up to export_statement). Fix: give the factory a wrapperNodeTypes option (TS/JS pass ['export_statement']) and retry from node.parent when previousNamedSibling is null/non-doc; add export-wrapper tests.

_(pr-tri-review [reproduced])_

gitnexus/src/core/ingestion/utils/ast-helpers.ts
```diff
@@ -926,6 +926,138 @@ export function findChild(node: SyntaxNode, type: string): SyntaxNode | null {
...
+  'Constructor',
+  'Class',
+  'Interface',
+  'Enum',
+  'Struct',
+  'Trait',
+  'Record',
+  'Union',
+  'Namespace',
+  'Module',
```
magyargergo: **[P2]** DOC_BEARING_LABELS membership is out of sync with embeddings/EMBEDDABLE_LABELS (embeddings/types.ts = CHUNKABLE ∪ SHORT), so two common shapes get an extracted-but-never-searchable description. (a) Module/Delegate/Annotation are in this set but NOT in EMBEDDABLE_LABELS → description is written to the DB but embedding-pipeline.ts only iterates EMBEDDABLE_LABELS, so it's never embedded → not findable via MCP query (wasted write; the factory JSDoc 'reaches the embedding metadata header' overstates). Affects Rust/Go/Ruby modules, C# delegates, Java @interface. (b) C++ templates: the C++ query emits @definition.template → getLabelFromCaptures (ast-helpers.ts:361, unchanged) maps it to label 'Template', which is in NEITHER set, so `/** marker */ template <typename T> T add(T v){}` drops its docblock at extraction. Fix: reconcile DOC_BEARING_LABELS with what's actually embeddable — drop the non-embeddable labels, or add them (incl. 'Template', or map template→inner Function/Class) to EMBEDDABLE_LABELS + a queryEmbeddableNodes branch — and narrow the JSDoc/claim accordingly.

_(pr-tri-review [code-read])_

magyargergo (commented):
## Multi-engine review — PR #2286 `feat(ingestion): make doc comments searchable across all languages` (Closes #2270)

**Engines.** Two independent engines, not five separate confirmations. **Claude (engine A)** ran the GitNexus PR swarm (risk-architect, test-ci-verifier, security-boundary, docs-dod) + the ce-code-review personas (correctness, adversarial, testing, maintainability, api-contract, performance, project-standards). **Codex GPT-5.5/xhigh (engine B)** ran swarm (Solo), ce (report-only), and adversarial — **all 3 Codex legs produced real findings.** A coordinator reproduction pass (treated as high-trust) verified the headline. The strong signal below is where a `codex-*` lane AND a `claude-*` lane reached the same finding independently, or where something was reproduced end-to-end; Claude-only or Codex-only agreement is labeled "single-engine."

### Verdict: `not production-ready` — one P1 to fix-or-scope before the "across ALL languages" claim ships. Strong, well-tested feature; the core mechanism works end-to-end, but the headline under-delivers for two very common code shapes. *(Posted as a collaborative `COMMENT`, not a blocking review event.)*

The core mechanism is sound and verified end-to-end by code-read across both engines (see credits). The blocker is a **marquee-language scope gap**: **exported TS/JS declarations silently drop their JSDoc** — and most documented top-level TS/JS symbols are exported. Two more common shapes (C++ templates, and Module/Delegate/Annotation labels) are extracted but never reach search. None are regressions or crashes; descriptions were empty before. But they materially narrow the issue-#2270 closure claim.

### Credit what's solid (validated, cross-engine)
- **Core end-to-end path WORKS** for the labels that matter (Function/Method/Constructor/Class/Interface/Struct/Enum/Trait/Impl/Record/Union/Namespace/TypeAlias/Macro). Traced `descriptionExtractor` (parse-worker.ts:2154) → node props (2290) → CSV `description` column → `queryEmbeddableNodes` (n.description) → `buildMetadataHeader` prepend (text-generator.ts:88-94) → embedding/vector search. Confirmed independently by Claude risk-architect, Claude adversarial, Claude/Codex ce, Codex swarm, and docs-dod. Python already worked this way — this PR brings the other languages to parity. "Closes #2270" is valid for the Java/Kotlin/etc. method+type Javadoc use case.
- **Test quality is high.** 47 assertions across 4 new unit files; no `if`-branching, no `any`; issue #2270's acceptance case literally reproduced (computeBalanceV2). Tests import grammars directly (no worker-build/`distWorkerExists` silent-no-op risk) → run in standard `npm test` on all platforms.
- **Standards clean.** RFC #909 language-neutrality preserved (shared code names no language in executable paths; per-language prefixes supplied at the call site). `gitnexus/CHANGELOG.md` untouched (verified, 0 diff lines). PHP Eloquent precedence preserved + tested (Eloquent wins over PHPDoc fallback; both helpers return non-empty-string|null, so the truthiness refactor is behavior-identical).
- **Refuted risks (validation is a feature):** ReDoS (reproduced safe, 200k chars in 9ms / 2000 comments 34ms), CSV injection, inline-Cypher injection, line-comment adjacency off-by-one, degenerate `/**/`, PHP precedence regression, annotation/decorator confusing previousNamedSibling, captureMap type-widening break, all api-contract breaks, and all performance concerns.

### Inline-eligible findings

| # | File:line | Sev | Engines + lanes | Trigger / Evidence | Fix | Tag | Postable inline? |
|---|-----------|-----|-----------------|--------------------|-----|-----|------------------|
| 1 | ast-helpers.ts:983 (export-wrapper anchor) | **P1** | **Codex** swarm (B1) + **coordinator-REPRODUCED** + codebase's own captures.ts:349 | `/** doc */ export function f(){}` / `export class` / `export default` → JSDoc DROPPED. TS query captures inner `function_declaration`/`class_declaration` (queries.ts:18,27); helper anchors `node.previousNamedSibling`, which is null because the doc precedes the wrapping `export_statement`. Reproduced: bare fn works, every export form returns undefined. | Factory `wrapperNodeTypes` option (TS/JS=`['export_statement']`); retry from `node.parent` when previousNamedSibling is null/non-doc — mirrors the existing walk-up at captures.ts:349-353. Add export-wrapper tests. | [reproduced] | **YES** |
| 2 | ast-helpers.ts:1027 (DOC_BEARING_LABELS block) | P2 | **Claude** risk-architect (RA-1) + adversarial + docs-dod (DD-2) + coordinator code-read. *Single-engine (Claude); Codex did not raise.* | Module/Delegate/Annotation are in DOC_BEARING_LABELS but not EMBEDDABLE_LABELS → description written to DB but `embedding-pipeline.ts` never queries those labels → never searchable. Wasted write; factory JSDoc overstates. Rust/Go/Ruby modules, C# delegates, Java @interface. | Trim the 3 from DOC_BEARING_LABELS, OR add to EMBEDDABLE_LABELS + queryEmbeddableNodes branch, OR narrow the claim/JSDoc. | [code-read] | YES (P2 ≥2 Claude lanes + coordinator; cross-engine blind spot noted) |
| 3 | ast-helpers.ts:1032 (DOC_BEARING_LABELS block; true label site getLabelFromCaptures:361, unchanged) | P2 | **Codex** adversarial (B3) + coordinator code-read. *Single-engine (Codex).* | `/** marker */ template <typename T> T add(T v){...}`. C++ query → `@definition.template` → label `'Template'` (ast-helpers.ts:361). `'Template'` is in neither DOC_BEARING_LABELS nor EMBEDDABLE_LABELS → docblock dropped at extraction. | Add `'Template'` to DOC_BEARING_LABELS here AND make searchable (EMBEDDABLE, or map template→inner Function/Class). | [code-read] | YES (verified end-to-end by coordinator) |
| 4 | ast-helpers.ts:987 (block-comment branch) | P2 | **Codex** swarm(B1)+ce(B2)+adversarial(B3) **AND Claude** correctness(F1, reproduced)+risk(RA-3)+adversarial(#3)+docs-dod(DD-7). **CROSS-ENGINE.** | File-top `/**`/`/*!` license block with no package/import sibling → absorbed as first symbol's description → pollutes embedding text. Block branch (987-988) has no row-adjacency check; the line branch (997-1004) does. Reproduced: TS class-first, package-less Java, C/C++ headers. `/* */` style is safe. Author-acknowledged, untested. | Apply the same `prev.endPosition.row >= node.startPosition.row - 1` adjacency check to block comments, or suppress SPDX/license/@file markers; add a negative-assertion test. | [code-read]/[reproduced] | YES (strongest corroboration) |

Finding #1 is the merge blocker. #2–#4 are non-blocking but worth fixing (or explicitly scoping out in the PR/issue claim) before "across all languages" goes in release notes.

### Lower-priority / advisory (P3 cluster — body-only)
- **Go/Ruby directive mis-attribution** (Codex adversarial #2, reproduced; Claude adversarial #2, reproduced — *both engines*, P2-rated by the adversarial lanes but advisory). `//go:build`, `//go:generate`, `//nolint`, and Ruby `# frozen_string_literal: true` sitting directly above a symbol (no blank line) are folded into `description`, polluting embeddings. This is godoc/RDoc adjacency convention vs. directive prefixes. Fix: skip known directive prefixes. *Worth elevating in the body — it's cross-engine reproduced and very common in real Go/Ruby.*
- **No try/catch around `descriptionExtractor`** (parse-worker.ts:2154). CONTESTED: Claude risk/correctness/adversarial say wrap-it-for-symmetry (the sibling `extractTemplateConstraints` at 2134-2148 IS wrapped; a throw drops the whole language group); Codex B1 says not-a-bug. PRE-EXISTING call site (Python/PHP already routed here unwrapped); no constructible throw found across 3 lanes. → P3 robustness follow-up, not blocking.
- **Rust `//!`/`/*!` inner-doc semantics** (Codex ce, single-engine). The shared path treats Rust inner docs (`//!`, `/*!`) as following-item docs (Doxygen semantics); they're really module/inner docs. Low incidence. P3.
- **bidi/control chars not stripped** from description (Claude security-boundary). Pre-existing `sanitizeUTF8` gap (applies to name/content too); description is now a higher-entropy attacker-controlled field. Client-side Trojan-Source rendering concern, not a server exploit. P3 follow-up.
- **Maintainability nits** (Claude maintainability): `ast-helpers.ts` crosses 1000 lines (author-deferred split to `doc-comment-extractor.ts`); `DEFAULT_LINE_DOC_PREFIXES` exported but no external consumer; dart/ruby/rust split imports could merge; `getLabelFromCaptures` not widened to `CaptureMap` while `getDefinitionNodeFromCaptures` was (half-migrated). All P2/P3 cosmetic.
- **Docs** (Claude docs-dod DD-5 + Codex B1): ARCHITECTURE.md language-provider table omits the now-universal `descriptionExtractor` row; PR body/release note should state that **existing embeddings need `--embeddings` regeneration** (plain `analyze` preserves vectors; description only affects embedding text when the pipeline re-runs — contentHashForNode churns them on re-embed). P3 each.

### Validated / refuted suspicions (credit — many lanes did this)
Refuted with evidence: headline common-case end-to-end break (REFUTED — core labels flow correctly); PHP precedence regression; CSV injection; inline-Cypher injection (lbug-adapter unchanged); ReDoS (reproduced safe); line-comment adjacency off-by-one (reproduced); degenerate `/**/` garble (reproduced); annotation/decorator confusing previousNamedSibling (reproduced); captureMap widening break; all api-contract breaks (description field is additive — no shape/schema/version change); all performance concerns (per-symbol cost is noise vs. surrounding parse work). PHP `Property` does NOT fall through to a doc (Property ∉ DOC_BEARING_LABELS) — confirmed by api-contract.

### Test gaps (none blocking)
- **No end-to-end test** (real Javadoc file → `runPipelineFromRepo` → DB `description` populated → searchable). Both engines traced the path WORKING by code-read (a credit), but there is no behavioral proof for the newly-registered languages. The Module/Delegate/Annotation + C++ template gaps (#2/#3) are exactly what an end-to-end test would have caught. P1-flagged by the testing lane as coverage debt, P2 here. Pattern exists in `integration/resolvers/typescript.test.ts`.
- No test for the export-wrapper case (the #1 blind spot — the suite uses bare `function add(...)`, not `export function`).
- No test for Go/Ruby directive mis-attribution, block-comment file-header mis-attribution, PHP Property fallthrough, or Dart/Swift behavioral extraction (registration-only).

### CI status
Now **green**: `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`. All gates pass — quality (format/lint/typecheck/typecheck-web), Analyze (python / js-ts), tests (ubuntu/macos/windows + coverage + tree-sitter ABI + packaged-install smoke), CodeQL, gitleaks, Build&Push (gitnexus + web), benchmarks (e2e chromium skipped — no web changes). Findings #1–#4 are not caught by CI (the test suite is green precisely because it doesn't exercise the export-wrapper / template / cross-label / file-header cases). Note: the PR head advanced during this review (`ce0e8ab7` → `4ba5e05`); the new commits only merge in unrelated main work (Spring routes, #2281) — none of this PR's files (ast-helpers.ts, the 11 providers, the 4 test files) changed, so every finding and anchor above still holds.

### Coverage note
This is an automated multi-tool digest (Claude swarm+ce, Codex swarm+ce+adversarial, + a coordinator reproduction pass). Cross-engine corroboration and reproductions are noted per finding; single-engine items are labeled so the human can weight them. **Verify findings before acting.** The single highest-value action is fixing finding #1 (export-wrapper JSDoc drop) — without it the headline benefit does not materialize for most exported TS/JS symbols.

gitnexus/src/core/ingestion/utils/ast-helpers.ts
```diff
@@ -926,6 +926,138 @@ export function findChild(node: SyntaxNode, type: string): SyntaxNode | null {
...
+ */
+export function extractLeadingDocComment(
+  node: SyntaxNode,
+  lineCommentPrefixes: readonly string[] = DEFAULT_LINE_DOC_PREFIXES,
+): string | undefined {
+  const prev = node.previousNamedSibling;
+  if (!prev) return undefined;
+
+  // Block doc comment: /** ... */ or /*! ... */
+  if (prev.text.startsWith('/**') || prev.text.startsWith('/*!')) {
```
magyargergo: **[P2]** Block-comment branch has no row-adjacency check (unlike the line-comment branch at 997-1004). A file-top `/**`/`/*!` license/overview block with no intervening package/import sibling above the first declaration is absorbed as that symbol's description — then flows into embedding text, polluting semantic search. Reproduced for TS (class-first file), package-less Java, and C/C++ headers; common `/* */` license style is safe (startsWith('/**') is false). Author-acknowledged in the JSDoc, untested. Fix: require prev.endPosition.row >= node.startPosition.row - 1 for block comments too, or suppress SPDX/license/@file markers; add a negative-assertion test.

_(pr-tri-review [code-read])_

PR #1743, opened 2026-05-21
fix(group): attribute Spring interface routes to controllers

JaysonAlbert:
## Background

This refreshes closed, unmerged PR #719 for open issue #714 against current `main`. Current `main` has the newer `http-patterns/java.ts` plugin architecture, but annotated Java interfaces can still be treated as provider routes and concrete controllers do not inherit interface-declared Spring mappings.

## Changes

- Add optional project-level HTTP scan hooks for language plugins.
- Teach the Java HTTP plugin to skip interface-only Spring provider emissions.
- Attribute inherited interface-declared Spring class/method mappings to concrete `@RestController` / `@Controller` implementations.
- Preserve empty-string mapping normalization behavior.
- Add focused regressions for interface non-emission and controller attribution.

## Validation

- `cd gitnexus && npm test -- test/unit/group/http-route-extractor.test.ts test/unit/group/http-route-multi-verb.test.ts`
- `cd gitnexus && npx tsc --noEmit`
- `cd gitnexus && npm run format:check`
- `node <built-cli>/dist/cli/index.js analyze --force --index-only --name spring-provider-attribution-refresh .`
- `node <built-cli>/dist/cli/index.js detect-changes --repo spring-provider-attribution-refresh` (medium risk; expected group HTTP provider extraction symbols)

## Notes

The old regex-era patch was not cherry-picked directly; this version is adapted to current `http-patterns` / `HttpRouteExtractor` architecture.

PR #2290, opened 2026-06-24
fix(ingestion/routes): resolve Spring interface-inherited routes (#2288)

### LINKED ISSUE #2288: Spring route extraction: ingestion does not resolve interface-inherited mappings (#2280)
Subtask of #2280.

## Gap

The ingestion-layer Spring extractor (`route-extractors/spring.ts`) is per-file and has no inheritance pass, so a mapping declared on an interface and inherited by the implementing controller is not resolved the way the group layer resolves it (the group layer cross-joins across files via `scanSpringInheritanceProject`).

## Example

```java
public interface OrderApi {
  @GetMapping("/orders") Object list();
}

@RestController
public class OrderController implements OrderApi {
  public Object list() { return null; }
}
```

Group recognises `GET /orders`; ingestion does not emit a Route node for it.

## Why it matters

This is one of the gaps keeping Java at `routeCoverage: 'partial'`. It must be closed (along with the multi-verb Route identity subtask) before Java can flip to `'complete'`.

## Sketch

A cross-file post-processing pass that maps interface-declared mappings onto implementing classes, mirroring the group layer's `scanSpringInheritanceProject` semantics — kept separate per #2078, with a parity test to prevent drift.

henry201605:
## What

The ingestion route extractor now resolves Spring **interface-inherited** routes: a concrete `@RestController`/`@Controller` that implements an interface declaring `@*Mapping`s gets `Route` nodes for those paths, attributed to the **controller** (handler = the controller method), with both the interface's and the controller's class prefixes joined.

Previously ingestion emitted an interface method's mapping as a standalone, prefix-less route on the interface file, and never linked it to the implementing controller. The group layer already resolved this (#1743); this brings the graph side to parity.

## Why

This is subtask #2288 of #2280. It's one of the three gaps keeping Java at `routeCoverage: 'partial'`. (`routeCoverage` is unchanged here — the flip to `'complete'` still waits on the remaining subtask, multi-verb Route identity.)

## How

Per #2078, the two Spring extractors stay separate but must not drift, so this shares the algorithm rather than duplicating it:

- **Shared primitive.** Lifted the language-agnostic inheritance algorithm (`resolveInheritedSpringRoutes` + `SharedSpringType` + the path-join helpers) down into `ingestion/route-extractors/spring-shared.ts` (the lower module the group plugins already import). The group `scanSpringInheritanceProject` is now a thin adapter over it, and `spring-consumer-shared.ts` re-exports the moved symbols so `java.ts`/`kotlin.ts` imports are unchanged. Group provider contracts are byte-identical (group suite green).
- **Ingestion collection + cross-file pass.** New `extractSpringTypes` builds the per-file `SharedSpringType` view; a new optional `extractRouteInheritanceTypes` language hook (wired for Java) ships it across the worker boundary. The parse phase aggregates project-wide and runs `resolveInheritedSpringRoutes` before the routes phase, appending inherited routes to `allDecoratorRoutes` — mirroring the existing FastAPI `include_router` cross-file pass.
- **Suppress the wrong route.** `extractSpringRoutes` no longer emits an interface-declared method route (`findEnclosingType`); the inheritance pass owns it.
- FQN annotation names (`@org.springframework...GetMapping`) are normalised to their trailing segment, matching the group layer.

## Testing

- New unit suite (`spring-interface-inheritance.test.ts`): behaviour + a **parity set asserting the ingestion cross-file pass and the group `JAVA_HTTP_PLUGIN.scanProject` yield identical inherited routes** — plain, interface+controller prefixes with an array mapping, FQN names, ambiguous duplicate-interface drop, multiple interfaces, non-controller (no inherit), and own+inherited mix.
- New integration test (`spring-interface-inheritance-pipeline.test.ts`): full pipeline emits the controller-attributed `Route` nodes with prefixes joined and no standalone unprefixed interface route.
- `npx tsc --noEmit` clean; group + route suites green; prettier / eslint clean.

Part of #2280.


gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -393,6 +394,15 @@ export interface ParseWorkerResult {
...
   routerImports: ExtractedRouterImport[];
+  /**
+   * Optional. Project-wide `SharedSpringType` view of route-defining
+   * class/interface declarations, produced by the provider's
+   * `extractRouteInheritanceTypes` hook (Java/Spring). parse-impl aggregates
+   * these and runs a cross-file pass that resolves interface-inherited routes
+   * into additional `decoratorRoutes` (#2288). Optional for cache backward
+   * compatibility; consumers must guard with `?? []`.
+   */
+  springTypes?: SharedSpringType[];
```
magyargergo, line 405: **[P2] [code-read]** Cache-version not bumped for this worker-output shape change (cross-engine: Codex adversarial HIGH + Claude ce-adversarial/ce-testing/api-contract/risk-architect). This PR adds `springTypes?` to `ParseWorkerResult` here AND changes `decoratorRoutes` semantics (interface routes suppressed at extraction; inherited routes appended in parse-impl), but `gitnexus/src/storage/parse-cache.ts:58` stays `SCHEMA_BUMP = 7` — last bumped for the analogous #2138 Part 2 (`handlerName`). Package version is `1.6.8` at both base and HEAD, so `PARSE_CACHE_VERSION = "7+1.6.8"` is unchanged. `slimParseWorkerResultsForCache` preserves `springTypes` (post-PR caches fine), but a warm chunk cache written PRE-PR (no `springTypes`) is accepted under the unchanged version: on replay `mergeChunkResults` sees `springTypes===undefined` → `allSpringTypes` empty → inheritance pass emits nothing, AND the stale pre-PR wrong unprefixed interface `decoratorRoutes` replay verbatim. Worst case is a partial-cache cascade (interface chunk cache-HIT + controller chunk cache-MISS) producing a graph neither version yields alone. Safe on the published `npm i -g` path (a release bumps the version); bites dev-from-source / CI same-version caches / MCP hot-reload. Fix: bump `SCHEMA_BUMP` 7→8 at parse-cache.ts:58 with a comment citing #2288 (invalidates the durable parsedfile-cache in lockstep via the shared PARSE_CACHE_VERSION). Per the repo's own #2138 precedent this is a should-fix before merge. Optionally add a warm-cache regression with a pre-PR-shaped shard lacking `springTypes`.

magyargergo (commented):
## Verdict: production-ready with minor follow-ups — one should-fix (cross-engine, reproduced) is a one-line parse-cache version bump; everything else is body notes and pre-existing shared follow-ups. CI is fully green; PR is **MERGEABLE**.

**Engines/legs.** Claude ran swarm (risk-architect, test-ci-verifier, security-boundary, docs-dod) + ce (correctness, adversarial, testing, api-contract, maintainability, project-standards, performance) — all live. Codex (gpt-5.5/xhigh) ran swarm (B1, live) + adversarial (B3, live, found the cache HIGH); its ce leg (B2) **failed mid-run** ("stream disconnected", no findings). So 2 of 3 Codex legs are live — cross-engine corroboration is valid; we do **not** claim three Codex confirmations.

**Merge state.** `MERGEABLE`; the earlier `BLOCKED` reflected only pending checks, now all green. Branch hygiene is clean — the two merge-from-main commits are harmless and merge-safe, and `git diff BASE..HEAD` is exactly the 11 expected Spring/ingestion/test files (no unrelated churn, no CHANGELOG edit).

---

### What's solid (credit)

- **Faithful algorithm move.** `scanSpringInheritanceProject` → `resolveInheritedSpringRoutes` in `spring-shared.ts` is byte-equivalent; the group layer thins to a thin `HttpDetection` adapter and re-exports `joinPath` / `joinInheritedSpringPath` / `SharedSpringType`. Ingestion reuses the *same* resolver for genuine parity, pinned by a set-equality parity test (`ingestionInheritedKeys == groupInheritedKeys`). (ce-correctness ran the resolver via tsx; api-contract confirmed the re-export surface is type-identical and non-breaking.)
- **Correct suppression trade.** Interface `@*Mapping`s are suppressed at extraction (`extractSpringRoutes`) and re-emitted on the implementing controller — trading a *wrong unprefixed* interface-owned route for a *correct prefixed* controller route (or, when no controller exists, a conservative miss). Exactly one emission per inherited route.
- **`lineNumber:0` handler resolution proven.** `resolveRouteHandlerSymbols` resolves by `handlerName` + `filePath` (not line); the integration test confirms the controller method resolves with `lineNumber:0`.
- **RFC#909 clean.** The cross-file pass in `parse-impl.ts` iterates `allSpringTypes` with no `if(language==='java')`; the hook lives on `LanguageProviderConfig` (carries no language name), only `javaProvider` populates it, grammar logic stays in `route-extractors/spring.ts` / `spring-shared.ts`.
- **Hygiene.** CHANGELOG untouched; strict typing (no `any`); security clean (ReDoS/OOM/recursion/Unicode all refuted); performance clean (query compiled once, linear walkers, non-Java repos pay ~0).
- **CI fully green** (verified live): Analyze (js-ts + python), Build & Push (gitnexus + web), CI Gate, CodeQL, gitleaks, quality/{format,lint,typecheck,typecheck-web}, tests/{ubuntu coverage, macos, windows, ABI×3, packaged install smoke ×2, benchmarks}, autofix, review. `Vercel` is a known auth-infra non-issue; `e2e (chromium)` correctly skips (no web changes). 12/12 new unit tests pass locally.

---

### Headline finding — parse-cache `SCHEMA_BUMP` not bumped (P2, **should-fix before merge**)

**Confidence: STRONG (cross-engine, reproduced).** Codex B3 adversarial (HIGH) + Claude ce-adversarial (P1, reproduced) + ce-testing (P1) + ce-api-contract (P2) + swarm risk-architect (P2) independently reached this. Five lanes / two engines.

- **Evidence (worktree-confirmed):** `gitnexus/src/storage/parse-cache.ts:58` is `const SCHEMA_BUMP = 7;` — last bumped for the analogous "#2138 Part 2: `ExtractedDecoratorRoute` gained `handlerName`". This PR adds `springTypes?` to `ParseWorkerResult` (`parse-worker.ts:405`) **and** changes `decoratorRoutes` semantics (interface routes suppressed at extraction; inherited routes appended in `parse-impl`), but does not bump. `gitnexus/package.json` version is `1.6.8` at **both** base and HEAD → `PARSE_CACHE_VERSION = "7+1.6.8"` is identical pre- and post-PR.
- **`slimParseWorkerResultsForCache` preserves `springTypes`** (spread), so caches *written post-PR* are fine. The bug is **cross-version**: a warm chunk cache written **pre-PR** (no `springTypes`) is accepted under the unchanged version. On replay, `mergeChunkResults` sees `result.springTypes === undefined` → `allSpringTypes` stays empty → the inheritance pass emits **nothing**, while the stale pre-PR **wrong unprefixed** interface `decoratorRoutes` replay verbatim. The PR silently no-ops *and* re-emits the old wrong route, with zero diagnostic.
- **Worst case — partial-cache cascade [reproduced reasoning]:** controller `C.java` changed (cache MISS → re-parsed → emits a `SharedSpringType` for `C`) but interface `Api.java` unchanged (cache HIT, pre-PR shard, no `springTypes`). `interfaceRoutes.get("Api")` → `undefined` → `C` inherits nothing, **and** `Api.java` replays its stale wrong route — a coherence artifact neither version produces alone.
- **Why P2, not P0/P1:** the published `npm i -g gitnexus@latest` upgrade path is safe (a release bumps `package.json` → version string changes → cache invalidates). It bites **dev-from-source / CI same-version caches / MCP hot-reload** — exactly the populations that re-analyze the same repo repeatedly on a pinned version. It is the repo's own #2138 convention that makes this a should-fix.
- **Fix (one line):** bump `SCHEMA_BUMP` 7 → 8 with a comment citing #2288 (new `springTypes` field + changed `decoratorRoutes` suppression semantics). This invalidates stale shards in lockstep with the durable `parsedfile-cache` (shared `PARSE_CACHE_VERSION`). A maintainer should confirm the one-time cold-reparse cost — the same call #2138 made. Optionally add a warm-cache regression (pre-PR-shaped shard lacking `springTypes`).

> Inline anchor note: `parse-cache.ts` is **not** in the diff, so the inline comment is anchored on the in-diff `springTypes?` field declaration at `parse-worker.ts:405` and references `parse-cache.ts:58` in the body.

---

### Secondary (body)

**P3 — PR-introduced doc-drift (recommend fixing in this PR).** Group `java.ts:684` `routeCoverage` comment still lists "(2) interface-inherited Spring routes" as **not** emitted by ingestion — now false (the PR makes ingestion emit them); the follow-up note at lines 688-689 also still lists "interface-inheritance emission" as pending in #2280. The `routeCoverage:'partial'` setting itself stays **correct** (reasons (1) class-array prefix + (3) per-verb identity remain). Caught by Codex B1 + ce-maintainability + risk-architect. Unchanged file → body, not inline. Update lines 682-689 to drop reason (2).

**Pre-existing shared limitations — NOT #2290 regressions.** These were faithfully moved from the already-shipped group inheritance layer; the group path exhibits each identically on the same fixtures. The PR's actual target — **direct** `class implements Interface` inheritance — is correct and well-tested. Recommend follow-up issue(s) + documenting tests; none block merge.
- *Transitive `interface B extends A`* unhandled — controller `implements B (extends A)` misses `A`'s mappings; the collector hard-sets `implementedInterfaces: []` for `kind==='interface'` and the resolver only follows one hop. Cross-engine (Codex B1+B3, ce-adversarial, ce-testing, risk-architect); consensus P3 (a couple flagged P2). Pre-PR this leaked a wrong unprefixed route; now suppressed → no route. The single genuinely-new wrinkle is suppression-without-re-emit in the transitive case, which is net-neutral-to-better under `routeCoverage:'partial'`.
- *Method-name overload* — `methodMap` keyed by method **name** drops all-but-last overload (and the per-controller-method `seen` dedup then duplicates the survivor). `[reproduced]` by ce-adversarial / ce-correctness via direct `resolveInheritedSpringRoutes` call. Shared; pathological in Spring (routes by URL, one `@Override foo` by name). P3.
- *`joinInheritedSpringPath` prefix doubling* — interface `/api` + controller `/v1` → `/v1/api/list`; real Spring `findMergedAnnotation` gives `/v1/list` (controller's own `@RequestMapping` wins). `[reproduced]`. The dominant pattern (controller has no own prefix) is correct, and the same-prefix case is guarded. Shared/pre-existing; the unit test *encodes* `/v1/api/list` as expected. P3 — follow-up to verify vs Spring.
- *Bare path-less `@GetMapping` on an interface method* → `annotationRoutePaths` returns `[]` → route dropped. Shared. P3.
- *`@FeignClient` interface non-discrimination* — **RESOLVED to parity (no divergence).** Several lanes initially split, but coordinator code-read is definitive: the group **inheritance feeder** `collectSpringTypes` → `scanRouteAnnotations().methodRoutes` pushes every method `@*Mapping` **unconditionally** (no `@FeignClient` check). The Feign consumer/provider discrimination lives only in the separate per-file `scan()` Strategy-B path (`java.ts:725`), which is **not** the inheritance feeder. So ingestion treats a `@FeignClient` interface identically to the group inheritance path → parity, not a regression. Impact is negligible (a `@RestController implements @FeignClient` is a non-pattern). P3 follow-up: add a Feign fixture to **pin** the shared behavior.

**Maintainability debt (P2/P3).** The *algorithm* is unified, but `extractSpringTypes` + 5 private helpers (`declarationAnnotations`, `annotationName`, `annotationRoutePaths`, `implementedInterfaceNames`, `directMethods`) in `spring.ts` are near-verbatim duplicates of the group's `collectSpringTypes` + helpers in `java.ts`. Drift between the two collectors is guarded **only** by a one-fixture-set parity test (no Feign/overload/transitive cases). Recommend lifting the five helpers into `spring-shared.ts` and unifying the collectors, or broadening the parity fixtures. (ce-maintainability F1; also notes `InheritedSpringRoute` is an exported-but-unused type — make it module-local.)

**Test-strength nits (body).**
- Parity test `spring-interface-inheritance.test.ts:220` asserts `ingestion == group` without pinning the expected set → **vacuous if both return empty**. This is the array-mapping fixture (prior root of #2265) — the highest-value case to pin. Add `expect(groupInheritedKeys(files)).toEqual(new Set([...]))`.
- Integration test uses `toContain` / `not.toContain` (no upper bound) rather than set-equality → a spurious extra Route node would pass undetected.
- `result-merge.ts` `springTypes` branch has no unit coverage (a `src.springTypes = undefined` case would also have surfaced the shape-change).
- No warm-cache regression test (the exact gap that let the missing `SCHEMA_BUMP` through).

---

### Refuted suspicions (validation is a feature)

Each of these was raised by at least one lane and disproven with evidence — worth crediting:
- **Double-emission within ingestion** — interface route is suppressed *before* the inheritance pass appends → exactly one Route node. Refuted.
- **Double-emission ingestion-vs-group** — separate subsystems (ingestion `decoratorRoutes`→Route nodes at parse time; group `HttpDetection`→post-analyze). Different node types/stores. Refuted.
- **`lineNumber:0` breaks handler resolution** — resolved by name + filePath; integration test proves it. Refuted.
- **Worker-cache backward-compat CRASH** — `springTypes` is optional and every path is `?.`/`if`/`?? []`-guarded; a stale shard can't crash. (Distinct from the cache **staleness**, which is the real headline finding.) Refuted.
- **ReDoS / OOM / unbounded recursion** — `joinPath`/`joinInheritedSpringPath` regexes are anchored & linear; recursion is tree-sitter-bounded; algorithm is a direct lift of pre-existing group code. Refuted.
- **`decoratorName:'inherited-mapping'` breaks a consumer switch** — it's a free `string`, no exhaustive switch; the HTTP verb is carried separately on the Route node (`graphMethod` wins over `methodFromRouteReason`). Refuted.
- **FastAPI prefix expansion hitting Spring routes** — gated to `.py` + `router` receiver; Spring is `.java` with no receiver, and ordering is correct. Refuted.

---

### CI & coverage

CI is **fully green** (all checks above). The new integration test cannot run in the worktree (no built `parse-worker.js`) but **does** run in CI (setup builds before tests) and is in the default vitest project glob. Coverage is strong on the happy/direct-inheritance path; the gaps above (warm-cache, transitive/overload/Feign parity, `result-merge` `springTypes`) are body recommendations, not blockers.

---

*This is an automated multi-tool digest (Claude swarm + ce personas; Codex gpt-5.5/xhigh swarm + adversarial; the Codex ce leg failed). Verify the cache finding against your release/cache policy before acting on the one-line bump.*

koriyoshi2041:
Did a fresh-build local pass on `0ef89edb`:\n\n- `npm run build` passed\n- `npm test -- --run test/unit/spring-interface-inheritance.test.ts`: 12 passed\n- `npm test -- --run test/integration/route-parse-skip.test.ts -t 'interface-inherited Spring route survives on the implementing controller'`: 1 passed\n- `npm test -- --run test/integration/spring-interface-inheritance-pipeline.test.ts`: 3 passed\n- `npx tsc --noEmit` passed\n- `git diff --check origin/main...HEAD` passed\n\nSmall local gotcha: running the new pipeline integration directly before rebuilding can exercise stale `dist/` worker output and show the old bare `/list` / `/create` routes. After `npm run build`, the same test passes, so the fresh-worker functional path looks good. The parse-cache schema bump already called out above still looks like the remaining should-fix before merge.

henry201605:
Thanks for the thorough review! Addressed the should-fix plus all the in-PR recommendations:

- **parse-cache SCHEMA_BUMP (should-fix)** → bumped 7→8 with a comment noting the `springTypes` addition + the `decoratorRoutes` semantics change, so a warm cache predating this PR is invalidated instead of replaying stale routes.
- **routeCoverage doc-drift** → dropped the now-false "interface-inherited Spring routes" item from the `java.ts` comment (ingestion emits them as of this PR); `routeCoverage` stays `'partial'` for the remaining class-array + per-verb gaps.
- **Vacuous array-form parity test** → pinned the expected set on the interface+controller-prefix array case so it can't pass on two empty sets.
- **Integration test bound** → switched from `toContain` to full set-equality, so a spurious/leaked extra Route node now fails the test.
- **`result-merge` springTypes coverage** → added union + undefined-when-absent unit cases for the new merge branch.
- **`InheritedSpringRoute`** → made module-local (was exported but unused).

These are out of scope for this PR: the transitive-`extends` / overload / prefix-doubling / Feign cases are pre-existing limitations faithfully inherited from the group layer (not regressions here); the collector-helper unification and a warm-cache regression test are nice-to-haves. Happy to open a tracking issue if you'd like any of them followed up.

Rebased onto the latest main. tsc / lint / prettier clean; group + route suites green.

gitnexus/src/core/group/extractors/http-patterns/spring-consumer-shared.ts
```diff
@@ -16,6 +16,20 @@
...
+  type SharedSpringType,
+} from '../../../ingestion/route-extractors/spring-shared.js';
+
+// `joinPath` / `joinInheritedSpringPath` / `SharedSpringType` now live in the
+// lower shared layer (`ingestion/route-extractors/spring-shared.ts`) so the
+// ingestion route extractor and the group plugins share one implementation.
+// Re-exported here so this module's existing importers (java.ts / kotlin.ts)
+// keep importing them from `spring-consumer-shared.ts` unchanged.
+export { joinPath, joinInheritedSpringPath };
+export type { SharedSpringType };
```
magyargergo: Can you please make sure there aren't re-exported? Also update the caller site. 🙏

gitnexus/src/core/ingestion/parsing-processor.ts
```diff
@@ -120,6 +124,7 @@ export const mergeChunkResults = (
     for (const item of result.routerIncludes ?? []) allRouterIncludes.push(item);
     for (const item of result.routerImports ?? []) allRouterImports.push(item);
     for (const item of result.routerModuleAliases ?? []) allRouterModuleAliases.push(item);
+    if (result.springTypes) for (const item of result.springTypes) allSpringTypes.push(item);
```
magyargergo: Nitpick: you can do this
```suggestion
    for (const item of result.springTypes ?? []) allSpringTypes.push(item);
```

gitnexus/src/core/ingestion/route-extractors/spring-shared.ts
```diff
@@ -86,3 +104,156 @@ export function unquoteSpringLiteral(raw: string): string | null {
...
+ * its routes are dropped (the `null` marker). The controller's own class
+ * prefix(es) cross-product the inherited routes; `joinInheritedSpringPath`
+ * avoids doubling a prefix the interface already baked in (#2057). Duplicate
+ * `(method, path)` results per controller method are de-duped.
+ */
+export function resolveInheritedSpringRoutes(types: SharedSpringType[]): InheritedSpringRoute[] {
+  // interface name → (method name → routes). `ownerPrefix` records the
+  // interface's own class prefix so the controller side avoids doubling it
+  // (#2057). `null` marks an ambiguous (duplicated) interface name.
+  type IntermediateRoute = { method: string; path: string; ownerPrefix: string };
```
magyargergo: Move this out put the function please

gitnexus/src/core/ingestion/route-extractors/spring-shared.ts
```diff
@@ -86,3 +104,156 @@ export function unquoteSpringLiteral(raw: string): string | null {
...
+ * by the ingestion route extractor and the group Java/Kotlin plugins so all
+ * three emit the same inherited routes.
+ *
+ * An interface name that resolves to two distinct interfaces is ambiguous and
+ * its routes are dropped (the `null` marker). The controller's own class
+ * prefix(es) cross-product the inherited routes; `joinInheritedSpringPath`
+ * avoids doubling a prefix the interface already baked in (#2057). Duplicate
+ * `(method, path)` results per controller method are de-duped.
+ */
+export function resolveInheritedSpringRoutes(types: SharedSpringType[]): InheritedSpringRoute[] {
```
magyargergo, line 204: We need to benchmark this solution

magyargergo (changes requested):
Can you please benchmark your solution as well end-to-end? 🙏

henry201605:
Thanks! Addressed the round-2 review:

- **Benchmark (end-to-end)** → added `test/integration/spring-inheritance-benchmark.test.ts` with three suites:
  - an ungated O(n²) tripwire that runs in normal CI (400 interface→controller pairs resolve in a few ms; 5s budget);
  - a gated isolated-resolver scaling suite — 250→500→1000 pairs, ~1.0x time_ratio/size_ratio;
  - a gated **end-to-end** suite that runs the full `runPipelineFromRepo` on synthetic Spring repos (100/250/500 pairs): 738ms → 1111ms → 2440ms, i.e. 0.60x / 1.10x — linear.
- **Re-exports** → dropped from `spring-consumer-shared.ts`; `java.ts` / `kotlin.ts` now import `joinPath` + `SharedSpringType` directly from `spring-shared.ts`.
- **`IntermediateRoute`** → moved out to module scope.
- **`springTypes` aggregation** → took your `?? []` suggestion.
- **SCHEMA_BUMP** → already bumped 7→8 in the prior round (parse-cache.ts) with a #2288 comment.

tsc / lint / prettier clean; group + route suites green. The gated benchmark suites skip in normal CI (the end-to-end one needs a built worker).

koriyoshi2041:
Rechecked the latest head `cf87b85d` after the round-2 changes.

Local validation:
- `npm test -- --run test/unit/spring-interface-inheritance.test.ts test/unit/result-merge.test.ts test/integration/spring-interface-inheritance-pipeline.test.ts test/integration/spring-inheritance-benchmark.test.ts` passed: 4 files, 22 passed, 2 skipped
- `GITNEXUS_BENCH=1 npx vitest run test/integration/spring-inheritance-benchmark.test.ts` passed: 4 passed
- `npx tsc --noEmit` passed
- `git diff --check origin/main...HEAD` passed

The changes cover the earlier blockers I called out: `SCHEMA_BUMP` is now 8 with the right cache-invalidation reason, `springTypes` merge has direct unit coverage, and the benchmark file now has both an always-on hotpath tripwire and gated scaling/end-to-end checks. The only remaining red state I see is the known Vercel team authorization context, not a code failure.

magyargergo (approved):

PR #1479, opened 2026-05-10
feat(analyze): incremental indexing (parse cache + DB writeback + scope-res short-circuit)

abhigyanpatwari:
## Summary

Incremental analyze is now the default. On a 1000-file TypeScript repo (this codebase, 24K nodes / 32K edges):

| Run | Time | vs. `--force` |
|---|---|---|
| Cold full reindex | **86s** | (was 143s; **40% faster**) |
| Warm cache, no source changes | **3s** | early-return |
| Warm cache + 1-file edit | **38s** | (was 134s; **72% faster**) |

All paths produce **byte-identical** `{nodes, edges, clusters, flows}`. Verified by running `analyze` and `analyze --force` from the same edited state and comparing graph stats — they match exactly. `--force` remains the escape hatch when you want to wipe and start over.

## How

Three independent layers compose to deliver the speedup. Each preserves correctness on its own; together they give the headline numbers.

1. **Incremental DB writeback.** Hash all source files at end of analyze, store in `meta.json.fileHashes`. Next run diffs current vs. stored: only rows for `changed ∪ added ∪ deleted` files are deleted+rewritten in LadybugDB. Unchanged-file rows are never touched. `Community`/`Process` (graph-wide) are always regenerated by their phases on the merged graph, preserving the "Leiden runs on the FULL graph" invariant. Crash recovery via the `incrementalInProgress` dirty flag in `meta.json` — set before destructive ops, cleared on success, forces a full rebuild on the next run if found set.

2. **Chunk-level parse cache.** Cache file at `.gitnexus/parse-cache.json`. Key is `sha256(sorted("filePath:fileContentHash" for each file in chunk))`; value is the raw `ParseWorkerResult[]` returned by the worker pool. Cache hit → skip worker dispatch entirely, replay via the extracted `mergeChunkResults` helper. Content-addressed, so it survives `--force`. Two stability fixes: a JSON replacer/reviver that round-trips `Map`/`Set` instances cleanly (without it, `ParsedFile.scopes[*].typeBindings` collapses to `{}` and crashes on iteration); alphabetical sort on `parseableScanned` before chunking, so filesystem ordering noise doesn't shift files between chunks.

[...]

PR #2300, opened 2026-06-25
fix(search): index description field for FTS so doc comments are keyword-searchable

### LINKED ISSUE #2299: description field not indexed by FTS —doc comments unsearchable via keyword query
Issue #2270 / PR #2286 added `descriptionExtractor` for all languages, so Javadoc/KDoc/JSDoc/Doxygen/godoc/RDoc etc.
  are now correctly extracted and stored in the `description` column of LadybugDB. However, `description` is **not
  indexed by FTS**, making doc comments still unsearchable via MCP `query` keyword search.

 ## Reproduction

  A Java class whose Javadoc describes its purpose in terms unrelated to the class name:

  ```java
  /**
   * Implements the circuit breaker pattern for distributed
   * service mesh fault tolerance. Automatically isolates
   * failing downstream dependencies to prevent cascade failures
   * and restores traffic once health checks pass.
   */
  public class RetryScheduler { ... }
  ```

  - `query("circuit breaker fault tolerance")` →❌ does NOT find `RetryScheduler`
  - `query("RetryScheduler")` →✅ finds it by name only
  - Cypher `MATCH (n:Class {name: 'RetryScheduler'}) RETURN n.description` →✅ description is correctly stored in DB

  The Javadoc keywords ("circuit breaker", "fault tolerance", "cascade failures") are the natural terms a developer
  would search for, but they appear only in `description`, not in `name` or `content`.

  ## Root cause

  `gitnexus/src/core/search/fts-schema.ts` only indexes `name` and `content` for 5 tables:

  ```typescript
  { table: 'File',      indexName: 'file_fts',      properties: ['name', 'content'] },
  { table: 'Function',  indexName: 'function_fts',   properties: ['name', 'content'] },
  { table: 'Class',     indexName: 'class_fts',      properties: ['name', 'content'] },
  { table: 'Method',    indexName: 'method_fts',     properties: ['name', 'content'] },
  { table: 'Interface', indexName: 'interface_fts',  properties: ['name', 'content'] },
  ```

  **Two problems:**

  1. **`description` is missing from all FTS index definitions.** The `description` column exists in the DDL for
  Function, Class, Interface, Method, and 19 other symbol tables (`gitnexus/src/core/lbug/schema.ts`), but it is never
  indexed for FTS. BM25 keyword search (`searchFTSFromLbug` in `bm25-index.ts`) only queries these 5 FTS indexes, so
  doc-comment text is invisible to keyword search.

  2. **FTS only covers 5 tables.** Struct, Enum, Trait, Constructor, Record, Union, Namespace, TypeAlias, Macro,
  Property, Const, Static, Variable, Delegate, Annotation, Template, Module, and Section all have a `description` column
  in the DDL but have no FTS index at all. Doc comments on Rust structs, C++ enums, Go types, etc. are also
  unsearchable by keyword.

  ## Proposed fix

  1. Add `description` to the `properties` array for the existing 5 FTS indexes.
  2. Add FTS indexes for additional symbol tables that carry a `description` column and are also in `EMBEDDABLE_LABELS`
  (to ensure descriptions are both keyword-searchable and semantically searchable).

  ## Impact

  **Without this fix, the `descriptionExtractor` added by #2286 stores doc-comment text in the database but it is
  invisible to keyword search (BM25).** The `description` field is only consumed by the embedding pipeline (truncated to
  150 chars in the metadata header, configurable via `EmbeddingConfig.maxDescriptionLength`), so semantic search may
  partially benefit. For languages where doc comments sit outside the symbol's AST range (Java Javadoc, Rust `///`, Go
  godoc, etc.), the `content` snippet (`startLine - 2` to `endLine + 2`) captures at most 2 lines of the preceding
  comment —typically just the closing `*/` —so the bulk of the doc text is absent from both `content` and the
  embedding body. **Keyword search —the primary way users find code by doc-comment terms —remains broken for all
  languages.**

magyargergo:
## Why

`descriptionExtractor` (#2286 / #2270) populates the `description` column with
doc-comment text — Javadoc, KDoc, JSDoc, Doxygen, godoc, RDoc — for every symbol
table. But that column was **never fed to full-text search**, so the natural terms
a developer searches for were invisible to keyword (`query()`) search:

```java
/** Implements the circuit breaker pattern for distributed service mesh
 *  fault tolerance ... prevents cascade failures ... */
public class RetryScheduler { ... }
```

- `query("circuit breaker fault tolerance")` → ❌ did not find `RetryScheduler`
- `query("RetryScheduler")` → ✅ found by name only

For languages where doc comments sit outside the symbol's AST range (Java Javadoc,
Rust `///`, Go godoc), the text is also largely absent from `content` and the
embedding body — so **keyword search was the path that had to carry it, and it was
broken for every language.**

Closes #2299.

## Root cause

`gitnexus/src/core/search/fts-schema.ts` indexed only `name` + `content` on **5**
tables (File, Function, Class, Method, Interface). `description` was indexed
nowhere, and 15+ other symbol tables that carry a `description` column (Struct,
Enum, Trait, Const, …) had no FTS index at all.

## What changed

`FTS_INDEXES` is the single source of truth driving index creation
(`createSearchFTSIndexes`), querying (`searchFTSFromLbug`), and verification
(`verifySearchFTSIndexes`), so the fix is localized to the schema array plus the
index-build lifecycle:

- **Index `description`** on Function/Class/Method/Interface. `File` is left at
  `name`+`content` — it has no `description` column, and adding it would make
  `CREATE_FTS_INDEX` throw.
- **Extend FTS coverage 5 → 20 tables**, covering every `EMBEDDABLE_LABELS`
  symbol table (Constructor, Struct, Enum, Trait, Impl, Macro, Namespace,
  TypeAlias, Typedef, Const, Property, Record, Union, Static, Variable), each with
  `['name', 'content', 'description']`.
- **Drop-then-create** in `createSearchFTSIndexes`. `createFTSIndex` is
  idempotent-by-name and skips an existing index, so without this the four
  upgraded indexes would stay silently stale on incremental re-analyze and
  `--repair-fts` (the most important tables would never get the fix). Dropping
  first guarantees the live property set wins on every analyze.
- **`--repair-fts` FTS-availability guard.** Because `createSearchFTSIndexes` now
  drops before creating, the repair path gained an explicit `loadFTSExtension`
  gate (mirroring the analyze path) so an unavailable extension fails loudly
  *before* any index is dropped — it can no longer leave the DB index-less.

## Re-index required

Existing indexed repos must re-run `gitnexus analyze` (or `gitnexus analyze
--repair-fts`) before description-based keyword search works. The drop-then-create
change ensures that re-run actually upgrades the schema rather than skipping the
already-present indexes.

## Testing

- `fts-schema.test.ts` — structural guards: `File` stays `name`+`content`; the
  four originals include `description`; **every indexed property is a real column
  on its table** (derived from the schema DDL — prevents the `File`-style trap for
  all 20 tables); every `EMBEDDABLE_LABELS` table is covered; index names unique/valid.
- `fts-indexes.test.ts` — drop precedes create for every entry; callback parity.
- `fts-description-search.test.ts` (integration, real LadybugDB) — a Java class
  and a Rust struct are found by keywords present **only** in `description`; name
  search still works; and a second scenario seeds an old `name`+`content`-only DB,
  runs the real `createSearchFTSIndexes()`, and proves drop-then-create **upgrades
  a live stale index** so description search starts working.
- `bm25-search.test.ts` assertions re-derived from `FTS_INDEXES` so they stay
  correct as coverage evolves.

All FTS unit + integration suites pass; `tsc` clean.

## Residual Review Findings

Non-blocking items surfaced during review and deferred by design:

- **P2** `gitnexus/src/core/search/bm25-index.ts:86` — `searchFTSFromLbug` now
  fans out 20 sequential FTS round trips per query (incl. empty tables for
  language-homogeneous repos). Accepted tradeoff: the loop is deliberately
  sequential to avoid pool connection contention. Future optimization: cache the
  per-repo set of non-empty tables and skip the rest.
- **P2** `gitnexus/src/core/search/fts-schema.ts:22` — `FTS_INDEXES` is a
  hand-maintained literal list parallel to `EMBEDDABLE_LABELS`. Drift is guarded
  at test time by the "covers every embeddable symbol table" assertion; could be
  derived by construction with a snake_case helper if preferred.
- **P3** `gitnexus/src/core/search/fts-indexes.ts:21` — the `// ponytail:`
  rebuild-every-analyze comment should reference a tracking issue for the
  schema-fingerprint gate that would skip unchanged rebuilds.

https://claude.ai/code/session_01CXkGHj5kQUAcDXRa6BMAbZ

## Test infrastructure

- The new `fts-description-search.test.ts` runs in the sequential `lbug-db`
  vitest project (like every other FTS-index-creating integration test), not
  the parallel `default` project — avoids LadybugDB native mmap file-lock
  conflicts in parallel forks (Windows).
- `withTestLbugDB` honors `GITNEXUS_REQUIRE_FTS=1`: when the FTS extension is
  unavailable it now fails loudly instead of silently `ctx.skip()`-ing. The CI
  test jobs (ubuntu coverage + windows/macOS cross-platform) set the flag, so
  FTS-dependent lbug integration suites can never silently vanish while CI stays
  green; local/offline runs leave it unset and keep skipping gracefully.


magyargergo (commented):
## Tri-review digest — PR #2300 `fix(search): index description field for FTS so doc comments are keyword-searchable`

> ⚠️ **Degraded run — single engine.** Engine A (Claude swarm + ce, 8 lanes) was **unavailable** this run (weekly limit). Only **Engine B (Codex GPT‑5.5/xhigh)** produced findings — swarm (B1), ce report-only (B2), adversarial (B3). The Claude synthesis + critic gate could not run, so **every finding below is single‑engine, then independently re‑verified by the coordinator against the worktree.** No cross‑engine corroboration was possible — weight accordingly.

### Verdict
No merge‑blocking issues found. CI is fully green (31 checks); branch is `CLEAN`/`MERGEABLE`. This is a focused, well‑documented, well‑tested change. The one finding worth acting on is a **CI coverage gap**, not a correctness defect.

### What's solid (verified)
- **Schema linchpin holds.** All 20 `FTS_INDEXES` entries map to real columns in `core/lbug/schema.ts`: `File` correctly stays `name+content` (no `description` column); the other 19 == `EMBEDDABLE_LABELS` exactly and each carries `name`+`content`+`description` (explicit for Function/Class/Interface/Method/Property; `CODE_ELEMENT_BASE` for the rest). `fts-schema.test.ts` guards this going forward.
- **drop‑then‑create is the right fix.** `createFTSIndex` is idempotent‑by‑name, so without the drop a schema change (adding `description`) would never reach an existing persisted `.lbug` DB. `dropFTSIndex` no‑ops when absent and clears the per‑connection memo.
- **`--repair-fts` gate is correctly placed** — `loadFTSExtension` is checked *before* any drop (`run-analyze.ts:665`), so an unavailable extension can't leave the DB index‑less. (Self‑caught P1 from development; confirmed fixed.)
- **Strong, non‑vacuous tests**, including a real **upgrade‑path** test: `fts-description-search.test.ts` seeds an old `name+content`‑only schema (no Struct index), runs the *real* `createSearchFTSIndexes()`, and asserts description‑only keywords become searchable — directly exercising drop‑then‑create. Search terms live **only** in `description`, so the test fails if description isn't indexed.
- CHANGELOG untouched (release‑ownership rule honored), no bidi/non‑ASCII, no security regressions, parameter‑bound FTS query text.

### Findings (single‑engine Codex, coordinator‑verified)

**1 — [P2 · coverage gap · confirmed] New FTS test never runs on macOS/Windows** — `gitnexus/scripts/cross-platform-tests.ts:54-77` `[code-read]`
The new `test/integration/fts-description-search.test.ts` is registered in vitest's sequential `lbug-db` project (so it runs in the ubuntu/coverage full suite), but it is **absent from `LBUG_NATIVE` / `ALL_CROSS_PLATFORM`**, which is the explicit allowlist the macOS/Windows "platform‑sensitive" jobs run via `run-cross-platform.ts`. So the new 20‑index/`description` path is **validated only on Linux**. This is doubly notable because the PR adds `GITNEXUS_REQUIRE_FTS=1` to those very cross‑platform jobs *specifically so FTS suites can't silently skip there* — yet the new FTS suite isn't in the subset those jobs execute. The hardening protects the old FTS fixtures (`search-core`, `search-pool`, which are in `LBUG_NATIVE`) but not the new code.
*Not merge‑blocking* (feature is validated on Linux). **Fix:** add `'test/integration/fts-description-search.test.ts'` to `LBUG_NATIVE`.

**2 — [P3 · robustness] drop‑then‑create loop is non‑atomic** — `gitnexus/src/core/search/fts-indexes.ts:25` `[code-read]`
Codex‑adversarial escalated this to "high/no‑ship" on a *legacy table missing the `description` column* trigger. **That destructive framing is refuted:** the analyze path wipes any pre‑`description` DB before recreating (`schemaVersion !== 4` → full rebuild → fresh tables), and every incremental‑eligible (v4) DB already has `description` on all these tables — the column (added 2026‑02‑25) *predates* `INCREMENTAL_SCHEMA_VERSION` itself (#1479, 2026‑05‑12). So "drop a working index, then fail to recreate because the column is missing" is not reachable in practice.
**Residual valid point:** the loop has no atomicity/rollback. If a `CREATE` ever fails mid‑loop (transient LadybugDB error, disk full, or — the real future risk — a new table/column added to `FTS_INDEXES` without a `schemaVersion` bump), FTS is left partially rebuilt. **Mitigated:** `verifySearchFTSIndexes` reports the missing indexes and the caller throws a clear "run `gitnexus analyze --force`" message → loud + recoverable, not silent corruption. The author already documented the full‑rebuild cost trade‑off in a `ponytail:` comment. Acceptable as‑is; a failure‑injection test would close it.

**3 — [P3 · nit] `verifySearchFTSIndexes` checks queryability, not column coverage** — `gitnexus/src/core/search/fts-indexes.ts:42-55` `[code-read]`
It probes each index with `QUERY_FTS_INDEX` and catches errors — it confirms the index is *queryable*, not that it covers `description`. So if a stale `name+content` index ever survived (it can't on the paths above, but defensively), verify would pass green while description search silently failed. Optional hardening: probe a description‑only term or assert indexed properties post‑rebuild. Low priority.

**4 — [P3 · perf] Query latency: 5→20 sequential FTS probes per search** — `gitnexus/src/core/search/bm25-index.ts:86,95` `[code-read]`
`searchFTSFromLbug` loops over all `FTS_INDEXES` **sequentially** (the comment notes FTS must run serially to avoid connection contention), so every keyword search now issues 20 round‑trips instead of 5. Unmeasured on large repos; likely acceptable but it's per‑search and 4×. Consider a benchmark or skipping empty/absent‑index tables if latency shows up.

**5 — [P3 · minor] End‑to‑end coverage stops at `searchFTSFromLbug`** — the integration test exercises the BM25 read helper but not the public `query()` / `LocalBackend.query` surface named in the PR body. Fair, but `searchFTSFromLbug` *is* the FTS read path; the public query only layers ranking/merge. Optional one extra assertion.

**6 — [P3 · by‑design] macOS/Windows now hard‑fail if FTS can't load/install** — the intended trade‑off (loud > silent skip). Green on all 3 platforms today via the `auto` install policy. Future flake risk if extension packaging/network regresses; not a defect.

### CI
All 31 checks pass (ubuntu coverage, win/macOS platform‑sensitive, tree‑sitter ABI ×3, packaged‑install smoke, CodeQL, gitleaks, zizmor, quality/typecheck/lint/format). Merge state `CLEAN`.

---
*Automated digest — **single‑engine (Codex) this run; Claude unavailable**, coordinator‑verified, no independent critic pass. Verify before acting; treat confidence as lower than a normal two‑engine tri‑review.*

gitnexus/vitest.config.ts
```diff
@@ -57,6 +57,7 @@ export default defineConfig({
             'test/integration/local-backend-calltool.test.ts',
             'test/integration/search-core.test.ts',
             'test/integration/search-pool.test.ts',
+            'test/integration/fts-description-search.test.ts',
```
magyargergo, line 60: **[P2 · coverage gap · code-read] The new FTS test never runs on macOS/Windows.**

This line registers `fts-description-search.test.ts` in vitest's sequential `lbug-db` project, so it runs in the **ubuntu/coverage** full suite. But the macOS/Windows *platform-sensitive* jobs don't use this config selection — they run `npx tsx scripts/run-cross-platform.ts`, which executes only the explicit `ALL_CROSS_PLATFORM` allowlist. The new test is **absent from `LBUG_NATIVE`** in `gitnexus/scripts/cross-platform-tests.ts` (lines 54-77), so the new 20-index/`description` path is validated **only on Linux**.

This undercuts the PR's own hardening: you added `GITNEXUS_REQUIRE_FTS=1` to the win/macOS jobs *specifically so FTS suites can't silently skip there* — but those jobs only run the old FTS fixtures (`search-core`, `search-pool`), not the new suite.

**Fix:** add `'test/integration/fts-description-search.test.ts'` to `LBUG_NATIVE` in `gitnexus/scripts/cross-platform-tests.ts`.

Not merge-blocking (the feature is validated on Linux). `[code-read]`

PR #2302, opened 2026-06-25
feat(ingestion/routes): give Route nodes a (method, url) identity (#2289)

### LINKED ISSUE #2289: Route identity is URL-only: same-URL multi-verb routes collapse into one node (#2280)
Subtask of #2280.

## Gap

`Route` node ids are derived from the URL only (`generateId('Route', routeURL)`), so two handlers on the same URL with different HTTP verbs (e.g. `GET /users` and `POST /users`) collapse into a single Route node. The second verb's handler is effectively lost from the graph.

## Blast radius (why this needs design alignment first)

Changing Route identity from URL to `(method, URL)` touches several places that all key on the URL today:
- `pipeline-phases/routes.ts` (node creation, ×2 sites)
- `call-processor.ts` (FETCHES edges)
- `pipeline-phases/processes.ts` (ENTRY_POINT_OF)
- `manifest-extractor.ts` and the group `HANDLES_ROUTE` queries
- the cross-repo contract bridge (#2269), which links by contract identity

## Open question for maintainers

Before implementing, I'd like to agree on the identity scheme — e.g. a compound id `(method, path)`, how case and path variables (`/users/{id}`) should be normalised, and how it should interact with the cross-repo bridge so provider/consumer matching stays correct. Happy to draft a short proposal once there's a preferred direction.

## Why it matters

The last of the gaps keeping Java at `routeCoverage: 'partial'`.

henry201605:
## What & why

Part of #2280 (Spring route-parity gaps). Closes the third gap: multi-verb route identity.

Today a `Route` node is keyed by URL alone. When one URL carries multiple HTTP
verbs — e.g. a Spring controller with `@GetMapping("/items")` **and**
`@PostMapping("/items")`, or an Express `app.get('/api/users')` +
`app.post('/api/users')` — the two routes collapse into a **single** Route node.
The second verb's resolved handler and its execution flow are silently dropped.

This PR makes a Route node's identity `(method, url)`.

## Design

Identity is computed by a single shared primitive, `routeNodeKey(method, url)`:

- A **known, specific** verb keys as `` `${method} ${url}` `` → `GET /items`,
  `POST /items` are two distinct nodes.
- A **method-less** route falls back to **URL-only**, byte-identical to the
  pre-#2289 id. This covers filesystem routes (Next.js / Expo / PHP) and Laravel
  `resource` / `apiResource` (which expand to several verbs at runtime and are
  intentionally left method-less).
- A **wildcard `*`** route (e.g. a Django function view that handles any verb)
  also falls back to URL-only.

So only genuine declaration-style multi-verb routes split; everything else keeps
its existing node id.

The same key is shared across the **three** phases that must agree on the Route
node id (this is why the change is not confined to the routes phase):

- **routes phase** (`pipeline-phases/routes.ts`): registry key + node id +
  handler-symbol lookup. The Route node still carries the bare URL as its
  `name`; middleware/fetch URL-matching now reads the explicit `entry.url`.
- **call-processor** (`call-processor.ts`): `resolveRouteHandlerSymbols` is
  re-keyed by identity, so each verb resolves its own handler. A verb-less
  `fetch()` consumer (a fetch call has no statically-known method) still matches
  by URL and connects to **every** Route node at that URL — one FETCHES edge per
  verb. Consumer matching deliberately stays URL-only; verb-aware consumer
  matching is out of scope.
- **processes phase** (`pipeline-phases/processes.ts`): `ENTRY_POINT_OF` targets
  the identity-keyed node id.

## Schema

Bumps `INCREMENTAL_SCHEMA_VERSION` 4 → 5. Persisted pre-v5 Route nodes use the
old URL-only ids; an incremental top-up would strand them alongside new
composite-keyed nodes, so a pre-v5 stamp now forces a full re-analyze.

## Tests

- `test/integration/multi-verb-route-identity.test.ts` (new, end-to-end via
  `runPipelineFromRepo`):
  - same-URL GET/POST → two distinct nodes, each with its own verb;
  - each verb resolves to its own handler symbol (re-keyed `routeHandlerSymbols`);
  - filesystem route stays URL-only (regression guard, byte-identical id);
  - a filesystem route and a same-URL decorator route coexist as separate nodes;
  - a verb-less `fetch()` consumer connects to every Route node at the URL.
- `test/unit/route-process-linking.test.ts` (new): one `ENTRY_POINT_OF` edge per
  verb node; the old URL-only id is **not** used.
- `test/unit/resolve-route-handler-symbols.test.ts`: re-keyed; adds a multi-verb
  case (GET/POST resolve independently) + first-writer-wins per identity.
- `test/integration/resolvers/express-routes.test.ts`: updated to the new
  semantics — same-path GET/POST is now one node per verb.
- `blade-template-routes.test.ts` / `call-summary-schema-version.test.ts`:
  updated for the registry key shape and the schema bump.

`tsc --noEmit` clean; route/process/fetch suites green; prettier + eslint clean.


koriyoshi2041:
Local validation on `bcfbb637` looks good from the route-identity side.

Ran:
- `npm test -- --run test/integration/multi-verb-route-identity.test.ts test/unit/route-process-linking.test.ts test/unit/resolve-route-handler-symbols.test.ts test/integration/resolvers/express-routes.test.ts test/unit/blade-template-routes.test.ts test/unit/call-summary-schema-version.test.ts` -> 6 files / 33 tests passed
- `npx tsc --noEmit` -> passed
- `git diff --check origin/main...HEAD` -> passed

Code-read note: `routeNodeKey()` is now the shared identity primitive across route creation, handler-symbol lookup, and `ENTRY_POINT_OF` linking, which is the important part for avoiding another split-brain route id. I did not find a blocker in this pass.

magyargergo (commented):
**Recommendation: worth addressing two issues before merge.** The route re-keying is internally coherent and well-tested for the new behavior, but two issues sit in the feature's own blast radius — a schema-v5 migration guard that the unchanged-commit fast path bypasses (silent stale index on the most common upgrade path), and an `ENTRY_POINT_OF` linker that still keys by file, cross-wiring same-file multi-verb controllers. Both are fixable in this PR with in-file precedents.

## Engines

Three review methods ran independently. **Claude** ran a GitNexus swarm (risk-architect, test-ci-verifier) plus ce personas (correctness, adversarial, api-contract, maintainability, testing, project-standards, performance). **Codex** (GPT-5.5/xhigh) ran swarm, ce, and adversarial — all three Codex legs produced real findings. The strongest signals below are Codex lanes agreeing with Claude lanes (different engine, same scaffolding); the headline P1 is a *cross-engine contradiction* — Codex caught a bug that two Claude lanes gave explicit false assurance on, which is exactly why a second engine runs.

## P1 — Schema v5 bump is bypassed by the unchanged-commit fast path

**Where it lives:** `run-analyze.ts:753` (NOT in this diff) — the fix and the schema bump are. **Anchor in diff:** `repo-manager.ts:254` (the `INCREMENTAL_SCHEMA_VERSION = 5` bump).

**Engine status — CODEX-UNIQUE, contradicts both Claude lanes.** Codex B1 (swarm risk) found it. Claude correctness (`claude-ce-correctness.md:73-77`) and Claude risk-architect both explicitly asserted the *opposite* — "v5 forces full re-analyze" — because they only checked the incremental gate at `run-analyze.ts:977` and never looked at the earlier fast path. This is the canonical reason to run a second engine.

**Trigger:** a repo indexed at commit X with schema v4 (old URL-only Route nodes). User upgrades GitNexus to this build (schema v5 + multi-verb code) and re-runs `analyze` on the **same commit X with a clean worktree**.

**Evidence (code-read):** `run-analyze.ts:753` `if (existingMeta && !options.force && existingMeta.lastCommit === currentCommit)` returns `{ alreadyUpToDate: true }` with **no schemaVersion check**. `schemaVersion` is first consulted only at `run-analyze.ts:977` inside `isIncremental` — never reached on the unchanged-commit fast path. So stale v4 URL-only Route nodes persist and POST routes stay missing. This **defeats the exact migration safety the bump's own comment promises** (`repo-manager.ts:247-254`: "a top-up against a pre-v5 index would strand old url-keyed Route nodes... force a full re-analyze instead"). The bump only protects the writeback path; it does nothing if analysis never runs.

**Fix (in-file precedent):** `run-analyze.ts:738` already does exactly this for a different meta-mismatch — the `pdgModeMismatch` block sets `options = { ...options, force: true }` **before** the fast path, with a comment that it MUST sit before the `alreadyUpToDate` early-return. Mirror it: before line 753, if `existingMeta.schemaVersion !== INCREMENTAL_SCHEMA_VERSION` set `options.force = true` (with a log). ~6 lines. Add a regression test: unchanged commit + v4 meta → analyze does NOT return `alreadyUpToDate`.

## P2 — `ENTRY_POINT_OF` links by filePath, ignoring per-verb handlerSymbolId

**Where:** `processes.ts:110-156`. **Anchor in diff:** `processes.ts:120` (the changed `list.push(routeNodeKey(entry.method, entry.url))`) feeding the link loop at `processes.ts:143-149`.

**Engine status — CROSS-CODEX corroborated.** Codex B2 (ce, confidence 100) and Codex B3 (adversarial, medium) independently flagged the identical line with the identical fix. Claude maintainability grazed the same line (`processes.ts:111`) but only as a "re-derived key" drift hazard, not the mislink — partial Claude corroboration.

**Trigger:** one controller file with `GET /items -> listItems()` and `POST /items -> createItem()`, each with its own detected process.

**Evidence (code-read):** `processes.ts:110-120` builds `routesByFile` keyed by `entry.filePath` (every verb's `routeNodeKey` collapsed under the file). `processes.ts:143-149` then links **every** routeKey for a file to **every** process whose entry point is in that file. `handlerSymbolId` — which this PR now stamps per verb on the Route node — is never consulted. Result: the `GET /items` Route node gets `ENTRY_POINT_OF` to **both** `listItems` and `createItem` processes, and `POST /items` likewise. The PR manufactures per-verb precision and then discards it here, so `route_map`/`impact`/process views can attribute the POST flow to GET and vice versa.

**Fix (in-file precedent):** the **Tool** linking 10 lines below (`processes.ts:122-160`) already does it right — `toolsByHandlerId` keyed by `td.handlerNodeId`, then links only when `proc.entryPointId` matches, with a filePath fallback. Mirror that for routes: build a `routesByHandlerId` map by reading each Route node's `handlerSymbolId` **from its graph properties** (it's stamped on the node in `routes.ts`, not on `RouteEntry`), link only when `handlerSymbolId === proc.entryPointId`, and keep the filePath fallback for routes without a resolved handler. Add a two-same-file-handler test asserting GET links only to the GET process.

## P2 — Roundtrip tests pin the old url-only id and no longer cover the composite-id persistence path

**Where:** `route-method-roundtrip.test.ts:36` and `route-handler-symbol-roundtrip.test.ts:33` — **neither file is in this PR's diff**, so this is body-only (not inlinable).

**Engine status — STRONG cross-engine:** Claude api-contract (Finding 1), maintainability (M2), testing (F2), risk-architect (F1), plus Codex B1 (Lane 4).

Both tests hand-build `id: 'Route:/api/orders'` with `method: 'POST'`, and their comments claim to "mirror what the routes phase now emits." Production now emits `Route:POST /api/orders` for a method-bearing declarative route. The tests stay **green** — they seed and match the same synthetic id via `buildTestGraph`, never running `routeNodeKey` — so this is **not a break**. But they no longer exercise the composite-id CSV→COPY→`HANDLES_ROUTE_QUERY` round-trip, and a space-in-id COPY failure (historically swallowed silently) would go uncaught. There is currently **no** persistence-layer test for the new `Route:POST /api/orders` id format. Fix: seed `generateId('Route', routeNodeKey('POST', '/api/orders'))` and update the Cypher MATCH; drop the inaccurate "mirroring" comment.

> Note: the A1 test-ci-verifier called this "P0, block merge" — that is an **over-call**. The tests are self-consistent and pass; this is a coverage/fidelity gap, correctly P2.

## Lower-priority / body-only

- **manifest-extractor `LIMIT 1` verb-ambiguity** (api-contract AC-2, single-lane). After the split, `GET /x` and `POST /x` share `name='/x'`; `WHERE route.name=$n ORDER BY filePath LIMIT 1` (`manifest-extractor.ts`) picks an arbitrary verb's handler for the cross-repo manifest. **Pre-existing** method-imprecision, explicitly deferred under the #2269 scope decision — recommend as the method-aware #2269 follow-up. Not introduced by this PR.
- **`api_impact` response-shape switch** (api-contract AC-3, single-lane, **UNVERIFIED**). `local-backend.ts` (~6642) returns flat `results[0]` for one node vs `{routes,total}` for many; a GET+POST URL now yields 2 nodes → wrapped shape where it was flat. Plausible but not coordinator-verified and depends on a specific MCP call shape — flagged to check, not asserted.
- **Stale comments (P3 doc-drift):** `route-path.ts:4` header still says "keyed by normalized URL"; `java.ts:684` still says "2nd verb dropped (Route nodes are URL-keyed)" — now false, this PR closes that gap; `manifest-extractor.ts:196` implies a url-only id.
- **`if`-branch in `multi-verb-route-identity.test.ts:88`** (P3 nit). A collector-loop filter inside `forEachRelationship`, not a vacuous gate — but the project forbids `if` in tests categorically.
- **Known gap (not blocking):** Java `routeCoverage` is still `'partial'` at `java.ts:679`. The issue's payoff (flip to `'complete'`, re-enable parse-skip) is a one-liner NOT in this PR — a #2280-closing follow-up.

## Validated / refuted (credit — validation is a feature)

- **Three-phase key alignment is byte-identical.** `routes.ts` (registry key + node id + handler lookup), `call-processor.ts` (`resolveRouteHandlerSymbols` re-keyed via `claim(httpMethod,...)`), and `processes.ts` (`ENTRY_POINT_OF`) all key through `normalizeRouteMethod → routeNodeKey`, fed from the same `allExtractedRoutes`/`allDecoratorRoutes` arrays. Confirmed by Claude correctness + both swarm lanes + Codex.
- **Delimiter collision refuted.** `normalizeExtractedRoutePath` always returns a leading-slash URL, so a method-less key starts with `/` and a method-keyed key starts with a verb letter; keys are only compared by full-string equality. A URL containing a space is harmless.
- **Downstream Route-id consumers unaffected.** group manifest-extractor, `route_map`/`api_impact`, and FTS join on `route.name` (still the bare URL), not a reconstructed id.
- **FETCHES fan-out is bounded** (~verbs per URL) and self-heals at the bridge `dedupeContracts` boundary before persistence (adversarial F1, low) — no wrong data persisted. **Performance: non-issue** (complexity class unchanged; the per-call regex scan was not moved by this PR).
- **Project standards clean:** RFC#909 language-agnostic (no language-identity branches in implementation), no `any`, CHANGELOG untouched. **Security/bidi clean.** All signature changes (e.g. `processNextjsFetchRoutes`) have callers updated.

## CI status

`quality` typecheck / lint / format are **green**. Test suites (ubuntu coverage, etc.) were still **pending** at review time. The Vercel failure is the known non-infra auth gate, not a code signal.

---

*Automated multi-engine digest (Claude swarm + ce + Codex GPT-5.5). Verify findings against the code before acting; the P1 fast-path interaction in particular is worth a quick local check.*

gitnexus/src/storage/repo-manager.ts
```diff
@@ -244,8 +244,14 @@ export interface RepoMeta {
...
  * This single bump covers the whole FU-C re-index window (and the later FU-B-2).
+ * v5: `Route` node identity changed to `(method, url)` (#2289 — a same-URL
+ * GET/POST pair is now two distinct Route nodes). Every declarative-route node
+ * id moved from `Route:/x` to `Route:GET /x` (filesystem routes keep their
+ * URL-only id). The incremental writeback preserves unchanged-file rows, so a
+ * top-up against a pre-v5 index would strand old url-keyed Route nodes alongside
+ * new composite-keyed ones — force a full re-analyze instead.
  */
-export const INCREMENTAL_SCHEMA_VERSION = 4;
+export const INCREMENTAL_SCHEMA_VERSION = 5;
```
magyargergo, line 254: **P1 — schema v5 bump is bypassed by the unchanged-commit fast path.** This bump only protects the incremental *writeback* path; it does nothing on the most common upgrade path (tool upgraded, repo commit unchanged, clean worktree). `run-analyze.ts:753` `if (existingMeta && !options.force && existingMeta.lastCommit === currentCommit)` returns `{ alreadyUpToDate: true }` with **no** schemaVersion check — `schemaVersion` is first consulted only at `run-analyze.ts:977` inside `isIncremental`, which the fast path never reaches. So a pre-v5 index keeps its stale URL-only Route nodes and POST routes stay missing — defeating the migration safety this very comment promises. Fix: mirror the `pdgModeMismatch` precedent at `run-analyze.ts:738`, which sets `options = { ...options, force: true }` *before* the fast path; add `if (existingMeta.schemaVersion !== INCREMENTAL_SCHEMA_VERSION) options.force = true` before line 753 (with a log) plus a regression test (unchanged commit + v4 meta must NOT return `alreadyUpToDate`). [code-read]

gitnexus/src/core/ingestion/pipeline-phases/processes.ts
```diff
@@ -107,13 +108,16 @@ export const processesPhase: PipelinePhase<ProcessesOutput> = {
...
         let list = routesByFile.get(entry.filePath);
         if (!list) {
           list = [];
           routesByFile.set(entry.filePath, list);
         }
-        list.push(url);
+        // Push the Route node identity (`routeNodeKey`), not the bare URL, so the
+        // ENTRY_POINT_OF edge targets the same node id the routes phase created
+        // (#2289: a same-URL GET/POST pair is two distinct Route nodes).
+        list.push(routeNodeKey(entry.method, entry.url));
```
magyargergo: **P2 — `ENTRY_POINT_OF` links by filePath, discarding the per-verb `handlerSymbolId` this PR just stamped.** `routesByFile` here (lines 110-120) keys by `entry.filePath`, collapsing every verb's `routeNodeKey` under one file; the link loop at lines 143-149 then attaches **every** routeKey for a file to **every** process whose entry point is in that file. For a single controller with `GET /items -> listItems()` and `POST /items -> createItem()`, the `GET /items` node gets `ENTRY_POINT_OF` to both processes and `POST /items` likewise — the PR's new per-verb precision is thrown away, mis-attributing flows in `route_map`/`impact`/process views. Precedent: the Tool linking just below (lines 122-160) keys `toolsByHandlerId` by `td.handlerNodeId` and links only when `proc.entryPointId` matches, with a filePath fallback. Mirror it: build a `routesByHandlerId` map by reading each Route node's `handlerSymbolId` from its graph properties (it's stamped on the node in `routes.ts`, not carried on `RouteEntry`), link only when `handlerSymbolId === proc.entryPointId`, filePath fallback otherwise. Add a two-same-file-handler test asserting GET links only to the GET process. [code-read]

koriyoshi2041:
I checked the review findings locally.

P1 looks valid: the same-commit `alreadyUpToDate` path in `src/core/run-analyze.ts` can return before the later `existingMeta.schemaVersion === INCREMENTAL_SCHEMA_VERSION` incremental reuse gate is evaluated, so a clean repo with a pre-v5 stamp can skip the v5 route-id rebuild.

I put together a small local patch that:
- guards the same-commit fast path with the current incremental schema version;
- adds a regression test where a same-commit stale schema stamp rebuilds instead of returning `alreadyUpToDate`;
- updates the route method / handler roundtrip tests to build Route ids with `generateId('Route', routeNodeKey('POST', '/api/orders'))`.

Validation:
- `npm test -- --run test/unit/incremental-orchestration.test.ts test/unit/route-process-linking.test.ts test/unit/call-summary-schema-version.test.ts test/integration/route-method-roundtrip.test.ts test/integration/route-handler-symbol-roundtrip.test.ts`
- `npx tsc --noEmit`
- `git diff --check`

Local patch commit: `df542df5`.

henry201605:
Thanks for the multi-engine pass — the P1 catch in particular is exactly the kind of blind spot a single reviewer would miss. I've pushed fixes for all P1/P2/P3 items in eb4d4287. Walking through each:

## P1 — Schema v5 fast-path bypass ✅

Confirmed and fixed in `run-analyze.ts`. The new block sits **above** the `alreadyUpToDate` early-return, mirroring the in-file `pdgModeMismatch` precedent you pointed to:

```ts
if (existingMeta && existingMeta.schemaVersion !== INCREMENTAL_SCHEMA_VERSION) {
  const stampedVersion = existingMeta.schemaVersion ?? 'pre-versioning';
  log(
    `index schema changed (stamped v${stampedVersion}, this build is v${INCREMENTAL_SCHEMA_VERSION}); ` +
      `forcing a full rebuild so persisted rows match the current schema.`,
  );
  options = { ...options, force: true };
}
```

The `?? 'pre-versioning'` is only a log-friendliness tweak — non-git repos stamp `schemaVersion: undefined` (see meta save site), so a literal `vundefined` would have shown up in their re-analyze log. Behavior is unchanged: the force is harmless there because non-git repos already take the `currentCommit === ''` rebuild branch below.

**Regression test** added in `test/unit/incremental-orchestration.test.ts`: analyze a fresh repo (stamps v5), rewrite meta to `schemaVersion: 4`, re-analyze on the same commit, assert `alreadyUpToDate === undefined` and that meta is re-stamped back to v5.

## P2 — `ENTRY_POINT_OF` links by filePath ✅

Confirmed and fixed in `processes.ts`. Followed your exact suggestion — built `routesByHandlerId` keyed by reading `handlerSymbolId` from the Route node's graph properties (the source of truth `routes.ts` stamps), with a `routesWithoutHandlerByFile` fallback. The link loop now mirrors the Tool linking precedent 10 lines below: `exactRouteKeys ?? fallbackRouteKeys`.

```ts
const routeNode = ctx.graph.getNode(generateId('Route', routeKey));
const handlerSymbolId = routeNode?.properties.handlerSymbolId as string | undefined;
const targetMap = handlerSymbolId ? routesByHandlerId : routesWithoutHandlerByFile;
const bucketKey = handlerSymbolId ?? entry.filePath;
```

Two regression tests added in `test/unit/route-process-linking.test.ts`:
- **weak form** — same-file `GET /items` + `POST /items` where only `listItems` has a detected process; asserts POST does not get spuriously attached to the listItems process.
- **strong form** (mirrors your trigger verbatim) — both verbs have their own distinct processes; asserts each Route links to **exactly** its own handler's `entryPointId` and that the total route→process edge count is `2`, not the pre-fix 4-edge cross-wire.

## P2 — Roundtrip tests pin old url-only id ✅

Fixed both `route-method-roundtrip.test.ts` and `route-handler-symbol-roundtrip.test.ts`. Each now seeds the Route node with `generateId('Route', routeNodeKey('POST', '/api/orders'))`, updates the Cypher MATCH to use the same composite id, and drops the inaccurate "mirroring" comment.

Also added an explicit `expect(routeCsv).toContain(ROUTE_ID)` sanity — a CSV escaping regression on the literal-space-in-id (`Route:POST /api/orders`) would surface here instead of being silently swallowed by the extractor's `catch`.

## P3 — Doc-drift + test `if` ✅

- `route-path.ts:4` — "keyed by the normalized URL" → "keyed by `(method, url)` via `routeNodeKey` — #2289"
- `java.ts:684` — removed "(Route nodes are URL-keyed)" and noted that the same-URL multi-verb gap is closed by #2289 (so it's no longer a `routeCoverage` partial-driver)
- `manifest-extractor.ts:196` — explicit that the Route node *id* is composite post-#2289 while `route.name` remains the bare URL (which is why this URL-keyed query keeps working)
- `multi-verb-route-identity.test.ts:88` — `forEachRelationship + if` → `.filter().map()` chain (no test-level conditional)

The two new route-process-linking tests above were also written `if`-free (`filter() → map() → new Map(...)`).

## Lower-priority items — assessed, not changed in this PR

- **manifest-extractor `LIMIT 1` verb-ambiguity** — Agreed it's pre-existing and explicitly out of the #2269 scope decision; leaving for the method-aware #2269 follow-up as you suggested.
- **`api_impact` response-shape switch** — Verified, the concern is real: same URL with GET+POST now yields `routes.length === 2` and trips the `results.length === 1 ? results[0] : { routes, total }` branch in `local-backend.ts` (~6644). The fix is a design choice (group by URL? always wrap? add a `groupByUrl` param?) so it deserves its own issue rather than being snuck into this PR. **This should be tracked as a follow-up issue.**
- **Java `routeCoverage` flip to `'complete'`** — Agreed, deferred under #2280 as you noted.

## Validation

- `npx tsc --noEmit` clean
- `npx prettier --check` clean on all touched files
- 9 touched suites / 43 tests pass locally:
  ```
  npm test -- --run \
    test/unit/route-process-linking.test.ts \
    test/integration/multi-verb-route-identity.test.ts \
    test/unit/resolve-route-handler-symbols.test.ts \
    test/integration/resolvers/express-routes.test.ts \
    test/integration/route-method-roundtrip.test.ts \
    test/integration/route-handler-symbol-roundtrip.test.ts \
    test/unit/blade-template-routes.test.ts \
    test/unit/call-summary-schema-version.test.ts \
    test/unit/incremental-orchestration.test.ts
  ```

Ready for another pass when you have a moment.

gitnexus/src/core/ingestion/pipeline-phases/routes.ts
```diff
@@ -138,39 +149,8 @@ function escapeRegex(s: string): string {
...
-  if (typeof raw !== 'string') return undefined;
-  const verb = raw.trim().toUpperCase();
-  // '*' marks a method-agnostic route (e.g. a Django function view handles any
-  // verb). Preserve it so the contract layer emits a wildcard provider that
-  // matches consumers of any method, instead of silently narrowing to GET.
-  if (verb === '*') return '*';
-  return VALID_HTTP_METHODS.has(verb) ? verb : undefined;
-}
+// Re-exported for existing consumers/tests that import them from the routes phase.
+export { normalizeExtractedRoutePath, normalizeRouteMethod, routeNodeKey };
```
magyargergo: Can you please make sure we are not re-exporting? Please also update the caller sites.

magyargergo (changes requested):

gitnexus/src/core/ingestion/pipeline-phases/routes.ts
henry201605: Done in 39d6ef38.

- Removed the re-export in `routes.ts` (the `export { normalizeExtractedRoutePath, normalizeRouteMethod, routeNodeKey }` line plus its comment).
- Updated the two remaining caller sites that imported via the routes phase to import directly from the source module `route-extractors/route-path.js`:
  - `test/unit/blade-template-routes.test.ts` — split the import group so `normalizeExtractedRoutePath` + `routeNodeKey` come from `route-extractors/route-path.js` while `extractTemplateStaticFetchCalls` / `isTemplateRouteCandidate` / `routesPhase` stay on `pipeline-phases/routes.js` (those are routes.ts's own exports, not re-exports).
  - `test/unit/spring-route-extractor-parity.test.ts` — switched its single `normalizeExtractedRoutePath` import.
- `call-processor.ts` and `processes.ts` were already importing from the source, so no change needed there. Verified end-to-end: no remaining `pipeline-phases/routes` import in the tree references those three symbols.

Also rolled into the same commit: a fix for two pre-existing `run-analyze.test.ts` fast-path tests that started failing on CI after the P1 schema-version guard landed. The fixtures hand-built a `RepoMeta` with no `schemaVersion` field; the new guard (faithful to your literal `existingMeta.schemaVersion !== INCREMENTAL_SCHEMA_VERSION`) treated that pre-versioning shape as a mismatch and forced a rebuild, short-circuiting the fast path under test. Stamped the current `schemaVersion` on those fixtures so they reflect the post-#2289 meta shape `runFullAnalysis` actually writes.

magyargergo (approved):

PR #2305, opened 2026-06-26
fix(ingestion): index generator function declarations

### LINKED ISSUE #2303: Analyzer does not index async generator functions (async function*)
## Summary

The analyzer does not index **async generator functions** (`async function*`). Top-level `export async function*` declarations are missing from the symbol graph, so `context` / `impact` (and the MCP equivalents) return **"Symbol not found"** for them. Regular `function` and `async function` declarations in the *same files* index correctly, so the files are parsed — only the generator declarations are skipped.

This is a meaningful correctness gap: an `impact()` that returns "not found" reads as "no callers, safe to change" — a false green for anyone using GitNexus to gate edits.

## Environment

- gitnexus: **1.6.8**
- node: v24.13.0
- OS: Windows 11
- Project: TypeScript (ES modules)

## Repro

Given a TypeScript file containing both a regular function and an async generator:

```ts
// src/lib/coach.ts
export function userText(input: CoachInput): string { /* ... */ }          // regular

export async function* runCoachLoop(                                        // async generator
  input: CoachInput,
  deps: CoachDeps
): AsyncGenerator<CoachEvent> { /* ... */ }
```

After a clean rebuild (`gitnexus analyze --force`):

| Symbol | Declaration | Result |
|---|---|---|
| `userText` | `export function` | ✅ found (`Function:src/lib/coach.ts:userText`) |
| `dispatchTool` | `export async function` | ✅ found, full call graph |
| `runCoachLoop` | `export async function*` | ❌ `Symbol 'runCoachLoop' not found` |
| `streamComplete` | `export async function*` | ❌ `Symbol 'streamComplete' not found` |

Both missing symbols are async generators in different files; both files have other (non-generator) symbols that index fine.

## Expected

`async function*` declarations should be indexed as `Function` symbols (with their call edges), the same as `function` and `async function`.

## Impact

Async generators are common for streaming / agentic loops (SSE token streams, bounded tool-use loops). In this project the two missing symbols are the LLM stream and the agent loop — i.e. exactly the high-blast-radius code where impact analysis matters most. Their callers/callees are invisible at the symbol level, so `impact`/`context` give false-negative safety signals on that path.

## Likely cause

The AST visitor probably matches `FunctionDeclaration` (and async) but not the generator variant (`node.generator === true`, or a missing `async function*` / `MethodDefinition` generator case). Adding the generator case to the declaration handler should cover it.

koriyoshi2041:
Problem
async function* declarations parse as generator_function_declaration, but the legacy definition query only captured function_declaration. That leaves exported async generators out of the Function graph, so context/impact can report Symbol not found for streaming or loop entry points. Fixes #2303.

Fix
- Capture generator_function_declaration as @definition.function for TypeScript and JavaScript.
- Add a query-level regression for exported async function*.
- Add a worker-backed integration regression that checks the Function node is emitted.

Test
- npm run build
- npm test -- --run test/unit/tree-sitter-queries.test.ts test/integration/typescript-async-generator-functions.test.ts
- npx tsc --noEmit
- git diff --check

Risk
Low. This only adds another named function-declaration node type to the existing Function definition path; anonymous generators remain unchanged.

### DIFF
--- a/gitnexus/src/core/ingestion/tree-sitter-queries.ts
+++ b/gitnexus/src/core/ingestion/tree-sitter-queries.ts
@@ -27,6 +27,9 @@ export const TYPESCRIPT_QUERIES = `
 (function_declaration
   name: (identifier) @name) @definition.function
 
+(generator_function_declaration
+  name: (identifier) @name) @definition.function
+
 ; TypeScript overload signatures (function_signature is a separate node type from function_declaration)
 (function_signature
   name: (identifier) @name) @definition.function
@@ -372,6 +375,9 @@ export const JAVASCRIPT_QUERIES = `
   name: (identifier) @name) @definition.class
 
 (function_declaration
+  name: (identifier) @name) @definition.function
+
+(generator_function_declaration
   name: (identifier) @name) @definition.function
 
 (method_definition
--- a/gitnexus/test/integration/typescript-async-generator-functions.test.ts
+++ b/gitnexus/test/integration/typescript-async-generator-functions.test.ts
@@ -1,0 +1,28 @@
+import { describe, expect, it } from 'vitest';
+import { parseFilesWithWorkers } from '../helpers/worker-parse.js';
+
+describe('TypeScript async generator functions', () => {
+  it('indexes exported async function* declarations as Function nodes', async () => {
+    const { graph } = await parseFilesWithWorkers([
+      {
+        path: 'src/coach.ts',
+        content: `
+          export function userText(input: string): string {
+            return input.trim();
+          }
+
+          export async function* runCoachLoop(input: string): AsyncGenerator<string> {
+            yield userText(input);
+          }
+        `,
+      },
+    ]);
+
+    const functionNames = new Set(
+      graph.nodes.filter((node) => node.label === 'Function').map((node) => node.properties.name),
+    );
+
+    expect(functionNames).toContain('userText');
+    expect(functionNames).toContain('runCoachLoop');
+  });
+});
--- a/gitnexus/test/unit/tree-sitter-queries.test.ts
+++ b/gitnexus/test/unit/tree-sitter-queries.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest';
 import Parser from 'tree-sitter';
+import JavaScript from 'tree-sitter-javascript';
 import TypeScript from 'tree-sitter-typescript';
 import {
   TYPESCRIPT_QUERIES,
@@ -17,29 +18,30 @@ import {
   DART_QUERIES,
 } from '../../src/core/ingestion/tree-sitter-queries.js';
 
+function capturedDefinitionFunctionNames(
+  language: Parameters<Parser['setLanguage']>[0],
+  querySource: string,
+  src: string,
+): string[] {
+  const parser = new Parser();
+  parser.setLanguage(language);
+  const query = new Parser.Query(language, querySource);
+  const tree = parser.parse(src);
+  const names: string[] = [];
+  for (const match of query.matches(tree.rootNode)) {
+    let isFunction = false;
+    let name: string | undefined;
+    for (const capture of match.captures) {
+      if (capture.name === 'definition.function') isFunction = true;
+      if (capture.name === 'name') name = capture.node.text;
+    }
+    if (isFunction && name !== undefined) names.push(name);
+  }
+  return names;
+}
+
 describe('tree-sitter queries', () => {
   describe('TypeScript queries', () => {
-    function capturedDefinitionFunctionNames(src: string): string[] {
-      const parser = new Parser();
-      parser.setLanguage(TypeScript.typescript as Parameters<Parser['setLanguage']>[0]);
-      const query = new Parser.Query(
-        TypeScript.typescript as Parameters<Parser['setLanguage']>[0],
-        TYPESCRIPT_QUERIES,
-      );
-      const tree = parser.parse(src);
-      const names: string[] = [];
-      for (const match of query.matches(tree.rootNode)) {
-        let isFunction = false;
-        let name: string | undefined;
-        for (const capture of match.captures) {
-          if (capture.name === 'definition.function') isFunction = true;
-          if (capture.name === 'name') name = capture.node.text;
-        }
-        if (isFunction && name !== undefined) names.push(name);
-      }
-      return names;
-    }
-
     it('captures class declarations', () => {
       expect(TYPESCRIPT_QUERIES).toContain('class_declaration');
       expect(TYPESCRIPT_QUERIES).toContain('@definition.class');
@@ -56,12 +58,16 @@ describe('tree-sitter queries', () => {
     });
 
     it('captures async generator function declarations as function definitions', () => {
-      const names = capturedDefinitionFunctionNames(`
+      const names = capturedDefinitionFunctionNames(
+        TypeScript.typescript as Parameters<Parser['setLanguage']>[0],
+        TYPESCRIPT_QUERIES,
+        `
         export function userText() { return ''; }
         export async function* runCoachLoop(): AsyncGenerator<string> {
           yield 'ready';
         }
-      `);
+      `,
+      );
 
       expect(names).toContain('userText');
       expect(names).toContain('runCoachLoop');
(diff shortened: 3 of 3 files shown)


magyargergo (commented):
## Tri-review: `fix(ingestion): index generator function declarations`

This PR adds `(generator_function_declaration name: (identifier) @name) @definition.function` to both `TYPESCRIPT_QUERIES` and `JAVASCRIPT_QUERIES` in `gitnexus/src/core/ingestion/tree-sitter-queries.ts` (+6), so the structure-phase symbol-emit query finally captures `function*` / `async function*` declarations. It adds one TS unit test (real-tree-sitter parse) and one TS worker integration test. **Closes #2303** (async generators were silently un-indexed → `impact()` returned "not found", a false-green safety signal for exactly the streaming / agentic code with the highest blast radius).

### How it was reviewed
Two engines ran the same swarm + ce scaffolding:
- **Engine A (Claude):** GitNexus swarm (risk-architect, facts-historian) + ce personas (correctness, testing, maintainability, project-standards).
- **Engine B (Codex, GPT-5.5 / xhigh):** swarm (B1, Solo) **APPROVE + P3**, ce report-only (B2) **P2**, adversarial (B3) **approve / no material findings**. All three Codex legs were live. (B2's correctness + learnings sub-lanes crashed with platform stream errors, but B2 still produced its merged JS-coverage finding; its maintainability and agent-native sub-lanes ran.)

### What's solid (credited)
- **Correctness is verified, not assumed.** The Claude correctness lane parsed real source against the declared grammars (`tree-sitter-typescript@0.23.2` / `tree-sitter-javascript`): `function*` and `async function*` both parse to a single `generator_function_declaration` node (async is a child keyword), the `name: (identifier)` field binds for both, and `export`-wrapped forms still match the nested node. No defect found.
- **This aligns the two parse paths.** `generator_function_declaration` was *already* recognized everywhere downstream — registry-primary scope queries (`languages/typescript/query.ts`, `languages/javascript/query.ts`), `captures.ts`, CFG visitors, method/type/embeddings extractors, `ast-helpers.ts` `FUNCTION_NODE_TYPES`. This PR fills the **one** missing structure-phase node-emit step, closing a parity gap rather than introducing inconsistency.
- **CI is green** on every substantive check across both engines' facts: Analyze (js-ts + python), CodeQL, gitleaks, quality/{format,lint,typecheck,typecheck-web}, tests on ubuntu/macos/windows + benchmarks + packaged-install smoke, and the **tree-sitter ABI job on all 3 platforms** (which compiles the new JS query against the real grammar — so it is syntactically valid, though it does not assert capture semantics).

### Headline finding (the only actionable item)

**The `JAVASCRIPT_QUERIES` half of the fix has no behavioral test. — P3, [code-read]**

- **Evidence:** the JS rule is added at `gitnexus/src/core/ingestion/tree-sitter-queries.ts:380`, but every JS-side test is string-containment: `expect(JAVASCRIPT_QUERIES).toContain('@definition.function')` (`test/unit/tree-sitter-queries.test.ts:93`) — an assertion that **already passed before this PR** because `function_declaration` contributes the same tag, so it gives zero delta for the new rule. The new parse-based helper `capturedDefinitionFunctionNames` (`tree-sitter-queries.test.ts:22-41`) builds its `Parser.Query` only from `TypeScript.typescript` + `TYPESCRIPT_QUERIES`, and the integration fixture is `src/coach.ts` (TypeScript). No test parses JS generator source.
- **Trigger:** a future `tree-sitter-javascript` node-type rename, or an accidental removal of the JS rule (while the TS rule stays), would ship un-indexed JS generators with no failing test to catch it — re-opening exactly the #2303 false-green for `.js`/`.jsx` code.
- **Why P3 (not a blocker):** the JS path is **verified-correct today** (the correctness lane's direct parse: `[JS] "async function* agen(){}" → ["agen"]`) and the ABI CI job compiles the JS query on 3 platforms, so present exposure is nil — this is regression-detection coverage on already-working code, not a latent bug. Most lanes landed on P3; Codex's ce lane rated it **P2** on a stricter "changed behavior must have a behavioral test" DoD reading, which is a fair view to weigh.
- **Cross-engine corroboration: STRONG** (two engines, not five independent reviewers). Both engines reached it: **Claude A** — risk-architect (F2, P3), ce-testing (T1, P3/conf 100), maintainability (residual); **Codex B** — B1 swarm (Lanes 4 & 6, P3), B2 ce (P2/conf 100).
- **Fix:** mirror the TS unit helper for JavaScript (`new Parser.Query(JavaScript, JAVASCRIPT_QUERIES)` over `function* / async function*` source), or add a `.js` fixture to the integration test asserting `runCoachLoop` is captured.

### Lower-priority items (body only, not blocking)
- **Plain non-async `function*` is never exercised** — both new tests use `async function*`. Same node type, so correctness holds; a one-line bare-`function*` fixture would prove it. (Claude-only, P3.)
- **No exact-count / double-count assertion** — the integration test collects names into a `Set` and the unit helper uses `toContain`, so neither would catch a hypothetical double-match from a future pattern-widening. Low priority given the node types are distinct. (Claude-only, P3.)
- **Maintainability nits** — the `as Parameters<Parser['setLanguage']>[0]` cast is repeated twice inside the helper (`tree-sitter-queries.test.ts:24,26`); a local `const lang = …` would DRY it (M-01, P3). The new generator rule ships without the explanatory comment its adjacent `function_signature` sibling carries (M-02, P3).

### Suspicions raised and dropped (validation is a feature)
- **Double-emit / double-count: REFUTED** by both engines and reproduced — `generator_function_declaration` and `function_declaration` are distinct, mutually-exclusive node types; a generator matches exactly one rule (correctness lane: every declaration yielded exactly one captured name).
- **"No language names in shared `core/ingestion/`": REFUTED** by both engines. `tree-sitter-queries.ts` is by design a file of per-language query-string **constants** that live `LanguageProvider`s wire into `provider.treeSitterQueries`; editing them is the canonical extension point, not the resolution-logic language-branching the rule forbids.
- **`async function*` coverage: CONFIRMED** — same `generator_function_declaration` node, `async` is a child keyword; the single rule covers both forms.
- **Out of scope, not a regression:** generator *expressions* (`const f = function*(){}`) remain uncaptured — consistent with the existing `function_expression` treatment, a pre-existing gap, not introduced here.

### CI status
MERGEABLE; `mergeStateStatus` UNSTABLE is **solely** the Vercel deploy-authorization gate (fork-contributor auth infra, zero code signal). All substantive checks green; 13,487 passed / 0 failed / 57 known-skips; coverage +0.4–0.6pp.

### Verdict
**Production-ready with minor follow-ups.** Clean feature/fix PR; `mergeable`. The one actionable item (JS behavioral-test gap) is a non-blocking P3.

### Coverage
Small PR (+69/-0, 3 files) read in full by both engines; production change parse-verified against the real grammars. This is an automated multi-tool digest (Claude + Codex/GPT-5.5) — please sanity-check before acting.

magyargergo:
### Empirical reproduction — scope resolution, with and without the fix

To confirm the change does more than create a node (the PR's tests assert node existence, not resolution), I ran the repo's real pipeline (`runPipelineFromRepo`) on a small fixture, comparing released `gitnexus@1.6.8` against the same build with this PR's `generator_function_declaration` capture applied.

**Fixture** — each generator calls a function in its body; `main` calls both generators (covers calls *to* and *from* generators, async **and** plain):

```ts
export function userText(input: string): string { return input.trim(); }

export async function* runCoachLoop(input: string): AsyncGenerator<string> {
  yield userText(input);            // call FROM an async-generator body
}
export function* plainGen(input: string): Generator<number> {
  yield userText(input).length;     // call FROM a plain-generator body
}
export async function main(): Promise<void> {
  for await (const t of runCoachLoop('hi')) void t;   // call TO async generator
  for (const n of plainGen('yo')) void n;             // call TO plain generator
}
```

**Result**

| | `Function` nodes | `CALLS` edges |
|---|---|---|
| **without** (released) | `main`, `userText` | `coach.ts -> userText` |
| **with this PR** | `main`, `userText`, `runCoachLoop`, `plainGen` | `main -> runCoachLoop`, `main -> plainGen`, `runCoachLoop -> userText`, `plainGen -> userText` |

**What it shows**
- **Without:** both generators are missing as symbols; calls *to* them don't exist; and the call *from* the generator body mis-attributes to the **file** (`coach.ts -> userText`) since the generator has no symbol to own the scope — so it's not just a missing node, it's a wrong edge source. (This is #2303.)
- **With:** both `function*` and `async function*` become proper `Function` symbols, and **all four** edges resolve — calls to the generators, and calls from their bodies now correctly owned by the generator rather than the file. The downstream scope-resolution queries already listed `generator_function_declaration`; this fills the one missing structure-phase capture so the two line up.

**One follow-up:** none of those four `CALLS` edges are covered by the PR's tests (they assert only that the `Function` node exists). A single integration assertion on, say, `main -> runCoachLoop` and `runCoachLoop -> userText` would lock in the resolution behavior — and a JS (`.js`) variant would also close the JavaScript-path coverage gap noted in the review.

<sub>Automated reproduction via the repo's own pipeline; please verify before relying on it.</sub>

PR #2307, opened 2026-06-26
fix(search): make FTS stemmer configurable

### LINKED ISSUE #2306: FTS keyword search is unreliable for non-English languages due to hardcoded Porter stemmer
# FTS keyword search is unreliable for non-English languages due to hardcoded Porter stemmer

## Summary

GitNexus's BM25/FTS keyword search uses LadybugDB's `CREATE_FTS_INDEX` with `stemmer := 'porter'`, which is an **English-only** stemming algorithm. This makes keyword search for Chinese, Japanese, Korean, and other non-English doc comments (Javadoc/KDoc/JSDoc/etc.) unreliable — some terms match unpredictably while others fail entirely.

There is **no environment variable, CLI flag, or configuration option** to change the stemmer.

### Source code evidence

The stemmer is hardcoded in two places with no override mechanism:

1. **`gitnexus/src/core/lbug/lbug-adapter.ts`** — `createFTSIndex` and `ensureFTSIndex` both default to `'porter'`:
   ```typescript
   // Line 2348
   export const createFTSIndex = async (
     tableName: string, indexName: string, properties: string[],
     stemmer: string = 'porter',   // ← hardcoded, no env var
   ): Promise<void> => { ... }

   // Line 2444
   export const ensureFTSIndex = async (
     tableName: string, indexName: string, properties: string[],
     stemmer: string = 'porter',   // ← same
   ): Promise<void> => { ... }
   ```

2. **`gitnexus/src/core/search/fts-indexes.ts`** — `createSearchFTSIndexes` calls `createFTSIndex` without passing any stemmer parameter, so the `'porter'` default always applies:
   ```typescript
   // Line 26
   await createFTSIndex(table, indexName, [...properties]);
   // No stemmer argument → defaults to 'porter'
   ```

### Why Porter stemmer breaks non-English search

The Porter stemmer applies English-specific morphological rules (strip "-ing", "-ed", "-tion", etc.) to **all text regardless of language**. For CJK (Chinese/Japanese/Korean) text that doesn't use spaces as word boundaries, this produces nonsensical token fragments, causing:

- **Unpredictable partial matches**: A Chinese Javadoc phrase like `/** * 订单审批流程自动推送 */` may match when searching the full phrase, but sub-terms like "订单审批" fail to match
- **Complete misses**: Searching for "审批结果通知到邮件" returns zero results even when the exact doc comment exists in the indexed codebase
- **European languages also affected**: German "Bearbeitung" → stemmer strips to "bearbeit", French "connexion" → "connex", etc. — while still functional, the aggressive English rules produce suboptimal stems for non-English European languages

### Impact on GitNexus hybrid search

GitNexus's `query()` tool uses a **hybrid search** pipeline: BM25 (FTS keyword) + Vector (embedding semantic) merged via Reciprocal Rank Fusion (RRF). When FTS returns poor or zero results for non-English keywords, the BM25 contribution to the RRF merge is weakened or absent, degrading overall search quality — even when vector search is working correctly.

Note: Vector/embedding search is **not affected** by this issue — it already supports model configuration via `GITNEXUS_EMBEDDING_MODEL`, `GITNEXUS_EMBEDDING_URL`, and `GITNEXDING_EMBEDDING_DIMS` environment variables.

## LadybugDB stemmer options

LadybugDB's FTS extension supports the following stemmers (verified against the current version):

| Category | Stemmers |
|----------|----------|
| Language-specific Snowball | `arabic`, `basque`, `catalan`, `danish`, `dutch`, `english`, `finnish`, `french`, `german`, `greek`, `hindi`, `hungarian`, `indonesian`, `irish`, `italian`, `lithuanian`, `nepali`, `norwegian`, `portuguese`, `romanian`, `russian`, `serbian`, `spanish`, `swedish`, `tamil`, `turkish` |
| Generic | `porter` (English Snowball), `none` (no stemming) |
| **Not supported** | No CJK-specific stemmer (`chinese`, `japanese`, `korean`, `cjk` — all rejected) |

For CJK languages, `stemmer := 'none'` is the best available option — it disables stemming entirely and allows character-level matching, which works significantly better than Porter for Chinese/Japanese/Korean text.

## Suggested fix

Add a configuration option (environment variable or CLI flag) to allow users to choose the FTS stemmer. For example:

- **Environment variable**: `GITNEXUS_FTS_STEMMER` (default: `'porter'` for backward compatibility)
- **CLI flag**: `--fts-stemmer <name>` on `gitnexus analyze`

This would pass the configured stemmer value through `createSearchFTSIndexes` → `createFTSIndex` / `ensureFTSIndex`.

For CJK-heavy codebases, users could set `GITNEXUS_FTS_STEMMER=none`. For German-heavy codebases, `GITNEXUS_FTS_STEMMER=german`. The default `'porter'` preserves current behavior for English-centric projects.

A more advanced option could allow **per-index stemmer configuration** (e.g., different stemmers for different symbol tables), but a single global setting would be a sufficient first step.

## Reproduction

1. Index a Java project containing Chinese Javadoc comments (e.g., `/** * 订单审批流程自动推送 */`)
2. Run `gitnexus analyze --force` to build FTS indexes with description content
3. Search via MCP `query({search_query: "订单审批"})` — observe poor/zero BM25 results
4. Search via MCP `query({search_query: "审批结果通知到邮件"})` — observe zero results despite exact match in indexed doc comment

## Environment

- GitNexus version: current `main` branch (post #2286, #2300)
- LadybugDB: current bundled version
- Affected languages: Chinese, Japanese, Korean (worst), all non-English European languages (moderate)

koriyoshi2041:
## Summary
- add `GITNEXUS_FTS_STEMMER` for BM25/FTS index rebuilds, defaulting to the existing `porter` behavior
- validate stemmer names before passing them into `CREATE_FTS_INDEX`
- document `none` for CJK-heavy repositories and `--repair-fts` after changing the setting

Fixes #2306.

## Tests
- `npm test -- --run test/unit/fts-indexes.test.ts`
- `npx tsc --noEmit`
- `git diff --check`

## Risk
Low for existing installs because unset configuration keeps the current `porter` stemmer. Invalid configured values now fail fast before FTS indexes are rebuilt.


magyargergo (commented):
## Tri-review — PR #2307 `fix(search): make FTS stemmer configurable` (fixes #2306)

**One-line:** Functionally sound, backward-compatible feature with an airtight injection mitigation — but it ships a **deterministic, CI-confirmed test break** (a known ripple site the author missed) that must be fixed before merge.

**Engine breakdown.** Claude ran the **swarm** (test-ci, security-boundary, risk) and **ce** personas (correctness, testing, maintainability, project-standards, security, adversarial). Codex (GPT-5.5) was scheduled for swarm + ce + adversarial, but only the **Codex ce** leg ran live — the **Codex swarm** lane failed (`stream disconnected before completion`) and the **Codex adversarial** lane hit a `bwrap` sandbox error, could not read any files, and then stream-disconnected. Its "approve / no material findings" is a **tooling failure, not a refutation**, and is not counted as evidence of safety. So the headline has genuine cross-engine corroboration (Claude + Codex ce); the adversarial cross-engine pair did **not** materialize.

---

### 1. Headline (P1 · merge-blocking · CI-confirmed)

**`test/unit/bm25-search.test.ts:37-39` asserts 3-arg `createFTSIndex` calls; the PR makes them 4-arg → deterministic CI failure.**

- **Trigger:** the PR changes `createSearchFTSIndexes` to pass a 4th `stemmer` argument at `gitnexus/src/core/search/fts-indexes.ts:70` (`await createFTSIndex(table, indexName, [...properties], stemmer)`). `bm25-search.test.ts` was **not** in the PR diff; its assertion still maps to 3-tuples:
  ```ts
  // bm25-search.test.ts:37-39
  expect(vi.mocked(createFTSIndex).mock.calls).toEqual(
    FTS_INDEXES.map((i) => [i.table, i.indexName, [...i.properties]]),
  );
  ```
  The mock at line 10 silently accepts the extra arg, but `mock.calls` now records 4-tuples (`[..., 'porter']`) → Vitest deep-equality fails. **[CI-confirmed]** — the coverage job already failed on exactly this (1 file failed / 615 passed, per the test-ci lane and the CI log).
- **Fix:** add `'porter'` as the 4th element at `bm25-search.test.ts:38`:
  ```ts
  FTS_INDEXES.map((i) => [i.table, i.indexName, [...i.properties], 'porter']),
  ```
  `'porter'` is correct: this test sets no env stub, so `getSearchFTSStemmer()` returns `DEFAULT_FTS_STEMMER`.
- **Root cause:** the PR-body test plan ran **only** the touched `fts-indexes.test.ts`, not the full suite. `bm25-search.test.ts` is a **known ripple site** for `fts-indexes.ts` — it was already updated as a coupled file during the #2299/#2300 description-indexing work. The author correctly updated the sibling `fts-indexes.test.ts` but missed this one. **Recommendation: run the full `gitnexus` unit suite, not just the touched file, before pushing.**
- **Corroboration: STRONG (cross-engine).** Claude swarm/test-ci (P0, conf 100), Claude swarm/risk (P1), Claude ce/testing (P0, conf 100), Claude ce/maintainability (T1 BLOCKING), **and Codex ce (B2)** which independently reached `bm25-search.test.ts:37 P1 conf 100` with the same fix. Plus claude-mem prior-session obs #44183/#44189.

---

### 2. Recommended follow-up (P2 · single-engine Claude · concrete code-traced trigger · postable)

**`GITNEXUS_FTS_STEMMER` is validated only at the FTS phase (~85% of analyze), not at startup — a typo aborts a near-complete run and forces a full rebuild.**

- **Trigger:** `getSearchFTSStemmer()` (the only validator) has exactly one caller — inside `createSearchFTSIndexes` (`fts-indexes.ts:55`), invoked at `run-analyze.ts:1229` ("Phase 3: FTS, 85-90%") and `:678` (`--repair-fts`). There is **no startup preflight** (analyze early-validates `--branch`/`--default-branch`, but not this env var). Setting `GITNEXUS_FTS_STEMMER=porterr` in a shell profile and running `gitnexus analyze` on a large repo runs the full parse/scope-resolution/graph-COPY work, then throws at ~85%.
- **Cascade (ce-adversarial, [code-read], conf 80):** the throw reaches the outer catch (`run-analyze.ts:1609`) which checkpoints then **rethrows** (`:1630`); the final `saveMeta` (~`:1471`) never runs. On a re-analyze the `incrementalInProgress` dirty flag stays set, so the crash-recovery gate (`:715`) **forces the next run into a full rebuild**. Net: a one-character env typo discards an ~85%-complete analyze, and *every* subsequent run keeps failing at 85% until the operator notices. The repair path's own comment (`run-analyze.ts:660-666`, "Gate ... BEFORE touching any index ... Fail loudly first") articulates the opposite, fail-fast philosophy.
- **Fix:** call the already-exported, side-effect-free `getSearchFTSStemmer()` once early in `runAnalyze` (and/or at CLI arg-parse for `analyze`) so a bad value fails in milliseconds. No interface change. **Blocks merge: NO** — recommendation.
- **Corroboration: MODERATE (Claude-only: swarm/risk + ce/correctness R2 + ce/adversarial F1).** No Codex confirmation (the live Codex ce leg scoped to the test break only). Kept on severity + concrete code-traced trigger.

---

### 3. Minor / maintainability (P2–P3 · body-text, not inline)

- **`DEFAULT_FTS_STEMMER = 'porter'` (`fts-indexes.ts:4`) duplicates the adapter's inline defaults** at `lbug-adapter.ts:2348` and `:2444`. In sync today (and the analyze path always passes an explicit stemmer, so the adapter default is dead on that path), but the literals aren't linked — change one and the analyze path silently diverges from the read-only MCP `ensureFTSIndex` path. *Suggest:* export one `DEFAULT_FTS_STEMMER` const from the adapter and import it. (ce/maintainability M1 conf 100; ce/correctness R1.)
- **The 28-name `SUPPORTED_FTS_STEMMERS` allowlist is a static mirror with no version-source comment.** It is byte-for-byte correct for the installed extension *today* (see §4), but it can drift on a future LadybugDB minor. *Suggest (minimum):* a comment citing the `@ladybugdb/core` version the list was sourced from; *better:* a CI guard test that derives the allowlist from the installed extension so a dependency bump fails loudly. (ce/maintainability M2 conf 75; risk Finding 4; ce/adversarial R1.)
- **P3 doc/UX nits (non-blocking):** `GITNEXUS_FTS_STEMMER` is not surfaced in `gitnexus analyze --help` / the CLI i18n strings, only in the READMEs (maintainability R1); and in `gitnexus/README.md:366` it was added to the *extension-install* troubleshooting table rather than a general env-var table — though this follows the pre-existing `GITNEXUS_WAL_CHECKPOINT_THRESHOLD` precedent, not a new violation (project-standards PS-001 conf 75).

---

### 4. Refuted suspicions — validation is a feature, credit it

- **Cypher injection via `GITNEXUS_FTS_STEMMER` — REFUTED.** The sink at `lbug-adapter.ts:2365` interpolates `stemmer := '${stemmer}'` unescaped (pre-existing), but `getSearchFTSStemmer()` does `.trim().toLowerCase()` then `SUPPORTED_FTS_STEMMERS.has(raw)` against a closed set of 28 plain `[a-z]+` names, returning `raw` *only* on a `.has()` hit — so a hit **is** proof the value is a metacharacter-free safe name. `toLowerCase()` is locale-independent (no Turkish-ı fold); `Set.has` is SameValueZero (no prototype pollution). The PR test feeds `"none'); DROP TABLE File; --"` and asserts rejection with zero adapter calls. *Refuted by both* swarm/security-boundary (HIGH conf) *and* ce/security (independent trace) — Claude cross-lane agreement; CodeQL also passed clean.
- **Allowlist-drift as a *current* bug — REFUTED.** ce/adversarial extracted the installed FTS extension binary (`/home/node/.lbdb/extension/0.17.0/linux_amd64/fts/libfts.lbug_extension`, matching the pinned `@ladybugdb/core@^0.17.0`) and confirmed its 28 lowercase stemmer tokens are **byte-for-byte** the PR's allowlist. The PR is a strict subset (omits `armenian`, which appears only capitalized = a Unicode *script* name, not a stemmer) — never a superset, so there is no "passes allowlist, fails CREATE" case today. (This also downgrades ce/correctness F1's `armenian` lead to a non-issue.) Residual is **version-drift only** (the caret range), captured in §3.
- **Drop-then-fail data loss & stale-index coexistence — REFUTED.** The drop-then-fail mechanism is real (`dropFTSIndex` swallows errors + evicts the cache key in `finally`), but validation throws *before* the loop (test asserts zero drops), and on the full path the DB is wiped in Phase 2 so Phase-3 drops are no-ops; a stemmer change cleanly replaces each index (drop keyed on the same `indexName`).

---

### 5. What's solid (credit)

- **Backward-compatible:** unset/blank/whitespace env → `porter`; the 4th-arg addition is safe because `createFTSIndex` already defaults `stemmer='porter'`, so all other 3-arg callers are unaffected.
- **Validation provably precedes any index mutation:** `getSearchFTSStemmer()` is the first statement of `createSearchFTSIndexes`, before the drop/create loop — an invalid value throws before any drop, so `--repair-fts` can never destroy pre-existing indexes on a bad stemmer (test locks this with `expect(calls).toEqual([])`).
- **Injection mitigated and tested** (the DROP-TABLE payload is rejected); **production routing correct** on both the full-analyze and `--repair-fts` paths.
- **Project-standards clean:** no CHANGELOG touched (release-owned), no `any`/`as any`, no `if`-branching in tests, mock signatures match the real `createFTSIndex` types, `--repair-fts` is a real flag, and the new `fts-indexes.test.ts` assertions are strong (non-vacuous `toEqual` against fully-enumerated expectations).

---

### 6. CI status

**Red — blocking.** The coverage / unit job fails on exactly one file: `test/unit/bm25-search.test.ts` (1 file failed / 615 passed; 1 test failed / 13196 passed). This is the §1 headline and is fully explained; the §1 one-line fix turns it green. CodeQL passed clean (corroborates the injection refutation). The ripple audit found **no other** failing test files (`run-analyze-fts-repair.test.ts` mocks `createSearchFTSIndexes` wholesale; the integration tests use the real adapter via the `porter` default).

---

*This is an automated multi-tool digest (Claude swarm + Claude ce + a partially-live Codex GPT-5.5 leg). Verify the findings against the code before acting — especially the single-engine §2 follow-up.*

gitnexus/src/core/search/fts-indexes.ts
```diff
@@ -23,7 +67,7 @@ export async function createSearchFTSIndexes(
     // runs inside the existing FTS phase. Gate on a stored schema fingerprint if
     // this rebuild cost ever shows up in analyze profiles.
     await dropFTSIndex(table, indexName);
-    await createFTSIndex(table, indexName, [...properties]);
+    await createFTSIndex(table, indexName, [...properties], stemmer);
```
magyargergo, line 97: **[P1 · merge-blocking · CI-confirmed]** This new 4th `stemmer` arg is a breaking change for a sibling test that the PR didn't update.

`bm25-search.test.ts:37-39` still asserts 3-arg calls:
`FTS_INDEXES.map((i) => [i.table, i.indexName, [...i.properties]])`
After this line, `vi.mocked(createFTSIndex).mock.calls` records 4-tuples (`[..., 'porter']`), so the `toEqual` deep-equality fails deterministically — this is the CI failure (coverage job: 1 file failed / 615 passed). The mock accepts the extra arg silently; only the assertion breaks.

`bm25-search.test.ts` is a known ripple site for `fts-indexes.ts` (it was last touched for the #2299/#2300 description-indexing work). The PR correctly updated `fts-indexes.test.ts` but missed this sibling because the test plan ran only the touched file.

**Fix** — update `bm25-search.test.ts:38` to:
`FTS_INDEXES.map((i) => [i.table, i.indexName, [...i.properties], 'porter']),`
(`'porter'` = the default `getSearchFTSStemmer()` returns when no env stub is set.) Please also run the full gitnexus unit suite, not just the touched file, before pushing. [CI-confirmed]

gitnexus/src/core/run-analyze.ts
```diff
@@ -547,6 +551,9 @@ export async function runFullAnalysis(
   const progress = (phase: string, percent: number, message: string) =>
     callbacks.onProgress(phase, percent, message);
 
+  // Validate operator-provided FTS config before the expensive parse/load phases.
+  getSearchFTSStemmer();
```
magyargergo: Could we refactor this into an initialise and get? so we can call initialise here and the get in other places.

PR #2309, opened 2026-06-26
fix(mcp): stabilize api_impact response shape for same-URL multi-verb routes (#2308)

### LINKED ISSUE #2308: Track api_impact response shape for same-URL multi-verb routes
## Context

Follow-up from #2302 after Route ids became method-aware.

That PR correctly allows same-URL multi-verb routes to exist as distinct Route nodes, for example `GET /api/orders` and `POST /api/orders`. One remaining API contract question is how `api_impact` should shape its response when the caller passes only the URL.

## Current behavior to verify

`LocalBackend.apiImpact()` currently unwraps a single result and wraps multiple results:

- `gitnexus/src/mcp/local/local-backend.ts:6642-6647`
- single match: returns the route object directly
- multiple matches: returns `{ routes, total }`

After #2302, a previously singular route lookup like `api_impact({ route: "/api/orders" })` can become multi-match solely because the same URL has multiple HTTP method exports. That changes the response envelope from the direct route object to `{ routes, total }`, even though the caller supplied the same URL.

There is existing coverage for broad multi-route matching:

- `gitnexus/test/unit/calltool-dispatch.test.ts:1426-1454`

But I did not see a test or documented contract for the method-aware same-URL case.

## Why this matters

This is not necessarily a bug in #2302. It is a contract/design choice:

- keep the current polymorphic shape and document that same-URL multi-verb routes return the wrapped form;
- group results by URL while preserving per-method details;
- add a method parameter to let callers target one verb;
- always return a stable envelope from `api_impact`.

The main risk is downstream MCP/tool callers that assume a URL lookup returns a direct route object when the route string is exact.

## Suggested acceptance criteria

- Add a regression test for same-URL GET+POST `api_impact({ route })` behavior.
- Decide and document whether the response shape is intentionally polymorphic or should be normalized.
- If a method selector is added, make the no-method URL lookup behavior explicit.

Related PR: #2302

magyargergo:
## What & why

Closes #2308 (follow-up to #2302).

After #2302 makes Route node identity method-aware, the same URL exposes one Route node per HTTP verb — `GET /api/orders` and `POST /api/orders` become distinct nodes that share the bare URL as their `name`. That silently destabilized the `api_impact` MCP tool: a bare-URL lookup like `api_impact({ route: "/api/orders" })` flips from a **direct route object** (single match) to a **wrapped `{ routes, total }`** (multiple matches) the moment a URL gains a second verb. A caller reading `result.route` / `result.handler` gets `undefined` — a latent break that ships the day someone adds a `POST` next to an existing `GET`. The `file` lookup path breaks identically (one handler file → multiple verb nodes).

## Decision

The shape is now **a function of how many routes match the query, not of the indexed data** — and the polymorphism is documented rather than incidental:

- **Surface `method`** on every route the impact path returns. Previously both verbs came back as indistinguishable `name: "/api/orders"`; now each result self-describes its verb, so a caller handling the wrapped form can tell them apart.
- **Add an optional `method` selector** to `api_impact`. Passing `method` narrows a multi-verb URL/file to one verb. A URL/file that exists but has no route for that verb returns a distinct, clear error.
- **Keep** singular-on-single-match / wrapped-on-multi-match (rejected "always return an envelope" — it would break every existing single-route caller, including the `file`-param path, for no correctness gain).

This satisfies the issue's three acceptance criteria: regression coverage, a documented contract, and a method selector with explicit no-method behavior.

## Sequencing note (#2302 not yet on main)

This is intentionally a **follow-up to #2302**, which is still a sibling PR off the same parent. On `main` today, `routes.ts` still keys Route nodes by URL only, so a same-URL second verb is deduped at ingestion and the wrapped multi-verb path is not yet reachable. This change is **forward-compatible**: Route nodes already carry `method`, so the surfacing and the selector work today on the single node per URL; the wrapped multi-verb behavior activates automatically once #2302 lands. No schema bump here.

## Also fixed (review pass)

- `route_map` and `shape_check` now surface `method` too — the shared fetch already returned it and they were dropping it. `route_map` is the discovery tool agents use *before* `api_impact`, so it needs to show verbs.
- A method-scoped `api_impact` now still flags a multi-verb handler's partial-middleware warning (the verb count is taken from the full match, not the post-filter set).
- Tool description corrected: `method` narrows to one verb, but the singular shape only holds when exactly one route remains after filtering — a substring `route`/`file` match spanning several URLs can still wrap.

## Tests

`test/unit/calltool-dispatch.test.ts` — same-URL GET+POST via `route` and via `file` (wrapped, distinct methods); the `method` selector collapsing to the singular shape (case-insensitively); verb-not-found error; verbless (method-less) route surfacing `null` and being excluded from a selector; single-route backward-compat guard. `tsc`, `prettier`, and the `ai-context` / `skill-gen` descriptor snapshots all green.

## Out of scope / residual

- Method-aware matching in the group-layer `manifest-extractor` (queries Route by URL only — a documented deferral from #2302).
- Pre-existing: `apiImpact` returns `Promise<any>`; the `route`-or-`file` either-or constraint isn't expressed in the input schema.
- No real-LadybugDB integration test asserting `n.method` round-trips through the live query (unit tests mock the DB; #2302's `multi-verb-route-identity` integration test covers ingestion-side `method` population).

Plan: `docs/plans/2026-06-26-001-fix-api-impact-response-shape-plan.md`


gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -6521,21 +6530,28 @@ export class LocalBackend {
       queryParams.file = params.file;
     }
 
-    const routes = await this.fetchRoutesWithConsumers(repo.lbugPath, routeFilter, queryParams);
+    // After #2302 the same URL/handler can expose one Route node per HTTP verb.
+    // An optional `method` narrows to that one verb so the response collapses to
+    // the singular shape; verbless routes (null method) never match a selector.
+    const wantedMethod = params.method?.toUpperCase();
+    const matched = await this.fetchRoutesWithConsumers(repo.lbugPath, routeFilter, queryParams);
+    const routes = matched.filter((r) => !wantedMethod || r.method?.toUpperCase() === wantedMethod);
```
magyargergo: **P2 — wildcard `'*'` routes are excluded by the method selector (false negative).** Django function-based views (and any method-agnostic route) persist with `method = '*'`, **not** `null`: `inferHttpMethod` (django.ts:78) returns `'*'` for a view with no `.get/.post` suffix, `normalizeRouteMethod` (route-path.ts:50) preserves it, and `routes.ts:286` stores `{ method: '*' }`. Here `'*'.toUpperCase() === 'POST'` is `false`, so `api_impact({ route: '/orders/', method: 'POST' })` returns `"No routes found … with method \"POST\""` for a route that genuinely handles POST — a false negative in a tool whose whole purpose is impact safety.

Fix: `!wantedMethod || r.method === '*' || r.method?.toUpperCase() === wantedMethod`, and add a `verbRow('*', …)` regression test asserting a specific-method lookup returns the route, not an error.

_Found independently by both Codex legs (ce + adversarial); missed by all 9 Claude lanes; coordinator-verified end-to-end._ `[code-read]`

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -6265,11 +6267,16 @@ export class LocalBackend {
       const consumerName = row.consumerName ?? row[6];
       const consumerFile = row.consumerFile ?? row[7];
       const fetchReason: string | null = row.fetchReason ?? row[8] ?? null;
+      // Verb is absent (null) for method-less routes (filesystem, Laravel
+      // resource, Django wildcard). Appended last in RETURN so positional
```
magyargergo: **P3 — this comment is factually wrong, and it's the root of the wildcard bug.** Django wildcard routes persist as `method = '*'`, not `null`. Only filesystem routes and Laravel `resource`/`apiResource` are `null`; Django function views are `'*'` (see `inferHttpMethod` / `normalizeRouteMethod`). Correct this when fixing the selector. `[code-read]`

gitnexus/src/mcp/tools.ts
```diff
@@ -741,13 +741,18 @@ WHEN TO USE: BEFORE modifying any API route handler. Shows what consumers depend
...
 
-Returns: single route object when one match, or { routes: [...], total: N } for multiple matches. Combines route_map, shape_check, and impact data.`,
+Response shape is keyed on how many routes match, not on the data: exactly one match returns a single route object; two or more return { routes: [...], total: N }. The same URL can expose multiple HTTP verbs (e.g. GET and POST /api/orders are distinct routes that share the URL), so a bare-URL lookup may return the wrapped form — every route object carries its own "method" so verbs are distinguishable. Pass "method" to narrow to one verb; the single-object shape is returned only when exactly one route remains after filtering — a substring route/file match spanning several URLs can still return the wrapped form. A URL/file that exists but has no route for the given verb returns an error. Each route's "method" is null for method-less routes (filesystem, Laravel resource, Django wildcard). Combines route_map, shape_check, and impact data.`,
```
magyargergo: **P3 — the tool description repeats the same `null`-vs-`'*'` error.** "Django wildcard" is listed among routes whose `method` is `null`, but Django wildcards are `'*'`. Agents read this description as the response contract, so distinguish wildcard `'*'` (matches any verb) from verbless `null` (never matches a selector). `[code-read]`

magyargergo (commented):
## Tri-engine review — PR #2309 (`api_impact` multi-verb response shape, #2308)

**Engines:** Claude ran swarm (4 lanes) + ce-code-review (5 lanes) = 9 lanes; Codex (gpt-5.5/xhigh) ran swarm + ce + adversarial = 3 legs, all live with real findings. Base `c45d38f2` → head `cf97a646`, 3 files (~29 non-test lines).

### What's solid (validated across lanes)
- **Mechanically correct.** The positional fallback is safe — `n.method AS method` is appended LAST in the RETURN (index 9), so `row.consumerName ?? row[6]`, `row.consumerFile ?? row[7]`, `row.fetchReason ?? row[8]` keep their indices (correctness, maintainability, risk-architect, all [code-read]).
- **No injection.** `method` flows only into a JS array filter + the error string — never into the Cypher text; `route`/`file` stay parameterized via `$route`/`$file` (security-boundary, risk-architect, correctness — cross-engine with Codex swarm).
- **Backward-compatible.** The `method` key is purely additive (always present, incl. `null`); no whole-object `toEqual`/`toStrictEqual` assertion exists anywhere in unit or integration tests, so the new key breaks no consumer. Non-`method` callers are byte-identical (`matched === routes` when no selector).
- **Strong new unit tests.** 8 new tests with unconditional assertions (no `if`-branching), faithful mock keys matching `fetchRoutesWithConsumers`'s row parser so the real parse + filter run; wrapped-vs-singular shape, case-insensitive narrowing (URL + file), verb-not-found error, and `null`-method surfacing are all well covered (test-ci-verifier, ce-testing, correctness — 15 api_impact tests pass locally).

### Headline — P2: wildcard `'*'` routes are excluded by the method selector (false negative)
`local-backend.ts:6538` — `matched.filter((r) => !wantedMethod || r.method?.toUpperCase() === wantedMethod)`.

Django function-based views (and any method-agnostic route) persist with Route `method = '*'`, **not** `null`: `inferHttpMethod` (django.ts) returns `'*'` for a view with no `.get/.post` suffix, `normalizeRouteMethod` (route-path.ts) preserves `'*'`, and routes.ts stores `{ method: '*' }`. The new filter evaluates `'*'.toUpperCase() === 'POST'` → `false`, so `api_impact({ route: '/orders/', method: 'POST' })` returns `{ error: "No routes found ... with method \"POST\"" }` for a Django route that genuinely handles POST — a false negative in an impact-safety tool. Existing contract matching already treats a `'*'` provider as matching specific-method consumers, so the selector is inconsistent with the rest of the system.

- **Corroboration:** found independently by **both Codex legs** — ce (correctness + api-contract, confidence 100) and adversarial (medium, "no-ship"). **Missed by all 9 Claude lanes.** Coordinator-verified end-to-end [code-read].
- **Trigger:** any indexed Django project with function-based views + a `method` selector.
- **Fix:** include `r.method === '*'` in the selector predicate (treat wildcard as matching any requested verb); add a `verbRow('*', …)` regression test asserting a specific-method lookup returns the route, not an error.

### Linked — P3: tool docs wrongly say Django wildcard → `null`
The PR's own comment (`local-backend.ts:6270-6271`) and the `api_impact` description (`tools.ts:744`) both state `method` is "null for method-less routes (filesystem, Laravel resource, **Django wildcard**)". This is factually wrong — Django wildcard → `'*'`, not `null` — and is the root misconception behind the P2 bug. Fix both strings when fixing the filter.

### Remaining findings (P2 → P3)

| Sev | Finding | File:line | Found by |
|-----|---------|-----------|----------|
| P2 | No real-LadybugDB integration test for the new `n.method` query column — only unit-mocked; the e2e seed creates Route nodes without `method`. Mitigated by `route-method-roundtrip.test.ts` proving the column is queryable. | `fetchRoutesWithConsumers` / `test/fixtures/api-impact-seed.ts:23` | risk-architect (P2), Codex swarm L4-F2 (P3), docs-dod, ce-testing — **cross-engine** |
| P2 | `method` field added to `route_map` (6377) and `shape_check` (6445) responses has zero test coverage — no assertion anywhere; silent if dropped/renamed. | `local-backend.ts:6377,6445` | ce-testing F2, api-contract TG-1, test-ci-verifier |
| P3 | `route_map` + `shape_check` tool **descriptions** weren't updated for the new always-present `method` key — the MCP description is the machine-readable contract agents read. | `tools.ts` route_map ~673-678, shape_check ~713-719 | docs-dod F1, api-contract AC-1, project-standards RR-1 — **3 Claude lanes** [code-read] |
| P3 | `middlewareDetection: 'partial'`-under-method-filter (the diff's core behavioral line, `for (const r of matched)`) is untested: every `verbRow` hardcodes `middleware:null`, so reverting `matched`→`routes` leaves all tests green. Test gap, not a bug (behavior is correct). | `local-backend.ts:6554` | ce-testing F1 (rated P1; downgraded — it's a coverage gap), project-standards |
| P3 | Non-string `method` (e.g. `5`) → `TypeError` from `.toUpperCase()`. Caught by the MCP server try/catch → structured `{isError}`, no crash/leak. Schema declares `type:string`; consistent with repo's trust-the-schema convention for `route`/`file`. | `local-backend.ts:6536` | security-boundary D, correctness |
| P3 | Empty-string `method:""` silently acts as no-filter (`!"" === true`). | `local-backend.ts:6538` | risk-architect F2 |
| P3 | Verb suffix (`with method "GET"`) appears in the error even when the URL itself doesn't exist — can mislead a caller into thinking the URL exists with other verbs. Cosmetic. | `local-backend.ts:6540-6543` | correctness |
| P3 | `middlewareNote` can claim "other methods in this handler" when the sibling is actually a verbless (`null`) route, not another HTTP verb. Cosmetic. | `local-backend.ts:6553-6558, 6640-6646` | risk-architect F3 |
| P3 | PR **body** has stale sequencing facts ("#2302 not yet on main", "routes.ts still keys Route nodes by URL only") — #2302 is merged (commit 8bef64ba; `schema.ts:197` has `method STRING`), so this path is LIVE. Same misconception as the wildcard bug. | PR body | Codex swarm L1-F1 + L6-F1 |
| P3 | Pre-existing maintainability debt this PR touches (not introduced): identical anonymous route-shape type declared twice in `fetchRoutesWithConsumers`; `verbRow`/`ordersVerbRows` fixtures placed mid-describe. | `local-backend.ts:6205-6220 & 6242-6259`; `calltool-dispatch.test.ts:1460-1475` | maintainability MAINT-1/2 |

### Refuted / validated suspicions
- **No Cypher injection** via `method` — JS-filtered only; route/file parameterized (traced all write-sites).
- **Positional fallback integrity** — `row[9]` reads the right column; indices 0-8 unchanged.
- **Turkish-i case-fold hazard** — refuted: both sides use the same `toUpperCase` at the same locale, and HTTP verbs contain no `i`.
- **`routeCountByHandler` double-count** — refuted: dedup by route `id`, per-verb ids distinct; counting over `matched` is the intended behavior so `middlewarePartial` survives a method scope.
- **Additive `method` key non-breaking** — confirmed; no breaking contract change.

### CI
All checks green except `tests / ubuntu / coverage` = **PENDING** (runner availability). The claim that this is an unrelated Dart CFG visitor / worker-pool flake is supported — this PR touches no Dart/CFG files — but the coverage job must complete green before merge; if it fails, confirm the failure is unrelated before treating as benign.

### Verdict
- **Branch hygiene:** clean feature/fix PR — 3 scoped files, one harmless merge-from-main commit, no unrelated churn.
- **Merge state:** checks pending — `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` only because the coverage job hasn't finished.
- **Final:** **not production-ready (yet).** The core implementation is correct, well-tested for the non-wildcard cases, injection-free, and backward-compatible — but the P2 wildcard `'*'` false-negative is real (Codex-confirmed, coordinator-verified against the full `inferHttpMethod` → `normalizeRouteMethod` → `routes.ts` → filter chain) and defeats the impact-safety tool for Django function-based views whenever a `method` selector is passed. Fix the one-line selector predicate (`r.method === '*'`), correct the two `null`-vs-`'*'` doc strings, add a `verbRow('*', …)` regression test, and let the coverage job go green. None of the other items block.

---
*This is an automated multi-tool digest (Claude 9 lanes + Codex 3 legs). Verify findings against the code before acting — especially the P2 wildcard bug, which is the one merge-relevant correctness item.*

Issue #2175, opened 2026-06-12
MCP `query` tool: `query` parameter not transmitted by Claude Code —rename parameter as workaround


Syakaraka:
## Summary

  The `query` MCP tool's required parameter named `query` is consistently not transmitted when called from Claude Code
  (the most widely used MCP client). The server receives an empty `query` value and returns:

  "query parameter is required and cannot be empty."

  The same issue affects the `cypher` tool, which also has a required parameter named `query`.

  ## Verified: Server-side is correct

  Direct MCP protocol calls confirm the server handles `query` correctly —the parameter IS transmitted when bypassing
  Claude Code:

  | Call Method | `query` Param Received | Result |
  |-------------|----------------------|--------|
  | curl →remote HTTP MCP (`/api/mcp`) | ✅ Yes | ✅ Returns search results |
  | Node.js stdio →local MCP (`gitnexus mcp`) | ✅ Yes | ✅ Returns results |
  | Claude Code →MCP (stdio or HTTP via `mcp-remote`) | ❌ No | ❌ "query parameter is required and cannot be empty." |

  All tools with non-`query` parameter names work correctly from Claude Code:

  | Tool | Required Param | Works from Claude Code? |
  |------|---------------|------------------------|
  | `query` | `query` | ❌ |
  | `cypher` | `query` | ❌ |
  | `context` | `name` | ✅ |
  | `impact` | `target` | ✅ |
  | `rename` | `new_name` | ✅ |
  | `list_repos` | (none required) | ✅ |

  ## Root Cause

  The parameter name `query` conflicts with Claude Code's internal parameter handling. When Claude Code constructs MCP
  tool call arguments, the `query` key is dropped. This is reproducible across:

  - Local stdio MCP (tested with `gitnexus setup`)
  - Remote HTTP MCP via `mcp-remote` bridge
  - Multiple Claude Code versions (including latest)

  Since Claude Code is not open source, this cannot be reported upstream. The practical fix is on GitNexus's side.

  ## Suggested Fix

  Rename the affected parameters to avoid the conflict:

  - `query` tool: `query` →`search_query`
  - `cypher` tool: `query` →`statement`

  Files to update:
  - `gitnexus/src/mcp/tools.ts` —inputSchema definitions (lines 123, 217)
  - `gitnexus/src/mcp/local/local-backend.ts` —backend parameter references (lines 1282-1293, etc.)

  ## Impact

[...]

Issue #2106, opened 2026-06-09
Support multi-branch indexing and cross-branch search


NicolasKim:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Problem or opportunity

GitNexus currently indexes and queries repositories without clear support for multiple branches as independent, queryable states.

In real-world development workflows, multiple branches (e.g. `main`, feature branches, hotfix branches) often coexist for extended periods. These branches may contain meaningful divergences in code structure and call graphs.

At the moment, it is unclear whether:

* indexing is branch-specific or branch-agnostic
* search can be scoped to a specific branch
* comparison across branches is supported

This limits the ability to:

* reason about code evolution across branches
* perform branch-specific analysis (e.g. feature isolation)
* compare graph-level changes before merge

### Proposed solution

Introduce **multi-branch indexing and branch-scoped querying**:

1. **Branch-aware indexing**

   * Allow each branch to be indexed independently under the same repository
   * Maintain separate graph/index snapshots per branch

2. **Branch selection in queries**

   * Extend query model to optionally specify branch scope, e.g.:

     * repo + branch selector
   * Default behavior remains current branch or default branch (backward compatible)

3. **Cross-branch query capability (optional extension)**

   * Enable querying across multiple branches within the same repo
   * Example use cases:

     * search symbol existence across branches
     * compare graph differences between two branches

This approach keeps backward compatibility while enabling more advanced analysis workflows.

### Alternatives considered

* **External git checkout per branch indexing**

  * Requires manual orchestration outside GitNexus
  * Does not scale for multiple branches or frequent updates

* **Single unified index across all branches**

  * Loses branch-specific context
  * Introduces ambiguity in graph relationships

* **Relying on git diff tooling outside the system**

  * Does not leverage graph-based indexing benefits
  * Requires additional tooling integration

### Acceptance criteria

* Ability to index a repository per branch without overwriting existing branch data
* Ability to query a specific branch explicitly
* Existing single-branch workflow remains unchanged
* Clear separation of indexed data between branches
* (Optional) ability to query across multiple branches when explicitly requested

### Constraints

* Must preserve backward compatibility with current indexing behavior
* Should not significantly increase indexing cost for single-branch usage
* Should avoid breaking existing graph/query APIs
* Prefer minimal schema changes if possible (e.g., namespace or metadata-based separation rather than full redesign)

### Contribution

- [ ] I am willing to open a PR for this (may need design discussion first).

PR #2310, opened 2026-06-26
fix(cli): --limit i18n, 0/negative guard, and correct truncation paths

S23Web3:
## `--limit` flag: i18n, 0/negative guard, and correct truncation property paths

Adds a `--limit` (`-l`) flag to the `context`, `cypher`, and `detect-changes` CLI commands (`query` and `impact` already carry a limit on `main`), with matching help text in English and zh-CN, plus a `-q, --query` alias for the `query` command.

### Changes

- **i18n** — en + zh-CN strings for the new `--limit` options and the `-q, --query` alias.
- **Guard** — `Math.max(0, parseInt(n, 10))` clamps negatives; the truncation check is `if (limit !== undefined)` so an explicit `--limit 0` returns an empty payload instead of being silently dropped as falsy.
- **Correct truncation paths** (per-command result shapes):
  - `context` → `incoming.calls`, `outgoing.calls`, `outgoing.accesses`, `processes`
  - `impact` → `affected_processes`, `affected_modules`, `byDepth`
  - `cypher` → caps `row_count` (rows are embedded in the markdown string, so they cannot be spliced)
  - `detect-changes` → `changed_symbols`, `affected_processes`
- **Tests** — `cli-limit-e2e.test.ts`: end-to-end `--limit` truncation across all 5 commands (14 tests, green).

### Rebased onto `main`

Rebased onto the current `main`. Out-of-scope changes from the original branch were dropped: the `@ladybugdb/core` / TypeScript dependency bumps, the tree-sitter-kotlin native rebuild, and the skill-doc / `.codewhale/` / `CONTEXT.md` edits. Conflicts were resolved in favour of main's CLI evolution (`#2175` `search_query`/`statement` param names, `createLbugLazyAction`, the existing `impact` pagination, `--branch`, and the `check`/`trace` commands). The `impact -l` alias was dropped because `main` already registers `--limit` on that command.


gitnexus/src/cli/tool.ts
```diff
@@ -99,6 +102,7 @@ export async function contextCommand(
     process.exit(1);
   }
 
+  const limit = options?.limit ? Math.max(0, parseInt(options.limit, 10)) : undefined;
```
magyargergo: **[P1 · reproduced] `--limit` does no numeric validation → silently empties results.**
The gate is *string truthiness*, not numeric validity. `--limit abc` → `parseInt("abc",10)` = `NaN` → `Math.max(0, NaN)` = `NaN` → `NaN !== undefined` is `true` → `.slice(0, NaN)` returns `[]`. So `impact`/`context`/`detect-changes` return **empty results with exit 0, no error** — and these are the guardrail tools AGENTS/CLAUDE tell you to run before editing, so a typo'd limit reads as "nothing breaks." `--limit 0` (the string `"0"` is truthy) and `--limit -5` (`Math.max(0,-5)`=0) both also empty. Same parse is copy-pasted at `:83/:141/:190/:214`.
Fix once in a shared helper: `const n = Number.parseInt(s,10); return Number.isInteger(n) && n>0 ? n : undefined;` (or reject non-numeric with a clear error + non-zero exit) and apply uniformly across all 5 commands. Verified via node repro.

gitnexus/src/cli/tool.ts
```diff
@@ -107,6 +111,15 @@ export async function contextCommand(
     include_content: options?.content ?? false,
     repo: options?.repo,
   });
+    if (limit !== undefined) {
```
magyargergo: **[P2 · code-read] Over-indented block → fails `quality/format` (prettier) CI.**
This `if (limit !== undefined)` block sits at 4/6-space indentation (it reads as if nested inside the `callTool({...})` argument), but it's function-body scope where 2/4 spaces is correct — the other commands' limit blocks are indented correctly. `prettier --check` will fail the CI gate. Dedent to align with `output(result);`.

gitnexus/src/cli/tool.ts
```diff
@@ -152,31 +179,50 @@ export async function cypherCommand(
...
     query,
     repo: options?.repo,
   });
+  if (limit !== undefined) {
+    if (Array.isArray(result)) {
+      result.splice(limit);
+    } else if (result && typeof result === 'object' && typeof result.row_count === 'number') {
+      // Cypher returns {markdown, row_count} — rows are embedded in markdown string.
+      // We can't slice the markdown, but cap the reported row_count.
+      result.row_count = Math.min(result.row_count, limit);
```
magyargergo, line 300: **[P2 · reproduced] `cypher --limit` reports a row count it can't back.**
The cypher backend returns `{markdown, row_count}` (an object), so the `Array.isArray(result)` `splice` branch above is dead and only this line runs: it caps `row_count` but the markdown table is never sliced. Result: `cypher "..." --limit 2` over a 50-row result still prints all 50 rows but reports `row_count: 2` — a consumer bounding iteration on `row_count` is silently wrong. On bad input `row_count` becomes `NaN`. Either slice the parsed rows before formatting (keep header), push the limit into the query/backend, or drop `--limit` for cypher rather than fabricate a count.

gitnexus/test/integration/cli-limit-e2e.test.ts
```diff
@@ -0,0 +1,467 @@
...
+        if (Array.isArray(l)) limitedArr = l;
+        if (Array.isArray(b)) baselineArr = b;
+
+        if (limitedArr && baselineArr) {
+          expect(baselineArr.length).toBeGreaterThanOrEqual(limitedArr.length);
+        }
+      }
+    });
+
+    it('treats --limit 0 as no limit (falsy, skipped)', () => {
```
magyargergo: **[P1 · code-read] This test asserts the wrong behavior and passes only vacuously.**
Title says `--limit 0` is "falsy, skipped," but `"0"` is a *truthy* string, so the impl computes `limit=0` → `slice(0,0)` = `[]` (empties), not baseline-equal. The assertion is wrapped in `if (Array.isArray(...))` and the body is gated by `if (result.status === null) return` / `if (!data) return`, so it never fails — it just skips (violates the project's no-`if`-in-vitests / no-vacuous-pass rule). The whole file has this shape across all 14 `it()` bodies, and there is **no** test for `--limit abc` / `-5` / `""` — so the PR's headline "negative guard" is entirely unexercised, and the suite cannot catch the NaN bug it ships. De-vacuum (assert unconditionally) and decide + test the real `--limit 0` contract.

gitnexus/src/cli/i18n/en.ts
```diff
@@ -41,7 +41,7 @@ export const en = {
   'remove.removed': 'Removed: {{name}}',
   'remove.failed': 'Failed to remove {{name}}: {{message}}',
   'tool.noIndexed': 'GitNexus: No indexed repositories found. Run: gitnexus analyze',
-  'tool.usage.query': 'Usage: gitnexus query <search_query>',
+  'tool.usage.query': 'Usage: gitnexus query <search_query>  or  gitnexus query -q, --query <search_query>',
```
magyargergo: **[P2 · code-read] Malformed usage string.**
Two issues: (1) it shows `<search_query>` (required notation) although the positional is now optional (`query [search_query]`); (2) `-q, --query` is option-*definition* syntax, not an invocation — a user copying it verbatim passes a literal comma. Suggest: `Usage: gitnexus query [search_query]  or  gitnexus query --query <text>`. Same fix needed in `zh-CN.ts:48`.

CONTEXT.md
```diff
@@ -0,0 +1,82 @@
+# Context — GitNexus
+
+> Auto-generated 2026-06-06 via grill-with-docs. Source: `D:\repos\gitnexus\`
+
+## Domain Glossary
+
+| Term | Definition |
+|---|---|
+| **GitNexus** | Cross-repo code knowledge graph — indexes codebases into Neo4j |
```
magyargergo: **[P1 · code-read] Drop this file (and `.codewhale/`) — scope creep + factual errors.**
This line states GitNexus "indexes codebases into **Neo4j**" — it uses LadybugDB/Kuzu, not Neo4j (also lines 10, 14, 67). The file is a contributor's auto-generated personal context doc: it embeds machine-specific paths (`D:\repos\gitnexus`, `C:\Users\HP\.codewhale\...`), which trips the CONTRIBUTING pre-PR checklist, and lists private indexed repos. Along with `.codewhale/config.toml` (forces `deepseek-v4-flash` on any CodeWhale contributor who clones) and `.codewhale/mcp.json`, none of these belong in the PR — please remove them.

magyargergo (commented):
## PR #2310 review — `fix(cli): add i18n support and negative --limit guard`

Claude ran the GitNexus swarm (risk, branch-hygiene, test-ci, security) + ce-code-review personas (correctness, testing, maintainability, project-standards, adversarial, api-contract). **Codex (GPT-5.5) was attempted for all 3 legs but evaporated mid-run — this digest is Claude-only; no cross-engine corroboration.** Confidence rests on end-to-end node reproduction (the NaN and `--limit 0` paths) plus convergence across many independent Claude lanes; treat that convergence as correlated (same engine), not independent corroboration.

### What's solid (validated, no action needed)
- **i18n parity is clean.** `en.ts` and `zh-CN.ts` both received the same 5 new keys (`help.option.query.flag`, `context/impact/cypher/detectChanges.limit`) and the mirrored `tool.usage.query` update; all descriptions are single-line per the Issue #2106 zh-CN convention.
- **`gitnexus/CHANGELOG.md` untouched** — respects the feature-PR rule.
- **No security exposure.** Bidi scan 0, zero-width scan 0; non-ASCII is only CJK i18n + pre-existing em-dashes. No secrets, no injection, no eval/shell/template interpolation of user input, no unsafe deserialization.
- **The positional→`--query` fallback is correct.** `queryText?.trim() || options?.query?.trim()` — positional wins; falls back to `--query`; clean usage error + exit 1 when both empty. No bug.
- **`query` is immune to the `--limit` bug.** Its backend does `params.limit || 5` (`local-backend.ts:920`), coercing `NaN`/`0`/clamped-negative to the default 5.

---

### Headline findings

**1. (P1) Non-numeric / NaN `--limit` silently empties results on `context`/`impact`/`detect-changes`** — `[reproduced]`
Trigger: `gitnexus impact AuthService --limit abc` (also `Infinity`, `"  "`, `0x10`). The gate is string truthiness, not numeric validity: `options?.limit ? Math.max(0, parseInt(options.limit, 10)) : undefined`. `"abc"` is truthy → `Math.max(0, NaN)` = `NaN` → `NaN !== undefined` is `true` → `.slice(0, NaN)` returns `[]`, and `impact`'s `byDepth` loop rewrites every depth bucket to `[]`. Exit 0, no throw. These are the project's **safety/guardrail tools** (AGENTS/CLAUDE mandate running `impact`/`detect_changes` before edits): a fat-fingered `--limit` makes them report empty blast radius / no changed symbols with a clean exit, so an operator or agent reading the JSON concludes "nothing breaks." Evidence `tool.ts:105/141/214` (parse) + `:114-122`/`:148-160`/`:221-226` (slice). **`cypher` fails differently** — its backend returns `{markdown, row_count}` (not an array), so the `splice` branch is dead and instead `row_count` becomes `NaN` while all rows still print (see the dedicated cypher finding below); it does not empty. **`query` is NOT affected** (backend `|| 5`). Converged across correctness, adversarial, maintainability, api-contract, risk-architect, security, test-ci, testing lanes. Fix: parse once, validate numerically (`Number.isInteger(n) && n > 0`), reject or treat-as-no-limit uniformly across all five commands.

**2. (P1) Mergeability + scope-creep: PR is CONFLICTING and ~30% churn/already-merged** — `[code-read]`
GitHub `mergeStateStatus: DIRTY`, `mergeable: CONFLICTING`. Base is **253 commits behind main** with only 4 own commits. The genuinely-new substance is narrow (CLI `--limit` across 5 commands + i18n + a 467-line E2E test). Three scope-creep clusters must be dropped before this is reviewable:
- **Personal Windows tooling artifacts** (`.codewhale/config.toml`, `.codewhale/mcp.json`, `CONTEXT.md`) — tracked, not gitignored, unrelated to the i18n fix. `CONTEXT.md` embeds machine-specific paths (`D:\repos\gitnexus`, `D:\renaissance`, `C:\Users\HP\.codewhale\...`) — a direct hit on the CONTRIBUTING pre-PR checklist ("No secrets, tokens, or machine-specific paths committed") — and is **factually wrong about the architecture**: it states GitNexus "indexes codebases into Neo4j" / "Neo4j-backed graph" / "Cypher = Neo4j query language" (`CONTEXT.md:9,10,14,67`). GitNexus uses LadybugDB/Kuzu.
- **Skill-doc changes are stale and partly regressive** *(corrected after a per-file diff vs current `main`)* — only **2 of 6** are identical to `main` (`gitnexus-cli`, `gitnexus-impact-analysis` = pure churn / conflict noise). The other **4** (`debugging`, `exploring`, `guide`, `refactoring`) are the fork's **older** versions, so merging them would actively **regress** `main`: they revert the `query({search_query})` / `cypher({statement})` docs back to the old `query` param (undoing #2175) and **delete** the documentation for the `trace`, `explain`, `pdg_query`, and `check` tools that exist on `main` today (the guide loses 3 whole sections). None are necessary; 4 are regressions — drop all 6.
- **Unrelated dependency bumps** — `@ladybugdb/core ^0.16.1→^0.17.1` (runtime/graph-DB; main already at `^0.17.0`, so it collides rather than advances) + `typescript ^5.4.5→^5.9.3`, dragging 286 lines of lockfile churn. None required by the i18n change; per GUARDRAILS, dep changes belong in a dedicated `chore(deps)` PR with CI green.
- Note: commit `2640c63b` "rebuild tree-sitter-kotlin for Node 24 Windows" touches zero tree-sitter files (package.json/lock keep it at `^0.3.8`) and ships nothing in-tree — the commit subject mislabels its content.

---

### Findings table (severity-ordered)

| Sev | Blocks | Finding | File:line | Trigger | Tag |
|-----|--------|---------|-----------|---------|-----|
| P1 | yes | NaN/non-numeric `--limit` silently empties `context`/`impact`/`detect-changes` (slice on `NaN`); silent false-negative in guardrail tools | `tool.ts:105,141,214` (+114-122/148-160/221-226) | `impact X --limit abc` | reproduced |
| P1 | yes | Scope creep: personal `.codewhale/` + `CONTEXT.md` (machine paths, Neo4j factual error) | `.codewhale/config.toml`, `.codewhale/mcp.json`, `CONTEXT.md:3,9,24,67` | n/a (tracked additions) | code-read |
| P1 | yes | Test file passes vacuously — cannot catch the bug it ships | `cli-limit-e2e.test.ts:118,155,195,…,460` | `status===null return` + `!data return` + `if(Array.isArray)` guards | code-read |
| P2 | yes | `--limit 0` empties slice commands but PR's own tests assert "no limit"; same flag = 3 behaviors (query→5, slice→empty, cypher→count 0) | `tool.ts:105`; `cli-limit-e2e.test.ts:186,201` | `impact X --limit 0` | reproduced |
| P2 | maybe | `cypher --limit` caps `row_count` but cannot slice markdown rows → reported count desyncs from displayed rows; `NaN` on bad input | `tool.ts:196-204` (cap `:202`) | `cypher "…" --limit 2` over >2 rows | reproduced |
| P2 | yes | contextCommand limit block over-indented (4/6 spaces vs 2/4) → fails `quality/format` (prettier) CI | `tool.ts:114-122` | n/a | code-read |
| P2 | yes | Skill-doc: 2/6 identical to `main` (churn); 4/6 stale **reverts** that delete `trace`/`explain`/`pdg_query`/`check` docs + undo #2175 `search_query` rename | `.claude/skills/gitnexus/*/SKILL.md` | merge regresses main docs | code-read |
| P2 | maybe | Unrelated dep bumps bundled into a `fix` PR (`@ladybugdb/core`, `typescript`, +286 lock lines) | `package.json:59,109`; `package-lock.json` | n/a | code-read |
| P2 | no | `tool.usage.query` malformed: still shows required `<search_query>` (now optional) and writes `-q, --query` as if invocation (literal comma) | `en.ts:44`; `zh-CN.ts:48` | trigger usage error on empty `query` | code-read |
| P3 | no | radix-10 `parseInt` truncates: `--limit 1e9`→1, `12abc`→12, `0x10`→0 | `tool.ts:83,105,141,190,214` | `impact X --limit 1e9` | reproduced |
| P3 | no | `--limit N` applies per-array (context→4 arrays, impact→per-depth) so total items ≠ N; help text implies a unified cap | `tool.ts:114-122,148-160` | `context X --limit 5` | code-read |

---

### Refuted suspicions (validation is a feature)
- **Supply-chain risk of the new deps — REFUTED.** All transitive additions (apache-arrow ASF Apache-2.0, command-line-*, chalk, @swc/helpers) are official npmjs packages with SHA-512 integrity, no private registries, git-URLs, or suspicious postinstall. The dep *bumps* are still a scope/hygiene problem (above), not a security one.
- **Injection via `--limit` / `--query` — REFUTED.** The `NaN` reaches only in-process `slice`/`splice`; `query`'s value reaches FTS as a typed numeric param. No shell, eval, or string interpolation of user input.
- **`--limit ""` — SAFE no-op.** Empty string is falsy → `limit` stays `undefined` → the slice block is skipped entirely.
- **`--limit 99999999` — harmless over-slice.** `.slice(0, huge)` / `.splice(huge)` on shorter arrays returns/keeps the full array (effectively no cap). (Residual perf note only: `query --limit huge` amplifies to `processLimit * maxSymbolsPerProcess` into FTS/vector search — performance-reviewer territory, not a correctness bug.)
- **"no-any" violation — NOT POSTED.** It is not a citable repo standard (`tsconfig` `strict:false`, eslint `no-explicit-any: 'warn'`, no `--max-warnings 0`, no STANDARDS.md), and the new `tool.ts` code adds no new explicit `any` (the `any` flows from the pre-existing `callTool(): Promise<any>`).
- **`-l` short-flag collision / `query` positional-vs-`-q` ambiguity — REFUTED.** Each command registers `-l, --limit` exactly once (commander does not throw); positional-vs-`-q` is deterministic (positional wins).
- **Codex "No material findings" is NOT a refutation.** All 3 Codex legs evaporated (`stream disconnected before completion`); the trailing line is a rescue fallback, not a clearing verdict.

---

### CI status
No GitHub Actions ran — this is a fork PR **awaiting maintainer approval**, so the workflows are effectively un-run. The only check that executed is Vercel, which **auth-failed (normal for fork PRs)**. The contributor's "294 tests pass" is local-only, and the new 467-line E2E file is integration-tier and never executed in CI. Nothing here is CI-validated.

### Branch hygiene / mergeability verdict
- **Merge state:** `blocked by conflicts`
- **Branch hygiene:** `rebase/split required`
- **Final verdict:** `rebase/split required before final review`

`CONFLICTING` / `DIRTY`, 253 commits behind main, with ~30% of the diff being churn already on main. Recommended path: **rebase onto current `main`**, drop the 6 skill-doc edits, the `.codewhale/` artifacts, `CONTEXT.md`, and the dependency bumps; fix the prettier indentation and add real numeric validation; de-vacuum the tests. Cleanest outcome is a single focused CLI-`--limit` + i18n PR (the dep bumps, if wanted, in a separate `chore(deps)` PR).

---

*This is an automated multi-tool digest (Claude swarm + ce personas; Codex unavailable). Verify each finding at the cited `file:line` before acting.*

magyargergo:
**Correction / detail on the skill-doc (`.md`) changes** — follow-up to my tri-review above.

My original review called the 6 `.claude/skills/gitnexus/*/SKILL.md` edits "already on main verbatim." That was only half right. Diffing each file against **current `main`** (not the PR's stale base — this branch is 253 commits behind) shows they are **not necessary, and 4 of the 6 would regress `main`**:

| File | vs current `main` | Effect of merging |
|------|------|--------|
| `gitnexus-cli/SKILL.md` | identical | No-op churn (only adds a conflict) |
| `gitnexus-impact-analysis/SKILL.md` | identical | No-op churn |
| `gitnexus-debugging/SKILL.md` | **older** | Reverts `query({search_query})` / `cypher({statement})` → old `query` param (undoes #2175); **deletes** the `trace` tool section |
| `gitnexus-exploring/SKILL.md` | **older** | Reverts `search_query` docs → `query` |
| `gitnexus-refactoring/SKILL.md` | **older** | Reverts `search_query` docs → `query` |
| `gitnexus-guide/SKILL.md` | **older** | **Deletes** the `trace`, `explain`, `pdg_query`, `check` tool rows + 3 whole doc sections |

**Why:** the fork's SKILL.md copies predate the param rename (#2175) and the `trace` / `explain` / `pdg_query` / `check` tools — all of which exist on `main` today. Merging rolls those docs backward rather than adding anything.

**Recommendation:** drop all 6 `SKILL.md` edits (plus `CONTEXT.md` and `.codewhale/`); rebasing onto current `main` makes them unnecessary. The PR's keepable substance is the CLI `--limit` + i18n code and (after the fixes noted inline) the E2E test.

*Automated multi-tool digest (Claude lanes; Codex unavailable this run). Verify before acting.*

gitnexus/src/cli/tool.ts
magyargergo: Fixed in 3b71abf9: extracted a shared `parseLimit()` helper — `Number.isInteger(n) && n > 0 ? n : undefined` — used by all 5 commands. `--limit abc` / `0` / negative now resolve to `undefined` (no limit) instead of `NaN → slice(0, NaN) → []`. Added an E2E regression test asserting `--limit abc` ≡ no-limit (full result, not empty).

gitnexus/src/cli/tool.ts
magyargergo: Fixed in 3b71abf9: the markdown table is now sliced to `--limit` data rows (`markdown.split("\n").slice(0, 2 + limit)`, keeping the 2 header lines) before `row_count` is capped, so the reported count matches what is actually printed.

gitnexus/src/cli/i18n/en.ts
magyargergo: Fixed in 3b71abf9 (en + zh-CN): `Usage: gitnexus query [search_query]  or  gitnexus query --query <text>` — optional-positional notation and an invocation form, not the option-definition `-q, --query <search_query>` comma syntax.

gitnexus/test/integration/cli-limit-e2e.test.ts
magyargergo: Addressed in 3b71abf9: with `parseLimit()`, `--limit 0` now resolves to no-limit, so the "treats 0 as no limit" assertion holds for real rather than vacuously. Added a dedicated non-numeric `--limit` regression test next to it.

gitnexus/src/cli/tool.ts
magyargergo: Resolved: prettier-clean now — `prettier --check` passes on the changed files, and the `/autofix` bot pass (4607a44d) is already incorporated.

CONTEXT.md
magyargergo: Dropped during the rebase onto `main`: `CONTEXT.md` and `.codewhale/` are no longer part of the PR (along with the dependency bumps and the tree-sitter-kotlin rebuild).

gitnexus/src/cli/tool.ts
```diff
@@ -209,20 +252,41 @@ export async function cypherCommand(
...
+    if (Array.isArray(result)) {
+      // Non-tabular result: a raw row array.
+      result.splice(limit);
+    } else if (result && typeof result === 'object' && typeof result.row_count === 'number') {
+      // Tabular result: { markdown, row_count }. The markdown is a table built as
+      // [header, separator, ...dataRows].join('\n'), so slice it to `limit` data
+      // rows (keeping the 2 header lines) and report a row_count that matches what
+      // is actually printed — otherwise `--limit 2` over 50 rows prints all 50 but
+      // claims row_count: 2.
+      if (typeof result.markdown === 'string' && result.row_count > limit) {
```
magyargergo, line 294: **P2 — `cypher --limit` corrupts multi-line cells** `[reproduced]`

This slice assumes one physical line per row, but the backend builds cells with a bare `String(v)` (no `\n` escaping — `local-backend.ts:2318-2325`) and symbol nodes carry a multi-line `content` property that the backend itself returns (`local-backend.ts:1844`, `:2587` for `include_content`). So a realistic query —

```
cypher "MATCH (n:Function) RETURN n.name, n.content LIMIT 5" --limit 3
```

— splits one logical row across many physical lines, so `slice(0, 2 + limit)` keeps the **wrong** number of complete rows, often cutting the first row mid-cell and printing **zero** complete rows while `row_count` claims 3 (exit 0). An agent consuming the markdown gets a mangled table with a `row_count` that says everything is fine. This is the same count-vs-display desync class the PR fixed for *single*-line rows, re-introduced (worse) for multi-line cells. Both engines' adversarial passes independently reproduced it with the same `n.content` trigger (Codex: "do not ship yet").

**Fix:** slice the structured rows before `formatCypherAsMarkdown`, or escape `\n`→`\\n`/`<br>` in cells (also fixes the pre-existing un-`--limit`ed table corruption), or push `LIMIT` into the Cypher layer. Add a regression returning `n.content` that asserts the markdown holds exactly `limit` complete rows.

gitnexus/test/integration/cli-limit-e2e.test.ts
```diff
@@ -0,0 +1,507 @@
...
+describe('CLI --limit flag E2E', () => {
+  // ─── context ────────────────────────────────────────────────────────────
+
+  describe('context --limit', () => {
+    it('truncates callers/callees/processes to --limit 1', () => {
+      const result = runCliRaw(
+        ['context', 'validateInput', '--limit', '1', '--repo', 'mini-repo'],
+        MINI_REPO,
+      );
+      if (result.status === null) return;
```
magyargergo: **P2 — truncation tests can pass vacuously** `[code-read]`

This `if (result.status === null) return;` early-return, the `if (Array.isArray(...)) { expect(...) }` guards, and the bounds-only `toBeLessThanOrEqual` assertions violate **DoD.md:82** ("avoid `toBeGreaterThanOrEqual` and other bounds-only assertions that mask regressions"). Concretely: the fixture's `validateInput` has exactly **one** caller (`handler.ts:7`; `index.ts` only re-exports), so the context/impact/query `--limit 1` comparison tests assert `1 >= 1` and would stay green **even if the entire slicing block were deleted**; and `createFileChange()` appends a comment (no AST symbol), so the detect-changes tests likely run against 0 changed symbols. Only the line-197 `non-numeric --limit` regression test is unconditional and actually guards the shipped fix.

**Fix:** drop the `status===null`/null-parse early returns (assert the subprocess completed and produced the expected shape); replace conditional `expect`s with unconditional asserts; and make the fixture yield `> limit` items (or assert `baseline.length > limit`) so truncation genuinely bites.

gitnexus/test/integration/cli-limit-e2e.test.ts
```diff
@@ -0,0 +1,507 @@
...
+});
+
+afterAll(() => {
+  if (tmpParent) cleanupTempDirSync(tmpParent);
+  if (suiteGitnexusHome) cleanupTempDirSync(suiteGitnexusHome);
+});
+
+// ─── Tests ───────────────────────────────────────────────────────────────────
+
+describe('CLI --limit flag E2E', () => {
```
magyargergo, line 141: **P2 — this spawning suite isn't registered for Windows/macOS CI** `[code-read]`

This is a real `child_process.spawn` CLI test, but it's missing from the `SPAWN_CLI` array in `gitnexus/scripts/cross-platform-tests.ts:83` (which registers the sibling `cli-e2e.test.ts`). TESTING.md:91/96 classify "Process spawning / CLI" as a cross-platform category and mandate registration; the `npm run test:cross-platform` check only fails on *listed-but-missing* files, not the reverse — so CI stays green while the P1 `--limit` regression guard **never runs on Windows/macOS**, where path separators, CRLF, the `→`-arrow regex, and spawn semantics genuinely differ.

**Fix:** add `'test/integration/cli-limit-e2e.test.ts'` to `SPAWN_CLI` in `gitnexus/scripts/cross-platform-tests.ts`.

gitnexus/src/cli/tool.ts
```diff
@@ -189,6 +217,21 @@ export async function impactCommand(
       offset: parsedOffset,
       summaryOnly: options?.summaryOnly ?? undefined,
     });
+    // Client-side cap of the affected-list payload to --limit (parity with the
+    // other tool commands). The backend already paginates byDepth per level; this
+    // additionally bounds affected_processes/modules to the same cap.
+    if (parsedLimit !== undefined) {
```
magyargergo, line 242: **P2 — `impact --limit` silently expanded its contract** `[code-read]`

Before this PR, `impact --limit` only paginated the backend's `byDepth`. This block now also client-caps `affected_processes`/`affected_modules` to the same value — but the flag's help still says only "Max symbols per depth level (default: 100)" (`en.ts:254`). An existing eval/CI caller parsing `affected_processes` now silently gets a truncated array, and `affected_processes.length` disagrees with `summary.processes_affected` (the authoritative count). It's guardrail-**safe** (risk/`impactedCount`/`summary.*` are computed pre-slice from the full BFS), but it's an undocumented behavior change to a pre-existing flag. The `byDepth` re-slice just below is also a no-op — the backend already paginated `byDepth` with the same limit.

**Fix:** update the help/i18n to state `--limit` also caps the affected lists (or expose `total` vs `displayed` counts), and drop the redundant `byDepth` re-slice.

magyargergo (commented):
## PR #2310 review — `fix(cli): --limit i18n, 0/negative guard, and correct truncation paths`

**Engines:** Claude ran the GitNexus swarm (risk / test-ci / security) + ce personas (correctness, testing, maintainability, project-standards, adversarial, api-contract). Codex (GPT-5.5 / xhigh) ran swarm + ce + adversarial — **all 3 Codex legs live and produced real findings** (swarm verdict "production-ready with minor follow-ups"; ce "ready with fixes"; adversarial "needs-attention / do not ship yet"). This is **two independent engines**, not five independent confirmations — weight findings by whether *both* Claude and Codex hit them.

**Bottom line:** clean feature/fix PR (7 files, no merge commits, scope-clean rebase). **Merge state: `MERGEABLE`** — every code check (CI Gate, tests, typecheck, lint) is green; the only red is the Vercel "Authorization required to deploy" infra/auth gate, not a code failure. **The core `--limit` guard fix is correct and resolves the Jun-27 P1.** The one thing I'd fix before merging is the **new `cypher --limit` slicer's multi-line-cell corruption (P2 headline)** — both engines' adversarial passes independently reproduced it (Codex B3: "do not ship yet"); it's a defect in a *new* feature, not a regression. The remaining P2s are test-quality/coverage. Net recommendation: **address the cypher P2 (and de-vacuum the truncation tests) before merge; the rest are follow-ups.** Codex legs split B1 "production-ready w/ minor follow-ups" · B2 "ready with fixes" · B3 "needs-attention — do not ship yet."

---

### What's solid (credit where due)

- **The `parseLimit()` root-cause fix is correct** (tool.ts:71-75): `Number(raw); Number.isInteger(n) && n>0 ? n : undefined`. Every invalid/0/negative/non-numeric input → `undefined` → "show everything", never the old `slice(0,NaN) → []` silent-empty. Cross-engine confirmed (all Claude lanes + Codex B1/B3, B3 reproduced empirically across `0x10 / 1e9 / 3.5 / 0 / -1 / abc / 1_000`). **The headline P1 from the Jun-27 review is resolved.**
- **That fix is guarded by a robust regression test** — `cli-limit-e2e.test.ts:197` (`treats a non-numeric --limit as no limit`) is unconditional and would go RED if `parseLimit('abc')` regressed (confirmed Claude testing + test-ci + Codex).
- **Scope is clean** — 7 files, all under `gitnexus/src/cli/` + tests. The Jun-27 scope creep (`.codewhale/`, `CONTEXT.md`, dep bumps, stale-fork base) is fully gone after the rebase. No merge commits. `CHANGELOG.md` untouched (correct — owned at release).
- **i18n parity** — 4 new keys + `tool.usage.query` present in en/zh-CN, zh-CN single-line (project-standards + both engines).
- **Security clean** — no injection (`--limit`/`-q` reach only `slice`/`splice`/`Math.min`, never the Cypher statement); bidi/zero-width scans return 0; non-ASCII only benign zh-CN. Both engines.
- **Guardrail safety survives truncation** — `risk`, `impactedCount`, `summary.{direct,processes_affected,modules_affected,changed_count}` are computed pre-slice from the full BFS. `--limit` slices only the *lists*, never the *counts*, so **no blast-radius under-report** for a consumer reading the count/risk (verified Claude risk + adversarial [reproduced]).

---

### Headline finding — Cypher `--limit` corrupts multi-line-cell output (P2)

The PR's stated goal was "correct truncation paths." It fixed cypher's old `row_count`-vs-printed desync for single-line rows but **re-introduced a worse one for multi-line cells.** The new slice (tool.ts:281-287) does:

```ts
result.markdown = result.markdown.split('\n').slice(0, 2 + limit).join('\n');
result.row_count = Math.min(result.row_count, limit);
```

This assumes one physical line per row. But the backend builds cells with a bare `String(v)` and **no newline escaping** (`local-backend.ts:2318-2325`, verified), and GitNexus symbol nodes carry a multi-line `content` property that the backend itself returns (`local-backend.ts:1844`, `:2587` for `include_content`). So a realistic query —

```
cypher "MATCH (n:Function) RETURN n.name, n.content LIMIT 5" --limit 3
```

— splits one logical row across many physical lines, and `slice(0, 2+limit)` keeps the **wrong** number of complete rows, often cutting the first row mid-cell and printing **ZERO** complete rows while `row_count` claims 3, exit 0. An LLM/agent consuming the markdown gets a mangled, under-populated table with a `row_count` that says everything is fine.

**Corroboration: 5 lanes / both engines / reproduced twice.** Claude ce-adversarial F1 `[reproduced]` (the `n.content` repro), Claude ce-correctness FINDING-1, Claude ce-api-contract AC-2, Codex B1 swarm Lane 3, Codex B2 ce #1, Codex B3 adversarial `[reproduced]` (independently built the same `n.content` trigger → verdict "do not ship yet"). Both adversarial passes built the same trigger independently. **Fix:** slice structured rows before `formatCypherAsMarkdown`, or escape `\n`→`\\n`/`<br>` in cells (also fixes pre-existing un-`--limit`ed table corruption), or push `LIMIT` into the Cypher layer.

---

### Findings table (severity-ordered)

| Sev | Finding | file:line | Engines / lanes | Trigger | Tag | Inline? |
|-----|---------|-----------|-----------------|---------|-----|---------|
| **P2** | Cypher `--limit` slices rendered markdown by physical line → multi-line cells cut mid-row, fewer complete rows than `row_count` (often 0), exit 0 | `gitnexus/src/cli/tool.ts:281` (+ `local-backend.ts:2318` no-escape, `:1844`/`:2587` returns `n.content`) | **Both** — Claude adversarial+correctness+api-contract, Codex B1+B2+B3 | `cypher "MATCH (n:Function) RETURN n.name, n.content LIMIT 5" --limit 3` | `[reproduced]` ×2 | **YES** |
| **P2** | Truncation `it()`s use repo-banned `if(status===null)return` + `if(Array.isArray)` + bounds-only `toBeLessThanOrEqual` → vacuous; pass even if the slice block were deleted. Forbidden by **DoD.md:82** (tracked: "avoid bounds-only assertions that mask regressions"). Fixture's `validateInput` has ≤1 caller so context/impact/query `--limit 1` compare `1>=1`; detect-changes appends a comment → 0 changed symbols. (Only the line-197 regression test is robust.) | `cli-limit-e2e.test.ts:114` (+ 122-139, 346-353, 401-463) | **Both** — Claude testing(F1/F2)+maintainability(M-1)+test-ci+risk+correctness+project-standards, Codex B1 Lane 4 + B2 #4/#6 | slow CI → `status===null` → vacuous green; or fixture ≤ limit | `[code-read]` | **YES** |
| **P2** | New subprocess-spawning test not registered in `gitnexus/scripts/cross-platform-tests.ts` `SPAWN_CLI` (which lists the sibling `cli-e2e.test.ts` at :83). TESTING.md:96 mandates registration; the check only catches listed-but-missing, not the reverse → Windows/macOS CI never runs the P1 regression guard | `cli-limit-e2e.test.ts` (whole file) vs `gitnexus/scripts/cross-platform-tests.ts:83` | Claude project-standards (1 lane, **coordinator-verified**) | `npm run test:cross-platform` skips the file | `[code-read]` | **YES** |
| **P2** | `impact --limit` now **also** client-caps `affected_processes`/`affected_modules` (tool.ts:219-232) — undocumented behavior change to a pre-existing flag whose help says only "Max symbols per depth level" (en.ts:254). JSON consumers reading `affected_processes.length` get the capped number while `summary.processes_affected` holds the true count (two fields disagree). Safety-field-safe (no blast-radius under-report). | `gitnexus/src/cli/tool.ts:219` (+ en.ts:254) | Claude api-contract(AC-1)+risk (2 Claude lanes; Codex did not flag) | `gitnexus impact <sym> --limit N` + parse `affected_processes` | `[code-read]` | **YES (moderate)** |
| **P3** | detect-changes slices `changed_symbols`/`affected_processes` but `formatDetectChangesResult` header counts come from un-sliced `summary.changed_count` while the "+N more" derives from the **sliced** length → header says "50 symbols", body lists N with wrong/absent overflow marker. True total still in header → cosmetic, guardrail-safe | `gitnexus/src/cli/tool.ts:308` × `detect-changes-format.ts:44-64` | **Both** — Claude correctness(F2)+risk+testing(F5)+adversarial(F3), Codex B2 #3 | `detect-changes --limit 5` with 50 changed; or `--limit 20` → false "+5 more" | `[code-read]` | YES (low-pri) |
| **P3** | `impact` lacks the `-l` shorthand the PR adds to context/cypher/detect-changes → `impact -l 5` errors "Unknown option" while `context -l 5` works | `gitnexus/src/cli/index.ts:362` | Claude risk (1 lane) | `gitnexus impact <sym> -l 5` | `[code-read]` | body |
| **P3** | `context --limit` slices `incoming.calls`/`outgoing.calls`/`outgoing.accesses`/`processes` but leaves other array categories (e.g. `typed_properties`) unbounded | `gitnexus/src/cli/tool.ts:141` | Codex B2 #2 (1 lane, Codex-only) | `context <sym> --limit 1` on a symbol with typed props | `[code-read]` | body |
| **P3** | Maintainability nits: redundant `result.incoming?.calls && Array.isArray(...)` double-guard (Array.isArray already subsumes truthiness; line 147 uses the concise form); `parsedOffset` left on legacy `parseInt`/`Number.isFinite` idiom while limit migrated to `parseLimit` — two contradictory idioms adjacent | `gitnexus/src/cli/tool.ts:141` & `:192-194` | Claude maintainability (M-2/M-3, 1 lane) | n/a (style) | `[code-read]` | body |

---

### Lower-priority / refuted suspicions ("validation is a feature")

- **`query` positional `||` `--query` precedence — REFUTED as a defect.** A non-empty positional shadows `--query` (tool.ts:89), so `query "stale" --query "real"` silently drops `--query` (exit 0). But this is **documented** ("alias for positional", positional wins) — a footgun, not a defect. Both Claude adversarial (F2) and risk agree. Not posted as a finding; optionally a one-line "warn when a non-empty `--query` is ignored" suggestion.
- **`parseLimit` `0x10`→16 / `1e9`→1e9 — harmless.** Hex/exponent accepted as caps; never empties, never crashes, matches old `parseInt("0x10")`=16. No regression. Refuted.
- **No injection — security clean, both engines.** `--limit`/`-q` validated values reach only array/string slicing and `Math.min`; never concatenated into the Cypher statement.
- **impact `byDepth` client re-slice — redundant/dead but harmless.** Backend already paginates `byDepth` with the same limit; the client `slice(0, parsedLimit)` is a no-op today (latent footgun if backend pagination changes, but not a current bug). Refuted as a finding.
- **Supply-chain — N/A** (no dependency changes).

### Test-coverage gaps (follow-up, not merge-blocking on their own)

- No cypher `--limit` test with a multi-line cell (`RETURN n.content`) asserting the markdown contains exactly `limit` complete rows — the existing case uses single-line `n.name` and only checks `row_count`, so it cannot catch the headline finding.
- No isolated `parseLimit()` unit test for the edge matrix (undefined/0/neg/NaN/"abc"/"3.5"/"1e9") — only exercised indirectly via subprocess.
- No negative / fractional `--limit` test; no detect-changes test with `changed_count > limit > 15` (mini-repo too small).
- detect-changes `createFileChange` appends a comment (0 AST symbols) → those `it()`s likely run against 0 changed symbols; mutate real symbol content to make them bite.

### CI status

All green **except Vercel "Authorization required to deploy"** — an infra/auth gate, **not a code failure**. `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE` solely because of that Vercel check. The Jun-27 review (4584478630) is superseded by the rebase and should be dismissed at post time.

---

*Automated multi-tool digest (Claude swarm + ce, Codex GPT-5.5 swarm + ce + adversarial). Verify each finding against the diff before acting.*

PR #1877, opened 2026-05-28
feat(ingestion): resolve FastAPI include_router(prefix=...) cross-file routes

henry201605:
## Background & Problem

A very common FastAPI pattern: sub-route files declare paths via `APIRouter`,
and the entry file mounts the whole router under a prefix using
`app.include_router(<router>, prefix='/x')`.

```python
# users.py
from fastapi import APIRouter
router = APIRouter()

@router.get("/list")
def list_users(): ...

# main.py
from fastapi import FastAPI
from . import users
from .calls import router as calls_router

app = FastAPI()
app.include_router(users.router, prefix="/users")  # Shape A: attribute access
app.include_router(calls_router,  prefix="/calls") # Shape B: named import
```

The expected routes are `GET /users/list`, `GET /calls/list`, etc.

But today both of GitNexus's route-producing paths only see the bare `/list`
declared inside the sub-route file — the `prefix` from the entry file is
completely lost:

1. **Ingestion layer** — when writing `Route` graph nodes,
   `routes.ts` uses the decorator's `routePath` directly as the URL,
   with no mechanism to thread the cross-file `prefix` in.
2. **Group layer** — `HttpRouteExtractor` only scans decorators file-by-file;
   there is no "scan the repo first to build a prefix map" pre-pass,
   so `ExtractedContract` URLs are also bare `/list`.

The direct impact: cross-repo contract matching (HTTP provider ↔ HTTP consumer)
fails to find the corresponding provider — a frontend call to `/users/list`
ends up disconnected on the graph.

## Solution

Two independent but thematically aligned changes, both centered on
"thread the cross-file `include_router(prefix=)` info onto sub-route URLs."

### 1. Ingestion layer (so `Route` graph nodes carry the correct URL)

Add extraction & cross-file aggregation for "router imports / mounts":

- **`parse-worker.ts`**
  - `ExtractedDecoratorRoute` gains `decoratorReceiver?: string` and `prefix?: string | null`
  - New `ExtractedRouterInclude` / `ExtractedRouterImport` interfaces
  - `ParseWorkerResult` adds `routerIncludes` / `routerImports` (3 init sites + `mergeResult`)
  - New `extractFastAPIRouterBindings()`: scans `app.include_router(...)`
    (both attribute and named shapes) and `from <module> import router [as <alias>]`
- **`parsing-processor.ts`**
  - `WorkerExtractedData` / `mergeChunkResults` pass the two new fields through
- **`pipeline-phases/parse-impl.ts`**
[...]

PR #2312, opened 2026-06-27
fix(fastapi): apply APIRouter constructor prefixes

### LINKED ISSUE #2311: GitNexus does not capture in-file `APIRouter(prefix=...)` for FastAPI Python routes
### Area

### Summary

When a FastAPI router defines its URL prefix directly via `APIRouter(prefix="/gde-envs")` in the same file where `@router.get(...)` decorators are used, GitNexus does not capture the prefix. Routes are registered under `/` instead of `/gde-envs`, making `route_map` results inaccurate for the most common FastAPI project layout.

### Context

I'm indexing a FastAPI Python project with a typical router-file pattern:

```python
# app/routers/env_router.py
router = APIRouter(prefix="/gde-envs", tags=["gde-envs"])

@router.get("")
async def list_envs_gde():
    ...

@router.get("/{env_id}")
async def get_env_gde(env_id: str):
    ...
```

After running `gitnexus analyze`, I called `route_map({ route: "/gde-envs" })` and got zero results. Searching for `/` returned one result, confirming the route is partially captured but the prefix is lost.

Relevant code area:
- [`gitnexus/src/core/ingestion/route-extractors/fastapi-router-bindings.ts`](https://github.com/org/repo/blob/main/gitnexus/src/core/ingestion/route-extractors/fastapi-router-bindings.ts) — handles FastAPI prefix extraction but only for the cross-file `include_router(prefix=...)` pattern
- [`gitnexus/src/core/ingestion/workers/parse-worker.ts`](https://github.com/org/repo/blob/main/gitnexus/src/core/ingestion/workers/parse-worker.ts) — decorator route generation and prefix resolution

### Expected behavior

`route_map` should return routes with their full, prefix-included paths:

| Source code | Expected `route_map` path |
|---|---|
| `APIRouter(prefix="/gde-envs")` + `@router.get("")` | `/gde-envs` |
| `APIRouter(prefix="/gde-envs")` + `@router.get("/{id}")` | `/gde-envs/{id}` |
| `APIRouter(prefix="/api/v2")` + `@router.post("/items")` | `/api/v2/items` |

### Actual behavior

Only the decorator-level path is captured. The `APIRouter(prefix=...)` is silently ignored:

| Source code | Actual `route_map` path |
|---|---|
| `APIRouter(prefix="/gde-envs")` + `@router.get("")` | `/` |
| `APIRouter(prefix="/gde-envs")` + `@router.get("/{id}")` | `/{id}` |

`route_map({ route: "/gde-envs" })` returns `{ routes: [], total: 0 }`.

### Steps to reproduce

1. Create a minimal FastAPI project:

   ```python
   # app/routers/items.py
   from fastapi import APIRouter

   router = APIRouter(prefix="/api/items", tags=["items"])

   @router.get("")
   async def list_items():
       return []

   @router.post("")
   async def create_item():
       return {}
   ```

2. Run the GitNexus analyzer:

   ```bash
   NODE_ENV=development node .gitnexus/run.cjs analyze
   ```

3. Query the route map:

   ```
   route_map({ route: "/api/items" })
   // Expected: 2 routes
   // Actual:   0 routes

   route_map({ route: "/" })
   // Returns 1 route — the `@router.get("")` handler, but with path "/"
   // instead of "/api/items"
   ```

## Root cause

`extractFastAPIRouterBindings()` in `fastapi-router-bindings.ts` only detects the **cross-file** pattern:

```python
# ✅ Detected: include_router with explicit prefix
app.include_router(items.router, prefix="/api/items")
```

The **in-file** pattern is not handled:

```python
# ❌ Not detected: APIRouter constructor with prefix in the router file itself
router = APIRouter(prefix="/api/items", tags=["items"])
```

The function returns early when the file does not contain the literal string `include_router` (line 242):

```typescript
if (!content.includes('include_router')) return;
```

## Proposed fix direction

A lightweight same-file scan for `identifier = APIRouter(prefix="...")` in the Python worker path of `parse-worker.ts`, applied to `decoratorRoutes` entries whose `decoratorReceiver` matches the captured variable name. The `ExtractedDecoratorRoute.prefix` field already exists and `normalizeExtractedRoutePath` already supports prefix concatenation — no downstream changes are needed.

See also: the cross-file `include_router` prefix resolution in `parse-impl.ts` lines 1006–1144, which handles the complementary case but never runs for router files that lack `include_router` calls.

### Environment

_No response_

### Logs / screenshots

_No response_

koriyoshi2041:
Fixes #2311.

## What changed
- Capture same-file `router = APIRouter(prefix="...")` declarations in the FastAPI router binding pass.
- Apply that constructor prefix to `@router.<verb>` routes before any outer `include_router(prefix=...)` prefix is stacked.
- Keep the group HTTP extractor aligned with the ingestion route graph behavior.

## Verification
- `npm run build`
- `npm test -- test/unit/fastapi-router-bindings.test.ts`
- `npm test -- test/integration/fastapi-prefix-pipeline.test.ts`
- `npm test -- test/unit/group/http-route-extractor.test.ts`

## Risk
Low. The new handling is limited to literal FastAPI `APIRouter(prefix=...)` constructor prefixes and preserves the existing include-router prefix expansion path.

### DIFF
--- a/gitnexus/src/core/group/extractors/http-patterns/python.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/python.ts
@@ -6,6 +6,7 @@ import {
   unquoteLiteral,
   type LanguagePatterns,
 } from '../tree-sitter-scanner.js';
+import { normalizeExtractedRoutePath } from '../../../ingestion/route-extractors/route-path.js';
 import type { HttpDetection, HttpLanguagePlugin, RepoContext } from './types.js';
 
 /**
@@ -162,7 +163,7 @@ const INCLUDE_ROUTER_NAME_PATTERNS = compilePatterns({
   ],
 } satisfies LanguagePatterns<Record<string, never>>);
 
-const APIRouter_PREFIX_PATTERNS = compilePatterns({
+const API_ROUTER_PREFIX_PATTERNS = compilePatterns({
   name: 'python-fastapi-apirouter-prefix',
   language: Python,
   patterns: [
@@ -857,10 +858,6 @@ interface PythonRepoContext {
   prefixesByLongKey: Map<string, Set<string>>;
   /** stem only → set of prefixes (basename fallback, may collide) */
   prefixesByShortKey: Map<string, Set<string>>;
-  /** `<parent>/<stem>` → APIRouter(prefix=...) declared in that router file */
-  constructorPrefixesByLongKey: Map<string, string>;
-  /** stem only → APIRouter(prefix=...) fallback for root/single-segment files */
-  constructorPrefixesByShortKey: Map<string, string>;
 }
 
 /** Strip `.py` and return the bare basename (e.g. `api/users.py` → `users`). */
@@ -928,40 +925,19 @@ function buildPythonRepoContext(
 ): PythonRepoContext {
   const prefixesByLongKey = new Map<string, Set<string>>();
   const prefixesByShortKey = new Map<string, Set<string>>();
-  const constructorPrefixesByLongKey = new Map<string, string>();
-  const constructorPrefixesByShortKey = new Map<string, string>();
 
-  // Pre-pass over .py files. We deliberately run this even on files
-  // that don't contain `include_router` — the cost of an extra parse
-  // is bounded by the file count, and detecting `include_router`
-  // beforehand would require its own grep/scan.
+  // Cross-file pre-pass: only `include_router` sites need it — they bind a
+  // prefix declared in one file to a router defined in another. Same-file
+  // `APIRouter(prefix=...)` is resolved in scan() from the file's own tree, so
+  // APIRouter-only files are left out here and never parsed twice.
   for (const rel of files) {
     if (!rel.endsWith('.py')) continue;
     const src = readFile(rel);
     if (!src) continue;
-    if (!src.includes('include_router') && !src.includes('APIRouter')) continue;
+    if (!src.includes('include_router')) continue;
     parser.setLanguage(Python);
     const tree = parseSource(parser, src);
     if (!tree) continue;
-
-    // Same-file FastAPI router prefix:
-    //   router = APIRouter(prefix="/items")
-    //   @router.get("/{id}")  -> /items/{id}
-    // This stacks with include_router(prefix=...) later in scan().
-    for (const m of runCompiledPatterns(APIRouter_PREFIX_PATTERNS, tree)) {
-      const prefixNode = m.captures.prefix;
-      if (!prefixNode) continue;
-      const prefix = unquoteLiteral(prefixNode.text);
-      if (prefix === null) continue;
-      const longKey = fileLongKey(rel);
-      if (longKey) {
-        constructorPrefixesByLongKey.set(longKey, prefix);
-      } else {
-        constructorPrefixesByShortKey.set(fileShortKey(rel), prefix);
-      }
-    }
-
-    if (!src.includes('include_router')) continue;
 
     // Local name → (short, long) map for the current file, populated
     // from `from <module> import router [as <alias>]` statements. The
@@ -1049,17 +1025,14 @@ function buildPythonRepoContext(
   return {
     prefixesByLongKey,
     prefixesByShortKey,
-    constructorPrefixesByLongKey,
-    constructorPrefixesByShortKey,
   };
 }
 
 function joinPrefix(prefix: string, route: string): string {
-  // Mirror FastAPI's path joining: trim trailing slash off prefix,
-  // ensure exactly one leading slash on the result.
-  const p = prefix.replace(/\/+$/, '');
-  const r = route.startsWith('/') ? route : `/${route}`;
-  return `${p}${r}`;
+  // Delegate to the shared route-path normalizer so the group contract and the
+  // ingestion Route node join prefixes identically — one helper, no
+  // trailing-slash drift on empty routes (`APIRouter(prefix="/x")` + `@get("")`).
+  return normalizeExtractedRoutePath(route, prefix);
 }
 export const PYTHON_HTTP_PLUGIN: HttpLanguagePlugin = {
   name: 'python-http',
@@ -1121,6 +1094,18 @@ export const PYTHON_HTTP_PLUGIN: HttpLanguagePlugin = {
     // the ingestion route extractor), not a per-file source scan — see the note
     // at the top of this file.
 
+    // Same-file `router = APIRouter(prefix="/x")` (router-only). Read from this
+    // file's own tree, so there is no cross-file map and no prefix bleed across
+    // same-stem files; it stacks under any include_router(prefix=...) below.
+    let constructorPrefix: string | undefined;
+    for (const m of runCompiledPatterns(API_ROUTER_PREFIX_PATTERNS, tree)) {
+      const prefixNode = m.captures.prefix;
+      if (!prefixNode) continue;
+      const p = unquoteLiteral(prefixNode.text);
+      if (p === null) continue;
+      constructorPrefix = p;
+    }
+
     // Providers: FastAPI @router.<verb>("/path") — must be joined
     // with the prefix(es) declared at the include_router site. When
     // no prefix is found we still emit the unprefixed path so this
@@ -1145,16 +1130,8 @@ export const PYTHON_HTTP_PLUGIN: HttpLanguagePlugin = {
       const shortPrefixes =
         longPrefixes || !shortKey ? undefined : ctx?.prefixesByShortKey.get(shortKey);
       const prefixSet = longPrefixes ?? shortPrefixes;
-      // Constructor prefixes follow the same keying contract as
-      // include_router prefixes: the short-key map is only a fallback for
-      // repo-root/single-segment files. If `longKey` exists and has no
-      // constructor prefix, do not fall back or a root `users.py` prefix can
-      // bleed onto `admin/users.py`.
-      const constructorPrefix = longKey
-        ? ctx?.constructorPrefixesByLongKey.get(longKey)
-        : shortKey
-          ? ctx?.constructorPrefixesByShortKey.get(shortKey)
-          : undefined;
+      // Stack the same-file APIRouter(prefix=...) under any cross-file
+      // include_router prefix.
       const localPath = constructorPrefix ? joinPrefix(constructorPrefix, rawPath) : rawPath;
       const paths =
         prefixSet && prefixSet.size > 0
--- a/gitnexus/src/core/group/extractors/http-patterns/python.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/python.ts
@@ -1145,9 +1145,16 @@ export const PYTHON_HTTP_PLUGIN: HttpLanguagePlugin = {
       const shortPrefixes =
         longPrefixes || !shortKey ? undefined : ctx?.prefixesByShortKey.get(shortKey);
       const prefixSet = longPrefixes ?? shortPrefixes;
-      const constructorPrefix =
-        (longKey ? ctx?.constructorPrefixesByLongKey.get(longKey) : undefined) ??
-        (shortKey ? ctx?.constructorPrefixesByShortKey.get(shortKey) : undefined);
+      // Constructor prefixes follow the same keying contract as
+      // include_router prefixes: the short-key map is only a fallback for
+      // repo-root/single-segment files. If `longKey` exists and has no
+      // constructor prefix, do not fall back or a root `users.py` prefix can
+      // bleed onto `admin/users.py`.
+      const constructorPrefix = longKey
+        ? ctx?.constructorPrefixesByLongKey.get(longKey)
+        : shortKey
+          ? ctx?.constructorPrefixesByShortKey.get(shortKey)
+          : undefined;
       const localPath = constructorPrefix ? joinPrefix(constructorPrefix, rawPath) : rawPath;
       const paths =
         prefixSet && prefixSet.size > 0
--- a/gitnexus/src/core/group/extractors/http-patterns/python.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/python.ts
@@ -162,6 +162,26 @@ const INCLUDE_ROUTER_NAME_PATTERNS = compilePatterns({
   ],
 } satisfies LanguagePatterns<Record<string, never>>);
 
+const APIRouter_PREFIX_PATTERNS = compilePatterns({
+  name: 'python-fastapi-apirouter-prefix',
+  language: Python,
+  patterns: [
+    {
+      meta: {},
+      query: `
+        (assignment
+          left: (identifier) @router_name (#eq? @router_name "router")
+          right: (call
+            function: (identifier) @factory (#eq? @factory "APIRouter")
+            arguments: (argument_list
+              (keyword_argument
+                name: (identifier) @kw (#eq? @kw "prefix")
+                value: (string) @prefix))))
+      `,
+    },
+  ],
+} satisfies LanguagePatterns<Record<string, never>>);
+
 // `from .api.assistant import router` style — used together with
 // INCLUDE_ROUTER_NAME so we can map a local name back to its module
 // path, then back to the file the router was declared in.
@@ -837,6 +857,10 @@ interface PythonRepoContext {
   prefixesByLongKey: Map<string, Set<string>>;
   /** stem only → set of prefixes (basename fallback, may collide) */
   prefixesByShortKey: Map<string, Set<string>>;
+  /** `<parent>/<stem>` → APIRouter(prefix=...) declared in that router file */
+  constructorPrefixesByLongKey: Map<string, string>;
+  /** stem only → APIRouter(prefix=...) fallback for root/single-segment files */
+  constructorPrefixesByShortKey: Map<string, string>;
 }
 
 /** Strip `.py` and return the bare basename (e.g. `api/users.py` → `users`). */
@@ -904,6 +928,8 @@ function buildPythonRepoContext(
 ): PythonRepoContext {
   const prefixesByLongKey = new Map<string, Set<string>>();
   const prefixesByShortKey = new Map<string, Set<string>>();
+  const constructorPrefixesByLongKey = new Map<string, string>();
+  const constructorPrefixesByShortKey = new Map<string, string>();
 
   // Pre-pass over .py files. We deliberately run this even on files
   // that don't contain `include_router` — the cost of an extra parse
@@ -913,10 +939,29 @@ function buildPythonRepoContext(
     if (!rel.endsWith('.py')) continue;
     const src = readFile(rel);
     if (!src) continue;
-    if (!src.includes('include_router')) continue;
+    if (!src.includes('include_router') && !src.includes('APIRouter')) continue;
     parser.setLanguage(Python);
     const tree = parseSource(parser, src);
     if (!tree) continue;
+
+    // Same-file FastAPI router prefix:
+    //   router = APIRouter(prefix="/items")
+    //   @router.get("/{id}")  -> /items/{id}
+    // This stacks with include_router(prefix=...) later in scan().
+    for (const m of runCompiledPatterns(APIRouter_PREFIX_PATTERNS, tree)) {
+      const prefixNode = m.captures.prefix;
+      if (!prefixNode) continue;
+      const prefix = unquoteLiteral(prefixNode.text);
+      if (prefix === null) continue;
+      const longKey = fileLongKey(rel);
+      if (longKey) {
+        constructorPrefixesByLongKey.set(longKey, prefix);
+      } else {
+        constructorPrefixesByShortKey.set(fileShortKey(rel), prefix);
+      }
+    }
+
+    if (!src.includes('include_router')) continue;
 
     // Local name → (short, long) map for the current file, populated
     // from `from <module> import router [as <alias>]` statements. The
@@ -1001,7 +1046,12 @@ function buildPythonRepoContext(
     }
   }
 
-  return { prefixesByLongKey, prefixesByShortKey };
+  return {
+    prefixesByLongKey,
+    prefixesByShortKey,
+    constructorPrefixesByLongKey,
+    constructorPrefixesByShortKey,
+  };
 }
 
 function joinPrefix(prefix: string, route: string): string {
@@ -1095,10 +1145,14 @@ export const PYTHON_HTTP_PLUGIN: HttpLanguagePlugin = {
       const shortPrefixes =
         longPrefixes || !shortKey ? undefined : ctx?.prefixesByShortKey.get(shortKey);
       const prefixSet = longPrefixes ?? shortPrefixes;
+      const constructorPrefix =
+        (longKey ? ctx?.constructorPrefixesByLongKey.get(longKey) : undefined) ??
+        (shortKey ? ctx?.constructorPrefixesByShortKey.get(shortKey) : undefined);
+      const localPath = constructorPrefix ? joinPrefix(constructorPrefix, rawPath) : rawPath;
       const paths =
         prefixSet && prefixSet.size > 0
-          ? [...prefixSet].map((p) => joinPrefix(p, rawPath))
-          : [rawPath];
+          ? [...prefixSet].map((p) => joinPrefix(p, localPath))
+          : [localPath];
 
       for (const p of paths) {
         out.push({
--- a/gitnexus/src/core/ingestion/pipeline-phases/parse-impl.ts
+++ b/gitnexus/src/core/ingestion/pipeline-phases/parse-impl.ts
@@ -1212,8 +1212,11 @@ export async function runChunkedParseAndResolve(
     // without a corresponding import statement).
     const prefixesByLongKey = new Map<string, Set<string>>();
     const prefixesByShortKey = new Map<string, Set<string>>();
-    const constructorPrefixesByLongKey = new Map<string, Map<string, string>>();
-    const constructorPrefixesByShortKey = new Map<string, Map<string, string>>();
+    // Constructor prefixes are `router`-only (the apply gate below and the
+    // group-layer tree-sitter both pin to the literal name `router`), so a
+    // flat file-key → prefix map suffices — mirrors the group layer's shape.
+    const constructorPrefixesByLongKey = new Map<string, string>();
+    const constructorPrefixesByShortKey = new Map<string, string>();
 
     const recordPrefix = (target: Map<string, Set<string>>, key: string, prefix: string): void => {
       let set = target.get(key);
@@ -1222,20 +1225,6 @@ export async function runChunkedParseAndResolve(
         target.set(key, set);
       }
       set.add(prefix);
-    };
-
-    const recordConstructorPrefix = (
-      target: Map<string, Map<string, string>>,
-      key: string,
-      routerName: string,
-      prefix: string,
-    ): void => {
-      let byRouter = target.get(key);
-      if (!byRouter) {
-        byRouter = new Map();
-        target.set(key, byRouter);
-      }
-      byRouter.set(routerName, prefix);
     };
 
     for (const inc of allRouterIncludes) {
@@ -1297,19 +1286,9 @@ export async function runChunkedParseAndResolve(
       for (const ctor of allRouterConstructorPrefixes) {
         const longKey = fileLongKey(ctor.filePath);
         if (longKey) {
-          recordConstructorPrefix(
-            constructorPrefixesByLongKey,
-            longKey,
-            ctor.routerName,
-            ctor.prefix,
-          );
+          constructorPrefixesByLongKey.set(longKey, ctor.prefix);
         } else {
-          recordConstructorPrefix(
-            constructorPrefixesByShortKey,
-            fileShortKey(ctor.filePath),
-            ctor.routerName,
-            ctor.prefix,
-          );
+          constructorPrefixesByShortKey.set(fileShortKey(ctor.filePath), ctor.prefix);
         }
       }
 
@@ -1334,10 +1313,9 @@ export async function runChunkedParseAndResolve(
         // returns ''. Do not fall back from a missing long-key match to the
         // short key or a root `users.py` prefix can leak onto
         // `admin/users.py`.
-        const constructorPrefixesForFile = longKey
+        const constructorPrefix = longKey
           ? constructorPrefixesByLongKey.get(longKey)
           : constructorPrefixesByShortKey.get(fileShortKey(dr.filePath));
-        const constructorPrefix = constructorPrefixesForFile?.get(dr.decoratorReceiver);
         const routePath = constructorPrefix
           ? normalizeExtractedRoutePath(dr.routePath, constructorPrefix)
           : dr.routePath;
--- a/gitnexus/src/core/ingestion/pipeline-phases/parse-impl.ts
+++ b/gitnexus/src/core/ingestion/pipeline-phases/parse-impl.ts
@@ -1328,11 +1328,16 @@ export async function runChunkedParseAndResolve(
           ? undefined
           : prefixesByShortKey.get(fileShortKey(dr.filePath));
         const prefixes = longPrefixes ?? shortPrefixes;
-        const constructorPrefix =
-          (longKey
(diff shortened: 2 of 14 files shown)


gitnexus/src/core/ingestion/pipeline-phases/parse-impl.ts
```diff
@@ -1278,12 +1328,20 @@ export async function runChunkedParseAndResolve(
           ? undefined
           : prefixesByShortKey.get(fileShortKey(dr.filePath));
         const prefixes = longPrefixes ?? shortPrefixes;
+        const constructorPrefix =
+          (longKey
+            ? constructorPrefixesByLongKey.get(longKey)?.get(dr.decoratorReceiver)
+            : undefined) ??
+          constructorPrefixesByShortKey.get(fileShortKey(dr.filePath))?.get(dr.decoratorReceiver);
```
magyargergo: **[P1] Short-key constructor-prefix bleed — a new wrong route.** `constructorPrefixesByShortKey` is populated *only* by repo-root files (the `else` branch where `fileLongKey` is empty). A root `items.py` with `router = APIRouter(prefix="/root")` lands under short key `items`; a same-stem subdir `pkg/items.py` that declares no prefix of its own misses the long key and falls through here to `/root`, so `@router.get("/b")` is indexed as `/root/b` (actual FastAPI: `/b`). This turns a previously-correct route wrong. Reproduced on both the ingestion pipeline and the group extractor; GitHub Copilot independently flagged this line. Same short-key-collision class as #2281, and the new `/local` fixture doesn't exercise the root↔subdir case. Fix: suppress the short-key fallback when the file has a non-empty long key, and add a root + same-stem-subdir regression asserting the subdir route stays unprefixed on both layers. `[reproduced]`

gitnexus/src/core/group/extractors/http-patterns/python.ts
```diff
@@ -1095,10 +1145,14 @@ export const PYTHON_HTTP_PLUGIN: HttpLanguagePlugin = {
       const shortPrefixes =
         longPrefixes || !shortKey ? undefined : ctx?.prefixesByShortKey.get(shortKey);
       const prefixSet = longPrefixes ?? shortPrefixes;
+      const constructorPrefix =
+        (longKey ? ctx?.constructorPrefixesByLongKey.get(longKey) : undefined) ??
+        (shortKey ? ctx?.constructorPrefixesByShortKey.get(shortKey) : undefined);
```
magyargergo: **[P1] Same short-key bleed on the group/contract layer.** Mirror of the ingestion-side comment on `parse-impl.ts:1335` — a root file's `APIRouter(prefix=...)` (stored under the short key) attaches to a same-stem subdir file that declares no prefix, because the short-key fallback isn't gated on long-key presence. Copilot flagged the adjacent `joinPrefix` at line 1151. Both layers bleed identically, so fix them together to keep the graph Route node and the HTTP contract consistent. `[reproduced]`

gitnexus/src/core/ingestion/route-extractors/fastapi-router-bindings.ts
```diff
@@ -122,6 +128,8 @@ const INCLUDE_ROUTER_NAME_RE =
 // The latter is the common case and the only one we can map back to
 // a module stem.
 const FROM_IMPORT_ROUTER_RE = /^\s*from\s+(\.+|\.*[A-Za-z_][\w.]*)\s+import\s+([^#\n]+)/gm;
+const APIRouter_PREFIX_RE =
+  /\b([A-Za-z_]\w*)\s*=\s*APIRouter\s*\([^)]*?\bprefix\s*=\s*(['"])([^'"]*)\2/g;
```
magyargergo: **[P2] `[^)]*?` can't cross a nested `)` → prefix silently dropped, graph ≠ contract.** `router = APIRouter(dependencies=[Depends(get_db)], prefix="/y")` (the idiomatic router-level auth pattern) makes this regex return `null`: `[^)]*?` stops at the `)` of `Depends(...)` before reaching `prefix`. So the ingestion Route node stays unprefixed (`/x`) while the group tree-sitter contract correctly gets `/y/x` — a *new* Route-node-vs-HTTP-contract split (pre-PR both agreed). Since FastAPI routes are graph-authoritative for `api_impact` (#2308/#2309), the graph and the cross-service contract disagree. Reproduced via `runPipelineFromRepo` and direct `exec`; flagged by both Codex legs. Fix: balanced-paren scan, or (preferred) move ingestion constructor-prefix extraction onto the worker's tree-sitter pass so both layers share one extractor; add a `Depends(...)`-before-`prefix=` regression asserting Route == contract. `[reproduced]`

magyargergo (commented):
# Tri-review digest — PR #2312 "fix(fastapi): apply APIRouter constructor prefixes"

_Two-engine, three-method review. **Claude** ran the GitNexus production-readiness swarm (risk-architect, test-ci, security-boundary, facts-historian, branch-hygiene, docs-dod) + ce personas (correctness, adversarial, api-contract, performance, testing, maintainability, project-standards). **Codex (GPT-5.5/xhigh)** ran all three legs live with real findings: B1 GitNexus swarm Solo, B2 ce-code-review, B3 adversarial red-team. Every finding below was re-verified by reading the PR-head worktree; `[reproduced]` = a lane executed regex/tree-sitter/pipeline, `[code-read]` = traced statically._

## Posture: COMMENT (collaborative) — not approve/reject

This is a genuine, correct improvement for the happy path: literal `router = APIRouter(prefix="/x")` with prefix-first or paren-free args, stacking correctly with `include_router(prefix=...)` (`/v1/items/{item_id}`). It correctly follows the #1836 / #1877 precedent — the worker only harvests `routerConstructorPrefixes`, and the cross-file join runs **main-thread** in `parse-impl.ts`. Credit where due:
- **CI is green** — coverage 13527/13527, win/mac/ubuntu, CodeQL, gitleaks all pass. The sole `UNSTABLE` is the Vercel fork auth-wall (benign, not code).
- **Branch is clean** — single linear commit `038cd951` on the exact fork point `57e4afa4`, 12 files, one coherent domain, no CHANGELOG edit, no unrelated churn.
- **No security issues** — ReDoS on `APIRouter_PREFIX_RE` refuted (single lazy `[^)]*?`, O(k) per anchor); CSV/Cypher injection refuted (prefix is `[^'"]*`, RFC4180 CSV + parameterized Cypher); no hidden Unicode/bidi.
- **Closes issue #2311** ("GitNexus does not capture in-file `APIRouter(prefix=...)`") on merge. GitHub **Copilot** left a COMMENTED review (not approve) with 2 inline findings — both overlap finding **A** below. No human approver yet.

The findings are: one **new wrong-route** bug (A), one **graph-vs-contract divergence** (B), one **warm-cache correctness gap** that silently no-ops the fix on incremental re-index (C), and one **common-pattern silent miss + dead plumbing** (D), plus coverage gaps. **A, B and C are the strongest "should-fix-before-merge" items** — A and B both make the indexed route _wrong_ (A invents a prefix; B introduces a *new* Route-node-vs-HTTP-contract split the PR did not have before), and C silently disables the fix on warm re-index. **D is a consistent miss (both layers agree), so it can be fixed _or_ documented as an explicit `router`-only limitation with a test that pins the boundary.**

---

## A — HEADLINE · P1 · Short-key constructor-prefix BLEED across same-stem files → a NEW wrong route `[reproduced, both layers]`

**Problem.** The short-key fallback attaches a repo-root file's `APIRouter(prefix=...)` to a same-stem **subdir** file that declares **no prefix of its own** — inventing a prefix on a route that was previously correct. This is the only finding that turns a correct route *wrong* (not just incomplete).

**Sites:** ingestion `parse-impl.ts:1331-1335`; group `python.ts:1148-1150`.

**Trigger (reproduced both layers):**
```python
# items.py  (repo ROOT)      router = APIRouter(prefix="/root");  @router.get("/a")
# pkg/items.py (NO prefix)    router = APIRouter();                @router.get("/b")
```
Output (identical on both layers): `/root/a` (correct) **and `/root/b` (WRONG — actual FastAPI = `/b`)**.

**Mechanism (verified).** Root files have an empty long key → stored under short key `items` (`constructorPrefixesByShortKey`). `pkg/items.py` (long key `pkg/items`) has no constructor entry, so `constructorPrefixesByLongKey.get('pkg/items')?.get('router')` is `undefined`, and `?? constructorPrefixesByShortKey.get('items')?.get('router')` falls through to the root's `/root`. `normalizeExtractedRoutePath('/b','/root') = '/root/b'`.

**Accuracy note (for the critic):** the short-key fallback is the *same* mechanism the `include_router` path uses; both layers' fallbacks are structurally equivalent (`(longKey ? long : undefined) ?? short`). So this is not literally "a guard `include_router` has and the constructor dropped" — it's that the short-key fallback is **inherently collision-prone for same-stem files**, and the constructor feature introduces a **new instance** of the #2281 short-key-collision class. The PR's own guard test only covers the *long-key* no-bleed case (`api/users` vs `admin/users`, a file that has its own prefix); it does **not** cover the root↔subdir bleed for either include_router or constructor.

**Corroboration:** Claude ce-adversarial F3 [reproduced both layers, ran `HttpRouteExtractor.extract` + `runPipelineFromRepo`], facts-historian F1 [code-read], risk-architect (cross-domain), **+ GitHub Copilot independently flagged both exact lines** (`parse-impl.ts:1335`, `python.ts:1151`). Strong within-Claude + reproduced + external-tool corroboration. (Codex's three legs focused on B/C/D, not A — so this is *not* cross-engine to Codex; it stands on reproduction + Copilot.) Severity override applies (concrete reproduced trigger, makes a route wrong).

**Fix.** Gate the constructor short-key fallback so a file with a non-empty long key never reads the short-key map (mirror the include_router intent at file granularity), OR drop the short-key map entirely for constructors. Add a root + same-stem-subdir regression fixture asserting the subdir route stays unprefixed on **both** layers.

---

## B — P2 · Nested-paren regex drops the prefix → graph-vs-contract divergence `[reproduced]`

**Problem.** `APIRouter(dependencies=[Depends(x)], prefix="/y")` — the regex `[^)]*?` cannot cross the inner `)` of `Depends(x)`, so the ingestion `Route` node stays **unprefixed** while the group tree-sitter contract (immune to arg order/nesting) gets `/y/...`. Pre-PR both layers agreed on the (wrong-but-consistent) `/x`; the PR makes the group layer correct and the Route node still wrong → a **new** Route-node-vs-HTTP-contract split. FastAPI routes are graph-authoritative for `api_impact` (#2308/#2309), so the graph reports `/x` while the cross-service contract uses `/y/x`. `dependencies=[Depends(...)]` is the idiomatic FastAPI router-auth pattern.

**Site:** `fastapi-router-bindings.ts:131-132`. **Trigger:** `router = APIRouter(dependencies=[Depends(get_db)], prefix="/y")` + `@router.get("/x")` → GROUP `http::GET::/y/x`, INGESTION `Route.name = /x`, runtime `/y/x`.

**Corroboration (STRONG, cross-engine + reproduced):** Claude ce-correctness [reproduced], ce-adversarial F1 [reproduced via `runPipelineFromRepo`], ce-testing F2 [reproduced 100%, `exec(...) === null`], risk-architect, security (informational); **Codex B1 (risk-architect, blocks-merge), B2 (regex finding, conf 100), B3 (high)**.

**Fix.** Make the constructor harvester paren/bracket-balanced, or (preferred) move ingestion constructor-prefix extraction onto the worker's tree-sitter pass so both layers share one extractor. Add a `Depends(...)`-before-`prefix=` regression asserting Route node == HTTP contract.

---

## C — P1 · Parse-cache `SCHEMA_BUMP` not incremented → fix silently no-ops on warm re-index `[code-read]`

**Problem.** The PR adds `routerConstructorPrefixes?: ExtractedRouterConstructorPrefix[]` to `ParseWorkerResult` (`parse-worker.ts:398`) but `SCHEMA_BUMP` stays `8` (`parse-cache.ts:58`, **not in this PR's diff** — confirmed `git diff … -- gitnexus/src/storage/parse-cache.ts` is empty). On a warm-cache run that replays old `ParseWorkerResult[]` for unchanged Python files, the field is absent; the `?? []` guard (`parsing-processor.ts:129`) prevents a crash but yields **empty** constructor-prefix maps — so FastAPI routes whose prefix comes from `APIRouter(prefix=...)` are indexed with **no prefix** on cache-hit chunks. The cache header itself mandates this: *"any future change to the `ParsedFile` serialization shape MUST bump SCHEMA_BUMP."* `SCHEMA_BUMP` was last bumped at #2288 for exactly this kind of `ParseWorkerResult` field addition (`springTypes`).

**Corroboration (STRONG, cross-engine):** Claude ce-api-contract AC-1 (conf 75); **Codex B2 — both its api-contract subagent (P1) and its adversarial subagent independently raised it** (headline of Codex B2's findings table, P1/75).

**Fix.** Bump `SCHEMA_BUMP` 8 → 9 and update the comment to reference this PR. Add a warm-cache regression replaying a `ParseWorkerResult` missing `routerConstructorPrefixes`.

---

## D — P2 · Non-`router` receiver dropped by BOTH layers + dead per-routerName map `[reproduced + code-read]`

**Problem.** The regex captures **any** identifier (`([A-Za-z_]\w*)`) into `Map<routerName, prefix>`, but the apply-gate is `router`-only (`parse-impl.ts:1318`: `if (dr.decoratorReceiver !== 'router' …) continue`) **and** the group tree-sitter is `router`-only (`python.ts:173`: `#eq? @router_name "router"`). So `api_router = APIRouter(prefix="/api")` + `@api_router.get("/x")` is dropped by **both** layers (a consistent MISS, not a divergence), and the per-`routerName` second map dimension is **dead plumbing** — only the `router` slot is ever read.

**RESOLVED DISAGREEMENT:** ce-maintainability **M1** claims "the group layer emits the wrong (unprefixed) path while the INGESTION layer emits the CORRECT `/api/{id}`." That is **wrong** and is **not propagated**. I read the gate myself: for `@api_router.get`, `dr.decoratorReceiver === 'api_router' !== 'router'` hits the `continue` at `parse-impl.ts:1318` and pushes the route **unprefixed** *before* any lookup. ce-adversarial F4 [reproduced, full pipeline] and Codex B1/B3 confirm INGESTION drops it too. Correct framing: **both layers miss non-`router` names; the per-routerName map is dead.**

**Corroboration:** Claude ce-correctness (residual), ce-adversarial F4 [reproduced], ce-testing F1, risk-architect F2, facts-historian F3; **Codex B1 (risk-architect, blocks-merge, conf 100), B2 (receiver-gate, conf 100), B3 (high)**. Strong cross-engine + reproduced.

**Fix (pick one).** Either (a) document the `router`-only restriction in a comment and **narrow the regex** to `router` so the captured map matches what is applied, or (b) wire `dr.decoratorReceiver` through both layers (regex any-name + pin the group `@router_name`/`@obj` to the same set) and add `api_router` + two-routers-in-one-file tests. Do not relax the `!== 'router'` gate without first fixing A — doing so would activate the bleed across all router names.

---

## Lower-priority / body-only

- **E (P3, latent) — `joinPrefix` vs `normalizeExtractedRoutePath` duplication / trailing-slash.** `joinPrefix("/local","") → "/local/"` while `normalizeExtractedRoutePath` → `/local`; **masked today** by `normalizeHttpPath` stripping the trailing slash before the contractId, so no observable wrong output. Two join helpers will drift. (ce-correctness P3 [reproduced], ce-maintainability M3, facts-historian F5, Codex B1.) Suggest `joinPrefix` delegate to the shared normalizer.
- **F (P2/P3) — Coverage gaps.** No tests for B (nested-paren), C (warm-cache replay), D (non-`router` name, two routers/file), the root↔subdir bleed (A), empty `prefix=""`, or cross-layer agreement on one fixture (the two suites use disjoint fixtures, so a B-style divergence passes both). ce-testing F1-F4, Codex B1 lane-4.
- **G (P3, nit) — naming `APIRouter_PREFIX_PATTERNS` / `APIRouter_PREFIX_RE`** break the file-wide SCREAMING_SNAKE convention. ce-maintainability M2 (conf 100), **but ce-project-standards confirms NO citable rule** (no eslint naming-convention rule) → style nit, body-only, not inline.
- **H (P3) — perf double-parse.** The widened gate (`… && !src.includes('APIRouter')`) makes APIRouter-only files parse twice in the group phase (pre-pass + `getScanInput`, no shared cache); bounded by file count (~50-150ms / 100 files). ce-performance PERF-1 (conf 75, single lane).

## Validated / refuted (validation is a feature)

- **ReDoS** on `APIRouter_PREFIX_RE` — refuted (single lazy `[^)]*?`, no nested quantifier, O(k)/anchor). security-boundary, ce-performance, ce-adversarial.
- **CSV / Cypher injection** — refuted (prefix `[^'"]*`, RFC4180 CSV, parameterized Cypher reader). security-boundary.
- **Empty `prefix=""` / trailing-slash wrong output** — refuted; `""` is falsy → no-op consistently on both layers; trailing slash normalized away. ce-adversarial, ce-correctness, Codex B1.
- **RFC#909 "must not name languages" placement** — refuted; route extraction is an established language-named subsystem and `parse-impl.ts` is the documented home of the FastAPI/Django cross-file join. ce-project-standards (conf 80).
- **Branch hygiene / CHANGELOG / hidden Unicode** — clean. branch-hygiene, project-standards, security-boundary, Codex B1.
- **M1's "ingestion is correct for `api_router`"** — refuted by code-read + reproduction (see D).

_Note: two Claude swarm lanes (test-ci, docs-dod) ended without a final block (known gitnexus-lane behavior); their domains are covered cross-stream by ce-testing, ce-project-standards, Codex B1 lanes 4 & 6, and facts-historian — not left unreviewed._

---

## INLINE COMMENTS (candidates) — all anchors verified in-diff (RIGHT side)

**1. [P1] `gitnexus/src/core/ingestion/pipeline-phases/parse-impl.ts` — line 1335** (added: `constructorPrefixesByShortKey.get(fileShortKey(dr.filePath))?.get(dr.decoratorReceiver);`)
> **Short-key constructor-prefix bleed (new wrong route).** A repo-root `items.py` with `APIRouter(prefix="/root")` is stored under short key `items`; a same-stem subdir `pkg/items.py` that declares no prefix misses the long key and falls through here to `/root`, so `@router.get("/b")` is indexed as `/root/b` (actual FastAPI: `/b`). Reproduced on both the ingestion pipeline and the group extractor; GitHub Copilot flagged this exact line. Same short-key-collision class as #2281. Fix: gate the short-key fallback on long-key absence at file granularity, and add a root + same-stem-subdir regression. `[reproduced]`

**2. [P1] `gitnexus/src/core/group/extractors/http-patterns/python.ts` — line 1150** (added: `(shortKey ? ctx?.constructorPrefixesByShortKey.get(shortKey) : undefined);`)
> **Same short-key bleed on the group layer.** Mirror of comment 1 — the group contract also attaches a root file's constructor prefix to a same-stem subdir file with no prefix (Copilot flagged the adjacent `joinPrefix` at 1151). Both layers bleed identically; fix together so the graph node and HTTP contract stay consistent. `[reproduced]`

**3. [P2] `gitnexus/src/core/ingestion/route-extractors/fastapi-router-bindings.ts` — line 132** (added: the `APIRouter_PREFIX_RE` literal)
> **`[^)]*?` can't cross a nested `)` → prefix silently dropped, graph≠contract.** `router = APIRouter(dependencies=[Depends(x)], prefix="/y")` (the idiomatic router-auth pattern) makes this regex return `null`, so the Route node stays `/x` while the group tree-sitter contract correctly gets `/y/x`. Since FastAPI routes are graph-authoritative for `api_impact`, the graph and the cross-service contract disagree. Reproduced via `runPipelineFromRepo` and direct `exec`. Fix: balanced-paren scan, or move to the worker tree-sitter pass shared with the group layer. `[reproduced]`

**4. [P1] `gitnexus/src/core/ingestion/workers/parse-worker.ts` — line 398** (added: `routerConstructorPrefixes?: ExtractedRouterConstructorPrefix[];`)
> **New `ParseWorkerResult` field without a `SCHEMA_BUMP` bump → warm-cache silently empty.** This field is added to the worker result, but `SCHEMA_BUMP` stays `8` in `gitnexus/src/storage/parse-cache.ts:58` (real location — not in this diff). On a warm-cache re-index, replayed pre-PR chunks lack this field; the `?? []` guard avoids a crash but yields empty constructor-prefix maps, so the fix no-ops and routes are indexed with no prefix until a content change. The cache comment mandates bumping on any serialization-shape change (last bumped at #2288 for `springTypes`). Fix: bump to 9 + a warm-cache replay test. `[code-read]`

**5. [P2] `gitnexus/src/core/group/extractors/http-patterns/python.ts` — line 173** (added: `left: (identifier) @router_name (#eq? @router_name "router")`)
> **Constructor prefixes are `router`-only across BOTH layers + dead per-name map.** This `#eq?` pins the group capture to the literal name `router`; the ingestion apply-gate is also `router`-only (`parse-impl.ts:1318` `dr.decoratorReceiver !== 'router'`). So `api_router = APIRouter(prefix="/api")` + `@api_router.get(...)` is dropped by both layers (reproduced), and the ingestion regex's any-name `Map<routerName,prefix>` is dead — only the `router` slot is read. (Contradicts a maintainability note that claimed ingestion handles `api_router` correctly; it does not.) Fix: either document + narrow the regex to `router`, or wire the real receiver through both layers with `api_router`/two-routers tests. `[reproduced]`

---

_This is an automated multi-tool digest (Claude swarm + ce, Codex B1/B2/B3, GitHub Copilot cross-check) merged with cross-engine confidence weighting. Verify each finding against the worktree before acting; the inline anchors above were confirmed present on the RIGHT side of the PR diff._

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -394,6 +395,7 @@ export interface ParseWorkerResult {
   decoratorRoutes: ExtractedDecoratorRoute[];
   routerIncludes: ExtractedRouterInclude[];
   routerImports: ExtractedRouterImport[];
+  routerConstructorPrefixes?: ExtractedRouterConstructorPrefix[];
```
magyargergo, line 398: **[P1] New `ParseWorkerResult` field without a `SCHEMA_BUMP` → warm-cache silently empty.** This adds `routerConstructorPrefixes?` to the worker result, but `SCHEMA_BUMP` stays `8` in `gitnexus/src/storage/parse-cache.ts:58` (the real location — not in this diff). On a warm-cache re-index, replayed pre-PR chunks lack this field; the `?? []` guard in `parsing-processor.ts` avoids a crash but yields empty constructor-prefix maps, so the fix no-ops and FastAPI routes are indexed with no prefix until the file content changes. The cache header mandates bumping on any `ParseWorkerResult` serialization-shape change — `SCHEMA_BUMP` was last bumped at #2288 for exactly this (the `springTypes` field, per its own comment). Fix: bump `SCHEMA_BUMP` 8 → 9 and add a warm-cache replay regression. `[code-read]`

gitnexus/src/core/group/extractors/http-patterns/python.ts
```diff
@@ -162,6 +162,26 @@ const INCLUDE_ROUTER_NAME_PATTERNS = compilePatterns({
...
 
+const APIRouter_PREFIX_PATTERNS = compilePatterns({
+  name: 'python-fastapi-apirouter-prefix',
+  language: Python,
+  patterns: [
+    {
+      meta: {},
+      query: `
+        (assignment
+          left: (identifier) @router_name (#eq? @router_name "router")
```
magyargergo, line 174: **[P2] Constructor prefixes are `router`-only across BOTH layers, and the per-name map is dead.** This `#eq? @router_name "router"` pins the group capture to the literal name `router`; the ingestion apply-gate is also `router`-only (`parse-impl.ts:1318` `dr.decoratorReceiver !== 'router'` → `continue` before any lookup). So `api_router = APIRouter(prefix="/api")` + `@api_router.get(...)` is dropped by **both** layers (reproduced), and the ingestion regex's any-name `Map<routerName, prefix>` is dead plumbing — only the `router` slot is ever read. (A maintainability note claimed ingestion handles `api_router` correctly; the gate shows it does not.) Fix: either document + narrow the regex to `router` so the captured map matches what's applied, or thread `dr.decoratorReceiver` through both layers with `api_router` / two-routers-in-one-file tests. Note: don't relax the gate before fixing the short-key bleed, or it would activate the bleed across all router names. `[reproduced]`

magyargergo (commented):
## GitNexus Swarm Review — `fix(fastapi): apply APIRouter constructor prefixes` (closes #2311)

_Automated production-readiness swarm: 6 review lanes (facts/history, branch-hygiene, risk, security-boundary, test/CI, docs-DoD) + a synthesis-critic gate. Verify findings against the code before acting._

### 1. Review bar
Bug-fix extending FastAPI route extraction to same-file `router = APIRouter(prefix=...)`. Bar: the joined Route URL must be correct **and identical across the ingestion graph layer and the group contract layer** (FastAPI Route nodes are graph-authoritative for `api_impact`, #2308/#2309); no regression to existing routes; warm-cache invalidates; tests + standards (no CHANGELOG, no `any`, no if-branching in tests). Security bar: the prefix is untrusted source text reaching graph/contract sinks.

### 2. Problem being solved
#2311: `router = APIRouter(prefix="/gde-envs")` + `@router.get("")` indexed as `/`; `route_map({route:"/gde-envs"})` returned nothing. Pre-PR `extractFastAPIRouterBindings` only handled cross-file `include_router(prefix=...)` and bailed when a file lacked `include_router`.

### 3. Current PR state
OPEN, 7 linear commits, head `52817904`. The last two commits (`dcff2bc8`, `52817904`) address the prior tri-review. 14 files (7 source, 7 test/fixture).

### 4. Merge status and mergeability
Classification: **`review blocked`.** `mergeable=true` (no conflicts; 0 commits behind `main`). `mergeStateStatus=BLOCKED` solely from the missing approving review — CI is green on all platforms; the only red check is Vercel ("Authorization required to deploy"), a fork deploy-auth infra failure, not a code gate.

### 5. Repository history considered
#1877 (the include_router cross-file pipeline this extends); #2281 (short-key/long-key collision fix pattern); #2288/#2290 (last SCHEMA_BUMP 7→8 for `springTypes` — same "new ParseWorkerResult field → bump", now 8→9); #2308/#2309 (FastAPI Route nodes graph-authoritative for `api_impact`). `parse-impl.ts`'s FastAPI block is the historically fragile spot; the new logic sits adjacent.

### 6. Branch hygiene assessment
Classification: **`clean feature/fix PR`.** 7 commits, zero merge commits, single domain, no CHANGELOG edit, `git diff --check` clean, conventional title. No rebase/split required.

### 7. Understanding of the change
Same-file `APIRouter(prefix="/x")` is captured and stacked onto `@router.<verb>` on **both** layers. Ingestion: `fastapi-router-bindings.ts` (`API_ROUTER_ASSIGN_RE` + `findMatchingParen()` + `API_ROUTER_PREFIX_ARG_RE`) → new `routerConstructorPrefixes` worker field → result-merge/parsing-processor → applied in `parse-impl.ts` via flat maps + `normalizeExtractedRoutePath`, stacked under any include_router prefix. Group: `python.ts` `scan()` reads the prefix from the file's own tree; `joinPrefix` now delegates to `normalizeExtractedRoutePath`. Constructor support is intentionally `router`-only. SCHEMA_BUMP 8→9.

### 8. Findings

**P2 — Multiple `router = APIRouter(...)` reassignments in one file: last-write-wins**
- **Risk:** a file reassigning `router` applies the LAST prefix to all `@router.*` routes (incl. earlier-decorated ones) → wrong but layer-consistent Route URL.
- **Evidence to check:** `parse-impl.ts:1286-1293` (flat `Map.set` overwrites); `python.ts:1101-1106` (for-loop overwrites). Both source-order, so they agree — no graph↔contract divergence. No two-router test.
- **Recommended fix:** one-line code comment (or follow-up issue); a real fix (decorator→router-var binding) is larger and out of scope.
- **Blocks merge:** no (exotic, not a regression, layer-consistent).

**P3 — Triple-quoted string prefix diverges between layers**
- **Risk:** `prefix="""x"""` → ingestion captures empty (`/`), group captures `x` (`/x`) — a graph↔contract split.
- **Evidence to check:** `API_ROUTER_PREFIX_ARG_RE` `([^'"]*)` vs the group `unquoteLiteral` triple-quote branch.
- **Recommended fix:** none required; optional comment. No real FastAPI code uses triple-quoted URL prefixes.
- **Blocks merge:** no.

**P3 — Raw-text regex comment-blindness (ingestion)**
- **Risk:** `API_ROUTER_ASSIGN_RE` matches a commented-out `router = APIRouter(prefix="/old")` → ingestion prefixes while group tree-sitter ignores it.
- **Evidence to check:** `fastapi-router-bindings.ts:279-302` (raw-text scan) — **same behavior** as the pre-existing sibling `INCLUDE_ROUTER_*`/`FROM_IMPORT_*` regexes, not a new class.
- **Recommended fix:** none for this PR; a module-wide tree-sitter migration is a separate effort.
- **Blocks merge:** no.

**P3 — Coverage follow-ups**
- **Risk:** latent regressions slipping past the suite.
- **Evidence to check:** warm-cache SCHEMA_BUMP replay has no dedicated test (generic version-gated invalidation is covered by `parse-impl-warm-cache-parsedfile-coverage.test.ts:103`); empty `prefix=""` is only group-covered (`http-route-extractor.test.ts:7002`); no cross-layer-agreement test; multiple-routers untested.
- **Recommended fix:** add empty-prefix unit/integration, a cross-layer-agreement test, optionally a warm-cache replay test.
- **Blocks merge:** no.

### 9. PR-specific assessment
- **Correctness/consistency (credit):** both layers join via `normalizeExtractedRoutePath`, so graph URL == contract for realistic cases incl. the nested `dependencies=[Depends(get_db)], prefix=` case (verified e2e).
- **Bleed safety (credit):** ingestion gates the short-key fallback on long-key absence; group reads from the file's own tree in `scan()` → cross-file bleed structurally impossible; root↔subdir regression-tested on both layers (`fastapi-prefix-pipeline.test.ts:124-139`, `http-route-extractor.test.ts:7024-7059`).
- **Persistence/worker (credit):** SCHEMA_BUMP 8→9 (`parse-cache.ts:58`); new field is clone-safe; `??=` guards throughout.
- **Performance (credit):** same-file extraction moved into `scan()` → no double-parse.
- **Security (all suspicions refuted):** `([^'"]*)` excludes quotes; CSV-escaped on the graph path, prepared-statement `$contractId` on the bridge path; `findMatchingParen` is O(n), no ReDoS; bidi/zero-width grep clean; same safety class as include_router; no auth/secrets/workflow/Docker touched.
- **Standards/DoD:** #2311 acceptance met; no user docs needed; router-only documented; no CHANGELOG/`any`/if-in-tests; FastAPI-in-`parse-impl` is a pre-existing pattern.

### 10. Back-and-forth avoided by verifying
Confirmed the multiple-router and comment-blindness cases are layer-**consistent** (both wrong identically) → P2/P3, not `api_impact`-corrupting P1; confirmed CI green for HEAD `52817904` and Vercel red = infra; confirmed SCHEMA_BUMP=9 + the generic invalidation test cover the warm-cache no-op; ran the FastAPI suites after `npm run build` → **225/225 pass**.

### 11. Open questions
- Fix the `router`-reassignment last-write-wins now, comment it, or track as follow-up?
- Is broader router-variable-name support (`api_router`, …) desired as a follow-up, or is `router`-only the accepted scope?

### 12. Final verdict
**`production-ready with minor follow-ups`.** The PR correctly and consistently indexes same-file `APIRouter(prefix=...)` across the ingestion graph and group contract layers through one shared normalizer, is bleed-proof, invalidates warm cache, and carries unit + group + integration tests (225/225) with green cross-platform CI. The security lane refuted every injection/ReDoS/Unicode suspicion — same safety class as the existing include_router prefix. No P0/P1; the residuals are P2/P3 exotic-shape gaps and coverage follow-ups, none merge-blocking. Merge once a maintainer approves; optionally document the multiple-`router`-reassignment limitation first.

---
🤖 Automated GitNexus swarm review (6 lanes + synthesis-critic gate) generated with [Claude Code](https://claude.com/claude-code).

koriyoshi2041:
Thanks, this matches how I read the remaining edge cases too. I would not broaden this PR to arbitrary router variable names; keeping the scope to the common `router = APIRouter(...)` pattern feels right for #2311.

For multiple reassignments in one file, I would track it as a follow-up rather than churn this branch again before approval. The current behavior is layer-consistent, and fixing it properly probably needs decorator-to-binding source-order tracking rather than a small patch.

PR #2313, opened 2026-06-27
fix(group): cache read-only bridge handle to fix Windows @group reopen (#2274)

### LINKED ISSUE #2274: Group bridge.lbug in-process read-only reopen fails on Windows
## Limitation

The group bridge (`bridge.lbug`) can be reopened read-only in-process on Linux/macOS after #2269 (root cause: `closeBridgeDb` issued `CHECKPOINT` on read-only handles, leaving a WAL/shadow lock artifact that failed the next read-only open; fix = `BridgeHandle._readOnly` flag + skip CHECKPOINT for it). **Windows still fails** the in-process read-only reopen, so *repeated* `@group` trace/impact calls in a long-lived MCP server are affected on Windows.

## Root cause (so far)

The writable-close → read-open handoff plus `writeBridge`'s atomic sidecar rename does not release the OS file handle on Windows, and the open-side `LBUG_OPEN_RETRY` only retries lock-*pattern* errors, not the post-rename db-id mismatch. The core LadybugDB adapter's own reopen (`safeClose` / `lbug-close-handle-release`) passes on Windows, so this is **bridge + Windows specific**.

An attempt to add `waitForWindowsHandleRelease` + `finalizeLbugSidecarsAfterClose` to `closeBridgeDb` and flip the reopen tests to run on Windows was **reverted** — Windows CI proved it ineffective and it added latency. The reopen tests (`itLbugReopen` / `describeReopen`) stay Windows-skipped.

## Suggested direction

- Investigate the bridge close/reopen handle-release on Windows specifically; align with the core adapter's `safeClose` win32 path more carefully (win32-gated), or
- Handle the post-rename db-id mismatch in the open-side retry (distinct from the lock-pattern retry).

## Pointers

- `gitnexus/src/core/group/bridge-db.ts` — `closeBridgeDb`, `openBridgeDbReadOnly`
- `gitnexus/test/unit/group/bridge-db.test.ts` — `itLbugReopen` (Windows-skipped)

Follow-up from #2269.

prajapatisparsh:
Closes #2274.

## Problem
In a long-lived MCP `serve` process, every `@group` trace/impact call opened `bridge.lbug` read-only, queried, and closed it. On Windows the in-process reopen of the same file fails, so repeated `@group` calls broke. #2269 fixed Linux/macOS (skip CHECKPOINT on read-only handles); Windows stayed broken. A prior close-side parity attempt (waitForWindowsHandleRelease + finalizeLbugSidecarsAfterClose) was reverted as ineffective on Windows CI.

## Approach — stop reopening
Cache one read-only handle per groupDir and reuse it across calls; the open-once-per-process pattern already works on Windows. Cache invalidates on `bridge.lbug` mtime change or explicitly before same-process writes; concurrent first-open is guarded by an in-flight promise (no handle leak); handles close on process exit. The cached handle is shared across concurrent `@group` callers, so `queryBridge` serializes per handle (a per-handle FIFO lock — a LadybugDB `Connection` is not safe for concurrent queries) and a refcount lease defers the native close until in-flight readers drain (so an in-process `group_sync` can never close a handle a reader is still using). `closeBridgeDb` is a lease release for cached handles; uncached/writable handles are unchanged.

## Scope / known limitation
The in-process write→read reopen of the same bridge.lbug file is unchanged (still a LadybugDB Windows limitation), so the unit-level reopen tests remain win32-skipped. This PR removes the *production* symptom (repeated reopen in serve), not the underlying engine behavior.

## Verification (CI-gated Windows answer)
Cross-process integration coverage lives in `test/integration/group/bridge-cache-reopen.test.ts` and **runs on win32 CI** (it is *not* skipped). `bridge.lbug` is seeded in a **separate process**, so the read-only opens are clean cross-process opens rather than the in-process write→read reopen (the unfixed limitation that otherwise forces a win32 skip). Two things are proven on Windows there:

1. **Production fix:** repeated `getCachedBridgeReadOnly` returns one reused handle with no reopen — the symptom that broke repeated `@group` calls.
2. **B2 rename-clash probe:** while a cached RO handle is held, an external atomic rename over `bridge.lbug` is asserted to succeed. This answers, on Windows CI, whether the cache introduces a serve-vs-sync rename clash. LadybugDB opens RO with `FILE_SHARE_DELETE`, which should permit the rename; if the probe fails on Windows CI, then `writeBridge`'s invalidate-before-rename (and the win32 drain) are load-bearing rather than belt-and-suspenders.

Local (macOS/Linux): group unit + integration suites green; `tsc` and Prettier clean.


koriyoshi2041:
I think the Windows failure is from the test gate, not the cache implementation.

The failing assertion is the first `getCachedBridgeReadOnly(tmpDir)` in `cache reopens after bridge.lbug mtime changes`:

- CI: `tests / windows-latest (platform-sensitive)` fails at `test/unit/group/bridge-db.test.ts:508` with `expected null not to be null`.
- That test shares the same `beforeEach writeBridge(...) -> read-only open` setup described in lines 485-491 as the known Windows write->read reopen limitation.
- The neighboring cache tests use `itCacheReopen` for that exact setup, but this mtime test is still a plain `it(...)`, so Windows reaches the known unsupported setup before it can exercise mtime invalidation.

Smallest fix is probably to make this one `itCacheReopen(...)` too. If you want Windows coverage for the invalidation path, it likely needs a different setup that does not begin with same-process writable close -> read-only reopen.

gitnexus/test/unit/group/bridge-db.test.ts
```diff
@@ -436,6 +449,115 @@ describe('writeBridge + read', () => {
...
+    const first = await getCachedBridgeReadOnly(tmpDir);
+    expect(first).not.toBeNull();
+
+    const second = await getCachedBridgeReadOnly(tmpDir);
+    expect(second).not.toBeNull();
+    // Must be the SAME object — not a new open
+    expect(second).toBe(first);
+  });
+
+  it('cache reopens after bridge.lbug mtime changes', async () => {
```
magyargergo: **P0 (CI red): this must be `itCacheReopen`, not `it`.** The two sibling tests in this describe (lines 493, 534) use `itCacheReopen` (= `it.skip` on win32) because the block's `beforeEach` calls `writeBridge`, making the first `getCachedBridgeReadOnly(tmpDir)` an in-process write→read reopen — the LadybugDB Windows limitation this PR does **not** fix. On `windows-latest` it returns `null`, so `expect(first).not.toBeNull()` (line 508) throws — this is the failing `tests / windows-latest (platform-sensitive)` job and the downstream `CI Gate`. Same shape that got PR #2269 reverted (issue #2274 body). Fix: change `it(` → `itCacheReopen(`. _[reproduced — CI log]_

gitnexus/src/core/group/cross-impact.ts
```diff
@@ -369,7 +377,10 @@ export async function ensureBridgeReady(
       error: `No bridge.lbug in this group directory. Run gitnexus group sync (schema ${BRIDGE_SCHEMA_VERSION}).`,
     };
   }
-  const handle = await openBridgeDbReadOnly(groupDir);
+  // Use the cached read-only handle if available — avoids reopening the same
+  // bridge.lbug in a long-lived MCP server, which fails on Windows because
+  // the OS handle isn't fully released before the next open races in.
+  const handle = await getCachedBridgeReadOnly(groupDir);
```
magyargergo, line 380: **P1: this swaps per-call connection isolation for one shared native connection with no serialization.** Pre-PR `ensureBridgeReady` opened a fresh `lbug.Connection` per request; it now returns the cached handle, so every concurrent `@group` caller for a groupDir shares one connection. `queryBridge` (bridge-db.ts:352-370, unchanged) runs `conn.prepare`/`execute`/`query`/`getAll` with **no mutex**, and the MCP SDK dispatches tool handlers concurrently (the handler is fired as a microtask, not awaited) — so two parallel `group_impact`/`group_trace` calls on the same group drive two queries onto one connection. `conn-lock.ts:1-16` documents that hazard class ("dispatching two queries on one connection at the same time … corrupting the heap" → "double free or corruption / SIGSEGV"). **Caveat (precise):** conn-lock's observed crash was a CHECKPOINT/COPY *write* overlap (`analyze --pdg`); here the overlap is read-only queries, which is **not** confirmed safe and has no documented RO exemption. Confirm RO-concurrency safety with the LadybugDB contract, or serialize per cached handle (a `withConnLock`-style chain keyed by groupDir), or give each caller its own RO connection from a small pool. The long-lived MCP server is this cache's target deployment, so this is the exposed path. _[code-read]_

gitnexus/src/core/group/bridge-db.ts
```diff
@@ -379,6 +519,13 @@ export async function writeBridge(
   input: WriteBridgeInput,
 ): Promise<WriteBridgeReport> {
   await fsp.mkdir(groupDir, { recursive: true });
+
+  // Invalidate the RO cache before writing. On Windows the cached handle
+  // would block the atomic rename (tmp → bridge.lbug) because the OS keeps
+  // a shared-mode lock on the open file. Closing it first guarantees the
+  // rename succeeds without EBUSY.
+  await invalidateBridgeCache(groupDir);
```
magyargergo, line 756: **P1: closing the cached handle (here and on mtime-evict) has no refcount, so it can close a connection an in-flight read is still using.** `writeBridge` calls `invalidateBridgeCache(groupDir)` → `closeBridgeHandle` on the cached handle. `group_sync` is in-process (`service.ts:340` → `syncGroup` → `writeBridge`), so a concurrent `@group` impact/trace can be mid-`queryBridge` on that same handle when it is closed → "connection closed" at best, native use-after-free at worst. The mtime-change evict path (lines 135-137) has the same shape: two concurrent callers observing the same stale entry both `closeBridgeHandle` it (double-close), and an evicted handle's `finally { closeBridgeDb(h) }` (cross-impact.ts:670, cross-trace.ts:1026/1258) fails the identity guard (`cached.handle === handle`, line 398) and closes it a second time. The identity guard correctly protects the *current* cached handle — nothing protects an *evicted* one. Add a refcount/lease: mark the entry draining, defer the native close until in-flight queries finish, and close exactly once. _[code-read]_

gitnexus/src/core/group/bridge-db.ts
```diff
@@ -33,6 +35,163 @@ const bridgeLogger = createLogger('bridge-db', { debugEnvVar: 'GITNEXUS_DEBUG_BR
...
+  }
+
+  // TOCTOU guard: if another caller is already opening for this key,
+  // await their in-flight promise instead of starting a second open.
+  const inFlight = inFlightOpens.get(key);
+  if (inFlight) return inFlight;
+
+  const openPromise: Promise<BridgeHandle | null> = (async () => {
+    try {
+      const handle = await openBridgeDbReadOnly(groupDir);
```
magyargergo, line 297: **P2: the cache fixes read→read reuse, not the write→read open — so the first post-sync `@group` read on Windows still returns null.** A cache miss falls straight to `openBridgeDbReadOnly` here. Since `group_sync` is in-process and invalidates the cache, the realistic workflow "agent runs `group_sync`, then immediately `group_impact` on the same group" is a cache miss → first RO open right after the writable close = the unfixed Windows write→read reopen (your own NOTE at lines 403-413 describes the post-rename sidecar db-id mismatch that the lock-pattern retry does not match → immediate null). This is exactly what the failing CI test encodes. Scope the "closes the Windows gap" claim to read→read, or fix the write→read reopen itself (e.g. hand the writer's open handle to the cache so the first post-sync read reuses it). _[reproduced — CI log]_

gitnexus/src/core/group/bridge-db.ts
```diff
@@ -231,47 +390,28 @@ function unwrapQueryResult(queryResult: lbug.QueryResult | lbug.QueryResult[]):
...
-  try {
-    await (handle._db as lbug.Database).close();
-  } catch {
-    /* ignore */
+  // If this handle is the currently cached RO handle, the cache owns its
+  // lifetime. Only close on explicit invalidation or process shutdown.
+  const key = bridgeCacheKey(handle.groupDir);
+  const cached = cachedBridgeHandles.get(key);
+  if (cached && cached.handle === handle) {
+    return; // cache owns this handle
```
magyargergo: **P2: `closeBridgeDb` now silently no-ops for cached handles, but its exported signature is unchanged — callers can't tell.** All three production callers (cross-impact.ts:670, cross-trace.ts:1026/1258) call it in a `finally` as if closing what they opened; they're now no-ops (correct today, since `writeBridge` calls `invalidateBridgeCache` before renaming). But a future caller using the open-in-try / close-in-finally idiom and then renaming/deleting bridge.lbug would hit EBUSY on Windows because the handle is still open. Document it (JSDoc: "no-op for cache-owned handles; call `invalidateBridgeCache` before renaming/deleting") or rename to `releaseBridgeHandle`. Also: no test exercises this no-op branch — if it were inverted to `closeBridgeHandle`, every caller's `finally` would defeat the cache and no test would catch it. _[code-read]_

gitnexus/test/unit/group/bridge-db.test.ts
```diff
@@ -436,6 +449,115 @@ describe('writeBridge + read', () => {
...
+    const { getCachedBridgeReadOnly, invalidateBridgeCache } =
+      await import('../../../src/core/group/bridge-db.js');
+    const first = await getCachedBridgeReadOnly(tmpDir);
+    expect(first).not.toBeNull();
+
+    // Rewrite the bridge to bump mtime
+    await invalidateBridgeCache(tmpDir);
```
magyargergo: **P2: this test proves manual invalidation, not `writeBridge`'s self-invalidation.** The explicit `await invalidateBridgeCache(tmpDir)` before `writeBridge` clears the cache, so `writeBridge`'s own internal `invalidateBridgeCache` (impl line 527) is a no-op and the stat-based mtime branch (impl line 127) is never exercised. The load-bearing Windows invariant — `writeBridge` must close a *live* cached handle before its atomic rename or the rename fails with EBUSY — is therefore untested. Drop the manual call and let `writeBridge` be the sole invalidator, and add an external-writer case that bumps mtime without calling `invalidateBridgeCache`. _[code-read]_

gitnexus/test/unit/group/bridge-db.test.ts
```diff
@@ -625,3 +759,81 @@ describe('findContractNode', () => {
...
+    await fsp.writeFile(tmpPath, await fsp.readFile(dbPath));
+
+    let renameError: string | null = null;
+    try {
+      await fsp.rename(tmpPath, dbPath);
+    } catch (err: unknown) {
+      renameError = err instanceof Error ? err.message : String(err);
+    }
+
+    if (renameError) {
```
magyargergo: **Nit (repo "no if-branching in vitests"): drop the `if (renameError)` block.** The `expect(renameError).toBeNull()` at line 828 is unconditional, so there's no vacuous pass here, but the repo convention forbids branching in test bodies. Fold the diagnostic into the assertion message: `expect(renameError, '[B2] rename failed while cached RO handle held').toBeNull();`. Separately, this B2 probe is `it.skip` on win32, so it can't actually probe the Windows share-mode behavior its comment targets. _[code-read]_

magyargergo (commented):
**Verdict: not production-ready — request changes; do not merge until the P0 (CI-red) and both P1s (shared-connection concurrency, use-after-close) are resolved.** This is a collaborative comment review, not a formal block. The cache is a legitimate, well-motivated fix with a real performance win, but it ships CI red and introduces a shared native `lbug.Connection` into the long-lived MCP server that is its own target deployment — without serialization or lifetime ownership. (Branch hygiene: **clean feature/fix PR** — 1 non-merge commit, 3 scoped files. Merge state: **checks failing** — `mergeable: MERGEABLE`, `BLOCKED` only by the red checks.)

## Engines

- **Engine A — Claude:** PR swarm (risk-architect, test-ci-verifier) + 8 ce personas (correctness, adversarial, performance, reliability, maintainability, testing, project-standards, api-contract).
- **Engine B — Codex / GPT-5.5:** swarm + ce + adversarial — all three legs live and substantive.

**Cross-engine corroboration (honest):** Engine A and Engine B share the same review scaffolding, so same-engine agreement is "consistent," not independent. The high-confidence signal here is where a **Codex lane and a Claude lane independently land on the same finding**. That happened on four findings: (1) the line-504 gating bug, (2) the shared-connection concurrency hazard — both engines cite `conn-lock.ts` by line, (3) the in-process sync→read Windows efficacy gap, and (4) the `closeBridgeDb` no-op contract change. Those four are the spine of this review. The use-after-close P1 is mostly Engine A (correctness, adversarial, reliability, swarm-risk) with Codex adversarial corroborating the `groupSync`-invalidates-mid-query trigger.

## Headline

1. **CI is RED and it's a one-line test-gating bug.** `bridge-db.test.ts:504` is a plain `it(` where its two siblings (493, 534) use `itCacheReopen` (= `it.skip` on win32). The describe's `beforeEach` does an in-process `writeBridge` then the body opens RO — the exact write→read reopen that still fails on Windows — so `expect(first).not.toBeNull()` (508) throws on `windows-latest`. This is the second time un-skipping a Windows bridge-reopen test broke CI (PR #2269 was reverted for the same shape; issue #2274 body confirms). Fix: `itCacheReopen(` at 504.
2. **The cache trades the Windows reopen problem for a native-concurrency problem in the exact deployment it targets.** Before this PR every `@group` call opened its own connection (isolated). The cache hands one shared `lbug.Connection` per groupDir to every concurrent caller, and `queryBridge` has no mutex. The MCP SDK dispatches tool handlers concurrently, so two parallel `@group` impact/trace calls — or a `group_sync` racing an in-flight read — can drive two queries onto one connection at once. `conn-lock.ts:1-16` documents that exact hazard class ("dispatching two queries on one connection at the same time … corrupting the heap" → "double free or corruption / SIGSEGV"); the crash it cites was a CHECKPOINT/COPY **write** overlap, and concurrent read-only queries on one connection have **not** been confirmed safe (no RO exemption is documented). No serialization, no refcount, no lease.

## Findings

| Sev | Finding | File:line | Engines | Trigger | Tag |
|-----|---------|-----------|---------|---------|-----|
| **P0** | Line-504 mis-gated `it` runs the Windows write→read reopen → CI red | `test/unit/group/bridge-db.test.ts:504` (assert 508) | **Both** (A: swarm test-ci, correctness, testing, maintainability, api-contract, project-standards; B: swarm lane-1, ce #3) | Plain `it` runs on win32; `beforeEach` writeBridge → `getCachedBridgeReadOnly` returns null | [reproduced] |
| **P1** | Shared `lbug.Connection`, no query serialization → concurrent `@group` calls can corrupt native heap (conn-lock.ts hazard class; RO-concurrency unconfirmed) | adopt: `cross-impact.ts:383` / `bridge-db.ts:130`; real loc `queryBridge` 352–370 (unchanged), `conn-lock.ts:1-16` | **Both** (A: swarm risk-A, correctness F1, adversarial F1, reliability F1, performance F2, maintainability, api-contract AC-3, project-standards RR1; B: swarm risk, ce #1, adversarial high-1) | Two concurrent `group_impact`/`group_trace` on one group → both run `conn.execute` on shared conn | [code-read] |
| **P1** | Use-after-close / double-close: in-process `group_sync`→`invalidateBridgeCache` (or concurrent mtime-evict) closes a handle an in-flight read holds; no refcount | `bridge-db.ts:527` (invalidate in writeBridge); evict+close 135-137; identity guard 392-401 | A: swarm risk-B, correctness F2/F3, adversarial F2, reliability F2/F3; **B: adversarial high-1** (groupSync-invalidates-mid-drain) | `group_sync` writes bridge while a concurrent `@group` query drains on the same handle | [code-read] |
| **P2** | Windows efficacy gap: first in-process `group_sync`→`@group` read still returns null on Windows (CI-proven); fix only covers read→read | `bridge-db.ts:147` (cache-miss → openBridgeDbReadOnly); NOTE 403-413; `service.ts:340` | **Both** (A: swarm risk-F, correctness F4, adversarial F5; B: swarm risk lane-3, ce #2, adversarial high-2) | Agent runs `group_sync` then `group_impact` in one MCP process on Windows | [reproduced] |
| **P2** | `closeBridgeDb` silently no-ops for cached handles — undocumented exported-contract change + no test covers the branch | `bridge-db.ts:398` (392-401) | **Both** (A: maintainability, api-contract AC-2, testing F5, swarm risk-E; B: swarm risk P2) | Future caller does open→use→`closeBridgeDb`→rename, hits EBUSY on Windows | [code-read] |
| **P2** | mtime-change test proves manual invalidation, not `writeBridge` self-invalidation (redundant `invalidateBridgeCache` before `writeBridge`) | `test/unit/group/bridge-db.test.ts:511` | **Both** (A: swarm test-ci, testing F3, maintainability; B: swarm test-ci lane-4) | Manual clear at 511 masks the stat-based mtime branch (bridge-db.ts:127) | [code-read] |
| **P2** | `if (renameError)` branch in vitest B2 probe — repo "no if-branching in vitests" rule; assertion is unconditional so vacuous-pass risk is low | `test/unit/group/bridge-db.test.ts:823` | A only (swarm test-ci, testing F4); **project-standards lane disputes** the vacuous framing | Style/DoD nit; `expect(renameError).toBeNull()` at 828 still guards | [code-read] |
| **P2** | Concurrency untested: no test fires concurrent `queryBridge` on one cached handle (ties to P1) | `test/unit/group/bridge-db.test.ts:534` (open-dedupe only) | **Both** (A: testing, correctness, adversarial, api-contract TG-1, reliability; B: adversarial rec) | Existing concurrent test asserts handle identity, then queries serially | [code-read] |
| **P2** | Windows coverage gap: after the 504 fix, all 3 cache tests skip on win32 → cache efficacy unproven on the target platform | `test/unit/group/bridge-db.test.ts:491,493,504,534` | A (swarm test-ci, testing F2); B notes B2/PR-body mismatch | A child-process or fixture-seeded read→read test would run on Windows | [code-read] |
| P3 | `writeBridge` invalidate→rename is not atomic; a concurrent reader re-locks bridge.lbug in the window → Windows EBUSY can recur + stale handle cached | `bridge-db.ts:527`→760 | **B adversarial F3** (Codex-only — worth noting) | Concurrent `@group` read between invalidate and retryRename | [code-read] |
| P3 | `beforeExit` cleanup never fires on `process.exit()`/SIGTERM/SIGINT (every MCP shutdown path); stated WAL benefit is moot for RO handles | `bridge-db.ts:191` | A (adversarial F6, reliability F4); B notes cleanup-path weakness | Server killed by signal → handles only OS-reclaimed | [code-read] |
| P3 | mtime strict-equality false-negative serves a stale handle on coarse FS (FAT/SMB/NFS) or rename-preserved mtime | `bridge-db.ts:129` | A (correctness F5, adversarial R1) | External writer rewrites bridge.lbug within one mtime tick | [code-read] |
| P3 | post-open ENOENT stores `mtime=0` → cache bypassed every call until file reappears | `bridge-db.ts:150-155` | A (swarm risk-C) | bridge.lbug absent at open time | [code-read] |
| P3 | ~16% of the diff (~80/515 lines) is incidental Prettier churn within printWidth — inflates a ~160-line change to 447, harms reviewability | bridge-db.ts / cross-impact.ts / test (multiple) | A (maintainability) | n/a | [code-read] |
| P3 | describe comment claims cache tests verify mtime invalidation "on macOS/Linux/Windows" — false given win32 skips | `test/unit/group/bridge-db.test.ts:~460` | A (swarm test-ci) | n/a | [code-read] |

## Inline comments (postable — diff-present RIGHT-side lines)

**1. P0 — `test/unit/group/bridge-db.test.ts:504` [reproduced]**
> **P0 (CI red): this test must be `itCacheReopen`, not `it`.** The two sibling tests in this describe block (lines 493, 534) use `itCacheReopen` (= `it.skip` on win32) precisely because the block's `beforeEach` calls `writeBridge`, making the first `getCachedBridgeReadOnly(tmpDir)` an in-process write→read reopen — the LadybugDB Windows limitation this PR does **not** fix. On `windows-latest` this returns `null` and `expect(first).not.toBeNull()` (line 508) throws; this is the failing `tests / windows-latest (platform-sensitive)` job and the downstream `CI Gate` failure. This is the same shape that got PR #2269 reverted. Fix: change `it(` to `itCacheReopen(` at line 504.

**2. P1 — `gitnexus/src/core/group/cross-impact.ts:383` [code-read]**
> **P1: this swaps per-call connection isolation for a shared native connection with no serialization.** Before this PR `ensureBridgeReady` called `openBridgeDbReadOnly` (a fresh `lbug.Connection` per request); it now returns the cached handle, so every concurrent `@group` caller for a groupDir shares one `lbug.Connection`. `queryBridge` (bridge-db.ts:352-370, unchanged) issues `conn.prepare`/`conn.execute`/`conn.query`/`getAll` with **no mutex**. The MCP SDK dispatches tool handlers concurrently (`protocol.js` fires the handler as a microtask without awaiting), so two parallel `group_impact`/`group_trace` calls on the same group interleave two queries on one connection. That is the hazard class `conn-lock.ts:1-16` documents — its rule is general ("dispatching two queries on one connection at the same time … corrupting the heap" → "double free or corruption / SIGSEGV"). Caveat (be precise): the crash it actually observed was a CHECKPOINT/COPY **write** overlap (`analyze --pdg`); here the overlap is read-only queries, which has **not** been confirmed safe and has no documented RO exemption. So: confirm with the LadybugDB contract that concurrent RO `prepare`/`execute` on one connection is safe, or serialize per cached handle (a `withConnLock`-style promise chain keyed by groupDir), or give each concurrent caller its own RO connection from a small pool. The long-lived MCP server is this cache's target deployment, so this is the exposed path, not a corner case.

**3. P1 — `gitnexus/src/core/group/bridge-db.ts:527` [code-read]**
> **P1: closing the cached handle here (and on mtime-evict) has no refcount, so it can close a connection an in-flight read is still using.** `writeBridge` calls `invalidateBridgeCache(groupDir)` → `closeBridgeHandle` on the cached handle. `group_sync` is in-process (`service.ts:340` → `syncGroup` → `writeBridge`), so a concurrent `@group` impact/trace can be mid-`queryBridge` on that same handle when it's closed → "connection closed" at best, native use-after-free at worst. The same applies on the mtime-change evict path (135-137): two concurrent callers observing the same stale entry both call `closeBridgeHandle` on it (double-close), and an evicted handle's `finally { closeBridgeDb(h) }` (cross-impact.ts:670, cross-trace.ts:1026/1258) fails the identity guard (`cached.handle === handle`, line 398) and closes it a second time. The identity guard correctly protects the *current* cached handle, but nothing protects an *evicted* one. Add a refcount/lease: mark the entry draining, defer the native close until in-flight queries finish, and ensure a handle is closed exactly once.

**4. P2 — `gitnexus/src/core/group/bridge-db.ts:147` [reproduced]**
> **P2: the cache fixes read→read reuse, not the write→read open — so the first post-sync `@group` read on Windows still returns null.** A cache miss falls straight to `openBridgeDbReadOnly` here. Since `group_sync` is in-process and invalidates the cache, the realistic workflow "agent runs `group_sync`, then immediately `group_impact` on the same group" is a cache miss → first RO open right after the writable close = the unfixed Windows write→read reopen (your own NOTE at 403-413 describes the sidecar db-id mismatch that `isTransientLockError` doesn't match → no retry → immediate null). This is what the failing CI test actually encodes. The PR/NOTE claim that the cache "closes the Windows gap" should be scoped to read→read, or the write→read reopen itself needs fixing (e.g. hand the writer's open handle to the cache so the first post-sync read reuses it).

**5. P2 — `gitnexus/src/core/group/bridge-db.ts:398` [code-read]**
> **P2: `closeBridgeDb` now silently no-ops for cached handles, but its exported signature is unchanged — callers can't tell.** All three production callers (cross-impact.ts:670, cross-trace.ts:1026/1258) call it in a `finally` as if closing what they opened; they're now no-ops (correct today, since `writeBridge` calls `invalidateBridgeCache` before renaming). But a future caller following the open-in-try/close-in-finally idiom and then renaming/deleting bridge.lbug would hit EBUSY on Windows because the handle is still open. Document it (JSDoc: "no-op for cache-owned handles; call `invalidateBridgeCache` before renaming/deleting") or rename to `releaseBridgeHandle`. Also: no test exercises this no-op branch — if it were inverted to `closeBridgeHandle`, every caller's `finally` would defeat the cache and no test would catch it.

**6. P2 — `gitnexus/src/core/group/bridge-db.ts:511`... (test) — `test/unit/group/bridge-db.test.ts:511` [code-read]**
> **P2: this test proves manual invalidation, not `writeBridge`'s self-invalidation.** The explicit `await invalidateBridgeCache(tmpDir)` before `writeBridge` clears the cache, so `writeBridge`'s own internal `invalidateBridgeCache` (impl line 527) is a no-op and the stat-based mtime branch (impl line 127) is never exercised. The load-bearing Windows invariant — `writeBridge` must close a *live* cached handle before its atomic rename or the rename fails with EBUSY — is therefore untested. Drop the manual call and let `writeBridge` be the sole invalidator (and add an external-writer case that bumps mtime without calling `invalidateBridgeCache`).

**7. P2/P3 — `test/unit/group/bridge-db.test.ts:823` [code-read]**
> **Nit (repo "no if-branching in vitests"): drop the `if (renameError)` block.** The `expect(renameError).toBeNull()` at line 828 is unconditional, so there's no vacuous-pass here (the project-standards lane agreed this isn't a hard violation), but the repo convention forbids branching in test bodies. Fold the diagnostic into the assertion message: `expect(renameError, '[B2] rename failed while cached RO handle held').toBeNull();`. Separately, this B2 probe is `it.skip` on win32, so it can't actually probe the Windows share-mode behavior its comment targets.

## Credit — solid work and validated-correct (refuted suspicions)

What's genuinely good:
- **Performance win is real.** On a cache hit the heavy native open + `db.init()`/`conn.init()` + up to ~3 s of Windows open-retry backoff collapse to a single `fsp.stat`. The hot `@group` path gets materially cheaper (performance lane).
- **The TOCTOU open guard (`inFlightOpens`) is correct.** Concurrent cache-miss opens dedupe to one open; the second caller awaits the first's promise. No double-open, no handle leak on the open side. (Verified by adversarial, reliability, api-contract, correctness.)
- **The `closeBridgeDb` identity guard correctly protects the *live* cached handle.** The per-request `finally { closeBridgeDb(handle) }` is a safe no-op while the handle is current — only the *evicted* handle is at risk (finding P1 #3).
- **Clean hygiene / mergeable.** One non-merge commit, 3 files, scoped to the group bridge layer; no CHANGELOG edit; conventional `fix(group):` title; no new `any`/`as any`; `git diff --check` clean; no bidi/hidden-Unicode; `mergeable: MERGEABLE` (`BLOCKED` only by the failing checks). `writeBridge` invalidates before opening the staging temp DB (no two-writable-handle window).

Refuted / validated-correct (good that the review didn't chase these):
- **Open-side TOCTOU double-open — refuted.** `inFlightOpens` handles it (the close side is where the risk lives).
- **"cross-trace bypasses the cache" — refuted** (prior-context #44383 is false). `cross-trace.ts` gets its handle via `ensureBridgeReady` → `getCachedBridgeReadOnly`; both impact and trace share the cache — which is *why* the P1s apply to both.
- **Repeated `beforeExit` → double-close — refuted.** `closeAllCachedBridges` clears the map synchronously before awaiting closes; a second fire finds an empty map.
- **`closeBridgeDb` closing the live cached handle mid-use — refuted for steady state.** Identity guard makes it a no-op; only the evicted handle is exposed.

## CI status

**RED.** `tests / windows-latest (platform-sensitive)` fails at `bridge-db.test.ts:508`; `CI Gate` fails downstream (`TESTS=failure`). `macOS`, `ubuntu`, `packaged-install`, and `tree-sitter-ABI` all pass. CI also logged "Unhandled Error: Worker exited unexpectedly" in the bridge-db worker — likely a side effect of the failed RO open / afterEach EBUSY on Windows; worth confirming it doesn't mask another failure. The diff is small (3 files, no CI/workflow changes), so review coverage is full.

---
*Automated multi-tool digest (Claude swarm + ce, Codex/GPT-5.5 swarm + ce + adversarial). Verify findings against the diff before acting; inline anchors point at the cache-adoption/getCachedBridgeReadOnly lines, with the real unchanged `queryBridge` location referenced in the comment text.*

magyargergo (changes requested):
Please review the comments above.

gitnexus/test/unit/group/bridge-db.test.ts
magyargergo: **Fixed (P0 / CI-red).** The mtime test is now gated with `itCacheReopen` (= `it.skip` on win32), so the `beforeEach` `writeBridge` → RO-open — the in-process write→read reopen — no longer runs on `windows-latest`. The failing `expect(first).not.toBeNull()` is skipped there and the `CI Gate` goes green; macOS/ubuntu still run it. Same conclusion @koriyoshi2041 and the Copilot note reached.

After the fix every test in the `bridge handle cache` describe is `itCacheReopen` (the block always opens RO right after a writable `writeBridge`), so none of them run the unsupported reopen on Windows. The test itself was also restructured for the self-invalidation finding — see my reply on the `invalidateBridgeCache(tmpDir)` thread (`test:511`).

Verified locally: `tsc --noEmit` clean, `vitest run test/unit/group/bridge-db.test.ts` → 42/42.

gitnexus/src/core/group/cross-impact.ts
magyargergo: **Fixed (P1 — query serialization).** Added a **per-handle** FIFO lock (the `conn-lock.ts:30-55` chain mechanic, keyed per cache entry) — deliberately *not* the global `withConnLock`, which would over-serialize every group against the adapter singleton on the hot `@group` path. `queryBridge` runs every conn op on that chain, so two concurrent `@group` callers sharing one cached connection can never dispatch two queries at once. Uncached/writable handles (the `writeBridge` temp DB) are single-threaded and skip the lock at zero cost.

As you flagged, I did **not** assert RO-concurrency is safe (conn-lock documents no RO exemption) — I serialize.

```ts
export async function withHandleLock<T>(lock: { lockTail: Promise<void> }, fn: () => Promise<T>): Promise<T> {
  const prior = lock.lockTail;
  let release!: () => void;
  lock.lockTail = new Promise<void>((resolve) => { release = resolve; });
  await prior;
  try { return await fn(); } finally { release(); }
}

// queryBridge:
const entry = bridgeEntryByHandle.get(handle);
return entry ? withHandleLock(entry, run) : run(); // uncached/writable → no lock
```

New deterministic test (mirrors `conn-lock.test.ts`) fires 8 concurrent ops through one lock and asserts `Math.max(active) === 1` (plus releases-on-throw). Verified green.

gitnexus/src/core/group/bridge-db.ts
magyargergo: **Fixed (P1 — use-after-close / double-close).** Added a refcount lease (mirrors `pool-adapter.ts` `checkedOut` + idempotent release). `getCachedBridgeReadOnly` takes a lease (`refs++`), `closeBridgeDb` releases it (`refs--`) — your existing consumer `finally { closeBridgeDb(handle) }` is now the release point, **no API/caller change**. Both close sites — `invalidateBridgeCache` **and** the mtime-evict branch (the *other* live close-under-lease path) — route through one `evictBridgeEntry` → `finalizeBridgeClose`, which:

- **defers** the native close until `refs === 0` (no in-flight reader) → kills the use-after-close;
- closes **exactly once** via a `closeStarted` guard → kills the double-close (incl. the evicted-handle `finally{closeBridgeDb}` you described).

One refinement from a deepening pass on top of your suggestion: the drain must **not** block `writeBridge` on active readers (the lease spans the whole `@group` fan-out, e.g. `cross-impact.ts:598-668`, not just the bridge queries — blocking there would starve `group_sync` unboundedly). So `evictBridgeEntry` awaits the close **only when `refs === 0`** (the common single-threaded `group_sync` — preserving the Windows close-before-rename guarantee); when a reader holds a lease it defers and returns, and `retryRename` covers the rare concurrent-Windows window.

```ts
async function evictBridgeEntry(key, entry) {
  if (!entry.evicted) { entry.evicted = true; if (cachedBridgeHandles.get(key) === entry) cachedBridgeHandles.delete(key); }
  if (entry.refs <= 0) await finalizeBridgeClose(entry);   // refs>0 → deferred to last closeBridgeDb release
}
async function finalizeBridgeClose(entry) {
  if (entry.closeStarted) return;          // close-once guard
  entry.closeStarted = true;
  bridgeEntryByHandle.delete(entry.handle);
  await closeBridgeHandle(entry.handle);
}
```

New tests prove: invalidate defers close while leases are held and closes once on the last release; `refs` counts every `inFlightOpens` awaiter (N-1 releases don't close, the Nth does); and the mtime-evict path defers identically. Verified green.

gitnexus/src/core/group/bridge-db.ts
magyargergo: **Fixed (P2 — scoped the claim).** Reworded the NOTE so it no longer claims the cache "closes the Windows gap." It now states the cache removes the **production** symptom — a long-lived serve process reopening `bridge.lbug` on every call — for **read→read** reuse, and explicitly does **not** fix write→read:

> Scope of the RO bridge-handle cache: it removes the PRODUCTION symptom ... by keeping one RO handle alive for read→READ reuse. It does NOT fix the write→READ reopen: the first `@group` read right after an in-process `group_sync` is a cache miss → `openBridgeDbReadOnly`, i.e. the same unfixed reopen, so on Windows that first post-sync read still returns null.

This matches the realistic `group_sync` → `group_impact` workflow you described and lines up with the win32 test skips.

gitnexus/src/core/group/bridge-db.ts
magyargergo: **Fixed (P2 — documented the contract + tested the branch).** Kept the name `closeBridgeDb` (it's genuinely dual-use — a lease *release* for cache-owned handles, a real *close* for the writable `writeBridge` temp handle; renaming to `releaseBridgeHandle` churns 3 call sites and is wrong for the writable case) and spelled the contract out in JSDoc:

```ts
/**
 * - Cache-owned handle (from getCachedBridgeReadOnly): the matching release —
 *   decrements the lease refcount, does NOT close. The cache owns the lifetime.
 *   If already evicted and this is the last lease, the deferred native close fires here.
 * - Uncached/writable handle (e.g. the writeBridge temp DB): closes for real.
 * Contract: before renaming/deleting bridge.lbug, call invalidateBridgeCache (not this).
 */
```

And the no-op/release branch is now covered: the new refcount tests assert that releasing the *live* cached handle decrements without closing, while the evicted handle closes on the last release (so inverting it to an unconditional close would now fail a test).

gitnexus/test/unit/group/bridge-db.test.ts
magyargergo: **Fixed (P2 — test now proves self-invalidation).** Dropped the manual `await invalidateBridgeCache(tmpDir)`. The test (`writeBridge self-invalidates the cache (no manual invalidate)`) now relies solely on `writeBridge`'s own internal invalidate, so impl line 527 **and** the stat-based mtime branch are actually under test — a fresh `getCachedBridgeReadOnly` after `writeBridge` returns a different handle with the updated data.

Added a separate external-writer test (`cache reopens when an external writer bumps mtime`) that exercises the stat-mtime branch directly **without** going through `writeBridge`. It bumps mtime with `fsp.utimes` — not `fsp.writeFile`, since `bridge.lbug` is a native binary and rewriting its bytes corrupts it (the reopen would then return `null` and the test would throw):

```ts
const future = new Date(Date.now() + 5000);
await fsp.utimes(dbPath, future, future);
const second = await getCachedBridgeReadOnly(tmpDir);
expect(second).not.toBe(first);          // mtime moved → fast path missed → fresh handle
```

Both are `itCacheReopen` (their first open is the write→read reopen, win32-skipped). Verified green.

gitnexus/test/unit/group/bridge-db.test.ts
magyargergo: **Fixed (P2 — no if-branching in vitests).** Dropped the `if (renameError) { console.warn(...) }` block and folded the diagnostic into the assertion message, so the test body has no branch:

```ts
// No `if`-branch in the test body (repo convention). The diagnostic — on POSIX
// this should never fail; on Windows a failure means the share mode is
// insufficient (FILE_SHARE_DELETE not set) — rides on the assertion message.
expect(renameError, `[B2] rename failed while cached RO handle held: ${renameError}`).toBeNull();
```

The B2 probe stays `it.skip` on win32 (its `beforeEach` puts it in the write→read reopen class) — the in-code comment already states that correctly. The PR-body claim that it's "intentionally NOT win32-skipped" is the part that's stale; see my separate comment on Windows coverage (F9) for the body correction.

gitnexus/src/core/group/bridge-db.ts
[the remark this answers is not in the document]
magyargergo: **Re: the writer-vs-new-reader race during the write window** (also the tri-review P3 "invalidate→rename not atomic"). With the refcount fix, `invalidateBridgeCache` no longer blocks the rename when readers are active and the entry is dropped from the cache **synchronously**, so I deliberately did **not** add a hard write-gate. The residual window — a *new* reader cache-missing between invalidate and the atomic rename — is covered by:

1. `retryRename` (EBUSY ×3, exp backoff) for the rare Windows case where a freshly-opened RO handle briefly holds the file, and
2. mtime/size self-healing — a handle opened pre-rename is evicted+reopened on the next call once the rename bumps mtime (so it can't serve stale data indefinitely).

I scoped this as the lazy-cut (P3, single-engine + your note): adding a `writingGroups` gate would close the window entirely but adds per-group serialization for a window already retry-covered. Happy to add the gate if you'd rather close it hard — say the word.

magyargergo: **Windows hardening (follow-up to the refcount fix above).** The non-blocking invalidate I described left one rough edge on Windows: a `group_sync` racing an in-flight read would defer the close and hit `retryRename` EBUSY (recoverable, but not clean). Added a **platform-aware bounded drain** in `evictBridgeEntry` — this is exactly ADV-1's recommended option (b):

```ts
if (entry.refs <= 0) { await finalizeBridgeClose(entry); return; }
// refs > 0: close deferred to the last closeBridgeDb release.
if (process.platform === 'win32') {
  // Windows can't rename over an open handle — wait (bounded) for readers to
  // release, then the rename is clean; on timeout, retryRename copes.
  let timer; const timeout = new Promise<void>((r) => { timer = setTimeout(r, WINDOWS_DRAIN_TIMEOUT_MS); });
  await Promise.race([entry.drained, timeout]).finally(() => clearTimeout(timer));
}
```

- **POSIX:** unchanged — fully non-blocking (rename succeeds over the open RO handle; old inode survives the in-flight reader). No starvation.
- **Windows:** `group_sync` waits up to `WINDOWS_DRAIN_TIMEOUT_MS` (5 s) for the reader to release so the atomic rename is clean; past that it falls through to `retryRename`, so a pathologically long reader can never wedge `group_sync` (addresses ADV-1's unbounded-hang concern).

Single-threaded `group_sync` (`refs===0`, the common case) still closes-before-rename synchronously on all platforms — the original Windows guarantee. The win32 branch has a unit test (mocks `process.platform` on the Linux runner: with a lease held, `invalidateBridgeCache` does NOT resolve until the reader releases). Verified: tsc + bridge-db unit + group integration green.

magyargergo:
@prajapatisparsh can you please test this on your side? 🙏

magyargergo (approved):

prajapatisparsh:
Okay I'll test em

PR #2323, opened 2026-06-30
fix(indexing): keep full text file content searchable

### LINKED ISSUE #2317: GitNexus only indexes the first ~10 KB of text files
### Area

### Summary

GitNexus only indexes the first ~10 KB of .txt files, making content beyond that limit unavailable for queries.

### Context

I was testing GitNexus's ability to query information from large text files. The repository contains a .txt file larger than 10 KB. Queries against content within the first ~10 KB succeed, while queries for content located later in the file consistently fail.

### Expected behavior

GitNexus should index and query the entire contents of supported text files, allowing information anywhere in the file to be retrieved.

### Actual behavior

Only the first ~10 KB of the .txt file appears to be indexed. Any content beyond that point is ignored during querying, resulting in incomplete or missing responses.

### Steps to reproduce

1. Create or add a .txt file larger than 10 KB to a repository.
2. Add unique text near the beginning of the file and additional unique text after the first ~10 KB.
3. Index the repository with GitNexus.
4. Query for text located near the beginning of the file (works as expected).
5. Query for text located beyond the first ~10 KB (no results are returned).

### Environment

OS: macOS 10.15.7
Node: (fill in your version, e.g., node -v)
Browser (if applicable): Chrome 149
Commit / Version: (fill in GitNexus version or commit SHA)

### Logs / screenshots

No errors or warnings are displayed during indexing or querying. The issue is reproducible consistently. Screenshots or sample files can be provided if needed.

koriyoshi2041:
## What changed

- Stop truncating File node content at 10,000 characters during LadybugDB CSV emit.
- Keep existing binary-file suppression and symbol-level snippet limits unchanged.
- Add a regression test with a .txt file whose unique query term appears after the old 10KB cutoff.

## Why

Fixes #2317. The indexed File.content column was receiving only the first 10,000 characters, so FTS/query could not find text later in otherwise-supported text files.

## Verification

- npm test -- test/integration/csv-pipeline.test.ts
- npx tsc --noEmit
- npx prettier --check gitnexus/src/core/lbug/csv-generator.ts gitnexus/test/integration/csv-pipeline.test.ts
- git diff --check
- npm run lint -- gitnexus/src/core/lbug/csv-generator.ts gitnexus/test/integration/csv-pipeline.test.ts (exited 0; repo-wide warnings only)


koriyoshi2041:
Checked the red Windows platform-sensitive job. The log shows vitest hit the 15 minute timeout after the aggregate report had 13,542 passed / 0 failed / 56 skipped, so I don't see a failing assertion tied to this change.

This PR's diff is still limited to:
- `gitnexus/src/core/lbug/csv-generator.ts`
- `gitnexus/test/integration/csv-pipeline.test.ts`

Local recheck on head `f3b08798`:
- `npx tsc --noEmit`: passed
- `npx prettier --check src/core/lbug/csv-generator.ts test/integration/csv-pipeline.test.ts`: passed
- `npm test -- test/integration/csv-pipeline.test.ts`: passed, 17 tests

I tried to rerun the failed job, but GitHub requires repo admin rights for that run rerun. Holding for a maintainer rerun or a concrete Windows failure log.

gitnexus/src/core/lbug/csv-generator.ts
```diff
@@ -156,10 +156,7 @@ const extractContent = async (node: GraphNode, contentCache: FileContentCache):
   if (isBinaryContent(content)) return '[Binary file - content not stored]';
 
   if (node.label === 'File') {
-    const MAX_FILE_CONTENT = 10000;
-    return content.length > MAX_FILE_CONTENT
-      ? content.slice(0, MAX_FILE_CONTENT) + '\n... [truncated]'
-      : content;
+    return content;
```
magyargergo: **[P2 default / P1 once `GITNEXUS_MAX_FILE_SIZE` is raised] Uncapped File content makes the count-based flush buffer byte-unaware.** Blocks merge: maybe.

Returning the full File body removes the only tight per-row size bound — content is now bounded only by the walker cap (512KB default / 32MB override; `filesystem-walker.ts:66`, `max-file-size.ts:5`, `constants.ts:13`). `BufferedCSVWriter.flush` does `this.buffer.join('\n')` over `FLUSH_EVERY = 500` rows (`csv-generator.ts:210`, `:48`): ~256MB transient spike per flush at the 512KB default (vs ~5MB pre-PR), and — with `GITNEXUS_MAX_FILE_SIZE` raised, the big-file case this PR exists to enable — a 500-row batch can exceed V8's 2^29 (536,870,912)-char string limit, throwing `RangeError: Invalid string length` synchronously inside `flush()`, aborting the whole analyze and leaking write-stream fds.

_Fix:_ add a decoupled FTS/CSV-write content cap (~64–256KB) here in the File branch — separate from the walker's parse/OOM knob — or make `BufferedCSVWriter` flush on accumulated bytes rather than row count.

_Cross-engine (Codex B1+B2 + Claude performance/reliability/adversarial). [code-read]_

gitnexus/test/integration/csv-pipeline.test.ts
```diff
@@ -166,6 +166,30 @@ describe('streamAllCSVsToDisk', () => {
...
+        filePath: 'src/large.txt',
+      },
+    ]);
+
+    const result = await streamAllCSVsToDisk(graph, repoDir, csvDir);
+    const fileCsv = result.nodeFiles.get('File');
+    expect(fileCsv).toBeDefined();
+
+    const content = await fs.readFile(fileCsv!.csvPath, 'utf-8');
+    expect(content).toContain(lateNeedle);
```
magyargergo, line 193: **[P2] The test proves the CSV cell, not end-to-end FTS searchability.** Blocks merge: no.

This asserts the late needle is present in the generated `file.csv` (`:188-190`), but never loads LadybugDB, never builds `file_fts`, and never calls `searchFTSFromLbug` — so #2317's actual symptom (FTS *search* can't find content past 10KB) is unproven end-to-end. The CSV→COPY→FTS→query round-trip over a >10KB multiline quoted cell is exactly what the PR claims to fix and the one thing left unexercised.

_Fix:_ add a follow-up FTS integration test (this suite deliberately avoids LadybugDB) that COPYs a >10KB File and asserts the late needle is returned by an FTS query. Non-blocking — a guard-coverage gap, not a defect.

_Cross-engine (Codex B1 + Claude risk-architect/test-ci/correctness). [code-read]_

magyargergo (commented):
## Tri-review digest — PR #2323 (remove 10KB `MAX_FILE_CONTENT` cap so File content past 10KB is FTS-searchable, fixes #2317)

**Two-engine review.** Engine A = Claude (swarm: risk-architect / test-ci / security-boundary; ce: correctness / performance / reliability / adversarial / maintainability / project-standards). Engine B = Codex / GPT-5.5 (swarm B1 live, ce B2 live, adversarial B3 partial — it surfaced the 512KB/32MB bound, then the stream disconnected). The two strong findings below are **cross-engine** (a Claude lane and a Codex lane reached each independently). The headline framing that removing the cap makes File content "unbounded" is **refuted**: the filesystem walker (`filesystem-walker.ts:66`) skips files larger than `maxFileSizeBytes` *before* they become File nodes, so worst-case File content is bounded at **512KB by default** (`max-file-size.ts:5`) and **32MB with the `GITNEXUS_MAX_FILE_SIZE` override** (clamped to `TREE_SITTER_MAX_BUFFER`, `constants.ts:13`) — a ~50× rise from the old 10KB, inside an existing configurable envelope, not a new unbounded DoS.

### What's solid (credit)
- **Fixes a real bug (#2317).** File rows are FTS-indexed on `['name','content']` (`fts-schema.ts:24`); the old 10KB slice truncated content *before* it reached the index, so text past 10KB was genuinely unsearchable. The one-line `return content;` is the correct root-cause fix and is minimal/targeted (2 files, +25/−4).
- **Correctness validated** (both engines): CSV escaping is untouched and still applied to the full body at `csv-generator.ts:461` (`escapeCSVField`, RFC-4180 quote-doubling); the COPY reader is `PARALLEL=false` and explicitly supports multiline quoted cells of any size (`lbug-adapter.ts:1234-1244`); no downstream consumer hardcodes a ≤10KB assumption (grep clean); the `content === ''` / Folder / binary short-circuits and the exactly-10000-char boundary are byte-identical to before; **FileContentCache RAM is unchanged** (the cache always stored the full body — the 10KB cap lived on the write path only).
- **New test is a genuine regression guard** and CI-wired: it writes a needle past the old cutoff and asserts both `toContain(lateNeedle)` and `not.toContain('[truncated]')` — both flip red pre-PR. Runs in the vitest `default` project on every PR (`ci-tests.yml`).
- **Conventions clean:** no `any`/`as any`, no if-branching in the new vitest, correct integration-test placement, CHANGELOG correctly untouched (release-owned), no hidden unicode/bidi in the diff (`git diff --check` clean). CI is green except Vercel (auth/infra, not code).

### Finding 1 (headline) — count-based flush buffer is byte-unaware: transient ~256MB spike at default, deterministic crash at raised cap. P2 default / P1 once `GITNEXUS_MAX_FILE_SIZE` is raised. **Cross-engine** (Codex B1#2 + Codex B2 + Claude performance/reliability/adversarial). `[code-read]`
- `BufferedCSVWriter.flush` does `this.buffer.join('\n')` over up to `FLUSH_EVERY = 500` File rows (`csv-generator.ts:210`, `:48`). Pre-PR a File row was ≤10KB, so a 500-row flush was ≤~5MB. With the cap gone, the per-row ceiling is the walker cap.
- **Default 512KB:** 500 × 512KB ≈ a **256MB transient join string per flush** (plus the source array and the `ws.write` copy) — a ~50× memory regression, survivable on a roomy host, a P3 OOM risk on a constrained CI heap.
- **`GITNEXUS_MAX_FILE_SIZE` raised (the PR's *own* intended big-file config):** a 500-row batch can sum past V8's max string length (2^29 = 536,870,912 chars). `Array.prototype.join` then throws **`RangeError: Invalid string length` synchronously inside `flush()`**, which aborts the entire `streamAllCSVsToDisk` run, writes no `rel_*.csv`, runs no COPY, and leaks the open write-stream fds — a hard crash that recurs until the operator lowers the cap. This lives exactly in the use case the PR exists to enable, which is why it escalates to P1 there. Both Codex legs independently rated it **P2**; Claude perf/reliability said P1 — synthesis lands on P2-default / P1-at-override.
- **Secondary (single-engine, Claude):** the same uncapped content also inflates the `file_fts` index, which is dropped+rebuilt every analyze (`fts-indexes.ts:96-97`) and probed sequentially across ~20 indexes per query (`bm25-index.ts:86-103`) → larger index, slower BM25 (compounding the already-"unmeasured" 20-index latency from #2300). Same root cause, same fix.
- **Convergent fix** (independently proposed by performance, reliability, security, risk-architect): a separate, documented FTS/CSV-write content cap (~64–256KB) in the `extractContent` File branch, **decoupled** from the walker's `GITNEXUS_MAX_FILE_SIZE` (which is a parse/OOM knob). Keeps most files fully searchable while restoring a tight write-path bound. A byte-aware flush would also bound the spike.
- **Blocks merge: maybe** — at default config the growth is bounded and survivable (not blocking); the hard `RangeError` crash only manifests once `GITNEXUS_MAX_FILE_SIZE` is raised (a non-default action, but the one this PR exists to enable), so the byte-aware flush / decoupled write-cap is a strongly-recommended follow-up rather than a hard gate.

### Finding 2 — the new test proves the CSV cell, not end-to-end FTS searchability. P2. **Cross-engine** (Codex B1#1 + Claude risk-architect/test-ci/correctness). `[code-read]`
- The test writes ~14.4KB (`'…\n'.repeat(400)` + needle), calls `streamAllCSVsToDisk`, reads `file.csv`, and asserts the needle is in the CSV text (`csv-pipeline.test.ts:184-190`). It never loads LadybugDB, never builds `file_fts`, never calls `searchFTSFromLbug` — so #2317's actual symptom (FTS *search* can't find content past 10KB) is unproven end-to-end. The CSV→COPY→FTS→query round-trip over a >10KB multiline quoted cell is the thing the PR claims to fix, and it's the one thing the test doesn't exercise.
- Honest mitigation: `csv-pipeline.test.ts` deliberately avoids LadybugDB (the `lbug-db` project is excluded from the default vitest run), so this is partly a suite-level limitation rather than negligence — recommend a follow-up FTS integration test that COPYs a >10KB File and asserts the late needle is returned by an FTS query.
- **Blocks merge: no** — a regression-guard completeness gap, not a defect in the change; fine to merge with the FTS end-to-end test as a follow-up.

### Secondary / lower-priority (body only — single-engine or pre-existing)
- **BM25 query-time cost (Codex-unique, B2, P2 conf 75):** the BM25 path returns `RETURN node, score` (`bm25-index.ts:38`), so full `File.content` *may* be materialized per result row at query time, not just at index time. *Unverified* — depends on whether LadybugDB materializes the full node struct or only consumed fields on `RETURN node`; check a query plan before acting. If confirmed, consider projecting only `id`/`filePath`/`score` for nodeId-only paths.
- **DB/disk inflation (Claude perf F3, P2):** the File table `content` column and the `.lbug` file grow proportionally; bigger COPY + WAL on re-analyze.
- **Binary-sampler head-only bypass (Claude security F3, pre-existing, amplified):** `isBinaryContent` samples only the first 1000 chars (`csv-generator.ts:93`), so an all-ASCII head + high-byte tail passes; the old 10KB cap was the implicit backstop. Garbage-into-FTS, not an exec/render exploit.
- **Bidi pass-through (Claude security F5, P3, pre-existing):** U+202A–202E etc. survive `sanitizeUTF8` into `file_fts`; display-layer only.
- **Maintainability (Claude, P2):** `return content;` sits 11 lines above the still-capped snippet path (`MAX_SNIPPET = 5000`, `:170`). Add a one-line comment naming the deliberate "File is intentionally unbounded for FTS; snippets stay capped" asymmetry so the next reader doesn't read it as a forgotten guard.
- **Project-standards (Claude, P3 conf 50):** CONTRIBUTING.md line 69 asks for a docs update on a behavior change; a one-line ARCHITECTURE/RUNBOOK note (full File content now stored; re-index repopulates the column) would satisfy it, or mark the DoD item N/A.

### Refuted / validated (these are features of the review, not gaps)
- **Unbounded / 50MB-file DoS — REFUTED** (both engines): walker cap hard-bounds File content at 512KB/32MB (`filesystem-walker.ts:66`); import-target `generateId('File',…)` calls create edge endpoints only, not content nodes, so no cap bypass.
- **Cache RAM growth — REFUTED**: cache always held the full body; the cap was post-cache.
- **CSV-injection / COPY field-size corruption — REFUTED**: escaping is RFC-4180-correct; `PARALLEL=false` handles embedded newlines/quotes at any size; `content STRING` is unbounded VARCHAR with no field limit found in repo.
- **Lockfile / minified FTS pollution — LARGELY MITIGATED**: the default ignore list already excludes lockfiles, `.min.js`/`.bundle.js`/`.d.ts`/`.map`, and `node_modules`/`dist`/`build`/`vendor` (`ignore-service.ts`). Residual = large hand-written source/markdown/data files just under 512KB.
- **IGNORE_ERRORS silent row-drop** (`lbug-adapter.ts:923-925`): real pre-existing code, but no LadybugDB cell/field limit was demonstrated to trip at ≤512KB → trigger unproven, not postable; only a theoretical concern at a 32MB override.

### Verdict & state
- **Final verdict: production-ready with minor follow-ups.** The change is the correct root-cause fix for a real bug (#2317), is minimal and targeted, and correctness/escaping/conventions are validated with green CI. At the default 512KB cap the resource growth is bounded and survivable — nothing here corrupts data or fails at default config. The two follow-ups are real but non-blocking: (1) a byte-aware flush / decoupled FTS-write content cap before anyone runs a large-file repo with `GITNEXUS_MAX_FILE_SIZE` raised (where Finding 1 becomes a hard `RangeError` crash), and (2) an end-to-end FTS test proving #2317's search symptom is actually fixed. Neither needs to gate this merge, but the write-path cap should land soon.
- **Branch hygiene:** merge-from-main commit present but harmless and merge-safe — the diff against merge-base is exactly the two intended files (+25/−4), no unrelated churn.
- **Merge state:** `mergeable` — no conflicts, CI green except infra Vercel; `mergeStateStatus = BEHIND` main, so update/rebase onto `main` before merging, but nothing blocks it.
- CI green except **Vercel** (auth/infra, not a code gate).

---
*This is an automated multi-tool (two-engine: Claude + Codex/GPT-5.5) digest. Verify the findings against the code before acting.*

PR #2327, opened 2026-06-30
fix(group): replace LadybugDB-incompatible multi-label Cypher (#2325)

### LINKED ISSUE #2325: [Bug] manifest-extractor resolveSymbol fails with multi-label Cypher syntax unsupported by LadybugDB
## Description

  When using `gitnexus group sync` with manifest links (custom contract type),
  the `manifest-extractor`'s `resolveSymbol` function generates a Cypher query with multi-label
  disjunction syntax (`n:Function|Method|Class|Interface|...`), which LadybugDB does not support.

  The cross-links are still established via synthetic UIDs (name-only matching), but without
  resolved file paths — `symbolRef.filePath` is empty for all manifest links.

  ## Environment

  - GitNexus: v1.6.7 (via npx)
  - Node.js: v25.8.2
  - OS: macOS

  ## Reproduction Steps

  1. Create a group with multiple Java repos
  2. Add manifest links in `group.yaml` for custom contracts:

  ```yaml
  links:
    - contract: "custom::MyServiceFacade"
      type: custom
      from: repo-a
      to: repo-b
      role: consumer

  3. Run gitnexus group sync <group-name> --verbose

  Actual Behavior

  Every manifest link produces a warning:

  [manifest-extractor] resolveSymbol failed for custom:custom::MyServiceFacade in repo-b:
  Prepare failed: Parser exception: Invalid input
  <MATCH (n:Function|Method|Class|Interface|Struct|Enum|Trait|Constructor|TypeAlias|Impl|Macro>:
  expected rule oC_SingleQuery (line: 1, offset: 86)

  Cross-links are created with confidence: 1 and matchType: "manifest", but symbolRef.filePath
  is empty. group impact partially fails due to missing symbol resolution.

  Expected Behavior

  resolveSymbol should use LadybugDB-compatible Cypher syntax. Replace:

  MATCH (n:Function|Method|Class|Interface|Struct|Enum|Trait|Constructor|TypeAlias|Impl|Macro)
  WHERE n.name = 'MyServiceFacade'
  RETURN n

  with:

  MATCH (n)
  WHERE labels(n) IN ['Function','Method','Class','Interface','Struct','Enum','Trait',
        'Constructor','TypeAlias','Impl','Macro']
  AND n.name = 'MyServiceFacade'
  RETURN n

  The labels() function works in LadybugDB (confirmed in issue #1216).

  Impact

  - Not blocking: manifest links still create valid cross-links via synthetic UIDs
  - Degraded: group impact cannot trace from contract to actual code symbol/file
  - Affects: any type: custom manifest link (and potentially grpc/thrift if they use the same resolveSymbol path)

  ---

magyargergo:
## Summary

Closes #2325. Cross-repo group links that pointed at a manifest `custom` contract, or at a library (`lib`) contract, were silently losing their target symbol: the resolved `filePath` came back empty and the link fell back to a synthetic UID. The cause was a Cypher-dialect issue — both group extractors built queries with the openCypher label disjunction `MATCH (n:A|B|C)`, and **LadybugDB's parser rejects such a disjunction when a label in it is a reserved keyword (`Macro` and `Union` both are) or names a node table that doesn't exist**. The error was swallowed by a `try/catch`, so nothing surfaced; the contract just resolved to nothing.

All seven affected queries now use `MATCH (n) WHERE labels(n) IN [...]`. LadybugDB returns `labels(n)` as a single string, so this is an exact label allowlist — the same set the disjunction intended. It is a 1:1 behavior-preserving translation (the label sets are unchanged), validated against LadybugDB 0.17.1.

> **Diagnosis refined during review.** An earlier framing claimed LadybugDB rejects `MATCH (n:A|B)` universally. Verified against the real parser, it does **not** — `Function|Method|CodeElement`, `Function|Const`, `Function|Static`, and a 7-label no-reserved list all parse fine. The genuine trigger is a reserved-keyword label (`Macro`/`Union`) or a missing node table.

## What was actually broken

Only the queries whose label list hit one of those two conditions:

- **manifest-extractor `custom` branch** — its 21-label allowlist contains `Macro` and `Union` (reserved keywords) → parser rejection → every `custom` contract resolved to a synthetic `manifest::<repo>::<contract>` UID with an empty `filePath`.
- **manifest-extractor `lib` branch** — allowlisted a non-existent `Package` node table → binder error → `lib` contracts silently unresolved.

The `topic` / `grpc` / `thrift` branches and both **http-route-extractor** queries (`Function|Method|CodeElement`) contain no reserved keyword and **parsed fine — they were never broken**. They were converted to the `labels(n) IN` form for consistency and to stay immune if a reserved-keyword label is added later, not because they failed.

It shipped undetected because every extractor unit test mocks the Cypher executor — nothing ran these query strings through the real parser. This PR adds that missing layer.

## Key decision

`labels(n) IN [...]` over a per-label `UNION ALL`: the `custom` query allowlists 21 labels, which `UNION ALL` would turn into 21 stacked `RETURN` blocks (and complicate the single `ORDER BY ... LIMIT`). The existing `UNION ALL` form in `CONTAINING_QUERY` is left as-is. The fix changes only Cypher syntax, never which labels each query allows.

## Tests

- Two new real-DB integration suites run the **exact production queries** against a seeded LadybugDB — the coverage layer whose absence let the bug ship. `RESOLVE_BY_NAME_QUERY` / `RESOLVE_IN_MODULE_QUERY` and the `custom` query are exported for this, and both suites are registered in the sequential `lbug-db` Vitest project per `TESTING.md`.
- Every per-branch case seeds a same-named **wrong-label decoy whose `filePath` sorts first**, so a widened or dropped allowlist would surface the decoy and flip the assertion (exclusion is non-vacuous). Includes a `thrift` package-strip case.
- Rejection canaries pin the **real** trigger: the manifest `custom` disjunction (its list names `Macro`/`Union`) throws, and so does a `MATCH (n:Function|Macro|Union)` form — so a reintroduced reserved-keyword disjunction fails loudly. A `labels()`-is-a-string canary guards against a future LadybugDB semantics change.
- `tsc` clean, eslint clean, full group suite + new cases green.

## Out of scope (pre-existing, not regressions)

Real coverage gaps the audit surfaced, but widening them would make this stop being a behavior-preserving syntax fix — each deserves its own change + tests:

- `custom` allowlist omits emitted `Variable` / `Namespace` labels.
- gRPC/thrift **service** allowlist omits `Trait` (Rust tonic) and `Struct` (Go gRPC service impls).
- Separately, the codebase's `labels(n)[0] AS type` projections (augmentation / wiki) return an empty string on LadybugDB 0.17.x — a pre-existing concern that deserves its own issue.

> The dead `'Package'` label in the `lib` allowlist — previously listed here as deferred — is **now removed** in this PR.

## Post-deploy validation

After a `group sync`: the `[manifest-extractor] resolveSymbol failed ... Parser exception` warnings should disappear, and previously-unresolved `custom`/`lib` manifest cross-links should resolve to real symbol UIDs. A rise in realized cross-repo hit counts is **expected** here (manifest links that were silently dead now produce hits, and newly-matching UIDs let manifest-preference dedup fire) — not a regression.

---

Claude Code


gitnexus/test/integration/group/manifest-resolve-symbol-2325.test.ts
```diff
@@ -0,0 +1,151 @@
...
+ * These cases run the REAL production query (via `extractFromManifest`) against a
+ * real LadybugDB — the only layer that can catch a parser rejection, since the
+ * unit tests mock the executor. Each per-branch case seeds a wrong-label decoy
+ * whose `filePath` sorts BEFORE the target so a widened/broken allowlist would
+ * surface the decoy under `ORDER BY n.filePath ASC LIMIT 1` and flip the
+ * assertion — making the exclusion check non-vacuous.
+ */
+import { it, expect, afterEach } from 'vitest';
+import { ManifestExtractor } from '../../../src/core/group/extractors/manifest-extractor.js';
+import type { GroupManifestLink, CypherExecutor } from '../../../src/core/group/types.js';
```
magyargergo: **[P2] `CypherExecutor` is imported from a module that doesn't export it (latent `TS2305`).** `[reproduced]`

`types.js` exports `GroupManifestLink` (✅) but **not** `CypherExecutor` — that type is defined only at `gitnexus/src/core/group/contract-extractor.ts:13`, and all 13 production extractors import it from `'../contract-extractor.js'`.

This reproduces as a real type error:
```
$ npx tsc --noEmit -p tsconfig.test.json
test/integration/group/manifest-resolve-symbol-2325.test.ts(18,34): error TS2305: Module '"../../../src/core/group/types.js"' has no exported member 'CypherExecutor'.
```
It's masked today because CI's `quality / typecheck` runs `tsc --noEmit` on the default `tsconfig.json` (`include: ["src/**/*"]`, so `test/` is out of scope), and `import type {…}` is erased at runtime so vitest never resolves it. It still reddens IDE / local `tsconfig.test.json` runs and would break the moment test typechecking is wired into CI.

**Fix:** split the import — keep `GroupManifestLink` from `types.js`, import `CypherExecutor` from `'../../../src/core/group/contract-extractor.js'`.

gitnexus/test/integration/group/manifest-resolve-symbol-2325.test.ts
```diff
@@ -0,0 +1,151 @@
...
+import { initLbug, executeParameterized, closeLbug } from '../../../src/core/lbug/pool-adapter.js';
```
magyargergo, line 28: **[P2] These new native-LadybugDB tests aren't registered in the sequential `lbug-db` Vitest project.** `[code-read]`

Both new integration files open a real LadybugDB (`pool-adapter` / `withTestLbugDB`, `@ladybugdb/core`) but appear in **neither** the `lbug-db` `include` allowlist nor the `default` `exclude` list in `gitnexus/vitest.config.ts` — so they fall through to the `default` project, which runs files in **parallel forks**.

`TESTING.md:68` states the policy verbatim: *"When adding a new test that uses native LadybugDB (`@ladybugdb/core`), add it to the `lbug-db` project's explicit include list and the `default` project's exclude list."* The `lbug-db` project is sequential specifically to *"prevent file-lock conflicts from [the] native mmap addon"* — every one of the 23 sibling lbug integration tests is listed there. Running these two in parallel risks exactly that mmap file-lock flake on `tests / windows-latest`.

**Fix:** add **both** new files (`manifest-resolve-symbol-2325.test.ts` and `http-route-resolve-symbol.test.ts`) to `lbug-db.include` and `default.exclude`.

_Cross-engine: independently flagged by Codex (swarm) and the Claude test-CI lane._

magyargergo (commented):
# Tri-review digest — PR #2327 · fix(group): replace LadybugDB-incompatible multi-label Cypher (#2325)

**Verdict: APPROVE with 2 non-blocking P2 hygiene follow-ups.** The core fix is correct, empirically reproduced, and the root cause is fully eradicated in `src/`. Two P2s are test-side only (one latent type error, one test-isolation policy violation); neither affects production behavior. A handful of P3 doc/test-quality polish items.

## Engine / liveness breakdown
Three engines, six lanes, all live and returning real findings:
- **Engine A (Claude):** GitNexus swarm (risk-architect, test-ci-verifier — both partial, coordinator-backfilled) + ce-code-review personas (correctness, testing, maintainability, project-standards).
- **Engine B (Codex GPT-5.5/xhigh), all 3 legs LIVE — verified each artifact is a real report, not a "skill not found"/empty marker:**
  - **Codex swarm (Solo)** — full PR review, P3 lbug-db wiring finding.
  - **Codex ce-code-review (report-only)** — JSON report, P2 `CypherExecutor` import finding.
  - **Codex native adversarial-review** — `verdict: approve`, no material blocker; explicitly refuted the failure probes (labels()/guards/exported-strings).
- **Coordinator verification** — independently REPRODUCED the two P2s (tsc + config read). Highest-confidence evidence.

## Validated correctness — the fix is sound (credit, not findings)
- **`labels(n)`-is-a-scalar-string: TRIPLE-confirmed** (ce-correctness empirical probe against `@ladybugdb/core@0.17.1` + Codex B3 adversarial + Codex B1 swarm MCP Cypher probe) and **REPRODUCED** — new integration tests 10/10 pass, unit 25/25. The comment at `http-route-extractor.ts:90` is empirically correct, not overstated. `labels(n)[0]` would have returned `""` and matched nothing, so the chosen form is the *only* correct one for this engine.
- **Multi-label nodes impossible:** Codex B3 confirmed LadybugDB rejects multi-label CREATE; `schema.ts` = one node table per label (Kùzu-style). `IN [...]` as an allowlist is sound.
- **Guards preserved (diff-verified, all lanes):** `LIMIT 2` ambiguity cap, `ORDER BY n.filePath ASC LIMIT 1`, `n.filePath <> ''` synthetic-node exclusion, `toResolvedSymbol` `uid`-only-when-`rows.length===1`.
- **No remaining `MATCH (n:A|B)` disjunction in `src/`** (grep + B1 + B3 + coordinator) — root cause fully eradicated, not just the ticketed path.
- **No injection:** label allowlists are hardcoded string literals; `$name`/`$contract`/`$symbolName`/`$methodName` etc. remain prepared-statement-bound.
- **Branch hygiene clean** (1 commit, 5 files, no merge commits, no CHANGELOG/lockfile churn); **security clean** (bidi/`git diff --check` clean — B1); **standards clean** (project-standards + Codex correctness/project-standards/performance all clean); behavioral delta is strictly net-positive (old form threw → swallowed → null/synthetic; new form resolves the intended node).

## Headline findings — 2× P2 (both test-side, non-blocking)

### [P2] `CypherExecutor` imported from `types.js` → TS2305 latent type error  `[reproduced]`
- **Where:** `gitnexus/test/integration/group/manifest-resolve-symbol-2325.test.ts:18` (new file, in-diff).
- **Trigger:** `npx tsc --noEmit -p tsconfig.test.json` → `manifest-resolve-symbol-2325.test.ts(18,34): error TS2305: Module '"../../../src/core/group/types.js"' has no exported member 'CypherExecutor'.` — re-run and confirmed by synthesis.
- **Evidence:** `import type { GroupManifestLink, CypherExecutor } from '.../group/types.js'`. `GroupManifestLink` IS exported (`types.ts:16`) but `CypherExecutor` is NOT — it is defined only at `contract-extractor.ts:13` (`export type CypherExecutor = (...)`). All 13 production extractors import it from `'../contract-extractor.js'`. Only the second import member is wrong.
- **Why CI is green (masking):** CI `quality / typecheck` runs `tsc --noEmit` on the default `tsconfig.json` (`include: ["src/**/*"]`), which excludes `test/`; no CI step runs `tsc -p tsconfig.test.json`. The runtime test passes because `import type {…}` is fully erased by esbuild/vitest. Latent: reddens IDE/local `tsconfig.test.json` runs and breaks the moment test typechecking is wired into CI.
- **Note:** `tsconfig.test.json` also surfaces pre-existing TS2305s elsewhere (`NodeLabel`, `ScopeResolutionIndexes`) — those are NOT this PR's; this one IS introduced by the diff. Sibling `http-route-resolve-symbol.test.ts` does not have this problem.
- **Cross-engine:** Codex B2 (ce, confidence 100) found it; Claude ce-testing missed it → exactly the cross-engine blind-spot the setup targets. Coordinator + synthesis REPRODUCED → **Strong**. Severity-override + reproduced trigger → **INLINE on line 18**.
- **Fix:** split the import — keep `GroupManifestLink` from `types.js`, import `CypherExecutor` from `'../../../src/core/group/contract-extractor.js'`.

### [P2] New native-LadybugDB tests not registered in the sequential `lbug-db` vitest project  `[code-read]`
- **Where:** policy at `TESTING.md:68`; wiring at `gitnexus/vitest.config.ts` (lbug-db include = 23-file allowlist, lines 51–74, `fileParallelism:false`; default exclude = mirror, lines 86–115). Both new test files in-diff; `vitest.config.ts` and `TESTING.md` are NOT in diff.
- **Trigger:** `tests / windows-latest` runs the `default` project with `fileParallelism:true` (`pool:'forks'`). The two new files open a real LadybugDB via `withTestLbugDB`/`pool-adapter` (confirmed by grep) but appear in NEITHER list → they fall through to `default` and can be scheduled in concurrent forks → native mmap file-lock conflict on Windows — the exact failure the `lbug-db` project exists to prevent.
- **Evidence:** `TESTING.md:68` verbatim — *"When adding a new test that uses native LadybugDB (`@ladybugdb/core`), add it to the `lbug-db` project's explicit include list and the `default` project's exclude list."* Every one of the 23 sibling lbug integration tests is listed; these two are not. Confirmed by synthesis: `grep integration/group vitest.config.ts` → no match.
- **Cross-engine:** Codex B1 swarm (P3, cited TESTING.md:68) + Claude test-ci-verifier (P2 for Windows flake). Reconciled to **P2, cross-engine corroborated** → **Strong**. Real flake risk; 2-line config fix.
- **Fix:** add both files to the `lbug-db` `include` allowlist AND the `default` `exclude` list (mirror every sibling).
- **Posting:** body lead; optional one inline anchored on a new test file's import line (config file is not in-diff, can't inline there).

## P3 polish (body only)
- **`LABELS_CUSTOM_QUERY` reconstructs the production 21-label `custom` query locally** (`manifest-resolve-symbol-2325.test.ts:49–55`) instead of importing it — silent drift risk. http-route avoided this by exporting `RESOLVE_BY_NAME_QUERY` / `RESOLVE_IN_MODULE_QUERY` (confirmed exported at `http-route-extractor.ts:101,113`); the manifest custom-branch query is an un-exported inline literal. Mitigated: the e2e per-branch cases still run the real production query via `extractFromManifest`. (Claude maintainability, P2/conf-80; reconciled to P3 — test-quality only.) Fix: export the query and import it, or add an inline-snapshot guard.
- **`PIPELINE.md:115–119` still documents the banned `:A|B` form** (`(n:Function|Method|Class|Interface)`, `(n:Package|Module)` etc.). Confirmed present; file NOT in diff → body only, can't inline. A contributor copying from it would re-introduce #2325. (ce-correctness F2.)
- **http-route test lacks a `:A|B`-rejection canary** — symmetry gap vs the manifest test's `rejects.toThrow` at lines 91–98. (ce-testing F1.)
- **Unit-test 7-label substring couples to label order/spacing** (`manifest-extractor.test.ts:845`). The real guard is the unconditional `.not.toContain('Function|Method')` at :848. (ce-testing F2 + maintainability residual.)
- **`thrift` branch not exercised through a real executor** — identical query strings to `grpc` (covered), but the package-strip path (`rawServiceName.split('.').pop()`) has only mocked-executor unit coverage. (ce-testing F3.) `http`/`include` branches were never `:A|B`-bugged → out of scope (ce-testing F4, suppress).
- **`'Package'` is a vestigial allowlist entry** in `manifest-extractor.ts:280` (`labels(n) IN ['Package','Module']`) — no `Package` node table exists. Harmless no-op now (was a hard "Table Package does not exist" error under the old form); pre-existing, optional cleanup. (ce-correctness F1.)

## Out-of-scope note (NOT inline, pre-existing)
Two lanes (ce-correctness empirical probe + Codex B1 GitNexus Cypher probe) independently observed `labels(n)[0]` returns `""` (empty) on `@ladybugdb/core@0.17.1`, while `labels(n)` returns the scalar label. This **validates** this PR's `labels(n)` choice. It ALSO implies the codebase's prevailing `labels(n)[0] AS type` idiom (`augmentation/engine.ts:129/155`, `wiki/graph-queries.ts:79/84/251/299`, `mcp/local/local-backend.ts`) may yield empty `type` strings — a **separate, pre-existing** concern (`local-backend.ts` already carries fallbacks for exactly this), out of scope for #2327. Flag for a follow-up look; do not assert a confirmed bug without checking those call sites' downstream handling.

## CI status (at review time)
- `quality` (lint / typecheck / format): **green** — note the typecheck masks the P2 above (test/ out of scope).
- `tests/*` (ubuntu coverage, Windows/macOS platform-sensitive): **pending** at review time — the lbug-db wiring P2 means a green Windows run would be schedule-luck, not proof.
- Vercel: **fail — auth/deploy authorization only, not code.**

---
*Automated multi-tool digest (GitNexus swarm + ce-code-review + Codex GPT-5.5, merged by a synthesis agent). Cross-engine-weighted and worktree-verified, but verify each finding before acting.*

PR #2314, opened 2026-06-28
fix(test): stabilize local Windows gate baselines

azizur100389:
﻿Summary
- Stabilize Windows local-gate baselines for git root, rename, literal collector, and HTTP source-scan tests.
- Normalize HTTP source-scan repo-relative paths before symbol containment lookup.

Validation
- npx vitest run test/unit/git-utils.test.ts test/unit/calltool-dispatch.test.ts test/integration/literal-collectors.test.ts test/integration/http-inline-handler-symbol-roundtrip.test.ts
- npx vitest run test/integration/typescript-async-generator-functions.test.ts test/unit/git-utils.test.ts test/unit/calltool-dispatch.test.ts test/integration/spring-interface-inheritance-pipeline.test.ts test/integration/spring-route-pipeline.test.ts test/integration/doc-comment-description-e2e.test.ts test/integration/multi-verb-route-identity.test.ts test/integration/http-inline-handler-symbol-roundtrip.test.ts test/integration/literal-collectors.test.ts test/integration/cfg/pipeline-pdg.test.ts
- npx tsc --noEmit
- npx prettier --check touched files
- gitnexus detect-changes: low risk, 0 affected processes

PR #1325, opened 2026-05-04
fix(server): harden git-clone — close 6 path-injection / CLI-injection / ReDoS alerts (U3)

magyargergo:
## Summary

**U3** of [the security remediation plan](docs/plans/2026-05-04-001-fix-medium-to-critical-security-findings-plan.md) — closes the six high-severity CodeQL alerts in `gitnexus/src/server/git-clone.ts`. This is the highest-real-attacker-risk cluster remaining: `git-clone.ts` is the subprocess pipeline that takes user-supplied URLs and shells out to \`git clone\`.

| Alert | Rule | Line |
|---|---|---|
| #185 | `js/polynomial-redos` | 16 |
| #176 | `js/path-injection` | 209 |
| #177 | `js/path-injection` | 219 |
| #178 | `js/path-injection` | 230 |
| #166 | `js/second-order-command-line-injection` | 221 |
| #167 | `js/second-order-command-line-injection` | 221 |

Tracking: #1318.

## Threat model addressed

1. **Polynomial ReDoS via trailing slashes** — `extractRepoName(url)` used `url.replace(/\/+$/, '')`. CodeQL flagged it as polynomial; fixed via O(n) charCode loop.
2. **Path-traversal via crafted URL → bad repoName → bad targetDir** — A URL like `https://github.com/owner/repo:..` would yield `repoName === '..'` from `extractRepoName`, then `getCloneDir('..')` would resolve to `~/.gitnexus/` (parent escape).
3. **Git option-injection via URL** — `runGit(['clone', '--depth', '1', url, targetDir])` lets a URL beginning with `--` (e.g. `--upload-pack=evil ...`) be parsed as a git option, executing an attacker-chosen subprocess.

### `extractRepoName(url)`
- Replaced regex-based trailing-slash trim with an O(n) charCode loop (closes #185).
- Now throws when the last segment isn't filesystem-safe (`^[a-zA-Z0-9._-]+$`, with `.` and `..` explicitly rejected). This closes the upstream of the path-injection chain.

### `getCloneDir(repoName)`
- Defense-in-depth: re-validates repoName against the same safe pattern at the boundary, so callers that don't go through `extractRepoName` (test helpers, future scripts) still can't construct an escape.

### `cloneOrPull(url, targetDir, ...)`
- **Containment barrier at function entry** (closes #176/#177/#178):
  ```ts
  const safeTarget = path.resolve(targetDir);
  const rel = path.relative(CLONE_ROOT, safeTarget);
[...]

PR #2329, opened 2026-06-30
feat: adds an opt-in auto sync and analysis loop for GitNexus

im-fan:
## Feature Overview

This PR adds an opt-in auto sync and analysis loop for GitNexus. When `AUTO_UPDATE_AND_ANALYZE_FLAG=1` is set, GitNexus reads `.gitnexus/sync_config.yml`, periodically pulls
configured repositories, analyzes only changed commits on allowed branches, registers updated repo metadata, and syncs configured GitNexus groups.

## Use Cases

- Keep frequently changing repositories indexed without manually running `gitnexus analyze`
- Maintain a GitNexus group view for multiple backend/service repositories
- Avoid redundant analysis when the current branch commit has not changed
- Retry analysis automatically after a previous failed analysis
- Safely isolate auto-cloned repositories under operator-controlled clone roots
- Quarantine partial clone outputs after clone failures for later inspection

## Summary

- Add opt-in auto sync/analyze runner gated by `AUTO_UPDATE_AND_ANALYZE_FLAG=1`
- Load `.gitnexus/sync_config.yml` to clone/pull configured remotes, filter branches, run analysis only for changed commits, register repos, and sync GitNexus groups
- Harden clone target handling for auto-sync paths: strict repo-name validation, controlled clone roots, symlink/realpath containment checks, unsafe permission rejection, and
partial-clone quarantine
- Add unit coverage for config parsing, state persistence, runner behavior, branch/commit skip logic, group sync isolation, clone-root safety, and clone failure quarantine
- Ignore generated `.tmp-test` and `/log/` artifacts

## Validation

- Added/updated unit tests:
  - `gitnexus/test/unit/auto-sync.test.ts`
  - `gitnexus/test/unit/auto-sync-runner.test.ts`
  - `gitnexus/test/unit/git-clone.test.ts`


gitnexus/src/core/auto-sync/config.ts
```diff
@@ -0,0 +1,163 @@
...
+        errors.push(`projects[${index}].local_path is required`);
+      } else {
+        try {
+          normalizeConfiguredCloneRoot(localPath);
+        } catch (err: unknown) {
+          errors.push(`projects[${index}].local_path ${(err as Error).message}`);
+        }
+      }
+
+      const remoteUrls = Array.isArray(project.remote_urls)
```
magyargergo, line 155: **P1 — SSH remotes fail on every sync cycle. [reproduced]**

`remote_urls` accepts any non-empty string here, but `cloneOrPull` → `validateGitUrl` (`git-clone.ts:93`) permits only `http:`/`https:`, so `new URL('git@gitee.com:owner/repo.git')` throws (reproduced via live Node). The error is caught at `runner.ts:152` → `failed++` → silent perpetual failure. SSH is the standard private-repo mechanism, and **this PR's own auto-sync test fixtures use `git@gitee.com:...` SSH URLs** — but every runner test mocks `cloneOrPull`, so `validateGitUrl` is never exercised.

**Fix:** reject SSH/scp-style remotes at config load with a clear error and switch fixtures/docs to HTTPS, or add a real SSH clone path with equivalent safety checks.

_Cross-engine: Codex B3 [high] + Codex B2 [P2] + Claude risk-architect (reproduced) + correctness._

gitnexus/src/server/git-clone.ts
```diff
@@ -39,12 +42,16 @@ export const REPO_NAME_PATTERN = /^[a-zA-Z0-9._-]+$/;
  * clone root via path traversal.
  */
 export function extractRepoName(url: string): string {
-  const name = parseRepoNameFromUrl(url);
+  let trimmed = url.trim();
+  while (trimmed.endsWith('/')) trimmed = trimmed.slice(0, -1);
+  const withoutGit = trimmed.toLowerCase().endsWith('.git') ? trimmed.slice(0, -4) : trimmed;
```
magyargergo, line 49: **P2 — `extractRepoName` rewrite is an undocumented behavior change on the shipped web clone endpoint. [code-read]**

The old path delegated to `sanitizeRepoName` (`storage/git.ts:344`, `.replace(/[^a-zA-Z0-9._-]/g,'_')`) which **rewrote** bad chars; this inline parser **throws**. The real consumer is the web `/api/analyze` clone at **`api.ts:1585`**, where the throw surfaces as a **failed clone job** (`job.status:'failed'`). Concrete trigger: `https://dev.azure.com/org/proj/_git/My%20Repo` (the `%`/space fails `REPO_NAME_PATTERN`) previously cloned under a cleaned name and now hard-fails. It also now **accepts** Windows-reserved names (`CON`/`NUL`/`COM1`) the old reserved-name check blocked.

**Fix:** document the policy break and/or keep a sanitize path at the web boundary; restore the reserved-name guard.

_Cross-engine: Codex B3 [medium] + Claude facts/api-contract/correctness/risk._

gitnexus/src/core/auto-sync/runner.ts
```diff
@@ -0,0 +1,201 @@
...
+        const targetDir = getConfiguredRepoPath({ localPath: cloneRoot.root }, repoName);
+        await deps.cloneOrPull(remoteUrl, targetDir, undefined, {
+          allowedCloneRoot: cloneRoot.root,
+          expectedRepoName: repoName,
+          quarantineRoot: cloneRoot.quarantineRoot,
+        });
+        result.synced += 1;
+
+        const currentBranch = deps.getCurrentBranch(targetDir);
+        if (!currentBranch || !project.branches.includes(currentBranch)) {
```
magyargergo: **P2 — Multi-branch config is non-functional. [code-read]**

`buildCloneArgs` has no `--branch`; the runner only **filters** `project.branches.includes(currentBranch)` against the cloned *default* branch. `branches:[dev]` with default `main` → `['dev'].includes('main')` is false → `skippedAnalysis++` forever with no error; `branches:[main,dev]` → only the default is ever analyzed. The plural field + array/CSV parsing imply multi-branch support that doesn't exist.

**Fix:** fetch/checkout each configured branch (or per-branch worktrees), or reject multi/non-default-branch configs at parse time and document the single-branch limit.

_Cross-engine: Codex B3 [high] + Codex B2 [P2] + Claude correctness/risk._

gitnexus/src/core/auto-sync/runner.ts
```diff
@@ -0,0 +1,201 @@
...
+  const groupsToSync = new Set<string>();
+  const result: AutoSyncRunResult = { synced: 0, analyzed: 0, skippedAnalysis: 0, failed: 0 };
+
+  for (const project of config.projects) {
+    for (const remoteUrl of project.remoteUrls) {
+      try {
+        const repoName = extractRepoNameFromRemoteUrl(remoteUrl);
+        const cloneRoot = await deps.resolveCloneRoot(project.localPath);
+        const targetDir = getConfiguredRepoPath({ localPath: cloneRoot.root }, repoName);
+        await deps.cloneOrPull(remoteUrl, targetDir, undefined, {
```
magyargergo: **P1 — No timeout on the git subprocess → a hung remote freezes the whole loop. [code-read]**

`runGit`'s `spawn('git', …)` (`git-clone.ts:737`) has no `timeout`/`AbortSignal` and settles only on child `close`/`error`. The awaited `cloneOrPull` here never returns against a slow-loris / never-closing remote → `starter.ts` `.finally(() => running = false)` never runs → `running` stays `true` forever → every later interval tick hits the overlap guard and is skipped. The sequential loop also starves repos #2..N in the same run. This is an unattended loop pointing an unbounded git op at attacker-influenceable remotes on a timer.

**Fix:** wrap each `cloneOrPull`/`runFullAnalysis` in a wall-clock timeout (`AbortController` + `spawn(…, { signal })`) so a hang becomes a caught per-repo throw; add a watchdog that resets `running` after `k * intervalMs`.

_Single-engine, high-confidence: Claude reliability [P1] + adversarial; Codex did not explicitly hit it._

gitnexus/src/core/auto-sync/runner.ts
```diff
@@ -0,0 +1,201 @@
...
+        state[stateKey] = {
+          codeCommitId: currentCommit,
+          analyzedCommitId,
+          lastAnalyzeStatus: analyzeStatus,
+          lastSyncTime: now().toISOString(),
+        };
+
+        if (project.gitnexusGroup) {
+          const added = await deps.addRepoToGroup(project, repoName);
+          if (added) groupsToSync.add(project.gitnexusGroup);
```
magyargergo: **P1 — Grouped repos stop re-syncing after first membership. [code-read]**

`if (added) groupsToSync.add(...)`, and `addRepoToGroup` (line 185) returns `true` **only** when the repo is newly added to `group.yaml`. On every later cycle the repo is re-analyzed with new commits (`analyzed++`) but the group is **never re-synced** → the cross-repo graph goes stale and silently diverges from the per-repo indexes.

**Fix:** queue a group sync whenever a grouped repo was successfully analyzed (or has a pending failed group-sync), persisting pending state until `syncGroupByName` succeeds.

_Codex B2 [P1, confidence 100] + coordinator [code-read]._

gitnexus/src/core/auto-sync/runner.ts
```diff
@@ -0,0 +1,201 @@
...
+        }
+      } catch (err: unknown) {
+        result.failed += 1;
+        logger.error(`[auto-sync] Repository sync failed for ${remoteUrl}: ${(err as Error).message}`);
```
magyargergo: **P2 — Credential leak: raw `remote_url` is logged verbatim. [code-read]**

This logs the raw `remoteUrl`, and `buildCloneArgs` also places it in git argv (`/proc/<pid>/cmdline`). Auto-sync has **no token field**, so credentials-in-URL (`https://user:token@host/repo.git`) is the only private-repo path — meaning a PAT lands in stderr/journald/docker logs. The same file already redacts userinfo elsewhere (`normalizeGitUrlForCompare`).

**Fix:** redact userinfo before logging and reject/ warn on credential-bearing `remote_urls` at config load; add a regression test for `https://user:token@example.com/org/repo.git`.

_Cross-engine: Codex B2 [P1] + Claude ce-security [P2] + security-boundary [P3]._

gitnexus/src/core/auto-sync/path-security.ts
```diff
@@ -0,0 +1,173 @@
...
+      throw err;
+    }
+    if (stat.isSymbolicLink()) throw new Error(`Refusing symlink in auto-sync clone root path: ${current}`);
+  }
+}
+
+export async function assertDirectoryOwnerAndPermissions(root: string): Promise<void> {
+  const stat = await fs.stat(root);
+  if (!stat.isDirectory()) throw new Error(`auto-sync clone root is not a directory: ${root}`);
+  if (typeof process.getuid === 'function' && stat.uid !== process.getuid()) {
```
magyargergo, line 170: **P2 — Ownership / world-writable clone-root gate is a silent no-op on Windows. [code-read]**

The uid check is gated on `typeof process.getuid === 'function'` (absent on Windows), and `DANGEROUS_ROOTS`/`DANGEROUS_PARENT_ROOTS` are POSIX-only (no `C:\`, `%SystemRoot%`, `System32`). So on Windows any directory — including another user's, or `C:\Windows\System32\Tasks\…` — passes all guards.

**Fix:** on Windows verify writability/ownership via ACLs (or at minimum `fs.access(root, W_OK)` + a warning when `getuid` is absent) and add Windows system roots to the dangerous sets.

_Within-Claude: security-boundary [P2] + ce-security [P3]. (The distinct Windows reserved-name pass-through in `extractRepoName` is the cross-engine one — see the `git-clone.ts` comment.)_

gitnexus/src/core/auto-sync/state.ts
```diff
@@ -0,0 +1,61 @@
...
+  return `${path.resolve(repoPath)}|${branch}`;
+}
+
+export function shouldAnalyzeCommit(input: {
+  currentCommit: string;
+  previousAnalyzedCommit?: string;
+  previousStatus?: AutoSyncAnalyzeStatus;
+}): boolean {
+  if (!input.currentCommit) return false;
+  if (input.previousStatus === 'failed') return true;
```
magyargergo, line 40: **P2 — Failed analysis retries forever with no backoff. [code-read]**

`previousStatus === 'failed'` always returns `true`, with no failure counter, cooldown, or cap. A repo whose `runFullAnalysis` consistently throws (OOM, parser crash, permission error) triggers a full re-analysis (tree-sitter workers, GBs of heap) every `sync_interval_minutes` forever.

**Fix:** persist a failure count + last-failure time per repo/branch; retry immediately on a *new* commit but apply backoff / a max-retry cap for the same failing commit.

_Cross-engine: Codex B2 [P2] + Claude reliability [P2]._

magyargergo (commented):
# Tri-review digest — PR #2329 "adds an opt-in auto sync and analysis loop"

This PR adds an opt-in background loop (`AUTO_UPDATE_AND_ANALYZE_FLAG=1` + `~/.gitnexus/sync_config.yml`) that clones/pulls configured remotes on a timer and runs full analysis, plus a hardening pass on `server/git-clone.ts` (realpath/symlink containment, quarantine, owner/permission gates) that fulfils the deferred follow-up from #1325.

**Engines:** Claude ran the GitNexus swarm (6 lanes) + ce-review (8 personas). Codex (GPT-5.5/xhigh) ran ce-review **(B2, live)** and adversarial **(B3, live)**; its swarm leg **(B1) failed with a stream disconnect and produced no findings.** So cross-engine corroboration here means a Codex B2/B3 leg agreeing with a Claude lane — Engine B is only partially live; this is *not* "both engines ran all three methodologies." Codex B2 independently ran `tsc --noEmit` → **PASS**. *This is an automated multi-tool digest; verify each finding against the cited file:line before acting.*

---

## Verdict

- **Final verdict:** not production-ready — P1 blockers (SSH remotes rejected every cycle; no-timeout loop-freeze; grouped repos never re-sync) plus P2 regressions (web-clone `extractRepoName` break, non-functional multi-branch config) that would generate immediate bug reports. The hardening design itself is sound; the gating issues are the feature's clone/runtime semantics, not its security posture.
- **Branch hygiene:** clean feature/fix PR — single coherent commit, no unrelated churn, no merge-from-main noise.
- **Merge state:** mergeable (no conflicts), but BEHIND `main` by 2 commits — routine rebase before merge.

---

## What's solid (credit)

- **The path-containment / quarantine design is layered and sound.** SSRF ordering is correct (`validateGitUrl` runs before `fs.mkdir` on both clone and pull branches); arg-injection is closed (`REPO_NAME_PATTERN` + leading-`-` throw + `--` separator, repo name only ever a directory basename); YAML uses `JSON_SCHEMA` so `!!js/*` tags can't execute (no RCE); Azure PAT is injected via `GIT_CONFIG_*` not argv; config never silently drops a project (always raises an error); `shouldAnalyzeCommit` is correct; state save is atomic (tmp+rename) with corrupt-load → `{}`. Both engines independently attacked these and refuted them — the validation is a real feature, not theatre.
- **Tests + types are green locally.** Claude test-CI ran `npx vitest run` on the 3 suites → **125/125 pass, 0 skip [reproduced]**; Codex B2 ran `tsc --noEmit` → **PASS**.
- **Safe by default.** Flag unset/`0` → null, no timer, no I/O (confirmed in `parseAutoSyncFlag`/`maybeStartAutoSyncFromEnv`).
- **CHANGELOG correctly untouched** (release-owned per repo convention); branch is hygienically clean (single coherent commit, no unrelated churn).

---

## Headline findings (P1/P2 — cross-engine or single-lane with a concrete trigger)

### 1. SSH remotes fail on every cycle — `config.ts:139`, `git-clone.ts:93` · P1 · [reproduced]
`remote_urls` accepts any non-empty string, but `cloneOrPull`→`validateGitUrl` permits only `http:`/`https:`, so `new URL('git@gitee.com:owner/repo.git')` throws → `runner.ts:152` catch → `failed++` → silent perpetual failure. SSH is the standard private-repo mechanism, and **the PR's own auto-sync test fixtures use `git@gitee.com:...` SSH URLs** (but every runner test mocks `cloneOrPull`, so `validateGitUrl` is never exercised). **Fix:** reject SSH at config-load with a clear error and switch fixtures/docs to HTTPS, or add a real SSH clone path with equivalent safety checks. **Strongest finding — agreed by Codex B3 [high], Codex B2 [P2], Claude risk-architect [P1, reproduced via live Node], correctness, and the coordinator.**

### 2. `extractRepoName` sanitize→throw regresses the web clone endpoint — `git-clone.ts:44` (real site `api.ts:1585`) · P2 · [code-read]
The old `sanitizeRepoName` (`storage/git.ts:344`, `.replace(/[^a-zA-Z0-9._-]/g,'_')`) rewrote bad chars; the new inline parser **throws**. `api.ts:1585` calls it inside the web `/api/analyze` async job — this surfaces as a **failed clone job** (`job.status:'failed'`), not an HTTP 500. Concrete trigger: Azure DevOps `https://dev.azure.com/org/proj/_git/My%20Repo` (the `%`/space fails `REPO_NAME_PATTERN`) previously cloned under a cleaned name and now hard-fails. The PR's own tests flip 4 `'sanitizes…'` cases to `'rejects…'`, confirming intent — but it's an undocumented behavior change on a shipped endpoint, and it also now **passes** Windows reserved names (`CON`/`NUL`/`COM1`) the old path blocked. **Fix:** document the policy break and/or keep a sanitize path at the web boundary; restore the reserved-name guard. **Codex B3 [medium] + Claude facts/api-contract/correctness/risk.**

### 3. Multi-branch config is non-functional — `runner.ts:92` · P2 · [code-read]
`buildCloneArgs` has no `--branch`; the runner only **filters** `project.branches.includes(currentBranch)` against the cloned *default* branch. `branches:[dev]` with default `main` → `['dev'].includes('main')` is false → `skippedAnalysis++` forever with no error; `branches:[main,dev]` → only the default is ever analyzed. The plural field + array/CSV parsing imply multi-branch support that doesn't exist. **Fix:** fetch/checkout each configured branch (or per-branch worktrees), or reject multi/non-default-branch configs at parse time and document the single-branch limit. **Codex B3 [high] + Codex B2 [P2] + Claude correctness [P2] + risk [P2].**

### 4. No timeout on the git subprocess → a hung remote freezes the whole loop — `runner.ts:84` (root cause `git-clone.ts:737`) · P1 · [code-read]
`runGit`'s `spawn('git', …)` has no `timeout`/`AbortSignal` and settles only on child `close`/`error`. The awaited `cloneOrPull` at `runner.ts:84` therefore never returns against a slow-loris / never-closing remote → `starter.ts` `.finally(() => running = false)` never runs → `running` stays `true` forever → every later interval tick hits the overlap guard and is skipped. The sequential loop also starves repos #2..N in the same run. This is the auto-sync feature pointing an unbounded git op at attacker-influenceable remotes on a timer. **Fix:** wrap each `cloneOrPull`/`runFullAnalysis` in a wall-clock timeout (`AbortController` + `spawn(…, { signal })`) so a hang becomes a caught per-repo throw; add a watchdog that resets `running` after `k * intervalMs`. **Single-engine but high-confidence: Claude reliability [P1, conf 100] + adversarial [P2]** (Codex did not explicitly hit it).

### 5. Grouped repos stop re-syncing after first membership — `runner.ts:150` · P1 · [code-read]
`if (added) groupsToSync.add(...)`, and `addRepoToGroup` (line 185) returns `true` **only** when the repo is newly added to `group.yaml`. On every later cycle the repo is re-analyzed with new commits but the group is **never re-synced** → the cross-repo graph goes stale and silently diverges from the per-repo indexes. **Fix:** queue a group sync whenever a grouped repo was successfully analyzed (or has a pending failed group-sync), persisting pending group-sync state until `syncGroupByName` succeeds. **Codex B2 [P1, confidence 100] + coordinator [code-read].**

---

## Other findings (P2/P3)

- **Credentials in `remote_urls` leak** — `runner.ts:154` logs the raw URL verbatim, and `buildCloneArgs` puts it in git argv (`/proc/<pid>/cmdline`). Auto-sync has no token field, so creds-in-URL is the only private-repo path → a PAT lands in stderr/journald/docker logs. Redact userinfo before logging (the file already does this in `normalizeGitUrlForCompare`). **P2 — Codex B2 [P1] + Claude ce-security [P2] + security-boundary [P3].**
- **Windows ownership/world-writable gate is a no-op** — `path-security.ts:159` skips the uid check when `process.getuid` is absent, and `DANGEROUS_ROOTS` are POSIX-only, so on Windows any directory (incl. another user's, or `C:\Windows\System32\Tasks\…`) passes. **P2 — within-Claude: security-boundary [P2] + ce-security [P3].** (The distinct Windows *reserved-name* pass-through — `extractRepoName` now accepts `CON`/`NUL`/`COM1` — is the cross-engine one, Codex B2 [P2] + Claude api-contract, covered in headline #2.)
- **Failed analysis retries forever, no backoff** — `state.ts:30` always re-runs on `previousStatus==='failed'`; no failure counter, cooldown, or cap → an OOM/parser-crash repo hammers full analysis every interval. **P2 — Codex B2 [P2] + Claude reliability [P2].**
- **Same-basename remotes collide into one `targetDir`** — `runner.ts:81-83` derives the dir from the repo basename only, so two distinct remotes ending in `repo.git` map to one checkout; the second mismatches origin and fails every cycle. **P2 — Codex B2 [P2] + coordinator.**
- **No cross-process lock** — the `running` flag is intra-process only; auto-sync's `runFullAnalysis` bypasses the web API's per-repo job lock, so `gitnexus analyze /repo` + the background loop hit the same LadybugDB concurrently; two processes sharing `GITNEXUS_HOME` also last-writer-wins on the state/group files. **P2 — Codex B2 [P2] + Claude risk [P2] + reliability/maintainability residuals.**
- **Cross-device quarantine wedges a repo** — on clone failure, `quarantineAutoSyncPartial` does `fs.rename` from `local_path` to `~/.gitnexus/quarantine`; if those are different mounts the `EXDEV` is swallowed by `.catch(()=>{})`, the partial clone stays put, and the next run's `targetExists && !.git` throw (outside the try/catch) wedges the slot forever. **P3 — Claude correctness [E] + adversarial [F2].**
- **`targetExists && !.git` now throws** (`git-clone.ts:516`, outside the clone try/catch) — behavior change from "git clone into an empty dir self-heals"; a stale/empty dir now permanently wedges the slot on both web and auto-sync paths, never quarantined. **P3 — Claude adversarial [F3] + risk.**
- **`sync_interval_minutes` has no upper bound** — a huge value (`*60000 > 2^31-1`) overflows `setInterval`, clamps to ~1 ms, and becomes a back-to-back hot loop instead of "essentially never." **P3 — Claude adversarial [F4].**
- **`core → server` layering precedent** — `runner.ts`/`repo.ts` import `extractRepoName` from `server/git-clone.ts`, while `git-clone.ts` imports from `core/auto-sync/path-security.ts` → the first mutual `core↔server` dependency in the repo (git-grep-confirmed). Move the shared git-URL/path-security helpers into `core/` to keep the conventional server→core direction. **P1/P2 — Claude maintainability [P1] + docs-dod [P2].** (No written standard forbids it, per project-standards — treat as design debt, not a gate.)

---

## Test / CI gaps

- **Cross-platform registration missing.** The new platform-sensitive tests (symlink rejection, `chmod 0o777` world-writable, uid/permission gates) in `auto-sync.test.ts` and the additions to `git-clone.test.ts` are **not** in `scripts/cross-platform-tests.ts`, so Windows/macOS CI skips them — exactly the platform-divergent code that goes unverified. The tests also lack `it.skipIf(win32)` and would regress the #2314 Windows gate (`fs.symlink` → EPERM, `chmod` no-op). **Triply confirmed: test-ci + project-standards + testing personas + the TESTING.md rule.**
- **CLI activation gate untested** — `cli/index.ts:18` (`?.trim()==='1'` → fire-and-forget import) has no test; a module rename/typo would silently disable the whole feature.
- **Web-path regression untested** — no test proves the existing web `/api/analyze` clone still succeeds under the new *unconditional* symlink/realpath checks, nor that the `extractRepoName` throw path matches the documented API error shape.
- **No test/tsc/lint CI ran on this fork PR.** Tests are green locally only; the merged code is otherwise CI-unverified until a maintainer runs them.

---

## Docs / process

- **No user-facing docs** for `AUTO_UPDATE_AND_ANALYZE_FLAG` or the `sync_config.yml` schema — absent from README/RUNBOOK/`analyze --help`/i18n. A new operator can't use the feature without reading source; no example config is provided. **DoD §2.4 gap.**
- **PR-body path mismatch** — the description says `.gitnexus/sync_config.yml` (per-repo), but `getAutoSyncConfigPath()` resolves the **global** `~/.gitnexus/sync_config.yml` (`GITNEXUS_HOME`-overridable). Misleads operators creating the file.
- **PR title is not conventional-commit** (`"adds an opt-in auto sync…"`) → `pr-labeler.yml` applied no label; release-notes grouping breaks. Suggest `feat(auto-sync): add opt-in auto-sync and analysis loop`.

---

## CI status

`statusCheckRollup`: Vercel = auth-required infra **noise** (not a code failure); "Apply conventional label" = **pass**; "Validate PR title" = **skipped** (`pull_request_target` restricted on forks). **No unit-test / tsc / lint CI ran on this PR head.** Branch is `MERGEABLE` but 2 commits `BEHIND` main — routine rebase before merge.

---

*Automated digest from a two-engine tri-review (Claude swarm + ce-review; Codex ce + adversarial live, Codex swarm leg failed). Findings are evidence-tagged `[reproduced]`/`[code-read]` and were re-confirmed against the PR-head worktree, but this is tooling output — confirm each file:line and severity before merging or editing.*

magyargergo:
Suggestion, we could just create a `gitnexus watch` command?

im-fan:
## Summary

Adds an explicit `gitnexus watch` CLI entrypoint for long-running auto-sync, replacing env-flag startup with config-driven watch lifecycle controls.   The watch flow now supports
branch fallback, per-repo git timeouts, status/lock files, commit info output, and analyze failure threshold skipping.

## Motivation / context

Auto-sync needed a safer and more inspectable operating model than implicit startup via `AUTO_UPDATE_AND_ANALYZE_FLAG`.   This change makes watch behavior explicit through
`GITNEXUS_HOME/watch_config.  yml`, prevents duplicate watch processes per `GITNEXUS_HOME`, and records runtime state for operators.

## Areas touched

- [ ] `.  github/` (workflows, actions)
- [x] Docs / agent config only (`README.  md`, package README)

## Scope & constraints

**In scope**

- Add `gitnexus watch [init|start|restart|stop|status]`.
- Load watch config from `GITNEXUS_HOME/watch_config.yml`.
- Move watch runtime files under `GITNEXUS_HOME/watch/`.
- Add PID, lock, and status file handling for single-instance watch execution.
- Support `branches` with legacy `branch` compatibility.
- Try configured branches in order and skip repos when all branches fail or time out.
- Restrict auto-sync remotes to SSH SCP URLs for `github.com`, `gitlab.com`, and `gitee.com`.
- Add per-repo git timeout parsing via `repo_git_timeout`.
- Add analyze consecutive failure tracking and `analyze_failure_threshold`.
- Write human-readable `project_commit_info.txt`.
- Update CLI help, i18n descriptions, README docs, and unit tests.

- No Web UI changes.
- No change to ordinary `gitnexus analyze`.
- No reset command for auto-sync failure counters.
- No support for multiple concurrent watch instances under one `GITNEXUS_HOME`.
- No migration tool for older auto-sync config/state files.

## Implementation notes

- Removed implicit env-flag startup from CLI initialization;   watch is now an explicit command.
- `watch init` creates a default config without overwriting an existing file.
- Auto-sync state now lives at `GITNEXUS_HOME/watch/auto-sync-state.json`.
- `project_commit_info.txt` is inspection-only;   machine state remains JSON.
- Web/API clone repo-name compatibility is preserved through `extractWebRepoName`, while auto-sync keeps stricter remote/repo validation.
- Existing clone/pull gained auto-sync-only SSH support and branch-specific clone/pull behavior.

## Testing & verification

- [x] `cd gitnexus && npm test`

Focused unit coverage added/updated for:

- watch config parsing and validation
- branch fallback and unavailable branch handling
- analyze failure threshold skipping
- watch PID/lock/status lifecycle
- `watch init` and CLI help output
- auto-sync SSH clone path and timeout behavior
- web clone repo-name compatibility

## Risk & rollout

This changes the auto-sync operating model from implicit env startup to explicit `gitnexus watch`.   Operators using `AUTO_UPDATE_AND_ANALYZE_FLAG` must switch to `gitnexus watch
start` and create `GITNEXUS_HOME/watch_config.yml`.

Runtime artifacts now move under `GITNEXUS_HOME/watch/`;   stale older auto-sync state files outside that directory are not migrated automatically.

Index refresh is not required for existing repositories, but watch-managed repos will re-analyze when their configured branch commit changes or prior analysis failed.

im-fan:
@magyargergo  I have adjusted my PR in accordance with the review opinions. I hope you can review my PR for me again. Thank you

koriyoshi2041:
I checked the current head (`60b00c3d`) after the latest update. The remaining red checks look pretty mechanical now:

- `npx prettier --check ...` still reports 10 files needing formatting: the new watch/auto-sync files plus `git-clone.ts` and the three auto-sync test files. The PR autofix comment should probably handle most of this, or locally `npx prettier --write` on those files.
- Focused lint on the touched files has 2 hard errors, both in `gitnexus/src/core/auto-sync/path-security.ts`:
  - line 122: `path.parse(root)`
  - line 142: `path.parse(root)`
  The repo's custom rule wants `parseSourceSafe(...)` instead of direct `path.parse(...)` because of the Windows long-path crash guard.
- The one failed test from the uploaded `test-results.json` is:
  `test/unit/hooks.test.ts > windowsHide regression > gitnexus/src/server/git-clone.ts: every spawn-family options object contains windowsHide: true`

I reproduced that locally with:

```bash
cd gitnexus
npx vitest run test/unit/hooks.test.ts -t 'windowsHide regression'
```

Result: `expected 2 to be 1` at `test/unit/hooks.test.ts:447`, so the test sees two spawn-family sites in `git-clone.ts` but only one `windowsHide: true`. The existing `spawn('git', ['config', ...])` path has it, and `runGit()` has it; the mismatch is likely from a new/changed spawn-family options object without the field.

So the shortest CI cleanup path seems to be: run Prettier, replace the two direct `path.parse` calls with the safe helper, then add `windowsHide: true` to the remaining spawn-family options object (or refactor so the regression scanner still sees it).

im-fan:
@magyargergo  The gitnexus watch command has been added，and i have fixed the error mentioned in the pr. Please help review it again. Thank you

im-fan:
closed, add a new pr

PR #2330, opened 2026-06-30
fix(search): make vector distance threshold configurable

### LINKED ISSUE #2326: Search fails for symbols with short descriptions (CJK worst case)

## Summary

## Problem

```
```

```
Repo: my-project
Export: false
更新物料有效期 @param paramMap
```

## Combined Impact

## Reproduction

```
Original:      "采购订单自动审批流程"
```

```typescript
```

```
updateMaterialExpiryDate
更新物料有效期 @param paramMap
```

```
更新物料有效期 @param paramMap
```

## Related

### LINKED ISSUE #2332: Make vector search maxDistance configurable
Sub-issue of #2326

## Problem

Vector similarity thresholds are hardcoded with no env var or config:

- `local-backend.ts` — `WHERE distance < 0.6` (MCP vector index query)
- `local-backend.ts` — `rankExactEmbeddingRows(..., 0.6)` (exact scan fallback)

This disproportionately filters symbols with short descriptions (especially CJK), because:

1. Different embedding models have different distance distributions (multilingual/CJK often needs a higher threshold).
2. Short descriptions in long embedding text produce higher query–stored distances (see sibling sub-issue on embedding format).

## Proposed fix

```typescript
```

Or add to `EmbeddingConfig` for per-deployment tuning. Use one consistent default across MCP backend and embedding pipeline.

## Acceptance criteria

- [ ] `maxDistance` configurable via env and/or `EmbeddingConfig`
- [ ] All hardcoded `0.5` / `0.6` sites use the shared setting
- [ ] Documented in env/config reference
- [ ] Tests for override behavior

## Related

- Parent: #2326

koriyoshi2041:
## Summary

Adds `GITNEXUS_VECTOR_MAX_DISTANCE` so deployments can tune the semantic/vector cosine-distance cutoff without patching code. This addresses the hardcoded maxDistance slice from #2326 while leaving the CJK tokenizer and embedding-text weighting ideas for separate changes.

Defaults are unchanged:
- MCP local backend keeps `0.6`
- shared semantic search keeps `0.5`
- invalid, empty, or non-positive env values fall back to the path default

## Tests

- `npm test -- test/unit/exact-search.test.ts`
- `npx tsc --noEmit`
- `npx prettier --check src/core/embeddings/exact-search.ts src/core/embeddings/embedding-pipeline.ts src/mcp/local/local-backend.ts test/unit/exact-search.test.ts src/cli/i18n/en.ts src/cli/i18n/zh-CN.ts`

## Notes

This is intentionally the smallest operational tuning hook for #2326. It does not change tokenization, embedding text generation, or default result thresholds.


koriyoshi2041:
CI follow-up: the only red surface I see is tests / packaged install smoke (windows-latest) being cancelled, which then made CI Gate fail. I could not rerun it because GitHub requires repo admin rights for this run.

The related checks passed on this head: format, lint, typecheck, ubuntu coverage, Windows/macOS platform-sensitive, tree-sitter ABI on all three OSes, ubuntu packaged install smoke, benchmarks, CodeQL, Docker builds, gitleaks, dependency review.

Local checks before opening were:
- npm test -- test/unit/exact-search.test.ts: passed, 5 tests
- npx tsc --noEmit: passed
- npx prettier --check src/core/embeddings/exact-search.ts src/core/embeddings/embedding-pipeline.ts src/mcp/local/local-backend.ts test/unit/exact-search.test.ts src/cli/i18n/en.ts src/cli/i18n/zh-CN.ts: passed

Holding for a maintainer rerun or a concrete Windows packaged-install failure log.

gitnexus/src/core/embeddings/exact-search.ts
```diff
@@ -14,6 +14,17 @@ export interface ExactSearchChunk {
...
 
+export const DEFAULT_VECTOR_MAX_DISTANCE = 0.5;
+export const DEFAULT_MCP_VECTOR_MAX_DISTANCE = 0.6;
+
+export const getVectorMaxDistance = (fallback: number = DEFAULT_VECTOR_MAX_DISTANCE): number => {
+  const raw = process.env.GITNEXUS_VECTOR_MAX_DISTANCE;
+  if (raw === undefined || raw.trim() === '') return fallback;
+
+  const parsed = Number(raw);
+  return Number.isFinite(parsed) && parsed > 0 ? parsed : fallback;
```
magyargergo: **P3 / non-blocking nit.** Cosine distance is bounded to `[0, 2]`, but this accepts any finite positive number with no ceiling. `GITNEXUS_VECTOR_MAX_DISTANCE=5` (or a fat-fingered `50` for `0.50`) makes `WHERE distance < 5` and the JS filter `row.distance < 5` always-true, silently disabling relevance filtering and returning top-k for every query; a typo'd value silently falls back to the default with no signal. Both sibling env-readers here surface invalid input loudly — `getMaxFileSizeBytes` (`core/ingestion/utils/max-file-size.ts`) `warnOnce`s + clamps to a ceiling, and `parsePositiveInt` (`core/embeddings/config.ts`) throws — so this silent fallback is the odd one out. Consider accepting only `0 < parsed <= 2`, else `warnOnce` + fall back, so the effective threshold is observable. Not a merge blocker.

(This is also where a regression test asserting the env value reaches `WHERE distance < ${maxDistance}` would add the most value — that wiring is currently unasserted.)

_[code-read] — multi-tool tri-review_

magyargergo (commented):
# Tri-review digest — PR #2330 `fix(search): make vector distance threshold configurable`

Base `9c5a1743` → head `b5a5c7aa`. Diff: 6 files, +83/−8 (≈30 source lines). Worktree-verified.

## Verdict block
- **Final verdict:** `production-ready with minor follow-ups`. No blocking code defect survives verification; every defect probe was refuted across both engines. The five items below are all optional quality follow-ups. The PR is **BEHIND base** and should be rebased before merge, and its red CI is infra-only (see CI section), waivable per repo policy.
- **Branch hygiene:** `clean feature/fix PR` — single commit, minimal on-topic diff, no merge-from-main churn, no drive-by refactors.
- **Merge status:** `checks failing` — but **only** the aggregate `CI Gate`, which is red solely because of the flaky Windows packaged-install-smoke 20-minute timeout + Vercel auth-required; `mergeable = MERGEABLE`, all code-quality checks pass. Code-wise this is mergeable after a rebase.

## Engine breakdown (leg liveness)
- **Claude (8 lanes, all live):** swarm = `risk-architect`, `security-boundary`; ce = `correctness`, `adversarial`, `api-contract`, `testing`, `maintainability`, `project-standards`. (A 9th swarm lane, `test-ci-verifier`, returned a partial thought with no findings — its domain is covered by ce-testing + direct CI inspection.)
- **Codex / GPT-5.5 (xhigh), 3 legs:**
  - **B1 swarm (Solo)** — **fully live**: full multi-lane verdict + adversarial probes.
  - **B2 ce (report-only)** — **stream died mid-analysis** (`response.failed`); produced only a scope preamble, **no findings**. Not counted as an independent source.
  - **B3 adversarial** — **live**: verdict `approve`, "no material findings."
- **Two engines, not five confirmations.** Cross-engine corroboration here rests on **B1 (live) + B3 (live) + Claude**; B2 is excluded.

## Review bar for this PR (DoD)
Make the semantic/vector cosine-distance cutoff configurable via one env var, without changing default behavior when unset, across **every** search path (pipeline + MCP, vector-index + exact-scan fallback), with tests and bilingual help docs. **Met:** all four filter sites wired; defaults preserved when unset; en/zh help updated; unit tests added.

## Bottom line
**Clean PR. No blocking issues. No P0/P1/P2 defect survives verification.** Every correctness, injection, completeness, sci-notation, and caching probe across *both* engines was **refuted**. The only postable content is five non-blocking quality suggestions, plus one optional inline nit.

## Repository history considered
`fix(search): make X configurable` matches a direct, maintainer-merged in-repo precedent — `a7df8f86` *fix(search): make FTS stemmer configurable (#2307)* — so the `fix` type and shape are conventional. No prior regression in these embedding files surfaced in recall (prior-context.md). The bare-catch observability gap (#3) predates this PR.

## What is solid (credit)
- **Completeness verified across both engines.** All four vector-distance filter sites use the env-derived value: pipeline Cypher `WHERE distance < ${maxDistance}` (`embedding-pipeline.ts:629`) + `rankExactEmbeddingRows` (`exact-search.ts:58`); MCP Cypher (`local-backend.ts:2146`) + `rankExactEmbeddingRows` (`local-backend.ts:2196`). No hardcoded `0.5`/`0.6` distance literal remains in any search path. `[reproduced]`
- **Validation is genuinely well-defended.** `Number(raw)` (full-string parse, not `parseFloat`) + `Number.isFinite && > 0` rejects every injection/garbage string at the guard, before interpolation.
- **i18n parity clean and accurate.** `en.ts` + `zh-CN.ts` move in lockstep, single-line, semantically equivalent; documented defaults (`0.6` MCP / `0.5` elsewhere) match the constants. `[reproduced]`
- **Project standards clean.** No `any`/`as any` in code or tests; no bidi/hidden-Unicode; `git diff --check` clean; CHANGELOG correctly untouched (release-owned).
- **Default-parameter re-read is correct.** `maxDistance = getVectorMaxDistance(...)` is an ES default-param expression evaluated **per call**, so the env var is honored on every invocation — no module-load caching. `[reproduced]`

## Non-blocking suggestions (ranked)

### 1. No `(0,2]` upper-bound clamp and silent fallback (no warning) on `getVectorMaxDistance`
- **Risk:** cosine distance ∈ `[0,2]`, but the reader accepts any finite positive number with no ceiling and no log. `GITNEXUS_VECTOR_MAX_DISTANCE=5` (or a fat-fingered `50` for `0.50`, or `0x15`→`21`) makes `WHERE distance < 5` (`embedding-pipeline.ts:629`, `local-backend.ts:2146`) and the JS filter `row.distance < 5` (`exact-search.ts:58`) **always-true** → relevance filter silently disabled, top-k returned for every query (pollutes hybrid-search RRF). A typo'd value silently falls back to the default with **no signal**. (Lower-bound mirror: `1e-7` → near-every row rejected → silent-empty, compounded by the bare `catch` in #3.) `[reproduced]` (code-read both readers + node-repro of the parse)
- **Evidence to check:** `exact-search.ts:25` `return Number.isFinite(parsed) && parsed > 0 ? parsed : fallback;` (no clamp); `cosineDistance` range `exact-search.ts:28-42`. Both existing env-readers in this area surface invalid input **loudly**: `getMaxFileSizeBytes` (`core/ingestion/utils/max-file-size.ts:23-45`) `warnOnce`s and clamps to a ceiling; `parsePositiveInt` (`core/embeddings/config.ts:7-10`) **throws**. `getVectorMaxDistance`'s silent fall-back is inconsistent with both.
- **Recommended fix:** accept only `0 < parsed <= 2`, otherwise `warnOnce` + fall back, mirroring `getMaxFileSizeBytes`, so the effective threshold is observable.
- **Corroboration:** 5 **Claude** lanes (`security-boundary S3`, `ce-adversarial F1`, `ce-correctness R3`, `ce-api-contract RR-2`, `risk-architect F3`). Codex probed the same surface and deemed nothing material — i.e. strong intra-Claude convergence, **not** independent cross-engine. **Severity P3 / quality.**
- **Blocks merge:** no. *(This is the one item proposed as an optional inline comment.)*

### 2. Test gap — the env→Cypher `WHERE`-clause wiring is unasserted; two `getVectorMaxDistance` branches untested
- **Risk:** no test asserts the env value reaches the Cypher `WHERE` clause. The MCP dispatch test only asserts `cypher.includes('QUERY_VECTOR_INDEX')` — never the threshold value — so a regression that re-hardcoded `0.6` or memoized the threshold at module init would pass undetected. Separately, the empty/whitespace guard (`exact-search.ts:22`) and the non-finite branch (`exact-search.ts:25`) are unreachable by the current invalid-values loop `['0','-0.1','not-a-number']`.
- **Evidence to check:** `calltool-dispatch.test.ts:467` (`...includes('QUERY_VECTOR_INDEX')).toBe(true)`, no threshold assertion); the invalid-values loop in `exact-search.test.ts`. `semanticSearchWithContext` is also untested — but it has **no live callers** (grep-confirmed), so that sub-gap is informational. `[reproduced]`
- **Recommended fix:** a minimal test mocking `executeQuery`, set `GITNEXUS_VECTOR_MAX_DISTANCE`, call `semanticSearch`, assert the captured Cypher contains the threshold; add `''`/`'  '` and `'Infinity'` to the invalid-values loop.
- **Corroboration:** **STRONG cross-engine** — Codex **B1 Lane 4 (L4-F1, P3, live leg)** + Claude `ce-testing`, `risk-architect F4`, `security-boundary S4`, `api-contract`. **Severity P3 / quality.**
- **Blocks merge:** no.

### 3. Pre-existing bare `catch { … = new Map() }` swallows vector-query errors with no log
- **Risk:** any vector-query failure is swallowed silently; the operator cannot distinguish "no matches" from "query failed." **Pre-existing — NOT introduced by this PR**, but the new env var adds a (refuted) path to it.
- **Evidence to check:** `embedding-pipeline.ts:644-646` and `local-backend.ts:2161-2163`, both `catch { bestChunks = new Map(); }`, no `logger.warn`. `[reproduced]`
- **Recommended fix:** add a `logger.warn` in both catch blocks (out of scope for this PR; note for a follow-up).
- **Corroboration:** Claude `risk-architect F1` + `ce-adversarial R1` + `ce-correctness R1`; Codex did not raise it. **Severity: informational (pre-existing).**
- **Blocks merge:** no.

### 4. `getVectorMaxDistance` / `DEFAULT_*` placed in `exact-search.ts` rather than `embeddings/config.ts`
- **Risk:** `exact-search.ts` is otherwise pure cosine math with no `process.env` access; the env reader breaks that and is harder to find. `embeddings/config.ts` already owns the other `GITNEXUS_*` embedding readers (`resolveEmbeddingConfig`).
- **Evidence to check:** `exact-search.ts:17-26` vs `config.ts` (`GITNEXUS_EMBEDDING_BATCH_SIZE/SUB_BATCH_SIZE/THREADS/DEVICE`). `[code-read]`
- **Recommended fix:** move the three exports to `config.ts`, update the two import sites. Pure structure preference; no behavioral impact.
- **Corroboration:** Claude `maintainability M1` only (single-lane). **Severity P3 / nit.**
- **Blocks merge:** no.

### 5. `semanticSearchWithContext` dropped its explicit `0.5` → now env-sensitive with an unchanged TS signature
- **Risk:** the function is now silently env-dependent while its type signature is unchanged — a quiet semantic-contract shift. But **zero realized impact**: it has no live callers, and the env var is brand-new so no operator has it set.
- **Evidence to check:** `embedding-pipeline.ts:749` `semanticSearch(executeQuery, query, k)` (was `..., k, 0.5`). `[reproduced]`
- **Recommended fix:** optional one-line JSDoc noting the threshold now respects the env var. The change itself is intentional (makes all paths consistently env-aware).
- **Corroboration:** Claude `api-contract AC-1` (flagged shift) + `correctness probe 2` / `adversarial` / `testing` (all confirm zero impact); Codex did not flag. **Severity: effectively informational.**
- **Blocks merge:** no.

## Probed and refuted (credit the validation — NOT findings)
- **Cypher injection via the env var — REFUTED (HIGH conf, both engines).** `security-boundary S1` + Codex B1 Lane 5 + B3 + `ce-adversarial R-a`. `Number(raw)` is a full-string parse; any trailing payload (`"0.6 OR 1=1"`, `"0.6; DROP …"`) → `NaN` → fallback. The `isFinite && >0` guard (`exact-search.ts:24-25`) sits **before** both interpolation sites, so only a bare numeric literal can reach Cypher.
- **Scientific-notation breaks the Cypher parser — REFUTED (decisive, both engines).** `ce-adversarial R-c` + `risk-architect F2` + `security-boundary S2` + Codex B1. **Decisive:** `queryVecStr = `[${queryVec.join(',')}]`` (`embedding-pipeline.ts:611`, `local-backend.ts:2131`) already interpolates `String(float)` sci-notation embedding components (e.g. `3e-8`) into the **same** `CAST(... AS FLOAT[dims])` Cypher on **every** query — so the DB already accepts `e-` float literals in production; `WHERE distance < 1e-7` parses by the same grammar. Only the `e+` exponent at ≥1e21 (an absurd threshold) is un-exercised — folded into nit #1.
- **Module-load / default-param caching — REFUTED.** Per-call evaluation; env honored on the next search.

## Back-and-forth avoided by verifying
Rather than assume, the coordinator directly confirmed in the worktree: all five cited line numbers; the `cosineDistance` `[0,2]` range; the `max-file-size.ts` warnOnce+clamp and `config.ts` `parsePositiveInt` throw patterns; `semanticSearchWithContext` having no live callers; the `calltool-dispatch.test.ts:467` assertion that checks presence-not-value; the B2-died (`response.failed`) liveness fact; and the CI red being infra-only (`gh pr checks`). Each surfaced item is tagged `[reproduced]`/`[code-read]`.

## CI status
**RED is infra, not code.** All quality/typecheck/lint/format and all tests pass (ubuntu + windows + macos coverage, tree-sitter ABI, benchmarks, CodeQL, gitleaks). The only red checks: `tests / packaged install smoke (windows-latest)` (20-minute timeout, historically flaky/slow infra job, unrelated to this diff); `Vercel` (auth-required infra). `CI Gate` aggregates the Windows smoke failure. Codex B1 independently reached the same conclusion — "not production-ready" **solely** because `CI Gate` is red, code itself a "low-risk merge."

## Open questions
None remain after verification.

## Merge guidance
- No code changes required to merge; the five suggestions are optional follow-ups.
- **Rebase onto `main`** (BEHIND base) and re-run / waive the flaky Windows smoke + Vercel infra checks per repo policy.

---

*Automated multi-tool tri-review digest (Claude swarm+ce, Codex GPT-5.5 swarm+ce+adversarial, synthesis-weighted). Two engines, not five confirmations. Verify findings against the live diff before acting.*

magyargergo:
@koriyoshi2041 Thanks for this PR, please let me know if you have any questions 🙏

PR #2334, opened 2026-06-30
feat(embeddings): compact, description-forward embedding text (#2333)

### LINKED ISSUE #2326: Search fails for symbols with short descriptions (CJK worst case)

### LINKED ISSUE #2333: Reduce description dilution in embedding text for short doc comments
Sub-issue of #2326

## Problem

`buildMetadataHeader` (`text-generator.ts`) builds embedding text where description is a tiny fraction of the vector input:

```
Repo: my-project
Path: src/main/java/.../MaterialServiceImpl.java
Export: false
更新物料有效期 @param paramMap
[+ ~500 chars code body]
```

For short descriptions + large bodies, description is often **<3%** of embedding text; metadata + code dominate the vector. Description-based semantic search weakens for **all languages**; CJK is worse because descriptions are often 4–20 characters.

This compounds with hardcoded `maxDistance` (sibling sub-issue) and FTS gaps for CJK.

## Proposed directions (pick one or hybrid)

**Option A — Compact embedding header:** drop verbose metadata from embedding text; keep full metadata on graph nodes for display.

**Option B — Description first:** put description before metadata/code in embedding text.

**Option C — Description-only embedding for short docs:** when `description.length < threshold`, add a second embedding with description only; hybrid search combines both.

## Acceptance criteria

- [ ] Short-description symbols rank higher for description-shaped queries in vector/RRF search
- [ ] Full metadata still available in graph/UI context
- [ ] Regression tests for English and CJK short-description fixtures
- [ ] Approach documented (which option chosen and why)

## Related

- Parent: #2326

magyargergo:
## Summary

Symbols with **short doc comments** are weakly matched by semantic search. `buildMetadataHeader` (now `buildEmbeddingHeader`) prepended a verbose metadata block (`Repo`/`Server`/`Path`/`Export`) ahead of the description in every embedding text, so for a 4–20 char CJK comment or a brief English Javadoc the description was **<3%** of the vector input while metadata (~25–30%) and code body (~70%) dominated. This PR makes the embedding header **compact and description-forward**: name + description lead, the low-signal metadata lines are dropped from embedding text. Full metadata is untouched on the graph nodes (display/context tools read node properties, not embedding text).

Sub-issue of #2326, scoped to **Issue 3** (embedding-text dilution) only. The sibling sub-issues — hardcoded `maxDistance` and the missing CJK FTS tokenizer — are out of scope.

## What changed

- **`text-generator.ts`** — `buildEmbeddingHeader` emits `Label: name` then the (truncated) description, and no longer includes `Repo`/`Server`/`Export`/`Path`. The structural `Container:`/`Methods:`/`Properties:` lines for Class/Interface/Struct are preserved. Renamed from `buildMetadataHeader` (the old name no longer described the behavior).
- **`embedding-pipeline.ts`** — bump `EMBEDDING_TEXT_VERSION` `v2 → v3` so existing vectors are re-embedded with the new template on the next `analyze`. Removed the now-dead context-injection loop (the header no longer reads `node.repoName`/`serverName`).
- **Tests** — updated existing assertions to the new contract; added English + CJK short-description fixtures, short-label and structural-node coverage, description-before-body ordering guards, and `contentHashForNode` independence from `repoName`/`serverName`/`isExported`/`filePath`.

## Why Option A (and not B or C)

The issue offered three directions; the rationale is documented inline in `text-generator.ts`:

- **Option B (reorder only, keep metadata)** — rejected. The embedding model mean-pools over tokens, so the dilution is about token *proportion*, not position; reordering alone barely moves the signal.
- **Option C (separate description-only embedding + hybrid merge)** — deferred. Much larger change (new vector rows, schema, query-time merge); build only if Option A is measured insufficient.
- **Option A (compact header)** — chosen. Smallest diff that directly raises the description's proportional weight by removing ~25–30% metadata noise, and folds in B's ordering for free. Full metadata stays on graph nodes, so context/UI is unaffected.

## Rollout note

The embedding-text version bump invalidates all stored embedding vectors: the **first `analyze` after upgrade is a full re-embed**, not incremental. This is the intended, documented invalidation mechanism (same as prior template changes), not a regression. The tri-review follow-up commits (below) bump the version again to `v4`.

## Testing

- `text-generator`, `embedding-pipeline`, `embedding-chunking` unit suites green; `tsc` and prettier clean; pre-commit hooks pass.
- Ranking improvement (acceptance criterion: short-description symbols rank higher for description-shaped queries) is validated deterministically at the text-composition level. End-to-end ranking is best confirmed by re-analyzing a repo with CJK + short English descriptions and running an MCP semantic search — recommended manual check before release (note: `generateEmbeddingText` has one production caller — the analyze write path; MCP semantic search embeds the raw query via `embedQuery` and searches the stored vectors, so it is the end-to-end verification path, not a second consumer of the shared generator).

## Tri-review follow-up fixes

The PR #2334 tri-review (Claude swarm + ce personas; Codex gpt-5.5/xhigh swarm + ce + adversarial) findings are addressed by four further commits, each self-contained:

- **U3 — `fix(embeddings): restore bounded location signal` (headline).** The cross-engine adversarial finding: dropping `Path` entirely regressed path/service-qualified semantic search (e.g. `billing/handler` vs `identity/handler`), and FTS indexes only name/content/description — never `filePath` — so there is no backfill. Reinstates a **bounded** location signal (last 1–2 path segments, after the description) and bumps `EMBEDDING_TEXT_VERSION` `v3 → v4`.
- **U4 — `refactor(embeddings): remove dead context/EmbeddingContext plumbing`.** Resolves the residual below: the `context` parameter, the `EmbeddingContext` type, and the caller plumbing (`run-analyze.ts`, `api.ts`) are removed entirely.
- **U5 — `test(embeddings): harden structural ordering + header-integrity guards`.** Adds Interface/Struct description-ordering tests and replaces the narrow `not.toContain('\ntrue')` guard with a start-of-text header check.
- **U6 — `fix(embeddings): bound the migration re-embed failure window`.** Pre-existing hardening: moves the stale-row DELETE from a single up-front bulk delete to per-batch (just before each batch's INSERT), so an interrupted migration loses at most one batch instead of the whole index. Kept separate so it can be split to its own PR.

Closes #2333.


gitnexus/src/core/embeddings/text-generator.ts
```diff
@@ -58,33 +58,32 @@ const cleanContent = (content: string): string => {
...
+ * what display/context tools read. Only the embedding text changes here.
+ *
+ * Option B (reorder only, keep metadata) was rejected — mean-pooled embeddings
+ * weight by token proportion, not position, so reordering alone barely moves the
+ * signal. Option C (a separate description-only embedding + hybrid merge) is
+ * deferred to follow-up; build it only if Option A proves insufficient against
+ * real measurement. Any change to this template MUST bump EMBEDDING_TEXT_VERSION.
  */
-const buildMetadataHeader = (node: EmbeddableNode, config: Partial<EmbeddingConfig>): string => {
+const buildEmbeddingHeader = (node: EmbeddableNode, config: Partial<EmbeddingConfig>): string => {
```
magyargergo, line 98: **P2 — Path/service-qualified semantic-search recall regression** `[code-read]`

Dropping `Path`/`Repo`/`Server`/`Export` from the embedding header removes the only location discriminator from the stored vector — and FTS does **not** backfill it: `FTS_PROPERTIES = ['name','content','description']` (`gitnexus/src/core/search/fts-schema.ts:20`), and no symbol table indexes `filePath`. So a path/service token is absent from both the vector and BM25.

**Trigger:** a monorepo with two same-named symbols (`billing/handler` vs `identity/handler`, generic bodies). A query like `"billing handler"` loses the `billing` discriminator in the vector, and BM25 has no symbol-path field to recover it → arbitrary ranking between them. (Within one repo's index — cross-repo confusion is correctly avoided via per-repo DBs.)

**Cross-engine:** Claude (ce-adversarial + risk-architect) and Codex gpt-5.5 (adversarial) independently flagged this; Codex verified the FTS-no-`filePath` fact.

This is a deliberate #2333 Option-A tradeoff and is defensible — but the "FTS handles path-qualified queries" reassurance is false. Consider preserving a bounded location signal (last 1–2 path segments in the header) or making the hybrid/FTS path `filePath`-aware before relying on this; at minimum, make this the explicit query class to measure. `[code-read]`

magyargergo (commented):
# Tri-review digest — PR #2334 (`feat(embeddings): compact, description-forward embedding text` / #2333)

Base sha: `e148bc089a9323aead9d9029b7dadbbc9b0a0e31` · Head `5038c00` (merge-from-main; net diff = the 5-file embedding-text change).

## 1. Engine breakdown (all lanes live)

- **Engine A — Claude:** swarm `risk-architect` + `test-ci-verifier`; ce `correctness` + `adversarial` + `testing`. All produced real findings.
- **Engine B — Codex (gpt-5.5/xhigh):** `swarm` (6-lane) + `ce` + `adversarial`. All three artifacts verified to hold real findings (not just logs): codex-adversarial carries the headline path-drop finding; codex-ce returns "Ready to merge" with two residual risks; codex-swarm returns "production-ready with minor follow-ups" with a 6-lane table.

Cross-engine corroboration is strong on the headline; everything else is single-engine-verified or operational.

## 2. Credit — what's solid (both engines agree)

- **Change is correct.** Claude ce-correctness = CLEAN (every focus-area concern traced and refuted); Codex ce = "Ready to merge"; both swarms = "production-ready". The version-bump invalidation (`v2→v3` folded into `contentHashForNode`, embedding-pipeline.ts:100) correctly forces a full re-embed; `filePath`-independence of the hash is consistent because those fields no longer affect the vector either.
- **CI fully green.** CI Gate, typecheck, lint, tests, CodeQL, gitleaks, Docker builds, packaged smoke checks all pass. No workflow changes; `git diff --check` clean; no bidi/hidden-unicode.
- **Tests are strong, not weakened.** Negative `not.toContain` assertions are each paired with a positive presence check (no vacuous passes); hash-independence asserted for `filePath` AND `repoName`/`serverName`/`isExported`; no `it.skip`/`xit`, no deleted tests, no if-branching in vitests.
- **Refuted suspicions (validation is a feature).** Cross-repo vector confusion (per-repo DBs); empty/degenerate embedding text (header always emits `Label: name`); orphan/duplicate rows from path-independent hashing (compare is per-`nodeId`); `void context` masking a real bug (fields only ever read by the deleted loop); FTS impact (FTS indexes columns, not embedding text); hash collision; non-string-description crash (truncation path unchanged). All probed with a concrete trigger and refuted.

## 3. Headline finding (postable, inline-worthy)

**Path/service-qualified semantic-search recall regression — P2 [code-read].**
*Severity rationale:* the contributing lanes rated it P3 advisory (Claude ce-adversarial, as a "deferred-to-measurement tradeoff") and medium / "needs-attention" (Codex B3). Escalated one step to P2 because (a) it is cross-engine corroborated — the synthesis promotion rule lifts a corroborated finding one anchor step — and (b) the coordinator-verified FTS-no-`filePath` fact (below) removes the "FTS backfills path-qualified queries" mitigation that had justified the P3/advisory rating. It remains a deliberate, narrow-scope tradeoff, not a defect — hence inline-as-discussion, not merge-blocking.

Dropping `Path`/`Repo`/`Server`/`Export` from the embedding header removes the only location discriminator from the vector. Codex independently verified the reassuring half is false: **FTS does NOT backfill it** — `FTS_PROPERTIES = ['name','content','description']` (gitnexus/src/core/search/fts-schema.ts:20) and `File` is indexed on `['name','content']` only (line 24); **no symbol table indexes `filePath`**. So a path/service token is absent from both the vector and BM25.

- **Trigger:** a monorepo with two same-named symbols (e.g. `billing/handler` vs `identity/handler`, generic bodies, no/identical doc-comment). A query like `"billing handler"` loses the `billing` token in the vector and BM25 has no symbol-path field to recover it → arbitrary/wrong ranking between them. Within one repo's index (cross-repo confusion is correctly refuted — per-repo DBs).
- **Cross-engine:** Claude ce-adversarial (Finding 2) + Claude risk-architect (domain) + **Codex B3 adversarial** (`needs-attention`, "removes the only path/service discriminator… not backfilled by FTS"). Codex added the verified no-filePath FTS fact.
- **Frame:** this is a deliberate #2333 Option-A tradeoff and is defensible — but the project's "FTS handles path-qualified queries" reassurance is **false**. Recommend either preserving a bounded location signal (last 1–2 path segments in the header) OR adding `filePath` to the hybrid/FTS path **before relying on this**; at minimum, this is the explicit query class to measure.

## 4. Inline findings to post (exactly one)

| # | File:line (RIGHT side, in-diff) | Severity | Tag | Theme |
|---|---|---|---|---|
| 1 | `gitnexus/src/core/embeddings/text-generator.ts:79-96` (`buildEmbeddingHeader`, dropped Path/Repo/Server/Export block) | P2 | [code-read] | Path/service-qualified recall regression; FTS does not backfill `filePath` (fts-schema.ts:20,24) |

Anchor the comment to the `buildEmbeddingHeader` change; cite the trigger and the FTS fact. Model can't be run in the worktree, hence [code-read].

## 5. Body items (non-blocking)

- **Pre-existing re-embed durability cascade (operational, NOT inline).** The `v3` bump marks 100% of nodes stale, so the first migration `analyze` issues one bulk `DELETE` of all stale rows up front (embedding-pipeline.ts:353-373) before re-embedding; `executeWithReusedStatement` is autocommit (lbug-adapter.ts), so an interruption mid-re-embed leaves rows deleted until a clean re-run. **Pre-existing:** the diff does not touch lines 353-373; every prior `EMBEDDING_TEXT_VERSION` bump hit the same path; it self-heals on the next `analyze`. Raised by Claude ce-adversarial (Finding 1, P2) only; Codex B3 did not corroborate and Codex B1 risk found no blocker; risk-architect + ce-correctness confirm the full re-embed is intended/documented (GUARDRAILS "Embeddings vanished after analyze"). Optional pre-existing hardening: wrap migration delete+re-embed atomically (or defer the per-node delete to just before its re-insert).
- **Measure-before-rely.** R1 ("description ranks higher") is verified only at the text-composition level; true recall/ranking is unmeasured. Cross-engine: Claude risk-architect F-2 + ce-testing Q4/R1 + ce-correctness residual + Codex B1 Lane 4 + Codex B2 residual all agree. This is the manual verification gating the headline: re-analyze a repo with short English+CJK descriptions, run MCP semantic search, and specifically test path-token / duplicate-name queries against v3 vs v2.
- **Dead `context` plumbing (low follow-up).** `runEmbeddingPipeline` now does `void context;` (embedding-pipeline.ts:325) but `run-analyze.ts:1386` still computes `serverName = await readServerMapping(projectName)` and passes `{repoName, serverName}` — now pure overhead + a misleading call site. Claude-only (risk-architect F-1, ce-correctness, ce-adversarial); Codex B1 noted the no-op but didn't flag it. Already in the PR's Residual section; suggest a tracking issue + TODO with the issue number at the `void` line.
- **PR-body wording (Codex-only, verified true).** The body says analyze and MCP search-time "both consume the shared text generator," but `generateEmbeddingText` has exactly one production caller (embedding-pipeline.ts); MCP search-time embeds the raw query via `embedQuery` (mcp/core/embedder.ts:158) — it is the end-to-end verification path over stored embeddings, **not** a second consumer of the generator. Codex B1 Lane 6 + B2. Reword for accuracy.
- **Infra/hygiene (non-blocking).** Vercel = `FAILURE` "Authorization required to deploy" is a fork deploy-preview auth gate, not a code failure (both engines note it; all real CI checks green). Head `5038c00` is a merge-from-main commit; net diff is still the 5-file change. Confirm Vercel is non-required (or waive) before merge per branch policy.
- **Test gaps (low).** (a) `not.toContain('\ntrue')` (embedding-pipeline.test.ts:434) is a narrow supplementary guard — the real protection is the `toContain('Parses typed payloads.')` description presence check; (b) Interface+description and Struct+description ordering is untested though they share `buildEmbeddingHeader` (only Class+desc is asserted); (c) no pipeline-level test passing a live `{repoName,serverName}` context to prove it is now ignored. Claude ce-testing F1/F2/F3 + test-ci P3s; all low / correct-by-construction today.

## 6. CI status

`MERGEABLE` / `mergeStateStatus: UNSTABLE` solely due to Vercel auth gate. CI Gate, typecheck, lint, tests, CodeQL, gitleaks, Docker, packaged smoke = all green. No workflow changes; `ci-tests.yml` runs `npx vitest run --coverage` across ubuntu + cross-platform jobs.

## 7. Verdict

**production-ready with minor follow-ups** — no code blocker. The one item worth acting on **before relying on the feature** is the path-qualified recall regression: measure it (re-analyze + MCP semantic search on path-token/duplicate-name queries), and consider preserving a bounded location signal or making FTS/hybrid `filePath`-aware. Everything else is operational, pre-existing, or low-severity follow-up.

---
*Automated multi-tool tri-review digest (Claude swarm + ce personas; Codex gpt-5.5/xhigh swarm + ce + adversarial). Cross-engine confidence weighting applied; verify findings against the diff before acting.*

PR #2339, opened 2026-07-01
feat(search): add opt-in CJK bigram segmentation for FTS search

### LINKED ISSUE #2326: Search fails for symbols with short descriptions (CJK worst case)

### LINKED ISSUE #2331: CJK FTS preprocessing: segment text before indexing (workaround for missing tokenizer)
Sub-issue of #2326

## Problem

`CREATE_FTS_INDEX` only accepts `stemmer` and `stopwords` — there is no `tokenizer` parameter. LadybugDB's built-in tokenizer splits on whitespace only, so contiguous CJK text becomes a single token:

```
Indexed:   "采购订单自动审批流程"
Tokenized: ["采购订单自动审批流程"]
```

Sub-phrase searches (`审批流程`, `采购订单`, `自动审批`) fail. Mixed CJK+Latin without spaces (e.g. `ERP审批`) is also one token.

`GITNEXUS_FTS_STEMMER=none` (#2307) avoids English stemming on CJK, but does not fix tokenization.

Upstream: LadybugDB `CREATE_FTS_INDEX` needs a `tokenizer` parameter (separate feature request).

## Proposed fix

Before writing text to FTS-indexed properties, apply CJK word segmentation and insert spaces between words:

```
Original:      "采购订单自动审批流程"
```

Toggle via `GITNEXUS_FTS_CJK_SEGMENTATION` (values: `none`, `jieba`, etc.), default `none` for backward compatibility.

## Acceptance criteria

- [ ] CJK segmentation hook runs on FTS-indexed text when env var is set
- [ ] Sub-phrase CJK queries return BM25 hits after re-analyze
- [ ] Default (`none`) behavior unchanged
- [ ] Tests cover segmented vs unsegmented indexing

## Related

- Parent: #2326
- #2299, #2306, #2307

magyargergo:
## Summary

LadybugDB's full-text-search tokenizer splits only on the space character, so contiguous CJK (Chinese/Japanese) text indexes as one giant token. Searching for a real sub-phrase like `审批流程` ("approval process") inside a symbol description or file containing `采购订单自动审批流程` ("purchase order automatic approval process") returns nothing — the whole 10-character string is one opaque token. This adds an opt-in `GITNEXUS_FTS_CJK_SEGMENTATION=bigram` mode that makes that search actually work, with zero effect on default behavior when unset.

Closes abhigyanpatwari/GitNexus#2331.

## The premise dispute, resolved empirically

The issue assumed no native tokenizer option exists. A comment disputed that, citing LadybugDB's `tokenizer := 'jieba'` parameter as a simpler fix. Both claims were tested directly against the real `@ladybugdb/core@0.17.1` binary before any design work started:

- The comment was right that the parameter exists, but the fix it proposed is unsafe as stated: `tokenizer := 'jieba'` without a bundled dictionary (none ships in the npm package) causes an uncatchable native process abort (SIGABRT), not a JS exception.
- Even with a dictionary supplied and the crash avoided, `QUERY_FTS_INDEX` has no way to apply the same segmentation to a query string — so a query for the exact original indexed text still returns zero matches, because query-time tokenization is always whitespace-only regardless of how the index was built.

Since application-level segmentation of the query string is unavoidable either way, this PR uses character-bigram overlap instead — the same technique MySQL's `ngram` parser, Elasticsearch's `cjk` analyzer, and Lucene's `CJKBigramFilter` use by default. A shared `applyCjkSegmentationIfEnabled` function is called from both the index-write path and the query-read path, so the two can never diverge the way native tokenizer + app-side query segmentation would.

## What changed

- New `gitnexus/src/core/search/cjk-segmentation.ts`: `segmentCjkSpans` rewrites contiguous CJK Unified Ideograph runs into space-separated overlapping bigrams (`采购订单` → `采购 购订 订单`), plus the `GITNEXUS_FTS_CJK_SEGMENTATION` env-var triad mirroring the existing `GITNEXUS_FTS_STEMMER` pattern in `fts-indexes.ts`.
- `csv-generator.ts`: `content`/`description` are segmented before the existing #2317 whitespace-flatten, at the same six emission sites.
- `bm25-index.ts`: the search query is segmented once, up front, before either the pool-executor or direct `queryFTS` branch — still bound via `$query`, never string-concatenated.
- `local-backend.ts`: the MCP `query` tool now warns when a query contains CJK characters but segmentation is off, so an agent driving GitNexus has a way to discover the capability exists.
- `README.md`: documents the new env var, including that enabling it on an already-indexed repo needs a full `gitnexus analyze --force` — neither `--repair-fts` nor a plain incremental `analyze` re-segments already-stored content.

Scoped to CJK Unified Ideographs (Chinese text, Japanese Kanji) only; Hiragana/Katakana/Hangul are deliberately out of scope for this PR.

## Review process

This branch went through two independent review passes before this PR: a plan-level review (coherence/feasibility/adversarial personas against the technical design) and a Tier-2 code review (8 personas: correctness, testing, maintainability, project-standards, agent-native, learnings, performance, adversarial) against the actual diff. The code review caught and fixed a real bug worth calling out:

**A genuine O(n²) performance bug**, found and benchmarked by the performance reviewer: the original run-join loop indexed into the accumulated output string on every run boundary, which forces V8 to repeatedly flatten its internal rope representation. On realistic content (source code with inline CJK comments — this feature's actual target, not a contrived case), a single file at the *default* 512KB size cap took ~2.2 seconds to segment. Fixed by tracking state in a plain variable instead of re-indexing the growing string; verified a ~190x speedup (648KB: 4.76s → 25ms) with byte-identical output.

The adversarial pass also found an unbounded query-length path (nothing bounds search-query length before it reaches segmentation, unlike file content) — fixed with a 2000-character cap, past which a query still searches correctly, just without CJK segmentation.

### Residual findings (not blocking, tracked for follow-up)

Three findings didn't get fixed in this PR because they need a design decision or are narrow edge cases rather than mechanical fixes. Filing them as GitHub issues was blocked by this session's permission policy (autonomously creating public issues on a third-party repo without explicit authorization), so they're recorded here instead:

1. **P2 — Invalid `GITNEXUS_FTS_CJK_SEGMENTATION` throws inconsistently across query-serving processes.** `analyze` validates the env var once at startup and fails fast; long-lived server processes (MCP, web) never do, so a bad value only surfaces on the first query — and gets swallowed/mischaracterized differently by different callers (`hybrid-search.ts`'s bare `catch {}`, `local-backend.ts`'s "FTS indexes may not exist" mischaracterization, vs. `server/api.ts`'s correct surfacing). Needs a decision between fail-fast-at-startup vs. a distinguishable error class. (`gitnexus/src/core/search/bm25-index.ts:80`)
2. **P2 — Memory allocation profile for extreme all-CJK files.** `segmentCjkSpans`'s per-character array allocations create a large transient memory spike (~1.85GB estimated) for a File approaching the 32MB hard cap when `GITNEXUS_MAX_FILE_SIZE` is raised and the content is close to 100% contiguous CJK. Needs an algorithmic rewrite to bound peak memory independent of input size; raised independently by two separate review passes. (`gitnexus/src/core/lbug/csv-generator.ts`, `cjk-segmentation.ts`)
3. **P2 (advisory) — Invisible/combining Unicode characters** interspersed in a CJK run get isolated into their own run and wrapped with new visible separator spaces rather than treated as part of the adjacent run. Narrow precision edge case, not a crash. (`gitnexus/src/core/search/cjk-segmentation.ts:89`)

## Test plan

96 tests across the affected surface, including a real end-to-end proof (not a Cypher-seeded shortcut): a file with genuine contiguous CJK text goes through `loadGraphToLbug` → `streamAllCSVsToDisk` → COPY → `createFTSIndex` → `searchFTSFromLbug`, and a real sub-phrase query returns a BM25 hit with a positive score.

- [x] `segmentCjkSpans` unit tests: happy path, mixed CJK+Latin boundaries, punctuation resets, ASCII no-op, empty string, growth-factor and linear-scaling regression guards
- [x] Env-var triad: valid/invalid/default/caching, mirroring `GITNEXUS_FTS_STEMMER`'s test shape
- [x] Index-time: byte-identical default output, bigram-segmented `content` and `description` when enabled, CSV-flush safety margin re-verified with the new growth factor folded in
- [x] Query-time: unchanged by default, bigram-segmented when enabled (direct and MCP pool paths), existing `$query`-parameterization guard test unaffected
- [x] End-to-end: real sub-phrase search returns a positive-score BM25 hit
- [x] `npx tsc --noEmit` clean; `npx eslint` clean on all changed lines (pre-existing warnings on files only incidentally touched, not introduced by this diff)

No additional operational monitoring required beyond what's already in place: this ships fully opt-in (`GITNEXUS_FTS_CJK_SEGMENTATION` defaults to `none`), so default behavior is unaffected for every existing user and there is no rollout risk to a currently-running process. For operators who do enable it:
- **Healthy signal:** `gitnexus analyze --force` completes normally; a subsequent sub-phrase CJK query via `gitnexus query` or the MCP `query` tool returns expected results.
- **Failure signal:** an invalid value (anything other than `none`/`bigram`) throws immediately in `gitnexus analyze` with a message naming the valid options — this is expected fail-fast behavior, not a bug.
- **Rollback:** unset the env var (or set it to `none`) and re-run `gitnexus analyze --force`; there is no persisted schema change, only re-derivable indexed text.

## CI Failures Unresolved

One check remains red and is not fixable by any code change in this PR:

- **Vercel — "Authorization required to deploy."** ([authorization link](https://vercel.com/git/authorize?team=NexusCore&slug=nexus-core-77cd8306&teamId=team_VFCAhEd9yGJyInGx87eQQT9V&type=github&job=%7B%22headInfo%22%3A%7B%22sha%22%3A%22b8c265189cba618834bfb498be683a90c49e1a3a%22%7D%2C%22id%22%3A%22QmdfjeCT95uS4PFGLykDbhk33KCCjfmE8kFDt7SpHHeHrh%22%2C%22org%22%3A%22abhigyanpatwari%22%2C%22prId%22%3A2339%2C%22repo%22%3A%22GitNexus%22%7D)) — a Vercel deployment-preview authorization gate for PRs from a fork, not a test or build failure. It requires a maintainer with access to the `NexusCore` Vercel team to click through the link; no commit to this branch can satisfy it.

Every other check passed: CI Gate, the full test matrix (Ubuntu/macOS/Windows), tree-sitter ABI and packaged-install smoke tests on all three platforms, typecheck (core + web), lint, format, both CodeQL scans, gitleaks, dependency review, and PR metadata validation.

---

Claude Code


gitnexus/test/integration/fts-cjk-segmentation-search.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+afterAll(() => {
+  delete process.env.GITNEXUS_FTS_CJK_SEGMENTATION;
+});
+
+// Issue #2331's own example: "purchase order automatic approval process".
+const CJK_PHRASE = '采购订单自动审批流程';
+const FILE_BODY = `// ${CJK_PHRASE}\nexport function approve() {}\n`;
+
+withTestLbugDB(
```
magyargergo, line 33: **[P1] New LadybugDB integration test isn't registered in the sequential `lbug-db` Vitest project** `[reproduced]`

This test calls `withTestLbugDB(...)`, opening a real native LadybugDB connection exactly like its siblings `fts-description-search.test.ts` and `fts-fullfile-search.test.ts` — but unlike those siblings, it's absent from both `vitest.config.ts`'s `lbug-db` project `include` list (lines 51-78) and the `default` project's `exclude` list (lines 89-117). It therefore runs under vitest's normal parallel file execution instead of the `fileParallelism: false` sequential pool that exists specifically to avoid Windows file-lock/N-API destructor conflicts with the native mmap addon (see the comment at `vitest.config.ts:38-45`).

Confirmed independently by 4 lanes across both review engines (Claude ce-project-standards, Claude ce-testing, Codex swarm, Codex ce-code-review) — the strongest cross-engine signal in this review.

**Fix:** add `'test/integration/fts-cjk-segmentation-search.test.ts'` to both lists in `gitnexus/vitest.config.ts`, matching the two existing sibling entries. Also worth a look (lower confidence — its sibling `fts-fullfile-search.test.ts` is absent too, so this may be intentional): `gitnexus/scripts/cross-platform-tests.ts:74` lists `fts-description-search.test.ts` for the Windows/macOS cross-platform CI subset but not this new test.

gitnexus/src/core/search/bm25-index.ts
```diff
@@ -71,6 +82,17 @@ export const searchFTSFromLbug = async (
...
   repoId?: string,
 ): Promise<FTSSearchResponse> => {
+  // Applied once, up front, so every downstream branch searches with the
+  // same text the index was built from (#2331) — index-time and query-time
+  // segmentation must never diverge, since QUERY_FTS_INDEX cannot derive a
+  // tokenizer from the index it queries. No-op when segmentation is disabled
+  // (default). Skipped for pathologically long queries (see
+  // MAX_CJK_SEGMENTATION_QUERY_LENGTH) — the query still searches correctly,
+  // just without CJK sub-phrase segmentation.
+  const searchQuery =
```
magyargergo: **[P1] Two composition gaps between write-path and query-path text transforms** `[reproduced]`

**1. Queries over 2000 chars silently return zero results for verbatim-indexed text.** `MAX_CJK_SEGMENTATION_QUERY_LENGTH` caps query-side segmentation at 2000 chars, but the write path (`csv-generator.ts`) has no equivalent cap — File content is segmented up to the full 32MB `TREE_SITTER_MAX_BUFFER` ceiling. Empirically reproduced against the real embedded LadybugDB FTS engine: indexing `'審'.repeat(2001)` under `GITNEXUS_FTS_CJK_SEGMENTATION=bigram`, then querying with the identical 2001-char string returns `[]`; trimmed to exactly 2000 chars, it matches. This isn't "less precise" search as the adjoining comment claims — it's a guaranteed zero-result miss for text that's verbatim in the repo and indexed. The new MCP degraded-signal warning (`local-backend.ts:1999`) doesn't fire for this case either (it only fires when segmentation is *disabled*), so the failure is completely silent.

**2. The query never calls `normalizeFtsText`.** The write path composes `normalizeFtsText(applyCjkSegmentationIfEnabled(text))` (collapses `\r\n\t` to a space), but this query path only calls `applyCjkSegmentationIfEnabled(query)` — confirmed by direct grep, `normalizeFtsText` never appears in this file. A query containing a literal tab/newline adjacent to a CJK run (e.g. copy-pasted from an editor/log) won't match whitespace-normalized indexed text. Reproduced: query `审批\t流程` (tab) returns `[]`; `审批 流程` (real space) matches. This partially predates this PR (#2333), but this PR's own comment asserts index-time/query-time text "must never diverge" — both scenarios here falsify that.

**Suggested fix:** extend the existing degraded-signal warning to also cover `query.length > MAX_CJK_SEGMENTATION_QUERY_LENGTH` (cheapest fix for #1), and apply `normalizeFtsText` to the query alongside `applyCjkSegmentationIfEnabled` (fix for #2, restores the stated invariant).

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -1989,6 +1989,21 @@ export class LocalBackend {
...
       );
     }
+    // #2331: a CJK query against a server process resolving
+    // GITNEXUS_FTS_CJK_SEGMENTATION to 'none' silently misses sub-phrase
+    // matches with no other signal — this is the only place an agent driving
+    // GitNexus through the query tool can learn the capability exists.
+    try {
+      const { containsCjkIdeograph, getSearchFTSCjkSegmentation } =
+        await import('../../core/search/cjk-segmentation.js');
+      if (containsCjkIdeograph(searchQuery) && getSearchFTSCjkSegmentation() !== 'bigram') {
```
magyargergo: **[P2] Index/mode-drift observability gap** `[code-read]`

This warning only detects "CJK query while server segmentation mode is `none`" — it doesn't detect the reverse: server mode is `bigram` but the actual index was built under `none`/legacy (e.g. env var changed without a full `--force` re-analyze, or a plain/`--repair-fts` analyze ran after toggling the setting). In that case query text gets segmented into bigrams while the indexed rows are still unsegmented, causing silent recall regression with no warning for this direction.

Raised independently by both engines (Codex adversarial-review as "high"/no-ship, Codex ce-code-review, Claude risk-architect, Claude ce-correctness). To be clear about severity: this is *not* data corruption — no exceptions, no false positives, no ID corruption (traced and confirmed directly). It's a real, silent recall-loss risk for an opt-in feature that currently relies entirely on README documentation ("set the same value everywhere") with no runtime enforcement.

**Suggested fix:** persist the resolved CJK segmentation mode in repo metadata (mirroring the existing `schemaVersion` full-rebuild-trigger pattern in `run-analyze.ts`), and either force a full rebuild when it differs from the current env var, or at minimum extend this warning to also cover "server mode = bigram but index mode = none/legacy."

gitnexus/test/integration/csv-pipeline.test.ts
```diff
@@ -194,6 +195,93 @@ describe('streamAllCSVsToDisk', () => {
...
+        '订单',
+        '单自',
+        '自动',
+        '动审',
+        '审批',
+        '批流',
+        '流程',
+      ];
+      for (const bigram of expectedFileBigrams) {
+        expect(fileContent).toContain(bigram);
```
magyargergo, line 274: **[P2] Weak assertion — this test would pass even if segmentation silently became a no-op** `[reproduced]`

`expect(fileContent).toContain(bigram)` checks each of 9 expected 2-char bigram substrings, but every one of them is already present in the raw, unsegmented source phrase (`采购订单自动审批流程` contains `'采购'` as its literal first two characters) — so `toContain` passes whether or not segmentation actually ran. Unlike the `mode: none` test a few lines above (line 231: `expect(funcContent).not.toContain(FILE_CJK_PHRASE)`), there's no negative assertion proving the original contiguous run is gone / proving these are actually space-delimited tokens. A regression that silently turned `applyCjkSegmentationIfEnabled` into a no-op on the write path would not be caught here.

**Suggested fix:** assert the full space-delimited segmented phrase (or check token boundaries explicitly), and add a negative assertion that the raw contiguous phrase is absent from the transformed column — mirroring the `mode: none` test's negative-assertion pattern one test up.

magyargergo (commented):
# PR #2339 Tri-Review Digest — feat(search): add opt-in CJK bigram segmentation for FTS search

**Base:** `main` @ `316aaed928a03fa52be2814083040e64b9a85ae9` · **Head:** `feat/cjk-fts-bigram-segmentation` @ `4717a044c`
**Closes:** #2331 (sub-issue of #2326) · **Worktree:** `/workspace/.worktrees/pr-2339`

---

## 1. Engine breakdown

- **Claude engine, GitNexus PR swarm (A1)** — 6 lanes, all produced output: facts-historian, branch-hygiene, risk-architect, test-ci-verifier, security-boundary, docs-dod.
- **Claude engine, ce-code-review personas (A2)** — 7 personas, all produced output: correctness, testing, maintainability, project-standards, adversarial, performance, security.
- **Codex (gpt-5.5/xhigh)** ran all three requested legs, and **all three produced real findings** (none came back empty):
  - **B1 — swarm** (`gitnexus-pr-review`, Solo mode): 3 findings, verdict `not production-ready`.
  - **B2 — ce-code-review** (5 sub-personas: correctness, testing, maintainability/agent-native, performance, project-standards; one adversarial/reliability sub-lane disconnected mid-run and was excluded from its own synthesis, so B2's 5 findings are from the 5 personas that did complete): verdict `Ready with fixes`, 5 findings.
  - **B3 — native adversarial pass**: verdict `needs-attention` / "no-ship" framing, 2 findings (both `high`/`medium`).

13 Claude lane agents + 3 Codex legs (8 effective sub-lanes within B1–B3) reviewed this PR. Total distinct themes after dedup: ~9, of which 2 are top-tier merge-blocking/near-blocking, and the rest are follow-up-grade.

---

## 2. What's solid (credit where due)

- **Write/read symmetry via a single seam.** Both the write path (`csv-generator.ts`) and the read path (`bm25-index.ts`) route exclusively through the one exported `applyCjkSegmentationIfEnabled` entry point — no parallel/duplicated transform logic exists. Confirmed independently by Claude risk-architect, Claude maintainability, Claude project-standards, and Codex B1/B2. This is the right architectural seam for the feature's core invariant.
- **FLUSH_BYTES margin math is correct.** Re-derived independently by Claude facts-historian, Claude risk-architect, and Claude performance (3 lanes, cross-persona within Claude) plus spot-checked by Codex B1: `8MB + 2×(7/3)×32MB ≈ 157.3MB`, safely under the `MAX_STRING_LENGTH/2` (~256MB) test ceiling (~1.63x margin as stated, and Claude performance additionally showed the *true* unit-corrected margin is actually larger, ~8x, because the byte-vs-code-unit comparison is conservative in the safe direction — a documentation-precision nit only, not a defect). This is notable because a prior version of this exact PR (commit `823ce7ca`) got this math wrong once — it's now fixed and independently re-verified.
- **Regex is ReDoS-safe and O(n), confirmed empirically.** Claude security-boundary and Claude ce-security both traced `CJK_RUN_RE = /[一-鿿]{2,}/g` and found no backtracking hazard. Claude ce-performance went further and actually benchmarked `segmentCjkSpans` outside the test sandbox from 1M to 16M+ chars (including the real 32MB/11.18M-char TREE_SITTER_MAX_BUFFER ceiling) and confirmed flat linear scaling (~35–110ms/Mchar, worst-case single file ~450ms) — the single-regex-replace refactor in the final commit (`4717a044`) did not reintroduce the O(n²) bug that an earlier commit (`442ad2c7`) fixed.
- **No injection surface.** Claude security-boundary and Claude ce-security both confirmed the query reaches `QUERY_FTS_INDEX` only as a parameterized `$query` value (never interpolated), and CSV output only gets ASCII space separators injected into already-matched CJK spans — no delimiter/formula-injection vector introduced.
- **Opt-in, zero regression on default path.** `GITNEXUS_FTS_CJK_SEGMENTATION` defaults to `none`, and `applyCjkSegmentationIfEnabled` is a pure identity function when disabled — unanimous across every lane that checked it.
- **Mature diff, not a first draft.** Git log shows 10 commits: 5 initial feature commits followed by 5 explicit self-correction commits (margin-arithmetic fix, O(n²) fix, query-length guard + test-isolation fix, MCP-warning surfacing, single-regex simplification). This PR already went through real self-review cycles before any external reviewer touched it (see §7).

---

## 3. Headline findings (near-top / merge-relevant)

### F1 — New LadybugDB integration test not registered in the sequential `lbug-db` Vitest project
**Severity: P1 · 4-way cross-engine corroboration (strongest signal in this review)**

- **Lanes:** Claude ce-project-standards (conf 85), Claude ce-testing (conf 80, "T2"), Codex B1-swarm (finding #2, rated P2 by B1 itself but flagged "blocks merge: yes"), Codex B2-ce-code-review (finding #1, P1, conf 100). Two Claude personas + both independent Codex legs — genuine cross-engine agreement, not "consistent within one engine." (Synthesis note: B1 rated this P2; this digest upgrades it to P1 given the 4-way corroboration and concrete CI-stability trigger — flagging the upgrade for transparency rather than silently overriding B1's own rating.)
- **Trigger:** The new `gitnexus/test/integration/fts-cjk-segmentation-search.test.ts` calls `withTestLbugDB(...)`, opening a real native LadybugDB connection exactly like its siblings `fts-description-search.test.ts` and `fts-fullfile-search.test.ts` — but unlike those siblings, it is absent from both `vitest.config.ts`'s `lbug-db` project `include` list (lines 51–78) and the `default` project's `exclude` list (lines 89–117). It therefore runs under vitest's normal parallel file execution instead of the `fileParallelism: false` sequential pool that exists specifically to avoid Windows file-lock/N-API destructor conflicts with the native mmap addon.
- **Evidence [reproduced]:** `grep -n "fts-cjk-segmentation-search" gitnexus/vitest.config.ts` returns nothing (verified directly by the coordinator and by this synthesis agent). `gitnexus/vitest.config.ts:60-61` lists the two sibling files explicitly within the 51-78 include range.
- **Inline comment target:** `vitest.config.ts` is **not modified by this PR's diff**, so it cannot be inlined there directly. Anchor the inline comment to the new test file itself:
  - **File:** `gitnexus/test/integration/fts-cjk-segmentation-search.test.ts`
  - **Line:** 33 (`withTestLbugDB(`)
  - **Body should reference the real fix location:** `gitnexus/vitest.config.ts` lines 51–78 (`lbug-db` project `include`) and lines 89–117 (`default` project `exclude`).
- **Suggested fix:** Add `'test/integration/fts-cjk-segmentation-search.test.ts'` to both lists, matching the two existing sibling entries. **Additionally** (lower-confidence, third location — the critic gate flagged this and the coordinator confirmed it): `gitnexus/scripts/cross-platform-tests.ts:74` lists `fts-description-search.test.ts` for the Windows/macOS cross-platform CI subset, but not the new CJK test. Note this list is manually curated, not an exhaustive mirror of `lbug-db` — its OTHER sibling, `fts-fullfile-search.test.ts`, is also absent from it, suggesting omission here isn't automatically a bug. Worth a maintainer judgment call on whether the new test needs cross-platform coverage; the `vitest.config.ts` fix above is the load-bearing one.
- **Tag:** `[reproduced]` (direct grep against the real worktree, by coordinator and independently by this agent and the synthesis-critic).

### F2 — CJK query over 2000 characters silently returns zero results for verbatim-indexed text
**Severity: P1 (highest standalone severity in this review) · primarily single-lane, with a partial cross-engine echo**

- **Lane:** Claude ce-adversarial, confidence 90 — **empirically reproduced against the real embedded LadybugDB FTS engine** (temporary vitest probe, run then deleted, not committed): indexed a file with `'審'.repeat(2001)` under `GITNEXUS_FTS_CJK_SEGMENTATION=bigram`; querying with the identical 2001-char string returned `[]`; the same string trimmed to exactly 2000 chars matched; a 50-char sub-phrase from the same run also matched.
- **Root cause:** `bm25-index.ts` caps query-side segmentation at `MAX_CJK_SEGMENTATION_QUERY_LENGTH = 2000` (line 33) but the write path (`csv-generator.ts`) has **no equivalent cap** on indexed content — File content is segmented up to the full 32MB `TREE_SITTER_MAX_BUFFER` ceiling. A long, space-free, contiguous CJK query longer than 2000 chars is sent to `QUERY_FTS_INDEX` completely unsegmented, but the index only ever contains the bigram-segmented form — the raw contiguous token the query represents was never written, so the query returns nothing, not "degraded" results, contradicting the in-code comment's claim ("the query still searches correctly, just without CJK sub-phrase segmentation").
- **Compounding factor:** the PR's own new MCP degraded-signal warning (`local-backend.ts:1999`) only fires when `getSearchFTSCjkSegmentation() !== 'bigram'` — it does **not** fire for this case (segmentation is enabled, but the query exceeded the cap), so the failure is silent even to the PR's own new diagnostic.
- **Partial cross-engine echo:** Codex B2's `residual_risks` independently flags "long CJK queries over the 2000-character cap skip query-side segmentation silently... a narrow recall risk" — a weaker, lower-confidence version of the same root cause, reached independently by the other engine. This elevates F2 from "one lane's opinion" to a real, partially cross-confirmed, severity-overriding finding per the review's corroboration rules (severity + a verified trigger override single-lane sourcing here).
- **Evidence:** `gitnexus/src/core/search/bm25-index.ts:33,90-95`; `gitnexus/src/mcp/local/local-backend.ts:1999`.
- **Inline comment target:**
  - **File:** `gitnexus/src/core/search/bm25-index.ts`
  - **Line:** 92 (`const searchQuery =`)
  - **Severity:** P1
  - **Problem:** Queries above 2000 chars silently skip segmentation while the write path has no matching cap, producing guaranteed zero-result misses for indexed, verbatim-present CJK text — not "less precise" search as the adjoining comment claims.
- **Suggested fix:** Either extend the existing degraded-signal warning in `local-backend.ts` to also cover `query.length > MAX_CJK_SEGMENTATION_QUERY_LENGTH` (cheapest fix — makes the failure visible instead of silent), or raise/remove the cap and address the per-character allocation cost with a cheaper CJK-presence pre-check before running full segmentation.
- **Tag:** `[reproduced]` (real embedded LadybugDB FTS engine, temporary probe test, executed and deleted).

---

## 4. Index/mode-drift observability gap (merged theme)
**Severity: P2 · cross-engine, multi-thread — present as a recall-regression risk, not corruption**

This merges four independently-arrived-at threads describing the same underlying root cause, per coordinator-verification.md's explicit merge guidance:
- Codex B3 (native adversarial): "high"/no-ship framing — "Runtime CJK mode can diverge from how the index was built" (`bm25-index.ts:92-95`). Scenario: index built with `none`, server/CLI later runs with `bigram` (or vice versa), or a plain/`--repair-fts` analyze runs instead of a full rebuild — query text gets segmented against unsegmented indexed rows (or vice versa), causing silent recall loss with no warning for this specific direction.
- Codex B2 (ce-code-review): P2, "Server mode warning misses stale index-mode mismatches" (`local-backend.ts:1999`) — the MCP warning only detects "CJK query while server mode is `none`"; it does not detect "server mode is `bigram` but the index was built under `none`/legacy."
- Claude risk-architect: F1 ("CJK mismatch warning absent on HTTP server and CLI search paths") and F2 ("Incremental analyze after mode change silently corrupts index consistency" — risk-architect's own wording uses "corrupts" but the body clarifies this means *recall loss on stale rows*, not data corruption).
- Claude ce-testing: T3 ("README's '--force is required' claim is documentation-only; no code enforces it") — traced `run-analyze.ts`'s incremental-eligibility logic directly and confirmed no `meta.json` field records which segmentation mode built the existing index, unlike the precedent `schemaVersion` mechanism already used for other index-shape changes.
- Claude ce-adversarial explicitly **traced and refuted** the "genuine corruption" framing: incremental analyze without `--force` early-returns cleanly, `--repair-fts` only rebuils index structure over existing (unmigrated) columns, and un-migrated CJK spans just silently lose recall — no exceptions, no false positives, no ID corruption. This does not contradict the other four threads; per coordinator-verification.md, ce-adversarial refutes *data corruption* specifically while agreeing the recall-loss scenario is real.

**Net finding:** this is a real, moderate-severity, opt-in-feature **observability gap**: no persisted mode-stamp in `meta.json`/repo metadata, no drift check at analyze or query time, and MCP-only (not HTTP/CLI) partial warning coverage for one direction of the mismatch. Not a blocker (feature is opt-in, default `none`, and the README already documents "keep the env var identical everywhere" — readers just get no runtime enforcement of that requirement).

- **Inline comment target:**
  - **File:** `gitnexus/src/mcp/local/local-backend.ts`
  - **Line:** 1999 (`if (containsCjkIdeograph(searchQuery) && getSearchFTSCjkSegmentation() !== 'bigram') {`)
  - **Severity:** P2
  - **Suggested fix:** Persist the resolved CJK segmentation mode in repo metadata (mirroring the existing `schemaVersion` full-rebuild-trigger pattern in `run-analyze.ts`), and either force a full rebuild when it differs from the current env var, or at minimum warn loudly at analyze/query time. Also extend the warning's condition to cover "server mode = bigram but index mode = none/legacy," not just the reverse.
- **Tag:** `[code-read]` (all four threads reasoned from tracing `run-analyze.ts` control flow directly; no live-DB reproduction of the drift scenario itself was performed, though the individual code paths it depends on were independently traced by ce-adversarial).

---

## 5. Query never normalizes whitespace before segmentation
**Severity: P2 · single-lane finding, coordinator-verified, partially pre-existing**

- **Lane:** Claude ce-adversarial (confidence 65, discounted because root cause predates this PR).
- **Coordinator-verified [reproduced-equivalent]:** `grep -n "normalizeFtsText\|applyCjkSegmentationIfEnabled" gitnexus/src/core/search/bm25-index.ts` shows only `applyCjkSegmentationIfEnabled` is ever called — `normalizeFtsText` never appears in this file.
- **Trigger:** Write path composes `normalizeFtsText(applyCjkSegmentationIfEnabled(text))` (collapses `\r\n\t` → space, applied *after* segmentation). Query path only calls `applyCjkSegmentationIfEnabled(query)` — never `normalizeFtsText`. A query containing a literal tab/newline inside or adjacent to a CJK run (e.g. copy-pasted from an editor/log/diff) won't match the space-joined indexed tokens. Reproduced against the real FTS engine: `审批\t流程` (tab) → `[]`; `审批 流程` (real space) → matches.
- **Scope note:** the whitespace-normalization asymmetry itself predates this PR (#2333) and is not new code added here — this PR's contribution is composing a *second* transform (CJK segmentation) on top of a pre-existing gap, while its own code comment asserts index-time/query-time text "must never diverge," a claim this scenario falsifies (alongside F2 above).
- **Evidence:** `gitnexus/src/core/lbug/csv-generator.ts:204-208` (write-path composition) vs. `gitnexus/src/core/search/bm25-index.ts:92-95` (query-path, no `normalizeFtsText` call).
- **Inline comment target:**
  - **File:** `gitnexus/src/core/search/bm25-index.ts`
  - **Line:** 92
  - **Severity:** P2
  - **Suggested fix:** Apply `normalizeFtsText` to the query alongside `applyCjkSegmentationIfEnabled`, matching the write-path composition. Since this is partially a pre-existing gap (#2333), a minimal fix scoped to this PR would compose both transforms on the query side to at least restore the "must never diverge" invariant this PR itself asserts.
- **Tag:** `[reproduced]` (real FTS engine probe by ce-adversarial; import/call-site absence independently confirmed by coordinator via direct grep).

---

## 6. CSV bigram test's weak assertion
**Severity: P2 · cross-engine (Codex B2 + coordinator-verified), test-quality gap not a runtime bug**

- **Lanes:** Codex B2 (finding #3, conf 100) and coordinator-verification.md (direct read, confirmed true).
- **Trigger:** `gitnexus/test/integration/csv-pipeline.test.ts`'s `'bigram-segments both File content and Function description when enabled'` test (lines 237–282) asserts `expect(fileContent).toContain(bigram)` / `expect(funcContent).toContain(bigram)` for 9 expected 2-char bigram substrings. Every one of those substrings is already present in the raw, unsegmented source phrase (`采购订单自动审批流程` contains `'采购'` as its literal first two characters) — so `toContain` passes whether or not segmentation actually ran. Unlike the adjacent `mode: none` test a few lines above (line 231: `expect(funcContent).not.toContain(FILE_CJK_PHRASE)`), there is no negative assertion proving the original contiguous run is gone. A regression that silently turned `applyCjkSegmentationIfEnabled` into a no-op on the write path would not be caught by this test.
- **Evidence [reproduced]:** `gitnexus/test/integration/csv-pipeline.test.ts:274,280` — confirmed by direct read (coordinator) and independently by Codex B2.
- **Inline comment target:**
  - **File:** `gitnexus/test/integration/csv-pipeline.test.ts`
  - **Line:** 274 (`expect(fileContent).toContain(bigram);`)
  - **Severity:** P2
  - **Suggested fix:** Assert the full space-delimited segmented phrase (or explicit token boundaries via split/includes-as-whole-token), and add a negative assertion that the raw contiguous phrase is absent from the transformed column — mirroring the existing `mode: none` test's negative-assertion pattern one test up.
- **Tag:** `[reproduced]` (coordinator direct read; Codex B2 independently produced the identical finding).

---

## 7. Lower-priority items

- **Unjustified dynamic `import()` + defensive try/catch in `local-backend.ts` for a zero-dependency module** — Claude ce-maintainability (P2, conf 80, `[code-read]`). `cjk-segmentation.ts` has zero imports (pure regex/string code, no native bindings) — no circular-import or sandboxed-native-binding risk exists that would justify a dynamic import + bare catch, unlike the file's one legitimate precedent (`bm25-index.js`'s dynamic import, which has a cited native-binding failure mode, `#1489`). Suggested fix: convert to a static top-of-file import. Single-lane, `[code-read]`, not blocking.
- **Invalid env value silently swallowed as generic "FTS unavailable"** — cross-engine (Codex B2 finding #4, conf 75; Claude ce-correctness residual risk conf 35; Claude ce-security residual risk conf 35; Claude risk-architect S2). On a long-lived MCP/HTTP server process (which never calls `initialiseSearchFTSCjkSegmentation()` at startup, unlike the CLI `analyze` path), a typo'd `GITNEXUS_FTS_CJK_SEGMENTATION` value throws on every query; the bare `catch {}` around the CJK diagnostic block (`local-backend.ts:2005`, "Best-effort diagnostic only") and the outer `bm25Search` catch both swallow it without a log line, so the operator sees generic degraded/semantic-only search instead of a diagnosable config error. Fix: log at `debug`/`warn` in the catch, matching the sibling `bm25-index.js` import-failure pattern in the same file. `[code-read]`, low severity, not blocking.
- **Transient memory during bigram expansion — flag as low-confidence/unresolved, not a solid finding.** Codex B2 and B3 (same engine — "consistent within one engine," not cross-engine corroboration) both raised "max-size CJK runs allocate one string per bigram, transient heap pressure near the 32MB file cap." Claude ce-performance specifically benchmarked this scenario up to 11.18M chars and found no problem, but only measured wall-clock time, not peak RSS/heap — so it does not actually disprove the memory-safety concern, it simply didn't test for it. Net: neither side has direct evidence on peak memory; present as an unresolved residual risk requiring a real memory-profiling run, not as a contradicted or an inline finding.
- **Single-character CJK query over-triggers the MCP warning (Claude risk-architect, P3, `[code-read]`).** `containsCjkIdeograph` (used to gate the `local-backend.ts:1999` warning) matches any single CJK character, but `CJK_RUN_RE` requires `{2,}` — so a genuinely single-character CJK query (e.g. `审`) triggers the "enable `GITNEXUS_FTS_CJK_SEGMENTATION=bigram`" warning even though bigram mode can never segment a lone character and would not change that query's behavior. Minor UX/wording inaccuracy, not a functional bug.
- **Bigram transformation applied to non-FTS-indexed columns (Community, Tool, Section, Annotation, Delegate, Template, Module)** — Claude risk-architect F3, `[code-read]`. In bigram mode, graph API responses for these node types' `description` fields return bigram-transformed text even though these types aren't in `FTS_INDEXES` — no search benefit, only a storage/API-surface side effect. Pre-existing inconsistency (these types were never FTS-indexed), but this PR makes the visible effect larger. Follow-up grade, not blocking.
- **Doc/precision nits (Claude docs-dod):** three code comments reference an unresolvable offline "the plan" artifact (`cjk-segmentation.ts:18,81`, `csv-generator.ts:202`) — also independently flagged by Claude ce-maintainability; README's Hiragana/Katakana/Hangul-exclusion phrasing is technically correct but could be more prominent; MCP warning says "CJK characters" where "Han/CJK Unified Ideographs" would be more precise; README's "--force after enabling" phrasing doesn't mention that disabling (bigram→none) equally requires a full rewrite (Codex B1, P3). All P2/P3, non-blocking, easy follow-ups.
- **Testing-gap only items (no runtime bug):** `containsCjkIdeograph` has zero direct unit tests and the documented Hiragana/Katakana/Hangul no-op scope boundary is entirely unverified by the suite (Claude ce-testing T1, P1-for-coverage but not a runtime defect — verified correct by hand via tsx outside the suite); no boundary test at exactly 2000/2001 chars (Claude ce-testing T5, Claude ce-adversarial, Claude ce-security); O(n)/O(n²) timing regression test is a single-sample wall-clock comparison under CI coverage instrumentation (Claude ce-testing T4, flakiness risk not a correctness gap); only 1 of 6 `formatFtsDescription` call sites is directly tested (Claude ce-testing T7, low risk given shared-helper design).

---

## 8. Refuted / validated-safe (do not re-raise as findings)

- **FLUSH_BYTES growth-factor math** — confirmed correct after a prior wrong version (commit `823ce7ca` fixed it); independently re-derived by 3 Claude lanes + spot-checked by Codex B1. See §2.
- **Boundary-peek logic in `segmentCjkSpans`** (adjacent-run spacing, astral/surrogate neighbors, run-at-string-boundary) — Claude ce-adversarial mechanically tested every boundary shape (adjacent runs, existing-space neighbor, string start/end, astral-plane/Hiragana neighbor, lone single CJK char) and found correct behavior in every case; confirmed no read/write divergence is possible since both paths share the one function.
- **No ReDoS** — `CJK_RUN_RE = /[一-鿿]{2,}/g` has no nested/ambiguous quantifiers; confirmed by both Claude security-boundary and Claude ce-security, and independently benchmarked (2000-char cap × 10,000 calls ≈ 491ms total, ~0.05ms/call amortized).
- **No CSV/Cypher injection** — query reaches `QUERY_FTS_INDEX` only as a bound `$query` parameter (never interpolated); CSV output only gets ASCII-space separators inserted into already-matched CJK spans, verified against comma/quote/newline-mixed test vectors.
- **Toggling `GITNEXUS_FTS_CJK_SEGMENTATION` without `--force` is documented-graceful-degradation, not corruption** — Claude ce-adversarial traced the actual incremental-analyze control flow directly: a plain non-`--force` run early-returns and touches nothing when the tree is clean; `--repair-fts` only rebuilds FTS index structure over existing (unmigrated) stored columns; a genuinely incremental dirty-tree run re-segments only changed files. No exceptions, no false-positive matches, no ID corruption — only silent recall loss on stale rows, which is the same root cause merged into the observability-gap finding in §4 (present that as an observability gap, not as "corruption").
- **Regex `lastIndex` reentrancy** — `String.prototype.replace` with a global regex and a synchronous callback runs to completion atomically with respect to the event loop; no interleaving hazard.
- **No test silently skips** — the new integration test's `withTestLbugDB` FTS-availability gate is the same shared infrastructure used by every FTS-dependent suite in the repo, and `GITNEXUS_REQUIRE_FTS=1` is hard-coded in CI (`ci-tests.yml`), converting any skip into a hard failure — confirmed by Claude ce-testing (R1) and Claude test-ci-verifier. This is a residual local/dev-only risk, not a CI gap (separate from the F1 project-registration issue above).

---

## 9. CI status (from coordinator-verification.md, verbatim)

`gh pr checks 2339` (re-checked live, after all lanes finished) shows ALL jobs passing except:
- `tests / benchmarks (GITNEXUS_BENCH)` — FAIL (2m28s). Facts-historian compared this exact step across two CI runs of this same PR/branch and found it PASSED on the prior run (triggered by commit `b8c265189`); the PR touches zero scope-capture files. **Pre-existing flaky test, not a regression caused by this diff.**
- `CI Gate` — FAIL (2s). The aggregate required-checks gate; fails purely because the benchmark job above failed. Not an independent issue.
- `Vercel` — FAIL. "Authorization required to deploy" — fork-contributor auth gate, not a code issue (expected for any fork PR without prior Vercel team access).
- All previously-pending jobs (ubuntu/coverage, windows-latest, macos-latest platform-sensitive, Docker builds) now show PASS.

---

## 10. This PR is a mature diff, not a first draft

Git log for this branch shows 10 commits, all same-day, same-author: 5 initial feature commits followed by 5 explicit self-correction commits — `823ce7ca` "fix(search): correct CJK margin arithmetic, dedupe growth-factor constant", `442ad2c7` "fix(search): eliminate O(n^2) behavior in CJK segmentation run-join", `51aa3dcc` "fix(search): guard query-length, fix test isolation, correct scope claims", `b8c26518` "feat(mcp): surface CJK segmentation hint in query tool warnings", `4717a044` "refactor(search): simplify CJK bigram segmentation to a single regex replace". This PR had already caught and fixed a margin-arithmetic bug and an O(n²) performance bug on its own before any of the three review engines in this tri-review ever touched it. The two headline findings above (F1 vitest registration gap, F2 query-length-cap silent zero-results) are genuinely new discoveries beyond that self-review, not re-litigations of already-fixed issues.

---

## 11. Bottom line

- **Branch hygiene:** clean feature/fix PR — 10 linear, single-author, same-day commits, no merge noise, no unrelated churn (Claude branch-hygiene lane + Codex B1, independently).
- **Merge state:** effectively green — every required check passes except one confirmed pre-existing flaky benchmark job (and the aggregate gate that depends on it) and a fork-auth-only Vercel failure; neither blocks on code grounds.
- **Overall:** **production-ready with minor follow-ups.** The two P1 findings (§3) are real, verified, and each a small, well-scoped fix (a vitest project-registration line, and either a warning-condition extension or a query-cap change) — neither requires re-architecting the feature, and the feature ships opt-in/off-by-default either way. Recommend landing after F1 and F2 are addressed (or, at minimum, F1 given its CI-stability angle and F2 given it's a real correctness gap in the feature's own value proposition); §4–§7 are legitimate follow-up-grade issues worth tracking as separate issues, not silent deferrals.

## 12. Recommended merge-blocking action items (summary)

1. **F1 (P1):** Register `fts-cjk-segmentation-search.test.ts` in `vitest.config.ts`'s `lbug-db` include list and `default` exclude list. One-line fix, 4-way cross-engine corroborated, real CI-stability risk.
2. **F2 (P1):** Address the query-length-cap silent-zero-results gap — at minimum, extend the existing MCP degraded-signal warning to cover `query.length > MAX_CJK_SEGMENTATION_QUERY_LENGTH`. Empirically reproduced against the real FTS engine.
3. Everything else in §4–§7 is real but follow-up grade (P2/P3): not blocking given the feature is opt-in with an unchanged default, but worth tracking as issues rather than silently dropping (per this repo's own norm against deferring known issues).

---

*This is an automated multi-tool digest synthesizing 13 Claude review-lane agents and 3 Codex (gpt-5.5/xhigh) review legs, cross-referenced against direct coordinator verification in the live worktree. Verify findings independently before acting on them — in particular, re-run the two empirically-reproduced probes (F2's query-length-cap scenario and the whitespace-normalization scenario in §5) yourself if you intend to treat them as accepted before merge.*

Issue #623, opened 2026-04-01
1.5.x: Maximum call stack size exceeded for large repo


fastfrwrd:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

When attempting to run the analyze on the latest version of gitnexus in [grafana](https://github.com/grafana/grafana), I get the following error:

```sh
$ > npx gitnexus analyze --force

  GitNexus Analyzer

  Skipped 27 large files (>512KB, likely generated/vendored)
  █████░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  12% | Parsing code (20s)
  Analysis failed: Maximum call stack size exceeded
```

The same command seemingly works fine in 1.4.10:

```sh
$ > npx gitnexus@1.4.10 analyze --force

  GitNexus Analyzer

  Skipped 27 large files (>512KB, likely generated/vendored)
  ████████████████████████████████████████ 100% | Done
  Repository indexed successfully (130.8s)

  136,934 nodes | 531,698 edges | 5394 clusters | 300 flows
  LadybugDB 16.0s | FTS 12.9s | Embeddings off (use --embeddings to enable)
  /Users/foo/Workspace/grafana
  Context: AGENTS.md (updated), CLAUDE.md (updated), .claude/skills/gitnexus/ (6 skills)
  Note: 4 edges across 3 types inserted via fallback (schema will be updated in next release)
```

### Expected behavior

Expect the repo to be analyzed still

### Actual behavior

Analyze fails with an error

### Steps to reproduce

1. checkout [grafana](https://github.com/grafana/grafana)
2. run `npx gitnexus analyze`

### Environment

- OS: macOS 26.3
- Node: 24
- Version: 1.5.2

songjianping-cloud:
Same here

xulongwu4:
Explanation by AI:

<img width="1505" height="373" alt="Image" src="https://github.com/user-attachments/assets/6631247c-34ae-4a22-af87-89c092cfe35c" />

Arkh74278:
> Explanation by AI:
This helps, thank you!

seaward:
> Explanation by AI:
> 
[...]

PR #612, opened 2026-04-01
ci: use CHANGELOG.md for GitHub Release notes

abhigyanpatwari:
## Summary

Replaces `generate_release_notes: true` in the publish workflow with a step that extracts the current version's section from `gitnexus/CHANGELOG.md` and uses it as the GitHub Release body.

## Problem

The auto-generated release notes just list PR titles grouped by label — not useful for announcements and requires manual editing after every release.

## Solution

Since we already write a thorough, reviewed CHANGELOG entry as part of every release PR, the publish workflow now:

1. Extracts the section between `## [X.Y.Z]` and the next `## [` from `CHANGELOG.md`
2. Uses that as the GitHub Release body via `body_path`
3. Falls back to auto-generated notes if no CHANGELOG entry is found (safety net)

**Before:** Tag push → publish → auto-generated PR list → maintainer manually edits release notes
**After:** Tag push → publish → CHANGELOG section used automatically → done

## How it works

```bash
# awk extracts lines between ## [1.5.0] and the next ## [
awk '/^## \[1.5.0\]/{found=1; next} /^## \[/{if(found) exit} found' gitnexus/CHANGELOG.md
```

The CHANGELOG.md becomes the single source of truth — written and reviewed during the release PR, automatically used by the publish workflow.

Made with [Cursor](https://cursor.com)

PR #605, opened 2026-03-31
feat: gitnexus:keep marker preserves custom context sections

dp-web4:
## Summary

When `<!-- gitnexus:keep -->` is placed inside the GitNexus-generated block in CLAUDE.md or AGENTS.md, `analyze` only updates the stats line (node/edge/flow counts) instead of replacing the entire section with the verbose default template.

This lets users trim the auto-generated context to a lean custom format without it being overwritten on every reindex.

## Problem

Every `analyze` run regenerates the full GitNexus section in CLAUDE.md/AGENTS.md with the verbose template (tool descriptions, usage instructions, risk levels, etc.). Teams that maintain a lean custom format have to re-edit the file after every reindex.

## Solution

- If `<!-- gitnexus:keep -->` is present inside the GitNexus block, only the stats line is updated (e.g., "Indexed as **repo** (N symbols, N relationships, N execution flows)")
- If no keep marker is present, behavior is unchanged — full template replacement
- Broadened stats-line regex to match both "Indexed as" and "indexed by GitNexus as" formats
- If keep marker is present but no stats line is found, the section is preserved as-is

## Files changed

- `gitnexus/src/cli/ai-context.ts` — keep marker detection + stats-only update logic
- `gitnexus/test/unit/ai-context.test.ts` — unit tests for keep preservation and no-keep replacement

## Test plan

- [x] Unit tests for keep marker preservation
- [x] Unit tests for no-keep full replacement
- [x] Verified on 10+ repos in the dp-web4 collective with custom CLAUDE.md formats

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2340, opened 2026-07-01
fix(deps): pin Ladybug 0.18.0, validate the multi-writer deadlock fix

### LINKED ISSUE #2338: Validate Ladybug multi-writer deadlock fix (PR #605) against GitNexus large-repo ingestion
## Background

GitNexus's parsing pipeline already parallelizes, but LadybugDB writes remain one of the main constraints when indexing very large repos. Per @adsharma, Ladybug 0.17.1 either isn't performant enough or deadlocks once multi-writer is enabled.

## Fix to validate

[LadybugDB/ladybug#605](https://github.com/LadybugDB/ladybug/pull/605) — a classic lock-order inversion between two `TransactionManager` mutexes, hit when `TransactionContext::commit()` triggers an auto/force checkpoint concurrently with `TransactionContext::beginAutoTransaction()`. Resolved by switching to atomics.

Testing it requires a fresh Python-bindings build (as of 2026-06-21 per adsharma — "tonight's" build); we need to confirm when/whether an equivalent Node build is available before we can try it from GitNexus's ingestion path.

## Action items

- [ ] Confirm Node bindings incorporating #605 are available (or build from source) and get a version/tag to pin.
- [ ] Re-run GitNexus's large-repo ingestion benchmark with multi-writer enabled — adsharma reports this fix lets them import repos as large as Ladybug core itself, which is a useful upper-bound comparison.
- [ ] Verify no deadlock/regression under GitNexus's actual write pattern (concurrent per-file parse workers writing into one LadybugDB instance) before considering multi-writer as a default for large-repo `analyze` runs.

---
Collaboration with @adsharma (Ladybug). Discussion log: Discord, 2026-06-21.

magyargergo:
## Summary

GitNexus pins `@ladybugdb/core` at `^0.17.0`. Ladybug `0.18.0` bundles [LadybugDB/ladybug#605](https://github.com/LadybugDB/ladybug/pull/605), a `TransactionManager` lock-order-inversion deadlock fix @adsharma flagged as blocking multi-writer use ([issue #2338](https://github.com/abhigyanpatwari/GitNexus/issues/2338)). This PR pins that version, re-validates every version-coupled comment/regex the bump touches, and adds a test that directly forces and confirms the deadlock is gone — without changing GitNexus's production write path, which still serializes every write through one connection (`conn-lock.ts`).

## Findings — issue #2338 action items

**1. Confirm Node bindings incorporating #605 are available, pin a version.**
Confirmed and pinned. `@ladybugdb/core@0.18.0` (released 2026-06-30) contains #605's merge commit (`v0.18.0` is 31 commits ahead of it). The five platform `optionalDependencies` moved in lockstep. Checked Ladybug's issue tracker for known `0.18.0`-specific regressions before pinning — none found.

**2. Re-run GitNexus's large-repo ingestion benchmark with multi-writer enabled.**
Zero implementation progress on this item — not even a scaffold. A full real-world large-repo wall-clock benchmark is out of scope here; `bench/parse-throughput.md`'s "methodology + harness scaffold first, numbers later" convention is named only as the pattern a future follow-up should use.

**3. Verify no deadlock/regression under GitNexus's actual write pattern before considering multi-writer as a default.**
Added `test/integration/lbug-multiwriter-deadlock.test.ts`, which opens its own `Database` and multiple raw `Connection`s directly against `@ladybugdb/core` (bypassing `conn-lock.ts` on purpose, to prove the *native* fix rather than that GitNexus's app-level lock hides the bug) and forces the exact interleaving #605 targets (a checkpoint racing a concurrent reader/writer handoff).

- Ran this test 12 consecutive times against `0.18.0`: all passed cleanly (~2.5-5.5s each).
- Ran the identical test 4 times against `0.17.1` (pre-fix) as a comparison: 1 of 4 runs hung for the full 60s timeout and failed — a direct reproduction of the lock-order-inversion deadlock, consistent with #605's own description of it as timing-dependent rather than deterministic.
- **Production is unchanged.** `conn-lock.ts` still serializes every write through one connection. This validates the engine-level fix; it does not enable multi-writer as a default, per this action item's own explicit gate.
- `conn-lock.ts`'s serialization rationale was re-checked against `0.18.0`'s other corruption-adjacent fixes (`#612`: an out-of-bounds CSR index bug after delete+checkpoint, single-threaded and unrelated to concurrency; `#623`: a "narrow" fix for a database close/destroy-vs-GC race, explicitly not a full concurrent-access fix per the PR's own description). Neither addresses the risk `conn-lock.ts` guards against (two queries dispatched concurrently on one connection) — the lock's justification is unchanged and still required.

## Other changes

- Re-validated every `LADYBUGDB-CONTRACT`-marked comment/regex against `0.18.0` (`lbug-config.ts`, `lbug-adapter.ts`, `sidecar-recovery.ts`, `fts-indexes.ts`), plus two version-coupled spots the marker convention doesn't catch (`bridge-db.ts`'s `LBUG_OPEN_RETRY_PATTERNS`, `conn-lock.ts`'s docstring). Verified via upstream source diff (`v0.16.1..v0.18.0`) that every matched error-text string is unchanged.
- Added a stemmer-sweep test (`fts-stemmer-sweep.test.ts`) proving the bundled `0.18.0` FTS extension accepts all 28 entries in `SUPPORTED_FTS_STEMMERS`, not just the default `porter` (existing FTS tests only ever exercised `porter`).
- A live-trigger test for `isMissingShadowSidecarError` was attempted but abandoned after empirical probing showed it isn't reliably reproducible — even a SIGKILL-simulated crash at the exact moment a `.shadow` file existed on disk didn't reproduce the error on reopen. Documented as inspection-verified (source/changelog diff) instead of overclaiming test coverage that doesn't exist.
- Full `npm test` showed 41-55 failing test files with non-deterministic membership across two runs (`Worker crashed during startup` — a parse-worker resource-contention pattern under sandbox constraints running all 13,331 tests at once). None of the failing files touch `lbug`/`ladybug` or any file changed here. A sample of 6 failed files (132 tests) all passed cleanly re-run in isolation, confirming pre-existing environmental flakiness. The `lbug-db` vitest project (28 files, 336 tests — the group most directly affected by this bump) passed cleanly end-to-end, including after all code review fixes below.
- JSDoc comments across the touched `lbug` files were trimmed of task-narrative "re-validated on this bump" prose that would otherwise grow monotonically with every future version bump; only durable current-state facts remain (see the tri-engine review round below for the last of this cleanup).

## Code review (8-angle parallel review, medium effort)

Fixed in this PR:
- `lbug-multiwriter-deadlock.test.ts` hand-duplicated the native `Database` constructor call instead of reusing `lbug-config.ts`'s `createLbugDatabase` — and the comment justifying the duplication (needing to bypass it for a custom `checkpointThreshold`) was factually wrong, since `GITNEXUS_WAL_CHECKPOINT_THRESHOLD` already provides that override. Switched to `createLbugDatabase` + the env var.
- Every `QueryResult` from `conn.query()` (write loop, read loop, verify query) was never closed, leaking native cursors — fixed by routing through the existing `closeQueryResults` helper (`query-result-utils.ts`), matching `lbug-adapter.ts`'s established pattern.
- Cleanup (timers, connection/DB close, env var restore) only ran on the happy path — a timeout or a writer exhausting its retry budget skipped it. Moved all cleanup into the outer `finally`.
- Verified: 8 consecutive runs after the fix, all passing cleanly, and the full `lbug-db` project (336 tests) still green.

Residual (not fixed at that point, later resolved — see tri-engine review round below):
- `isDbBusyError` (`lbug-config.ts`) does not recognize LadybugDB's "Only one write transaction at a time is allowed in the system." message — documented in GUARDRAILS.md/RUNBOOK.md as a known, out-of-scope-for-this-PR production gap.
- The `.shadow`-file poll (5ms interval, used as proof the checkpoint-vs-reader race was actually exercised) is a best-effort proxy for a transient native artifact; a CI runner under different scheduling could in principle miss the window.
- The entire new test is skipped on Windows (`process.platform === 'win32'`) with no tracked follow-up issue — Windows-specific timing for this exact race is unvalidated.

- **Watch the cross-platform CI matrix** (`LBUG_NATIVE` group, all three OS jobs) on the first few runs after merge — this bump touches every platform's native binary, and `lbug-multiwriter-deadlock.test.ts`'s 60s timeout is only empirically sized against local Linux runs; a Windows-specific timing issue would show up as a new, unexplained timeout in that job specifically.
- **Watch for new GitHub issues mentioning**: `Could not set lock on file`, `Cannot open file .shadow`, `replay shadow pages under read-only mode`, or WAL checkpoint rename/remove errors — these are the exact strings this PR's re-validated regexes match.
- **Expected healthy signal**: no change in `gitnexus analyze`/`serve` crash-recovery behavior (WAL corruption recovery, orphan sidecar cleanup, FTS index creation) compared to `0.17.1` — this is purely a dependency bump plus new test coverage, no production code path changed.
- **Failure/rollback trigger**: a new crash or hang during `analyze`/`serve` that traces to `@ladybugdb/core` and reproduces only on `0.18.0` (not `0.17.1`) — revert the version bump in `package.json`/`package-lock.json` and re-open issue #2338 with the new findings.
- **Validation window**: through the next 1-2 GitNexus releases, since native dependency regressions often surface under real-world large-repo load this validation didn't cover (see the deferred large-repo benchmark item above). Owner: whoever merges/releases this PR.

## Test plan

- [x] `@ladybugdb/core` resolves to `0.18.0` in the lockfile
- [x] Full `lbug-db` vitest project (28 files, 336 tests) passes
- [x] New `fts-stemmer-sweep.test.ts`: all 28 stemmers accepted by the bundled extension
- [x] New `lbug-multiwriter-deadlock.test.ts`: 12/12 clean runs on 0.18.0; reproduced the deadlock 1/4 times on 0.17.1
- [x] `tsc --noEmit`, `eslint`, `prettier --check` all clean on the diff

---

## Second review round (`ce-code-review`, 8 personas: correctness, testing, maintainability, project-standards, reliability, adversarial, agent-native, learnings)

`correctness` independently re-verified this PR from scratch — re-cloned upstream LadybugDB at both `v0.16.1` and `v0.18.0` to check the source-diff claims, ran the new tests multiple times against the real binary, and ran the full `lbug-db` project. Zero findings. `agent-native` and `learnings` both came back clean/confirmatory (no parity gaps; no institutional-knowledge conflicts — this PR is confirmed as the expected follow-through on the adsharma/Ladybug collaboration thread that opened #2338).

Fixed in this PR:
- `skipUnlessFtsAvailable`/`FTS_UNAVAILABLE_NOTE` was duplicated verbatim between `lbug-core-adapter.test.ts` and the new `fts-stemmer-sweep.test.ts` — extracted to `test/helpers/fts-availability.ts`, both files now import it.
- `verifyConn` in the deadlock test's final verification block leaked a native connection if the count query threw before `.close()` — now pushed into the `readers` array the outer `finally` already closes.
- A latent TypeScript error (`tsconfig.test.json` catches it, the narrower `tsconfig.json` used for `src/` doesn't): `conn.query()` types as `QueryResult | QueryResult[]`; the verify block called `.getAll()` directly without narrowing. Fixed with an `Array.isArray` check.
- Repeated inline `InstanceType<typeof import('@ladybugdb/core').Connection>` expressions replaced with local `LbugConnection`/`LbugDatabase` type aliases.
- Verified: 12 consecutive runs of the deadlock test after these fixes, all clean; full `lbug-db` project (336 tests) green.

Residual at that point (later resolved or scoped further — see tri-engine review round below):
- `isDbBusyError` gap (same one noted above) — `reliability` and `adversarial` both independently re-derived this and rated it P1/confidence-100; `correctness` also flagged it as a residual risk.
- The deadlock test's timeout path doesn't cancel in-flight writer/reader loops before closing their connections (`reliability` + `adversarial`, both P2) — left open, requires a cancellation/abort-flag design beyond a one-line patch.
- `SUPPORTED_FTS_STEMMERS` exported as a live mutable `Set` (`maintainability` + `adversarial`) — **fixed in the tri-engine review round below.**
- Reader loop has no bounded retry for transient errors during the race window (`adversarial`) — left open; not observed to fail in 12+ local runs, but not proven safe either.
- `^0.18.0` caret range (`adversarial`) — a deliberate, already-documented trade-off matching the existing caret convention for this dependency, not a new gap.

## Third review round (tri-engine: Claude swarm + Claude `ce-code-review` + Codex GPT-5.5/xhigh running the same two methodologies + Codex native adversarial pass, [posted as a PR review](https://github.com/abhigyanpatwari/GitNexus/pull/2340#pullrequestreview-4608871890))

This round ran the identical swarm and `ce-code-review` methodologies on both Claude and Codex (GPT-5.5), plus a Codex-only adversarial pass, then synthesized and critic-gated the results before posting. Headline finding, cross-engine confirmed (Claude `ce-correctness` + Codex swarm independently, from the same source line, without seeing each other's output):

**Fixed in this PR:**
- The `withRetry` swap (from the second review round above) called `isRetryable` with no `afterMs`, so `computeBackoffMs`'s exponential-jitter formula gave a *deterministic zero-delay* on the very first retry (`floor(random() * min(capDelayMs, baseDelayMs))` is always `0` at `attempt=0`) — reintroducing the exact zero-delay hammering pattern the original hand-rolled loop's jitter was tuned to avoid. Fixed by returning an explicit `{ retry: true, afterMs: 1 + Math.floor(Math.random() * 3) }` on the retryable branch, which bypasses the exponential formula entirely and restores the guaranteed 1-3ms non-zero range. Re-validated: 12/12 consecutive clean runs against genuine `0.18.0`.
- The deadlock test's JSDoc referenced a nonexistent `docs/plans/2026-07-01-001-...` path (`docs/` is gitignored — a local planning artifact from the implementing session that was never committed) and unresolvable "see the PR description" pointers. Replaced with self-contained prose and durable references (issue/PR numbers, commit SHAs, GUARDRAILS.md/RUNBOOK.md) that stay resolvable after this PR merges.
- `SUPPORTED_FTS_STEMMERS` (flagged as a mutable-`Set` export in the second review round) is now typed `ReadonlySet<string>`, matching this codebase's own established convention for exported validation allowlists (`EVAL_SERVER_TOOLS`, `STRUCTURAL_LABELS`). Type-only change; both existing consumption patterns continue to work unchanged.
- GUARDRAILS.md's new "Known gap" bullet introduced a 4th label where the file's own documented convention is strictly Trigger/Do/Why — folded into the existing Why line.

**Re-confirmed, not escalated (already-disclosed residuals; both engines independently validated or refuted the same suspicions the earlier rounds raised — none newly blocking):**
- `isDbBusyError` gap — confirmed real by essentially every lane across both engines; two consistent framings: production-safe *today* given `conn-lock.ts`'s existing serialization (no new-architecture change makes it reachable yet), vs. a real gap that should close before any future loosening of the single-writer assumption. Still out of scope for this validation PR.
- Timeout-doesn't-cancel-in-flight-loops, reader-loop-no-bounded-retry, `.shadow`-poll-as-proxy, Windows-skip-without-issue-number — all re-confirmed present, none escalated by this round.
- Explicitly refuted this round (verified, not just re-asserted): partial-platform-lockstep concerns, `conn-lock.ts`'s rationale being invalidated by `0.18.0`'s other fixes, a suspected TypeScript error in the `withRetry` call, duplicate test registration, Cypher injection in the new test's query strings, the shadow-file-proxy and 60s-timeout-masking suspicions raised adversarially.
- Independent within-review reproduction of the PR's central claim: two Claude lanes separately ran the deadlock test against the pre-fix `0.17.1` binary (before a stale `node_modules` symlink in the review's own worktree was corrected) and got a 3/3 and a 3/6 hang rate respectively — both consistent with #605's own "timing-dependent, not deterministic" description, and neither contradicting the 0.18.0 fix.

Full digest: [PR review #4608871890](https://github.com/abhigyanpatwari/GitNexus/pull/2340#pullrequestreview-4608871890).

---

Claude Code


magyargergo (commented):
# Tri-Engine Digest — PR #2340 (abhigyanpatwari/GitNexus)

**"fix(deps): pin Ladybug 0.18.0, validate the multi-writer deadlock fix"**
Base `316aaed9` → Head `2c00c278` · 16 files, +358/-67 · Author magyargergo

---

## (a) Engine/leg breakdown — liveness confirmed

**Claude — GitNexus PR-swarm methodology (6 lanes, all live, all substantive):**
facts-historian, branch-hygiene, risk-architect, test-ci-verifier, security-boundary, docs-dod.

**Claude — ce-code-review personas (6 lanes, all live, all substantive):**
correctness, testing, maintainability, project-standards, reliability, adversarial.

**Codex (gpt-5.5/xhigh) — 3 legs, all live:**
- `codex-swarm` (B1) — solo Codex run adopting all 6 swarm personas. Live, full findings + verdict ("production-ready with minor follow-ups").
- `codex-ce` (B2) — solo Codex run adopting ce-code-review personas, `mode:report-only`. Live, completed cleanly with **"No actionable findings" / "Actionable findings: none"** — this is a genuine clean result, not a dead or empty leg (confirmed substantive tool-use transcript: diffs read, `gitnexus/impact` and `gitnexus/context` called, lockstep/`git diff --check` verified before concluding). Verdict: "Ready to merge."
- `codex-adversarial` (B3) — Codex's native red-team pass, no Claude equivalent. Live, narrow-scope: one finding only, verdict **"needs-attention"** on the `isDbBusyError` gap (see (b) below).

All 15 lanes/legs across both engines produced real, evidence-backed output. No empty or failed legs.

---

## (b) What's solid

**Core deadlock-fix claim — genuinely cross-engine reproduced.** The upstream LadybugDB #605 fix (native `TransactionManager` deadlock) landing in `0.18.0` was independently corroborated by: (1) the PR author's own reported 9/9-clean-on-0.18.0 vs 1/4-hang-on-0.17.1; (2) `ce-testing` (Claude) running the test 3x against 0.17.1 and getting 3/3 correct failures (`expected 'timeout' to be 'completed'`) — proof the assertion is a real, non-vacuous detector, not proof of the fix itself; (3) `ce-adversarial` (Claude) running it 6x against 0.17.1 and getting 3 hangs/3 passes (50%, vs the PR's own 25%) — same conclusion, plus a useful data point that reproduction probability is environment-sensitive; (4) **the coordinator's own final re-verification after fixing a stale `node_modules` symlink** (see (g) below) — 4/4 clean runs of the deadlock test and 28/28 of the stemmer sweep against a genuine, correctly-provisioned `@ladybugdb/core@0.18.0` binary. This last one is the cleanest [reproduced] evidence in the whole review that the fix, as pinned, actually works.

**Validated / refuted concerns** (explicitly tested and found not to hold — do not treat as open risks):
- Shadow-file-proxy (`.shadow` poll) unreliability — mostly REFUTED for this test's actual parameters (`ce-adversarial`, Q3): dense delay-free loops and ~10+ checkpoints per run make "idle reader at checkpoint time" implausible in practice.
- 60s-timeout masking a slow-but-not-deadlocked race — REFUTED (`ce-adversarial`, Q2): binary completed-vs-timeout race has no masking mechanism.
- Vitest 4 dropping numeric third-arg per-test timeouts — REFUTED (`ce-adversarial`): all 3 observed hang runs reported 60034-60038ms with the test's own custom message, confirming the numeric form is honored in v4.1.9.
- `conn-lock.ts`'s new docstring conflating the cross-connection deadlock (#605) with the same-connection concurrent-query-corruption hazard it itself guards against — REFUTED (`ce-adversarial` + independently re-confirmed by `ce-correctness`'s "verified clean" pass): the added text explicitly keeps the two hazard classes distinct.
- A suspected TypeScript type error in `withRetry`'s `isRetryable` lambda (missing `attempt` param) — REFUTED (`risk-architect`): `tsc --noEmit` confirmed clean; TS permits fewer params than the type requires.
- Partial-platform lockstep bump — REFUTED (`risk-architect`, `facts-historian`, `ce-correctness` all independently confirm all 5 platform `optionalDependencies` moved 0.17.1→0.18.0 together, no straggler).
- `conn-lock.ts`'s serialization rationale invalidated by 0.18.0's other fixes (#612, #623) — REFUTED (`risk-architect`): neither addresses the same-connection-concurrent-dispatch risk `conn-lock.ts` guards against.
- Duplicate test registration between `vitest.config.ts` and `cross-platform-tests.ts` — REFUTED (`risk-architect`): include-list vs. default-exclude-list, not a double-run.
- `TEST_FIXTURE_PREFIXES` gap for the new temp-dir prefix — REFUTED (`risk-architect`): transitively covered by the existing `gitnexus-lbug-` prefix via `startsWith`.
- Cypher injection in the new test's query strings — REFUTED (`security-boundary`, F5): all interpolated values are pure integer loop arithmetic or compile-time constants, no user input reaches any query string.

Note: these refutations are almost entirely **Claude-internal** (`risk-architect`, `ce-adversarial`, `security-boundary`, `ce-correctness` cross-checking each other) — consistent-within-one-engine, not cross-engine-confirmed. `codex-adversarial` (B3)'s red-team pass this round was narrowly scoped to the `isDbBusyError` classifier and did not independently re-litigate this list; it should not be read as a second-engine confirmation of these specific refutations.

---

## (c) Headline finding

### `withRetry` swap gives the deadlock test's first retry a deterministic **zero-millisecond** delay — contradicts its own "jittered delay" narrative
**Severity: P2 · Corroboration: STRONG — genuine cross-engine match** (Claude `ce-correctness` + Codex `codex-swarm`, independently) · **Tag: [code-read]** (verified by direct, deterministic computation of the formula — not sampled/executed, but mathematically certain)

Commit `2c00c278` replaced a hand-rolled retry loop (`sleep(1 + Math.floor(Math.random()*3))`, uniform 1-3ms, every attempt) with `gitnexus-shared`'s `withRetry`, called with `{ baseDelayMs: 1, capDelayMs: 3 }`. Tracing `computeBackoffMs` (`gitnexus-shared/src/integrations/retry.ts:52-65`):
- attempt=0: `exponential = 1*2^0 = 1`, `upper = min(3,1) = 1`, `delay = floor(random()*1)`. Since `Math.random()` is always `< 1`, this is **always 0** — the very first retry after a collision has zero delay, deterministically, on every call.
- attempt=1: delay ∈ {0,1}. attempt≥2: delay ∈ {0,1,2}.

Both `claude-ce-correctness.md` (F1, P2/confidence 75) and Codex's `codex-swarm.md` (Lane 4, P2/confidence high) derived this same fact independently, from the same source line, in the same session, without seeing each other's output. That is the strongest possible corroboration pattern this review produced (different engine, same contract, same conclusion) — per this review's own weighting rules this is leading and should not be undercounted despite its "only" P2 severity.

**Why it matters:** the surrounding JSDoc (unchanged by the refactor) still says the config was "validated across 5 consecutive local runs" with "a small jittered retry delay" that avoids tripping a *different* native guard (`"Timeout waiting for active write transactions..."`). All of the PR's cited "9/9" and "12 consecutive" validation runs predate this refactor commit — there is no recorded re-validation count for the actual timing that ships. `claude-ce-maintainability.md` and `claude-ce-adversarial.md` (F2) both independently touch the same math as a secondary observation, reinforcing it further (all within Claude, but corroborating the core fact).

**Suggested fix** (Codex's `codex-swarm.md` Lane 4 recommends this specific shape; it is the only fix that actually works): have `isRetryable` return an explicit `{ retry: true, afterMs: 1 + Math.floor(Math.random()*3) }` to preserve the old 1-3ms jitter range exactly. Note: bumping `baseDelayMs` alone does **not** fix this — no value of `baseDelayMs` eliminates 0 from `floor(random() * upper)`; e.g. `baseDelayMs: 2` only reduces attempt=0's zero-delay probability from 100% to 50%, it doesn't eliminate it. The `afterMs` override is the only mechanism in `withRetry` that bypasses the exponential-backoff formula entirely. Either way, re-run the consecutive-pass validation against real 0.18.0 post-fix and update the JSDoc/PR counts to match the code that actually ships.

---

## (d) Other findings, ranked

### `isDbBusyError` does not recognize `"Only one write transaction at a time is allowed in the system."`
**This is the PR's own already-disclosed, headline residual — not a new finding.** Confirmed by essentially every lane across both engines (`lbug-config.ts:381-389`; matcher only checks `busy`/`lock`/`already in use`). Already documented in the PR body, `GUARDRAILS.md`, and `RUNBOOK.md`. Two framings coexist and are **not actually contradictory**:
- `risk-architect` (Claude): production-safe **today**, given the current architecture — `conn-lock.ts` already prevents the concurrent-connection scenario that triggers this exact error in production; cross-process races instead hit the already-recognized "Could not set lock" file-lock error.
- `codex-adversarial` (B3): rates it more urgently — verdict `needs-attention`, explicit "No-ship" framing — arguing the classifier should be fixed regardless, since any future architecture change (or the MCP-server-plus-concurrent-`analyze` scenario `ce-reliability` traces through `withLbugDb`'s own stated cross-process-contention purpose) could make it reachable.

Read together: safe under the architecture as it stands today, but a real gap that should be closed before any future loosening of the single-writer assumption — both framings are correct at their own scope. `ce-reliability` (Claude, F1, P1/confidence 100) independently re-derived the same gap with the strongest reachability argument (`withLbugDb`'s docstring explicitly names cross-process DB contention, e.g. MCP server + a concurrent `analyze` run, as its purpose) — worth noting this is in some tension with `risk-architect`'s "only cross-process races occur and those hit a different, already-recognized error" claim; neither lane fully reconciles this, and it's a legitimate open question for a human to settle, not a resolved contradiction.

### Test file's own new JSDoc header still has task-narrative artifacts the sibling JSDoc-trim commit's stated policy should have caught
**Severity: P2 · Corroboration: single Claude lane (`ce-maintainability`) · Tag: [reproduced]** (ran `find . -path "*/docs/plans*"` and `find . -iname "2026-07-01*"`, both empty)

`lbug-multiwriter-deadlock.test.ts:19` references `docs/plans/2026-07-01-001-...` — a path that does not exist anywhere in the repo (confirmed by execution, not just inspection) and never will for anyone but the PR author's own session. Lines 38-48 also carry "flagged in the PR description... see the PR description for the full record" pointers. Commit `f999bdd5` (the JSDoc-trim commit, same PR) explicitly states its own rationale for removing exactly this kind of narrative from 5 production files — but never touched this test file, which was added/re-touched across 3 earlier commits in this same PR. The policy was applied inconsistently within the PR's own diff. Single-lane, but execution-verified — postable.

### Test's concurrency shape doesn't reflect "GitNexus's actual write pattern" (scope-precision point, not a defect)
**Severity: informational/P2 · Corroboration: single Claude lane (`ce-adversarial`) · Tag: [code-read]**

`conn-lock.ts` serializes every write through one connection; `bridge-db.ts`'s `withHandleLock` does the same per-handle for the group/cross-repo DB; `pool-adapter.ts`'s MCP pool is read-only by design. Net: GitNexus has **zero concurrent writers anywhere in its own code today**, by design. The test's 2-writer/3-reader shape validates the upstream engine fix for a hypothetical future multi-writer mode (issue #2338's own action item 3 language, "before considering multi-writer as a default," already scopes this correctly) — but the test docstring and its own commit message ("under a shape close to GitNexus's real concurrent-writer load") overstate present-day relevance. This is about how the PR/issue answers action item 3 precisely, not a code defect.

### GUARDRAILS.md/RUNBOOK.md's "stop the overlapping process" advice describes the wrong process model
**Severity: P3 · Corroboration: single Claude lane (`risk-architect`, F5, confidence 85) · Tag: [code-read]**

The operator advice frames the fix as cross-*process*, but the actual cause is concurrent *connections within one process*. The practical advice still works and the error is unreachable in production under `conn-lock.ts` anyway — minor precision gap only.

### GitHub-visible facts
**Tag: [reproduced]** (via `gh pr view`) — `facts-historian`: PR has **0 human reviews** recorded despite multiple AI-agent review rounds; issue **#2338 is NOT wired as a closing issue** (`closingIssuesReferences` empty) — action item 2 (large-repo benchmark) is explicitly deferred, so a full auto-close may be intentional, but this isn't stated anywhere in the PR.

### Hardware/environment sensitivity of the pre-fix deadlock reproduction rate
**Tag: [reproduced]** (single Claude lane, `ce-adversarial`, live 6-run execution) — 50% hang rate (3/6) against 0.17.1 in this sandbox vs. the PR's own reported 25% (1/4) locally. Both are consistent with "timing-dependent, not deterministic" (matching #605's own description), but the 2x shift purely from environment/scheduling is worth flagging as evidence the safety margin (and the new test's 60s timeout sizing, empirically set from local Linux runs only) could vary meaningfully by CI hardware.

---

## (e) Lower-priority items and doc-only inaccuracies (grouped)

- **Stemmer count: PR body says 27, actual is 28.** Found independently by `facts-historian`, `docs-dod`, `risk-architect`, and `test-ci-verifier` — **4 Claude lanes, consistent within one engine, not cross-engine**. Doc-only; the sweep test itself iterates the live exported `Set`, so test coverage is unaffected.
- **PR body stale by 2 commits** (`f999bdd5` JSDoc trim + GUARDRAILS/RUNBOOK notes; `2c00c278` withRetry swap) not described in the PR body. Found by `facts-historian`, `branch-hygiene`, `docs-dod` — 3 Claude lanes, consistent within one engine.
- **`SUPPORTED_FTS_STEMMERS` exported as a live mutable `Set`.** Found by `security-boundary`, `ce-maintainability`, `risk-architect` — 3 Claude lanes, P2/P3. Already a documented residual from an earlier review round per the PR body; only one current consumer (test file), read-only. Suggested fix (unchanged from prior round): `ReadonlySet<string>` or a frozen array + predicate helper.
- **GUARDRAILS.md's new "Known gap" bullet breaks the file's own strict 3-part Sign format** (`ce-standards`, F1, P3) — introduces a 4th bolded label type never used elsewhere; content is accurate, only a format-consistency nit.
- **Windows skip has no tracked follow-up issue number** (`docs-dod`, P3) — matches an established pattern elsewhere in the suite (`ce-standards` confirms byte-identical idiom used in 3 other test files), just missing an issue number.
- **PR title `fix(deps)` borderline vs. `chore(deps)`** (`docs-dod`, F4, P3/confidence 50) — the fixed bug is upstream in LadybugDB; no hard rule violated.
- Minor residual test-coverage notes from `ce-testing` (single lane, P3): `isDbBusyError`'s accepted-false behavior has no pinning unit test; `fts-stemmer-sweep.test.ts` only proves stemmers are *accepted*, not that they're *correctly applied* (matches a pre-existing convention, not a regression).
- Already-known, PR-body-disclosed residuals re-confirmed present and unchanged by this round (not new): timeout path doesn't cancel in-flight writer/reader loops before `finally` closes connections (P2, `risk-architect`/`ce-reliability`/`ce-testing` all re-confirm); reader loop has no bounded retry (P2/P3); `.shadow` poll is a best-effort proxy (already conceded in the PR's own text); `^0.18.0` caret range trade-off (documented, accepted convention). None of these are escalated by this round beyond what the PR already discloses.

---

## (f) CI status

All green. Per `facts-historian`: **13,629 passed, 0 failed, 56 skipped** (all documented expected-skips). All 26 status checks COMPLETED/SUCCESS (format, lint, typecheck, cross-platform tests on ubuntu/windows/macos, tree-sitter ABI ×3, packaged install smoke, benchmarks, CodeQL ×2, Dependency Review, Docker builds, Gitleaks, Vercel). No workflow files changed by this PR. Both new tests correctly wired into `vitest.config.ts`'s `lbug-db` project and `cross-platform-tests.ts`'s `LBUG_NATIVE` group.

---

## (g) A note on test-execution provenance in this review session

Mid-review, the coordinator discovered and fixed a **stale `node_modules` symlink**: this worktree's `gitnexus/node_modules` initially pointed at the main checkout's `node_modules`, which is still pinned to `@ladybugdb/core@0.17.1` (main hasn't merged this PR). This is a review-harness artifact, not a PR defect — GitHub Actions CI always does a clean `npm ci` off this PR's own lockfile and genuinely ran 0.18.0.

This affects how much weight to give in-session "I ran it and it passed/failed" claims:
- **Before the fix / unaware of it:** `test-ci-verifier`'s "passed in 2.90s locally" claim for the deadlock test does not mention the binary version at all — given the PR's own reported ~25% hang rate on 0.17.1, a single pass is the *more likely* outcome even on the buggy binary, so this claim cannot be read as 0.18.0 confirmation. `ce-correctness` self-discovered the stale-symlink gap during its own investigation and explicitly caveated its findings accordingly (did not claim to have validated 0.18.0 behavior).
- **After discovery, deliberately used as a before/after comparison:** `ce-testing` (3 runs, 3/3 correctly failed) and `ce-adversarial` (6 runs, 3 hangs/3 passes) both explicitly identified they were running against 0.17.1 and framed their results as *pre-fix mechanism confirmation* ("the assertion is a real, non-vacuous deadlock detector"), not as validation of the fix itself — a self-aware, correctly-scoped use of the wrong-but-known binary.
- **Genuine post-fix confirmation:** after re-pointing the symlink to the PR author's own correctly-provisioned `0.18.0` implementation worktree, the coordinator re-ran both new tests directly: `lbug-multiwriter-deadlock.test.ts` passed **4/4 consecutive runs** (~2.4s each), `fts-stemmer-sweep.test.ts` passed **28/28**. These are the only genuinely 0.18.0-verified local results produced during this review session, and they corroborate (do not merely repeat) the PR author's own originally-reported 9/9-clean claim.

---

This is an automated multi-engine digest (2 Claude review scaffolds + 3 Codex legs). Verify findings against the actual diff before acting on them.

gitnexus/test/integration/lbug-multiwriter-deadlock.test.ts
```diff
@@ -0,0 +1,227 @@
...
+  maxAttempts = 500,
+): Promise<void> {
+  await withRetry(
+    async () => {
+      const result = await conn.query(query);
+      await closeQueryResults(result);
+    },
+    {
+      maxAttempts,
+      baseDelayMs: 1,
```
magyargergo, line 100: **[P2 · STRONG cross-engine corroboration: Claude `ce-correctness` + Codex `codex-swarm`, independently · [code-read] verified by direct computation]**

`withRetry` is called here with `baseDelayMs: 1, capDelayMs: 3`. Tracing `computeBackoffMs` (`gitnexus-shared/src/integrations/retry.ts:52-65`) for `attempt=0`: `exponential = 1*2^0 = 1`, `upper = min(3,1) = 1`, `delay = floor(random()*1)`. Since `Math.random()` is always `< 1`, this is **deterministically 0** on the very first retry, every time — `withRetry` then skips `sleep()` entirely (`retry.ts:101`, `if (delayMs > 0) await sleep(delayMs)`).

This contradicts the surrounding JSDoc (unchanged by this refactor), which says the config was "validated across 5 consecutive local runs" specifically because a small jittered delay (never zero) was needed to avoid tripping a *different* native guard (`"Timeout waiting for active write transactions..."`). The old hand-rolled loop (`sleep(1 + Math.floor(Math.random()*3))`) always slept 1-3ms; the new one can skip sleeping entirely. All of the PR's cited "9/9"/"12 consecutive" validation runs predate this refactor commit (2c00c278) — there's no recorded re-validation count for the timing that actually ships.

**Suggested fix** (Codex `codex-swarm` Lane 4's recommendation — the only shape that actually works): have `isRetryable` return `{ retry: true, afterMs: 1 + Math.floor(Math.random()*3) }` to preserve the exact old 1-3ms jitter range. Bumping `baseDelayMs` alone does not fix this — no value eliminates 0 from `floor(random()*upper)`, it only lowers the probability (e.g. `baseDelayMs: 2` still yields 0 delay 50% of the time at attempt=0). Re-run the consecutive-pass validation after the fix and update the JSDoc/PR counts to match.

gitnexus/test/integration/lbug-multiwriter-deadlock.test.ts
```diff
@@ -0,0 +1,227 @@
...
+ * `Connection`s directly against `@ladybugdb/core`, so it proves the
+ * *native* engine no longer deadlocks — not merely that GitNexus's app-level
+ * serialization hides the problem. Production still routes every write
+ * through the single serialized connection (see `conn-lock.ts`); this test
+ * does not change that. It does reuse `lbug-config.ts`'s `createLbugDatabase`
+ * for the constructor call itself, so it stays in sync with any future
+ * signature change instead of hand-maintaining a second copy of the
+ * positional arg list.
+ *
+ * Empirical grounding (see plan docs/plans/2026-07-01-001-...):
```
magyargergo: **[P2 · single Claude lane (`ce-maintainability`) · [reproduced] — confirmed via `find`, path doesn't exist]**

This references `docs/plans/2026-07-01-001-...`, which doesn't exist anywhere in this repo (`docs/` is gitignored — this was a local planning artifact from the implementing session, never committed) and never will for anyone else. Commit `f999bdd5` (same PR) explicitly trims exactly this kind of task-narrative/session-referencing JSDoc from 5 production files, but this test file — added/touched across 3 earlier commits in this same PR — was never brought in line with that same policy. Lines 38-40 and 47-48 similarly say "flagged in the PR description" / "see the PR description for the full record" — pointers that rot the moment the PR merges and the description is no longer the living source of truth.

Suggested fix: replace the dangling path with either nothing (the surrounding prose already carries the empirical grounding inline) or a link to wherever this record should durably live (e.g. a code comment with the actual measured numbers, not a pointer to an ephemeral doc).

Issue #307, opened 2026-03-16
doesn't work with pnpx


wmertens:
```
$ pnpx gitnexus
Progress: resolved 79, reused 38, downloaded 32, added 0
Downloading kuzu@0.11.3: 6.63 MB/184.69 MB
Downloading @huggingface/transformers@3.8.1: 10.48 MB/10.48 MB, done
 WARN  7 deprecated subdependencies found: are-we-there-yet@3.0.1, boolean@3.2.0, gauge@4.0.4, glob@11.1.0, kuzu@0.11.3, npmlog@6.0.2, tar@6.2.1
Progress: resolved 294, reused 204, downloaded 67, added 0: 8.70 MB/20.2
Packages: +274
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++.2
Progress: resolved 294, reused 204, downloaded 67, added 0
Downloading onnxruntime-web@1.22.0-dev.20250409-89f8206ba4: 20.24 MB/20.Downloading onnxruntime-node@1.21.0: 80.02 MB/80.02 MB, done8.70 MB/20.2
Downloading kuzu@0.11.3: 184.69 MB/184.69 MB, done
Progress: resolved 296, reused 204, downloaded 72, added 274, done
.cache/pnpm/dlx/a24b89e5f3e4e882df8d1e8bde8fead2cc0d76293a09a3d153c55be77c7da9ab/19cf541ab5d-1a6521/node_modules/.pnpm/gitnexus@1.4.1_graphology-types@0.24.8_zod@4.3.6/node_modules/gitnexus: Running postinstall scrip.cache/pnpm/dlx/a24b89e5f3e4e882df8d1e8bde8fead2cc0d76293a09a3d153c55be77c7da9ab/19cf541ab5d-1a6521/node_modules/.pnpm/gitnexus@1.4.1_graphology-types@0.24.8_zod@4.3.6/node_modules/gitnexus: Running postinstall script, done in 115ms
node:internal/modules/esm/resolve:274
    throw new ERR_MODULE_NOT_FOUND(
          ^

[...]

Issue #165, opened 2026-03-04
Requesting support for cuda 13


nimishchaudhari:
Cannot install gitnexus via npm due to cuda

nimish@jetpack-ai:~$ nvidia-smi
Wed Mar  4 11:24:36 2026       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 590.48.01              Driver Version: 590.48.01      CUDA Version: 13.1     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 3090        On  |   00000000:01:00.0  On |                  N/A |
|  0%   35C    P8             17W /  230W |    7209MiB /  24576MiB |      3%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA GeForce RTX 3060        On  |   00000000:04:00.0 Off |                  N/A |
|  0%   39C    P8              9W /  180W |      14MiB /  12288MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    0   N/A  N/A           13372      G   /usr/bin/gnome-shell                    674MiB |
|    0   N/A  N/A           13756    C+G   /usr/bin/gnome-software                  34MiB |
|    0   N/A  N/A           13870      G   /usr/bin/Xwayland                         6MiB |
|    0   N/A  N/A           16528      G   /usr/bin/kitty                           69MiB |
|    0   N/A  N/A          712517    C+G   /usr/bin/nautilus                        89MiB |
|    0   N/A  N/A          939385    C+G   /usr/bin/gnome-text-editor               57MiB |
|    0   N/A  N/A         1849819      G   ...share/Steam/ubuntu12_32/steam          6MiB |
|    0   N/A  N/A         1850314      G   ./steamwebhelper                         66MiB |
[...]

PR #2341, opened 2026-07-01
fix(embeddings): use system-matched onnxruntime-node CUDA build so CUDA 13 hosts use the GPU

ACT900:
## What

On a CUDA 13 host, `gitnexus analyze --embeddings --embedding-device cuda` silently runs embeddings on **CPU** instead of the GPU. This teaches the embedder to pick the `onnxruntime-node` CUDA build that matches the host's CUDA runtime (12 or 13).

## Why

The embedding pipeline runs on `@huggingface/transformers`, which exact-pins `onnxruntime-node@1.24.3` (a **CUDA 12** build). gitnexus' own `onnxruntime-node: ^1.24.0` floats to the latest 1.x (a **CUDA 13** build). npm/pnpm can't dedupe an exact pin against a range, so an install carries two copies — gitnexus' top-level CUDA-13 build (unused) and transformers' nested CUDA-12 build (the one that actually loads).

The `overrides` block in `gitnexus/package.json` that would reconcile them is honoured only from a **root** manifest, so it is inert once gitnexus is installed as a dependency (`npm i -g`, `npx`, transitive) — the same transitive-override limitation already documented in `onnxruntime-common-resolver.ts` (#307).

Two coupled CUDA-12 assumptions then produce the bug:

1. `isCudaAvailable()` probed **only** `libcublasLt.so.12`, so on a CUDA-13-only host (`.so.13`, no `.so.12`) it returned `false` → `auto` picks CPU.
2. Even when forced, the nested CUDA-12 provider can't load `libcublasLt.so.12` → the CUDA execution provider fails → CPU fallback.

## What changed

- **New `onnxruntime-node-resolver.ts`** — mirrors the existing `onnxruntime-common-resolver.ts` `module.registerHooks` pattern. Before transformers is imported, if the system's cuBLASLt major (12/13) does not match the build transformers would load **but gitnexus' own copy does**, it redirects `onnxruntime-node` (and its version-matched `onnxruntime-common`) to that copy. Detection-based and conservative — a **no-op** when the default already matches, when there is no system CUDA, on non-Linux, when only one copy is present, or when neither copy matches (never makes it worse). Requires `module.registerHooks` (Node ≥ 22.15); a graceful no-op below that.
- **`embedder.ts`** — `isCudaAvailable()` now probes for `libcublasLt.so.12` **or** `.so.13`, matched against the CUDA major of the copy that will *actually* load (via `getEffectiveOnnxRuntimeNodeDir()`), replacing the CUDA-12-only probe. Net −45 lines.
- **`test/unit/onnxruntime-node-resolver.test.ts`** — 12 cases covering cuBLASLt-major detection (ldconfig / ldd / `LD_LIBRARY_PATH` scan, CUDA 12 vs 13, unresolved-lib ldd exit) and the graceful / no-op / idempotent installation, in the same style as the sibling resolver test.

## How to verify

```bash
# in gitnexus/
npx tsc --noEmit                                             # clean
npx vitest run test/unit/onnxruntime-node-resolver.test.ts  # 12 passed
```

On a CUDA-13 host, `gitnexus analyze --embeddings --embedding-device cuda` now loads the CUDA execution provider (GPU) instead of logging `Failed to load library … libcublasLt.so.12` and running on CPU. Verified end-to-end: `@huggingface/transformers@4.2.0` + `onnxruntime-node@1.27.0` produces 384-dim embeddings on the GPU with no `LD_LIBRARY_PATH` shim.

## Risk / rollback

Low and self-contained. The redirect is a no-op on CUDA-12 hosts, Windows (DirectML), macOS, CPU-only hosts, and Node < 22.15; any failure is swallowed and leaves resolution exactly as before. Rollback is reverting the change — no state, no migration.


gitnexus/src/core/embeddings/embedder.ts
```diff
@@ -183,6 +132,12 @@ export const initEmbedder = async (
       // Under pnpm-strict / `pnpm dlx`, transformers' phantom `onnxruntime-common`
       // import is unresolvable; register the fallback resolver first (#307).
       ensureOnnxRuntimeCommonResolvable();
+      // Registered AFTER the common fallback so this hook resolves FIRST (Node
+      // runs the most-recently-registered hook first): on CUDA-13 hosts it
+      // redirects onnxruntime-node (and its version-matched onnxruntime-common)
+      // to the CUDA-13 build before transformers imports them. No-op on matching
+      // layouts, non-CUDA, Windows/DirectML, and macOS.
+      ensureOnnxRuntimeNodeMatchesSystem();
```
magyargergo, line 115: **P1.** This hook is only installed here, not in the separate MCP query-time embedder at `gitnexus/src/mcp/core/embedder.ts:71-72`, which still calls only `ensureOnnxRuntimeCommonResolvable()` before importing `@huggingface/transformers`. Confirmed reachable via `local-backend.ts:2130`'s `semanticSearch()`. On a CUDA-13 host with `--embedding-device cuda` set for MCP search, this fix does not apply — the redirect is never installed there regardless of device config. Found independently by 2 of 3 Codex lanes (ce-code-review P1, adversarial "no-ship") + separately noted by Claude's correctness lane.

Fix: call `ensureOnnxRuntimeNodeMatchesSystem()` in the MCP embedder too, mirroring this call site; add a cross-entrypoint regression test.

`[code-read]`

gitnexus/src/core/embeddings/onnxruntime-node-resolver.ts
```diff
@@ -0,0 +1,210 @@
...
+ * Idempotently install the CUDA-build-matching redirect. Call once immediately
+ * before the dynamic `import('@huggingface/transformers')` on the local
+ * embedding path (after the runtime guard, alongside the onnxruntime-common
+ * fallback). No-op unless a strictly-better matching copy exists.
+ */
+export const ensureOnnxRuntimeNodeMatchesSystem = (): void => {
+  if (attempted) return;
+  attempted = true;
+  try {
+    if (typeof registerHooks !== 'function') return; // Node < 22.15: graceful no-op
```
magyargergo, line 287: **P1.** `decide()` (used by `isCudaAvailable()` in embedder.ts) is indifferent to whether `registerHooks` exists, but this line makes the actual hook install a no-op on Node 22.0-22.14 (still allowed by `package.json`'s `engines: ">=22.0.0"`). Result: on a CUDA-13 host in that Node range, `isCudaAvailable()` can return `true` (decide() says redirect matches) while this function silently installs nothing — transformers then loads the default CUDA-12 build and the code requests `device:'cuda'` against it, i.e. exactly the "uncatchable native crash" this probe exists to prevent (see the comment at `embedder.ts:39`).

Coordinator-verified via direct trace (single Codex lane found it; Claude's docs-dod lane flagged the same Node-version fact but under-called the severity).

Fix: raise the engines floor to `>=22.15`, or have `decide()` treat the redirect as unavailable when `registerHooks` isn't a function.

`[code-read]`

gitnexus/test/unit/onnxruntime-node-resolver.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        throw Object.assign(new Error('ldd failed'), {
+          stdout: 'libcublasLt.so.12 => not found',
+        });
+      },
+    });
+    expect(mod.ortCudaMajor('/pkg/onnxruntime-node')).toBe(12);
+  });
+});
+
+describe('ensureOnnxRuntimeNodeMatchesSystem', () => {
```
magyargergo, line 326: **P1 (testing).** None of the 4 tests in this `describe` block ever drives `decide()` into `redirect: true` — all force `platform:'darwin'` or `registerHooks: undefined`. The actual redirect logic (the `resolve()` closure at `onnxruntime-node-resolver.ts:192-198`) and `isCudaAvailable()`'s 3-function composition (`embedder.ts:51-57`) have zero test executions anywhere in the repo — confirmed by running the suite (12/12 pass without touching the redirect branch). Compare `onnxruntime-common-resolver.test.ts:71-131`, whose sibling suite *does* capture and assert its resolve closure. Broadly corroborated across nearly every lane in both engines.

Fix: inject `require`/`createRequire` as a seam, drive a real `redirect:true` case, and assert the closure's branches.

`[code-read]` + `[reproduced]`

gitnexus/src/core/embeddings/onnxruntime-node-resolver.ts
```diff
@@ -0,0 +1,210 @@
...
+import { join, dirname } from 'node:path';
+import { execFileSync } from 'node:child_process';
+import { logger } from '../logger.js';
+
+export type CudaMajor = 12 | 13;
+
+const require = createRequire(import.meta.url);
+
+/** Read a shared object's NEEDED entries, tolerating ldd's non-zero exit when a lib is unresolved. */
+const readSoNeeded = (soPath: string): string => {
```
magyargergo: **P2 (Strong — reproduced).** `readSoNeeded` returns `''` on ANY `execFileSync` failure with no `.stdout` (missing `ldd` binary, permission-denied `.so`, sandboxed exec) — indistinguishable from "no CUDA provider shipped." The pre-PR code (`hasOrtCudaProvider`) used only `existsSync`, never `ldd`, so this is a genuine regression for CUDA-12 hosts that worked before this PR. Empirically reproduced (chmod 000 + missing-binary repros) by the Claude adversarial lane; independently flagged by the correctness lane too.

Fix: log distinctly / retain an `existsSync`-based fallback when `ldd` itself fails to run.

`[reproduced]`

magyargergo (commented):
# PR #2341 Tri-Review Digest — fix(embeddings): use system-matched onnxruntime-node CUDA build so CUDA 13 hosts use the GPU

Head: `4a56d51b632f4d32bad14bc0a605001d79b82efc` (ACT900:fix/cuda13-onnxruntime-node-build-match) · Base: `main` · Merge-base: `316aaed928a03fa52be2814083040e64b9a85ae9`
Worktree reviewed: `/workspace/.worktrees/pr-2341`

## 1. What this PR does

`@huggingface/transformers` exact-pins a CUDA-12 `onnxruntime-node` build while gitnexus' own dependency floats to a CUDA-13 build; npm/pnpm can't dedupe an exact pin against a range, so both copies end up on disk and transformers always loads its own (CUDA-12) copy — silently forcing CPU on CUDA-13-only hosts even with `--embedding-device cuda`. The PR adds `gitnexus/src/core/embeddings/onnxruntime-node-resolver.ts` (new, 210 lines): a CUDA-major probe (`detectSystemCudaMajor`, `ortCudaMajor` — via `ldconfig -p` / `ldd` / path scans) plus a `module.registerHooks`-based ESM resolve hook (`ensureOnnxRuntimeNodeMatchesSystem`) that redirects `onnxruntime-node` (and its paired `onnxruntime-common`) to gitnexus' own CUDA-13 copy when it's a strictly better match than transformers' default. `gitnexus/src/core/embeddings/embedder.ts` is rewired to call it (-73/+28 net), and a new 163-line, 12-test unit-test file covers the pure detection primitives. 3 files changed total, all within `gitnexus/src/core/embeddings/`.

## 2. Engine / lane breakdown (all confirmed live)

| Engine | Methodology | Lanes | Status |
|---|---|---|---|
| Claude | GitNexus PR swarm | facts-historian, risk-architect, test-ci, security-boundary, docs-dod, branch-hygiene | 6/6 live, substantive content |
| Claude | ce-code-review personas | correctness, testing, maintainability, project-standards, adversarial, reliability | 6/6 live, substantive content |
| Codex (gpt-5.5/xhigh) | GitNexus PR swarm, Solo mode | `codex-swarm-final.md` | Live — real per-lane findings (not empty/error) |
| Codex (gpt-5.5/xhigh) | ce-code-review, report-only | `codex-ce.md` | Live — full JSON findings block at tail (2 scored findings + testing_gaps + triage_groups), confirmed non-empty |
| Codex (gpt-5.5/xhigh) | native adversarial red-team | `codex-adversarial.md` | Live — structured finding block at tail after tool-call log, confirmed non-empty |

Two of the three Codex legs (ce-code-review and the native adversarial-review) independently discovered the same fact — `gitnexus/src/mcp/core/embedder.ts` (a second, unmodified local-embedding entrypoint) never gets this PR's fix — via two different review contracts; the third leg (the GitNexus PR swarm) did not surface it. Claude's own ce-correctness lane separately noted the identical underlying fact (see the P1 finding below for how it characterized it).

## 3. Headline findings (ranked by severity, then corroboration)

### P1 — MCP query-time embedder never installs the CUDA-13 redirect
**Corroboration: Strong** (Codex ce-code-review P1 + Codex adversarial "high"/no-ship + Claude ce-correctness lane independently noted the identical underlying fact, though it declined to file it as a defect, framing it as a possible intentional scope boundary)
**File: `gitnexus/src/mcp/core/embedder.ts:71-72`** (NOT touched by this diff)
**Evidence tag: `[code-read]`** — verified directly:
```
71:      ensureOnnxRuntimeCommonResolvable();
72:      const { pipeline, env } = await import('@huggingface/transformers');
```
No call to `ensureOnnxRuntimeNodeMatchesSystem()` exists anywhere in this file — this holds regardless of `embeddingConfig.device`, so the redirect is never installed for MCP embedding no matter how the user configures the device. Confirmed this is a real, reachable entrypoint: `gitnexus/src/mcp/local/local-backend.ts:2130` — `const { embedQuery, getEmbeddingDims } = await import('../core/embedder.js');` inside `semanticSearch()`, the MCP query-time embedding path. Separate, additional gap also verified: `devicesToTry` (mcp/core/embedder.ts:84-87) only attempts the GPU device when `embeddingConfig.device` is literally `'dml'`/`'cuda'` — never `'auto'` — meaning MCP also lacks the core embedder's `isCudaAvailable()`-driven auto-detection. A user who explicitly sets `--embedding-device cuda` for MCP search DOES attempt the CUDA device (unlike an `'auto'` user, who never tries CUDA on this path at all) — but that attempt still resolves the unredirected, mismatched CUDA-12 build on a CUDA-13-only host, because the hook that would fix that was never called.
**Trigger:** CUDA-13 host running the GitNexus MCP server for semantic search with `device: 'cuda'` explicitly configured.
**Suggested fix:** import and call `ensureOnnxRuntimeNodeMatchesSystem()` in `mcp/core/embedder.ts` immediately after `ensureOnnxRuntimeCommonResolvable()`, mirroring the core embedder; add a regression test asserting both local-embedding entrypoints register the hook before importing transformers.

### P1 — On Node 22.0–22.14, the probe can report "CUDA available" while the redirect hook silently never installs — reproducing the exact crash the probe exists to prevent
**Corroboration: coordinator-verified single Codex lane** (`codex-ce.md` finding #2, "CUDA probe trusts uninstalled redirect"), with a partial, weaker cross-engine signal: Claude's `docs-dod` lane independently flagged the same Node-version boundary (F1: "Node 22.0-22.14 silent non-fix not disclosed") but characterized the consequence as a benign "stays on CPU fallback" — I traced the code and confirmed **Codex's characterization is the one that actually holds**, not Claude's. Per synthesis rule 2, a verified P1 with a concrete trigger is postable regardless of lane count.
**File: `gitnexus/src/core/embeddings/onnxruntime-node-resolver.ts:176`**, combined with `embedder.ts:51-57` and `:124`/`:140`
**Evidence tag: `[code-read]`** — traced the exact call sequence in the worktree:
- `embedder.ts:124`: `const gpuDevice = isCudaAvailable() ? 'cuda' : 'cpu';` runs **synchronously**, before the async IIFE that calls `ensureOnnxRuntimeNodeMatchesSystem()` at `embedder.ts:140` even starts.
- `isCudaAvailable()` (`embedder.ts:51-57`) calls `getEffectiveOnnxRuntimeNodeDir()` → `decide()`. `decide()` (`onnxruntime-node-resolver.ts:136-153`) is **entirely indifferent to whether `registerHooks` exists** — it only uses `require.resolve`. On a CUDA-13-only host with gitnexus' own top-level CUDA-13 build present (the PR's target scenario), `decide()` legitimately returns `{redirect: true, effectiveDir: ourDir}`, so `isCudaAvailable()` returns `true` and `gpuDevice = 'cuda'`.
- `ensureOnnxRuntimeNodeMatchesSystem()` then runs; on Node 22.0-22.14 it hits `if (typeof registerHooks !== 'function') return;` (line 176) and returns having installed **nothing**.
- `import('@huggingface/transformers')` then resolves the **default** (unredirected, CUDA-12) `onnxruntime-node`. The pipeline is created with `device: 'cuda'` against a CUDA-12 binary on a host that (per this exact scenario) lacks CUDA-12 libs — the code's own comment (`embedder.ts:39`, "ONNX Runtime's native layer crashes (uncatchable) if we request CUDA without the required shared libraries") describes precisely this failure.
- Confirmed `gitnexus/package.json:119-120` keeps `engines: { "node": ">=22.0.0" }` — Node 22.0-22.14 is an allowed, undetected runtime for this desync.
**Trigger:** Node 22.0-22.14 + CUDA-13-only host + gitnexus' own top-level onnxruntime-node build present + no CUDA-12 compat libs on the system.
**Suggested fix:** either raise `engines.node` to `>=22.15`, or make `decide()`/`getEffectiveOnnxRuntimeNodeDir()` return the redirected dir only when `registerHooks` is confirmed available, falling back to transformers' actual default dir otherwise, so the probe and the loader can never disagree.

### P1 (testing) — The PR's actual shipped behavior (the `registerHooks` redirect closure, and `isCudaAvailable()`'s composition) has zero executed test coverage anywhere in the repo
**Corroboration: Strong, broad cross-engine** — Claude ce-testing (its own P1 x2: Finding 1 "redirect:true path structurally unreachable," Finding 2 "isCudaAvailable() invoked by zero tests anywhere"), Claude swarm-risk-architect (P2), Claude ce-correctness F4 (P2), Claude ce-maintainability (testing gap), Codex swarm-final Lane 3 (P2), Codex-ce (`testing_gaps` + both filed findings' evidence trail), Codex-adversarial (testing gaps).
**File: `gitnexus/test/unit/onnxruntime-node-resolver.test.ts:134-163`** (the `ensureOnnxRuntimeNodeMatchesSystem` describe block); source at `onnxruntime-node-resolver.ts:172-210` (esp. the `resolve()` closure, lines 192-198) and `embedder.ts:51-57`.
**Evidence tag: `[code-read]` + `[reproduced]`** — read the test file directly: all 4 tests either force `platform: 'darwin'` (zeroing `systemMajor` before `decide()` ever inspects directories) or set `registerHooks: undefined`; none mocks `require.resolve`/`createRequire` to drive `decide()` into `redirect: true`. Ran `npx vitest run test/unit/onnxruntime-node-resolver.test.ts` myself → 12/12 pass, confirming the suite is green while never exercising the redirect branch. Also confirmed by direct comparison: the sibling `onnxruntime-common-resolver.test.ts:71-131` **does** capture and directly assert on its resolve closure via a `captureResolve()` helper — this PR's test file has no equivalent.
**Suggested fix:** make `resolveDefaultOrtNodeDir`/`resolveOurOrtNodeDir` injectable (seam for `require`/`createRequire`), add a test driving `ortCudaMajor(defaultDir) !== systemMajor && ortCudaMajor(ourDir) === systemMajor`, and assert the captured `resolve()` closure's three branches (`onnxruntime-node` redirect, `onnxruntime-common` redirect, passthrough) — mirroring the sibling file's existing pattern.

### P2 — New hard dependency on `ldd` succeeding regresses CUDA detection on CUDA-12 hosts that worked fine under the old `existsSync`-only check
**Corroboration: Strong via reproduction** — two independent Claude lanes converge (ce-correctness F3, confidence 35; ce-adversarial Finding 1, confidence 75) AND ce-adversarial **empirically reproduced both failure shapes live in its own sandbox** (`chmod 000` on the `.so` → permission-denied throw with no stdout; nonexistent `ldd` binary → `ENOENT` with no stdout) — this satisfies the "reproduced end-to-end" bar for Strong even though both lanes are the same engine.
**File: `onnxruntime-node-resolver.ts:58-65`** (`readSoNeeded`), `:68-82` (`ortCudaMajor`), consumed at `embedder.ts:51-56`.
**Evidence tag: `[reproduced]`**. Confirmed via `git show <merge-base>:gitnexus/src/core/embeddings/embedder.ts` that the pre-PR `hasOrtCudaProvider()` used only `existsSync` — it never shelled out to `ldd` at all.
**Trigger:** `ldd` absent from `PATH` (minimal/hardened container images), or the provider `.so` lacks read permission for the running user (root-installed `node_modules` run as an unprivileged user — a real, documented Docker misconfiguration pattern), or exec blocked by sandbox/seccomp/AppArmor. In all three, `readSoNeeded` returns `''` (the catch block's `.stdout` check fails), both CUDA-major regexes miss, and `ortCudaMajor` returns `null` — indistinguishable from "no CUDA provider shipped at all," so `isCudaAvailable()` unconditionally returns `false` even though the host's CUDA-12 libs are genuinely present and worked before this PR.
**Suggested fix:** distinguish "detection failed" from "definitely no provider" (e.g. log at `warn` when `readSoNeeded` throws with no `stdout`), and/or retain the old `existsSync`-based check as a secondary corroboration before giving up entirely.

## 4. Checked and refuted (validated code, not defects — a real signal of quality)

- **Hook LIFO ordering is correct, not a bug.** Node's `module.registerHooks` runs the most-recently-registered hook first; the PR's own comment relies on this and it's accurate. Confirmed against official Node docs and live repro by risk-architect, ce-adversarial, and codex-adversarial independently.
- **`napi-v6` hardcoded path** in `ortCudaMajor()` — confirmed pre-existing (carried over from the removed `hasOrtCudaProvider`), not a regression (facts-historian, risk-architect, maintainability, via `git show`/`grep`).
- **Security surface fully clean** (security-boundary lane, all P4): `execFileSync` calls are array-form with no `shell: true` (injection-safe by construction, no user input reaches the args); `CUDA_PATH`/`LD_LIBRARY_PATH` path traversal via `existsSync` is read-only (worst case a mis-detected CUDA major, never write/exec); the ESM hook only ever redirects to installed-package-derived paths, never user input; subprocess-output parsing has no ReDoS (fixed-literal regexes/`.includes()`, bounded 3s/5s timeouts). Zero merge-blocking security findings.
- **`ldd` output locale-dependence** — refuted (ce-adversarial): `DT_NEEDED` sonames are verbatim ASCII from the ELF table, not locale-sensitive prose.
- **Transitive cuBLASLt pull-in causing a wrong-major read** — refuted (ce-adversarial): verified against the repo's actual installed `onnxruntime-node@1.27.0` CUDA provider binary; cuBLASLt is a direct `NEEDED` entry, not indirect.
- **npm global-install layout collapsing `ourDir`/`defaultDir` onto the same copy** (which would silently disable the fix for the PR's own target scenario) — refuted (ce-adversarial): built a concrete two-copy fixture and confirmed both resolve to distinct, correct directories as designed.
- **`attempted`-flag TOCTOU / concurrent-init race** — refuted (ce-reliability): the entire call chain is synchronous with no `await`, so two "concurrent" `initEmbedder()` calls cannot interleave inside it regardless of flag-set ordering.
- **Partial-hook-install / degraded state on throw** — refuted (ce-reliability): `nodeUrl` resolution and `registerHooks(...)` are both inside the same outer `try`, so a throw before `registerHooks()` leaves nothing installed — clean degradation, matches the documented contract.
- **Project-standards: zero violations** — no `any`/`as any`, no absolute/machine-specific paths, naming/file conventions consistent with the sibling resolver, PR title is conventional-commit compliant, diff scope is minimal and causally connected (branch-hygiene: single clean commit, zero conflict risk, zero unrelated churn).

## 5. Lower-priority items, test gaps, and disclosure gaps (single-lane, not reproduced — per rule 2 these need ≥2-lane/reproduction to be inline-postable, so listed here instead)

**Correctness edge cases** (all `[code-read]`, verified directly, all single-lane P2/P3):
- `getEffectiveOnnxRuntimeNodeDir()` can return `null` even when `ourDir` would resolve and match, because the `ourDir` fallback lookup is gated behind `defaultDir` being truthy (`onnxruntime-node-resolver.ts:142`) — affects the narrower case where transformers' own nested copy fails to resolve at all (ce-correctness F1, confidence 65).
- `CUDA_PATH`/`LD_LIBRARY_PATH` scan only prefers CUDA-13-over-12 within a single `dir`/`sub` pair, not globally across the whole search path (`:98-105`) (ce-correctness F2, confidence 45).
- Stale `ldconfig` cache (library physically removed without re-running `ldconfig`) could make `detectSystemCudaMajor` report a phantom CUDA-13, driving `redirect:true` + `isCudaAvailable()===true` against a genuinely-missing runtime lib → the crash scenario. Not independently reproducible without real CUDA hardware (ce-adversarial Finding 2, confidence 55).
- `npm link`/symlinked dev-checkout layouts resolve `resolveOurOrtNodeDir`/`resolveDefaultOrtNodeDir` against the resolver module's own realpath, not the consuming app's tree — mechanism was empirically reproduced, but blast radius is narrow (dev-only symlink workflows), not representative of gitnexus' normal install paths (ce-adversarial Finding 3, confidence 70).
- `onnxruntime-common-resolver.ts`'s independently-computed fallback `redirectUrl` could pair the wrong `onnxruntime-common` with an already-redirected `onnxruntime-node` in a compound failure scenario; no concrete trigger constructed (ce-adversarial Finding 4, confidence 45).
- Duplicated CJS resolve-walk (the "resolve via transformers' main entry, not `package.json`" gotcha) between `onnxruntime-node-resolver.ts:110-117` and `onnxruntime-common-resolver.ts:74-84`, undocumented in the new file — a future "simplification" could silently reintroduce #307 (ce-maintainability F1, confidence 75).

**Reliability / observability** (all Claude ce-reliability, single-lane, `[code-read]`/reasoned):
- Up to ~13s of synchronous `execFileSync` calls (1x `ldconfig` + up to 2x `ldd`) can block the entire Node event loop on the first `initEmbedder()` call in a long-lived server process (F1, confidence 70).
- `execFileSync`'s `timeout` is a kill-signal deadline, not a hard wall-clock bound — a child stuck in uninterruptible I/O (e.g. an NFS-stalled `.so` read) isn't guaranteed to be bounded by it; theoretical, not reproducible in this sandbox (F2, confidence 55).
- A transient first-probe failure is cached for the life of the process (`cached` is set unconditionally, including on caught-exception "null" results) — permanently locks a CUDA-capable host onto CPU until restart (F3, confidence 80, verified via code read).
- Redirect success/failure is logged only at `logger.debug` (filtered by the default `info` level, confirmed in `logger.ts`), and `gitnexus/src/cli/doctor.ts` never surfaces the resolved CUDA major or redirect state (only the configured device and an ONNX version fingerprint) — no diagnostic path for "why is my CUDA-13 host still on CPU" (F4, confidence 85, verified via code read).
- Redundant `ldconfig`/`ldd` subprocess spawns: `isCudaAvailable()` re-invokes `ortCudaMajor`/`detectSystemCudaMajor` even though `decide()` already computed both, since `Decision` doesn't cache the effective dir's CUDA major (swarm-risk-architect Finding 2, P3).

**Test-hygiene / disclosure gaps** (docs-dod + test-ci, mostly P3-P4):
- PR body doesn't disclose the Node 22.0-22.14 boundary (docs-dod F1) — now known (via P1 finding #2 above) to be a more serious crash-risk desync than the "silent non-fix" the PR body's absence implies.
- No linked GitHub issue — #165 ("Requesting support for cuda 13") is the closest match but is a distinct, earlier-layer bug (postinstall version rejection) and isn't formally closed by this PR (facts-historian, docs-dod F2).
- PR body doesn't explicitly name the CI-unverifiable GPU-hardware validation gap (docs-dod F3).
- The new platform-mutating test file is absent from `scripts/cross-platform-tests.ts`'s `PLATFORM_LOGIC` list, which `TESTING.md:96` says platform-sensitive tests should be added to. **Verified directly**: zero `"onnxruntime"` hits in `cross-platform-tests.ts`. Note a real disagreement between lanes here: Claude's own swarm test-ci lane asserted the opposite ("correctly excludes this Linux-only test"), which does not hold up — nothing "excludes" it, it's simply absent; Codex's swarm lane caught this correctly per the written convention. Operational risk is low (the code's non-Linux paths are simple `process.platform !== 'linux'` guards, and the sibling `onnxruntime-common-resolver.test.ts` has the identical, pre-existing gap).
- Weak/vacuous assertions in 3 of the 4 `ensureOnnxRuntimeNodeMatchesSystem` tests — assert only "doesn't throw," not concrete return values; the test literally named "is best-effort and idempotent" never asserts a call count on its own spy and (per the P1 testing finding above) can't even reach the guarded path under its own `darwin` mock (ce-testing Finding 6, confidence 82, verified directly against the test file).
- No test for `ortCudaMajor`'s "garbled output matches neither regex" case; both detectors use unanchored substring/regex matches with no boundary (e.g. a hypothetical `libcublasLt.so.130` would false-match `.so.13`) — low real-world likelihood today (ce-testing Finding 5).
- Minor/cosmetic: `CUDA_PATH` fallback branch untested vs. `LD_LIBRARY_PATH` (facts-historian); `process.env = {...}` reassignment inconsistent with the file's own `Object.defineProperty` platform-stub pattern (swarm test-ci, P4); Windows (`win32`) not separately exercised from `darwin` despite sharing an identical guard (ce-project-standards, P4).

## 6. Postable inline comments

| # | File | Line | Body |
|---|------|------|------|
| 1 | `gitnexus/src/core/embeddings/embedder.ts` | 140 | **P1.** This hook is only installed here (`gitnexus/src/core/embeddings/embedder.ts`), not in the separate MCP query-time embedder at `gitnexus/src/mcp/core/embedder.ts:71-72`, which still calls only `ensureOnnxRuntimeCommonResolvable()` before importing `@huggingface/transformers`. Confirmed reachable via `local-backend.ts:2130`'s `semanticSearch()`. On a CUDA-13 host with `--embedding-device cuda` set for MCP search, this fix does not apply. Found independently by 2 Codex lanes (ce-code-review P1, adversarial "no-ship") + noted by Claude's correctness lane. Fix: call `ensureOnnxRuntimeNodeMatchesSystem()` in the MCP embedder too, mirroring this call site; add a cross-entrypoint regression test. `[code-read]` |
| 2 | `gitnexus/src/core/embeddings/onnxruntime-node-resolver.ts` | 176 | **P1.** `decide()` (used by `isCudaAvailable()` in embedder.ts) is indifferent to whether `registerHooks` exists, but this line makes the actual hook install a no-op on Node 22.0-22.14 (still allowed by `package.json`'s `engines: ">=22.0.0"`). Result: on a CUDA-13 host in that Node range, `isCudaAvailable()` can return `true` (decide() says redirect matches) while this function silently installs nothing — transformers then loads the default CUDA-12 build and the code requests `device:'cuda'` against it, i.e. exactly the "uncatchable native crash" this probe exists to prevent. Coordinator-verified via direct trace (single Codex lane found it; Claude's docs-dod lane flagged the same Node-version fact but under-called the severity). Fix: raise the engines floor to `>=22.15`, or have `decide()` treat the redirect as unavailable when `registerHooks` isn't a function. `[code-read]` |
| 3 | `gitnexus/test/unit/onnxruntime-node-resolver.test.ts` | 134 | **P1 (testing).** None of the 4 tests in this `describe` block ever drives `decide()` into `redirect: true` — all force `platform:'darwin'` or `registerHooks: undefined`. The actual redirect logic (the `resolve()` closure at `onnxruntime-node-resolver.ts:192-198`) and `isCudaAvailable()`'s 3-function composition (`embedder.ts:51-57`) have zero test executions anywhere in the repo — confirmed by running the suite (12/12 pass without touching the redirect branch) and grepping every test file that imports the core embedder. Compare `onnxruntime-common-resolver.test.ts:71-131`, whose sibling suite *does* capture and assert its resolve closure. Broadly corroborated (5 Claude findings + 3 Codex references to the same gap). Fix: inject `require`/`createRequire` as a seam, drive a real `redirect:true` case, and assert the closure's branches. `[code-read]`+`[reproduced]` |
| 4 | `gitnexus/src/core/embeddings/onnxruntime-node-resolver.ts` | 58 | **P2 (Strong — reproduced).** `readSoNeeded` returns `''` on ANY `execFileSync` failure with no `.stdout` (missing `ldd` binary, permission-denied `.so`, sandboxed exec) — indistinguishable from "no CUDA provider shipped." The pre-PR code (`hasOrtCudaProvider`) used only `existsSync`, never `ldd`, so this is a genuine regression for CUDA-12 hosts that worked before. Empirically reproduced (chmod 000 + missing-binary repros) by the Claude adversarial lane; independently flagged by the correctness lane too. Fix: log distinctly / retain an `existsSync`-based fallback when `ldd` itself fails to run. `[reproduced]` |

## 7. Bottom line

- **Verdict:** `not production-ready` (2 blocking correctness/scope gaps + 1 blocking test-coverage gap; core mechanism otherwise sound)
- **Branch hygiene:** `clean feature/fix PR` — single commit, causally scoped to `gitnexus/src/core/embeddings/`, no unrelated churn. Note: `origin/main` has advanced one commit since this PR branched (`400cc6a4`, an unrelated CJK-bigram FTS change, disjoint files, zero conflict risk) — a rebase will be needed before landing, and CI should re-run on the rebased head.
- **Merge state:** `mergeable` per GitHub (`mergeStateStatus: BEHIND`, `mergeable: MERGEABLE`, zero conflicts) — but **CI visibility is incomplete**: `gh pr checks` shows only the Vercel deploy-auth gate (fails on every fork PR, not a code signal) and the conventional-label/PR-title checks; the actual unit-test/typecheck CI run's pass/fail status was not observable from this review and should be confirmed separately before merge.

## 8. Overall verdict

**Not safe to merge as-is; needs-attention, not no-ship.** The core mechanism (detection primitives, hook composition, LIFO ordering, security posture, scope/hygiene) is sound and was genuinely stress-tested — several plausible attack angles were tried and refuted, which is real signal. But two P1s should block: (1) the fix's own title claim ("CUDA 13 hosts use the GPU") is incomplete — the MCP query-time embedding path never gets the redirect, found independently by two of the three Codex legs plus separately noted by Claude's own correctness lane; and (2) on Node versions the project still declares supported (22.0-22.14), the probe and the actual loader can disagree in exactly the way the probe was built to prevent, risking the "uncatchable native crash" the code's own comments warn about. The third P1 (zero executed test coverage of the PR's actual redirect behavior) means neither of the above would have been caught by CI even if the author's manual verification on a real CUDA-13 host was accurate. Recommended before merge: wire the MCP embedder into the same resolver call, close the Node-version probe/loader agreement gap (or raise the engines floor), and add at least one test that drives `redirect: true` end-to-end. Good follow-ups (non-blocking): the `ldd`-hard-dependency regression on previously-working CUDA-12 hosts, the reliability/observability gaps (debug-only logging, no `doctor` visibility, up-to-13s synchronous blocking), and the narrower correctness edge cases (defaultDir-null gating, LD_LIBRARY_PATH scan order, stale-ldconfig-cache, npm-link resolution, common/node pairing under compound failure).

ACT900:
Thank you for the exceptionally thorough review, @magyargergo — the tri-review caught real issues I'd missed (the MCP query-time entrypoint, the Node 22.0–22.14 `registerHooks` desync, the redirect tests that never actually drove the redirect branch, and the `ldd` regression on minimal containers), and I really appreciate you taking the time to implement the fixes directly. The effective-major caching, the `doctor` surfacing, and the cross-platform/OS-agnostic test hardening are all clear improvements. Happy to help with anything else the PR needs.

magyargergo:
Thank you @ACT900 for you contribution! It's a real game changer. 🔥

PR #2353, opened 2026-07-02
fix: resolve Java cast-wrapped and this.method() call edges

xi3w3n:
Two fixes for missing call edges in Java method resolution:

1. compound-receiver.ts — cast expression handling:
   - Strip (Type) cast wrappers from receiver text, tracking the outermost meaningful cast type
   - Resolve directly to the cast type class (not the field's declared type), since the cast narrows the receiver type
   - Add this.field chain walker for field-access receivers
   - Replace text → workingText throughout the function body

2. scope-resolver.ts:
   - Enable resolveThisViaEnclosingClass: true for Java (activates Case 0.5 in receiver-bound-calls.ts)

Verified on a large-scale Java codebase with no regressions.

  ## Summary

  Fix two root causes of missing call graph edges in Java: `this.method()` calls and cast-wrapped `(Type)this.field.method()` receivers.

  ## Motivation / context

  Java code that calls `((SubInterface) this.baseField).method()` produces a `cast_expression` node as the method call receiver. The compound receiver resolver couldn't parse cast syntax — the `(Type)` prefix broke chain splitting and
  head resolution. Additionally, bare `this.method()` calls weren't resolved for Java because `resolveThisViaEnclosingClass` was not enabled.

  ## Areas touched

  ## Scope & constraints

  **In scope**

  - Java cast-stripping + cast type tracking in `resolveCompoundReceiverClass`
  - `this.field[.subfield]` chain walker via enclosing class `typeBindings`
  - Enable `resolveThisViaEnclosingClass` for Java scope resolver

  - Tree-sitter query-level cast unwrapping
  - Other C-family languages (cast-stripping is safe for non-Java — passes through unchanged when no cast matches)

  ## Implementation notes

  Cast stripping peels `(Type)` layers left-to-right, tracking only the outermost meaningful type (skipping intermediate casts like `(Object)` introduced by decompilers). When a cast type is found, the resolver returns it directly — the
  cast IS the receiver's effective type. The `this.field` fallback walks the enclosing class's `typeBindings` only when no cast type was identified.

  ## Testing & verification

  - [ ] `cd gitnexus && npm run test:integration`

  Verified on a large-scale Java codebase — no edge regressions, no index corruption.

  ## Risk & rollout

  - No breaking changes. `workingText === text` when no cast is present.
  - Cast-stripping is gated on balanced-paren scanning; non-cast expressions pass through unchanged.
  - Re-indexing recommended for Java repos to pick up newly resolvable edges.

  ## Checklist

  - [ ] If `AGENTS.md` / overlays changed: headers and changelog updated

### DIFF
--- a/gitnexus/src/core/ingestion/languages/java/scope-resolver.ts
+++ b/gitnexus/src/core/ingestion/languages/java/scope-resolver.ts
@@ -52,6 +52,7 @@ const javaScopeResolver: ScopeResolver = {
   populateOwners: (parsed: ParsedFile) => populateClassOwnedMembers(parsed),
 
   isSuperReceiver: (text) => text.trim() === 'super',
+  resolveThisViaEnclosingClass: true,
 
   fieldFallbackOnMethodLookup: false,
   propagatesReturnTypesAcrossImports: true,
--- a/gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
+++ b/gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
@@ -25,6 +25,7 @@ import type { ScopeResolutionIndexes } from '../../model/scope-resolution-indexe
 import type { WorkspaceResolutionIndex } from '../workspace-index.js';
 import {
   findClassBindingInScope,
+  findEnclosingClassDef,
   findExportedDefByName,
   findReceiverTypeBinding,
 } from '../scope/walkers.js';
@@ -83,12 +84,107 @@ export function resolveCompoundReceiverClass(
   if (text.length === 0) return undefined;
   const fieldFallback = options.fieldFallback ?? true;
 
+  // ── Pre-processing: strip Java-style cast expressions ───────────
+  // Java cast-wrapped receivers like ((Type)((Object)this.field)).method()
+  // produce parenthesized_expression receiver text. Peel outer (Type)
+  // layers so the resolver sees the actual receiver (e.g. this.field).
+  // Track the outermost meaningful cast type — the cast narrows the
+  // receiver's declared type, so we resolve to the CAST type, not the
+  // underlying expression's type.
+  let castType: string | null = null;
+  let workingText = text;
+  while (true) {
+    if (!workingText.startsWith('(')) break;
+    let d = 1;
+    let closeIdx = -1;
+    for (let i = 1; i < workingText.length; i++) {
+      if (workingText[i] === '(') d++;
+      else if (workingText[i] === ')') {
+        d--;
+        if (d === 0) { closeIdx = i; break; }
+      }
+    }
+    if (closeIdx === -1) break;
+    // Extract the type from inside this cast's parens — only capture
+    // the FIRST (outermost) non-parenthesized simple type name as the
+    // effective cast target. CFR patterns like ((Target)((Object)expr))
+    // have a meaningful outer cast and a noise inner cast; we want the
+    // meaningful outer one.
+    const insideParens = workingText.slice(1, closeIdx).trim();
+    if (castType === null && !insideParens.startsWith('(') && /^[a-zA-Z_]\w*$/.test(insideParens)) {
+      castType = insideParens;
+    }
+    const remainder = workingText.slice(closeIdx + 1).trim();
+    // CFR decompiler patterns: (Type)expr, ((Object)expr), etc.
+    if (
+      remainder.length === 0 ||
+      remainder.startsWith('(') ||
+      remainder.startsWith('this') ||
+      /^[a-zA-Z_]/.test(remainder)
+    ) {
+      workingText = remainder.length === 0
+        ? workingText.slice(1, closeIdx).trim()
+        : remainder;
+      continue;
+    }
+    break;
+  }
+
+  // If we found a cast type, resolve it directly — the cast tells us
+  // the exact receiver type for method resolution.
+  if (castType !== null) {
+    const cls = findClassBindingInScope(inScope, castType, scopes);
+    if (cls !== undefined) return cls;
+  }
+
+  // Handle this.field[.subfield] — walk enclosing class's field typeBindings
+  if (workingText.startsWith('this.')) {
+    const enclosingClass = findEnclosingClassDef(inScope, scopes);
+    if (enclosingClass !== undefined) {
+      const chainRest = workingText.slice(5); // after 'this.'
+      const parts = chainRest.split('.');
+      let curClass: SymbolDefinition | undefined = enclosingClass;
+      for (let i = 0; i < parts.length && curClass !== undefined; i++) {
+        const seg = parts[i];
+        if (seg === undefined) break;
+        const cleanName = seg.endsWith(')') ? seg.slice(0, seg.indexOf('(')) : seg;
+        if (cleanName.length === 0) continue;
+        const cs = classScopeByDefId.get(curClass.nodeId);
+        if (cs === undefined) { curClass = undefined; break; }
+        let memberType = cs.typeBindings.get(cleanName);
+        if (memberType === undefined && options.hoistTypeBindingsToModule === true) {
+          let curId: ScopeId | null = cs.parent;
+          while (curId !== null) {
+            const curScope = scopes.scopeTree.getScope(curId);
+            if (curScope === undefined) break;
+            const cand = curScope.typeBindings.get(cleanName);
+            if (cand !== undefined) { memberType = cand; break; }
+            curId = curScope.parent;
+          }
+        }
+        if (memberType === undefined && fieldFallback) {
+          for (const [, mb] of cs.typeBindings) {
+            const fb = findClassBindingInScope(mb.declaredAtScope, mb.rawName, scopes);
+            if (fb === undefined) continue;
+            const fcs = classScopeByDefId.get(fb.nodeId);
+            const found = fcs?.typeBindings.get(cleanName);
+            if (found !== undefined) { memberType = found; break; }
+          }
+        }
+        if (memberType === undefined) { curClass = undefined; break; }
+        curClass = findClassBindingInScope(memberType.declaredAtScope, memberType.rawName, scopes);
+      }
+      if (curClass !== undefined) return curClass;
+    }
+  }
+  // ── End pre-processing ─────────────────────────────────────────
+
   // Bare identifier — resolve via typeBinding first, then fall back to
   // a direct class-name lookup. The class-name fallback handles
   // "static receiver" shapes like `UserService.findUser()` where
   // `UserService` isn't a variable but a class imported into scope.
-  if (!text.includes('.') && !text.includes('(')) {
-    const mapTuple = parseMapTupleSentinel(text);
+  if (!workingText.includes('.') && !workingText.includes('(')) {
+    const mapTuple = parseMapTupleSentinel(workingText);
     if (mapTuple !== null) {
       const rhsTb = findReceiverTypeBinding(inScope, mapTuple.rhs, scopes);
       if (rhsTb === undefined) return undefined;
@@ -97,7 +193,7 @@ export function resolveCompoundReceiverClass(
       return findClassBindingInScope(rhsTb.declaredAtScope, arg, scopes);
     }
 
-    const tb = findReceiverTypeBinding(inScope, text, scopes);
+    const tb = findReceiverTypeBinding(inScope, workingText, scopes);
     if (tb !== undefined) {
       // Map for-of: binding name is `user` but rawType is
       // `__MAP_TUPLE_i__:entries` (see captures.ts) — same extraction as
@@ -167,17 +263,17 @@ export function resolveCompoundReceiverClass(
         if (compound !== undefined) return compound;
       }
     }
-    return findClassBindingInScope(inScope, text, scopes);
+    return findClassBindingInScope(inScope, workingText, scopes);
   }
 
   // Trailing `()` — call expression. Strip it and resolve the function
   // expression's return type. We only handle the canonical `f()` /
   // `obj.method()` shape; nested-arg expressions like `f(g())` are
   // out of scope for V1 (depth-capped recursion catches infinite loops).
-  if (text.endsWith(')')) {
-    const openIdx = matchingOpenParen(text);
+  if (workingText.endsWith(')')) {
+    const openIdx = matchingOpenParen(workingText);
     if (openIdx === -1) return undefined;
-    const fnExpr = text.slice(0, openIdx).trim();
+    const fnExpr = workingText.slice(0, openIdx).trim();
     if (fnExpr.length === 0) return undefined;
 
     const lastDot = fnExpr.lastIndexOf('.');
@@ -286,7 +382,7 @@ export function resolveCompoundReceiverClass(
   // (method return-type). We accept both on each hop because class
   // scopes store both method return types and field types under
   // `typeBindings` keyed by the member name.
-  const parts = splitChainAtTopLevel(text);
+  const parts = splitChainAtTopLevel(workingText);
 
   // Language-specific collection-accessor suffix (C#'s `data.Values`
   // on Dictionary<K,V>, etc.). When the provider hook recognizes


gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -83,12 +84,107 @@ export function resolveCompoundReceiverClass(
...
+      }
+    }
+    if (closeIdx === -1) break;
+    // Extract the type from inside this cast's parens — only capture
+    // the FIRST (outermost) non-parenthesized simple type name as the
+    // effective cast target. CFR patterns like ((Target)((Object)expr))
+    // have a meaningful outer cast and a noise inner cast; we want the
+    // meaningful outer one.
+    const insideParens = workingText.slice(1, closeIdx).trim();
+    if (castType === null && !insideParens.startsWith('(') && /^[a-zA-Z_]\w*$/.test(insideParens)) {
```
magyargergo, line 114: **P0 — Cast-stripping silently discards qualified/generic/array cast types, producing a confidently WRONG `CALLS` edge (not a safe fallback)** `[reproduced]`

This regex only accepts a bare simple identifier as `castType` — correctly rejecting `List<String>`, `Type[]`, `com.example.Foo`. But when it doesn't match, the surrounding loop (lines 96-131) still unwraps the parens down to the underlying expression, and execution falls through to resolve *that* stripped expression via its own pre-cast declared type (~line 266) — silently discarding the cast the source actually wrote.

Verified end-to-end: built and ran the real pipeline (`runPipelineFromRepo`) against this PR's HEAD and its merge-base for `((Box<String>) obj).open()`, `((Foo[]) obj).act()`, `((com.example.Foo) obj).act()`, and `((java.util.List<Foo>) this.field).act()`. BASE emits no `CALLS` edge for any of these (safe no-op, pre-PR behavior). HEAD emits a confident `CALLS` edge to the **wrong class** (`obj`'s own declared type, not the cast target) with `reason: "import-resolved"` for all four — indistinguishable from a correct resolution. The simple-identifier control case (`((Foo) obj).act()`) resolves correctly on HEAD, isolating the regression to exactly generic/array/FQN cast shapes.

This matters specifically for this PR's stated goal: CFR-decompiled Java (the PR's own motivating example) disproportionately produces fully-qualified and generic casts, not simple-identifier casts.

Suggested fix: distinguish "this paren-group looks like a real cast I can't fully parse" (broaden the regex to also match dotted/generic/array shapes) from "this isn't a cast at all," and return `undefined` (matching pre-PR behavior) in the former case instead of falling through to the stripped expression's own type.

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -83,12 +84,107 @@ export function resolveCompoundReceiverClass(
   if (text.length === 0) return undefined;
   const fieldFallback = options.fieldFallback ?? true;
 
+  // ── Pre-processing: strip Java-style cast expressions ───────────
```
magyargergo, line 87: **P1 — New cast-stripping and `this.field` pre-processing run unconditionally for every language, not just Java** `[code-read]`

`compound-receiver.ts` is shared across every C-family language's compound-receiver resolution (cpp, csharp, go, java, javascript, kotlin, php, rust, typescript, vue all route through `resolveCompoundReceiverClass`). This new block (through line 179) has no opt-in gate — unlike every other language-specific behavior already in this file (`options.unwrapCollectionAccessor`, `options.hoistTypeBindingsToModule`, `options.fieldFallback`, all threaded from a per-language `ScopeResolver` flag). It's explicitly documented as Java/CFR-specific ("strip Java-style cast expressions", "CFR decompiler patterns") while running for all languages regardless.

This directly contradicts this repo's own architecture rule (AGENTS.md/CLAUDE.md: "Shared pipeline code in `gitnexus/src/core/ingestion/` must not name languages — use `LanguageProvider`/`ScopeResolver` hooks instead"). Concretely: a C#-style `((IFoo)obj).Method()` receiver previously resolved to `undefined` (safe no-op); it now resolves via this new cast-stripping early-return — a behavior change for a language this PR never intended to touch, with zero test coverage.

Suggested fix: add an opt-in field (e.g. `stripCastExpressions?: boolean`) to `ResolveCompoundReceiverOptions`, thread it from a new `ScopeResolver` hook, and enable it only for Java — mirroring the file's own existing pattern.

gitnexus/src/core/ingestion/languages/java/scope-resolver.ts
```diff
@@ -52,6 +52,7 @@ const javaScopeResolver: ScopeResolver = {
   populateOwners: (parsed: ParsedFile) => populateClassOwnedMembers(parsed),
 
   isSuperReceiver: (text) => text.trim() === 'super',
+  resolveThisViaEnclosingClass: true,
```
magyargergo, line 55: **P1/P2 — `resolveThisViaEnclosingClass: true` activates C++-authored dispatch semantics for Java, untested and currently only coincidentally safe** `[code-read + reproduced]`

Before this PR, C++ was the only language with this flag (`cpp/scope-resolver.ts:268`). It activates 'Case 0.5' in `receiver-bound-calls.ts` (~lines 452-624), a chain-walk written and tuned for C++ member-name semantics. Java doesn't define a `resolveReceiverMember` override (only C++ does, at `cpp/scope-resolver.ts:269`), so Java falls through to the raw C++-tuned logic with no per-language override. Two C++-specific assumptions don't hold for Java: (1) if the nearest class in the MRO declares any overload of the name but none match arity, the walk stops entirely with no edge — Java doesn't hide inherited overloads this way; (2) method lookup unconditionally precedes field lookup, but Java fields/methods can share a name (e.g. records synthesize accessor `x()` alongside field `x`) — a `this.x` field read could resolve to the method instead.

One review lane instrumented the actual pipeline for both scenarios: on today's code, HEAD and BASE produce the *same* (independently wrong, pre-existing) output — not because Case 0.5 is safe, but because two *unrelated* bugs (a duplicate tree-sitter reference-site match, and an old Case-4 arity-blind shortcut) happen to currently mask it. That masking isn't designed or tested, so an unrelated future fix could silently change Java's call graph with no diagnostic trail.

P.S. — this PR also makes two pre-existing comments stale: `receiver-bound-calls.ts:459-461` and `contract/scope-resolver.ts:756-758` both say C++ is 'the sole `resolveThisViaEnclosingClass === true` language' — worth a two-line fix alongside.

Suggested fix: give Java (and future languages) their own `resolveReceiverMember` hook rather than reusing the all-purpose flag, or at minimum add a regression test locking in today's masked behavior.

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -83,12 +84,107 @@ export function resolveCompoundReceiverClass(
...
+
+  // If we found a cast type, resolve it directly — the cast tells us
+  // the exact receiver type for method resolution.
+  if (castType !== null) {
+    const cls = findClassBindingInScope(inScope, castType, scopes);
+    if (cls !== undefined) return cls;
+  }
+
+  // Handle this.field[.subfield] — walk enclosing class's field typeBindings
+  if (workingText.startsWith('this.')) {
```
magyargergo, line 141: **P1 — New `this.field` chain-walker duplicates the existing generic per-segment walker (lines 453-500) and its field-fallback sub-loop (lines 338-356)** `[code-read]`

Verified: this block re-implements, with renamed variables (`curClass`/`seg`/`cleanName` vs. `currentClass`/`segment`/`memberName`), the same per-segment walk and `hoistTypeBindingsToModule` fallback already present lower in this file, plus a near-verbatim copy of the field-fallback sub-loop. This isn't just style debt — it's a real inconsistency: `this.field.field2` now gets an extra field-fallback heuristic that structurally-identical `obj.field.field2` chains (any non-`this.`-prefixed chain) still don't get, because the generic walker they fall through to is unchanged.

Suggested fix: seed the existing generic walker's `currentClass` from `findEnclosingClassDef` when the head segment is `this`, instead of adding a second, narrower implementation.

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -83,12 +84,107 @@ export function resolveCompoundReceiverClass(
...
+    const enclosingClass = findEnclosingClassDef(inScope, scopes);
+    if (enclosingClass !== undefined) {
+      const chainRest = workingText.slice(5); // after 'this.'
+      const parts = chainRest.split('.');
```
magyargergo, line 145: **P2 — `chainRest.split('.')` doesn't skip dots inside balanced parens, unlike this file's own `splitChainAtTopLevel` helper** `[code-read]`

For a receiver like `this.getMap("a.b").field`, the naive split corrupts segments and the chain silently returns `undefined` instead of resolving — exactly the kind of shape CFR-decompiled Java can produce.

Suggested fix: use `splitChainAtTopLevel(chainRest)` (already defined lower in this file) instead of the bare `.split('.')`.

magyargergo (commented):
# PR #2353 Tri-Review Digest — "fix: resolve Java cast-wrapped and this.method() call edges"

Fork `javasecskill/GitNexus` branch `java-call-graph-fix`, head `3905f0b7d`, base `main`
(`859e4b75a`). Two-file diff: `gitnexus/src/core/ingestion/languages/java/scope-resolver.ts` (+1)
and `gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts` (+104/-8).

## Engines and lanes that ran

- **Claude:** GitNexus swarm (6 lanes — facts-historian, branch-hygiene, risk-architect,
  security-boundary, docs-dod, test-ci-verifier) + ce-code-review (5 personas — correctness,
  testing, maintainability, project-standards, adversarial).
- **Codex (GPT-5.5/xhigh):** swarm (solo mode, all 6 lane specs applied by one thread),
  ce-code-review (report-only mode), and a native adversarial-review pass.
- All three Codex legs produced real, substantive, source-grounded findings (git diffs read,
  files inspected line-by-line, prettier/vitest actually invoked where the sandbox allowed) —
  this was a genuine two-engine review, not a degraded run. `claude-swarm-test-ci-verifier.md` is
  the one INCONCLUSIVE lane (agent hung running `vitest` inside the worktree — pre-existing
  worktree/`node_modules` infra issue, not a code problem) but its coordinator-recorded facts
  (CI numbers, prettier diff, zero fixture overlap) are used below since they were verified by
  other lanes independently.
- Coverage: this is a small, fully-read 2-file diff — every lane read both files in full; no
  partial-coverage caveat applies.

---

## Headline finding — P0, cast-stripping regression on generic/array/FQN casts (STRONG, cross-engine, EMPIRICALLY REPRODUCED)

**File:** `gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts:94-138`
(verified directly in the worktree at these line numbers).

The new cast-stripping loop's type-capture regex (line 114: `/^[a-zA-Z_]\w*$/`) only accepts a
bare simple identifier as a cast type — by design, it correctly declines generic (`List<String>`),
array (`Type[]`), and fully-qualified (`com.example.Foo`) cast types. The bug is what happens
*next*: the loop still unwraps the parens and keeps stripping down to the bare underlying
expression regardless, and resolution falls through to resolve *that* stripped expression via its
own pre-cast declared type — silently discarding the cast the source code actually wrote.

**Why this is P0 and not a lower-severity "silent fallback":** `claude-ce-adversarial` didn't just
trace this — it built and ran the real pipeline (`runPipelineFromRepo`) against both HEAD and the
merge-base (`859e4b75a`), reusing the worktree's `node_modules`, on four independent input shapes:
`((Box<String>) obj).open()`, `((Foo[]) obj).act()`, `((com.example.Foo) obj).act()`, and
`((java.util.List<Foo>) this.field).act()`. Result: **BASE emits zero CALLS edges for all four
(safe no-op)**; **HEAD emits a confident CALLS edge to the WRONG class (the pre-cast
variable/field's own declared type) with `reason: "import-resolved"` for all four** — indistinguishable
from a legitimate resolution, no low-confidence marker. The control case (`((Foo) obj).act()`,
simple-identifier cast) correctly resolves on HEAD, isolating the regression to exactly the
generic/array/FQN shapes. `codex-adversarial` independently (code-read only, not executed)
reached the same root cause via a different concrete trigger:
`((com.acme.Target) this.delegate).run()`, `((Target<User>) this.delegate).run()`, and
`((Target) this.delegate).child.run()` — "the resolver falls back to the underlying declared type
or fails the chain head, so the cast-wrapped call edge remains missing or wrong."

**This is genuine cross-engine corroboration on the PR's single most severe finding** — Claude
`[reproduced]` (full pipeline A/B execution) + Codex adversarial `[code-read]` (independent trace,
different concrete inputs, same mechanism). The mechanism was also independently flagged
(without the empirical proof or severity escalation) as a P2 by `claude-swarm-risk-architect`,
`claude-ce-correctness` (Finding 3, P2 conf 65), `claude-ce-testing` (F3, P1 conf 75, "zero
coverage for any of the four documented shapes"), and `codex-swarm` (Finding 3, P2). **This digest
weighs the empirically-verified severity higher: P0 (active regression producing a wrong, confident
edge, not a safe fallback) supersedes the code-read-only P2 reads**, since the "non-Java shape still
falls through to old-style safe fallback" hypothesis those lanes assumed turned out, on actual
execution, to be false for exactly the cast-shapes CFR decompiler output is most likely to produce
(decompilers routinely emit fully-qualified/generic casts) — an empirically-executed result
outweighs a code-read-only trace when they disagree on the same mechanism.

**Why this matters for the PR's stated purpose:** the code's own comments cite "CFR decompiler
patterns" as the motivating input domain. Decompiled Java disproportionately produces exactly the
qualified/generic/array cast shapes this bug mishandles — the simple bare-identifier case this PR
fixes correctly is arguably the *less* common decompiler shape.

**Fix direction (per both lanes):** distinguish "this looks like a type expression I can't fully
parse" (generics/arrays/dotted qualifiers — i.e., it clearly *is* a cast) from "this isn't a cast
at all," and return `undefined` (matching pre-PR safe behavior) in the former case instead of
falling through to the stripped expression's own type.

---

## Finding 2 — P1, shared pipeline code runs new logic unconditionally for every language (STRONG, both engines, many lanes)

**File:** `compound-receiver.ts:87-179` — both the cast-stripping block and the `this.field[.subfield]`
chain walker are inserted at the top of `resolveCompoundReceiverClass`, the single shared function
every C-family language's compound-receiver resolution routes through (cpp, csharp, go, java,
javascript, kotlin, php, rust, typescript, vue — confirmed by grep in multiple lanes). Neither
block is gated behind any `ResolveCompoundReceiverOptions` field or `ScopeResolver` hook — unlike
every other language-specific behavior already in this exact file (`unwrapCollectionAccessor`,
`hoistTypeBindingsToModule`, `fieldFallback`, all opt-in per-language). This directly violates this
repo's own documented rule (AGENTS.md/CLAUDE.md: *"Shared pipeline code in
`gitnexus/src/core/ingestion/` must not name languages — use `LanguageProvider` / `ScopeResolver`
hooks instead"*), and the new code's own comments explicitly name "Java-style" and "CFR decompiler"
while running for all languages regardless.

**Cross-engine corroboration:** `claude-swarm-risk-architect` (Finding 1, P1), `claude-swarm-docs-dod`
(F1, P2), `claude-ce-maintainability` (Finding 1, P1 conf 90), `claude-ce-project-standards`
(Finding 2, P2 conf 65 — judgment call, since gating is shape-based not literal
`if(language==='java')`), `codex-swarm` (Finding 1, "blocks merge: yes"), and `codex-ce`
(Finding 1). Both engines, multiple independent lanes each — this is one of the two most
corroborated findings in the review.

Concretely traced consequence: `claude-swarm-risk-architect` and `claude-ce-testing` (F4) both
hand-traced a C#-style `((IFoo)obj).Method()` receiver — pre-PR this resolved to `undefined`
(safe no-op via the old "trailing `)` — call expression" branch); post-PR it now resolves via the
new cast-stripping early-return, an entirely new, zero-test-coverage code path for a language this
PR never intended to touch.

**Fix direction:** add an opt-in field (e.g. `stripCastExpressions?: boolean`) to
`ResolveCompoundReceiverOptions`, threaded from a new `ScopeResolver` hook, enabled only for Java.

---

## Finding 3 — P1, zero test coverage for two new hand-rolled parsers (STRONG, both engines, nearly every lane)

Confirmed directly: `git diff --name-only` shows only the two source files touched; no fixture
under `test/fixtures/lang-resolution/java-*`, no `compound-receiver`-named test file anywhere in
the repo. `claude-swarm-test-ci-verifier`'s coordinator-verified fact: grepping the entire fixture
corpus for `this\.\w+\.\w+\(` (a `this.field.method()` shape) or any parenthesized-cast receiver
shape in `java-*` fixtures returns **zero matches** — the existing 13,709-test suite structurally
cannot catch a regression in either new code path, because nothing in the corpus exercises them.

**Cross-engine corroboration:** `codex-swarm` (Finding 2, "blocks merge: yes"), `codex-ce`
(Finding 2, "no tests cover new receiver branches"), `claude-swarm-docs-dod` (F4, "the decisive DoD
gate"), `claude-swarm-branch-hygiene` (P2), `claude-swarm-facts-historian`, `claude-ce-testing`
(F1/F2/F3/F4, all P1), `claude-ce-maintainability` (testing gaps), `claude-ce-adversarial`
(testing gaps — notes this is exactly what would have caught the P0 above). This is the second
of the two most corroborated findings in the review, appearing in essentially every lane on both
engines.

---

## Finding 4 — P1 design risk (STRONG mechanism corroboration, differing manifestations/severities across lanes) — Case 0.5 (`resolveThisViaEnclosingClass`) is C++-authored and untested for Java

**Mechanism:** `java/scope-resolver.ts:55` sets `resolveThisViaEnclosingClass: true` — before this
PR, C++ was the only language with this flag (`cpp/scope-resolver.ts:268`). It activates "Case
0.5" in `receiver-bound-calls.ts` (~lines 452-624), which was written and tuned for C++ member-name
semantics. **Coordinator-verified directly in the worktree:** Java does **not** define a
`resolveReceiverMember` hook (grep confirms `resolveReceiverMember` appears only in
`cpp/scope-resolver.ts:269`, nowhere in `languages/java/`), so Java falls through to the raw,
C++-tuned generic chain-walk with no per-language override — unlike C++, which supplies its own
`resolveCppReceiverMember`. Also verified directly: the stale-comment claim that C++ is "the sole
`resolveThisViaEnclosingClass === true` language" appears at both
`receiver-bound-calls.ts:459-461` and `contract/scope-resolver.ts:756-758`.

Two specific C++-shaped assumptions baked into Case 0.5 that don't hold for Java:
1. **Member-name hiding across arity** (`receiver-bound-calls.ts` ~552-559): if the nearest class
   in the MRO declares *any* overload of the called name but none match the call's arity, the walk
   stops entirely (`hiddenByName = true; break`) with **no edge emitted**, never trying an ancestor
   that has the matching overload. Real Java overload resolution considers the whole accessible
   hierarchy; an inherited method isn't "hidden" by an unrelated-arity overload on a subclass.
2. **Method lookup unconditionally precedes field lookup**, regardless of `site.kind` (call vs.
   read/write) — correct for C++ (one shared member namespace) but wrong for Java (fields and
   methods live in separate namespaces and can share a name — e.g. Java records synthesize an
   accessor `x()` alongside field `x`).

**How the lanes converge and diverge on severity:**
- `claude-ce-correctness` (Findings 1 & 2, P1, confidence 75 each) traced concrete failing Java
  inputs by hand against the actual code (`[code-read]`, not executed) — e.g. a `Derived.greet(String,int)`
  overload hiding an inherited `Base.greet(String)` via `this.greet("world")`, and a `Widget.size`
  field/`size()` method collision where `this.size` resolves to the method at confidence 1.0.
- `claude-ce-testing` (F2, P1 conf 75) independently confirmed via fixture grep that **zero**
  existing Java fixtures exercise `this.method()` with overloads, inheritance, or interface/virtual
  dispatch — the only covered case is a trivial zero-arg, zero-overload, single-class `this.save()`.
- `claude-ce-adversarial` (Finding 2, P2, confidence 55) went further and **instrumented the actual
  pipeline** `[reproduced]`: for both of the correctness lane's constructed scenarios, HEAD and
  BASE produce the **same wrong output today** — not because Case 0.5 is safe, but because of two
  *unrelated, incidental* pre-existing bugs (a duplicate tree-sitter reference-site match that lets
  `emitFreeCallFallback` independently reproduce the same wrong answer regardless of what Case 0.5
  decides, and Case 4's own pre-existing arity-blind single-overload shortcut) that happen to
  currently mask Case 0.5's C++-specific behavior. **This masking is not designed or tested — a
  future, unrelated fix to either masking bug would silently flip Java call-graph output with no
  diagnostic trail** (Case 0.5's suppression paths never call `recordResolutionOutcome`, unlike
  Case 4's equivalent path).
- `codex-adversarial` (medium confidence, code-read) independently found a **different concrete
  manifestation of the same root theme**: Case 0.5 doesn't share Case 4's post-resolution behavior
  for interface-owner targets, so `this.method()` calling an interface default method with
  overriding implementors can keep only the primary interface-method edge and drop the
  implementor-dispatch edges Case 4 used to emit.

**Synthesis call:** when lanes disagree on severity for the same mechanism, this digest prefers the
empirically-verified read — `claude-ce-adversarial`'s pipeline run is the only lane that actually
executed real inputs through this path, and it found **no live regression today** (masked by two
unrelated bugs). This is downgraded from "confirmed P1 live bug" to **"confirmed P1/P2 design risk:
untested C++-authored semantics now reachable for Java, currently coincidentally masked, not a
designed safety net."** It is real cross-engine corroboration (Claude correctness + testing +
adversarial, and Codex adversarial, all four independently converge on "Case 0.5 is risky/untested
for Java," via three different concrete mechanisms) — but the "is it live today" question is
answered by the one lane that ran it: no, not today, but fragile.

**Fix direction:** give Java (and any future language) a `resolveReceiverMember` hook that doesn't
hide by arity and doesn't prefer methods over fields for non-call sites, rather than silently
reusing the all-purpose `resolveThisViaEnclosingClass` flag; at minimum, add a regression test
locking in today's masked behavior so an unrelated fix elsewhere doesn't silently change Java's
call graph.

---

## Credited as solid / working correctly (verified by multiple lanes)

- **Simple-identifier cast case** (`((Foo)x).method()`, `((Target)((Object)expr))` nested-cast
  extraction order) — traced by `claude-ce-correctness` (Focus area 1) and independently
  re-verified via standalone script simulation by `claude-ce-adversarial` (scenario 4): correctly
  picks the outermost meaningful cast, correctly skips "noise" inner casts, correctly falls through
  for genuine chained-call receivers like `((A)x).method1()`. **REFUTED as a bug.**
- **`fieldFallback`/Java-static-typing interaction** — `claude-ce-correctness` (Focus area 3) and
  `codex-swarm`/`codex-ce` (verified via grep) both confirm `options.fieldFallback` is correctly
  threaded from `provider.fieldFallbackOnMethodLookup`, which Java sets `false`; the new
  `this.field` walker's fallback branch is correctly disabled for Java's primary paths. Not a live
  bug for Java today (though `claude-ce-maintainability` Finding 3 separately flags the *contract*
  of this flag as silently widened in scope — see below).
- **Cast-stripping loop termination bound** — verified independently by `claude-ce-adversarial`
  (residual risk), `claude-ce-maintainability` (residual risk), `claude-swarm-branch-hygiene` (P2),
  and `claude-swarm-security-boundary` (S-1): the `while(true)` loop provably terminates (every
  path strictly shrinks `workingText`), worst case O(n²) CPU for pathologically nested parens, not
  an infinite loop. Real but low-severity CPU-amplification risk on adversarial decompiled input,
  not a correctness or availability issue at realistic file sizes.
- **`(a || b).field` non-cast parenthesized expression** — REFUTED as mangled by
  `claude-ce-correctness` (Focus area 2): the loop correctly declines to treat it as a cast and
  leaves it unmodified.

## Checked, not an issue (explicitly refuted hypotheses)

- **Nested-cast extraction order `((Target)((Object)expr))`** — refuted, see above.
- **`(a || b).field` mangling** — refuted, see above.
- **`fieldFallback` reuse silently re-enabling Java's disabled dynamic-fallback heuristic** —
  refuted; correctly wired to Java's `fieldFallbackOnMethodLookup: false`.
- **Anonymous-class enclosing-class ambiguity** in the new `this.field` walker's
  `findEnclosingClassDef` vs. the pre-existing `synthesizeJavaReceiverBinding` path — refuted by
  `claude-ce-adversarial` (scenario 5): neither mechanism creates a scope for anonymous class
  bodies (Java's `@scope.class` query doesn't fire on `object_creation_expression`), so both walk
  past it to the same outer real class — a pre-existing shared limitation, not made worse by this
  diff.
- **Cast to an unresolvable/JDK-stdlib type** (`((String)obj).act()`) — working as designed, not a
  bug: falls through to `obj`'s own declared type, a reasonable best-effort default when no more
  specific type info was ever extracted. (Contrast with the P0 above: that case involves a cast
  type that WAS present in source but discarded by regex limits — a materially different, worse
  outcome than this genuinely-unresolvable-type case.)

---

## Other inline-candidate findings (lower priority, mostly Claude-only corroboration)

- **P2 — Naive `chainRest.split('.')` in the new `this.field` walker** (compound-receiver.ts:145)
  doesn't skip dots inside balanced parens/args, unlike the file's own `splitChainAtTopLevel`
  helper used elsewhere. `this.getMap("a.b").field`-shaped chains silently fail (return
  `undefined`) rather than resolving. Moderate corroboration, Claude-only:
  `claude-swarm-risk-architect` (Finding 2), `claude-swarm-security-boundary` (S-2),
  `claude-swarm-docs-dod` (F5). `[code-read]`.
- **P1 — New `this.field` walker duplicates the existing per-segment chain walker and field-fallback
  sub-loop** (compound-receiver.ts:140-179 vs. 338-356/453-500), already stylistically drifted
  (`curClass`/`seg` vs. `currentClass`/`segment`), and introduces a real inconsistency: `this.field.field2`
  now gets an extra field-fallback heuristic that structurally-identical `obj.field.field2` chains
  still don't get. Single-lane (Claude-only): `claude-ce-maintainability` (Finding 2, conf 85).
  Per rule 2, a single-lane P1 with a concrete traced trigger is postable on its own — flagging as
  inline-candidate despite lacking cross-engine corroboration. `[code-read]`.
- **P2 — `fieldFallback` option's contract silently widened** beyond its documented "method lookup
  fallback" scope to also gate the new field-lookup fallback, under a name/doc that only describes
  one of the two uses. Single-lane (Claude-only): `claude-ce-maintainability` (Finding 3, conf 70).
  `[code-read]`.
- **P2 — O(N²) CPU in the cast-stripping loop on adversarial nested-paren input** (e.g. 5000 levels
  of nesting, within the 512KB file cap → ~25M char comparisons per site). Moderate corroboration,
  Claude-only: `claude-swarm-security-boundary` (S-1), `claude-swarm-branch-hygiene` (P2),
  `claude-ce-adversarial` (residual risk, concurs bounded/not correctness). Not merge-blocking
  (no networked surface), real hardening item for `analyze` on untrusted repos. `[code-read]`.
- **P3 — Cast-stripping parser embedded inline rather than extracted as a testable pure helper**,
  unlike the file's other pure string-parsing helpers (`splitChainAtTopLevel`, `matchingOpenParen`).
  Single-lane: `claude-ce-maintainability` (Finding 4). `[code-read]`.
- **P3 — Paren scanner mishandles `)` inside string/char literals** (e.g.
  `((String)someMethod(")"))`— falls through safely, no crash, just a missed cast-type detection).
  Single-lane, medium confidence: `claude-swarm-security-boundary` (S-3). `[code-read]`.

---

## Stale comments (P3, low severity, high corroboration — both engines)

`receiver-bound-calls.ts:459-461` and `contract/scope-resolver.ts:756-758` both state C++ is "the
sole `resolveThisViaEnclosingClass === true` language" — now false after this PR. **Verified
directly in the worktree** (both citations above). No functional impact confirmed (Java doesn't
set `isStaticOnly`, so the specific interaction the comment warns about isn't currently triggered)
— documentation drift only. Corroborated by `claude-swarm-facts-historian`,
`claude-swarm-risk-architect` (Finding 4), `claude-swarm-security-boundary` (S-4),
`claude-swarm-docs-dod` (F2/F3), `claude-ce-project-standards` (Finding 5), and `codex-swarm`
(referenced in Lane 6 DoD findings). Six lanes, both engines — trivial fix (update "C++" to "C++
and Java" in both comments), should be folded in alongside the substantive fixes.

Separately, `claude-ce-project-standards` (Finding 3) explicitly checked and found the new
"Java"/"CFR decompiler" comment *language* in `compound-receiver.ts:87-93` is **not** a standards
violation — the repo's "must not name languages" rule governs branching logic, and this same file
already names languages pervasively in prose comments elsewhere (Python, Ruby, C#). The violation
is Finding 2's ungated *logic*, not this prose.

---

## CI status

- `quality/format` (prettier) **FAILS** — confirmed independently in this session (`npx prettier
  --check compound-receiver.ts` reproduces the failure) and by `claude-swarm-branch-hygiene`,
  `claude-ce-project-standards`, and `codex-ce`/`codex-swarm`. Purely mechanical: 6 sites (brace
  expansion of single-line `if {...; break;}` blocks, one ternary re-join), zero logic changes —
  confirmed by diffing prettier's own output against the current file. Trivial fix: `npx prettier
  --write` on the one file.
- `CI Gate` **FAILS solely by aggregating** the format failure — `quality/typecheck`: SUCCESS,
  `quality/lint`: SUCCESS, all OS test jobs: SUCCESS.
- `Vercel` **FAILS** on an unrelated fork-PR authorization gate ("Authorization required to
  deploy") — org/permissions issue, not code-attributable.
- **All actual tests pass: 13,709 / 13,709, 0 failed, 56 skipped** (`tests/ubuntu/coverage`,
  confirmed by `claude-swarm-facts-historian` and the CI run directly). This is *not* evidence the
  new logic is correct — the existing suite has zero fixture overlap with either new code path
  (confirmed by direct grep against the fixture corpus), so it structurally cannot catch a
  regression in the cast-stripping or `this.field` walker.
- PR body's own "Testing & verification" checklist (`npm test`, `npm run test:integration`, `npx
  tsc --noEmit`) is entirely unchecked, despite prose claiming "Verified on a large-scale Java
  codebase — no edge regressions."

---

## Bottom line

**Branch hygiene:** clean feature/fix PR (linear history, one commit, two causally-coupled files,
no mixed domains — confirmed by `claude-swarm-branch-hygiene`). **Merge state:** checks failing
(`quality/format` FAILS; `CI Gate` FAILS solely as an aggregate of that; all other checks pass).
**Verdict: not production-ready** — do not merge as-is.

Two P1-strength, both-engine-corroborated structural findings (ungated shared-pipeline code;
zero test coverage for two new hand-rolled parsers) plus one P0 finding that Claude's adversarial
lane *proved by executing the real pipeline* and Codex's adversarial lane independently reached by
code trace: the cast-stripping fallthrough turns a safe no-op into a confident wrong `CALLS` edge
for exactly the qualified/generic/array cast shapes CFR-decompiled Java disproportionately
produces — the opposite of this PR's stated goal for its primary target input. A fourth,
lower-confidence-but-multiply-corroborated finding (Case 0.5's C++-authored semantics reused
unmodified for Java) is a real design risk under all four lanes that examined it, but the one lane
that actually executed it found it's currently masked by unrelated bugs rather than live today.
Recommendation: do not merge as-is. Minimum bar to land: (1) fix the cast-stripping fallthrough to
return `undefined` rather than a wrong edge when a real-but-unparseable cast is detected, (2) gate
the two new pre-processing blocks behind an opt-in `ScopeResolver` hook instead of running
unconditionally for every language, (3) add fixtures for at least the cast-wrapped and
`this.field` chain shapes this PR claims to fix, (4) run `prettier --write` on the one file, (5)
fold in the two-line stale-comment fix.

---

*This is an automated multi-tool digest (2 engines × swarm + ce-code-review + adversarial
methodologies). Verify findings against the actual diff before acting on them — in particular,
re-check the P0 finding's empirical repro claim and the Case-0.5 masking analysis, since both rest
on non-trivial pipeline-execution traces rather than simple static reads.*

magyargergo:
Thanks @javasecskill for this fix — the cast-wrapped receiver resolution is a real gap and your simple-identifier cast handling was correct (we fixture-backed it as-is).

Since maintainer edits are disabled on this branch, the review follow-ups landed as #2357, which **supersedes this PR**: it builds on your commit (preserved with authorship) plus one commit per finding from the review above — including the P0 unparseable-cast fallthrough, opt-in gating for the shared pipeline, test coverage for both new behaviors, and an A/B-verified revert of the `resolveThisViaEnclosingClass` flag (the pinning fixtures showed Case 0.5 drops the interface-dispatch fan-out edges Java gets from Case 4).

A maintainer can close this PR in favor of #2357.

Issue #1438, opened 2026-05-09
Bug: gitnexus-hook.cjs findGitNexusDir() matches user-global ~/.gitnexus/ registry, triggers false-positive 'stale index' on every git commit outside indexed repos


albertorsesc:
### Area

`hooks/claude/gitnexus-hook.cjs` (`findGitNexusDir`) → PostToolUse staleness detection

### Summary

`findGitNexusDir(cwd)` walks up to 5 parent directories looking for any folder named `.gitnexus`. Because `gitnexus setup` creates the user-global registry at `~/.gitnexus/` (containing `registry.json`), **any `cwd` under `$HOME` that is NOT inside an indexed repo** still resolves to `~/.gitnexus/` and treats it like a per-repo index. Since the global registry has no `meta.json`, the hook falls through to:

```js
} catch {
  /* no meta — treat as stale */
}
// later:
sendHookResponse('PostToolUse',
  `GitNexus index is stale (last indexed: ${lastCommit ? ... : 'never'}). ` +
  `Run \`${analyzeCmd}\` to update the knowledge graph.`,
);
```

Result: every `git commit|merge|rebase|cherry-pick|pull` in any home-subdirectory git repo (notes, dotfiles, scratch, etc.) emits a spurious "GitNexus index is stale (last indexed: never)" message in Claude Code, prompting the user to `npx gitnexus analyze` a directory that was never meant to be indexed.

### Reproduction

```bash
# 1) Install + setup gitnexus globally
npm i -g gitnexus
gitnexus setup
# 2) Index ONE repo somewhere (e.g. ~/Projects/foo)
cd ~/Projects/foo && gitnexus analyze
# 3) cd to ANY OTHER git repo under $HOME (e.g. a Notes vault)
cd ~/Notes
# 4) Commit anything from inside Claude Code
# 5) PostToolUse hook fires with "GitNexus index is stale (last indexed: never)" pointing at ~/.gitnexus/
```

The hook also fires for any non-indexed repo whose path contains a parent that matches the 5-level walk to `~/.gitnexus/`.

### Expected behavior

`findGitNexusDir` should only match a per-repo GitNexus index — i.e. a directory containing both `meta.json` AND `lbug`. The user-global registry at `~/.gitnexus/` should be skipped because it never has those files (registry only).

### Proposed fix

Tighten the existence check in `findGitNexusDir` from `fs.existsSync(.gitnexus)` to require both index files:

```js
[...]

PR #2343, opened 2026-07-01
fix: consolidate icon imports, fix stale refs and package name collision

fix2015:
Noticed three components were importing directly from \`lucide-react\` instead of using the centralized \`@/lib/lucide-icons\` module that the rest of the codebase goes through. The module header comment even says all imports should go through it. Added the missing \`Keyboard\` and \`BarChart2\` exports and updated the imports in HelpPanel, ProcessesPanel, and ProcessFlowModal.

While I was reading through the code I also fixed a few other small things:
- duplicate mermaid init comment in ProcessFlowModal (same line twice)
- placeholder \`#XXX\` issue number in git.ts — replaced with a descriptive note
- stale "KuzuDB" reference in ARCHITECTURE.md — should be "LadybugDB" since the migration
- gitnexus-web's package.json had \`"name": "gitnexus"\` which collides with the CLI package — changed to \`"gitnexus-web"\`

PR #2363, opened 2026-07-03
fix(storage): rename index metadata to gitnexus.json with dual-write compatibility

LivioGama:
## Summary

Renames the per-repo index metadata file from `.gitnexus/meta.json` to `.gitnexus/gitnexus.json` — with **permanent backward compatibility**: `saveMeta` dual-writes both filenames on every analyze, each run reconciles the two (fresher `indexedAt` wins, nothing is ever deleted), and `loadMeta` falls back to the legacy mirror when the primary is absent. Includes a 16-commit review-fix series addressing all findings from the [tri-review](https://github.com/abhigyanpatwari/GitNexus/pull/2363#pullrequestreview-4624733676).

## Motivation / context

The original goal was giving the index metadata a filename distinct from the generic `meta.json`. Note this has **no bearing on git-worktree behavior** (each worktree already gets its own independent `.gitnexus/`; the directory stays fully git-ignored) — the module docs now state the honest rationale. Review findings: [review 4624733676](https://github.com/abhigyanpatwari/GitNexus/pull/2363#pullrequestreview-4624733676); fix series summarized in [this comment](https://github.com/abhigyanpatwari/GitNexus/pull/2363#issuecomment-4877031772).

## Areas touched

- [x] Docs / agent config only (`AGENTS.md`, `CLAUDE.md`, `.cursor/`, `llms.txt`, etc.) — plus all 4 shipped editor hooks (`gitnexus/hooks/claude`, `gitnexus/hooks/antigravity`, `gitnexus-claude-plugin`, `gitnexus-cursor-integration`)

## Scope & constraints

**In scope**

- `gitnexus.json` as the primary metadata filename; `meta.json` kept as an always-in-sync mirror (dual-write + per-analyze reconciliation, `storage/repo-manager.ts`)
- Strict `loadMeta` fallback: legacy mirror only on ENOENT/ENOTDIR; a corrupt/unreadable primary returns null → safe full rebuild
- All 4 shipped hooks read `gitnexus.json` first with `meta.json` fallback (+ drift-guard test)
- Collision-safe atomic writes (`storage/fs-atomic.ts`, random tmp suffix + `retryRename`)
- Registry validation, `index`/`remove` CLI, group staleness, `collectBranchCacheKeys` — all dual-file aware
- Docs: ARCHITECTURE/GUARDRAILS/RUNBOOK/`.cursor/index.mdc` + new MIGRATION.md entry
- Tests: reconciliation suite, remove-command suite, un-vacuumed staleness integration tests, dual-filename #1169 guards

- Removing the legacy `meta.json` mirror — deferred to a future major version, announced in MIGRATION.md first
- Making `.gitnexus/` git-trackable / any actual worktree behavior change
- Deduplicating the four copy-paste hook files into a shared module (drift-guard test mitigates)

## Implementation notes

- Design follows MIGRATION.md's `OVERRIDES → METHOD_OVERRIDES` precedent: dual-support at runtime, no forced deletion, removal only at a future major. This closes the mixed-version (old MCP server), rollback, and stale-shadow hazards the one-way migrate-then-delete had.
- Reconciliation (`reconcileMetadataFiles`) is best-effort: failures (EROFS/ENOSPC) warn and continue; per-branch errors are isolated so one bad branch dir can't starve siblings.
- `retryRename` moved from `core/group/bridge-db.ts` to `storage/fs-atomic.ts` (kept re-exported) to avoid a backwards `storage/ → core/group/` import.

## Testing & verification

- [x] `cd gitnexus && npm test` — full suite run locally; all touched suites green in isolation (remaining bulk-run failures are the documented parallel-load worker-exit flakes, verified pass-in-isolation 68/68)
- [x] `cd gitnexus && npm run test:integration` — incl. `cli-e2e` 34/34 (real analyzes assert BOTH metadata filenames + content equality), staleness integration tests now assert a real pool reinit
- [x] `cd gitnexus && npx tsc --noEmit` (also clean for `gitnexus-shared`)
- [ ] `gitnexus-web` unchanged
- Note: CI requires maintainer approval for fork-PR workflows; Vercel check is standing fork-auth noise

## Risk & rollout

- **No user action needed.** A pre-existing repo with only `meta.json` gets `gitnexus.json` bootstrapped on its next analyze; both files then stay in sync.
- **Rollback-safe:** downgrading to an older binary works — `meta.json` is always present and current (including the `incrementalInProgress` crash-recovery flag).
- **Mixed versions converge:** an older binary writing only `meta.json` is picked up by reconciliation (fresher stamp wins) instead of being shadowed.
- MIGRATION.md entry documents the change, the rollback story, and the future-major removal policy for the mirror.

## Checklist

- [x] If `AGENTS.md` / overlays changed: headers, scope block, and changelog updated per project conventions (`.cursor/index.mdc` embeddings rule updated)

### DIFF
--- a/gitnexus/package-lock.json
+++ b/gitnexus/package-lock.json
@@ -4222,6 +4222,12 @@
         "onnxruntime-common": "1.27.0"
       }
     },
+    "node_modules/onnxruntime-node/node_modules/onnxruntime-common": {
+      "version": "1.26.0",
+      "resolved": "https://registry.npmjs.org/onnxruntime-common/-/onnxruntime-common-1.26.0.tgz",
+      "integrity": "sha512-qVyMR4lcWgbkc4getFV+GQijsTnbg/siteoqcDwa3sI/LxbrMSNw4ePyvCq/ymdQaRomCA7YuWmhzsswxvymdw==",
+      "license": "MIT"
+    },
     "node_modules/onnxruntime-web": {
       "version": "1.26.0-dev.20260416-b7804b056c",
       "resolved": "https://registry.npmjs.org/onnxruntime-web/-/onnxruntime-web-1.26.0-dev.20260416-b7804b056c.tgz",
--- a/gitnexus/package-lock.json
+++ b/gitnexus/package-lock.json
@@ -4222,12 +4222,6 @@
         "onnxruntime-common": "1.27.0"
       }
     },
-    "node_modules/onnxruntime-node/node_modules/onnxruntime-common": {
-      "version": "1.26.0",
-      "resolved": "https://registry.npmjs.org/onnxruntime-common/-/onnxruntime-common-1.26.0.tgz",
-      "integrity": "sha512-qVyMR4lcWgbkc4getFV+GQijsTnbg/siteoqcDwa3sI/LxbrMSNw4ePyvCq/ymdQaRomCA7YuWmhzsswxvymdw==",
-      "license": "MIT"
-    },
     "node_modules/onnxruntime-web": {
       "version": "1.26.0-dev.20260416-b7804b056c",
       "resolved": "https://registry.npmjs.org/onnxruntime-web/-/onnxruntime-web-1.26.0-dev.20260416-b7804b056c.tgz",
--- a/gitnexus/src/cli/remove.ts
+++ b/gitnexus/src/cli/remove.ts
@@ -1,9 +1,12 @@
 /**
  * Remove Command (#664)
  *
- * Delete the `.gitnexus/` index for a registered repo and unregister it
- * from the global registry (~/.gitnexus/registry.json). The target is
- * identified by alias / basename-derived name / remote-inferred name /
+ * Delete the GitNexus index for a registered repo and unregister it
+ * from the global registry (~/.gitnexus/registry.json). This includes:
+ * - The gitnexus.json metadata file (new format)
+ * - The .gitnexus/ directory
+ *
+ * The target is identified by alias / basename-derived name / remote-inferred name /
  * absolute path — no `--repo` flag, just a positional argument so the
  * destructive-command ergonomics match `clean` (which is also
  * destructive but scoped to `process.cwd()`).
@@ -27,6 +30,7 @@
  */
 
 import fs from 'fs/promises';
+import path from 'path';
 import { logger } from '../core/logger.js';
 import { cliError } from './cli-message.js';
 import { t } from './i18n/index.js';
@@ -35,6 +39,7 @@ import {
   resolveRegistryEntry,
   assertSafeStoragePath,
   unregisterRepo,
+  INDEX_METADATA_FILE,
   RegistryNotFoundError,
   RegistryAmbiguousTargetError,
   UnsafeStoragePathError,
@@ -101,7 +106,17 @@ export const removeCommand = async (target: string, options?: { force?: boolean 
   // orphaned — `listRegisteredRepos({ validate: true })` prunes those on
   // next read, so the failure is self-healing.
   try {
+    // Delete the metadata file first so a partial rm cannot leave a valid-looking index.
+    const metadataPath = path.join(entry.storagePath, INDEX_METADATA_FILE);
+    try {
+      await fs.unlink(metadataPath);
+    } catch {
+      // File may not exist (legacy format) - ignore
+    }
+
+    // Delete the .gitnexus/ directory.
     await fs.rm(entry.storagePath, { recursive: true, force: true });
+
     await unregisterRepo(entry.path);
     console.log(t('remove.removed', { name: entry.name }));
     console.log(`   ${t('common.path')}:    ${entry.path}`);
--- a/gitnexus/src/cli/remove.ts
+++ b/gitnexus/src/cli/remove.ts
@@ -1,10 +1,10 @@
 /**
  * Remove Command (#664)
  *
- * Delete the GitNexus index for a registered repo and unregister it
- * from the global registry (~/.gitnexus/registry.json). This includes:
- * - The gitnexus.json metadata file (new format)
- * - The .gitnexus/ directory
+ * Delete the `.gitnexus/` index directory for a registered repo (including
+ * both metadata filenames — gitnexus.json and its legacy meta.json mirror —
+ * which live inside it) and unregister it from the global registry
+ * (~/.gitnexus/registry.json).
  *
  * The target is identified by alias / basename-derived name / remote-inferred name /
  * absolute path — no `--repo` flag, just a positional argument so the
@@ -30,7 +30,6 @@
  */
 
 import fs from 'fs/promises';
-import path from 'path';
 import { logger } from '../core/logger.js';
 import { cliError } from './cli-message.js';
 import { t } from './i18n/index.js';
@@ -39,7 +38,6 @@ import {
   resolveRegistryEntry,
   assertSafeStoragePath,
   unregisterRepo,
-  INDEX_METADATA_FILE,
   RegistryNotFoundError,
   RegistryAmbiguousTargetError,
   UnsafeStoragePathError,
@@ -106,17 +104,7 @@ export const removeCommand = async (target: string, options?: { force?: boolean 
   // orphaned — `listRegisteredRepos({ validate: true })` prunes those on
   // next read, so the failure is self-healing.
   try {
-    // Delete the metadata file first so a partial rm cannot leave a valid-looking index.
-    const metadataPath = path.join(entry.storagePath, INDEX_METADATA_FILE);
-    try {
-      await fs.unlink(metadataPath);
-    } catch {
-      // File may not exist (legacy format) - ignore
-    }
-
-    // Delete the .gitnexus/ directory.
     await fs.rm(entry.storagePath, { recursive: true, force: true });
-
     await unregisterRepo(entry.path);
     console.log(t('remove.removed', { name: entry.name }));
     console.log(`   ${t('common.path')}:    ${entry.path}`);
--- a/gitnexus/src/core/run-analyze.ts
+++ b/gitnexus/src/core/run-analyze.ts
@@ -55,6 +55,8 @@ import {
   registerRepo,
   isRepoRegistered,
   cleanupOldKuzuFiles,
+  migrateToCommittedIndex,
+  INDEX_METADATA_FILE,
   INCREMENTAL_SCHEMA_VERSION,
   type RepoMeta,
 } from '../storage/repo-manager.js';
@@ -363,12 +365,12 @@ export const primaryInversionWarning = (
 
 /**
  * Collect the recorded parse-cache chunk keys across the flat + every branch
- * meta under a flat `.gitnexus` storage, EXCLUDING `excludeDir` (the current
- * run's own meta dir) so a single-branch repo collects nothing and its prune
- * stays byte-identical to today (#2106 R6). `complete` is false when a sibling
- * meta.json exists but fails to parse — callers then retain the whole shared
- * cache rather than over-evict another branch's still-live shards. Exported for
- * testing.
+ * metadata directory under a flat `.gitnexus` storage, EXCLUDING `excludeDir`
+ * (the current run's own meta dir) so a single-branch repo collects nothing and
+ * its prune stays byte-identical to today (#2106 R6). `complete` is false when
+ * a sibling metadata file exists but fails to read or parse — callers then
+ * retain the whole shared cache rather than over-evict another branch's
+ * still-live shards. Exported for testing.
  */
 export const collectBranchCacheKeys = async (
   storagePath: string,
@@ -384,9 +386,20 @@ export const collectBranchCacheKeys = async (
     if (excludeDir && path.resolve(dir) === path.resolve(excludeDir)) continue;
     let raw: string;
     try {
-      raw = await fs.readFile(path.join(dir, 'meta.json'), 'utf-8');
-    } catch {
-      continue; // no meta here — not a branch index, not a failure
+      raw = await fs.readFile(path.join(dir, INDEX_METADATA_FILE), 'utf-8');
+    } catch (newErr) {
+      const newCode = (newErr as NodeJS.ErrnoException)?.code;
+      if (newCode !== 'ENOENT' && newCode !== 'ENOTDIR') {
+        complete = false;
+        continue;
+      }
+      try {
+        raw = await fs.readFile(path.join(dir, 'meta.json'), 'utf-8');
+      } catch (legacyErr) {
+        const legacyCode = (legacyErr as NodeJS.ErrnoException)?.code;
+        if (legacyCode !== 'ENOENT' && legacyCode !== 'ENOTDIR') complete = false;
+        continue; // no metadata here — not a branch index, not a failure
+      }
     }
     try {
       const parsed = JSON.parse(raw) as { cacheKeys?: unknown };
@@ -614,10 +627,12 @@ export async function runFullAnalysis(
   const branchLabel = options.branch ?? checkedOutBranch;
   const placement = await resolveBranchPlacement(repoPath, branchLabel);
   const { lbugPath, metaPath } = getStoragePaths(repoPath, placement.branch);
-  // Directory that owns this run's meta.json (flat `.gitnexus` for the primary
-  // slot, `branches/<slug>/` otherwise). loadMeta/saveMeta operate on it so
-  // each branch keeps its own lastCommit / fileHashes / incremental dirty flag.
+  // metaPath now points to the metadata file (gitnexus.json) in a branch-specific directory.
+  // metaDir is the directory containing the metadata file (and branch-specific DBs).
   const metaDir = path.dirname(metaPath);
+
+  // Auto-migrate from legacy .gitnexus/meta.json to gitnexus.json.
+  await migrateToCommittedIndex(repoPath);
 
   const existingMeta = await loadMeta(metaDir);
 
--- a/gitnexus/src/core/run-analyze.ts
+++ b/gitnexus/src/core/run-analyze.ts
@@ -56,6 +56,7 @@ import {
   isRepoRegistered,
   cleanupOldKuzuFiles,
   reconcileMetadataFiles,
+  isMissingFilesystemError,
   INDEX_METADATA_FILE,
   INCREMENTAL_SCHEMA_VERSION,
   type RepoMeta,
@@ -367,10 +368,13 @@ export const primaryInversionWarning = (
  * Collect the recorded parse-cache chunk keys across the flat + every branch
  * metadata directory under a flat `.gitnexus` storage, EXCLUDING `excludeDir`
  * (the current run's own meta dir) so a single-branch repo collects nothing and
- * its prune stays byte-identical to today (#2106 R6). `complete` is false when
- * a sibling metadata file exists but fails to read or parse — callers then
- * retain the whole shared cache rather than over-evict another branch's
- * still-live shards. Exported for testing.
+ * its prune stays byte-identical to today (#2106 R6 — the byte-identity claim
+ * is about the PRUNE result; the metadata FILENAME read here changed with
+ * PR #2363's rename, checking `gitnexus.json` first then the legacy
+ * `meta.json` mirror). `complete` is false when a sibling metadata file exists
+ * but fails to read or parse — callers then retain the whole shared cache
+ * rather than over-evict another branch's still-live shards. Exported for
+ * testing.
  */
 export const collectBranchCacheKeys = async (
   storagePath: string,
@@ -388,16 +392,14 @@ export const collectBranchCacheKeys = async (
     try {
       raw = await fs.readFile(path.join(dir, INDEX_METADATA_FILE), 'utf-8');
     } catch (newErr) {
-      const newCode = (newErr as NodeJS.ErrnoException)?.code;
-      if (newCode !== 'ENOENT' && newCode !== 'ENOTDIR') {
+      if (!isMissingFilesystemError(newErr)) {
         complete = false;
         continue;
       }
       try {
         raw = await fs.readFile(path.join(dir, 'meta.json'), 'utf-8');
       } catch (legacyErr) {
-        const legacyCode = (legacyErr as NodeJS.ErrnoException)?.code;
-        if (legacyCode !== 'ENOENT' && legacyCode !== 'ENOTDIR') complete = false;
+        if (!isMissingFilesystemError(legacyErr)) complete = false;
         continue; // no metadata here — not a branch index, not a failure
       }
     }
--- a/gitnexus/src/core/run-analyze.ts
+++ b/gitnexus/src/core/run-analyze.ts
@@ -55,7 +55,7 @@ import {
   registerRepo,
   isRepoRegistered,
   cleanupOldKuzuFiles,
-  migrateToCommittedIndex,
+  reconcileMetadataFiles,
   INDEX_METADATA_FILE,
   INCREMENTAL_SCHEMA_VERSION,
   type RepoMeta,
@@ -631,8 +631,17 @@ export async function runFullAnalysis(
   // metaDir is the directory containing the metadata file (and branch-specific DBs).
   const metaDir = path.dirname(metaPath);
 
-  // Auto-migrate from legacy .gitnexus/meta.json to gitnexus.json.
-  await migrateToCommittedIndex(repoPath);
+  // Keep gitnexus.json and the legacy meta.json mirror in sync (fresher
+  // indexedAt wins; nothing is deleted). Best-effort: loadMeta has its own
+  // legacy fallback, so a reconciliation failure (read-only mount, full disk)
+  // must never abort the analyze run — a repo that indexed fine read-only
+  // before the rename must keep doing so.
+  try {
+    await reconcileMetadataFiles(repoPath);
+  } catch (err) {
+    const code = (err as NodeJS.ErrnoException)?.code;
+    log(`Metadata reconciliation failed (non-critical${code ? `, ${code}` : ''}); continuing.`);
+  }
 
   const existingMeta = await loadMeta(metaDir);
 
--- a/gitnexus/src/storage/repo-manager.ts
+++ b/gitnexus/src/storage/repo-manager.ts
@@ -127,12 +127,13 @@ export interface RepoMeta {
    */
   fileHashes?: Record<string, string>;
   /**
-   * Crash-recovery dirty flag — a generic marker written to meta.json
-   * BEFORE any destructive DB mutation by BOTH writeback branches
-   * (incremental since its introduction; full rebuilds over an existing
-   * meta since #2099 F1); cleared on success by overwriting meta.json.
-   * If a run crashes between, the next run sees the flag and forces a
-   * full rebuild — the cheapest path back to a known-good index.
+   * Crash-recovery dirty flag — a generic marker written to the metadata
+   * file (gitnexus.json + its meta.json mirror) BEFORE any destructive DB
+   * mutation by BOTH writeback branches (incremental since its introduction;
+   * full rebuilds over an existing meta since #2099 F1); cleared on success
+   * by overwriting the metadata file. If a run crashes between, the next
+   * run sees the flag and forces a full rebuild — the cheapest path back
+   * to a known-good index.
    */
   incrementalInProgress?: {
     /** When the run started (epoch ms). */
@@ -143,9 +144,9 @@ export interface RepoMeta {
   };
   /**
    * Name of the git branch this index represents (#2106). Absent for the
-   * default/legacy single-branch case so the flat `meta.json` stays
+   * default/legacy single-branch case so the flat metadata file stays
    * byte-identical to pre-multi-branch output. When present in the FLAT
-   * `meta.json`, it records which branch "owns" the flat slot (the first
+   * metadata file, it records which branch "owns" the flat slot (the first
    * branch indexed); per-branch indexes under `branches/<slug>/` always carry
    * their own `branch`.
    */
@@ -412,7 +413,7 @@ export const cleanupOldKuzuFiles = async (
  * Returns null when the file is absent, unreadable, or unparseable — a
  * corrupt legacy file is treated the same as a missing one (safe rebuild).
  */
-export const loadMetaLegacy = async (metaDir: string): Promise<RepoMeta | null> =>
+const loadMetaLegacy = async (metaDir: string): Promise<RepoMeta | null> =>
   tryReadMetaFile(metaDir, LEGACY_METADATA_FILE);
 
 /**
@@ -504,7 +505,7 @@ export const hasIndex = async (repoPath: string): Promise<boolean> => {
   } catch {
     // Fall back to legacy location
     try {
-      await fs.access(path.join(paths.storagePath, 'meta.json'));
+      await fs.access(path.join(paths.storagePath, LEGACY_METADATA_FILE));
       return true;
     } catch {
       return false;
@@ -1165,11 +1166,11 @@ export class RegistryAmbiguousTargetError extends Error {
 
 /**
  * Thrown by {@link assertAnalysisFinalized} when a successful `analyze`
- * run did not actually persist `meta.json` or did not register the repo
- * in `~/.gitnexus/registry.json` (#1169).
+ * run did not actually persist the index metadata file or did not register
+ * the repo in `~/.gitnexus/registry.json` (#1169).
  *
  * Why this exists: on Windows, `gitnexus analyze` has been observed to
- * exit cleanly (code 0) with `lbug.wal` written but no `meta.json`,
+ * exit cleanly (code 0) with `lbug.wal` written but no metadata file,
  * leaving the repo invisible to `gitnexus list`/`status` and downstream
  * MCP discovery. The only signal to the user was an empty banner —
  * which is indistinguishable from a no-op early return. This invariant
@@ -1188,7 +1189,7 @@ export class AnalysisNotFinalizedError extends Error {
   ) {
     const detail =
       missing === 'meta'
-        ? `meta.json was not written to ${path.join(storagePath, 'meta.json')}`
+        ? `${INDEX_METADATA_FILE} was not written to ${path.join(storagePath, INDEX_METADATA_FILE)}`
         : `registry entry for ${repoPath} was not added to ${registryPath}`;
     super(
       `Analysis did not finalize for ${repoPath}: ${detail}. ` +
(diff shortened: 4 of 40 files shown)


magyargergo (commented):
# Tri-Review Digest — PR #2363 "Fix git worktree metadata compatibility"

**Verdict: not production-ready** — but the core registry/fallback rewrite is solid and the PR is fixable in place. Head `81be46a7`, base `1029a8dd`, 11 files (+412/−106). No linked issue; PR body is the unfilled template.

**Merge state:** MERGEABLE, 2 doc-only commits BEHIND main (no semantic-conflict risk on touched files); required CI never ran (fork gate). **Branch hygiene:** single linear commit, coherent single-domain change — do not split — but polluted by unrelated `package-lock.json` churn and a non-conventional title.

## Engine breakdown & leg liveness

| Leg | Engine | Status |
|---|---|---|
| A1 swarm (6 personas: facts-historian, risk-architect, security-boundary, test-ci, docs-dod, branch-hygiene) | Claude | live, full reports |
| A2 ce-code-review (6 personas: correctness, reliability, maintainability, testing, project-standards, adversarial) | Claude | live, full reports; 2 findings reproduced via live repros |
| B1 swarm | Codex GPT-5.5/xhigh | live |
| B2 ce | Codex GPT-5.5/xhigh | live |
| B3 adversarial | Codex GPT-5.5/xhigh | live |
| Coordinator verification | Claude (main) | live; 1 failing test reproduced |

Both engines ran the same swarm+ce scaffolding, so cross-engine agreement below is a clean A/B signal. Confidence tiers: **Strong** = Codex leg + Claude lane agree; **Moderate** = agreement within one engine only; **Weak** = single lane.

## CI status

The core suite **never ran**: fork-PR `action_required` gate — visible checks are only "Apply conventional label" (pass), "Validate PR title" (skipped), and Vercel (standing fork auth noise). ci-tests.yml / ci-quality.yml / ci-e2e.yml are all absent. This matters doubly because **CI would be red**: `test/unit/run-analyze-fts-repair.test.ts:419` fails on this head ([reproduced] by coordinator: 1 failed / 9 passed — `ENOENT: open '<tmp>/.gitnexus/meta.json'` because analyze now writes `gitnexus.json` while the test reads `meta.json` raw). `test/integration/cli-e2e.test.ts:393` and `:440` read the same path post-analyze (the `:393` one guards the #1169 silent-finalize symptom), and `test/integration/antigravity-hook-e2e.test.ts` writes `meta.json` fixtures throughout (17 references). Cross-engine: Codex B1 (P1) + B2 (P2) flagged all three sites independently of the coordinator's repro.

## What's verified solid (credit where due)

Multiple lanes in both engines attacked these and **refuted** themselves — the core of the PR is genuinely correct:

- **`listRegisteredRepos` two-file keep/prune rewrite is correct.** All four error combinations hand-traced and verified by 4 independent lanes (both-missing → prune; any transient error on either file → keep+warn; `firstNonMissingError` correctly dominates). The new `it.each` dual-file tests are well-asserted (disk persistence checked, not just return values). The 4 changed test files pass locally: 145/145.
- **local-backend staleness swap is behavior-preserving.** Conflict resolved by synthesis reading both versions: the pre-image's raw `readFile`/`JSON.parse` failures were swallowed by the same `catch { return; /* Can't read meta — assume pool is fine */ }` (base local-backend.ts, unchanged at head :1330-1332); the new `if (!meta) return;` (:1305) is semantically identical for same-version operation. The security lane's "silent-return behavior change" claim is **refuted**; the *mixed-version* variant remains real (headline 2).
- **`collectBranchCacheKeys` error handling is an improvement, not a regression** — non-missing errors now set `complete = false` (under-evict, cache retained) where old code silently continued (could over-evict a live branch's shards).
- **All five security probes refuted with mechanism** (remove.ts path safety via `assertSafeStoragePath`; symlink-following requires pre-existing local write access; JSON.parse prototype pollution inert; lockfile removal is correct dedup; no log injection). `git diff --check` clean, no bidi controls.
- Migration crash windows within a **single version** are safe: saveMeta is write+rename atomic; crash between rename and unlink leaves two identical files (benign until a legacy writer diverges them — headline 3).

## Headline findings

### 1. [P0, Strong, reproduced] All three shipped hooks still read `meta.json` → permanent false-stale nag for every user, after every commit
`gitnexus/hooks/claude/gitnexus-hook.cjs:446`, `gitnexus/hooks/antigravity/gitnexus-antigravity-hook.cjs:430`, `gitnexus-claude-plugin/hooks/gitnexus-hook.js:466` — all do `JSON.parse(fs.readFileSync(path.join(gitNexusDir, 'meta.json')))` with `catch { /* no meta — treat as stale */ }`. After the first new-version analyze deletes `meta.json` (migration runs unconditionally at run-analyze.ts:635), every successful `git commit|merge|rebase|pull` in an agent session fires "GitNexus index is stale (last indexed: never)" despite a fresh index — forever — and `hadEmbeddings` is lost so the suggested re-analyze drops `--embeddings`. **Reproduced** by ce-correctness (piped a PostToolUse commit event into the hook against a gitnexus.json-only repo → false stale emitted; legacy repo → correctly silent). This reintroduces the exact failure mode of closed issues **#1438 / #1137**. Found by all 3 Codex legs + 3 Claude lanes + coordinator (exact lines verified). Fix in-PR: gitnexus.json-first read with meta.json fallback in all three hook copies (plus the `isGlobalRegistryDir` meta.json sentinel at ~line 41/33 in each — misleading post-migration though its final answer stays correct), plus hook-test coverage for the new format.

### 2. [P1, Strong] One-way migration deletes `meta.json` — mixed-version blindness, rollback data loss, and a reproduced stale-shadow state
`repo-manager.ts:507` unlinks the legacy file after copying. Three concrete failure chains, each independently verified:
- **Old MCP server still running** (editor-embedded, pinned Docker, long-lived npx): its 5s staleness probe reads only `meta.json` → ENOENT → `catch { return }` → **silently serves the pre-analyze index until process restart**. The staleness probe is the only rebuild-pickup mechanism (#297). (Claude risk-architect + ce-adversarial F2; Codex B3 high + B1 P2.)
- **Rollback/downgrade**: old `loadMeta` reads only `meta.json` → null → "no prior index" → silent full rebuild, losing incremental state and the `incrementalInProgress` crash-recovery flag.
- **Stale-shadow [reproduced]** (ce-adversarial F1, tsx repro against the PR's own module): once `gitnexus.json` exists, `migrateToCommittedIndex` never looks at legacy again (`fs.access` gate, :499) and `loadMeta` blindly prefers the new file (:420-427). If an older binary re-analyzes (writing a *fresher* meta.json, possibly with the dirty flag), the new version permanently trusts the **stale** gitnexus.json → false `alreadyUpToDate` certification of a DB built at a different commit (run-analyze.ts:820), skipped crash recovery (:746), stale group/MCP staleness reads. Repro output: `migrated=false`, `lastCommit=AAAA` (stale), dirty flag LOST.

Recommendation (consensus of both engines): **do not delete `meta.json` in this PR**. Keep or dual-write legacy for a compatibility window (remove at a future major); if both files exist, reconcile by `indexedAt` instead of blind-preferring the new one.

### 3. [P1, Strong] `loadMeta` blanket catch masks a corrupt/unreadable `gitnexus.json` with stale legacy data
`repo-manager.ts:424` — `catch { return loadMetaLegacy(metaDir); }` catches **parse errors and EACCES/EIO**, not just ENOENT. Pre-PR, corrupt meta → null → forced full rebuild (the fail-safe the saveMeta docstring itself relies on). Now a corrupt new file + lingering legacy (one failed unlink suffices, and the unlink is warn-and-continue with no retry — the access gate then skips cleanup forever) → incremental run over a stale `fileHashes` baseline with no dirty flag. Cross-engine: Codex B2 P1 + B1 P2 + Claude ce-adversarial F4. Fix: fall back only on ENOENT/ENOTDIR — `isMissingFilesystemError` already exists in this file at :573 (and is inline-duplicated twice in run-analyze.ts:391/399 — export and reuse it).

### 4. [P1, Moderate→postable] `migrateToCommittedIndex`: destructive, runs on every analyze, zero tests, and an uncaught `saveMeta` aborts analyze on read-only/full disks
- **Zero test coverage** (grep of test/ = no matches) for a function that unlinks files unconditionally on every `gitnexus analyze` (Claude test-ci P0 + ce-testing P1; Codex B2 asks for migration tests — Strong).
- **Uncaught `saveMeta`** at :504, and the call site `run-analyze.ts:635` has no try/catch: EROFS/EACCES/ENOSPC on `.gitnexus/` → the whole analyze aborts, **repeatedly, on every run**, for a repo that worked fine read-only pre-PR (loadMeta's fallback makes the migration write non-load-bearing — its failure should never be fatal). The same file already has the `isReadOnlyFilesystemError` + warn pattern to copy. (ce-reliability P1 + ce-testing; Claude-only but concrete trigger → stays P1 per severity rule.)
- **Fixed-name `.tmp` collision [reproduced]**: `${metaPath}.tmp` (:445) — the tmp-name pattern itself is pre-existing (base `saveMeta` used the same name), but the unconditional migration adds a new concurrent writer on every analyze (CI matrix, auto-sync #2329) → loser's rename throws raw ENOENT and kills its analyze; interleavings can momentarily install a truncated JSON (feeding headline 3). Repro: `Promise.allSettled` of two saveMeta → `['REJECTED: ENOENT', 'ok']`. Fix: PID/random tmp suffix + make the migrate call non-fatal.
- **Branch-loop error swallowing**: the catch at :545 (written for "branchesDir may not exist") wraps the whole per-branch loop — one dangling symlink/EACCES dir silently aborts migration of all later branches, zero logging, permanently (stable readdir order). Per-branch try/catch + `logger.warn`. (3 Claude lanes.)

### 5. [P1 as review-blocker, Strong] The stated purpose is not achieved — the rename does nothing for git worktrees
`.gitnexus/` remains fully git-ignored (`ensureGitNexusIgnored` / `GITNEXUS_EXCLUDE_ENTRY = '.gitnexus/'` untouched, blanket `*` self-.gitignore intact), so renaming a file inside it is invisible to git; each worktree already had its own independent `.gitnexus/` by construction (`getStoragePath` is per-checkout), and linked-worktree handling (`.git`-file early return, :621) is unchanged. The name `migrateToCommittedIndex` and the "committed index / enables worktree compatibility" header comments describe a commit-the-index design **this diff does not implement** — likely a leftover of an abandoned approach. Confirmed by Claude risk-architect (P1), ce-adversarial F5, ce-maintainability, facts-historian; Codex B1 raised it as its open question. **Ask**: what concrete worktree failure does this fix? If none, the rename — and its entire migration risk surface above — may be droppable; if real, link the issue and rename the function (`migrateLegacyMetaJson`) + fix the header comments.

## Inline-comment candidates

All verified in the worktree by synthesis; lines are diff-anchorable unless marked proxy.

| # | Path : line | Sev | Problem | Evidence / tag | Suggested fix | Corroboration |
|---|---|---|---|---|---|---|
| 1 | `gitnexus/src/core/run-analyze.ts:635` (proxy anchor — hook files not in diff) | P0 | Migration this line triggers deletes `meta.json` that all 3 shipped hooks still read → permanent false-stale nag; reintroduces #1438/#1137 | hooks/claude/gitnexus-hook.cjs:446, antigravity:430, plugin:466 verified; hook repro with commit event | Dual-read gitnexus.json→meta.json in all 3 hook copies + isGlobalRegistryDir sentinels; hook tests | Strong (Codex B1+B2+B3; Claude correctness/docs-dod/facts + coordinator) [reproduced] |
| 2 | `gitnexus/src/storage/repo-manager.ts:507` | P1 | One-way unlink of legacy meta.json → old MCP silently stale forever; rollback = silent full rebuild + dirty-flag loss | unlink verified :507; old-binary read path byte-identical pre-PR | Don't unlink in this PR; dual-write window, delete at a major | Strong (Codex B3+B1; Claude risk/correctness/adversarial) [code-read] |
| 3 | `gitnexus/src/storage/repo-manager.ts:499` | P1 | `fs.access` gate never reconciles dual-file state → fresher legacy meta.json (old-writer) permanently shadowed → false `alreadyUpToDate`, lost dirty flag | tsx repro: migrated=false, stale lastCommit wins, dirty flag lost | When both files exist, take `max(indexedAt)`, rewrite new, unlink legacy | Moderate (Claude adversarial+correctness; Codex B2 adjacent) [reproduced] |
| 4 | `gitnexus/src/storage/repo-manager.ts:424` | P1 | Blanket catch falls back to legacy on parse/EACCES/EIO → corrupt gitnexus.json masked by stale legacy baseline | catch verified; pre-PR corrupt→null→safe full rebuild | Fall back only via `isMissingFilesystemError` (:573); corrupt → null | Strong (Codex B2 P1+B1; Claude adversarial F4) [code-read] |
| 5 | `gitnexus/src/storage/repo-manager.ts:504` | P1 | Uncaught `saveMeta` in migration + unguarded call site → EROFS/ENOSPC crashes **every** analyze; migration should be best-effort | no try/catch :504 nor run-analyze.ts:635; loadMeta fallback makes write non-load-bearing | Wrap migration in try/catch, `logger.warn`, return false | Moderate (Claude reliability+testing) [code-read] |
| 6 | `gitnexus/src/storage/repo-manager.ts:490` | P1 | Destructive migration run on every analyze has **zero tests** | `grep -rn migrateToCommittedIndex test/` → nothing | Unit tests: happy flat/branch, idempotency, unlink-failure, corrupt legacy, saveMeta-throw contract | Strong (Claude test-ci+testing; Codex B2) [code-read] |
| 7 | `gitnexus/src/storage/repo-manager.ts:445` | P2 | Fixed `.tmp` name (pre-existing pattern) now gets a new concurrent writer — unconditional migration on every analyze: loser's rename ENOENT aborts analyze; partial-write window installs truncated JSON | `Promise.allSettled` repro `['REJECTED: ENOENT','ok']` | Unique tmp suffix (pid+timestamp) + unlink-on-failure; make migrate non-fatal | Weak (ce-adversarial) but [reproduced] — retained per rule |
| 8 | `gitnexus/src/storage/repo-manager.ts:545` | P2 | Branch-loop catch swallows real errors (stat/saveMeta) → silently, permanently skips remaining branches, no log | catch wraps whole loop, comment says "may not exist" only | Per-branch try/catch + `logger.warn` with branch name | Moderate (Claude reliability/correctness/adversarial) [code-read] |
| 9 | `gitnexus/src/storage/repo-manager.ts:5` | P1 (design) | Header (+`getStoragePaths` doc, + function name) claims worktree compatibility/committed index; `.gitnexus/` stays ignored — rename achieves nothing for the stated goal | ensureGitNexusIgnored/exclude-entry unchanged; no worktree-detection change in diff | State the real problem + link issue, or drop the rename; rename fn to `migrateLegacyMetaJson`, rewrite comments | Strong (Claude risk/adversarial/maintainability/facts; Codex B1 open question) [code-read] |
| 10 | `gitnexus/src/storage/repo-manager.ts:326` | P3 | Retained "#2106 byte-identical to pre-multi-branch" sentence is now false for `metaPath` (same in run-analyze.ts:370 R6 comment) | sentence verified at :326 in edited hunk | Scope byte-identity to storagePath/lbugPath; document the metaPath change | Weak (project-standards) [code-read] |
| 11 | `gitnexus/src/core/run-analyze.ts:391` | P3 | Inline re-derivation (×2, :391/:399) of `isMissingFilesystemError` introduced in the same PR | both copies verified | Export helper from repo-manager.ts and import | Weak (maintainability) [code-read] |
| 12 | `gitnexus/src/cli/remove.ts:110` | P3 | Pre-unlink is deterministic-orphan-prone (unlink ok + rm fails → validate-prune orphans branch DBs unreachable by `gitnexus remove`) and incomplete (legacy `meta.json` not unlinked, so the stated invariant fails on unmigrated repos) | remove.ts:109-118 verified; hasIndex accepts legacy | Also best-effort-unlink `meta.json`; or unregister after successful rm; no data-safety issue either way | Mixed — Codex B3 medium vs Claude risk-architect refutation; synthesis: real but recoverable → P3 [code-read] |
| 13 | `gitnexus/package-lock.json` (removed nested onnxruntime-common block) | P3 | Unrelated lockfile churn — no dep changed; npm-version artifact (same class as PR #2343) | package.json untouched | `git checkout 1029a8dd -- gitnexus/package-lock.json` | Strong (Codex B1+B2; Claude hygiene/standards/security) [code-read] |

On #12, synthesis weighed risk-architect's refutation (validate-prune self-heals the registry; re-analyze recovers) against Codex B3's "breaks retry invariant": both agree there is no data-safety issue; the residue is manual-cleanup-only orphaned branch dirs. P3, worth a comment, not blocking.

## Body-only findings (files not in the diff)

- **[P1] Unchanged tests will fail / test the wrong thing**: `run-analyze-fts-repair.test.ts:419` **fails now** [reproduced]; `cli-e2e.test.ts:393/440` read `meta.json` post-analyze; `antigravity-hook-e2e.test.ts` builds meta.json fixtures throughout. Update to `INDEX_METADATA_FILE`/`loadMeta`, keeping deliberate legacy fixtures only in fallback tests.
- **[P2] `AnalysisNotFinalizedError` message points at the wrong file** (`repo-manager.ts:1100` — *not* in the diff hunks): says "meta.json was not written to …/meta.json" while `assertAnalysisFinalized` (:1142) now checks gitnexus.json — the #1169 Windows silent-finalize diagnostic becomes misdirection. Use `INDEX_METADATA_FILE`. (Cross-engine: Codex B2 + 2 Claude lanes.)
- **[P2] Operator docs now point at a deleted file**: ARCHITECTURE.md:386 (storage diagram), GUARDRAILS.md:22/38/44 (incl. a "Non-negotiable" and the crash-recovery Signs), RUNBOOK.md:59. **MIGRATION.md has no entry** for a one-way format change (precedent: the OVERRIDES→METHOD_OVERRIDES entry). Per repo convention, fix in this PR — and do **not** touch CHANGELOG (owned at release time).
- **[P2] Test-effectiveness gaps around the exact changed paths**: the only integration test driving the local-backend staleness branch has vacuous assertions ("didn't crash"), and the `vi.mock('repo-manager.js')` blocks in `local-backend-calltool.test.ts` / `staleness-and-stability.test.ts` omit `loadMeta` → the PR's new call site throws TypeError, swallowed by the same catch, so the swap is effectively untested end-to-end. Add `loadMeta` to the mocks and a reinit spy assertion. Also: no direct unit test for `loadMeta`/`hasIndex` legacy fallback; old transient-error tests (lines 97–184) now exercise the legacy-fallback path while keeping primary-path labels.
- **[P3] Help/i18n drift**: `cli/index.ts:198`, `i18n/en.ts:203`, `i18n/zh-CN.ts:190` still say "meta.json" for `index --force` while the runtime message this PR edited says both files. Prefer file-name-agnostic wording.
- **[P3] Compact quality items**: `loadMetaLegacy` needlessly exported; local `let hasIndex` shadows the exported `hasIndex` (repo-manager.ts:1313); ad-hoc `as {lastCommit?; indexedAt?}` cast in group/service.ts:580 instead of `Pick<RepoMeta,…>`; group status now reports corrupt metadata as `indexStale` instead of `missing` (observable change, arguably better — confirm intent); one new `catch (err: any)` + two new `as any` test casts (follow file convention, but eslint `no-explicit-any` is codified — easy to avoid); RepoMeta inline comments at :125/:128/:141/:143 and assertAnalysisFinalized JSDoc (~:1128) still say meta.json; migration doubles `fs.access` traffic per registry entry on validate (watch on NFS-scale registries).

## Process / DoD gaps

- PR title lacks the mandatory conventional-commit prefix (the validating check skipped, so nothing enforced it); suggested: `fix(storage): rename index metadata to gitnexus.json with legacy migration` — pending the headline-5 answer.
- PR body is the verbatim unfilled template — Risk & rollout empty on a one-way-migration PR; no linked issue; no testing notes. Fill in-PR.
- Repo convention: propose and land fixes **in this PR** — no "follow-up later" deferrals.

---
*This is an automated multi-engine review digest (Claude swarm+ce, Codex GPT-5.5 swarm+ce+adversarial, coordinator verification); verify findings before acting on them.*

gitnexus/src/core/run-analyze.ts
```diff
@@ -614,11 +627,13 @@ export async function runFullAnalysis(
...
   const { lbugPath, metaPath } = getStoragePaths(repoPath, placement.branch);
-  // Directory that owns this run's meta.json (flat `.gitnexus` for the primary
-  // slot, `branches/<slug>/` otherwise). loadMeta/saveMeta operate on it so
-  // each branch keeps its own lastCommit / fileHashes / incremental dirty flag.
+  // metaPath now points to the metadata file (gitnexus.json) in a branch-specific directory.
+  // metaDir is the directory containing the metadata file (and branch-specific DBs).
   const metaDir = path.dirname(metaPath);
 
+  // Auto-migrate from legacy .gitnexus/meta.json to gitnexus.json.
+  await migrateToCommittedIndex(repoPath);
```
magyargergo: **[P0] [reproduced]** The migration this line triggers deletes `meta.json` — which all three shipped hooks still read directly (`gitnexus/hooks/claude/gitnexus-hook.cjs:446`, `gitnexus/hooks/antigravity/gitnexus-antigravity-hook.cjs:430`, `gitnexus-claude-plugin/hooks/gitnexus-hook.js:466`; none are in this diff, hence the anchor here).

After the first new-version analyze, every successful `git commit|merge|rebase|pull` fires "GitNexus index is stale (last indexed: never)" despite a fresh index — permanently — and `hadEmbeddings` is lost so the suggested command drops `--embeddings`. Reproduced by piping a PostToolUse commit event into the hook against a `gitnexus.json`-only repo. This reintroduces the exact failure mode of closed issues #1438 / #1137.

**Fix (in-PR):** dual-read `gitnexus.json` → `meta.json` fallback in all three hook copies (and their `isGlobalRegistryDir` meta.json sentinels), plus hook-test coverage for the new format. Corroboration: all 3 Codex legs + 3 Claude lanes + coordinator.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -442,6 +482,73 @@ export const loadRepo = async (repoPath: string): Promise<IndexedRepo | null> =>
...
+  try {
+    await fs.access(primaryMetadataPath);
+  } catch {
+    // Metadata file doesn't exist, check for legacy
+    const legacyMeta = await loadMetaLegacy(storagePath);
+    if (legacyMeta) {
+      await saveMeta(storagePath, legacyMeta);
+      logger.info(`Migrated primary index from ${primaryLegacyPath} to ${primaryMetadataPath}`);
+      try {
+        await fs.unlink(primaryLegacyPath);
```
magyargergo: **[P1] [code-read]** One-way delete of the legacy file breaks mixed-version and rollback scenarios:

- An **old MCP server still running** (editor-embedded, pinned Docker, long-lived `npx`) reads only `meta.json` in its staleness probe → ENOENT → `catch { return }` → silently serves the pre-analyze index until process restart.
- **Downgrade/rollback:** old `loadMeta` reads only `meta.json` → `null` → "no prior index" → silent full rebuild, losing incremental state and the `incrementalInProgress` crash-recovery flag.

**Fix:** don't unlink in this PR — keep or dual-write legacy for a compatibility window; remove at a future major. Corroboration: Codex B3(high)+B1; Claude risk-architect/correctness/adversarial.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -442,6 +482,73 @@ export const loadRepo = async (repoPath: string): Promise<IndexedRepo | null> =>
...
+export const migrateToCommittedIndex = async (repoPath: string): Promise<boolean> => {
+  const storagePath = getStoragePath(repoPath);
+  let migrated = false;
+
+  // Migrate primary/flat metadata
+  const primaryMetadataPath = path.join(storagePath, INDEX_METADATA_FILE);
+  const primaryLegacyPath = path.join(storagePath, 'meta.json');
+
+  try {
+    await fs.access(primaryMetadataPath);
```
magyargergo: **[P1] [reproduced]** This `fs.access` gate never reconciles a dual-file state: once `gitnexus.json` exists, migration never looks at legacy again, and `loadMeta` blindly prefers the new file. If an older binary re-analyzes afterwards (writing a *fresher* `meta.json`, possibly carrying the dirty flag), the new version permanently trusts the **stale** `gitnexus.json` → false `alreadyUpToDate` certification of a DB built at a different commit, skipped crash recovery. Reproduced against this module: `migrated=false`, stale `lastCommit` wins, dirty flag lost.

**Fix:** when both files exist, reconcile by `max(indexedAt)` (rewrite new, then unlink legacy).

gitnexus/src/storage/repo-manager.ts
```diff
@@ -395,40 +410,65 @@ export const loadMeta = async (storagePath: string): Promise<RepoMeta | null> =>
...
+ * For feature branches: metaDir = <repo>/.gitnexus/branches/<slug>
+ * Falls back to legacy meta.json if gitnexus.json doesn't exist.
+ */
+export const loadMeta = async (metaDir: string): Promise<RepoMeta | null> => {
+  // Try new metadata file first
+  try {
+    const metaPath = path.join(metaDir, INDEX_METADATA_FILE);
+    const raw = await fs.readFile(metaPath, 'utf-8');
+    return JSON.parse(raw) as RepoMeta;
+  } catch {
```
magyargergo, line 442: **[P1] [code-read]** This blanket `catch` falls back to legacy on **parse errors and EACCES/EIO**, not just missing-file. Pre-PR, a corrupt meta → `null` → forced full rebuild (the fail-safe the `saveMeta` docstring relies on). Now a corrupt/unreadable `gitnexus.json` + a lingering `meta.json` (one failed unlink suffices — and the unlink is warn-and-continue with no retry) → incremental run over a stale `fileHashes` baseline with no dirty flag.

**Fix:** fall back only on ENOENT/ENOTDIR — `isMissingFilesystemError` already exists in this file (:573); corrupt new file should return `null` (safe rebuild). Corroboration: Codex B2(P1)+B1; Claude adversarial.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -442,6 +482,73 @@ export const loadRepo = async (repoPath: string): Promise<IndexedRepo | null> =>
...
+  const primaryMetadataPath = path.join(storagePath, INDEX_METADATA_FILE);
+  const primaryLegacyPath = path.join(storagePath, 'meta.json');
+
+  try {
+    await fs.access(primaryMetadataPath);
+  } catch {
+    // Metadata file doesn't exist, check for legacy
+    const legacyMeta = await loadMetaLegacy(storagePath);
+    if (legacyMeta) {
+      await saveMeta(storagePath, legacyMeta);
```
magyargergo: **[P1] [code-read]** This `saveMeta` is uncaught, and the call site (`run-analyze.ts:635`) has no try/catch either: EROFS/EACCES/ENOSPC on `.gitnexus/` → **every** analyze aborts, repeatedly, for a repo that worked read-only pre-PR. Since `loadMeta` already has the legacy fallback, this migration write is non-load-bearing — its failure should never be fatal. This file already has the `isReadOnlyFilesystemError` + warn pattern to copy.

**Fix:** wrap the migration body in try/catch → `logger.warn` → return false.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -442,6 +482,73 @@ export const loadRepo = async (repoPath: string): Promise<IndexedRepo | null> =>
   };
 };
 
+/**
+ * Migrate metadata from legacy .gitnexus/meta.json to gitnexus.json.
+ * Returns true if migration occurred, false if not needed.
+ * Migrates both primary and branch-specific metadata.
+ */
+export const migrateToCommittedIndex = async (repoPath: string): Promise<boolean> => {
```
magyargergo: **[P1] [code-read]** `migrateToCommittedIndex` unlinks files unconditionally on every `gitnexus analyze`, and has **zero tests** (`grep -rn migrateToCommittedIndex test/` → nothing).

**Fix:** unit tests for: flat + branch happy paths, idempotency (gitnexus.json already present), unlink-failure warn path, corrupt legacy, saveMeta-throw contract. Corroboration: Claude test-ci + ce-testing; Codex B2.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -395,40 +410,65 @@ export const loadMeta = async (storagePath: string): Promise<RepoMeta | null> =>
...
  * on `node:fs/promises` uses `MoveFileEx(REPLACE_EXISTING)`), so either
  * the old or the new file is observed at every moment.
  */
-export const saveMeta = async (storagePath: string, meta: RepoMeta): Promise<void> => {
-  await fs.mkdir(storagePath, { recursive: true });
-  const metaPath = path.join(storagePath, 'meta.json');
+export const saveMeta = async (metaDir: string, meta: RepoMeta): Promise<void> => {
+  await fs.mkdir(metaDir, { recursive: true });
+  const metaPath = path.join(metaDir, INDEX_METADATA_FILE);
   const tmpPath = `${metaPath}.tmp`;
```
magyargergo: **[P2] [reproduced]** Fixed `.tmp` name (pre-existing pattern, but the unconditional migration adds a new concurrent writer on every analyze — CI matrix, auto-sync #2329): with two concurrent writers, the loser's `rename` throws raw ENOENT and kills its analyze; interleavings can momentarily install truncated JSON (feeding the corrupt-file fallback issue above). Repro: `Promise.allSettled` of two `saveMeta` → `['REJECTED: ENOENT', 'ok']`.

**Fix:** unique tmp suffix (pid+timestamp) + unlink-on-failure; make the migrate call non-fatal.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -442,6 +482,73 @@ export const loadRepo = async (repoPath: string): Promise<IndexedRepo | null> =>
...
+            await fs.unlink(branchLegacyPath);
+            logger.info(`Cleaned up legacy ${branchLegacyPath}`);
+          } catch (err) {
+            logger.warn({ err }, 'Failed to clean up legacy branch meta.json (non-critical)');
+          }
+          migrated = true;
+        }
+      }
+    }
+  } catch {
```
magyargergo: **[P2] [code-read]** This catch was written for "branchesDir may not exist" but wraps the whole per-branch loop — a single bad branch dir (dangling symlink → `stat` throws, EACCES) silently and permanently aborts migration for every branch after it (stable readdir order), with zero logging.

**Fix:** per-branch try/catch + `logger.warn` naming the branch; keep the outer catch for the missing-dir case only.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -1,9 +1,14 @@
 /**
  * Repository Manager
  *
- * Manages GitNexus index storage in .gitnexus/ at repo root.
- * Also maintains a global registry at ~/.gitnexus/registry.json
- * so the MCP server can discover indexed repos from any cwd.
+ * Manages GitNexus index storage:
+ * - Per-worktree metadata files (gitnexus.json) under .gitnexus/ for worktree compatibility
```
magyargergo: **[P1 design question] [code-read]** The header (and `getStoragePaths` docs, and the name `migrateToCommittedIndex`) claim worktree compatibility / a committed index — but `.gitnexus/` remains fully git-ignored (`ensureGitNexusIgnored` / `GITNEXUS_EXCLUDE_ENTRY` untouched, blanket `*` self-.gitignore intact), so renaming a file inside it is invisible to git, and each worktree already had its own independent `.gitnexus/` by construction. Nothing in this diff changes git-worktree behavior.

**Ask:** what concrete worktree failure does this fix? If none, the rename — and the entire migration risk surface — may be droppable. If real, please link the issue, rename the function (e.g. `migrateLegacyMetaJson`), and rewrite these comments. Corroboration: Claude risk/adversarial/maintainability/facts; Codex B1 raised the same open question.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -314,17 +321,25 @@ export const getStoragePath = (repoPath: string): string => {
  *
  * `storagePath` is ALWAYS the flat `<repo>/.gitnexus` — content-addressed
  * caches (`parse-cache/`, `parsedfile-store/`) live there and are shared
- * across branches (#2106 KTD7). When `branch` is provided, only `lbugPath` and
- * `metaPath` are scoped under `branches/<slug>/`; the flat call (no `branch`)
+ * across branches (#2106 KTD7). When `branch` is provided, both `lbugPath`
+ * and `metaPath` are scoped under `branches/<slug>/`; the flat call (no `branch`)
  * returns byte-identical paths to the pre-multi-branch behavior.
```
magyargergo: **[P3] [code-read]** The retained "#2106 … byte-identical" wording is now false for `metaPath` (this PR changes it from `meta.json` to `gitnexus.json`); same for the R6 comment at `run-analyze.ts:370`. Scope the byte-identity claim to `storagePath`/`lbugPath` and document the metaPath change.

gitnexus/src/core/run-analyze.ts
```diff
@@ -384,9 +386,20 @@ export const collectBranchCacheKeys = async (
     if (excludeDir && path.resolve(dir) === path.resolve(excludeDir)) continue;
     let raw: string;
     try {
-      raw = await fs.readFile(path.join(dir, 'meta.json'), 'utf-8');
-    } catch {
-      continue; // no meta here — not a branch index, not a failure
+      raw = await fs.readFile(path.join(dir, INDEX_METADATA_FILE), 'utf-8');
+    } catch (newErr) {
+      const newCode = (newErr as NodeJS.ErrnoException)?.code;
```
magyargergo: **[P3] [code-read]** This ENOENT/ENOTDIR check (here and at :399) re-derives `isMissingFilesystemError`, which this same PR adds as a private helper in `repo-manager.ts:573`. Export it and import here — three hand-rolled copies of the fallback predicate will drift.

gitnexus/src/cli/remove.ts
```diff
@@ -101,7 +106,17 @@ export const removeCommand = async (target: string, options?: { force?: boolean
   // orphaned — `listRegisteredRepos({ validate: true })` prunes those on
   // next read, so the failure is self-healing.
   try {
+    // Delete the metadata file first so a partial rm cannot leave a valid-looking index.
+    const metadataPath = path.join(entry.storagePath, INDEX_METADATA_FILE);
```
magyargergo: **[P3] [code-read]** Two issues with the pre-unlink: (a) it doesn't cover legacy `meta.json` or branch metadata, so the stated invariant ("partial rm cannot leave a valid-looking index") fails on unmigrated repos (`hasIndex` accepts legacy); (b) if the unlink succeeds and `fs.rm` then fails, `validate:true` prunes the registry entry while `lbug/` remains on disk — orphaned data no longer reachable via `gitnexus remove` (manual cleanup only; no data-safety issue). Either also best-effort-unlink `meta.json`, or drop the pre-unlink (fs.rm removes it anyway) and unregister only after a successful rm.

gitnexus/package-lock.json
```diff
@@ -4222,12 +4222,6 @@
         "onnxruntime-common": "1.27.0"
```
magyargergo, line 4222: **[P3] [code-read]** Unrelated lockfile churn: the nested `onnxruntime-node/node_modules/onnxruntime-common@1.26.0` block is dropped with no `package.json` change — an npm-version regeneration artifact (same class as the PR #2343 finding; the dedup to top-level 1.27.0 is semantically fine but doesn't belong in this PR). `git checkout <base> -- gitnexus/package-lock.json` unless intentional.

Issue #1741, opened 2026-05-21
GitNexus analyze gets stuck at 49% during “Resolving calls”.


miracleAi-new:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

I did "gitnexus analyze --skip-agents-md --worker-timeout 60 --verbose --force" on a project and after coming to 49% | Project structure analyzed + 49% |  Resolving calls (all chunks)... (9000/18066 files), it gets stuck on that no matter how long I await (I have wated for 24h). The process isn't even failing also,the CPU usage is still high。

### Expected behavior

The codebase indexing should get completed with gitnexus analyze getting to 100%

### Actual behavior

stuck on 49% | Resolving calls (all chunks)... (9000/18066 files)

### Steps to reproduce

gitnexus analyze --skip-agents-md --worker-timeout 60 --verbose --force

### Environment

- OS: macOs 14.7.8
- Node:v22.22.2

### Logs / screenshots

<img width="2104" height="210" alt="Image" src="https://github.com/user-attachments/assets/c153071e-9abe-410b-97d9-c77af7d1d052" />

sanguine59:
Hello, which version of gitnexus are you using? are the codebase size relatively large? there already an issue similar to this one at #1726

sanguine59:
perhaps it has something to do with the call processor ingestion... when indexing a files with large call site counts or overall large codebase

sanguine59:
@miracleAi-new , are you working with protobuf (gRPC) or maybe OpenAPI clients? they could be one of the cause. Either way, the ingestion might need some changes to handle this kind of issue

magyargergo:
Please see my comment at https://github.com/abhigyanpatwari/GitNexus/issues/1726#issuecomment-4505146873

miracleAi-new:
> 您好，请问您使用的是哪个版本的 GitNexus？代码库规模比较大吗？之前已经有一个类似的问题，编号为[#1726。](https://github.com/abhigyanpatwari/GitNexus/issues/1726)

I’m using GitNexus v1.6.4, and yes—our project is relatively large.

miracleAi-new:
> [@miracleAi-new](https://github.com/miracleAi-new)您是否在使用 protobuf（gRPC）或 OpenAPI 客户端？它们可能是问题的原因之一。无论如何，数据摄取流程可能需要进行一些更改才能处理此类问题。

[...]

Issue #1549, opened 2026-05-13
Documented Docker workflow (`gitnexus index /workspace/<repo>` on `:ro` mount) is broken in two places


Avicennasis:
## Summary

The README's documented "Make repos on your host machine indexable" Docker workflow is broken in two distinct ways. Both bugs together mean that the literal command shown in the README cannot be executed end-to-end against the published image.

## Reproduction (deterministic, 100% from clean state)

```bash
# Clean slate
rm -rf .gitnexus ~/.gitnexus
docker compose down -v

# Step 1 of the documented workflow (host-side analyze)
npx -y gitnexus@1.6.5-rc.19 analyze
# → succeeds, creates ./.gitnexus/ + ~/.gitnexus/registry.json

# Step 2 of the documented workflow (bring up stack with read-only workspace)
WORKSPACE_DIR="$PWD" \
  SERVER_IMAGE=ghcr.io/abhigyanpatwari/gitnexus:1.6.5-rc.19 \
  docker compose up -d
docker inspect gitnexus-server --format \
  '{{range .Mounts}}{{if eq .Destination "/workspace"}}{{.Destination}}: mode={{.Mode}} RW={{.RW}}{{end}}{{end}}'
# → /workspace: mode=ro RW=false   (matches .env.example "Optional read-only mount")

# Step 3 — the README's literal command:
docker compose exec gitnexus-server gitnexus index /workspace
# → exec: "gitnexus": executable file not found in $PATH

# Workaround for the PATH bug:
docker compose exec gitnexus-server node /app/gitnexus/dist/cli/index.js index /workspace
# → Error: EROFS: read-only file system,
#   open '/workspace/.gitnexus/.gitignore'
#     at async ensureGitNexusIgnored (file:///app/gitnexus/dist/storage/repo-manager.js:209:5)
#     at async indexCommand (file:///app/gitnexus/dist/cli/index-repo.js:99:5)
```

Same outcome on the latest stable image (`ghcr.io/abhigyanpatwari/gitnexus:latest` = v1.6.4, built 2026-05-10).

## Bug A — `gitnexus` is not on `PATH` inside the published image

The README documents:

```
docker compose exec gitnexus-server gitnexus index /workspace/my-repo
```

But `which gitnexus` inside the container is empty:

```bash
$ docker compose exec gitnexus-server sh -c 'echo $PATH; which gitnexus || echo MISSING'
[...]

PR #2364, opened 2026-07-03
feat: flat workspace index follows the checked-out branch

### LINKED ISSUE #2354: feat(cli): add a flag or setting to completely ignore Git branches and index solely based on local workspace disk state
### Area

### Problem or opportunity

When switching branches in a local repository, the GitNexus CLI throws a slot-ownership conflict error (e.g., "main" is not the primary index — "html-agent" owns the flat slot. Run gitnexus clean...).

For developers who frequently bounce between branches or prefer GitNexus to act as a purely branch-agnostic workspace analyzer, this introduces significant friction. It forces manual gitnexus clean wipes or requires managing explicit --branch flags. There is currently no out-of-the-box configuration or flag to tell GitNexus: "Completely ignore Git branches and just index whatever raw code is currently sitting on my local disk."

### Proposed solution

Introduce a way to completely bypass Git branch-matching logic in the analyzer pipeline.

1. Config Option: Support an ignoreBranches: true setting inside gitnexus.json.

2. CLI Flag: Provide a --no-branch or --disk-only flag on the analyzer (e.g., npx gitnexus analyze --no-branch).

User-visible behavior:
When this mode is active, gitnexus analyze will map everything to a single unified, static default slot. Switching branches locally via Git will not trigger a primary slot collision error. Instead, running npx gitnexus analyze will seamlessly perform a fast incremental update (changed=X, added=Y) matching whatever files currently exist in the workspace, completely oblivious to Git's current HEAD branch name.

### Alternatives considered

Manually running npx gitnexus clean every time: This kills developer velocity, destroys unchanged file cache rows unnecessarily, and increases indexing times for large repositories.

Using --branch <name> for every single branch: This fragments the local .gitnexus/ DB directory with multiple branch indices, which takes up unnecessary space when the developer only cares about the AI agent seeing their active screen context.

### Acceptance criteria

[ ] The analyzer respects a new CLI flag (e.g., --no-branch) or a config property (settings.ignoreBranches) in gitnexus.json.

[ ] When enabled, GitNexus does not check the current Git branch name or validate it against the current primary index slot owner.

[ ] Re-running npx gitnexus analyze after switching branches successfully performs an incremental file-delta update on a single default slot without throwing slot-ownership validation errors.

[ ] Appropriate unit tests added to the core analyzer verifying branch-agnostic indexing behavior.

### Constraints

Must not break the existing behavior for users who rely on explicit branch indexing (--branch).

Must safely fallback to a generic slot identifier (e.g., "default" or "workspace") in the local LadybugDB metadata.

### Contribution

- [x] I am willing to open a PR for this (may need design discussion first).

magyargergo:
## Summary

A plain `gitnexus analyze` now always updates the flat workspace index, which follows the checked-out working tree — switching branches performs a fast incremental update instead of auto-routing into `branches/<slug>/` sub-indexes or nagging with the primary-inversion "run `gitnexus clean`" warning. Per-branch sub-indexes remain available, but only via explicit `--branch`. No new CLI flag or config key: the smart behavior is the default.

## Motivation / context

Closes #2354. Developers who bounce between branches hit slot-ownership friction: the #2106 model pinned the flat slot to the first-indexed branch, so plain re-analyzes on other branches silently fragmented storage into per-branch indexes (disk bloat) and the R8 warning told users to wipe with `gitnexus clean`. #2358 and #2359 both solve this with a new `--no-branch`/`--disk-only` flag; this PR supersedes them by making the requested behavior the default instead of adding a third knob.

## Areas touched

## Scope & constraints

**In scope**

- Placement: only explicit `--branch` consults `resolveBranchPlacement`; a plain analyze always targets the flat slot. `meta.branch` becomes an informational "last analyzed branch" label, restamped each run. Cross-branch incremental is safe because the diff is pure content-hash based (`diffFileHashes`): deleted files are dropped, and the importer BFS reads the pre-run DB — exactly the rows to replace.
- Fast path: a same-commit clean-tree branch flip restamps the label and registry entry via a new narrow `adoptFlatBranchLabel` helper (no-op for unregistered repos, preserving the #2264/#1169 no-self-heal rule; `registerRepo` is unsafe there — it self-heals missing entries, can throw `RegistryNameCollisionError`, and shells out `git config`).
- Shadowed sub-index cleanup: when the flat slot adopts a label that has a pinned sub-index, that sub-index becomes permanently unreachable (query scoping serves the flat handle first; explicit `--branch` re-resolves to flat) — the directory and its registry summary are removed together, so `clean --branch` never sees an un-cleanable orphan.
- MCP: `applyBranchScope` now always falls back to the on-disk flat meta before throwing "not indexed", so a long-lived server resolves a freshly restamped workspace branch without a registry refresh (cached handles have no repo-resolution miss to refresh them after a flip).
- `status`: the "current branch not indexed" dead end is replaced by falling through to the workspace index with an informational line plus the usual commit-based staleness verdict (same-commit flip → up-to-date; different commit → stale).
- Deleted `primaryInversionWarning` (meaningless under follows-HEAD semantics); help text, MCP tool `branch` param description, i18n (en + zh-CN), and README one-liners updated.

- Explicit `--branch` pinning semantics, the checkout-mismatch guard, detached-HEAD/CI flat routing, and `clean --branch` are unchanged.
- No migration of pre-existing auto-created `branches/<slug>/` dirs — they stay queryable via `--branch` and cleanable via `clean --branch`.

## Implementation notes

Side benefit: this fixes the hooks' perpetual-staleness blind spot — the PostToolUse hooks compare the flat `meta.lastCommit` against HEAD, but the analyze they nagged for previously wrote a sub-index and never satisfied the flat comparison; now a plain analyze updates the flat slot the hooks read.

## Testing & verification

- [x] `cd gitnexus && npx vitest run test/unit/run-analyze.test.ts test/unit/repo-manager.test.ts test/unit/calltool-dispatch.test.ts test/unit/list-status-branch.test.ts test/unit/cli-i18n.test.ts test/unit/cli-index-help.test.ts test/integration/multi-branch-analyze.test.ts`
- [x] Manual CLI E2E (`node dist/cli/index.js` in a scratch two-branch repo, isolated `GITNEXUS_HOME`): plain analyze after checkout runs `Incremental: changed=0, added=3` into the flat slot with no `branches/` dir and no warning; `status` shows the new workspace line + stale on the other branch; `--branch` mismatch guard still refuses; pin → plain analyze reclaims the shadowed sub-index (dir and registry summary removed, empty `branches/` parent tidied); same-commit flip prints "Already up to date" and still restamps the label; `clean --branch nope` errors cleanly.

New/updated tests: workspace adoption on the fast path (restamp + shadow removal), explicit-pin `isPrimaryBranch:false`, `adoptFlatBranchLabel` (relabel, keep-other-pins, no-self-heal), MCP stale-handle fallback, integration round-trip (plain follow, pin, reclaim), and `status` fall-through.

## Risk & rollout

Semantic change to plain `analyze` on a non-owner branch: it now updates the flat index in place instead of writing a sub-index — the previous branch's flat content is not preserved unless pinned with `--branch`. Worth a release-note line. Backward compatible otherwise; no index rebuild required (the first plain analyze after upgrade just adopts the current branch incrementally).

## Checklist

## Review follow-ups (tri-review 4627328889)

> Rebased onto the branch tip after `main` (incl. #2363 gitnexus.json dual-write) was merged in — zero conflicts; all touched suites re-validated post-rebase (373 tests + tsc).

All four findings, the 7 listed test gaps, and the P3 notes are fixed — one commit per finding, plus five hardening commits from a post-fix review pass over the fix series itself.

**Finding → commit map**

| Finding | Commit | Fix |
|---|---|---|
| F1 (P1) stale cached handle, both arms | `6bf2cb98` | flat meta read before any cached-state trust; summary arm verifies the sub-index lbug on disk; no-readable-meta legacy trust preserved (#2106 R4); one `refreshRepos()` on staleness. Gaps 1–2 + a dangling-summary degraded-window regression |
| F2 (P2) rm before registry no-op check | `3003271f` | registry membership gate precedes the rm; no-self-heal now covers disk. Gap 3; the two bug-encoding tests re-anchored (registered fixture under isolated `GITNEXUS_HOME` + unregistered-survives variant) |
| F3 (P2) saveMeta-before-adopt + `:ro` regression | `82f046d7` | adopt-first / saveMeta-last (partial failure keeps the retry guard true), whole block best-effort with `isReadOnlyFilesystemError` (#1549). Gaps 4, 6, 7 in a new dedicated failure-mode test file |
| F4 (P2) rm failure swallowed, summary dropped | `f46e90e2` | summary dropped only when the dir is verifiably gone; warn with path+errno; later adopts retry. Gap 5 |
| P3 end-of-run adopt not best-effort | `e0d3e4b9` | warn-and-continue mirroring the parse-cache wrapper; integration test |
| P3 error guidance | `69e8f717` | leads with the checkout; `(primary only)` → `(workspace only)` |
| P3 vocab drift | `d91c210a` | flagged comments/JSDoc/test names only; no API renames |
| P3 `workspaceIndexLabel` wording | `bd25f87a` | en + zh-CN: "re-run gitnexus analyze to follow the current branch" |

**Hardening from the post-fix review pass** (self-review of the fix series, 8 finder angles)

- `ee64d2aa` — re-read the registry after the recursive rm (the F2 reorder had widened the #2106 R9 lost-update window across a multi-second delete)
- `3b62b78c` — both new existence probes treat only ENOENT/ENOTDIR as gone (an EACCES/EIO would have re-created F4's stranded-bloat / produced false not-indexed errors)
- `e8e4970b` — `applyBranchScope` coherence: truthful indexed-branches list (no more "not indexed: main / indexed: main"), refresh-once on every miss (newly-pinned branches heal on the next call), the rm-degraded duplicate-claim state errors honestly instead of serving a stale-vintage pin, and the meta that decides routing now also supplies the handle metadata
- `fc274488` — restamp warnings keep the real errno next to the #1549 hint; end-of-run retry-semantics comment corrected (retries on the next content-changing run; interim state is label-cosmetic only, never wrong routing)
- `87f1bb35` — test hygiene: mkdtemp-unique branch-scope fixture root (fixed shared tmpdir path could flake concurrent runs); dropped redundant dynamic imports

**Deferred (unchanged from the review's own notes):** registry-mutation locking / lost-update class retirement, canonicalize→findIndex + containment-guard helper extraction, refresh debounce for persistently-stale registries, test-helper consolidation (`GITNEXUS_HOME` isolation harness, `metaFor`/seed fixtures), the pre-existing hardcoded-English `analyze --branch` help description, `as any` cleanup in pre-existing test code. PR #2363 sequencing stays "this PR first, rebase #2363".

**Validation:** suite green per commit (targeted) + `tsc --noEmit` per commit (pre-commit hook enforced); full `npm test` from `gitnexus/`, with the 34 load-flaked files (worker-startup starvation under parallel review agents, the #1741 signature) all re-verified green in isolated batches; root `prettier --check` and `eslint` (0 errors) on every touched file.

- **Log queries:** `could not restamp the workspace branch label` and `could not sync the workspace branch label` (analyze warnings — expected only on RO mounts/locks, self-healing); `Could not remove the shadowed branch sub-index` (Windows EBUSY — summary retained by design, `clean --branch` remains able to target it).
- **Healthy signals:** no `Branch "X" is not indexed ... Indexed branches: X`-shaped self-contradictions in MCP error logs; `gitnexus list` branch labels match `status` after branch flips.
- **Failure signal / rollback trigger:** branch-scoped MCP queries returning wrong-branch content after a checkout flip (the F1 class) — revert to the pre-series head `c16cb7ea` behavior is NOT a mitigation (it contains the bugs); revert the whole PR instead.
- **Window/owner:** normal release soak; no infra changes.

### DIFF
--- a/gitnexus/src/core/run-analyze.ts
+++ b/gitnexus/src/core/run-analyze.ts
@@ -54,6 +54,7 @@ import {
   ensureGitNexusIgnored,
   registerRepo,
   adoptFlatBranchLabel,
+  isReadOnlyFilesystemError,
   isRepoRegistered,
   cleanupOldKuzuFiles,
   reconcileMetadataFiles,
@@ -862,8 +863,27 @@ export async function runFullAnalysis(
         // Detached HEAD / non-git (branchLabel === null) keeps the existing
         // stamp, mirroring the end-of-run meta write.
         if (!placement.branch && branchLabel && existingMeta.branch !== branchLabel) {
-          await saveMeta(metaDir, { ...existingMeta, branch: branchLabel });
-          await adoptFlatBranchLabel(repoPath, branchLabel);
+          // Adopt first, stamp last (#2364 review F3): this block's retry
+          // guard is `existingMeta.branch !== branchLabel`, so stamping the
+          // meta before the registry/shadow cleanup would flip the guard and
+          // lock in any partial failure — with saveMeta last, a failed adopt
+          // leaves the guard true and the next same-commit run self-heals
+          // (adopt is idempotent). The whole sync is best-effort: the label
+          // is informational and the flat DB content is byte-valid for both
+          // labels here (same commit, clean tree), so an "Already up to
+          // date" run must not fail over it; read-only storage — the
+          // documented Docker :ro workflow (#1549) — degrades to a warning.
+          try {
+            await adoptFlatBranchLabel(repoPath, branchLabel);
+            await saveMeta(metaDir, { ...existingMeta, branch: branchLabel });
+          } catch (err) {
+            const reason = isReadOnlyFilesystemError(err)
+              ? 'GitNexus storage is read-only (#1549)'
+              : (err as Error).message;
+            log(
+              `Warning: could not restamp the workspace branch label (${reason}); will retry on the next run.`,
+            );
+          }
         }
         await ensureGitNexusIgnored(repoPath);
         return {
--- a/gitnexus/src/core/run-analyze.ts
+++ b/gitnexus/src/core/run-analyze.ts
@@ -879,8 +879,11 @@ export async function runFullAnalysis(
             await adoptFlatBranchLabel(repoPath, branchLabel);
             await saveMeta(metaDir, { ...existingMeta, branch: branchLabel });
           } catch (err) {
+            // EACCES/EPERM also arise from ownership problems and transient
+            // Windows locks, so keep the real error visible alongside the
+            // #1549 read-only hint instead of replacing it.
             const reason = isReadOnlyFilesystemError(err)
-              ? 'GitNexus storage is read-only (#1549)'
+              ? `${(err as Error).message} — storage may be read-only (#1549)`
               : (err as Error).message;
             log(
               `Warning: could not restamp the workspace branch label (${reason}); will retry on the next run.`,
@@ -1623,9 +1626,12 @@ export async function runFullAnalysis(
     // Drop a now-shadowed `branches/<slug>/` sub-index for the same label
     // (unreachable once the flat slot serves it) and align the registry's
     // top-level branch label. Best-effort like the parse-cache save above
-    // (#2364 review F5): the index is already complete and registered, and
-    // adopt retries unconditionally on the next plain analyze — a registry
-    // write failure must not fail an otherwise successful multi-minute run.
+    // (#2364 review F5): the index is complete and registered, and a failure
+    // here leaves only a stale registry label / undeleted shadowed dir —
+    // never wrong routing, because the flat meta this run already stamped is
+    // what applyBranchScope trusts. Retried by the next content-changing run
+    // (same-commit fast-path runs skip it: their guard compares the
+    // already-stamped meta label).
     if (!placement.branch && branchLabel) {
       try {
         await adoptFlatBranchLabel(repoPath, branchLabel);
--- a/gitnexus/src/core/run-analyze.ts
+++ b/gitnexus/src/core/run-analyze.ts
@@ -283,10 +283,12 @@ export interface AnalyzeResult {
    */
   ftsSkipped?: boolean;
   /**
-   * True when the index this run produced/validated is the primary/flat slot
-   * (#2106 R2). `false` for a non-primary branch index. Lets the CLI skip
-   * repo-root AGENTS.md/CLAUDE.md refreshes (e.g. the base_ref fast-path) for a
+   * True when the index this run produced/validated is the flat workspace
+   * slot (#2106 R2, inverted by #2354 to follow the checked-out branch).
+   * `false` for a pinned `--branch` sub-index. Lets the CLI skip repo-root
+   * AGENTS.md/CLAUDE.md refreshes (e.g. the base_ref fast-path) for a pinned
    * branch analyze, mirroring the in-pipeline `if (!placement.branch)` gate.
+   * (The historical "primary" name is kept — it is public API surface.)
    */
   isPrimaryBranch?: boolean;
 }
--- a/gitnexus/src/core/run-analyze.ts
+++ b/gitnexus/src/core/run-analyze.ts
@@ -53,6 +53,7 @@ import {
   loadMeta,
   ensureGitNexusIgnored,
   registerRepo,
+  adoptFlatBranchLabel,
   isRepoRegistered,
   cleanupOldKuzuFiles,
   INCREMENTAL_SCHEMA_VERSION,
@@ -94,7 +95,6 @@ import {
 import {
   getCurrentCommit,
   getCurrentBranch,
-  getDefaultBranch,
   getRemoteUrl,
   hasGitDir,
   getInferredRepoName,
@@ -205,13 +205,13 @@ export interface AnalyzeOptions {
    */
   defaultBranch?: string;
   /**
-   * Index-branch selector (#2106). Distinct from `defaultBranch` (which only
-   * affects generated AGENTS.md/CLAUDE.md base_ref text). When set, this run is
-   * labelled as that branch and routed to a per-branch index slot unless it is
-   * the primary branch. When `undefined`, the branch is auto-detected from the
-   * checked-out HEAD (the flat/primary slot for the first-indexed branch, a
-   * `branches/<slug>/` sub-directory otherwise). Detached HEAD / non-git always
-   * maps to the flat slot.
+   * Index-branch selector (#2106, #2354). Distinct from `defaultBranch` (which
+   * only affects generated AGENTS.md/CLAUDE.md base_ref text). When set, this
+   * run is pinned to a per-branch index slot (`branches/<slug>/`) unless the
+   * label matches the flat slot's recorded branch. When `undefined`, the run
+   * always targets the flat workspace slot, which follows the checked-out
+   * working tree; the auto-detected branch is only recorded as the slot's
+   * informational label. Detached HEAD / non-git also map to the flat slot.
    */
   branch?: string;
   /**
@@ -340,27 +340,6 @@ export const PHASE_LABELS: Record<string, string> = {
  * the {@link AnalyzeCallbacks} interface — it never writes to stdout/stderr
  * directly and never calls `process.exit()`.
  */
-/**
- * Build the primary-inversion warning (#2106 R8), or `undefined` when there is
- * nothing to warn about. Pure + exported for testing. Both inputs are trimmed
- * (a diagnostic — a missed warning is low-harm; a false warning is the thing to
- * avoid). `defaultBranch` is the repo's `origin/HEAD` branch (null when unset,
- * e.g. fresh clones / CI), `flatOwner` is the branch that owns the flat slot.
- */
-export const primaryInversionWarning = (
-  defaultBranch: string | null | undefined,
-  flatOwner: string | null | undefined,
-): string | undefined => {
-  const norm = (s: string | null | undefined): string | undefined => s?.trim() || undefined;
-  const d = norm(defaultBranch);
-  const o = norm(flatOwner);
-  if (!d || !o || d === o) return undefined;
-  return (
-    `Warning: the default branch "${d}" is not the primary index — "${o}" owns the flat slot. ` +
-    `Run \`gitnexus clean --branch ${o}\` then re-index on "${d}", or query it explicitly with \`--branch ${d}\`.`
-  );
-};
-
 /**
  * Collect the recorded parse-cache chunk keys across the flat + every branch
  * meta under a flat `.gitnexus` storage, EXCLUDING `excludeDir` (the current
@@ -585,13 +564,15 @@ export async function runFullAnalysis(
   const repoHasGit = hasGitDir(repoPath);
   const currentCommit = repoHasGit ? getCurrentCommit(repoPath) : '';
 
-  // ── #2106: resolve which branch slot this run writes to ───────────────
+  // ── #2106/#2354: resolve which branch slot this run writes to ─────────
   // `branchLabel` is the branch identity recorded in meta.json (incl. the
-  // primary). `placement.branch` is undefined for the flat/primary slot (the
-  // lbug/meta paths stay byte-identical to single-branch behavior) and set for
-  // a `branches/<slug>/` sub-directory. Explicit `--branch` is always honored;
-  // otherwise auto-detect the checked-out branch (null for detached HEAD /
-  // non-git → flat slot).
+  // flat workspace slot). `placement.branch` is undefined for the flat slot
+  // (the lbug/meta paths stay byte-identical to single-branch behavior) and
+  // set for a `branches/<slug>/` sub-directory. Only an explicit `--branch`
+  // can route to a sub-directory; a plain analyze ALWAYS targets the flat
+  // slot, which follows the checked-out working tree (#2354) — the
+  // auto-detected branch (null for detached HEAD / non-git) is recorded as
+  // the slot's informational label only.
   // Normalize the auto-detected branch the same way an explicit `--branch` is
   // validated (#2106 R1): a git ref the branch-name rules forbid (backtick,
   // `~ ^ : ? *`, leading `-`, `..`) becomes `null` → the flat slot, matching
@@ -612,7 +593,7 @@ export async function runFullAnalysis(
     );
   }
   const branchLabel = options.branch ?? checkedOutBranch;
-  const placement = await resolveBranchPlacement(repoPath, branchLabel);
+  const placement = options.branch ? await resolveBranchPlacement(repoPath, branchLabel) : {};
   const { lbugPath, metaPath } = getStoragePaths(repoPath, placement.branch);
   // Directory that owns this run's meta.json (flat `.gitnexus` for the primary
   // slot, `branches/<slug>/` otherwise). loadMeta/saveMeta operate on it so
@@ -620,21 +601,6 @@ export async function runFullAnalysis(
   const metaDir = path.dirname(metaPath);
 
   const existingMeta = await loadMeta(metaDir);
-
-  // ── #2106 (R8): warn when the repo's default branch is not the primary ──
-  // A non-default branch can own the flat slot (it was indexed first). That
-  // index is still fully queryable via `--branch`, so this is an ergonomics
-  // wart, not data loss — we only warn (no risky relocation of a live DB).
-  if (repoHasGit) {
-    // Who owns the flat slot after this run? For a flat/primary run it is this
-    // run's resolved label (carrying an existing stamp forward); for a branch
-    // run the flat owner is unchanged, so read the flat meta.
-    const flatOwner = placement.branch
-      ? (await loadMeta(storagePath))?.branch
-      : (branchLabel ?? existingMeta?.branch);
-    const warning = primaryInversionWarning(getDefaultBranch(repoPath), flatOwner);
-    if (warning) log(warning);
-  }
 
   // ── FTS-only repair path ────────────────────────────────────────────
   if (options.repairFts) {
@@ -862,6 +828,17 @@ export async function runFullAnalysis(
       const healUnregistered =
         options.allowDuplicateName === true && !(await isRepoRegistered(repoPath));
       if (!dirty && !healUnregistered) {
+        // ── #2354: restamp the workspace label on a same-commit branch flip ──
+        // The flat slot follows the checked-out working tree; a branch switch
+        // at the SAME commit with a clean tree changes nothing the pipeline
+        // must rebuild, but the slot's informational `branch` label (and the
+        // registry copy that query-side branch scoping reads) would go stale.
+        // Detached HEAD / non-git (branchLabel === null) keeps the existing
+        // stamp, mirroring the end-of-run meta write.
+        if (!placement.branch && branchLabel && existingMeta.branch !== branchLabel) {
+          await saveMeta(metaDir, { ...existingMeta, branch: branchLabel });
+          await adoptFlatBranchLabel(repoPath, branchLabel);
+        }
         await ensureGitNexusIgnored(repoPath);
         return {
           // `resolveRepoIdentityRoot` collapses worktree roots to the
@@ -1593,6 +1570,14 @@ export async function runFullAnalysis(
       // top-level fields (#2106).
       branch: placement.branch,
     });
+
+    // ── #2354: the flat workspace slot has adopted this run's branch ──────
+    // Drop a now-shadowed `branches/<slug>/` sub-index for the same label
+    // (unreachable once the flat slot serves it) and align the registry's
+    // top-level branch label.
+    if (!placement.branch && branchLabel) {
+      await adoptFlatBranchLabel(repoPath, branchLabel);
+    }
 
     // Keep generated .gitnexus contents ignored without editing the user's root .gitignore.
     await ensureGitNexusIgnored(repoPath);
--- a/gitnexus/src/core/run-analyze.ts
+++ b/gitnexus/src/core/run-analyze.ts
@@ -1620,9 +1620,18 @@ export async function runFullAnalysis(
     // ── #2354: the flat workspace slot has adopted this run's branch ──────
     // Drop a now-shadowed `branches/<slug>/` sub-index for the same label
     // (unreachable once the flat slot serves it) and align the registry's
-    // top-level branch label.
+    // top-level branch label. Best-effort like the parse-cache save above
+    // (#2364 review F5): the index is already complete and registered, and
+    // adopt retries unconditionally on the next plain analyze — a registry
+    // write failure must not fail an otherwise successful multi-minute run.
     if (!placement.branch && branchLabel) {
-      await adoptFlatBranchLabel(repoPath, branchLabel);
+      try {
+        await adoptFlatBranchLabel(repoPath, branchLabel);
+      } catch (e) {
+        log(
+          `Warning: could not sync the workspace branch label (${(e as Error).message}); continuing.`,
+        );
+      }
     }
 
     // Keep generated .gitnexus contents ignored without editing the user's root .gitignore.
--- a/gitnexus/src/mcp/local/local-backend.ts
+++ b/gitnexus/src/mcp/local/local-backend.ts
@@ -396,7 +396,7 @@ interface RepoHandle {
   stats?: RegistryEntry['stats'];
   /** Primary/flat branch name, when known (#2106). */
   branch?: string;
-  /** Non-primary branch indexes available for this repo (#2106). */
+  /** Pinned `--branch` sub-indexes available for this repo, distinct from the flat workspace slot (#2106/#2354). */
   branches?: BranchSummary[];
 }
 
@@ -584,7 +584,7 @@ export interface RepoListing {
   siblings?: Array<{ name: string; path: string; lastCommit: string }>;
   /** Primary/flat branch name, when known (#2106). */
   branch?: string;
-  /** Non-primary branch indexes available for this repo (#2106). */
+  /** Pinned `--branch` sub-indexes available for this repo, distinct from the flat workspace slot (#2106/#2354). */
   branches?: Array<Omit<BranchSummary, 'stats'>>;
 }
 
--- a/gitnexus/src/mcp/local/local-backend.ts
+++ b/gitnexus/src/mcp/local/local-backend.ts
@@ -1185,10 +1185,14 @@ export class LocalBackend {
     const indexed = [handle.branch, ...(handle.branches?.map((b) => b.branch) ?? [])].filter(
       Boolean,
     );
-    const available = indexed.length > 0 ? indexed.join(', ') : '(primary only)';
+    const available = indexed.length > 0 ? indexed.join(', ') : '(workspace only)';
+    // Post-#2354 a bare `analyze --branch <X>` refuses to run unless X is
+    // checked out, so the guidance must lead with the checkout (#2364 F6).
     throw new Error(
       `Branch "${branch}" is not indexed for "${handle.name}". ` +
-        `Indexed branches: ${available}. Run: gitnexus analyze --branch ${branch}`,
+        `Indexed branches: ${available}. The workspace index follows the ` +
+        `checked-out branch — check out "${branch}" and re-run: gitnexus analyze ` +
+        `(add --branch ${branch} while it is checked out to pin a separate sub-index).`,
     );
   }
 
--- a/gitnexus/src/mcp/local/local-backend.ts
+++ b/gitnexus/src/mcp/local/local-backend.ts
@@ -1120,10 +1120,10 @@ export class LocalBackend {
   /**
    * Re-point a resolved repo handle at a specific branch index (#2106).
    *
-   * - No `branch` (default) → the primary/flat handle, unchanged (backward
+   * - No `branch` (default) → the flat workspace handle, unchanged (backward
    *   compatible: every existing caller passes no branch).
-   * - `branch` equal to the known primary → the flat handle.
-   * - `branch` matching an indexed non-primary branch → a handle whose
+   * - `branch` equal to the flat slot's recorded branch → the flat handle.
+   * - `branch` matching an indexed pinned branch → a handle whose
    *   `lbugPath` points at `branches/<slug>/lbug`; the connection pool keys by
    *   `lbugPath`, so this is the only change needed to scope every tool.
    * - `branch` that was never indexed → a clear error (never a silently-empty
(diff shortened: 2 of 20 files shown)


gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -1143,15 +1143,17 @@ export class LocalBackend {
...
+    // workspace slot follows the checked-out working tree (#2354), so a plain
+    // analyze after a branch switch restamps the flat meta without any repo-
+    // resolution miss that would refresh this handle. Also covers legacy
+    // pre-#2106 entries with no recorded primary `branch`. Read the flat
+    // meta.json (next to the flat handle's lbug) and serve the flat handle
+    // only when it actually matches — never serve flat for an arbitrary
+    // unindexed branch (#2106 R4). Error path only, so the extra read costs
+    // nothing on scoped-query hits.
+    const flatMeta = await loadMeta(path.dirname(handle.lbugPath));
+    if (flatMeta?.branch && flatMeta.branch === branch) return handle;
```
magyargergo: **P1 — stale cached handle serves wrong-branch data / routes to a deleted sub-index** (cross-engine: Claude adversarial/correctness/reliability/risk-architect + all 3 Codex legs; synthesis re-verified [code-read]).

This new flat-meta fallback is correct but sits **after** two checks that trust the in-memory cached handle, which is only refreshed on resolve miss/`list_repos` — never on a cache hit. This PR makes flat-slot ownership flips routine, so the stale window is now the common case and unbounded:

1. **Wrong data (old owner):** line 1134's `handle.branch === branch` early-return has no disk check. Server caches `branch='main'`; user checks out `feature` + plain analyze; `{branch:'main'}` cache-hits the stale label and returns the flat handle — whose pool the `indexedAt` staleness reinit (~1304-1327) has already hot-swapped to *feature* content (it never updates `handle.branch`). Feature's graph is silently served as "main"; correct behavior is the "not indexed" error (#2106 R4, this function's own contract).
2. **Deleted sub-index (new owner):** a stale cached `branches[]` summary (lines 1135-1145) matches **before** this fallback. After `adoptFlatBranchLabel` rm's `branches/<slug>/`, scoped calls route to the deleted dir → raw `LadybugDB not found … Run: gitnexus analyze` (a fast-path no-op, so the advice can't heal), or ghost reads from an already-open unlinked pool with the staleness check blinded (meta gone → "assume pool is fine", ~1333) — while the correct flat answer is one disk read away here.

The new test (calltool-dispatch.test.ts:3582) covers only the coherent-registry miss arm — no stale `branches[]`, no old-label request.

**Suggested fix:** make the on-disk flat meta authoritative: read it before trusting `handle.branch` equality and before honoring a `branches[]` summary (or stat the summary's lbugPath and fall through on miss); trigger one `refreshRepos()` on label mismatch. Regressions: old-label-must-error + stale-summary-must-serve-flat.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -910,6 +910,53 @@ export const removeBranchIndex = async (repoPath: string, branch: string): Promi
...
+ */
+export const adoptFlatBranchLabel = async (repoPath: string, branch: string): Promise<void> => {
+  const resolved = path.resolve(repoPath);
+  const { storagePath } = getStoragePaths(resolved);
+  // Remove a shadowed sub-index directory, mirroring `clean --branch`'s
+  // containment guard: the target MUST live under .gitnexus/branches/.
+  const branchDir = path.join(storagePath, BRANCHES_DIR, branchSlug(branch));
+  const branchesRoot = path.join(storagePath, BRANCHES_DIR) + path.sep;
+  if (branchDir.startsWith(branchesRoot)) {
+    await fs.rm(branchDir, { recursive: true, force: true }).catch(() => {});
```
magyargergo: **P2 — unregistered repo loses its pinned sub-index** (cross-engine: all 3 Codex legs + Claude correctness; synthesis **[reproduced]** — script with an isolated empty registry + materialized `branches/<slug>/` dir: `adoptFlatBranchLabel` deleted it).

The docstring (lines 925-927) promises "a missing registry entry is a no-op (never self-heals … #2264/#1169)", but this `fs.rm` runs before the registry lookup and `if (idx < 0) return;` at lines 944-948 — the no-self-heal guard only protects the registry write, not the disk deletion. Reachable: the documented #2264 half-finalized state (up-to-date but unregistered) + a same-commit branch flip → fast-path restamp (run-analyze.ts:838-840) fires → pinned index irreversibly deleted, then the run still hard-fails `assertAnalysisFinalized`. A run that reports failure has destroyed an index.

Note the PR's own "never self-heals" test (repo-manager.test.ts:~703) only asserts the registry stays empty — it never materializes a branch dir, so it encodes the buggy disk behavior.

**Suggested fix:** move the registry lookup + `idx < 0` early-return above the `fs.rm` block; add a regression that an unregistered repo's `branches/<slug>/` dir survives.

gitnexus/src/core/run-analyze.ts
```diff
@@ -862,6 +828,17 @@ export async function runFullAnalysis(
...
       if (!dirty && !healUnregistered) {
+        // ── #2354: restamp the workspace label on a same-commit branch flip ──
+        // The flat slot follows the checked-out working tree; a branch switch
+        // at the SAME commit with a clean tree changes nothing the pipeline
+        // must rebuild, but the slot's informational `branch` label (and the
+        // registry copy that query-side branch scoping reads) would go stale.
+        // Detached HEAD / non-git (branchLabel === null) keeps the existing
+        // stamp, mirroring the end-of-run meta write.
+        if (!placement.branch && branchLabel && existingMeta.branch !== branchLabel) {
+          await saveMeta(metaDir, { ...existingMeta, branch: branchLabel });
```
magyargergo: **P2 — saveMeta-before-adopt makes partial failure permanent; also a #1549 read-only regression** (Claude risk-architect + reliability + adversarial converged on the same fix; synthesis re-verified [code-read]).

The retry condition for this block is `existingMeta.branch !== branchLabel` (line 838), and this `saveMeta` flips it **before** `adoptFlatBranchLabel` runs. A crash between the two, or a `writeRegistry` failure inside adopt (ENOSPC, `~/.gitnexus` perms — the dir rm at repo-manager.ts:937 has already happened), leaves flat meta stamped with the new label while the stale registry summary and/or `branches/<slug>/` dir survive — and every subsequent same-commit run takes the fast path, sees the guard false, and **never retries the cleanup**. Branch-scoped queries then keep routing to the stale pinned sub-index until an unrelated commit-changing full run.

Separately: this block sits outside any try/catch (the pipeline `try` starts at line 1071) and neither `saveMeta` nor `writeRegistry` tolerates EROFS — unlike `ensureGitNexusIgnored`, which explicitly absorbs EROFS/EACCES/EPERM for the documented Docker `:ro` workflow (repo-manager.ts:461-501, #1549). Pre-PR the "Already up to date" path performed no other writes; post-PR a same-commit branch flip on a read-only mount makes a byte-for-byte-current analyze exit 1 over a purely informational label sync.

**Suggested fix:** reorder — `adoptFlatBranchLabel` first (idempotent: early-returns when already coherent), `saveMeta` last, so any partial failure leaves the guard true and the next run self-heals; wrap the block with the existing `isReadOnlyFilesystemError` pattern to degrade RO storage to a warning.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -910,6 +910,53 @@ export const removeBranchIndex = async (repoPath: string, branch: string): Promi
...
+  }
+
+  const canonicalInput = canonicalizePath(repoPath);
+  const entries = await readRegistry();
+  const idx = entries.findIndex((e) =>
+    registryPathEquals(canonicalizePath(e.path), canonicalInput),
+  );
+  if (idx < 0) return; // unregistered → no-op (no self-heal)
+  const entry = entries[idx];
+  const remaining = entry.branches?.filter((b) => b.branch !== branch);
```
magyargergo: **P2 — Windows-locked sub-index becomes un-cleanable disk bloat** (Claude reliability/correctness/facts/test-ci; Codex ce flags the non-recoverable ordering as residual risk; synthesis re-verified [code-read]).

The `fs.rm` at line 937 swallows all errors (`.catch(() => {})`, no log), yet the summary drop here proceeds unconditionally. On Windows, rm of an lbug held open (mmap) by a long-lived MCP server fails EBUSY/EPERM — routine, not exotic. Once the summary is gone, `clean --branch` cannot target the leftover: it resolves **only** via the recorded summary (clean.ts:45-48 → `branchNotIndexed`). Result: a stranded multi-hundred-MB `branches/<slug>/` dir — the exact "un-cleanable disk bloat" this function's docstring (lines 920-922) says it exists to prevent — whose cacheKeys also keep parse-cache chunks retained via the disk-scanning `collectBranchCacheKeys`. Same-commit fast-path runs never retry the rm (the restamp guard is already satisfied).

**Suggested fix:** only drop the `branches[]` summary when the rm verifiably succeeded (dir confirmed gone); on failure keep the summary so `clean --branch` still works, and log a warning with the errno.

magyargergo (commented):
# Tri-engine review digest — PR #2364 `feat: flat workspace index follows the checked-out branch` (#2354)

**Engines:** Claude ran 13 lanes (6 swarm reviewers + 7 ce-code-review personas); Codex GPT-5.5/xhigh ran 3 legs (swarm, ce, adversarial), all verified live. Honest caveat: that is **two engines, not 16 independent opinions** — same-engine convergence can share a blind spot, so findings below are weighted by *cross-engine* corroboration, and every postable finding was independently re-verified against the worktree at head `12dc32d2` (base `6e42040`).

## What's solid

- The **placement inversion itself is correct and well-tested**: plain analyze always targets the flat slot (`run-analyze.ts:596` ternary), explicit `--branch` still pins, detached-HEAD/non-git keep the existing stamp, and the `--branch`-must-match-HEAD guard holds. Both engines traced these paths and found no defect in the core semantics.
- **Shadow cleanup design is coherent**: `adoptFlatBranchLabel` drops the now-unreachable `branches/<slug>/` dir and registry summary in one pass, with direct unit + integration coverage (relabel+delete, keep-others, reclaim round-trip).
- **No-self-heal (#2264/#1169) is preserved on the registry side**, with a dedicated test (the *disk* side has a gap — Finding 2).
- **Security: all destructive-path probes refuted by both engines independently** — path traversal via branch name (slug = sanitize + sha256 suffix, no separator survives), `startsWith` containment guard, symlink-following `fs.rm`, MCP `branch` param reaching any filesystem/destructive sink, registry JSON injection. Clean bill.
- **Hygiene/i18n/DoD clean**: single focused commit, prettier/eslint clean, en/zh-CN key parity holds, no CHANGELOG edit, PR template complete, no hidden unicode, conventional title, `primaryInversionWarning` deleted cleanly with its tests.
- **CI fully green** at post time (quality, CodeQL, gitleaks, coverage, Windows/macOS platform tests, ABI, packaging, benchmarks, Docker); Vercel failure is the standard fork auth gate.

---

## Headline findings

### 1. P1 — `applyBranchScope` trusts stale cached branch state after a workspace flip (two arms: silent wrong-branch data, and routing to a deleted sub-index)

**Corroboration: STRONG (cross-engine).** Claude adversarial/correctness/reliability at P1 (risk-architect rated the branches[]-arm P2) + Codex swarm (P1), Codex adversarial (high), Codex ce (#1, confidence 100, both arms). **[code-read]** — mechanics re-verified line-by-line in the worktree. **Blocks merge: yes.**

The PR added a flat-meta disk fallback in `applyBranchScope` (`local-backend.ts:1155-1156`), but it sits **after** two checks that trust the in-memory cached handle, and the handle cache is only refreshed on resolve miss/ambiguity or `list_repos` — never on a successful cache-hit resolution. This PR is what makes flat-slot ownership flips routine (every hook-driven plain analyze after a branch switch), so the stale window is now the common case, unbounded in duration:

- **Arm (i) — wrong data for the old owner** (`local-backend.ts:1134`, unchanged context: `if (handle.branch && handle.branch === branch) return handle;`). Trigger: long-lived MCP server caches `handle.branch='main'`; user checks out `feature` (different commit) and runs plain `gitnexus analyze` → flat slot rebuilt with feature content, meta/registry restamped. A tool call `{branch:'main'}` cache-hits, matches the stale label, and returns the flat handle — whose pool the `indexedAt` staleness check (`local-backend.ts:1304-1327`) has already hot-swapped to **feature** content, while never updating `handle.branch`. Result: feature's graph silently served as "main". Correct post-#2354 behavior is the clear "not indexed" error — the exact "never a silently-empty result against the wrong DB" invariant the function's own doc-comment claims (#2106 R4), violated in the worse direction.
- **Arm (ii) — stale `branches[]` summary shadows the new fallback** (`local-backend.ts:1135-1145`, checked before the fallback). Trigger: user had pinned `analyze --branch feat`; server caches the summary; user checks out feat and runs plain analyze → `adoptFlatBranchLabel` deletes `branches/<slug>/` and drops the summary. The stale cached summary still matches first and routes to the **deleted** directory: either a raw `LadybugDB not found … Run: gitnexus analyze` (whose advice is a fast-path no-op — the labels already match, see Finding 3's guard), or — if the pool was already open on POSIX — ghost reads of the unlinked pre-adoption DB with staleness detection permanently blinded (the meta beside it is gone; `ensureInitialized` catch says "assume pool is fine", `local-backend.ts:~1333`). Meanwhile the correct answer (the flat slot now owns feat) is one disk read away at line 1155.

The PR's own regression test (`calltool-dispatch.test.ts:3582`) covers only the coherent-registry **miss** arm (no stale `branches[]` in the fixture, no old-label request) — verified by reading the fixture.

**Suggested fix:** make the on-disk flat meta authoritative: when `branch` is requested, read it (the PR already pays this read on the fallback path) **before** trusting `handle.branch` equality and **before** honoring a `branches[]` summary (or `stat` the summary's lbugPath and fall through on miss); on label mismatch trigger one `refreshRepos()`.

**Inline comment anchor:** `gitnexus/src/mcp/local/local-backend.ts:1156` (added line, side RIGHT).

### 2. P2 — `adoptFlatBranchLabel` deletes the pinned sub-index BEFORE the registry-miss no-op check — unregistered repos lose data on the fast path

**Corroboration: STRONG (cross-engine; severity split disclosed — Codex rates it P2 unanimously, the two Claude lanes that reached it rated it P3; the live reproduction justifies P2).** Codex swarm (P2), Codex adversarial (high), Codex ce (#2, confidence 100) + Claude correctness F4. **Blocks merge: yes (with Finding 3's reorder, the same fix).** **[reproduced]** — I executed a tsx script against the worktree: created `branches/<slug>/` for a repo with an **empty registry** (`GITNEXUS_HOME` isolated), called `adoptFlatBranchLabel` → directory deleted (`before: exists=true / after: exists=false`).

The function's docstring (`repo-manager.ts:925-927`) promises "a missing registry entry is a no-op (never self-heals … per #2264/#1169)", but the `fs.rm` at `repo-manager.ts:937` runs before the registry lookup and `if (idx < 0) return;` at `:944-948`. Reachable path: the documented #2264 half-finalized state (index up-to-date but unregistered after a `--name` collision) plus a same-commit branch flip → the fast-path restamp (`run-analyze.ts:838-840`) fires, the pinned sub-index is irreversibly deleted, and the run **still hard-fails** `assertAnalysisFinalized`. A run that reports failure has silently destroyed an index. Note: the PR's own "never self-heals" unit test (`repo-manager.test.ts:~703`) only asserts the registry stays empty — it never materializes a branch dir, so it encodes the current buggy disk behavior.

**Suggested fix:** move the registry lookup and `idx < 0` early-return above the `fs.rm` block; add the regression Codex ce listed (unregistered repo keeps its `branches/<slug>/` dir).

**Inline comment anchor:** `gitnexus/src/storage/repo-manager.ts:937` (added line, side RIGHT).

### 3. P2 — Fast-path restamp order: `saveMeta` before `adoptFlatBranchLabel` flips the retry guard, so a failed adopt is never retried; the block can also hard-fail read-only workflows

**Corroboration: MODERATE (Claude-only — risk-architect F1, reliability R2+R4, adversarial F3 — three lanes independently converging on the same fix; no Codex leg reached it, though Codex ce's residual-risks note the non-recoverable update order).** **Blocks merge: yes (same reorder as Finding 2).** **[code-read]** — verified: the guard at `run-analyze.ts:838` is `existingMeta.branch !== branchLabel`; `saveMeta` at `:839` flips it before `adoptFlatBranchLabel` at `:840` runs; inside adopt, the dir rm (`repo-manager.ts:937`) precedes the fallible `writeRegistry` (`:957`).

Two verified consequences:

- **Stuck split state:** a crash between 839 and 840, or a `writeRegistry` failure inside adopt (ENOSPC, `~/.gitnexus` perms), leaves flat meta stamped with the new label while the registry summary and/or `branches/<slug>/` dir survive. Every subsequent same-commit run takes the fast path, the guard is now false, and the cleanup is **never retried** — branch-scoped queries keep routing to the stale pinned sub-index (Finding 1 arm ii, but durable across server restarts) until an unrelated commit-changing full run.
- **Read-only regression (#1549 workflow):** the restamp block sits outside any try/catch (the pipeline `try` starts at `run-analyze.ts:1071`) and neither `saveMeta` nor `writeRegistry` tolerates EROFS — unlike `ensureGitNexusIgnored`, which explicitly absorbs EROFS/EACCES/EPERM for the documented Docker `:ro` bind-mount (verified at `repo-manager.ts:461-501`). Pre-PR the "Already up to date" path performed no other writes; post-PR a same-commit branch flip on a read-only mount makes a byte-for-byte-current analyze **exit 1** on a purely informational label sync.

**Suggested fix (lanes agree):** reorder — `adoptFlatBranchLabel` first (idempotent: early-returns when registry is already coherent), `saveMeta` last, so any partial failure leaves the guard true and the next fast-path run self-heals; wrap the block with the existing `isReadOnlyFilesystemError` pattern so RO storage degrades to a warning instead of a failed run.

**Inline comment anchor:** `gitnexus/src/core/run-analyze.ts:839` (added line, side RIGHT).

### 4. P2 — `fs.rm` failure swallowed but the registry summary is dropped unconditionally — a Windows-locked sub-index becomes invisible to `clean --branch`

**Corroboration: MODERATE (Claude reliability R3 / correctness F3 / facts-historian B / test-ci gap 1; Codex ce touches it only as a residual risk).** **Blocks merge: no (Windows-specific trigger; fix strongly recommended in-PR).** **[code-read]** — verified: `repo-manager.ts:937` is `await fs.rm(...).catch(() => {});` (error discarded, no log); the summary drop at `:950-957` proceeds regardless; and `clean --branch` resolves its target **only** via the recorded summary (`clean.ts:45-48` — `if (!summary) { … branchNotIndexed … return; }`), so once the summary is gone the on-disk dir can never be targeted.

Trigger: Windows user with a live MCP server holding `branches/<slug>/lbug` open (mmap) switches to that branch and runs plain analyze → rm fails EBUSY/EPERM, silently → summary dropped anyway → stranded multi-hundred-MB sub-index that `clean --branch` reports as "not indexed" — the exact "un-cleanable disk bloat" the function's docstring (`repo-manager.ts:920-922`) says it exists to prevent. Its recorded cacheKeys also keep parse-cache chunks retained via the disk-scanning `collectBranchCacheKeys`. Partial mitigation exists (a later full analyze on the same label retries the rm; whole-repo `clean` still works), but same-commit fast-path runs never retry (Finding 3's guard).

**Suggested fix:** only drop the `branches[]` summary when the rm verifiably succeeded (dir confirmed gone); on failure keep the summary (so `clean --branch` still works) and log a warning with the errno.

**Inline comment anchor:** `gitnexus/src/storage/repo-manager.ts:950` (added line, side RIGHT).

---

## Lower-priority notes (P3 / informational — no inline comments)

- **End-of-run adopt is not best-effort** (`run-analyze.ts:1578-1580`, inside the pipeline try whose catch rethrows): a `writeRegistry` ENOSPC after a successful multi-minute analyze fails the whole run even though the index is complete and registered — the parse-cache save two blocks earlier is deliberately wrapped for exactly this reason. Warn-and-continue is strictly safer (cleanup retries unconditionally on the next flat analyze). *(single Claude lane)*
- **Error guidance now sends users to a command that refuses to run:** `local-backend.ts:1161-1164` says `Run: gitnexus analyze --branch <X>`, but post-#2354 that command hard-errors unless X is checked out (`run-analyze.ts:~590`). This becomes the common goodbye message for a formerly-indexed branch; wording should explain that the workspace index followed the checkout.
- **Vocabulary drift:** the user-facing `'(primary only)'` string at `local-backend.ts:1160` sits three lines below the comment block this PR rewrote to "workspace" terms; `isPrimaryBranch` JSDoc (`run-analyze.ts:281-286`), two `branches?` JSDoc comments (`local-backend.ts:399,587`), a comment at `cli/analyze.ts:1321`, and 4 surrounding test names in `calltool-dispatch.test.ts` still speak pre-inversion "primary/non-primary".
- **i18n wording:** `status.workspaceIndexLabel` — "gitnexus analyze follows this branch" is ambiguous about *which* branch (recorded vs current); suggest "re-run gitnexus analyze to follow the current branch" (en + zh-CN). Also a **pre-existing** gap re-exposed: the `analyze --branch` help description is hardcoded English, bypassing `OPTION_DESCRIPTION_KEYS`.
- **Maintainability:** `adoptFlatBranchLabel` adds a fifth inline copy of the canonicalize→readRegistry→findIndex idiom, and a second copy of the safety-critical `branches/` containment guard (its own comment says "mirroring `clean --branch`'s guard") — both are one-helper extractions.
- **Registry concurrency:** adopt is a new unlocked read-modify-write that now fires on every branch flip, without `registerRepo`'s #2106-R9 re-read-before-write narrowing — a pre-existing lost-update class with materially increased write frequency; a shared registry-mutation helper (or lockfile) would retire the whole class.
- **Tests:** 5 new `as any` mock casts (warn-level, matches the pervasive house pattern in these two files; `vi.mocked(...)` preferred for new code). Full-suite attestation in the PR body is a targeted 7-file list rather than `npm test` (CI covers the rest).
- **PR #2363 collision (sequencing, not a defect):** this PR adds two new raw `meta.json` reads (`local-backend.ts:1155` fallback, fast-path restamp) that join the meta.json→gitnexus.json rename blast radius if both land; merge order is #2364 first, rebase #2363.

## Refuted suspicions (validation both engines performed)

Path traversal / guard bypass / symlink-following in `adoptFlatBranchLabel`'s `fs.rm` (slug sanitization + sha256 suffix means no separator or `..` survives; Node's rm is lstat-based); MCP `branch` param reaching any destructive/filesystem sink or DoS (branch is only string-compared; the new read derives from `handle.lbugPath`); registry/meta JSON injection via branch names; the flat-meta fallback racing a mid-rewrite flat DB (meta is written only after persistence, atomically); dirty-tree-at-same-commit skipping the restamp (the full path stamps everything); detached-HEAD / non-git / `--skip-git` regressions (label-null no-ops mirror pre-PR behavior); upgrade stranding of old auto-created sub-dirs (still queryable; converted on first plain analyze); slug collisions deleting a sibling branch's index; the "Windows case-sensitivity of `startsWith`" test-ci suspicion (false positive — both strings derive from the same `storagePath` in the same invocation); status fall-through regression (graceful, correct remedy text).

## Test gaps worth closing (in-PR or immediate follow-up)

1. Stale cached handle, **old** flat label: resolving the pre-flip branch must error, not serve the flat handle (Finding 1 arm i) — Codex ce regression #1.
2. Stale cached `branches[]` summary whose sub-index adopt deleted: must fall through to the flat handle (Finding 1 arm ii) — Codex ce regression #2.
3. Unregistered repo keeps its materialized `branches/<slug>/` dir across `adoptFlatBranchLabel` (Finding 2) — Codex ce regression #3; reproduced failing today.
4. Restamp partial-failure retry: make adopt throw and assert the next same-commit run retries the cleanup (Finding 3).
5. `fs.rm` failure retains the registry summary / emits a warning (Finding 4).
6. Detached-HEAD **fast-path** restamp skip (the existing integration test uses `force: true` and never reaches the new block).
7. "Already up to date" still succeeds on read-only `.gitnexus` after a same-commit branch flip (#1549 workflow).

## Branch hygiene assessment

**clean feature/fix PR** — single focused commit rebased directly onto the main tip, no merge commits, no unrelated churn, all 16 files causally coherent to #2354, no hidden unicode, prettier/eslint clean.

## Merge status and mergeability

**mergeable — all required checks green** (re-checked at post time): quality, CodeQL (2×), gitleaks, dependency review, ubuntu coverage, Windows + macOS platform tests, ABI, packaging smoke, benchmarks, Docker build — all SUCCESS at head `12dc32d2`. Vercel FAILURE is the standard fork auth gate, not a signal. E2E correctly skipped (no web changes).

Current visible state is otherwise complete for CI; what was NOT verified is runtime behavior on a real Windows host (Finding 4's trigger) and a live long-running MCP server deployment (Finding 1's window was verified by code trace and the pool-reinit code path, not a deployed reproduction).

## Final verdict

**not production-ready** — the core feature (placement inversion, shadow cleanup, status/UX, security posture) is correct, well-tested, and both engines validated it extensively. But Finding 1 (P1, cross-engine) means the primary MCP runtime surface can silently serve wrong-branch data or route to a deleted sub-index in exactly the workflow this PR makes routine, and Findings 2–3 (P2) put a data-deleting side effect ahead of its own no-op guard with a retry gate that locks in partial failure. All three share small, localized fixes (meta-first ordering in `applyBranchScope`; registry-check-first + adopt-before-saveMeta in the write path) — after those and the listed regressions, this is merge-ready.

## Coverage

Full 16-file diff reviewed by all lanes (16=16 sanity-checked); every postable finding above was re-verified in the worktree by the synthesis pass, with one live reproduction (Finding 2) and diff-anchored line checks for all four inline comments.

*Automated multi-tool digest (2 engines, 16 lanes/legs) — verify before acting.*

gitnexus/src/mcp/local/local-backend.ts
magyargergo: Fixed in 6bf2cb98 (+ hardening in e8e4970b, 3b62b78c): the flat meta is read before any cached-state trust; the summary arm serves only when the sub-index lbug exists on disk (ENOENT/ENOTDIR-only polarity); the no-readable-meta legacy trust arm is preserved (#2106 R4, pinned); every miss fires exactly one refreshRepos() so newly-pinned/restamped state heals on the next call. Regressions cover both arms plus the dangling-summary degraded window.

gitnexus/src/storage/repo-manager.ts
magyargergo: Fixed in 3003271f (+ ee64d2aa): the registry membership check now precedes the rm, so the no-self-heal rule covers disk state — the reproduced unregistered-repo deletion is pinned by a regression that materializes the dir and asserts survival. ee64d2aa additionally re-reads the registry after the (slow) rm before writing, keeping the #2106 R9 tight read-to-write window.

gitnexus/src/core/run-analyze.ts
magyargergo: Fixed in 82f046d7: adopt-first / saveMeta-last (any partial failure leaves the retry guard true; next same-commit run self-heals — pinned by a retry-after-partial-failure regression), and the block is best-effort: EROFS/EACCES/EPERM degrade to a #1549-citing warning (it.each-tested), other errors warn-and-retry. Detached-HEAD fast-path skip also pinned (gap 6).

gitnexus/src/storage/repo-manager.ts
magyargergo: Fixed in f46e90e2 (+ 3b62b78c): the branches[] summary is dropped only when the dir is verifiably gone — a resolved force:true rm proves absence; on rm failure the probe treats only ENOENT/ENOTDIR as gone (an EACCES would otherwise re-create this bug), the summary is retained so clean --branch keeps working, and a warning names the path and errno. Later adopts retry the rm (pinned).

Issue #1515, opened 2026-05-12
GitNexus fails on macOS Intel (x86_64) with 'Cannot find module ../bin/napi-v6/darwin/x64/onnxruntime_binding.node'


chendacai:
## Bug Description

GitNexus fails to run on macOS Intel (x86_64) architecture with a module not found error for onnxruntime-node's native binding.

## Steps to Reproduce
1. On a macOS Intel (x86_64) machine
2. Run `gitnexus analyze --embeddings` or `ONNX_WEB_BACKEND=wasm npx gitnexus analyze --embeddings`
3. See the error

## Expected Behavior
GitNexus should run successfully on macOS Intel (x86_64) or provide a clear error message indicating unsupported platforms.

## Actual Behavior
Error:
```
GitNexus Analyzer

  Analysis failed: Cannot find module '../bin/napi-v6/darwin/x64/onnxruntime_binding.node'
Require stack:
- .../@huggingface/transformers/node_modules/onnxruntime-node/dist/binding.js
- .../@huggingface/transformers/node_modules/onnxruntime-node/dist/backend.js
- .../@huggingface/transformers/node_modules/onnxruntime-node/dist/index.js
```

## Environment
- OS: macOS (Darwin)
- Architecture: x86_64 (Intel)
- GitNexus version: latest

## Root Cause Analysis
The issue is that @ladybugdb/core (GitNexus's core dependency) only includes precompiled onnxruntime-node binaries for:
- ✅ darwin-arm64 (Apple Silicon)
- ✅ linux-x64
- ✅ win32-x64
- ✅ linux-arm64
- ❌ **darwin-x64 (macOS Intel) - missing**

## Impacted Platforms
| Platform | Status |
|----------|--------|
| macOS Intel (x86_64) | ❌ Not supported |
| macOS Apple Silicon (arm64) | ✅ Supported |
| Linux x86_64 | ✅ Supported |
| Windows x86_64 | ✅ Supported |
| Linux arm64 | ✅ Supported |

## Possible Solutions
1. Add darwin-x64 precompiled binaries to the @ladybugdb/core package
2. Provide a fallback mechanism for unsupported architectures
3. Document platform support clearly in the README

xiaolu289:
also the same problem

lzsheng:
also the same problem
macbook pro 2019 (Intel Core i7)

magyargergo:
Could you please try the latest RC version and let me know if this still is an issue? 🙏

[...]

PR #2372, opened 2026-07-04
fix: proxy-blocked installs survive onnxruntime-node postinstall and self-heal embeddings (#2370)

### LINKED ISSUE #2370: npm install -g gitnexus fails behind HTTP proxy due to onnxruntime-node postinstall / global-agent issues
# `npm install -g gitnexus` fails behind HTTP proxy due to `onnxruntime-node` postinstall / `global-agent` issues

### Area

### Summary
`npm install -g gitnexus` fails in environments behind HTTP proxy due to `onnxruntime-node` postinstall script (via `global-agent`) not respecting standard `HTTP_PROXY`/`HTTPS_PROXY` env vars.

### Context
Installing GitNexus globally via npm on a machine that has `HTTP_PROXY`/`HTTPS_PROXY` environment variables set. The npm registry itself uses a Chinese mirror (`registry.npmmirror.com`) and works fine, but the `onnxruntime-node` dependency's postinstall script fails because it cannot reach `api.nuget.org` or Microsoft CDN to download the required binary.

### Expected behavior
`npm install -g gitnexus` should succeed out of the box, respecting standard proxy environment variables.

### Actual behavior
The install fails with one of two errors depending on configuration:

**Error 1** — Without proxy forwarding to NuGet:
```
npm error Error: Failed to download build list. HTTP status code = 302
```
The script at `onnxruntime-node/script/install-utils.js:57` receives a 302 redirect (likely because `api.nuget.org` redirects to `nuget.azure.cn` in certain regions) and treats non-200 as failure — it does not follow redirects.

**Error 2** — With `GLOBAL_AGENT_HTTPS_PROXY` set (to force proxy for `global-agent`):
```
npm error Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames
```
`global-agent` resolves the proxy hostname to `localhost` and then the TLS certificate check fails because the upstream Akamai CDN cert doesn't match `localhost`.

### Steps to reproduce
1. Set `HTTP_PROXY` and `HTTPS_PROXY` environment variables (e.g., `http://proxy.local:port`)
2. Run `npm install -g gitnexus`
3. Observe postinstall error from `onnxruntime-node`

**Workaround** (not ideal):
```bash
npm install -g gitnexus --ignore-scripts
# Then manually run:
cd $(npm root -g)/gitnexus
node scripts/materialize-vendor-grammars.cjs
node scripts/build-tree-sitter-dart.cjs
node scripts/build-tree-sitter-proto.cjs
node scripts/build-tree-sitter-swift.cjs
node node_modules/@ladybugdb/core/install.js
```

Or use pnpm:
```bash
pnpm --allow-build=@ladybugdb/core --allow-build=gitnexus --allow-build=tree-sitter install -g gitnexus@latest
```

### Root cause analysis
The problem has three layers:

1. **`onnxruntime-node` downloads binaries from NuGet/Microsoft CDN**, not from npm registry — so Chinese npm mirrors don't help. In China, `api.nuget.org` redirects (302) to `nuget.azure.cn`, but the download script at `install-utils.js:57` rejects any non-200 response instead of following redirects.

2. **`onnxruntime-node` uses `global-agent`** (not Node.js built-in `undici`/`fetch`) to honor proxy settings, but `global-agent` reads `GLOBAL_AGENT_HTTPS_PROXY`, not the standard `HTTPS_PROXY` env var. Users who set `HTTPS_PROXY` expecting it to work find that `global-agent` ignores it entirely, so the download goes direct.

3. **Even when `GLOBAL_AGENT_HTTPS_PROXY` is set**, `global-agent` has a bug where it resolves the proxy host incorrectly to `localhost`, causing an Akamai CDN TLS certificate mismatch error.

### Suggestions
- **Short-term**: Switch from `global-agent` to Node.js built-in `fetch` / `undici`, which natively respects `HTTP_PROXY`/`HTTPS_PROXY` and handles redirects automatically.
- **Medium-term**: Make `onnxruntime-node` an optional dependency and fall back gracefully when it fails to install — GitNexus doesn't need it for its core functionality (indexing, MCP, knowledge graph).
- **Alternative**: Add a `--skip-optional` or `GITNEXUS_SKIP_ONNX` flag to skip the onnxruntime install without `--ignore-scripts`.

### Environment
- OS: Ubuntu 22.04 (Linux 6.8.0-124-generic)
- Node: v22.23.1
- npm: 10.9.8
- GitNexus: 1.6.6 (latest)
- Proxy: http proxy via env vars `HTTP_PROXY` / `HTTPS_PROXY`

magyargergo:
Fixes #2370.

## Problem

`npm install -g gitnexus` dies with exit 1 behind HTTP proxies and regional firewalls: on every linux/x64 install, `onnxruntime-node`'s postinstall downloads CUDA EP binaries from `api.nuget.org` — outside the npm registry, so registry mirrors don't cover it. All three layers from the issue reproduced against `gitnexus@1.6.9`:

1. **Fatal install** — with a proxy configured for `global-agent` that can't connect, the whole install fails inside the postinstall (`ECONNREFUSED`).
2. **302 rejection** — `install-utils.js`'s `downloadJson` rejects any non-200: the regional `api.nuget.org → nuget.azure.cn` redirect produces the reported `Failed to download build list. HTTP status code = 302`.
3. **Standard proxy vars ignored** — `global-agent` (bootstrapped by the postinstall) makes the NuGet request **direct with zero proxy hits** when only `HTTP_PROXY`/`HTTPS_PROXY` are set; it reads exclusively `GLOBAL_AGENT_*`.

The key observation: those CUDA binaries are only needed on CUDA GPU hosts. The CPU ONNX binding ships inside the npm tarball, embeddings are opt-in, and the embedder already probe-gates CUDA with a CPU fallback — so a failed CUDA download should never brick the install.

## Fix (two commits)

**1. Installs can't brick (`optionalDependencies`)**
`@huggingface/transformers` + `onnxruntime-node` move from `dependencies` to `optionalDependencies` — the same pattern esbuild/sharp/swc use for fallible native deps. A failed postinstall now prunes the embedding stack instead of failing the install. `onnxruntime-common` stays a regular dependency (pure JS, no install script, needed by the #307 resolver fallback). The `$onnxruntime-node` override still resolves.

**2. Pruned installs self-heal (on-demand embedding runtime)**
A degraded install no longer needs a reinstall:

- The first `gitnexus analyze --embeddings` detects the missing stack and installs it into `~/.gitnexus/embedding-runtime` (`GITNEXUS_EMBEDDING_RUNTIME_DIR` to override) **through the user's npm registry config** — mirrors and proxies apply, i.e. exactly the network path that already worked for them — with `--ignore-scripts` + `ONNXRUNTIME_NODE_INSTALL=skip`, so **NuGet is never contacted**. CPU embeddings are fully functional without the CUDA binaries.
- New command: `gitnexus embeddings install [--cuda] [--force]` (en + zh-CN help). `--cuda` opts GPU hosts into the NuGet postinstall (`GLOBAL_AGENT_HTTPS_PROXY` respected).
- Resolution is package-first with a `registerHooks` fallback that re-anchors the stack's bare specifiers at the prefix (mirrors the #307 onnxruntime-common hook; registered before the CUDA-13 redirect so it sits last in the chain). The CUDA probe (`resolveDefaultOrtNodeDir`) also falls back to the prefix copy so device selection inspects the copy that actually loads.
- Install specs are read from gitnexus' own `optionalDependencies` manifest, so the on-demand install can never drift from a normal one.
- Runtime UX for the pruned state: `getMissingLocalEmbeddingStackMessage` (matches module-not-found codes + the quoted bare specifier, so #1515-style binding-path errors are never misreported) wired into both embedders, the `analyze` error handler (ahead of the generic "installation may be corrupt" hint), and `gitnexus doctor`.

The stack deliberately stays in `optionalDependencies` rather than being removed from install entirely: normal installs keep shipping embeddings preinstalled (no first-use download for everyone), and the Docker image (`npm prune --omit=dev`) is unaffected.

`ONNXRUNTIME_NODE_INSTALL=skip npm install -g gitnexus` remains documented as the way to keep the stack inside the global install itself (verified it works today, even against 1.6.9).

## Verification

All end-to-end runs used a tarball packed from this branch, with `GLOBAL_AGENT_HTTPS_PROXY` pointed at a **dead proxy the entire time**:

- Install exits **0** (1.6.9 exits 1); `gitnexus analyze` works on the degraded install.
- `doctor` → `✗ optional embedding stack not installed` with actionable guidance.
- `analyze --force --embeddings` prints the heal notice, installs the runtime from the npm registry, loads the model **from the prefix** (dev-mode log: `🧠 Loading embedding model` → `✅ Embedding model loaded successfully`), embeds the fixture repo; `doctor` then reports `✓ local embeddings supported`.
- A normal-network install still ships the full stack preinstalled (`✓ local embeddings supported`, ONNX 1.27.0).
- Unit: 94 tests across the six affected files, including the zh-CN help-i18n gate; full unit suite green apart from the 5 known load-flake files (all pass in isolation); `tsc` + pre-commit clean on both commits.

Follow-up worth filing upstream (microsoft/onnxruntime): make the CUDA EP download warn-and-continue instead of failing the install, follow redirects, and honour standard proxy env vars — that would let this repo revert the optional-deps move eventually.

https://claude.ai/code/session_01D4Kq6gNrEAE8uoyS7PRnTo


koriyoshi2041:
I found one compatibility edge while checking the new runtime fallback path.

`runtime-install.ts` says Node `<22.15` should degrade gracefully when `registerHooks` is unavailable, but the static named import prevents that on Node 22.0-22.14. GitNexus currently advertises `"node": ">=22.0.0"`, so this would make any path importing `runtime-install.ts` fail at module load time before the `typeof registerHooks` guard runs.

Repro:

```bash
npx -y node@22.14.0 --input-type=module -e "import { registerHooks } from 'node:module'; console.log(typeof registerHooks)"
```

Result:

```text
SyntaxError: The requested module 'node:module' does not provide an export named 'registerHooks'
```

A namespace import does preserve the intended fallback behavior on 22.14:

```bash
npx -y node@22.14.0 --input-type=module -e "import * as moduleApi from 'node:module'; console.log(typeof moduleApi.registerHooks)"
# undefined
```

So I think the options are either:

- switch to `import * as moduleApi from 'node:module'` and call `moduleApi.registerHooks` only when it is a function, or
- raise the package engine floor to `>=22.15.0` if the runtime-prefix fallback is required.

I also ran the focused new/related tests locally on this PR head (`ec06664`):

```bash
cd gitnexus
npx vitest run test/unit/embedding-runtime-install.test.ts test/unit/embedding-runtime-support.test.ts test/unit/doctor-format.test.ts test/unit/cli-index-help.test.ts
```

Result: `4 passed`, `53 passed`.

gitnexus/src/core/embeddings/runtime-install.ts
```diff
@@ -0,0 +1,212 @@
...
+ * from the user's configured npm registry — honouring their mirror and proxy
+ * settings, the part of their network setup that demonstrably works — with
+ * `--ignore-scripts`, so no NuGet download is attempted at all. The CPU ONNX
+ * binding ships inside the npm tarball; only CUDA GPU acceleration needs the
+ * postinstall, and `installEmbeddingRuntime({ cuda: true })` opts into it.
+ *
+ * Resolution is package-first: a normally-installed stack always wins, and the
+ * runtime prefix is only consulted when the bare specifier does not resolve.
+ */
+import { createRequire, registerHooks } from 'node:module';
```
magyargergo: **[P0] Node 22.0–22.14: this static named import is an ESM link-time SyntaxError — every `gitnexus analyze` invocation dies, and so does `gitnexus embeddings install` (confirms @koriyoshi2041, issuecomment-4884449781)** [reproduced]

`registerHooks` was added in Node 22.15.0; `engines` still allows `>=22.0.0` and the package is `"type": "module"`. In real Node ESM, a missing named export from a builtin fails at `ModuleJob._instantiate` — **before any statement runs**, so the `typeof registerHooks !== 'function'` guard at line 109 is unreachable dead code. Reproduced on a real node v22.14.0 binary: `SyntaxError: The requested module 'node:module' does not provide an export named 'registerHooks'`, `code: undefined` (so `getMissingLocalEmbeddingStackMessage` correctly doesn't match it — the user gets a raw, unexplained SyntaxError via the unhandled rejection from `cli/index.ts`'s `program.parse`).

**Regression vs base (verified via import-closure diff against 187c162f):** plain `gitnexus analyze` worked on 22.0–22.14 at base; at HEAD `cli/analyze.ts` statically imports this module (directly and via `runtime-support.ts`), so analyze crashes at command load with no `--embeddings` needed. The same pre-existing pattern sits at `onnxruntime-node-resolver.ts:54` and `onnxruntime-common-resolver.ts:60` (both from #2341 — their "graceful degradation" guards have never been reachable in the ESM dist either); fix all three together.

**Why CI is green:** `setup-node` `node-version: 22` resolves to ≥22.15, and vitest/tsx transforms mask the link error entirely (reproduced: the same import shape passes under the repo's vitest and under tsx on node 22.14).

**Fix:** namespace import (`import * as nodeModule from 'node:module'`) in all three files — verified to link on 22.14 with `nodeModule.registerHooks === undefined`, making the existing guards finally functional; use a typed optional narrow since `@types/node` declares the member always-present (no-`any` policy). Alternative: raise `engines` to `>=22.15.0` + CONTRIBUTING/README lockstep. Note the namespace import alone does NOT make self-heal work below 22.15 (see the `cli/embeddings.ts` comment). Add a dist-level smoke on a 22.14 matrix leg — vitest structurally cannot catch this class.

*Corroboration: all 3 Codex legs + 9 Claude lanes independently; reproduced on real node 22.14.0.*

gitnexus/src/core/embeddings/runtime-install.ts
```diff
@@ -0,0 +1,212 @@
...
+  opts: EmbeddingInstallOptions = {},
+): Promise<void> => {
+  const { args, env } = buildEmbeddingInstallCommand(opts);
+  await new Promise<void>((resolve, reject) => {
+    const child = spawn('npm', args, {
+      env,
+      windowsHide: true,
+      // Windows `npm` is a `.cmd` shim; without a shell spawn ENOENTs
+      // (mirrors getNpmMajorVersion in cli/resolve-invocation.ts).
+      shell: process.platform === 'win32',
```
magyargergo: **[P1] Windows: `shell: true` joins args unquoted — any profile path with a space breaks the self-heal, and cmd.exe eats the `^` in semver ranges** [reproduced mechanism / code-read for win32]

With `shell: true`, Node concatenates command+args into one cmd.exe string with **no quoting** (documented spawn behavior; the split was reproduced locally with `spawn('printf', ['[%s]','--prefix','/home/John Doe/x'], {shell:true})`). The cited precedent (`getNpmMajorVersion`, `cli/resolve-invocation.ts`) is only safe because it passes `['--version']` — no paths, no ranges.

Consequences: (a) `C:\Users\First Last\.gitnexus\embedding-runtime` word-splits — npm gets `--prefix C:\Users\First` plus a bogus extra spec; auto-heal and `gitnexus embeddings install` fail (or install to the wrong place) for every Windows user with a space in their profile path, which is a very common default. (b) Outside double quotes, cmd.exe consumes `^`, silently turning `@huggingface/transformers@^4.1.0` into an exact pin — contradicting this module's own no-drift guarantee (lines 43-45). (c) A hostile `GITNEXUS_EMBEDDING_RUNTIME_DIR` (e.g. `C:\x & calc.exe & rem`) is shell-interpreted — env-var-gated command injection.

**Fix:** quote args containing whitespace/cmd metacharacters when the shell is enabled (double quotes suppress `^` processing), or avoid the shell entirely by resolving npm's `npm-cli.js` and spawning `process.execPath` (also immune to the CVE-2024-27980 `.cmd` hardening). Add a builder/spawn unit test with a spaced prefix.

*Corroboration: 5 Claude lanes (adversarial, correctness, project-standards, security, security-boundary); mechanism demonstrated, no Windows host available.*

gitnexus/src/core/embeddings/runtime-install.ts
```diff
@@ -0,0 +1,212 @@
...
+ * Install (or update) the embedding stack into the runtime prefix via the
+ * user's npm — registry, mirror, and proxy configuration all apply. Rejects
+ * with npm's tail output when the install fails.
+ */
+export const installEmbeddingRuntime = async (
+  opts: EmbeddingInstallOptions = {},
+): Promise<void> => {
+  const { args, env } = buildEmbeddingInstallCommand(opts);
+  await new Promise<void>((resolve, reject) => {
+    const child = spawn('npm', args, {
```
magyargergo: **[P1] The npm child has no timeout, no parent-exit cleanup, and the `close` handler drops the `signal` argument — a blackholed proxy (this PR's exact audience) hangs `gitnexus analyze` forever** [reproduced]

Reproduced with a fake hanging `npm` on PATH: `installEmbeddingRuntime()`'s promise never settles (still pending after the harness's full wait), and when the parent exits mid-install the npm child **survives as an orphan** still writing into the shared `~/.gitnexus/embedding-runtime`. Notes: (1) no `timeout`/`AbortSignal` on the spawn and no deadline at either call site — `analyze --embeddings` behind a firewall that DROPs instead of rejecting prints "Downloading it now …" then freezes; in CI there is no diagnostic at all; (2) `analyze.ts` registers its SIGINT handler only *after* this block; (3) `child.on('close', (exitCode) => ...)` ignores the second `(code, signal)` argument, so an OOM-killed/signal-killed install reports a bare `exit null`.

**Fix:** mirror the repo's own precedent — `installDuckDbExtensionOutOfProcess` (`core/lbug/extension-loader.ts:81-160`) uses an env-overridable `setTimeout` + SIGKILL + settle-guard. Add a generous, env-overridable deadline (the stack is large; minutes, not 15s), kill the child on `exit`/`SIGINT`/`SIGTERM`, and fold `signal` into the failure message so the CLI can say "timed out — check your proxy" instead of a generic failure.

*Corroboration: Claude ce-reliability (reproduced, rated P0) + Codex B2 reliability (independent, static). Synthesis carries it as P1 rather than P0 because the forever-hang requires DROP-not-reject proxy behavior (many proxy configs fail fast via npm's own fetch retries) — but it remains squarely in this PR's target environment. An interrupted install also feeds the partial-prefix trap flagged on `resolveEmbeddingRuntime` below.*

gitnexus/src/core/embeddings/runtime-install.ts
```diff
@@ -0,0 +1,212 @@
...
+const prefixRequire = () => createRequire(join(getEmbeddingRuntimeDir(), 'noop.js'));
+
+/**
+ * Where the embedding stack resolves from, or `null` when it is not installed
+ * at all. Resolution only — nothing is imported, so this never loads native
+ * code and is safe on every platform.
+ */
+export const resolveEmbeddingRuntime = (): EmbeddingRuntimeResolution | null => {
+  try {
+    require.resolve('@huggingface/transformers');
```
magyargergo: **[P2] `resolveEmbeddingRuntime()` probes only `@huggingface/transformers` — a partial or stale prefix suppresses the self-heal and makes every diagnostic lie** [code-read; downstream matcher behavior reproduced]

Both probe tiers resolve only `'@huggingface/transformers'`; `onnxruntime-node` is never checked, and no version comparison against `getEmbeddingStackSpecs()` happens. Trigger: Ctrl-C/ENOSPC/kill mid-`npm install` (there is no timeout or lock — see the spawn comment) leaves transformers extracted but onnxruntime-node absent. Next `analyze --embeddings`: the gate at `analyze.ts:1105` sees non-null → **auto-heal skipped**; the embedder then fails with `Cannot find package 'onnxruntime-node'` → the (reproduced) matcher fires and tells the user analyze "does this automatically" — which it now never will. Loop. Meanwhile `doctor` reports local embeddings supported (same single-package probe via `isLocalEmbeddingStackInstalled`). Stale-version drift after a future gitnexus upgrade behind the same proxy has the same shape: old prefix silently satisfies the gate forever. Manual `gitnexus embeddings install` does escape (only `source === 'package'` short-circuits), but nothing tells the user that.

**Fix:** probe both `@huggingface/transformers` AND `onnxruntime-node` (treat a partial prefix as `null` — reinstall is idempotent and cheap), and consider a small manifest written at install time for a staleness check or warning.

*Corroboration: strong cross-engine — Claude ce-adversarial + risk-architect + Codex B3 adversarial independently.*

gitnexus/src/cli/embeddings.ts
```diff
@@ -0,0 +1,59 @@
...
+      recoveryHint: 'local-embedding-stack-missing',
+    });
+    process.exitCode = 1;
+    return;
+  }
+
+  cliInfo(
+    resolveEmbeddingRuntime() !== null
+      ? '✓ Embedding runtime installed. `gitnexus analyze --embeddings` is ready.'
+      : '✗ Install completed but the stack still does not resolve — check the output above.',
```
magyargergo: **[P2] Post-install verification can mislead in both directions: the `✗` failure path exits 0, and the `✓` is a false positive on Node 22.0–22.14** [code-read]

(1) When the post-install check prints `✗ Install completed but the stack still does not resolve`, the command still exits with code 0 — scripts and CI wrappers see success on a failed heal. Set `process.exitCode = 1` on this branch.

(2) Once the P0 import crash is fixed with a namespace import, on Node 22.0–22.14 this command will deposit packages into the prefix, `resolveEmbeddingRuntime()` (CJS `createRequire` — version-agnostic) will report `runtime-prefix`, and this line prints `✓ … ready` — but the actual ESM `import('@huggingface/transformers')` in the embedders will still fail, because the `registerHooks` fallback cannot be registered below 22.15. The success message should be gated on (or at least caveated by) hook availability, or the command should warn explicitly on Node < 22.15 that self-heal requires 22.15+.

*Corroboration: cross-engine — Codex B2 reliability (exit-code half, explicit); Claude docs-dod (false-positive half, explicit independent finding); Codex B1 (false-positive half implicit via its namespace-import nuance).*

gitnexus/src/core/embeddings/runtime-install.ts
```diff
@@ -0,0 +1,212 @@
...
+import { join } from 'node:path';
+import { logger } from '../logger.js';
+
+const require = createRequire(import.meta.url);
+
+/** The stack the runtime prefix provides; resolution fallback covers all three. */
+const EMBEDDING_STACK_PACKAGES = [
+  '@huggingface/transformers',
+  'onnxruntime-node',
+  'onnxruntime-common',
```
magyargergo, line 77: **[P2] Including `'onnxruntime-common'` in this hook's allowlist lets a stale runtime prefix hijack the #307 version-pairing hook** [reproduced]

Because this hook is registered first (correctly, as the last-resort fallback), it sits *below* the #307 common-resolver in the chain — so the #307 hook's initial `nextResolve` reaches this one **before** its own `catch` (whose entire purpose is pairing `onnxruntime-common` with `getEffectiveOnnxRuntimeNodeDir`'s exact onnxruntime-node copy) can fire. Reproduced with the PR's real `ensureEmbeddingStackResolvable()` via tsx on Node 22.16: with a populated prefix, the phantom `onnxruntime-common` import resolved to the **prefix copy** and the #307-mimic hook never fired; with an empty prefix, #307 fired as designed.

Trigger: `~/.gitnexus/embedding-runtime` is user-global, machine-wide state — created by any past heal or by `embeddings install --force` on a healthy install. Later, `pnpm dlx gitnexus analyze --embeddings` (pnpm-strict layout: transformers + onnxruntime-node fine from the store, phantom `onnxruntime-common` unresolvable) pairs the store's onnxruntime-node with the prefix's possibly version-drifted onnxruntime-common — exactly the mixed-copy/`Tensor`-identity class #307/#2069 exist to prevent.

**Fix:** only re-anchor `onnxruntime-common` when the stack itself is prefix-sourced (e.g. gate on `resolveEmbeddingRuntime()?.source === 'runtime-prefix'`), leaving #307 as the sole owner of the phantom-import case for package-sourced stacks.

*Corroboration: single lane (Claude ce-adversarial) — but reproduced end-to-end with the PR's actual hook code, which is why it is posted inline.*

magyargergo (commented):
# Tri-review digest — PR #2372 (`fix: proxy-blocked installs survive onnxruntime-node postinstall and self-heal embeddings`, #2370)

Head `ec06664a` · base `main` (merge-base `187c162f`) · two engines, 16 review lanes. This is an automated multi-tool synthesis; findings below are weighted by **cross-engine corroboration** and per-finding re-verification against the worktree, not by raw lane count.

---

## 1. Engine breakdown and leg liveness

**Claude engine (13 lanes, all completed):**
- 6 GitNexus PR-swarm personas: facts-historian, branch-hygiene, risk-architect, test-ci-verifier, security-boundary, docs-dod.
- 7 ce-code-review personas: correctness, adversarial, reliability, security, testing, maintainability, project-standards.
- Several lanes executed reproductions (including downloading and running a **real node v22.14.0 binary**, running the PR's actual hook code via tsx, fake-npm hang harnesses, and npm 10.9.2 `--prefix`/`.npmrc` experiments).

**Codex GPT-5.5 engine (3 legs):**
- **B1 codex-swarm: ran fully** (solo swarm). Sandbox blocked all `node` execution (`bwrap` namespace error), so its analysis is static-only; it also could not see live GitHub state. Note: the first B1 attempt crashed on a gitnexus-MCP schema/API incompatibility and was relaunched without MCP.
- **B2 codex-ce: DEGRADED.** Only 3 personas completed (maintainability, reliability, testing). Correctness, project-standards, and api-contract disconnected mid-run; security and adversarial never started. Treat B2's coverage as partial.
- **B3 codex-adversarial: ran fully** (static-only for the same sandbox reason).

Corroboration labels used below: **Strong** = independent agreement across both engines, or reproduced end-to-end; **Moderate** = same-engine multi-lane agreement; **Weak** = single lane.

---

## 2. Credit where due — what the review swarm probed and found solid

Validation is a feature; these were attacked from multiple angles and survived:

- **Hook registration order is correct in BOTH embedders** (`src/core/embeddings/embedder.ts:115-124`, `src/mcp/core/embedder.ts:78-88`): stack-prefix hook registered first → last-resort in `registerHooks`' most-recent-first chain; CUDA-13 hook registered last → runs first. Probed independently by risk-architect, facts-historian, ce-correctness, ce-adversarial. The suspected order bug was **refuted**.
- **Package-first resolution holds**: a healthy package install beats a stale prefix at every layer for `@huggingface/transformers` and `onnxruntime-node` (one exception, `onnxruntime-common` — see inline finding 6).
- **#1515 two-embedder parity**: core and MCP embedders wired identically (same `ensure*` ordering, same `import().catch` conversion, both reset init state on failure so in-process retry after healing works). The recurring one-sided-fix bug class did not recur.
- **optionalDependencies move is correct**: `onnxruntime-common` stays a regular dep (the #307 phantom-import fallback needs it); lockfile delta is minimal and coherent (transitive subtree flagged `optional`, no drift); `npm prune --omit=dev` does not prune optionals, so Docker images are unaffected.
- **Malicious-repo `.npmrc` registry-redirect attack REFUTED [reproduced]** on npm 10.9.2: with `--prefix`, npm reads project config from `<prefix>/.npmrc`, not the analyzed repo's cwd. The end-to-end "attacker repo → attacker tarballs → in-process load" chain is not reachable on bundled npm. (Residual exposure only for legacy npm v6 — see §5 hardening note.)
- **Command-injection via package specs REFUTED**: specs come from gitnexus' own manifest via `createRequire(import.meta.url)`, never from the analyzed repo; `--ignore-scripts` covers all transitive deps; the hook intercepts only a hardcoded 3-specifier allowlist and only on `ERR_MODULE_NOT_FOUND`/`ERR_PACKAGE_PATH_NOT_EXPORTED`.
- **The pruned-stack matcher is well designed [reproduced]**: #1515-style path-form binding errors, musl relocation failures, corrupt-subpath ESM misses, and the Node-22.x link SyntaxError all correctly do NOT match; the quote-delimited bare-specifier match fires on the real pruned case, including failures re-anchored at the prefix.
- **i18n complete** (en.ts + zh-CN.ts + help-i18n.ts + `allHelpCommands`, genuine translations); **CHANGELOG untouched** (correct per repo policy); **no `any`/`as any`**, no deferred work, conventional-commit title, clean whitespace/bidi scans (both engines); branch hygiene clean (2 linear commits, in-scope files only).
- **`getEmbeddingStackSpecs` manifest path verified for the published dist layout** (`dist/core/embeddings/../../../package.json` = package root), and a drift-guard unit test pins the specs.
- **Doctor DI (`stackInstalled` param) + tests correct; MCP server deliberately never auto-installs** (no user terminal) — right call.

**Branch hygiene classification: `clean feature/fix PR`** — 2 linear commits on the merge-base, all 21 files in-scope for #2370, no unrelated churn, clean whitespace/bidi scans.

---

## 3. Headline finding (merged across ~9 Claude lanes + all 3 Codex legs)

### P0 — Static named `registerHooks` import: ESM link-time SyntaxError on Node 22.0–22.14 breaks `gitnexus analyze` (every invocation) and the self-heal command itself — **confirms @koriyoshi2041's comment** (issuecomment-4884449781)

**Corroboration: Strong** (Codex B1 + B2 + B3 and Claude facts-historian, risk-architect, test-ci-verifier, branch-hygiene, ce-correctness, ce-adversarial, ce-testing, ce-project-standards, ce-maintainability all reached it independently). **[reproduced]** — ce-correctness downloaded and ran a real `node v22.14.0` binary:
- `typeof require('node:module').registerHooks` → `undefined` on 22.14.0 (API landed in 22.15.0, Apr 2025).
- Static `import { createRequire, registerHooks } from 'node:module'` in ESM → `SyntaxError: The requested module 'node:module' does not provide an export named 'registerHooks'` at `ModuleJob._instantiate` — **link time, before any statement runs**. The `typeof registerHooks !== 'function'` guard at `runtime-install.ts:109` is unreachable dead code.

**Three files carry the pattern** (all re-verified in the worktree):
- `gitnexus/src/core/embeddings/runtime-install.ts:21` — **NEW in this PR**.
- `gitnexus/src/core/embeddings/onnxruntime-node-resolver.ts:54` — pre-existing (PR #2341).
- `gitnexus/src/core/embeddings/onnxruntime-common-resolver.ts:60` — pre-existing (PR #2341, not in this PR's diff).
Consequence: #2341's "graceful degradation on Node < 22.15" guards (`runtime-install.ts:109`, `onnxruntime-node-resolver.ts:204/298`, `onnxruntime-common-resolver.ts:104`) have **never been reachable** in the shipped ESM dist on 22.0–22.14.

**Regression vs base (verified against `git archive 187c162f` import closures):** at base, plain `gitnexus analyze` had a clean static closure and **worked** on 22.0–22.14; `doctor` was already broken (pre-existing). At HEAD, `cli/analyze.ts` statically imports `runtime-install.js` (directly and via `runtime-support.ts`), so **every `gitnexus analyze` invocation — no `--embeddings` needed — dies at lazy-action import time**, as does the new `gitnexus embeddings install` (the remedy command itself). `cli/index.ts:444` uses `program.parse(...)` (not `parseAsync`), so the rejection surfaces as an unhandled-rejection dump of the raw SyntaxError with no mention of Node versions; `getMissingLocalEmbeddingStackMessage` correctly does NOT match it (`code: undefined`), so there is no misdiagnosis — but also no actionable message. MCP semantic search silently degrades (pre-existing swallow). `engines` floor is `>=22.0.0` (`gitnexus/package.json`), so this range is contractually supported (CONTRIBUTING.md:16).

**Why CI is green:** `actions/setup-node` uses `node-version: 22` → resolves to latest 22.x (≥22.15). And **vitest/tsx structurally mask this class [reproduced]**: ce-testing ran the exact import shape under the repo's vitest — passes; ce-correctness ran the same named import under tsx on node 22.14 — loads fine with `registerHooks === undefined`. Only real Node ESM linking of the built dist fails. No test in the suite can catch it.

**Fix options (both engines converge, with two nuances):**
1. **Namespace import** (`import * as nodeModule from 'node:module'`; guard `typeof nodeModule.registerHooks !== 'function'`) in **all three files** — **verified on real node 22.14.0**: links fine, member is `undefined`, the existing guards finally become functional. Nuance (a) — raised by Codex B1 and Claude docs-dod: the namespace import alone stops the crash but **self-heal still cannot work on 22.0–22.14** (no hooks API), and worse, `gitnexus embeddings install` would deposit packages, `resolveEmbeddingRuntime()` (CJS `createRequire`) reports `runtime-prefix`, and `cli/embeddings.ts:54-58` prints a **false-positive** "`✓ … ready`" while the later ESM `import('@huggingface/transformers')` fails — so pair the fix with an explicit Node-version message or a non-hook fallback (see inline finding 5). Nuance (b) — risk-architect: `@types/node` types `registerHooks` as always-present, so under the repo's no-`any` policy the fix needs a typed optional narrow (e.g. `const rh: typeof nodeModule.registerHooks | undefined = (nodeModule as Partial<typeof nodeModule>).registerHooks`).
2. **Raise `engines` to `>=22.15.0`** — simpler and honest about the self-heal dependency (Codex B1's preference), but breaking for existing 22.0–22.14 installs, npm only warns on engines mismatch by default, and it requires lockstep updates to CONTRIBUTING.md:16 + both READMEs.

Either way, add a **dist-level smoke test under the engines floor** (e.g. `node --input-type=module -e "await import('./dist/core/embeddings/runtime-install.js')"` on a 22.14 matrix leg) — vitest cannot cover this class. One lane (branch-hygiene) suggested deferring to a post-merge hotfix; that conflicts with this repo's fix-in-PR rule and is outvoted by every other lane on both engines: **fix in this PR**.

---

## 4. Inline findings (posted as review comments)

| # | File:line (head) | Sev | Finding | Tag | Corroboration |
|---|---|---|---|---|---|
| 1 | `gitnexus/src/core/embeddings/runtime-install.ts:21` | P0 | registerHooks named import → ESM link error on Node 22.0–22.14 (full picture in §3) | [reproduced] | Strong: 3/3 Codex legs + 9 Claude lanes |
| 2 | `gitnexus/src/core/embeddings/runtime-install.ts:192` | P1 | Windows `shell: true` joins args unquoted: spaced profile paths split `--prefix`; cmd.exe eats `^` in semver ranges; env-var path is shell-interpreted | [reproduced mechanism / code-read win32] | Moderate-strong: 5 Claude lanes (adversarial, correctness, project-standards, security, security-boundary); not reached by Codex |
| 3 | `gitnexus/src/core/embeddings/runtime-install.ts:187` | P1 | npm child has no timeout, no parent-exit cleanup, and `close` drops the `signal` arg — blackholed proxy (the PR's own audience) hangs `analyze` forever | [reproduced] | Strong: Claude ce-reliability (reproduced) + Codex B2 reliability |
| 4 | `gitnexus/src/core/embeddings/runtime-install.ts:74` | P2 | `resolveEmbeddingRuntime()` probes only `@huggingface/transformers` — a partial/stale prefix suppresses self-heal and misleads every diagnostic | [code-read + matcher half reproduced] | Strong: Claude ce-adversarial + risk-architect + Codex B3 |
| 5 | `gitnexus/src/cli/embeddings.ts:57` | P2 | Post-install verification: `✗` failure path leaves exit code 0; `✓` is a false positive on Node 22.0–22.14 (CJS probe vs ESM load) | [code-read] | Strong: Codex B2 reliability + B1 + Claude docs-dod |
| 6 | `gitnexus/src/core/embeddings/runtime-install.ts:34` | P2 | `'onnxruntime-common'` in the prefix-hook allowlist sits below the #307 pairing hook and hijacks its phantom-import case with a possibly version-drifted prefix copy | [reproduced] | Weak (single lane, ce-adversarial) but reproduced end-to-end with the PR's real hook code |

Full comment bodies are in the JSON section at the end. Note `onnxruntime-common-resolver.ts` is not in this PR's diff, so its pre-existing copy of the P0 pattern is referenced in comment text rather than anchored.

---

## 5. Lower-priority items, refuted suspicions, and the test-coverage story

### Body-only findings (real but lower severity, weaker corroboration, or not diff-anchorable)

- **P2/P3 — MCP surface: the new actionable message is dead-lettered.** `local-backend.ts:2402-2405` (`catch { return [] }`, pre-existing, not in diff) swallows the `getMissingLocalEmbeddingStackMessage` error that `mcp/core/embedder.ts` carefully constructs — semantic search silently degrades to BM25 on a pruned install, which is the exact silent-degradation mode #2370 exists to fix. Suggested: `logger.warn` once per process in the catch when the message matches. (Two Claude lanes: ce-adversarial, ce-correctness.)
- **P3 — `--cuda` doesn't clear an inherited `ONNXRUNTIME_NODE_INSTALL=skip`** (`runtime-install.ts:171-173`): a user who exported the skip var per this PR's own guidance and later runs `embeddings install --cuda` gets a scripts-enabled install whose postinstall still skips the NuGet download, then a `✓ ready` (post-check only tests resolvability). Fix: `delete env.ONNXRUNTIME_NODE_INSTALL` in the cuda branch. (Single lane, ce-correctness, mechanically solid.)
- **P2 (conf 55) — CUDA-13 redirect can't use the prefix's top-level `onnxruntime-node` after `embeddings install --cuda`**: the new prefix fallback was added to `resolveDefaultOrtNodeDir` (`onnxruntime-node-resolver.ts:158-172`) but not `resolveOurOrtNodeDir` (:176-182) — in the pruned+healed CUDA-13 state, `ourDir` is `null` → no redirect → GPU binaries downloaded but embeddings run on CPU, command reports success. (Single lane, ce-correctness; depends on pin/range split.)
- **P2 — no lock on the shared prefix; concurrent installs unsynchronized** (two `analyze --embeddings` runs, or analyze + manual install). Cross-lane tension acknowledged: risk-architect judged npm's own locking adequate (P3), ce-reliability's lightweight concurrent-install experiment saw no corruption but could not exercise the heavy real stack. Advisory mkdir-sentinel + post-wait re-check recommended, not blocking.
- **P3 — relative or empty `GITNEXUS_EMBEDDING_RUNTIME_DIR` → `createRequire` throws inside the probe → perpetual "not installed" and reinstall-every-run split-brain** ([reproduced] for the throw). Fix: `path.resolve()` once in `getEmbeddingRuntimeDir()`.
- **P3 — analyze auto-install UX**: runs with no `onOutput` (npm silent until failure) and before the cheap flag-conflict validations; a doomed `--repair-fts --force` invocation can do a multi-minute download first.
- **Security hardening (non-blocking)**: the analyzed-repo `.npmrc` attack was refuted on npm 10 [reproduced], but `cwd: homedir()` on the spawn (or `--userconfig`) removes the residual legacy-npm-v6 class for two tokens. Windows env-var command injection via `GITNEXUS_EMBEDDING_RUNTIME_DIR` + `shell:true` is covered by inline finding 2's fix. npm output tail (2000B) lands in error messages/pino logs — sanitize `://user:pass@` patterns if paranoid.
- **Docs (P2/P3)**: `GITNEXUS_EMBEDDING_RUNTIME_DIR` is user-facing but appears in neither README; the Node ≥ 22.15 self-heal requirement is documented nowhere user-visible (five surfaces need lockstep edits if engines moves); `gitnexus embeddings install` is missing from the README "CLI Commands" block.
- **Maintainability (P2s)**: the fragile registerHooks pattern is now duplicated in two files instead of centralized behind one guarded accessor; the ordering-sensitive 4-statement embedder wiring block is inlined verbatim in both embedders against the codebase's own shared-module convention (#1515 risk class); the `noop.js` anchor convention is hard-coded at three sites across two files (`prefixRequire` deserves an export); `analyzeCommandImpl` (1647 → 1695 lines) grew another inline orchestration block.

### Refuted / probed-and-solid (beyond §2)

- "ESM link error gets misdiagnosed as missing-stack" — **refuted [reproduced]** (`code: undefined` fails the matcher's code check; propagates raw).
- "Doctor probe breaks on ESM-only transformers / missing `./package.json` export" — **refuted [reproduced]** (CJS `require.resolve` of the bare specifier succeeds).
- "`require('../../../package.json')` wrong depth in dist" — **refuted** (rootDir strip verified).
- "Flag→env ordering breaks `isHttpMode()` at the auto-install gate" — **refuted** (flags written at analyze.ts:1024/1034, gate at :1105).
- "Package-location partial state deadlocks `embeddings install`" — **mostly refuted for npm** (lockfile shows transformers' `onnxruntime-node` is a regular dep — pruned together); residual for yarn/pnpm globals, `--force` is the escape hatch.
- "Spec drift vs lockfile" — **refuted as non-issue** (global installs also resolve manifest ranges fresh; matches the module's stated invariant).
- Double hook registration across the two embedders, hook leakage to unrelated specifiers, `ERR_PACKAGE_PATH_NOT_EXPORTED` handling, TOCTOU check-vs-load, SSRF, doctor-probe safety on macOS Intel, `registerHooks`-throwing degradation, HTTP-mode gating — all probed clean.

### Test-coverage story (consistent across ce-testing, test-ci-verifier, Codex B2-testing)

The pure helpers are genuinely well-tested (`buildEmbeddingInstallCommand` shapes, matcher positive/negative incl. the #1515 path-form negative, doctor DI, spec drift-guard, i18n extraction). But **every I/O path is untested**: `installEmbeddingRuntime`'s spawn flow (exit≠0, ENOENT, tail assembly), `ensureEmbeddingStackResolvable` (redirect/passthrough/rethrow/idempotency/registration order — the sibling resolver suites have an established `vi.doMock('node:module')` pattern that was not reused), the `runtime-prefix` resolution tier (the string `'runtime-prefix'` appears only in the source file), `embeddingsInstallCommand` (all four outcomes), and the analyze auto-install branch (2×2×2 trigger matrix). And structurally: **vitest masks the entire ESM link-error class [reproduced both directions]** — the suite's "graceful degradation" tests mock an `undefined` binding that real Node ESM would never deliver. One hygiene note: `embedding-runtime-support.test.ts` leaks real, never-deregistered global resolution hooks into shared vitest workers.

---

## 6. CI status on head

**Merge state classification: `checks failing`** — one check is failing: Vercel "Authorization required". That check is fork-PR infra noise, not code (`mergeable=MERGEABLE`, `UNSTABLE` solely from it); every functional check is green (coverage suite 14005 passed / 56 skipped, win/mac platform-sensitive suites, packaged smokes ubuntu+windows, CodeQL, Docker). Important caveat: CI runs `node-version: 22` → latest 22.x, which **masks the P0 entirely**; no leg covers 22.0–22.14, and the packaged-install smokes don't exercise `embeddings install` or the auto-heal.

---

## 7. Coverage limits of this review

- The worktree is strictly read-only: nobody ran `npm run build`, so the tsc-emitted dist itself was not executed (NodeNext emits the static imports 1:1; closure analysis was done on sources with type-only imports elided).
- Both Codex legs that completed ran **static-only** — the sandbox blocked every `node` invocation (`bwrap` namespace error) — and could not see live GitHub state.
- Codex B2 (ce personas) was degraded: correctness/project-standards/api-contract disconnected, security/adversarial never started.
- No Windows host was available; the `shell:true` findings verify the arg-joining mechanism on Linux plus documented cmd.exe semantics, not an end-to-end Windows run. macOS Intel (#1515) paths are unchanged and were not exercised.

---

## 8. Final verdict

**not production-ready**

The PR's intended fix — surviving proxy-pruned onnxruntime installs and self-healing embeddings — is sound in design, and much of the implementation withstood adversarial probing (§2). But the static named `registerHooks` import at `runtime-install.ts:21` introduces an ESM link-time SyntaxError on Node 22.0–22.14 that crashes `gitnexus analyze` — the flagship command — on **every** invocation, `--embeddings` or not. That is a regression against base (where plain `analyze` worked on those versions) and against the published `engines: >=22.0.0` contract, and the remedy command `gitnexus embeddings install` dies identically — making the PR counterproductive for the exact user population it targets on those Node versions. Additionally, `installEmbeddingRuntime` has no spawn timeout (a blackholed proxy hangs `analyze` forever — reproduced) and a Windows shell-quoting flaw breaks every spaced-profile-path user. Before merging: fix the `registerHooks` named import in all three affected files (or raise engines with lockstep doc updates), add a spawn timeout mirroring the repo's own DuckDB extension-loader precedent, quote the Windows spawn args, and add a dist-level smoke on a 22.14 matrix leg so this class can't pass CI silently again.

## 9. Closing

This is an automated two-engine, 16-lane review digest — verify findings before acting on them; severities are the synthesis' judgment, not gospel. (Format note: this digest intentionally uses the tri-review synthesis format for the GitHub-comment context; the swarm's internal 12-section framework lives in the underlying lane artifacts.)

Explicit credit to **@koriyoshi2041** (issuecomment-4884449781): the Node 22.0–22.14 `registerHooks` link-error report is **confirmed** — reproduced on a real node v22.14.0 binary, blast radius verified as a NEW regression for plain `gitnexus analyze` (clean at base), and the proposed namespace-import fix verified to link on 22.14 with functional guards. Two nuances for whichever fix is chosen: (a) namespace import alone still leaves self-heal non-functional below 22.15 and lets `embeddings install` print a false-positive success there (CJS probe vs ESM load), so pair it with an explicit version message or engines decision; (b) `@types/node` declares `registerHooks` as always-present, so the guard needs a typed optional narrow to stay within the repo's no-`any` policy.

koriyoshi2041:
Verified the current head (`3df925c7`) against the Node floor issue I flagged earlier.

Checks I ran locally:

```bash
npm run build
npx -y node@22.14.0 --input-type=module -e "await import(\"./dist/core/embeddings/runtime-install.js\"); await import(\"./dist/core/embeddings/onnxruntime-node-resolver.js\"); await import(\"./dist/core/embeddings/onnxruntime-common-resolver.js\"); const m = await import(\"./dist/core/embeddings/runtime-install.js\"); console.log(\"isPrefixRuntimeLoadable=\" + m.isPrefixRuntimeLoadable())"
npx vitest run test/unit/node-module-compat.test.ts test/unit/embedding-runtime-support.test.ts test/unit/embeddings-install-command.test.ts
```

Result: the Node 22.14 dist imports link cleanly and print `isPrefixRuntimeLoadable=false`; the focused tests pass (`34 passed`). This resolves my earlier `registerHooks` link-time concern from the supported 22.0-22.14 range.

PR #726, opened 2026-04-08
fix(fts): try local LOAD before INSTALL to avoid network failures

pratyush618:
## Summary

Follow-up to #781 (which landed the `loadFTSExtension` LOAD-first reorder). This PR extends that fix to the two places it was missed:

- **`pool-adapter.ts`** — `initPool` and `initLbugWithDb` only did `LOAD EXTENSION fts` and silently swallowed failures. If the extension was never installed (no prior `gitnexus analyze` run), FTS queries failed at runtime with no diagnostic. Both paths now call `loadFTSExtension(available[0])` so the core adapter's LOAD→INSTALL fallback applies here too.
- **`lbug-adapter.ts`** — `loadFTSExtension` now accepts an optional `Connection` and returns a success boolean, letting the pool adapter track `shared.ftsLoaded` without duplicating the error-handling chain.

Fixes the residual case from #712 for MCP pool users who haven't run `gitnexus analyze` locally.

**Contract change:** `loadFTSExtension` return type changed from `Promise<void>` to `Promise<boolean>`; existing callers that discard the result are unaffected.

## Test plan

- [x] TypeScript compiles cleanly (`npx tsc --noEmit`)
- [x] `lbug-core-adapter` integration tests pass — now includes a targeted test for the `loadFTSExtension(conn)` path on a fresh Connection
- [x] `search-pool` / `search-core` integration tests pass
- [ ] Verify MCP pool path: connect against a repo where FTS extension was never installed — should now install + load instead of failing silently

PR #2375, opened 2026-07-05
fix: surface real FTS extension LOAD errors and self-heal broken extension files (#2374)

### LINKED ISSUE #2374: FTS exists, but it still reports that LadybugDB does not exist.
### Area

### Summary

gitnexus analyze

  GitNexus Analyzer

  ████████████████████████░░░░░░░░░░░░░░░░  60% | Loading into LadybugDB...{"level":40,"time":1783251883187,"name":"git}
  Repository indexed successfully (31.1s)░  85% | Creating search indexes... (14s){"level":40,"time":1783251900774,"nam}
FTS extension unavailable; skipping search-index creation. Full-text/BM25 search will be disabled until the LadybugDB F.
  6,417 nodes | 16,507 edges | 247 clusters | 300 flows
  F:\McMod\RTSbuilding-1.7.10-port

  Warning: full-text/BM25 search is disabled — the LadybugDB FTS extension was unavailable.
  Install it once with network access (GITNEXUS_LBUG_EXTENSION_INSTALL=auto) then rerun, or
  run `gitnexus analyze --repair-fts` when connected. Run `gitnexus doctor` for details.

### Context

_No response_

### Expected behavior

the index is built successfully.

### Actual behavior

the index is missing when use gitnexus query

### Steps to reproduce

install gitnexus 1.6.9
download ladybug 0.18.0
add libfts.lbug_extension(version 0.18.0) under .lbdb/extension

### Environment

win 11 25h2

### Logs / screenshots

Microsoft Windows [版本 10.0.26200.8246]
(c) Microsoft Corporation。保留所有权利。

F:\McMod\RTSbuilding-main>gitnexus analyze

  GitNexus Analyzer

  ████████████████████████░░░░░░░░░░░░░░░░  60% | Loading into LadybugDB...{"level":40,"time":1783252318994,"name":"git}
  Repository indexed successfully (23.6s)░  85% | Creating search indexes...{"level":40,"time":1783252323108,"name":"gi}
FTS extension unavailable; skipping search-index creation. Full-text/BM25 search will be disabled until the LadybugDB F.
  9,802 nodes | 33,344 edges | 537 clusters | 300 flows
  F:\McMod\RTSbuilding-main

  Warning: full-text/BM25 search is disabled — the LadybugDB FTS extension was unavailable.
  Install it once with network access (GITNEXUS_LBUG_EXTENSION_INSTALL=auto) then rerun, or
  run `gitnexus analyze --repair-fts` when connected. Run `gitnexus doctor` for details.

F:\McMod\RTSbuilding-main>gitnexus analyze --repair-fts

  GitNexus Analyzer

  ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░   0% | Initializing...{"level":40,"time":1783252344163,"name":"gitnexus","ms}
{"level":40,"time":1783252344328,"name":"gitnexus","msg":"GitNexus: FTS extension unavailable; continuing without FTS f}
  Analysis failed: Cannot repair FTS indexes: the LadybugDB FTS extension is unavailable (not pre-installed and could not be installed on this machine). Run `gitnexus doctor` to install it, then retry `--repair-fts`.
Error: Cannot repair FTS indexes: the LadybugDB FTS extension is unavailable (not pre-installed and could not be installed on this machine). Run `gitnexus doctor` to install it, then retry `--repair-fts`.
    at runFullAnalysis (file:///C:/Users/Administrator/AppData/Roaming/npm/node_modules/gitnexus/dist/core/run-analyze.js:368:23)
    at async analyzeCommandImpl (file:///C:/Users/Administrator/AppData/Roaming/npm/node_modules/gitnexus/dist/cli/analyze.js:943:24)
    at async analyzeCommand (file:///C:/Users/Administrator/AppData/Roaming/npm/node_modules/gitnexus/dist/cli/analyze.js:507:9)
    at async Command.<anonymous> (file:///C:/Users/Administrator/AppData/Roaming/npm/node_modules/gitnexus/dist/cli/lazy-action.js:33:9)

F:\McMod\RTSbuilding-main>gitnexus doctor
GitNexus Doctor

Runtime
  OS:       win32/x64
  Node:     v24.18.0
  GitNexus: 1.6.9
  LadybugDB:unknown
  native    ✓ lbugjs.node loaded
  ONNX:     1.27.0

Capabilities
  Graph store:      available
  Full-text search: available
  VECTOR index:     unavailable
  Semantic mode:    exact-scan
  Ext install:      load-only (offline; load only, no network install)
  Exact scan limit: 10000 chunks
  Note:             LadybugDB VECTOR is disabled on this platform; semantic search uses exact scan when embeddings exist.

Embeddings
  Backend:    local
  Device:     auto
  Threads:    4
  Batch:      16 nodes
  Sub-batch:  8 chunks
  Support:    ✓ local embeddings supported
  CUDA:       n/a (no system CUDA detected)

magyargergo:
Fixes #2374.

## The problem

A user on Windows manually placed a valid-looking `libfts.lbug_extension` at the correct `~/.lbdb/extension/0.18.0/win_amd64/fts/` path, yet:

- `analyze` reported *"FTS extension unavailable"* with no reason
- `analyze --repair-fts` failed claiming the extension is *"not pre-installed and could not be installed"* — factually wrong, the file was there
- the error said *"Run `gitnexus doctor` to install it"* — doctor has no install action
- `doctor` contradicted everything with a static `Full-text search: available` and printed `LadybugDB: unknown`

Reproduced end-to-end on main: the entire experience is byte-identical when the extension file is **present but unloadable** (truncated/wrong-platform download). Four root causes, all fixed here:

## Root causes and fixes

**1. The real LOAD error was discarded** — `ExtensionManager.tryLoad()` swallowed `err.message`. LadybugDB's errors are precise (`Failed to load library: <path> … file too short` / `invalid ELF header` vs `fts is an official extension and has not been installed`), and they distinguish a *missing* file from a *broken* one. `tryLoad` now returns the error and every unavailable-reason carries it — through the analyze warning, the repair error, query/API response warnings, and doctor.

**2. Plain `INSTALL` silently no-ops over a broken file** — so `GITNEXUS_LBUG_EXTENSION_INSTALL=auto` could never self-heal: install "succeeds", the corrupt file survives, LOAD fails forever. Kuzu's docs confirm: *"Running INSTALL again will not perform any action if the extension is already installed."* The out-of-process installer now runs `FORCE INSTALL` — always correct because it only executes after LOAD already failed, so any existing file is broken (`UPDATE <ext>` is the Kuzu-documented re-download verb; both behave identically on LadybugDB 0.18, verified empirically). The Docker bake path is unaffected (fresh layer, single download either way).

**3. `doctor` printed a static capability** — now live-probes `LOAD EXTENSION fts` on a throwaway in-memory DB (dynamic import, so doctor still runs when the native module itself is broken) and prints the real error on failure. The `LadybugDB: unknown` bug reproduced on *every* platform: `@ladybugdb/core`'s `exports` map omits `./package.json`, so the version require threw `ERR_PACKAGE_PATH_NOT_EXPORTED`; `packageVersion()` now falls back to resolving the entry point and walking up to the package root.

**4. Query responses misdiagnosed the broken-extension case** — `query` said *"FTS indexes missing"* even when the indexes existed but the extension failed to load. New shared `ftsDegradedWarning()` (MCP backend + HTTP API) prefers the live capability reason.

The #726/#1161 offline-first contract is preserved: `FORCE` only changes the SQL verb inside the already-gated `auto`-path child; serve/query read paths stay `load-only` and never touch the network.

## After this change (broken file, offline)

```
GitNexus: FTS extension unavailable; continuing without FTS features. load-only policy
(no install attempted); LOAD fts failed: IO exception: Failed to load library:
/home/user/.lbdb/extension/0.18.0/linux_amd64/fts/libfts.lbug_extension which is needed
by extension: fts. Error: …: file too short.
```

And the reporter's exact failing command — broken file + `auto` + `--repair-fts` — now re-downloads and prints `FTS indexes repaired successfully`.

## Testing

**New mock-free E2E suite** (`test/integration/fts-extension-e2e.test.ts`, 11 cases, ~90s): spawns the real CLI per `cli-e2e.test.ts` conventions; LadybugDB resolves `.lbdb` from `HOME`/`USERPROFILE`, so each scenario owns a hermetic fake home — the machine's real `~/.lbdb` is never touched, and the extension path is discovered, never hardcoded. Matrix: happy offline (analyze/query/doctor/repair-fts), broken-file (all four surfaces assert the real reason **and** that the old misleading strings are gone), missing-file (distinct `has not been installed` reason), and network-gated self-heal (broken → `auto` repair re-downloads, asserted by file size on disk; fresh-machine auto install). Assertion strings are LadybugDB's own platform-independent messages. **Mutation-verified**: reverting `FORCE INSTALL` back to `INSTALL` makes the self-heal case fail exactly as the user's machine did.

**Unit**: reason-threading through all three failure modes, `ftsDegradedWarning` selector, LadybugDB version resolution regression guard; updated repair-fts assertion. Full unit suite green (5 unrelated 30s-timeout load-flakes pass in isolation); FTS integration suites green; `tsc`/eslint/prettier clean.

Impact analysis: `ExtensionManager.ensure` is CRITICAL-radius (embedding pipeline, MCP query, analyze), but its public boolean contract and success paths are byte-identical — only failure reason strings gained detail; a repo-wide consumer sweep confirmed nothing parses the changed strings.


magyargergo (commented):
# Tri-review digest — PR #2375 "fix: surface real FTS extension LOAD errors and self-heal broken extension files (#2374)"

Head `a35e3044` · base `cdad478c` · worktree `/workspace/.worktrees/pr-2375`

**Engine breakdown:** Claude — 6 GitNexus swarm lanes (facts-historian, branch-hygiene, risk-architect, test-ci, security-boundary, docs-dod) + 8 ce-code-review personas (correctness, testing, maintainability, project-standards, adversarial, security, api-contract, reliability), all produced full reports. Codex GPT-5.5/xhigh — swarm (Solo, verdict + 2 structured findings), ce-code-review report-only (1 finding, confidence 100), native adversarial red-team (2 findings). **All three Codex legs live** (real findings, no error/skill-not-found markers). Self-review disclosure: the PR was authored by the reviewing account in a prior session; the memory briefing was treated as leads only, and the load-bearing claims were independently re-executed by lanes (details below).

---

## What's solid

- **The core premise was reproduced end-to-end, not just re-read.** ce-correctness executed the exact heal sequence against the bundled @ladybugdb/core 0.18.0 with a hermetic HOME and a corrupt `libfts.lbug_extension`: LOAD fails with `IO exception: Failed to load library …`; plain `INSTALL fts` returns OK but is a **silent no-op** (LOAD still fails — this is #2374's trap); `FORCE INSTALL fts` succeeds and the retry LOAD then succeeds, on the same connection, exactly as `ensure()` does it. ce-adversarial independently reproduced the doctor probe returning `{loaded:true}` on a valid home and correct, distinct reasons for broken ("file too short") vs missing ("has not been installed"). The author-session's "INSTALL no-ops over a broken file" and "FORCE INSTALL parses on 0.18.0" claims — flagged in the briefing as needing re-verification — are now independently verified [reproduced], and additionally validated by CI: the new e2e suite ran 11/11 green on the networked ubuntu runner, exercising the real FORCE INSTALL self-heal.
- **Tests executed:** ce-testing ran all six touched/added unit files (59/59 pass) plus the e2e `doctor` tests live (2/2, real live-probe reason surfacing); ce-correctness ran a targeted 39/39 vitest pass and `npx tsc --noEmit` (clean). `test/unit/lbug-extension-loader.test.ts` covers every `ensure()` branch with exact reason-string assertions; no `any`/`as any` or if-branching in tests (grep-verified, per repo rules).
- **Branch hygiene:** clean — single commit, 11 minutes off main, 17 files all causally tied to #2374, no bidi/whitespace issues (verified by both engines' hygiene passes), CHANGELOG correctly untouched, conventional-commit title.
- **A large set of plausible regressions was attacked and refuted** (see the validated/refuted list) — including the two contract-critical ones: offline-first read paths (#726/#1161) and SQL injection via `FORCE INSTALL`.
- The `exports`-gap fix (`LadybugDB: unknown`) was re-verified live: `require('@ladybugdb/core/package.json')` still throws ERR_PACKAGE_PATH_NOT_EXPORTED in this worktree, and `getRuntimeFingerprint()` now returns `0.18.0` (executed by two lanes).

---

## Findings for inline comments

### 1. P2 — e2e suite crashes in `beforeAll` instead of skipping on offline machines with no seed extension
- **File/line:** `gitnexus/test/integration/fts-extension-e2e.test.ts:203` (also 100, 241–244)
- **Problem:** the documented skip contract ("Skip everything (visibly) when no valid extension exists", lines 192–196) is a file-level `beforeEach(ctx.skip)`, but the `happy` and `broken` describes call `makeHome(...)` in **beforeAll**, which vitest runs first. `makeHome` does `path.join(home, extensionRelPath)` (line 100) with `extensionRelPath` (line 42) left `undefined` when `resolveSeedExtension()` finds no local copy and the network install fails → `TypeError: The "path" argument must be of type string` in two describes' hooks.
- **Trigger:** offline dev machine with no `~/.lbdb/extension/<ver>` copy, `GITNEXUS_REQUIRE_FTS` unset → 8 tests fail red with hook TypeErrors instead of skipping.
- **Evidence:** verified in worktree — line 42 `let extensionRelPath: string;` assigned only in `resolveSeedExtension()` success paths; line 100 `path.join(home, extensionRelPath)`; beforeAll at 202–205/241–244 vs beforeEach skip at 194–196. `path.join('/a', undefined)` TypeError reproduced by claude-ce-adversarial.
- **Fix:** early-return from the describe-level `beforeAll` when `!seedExtensionFile` (or default `extensionRelPath` / build homes lazily inside the skippable tests).
- **Corroboration:** STRONG cross-engine — codex-ce (conf 100), codex-adversarial, claude-ce-adversarial F4. **[reproduced]**

### 2. P2 — the Windows-motivated regression suite never runs on Windows/macOS CI, and runs unregistered in the parallel `default` vitest project
- **File/line:** `gitnexus/test/integration/fts-extension-e2e.test.ts:29` (the `import lbug from '@ladybugdb/core';` — the config files it should be registered in are not in the diff)
- **Problem:** the new suite spawns the real CLI (`spawnSync`, lines 22/77/144) and matches the SPAWN_CLI category exactly, but is absent from `scripts/cross-platform-tests.ts` (SPAWN_CLI list, lines 98–113 — verified by grep) despite TESTING.md:96 and the script's own header rule; it's also absent from any sequential vitest project (`lbug-db` include / `cli-e2e` / `default` exclude, TESTING.md:68 as-written, since line 29 imports the native module — though only for `lbug.VERSION`).
- **Actual harm (reconciled across lanes + CI):** the suite **does** run in CI — it landed in the `default` project wildcard and passed 11/11 on ubuntu/coverage this run. So this is NOT "coverage never runs." The two real harms: (a) Windows/macOS CI runs only the cross-platform subset (TESTING.md:85), so the regression suite for a **Windows-reported** bug (#2374, Windows 11 + user-placed extension file; LadybugDB C++ error-string assertions unverified on Windows) never executes on Windows CI — the windows job passed this run without running it; (b) 11 CLI-spawning tests (44s on this run) execute file-parallel with the rest of the suite, aggravating the repo's known under-load 30s-timeout flake class — this very CI run had an unrelated `hooks.test.ts` concurrency-burst flake fail the coverage job.
- **Fix:** add `test/integration/fts-extension-e2e.test.ts` to the SPAWN_CLI array in `scripts/cross-platform-tests.ts`, and register it in a sequential vitest project (`cli-e2e` fits best — CLI spawning, 600s timeouts) + the `default` exclude list. The suite already self-gates offline (modulo finding 1).
- **Corroboration:** STRONG cross-engine — claude test-ci (P1), project-standards PS-1 (P1)/PS-2, docs-dod F2, codex-swarm Lane-4 #1 (P2, high conf). Absence from both lists re-verified by synthesis. Severity adjudicated P2 (not the P1 "do not merge" framing: the suite runs and passed; not P3: the Windows gap is real for a Windows bug). **[code-read]**

### 3. P2 — the headline `INSTALL`→`FORCE INSTALL` fix has no deterministic regression guard; its only test can silently skip
- **File/line:** `gitnexus/scripts/install-duckdb-extension.mjs:54`
- **Problem:** grep confirms exactly one production occurrence of `FORCE INSTALL` and zero unit tests asserting the SQL verb. The only guard is the network-gated e2e describe (`fts-extension-e2e.test.ts:297–300`, `if (!networkAvailable) ctx.skip()`), a **soft silent skip** not covered by the `GITNEXUS_REQUIRE_FTS=1` hard-fail contract that protects the rest of the file. A revert to plain `INSTALL` (which no-ops over a broken file — the exact #2374 regression, empirically confirmed) could merge on any day `extension.ladybugdb.com` is unreachable from CI.
- **Trigger:** revert `FORCE INSTALL` → `INSTALL` + CI egress hiccup ⇒ green build, #2374 reintroduced.
- **Evidence:** verified in worktree: mjs:54 sole production site; e2e skip at 298–300 outside the REQUIRE_FTS beforeAll throw (164–170). CI note: the self-heal describe DID run and pass on this networked runner — the guard works when the network does; the structural skip risk is what remains.
- **Fix:** cheap offline unit test — inject/mock the `@ladybugdb/core` connection in `install-duckdb-extension.mjs` (or extract the SQL choice) and assert `conn.query` was called with `FORCE INSTALL fts`, matching the rigor already given to `getExtensionInstallChildProcessArgs`.
- **Corroboration:** STRONG cross-engine — claude-ce-testing F3 (conf 80) + codex-swarm Lane-4 #2 (medium-high). **[code-read]**

### 4. P2 — FORCE INSTALL re-downloads the extension on every `analyze` run when LOAD persistently fails for a non-file reason
- **File/line:** `gitnexus/scripts/install-duckdb-extension.mjs:49` (the comment "any existing extension file is broken" encodes the unsound assumption)
- **Problem:** LOAD can fail permanently for reasons a re-download can never fix (missing MSVC runtime / glibc-vs-musl mismatch, seccomp/noexec, upstream shipping a broken build for one platform). The install-outcome cache is per-process only (`installAttempted`, extension-loader.ts:240–245; every `gitnexus analyze` is a fresh process, default policy `auto`), so such machines now re-download ~2.2 MB with up-to-15 s latency on **every analyze run, forever** — pre-PR, plain INSTALL no-op'd instantly over the present file. Worst case (upstream broken build) is fleet-wide per-analyze load on the extension repo with no kill switch besides manually setting `GITNEXUS_LBUG_EXTENSION_INSTALL=load-only`.
- **Trigger:** any host where the freshly downloaded extension still can't dlopen; repeat `gitnexus analyze` (CI pipelines, auto-sync).
- **Evidence:** claude-ce-adversarial reproduced the mechanics: `FORCE INSTALL fts` over an already-**valid** file re-downloads and rewrites it in place (mtime changed, same inode, 2,225,576 bytes). Per-process-only dedupe verified at extension-loader.ts:240–245.
- **Fix:** escalate to FORCE only when the prior LOAD error indicates file-level breakage (`Failed to load library` / `file too short`), keeping plain INSTALL otherwise; and/or persist a small marker (file hash + failed-LOAD fingerprint) so repeat runs skip a doomed FORCE within a cool-down.
- **Corroboration:** Moderate — consistent within one engine (claude risk-architect F1, ce-reliability F3, ce-adversarial F1), no codex lane hit it; postable on the reproduction. Risk-architect notes the README documents the tradeoff — this is a deliberate-tradeoff-with-a-sharp-edge, not an oversight, so a targeted follow-up is acceptable if not fixed in-PR. **[reproduced]** (mechanics)

### 5. P2 — `probeFtsExtensionLoad` hardening: can hang doctor forever (no timeout), leaks `db` if the Connection ctor throws, bare `r.close()` can misreport success
- **File/line:** `gitnexus/src/core/lbug/native-check.ts:115` (function 109–126)
- **Problem (three related hardening gaps in the new probe):** (a) no timeout on `import()`/`new Database`/`new Connection`/`conn.query('LOAD EXTENSION fts')` — the sibling installer child gets a 15 s SIGKILL precisely because unbounded native lbug calls are a known hang risk, and `@ladybugdb/core` even exposes `Connection.setQueryTimeout()` (unused repo-wide); (b) the try/finally starts only after both constructors, so a throwing `new lbug.Connection(db)` skips `db.close()`; (c) `r.close()` is called bare inside the same try whose catch produces `reason`, so a close-time error would report a genuinely successful LOAD as `{loaded:false}` — inverting the PR's own "surface the real error" goal (the repo's `closeQueryResult` pattern exists for exactly this).
- **Trigger:** (a) **reproduced** — mkfifo at `.lbdb/extension/<ver>/<platform>/fts/libfts.lbug_extension` blocks the probe indefinitely; `gitnexus doctor` prints the capabilities header and freezes (realistic variant: hung NFS/SMB home). (b)/(c) require a partially-degraded native layer — structural, not constructed.
- **Adjudication of the lane disagreement:** claude risk-architect's "doctor hang: refuted" addressed only the network-hang hypothesis (LOAD never touches the network; probe gated on nativeCheck.ok) — it does not contradict the filesystem-blocking hang claude-ce-adversarial reproduced. ce-reliability's P1 is overstated (trigger population is narrow, doctor is interactive/Ctrl-C-able, one-shot process bounds the leak); ce-adversarial's P3 undersells that this is doctor's **first** untimed live native call, added by this PR, in the tool the degradation warnings send users to. Adjudicated P2.
- **Evidence:** verified in worktree — native-check.ts:112–113 ctors outside the try (114); bare `r.close()` loop at 116; no timeout anywhere in the function; doctor.ts:138–140 bare `await`.
- **Fix:** race the probe against ~10 s (report `probe timed out (extension file or filesystem unresponsive)`), move `new lbug.Connection` inside the try (close `db` via optional chaining), wrap each `r.close()` in its own try/catch.
- **Corroboration:** Moderate — one engine (ce-reliability F1/F2/F5, ce-adversarial F2, risk-architect F2, ce-correctness CE-1), hang leg reproduced. **[reproduced]** (hang); [code-read] (leak/close). *Confidence note: P2 is a borderline one-level upgrade from the reproducing lane's P3 — the trigger population (unresponsive extension file / hung network home dir) is narrow and doctor is interactive; triage accordingly.*

### 6. P3 — LadybugDB LOAD error text (absolute HOME path, OS username) now flows into HTTP `/api/search` and MCP query warnings
- **File/line:** `gitnexus/src/core/search/fts-indexes.ts:17` (`(reason ? \` (${reason})\` : '')`; consumers api.ts:1322, local-backend.ts:2088 — both in diff)
- **Problem:** the broken-file reason embeds the absolute extension path **twice** (`Failed to load library: /home/<user>/.lbdb/extension/0.18.0/linux_amd64/fts/libfts.lbug_extension … file too short`) — leaking username, lbug version, platform triple to any client of `POST /api/search` (no auth/rate-limit on that route) and MCP query output; in auto-policy processes the reason can also carry up to 4000 chars of installer-child stderr including `err.stack` (extension-loader.ts:124–127, mjs:67).
- **Trigger:** `gitnexus serve --host 0.0.0.0` (documented; serve has no auth option) with a broken extension file; remote `POST /api/search {"query":"x"}` → `warning` contains the paths.
- **Evidence:** [reproduced] — claude-ce-security reproduced the exact error text against the vendored 0.18.0 (missing-file variant has no path; broken-file variant has it twice). Interpolation verified in worktree at fts-indexes.ts:14–17. Severity gating: default bind is loopback, and the unauthenticated API already discloses absolute repo paths (`GET /api/repos`) and `err.message` on 500s, so the marginal disclosure is consistent with — and weaker than — the pre-existing surface (hence P3, not the P2 one lane proposed).
- **Fix:** strip/relativize filesystem paths in `ftsDegradedWarning()` for the HTTP/MCP surfaces (e.g. replace the homedir prefix with `~`, or send only the error class), keeping the full path on CLI/doctor/log surfaces where it's useful. A `not.toMatch(/\/home\/|C:\\Users\\/)` test would currently fail — good regression guard.
- **Corroboration:** Moderate — one engine, two lanes (claude security-boundary F1, ce-security SEC-1); no codex lane reached it (codex swarm security lane: "no structured findings"). Postable on the reproduction. **[reproduced]**

### 7. P3 — README timeout row still says plain "INSTALL" after the verb changed
- **File/line:** `gitnexus/README.md:454` (context line inside the PR's own hunk — the sibling cells at 447/453 were updated to FORCE INSTALL in this diff)
- **Problem:** the `GITNEXUS_LBUG_EXTENSION_INSTALL_TIMEOUT_MS` row still reads "out-of-process `INSTALL` child". Same drift in internal log strings: extension-loader.ts:138 (`INSTALL ${name} timed out …` — the timer that now bounds a FORCE INSTALL) and :157–158 (not diff-anchorable; log-level only).
- **Trigger/Evidence:** verified in worktree (line 454 context inside the @@ -444,13 +444,13 @@ hunk).
- **Fix:** one-word README edit; optionally reword the two log strings.
- **Corroboration:** cross-engine — claude docs-dod F1, project-standards PS-3, risk-architect F3 + codex-ce independently noted "one leftover row still saying INSTALL". **[code-read]**

---

## Body-only findings

- **Default hybrid web search never emits the new warning even when FTS is broken** (codex-adversarial; api.ts:1221–1231): with embeddings ready, the hybrid branch calls `hybridSearch()` which swallows FTS failures internally; `ftsAvailable` stays `undefined`, so the `=== false` gate at api.ts:1321 never fires — web clients silently lose BM25 coverage on the default mode this PR aims to make honest. Mechanics verified by synthesis in the worktree. Pre-existing gap (the old static warning had the same gate), so a scope/DoD note rather than a regression. Single codex lane, [code-read]. **Adjudication of overlap (g): this is distinct from claude-ce-adversarial F3** — codex found *warning-never-fires* (hybrid mode), claude found *warning-fires-forever* (stale latch); same `ftsAvailable`/warning-propagation neighborhood, different bugs.
- **Stale latched `loaded:false` warning in long-running serve/MCP** (claude-ce-adversarial F3, P3): pool init loads FTS once per shared Database (pool-adapter.ts:701–703); after the user performs the warning's own remedy (`analyze --repair-fts` in another process), the still-running server keeps emitting "FTS extension failed to load (old reason)… run --repair-fts" until restart/idle-eviction — recreating a #2374-style contradiction with doctor. The latch is pre-existing; the PR makes the *message* claim live state it doesn't have. Note ce-reliability's soundness analysis (ensure() re-runs per call on the singleton path) covers the non-pool path only. Minimal fix: append "restart `gitnexus serve`/the MCP process if this was fixed since it started" to that warning branch. Single-lane, [code-read].
- **Doctor's native-broken fallback reason misdiagnoses "missing" as "failed to load"** (ce-correctness CE-2; doctor.ts:137): all three `checkLbugNative` failure modes collapse to 'LadybugDB native module (lbugjs.node) failed to load' — the same misdiagnosis class this PR fixes, in miniature. Reuse nativeCheck's specificity (`binaryPath ? 'failed to load' : 'missing'`).
- **`warning` field contract turned dynamic with no machine-readable discriminant** (ce-api-contract AC-1): `/api/search`'s `warning` was a fixed grep-able literal; now it branches on live state and splices unbounded third-party text. Consider a `warningCode` companion (`fts_indexes_missing` | `fts_extension_load_failed`). Also: `POST /api/search`'s degraded branch has zero test coverage before or after this PR.
- **Duplication trio** (ce-maintainability): `oneLine()` re-typed inline at native-check.ts:124 vs extension-loader.ts:54; trailing-period strip duplicated at run-analyze.ts:687 and fts-indexes.ts:14; four hand-assembled reason templates in `ensure()` (a structured `{stage, detail}` capability + one formatter would collapse them).
- **In-place FORCE INSTALL concurrent-writer window** (ce-adversarial F5, conf 50 + ce-correctness residual): FORCE rewrites the same inode with no temp+rename (reproduced); two analyze processes sharing HOME (CI shards) can tear the file, and a serve process dlopening mid-write latches a transient failure. Self-heals next run; advisory lockfile or scratch-home+rename would close it. (The 15 s SIGKILL truncation scenario was largely refuted — lbug writes only after the download completes.)
- **`warnedKeys` never resets after an interim recovery** (ce-reliability F4, P3 conf 45): a repeat identical failure after markLoaded is silently under-logged (operator log line only; user-facing warnings unaffected).
- **`packageVersion` walk-up edge quirks** (ce-correctness CE-3): a name-matching intermediate package.json without `version` ends the search with `undefined`; a malformed intermediate aborts the walk. Theoretical for both real callers; fails safe to pre-PR "unknown".
- **e2e harness robustness** (ce-correctness): the seed resolver trusts any >1 MB `libfts` file in the real home as known-good (a broken-but-large file would poison the happy path); the network gate's `HEAD` probe would skip self-heal coverage on servers answering 403/405 to HEAD; tests within each describe are order-dependent (filtering to a single `it` fails — consistent with cli-e2e convention, worth a comment).

---

## Refuted / validated (validation is a feature of this review)

- **SQL injection via `FORCE INSTALL ${extensionName}` — REFUTED, cross-engine.** `EXTENSION_NAME_PATTERN` allowlist validated in both parent (extension-loader.ts:110) and child (mjs:18) before any query; all production names are compile-time literals. (claude security-boundary F2, ce-security REF-1; codex swarm security lane clean.)
- **Offline-first read-path contract (#726/#1161) — PRESERVED, cross-engine.** FORCE lives only in the installer child, reachable only from `ensure()` under policy `auto`; serve/MCP pool paths hardcode `load-only` (pool-adapter.ts:702/768; lbug-adapter.ts:899 readOnly branch); doctor probe is LOAD-only on `:memory:`; Dockerfile.cli bake is a fresh layer (FORCE ≡ INSTALL) with a `--verify-only` LOAD-only gate. (ce-correctness R-8, ce-adversarial R6, security-boundary F3, project-standards, facts-historian; codex swarm lanes 3+5 independently concluded the same.)
- **`capabilities.ts` walk-up path traversal / wrong-package resolution — REFUTED.** Hardcoded caller names, ≤5 hops, `pkg.name === name` guard, readFileSync+JSON.parse only (no require of walked files). (security F4, ce-security REF-3, ce-correctness R-9, ce-adversarial R4 [executed: fingerprint → 0.18.0].)
- **Empty-string LOAD error flipping the null-check — REFUTED [reproduced].** Both call sites use strict `=== null`; `new Error('')` correctly classifies as failure. (ce-correctness R-1, ce-adversarial R1, maintainability.)
- **FORCE INSTALL unsupported on 0.18.0 / destroys a working file — REFUTED [reproduced + CI].** Full heal sequence executed; dead-repo FORCE over a valid file left it byte-identical; self-heal e2e passed on the networked CI runner (closes risk-architect's S1).
- **`alreadyAvailable` masking a broken-file error as success — REFUTED empirically** (neither real error contains the substrings); the "already exists in catalog" edge predates this PR unchanged (ce-adversarial R7, for the record).
- **`ftsDegradedWarning` firing spuriously / reading an unpopulated cache — REFUTED for reachable flows** (every FTS query route runs ensure() first; pool registers entries only after the FTS load attempt). (ce-correctness R-5, ce-adversarial R8, risk-architect.)
- **repair-fts reading a stale/foreign capability — REFUTED** (same-process singleton, ensure() immediately precedes the read). Old-message string consumers — none left un-updated (grep-verified). Warn-dedup spam via variable reasons — REFUTED (reason strings are stable within a process; installAttempted caches the install leg).
- **Infinite FORCE loops within a process / doctor probing the network — REFUTED** (per-process installAttempted; LOAD is offline by construction).

**Author-session claims: re-verified vs re-read.** Independently re-executed: INSTALL-no-op premise, FORCE INSTALL semantics, probe reason accuracy, exports-gap version fallback, e2e doctor surface. Re-read only (still open): the FORCE→INSTALL **mutation** claim (no lane re-ran the mutation; mitigated by CI's green self-heal run but see inline #3), and the OS-independence of the lbug C++ error strings — which Windows CI cannot confirm until inline #2 is fixed, since the suite doesn't run there.

---

## Test gaps (summary)

- `--repair-fts` reason-interpolation branch is vacuous at unit level (ce-testing F1, conf 85; **mechanically re-verified by synthesis**): `run-analyze-fts-repair.test.ts:286` mocks `loadFTSExtension` from lbug-adapter, bypassing the real ExtensionManager, so `getExtensionCapabilities()` returns `[]`, `ftsReason` is always undefined, and the updated regex (line 310) passes whether or not the interpolation fires. Only the e2e test (which is network/registration-fragile per inline #2/#3) protects the PR's headline surface. Fix: also mock `getExtensionCapabilities` to return a reason and assert it appears.
- `probeFtsExtensionLoad()` has **zero unit tests** (ce-testing F2): the two doctor branches are covered only via e2e with a healthy native module; the `nativeCheck.ok === false` fallback branch has no automated coverage at all.
- FORCE INSTALL has only network-gated coverage (inline #3).
- No tests for: probe timeout/hang behavior, the cross-run FORCE re-download loop (inline #4), the stale-server warning sequence, the `capabilities.ts` walk-up edge branches (hop bound, name-mismatch continue), or the composed-warning double-punctuation strip (`not.toMatch(/\.\)/)`).

---

## CI status

- Quality gates, CodeQL (both langs), dependency review, gitleaks, docker web, packaged-install smoke (ubuntu+windows), node-22.14 floor, benchmarks, macos platform-sensitive, tree-sitter ABI ×3: **green**.
- **tests/ubuntu/coverage: FAILED — unrelated to this PR.** Sole failure is `test/unit/hooks.test.ts` "Plugin: hook does not exceed MAX_INFLIGHT under simultaneous bursts (hard cap)" (1 failed / 647 passed / 8 skipped); the PR touches no hook code (empty diff for hooks.test.ts and hooks/). This matches the repo's known full-suite load-flake class — and, notably, is the flake class that inline #2's parallel-registration fix is meant to stop aggravating. CI Gate failure is downstream of it.
- **The new `fts-extension-e2e.test.ts` ran on that job and passed 11/11 (44.4 s)** in the `default` vitest project on a networked runner — the FORCE INSTALL self-heal path was exercised on real CI, validating the fix end-to-end and confirming the registration findings are about Windows coverage + parallelism hygiene, not "suite doesn't run."
- tests/windows-latest (platform-sensitive) and Docker Build & Push (gitnexus) both **passed** (the Docker pass exercises the Dockerfile.cli FORCE-INSTALL bake). Vercel failure is fork-authorization only, not a code signal.

---

## Verdict

- Branch hygiene: `clean feature/fix PR` · Merge state: `review blocked` (open P2 items; CI Gate currently red on an unrelated flake)
- **Not production-ready as-is — three small test-side P2s to fix in this PR** (per the repo's fix-in-PR convention): #1 offline-skip crash and #2 Windows/cross-platform + sequential-project registration (both engines converged on these), plus the deterministic FORCE INSTALL guard (#3). No production-code blocker was found by any lane; the offline-first contract and the PR's core self-heal premise were validated by execution on both engines' evidence plus a green networked CI run. #4–#7 are code-hardening/polish with clear, bounded fixes.

---

*Automated multi-engine review digest: Claude (GitNexus swarm + ce-code-review personas) and Codex GPT-5.5/xhigh (same swarm + ce contracts + native adversarial pass), merged by a dedicated synthesis agent and gated by an independent critic. Findings above carry honest [reproduced]/[code-read] tags — verify before acting.*

gitnexus/test/integration/fts-extension-e2e.test.ts
```diff
@@ -0,0 +1,332 @@
...
+beforeEach((ctx) => {
+  if (!seedExtensionFile) ctx.skip();
+});
+
+describe('happy path — extension pre-installed, fully offline (load-only)', () => {
+  let home: string;
+  let repo: string;
+
+  beforeAll(() => {
+    ({ home } = makeHome('valid'));
```
magyargergo, line 206: **[P2 · test-infra · reproduced]** On an offline machine with no seed extension (and `GITNEXUS_REQUIRE_FTS` unset), this `beforeAll` crashes with a `TypeError` instead of skipping: `makeHome('valid')` runs before the file-level `beforeEach(ctx.skip)` (L194–196) can fire, and `extensionRelPath` (L42) is still `undefined` when `resolveSeedExtension()` found nothing — so `path.join(home, extensionRelPath)` (L100) throws in this hook and again in the broken-file describe (L241–244). 8 tests go red instead of visibly skipping.

**Fix:** early-return from these `beforeAll`s when `!seedExtensionFile` (or default `extensionRelPath` / build homes lazily inside the skippable tests).

*Corroboration: cross-engine — Codex ce (conf 100), Codex adversarial, and Claude adversarial (reproduced `path.join('/a', undefined)` TypeError) all found this independently.*

gitnexus/test/integration/fts-extension-e2e.test.ts
```diff
@@ -0,0 +1,332 @@
...
+ */
+import { describe, it, expect, beforeAll, beforeEach, afterAll } from 'vitest';
+import { spawnSync } from 'child_process';
+import path from 'path';
+import fs from 'fs';
+import os from 'os';
+import { fileURLToPath, pathToFileURL } from 'url';
+import { createRequire } from 'module';
+
+import lbug from '@ladybugdb/core';
```
magyargergo, line 29: **[P2 · test-infra · code-read]** This suite spawns the real CLI (SPAWN_CLI category) and imports `@ladybugdb/core`, but it's registered in neither `scripts/cross-platform-tests.ts` (SPAWN_CLI list) nor any sequential vitest project — verified absent from both. It *does* run on CI (it landed in the `default` project wildcard and passed 11/11 on ubuntu/coverage this run), so the two real harms are: (a) the regression suite for a **Windows-reported** bug (#2374, user-placed extension file on Windows 11; lbug C++ error-string assertions unverified on Windows) never executes on Windows/macOS CI, which only runs the cross-platform subset; (b) 11 CLI-spawning tests run file-parallel with the rest of the suite, aggravating the repo's known under-load timeout-flake class (this very run had an unrelated `hooks.test.ts` burst-test flake fail the coverage job).

**Fix:** add the file to the SPAWN_CLI array in `scripts/cross-platform-tests.ts` and register it in a sequential vitest project (`cli-e2e` fits: CLI spawning, long timeouts) + the `default` exclude list. Precedent: commit 287aa233e registered the CJK integration test the same way.

*Corroboration: cross-engine — Claude test-ci + project-standards + docs-dod, and Codex swarm Lane-4, all flagged the registration gap.*

gitnexus/scripts/install-duckdb-extension.mjs
```diff
@@ -46,7 +46,12 @@ async function installDuckDbExtension(extensionName, verifyOnly = false) {
         `[install-ext] LOAD-only verify OK for '${extensionName}' (HOME=${process.env.HOME})`,
       );
     } else {
-      await conn.query(`INSTALL ${extensionName}`);
+      // FORCE: this installer only runs after LOAD already failed, so any
+      // existing extension file is broken (truncated/wrong-platform, #2374).
+      // Plain INSTALL sees the file exists, reports success without
+      // re-downloading, and LOAD keeps failing forever. (`UPDATE <ext>` is the
+      // Kuzu-documented re-download verb; both behave identically on LadybugDB.)
+      await conn.query(`FORCE INSTALL ${extensionName}`);
```
magyargergo: **[P2 · test-coverage · code-read]** The headline `INSTALL` → `FORCE INSTALL` fix has no deterministic regression guard. This is the sole production occurrence of the verb, zero unit tests assert it, and the only test that exercises it is the network-gated e2e describe (`fts-extension-e2e.test.ts:298–300`) whose `ctx.skip()` on `!networkAvailable` is a **soft silent skip** outside the `GITNEXUS_REQUIRE_FTS=1` hard-fail contract (that throw only guards seed-extension existence, L166–170). A revert to plain `INSTALL` — which no-ops over a broken file, the exact #2374 trap, empirically confirmed by this review — could merge green on any day CI can't reach `extension.ladybugdb.com`. (It did run and pass on this networked run.)

**Fix:** cheap offline unit test — inject/mock the `@ladybugdb/core` connection (or extract the SQL-verb choice) and assert `conn.query` is called with `FORCE INSTALL fts`, matching the rigor already applied to `getExtensionInstallChildProcessArgs`.

*Corroboration: cross-engine — Claude testing (conf 80) + Codex swarm Lane-4.*

gitnexus/scripts/install-duckdb-extension.mjs
```diff
@@ -46,7 +46,12 @@ async function installDuckDbExtension(extensionName, verifyOnly = false) {
         `[install-ext] LOAD-only verify OK for '${extensionName}' (HOME=${process.env.HOME})`,
       );
     } else {
-      await conn.query(`INSTALL ${extensionName}`);
+      // FORCE: this installer only runs after LOAD already failed, so any
```
magyargergo: **[P2 · reliability · reproduced (mechanics)]** "any existing extension file is broken" is unsound when LOAD fails for a reason a re-download can't fix (missing MSVC runtime, glibc/musl mismatch, noexec mount, upstream shipping a broken build). The install-outcome cache is per-process only (`installAttempted`, extension-loader.ts:240–245) and every `gitnexus analyze` is a fresh process under default `auto` policy — so such machines now re-download ~2.2 MB with up-to-15 s latency on **every analyze run, forever** (pre-PR, plain INSTALL no-op'd instantly; the only opt-out is manually setting `GITNEXUS_LBUG_EXTENSION_INSTALL=load-only`). Reproduced: `FORCE INSTALL` rewrites even a valid file in place on every invocation.

**Fix:** escalate to FORCE only when the prior LOAD error indicates file-level breakage (`Failed to load library` / `file too short`), keeping plain INSTALL otherwise; and/or persist a small failed-heal marker so repeat runs skip a doomed FORCE within a cool-down.

*Corroboration: three Claude lanes (risk-architect, reliability, adversarial) — consistent within one engine; README does document the load-only mitigation, so this is a deliberate tradeoff with a sharp edge rather than an oversight.*

gitnexus/src/core/lbug/native-check.ts
```diff
@@ -89,3 +89,38 @@ export function checkLbugNative(overridePkgDir?: string): NativeCheckResult {
...
+ * file from a present-but-broken one (wrong platform, truncated download).
+ * Dynamic import so doctor still runs when the native module itself is broken.
+ */
+export async function probeFtsExtensionLoad(): Promise<FtsProbeResult> {
+  try {
+    const { default: lbug } = await import('@ladybugdb/core');
+    const db = new lbug.Database(':memory:');
+    const conn = new lbug.Connection(db);
+    try {
+      const result = await conn.query('LOAD EXTENSION fts');
```
magyargergo: **[P2 · reliability · reproduced (hang) / code-read (leak, close)]** Three hardening gaps in the new probe: (a) **no timeout** on `import()` / `new Database` / `new Connection` / `conn.query('LOAD EXTENSION fts')` — reproduced: a blocking file at the extension path (mkfifo; realistic analog: hung NFS/SMB home) freezes `gitnexus doctor` forever, the very tool the new degradation warnings send users to. The sibling installer child gets a 15 s SIGKILL precisely because unbounded native lbug calls are a known hang class, and `Connection.setQueryTimeout()` exists unused. (b) `new lbug.Connection(db)` sits outside the try/finally — if it throws, `db.close()` never runs (bounded: doctor exits right after). (c) the bare `r.close()` loop runs inside the try whose catch produces `reason` — a close-time error would report a genuinely successful LOAD as `{loaded: false}`, inverting this PR's own goal (the repo's `closeQueryResult` pattern exists for exactly this).

**Fix:** race the probe against ~10 s (reporting `probe timed out — extension file or filesystem unresponsive`), move the `Connection` ctor inside the try and close `db` via optional chaining, wrap each `r.close()` in its own try/catch.

*Confidence note: borderline P2 (one-level up from the reproducing lane's P3) — narrow trigger population, doctor is interactive/Ctrl-C-able. One engine, four lanes.*

gitnexus/src/core/search/fts-indexes.ts
```diff
@@ -1,6 +1,26 @@
...
+ * generic indexes-missing message, so "indexes exist but the extension broke"
+ * is not misreported as missing indexes.
+ */
+export const ftsDegradedWarning = (): string => {
+  const fts = getExtensionCapabilities().find((c) => c.name === 'fts');
+  if (fts && !fts.loaded) {
+    const reason = fts.reason?.replace(/\.$/, '');
+    return (
+      'FTS extension failed to load — keyword search degraded' +
+      (reason ? ` (${reason})` : '') +
```
magyargergo, line 29: **[P3 · info-disclosure · reproduced]** The broken-file LOAD reason embeds the absolute extension path **twice** (`Failed to load library: /home/<user>/.lbdb/extension/0.18.0/linux_amd64/fts/... file too short`) — leaking username, lbug version, and platform triple into `POST /api/search` responses (route has no auth/rate-limit) and MCP query warnings; in auto-policy processes the reason can also carry up to 4000 chars of installer-child stderr incl. `err.stack` (extension-loader.ts:124–127). Trigger: `gitnexus serve --host 0.0.0.0` (documented mode, no auth option) with a broken extension file. Severity bounded to P3: default bind is loopback and the unauthenticated API already discloses absolute repo paths (`GET /api/repos`), so the marginal disclosure is consistent with the pre-existing surface.

**Fix:** strip/relativize filesystem paths in `ftsDegradedWarning()` for the HTTP/MCP surfaces (homedir → `~`, or error class only), keeping the full path on CLI/doctor/log surfaces. A `not.toMatch(/\/home\/|C:\\Users\\/)` unit test would currently fail — good regression guard.

*Corroboration: two Claude security lanes (boundary P2 / ce P3), reproduced error text; no Codex lane reached it.*

gitnexus/README.md
```diff
@@ -444,13 +444,13 @@ The prefix defaults to `~/.gitnexus/embedding-runtime`; set `GITNEXUS_EMBEDDING_
...
 | `GITNEXUS_LBUG_EXTENSION_INSTALL_TIMEOUT_MS` | positive integer             | `15000`             | Wall-clock budget for the out-of-process `INSTALL` child before it is killed.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |
```
magyargergo: **[P3 · docs · code-read]** This row still says "out-of-process `INSTALL` child" while the sibling cells at L447/L453 were updated to `FORCE INSTALL` in this PR. Same drift in the internal timeout message at `extension-loader.ts:138` (`INSTALL ${name} timed out …` — the timer that now bounds a FORCE INSTALL) and :157–158.

**Fix:** one-word README edit; optionally reword the two log strings.

*Corroboration: cross-engine — Claude docs-dod + project-standards + risk-architect, and Codex ce independently noted the leftover row.*

Issue #2377, opened 2026-07-05
startLine, endLine off-by-one downward (zero-based?) content contains two lines before/after


wtmorrison:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

startLine, endLine off-by-one downward (zero-based?) content contains two lines before/after for `MATCH (f:Function) WHERE f.filePath = 'gitnexus/src/server/api.ts' RETURN f.startLine, f.endLine, f.name, f.content;`

### Context

Using GitNexus 1.6.9

I indexed the code of GitNexus itself, commit 6e420400702d1fee9cb6619c142ffde5304947cd

Query:

```
MATCH (f:Function)
WHERE f.filePath = 'gitnexus/src/server/api.ts'
RETURN f.startLine, f.endLine, f.name, f.content;
```

Comparing results to `less -N gitnexus/src/server/api.ts' taking waitForDrain as example.

```
      "f.startLine": 246,
      "f.endLine": 279,
      "f.name": "waitForDrain",
      "f.content": "};\n\nconst waitForDrain = async (res: express.Response, signal?: AbortSignal): Promise<void> => {\n  ensureStreamIsWritable(res, signal);\n\n  await new Promise<void>((resolve, reject) => {\n    const cleanup = () => {\n      res.off('drain', onDrain);\n      res.off('close', onClose);\n      signal?.removeEventListener('abort', onAbort);\n    };\n\n    const onDrain = () => {\n      cleanup();\n      resolve();\n    };\n    const onClose = () => {\n      cleanup();\n      reject(new ClientDisconnectedError());\n    };\n    const onAbort = () => {\n      cleanup();\n      reject(new ClientDisconnectedError());\n    };\n\n    res.once('drain', onDrain);\n    res.once('close', onClose);\n    signal?.addEventListener('abort', onAbort, { once: true });\n\n    if (signal?.aborted || res.destroyed || res.writableEnded) {\n      onAbort();\n    }\n  });\n\n  ensureStreamIsWritable(res, signal);\n};\n\nconst isClientDisconnectWriteError = (err: unknown): boolean => {"
```

1. startLine and endLine are one less than expected

sed '246,279!d' gitnexus/src/server/api.ts

Returns the blank line before, does not return the trailing '};' 

Maybe Tree sitter uses 0-based line counts? Still, convention for line numbering is 1-based in IDEs/editors, sed, less -N, etc.

2. content includes two lines before and after the actual function content (see example above)

[...]

PR #2379, opened 2026-07-06
fix(lbug): store exact symbol content snippets

koriyoshi2041:
## Summary

Refs #2377.

This fixes the symbol `content` part of the report: Function/Method/Class/etc. rows now store the exact symbol span instead of adding the two neighboring lines before and after the symbol.

I intentionally left the `startLine` / `endLine` basis unchanged in this PR. Several PDG/taint paths currently document symbol spans as 0-based and shift them when joining to 1-based BasicBlock rows, so changing that looks broader than the content-padding fix.

## Verification

- `npm test -- --run test/integration/csv-pipeline.test.ts` -> 22 passed
- `npx prettier --check src/core/lbug/csv-generator.ts test/integration/csv-pipeline.test.ts` -> passed
- `npx eslint src/core/lbug/csv-generator.ts test/integration/csv-pipeline.test.ts` -> 0 errors, existing warnings in the test file only
- `git diff --check` -> passed

Non-blocking current-main note:
- `npx tsc --noEmit` currently fails on existing Spring DI type errors in `src/core/ingestion/di-extractors/spring.ts` and `src/core/ingestion/pipeline-phases/di.ts`, unrelated to this patch.


gitnexus/src/core/lbug/csv-generator.ts
```diff
@@ -233,8 +260,9 @@ const extractContent = async (node: GraphNode, contentCache: FileContentCache):
   if (startLine === undefined || endLine === undefined) return '';
 
   const lines = content.split('\n');
-  const start = Math.max(0, startLine - 2);
-  const end = Math.min(lines.length - 1, endLine + 2);
+  const exactSymbolContent = EXACT_SYMBOL_CONTENT_LABELS.has(node.label);
+  const start = Math.max(0, exactSymbolContent ? startLine : startLine - 2);
```
magyargergo, line 280: **P2 — COBOL/JCL exact snippets are off-by-one / wrong-line.**

This exact-path slice treats `startLine`/`endLine` as **0-based** array indices. That holds for tree-sitter symbols (`parse-worker.ts:2312` is 0-based), but **COBOL and JCL emit 1-based line numbers**, and every label they produce is in `EXACT_SYMBOL_CONTENT_LABELS`. So their stored FTS `content` is shifted by one: multi-line symbols drop their own declaration line and bleed one trailing line; **single-line** symbols (JCL Job/Step/DD, COBOL data items / EXEC blocks) store the **wrong line entirely**, or an **empty string** when the symbol is on the file's last physical line.

**Evidence:** 1-based source at `cobol-preprocessor.ts:1142` and `jcl-parser.ts:131` (`const lineNum = i + 1;`) flows to `cobol-processor.ts:480` (Function `para.line`), `:362` (Module hardcoded `startLine: 1`), `:445` (Namespace `sec.line`), `:514` (Property `item.line`), `:617` (CodeElement `call.line`), and `jcl-processor.ts:101/140/212/247` (CodeElement/Module). A first-line JCL JOB stores `lines[1]` instead of `lines[0]`. The old `Math.max(0, startLine-2)` window accidentally compensated for the +1 offset (it always reached index ≥0), so this is a **regression** exposed by removing the ±2 buffer for these labels.

**Fix:** normalize the COBOL/JCL emitters to 0-based (`-1`; full-file span `endLine: lines.length-1`) so they match every other emitter — the root-cause fix — **or** keep those labels on the ±2 window / exclude them from the exact path until their ranges are normalized. Caveat: `findOwningProgramName` / `findContainingSection` compare a `line` value against `startLine` from the **same** 1-based source, so normalize **both sides together** and check the route-containment / MCP-display consumers before flipping.

_Cross-engine: both engines, 5 lanes (claude risk-architect + ce-correctness; codex swarm + ce + adversarial). Coordinator-verified. `[code-read]` — no COBOL/JCL index was executed; confirm against a real mainframe index if precision matters._

gitnexus/test/integration/csv-pipeline.test.ts
```diff
@@ -171,6 +171,46 @@ describe('streamAllCSVsToDisk', () => {
...
+    const functionCsv = result.nodeFiles.get('Function');
+    expect(functionCsv).toBeDefined();
+    if (!functionCsv) {
+      throw new Error('Function CSV was not generated');
+    }
+
+    const content = await fs.readFile(functionCsv.csvPath, 'utf-8');
+    expect(content).toContain('export function target()');
+    expect(content).toContain('return before;');
+    expect(content).not.toContain('const before = 1;');
```
magyargergo, line 210: **P2 — this test doesn't actually pin exactness, and there's no 1-based (COBOL/JCL) coverage for the headline bug.**

The fixture places **blank lines** at array indices 1 and 5 — immediately adjacent to the exact `[2,4]` boundary — and the assertions only check the non-empty neighbors at indices 0 and 6. A ±1 slice shift at either edge (exactly the off-by-one class this PR is about) would pull in only a blank line, so all four assertions still **pass**. The test also exercises only a hand-built **0-based** `Function` node, so the confirmed COBOL/JCL 1-based regression (see the comment on `csv-generator.ts:264`) is entirely uncovered.

**Fix:** remove the blank separators so neighbor text sits directly adjacent to the boundary (or assert the exact normalized snippet string), **and** add a regression case for a 1-based producer (COBOL Function/Module or JCL CodeElement) asserting the stored snippet is the symbol's own line(s). `[code-read]`

magyargergo (commented):
## PR #2379 — fix(lbug): store exact symbol content snippets — tri-engine review

**Engines / legs (all live).** Claude (Engine A) ran the GitNexus swarm (risk-architect, test-ci) + ce personas (correctness, testing, maintainability, project-standards). Codex/GPT-5.5 @ xhigh (Engine B) ran swarm + ce + adversarial. Every Codex leg produced real findings — `codex-ce` recovered from an initial skill-path miss and still returned a structured P2. **The headline finding is cross-engine — independently reached by both engines across 5 lanes.**

**Verdict: not production-ready as-is.** One P2 regression (COBOL/JCL FTS content) should be fixed or scoped out of the exact-slice path before merge; the change is a clean improvement for every tree-sitter language. Branch: clean fix PR (one merge-from-main commit). Merge state: review-blocked pending the P2. CI: green except Vercel (auth-only infra noise).

---

### What's solid (validated, credit where due)

- **Core change is correct for every tree-sitter language.** `extractContent` now slices `lines.slice(startLine, endLine+1)` with no ±2 padding for `EXACT_SYMBOL_CONTENT_LABELS`. Graph node `startLine`/`endLine` from the worker path are uniformly **0-based** (`parse-worker.ts:2312` = `definitionNode.startPosition.row + lineOffset`, `lineOffset` default 0; doc-confirmed at `summary-harvest-driver.ts:21`). Boundary math checks out: single-line (`slice(L,L+1)`), file-start (`Math.max(0,…)`), file-end (`Math.min(lines.length-1,…)`) all safe. The added test passes.
- **The general "off-by-one for all symbols" worry was investigated and REFUTED.** The `row + 1` sites in cpp/dart `captures.ts` build `Range`/`Capture` scope objects, not `properties.startLine`. Dropping the ±2 context window is intentional per the PR title; no consumer was found to depend on the extra recall.
- **Standards clean:** no `any` introduced, PR title conventional-commit valid, diff minimal (2 files), CHANGELOG correctly absent (owned at release), test correctly placed in the existing integration file.
- **CI:** functional checks green; only Vercel is UNSTABLE and that's auth-only infra noise, not a real failure. `ci-tests.yml` runs `vitest run` over `gitnexus/` with `GITNEXUS_REQUIRE_FTS=1`, so `csv-pipeline.test.ts` (including the new case) does run in CI. test-ci lane executed it locally — passes.

---

### Headline finding — P2 regression: COBOL/JCL exact snippets are off-by-one / wrong-line

**Corroboration: STRONG — both engines, 5 lanes** (claude risk-architect P2 "blocks merge", claude ce-correctness P2 conf 0.80, codex-swarm P3, codex-ce P2 conf 75, codex-adversarial "medium / no-ship"). Coordinator + this synthesis confirmed by code-read.

`extractContent` (`csv-generator.ts:263-265`) treats `startLine`/`endLine` as 0-based array indices. That holds for tree-sitter, but **COBOL and JCL emit 1-based line numbers** — and every label they emit is in `EXACT_SYMBOL_CONTENT_LABELS`:

- Source of the 1-based values: `cobol-preprocessor.ts:1142` and `jcl-parser.ts:131`, both `const lineNum = i + 1;`.
- Flows to exact-set labels: `Function` (`cobol-processor.ts:480` `para.line`), `Namespace` (`:445` `sec.line`), `Property` (`:514` `item.line`, single-line), `CodeElement` (`:617` `call.line`; `jcl-processor.ts:101/140/212` job/step/dd `.line`, single-line), `Module` (`:362` hardcoded `startLine: 1`; `jcl-processor.ts:247` `proc.line`).

**Concrete triggers:**
- *Multi-line* symbol (COBOL paragraph at physical lines 4–6 → emitted `startLine=4,endLine=6` 1-based): exact path `slice(4,7)` = `lines[4..6]` → **drops the paragraph's own header line and bleeds one line into the next symbol** (window shifted +1).
- *Single-line* symbol (JCL JOB / COBOL data item / EXEC block; `startLine==endLine`): stores the **wrong line entirely** — the line *after* the symbol. A first-line JCL JOB (`job.line=1`) stores `lines[1]` instead of `lines[0]`. If it sits on the file's **last physical line**, `slice(len,len)` = `[]` → **empty stored content**.

**Why it's a PR regression, not pre-existing:** the old `Math.max(0, startLine-2)` / `Math.min(len-1, endLine+2)` window accidentally compensated for the 1-based offset (the ±2 buffer absorbed the +1 shift and started at 0), so pre-PR snippets were a correct superset. Removing the buffer for these labels exposes the latent mismatch. Impact is FTS/search-content fidelity only (no crash, no graph-edge corruption), and the languages are niche — hence P2, not higher.

**Fix direction (either):** (1) normalize the COBOL/JCL emitters to 0-based (`startLine-1`/`endLine-1`, full-file span `endLine: lines.length-1`) so they match every other emitter — the root-cause fix; or (2) keep those 1-based labels on the old ±2 window / exclude them from the exact path until their ranges are normalized. **Blast-radius caveat:** if you flip the emitters, `findOwningProgramName` / `findContainingSection` compare a `line` value against `startLine` drawn from the *same* 1-based source — normalize **both sides together**, and check the route-containment / MCP-display consumers before flipping. A localized `-1` inside `csv-generator` would work but conflicts with the repo's "shared pipeline code must not name languages" norm.

---

### Inline anchors (2)

1. **`csv-generator.ts:264` — P2, headline COBOL/JCL off-by-one** (both engines, 5 lanes). Exact anchor is the added `const start = …` line.
2. **`csv-pipeline.test.ts:210` — P2, test doesn't pin exactness + no 1-based coverage** (claude ce-testing F2; risk-architect / test-ci / codex-swarm on the missing COBOL/JCL case). Anchor is the added `.not.toContain('const before = 1;')` assertion.

Both are added (RIGHT-side) diff lines — verified against `git diff <base> HEAD`.

---

### Lower-priority / body items

- **P3 (single-lane, risk-architect) — same 1-based bug on scope-tree CodeElement nodes.** `emit-references.ts:143` stores `startLine: scope.range.startLine` (1-based via `ast-helpers.ts:1160` `row+1`) on CodeElement scope nodes. Gated behind `INGESTION_EMIT_SCOPES=1` (non-default), so opt-in — same root cause, lower reach.
- **P2 (single-engine, ce-maintainability) — DRY: third hand-maintained symbol-label list.** `EXACT_SYMBOL_CONTENT_LABELS` (24 labels) duplicates `CUSTOM_CONTRACT_RESOLVE_QUERY`'s allowlist (`manifest-extractor.ts:19`, 21 labels) almost verbatim (+Namespace/Variable/Module), with no shared source of truth. The repo already models the fix — `callable-labels.ts` exports a single reused `NodeLabel` predicate. Suggest deriving both from one exported `SYMBOL_NODE_LABELS` set.
- **P3 (ce-maintainability) — no rationale comment.** Every other module-level const in this file carries a "why" comment; the new set has none, and (ironically, given the headline) doesn't note the 0-based-line assumption the exact slice depends on.
- **Test-coverage gaps (test-ci, ce-testing F1):** the ±2 fallback branch is now reachable **only** via `Section` (the sole exact-set non-member on this path) and is untested; single-line / file-start / file-end / `MAX_SNIPPET`-truncation exact-path edges untested. Defense-in-depth P3s.
- **Style nit (project-standards, informational):** the new test narrows via `if (!functionCsv) throw` while every sibling in the file uses the `!` non-null assertion — real inconsistency, but no committed standards rule mandates a style, so not a scored finding.

---

### Refuted suspicions (validation credit)

- **General tree-sitter off-by-one — REFUTED.** `parse-worker.ts:2312-2313` is 0-based; cpp/dart `row+1` sites are Range objects, not `properties.startLine`. If any lane resurfaces this, it is refuted-on-verification.
- **±2 context-loss as a correctness bug — not a bug.** Intended per PR title; no consumer dependency found.
- **Boundary/clamp crashes (startLine=0, endLine<startLine, indices > len) — safe.** Clamps + `slice` tolerate all branches.

---

### CI status

Functional checks green. The single UNSTABLE signal is **Vercel — auth-only infra noise**, not a real CI failure. The new test runs in CI via `ci-tests.yml` (`vitest run`, `GITNEXUS_REQUIRE_FTS=1`) and passes locally.

---

*This is an automated two-engine (Claude + GPT-5.5) multi-lane digest. Verify the file:line evidence before acting — the headline is a code-read trace (no COBOL/JCL index was executed), so confirm against a real mainframe index if precision matters.*

PR #2380, opened 2026-07-06
fix(lbug/mcp): exact symbol content + 0-based line storage with 1-based MCP display (#2377, #2379)

### LINKED ISSUE #2377: startLine, endLine off-by-one downward (zero-based?) content contains two lines before/after

magyargergo:
## Summary

Fixes #2377 and carries forward / absorbs #2379 (thanks @koriyoshi2041 — the exact-content commit is preserved).

**#2377** reported two things about `RETURN f.startLine, f.endLine, f.content`:
1. `startLine`/`endLine` are **0-based** (don't line up with editors / `sed` / `less -N`);
2. `content` carried **±2 padding** lines.

**#2379** fixed the padding for tree-sitter languages via exact-span slicing — but regressed **COBOL/JCL**, because those are the only graph-node emitters that stored **1-based** line numbers, so the exact slice dropped each symbol's declaration line (and stored the wrong line, or empty, for single-line COBOL/JCL symbols). This PR fixes that root cause and adds the 1-based display half.

### Storage stays 0-based, uniformly (content half of #2377 + the #2379 regression)
- New `toZeroBasedLine` helper; the COBOL/JCL, scope-graph, and **markdown `Section`** emitters now store 0-based `startLine`/`endLine` — converted at the **graph-node emission boundary** only, so parser-internal `.line` values, `L${line}` node/edge IDs, and containment checks are untouched.
- `extractContent` is **unchanged** — the #2379 exact slice is correct once every emitter is 0-based.
- DRY: `EXACT_SYMBOL_CONTENT_LABELS` now derives from a single shared `SYMBOL_NODE_LABELS`.

### MCP tools present 1-based (line-number half of #2377)
- `context`, `query`, and `impact` present **1-based** line numbers (editor / `sed` aligned) via `toDisplayLine` at the response boundary.
- Raw `cypher` and stored values stay **0-based** and are documented in the `…/schema` resource. BasicBlock / PDG statement lines (already 1-based) and internal join params are untouched.

## Tests
- COBOL/JCL 0-based emitter assertions (Module=0; JCL job/step self-validated).
- Hardened exact-content slicing — the fixture's blank buffer is removed so a ±1 shift now **fails** — plus one-line-symbol and ±2-fallback (`Section`) cases.
- Markdown `Section` line numbers updated to 0-based.
- MCP: `context` reports 1-based while a raw `cypher` of the same node stays 0-based.

Validated locally: `tsc` clean; the COBOL/markdown/csv-pipeline/MCP/PDG/taint suites pass (200+ MCP/PDG tests, 583 in the ingestion sweep).

## Notes / follow-ups
- **Consistency:** `trace` / `explain` / `pdg_query` / resources still surface 0-based symbol lines — a follow-up for full 1-based parity across every tool.
- **Discovered, pre-existing (separate from this fix):** COBOL `COPY` expansion makes `para.line` an *expanded-source* coordinate while `extractContent` reads the *raw* file, so COBOL symbol `content` is misaligned for files containing `COPY`. Orthogonal to the 0/1-based issue; worth its own issue.


gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -2255,8 +2256,8 @@ export class LocalBackend {
               name: sym.name || sym[1],
               type: sym.type || sym[2],
               filePath: sym.filePath || sym[3],
-              startLine: sym.startLine || sym[4],
-              endLine: sym.endLine || sym[5],
+              startLine: toDisplayLine(sym.startLine || sym[4]),
```
magyargergo: **[P1 · blocks merge · reproduced] `query()` double-converts BM25/FTS line numbers.**
`bm25Search` applies `toDisplayLine` here (2259-2260), so its rows are already 1-based; `query()` then re-applies `toDisplayLine` at 1991-1992 → a symbol stored 0-based `41` is reported as **43** (want 42). `semanticSearch` leaves its lines raw, so vector-matched symbols come out correct (42), and RRF keeps the BM25 copy when both retrievers hit — so the reported line silently depends on which retriever surfaced the symbol. `bm25Search` is called only from `query()`, so the fix is local: **drop `toDisplayLine` at 2259-2260** and keep the single conversion at 1991-1992; add a `query()` test asserting stored 41 → 42. _(ce-correctness reproduced 5→6→7 with a failing test; Codex swarm + adversarial + Claude risk/adversarial concur.)_ [reproduced]

Also on this line: the `sym.startLine || sym[4]` **falsy-OR** drops `startLine` for 0-based first-line symbols — every COBOL `Module` (`toZeroBasedLine(1)=0`), markdown h1 (also `context()` at 3250) — use `??`, not `||`. [reproduced]

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -735,8 +736,8 @@ export class LocalBackend {
           name: s.name,
           type: s.type,
           filePath: s.filePath,
-          startLine: s.startLine,
-          endLine: s.endLine,
+          startLine: toDisplayLine(s.startLine),
```
magyargergo, line 757: **[P2 · code-read] Group/cross-repo `trace` mixes line-number bases in one response.**
`resolveSymbolForGroup` now returns 1-based endpoints (here + the ambiguous branch at 752), and `cross-trace.ts` `endpointFrom` (518) uses them for `from`/`to`, while `tagHops` (509) copies `port.trace()` hops **raw 0-based**. So one group/cross-repo trace segment reports 1-based `from`/`to` alongside 0-based `hops[]` for the same symbols (before this PR both were 0-based). Fix: convert the whole group-trace response at a single boundary, or keep the port raw and convert once at the response edge. _(Codex adversarial + Claude ce-api-contract.)_ [code-read]

magyargergo (commented):
# Tri-review digest — PR #2380 (fix(lbug/mcp): exact symbol content + 0-based storage / 1-based MCP display, #2377 #2379)

**Verdict: well-built, well-tested PR with exactly ONE merge-blocking bug — `query()` double-converts BM25 line numbers (a 2-line fix + 1 test). Everything else is scoped follow-ups.** The refactor's hard part (containment/ID safety, exact-content slicing) is genuinely correct and pinned by new tests.

- **Final verdict: not production-ready** (one P1 blocks; fix is a 2-line deletion + one test)
- **Branch hygiene: clean feature/fix PR**
- **Merge state: checks pending** (MERGEABLE; `BLOCKED` = pending CI, not a conflict)

## Engines
Two independent engines, not five confirmations:
- **Claude** ran both the GitNexus PR swarm (risk-architect, test-ci, docs-dod, branch-hygiene, pr-facts) and the ce-code-review personas (correctness, testing, maintainability, project-standards, adversarial, api-contract).
- **Codex / GPT-5.5** ran the swarm (B1) and adversarial (B3) legs. **The Codex ce leg (B2) FAILED** on a fatal tool-schema error (`Invalid schema for function 'api_impact'`) before producing findings — treat as unavailable. Cross-engine corroboration still holds via B1+B3, both of which independently reached the headline P1 and the group-trace mix.
- The **coordinator reproduced the P1 and group-trace mix first-hand** against the worktree (highest authority; it explicitly reversed an earlier provisional "no double-conversion" read once it traced that `bm25Search` itself converts).

## What's solid (validated — credit)
- **Containment / node-ID safety REFUTED as a risk.** `toZeroBasedLine` is applied ONLY at `startLine:`/`endLine:` property-assignment sites. Every containment + ID use reads the RAW 1-based `.line`: `findOwningProgramName(sec.line)`, `findContainingSection(para.line)`, nested-program `prog.startLine`, `nextLine = sections[i+1].line-1`, and all `L${line}` generateId calls. JCL follows the same pattern. Traced by risk-architect, ce-adversarial, ce-correctness, and the coordinator across every site. [reproduced]
- **The #2379 exact-content regression is genuinely fixed.** 0-based `startLine` → `lines.slice(start, end+1)` yields exactly the symbol span; single-line (start===end) yields one line. The new tests PIN exactness: csv-pipeline's "±1 boundary shift" puts non-blank neighbors DIRECTLY adjacent so any off-by-one at either edge fails; the one-line-symbol case pins start===end; cobol.test's JCL STEP1 asserts `expectedIdx>0` + `lines[startLine].includes('STEP1')` — a rigorous non-line-0 0-based proof. test-ci ran them green (csv-pipeline 24/24, cobol 64/64, markdown-crlf 5/5). [reproduced]
- **BasicBlock join `sym.startLine + 1`** (resolveBlockAnchor queryParams, 3358) correctly targets the 1-based BasicBlock id space and is NOT wrapped in `toDisplayLine`. **`toDisplayLine`'s undefined-passthrough is correct** (`typeof x === 'number' ? x+1 : undefined`; absent fields stay absent through `JSON.stringify`). [code-read]
- Branch hygiene clean, commits well-sequenced, CHANGELOG correctly untouched, no project-standards violations, no bidi/non-ASCII hygiene issues.

## HEADLINE — [P1, BLOCKS MERGE] `query()` double-converts FTS/BM25 line numbers
`bm25Search` applies `toDisplayLine` when building each result row — **local-backend.ts:2259-2260** `startLine: toDisplayLine(sym.startLine || sym[4])` → rows are already 1-based. Those rows flow verbatim through `scoreMap` (data:result, 1823) → `merged` → `sym = item.data` (1969), and `query()` re-applies `toDisplayLine` at **1991-1992** → **a second +1**. Net: a symbol stored 0-based 41 is reported as **43** (want 42).
- **Trigger:** any `query({search_query})` where the symbol is surfaced by the FTS/BM25 retriever. `semanticSearch` leaves its chunk lines RAW, so vector-matched symbols convert once and are correct (42). RRF inserts bm25 `data` first and keeps it when both retrievers hit — so the reported line silently depends on which retriever surfaced the symbol. `query()` is the exploration tool CLAUDE.md tells agents to use.
- **`bm25Search` is called ONLY from `query()`** (grep + GitNexus "Called by: query"), so the fix is local and safe: **drop `toDisplayLine` at 2259-2260** (return raw 0-based), keeping the single conversion at 1991-1992 that both retrievers already share.
- **Why CI was green:** the new mcp-line-display.test only drives `context()` + raw cypher — never `query()`. ce-correctness reproduced the exact `5 →(2259) 6 →(1991) 7` end-to-end.
- **Cross-engine:** Codex B1 [P1] + Codex B3 [high] + Claude risk-architect [P1] + ce-adversarial [P2] + ce-correctness [P1, reproduced] + coordinator reproduction. [reproduced]

## Inline findings (2)
1. **local-backend.ts:2259 — [P1] BM25 double-conversion** (the headline). Drop `toDisplayLine` at 2259-2260; add a `query()` test asserting stored 41 → 42 not 43.
2. **local-backend.ts:739 — [P2] Group/cross-repo trace mixes line bases in one response.** `resolveSymbolForGroup` (739-740, ambiguous 752) now returns 1-based endpoints; in cross-trace.ts `endpointFrom` (518) uses them for `from`/`to`, while `tagHops` (509) copies `port.trace` hops raw 0-based. So one group/cross-repo trace segment shows 1-based `from`/`to` + 0-based `hops[]` — same field name, same call, two bases (before this PR both were 0-based). Cross-engine: Codex B3 [medium] + Claude ce-api-contract + coordinator. Fix: keep the port raw and convert the whole group-trace response at one boundary, or convert `tagHops` too.

## Body findings
- **[P2] No schema/version bump for the storage + semantic change.**
  - `INCREMENTAL_SCHEMA_VERSION` (repo-manager.ts:275 = 5, file NOT in diff) is unbumped. Its own comment says a bump forces a one-time full re-analyze because incremental writeback preserves unchanged-file rows. On upgrade + incremental re-analyze, unchanged COBOL/JCL/markdown/scope files keep their OLD 1-based rows; the new +1 display renders them one line too high, and the repo ends up MIXING 0-based (new) and 1-based (stale) rows. The same stale 1-based data also makes the pdg-impact join (`s.startLine + 1 = fnLine`) miss silently. Recommend bumping to force a full re-analyze (or a mandatory release note). Claude-only (risk-architect + ce-adversarial). [code-read]
  - `PDG_RESULT_VERSION` (pdg-impact.ts:95 = 1) is the one MCP result shape with a version discriminator built for exactly this, documented in tools.ts:432 ("bumps on any breaking change to the PDG result shape"). The `startLine` field it covers changed semantics at pdg-impact.ts:813 (`toDisplayLine(s.startLine)`) with no bump — a consumer pinned to `pdgResultVersion === 1` gets no signal. Claude ce-api-contract. [code-read]
- **[P2/P3] Uneven conversion — new cross-tool and within-tool base inconsistency, under-documented.**
  - `explain`/`pdg_query` share `resolveBlockAnchor`: the PR converted its ambiguous-candidate branch to 1-based (local-backend.ts:3338 `line: toDisplayLine(c.startLine)`) but left the resolved-target `anchor.startLine`/`endLine` raw 0-based (3362-3363). Same shared resolver, same tool, two bases depending on whether the target is ambiguous.
  - Across tools: `context`/`query`/`impact` present 1-based; `trace`/`explain`/`pdg_query` anchors + raw cypher stay 0-based. Not a regression (nothing got worse), but the split is the main design tradeoff. The only doc is resources.ts:453, which names only context/query/impact — it does not enumerate the tools that stay 0-based, and no tool's own tools.ts description mentions the output-line convention at all. Cross-engine: Codex B1 + Claude docs-dod / ce-api-contract / ce-testing / risk-architect. Fix: pick one base per shared resolver and name the 0-based tools in the doc. [code-read]
- **[P2 test gap] The new mcp-line-display.test only covers `context()` + raw cypher.** `query()`, `impact()` (callgraph + pdg), pdg byDepth (pdg-impact.ts:813), the 3338 ambiguous-candidate conversion, and the `toZeroBasedLine`/`toDisplayLine` boundary functions themselves are all unasserted for their numeric output. This is WHY the P1 shipped green — every lane flagged it. Fold into the P1 fix: add a `query()` line-value assertion (stored 41 → 42), plus one `impact()`/anchor case. [reproduced]
- **[P2 maintainability] manifest-extractor.ts:19-23 is a second hand-maintained label list.** `CUSTOM_CONTRACT_RESOLVE_QUERY`'s `labels(n) IN [...]` (line 23) hand-lists 21 labels that must stay = `SYMBOL_NODE_LABELS` minus {Namespace,Variable,Module}, guarded only by the comment at 19 — the exact drift class (#2379) this PR's shared-set refactor eliminated everywhere else. Author documented it as intentional (#2325-test-locked), so it's a hardening nit. Fix: derive the string from `SYMBOL_NODE_LABELS`, or add a unit test asserting the subset relationship. Claude ce-maintainability + coordinator. [code-read]
- **[P2, reproduced] First-line (startLine=0) symbols lose their line number in `context()`/`query()`.** `toDisplayLine(sym.startLine || sym[4])` at local-backend.ts:2259 (query/bm25) and :3250 (context) — the falsy `||` collapses a legitimate 0-based `startLine:0` to `sym[4]` (undefined on object-shaped rows), so `toDisplayLine(undefined)=undefined` and the field is omitted from the response entirely. ce-correctness seeded a `Class` at `startLine:0` and `context()` returned it with `startLine`/`endLine` ABSENT. Although the `||` predates the PR, the PR's 0-based conversion makes `startLine:0` a COMMON real value: it newly fires for EVERY COBOL program `Module` node (`toZeroBasedLine(1)=0`), every markdown line-1 heading, and any symbol on line 1 of its file — a regression from the prior truthy `1`. Fix: `??` instead of `||` at 2259/2260 and 3250/3251 — trivial, and 2259 is the same line as the P1 fix. [reproduced]

## Refuted suspicions (do not report)
- Double-conversion in the NON-bm25 path, containment/ID drift, exact-slice off-by-one, pdg-impact byDepth mis-conversion, BasicBlock join corruption, degenerate/empty-file crash — all attacked and refuted. Test-misregistration of mcp-line-display.test (parallel `default` project) refuted: 20+ existing `withTestLbugDB` tests run there; established practice.

## Pre-existing / out-of-scope (follow-ups only, NOT findings against this PR)
- `rename` path local-backend.ts:4288/4292 (`sym.startLine - 1` under a falsy `if (sym.startLine)` guard) assumes 1-based; a first-line 0-based symbol is skipped/mis-targeted. Predates PR; rename untouched. The PR widens exposure to COBOL programs (now stored startLine=0).
- COBOL COPY-expansion coordinate mismatch (expanded `para.line` vs raw-file `extractContent`) — separate bug, documented in the cobol test comment.
- No tracking issues filed for the trace/explain/pdg_query parity follow-up or the COPY-expansion follow-up (pr-facts). `#2379` is an already-CLOSED PR (not an issue); "Fixes #2377" correctly closes the one open issue — nothing stale after merge.

## CI status
Green except Vercel (auth infra — "Authorization required to deploy", not code). 3 jobs still pending: tests/ubuntu/coverage, windows-latest (platform-sensitive — touches file I/O + line splitting, worth confirming green), Build & Push gitnexus. `mergeStateStatus: BLOCKED` = pending checks, not a conflict; GitHub reports MERGEABLE. No reviews yet.

---
This is an automated multi-tool digest (Claude swarm + ce personas, Codex B1+B3; Codex ce leg degraded). Re-verify each anchor before acting on it.

PR #2381, opened 2026-07-06
fix(web): improve repository dropdown search

### LINKED ISSUE #2419: Web UI repo dropdown treats duplicate repository names as the same identity
### Area

### Summary

The Web UI uses `repo.name` as the repository identity, which makes repository selection ambiguous when multiple indexed repositories share the  same display name.

### Context

This came up while reviewing the repository dropdown search work in PR #2381.

The dropdown can render multiple indexed repositories returned by `/api/repos`, but the UI currently uses `repo.name` for active-state comparison, switching, deletion, and re-analysis state. A unique React row key can stabilize rendering, but it does not solve the underlying repository identity problem.

### Expected behavior

Repositories with the same display name but different paths should be distinguishable in the Web UI.

The display label can remain `repo.name`, but active state, switching, delete, re-analyze, and backend resolution should use a unique repository  identity, such as `repo.path` or a backend-provided canonical repo id.

### Actual behavior

When two indexed repositories have the same `name`:

  - multiple rows can display the same repository name
  - active-state comparison uses `repo.name === projectName`
  - switching passes `repo.name` into `onSwitchRepo`
  - delete and re-analyze flows also track repositories by `repo.name`
  - backend HTTP repo resolution may still treat the `repo` parameter as a name

  As a result, duplicate display names can be rendered, but interactions are ambiguous.

### Steps to reproduce

1. Index two repositories that have the same display name but different paths.
2. Open the Web UI and open the repository dropdown.
3. Observe that both rows display the same repository name.
4. Try to identify the active repository or switch between the duplicate-name rows.
5. Observe that the UI still relies on `repo.name`, so the duplicate rows are not fully disambiguated.

### Environment

_No response_

### Logs / screenshots

_No response_

evander-wang:
## Summary

Makes the repository dropdown usable for large indexed repo lists by adding an internal scroll area and local repository-name search.

Also fixes duplicate repository row rendering by keying rows with the repo path instead of the display name.

## Motivation / context

When many repositories are indexed, the repository dropdown could extend beyond the viewport and hide lower entries.

After adding search, duplicate repository names such as `search_sync`, `tab_server`, and `reels` exposed a React key issue: rows keyed only by `repo.name` could be incorrectly reused during filtering, leaving stale non-matching rows visible.

## Areas touched

- [x] `gitnexus-web/` (Vite / React UI)

## Scope & constraints

**In scope**

- Bound the repository dropdown to the viewport.
- Keep the repository list itself scrollable.
- Add local search over visible repository names.
- Show an empty state when no repository name matches.
- Use stable unique row keys for duplicate repository names.
- Add unit coverage for scrolling, filtering, empty state, and duplicate-name filtering.

- No backend API changes.
- No changes to repository indexing or repo metadata.
- No changes to repository switching, deleting, or re-analyzing behavior.
- No graph rendering changes.

## Implementation notes

- Repository search is local-only and case-insensitive.
- Filtering intentionally matches only `repo.name`, not `repo.path`, because the UI displays repository names and hidden path matches were confusing.
- Repository rows now use `repo.path || repo.name` as the React key to avoid stale DOM reuse when multiple repos share the same name.
- The dropdown keeps fixed header/search controls while the result list scrolls internally.

## Testing & verification

- [x] `cd gitnexus-web && npm test` *(if web changed)*
  - `npm test -- test/unit/bug-fixes.test.ts test/unit/header.test.tsx test/unit/i18n.test.tsx`
- [x] `cd gitnexus-web && npx tsc -b --noEmit` *(if web changed)*

Additional checks run:

- `git diff --check`
- GitNexus `impact` on `Header`: LOW
- GitNexus `detect_changes`: Header-only affected flow
- Pre-commit hook: lint-staged + `gitnexus-web` typecheck passed

## Risk & rollout

Low-risk Web UI-only change. It affects only repository dropdown rendering and local filtering in the header.

Rollback is a normal revert of this PR. No migration or reindex is required.

## Checklist


magyargergo:
Can you please provide a video of your change? 🙏

evander-wang:
> Can you please provide a video of your change? 🙏

How does this look?

image
image

gitnexus-web/src/components/Header.tsx
```diff
@@ -196,145 +207,168 @@ export const Header = ({
...
+                                key={repo.path || repo.name}
+                                className={`group flex items-center gap-2 px-4 py-2 transition-colors ${
+                                  repo.name === projectName
+                                    ? 'border-l-2 border-accent bg-accent/10'
+                                    : 'hover:bg-hover'
+                                }`}
+                              >
+                                <button
+                                  onClick={() => {
+                                    if (repo.name !== projectName) onSwitchRepo?.(repo.name);
```
magyargergo, line 262: This still treats `repo.name` as the unique repo identity, even though the row key was changed to `repo.path || repo.name` to handle duplicate display names. With two indexed repos that share the same name, both rows can compare equal to `projectName`, and this click will pass the ambiguous name into `onSwitchRepo`. Can we switch/track active state using a unique repo identifier here (likely `repo.path`, or a canonical id from the backend), or add a regression test that documents this PR as render-only for duplicate names?

koriyoshi2041:
I did a focused local check on the web-only path here.\n\nCommands run from this PR head:\n\n```bash\ncd gitnexus-web\nnpm test -- test/unit/bug-fixes.test.ts test/unit/header.test.tsx test/unit/i18n.test.tsx\nnpx tsc -b --noEmit\ngit diff --check origin/main...HEAD -- gitnexus-web/src/components/Header.tsx gitnexus-web/src/locales/en/header.json gitnexus-web/src/locales/zh-CN/header.json gitnexus-web/test/unit/header.test.tsx\n```\n\nResults: the targeted web tests passed (3 files / 27 tests), typecheck passed, and diff whitespace check passed.\n\nThe current red CI looks like the Windows platform-sensitive Vitest job timing out after 15 minutes rather than a failing assertion: the sticky report shows 14223 passed / 0 failed / 56 skipped, and the Windows log ends with `vitest timed out after 15 minutes`.

koriyoshi2041:
Small follow-up after #2394 landed: the remaining red check here is still the older Windows platform-sensitive timeout shape, not a web assertion failure.

Since #2394 is now merged and its CI gate passed, rebasing this branch onto current `main` should be the cleanest way to separate this web change from that unrelated runner timeout. My earlier focused web-path check still looks valid from this PR side.

evander-wang:
> Small follow-up after #2394 landed: the remaining red check here is still the older Windows platform-sensitive timeout shape, not a web assertion failure.
> 
> Since #2394 is now merged and its CI gate passed, rebasing this branch onto current `main` should be the cleanest way to separate this web change from that unrelated runner timeout. My earlier focused web-path check still looks valid from this PR side.

Thanks for checking.

gitnexus-web/src/components/Header.tsx
evander-wang: > This still treats `repo.name` as the unique repo identity, even though the row key was changed to `repo.path || repo.name` to handle duplicate display names. With two indexed repos that share the same name, both rows can compare equal to `projectName`, and this click will pass the ambiguous name into `onSwitchRepo`. Can we switch/track active state using a unique repo identifier here (likely `repo.path`, or a canonical id from the backend), or add a regression test that documents this PR as render-only for duplicate names?

You’re right that `key={repo.path || repo.name}` only stabilizes React rendering and does not solve duplicate repository identity.

I opened #2419 to track the full duplicate-name fix, because it is larger than this search-only PR. The complete fix needs to update active-state tracking, `onSwitchRepo` / `connectToServer`, delete/re-analyze flows, and backend repo resolution to use a canonical repository identity instead  of `repo.name`.

For this PR, I’ll keep the behavior scoped to repository search by name.

PR #2383, opened 2026-07-06
fix(fts): diagnose Windows FTS missing-dependency load failures (#2374, Phase 1)

### LINKED ISSUE #2374: FTS exists, but it still reports that LadybugDB does not exist.

magyargergo:
## What & why

Follow-up to #2374 (continues #2375). PR #2375 made the real LadybugDB `LOAD` error visible; the rc.4 reproduction then confirmed the remaining defect: on Windows the FTS extension file installs fine but `LoadLibrary` fails with **error 126** (`找不到指定的模块` / "the specified module could not be found") because the CDN-built `libfts.lbug_extension` dynamically imports OpenSSL 3 / MSVC 14 DLLs that ship nowhere. For that class, telling the user to reinstall/redownload is wrong — the file is present; a **runtime dependency** is missing — and it traps them in a query → repair → doctor loop.

This is **Phase 1 (diagnosis)** of a planned three-part fix. It makes the error *actionable* without yet shipping any binary.

## Changes

- **New `classifyExtensionLoadError` (`src/core/lbug/extension-load-error.ts`)** — a pure-string, lbug-free four-way classifier (`missing_file` / `corrupt_file` / `missing_dependency` / `unknown`). It stays free of any `@ladybugdb/core` import so `native-check.ts` can use it and remain dynamic-import-safe.
  - **Windows catch-all guard:** LadybugDB wraps *every* Windows load failure in the same `Failed to load library … needed by extension` text, so `missing_dependency`'s *confident* remedy keys **strictly on the specific error-126 signal**, never that generic wrapper. Error 127 (wrong OpenSSL minor), 5 (AV/permission lock), and 1114 (DllMain failure) carry only that language-independent wrapper with a localized tail we do not enumerate, so they fall to a **hedged `missing_dependency`** remedy — non-committal (it points the user at their own OS error and notes reinstalling usually won't help), never a confident wrong instruction. A truncated/wrong-arch file (error 193) → `corrupt_file`. This avoids the over-broad branch that would re-trap users.
- **Wired into all four remedy surfaces** — `doctor`, the `--repair-fts` error, the analyze degrade log, and `ftsDegradedWarning()`. For the Windows missing-dependency class they now emit an accurate remedy (MSVC 2015–2022 redistributable first, per DuckDB's canonical answer for this exact error, then OpenSSL 3) instead of the old reinstall-over-network guidance. Other classes keep their existing routing. Client-facing path redaction is preserved.

## Scope / follow-ups

- **Phase 2 (ship OpenSSL 3 DLLs on Windows)** — deliberately *not* in this PR. It is gated on resolving the loader-search mechanism in-house on a Windows runner (next-to-extension placement is preferred over a `process.env.PATH` prepend, which is both a CWE-427 search-path-widening risk and ineffective across the separate `serve`/`query`/`doctor` processes). It also requires an MSVC-runtime preflight, since shipping OpenSSL alone can't repair an MSVC-missing host.
- **Phase 3 (upstream)** — a LadybugDB request to static-link OpenSSL into the extension (as DuckDB does for `httpfs`) or co-locate the DLLs on the CDN, which is the permanent cure.

The classifier is also the safety net for Phase 2: a future OpenSSL/MSVC ABI bump that outstrips a shipped DLL set surfaces as a named-missing-dependency doctor message rather than a silent break.

## Verification

- 60 unit tests green (`extension-load-error` ×17, `fts-degraded-warning`, `doctor-format`, `fts-indexes`, `run-analyze-fts-repair`, `native-check-probe`); `tsc` and `eslint` clean.
- The broken-file `doctor` e2e (`fts-extension-e2e.test.ts`) verifies the classified remedy end-to-end through the real CLI, including the catch-all negative (a corrupt file must not trigger the VC++/OpenSSL remedy).

Closes part of #2374 (Phase 1). No CHANGELOG edit (owned at release time).


koriyoshi2041:
Pulled the current head (`9624fd4`) locally for a focused pass.

Commands run:

```bash
cd gitnexus
npx vitest run test/unit/extension-load-error.test.ts test/unit/fts-degraded-warning.test.ts test/unit/doctor-format.test.ts test/unit/fts-indexes.test.ts test/unit/run-analyze-fts-repair.test.ts test/unit/native-check-probe.test.ts --maxWorkers=2 --testTimeout=120000
npx prettier --check src/cli/doctor.ts src/core/lbug/extension-load-error.ts src/core/run-analyze.ts src/core/search/fts-indexes.ts test/integration/fts-extension-e2e.test.ts test/unit/extension-load-error.test.ts test/unit/fts-degraded-warning.test.ts
npm run build --if-present   # in gitnexus-shared
npx tsc --noEmit
npx eslint src/cli/doctor.ts src/core/lbug/extension-load-error.ts src/core/run-analyze.ts src/core/search/fts-indexes.ts test/integration/fts-extension-e2e.test.ts test/unit/extension-load-error.test.ts test/unit/fts-degraded-warning.test.ts
```

Results:

- focused unit tests: `6 passed`, `60 passed`
- `gitnexus-shared` build: passed
- typecheck: passed
- eslint: 0 errors, 4 existing-style warnings in `run-analyze.ts`
- prettier reproduces the current CI format failure in `extension-load-error.ts`, `run-analyze.ts`, and `fts-indexes.ts`

I also ran `test/integration/fts-extension-e2e.test.ts`; on this macOS checkout it skipped all 11 cases because the FTS-extension e2e precondition was unavailable, so I would not count that as coverage. I did not find another blocker in the classifier/remedy path during this pass.

gitnexus/src/core/lbug/extension-load-error.ts
```diff
@@ -0,0 +1,340 @@
...
+ */
+export function diagnoseExtensionLoad(reason: string | undefined | null): ExtensionLoadDiagnosis {
+  const text = reason ?? '';
+  const stringResult = classifyExtensionLoadError(text);
+  const fileState = inspectExtensionBinary(extractExtensionPath(text));
+
+  if (fileState === 'corrupt') {
+    return { kind: 'corrupt_file', remedy: CORRUPT_FILE_REMEDY };
+  }
+  if (fileState === 'valid') {
```
magyargergo, line 343: **[P2 · reproduced] The `valid` branch discards a correct `corrupt_file` verdict → wrong remedy for a truncated download.**

When the structural probe returns `valid`, this branch unconditionally returns `missing_dependency`, throwing away a `corrupt_file` verdict that `classifyExtensionLoadError` already computed. `inspectExtensionBinary` reads only the first 4096 bytes, so a multi-MB extension truncated *after* its header reads `valid` — lbug reports "file too short" / "%1 is not a valid Win32 application" (error 193) → the string classifier correctly says `corrupt_file` → this override flips it to "install VC++/OpenSSL, reinstalling will NOT help" when re-downloading (`--repair-fts`) is the actual fix. That re-creates the #2374 "no way out" loop, and it contradicts this file's own docstring (lines 46-53, 148-153) which promises a truncated download is caught as `corrupt_file` first. Reproduced independently by Claude ce-correctness + ce-adversarial and Codex adversarial.

**Fix:** honor an explicit loader-corruption report before defaulting to dependency — add `if (stringResult.kind === 'corrupt_file') return stringResult;` at the top of this branch. (Localized non-English tails classify as *hedged*, not `corrupt_file`, so they still fall through to `missing_dependency` as intended.) `[reproduced]`

gitnexus/src/core/run-analyze.ts
```diff
@@ -1343,7 +1352,16 @@ export async function runFullAnalysis(
...
     } else {
-      log(FTS_UNAVAILABLE_MESSAGE);
+      // Append the classified remedy when a runtime dependency is missing
+      // (#2374) — the generic "install it with network access" guidance in
+      // FTS_UNAVAILABLE_MESSAGE is wrong for that class (the file is present).
+      const ftsReason = getExtensionCapabilities().find((c) => c.name === 'fts')?.reason;
+      const { kind, remedy } = diagnoseExtensionLoad(ftsReason);
+      log(
+        kind === 'missing_dependency'
+          ? `${FTS_UNAVAILABLE_MESSAGE} ${remedy}`
```
magyargergo: **[P2 · code-read] Self-contradictory guidance on the main analyze surface.**

For the `missing_dependency` branch this logs `FTS_UNAVAILABLE_MESSAGE` (lines 302-306: "…installed once with **network access** … or pre-installed") immediately followed by the remedy ("…**Reinstalling the extension will NOT help.** Install the Microsoft Visual C++…"). A Windows error-126 user — the core #2374 case — reads "install with network access" and "reinstalling will NOT help" in one line. The comment just above (1355-1357) even states the generic guidance "is wrong for that class (the file is present)," yet the code still prepends the full message. Corroborated cross-engine (Claude docs-dod + Codex ce-code-review correctness).

**Fix:** for the `missing_dependency` branch, use a class-specific degraded-FTS prefix that omits the install / network-access / pre-install line, then append `remedy`. `[code-read]`

gitnexus/src/core/search/fts-indexes.ts
```diff
@@ -24,10 +25,19 @@ export const ftsDegradedWarning = (): string => {
   const fts = getExtensionCapabilities().find((c) => c.name === 'fts');
   if (fts && !fts.loaded) {
     const reason = fts.reason ? redactPaths(fts.reason).replace(/\.$/, '') : undefined;
+    // A missing *runtime dependency* (Windows error 126, etc.) is not healed by
+    // reinstalling (#2374) — surface the classified remedy instead of the
+    // generic reinstall tail. Classify the raw reason (the class words survive
+    // redaction), but only ever surface the redacted text.
+    const { kind, remedy } = diagnoseExtensionLoad(fts.reason);
```
magyargergo: **[P2 · reproduced] `ftsDegradedWarning()` (pure at BASE) now does uncached synchronous disk I/O on every degraded request.**

`diagnoseExtensionLoad(fts.reason)` → `inspectExtensionBinary` → `openSync`/`readSync(4096)`/`closeSync`, with no cache and no timeout. This function is called per-request from `server/api.ts:1322` (`/api/search`, which has no rate limiter) and `mcp/local/local-backend.ts:2107` (MCP `query`); the degraded state is the steady state this PR targets, so every degraded search pays a blocking main-thread read — amplified on a slow/roaming/NFS home dir (the same enterprise setups where VC++/AV failures cluster), where one `openSync` freezes the event loop for all clients. `fts.reason` is byte-stable for the process lifetime, so the read is pure repeated waste, and it defeats the very hang-safety `native-check.ts:130` was hardened for. Reproduced by Claude ce-reliability / ce-adversarial + Codex ce-code-review reliability.

**Fix:** cache the `ExtensionLoadDiagnosis` in `ExtensionCapability` at `markUnavailable` time (extension-loader.ts) and read it here; or call the no-I/O `classifyExtensionLoadError` on the request path and reserve the disk-touching `diagnoseExtensionLoad` for one-shot surfaces (doctor / analyze / `--repair-fts`). `[reproduced]`

magyargergo (commented):
# PR #2383 — Two-Engine Tri-Review Digest

**fix(fts): diagnose Windows FTS missing-dependency load failures (#2374, Phase 1)**
Base `fbffa96554b6cba61b405163a80bcc8853b0cbbc` → head `354f8cfbe850ad3790ddab174c0e3a66db3a0e02` · 9 files (+819 / −8) · 5 source + 4 test.

## Headline verdict

**Not a hard merge-blocker, but three concrete P2s should be fixed first.** The classifier design is sound and the security/never-throws contracts hold, but a `valid`-branch override can flip a correct "re-download" verdict into "install VC++, reinstalling will NOT help" for a truncated-but-header-intact download (F1) — re-creating the exact #2374 "no way out" loop in the very truncated-download case the module's own docstring claims to handle. Two more P2s (self-contradictory analyze message F2; per-request blocking disk I/O F3) are concrete and inline-worthy. The rest are body-level correctness/maintainability/test-coverage notes.

**Blocks merge:** none are hard blockers. F1/F2/F3 are should-fix-before-merge (all three are message-quality/perf, not crashes or data loss); F4/F5/F6 and the hardening notes are safe as follow-ups. The primary #2374 path (present, valid DLL + missing VC++/OpenSSL) is handled correctly.

**Engine breakdown — cross-engine corroboration IS available (all three Codex legs produced real findings):**
- **Engine A (Claude):** GitNexus swarm (risk-architect, test-ci, security-boundary, docs-dod — inline) + ce-code-review personas (correctness, adversarial, security, reliability, testing, maintainability, project-standards).
- **Engine B (Codex / GPT-5.5, xhigh):** B1 swarm + B2 ce-code-review (4 sub-reviewers) + B3 adversarial. All three legs non-empty and independently reached F1/F3/F4.
- **Coordinator** independently verified F1/F2/F3/F6, branch hygiene, and the environmental e2e explanation against the worktree — treated as ground truth.

---

## Credit what's solid (validation is a feature)

- **The diagnosis design is the right idea.** A shared `diagnoseExtensionLoad` routing all four surfaces (doctor, `--repair-fts`, analyze-degrade log, `ftsDegradedWarning`) through one classifier is exactly the de-duplication the prior architecture called for. Confirmed all four call sites route through it (Codex B1 + Claude maintainability + coordinator).
- **The language-independent structural layer earns its complexity.** String-only classification fails on non-English Windows locales; the PE/ELF/Mach-O magic+arch sniff is a genuinely novel-for-this-repo helper (no pre-existing binary-sniffer duplicated — Claude maintainability grepped `src/`), exercised by both synthetic and real-binary tests.
- **Security boundary is sound — every enumerated concern refuted, cross-engine.** All 7 `*_REMEDY` values are static constants with **no template interpolation** (grepped for `${` — none), so no path/reason byte rides out to the HTTP `/api/search` or MCP surface; `redactPaths` unchanged; regexes linear (ReDoS refuted, measured ~0.6–7 ms/1 MB); the CJK literal decodes to exactly the legit Simplified-Chinese error-126 string (no bidi/zero-width/BOM); `aka.ms/vs/17/release/vc_redist.x64.exe` is the genuine Microsoft endpoint. (Claude security + risk-architect + adversarial + Codex B1.)
- **Never-throws + fd-safety + degrade-not-throw contracts HOLD.** `inspectExtensionBinary` wraps open/read/close, `finally { closeSync }` always runs, all `buf.readUIntXX` bounds-guarded against a fixed 4096-byte alloc (no RangeError reachable), `diagnoseExtensionLoad` provably cannot throw → the analyze degrade path stays graph-only, never converts a degrade into a crash. (Claude reliability traced every path + risk-architect + coordinator.)
- **Branch hygiene clean.** 9 coherent files, no unrelated churn/blobs, `git diff --check` clean, no bidi controls, MERGEABLE (a merge-from-main commit present but merge-safe).
- **Real CI is green** on the paths this PR touches (see CI section).

---

## Findings — most severe first

### F1 — [P2] [reproduced] — INLINE — `valid`-branch override discards a correct `corrupt_file` verdict → wrong "reinstalling will NOT help" remedy for a truncated download
- **File:** `gitnexus/src/core/lbug/extension-load-error.ts:323-331` (the `fileState === 'valid'` branch of `diagnoseExtensionLoad`).
- **Problem:** When the structural probe returns `valid`, the branch unconditionally returns `kind: 'missing_dependency'`, discarding a `corrupt_file` string verdict the classifier already computed correctly.
- **Trigger:** A multi-MB extension download truncated **after** its 4 KB header (the common shape for an interrupted network install). `inspectExtensionBinary` reads only `BINARY_HEADER_BYTES = 4096`, so the intact header reads `valid`; lbug reports "file too short" / "%1 is not a valid Win32 application" (Windows error 193, path-wrapped) → string classifier correctly says `corrupt_file` → the `valid` branch overrides it to `missing_dependency`. The user is told to install the VC++ 2015-2022 Redistributable / OpenSSL 3 and that "Reinstalling the extension will NOT help" — when re-downloading (`analyze --repair-fts`) is precisely the fix. This is the same "no way out" loop #2374/#2375 exist to kill, at the very entry point every surface calls.
- **Evidence:** Reproduced independently three ways — Claude correctness (real host binary's first 4 KB as an intact head + "file too short"/"not a valid Win32"/"truncated" tails → `diagnose=missing_dependency`), Claude adversarial (20-byte valid-ELF-magic file + real ld.so "file too short" → `missing_dependency`), and Codex B3 (same 20-byte ELF + large-DOS-stub PE). The file's own docstring (lines 46-53, 148-153) **promises** truncation is "caught here [corrupt_file], before the missing-dependency branch" — the structural override falsifies that promise.
- **Fix:** In the `valid` branch, honor an explicit loader corruption report before defaulting to dependency: `if (stringResult.kind === 'corrupt_file') return stringResult;` then keep the existing dependency logic. A header-only `valid` is weaker evidence than the loader's whole-file "too short / not a valid Win32 app" report. (Fix is safe: localized tails classify as hedged, not `corrupt_file`, so they still fall through to `missing_dependency`.)
- **Corroboration:** Claude ce-correctness (reproduced) + ce-adversarial (reproduced, conf 90) + risk-architect; Codex B3 (high) + B1 (P3). **Strong — cross-engine + reproduced.**

**F1 secondary family (same `valid`/`absent` over-trust root — BODY-only, lower confidence, note as a cluster):**
- **[P3] [code-read]** `valid` is too permissive on *type*: `classifyELF/PE/MachO` check magic + CPU arch but **not** that the file is a shared library (ELF `ET_DYN`, PE DLL characteristics, Mach-O dylib/bundle). Any valid host **executable** at a `*.lbug_extension` path reads `valid` → `missing_dependency`. The new real-binary test (`extension-binary-real.test.ts:87-91`) copies `process.execPath` (the Node executable, not a `.so`) to a `*.lbug_extension` path and asserts `missing_dependency` — **baking the false positive into the suite**. Codex B2 #3 (P2, conf 75). Consider validating library type and fixing that test assertion.
- **[P3] [reproduced]** `classifyPE` returns `corrupt` when a valid PE's `e_lfanew > 4090` (`peOffset + 6 > bytesRead` on the 4096-cap read) → tells the user to re-download a fine DLL. Harmless for the *real* fts DLL (MSVC `e_lfanew` ≈ 0xE8–0x200, proven `valid` by the real-binary test), so latent. Claude correctness (reproduced) + Codex B1/B3.
- **[P3] [reproduced]** Fat Mach-O (`0xcafebabe`) returns `valid` **unconditionally** without checking the host slice exists → a fat binary missing the host arch is called `missing_dependency` instead of wrong-arch/corrupt. Claude adversarial (reproduced) + Codex B1.
- **[P3] [reproduced]** Windows error 126 on an **absent/AV-quarantined** file: `inspectExtensionBinary` sees `absent`, but the `absent` branch defers to the string classifier which returns the **confident** VC++ remedy asserting "the file is present." The code comment (lines 333-338) acknowledges the stale/racy trade-off; the remedy is confident, not hedged, for this ambiguous case. Claude adversarial (reproduced, conf 60). Lower priority — the primary #2374 case (present valid DLL, missing VC++) is handled correctly.

### F2 — [P2] [code-read] — INLINE — Analyze degrade log concatenates contradictory guidance on the main analyze surface
- **File:** `gitnexus/src/core/run-analyze.ts:1361` (emits `` `${FTS_UNAVAILABLE_MESSAGE} ${remedy}` ``), with `FTS_UNAVAILABLE_MESSAGE` at `302-306`.
- **Problem:** For the `missing_dependency` branch the log prints `FTS_UNAVAILABLE_MESSAGE` ("…will be disabled until the LadybugDB FTS extension is **installed once with network access** … or pre-installed") immediately followed by the remedy ("…**Reinstalling the extension will NOT help.** Install the Microsoft Visual C++…"). Self-contradictory in one line, on the primary analyze surface — preserving exactly the confusing guidance this PR is meant to remove.
- **Trigger:** Any Windows host that fails to load a present, valid FTS DLL because VC++/OpenSSL is missing (the core #2374 case) running `analyze`.
- **Evidence:** Verified at the worktree; the code comment at `1356-1358` even states the generic install guidance "is wrong for that class (the file is present)" — yet the code still prepends the full `FTS_UNAVAILABLE_MESSAGE` verbatim.
- **Fix:** For `missing_dependency`, emit a class-specific degraded-FTS prefix that omits the install/network-access/pre-install line, then append the remedy.
- **Corroboration:** Claude docs-dod (P1) + Codex B2 #2 correctness (P2, conf 100). **Strong — cross-engine.**

### F3 — [P2] [reproduced] — INLINE — `ftsDegradedWarning()` (was pure) now does uncached synchronous disk I/O on every degraded HTTP/MCP request
- **File:** `gitnexus/src/core/search/fts-indexes.ts:32` (`diagnoseExtensionLoad(fts.reason)` → `inspectExtensionBinary` → `openSync`/`readSync(4096)`/`closeSync`).
- **Problem:** At BASE `ftsDegradedWarning()` did **zero** I/O (verified via `git show fbffa96`). It is now called per-request from `src/server/api.ts:1322` (`/api/search`, which has **no** rate limiter) and per-query from `src/mcp/local/local-backend.ts:2107`, performing an uncached synchronous file read on the Node main thread — no timeout, no memoization — for a warning string. The degraded state is the steady state this PR targets, so every degraded search pays it.
- **Trigger:** Sustained search/query QPS against an FTS-degraded server; amplified if the extension path is on a slow/hung/roaming filesystem (NFS/Citrix Windows home dirs — the same enterprise environments where VC++/AV failures cluster), where a single blocking `openSync` freezes the whole event loop for all clients.
- **Evidence:** Claude reliability confirmed the diff adds the only I/O to a previously I/O-free per-request function and contrasted it with the sibling `native-check.ts:130` probe, whose docstring says an unresponsive extension file "must never freeze `doctor`" — this new path defeats that same hardening one statement later at `doctor.ts:152`. `fts.reason` is byte-stable for process life (cached in `ExtensionManager` via `markUnavailable`), so the read is pure repeated waste.
- **Fix:** Cache the `ExtensionLoadDiagnosis` once in `ExtensionCapability` at `markUnavailable` time (extension-loader.ts) and have the warning read it; or use only the no-I/O string classifier (`classifyExtensionLoadError`) on the request path, reserving the disk-touching `diagnoseExtensionLoad` for one-shot surfaces (doctor / analyze / `--repair-fts`).
- **Corroboration:** Claude ce-reliability (P1, conf 80) + risk-architect (P2) + ce-adversarial (reproduced, conf 80); Codex B2 #1 reliability+adversarial (P1, conf 75). **Strong — cross-engine + reproduced.**

### F4 — [P2] [code-read] — BODY — Non-126 Windows errors (5/1114/127) route to `missing_dependency`, contradicting the PR body
- **File:** `gitnexus/src/core/lbug/extension-load-error.ts:175-176` (`LOAD_FAILURE_WRAPPER` fallback → `kind: 'missing_dependency'`, `HEDGED_LOAD_FAILURE_REMEDY`).
- **Problem:** After the specific 126/POSIX regexes miss, any `Failed to load library` wrapper (Windows error 5 "Access is denied", 1114 "DLL init failed", 127 "procedure not found") returns `missing_dependency`. The unit tests explicitly enshrine 127/5/bare-wrapper as `missing_dependency`. **The PR body states these route to `unknown`** — a doc/code mismatch. Consumers key on `kind === 'missing_dependency'` to suppress the generic repair/doctor tail, so permission/AV/DllMain failures are not distinguishable from a real missing runtime.
- **Nuance (why P2, not P1):** The *code matches its own comment* and the fallback remedy is **hedged/honest** ("never prescribes the wrong fix") — so there is no confidently-wrong output here, unlike F1. The defect is the stale PR body plus the loss of a distinct hedged/unknown class.
- **Fix:** Correct the PR body; and consider a separate hedged/`unknown-load-failure` kind so 5/1114/127 aren't labeled `missing_dependency`, with consumers branching on it.
- **Corroboration:** Codex B3 (medium) + B1 risk-architect (P2, High) + B1 docs-dod; coordinator verified the body text. **Moderate — Codex-only, but verified [code-read].**

### F5 — [P2/P3] [code-read] — BODY — Most consumers discard the per-class remedies; two duplications lack a drift guard
- **Files:** `run-analyze.ts:692-699` (`--repair-fts`), `run-analyze.ts:1358-1364` (degrade log), `fts-indexes.ts:32-36` (`ftsDegradedWarning`) — three of four consumers only forward `remedy` when `kind === 'missing_dependency'`, falling back to pre-existing generic tails for `missing_file`/`corrupt_file`/`unknown`. So `MISSING_FILE_REMEDY`/`CORRUPT_FILE_REMEDY`/`UNKNOWN_REMEDY` are live only via `doctor.ts` + unit tests — undercutting the module's stated "every surface gets the right remedy" goal. **[P2, conf 68]**
- **[P2, conf 65]** VC++ name+URL duplicated verbatim between `WINDOWS_MISSING_DEPENDENCY_REMEDY` (116-121) and `STRUCTURAL_MISSING_DEPENDENCY_REMEDY` (186-190), no shared constant, and the structural one's URL is asserted by no test → silent drift risk. Factor a shared `VC_REDIST_INSTALL_HINT`.
- **[P2, conf 60]** `FILE_CORRUPTION_SIGNATURES` (extension-load-error.ts:56-62) is byte-identical to `scripts/install-duckdb-extension.mjs:17-23` (7 regexes) — deliberate & disclosed (the `.mjs` can't import the `.ts`), but no parity test; a future edit to one silently desyncs FORCE-install verb from remedy classification. Add a parity test.
- **Corroboration:** Claude ce-maintainability only. **Moderate — single-engine but clean code-read; quality/maintainability, not a merge blocker.**

### F6 — [P2/P3] [code-read] — BODY — Test-coverage gaps (this is why F2 slipped)
- **(a) [P1-flagged by lane]** 3 of 4 consumer surfaces have **no test** reaching the `missing_dependency` remedy branch. The e2e doctor/`--repair-fts` tests use `makeHome('broken')` → always `corrupt_file`, and both `assert not.toContain('Visual C++')` — i.e. they assert the *false* branch. The only end-to-end `missing_dependency` assertion (`fts-degraded-warning.test.ts`) uses a non-existent path so the result comes via the string-classifier fallback, never the structural `valid` path. A typo/inversion of the `kind` check in `doctor.ts` or either `run-analyze.ts` site would go uncaught. Claude ce-testing (P1) + docs-dod; Codex B2 testing-gap.
- **(b) [P2]** Structural header branches untested: PE `e_lfanew`-beyond-window, recognized-platform/unmapped-arch, fat Mach-O `0xcafebabe`, PE valid-MZ-but-garbage-signature. Claude ce-testing + Codex B1/B3.
- **(c) [P3]** `extension-binary-real.test.ts` imports `@ladybugdb/core` (line 14) but is **not** in the vitest `lbug-db` project include list (TESTING.md:68 rule). Practical hazard LOW — it only reads `lbug.VERSION`, no DB open/mmap, so it runs harmlessly in `default`. Coordinator-verified P3. Claude ce-testing + ce-project-standards + Codex B2.
- **(d) [P3]** The two `.skipIf(!lbugNative)`/`.skipIf(!installedFts)` guards don't honor `GITNEXUS_REQUIRE_FTS=1` (CI sets it on ubuntu/coverage + cross-platform), so the real-binary structural assertions could silently vanish from a green require-FTS lane — the #2299 trap. Mitigated: the always-on `expect(inspectExtensionBinary(process.execPath)).toBe('valid')` assertion (line 70) does run everywhere. Claude ce-testing + ce-project-standards + Codex B1/B2.
- **Corroboration:** Claude ce-testing + project-standards + docs-dod; Codex B1 + B2. **Cross-engine on (b)(c)(d); (a) is the highest-value gap.**

---

## Refuted suspicions + hardening notes (credit — validation is a feature)

- **Info-leak on HTTP/MCP — REFUTED** (cross-engine): static remedy constants, no interpolation, `redactPaths` intact.
- **ReDoS — REFUTED [reproduced]**: linear regexes, measured ~0.6–7 ms/1 MB.
- **Degrade → throw — REFUTED [reproduced]**: `diagnoseExtensionLoad` cannot throw; graph-only degradation preserved.
- **Path-steering to FIFO/device + fd exhaustion — REFUTED**: inspected path is always the server's own configured extension path from lbug's cached reason, never request-controlled; fd released in `finally`.
- **Windows 126-vs-127/5/1114 string-tail misfire — REFUTED**: no FILE_CORRUPTION signature matches a 126 tail; the CJK string is 126-specific. (The real 126 issues are F1's structural override and the absent-file path, not the string guard.)
- **Two P3 hardening notes:** `inspectExtensionBinary` is exported without a `.lbug_extension` suffix guard at the export boundary (defense-in-depth); `openSync` has no FIFO/hung-FS timeout (folded into F3's fix).
- **Pre-existing (NOT this PR):** `redactPaths` stops at whitespace → a Windows path with a space in the username (`C:\Users\a b\…`) is only partially redacted on the HTTP warning. The regex is unchanged here, but this PR appends to that surface — suggest adding a regression test for a spaced-username path on the new `missing_dependency` branch. Codex B2 pre-existing + Claude security.

## Environmental note — do NOT report as a finding
- The full `fts-extension-e2e.test.ts` shows 9 failed / 2 passed **in this worktree only**: every failure is `analyze` exiting 1 with `Worker script not found: …/parse-worker.js` (verified in `analyze-direct.log`) because the worktree has no built `dist/` (symlinked node_modules, can't `npm run build`). The PR does not touch the ingestion/worker path; **both `doctor` tests pass** (the PR's actual FTS-diagnosis path). Conclusively environmental, not a regression. (Coordinator + Claude ce-testing both hit the same build wall.)

## CI status
- GitHub reports `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE`. **Vercel failure is authorization-only** (deploy auth), not a code failure. CI Gate + repo CI passed; `tests / macos-latest (platform-sensitive)` and `packaged install smoke (windows-latest / ubuntu-latest)` are **green** — i.e. the platform-sensitive + smoke jobs on this PR head pass. A few jobs pending. No real CI failure attributable to the diff.

---

*Automated multi-tool digest (Claude swarm + ce-code-review; Codex GPT-5.5 swarm + ce + adversarial). Findings are evidence-weighted by cross-engine corroboration; `[reproduced]` = a lane executed a repro, `[code-read]` = verified by reading source at the cited line. F1/F2/F3 are inline-worthy (concrete, verified, on diff lines); F4/F5/F6 and hardening are body-level. Verify each fix against the worktree before acting — several findings are edge/latent (F1 secondary family) rather than the primary #2374 path, which the PR handles correctly.*

PR #2386, opened 2026-07-06
fix: report custom HTTP embedding endpoint failures instead of huggingface download errors (#2385)

### LINKED ISSUE #2385: Custom HTTP embedding endpoint failures are misreported as huggingface.co download errors
### Area

### Summary

When a custom HTTP embedding endpoint (`--embedding-base-url` / `GITNEXUS_EMBEDDING_URL`) is unreachable, the CLI reports it as a **huggingface.co model-download failure** — even though HTTP mode never downloads a model and never contacts huggingface.

### Context

Reported by @kajiev in https://github.com/abhigyanpatwari/GitNexus/issues/628#issuecomment-4891590140. User pointed `--embedding-base-url` at a local server, the request to that server failed, and the CLI told them "The embedding model could not be downloaded / huggingface.co may be unreachable" — sending them down the wrong debugging path.

Root cause (misclassification in the CLI error handler):

- HTTP mode is active (`isHttpMode()` true), so `initEmbedder()` is skipped and **no local/HF download is attempted** — confirmed by the `Using custom embedding endpoint:` line that prints first.
- The first batch POST to `<base-url>/embeddings` fails to connect. `gitnexus-shared/src/integrations/resilient-fetch.ts` rethrows Node's raw `fetch failed` on a terminal network error, which `core/embeddings/http-client.ts` (`httpEmbedBatch`) wraps as `Embedding request failed (<url>, batch N): fetch failed`.
- That error propagates to the CLI catch in `cli/analyze.ts` (~L1647): `isHfDownloadFailure(msg) || msg.includes('Failed to download embedding model')`.
- `isHfDownloadFailure` → `isNetworkFetchError` (`core/embeddings/hf-env.ts:96`) matches **any** string containing `fetch failed` / `ECONNREFUSED` / `ENOTFOUND` / `ETIMEDOUT` / `ECONNRESET`.
- There is **no `isHttpMode()` guard** on that branch, so a custom-endpoint connection failure is attributed to a huggingface download.

### Expected behavior

In HTTP mode, a failure to reach the custom endpoint should surface an endpoint-specific message (the actual URL — masked via `safeUrl` — and the underlying reason: connection refused / timeout / DNS), with no mention of huggingface.co or model downloads.

### Actual behavior

The CLI prints the huggingface-download remediation block (`HF_ENDPOINT=…`, "the model is cached — future runs work offline", etc.), which is irrelevant and misleading when a custom HTTP endpoint is configured.

### Steps to reproduce

1. Point the embedding endpoint at an address where nothing is listening (or a wrong port):
   ```
   gitnexus analyze --embeddings \
     --embedding-base-url http://127.0.0.1:1/v1 \
     --embedding-model some-model --embedding-dims 1024
   ```
2. `Using custom embedding endpoint: http://127.0.0.1:1/v1 (model: some-model)` prints (HTTP mode confirmed).
3. On the first batch the run fails with "The embedding model could not be downloaded / huggingface.co may be unreachable from your network" instead of an endpoint-connection error.

### Environment

- Reporter: GitNexus v1.6.8
- Applies to current `main` as well (code paths cited above are unchanged).

### Logs / screenshots

From #628:
```
Embedding 0/7254{"level":50,...,"embedErr":{},"msg":"❌ embedBatch failed for 8 texts ..."}
  ████████████████████████████████████░░░░  90% | Embedding 0/?
  The embedding model could not be downloaded.
  huggingface.co may be unreachable from your network
```

### Proposed fix

Gate the HF-download branch in `cli/analyze.ts` on `!isHttpMode()`, and/or classify the wrapped `httpEmbedBatch` failures explicitly so the CLI prints an endpoint-connection message (with the masked URL) rather than the HF guidance. The dimension-mismatch path already produces a correct, endpoint-specific message — connection/timeout failures should too.

magyargergo:
## Summary

`gitnexus analyze --embeddings` against a custom HTTP embedding endpoint (`--embedding-base-url` / `GITNEXUS_EMBEDDING_URL`) now reports endpoint failures **as endpoint failures** — surfacing the masked URL and the real reason — instead of the misleading *"The embedding model could not be downloaded / huggingface.co may be unreachable"* remediation. HTTP mode never downloads a model, so the old message sent users down the wrong debugging path (set `HF_ENDPOINT`, check the model cache). Fixes #2385.

## Root cause

`httpEmbedBatch` wrapped every endpoint failure as a plain `Error` whose message contained `fetch failed` / `ECONNREFUSED`. The analyze CLI catch classified errors by matching those substrings (`isHfDownloadFailure` → `isNetworkFetchError`) with no `isHttpMode()` guard, so a custom-endpoint connection failure was attributed to a HuggingFace download.

## What changed

- **Typed error, type-based classification.** The HTTP embedding path (`http-client.ts`) now throws a typed `HttpEmbeddingError` from every endpoint-failure site; the CLI discriminates on the **error type**, not its message text — locale-proof and rewording-proof, matching the codebase's existing `CircuitOpenError` / `err.name === 'TimeoutError'` idiom.
- **Accurate, neutral message.** The endpoint branch surfaces the real reason (masked URL + cause) and covers both *never-reached* failures (connection / timeout / DNS / circuit-open) and *reached-but-failed* ones (4xx/5xx, dimension/shape mismatch, unparseable/empty body) without falsely asserting "unreachable". New `recoveryHint: http-embedding-endpoint-error`.
- **Complete classification.** A reachable-but-wrong endpoint that answers `200` with an HTML/truncated body is now classified as an endpoint failure instead of escaping as a raw `SyntaxError` to the generic stack-dump path.
- **HF branch gated on `!isHttpMode()`**, resolved defensively so a malformed `GITNEXUS_EMBEDDING_DIMS` can't throw out of the error handler.

## Testing

- `test/unit/http-embedder.test.ts` — typed-error classification (instance, cross-realm `name` fallback, negatives) plus type-completeness fences on the connection, timeout, non-OK-status, and non-JSON-body paths.
- `test/unit/analyze-http-endpoint-error.test.ts` — CLI catch routing: endpoint message + masked URL, no huggingface leakage, HF branch suppressed in HTTP mode, real HF path unchanged when HTTP mode is off, unrelated errors not captured, and a regression guard that a dimension mismatch is not mislabeled "unreachable".
- Full-package typecheck + lint clean; sibling embedding/analyze suites green.

Reviewed by a 6-persona pass (correctness, reliability, testing, maintainability, adversarial, project-standards); all findings (2× P2, 3× P3) resolved in this branch.

No production/runtime service impact — this is a CLI diagnostic-message and error-classification change. Validate by running `gitnexus analyze --embeddings --embedding-base-url http://127.0.0.1:1/v1 --embedding-model m --embedding-dims 1024` against a dead endpoint and confirming the endpoint-connection message appears with **no** huggingface text. Failure signal to watch: a new report that custom-endpoint failures still mention huggingface.co.

## CI Failures Unresolved

Three checks are red; none is caused by this diff (all substantive suites pass: Ubuntu/macOS tests, node-floor compat, tree-sitter ABI ×3, coverage, quality lint/format/typecheck, CodeQL, packaged-install smoke).

- **`tests / windows-latest (platform-sensitive)`** — the Windows job hit a global **15-minute vitest wall-clock timeout** (`vitest timed out after 15 minutes`), not a test assertion failure. This is a pre-existing, recurring Windows-runner flake: the identical job also timed out on the unrelated PR `fix/line-numbers-2377-2379` ([CI run 28792120564](https://github.com/abhigyanpatwari/GitNexus/actions/runs/28792120564)), and Ubuntu coverage runs ~15m37s — right at the budget — so the slower Windows runner tips over. This diff adds only 3 small error-handling edits + 2 fast unit-test files; it cannot add 15 minutes to the suite. Re-ran the job twice; the timeout reproduced both times. No code-fix path in this PR. ([failing run](https://github.com/abhigyanpatwari/GitNexus/actions/runs/28807075877/job/85425242051))
- **`Vercel`** — "Authorization required to deploy." Deploy-authorization gate for fork PRs; requires a maintainer to approve the deployment. Infra gate, not a code failure.
- **`CI Gate`** — aggregate gate; red solely as a consequence of the two items above.

Suggested resolution: a maintainer re-runs the Windows job (or the repo raises the Windows vitest timeout / shards the platform-sensitive suite) and authorizes the Vercel deploy.


gitnexus/src/core/embeddings/http-client.ts
```diff
@@ -140,33 +175,56 @@ const httpEmbedBatch = async (
...
+  let data: { data: EmbeddingItem[] };
+  try {
+    data = (await resp.json()) as { data: EmbeddingItem[] };
+  } catch (err) {
+    throw new HttpEmbeddingError(
+      `Embedding endpoint returned an unparseable response (${safeUrl(url)}, batch ${batchIndex})`,
+      { cause: err },
+    );
+  }
+  if (!Array.isArray(data?.data)) {
```
magyargergo: **P3 (completeness gap, non-regression) — cross-engine, reproduced.** The new shape guard only checks the outer array, so a malformed item escapes the typed boundary.

**Trigger:** a reachable-but-wrong endpoint answers HTTP 200 with `{"data":[null]}` (or any `data` array with a literal `null` slot, matching batch count).

**Evidence:** `Array.isArray([null])` is `true`, so `[null]` passes here; then `httpEmbed` runs `new Float32Array(item.embedding)` (~:267) on `item === null` → raw `TypeError`, thrown *outside* `httpEmbedBatch`'s typed try/catch. In analyze, `isHttpEmbeddingError(err)` is `false` and the HF branch is gated off by `!inHttpMode`, so it lands on `writeFatalToStderr('Analysis failed', err)` (analyze.ts:1717) — the raw stack-dump path #2385 closes. This contradicts this module's own JSDoc, which claims a "wrong-shape response body" is covered. Narrow: only `null`/`undefined` items escape (`{}`, strings, numbers → length-0 → typed dimension-mismatch). `httpEmbedQuery` shares the crash at `items[0].embedding`.

**Severity split (honest):** Codex rated this P2/blocking; Claude rated it P3/non-regression (pre-PR `{data:[null]}` already stack-dumped). Flagged inline because it's reproduced by both engines, sits on this changed line, and undermines the PR's stated "all endpoint failures are typed" invariant.

**Fix (~2 lines + test):** `if (!Array.isArray(data?.data) || data.data.some((it) => !it || !Array.isArray(it.embedding)))` → throw `HttpEmbeddingError` ("unexpected response shape"). Add a `{"data":[null]}` fixture on the batch + query paths asserting `isHttpEmbeddingError(err) === true` (this also closes test-gap F1: the shape-check branch currently has zero coverage). [reproduced]

magyargergo (commented):
## Two-engine tri-review — PR #2386 (fix: report custom HTTP embedding endpoint failures, #2385)

**Engines:** Claude ran the GitNexus swarm + ce-code-review persona set; Codex (GPT-5.5 / xhigh) ran swarm + ce + adversarial. All three Codex legs ran live and returned real findings. Coverage is not degraded by the one Claude swarm lane that ended mid-investigation (risk-architect): its domain — regression to other consumers and mergeability — is fully covered by Codex B1, ce-correctness, and the adversarial MCP-consumer probe.

### Verdict: sound, mergeable-on-its-merits

This is a well-executed fix. The typed-error classification (`HttpEmbeddingError` + `isHttpEmbeddingError`, keyed on error **type** not message text) is the correct axis for the reported bug — a connection failure has no status/message to match on — and it is a deliberate, user-driven design choice. Both engines independently confirmed the branch ordering, same-process `instanceof` propagation, and security hygiene. **None of the findings below are regressions**; every one is either a pre-existing edge the PR narrows but doesn't fully close, or test-completeness. The one headline finding is a genuine cross-engine agreement but carries an honest severity split (Codex P2/blocking vs Claude P3/non-regression) — presented below as the coordinator reconciled it: a P3 completeness gap that is still inline-worthy.

### What's solid (credit — much was probed and refuted)

- **Design is correct.** Typed classification over message-matching is locale-proof and survives rewording; the JSDoc explains exactly why `isNetworkFetchError` can't separate a custom-endpoint failure from an HF-download failure. User-driven, not to be re-litigated.
- **Branch ordering verified by both engines.** The two message-matcher branches above the typed branch (`isLocalEmbeddingRuntimeBlockerMessage`, `isMissingLocalEmbeddingStackMessage`) are disjoint from every `HttpEmbeddingError` message; the typed branch fires first, and the HF branch is additionally `!inHttpMode`-gated. [reproduced by ce-correctness]
- **Same-process propagation holds.** `embedding-pipeline` rethrows the *same object* (530/600), `run-analyze` rethrows it (1738), no worker/`structuredClone` boundary — so `instanceof` survives to the analyze catch; the `.name` fallback is belt-and-suspenders. Confirmed by both engines (Claude ce-correctness/ce-reliability + Codex adversarial/swarm).
- **Security clean.** All 11 `HttpEmbeddingError` throws use `safeUrl(url)`; the API key travels only as an `Authorization: Bearer` header, never in error messages (existing test asserts this). 7 security vectors probed and refuted (Claude security-boundary). One residual URL-embedded-credential sub-case surfaced by Codex — see Finding 1.
- **Standards clean.** No CHANGELOG edit, no version bump, conventional `fix:` PR title, new `RecoveryHint` union member wired at its one call site with no exhaustive-switch consumer needing a new case, no `any`. (ce-project-standards: zero violations.)
- **Captive-portal win.** Moving `resp.json()` inside the typed boundary means a 200-with-HTML body now classifies as `HttpEmbeddingError` instead of leaking a raw `SyntaxError` — a real reliability improvement, with a test.
- **Prior 6-persona review's fixes all still hold** (neutral CLI header, shape-check inside the typed try/catch, defensive `inHttpMode`, dropped speculative `status?` field, type-completeness fences).

---

### Findings (all non-blocking)

#### Finding 2 — HEADLINE — malformed `null` response item escapes the typed boundary  ·  P3 (completeness gap, non-regression) · will be inlined
- **Trigger:** a reachable-but-wrong endpoint returns HTTP 200 with body `{"data":[null]}` (or any `data` array containing a literal JSON `null` slot, matching batch count).
- **Evidence:** `gitnexus/src/core/embeddings/http-client.ts:223` — the new shape guard only checks `Array.isArray(data?.data)`. `Array.isArray([null])` is `true`, so `[null]` passes; then `httpEmbed` at `:267` runs `new Float32Array(item.embedding)` on `item === null` → raw **`TypeError`**, thrown *outside* `httpEmbedBatch`'s typed try/catch. In analyze, `isHttpEmbeddingError(err)` is `false`, the HF branch is gated off by `!inHttpMode` (HTTP mode is true), so it falls to `writeFatalToStderr('Analysis failed', err)` at `analyze.ts:1717` — a raw stack dump, the exact failure class #2385 closes. This contradicts the PR's own JSDoc claiming "wrong-shape response body" is covered.
- **Scope (narrow):** only literal `null`/`undefined` items escape. `{}`, strings, and numbers coerce to `Float32Array` length 0 → caught as a *typed* dimension-mismatch. `httpEmbedQuery` shares the crash at `items[0].embedding` but there it's swallowed by `semanticSearch` → silent empty results (still a missed classification).
- **Not a regression:** pre-PR, `{data:[null]}` also produced a raw TypeError → stack dump. This is a completeness gap in the new guard, not new breakage.
- **Cross-engine — strongest signal in the review.** Codex B1 swarm [P2, "blocks merge: yes"] + B3 adversarial [medium] + B2 ce (×4 personas: reliability, testing, correctness, maintainability) AND Claude ce-adversarial [P3, reproduced] + ce-correctness [P3 residual]. **Severity split reported honestly:** Codex = P2/blocking; Claude = P3/non-regression. Coordinator call: **P3 completeness gap, inline-worthy** (reproduced, both-engine, on a changed line, contradicts the code's own contract). [reproduced]
- **Fix (~2 lines + test):** reject any item that isn't an object with an array `embedding` at the shape guard, e.g. `if (!Array.isArray(data?.data) || data.data.some((it) => !it || !Array.isArray(it.embedding)))`, throwing `HttpEmbeddingError` ("unexpected response shape"). Add a `{"data":[null]}` fixture on the batch and query paths asserting `isHttpEmbeddingError(err) === true`.

#### Test-hardening cluster  ·  P2/P3 · Claude ce-testing (all REPRODUCED by reverting the shipped code and confirming all 43 tests stay green)
The PR added a "type-completeness fence" pattern (`expect(isHttpEmbeddingError(err)).toBe(true)`) to some new throw-sites but not uniformly. Each item below was verified by mutating a `HttpEmbeddingError` throw back to `Error` (or removing a defensive `try/catch`) — reintroducing #2385's exact bug class — with zero test failures.
- **F1 (P2) — shape-check branch has zero coverage.** `http-client.ts:223-227`. No test drives a non-array `data.data`. (Overlaps Finding 2's location — the same test fixture closes both.) [reproduced]
- **F3 (P2) — four converted throw-sites assert message-text only, not type.** `http-client.ts:259-262` (vector-count), `:270-278` and `:317-323` (dimension mismatch ×2), `:306` (empty response). These were switched `Error`→`HttpEmbeddingError` in this diff — exactly the "reached-but-failed" class the PR targets — but their tests still use `.rejects.toThrow('...')` only. A future refactor reverting any one would silently reintroduce #2385 with no test failing. Fix: add `expect(isHttpEmbeddingError(err)).toBe(true)` to the four tests. [reproduced]
- **F2 (P2) — terminal 4xx (`!resp.ok`, e.g. 404) throw untested.** `http-client.ts:202-207`. The single most realistic `--embedding-base-url` misconfiguration (wrong path → 404, bad key → 401/403); all existing "status" tests use 500 (different catch branch) or 429/503 (retry-then-succeed). Fix: stub `fetch` → `{ ok:false, status:404 }`, assert typed. [reproduced]
- **F4 (P3) — defensive `inHttpMode` try/catch never exercised.** `analyze.ts:1674-1680`. `isHttpMode` mock only ever returns; nothing makes it throw, so the `catch { return false }` arm is dead from the suite's view. Fix: mock `isHttpMode` to throw + an HF-shaped rejection, assert HF branch still fires. [reproduced]
- **F5 (P3, cosmetic) — stale "Kuzu" fixture name.** `analyze-http-endpoint-error.test.ts:214` uses `new Error('Kuzu write failed')`; the engine is LadybugDB now. Logic is correct (just needs any non-network error); rename to something dependency-agnostic. [code-read]

#### Finding 3 — malformed `GITNEXUS_EMBEDDING_DIMS` env var → undecorated stack dump  ·  P3 · Claude ce-reliability [P2] / ce-correctness [P3 residual], Codex did not emphasize
- **Trigger:** `GITNEXUS_EMBEDDING_URL` + `MODEL` set with a malformed ambient `GITNEXUS_EMBEDDING_DIMS` (e.g. `"1024abc"`) — a plausible Docker/K8s pattern that bypasses the `--embedding-dims` CLI-flag guard at `analyze.ts:1052`.
- **Evidence:** `readConfig()` (`http-client.ts:40-49`) throws a plain `Error`, called at the top of `httpEmbed`/`httpEmbedQuery` and every `isHttpMode()` — *outside* `httpEmbedBatch`'s typed try/catch. So it's neither an `HttpEmbeddingError` nor HF-matched → falls to the raw stack dump. [code-read]
- **Non-regression, and the cause text stays VISIBLE** (`writeFatalToStderr` prints `err.message`, so `"...must be a positive integer, got \"1024abc\""` is shown — just undecorated). The PR's defensive `inHttpMode` try/catch correctly prevents the handler itself from crashing here. Narrow trigger.
- **Fix (optional):** wrap the DIMS-validation throw as `HttpEmbeddingError`, or add a message-recognizing branch mirroring the existing `EMBEDDING_DIMS_ERROR` early-exit, so env-var and CLI-flag paths get the same clean output.

#### Finding 1 — credential-in-URL leak via `reason`  ·  P3 hardening (cross-engine split) · Codex B3 [high] + coordinator [reproduced] vs Claude security clean
- **Trigger:** `GITNEXUS_EMBEDDING_URL=https://user:secret@host/v1`.
- **Evidence (REPRODUCED, coordinator, Node 22.16):** undici hard-fails a credential-bearing URL at `Request` construction with `TypeError: Request cannot be constructed from a URL that includes credentials: https://user:secret@…`. `http-client.ts:195-198` interpolates `reason = err.message` verbatim after `safeUrl(url)` — so the full `user:secret` URL re-enters the message, which analyze prints to stderr and captures in structured logs. `safeUrl` masks only the first arg, not `reason`.
- **Why Claude rated the surface clean:** the security-boundary lane refuted only the API-key-in-header sub-case; it never tested URL-embedded credentials. This is a genuine cross-engine split, not a contradiction.
- **Non-regression** (`reason`-interpolation predates the PR) and narrow (requires creds embedded in the URL, not the default `GITNEXUS_EMBEDDING_API_KEY` header). But this PR re-touches this exact throw and elevates the message to first-class user-facing + logged, and undici fails such a URL on *every* run — so it's the right place to sanitize.
- **Fix:** scrub the configured raw URL / userinfo / query from `reason` before interpolation; add a regression test asserting the message omits `user:secret`. Body note, not inline.

#### Maintainability notes  ·  P3 · Claude ce-maintainability (brief, future-refactor)
- **analyze.ts catch-cascade growth.** `analyze.ts:1546-1768` is now a ~12-branch if/else-if classification cascade whose correctness depends on hand-maintained "must come before X" ordering comments. The PR **correctly follows the existing pattern** (ce-project-standards confirmed), so this is a future-refactor suggestion (extract to an ordered classifier array), not a demand on this PR.
- **Duplicated rationale comment.** `analyze.ts:1670-1673` and `1685-1690` state the "readConfig can throw on malformed DIMS, handler must never throw" fact nearly verbatim, four lines apart — state it once.
- **Out-of-scope follow-up:** `doctor.ts` calls `isHttpMode()` unguarded 3× (~166/179/188) — the same crash this PR defends against in analyze. Worth a follow-up issue, not this PR.

---

### CI status — nothing PR-introduced

`tests / windows-latest` is red because the **whole vitest suite hit the 15-min timeout** on unrelated heavy integration tests (e.g. `cli-limit-e2e ~109s`); the PR's own fast mocked tests never ran before the cap. This is the known Windows full-suite load-flake, not a PR assertion failure. `CI Gate` red is downstream of that timeout. Vercel red is a fork deploy-auth infra gate. **macOS, node-floor (22.14), ubuntu coverage all PASS; typecheck PASS.** GitHub reports `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` solely on that red required check.

---

*Automated two-engine, multi-lane digest (GitNexus swarm + ce personas + Codex GPT-5.5 xhigh). Findings were cross-checked against the worktree; the headline was reproduced by both engines. Verify before acting — the single recommended change is the item-shape guard at the shape-check line, which closes Finding 2 and its overlapping test gap (F1) together.*

Issue #1811, opened 2026-05-25
Encoding Failure in Windows PowerShell with Chinese Directory Paths


yisen:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

PowerShell Fails to Resolve Directory Paths Containing Chinese Characters Due to Encoding Mismatch

### Expected behavior

running sucessful

### Actual behavior

Analysis failed: COPY failed for File: IO exception: Cannot open file. path: C:\Project\XXXX\code\.gitnexus\lbug - Error 3: The system cannot find the path specified.
Error: COPY failed for File: IO exception: Cannot open file. path: C:\Project\XXXX\code\.gitnexus\lbug - Error 3: The system cannot find the path specified.
    at loadGraphToLbug (file:///C:/Users/yisen/AppData/Roaming/npm/node_modules/gitnexus/dist/core/lbug/lbug-adapter.js:557:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async runFullAnalysis (file:///C:/Users/yisen/AppData/Roaming/npm/node_modules/gitnexus/dist/core/run-analyze.js:415:13)
    at async analyzeCommand (file:///C:/Users/yisen/AppData/Roaming/npm/node_modules/gitnexus/dist/cli/analyze.js:314:24)
    at async Command.<anonymous> (file:///C:/Users/yisen/AppData/Roaming/npm/node_modules/gitnexus/dist/cli/lazy-action.js:16:9)

### Steps to reproduce

npx gitnexus analyze

### Environment

windows 11
powershell

PR #2387, opened 2026-07-06
fix(lbug): recognize Windows missing-shadow error so serve repo-switch recovers (#2382)

### LINKED ISSUE #2382: Error 2: The system cannot find the file specified
### Area

gitnexus (CLI/ core / indexing / MCP server)

### Summary

When opening the visual graph via gitnexus serve, if you start with repository A, switch to repository B, and then switch back to repository A, an "Error 2: The system cannot find the file specified" will be reported. Furthermore, the repository becomes unavailable, requiring you to re-run gitnexus analyze before it can be used normally again.

### Context

_No response_

### Expected behavior

Normal switching and normal usage are possible.

### Actual behavior

Normal switching and usage are not possible.

### Steps to reproduce

Run gitnexus serve
Open http://localhost:4747/
Select repository A
Switch to repository B
Switch back to repository A

### Environment

win11 25h2
NPM 11.16.0
gitnexus 1.6.9

### Logs / screenshots

服务器错误：IO exception: Cannot open file. path: F:\McMod\RTSbuilding-main\.gitnexus\lbug.shadow - Error 2: The system cannot find the file specified.

magyargergo:
Fixes #2382.

## Problem

On Windows, switching repositories in `gitnexus serve` (A → B → A) throws a raw
`IO exception: Cannot open file. path: ...\.gitnexus\lbug.shadow - Error 2: The system cannot find the file specified.`
and leaves repo A unusable until `gitnexus analyze` is re-run. POSIX users never hit it.

## Root cause

The missing-shadow recovery machinery already exists and works — it just never engages on Windows. `isMissingShadowSidecarError` (`sidecar-recovery.ts`) matched only the POSIX error text (`Cannot open file …shadow: No such file or directory`), not the Windows form (`…shadow - Error 2: <localized text>`) — different punctuation *and* a different reason phrase. On the read-only reopen that a repo-switch triggers, the matcher returns `false`, `ensureReadOnlyConnectionUsable` rethrows the raw error, `/api/graph` returns it as a 500, and the orphan WAL is never quarantined — so every subsequent open fails identically.

Windows OS error *text* is localized (the reporter is on non-English Windows), so the fix keys on the **locale-invariant Win32 code** (`Error 2` = `ERROR_FILE_NOT_FOUND`), not the English phrase.

## What changed

- **Matcher** (`sidecar-recovery.ts`): recognize the Windows `Error 2` form, matched **only in the reason after `.shadow`** so a repo *path* containing e.g. `\error 2\` can't trip it. POSIX form unchanged.
  - Excludes **`Error 3`** (`ERROR_PATH_NOT_FOUND`) — that's the [#1811](https://github.com/abhigyanpatwari/GitNexus/pull/1811) non-ASCII path-garble artifact where the shadow is **present on disk**; matching it would quarantine a live WAL (data loss).
  - Excludes `Error 5` / `Error 32` / POSIX `Permission denied` (present-but-locked).
- **Data-loss belt** (`lbug-adapter.ts`): `refuseLargeWalQuarantine` now also refuses to quarantine when the shadow is present on disk (`wal-with-shadow`), guarding both read-only and writable recovery paths.

The matcher is centralized, so the fix covers **all** consumers: serve `/api/graph`, the analyze/writable path, and the MCP/wiki/augmentation pool. A tiny orphan WAL self-heals (quarantine + reopen); a large orphan WAL yields the actionable `analyze --force` message — both replace the raw `Error 2` 500.

## Testing

- `test/unit/sidecar-recovery.test.ts`: new direct coverage for `isMissingShadowSidecarError` (there was none) — POSIX, English-Windows, **localized-Windows**, and negatives (`Error 3`/`5`/`32`, path-embedded number, `Permission denied`, non-shadow file) + predicate-distinctness from `isReadOnlyShadowReplayError`.
- `test/unit/lbug-adapter-wal-schema.test.ts`: Windows-format consumer regression across read-only self-heal, large-WAL actionable message, writable path, and the present-shadow guard.
- `test/unit/pool-wal-recovery.test.ts`: MCP-pool Windows-format recovery.
- `tsc --noEmit` clean; a 21-file / 459-test sweep of the recovery/matcher/pool surface passes with no regressions.

## Caveats

The Windows format is derived from the **issue report string**, not a self-produced live crash; the tests inject that same string, so **green tests do not prove the byte-exact 0.18.0 Windows format**. Please confirm the matcher against a real Windows run (reporter or Windows CI) before closing #2382. The code comment and the plan both flag this.


gitnexus/src/core/lbug/sidecar-recovery.ts
```diff
@@ -109,16 +109,38 @@ const warnOnce = (logger: SidecarRecoveryLogger, key: string, message: string):
...
+//
+// The Windows branch is derived from the issue #2382 reported string, not a
+// self-produced live crash; unit/consumer tests inject that same string, so
+// GREEN TESTS DO NOT PROVE the byte-exact 0.18.0 Windows format — confirm
+// against a real Windows run before closing #2382.
 export const isMissingShadowSidecarError = (err: unknown): boolean => {
   const msg = err instanceof Error ? err.message : String(err);
-  return /Cannot open file .*\.shadow: No such file or directory/i.test(msg);
+  if (!/cannot open file/i.test(msg)) return false;
+  const shadowAt = msg.search(/\.shadow\b/i);
```
magyargergo: **[P2 · cross-engine: Codex (swarm + ce + adversarial) & Claude (security-boundary, ce-adversarial, ce-correctness, ce-reliability) · [reproduced]]** First-`.shadow` anchor false-positive.

`msg.search(/\.shadow\b/i)` returns the **first** `.shadow` in the message. If the dbPath has an earlier `.shadow`-suffixed directory segment (e.g. a `branch=subdir` dir like `snap.shadow\`) **and** a later `\error 2\` path segment, `reason = msg.slice(shadowAt)` starts at the wrong token, so `/\berror\s+2\b/` matches the **directory name** even when the real trailing Win32 code is an EXCLUDED one — defeating the comment on lines 140-141 ("a path-embedded number/phrase before it can't satisfy the discriminator").

Reproduced against this exact matcher body:
- `…F:\snap.shadow\error 2\repo\.gitnexus\lbug.shadow - Error 32:` → `true` (WRONG — real code 32 = sharing violation)
- `…F:\repos\.shadow\error 2\project\.gitnexus\lbug.shadow - Error 5:` → `true` (WRONG — real code 5 = locked)
- shipped KTD2 (`error 2` *before* the sole `.shadow`) → `false` — so the green test only validates the easy arrangement.

On the serve path the new `wal-with-shadow` belt catches this; on the **pool** path (`pool-adapter.ts`, no belt — see the digest's Finding B) it renames a live WAL.

**Fix (one line):** anchor on the LAST `.shadow` — the LadybugDB reason text never contains `.shadow`, so the last match is always the real sidecar:
```ts
const m = [...msg.matchAll(/\.shadow\b/gi)].pop();
if (!m) return false;
const reason = msg.slice(m.index);
```
Add a regression test: earlier `.shadow` segment + a later `error 2` + a real `Error 32`/`Error 5` → expect `false`.

gitnexus/src/core/lbug/lbug-adapter.ts
```diff
@@ -507,22 +507,41 @@ const queryAndDrain = async (targetConn: lbug.Connection, cypher: string): Promi
...
   triggeringErr: unknown,
 ): Promise<void> => {
   const state = await inspectLbugSidecars(dbPath);
+  if (state.kind === 'wal-with-shadow') {
+    logger.warn(
+      `GitNexus: refusing to quarantine WAL at ${dbPath}.wal during ${mode} recovery — ` +
+        'the .shadow sidecar is present on disk, so the open likely failed on path reachability or a lock ' +
+        'rather than a missing shadow. Run `gitnexus analyze --force <repo-path> --index-only` if the index is genuinely broken.',
+    );
+    throw new Error(shadowSidecarRecoveryMessage(dbPath, triggeringErr));
```
magyargergo: **[P3 · Claude ce-correctness + ce-reliability · [code-read]]** Misleading recovery message on the present-shadow branch.

This branch fires only when the shadow is **present** on disk (`wal-with-shadow`), but `shadowSidecarRecoveryMessage` says "LadybugDB checkpoint sidecar is **missing**" and tells the operator to rebuild the index — contradicting the accurate `logger.warn` just above (lines 538-541), which correctly states the shadow is present and the open likely failed on path reachability or a lock. Consider a distinct present-but-unreachable message; the remedy differs (fix the lock/path, not a full reindex). This PR's own KTD7 test pins the misleading wording.

magyargergo (commented):
## Two-engine tri-review — PR #2387 `fix(lbug): recognize Windows missing-shadow error so serve repo-switch recovers (#2382)`

**What this PR does.** Rewrites `isMissingShadowSidecarError` from a POSIX-only regex to a `.shadow`-suffix-anchored discriminator that also recognizes the Windows native form (keying on the locale-invariant Win32 code `Error 2` = FILE_NOT_FOUND, not the localized OS text), and adds a present-shadow (`wal-with-shadow`) refusal belt to the serve-path `refuseLargeWalQuarantine`. Net effect: `gitnexus serve` repo-switch A→B→A on non-English Windows recovers instead of surfacing a raw HTTP 500. +296/−20, 5 files (2 source + 3 unit tests).

**Coordinator verdict: changes requested — fix both headline findings in-PR before merge.** The core approach is correct and a genuine improvement over the raw-500 status quo, and every load-bearing safety property (the Error-3 data-loss guard, the code-not-text exclusions, belt ordering, hygiene) was actively validated — several by reproduction. But the review surfaced **two cross-engine P2 findings on a destructive (`fs.rename`) path**, both with concrete reproduction and both fixable with small in-PR changes: a first-`.shadow` anchor false-positive (Finding A) and the newly-reachable unguarded pool quarantine (Finding B). Given the user's standing "fix-in-PR, don't defer" preference and that the fixes are a one-line anchor change + a pool guard + two regression tests, the honest call is to address both here rather than merge-and-follow-up. **PR state:** open, single non-merge commit atop base (clean feature/fix PR — no unrelated churn, `git diff --check` clean, no bidi); `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` on **checks pending** (quality gates green; test matrix jobs still running; Vercel fails on external auth only).

**Engines & liveness.** Two independent engines, A/B. **Claude** ran the GitNexus swarm (risk-architect, security-boundary, test-ci-verifier) + 6 ce personas (correctness, adversarial, reliability, testing, maintainability, project-standards). **Codex (GPT-5.5/xhigh)** ran three **live** legs — swarm (solo), ce (report-only, spawned correctness/reliability/testing sub-reviewers), and adversarial. All three Codex legs produced real findings. The strong signal below is a Codex lane and a Claude lane reaching the same finding independently. (Codex-swarm's final turn was cut by an unrelated `api_impact` tool-schema error; its coordinator-distilled summary block is used and corroborates the ce leg.)

Verdict spread: Claude risk-architect "production-ready with minor follow-ups"; Codex-swarm/ce "Not ready"; Codex-adversarial "needs-attention". The disagreement is entirely about whether the two headline findings are fix-in-PR blockers or acceptable follow-ups — the mechanism is agreed by everyone.

---

### What's solid (validated by the reviews — credit)

The reviews actively confirmed the load-bearing safety properties, several by reproduction:

- **Error-3 data-loss guard is CORRECT and holds.** risk-architect's mechanism (independently reproduced here): `CreateFileA` garbles the non-ASCII **directory** bytes, never the always-ASCII `lbug.shadow` filename → a garbled present-shadow path surfaces as **Error 3** (PATH_NOT_FOUND), while **Error 2** (FILE_NOT_FOUND) reliably means the ASCII sidecar is genuinely gone. Matching Error 2 only, excluding Error 3, is sound. Reproduced: the Error-3 garble string returns `false`. (risk-architect, security-boundary, ce-correctness, ce-reliability, Codex-swarm; `[reproduced]`)
- **`/\berror\s+2\b/` correctly rejects Error 3 / 5 / 12 / 20 / 22 / 32** and localized trailing text alone (`系统找不到指定的文件。` → false). Reproduced across both engines. (`[reproduced]`)
- **`isReadOnlyShadowReplayError` stayed distinct** — no lexical overlap, KTD5 asserts both directions. Not broadened. (security-boundary, maintainability)
- **Belt ordering correct across all 5 `LbugSidecarState` kinds** — `wal-with-shadow`/`orphan-wal` are mutually exclusive, each throws; clean/tiny/orphan-shadow proceed safely. (risk-architect, ce-correctness)
- **Hygiene clean:** `git diff --check` clean; no bidi/whitespace issues; all added non-ASCII benign (em-dashes, one `⚠️`, and the intentional load-bearing Chinese test string that pins locale-invariant matching). (security-boundary, Codex-swarm)
- **Standards:** conventional-commit title, no CHANGELOG edit, no version bump, no `.skip`/TODO, `LADYBUGDB-CONTRACT` marker retained, scope tight (+296/-20, 5 files, all in blast radius). (project-standards)

---

### Headline findings (both cross-engine, both fix-in-PR; they share one root cause)

Root-cause note up front: the two findings overlap but are **not** both closed by one fix. Finding A's anchor fix (one line, central matcher) closes the **false-positive** chain — a present/locked shadow (Error 3/5/32) misclassified as missing, which the unguarded pool would then quarantine. It does **not** cover Finding B's **independent** exposure: a *genuine* missing shadow with a *large* orphan WAL now reaches the pool's guardless `tryQuarantineForMissingShadow` on Windows (pre-PR the Windows matcher returned `false` → `throw err`, **no filesystem mutation**), where the serve path refuses. So both are needed — fix the anchor (A) **and** add a present-shadow/large-WAL guard to the pool path (B).

#### FINDING A — first-`.shadow` anchor false-positive misclassifies a present/locked shadow as missing  ·  P2  ·  IN-DIFF (inline-able)

- **path:line:** `gitnexus/src/core/lbug/sidecar-recovery.ts:138-143`
- **Engines/lanes:** Codex (swarm + ce P2/conf75 + adversarial `[high]`) **and** Claude (security-boundary P2, ce-adversarial P2/conf72, ce-reliability F3 P2/conf65, ce-correctness F1). Both engines, 7 lanes. `[reproduced]` (here + multiple lanes).
- **Trigger:** `msg.search(/\.shadow\b/i)` anchors on the **first** `.shadow` in the message. A dbPath whose *parent* directory segment ends in `.shadow` (e.g. a branch dir `v2.shadow/`, `snap.shadow\`, `my.shadow-backup\` — dots are legal in git refs, and GitNexus stores branch=subdir) plus a later `\error 2\` path segment makes `reason = msg.slice(shadowAt)` start at the wrong token, so `/\berror\s+2\b/` matches the **directory name** even when the real trailing Win32 code is an EXCLUDED one (3/5/32). This defeats the in-code comment's explicit "a path-embedded number before it can't satisfy the discriminator" claim. Reproduced here against the real matcher body:
  - `…F:\error 2\repo\.gitnexus\lbug.shadow - Error 32:` → **false** (shipped KTD2 — first==last `.shadow`, gives false confidence)
  - `…F:\snap.shadow\error 2\repo\.gitnexus\lbug.shadow - Error 32:` → **true (WRONG — real code is 32, sharing violation)**
  - `…F:\repos\.shadow\error 2\project\.gitnexus\lbug.shadow - Error 5:` → **true (WRONG — real code is 5, locked)**
- **Consequence:** a present-but-locked shadow (Error 5/32) is classified as missing-shadow. On the **serve** path the new `wal-with-shadow` belt catches it (present shadow → throws, no loss). On the **pool** path there is no belt (Finding B) → live WAL renamed aside.
- **Why the green test doesn't catch it:** the shipped KTD2 test only places `error 2` *before* the sole `.shadow` (first==last match), so it validates the easy arrangement and never exercises an earlier `.shadow`-suffixed segment.
- **Fix (one line):** anchor on the LAST `.shadow` — the LadybugDB reason text (`No such file or directory` / `Error N: …`) never contains `.shadow`, so the last match is always the real sidecar: `const m = [...msg.matchAll(/\.shadow\b/gi)].pop(); if (!m) return false; const reason = msg.slice(m.index);`. Add a regression test with an earlier `.shadow` segment + `error 2` + a real Error 32/5.

#### FINDING B — pool quarantine path is unguarded; the broadened matcher newly routes Windows Error 2 into it  ·  P2  ·  BODY-only (pool-adapter.ts unchanged, NOT inline-able)

- **path:line (reference):** `gitnexus/src/core/lbug/pool-adapter.ts:438-463` (`tryQuarantineForMissingShadow`; `fs.rename` via `quarantineWalForMissingShadow` at `:443`), reached from `:486`/`:517`. Contrast `gitnexus/src/core/lbug/lbug-adapter.ts:531-543` (`refuseLargeWalQuarantine`, the new belt).
- **Engines/lanes:** Codex (ce P1/conf100, swarm, adversarial) **and** Claude (security-boundary F2 P2, ce-reliability F4 P1/conf80, ce-adversarial F2 P2/P1, ce-testing F1 P1/conf60; risk-architect F4 rates it **pre-existing-acceptable**). Both engines. `[code-read]` + `[reproduced]` (ce-reliability scratch test).
- **Trigger:** `tryQuarantineForMissingShadow` renames the WAL with **no** present-shadow (`wal-with-shadow`) and **no** large-WAL (`orphan-wal`) refusal — unlike the belt this PR *added* to the serve path. Before this PR the matcher was POSIX-only, so on Windows `isMissingShadowSidecarError` returned false and the pool's probe error fell through to `throw err` (surface error, **no filesystem mutation**). After this PR, Windows Error 2 matches → the pool now calls `fs.rename` on a live WAL. Combined with Finding A, a present-shadow/locked error or a large uncheckpointed WAL is renamed aside on the common Windows MCP/wiki/augmentation read path — the exact #2382 scenario, via the pool instead of serve.
- **Severity settle:** net **P2**. In favor of "acceptable" (risk-architect): the rename is **recoverable** (not delete), POSIX already behaved this way, and without the shadow the uncheckpointed pages are unreplayable in either path (question is preserved-in-place vs moved-aside). Against: it is silent, on the common Windows read path, newly reachable, and defeats the very belt this PR adds elsewhere; four lanes across both engines rate it P1–P2.
- **Fix:** fixing Finding A's anchor closes the *false-positive* half here for all consumers, but does **not** address the *genuine* large-orphan-WAL-on-Windows half — for that the pool needs its own guard. Route `tryQuarantineForMissingShadow` through `refuseLargeWalQuarantine` (or a shared `guardWalQuarantine(dbPath, mode, err)`) before the rename, and add a pool test asserting a present-shadow / large WAL is **refused, not renamed** (no pool equivalent of the serve KTD7 test exists today). So: anchor fix (A) **and** pool guard (B) are both required, not either/or.

---

### Secondary findings (verified, lower severity)

| # | path:line | Sev | In-diff? | Lanes / engines | Note |
|---|-----------|-----|----------|-----------------|------|
| S1 | `gitnexus/test/unit/lbug-adapter-wal-schema.test.ts` (new Windows `describe` ~730-812) + `gitnexus/scripts/cross-platform-tests.ts` PLATFORM_LOGIC | P3 (standards) | test IN-DIFF; target file not | test-ci-verifier + project-standards (Claude, 2 lanes) | **Confirmed:** its two sibling test files (`sidecar-recovery.test.ts:44`, `pool-wal-recovery.test.ts:45`) ARE registered; this one is NOT, so its Windows-format cases run only on ubuntu. Mock-driven (`vi.doMock`, no real FS) → platform-agnostic, so low real impact, but a documented TESTING.md gap. **Fix:** add the one line to PLATFORM_LOGIC. `[code-read]` |
| S2 | `gitnexus/src/core/lbug/lbug-adapter.ts:543` | P3 | IN-DIFF (inline-able) | ce-correctness F2 + ce-reliability F2 (Claude, 2 lanes) | The `wal-with-shadow` branch throws `shadowSidecarRecoveryMessage` ("checkpoint sidecar is **missing**") though it fires only when the shadow is **present** — the adjacent `logger.warn` is accurate, the thrown text contradicts it and points the operator at a full reindex instead of a lock/path-reachability fix. **Fix:** distinct present-but-unreachable message. `[code-read]`, and this PR's own KTD7 test pins the wrong wording. |
| S3 | `gitnexus/src/core/lbug/lbug-adapter.ts:538-542` | P3 | IN-DIFF | ce-reliability F5 (Claude, 1 lane) | Belt's `logger.warn` bypasses the `warnOnce` anti-spam wrapper → one warn per request under a sustained stuck condition. Mirrors the pre-existing `orphan-wal` branch (consistency, not a regression). Minor. `[code-read]` |
| S4 | `gitnexus/src/core/lbug/pool-adapter.ts` `openReadOnlyDatabase` (~511-549) | P1-reported, single-lane | BODY-only | ce-reliability F1 (Claude, 1 lane, `[reproduced]`) | Distinct from Finding B: on a **persistent** (not transient) missing-shadow, the pool's second probe rethrows the **raw** native error unwrapped, so an MCP caller sees the bare string — the #2382 symptom class, relocated. `lbug-adapter.ts` wraps this via `shadowSidecarRecoveryMessage`; the pool does not. Single-lane/single-engine and partly pre-existing, so surfaced as a note, not a corroborated headline. **Fix:** wrap the persistent second probe like `reopenReadOnlyAfterMissingShadow`. |

Also noted (not scored): maintainability F1 — the `refuseLargeWalQuarantine` doc-comment's "returns silently only when the shadow is absent AND the WAL absent/tiny" omits the `orphan-shadow` variant (harmless, doc-accuracy only).

---

### Refuted / dismissed suspicions (validation is a feature)

- **"non-ASCII garble could produce Error 2 not Error 3"** — REFUTED. The filename `lbug.shadow` is always pure ASCII; only the directory garbles → Error 3. (risk-architect, reproduced)
- **"localized OS text could contain a word-bounded `error 2`"** — REFUTED as not-real-on-Windows. FormatMessage strings for codes 1–32767 don't embed "Error 2"; the matcher keys on the numeric code, not English text. (security-boundary, ce-adversarial)
- **`error 2.5` matches `/\berror\s+2\b/`** — true but **zero practical impact**: Win32 GetLastError returns an integer DWORD; "Error 2.5" cannot appear. Optional maximal-precision nit only. (risk-architect P3)
- **Belt-ordering / mutual-exclusion concerns** — REFUTED; `inspectLbugSidecars` returns exactly one kind, ordering is moot.

**Test gaps (all disclosed / non-blocking):**
- **Self-referential validation (self-disclosed in the `LADYBUGDB-CONTRACT` comment):** all Windows-format tests inject the same hand-copied #2382 string, so green ≠ proof of the real `@ladybugdb/core` 0.18.0 Windows byte-format. Everyone agrees it's honestly disclosed and the fix still strictly beats the raw 500. **Real post-merge action:** capture the byte-exact error text from a live Windows missing-shadow run and diff against the fixtures *before closing #2382*.
- No test exercises Finding A's misanchor (earlier `.shadow` segment + later `error 2`), the pool present-shadow/large-WAL refusal (Finding B), or a persistent-failure pool rethrow (S4).

---

### CI status

Quality gates (lint / format / typecheck / typecheck-web) **green**; `npx tsc --noEmit` and the 72 targeted lbug unit tests pass locally (Codex ce reproduced). Several test jobs (ubuntu coverage, windows/macos platform-sensitive, node-floor, packaged-smoke) were **still PENDING** at review time — worth confirming before merge, especially given S1 (the new Windows-format tests aren't in the Windows CI lane) and the disclosed real-Windows validation gap. Vercel fails on an external authorization issue, not a code problem.

---

*This is an automated multi-tool digest (Claude swarm + 6 ce personas, Codex swarm/ce/adversarial). Verify the two headline findings against the diff before acting — the anchor fix (Finding A) is one line and central; the pool belt (Finding B) is also required, guarding the genuine large-orphan-WAL-on-Windows path that Finding A's anchor fix does not cover.*

PR #2393, opened 2026-07-07
fix: resolve imported/composed FastAPI route path constants (#2391)

### LINKED ISSUE #2391: FastAPI route extraction does not resolve imported or composed path constants
### Summary

GitNexus 1.6.9 does not appear to resolve FastAPI route paths when a decorator uses an imported constant composed from other constants.

The handler and execution flows are indexed, but the route may be recorded as `/`. Consequently, `api_impact` cannot find the actual endpoint.

### Minimal reproduction

```python
# constants.py
API_V1 = "/api/v1"
API_V1_WIDGETS = API_V1 + "/widgets"
API_V1_WIDGETS_GET = API_V1_WIDGETS + "/get"
```

```python
# routes.py
from fastapi import APIRouter

from .constants import API_V1_WIDGETS_GET

router = APIRouter()


@router.post(API_V1_WIDGETS_GET)
async def get_widgets():
    return {"success": True}
```

Run:

```bash
npx gitnexus analyze
```

Then query:

```text
api_impact(
  route="/api/v1/widgets/get",
  method="POST"
)
```

### Actual behavior

GitNexus reports:

```json
{
  "error": "No routes found matching \"/api/v1/widgets/get\"."
}
```

`route_map` records the handler under `POST /` rather than `POST /api/v1/widgets/get`.

### Expected behavior

GitNexus should evaluate the imported constant and index:

```text
POST /api/v1/widgets/get
```

`api_impact` should then resolve the endpoint and associate only the appropriate handler and execution flows.

### Confirmed workaround

Replacing the decorator argument temporarily with a string literal works:

```python
@router.post("/api/v1/widgets/get")
```

After re-running `gitnexus analyze`, `api_impact` resolves the route successfully.

### Additional observation

When the literal route is recognized, execution-flow attribution may include flows belonging to other handlers in the same route file. This may indicate that route-to-handler associations are being calculated at file scope rather than handler scope.

### Suggested direction

Consider adding static resolution for:

- Imported string constants
- Constants composed through string concatenation
- Chained constants defined in another module

If complete static evaluation is not possible, reporting the unresolved decorator expression would be preferable to silently indexing it as `/`.

### Environment

- GitNexus: 1.6.9
- Framework: FastAPI
- Language: Python

magyargergo:
Fixes #2391.

## Problem

FastAPI route decorators whose path is a **non-string-literal** — an imported constant (`@router.post(API_V1_WIDGETS_GET)`) or a `+`-concatenation — were indexed with an empty path that normalized to `POST /`. `api_impact` / `route_map` then couldn't resolve the endpoint, and the group HTTP-contract layer dropped the route entirely.

Root cause: two independent literal-only gates (the Python decorator tree-sitter query and the group `FASTAPI_ROUTER_PATTERNS`), plus a fallback that conflated "no argument" with "non-literal argument" — both collapsing to `/`.

## Fix

A shared, transitive, multi-module Python string-constant resolver that both extraction subsystems consume, so provider contracts and graph Route nodes agree on the path.

- **`route-extractors/python-const-resolver.ts`** (new, pure): folds `+`-concatenations of string literals/constants across arbitrarily deep import chains. Keyed by **unique file path** and resolves relative imports package-relative, so two packages' same-named `constants.py` never cross-resolve (a wrong path would be worse than an unresolved one). Cycle-guarded, depth-capped. Reuses the `django.ts` `binary_operator` fold + `MAX_INCLUDE_DEPTH` prior art.
- **Ingestion** (`tree-sitter-queries.ts`, `parse-worker.ts`, `parse-impl.ts`): the query now captures a non-literal first arg; the worker emits the expression + operands and each file's module constants; parse-impl resolves them cross-file before the `include_router`/`APIRouter` prefix pass.
- **Group** (`http-patterns/python.ts`): matches and resolves non-literal decorator args over the **same file closure** ingestion uses (cost-gated — a literal-only repo pays zero extra parse), for provider/graph parity.
- **Correctness floor**: an argument that can't be folded to a literal is **skipped** in both subsystems — no Route node, never `POST /`, never a garbage raw-expression URL.
- **`parse-cache.ts` `SCHEMA_BUMP` 10 → 11**: the new `ParseWorkerResult` fields change the cache shape, so warm caches invalidate — otherwise a same-version re-analyze would silently replay pre-upgrade shards and keep serving `POST /`.

## Out of scope (deferred)

f-strings, `.format()`/`%`, `os.path.join`, `Enum`/`settings.X` attribute references, `__init__.py` package re-exports, and conditional/non-top-level assignments — all fold to the skip floor (never a wrong path).

## Tests

- Pure resolver + tree extractor (literals, concat, `+=`, rebind, multi-hop, aliasing, **package collision**, cycle, depth cap, unknown → null)
- Decorator query capture (literal, empty-literal `""`, identifier, concat, attribute-skip, no-arg)
- `mergeResult` field coverage (regression lock for the worker-accumulator drop)
- End-to-end pipeline: the issue repro, prefix stacking, multi-hop chain, **ingestion↔group parity**, and a **warm parse-cache** round-trip proving the new Map fields survive serialization
- Group provider resolution + cost gate

The original repro now indexes `POST /api/v1/widgets/get`. Full blast-radius suite green, no regressions.

### DIFF
--- a/gitnexus/src/core/group/extractors/http-patterns/python.ts
+++ b/gitnexus/src/core/group/extractors/http-patterns/python.ts
@@ -7,6 +7,13 @@ import {
   type LanguagePatterns,
 } from '../tree-sitter-scanner.js';
 import { normalizeExtractedRoutePath } from '../../../ingestion/route-extractors/route-path.js';
+import {
+  extractPythonModuleConstants,
+  parseConstOperands,
+  resolveOperands,
+  type ModuleConstants,
+  type Operand,
+} from '../../../ingestion/route-extractors/python-const-resolver.js';
 import type { HttpDetection, HttpLanguagePlugin, RepoContext } from './types.js';
 
 /**
@@ -75,6 +82,46 @@ const FASTAPI_ROUTER_PATTERNS = compilePatterns({
               object: (identifier) @obj (#eq? @obj "router")
               attribute: (identifier) @method (#match? @method "^(get|post|put|delete|patch)$"))
             arguments: (argument_list . (string) @path)))
+      `,
+    },
+  ],
+} satisfies LanguagePatterns<Record<string, never>>);
+
+// #2391: `@router.<verb>` / `@app.<verb>` whose first argument is a non-literal
+// path — a bare imported constant or a `+`-concatenation. The path is resolved
+// against the repo-wide constant map (parity with the ingestion side) and, on
+// failure, the route is skipped (no provider contract) exactly like ingestion.
+const FASTAPI_ROUTER_EXPR_PATTERNS = compilePatterns({
+  name: 'python-fastapi-router-expr',
+  language: Python,
+  patterns: [
+    {
+      meta: {},
+      query: `
+        (decorator
+          (call
+            function: (attribute
+              object: (identifier) @obj (#eq? @obj "router")
+              attribute: (identifier) @method (#match? @method "^(get|post|put|delete|patch)$"))
+            arguments: (argument_list . [(identifier) (binary_operator)] @path)))
+      `,
+    },
+  ],
+} satisfies LanguagePatterns<Record<string, never>>);
+
+const FASTAPI_APP_EXPR_PATTERNS = compilePatterns({
+  name: 'python-fastapi-app-expr',
+  language: Python,
+  patterns: [
+    {
+      meta: {},
+      query: `
+        (decorator
+          (call
+            function: (attribute
+              object: (identifier) @obj (#eq? @obj "app")
+              attribute: (identifier) @method (#match? @method "^(get|post|put|delete|patch)$"))
+            arguments: (argument_list . [(identifier) (binary_operator)] @path)))
       `,
     },
   ],
@@ -858,6 +905,13 @@ interface PythonRepoContext {
   prefixesByLongKey: Map<string, Set<string>>;
   /** stem only → set of prefixes (basename fallback, may collide) */
   prefixesByShortKey: Map<string, Set<string>>;
+  /**
+   * File-path-keyed module string constants (#2391), for resolving non-literal
+   * `@router`/`@app` decorator paths. Empty when the repo has no composed-constant
+   * route (cost gate). Keyed identically to the ingestion aggregate so provider
+   * contracts and graph Route nodes resolve the same paths (R4 parity).
+   */
+  constantsByFile: Map<string, ModuleConstants>;
 }
 
 /** Strip `.py` and return the bare basename (e.g. `api/users.py` → `users`). */
@@ -1022,11 +1076,48 @@ function buildPythonRepoContext(
     }
   }
 
+  // #2391: build the repo-wide constant map for resolving non-literal decorator
+  // paths. Cost gate (KTD6): only PARSE for constants when some file actually has
+  // a non-literal `@router`/`@app` decorator (cheap content pre-filter); a
+  // literal-only repo pays just one read pass. When active, parse EVERY `.py`
+  // file so the resolvable set matches the ingestion aggregate (R4 parity) — a
+  // narrower set would return null where ingestion resolves.
+  const constantsByFile = new Map<string, ModuleConstants>();
+  const pyContents = new Map<string, string>();
+  let hasComposedRoute = false;
+  for (const rel of files) {
+    if (!rel.endsWith('.py')) continue;
+    const src = readFile(rel);
+    if (!src) continue;
+    pyContents.set(rel, src);
+    if (!hasComposedRoute && NONLITERAL_ROUTE_DECORATOR_RE.test(src)) hasComposedRoute = true;
+  }
+  if (hasComposedRoute) {
+    for (const [rel, src] of pyContents) {
+      parser.setLanguage(Python);
+      const tree = parseSource(parser, src);
+      if (!tree) continue;
+      const mc = extractPythonModuleConstants(tree);
+      if (mc.literals.size > 0 || mc.exprs.size > 0 || mc.imports.size > 0) {
+        constantsByFile.set(rel, mc);
+      }
+    }
+  }
+
   return {
     prefixesByLongKey,
     prefixesByShortKey,
+    constantsByFile,
   };
 }
+
+// Cheap cost-gate pre-filter: a `@router`/`@app.<verb>(` call whose first
+// non-space argument character starts an identifier (not a quote or `)`), i.e. a
+// bare constant or the head of a `+`-concatenation. Deliberately loose — a false
+// positive only costs a parse pass; a false negative would silently drop the
+// feature, so it errs toward matching.
+const NONLITERAL_ROUTE_DECORATOR_RE =
+  /@\s*(?:app|router)\s*\.\s*(?:get|post|put|delete|patch)\s*\(\s*[A-Za-z_]/;
 
 function joinPrefix(prefix: string, route: string): string {
   // Delegate to the shared route-path normalizer so the group contract and the
@@ -1065,6 +1156,36 @@ export const PYTHON_HTTP_PLUGIN: HttpLanguagePlugin = {
     // is an imported (possibly aliased) symbol resolves to its real definition.
     const importMap = buildPythonImportMap(tree);
 
+    // #2391: fold a non-literal decorator argument (bare constant or
+    // `+`-concatenation) to its literal path against the repo constant map, or
+    // `null` → skip (the same floor the ingestion side applies, so provider
+    // contracts and graph Route nodes agree on both resolved and dropped routes).
+    const resolveExprArg = (argNode: Parser.SyntaxNode): string | null => {
+      const cbf = ctx?.constantsByFile;
+      if (!cbf || !fileRel) return null;
+      const operands: Operand[] | null =
+        argNode.type === 'identifier'
+          ? [{ kind: 'ref', name: argNode.text }]
+          : parseConstOperands(argNode);
+      return operands ? resolveOperands(fileRel, operands, cbf) : null;
+    };
+    const emitAppProvider = (httpMethod: string, pathVal: string, line: number): void => {
+      out.push({
+        role: 'provider',
+        framework: 'fastapi',
+        method: httpMethod,
+        path: pathVal,
+        name: null,
+        // The decorated handler has no captured name → resolve by line-span
+        // containment. Best-effort fallback: FastAPI routes are graph-backed
+        // (ingestion decorator routes) and the function span starts at `def`
+        // (decorators excluded), so this lands the single-decorator case and
+        // degrades to file-level for multi-decorator stacks.
+        line,
+        confidence: 0.8,
+      });
+    };
+
     // Providers: FastAPI @app.<verb>("/path") — already absolute path.
     for (const match of runCompiledPatterns(FASTAPI_APP_PATTERNS, tree)) {
       const methodNode = match.captures.method;
@@ -1074,20 +1195,18 @@ export const PYTHON_HTTP_PLUGIN: HttpLanguagePlugin = {
       if (!httpMethod) continue;
       const path = unquoteLiteral(pathNode.text);
       if (path === null) continue;
-      out.push({
-        role: 'provider',
-        framework: 'fastapi',
-        method: httpMethod,
-        path,
-        name: null,
-        // The decorated handler has no captured name → resolve by line-span
-        // containment. Best-effort fallback: FastAPI routes are graph-backed
-        // (ingestion decorator routes) and the function span starts at `def`
-        // (decorators excluded), so this lands the single-decorator case and
-        // degrades to file-level for multi-decorator stacks.
-        line: pathNode.startPosition.row + 1,
-        confidence: 0.8,
-      });
+      emitAppProvider(httpMethod, path, pathNode.startPosition.row + 1);
+    }
+    // Providers: FastAPI @app.<verb>(CONST | A + "/x") — resolved composed path.
+    for (const match of runCompiledPatterns(FASTAPI_APP_EXPR_PATTERNS, tree)) {
+      const methodNode = match.captures.method;
+      const pathNode = match.captures.path;
+      if (!methodNode || !pathNode) continue;
+      const httpMethod = FASTAPI_VERBS[methodNode.text];
+      if (!httpMethod) continue;
+      const resolved = resolveExprArg(pathNode);
+      if (resolved === null) continue; // skip floor
+      emitAppProvider(httpMethod, resolved, pathNode.startPosition.row + 1);
     }
 
     // Django providers come from the graph Route nodes (includes composed by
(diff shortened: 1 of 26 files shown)


gitnexus/src/core/ingestion/route-extractors/python-const-resolver.ts
```diff
@@ -0,0 +1,226 @@
...
+export function extractPythonModuleConstants(tree: Parser.Tree): ModuleConstants {
+  const literals = new Map<string, string>();
+  const exprs = new Map<string, readonly Operand[]>();
+  const imports = new Map<string, { module: string; originalName: string }>();
+
+  // Apply an assignment result, honoring last-wins: clear any prior binding for
+  // `name`, then set the new one (a `null` rep leaves it cleared = unresolvable).
+  const setName = (name: string, ops: Operand[] | null): void => {
+    literals.delete(name);
+    exprs.delete(name);
```
magyargergo, line 211: **[P2] Import binding not cleared on rebind → confidently wrong route path.** `setName` deletes `literals`+`exprs` but never `imports`, and `handleImport` never clears a prior literal/expr; `foldName` precedence (literals>exprs>imports) then ignores source order. A name that is **both imported and locally (re)assigned** keeps both bindings and the wrong one wins:
- `from .constants import BASE; BASE += "/v1"` → resolves `/api` (stale import), not `/api/v1`.
- `from .constants import ROUTE; ROUTE = os.getenv(...)` → resolves the stale import instead of dropping.

This emits a **non-empty wrong path** — the exact case the skip-floor exists to prevent — and affects **both** ingestion and the group layer (shared `extractPythonModuleConstants`). Fix: make top-level binding writes mutually exclusive across {literals, exprs, imports}, honoring source order (assignment → `imports.delete(name)`; import → clear any prior literal/expr). _[reproduced via tsx against the real resolver]_

gitnexus/src/core/ingestion/route-extractors/constant-resolver.ts
```diff
@@ -0,0 +1,173 @@
...
+function foldName(
+  fileKey: string,
+  name: string,
+  state: ResolveState,
+  depth: number,
+): string | null {
+  if (depth > MAX_RESOLVE_DEPTH) return null;
+  const guard = `${fileKey}::${name}`;
+  if (state.visited.has(guard)) return null; // cycle
+  state.visited.add(guard);
```
magyargergo, line 141: **[P2] `visited` cycle guard is never cleared on unwind → legit repeated/diamond refs silently dropped.** `visited.add(guard)` here is a *whole-fold* set that is never removed, so the 2nd occurrence of any `(fileKey,name)` in one fold is a **false cycle** → `foldName` returns `null` → route dropped. Triggers:
- `X = A + A`
- `ROUTE = SLASH + PATH + SLASH` (reused separator — realistic)
- diamond `X = P + Q` where `P`,`Q` share a common `BASE`

Skip-floor-*safe* (a drop, not a wrong path) but it defeats the feature for these shapes and is a latent trap for the advertised cross-language (JVM) binding that reuses this core. Fix (one line): recursion-stack semantics — `try { … } finally { state.visited.delete(guard); }` around the literal/expr/import resolution branches. _[reproduced via tsx]_

gitnexus/src/core/group/extractors/http-patterns/python.ts
```diff
@@ -1022,12 +1076,49 @@ function buildPythonRepoContext(
...
   };
 }
 
+// Cheap cost-gate pre-filter: a `@router`/`@app.<verb>(` call whose first
+// non-space argument character starts an identifier (not a quote or `)`), i.e. a
+// bare constant or the head of a `+`-concatenation. Deliberately loose — a false
+// positive only costs a parse pass; a false negative would silently drop the
+// feature, so it errs toward matching.
+const NONLITERAL_ROUTE_DECORATOR_RE =
+  /@\s*(?:app|router)\s*\.\s*(?:get|post|put|delete|patch)\s*\(\s*[A-Za-z_]/;
```
magyargergo: **[P2] Cost-gate regex misses a literal-leading concat → R4 (ingestion↔group) parity break.** `NONLITERAL_ROUTE_DECORATOR_RE` requires the first arg char to start an identifier (`[A-Za-z_]`), so `@router.get("/api" + SUFFIX)` (string-literal-leading concat) never trips `hasComposedRoute`. When that is the **only** composed-route shape in a repo, `constantsByFile` stays empty and the group **drops** the route, while ingestion (no such gate) resolves `/api/items` and emits a `Route` node — a graph↔provider divergence that flaps (it's masked by any identifier-first composed route elsewhere in the repo). Fix: widen the gate to also match a string-literal-leading `+`-concat (e.g. allow a leading quote followed by `... + <ident>`). _[regex miss reproduced + downstream drop code-read]_

magyargergo (commented):
## Automated tri-review — PR #2393 (fixes #2391: cross-file FastAPI composed route constants)

**This is an automated multi-tool digest. Verify each finding against the code before acting.** Two review engines were used — **Claude** (swarm: risk-architect, test-ci, security-boundary, docs-dod + ce: correctness, adversarial, performance, api-contract, maintainability, testing, project-standards) and **Codex/GPT-5.5 xhigh** (ce + adversarial completed clean; the Codex *swarm* leg died mid-run on an `api_impact` MCP schema error, so it corroborated facts but produced no standalone structured findings). So: **two engines, Codex partially degraded** — not five independent confirmations. The coordinator additionally **reproduced** the two resolver bugs (findings #1, #2) by executing the real resolver via `tsx`, and the regex miss (#3) via a direct regex test.

### Verdict & PR state
- **Verdict: not production-ready as-is — three small, local P2 fixes recommended first** (finding #1 can emit a *wrong* route path; #2 and #3 are skip-floor-safe under-resolution). Everything else is solid; all three fixes are a few lines each.
- **Branch hygiene:** clean feature PR (linear U1–U6 + refactor stack; no unrelated churn or merge-from-main pollution).
- **Merge state:** `mergeable=MERGEABLE`; CI green except the auth-gated Vercel deploy preview (not a code failure); `mergeStateStatus=UNSTABLE` is Vercel-only.
- **Open questions:** none — every suspicion was resolved by verification (see Refuted section).

### What's solid (lead with the wins)
- **Both prior-review P1s are genuinely closed.** The P0-class `moduleConstants` field-drop from the build is fixed and now **regression-locked** (`result-merge.test.ts` asserts cross-chunk accumulation; the integration warm-cache test round-trips it). The warm parse-cache path (SCHEMA_BUMP 10→11, `mapReplacer`/`mapReviver` nested-Map round-trip) reproduced **GREEN**. Relative-import keying (file-path keys + package-relative dot-walk + ambiguity→null) holds for standard layouts.
- **No breaking contract changes; cache round-trip sound.** The language-agnostic core split (`constant-resolver.ts` + `python-const-resolver.ts` binding) matches the existing `route-extractors/` precedent and is standards-compliant.
- **Strong test suite**: exact-value assertions (not truthiness), R4 ingestion↔group parity on the *same* fixture dir, KTD4 suffix-collision covered twice, multi-hop `deep/` chain. Coordinator ran **51 unit + 10 integration tests GREEN** on the built checkout.
- **CI green except Vercel** (auth-gated deploy preview, not code). `mergeStateStatus=UNSTABLE` is Vercel-only; `mergeable=MERGEABLE`.

---

## Headline findings (3 × P2)

### 1. [P2] Import binding not cleared on rebind → CONFIDENTLY WRONG route path — *violates the PR's core invariant*
`gitnexus/src/core/ingestion/route-extractors/python-const-resolver.ts` — `extractPythonModuleConstants` `setName` (line ~168) deletes `literals`+`exprs` but **never** `imports`; `handleImport` (line ~181) sets `imports` but never clears `literals`/`exprs`; `foldName` precedence is literals>exprs>imports regardless of source order. So a name that is both imported and locally (re)assigned keeps both bindings and the wrong one wins.
- **Trigger A** [reproduced]: `from .constants import ROUTE; ROUTE = os.getenv("ROUTE")` → `@router.get(ROUTE)` resolves to the **stale import value** `/imported` instead of dropping (the live value is dynamic).
- **Trigger B** [reproduced]: `ROUTE = "/local"; from .constants import ROUTE` → resolves `/local`, but Python's last binding is the import.
- This emits a **non-empty wrong path**, the exact failure the skip-floor design exists to prevent. Affects **both** ingestion and group (shared extractor).
- **Agreement**: Codex adversarial B3 `[high]` + Codex swarm B1 (truncated lead, same finding) + Claude ce-correctness F2 + coordinator repro. **[reproduced]**
- **Fix**: make top-level binding writes mutually exclusive across {literals, exprs, imports}, honoring source order (assignment must `imports.delete(name)`; import must clear any prior literal/expr).

### 2. [P2] `visited` cycle guard never cleared → legit repeated-ref / diamond routes silently DROPPED
`gitnexus/src/core/ingestion/route-extractors/constant-resolver.ts` — `foldName` (lines 120–122): `visited` is a *whole-fold* set, added at line 122 but never removed on unwind, so the 2nd occurrence of any name is a false cycle → `null` → route dropped.
- **Triggers** [reproduced]: `X = A + A` → null; `ROUTE = SLASH + PATH + SLASH` (reused separator — realistic) → dropped; diamond `X = P + Q` where `P`,`Q` share `BASE` → null.
- Skip-floor **safe** (a drop, not a wrong path) but defeats the feature and is a trap for the advertised JVM binding that reuses this core.
- **Agreement**: Codex ce B2 (P2, conf 100) + Codex adversarial B3 + Claude ce-correctness F1 + Claude ce-adversarial F2 + coordinator repro. **[reproduced]**
- Note: the Claude **risk-architect** lane called this "correct and tested" — a **false negative** the Codex engine caught; the reproduction overrides it.
- **Fix**: one line — recursion-stack semantics, `try { … } finally { state.visited.delete(guard); }` around the literal/expr/import branches (or push/pop).

### 3. [P2] Group cost-gate regex misses literal-leading concat → R4 parity break
`gitnexus/src/core/group/extractors/http-patterns/python.ts` — `NONLITERAL_ROUTE_DECORATOR_RE` (line 1120): `/@\s*(?:app|router)\s*\.\s*(?:get|post|put|delete|patch)\s*\(\s*[A-Za-z_]/` requires the first arg char to start an **identifier**, so it misses a string-literal-leading concat.
- **Trigger** [reproduced-regex]: `@router.get("/api" + SUFFIX)` — regex sees `("` → no match. When this is the **only** composed-route shape in a repo, `hasComposedRoute` stays false → `constantsByFile` stays empty → the group **drops** the route, while ingestion (no gate) resolves `/api/items` and emits a `Route` node → graph/provider divergence. "Flaps": masked by any identifier-first composed route elsewhere in the repo.
- **Agreement**: Codex adversarial B3 `[medium]` + Claude ce-correctness F3 + Claude ce-adversarial F1 + coordinator regex repro. **[reproduced-regex/code-read]**
- **Fix**: widen the gate to also match a string-literal-leading `+`-concat (e.g. allow `"`/`'` then look for a following `+ident`), or drop the identifier-first constraint.

---

## Inline comments (anchored — post on these exact current lines)
- `gitnexus/src/core/ingestion/route-extractors/python-const-resolver.ts:168` — **[P2]** `setName` clears literals+exprs but never `imports`; a name imported AND locally (re)assigned keeps both → wrong non-empty route path. Add `imports.delete(name)` and make binding writes source-order mutually exclusive.
- `gitnexus/src/core/ingestion/route-extractors/constant-resolver.ts:122` — **[P2]** `visited.add(guard)` is never removed on unwind, so a repeated/diamond ref (`A + A`, shared `BASE`, reused separator) is a false cycle → route silently dropped. Delete the guard on the way out (stack semantics).
- `gitnexus/src/core/group/extractors/http-patterns/python.ts:1120` — **[P2]** the `[A-Za-z_]` gate requires an identifier-first arg, so `@router.get("/api" + SUFFIX)` never trips `hasComposedRoute`; group drops a route ingestion resolves → R4 parity break. Widen to match literal-leading concat.

---

## Lower-priority / quality items (body-only)
- **[P2, Claude-only, multi-lane — non-blocking] Group-layer double read + memory.** `buildPythonRepoContext` (`python.ts:1085–1105`) reads **every** `.py` into `pyContents` (uncached) — an unconditional **2× read** duplicating the existing `include_router` read pass on every Python repo — and, when any composed route exists, re-parses every `.py` with no cross-run cache; `pyContents` holds all file content in memory unconditionally. Real on >2000-file repos. Sources: Claude ce-performance (P2 conf 75) + risk-architect (P2×2) + maintainability + security (P3).
- **[P3] `resolvePythonImport` edge cases.** `from . import X` resolves to `<dir>.py` rather than `<dir>/__init__.py`; over-deep relative imports clamp to a root file instead of returning null. Rare layouts; otherwise a safe drop. Sources: Claude ce-correctness F4 + ce-adversarial F5.
- **[Quality] YAGNI / dead surface.** `resolveConstant` (by-name entry) has **zero production callers** (dead outside tests). The agnostic-core split is a documented reversal — commit `c454e261` deferred it; the next commit did it with only a fabricated Java-style test as the reuse proof — and pushes three files past 1k lines. Source: Claude ce-maintainability.

## Test gaps (credit the otherwise-strong suite)
- **[P1-test] Zero coverage of the group `@app.<verb>(CONST)` EXPR branch** (`FASTAPI_APP_EXPR_PATTERNS`): every group/integration fixture is `@router.*`, never `@app.*`, yet the group layer special-cases the `app` receiver in a distinct compiled query. A regression there ships silently. Source: Claude ce-testing F1.
- **None of the 4 defects above are tested** — every fixture is identifier-first, so the cost-gate miss, the repeated-ref cycle, and the import-rebind wrong-path all slip past.
- **Warm-cache test proves a current-format round-trip, not real stale-shard invalidation** — it never constructs a pre-#2391 (`version: '10+…'`) blob and asserts it's discarded. General version-mismatch path is covered elsewhere, hence P2. Sources: Claude ce-testing F2 + test-ci.

## Docs / DoD nits (P3)
- `parseConstOperands` jsdoc omits `conditional_expression`; `NONLITERAL_ROUTE_DECORATOR_RE` is defined **after** its use; an integration-test comment calls `/v2/api/v1/widgets/get` "ingestion-only garnish" when the group side **does** emit it (`fastapi-composed-provider.test.ts:78–90`). Sources: Claude docs-dod + risk.

---

## Refuted suspicions (validation is a feature — do NOT re-post these)
- **Stack overflow via `parseConstOperands` unbounded recursion** (security lane called it a P2 merge-blocker) → **REFUTED**. Coordinator empirically proved tree-sitter **throws** on a ~5000-deep `+` chain *before* `parseConstOperands` runs; at depths tree-sitter accepts (1000), the function is safe. tree-sitter's own depth cap sits below Node's stack limit → unreachable. A depth cap would be defense-in-depth only. (The ~5000-deep tree-sitter crash is pre-existing, not this PR.) Adversarial lane + coordinator repro agree.
- **Prior-review P1 "relative-import key collapse → wrong file"** → REFUTED (file-path keying + package-relative dot-walk + ambiguity→null holds for standard layouts).
- **Warm-cache loses constants on replay** → REFUTED (SCHEMA_BUMP 10→11; nested-Map round-trip via `mapReplacer`/`mapReviver`; warm-cache integration test ran GREEN).
- **Skip-floor regression (drops a previously-correct route)** → REFUTED (pre-PR a bare-identifier arg produced routePath `''` → `POST /`, itself a wrong route; the skip floor removes only phantom routes).
- **tree-sitter 0.21.1 alternation-predicate hazard** → N/A (new `[...]` alternations are capture-only). **ReDoS on the gate regex, Cypher/SQL injection, path-traversal via `resolvePythonImport`, bidi/whitespace** → all REFUTED (non-ASCII only in JSDoc prose). f-string truncated-path is a **pre-existing** literal fast-path, not introduced here.

---

*Automated tri-review digest (Claude + Codex engines, Codex swarm leg degraded). Findings #1 and #2 are reproduced by executing the real resolver via `tsx`; finding #3 is confirmed by a direct regex test plus a code read of the downstream group path. None of the three block a build, and skip-floor keeps #2 and #3 from ever emitting a wrong path — but finding #1 genuinely can. All three have small, local fixes; verify before acting.*

gitnexus/src/core/ingestion/route-extractors/python-const-resolver.ts
magyargergo: Addressed: `2a7cfbb0` makes the three binding maps mutually exclusive (setName clears imports; bindImport clears prior literals/exprs), and `c08858cf` generalizes it — RHS refs are snapshotted at the assignment line so an aliased-then-rebound constant can no longer publish a stale/wrong path. Locked by unit + E2E tests (app/snapshot.py). Resolving.

gitnexus/src/core/ingestion/route-extractors/constant-resolver.ts
magyargergo: Addressed: `dd50ca23` pops the `visited` guard in `finally` (active-stack semantics) so repeated/diamond refs fold instead of false-cycling; `546223f3` adds a never-popped memo + a MAX_FOLD_LENGTH cap so the popped guard cannot reintroduce exponential recompute / OOM. Real cycles still return null. Locked by unit tests. Resolving.

PR #2369, opened 2026-07-04
feat: full Codex support — hooks, plugin marketplace, and setup (#2328, supersedes #1131)

magyargergo:
## Summary

Completes Codex CLI support by merging the intent of three open workstreams into one verified implementation: **hooks** (#2328), **plugin/marketplace install** (#1131), and the already-shipped MCP + skills setup (#244). Every integration decision was verified against the official Codex docs (developers.openai.com/codex) rather than the assumptions in the original PRs.

> **Stacked on #2368** (CodeBuddy/Qoder) — it reuses that PR's shared `isEnoent` / editor-targets helpers. Merge #2368 first; this diff then reduces to the 3 commits listed below.

### 1. Codex hooks — closes #2328, completes #244 (commit `87501a0c`)

Codex CLI's hooks use **Claude Code's exact schema**: the same `{hooks: {Event: [...]}}` JSON shape, the same stdin payload, and the same `hookSpecificOutput.additionalContext` response contract, registered in a dedicated `~/.codex/hooks.json` ([docs](https://developers.openai.com/codex/hooks)). So:

- `installClaudeCodeHooks` is parameterized into `installClaudeSchemaHooks('claude' | 'codex')` — both runtimes share one installer and the **same bundled `gitnexus-hook.cjs` adapter**, unchanged.
- A `codex` `HookTarget` in `editor-targets.ts` gives uninstall + the setup→uninstall round-trip tripwire full coverage with **zero uninstall.ts changes**.
- `gitnexus setup -c codex` now delivers PreToolUse graph enrichment + PostToolUse stale-index detection, matching Claude Code.
- **SessionStart deliberately not registered**: Codex reads `AGENTS.md` natively ([docs](https://developers.openai.com/codex/guides/agents-md)), which already carries the GitNexus context block — an echo hook would duplicate it.

### 2. Codex plugin install — supersedes #1131 (commit `d0df2bb3`, co-authored with @jublin)

#1131's approach was right, but its implementation duplicated 7 (now-stale) skills into a parallel `gitnexus-codex-plugin/` tree. Verified against the [plugin docs](https://developers.openai.com/codex/plugins/build): Codex sets `CLAUDE_PLUGIN_ROOT`/`CLAUDE_PLUGIN_DATA` for compatibility and loads the same SKILL.md skills, `hooks/hooks.json`, and `.mcp.json` — so the existing `gitnexus-claude-plugin/` is **already Codex-compatible as-is**. This PR adds only:

[...]

Issue #2163, opened 2026-06-11
bug(hooks): Claude db-lock probe leaks orphaned lsof processes on busy Linux hosts — snowballs to sustained 100% CPU


Minidoracat:
## Area

Claude Code hooks (`gitnexus/hooks/claude/hook-db-lock-probe.cjs`)

## Summary

On Linux hosts with a high process count, the PreToolUse db-lock probe leaks **orphaned `lsof` processes** (PPID 1, stuck in R state). Each leaked `lsof` adds CPU load, which makes the next probe slower and more likely to leak again — a self-reinforcing snowball. On a 16-core server this accumulated **114 orphaned `lsof` processes and a load average of 122**, pinning all cores at 100% for ~40 minutes until manually cleaned up.

The probe was introduced in #1493 (to fix #1492), so every install since v1.6.x is exposed. This is distinct from #1546 (orphaned `gitnexus mcp` processes), though both were present on the affected host.

## Environment

- gitnexus **1.6.7** (hooks installed by `gitnexus analyze` / `setup`, registered with a 10s timeout on `PreToolUse: Grep|Glob|Bash` and `PostToolUse: Bash`)
- Node v24.15.0, Ubuntu (Proxmox VM), 16 cores, 64 GB RAM
- ~700 processes / ~580 tasks at steady state (busy dev server: MySQL, Redis, nginx, Docker, several editor/agent sessions)
- 2 concurrent Claude Code sessions (with subagents → frequent Grep/Glob/Bash tool calls) open in a repo indexed by GitNexus

## What I observed

```
top - load average: 122.63, 124.17, 127.1   (16 cores)
Tasks: 707 total, 120 running
%Cpu(s): 57.8 us, 41.3 sy, 0.0 id

  PID    USER  ... %CPU  TIME+   COMMAND
2872721  root  ... 100.0 10:07   lsof
2873936  root  ... 100.0  9:09   lsof
...  (114 total, all:)
/usr/bin/lsof -nP -t -- /path/to/repo/.gitnexus/lbug
```

- All 114 `lsof` had **PPID 1** (orphaned) and were in **R state**, each having burned 4–10 minutes of CPU time.
- A **new `lsof` appeared roughly every ~30s** (matching the tool-call cadence of the active agent sessions), replacing any that finished — a stable pile-up.
[...]

Issue #1913, opened 2026-05-30
Make GitNexus hook silent on MCP-owned DB skip for strict Codex hook runners


lishixiang0705:
## Summary

When the GitNexus Claude hook is wired into Codex as a `PreToolUse` hook, Codex validates hook output against a strict JSON schema. The normal GitNexus skip path for an MCP-owned DB can emit a non-JSON diagnostic:

```text
[GitNexus] augment skipped: MCP server owns DB
```

That makes Codex hook runs noisy and can surface as `hook returned invalid pre-tool-use JSON output` in strict hook environments. The hook should be silent on normal skip paths, or only emit diagnostics behind `GITNEXUS_DEBUG`.

## Environment

- GitNexus: 1.6.5
- Codex CLI: 0.135.0
- Hook file: `hooks/claude/gitnexus-hook.cjs`
- Usage: GitNexus hook registered from `~/.codex/hooks.json` for `PreToolUse` / `PostToolUse`

## Reproduction shape

With GitNexus MCP server owning the DB, invoke the hook as Codex would:

```bash
printf '%s' '{"hook_event_name":"PreToolUse","tool_name":"Bash","tool_input":{"command":"rg someSymbol ."},"cwd":"/path/to/indexed/repo"}' \
  | node ~/.claude/hooks/gitnexus/gitnexus-hook.cjs
```

Current normal skip behavior can write the diagnostic above to stderr.

## Expected behavior

For normal non-error skip paths, hook output should be either:

- empty stdout/stderr, exit 0, or
- valid hook JSON matching the target runner schema.

Diagnostics such as `MCP server owns DB` should be debug-only, e.g. behind `GITNEXUS_DEBUG=1`.

## Suggested fix

In `handlePreToolUse`, change the MCP-owned-DB skip path from unconditional stderr output to debug-only output:

```js
if (hasGitNexusServerOwner(gitNexusDir)) {
  if (process.env.GITNEXUS_DEBUG === '1' || process.env.GITNEXUS_DEBUG === 'true') {
    process.stderr.write('[GitNexus] augment skipped: MCP server owns DB\\n');
  }
  return;
}
```

A stronger follow-up would be a small Codex-compatible wrapper/self-test so GitNexus upgrades can verify that `PreToolUse` and `PostToolUse` never emit invalid hook output for Codex.

Issue #2180, opened 2026-06-12
perf(hooks): replace the O(procs×fds) Linux db-lock scan with a cmdline-first sweep and drop the lsof fallback (proposal)


Minidoracat:
## Area

Claude Code hooks (`gitnexus/hooks/claude/hook-db-lock-probe.cjs`) — Linux db-lock probe

## Context

#2163 → #2165 → #2169 contained the orphan storm: probe and augment subprocesses are now guarded, concurrency is slot-bounded, and a busy host can no longer snowball to 100% CPU. What those PRs deliberately did **not** change is the probe's detection algorithm — and on busy hosts it still cannot answer the question it exists for:

- `linuxProcScanFindGitNexusServer` walks `/proc/<pid>/fd/*` with `stat(2)` for **every process** — O(processes × fds). Measured on the incident-class host (~660 processes / ~18,300 fds): **~280 ms** for a full scan on an idle box, and a guaranteed `GITNEXUS_HOOK_LINUX_PROC_BUDGET_MS` (1200 ms) blowout under load.
- The `lsof` fallback then burns its full 1 s `spawnSync` budget (real `lsof` takes ~2 s there) and lands ETIMEDOUT → fail-closed.

Net effect on busy hosts: every Grep/Glob/Bash hook invocation spends ~2.2 s of CPU to conclude "couldn't tell" and skip the augment. The guards from #2165/#2169 make this safe; this proposal makes it *fast and correct*.

## Proposal: invert the scan — identify candidates by name first, stat only their fds

Rewrite `linuxProcScanFindGitNexusServer` (name kept) as a tri-state (`'owned' | 'not-owned' | 'timeout'`) three-phase scan:

1. **Phase 0 — `/proc/<pid>/comm`**: one tiny read per process (kernel serves it from `task->comm`, never touches the target's memory maps; microseconds each). Keep only processes whose comm could plausibly be a Node/gitnexus runner (`node`, `gitnexus`, `bun`, `npm`, …).
2. **Phase 1 — `/proc/<pid>/cmdline`** for survivors only, read via `openSync`+`readSync` capped at 4 KB (guards against D-state processes wedged on `mmap_lock` and against giant cmdlines). Prefilter with the existing `isGitNexusServerCommand`.
3. **Phase 2 — fd stat** for the 0–2 remaining candidates only: compare `dev`+`ino` against the lbug file, exactly as today.

[...]

PR #2397, opened 2026-07-08
fix(hook): emit MCP query hint when server owns DB lock (#2396)

### LINKED ISSUE #2396: PreToolUse hook silently skips when MCP server owns DB lock — defeats the purpose
## Problem

The PreToolUse hook's `augment` path silently skips when the GitNexus MCP server is running, which is exactly the scenario where augmentation is most useful.

### Current logic (hook source: `gitnexus-hook.cjs` L387-395)

```js
if (hasGitNexusServerOwner(gitNexusDir)) {
    // MCP server holds the lock → skip
    return;
}
```

The hook spawns `gitnexus augment -- <pattern>` as a CLI process that needs to **exclusively open** the KuzuDB file. But the MCP server already holds the DB lock, so the hook detects the conflict and silently returns — no augmentation, no feedback to the agent.

### Why this is backwards

| Scenario | Hook behavior | Actual need |
|---|---|---|
| MCP server **online** (most common) | **Skipped** — DB locked | Should work — this is when agents benefit most from graph context |
| MCP server **offline** | Runs CLI augment | Lower value — agent has no MCP tools to complement the augmentation |

The hook's primary value-add — injecting graph context before searches — is disabled in the exact deployment mode most users will use (MCP server running as a long-lived process).

### Proposed fix

When the MCP server owns the DB lock, the hook should **query the MCP server** (via its stdio/HTTP interface) instead of trying to run a competing CLI process. Fallback to the CLI path only when the MCP server is not running.

Alternatively, a simpler approach: when the MCP server is detected, emit a lightweight `additionalContext` hint like `"GitNexus graph context available via MCP query tool"` so the agent can decide whether to call `mcp__gitnexus__query` itself.

### Environment

- macOS, Node v22
- GitNexus installed globally via npm
- Claude Code with MCP server configured
- Hook configured in `~/.claude/settings.json` PreToolUse for `Grep|Glob|Bash`

magyargergo:
## Problem

Fixes #2396. The PreToolUse hook's CLI `augment` path silently skips **exactly when the GitNexus MCP server is running** — the most common deployment. The `augment` child needs to open the `lbug` KuzuDB/LadybugDB file, but LadybugDB is single-writer and the server already holds the write lock, so the hook detects the owner and `return`s with no augmentation and no feedback. Reproduced on `main` (linux/x64): with a `gitnexus … mcp` process holding the DB fd, the hook produced zero output (only visible under `GITNEXUS_DEBUG=1`: `[GitNexus] augment skipped: MCP server owns DB`).

## Why not "route the hook to the server"

The issue floats having the hook query the running server. Rejected as disproportionate:

- The MCP server runs over **stdio**, spawned by the client — the hook is a *separate* process with no port or pipe to it.
- LadybugDB is **single-writer**; the server holds the DB in write mode, so a second-process read-only `augment` can't coexist with the lock either.

Both would be large, fragile architecture changes.

## Fix

The decisive fact: when the server owns the lock, **the same session that fired the hook already has the GitNexus MCP `query` tool live**. So instead of fighting for the DB, hand the augmentation to the agent.

The owner branch now emits an `additionalContext` hint pointing the agent at `mcp__gitnexus__query` for that pattern, via the **same `hookSpecificOutput.additionalContext` stdout channel the successful augment already uses** (so it's strict-runner-safe — Codex accepts the same schema, #2369):

```
[GitNexus] Knowledge graph is live via the MCP server. For graph-ranked context
on "<pattern>", call the GitNexus `query` MCP tool (e.g. mcp__gitnexus__query)
with search_query "<pattern>".
```

- `return;` → `result = buildMcpQueryHint(pattern)` (a one-line helper); the existing emit path sends it.
- Applied to all **three gated hook copies**: `hooks/claude/gitnexus-hook.cjs`, `gitnexus-claude-plugin/hooks/gitnexus-hook.js`, `hooks/antigravity/gitnexus-antigravity-hook.cjs` (matching the repo's existing per-file duplication — no new shared module).
- The cursor hook has no owner gate (augment fails gracefully; probe deferred per #2163) and is left untouched.
- The stderr `augment skipped: MCP server owns DB` diagnostic is preserved and stays `GITNEXUS_DEBUG`-gated (#1913).

## Tests

- Owner-path assertions in `test/unit/hooks.test.ts` and `test/integration/antigravity-hook-e2e.test.ts` flipped from *stdout-empty* to *hint-present* (pattern + `mcp__gitnexus__query`), keeping the gated-stderr contract and the "CLI never ran" marker check.
- `vitest run test/unit/hooks.test.ts` → **165 pass / 36 skip**.
- The owner-path unit assertions are `skipIf(linux)` (lsof lane = macOS CI); the Linux emission was validated end-to-end against a real `/proc` + a fake `…/node_modules/gitnexus/dist/cli/index.js mcp` process holding the `lbug` fd.

### DIFF
--- a/gitnexus/hooks/claude/gitnexus-hook.cjs
+++ b/gitnexus/hooks/claude/gitnexus-hook.cjs
@@ -353,15 +353,19 @@ function runGitNexusCli(cliPath, args, cwd, timeout) {
  * PreToolUse handler — augment searches with graph context.
  */
 /**
- * Fallback augmentation for the #2396 path: when the MCP server owns the DB the
- * CLI `augment` can't run, so hand the agent a pointer to the live MCP `query`
- * tool for this pattern. `pattern` is JSON-escaped by sendHookResponse.
+ * Fallback augmentation for the #2396 path: when a GitNexus process holds the
+ * lbug DB write lock the CLI `augment` can't run, so point the agent at the MCP
+ * `query` tool instead. Phrased conditionally ("if the MCP tools are live") so it
+ * stays truthful on every owner path — a confirmed MCP owner, a `serve` owner, or
+ * a fail-closed probe where no server is actually confirmed. `pattern` is embedded
+ * verbatim; the caller (sendHookResponse) JSON-escapes it structurally.
  */
 function buildMcpQueryHint(pattern) {
   return (
-    `[GitNexus] Knowledge graph is live via the MCP server. For graph-ranked ` +
-    `context on "${pattern}", call the GitNexus \`query\` MCP tool ` +
-    `(e.g. mcp__gitnexus__query) with search_query "${pattern}".`
+    `[GitNexus] Local augment is unavailable (the graph DB is held by another ` +
+    `GitNexus process). If the GitNexus MCP tools are live in this session, call ` +
+    `the GitNexus \`query\` MCP tool (e.g. mcp__gitnexus__query) with ` +
+    `search_query "${pattern}".`
   );
 }
 
--- a/gitnexus/hooks/claude/gitnexus-hook.cjs
+++ b/gitnexus/hooks/claude/gitnexus-hook.cjs
@@ -352,6 +352,19 @@ function runGitNexusCli(cliPath, args, cwd, timeout) {
 /**
  * PreToolUse handler — augment searches with graph context.
  */
+/**
+ * Fallback augmentation for the #2396 path: when the MCP server owns the DB the
+ * CLI `augment` can't run, so hand the agent a pointer to the live MCP `query`
+ * tool for this pattern. `pattern` is JSON-escaped by sendHookResponse.
+ */
+function buildMcpQueryHint(pattern) {
+  return (
+    `[GitNexus] Knowledge graph is live via the MCP server. For graph-ranked ` +
+    `context on "${pattern}", call the GitNexus \`query\` MCP tool ` +
+    `(e.g. mcp__gitnexus__query) with search_query "${pattern}".`
+  );
+}
+
 function handlePreToolUse(input) {
   const cwd = input.cwd || process.cwd();
   if (!path.isAbsolute(cwd)) return;
@@ -385,18 +398,23 @@ function handlePreToolUse(input) {
   let result = '';
   try {
     if (hasGitNexusServerOwner(gitNexusDir)) {
-      // Normal skip path: the MCP server owns the DB, so the CLI augment would
-      // contend on the lock. Stay silent for strict hook runners (issue #1913);
-      // surface the reason only when diagnostics are explicitly requested.
+      // #2396: the MCP server holds the DB write lock, so a competing CLI
+      // `augment` would only contend on it (LadybugDB is single-writer). But the
+      // session that triggered this hook has the GitNexus MCP tools live — route
+      // the augmentation to the agent via additionalContext instead of silently
+      // doing nothing. Mirror the skip reason to stderr only under GITNEXUS_DEBUG
+      // (strict-runner contract, #1913); the hint itself rides the sanctioned
+      // additionalContext stdout channel the successful augment already uses.
       if (isDebugEnabled()) {
         process.stderr.write('[GitNexus] augment skipped: MCP server owns DB\n');
       }
-      return;
-    }
-    const cliPath = resolveCliPath();
-    const child = runGitNexusCli(cliPath, ['augment', '--', pattern], cwd, 7000);
-    if (!child.error && child.status === 0) {
-      result = extractAugmentContext(child.stderr || '');
+      result = buildMcpQueryHint(pattern);
+    } else {
+      const cliPath = resolveCliPath();
+      const child = runGitNexusCli(cliPath, ['augment', '--', pattern], cwd, 7000);
+      if (!child.error && child.status === 0) {
+        result = extractAugmentContext(child.stderr || '');
+      }
     }
   } catch {
     /* graceful failure */
--- a/gitnexus/hooks/claude/gitnexus-hook.cjs
+++ b/gitnexus/hooks/claude/gitnexus-hook.cjs
@@ -367,6 +367,32 @@ function buildMcpQueryHint(pattern) {
 }
 
 /**
+ * #2396 throttle: emit the MCP-query hint at most once per repo per window, so an
+ * owner-locked session isn't nudged on every search. Window (ms) via
+ * GITNEXUS_MCP_HINT_THROTTLE_MS (default 10min; 0/invalid disables). Best-effort —
+ * any fs error falls back to emitting.
+ * ponytail: per-repo mtime marker, shared across concurrent sessions on the same
+ * repo; add per-session dedup only if that sharing becomes a problem.
+ */
+function shouldEmitMcpHint(gitNexusDir) {
+  const raw = process.env.GITNEXUS_MCP_HINT_THROTTLE_MS;
+  const windowMs = raw === undefined || raw === '' ? 600000 : Number(raw);
+  if (!Number.isFinite(windowMs) || windowMs <= 0) return true;
+  const marker = path.join(gitNexusDir, '.mcp-hint-shown');
+  try {
+    if (Date.now() - fs.statSync(marker).mtimeMs < windowMs) return false;
+  } catch {
+    /* marker missing/unreadable → emit */
+  }
+  try {
+    fs.writeFileSync(marker, '');
+  } catch {
+    /* best-effort; still emit */
+  }
+  return true;
+}
+
+/**
  * PreToolUse handler — augment searches with graph context.
  */
 function handlePreToolUse(input) {
@@ -412,7 +438,9 @@ function handlePreToolUse(input) {
       if (isDebugEnabled()) {
         process.stderr.write('[GitNexus] augment skipped: MCP server owns DB\n');
       }
-      result = buildMcpQueryHint(pattern);
+      if (shouldEmitMcpHint(gitNexusDir)) {
+        result = buildMcpQueryHint(pattern);
+      }
     } else {
       const cliPath = resolveCliPath();
       const child = runGitNexusCli(cliPath, ['augment', '--', pattern], cwd, 7000);
--- a/gitnexus/hooks/claude/gitnexus-hook.cjs
+++ b/gitnexus/hooks/claude/gitnexus-hook.cjs
@@ -350,9 +350,6 @@ function runGitNexusCli(cliPath, args, cwd, timeout) {
 }
 
 /**
- * PreToolUse handler — augment searches with graph context.
- */
-/**
  * Fallback augmentation for the #2396 path: when a GitNexus process holds the
  * lbug DB write lock the CLI `augment` can't run, so point the agent at the MCP
  * `query` tool instead. Phrased conditionally ("if the MCP tools are live") so it
@@ -369,6 +366,9 @@ function buildMcpQueryHint(pattern) {
   );
 }
 
+/**
+ * PreToolUse handler — augment searches with graph context.
+ */
 function handlePreToolUse(input) {
   const cwd = input.cwd || process.cwd();
   if (!path.isAbsolute(cwd)) return;
--- a/gitnexus/test/unit/hooks.test.ts
+++ b/gitnexus/test/unit/hooks.test.ts
@@ -2012,6 +2012,87 @@ describe.skipIf(SKIP_LSOF_PATH)('MCP-owner hint throttle (#2396)', () => {
   }
 });
 
+// #2396: buildMcpQueryHint and its throttle are triplicated across the three hook
+// copies (the repo's deliberate no-shared-module hook convention). Guard against
+// silent drift with a source-level byte-identity check — runs on every platform,
+// unlike the owner-path behavior tests which are macOS-only.
+describe('hook copy drift guard (#2396)', () => {
+  const ANTIGRAVITY_HOOK = path.resolve(
+    __dirname,
+    '..',
+    '..',
+    'hooks',
+    'antigravity',
+    'gitnexus-antigravity-hook.cjs',
+  );
+  const HOOK_SOURCES: ReadonlyArray<readonly [string, string]> = [
+    ['claude', CJS_HOOK],
+    ['plugin', PLUGIN_HOOK],
+    ['antigravity', ANTIGRAVITY_HOOK],
+  ];
+
+  function extractFn(source: string, name: string): string {
+    const match = source.match(new RegExp(`function ${name}\\([^)]*\\) \\{[\\s\\S]*?\\n\\}`));
+    return match ? match[0] : `<${name} not found>`;
+  }
+
+  for (const fnName of ['buildMcpQueryHint', 'shouldEmitMcpHint']) {
+    it(`${fnName} is byte-identical across all three hook copies`, () => {
+      const [claude, plugin, antigravity] = HOOK_SOURCES.map(([, p]) =>
+        extractFn(fs.readFileSync(p, 'utf-8'), fnName),
+      );
+      expect(claude).toContain(`function ${fnName}`);
+      expect(plugin).toBe(claude);
+      expect(antigravity).toBe(claude);
+    });
+  }
+});
+
+// #2396: an adversarial search pattern (embedded quote + newline) must not break
+// the additionalContext JSON envelope — JSON.stringify in the emit path escapes it
+// structurally. Owner-path only (macOS/other-Unix lsof+ps lane, SKIP_LSOF_PATH).
+describe.skipIf(SKIP_LSOF_PATH)('MCP hint pattern escaping (#2396)', () => {
+  for (const [label, hookPath] of [
+    ['CJS', CJS_HOOK],
+    ['Plugin', PLUGIN_HOOK],
+  ] as const) {
+    it(`${label}: quote+newline pattern stays JSON-safe in additionalContext`, () => {
+      const markerPath = path.join(os.tmpdir(), `gn-hook-esc-${process.pid}-${label}`);
+      const lbugPath = path.join(gitNexusDir, 'lbug');
+      fs.writeFileSync(lbugPath, '');
+      fs.rmSync(markerPath, { force: true });
+      const binDir = createHookToolDir({
+        gitnexusMarkerPath: markerPath,
+        lsofOutput: '12345\n',
+        psOutput: 'node /tmp/node_modules/.bin/gitnexus mcp\n',
+      });
+      const evilPattern = 'foo"bar\nbaz';
+      try {
+        const result = runHook(
+          hookPath,
+          {
+            hook_event_name: 'PreToolUse',
+            tool_name: 'Grep',
+            tool_input: { pattern: evilPattern },
+            cwd: tmpDir,
+          },
+          undefined,
+          { env: hookEnv(binDir) },
+        );
+        // parseHookOutput JSON.parses stdout — a broken envelope would throw/return null.
+        const output = parseHookOutput(result.stdout);
+        expect(output!.additionalContext).toContain('foo"bar');
+        expect(output!.additionalContext).toContain('search_query');
+        expect(result.status).toBe(0);
+      } finally {
+        fs.rmSync(lbugPath, { force: true });
+        fs.rmSync(markerPath, { force: true });
+        fs.rmSync(binDir, { recursive: true, force: true });
+      }
+    });
+  }
+});
+
 describe.skipIf(SKIP_LSOF_PATH)(
   'Ladybug DB owner guard — production-shaped ps + failure modes (#1493)',
   () => {
--- a/gitnexus/test/unit/hooks.test.ts
+++ b/gitnexus/test/unit/hooks.test.ts
@@ -1817,13 +1817,14 @@ describe('PreToolUse augmentation filtering (integration)', () => {
       }
     });
 
-    // Issue #1913: the MCP-owned-DB skip is a NORMAL (non-error) path, so by
-    // default it must stay completely silent — empty stdout AND empty stderr,
-    // exit 0 — so strict hook runners (e.g. Codex `PreToolUse`) never see
-    // unexpected output. GITNEXUS_DEBUG is forced off to keep the assertion
-    // deterministic regardless of the ambient environment.
+    // #2396: when a GitNexus MCP process owns the repo DB the CLI augment can't
+    // run, so the hook hands the agent the MCP-query hint on stdout (the sanctioned
+    // additionalContext channel). By default (GITNEXUS_DEBUG unset) the stderr skip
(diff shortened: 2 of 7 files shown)


gitnexus/test/unit/hooks.test.ts
```diff
@@ -1859,11 +1859,13 @@ describe('PreToolUse augmentation filtering (integration)', () => {
...
-    // via GITNEXUS_DEBUG=1 — stdout stays empty (no augment ran), the diagnostic
-    // appears on stderr.
+    // #2396: when the MCP server owns the DB the CLI augment can't run, so the
+    // hook hands the agent an MCP-query hint on stdout (the sanctioned
+    // additionalContext channel) instead of doing nothing. The CLI still never
+    // spawns (marker absent). #1913: the stderr skip diagnostic stays gated
+    // behind GITNEXUS_DEBUG.
     it.skipIf(SKIP_LSOF_PATH)(
-      `${label}: surfaces the MCP-owner skip reason only under GITNEXUS_DEBUG`,
+      `${label}: MCP-owner path emits the MCP query hint; stderr reason gated by GITNEXUS_DEBUG`,
```
magyargergo, line 1871: **P0 · Blocks merge: yes · [reproduced] (CI-confirmed, both engines)**

This is the one owner-path test the PR updated — but the same fix broke a whole **class** of sibling owner-path tests that were left asserting the old silent behaviour, and the macOS `platform-sensitive 2/3` shard is red because of it (run 28956185902, job 85915692366).

The owner branch now emits the hint on stdout, and the probe collapses *timeout/guard-death* and *confirmed-owned* into the same `true`, so **both** the definite-owner tests **and** every fail-closed test now receive the hint JSON where they still `expect(result.stdout.trim()).toBe('')`. CI-confirmed failing assertions (all in this file): **1850, 1998, 2062** (definite owner) and **2150, 2186, 2234, 2287, 2343, 2399** (fail-closed) — each parametrized over `['CJS','Plugin']`. `ce-testing` also predicts **2524** (multi-PID) on the same branch.

These only run on macOS (`SKIP_LSOF_PATH = win32 || linux`, L100), which is why Linux/Windows are green.

**Fix (recommended — Option B, see body §2):** narrow the code to emit the hint only on *confirmed* ownership; flip just the 3 definite-owner tests (1850/1998/2062) to parse stdout and assert `output!.additionalContext` contains `mcp__gitnexus__query`; keep the fail-closed tests asserting silence. Also add back a `GITNEXUS_DEBUG=''` (unset) owner-hint case — the new loop only covers `['0','false']`. (Option A = accept fail-closed→emit and flip all ~10 assertions.)

gitnexus/hooks/claude/gitnexus-hook.cjs
```diff
@@ -385,18 +398,23 @@ function handlePreToolUse(input) {
...
       if (isDebugEnabled()) {
         process.stderr.write('[GitNexus] augment skipped: MCP server owns DB\n');
       }
-      return;
-    }
-    const cliPath = resolveCliPath();
-    const child = runGitNexusCli(cliPath, ['augment', '--', pattern], cwd, 7000);
-    if (!child.error && child.status === 0) {
-      result = extractAugmentContext(child.stderr || '');
+      result = buildMcpQueryHint(pattern);
```
magyargergo: **P2 · Blocks merge: maybe · [code-read]** (Codex ce, 5 reviewers; ce-adversarial partial)

`hasGitNexusServerOwner()` returns `true` not only for a confirmed `gitnexus mcp` owner but also on **fail-closed** paths — lsof/ps `ETIMEDOUT`, guard exit 124/137, guard signal-death (`hook-db-lock-probe.cjs` L629/642/662/668). So on a probe timeout with **no server actually running**, this line still emits *"Knowledge graph is live via the MCP server … call `mcp__gitnexus__query`"* — an unverified claim that points the agent at a tool that may not exist.

This is the code side of the §2 design decision. **Recommended:** expose the probe's tri-state verdict and emit the hint only for *confirmed* MCP ownership; keep fail-closed silent (debug-gated stderr only), in all three copies (`:411`, plugin `:443`, antigravity `:435`). This also resolves the C1 test breakage on the fail-closed lines and the C4 `serve`-owner concern.

gitnexus/hooks/claude/gitnexus-hook.cjs
```diff
@@ -352,6 +352,19 @@ function runGitNexusCli(cliPath, args, cwd, timeout) {
 /**
  * PreToolUse handler — augment searches with graph context.
  */
+/**
+ * Fallback augmentation for the #2396 path: when the MCP server owns the DB the
+ * CLI `augment` can't run, so hand the agent a pointer to the live MCP `query`
+ * tool for this pattern. `pattern` is JSON-escaped by sendHookResponse.
+ */
+function buildMcpQueryHint(pattern) {
```
magyargergo, line 360: **P3 · Blocks merge: no · [code-read]** (cross-engine: ce-maintainability, ce-correctness, codex-swarm, codex-ce)

Inserting `buildMcpQueryHint` here orphaned the pre-existing `/** PreToolUse handler — augment searches with graph context. */` JSDoc (L353) onto this helper, leaving `handlePreToolUse` (L368) undocumented. The **plugin copy** placed the helper cleanly (after `sendHookResponse`, before the handler doc) and antigravity has no orphan — so this is a copy-specific inconsistency.

**Fix:** move `buildMcpQueryHint` + its own doc block above the "PreToolUse handler" comment, matching `gitnexus-claude-plugin/hooks/gitnexus-hook.js`.

magyargergo (commented):
## Tri-review — PR #2397 (fix #2396: emit MCP-query hint when the MCP server owns the lbug DB lock)

Base (merge-base) `840296319818177b3195c5b13d92c8f57224b0ee` → HEAD `7166137b7`. Files: 3 hook copies + 2 test files (+113 / −48).

### Final verdict: **not production-ready**

The change is directionally correct and the code's runtime behavior is contract-safe, but the PR's own test suite contradicts the new behavior: the macOS `tests / macos-latest (platform-sensitive) 2/3` shard **fails** with 9 assertion errors because a whole class of pre-existing owner-path unit tests still asserts the old silent-skip behavior. Beyond the mechanical test update, the fix bakes in an unresolved design choice — the fail-closed/timeout probe paths now emit a hint claiming the graph "is live via the MCP server" even when no server was confirmed. Both must be resolved before merge. Nothing in the *code* is unsafe (the #1913 stdout contract, slot/release, and 3-copy consistency all validated), so this is a finish-the-tests-plus-one-decision blocker, not a redesign.

### Review bar for this PR

Production hook code shipped to end-user machines across three editor integrations, changing a normal (non-error) control path that fires on nearly every agent search in the most common deploy (MCP server running). Bar: (1) CI green on the platform that actually exercises the path; (2) the new behavior is truthful on every path that reaches it; (3) the three hook copies stay identical; (4) the #1913 strict-runner silence contract is preserved. Test correctness is weighted heavily — the changed path is only CI-exercised on one platform.

### Problem being solved

Issue #2396: in the common production deploy the GitNexus MCP server holds the lbug single-writer lock, so the hook's CLI `augment` can't run and the hook was **silently** skipping search-augmentation — the agent got nothing. The fix adds `buildMcpQueryHint(pattern)` to all three hook copies and, on the owner branch, hands the agent an MCP-`query` hint via the sanctioned `additionalContext` stdout channel instead of returning empty.

### Current PR state

Open, cross-repository (fork `magyargergo:fix/2396-hook-mcp-query-hint` → `abhigyanpatwari:main`), single commit `7166137b`, `MERGEABLE` but `mergeStateStatus: BLOCKED`. Quality gates (lint, format, typecheck, typecheck-web) green; ubuntu coverage 1–3 green; windows platform-sensitive 2–3 green; **macOS platform-sensitive 2/3 red** (the blocker). Vercel fail is auth-only, ignore.

### Merge status and mergeability

**Classification: `checks failing`.** `mergeable: MERGEABLE` (no conflicts) but `mergeStateStatus: BLOCKED` by a required check: run 28956185902, job 85915692366 (`tests / macos-latest (platform-sensitive) 2/3`) failed with 9 assertion errors. Not mergeable until the owner/fail-closed test class is reconciled and CI is green.

### Branch hygiene assessment

**Classification: `clean feature/fix PR`.** One commit, five causally connected files (3 hook copies implementing the same fix + their 2 test files), no merge-from-main commits, no unrelated churn, no mixed domains, diff hygiene clean (`git diff --check` clean, no bidi/hidden-Unicode). No rebase or split needed.

---

### 1. Engine breakdown & leg liveness

**Engine A — Claude:** gitnexus-pr-review **swarm** (risk-architect, test-ci-verifier, security-boundary) + **ce-code-review** (correctness, testing, maintainability, project-standards, adversarial). All lanes live.

**Engine B — Codex / GPT-5.5 (xhigh):** all three legs LIVE — `codex-swarm` (Solo, REQUEST CHANGES, P1 stale test), `codex-ce` (7 personas, 4 findings incl. P1 stale test + P2 hint-on-inconclusive-owner), `codex-adversarial` ([high], stale macOS assertions).

Both engines ran the same swarm + ce scaffolding, so the strong signal is a Codex lane agreeing with a Claude lane. The headline P0 is **CI-reproduced + both engines** — the highest-confidence class.

---

### 2. HEADLINE P0 — stale owner-path stdout assertions break the macOS CI shard

**Severity P0 · Blocks merge: yes · [reproduced] (CI-confirmed + empirical repro) · strongest possible corroboration.**

**In-diff anchor:** `gitnexus/test/unit/hooks.test.ts:1868` (the one owner-test the PR *did* update). **Root cause in code:** `gitnexus/hooks/claude/gitnexus-hook.cjs:411` — the owner branch now runs `result = buildMcpQueryHint(pattern)` unconditionally (only the stderr diagnostic stays behind `isDebugEnabled()`), emitted at `:425` via `sendHookResponse('PreToolUse', result)`. Same in `gitnexus-claude-plugin/hooks/gitnexus-hook.js:443` and, via `writeAdditionalContext`, `gitnexus/hooks/antigravity/gitnexus-antigravity-hook.cjs:435`.

**Why a whole CLASS breaks, not 2 tests.** The probe collapses *timeout/guard-death* and *confirmed-owned* to the same `true` (`hook-db-lock-probe.cjs` returns `true` for a real MCP-owner ps match, lsof/ps ETIMEDOUT, guard signal-death, and guard exit 124/137 — verified L629/642/662/668). So BOTH the definite-owner tests AND every fail-closed/timeout/guard-death test route through the owner branch that now emits the hint. The PR updated only 3 sibling tests (the `DEBUG=1` one ~L1867 and the `'0'`/`'false'` loop ~L1911) and missed the baseline "skips SILENTLY" test (L1826/1850) plus the entire `Ladybug DB owner guard … (#1493)` describe (`describe.skipIf(SKIP_LSOF_PATH)`, L1956).

**Why only macOS surfaces it.** `SKIP_LSOF_PATH = process.platform === 'win32' || process.platform === 'linux'` (hooks.test.ts:100). Linux "coverage" and the Windows leg skip all of these; only macOS has `SKIP_LSOF_PATH === false` and executes them. Both `hooks.test.ts` and `antigravity-hook-e2e.test.ts` are in the macOS fixed-file list (`cross-platform-tests.ts`).

**Authoritative failing assertions** (CI check-runs annotations, level=failure; each received the exact hint JSON where it asserted `''`). All `expect(result.stdout.trim()).toBe('')` in `gitnexus/test/unit/hooks.test.ts`:

| Line | Test | Owner path |
|------|------|-----------|
| 1850 | skips augment SILENTLY when a GitNexus MCP process owns the repo DB | DEFINITE owner |
| 1998 | skips augment for real node_modules/gitnexus ps line (npx child) | DEFINITE owner |
| 2062 | skips augment for gitnexus serve child | DEFINITE owner |
| 2150 | ETIMEDOUT lsof → augment skipped (fail-closed) | fail-closed |
| 2186 | ETIMEDOUT lsof → augment skipped SILENTLY by default | fail-closed |
| 2234 | ETIMEDOUT lsof with wrapper disabled → identical fail-closed skip | fail-closed |
| 2287 | bad wrapper rejected by self-test → still fail-closed | fail-closed |
| 2343 | guard exit 124 (budget expiry) → fail-closed skip | fail-closed |
| 2399 | guard signal-death (status null + signal) → fail-closed skip | fail-closed |

9 distinct assertions (10 CI annotations — L1850 double-counts stdout+stderr). Each `it(...)` is parametrized over `label ∈ ['CJS','Plugin']`, so ~18 case failures. **ce-testing additionally predicts a 10th sibling on the same branch — L2524 "multiple PIDs — skip if any ps line is GitNexus MCP" [code-read]** — not in the CI annotation set (cap/ordering) but it reaches the identical collapsed-`true` branch; the fix sweep must catch it too.

**Also (ce-testing):** after correcting L1826/1850, the ONLY remaining `GITNEXUS_DEBUG=''` (unset/default) coverage for the owner-hint scenario disappears — the PR's new loop only iterates `['0','false']`. Add a like-for-like `''` case; don't just delete the stale test.

**Corroboration:** Engine A — coordinator CI-oracle (authoritative) + Claude-swarm test-ci-verifier (P0) + ce-testing (P0, empirical repro). Engine B — codex-swarm (P1), codex-adversarial ([high]), codex-ce (P1). **CI-reproduced + both engines.**

#### The design decision baked into the fix (author must resolve — not a mechanical test update)

The fail-closed paths (L2150–L2399) reach the hint branch because the probe COULD NOT confirm ownership and fails closed (assume-owned). The hint asserts *"Knowledge graph is live via the MCP server"* and points at `mcp__gitnexus__query` — but on a genuine probe timeout with no server running, that claim is unverified/possibly false.

- **Option A — accept fail-closed → emit hint.** Update all ~9–10 owner/fail-closed tests to expect the hint. Simplest, but the "is live" wording is then sometimes untrue and the guidance may point at a non-existent tool.
- **Option B — emit ONLY on CONFIRMED owner; keep fail-closed silent (RECOMMENDED).** Narrow the code so fail-closed still returns empty. Only the 3 definite-owner tests (1850/1998/2062) flip to the hint; the fail-closed tests keep asserting silence. Keeps the hint's claim truthful, addresses C4, and is fewer test changes. Requires exposing the probe's tri-state verdict instead of the collapsed boolean.

Option B is independently corroborated by **codex-ce (P2, 5 reviewers)** and the coordinator. Recommend Option B.

---

### 3. Inline-candidate findings

#### C1 — P0 — Stale owner/fail-closed stdout assertions (the headline) · Blocks merge: yes
See §2. Anchor `gitnexus/test/unit/hooks.test.ts:1868`. Fix per Option B (preferred): narrow the code to emit only on confirmed ownership, flip the 3 definite-owner tests to parse stdout and `expect(output!.additionalContext).toContain('mcp__gitnexus__query')`, keep fail-closed tests asserting silence, add a `GITNEXUS_DEBUG=''` case. If Option A: flip all ~10 assertions. `[reproduced]` · Engines A+B, CI-confirmed.

#### C2 — P2 — Hint emitted on inconclusive/fail-closed owner probes (untruthful "is live") · Blocks merge: maybe
`gitnexus/hooks/claude/gitnexus-hook.cjs:411` (+ plugin `:443`, antigravity `:435`). The owner boolean collapses timeout/guard-death/EACCES into "owned", so the hint fires — claiming the graph is "live via the MCP server" and to call `mcp__gitnexus__query` — even when no server was confirmed. **Trigger:** lsof/ps ETIMEDOUT or guard exit 124 with no MCP server up → agent is told to call a tool that may not exist. **Fix:** consume the probe's tri-state verdict; emit only for confirmed MCP ownership; keep fail-closed silent (debug-gated stderr only) across all 3 copies. This is the code-side of Option B. `[code-read]` · Engine B strong (codex-ce, 5 lanes); Engine A partial (ce-adversarial flagged fail-closed reaches the branch); coordinator-elevated.

#### C3 — P2 — Owner-path hint is unconditional & un-deduped → context bloat + ~2× query amplification · Blocks merge: no
`gitnexus/hooks/claude/gitnexus-hook.cjs:411` emitted `:425` (+ plugin, antigravity). Unlike the non-owner path (relevance-gated: `extractAugmentContext` returns `''` with no marker), the owner path emits the ~200-char hint for ANY qualifying pattern, no dedup, no graph-relevance check. **Trigger (common deploy — MCP server up all session):** 80 Grep/Glob/Bash calls → ~80 near-identical hints (~4,000 tokens), including for patterns the graph knows nothing about; on PreToolUse an obedient agent interleaves a `query` before every search → ~2× tool calls. **Fix (advisory):** dedup once per session/pattern, and/or reword as a one-time capability reminder rather than a per-search imperative. `[code-read]` · Engine A (ce-adversarial P2); Engine B noted but non-blocking.

#### C4 — P2 — `gitnexus serve` (web-UI) owners get an MCP-only hint that may not apply · Blocks merge: no
`isGitNexusServerCommand` matches `(mcp|serve)` (`hook-db-lock-probe.cjs:64`, verified). A `gitnexus serve` process (web server) trips the owner probe and receives *"live via the MCP server … call mcp__gitnexus__query"* — but a serve-only deploy need not expose MCP tools to the session, so the hint can misdirect. **Fix:** split the probe result by mode; emit the MCP hint only for `gitnexus mcp` owners; keep serve silent/debug-only. `[code-read]` (predicate confirmed) · **Engine B only** (codex-ce, api-contract+adversarial). Single-engine → body-level.

#### C5 — P3 — Orphaned JSDoc in the claude `.cjs` copy; `handlePreToolUse` left undocumented · Blocks merge: no
`gitnexus/hooks/claude/gitnexus-hook.cjs:353` — the pre-existing `/** PreToolUse handler … */` doc now sits directly above `buildMcpQueryHint` (L360); `handlePreToolUse` (L368) lost its doc. The plugin copy placed the helper cleanly (after `sendHookResponse`); antigravity has no orphan. **Fix:** move `buildMcpQueryHint` (with its own doc) above the "PreToolUse handler" comment, matching the plugin copy. Cosmetic. `[code-read]` · **CROSS-ENGINE, 4 lanes:** ce-maintainability, ce-correctness, codex-swarm, codex-ce.

---

### 4. Lower-priority / body items

- **Stale README `gitnexus/README.md:538-544` — P2 · Blocks merge: no · [code-read].** The "Hook augmentation … silently skipped" section still lists "when the GitNexus MCP server owns the repo DB" as a silent-skip trigger — the exact case this PR moved to emitting a hint. Update it. **CROSS-ENGINE:** ce-project-standards + codex-ce. (Correctly, no CHANGELOG edit in a feature PR.)
- **Hardcoded `mcp__gitnexus__query` tool name — P3 · Blocks merge: no.** Claude-Code-specific `mcp__<server>__<tool>` form; wrong for a renamed server key or the Gemini/Antigravity client the antigravity copy serves. Mitigated in-text (hedged "(e.g. …)", also names bare `query` + `search_query`; server registers bare `query`, #2059/#2175). codex-swarm judged **acceptable as written**; ce-adversarial P3. Optional hardening.
- **pattern echoed into `additionalContext` — REFUTED as an exploit (presented honestly) · Blocks merge: no.** ce-adversarial (attack #4) + ce-correctness (R3) + the Codex security sweep refute it: `pattern` is the agent's OWN search query (near-zero amplification), Glob is constrained to `[a-zA-Z][a-zA-Z0-9_-]{2,}`, and `JSON.stringify` escapes quotes/backslashes/newlines so no envelope breakout. **Not a blocker** — at most a low-priority note that arbitrary search text is re-emitted into a `[GitNexus]`-branded block.
- **Test gaps (minor) · Blocks merge: no.** (a) No drift-guard test asserting the 3 `buildMcpQueryHint` bodies stay byte-identical, despite the repo institutionalizing exactly this guard elsewhere (ce-maintainability). (b) No hint escaping test feeding a `"`/`\`/newline pattern end-to-end (ce-testing R1). (c) Owner-path hint emission is CI-exercised only on macOS; a platform-independent unit test injecting the probe verdict would close the gap.

---

### 5. What the reviews VALIDATED (refuted attacks — the code is sound)

- **#1913 strict-runner stdout contract is SAFE.** The hint rides the *same* sanctioned `hookSpecificOutput.additionalContext` stdout channel the non-owner augment-success path already emits; #1913 (commit 292f26ec) targeted **stderr** noise and explicitly left "stdout JSON the agent consumes unaffected". Refuted with strong evidence by ce-adversarial, ce-correctness (R5), project-standards, codex-swarm, codex-ce, + coordinator.
- **Slot/release is SAFE (strictly safer).** `release()` is in `finally`, runs exactly once on every acquired path, before stdout emission; the early-`return` → `result=…` change is behavior-equivalent for the slot. ≤3-per-repo cap (#2163) untouched. (ce-adversarial #5, ce-correctness R1.)
- **Three copies byte-identical.** `buildMcpQueryHint` body + hint template identical across all three hook files (verified by coordinator + every lane).
- **No injection exploit; no throw.** `pattern` is JSON-escaped; interpolation is inert data; `buildMcpQueryHint` is pure string concat inside the try/catch/finally. (ce-correctness R2/R3/R4.)
- **No missed 4th copy.** The cursor hook is architecturally distinct (PostToolUse, no DB-owner gate) — correctly untouched.

---

### 6. Repository history considered

#2396 is this session's own root-cause work (the silent skip was introduced when #1913 gated the owner path behind `GITNEXUS_DEBUG`). The 3-copy hook duplication is a deliberate no-shared-module convention (#2059/#1913). The `SKIP_LSOF_PATH` macOS-only gating traces to #2180 (Linux dropped the lsof/ps fallback). The `additionalContext` channel as the sanctioned augmentation path is established (#1913, Codex-compat #2369). No prior regression in these exact files beyond the #2396 chain.

### 7. Back-and-forth avoided by verifying

Confirmed the merge-base and 5-file diff; ran the CI check-runs annotations API to get the authoritative failing-line set (rather than guessing); read `hook-db-lock-probe.cjs` to confirm the timeout/owned collapse (so the fail-closed breakage is understood, not assumed); confirmed `isGitNexusServerCommand` matches `(mcp|serve)`; verified the 3 hint bodies are byte-identical; and confirmed the #1913-safe claim against the actual channel. The empirical repro (ce-testing, staged hook + stubbed probe) reproduced non-empty stdout on the owner path.

### 8. CI status

Run 28956185902: **`tests / macos-latest (platform-sensitive) 2/3` = FAIL** (job 85915692366) — the merge blocker. macOS 1/3 & 3/3 pass; ubuntu coverage 1–3 pass; windows 2–3 pass; all else green except Vercel (auth-only). `mergeStateStatus: BLOCKED`. **Owner-path coverage is macOS-only** — Linux + Windows legs pass green while never exercising the changed behavior; the one lane that runs it is the one that's red.

### 9. Open questions

- Which fix direction does the author want — Option A (emit on fail-closed, update all tests) or Option B (confirmed-owner-only, keep fail-closed silent)? This determines both the code change and how many tests flip.
- Is the repeated per-search hint (C3) acceptable UX for the common always-on-server deploy, or should it dedup?

---

*Automated multi-tool digest (Claude ce + swarm, Codex/GPT-5.5 ce + swarm + adversarial, coordinator CI-oracle) merged with cross-engine confidence weighting. The P0 is CI-reproduced and double-engine; treat C2–C5 and §4 as verify-then-act. Recommended path: adopt **Option B**, which resolves the P0, C2, C4, and the "is live" truthfulness concern together.*

Issue #2409, opened 2026-07-09
Incremental analyze silently fails to commit (commitOK=False) on every large BFS importer-set expansion (>=~150 importers) — 100% reproducible on gitnexus 1.6.8


kam-ship-it:
### Summary

On `gitnexus 1.6.8`, an **incremental** `analyze` run **silently fails to commit** whenever the change touches enough of the graph that the importer-set BFS expansion is large (empirically **≥ ~150 importers added to the writable set** at `BFS depth ≤ 4`). The analyze process exits reporting `commitOK=False` with **no error, stack trace, or non-zero signal** surfaced — the incremental fast path just doesn't persist. A subsequent forced full rebuild (`analyze -f`) then frequently dies too, right after a WAL warning, and only a **rename-aside of the index dir + a fresh from-scratch rebuild** recovers a known-good index.

Small importer expansions (**+1, +2, +8** importers) commit cleanly. The failure is **deterministic by size**, not intermittent.

### Environment

- `gitnexus` **1.6.8** (npm global), CLI: `node --stack-size=2000 …/gitnexus/dist/cli/index.js analyze <repo> --embeddings 0 --skip-agents-md`
- Windows 11, PowerShell 7.5.x, Node (system)
- Repo under analysis: ~4,670 tracked files, ~72,000 graph nodes
- `--stack-size=2000` is already set on every invocation (raising it did **not** fix this — see below)

### Reproduction (observed, 100% of large-BFS runs over 2 days of hourly incrementals)

The hourly incremental daemon logged this exact pattern **on every run where the importer expansion was large**. Representative slice (importer expansion of **+689**):

```
Incremental: changed=12, added=13, deleted=0 (skipping wipe + 4673 unchanged file rows preserved)
Incremental: +689 importer(s) added to writable set (BFS depth ≤ 4)
escalating: commitOK=False force=False           ← incremental did NOT commit, no error
analyze: node --stack-size=2000 …analyze <repo> --embeddings 0 --skip-agents-md -f
Previous analyze run did not complete cleanly (incrementalInProgress flag set); forcing full rebuild…
{"level":40,"name":"gitnexus","msg":"GitNexus: found 968243 byte lbug.wal without lbug.shadow before write open; will rely on LadybugDB replay/recovery instead of deleting pending WAL data."}
forced rebuild did not reach HEAD - attempting safe nuke   ← the -f rebuild also failed
analyze: … -f    (after renaming the index dir aside → succeeds from scratch)
[...]

PR #2410, opened 2026-07-09
fix: surface incremental dirty state diagnostics

koriyoshi2041:
## Summary

- extend the `incrementalInProgress` dirty flag with the last writeback phase and write-set counters
- refresh the flag after importer-BFS, effective-write-set calculation, and before LadybugDB load
- include those details in the recovery log before forcing a full rebuild

## Why

This gives issue #2409 a better failure receipt than a stale `commitOK=False` outcome. If an incremental run stops during a large importer expansion, the next run can report the last known phase plus `toWrite`, importer expansion, effective write-set size, and delete count. For native hard exits, the pre-written metadata still captures the last completed JS-side phase.

## Verification

- `npx tsc --noEmit`
- `npx vitest run test/unit/incremental-orchestration.test.ts`

## Risk

Low. This does not change the recovery strategy or LadybugDB write semantics; it only records extra optional metadata and logs it when the dirty flag is found.


gitnexus/src/storage/repo-manager.ts
```diff
@@ -138,9 +138,23 @@ export interface RepoMeta {
...
     /** When the run started (epoch ms). */
     startedAt: number;
+    /** Last dirty-flag refresh (epoch ms). */
+    updatedAt?: number;
     /** Number of files in the writable set, for diagnostic logs.
      *  `0` on the full-rebuild path (no incremental write set exists). */
     toWriteCount: number;
+    /** Last completed writeback phase before the process stopped. */
+    phase?: string;
+    /** Directly changed/added/deleted files before importer expansion. */
```
magyargergo: This comment says `directWriteCount` includes deleted files, but the value is written from `hashDiff.toWrite.length` / `writableFiles.size`, and `toWrite` is changed + added only. Deleted files are tracked separately via `hashDiff.deleted` and later `deleteCount`. Can we either change the comment to changed/added files, or include deleted files in the counter if that is the intended meaning?

gitnexus/src/core/run-analyze.ts
```diff
@@ -744,10 +744,25 @@ export async function runFullAnalysis(
   // clear it, the on-disk index may be in a half-state. Cheapest path
   // back to a known-good index is to wipe + rebuild from scratch.
   if (existingMeta?.incrementalInProgress) {
+    const dirty = existingMeta.incrementalInProgress;
+    const dirtyDetails = [
+      dirty.phase ? `phase=${dirty.phase}` : undefined,
+      `toWrite=${dirty.toWriteCount}`,
```
magyargergo: This formatter assumes `incrementalInProgress` is the new object shape, but PR head still preserves the legacy boolean shape (`incrementalInProgress: true`) through repo-manager reconciliation. Recovery still works, but the log can become `toWrite=undefined`. Can we guard with `typeof dirty === "object"` and either log a legacy fallback or omit the detailed fields for the old shape?

PR #1850, opened 2026-05-27
fix(web): replace broken Browse-for-folder with upload directory picker

magyargergo:
## Summary

Fixes the "Browse for folder" button in the local path analysis UI, which was completely broken on all platforms (Docker, local Windows, local Linux). The button used `<input type="file" webkitdirectory>` which only exposes relative paths via `webkitRelativePath` — the code extracted just the folder name (e.g. `myproject`), causing the server to reject it with *"path must be an absolute path"*. No browser API can expose absolute filesystem paths, so the approach was fundamentally unfixable.

This PR replaces the broken browser-side directory picker with a server-side directory listing endpoint + a new `DirectoryPicker` modal component. Docker users can now browse `/workspace/` and other container paths directly from the UI.

### Changes

- **`GET /api/fs/list`** endpoint in `api.ts` — lists subdirectories at a given absolute server-side path (rate-limited via `createRouteLimiter`, validates absolute path + rejects traversal)
- **`listDirectories()`** client function in `backend-client.ts`
- **`DirectoryPicker`** modal component — breadcrumb navigation, sorted directory listing, folder selection, loading/error/empty states
- **`RepoAnalyzer.tsx`** — removed broken `webkitdirectory` file input, wired in `DirectoryPicker`
- **i18n** — updated en + zh-CN locale files with new strings
- **Unit tests** — 9 tests covering happy paths, validation, error handling, and rate-limit wiring

### Before / After

| Before | After |
|--------|-------|
| "Browse for folder" sends only folder name → server rejects with "path must be an absolute path" | "Browse server directories" opens a modal that navigates real server-side directories and returns the full absolute path |
| Docker users have no way to discover container paths | Docker users can browse `/workspace/`, `/data/gitnexus/`, etc. directly |
| Manual path typing works (unchanged) | Manual path typing still works (unchanged) |

## Test plan

- [x] `npx vitest run test/unit/api-fs-list.test.ts` — 9 tests passing
- [x] Existing API tests unaffected (`api-file-route`, `server-validation`, `rate-limit` — 51 tests passing)
- [x] TypeScript type check passes (`tsc --noEmit` for both gitnexus and gitnexus-web)
- [x] Pre-commit hooks pass (eslint, prettier, tsc)
- [ ] Manual test: start server, open web UI, click "Browse server directories", navigate to a folder, select → path appears in input field
- [ ] Manual test: Docker compose up, browse to `/workspace/<repo>`, analyze

Closes #1518

PR #2420, opened 2026-07-09
fix(web): use repo path identity in switcher

### LINKED ISSUE #2419: Web UI repo dropdown treats duplicate repository names as the same identity

koriyoshi2041:
## Summary
- use the backend repo path as the Web UI repository identity while keeping `repo.name` as the display label
- preserve path identity through landing selection, header switching, re-analyze, delete, URL persistence (`?repo=` param, restored in preference to the cosmetic `?project=`), analyze completion, the code-reference viewer, and agent/backend calls
- server: `resolveRegisteredRepoEntry` resolves `?repo=` path-first with fail-closed path misses (a stale or wrong path can never silently retarget a same-named sibling); relative slash-shaped params are rejected without a filesystem probe; `GET /api/repo` is rate-limited; the SSE completion payload carries `repoPath`; `/api/repos` emits `repoPath`; delete only removes a clone dir the entry actually owns
- add duplicate-name coverage across header switching/delete, the server resolver (registered in `cross-platform-tests.ts` so Windows CI runs it), URL restore, analyze completion, and the code viewer

Fixes #2419.

## Verification
- `cd gitnexus-web && npx vitest run test/unit && npx tsc -b --noEmit`
- `cd gitnexus && npx vitest run test/unit/repo-manager.test.ts test/unit/server-api-repo-resolution.test.ts test/unit/rate-limit.test.ts test/unit/analyze-api.test.ts test/unit/server-sse-payload.test.ts && npx tsc --noEmit`
- `git diff --check`


gitnexus/src/server/api.ts
```diff
@@ -561,6 +568,38 @@ const requestedRepo = (req: express.Request): string | undefined => {
...
+
+  if (looksLikePath) {
+    const requestedPath = canonicalizePath(repoName);
+    const pathMatch = repos.find((r) =>
+      registryPathEquals(canonicalizePath(r.path), requestedPath),
+    );
+    if (pathMatch) return pathMatch;
+  }
+
+  const normalizedName = repoParamBasename(repoName);
```
magyargergo, line 620: **[P1] Silent basename fallback after path miss re-targets duplicate-name siblings (destructive reach)**

**P1.** After a path-shaped `?repo=` misses the path match, this line starts an unconditional basename fallback — first duplicate-name repo wins. Duplicate basenames register under the identical name (repo-manager.ts:950-955). Trigger: tab 1 deletes `/b/reels`; tab 2's stale row re-sends `DELETE ?repo=/b/reels` → path miss → basename `reels` → resolves `/a/reels` → the sibling's index, clone dir, and registration are destroyed. The same resolver serves `POST /api/embed` and all read routes. LocalBackend's resolver returns null for path-shaped misses (it never basenames) and the CLI resolver throws `RegistryAmbiguousTargetError` on duplicates — this resolver alone basenames a failed path. Fix: return null on path-shaped misses, or fall back only when the basename is unique. Flagged by both engines. [reproduced]

gitnexus-web/src/hooks/useAppState.tsx
```diff
@@ -1225,12 +1235,13 @@ const AppStateProviderInner = ({ children }: { children: ReactNode }) => {
         const repoPath = result.repoInfo.repoPath ?? result.repoInfo.path;
         // Prefer the registry name, then normalize Windows \ and Unix / paths
         const pName =
-          repoName ||
           result.repoInfo.name ||
           (repoPath || '').replace(/\\/g, '/').split('/').filter(Boolean).pop() ||
+          repoName ||
           'server-project';
+        repoIdentity = repoName || repoPath || pName;
```
magyargergo, line 1262: **[P2] URL persistence still keys on the ambiguous display name — #2419 recurs on refresh**

**P2.** `repoIdentity` is computed here, but the persist block just below (line 1280, and App.tsx:84) still writes `?project=<display name>`. With two repos named `reels`, switching to `/b/reels` then pressing F5 restores by name → server picks the first `reels` → user is silently back on `/a/reels` with the Active badge confidently wrong. Commit 24cf50dc shows this is a deliberate scope cut — fine, but the PR body claims path identity is preserved through "URL persistence", which it isn't. Please either persist `?repo=<identity>` (preferred on restore, keep `?project=` cosmetic) or correct the PR description and note the limitation in #2419. Flagged by both engines (all 3 Codex legs; Claude adversarial/correctness/facts). [code-read]

gitnexus/src/server/api.ts
```diff
@@ -833,20 +872,9 @@ export const createServer = async (port: number, host: string = '127.0.0.1') =>
...
-      found =
-        repos.find((r) => r.name === normalizedName) ||
-        repos.find((r) => r.name.toLowerCase() === normalizedName.toLowerCase()) ||
-        null;
-    } else if (repos.length > 0) {
-      found = repos[0]; // default to first repo
-    }
+    const found = resolveRegisteredRepoEntry(repos, repoName);
+
+    const normalizedName = repoName ? repoParamBasename(repoName) : undefined;
```
magyargergo, line 904: **[P2] Recovery paths drop the path identity back to basename (hold-queue at 916, retry at 937)**

**P2 (pre-existing, now inconsistent).** `normalizedName` computed here feeds two unchanged branches that discard the caller's path: hold-queue completion resolves by name-only `find` (line 916) and the deep-init retry recurses with the basename (line 937). These fire exactly during the registration races they exist to handle: with a duplicate-name sibling the held/retried request resolves the wrong repo; for an aliased repo (name ≠ basename — your own test fixture shape) the basename retry 404s a path that IS registered post-init. Fix: `resolveRepo(repoName, true, req)` at 937 and `resolveRegisteredRepoEntry(freshRepos, repoName)` at 916. Flagged by both engines (Codex swarm; Claude correctness/api-contract/adversarial). [code-read]

gitnexus-web/src/components/Header.tsx
```diff
@@ -248,125 +252,130 @@ export const Header = ({
...
+                                          force: true,
+                                        });
+                                        reanalyzeSseRef.current = streamAnalyzeProgress(
+                                          jobId,
+                                          (p) => setReanalyzeProgress(p),
+                                          () => {
+                                            setReanalyzing(null);
+                                            setReanalyzeProgress(null);
+                                            reanalyzeSseRef.current = null;
+                                            onAnalyzeComplete?.(identity);
```
magyargergo, line 313: **[P2] onAnalyzeComplete mixes vocabularies: re-analyze passes a PATH, RepoAnalyzer passes a NAME**

**P2.** This branch correctly passes `identity` (a path), but the RepoAnalyzer branch above (lines 219-223) passes the worker's `repoName` into the same callback. App.tsx:280 feeds it to `connectToServer` unchanged. Trigger: `/a/reels` indexed → "Analyze new" on `/b/reels` (registers as a second `reels`) → completion connects by name → server resolves the FIRST `reels` → user lands on `/a/reels` marked Active, believing it's the repo they just analyzed. Same class in the onboarding flow (RepoLanding.tsx:150). Fix: pass the analyzed path (the client supplied it to `startAnalyze`; the SSE done payload already carries `repoPath`). [code-read]

gitnexus/test/unit/server-api-repo-resolution.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    const aliased = entry({
+      name: 'e2e-mini-repo',
+      path: '/tmp/gitnexus-e2e-repo',
+      storagePath: '/tmp/gitnexus-e2e-repo/.gitnexus',
+    });
+
+    expect(resolveRegisteredRepoEntry([aliased], '/tmp/gitnexus-e2e-repo')).toBe(aliased);
+  });
+
+  it('falls back to basename/name matching for older callers', () => {
```
magyargergo, line 25: **[P2] The PR's core invariant — duplicate name, distinct paths — is never tested**

**P2 (rated P1 by two lanes).** Both tests use single-entry registries, so the path-match branch could be deleted and this suite stays green. Please add:
```ts
const a = entry({ name: 'reels', path: '/ws/group-a/reels' });
const b = entry({ name: 'reels', path: '/ws/group-b/reels' });
expect(resolveRegisteredRepoEntry([a, b], '/ws/group-a/reels')).toBe(a);
expect(resolveRegisteredRepoEntry([a, b], '/ws/group-b/reels')).toBe(b);
expect(resolveRegisteredRepoEntry([a, b], 'reels')).toBe(a); // lock in first-wins
```
Also worth adding: a Windows-shaped input (`'C:\\ws\\reels'` — empirically falls into the ambiguous fallback on POSIX), the `!repoName → repos[0]` branch, and registering this file in `scripts/cross-platform-tests.ts` (the repo-manager tests for these same helpers are listed there; Windows CI won't run this file otherwise). Flagged by both engines. [reproduced]

gitnexus/src/server/api.ts
```diff
@@ -561,6 +568,38 @@ const requestedRepo = (req: express.Request): string | undefined => {
...
+
+export const resolveRegisteredRepoEntry = (
+  repos: RegistryEntry[],
+  repoName?: string,
+): RegistryEntry | null => {
+  if (!repoName) return repos[0] ?? null;
+
+  const looksLikePath =
+    path.isAbsolute(repoName) ||
+    repoName.includes(path.sep) ||
```
magyargergo: **[P3] looksLikePath is broader than needed — relative slash-strings trigger FS probes; path.sep check is redundant**

**P3.** `includes(path.sep)` is subsumed by the `'/'` and `'\\'` checks on every platform, and matching relative slash-strings (`org/name`) sends CWD-relative, attacker-influenced input into `realpathSync.native` (sync FS syscall) on un-rate-limited GET routes (/api/repo, /api/graph). Registered identities are always absolute (`registerRepo` does `path.resolve`), so narrowing to `path.isAbsolute(repoName) || repoName.includes('\\')` keeps Unix-absolute + Windows-on-Linux handling while dropping CWD probing. Optionally add `createRouteLimiter()` to GET /api/repo for parity with /api/file//api/grep. (Claude security-boundary + ce-security, same mechanism; impact bounded — timing-only side channel.) [code-read]

gitnexus-web/src/components/Header.tsx
```diff
@@ -248,125 +252,130 @@ export const Header = ({
...
+                                        {t('header:active')}
+                                      </span>
+                                    )}
+                                  </button>
+                                  {/* Re-analyze */}
+                                  <button
+                                    onClick={async (e) => {
+                                      e.stopPropagation();
+                                      if (reanalyzing) return; // already running
+                                      setReanalyzing(identity);
```
magyargergo, line 295: **[P3] Re-analyze progress label now renders the absolute path**

**P3.** `reanalyzing` now stores the identity (path), and the progress bar at lines 396-401 interpolates it directly: users see "Re-analyzing /workspace/group-b/reels: …" in a 320px dropdown (line 333 still shows `repo.name` for the same repo). Same class: the switch overlay shows "Loading /abs/path…" (useAppState.tsx:1173) and `loadGraphAnyway` feeds the path into the agent prompt as the project name (useAppState.tsx:1405). Keep the identity for state matching but store/look up the display name for rendering — e.g. `{identity, name}` in state or `availableRepos.find(...)` at render time. (Claude correctness + adversarial, deterministic on every re-analyze.) [code-read]

magyargergo (commented):
# Two-engine tri-review digest — PR #2420 "fix(web): use repo path identity in switcher"

**Engines & legs that ran.** Claude: 5 swarm lanes (facts-historian, branch-hygiene, risk-architect, test-ci-verifier, security-boundary) + 7 ce-code-review personas (correctness, adversarial, security, testing, maintainability, project-standards, api-contract). GPT-5.5 (Codex): gitnexus swarm Solo (full 6-lane report, 5 findings), ce-code-review report-only (its internal sub-agent dispatch failed with deployment 404s, so it completed single-session; 3 findings), and a native adversarial red-team (2 findings). All three Codex legs produced substantive final reports — leg liveness confirmed. Every finding below was re-verified against PR-head code (`24c442e6`) by the synthesis pass; line numbers are PR-head.

**Verdict.** The core fix is sound and well-made: multiple lanes on both engines independently traced the switch, delete, re-analyze (header branch), landing-select, and active-badge flows end-to-end and found them correct — the dropdown now sends registry paths, `resolveRegisteredRepoEntry` path-matches first via the canonical `canonicalizePath`/`registryPathEquals` pair, React keys are collision-free, the #2381 search behavior is preserved, and both changed/new test files pass locally (test-ci lane executed them: 2/2 and 6/6). The residual risk is concentrated in the *fallback* paths: a silent basename fallback after a path miss (reachable from destructive endpoints — the one issue we'd fix before merge), name-based recovery/completion paths that predate the PR but now contradict its guarantee, and a deliberate URL-persistence scope cut that the PR description overstates. Branch hygiene is acceptable (merge-from-main commit follows repo convention; intra-PR commit style is compliant per CONTRIBUTING).

---

## Findings

### 1. P1 — Silent basename fallback after a path-shaped miss re-targets a duplicate-name sibling, including on DELETE and embed
- **File:** `gitnexus/src/server/api.ts:586-600` (new code, in diff)
- **Agreement:** Codex: swarm (P1, blocks-merge) + ce (P1) + adversarial (high) — Claude: adversarial (P1), correctness (P2), testing (empirically executed the resolver), api-contract (P2). Strongest cross-engine corroboration in the review.
- **Trigger (deterministic, two tabs):** Registry holds `/a/reels` and `/b/reels` — both named `reels` (verified: `repo-manager.ts:950-955` registers un-aliased basename collisions silently under the *identical* name; no `-2` suffix). Tab 1 deletes `/b/reels` (path match — correct). Tab 2's stale row re-issues `DELETE /api/repo?repo=%2Fb%2Freels` → path match misses (entry unregistered; the local dir still realpaths cleanly) → fallback basenames to `reels` → `find` returns `/a/reels` → **the sibling's index, clone dir, and registration are destroyed**, and the response `{deleted:'reels'}` looks correct. Same resolver serves `POST /api/embed` (api.ts:1724 — writes embeddings into the sibling's DB) and all read routes (stale tabs silently receive the sibling's data with no 404 signal); a same-basename sibling also makes the hold-queue/404 machinery unreachable for path requests (`found` is truthy at api.ts:883).
- **Evidence:** `api.ts:586-592` path tier, `594-600` unconditional basename/name fallback; DELETE routes through `resolveRepo` at `api.ts:1045`. Both sibling resolvers are stricter for exactly this input class: `LocalBackend.resolveRepoFromCache` (src/mcp/local/local-backend.ts:1274-1331) never extracts a basename from a path-shaped param — its name/id/partial tiers compare the **full param string** against registry names, so a stale path miss returns null — and the CLI's `resolveRegistryEntry` (repo-manager.ts:1480-1506) throws `RegistryAmbiguousTargetError` on duplicate names. The new api.ts resolver is the only one that basenames a failed path-shaped lookup and retargets. (Once one sibling is deleted, a bare-*name* lookup legitimately resolves the survivor in every resolver — the divergence is specifically path-shaped input.) `[code-read]` + `[reproduced]` (Claude testing lane ran the real function: stale path among duplicates → first sibling).
- **Fix:** When `looksLikePath` is true and the path match misses, return null (letting the existing 404/hold-queue/deep-init machinery run), or at minimum fall back only when exactly one repo bears that basename; align with `resolveRegistryEntry` semantics. If lenient fallback must stay for legacy read callers, gate destructive endpoints (DELETE, embed) with a strict flag. Optional hardening: disable the Header delete button while a delete is in flight (the re-analyze button has this guard; delete does not — Header.tsx:341-375).
- **Postable-inline:** yes (api.ts:594).

### 2. P2 — URL persistence/restore still keys on the ambiguous display name: #2419 recurs on every refresh, and the PR body overstates
- **File:** `gitnexus-web/src/hooks/useAppState.tsx:1280`, `gitnexus-web/src/App.tsx:84` (unchanged lines); restore at `App.tsx:124/173`; server name-match `api.ts:597`
- **Agreement:** Codex: swarm (P1) + ce (P1) + adversarial (high) — Claude: adversarial (P2), correctness (P3), facts (P3, "confirmed design decision").
- **Adjudication:** Commit `24cf50dc` ("keep repo URL project names stable") *deliberately* reverted path-in-URL, and the behavior is byte-identical to pre-PR main — Codex's "P1 blocks-merge" overstates a non-regression, but Claude's "old behavior preserved" understates the consequence: the PR body claims path identity is preserved through "URL persistence", which is false, and the headline #2419 failure resurrects on every F5 — switch to `/b/reels`, refresh, and you are silently back on `/a/reels` with the Active badge confidently marking the wrong row. Landed at **P2**: intentional scope cut that needs explicit maintainer acknowledgment, a corrected PR description, and ideally a small follow-up (persist `?repo=<identity>` alongside the cosmetic `?project=`, preferred on restore). `[code-read]` — persist/restore chain re-verified at both sites.
- **Postable-inline:** yes — anchored at useAppState.tsx:1242 (in-diff identity line adjacent to the persist block), cross-referencing 1280 / App.tsx:84.

### 3. P2 — resolveRepo's recovery paths degrade the path identity to a basename (hold-queue completion + deep-init retry)
- **File:** `gitnexus/src/server/api.ts:916` (`freshRepos.find((r) => r.name === normalizedName)`), `:937` (`resolveRepo(normalizedName, true, req)`) — both PR-head line numbers on unchanged lines; the feeding `normalizedName` at `:877` is an added line in the diff (the inline anchor)
- **Agreement:** Codex: swarm (P2, cites 916) — Claude: correctness (P2), api-contract (P2), adversarial (P3), facts (P3). Cross-engine.
- **Adjudication:** Both branches are pre-existing and byte-identical to main (old web sent names, and `path.basename(name) === name`, so nothing was lost). This is a **consistency gap, not a regression**: the fallbacks fire precisely during the registration races they exist to handle, and they now silently discard the disambiguating path the new web supplies. With a duplicate sibling, the held/retried request resolves the wrong repo (and `switchRepo` stamps `currentRepo` with the *requested* path, so the badge marks the requested row while the graph serves the sibling); for an aliased repo (name ≠ basename — the PR's own test fixture shape), the basename retry 404s a path that IS registered post-init. `[code-read]`
- **Fix:** Retry with the original identity — `resolveRepo(repoName, true, req)` at 937 — and re-run `resolveRegisteredRepoEntry(freshRepos, repoName)` at 916.
- **Postable-inline:** yes — anchored at api.ts:877 with cross-reference to 916/937.

### 4. P2 — onAnalyzeComplete identity mismatch: analyze-new completion passes the NAME and can connect the wrong sibling
- **File:** `gitnexus-web/src/components/Header.tsx:219-223` (RepoAnalyzer branch passes the worker's `repoName`) vs `:309` (re-analyze branch passes `identity`); consumer `App.tsx:280` (`connectToServer(url, undefined, undefined, repoName)`); same class at `RepoLanding.tsx:150` (onboarding)
- **Agreement:** Claude: adversarial (P2, 0.75) — Codex swarm's finding 3 corroborates the *outcome* ("analyze completion can reconnect the wrong repo") via the server-side path. Claude risk traced only the re-analyze branch (identity — sound) and did not examine the analyze-new branch, so its "resolves correctly server-side" does not refute this. Adjudicated real by code read.
- **Trigger:** `/a/reels` indexed; header dropdown → "Analyze new" → analyze `/b/reels` (registers silently as a second `reels`). Completion fires `onComplete('reels')` → `connectToServer(..., 'reels')` → server name-matches the FIRST entry → user lands on `/a/reels`, marked Active, believing they're on the repo they just analyzed. `[code-read]`
- **Fix:** Pass the analyzed path from the completion flow — the client supplied `path` to `startAnalyze`, and the SSE done payload already carries `repoPath` (`analyze-worker-ipc.ts`). Type the callback param so name/path vocabularies can't silently mix.
- **Postable-inline:** yes — anchored at Header.tsx:309 (in diff), cross-referencing the RepoAnalyzer branch at 219-223.

### 5. P2 — The PR's core invariant (duplicate name, distinct paths) has no test
- **File:** `gitnexus/test/unit/server-api-repo-resolution.test.ts:13-30` (new file, in diff)
- **Agreement:** Claude: test-ci (rated P1), ce-testing (rated P1, `[reproduced]` — probed the real function via tsx), risk (P2) — Codex: swarm lane-4 (missing duplicate-name coverage). Cross-engine.
- **Evidence:** Both `it` blocks build single-entry registries; no test constructs two same-name entries, so a refactor deleting the path-match branch ships green. The empirical probe confirmed ambiguous bare-name input resolves to the first-registered duplicate — behavior worth locking in explicitly. Also missing: Windows-shaped input (`'C:\\ws\\reels'` on POSIX empirically falls into the ambiguous fallback), the `!repoName → repos[0]` branch, and a negative case (unregistered path-shaped input → null).
- **Suggested test:**
  ```ts
  it('disambiguates duplicate names by exact path', () => {
    const a = entry({ name: 'reels', path: '/ws/group-a/reels' });
    const b = entry({ name: 'reels', path: '/ws/group-b/reels' });
    expect(resolveRegisteredRepoEntry([a, b], '/ws/group-a/reels')).toBe(a);
    expect(resolveRegisteredRepoEntry([a, b], '/ws/group-b/reels')).toBe(b);
    expect(resolveRegisteredRepoEntry([a, b], 'reels')).toBe(a); // documented first-wins
  });
  ```
- Additionally (project-standards, precedent verified): the helpers under test are platform-sensitive (`path.isAbsolute`/`path.sep`, `realpathSync.native`, win32-cased `registryPathEquals`), and the sibling `repo-manager` tests ARE listed in `gitnexus/scripts/cross-platform-tests.ts:41-42` — this new file is not, so Windows CI never runs it. Add it there.
- **Postable-inline:** yes (test file line 24).

### 6. P2 (body-only) — Delete handler's clone-dir cleanup is name-keyed: deleting duplicate B can rm -rf same-named clone A
- **File:** `gitnexus/src/server/api.ts:1074-1089` — NOT in the diff (pre-existing), so body-only.
- **Agreement:** Claude correctness (P2, conf 60), single lane — verified by synthesis `[code-read]`.
- **Evidence/trigger:** Step 2 of the delete handler derives the clone dir from the entry NAME (`getCloneDir(entry.name)`, api.ts:1076) with no ownership check, while step 2b's own comment (api.ts:1091-1093) explains why cleanup must be driven off `entry.path` "so a same-named clone is never affected". With a server-cloned repo A named `reels` (clone at `~/.gitnexus/repos/reels`) plus a local duplicate B also named `reels`, the PR's now-path-precise delete of B correctly resolves B — then wipes A's working clone. Pre-existing latent bug, but this PR is what makes targeting the non-first duplicate reachable from the web UI.
- **Fix:** Only remove `cloneDir` when it is the deleted entry's own path (`registryPathEquals(canonicalizePath(cloneDir), canonicalizePath(entry.path))`).

### 7. P2 (body-only) — Code viewer still scopes file reads by display name
- **File:** `gitnexus-web/src/components/CodeReferencesPanel.tsx:229-236` — NOT in the diff, so body-only.
- **Agreement:** Codex ce (rated P1, single leg) — verified by synthesis `[code-read]`; no Claude lane examined this file.
- **Evidence/trigger:** `readFile(selectedFilePath, { ...options, repo: projectName || undefined })` sends the display name, while the agent's readFile now uses the path identity (useAppState.tsx:660). With duplicates, after switching to `/b/reels`, clicking a node's code view resolves `reels` server-side → first sibling → the panel renders the WRONG repo's file contents under the right filename (both siblings often share file paths, so it fails silently). Demoted from Codex's P1 to P2: wrong-content display, non-destructive, single leg — but a genuine incomplete-fix hole in the PR's identity model.
- **Fix:** Read `currentRepo` from `useAppState()` and pass `currentRepo ?? projectName` (all three `repo:` sites in the effect).

---

## Validated / refuted suspicions

Probed and cleared (lane attribution in parentheses):
- **Case-sensitivity divergence** web `===` vs server `registryPathEquals` — refuted: every compared identity string originates from the same registry `entry.path`; server compares run through one `canonicalizePath` on one machine (Claude adversarial).
- **`repoPath` vs `path` divergence via /api/repos shape** — refuted: `/api/repos` returns `path`, `/api/repo` returns `repoPath`, both serialize `entry.path` (api.ts:996/1025, re-verified); `fetchRepoInfo` normalizes; both sides of the active-badge compare land on the same string (Claude risk + adversarial).
- **React duplicate keys / render loops** — refuted: registry dedups by canonical path at write time; `setCurrentRepo` is a stable `useCallback([])`, dependency arrays correct, no loop (Claude risk, adversarial). Caveat for pre-canonicalization registries noted below.
- **Path traversal / arbitrary FS ops via `?repo=`** — refuted: raw input is only a lookup key; every FS sink (delete, /api/file root, storage paths) derives from the resolved registry entry; the PR #1850 containment idiom is intact (Claude security + security-boundary, independently).
- **#2381 dropdown-search regression** — refuted: filter still matches `repo.name` (Header.tsx:105) (Claude risk, facts; Codex swarm lane 2).
- **Version skew, common case** — both directions sound when `name === basename(path)`: new-web→old-backend degrades to pre-PR behavior; old-web→new-backend takes the unchanged name path (Claude risk, adversarial, api-contract; Codex swarm partially dissents — see alias note below).
- **Codex swarm finding 4 (alias version-skew) vs Claude clearances — adjudicated:** the Claude refutations only covered `name === basename`; for an *aliased* repo (`--name`, name ≠ basename) the new web sends the path and an OLD backend's `path.basename` resolver (verified at base: old api.ts:840) yields a basename that matches no name → 404 where the old web worked. Mechanism real, population narrow (hosted web + outdated backend + aliased repo): landed as P3, not Codex's P2 — see lower-priority notes.
- **Risk R2 (switch-test selector ambiguity, header.test.tsx `getAllByText('reels')[1]`) vs test-ci — adjudicated for test-ci:** the suite was actually run (6/6 green) and ce-testing hand-traced the DOM order (badge→row-a→row-b) confirming the index is currently correct. Kept only as a P3 brittleness note, not a defect.
- **header.test.tsx delete test hitting the network** — refuted: `backend-client` is properly `vi.mock`'d (Claude testing).
- **api.ts import side effects breaking the unit test** — refuted: only a `package.json` require at module scope; JobManager/LocalBackend instantiate inside `createServer` (Claude facts, test-ci, risk).
- **Full FS paths leaking into shared URLs** — refuted: `?project=` still carries the display name (Claude security) — which is precisely why finding 2 exists.
- **Commit style / no-new-`any` / vitest-branching / i18n / CHANGELOG / scope-minimality** — all compliant; CONTRIBUTING explicitly allows any intra-PR commit style, and the PR title is conventional (Claude project-standards, branch-hygiene).
- **Delete-during-re-analyze corruption** — refuted: `reanalyzing === identity` scopes the SSE abort; server-side repo lock returns 409 (Claude adversarial).

## Lower-priority notes (P3 — no inline comments)

- **Full path rendered in user-facing labels** (Claude correctness + adversarial): re-analyze progress reads "Re-analyzing /abs/path…" (`Header.tsx:396-401`, `reanalyzing` now holds the identity set at :291) and the switch overlay shows "Loading /abs/path…" (`useAppState.tsx:1173`, `repoName` is now the path). Deterministic cosmetic regression; keep identity for state keys, pass `repo.name` to i18n. (Inline comment included at the in-diff `setReanalyzing(identity)` line.)
- **`loadGraphAnyway` feeds the path into the agent prompt as the project name** (`useAppState.tsx:1405`, unchanged line whose semantics changed: `initializeAgent(repo, …)` puts `repoRef.current` — now a path — into `effectiveProjectName` → `buildCodebaseContext`). Chat-only connect → "Load graph anyway" → the LLM context header/file-tree root becomes the absolute path. Fix: `initializeAgent(undefined, { chatOnly: false, repo })`. (Claude correctness + adversarial; the security lane's "no path to LLM" clearance covered the standard flow, not this call site.)
- **Latent trap:** `initializeAgent`'s `setCurrentRepo(opts?.repo ?? overrideProjectName)` fallback (`useAppState.tsx:644-646`, in diff) lets any future `initializeAgent(projectName)` caller — the pre-PR idiom — overwrite the path identity with a display name, killing the Active badge and re-admitting #2419 through the agent path. All current call sites verified safe (Claude adversarial). Only `opts.repo` should ever write identity.
- **`looksLikePath` breadth + sync realpath on user input** (Claude security-boundary P2 / ce-security P3 — same mechanism, landed at P3 given bounded realpath cost, uniform responses, and CORS-gated reads): `includes(path.sep)` is redundant (`/`+`\\` cover both platforms), and matching relative slash-strings (`org/name`) sends CWD-relative probes into `realpathSync.native` on un-rate-limited `GET /api/repo`/`/api/graph`. Narrow to `path.isAbsolute(repoName) || repoName.includes('\\')` (or pre-check registry emptiness); consider a `createRouteLimiter()` on `GET /api/repo` for parity with `/api/file`/`/api/grep`. (Inline comment included — crisp one-line fix on an in-diff line.)
- **`repoIdentity` helper duplicated** (Header.tsx:108 in-body per-render vs RepoLanding.tsx:90 module-level), while `backend-client.ts` already owns the `repoPath ?? path` normalization (`fetchRepoInfo`:559) but `fetchRepos` doesn't normalize and `/api/repos` doesn't emit `repoPath` at all (api.ts:994-1001) — the whole scheme silently relies on `/api/repo.repoPath === /api/repos.path`. Export one `repoIdentity` from backend-client (or normalize in `fetchRepos`), and consider emitting `repoPath` from `/api/repos`. (Claude maintainability, project-standards, risk, facts — multi-lane, one engine.)
- **Alias version-skew** (Codex swarm only, mechanism verified): hosted (auto-updated) web + old local backend + `--name`-aliased repo → new web sends the path, old server basenames it, alias never matches → connect/switch 404s where the old web worked. Narrow population; worth a release-note line rather than a code change.
- **Pure resolver lives in api.ts** (Codex swarm P3 + Claude facts F6): the unit test imports the full server module (heavy import chain), and `src/server/**` is excluded from vitest coverage, so the new tests don't count toward coverage. Moving `resolveRegisteredRepoEntry`/`repoParamBasename` to a small pure module fixes both. Related: a short JSDoc documenting the intentional name-fallback contract (project-standards PS-4).
- **CodeQL bot already flagged** the dead initializers at `useAppState.tsx:1196-1197` (`pNameStr`, `repoIdentity` initial values provably never read — confirmed genuine by the maintainability lane). Not re-flagged inline per bot-dedup policy; drop the initializers when convenient.
- **PR description gap** (project-standards PS-3): the body describes only the web half; the server resolver change and `cd gitnexus && npx tsc --noEmit && npm test` verification are undocumented. Maintainer-editable without fork access — and per finding 2, the "URL persistence" claim should be removed/corrected at the same time.
- **Pre-canonicalization registries** (Claude adversarial residual): two entries whose raw paths differ but canonicalize identically can both render and cross-target each other's (genuinely distinct) storage dirs. Stale-registry population only.

## Test gaps

1. Duplicate-name + distinct-path disambiguation for `resolveRegisteredRepoEntry` (finding 5 — the PR's core invariant; suggested code above), plus documented-first-wins for ambiguous bare names.
2. Negative/containment case: unregistered path-shaped `?repo=` → null (no default-to-sibling) — currently the "falls back for older callers" test *asserts the fallback as desired* with a single-entry registry, masking finding 1.
3. Windows-shaped input (`'C:\\ws\\reels'`) — pure string test, runs on POSIX; plus register the new test file in `scripts/cross-platform-tests.ts` (precedent: repo-manager tests, lines 41-42).
4. `!repoName → repos[0] ?? null` default branch (hit on every no-param connect).
5. Client identity threading: renderHook test that `switchRepo('/ws/group-b/reels')` sets `currentRepo` to that identity and threads it into the next backend call (the existing load-graph-anyway fixture can't distinguish old from new behavior — name/path/requested all coincide).
6. Duplicate-name URL-restore flow (switch to second sibling → simulate `?project=` reconnect → assert identity) — locks in whatever finding-2 decision the maintainer takes.
7. RepoLanding has no test file at all; the landing analog of the header duplicate-name tests is uncovered.

## CI status (as of review)

By the end of this review, every CI check on head `24c442e6` has completed **green** — quality/lint/format/typecheck, CodeQL, gitleaks, e2e chromium, node-floor, benchmarks, packaged-install smoke (ubuntu+windows), tree-sitter ABI (3 platforms), all coverage and platform-sensitive shards (ubuntu/macOS/windows), Docker builds — except **Vercel**, which fails with "authorization required to deploy" (a fork-PR deployment gate, not a code signal); mergeStateStatus is UNSTABLE on that check alone. Earlier-SHA full report: 14335 passed / 60 skipped / 0 failed. Both changed/new test files were also executed locally in the PR worktree and pass (server-api-repo-resolution 2/2; header 6/6). CodeQL bot left 2 inline comments (dead initializers, see notes) — real but cosmetic.

---

*This is an automated two-engine, multi-lane review digest (Claude: gitnexus swarm + ce-code-review personas; GPT-5.5: gitnexus swarm + ce-code-review + native adversarial), cross-checked and re-verified against PR-head code by a synthesis pass. Findings should be verified before acting.*

koriyoshi2041:
Thanks for the detailed review. I pushed `439bd8a6` to address the resolver-path findings here:

- path-shaped registry misses now return null instead of falling back to a duplicate basename, so stale DELETE/embed/read requests cannot retarget a sibling repo
- the hold-queue completion and deep-init retry paths now re-resolve with the original repo identity instead of the basename
- added a duplicate-basename regression test covering path miss vs exact path match

Verified locally:
`npm run test -- test/unit/server-api-repo-resolution.test.ts`
`git diff --check HEAD~1 HEAD`

magyargergo:
@koriyoshi2041 can you please test the recent changes in the web ui? 🙏

koriyoshi2041:
@magyargergo I checked the new head `9912584f` locally for the recent web UI identity changes. After building `gitnexus-shared`, this focused web test set passes:

```
npm run test -- test/unit/header.test.tsx test/unit/switch-repo-url.test.tsx test/unit/url-restore.test.ts test/unit/backend-client-repo-identity.test.ts test/unit/code-references-panel.test.tsx test/unit/repo-analyzer-complete-identity.test.tsx test/unit/initialize-agent-identity.test.tsx test/unit/load-graph-anyway.test.tsx
```

Result: `8 passed`, `36 passed`.

This covers the UI identity paths I was worried about: duplicate-name repo switching in the header, URL restore with `repo` identity, backend `repoIdentity` normalization, code-reference reads scoped to `currentRepo`, analyze completion reconnecting by path identity, `initializeAgent` not clobbering the active repo identity, and `loadGraphAnyway` keeping display labels separate from path identity.

I did not run the full browser e2e/manual UI pass here; this was targeted unit/component coverage for the repo-identity UI paths.

magyargergo:
@koriyoshi2041 can you please manually check it? 🙏 and provide screenshots and videos

magyargergo:
I'll add playwrigt e2e tests so it will be covered.

magyargergo:
image

image

image

PR #2394, opened 2026-07-07
fix(ci): shard platform-sensitive matrix + spawn built CLI to fix Windows cross-platform timeout

magyargergo:
## Problem

The `windows-latest (platform-sensitive)` CI job was **failing at its 15-min internal vitest watchdog** (`scripts/run-cross-platform.ts` → `"vitest timed out after 15 minutes"`), consistently across retries — e.g. run 28865801028 (18.2 min, timed out) while macOS ran the identical set in ~6 min total.

It's **cumulative slowness, not a hang**: the fixed 72-file cross-platform suite is dominated by ~50 CLI/worker **process spawns**, each doing a full `analyze`. Windows is ~5× slower than macOS at process startup, and every e2e test spawned the CLI via `node --import tsx src/cli/index.ts` — re-transpiling the whole CLI source graph on **every** spawn. The suite had crept past the watchdog (main history: 9–14 min, this run tipped over).

## Fix (two complementary changes, no test assertions touched)

1. **Shard the platform-sensitive matrix** — `os: [windows, macos] × shard: [1, 2]`, forwarding `--shard=i/2` through `run-cross-platform.ts` to vitest, which partitions the fixed file list deterministically (sha1, equal file-count, never splits a file). Each runner now does ~half the work with comfortable headroom.

2. **Spawn the built CLI in CI** — new `test/helpers/cli-entry.ts` (`CLI_SPAWN_PREFIX`) spawns `dist/cli/index.js` when `GITNEXUS_E2E_CLI=dist` (set on the cross-platform job, which already runs `build: true`), skipping the per-spawn tsx transpile. It **defaults to tsx-on-source** so local runs always reflect current source; `GITNEXUS_E2E_CLI=dist` on an unbuilt tree throws an actionable *"run `npm run build`"* error. dist is **opt-in only** — never inferred from a generic `CI` env — so an ambient `CI=1` can't silently run a stale build. Converted 8 spawn-based e2e suites + added `test/unit/cli-entry.test.ts`.

The Ubuntu coverage job leaves `GITNEXUS_E2E_CLI` unset, so the **tsx-on-source path stays exercised in CI** too — both entry points covered.

## Measured impact (Linux; larger on Windows where transpile is a bigger share)

| suite | tsx | dist | Δ |
|---|---|---|---|
| `cli-limit-e2e` | 121.5s | 91.0s | −25% |
| `cli-e2e` | 289s | 217s | −25% |

## Notes for reviewers

[...]

PR #2424, opened 2026-07-10
fix(cli): actionable diagnostics for non-4K page-size buffer manager failures

### LINKED ISSUE #1231: ARM64 16K page size: COPY fails with 'Releasing physical memory associated with a frame failed' on Raspberry Pi 5
## Problem

`gitnexus analyze` fails on Raspberry Pi 5 (16 GB RAM, aarch64) with:

```
  ██████████████████████████░░░░░░░░░░░░░░  64% | Loading nodes 1/12: File (166 rows)
  Analysis failed: COPY failed for File: Buffer manager exception: Releasing physical memory associated with a frame failed with error code -1: Invalid argument.
```

This is **distinct from** #785 (8TB mmap on low-RAM ARM64). The patch from #785 (`maxDBSize=256MB` on all 4 `new lbug.Database()` calls) was applied but does not help — same error.

## Root Cause

Raspberry Pi 5 default kernel uses **16 KB page size**, while LadybugDB's buffer manager assumes 4 KB. `munmap()` returns `EINVAL` when the unmap length isn't a multiple of the actual page size.

```
$ getconf PAGE_SIZE
16384
$ uname -r
6.12.47+rpt-rpi-2712
```

The error "Releasing physical memory associated with a frame failed with error code -1" matches `munmap(2)` returning -1 with errno=EINVAL on misaligned size.

## Environment

- **Hardware:** Raspberry Pi 5 Model B Rev 1.1
- **OS:** Raspberry Pi OS Bookworm
- **Kernel:** 6.12.47+rpt-rpi-2712 (16 KB pages)
- **Architecture:** aarch64
- **RAM:** 16 GB (14 GB free)
- **Node:** 20.19.2
- **GitNexus:** 1.6.2 and 1.6.4-rc.25 (both fail identically)

## Reproduction

Any moderately sized codebase. Crashes deterministically at ~64% during `Loading nodes` phase regardless of repo content. Crashes before any per-file scope-extraction warnings, so it's not file-specific.

## Suggested Fix

LadybugDB / its buffer manager needs to query `sysconf(_SC_PAGESIZE)` at runtime instead of assuming 4 KB. Alternatively, GitNexus can detect Pi 5 / 16 K pages and configure LadybugDB to use 16 K-aligned buffer sizes, or fall back to a non-mmap allocator.

Workaround for users: switch Pi 5 to a 4 KB kernel via `config.txt` (`kernel=kernel_2712.img` + reboot) — but that disables Pi 5 specific optimizations.

This is going to affect more users as ARM64 desktops/SBCs ship with 16 K pages by default.

vlisitskii:
## What

Addresses #1231 (ARM64 16K page size: COPY fails with "Releasing physical memory associated with a frame failed" on Raspberry Pi 5).

- **`lbug-config.ts`**: adds `isLbugPageSizeFrameError()` (matches the frame-release failure and the 0.18.0 residual `Unsupported page size combination` guard; deliberately does **not** match generic buffer-pool exhaustion), `isPageSizeAwareLadybug()` (version gate at 0.18.0), and `getOsPageSize()` (cached POSIX `getconf PAGE_SIZE` probe — Node exposes no page-size API).
- **`analyze.ts`**: new error branch (same pattern as the WAL-corruption / checkpoint-IO branches) that reports the detected OS page size and gives version-dependent guidance: upgrade to a release bundling `@ladybugdb/core >= 0.18.0` when the installed one is older, or a report-this-bug hint with the exact triage fields the maintainer asked for in #1231 when it is already >= 0.18.0.
- **`doctor.ts`**: prints the OS page size next to the LadybugDB version and warns on the non-4K-pages + `< 0.18.0` combination *before* the user runs analyze.
- **`cli-message.ts`**: adds the `lbug-page-size` recovery hint to the `RecoveryHint` union.

## Why — root cause findings

The crash in #1231 is an upstream LadybugDB bug that is **already fixed by the 0.18.0 pin (#2340)**, but nobody on a 16K-page machine has confirmed it on the issue, and users on older installed versions still get a raw truncated native error with zero guidance. Findings from investigating with the published npm binaries:

1. Disassembling `VMRegion::releaseFrame` in `@ladybugdb/core-linux-arm64` **0.17.1** shows frame eviction calls `madvise(frame, frameSize, MADV_DONTNEED)` on 4 KiB-aligned frames and throws `BufferManagerException` on any non-zero return. On a 16 KiB-page kernel (Raspberry Pi 5 default `kernel_2712.img`, Asahi Linux) most 4 KiB frame addresses are not page-aligned, so `madvise` returns `EINVAL` — exactly the reported `error code -1: Invalid argument`.
2. **0.18.0** rewrote that path: the binary gains runtime OS-page-size detection (`"Failed to detect the operating system page size."`) and discard-granule-aligned release with per-frame-group refcounting (`"Unsupported page size combination: frame size {}, discard granule size {}, frame group size {}."`). Versions 0.15.4 / 0.16.0 / 0.16.1 / 0.17.0 / 0.17.1 contain **none** of these strings — the reporter's gitnexus 1.6.2 / 1.6.4-rc.25 all shipped affected versions.
3. The residual 0.18.0 guard can still throw on exotic frame-size/page-size combinations, so the new diagnostics keep users on >= 0.18.0 covered too (they are pointed back to #1231 with the triage fields).

## How to verify

- `cd gitnexus && npx tsc --noEmit` — clean.
- `npx vitest run test/unit/lbug-config-pagesize.test.ts test/unit/analyze-pagesize-error.test.ts test/unit/lbug-config-wal.test.ts test/unit/analyze-wal-error.test.ts` — 52 tests pass (new matcher/probe/CLI-branch tests mirror the shape of the existing WAL-error tests).
- End-to-end on a real 16 KiB-page machine (darwin/arm64, `vm.pagesize` 16384): full self-analyze of this repo with the built CLI and `@ladybugdb/core` 0.18.0 succeeds (19,557 nodes / 52,650 edges), and `gitnexus doctor` prints `page size 16384` with no warning. With the fingerprint pinned to 0.17.1 the analyze branch renders the upgrade guidance (covered by unit test).

## Risk / rollback

Purely additive: one new early-return branch in the analyze error chain (placed after the checkpoint-IO branch, before the embedding branches), two doctor output lines, new exports in `lbug-config.ts`. No behavior change on the success path. Revert the single commit to roll back.

Closes #1231 pending the reporter's Pi 5 retest on a `>= 1.6.9` build — happy to adjust wording/placement per review.

### DIFF
--- a/gitnexus/src/cli/analyze.ts
+++ b/gitnexus/src/cli/analyze.ts
@@ -16,7 +16,10 @@ import cliProgress from 'cli-progress';
 import { isLbugReady, LbugWipeError } from '../core/lbug/lbug-adapter.js';
 import { boundedCheckpointBeforeExit } from '../core/lbug/shutdown-helpers.js';
 import {
+  getOsPageSize,
   isLbugCheckpointIoError,
+  isLbugPageSizeFrameError,
+  isPageSizeAwareLadybug,
   isWalCorruptionError,
   parseWalCheckpointThreshold,
   WAL_RECOVERY_SUGGESTION,
@@ -37,6 +40,7 @@ import {
   GitNexusRcError,
 } from './analyze-config.js';
 import { runFullAnalysis } from '../core/run-analyze.js';
+import { getRuntimeFingerprint } from '../core/platform/capabilities.js';
 import { getMaxFileSizeBannerMessage } from '../core/ingestion/utils/max-file-size.js';
 import { warnMissingOptionalGrammars, getOptionalGrammarExtensions } from './optional-grammars.js';
 import { glob } from 'glob';
@@ -1631,6 +1635,39 @@ const analyzeCommandImpl = async (
       cliError(`  ${msg.replace(/\n/g, '\n  ')}\n`, {
         recoveryHint: 'lbug-wipe-failed',
       });
+      process.exitCode = 1;
+      return;
+    }
+
+    // Buffer-manager frame-release failure on non-4K-page kernels (#1231).
+    // LadybugDB <= 0.17.x assumed 4 KiB OS pages when releasing evicted
+    // frames; Raspberry Pi 5 (16 KiB kernel pages) and other arm64 systems
+    // crash mid-COPY with a raw native message. 0.18.0 detects the page size
+    // at runtime, so the actionable fix depends on which side of that
+    // boundary the installed @ladybugdb/core is.
+    if (isLbugPageSizeFrameError(err)) {
+      const pageSize = getOsPageSize();
+      const ladybug = getRuntimeFingerprint().ladybugdb;
+      const pageLine =
+        pageSize !== undefined && pageSize !== 4096
+          ? `  Detected OS page size: ${pageSize} bytes (non-4K — e.g. Raspberry Pi 5 16K kernel, Asahi Linux).\n`
+          : '';
+      const guidance = isPageSizeAwareLadybug(ladybug)
+        ? `  The installed @ladybugdb/core (${ladybug}) already detects the OS page size at runtime,\n` +
+          `  so this configuration was expected to work. Please report it:\n` +
+          `    https://github.com/abhigyanpatwari/GitNexus/issues/1231\n` +
+          `  and include: gitnexus --version, node --version, getconf PAGE_SIZE, uname -a,\n` +
+          `  and the full error message above.\n`
+        : `  The installed @ladybugdb/core (${ladybug ?? 'unknown'}) assumes 4 KiB pages in its buffer\n` +
+          `  manager. Upgrade GitNexus to a release that bundles @ladybugdb/core >= 0.18.0\n` +
+          `  (gitnexus >= 1.6.9), which detects the OS page size at runtime:\n` +
+          `    npm install -g gitnexus@latest\n` +
+          `  Last-resort workaround on Raspberry Pi 5: boot the 4 KiB-page kernel\n` +
+          `  (config.txt: kernel=kernel8.img), at the cost of Pi 5 optimizations.\n`;
+      cliError(
+        `  LadybugDB's buffer manager failed to release frame memory.\n` + pageLine + guidance,
+        { recoveryHint: 'lbug-page-size', pageSize, ladybugVersion: ladybug },
+      );
       process.exitCode = 1;
       return;
     }
--- a/gitnexus/src/cli/analyze.ts
+++ b/gitnexus/src/cli/analyze.ts
@@ -1652,20 +1652,35 @@ const analyzeCommandImpl = async (
         pageSize !== undefined && pageSize !== 4096
           ? `  Detected OS page size: ${pageSize} bytes (non-4K — e.g. Raspberry Pi 5 16K kernel, Asahi Linux).\n`
           : '';
+      // The upgrade variant must not assert version facts about an unknown
+      // version — mirror the doctor-side wording rule (#2424 review R2).
+      const upgradeIntro =
+        ladybug === undefined
+          ? `  The installed @ladybugdb/core version is unknown — it may predate the\n` +
+            `  runtime OS-page-size detection added in 0.18.0.\n`
+          : `  The installed @ladybugdb/core (${ladybug}) assumes 4 KiB pages in its buffer\n` +
+            `  manager.\n`;
       const guidance = isPageSizeAwareLadybug(ladybug)
         ? `  The installed @ladybugdb/core (${ladybug}) already detects the OS page size at runtime,\n` +
           `  so this configuration was expected to work. Please report it:\n` +
           `    https://github.com/abhigyanpatwari/GitNexus/issues/1231\n` +
           `  and include: gitnexus --version, node --version, getconf PAGE_SIZE, uname -a,\n` +
           `  and the full error message above.\n`
-        : `  The installed @ladybugdb/core (${ladybug ?? 'unknown'}) assumes 4 KiB pages in its buffer\n` +
-          `  manager. Upgrade GitNexus to a release that bundles @ladybugdb/core >= 0.18.0\n` +
+        : upgradeIntro +
+          `  Upgrade GitNexus to a release that bundles @ladybugdb/core >= 0.18.0\n` +
           `  (gitnexus >= 1.6.9), which detects the OS page size at runtime:\n` +
           `    npm install -g gitnexus@latest\n` +
           `  Last-resort workaround on Raspberry Pi 5: boot the 4 KiB-page kernel\n` +
           `  (config.txt: kernel=kernel8.img), at the cost of Pi 5 optimizations.\n`;
+      // Embed the raw native text (indented, no stack) so "the full error
+      // message above" is fulfillable — same idiom as the LbugWipeError
+      // branch. The errno suffix and the 0.18.0 guard's frame/granule numbers
+      // are the discriminating triage content (#2424 review P2).
       cliError(
-        `  LadybugDB's buffer manager failed to release frame memory.\n` + pageLine + guidance,
+        `  LadybugDB's buffer manager failed to release frame memory.\n` +
+          `    ${msg.replace(/\n/g, '\n    ')}\n` +
+          pageLine +
+          guidance,
         { recoveryHint: 'lbug-page-size', pageSize, ladybugVersion: ladybug },
       );
       process.exitCode = 1;
--- a/gitnexus/src/cli/doctor.ts
+++ b/gitnexus/src/cli/doctor.ts
@@ -113,6 +113,39 @@ export function localEmbeddingDoctorStatus(opts: {
   return { status: '✓ local embeddings supported', detail: null };
 }
 
+/**
+ * Page-size lines for the `doctor` Runtime section (#1231). Pure so the
+ * warning gate can be unit-tested without running the whole command (the
+ * `localEmbeddingDoctorStatus` precedent above) — but takes the probed
+ * values as plain params rather than injectable probes, because `undefined`
+ * is a *meaningful* pageSize state here (probe unavailable / win32) and
+ * would collide with a "not provided → use default" DI convention.
+ *
+ * Returns 0 lines (page size unknown), 1 line (page size), or 2 lines
+ * (page size + non-4K warning when the installed @ladybugdb/core does not
+ * detect the OS page size at runtime).
+ */
+export function pageSizeDoctorLines(
+  pageSize: number | undefined,
+  ladybugVersion: string | undefined,
+): string[] {
+  if (pageSize === undefined) return [];
+  const lines = [`  ${padDisplayEnd('page size', 10)}${pageSize}`];
+  if (pageSize > 4096 && !isPageSizeAwareLadybug(ladybugVersion)) {
+    // Don't assert "< 0.18.0" as fact when the version is unresolvable
+    // (#2424 review R2) — name the unknown state instead.
+    const versionClause =
+      ladybugVersion === undefined
+        ? 'an unknown @ladybugdb/core version (may predate 0.18.0)'
+        : `@ladybugdb/core < 0.18.0`;
+    lines.push(
+      `  ${padDisplayEnd('', 10)}⚠ non-4K page size with ${versionClause} — ` +
+        `'gitnexus analyze' may fail during COPY (#1231). Upgrade gitnexus (npm install -g gitnexus@latest).`,
+    );
+  }
+  return lines;
+}
+
 export const doctorCommand = async () => {
   const fingerprint = getRuntimeFingerprint();
   const capabilities = getRuntimeCapabilities();
@@ -128,15 +161,8 @@ export const doctorCommand = async () => {
   // @ladybugdb/core < 0.18.0 assumed 4 KiB pages in its buffer manager and
   // crashes mid-COPY on 16 KiB/64 KiB-page kernels (#1231). Literal label
   // (like the 'native' line below) to avoid adding i18n keys.
-  const osPageSize = getOsPageSize();
-  if (osPageSize !== undefined) {
-    console.log(`  ${padDisplayEnd('page size', 10)}${osPageSize}`);
-    if (osPageSize > 4096 && !isPageSizeAwareLadybug(fingerprint.ladybugdb)) {
-      console.log(
-        `  ${padDisplayEnd('', 10)}⚠ non-4K page size with @ladybugdb/core < 0.18.0 — ` +
-          `'gitnexus analyze' may fail during COPY (#1231). Upgrade gitnexus (npm install -g gitnexus@latest).`,
-      );
-    }
+  for (const line of pageSizeDoctorLines(getOsPageSize(), fingerprint.ladybugdb)) {
+    console.log(line);
   }
   const nativeCheck = checkLbugNative();
   if (nativeCheck.ok) {
--- a/gitnexus/src/cli/doctor.ts
+++ b/gitnexus/src/cli/doctor.ts
@@ -13,6 +13,7 @@ import {
 } from '../core/embeddings/runtime-install.js';
 import { cudaRedirectDoctorStatus } from '../core/embeddings/onnxruntime-node-resolver.js';
 import { checkLbugNative, probeFtsExtensionLoad } from '../core/lbug/native-check.js';
+import { getOsPageSize, isPageSizeAwareLadybug } from '../core/lbug/lbug-config.js';
 import { diagnoseExtensionLoad } from '../core/lbug/extension-load-error.js';
 import { getExtensionInstallPolicy } from '../core/lbug/extension-loader.js';
 import { t } from './i18n/index.js';
@@ -123,6 +124,20 @@ export const doctorCommand = async () => {
   console.log(`  ${label('doctor.labels.node', 10)}${fingerprint.node}`);
   console.log(`  ${label('doctor.labels.gitnexus', 10)}${fingerprint.gitnexus}`);
   console.log(`  ${label('doctor.labels.ladybugdb', 10)}${fingerprint.ladybugdb ?? 'unknown'}`);
+  // OS page size next to the LadybugDB version because the two interact:
+  // @ladybugdb/core < 0.18.0 assumed 4 KiB pages in its buffer manager and
+  // crashes mid-COPY on 16 KiB/64 KiB-page kernels (#1231). Literal label
+  // (like the 'native' line below) to avoid adding i18n keys.
+  const osPageSize = getOsPageSize();
+  if (osPageSize !== undefined) {
+    console.log(`  ${padDisplayEnd('page size', 10)}${osPageSize}`);
+    if (osPageSize > 4096 && !isPageSizeAwareLadybug(fingerprint.ladybugdb)) {
+      console.log(
+        `  ${padDisplayEnd('', 10)}⚠ non-4K page size with @ladybugdb/core < 0.18.0 — ` +
+          `'gitnexus analyze' may fail during COPY (#1231). Upgrade gitnexus (npm install -g gitnexus@latest).`,
+      );
+    }
+  }
   const nativeCheck = checkLbugNative();
   if (nativeCheck.ok) {
     console.log(`  ${padDisplayEnd('native', 10)}✓ lbugjs.node loaded`);
--- a/gitnexus/src/core/lbug/lbug-config.ts
+++ b/gitnexus/src/core/lbug/lbug-config.ts
@@ -355,6 +355,97 @@ export const isLbugCheckpointIoError = (err: unknown): boolean => {
   return LBUG_CHECKPOINT_PERMISSIVE_RE.test(msg);
 };
 
+// ─── Ladybug non-4K page-size frame-release matcher (#1231) ─────────────────
+//
+// LadybugDB <= 0.17.x hardcoded a 4 KiB OS-page assumption in its buffer
+// manager: evicting a frame released physical memory with
+// `madvise(frame, frameSize, MADV_DONTNEED)` on 4 KiB-aligned frame
+// addresses (verified by disassembling `VMRegion::releaseFrame` in
+// @ladybugdb/core-linux-arm64 0.17.1 — `mov w2, #0x4` = MADV_DONTNEED,
+// throw on non-zero return). On kernels with 16 KiB pages (Raspberry Pi 5
+// default 2712 kernel, Asahi Linux) or 64 KiB pages (some enterprise arm64
+// distros), madvise rejects addresses that are not multiples of the real
+// page size with EINVAL, surfacing as:
+//   "Buffer manager exception: Releasing physical memory associated with a
+//    frame failed with error code -1: Invalid argument."
+// which aborts `gitnexus analyze` mid-COPY.
+//
+// @ladybugdb/core 0.18.0 rewrote the release path with runtime OS-page-size
+// detection and discard-granule-aligned madvise (new binary strings:
+// "Failed to detect the operating system page size.", "Unsupported page
+// size combination: frame size {}, discard granule size {}, frame group
+// size {}."), so upgrading is the fix. The residual 0.18.0 guard
+// ("Unsupported page size combination") is matched here too so exotic
+// configurations receive the same actionable guidance instead of a raw
+// native message.
+const LBUG_FRAME_RELEASE_RE = /releasing physical memory associated with a frame failed/i;
+const LBUG_PAGE_COMBO_RE = /unsupported page size combination/i;
+
+/**
+ * True when `err` looks like the LadybugDB buffer manager failing to release
+ * frame memory — the failure mode of a 4 KiB page-size assumption on a
+ * 16 KiB/64 KiB-page kernel (#1231). Deliberately does NOT match the
+ * generic "buffer pool is full" exhaustion error, which is a sizing
+ * problem, not a page-size one.
+ */
+export const isLbugPageSizeFrameError = (err: unknown): boolean => {
+  if (!err) return false;
+  const msg = err instanceof Error ? err.message : String(err);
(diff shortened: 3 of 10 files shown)


koriyoshi2041:
Pulled this branch locally and ran the focused verification from the PR body:

- `npx vitest run test/unit/lbug-config-pagesize.test.ts test/unit/analyze-pagesize-error.test.ts test/unit/lbug-config-wal.test.ts test/unit/analyze-wal-error.test.ts` -> 52 tests passed
- `npx tsc --noEmit` -> passed
- `git diff --check origin/main...HEAD` -> passed

The new page-size matcher also keeps the buffer-pool exhaustion case out of this recovery path, which is the boundary I was most interested in checking. I did not find a blocking issue in this pass.

gitnexus/src/cli/analyze.ts
```diff
@@ -1635,6 +1639,39 @@ const analyzeCommandImpl = async (
...
+      const pageLine =
+        pageSize !== undefined && pageSize !== 4096
+          ? `  Detected OS page size: ${pageSize} bytes (non-4K — e.g. Raspberry Pi 5 16K kernel, Asahi Linux).\n`
+          : '';
+      const guidance = isPageSizeAwareLadybug(ladybug)
+        ? `  The installed @ladybugdb/core (${ladybug}) already detects the OS page size at runtime,\n` +
+          `  so this configuration was expected to work. Please report it:\n` +
+          `    https://github.com/abhigyanpatwari/GitNexus/issues/1231\n` +
+          `  and include: gitnexus --version, node --version, getconf PAGE_SIZE, uname -a,\n` +
+          `  and the full error message above.\n`
```
magyargergo, line 1668: **[P2 — correctness]** The ≥0.18.0 bug-report path asks users to include "the full error message above", but this branch never prints the raw error: the `return` at line 1672 skips `writeFatalToStderr('Analysis failed', err)` (line 1780), the only place the native message/cause chain reaches stderr. The original text only ever hits the transient progress-bar frame (`run-analyze.ts:1111-1112` appends `detail` to the bar message — width-truncated and overwritten). Empirically confirmed by running this PR's own tests and grepping the captured output: the page-size branch emits only the guidance block. Because `package.json` pins `@ladybugdb/core ^0.18.0`, this report-a-bug path is the **primary** path on shipping builds — and the suppression also hides the discriminating detail: the errno suffix ("Invalid argument" vs "Cannot allocate memory") and, for the `LBUG_PAGE_COMBO_RE` case, the actual frame/discard-granule/frame-group sizes that are the entire diagnostic content of that error.

Suggested fix: surface `msg` inside the branch — e.g. prepend `` `  ${msg.replace(/\n/g, '\n  ')}\n` `` to the `cliError` body (the LbugWipeError branch at ~1635 embeds the message the same way), or call `writeFatalToStderr('Analysis failed', err)` before `cliError` for the aware branch (the AnalysisNotFinalizedError branch at ~1585 is the in-file model), and adjust the "above"/"below" wording to match. Add a test asserting the `PI5_COPY_ERROR` text appears in the captured records — the current tests assert hint presence and stack absence but never raw-message presence, exactly the gap that lets this pass green. While there, mocking `getOsPageSize()` to 16384 would also exercise the `Detected OS page size` line, which is currently dead in the suite on 4K CI boxes.

Strongest cross-engine finding of this review: Claude risk-architect + correctness + adversarial (all empirical) and all three Codex legs (swarm risk-architect — rated merge-blocking, ce-review, native adversarial) converged independently. [reproduced]

gitnexus/test/unit/lbug-config-pagesize.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    expect(isPageSizeAwareLadybug(undefined)).toBe(false);
+    expect(isPageSizeAwareLadybug('')).toBe(false);
+    expect(isPageSizeAwareLadybug('unknown')).toBe(false);
+    expect(isPageSizeAwareLadybug('v0.18.0')).toBe(false);
+  });
+});
+
+// ─── #1231: OS page-size probe ───────────────────────────────────────────────
+
+describe('getOsPageSize', () => {
```
magyargergo, line 93: **[P2 — CI coverage]** This file never runs on the Windows/macOS CI matrix: `.github/workflows/ci-tests.yml` runs windows-latest/macos-latest with only the fixed list from `gitnexus/scripts/cross-platform-tests.ts` ("Platform-sensitive subset only — the full suite runs on Ubuntu"), and this file isn't in it. TESTING.md says to add platform-sensitive tests to that script, and this file qualifies twice over: an explicit `process.platform === 'win32'` guard (the script's own PLATFORM_LOGIC category definition: "contain explicit process.platform guards") and a real `getconf` spawn. Two consequences: the win32 `toBeUndefined()` assertion is dead code in CI, and — the missed opportunity — macos-latest runners are arm64 with **real 16384-byte pages**, the exact hardware class #1231 targets; registering the file gives the probe genuine non-4K coverage instead of only Ubuntu's 4096.

Suggested fix: add `'test/unit/lbug-config-pagesize.test.ts'` to the `PLATFORM_LOGIC` array in `gitnexus/scripts/cross-platform-tests.ts` with a one-line rationale (precedent: `test/unit/platform-capabilities.test.ts` sits there for the same reasons). `analyze-pagesize-error.test.ts` does **not** need registration — its fingerprint is mocked and its outcome is platform-invariant.

Claude project-standards verified this against ci-tests.yml + TESTING.md + the script's six lists; Codex flagged the same block's host-dependence. [code-read]

magyargergo (commented):
## Tri-engine review digest

**Engines:** Claude ran the GitNexus PR swarm lanes (facts-historian, risk-architect, test-ci-verifier, security-boundary) plus six ce-code-review personas (correctness, adversarial, testing, maintainability, project-standards, reliability). Codex (GPT-5.5, xhigh reasoning) independently ran the same swarm in Solo mode and the ce-code-review contract, plus a native adversarial pass — all three Codex legs were live. Findings were merged with cross-engine corroboration weighting; everything below was re-verified against the PR head (b30c54c) in a local worktree. Coverage limits: none significant — 6-file, purely additive PR, fully read by every lane; all 52 new+neighbor tests run locally (pass), `tsc --noEmit` and prettier clean.

### What's solid

This is a well-built diagnostics PR and both engines said so before finding anything:

- Purely additive (+402/-0) and it follows the WAL-branch template faithfully — matcher-fn in `lbug-config.ts`, `cliError` + `recoveryHint` union procedure done exactly as `cli-message.ts`'s own doc comment prescribes, `_resetXForTest` and tri-state-cache conventions matching house precedent.
- The matcher negative tests are strong: buffer-pool exhaustion, the 8TB mmap failure (#785), WAL corruption, and lock contention are all proven non-matching, and re-verified at the `analyzeCommand` integration level.
- The hardcoded version claims check out end-to-end: `package.json` is 1.6.9 with `@ladybugdb/core ^0.18.0` (lockfile resolves 0.18.0), the #2340 pin is an ancestor of the v1.6.9 release commit, and `npm view gitnexus@1.6.9` agrees. Deliberately not anchoring the matcher on the strerror suffix ("Invalid argument") is the right locale-safety call per the #2385 precedent.
- Security-clean: static exec args, no shell, no ReDoS surface in either regex, hostile-getconf-output parse guard verified by execution, bidi/hidden-unicode scan clean.
- A large number of adversarial probes were constructed and **refuted** (see "Validated clean" below) — that's a feature of this review, and a credit to the code.

### Headline finding

**The bug-report path suppresses the very error it asks users to attach (P2, strongest cross-engine agreement).** All matched branches early-return before `writeFatalToStderr` (analyze.ts:1780), so the raw native message only ever reaches the transient progress-bar frame — width-truncated and overwritten. The ≥0.18.0 guidance ends with "and include: … the full error message above", but nothing above printed it. Since `package.json` pins `^0.18.0`, on shipping builds `isPageSizeAwareLadybug()` is true, making this report-a-bug path the **primary** path, not the edge case. The suppression also hides the discriminating content: the errno suffix (EINVAL vs ENOMEM) and, for the "Unsupported page size combination" 0.18.0-guard error, the actual frame/discard-granule sizes that *are* the diagnostic. Empirically confirmed by three Claude lanes (by running the PR's own tests and grepping captured output) and independently by all three Codex legs — Codex's swarm rated it merge-blocking. This fits the repo's recurring silent-error-surfacing theme (#2374, #1913, #2409). Concrete fix in the inline comment; it's small (embed `msg` in the cliError body or `writeFatalToStderr` first, plus one test assertion).

### Other findings (details in inline comments)

- **Cross-platform CI registration (P2):** `lbug-config-pagesize.test.ts` has an explicit `process.platform === 'win32'` guard and a real `getconf` spawn, but isn't registered in `gitnexus/scripts/cross-platform-tests.ts` — the windows/macos CI matrix runs *only* that fixed list (TESTING.md requires registration). Bonus: macos-latest arm64 runners have real 16 KiB pages — exactly the hardware #1231 is about — so registering buys genuine non-4K probe coverage for free.
- **doctor.ts additions have zero test coverage (P2):** the `page size` line and the ⚠ warning gate (`osPageSize > 4096 && !isPageSizeAwareLadybug(...)`) are asserted nowhere; the only `doctorCommand()` test is a doesn't-throw smoke test. Lanes ranged P1–P3 on this; merged at P2 (regression-protection gap on one of the PR's two advertised surfaces, not a live defect). Prior polarity-regression precedent: PR #2341.
- **Test hygiene in the getOsPageSize block (P2, cross-engine):** platform if/else inside the `it()` body (house rule; the WAL mirror files have none; repo idiom is `it.skipIf`/`it.runIf`), plus the test depends on the host's real `getconf` while the probe's failure paths (ENOENT/garbage/timeout → undefined) — the fail-safe that protects the analyze error path — have no coverage, and the "caches the probe" test is vacuous without a spy.
- **NODE_OPTIONS leak in analyze-pagesize-error.test.ts (P3):** `beforeEach` appends `--max-old-space-size=8192` with no restore; `analyze-worker-pool-size.test.ts` documents this exact duplicate-token failure mode and the two-line fix. (Downgraded from Codex's P2: per-file fork isolation contains the leak to this file; the WAL template file shares the same gap.)
- **getconf probe hardening (P3, reproduced):** the 5s timeout is not a hard bound — a SIGTERM-ignoring `getconf` blocked 9s in reproduction; `killSignal: 'SIGKILL'` makes it one. And `doctor` now forks getconf unconditionally on every POSIX run with a worst case of ~5s on the happy path; real getconf answers in ~2ms, so a much shorter timeout loses nothing.

### Lower-priority notes (no action required to merge)

- **Prerelease gate — genuine cross-engine disagreement, adjudicated:** Codex's adversarial leg argued treating `0.18.0-rc.x` as page-size-aware is a bug (semver-lower than 0.18.0); Claude's lanes treated it as acceptable, and the PR's own test pins `0.18.0-dev.20260708 → true` as intended. Practical impact is only *which guidance text* is shown, the shipping `^0.18.0` pin resolves stable releases, and #1231's reporter ran gitnexus 1.6.4-rc builds bundling 0.17.x (which reads 0.17.x and correctly gets the upgrade path). Verdict: fine as-is; a one-line comment documenting the intent would settle it permanently.
- The `Detected OS page size: N bytes` line (the PR's headline scenario) is dead in the test suite — on 4K CI boxes `pageLine` is always empty and no test mocks `getOsPageSize`; mutation-testing showed deleting the line passes all 22 tests. Cheap to pin alongside the headline fix.
- The structured fields (`recoveryHint: 'lbug-page-size'`, `pageSize`, `ladybugVersion`) are never asserted; five other `analyze-*-error.test.ts` files do assert `recoveryHint`.
- A third 0.18.0 page-size binary string, `"Failed to detect the operating system page size."`, is quoted in the PR's own comment block but not matched by either regex — a 0.18.0 user hitting it gets the generic stack dump instead of the #1231 guidance. If the exclusion is deliberate, a comment would help.
- Because `LBUG_FRAME_RELEASE_RE` matches any frame-release errno, a rare non-page-size madvise failure (e.g. ENOMEM) on a genuine 4K box takes this branch and gets a page-size root-cause claim; surfacing the raw message (headline fix) makes this self-diagnosing.
- Pre-existing-adjacent: a frame-release failure during *relationship* COPY never reaches this branch — rel-COPY warnings are truncated to 80 chars, the fallback swallows row errors, and every `loadGraphToLbug` call site discards the returned warnings; analyze exits 0 with missing edges and no #1231 guidance. Out of scope here (#2226-era design), but worth knowing the fix's coverage boundary.
- Maintainability (weak signals, judgment calls): the 0.18.0/#1231 guidance prose is duplicated between analyze.ts and doctor.ts with no shared constant (`WAL_RECOVERY_SUGGESTION` is the house pattern for exactly this); `getOsPageSize` is a pure OS fact living in the Ladybug-specific config module (`platform/capabilities.ts` is arguably its home); the hardcoded "gitnexus >= 1.6.9" floor in prose isn't tied to any source of truth; and doctor's warning asserts "< 0.18.0" even when the fingerprint is actually unresolvable (analyze handles that case with `?? 'unknown'`).

### Validated clean (probes constructed, then refuted — worth having on record)

- **No shadowing:** the WAL-corruption and checkpoint-IO regexes and all typed branches ahead of the new one cannot claim the frame-release message; ordering verified on this head (post-#2425 rebase). Re-verify if the chain changes again.
- **Matcher breadth:** the frame-release phrase is unique in the shipped 0.18.0 binary (`strings`-verified); wrapped COPY messages match (200-char slice can't cut the phrase).
- **Error-handler robustness:** `getOsPageSize`/`getRuntimeFingerprint` cannot throw in the handler — 9 hostile getconf variants (ENOENT, EPERM, garbage, multi-line, empty, non-zero exit, slow-timeout…) all executed → `undefined`, never a crash.
- **Cleanup parity:** the new branch inherits identical cleanup to all siblings (bar.stop, SIGINT restore, console restore happen once, before the chain); exit-code handling consistent.
- **Version-gate polarity:** correct for all realistic inputs; `'v0.18.0' → false` is unreachable-from-production conservatism (npm version fields never carry "v").
- **Cache tri-state:** `undefined`=unprobed / `null`=unavailable handled correctly; at most one fork per process; no test-order pollution (per-file forks + resets).
- **Standards:** conventional-commit title, no CHANGELOG edit, RecoveryHint union procedure followed, i18n literal-label precedent is real (matches the existing 'native'/'Ext install' lines), lbug-db vitest project routing correctly N/A (type-only import).

### CI status

The matrix has since completed: everything passes except `tests / windows-latest (platform-sensitive) 1/3` (and the downstream CI Gate) plus the Vercel fork-authorization check. The Windows failure is `test/integration/cli-e2e.test.ts:369` — the #1169-guard "analyze timed out before asserting finalization artifacts" assertion on the spawn-heavy shard (703s total) — a file this PR doesn't touch, in a known-flaky class (the shard-matrix exists because this suite crept past the Windows watchdog before, #2394-era); the new page-size tests don't even run in that matrix (see inline comment 2), and the PR's runtime changes are confined to the analyze *error* path and doctor. A re-run should clear it; worth confirming before merge. The Vercel failure is fork-authorization, not code.

### Verdict

**Final verdict: production-ready with minor follow-ups.** The change is purely additive, correctly placed in the error chain with no shadowing (verified post-#2425-rebase), security-clean, and its matcher/version-gate/probe core survived a large adversarial battery on both engines. The follow-ups are real but small and should land in this PR: the headline P2 (surface the raw native error the report path asks for — Codex's swarm rated this one merge-blocking, and we split the difference: the feature works, but the report-a-bug path is the primary path on shipping `^0.18.0` builds, so fix before merge), the cross-platform test registration, doctor coverage, and the test-hygiene items. None of them changes runtime behavior on the success path.

**Branch hygiene: clean feature/fix PR.** One commit, six causally related files, no unrelated churn, rebased onto the current base (the merge-base is #2425's merge commit).

**Merge state: checks failing** — but the only code-check failure is the pre-existing Windows e2e flake described above (re-run to confirm), plus the expected Vercel fork-auth block; `mergeable: MERGEABLE`, no conflicts.

---
*This is an automated multi-tool review digest (Claude swarm + ce personas, Codex GPT-5.5 swarm/ce/adversarial). Findings were cross-verified against the PR head, but please verify before acting on any individual item.*

gitnexus/src/cli/doctor.ts
```diff
@@ -123,6 +124,20 @@ export const doctorCommand = async () => {
...
   console.log(`  ${label('doctor.labels.gitnexus', 10)}${fingerprint.gitnexus}`);
   console.log(`  ${label('doctor.labels.ladybugdb', 10)}${fingerprint.ladybugdb ?? 'unknown'}`);
+  // OS page size next to the LadybugDB version because the two interact:
+  // @ladybugdb/core < 0.18.0 assumed 4 KiB pages in its buffer manager and
+  // crashes mid-COPY on 16 KiB/64 KiB-page kernels (#1231). Literal label
+  // (like the 'native' line below) to avoid adding i18n keys.
+  const osPageSize = getOsPageSize();
+  if (osPageSize !== undefined) {
+    console.log(`  ${padDisplayEnd('page size', 10)}${osPageSize}`);
+    if (osPageSize > 4096 && !isPageSizeAwareLadybug(fingerprint.ladybugdb)) {
```
magyargergo: **[P2 — test coverage]** The doctor half of this PR — the `page size` line and this ⚠ warning gate — has no test coverage. The only `doctorCommand()` test (`doctor-format.test.ts:149-159`) asserts it doesn't throw and never captures output, so nothing catches this two-clause boolean gate regressing (flipped comparison or polarity — this repo has a prior polarity-regression precedent, PR #2341). Suggested: assert the matrix — warning present for (16384, '0.17.1'); absent for (16384, '0.18.0'), (4096, '0.17.1'), and `osPageSize === undefined` — either by mocking `lbug-config.js`/`capabilities.js` the way `analyze-pagesize-error.test.ts` already does, or by extracting an injectable helper the way `localEmbeddingDoctorStatus` (this file's own tested precedent) does it. Nit while there: when the fingerprint is unresolvable, the warning still asserts "@ladybugdb/core < 0.18.0" — a comparison it never actually determined; analyze handles the same case with `${ladybug ?? 'unknown'}`.

Flagged by five Claude lanes + Codex's swarm test-ci lane; severity ranged P1 (one lane called it merge-blocking) to P3, merged at P2 — it's a regression-protection gap on one of the PR's two advertised surfaces, not a live defect. [code-read]

gitnexus/test/unit/lbug-config-pagesize.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  afterEach(() => {
+    _resetOsPageSizeCacheForTest();
+  });
+
+  it('returns a positive power-of-two page size on POSIX platforms', () => {
+    const pageSize = getOsPageSize();
+    if (process.platform === 'win32') {
```
magyargergo: **[P2 — test hygiene]** Two related issues in this block. (a) Platform if/else inside the `it()` body — the maintainer house rule is no if-branching in vitests, the WAL "mirror" neighbors this PR cites contain none, and the dominant in-repo idiom is `it.skipIf`/`it.runIf` (~100 usages, e.g. `hooks.test.ts`, `hook-db-lock-probe.test.ts`); on every current CI runner only the `else` branch executes. (b) The test calls the real `getOsPageSize()` against the host's `getconf` and expects a positive value on every non-Windows host — but the production code deliberately tolerates missing/sandboxed `getconf` (returns `undefined`), so this can fail for environment reasons rather than code regressions on constrained POSIX hosts. Meanwhile the probe's failure paths (execFileSync throw, malformed stdout, win32 short-circuit → `undefined` — the fail-safe that keeps the analyze error path from crashing) have zero unit coverage, and the "caches the probe" test at lines 92-94 is vacuous: real getconf is deterministic, so `expect(getOsPageSize()).toBe(getOsPageSize())` passes even with the cache deleted.

Suggested fix: split into `it.skipIf(process.platform === 'win32')` / `it.runIf(...)` tests with unconditional assertions (also removes the two `as number` casts), and mock `child_process.execFileSync` to drive throw / garbage-stdout / win32 deterministically, plus `expect(execFileSync).toHaveBeenCalledTimes(1)` to make the cache test real (the repo already mocks execFileSync this way, e.g. `hook-db-lock-probe.test.ts`, `git.test.ts`).

Cross-engine: Claude facts-historian, risk-architect, test-ci-verifier, testing, project-standards + Codex ce (testing + project-standards) and Codex swarm lane 4. [code-read]

gitnexus/test/unit/analyze-pagesize-error.test.ts
```diff
@@ -0,0 +1,163 @@
...
+const PI5_COPY_ERROR =
+  'COPY failed for File: Buffer manager exception: Releasing physical memory ' +
+  'associated with a frame failed with error code -1: Invalid argument.';
+
+describe('analyzeCommand non-4K page-size error handling (#1231)', () => {
+  beforeEach(() => {
+    vi.resetModules();
+    runFullAnalysisMock.mockReset();
+    process.exitCode = undefined;
+    process.env.NODE_OPTIONS = `${process.env.NODE_OPTIONS ?? ''} --max-old-space-size=8192`.trim();
```
magyargergo, line 99: **[P3 — test isolation]** `beforeEach` appends `--max-old-space-size=8192` to `process.env.NODE_OPTIONS` and nothing restores it, so within this file each test appends another copy (three duplicated tokens by the last test). The repo already documents this exact failure mode and its fix: `analyze-worker-pool-size.test.ts:33-51` captures `ORIGINAL_NODE_OPTIONS` before the suite and restores (or deletes) it in `afterEach`, with a comment explaining why accumulation matters. To be fair: the WAL template this file mirrors (`analyze-wal-error.test.ts:63`) has the same gap, and vitest's per-file fork isolation stops the leak at the file boundary — so this merges at P3 (Codex rated it P2 on a cross-file-leak theory that doesn't hold under `pool: 'forks'` with default isolation). Still worth the two-line fix while the file is fresh; consider giving `analyze-wal-error.test.ts` the same treatment.

Found by Codex ce (testing + adversarial personas); mechanism re-verified in the worktree by synthesis, blast radius corrected. [code-read]

gitnexus/src/core/lbug/lbug-config.ts
```diff
@@ -355,6 +355,97 @@ export const isLbugCheckpointIoError = (err: unknown): boolean => {
...
+export const getOsPageSize = (): number | undefined => {
+  if (cachedOsPageSize !== undefined) return cachedOsPageSize ?? undefined;
+  if (process.platform === 'win32') {
+    // Windows allocation granularity is not what madvise alignment is about;
+    // the #1231 failure mode is POSIX-only.
+    cachedOsPageSize = null;
+    return undefined;
+  }
+  try {
+    const out = execFileSync('getconf', ['PAGE_SIZE'], {
```
magyargergo, line 440: **[P3 — hardening]** Two bounded-latency notes on this probe, both empirically driven. (a) The 5s `timeout` is not a hard bound: `spawnSync` sends the default `killSignal` (SIGTERM) once at timeout, and a `getconf` that ignores it kept the sync call blocked for 9s in reproduction (`trap '' TERM`); a D-state child (hung network mount) behaves the same. Adding `killSignal: 'SIGKILL'` to the options makes 5s a real ceiling for everything except true D-state — a one-line change. (b) `doctor` now runs this fork unconditionally on every POSIX invocation (doctor.ts:131), so the ~5s worst case lands on the doctor *happy path* (reproduced with a stalling stub: `elapsed ms 5005`); a real `getconf` answers in ~2ms, so a much shorter timeout (500ms–1s) loses nothing — the 5s value was sized for the slower Windows cmd.exe 8.3 probe this code cites as its model. The analyze-side call is fine as-is: it only runs after the run has already failed.

Claude adversarial (SIGTERM escape, reproduced) + Claude reliability (5s doctor stall, reproduced); both rated hardening, not blocking. For the record, the sibling attacks all failed: ENOENT, EPERM, garbage output, multi-line output, empty, non-zero exit, and plain timeout are all swallowed → `undefined`. [reproduced]

magyargergo:
@vlisitskii Can you please test it on you arm cpu? 🙏

vlisitskii:
@magyargergo Tested on ARM as requested — two legs, both at the current head (`d6ff5ad`, including all your review follow-ups). Thanks for landing those!

**Leg 1 — darwin/arm64, 16 KiB pages (Apple Silicon, `vm.pagesize` = 16384, Node v22.21.1)** — the non-4K hardware class this PR is about:

| Check | Result |
|---|---|
| `npx tsc --noEmit` | ✅ clean |
| Targeted suites (`lbug-config-pagesize`, `analyze-pagesize-error`, `lbug-config-wal`, `analyze-wal-error`, `doctor-format`) | ✅ 78 passed, 1 skipped (the win32-only guard) |
| `npm run test:cross-platform` (the matrix the probe test is now registered in) | ✅ 1652 passed, 31 skipped, 2 failed — both failures are `hooks-e2e.test.ts` "prefers pnpm dlx" assertions that reproduce **identically on current main**, so pre-existing and unrelated |
| `gitnexus doctor` (built dist) | ✅ prints `page size 16384` next to `LadybugDB: 0.18.0`, no ⚠ (correct — 0.18.0 is page-size aware) |
| Full `gitnexus analyze . --force` of this repo (built dist) | ✅ 19,671 nodes / 52,951 edges in ~40s — bulk COPY completes on real 16 KiB pages with `@ladybugdb/core` 0.18.0, no buffer-manager errors |

**Leg 2 — linux/arm64 (aarch64) container, 4 KiB pages, `@ladybugdb/core` 0.18.0 native verified loading:**

| Check | Result |
|---|---|
| Same 5 targeted suites | ✅ 78 passed, 1 skipped |

One honest caveat: I don't have a 16 KiB-page *Linux* kernel to hand (Raspberry Pi 5 / Asahi) — Docker's VM kernel is 4K, and macOS's Mach VM tolerates the sub-page `madvise` pattern that Linux rejects with EINVAL, which is why the crash was never reproducible on Macs on ≤ 0.17.x either. So the closest-to-Pi evidence remains the binary analysis in the PR body (0.17.1's `releaseFrame` hardcodes 4 KiB `madvise`; 0.18.0 adds runtime page-size detection) plus the 16 KiB-macOS end-to-end run above. A retest from @evefall's Pi 5 on a ≥ 1.6.9 build would close the loop completely.

Issue #1300, opened 2026-05-03
Corrupted wal file. Read out invalid WAL record type


herosimith:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

After building, when using Skills, the error "Corrupted WAL file. Read out invalid WAL record type" frequently occurs. There's no solution other than force deletion and rebuilding.

### Expected behavior

After building, when using Skills, the error "Corrupted WAL file. Read out invalid WAL record type" frequently occurs. There's no solution other than force deletion and rebuilding.

### Actual behavior

After building, when using Skills, the error "Corrupted WAL file. Read out invalid WAL record type" frequently occurs. There's no solution other than force deletion and rebuilding.

### Steps to reproduce

After building, when using Skills, the error "Corrupted WAL file. Read out invalid WAL record type" frequently occurs. There's no solution other than force deletion and rebuilding.

magyargergo:
Please try the latest rc version

magyargergo:
This should be fixed since `v.1.6.4-rc.66`

PR #2425, opened 2026-07-10
fix: make large incremental writebacks commit reliably (#2409)

### LINKED ISSUE #2409: Incremental analyze silently fails to commit (commitOK=False) on every large BFS importer-set expansion (>=~150 importers) — 100% reproducible on gitnexus 1.6.8

magyargergo:
Fixes #2409 — the root-cause follow-up to #2410 (which surfaced the diagnostics; this PR fixes what they diagnose).

## Root cause

**Defect 1 — the writeback itself.** The incremental path deleted rows with a count + `DETACH DELETE` per node table per **file** — ~13,000 single-row write transactions for a ~700-file write set, every one a WAL append interleaved with checkpoint ticks. Reproduced at the issue's exact profile (4,515 files / 38.8k symbols, `changed=12, added=13`, +690 importer expansion): the delete phase alone took **37s of a 63s run** — the "fast path" was 2.7× slower than a full rebuild — and that txn/WAL storm is the write pattern behind the native mid-writeback deaths on Windows. The loop also swallowed **all** errors per file, which is why `commitOK=False` never came with a reason (the issue's ask #1).

**Defect 2 — recovery death spiral.** On the next run, the dirty-flag recovery forces a full rebuild — but the embedding-cache preservation step opens the DB **before** the rebuild wipe (`run-analyze.ts` open at the cache-load step vs. wipe far below it). That open replays the crashed writeback's poisoned WAL and dies natively, so recovery never runs. Manually renaming the index dir aside worked precisely because it avoided opening those files. This also answers the issue's ask #3: it's not the #1300/#1611 corrupt-WAL class — it's a replay-of-poisoned-WAL-before-wipe ordering bug.

**Bonus latent bug.** LadybugDB's parser **rejects** SQL-style `''` quote doubling (`'we''ird'` is a parser error, not an escape) — the old delete/importer queries used it and swallowed the failure, so files with quotes in their paths were silently never deleted or BFS-expanded.

## The fix

| Change | Effect |
|---|---|
| `deleteNodesForFiles` — chunked `WHERE n.filePath IN […]` batch deletes (200/chunk), **no error swallowing** | delete phase 37s → 7s; ~40 statements instead of ~13,000; real engine failures now propagate to the CLI error path |
| `queryImportersBatch` — importer BFS batched per depth | BFS 5.6s → 0.9s; ~8 queries instead of ~1,400 |
| Escalation valve — effective write set > 50% of files (and ≥ 50 files) → switch to the proven wipe + bulk-COPY full write plan (the pipeline already built the full graph) | 100%-write-set case 38s → 7.7s; the surgical path no longer runs at the scale where it failed |
| `quarantineSidecarsForDirtyRecovery` — park `lbug.wal`/`lbug.shadow` (rename, never delete; fixed names cap accumulation at one file each) **before any open** on dirty-flag recovery | recovery rebuild can no longer be killed by the crashed run's WAL |
| Full-rebuild wipe list now includes `lbug.shadow` | a stale shadow next to a fresh DB file was replay poison |
| `checkpointOnce()` boundary drain after the incremental writeback | bounds the WAL any later crash can leave behind |
| `escapeCypherString` (backslash escaping) on the writeback path | quoted paths delete/expand correctly |

End-to-end at the report's scale: incremental **63s → 31s** (surgical) and **38s → 7.7s** (escalated), committing cleanly (`incrementalInProgress` cleared, `lastCommit == HEAD`) in both cases.

## Correctness

- The load-bearing **incremental ≡ `--force` stats-equality** invariant tests pass unchanged, and the new escalation test asserts the same equality for the escalated plan (any stale row surviving the wipe would show as a stats delta).
- New integration test for `deleteNodesForFiles`: exact-set deletion across a >1-chunk write set, DETACH semantics at the kept/deleted boundary, quote-in-path escaping, zero-match no-op, cumulative progress callbacks.
- New orchestration tests: hub-edit escalation end-to-end; dirty-flag recovery parks garbage `wal`/`shadow` bytes verbatim before reopening and completes the rebuild.
- New sidecar unit tests: parks both sidecars, no-op when absent, stale parked copy replaced.
- Suites run locally: incremental-orchestration 11/11, sidecar-recovery + delete-batch 47/47, adapter/conn-serialization/mock-dependent 56/56, analyze-adjacent sweep 106/106. `tsc --noEmit` and prettier clean.

## Follow-ups — addressed in-PR (commit 938a6169)

- **All remaining `''` escaping sites fixed.** The invalid doubling also lived in `augmentation/engine.ts` (3 sites), `embedding-pipeline.ts` (1), and `wiki/graph-queries.ts` (5) — nine queries that never parsed, silently. All now share `escapeCypherString` (`src/core/lbug/cypher-escape.ts`), unit-tested; the `csv-generator.ts` `''` usages are CSV-reader grammar, not Cypher, and are documented as deliberately untouched on the helper.
- **Windows CI now exercises the #2409 paths directly.** The dev environment is Linux/WSL2, so native-Windows validation is routed through the `windows-latest` cross-platform matrix: `lbug-delete-nodes-for-files.test.ts` (chunked IN-list DETACH DELETE + escaping against the real native addon) joins `LBUG_NATIVE`, and the dirty-recovery parking test is split into `incremental-dirty-recovery.test.ts` so the matrix runs it without the full orchestration suite. `sidecar-recovery.test.ts` was already in the matrix.
- The Windows-native death itself was never reproducible on Linux; this PR removes the write pattern that triggers it, makes any residual failure loud instead of silent, and makes recovery survive it. If a Windows repro still dies mid-writeback after this, the parked `*.dirty-recovery` sidecars are exactly the post-mortem artifact to attach.
- `deleteNodesForFile` (singular) is no longer used by the orchestrator but keeps its per-query-connection API and tests.


gitnexus/src/core/run-analyze.ts
```diff
@@ -1305,66 +1327,128 @@ export async function runFullAnalysis(
...
+            await fs.rm(f, { recursive: true, force: true });
+          } catch {
+            /* swallow */
+          }
         }
-        if (i % 20 === 0) {
-          progress('lbug', 62, `Removing rows for changed files (${i}/${filesToDelete.length})...`);
+        await initLbug(lbugPath);
+        walCheckpointDriver = startWalCheckpointDriver();
+        await loadGraphToLbug(pipelineResult.graph, pipelineResult.repoPath, storagePath, (msg) => {
```
magyargergo, line 1462: **P1 — the escalation valve wipes the HNSW vector index and nothing on the preserve-only path recreates it; meta stamps `vector-index` anyway.**

The wipe above (1365-1371) destroys the index with the DB file. An escalated run is by definition `force=false`, so `deriveEmbeddingMode` (embedding-mode.ts:74-78) yields `shouldGenerateEmbeddings=false` for any previously-embedded repo; Phase 3.5 restores embedding **rows** only, and `createVectorIndex` is reachable solely via `buildVectorIndex` inside the skipped `runEmbeddingPipeline` (embedding-pipeline.ts:228/237, callers :401/:568 — grep-verified sole path). Meta then falls back to `runtimeCapabilities.semanticMode` (run-analyze.ts:1659-1661) and stamps `vectorSearch.status: 'vector-index'` because the restored rows make `embeddingCount > 0` (:1705-1707).

**Trigger:** embedded repo (>10k CodeEmbedding rows for the hard variant) on macOS/Linux, routine `gitnexus analyze` after a hub edit that trips this valve — the exact #2409 scenario. `semanticSearch`'s `QUERY_VECTOR_INDEX` throw is swallowed (embedding-pipeline.ts:653-655) and the exact-scan fallback caps at 10k rows (capabilities default), so semantic search silently returns `[]` forever until `--force`/`--embeddings`. Plain `--force` is safe (it sets `forceRegenerateEmbeddings` → pipeline runs → index rebuilt), so this breaks the valve's own "indistinguishable from a --force rebuild" invariant. The new escalation test can't see it: its fixture has no embeddings and the stats-parity block omits `embeddings`/index presence.

**Fix:** after the escalated load + Phase 3.5 restore, recreate the index when cached rows were re-inserted (policy-gated `createVectorIndex`, warn on failure like the pipeline does) — or treat escalation as `--force` for embedding-mode purposes, or at minimum stamp `exact-scan`/`unavailable` when the wipe ran without the pipeline. Extend the escalation test with an embedded fixture asserting index presence (`SHOW_INDEXES` / a `semanticSearch` round-trip) and `stats.embeddings` parity.

[code-read] — chain verified link-by-link by two independent Claude passes (correctness + adversarial), the coordinator, and synthesis. Introduced by this PR (first `force=false` wipe on the mainstream incremental flow).

gitnexus/src/core/lbug/lbug-adapter.ts
```diff
@@ -2045,6 +2046,62 @@ export const deleteNodesForFile = async (
...
+      // Community/Process are graph-wide (no filePath); the orchestrator
+      // drops them wholesale via deleteAllCommunitiesAndProcesses.
+      if (tableName === 'Community' || tableName === 'Process') continue;
+      const tn = escapeTableName(tableName);
+      await queryAndDrain(
+        targetConn,
+        `MATCH (n:${tn}) WHERE n.filePath IN ${listLiteral} DETACH DELETE n`,
+      );
+    }
+    // Embedding rows key on nodeId ('<filePath>:…'), so IN can't apply —
```
magyargergo: **P2 — this comment is factually wrong and the embedding delete below is a provable no-op.** Embedding `nodeId` is the graph node id, which is **label-first** (`generateId` → `` `${label}:${name}` ``, src/lib/utils.ts:1-3; `nodeId: node.id` at embedding-pipeline.ts:501; this PR's own fixture uses `Function:${fp}:fn${i}:1`), so `e.nodeId STARTS WITH '<filePath>'` never matches anything.

Reproduced against the real indexed DB: `MATCH (e:CodeEmbedding) RETURN e.nodeId LIMIT 5` → `Function:docker-server.mjs:isValidUrl`, … — no bare-path prefixes exist. The old per-file variant (~line 2036) has the identical bug, so the batched form faithfully mirrors a pre-existing no-op rather than regressing — but this PR adds a comment enshrining the wrong id format.

**Impact:** deleted files' CodeEmbedding rows linger forever on the surgical path (table bloat, inflated `embeddingCount` — which also feeds the 10k exact-scan gate — ghost KNN candidates dropped only at the node join). Correctness holds today via downstream guards (hash-stale deletes use exact ids; `semanticSearch` filters dead nodes). The new integration test can't detect it: `withTestLbugDB` wipes the embedding table and nothing repopulates it.

**Fix (small):** delete by join before the node DETACH DELETE (`MATCH (n:<T>) WHERE n.filePath IN […] MATCH (e:CodeEmbedding) WHERE e.nodeId = n.id DELETE e`), or build prefixes as `<Label>:<filePath>:` per embeddable label — at minimum correct this comment so the false invariant doesn't propagate. And seed CodeEmbedding rows (deleted + survivor + quoted path) in lbug-delete-nodes-for-files.test.ts.

[reproduced] — live-DB query by the coordinator; independently found by Claude security + correctness lanes and Codex (GPT-5.5) correctness at confidence 100.

gitnexus/src/core/lbug/lbug-adapter.ts
```diff
@@ -6,6 +6,7 @@ import { finished } from 'stream/promises';
 import path from 'path';
 import lbug from '@ladybugdb/core';
 import { closeQueryResults } from './query-result-utils.js';
+import { escapeCypherString } from './cypher-escape.js';
```
magyargergo, line 9: **P2 — the escaping sweep is incomplete in this same file:** three Cypher builders still use the SQL-style `''` doubling that cypher-escape.ts's own docstring says LadybugDB **rejects** (parser error — premise confirmed live against the real engine during this review):

- `fallbackRelationshipInserts`'s `esc` (lines 1288-1289) — quotes doubled *before* backslashes, so doubly wrong — and **production-reachable**: called from `loadGraphToLbug` (line 1175) whenever a relationship COPY fails for missing schema pairs, where the per-row `catch { // skip }` (1298-1300) silently drops every relationship whose id/type/reason contains a quote. Node ids embed file paths — this is exactly the silent-loss class #2409 is about.
- `insertNodeToLbug`'s `escapeValue` (line 1378) and `batchInsertNodesToLbug`'s (line 1455) — exported adapter API; currently test-only callers, same latent defect.

All three are pre-existing lines (not in this diff), but commit 938a6169 claims "apply valid Cypher escaping everywhere", and this PR's shared helper is already imported here (this line).

**Fix (cheap):** route the string branch of all three through `escapeCypherString` (keep the NULL/number branches; the `\n`/`\r` escapes can stay or go — the adversarial probe confirmed raw LF is legal inside single-quoted literals). Add one quoted-value regression test through the fallback-insert path (currently zero coverage — no test references `fallbackRelationshipInserts`).

[code-read] — coordinator-verified in the worktree; converged independently by Codex (GPT-5.5) swarm (its single merge-gating finding) + Codex maintainability (confidence 100) and Claude facts/security/maintainability lanes.

gitnexus/src/core/lbug/sidecar-recovery.ts
```diff
@@ -429,6 +429,59 @@ export async function finalizeLbugSidecarsAfterClose(
...
+    try {
+      if (!(await statIfExists(from))) continue;
+      const to = `${from}.dirty-recovery`;
+      // Windows rename-onto-existing fails — drop a stale parked copy first.
+      await fs.rm(to, { force: true });
+      await fs.rename(from, to);
+      moved.push(to);
+    } catch (err) {
+      if (!missing(err)) {
+        log(
```
magyargergo, line 629: **P2 — on the platform this fix targets, a failed park rename falls through to the exact pre-wipe open the parking exists to protect.** If this rename fails EBUSY/EPERM (Windows AV, or an MCP/serve process holding `lbug.wal` — the lock class this repo's own checkpoint-driver docs describe), the poisoned WAL stays in place, and the recovery analyze's embedding-cache preservation open (`initLbug` at run-analyze.ts:1015, reached for any repo with `stats.embeddings > 0`) replays it **before** the rebuild wipe. If the WAL is genuinely poisoned, that open dies natively (a JS try/catch can't intercept it — the PR's own reasoning), so the #2409 death loop persists on this branch for as long as the external lock does. The warning text also over-promises: "the rebuild will wipe it in place instead" — if the pre-wipe open crashes, no wipe ever happens.

Both engines converged on the mechanism (Codex adversarial: high; Codex reliability: P1; Claude reliability: P1; Claude risk: P2-by-design). Kept at **P2**: pre-PR there was *no* parking at all, so this is a residual gap in a strict improvement; it self-heals once the lock clears; and the docstring honestly declares best-effort.

**Fix (cheap hardening):** when a sidecar fails to park, skip the embedding-cache-preservation open for that run (regenerate instead of preserve), or fall back to `fs.rm(from)` — the rebuild discards those bytes regardless; and make the log accurate (reuse `renameFailureMessage`'s EPERM/EBUSY guidance — stop the MCP server / AV exclusion / retry). Two sharp sub-edges: a successful `rm(to)` followed by a failed rename destroys the *previous* crash's parked post-mortem copy while keeping the live poison; and `fs.rm(force:true)` doesn't clear a Windows read-only attribute. No test covers this branch — the three new quarantine tests are all happy-path (a `vi.spyOn(fs,'rename')` EBUSY case would pin it).

[code-read] — ordering verified in worktree (catch 467-474 → run-analyze.ts:784 caller → :1015 open → wipe later).

gitnexus/src/core/run-analyze.ts
```diff
@@ -1305,66 +1327,128 @@ export async function runFullAnalysis(
...
+        });
+        // Strategy switch: stop the checkpoint driver around the close so its
+        // in-flight CHECKPOINT can't race the reopen, drop the DB files
+        // (sidecars included), and bulk-load the full graph into a fresh DB —
+        // byte-for-byte the full-rebuild write plan.
+        await walCheckpointDriver.stop();
+        await closeLbug();
+        for (const f of [lbugPath, `${lbugPath}.wal`, `${lbugPath}.shadow`, `${lbugPath}.lock`]) {
+          try {
+            await fs.rm(f, { recursive: true, force: true });
```
magyargergo: **P2 — a swallowed wipe failure here leads to silently certified corruption.** If this `rm` fails with the process alive (a Windows handle without FILE_SHARE_DELETE — the same lock class this PR blames for #2409), the `catch { /* swallow */ }` hides it, `initLbug` (1372) reopens the **old, fully-populated** DB, and the full-graph COPY runs against non-empty tables.

The consequence was **empirically proven against the real `@ladybugdb/core` engine** by the adversarial pass: a duplicate-PK node COPY fails, `copyCsvWithRetry`'s IGNORE_ERRORS retry then "succeeds" while keeping the stale row's **old** values (new content dropped), and the PK-less `CodeRelation` COPY **duplicates every relationship** (count 2 → 4 in the probe). No error surfaces; FTS rebuilds over stale content; meta saves clean with fresh fileHashes, so subsequent runs see "unchanged" — permanent until a manual `--force`.

The identical swallow pre-exists in the full-rebuild wipe (1162-1169), so the hazard class isn't new — but the valve routes **routine incremental runs** through it for the first time. Trigger is environmental (rm fails yet the reopen succeeds), hence P2 rather than P1.

**Fix:** after the rm loop, verify `lbugPath` is actually gone (`fs.access` → throw a loud, actionable error) before `initLbug` — the dirty flag is already durably saved (1352-1358), so failing here recovers cleanly next run. Same guard belongs on the full-rebuild wipe; while there, consider extracting the now-duplicated 4-file wipe list into one helper — list drift between these two copies is exactly the `.shadow` omission that was one of #2409's root defects.

[reproduced] — COPY/IGNORE_ERRORS semantics proven live (Claude adversarial); independently flagged P1 by Codex (GPT-5.5) reliability.

gitnexus/src/core/lbug/lbug-adapter.ts
```diff
@@ -2096,6 +2153,56 @@ export const queryImporters = async (targetFilePath: string): Promise<string[]>
...
+      try {
+        queryResult = await c.query(cypher);
+        const result = Array.isArray(queryResult) ? queryResult[0] : queryResult;
+        const rows = await result.getAll();
+        for (const row of rows) {
+          const v = (row as { importer?: unknown }).importer;
+          if (typeof v === 'string' && v.length > 0) importers.add(v);
+        }
+      } catch {
+        /* mirror queryImporters: degrade to a smaller expansion, stay writable */
```
magyargergo: **P2 — this catch silently drops up to 200 files' importer expansion per failed chunk, with zero logging.** Verified: no `logger.*`/`log()` call on this path. That is inconsistent with this same PR's rationale for `deleteNodesForFiles` ("silently skipping was exactly how #2409 hid its root cause") and with DoD 2.8 ("no silent catches that swallow diagnostics").

**Consequence chain** (Codex reliability): a swallowed chunk under-expands `writableFiles`/`effectiveWriteSet`, files that import a changed file keep their stale CALLS/IMPORTS edges, then the run exits 0, clears the dirty flag, and saves fresh fileHashes — the next run's hash diff sees "unchanged", so the staleness is permanent until `--force`. The old per-file `queryImporters` had the same swallow shape, but at 1-file granularity; a systematic engine failure now collapses the whole BFS to zero expansion with the only tell being an absent progress line. Silent-skip is this repo's documented recurring failure mode (#1913 / #2374 / #2396).

**Fix (minimal, keeps the deliberate degrade-don't-fail contract):** one `log`/`logger.warn` per failed chunk (depth, chunk index, first error message), optionally a dropped-chunk counter in the `incrementalInProgress` diagnostics that #2410 established. Fail-closed (propagate → dirty flag → full rebuild) is the stronger alternative if silent staleness at 200-file granularity is unacceptable. No test exercises this branch either way.

[code-read] — cross-engine: Claude reliability (P2/72) + Codex (GPT-5.5) reliability (P2/75) + Claude project-standards (DoD citation); the adversarial lane lists it as its top residual risk.

gitnexus/scripts/cross-platform-tests.ts
```diff
@@ -99,6 +99,15 @@ const LBUG_NATIVE = [
   'test/integration/analyze-wal-checkpoint-failure.test.ts',
   'test/integration/fts-stemmer-sweep.test.ts',
   'test/integration/lbug-multiwriter-deadlock.test.ts',
+  // #2409 batched incremental writeback: chunked IN-list DETACH DELETEs +
+  // backslash quote escaping against the REAL native engine — the failing
+  // environment for #2409 was Windows, so the write pattern must be proven
+  // on the windows-latest native addon, not just Ubuntu.
+  'test/integration/lbug-delete-nodes-for-files.test.ts',
```
magyargergo, line 106: **P2 — both new native-DB tests are registered here for the Windows matrix but neither was added to the `lbug-db` vitest project.** TESTING.md line 68 requires it verbatim: "When adding a new test that uses native LadybugDB (`@ladybugdb/core`), add it to the `lbug-db` project's explicit include list and the `default` project's exclude list."

Verified in the worktree: `gitnexus/vitest.config.ts`'s include list (lines 60-91) contains neither `test/integration/lbug-delete-nodes-for-files.test.ts` nor `test/unit/incremental-dirty-recovery.test.ts`, so both run in the **parallel-forks `default` project**. `run-cross-platform.ts` uses the main vitest config, so on windows-latest — the platform this PR is proving #2409 behavior on — they run in exactly the parallel-fork configuration the `lbug-db` project exists to prevent ("LadybugDB's native mmap addon causes file-lock conflicts when vitest runs lbug test files in parallel forks on Windows", vitest.config.ts:46-48). Precedent: 12 of 15 sibling `lbug-*` integration tests ARE registered; the unit-test half has softer standing (its sibling `incremental-orchestration.test.ts` is likewise unregistered and the list holds no `test/unit` files today). Current CI is green and per-suite temp-dir DBs mitigate — but per the config's own comment, isolation does not eliminate the hazard; this is flake insurance for the exact shards that gate this PR.

**Fix (2 lines × 2 lists):** add the integration test — and, for TESTING.md consistency, the unit test — to the `lbug-db` include list and the `default` exclude list.

[code-read] — rule text, both config lists, and sibling precedent verified by synthesis; found by the Claude project-standards lane.

gitnexus/src/core/lbug/sidecar-recovery.ts
```diff
@@ -429,6 +429,59 @@ export async function finalizeLbugSidecarsAfterClose(
...
+ * died mid-writeback — its WAL can be poisoned in a way that natively
+ * kills the process on replay. The recovery run used to open the DB
+ * BEFORE the rebuild wipe (the embedding-cache preservation open), replay
+ * the poisoned WAL, and die on the spot — so recovery never happened and
+ * only a manual rename-aside of the index dir escaped the loop. The
+ * rebuild discards every pending WAL byte anyway (the DB files are wiped),
+ * so parking the sidecars first costs nothing and makes every subsequent
+ * open replay-free.
+ *
+ * Renamed, never deleted — same philosophy as
```
magyargergo: **P2 — the "same philosophy as `quarantineWalForMissingShadow`" parity claim is inaccurate in the one way users will notice.** Missing-shadow quarantine files are surfaced and deletable via `gitnexus clean --lbug-sidecars`; the new `*.dirty-recovery` files are invisible to it — `listQuarantinedMissingShadowWals` filters only `` `${base}.wal.missing-shadow.` `` (sidecar-recovery.ts:496), `clean.ts:103` calls only `cleanQuarantinedMissingShadowWals`, and there is no `.dirty-recovery` reference anywhere in the CLI (grep-verified).

So after any dirty recovery, up to two files — a full WAL of a large repo plus shadow, potentially hundreds of MB, containing raw DB page bytes of indexed source — sit in the user's `.gitnexus/` while `clean --lbug-sidecars` reports nothing to clean. No user-facing doc (GUARDRAILS/ARCHITECTURE/RUNBOOK) mentions them, even though the PR description calls them "exactly the post-mortem artifact to attach". Mitigations verified: fixed destination names cap accumulation at one copy per sidecar, and a full `gitnexus clean` removes the whole directory.

**Fix:** extend the lister (or add a `.dirty-recovery` twin) so `clean --lbug-sidecars` surfaces them; correct this docstring line; and add one GUARDRAILS/RUNBOOK bullet ("parked sidecars from an interrupted run, kept for debugging, safe to delete"). Sibling PR #2410's review feedback was this exact docs-gap class.

[code-read] — lister filter + clean.ts verified by synthesis; converged by Claude docs-DoD (two P2s), security (P3), risk (P3); Codex (GPT-5.5) swarm lists the RUNBOOK mention as a non-blocking follow-up.

magyargergo (commented):
# Tri-engine review digest — PR #2425 `fix: make large incremental writebacks commit reliably (#2409)`

Reviewed at head `4b34519f` (merge of `9bc95961` + `938a6169` into main), base `56186a4a`, all 13 changed files
read in full.

**How this review ran.** Two engines, five legs. Claude ran the 6-lane GitNexus PR swarm (facts, hygiene, risk,
security, test/CI, docs/DoD) plus a 7-persona code review (correctness, adversarial, reliability, testing,
performance, maintainability, project-standards) — 13 lanes. Codex (GPT-5.5, xhigh reasoning) ran the swarm in
solo mode, a persona code review (6 personas completed: correctness, testing, maintainability,
project-standards, performance, reliability; its adversarial/security/agent-native personas failed at provider
setup and did not run), and a dedicated adversarial pass. All three Codex legs produced full reports; the swarm
leg lost its stream right before emitting and was recovered via thread resume with its findings intact. The
coordinator additionally reproduced two claims against the real engine/DB. Findings below are weighted by
cross-engine corroboration — same-engine agreement is never counted as independent confirmation.

## What's solid

The core design holds up under hostile review, and several suspicions were **empirically refuted** against the
real native engine:

- **Crash-safety of the escalation valve is real.** Every crash boundary was traced (Claude risk + adversarial,
  independently by Codex): the `'escalated-full-write'` dirty flag is durably saved (atomic tmp+rename) strictly
  before the driver stop / close / wipe; recovery is phase-agnostic (any truthy flag → park sidecars → full
  rebuild); `stop()` is idempotent (pinned by a pre-existing test); `closeLbug()` cannot throw.
  Kill-at-every-boundary converges.
- **Embedding-cache preservation survives escalation.** The suspected "valve forces a full re-embed" regression
  is a false alarm: the cache loads pre-pipeline, Phase 3.5 re-inserts rows on both branches, and the hash-skip
  map avoids ML recomputation. The escalated path pays only bounded batch-insert I/O — same as `--force` always
  has. (But see the headline: rows are preserved, the *index* is not.)
- **Escaping is correct for practical inputs.** Live probes: `'we''ird'` confirmed a parser error (the PR's
  premise); quotes/backslashes/real-newline filenames round-trip end-to-end under `escapeCypherString` (insert →
  IN-list match → OR-chain match → DETACH DELETE, 4/4); raw LF inside single-quoted literals is legal; 120KB
  IN-lists and 1000-clause OR-chains parse fine. No injection vector found across the converted sites (11 call
  sites in this diff).
- **Escalated ≡ `--force` for graph data.** FTS indexes are unconditionally dropped and recreated after both
  branches with a hard verify; the missing `pdgEmitManifest` arg is a no-op (`resolveStreamPdgEmit` requires
  `force === true` — verified three times independently); communities/processes/INJECTS handling is correct on
  both branches.
- **The perf fix is genuine** — per-file transaction storm (~13k single-row write txns) collapsed to ~124
  batched statements for the 700-file case, BFS round-trips cut ~175×, no quadratic dedup, memory trivial. Both
  engines' perf analyses agree the win is round-trip/lock count, robust to either IN-evaluation strategy.
- **Test craftsmanship**: the cypher-escape suite pins the order-sensitivity trap with a well-chosen adversarial
  input; the DETACH boundary arithmetic in the new integration test was hand-verified correct; chunk assertions
  derive from the exported constant; repo rules (no `any`, no test if-branching, no CHANGELOG edit, NodeNext
  `.js` imports) are clean; branch hygiene is clean (13 files, zero unrelated churn, lockfile-only merge
  commit).

## Headline finding

**P1 — the escalation valve wipes the HNSW vector index, and the preserve-only incremental path never recreates
it; meta certifies `vector-index` anyway.** An escalated run is by definition `force=false`, so
`deriveEmbeddingMode` yields `shouldGenerateEmbeddings=false` for any previously-embedded repo. The valve's wipe
destroys the index with the DB file; Phase 3.5 restores embedding **rows** only; `createVectorIndex` is
reachable solely through `buildVectorIndex` inside the (skipped) embedding pipeline. Meta then falls back to
`runtimeCapabilities.semanticMode` and stamps `vectorSearch.status: 'vector-index'` because the restored rows
make `embeddingCount > 0`. At query time, `QUERY_VECTOR_INDEX` throws, is swallowed, and the exact-scan fallback
caps at 10k rows — above that, semantic search silently returns `[]` forever until a manual
`--force`/`--embeddings`.

**Trigger:** embedded repo on macOS/Linux, routine `gitnexus analyze` after a hub edit that trips the valve —
the exact scenario this PR targets. This breaks the PR's own "indistinguishable from a `--force` rebuild"
invariant (`--force` regenerates and rebuilds the index), it is introduced *by* this PR (the valve is the first
`force=false` wipe on the mainstream git-repo flow), and the new escalation test cannot see it (fixture has no
embeddings; stats parity omits `embeddings`/index presence). Found independently by two Claude review passes;
every link verified by code-read (coordinator + synthesis).

## Inline findings (posted as review comments)

1. **P1** Escalated path never rebuilds the vector index; meta misreports (`run-analyze.ts:1374`).
2. **P2** `deleteNodesForFiles`' embedding delete is a provable no-op — embedding nodeIds are label-first,
   `STARTS WITH '<filePath>'` never matches — and the new comment enshrines the wrong format. Reproduced against
   the live DB. Pre-existing behavior, mirrored into new code (`lbug-adapter.ts:2094`).
3. **P2** Escaping sweep incomplete: three `''`-doubling Cypher builders remain in `lbug-adapter.ts` (1289 /
   1378 / 1455) despite the commit claiming "everywhere"; the `fallbackRelationshipInserts` one is
   production-reachable and silently drops quoted-id relationships (`lbug-adapter.ts:9` anchor).
4. **P2** Windows locked-WAL fallthrough: if the sidecar park rename fails, the pre-wipe embedding-cache open
   still replays the poisoned WAL — the death loop persists on that branch while the lock does; the warning text
   over-promises (`sidecar-recovery.ts:469`).
5. **P2** Valve wipe-`rm` failures are swallowed; reopening the still-populated DB and bulk-COPYing into it was
   empirically shown to keep stale rows and duplicate every relationship, silently (`run-analyze.ts:1367`).
6. **P2** `queryImportersBatch` swallows per-chunk failures with zero logging — up to 200 files' importer
   expansion lost per event, then the run exits 0 and clears the dirty flag (`lbug-adapter.ts:2195`).
7. **P2** Both new native-DB tests missing from the `lbug-db` vitest project include/exclude lists that
   TESTING.md requires — on windows-latest they run in the parallel-fork configuration the project exists to
   prevent (`cross-platform-tests.ts:106`).
8. **P2** `*.dirty-recovery` parked files are invisible to `gitnexus clean --lbug-sidecars` and undocumented;
   the "same philosophy as quarantineWalForMissingShadow" claim is inaccurate — those ARE surfaced by the
   command (`sidecar-recovery.ts:446`).

## Lower-priority notes (no inline comment)

- `DELETE_FILES_CHUNK_SIZE` docstring says a 700-file set collapses "to ~40" statements; the real figure is 124
  (30 tables + 1 embedding statement × 4 chunks). Qualitative claim fine, arithmetic off ~3×
  (`lbug-adapter.ts:2051`).
- `writeFraction`'s numerator and denominator come from different populations: deleted importers found by the
  BFS inflate the numerator (fraction can exceed 1; log can print >100%), while `hashDiff.deleted` alone doesn't
  count toward it (mass-delete runs stay surgical). Both skews were analyzed as direction-safe — early
  escalation is conservative, and batched deletes make a delete-heavy surgical run cheap (a 900-delete run is ~5
  chunks) — but the knob doesn't measure what its name says. Clamp/intersect or document.
- `checkpointOnce()` at the writeback boundary converts a >750ms checkpoint-IO flake (e.g. AV holding the WAL)
  into a failed analyze + forced full rebuild next run, where the pre-PR periodic driver absorbed the same
  flake. Three lanes independently concluded this is acceptable (loud, crash-safe, self-healing) — flagged only
  so the trade-off is confirmed intentional; consider a doc note or one extra `tryFlushWAL` in the outer catch.
- `analyze --repair-fts` opens the DB (line 678) *before* the dirty-flag quarantine (line 784) — the
  poisoned-WAL death remains reachable through that maintenance flag. Narrow; worth hoisting the check or
  refusing on a dirty flag.
- The singular `deleteNodesForFile`/`queryImporters` now have zero production callers; doc comments in
  `run-analyze.ts`/`subgraph-extract.ts`/`shadow-candidates.ts` still name them, and two pre-existing test files
  (`run-analyze-fts-repair.test.ts`, `repo-manager-reconcile.test.ts`) build full module-mock factories listing
  only the old names — verified currently harmless (those tests never reach the incremental branch) but a
  `TypeError` trap for the next default-options test.
- The 4-file wipe list is now duplicated verbatim (full-rebuild wipe + valve wipe). List drift is exactly how
  `.shadow` went missing pre-PR — one root cause of #2409. Extract a shared helper.
- **Test gaps** (follow-up material, both engines' testing lanes): the valve's AND-gate is only tested at
  both-true/both-false corners (an `&&`→`||` mutation would pass CI); `queryImportersBatch` has no direct test
  (chunk boundary, failure branch, sort, dedup all unexercised); `deleteNodesForFiles`' errors-propagate
  contract is untested; the quarantine failure branch is untested; the dirty-recovery test never exercises the
  pre-wipe embedding-cache open it exists to protect (fixture has `stats.embeddings = 0`, so `shouldLoadCache`
  is false — the sharpest version of this came from the Codex testing persona); `semanticSearch`'s real body has
  zero test coverage from any angle while this diff changes one of its lines; the `'escalated-full-write'` phase
  value is never driven through recovery.
- **Docs**: `lbug.shadow` is absent from ARCHITECTURE.md's storage table; the escalation valve makes
  GUARDRAILS.md's "only changed-file rows... are rewritten" conditionally stale; `.dirty-recovery` needs a "safe
  to delete" mention. Sibling PR #2410's review theme was this exact class.

## Suspicions raised and refuted (credit where due)

The review spent real effort trying to break this PR; these did not survive scrutiny: valve crash windows at
every boundary (converges; flag-before-destruction verified durable); embedding-cache loss / forced re-embed on
escalation (Phase 3.5 is unconditional); `pdgEmitManifest` omission (streaming is force-gated); IN-list/OR-chain
statement-size limits (120KB/1000 clauses pass on the real engine); hostile-filename injection or DoS (all
probes safe; raw newline is a legal literal char); FTS gap on the escalated path (unconditional
drop+recreate+verify); valve thrash between runs (stateless, AND-gated, identical bookkeeping); WAL-driver
double-stop (idempotent, test-pinned); `deleteNodesForFiles` bricking on missing tables (every non-skipped table
has `filePath`; schema created on every open); BFS closure inequivalence vs the old per-file loop (proved
set-identical per depth); recovery mishandling the new `toWriteCount: 0` sentinel (recovery is phase-agnostic;
the sentinel has one log-only consumer).

## CI status

All checks green at synthesis time (38 pass, 1 skip: `e2e (chromium)`, gated by its own no-web-changes check):
quality lint/format/typecheck ×2, ubuntu coverage 3/3 + merge, windows-latest platform-sensitive 3/3, macos 3/3,
benchmarks, node-floor 22.14, packaged-install smoke (ubuntu+windows), tree-sitter ABI ×3, CodeQL, gitleaks,
Docker builds, Vercel. Both new tests are correctly registered in `LBUG_NATIVE` and ran on the windows shards.
Zero human reviews so far.

## Verdict

- **Final verdict: `not production-ready`** — the core design is sound and most hostile hypotheses were refuted
  under real-engine probing, but one verified P1 introduced by this PR remains: an escalated run destroys the
  HNSW vector index, nothing on the preserve-only path recreates it, and meta then certifies `vector-index`
  anyway — breaking the valve's own "indistinguishable from `--force`" invariant on the mainstream scenario the
  valve targets. It never self-heals without a manual `--force`/`--embeddings`, and the new escalation test
  cannot detect it (no embeddings in its fixture). The fix is small and in-scope (recreate the index after the
  escalated restore, or stamp the degraded mode honestly); with it and the cheap P2 hardenings below addressed,
  this PR is otherwise in strong shape.
- **Branch hygiene: `clean feature/fix PR`** — 13 files, zero unrelated churn, lockfile-only merge-from-main
  commit consistent with repo precedent.
- **Merge state: `review blocked`** — mergeable with all checks green, but zero human approvals and an
  outstanding verified P1.

## Recommendation

Comment review (no approve/request-changes event). Suggested split, given this repo's fix-in-PR preference:

**Fix in this PR (all small):** the P1 vector-index recreate + an embedded-fixture escalation assertion; reuse
`escapeCypherString` at the three leftover sites (+1 quoted fallback-insert test); fix the embedding-delete
predicate or at least its comment; a `warn` in the `queryImportersBatch` catch; verify-gone after the wipe `rm`
loops; parking-failure hardening (skip the pre-wipe cache open when parking failed) + accurate warning text; the
2×2-line vitest `lbug-db` registration; `clean --lbug-sidecars` coverage for `.dirty-recovery` + the
parity-comment and "~40"→124 docstring fixes.

**Fine as follow-ups:** the test-architecture items (valve boundary corners, direct `queryImportersBatch` test,
error-propagation pin, stale mock-factory rename, `semanticSearch` coverage), GUARDRAILS/ARCHITECTURE doc
updates, `--repair-fts` parking order, the checkpointOnce trade-off note, and the shared wipe-helper extraction.

---
*Automated multi-engine review digest (Claude swarm + personas; Codex GPT-5.5 swarm/personas/adversarial;
coordinator-verified evidence). Findings were individually verified against the worktree, but verify before
acting — especially the empirical claims, which were probed on Linux x64 against the bundled native engine.*

magyargergo:
## Tri-review follow-ups: all findings fixed (review 4669518496)

Five commits land every finding from the multi-engine review, plus hardening from an 8-angle shipping review of the fixes themselves.

### Review findings → fixes

| Finding | Fix | Commit |
|---|---|---|
| **P1** escalated run destroys the HNSW vector index; meta stamps `vector-index` anyway | `buildVectorIndex` exported and called after the cached-row restore on any wiped preserve-mode run (cap-aware gate `restored>0 && dbWasWiped && embeddingSkipped`); `semanticMode` set from the actual result; `RepoMeta.capabilities` typed | 594fefc0 |
| **P2** embedding `STARTS WITH '<filePath>'` delete was a no-op (nodeIds are label-first) | Both delete variants join through the node tables before the DETACH DELETE (order is load-bearing); the factually wrong comment replaced with the verified id contract | 594fefc0 |
| **P2** three `''`-doubling Cypher builders remained | All three route through `escapeCypherString` via one `formatCypherValue`; `fallbackRelationshipInserts` exported + quoted-id regression tests against the real engine | 02fc3682 |
| **P2** locked-WAL park failure fell through to the pre-wipe open | Redesigned (see shipping review below): retry → collision probe → rm-fallback → fail-fast typed error. The poisoned-WAL replay window is closed on every branch | 7b49d77e + 9b034973 |
| **P2** valve wipe swallowed rm failures → silent COPY corruption | Shared `wipeLbugDbFiles` with ENOENT-verified removal, bounded retry, typed `LbugWipeError` + dedicated CLI handler; both wipe sites use it | 02fc3682 |
| **P2** `queryImportersBatch` swallowed chunk failures silently | `logger.warn({err})` + `onChunkFailure`; `droppedImporterChunks` persisted into the #2410 crash diagnostics (stamped only when >0) and printed by the recovery log | d26e048c |
| **P2** native-DB tests missing from the vitest `lbug-db` project | All four suites registered (include + exclude lists), incl. `incremental-orchestration` which gained native tests here; registration audit in the commit | d26e048c |
| **P2** `*.dirty-recovery` files invisible to `clean --lbug-sidecars` | Aggregate lister/cleaner over both sidecar families (+ `.next` residues), per-file unlink tolerance, en/zh-CN strings, docstring parity claim corrected | d26e048c + 9b034973 |
| **P3 sweep** | docstring math (124, not ~40), `writeFraction` documented + display-clamped via a pure `shouldEscalateIncrementalWrite` predicate with 4-corner tests, stale singular-API comments, mock-factory hygiene, `--repair-fts` refuses on a dirty flag, GUARDRAILS/ARCHITECTURE updates | 02fc3682 / 7b49d77e / d26e048c |
| **Test gaps** | Escalated-with-embeddings test (SHOW_INDEXES actually runs on linux — VECTOR is statically linked), surgical embeddings-parity test, dirty-recovery test now seeds embeddings so it traverses the exact pre-wipe open it protects, park-failure injection tests, `queryImportersBatch` direct suite with a real DROP TABLE failure block | 594fefc0 / 7b49d77e / d26e048c |

### Scope disclosure (self-identified during planning)

Making the embedding delete real **un-masked a latent restore defect**: Phase 3.5 re-inserted all cached rows, PK conflicts aborted whole 200-row batches (silently losing changed-file embeddings in preserve mode), and deleted files' rows were resurrected. The restore now derives its scope in memory (no extra table scan), filters to live graph nodes, and a guarded exact-id sweep removes the no-op-era legacy orphans on surgical runs — so pre-existing orphaned rows heal without a schema bump. This is beyond the review's findings but entailed by them; flagging it explicitly for re-review attention (it touches the hottest data path).

### Shipping review (8 angles over the fix commits themselves) → 9b034973

The strongest finding replaced the original park-failure design: dropping embeddings on a failed park was wrong twice over — an explicit `--embeddings` recovery invocation still reached the embedder, and under a persistent lock the run wasted the whole pipeline and then died at the wipe on the same handle. Now: rename retries over the shared handle-release budget, a probe rename discriminates true collisions, an rm-fallback removes the un-parkable sidecar (no WAL ⇒ nothing to replay ⇒ **embeddings still preserved**), and only total failure throws a typed, holder-naming error in seconds. Also from that review: `.lock` survivors no longer abort a sound rebuild; a persisted honest `exact-scan` stamp is carried forward instead of being reverted by the platform fallback on the next surgical run; the embedding join scopes to `EMBEDDABLE_LABELS` (probe-proven multi-label alternation) instead of scanning every node table per chunk; a locked parked file no longer crashes `clean --lbug-sidecars` mid-command.

### Deferred (follow-up material, deliberate)

- A real `semanticSearch` end-to-end harness (its silent-empty behavior above the 10k exact-scan cap on a missing index is the motivating symptom; the SHOW_INDEXES assertions cover the index half).
- `formatCypherValue`'s long-term home in `cypher-escape.ts`, and an empirical probe of `''`-doubling inside LadybugDB **CSV list elements** (csv-generator builds array literals inside CSV fields — different sub-grammar, pre-existing, unproven either way).
- Removing/wrapping the now-production-dead singular `deleteNodesForFile`/`queryImporters` (test callers pin a different error contract).
- `gitnexus doctor` prints platform capabilities and can disagree with a repo-level `exact-scan` stamp (per-platform by design).

### Post-deploy monitoring & validation

Log lines worth watching after this lands: `switching to a full DB write` (expected on >50% write sets — not a bug), `Parked lbug.wal.dirty-recovery` (crash recovery worked; artifact removable via `clean --lbug-sidecars`), `Importer BFS chunk failed` (new visibility — was silent; investigate if recurring), `Vector index creation failed; semantic search will use exact-scan fallback` (extension/platform issue), and `LbugWipeError`/lock guidance (persistent holder — stop MCP/serve or AV-exclude; an immediate re-run often succeeds for transient AV lag). Healthy signal: none of these on routine incremental runs; `stats.embeddings` stable across surgical runs (new orchestration test pins this). Recovery diagnostics now include `droppedImporterChunks` when any BFS chunk was lost.

*Automated follow-up batch (plan → 3-agent deepening → 4-persona doc review → 4 implementation batches → 8-angle shipping review); all suites green locally incl. the two heavy orchestration suites under the serialized `lbug-db` project. CI on this head is the authoritative gate.*

PR #2429, opened 2026-07-11
fix(parser): replace null bytes before safe parse

### LINKED ISSUE #2426: Null byte (`\x00`) in source causes tree-sitter to silently truncate parsing — methods dropped from index with no warning
[NullByteDemoService.java](https://github.com/user-attachments/files/29887572/NullByteDemoService.java)

## Summary

When a Java source file contains a null byte (`\x00`, e.g. in a corrupted Javadoc `@example` value), tree-sitter's C binding treats it as a null-terminated C-string terminator. This causes **silent truncation** of the parse tree: all code after the null byte position is invisible to the parser, and every method defined after that point is **dropped from the GitNexus index** without any warning or error.

## Root Cause

tree-sitter's Node native binding converts JavaScript strings to C strings (`char*`) internally. In C, `\x00` is the string terminator — the parser stops reading at that byte. This affects both parse paths in `safe-parse.ts`:

- **Direct path** (`parser.parse(string)` for files ≤ 16KB): The entire JS string is converted to a C buffer; `\x00` terminates it.
- **Chunked path** (`Parser.Input` callback for files > 16KB): Each chunk returned by `sourceText.slice()` is converted to a C buffer individually; `\x00` within a chunk terminates that chunk. tree-sitter internally concatenates chunks, but the null byte in the concatenated buffer still acts as a terminator.

## Impact

- **Silent data loss**: Methods, classes, and any code after the null byte are never indexed. No error, no warning, no degraded-parse log — the file simply appears to have fewer symbols than it actually does.
- **Search failure**: `query()`, `context()`, and BM25/vector search cannot find any symbol defined after the null byte.
- **Blast radius blindness**: `impact()` analysis is incomplete for symbols near/after the truncation point.
- **Hard to diagnose**: The null byte is invisible in most text editors and `cat`/`head` output. Only `xxd` or binary inspection reveals it.

### Test file

See attached `NullByteDemoService.java` (68.6 KB, 110 methods, null byte at line 1134).

The null byte is embedded in a Javadoc `@example` line:

```java
@example paramsMap={"idList":["ITEM-001"],"allLevel":true,"dataStyle":"<NULL_BYTE_HERE>"}
```

### Step 1: Index WITH null byte

```
gitnexus analyze --force --skip-git
```

**Result**: Only 79 of 110 methods indexed. All 31 methods after the null byte (including `batchGetStructure`, `getItemUsesOccurrences`, `getSubstituteFor`, etc.) are missing.

Cypher verification:
```cypher
MATCH (m:Method) WHERE m.filePath CONTAINS 'NullByteDemoService' RETURN count(m)
// → 79 (expected 110)
```

### Step 2: Remove null byte, re-index

Remove the `\x00` byte from the file, then:

```
gitnexus analyze --force --skip-git
```

**Result**: All 110 methods indexed, including every method that was previously missing.

Cypher verification:
```cypher
MATCH (m:Method) WHERE m.filePath CONTAINS 'NullByteDemoService' RETURN count(m)
// → 110 (correct)
```

### Comparison

| Scenario | Methods indexed | Missing |
|----------|----------------|---------|
| With `\x00` | 79 | 31 (all after null byte) |
| Without `\x00` | 110 | 0 |

## Suggested Fix

Strip or replace null bytes from `sourceText` before passing it to tree-sitter, regardless of which parse path is used. A minimal fix in `safe-parse.ts`:

```typescript
// Before parsing, sanitize null bytes that would truncate tree-sitter's C-string handling
const sanitizedText = sourceText.replace(/\0/g, '');
```

This should be applied **before** the `sourceText.length` check that selects the parse path, so both direct and chunked paths receive sanitized input.

### Considerations

- **Position mapping**: Stripping `\x00` shifts character positions by 1 per removal. If tree-sitter node positions are used to locate symbols in the original file, a position offset map may be needed. However, null bytes are rare enough that a simple replacement with a space (`\x00` → ` `) would preserve positions and still prevent truncation (a space in a Javadoc comment is harmless).
- **Alternative**: Replace `\x00` with a space character instead of removing it, to preserve byte offsets:
  ```typescript
  const sanitizedText = sourceText.replace(/\0/g, ' ');
  ```
- **Logging**: When null bytes are detected, emit a `warn`-level log so operators can identify and fix the corrupted source files.

## Affected Code

- `gitnexus/src/core/tree-sitter/safe-parse.ts` — `parseSourceSafe()` passes `sourceText` directly to tree-sitter without sanitization.
- `gitnexus/src/core/ingestion/workers/parse-worker.ts` — line ~1282: `getProvider(language).preprocessSource?.(parseContent, file.path) ?? parseContent` — Java has no `preprocessSource` hook, so null bytes pass through unchanged.

## Environment

- GitNexus version: latest (main branch, commit `5f4964b4`)
- tree-sitter: 0.21.x (Node native binding)
- OS: Windows 11
- Language: Java (also affects any language parsed by tree-sitter)

koriyoshi2041:
## What changed

Fixes #2426.

- replace `\0` bytes with spaces inside `parseSourceSafe` before either the direct string path or callback input path reaches tree-sitter
- keep source length stable so node offsets are not shifted
- emit a warning with the file label and null-byte count when sanitization happens
- add direct-path and callback-path regression coverage so source after a null byte is still parsed

## Verification

- `npm run test -- test/unit/safe-parse.test.ts`
- `npx tsc --noEmit`
- `npx prettier --check gitnexus/src/core/tree-sitter/safe-parse.ts gitnexus/test/unit/safe-parse.test.ts`
- `npm run build`
- `git diff --check`

## Risk

Low. The sanitizer only changes source text when a file contains null bytes, and replacement uses spaces rather than deletion to preserve positions.

### DIFF
--- a/gitnexus/src/core/tree-sitter/safe-parse.ts
+++ b/gitnexus/src/core/tree-sitter/safe-parse.ts
@@ -34,6 +34,22 @@ const DIRECT_PARSE_LIMIT_CHARS = 16 * 1024;
  * (unlimited parse time — restore the historical behaviour for debugging).
  */
 const DEFAULT_PARSE_TIMEOUT_MS = 15_000;
+const NULL_BYTE_PATTERN = /\0/g;
+
+function replaceNullBytes(sourceText: string, label?: string): string {
+  if (!sourceText.includes('\0')) return sourceText;
+
+  const nullByteCount = sourceText.match(NULL_BYTE_PATTERN)?.length ?? 0;
+  const sanitizedText = sourceText.replace(NULL_BYTE_PATTERN, ' ');
+  logger.warn(
+    {
+      ...(label ? { file: label } : {}),
+      nullByteCount,
+    },
+    'tree-sitter input contained null bytes; replaced them with spaces before parsing',
+  );
+  return sanitizedText;
+}
 
 /**
  * Resolve the per-parse budget in milliseconds from the environment. A
@@ -196,7 +212,12 @@ export function getParseDiagnostics(tree: Parser.Tree): {
  *     `finally` so it never leaks to the next parse on a reused/singleton
  *     parser (`loadParser()` and the worker both reuse one `Parser`).
  *
- *  3. **Intrinsic error detection.** On a successful parse, a degraded tree
+ *  3. **Null-byte truncation guard.** Inputs containing `\0` have those bytes
+ *     replaced with spaces before either parse path so tree-sitter's native
+ *     binding cannot treat them as C-string terminators. Replacement preserves
+ *     source length and node offsets while surfacing a warning for the file.
+ *
+ *  4. **Intrinsic error detection.** On a successful parse, a degraded tree
  *     (`rootNode.hasError`) is logged at `debug` level with throttling, then
  *     the tree is **returned anyway** — error recovery is a downgrade, never a
  *     drop. Callers wanting the boolean use {@link parseHadErrors}.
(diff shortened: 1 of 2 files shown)


magyargergo:
Superseeded by https://github.com/abhigyanpatwari/GitNexus/pull/2430. Thanks for your contribution @koriyoshi2041 !

PR #2198, opened 2026-06-14
fix: remove hardcoded 300-flows cap for large repositories

ChunxueLi:
## Problem

The `dynamicMaxProcesses` was capped at 300 via `Math.min(300, ...)`, causing large repositories (280K+ nodes) to lose execution flows.

## Solution

Remove the `Math.min(300, ...)` cap, keep dynamic calculation based on symbol count.

## Effect

- **Before**: 280K-node repo → 300 flows (hard cap)
- **After**: 280K-node repo → 1617 flows (dynamic)

## Code Change

```diff
- const dynamicMaxProcesses = Math.max(20, Math.min(300, Math.round(symbolCount / 10)));
+ const dynamicMaxProcesses = Math.max(20, Math.round(symbolCount / 10));
```

## Testing

Verified on a 280,951-node Java repository:
- Before: 300 flows
- After: 1617 flows

Small repositories (< 3000 symbols) are unaffected (dynamic calculation already < 300).

---

*Contributed by [卫宁健康科技集团](https://www.winning.com.cn) — WiNEX门诊医生站团队*

Issue #15, opened 2026-01-10
The packages folder is being ignored, but it’s actually part of the core project.


them7d:
I tried running https://github.com/GrapesJS/grapesjs, but the packages folder, which is the core project folder, was ignored.

abhigyanpatwari:
<img width="592" height="423" alt="Image" src="https://github.com/user-attachments/assets/26797d56-08b2-4782-9436-8b31a27b2257" /> 

Did u see this error?  Interestingly, uploading the zip directly works fine but the git clone shows error. Something might be wrong with the proxy service. Will check.

Thanks for reporting this

abhigyanpatwari:
Fixed it. 

The issue was symlink. ( a type of file that serves as a reference or pointer to another file or directory within a file system ).
This was breaking the parsing logic. Basic try catch fix

them7d:
thanks for fixing enoent error i faced the **enoent** error and tried to upload .zip file

the point that i mean that folder `/packages` is ignored when parsing the folders and files of project
@abhigyanpatwari 
<img width="481" height="541" alt="Image" src="https://github.com/user-attachments/assets/e3572a2a-8d1c-40aa-ac34-1bcec44a1799" />

in gitnexus
<img width="352" height="406" alt="Image" src="https://github.com/user-attachments/assets/65afc156-23ee-41d7-9d31-5546215f5d2e" />

abhigyanpatwari:
Ok got it, will check. Reopening issue for now

abhigyanpatwari:
Fixed the issue. Also there are several improvements overall ( deeper context, cyfer validation, UI improvements, etc).
Let me know if it worked for you.

abhigyanpatwari:
<img width="2502" height="1280" alt="Image" src="https://github.com/user-attachments/assets/050c6e09-0764-4009-827d-9bb32f7a416a" />

this pattern looks so cool :-)  🥷⭐

PR #14, opened 2026-01-10
prompt changes

abhigyanpatwari:

PR #2431, opened 2026-07-11
feat(skills): engineering skill family — plan/work/review/lfg

magyargergo:
## What

Adds the **GitNexus-native engineering skill family** — four canonical, CLI-neutral specs under `.claude/skills/` that work from Claude Code (slash commands), Cursor, Codex (via `AGENTS.md` + documented user-level install), and any AGENTS.md-aware agent:

| Skill | Role |
|---|---|
| **`gitnexus-plan`** | Deep, implementation-ready plans: graph navigation (`query` → `context` → `impact`/`trace`), bounded statement-level PDG slices, targeted source verification. Emits a 13-section plan into `docs/plans/` whose §11 is a machine-readable **implementation context pack**; asks plan depth up front (quick/standard/deep); Deepen mode strengthens an existing plan in place. |
| **`gitnexus-work`** | Executes a plan as verified atomic commits: drift-checks the plan's pinned evidence commit against HEAD, `impact` before every symbol edit, `detect_changes` before every commit, tests from the plan's scenarios — then **refreshes the knowledge graph** (`analyze --index-only`, keeping `--pdg` when present) so review and later sessions query the finished work. |
| **`gitnexus-review`** | (renamed from `gitnexus-pr-review`) Read-only review of a **PR, branch/range, or local changes**: exact-SHA checkout/index alignment, `detect_changes` + upstream `impact` evidence, a **PDG-backed taint pass** on trust-boundary diffs, and **expert lenses** — per-domain reviewers derived from the graph's clusters, plus four cross-cutting checks (architectural fit, language conformance per the repo's own config, Definition of Done, simplicity/YAGNI), scaled to the diff. |
| **`gitnexus-lfg`** | Thin pipeline: plan (depth asked up front) → **blocking user gate** (proceed / stop) → work → review. One bounded fix cycle; never pushes or opens a PR on its own. |

## Key design decisions

- **Freshness is a gate, not advisory**: a stale index or missing PDG layer is refreshed via `analyze --index-only [--pdg]` through a resolved runner ladder (`node .gitnexus/run.cjs` → installed `gitnexus` → `npx gitnexus`). `--index-only` is load-bearing: it writes only the `.gitnexus` store, so planning/review stay read-only and work's post-commit refresh leaves the tree coherent.
- **Knowledge stays fresh through the pipeline**: work's Phase 4 re-indexes after the DoD walk, so lfg's review lane and every later session see the finished work, not the pre-work graph.
- **Review evidence model**: graph evidence (`detect_changes` at the merge-base, d=1 dependents) plus source proof; taint findings are judged against the diff (an introduced flow or removed sanitizer is a finding, a pre-existing flow is context); expert lenses dispatch once after the evidence-gathering steps and report through a single finding standard.
- **Auditable plans**: evidence pinned to a commit, claims tagged `[verified]/[graph]/[inferred]/[assumed]`, a context ledger with symbol budgets; `context-pack.md` fixes the §11 field names `gitnexus-work` consumes.
- **Data-safe rename migration**: `gitnexus setup` warns on a legacy `gitnexus-pr-review` dir and leaves it in place (the installer can't prove ownership); `gitnexus uninstall` removes it (ownership-by-name is its documented contract); a sync-test guard keeps the legacy name from resurrecting in any shipped tree.

## Validation

- Multi-agent adversarial review of the plan/work/lfg cross-skill contracts in earlier rounds (21 findings, all fixed) plus two live end-to-end scenario runs with independent evidence auditors.
- A **six-lens expert review at the current head** (CLI/TypeScript, skill-content, eval-harness, architecture+YAGNI, DoD/docs, CI/security) produced 31 findings; every actionable one is fixed in the last five commits. Highlights: `ci.yml`'s `paths-ignore: '**.md'` exempted md-only skill edits from the drift guard (fixed with the paths-triggered `skill-sync.yml` workflow), doc routing tables pointed at untracked paths, uninstall orphaned the legacy skill name, and the benchmark's promotion gate compared arms on main-loop-only token counts.
- 56 vitest (sync/setup/uninstall) + 36 pytest (eval) green; `tsc --noEmit` clean.

## Benchmark (`eval/workflow_bench/`)

Real headless `claude -p --output-format json` sessions in fresh detached worktrees: baseline vs workflow arms plus external comparators (`ce_workflow`, `ce_workflow_direct`, `review`, `ce_review`). The deterministic promotion gate now defaults to **`cost_usd`** (the only metric that includes subagent spend; token metrics carry an explicit main-loop-only warning), classifies failures by `error_kind` (session-error / verify-failed / infra-error) and excludes them from efficiency medians, **verifies from the session transcript that each arm actually invoked its skill** (fail-closed), treats a one-run resolution edge as noise, and rejects candidate overlays no arm exercises. Free-model routing supported (`free-model.litellm.yaml`), so benchmarking needn't burn paid tokens.

## Self-evolution (`workflow_bench.evolve`)

The benchmark's candidate/gate machinery now closes into a loop. `uv run python -m workflow_bench.evolve --tasks … --model <pinned> --generations N` runs: **proposer** (a headless session in a throwaway clone reads the incumbent skills, the prior generation's loser rows, their transcripts/patches, and the learning queue, then writes ONE bounded candidate overlay plus a reviewer-facing `proposal.md`) → **paired benchmark** → **deterministic gate**. `promote` stops the loop; `--apply` copies the overlay onto the canonical skills and all shipped mirrors as an ordinary working-tree diff — committing, CI, and the merge stay human. `keep_incumbent` feeds that generation's trajectories to the next proposer. The overlay trust boundary is unchanged (`candidate_overlay_files` re-validates before every consume), and each family skill gained a "Skill feedback" section routing live-task friction into a gitignored `learnings.jsonl` queue instead of self-edits — a learning reaches a shipped skill only by beating the incumbent on the paired benchmark.

## Review agent (`gitnexus-review-agent.yml`)

Commenting `@gitnexus review` on a PR (maintainers only; the action re-validates write access) runs the `gitnexus-review` skill headlessly on a GitHub runner and posts the graph-backed review as a sticky comment — remote triggering with no local setup, plus a `workflow_dispatch` lane for the Actions UI. Uses the repo's existing `CLAUDE_CODE_OAUTH_TOKEN` secret (same as `claude.yml`) — no new secret needed. Read-only by construction: `contents: read` + `pull-requests: read` token (only `issues: write` for posting the comment), Write/Edit/web tools disallowed, `analyze` resolves ONLY through `npx gitnexus` (npm registry) never the checked-out tree, and `gh api` is excluded from the allowlist — closing a fork-PR confused-deputy path a live end-to-end test run against this PR surfaced (a fork could otherwise smuggle an executable `.gitnexus/run.cjs` into a job holding write-scoped secrets). `issue_comment` workflows run from the default branch, so the comment trigger activates on merge; `workflow_dispatch` needs the same.

## Distribution (`npm i -g gitnexus`)

The family ships in the npm package (`gitnexus/skills/`, installed to every editor target by `gitnexus setup`) and the Claude Code plugin (+ per-skill `mcp.json`); `gitnexus-review` also has a standalone Cursor mirror. `shipped-skills-sync.test.ts` keeps all copies byte-identical to the canonical `.claude/skills/` family, and the new `skill-sync.yml` workflow runs it on exactly the guarded paths — closing the md-only-edit blind spot.

## Registration

`.gitignore` whitelists the four skill dirs and un-ignores `docs/plans/` (plan output travels with the work it drives); `AGENTS.md` 1.13.0 / `CLAUDE.md` 1.7.0 with headers matching their changelogs; per-skill READMEs carry Codex install instructions, and lfg's threshold-governance policy lives in its README rather than the shipped prompt.


magyargergo:
## Workflow benchmark — ground base results (2026-07-11)

First measured ground base for the skill family (`eval/workflow_bench/`, real headless Claude Code 2.1.207 sessions, default model, fresh detached worktrees, n=1/cell). **Every arm resolved every task**, so at this difficulty the comparison is pure cost:

| task (class) | arm | resolved | cost $ | wall | turns | vs baseline cost |
| --- | --- | --- | --- | --- | --- | --- |
| trivial-version-alias | workflow | ✅ | 9.16 | 16m | 63 | −333% |
| trivial-version-alias | baseline | ✅ | 2.11 | 2.8m | 16 | — |
| inv-bug-pdg-note | workflow | ✅ | 14.56 | 21m | 83 | −331% |
| inv-bug-pdg-note | workflow_direct | ✅ | 5.23 | 7.5m | 32 | −55% |
| inv-bug-pdg-note | baseline | ✅ | 3.38 | 4.7m | 22 | — |
| inv-feature-list-repos-filter | workflow | ✅ | 13.22 | 19m | 84 | −211% |
| inv-feature-list-repos-filter | workflow_direct | ✅ | 4.87 | 4.8m | 38 | −15% (wall 14% faster) |
| inv-feature-list-repos-filter | baseline | ✅ | 4.25 | 5.5m | 32 | — |

**Findings, honestly:**

- The full plan→work workflow never amortized its ~$9–11 fixed cost (freshness gate incl. analyzer rebuild + re-index, 13-section plan, work-phase re-anchoring) on tasks a baseline finishes in ≤35 turns — including the investigation-class tasks.
- `workflow_direct` (gitnexus-work without a planning pass) sits near baseline (−15% to −55% cost, once faster wall) and produced visibly more test coverage — the execution discipline is cheap; the planning pass is what's expensive.
- Quality didn't differentiate because nothing failed: pass/fail saturates at this difficulty. The workflow's quality case must show up as resolve-rate deltas on harder tasks — the `cross-module-parse-retry` cell is deliberately still unmeasured.
- Mid-analysis measurement fix: churn now uses `--intent-to-add` + `:(exclude)docs/plans` so non-committing arms don't undercount and the committed plan doc doesn't inflate workflow churn (this run's churn excluded from the record accordingly).

**Provisional routing rule** (recorded in the benchmark README, backing the gates the skills already ship): below this task scale route to `gitnexus-work` direct mode or a plain agent; reserve full `gitnexus-plan` → `gitnexus-work` for cross-module work, multi-session execution where the context pack amortizes, or plan-as-deliverable flows. Also a measured argument for making the plan skill's freshness refresh cheaper.

**Next cells:** cross-module (where baselines should start failing), `--runs 3+` variance on a free backend (`free-model.litellm.yaml`), and a plan-once/execute-twice amortization scenario.

Raw per-session data (session ids, full usage) in the run's `results.jsonl`; methodology and this table recorded in `eval/workflow_bench/README.md` (987fe9ed).

magyargergo:
## Cross-module cell complete — the routing table has its answer

Final cell of the ground base (`cross-module-parse-retry`: bounded retry with backoff across the worker-pool/pipeline seams, transient-vs-deterministic classification; clone-isolated arms, n=1):

| arm | resolved | cost $ | wall | turns | churn |
| --- | --- | --- | --- | --- | --- |
| workflow (plan→work) | ✅ | 18.32 | 37m | 107 | 4/+373/−17 |
| **workflow_direct** | ✅ | **9.53** | **15m** | **52** | 11/+244/−66 |
| baseline | ✅ | 18.03 | 34m | 98 | 6/+345/−69 |

**The measured story across all four task classes:**

- `gitnexus-work` **direct mode is the consistent sweet spot** — near-baseline cost on small/investigation tasks, and on the hardest class it beat a plain agent by **47% cost and 56% wall time**: impact-first navigation and detect_changes-gated commits prevented the flailing that baseline's 98 turns represent. That's the token-savings claim, measured.
- The **full workflow's premium amortizes to ~0 at cross-module scale** (−1.6% vs baseline; −211%…−333% on smaller classes) with a less destructive diff and a durable plan artifact — but with the plan consumed once, it never beat direct mode. The planning pass is bought for its artifact (multi-session reuse, review, handoff), not same-session savings.
- **Resolve rate stayed tied in every cell** — quality differentiation needs harder tasks or n>1.

**Benchmark integrity note:** the first `workflow_direct` run of this cell was invalidated — `git worktree add` shares the repo ref namespace, the workflow arm's slug branch survived worktree removal, and the direct arm found and adopted the finished work (caught by byte-identical churn fingerprints). Arms now run in isolated `git clone --shared` copies (43fad1e8); the table above is the clean re-run. Identical churn across arms is now documented as a contamination alarm.

Full record and methodology: `eval/workflow_bench/README.md` (abd53ab6). Next: n≥3 variance on a free backend, and a plan-reuse scenario to price the amortization case directly.

magyargergo:
We need to do this measurment when there's no changing files in any ways.

zander-raycraft:
This is really good @magyargergo lfg mode is nice! @abhigyanpatwari check this out when you get a chance

magyargergo:
> Skills and prompts expire as models improve; can we build a self-learning loop around agent trajectories?

Implemented a safe prompt-policy loop in f8415442:

- paired `candidate_workflow*` arms apply Markdown-only skill overlays in isolated clones;
- every run records the pinned model, task/base commit, prompt and skill digests, timestamp, and session IDs for trajectory drill-down;
- `promotion.json` uses a deterministic quality-first gate: at least 3 paired runs, no task-level resolution regression, at least 5% median efficiency gain, and a 20% per-task regression cap;
- promotion evidence expires after 90 days and is invalidated earlier by a model or tool-harness change;
- promoted changes still require normal human review and copy synchronization—live runs never rewrite production prompts.

This intentionally stops at offline prompt optimization rather than claiming online model-weight RL. Validation: 11 eval tests, 6 shipped-skill sync tests, Ruff, Prettier, GitNexus TypeScript, and LOW-risk `detect_changes` with zero affected flows.

magyargergo:
## Cross-module benchmark rerun after merging `origin/main`

The routing result flipped on the current model: all three arms resolved, but the skill-disabled baseline won every efficiency metric.

| arm | resolved | cost | wall | turns | output tokens | churn |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| `workflow` | 1/1 | $23.7831 | 37m 28s | 144 | 106,743 | 6 files / +413 / -15 |
| `workflow_direct` | 1/1 | $21.6433 | 38m 04s | 117 | 77,570 | 15 files / +1,073 / -80 |
| `baseline` | 1/1 | $17.9995 | 22m 32s | 83 | 51,943 | 6 files / +487 / -3 |

### Same-run routing result

Compared with `baseline`:

- `workflow`: **+32.1% cost**, **+66.2% wall time**, **+73.5% turns**, and **+105.5% output tokens**.
- `workflow_direct`: **+20.2% cost**, **+68.9% wall time**, **+41.0% turns**, and **+49.3% output tokens**.

The full workflow produced fewer insertions than baseline, but that did not compensate for its token, turn, and latency overhead. Direct routing was especially poor here: it made the broadest patch by far while still costing more and taking longer.

### Directional comparison with the July 11 cell

Relative to [the previous published run](https://github.com/abhigyanpatwari/GitNexus/pull/2431#issuecomment-4945850575):

| arm | cost change | wall-time change | turn change |
| --- | ---: | ---: | ---: |
| `workflow` | +29.8% | about +1.3% | +34.6% |
| `workflow_direct` | +127.1% | about +153.8% | +125.0% |
| `baseline` | -0.2% | about -33.7% | -15.3% |

That historical comparison is directional only because the model, merged tree, and harness/index plumbing changed, and the old wall times were rounded. The within-run comparison above is controlled and is the stronger routing signal.

### Reproducibility and interpretation

- Exact base: `e6ec8cd588cc5ba4c5aa877949f75b33b663ef4b` (the merge of current `origin/main` into this PR).
- Model: `claude-fable-5[1m]`; prompt digest: `da85a08801110fe2ab0e079883ed20fea507620ee0a9da18245bc331b235f49f`.
- Same task prompt and verifier in a fresh clone for each arm; verifier: CLI typecheck plus `parse-retry.test.ts`.
- Each clone received the same schema-v6 GitNexus graph for the exact base. Harness-only dependency/index symlinks were excluded from churn.
- `baseline` disabled skill invocation but retained the same MCP availability. All session IDs and token accounting are retained in the local JSONL artifact.

This is evidence for the prompt-expiry concern behind the PR: a routing rule that helped on the earlier cell is no longer justified on this model/task cell. It should **not** trigger an automatic rewrite from `n=1`, but it does justify the PR's model-pinned, time-bounded evidence gate and a repeatable evolution loop. The next useful extension is to give `gitnexus-pr-review`—the review stage used by `gitnexus-lfg`—the same benchmark, expiry, and candidate-promotion treatment.

magyargergo:
## Family-vs-family benchmark: this PR's skills vs compound-engineering (`ce-*`) — haiku cell

Follow-up to [this morning's fable rerun](https://github.com/abhigyanpatwari/GitNexus/pull/2431#issuecomment-4990713097): same harness, now comparing the GitNexus skill family against the compound-engineering plugin family (v3.9.2) as an external comparator, on `claude-haiku-4-5-20251001`.

**Method.** New runner arms (`34c4e201`): `ce_workflow` (ce-plan → ce-work) and `ce_workflow_direct` (ce-work, no planning pass) mirror the `workflow`/`workflow_direct` prompts verbatim, only the skill family differs; a new review cell runs `gitnexus-review` vs `ce-code-review` on an **identical local diff** (open PR #2198 applied by task setup) with a fresh index built at the pinned base. Every arm runs in its own throwaway shared clone of a dedicated worktree at `437c613c`, n=1 per arm, same task as the previous cells (`cross-module-parse-retry`). Token columns below are **subagent-inclusive** (summed over each session's transcript tree, deduped per API call) because the CLI's `usage` field is main-loop-only — a subagent-heavy skill looks token-cheap in `usage` while its cost lands in `total_cost_usd`. Skill invocation was verified per arm in the transcripts.

### Execution cell (haiku, n=1)

| arm | resolved | cost $ | wall | turns | output tok | cache-create | subagents | churn |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | --- |
| `workflow` (gitnexus-plan → work) | yes¹ | 1.12 | 11.2m | 97 | 31,009 | 173,852 | 0 | 2/+390/−25 |
| `ce_workflow` (ce-plan → ce-work) | yes | 1.45 | 10.0m | 124 | 41,254 | 194,490 | 0 | 4/+574/−8 |
| `workflow_direct` (gitnexus-work) | yes | **0.66** | **3.9m** | 45 | 15,300 | 104,604 | 0 | 3/+481/−1 |
| `ce_workflow_direct` (ce-work) | yes | 0.68 | 6.6m | 44 | 14,141 | 103,257 | 0 | 2/+342/−0 |
| `baseline` (no skills) | yes | 1.21 | 15.5m | 63 | 27,282 | 212,481 | 1 | 4/+460/−7 |

¹ corrected row ([addendum](https://github.com/abhigyanpatwari/GitNexus/pull/2431#issuecomment-4996629319)): the original run implemented but reported done without observing its slow test run — backgrounded tests plus scheduled wakeups that can never fire in a headless `-p` session, while two of its generated real-sleep tests hit the 30s vitest timeout. All four work-arm prompts (both families, symmetric) now require verification output to be observed in the foreground (`0ef545c4`); the re-run verified 20/20 tests + clean `tsc` before declaring done. Original row: no / $1.32 / 16.0m / 135 turns, preserved in `…-haiku-exec2/`.

- **Direct execution is cost-parity between families** (−$0.01), with gitnexus-work 42% faster wall-clock; **both direct arms cost ~45% less than the skill-less baseline** — on haiku the graph-tools discipline beats a plain agent, reversing this morning's fable result.
- **Plan→work: gitnexus resolves at 23.0% lower cost and 24.8% fewer output tokens than ce** (97 vs 124 turns), and is the only plan-based arm under baseline (−7.2%).

### Review cell (haiku, n=1) — identical diff, both reviews substantive

| arm | resolved | cost $ | wall | turns | output tok | cache-create | subagents |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `review` (gitnexus-review) | yes | **0.28** | 6.1m | 29 | 8,252 | 50,131 | 0 |
| `ce_review` (ce-code-review) | yes | 1.01 | 0.8m | 2 | 13,991 | 292,599 | 4 |

**This is the clearest cell for the traversal hypothesis: `gitnexus-review` is 72% cheaper with 41% fewer output tokens and 83% less cache-creation.** The structural difference is visible in the shape: gitnexus-review answers "what does this change break" from the persisted graph in a single loop, while ce-code-review pays 4 persona subagents to re-read the code from scratch (each with its own context). ce's parallel personas do win wall-clock (0.8m vs 6.1m). Both reviews correctly analyzed the #2198 flows-cap change (7.8k vs 12.1k chars).

### Fable context (this morning, base `e6ec8cd5`, n=1)

The same comparison on `claude-fable-5[1m]` couldn't complete: `ce_workflow` burned **$89.44 in ~19 minutes without resolving** — ce-plan ran a 7-agent doc-review panel, ce-work implemented and then launched a 12-agent ce-code-review panel, and the burst tripped the account's 5h session limit, killing the remaining ce arm at $0. The full gitnexus workflow had completed and resolved for **$23.78** ([details](https://github.com/abhigyanpatwari/GitNexus/pull/2431#issuecomment-4990713097)). Subagent-inclusive tokens up to the cutoff: ce plan+work 277,849 output / 2.14M cache-create vs gitnexus plan+work 106,743 / 315.7k — **~2.6× the output tokens and ~6.8× the cache-creation for an unfinished run**. On big models the ce family's fan-out is the dominant cost driver; the gitnexus family stays single-loop by replacing exploratory reading with graph queries.

### Caveats & provenance

- n=1 per arm per cell, one repo, one exec task, one review target; treat magnitudes as directional. Subagent counts on haiku are themselves variance (an earlier haiku exec run saw ce-plan spawn an agent; this one didn't).
- The first haiku exec run was invalidated by a harness bug this exposed: committed example plans in `docs/plans/` tie on clone mtimes and the name-glob handed `gitnexus-work` a stale plan (it correctly concluded that plan was already merged and changed nothing). Plan handoff is snapshot-based as of `34c4e201`; the invalid run is preserved in `eval/results/wfbench-20260716-437c613c-haiku-exec/`.
- Full records (session ids, verify output tails, final patches): `eval/results/wfbench-20260716-437c613c-haiku-exec2/`, `…-haiku-workflow-rerun/` (corrected row), and `…-haiku-review/` on the benchmark host; `ce-lfg`/`gitnexus-lfg` were not run as arms — both are thin pipelines over plan → work → review, and their push/PR/CI stages can't run in throwaway clones.

azizur100389:
﻿Thanks Gergo, I reviewed #2431.

The direction looks valuable, and the skill install/sync paths have decent coverage, but I found a few things I think should be addressed before merge.

Local validation I ran against latest head `34c4e201`:

- `cd gitnexus && npm run build` — passed
- `cd gitnexus-web && npm test` — passed, 384 tests
- `cd eval && .venv\Scripts\python.exe -m pytest tests/test_workflow_bench.py -q` — passed, 11 tests
- `cd gitnexus && npx vitest run test/integration/setup-skills.test.ts test/unit/shipped-skills-sync.test.ts --maxWorkers=1 --testTimeout=120000` — passed, 18 tests
- `cd gitnexus && npm test -- --maxWorkers=2 --testTimeout=120000` — failed locally: 18 files failed, 30 tests failed, 6 worker errors

The full-suite failures look mostly Windows/local-env sensitive rather than obviously caused by this PR: FTS/LadybugDB, CLI subprocess/stdout timeouts, `--skip-git`, bridge cache reopen `EPERM`, Swift scaling timeout, CLI help timeout, MCP startup timeout. Still, I’m recording that local full-suite was not green.

Main concerns:

1. `gitnexus setup` can delete user-customized legacy skill dirs

`gitnexus/src/cli/setup.ts` removes `gitnexus-pr-review` with forced recursive deletion after installing `gitnexus-review`, but it does not verify the old directory is installer-owned. The new test creates arbitrary `"legacy review skill"` content and expects deletion, which suggests a user-customized local skill under Cursor/Codex/etc could be removed during upgrade.

Could we either verify the legacy contents match the shipped legacy copy before deleting, or leave the old directory in place and print a migration warning?

2. `workflow_bench` can accept malformed Claude JSON as valid benchmark evidence

`eval/workflow_bench/runner.py` falls back to `{}` on empty/malformed `--output-format json`, then marks `ok` true if the process exited 0 and `is_error` is absent. If the verifier passes, this can produce a resolved run with zero usage/cost/turns, which would corrupt the benchmark and promotion evidence.

I think this should fail closed when JSON parsing fails or required usage fields are missing.

3. Benchmark trust boundary should be explicit/safe by default

The workflow bench loads task YAML and executes `setup` / `verify` via `shell=True`, while the launched agent defaults to `--permission-mode bypassPermissions` and inherits the full parent environment. That seems acceptable only for trusted local benchmark inputs, but the CLI/docs currently make it easy to run untrusted task files or candidate Markdown overlays with broad host/env access.

Could we add an explicit trusted-input flag / safer default / environment scrubbing, or at least very prominent docs saying these task files and candidate overlays are trusted executable input?

4. Static LiteLLM master key in the example

`eval/workflow_bench/free-model.litellm.yaml` uses `master_key: sk-wfbench`, and the README uses that same token. For a local-only example this is low risk, but if the proxy binds beyond loopback it exposes the configured backend. Suggest sourcing the key from env or documenting `127.0.0.1` binding clearly.

5. Eval tests are not CI-wired

The new `eval/tests/test_workflow_bench.py` passes locally, but I could not find CI wiring that runs the eval pytest suite. Since this PR adds a sizeable Python benchmark/evolution surface, please either wire those tests into CI or explicitly mark the eval harness as not CI-protected.

magyargergo:
Thanks for the thorough review @azizur100389 — all five points are addressed in `ff4b32a3`.

**1. Legacy skill dir deletion → never delete, warn instead.** You're right that the "installer-owned" claim in the comment wasn't enforced in the code. Digest-verifying against the shipped legacy copy isn't actually possible here: `gitnexus-pr-review` was renamed within this PR and has never shipped in any release, so an installer in the field has no authoritative copy to compare against — and a hand-written user skill under that name is indistinguishable from an installed one. So the migration now takes your second option: the legacy directory is left in place untouched and setup prints a migration warning with the exact path. The integration test now asserts the customized legacy dir **survives** with its content intact.

**2. Malformed session JSON → fail closed.** `run_claude` now marks a run `ok: false` whenever the `--output-format json` report is empty, malformed, or missing any required usage field — an exit-0 process can no longer produce a "resolved" run with zero usage/cost that would poison medians or promotion evidence. Covered by 5 new parametrized regression tests (`test_run_claude_fails_closed_on_bad_reports`).

**3. Trust boundary → documented prominently.** Added a **Trust model — task files are executable input** section to the README (shell-executed `setup`/`verify`, `bypassPermissions` default, inherited parent environment, candidate overlays as prompt-injection surface, "read a tasks YAML like a shell script") plus the same warning in the runner's module docstring. I went with docs rather than a trust flag: the same local user supplies both the flag and the file, so a `--trusted` switch adds friction without adding proof — the contract needs to be visible, and now it is. Happy to add env scrubbing as a follow-up if you think docs aren't enough.

**4. Static LiteLLM key → env-sourced.** `master_key` is now `os.environ/LITELLM_MASTER_KEY` with no static default; README/yaml examples generate a random key (`openssl rand -hex 16`) and both now warn to keep the proxy on loopback.

**5. Eval tests → CI-wired.** New `eval-tests` job in `ci-tests.yml` runs the workflow_bench pytest suite on Ubuntu (pinned `setup-python`, installs only `pytest pyyaml` rather than the full eval stack, `persist-credentials: false` like the other jobs). It ran the 16-test suite green locally.

On the local full-suite failures: agreed those match the known Windows/env-sensitive families (FTS/LadybugDB extension loading, CLI subprocess/stdout timeouts, bridge-cache `EPERM` on reopen) — same suites that motivated the Windows shard watchdog widening in #2449. The Linux CI shards are the merge gate for those.

azizur100389:
Rechecked the current head and the original review points look addressed.

I also reproduced one extra local portability issue while rerunning the new eval pytest suite on Windows with Python 3.13: the transcript-verification tests monkeypatched `HOME`, but `Path.home()` on Windows resolves from `USERPROFILE`, so the code looked under the real user profile instead of the temp test home. That made the skill-invocation tests fail locally even though the Ubuntu CI shape is fine.

I pushed a small follow-up fix in `f06f90ab`:

- `eval/tests/test_workflow_bench.py` now sets `USERPROFILE` alongside `HOME` for the fake transcript home used by those tests.
- This keeps the tests isolated on Windows and Unix.

Local validation after the fix:

```text
cd eval
py -3.13 -m pytest tests/test_workflow_bench.py -q
36 passed in 0.85s
```

GitNexus pre-commit scope check for this follow-up: `detect_changes` reports low risk, 1 changed file, 0 indexed changed symbols, and 0 affected execution flows.

azizur100389:
Follow-up: GitHub CI is green on the current head.

All GitHub Actions checks, including CI Gate, CodeQL, quality, eval pytest, coverage shards, platform shards, benchmarks, Docker builds, and shipped-skills drift guard, are passing. The remaining Vercel status is the usual deployment authorization block and is not a code/test failure.

From my side, this is ready for merge.

magyargergo:
@gitnexus review

_(pre-merge trigger test: expected to be a no-op — `issue_comment` workflows only read the workflow file from the default branch, and the dispatch API likewise refuses unregistered workflows (verified 404). The equivalent headless session is being exercised out-of-band against this PR; the comment trigger goes live once this PR merges.)_

magyargergo:
**Update on the `@gitnexus review` trigger test above:** confirmed no-op pre-merge as expected — `issue_comment` workflows only read from the default branch (0 runs of `GitNexus review agent`; the pre-existing `Claude Code` workflow correctly skipped too, since `@gitnexus review` matches neither of its trigger phrases).

To actually exercise the session, I ran the exact workflow prompt/allowlist headlessly against this PR out-of-band (66 turns, full `gitnexus-review` pass with a rebuilt PDG index). It came back **REQUEST CHANGES** and caught a real HIGH-severity issue in the workflow itself: `.gitnexus/` is gitignored but not blocked from being committed, `issue_comment` checks out PR-head content, and the skill's runner ladder tries `node .gitnexus/run.cjs` first — so a fork PR could smuggle its own `run.cjs` into a job holding `CLAUDE_CODE_OAUTH_TOKEN` and a write-scoped `GITHUB_TOKEN`. Fixed in 427fae01: `analyze` now resolves only through `npx gitnexus` (npm registry, never the checked-out tree), `gh api` (not read-only) is off the allowlist, and `pull-requests: write` is downgraded to `read`. Also fixed a null-safety bug the same session's evidence-handling code path could hit. Full context in the commit message.

magyargergo:
## Code Review Results

**Scope:** PR #2431, main -> fork head 427fae01e365736344cedeea4ab3053671378d1e (73 files, 10,062 diff lines)
**Intent:** Ship a cross-CLI GitNexus plan/work/review/lfg skill family, synchronized distribution, a benchmark/evolution harness, installer rename handling, and remote review automation.
**Mode:** markdown report-only

**Reviewers:** correctness and project standards, plus testing, maintainability, agent-native, security, performance, reliability, adversarial verification, and prior-feedback review.

- Testing covered the new Python harness, CI guards, mocks, and TypeScript installer tests.
- Maintainability covered the large structural diff and duplicated distribution surfaces.
- Agent-native covered skills, prompts, MCP configuration, and cross-CLI contracts.
- Security covered fork-triggered workflows, secrets, subprocesses, and LLM-proposed overlays.
- Performance covered paid session fan-out and expensive scenario selection.
- Reliability covered subprocess timeouts, worktrees, failure classification, and exact-head alignment.
- Adversarial verification covered CI and benchmark mechanisms that can silently pass against the wrong behavior.
- Prior-feedback review found no unresolved actionable comment.

### Triage Groups

| Group | Findings | Context | Preferred Resolution | Why |
|-------|----------|---------|----------------------|-----|
| Graph freshness and drift (apply queue) | #3, #12, #13 | Work can trust stale plan evidence or a stale graph between commits and at final refresh | Fix dirty-state re-anchoring and per-step freshness first, then the final analyzer build-current check | One coherent freshness contract protects edits and post-work review evidence |
| Canonical skill distribution (apply queue) | #4, #6 | Duplicate definitions and an incomplete Markdown-only CI path can let shipped instructions drift | Remove nested copies, then run both sync and steering guards | Restore one source of truth before strengthening its guard |
| Remote review workflow (apply queue) | #5, #14 | The action cannot resolve, and manual dispatch cannot align the graph to the requested PR head | Replace the invalid pin, then perform a trusted exact-head checkout before the agent starts | A runnable action is necessary; exact-head alignment makes its output trustworthy |
| Benchmark resource and host safety (apply queue) | #7, #11, #15, #16 | Defaults waste paid runs while proposer and timeout boundaries expose the host | Constrain proposer permissions and process groups, then filter irrelevant arms and expensive tasks | Contain host risk before optimizing cost |
| Promotion gate integrity (apply queue) | #8, #9, #10 | Promotion can succeed after an arm rejects, a transcript is missing, or work never ran | Define evidence per touched skill and require all relevant decisions, invocations, and handoffs | These are three fail-open symptoms of one evidence model |
| Governed workflow reproducibility (apply queue) | #17, #18, #21 | Planning, eval CI, and benchmark commands diverge from written mutation, Python-runner, and model rules | Align all three execution paths with the governing standards | Documented and executed workflows should match |

### P1 -- High

| # | File | Issue | Reviewer | Confidence |
|---|------|-------|----------|------------|
| 3 | .claude/skills/gitnexus-work/SKILL.md:47 | Dirty files bypass the plan drift check | correctness | 100 |
| 4 | .claude/skills/gitnexus/gitnexus-guide/SKILL.md:2 | Nested skills duplicate canonical definitions | maintainability, project-standards | 100 |
| 5 | .github/workflows/gitnexus-review-agent.yml:76 | Pinned action commit does not exist | adversarial | 100 |
| 6 | .github/workflows/skill-sync.yml:59 | Markdown CI skips the steering contract | testing | 100 |
| 7 | eval/workflow_bench/evolve.py:315 | Plan-only evolution runs an untouched direct-mode pair | performance | 100 |
| 8 | eval/workflow_bench/evolve.py:413 | One passing arm can apply a rejected overlay | adversarial, correctness, reliability, testing | 100 |
| 9 | eval/workflow_bench/runner.py:261 | Missing transcripts remain promotion-eligible | adversarial, correctness, reliability, testing | 100 |
| 10 | eval/workflow_bench/runner.py:513 | Workflow can pass without running work skill | adversarial, correctness, reliability | 100 |
| 11 | eval/workflow_bench/tasks.scenarios.yaml:75 | Quick start includes an explicitly opt-in expensive scenario | performance | 100 |
| 12 | .claude/skills/gitnexus-work/SKILL.md:131 | Later steps query the graph before prior commits land | adversarial | 75 |
| 13 | .claude/skills/gitnexus-work/SKILL.md:133 | Post-work refresh can use the old analyzer | correctness | 75 |
| 14 | .github/workflows/gitnexus-review-agent.yml:102 | Manual reviews cannot align to the PR checkout | reliability | 75 |
| 15 | eval/workflow_bench/evolve.py:213 | Poisoned evidence escapes the proposer clone | security | 75 |
| 16 | eval/workflow_bench/runner.py:393 | Timeouts leave descendant processes running | reliability | 75 |

- **#3** - HEAD equality is treated as proof that citations are unchanged even though planning supports uncommitted work. Pin or compare dirty-state evidence and re-read changed cited ranges.
- **#4** - Six nested standard skills duplicate the canonical direct-child paths and already drift. Remove them, move intended edits to canonical copies, and add a resurrection guard.
- **#5** - GitHub returns 404 for the pinned anthropics/claude-code-action commit, so the new workflow stops before review begins. Pin a verified immutable commit and smoke-test both triggers.
- **#6** - The only guaranteed CI path for skill-only Markdown changes runs the byte-sync test but not skills-steering.test.ts. Run both guards.
- **#7** - A plan-only overlay still launches the direct-work incumbent/candidate pair, adding paid sessions that cannot measure the changed prompt. Filter arm pairs by touched skills.
- **#8** - Any promote decision makes the list truthy and --apply copies the full overlay, even when another relevant arm rejects it. Require every arm relevant to each touched skill to promote.
- **#9** - An unlocatable transcript leaves the row successful and valid for promotion. Mark missing invocation evidence ineligible and invert the permissive test.
- **#10** - A missing or overwritten plan skips the work session, but a passing verifier can still mark the workflow resolved. Require plan production and every expected skill invocation.
- **#11** - The documented quick start loads the scenario marked run deliberately, not by default; the PR's own figures imply substantial extra cost and runtime. Add an explicit expensive-task opt-in.
- **#12** - Atomic commits can change callers before later impact checks, while the graph refresh happens only at the end. Check indexed HEAD and refresh before later dependent edits.
- **#13** - Final refresh uses the generic runner ladder without the planner's analyzer build-current check. Rebuild and invoke fresh analyzer output when analyzer source changed.
- **#14** - workflow_dispatch starts from the dispatched ref and lacks checkout/worktree commands needed by the review skill. Resolve and check out the requested PR head in a trusted step before Claude starts.
- **#15** - The proposer receives prompt-derived evidence while using bypassPermissions, the full parent environment, and unrestricted host writes. Use an OS-enforced sandbox, an environment allowlist, and clone-local outputs.
- **#16** - subprocess.run timeouts terminate the immediate process but do not reliably reap descendants. Use dedicated process groups and verify cleanup with a sentinel regression test.

### P2 -- Moderate

| # | File | Issue | Reviewer | Confidence |
|---|------|-------|----------|------------|
| 17 | .claude/skills/gitnexus-plan/SKILL.md:26 | Plan skill permits unregistered repository state changes | project-standards | 100 |
| 18 | .github/workflows/ci-tests.yml:540 | Eval CI bypasses the required uv runner | project-standards | 100 |
| 20 | eval/workflow_bench/evolution.py:19 | Review and LFG cannot enter the evolution gate | correctness, maintainability | 100 |
| 21 | eval/workflow_bench/runner.py:674 | Benchmark defaults to an unnamed model | project-standards | 100 |

- **#17** - The planning skill permits dist rebuilds and learnings.jsonl writes despite the governing planning boundary naming index refresh as the sole extra state change. Remove or explicitly reconcile those writes.
- **#18** - Eval CI installs with pip and runs python -m pytest under eval instead of the repository-required uv run path. Use a pinned uv setup and the documented runner.
- **#20** - Decision gate: either add paired candidate lanes for review and LFG, or narrow feedback, overlay, apply, and evolution claims to plan/work until those lanes exist.
- **#21** - The runner and quick start allow the configured default model, making governed comparisons non-reproducible. Require a named model and record it.

### Actionable Findings

| # | File | Issue | Route | Notes |
|---|------|-------|-------|-------|
| 3 | .claude/skills/gitnexus-work/SKILL.md:47 | Dirty-state drift check | gated_auto -> downstream-resolver | Suggested fix present; verification required |
| 4 | .claude/skills/gitnexus/gitnexus-guide/SKILL.md:2 | Duplicate nested skills | gated_auto -> downstream-resolver | Suggested fix present; verification required |
| 5 | .github/workflows/gitnexus-review-agent.yml:76 | Invalid action pin | gated_auto -> downstream-resolver | Suggested fix present; trigger smoke test required |
| 6 | .github/workflows/skill-sync.yml:59 | Missing steering guard | gated_auto -> downstream-resolver | Suggested fix present; CI verification required |
| 7 | eval/workflow_bench/evolve.py:315 | Irrelevant paid arm pair | gated_auto -> downstream-resolver | Suggested fix present; selection test required |
| 8 | eval/workflow_bench/evolve.py:413 | Partial promotion applies full overlay | manual -> downstream-resolver | Per-skill evidence mapping required |
| 9 | eval/workflow_bench/runner.py:261 | Missing transcript fails open | gated_auto -> downstream-resolver | Suggested fix present; gate test required |
| 10 | eval/workflow_bench/runner.py:513 | Missing work session still resolves | gated_auto -> downstream-resolver | Suggested fix present; workflow test required |
| 11 | eval/workflow_bench/tasks.scenarios.yaml:75 | Expensive task included by default | gated_auto -> downstream-resolver | Suggested fix present; task-selection test required |
| 12 | .claude/skills/gitnexus-work/SKILL.md:131 | Inter-step stale graph | gated_auto -> downstream-resolver | Suggested fix present; multi-step scenario required |
| 13 | .claude/skills/gitnexus-work/SKILL.md:133 | Final refresh can use stale build | gated_auto -> downstream-resolver | Suggested fix present; analyzer scenario required |
| 14 | .github/workflows/gitnexus-review-agent.yml:102 | Manual lane reviews wrong checkout | gated_auto -> downstream-resolver | Suggested fix present; exact-SHA smoke test required |
| 15 | eval/workflow_bench/evolve.py:213 | Proposer is not host-isolated | manual -> downstream-resolver | Sandbox design and canary test required |
| 16 | eval/workflow_bench/runner.py:393 | Timeout leaks descendants | gated_auto -> downstream-resolver | Suggested fix present; cleanup test required |
| 17 | .claude/skills/gitnexus-plan/SKILL.md:26 | Planning mutation boundary mismatch | gated_auto -> downstream-resolver | Suggested fix present; resync mirrors |
| 18 | .github/workflows/ci-tests.yml:540 | Eval job does not use uv | gated_auto -> downstream-resolver | Suggested fix present; rerun eval CI |
| 21 | eval/workflow_bench/runner.py:674 | Model is not required | gated_auto -> downstream-resolver | Suggested fix present; CLI test required |

### Agent-Native Gaps

- The agent-native pass independently corroborated #9 and #20.
- Residual: exact-SHA worktree creation is not consistently paired with an explicit repo selector on every graph query; verify graph identity before enabling remote review.
- The Codex and Cursor copies are synchronized, but no runtime smoke fixture exercises those entrypoints end to end.

### Coverage

- Scope was a remote fork diff. The prescribed origin fetch could not obtain the PR head, so reviewers used the complete GitHub diff; the local untracked gitnexus-2392-fts-repro/ directory was excluded.
- Independent cross-model adversarial review: not run for remote-only scope; the in-process adversarial fallback ran.
- Validator shortcut: 0. One independent batch checked 21 findings; 17 validated and 4 were dropped (#1, #2, #19, #22). No blocker is validation-degraded.
- Preliminary fast-pass items withdrawn: 1. The validator could not prove that the current .mcp.json path is fork-controlled.
- Mechanics: 0 malformed returns, 0 malformed findings, 0 confidence suppressions, 0 quote-line demotions, and 0 soft-bucket demotions.
- No implementation plan was discovered, so requirements completeness and settlement suppression were not evaluated.
- GitHub reports core tests, eval pytest, typechecks, lint, formatting, CodeQL, actionlint, workflow lint, and skill-sync green. The external Vercel check is failing, and browser e2e was skipped.
- No local PR-head tests were run because this report-only remote review did not check out the fork branch.
- Removable surface: about 620 added lines across 6 duplicate nested skill files in #4, less any intended edits moved to canonical copies.
- Residual risks: moving task refs can weaken same-commit pairing; actual model identity is not compared across sessions; overlay apply is non-atomic and lacks a final digest recheck; directory-level exercise checks can miss unread files; the repaired action must re-establish which checkout supplies .mcp.json.
- Testing gaps: same-HEAD dirty drift, multi-step graph refresh, exact-head workflow triggers, descendant cleanup, real transcript fixtures, overlay symlink/digest tampering, and candidate review/LFG lanes lack end-to-end coverage.
- Run artifacts: /tmp/compound-engineering-1000/ce-code-review/20260718-045637-b4efbbc8

---

### Verdict

> **Not ready**
>
> The new remote review action does not resolve at its pinned commit, its manual lane cannot establish an exact-head graph, and the benchmark can promote evidence without the required skill or despite a rejecting arm. These are merge-blocking behavioral defects even though the existing CI jobs are mostly green.
>
> **Fix order:** #5 and #14 -> #15 -> #8, #9, #10 -> #3, #12, #13 -> #4 and #6 -> remaining P1/P2 items.

### Actionable Findings Recap

| # | Severity | File:line | Title | Class | Suggested fix | Confidence |
|---|----------|-----------|-------|-------|---------------|------------|
| 3 | P1 | .claude/skills/gitnexus-work/SKILL.md:47 | Dirty files bypass the plan drift check | gated_auto | yes | 100 |
| 4 | P1 | .claude/skills/gitnexus/gitnexus-guide/SKILL.md:2 | Nested skills duplicate canonical definitions | gated_auto | yes | 100 |
| 5 | P1 | .github/workflows/gitnexus-review-agent.yml:76 | Unavailable action pin prevents every remote review | gated_auto | yes | 100 |
| 6 | P1 | .github/workflows/skill-sync.yml:59 | Markdown CI skips the steering contract | gated_auto | yes | 100 |
| 7 | P1 | eval/workflow_bench/evolve.py:315 | Plan-only evolution runs an untouched direct-mode pair | gated_auto | yes | 100 |
| 8 | P1 | eval/workflow_bench/evolve.py:413 | One passing arm can apply a rejected overlay | manual | yes | 100 |
| 9 | P1 | eval/workflow_bench/runner.py:261 | Missing transcripts remain promotion-eligible | gated_auto | yes | 100 |
| 10 | P1 | eval/workflow_bench/runner.py:513 | Workflow can pass without running work skill | gated_auto | yes | 100 |
| 11 | P1 | eval/workflow_bench/tasks.scenarios.yaml:75 | Quick start includes the opt-in expensive scenario | gated_auto | yes | 100 |
| 12 | P1 | .claude/skills/gitnexus-work/SKILL.md:131 | Later steps query the graph before prior commits land | gated_auto | yes | 75 |
| 13 | P1 | .claude/skills/gitnexus-work/SKILL.md:133 | Post-work refresh can use the old analyzer | gated_auto | yes | 75 |
| 14 | P1 | .github/workflows/gitnexus-review-agent.yml:102 | Manual reviews cannot align to the PR checkout | gated_auto | yes | 75 |
| 15 | P1 | eval/workflow_bench/evolve.py:213 | Poisoned evidence escapes the proposer clone | manual | yes | 75 |
| 16 | P1 | eval/workflow_bench/runner.py:393 | Timeouts leave descendant processes running | gated_auto | yes | 75 |
| 17 | P2 | .claude/skills/gitnexus-plan/SKILL.md:26 | Plan skill permits unregistered state changes | gated_auto | yes | 100 |
| 18 | P2 | .github/workflows/ci-tests.yml:540 | Eval CI bypasses required uv runner | gated_auto | yes | 100 |
| 21 | P2 | eval/workflow_bench/runner.py:674 | Benchmark defaults to an unnamed model | gated_auto | yes | 100 |

magyargergo:
Implemented the review findings in replacement PR #2566: https://github.com/abhigyanpatwari/GitNexus/pull/2566

The replacement preserves this PR's full history, adds the reviewed hardening commit, and merges current `main` without a force-push. It covers all 18 actionable findings plus adversarial follow-ups around evidence provenance, exact-head workflow validation, analyzer invalidation, proposer/process isolation, and atomic promotion recovery.

Local validation: build + typecheck pass; changed GitNexus suite 310/310; eval 283 passed / 10 skipped; actionlint and zizmor pass; two follow-up review lanes found no remaining issue. The PR body documents the known container-native full-suite worker-startup limitation for hosted CI to adjudicate.

PR #2434, opened 2026-07-11
fix(cli): make Claude skills discoverable

### LINKED ISSUE #2433: analyze installs Claude Code skills one level too deep (.claude/skills/gitnexus/<name>/), so none of them register
## Bug

`gitnexus analyze` installs its six standard skills to `.claude/skills/gitnexus/<name>/SKILL.md`. Claude Code discovers skills at exactly `.claude/skills/<name>/SKILL.md` — one level below `.claude/skills/`, with **no recursive scan** — so every GitNexus skill lands inert. They are never registered and can never be invoked.

This is the same class of bug as #1098 / #1381 (OpenCode `skill` vs `skills`), but for Claude Code, and it is currently silent: the skills look installed, `analyze` reports them as created, and the generated `CLAUDE.md` instructs the agent to read them — but the agent has no such skills.

## Source

[`gitnexus/src/cli/ai-context.ts:371`](https://github.com/abhigyanpatwari/GitNexus/blob/main/gitnexus/src/cli/ai-context.ts#L371):

```ts
const skillsDir = path.join(repoPath, '.claude', 'skills', 'gitnexus');
//                                                          ^^^^^^^^^^ extra level
```

combined with [`:409`](https://github.com/abhigyanpatwari/GitNexus/blob/main/gitnexus/src/cli/ai-context.ts#L409):

```ts
const skillDir = path.join(skillsDir, skill.name);
```

which yields `.claude/skills/gitnexus/gitnexus-exploring/SKILL.md`.

## Docs

https://code.claude.com/docs/en/skills.md — a skill is a directory directly under `.claude/skills/`, and the **directory name becomes the command**. The only other layout recognized is a nested `.claude/skills/` inside a monorepo package (e.g. `apps/web/.claude/skills/deploy/` → `/apps/web:deploy`), which is not what is produced here. There is no setting that enables recursive scanning of a grouping directory.

## Expected

`.claude/skills/gitnexus-exploring/SKILL.md` → registers as a usable skill.

## Actual

`.claude/skills/gitnexus/gitnexus-exploring/SKILL.md` → never registered, inert markdown.

## Steps to reproduce

1. `gitnexus analyze` in any repo.
2. Open Claude Code in that repo and list available skills.
3. None of `gitnexus-exploring`, `gitnexus-debugging`, `gitnexus-impact-analysis`, `gitnexus-refactoring`, `gitnexus-guide`, `gitnexus-cli` are present.

## Suggested fix

`skill.name` is *already* namespaced (`gitnexus-exploring`, `gitnexus-cli`, …), so simply dropping the extra path segment on line 371 produces correctly-namespaced, discoverable directories with no collision risk:

```diff
-const skillsDir = path.join(repoPath, '.claude', 'skills', 'gitnexus');
+const skillsDir = path.join(repoPath, '.claude', 'skills');
```

The skills table template at [`:168-176`](https://github.com/abhigyanpatwari/GitNexus/blob/main/gitnexus/src/cli/ai-context.ts#L168-L176) needs the same one-segment change, since it currently emits the nested paths into `AGENTS.md` / `CLAUDE.md`.

Worth checking whether `.claude/skills/generated/<name>/` (community skills from `--skills`) has the identical problem — it looks like it does.

## Secondary damage: it silently reverts the user's fix

A user who moves the skills up to the working layout has that fix undone by the next `analyze`: the skills are reinstalled nested, *and* the managed block in `AGENTS.md`/`CLAUDE.md` is rewritten to point back at the dead paths. We hit exactly this and had to add a CI guard to catch the regression.

## Workaround

`gitnexus analyze --skip-skills` — it suppresses both the nested install and the nested-path rows in the generated table (the `skipSkills` branch at `:168` correctly omits them). Users then vendor the six skills themselves at `.claude/skills/gitnexus-*/`.

## Environment

gitnexus 1.6.5, macOS (arm64), Claude Code.

magyargergo:
## Summary

`gitnexus analyze` now installs Claude Code skills where Claude actually discovers them. Standard skills are direct children of `.claude/skills/`, while generated community skills use a collision-safe `gitnexus-area-*` namespace instead of an undiscoverable grouping directory.

Legacy GitNexus-owned outputs are migrated on repeat analysis without clearing the shared skill root or deleting unrelated project skills. Generated routing tables, CLI/help copy, checked-in standard skills, and skip-mode behavior now describe the same filesystem contract.

## Validation

- 68 review-focused unit tests pass, including the 64-character name boundary and preserved nested-skill coverage.
- 91 CLI skill-generation, steering, help, and analyze-bridge tests pass.
- All 25 `skills-e2e` language fixtures pass.
- `npx tsc --noEmit` passes; full-project ESLint reports 0 errors (existing warnings only).
- The broad `npm test` attempt reached 11,905 passing tests but was not green in this container: native worker replacement/startup failures affected 70 unrelated parsing suites. All changed-area suites remained green.
- GitNexus `detect_changes` reports low risk and no affected execution flows.

Fixes #2433.

---

Codex


koriyoshi2041:
I did a raw-checkout verification of the changed CLI help / skill path surface and found one actionable gap: the source files are updated, but the built `dist` artifacts are still carrying the old paths.

Repro on this PR head (`054c5160`):

```bash
cd gitnexus
npx vitest run test/unit/ai-context.test.ts test/unit/skill-gen.test.ts test/unit/skills-steering.test.ts test/unit/skip-git-cli.test.ts test/unit/analyze-gitnexusrc.test.ts test/unit/analyze-no-stats-bridge.test.ts --reporter=dot
```

Result: `91 passed, 1 failed`. The failure is `test/unit/skip-git-cli.test.ts` because `node dist/cli/index.js analyze --help` still prints:

```text
--skip-skills ... .claude/skills/gitnexus/ ... .claude/skills/generated/
```

`rg` confirms the stale strings are still in `gitnexus/dist/cli/index.js`, `gitnexus/dist/cli/i18n/*.js`, and `gitnexus/dist/cli/ai-context.js`.

After running:

```bash
npm run build
npx vitest run test/unit/skip-git-cli.test.ts --reporter=dot
```

that focused test passes (`7 passed`). So the implementation direction looks sound; this just needs the generated build outputs refreshed in the PR, or the dist-backed test adjusted if `dist` is not meant to be committed for this change.

Issue #1381, opened 2026-05-06
Skills are installed to wrong directory (OpenCode)


JustSayBien:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

> Skills installed to: OpenCode skills (7 skills → ~/.config/opencode/skill/)

the `setup` command installs the skills to `~/.config/opencode/skill` but it should be `~/.config/opencode/skills` according to the OpenCode docs: https://opencode.ai/docs/skills/

### Expected behavior

Skills are installed to `~/.config/opencode/skills`

### Actual behavior

Skills are installed to `~/.config/opencode/skill`

### Steps to reproduce

Run `npx gitnexus setup`

rwb-truelime:
Use the rc version, this has been fixed already.

magyargergo:
Apologies for the delay with the release but we found a major bug we must address before the release.

rwb-truelime:
No worries @magyargergo, good luck with squashing that bug.

Issue #1098, opened 2026-04-27
Bug: `gitnexus setup` installs OpenCode skills to wrong directory (`skill` vs `skills`)


rwb-truelime:
## Bug

`gitnexus setup` installs OpenCode skills to `~/.config/opencode/skill/` (singular), but OpenCode only discovers skills from `~/.config/opencode/skills/` (plural). Skills installed by `gitnexus setup` are invisible to OpenCode.

## Source

[`gitnexus/src/cli/setup.ts` line 588-590](https://github.com/abhigyanpatwari/GitNexus/blob/main/gitnexus/src/cli/setup.ts#L588-L590):

```typescript
/**
 * Install global OpenCode skills to ~/.config/opencode/skill/gitnexus/
 */
async function installOpenCodeSkills(result: SetupResult): Promise<void> {
  const opencodeDir = path.join(os.homedir(), '.config', 'opencode');
  if (!(await dirExists(opencodeDir))) return;

  const skillsDir = path.join(opencodeDir, 'skill');  // ← should be 'skills'
```

## OpenCode docs

From https://opencode.ai/docs/skills/ — **Place files**:

> Global config: `~/.config/opencode/skills/<name>/SKILL.md`

**Understand discovery**:

> Global definitions are also loaded from `~/.config/opencode/skills/*/SKILL.md`

There is no `skill` (singular) path in OpenCode's discovery. Only `skills` (plural).

## Fix

Line 590: `'skill'` → `'skills'`
Line 587 JSDoc: `skill/gitnexus/` → `skills/gitnexus/`

PR #2435, opened 2026-07-11
fix(analyze): install Claude Code skills one level up so they actually register

### LINKED ISSUE #2433: analyze installs Claude Code skills one level too deep (.claude/skills/gitnexus/<name>/), so none of them register

ryanchenglin-cyber:
Fixes #2433.

## The bug

`analyze` installs the six standard skills at `.claude/skills/gitnexus/<name>/SKILL.md`.

Claude Code discovers skills at exactly `.claude/skills/<name>/SKILL.md` and does **not** scan recursively ([docs](https://code.claude.com/docs/en/skills.md)). The extra grouping directory means none of the skills are ever registered — they ship as inert markdown, while the `CLAUDE.md` written by the same command tells the agent to go read them.

It fails silently: `analyze` reports the files as created, and they are on disk. Nothing surfaces that they do nothing. Same class as #1098 / #1381 (OpenCode `skill` vs `skills`), but quieter.

## The fix

```diff
-const skillsDir = path.join(repoPath, '.claude', 'skills', 'gitnexus');
+const skillsDir = path.join(repoPath, '.claude', 'skills');
```

Skill names are already namespaced (`gitnexus-exploring`, `gitnexus-cli`, …), so removing the grouping level cannot collide with a user's own skills.

This also makes `analyze` agree with `setup`, which **already** installs these same skills flat — `editor-targets.ts` maps Claude Code to `~/.claude/skills/`, with no `gitnexus/` segment. The two installers in this repo disagreed; the global one was right.

Also in this PR:

- The routing table injected into `AGENTS.md`/`CLAUDE.md` now emits the flat paths.
- **This repo's own 9 dogfooded skills** were nested too, so they were inert for GitNexus's own contributors. Moved up a level. (The `gitnexus-pr-swarm-review` sibling was already flat — that inconsistency is what confirmed the intended layout.)
- `--skip-skills` help text, en/zh-CN i18n, READMEs, and JSDoc updated to match.

## Why it shipped: the test swallowed its own failure

```ts
try {
  const entries = await fs.readdir(skillsDir, { recursive: true });
  expect(entries.length).toBeGreaterThan(0);
} catch {
  // Skills dir may not be created if skills source doesn't exist in test context
}
```

The `catch` discards the assertion, so the test passes whatever happens. It is now a real guard: the six skills must exist at the discoverable layout, **and** the nested directory must not exist.

## Verification

- `test/unit/ai-context.test.ts` + `test/unit/skills-steering.test.ts`: 44/44 pass.
- Full `test/unit` suite: failure count identical to pristine `main` in the same environment (203 = 203). The only differing test (`hooks.test.ts`, "Ladybug DB owner guard") is flaky — it inspects live `ps` output and flips between runs on an unmodified checkout.
- `tsc` and `scripts/build.js` fail in my environment on **unmodified `main`** as well (unbuilt `gitnexus-shared` sibling), so I could not use them as a signal. Nothing I touched is implicated.

## Deliberately not included

`--skills` community skills still install to `.claude/skills/generated/<name>/`, which has the identical discovery problem. Those names are un-namespaced kebab-cased community labels (`auth`, `parser`, fallback `skill`), so flattening them risks colliding with a user's own skills. That needs a naming decision (`gitnexus-gen-<name>`?) which is yours to make, so I left it out rather than guess. Happy to follow up.

## Migration

Repos analyzed with an older version keep a stale `.claude/skills/gitnexus/` directory. It is inert either way, so nothing breaks, but it is dead weight — worth a cleanup step if you want one; I did not add automatic deletion of files under a user's `.claude/`.


magyargergo:
It's been done as part of https://github.com/abhigyanpatwari/GitNexus/pull/2434.

PR #2453, opened 2026-07-13
fix(embeddings): make batch inserts retry-safe

### LINKED ISSUE #2452: [Regression] --force --embeddings fails with duplicate CodeEmbedding primary keys after an interrupted run
### Area

### Area

gitnexus (embedding pipeline / analyze crash recovery)

### Summary

On GitNexus 1.6.9, retrying `analyze --force --embeddings` after a
previous interrupted or partially completed embedding run can fail with
a duplicate `CodeEmbedding` primary key.

`--force` rebuilds the graph but preserves existing embeddings by
default. Partial embedding rows can therefore survive and collide with
the retry because `batchInsertEmbeddings()` currently uses `CREATE`.

This appears to be a regression of the idempotency behavior addressed
by #822 / PR #823.

### Error

```text
Batch execution failed for rows 1-4: Runtime exception:
Found duplicated primary key value
Variable:src/inkmagic/models/plot.py:type:0,
which violates the uniqueness constraint of the primary key column.

CREATE (e:CodeEmbedding {
  id: $id,
  nodeId: $nodeId,
  chunkIndex: $chunkIndex,
  ...
})
```

The final `:0` is the embedding chunk index, so the collision is on an
existing `(nodeId, chunkIndex)` embedding row rather than two source
symbols sharing the same name.

### Expected behavior

An embedding retry should be crash-safe and idempotent:

- detect and reuse unchanged existing embeddings, or
- delete/reinsert the affected `(nodeId, chunkIndex)` rows safely.

`--force` should not require users to know that partial embeddings must
be dropped manually.

### Actual behavior

The pipeline retries `CREATE` for an embedding row left by the previous
partial run, causing a primary-key violation and aborting analysis.

### Steps to reproduce

I‘ve write a reproduce script, please view the attachment.
reproduce it via `python reproduce.py`

[reproduce.py](https://github.com/user-attachments/files/29966850/reproduce.py)

### Environment

- OS: Windows 11 x64
- Node.js: 22.20.0
- GitNexus: 1.6.9
- LadybugDB: 0.18.1
- Embedding backend: local
- Device: DirectML
- VECTOR unavailable; exact-scan fallback enabled

### Logs / screenshots

_No response_

koriyoshi2041:
## Summary
- delete target CodeEmbedding row ids immediately before each batch insert
- keep the delete scoped to the current batch so interrupted retries do not require a manual embedding wipe
- add unit coverage for the row-id delete-before-create path
- add a local type shim for the optional @huggingface/transformers module so tsc still passes when the optional embedding stack is pruned

Closes #2452

## Tests
- npm test -- --run test/unit/embedding-pipeline.test.ts
- npx tsc --noEmit
- temporarily moved node_modules/@huggingface/transformers aside, then ran npx tsc --noEmit
- git diff --check

Note: I also tried npm ci --omit=optional locally to mirror the failed Ubuntu setup path, but local install stopped in @ladybugdb/core prebuild/source-build setup before reaching tsc. The targeted optional-stack absence check above covers the TypeScript failure shown by CI without changing package install behavior.


magyargergo (changes requested):
## Review: PR #2453 — fix(embeddings): make batch inserts retry-safe

Reviewed at head `7d573e7` against base/merge-base `3ee1e2c` (merge-base equals the PR base), in a detached worktree at the exact head SHA.

### Findings

- **[HIGH]** `gitnexus/test/unit/embedding-pipeline.test.ts:810` — the PR head deterministically fails `announces each checkpoint window before mutating any node in that window`: expected `mutationCountsAtWindowStart` `[0, 2]`, got `[0, 4]`. This is a direct interaction with this PR's core change: each batch now issues **two** `executeWithReusedStatement` calls (id-scoped DELETE + CREATE) instead of one, and the test counts raw `stmtCalls.length` at window start. The test arrived on main via `711ff87` (HTTP-generation resumability) and entered this branch through your own merge-of-main head commit — the "npm test" run in the PR body predates that merge. This is exactly why `tests / ubuntu / coverage 3/3` and the CI Gate are red; I reproduced it locally at the head SHA (1 failed / 53 passed in the file). Remediation: update the expectation to `[0, 4]` — or better, count only CREATE statements so the announce-before-mutate invariant stays robust to statement-shape changes — and re-run the suite on the merged head.

- **[LOW]** `gitnexus/src/types/huggingface-transformers.d.ts:1` — the ambient `declare module` **shadows the real package types even when `@huggingface/transformers` is installed**. Verified empirically at head: importing `AutoModel` (a genuine export of the real package) fails typecheck with "has no exported member", proving tsc always compiles against the shim now. The shim precisely covers today's usage in both embedders (`pipeline('feature-extraction', …)`, `FeatureExtractionPipeline`, `ProgressInfo`, `env.allowLocalModels/cacheDir/remoteHost`, `result.data`, optional `dispose`), so nothing breaks — but future upgrades or new API usage will type-check against stale shim shapes instead of the real definitions. Non-blocking; please add a header comment to the .d.ts stating that it shadows the real types (its purpose is optional-dep-pruned CI, per the coverage-shard TS2307 flake) and that any new API surface must be mirrored here.

- Diff: 4 files, +97/−8. Core change: `batchInsertEmbeddings` deletes the exact target row ids (`MATCH (e:CodeEmbedding {id: $id}) DELETE e`) immediately before each batch CREATE, plus an empty-batch early return.
- The fix lands at the right chokepoint: both insert paths — the `runEmbeddingPipeline` batch loop and the `run-analyze.ts:1702` vector-index wipe-and-restore path (dynamic import, invisible to the call graph; verified by source) — route through it, so retry idempotency covers every caller. `id` is the table's primary key (the #2452 error text proves the uniqueness constraint is on `id`), so the delete is an indexed point lookup, not a scan.
- Crash between DELETE and CREATE self-heals: missing rows fail the content-hash reuse check on the next run and are re-embedded.
- `detect_changes(compare, base=3ee1e2c)`: 4 changed symbols, **0 affected execution flows, LOW risk**. `impact(batchInsertEmbeddings, upstream)`: 1 direct caller (`runEmbeddingPipeline`), embeddings module only.

- New unit test pins delete-before-create ordering with exact id params; existing tests correctly narrowed to distinguish nodeId-scoped stale deletes from the new id-scoped row deletes; the fts-repair restore expectation update is correct. `run-analyze-fts-repair.test.ts` passes at head; `embedding-pipeline.test.ts` fails only on the checkpoint-window test above.
- The reproduction path in #2452 (`--force --embeddings` retry after interruption) is addressed by design; no regression test drives the actual duplicate-PK scenario end-to-end, but the unit-level ordering proof plus PK semantics make that acceptable.
- Vercel failure is the pre-existing fork-deploy authorization gate, not this PR.

### Verdict

REQUEST CHANGES — the fix itself is correct and well-scoped; it's blocked only on the checkpoint-window test expectation (one-line update) and a green re-run of the merged head. The LOW shim note is take-it-or-leave-it.

koriyoshi2041:
Fixed the blocking checkpoint test by counting only CREATE statements, so the announce-before-mutate invariant no longer depends on the new delete statement shape. I also added the optional-package shadowing note to the transformers shim. Focused embedding tests pass (54/54); new head is `8a85aa42` and fresh CI is running.

PR #2455, opened 2026-07-13
fix(embeddings): fall back to text-bearing file nodes

### LINKED ISSUE #2454: [Bug] Embeddings skip text-only repositories with no code symbols
## Summary

`gitnexus analyze --embeddings` can produce zero embeddings for documentation, static-site, and configuration-heavy repositories even when their `File` nodes contain useful searchable text.

## Current behavior

The embedding selector queries code-symbol labels. When none exist, it returns an empty selection and never considers persisted text-bearing `File` nodes. Semantic retrieval is therefore unavailable for a repository that indexed successfully but contains no supported code symbols.

## Expected behavior

When and only when the symbol query returns no nodes, select non-empty, non-binary `File` nodes as a fallback. Repositories with code symbols must retain the existing symbol-first behavior and must not add duplicate File embeddings.

## Proposed regression contract

- a text-only repository embeds a non-empty `File` node;
- empty and binary-placeholder File content is excluded;
- a normal code repository does not invoke the File fallback;
- existing symbol selection and counts are unchanged.

## Evidence

A two-file implementation and neutral unit fixture pass all 33 focused embedding-pipeline tests on upstream `c6445096eb3fdfa70bc1d13995918e0cbd783a38`. Fork proof and history: https://github.com/electricsheephq/evaOS-gitnexus/issues/45 and https://github.com/electricsheephq/evaOS-gitnexus/pull/66.

Related embedding issues and PRs (#283, #702, #974, #2047, #2452, #2453) were checked and do not implement this zero-symbol fallback.

100yenadmin:
Closes #2454

## Summary
- use text-bearing `File` nodes only when the normal code-symbol selection is empty
- exclude empty content and the binary-content placeholder
- preserve current symbol-first behavior for normal code repositories

## Why
Documentation, static-site, and configuration-heavy repositories can index useful File content without producing supported code-symbol nodes. Their embedding run currently selects nothing, leaving semantic retrieval unavailable.

## Validation
- `vitest run --project default test/unit/embedding-pipeline.test.ts` (33/33)
- focused Prettier check
- diff whitespace/sanitization checks
- impact analysis: LOW, one direct caller in the Embeddings module

## Provenance
The contribution is based directly on upstream `c6445096eb3fdfa70bc1d13995918e0cbd783a38` and contains only neutral source and tests. Fork proof: https://github.com/electricsheephq/evaOS-gitnexus/issues/45 and https://github.com/electricsheephq/evaOS-gitnexus/pull/66.


100yenadmin:
Current-upstream self-review at `6eb82713`: no additional correctness gap found within this PR’s explicit zero-symbol fallback contract. The branch is current with upstream `main`; File nodes are queried only when supported symbol selection is empty, empty/binary content is excluded, and normal code repositories retain symbol-first behavior. Focused embedding-pipeline proof: 33/33 tests pass. The disposable worktree required `gitnexus-shared` build plus `npm rebuild @ladybugdb/core` after a lifecycle-disabled install before tests could collect.

gitnexus/src/core/embeddings/embedding-pipeline.ts
```diff
@@ -180,7 +180,50 @@ const queryEmbeddableNodes = async (
...
-  return allNodes;
+  return allNodes.length > 0 ? allNodes : queryFallbackFileNodes(executeQuery);
+};
+
+/**
+ * Static and documentation repositories may contain no code symbols while
+ * still persisting useful text on File nodes. Keep File embeddings as a
+ * zero-symbol fallback so code repositories retain symbol-first selection.
+ */
+const queryFallbackFileNodes = async (
```
magyargergo, line 191: The write side is covered here, but the incremental DELETE sweeps in lbug-adapter.ts (embeddableLabelMatch at the filePath scoped deletes, around lines 2187 and 2313) join embedding rows through EMBEDDABLE_LABELS only, which excludes File. In a docs repo a changed file re embeds without deleting its old rows, so duplicates accumulate on every reanalyze, and a deleted file leaves orphaned rows that eat vector search result slots. Including File in that alternation is harmless for code repos since they have no File embedding rows to match. I can add it when landing.

magyargergo (commented):
Reviewed. The gating is right: the fallback only fires with zero symbol nodes, blank and binary sentinel content is excluded, failures degrade to an empty list, and the code repo test proves symbol first selection never calls the File query. I also verified the read side end to end: semantic search already resolves File embeddings (VALID_NODE_LABELS includes File and local-backend has an explicit File branch), so results actually surface. One real gap inline.

gitnexus/src/core/embeddings/embedding-pipeline.ts
magyargergo: Added in e814c5a1 on the merge train. File is now part of the delete alternation with the rationale documented, and the real database delete suite passes with the widened match.

magyargergo:
Merged via #2512. The delete sweep gap was closed on the train in e814c5a1, so docs repos no longer accumulate duplicate embedding rows on reanalyze. The search side already handled File nodes, nice.

100yenadmin:
Thanks, confirmed the landing fix in `e814c5a1`. I added the missing regression coverage for both batched and singular `File`-owned embedding deletion in #2513; the focused Ladybug integration suite passes 4/4.

PR #2458, opened 2026-07-13
feat(eval): require bearer auth for remote binding

### LINKED ISSUE #2457: security(eval): require authentication for non-loopback server binding
## Problem

`gitnexus eval-server` can bind to a non-loopback address without authentication. That makes the HTTP evaluation surface reachable by other hosts whenever an operator selects `0.0.0.0` or a LAN address.

The default loopback binding is safe, but the CLI currently permits remote exposure without an explicit security gate.

## Proposed contract

- Loopback hosts (`127.0.0.1`, `localhost`, `::1`) may run without credentials.
- Non-loopback startup refuses unless `GITNEXUS_AUTH_TOKEN` is non-empty.
- When a token is configured, every HTTP request requires an exact `Authorization: Bearer <token>` header.
- Compare credentials with `crypto.timingSafeEqual`.
- Return 401 for unauthorized requests.
- Never print the configured credential.

## Acceptance tests

- Non-loopback plus no token exits non-zero before listening.
- IPv4 and IPv6 loopback still work without a token.
- Missing, malformed, or incorrect authorization returns 401.
- The correct bearer token succeeds.
- stdout, stderr, and response bodies do not contain the token.
- Existing loopback workflows remain unchanged.

I found related work around configurable eval-server hosts, but no issue or PR implementing this authentication contract.

100yenadmin:
## Summary

- refuse non-loopback eval-server startup unless `GITNEXUS_AUTH_TOKEN` is configured
- require an exact Bearer credential when authentication is enabled
- compare credentials with `crypto.timingSafeEqual` and keep credentials out of logs and responses
- preserve unauthenticated IPv4 and IPv6 loopback workflows

## Validation

- `vitest run test/unit/eval-server-auth.test.ts` (12 passing)
- focused Prettier check for all six changed files
- full fork CI for the equivalent capability is green across build, typecheck, lint, unit, integration, packaged install, and platform-sensitive jobs

Closes #2457


gitnexus/src/cli/eval-server.ts
```diff
@@ -62,6 +62,40 @@ export function validateHost(raw: string): string | null {
   return null;
 }
 
+/** Resolve the eval-server bearer token without retaining surrounding shell whitespace. */
+export function resolveEvalServerAuthToken(env: NodeJS.ProcessEnv): string | undefined {
```
magyargergo: I think, we need to support reading this from .env or .env.local as well.

gitnexus/src/cli/eval-server.ts
```diff
@@ -62,6 +62,40 @@ export function validateHost(raw: string): string | null {
...
+  return env.GITNEXUS_AUTH_TOKEN?.trim() || undefined;
+}
+
+/** True only for bind hosts that are local to this machine. */
```
magyargergo: This comment is a bit misleading. This also enables anything that is in the local network, because of the IPv4 check. We also should resolve dns that points to a local IPv4.

magyargergo (changes requested):
Please update the relevant readme as well. 🙏

gitnexus/src/cli/eval-server.ts
100yenadmin: Implemented in c9fdab17. The eval server now reads only GITNEXUS_AUTH_TOKEN from the process environment, then .env.local, then .env (shell > .env.local > .env), without importing unrelated file values into process.env. Explicit blank shell values suppress file fallback. Added unit coverage for precedence, fallback, trimming, and explicit blank behavior, plus README documentation.

gitnexus/src/cli/eval-server.ts
100yenadmin: Implemented in c9fdab17. Hostnames are now resolved once to a concrete IPv4 address, and that same address is used for both the loopback/auth decision and listen(), avoiding a validation/bind mismatch. A hostname resolving to loopback remains token-optional; a LAN-resolved hostname requires GITNEXUS_AUTH_TOKEN. I also replaced the misleading comment and updated CLI/i18n help and tests.

100yenadmin:
Addressed the requested changes in c9fdab17 and updated the branch to current upstream main without rewriting history. Added README guidance, targeted .env.local/.env token loading, single-resolution DNS hostname binding, localized help, and regression coverage. Local proof: 78 focused unit/help tests pass, 5 eval-server CLI integration cases pass, TypeScript passes, Prettier passes, and ESLint reports no errors (only the file existing no-explicit-any warnings). @magyargergo ready for re-review.

100yenadmin:
The remaining Gitleaks failure was a historical false positive, not a current-tree credential: commit `c9fdab17` briefly used the literal documentation placeholder `replace-me` in a curl authorization example, and the next commit removed that curl line. Gitleaks correctly continued scanning the branch history.

Fixed at `8a29f3c3` without rewriting branch history or broadening the scanner policy: `.gitleaksignore` contains only the exact reported fingerprint (`c9fdab17...:README.md:curl-auth-header:348`) and documents why it is safe.

I downloaded and checksum-verified Gitleaks `8.24.3`, the same version shown in the failed job, then reran the CI-equivalent first-parent range through the new exact head. Result: 4 commits scanned, the one fingerprint skipped, no leaks found, exit 0.

GitHub CI has restarted on this exact head.

gitnexus/src/cli/eval-server.ts
```diff
@@ -62,6 +66,90 @@ export function validateHost(raw: string): string | null {
...
+    return null;
+  }
+}
+
+function readAuthTokenFile(filePath: string): string | undefined {
+  try {
+    return parseEnv(readFileSync(filePath, 'utf8')).GITNEXUS_AUTH_TOKEN?.trim() || undefined;
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code === 'ENOENT') return undefined;
+    throw new Error(`Unable to read eval-server authentication from ${filePath}`, { cause: error });
```
magyargergo, line 106: If .env or .env.local exists but cannot be read, say a permission error or it is actually a directory, this throws and the server refuses to start even for a loopback bind that needs no token at all. Fail closed is a fine default, but consider deferring this error until a token is actually required, or documenting the behavior.

gitnexus/src/cli/eval-server.ts
```diff
@@ -62,6 +66,90 @@ export function validateHost(raw: string): string | null {
...
+        label.length <= 63 &&
+        /^[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$/.test(label),
+    );
+}
+
+/** Resolve a DNS bind name once so validation and listen() use the same concrete address. */
+export async function resolveEvalServerBindHost(
+  raw: string,
+  resolveHostname: EvalServerHostLookup = async (hostname) =>
+    (await lookup(hostname, { family: 4 })).address,
```
magyargergo, line 87: Forcing family 4 here means host localhost fails on a machine where localhost only resolves to ::1. Narrow case, and ::1 works as a workaround, so not blocking. Maybe worth a mention in the help text.

magyargergo (commented):
This addresses everything from my earlier review, thanks. I went through the auth core and it holds up: the length equalized constant time compare, the 401 before the idle timer reset so probes cannot keep the server alive, and duplicate Authorization headers get rejected. The single DNS resolution so validation and listen use the same address is a nice touch, and the Object.hasOwn precedence gives the tests clean isolation.

Two small notes inline, neither blocks. One thing for the release notes when this ships: an unauthenticated bind to 0.0.0.0 now exits 1 where it used to start. That is the point of the PR, but existing scripts will notice, so it needs a line.

gitnexus/src/cli/eval-server.ts
magyargergo: Addressed in 277dfbbe on the merge train (#2511). Loopback binds now warn and start when the env file cannot be read; non loopback binds keep the hard failure.

gitnexus/src/cli/eval-server.ts
magyargergo: Help text now notes that hostnames resolve to IPv4, same commit 277dfbbe. Landed via #2511.

magyargergo:
Merged to main via the merge train #2511 (merge commit 2e2db8b7). Both review notes were addressed on the train in 277dfbbe. Thanks for the careful auth work and the quick turnaround on the earlier feedback.

PR #2460, opened 2026-07-13
feat(mcp): add deterministic response budgets

### LINKED ISSUE #2459: feat(mcp): add deterministic response budgets
## Problem

The `query`, `context`, and `impact` MCP tools can return very large formatted responses. Clients can limit their own display, but the server currently has no deterministic response budget, so large repositories can consume excessive model context before the client can intervene.

## Proposed contract

- Add an optional positive-integer `maxTokens` parameter to `query`, `context`, and `impact`.
- Add optional `GITNEXUS_MCP_DEFAULT_MAX_TOKENS` configuration.
- An explicit request value overrides the environment default.
- Apply the budget to the complete formatted response, including next-step hints and the truncation marker.
- Keep truncation deterministic and UTF-8 safe.
- Preserve byte-for-byte current output when no budget is configured.
- Reject malformed, zero, negative, or unsafe-integer budgets.

The budget is an intentionally conservative token estimate rather than tokenizer-specific behavior, so it stays deterministic across clients and models.

## Acceptance tests

- No configured budget preserves current output exactly.
- Explicit request values override the default.
- The complete response fits the configured estimate, including its marker.
- Truncation never splits a UTF-8 code point.
- Small budgets remain valid.
- Budget-only arguments are not forwarded to the backend.
- Errors are also bounded when a valid budget was resolved.

Searches for MCP response budgets, truncation, token limits, and `maxTokens` found no matching issue or pull request.

100yenadmin:
## Summary

- add optional `maxTokens` response budgets to `query`, `context`, and `impact`
- support `GITNEXUS_MCP_DEFAULT_MAX_TOKENS`, with explicit request values taking precedence
- apply deterministic UTF-8-safe truncation to the complete formatted response
- preserve current output exactly when no budget is configured

## Validation

- `vitest run test/unit/mcp-output-budget.test.ts test/unit/server.test.ts test/unit/tools.test.ts` (56 passing)
- focused Prettier check for all six changed files
- equivalent capability passed the full fork CI matrix, including build, typecheck, lint, unit, packaged install, benchmarks, and platform-sensitive jobs

Closes #2459


magyargergo:
Can you please explain the motivation of this PR? 🙏 my understanding is to reduce the response length if it exceeds the limit, so it wont waste tokens?

100yenadmin:
Thanks for asking. The motivation is to give MCP callers an opt-in context guardrail when `query`, `context`, or `impact` produces a large fully formatted payload. Existing pagination and summary controls shape the semantic result, but they do not guarantee a bounded transport response, so an agent can still spend a disproportionate amount of its context window on one tool call.

The updated contract is deliberately conservative:
- no configured budget preserves the existing response byte-for-byte
- explicit positive-integer `maxTokens` wins over the optional global default
- the budget covers the complete formatted response, including hints and errors
- truncation is deterministic and UTF-8 safe
- `maxTokens` is removed before backend dispatch
- the four-bytes-per-token estimate is documented as a transport guardrail, not exact model tokenization or a substitute for pagination

I also added README documentation and a server-level test proving an invalid global default fails before backend execution. Focused result at `0506f91e`: 57 tests pass; typecheck and Prettier pass; ESLint has zero errors (pre-existing warnings only).

gitnexus/src/mcp/output-budget.ts
```diff
@@ -0,0 +1,58 @@
+const BUDGETED_TOOLS = new Set(['query', 'context', 'impact']);
+
+export const MCP_TOKEN_ESTIMATE_BYTES = 4;
+export const MCP_TRUNCATION_MARKER = '\n…';
+
+function parsePositiveInteger(value: unknown, source: string): number {
```
magyargergo, line 6: This accepts numeric strings like '20' where the schema says integer. Harmless leniency, just confirming it is intentional and not an accident of the env var sharing this parser.

magyargergo (commented):
Nice and contained. I like the error path handling: a budget that failed validation never applies, but a backend error after a valid budget gets capped, and both paths have tests. The mid structure JSON truncation is fine as documented, this is a transport cap, not pagination, and the byte identical behavior when unset is what makes it safe.

One tiny question inline. Merge order: this conflicts with #2464 and #2465 on server.ts and with #2462 on tools.ts, so this lands last in the MCP series and will need a rebase.

gitnexus/src/mcp/output-budget.ts
magyargergo: Confirmed as intentional, the env default shares this parser. Landed as is via #2511.

magyargergo:
Merged via #2511, composed with the read only and repository policies in the tool call handler. The numeric string leniency stays as is.

PR #2462, opened 2026-07-13
feat(mcp): normalize impact and context parameter aliases

### LINKED ISSUE #2461: feat(mcp): normalize impact and context parameter aliases
## Problem

MCP clients and wrappers commonly use different field names for the same GitNexus operation. `impact` currently expects `target`, while some callers send `name` or `symbol`; `context` expects `file_path`, while some callers send `file`.

Rejecting these otherwise unambiguous requests creates avoidable integration failures. Silently choosing one value when multiple fields disagree would be unsafe.

## Proposed contract

- `impact` accepts `target`, `name`, or `symbol`.
- `context` accepts `file_path` or `file`.
- Canonical and alias values may be supplied together only when their trimmed values agree.
- Conflicting values return a validation error before repository resolution or backend dispatch.
- Normalize once to the canonical field before single-repository or group dispatch.
- Remove alias fields before downstream dispatch.

## Acceptance tests

- Every alias works independently.
- Agreeing canonical and alias fields are accepted.
- Conflicting fields fail before repository resolution.
- Group-mode forwarding receives only canonical parameters.
- Real-index integration tests prove all aliases reach the expected backend operation.
- Tool schemas advertise the accepted alternatives.

Searches for the exact `impact` and `context` alias combinations found no matching issue or pull request.

100yenadmin:
## Summary

- accept `impact.target`, `impact.name`, and `impact.symbol`
- accept `context.file_path` and `context.file`
- reject conflicting canonical and alias values before repository resolution
- normalize once before single-repository or group dispatch

## Validation

- `vitest run test/unit/tools.test.ts` (32 passing)
- `vitest run test/unit/calltool-dispatch.test.ts` (216 passing)
- `vitest run test/integration/local-backend-calltool.test.ts` (34 passing against LadybugDB)
- focused Prettier and diff checks

The older stacked fork PR had one unrelated typecheck job fail because its historical base lockfile omitted `@huggingface/transformers`; all runtime and platform jobs passed. This branch starts from current upstream and passes a clean local typecheck. Its own current-head CI is the remote acceptance gate.

Closes #2461


100yenadmin:
Self-review follow-up at `3f3494fd`:

- rebased behavior onto current upstream `main` via a normal merge (no history rewrite)
- removed the advertised top-level `anyOf`, which some MCP providers reject
- moved the at-least-one impact target contract into runtime normalization
- preserved valid `target_uid`-only impact calls
- reject empty, non-string, and conflicting aliases before repository resolution
- retain canonicalization before group forwarding

Focused proof: 287 dispatch/schema/real-index tests pass; typecheck and Prettier pass; ESLint reports zero errors (repository baseline warnings remain).

100yenadmin:
The current-head CI failures exposed an internal CLI compatibility case that the MCP-only tests did not cover: Commander materializes omitted optional flags as properties whose value is JavaScript `undefined`. Alias normalization treated those as explicitly supplied invalid values, so ordinary `context` CLI calls returned a validation error before repository resolution.

Fixed at `d4576630`: undefined optional alias keys are now treated as absent. Explicit nulls, non-strings, blanks, and conflicts still fail closed.

Focused proof:
- 289 alias schema, dispatch, and real-index tests
- exact previously failing `cli-limit-e2e` and `skip-git-cli` suites: 23/23 pass
- build and `npx tsc --noEmit`
- focused Prettier and ESLint: zero errors (pre-existing warning baseline only)

GitHub CI has restarted on the repaired exact head.

gitnexus/src/mcp/tools.ts
```diff
@@ -578,7 +590,7 @@ SERVICE: optional monorepo path prefix (case-sensitive path segments). When "rep
           maximum: 3600000,
         },
       },
-      required: ['target', 'direction'],
+      required: ['direction'],
```
magyargergo, line 593: Two sides to this one. Dropping target from required fixes a real bug, the old schema made a target_uid only call invalid. But your reason for avoiding anyOf was that some MCP providers reject combinators, and api_impact already uses a top level anyOf for exactly this pattern, see #2308 and its test in tools.test.ts. So either api_impact breaks those providers too and we should file that, or anyOf is fine and this tool should use it for consistency. Which is it? I can live with the runtime guard either way, I just want both tools telling the same story.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -142,6 +142,61 @@ function resolveAliasString(canonical: unknown, legacy: unknown): string | undef
...
+    const supplied: Array<{ key: string; value: string }> = [];
+    for (const key of keys) {
+      if (!Object.prototype.hasOwnProperty.call(input, key)) continue;
+      const value = input[key];
+      // Internal CLI callers materialize omitted optional flags as undefined.
+      if (value === undefined) continue;
+      if (typeof value !== 'string' || !value.trim()) {
+        return { error: `MCP parameter ${method}.${key} must be a non-empty string.` };
+      }
+      supplied.push({ key, value: value.trim() });
```
magyargergo, line 175: This trims the canonical value too, so target ' validate ' becomes 'validate' even with no alias in play. Almost certainly an improvement, just flagging it as a behavior change on the canonical path.

magyargergo (commented):
The normalizer sits at the right chokepoint, after the group early return but before repo routing, and the test proving groupImpact receives the canonical target is exactly the coverage I wanted. Good catch on the Commander undefined flags regression, that chokepoint touches every CLI command and the dedicated test locks it in.

One design question and one behavior note inline. Merge order: conflicts with #2460 on tools.ts, whoever lands second rebases.

gitnexus/src/mcp/tools.ts
magyargergo: Decision: keep the runtime guard. The combinator concern looks real after all, open PR #2489 proposes removing the top level anyOf from api_impact for the same reason, so the two tools should end up telling the same story from the other direction.

gitnexus/src/mcp/local/local-backend.ts
magyargergo: No change needed, flagged for the record. Landed as is via #2511.

magyargergo:
Merged via #2511. I kept the runtime guard over anyOf; #2489 proposing to drop the api_impact combinator supports that call from the other direction.

PR #2464, opened 2026-07-13
feat(mcp): add fail-closed read-only mode

### LINKED ISSUE #2463: feat(mcp): add fail-closed read-only mode
## Problem

GitNexus MCP exposes read operations alongside mutation-capable operations and raw Cypher. Operators embedding the server in agent workflows need a fail-closed mode where discovery and dispatch agree on a provably read-only surface.

Filtering only tool discovery is insufficient because a client can call an undiscovered name directly. Resources, aliases, group routes, and dynamic dispatch also need the same policy boundary.

## Proposed contract

- Enable with `GITNEXUS_MCP_READ_ONLY=1`; accept only `0`, `1`, blank, or unset.
- Discover only explicitly reviewed single-repository read tools.
- Omit raw `cypher`, `rename`, group mutation tools, and all unknown dynamic tool names.
- Reject omitted tools at dispatch before the backend is called.
- Keep only proven legacy read aliases.
- Omit group resource templates and reject direct group resource reads.
- Reject `@group` tool routing.
- Scrub hidden operations and group routes from generated discovery resources.
- Preserve the current MCP surface byte-for-byte when read-only mode is disabled.

Raw Cypher is excluded even though the current backend attempts to reject writes; this contract does not rely on query-string classification as a transaction-boundary guarantee.

## Acceptance tests

- Discovery contains the exact allowlisted read set.
- Hidden and unknown tool names fail before backend dispatch.
- Raw Cypher fails before backend dispatch.
- Tool and resource group routes are absent or rejected.
- Legacy read aliases still work.
- Generated setup/context resources do not advertise forbidden operations.
- Malformed configuration fails startup.

Searches for a GitNexus MCP read-only mode and the configuration name found no matching issue or pull request.

100yenadmin:
## Summary

- add fail-closed `GITNEXUS_MCP_READ_ONLY=1` mode
- expose only reviewed single-repository read operations
- omit and reject raw Cypher, mutation, group routing, and unknown dynamic dispatch
- apply the same policy to tool discovery, aliases, resource templates, direct reads, and generated discovery text

## Validation

- `vitest run test/unit/mcp-read-only.test.ts` (19 passing)
- focused Prettier, diff, and sanitization checks
- equivalent fork capability passed typecheck, lint, benchmarks, Linux coverage, Ubuntu packaged install, and all jobs that started; its other jobs were interrupted before checkout by a GitHub Actions service outage

This upstream branch intentionally contains only the neutral `GITNEXUS_MCP_READ_ONLY` switch. Product-specific compatibility aliases remain outside this contribution.

Closes #2463


100yenadmin:
Security self-review follow-up at `d4eb7560`:

- fixed a direct-resource bypass where case-varied or user-info group URIs could evade a raw prefix check and still normalize to the `group` host
- group resource enforcement now parses and compares the canonical URL protocol/hostname, with malformed group-shaped inputs failing closed
- added adversarial coverage for canonical, uppercase, and user-info URI forms
- added a compatibility test proving normal mode still discovers and dispatches `cypher`, rename, and group tools
- documented `GITNEXUS_MCP_READ_ONLY` and its fail-closed configuration contract

Focused proof: 71 read-only/server/HTTP tests pass; typecheck and Prettier pass; ESLint reports zero errors (four existing server warnings).

gitnexus/src/mcp/read-only-policy.ts
```diff
@@ -0,0 +1,107 @@
+import type { GITNEXUS_TOOLS } from './tools.js';
+
+type GitNexusTool = (typeof GITNEXUS_TOOLS)[number];
+
+export const MCP_READ_ONLY_TOOLS = new Set([
```
magyargergo, line 5: Signing off on this list as the policy decision: these 13 are the read surface, and anything added to GITNEXUS_TOOLS later stays blocked until someone allowlists it on purpose. That fail closed default for future tools is the right call.

gitnexus/src/mcp/read-only-policy.ts
```diff
@@ -0,0 +1,107 @@
...
+  const repo = properties.repo;
+  if (repo && typeof repo === 'object') {
+    properties.repo = {
+      ...repo,
+      description:
+        'Indexed repository name or path. Group-mode values beginning with @ are unavailable in MCP read-only mode.',
+    };
+  }
+  delete properties.subgroup;
+  delete properties.crossDepth;
```
magyargergo, line 100: You strip these from the advertised schema but dispatch still accepts them. Harmless today since crossDepth only matters on the @group path, which you already reject. A one line rejection at dispatch would make the contract self documenting. Optional.

gitnexus/src/mcp/read-only-policy.ts
```diff
@@ -0,0 +1,107 @@
...
+    // keeps obviously group-shaped malformed inputs fail-closed as well.
+    isGroupResource = /^gitnexus:\/\/group(?:\/|$)/iu.test(uri);
+  }
+
+  if (isGroupResource) {
+    throw new Error('Group resources are not available in GitNexus MCP read-only mode.');
+  }
+}
+
+export function filterMcpReadOnlyResourceContent(content: string, readOnly: boolean): string {
```
magyargergo, line 67: These line regexes will silently stop matching if the resource format drifts. Not blocking since dispatch is the real boundary and a stale mention just leads to a clean error, but add a comment saying the scrub is cosmetic so nobody mistakes it for enforcement.

magyargergo (commented):
I read through the whole policy and the layering is right: filtered discovery, enforcement at dispatch, resource guards, and strict env parsing that fails startup on anything but 0 or 1. I also chased the crossDepth question myself: it is only forwarded inside the @group branch of callTool, which this mode rejects, so there is no cross repo leak through a plain repo call. Testing this surface through a real in memory MCP client is exactly right.

A few notes inline. On merge order, this conflicts with #2460 and #2465 on server.ts and the README table. I want to land this one first, the others can rebase.

gitnexus/src/mcp/read-only-policy.ts
magyargergo: Sign off recorded. Landed via #2511.

gitnexus/src/mcp/read-only-policy.ts
magyargergo: Done in aa0d6ee0 on the merge train. Dispatch now rejects crossDepth and subgroup in read only mode, with tests.

gitnexus/src/mcp/read-only-policy.ts
magyargergo: Documented as cosmetic in aa0d6ee0.

magyargergo:
Merged via #2511 with two small hardenings in aa0d6ee0: dispatch now rejects crossDepth and subgroup in read only mode, and the resource scrub is documented as cosmetic. The 13 tool allowlist is signed off as the policy.

Issue #1308, opened 2026-05-04
Select multiple repositories in mcp tools


mohammad-kasaei-mohaymen:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Problem or opportunity

When the agent is calling mcp tools it may need to search in multiple repositories not just one repo
It is better to have a feature that the repo argument of tools would be able to run on more than one repo
this will reduce the number of calls to the gitnexus
and I think it can help reranking things like that

It is specially useful when we have many indexed repos (imagine 100) and the agent wants to get a value from any repo and find sth

### Proposed solution

In repo parameter of tools (for example query and probably others) it is good to have a list of values inside that

### Alternatives considered

In current setup the agent have to search multiple times and gets many many results

### Acceptance criteria

With a single call search on more than one repo(must have)
A good non functional requirement is that it can run about 10 repos at once (should have)

### Contribution

- [ ] I am willing to open a PR for this (may need design discussion first).

PR #2465, opened 2026-07-13
feat(mcp): enforce repository allowlist and default

### LINKED ISSUE #1203: Mcp server (just serve one repo)
### Area

### Problem or opportunity

When I run mcp I get all the repos inside that but it is better to have an option just to select a set of repos not all of them, because sometimes I don't want to give user access to a proper code base

### Proposed solution

in the command `gitnexus mcp --repos "a,b,c,d"` or any other syntax that make explicit which one is used and just a subset of repos is accessible

### Alternatives considered

Have separate docker images with custom script that expose mcp

### Acceptance criteria

When user wants would be able to run an mcp server just for one or a subset of repos not all of them

### Constraints

I want to have indexed all of my repos but just one of them is used and served in MCP server

### Contribution

- [ ] I am willing to open a PR for this (may need design discussion first).

100yenadmin:
## Summary

- add `GITNEXUS_MCP_ALLOWED_REPOS` for a canonical, trimmed, deduplicated repository allowlist
- add `GITNEXUS_MCP_DEFAULT_REPO` for deterministic implicit selection
- validate names and paths against the registry before any stdio or HTTP server binds
- enforce the same policy across tools, aliases, resources, templates, group routes, session factories, and implicit defaults
- filter `list_repos` before pagination and redact unavailable registry entries from errors

## Behavior

- one allowed repository becomes the implicit default
- multiple allowed repositories require an explicit selection unless a default is configured
- a configured default must be inside the allowlist
- ambiguous names fail startup; unique absolute paths remain supported
- no configuration preserves current behavior
- setting only a default does not restrict explicit repository selections

## Validation

- `vitest run test/unit/mcp-repository-policy.test.ts` (19 passing after final formatting)
- focused Prettier, diff, and sanitization checks
- equivalent fork capability passed typecheck, lint, Linux coverage, packaged install, and most platform jobs; the remaining historical failures were a GitHub Actions service outage before checkout

This addresses the repository-subset request in #1203. Multi-repository fan-out from #1308 is intentionally out of scope.

Closes #1203


100yenadmin:
Completeness self-review follow-up at `575dc681`:

- embedded `/api/mcp` now validates repository policy before route registration instead of deferring invalid configuration to the first request
- `gitnexus serve`, standalone HTTP/SSE, and stdio therefore share the same fail-startup contract
- added a regression proving invalid embedded policy registers no MCP route
- documented allowlist/default resolution, the default-only non-restricting behavior, and every enforced surface

Focused proof: 69 repository-policy/server/HTTP tests pass; typecheck and Prettier pass; ESLint reports zero errors (existing `api.ts` warnings remain). The first local attempt failed before test collection only because this fresh worktree was installed with lifecycle scripts disabled and lacked LadybugDB’s native artifact; `npm rebuild @ladybugdb/core` restored the disposable test dependency and the identical test command then passed.

100yenadmin:
Follow-up from live operator validation: the fail-closed startup behavior works as designed, but a long allowlist containing one stale or ambiguous entry leaves the operator with only the aggregate sanitized error and no safe way to identify the configured item.

I filed #2509 with a neutral, non-leaking preflight contract. This does not need to weaken or block the current policy implementation; it can land in this PR if maintainers prefer, or remain a linked follow-up.

100yenadmin:
Routing correction: the diagnostic tracker is now canonical in the shipping fork at electricsheephq/evaOS-gitnexus#127 because repository policy is not yet on upstream main. I closed upstream issue #2509 to avoid treating proposed behavior as an upstream release defect. This PR remains the upstream contribution surface.

gitnexus/src/mcp/repository-policy.ts
```diff
@@ -0,0 +1,362 @@
...
+    const selected = this.repoForArgs(repo === undefined ? undefined : { repo });
+    return backend.resolveRepo(selected?.path, branch);
+  }
+
+  assertResourceUri(uri: string): void {
+    if (!this.restricted) return;
+    let parsed: URL;
+    try {
+      parsed = new URL(uri);
+    } catch {
```
magyargergo, line 243: When new URL throws you return and allow the read. #2464 handles the same situation with a fail closed regex fallback. Today this is safe because resources.ts parses with the same new URL call, so both sides reject the same inputs, but if either parser changes they can disagree silently. Please mirror the fallback here.

gitnexus/src/mcp/repository-policy.ts
```diff
@@ -0,0 +1,362 @@
...
+          policy.restricted &&
+          (property === 'readGroupContractsResource' || property === 'readGroupStatusResource')
+        ) {
+          return async () => {
+            throw new Error(
+              'Group resources are unavailable when an MCP repository allowlist is set.',
+            );
+          };
+        }
+        const value = Reflect.get(target, property, receiver);
```
magyargergo, line 311: The proxy covers callTool, listRepos, resolveRepo, getContext and the group resource readers, but not queryClusters, queryProcesses, queryClusterDetail and queryProcessDetail. Those are only safe because assertResourceUri runs first in the read handler. Any future readResource caller or new resource route bypasses the policy for those four. Intercepting them here is cheap and makes the proxy self contained.

magyargergo (commented):
Solid piece of work. Startup validation before any bind, sanitized errors that leak nothing about the registry, and I verified all five entry points pass the policy, including the embedded server after your last fix. The bypass guards on direct construction and custom factories are a good instinct.

Two hardening asks inline. One structural question: this overlaps a lot with #2464. Both scrub group text from descriptions, filter templates and guard resources, and they edit the same server.ts regions. I plan to land #2464 first, so please rebase this on top and reuse its helpers rather than keeping two copies of the scrub logic.

One doc note: an allowed repo whose name is duplicated in the registry returns no context resource because of the unique name filter. Consistent with the ambiguity rules, but worth a line in the README.

gitnexus/src/mcp/repository-policy.ts
magyargergo: Done in 48a145a4. Testing this spot also caught a real bypass: gitnexus is a non special URL scheme so the host keeps its case, and GITNEXUS://GROUP/... walked past the case sensitive compare. Fixed with the fallback and a case insensitive host check, both tested.

gitnexus/src/mcp/repository-policy.ts
magyargergo: Done in 48a145a4. The proxy now intercepts queryClusters, queryProcesses, queryClusterDetail and queryProcessDetail.

magyargergo:
Merged via #2511 with hardening in 48a145a4: case insensitive host comparison in assertResourceUri (the new test caught a GITNEXUS://GROUP bypass), a fail closed fallback for unparseable URIs, proxy interception of the four repo scoped query methods, and the scrub helper shared with the read only policy. Good work on this one.

PR #2047, opened 2026-06-04
feat(embeddings): add GITNEXUS_EMBEDDING_OMIT_DIMENSIONS for provider…

CheckPickerUpper:
## Summary

Adds an opt-in `GITNEXUS_EMBEDDING_OMIT_DIMENSIONS` flag so embedding providers that reject the `dimensions` request field (e.g. Voyage) can be used, while the returned vector length is still validated against `GITNEXUS_EMBEDDING_DIMS`.

## Motivation / context

`GITNEXUS_EMBEDDING_DIMS` is overloaded: it is both the `dimensions` field sent in the request and the length the returned vector is validated against. Some OpenAI-compatible endpoints reject the `dimensions` field with a 400 while still returning a fixed-size vector. Voyage, for example, returns `400 "Argument 'dimensions' is not supported by our API"`. That leaves no working configuration: setting DIMS triggers the 400, and leaving it unset makes validation expect the 384 default and reject the provider's real 1024-d vector.

## Areas touched

- [x] `gitnexus/` (CLI / core / MCP server)
- [ ] `gitnexus-web/` (Vite / React UI)
- [ ] `.github/` (workflows, actions)
- [ ] `eval/` or other tooling
- [ ] Docs / agent config only

## Scope & constraints

**In scope**

- New opt-in `GITNEXUS_EMBEDDING_OMIT_DIMENSIONS` env flag (`1`/`true`) read in the HTTP embedding client.
- Suppresses only the sent `dimensions` field, on both the batch (`httpEmbed`) and single-query (`httpEmbedQuery`) paths.
- `GITNEXUS_EMBEDDING_DIMS` is still honoured to validate the returned vector length.
- README env-var docs plus two unit tests.

**Explicitly out of scope / not done here**

- No change to the local (transformers.js) embedding path.
- No provider auto-detection. The flag is explicit and off by default.
- No change to default behaviour: with the flag unset, the request body is byte-for-byte identical to today.

## Implementation notes

`readConfig()` gains an `omitDimensionsField` boolean derived from the env var. Both call sites pass `undefined` for the request dimensions when it is set, so the field is omitted; the existing validation (`config.dimensions ?? DEFAULT_DIMS`) is untouched, so a mismatched response is still rejected.

## Testing & verification

[...]

Issue #283, opened 2026-03-14
gitnexus analyze --embeddings （error undici）


gcggcg:
████████████████████████████████████░░░░  90% | Embedding 0/?node:internal/deps/undici/undici:14902
      Error.captureStackTrace(err);
            ^

TypeError: fetch failed
    at node:internal/deps/undici/undici:14902:13
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at runNextTicks (node:internal/process/task_queues:64:3)
    at listOnTimeout (node:internal/timers:545:9)
    at process.processTimers (node:internal/timers:519:7)
    at async getModelFile (file:///usr/lib/node_modules/gitnexus/node_modules/@huggingface/transformers/dist/transformers.node.mjs:32855:24)
    at async getModelText (file:///usr/lib/node_modules/gitnexus/node_modules/@huggingface/transformers/dist/transformers.node.mjs:32994:20)
    at async getModelJSON (file:///usr/lib/node_modules/gitnexus/node_modules/@huggingface/transformers/dist/transformers.node.mjs:33014:18)
    at async Promise.all (index 0)
    at async loadTokenizer (file:///usr/lib/node_modules/gitnexus/node_modules/@huggingface/transformers/dist/transformers.node.mjs:25947:18) {
  [cause]: AggregateError [ETIMEDOUT]:

naicud:
Check on pr 71 I have tried on local with ollama and  gemmaembeddings all works

ShunsukeHayashi:
This is typically caused by Node.js version incompatibility with the `undici` HTTP client used by `@huggingface/transformers`.

**Fix:**
```bash
# Ensure Node.js 20+ (recommended: 22 LTS)
node --version

# If on Node 18, upgrade:
nvm install 22
nvm use 22

# Then retry:
npx gitnexus analyze --embeddings
```

If you are on Node 22+ and still see this, please share the full error output.

1353604736:
API> gitnexus analyze --embeddings --skills --verbose

  GitNexus Analyzer

  ████████████████████████████████████░░░░  90% | Loading embedding model...

  Analysis failed: Cannot find package 'onnxruntime-common' imported from D:\ProgramData\pnpm\store\v11\links\@huggingface\transformers\4.2.0\a4656ca5db838bdb956f51be7e35730dcc9dd3bb94bf332795ca75a0f760bb1a\node_modules\@huggingface\transformers\dist\transformers.node.mjs
[...]

PR #2468, opened 2026-07-13
fix(embeddings): make HTTP generation resumable

### LINKED ISSUE #2467: Embedding HTTP resilience can silently skip partial multi-chunk nodes on resume
## Problem

HTTP embedding generation currently treats transient endpoint failures as terminal, has no provider-neutral request pacing controls, and cannot reliably cancel in-process embedding work. Long runs can therefore fail late on a short-lived 429/5xx/transport error or continue holding work after an API job is cancelled.

There is also a resume-integrity gap. A node can produce multiple embedding chunks. If the process stops after only some chunks are inserted, any persisted chunk already carries the current node content hash. A subsequent run may treat that matching hash as proof that the whole node is complete and skip the missing chunks.

## Expected behavior

- Retry transient transport errors, 429, and 5xx responses with bounded total attempts.
- Honor `Retry-After`, but cap the wait with a configurable upper bound.
- Optionally pace requests with a provider-neutral minimum interval.
- Propagate cancellation through queued pacing, retry sleeps, fetch timeouts, local embedding batches, and server embedding jobs.
- Persist a bounded pending-node window before any embedding mutation.
- After the database checkpoint, clear the pending window while retaining progress.
- On restart, delete and regenerate every pending node even when a persisted chunk has the current content hash.
- Fail before mutation if the checkpoint model or dimensions do not match the current runtime.
- Keep repository locks held until cancelled work has actually stopped.
- Redact endpoint credentials and API keys from surfaced transport errors.

## Proposed configuration

- `GITNEXUS_EMBEDDING_MAX_ATTEMPTS` (positive integer, default 3, maximum 20)
- `GITNEXUS_EMBEDDING_RETRY_CAP_MS` (positive integer, default 5000)
- `GITNEXUS_EMBEDDING_MIN_INTERVAL_MS` (non-negative integer, default 0)

Provider-specific rate limits remain configuration, not built-in policy.

## Acceptance tests

- Retryable status and transport failures stop at the configured attempt bound.
- `Retry-After` and exponential backoff cannot exceed the configured cap.
- Pacing is deterministic under concurrent callers and cancellation interrupts both the queue and sleep.
- Error messages and causes contain neither URL credentials nor the API key.
- A pending node with a matching hash is deleted and fully re-embedded.
- Pending nodes that disappeared or became non-embeddable have residual rows deleted idempotently.
- The pending window is saved before its first mutation and cleared only after a database checkpoint.
- CLI and server embedding paths both consume and maintain the marker.
- Timeout cancellation does not release the repository lock before work unwinds.

## Related work, not duplicates

- #283 covers local Hugging Face model-download failures, not the configured HTTP embedding endpoint or durable resume.
- #2452 / #2453 cover retry-safe database row insertion, not HTTP request resilience or partial multi-chunk resume.
- #2047 covers optional provider dimensions.
- #974 combines backend selection and retrieval/ranking changes and is currently conflicting; it does not provide this bounded retry and pending-window contract.

100yenadmin:
## Summary

- add provider-neutral HTTP embedding retry, Retry-After capping, request pacing, cancellation, and credential-safe errors
- persist bounded pending-node windows so interrupted multi-chunk nodes cannot be mistaken for complete work
- resume pending nodes idempotently in both CLI analysis and the server embedding endpoint
- keep repository locks held until cancelled server work has actually stopped
- document the three generic resilience settings

## Failure mode fixed

Each persisted chunk carries the node content hash. Previously, an abrupt stop after a partial multi-chunk insert could leave a current hash in the database; restart then skipped the node and silently retained missing chunks. This change records the whole bounded window before mutation, checkpoints LadybugDB at the window boundary, and forces all still-pending nodes through delete plus regeneration on resume.

## Configuration

- `GITNEXUS_EMBEDDING_MAX_ATTEMPTS` (default 3, maximum 20)
- `GITNEXUS_EMBEDDING_RETRY_CAP_MS` (default 5000)
- `GITNEXUS_EMBEDDING_MIN_INTERVAL_MS` (default 0)

Explicit configuration wins; no provider-specific defaults are introduced.

## Validation

- `npm run build`
- `npx tsc --noEmit`
- `npm test -- --run test/unit/embedding-pipeline.test.ts test/unit/run-analyze.test.ts test/unit/http-embedder.test.ts test/unit/integrations/resilient-fetch.test.ts test/unit/analyze-job.test.ts test/unit/api-readonly-wiring.test.ts` (181 tests)
- `git diff --check`

The tests cover transient statuses, transport failures, capped Retry-After, deterministic pacing, cancellation, secret redaction, checkpoint-window ordering, matching-hash forced resume, disappeared pending nodes, model/dimension mismatch, server cancellation, and lock lifetime.

## Scope

This is intentionally separate from local Hugging Face download failures (#283), retry-safe database insert semantics (#2452 / #2453), optional dimensions (#2047), and backend/ranking policy (#974).

Closes #2467


100yenadmin:
Adversarial resume review follow-up at `1821b01d`:

- checkpoint identity now includes `local` or a secret-free SHA-256 fingerprint of the canonical HTTP endpoint identity
- resume fails before mutation if the endpoint/provider changes even when model name and dimensions happen to match
- URL credentials and query parameters are removed before fingerprinting; metadata contains no endpoint or credential-derived hash
- model-only env configuration no longer mislabels local embeddings as HTTP embeddings
- CLI analyze and the server embedding route use the same identity helper
- checkpoints from an unidentifiable older shape fail closed with the existing drop/rebuild recovery path

Focused proof: 182 retry/pacing/cancellation/checkpoint/job/analyze tests pass; typecheck and Prettier pass; ESLint has zero errors (repository baseline warnings remain).

gitnexus/src/core/run-analyze.ts
```diff
@@ -1826,6 +1920,21 @@ export async function runFullAnalysis(
         {},
         cachedEmbeddingNodeIds.size > 0 ? cachedEmbeddingNodeIds : undefined,
         existingEmbeddings,
+        {
+          forceReembedNodeIds: pendingEmbeddingNodeIds,
+          onCheckpointWindowStart: async ({ nodeIds, ...checkpoint }) => {
+            await saveEmbeddingCheckpoint(checkpoint, nodeIds, existingMeta?.stats?.embeddings);
+          },
+          onCheckpoint: async (checkpoint) => {
+            await checkpointOnce();
```
magyargergo, line 1929: Confirmed checkpointOnce runs a real checkpoint on every call (the once contrasts with the periodic driver), so the durability step holds per window. One nuance for the record: under GITNEXUS_WAL_MANUAL_CHECKPOINT=0 this no ops, so pending ids clear without the consolidation step. WAL replay still protects the written rows, so I think that is acceptable, just worth knowing.

gitnexus/src/core/run-analyze.ts
```diff
@@ -1735,7 +1781,7 @@ export async function runFullAnalysis(
     if (shouldGenerateEmbeddings) {
       const { skipForCap, capDisabled, nodeLimit } = deriveEmbeddingCap(
         stats.nodes,
-        options.embeddingsNodeLimit,
+        resumeEmbeddingCheckpoint ? 0 : options.embeddingsNodeLimit,
```
magyargergo, line 1784: Checked deriveEmbeddingCap: zero disables the cap, so resume deliberately bypasses it. Coherent, since a capped skip run never writes a checkpoint in the first place. Confirming this is intentional.

magyargergo (commented):
Reviewed the whole thing and this is solid work. The write ahead pending window protocol closes exactly the hole from #2467: pending node ids are persisted before any node in the window mutates, cleared after the durable checkpoint, and resume force re embeds them even when the persisted hash matches. The test that asserts each window is announced before any mutation is the one that pins the property, good.

Also verified along the way: the identity fingerprint keeps secrets and endpoints out of metadata, the early return fast path no longer swallows a pending checkpoint, and moving the /api/embed lock release into finally plus routing the timeout through cancelJob fixes a pre existing early release bug. The sanitizeReason change is a real security fix too, the raw cause could previously carry the API key into logs.

On scope: this includes the resilience knob half that the issue over claimed (the retry machinery already existed via resilientFetch), but it is built as extensions to that machinery rather than a reimplementation, with strict env validation and fake timer tests, so I am fine with it.

Sequencing: this conflicts with #2455 on embedding-pipeline.ts and its test file. #2455 is tiny, so that lands first and this rebases.

gitnexus/src/core/run-analyze.ts
magyargergo: Noted for the record and landed as is via #2512.

gitnexus/src/core/run-analyze.ts
magyargergo: Confirmed intentional and landed as is via #2512.

magyargergo:
Merged via #2512, no changes needed. The pending window protocol, the identity guard and the two bonus fixes (cause chain key redaction and the lock release in finally) all held up under review and the combined suites pass with #2455 stacked underneath. Strong work.

PR #2470, opened 2026-07-13
fix(cli): fail cypher errors loudly

### LINKED ISSUE #2469: CLI: cypher backend errors exit with status 0
## Summary

`gitnexus cypher` prints backend error payloads but exits with status 0. Shell automation therefore cannot distinguish an invalid query from a successful query without parsing stdout.

## Reproduction

Run a query that references a missing relationship property against any valid index:

```sh
gitnexus cypher "MATCH ()-[r:CodeRelation]->() RETURN r.missing" -r <repo>
echo $?
```

Observed output is a structured binder error such as:

```json
{"error":"Prepare failed: Binder exception: Cannot find property missing for r."}
```

Observed status is `0`.

## Expected

The structured error should still be printed, but the command should exit non-zero. Successful array and tabular results should remain exit zero, and `--limit` behavior should be unchanged.

## Proposed test

Mock `LocalBackend.callTool` to return `{ error: "..." }`, invoke `cypherCommand`, and assert `process.exitCode === 1`; retain a successful-result control.

100yenadmin:
Fixes #2469.

## Summary

- keep printing the existing structured Cypher backend error payload;
- set `process.exitCode = 1` when that payload contains a non-empty error string;
- preserve exit-zero behavior for successful tabular results and existing `--limit` handling.

## Tests

- added error and success controls to `tool-direct-cli.test.ts`;
- focused suite: 12 passed;
- `git diff --check`: passed.

The change is limited to the direct CLI wrapper and does not alter MCP behavior or database execution.


100yenadmin:
Current-upstream self-review at `5d1911b2`: the backend Cypher contract consistently returns non-empty string `error` payloads for validation, read-only, WAL, and query failures, so the exit-code check covers the false-success class without changing successful output or resetting a pre-existing process exit code. Focused direct-CLI proof: 12/12 tests pass.

gitnexus/src/cli/tool.ts
```diff
@@ -301,6 +301,15 @@ export async function cypherCommand(
...
   }
   output(result);
+  if (
+    result &&
+    typeof result === 'object' &&
+    'error' in result &&
+    typeof result.error === 'string' &&
+    result.error.trim().length > 0
+  ) {
+    process.exitCode = 1;
```
magyargergo, line 311: tool.ts has seven output(result) call sites sharing the same backend error payload contract, and query or context have the same silent exit today. Rather than repeating this block per command, the one guard that fixes them all is inside output() itself (or a small wrapper both use). Happy to widen it that way when landing so the siblings get the semantics at once.

magyargergo (commented):
The guard is correct and the two polarity tests are right. The two failing checks are not this PR: coverage merge died on a missing optional @huggingface/transformers module during a tsc step (runner infra) and Vercel is the usual fork authorization gate. One scope note inline.

gitnexus/src/cli/tool.ts
magyargergo: Widened in e20e3262 on the merge train. The guard now lives in output() so all seven commands share the semantics, with query and context cases added. One existing test relied on exit zero for a not found payload and was updated to pin the new contract in 36d25b5a.

magyargergo:
Merged via #2512. The guard was widened into output() in e20e3262 so every tool command shares the exit semantics, and one existing test that relied on exit zero for a not found payload now pins the new contract.

PR #2476, opened 2026-07-13
fix(cache): bound parsedfile generations

### LINKED ISSUE #2475: Durable ParsedFile cache accumulates stale worker shard generations
## Summary

Durable ParsedFile cache directories are keyed by chunk content hash, but cache-miss regeneration appends worker shard names without clearing the previous complete generation. Repeated analyses can therefore grow one supposedly content-addressed chunk directory without bound.

## Public scale evidence

A disposable OpenClaw index using two workers had:

- 41 live chunk hashes;
- 357 durable shard files;
- 22,544 ParsedFile rows for 18,046 unique paths;
- 4,498 extra copies;
- 6 to 13 shards in sampled chunk directories.

All duplicate ParsedFiles were byte-identical in that run, so the demonstrated impact is unbounded disk use and replay work rather than data corruption.

## Root cause

`persistDurableParsedFileShardSync` names shards with chunk hash, worker thread id, and sequence. A later cache-miss generation can use different worker/sequence assignments, leaving old names in the same chunk directory. Pruning removes dead chunk directories but does not remove stale shard files inside a still-live hash.

## Proposed contract

- The main thread clears and recreates a chunk directory once before dispatching a cache miss.
- Workers only write after preparation and never delete the shared directory.
- Directory recreation preserves worker-side mkdir memoization safety.
- The parse-cache schema version bumps once to invalidate existing accumulated stores.
- Tests cover both the storage seam and two repeated cache misses through the dispatcher.

100yenadmin:
Fixes #2475.

## Summary

- prepare a fresh durable ParsedFile directory once before workers dispatch a cache-miss chunk;
- preserve concurrent worker writes and worker-side directory memoization;
- bump the parse-cache schema so existing accumulated directories are invalidated once;
- add storage-level and dispatcher-level generation replacement tests.

## Tests

- 10 focused parse/cache/worker suites;
- 61 tests passed;
- repository formatting and `git diff --check` passed.

The change is limited to durable parse-cache lifecycle. It does not alter source parsing, symbol resolution, graph schema, or warm-cache hit behavior.


100yenadmin:
Current-upstream self-review at `a6441b34`: no additional gap found. Generation cleanup happens only on a cache miss, before worker dispatch; recreating the directory keeps the per-process mkdir memo valid, a crash cannot poison a later miss because the next miss prepares again, and warm hits remain read-only. The schema bump invalidates previously accumulated generations once. Focused durable ParsedFile/warm-cache proof: 10/10 tests pass.

gitnexus/src/core/ingestion/pipeline-phases/parse-impl.ts
```diff
@@ -984,6 +985,9 @@ export async function runChunkedParseAndResolve(
         // Cache miss: dispatch to workers, capture the raw results, store
         // them under the chunk hash for the next run.
         chunkCacheMisses++;
+        if (durableParsedFileDir !== undefined && chunkHash !== null) {
+          await prepareDurableParsedFileChunk(durableParsedFileDir, chunkHash);
```
magyargergo, line 989: This await is unguarded, so an fs failure here (read only cache dir, permission problem) kills the whole analyze over a cache optimization. The neighboring durable store calls (restoreDurableParsedFileShard, loadDurableParsedFileIndex) swallow fs errors and degrade. Workers recreate the directory on write anyway, so a try catch with a warn here degrades to at worst the old appending behavior. I can add this when landing.

magyargergo (commented):
Reviewed. The fix shape is exactly what the issue needed: one generation per chunk directory, replaced on the miss path before any worker writes, and the schema bump flushes the stores that already accumulated shards. The bump also retires the cross version stale shape hazard where an old generation shard could win the last writer race after an upgrade. Tests at both seams (storage and through the dispatcher) are the right pair. One hardening note inline.

gitnexus/src/core/ingestion/pipeline-phases/parse-impl.ts
magyargergo: Added in ee7161ef on the merge train (#2512). The reset now warns and continues, with a regression test I verified fails on the unguarded code.

magyargergo:
Merged via the merge train #2512 (merge commit c4fb511a). The fault tolerance note was addressed on the train in ee7161ef. Clean fix, thanks.

PR #2478, opened 2026-07-13
fix(communities): canonicalize projection order

### LINKED ISSUE #2477: Community projection is insertion-order dependent despite seeded Leiden RNG
## Summary

Community detection uses a fixed Leiden RNG seed, but the projection currently preserves `KnowledgeGraph` insertion order for nodes and first-seen order for edges. Equivalent graph sets reconstructed through cold parsing, warm cache replay, or incremental/full paths can therefore receive different numeric node indexes despite the fixed seed.

## Expected contract

- Equivalent symbol and clustering-edge sets produce the same projection regardless of node insertion order, relationship insertion order, or undirected edge direction.
- Projection nodes use a canonical ID order.
- Deduplicated undirected edge pairs use canonical endpoint indices and numeric pair order.
- Existing large-graph degree/confidence filters remain unchanged.

## Reproduction shape

Build two three-node `KnowledgeGraph` values with identical undirected connectivity, add nodes in different orders, add edges in reverse order and direction, and compare `buildCommunityProjection` results. They differ today in both node and edge arrays.

This issue concerns projection determinism only; it should not be used to hide genuine relationship-set differences.

100yenadmin:
Fixes #2477.

## Summary

- sort eligible community-projection nodes by canonical ID;
- deduplicate undirected edges using canonical endpoint indices;
- sort the final numeric edge pairs before Graphology or CSR construction;
- keep all existing confidence, degree, symbol, and large-graph filters.

## Tests

- added a red-to-green insertion-order and direction invariance regression;
- community and cohesion suites: 20 passed;
- formatting and `git diff --check` passed.

This removes input insertion order as a source of Leiden nondeterminism. It does not change or conceal the underlying relationship set.


100yenadmin:
Current-upstream self-review at `ff385a54`: no further change needed. Node IDs are canonicalized before index assignment, undirected endpoints are normalized before deduplication, and the final integer edge list is sorted; the change therefore removes both node- and relationship-insertion-order variance without changing eligibility or large-graph filtering. Focused community projection/CSR proof: 14/14 tests pass.

100yenadmin:
The macOS golden failure was reproducible locally and represented an expected deterministic-output change, not a runner flake.

I captured the complete semantic edge sets from upstream `main` (`a05b5011`) and this PR head. Both have 33 nodes, 69 relationships, and 4 processes. The only differences are six `MEMBER_OF` edges where two unchanged membership groups exchange the opaque names `Cluster_0` and `Cluster_1`; no symbol moves between groups and no call, import, process, or structural edge changes.

Head `b37dcef7` updates the golden digest to the reviewed canonical projection. Proof:
- full semantic edge-set comparison retained in the review evidence
- golden pipeline test passes three consecutive runs
- 14/14 focused community tests pass
- build and `npx tsc --noEmit`
- focused Prettier and ESLint: zero errors (two pre-existing warnings)

GitHub CI has restarted on this exact head.

gitnexus/test/fixtures/pipeline-golden/mini-repo/expected-graph.json
```diff
@@ -24,5 +24,5 @@
     "MEMBER_OF": 12,
     "STEP_IN_PROCESS": 12
   },
-  "edgeDigest": "1e80aba78cb1784276d387e9debd006ae4e82e60ce33c59e338aac4886d98f61"
+  "edgeDigest": "02858462a2acca13f09b7ae2a81d56fe858e67064552ea966cd9ca242371e029"
```
magyargergo, line 27: This rebaseline is incomplete. The coverage 1/3 shard fails in pipeline-pdg.test.ts on the stable golden digest assertions for C#, Java and Go, because the projection reorder changes their community output too. C and C++ happen to pass. Please rebaseline the per language pipeline pdg goldens as well.

magyargergo (changes requested):
The change itself is exactly what #2477 called for and the invariance test is the right test. But CI is red because of this PR, see the inline comment. Once the goldens are green this is an easy approve.

Sequencing thought: #2482 shifts these same digests again through the walker sort. I would land #2482 first, rebase this on top, and rebaseline everything in one pass so we only churn the fixtures once. Same for the release notes, existing users get a one time cluster renumbering on reindex, we should batch that message with #2482 and #2480.

gitnexus/test/fixtures/pipeline-golden/mini-repo/expected-graph.json
magyargergo: Rebaselined in b685ba4c on the merge train after stacking on #2482. The three per language snapshots were regenerated in one pass.

magyargergo:
Merged via #2511, stacked on #2482 so the goldens were rebaselined in one pass, in b685ba4c. That covered the three per language snapshots that made CI red here.

PR #2480, opened 2026-07-13
fix(scope): stabilize graph lookup collisions

### LINKED ISSUE #2479: Scope graph lookup depends on insertion order when keys collide
## Summary

`buildGraphNodeLookup` populates first-wins maps directly from `graph.iterNodes()`. When two linkable nodes produce the same simple or qualified lookup key, the selected target depends on graph insertion order rather than source order.

## Reproduction

Create two `Method` nodes in the same file with the same name and qualified name, at lines 10 and 20. Build the lookup once with the later node inserted first and once with the earlier node inserted first. Current `main` resolves the colliding keys to different node IDs. The same failure occurs when source positions are equal and insertion order is reversed.

## Impact

Equivalent analyses can select different graph targets for ambiguous first-wins keys. That can rewire downstream relationships without changing aggregate relationship counts, making clean rebuilds non-reproducible.

## Expected behavior

Collision selection should be deterministic. A minimal policy is source path, source line, then stable node ID. Unique-key behavior should remain unchanged.

## Proposed validation

- prove both reversed-insertion cases fail on current `main`
- sort only linkable lookup candidates before map construction
- verify the reversed-insertion cases and the affected pipeline/community golden suites

I found no existing issue matching this insertion-order collision.

100yenadmin:
## Summary
- canonicalize linkable graph candidates before building first-wins lookup maps
- select collisions by source path, source line, then stable node ID
- preserve existing behavior for unique lookup keys

## Test plan
- red on current `main`: 2/2 reversed-insertion cases fail
- green with fix: 2/2 determinism cases pass
- affected scope/community/pipeline matrix: 53/53 tests pass after build
- production build passes
- targeted formatting and `git diff --check` pass

The first broader local probe failed before execution because the fresh worktree had not generated its worker scripts. After the required build, the same 53-test matrix passed.

Fixes #2479


gitnexus/src/core/ingestion/scope-resolution/graph-bridge/node-lookup.ts
```diff
@@ -67,17 +67,41 @@ export function simpleKey(filePath: string, name: string): string {
...
+    : Number.MAX_SAFE_INTEGER;
+  if (leftLine !== rightLine) return leftLine - rightLine;
+
+  return compareText(left.id, right.id);
+}
+
 export function buildGraphNodeLookup(graph: KnowledgeGraph): GraphNodeLookup {
   const lookup = new Map<string, string>();
-  for (const node of graph.iterNodes()) {
+  const linkableNodes = Array.from(graph.iterNodes()).filter((node) => {
```
magyargergo: This may cause performance issues, and we can do this much effectively earlier in the pipeline.

magyargergo (changes requested):
Thanks for the pull requests you submitted I'll work my way through slowly of them. I also need to understand the big picture of the issue sets.

100yenadmin:
@magyargergo thanks this is just my personal fork. I'm pushing up our work as our AI community in asia uses my fork so pushing up fixes we've found for much larger codebases etc.

magyargergo:
> @magyargergo thanks this is just my personal fork. I'm pushing up our work as our AI community in asia uses my fork so pushing up fixes we've found for much larger codebases etc. 

Thank you 🙏 we appreciate any contribution that makes GitNexus more useful and of course stable and reliable.

I hope, we can get your fixes merged in as soon as possible, we just need to fine tune them. :)

gitnexus/src/core/ingestion/scope-resolution/graph-bridge/node-lookup.ts
100yenadmin: Reworked in 1f7036db based on this feedback. The lookup no longer materializes or sorts the graph. Canonicalization now happens once at the earlier ParseWorkerResult -> KnowledgeGraph merge boundary, covering both fresh worker output and parse-cache replay. The common already-ordered path is allocation-free O(n); only an out-of-order result copies/sorts once. Worker jobs and input paths are already merged deterministically, so this closes the remaining per-result ordering boundary before every downstream first-wins index. Focused proof: 91 tests across lookup, cache replay, worker fallback/quarantine, clone safety, pipeline golden, and community behavior, plus build/typecheck/format/lint.

100yenadmin:
Big-picture clarification after the redesign: this PR has one narrow contract only. It canonicalizes parse-node insertion at the worker/cache merge boundary so existing first-wins graph indexes receive stable source order. It does not change resolution rules, graph IDs, or collision semantics, and it no longer adds repeated work to buildGraphNodeLookup. Updated to current upstream main without history rewriting; ready for re-review at 1f7036db.

gitnexus/src/core/ingestion/parsing-processor.ts
```diff
@@ -58,6 +58,39 @@ export interface WorkerExtractedData {
...
+  if (fileOrder !== 0) return fileOrder;
+
+  const leftLine = sourceLine(left);
+  const rightLine = sourceLine(right);
+  if (leftLine !== rightLine) return leftLine < rightLine ? -1 : 1;
+
+  return compareText(left.id, right.id);
+}
+
+function nodesInSourceOrder(nodes: readonly ParsedGraphNode[]): readonly ParsedGraphNode[] {
```
magyargergo, line 85: Interplay with #2482: until the walker sort lands, a multi file chunk whose internal file order is not lexicographic hits the sort branch on every merge. Still correct, just not the free path. With #2482 merged first the chunks arrive ordered and this stays linear. So #2482 goes first, then this.

magyargergo (commented):
This rework answers my perf concern. The sortedness scan keeps the common path allocation free and the sort only runs when something is actually out of order. The source order semantics also match what extraction produces today, so real inputs should see no behavior change, this just pins the invariant. One sequencing note inline. Once #2482 is in, ping me and I will take another look and approve.

gitnexus/src/core/ingestion/parsing-processor.ts
magyargergo: Sequenced exactly that way on the merge train, #2482 first, then this. Landed via #2511.

magyargergo:
Merged via #2511, sequenced after #2482 as discussed. The perf rework was exactly what I asked for, thanks.

Issue #95, opened 2026-02-27
can gitnexus wiki  support  language setup ?


airclear:
the wiki feature is cool.  default is english ,  can add a config option to support zh-CN

MPED666:
@airclear How to configure Chinese? tell me the command
➜  dataadmin git:(release) ✗ gitnexus wiki --help
Usage: gitnexus wiki [options] [path]

Generate repository wiki from knowledge graph

Options:
  -f, --force        Force full regeneration even if up to date
  --model <model>    LLM model name (default: minimax/minimax-m2.5)
  --base-url <url>   LLM API base URL (default: OpenAI)
  --api-key <key>    LLM API key (saved to ~/.gitnexus/config.json)
  --concurrency <n>  Parallel LLM calls (default: 3) (default: "3")
  --gist             Publish wiki as a public GitHub Gist after generation
  -h, --help         display help for command

magyargergo:
## Triage verification (2026-05-11): **still valid — keep open**

### Check performed

Ran **`gitnexus wiki --help`** on current `main` / dev build.

- There is **no** `--language` / `zh-CN` / locale flag exposed on the wiki CLI today (only provider/model/base-url/api-key/concurrency/gist/review, etc.).

### Mini-checklist

- Issue: #95  
- Current behaviour: wiki generation path does not advertise output language selection  
- Expected behaviour: configurable wiki output language (e.g. `zh-CN`)  
- Can reproduce “missing feature” on main? **yes**  
- Fixed by existing PR/release? **no** (PR #196 is not landed as a user-visible flag in `--help`)  
- Regression test when fixed? **yes** — extend CLI tests under `gitnexus/test/` for `wiki` command argv + a small golden-string check on generated copy when language is set  

**Action:** needs owner / implementation; linking any future PR should update `gitnexus wiki --help` + docs.

magyargergo:
**Release triage (v1.6.5)** — issue **still open**.

Please verify against **v1.6.5**; comment with version and repro if still relevant._

sanguine59:
hello @airclear 

Check out PR #1613, where I added the --lang parameter to generate wiki content based on the natural language of your choice.
example of usage `npx gitnexus wiki --lang chinese`

related to a closed issue #1585

PR #2482, opened 2026-07-13
fix(scan): canonicalize traversal without order-sensitive bindings

### LINKED ISSUE #2481: Repository traversal order can rewire Rust and PHP graph edges
## Summary

Equivalent analyses can produce different graph endpoints because repository traversal order is not stable across filesystems or repeated scans. Canonicalizing the scan order exposes two consumers whose results currently depend on file arrival order:

- Rust field and identity-method bindings may be consumed before the defining file publishes them.
- PHP PSR-4 `use function` / `use const` fallback may choose the first file in a namespace directory instead of the file that declares the imported symbol.

The result can retain aggregate node and relationship counts while rewiring individual endpoints.

## Reproduction

1. Return the same accepted files from the repository walker in two different glob orders.
2. Analyze a Rust fixture where an importing file sorts before the type-definition file.
3. Analyze a PHP fixture where `Main.php` imports a function declared in `UserFactory.php` while `User.php` shares the namespace directory.

Without the complete fix, one or more of these outcomes depends on traversal order:

- repository entries are not returned in a canonical order;
- the Rust `City#save` call edge is absent;
- PHP `Main.php` imports `User.php` instead of `UserFactory.php`.

## Expected behavior

- The repository walker returns accepted paths in deterministic code-point order without locale dependence.
- Rust foundational member bindings are published for the whole workspace before assignment and method-chain resolution consumes them.
- PHP function and constant imports resolve to the unique declaring file; ambiguous declarations fail closed.
- Normal symbol-first behavior and existing graph semantics remain unchanged.

## Proposed validation

- reversed glob order yields the same canonical repository path sequence;
- the Rust method-chain binding regression stays green after scan canonicalization;
- PHP `Main.php -> UserFactory.php` import/call edges exist and no false `Main.php -> User.php` import is emitted;
- duplicate PHP declarations do not select an arbitrary target;
- build, typecheck, affected resolver suites, scope-capture benchmark, and cross-platform CI pass.

I found no existing issue matching this full traversal-order contract. The earlier suffix-index-only hypothesis was superseded after whole-repository regression testing localized the remaining order dependencies.

100yenadmin:
## Summary

- canonicalize accepted repository paths once at the scan boundary using deterministic code-point order
- publish Rust field and identity-method bindings for the whole workspace before dependent resolution
- resolve PHP function and constant imports by their unique declaring file instead of namespace-directory arrival order
- preserve the shared resolver API with optional import context and fail closed on ambiguous declarations
- rebaseline the PHP scope-capture fingerprint for the intentional import-target correction

## Why this shape

Sorting only a suffix lookup table did not address the source of nondeterminism: every downstream phase still received filesystem-dependent repository order. Sorting the scan result is the correct boundary, but it must be paired with fixes for consumers that had accidentally relied on arrival order. The Rust and PHP changes make those consumers order-independent rather than preserving an unstable phase order.

## Test plan

- red on current `main`: mocked reversed glob order is returned unchanged
- green with fix: accepted paths are returned in canonical code-point order
- Rust method-chain binding retains the expected `City#save` edge when `main.rs` sorts before `models.rs`
- PHP function import resolves `Main.php -> UserFactory.php` and emits no false `Main.php -> User.php` import
- ambiguous PHP declarations fail closed
- affected scanner, filesystem, shared-finalization, node-lookup, Rust, PHP, and cross-file suites pass
- build and TypeScript no-emit checks pass
- scope-capture benchmark retains its scaling budget with the reviewed fingerprint change
- fork cross-platform CI and two disposable large-repository canaries provide integration evidence before promotion

Fixes #2481


100yenadmin:
Reviewed the current branch against upstream `main` and found one scale issue in the PHP function/constant import path: each import scanned every parsed file.

Updated head `846b5456` now builds a directory index once per stable parse-result array and reuses it for subsequent import resolution. Absolute, relative, and namespace-suffix path behavior is preserved, and candidate sets remain deduplicated and fail closed on ambiguous declarations.

Focused proof:
- 17 filesystem-order, PHP declaration-selection, and range-binding timeout tests
- 221 PHP resolver integration tests after building the worker artifact
- `npx tsc --noEmit`
- focused Prettier and ESLint checks

The first integration attempt correctly failed before parsing because this clean worktree had no generated `parse-worker.js`; after `npm run build`, all 221 pipeline tests passed. GitHub Actions are now queued on this exact head. The Vercel authorization status is external to the code path.

gitnexus/src/core/ingestion/languages/rust/range-binding.ts
```diff
@@ -89,6 +89,13 @@ export function populateRustRangeBindings(
         }
       }
     }
+
+    // Publish per-type member bindings for the whole workspace before resolving
+    // assignments. Otherwise an importer processed before its defining file can
+    // miss a field or identity-method type solely because of file order.
+    const scopeMap = new Map(parsed.scopes.map((scope) => [scope.id, scope]));
+    processFieldTypeBindings(tree.rootNode, parsed, scopeMap);
```
magyargergo, line 97: The move to the publish loop is correct, but nothing locks it in. A refactor could drift these two calls back into the consume loop and no test would fail. The Rust case is the headline repro from #2481, yet PHP and the walker got regression tests and Rust did not. Please add one that feeds populateRustRangeBindings the importer file before the defining file and asserts the cross file method edge still shows up.

Small note: this publish now also runs for files without a Module scope, which the old placement skipped. Looks like a strict improvement, just confirming it is intentional.

gitnexus/src/core/ingestion/languages/php/import-target.ts
```diff
@@ -26,6 +27,89 @@ export interface PhpResolveContext {
...
+  return [...directories];
+}
+
+// A scope-resolution pass shares one stable parsedFiles array across imports.
+const phpDirectoryIndexCache = new WeakMap<
+  readonly ParsedFile[],
+  ReadonlyMap<string, readonly ParsedFile[]>
+>();
+
+function directoryAliases(filePath: string): string[] {
```
magyargergo, line 81: This registers every path suffix as an alias, so two roots that share a suffix pool their files. Take app/Models and vendor/x/app/Models. If a function only exists in the vendor copy, the import resolves to the wrong root. If it exists in both, an edge that used to resolve goes null. Your test covers Models vs Services but not two nested roots sharing a suffix. Can we match only the longest or most specific alias, or fail closed when suffix aliases collide? At minimum a comment documenting the ceiling plus a test for the nested root case.

gitnexus/bench/scope-capture/baselines.json
```diff
@@ -39,9 +39,9 @@
...
   },
   "php": {
-    "fingerprint": "bc2c27c5ba26d5aea61142a2a99fb772222f5b969205260eb7a71b4c0bd73cdb",
+    "fingerprint": "31c9e3f3cb7094a2bf9021cf9db859036e002f8b44605cd993b470fc600e97cb",
     "scaling_budget": 1.5,
-    "_rebaselined": "#1956: heritage-bearing scale source (class extends Base + use trait); both forms gated at scale; linear (~1.04).",
+    "_rebaselined": "#1956: heritage-bearing scale source (class extends Base + use trait); both forms gated at scale; linear (~1.04). | #95: PHP imports carry a symbol-kind capture so function/constant imports resolve by declaring file; capture shape changes, scaling remains linear (~1.04).",
```
magyargergo, line 44: This #95 is your fork's issue number. In this repo it points at an unrelated old issue. Please reference #2481 or this PR instead.

magyargergo (changes requested):
Went through all three legs of this and the approach holds up. I checked the contract change against every other language resolver, the optional param breaks nothing, and run.ts is the only live wiring path so full and incremental runs stay consistent. Good choice on the raw string comparison for the walker sort, localeCompare would have reintroduced platform variance.

Two things to address before merge, see the inline comments. One heads up for later: the walker sort changes scan order for every language, so existing users will see a one time graph and cluster churn on their next reindex. We should batch this with #2478 and #2480 in the same release and call it out in the release notes.

Also spotted while reading, not for this PR: the allReturnTypes and allFieldTypes dedup in range-binding.ts toggles between set and delete, so a name that appears three times keeps the third file's value instead of staying ambiguous. Deterministic after your sort but still wrong for three or more duplicates. Worth a follow up issue.

gitnexus/src/core/ingestion/languages/rust/range-binding.ts
magyargergo: Added in 6b013e7d, and I verified the new test fails on the pre fix ordering before keeping it.

gitnexus/src/core/ingestion/languages/php/import-target.ts
magyargergo: Done in 6b013e7d. The fallback now fails closed when a suffix matches directories under different roots, with tests for both the single declaration and dual declaration cases.

gitnexus/bench/scope-capture/baselines.json
magyargergo: Fixed in 6b013e7d, the note now cites #2481/#2482.

magyargergo:
Merged via #2511. The three review asks landed on the train in 6b013e7d: the Rust order regression test (verified to fail on the pre fix code), the fail closed guard for suffix colliding roots, and the corrected baselines reference. Heads up for everyone watching: existing indexes will see a one time graph churn on the next reanalyze.

100yenadmin:
Captured the out-of-scope Rust ambiguity finding in #2514 after re-verifying both toggle sites on current `main`. The issue specifies two/three-duplicate and input-order coverage for return and field inference.

Issue #1233, opened 2026-04-30
Don't modify gitignore of analyzed repo, usig gitignore with * inside .gitnexus instead


faysou:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Problem or opportunity

gitnexus modifies gitignore of analyzed repo which is something we don't want to commit in an open source project, and require skip-worktree

### Proposed solution

Instead add gitignore file inside .gitnexus with * line inside. This does the same thing as modifying the analyzed repo gitignore without having to use skip-worktree, which is annoying to deal with.

### Contribution

- [ ] I am willing to open a PR for this (may need design discussion first).

PR #2488, opened 2026-07-14
feat(cli): mirror skills to .agents/skills/ when .agents/ exists

ArgonarioD:
Some agents prefer repo-local .agents/skills/ over the global install. When .agents/ is present, mirror the standard and generated skills written to .claude/skills/ so those agents serve up-to-date copies. Opt-in via .agents/; absent directory leaves the layout untouched.

Summary

analyze writes repo-specific skills to .claude/skills/. Agents like the Codex app read skills from a repo-local .agents/skills/ directory and do not read .claude/skills/. This PR mirrors the standard and generated skills into .agents/skills/ when the repo contains an .agents/ directory, so those agents stay in sync without a manual copy. Opt-in via .agents/; absent directory leaves the layout untouched.

Motivation / context

setup installs the bundled GitNexus skills globally to ~/.agents/skills/ (Codex). But analyze generates repo-specific skills (community skills from --skills, plus the standard skill set) only under .claude/skills/. The Codex app migrates .claude/skills/ into .agents/skills/ once on first entry, then reads only .agents/skills/ — so when analyze later updates .claude/skills/, Codex never picks up the changes and keeps serving stale migrated copies. Mirroring on every analyze closes that gap.

Areas touched

- [x] gitnexus/ (CLI / core / MCP server)
- [ ] gitnexus-web/ (Vite / React UI)
- [ ] .github/ (workflows, actions)
- [ ] eval/ or other tooling
- [x] Docs / agent config only (README.md, CLI help)

Scope & constraints

In scope

- ai-context.ts — installSkills mirrors the standard skills to .agents/skills/<name>/ when .agents/ exists; new shouldMirrorSkillsToAgents(repoPath) gate. Mirror writes are best-effort (per-skill try/catch + logger.warn).
- skill-gen.ts — generateSkillFiles mirrors generated community skills to .agents/skills/gitnexus-area-*/, clearing the gitnexus-area-* namespace first. Mirror writes are best-effort (per-skill try/catch + console.log warn); a mirror-root mkdir failure disables mirroring for the run without aborting canonical generation.
- git.ts — isWorkingTreeDirty excludes .agents/ + .agents/** so a tracked .agents/ dir doesn't defeat the up-to-date fast path.
- README.md + --skip-skills help (en/zh) — note the .agents/skills/ mirror.

Out of scope

- No change to setup / editor-targets.ts — the global ~/.agents/skills/ install path is untouched.
- No change to AGENTS.md / CLAUDE.md generation content.
- Gate stays on .agents/ existing (not .agents/skills/) — see review reply for rationale.

Implementation notes

- Opt-in gate: shouldMirrorSkillsToAgents does a single fs.stat on <repo>/.agents; absent or not-a-directory returns false.
- Flat layout (post-#2434): standard skills → .agents/skills/gitnexus-*/, community skills → .agents/skills/gitnexus-area-*/.
- Mirror is a weak side-flow: canonical .claude/skills/ is the source of truth; mirror failures log a warning and never abort the canonical install or the analyze run.
- skill-gen.ts clears only the gitnexus-area-* namespace under .agents/skills/ (preserving mirrored standard skills and user-authored skills), matching the canonical cleanup.

Testing & verification

- [x] cd gitnexus && npx tsc --noEmit — clean
- [x] cd gitnexus && npm test — 494 files passed; 9 failures are pre-existing macOS /var vs /private/var symlink issues in analyzer-identity.test.ts / evidence-provenance-helper.test.ts (verified failing on pristine upstream main, unrelated to this PR)
- [x] cd gitnexus && npm run test:integration — 200 files, 4354 passed / 19 skipped, 0 failed
- [x] Manual / local verification — built locally and ran gitnexus analyze (with and without --skills) against a separate local repo with .agents/; confirmed standard skills mirror to .agents/skills/gitnexus-*/ and community skills to .agents/skills/gitnexus-area-*/, byte-identical to .claude/skills/. Repos without .agents/ unchanged. Mirror-failure path (.agents/skills as a file) degrades gracefully with canonical intact.
- 18 new tests: mirror failure paths (root-is-file, per-skill fail, delete-then-rewrite ordering, namespace-scoped cleanup), dirty-check excludes (real-edit regression, prefix collision .agentsrc, subdir .agents/, non-git/git-missing conservative fallback), gate on file-not-dir, idempotency, --skip-skills.

Risk & rollout

- Low risk. Strictly additive and gated on .agents/ existing; repos without .agents/ are byte-for-byte unchanged.
- No breaking changes, no migration, no index refresh required. Existing indexes pick up the mirror on the next analyze.
- Release note: "Skills generated by analyze are now mirrored to .agents/skills/ when the repo contains an .agents/ directory, for agents that read repo-local skills (e.g. Codex)."

Checklist

- [x] Docs/help updated in the same change (DoD.md L62): README.md + --skip-skills help (en/zh)


magyargergo:
Thank you for your contribution! We appreciate the time and effort you put into this but there's one ready PR in the queue already for this https://github.com/abhigyanpatwari/GitNexus/pull/2434

ArgonarioD:
> Thank you for your contribution! We appreciate the time and effort you put into this but there's one ready PR in the queue already for this #2434

@magyargergo Thanks for pointing me to #2434. I've read through it — it flattens .claude/skills/ (standard skills as direct children, community skills under a gitnexus-area-* namespace), which fixes discoverability for Claude Code. That covers the Claude side.

My PR targets a different gap #2434 doesn't touch: .agents/. Concretely, the current Codex app asks users to migrate repository-level skills from .claude/ into .agents/, then reads only .agents/ afterward. So when gitnexus analyze updates skills under .claude/skills/ (including #2434's new community skills), Codex never picks up those updates — it keeps serving the migrated stale copies. Mirroring the writes to .agents/skills/ on every analyze is what keeps Codex in sync.

#2434 has zero references to .agents/, so this isn't covered. Is the .agents/ mirroring intentionally out of scope, or should it land as a follow-up?

gitnexus/src/cli/skill-gen.ts
```diff
@@ -135,6 +160,9 @@ export const generateSkillFiles = async (
   // Step 4: Ensure the shared project-skill root exists. Never clear it: it
   // also contains user-authored and standard GitNexus skills.
   await fs.mkdir(outputDir, { recursive: true });
+  if (mirrorToAgents) {
```
magyargergo, line 167: **MEDIUM — unwrapped mirror writes silently abort canonical community-skill generation and destroy prior output (empirically reproduced).**

This `mkdir` — and the per-skill mirror `mkdir`/`writeFile` at lines 218-222 — are the only unwrapped filesystem ops in the mirror feature. Reproduced through this repo's own test harness with `.agents/` present but `.agents/skills` occupied by a file:

- `generateSkillFiles` throws `EEXIST` here, **after** the canonical `gitnexus-area-*` cleanup (lines 89-98) already ran → previously generated canonical skills are deleted and nothing is written back (`canonicalExists: false` for every community skill);
- a mid-loop failure at line 219-221 aborts canonical writes for all *remaining* skills, leaving a partial canonical set;
- the caller (`analyze.ts` ~1508) swallows the rejection with a silent `catch { /* best-effort */ }`, so the user gets no community skills and no message.

The PR body's stated contract — "Mirror writes are best-effort and per-skill: a failed mirror copy logs a warning and continues … it never fails the whole analyze run" — is implemented in `ai-context.ts` (per-skill try/catch + `logger.warn`) but not here. No test covers any mirror failure path.

**Remediation:** wrap the mirror ops (this mkdir and the loop write) in try/catch + warn, mirroring `ai-context.ts`'s pattern — simplest shape is a shared `mirrorSkill(dir, name, content)` helper used by both sites, which also prevents the two implementations drifting again. A regression test with `.agents/skills` as a file would pin the contract.

gitnexus/src/cli/ai-context.ts
```diff
@@ -363,14 +363,33 @@ async function upsertGitNexusSection(
...
+ * Mirrored to .agents/skills/ when .agents/ exists.
  */
-async function installSkills(repoPath: string): Promise<string[]> {
+async function installSkills(
+  repoPath: string,
+): Promise<{ skills: string[]; agentsMirror: boolean }> {
   const skillsDir = path.join(repoPath, '.claude', 'skills');
   const legacySkillsDir = path.join(skillsDir, 'gitnexus');
   const installedSkills: string[] = [];
+  const agentsMirror = await shouldMirrorSkillsToAgents(repoPath);
```
magyargergo, line 390: **MEDIUM — `.agents` is missing from the up-to-date fast-path dirty-check exclusion list, permanently defeating "Already up to date" for exactly the repos this feature targets.**

`run-analyze.ts:960-977` runs `git status --porcelain` with `:(exclude)` pathspecs for every path GitNexus itself writes during analyze — `.gitnexus`, `.claude/**`, `.cursor/**`, `AGENTS.md`, `CLAUDE.md` — with the contract spelled out in the comment above it: counting self-written paths as dirty "would perpetually defeat the up-to-date fast path" (#1233 regression guard). This PR adds a new analyze-time write surface (`.agents/skills/**`, from this call plus `generateSkillFiles` on `--skills`) without adding it to that list.

**Failing scenario:** a repo with a tracked `.agents/` (e.g. a committed plugin marketplace) and no `.agents/*` ignore rule. Run 1 writes untracked `.agents/skills/gitnexus-*/SKILL.md` mirrors; every subsequent same-commit `analyze` sees `?? .agents/skills/…`, reports dirty, and falls through to the full incremental pipeline (lbug write lock, parse-cache diff, context rewrite — which rewrites the mirror, keeping the loop alive) instead of the instant early return. Permanent until the user commits or ignores the files, and amplified by hook-triggered analyzes contending for the lock. The PR body's "strictly additive… no index refresh required" doesn't hold for this population. The existing fast-path test (#1233) never creates `.agents/`, so nothing catches it.

**Remediation:** add `':(exclude).agents'` + `':(exclude).agents/**'` to the pathspec array at `run-analyze.ts:969-976`, and a fast-path test with `.agents/` present.

gitnexus/src/cli/ai-context.ts
```diff
@@ -363,14 +363,33 @@ async function upsertGitNexusSection(
   return 'appended';
 }
 
+/**
+ * Some agents read skills from a repo-local `.agents/skills/` directory and
+ * prefer it over the global `~/.agents/skills/` install. When the repo contains
+ * an `.agents/` directory, skills written to `.claude/skills/` are mirrored
+ * there too so those agents serve the up-to-date copies.
+ */
+export async function shouldMirrorSkillsToAgents(repoPath: string): Promise<boolean> {
```
magyargergo, line 370: **LOW — the opt-in gate fires on `.agents/` directories that exist for non-skill purposes.**

`.agents/` is also the Codex plugin-marketplace convention — this repository itself carries a checked-in `.agents/plugins/marketplace.json` (see `CONTRIBUTING.md:179`, with the rest of `.agents/` gitignored). A repo carrying `.agents/plugins/` only — no agent reading repo-local skills — gets 6 (up to ~26 with `--skills`) unrequested skill directories injected, and absent an ignore rule, the fast-path defeat from the other comment. Harmless in this repo only because its `.gitignore` happens to swallow the rest of `.agents/`.

**Cheap option:** gate on `.agents/skills/` existing rather than `.agents/` (a repo that wants the mirror creates the directory once), or document the trigger in the README so plugin-only repos know how to avoid it. Flagged as a design cost with a concrete carrying scenario, not a blocker — author's call.

magyargergo (changes requested):
## Review: mirror skills to `.agents/skills/` when `.agents/` exists

Reviewed at head `7ba5483f` against merge-base `91955e65`, in a dedicated worktree with a fresh GitNexus index. Evidence: full diff read; `detect_changes` (exactly the 5 changed symbols, zero affected execution flows); `tsc --noEmit` clean; all 67 tests in the two touched files pass locally; CI green (fork-level Vercel auth noise aside); a combined domain + cross-cutting expert pass; and the headline finding **reproduced empirically through this repo's own test harness**.

The feature itself is sound and well-scoped: the opt-in gate is a single `stat`, canonical writes always precede mirror writes, the `gitnexus-area-*` mirror cleanup is correctly namespace-scoped (standard and user-authored skills survive it, symmetric with the canonical cleanup), the mirrored content is byte-identical and contains no `.claude`-specific references, and the skill-gen→ai-context import is safe. Flag interplay (`--skills` / `--skip-skills` / `--index-only`) produces no state where the mirror is written but canonical is skipped.

Two defects need fixing before merge, both cheap:

### Empirical repro for the headline finding

With `.agents/` present but `.agents/skills` occupied by a file (harness-driven, this repo's own test helpers):

```
threw:            Error: EEXIST: file already exists, mkdir '…/.agents/skills'
canonicalExists:  false   ← zero .claude/skills/gitnexus-area-* written
```

And because the canonical `gitnexus-area-*` cleanup at `skill-gen.ts:89-98` runs **before** the throwing mirror `mkdir`, a rerun in that state deletes previously generated canonical skills and writes nothing back — all swallowed silently by `analyze.ts`'s pre-existing `catch { /* best-effort */ }`. This contradicts the PR body's own contract ("Mirror writes are best-effort and per-skill: a failed mirror copy logs a warning and continues"), which the `ai-context.ts` half honors and the `skill-gen.ts` half does not. Details inline.

### Findings

- **MEDIUM** (inline, `skill-gen.ts:163`) — unwrapped mirror `mkdir`/`writeFile` silently abort canonical community-skill generation and destroy prior output
- **MEDIUM** (inline, `ai-context.ts:392`) — `.agents` missing from the up-to-date fast-path dirty-check exclusion list; repos with a tracked `.agents/` lose the fast path permanently
- **LOW** (inline, `ai-context.ts:372`) — the gate fires on `.agents/` directories that exist for non-skill purposes (e.g. Codex plugin marketplaces — this repo itself carries `.agents/plugins/marketplace.json`)
- **LOW** — docs/help staleness and PR-body drift: `README.md:184` ("6 agent skills installed to `.claude/skills/`"), `README.md:392`/`:448` and the `--skip-skills` help text (`src/cli/index.ts:93-94`) all omit the mirror (`--skip-skills` now also suppresses the standard-skill mirror; `--skills` community skills also land in `.agents/skills/`). The PR body also describes a nested `.agents/skills/gitnexus/` + `generated/` layout and manual verification of those paths — the shipped code writes the flat `.agents/skills/<name>/` layout and clears the `gitnexus-area-*` namespace, so the description and the recorded manual verification predate the final revision; please re-verify against head and update the body. DoD.md line 62 asks for the doc updates in the same change.

The 4 new tests cover the mirror-present/mirror-absent happy paths for both write sites, with no conditionals around assertions. Missing: any failure-path test (`.agents/skills` as a file — the headline trigger; the `ai-context.ts` warn path), and a fast-path regression test with `.agents/` present (the existing #1233 test never creates it). CHANGELOG correctly untouched. Verified clean: architectural fit (repo-local analyze-time injection already lives in these two files; `editor-targets.ts` remains the global setup/uninstall source of truth, untouched), TypeScript conformance (no `any`, `unknown`-typed catches, house test idioms), and YAGNI (the per-site `stat` and the `{skills, agentsMirror}` return-shape change are both fine).

### Verdict

**REQUEST CHANGES** — both MEDIUMs have small, mechanical fixes: a shared per-skill try/catch mirror helper used by both sites (which also restores the PR body's stated contract), and two `:(exclude)` pathspecs in `run-analyze.ts`. Evidence is valid for head `7ba5483f`; re-verify if the head moves.

azizur100389:
Thanks for the update. I rechecked the current head against current `main`.

I agree with the concern that this overlaps with the already-ready skill-layout work in #2434, so I would not merge this as-is unless the maintainers explicitly want this narrower `.agents/skills` mirror as a separate follow-up.

There is also still a correctness gap in the generated-skill mirror path. Standard skill mirroring in `ai-context.ts` is best-effort: if writing `.agents/skills/...` fails, it logs a warning and the canonical `.claude/skills/...` install still succeeds. Generated community skills in `skill-gen.ts` do not have the same error boundary: after writing the canonical `.claude/skills/gitnexus-area-*/SKILL.md`, the mirror write to `.agents/skills/...` can throw and fail the whole generation, leaving a partial state.

If this PR continues, generated skill mirroring should match the standard-skill behavior: canonical `.claude` output should remain the source of truth, `.agents` mirroring should be best-effort with a warning, and there should be a regression test where the `.agents` mirror write fails but `generateSkillFiles` still succeeds with the canonical skills present.

azizur100389 (changes requested):
Requesting changes.

`gitnexus/src/cli/skill-gen.ts` should treat generated-skill mirroring to `.agents/skills/generated/` as best-effort, the same way standard skill mirroring is handled. Right now a mirror mkdir/write failure can escape after canonical cleanup and leave `.claude/skills/generated/` missing or partial. Please wrap the mirror path with a warning-only error boundary and add a regression test where `.agents/skills` fails but canonical `.claude` output still succeeds.

Please also exclude `.agents` / `.agents/**` from the analyze dirty-check pathspecs in `gitnexus/src/core/run-analyze.ts`, otherwise repos that track `.agents/` can lose the up-to-date fast path after mirror writes.

Since this changes user-visible analyze output, the README/help text should mention the `.agents` mirror behavior and how `--skip-skills` affects it.

gitnexus/src/cli/ai-context.ts
ArgonarioD: I understand your concern, but from my own testing, agents like the Codex app do not read skills under .claude/skills/. For a repo like GitNexus where .agents/ has no skills/, Codex prompts the user to manually mirror skills on first entering the repo — but if they're updated later, Codex can't detect the asymmetry between the two sides. So I think it's best to keep creating mirror skills whenever .agents/ exists. If the repo owner doesn't want it polluting git commits, they can add a .gitignore entry (a wildcard ignore solves this simply, at low cost) or use the --skip-skills flag.

magyargergo:
@gitnexus review

magyargergo:
@gitnexus review

magyargergo (approved):

PR #2489, opened 2026-07-14
fix(mcp): avoid top-level api_impact schema combinators

### LINKED ISSUE #2487: api_impact tool uses anyOf at top level of inputSchema → breaks Claude via AWS Bedrock (TOOL_SCHEMA_INVALID)
### Area

### Summary

The api_impact MCP tool declares anyOf at the top level of its inputSchema, which AWS Bedrock rejects with TOOL_SCHEMA_INVALID, breaking all Claude tool-calling whenever gitnexus is used through any Bedrock-routed provider.

### Context

Use gitnexus MCP tools alongside a Claude model served through AWS Bedrock. I expected the gitnexus tools to be available to the agent exactly like any other MCP server (as they already work with OpenAI / Anthropic-native models)

### Expected behavior

The api_impact tool — and the full gitnexus tool set — should be accepted by Bedrock and callable by Claude, the same way they work with non-Bedrock providers.

### Actual behavior

Every agent call that includes the gitnexus tool set fails with 400 TOOL_SCHEMA_INVALID: input_schema does not support oneOf, allOf, or anyOf at the top level. Because MCP clients send tool calls with stream: true, the 400 is returned as a streaming error and gets silently swallowed — the session just hangs with no visible error. GPT/OpenAI models are unaffected.

### Steps to reproduce

1. Configure gitnexus as an MCP server (e.g. npx -y gitnexus@latest mcp).
2. In an MCP client (OpenCode / Claude Code), select a Claude model served through Amazon Bedrock (e.g. anthropic.claude-opus-4-... via the amazon-bedrock provider).
3. Send any prompt that triggers the primary agent (which sends the full tool list, including api_impact).
4. Result: request returns 400 TOOL_SCHEMA_INVALID from Bedrock; the client hangs with no surfaced error.

### Environment

- OS: macOS (darwin)
- Node: v22.22.3
- Browser: n/a (CLI / MCP server)
- Commit / version: gitnexus 1.6.9 (latest stable) and 1.6.10-rc.21 (latest RC) — both affected

### Logs / screenshots

Error returned by Bedrock:
400 TOOL_SCHEMA_INVALID
"input_schema does not support oneOf, allOf, or anyOf at the top level"
Offending schema — dist/mcp/tools.js, tool api_impact:
inputSchema: {
    type: 'object',
    properties: {
        route:  { type: 'string', description: 'Route path (e.g., "/api/grants")' },
        file:   { type: 'string', description: 'Handler file path (alternative to route)' },
        method: { type: 'string', description: 'Optional HTTP verb — GET, POST, PUT, PATCH, DELETE, etc.' },
        repo:   { type: 'string', description: 'Repository name or path.' },
    },
    required: [],
    // Exactly one lookup key is needed...
    anyOf: [{ required: ['route'] }, { required: ['file'] }],
}
Suggested fix: drop the top-level anyOf (the "at least one of route/file" rule is already enforced at runtime — the tool returns an error if neither is provided), making route/file simply optional.

koriyoshi2041:
## Summary
- remove the top-level `anyOf` from the `api_impact` MCP input schema
- keep `route` and `file` optional at the transport layer so stricter provider validators can load the tool list
- add a regression test that rejects top-level `anyOf` / `oneOf` / `allOf` on this schema while relying on the existing runtime guard for missing lookup keys

Fixes #2487.

## Verification
- `npx vitest run test/unit/tools.test.ts`
- `npx vitest run test/unit/calltool-dispatch.test.ts`
- `npx tsc --noEmit`

## Risk
Low. The transport schema is looser, but `callTool('api_impact', {})` still returns the existing structured error when neither `route` nor `file` is provided.


100yenadmin:
Downstream validation from the Electric maintenance fork is now complete: https://github.com/electricsheephq/evaOS-gitnexus/pull/123

The fork generalized the compatibility contract across every advertised MCP tool: no top-level `anyOf`, `oneOf`, or `allOf`; existing property names remain stable; and the `api_impact` route/file plus `impact` target/name/symbol requirements are enforced during runtime normalization before repository resolution. The focused matrix covers canonical, agreeing, conflicting, missing, grouped, read-only, allowlisted, stdio, and HTTP paths. All 317 focused cases and the complete cross-platform PR matrix passed at exact head `3d0daebd9738a7f53b71dae3a458b72550e9f766`.

This is provider-neutral implementation evidence; the tests use deterministic temporary data and contain no credentials or private repository policy.

koriyoshi2041:
CI readback after the rebase: all quality/typecheck/build/CodeQL jobs passed. The only non-Vercel failure is `tests / windows-latest (platform-sensitive) 1/3`, where `test/integration/lbug-multiwriter-deadlock.test.ts` hit the 15-minute Vitest timeout after the real checkpoint-vs-reader race test retried once.

That test looks unrelated to this PR's `api_impact` schema diff; the same run has Windows shards 2/3 and 3/3 green. I tried to rerun failed jobs, but GitHub requires repo admin rights for that action on this repo.

gitnexus/test/unit/tools.test.ts
```diff
@@ -128,14 +128,15 @@ describe('GITNEXUS_TOOLS', () => {
     expect(contextTool.inputSchema.required).toEqual([]);
   });
 
-  it('api_impact tool expresses the route-or-file requirement via anyOf (#2308)', () => {
+  it('api_impact tool avoids top-level schema combinators for Bedrock compatibility (#2487)', () => {
```
magyargergo, line 137: [MEDIUM — the one blocking item] This file conflicts with current `main`: since your branch point (a05b5011), main added adjacent tests in this exact region — `context tool advertises file as a compatibility alias` right above, and `impact tool requires direction … without combinators` right below (from 3f3494fd/a75844b6) — while this PR rewrites the `api_impact` test in between, so the merge is positional-conflicted.

Resolution is mechanical: rebase onto main, keep both of main's neighboring tests unchanged, and take this PR's rewritten `api_impact` test in place of the old anyOf one. Your interface-field removal in `tools.ts` auto-merges cleanly (verified with `git merge-tree`), and post-merge nothing else reads or writes `inputSchema.anyOf`.

For the record, the red `tests / windows-latest 1/3` shard is `lbug-multiwriter-deadlock.test.ts` — unrelated to this change; the rebase push will re-run CI.

magyargergo (changes requested):
## Review: PR #2489 (`rios/bedrock-api-impact-schema` @ 470384b6 vs merge-base a05b5011)

### Findings

- [MEDIUM] `gitnexus/test/unit/tools.test.ts` — the branch is **CONFLICTING** with `main` and cannot be merged as-is (inline comment with resolution guidance). This is the only blocking item.

No defects found in the change itself.

- The fix is real and verified end-to-end: `server.ts:157` forwards `inputSchema` verbatim in ListTools, so the top-level `anyOf` did reach Bedrock's validator; removing it addresses `TOOL_SCHEMA_INVALID` (#2487).
- The runtime guard the PR relies on exists and is structured: `local-backend.ts` `apiImpact()` returns `{ error: 'Either "route" or "file" parameter is required.' }` when both are missing — no behavioral hole opens up from loosening the transport schema.
- The tool description already documents the constraint (`tools.ts:744` — `Requires at least "route" or "file" parameter`), and the new test pins it.
- Removing the `anyOf` field from the `ToolDefinition` interface is safe: after merging with current main there are no remaining writers or readers of `inputSchema.anyOf` anywhere in `gitnexus/` or `gitnexus-web/` (only negative `toHaveProperty` assertions in tests, which don't depend on the TS type).
- This matches the repo's established direction — main's 3f3494fd ("fix(mcp): validate aliases without schema combinators") applied the same no-top-level-combinators pattern to the `impact` tool.
- GitNexus `detect_changes` (compare vs merge-base): LOW risk, 0 affected execution flows. Graph `impact` on `GITNEXUS_TOOLS` returned 0 upstream hits, so consumers were verified by source instead: `server.ts` (ListTools) and the schema self-check loop at `tools.ts:896`.
- No competing fix landed on main; #2487 is still open and this PR is the only candidate.

- The regression test locks out `anyOf`/`oneOf`/`allOf` at the top level of the `api_impact` schema and asserts the description carries the requirement; `calltool-dispatch.test.ts` already covers the runtime rejection path.
- The failing `tests / windows-latest 1/3` shard is `test/integration/lbug-multiwriter-deadlock.test.ts` — an LadybugDB multi-writer integration test unrelated to this 2-file MCP schema change (known Windows flake family). CI Gate is red only transitively; the rebase this review requires will trigger fresh CI anyway.
- Out of scope but worth tracking from the #2487 discussion: the client-side swallowing of Bedrock's 400 into a generic streaming error is a separate diagnosability issue this PR doesn't (and needn't) address.

### Verdict

REQUEST CHANGES — solely for mergeability. Rebase onto main and resolve the `tools.test.ts` conflict; the change itself is correct and I'll approve once the branch is mergeable with green (or flake-explained) CI.

---
*Reviewed with the `gitnexus-review` skill (PR #2431) — GitNexus `detect_changes`/`impact` + source verification in a detached worktree at head.*

koriyoshi2041:
Rebased onto current `main` and resolved the `tools.test.ts` conflict while preserving the newer context-alias and impact-schema coverage. Focused MCP tests pass (258/258); new head is `cb33f6ba` and fresh CI is running.

magyargergo (approved):

PR #2490, opened 2026-07-14
feat(taint): expand TS/JS sink model

### LINKED ISSUE #2486: feat(taint): expand TS_JS_TAINT_MODEL for modern DB libraries (Drizzle/better-sqlite3/Knex), argv-form spawning, and SSTI sinks
## Summary

The built-in TS/JS taint source/sink model (`TS_JS_TAINT_MODEL` in `gitnexus/dist/core/ingestion/taint/typescript-model.js`) covers vanilla Express + Node `child_process.exec[Sync]` + `eval`/`new Function` + `fs.readFile[Sync]` + `.query()`/`.execute()` SQL. It does not cover several extremely common patterns in modern TS/JS codebases, causing `gitnexus analyze --pdg` to report `0 TAINTED, 0 SANITIZES (N function(s) analyzed, M skipped: no source/sink match)` on repos that demonstrably contain source→sink flows.

This is a sibling of #2204 — that issue tracks adding models for *new languages*; this one tracks expanding the *existing* TS/JS model to cover modern ecosystem libraries. The engine (per-function summaries, interprocedural fixpoint, `TAINTED`/`TAINT_PATH` emit, `explain`) is already language-agnostic and ready to consume these.

## Current model recap

For reference, the existing sink set:

```js
sinks: [
  { name: 'exec', kind: 'command-injection', args: [0], module: 'child_process' },
  { name: 'execSync', kind: 'command-injection', args: [0], module: 'child_process' },
  { name: 'spawn', kind: 'command-injection', args: [0], module: 'child_process' },
  { name: 'eval', kind: 'code-injection', args: [0], global: true },
  { name: 'Function', kind: 'code-injection', global: true, newOnly: true },
  { name: 'readFile', kind: 'path-traversal', args: [0], module: 'fs' },
  { name: 'readFileSync', kind: 'path-traversal', args: [0], module: 'fs' },
  { name: 'writeFile', kind: 'path-traversal', args: [0], module: 'fs' },
  { name: 'writeFileSync', kind: 'path-traversal', args: [0], module: 'fs' },
  { name: 'query', kind: 'sql-injection', args: [0], anyReceiver: true },
  { name: 'execute', kind: 'sql-injection', args: [0], anyReceiver: true },
  { name: 'send', kind: 'xss', args: [0], receivers: ['res'] },
  { name: 'write', kind: 'xss', args: [0], receivers: ['res'] },
],
```

Sources are limited to Express-convention request member reads on `req`/`request` receivers (`body`, `query`, `params`, `headers`, `cookies`). Sanitizers: `encodeURIComponent`, `escape-html` default, `he.encode`, `path.basename`, `validator.escape`.

### 1. SQL — Drizzle, better-sqlite3, Knex, Sequelize, mysql2, pg

`db.query()` and `db.execute()` are matched, but the dominant idioms in modern TS/JS codebases use other method names:

| Library | Common sink idiom | Currently matched? |
|---|---|---|
| Drizzle ORM | `db.run(sql\`...\`)`, `db.all(...)`, `db.get(...)`, `db.values(...)` | ❌ |
| better-sqlite3 | `db.prepare(sql).run()`, `db.prepare(sql).get()`, `db.prepare(sql).all()`, `db.exec(sql)` | ❌ (only `exec` would match — and only if not gated by `module`) |
| Knex | `knex.raw(sql)`, `knex(...).toSQL()`, query builder | ❌ |
| Sequelize | `sequelize.literal(sql)`, `sequelize.query(sql, { raw: true })` | partial (`query` matches) |
| mysql2 / pg | `conn.query(sql)` (works), `conn.execute(sql)` (works), prepared statements | partial |

**Realistic miss example** — flows straight from `req.body` into a parameterized SQL template, never flagged:

```ts
app.post('/search', (req, res) => {
  const q = req.body.query;            // source: matches
  db.run(sql`SELECT * FROM items WHERE name LIKE '%${q}%'`);  // sink: NOT matched
});
```

(Whether Drizzle's `sql\`\`` tagged template parameterizes `${q}` automatically is a separate question — the model can't even see the sink to ask it.)

### 2. Command execution — argv-form spawning

The model matches `exec`, `execSync`, `spawn` (all from `child_process`). It does **not** match the argv-form variants, which are generally safer (no shell) but still vulnerable when the **binary path** (arg 0) is tainted:

| Sink | Currently matched? | Realistic risk |
|---|---|---|
| `execFile(cmd, args, opts)` | ❌ | tainted `cmd` (binary path) |
| `execFileSync(cmd, args, opts)` | ❌ | tainted `cmd` |
| `spawnSync(cmd, args, opts)` | ❌ | tainted `cmd` |

A common pattern: `execFileSync(req.body.tool, ['--flag'])` — currently invisible.

### 3. Sources beyond Express `req.*`

Modern TS/JS codebases receive tainted data through many channels the current source set doesn't recognize:

- HTTP client response bodies (`(await axios.post(url, ...)).data`, `(await fetch(url).then(r => r.json()))`)
- gRPC / WebSocket / message-queue message bodies
- Plugin / MCP / RPC handler arguments
- Process arguments / environment (already partially covered by other paths but worth noting)

These are harder to match name-based (no convention as strong as Express's `req`), so I'd suggest treating them as a separate work item — but worth tracking here so the source-side gap is visible.

### 4. Response XSS — `res.json`, `res.render`, `res.sendStatus`

`res.send` and `res.write` are matched; the equally common variants are not:

- `res.json(body)` — JSON responses can carry XSS payloads when later embedded in HTML by consumers; also a response-splitting vector when status/message is tainted
- `res.render(template, data)` — server-side template injection when `template` or `data` is tainted
- `res.sendStatus(code)` — minor; status message reflection

`res.json` is the weakest case (JSON itself isn't HTML-executable), but `res.render` is a genuine server-side template injection sink worth covering.

## Proposed additions

Using the existing model's JSON shape, against `TS_JS_TAINT_MODEL.sinks`:

```js
// ── SQL: Drizzle / better-sqlite3 / Knex / Sequelize ─────────────
{ name: 'run',   kind: 'sql-injection', args: [0], anyReceiver: true },  // Drizzle db.run, better-sqlite3 stmt.run, sequelize.query variants
{ name: 'all',   kind: 'sql-injection', args: [0], anyReceiver: true },  // better-sqlite3 stmt.all, Drizzle db.all
{ name: 'get',   kind: 'sql-injection', args: [0], anyReceiver: true },  // better-sqlite3 stmt.get, Drizzle db.get — see FP note below
{ name: 'raw',   kind: 'sql-injection', args: [0], anyReceiver: true },  // knex.raw, sequelize.literal, drizzle sql.raw
{ name: 'exec',  kind: 'sql-injection', args: [0], anyReceiver: true },  // better-sqlite3 db.exec(string) — collides with existing command-injection sink name; needs disambiguation (see notes)

// ── Command execution: argv-form child_process ───────────────────
{ name: 'execFile',    kind: 'command-injection', args: [0], module: 'child_process' },
{ name: 'execFileSync', kind: 'command-injection', args: [0], module: 'child_process' },
{ name: 'spawnSync',   kind: 'command-injection', args: [0], module: 'child_process' },

// ── XSS / template injection: Express response variants ─────────
{ name: 'render', kind: 'xss', args: [0], receivers: ['res'] },  // res.render(template) — SSTI
{ name: 'render', kind: 'xss', args: [1], receivers: ['res'] },  // res.render(template, data) — SSTI via data
```

### Sources (stretch goal — lower confidence)

```js
// HTTP client response bodies — convention as `response`/`res` (collision with Express `res` receiver needs care)
{ kind: 'remote-input', objects: ['response'], properties: ['data', 'body'] },        // axios
{ kind: 'remote-input', objects: ['axiosResponse', 'apiResponse'], properties: ['data'] },
```

These are speculative — name-based matching on HTTP client responses is brittle because there's no universal convention like Express's `req`. May be better left for a typed-source-resolution future milestone.

## Trade-offs and false-positive notes

1. **`anyReceiver: true` on common method names (`run`, `all`, `get`) will produce false positives** on non-DB receivers (`promise.run()`, `map.get()`, `iterator.next()`, etc). This is consistent with the existing `query`/`execute` posture, but the FP rate for `get` especially will be high. Two options:
   - Accept it and let users mark SANITIZES edges for known-safe call sites.
   - Constrain via `receivers` to DB-conventional names (`db`, `database`, `conn`, `client`, `pool`, `stmt`, `prepared`) — brittle if users rename handles but lower FP. Probably the right trade-off for `get`.
2. **`exec` name collision** — there's already a `command-injection` sink `{ name: 'exec', module: 'child_process' }`. Adding `{ name: 'exec', anyReceiver: true }` for SQL would conflict. Two options:
   - Use a separate `dbExec` matcher alias (requires matcher changes).
   - Skip the `exec` addition and document that better-sqlite3's `db.exec(string)` is uncovered (lowest-friction).
3. **`sql\`\`` tagged template as sanitizer** — Drizzle's `sql\`...\`` and Knex's `knex\`...\`` tagged templates parameterize `${var}` automatically. The model has no concept of "tagged-template-as-sanitizer"; it would need a new matcher shape. Worth tracking as a follow-up so the new SQL sinks don't flag every parameterized Drizzle call.
4. **Argv-form `execFile*` is genuinely lower-risk than shell-form `exec`** — the binary path is the only injection vector, and arg array values are passed-through (not shell-interpreted). Still worth matching because tainted binary path is a real bug, but reviewers should know the FP rate will be high.

## Suggested test fixtures

Concrete code patterns a unit fixture could assert produce `TAINTED` findings after the additions:

```ts
// Fixture 1 — Drizzle SQL injection via db.run
import { sql } from 'drizzle-orm';
import { db } from './db.js';
app.post('/a', (req, res) => {
  db.run(sql`SELECT * FROM t WHERE name = '${req.body.name}'`);  // expect TAINTED sql-injection
});

// Fixture 2 — better-sqlite3 prepared statement
import Database from 'better-sqlite3';
const stmt = db.prepare('SELECT * FROM t WHERE id = ?');
app.post('/b', (req, res) => {
  stmt.get(req.body.id);  // expect TAINTED sql-injection (if `get` addition lands)
});

// Fixture 3 — tainted binary path via execFileSync
import { execFileSync } from 'node:child_process';
app.post('/c', (req, res) => {
  execFileSync(req.body.tool, ['--flag']);  // expect TAINTED command-injection
});

// Fixture 4 — SSTI via res.render
app.post('/d', (req, res) => {
  res.render(req.body.template, { user: req.body.user });  // expect TAINTED xss
});
```

## Scope I'm NOT proposing

- New languages — tracked in #2204.
- Tagged-template sanitizers (Drizzle `sql\`\``, Knex `knex\`\``) — needs matcher work, separate PR.
- Type-aware sink resolution (resolve the DB handle's origin to disambiguate `db.get` from `map.get`) — would massively reduce false positives but is a much larger change.
- HTTP-client / plugin / MCP sources — noted above as a stretch goal; better as its own issue once a convention emerges.

## References

- Parent: #2204 (register source/sink models for more languages)
- Engine: #2164 (intra-procedural taint analysis), #2083
- Cross-language validation that surfaced the per-language model gap: #2197
- The `TS_JS_TAINT_MODEL` constant: `gitnexus/dist/core/ingestion/taint/typescript-model.js`

azizur100389:
## Summary

Expands the built-in TS/JS taint model so `--pdg` can detect more common source-to-sink flows in modern Node/TypeScript codebases.

Closes #2486.

## Motivation / context

The existing `TS_JS_TAINT_MODEL` covered Express request sources plus classic `child_process.exec`, `eval`, filesystem, `.query()`, `.execute()`, and `res.send` / `res.write` sinks. Issue #2486 documents several common gaps: argv-form child process APIs, modern DB helper methods, and Express template rendering.

## Areas touched

## Scope & constraints

**In scope**

- Add command-injection sinks for `child_process.spawn`, `spawnSync`, `execFile`, and `execFileSync`, including argv-array position coverage.
- Add receiver-conventional SQL sinks for `run`, `all`, `get`, `values`, and `raw`.
- Add Express `res.render` as an XSS/template-render sink for template and data arguments.
- Add matcher and propagation tests for the new coverage.

- New language taint models; that remains tracked separately by #2204.
- Treating `db.exec(...)` as SQL, because it collides with existing `child_process.exec` semantics.
- Tagged-template sanitizer modeling for Drizzle/Knex SQL helpers.
- Type-aware DB handle resolution or broader non-Express source modeling.

## Implementation notes

Short DB method names such as `run`, `all`, and `get` are intentionally receiver-conventional rather than `anyReceiver: true`. This keeps obvious false positives like `map.get(...)` and `task.run(...)` out of the model while still covering common DB handle names such as `db`, `database`, `conn`, `client`, `pool`, `stmt`, `statement`, and `prepared`.

## Testing & verification

- [x] `cd gitnexus && npx vitest run test/unit/taint/model-match.test.ts test/unit/taint/propagate.test.ts --maxWorkers=1 --testTimeout=120000`
- [x] `cd gitnexus && npx vitest run test/unit/taint --maxWorkers=1 --testTimeout=120000`
- [x] `cd gitnexus && npx vitest run test/integration/cfg/taint-snapshot.test.ts test/integration/cfg/interproc-taint.test.ts test/integration/pipeline-graph-golden.test.ts --maxWorkers=1 --testTimeout=180000`
- [x] `cd gitnexus && npx eslint src/core/ingestion/taint/typescript-model.ts test/unit/taint/model-match.test.ts test/unit/taint/propagate.test.ts`
- [x] `cd gitnexus && npx prettier --check src/core/ingestion/taint/typescript-model.ts test/unit/taint/model-match.test.ts test/unit/taint/propagate.test.ts`
- [x] `cd gitnexus && npm run build`
- [x] `git diff --check`
- [x] `node gitnexus/dist/cli/index.js detect-changes --repo C:\Users\erazi\Documents\PAPER_CODE\GitNexus` - low risk, 0 affected processes

## Tri review

Current visible state is incomplete. I could verify local diff scope, local branch hygiene, issue state, local validation, and GitNexus `detect-changes`, but not remote PR CI, review comments, and GitHub mergeability until this PR exists. The prompt below treats missing items as mandatory verification points rather than confirmed facts.

No production-readiness issues found against the current DoD bar.

- Review bar for this PR: model-only TS/JS taint coverage expansion, scoped diff, matcher + propagation tests, no schema or matcher contract changes.
- Problem being solved: #2486 coverage gaps for modern DB methods, argv-form process execution APIs, and Express rendering sinks.
- Current PR state: local branch before creation; remote CI pending after PR open.
- Merge status and mergeability: visibility incomplete until GitHub computes the PR; locally the branch was created from current `upstream/main` and had no conflicts.
- Branch hygiene assessment: clean feature/fix PR; tracked diff contains only the taint model and taint tests.
- Findings: initial Tri findings for missing `db.all` coverage and argv-array false negatives were fixed and rechecked.
- Final verdict: production-ready with minor follow-ups, pending remote CI.

## Risk & rollout

This changes the built-in taint model content, so existing `--pdg` indexes should refresh taint findings through the existing `taintModelVersion` mechanism. Default non-PDG analysis behavior is unchanged.

Residual precision trade-offs are expected for name-conventional taint models: uncommon DB receiver names may remain false negatives, while conventional receiver names may still produce conservative findings.

## Checklist

- [x] No `AGENTS.md` / overlay changes


gitnexus/src/core/ingestion/taint/typescript-model.ts
```diff
@@ -56,9 +60,37 @@ export const TS_JS_TAINT_MODEL: SourceSinkSanitizerSpec = {
     // (mysql2/pg/knex handles go by many names; receiver-conventional).
     { name: 'query', kind: 'sql-injection', args: [0], anyReceiver: true },
     { name: 'execute', kind: 'sql-injection', args: [0], anyReceiver: true },
+    // Modern DB libraries expose shorter method names with high collision
```
magyargergo, line 63: Could we write integration tests for them to ensure they get disambiguated? Basically, we need to make sure we only taint paths if they are the right symbols.

magyargergo (changes requested):

gitnexus/src/core/ingestion/taint/typescript-model.ts
azizur100389: Added the requested end-to-end coverage. The new real-pipeline integration test indexes a temporary TypeScript repo with both intended sinks and same-name/collision-prone non-sinks, then asserts persisted TAINTED findings only appear for the imported child_process calls, conventional DB receiver, and Express res.render receiver.

magyargergo (approved):
## Review: PR #2490 — feat(taint): expand TS/JS sink model

Reviewed at head `0bf9fe4` against base/merge-base `c4fb511` (merge-base equals the PR base, so the GitHub diff and the compare surface agree), in a detached worktree at the exact head SHA with a synced GitNexus index.

### Findings

None blocking.

The earlier requested change (end-to-end disambiguation coverage) is addressed by `test/integration/taint-explain.test.ts` — the new real-pipeline test proves the negative space, not just the positives: a module-level user `execFile` shadowing the import, `map.get`/`task.run` collision receivers, and `out.render` all produce **no** findings, while the aliased `childExecFile`, `spawnSync`, `db.run`, and `res.render` sites do.

- Diff: 4 files, +200/−2 — `TS_JS_TAINT_MODEL` sinks only, plus matcher/propagation/integration tests. No matcher, schema, or engine changes; the `receivers`/`args` mechanisms used are pre-existing (`source-sink-config.ts`, `match.ts:334-348`).
- `detect_changes(compare, base=c4fb511)`: 4 changed symbols in `typescript-model.ts`, **0 affected execution flows, LOW risk**. `impact(TS_JS_TAINT_MODEL, upstream)` returns 0; verified by source that the only consumers are `BUILTIN_TAINT_MODELS` (digest) and `registerBuiltinTaintModels` — blast radius is confined to taint matching.
- Persisted-finding invalidation is handled by design: `taintModelVersion` is a content digest of the models, so this change ships a new digest and trips full `--pdg` writeback (KTD7/R7) — no stale-cache hazard.

- Verified locally at head: `model-match` + `propagate` unit suites **83/83 pass**; full `taint-explain` integration suite **20/20 pass** (after building `dist/` — the worker-pool failure without a build is environmental, not the PR).
- Argv-position semantics check out in source: `positionMatches` only reports positions with actual value occurrences, so `spawnSync(value, [])` fires on arg 0 only, and `childExecFile('git', [value])` fires via the array element at arg 1 — both pinned by tests.
- Accepted precision trades (documented in code comments and the PR body, not defects): `client.get(x)` will flag Redis-style cache reads as sql-injection, and two-arg `spawn(cmd, opts)` treats a tainted options object as command-injection surface. Both are noise-side, consistent with the model's Semgrep-convention design; `this.db.run(x)` (3-segment path) remains a known false negative.
- CI: the red **CI Gate** comes solely from `tests / ubuntu / coverage 1/3`, which died in its *build step* (`tsc`: TS2307 `@huggingface/transformers` not found) before running any tests — an optional-dependency install flake, unrelated to this diff (coverage 2/3, 3/3 and all six platform-sensitive shards are green). I'm rerunning the failed job; this must be green before merge. The Vercel failure is the pre-existing fork-deploy authorization gate.

### Verdict

APPROVE — pending only the CI rerun of the flaked shard.

Issue #1225, opened 2026-04-30
Improve analysis for Java projects built with Spring and Spring Boot


PieceOfFall:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Problem or opportunity

# Enhancement: Improve analysis for Java projects built with Spring and Spring Boot
## Summary

GitNexus currently provides strong static code intelligence through parsing, symbol extraction, call relationships, route detection, and process graphs. However, many Java projects built with Spring and Spring Boot express core behavior through framework metadata rather than direct source-level calls.

This affects not only Spring Boot Web applications, but also service applications, batch jobs, messaging systems, scheduled tasks, data-processing services, integrations, and enterprise backend systems built on the Spring ecosystem.

Spring and Spring Boot rely heavily on annotations, IoC container wiring, auto-configuration, runtime proxies, AOP, reflection, configuration files, classpath scanning, and third-party starters. Without Spring-aware analysis, GitNexus may miss important relationships that determine how the application actually behaves.

## Why this matters

Spring is one of the most widely used Java application frameworks, especially in enterprise and backend systems. A large number of production Java codebases depend on Spring or Spring Boot for dependency injection, configuration, transactions, security, data access, messaging, scheduling, and integration behavior.

For these projects, the most important relationships are often not written as explicit method calls. They are inferred or created by the framework at build time, startup time, or runtime.

This has direct impact on GitNexus features such as:

- impact analysis
- process tracing
- route and handler discovery
- dependency flow understanding
- service boundary analysis
- configuration impact analysis
- refactoring safety
- AI-agent assisted code changes

Without Spring-specific semantic modeling, an agent may underestimate the blast radius of changes in Java projects that are otherwise very common in real-world enterprise environments.

## Problem

Spring and Spring Boot applications rely heavily on framework-managed behavior. A static Java call graph alone cannot accurately capture many important relationships, including:

- Annotation-declared behavior
  - `@Component`, `@Service`, `@Repository`, `@Controller`, `@RestController`
  - `@Configuration`, `@Bean`
  - `@RequestMapping`, `@GetMapping`, `@PostMapping`
  - `@Scheduled`, `@Async`
  - `@EventListener`
  - `@Transactional`
  - `@Cacheable`, `@CacheEvict`
  - `@PreAuthorize`, `@Secured`
- Reflection
  - `Class.forName`
  - `Method.invoke`
  - framework-driven reflective invocation
- Runtime proxies
[...]

PR #2494, opened 2026-07-15
feat(spring): build bean candidate inventory

### LINKED ISSUE #2411: feat(spring): build bean candidate inventory
Parent tracking issue: #1225

Status: remaining work

Goal
Detect Spring stereotype annotations and build a bean candidate inventory that later Spring container features can use.

Scope
Recognize common Spring bean annotations such as Component, Service, Repository, Controller, RestController, and Configuration. Record the bean role and enough metadata for context, impact, and later dependency resolution.

Acceptance criteria
1. Spring bean candidates are represented in analysis results with their role.
2. Existing route discovery keeps working as it does today.
3. A small Spring Boot fixture covers the common stereotype annotations.
4. Ambiguous or unsupported annotations fail closed instead of creating misleading relationships.

PieceOfFall:
## Summary

Adds a Java-only Spring bean candidate inventory for the six common stereotype annotations. The inventory is persisted as canonical framework annotation evidence on existing `Class` nodes and exposed as derived Bean metadata from `context` and `impact`.

## Motivation / context

Closes #2411.

Later Spring dependency-resolution work needs a reliable inventory of candidate classes without introducing framework-specific graph nodes or relationships. This change keeps Spring recognition in Java scope resolution while giving storage and MCP consumers a small, framework-neutral evidence field.

## Areas touched

## Scope & constraints

**In scope**

- Recognize `Component`, `Service`, `Repository`, `Controller`, `RestController`, and `Configuration` on Java class declarations.
- Resolve fully-qualified annotations, unique explicit imports, and a single unambiguous Spring wildcard import after Java scope/import resolution.
- Fail closed for lexical or same-package shadowing, multiple wildcard imports, composed annotations, conflicting stereotypes, and unresolved ambiguity.
- Persist canonical evidence in `Class.frameworkAnnotations` and derive a stable `{ framework, role, annotation }` Bean shape for MCP output.
- Preserve existing Spring route discovery and `HANDLES_ROUTE` behavior.

- Kotlin support, `@Bean` methods, Bean names, qualifiers, `@Primary`, conditions, profiles, or composed annotations.
- Full external classpath resolution. When multiple on-demand imports could expose the same simple annotation name, inventory generation remains conservative.
- New Bean nodes, relationships, pipeline phases, or impact traversal behavior.
- Runtime guarantees that a candidate is actually instantiated by Spring.

## Implementation notes

- The existing Java scope query captures class annotation syntax during its normal AST traversal; no second parse or tree traversal is added.
- Annotation facts are serialized through the Java worker capture side channel.
- Spring stereotype resolution runs from the Java scope resolver's post-resolution hook, after imports, lexical declarations, and same-package symbols are available.
- Class extraction remains framework-neutral; Bean evidence is attached directly to matching `Class` graph nodes only after resolution.
- LadybugDB stores only canonical framework annotation evidence. The Spring FQN-to-role mapping remains in a data-only Java module shared by resolution and MCP enrichment.
- Parse-cache schema version 13 and incremental schema version 7 force older indexes to rebuild instead of producing a partial inventory.
- Older or partially upgraded indexes degrade gracefully by omitting Bean enrichment.

## Testing & verification

- [x] `cd gitnexus && npm test`
- [x] `cd gitnexus && npm run test:integration`
- [ ] `cd gitnexus-web && npm test` *(not applicable)*
- [ ] `cd gitnexus-web && npx tsc -b --noEmit` *(not applicable)*
- [ ] Manual / Playwright E2E *(not applicable)*
- [x] `cd gitnexus && npm run test:unit`
- [x] `cd gitnexus && node --import tsx bench/emit-persistence/measure.mjs --check`

Targeted tests cover all six roles, explicit imports, FQNs, annotation arguments, nested classes, a single valid wildcard import, multiple-wildcard fail-closed behavior, lexical and cross-file same-package shadowing, route regression, persistence write paths, schema-version gates, and MCP `context` / `impact` output.

## Risk & rollout

The MCP change is additive and does not alter impact traversal or risk calculation. Existing indexes require a full re-analysis because the `Class` table gains a nullable list column; schema-version gates already enforce that rebuild.

## Checklist

- [x] If `AGENTS.md` / overlays changed: not applicable; none are included in this PR


gitnexus/src/core/ingestion/languages/java/spring-bean-metadata.ts
```diff
@@ -0,0 +1,141 @@
+import Parser from 'tree-sitter';
```
magyargergo: We need to this step later in the process when everything is ready for is and we don't need to double build and traverse the actual file, so the cross file imports will resolve nicely.

magyargergo (changes requested):

gitnexus/src/core/ingestion/languages/java/spring-bean-metadata.ts
PieceOfFall: Thanks, that makes sense. I’ll move this step later in the pipeline so it can use the fully built project context and handle cross-file imports without rebuilding and traversing the file twice.

PieceOfFall: https://github.com/abhigyanpatwari/GitNexus/pull/2494/commits/5882aa7e212fcfeaac53b84e9797cdfab8835f59  Moved Spring bean detection to post-scope-resolution. The existing Java scope query now captures class annotations via a worker side channel, and metadata is applied only after imports and same-package symbols are resolved. This removes the duplicate AST/import traversal and correctly handles cross-file shadowing.

PieceOfFall:
@magyargergo I’ve pushed the fix for the lint failure reported by CI.

The requested implementation changes have already been addressed, and the PR is currently only blocked by the required checks. Could you please take another look or re-run/approve the workflows when convenient?

The CI fix is in commit `1df1ed3`.

gitnexus/src/core/ingestion/languages/java/spring-bean-metadata.ts
```diff
@@ -0,0 +1,133 @@
...
+  if (wildcardTarget === undefined) return undefined;
+
+  // Same-package types and resolved wildcard imports are available only after
+  // finalize/populateNamespaceSiblings. Any such binding wins over guessing
+  // that an unresolved external Spring wildcard supplied the annotation.
+  return hasVisibleTypeBinding(enclosingScope, rawName, indexes) ? undefined : wildcardTarget;
+}
+
+/** Attach Spring stereotype evidence after Java cross-file resolution is complete. */
+export function attachSpringBeanCandidateMetadata(
```
magyargergo: **HIGH — incremental writeback persists a stale bean attribution; incremental and full rebuild diverge.**

`frameworkAnnotations` is (as far as I can find) the first persisted node *property* whose value depends on **other files'** content: the wildcard leg consults package siblings injected by `populateJavaPackageSiblings`. But the incremental write set is edge-based — changed files ∪ importer-BFS over `IMPORTS` edges (`subgraph-extract.ts` walks edges only; no property diffing) — and Java same-package implicit visibility produces no `IMPORTS` edge.

**Reproduced with this PR's built CLI:**
1. Full analyze of `package com.other; import org.springframework.stereotype.*; @Service public class WildcardService {}` → row = `["org.springframework.stereotype.Service"]` (correct — the wildcard fixture asserts exactly this).
2. Add `com/other/Service.java`: `package com.other; public @interface Service {}` (per JLS 6.4.1 the package member now shadows the on-demand import). Incremental analyze logs `changed=0, added=1 … 1 unchanged file rows preserved` → row **still** `["org.springframework.stereotype.Service"]`.
3. `analyze --force` on the identical tree → `[]` (correct).

This pass recomputes the right answer in memory each run — but the recomputed value never reaches the DB for unchanged files, so `context`/`impact` report a Spring bean indefinitely until something else touches that file. The symmetric direction (deleting the shadow type) leaves a stale under-report. The `INCREMENTAL_SCHEMA_VERSION` 6→7 bump only covers the upgrade top-up, not ongoing runs.

**Remediation options:** (a) when the changed/added/deleted set (re)declares a top-level Java type whose simple name is one of the six stereotype simple names, seed that package's other `.java` files into the write set (cheap — six names); or (b) diff computed `frameworkAnnotations` against the pre-run Class rows and pull mismatches into the effective write set.

gitnexus/src/core/ingestion/languages/java/spring-bean-metadata.ts
```diff
@@ -0,0 +1,133 @@
...
+    return SPRING_BEAN_STEREOTYPES.has(imported) ? imported : undefined;
+  }
+
+  const wildcardTarget = wildcardImportTarget(parsed, rawName);
+  if (wildcardTarget === undefined) return undefined;
+
+  // Same-package types and resolved wildcard imports are available only after
+  // finalize/populateNamespaceSiblings. Any such binding wins over guessing
+  // that an unresolved external Spring wildcard supplied the annotation.
+  return hasVisibleTypeBinding(enclosingScope, rawName, indexes) ? undefined : wildcardTarget;
```
magyargergo: **MEDIUM — the wildcard leg fails open on three legal-Java shadowing channels, attributing the wrong bean (the direction this PR itself treats as severe).**

The three gates (lexical scope chain, explicit imports, visible bindings) miss:

1. **Inherited member annotation types** (JLS 6.3 + 6.4.1): `class Base { @interface Service {} } class Holder extends Base { @Service static class Inner {} }` with a lone `import org.springframework.stereotype.*;` — javac binds `Base.Service`, but no gate consults the resolved EXTENDS/MRO chain, so `Inner` gets attributed Spring's `Service`.
2. **Single-static-import of an external nested annotation type**: `import static app.Annotations.Service;` shadows on-demand imports, but static imports are interpreted as `kind:'named'` **without** `targetIncludesImportedName` (see the `static` case in `interpret.ts`), so `explicitImportTargets` skips them; if `app.Annotations` is an external JAR, `hasVisibleTypeBinding` sees nothing either → Spring attributed.
3. **Packages over the sibling cap**: `populateJavaPackageSiblings` disables same-package visibility beyond `MAX_PACKAGE_FILES = 500` (`package-siblings.ts:80-89`), and this suppression path depends entirely on that channel — in a >500-file package the gate silently fails *open*, not closed.

All three need an annotation type named exactly like a stereotype plus a lone Spring wildcard — rare, but each is compiling Java recorded wrongly. Fixtures cover lexical/top-level/member/package-sibling/explicit/multi-wildcard shadows only.

**Remediation:** in this leg, fail closed when (i) *any* named/alias `ParsedImport` binds the simple name regardless of `targetIncludesImportedName`, (ii) the class has a resolved superclass whose owned members include the simple name (the MRO machinery already exists), and (iii) the file's package was sibling-capped (have `populateJavaPackageSiblings` report capped packages).

gitnexus/src/core/ingestion/languages/java/capture-side-channel.ts
```diff
@@ -0,0 +1,54 @@
...
+  readonly classScopeId: ScopeId;
+  readonly annotationNames: readonly string[];
+}
+
+export interface JavaCaptureSideChannel {
+  readonly kind: 'java';
+  readonly classAnnotations: readonly JavaClassAnnotationFact[];
+}
+
+const classAnnotationsByFile = new Map<string, readonly JavaClassAnnotationFact[]>();
```
magyargergo: **MEDIUM — this module-level map is missing the per-pass clear its own precedents define.**

The repo has an explicit lifecycle discipline for exactly this construct: C clears via `clearStaticNames()` and Kotlin via `clearCompanionScopes()`, both from `loadResolutionConfig` (`c/scope-resolver.ts:64-67`, `kotlin/scope-resolver.ts:60-70`), with Kotlin's comment calling that hook "the right lifecycle seam for clearing per-language side-channel state" — naming long-lived server-mode/multi-repo processes and stale cross-run facts as the hazards. The Java resolver defines no `loadResolutionConfig`, and this map has no clear function at all.

Carrying scenario: an MCP server or auto-sync process analyzing multiple repos/branches accumulates one entry per annotated Java class file for the life of the process. Correctness currently survives only on an implicit invariant — `attachSpringBeanCandidateMetadata` reads only current-run `parsedFiles`, each freshly refreshed — which a future reader of `getJavaClassAnnotationFacts` won't know about.

**Remediation:** add `clearJavaClassAnnotationFacts()` and a `loadResolutionConfig` on `javaScopeResolver` mirroring `kotlin/scope-resolver.ts:60-72`.

gitnexus/test/integration/spring-bean-metadata-roundtrip.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      await adapter.executeQuery(
+        `MATCH (c:Class {id: 'Class:src/Widget.java:Widget'}) RETURN c.frameworkAnnotations AS frameworkAnnotations`,
+      ),
+    ).toEqual([
+      {
+        frameworkAnnotations: ['org.springframework.stereotype.Component'],
+      },
+    ]);
+
+    if (process.platform === 'win32') return;
```
magyargergo: **MEDIUM — mid-test `return` silently *passes* the batch-insert third of this test on Windows instead of skipping it.**

On win32 the test reports green while the `batchInsertNodesToLbug` assertions below never ran, so Windows CI shows coverage for a write path it never exercises. The repo's established idiom for this exact situation (batch helper close/reopen vs win32 native locks) is a visible skip: `const itLbugReopen = process.platform === 'win32' ? it.skip : it;` (`lbug-core-adapter.test.ts:26`, also `lbug-close-handle-release.test.ts`, `lbug-orphan-sidecar-recovery.test.ts`).

**Remediation:** split the batch-insert section into a second `it` gated with the existing `it.skip` pattern so the skip is visible in test output.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -5524,6 +5527,7 @@ export class LocalBackend {
     }> = opts.skipEpistemic
       ? Promise.resolve({})
       : this.computeEpistemicBoundary(repo, symId, symType, (sym.name || sym[1]) as string);
+    const beanMetadataPromise = queryClassBeanMetadata(repo.lbugPath, symId, symType);
```
magyargergo: **LOW — the bean lookup starts unconditionally here, ignoring the suppression flags the sibling epistemic probe honors (`opts.skipEpistemic` ternary just above), and on several paths its result is dead work:**

- group cross-repo fan-out `impactByUid` sets `skipEpistemic: true` precisely because per-neighbor enrichment is dead work on the highest-volume path — yet this adds one pool round-trip per Class-typed neighbor per repo, and `cross-impact.ts` consumes only `byDepth`, never `target.bean`;
- the ambiguous-candidate probe (`skipEpistemic + summaryOnly`, up to 6 candidates in a `Promise.all`, documented in-file as pool-exhaustion-prone) fires a bean point-lookup per Class candidate and reads only `impactedCount`/`risk`;
- on the pdg leg, `composeUnifiedPdgImpactResult` rebuilds the target from `pdgResult`, so the bean computed here is dropped before output.

**Remediation:** mirror the epistemic pattern — `const beanMetadataPromise = opts.skipEpistemic || summaryOnly ? Promise.resolve(undefined) : queryClassBeanMetadata(…)`. Each skipped path provably never consumes `target.bean`.

gitnexus/src/core/lbug/csv-generator.ts
```diff
@@ -117,6 +117,15 @@ export const escapeCSVNumber = (
   return String(value);
 };
 
+const formatCSVStringArray = (value: unknown): string => {
```
magyargergo, line 120: **LOW — bare `items.join(',')` silently splits comma-bearing items, diverging from the Cypher write path.**

Empirically probed against real `@ladybugdb/core` COPY with this exact schema + options: a field of `"[has,comma item,plain]"` loads as **three** list items, while the same value through `formatCypherStringArray` round-trips intact — so COPY and CREATE/MERGE persist *different* data for the same input, guarded only by the comment. Unreachable in production today (the sole writer gates every item through `SPRING_BEAN_STEREOTYPES.has(...)`, so items are always one of six canonical FQNs), and the roundtrip test uses comma-free items, which is why nothing catches it.

**Remediation:** quote items the way the sibling list columns do (Route/Community/Process use `'…'`-quoted elements below), or add a dev-time assertion rejecting items matching `[,\[\]'"\n]` so a future non-Spring writer (e.g. the NestJS-style names this PR's own MCP test already simulates) can't silently corrupt the inventory.

magyargergo (changes requested):
## Review: Spring bean candidate inventory

Reviewed at head `d1b2d82d` against merge-base `c2dc9c11`, in a dedicated worktree with a fresh `--pdg` GitNexus index. Evidence: full diff read; `detect_changes` (maps exactly the diff's 48 symbols); upstream `impact` on the persistence/MCP hubs; taint pass (clean — the new list writers are escaped or fed from the closed stereotype set, and the MCP lookup is parameterized); `tsc --noEmit` clean; all 6 PR test files pass locally; CI green. Four domain/cross-cutting expert passes followed, and the headline finding below was **reproduced empirically with this PR's own built CLI**.

The feature is well-architected: it reuses the established `collectCaptureSideChannel`/`applyCaptureSideChannel`/`emitPostResolutionEdges` seams exactly as C/C++/Kotlin do, keeps all Spring knowledge in `languages/java/`, keeps the three Class write paths (CREATE TABLE / COPY / MERGE) consistent, and the v7 `INCREMENTAL_SCHEMA_VERSION` bump correctly forces a full re-analyze on upgrade. The fixture matrix for the fail-closed cases is strong. The verdict is driven by two correctness gaps in the fail-closed guarantee itself:

### Empirical repro for the HIGH finding (incremental staleness)

| Step | `WildcardService.frameworkAnnotations` |
|---|---|
| Full analyze (`import org.springframework.stereotype.*;` + `@Service`) | `["org.springframework.stereotype.Service"]` ✓ |
| Add same-package `public @interface Service {}` → **incremental** analyze ("1 unchanged file rows preserved") | `["org.springframework.stereotype.Service"]` ✗ stale false positive |
| Same tree, `analyze --force` full rebuild | `[]` ✓ |

Incremental and full rebuild diverge, and MCP `context`/`impact` serve a false-positive bean indefinitely — breaking #2411's criterion 4 (fail closed, no misleading results) and the incremental ≡ full-rebuild invariant. Details and remediation options inline.

### Findings (inline)

- **HIGH** — incremental writeback persists stale bean attributions (`spring-bean-metadata.ts`)
- **MEDIUM** — wildcard gate fails open on three legal Java shadowing channels (`spring-bean-metadata.ts`)
- **MEDIUM** — new side-channel map lacks the per-pass clear its C/Kotlin precedents have (`capture-side-channel.ts`)
- **MEDIUM** — mid-test `return` silently passes the batch-insert assertions on Windows (`spring-bean-metadata-roundtrip.test.ts`)
- **LOW** — bean lookup runs as dead work on the deliberately-slimmed impact paths (`local-backend.ts`)
- **LOW** — CSV list items with commas silently split, diverging from the Cypher path (`csv-generator.ts`)

Tested well: all six stereotypes incl. `RestController`'s distinct package, eight negative shadow fixtures, real-DB roundtrip incl. two-item lists, MCP shapes for context+impact, version-gate assertions, and a route-discovery regression assertion in the same fixture. Missing: an incremental-analyze test that mutates annotation-relevant sibling files (HIGH finding), inherited-member / static-import / capped-package shadow cases (MEDIUM finding), and Windows coverage of the batch write path. Records/interfaces/enums are deliberately out of scope and pinned by a unit test — noted as a product boundary, not a defect (`@Component` on records is legal Spring 6, worth a follow-up).

Evidence is valid for head `d1b2d82d`; findings were verified in source (and the HIGH one end-to-end) rather than inferred from graph output alone.

gitnexus/src/core/ingestion/languages/java/spring-bean-metadata.ts
PieceOfFall: Fixed in 6fc1eeda. Incremental Java analysis now compares persisted and fresh frameworkAnnotations and rewrites only drifted files. Added add/remove shadowing regression tests.

gitnexus/src/core/ingestion/languages/java/spring-bean-metadata.ts
PieceOfFall: Fixed in 6fc1eeda. Wildcard resolution now fails closed for conflicting imports, inherited member types, and capped package scans. Added coverage for all three cases.

gitnexus/src/core/ingestion/languages/java/capture-side-channel.ts
PieceOfFall: Fixed in 6fc1eeda. Java annotation facts are now cleared before each workspace analysis pass. Added a stale-state regression test.

gitnexus/test/integration/spring-bean-metadata-roundtrip.test.ts
PieceOfFall: Fixed in 6fc1eeda. Removed the mid-test return and moved the unsupported reopen case into an explicitly skipped Windows test.

gitnexus/src/mcp/local/local-backend.ts
PieceOfFall: Fixed in 6fc1eeda. Bean metadata queries are now skipped for skipEpistemic and summaryOnly paths. Added query-count assertions.

gitnexus/src/core/lbug/csv-generator.ts
PieceOfFall: Fixed in 6fc1eeda. Unsafe CSV list items now fail explicitly instead of being silently corrupted. Added a comma-containing integration test.

magyargergo:
@PieceOfFall Can you please resolve the conflicts? 🙏

magyargergo:
@PieceOfFall could we please include kotlin in this implementation? 🙏 We need to share common functionalities with java.

PieceOfFall:
OK. I’ll add Kotlin support and share the common parts with Java.

azizur100389:
Thanks for the follow-up fixes. I rechecked the current head against current `main`.

The earlier technical review findings look addressed in the Java path: the bean attribution now runs after scope resolution, stale incremental `frameworkAnnotations` are handled by comparing persisted vs fresh values, the side-channel state is cleared per pass, CSV list encoding now fails explicitly on unsafe items, and the skip-epistemic / summary-only lookup path avoids extra bean queries.

I would still keep this as changes-requested for the remaining maintainer scope: the implementation is still Java-only. The current diff adds the Spring bean metadata extraction under `languages/java/*`, Java fixtures, and Java pipeline/MCP tests, but there is no Kotlin equivalent and no shared Java/Kotlin abstraction yet. Since Gergo asked to include Kotlin and share common functionality with Java, this is not ready until the Kotlin provider participates in the same Spring bean inventory behavior or the PR is explicitly narrowed.

One useful next step would be to extract the stereotype/metadata derivation and post-resolution attachment shape into a JVM-shared helper, then wire both Java and Kotlin provider hooks with Kotlin fixtures covering `@Service` / `@Component` / `@RestController` and shadowing behavior.

PieceOfFall:
I’m happy to add the Kotlin support and shared JVM logic. Since `main` is moving quickly, I’d prefer to finish the implementation first and sync with the latest `main` once before final review, rather than repeatedly resolving conflicts along the way. Hope that works.

PieceOfFall:
**Would you be okay with bumping INCREMENTAL_SCHEMA_VERSION from 7 to 8?**

This PR does not change the physical database schema, but it changes the index completeness guarantee: v7 contains Java Spring Bean evidence, while this PR adds Kotlin evidence.

A same-commit v7 index can hit alreadyUpToDate before parse-cache v14 invalidation runs. Kotlin frameworkAnnotations would then remain missing until a forced rebuild or relevant source change.v8 would trigger a one-time full re-analysis and mark the index complete for both Java and Kotlin.

Alternatively, I can add a dedicated Spring Bean feature epoch, but that introduces additional metadata and migration logic. I recommend the v8 bump. Would that approach work for you?

PieceOfFall:
One additional thought before you decide:
I also remembered that #2411 is the first of the eight planned Spring follow-up issues under #1225. Several later tasks will likely add more persisted Spring semantics, including injection, @Bean methods, configuration binding, conditions, AOP, and non-HTTP handlers.
This does not change the correctness issue described above: some durable version gate must run before alreadyUpToDate. The v8 bump remains the smallest fix for this PR.
However, repeatedly using the global schema version for the later Spring phases could force unrelated repositories to re-index. If you prefer to reserve INCREMENTAL_SCHEMA_VERSION for global index invariants, I can instead introduce a dedicated Spring analysis feature stamp, starting with spring.beanInventory: 1.
I am happy to follow either approach. I just wanted to include this longer-term implication before finalizing the implementation.

azizur100389 (changes requested):
Requesting changes for current-head readiness.

`gitnexus/test/unit/spring-bean-schema.test.ts` still expects `PARSE_CACHE_VERSION` to match `^13+`, but current head has `SCHEMA_BUMP = 17` in `gitnexus/src/storage/parse-cache.ts`, so the unit suite should fail after the latest merge-from-main.

The feature is also still Java-only. Since the current maintainer scope asks for Kotlin/shared JVM support, this is not ready unless the PR is explicitly narrowed to Java-only. If Kotlin support lands here, please also settle the durable version gate, either with `INCREMENTAL_SCHEMA_VERSION = 8` or a dedicated Spring feature stamp checked before the up-to-date fast path.

Please rerun the required GitHub Actions checks after the fix; current head only shows the Vercel authorization failure.

magyargergo:
@PieceOfFall Thanks, the latest fixes address all six findings and @azizur100389 confirmed the Java path independently, so the technical review is settled. Answers to your questions:

1. Version gate: go with the bump to 8. It follows what this PR already did for the Java evidence and it is the smallest fix that closes the gap for Kotlin. Let's hold off on the dedicated Spring feature stamp until a later phase of #1225 actually needs it, no point building that machinery now.

2. Merge strategy: agreed, finish the Kotlin and shared JVM work first and sync with main once at the end. No need to chase conflicts along the way.

3. One thing for when you sync: please change the pinned versions in the schema test to minimums instead of exact values. The exact asserts are what broke after the last merge from main and they will keep breaking on unrelated bumps.

Once the Kotlin support is pushed I will approve the workflow runs and do the final review pass.

PieceOfFall:
Thanks, that is clear. I’ll proceed with v8, defer the dedicated Spring feature stamp, finish the Kotlin/shared JVM work, and update the schema assertions to minimum-version checks when syncing with main.

PieceOfFall:
Thanks for the guidance. I followed the suggested direction and extracted the Spring Bean logic into shared framework modules.

### Shared Spring Bean Resolution

The stereotype catalog, metadata derivation, annotation resolution, and post-resolution attachment are now shared by Java and Kotlin through thin language adapters.

Resolution supports:

- Fully qualified annotation names
- Explicit and aliased imports
- A single wildcard import
- Conservative, fail-closed behavior for lexical, inherited, same-package, multiple-wildcard, and incomplete-visibility conflicts

### Kotlin Support

Kotlin now participates in the same Bean inventory pipeline without requiring a second AST traversal. Annotation and package facts are captured during the existing scope traversal, persisted through the Kotlin `ParsedFile` side-channel, and restored from cache.

The fixtures cover:

- `@Service`, `@Component`, and `@RestController`
- Fully qualified annotations and import aliases
- Nested classes
- Abstract, data, sealed, and value classes
- Lexical and same-package shadowing

Interfaces, objects, enums, annotation classes, and composed annotations remain excluded.

Java and Kotlin now produce the same canonical `frameworkAnnotations` values and MCP `Bean` output.

### Shared JVM Package Visibility

I also extracted Java’s package-sibling handling into a shared JVM helper used by both languages.

It distinguishes the default package from package parsing failures and marks visibility as incomplete when source content is missing, parsing fails, or the package-size limit is reached. Wildcard attribution therefore remains conservative.

Incremental annotation drift detection now covers `.java`, `.kt`, and `.kts`, including same-package shadow additions and removals.

### Cache and Index Compatibility

After syncing with the current `main`, incremental schema v8 was already assigned to [[#2550](https://github.com/abhigyanpatwari/GitNexus/issues/2550)](https://github.com/abhigyanpatwari/GitNexus/issues/2550). Rather than immediately advancing to v9, I kept v8 and added independently versioned analysis capability stamps:

- `graph.class-framework-annotations@1`
- `spring.bean-inventory@1`, applied only to JVM repositories

Missing, malformed, mismatched, or unknown capability stamps trigger a one-time full rebuild before the fast path can be used.

The parse-cache schema is now v20. Upstream v19 is reserved for the Java enum-constant and anonymous-class identity changes, while v20 invalidates caches created before the new Java/Kotlin Spring side-channel data was available.

### Warm-Cache Behavior

Kotlin also opts into a generic cached-source policy. Warm durable-`ParsedFile` runs now load source content only for cache misses instead of loading every Kotlin source file.

Other resolvers retain their existing behavior, and the shared ingestion pipeline contains no language-specific checks.

### Validation

The full test suite passes:

- 699 test files passed
- 14,466 tests passed
- 49 tests skipped

Existing Java Bean behavior, MCP context and impact queries, route and `HANDLES_ROUTE` behavior, incremental analysis, parse caching, and storage regressions remain covered.

magyargergo:
I'll get back to you as soon as possible but now I'm busy working on getting the GitNexus Engineering Tool Kit done.

PR #2513, opened 2026-07-16
test(embeddings): cover File-row deletion

100yenadmin:
## Summary

- pin batched incremental deletion of fallback `File` embedding rows while preserving survivor rows
- pin the singular `deleteNodesForFile` path for a `File`-owned embedding
- correct the stale test comment that said `File` could not own embeddings

## Why

Follow-up coverage for #2455 and the landing fix in `e814c5a1`. The production matcher now includes `File`, but the existing integration fixture only seeded `Function` embeddings, so a regression could silently restore orphaned vectors in docs/static repositories.

## Test

- `npx vitest run test/integration/lbug-delete-nodes-for-files.test.ts --reporter=dot` (4 passed)


magyargergo:
Thanks 🙏

PR #2522, opened 2026-07-16
fix(scope-resolution): resolve callable reference flows (#2437)

### LINKED ISSUE #2437: impact/context report 0 upstream dependents for provider-hook implementations (emitScopeCaptures et al.)
Found while reviewing PR #2436 (fix for #2432).

`impact({target: "emitCppScopeCaptures", direction: "upstream"})` returns **0 impacted / LOW risk**, but the function is live on every C++ parse: it is consumed through the `LanguageProvider` hook indirection —

- `gitnexus/src/core/ingestion/languages/c-cpp.ts` assigns it as an object property: `emitScopeCaptures: emitCppScopeCaptures`
- `gitnexus/src/core/ingestion/scope-extractor-bridge.ts` (`extractParsedFile`) invokes it as `provider.emitScopeCaptures(...)`

The call graph does not model "function referenced as an object-literal property value, later invoked through that property", so every provider hook implementation (`emitScopeCaptures`, `cfgVisitor`, `preprocessSource`, `collectCaptureSideChannel`, …) shows a false-safe empty blast radius. Reviewers following the repo's own "MUST run impact before editing" rule get a misleading LOW on some of the hottest code in the pipeline.

Possible directions: emit a CALLS/REFERENCES edge for shorthand property values whose identifier resolves to a function symbol, or a dedicated `IMPLEMENTS_HOOK` edge from the provider-object property to the function.

magyargergo:
## Summary

Fixes #2437. GitNexus now models callable values from registration through assignment, aliasing, storage, argument passing, and the eventual invocation site. This removes the false `0 upstream / LOW risk` result for provider hooks while keeping registration as `USES` and emitting `CALLS` only at real call sites.

This also addresses the review feedback in #2522:

- refreshes the JavaScript/TypeScript scope-capture benchmark fingerprints after verifying the capture expansion remains within the scaling budget;
- reports over-cap property-dispatch keys at warning level and includes them in resolver statistics.

## Implementation

- Adds JSON-safe callable-flow facts to `ParsedFile` and bumps the parse cache schema to 17.
- Adds a shared, bounded, dependency-indexed inclusion solver for callable seeds, copies, aliases, address/reference operations, store/load pairs, actual-to-formal propagation, and invocation.
- Resolves calls at the actual call site across all 16 registered language providers, including callable values assigned to variables and passed through function or method parameters.
- Preserves lexical/import resolution, overload context, shadowing, static linkage, and existing precise call edges.
- Runs property dispatch before callable propagation so wrapper calls can seed actual-to-formal flow.
- Keeps shared ingestion language-neutral; language-specific syntax is supplied through provider options.

## C and C++ coverage

- C function pointers, copied pointers, pointer-to-pointer flows, prototypes, and static linkage.
- C++ function pointers and references, references to pointer variables, member-function pointers, cv/overload context, split declaration/assignment, imported targets, and actual-to-formal propagation.
- Virtual member-pointer calls expand through the MRO; non-virtual calls remain exact.

## Safety and observability

- Bounds callable targets, addresses, facts, and worklist propagation; partial results are not emitted after a bound is exceeded.
- Validates cached callable facts before use.
- Emits auditable reasons/confidence on synthesized edges.
- Warns and counts skipped over-cap property keys.
- PDG taint checks found no source-to-sink findings in the changed ingestion/storage paths.

## Validation

- 260 focused CLI/Core tests across the worker round-trip, callable-flow/provider matrix, TypeScript value references, ParsedFile storage, PDG callee IDs, worklist behavior, captures, and scope extraction.
- 384 GitNexus Web tests.
- `node --import tsx bench/scope-capture/measure.mjs --check` — all 14 language benchmark fixtures pass; every measured ratio remains below the 1.5 ceiling.
- `npm run build` — shared, CLI/Core, and Web production builds pass.
- `npx tsc --noEmit` and `npx tsc -b --noEmit` — pass.
- `npx eslint .` — 0 errors (repository baseline warnings remain).
- Full PDG index refresh completed successfully: 216,155 nodes, 459,248 edges, 997 clusters, 300 flows.

The monolithic local `npm test` run was also attempted; native LadybugDB workers exited with code 134 in this container, matching the repository's documented environment/file-locking limitation. The affected test matrix above passes, and the pushed head will receive the full platform CI matrix.

## Result

After a PDG-enabled reindex, `impact({ target: "emitCppScopeCaptures", direction: "upstream" })` reports 10 affected symbols at MEDIUM risk instead of the previous false-safe `0 / LOW` result.


azizur100389:
﻿Thanks Gergo, I reviewed #2522.

The design makes sense to me: separating registration as `USES` from invocation as synthesized `CALLS` at member-call sites is the right shape, and the tests cover the main intended cases: same-file/cross-file/aliased/shorthand registrations, JS parity, non-callable negatives, per-language partitioning, and fan-out cap behavior.

A couple of things I think should be addressed before merge:

1. CI is currently red because `tests / benchmarks (GITNEXUS_BENCH)` fails on scope-capture fingerprints.

The benchmark log shows expected drift from this PR's TS/JS capture additions:

```text
[scope-capture --check] FAIL: typescript: capture fingerprint drift
  got 25de86fd3377132c4e35d3d98f4f94a58e0cfeb7c22948a8ea3be4e793be74fd
  expected 3f44a4a6892698df2d145c8ff2812c3b318807648983c88aca28fbd694f172f9

[scope-capture --check] FAIL: javascript: capture fingerprint drift
  got 5567dd47e7ba29821a518c4a9852adc3b774e25ef3e7a6e2b3ecb7b59ddab73c
  expected d72f03c6c502235d2d4b74d66baa5c7d361f040d7a1b72e84acad61210d05ae8
```

Since the new `@reference.value-ref` / `@reference.property-key` captures intentionally change TS and JS capture output, could you update `gitnexus/bench/scope-capture/baselines.json` with the new fingerprints after confirming the scaling ratios stay within budget? Right now this makes CI Gate fail, separate from the Vercel auth failure.

2. Dropped property-dispatch keys are still effectively debug-only, not surfaced in returned stats/normal diagnostics.

`emitPropertyDispatchCalls()` returns `skippedKeys`, and `runScopeResolution()` logs it with `logger.debug(...)`, but the value is not returned through `RunScopeResolutionStats` / phase aggregate and won't normally be visible unless debug logging is enabled. Since an over-cap key means no synthesized `CALLS` are emitted for that key and the original false-safe gap can reappear, I think this should be more visible.

Could we either include the skipped-key count in scope-resolution stats/progress output, or log it at warning level with enough context? That would better match the PR description's "dropped over-cap keys are logged, never silent" claim.

Other notes:

- The implementation's `USES` vs `CALLS` split looks consistent in `reference-site.ts`, `edges.ts`, `emit-references.ts`, and `references-to-edges.ts`.
- The PDG resolved-callee sink is correctly gated to real call refs in `emitReferencesViaLookup`, while property-dispatch calls feed the sink as call sites through `tryEmitEdge`.
- Most checks are passing: quality, typecheck, platform-sensitive tests, coverage shards, CodeQL, packaged install smoke, etc. The remaining code-related blocker I saw is the benchmark fingerprint failure above.

azizur100389:
Thanks Gergo, I rechecked the current head against current `main`.

I think this still needs changes before merge. The failing GitHub CI is in the same resolver surface this PR changes: `tests / ubuntu / coverage 2/3` fails 8 assertions in `test/integration/resolvers/cpp.test.ts`, mostly duplicate or incorrect `CALLS` edges in C++ inheritance-lattice / dependent-base lookup cases. Examples from the log:

- virtual diamond lookup expected 1 call and got 2;
- non-virtual diamond suppression expected 0 calls and got 1;
- nested templated base normalization expected 1 call and got 2;
- dependent-base / name-hiding cases emit extra calls where the tests expect zero or one.

`tests / windows-latest (platform-sensitive) 1/3` also fails in `test/integration/cli-limit-e2e.test.ts` because `analyze --force` exits abnormally after a long run.

Given the PR changes core scope-resolution / callable-flow behavior and the current CI regressions are real resolver assertions, I would hold merge until those C++ call-resolution regressions are fixed and the CI is green.

PR #2539, opened 2026-07-17
fix(dart): extract extension type symbols

### LINKED ISSUE #2538: Dart `extension type` declarations produce zero symbol nodes (grammar recognizes them, resolution layer skips them)
## Summary

Dart **`extension type` declarations** (Dart 3.3+) produce **zero symbol nodes** in the knowledge graph. The `extension type` itself is never extracted, so `context` / `impact` / `rename` / `cypher` all return "not found" for it — even though the file is indexed and the tree-sitter grammar recognizes the syntax.

This is distinct from the already-tracked #584 (general "no Dart symbols") and #1926 (six other parsing-layer gaps). Neither covers `extension type` — a search for `extension_type` across issues returns nothing. Regular Dart `extension` (extension methods) and plain classes **are** extracted correctly, as is `describe` (top-level function, fixed since a recent release). Only `extension type` is affected.

`extension type` is the idiomatic Dart 3 way to build zero-cost wrapper types (identity types, units, typed IDs). Codebases that lean on it for domain modeling end up with their core value objects invisible to GitNexus.

## Environment

- GitNexus: **1.6.9** (latest stable)
- Node.js: v22.23.1
- Dart SDK: 3.11.5 (stable)
- OS: macOS 26.5 / arm64

## Minimal reproduction

A single self-contained file — no dependencies, no `pub get` needed.

`pubspec.yaml`:
```yaml
name: exttype_repro
environment:
  sdk: ^3.3.0
```

`lib/types.dart`:
```dart
/// A zero-cost wrapper — Dart 3.3 extension type. NOT indexed.
extension type const UserId(String value) {}

/// Extension type with a body and a method.
extension type Celsius(double degrees) {
  double toFahrenheit() => degrees * 9 / 5 + 32;
}

/// A plain class — indexed correctly, for comparison.
class PlainBox {
  const PlainBox(this.contents);
  final String contents;
  int size() => contents.length;
}

/// A top-level function — indexed correctly.
String describe(PlainBox box) => 'box of ${box.size()}';
```

Then:
```
git init && git add -A && git commit -m repro
npx gitnexus analyze --index-only
npx gitnexus cypher --repo <name> "MATCH (n) WHERE n.name IS NOT NULL RETURN n.name, n.filePath"
```

## Expected vs. actual

Symbol node count for each declaration in `lib/types.dart`:

| Declaration | Kind | Expected | **Actual** |
| --- | --- | --- | --- |
| `UserId` | `extension type` | 1 | **0** |
| `Celsius` | `extension type` | 1 | **0** |
| `toFahrenheit` | method of `Celsius` | 1 (child of `Celsius`) | 1 — **orphaned** (parent missing) |
| `PlainBox` | `class` | 1 | 1 ✅ |
| `size` / `contents` | member of `PlainBox` | 1 each | 1 each ✅ |
| `describe` | top-level function | 1 | 1 ✅ |

Two things stand out:

1. **The `extension type` declaration node is never created.** `UserId` and `Celsius` are absent from the graph entirely.
2. **The extension type's *body* is still walked** — `toFahrenheit` (Celsius's method) **is** extracted, but as an **orphan**: its enclosing type `Celsius` doesn't exist as a node, so the member has no owner. This suggests the traversal descends into the body but the declaration-level query/extractor doesn't match the `extension_type_declaration` node.

## Likely cause (pointer for the fix)

The tree-sitter-dart grammar **does** expose the node — `grep extension_type node-types.json` → `"extension_type_declaration"`. So the parser recognizes the construct; what's missing is a resolution-layer query/extractor mapping that node to a symbol (the same shape as the F24–F29 findings in #1926, e.g. F26 for `static const` fields).

For reference, the two `extension_type` header shapes to handle:
- primary-constructor form: `extension type const UserId(String value) {}` (and non-`const`)
- named-representation form: `extension type Celsius(double degrees) { ... }` with a body

## Real-world impact

In our production Flutter codebase, the three identity types (`PlayerId`, `PlayerSeat`, `GameId`), all declared `extension type const ...(String value) {}`, have **zero** nodes. They are the most-referenced value objects in the domain, and they're invisible to impact analysis — the exact case GitNexus is meant to help with. We currently fall back to `grep` for anything touching them.

Happy to test a patch against the repro above or our larger codebase.

azizur100389:
## Summary
- Add a length-preserving Dart preprocessor for `extension type` headers so existing Dart extension extraction can materialize extension type symbols.
- Keep declaration name offsets stable by replacing `type` / `const` header text with spaces before parsing.
- Map Dart `extension_declaration` containers to `Class` ownership so extension type methods attach to their symbol.
- Add regression coverage for empty-body, method-body, and generic extension type declarations.

Closes #2538

## Validation
- `npx vitest run test/integration/resolvers/dart-coverage.test.ts test/integration/resolvers/dart.test.ts test/unit/method-extraction.test.ts --maxWorkers=2 --testTimeout=120000`
- `npx vitest run test/integration/resolvers/dart-coverage.test.ts --maxWorkers=2 --testTimeout=120000`
- `npx prettier --check --ignore-unknown src/core/ingestion/languages/dart.ts src/core/ingestion/languages/dart/captures.ts src/core/ingestion/languages/dart/extension-type-preprocess.ts src/core/ingestion/utils/ast-helpers.ts test/integration/resolvers/dart-coverage.test.ts test/fixtures/lang-resolution/dart-extension-types/models.dart`
- `npx eslint src/core/ingestion/languages/dart.ts src/core/ingestion/languages/dart/captures.ts src/core/ingestion/languages/dart/extension-type-preprocess.ts src/core/ingestion/utils/ast-helpers.ts test/integration/resolvers/dart-coverage.test.ts` (0 errors; existing non-null assertion warnings in pre-existing files)
- `npx tsc --noEmit`
- `npm run build`
- `npm test -- --maxWorkers=2 --testTimeout=120000` ran to completion locally on Windows but failed in unrelated native/env-sensitive suites: FTS extension unavailable, symlink/EPERM behavior, MCP startup timeout, skip-git timeouts, and worker fork errors.

## Review
- Pre-PR risk review: no blocker.
- Pre-PR security/source-integrity review: no blocker after preserving declaration name offsets.


azizur100389:
Addressed the Tri review implementation finding in \

gitnexus/src/core/ingestion/languages/dart/captures.ts
```diff
@@ -522,6 +528,25 @@ function emitHeritage(classNode: SyntaxNode, out: CaptureMatch[]): void {
...
+  if (nameNode === null) return;
+
+  const bodyStart = extensionNode.text.indexOf('{');
+  const header = bodyStart === -1 ? extensionNode.text : extensionNode.text.slice(0, bodyStart);
+  const implementsIndex = header.indexOf('implements');
+  if (implementsIndex === -1) return;
+
+  const className = nameNode.text;
+  const interfaces = header.slice(implementsIndex + 'implements'.length);
+  for (const rawInterface of interfaces.split(',')) {
```
magyargergo: **[MEDIUM] Naive comma-split breaks on multi-arg generic interfaces**

`emitExtensionImplementsHeritage` splits the `implements` clause text on every comma, so a multi-type-argument generic interface gets split mid-generic and emits a spurious extra IMPLEMENTS edge.

Empirically reproduced against this PR's code:
```dart
class Comparator<A, B> {}

extension type Pair(String a) implements Comparator<String, int> {
  String describe() => a;
}
```
produces **two** heritage markers instead of one:
- `implements:Comparator:Pair` (correct)
- `implements:int:Pair` (bogus — `Pair` does not implement `int`)

The pre-existing structural heritage path for regular classes (`emitHeritageMarkers`, ~line 550 in this same file) doesn't have this problem because it walks the AST's `type_identifier` named children instead of parsing raw text — tree-sitter already segments `Comparator` from its `<String, int>` type_arguments as separate children. `extension_declaration` has no `interfaces` field in the vendored grammar (checked `vendor/tree-sitter-dart/src/node-types.json`), which is presumably why this fell back to text parsing — but the fallback needs to be bracket-depth-aware when splitting on commas (skip commas nested inside `<...>`).

Not covered by the new tests — the fixture only exercises single-arg generics (`SequenceLike<T>`). Suggest adding a regression case with a 2-arg generic interface alongside the fix.

gitnexus/bench/scope-capture/baselines.json
```diff
@@ -79,8 +79,10 @@
     "_rebaselined_2522_review_fixes": "PR #2522 review fixes: assignment target:/result: fields join the shared fallback. Prior 7687ee2466e16020a12440a03fbda53e63aa05f94b4481f6133c09867a0d560d -> 115c5da807e36bb12fdeba28e44f2b6484ef322ff26c19fa0f191febaf774248; scaling ratio re-verified within budget."
   },
   "dart": {
-    "fingerprint": "66a46d5ff09f3d11b2771db0f48596fe7057e95c5bc8f56241fdb911137298c3",
+    "fingerprint": "ab1208df37ee83371ebea312cc837dd94d3f963febd65acb501409a675269108",
     "scaling_budget": 1.5,
+    "_rebaselined_2538": "#2538: Dart extension type headers are preprocessed into normal extension declarations before scope capture, so extension type symbols and their methods are now emitted. Intentional Dart-only capture fingerprint drift; CI measured scaling 1.042 < 1.5.",
+    "_rebaselined_2538_implements": "#2538 tri-review follow-up: Dart extension type implements clauses now emit heritage markers and fixture coverage asserts IMPLEMENTS edges. Prior 70b882239d6d289541b9cbc2ae1ab86151277c238ec53f44e1066f6b9b6ad539 -> ab1208df37ee83371ebea312cc837dd94d3f963febd65acb501409a675269108; scaling 1.024 < 1.5.",
```
magyargergo: **[LOW] Changelog note cites a fingerprint that was never committed**

The `_rebaselined_2538_implements` note claims "Prior `70b882239d6d...`" but the diff shows the actual prior committed value was `66a46d5ff09f...` (the `_rebaselined_callable_flow_2522_review` line below still shows that as its own "New" value, confirming it was the real baseline immediately before this PR). Cosmetic only — the `--check` gate validates correctly against the real current value (verified: fingerprint matches, scaling 1.039 < 1.5). Worth fixing so a future rebaseline investigation isn't misled about the history.

magyargergo (changes requested):
GitNexus-assisted review. Verified in a temporary worktree at the PR head SHA: full Dart test suite (465 tests across dart-coverage, dart, dart-visitor, dart-import-resolver, dart-type-extractor, method-extraction) passes, `tsc --noEmit` clean, and `bench/scope-capture/measure.mjs --check` confirms the dart fingerprint rebaseline is legitimate. One correctness finding below on the new IMPLEMENTS-heritage extraction, plus a minor changelog-note nit. Confirmed `extension_declaration` is Dart-only in the tree-sitter grammar set (Swift uses a different node type, `extension`), so the shared `CONTAINER_TYPE_TO_LABEL` edit is safely scoped to Dart.

**Verdict: REQUEST CHANGES** — the heritage-parsing bug silently produces incorrect graph edges (no crash, no failing test) for a plausible, valid Dart pattern, and IMPLEMENTS heritage is one of this PR's two headline additions.

azizur100389:
Pushed review fixes in \777e4ee8\:\n\n- Fixed extension-type \implements\ parsing for multi-arg generics by splitting only on top-level commas.\n- Added regression coverage for \Comparator<String, int>\, including negative assertions that no bogus \int\ heritage marker/edge is emitted.\n- Corrected the Dart scope-capture rebaseline note to cite the actual prior committed fingerprint and the new measured fingerprint/scaling.\n\nValidation run locally:\n- \
px vitest run test/integration/resolvers/dart-coverage.test.ts --maxWorkers=2 --testTimeout=120000\\n- \
px vitest run test/integration/resolvers/dart-coverage.test.ts test/integration/resolvers/dart.test.ts test/unit/method-extraction.test.ts --maxWorkers=2 --testTimeout=120000\\n- \
pm run build\\n- \
px tsc --noEmit\\n- Prettier check + ESLint (0 errors; existing warnings only)\n- \
ode --import tsx bench/scope-capture/measure.mjs --check\\n- \detect_changes\ low risk, no affected processes

gitnexus/src/core/ingestion/languages/dart/captures.ts
azizur100389: Fixed in `777e4ee8`: replaced the raw comma split with a bracket-depth-aware top-level comma splitter for extension-type implements clauses. Added regression coverage for `Comparator<String, int>`, including negative assertions that no bogus `int` heritage marker/edge is emitted.

gitnexus/bench/scope-capture/baselines.json
azizur100389: Fixed in `777e4ee8`: corrected the rebaseline note to cite the actual prior committed Dart baseline `66a46d5ff09f3d11b2771db0f48596fe7057e95c5bc8f56241fdb911137298c3`, the new fingerprint `ba93c90dcd341259e8e088816bc8c76ad27882419f665e35c056dc22fa54cf73`, and the clean benchmark scaling `0.945 < 1.5`.

magyargergo (approved):

PR #2553, opened 2026-07-18
fix: rewrite CodeRelation.type IN predicates to scalar equality — context()/impact() dropped callers (#2508)

### LINKED ISSUE #2508: context()/impact() drop Function callers when filtering CodeRelation.type with IN
## Summary

`context()` and `impact(upstream)` can omit a production Function caller even though its `CALLS` edge exists. The test-file caller is returned twice instead.

## Public reproduction

Repository: `abhigyanpatwari/GitNexus`  
Commit: `226ec6cf99d7728321561fc45fa9fa62b741f8b4`  
GitNexus: 1.6.9  
LadybugDB: 0.18.1

```sh
git clone https://github.com/abhigyanpatwari/GitNexus.git
cd GitNexus
git checkout 226ec6cf99d7728321561fc45fa9fa62b741f8b4
gitnexus analyze

gitnexus context classifyOutcome \
  --repo "$PWD" \
  --file gitnexus-shared/src/integrations/resilient-fetch.ts

gitnexus impact classifyOutcome \
  --repo "$PWD" \
  --file gitnexus-shared/src/integrations/resilient-fetch.ts \
  --direction upstream
```

The graph contains:

```text
resilientFetch          --CALLS--> classifyOutcome
resilient-fetch.test.ts --CALLS--> classifyOutcome
```

Actual results:

- `context()` returns `resilient-fetch.test.ts` twice and omits `resilientFetch`.
- `impact()` returns `impactedCount: 0`, `risk: LOW`.
- With `--include-tests`, impact returns only the test caller.

## Root-cause evidence

This command returns the test caller twice:

```sh
gitnexus cypher --repo "$PWD" \
  "MATCH (caller)-[r:CodeRelation]->(target)
   WHERE target.id = 'Function:gitnexus-shared/src/integrations/resilient-fetch.ts:classifyOutcome'
     AND r.type IN ['CALLS']
   RETURN caller.id AS callerId, caller.name AS caller"
```

Changing only the predicate returns both correct callers:

```sh
gitnexus cypher --repo "$PWD" \
  "MATCH (caller)-[r:CodeRelation]->(target)
   WHERE target.id = 'Function:gitnexus-shared/src/integrations/resilient-fetch.ts:classifyOutcome'
     AND r.type = 'CALLS'
   RETURN caller.id AS callerId, caller.name AS caller"
```

## Suggested fix

Replace relationship-property list predicates such as:

```cypher
r.type IN $relTypes
```

with individually bound scalar comparisons:

```cypher
(r.type = $relType0 OR r.type = $relType1)
```

Apply this to context and impact queries, then audit other `CodeRelation.type IN` usages.

Add a real-LadybugDB regression test asserting exact caller IDs, not only result counts.

Related: #2129 reported the same under-reporting symptom but was closed with a different root cause.

magyargergo:
Fixes #2508.

## Root cause (empirically confirmed, sharper than the issue's claim)

An index **written** by `@ladybugdb/core` >= 0.18.1 through the analyze pipeline's bulk CSV `COPY` into the multi-pair `CodeRelation` rel table gets an on-disk layout on which relationship-property `IN` predicates (`r.type IN [...]` and `r.type IN $param`) return wrong rows — the issue's repro returns the test-file caller **twice** and **drops** `resilientFetch` — while scalar `=` / OR-chains return correct rows on the same DB.

Version matrix (built with the issue's exact recipe — gitnexus 1.6.9 at commit `226ec6cf`, `@ladybugdb/core` pinned per cell):

| DB written by (COPY pipeline) | read 0.18.0 | read 0.18.1 | read 0.18.2 |
|---|---|---|---|
| 0.18.0 | correct | correct | — |
| 0.18.1 | **wrong** | **wrong** | **wrong** |
| 0.18.2 | — | — | **wrong** |

Scalar-equality predicates are correct in every cell; node-property `IN` (`n.id IN $ids`) is unaffected in every cell. Small synthetic DBs (CREATE or COPY, up to 200k rels / 2 pairs) do **not** trigger it — the layout trigger needs full-pipeline scale, so no cheap CI fixture exists.

**Consequence:** pinning the dependency cannot fix this — indexes already written by >= 0.18.1 stay broken under every reader, and 0.18.2 still writes the affected layout. The query rewrite the issue suggested is the only robust fix. An upstream LadybugDB report is a follow-up (repro assets preserved).

## Change

- New `relTypeEquals(alias, types, paramPrefix?)` in `core/lbug/query-params.ts` — builds `(r.type = $relType0 OR ...)` with per-type bound params; empty list keeps `IN []` match-nothing semantics via `FALSE`.
- All **13** rel-type `IN` sites rewritten: `_contextImpl` (5 queries), `_runImpactBFS` (upstream/downstream), `_traceImpl`, `computeEpistemicBoundary` (2), `pdgLayerStatus`, `bfsReachableBlocks`, wiki `getFilesWithExports`.
- **Tripwire test**: source scan fails on any reintroduced rel-type `IN` predicate (verified failing against the pre-rewrite tree, listing exactly the un-rewritten sites).
- **Regression test** (real LadybugDB, per the issue's ask): seeds the #2508 cross-pair shape — production `Function -> Function` caller + test `File -> Function` caller — and asserts **exact caller IDs** (never counts) through the real `context`/`impact` tool paths, including `includeTests` filtering and no-duplicate guarantees.
- Four unit suites mocked `executeParameterized` by matching the literal `'r.type IN'` query text; their dispatch is re-keyed to `$frontierIds` with every assertion unchanged.

## Verification

- End-to-end against a genuinely broken 0.18.1-written index at the issue's pinned commit: post-fix `context classifyOutcome` returns **both** callers and `impact --direction upstream` reports `impactedCount: 1` with `resilientFetch` (pre-fix: doubled test caller / dropped prod caller / `impactedCount: 0, risk: LOW`).
- Full vitest suite: 12,679 passed; remaining failures reproduced as the repo's known load-flake class (all pass in isolated re-runs).
- `detect_changes(compare: main)`: exactly the intended 18 symbols across 11 files, no unexpected execution flows.

Note for users: on indexes written by ladybug >= 0.18.1, `impact`/`context` results may report **more** callers after this change — that is the fix working, not a regression. Raw `cypher` queries users write themselves can still hit the DB bug; a docstring warning is left as a follow-up.

## Upstream status (added after triage)

The defect is tracked upstream as [LadybugDB/ladybug#692](https://github.com/LadybugDB/ladybug/issues/692) (open): *"REL TABLE GROUP property IN predicate duplicates rows and omits another endpoint table"* — reported there on core **0.16.1**, so the evaluation defect predates 0.18.1 and is data/layout-dependent; the version matrix above describes when **this repository's data** starts triggering it (deterministically with >= 0.18.1 writers). Verified in this investigation: the newest nightly (`0.19.0-dev.20260717`) still misevaluates our affected index, and the v0.18.1→v0.18.2 window contains no relevant storage change — **no upstream fix exists yet**, which makes the scalar-equality rewrite the right call regardless of pinned version.


magyargergo:
Superseeded by https://github.com/LadybugDB/ladybug/pull/699

Issue #2550, opened 2026-07-18
scope-resolution: instance-scoped free-call resolution for Java (extensible to C#/other class-instance languages)


magyargergo:
## Context

Follow-up to #2545 / PR #2549.

#2545 was a TypeScript bug (unqualified `fetch()` resolving to an unrelated same-file object-literal method) with two root causes:

1. **Grammar-level scope leak**: object literals had no scope boundary, so a method's name auto-hoisted past the literal into the enclosing scope.
2. **The actual leak vector**: finalize's per-file bindings bucket (`materializeBindings` in `gitnexus-shared`) flattens every local declaration in a file onto module scope regardless of true nesting, so free-call resolution could still hit the leaked binding through that channel.

Both were fixed for TS/JS, and audited/extended to every other language with the same grammar shape. Two more real cases were found: **Kotlin** (`object { ... }` anonymous expressions) and **Java** (`new Runnable() { ... }` anonymous class bodies).

## What shipped (PR #2549)

- Kotlin: `(object_literal) @scope.class` + wired into the existing `isBuiltInName` guard (Kotlin already populates `LanguageProvider.builtInNames` with real platform/stdlib names like `println`) — **fully closes the equivalent leak**, verified.
- Java: `(object_creation_expression (class_body) @scope.class)` — closes the grammar-level scope leak (root cause 1). This part is real, safe, and shipped.

## What's still open for Java

Java has no `builtInNames` list — it has no free-standing global functions the way TS's `fetch` or Kotlin's `println` are. The names that leak for Java (`run`, `get`, `apply`, `compare`, `compareTo`, `accept`, `test`, ...) are the single-abstract-method names of core JDK functional interfaces — **real, extremely common method names**, not platform globals.

We attempted to close this with a narrowly-scoped `ScopeResolver.freeCallBuiltInNames` hook (consumed only by `free-call-fallback.ts`'s guard, decoupled from `LanguageProvider.builtInNames`'s other, unrelated consumers — `parse-worker.ts`'s call-extraction gate and `type-env.ts`'s return-type lookup, both of which assume a builtin name is *never* a real declaration, which populating `LanguageProvider.builtInNames` directly with SAM method names would have violated — verified that approach breaks legitimate `handler.run()`-style explicit-receiver calls).

[...]

PR #2558, opened 2026-07-18
feat(java): model enum constant bodies as first-class instances; JLS 13.1 anonymous naming

### LINKED ISSUE #2555: scope-resolution: Java anonymous-class instance model leftovers — enum constant bodies unmodeled, nested-host naming deviates from javac
## Context

Follow-up to #2550 / PR #2549, which modeled Java anonymous class bodies (`new Runnable() { ... }`) as first-class instances: synthesized `Worker$N` Class nodes, method ownership, receiver typeBindings, inheritance edges to the constructed type, and an instance-ownership gate on free-call resolution. An empirical review pass (probe fixtures run against both the PR head and the pre-PR baseline) verified the model across class/enum/interface/record hosts, nested anonymous bodies, field initializers, and local classes — and identified two remaining gaps that ship as documented limitations rather than fixes. This issue tracks them.

## 1. Enum constant bodies are unmodeled (the remaining leak shape)

```java
public enum EnumHost {
    A {
        public void hook() { System.out.println("constant body"); }
    };

    public abstract void hook();

    public void caller() {
        hook(); // fine — but see below
    }
}
```

An enum constant body (`A { ... }`) is javac's other anonymous-class shape (compiled as `EnumHost$1`), but in tree-sitter-java it is an `enum_constant` with a body — **not** an `object_creation_expression` — so none of the #2550 machinery fires:

- No `EnumHost$1` Class node is synthesized; the body's methods attribute to the **enum itself** (`Method:...:EnumHost.hook`).
- Because ownership lands on the host, an unrelated same-file bare call matching a constant-body method name still resolves through finalize's per-file module-scope bucket — the exact defect class #2550 closed for `object_creation_expression` bodies remains open for this shape.

Empirically verified during the PR #2549 review (scenario S1b): behavior is byte-identical to pre-PR, so this is an unchanged-old limitation, not a regression. The fix direction is the same synthesis pattern extended to the `enum_constant`-with-body node shape (naming shares the host's `$N` counter with anonymous bodies, matching javac).

## 2. Nested-host naming deviates from javac binary names

For an anonymous body inside a **nested** type, the synthesized name anchors on the **top-level** host:

```java
public class EnumWrap {
    enum Mode {
        ON;
        void install() {
            Runnable r = new Runnable() { public void run() {} }; // GitNexus: EnumWrap$1
        }                                                          // javac:    EnumWrap$Mode$1-style
    }
}
```

All four layers (structure-phase node id, owner walk, scope-side declaration, receiver typeBinding) share one naming authority (`synthesizeJavaAnonymousClassName`), so the model is internally consistent and collision-free — numbering is per top-level host in source order, verified for multiple hosts per file and nested anon-in-anon. The deviation is cosmetic unless a repo literally declares a source class named like `X$1` (legal but conventionally reserved), in which case the synthesized id could collide with a real class node.

## Also noted during the same review (pre-existing, unrelated to the PR — split out if picked up)

- Record container nodes never materialize, so `Record:` HAS_METHOD edges dangle (reproduces on a plain record with no anonymous classes).
- A chained call on a new-expression (`new Local().inner()`) emits only the constructor edge (compound-receiver limitation).

## Extension roadmap (deliberate deferrals from #2550, for reference)

- **C#** opt-in to `ScopeResolver.freeCallsRequireInstanceOwnership` — needs C#'s own instance modeling first (lambdas/local functions, not anonymous class bodies).
- **Kotlin** structural opt-in — its #2545 fix is name-keyed via `builtInNames`; the structural gate would also cover non-builtin same-file collisions.

magyargergo:
## Summary

Closes #2555.

Completes the Java anonymous-class instance model from #2550 (PR #2549) by covering its two documented leftovers:

### 1. Enum constant bodies (`enum E { A { void hook(){} } }`)

javac's other anonymous-class shape — a tree-sitter `enum_constant` with a `body: class_body` field, the **same `class_body` anchor** the #2550 machinery uses — now joins the instance model:

- Synthesized `Class` node `E$N` (structure query + `javaClassConfig`, with the skip-guard extended so the extractor's name fallback can't fabricate a class from the constant's own identifier).
- Scope boundary + body-anchored `@declaration.class`, so the body's methods are **owned** and re-keyed (`Method:...:E$1.hook#0`).
- A body-anchored `@reference.inherits` naming the **host enum** (javac semantics: `E$N extends E`), so `mroFor(E$N) ∋ E` and bare calls from the body to enum helpers pass the ownership gate's MRO arm.
- The #2550 same-file bare-call leak is closed for this shape (an unrelated class's bare `hook()` emits zero edges; discrimination evidenced by the archived pre-fix probe from the #2549 review).

### 2. JLS 13.1 immediate-host naming

Anonymous names switch from topmost-host anchoring to the immediately-enclosing-type chain (`EnumWrap$Mode$1`, matching JLS 13.1 binary names), numbered per immediate host in source order **across both shapes** (javac's shared counter). Every existing fixture's immediate host is its top-level type, so existing names are unchanged — proven by the untouched #2550 test suite, not assumed.

### Review hardening (empirical lens probes)

Five uncovered edge shapes probed on the implemented head: mixed bodied/body-less constants, constructor-args + body, interface-hosted anons, anon-inside-constant-body numbering, and deep nesting (`Outer$Inner$1`) — four correct first-try, one HIGH found and fixed: **same-named methods across sibling constant bodies collapsed onto the first body's node**. The scope-side qualifier chained the synthesized def to `M3.M3$2` (a constant body's Class scope parents into the enum's Class scope, unlike method-hosted anons which parent into a Function scope), desyncing from the structure id and falling to the file-wide simple-name lookup. `qualify()` now exempts class-like defs whose name already carries the `$`-chain — a complete binary name by construction — while `$`-named *members* (real in JS/TS) still qualify normally. Discriminating regression test included.

### Identity window

Node identities move on unchanged files (`E.hook` → `E$1.hook`; nested-host anons re-key), so the full lockstep checklist ships in the first commit: `INCREMENTAL_SCHEMA_VERSION` 8→9 (with the reuse-gate doc entry), parse-cache `SCHEMA_BUMP` 18→19, the U-C5 pin test extended with the v8-stamp rejection, and the scope-capture java bench fingerprint rebaselined (`measure.mjs --check` passes all 14 languages).

Known pre-existing gap (proven not a regression by an A/B probe): `E.CONST.method()` receiver dispatch emits no CALLS edge for bodied *or* body-less constants — noted for future work.

## Test plan

- New fixtures/scenarios: constant bodies modeled + owned + EXTENDS; leak closed; intra-body MRO-arm call; mixed bodied/body-less numbering; nested-host `EnumWrap$Mode$1`; same-name sibling attribution (the review fix's discriminator).
- Full `java.test.ts` 231/231 (twice sequentially pre-fix at 230, once post-fix); TS 254, JS/Kotlin 289, Python+Kotlin 459 (shared-file spot sets); schema/scope/owner units 90; bench `--check` PASS.
- `detect_changes` gated every commit; final graph check clean at head.


azizur100389:
Reviewed current head. No code blockers found. The enum constant body model covers synthesized class ownership, host-enum inheritance, same-file leak prevention, same-name sibling attribution, nested anonymous naming, and the schema/cache invalidation window. CI Gate is green. I am comfortable approving the code; only Vercel authorization remains if that check is required for merge.

PR #2559, opened 2026-07-18
fix(core): ensure path prefix and traversal guards support root directories

SyedaAnshrahGillani:
## Summary

This PR fixes a bug where repository path comparison and sandbox traversal guards duplicate slashes when the target path ends with a separator (like the root directory `/` on Unix or drive root `C:\` on Windows).

## Motivation / context

When a repository or target directory is located at a drive root (e.g., `C:\` on Windows) or the root filesystem (e.g., `/` on Unix), suffixing `path.sep` (e.g. `root + path.sep`) duplicates the separator (forming `//` or `C:\\\`). This causes prefix matching and traversal guards (which check `startsWith`) to fail. This change handles the trailing separator dynamically.

## Areas touched

## Scope & constraints

**In scope**

- Fixed `checkCwdMatch` in `git-staleness.ts` to support root-level directories.
- Fixed `assertSafePath` in `validation.ts` to support root-level directories.
- Fixed `resolveContainedDest` in `upload-ingest.ts` to correctly handle root-level staging directory sandbox checks.
- Fixed database deletion guard check in `lbug-adapter.ts` for root-level parents.
- Fixed local backend `assertSafePath` in `local-backend.ts`.
- Fixed `UPLOAD_ROOT` deletion check and file search path traversal guard in `api.ts`.
- Fixed `isWithinRepo` in `nuxt-auto-imports.ts` for root-level repositories.
- Added a unit test in `server-validation.test.ts` verifying `assertSafePath` with root-level `/`.

- Non-prefix path validation logic.

## Implementation notes

- Normalizes the prefix before `startsWith` check by checking `endsWith(path.sep) ? root : root + path.sep` instead of blindly appending `path.sep`.

## Testing & verification

- [x] `cd gitnexus && npm test` (Verified through specific unit tests suite)

### Ran:
- `npx vitest run test/unit/staleness.test.ts test/unit/sibling-clone-drift.test.ts test/unit/server-validation.test.ts`
- `npx tsc --noEmit`

## Risk & rollout

Low risk. Changes only refine boundary comparison for root-level directory path checks.

## Checklist

- [x] If `AGENTS.md` / overlays changed: headers, scope block, and changelog updated per project conventions


azizur100389:
I would hold this until full CI runs. Right now I only see labeler checks plus a Vercel authorization failure, with no CI Gate. This PR touches multiple path traversal/deletion guard paths, so it needs the normal required checks before review/merge. Please also update the title to conventional format and add coverage for the affected guard surfaces, especially Windows drive-root behavior, not just the single / case in server-validation.test.ts.

koriyoshi2041 (changes requested):
The root-prefix fix is incomplete in the same matching path. `gitnexus/src/core/augmentation/engine.ts` still checks `normalizedRepo + sep` and `normalizedCwd + sep` directly. With a registered repo at `/`, `findRepoForCwd("/src")` compares against `"//"` and returns no match, so hook augmentation silently disappears; the inverse root-CWD case has the same problem. Please route these comparisons through the same separator-aware prefix logic and add a focused regression for both directions. The current root test only exercises `server/validation.ts` and does not cover this remaining failure (or the requested Windows drive-root cases).

koriyoshi2041 (approved):
Re-reviewed at d8b7ff1. The augmentation matcher now handles both root-prefix directions, the focused Unix/Windows regressions are present, and the repository-owned CI matrix is green. The remaining Vercel failure is fork authorization rather than a code failure.

magyargergo:
@gitnexus review

magyargergo:
@SyedaAnshrahGillani Thank you for your contribution :)

PR #2560, opened 2026-07-18
fix(lbug): bound the LadybugDB buffer pool instead of the native 80%-of-RAM default

### LINKED ISSUE #2557: detect_changes via MCP balloons to ~19.5GB (80% of RAM) and OOM-kills the host session — bufferManagerSize=0 leaves LadybugDB buffer pool unbounded
## Summary

The `gitnexus mcp` server process ballooned from ~5 GB to ~19.5 GB RSS in about 60 seconds while handling a single `detect_changes` MCP tool call, triggering the kernel OOM killer and taking down the whole terminal session (24 GB RAM VM). The ceiling it climbed to matches LadybugDB's buffer-manager default of 80% of system RAM, which GitNexus never constrains.

## Environment

```text
OS: Linux x64 (Debian-based VM, 24 GB RAM + 8 GB swap)
Node: v24.x
GitNexus: 1.6.9 (npx cache, installed 2026-07-04)
@ladybugdb/core: 0.18.0
Repo: TypeScript/Next.js app — 10,249 symbols, 24,666 relationships, index (.gitnexus/lbug) 105 MB
```

Not #2233 — we are already on 1.6.9 / `@ladybugdb/core` 0.18.0, and this is memory growth to OOM, not a native SIGBUS crash.

## Timeline (from a 15-second memory sampler + systemd journal)

```text
16:39:39  MCP tool call: detect_changes (working tree had a large dirty diff:
          pnpm-lock.yaml + many generated JSON report files)
16:39:54  gitnexus mcp process (comm shows as "MainThread") at 5.1 GB RSS
16:40:09  8.9 GB
16:40:24  13.6 GB
16:40:40  17.3 GB
16:40:57  19.5 GB, swap exhausted
16:40:59  kernel OOM killer fires; systemd: scope Failed with result 'oom-kill',
          19.8G memory peak
```

Growth rate was roughly 5 GB per 15 s, i.e. the process consumed essentially all machine memory in one minute. This happened repeatedly across the day (9 OOM kills) — each coinciding with gitnexus analyze/detect_changes activity, and was initially misdiagnosed as a Next.js/turbopack build issue because the process `comm` is `MainThread`.

## Suspected cause

`createLbugDatabase()` in `dist/core/lbug/lbug-config.js` (source: `gitnexus/src/core/lbug/lbug-config.ts:543`) passes `bufferManagerSize = 0`:

```js
return new lbugModule.Database(databasePath, 0, /* bufferManagerSize */ ...)
```

`0` means "native default", and the LadybugDB/Kuzu-family default buffer pool is **80% of physical RAM** — on our 24 GB VM that is 19.2 GB, which matches the observed 19.5 GB plateau exactly. So under a heavy `detect_changes` workload (large dirty tree → FTS delete/update + incremental re-index) the buffer manager is *allowed* to grow until the machine dies, even though the on-disk graph is only ~105 MB.

`GITNEXUS_LBUG_MAX_DB_SIZE` does not help — it caps the mmap'd DB file size, not the buffer pool.

This is especially painful for the MCP server: it is a long-lived process spawned by the coding assistant, so users cannot wrap individual operations with cgroup limits the way they can for CLI invocations.

## Suggested fix

Any of:

1. Expose a `GITNEXUS_LBUG_BUFFER_POOL_SIZE` env var (mirroring `GITNEXUS_LBUG_MAX_DB_SIZE`) and pass it as `bufferManagerSize`.
2. Pick a sane fixed default (e.g. 1–2 GiB, configurable) instead of the native 80%-of-RAM default — the GitNexus self-index uses < 50 MB, so 80% of RAM is far beyond any realistic need.
3. At minimum, document the behaviour and the workaround.

## Workaround we applied

Wrapped the MCP server in a systemd scope in the assistant's MCP config:

```json
{
  "command": "systemd-run",
  "args": ["--user", "--scope", "--collect", "-q", "-p", "MemoryMax=6G", "-p", "MemorySwapMax=2G", "npx", "-y", "gitnexus@latest", "mcp"]
}
```

so a future balloon kills only the gitnexus process rather than the whole session. Happy to provide the full sampler logs or test a patched build.

magyargergo:
Closes #2557.

## Problem

`createLbugDatabase` passed `bufferManagerSize = 0`, which the LadybugDB native runtime sizes at **80% of physical RAM**. A long-lived `gitnexus mcp` process (or a large incremental `analyze`) could balloon to that ceiling — 19.5 GiB observed against a 105 MiB on-disk index — and OOM-kill the whole host session. `GITNEXUS_LBUG_MAX_DB_SIZE` caps only the mmap'd DB file, not the pool.

## Fix

All three paths suggested in the issue, at the single construction chokepoint every surface (CLI analyze, MCP server, worker pool, group bridges) routes through:

- **Bounded default**: `min(2 GiB, max(64 MiB, floor(0.8 × os.totalmem())))`, resolved at call time. The pool is a page cache with eviction, so oversized working sets get slower instead of killing the machine; the `min` keeps sub-2.5-GiB machines at the native-equivalent bound.
- **`GITNEXUS_LBUG_BUFFER_POOL_SIZE`** (bytes): positive values override; `0` deliberately restores the native unbounded default; invalid values warn and fall back (mirrors the `GITNEXUS_WAL_CHECKPOINT_THRESHOLD` pattern).
- **Docs**: both README env tables gain the new variable plus the previously code-comment-only `GITNEXUS_LBUG_MAX_DB_SIZE`, with the mmap-vs-memory distinction spelled out.

The short-lived `:memory:` FTS probe in `native-check.ts` is deliberately left unbounded (pool allocation is lazy; the probe lives milliseconds). A CLI flag (`--lbug-buffer-pool-size`) is deferred until someone needs it.

## Testing

- 9 new scenarios in `lbug-config-wal.test.ts` (default formula across machine sizes incl. the 64 MiB floor, env override, `0` passthrough, fractional flooring, invalid → warn + fallback, empty = unset); proven discriminating — 7 fail against the pre-fix tree.
- Empirical probe: the real `@ladybugdb/core` 0.18.0 native module constructs and queries cleanly with both 64 MiB and 2 GiB pools.
- Full unit suite: 492 files, 10,063 passed, 0 failed; `tsc --noEmit` clean.


azizur100389:
Reviewed current head. No findings from me. The buffer pool is now bounded at the shared createLbugDatabase construction point, the default/override/invalid-env behavior is covered, docs distinguish buffer-pool memory from DB mmap size, and CI/Vercel are green. Comfortable approving.

PR #2566, opened 2026-07-18
feat(skills): GitNexus Engineering Tool Kits

magyargergo:
## Summary

Supersedes #2431 with all findings from the [structured code review](https://github.com/abhigyanpatwari/GitNexus/pull/2431#issuecomment-5010276719) resolved, then merges the current `main` without rewriting the original branch history.

- makes plan/work evidence provenance and analyzer freshness fail closed across dirty trees, dependency changes, symlink retargets, and every lexical package boundary
- removes duplicate nested skills and strengthens byte-sync, steering, and distribution guards
- pins and authenticates the remote review workflow, checks out the exact trusted PR head, and validates a bounded name-status manifest before review
- requires complete per-skill promotion evidence, immutable task/oracle assets, named models, expected skill invocations, and transcript integrity
- isolates the proposer and benchmark runners, bounds descendant process trees, and classifies infrastructure failures separately
- applies promoted overlays with descriptor-relative identity checks, durable staging, atomic exchange, conservative rollback, and recovery records
- keeps the quick benchmark path free of opt-in expensive tasks and aligns eval CI with the documented `uv` workflow

## Review findings

All 18 actionable findings from #2431 are covered: dirty-state drift; duplicate skill definitions; invalid action pin; missing steering CI; irrelevant paid arms; partial promotion; missing transcript/work evidence; expensive defaults; stale inter-step/final graphs; wrong manual-review checkout; proposer host escape; leaked timeout descendants; planning mutation drift; non-`uv` eval CI; and unnamed models.

The implementation also closes adversarial edge cases found while validating those fixes, including staged/untracked gitlinks, rename/copy manifest ambiguity, dependency symlink retargeting, interrupted process ownership, and promotion-time filesystem races.

## Validation

- `cd gitnexus && npm run build` — pass
- `cd gitnexus && npx tsc --noEmit` — pass
- changed GitNexus suite, single worker — **14 files / 310 tests passed**
- `cd eval && uv run --locked --extra dev pytest -q` — **283 passed, 10 skipped**
- focused workflow/provenance checks — pass
- `actionlint` and `zizmor` — pass
- staged formatting, ESLint, and pre-commit typecheck — pass
- two independent adversarial follow-up reviews — no remaining findings
- GitNexus `detect_changes` — CRITICAL breadth, expected for 140 changed files and the CLI/eval/workflow surfaces; affected paths are covered by the gates above

A default-parallel `npm test` run cannot complete reliably in this container: native parse workers enter the repository's already-documented `Replacement worker did not report ready` / `Napi::Error` startup storm. The same changed tests pass 310/310 under `--maxWorkers=1 --no-file-parallelism`; hosted CI remains the authoritative full-suite gate.

## Relationship to #2431

This branch contains #2431's complete commit history, one reviewed hardening commit, and a clean merge of current `main`. No force-push or history rewrite was used.


eval/workflow_bench/evolve.py
```diff
@@ -0,0 +1,1021 @@
...
+            raise SandboxError(f"results artifact parent must be a real directory: {current}")
+        if transcript and stat.S_IMODE(metadata.st_mode) & 0o077:
+            raise SandboxError(f"transcript artifact parent must be owner-only: {current}")
+    return root / Path(*relative.parts)
+
+
+def _transcript_artifact_metadata(metadata: Any) -> tuple[str, str, int]:
+    """Validate transcript metadata without touching any host path."""
+
+    if not isinstance(metadata, dict) or set(metadata) != {"path", "sha256", "bytes"}:
```
magyargergo: **[P1 · reproduced] Evolution pipeline rejects its own runner's transcript metadata → seeded / multi-generation runs abort.**

This validator requires the metadata key set to be *exactly* `{path, sha256, bytes}`, but the runner's producer `_persist_parent_event_stream` emits a 4th key `"source": "parent-captured-stream-json"` (`runner_sessions.py:197-202`), and `sum_sessions` forwards each dict **unchanged** into `transcript_artifacts` (`runner_sessions.py:510-512`). So any real results row selected as proposer evidence — a `--seed-results` run (the documented usage) or generation ≥2 — hits `_preflight_transcript_artifacts` here and raises `SandboxError("transcript artifact metadata must contain only path, sha256, and bytes")` **before any proposal or promotion runs**.

**Reproduced** end-to-end (`uv run --locked`): real `sum_sessions` output → JSON round-trip (as results.jsonl seeding does) → this preflight → `SandboxError`; observed artifact keys `['bytes','path','sha256','source']`. Existing tests miss it because they build 3-field fixtures instead of round-tripping real producer output.

**Fix:** align producer and consumer on one schema — accept/validate (or ignore) `source` here, or strip it before it enters `transcript_artifacts`. Add an integration test that round-trips real `_persist_parent_event_stream` / `sum_sessions` output through `proposer_evidence_entries` across two generations.

gitnexus/src/cli/status.ts
```diff
@@ -49,16 +73,59 @@ export const statusCommand = async () => {
...
+          commit: activeMeta.lastCommit,
+          runnerIdentity: activeMeta.runnerIdentity ?? null,
+          runnerIdentityStatus: runnerIdentityIsCurrent ? 'current' : 'stale-or-unknown',
+          incompleteReasons,
+        },
+        current: {
+          commit: currentCommit,
+          runnerIdentity: currentRunnerIdentity,
+        },
+        status: isUpToDate ? 'up-to-date' : 'stale',
```
magyargergo, line 114: **[P2 · code-read] `status --json` reports a dirty working tree as `up-to-date`.**

`isUpToDate` (lines 82-85) tests only `currentCommit === activeMeta.lastCommit`, runner identity, and completeness — there is **no working-tree dirtiness check**. But `analyze` deliberately does *not* trust `HEAD === lastCommit` alone: it runs `git status --porcelain` (excluding self-written paths) and re-indexes a dirty tree (`run-analyze.ts:982-994`). So the two freshness contracts disagree — edit a tracked source file without committing and this new machine-readable API returns `"up-to-date"` while the graph omits your changes. A graph-backed agent gating on this JSON skips re-analysis and reasons on a stale graph.

**Fix:** include analyzer-visible dirtiness in `statusCommand` (share the `run-analyze` porcelain/hash logic; prefer comparing analyzer-visible file hashes vs `activeMeta.fileHashes` to distinguish new edits from an already-indexed dirty tree). Add staged / unstaged / untracked dirty-tree tests for both human and JSON output. *(Codex ce-review + coordinator.)*

gitnexus-claude-plugin/skills/gitnexus-plan/mcp.json
```diff
@@ -0,0 +1,8 @@
+{
+  "mcpServers": {
+    "gitnexus": {
+      "command": "npx",
+      "args": ["-y", "gitnexus@latest", "mcp"]
```
magyargergo: **[P2 · non-blocking · code-read] Plugin MCP configs launch an unpinned `gitnexus@latest`.**

This is an executable MCP definition (not docs): starting any of the four plugin skills runs whatever package currently owns `gitnexus@latest`, independent of the reviewed plugin version — non-reproducible, and a supply-chain execution boundary. The persisted setup path already pins the installed version for exactly this reason (`gitnexus/src/cli/setup.ts`). Same string in `gitnexus-{work,review,lfg}/mcp.json:5`.

**Fix:** pin the current release (`gitnexus@1.6.9`), generated from `gitnexus/package.json`, and bump it through the release/version tooling; keep the four copies synchronized via the existing mirror test. *(Most-corroborated non-blocking item: both Codex legs — swarm + ce — plus Claude facts/risk/security lanes.)*

.github/workflows/gitnexus-review-agent.yml
```diff
@@ -0,0 +1,2184 @@
...
+          claude_args: |
+            --model claude-sonnet-4-5-20250929
+            --add-dir "${{ runner.temp }}/gitnexus-review-pr-target"
+            --setting-sources user
+            --disable-slash-commands
+            --strict-mcp-config
+            --mcp-config "${{ runner.temp }}/gitnexus-review-mcp.json"
+            --tools "Read,Glob,Grep"
+            --allowedTools "Read(./**),Read(${{ runner.temp }}/gitnexus-review-pr-target/**),mcp__gitnexus__list_repos,mcp__gitnexus__query,mcp__gitnexus__context,mcp__gitnexus__check,mcp__gitnexus__impact,mcp__gitnexus__explain,mcp__gitnexus__pdg_query,mcp__gitnexus__route_map,mcp__gitnexus__tool_map,mcp__gitnexus__shape_check,mcp__gitnexus__api_impact,mcp__gitnexus__trace"
+            --disallowedTools "Bash,Write,Edit,MultiEdit,NotebookEdit,WebFetch,WebSearch,Skill,Task,Agent,Read(//proc/**),Read(//sys/**),Read(//dev/**),Read(/${{ github.workspace }}/**),mcp__github,mcp__gitnexus__detect_changes,mcp__gitnexus__rename,mcp__gitnexus__cypher,mcp__gitnexus__group_list,mcp__gitnexus__group_sync"
```
magyargergo: **[P2 · non-blocking · not exploitable · code-read] Double-slash deny globs in `--disallowedTools` may silently no-op.**

`${{ github.workspace }}` already expands to an absolute path, so `Read(/${{ github.workspace }}/**)` becomes `Read(//…)`; `Read(//proc/**)`, `Read(//sys/**)`, `Read(//dev/**)` are double-slash too. Path matchers that normalize `//`→`/` won't match these, silently disabling the deny layer.

**Impact is bounded to zero** at the current config: the `--allowedTools` allowlist (line 1186) grants only `Read(./**)`, the `pr-target` dir, and read-only gitnexus MCP tools — never `/proc`, `/sys`, `/dev`, or the workspace abs path — and `--permission-mode dontAsk` denies un-allowed reads. Pure defense-in-depth hygiene. **Fix:** drop the leading `/` (`/proc /sys /dev`, and `Read(${{ github.workspace }}/**)`). *(Two Claude lanes + Codex ce-security.)*

magyargergo (commented):
## Tri-review digest — PR #2566 (feat(skills): harden engineering skill family and evolution pipeline)

*Two-engine automated review.* **Claude** (Engine A) ran a production-readiness **swarm** (facts, hygiene, risk, security, test/CI) plus a **ce** persona pass (correctness, security, adversarial, reliability, project-standards, maintainability). **Codex** (Engine B, `gpt-5.6-sol`/xhigh) ran all three legs — **swarm-Solo** (B1 ✓), **ce-code-review** (B2 ✓), and **adversarial** (B3 ✓). So cross-engine corroboration is strong: both engines ran both a broad swarm pass and an adversarial pass. This is a genuine two-engine result, not five independent ones. All file:line claims below were re-opened in the PR-head worktree and confirmed before posting.

This is a well-hardened change and the reviews **largely validated it** — the security architecture in particular survived every attack both engines constructed. There is **one P1** worth blocking on, plus a small set of non-blocking P2/P3 follow-ups.

**Review bar for this PR:** high — it adds a secret-bearing CI review agent, an offline skill-evolution pipeline that promotes changes to shipped skills, and sandbox/containment code; correctness of the promotion gate and the trust boundaries is load-bearing.
**Problem being solved:** supersedes the now-closed #2431; hardens the GitNexus engineering skill family (plan/work/review/lfg), closes the skill-evolution loop (propose→benchmark→gate), and adds a comment-triggered `@gitnexus review` GitHub agent — folding in the #2431 review findings.
**Repository history considered:** predecessor **#2431 is CLOSED (not merged)**, superseded by this branch (which carries its full history + hardening commits); no linked closing issues; the rename `gitnexus-pr-review → gitnexus-review` and its uninstall migration are the main compatibility surface.
**Understanding of the change:** 140 files, +50,612/-842 — dominated by 4×-mirrored skill markdown + the new `eval/workflow_bench` Python pipeline, a new 2,184-line review-agent workflow, new pinned CI runtimes, and CLI/core changes (`status --json`, `analyzer-identity.ts`, uninstall seeding).
**Merge state:** `checks failing` — **sole cause is the Vercel "Authorization required" infra-auth check, not a code defect**; all code/test/security checks are green at HEAD (CodeQL passed; the GH-Advanced-Security "20 problems" comment is stale, from intermediate commit `2c078c90`).
**Branch hygiene:** `merge-from-main commit present but harmless and merge-safe` — 20 of 66 commits are clean main→branch syncs; conflict-free; single cohesive feature; rebase is cosmetic, not required.
**Back-and-forth avoided by verifying:** the P1 was **reproduced end-to-end** rather than asserted; all four inline anchors were re-opened in the worktree and confirmed on added diff lines; the prior-context security suspicions were each traced to a specific guarding line and refuted rather than re-filed.
**Open questions:** (1) the PR body's "two independent adversarial follow-up reviews" claim has no traceable artifact — can the author link them? (2) does the benchmark review arm re-index after task `setup` (bearing on the severity of `runner.py:1039`)?

---

### Headline (P1): the evolution loop rejects its own runner's transcript metadata

The benchmark's evidence validator and its producer disagree on the transcript-artifact schema, which breaks the documented multi-generation / `--seed-results` path before any proposal or promotion can run.

- Producer `_persist_parent_event_stream` returns a **4-key** dict `{"path","sha256","bytes","source":"parent-captured-stream-json"}` — `eval/workflow_bench/runner_sessions.py:197-202`.
- `sum_sessions` forwards each dict **unchanged** into `total["transcript_artifacts"]` — `runner_sessions.py:510-512`.
- Validator `_transcript_artifact_metadata` requires `set(metadata) == {"path","sha256","bytes"}` **exactly**, else raises `SandboxError` — `evolve.py:323`. It is reached via `proposer_evidence_entries → _preflight_transcript_artifacts` on every selected evidence row.

**Trigger:** any real results row that carries a runner-produced `transcript_artifact` is later selected as proposer evidence — i.e. a `--seed-results` run, or generation ≥2 feeding the prior generation's results forward. Evidence preflight raises `SandboxError("transcript artifact metadata must contain only path, sha256, and bytes")` and aborts. Codex B3 found it; the coordinator **reproduced** it end-to-end (`uv run --locked`: built the producer's exact 4-key artifact → real `sum_sessions` → JSON round-trip → `_preflight_transcript_artifacts` → `SandboxError`). Existing tests miss it because they construct 3-field fixtures instead of round-tripping real `_persist_parent_event_stream`/`sum_sessions` output.

**Fix:** align producer and consumer on one schema — have the validator accept (and validate/ignore) the runner's `source` field, or strip `source` before it enters `transcript_artifacts`. Add an integration test that round-trips real runner output through `sum_sessions → proposer_evidence_entries` across two generations.

---

### What's solid (validated by multiple lanes / both engines)

Both engines tried hard to break the security model and could not. Independently confirmed **working as designed**:

- **Fork-PR RCE / confused-deputy is closed.** Two-job split: the secret-bearing `analyze` job is `contents:read`+`pull-requests:read` only; the `publish` job holds `issues:write` but no model secret. The PR head is checked out at an exact SHA as *passive data*; every step that could execute hostile bytes (native parser, MCP reader, runtime installs) runs inside Bubblewrap (`--unshare-user/pid/net`, `--die-with-parent`, `/usr/bin/env -i`) and completes *before* the OAuth token is introduced. No in-tree `run.cjs`/npm-script executes under the secret. (Claude risk+security+adversarial+project-standards; Codex B2 "prior leads verified as addressed".)
- **Promotion gate is fail-CLOSED, not fail-open.** `evidence-unverified` is excluded; any excluded run in either arm → `insufficient_evidence`; a `skill-not-invoked` run counts as valid-but-unresolved and fails the 100% quality floor → `keep_incumbent`. No un-invoked-skill candidate can be promoted. (Claude correctness R1/R2, adversarial R1/R2; Codex B2.)
- **Rename migration is data-safe.** `RENAMED_SKILL_DIRS`/`LEGACY_SKILL_DIR_NAMES` seed the uninstall removal set, so a pre-rename `gitnexus-pr-review` dir is removed by name even when absent from bundled source; covered by `uninstall.test.ts:161-176`. Non-`gitnexus`-named dirs cannot be deleted. (Claude + Codex.)
- **Proposer/overlay sandbox is sound.** Env built from scratch (never copies `os.environ`), `O_NOFOLLOW`/`resolve(strict)` on every mount, overlay restricted to `.md` under `gitnexus-{plan,work}`, PID-namespace bounding, secret redaction, TOCTOU-safe freeze→validate→`renameat2(RENAME_EXCHANGE)` apply. No escape surface found. (Claude security R5, adversarial R4/R5; Codex B2.)
- **Drift guard + rename cleanup have real teeth.** `skill-sync.yml` triggers on all four mirror trees including `.md` (fixing the old `paths-ignore:['**.md']` gap); `shipped-skills-sync.test.ts` enforces byte-identity + resurrection guard; nested `gitnexus-pr-review` duplicate tree deleted everywhere.

Codex's own swarm leg (B1) **independently re-confirmed** these: Lane 5 cleared the fork-PR RCE / confused-deputy class (empty top-level permissions, read-only analyze token, PR content quarantined before the OAuth secret, no in-tree `run.cjs`/`gh api`/write-token regression), and Lanes 1/3 cleared the promotion-gate, uninstall-seeding, and proposal-placement fixes. So **both engines, across both their swarm and adversarial passes, cleared the prior-context security suspicions** — the strongest form of corroboration available here.

**Suspicions the reviews refuted (do not re-file):** fork-PR RCE, promotion-gate fail-open, uninstall orphan, proposer sandbox escape, overlay freeze→apply TOCTOU, `mcp__gitnexus__*` data-exfil, Cypher-injection in the prescan, cost double-count. All traced to a specific guarding line.

---

### Inline findings (meet trigger + evidence + tag; cross-engine or coordinator-verified)

| # | Location | Sev | Problem | Tag | Lanes / engines |
|---|----------|-----|---------|-----|-----------------|
| 1 | `eval/workflow_bench/evolve.py:323` | **P1** | Transcript-metadata schema mismatch aborts seeded/gen≥2 evolution runs (see headline) | **[reproduced]** | Codex B3 + coordinator (Claude lanes missed it) |
| 2 | `gitnexus/src/cli/status.ts:104` | **P2** | `status --json` returns `"up-to-date"` for a dirty working tree — no porcelain/hash check (`isUpToDate` at :82-85 tests commit+runner-identity+incompleteReasons only), disagreeing with `analyze`, which re-indexes a dirty tree (`run-analyze.ts:982-994`). A graph-backed agent gating on this JSON skips re-analysis on a stale graph. | [code-read] | Codex B2 #3 + coordinator |
| 3 | `gitnexus-claude-plugin/skills/gitnexus-plan/mcp.json:5` | **P2 (non-blocking, most-corroborated item)** | All four plugin skills' MCP configs run `npx -y gitnexus@latest` — executable, not docs. Makes released workflows non-reproducible and adds a supply-chain execution boundary; the persisted setup path already pins the version (`setup.ts`). Same string in `gitnexus-{work,review,lfg}/mcp.json:5`. Fix: pin `gitnexus@1.6.9` and update it through the release/version tooling that already pins the installed setup path. | [code-read] | **Both Codex legs (B1 Finding 4 + B2 #2)** + Claude facts/risk/security lanes |
| 4 | `.github/workflows/gitnexus-review-agent.yml:1187` | **P2 (non-blocking, NOT exploitable)** | Double-slash deny globs `Read(//proc/**),Read(//sys/**),Read(//dev/**),Read(/${{ github.workspace }}/**)` — `github.workspace` is already absolute, so these become `//…` and may not match normalized paths, silently no-op'ing the deny layer. **Bounded to zero impact:** the `--allowedTools` allowlist (1186) grants only `Read(./**)` + the pr-target dir + read-only gitnexus MCP tools — it never grants /proc, /sys, /dev, or the workspace abs path, and `--permission-mode dontAsk` denies un-allowed reads. Pure defense-in-depth hygiene. Fix: drop the leading `/` (single-slash `/proc /sys /dev`). | [code-read] | Claude risk F1 + Claude security F1 + Codex CE-security residual #1 |

Findings 1–4 are **inline-anchorable** (all four cited lines are added/new in the diff). Suggested severities: block on #1; #2 and #3 are pre-ship follow-ups; #4 is optional hardening.

---

### Body-only P2 (verified ordering, consequence not fully traced)

- **`eval/workflow_bench/runner.py:1039` — benchmark task `setup` runs after graph materialization + candidate overlay.** `graph_snapshot.materialize` (`runner.py:987`) builds the cached graph *before* `if task.get("setup")` (`runner.py:1039`) applies the diff-under-review as local changes; the candidate overlay is applied at `1024-1030`, also before setup. **Ordering confirmed** (Codex B2 #1 + coordinator). Plausible consequence: the review arm's cached graph omits the changed symbols it should review, and setup runs with the candidate overlay already present so incumbent/candidate arms can diverge on setup-produced state. Whether the arm re-indexes post-setup was **not fully traced**, so this is carried as a body P2, not a hard inline assertion. Fix: prepare + content-bind the task workspace before graph construction and before any candidate overlay; build the graph from setup-applied state.

---

### Codex B1 (swarm-Solo) integrity / defense items — body-only, non-blocking P2

Single-engine Codex findings (its now-complete swarm leg), coordinator-verified by code-read where noted. None is a new inline-worthy item, but two are in the same evidence-integrity family as the P1 and worth folding into the fix:

- **B1 F1 — cost silently defaults to `$0` (P2, evidence-integrity family).** `runner_sessions.py:425` records `cost_usd = data.get("total_cost_usd", 0.0)`, but the session validator `well_formed` (`runner_sessions.py:400`) checks only the nested `usage` fields — **not** `total_cost_usd`. Since `cost_usd` is the default promotion metric (lower = better), an otherwise-successful session that is missing `total_cost_usd` scores as **free** and can be wrongly promoted. Same failure shape as CV1 (producer/validator disagreement in the evidence pipeline). Fix alongside the P1: require + validate `total_cost_usd` (finite, ≥0, measured-zero distinguished from absent) on successful sessions, and reject promotion if either arm lacks the selected metric. [code-read, coordinator-confirmed]
- **B1 F2 — drift guard is symlink-defeatable (P2).** `shipped-skills-sync.test.ts:21-38`: `listFilesRecursive` (`readdirSync`, `entry.isDirectory()`) and `snapshotDir` (`fs.readFileSync`) both follow symlinks and never `lstat`/reject them, so a mirror file symlinked to the canonical tree passes the byte-compare — and would ship a symlink. Fix: `lstat` every entry, reject symlinks/non-regular files (and symlinked parents); add negative tests that symlink each mirror's `SKILL.md`/scripts. [code-read]
- **B1 F3 — real containment canary doesn't exercise the proposer's built-in file tools (P2).** `test_proposer_sandbox.py:703-744` asserts Bash + MCP containment, but production `run_proposer` (`evolve.py:465-493`) also enables `Read/Grep/Glob/Edit/Write` against the `/evidence` (read) and output (write) mounts. So the canary proves *process/MCP* isolation but not that Claude's built-in file tools are confined to `/evidence`+output — a distinction that matters because file tools use Claude's permission system while the OS sandbox mainly constrains Bash/subprocesses. Corroborates CE-adversarial TG4 + CE-reliability. Fix: add a real-Claude canary using `run_proposer`'s exact allowlist that reads a sentinel from `/evidence`, writes only to output, and asserts denial elsewhere. [code-read]
- **B1 F5 — review-agent DoD gate unmet pre-merge (P2, corroborates facts + risk lanes).** `gitnexus-review-agent.yml:4-10` ("Off by default"; needs same-repo + fork validation) vs `DoD.md:137` ("dry-run or triggered run") — but **no run can exist pre-merge**: `issue_comment` executes the default-branch workflow, and an unregistered workflow can't be dispatched from this PR (the "dead until merge" property the facts/risk lanes independently noted). Not a code defect; a rollout-sequencing gap. Recommendation: keep the review agent disabled until a two-phase rollout validates same-repo + fork execution *post-merge* — either stage activation explicitly, or revise the DoD wording to "before enabling."

### Lower-priority / single-lane items (not posted inline)

Each is a single Claude lane, [code-read], not cross-engine or reproduced — noted for the author, not blocking:

- **`evolve.py:940` (P3, correctness, conf 70):** an overlay that *adds* a new `.md` under `gitnexus-{plan,work}` (absent at HEAD) passes `candidate_overlay_files` (shape-only check) but `committed_destination_base_digests` raises `ValueError` uncaught → the whole evolve driver dies mid-run instead of reporting `NOT PROMOTED`. Nothing incorrect is *applied* (fails before the benchmark); it's an ungraceful crash / lost run. Fix: make the overlay trust boundary modify-only, or catch and treat as `NOT PROMOTED`.
- **`evolution.py:48` (P3, correctness):** `num_turns` is a selectable promotion metric but is main-loop-only (excludes subagent turns) and, unlike `output_tokens`, emits no `metric_warning` — a subagent-heavy candidate looks artificially efficient. Fix: add `num_turns` to `MAIN_LOOP_ONLY_METRICS`.
- **`runner.py` sweep loop (P2, reliability, conf 50):** no circuit-breaker / consecutive-failure abort; a sustained API outage mid-sweep pays out every remaining `--timeout` window one session at a time (hours of wall-clock) with no distinct "systemic outage" signal. Partial rows are preserved (no data loss).
- **`promotion_apply.py:24` `MIRROR_SKILL_ROOTS` (residual, adversarial RR1):** the 2-mirror apply set is narrower than the 4-tree drift guard; safe **today** only because `CANDIDATE_SKILLS` excludes the cursor-mirrored `gitnexus-review`. If a cursor-mirrored skill is ever added to `CANDIDATE_SKILLS`, `apply_promoted_overlay` would rewrite 3 of 4 trees and skip cursor — the exact PR #2488 asymmetric-sync bug class, latent one edit away. Suggest deriving the mirror set (or an assertion) from the drift-guard catalog.
- **`skill-sync.yml:59` (P3, security, conf 40):** `npm install && npm run build` for `gitnexus-shared` (vs `npm ci` immediately below for `gitnexus`) — floats deps in a `contents:read`, secretless job. Replicated pre-existing pattern; suggest `npm ci` + committed lockfile. **Divergence to note:** the canary `npm ci --ignore-scripts=false` (`ci-tests.yml`) was flagged P3 by swarm-security but assessed "acceptable" by CE-security (integrity-pinned, no secret in that job) — the lanes disagree; treat as non-blocking hygiene either way.

### Maintainability debt (quality notes, NOT merge blockers)

- **`evidence-provenance.mjs` — 2084 lines, 6 byte-identical hand-copies** (same md5 across `.claude`/`gitnexus`/`gitnexus-claude-plugin` × plan/work), guarded only post-hoc by a CI byte-compare (drift caught, never prevented) — 21× larger than the repo's own `run.cjs↔resolve-invocation.ts` precedent for this discipline. Suggest one canonical source + a generate/copy step.
- **`gitnexus/src/core/analyzer-identity.ts` — new 2462-line single file**, ~70 internal functions behind a 10-symbol public surface, with clean internal seams (build-tree / dependency-runtime / cache) that would split into a directory barrel without touching any of its 4 call sites.
- Plus P2 structural notes: the eval "arm" taxonomy is defined piecemeal across 4 files; `runner.py:main()` is 433 lines; `task_assets.py` imports underscore-"private" helpers from `proposer_sandbox.py`; the atomic-write-then-freeze idiom is hand-duplicated across ~9 sites; the 621-line inline-bash step in `gitnexus-review-agent.yml` isn't extracted/shellcheckable.

---

### Coverage, CI, and things not deep-verified

- **CI at HEAD:** green except **Vercel** ("Authorization required" — infra auth, not a code defect; this is the sole source of the UNSTABLE merge state) and skipped e2e Chromium (no web changes). New jobs pass: eval locked-pytest, eval-containment (ubuntu + windows), skill drift guard. CodeQL **passed at HEAD**; the github-advanced-security "20 problems" comment was at intermediate commit `2c078c90`, now **stale**. The 2 `PinnedDependenciesID` mediums in changed files are **pre-existing** (single appended `ci-tests.yml` hunk), not introduced here.
- **Unverifiable PR-body claim:** "two independent adversarial follow-up reviews — no remaining findings" has **no traceable artifact** in the repo or GitHub review history (facts + project-standards lanes). Not asserting it's false — just noting it can't be confirmed from public state.
- **Scope:** 140 files, +50,612/-842. The ~50k-line diff is dominated by **4×-mirrored skill markdown and generated lockfiles**, which were **not** read line-by-line — the byte-identity drift-guard tests are the backstop there. Deep verification focused on the eval pipeline (`evolve.py`, `runner*.py`, `promotion_apply.py`, `proposer_sandbox.py`, `process_control.py`), the review-agent workflow trust boundaries, and the `status.ts`/`analyzer-identity.ts` CLI surface. Not independently re-run: the full vitest/pytest suites (worktree has symlinked deps; CI's 15,000-test suite is the authoritative gate) — CV1 is the one path executed directly.
- **Engine liveness:** all three Codex legs completed — swarm-Solo B1 ✓ (5 findings), ce B2 ✓ (3 P2s), adversarial B3 ✓ (the P1). B1's items are single-engine Codex, folded into the body above; the four inline findings stand unchanged.

---

### Final verdict: **not production-ready** (one blocking P1; everything else non-blocking)

The confirmed, reproduced P1 makes the evolution pipeline's own documented `--seed-results` / multi-generation path abort (`SandboxError`) before any proposal or promotion can run — the producer emits a `source` key the validator rejects — so that headline feature is non-functional as shipped, and the same producer/validator-disagreement shape also lets a cost-less session score as free (B1 F1). Everything else is solid: both engines' swarm **and** adversarial passes failed to break the security model — the fork-PR RCE class is closed, the promotion gate is fail-closed, the proposer sandbox has no escape surface, and the rename migration is data-safe. The unblocking action is small and local: align `_transcript_artifact_metadata` with the runner's transcript schema (accept/validate or strip `source`), validate `total_cost_usd`, and add one integration test that round-trips real runner output across two generations. The remaining P2/P3s (`status --json` freshness, `gitnexus@latest` pinning, the double-slash deny globs, the symlink-defeatable drift guard, the review-agent DoD/rollout gate) are pre-ship follow-ups, none merge-blocking on their own. Recommendation: fix the P1 (+ B1 F1) before relying on the benchmark/promotion path; keep the review agent disabled until its post-merge same-repo + fork validation runs.

---

*Automated two-engine digest (Claude swarm + ce; Codex swarm + ce + adversarial). Findings were re-checked against the PR-head worktree, but please verify each against current HEAD before acting — treat the inline anchors as starting points, not final line references.*

magyargergo:
## GitNexus review — reproduced out-of-band

This is the `@gitnexus review` agent's output for this PR. Because `issue_comment`/`workflow_dispatch` only ever run the **default-branch** copy of `.github/workflows/gitnexus-review-agent.yml`, the agent can't self-trigger on the branch that introduces it. So this was produced by running the agent's *exact* session out-of-band — same `gitnexus-review` skill, prompt, read-only tool allowlist + GitNexus MCP, and model `claude-sonnet-4-5-20250929` — against a freshly rebuilt index at the current head. (19 turns, ~$0.91.)

---

# Review: PR #2566

## Findings

**No critical defects found.** The PR is well-structured with comprehensive test coverage. Several observations below are for maintainability and documentation clarity, not blocking issues.

**Target:** PR #2566  
**Head:** 5c7eb052d95c36c5b565487fd3b2cc9b8a700b94  
**Merge-base:** 12600000e36aa890cd5f6daaa7f3293fd2190b18

**Scope:** Large-scale feature addition introducing a complete engineering workflow system.

**Changed symbols and major additions:**

1. **Four new skills** (complete implementations):
   - `.claude/skills/gitnexus-lfg/` — pipeline orchestrator (plan → gate → work → review)
   - `.claude/skills/gitnexus-plan/` — implementation planning with PDG-backed evidence provenance
   - `.claude/skills/gitnexus-work/` — plan executor with atomic commits
   - `.claude/skills/gitnexus-review/` — code review skill (replaces `gitnexus-pr-review`)

2. **Core infrastructure** (new TypeScript modules):
   - `gitnexus/src/core/analyzer-identity.ts` — analyzer source/build provenance tracking
   - `gitnexus/src/core/index-freshness.ts` — index staleness detection
   - `gitnexus/src/cli/standard-skills.ts` — shipped skill manifest

3. **Evaluation framework** (`eval/workflow_bench/`):
   - Complete benchmark harness for multi-agent workflow evaluation
   - 12 new Python test files covering sandbox isolation, graph sanitization, CE plugin runtime
   - Oracle-based verification with deterministic promotion gates
   - Sanitized graph construction from history-pruned snapshots

4. **Evidence provenance system** (shared between `gitnexus-plan` and `gitnexus-work`):
   - `scripts/evidence-provenance.mjs` (2084 lines) — cryptographically-signed evidence tracking
   - Schema v2 with NUL-framed UTF-8 canonicalization
   - Descriptor-anchored safe file reader/writer with TOCTOU guards
   - Git-admin backup for plan replacements (Deepen mode)

5. **Modified core behavior**:
   - `gitnexus/src/cli/analyze.ts` — added `runnerIdentityAtBootstrap` parameter
   - `gitnexus/src/cli/lazy-action.ts` — new `createAnalyzerLbugLazyAction` for identity capture
   - `gitnexus/src/cli/status.ts` — added `--json` output with analyzer provenance
   - `gitnexus/src/storage/repo-manager.ts` — added `AnalyzerRunnerIdentity` persistence
   - `gitnexus/src/mcp/resources.ts` — schema resource now includes analyzer identity

6. **Documentation & workflow changes**:
   - `.github/workflows/gitnexus-review-agent.yml` — new CI review workflow
   - `.github/workflows/skill-sync.yml` — skill sync workflow
   - Updated `CLAUDE.md`, `AGENTS.md`, `DoD.md`, `README.md`
   - Skill renames: `gitnexus-pr-review` → `gitnexus-review` (broader scope)

**Execution flows:** The new skills form a cohesive pipeline, but each can also operate independently. `gitnexus-lfg` orchestrates the others; `gitnexus-plan` generates implementation-ready plans; `gitnexus-work` executes them; `gitnexus-review` validates results.

**Test coverage:**
- **New tests added:** 17 TypeScript test files, 12 Python workflow bench tests, 4 oracle test files
- **Key new test coverage:**
  - `analyzer-identity.test.ts` — identity computation and persistence
  - `analyzer-identity-cli.test.ts` — CLI integration
  - `evidence-provenance-helper.test.ts` — provenance serialization
  - `review-agent-workflow.test.ts` — review workflow validation
  - `engineering-skills-contract.test.ts` — skill API stability
  - `shipped-skills-sync.test.ts` — manifest synchronization
  - Comprehensive workflow bench tests for sandbox isolation, graph sanitization, oracle assets

**Evidence provenance security model:**
- The 2084-line `evidence-provenance.mjs` script implements defense-in-depth for file operations:
  - Descriptor-anchored reads with O_NOFOLLOW
  - Before/after stat identity checks (TOCTOU guards)
  - Atomic no-replace moves via Python `renameat2(RENAME_NOREPLACE)`
  - Symlink traversal prevention in all parent components
  - Git-admin vault for displaced plans (Deepen mode)
  - Fail-closed on platforms lacking POSIX ownership APIs or `/proc/self/fd`
- The implementation is substantial and security-critical. Test coverage exists (`evidence-provenance-helper.test.ts`), but comprehensive adversarial testing (symlink races, concurrent modifications, partial writes, RENAME_NOREPLACE platform variations) would strengthen confidence.

**Analyzer identity tracking:**
- New `analyzer-identity.ts` computes a fingerprint of the analyzer source/build state
- Persisted in repo metadata and checked by freshness gates
- Enables planning skills to detect stale analyzer builds before trusting graph output
- Test coverage exists (`analyzer-identity.test.ts`, `analyzer-identity-cli.test.ts`)

**Workflow bench evaluation framework:**
- Introduces a complete multi-agent benchmark system with:
  - Sandbox isolation (read-only mounts, PID namespaces, managed processes)
  - Sanitized graph construction (history-pruned clones, benchmark marker scrubbing)
  - CE plugin snapshot verification
  - Oracle-based correctness checking with deterministic promotion
- Test coverage is comprehensive (12 Python test files)
- Framework design prevents benchmark leakage into target graphs via marker scrubbing

**Residual risk:**
- **Evidence provenance complexity:** The safe reader/writer is intricate (descriptor anchoring, identity checks, atomic moves, Git-admin backups). While test coverage exists, the attack surface is large. Consider:
  - Fuzz testing with concurrent file system mutations
  - Platform-specific RENAME_NOREPLACE behavior validation (Linux, macOS, BSD)
  - Symlink race scenario coverage (especially parent directory races)
- **Skill orchestration state:** `gitnexus-lfg` routes between plan/work/review but has no persistent execution state. A crashed session loses progress. The design is intentional (stateless orchestration), but users may expect resumability.
- **Plan versioning:** Schema v2 is introduced. Legacy schema v1 plans require re-anchoring. Migration path is documented but not automated.
- **Performance:** `evidence-provenance.mjs` performs extensive filesystem probing (stat checks, descriptor anchoring, Git object reads, directory traversals). Impact on large repositories or slow filesystems is untested in CI.

**Graph limitations:** The changes correctly document PDG limitations (intra-procedural control/data dependence, function-level cross-function taint, no closure/callback flow). These are not regressions — the PR adds features that leverage existing PDG capabilities.

### Verdict

**APPROVE**

This PR delivers a major feature addition with solid engineering discipline:
- Clear architectural separation (lfg orchestrates, plan produces, work executes, review validates)
- Comprehensive test coverage for new functionality
- Security-conscious file operations (descriptor anchoring, TOCTOU guards, fail-closed)
- Thorough documentation (skill READMEs, reference docs, plan templates)
- Backward compatibility maintained (skill renames, schema versioning)
- CI integration (review agent workflow, skill sync workflow)

**Strengths:**
- Evidence provenance system is production-grade (cryptographic canonicalization, atomic writes, Git-admin backups)
- Workflow bench provides deterministic evaluation infrastructure
- Analyzer identity tracking prevents stale-graph issues in planning workflows
- Extensive inline documentation and examples

**Recommendations for follow-up (not blocking):**
1. Adversarial testing for `evidence-provenance.mjs` (symlink races, concurrent mutations, platform variations)
2. Performance profiling on large repositories (evidence snapshot overhead, directory traversal bounds)
3. Resumable execution for `gitnexus-lfg` (optional; current stateless design is valid)
4. Schema v1 → v2 migration helper (currently manual re-anchoring)

**Deployment notes:**
- Skills are opt-in (user invokes `/gitnexus-lfg`, `/gitnexus-plan`, etc.)
- Analyzer identity tracking is passive (computed during `analyze`, reported in `status --json`)
- No breaking changes to existing MCP tools or CLI commands
- New `--json` flag on `gitnexus status` is additive

---

**Evidence:**
- 153 files changed across skills, core, eval, docs, workflows
- 17 new TypeScript tests, 12 new Python workflow bench tests
- Key modules verified via context query attempts (new files not in merge-base graph as expected)
- Diff spans skill implementations, evidence provenance, analyzer identity, workflow bench, CI workflows
- No deprecated APIs removed without migration path
- Test suite includes integration tests (`analyzer-identity-cli.test.ts`, `setup-skills.test.ts`) and unit tests for all new modules


---
Analyzed head: `7ad79703d7cb9905704231a7b1858e0f848aa33d`  
Merge-base: `12600000e36aa890cd5f6daaa7f3293fd2190b18`  
Base (origin/main): `12600000e36aa890cd5f6daaa7f3293fd2190b18`

PR #2571, opened 2026-07-19
feat(eval): run the skill-evolution loop online

magyargergo:
Makes the skill-evolution loop **online**: the existing offline propose → benchmark → gate pipeline (`eval/workflow_bench/evolve.py`, from #2431/#2566) now runs on a weekly schedule in CI, with the same staged-rollout pattern as the review agent.

## What it does

1. **Weekly (Sat 03:00 UTC) or on dispatch**, spins up the proven eval environment: pinned Claude canary runtime (2.1.214), bubblewrap containment with the AppArmor userns fix, pinned uv/Python, built GitNexus runtimes — the exact setup `eval / containment (ubuntu)` already validates on every CI run.
2. Runs `workflow_bench.evolve` (default: 1 generation, 3 runs/arm/task, `claude-sonnet-5`, expensive tasks excluded) and **always uploads the full results directory as an artifact**.
3. If the deterministic gate promotes an overlay, the applied diff is **bounded to the three skill trees** (`.claude/skills/`, `gitnexus/skills/`, `gitnexus-claude-plugin/skills/`) — any path outside that boundary fails the run. A bounded promotion becomes a branch + **human-reviewed PR** opened with the release App token (so CI runs on it), with the `promotion.json` decisions inlined.

The gate stays evidence *for* a PR, never a bypass: nothing merges without review.

## Activation checklist (scheduled lane ships OFF)

- [x] Add repository secret `GITNEXUS_BENCH_AUTH_TOKEN` — an **Anthropic API key**; benchmark sessions bill real usage (the Claude Code OAuth subscription token does not work here — the bench injects it as `ANTHROPIC_API_KEY`).
- [x] Validate once via `workflow_dispatch` (containment preflight, runtime inside the 355-min timeout, artifact upload, PR shape on promotion).
- [x] `gh variable set GITNEXUS_EVOLUTION_ENABLED --body true`

## Scope note

Today only `gitnexus-plan` and `gitnexus-work` are promotable (`CANDIDATE_SKILLS`); `gitnexus-review` is the protected evaluation arm. Making the review skill itself evolvable needs its own bench arm with independent ground truth (seeded-defect tasks) — follow-up work, orthogonal to this workflow.

### DIFF
--- a/.github/workflows/gitnexus-skill-evolution.yml
+++ b/.github/workflows/gitnexus-skill-evolution.yml
@@ -204,12 +204,21 @@ jobs:
             esac
           done <<< "${changed}"
           echo "promoted=true" >> "${GITHUB_OUTPUT}"
+          # The loop returns on the first promotion, so the highest-numbered
+          # gen-N/bench/promotion.json is the decision that actually fired.
+          # Emit only that one — never every generation's, or a rejected
+          # generation's decisions could surface in the PR body. The heredoc
+          # uses a per-run random delimiter so a summary value that ever
+          # contains the marker cannot close the block early and inject keys.
+          promotion_file="$(find "${OUT_ROOT}" -name promotion.json | sort -V | tail -1)"
+          delim="PROMOTION_EOF_$(openssl rand -hex 16)"
           {
-            echo 'summary<<PROMOTION_EOF'
-            find "${OUT_ROOT}" -name promotion.json -print0 |
-              xargs -0 -r cat | tail -c 8000
+            echo "summary<<${delim}"
+            if [[ -n "${promotion_file}" ]]; then
+              tail -c 8000 "${promotion_file}"
+            fi
             echo
-            echo 'PROMOTION_EOF'
+            echo "${delim}"
           } >> "${GITHUB_OUTPUT}"
 
       - name: Mint GitHub App token
@@ -229,7 +238,11 @@ jobs:
           RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: |
           set -euo pipefail
-          branch="evolution/skills-${GITHUB_RUN_ID}"
+          # Include the run attempt: GITHUB_RUN_ID is stable across re-runs, so
+          # a re-run after a push-succeeds/PR-create-fails partial failure needs
+          # a fresh branch to push (a non-force push to the existing branch
+          # would be rejected non-fast-forward and wedge the lane).
+          branch="evolution/skills-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
           git config user.name 'gitnexus-evolution[bot]'
           git config user.email 'gitnexus-evolution[bot]@users.noreply.github.com'
           git checkout -b "${branch}"
--- a/.github/workflows/gitnexus-skill-evolution.yml
+++ b/.github/workflows/gitnexus-skill-evolution.yml
@@ -131,12 +131,14 @@ jobs:
 
       - name: Build pinned shared runtime
         run: |
+          set -euo pipefail
           npm ci
           npm run build
         working-directory: gitnexus-shared
 
       - name: Install and build pinned GitNexus runtime
         run: |
+          set -euo pipefail
           npm ci
           npm run build
         working-directory: gitnexus
@@ -176,7 +178,7 @@ jobs:
 
       - name: Upload benchmark evidence
         if: always() && steps.loop.outputs.out_root != ''
-        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v6.0.0
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: gitnexus-evolution-${{ github.run_id }}-${{ github.run_attempt }}
           path: ${{ steps.loop.outputs.out_root }}
--- a/.github/workflows/gitnexus-skill-evolution.yml
+++ b/.github/workflows/gitnexus-skill-evolution.yml
@@ -8,15 +8,31 @@
 # [ ] Configure the repository secret GITNEXUS_BENCH_AUTH_TOKEN (an Anthropic
 #     API key — benchmark sessions bill real usage; the Claude Code OAuth
 #     subscription token does not work here).
+# [ ] Configure the RELEASE_APP_ID and RELEASE_APP_PRIVATE_KEY secrets (the
+#     App that opens the promotion PR). The Mint-App-Token step hard-fails
+#     without them once a promotion is detected. Verify the App installation
+#     is scoped to this repo with only Contents: RW + Pull requests: RW.
+# [ ] Create the protected Environment `gitnexus-evolution` with a
+#     deployment-branch rule restricting it to `main`, and ideally scope the
+#     three secrets above to that Environment. workflow_dispatch runs this
+#     workflow (and eval/workflow_bench/evolve.py) from the *dispatched ref*,
+#     so this server-side rule — not a code-side guard the branch could edit
+#     away — is what stops a non-main branch from running with the secrets.
 # [ ] Run workflow_dispatch once and confirm: containment preflight passes,
 #     the benchmark completes inside the job timeout, the results artifact
 #     uploads, and a promotion (if any) opens a well-formed PR.
 # [ ] Set the repository variable GITNEXUS_EVOLUTION_ENABLED=true.
-# Roll back by setting that variable to false; workflow_dispatch remains available.
+# Roll back by setting that variable to false. Note: workflow_dispatch always
+# runs the full benchmark loop regardless of GITNEXUS_EVOLUTION_ENABLED and
+# bills real API usage on GITNEXUS_BENCH_AUTH_TOKEN.
 name: GitNexus skill evolution
 
 on:
   schedule:
+    # Weekly is a deliberate cadence to catch model/harness drift promptly; a
+    # no-promotion week only costs one benchmark run (the gate keeps the
+    # incumbent unless quality improves). Dial back toward the README's ~90-day
+    # re-evaluation guidance if the recurring spend is not worth it.
     - cron: '0 3 * * 6' # weekly, Saturday 03:00 UTC
   workflow_dispatch:
     inputs:
@@ -62,7 +78,7 @@ jobs:
     # it — server-side enforcement a dispatched non-main ref cannot bypass by
     # editing its own workflow copy. See the activation checklist above.
     environment: gitnexus-evolution
-    timeout-minutes: 355
+    timeout-minutes: 355 # ceiling just under GitHub's 360-minute hard cap
     permissions:
       contents: read # The promotion PR uses a short-lived App token minted below.
     env:
--- a/.github/workflows/gitnexus-skill-evolution.yml
+++ b/.github/workflows/gitnexus-skill-evolution.yml
@@ -1,0 +1,240 @@
+# GitNexus skill evolution: runs the offline propose → benchmark → gate loop
+# (eval/workflow_bench/evolve.py) on a schedule and, when the deterministic
+# promotion gate passes, opens a human-reviewed PR with the promoted skill
+# overlay. The gate is evidence FOR a PR, never a bypass of one — nothing
+# merges without review.
+#
+# Activation checklist (the scheduled lane is OFF by default).
+# [ ] Configure the repository secret GITNEXUS_BENCH_AUTH_TOKEN (an Anthropic
+#     API key — benchmark sessions bill real usage; the Claude Code OAuth
+#     subscription token does not work here).
+# [ ] Run workflow_dispatch once and confirm: containment preflight passes,
+#     the benchmark completes inside the job timeout, the results artifact
+#     uploads, and a promotion (if any) opens a well-formed PR.
+# [ ] Set the repository variable GITNEXUS_EVOLUTION_ENABLED=true.
+# Roll back by setting that variable to false; workflow_dispatch remains available.
+name: GitNexus skill evolution
+
+on:
+  schedule:
+    - cron: '0 3 * * 6' # weekly, Saturday 03:00 UTC
+  workflow_dispatch:
+    inputs:
+      generations:
+        description: 'Propose→bench→gate generations to run'
+        required: false
+        default: '1'
+        type: string
+      runs:
+        description: 'Runs per arm per task (the gate needs at least 3)'
+        required: false
+        default: '3'
+        type: string
+      model:
+        description: 'Model for benchmark and proposer sessions'
+        required: false
+        default: 'claude-sonnet-5'
+        type: string
+      include_expensive:
+        description: 'Include tasks marked expensive: true'
+        required: false
+        default: false
+        type: boolean
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+permissions: {}
+
+jobs:
+  evolve:
+    name: Propose, benchmark, and gate skill candidates
+    if: >-
+      github.repository == 'abhigyanpatwari/GitNexus' &&
+      (
+        github.event_name == 'workflow_dispatch' ||
+        vars.GITNEXUS_EVOLUTION_ENABLED == 'true'
+      )
+    runs-on: ubuntu-latest
+    timeout-minutes: 355
+    permissions:
+      contents: read # The promotion PR uses a short-lived App token minted below.
+    env:
+      GENERATIONS: ${{ inputs.generations || '1' }}
+      RUNS: ${{ inputs.runs || '3' }}
+      MODEL: ${{ inputs.model || 'claude-sonnet-5' }}
+      INCLUDE_EXPENSIVE: ${{ inputs.include_expensive && '1' || '' }}
+    steps:
+      - name: Require the benchmark auth secret
+        env:
+          HAS_TOKEN: ${{ secrets.GITNEXUS_BENCH_AUTH_TOKEN != '' }}
+        run: |
+          set -euo pipefail
+          if [[ "${HAS_TOKEN}" != 'true' ]]; then
+            echo '::error::GITNEXUS_BENCH_AUTH_TOKEN is not configured. The evolution loop runs real benchmark sessions and needs an Anthropic API key (not the Claude Code OAuth token).'
+            exit 1
+          fi
+
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: '22.16.0'
+          cache: npm
+          cache-dependency-path: |
+            gitnexus/package-lock.json
+            gitnexus-shared/package-lock.json
+
+      - uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
+        with:
+          version: '0.11.23'
+          python-version: '3.13'
+          enable-cache: true
+          cache-dependency-glob: eval/uv.lock
+
+      - name: Install sandbox runtime and pinned Claude CLI
+        run: |
+          set -euo pipefail
+          sudo apt-get update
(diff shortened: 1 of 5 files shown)


magyargergo:
**Why this stays on `GITNEXUS_BENCH_AUTH_TOKEN` (API key) instead of `CLAUDE_CODE_OAUTH_TOKEN`** — verified empirically on the pinned Claude Code 2.1.214:

The bench runs every session in `--bare` mode, which deliberately ignores OAuth/keychain credentials (documented in `eval/workflow_bench/proposer_sandbox.py`, which injects the token as `ANTHROPIC_API_KEY`). Both possible wirings of the subscription token fail:

- `CLAUDE_CODE_OAUTH_TOKEN` env + `--bare` → `Not logged in · Please run /login`
- OAuth token passed as `ANTHROPIC_API_KEY` + `--bare` → `Invalid API key · Fix external API key`

Making the OAuth token work would mean removing `--bare` from the bench sessions, which is part of the eval sandbox's hermeticity model — out of scope here. The scheduled lane therefore needs a real Anthropic API key in `GITNEXUS_BENCH_AUTH_TOKEN` before activation.

magyargergo (commented):
## Automated tri-review — PR #2571 `feat(eval): run the skill-evolution loop online`

**Scope:** one added file, `.github/workflows/gitnexus-skill-evolution.yml` (+262). Base `94a528f5` → head `a8fdac52`.

**Engines.** Two independent review engines ran the change:
- **Engine A (Claude):** the GitNexus reviewer swarm (risk-architect, security-boundary, test-CI, docs-DoD, facts-historian) plus the `ce-code-review` persona set (correctness, reliability, adversarial, security, project-standards).
- **Engine B (Codex, gpt-5.6-sol / xhigh):** the GitNexus swarm (solo), `ce-code-review`, and a dedicated adversarial pass — **all three legs live and productive** (real findings, not skill-not-found).

Findings below are weighted by cross-engine corroboration; single-lane items are only elevated when a concrete trigger was verified in the worktree.

**Final verdict: not production-ready.** The workflow cannot open a promotion PR because `--apply` is never passed, so the entire App-token/PR path is unreachable dead code (Blocker 1). Even with that flag added, the first real run would fail at task-binding because the tasks hardcode `repo: ~/GitNexus`, which does not exist on a hosted runner, and the workflow never provisions it (Blocker 2). Neither blocker is caught pre-merge because no CI job runs this loop end-to-end. The one failing required check (CodeQL) is an infrastructure timeout, not a code defect. The remaining P2/P3 items are hardening that should be fixed alongside the blockers but do not independently block. **Merge state: checks failing** (the CodeQL infra flake leaves the required check red, so branch protection reports `BLOCKED`; re-running it should clear it). **Branch hygiene: clean feature/fix PR** (one new file on a clean branch, no unrelated churn).

---

### What's solid (validated, not merely unreviewed)

The security architecture holds up under both engines' probing:
- `permissions: {}` at top level, job-level `contents: read` only; the write path uses a short-lived, SHA-pinned `create-github-app-token@v3.2.0` App token (same pattern as `publish.yml`).
- `persist-credentials: false` on checkout; App token reaches `git push` only through a `0700` `GIT_ASKPASS` file built from a single-quoted heredoc — never in argv, git config, or the checkout.
- Benchmark secret scoped to the loop step only; `build_sandbox_environment` builds the session env **from scratch** (never copies `os.environ`), subprocess env-scrub + total egress deny inside the sandbox, and transcript redaction of the literal token + `sk-ant-…` pattern.
- A deterministic Python gate runs before anything would touch the tree; the boundary check (`git status --porcelain` allowlist) fails closed on any path outside the three skill trees.
- All actions SHA-pinned; canary Claude CLI double-pinned (lockfile + `--version` assertion).

Both engines also **refuted several suspicions** — recording these is a feature of the review:
- **Cursor-mirror drift is NOT a current bug.** [code-read] The loop can only promote `gitnexus-plan`/`gitnexus-work` (`evolution.py:25-27` `CANDIDATE_SKILLS`, hard-rejected otherwise), and neither is cursor-distributed; the cursor byte-identity test targets `gitnexus-review` only, which is an evaluation arm, never a candidate. Both engines cleared this. It becomes a latent P3 the day a cursor-distributed skill joins `CANDIDATE_SKILLS` — the PR body already defers that.
- **`skill_invoked` fail-open: cleared.** The current runner marks missing skill-invocation / unverifiable transcript as failed evidence (fail-closed).
- **Benchmark-token exposure to LLM/bwrap sessions: refuted** via three traced layers (from-scratch sandbox env, subprocess scrub, transcript redaction).
- **`PROMOTION_EOF` GITHUB_OUTPUT injection: not exploitable today** — `promotion.json` is machine `json.dumps` (newlines escaped, no LLM free text). Kept below only as hardening.
- Askpass token expansion, `workflow_dispatch` input injection, and boundary rename/quotepath escapes were each probed and refuted with concrete traces.

---

### Headline — two merge blockers keep the workflow from doing its job

Both engines independently conclude the workflow **cannot accomplish its advertised purpose** as written. Engine B (Codex swarm) reached both; Engine A + coordinator source-reads confirm each.

#### Blocker 1 — P1, cross-engine + source-verified: the promotion-PR path is unreachable dead code (missing `--apply`)

**`.github/workflows/gitnexus-skill-evolution.yml:151`** [code-read]

The loop invocation (lines 151-158) passes `--tasks --model --generations --runs --claude-bin --out-root [--include-expensive]` but **never `--apply`**. In `evolve.py`, `--apply` is `action="store_true"` defaulting to `False` (`evolve.py:814-818`), and the **only** writer into the checkout — `apply_promoted_overlay` — is gated behind `if args.apply:` (`evolve.py:1017`). Without the flag, a gate-passing generation prints `PROMOTED — evidence in …`, falls into the else branch that prints `"Re-run with --apply…"` (`evolve.py:1033`), and returns 0 without mutating the tree.

Consequently the "Detect and bound the applied promotion" step (`git status --porcelain`, line 176) always sees a clean tree → sets `promoted=false` (line 179) → the Mint-App-Token (line 206) and Open-PR (line 213) steps, both gated on `promoted == 'true'`, **never run**. A gate-PASS run goes green printing "No promotion this run; the incumbent skills stand." — which is false on a promotion. The workflow's entire advertised deliverable (header lines 2-5: "opens a human-reviewed PR with the promoted skill overlay") can never fire.

**Corroboration:** Engine A — risk-architect (P0), reliability (P0), adversarial (P0), correctness (P1). Engine B — swarm ("confirmed blocker"), ce (P1, confidence 100), adversarial ("the promotion path is dead"). Coordinator source-read confirms.

**Fix:** append `--apply` to the invocation. `validate_promotion_for_apply` already runs in the `try` *before* the apply gate (`evolve.py:1003-1012`), so the deterministic gate is not weakened by adding the flag. Add a workflow-contract test asserting `--apply` stays present.

#### Blocker 2 — P1, Codex swarm + source-verified: the benchmark can't resolve its task repo on a hosted runner

**`.github/workflows/gitnexus-skill-evolution.yml` (loop step, lines 151-159 / checkout 79-82)** [code-read]

Every scenario in `eval/workflow_bench/tasks.scenarios.yaml` hardcodes `repo: ~/GitNexus` (lines 38, 69, 93, 119), and `runner_tasks.py:115` resolves it with `Path(task["repo"]).expanduser().resolve()` then immediately runs `git -C <repo> rev-parse --show-toplevel` (lines 116-118, via `run_checked`, which raises on failure). On a GitHub-hosted runner `~` is `/home/runner`, so `~/GitNexus` → `/home/runner/GitNexus`, but `actions/checkout` places the repo at `$GITHUB_WORKSPACE` = `/home/runner/work/GitNexus/GitNexus`. **The workflow never creates `~/GitNexus`, and never exports `GITNEXUS_BENCH_ORACLE_ROOT`** (which the oracle tasks reference, e.g. tasks.scenarios.yaml:107). So the first scheduled/dispatched run fails at task-binding — a *loud* red failure (unlike Blocker 1's silent green) — before any benchmark session runs.

Nothing catches this pre-merge: the passing `benchmarks (GITNEXUS_BENCH)` CI job runs npm/vitest micro-benchmarks (ci-tests.yml:455-514), not the `evolve.py` loop over `tasks.scenarios.yaml`, and the eval-containment jobs exercise unit/session tests with synthetic repos — so no CI job runs this workflow's actual code path (see the test gap below).

**Corroboration:** Engine B — Codex swarm (P1, "cannot resolve the benchmark repository on a hosted runner"). Coordinator source-read confirms. (Engine A lanes reviewed the workflow control-flow and promotion path; none traced the task-repo resolution against the runner filesystem — a genuine cross-engine catch by Codex's GitNexus-graph-assisted lane.)

**Fix:** provision the task repo/oracle roots the workflow's own checkout already contains — e.g. symlink `~/GitNexus` → `$GITHUB_WORKSPACE` and `export GITNEXUS_BENCH_ORACLE_ROOT=…` in the loop step — or make the task `repo`/oracle-root paths configurable and pass the CI workspace paths. Cover it with the same workflow-contract/integration test that Blocker 1 needs.

---

### Inline findings (posted on the diff)

1. **P1 — missing `--apply` (line 151)** — Blocker 1 above. (Blocker 2, the unresolvable task repo, is a diffuse "missing setup" spanning the loop step + checkout, so it's written up in the headline and body rather than pinned to one line.) [code-read]
2. **P2 — promotion branch name omits run attempt (line 221).** `branch="evolution/skills-${GITHUB_RUN_ID}"`; `GITHUB_RUN_ID` is stable across re-runs (only `GITHUB_RUN_ATTEMPT` increments). If `git push` (236) succeeds but `gh pr create` (257) fails, a re-run re-benchmarks, commits a different SHA, and its non-force push to the existing branch is rejected non-fast-forward — the lane is stuck until someone manually deletes the branch. Cross-engine (A risk/correctness/adversarial/reliability P2; B ce/adversarial P2). Fix: add `${GITHUB_RUN_ATTEMPT}` to the branch name (the artifact name already does, line 165). [code-read]
3. **P3 — mislabeled action pin (line 163).** `actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v6.0.0` — that exact SHA is labeled `# v7.0.1` in the other ~9 workflows in this repo that pin it, and the upstream tag `v7.0.1` resolves to this SHA (`v6.0.0` is a different SHA). Pin is safe; the comment is wrong and would mislead a future bump. Engine A project-standards (confidence 97). Fix: change the comment to `# v7.0.1`. [code-read]
4. **P3 — fixed `PROMOTION_EOF` GITHUB_OUTPUT delimiter (line 197).** Non-exploitable today (machine JSON, newlines escaped), but the repo already uses a random per-run delimiter for this exact hazard class in `tree-sitter-upgrade-readiness.yml`. Both engines refute as exploitable; keep as one-line hardening. [code-read]

---

### Lower-priority / body-only items

- **P2 hardening — `workflow_dispatch` runs from the dispatched ref with the secrets (cross-engine severity split).** Manual dispatch executes this workflow and in-tree `evolve.py` from the selected branch (checkout has no pinned `ref`); a write-access collaborator can dispatch a modified branch that runs with `GITNEXUS_BENCH_AUTH_TOKEN` and can mint the `RELEASE_APP` token. **Engine B (Codex) rates this a blocking-class concern (high/P2); Engine A security lanes treated "requires write access" as accepted trust and did not rate it merge-blocking** — an honest cross-engine disagreement. The precondition is an already-write-privileged actor, and it is a general property of dispatch + secrets, so we surface it as a real hardening recommendation rather than an active exploit: bind the job to a protected Environment restricted to `main`, and/or reject non-`main` dispatch refs + checkout the trusted `main` SHA before running repo code.
- **P3 — multi-generation PR-body summary.** `find … promotion.json | xargs cat | tail -c 8000` (lines 197-201) concatenates every generation's `promotion.json` (including rejected ones) in filesystem order; with `generations > 1` the PR body can show a losing generation's decisions or a truncated mix. Evidence artifact retains ground truth. Fix: emit/select only the promoted generation's file.
- **P3 — timeout vs internal watchdog mismatch.** `timeout-minutes: 355` is well below `evolve.py`'s own per-generation watchdog budget, so the internal watchdog can't fire in CI and no upper bound is enforced on operator-supplied `generations`/`runs`. Operational, not a correctness bug.
- **P3 — docs/checklist gaps.** (a) The activation checklist doesn't state that `workflow_dispatch` runs the full billing loop even when `GITNEXUS_EVOLUTION_ENABLED` is false; (b) `RELEASE_APP_ID`/`RELEASE_APP_PRIVATE_KEY` are absent from the checklist yet the Mint-App-Token step hard-fails (not graceful-skips) if missing on a promotion; (c) weekly cron vs README's "not on a tight schedule" guidance.
- **P3 — App-token least-privilege (Codex swarm).** The Mint-App-Token step (lines 204-210) passes only `app-id` + `private-key` with no `permission-*` inputs, so the minted token inherits *all* of the installation's permissions (the Release App is documented as having `Workflows: write` at publish.yml:307) — broader than the `contents: write` + `pull-requests: write` this job actually needs to push a branch and open a PR. This matches the repo's existing App-token usage (publish.yml, grammar-update-monitor.yml, build-tree-sitter-prebuilds.yml all mint it unscoped), so it is a pre-existing pattern rather than a regression, but this workflow's final shell step handles LLM-authored content immediately before minting — a good place to tighten. Fix: add `permission-contents: write` and `permission-pull-requests: write` (and prefer the `client-id` input, as publish.yml already does). This also blunts the `workflow_dispatch` hardening item above.
- **P3 — App-token scope is externally unverifiable** (config check: confirm the App install is Contents:RW + PullRequests:RW on this repo only).
- **P3 — defense-in-depth (pre-existing eval code, out of this PR's diff):** proposer overlay/`proposal.md` and `results.jsonl` `stderr_tail` are uploaded to the 14-day artifact without `redact_text`. Fix belongs upstream in the eval code, not this workflow.
- **Near-non-issue:** two `run:` blocks (lines 128, 134) omit an explicit `set -euo pipefail`, but GitHub's default `run:` shell is already `bash -eo pipefail` and neither block references shell vars, so `-u` adds nothing here. Mentioned only for completeness.

**Test gap (both engines) — the root cause of both blockers reaching this point.** No workflow-contract or integration test exercises this workflow's actual path: that a gate-passing run resolves its task repo, mutates the checkout, and activates the PR path (or even that `--apply` is present). A no-promotion run is observationally identical to the Blocker-1 state, and no pre-merge job runs the `evolve.py` loop against `tasks.scenarios.yaml` on a hosted runner, so Blocker 2 is invisible until the first real run. One integration test that dispatches the loop end-to-end (or a lighter contract test parsing the workflow + a smoke run) would have caught both.

---

### CI status

`mergeStateStatus: BLOCKED` is branch protection; `mergeable: MERGEABLE`. The only failing check is CodeQL `Analyze (javascript-typescript)` — `Connect Timeout Error (api.github.com:443)` during SARIF upload, an infrastructure flake; CodeQL passes on `main` at the exact merge-base. actionlint, zizmor, quality gates, the full test matrix, benchmarks, and eval-containment all pass. Re-running the CodeQL job should clear the check.

---

*This is an automated multi-tool digest (Claude swarm + ce personas + Codex). Verify the file:line evidence before acting on any item.*

---

.github/workflows/gitnexus-skill-evolution.yml
```diff
@@ -0,0 +1,262 @@
...
+          GITNEXUS_BENCH_AUTH_TOKEN: ${{ secrets.GITNEXUS_BENCH_AUTH_TOKEN }}
+        run: |
+          set -euo pipefail
+          out_root="${RUNNER_TEMP}/wfevolve"
+          echo "out_root=${out_root}" >> "${GITHUB_OUTPUT}"
+          extra=()
+          if [[ -n "${INCLUDE_EXPENSIVE}" ]]; then
+            extra+=(--include-expensive)
+          fi
+          uv run --locked --extra dev python -m workflow_bench.evolve \
```
magyargergo, line 190: **P1 — the promotion-PR path is unreachable dead code (missing `--apply`).** `[code-read]`

This invocation never passes `--apply`. In `evolve.py`, `--apply` is `store_true` / default-False (`evolve.py:814`), and the only writer into the checkout — `apply_promoted_overlay` — is gated behind `if args.apply:` (`evolve.py:1017`); without it a gate-passing generation prints "Re-run with --apply…" (`evolve.py:1033`) and returns 0 without touching the tree. So the "Detect and bound the applied promotion" step (`git status --porcelain`, line 176) always sees a clean tree → `promoted=false` (line 179) → the Mint-App-Token (line 206) and Open-PR (line 213) steps, both gated on `promoted == 'true'`, never run. A gate-PASS run goes green as "No promotion this run". Confirmed by both engines (Claude risk/reliability/adversarial/correctness; Codex swarm/ce/adversarial).

**Fix:** append `--apply` to the invocation. `validate_promotion_for_apply` already runs before the apply gate (`evolve.py:1003-1012`), so the deterministic gate is not weakened. Add a workflow-contract test asserting `--apply` stays present (that test would also catch Blocker 2 in the summary).

.github/workflows/gitnexus-skill-evolution.yml
```diff
@@ -0,0 +1,262 @@
...
+      - name: Open the promotion PR
+        if: steps.promotion.outputs.promoted == 'true'
+        env:
+          APP_TOKEN: ${{ steps.app-token.outputs.token }}
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          PROMOTION_SUMMARY: ${{ steps.promotion.outputs.summary }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          set -euo pipefail
+          branch="evolution/skills-${GITHUB_RUN_ID}"
```
magyargergo: **P2 — promotion branch name omits the run attempt.** `[code-read]`

`GITHUB_RUN_ID` is stable across re-runs (only `GITHUB_RUN_ATTEMPT` increments). If `git push` (line 236) succeeds but `gh pr create` (line 257) fails, a re-run re-benchmarks, commits a different SHA, and its non-force push to the already-existing branch is rejected non-fast-forward — the lane is stuck until someone manually deletes the branch. Cross-engine (Claude risk/correctness/adversarial/reliability; Codex ce).

**Fix:** include `${GITHUB_RUN_ATTEMPT}` in the branch name (the artifact name at line 165 already does).

.github/workflows/gitnexus-skill-evolution.yml
```diff
@@ -0,0 +1,262 @@
...
+            --generations "${GENERATIONS}" \
+            --runs "${RUNS}" \
+            --claude-bin "${RUNNER_TEMP}/claude-canary/node_modules/@anthropic-ai/claude-code-linux-x64/claude" \
+            --out-root "${out_root}" \
+            "${extra[@]}"
+        working-directory: eval
+
+      - name: Upload benchmark evidence
+        if: always() && steps.loop.outputs.out_root != ''
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v6.0.0
```
magyargergo: **P3 — mislabeled action pin.** `[code-read]`

This exact SHA (`043fb46d…`) is `actions/upload-artifact` **v7.0.1**, and it is labeled `# v7.0.1` in the other workflows in this repo that pin it (ci.yml, ci-tests.yml, ci-e2e.yml, build-tree-sitter-prebuilds.yml, pr-autofix.yml, scorecard.yml, impact-pdg-mutation-report.yml). The pin itself is safe (it's the SHA that binds); only the `# v6.0.0` comment is wrong and would mislead a future version bump.

**Fix:** change the comment to `# v7.0.1`.

.github/workflows/gitnexus-skill-evolution.yml
```diff
@@ -0,0 +1,262 @@
...
+              .claude/skills/*|gitnexus/skills/*|gitnexus-claude-plugin/skills/*) ;;
+              *)
+                echo "::error::Promotion touched a path outside the skill trees: ${path}"
+                exit 1
+                ;;
+            esac
+          done <<< "${changed}"
+          echo "promoted=true" >> "${GITHUB_OUTPUT}"
+          {
+            echo 'summary<<PROMOTION_EOF'
```
magyargergo: **P3 — fixed `GITHUB_OUTPUT` heredoc delimiter (hardening).** `[code-read]`

`PROMOTION_EOF` is a fixed delimiter. Not exploitable today — `promotion.json` is machine `json.dumps` output (newlines escaped, no LLM free text), so a bare `PROMOTION_EOF` line can't occur — but the repo already uses a per-run random delimiter for this exact class in `tree-sitter-upgrade-readiness.yml`. Both engines refuted the injection as reachable; filing only as one-line hardening.

**Fix:** use a per-run random delimiter, e.g. `DELIM="PROMOTION_EOF_$(openssl rand -hex 16)"` for both the open and close markers.

PR #2572, opened 2026-07-19
feat(ci): review agent runs as a coordinated reviewer swarm

magyargergo:
Converts the CI review agent into a coordinated reviewer swarm, with the orchestration owned by the `gitnexus-review` skill (not the workflow prompt), so any runner of the skill triggers the lanes consistently.

## Architecture

- **Skill-owned dispatch** — the skill's new "Swarm lanes" section defines the roster, dispatch shape (parallel finders, one message, per-lane context), verification contract (lane reports are unverified claims: re-anchor, dedup, drop unanchored), and the critic gate. The workflow prompt keeps only CI deltas: trusted-control-SHA install provenance, Task-tool surface, and the publisher's orchestrator-only evidence gate.
- **Five finder lanes**, dispatched in parallel: `ci-correctness-lens`, `ci-security-lens`, `ci-blast-radius-lens`, `ci-coverage-lens`, and `ci-adversarial-lens` (assumes the change is broken; constructs reachable failure scenarios — interleavings, hostile inputs, state corruption, new-surface abuse — each verified to a concrete entry point).
- **One critic gate**, `ci-critic-lens`, run last on the finished draft: audits anchoring, concreteness, severity calibration, conformance, and honesty; returns `PASS` or a numbered defect list with the smallest repair per item. Bounded to two passes — it hardens the review, never blocks it.
- Personas live in the canonical skill tree (`.claude/skills/gitnexus-review/ci-personas/`), mirrored byte-identical to all three shipped trees, and are installed into the CI reviewer's user-scope agents dir **from the exact control SHA** — a hostile PR head can never define a lane. All lanes are read-only: `Read`/`Glob`/`Grep` + safe graph MCP tools.

## Security posture (unchanged guarantees, one fix)

- `Task` allowed; `Bash` and `Agent` stay denied for every context.
- **Evidence-gate fix** (found by this workflow's own review of this PR): `proveGraphReview()` now excludes sidechain transcript entries (non-null `parent_tool_use_id`) from evidence, so only the orchestrator's own `context` call satisfies the publisher gate; malformed parent linkage fails the transcript. Regression fixtures prove sidechain-only evidence is rejected and mainline evidence beside sidechain turns still passes.
- `--max-turns` 100 → 150 and analyze timeout 45 → 75 min for swarm headroom.
- The workflow security-contract test pins the swarm posture; 122 tests across the four skill/workflow contract suites pass.

### DIFF
--- a/.github/workflows/gitnexus-review-agent.yml
+++ b/.github/workflows/gitnexus-review-agent.yml
@@ -1175,11 +1175,20 @@ jobs:
             gate. Adapt the skill's checkout/index steps to this pre-aligned environment.
 
             Return one structured field named body containing the complete Markdown
-            review: prioritized findings with file:line evidence, change and blast-radius
-            summary, coverage and residual risk, and the skill's verdict wording. Do not
+            review, structured exactly as: first a short opening paragraph that leads
+            with the skill's verdict wording and a plain-language summary of what the
+            PR does; then "### Findings" ordered by severity (CRITICAL, HIGH, MEDIUM,
+            LOW), one bold-severity bullet per finding stating the one-sentence claim
+            followed by an indented evidence line; then "### Change summary and blast
+            radius"; then "### Coverage and residual risk". Every file, line, or symbol
+            reference anywhere in the body must be a clickable Markdown link — never
+            bare `path:line` text. Link head files as
+            https://github.com/${{ github.repository }}/blob/${{ steps.context.outputs.head_sha }}/PATH#L10-L20
+            (exact analyzed head SHA, real line range) and deleted or rename-old paths
+            as the same URL shape at ${{ steps.inputs.outputs.merge_base }}. Do not
             include an HTML publication marker and do not mention users or teams.
           claude_args: |
-            --model claude-sonnet-4-5-20250929
+            --model claude-sonnet-5
             --add-dir "${{ runner.temp }}/gitnexus-review-pr-target"
             --setting-sources user
             --disable-slash-commands
--- a/.github/workflows/gitnexus-review-agent.yml
+++ b/.github/workflows/gitnexus-review-agent.yml
@@ -848,10 +848,11 @@ jobs:
           # This passive tree is mounted with --add-dir, which the runtime scans
           # for spawnable subagent definitions in .claude/agents/, and there is no
           # env to suppress that on the pinned runtime. Drop any PR-controlled
-          # agent definitions so only the trusted control-SHA personas installed
-          # into CLAUDE_CONFIG_DIR/agents can ever be dispatched. Skills are left
+          # agent definitions (at any depth, to also cover monorepo subpackages)
+          # so only the trusted control-SHA personas installed into
+          # CLAUDE_CONFIG_DIR/agents can ever be dispatched. Skills are left
           # intact so a PR that legitimately edits skills stays reviewable.
-          rm -rf -- "${review_dir}/.claude/agents"
+          find "${review_dir}" -type d -path '*/.claude/agents' -prune -exec rm -rf -- {} +
           export GIT_ALTERNATE_OBJECT_DIRECTORIES="${GITHUB_WORKSPACE}/.git/objects"
 
           merge_base="$(git -C pr-target merge-base "${BASE_SHA}" "${HEAD_SHA}")"
@@ -1227,7 +1228,7 @@ jobs:
             --strict-mcp-config
             --mcp-config "${{ runner.temp }}/gitnexus-review-mcp.json"
             --tools "Read,Glob,Grep,Agent"
-            --allowedTools "Agent(ci-correctness-lens,ci-security-lens,ci-blast-radius-lens,ci-coverage-lens,ci-adversarial-lens,ci-critic-lens),Glob,Grep,Read(./**),Read(${{ runner.temp }}/gitnexus-review-pr-target/**),Read(${{ runner.temp }}/gitnexus-review-merge-base/**),mcp__gitnexus__list_repos,mcp__gitnexus__query,mcp__gitnexus__context,mcp__gitnexus__check,mcp__gitnexus__impact,mcp__gitnexus__explain,mcp__gitnexus__pdg_query,mcp__gitnexus__route_map,mcp__gitnexus__tool_map,mcp__gitnexus__shape_check,mcp__gitnexus__api_impact,mcp__gitnexus__trace"
+            --allowedTools "Agent(ci-correctness-lens,ci-security-lens,ci-blast-radius-lens,ci-coverage-lens,ci-adversarial-lens,ci-critic-lens),Read(./**),Read(${{ runner.temp }}/gitnexus-review-pr-target/**),Read(${{ runner.temp }}/gitnexus-review-merge-base/**),mcp__gitnexus__list_repos,mcp__gitnexus__query,mcp__gitnexus__context,mcp__gitnexus__check,mcp__gitnexus__impact,mcp__gitnexus__explain,mcp__gitnexus__pdg_query,mcp__gitnexus__route_map,mcp__gitnexus__tool_map,mcp__gitnexus__shape_check,mcp__gitnexus__api_impact,mcp__gitnexus__trace"
             --disallowedTools "Bash,Write,Edit,MultiEdit,NotebookEdit,WebFetch,WebSearch,Skill,Read(/proc/**),Read(/sys/**),Read(/dev/**),Read(${{ github.workspace }}/**),mcp__github,mcp__gitnexus__detect_changes,mcp__gitnexus__rename,mcp__gitnexus__cypher,mcp__gitnexus__group_list,mcp__gitnexus__group_sync"
             --permission-mode dontAsk
             --no-session-persistence
--- a/.github/workflows/gitnexus-review-agent.yml
+++ b/.github/workflows/gitnexus-review-agent.yml
@@ -47,7 +47,7 @@ jobs:
         )
       )
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 75
     permissions:
       contents: read # Check out trusted control code and the passive PR tree.
       pull-requests: read # Resolve and revalidate the exact PR head/base tuple.
@@ -1179,17 +1179,19 @@ jobs:
             graph tools remain available for the review, but do not satisfy this evidence
             gate. Adapt the skill's checkout/index steps to this pre-aligned environment.
 
-            Run the expert-lens pass as the swarm defined in the skill's "Swarm lanes"
-            section: the lanes — finder lanes and the critic gate alike — are
-            installed as spawnable agents from the exact control SHA. Dispatch and
-            verify them via the Task tool exactly as that section instructs, handing
-            each lane the PR number, the exact head and merge-base SHAs, the absolute
-            paths of review-input/pr.diff and review-input/changed-paths.json, the
-            passive head directory, and the merge-base directory. In this environment,
-            lane tool calls never
+            The skill's "Swarm lanes" section governs the expert-lens pass, including
+            lane dispatch, verification, the critic gate, and every fallback. All six
+            lanes are pre-installed as spawnable agents from the exact control SHA;
+            the Task tool exists solely to dispatch them. Map the section's generic
+            context to this environment when handing lanes their inputs: the diff is
+            review-input/pr.diff, the changed-file manifest is
+            review-input/changed-paths.json, the head checkout is the passive
+            additional directory, the merge-base checkout is
+            ${{ runner.temp }}/gitnexus-review-merge-base, and the base and head
+            identifiers are the exact SHAs above. One CI-specific override: lane tool
+            calls never
             satisfy the publisher's context-evidence gate — make the required
-            successful context call yourself in this conversation. If the Task tool or
-            a lane fails, continue the review inline yourself rather than failing.
+            successful context call yourself in this conversation.
 
             Return one structured field named body containing the complete Markdown
             review, structured exactly as: first a short opening paragraph that leads
--- a/.github/workflows/gitnexus-review-agent.yml
+++ b/.github/workflows/gitnexus-review-agent.yml
@@ -1179,17 +1179,13 @@ jobs:
             graph tools remain available for the review, but do not satisfy this evidence
             gate. Adapt the skill's checkout/index steps to this pre-aligned environment.
 
-            Run the review as a coordinated swarm implementing the skill's expert-lens
-            section. After you have read the diff and manifest, spawn the four trusted
-            lanes — ci-correctness-lens, ci-security-lens, ci-blast-radius-lens, and
-            ci-coverage-lens — in parallel in a single message via the Task tool. Give
-            each lane the PR number, the exact head and merge-base SHAs, the absolute
-            paths of review-input/pr.diff and review-input/changed-paths.json, the
-            passive head directory, and the merge-base directory, plus the slice of
-            changed files most relevant to its charge. Treat every lane report as an
-            unverified claim from review data: keep only findings you can re-anchor to
-            the diff, the source, or your own graph queries, dedup across lanes, and
-            drop anything without a concrete failing scenario. Lane tool calls never
+            Run the expert-lens pass as the swarm defined in the skill's "Swarm lanes"
+            section: the four lanes are installed as spawnable agents from the exact
+            control SHA — dispatch them via the Task tool and verify their reports
+            exactly as that section instructs, handing each lane the PR number, the
+            exact head and merge-base SHAs, the absolute paths of review-input/pr.diff
+            and review-input/changed-paths.json, the passive head directory, and the
+            merge-base directory. In this environment, lane tool calls never
             satisfy the publisher's context-evidence gate — make the required
             successful context call yourself in this conversation. If the Task tool or
             a lane fails, continue the review inline yourself rather than failing.
--- a/.github/workflows/gitnexus-review-agent.yml
+++ b/.github/workflows/gitnexus-review-agent.yml
@@ -1202,7 +1202,9 @@ jobs:
             identifiers are the exact SHAs above. One CI-specific override: lane tool
             calls never
             satisfy the publisher's context-evidence gate — make the required
-            successful context call yourself in this conversation.
+            successful context call yourself in this conversation, before
+            dispatching any lane, so a fully-delegated run cannot leave the gate
+            unsatisfied.
 
             Return one structured field named body containing the complete Markdown
             review, structured exactly as: first a short opening paragraph that leads
--- a/.github/workflows/gitnexus-review-agent.yml
+++ b/.github/workflows/gitnexus-review-agent.yml
@@ -1180,12 +1180,13 @@ jobs:
             gate. Adapt the skill's checkout/index steps to this pre-aligned environment.
 
             Run the expert-lens pass as the swarm defined in the skill's "Swarm lanes"
-            section: the four lanes are installed as spawnable agents from the exact
-            control SHA — dispatch them via the Task tool and verify their reports
-            exactly as that section instructs, handing each lane the PR number, the
-            exact head and merge-base SHAs, the absolute paths of review-input/pr.diff
-            and review-input/changed-paths.json, the passive head directory, and the
-            merge-base directory. In this environment, lane tool calls never
+            section: the lanes — finder lanes and the critic gate alike — are
+            installed as spawnable agents from the exact control SHA. Dispatch and
+            verify them via the Task tool exactly as that section instructs, handing
+            each lane the PR number, the exact head and merge-base SHAs, the absolute
+            paths of review-input/pr.diff and review-input/changed-paths.json, the
+            passive head directory, and the merge-base directory. In this environment,
+            lane tool calls never
             satisfy the publisher's context-evidence gate — make the required
             successful context call yourself in this conversation. If the Task tool or
             a lane fails, continue the review inline yourself rather than failing.
--- a/.github/workflows/gitnexus-review-agent.yml
+++ b/.github/workflows/gitnexus-review-agent.yml
@@ -47,7 +47,7 @@ jobs:
         )
       )
     runs-on: ubuntu-latest
-    timeout-minutes: 45
+    timeout-minutes: 60
     permissions:
       contents: read # Check out trusted control code and the passive PR tree.
       pull-requests: read # Resolve and revalidate the exact PR head/base tuple.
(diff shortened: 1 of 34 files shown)


magyargergo:
@gitnexus review

magyargergo:
Both findings addressed in df15057c:

**HIGH — sidechain evidence bypass: fixed.** `proveGraphReview()` now reads `parent_tool_use_id` on every `assistant`/`user` transcript entry. Sidechain entries (non-null parent) are still strictly validated — a malformed linkage fails the whole transcript with `invalid_execution_transcript` — but they are excluded from both candidate `context` calls and qualifying results, so only the orchestrator's own mainline call can satisfy the gate, exactly as the prompt promises. The exact regression fixture the review asked for is added: a transcript whose only successful context call is sidechain now produces `status: failure` / `missing_graph_evidence`, and a companion case proves mainline evidence beside sidechain turns still passes (39/39 contract tests green — the harness executes the real embedded validator script).

**LOW — turn budget: addressed.** `--max-turns` raised 100 → 150 to cover the orchestrator's lane dispatch and verification turns (lane sessions have their own budgets; the main loop only spends turns on Task calls and re-verification, so 150 is generous headroom).

One note on the review's caveat: its HIGH mechanism was source-verified but not live-verified — the fix is defensive in both directions regardless: if the CLI's execution file includes sidechain turns they can no longer qualify, and if it never includes them the filter is inert.

.github/workflows/gitnexus-review-agent.yml
```diff
@@ -1194,12 +1213,12 @@ jobs:
...
-            --disallowedTools "Bash,Write,Edit,MultiEdit,NotebookEdit,WebFetch,WebSearch,Skill,Task,Agent,Read(/proc/**),Read(/sys/**),Read(/dev/**),Read(${{ github.workspace }}/**),mcp__github,mcp__gitnexus__detect_changes,mcp__gitnexus__rename,mcp__gitnexus__cypher,mcp__gitnexus__group_list,mcp__gitnexus__group_sync"
+            --tools "Read,Glob,Grep,Task"
+            --allowedTools "Task,Read(./**),Read(${{ runner.temp }}/gitnexus-review-pr-target/**),mcp__gitnexus__list_repos,mcp__gitnexus__query,mcp__gitnexus__context,mcp__gitnexus__check,mcp__gitnexus__impact,mcp__gitnexus__explain,mcp__gitnexus__pdg_query,mcp__gitnexus__route_map,mcp__gitnexus__tool_map,mcp__gitnexus__shape_check,mcp__gitnexus__api_impact,mcp__gitnexus__trace"
+            --disallowedTools "Bash,Write,Edit,MultiEdit,NotebookEdit,WebFetch,WebSearch,Skill,Agent,Read(/proc/**),Read(/sys/**),Read(/dev/**),Read(${{ github.workspace }}/**),mcp__github,mcp__gitnexus__detect_changes,mcp__gitnexus__rename,mcp__gitnexus__cypher,mcp__gitnexus__group_list,mcp__gitnexus__group_sync"
```
abhigyanpatwari: **P1 — the swarm can never dispatch on the pinned runtime (Task/Agent inversion).** On the pinned Claude Code 2.1.214 the subagent-dispatch tool is `Agent` (`Task` was renamed to `Agent` in 2.1.63 and survives only as a legacy alias), and permission rules evaluate deny → ask → allow, so a deny always beats an allow. Here `Agent` is in `--disallowedTools` (this line), only `Task` is in `--allowedTools` (1217), and `Agent` is absent from `--tools` (1216) — so the orchestrator can never dispatch a lane. Every review silently falls back to the SKILL.md inline single-agent path ("If subagent dispatch is unavailable… run that lane's charge inline", SKILL.md:213-215), so the swarm this PR introduces never runs in CI. Fail-safe (the fallback still yields a valid single-agent review; no crash, no wrong verdict, no security hole) but the feature is inert, and the tests at `review-agent-workflow.test.ts:1147-1161` assert the inverted split so CI certifies it as correct.

Ground truth: a grep of the installed 2.1.214 binary shows the dispatch tool registered as `Agent`; current Claude Code docs confirm the 2.1.63 rename and deny-over-allow precedence; the repo's own `gitnexus-pr-swarm-review` skill already dispatches via `Agent`. Independently reached by two Codex review legs.

**Fix:** use `Agent` consistently — add it to `--tools`, allow it scoped to the six personas (`Agent(ci-correctness-lens)`, `Agent(ci-security-lens)`, `Agent(ci-blast-radius-lens)`, `Agent(ci-coverage-lens)`, `Agent(ci-adversarial-lens)`, `Agent(ci-critic-lens)` — which also closes the `--add-dir` agent-discovery concern), remove `Agent` from `--disallowedTools`, and update the prompt at 1163 and the test assertions. Add a pinned-CLI smoke test that actually dispatches one installed persona — a text-only assertion cannot catch this class. `[reproduced via 2.1.214 binary + docs + Codex B1/B3]`

.github/workflows/gitnexus-review-agent.yml
```diff
@@ -1194,12 +1213,12 @@ jobs:
...
             --permission-mode dontAsk
             --no-session-persistence
-            --max-turns 100
+            --max-turns 150
```
abhigyanpatwari, line 1325: **P2 — swarm transcript volume can overflow the unchanged hard caps; no per-agent turn budget.** `--max-turns` was raised 100→150 for the swarm (this line), but the transcript caps were not: `MAX_TRANSCRIPT_MESSAGES = 1_000` / `MAX_TRANSCRIPT_BYTES = 8_000_000` (1261-1262) are byte-identical to base, and both are hard throws — the envelope check at 1633 and `readStrictJsonFile` at 1627 — mapping to `failure_code: invalid_execution_transcript`, which discards the entire review. All lane sidechains stream into the same execution file (the premise of this PR's sidechain exclusion), and no persona carries a `maxTurns` frontmatter bound, so a thorough six-lane run on a large PR can plausibly exceed either cap and brick a valid review fail-closed. Latent today because of the P1 (no lanes dispatch), but live the moment the P1 is fixed. Fix alongside the P1: raise the caps for swarm volume (or exempt sidechain entries from size accounting — they can never carry evidence) and add per-persona `maxTurns`; add a boundary test with a sidechain-inflated transcript. `[code-read; corroborated by Claude ce-correctness/ce-adversarial/ce-reliability + Codex B1]`

gitnexus/test/unit/review-agent-workflow.test.ts
```diff
@@ -1355,6 +1371,38 @@ describe('gitnexus review-agent workflow security contract', () => {
...
+      toolUseId: 'tool-side-1',
+    });
+    const main = reviewTranscript();
+    const combined = [main[0], side[1], side[2], main[1], main[2], main[3]];
+    const withMainline = runArtifactScenario({
+      rawTranscript: JSON.stringify(combined),
+    });
+    expect(withMainline.artifact).toMatchObject({
+      status: 'success',
+      failure_code: null,
```
abhigyanpatwari, line 1522: **P2 — neither `!sidechain` guard is independently pinned (mutation-verified).** This test's three scenarios cannot observe the two guards separately: deleting only the acceptance-side guard (workflow:1736), or only the registration-side guard (workflow:1702), still passes **all 39 tests** — verified two independent ways (direct workflow mutation + vitest rerun, and an extracted-heredoc mutant harness). Every fixture keeps a tool_use/tool_result pair's sidechain status consistent, so each guard is redundant on all tested inputs; only deleting both is caught. This combined/success scenario also passes identically with both guards reverted (the mainline pair alone suffices), so it exercises but does not pin the exclusion — a future simplification dropping one side would ship green and leave the subagent-satisfies-the-gate bug one producer-drift from reopening. One-line fix: add a cross-wired fixture — a mainline `tool_use` whose matching `tool_result` carries a non-null `parent_tool_use_id` (take `reviewTranscript()` and set `parent_tool_use_id: 'toolu-parent-1'` on the result entry only) — expecting `missing_graph_evidence`; that kills the acceptance-guard mutant, and with the existing sidechain-only case pins both guards. `[reproduced — mutation-verified by two lanes independently]`

abhigyanpatwari (commented):
# GitNexus tri-review digest — PR #2572 (coordinated reviewer swarm)

**Verdict: not production-ready as-is — the swarm-dispatch permission fix below should land before this feature is enabled.** The PR's architecture is sound and every security surface probed held up, but on the pinned Claude Code 2.1.214 runtime the workflow denies the very tool that dispatches subagents, so the six-lane swarm — the PR's entire purpose — can never run in CI; every review would silently fall back to the existing inline single-agent path, and the new tests certify the broken configuration as correct. This is fail-safe (the fallback still produces a valid single-agent review; no crash, no wrong verdict, no security hole), but the feature is inert, which is merge-blocking for a PR titled "review agent runs as a coordinated reviewer swarm."

Engine breakdown for this digest: **Claude** ran the six-lane review swarm (facts/history, branch hygiene, risk, security-boundary, test/CI, docs/DoD) plus seven ce-code-review personas (correctness, adversarial, testing, security, maintainability, project-standards, reliability). **Codex (GPT-5.6-sol)** ran three legs: a full seven-persona swarm review (B1, fully productive), an adversarial review (B3, fully productive), and a ce-code-review leg (B2) that **wedged in a collaboration wait loop and produced only a fast-pass ("no urgent P0/P1 evident from changed hunks") — treat B2 as degraded**; the Codex engine is live overall via B1 and B3, and cross-engine corroboration below rests on those two legs. The headline finding was additionally ground-truth verified by the coordinator against the actual pinned binary and current Claude Code documentation.

One structural caveat that shaped the whole review: **the swarm path has never executed end-to-end.** Both trigger lanes (`issue_comment` and `workflow_dispatch`) run the default-branch copy of the workflow, so this PR's changes cannot be exercised by its own CI — green checks here validate workflow text, not runtime dispatch behavior.

---

## What holds up (validated, cross-engine)

These were active suspicions going in; validating them is a result, not filler:

- **Fork persona injection: refuted.** Personas are installed via `cp -a -- .claude/skills/gitnexus-review/ci-personas/. "${claude_config}/agents/"` (workflow:410-411) executed from the workspace root, which holds the trusted control-SHA checkout (asserted `git rev-parse HEAD == CONTROL_SHA` before any later step); the PR head is confined to `pr-target/` with symlink validation. A hostile fork cannot define, replace, or bias a lane. Verified independently by Claude ce-security, Claude security-boundary, Claude facts, and Codex B1. [code-read, cross-engine]
- **Sidechain evidence-gate exclusion is correct on both sides.** Registration (`block.name === CONTEXT_EVIDENCE_TOOL && !sidechain`, workflow:1702) and acceptance (`!sidechain && …`, workflow:1736) both gate; ids are globally deduplicated with fail-closed throws; a cross-wired transcript was **empirically executed** against the extracted validator and correctly yields `missing_graph_evidence`. No path lets sidechain activity satisfy the gate. Verified by Claude ce-correctness (mutation harness), ce-adversarial, ce-security, and Codex B1. [reproduced, cross-engine]
- **Mirror sync covers the new `ci-personas/` subdirectory.** `shipped-skills-sync.test.ts` recurses via `listFilesRecursive`/`snapshotDir` (no shallow glob), and `skill-sync.yml` path globs match the new files. The npm-package copy (`gitnexus/skills/`) and the Cursor copy are byte-identical to canonical; the plugin copy (`gitnexus-claude-plugin/skills/`) equals canonical + `mcp.json` only, which is the expected state per `shipped-skills-sync.test.ts:158-164`. The prior-session "sync test may glob only top-level" lead is refuted. [reproduced, cross-engine]
- **MCP runtime pinned:** `.github/gitnexus-review-runtime` pins `gitnexus@1.6.9`, not `@latest` — the stale concern from the #2566 review does not apply at head. [code-read]
- **Permissions, triggers, secrets unchanged.** Top-level `permissions: {}`, job-level read scopes, the comment-lane authorization chain, the OAuth token confinement to the analyze job, action SHA pins, and the sandbox settings are byte-identical to base. The only widenings are resource ceilings (timeout 45→75 min, max-turns 100→150). [code-read, cross-engine]
- **No bypass, escape, or injection found**: no bidi/hidden Unicode, no whitespace errors, the old double-slash deny-pattern no-op is gone, prompt-injection escalation is structurally closed (no Bash/Write/publish tools; publication is a separate job consuming an immutable artifact), and the parent-linkage regex (`TOOL_ID_RE`) cannot false-reject legitimate 2.1.214 transcripts. [code-read, cross-engine]

---

## Headline finding — P1: the swarm can never dispatch on the pinned runtime (Task vs Agent inversion)

**`.github/workflows/gitnexus-review-agent.yml:1216-1218`** · corroborated by Codex B1 (P1, across three of its personas) + Codex B3 (adversarial) + coordinator ground-truth verification. [reproduced]

On the pinned Claude Code **2.1.214**, the subagent-dispatch tool is named **`Agent`** — `Task` was renamed to `Agent` in 2.1.63 and survives only as a legacy alias — and permission precedence is **deny > allow**. The workflow configures:

- `1216: --tools "Read,Glob,Grep,Task"` — `Agent` absent from the available-tool set
- `1217: --allowedTools "Task,…"` — allows the legacy alias only
- `1218: --disallowedTools "Bash,Write,…,Skill,Agent,…"` — **denies the canonical dispatch tool**, and deny beats allow
- `1163` (prompt): "Use only Read/Glob/Grep/Task…" — also names the alias

Net effect: the orchestrator can never dispatch a lane. Per the skill's own fallback instruction ("If subagent dispatch is unavailable or any lane fails, run that lane's charge inline", SKILL.md:213-215), every CI review silently degrades to the pre-existing inline single-agent path. The swarm never runs, and nothing surfaces the degradation. The text-only tests (`gitnexus/test/unit/review-agent-workflow.test.ts:1147-1161`) assert exactly this broken split — `Task` ∈ allowed, `Agent` ∈ disallowed, `Task` ∉ disallowed — so CI pins the defect as correct.

Ground truth: a binary grep of the actual installed 2.1.214 (`@anthropic-ai/claude-code-linux-x64`) shows the dispatch tool registered as `Agent` (the `subagent_type` schema and subagent-spawn strings sit under the `Agent` marker); current Claude Code docs confirm the 2.1.63 rename, the deny-over-allow precedence, and that unknown names in allow lists are ignored with only a startup warning. The repo's own pre-existing `gitnexus-pr-swarm-review` skill dispatches via `Agent`.

Notable single-engine blind spot: every Claude review lane assumed `Task` is the dispatch tool and reasoned "Task allowed, Agent disallowed = correct posture" (the ce-project-standards lane recorded the alias-conflict possibility as a residual risk but deferred it). The Codex legs plus the ground-truth pass are what caught it — this is exactly the cross-engine failure mode this review format exists for.

**Fix:** use `Agent` consistently — add it to `--tools`, allow it **scoped to the six personas** (`Agent(ci-correctness-lens)`, `Agent(ci-security-lens)`, `Agent(ci-blast-radius-lens)`, `Agent(ci-coverage-lens)`, `Agent(ci-adversarial-lens)`, `Agent(ci-critic-lens)`), remove `Agent` from `--disallowedTools`, update the prompt at 1163 and the test assertions at 1147-1161. Scoping the allow to `Agent(ci-*)` also closes the add-dir agent-discovery concern below. Add a pinned-CLI smoke test that actually dispatches one installed persona — a text assertion cannot catch this bug class (see the test-gap section).

---

## Inline comments

Three findings have diff-line anchors, verified triggers, and evidence, and are posted inline:

- **`.github/workflows/gitnexus-review-agent.yml:1218`** — P1: Task/Agent inversion blocks all swarm dispatch on the pinned 2.1.214 runtime. `[reproduced via binary + docs + Codex B1/B3]`
- **`.github/workflows/gitnexus-review-agent.yml:1221`** — P2: `--max-turns` raised for the swarm but the 1,000-message / 8 MB transcript caps (1261-1262, hard throws) were not, and personas carry no per-agent `maxTurns`; a full six-lane run on a large PR can overflow and brick a valid review fail-closed (latent until the P1 is fixed). `[code-read; cross-lane corroborated]`
- **`gitnexus/test/unit/review-agent-workflow.test.ts:1394`** — P2: neither `!sidechain` guard is independently pinned — deleting either (workflow:1702 or 1736) alone still passes all 39 tests. `[reproduced — mutation-verified by two lanes]`

---

## Lower-priority findings (body only)

Severity-ordered; each verified against the worktree but either lower-impact, judgment-dependent, or lacking a clean diff anchor.

- **P2 — Rollout gate is currently open** (Codex B1, confirmed via the live repo variable): `GITNEXUS_REVIEW_COMMENT_ENABLED` is already `true`, yet the swarm path cannot be dry-run pre-merge (trigger lanes run the default-branch workflow), and DoD.md:137 requires merge-disabled → same-repo canary → fork canary → enable. As it stands, the first post-merge authorized review goes straight to production behavior with zero swarm executions ever observed. Recommend disabling the variable before merge and running both canaries via `workflow_dispatch` after the P1 fix lands.
- **P2 — Critic fails open by design; cross-engine severity split — a design choice to reconsider** (SKILL.md:208-215): after two critic passes, the skill instructs the orchestrator to proceed with objections unresolved. Codex B3 calls this a defect (the bot can post a review it knows is defective, and the repo's other swarm — `pr-swarm-review` lane 7 — is a hard gate); the Claude risk lane calls it correct-by-design (bounded, cannot deadlock, review still gated by evidence and schema checks). Both readings are internally consistent; presenting as a deliberate trade-off that should be made explicit — either document why fail-open is intended here, or align with the hard-gate contract by stripping critic-challenged findings before publication.
- **P2 — The evidence gate now hangs on one prompted sentence** (Claude ce-adversarial, ~0.55): with all graph work delegated to lanes, the only thing forcing the orchestrator's own mainline `context` call is prose (workflow prompt ~1189-1193 + SKILL.md). If the model delegates everything and never makes its own call, `missing_graph_evidence` bricks a genuinely complete review — fail-closed, but user-visible on a live agent. Cheap hardening: have the skill require the orchestrator's own `context` call on a changed symbol *before* lane dispatch, and add a fully-delegated-transcript fixture.
- **P2 (uncertain) — add-dir agent-discovery gap** (Claude security-boundary): the `--add-dir` PR-target contains the repo's own `.claude/agents/` (seven pr-swarm agents). `CLAUDE_CODE_ADDITIONAL_DIRECTORIES_CLAUDE_MD: '0'` suppresses CLAUDE.md pickup, but no analogous agents-suppression knob is set; whether 2.1.214 can discover agents from an additional directory under `--setting-sources user` was not conclusively established. Bounded (any such agent still runs under the session deny list; no code exec), and **scoping the allow to `Agent(ci-*)` in the P1 fix closes it structurally** — one more reason to prefer the scoped form.
- **P3 — Docs not updated for a shipped capability change** (Claude docs-dod + ce-project-standards): AGENTS.md (changelog stuck at 1.13.0, gitnexus-review description omits swarm/ci-personas) and CLAUDE.md are untouched; every prior functional change to the skill family bumped them. Also `.claude/README-gitnexus-reviewer-swarm.md`'s differentiator ("fixed-roster swarm vs per-domain lenses") is now stale since gitnexus-review ships its own fixed six-lane roster.
- **P3 — PR body understates the timeout change** (Claude facts + Codex B1): the description says 45→60 minutes; the workflow says `timeout-minutes: 75` (raised 60→75 in the head commit). Documentation accuracy only.
- **P3 — Glob/Grep in `--tools` but not `--allowedTools`** (pre-existing posture, noted by three lanes with mild disagreement): under `dontAsk`, lane Glob/Grep calls may be auto-denied while every persona's frontmatter declares them. Either allow them explicitly or drop them from persona frontmatter, so denied-tool error results are not manufactured inside lane transcripts.
- **P3 — "Swarm" naming collision** (Claude ce-maintainability): the repo now has two unrelated "coordinated reviewer swarm" systems — pre-existing `gitnexus-pr-swarm-review` (numbered personas, `Agent` dispatch, hard critic gate) and this PR's `ci-personas` lanes — with near-identical vocabulary and no cross-reference. One disambiguating sentence in each SKILL.md would spare future maintainers. Relatedly, the new fixed five-lane catalog sits inside the pre-existing domain-scaled "Expert lenses" section without stating how they compose.
- **P3 — Merge-base checkout not readable by lanes** (Claude risk + ce-security): the raw merge-base source checkout at `${runner.temp}/gitnexus-review-merge-base` is not covered by `--add-dir` (only `gitnexus-review-pr-target` is) and is not in the working directory, so it is inaccessible to lanes' `Read` under `dontAsk`. Fail-closed (lanes degrade to MCP queries); a lane-efficacy gap for deleted-file analysis, not a security issue.
- **P3 — Local-harness persona registration undocumented** (Claude docs-dod): SKILL.md says "a local harness may register them from this directory" but neither `gitnexus setup` nor any doc provides a concrete install path.
- **P3 (cosmetic) — branch name** `fix/review-agent-sonnet-5` carries feature work; PR title already clarifies.

---

## Why the P1 slipped through: the test gap

The two defects share one root cause: **the workflow tests assert only text, never behavior.** `review-agent-workflow.test.ts` greps the YAML for flag strings — it never invokes the pinned CLI, so a permission config that can never dispatch a subagent passes identically to one that can (and the assertions at 1147-1161 actively pin the inverted Task/Agent split). Symmetrically, the sidechain guards are only property-tested as a pair: deleting either `!sidechain` guard alone (workflow:1702 or 1736) leaves all 39 tests green — mutation-verified independently by two Claude lanes using two different harnesses, corroborated by the swarm test-CI lane's code-read. Recommended additions: (1) a pinned-CLI dispatch smoke test that installs one persona into a throwaway `CLAUDE_CONFIG_DIR` and proves it can spawn (and that an unlisted agent type cannot); (2) the one-line cross-wired sidechain fixture from inline comment 3; (3) a transcript-cap boundary test with sidechain-inflated volume; (4) a small persona-contract test (frontmatter `name:` ↔ filename, `tools:` ⊆ workflow allowlist, SKILL.md lane names ⊆ `ci-personas/` stems).

## CI status

All checks green at head (actionlint, zizmor, gitleaks, CodeQL, typecheck, unit tests including the new sidechain test at 39/39, shipped-skills drift guard at 52/52) — but green is insufficient here: CI validates workflow text and mirror byte-identity, not runtime dispatch, and the changed workflow itself only executes post-merge from the default branch.

---

**Coverage note:** the reviewed head was `0ef716cc`; the PR head then advanced to `6a297f58` via a merge of `main` that pulled in the already-merged #2571 (skill-evolution) and touched **none** of the reviewed swarm files (the review-agent workflow, `ci-personas/`, SKILL.md, and the workflow test are byte-identical at both SHAs), so every finding and line anchor below holds at the current head. This digest merges thirteen Claude lane reports (six swarm lanes + seven ce personas), two fully productive Codex legs (swarm B1, adversarial B3), one degraded Codex leg (ce B2, fast-pass only), and a coordinator ground-truth pass against the pinned 2.1.214 binary and current Claude Code documentation. Findings are weighted by cross-engine corroboration and reproduction; single-lane items are labeled as such. Refuted suspicions are reported above rather than dropped, since ruling them out was part of the review's value.

*This is an automated multi-tool review digest — verify findings before acting on them.*

PR #2575, opened 2026-07-20
fix(ci): install root and shared node_modules for the evolution benchmark

abhigyanpatwari:
## What

The first real `workflow_dispatch` of the skill-evolution loop ([run 29723409846](https://github.com/abhigyanpatwari/GitNexus/actions/runs/29723409846)) failed at **task binding**, before any agent ran:

```
workflow_bench.proposer_sandbox.SandboxError: sandbox_copy path is unavailable: node_modules: [Errno 2] No such file or directory: 'node_modules'
```

## Why

The benchmark tasks sandbox-copy `node_modules` from three locations (`eval/workflow_bench/tasks.scenarios.yaml`), mirroring a full dev checkout: the monorepo **root**, `gitnexus-shared`, and `gitnexus`. This is a non-workspaces monorepo — each installs independently. The workflow already builds the two subpackage trees in their own steps ("Build pinned shared runtime", "Install and build pinned GitNexus runtime"), but never installed the **root** tree, so `capture_task_dependency_binding` aborted on the first (root) dependency. Steps 1–9 passed, including the `~/GitNexus` task-repo symlink from #2571 — this is the next layer down.

## How

Add a single "Install monorepo root dependencies" step (`npm ci` at the repo root); the two subpackage build steps are unchanged. A contract test in `skill-evolution-workflow.test.ts` pins all three installs — the same "fail in CI, not on the next real run" guard the workflow's two prior P1 fixes got.

## Testing

`npx vitest run test/unit/skill-evolution-workflow.test.ts` → 10 passed. End-to-end verification is the next `workflow_dispatch` after merge (the workflow only runs from `main`); it bills real API usage on `GITNEXUS_BENCH_AUTH_TOKEN`.

Re-dispatch the workflow after merge and confirm the loop gets **past task binding** into the benchmark (no `SandboxError`); watch the "Run the propose → benchmark → gate loop" step.


magyargergo:
@gitnexus review

magyargergo:
@gitnexus review

PR #2582, opened 2026-07-20
perf(lbug): size the buffer pool to the repo (adaptive, COPY-safe)

magyargergo:
## What

Size the LadybugDB buffer pool to the repo being analyzed instead of always allocating the fixed `min(2 GiB, 80%×RAM)` default. LadybugDB **eagerly commits** the buffer pool at DB open, so the oversized default wastes wall-time on every analyze — most visibly on large repos.

**Measured (GitNexus checkout itself, full `analyze --force`):**

| Buffer pool | Wall time |
|---|---|
| 2 GiB (current default) | **50.8s** |
| adaptive (~414 MiB here) | **35.3s** |

→ **31% faster** large-repo analyze, and every repo uses less memory.

## How

- `resolveBufferManagerSize()` gains a precedence chain: `GITNEXUS_LBUG_BUFFER_POOL_SIZE` env override (unchanged, still wins, `0` = native default) → a per-run **size hint** clamped to `[ADAPTIVE_POOL_FLOOR, default]` → the default. The hint can only *shrink* the pool from today's default, never grow past the 2 GiB / 80%-RAM cap.
- `runFullAnalysis` sets the hint from the built graph's `nodeCount + relationshipCount` (after the pipeline, before the DB open) and clears it at the top of each run. The pool is a page cache over the on-disk index, which scales with graph size.
- The DB-open path (`openLbugConnection`) is a 16-caller hub, so the hint is module-scoped rather than threaded through it — the diff stays confined to `lbug-config.ts` + one call site in `run-analyze.ts`.

## The COPY-safety floor (why `ADAPTIVE_POOL_FLOOR = 256 MiB`, not 64 MiB)

LadybugDB's bulk `COPY` needs working buffer-pool memory that scales with the repo. A pool that's too small fails with `Buffer manager exception: the buffer pool is full and no memory could be freed`. Empirically: a 6-file fixture needs ≥128 MiB; the ~1800-file GitNexus checkout needs ≥256 MiB. So the adaptive size never drops below a distinct 256 MiB floor (kept separate from `BUFFER_POOL_FLOOR = 64 MiB`, which only guards the default on tiny-RAM machines; the hint is still clamped up to the machine default, so it can never over-commit).

## Scope / honest notes

- This is a **large-repo optimization**. Small repos now open COPY-safely at 256 MiB instead of 2 GiB — same wall time on Linux (lazy overcommit), and the far cheaper eager commit should help on Windows, though that isn't verified here.
- Read-only opens (MCP serve, standalone `detect_changes`, `native-check :memory:`) don't set the hint and keep the default — unchanged. Sizing those from the on-disk index size is a possible follow-up.
- The `docs/plans/` plan started from a "small repos get a tiny fast pool" premise; implementation showed COPY needs ≥256 MiB, so the floor and framing were corrected (commit `fix(lbug): raise the adaptive pool floor to a COPY-safe 256 MiB`).

## Verification

- Full unit suite green: **503 files, 10327 passed**; new `lbug-config-wal` tests cover the precedence, clamping, and estimate boundaries.
- `skills-e2e` Idempotency integration test passes with **no env override** (previously failed `COPY: buffer pool is full` under the too-small floor).
- Large-repo benchmark above; env override + `0`-escape-hatch preserved.

### DIFF
--- a/gitnexus/src/core/lbug/lbug-config.ts
+++ b/gitnexus/src/core/lbug/lbug-config.ts
@@ -321,6 +321,16 @@ const resolveCheckpointThreshold = (): number => {
 const DEFAULT_BUFFER_POOL_CAP = 2 * 1024 * 1024 * 1024;
 const BUFFER_POOL_FLOOR = 64 * 1024 * 1024;
 
+// COPY-safety floor for the adaptive hint (below). LadybugDB's bulk COPY needs
+// working buffer-pool memory that scales with the repo: a 64 MiB pool fails
+// ("buffer pool is full and no memory could be freed") on any non-trivial repo,
+// and even the ~1800-file GitNexus checkout needs ≥256 MiB. So the adaptive
+// size never drops a repo below this — a distinct, higher floor than
+// BUFFER_POOL_FLOOR, which only guards defaultBufferPoolSize on tiny-RAM
+// machines. It is still clamped up to defaultBufferPoolSize, so a machine whose
+// default is below this floor keeps its default rather than over-committing.
+const ADAPTIVE_POOL_FLOOR = 256 * 1024 * 1024;
+
 const parseBufferPoolSize = (raw: string | undefined): number | undefined => {
   if (raw === undefined) return undefined;
   const normalized = raw.trim();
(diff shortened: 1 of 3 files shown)


magyargergo:
@gitnexus review

PR #2584, opened 2026-07-20
fix(eval): stop using --bare for arms that need Skill or MCP tools

magyargergo:
## Summary

The last skill-evolution run ([run 29754004954](https://github.com/abhigyanpatwari/GitNexus/actions/runs/29754004954/job/88391579361)) scored 0/3 resolution on both the incumbent and candidate arms across every task, with `error_kind: "skill-not-invoked"` on all 18 benchmark sessions. That looked like a candidate-quality result but wasn't: the benchmark harness could never invoke either arm's skill at all.

Root cause, confirmed directly against the pinned `claude-code@2.1.214` binary: **`--bare` hard-disables the `Skill` tool and every `mcp__*` tool by Claude Code design** — `--allowedTools` cannot restore what `--bare` removes (this is documented behavior, not a bug; a GitHub feature request asking for it was closed "Not planned"). Every `workflow_bench` arm except `baseline_nomcp` needs `Skill` and/or the GitNexus MCP tools (`gitnexus-plan`, `gitnexus-work`, `gitnexus-review`, the CE comparator skills, and even plain `baseline`'s read-only graph tools), so those sessions have silently never been able to invoke their skill since `--bare` was adopted for auth isolation.

- Only `baseline_nomcp` keeps `--bare` — it's the one arm that explicitly wants zero Skill/MCP access anyway.
- Every other arm drops `--bare` and relies on `ANTHROPIC_API_KEY` alone. The sandboxed `HOME` has no OAuth/keychain state to conflict with it, and there's no committed `.claude/settings.json` in this repo for dropping `--bare` to newly pick up.
- Outside `--bare` the built-in toolset defaults to everything (WebFetch, Task, subagents, ...), and `--allowedTools` only pre-approves within whatever's available — it doesn't narrow it. Added `--tools` for non-bare sessions so the intended per-arm scope is still enforced instead of silently widening.

## Test plan

- [x] `uv run pytest eval/tests/ -q` — 306 passed, 9 skipped, 0 failures
- [x] `uv run ruff check` on changed files — clean
- [x] Verified `--tools`/`--allowedTools`/skill auto-discovery behavior directly against the pinned binary (space- and comma-separated forms, MCP tool names silently ignored by `--tools`, project-local `.claude/skills/*` auto-discovers without `--bare` and without `--add-dir`)
- [ ] Live CI run of `gitnexus-skill-evolution` to confirm `ANTHROPIC_API_KEY` auth still works cleanly without `--bare` and that `Skill`/MCP tool calls aren't blocked by `build_claude_settings()`'s `permissions.allow` list (currently `Read/Grep/Glob/Bash` only — untested whether `--permission-mode dontAsk` covers the newly-available tools too, though it evidently already covers `Edit`, which also isn't allow-listed, based on real diffs produced in past `--bare` runs)

### DIFF
--- a/eval/tests/test_workflow_bench_sessions.py
+++ b/eval/tests/test_workflow_bench_sessions.py
@@ -167,6 +167,56 @@ def test_run_claude_forwards_the_named_model_to_every_session(monkeypatch, tmp_p
     assert captured[captured.index("--model") + 1] == "claude-sonnet-4-20250514"
 
 
+def test_run_claude_restricts_tools_via_tools_flag_outside_bare(monkeypatch, tmp_path):
+    # Outside --bare, the built-in toolset defaults to everything (subagents,
+    # WebFetch, Task, ...) and --allowedTools only pre-approves within that —
+    # it does not narrow it. --tools is what actually restricts the set, so a
+    # non-bare arm session must pass it or it silently gets a far wider
+    # toolset than intended.
+    captured: list[str] = []
+
+    def fake_run(command, **kwargs):
+        captured.extend(command)
+        return fake_cli_result(VALID_REPORT)
+
+    monkeypatch.setattr(runner_sessions, "run_managed", fake_run)
+    runner.run_claude(
+        "task",
+        tmp_path,
+        claude_bin="claude",
+        timeout=5,
+        bare=False,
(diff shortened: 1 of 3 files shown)


magyargergo:
@gitnexus review

PR #2597, opened 2026-07-21
fix(java): record container node + new-expression chained call receiver typing (#2564)

### LINKED ISSUE #2564: graph: dangling Record HAS_METHOD edges + missing method edge on new-expression chained calls (pre-existing, lens-observed)
## Context

Two pre-existing graph gaps observed (and A/B-proven unrelated to the changes under review) by the empirical review lenses on PRs #2549 and #2558. Filing so they stop being rediscovered; they are independent and can be split if picked up separately.

## 1. Record container nodes never materialize → dangling `Record:` HAS_METHOD edges

A plain Java record with methods — no anonymous classes involved — emits `HAS_METHOD` edges whose source id is a `Record:...` node that does not exist in the node set (probe R1 in the #2549 review: the dangling edges reproduce identically on a record-only fixture at the pre-PR baseline). Either the record container node should be emitted (record_declaration is in `javaClassConfig.typeDeclarationNodes`, so the drop is downstream of extraction) or the owner edge should key to a node that exists.

## 2. Chained call on a new-expression emits only the constructor edge

```java
void m() {
    class Local { void inner() {} }
    new Local().inner();   // CALLS: only the constructor edge; no edge to Local.inner
}
```

Probe S5 (#2549 review): `new Local().inner()` produces the ctor CALLS edge but no method edge — the compound-receiver machinery does not treat a new-expression as a typed receiver for the immediate chained call. Pre-existing; also reproduces with top-level classes.

Both observations have archived probe outputs referenced in the #2549/#2558 review reports.

magyargergo:
## Summary

Fixes #2564 — two independent, pre-existing bugs in the Java ingestion path found by review lenses on PRs #2549 and #2558.

- **Dangling `Record:` `HAS_METHOD` edges.** `JAVA_QUERIES` had no `@definition.record` capture for `record_declaration` — unlike its `class_declaration`/`interface_declaration`/`enum_declaration` siblings, and unlike C#'s own `record_declaration` pattern in the same file. A Java record's container node was never created, so its methods existed as ownerless `Method` nodes with no `HAS_METHOD` edge, even though ownership resolution already computed a valid `ownerId` for them.
- **Wrong-target resolution on `new Local().inner()`.** The receiver capture for a chained call binds the *entire* `object_creation_expression` node, so its raw text (`"new Local()"`) became the receiver name — a string that can never match a scope binding. The call silently fell through to name-only fallback resolution and could resolve to an unrelated same-named method on a collision. Fixed by normalizing the receiver to the constructed type's simple name in `java/captures.ts`, mirroring the existing `normalizePhpReceiver` precedent — a language-local capture rewrite, no shared-pipeline change.
- **`INCREMENTAL_SCHEMA_VERSION` bump.** The record fix makes previously-uncaptured `Record` nodes and `HAS_METHOD` edges appear for the first time, but the incremental write set only covers changed files. Bumped 9→10 so an existing index gets a one-time full re-analyze instead of silently continuing to omit these for unchanged record files, matching the v7/v8/v9 precedent for the same class of change (#2437/#2522, #2550, #2555).

## Test plan

- [x] New fixture `java-record-methods/Point.java` + `java.test.ts` assertions: `Record` node exists for `Point`, `HAS_METHOD` edges to `sum`/`scaled`, internal `scaled → sum` call resolves
- [x] New fixture `java-new-expr-chain-call/{LocalChain,Other}.java` + assertions: `new Local().inner()` resolves to the local `Local#inner`, not the unrelated same-named `Other#inner`
- [x] Both fixes verified independently (isolated via `git stash` to confirm each closes its gap without the other)
- [x] Full `java.test.ts` (239/239) and `java-1928.test.ts` (5/5) pass
- [x] `npx tsc --noEmit` clean for `gitnexus/` and `gitnexus-shared/`
- [x] `eslint` clean on changed source files
- [x] All 5 test files referencing `INCREMENTAL_SCHEMA_VERSION` pass (73/73) after the bump


magyargergo:
@gitnexus review

PR #2598, opened 2026-07-21
fix(analyze): drop FTS indexes before the incremental DETACH DELETE

### LINKED ISSUE #2589: [Bug] Incremental analyze fails with "FTS index 'file_fts' is inconsistent: term '...' is missing during delete" after markdown-only commits; --repair-fts also fails in that state
## Summary

Incremental `gitnexus analyze` intermittently fails with an FTS consistency error when re-indexing after commits that modify markdown files. Happened twice in one day on the same repository, each time with a different term. While the index is in this state, `analyze --repair-fts` also exits with an error; a subsequent plain `analyze` run recovers the index.

## Environment

- GitNexus CLI: `1.6.10-rc.47` (global npm install, invoked via `.gitnexus/run.cjs`)
- Node.js: `v24.14.0`
- OS: Linux (WSL2, kernel `6.6.87.2-microsoft-standard-WSL2`)
- Graph/FTS provider (from `.gitnexus/gitnexus.json`): `ladybugdb` / `ladybugdb-fts`, `schemaVersion: 8`
- Repository: private monorepo, ~1,860 indexed files, TypeScript + Markdown docs (docs contain non-ASCII/Cyrillic text)

## Reproduction (as observed; no minimal repro yet)

1. Repository indexed and up to date.
2. Make a commit touching only markdown files (docs), then run incremental re-index:

   ```bash
   node .gitnexus/run.cjs analyze --no-stats
   ```

3. Analysis fails:

   ```text
   Analysis failed: Runtime exception: FTS index 'file_fts' is inconsistent: term 'wiki' is missing during delete. Drop and recreate the FTS index.
   Error: Runtime exception: FTS index 'file_fts' is inconsistent: term 'wiki' is missing during delete. Drop and recreate the FTS index.
   ```

The second occurrence (same day, after another markdown-only commit) produced the identical error with a different term — a non-ASCII (Cyrillic) word from the docs. So the failing term is not charset-specific: both a plain-ASCII term (`wiki`) and a Cyrillic term triggered it.

## `--repair-fts` fails while the index is inconsistent

With the index in the broken state:

```bash
node .gitnexus/run.cjs analyze --repair-fts
```

exits with an error as well (stack trace through `analyzeCommandImpl` at `dist/cli/analyze.js:991` → `analyzeCommand` at `analyze.js:509`; unfortunately the full message was not captured).

On a **healthy** index the same command succeeds (`FTS indexes repaired successfully`), so the repair path appears to fail exactly in the situation it exists for.

## Workaround / recovery

Running plain `analyze` again right after the failed `--repair-fts` completes successfully and fully restores FTS (BM25 search verified working afterwards). Recovery sequence both times:

```bash
node .gitnexus/run.cjs analyze --repair-fts   # errors
node .gitnexus/run.cjs analyze --no-stats     # succeeds, FTS healthy again
```

## Expected

- Incremental analyze keeps `file_fts` consistent when files are modified/deleted, or transparently drops and rebuilds the FTS index instead of failing the whole analysis.
- `analyze --repair-fts` should be able to repair an inconsistent index (its primary use case), not only run on a healthy one.

## Notes

- Both failures followed commits that touched only markdown files.
- No multi-branch indexing, no `--pdg`, no embeddings; workspace index only.
- Happy to run diagnostics or a debug build if that helps.

magyargergo:
## Summary

Fixes #2589: incremental `gitnexus analyze` intermittently crashed with `Runtime exception: FTS index 'file_fts' is inconsistent: term 'X' is missing during delete` after commits that touched only markdown files. `analyze --repair-fts` also failed while the index was in this state; only a subsequent full `analyze` recovered (via the unrelated full-rebuild escalation path).

**Root cause:** the incremental writeback (`runFullAnalysis`, `run-analyze.ts`) ran `deleteNodesForFiles`'s batched `DETACH DELETE` against tables that still carried the FTS index built at the end of the *previous* analyze run — `createSearchFTSIndexes` only drops+rebuilds every FTS index in Phase 3, well after that delete already ran. LadybugDB's FTS extension is not proven to survive DML against an indexed table (its own docs never demonstrate the sequence).

**Fix:**
- Extract `dropSearchFTSIndexes()` (new) in `fts-indexes.ts` and call it up front in the incremental writeback's non-escalated branch, before `deleteNodesForFiles` — `createSearchFTSIndexes` still does its own unchanged drop+create pass in Phase 3.
- Harden `dropFTSIndex` (`lbug-adapter.ts`): it previously swallowed *every* `DROP_FTS_INDEX` error unconditionally. It now only tolerates the two benign "nothing to drop" cases (Binder/Catalog exceptions, verified end-to-end against `@ladybugdb/core` 0.18.x's real error text) and rethrows anything else — so a genuinely corrupted index can no longer persist across analyze runs undetected.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npm run build`
- [x] New: `test/unit/drop-fts-index-error-classification.test.ts` — pure classifier unit tests (including the exact reported error text) + a real-DB integration case
- [x] New: `test/unit/incremental-fts-drop-ordering.test.ts` — drives a real `runFullAnalysis` full+incremental cycle and asserts `SHOW_INDEXES()` reports every FTS index absent by the time `deleteNodesForFiles` runs; manually verified this test fails without the fix (indexes still present) and passes with it
- [x] Full targeted sweep: 14 test files / 89 tests across FTS + incremental-orchestration coverage, all green

### DIFF
--- a/gitnexus/src/core/lbug/lbug-adapter.ts
+++ b/gitnexus/src/core/lbug/lbug-adapter.ts
@@ -2904,7 +2904,26 @@ export const queryFTS = async (
 };
 
 /**
- * Drop an FTS index
+ * True for the two benign "nothing to drop" `DROP_FTS_INDEX` failures —
+ * both catalog/binder exceptions, LadybugDB's classes for "this name isn't
+ * bound to anything right now" (probe-verified end-to-end through
+ * `dropFTSIndex`'s real `conn.query()` path against @ladybugdb/core
+ * 0.18.x): the named index was never created (`Binder exception: Table <T>
+ * doesn't have an index with name <name>.`), or the FTS extension/function
+ * isn't registered at all (`Catalog exception: function DROP_FTS_INDEX is
+ * not defined...`). A real engine failure — e.g. the `Runtime exception:
+ * FTS index '<name>' is inconsistent: ...` class from #2589 — is a
+ * DIFFERENT exception class (an execution-time failure, not a catalog/bind
+ * lookup miss), so this returns false for it. Pure string logic so it is
+ * unit-testable without a native LadybugDB connection.
+ */
+export const isBenignDropFtsIndexError = (message: string): boolean =>
+  message.includes('Binder exception:') || message.includes('Catalog exception:');
+
+/**
+ * Drop an FTS index. Tolerates only {@link isBenignDropFtsIndexError} —
+ * anything else rethrows instead of being silently masked, which previously
+ * let a corrupted index persist across analyze runs undetected.
  */
 export const dropFTSIndex = async (tableName: string, indexName: string): Promise<void> => {
   if (!conn) {
@@ -2913,8 +2932,11 @@ export const dropFTSIndex = async (tableName: string, indexName: string): Promis
 
   try {
     await queryAndDrain(conn, `CALL DROP_FTS_INDEX('${tableName}', '${indexName}')`);
-  } catch {
-    // Index may not exist
+  } catch (e: unknown) {
+    const msg = e instanceof Error ? e.message : String(e);
+    if (!isBenignDropFtsIndexError(msg)) {
+      throw e;
+    }
   } finally {
     ensuredFTSIndexes.delete(ftsIndexKey(tableName, indexName));
   }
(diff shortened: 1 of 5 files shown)


magyargergo:
@gitnexus review

abhigyanpatwari:
@gitnexus review

PR #2600, opened 2026-07-21
fix(eval): self-hosted skill-evolution runner + sandbox Python 3 trust fix

magyargergo:
## Summary

- Moves `gitnexus-skill-evolution.yml` off `ubuntu-latest` onto a self-hosted runner (`runs-on: [self-hosted, linux, x64, gitnexus-evolution]`, `timeout-minutes: 1440`). GitHub-hosted runners hard-cap job execution at 6 hours, which is too short now that benchmark sessions actually invoke Skill/MCP tools for real (post the `--bare` fix in #2584). Self-hosted caps at 5 days instead. Documents the runner-registration step in the workflow's own activation checklist.
- Fixes a real bug the self-hosted validation run surfaced: `gitnexus-plan` sessions inside the bwrap sandbox failed with `planning must create or modify exactly one plan artifact; observed 0`. Root cause: `evidence-provenance.mjs`'s atomic plan-writer only trusts a Python 3 binary owned by root or by the current process. Inside a `--unshare-user` sandbox where root isn't mapped, the real (root-owned) `/usr/bin/python3` surfaces as the kernel's overflow uid and gets correctly refused as untrusted. Fix: provision a small, self-owned wrapper script in `proposer_sandbox.py` (same pattern already used for `shell-prefix`) that execs the real interpreter, so the sandbox has a Python 3 candidate the existing trust check can actually accept, without touching that security-sensitive validation logic.
- Adds visibility so this class of failure isn't quiet next time: `report.md` now shows *why* each row failed (`error_kinds`), not just resolved 0/1. The benchmark now exits non-zero when an incumbent arm resolves zero across every task, since that reads as a broken harness rather than a normal candidate miss (a candidate merely underperforming is expected and does not trip this).

## Test plan

- [x] `uv run --locked --extra dev python -m pytest tests -q` — 310 passed, 10 skipped (pre-existing, real-bwrap-gated)
- [x] `ruff check` clean on all touched files
- [x] Dispatched a real `workflow_dispatch` validation run against the self-hosted runner before this fix, confirming the failure this PR addresses; end-to-end re-validation with the fix applied is a natural next step post-merge

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,5 @@
+# Custom self-hosted runner labels actionlint can't discover on its own.
+# gitnexus-evolution: the skill-evolution EC2 runner (infra/gitnexus-evolution/).
+self-hosted-runner:
+  labels:
+    - gitnexus-evolution
--- a/.github/workflows/gitnexus-skill-evolution.yml
+++ b/.github/workflows/gitnexus-skill-evolution.yml
@@ -18,6 +18,15 @@
 #     workflow (and eval/workflow_bench/evolve.py) from the *dispatched ref*,
 #     so this server-side rule — not a code-side guard the branch could edit
 #     away — is what stops a non-main branch from running with the secrets.
+# [ ] Register a self-hosted runner labeled `gitnexus-evolution` (a dedicated
+#     EC2 box works well). GitHub-hosted runners hard-cap job execution at 6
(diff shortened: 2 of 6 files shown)


magyargergo:
@gitnexus review

magyargergo:
@gitnexus review

PR #2602, opened 2026-07-21
fix(java): resolve E.CONST.method() enum-constant receiver dispatch (#2561)

### LINKED ISSUE #2561: scope-resolution: enum-constant receiver dispatch — E.CONST.method() emits no CALLS edge
## Context

Follow-up to #2555 / PR #2558. During that PR's review, an empirical A/B probe established that calling a method **on an enum constant receiver** emits no CALLS edge at all:

```java
enum M4 {
    C { public void hook() { ... } };   // bodied constant — hook lives on M4$1
    public abstract void hook();
}

class User {
    void go() {
        M4.C.hook();   // → no CALLS edge (neither to M4$1.hook nor M4.hook)
    }
}
```

The probe's body-less control (`Plain.A.m()` where `m` is a plain enum method) **also** emits nothing on the pre-#2555 baseline — so this is a pre-existing receiver-machinery gap, not a regression from the instance model.

## Why it is newly actionable

Before #2555 there was nothing correct to dispatch to: a bodied constant's override had no node of its own. Now `M4.C`'s body is a first-class `Class M4$1` with `EXTENDS M4$1 → M4`, owned re-keyed methods, and an MRO. The missing piece is receiver-side: the receiver `M4.C` (a `field_access` on the enum type) needs to resolve to the constant's synthesized class (bodied constants) or the enum itself (body-less), after which the existing Case-4 chain walk (`ownerDef → mroFor`) does the rest — the constant-name → `E$N` mapping is derivable from the same source-order counter the naming authority uses.

## Suggested shape

- Bind `E.CONST` receivers: bodied constant → `E$N` def; body-less → the enum def (so `Plain.A.m()` finally resolves too).
- Tests: bodied-override dispatch (`M4.C.hook()` → `M4$1.hook#0`), body-less inherited dispatch (`Plain.A.m()` → `Plain.m#0`), and the existing enum-static describe (java.test.ts "enum static method call resolution") as the regression gate.

magyargergo:
## Summary

Fixes #2561 — calling a method on an **enum-constant receiver** (`E.CONST.method()`) emitted no `CALLS` edge.

`E.CONST` is a two-segment compound receiver. `resolveCompoundReceiverClass` resolves it by walking each dotted segment through the owning class scope's `typeBindings` map — but enum constants had no `typeBinding`, so the constant segment dead-ended and no target was ever resolved. #2555/#2558 gave bodied constants a first-class synthesized `E$N` class with an MRO that includes the host enum; this is the receiver-side follow-up they deliberately left open.

## Fix

`synthesizeJavaAnonymousClassDeclarations` (`gitnexus/src/core/ingestion/languages/java/captures.ts`) now emits a class-scope `typeBinding` for every enum constant's simple name:

- **bodied constant** → its synthesized `E$N` class (whose MRO includes `E`, so both the override and inherited enum helpers resolve);
- **body-less constant** → the host enum `E` itself (so `Plain.A.m()` finally resolves too).

This reuses the exact mechanism a field declaration already uses (`private User user;` binds `user` on the class scope). The generic compound-receiver chain walk then resolves `E.CONST.method()` with **no change to any shared scope-resolution code** — the fix is confined to the Java capture layer. The binding lives on the enum's own class scope, so it is reachable from inside the enum (unqualified sibling access, valid Java) and via a qualified `E.CONST` receiver from outside, but does not leak into unrelated scopes.

## Tests

New `describe('Java enum-constant receiver dispatch (#2561)')` block in `gitnexus/test/integration/resolvers/java.test.ts`, backed by fixture additions (`EnumConst.java` + new `Plain.java`):

- bodied: `EnumConst.A.hook()` → `Method:src/EnumConst.java:EnumConst$1.hook#0`
- body-less: `Plain.A.m()` → `Method:src/Plain.java:Plain.m#0`

Both were verified to **fail against the pre-fix tree** and pass after the fix (genuine discriminators).

## Verification

- Full integration suite: **4380 passed / 0 failed**
- `java.test.ts`: **241 passed** (includes the #2555 and enum-static regression blocks, unchanged)
- `tsc --noEmit` clean · eslint 0 errors · changed files prettier-clean


koriyoshi2041 (approved):
Verified exact head `d9437e6d` locally after refreshing the lockfile-backed install:

- `npm test -- test/integration/resolvers/java.test.ts --reporter=dot` — 241/241 passed
- `npx tsc --noEmit` — passed
- `git diff --check origin/main...HEAD` — passed

The capture-layer binding is appropriately narrow: body-less constants bind to the host enum, while bodied constants bind to the synthesized class whose MRO already reaches the host. The two regressions exercise both branches and assert exact target IDs. I don't see a blocking issue in this scope.

magyargergo:
@gitnexus review

magyargergo:
Addressed both findings from the review agent in `70e0a776`:

- **MEDIUM (untested inherited-dispatch MRO path):** added a third assertion `EnumConst.A.log()` → `Method:src/EnumConst.java:EnumConst.log#0`. `log()` is defined only on the host enum and not overridden by constant `A`, so this exercises `EnumConst$1`'s `@reference.inherits` MRO arm reaching the host enum's own method through the qualified receiver — the exact capability the comment claimed. Passes end to end.

- **LOW (bodied-constant fail-safe):** replaced `constantType = bodiedName ?? hostEnum` with `isBodied ? bodiedName : hostEnum`. A bodied constant now binds **only** to its `E$N` class; if name synthesis fails on a malformed/error-recovery tree, it emits no type-binding rather than silently misattributing an overriding constant's receiver to the host enum — mirroring the `object_creation_expression` branch's skip-on-synthesis-failure. Verified output-neutral on the well-formed bench corpus (fingerprint identical with and without the change).

The java scope-capture fingerprint was rebaselined again (`a822cef9` → `d04298a9`); the drift is purely the new `dispatchInherited` fixture method (this bench's corpus **is** `test/fixtures/lang-resolution`), confirmed by isolating the fixture-only fingerprint from the logic change.

Verification: `java.test.ts` 242 passed (3 `#2561` tests incl. the new inherited-dispatch case), `measure.mjs --check` PASS (14 languages), `tsc --noEmit` / prettier / eslint clean.

@gitnexus review

magyargergo:
@gitnexus review

PR #2603, opened 2026-07-21
feat(infra): codify skill-evolution runner + GitHub environment as Terraform

magyargergo:
## Summary

- Codifies the AWS side of the `gitnexus-skill-evolution` self-hosted runner (EC2 instance, IAM role/instance profile for SSM, security group, EventBridge start/stop schedules matching the workflow's own Saturday cron) and the GitHub side (the `gitnexus-evolution` environment and its main-only deployment branch policy) as Terraform under `infra/gitnexus-evolution/`.
- These resources were originally created by hand across a live session -- not reproducible, reviewable, or documented anywhere but shell history. This PR imports all ten of them into Terraform state so they're adopted rather than duplicated.
- `terraform plan` against the imported state came back clean after two real fixes the import surfaced:
  - `associate_public_ip_address` read back as `false` on import (a known AWS provider limitation -- it isn't directly exposed by the API), which would have forced a destructive replace of the already-registered runner to "fix" a value that was never actually wrong. Handled via `lifecycle.ignore_changes`.
  - The security group's egress rule carried a `description` field the hand-created rule didn't have, showing as a spurious delete+recreate of the rule. Removed the field to match reality exactly.
- Secret values (`GITNEXUS_BENCH_AUTH_TOKEN`, `RELEASE_APP_ID`, `RELEASE_APP_PRIVATE_KEY`) and the runner's GitHub Actions registration stay deliberate manual steps -- a real API key has no business in Terraform state, and a registration token expires in about an hour, so there's nothing durable for Terraform to manage there. Registration is now a documented, repeatable script (`register-runner.sh`) instead of a one-off sequence of hand-typed SSM commands.

## Test plan

- [x] `terraform validate` -- clean
- [x] All 10 pre-existing resources imported into state
- [x] `terraform plan` after import -- 0 destroy, 2 benign in-place changes (both fixed and re-verified)
- [x] `terraform apply` run against the real, already-live infrastructure -- applied cleanly
- [x] Follow-up `terraform plan` -- "No changes. Your infrastructure matches the configuration."

Related to #2600 (the workflow/harness fix this infra supports), but kept as its own PR since it's a distinct ops/infra concern independently reviewable from the application-code fix.

### DIFF
--- a/infra/gitnexus-evolution/.gitignore
+++ b/infra/gitnexus-evolution/.gitignore
@@ -1,0 +1,16 @@
+# Local plugin cache, not portable across machines/OSes.
+.terraform/
+
+# State can contain sensitive values (IAM policy documents, ARNs) and is
+# machine-specific in this local-backend setup -- never commit it.
+*.tfstate
+*.tfstate.*
+
+# Local variable overrides (may contain tokens if someone ignores the
+# GITHUB_TOKEN-env-var guidance in the README).
+*.tfvars
+*.tfvars.json
+
+.terraform.tfstate.lock.info
+crash.log
+crash.*.log
--- a/infra/gitnexus-evolution/.terraform.lock.hcl
+++ b/infra/gitnexus-evolution/.terraform.lock.hcl
@@ -1,0 +1,47 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/hashicorp/aws" {
+  version     = "5.100.0"
+  constraints = "~> 5.0"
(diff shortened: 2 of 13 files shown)


magyargergo:
Closing this -- on reflection, publishing the exact IAM roles, security group rules, and self-hosted runner architecture for a real, live AWS account isn't something that should live in a public repo, even without literal secrets or resource IDs in the diff. Keeping this infra documented privately instead. The workflow's own activation checklist (in .github/workflows/gitnexus-skill-evolution.yml) still covers what a maintainer needs to set this up by hand.

PR #2607, opened 2026-07-21
fix(eval): bind the resolved node binary into the sandbox on self-hosted runners

magyargergo:
## Summary

The first real `workflow_dispatch` run against the new self-hosted runner (#2600) failed: every session hit `error_kind: infra-error` with `bwrap: execvp /usr/local/bin/node: No such file or directory`, tripping the outage-streak breaker after 5 consecutive failures. Run: https://github.com/abhigyanpatwari/GitNexus/actions/runs/29836411744

`sanitized_graph.py` and `runner_sessions.py` invoke the sandboxed graph CLI at the fixed path `/usr/local/bin/node`. `_runtime_mount_args` only binds `/usr`, `/bin`, `/lib`, `/lib64` wholesale, so that path resolves correctly when `node` happens to live under `/usr/local/bin` on the host -- true on GitHub-hosted runner images, but not on a self-hosted runner, where `actions/setup-node` installs into its own tool-cache directory instead (outside all four bound trees, invisible to the sandbox regardless of what `PATH` says on the host process).

Fix lives entirely in the mount construction: resolve `node` via `shutil.which` (correctly picks up wherever `actions/setup-node` put it, since its tool-cache dir is already on `PATH` by the time this runs) and bind it read-only to the same fixed sandbox path the two call sites already expect. Neither call site needed to change. Backward compatible with GitHub-hosted runners -- there this resolves to the same path and binds a harmless no-op self-mount.

## Test plan

- [x] `uv run --locked --extra dev python -m pytest tests -q` -- 313 passed, 10 skipped (pre-existing, real-bwrap-gated)
- [x] `ruff check` / `ruff format --check` clean
- [x] New tests cover both the self-hosted case (node resolved outside the bound trees) and the no-node-found case (bind skipped, not a hard error)
- [ ] Re-dispatch `workflow_dispatch` against the self-hosted runner with this fix to confirm the actual failure is resolved (next step after merge)

### DIFF
--- a/eval/tests/test_proposer_sandbox.py
+++ b/eval/tests/test_proposer_sandbox.py
@@ -24,6 +24,7 @@ from workflow_bench.proposer_sandbox import (
     SANDBOX_USER_SKILLS,
     ReadOnlyMount,
     SandboxError,
+    _runtime_mount_args,
     build_claude_settings,
     build_sandbox_environment,
     prepare_sandbox,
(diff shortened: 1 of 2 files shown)


magyargergo:
@gitnexus review

PR #2608, opened 2026-07-21
fix(mcp): report every rename edit that apply writes (#2605)

### LINKED ISSUE #2605: rename reports total_edits: 1 while rewriting 4 occurrences, in both dry-run and apply
## Summary

`rename` reports `total_edits: 1` but rewrites 4 occurrences. The under-report happens in both `dry_run: true` and `dry_run: false`, so the preview is not a faithful preview of the apply, and the post-apply result does not describe what landed on disk.

The rename itself is correct. Only the reported edit set is wrong.

## Environment

- GitNexus 1.6.9
- `@ladybugdb/core` ^0.18.0
- node v26.4.0, Linux x86_64

## Reproduction

`src/lib.rs` in a fresh crate (`edition = "2021"`):

```rust
/// Private free function with exactly 4 textual occurrences:
/// this definition, one production call, two test calls.
fn rename_target(x: u32) -> u32 {
    x + 1
}

pub fn prod_call() -> u32 {
    rename_target(1)
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn unit_one() {
        assert_eq!(rename_target(1), 2);
    }

    #[test]
    fn unit_two() {
        assert_eq!(rename_target(2), 3);
    }
}
```

Ground truth from `grep -n rename_target src/lib.rs`: lines 31, 36, 45, 50.

```sh
git init && git add -A && git commit -m "repro"
gitnexus analyze .
```

Then call the `rename` MCP tool twice against that one index:

```json
{"name": "rename", "arguments": {
  "symbol_name": "rename_target", "new_name": "renamed_fn", "dry_run": true
}}
```

```json
{"name": "rename", "arguments": {
  "symbol_name": "rename_target", "new_name": "renamed_fn", "dry_run": false
}}
```

## Results

| call | `applied` | `total_edits` | reported lines |
| --- | --- | --- | --- |
| `dry_run: true` | `false` | 1 | `[31]` |
| `dry_run: false` | `true` | 1 | `[31]` |

Both report only the definition. The full dry-run payload:

```json
{
  "status": "success",
  "old_name": "rename_target",
  "new_name": "renamed_fn",
  "files_affected": 1,
  "total_edits": 1,
  "graph_edits": 1,
  "text_search_edits": 0,
  "changes": [
    {
      "file_path": "src/lib.rs",
      "edits": [
        {
          "line": 31,
          "old_text": "fn rename_target(x: u32) -> u32 {",
          "new_text": "fn renamed_fn(x: u32) -> u32 {",
          "confidence": "graph"
        }
      ]
    }
  ],
  "applied": false
}
```

After the `dry_run: false` call, the file actually contains 4 renamed occurrences and 0 stragglers:

```sh
grep -c renamed_fn src/lib.rs      # 4
grep -c rename_target src/lib.rs   # 0
cargo check                        # exit 0
```

So lines 36, 45 and 50 were rewritten without ever appearing in `changes`.

Note `text_search_edits: 0` despite three plain-text occurrences of `rename_target` remaining after the definition. The documented regex fallback either did not run or its hits are not counted.

## Why it matters

The tool description says "Preview by default" and tags each edit with a confidence so the caller can decide what to accept. An agent gating on that preview sees a single low-blast-radius edit to a definition and cannot see that three call sites, two of them in tests, are about to change too. The same wrong edit list comes back after the apply, so there is no point at which the caller learns what happened.

The failure is silent: `status` is `"success"` and no field indicates the edit list is partial.

magyargergo:
## Summary

`rename` reported `total_edits` from a **partial** enumeration while the apply step does a **whole-file** `\bold\b` global replace on every touched file. When a symbol's definition and all its call sites live in one file (private/local symbols), only the definition line was reported (`total_edits: 1`) even though apply rewrote every occurrence — in both `dry_run: true` and `dry_run: false`. The preview was not a faithful preview of the apply, and the same wrong list came back after apply.

Fixes #2605.

## Root cause

`gitnexus/src/mcp/local/local-backend.ts` `rename()` built the reported `changes` map from three under-counting sources:

- the **definition line** only;
- **one edit per graph-ref file**, then `break` (a file with N call sites reported 1);
- **text search that skipped any file already covered by the graph** (`if (graphFiles.has(...)) continue`) — so additional occurrences inside the definition/graph files were never enumerated.

Apply, meanwhile, ignored that list and did a whole-file `content.replace(/\bold\b/g, new)` on each file. In the issue's Rust repro the definition and all three call sites live in `src/lib.rs`, the graph has no incoming refs for the private fn, so the report was `total_edits: 1` while all 4 occurrences were rewritten on disk.

## Fix

Rebuild `changes` / `total_edits` / `graph_edits` / `text_search_edits` from a single source of truth:

1. **Classify** each file to rewrite into `graph` (definition + graph-ref files) or `text_search` (ripgrep-only files, never downgrading a graph file).
2. **Enumerate** every `\bold\b` line in each classified file using the same escaped global regex the apply step uses.

Because apply's per-file global replace rewrites exactly the lines the enumeration reports, the preview and the applied result are now provably equal. The apply loop, the result object, and the ripgrep discovery call are unchanged — the rename result itself was already correct.

## Testing

- New `gitnexus/test/unit/rename-edit-report.test.ts` reproduces the exact single-file / empty-graph case (definition + 3 same-file call sites). It **fails on the pre-fix tree** (`expected 1 to be 4`) and passes after the fix; occurrence lines are computed from the fixture, and it asserts the reported edit set equals what lands on disk in both dry-run and apply.
- `calltool-dispatch` + `tools` unit suites green (258 tests); `tsc --noEmit` clean.


magyargergo:
@gitnexus review

abhigyanpatwari:
Thanks for the review. Addressed in `54c44d91`:

- **MED — partial-write inflates totals** ✅ Fixed. On a non-dry-run where a file's `writeFile` throws, its edits are now dropped from `total_edits`/`graph_edits`/`text_search_edits`/`changes`; counts are derived once from the reported (landed) set, and `failed_files` still lists the drops. This is the same "report must match what actually landed" invariant #2605 is about. New test `reports only files that landed when a write fails mid-apply` covers it.
- **MED — per-line `RegExp` recompilation** ✅ Fixed. Hoisted `wordTest` (non-global, for `.test()`) and `wordReplace` (global, for `.replace()`) out of the per-line loop — one compile each per call instead of per line — and the apply loop reuses `wordReplace`.
- **MED — enumerate/apply two-read TOCTOU under concurrency** ⏸️ Deferred as pre-existing (out of scope). As the review notes, the two-read pattern predates this diff; closing it means a per-repo mutex/queue on `callTool` dispatch, which is a separate architectural change affecting every tool. I removed the over-claiming "provably in sync" wording and left an explicit comment that this method does not lock against an external write landing between preview and apply. Happy to file a follow-up for a dispatch-level lock.
- **LOW — graph-ref multi-occurrence never exercised** ✅ Fixed. New multi-file test drives a graph-referencing caller file (via `incoming.calls`) with multiple occurrences, so the old "one edit per file then `break`" trigger is now covered.
- **LOW — confidence classification / never-downgrade only trivially exercised** ✅ Fixed. The multi-file test asserts concrete per-file `confidence` values and the `graph`/`text_search` split (not just their sum), including a file reached by both the graph and ripgrep staying `graph` (never downgraded, via a stubbed `rg`) and a pure `text_search`-only file.
- **LOW — duplicated `oldName` escape formula** ✅ Fixed. The apply loop now reuses `escapedOldName`/`wordReplace` instead of recomputing the escape inline.

Apply's on-disk behavior is unchanged (still a whole-file word-boundary global replace). Diff remains the one method + its test file; `tsc --noEmit` clean, unit suites green.

abhigyanpatwari:
@gitnexus review

Re-review requested on head `54c44d91` — the prior REQUEST CHANGES findings (partial-write totals, per-line regex, confidence-classification coverage, duplicated escape) were addressed in `54c44d91`; the enumerate/apply TOCTOU was left as pre-existing with rationale above.

magyargergo:
@gitnexus review

PR #2610, opened 2026-07-21
fix(rust): resolve trait-object (&dyn Trait) dispatch producing no CALLS edge

### LINKED ISSUE #2604: impact returns 0 callers for Rust trait-object dispatch: no CALLS edge is emitted for `&dyn Trait` call sites
## Summary

In Rust, a call made through a trait object (`&dyn Trait`) produces no `CALLS` edge, so `impact` reports `impactedCount: 0` for a method that has a live caller. Every other dispatch shape I tested resolves correctly, so this looks specific to trait objects rather than a general call-graph gap.

This is distinct from #2508. There the `CALLS` edge exists and a predicate drops it on read. Here the edge is never written.

## Environment

- GitNexus 1.6.9
- `@ladybugdb/core` ^0.18.0
- node v26.4.0, Linux x86_64

## Reproduction

`Cargo.toml`:

```toml
[package]
name = "gnx-repro"
version = "0.1.0"
edition = "2021"
```

`src/lib.rs`:

```rust
pub struct Inner;

impl Inner {
    /// Called only through a struct field.
    pub fn field_dispatch_target(&self) -> u32 {
        1
    }
}

pub struct Outer {
    inner: Inner,
}

impl Default for Outer {
    fn default() -> Self {
        Self { inner: Inner }
    }
}

impl Outer {
    pub fn calls_through_field(&self) -> u32 {
        self.inner.field_dispatch_target()
    }
}

fn rename_target(x: u32) -> u32 {
    x + 1
}

pub fn prod_call() -> u32 {
    rename_target(1)
}

pub struct Assoc;

impl Assoc {
    fn via_self_target(x: u32) -> u32 {
        x + 1
    }

    fn via_type_target(x: u32) -> u32 {
        x + 2
    }

    pub fn calls_via_self(&self) -> u32 {
        Self::via_self_target(1)
    }

    pub fn calls_via_type(&self) -> u32 {
        Assoc::via_type_target(1)
    }
}

pub trait Behaviour {
    fn trait_target(&self) -> u32;
}

pub struct Impl1;

impl Behaviour for Impl1 {
    fn trait_target(&self) -> u32 {
        7
    }
}

/// Dispatch through a trait object.
pub fn calls_via_dyn(b: &dyn Behaviour) -> u32 {
    b.trait_target()
}
```

Then:

```sh
git init && git add -A && git commit -m "repro"
gitnexus analyze .
for t in field_dispatch_target via_self_target via_type_target trait_target; do
  gitnexus impact "$t" --direction upstream
done
```

## Results

Each target has exactly one caller in the source.

| target | call shape | `impactedCount` |
| --- | --- | --- |
| `field_dispatch_target` | `self.inner.method()` | 1 |
| `via_self_target` | `Self::assoc()` | 1 |
| `via_type_target` | `Type::assoc()` | 1 |
| `trait_target` | `b.method()` where `b: &dyn Behaviour` | **0** |

`impact trait_target --direction upstream` returns `impactedCount: 0`, `risk: LOW`, even though `calls_via_dyn` calls it.

## Root cause evidence

The whole graph contains four `CALLS` edges, and the trait-object one is absent:

```sh
gitnexus cypher "MATCH (c)-[r:CodeRelation {type:'CALLS'}]->(t) RETURN c.name AS caller, t.name AS callee"
```

```text
| caller               | callee                |
| -------------------- | --------------------- |
| calls_through_field  | field_dispatch_target |
| prod_call            | rename_target         |
| calls_via_self       | via_self_target       |
| calls_via_type       | via_type_target       |
```

`calls_via_dyn` has no outgoing edges at all:

```sh
gitnexus cypher "MATCH (c)-[r:CodeRelation]->(t) WHERE c.name = 'calls_via_dyn' RETURN t.name AS target, r.type AS type"
```

```text
[]
```

The only edges pointing at `trait_target` come from the trait and impl declarations, never from the call site:

```sh
gitnexus cypher "MATCH (c)-[r:CodeRelation]->(t) WHERE t.name = 'trait_target' RETURN c.name AS caller, r.type AS type"
```

```text
| caller       | type              |
| ------------ | ----------------- |
| trait_target | METHOD_IMPLEMENTS |
| Impl1        | HAS_METHOD        |
```

`context trait_target` likewise reports no incoming calls.

## Control

`calls_via_dyn` is a free function. Every passing case is a method, so I checked that free functions are not the differentiator. `prod_call` is a free function calling a free function and its edge is present:

```sh
gitnexus cypher "MATCH (c)-[r:CodeRelation]->(t) WHERE c.name = 'prod_call' RETURN t.name AS target, r.type AS type"
```

```text
| target        | type      |
| ------------- | --------- |
|               | MEMBER_OF |
| rename_target | CALLS     |
```

So the differentiator is the trait-object receiver, not the enclosing item kind.

## Why it matters

A 0-caller result reads as "nothing depends on this, safe to change or delete". Trait objects are common in Rust for plugin registries and anything boxed behind an interface, so the symbols most likely to be dynamically dispatched are the ones this silently reports as unreferenced.

magyargergo:
## Summary

Fixes #2604. In Rust, a call through a trait object (`&dyn Trait`) produced no `CALLS` edge, so `impact` reported 0 callers for a method reached only through dynamic dispatch. Two independent, compounding gaps in the Rust language provider caused this:

1. `normalizeRustTypeName`/`normalizeRustReturnType` never stripped the `dyn` keyword, so a `&dyn Trait`-typed receiver normalized to the literal string `"dyn Trait"` instead of `"Trait"` — an unmatchable name.
2. Abstract trait methods (`fn foo(&self) -> T;`, no body) parse as `function_signature_item`, a grammar node distinct from `function_item` that the Rust scope-extractor query never captured — so even with (1) fixed, the trait's method had no scope/declaration for the receiver-bound-calls resolver to dispatch to.

Both gaps needed fixing together — verified empirically (built the analyzer from source, reproduced the issue's exact repro, confirmed the `CALLS` edge appears and `impact()` goes from 0 to correct callers after both fixes, reverted, then re-applied for real).

A third commit bumps `INCREMENTAL_SCHEMA_VERSION` (10 → 11): the new capture only takes effect on files that get re-parsed, so an existing incrementally-maintained index would otherwise never backfill these edges for unchanged Rust trait files — same contract as the `v7`/`v10` bumps already in this file for the identical defect class.

## Test plan

- [x] New regression test (`rust.test.ts`, "Rust dyn trait-object dispatch (#2604)") with a dedicated minimal fixture, proven to discriminate (fails pre-fix, passes post-fix)
- [x] `npx vitest run test/integration/resolvers/rust.test.ts` — full Rust integration suite green
- [x] `UPDATE_GOLDEN=1 npx vitest run test/unit/scope-resolution/rust/rust-captures-golden.test.ts` — regenerated and committed (expected drift: 8 fixtures with required trait methods)
- [x] `node scripts/build.js` succeeds
- [x] `detect_changes(scope: "compare", base_ref: <fork point>)` — only the expected Rust-provider symbols affected, low risk


magyargergo:
@gitnexus review

PR #2613, opened 2026-07-21
fix(config): honor core.excludesFile and .git/info/exclude for global ignores (#2606)

### LINKED ISSUE #2606: IgnoreService has no user-level ignore file: global exclusions must be repeated per repo or patched into node_modules
## Summary

`IgnoreService` reads only per-repo ignore files. There is no user-level or global ignore path, so an exclusion that should apply everywhere has to be repeated in every indexed repo, or patched into the shipped defaults where the next upgrade wipes it.

## Current behaviour

In 1.6.9, `dist/config/ignore-service.js` consults exactly two sources:

- `.gitignore` (line 229)
- `.gitnexusignore` (line 258)

There is no reference to `homedir` or `HOME` anywhere in that file, and none of these are read:

- `.git/info/exclude`
- git's `core.excludesFile`
- any `~/.config/gitnexus/` path
- any environment variable naming an ignore file

`DEFAULT_IGNORE_LIST` and `IGNORED_EXTENSIONS` in the same file are the only global layer, and they ship inside the package.

## Why per-repo files are not enough

The concrete case: I index 15 repos and want documentation excluded in all of them, because fenced code blocks in `.md` files become graph symbols under `src/*` paths that do not exist as code. Doing that today means one of:

1. Adding `.gitnexusignore` to all 15 repos, then to every repo added later. Several of those repos are not mine to add a file to.
2. Putting `docs` in git's `core.excludesFile`, which GitNexus does not read.
3. Editing `DEFAULT_IGNORE_LIST` and `IGNORED_EXTENSIONS` in `node_modules`, which is what I ended up doing. `bun add -g gitnexus` wipes it on every upgrade, so it needs a post-upgrade hook to reapply.

Option 3 is the only one that scales, and it is the one that silently reverts.

## Request

Read a user-level ignore file in addition to the per-repo ones. Any of these would solve it:

- honour git's own `core.excludesFile` and `.git/info/exclude`, which is what users already expect from a tool that reads `.gitignore`
- a `~/.config/gitnexus/ignore` file with the same syntax as `.gitnexusignore`
- an env var pointing at an ignore file

Precedence that would work: shipped defaults, then user-level, then per-repo, with later layers able to negate earlier ones the way `.gitnexusignore` negation already works.

## Related

#2510 covers where tool-usage policy should live rather than what gets indexed, so this is a separate axis. #1265 shipped `--skip-agents-md` for the file-writing side, which is the closest existing opt-out and does not affect index scope.

magyargergo:
## Summary

- `IgnoreService` only read per-repo `.gitignore`/`.gitnexusignore`, so an exclusion meant to apply across every indexed repo had to be repeated per repo or hand-patched into `node_modules` (wiped on every upgrade). Closes #2606.
- `loadIgnoreRules` now also reads the same two sources real `git` itself consults for exactly this purpose ([gitignore(5)](https://git-scm.com/docs/gitignore)):
  - **`core.excludesFile`** — git's own all-repos global ignore file (defaults to `$XDG_CONFIG_HOME/git/ignore` when unconfigured).
  - **`$GIT_COMMON_DIR/info/exclude`** — per-repo, untracked, so it works without push/commit access to the repo (shared across every linked worktree, same as git's own resolution).
- Precedence mirrors git's own exactly (lowest → highest): `core.excludesFile` → `.git/info/exclude` → `.gitignore` → `.gitnexusignore`. Each later source can negate an earlier one via a `!pattern` line (last-match-wins), same semantics `.gitignore` → `.gitnexusignore` already used.
- `GITNEXUS_NO_GLOBAL_IGNORE` (or `noGlobalIgnore` option) skips both global sources, mirroring the existing `GITNEXUS_NO_GITIGNORE`.
- No signature changes to any exported function, so all existing callers (verified via `impact()` — 8 direct, 36 total upstream, all passing no options) are unaffected.
- New `getCoreExcludesFilePath`/`getGitInfoExcludePath` helpers added to `git.ts`, following the same `execSync` + `--git-common-dir` pattern as the existing `getCanonicalRepoRoot`, and memoized per repo path — see the performance note below.

<details>
<summary>Design note: an earlier revision of this PR used a custom <code>~/.gitnexus/ignore</code> file instead</summary>

The issue listed three mechanisms that would each solve the problem; the first push of this PR picked the "new global file" option, reusing GitNexus's existing `~/.gitnexus` directory (same place as `registry.json`/`config.json`). Review feedback pointed out that reusing real git mechanisms (`core.excludesFile` + `.git/info/exclude`) is more consistent with what a tool that already reads `.gitignore` is expected to honor, and avoids inventing a new file location. This PR now does that instead — no custom file, no new directory, just the two sources git itself already defines for this exact use case.
</details>

<details>
<summary>Performance note: memoization added after a self-review pass</summary>

`loadIgnoreRules` is called once per repo, *per language/contract extractor*, during a `group sync` — `discoverWorkspaceLinks` sequentially runs all six workspace extractors (rust/node/python/go/java/elixir), and each one's internal loop calls `loadIgnoreRules` once per package/repo in the group. Without caching, an N-repo group would spawn up to ~12×N synchronous `git` subprocesses just to resolve the two new global sources — directly working against the "I index 15 repos" scenario issue #2606 describes. `getCoreExcludesFilePath`/`getGitInfoExcludePath` now memoize by `fromPath` in a process-lifetime `Map`, so repeated calls with the same path spawn `git` exactly once. Locked in with dedicated cache-hit tests in `git.test.ts`.
</details>

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `gitnexus/test/unit/git.test.ts`: `getGitInfoExcludePath`/`getCoreExcludesFilePath` coverage (absolute path join, worktree-shared common dir, not-a-git-repo → null, configured vs. default-fallback, git-unavailable fallback) plus dedicated caching tests (repeated calls with the same path spawn `git` exactly once, null results are cached too, different paths get independent cache entries)
- [x] `gitnexus/test/unit/ignore-service.test.ts`: 187/187 passing, including coverage for: each global source honored independently, `.git/info/exclude` negating a `core.excludesFile` rule, per-repo `.gitnexusignore` negating both, graceful skip when not a git repo, `GITNEXUS_NO_GLOBAL_IGNORE`/`noGlobalIgnore` bypass, missing-file no-op, four-layer combination, unreadable-file warns without throwing
- [x] Full `npm run test:unit` suite: 10375/10429 passing (remaining failures are pre-existing full-suite parallel-load timeouts — heap-respawn, skip-git-cli, pdg-mode-flip, etc. — confirmed by re-running all affected files in isolation at 121/121 green; none touch `ignore-service.test.ts` or `git.test.ts`)
- [x] `detect_changes` against the merge-base: risk low, 0 unexpected affected symbols/processes
- [x] Empirically verified both `git` CLI invocations (`git rev-parse --path-format=absolute --git-common-dir`, `git config --get --type=path core.excludesFile`) against a real repo, not just mocked unit tests — including `~/` expansion and exit-1-on-unset behavior (git 2.49.0)

### DIFF
--- a/gitnexus/README.md
+++ b/gitnexus/README.md
@@ -511,9 +511,16 @@ For very large repositories:
 # Increase Node.js heap size
 NODE_OPTIONS="--max-old-space-size=16384" npx gitnexus analyze
 
-# Exclude large directories
+# Exclude large directories (this repo only)
 echo "vendor/" >> .gitnexusignore
 echo "dist/" >> .gitnexusignore
+
+# Exclude a directory across every repo you index, without touching each
+# repo's own .gitnexusignore. Same syntax; lives next to registry.json and
+# config.json under the global GitNexus directory ($GITNEXUS_HOME, default
+# ~/.gitnexus). A repo's own .gitignore/.gitnexusignore can still override
+# it with a `!pattern` negation. Skip it entirely with GITNEXUS_NO_GLOBAL_IGNORE=1.
+mkdir -p ~/.gitnexus && echo "docs/" >> ~/.gitnexus/ignore
 ```
 
 ### Large files are being skipped
--- a/gitnexus/README.md
+++ b/gitnexus/README.md
@@ -516,11 +516,14 @@ echo "vendor/" >> .gitnexusignore
 echo "dist/" >> .gitnexusignore
 
 # Exclude a directory across every repo you index, without touching each
-# repo's own .gitnexusignore. Same syntax; lives next to registry.json and
-# config.json under the global GitNexus directory ($GITNEXUS_HOME, default
-# ~/.gitnexus). A repo's own .gitignore/.gitnexusignore can still override
-# it with a `!pattern` negation. Skip it entirely with GITNEXUS_NO_GLOBAL_IGNORE=1.
-mkdir -p ~/.gitnexus && echo "docs/" >> ~/.gitnexus/ignore
+# repo's own .gitnexusignore or needing push/commit access to it. GitNexus
+# reads the same sources `git` itself does: core.excludesFile (all repos)
+# and $GIT_DIR/info/exclude (this repo only, untracked). A repo's own
+# .gitignore/.gitnexusignore can still override either with a `!pattern`
+# negation. Skip both entirely with GITNEXUS_NO_GLOBAL_IGNORE=1.
+git config --global core.excludesFile ~/.gitignore_global   # applies to every repo
+echo "docs/" >> ~/.gitignore_global
+echo "build/" >> .git/info/exclude                          # this repo only, untracked
 ```
 
 ### Large files are being skipped
--- a/gitnexus/src/config/ignore-service.ts
+++ b/gitnexus/src/config/ignore-service.ts
@@ -3,7 +3,7 @@ import fs from 'fs/promises';
 import nodePath from 'path';
 import type { Path } from 'path-scurry';
 import { logger } from '../core/logger.js';
-import { getGlobalDir } from '../storage/repo-manager.js';
+import { getCoreExcludesFilePath, getGitInfoExcludePath } from '../storage/git.js';
 
 const DEFAULT_IGNORE_LIST = new Set([
   // Version Control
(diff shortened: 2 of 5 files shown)


magyargergo:
@gitnexus review

magyargergo:
@gitnexus review

PR #2614, opened 2026-07-21
fix(lbug/analyze): atomic index swap + read-pool staleness invalidation

### LINKED ISSUE #2599: Windows: analyze fails WAL checkpoint rotation while gitnexus mcp holds the store open; recovery hint points at the wrong cause
### Area

### Summary

On Windows, `analyze` fails during WAL checkpoint rotation whenever a `gitnexus mcp` server process holds the store open, and the error's recovery hint (checkpoint-threshold tuning) points at the wrong cause.

### Context

Using gitnexus for day-to-day code intelligence on a private repo (~2,700 files, ~111k nodes, PDG layers enabled). The MCP server runs persistently inside an MCP client (Claude Code) while `analyze --pdg` is invoked from the CLI to refresh the index mid-session. Upgraded 1.6.8 to 1.6.9 the same day; the related version-skew datapoint is at the bottom.

### Expected behavior

`analyze --pdg` completes and swaps in the fresh index, or, if the store is held by another process, fails with a message naming the actual conflict (e.g. "store is held open by another gitnexus process (pid N)").

### Actual behavior

The run dies partway with a checkpoint-threshold hint that misdirects:

```
LadybugDB failed while rotating/removing WAL checkpoint files.
This can happen when auto-checkpoint runs at the default threshold (~16MB).
Retry with a larger checkpoint threshold:
  gitnexus analyze --wal-checkpoint-threshold 67108864
```

Retrying with `GITNEXUS_WAL_CHECKPOINT_THRESHOLD=67108864` fails identically and quickly, so checkpoint frequency is not the cause. The same `analyze --pdg` on the same commit succeeds when no live MCP server holds the store (verified in both directions the same morning). Diagnosis: Windows file-sharing semantics; WAL rotation needs delete/rename on files the MCP server has open handles to, which Windows refuses.

### Steps to reproduce

1. On Windows, index a repo (`gitnexus analyze --pdg`) so a root store exists.
2. Start the MCP server (`gitnexus mcp`, e.g. spawned by an MCP client such as Claude Code) and serve at least one query so the server holds `.gitnexus/lbug` open.
3. With that server still running, run `gitnexus analyze --pdg` in the same repo.
4. Observe the WAL checkpoint rotation failure above; retry with the suggested threshold and observe the identical failure.
5. Stop the MCP server and re-run the same command; observe success.

### Environment

- OS: Windows 11 Home 10.0.26200
- Node: 22.x (via npm-global install)
- Browser (if applicable): n/a
- Commit / version: gitnexus 1.6.9 (npm), reproduced identically after upgrading from 1.6.8

### Logs / screenshots

```
{"level":50,"name":"gitnexus","recoveryHint":"wal-checkpoint-threshold","msg":"  LadybugDB failed while rotating/removing WAL checkpoint files.\n  This can happen when auto-checkpoint runs at the default threshold (~16MB).\n  Retry with a larger checkpoint threshold to reduce checkpoint frequency:\n    gitnexus analyze --wal-checkpoint-threshold 67108864\n    (or set GITNEXUS_WAL_CHECKPOINT_THRESHOLD=67108864)\n    (Try 33554432 = 32 MiB on small-disk / CI runners.)\n"}
```

Related datapoint from the same session (possibly just a docs note): the MCP server binds its binary version for the process lifetime. After upgrading 1.6.8 to 1.6.9, the still-running old server could not read the store the new CLI wrote (`Trying to read a database file with a different version. Database file version: 42, Current build storage version: 41`) until reconnected. Expected individually, but together the two mean "upgrade + reindex mid-session" has two separate traps on Windows.

abhigyanpatwari:
## Summary

While `analyze` runs, the MCP read pool can serve **indefinitely-stale** or **half-built** index data. This PR closes both windows and folds in a set of related lbug/analyze hardening. LadybugDB is single-writer with concurrent read-only readers (MVCC) and exposes **no lock finer than the whole database**, so this is snapshot-generation coordination, not locking.

## Core changes

**1 — read-pool invalidation (`pool-adapter.ts`)** — `initLbug` early-returned on an existing pool entry with no freshness check, so after a rebuild the pool served the old (POSIX: unlinked-but-open) inode for up to `IDLE_TIMEOUT_MS` (5 min). Each `PoolEntry` now records `{ino, mtimeMs, size}` at open and `initLbug` re-stats: unchanged → reuse; changed & idle → `closeOne` + reopen; changed while a query is in flight → serve the current handle (a later idle `initLbug` reopens). A stat failure (ENOENT during a rebuild's unlink window) is treated as unchanged. Mirrors the bridge cache's mtime-invalidation.

**2 — atomic full-rebuild swap (`run-analyze.ts`)** — the full-rebuild path wiped the live index and rebuilt in place. Now the build targets `<lbugPath>.new` and is published over the live index in one atomic `rename`; the close checkpoint-consolidates the build to a single file (verified: no residual `.wal`/`.shadow`). A concurrent reader only ever sees the previous complete index, and a failure before the swap leaves the previous index byte-for-byte intact.

## Folded-in follow-ups

- **`setQueryTimeout`** — pooled read connections now set the engine-level query timeout, so a pathological query is bounded at the source, not just by the JS-side `Promise.race`.
- **`#2599` checkpoint-busy** — `isLbugCheckpointBusyError` classifies a WAL-checkpoint IO error that also carries a busy/lock signal, and the driver names the held-open cause (another `gitnexus mcp` server / stale reader) instead of a raw IO string when it exhausts retries. *(Overlaps in-flight `repro/issue-2599`; bundled at maintainer request.)*
- **Retry-budget consolidation** — the read-pool and bridge open-retry budgets move into the `lbug-config` retry registry (`POOL_OPEN_LOCK_RETRY_*`, `BRIDGE_OPEN_RETRY_*`); one tuning surface, no behavior change.
- **Atomic incremental (opt-in)** — `GITNEXUS_ATOMIC_INCREMENTAL=1` copies the live index into the temp, applies the incremental writeback to the copy, and swaps atomically. **Off by default** because the whole-file copy negates incremental's speed premise — kept behind a flag pending a benchmark.
- **Windows atomic swap (best-effort, UNVERIFIED)** — the swap extends to Windows for non-`--pdg` runs, where a real close is safe to release the build handle before the rename (`--pdg` keeps the `#2264` in-place behavior). **Not verified on a Windows runner** — any failure falls back to in-place, so it cannot corrupt, but Windows CI should confirm before relying on it.

## Testing

- `pool-freshness-invalidation.test.ts` — identity mechanism against real temp files + the pure decision logic.
- `checkpoint-busy-2599.test.ts` — the busy classifier and the driver's held-open hint (via its test seam).
- `analyze-atomic-swap.test.ts` (POSIX integration): full-rebuild no-temp-leak + inode swap; **crash-safety** (a rebuild that fails before the swap leaves the previous index byte-for-byte intact); **end-to-end** (analyze → pool serves it → rebuild → pool re-opens onto the new index, serving the renamed function not the stale one — Change 1 + Change 2 together); **opt-in atomic incremental** (no temp leak, the incremental change is reflected after the swap).
- `tsc --noEmit` clean; existing pool / bridge / checkpoint / full-analyze suites green.

## Verification status / caveats

- **POSIX**: all changes verified (unit + integration, real `analyze` end to end).
- **Windows atomic swap**: implemented but **not runnable here** — needs a Windows runner to confirm; safe-by-construction fallback to in-place.
- **Atomic incremental**: correct on POSIX but **off by default** pending a copy-cost benchmark.
- **#2599**: overlaps a maintainer branch already in flight — may need reconciliation on merge.

### DIFF
--- a/gitnexus/src/cli/analyze.ts
+++ b/gitnexus/src/cli/analyze.ts
@@ -18,6 +18,7 @@ import { boundedCheckpointBeforeExit } from '../core/lbug/shutdown-helpers.js';
 import {
   getOsPageSize,
   isLbugCheckpointIoError,
+  isLbugCheckpointBusyError,
   isLbugPageSizeFrameError,
   isPageSizeAwareLadybug,
   isWalCorruptionError,
@@ -1624,8 +1625,16 @@ const analyzeCommandImpl = async (
     }
 
     if (isLbugCheckpointIoError(err)) {
+      // #2599: when the checkpoint IO error also looks busy/locked, another
+      // handle holds the store open — name that actionable cause alongside the
+      // threshold hint (the original error is preserved so the hint still fires).
+      const heldOpen = isLbugCheckpointBusyError(err)
+        ? `  Another process may hold the store open (a running \`gitnexus mcp\` server, or a\n` +
+          `  stale reader) — close other GitNexus processes on this repo, then retry.\n`
+        : '';
       cliError(
         `  LadybugDB failed while rotating/removing WAL checkpoint files.\n` +
+          heldOpen +
           `  This can happen when auto-checkpoint runs at the default threshold (~16MB).\n` +
           `  Retry with a larger checkpoint threshold to reduce checkpoint frequency:\n` +
           `    gitnexus analyze --wal-checkpoint-threshold ${RECOMMENDED_WAL_CHECKPOINT_THRESHOLD}\n` +
--- a/gitnexus/src/core/group/bridge-db.ts
+++ b/gitnexus/src/core/group/bridge-db.ts
@@ -8,9 +8,6 @@ import { BRIDGE_SCHEMA_QUERIES, BRIDGE_SCHEMA_VERSION } from './bridge-schema.js
 import {
   closeLbugConnection,
   openLbugConnection,
-  BRIDGE_OPEN_RETRY_ATTEMPTS,
-  BRIDGE_OPEN_RETRY_BASE_MS,
-  BRIDGE_OPEN_RETRY_MAX_MS,
   type LbugConnectionHandle,
 } from '../lbug/lbug-config.js';
 import { dedupeContracts, dedupeCrossLinks } from './normalization.js';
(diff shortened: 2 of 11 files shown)


magyargergo:
@gitnexus review

Issue #1365, opened 2026-05-06
LadybugDB VECTOR is disabled on this platform; semantic search uses exact scan when embeddings exist.


mohammad-kasaei-mohaymen:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

LadybugDB VECTOR is disabled on this platform; semantic search uses exact scan when embeddings exist.
how to resolve this in windows

### Context

```
GitNexus Doctor

Runtime
  OS:        win32/x64
  Node:      v24.13.1
  GitNexus:  1.6.4-rc.71
  LadybugDB: unknown
  ONNX:      1.25.1

Capabilities
  Graph store:     available
  Full-text search: available
  VECTOR index:    unavailable
  Semantic mode:   exact-scan
  Exact scan limit:    10000 chunks
  Note:            LadybugDB VECTOR is disabled on this platform; semantic search uses exact scan when embeddings exist.

Embeddings
  Backend:   local
  Device:    auto
  Threads:   4
  Batch:     16 nodes
  Sub-batch: 8 chunks
```

### Expected behavior

To be able to use vector too

### Actual behavior

this error
LadybugDB VECTOR is disabled on this platform; semantic search uses exact scan when embeddings exist.

### Steps to reproduce

on windows install this version

### Environment

Windows

mohammad-kasaei-mohaymen:
I aslo see this 
```
GitNexus: FTS extension unavailable; continuing without FTS features. INSTALL fts failed with exit code 1: Error: IO ex)
```
It is not clear what is the problem

smallPlan:
same question: GitNexus: FTS extension unavailable; continuing without FTS features. INSTALL fts timed out after 15000ms
(node:5548) Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTT.
(Use `node --trace-warnings ...` to show where the warning was created)
  ████████████████████████████████████░░░░  90% | Embedding 0/?

  Analysis failed: fetch failed
TypeError: fetch failed
    at node:internal/deps/undici/undici:16416:13
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async loadResourceFile (file:///C:/Users/AppData/Roaming/npm/node_modules/gitnexus/node_modules/@huggingface/transformers/dist/transformers.node.mjs:6614:18)
[...]

PR #2624, opened 2026-07-22
fix(analyze): load VECTOR before the incremental writeback touches embedding rows (#2623)

### LINKED ISSUE #2623: [1.6.10-rc.83] incremental analyze --embeddings fails: "Trying to delete from an index on table CodeEmbedding but its extension is not loaded"
### Area
`gitnexus` (CLI / core / embeddings — LadybugDB adapter)

### Summary
On `1.6.10-rc.83`, incremental `analyze --embeddings` reliably fails with `Binder exception: Trying to delete from an index on table CodeEmbedding but its extension is not loaded` after the first embeddable node has a content hash change. #2453 fixed the PK-violation class of bugs but the new per-batch `deleteStaleEmbeddingRows` (and the equivalent deletes inside `lbug-adapter.js`'s `dropNodesAndEmbeddingsForFile` / `deleteNodesForFiles`) hit this second, related bug. Affects Linux/x64 with `@ladybugdb/core` 0.18.3 and the default `load-only` install policy.

### Context
- PR #2453 (merged 2026-07-16) — fixed `CREATE` PK collisions by introducing per-batch `DELETE` of stale `CodeEmbedding` rows before each `INSERT`.
- This issue: those new `MATCH (e:CodeEmbedding {…}) DELETE e` queries now fail at the LadybugDB layer when the `CodeEmbedding.code_embedding_idx` (VECTOR) HNSW index exists from a prior analyze but the `VECTOR` extension wasn't loaded this run. `doctor` reports `Ext install: load-only (offline; load only, no network install)` and `VECTOR index: available`, so an `analyze` from a previous program/binary version can leave the index orphaned.

### Environment
- gitnexus: **1.6.10-rc.83** (npm, global install)
- Node: v24.15.0
- OS: NixOS, linux/x64
- @ladybugdb/core: 0.18.3
- Embedding backend: local (CPU, Snowflake/snowflake-arctic-embed-xs, ONNX 1.27.0)
- Project: small (~200 nodes, ~1.6k edges, ~180 embeddings); a mixed-language repo (Python, JS, shell, config)
- VECTOR unavailable to auto-install (load-only)

### Steps to reproduce
1. `gitnexus clean --force` and `gitnexus analyze --embeddings` → succeeds, builds the `code_embedding_idx` on `CodeEmbedding` and stores the embeddings. Exit code 0.
2. Modify a real Python function body enough that the embedding-model input for that symbol changes (e.g. add/rename a top-level statement; pure docstring edits don't change `contentHashForNode` in `embedding-pipeline.js:62`).
3. Run `gitnexus analyze --embeddings` again → fails on the very first stale row.

### Actual behavior
```
Incremental: changed=1, added=0, deleted=0 (skipping wipe + 206 unchanged file rows preserved)
Incremental: +13 importer(s) added to writable set (BFS depth ≤ 4)

  Analysis failed: Binder exception: Trying to delete from an index on
  table CodeEmbedding but its extension is not loaded.
Error: Binder exception: Trying to delete from an index on table
CodeEmbedding but its extension is not loaded.
```

The stack trace varies — the same DELETE pattern appears in four places:
- `lbug-adapter.js:1961` — `dropNodesAndEmbeddingsForFile` (singular)
- `lbug-adapter.js:2086` — `deleteNodesForFiles` (bulk)
- `embedding-pipeline.js:284` — `deleteStaleEmbeddingRows` (per-batch)
- `embedding-pipeline.js:214` — `batchInsertEmbeddings` (row-level)

The `dirty-flag recovery` from #2546 reports `Last dirty state: phase=embedding, effectiveWrite=N` and then auto-forces a full rebuild on every subsequent run — masks the bug as "just slow" for users, but nothing actually fixes the underlying DELETE.

### Expected behavior
Incremental `analyze --embeddings` should succeed when content hashes change, regardless of whether the VECTOR `.so` is loaded this run. Kuzu/LadybugDB documents `DELETE` on indexed properties as forbidden; either the analyzer must not create the index without verifying the extension loaded, or it must drop+recreate the index around the DELETE.

### Suggested fix (mirrors an existing internal pattern)
Detect the `delete from an index … extension is not loaded` error and run `CALL DROP_VECTOR_INDEX('CodeEmbedding','code_embedding_idx')` → retry the DELETE → optionally `CALL CREATE_VECTOR_INDEX(...)` to restore the next-run index. Patch sites identified above; constants (`EMBEDDING_TABLE_NAME`, `EMBEDDING_INDEX_NAME`) already imported from `core/lbug/schema.js` (line 452 of the schema module = `'code_embedding_idx'`). Both `executeQuery` (for `CALL …`) and `executeWithReusedStatement` (for `MATCH…DELETE`) are already in scope at each site.

Sketch (one of the four):
```js
catch (err) {
  if (msg.includes('delete from an index') || msg.includes('extension is not loaded')) {
    await queryAndDrain(targetConn,
      `CALL DROP_VECTOR_INDEX('${EMBEDDING_TABLE_NAME}', '${EMBEDDING_INDEX_NAME}')`);
    await queryAndDrain(targetConn, /* original MATCH … DELETE … */);
    try {
      await queryAndDrain(targetConn,
        `CALL CREATE_VECTOR_INDEX('${EMBEDDING_TABLE_NAME}', '${EMBEDDING_INDEX_NAME}', 'embedding', metric := 'cosine')`);
    } catch (_) { /* next analyze will rebuild it */ }
    return;
  }
  throw err;
}
```

### Workaround (local, not a config flag)
I monkey-patched the four sites locally and the loop closes — 2 stale embeddings → success in 9.2s on the same repo. Diff is small (try/catch + DROP/RECREATE around the existing DELETE) and contained to the four listed sites; happy to share a patch file if useful.

A more durable upstream fix would also gate `buildVectorIndex` in `embedding-pipeline.js:215` and the clean-VECTOR-on-empty-index path so an orphaned VECTOR index is never left behind when the extension load state changes between runs.

### Related
- #2452, #2453 (PR), #2544, #2546 — same symptom family, all reported after 1.6.9.
- #2546 already notes: "incremental embeddings crash deterministically on 1.6.9 every routine incremental catch-up". This issue is the second part of that family.

magyargergo:
Fixes #2623.

## What was happening

On any repo that had built the `code_embedding_idx` HNSW index, every incremental `analyze` after a content change died with:

```
Analysis failed: Binder exception: Trying to delete from an index on table
CodeEmbedding but its extension is not loaded.
```

The next run then tripped the dirty-flag recovery and forced a full rebuild, which is why this mostly read as "incremental analyze is just slow" rather than as a hard failure.

## Root cause: ordering, not environment

The issue reports this as NixOS / offline / `load-only` specific. It is not — it reproduces end to end on a machine where the VECTOR extension loads perfectly well.

The surgical incremental writeback's first statement is the `CodeEmbedding` join-delete inside `deleteNodesForFiles` (`lbug-adapter.ts`), reached from `run-analyze.ts` right after the #2589 FTS drop seam. Nothing on that path loads VECTOR — the only `loadVectorExtension` call sites are `createVectorIndex` and the embedding pipeline's `ensureVectorExtensionAvailable`, both first reached in Phase 4, long after the delete. LadybugDB refuses all DML against a table carrying an HNSW index unless the extension is loaded on that connection, so the delete was rejected.

## Why the fixes proposed in the issue cannot work

Both suggestions in the thread route recovery through `DROP_VECTOR_INDEX`. Probed against `@ladybugdb/core` 0.18.2, the version `gitnexus/package-lock.json` pins, with an index present and VECTOR unloaded. Every case was also re-run on 0.18.0 and the results are identical:

| Statement | Result |
| --- | --- |
| `MATCH (e:CodeEmbedding) ... DELETE e` | `Binder exception: Trying to delete from an index ...` |
| `CREATE (e:CodeEmbedding {...})` | `Binder exception: Trying to insert into an index ...` |
| `MATCH (e:CodeEmbedding) ... SET e.contentHash='x'` (non indexed property) | SIGSEGV, exit 139 |
| `CALL DROP_VECTOR_INDEX('CodeEmbedding','code_embedding_idx')` | `Catalog exception: function DROP_VECTOR_INDEX is not defined` |
| `DROP TABLE CodeEmbedding` | `Binder exception: Cannot delete node table ... referenced by index` |
| `CALL SHOW_INDEXES() RETURN *` | works, reports `extension_loaded` per index |

`DROP_VECTOR_INDEX` is itself a VECTOR extension function, so it is undefined in exactly the state it would need to rescue. Without the extension the table is fully immutable: no delete, no insert, no drop. Loading the extension is the only in place repair.

`SHOW_INDEXES` does work without the extension, so detection needs no error string matching.

## The change

Two small pieces.

`ensureEmbeddingRowDmlSafe()` in `lbug-adapter.ts` loads VECTOR under the analyze install policy and reports whether embedding row DML is legal. Only if the load fails does it read `SHOW_INDEXES` to see whether an index actually exists to trip over, so a DB that never built one is not blocked.

`runFullAnalysis` calls it once, before the escalation gate and before any row is touched, and folds a blocked result into the existing wipe and bulk COPY escalation branch. This is the same seam `dropSearchFTSIndexes` occupies for FTS in #2589: index lifecycle first, then rows.

The call is unconditional rather than gated on `--embeddings`, because a DB carrying the index from an earlier embeddings run hits the identical wall on a plain incremental run.

One central seam rather than four independent catch and retry sites. It also covers the Phase 3.5 restore and the legacy orphan sweep, which mutate `CodeEmbedding` on the same connection later in the run. The index is deliberately kept rather than dropped: `run-analyze.ts` relies on the HNSW index surviving a surgical run, and dropping it would silently downgrade every incremental run to exact scan.

## Verification

Regression tests at both levels, and both were confirmed to discriminate. With the `run-analyze.ts` change reverted they fail with the exact reported Binder exception, and pass with it.

* `lbug-delete-nodes-for-files.test.ts` covers the three adapter cases: no index and VECTOR unavailable stays safe, index present and VECTOR unavailable is reported blocked and the raw delete genuinely throws, index present and VECTOR loadable is safe with the index surviving.
* `incremental-vector-extension-ordering.test.ts` is the sibling of the #2589 ordering suite. It drives the real `runFullAnalysis` incremental path against a real git repo and a real LadybugDB, and asserts `extension_loaded` is true at the exact moment `deleteNodesForFiles` is invoked.

Suites:

* `node scripts/build.js` exit 0, tsc clean for both packages
* `npm run test:unit` 513 files, 10401 passed, 0 failed
* `npm run test:integration` 205 files passed, 1 failed (`fts-extension-e2e.test.ts`) which is pre existing and reproduces identically at the base commit with these changes reverted
* `npm run format:check` clean for every changed file

Live end to end on a small mixed language repo: `analyze --embeddings`, edit a Python function body, `analyze --embeddings` again. Before, run 2 failed. Now it exits 0, stays on the surgical incremental path with no forced rebuild, the changed symbol's `contentHash` actually changes, and exactly one row per `nodeId` survives.

## Review follow ups, folded in

Reviewing the fix itself turned up three things, all fixed here rather than deferred.

**Embedding rows were destroyed on the blocked path.** Escalating wipes the DB files, and Phase 3.5 restores embedding rows from `cachedEmbeddings`, which `deriveEmbeddingMode` only populates when `meta.stats.embeddings > 0`. A DB holding embedding rows that its meta does not account for therefore had every vector destroyed silently by a rebuild it never asked for. Probe on a 3 file repo: 3 rows before, 0 after, no warning. That turns what used to be a loud crash into quiet data loss, so it needed fixing before merge, not after. The rows are now read before the escalation, which is a plain MATCH and needs no extension, and the log says what is happening. The blocked path test asserts the seeded rows survive exactly once, and that assertion fails without the rescue.

**The catalog is now checked before the extension.** The gate used to load VECTOR first and read `SHOW_INDEXES` only on failure, so every incremental analyze on a machine without VECTOR paid a bounded out of process install attempt plus an "extension unavailable" warning, including repos that never built an embedding index and can never hit this bug. One local catalog read settles that case first. The load is attempted only when an index actually gates DML, or when the catalog cannot be read.

**Removed a dead branch.** `targetConn` is always the module singleton there, so the `isSharedSingletonConn` ternary could never take its second arm.

## Review-agent findings and the Windows gate

The automated review approved with two findings; both are addressed, and chasing the first one surfaced something bigger.

**doctor now live-probes VECTOR (MEDIUM finding).** doctor's VECTOR line was a static platform check, so it printed "available" in exactly the failure state this PR fixes — which is precisely what the issue reporter saw (`VECTOR index: available` next to a crashing analyze). It now runs a real `LOAD` probe, mirrors the #2374 FTS treatment, prints the engine's actual error plus a classified remedy, and the semantic-mode line follows the probe. The remedy classifier is label-aware so the VECTOR row no longer dispenses FTS-only advice like `--repair-fts`.

**Catalog-failure fallback is now pinned (LOW finding).** A `Connection.prototype.query` spy forces `SHOW_INDEXES` to reject over a real DB and asserts the gate falls through to genuinely attempting the extension load.

**The Windows VECTOR gate was stale, and is now removed.** The codebase hard-refused VECTOR on win32 (`platform !== 'win32'` plus an early return in `loadVectorExtension`), based on an early-era report that in-process INSTALL could crash (#1365). Verified directly against the distribution infrastructure:

* the extension server hosts win_amd64 VECTOR artifacts for every 0.18.x extension version — v0.18.0 and v0.18.1 both serve a real 14 MB PE32+ x86-64 DLL (curl + `file` verified)
* the pinned 0.18.2 core resolves its extension directory to 0.18.1 (strace-verified `open()` during `LOAD`), so the pinned version's Windows artifact exists as well
* INSTALL runs in a spawned child process since the out-of-process installer landed, so the historical in-process crash cannot reach the parent any more — a failing installer degrades to "unavailable"

Windows now takes the same runtime path as every other OS: try `LOAD`, install out-of-process when policy allows, fall back to exact scan with an honest diagnostic when it truly fails. The MCP semantic-search lane loses its static platform gate the same way. The vector suites skip visibly only when the extension genuinely cannot load, so the Windows CI lanes on this PR are the enforcement — they now actually exercise VECTOR instead of skipping it by policy.

Verified on this branch: tsc clean, unit 513 files / 10,407 passed, integration 205 files passed (only the pre-existing `fts-extension-e2e.test.ts` environment failure, reproduced identically at the merge base), and the live end to end reproduction still passes on the final build.


magyargergo:
@gitnexus review

magyargergo:
@gitnexus review

magyargergo:
@gitnexus review

magyargergo:
Both findings addressed, in `fcdf7a96` and `671c8796`.

**MEDIUM, doctor's static VECTOR line:** doctor now live probes `LOAD EXTENSION vector` the same way #2374 did for FTS, prints the engine's real error plus a classified remedy, and the semantic mode line follows the probe. The remedy classifier takes a label so the VECTOR row no longer suggests `--repair-fts`, which repairs FTS indexes only.

**LOW, untested catalog-failure fallback:** pinned with a `Connection.prototype.query` spy over a real DB — `SHOW_INDEXES` is forced to reject and the test asserts the gate falls through to genuinely attempting the extension load.

Chasing the doctor finding surfaced a bigger one: the codebase's blanket Windows VECTOR refusal is stale. The extension server ships win_amd64 artifacts for every 0.18.x extension version (curl verified, real PE32+ DLLs; the pinned 0.18.2 resolves its extension dir to 0.18.1, strace verified), and INSTALL has run out of process since the installer child landed, so the #1365 era crash cannot reach the parent. Commit `26f7a98d` removes the gate everywhere — Windows now takes the same load-or-degrade path as every other OS, and the Windows CI lanes on this PR now genuinely exercise VECTOR instead of skipping it by policy.

magyargergo:
All three findings from the follow-up review are addressed in `603cde57`, `a9c2213f` and `37d6688e`.

**MEDIUM, dead pool vector lane:** the pool pre-warm now loads VECTOR at both load sites (`doInitLbug` and `initLbugWithDb`), under the same load-only contract as FTS, tracked by `SharedDB.vectorLoaded`. The new pool test is deliberately discriminating — it closes the writable core adapter first so the pool must open its own fresh Database, where only the pre-warm can make `QUERY_VECTOR_INDEX` legal. Confirmed failing with the exact Catalog exception at the pre-fix tree, passing with the fix.

**MEDIUM, Windows roster:** `incremental-vector-extension-ordering.test.ts` is on the cross-platform roster now. One correction to the review itself: its claim that `lbug-delete-nodes-for-files.test.ts` was also missing was wrong — that suite has been rostered since #2409.

**LOW, CI install cost:** `ensure-fts.ts` now pre-installs VECTOR alongside FTS (the existing extension cache already covers it), and the cross-platform job sets `GITNEXUS_REQUIRE_VECTOR=1` so an unavailable VECTOR is a loud failure rather than a silent skip.

Verified locally on Linux: full unit 513 files / 10,409 passed; integration 205 files passed with only the pre-existing `fts-extension-e2e` environment failure. Windows and macOS cannot be executed locally — this PR's CI lanes are the proof for the roster and workflow changes.

PR #2627, opened 2026-07-22
fix(eval): mount the node prefix for npx and catch nested Claude Code bootstrap noise

abhigyanpatwari:
## Why

Skill-evolution run [29861768554](https://github.com/abhigyanpatwari/GitNexus/actions/runs/29861768554) resolved **0/3 on every task and both arms**. That was not a bad candidate — the harness was broken in two independent ways, each fatal on its own. `runner.py`'s `broken_incumbent_arms` check and the promotion gate both behaved correctly; they were faithfully reporting a broken environment.

Evidence is the run's own artifact:

| symptom | count | field |
| --- | --- | --- |
| `/bin/sh: 1: npx: not found` | **18/18** | `verify_output` |
| `phase changed unauthorized workspace path(s): gitnexus/.claude/.cc-writes` | **13/18** | `error_detail` |

## Bug 1 — the sandbox had `node` but no `npx`/`npm`

`_runtime_mount_args()` bound only the `node` binary to `SANDBOX_NODE`. `npx` and `npm` are not standalone binaries — they are symlinks into `../lib/node_modules/npm/bin/*-cli.js` — so the **install prefix** (carrying both `bin/` and `lib/node_modules`) has to be mounted for them to resolve at all.

On GitHub-hosted images node lives in `/usr/local/bin`, whose prefix is already inside the wholesale `/usr` read-only bind, so `npm`/`npx` came along for free. That is exactly why this stayed invisible. A self-hosted runner's `actions/setup-node` installs into its own tool cache, outside `/usr`, so only the single `node` file was bound. Every task's verify command is `cd gitnexus && npx tsc --noEmit && npx vitest run <test>`, so no run could resolve regardless of model output. It reached the model too: 12 `npm: not found` failures across the session transcripts, with `gitnexus/scripts/build.js` dying on `npm ci` (status 127).

The fix binds the resolved prefix read-only at `/opt/claude/nodejs` and puts its `bin/` on `SANDBOX_PATH`. It is skipped when the prefix already sits inside `/usr`, `/bin`, `/lib` or `/lib64`, so the already-covered case does not widen the mount surface redundantly. `SANDBOX_NODE` is deliberately unchanged — `sanitized_graph.py` and `runner_sessions.py` invoke it directly.

The bind is gated on the property actually required: a working `npx` beside `node` in a real `bin/` directory. An earlier revision derived the prefix as `parent.parent` unconditionally, which read-only mounted an unrelated ancestor whenever node sat elsewhere — `/opt/bin/node` bound **all of `/opt`**, and a bare `<dir>/node` bound `<dir>`'s parent. That second shape is not hypothetical: the pre-existing `test_real_bubblewrap_runs_node_from_outside_the_bound_trees` canary builds exactly it, so `eval-containment-linux` would have silently over-mounted the whole pytest `tmp_path` inside a containment test.

## Bug 2 — the bootstrap-noise exclusion was root-anchored

`workspace_snapshot` tested only `relative.parts[0]` against `WORKSPACE_SNAPSHOT_BOOTSTRAP_NOISE`. Claude Code bootstraps into whatever directory it is **running in**, and the task prompts `cd gitnexus`, so the noise landed at `gitnexus/.claude/.cc-writes` — whose `parts[0]` is `"gitnexus"`, so it was never excluded. This is the same defect #2615 meant to fix; that change added the set but kept the root-anchored match, so it only ever covered workspace-root noise. The same code path also guards the review phase (`runner.py:499`), so review arms hit it as `review-evidence-invalid`.

Matching the whole set at any depth would be wrong — it also contains `package.json`, `package-lock.json`, `node_modules` and the `.env` family, and both `gitnexus/package.json` and `gitnexus/.claude/settings.local.json` are real tracked files whose edits must still be caught. So the root-anchored rule is untouched and one narrow rule matches `.cc-writes` inside a `.claude` directory at any depth.

Scoped to `.cc-writes` alone, deliberately. An earlier revision also excluded `agents` and `commands` on the theory they might appear nested; only `.cc-writes` ever was. Every excluded name is a blind spot: once a `.claude` directory exists, anything written under an excluded entry is invisible to the check, and Claude Code loads `.claude/agents` relative to its cwd — which these tasks point at `gitnexus/`. A planning phase could otherwise plant a definition the work phase reads, with no evidence in the boundary check.

## Verified on the actual self-hosted runner

The gated real-Bubblewrap canaries cannot run in a dev container (no unprivileged user namespaces, even under `sudo`), so every claim below was executed on the real runner over SSM:

- Guard predicate against the runner's real `which("node")` → `PREFIX BOUND = True`
- Real bwrap, before vs after: `npx: NOT FOUND` → `npx FOUND at /opt/claude/nodejs/bin/npx, 10.9.3`
- Full eval suite with `GITNEXUS_REQUIRE_BWRAP_CANARY=1` → **335 passed, 4 skipped**
- All five real-bwrap canaries pass, including the new `test_real_bubblewrap_runs_npx_from_outside_the_bound_trees`
- CI containment selection (the exact `eval-containment-linux` command) → 121 passed
- The real production command `cd gitnexus && npx tsc --noEmit` inside a real sandbox (read-only workspace, unshared network, PID namespace) → **passes**

Locally: 325 passed, 12 skipped; `ruff check` and `ruff format --check` clean.

## Known follow-up, not fixed here

Testing on the runner surfaced a **third, independent blocker** that `npx: not found` had been masking by short-circuiting the `&&` chain. With npx working, verify gets further and dies:

```
EROFS ... /workspace/gitnexus/node_modules/.vite-temp/vitest.config.ts.timestamp-*.mjs
```

Vite transpiles a TypeScript config before loading it, and the dependency `node_modules` is a read-only mount. This blocks the authored-test verify **and the hidden oracle** (same `run_verify(..., read_only_workspace=True)` path), so `resolved` will still be 0/N until it is addressed.

It is pre-existing and independent of this PR: it reproduces at the base commit with these changes stashed and npx bypassed entirely via `./node_modules/.bin/vitest`.

No fix is included because the obvious one does not generalize. Overlaying `--tmpfs <target>/.vite-temp` makes the complete verify pass (15/15 tests), but bwrap cannot create a mountpoint inside a read-only bind, so it only works where `.vite-temp` already exists on the host — it did for `gitnexus/node_modules`, not for `gitnexus-shared/node_modules`, and `npm ci` never creates it. The runner has bubblewrap 0.9.0, which predates `--tmp-overlay`/`--overlay-src`. Shipping a fix that passes by accident would be worse than none. Options, smallest first: create the empty `.vite-temp` in each dependency `node_modules` at staging time; upgrade bwrap to >= 0.10 and overlay the whole mount; or make the vitest config non-TS.

## Tests

Added to `test_proposer_sandbox.py` (the canary must live there — `test_workflow_bench.py` pins the set of files carrying the canary marker):
- tool-cache layout binds its prefix and keeps the `SANDBOX_NODE` bind
- already-bound prefix (`/usr/local`) emits no redundant bind
- unrecognized layout (bare `<dir>/node`) binds nothing extra
- `<prefix>/bin/node` without a working `npx` binds nothing extra
- gated real-bwrap canary running `npx --version` inside the sandbox

Added to `test_runner_hardening.py`:
- nested `gitnexus/.claude/.cc-writes` is ignored
- an excluded directory is not descended into
- nested `gitnexus/.claude/settings.local.json` edits are still rejected
- nested `gitnexus/package.json` edits are still rejected
- writes under a pre-existing nested `.claude/agents` are still caught

Exactly one pre-existing line changed across both test files: the hard-coded `SANDBOX_PATH` probe assertion, now derived from the constant so the two cannot drift. No test was weakened or removed.

### DIFF
--- a/eval/tests/test_proposer_sandbox.py
+++ b/eval/tests/test_proposer_sandbox.py
@@ -21,6 +21,8 @@ from workflow_bench.proposer_sandbox import (
     MAX_BUNDLE_BYTES,
     MAX_EVIDENCE_FILE_BYTES,
     SANDBOX_NODE,
+    SANDBOX_NODE_PREFIX,
+    SANDBOX_PATH,
     SANDBOX_PYTHON3,
     SANDBOX_SHELL_PREFIX,
     SANDBOX_USER_SKILLS,
@@ -165,7 +167,7 @@ def test_sandbox_command_has_minimal_mounts_and_no_host_root_bind(tmp_path: Path
             check=False,
         )
         assert probe.returncode == 0, probe.stderr
-        assert probe.stdout == "/home/agent|/opt/claude:/usr/local/bin:/usr/bin:/bin"
+        assert probe.stdout == f"/home/agent|{SANDBOX_PATH}"
 
         # The evidence-provenance.mjs plan-writer's PATH-scan trusts a Python 3
         # candidate only if it (and its directory) is owned by root or by the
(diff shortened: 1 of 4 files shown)


magyargergo:
@gitnexus review

PR #2628, opened 2026-07-22
fix(python): resolve calls through constructor-injected fields

abhigyanpatwari:
## Summary

Python indexes now retain the declared type of constructor-injected fields, so calls such as `self.service.method()` resolve to the receiver's real method instead of disappearing from the call graph or landing on a same-named decoy. This restores meaningful caller counts in `context()` and `impact()` for the common annotated dependency-injection pattern.

The binding is deliberately conservative: it accepts annotated constructor parameters and explicit field annotations, hoists the resulting type to class scope, and does not guess from arbitrary unannotated expressions. Nested functions and classes are excluded.

## Verification

- `npm run build`
- `npx tsc --noEmit`
- Focused Python ingestion/resolution suite: 309 tests passed.
- Exact-index CLI smoke test: all four production callers resolved to the intended method at confidence `0.85`; no production call resolved to the decoy.
- Full suite: 14,793 tests passed and 62 skipped. Nine `fts-extension-e2e` tests could not run because the native FTS extension was unavailable and its network install failed in this environment; no Python ingestion or resolver test failed.

## Risk and rollback

The behavior is confined to the Python provider and only emits bindings for evidence-backed `__init__` assignments. Existing capture goldens and negative cases cover drift and false-positive boundaries. Rollback is a single commit revert followed by re-indexing.


magyargergo:
@gitnexus review

magyargergo:
@gitnexus review

abhigyanpatwari:
> Reassigning a field to two different parameter-typed values inside one `__init__` silently keeps the first type.
>
> A field assigned only inside an `if`/`try`/`for`/`while` branch of `__init__` is bound as if it were always constructed with that type.

Addressed in `0a12d1b8`:

- repeated inferred writes now use the final assignment, while an explicit field annotation still takes precedence;
- branch-controlled writes under `if`, `try`, `for`, `while`, and `match` are excluded from class-scope synthesis;
- parameter-derived fields retain `parameter-annotation` provenance instead of being labeled as explicit annotations;
- explicit class-body field annotations now outrank parameter-derived constructor bindings, with collision coverage.

Added focused regressions for reassignment order, each conditional shape, provenance, and class-scope collisions. The 73-test Python capture/resolution set, typecheck, build, and capture benchmark guard pass. The full local suite remains blocked by the known container LadybugDB/native-addon initialization failure family.

koriyoshi2041 (approved):
Verified the constructor-field binding path at `2eea0ff0`. The implementation stays conservative: it requires a declared parameter or field type, hoists only the synthetic instance-field binding to class scope, skips nested/control-flow assignments, and preserves explicit class-body annotations over inferred constructor assignments. The receiver-constrained integration fixture also guards against the same-named free-function fallback.

Focused unit coverage passed locally (19/19). The integration import was blocked on my existing checkout's js-yaml 4.x install (`CORE_SCHEMA.withTags` is a js-yaml 5 API); hosted CI at this exact head is green, so I do not attribute that environment mismatch to this PR.

PR #2630, opened 2026-07-22
fix(eval): give vitest a writable .vite-temp inside read-only dependency mounts

abhigyanpatwari:
Follow-up to #2627, which fixed the `npx: not found` half of skill-evolution run [29861768554](https://github.com/abhigyanpatwari/GitNexus/actions/runs/29861768554). Testing that fix on the actual self-hosted runner surfaced a third, independent blocker that `npx: not found` had been hiding.

## The problem

Every task verify command and every hidden oracle ends in `npx vitest run <test>`, and both go through `run_verify` with `read_only_workspace=True`. Vite transpiles a TypeScript config by writing `<node_modules>/.vite-temp/<config>.timestamp-*.mjs` before it loads anything, so against a read-only dependency mount vitest dies before a single test executes:

```
EROFS ... /workspace/gitnexus/node_modules/.vite-temp/vitest.config.ts.timestamp-*.mjs
```

Because it blocks the **hidden oracle** as well as the authored-test verify, `resolved` stays 0/N without this — the evolution loop would keep reporting a clean sweep of failures with no promotion.

It is pre-existing, not a regression from #2627. Until that PR the verify command died at `npx: not found`, short-circuiting the `&&` chain before vitest ran. Confirmed by reproducing it at #2627's merge base with npx bypassed entirely via `./node_modules/.bin/vitest`, which makes it independent of the node-prefix mount.

## Why the obvious fix does not work

Overlaying `--tmpfs <target>/.vite-temp` is the right shape, but bwrap cannot create a mount point inside an already-read-only bind — the same constraint that put `SANDBOX_NODE` under `/opt/claude` rather than `/usr/local/bin`. It therefore only works when the directory already exists in the mounted bytes, and `npm ci` never creates it. On the runner it happened to exist under `gitnexus/node_modules` and not under `gitnexus-shared/node_modules`, which is exactly the kind of accidental pass that looks like a fix and is not.

The directory also cannot be `mkdir`'d into the dependency snapshot after capture: the snapshot is digest-bound, and `validate_dependency_binding` fails closed on drift.

## The fix

Capture the empty `.vite-temp` **during dependency capture**, before the manifest and both dependency digests are computed, so it is part of the snapshot rather than an untracked mutation of it. The sandbox then overlays a tmpfs on exactly that path.

Everything else in the dependency mount, and the whole workspace, stays read-only. The overlay lives only inside the sandbox, so it never reaches the host clone the credited patch is captured from. It is scoped to dependency mounts whose target basename is `node_modules`, so hidden-oracle and skill mounts stay wholly read-only with no writable island inside them.

`bubblewrap` 0.9.0 on the runner predates `--tmp-overlay`/`--overlay-src`, so a writable ephemeral overlay over the whole mount is not available.

## Digest impact

This shifts `sandbox_dependency_content_digest` and `sandbox_dependency_manifest_digest`, so promotion evidence recorded before this change is no longer comparable. That is already true of any harness fix that changes what the sandbox exposes, and the affected runs are the ones that resolved 0/3.

## Verification

On the self-hosted runner, through the real path — `TaskAssetCache.prepare` → `stage_task_assets` → `prepare_sandbox` → `run_verify` with the actual `trivial-version-alias` verify string:

```
verify : cd gitnexus && npx tsc --noEmit && npx vitest run test/unit/cli-commands.test.ts
PASSED : True
EROFS  : False | npx-missing: False
         Test Files  1 passed (1)
              Tests  15 passed (15)
```

Full eval suite on the runner with `GITNEXUS_REQUIRE_BWRAP_CANARY=1`: **337 passed, 4 skipped**. Locally: 329 passed, 12 skipped; `ruff check` and `ruff format --check` clean.

## Tests

`test_proposer_sandbox.py`
- a `node_modules` mount gets the tmpfs overlay, ordered **after** the read-only bind so the bind cannot mask it
- a non-`node_modules` mount (hidden oracle) gets no overlay

`test_task_assets.py`
- a `node_modules` dependency snapshot captures `payload/.vite-temp`, and it lands in the manifest entries and on disk
- a non-`node_modules` dependency is captured byte-for-byte as declared, with no injected directory

Note the `task_assets.py` diff includes a `ruff format` reflow of pre-existing lines; `git diff -w` shows the only semantic changes are the import, the capture block, and the new `_SnapshotBuilder.ensure_directory`.

### DIFF
--- a/eval/tests/test_proposer_sandbox.py
+++ b/eval/tests/test_proposer_sandbox.py
@@ -22,6 +22,7 @@ from workflow_bench.proposer_sandbox import (
     MAX_EVIDENCE_FILE_BYTES,
     SANDBOX_NODE,
     SANDBOX_NODE_PREFIX,
+    VITE_TEMP_DIR,
     SANDBOX_PATH,
     SANDBOX_PYTHON3,
     SANDBOX_SHELL_PREFIX,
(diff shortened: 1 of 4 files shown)


abhigyanpatwari:
Pushed `37d4f791`: the containment (ubuntu) job caught a real bug in the first revision. The tmpfs overlay keyed on the target basename being `node_modules`, which also matched the trusted GitNexus runtime mount at `/opt/gitnexus/node_modules` — whose source carries no `.vite-temp`, so bwrap failed with `Can't mkdir .../node_modules/.vite-temp: Read-only file system`.

Fixed by gating the overlay on the mount **source** actually containing the directory rather than the target name. That's correct by construction: a tmpfs can only overlay a mount point that already exists in the bound bytes, and `task_assets.py` captures `.vite-temp` only into dependency-snapshot `node_modules`. Added a regression test for the runtime-mount shape.

Verified on the self-hosted runner: the exact failing test `test_real_bubblewrap_runtime_mount_imports_cli_without_exposing_checkout` now passes, and the full containment selection is 124 passed / 4 skipped.

PR #2200, opened 2026-06-14
feat: add Spring DI resolver for @Autowired List<T> injection

ChunxueLi:
## Problem

Spring projects commonly use `@Autowired List<T>` to inject multiple implementations:

```java
@Component
public class SignInventoryValidationDecorator {
    @Autowired
    private List<SignInventoryValidationDecorator> subDecorators;
    
    public void execute() {
        for (SignInventoryValidationDecorator sub : subDecorators) {
            sub.execute();
        }
    }
}
```

GitNexus cannot resolve these injection points, breaking call chain tracing at the injection site.

## Solution

New pipeline phase `spring-di` that:
1. Scans `@Autowired List<T>` / `Set<T>` / `Collection<T>` field declarations
2. Resolves implementations via existing `EXTENDS`/`IMPLEMENTS` relationships
3. Creates `INJECTS` edges in the graph (Class -> Implementation)

## Implementation

**New file**: `spring-di.ts` (198 lines)
- Scans Java source files for `@Autowired` annotations
- Extracts generic type parameter `T`
- Queries graph for all classes implementing `T`
- Creates `INJECTS` edges with confidence 0.8

**Modified files**:
- `pipeline-phases/index.ts`: Export new phase
- `pipeline.ts`: Register phase after `mro`

## Effect

- **Before**: 0 `INJECTS` edges (Spring DI invisible)
- **After**: 82 `INJECTS` edges in 280K-node Java repo

## Limitations

- Only handles `@Autowired List<T>` pattern (most common)
- Does not handle `@Qualifier` for specific bean selection
- Does not handle `@ConditionalOnProperty` conditional injection
- Relies on grep scanning (GitNexus Java parser does not preserve generic parameters)

## Testing

Verified on a 280,951-node Java repository:
- 21 `@Autowired List<T>` fields detected
- 82 `INJECTS` edges created
- Spring DI call chains now traceable

---

*Contributed by [卫宁健康科技集团](https://www.winning.com.cn) — WiNEX门诊医生站团队*

PR #2632, opened 2026-07-22
feat(spring): resolve constructor and standard injection

### LINKED ISSUE #2414: feat(spring): resolve constructor and standard injection
Parent tracking issue: #1225

Status: remaining work

Goal
Model normal Spring dependency injection beyond the narrow collection injection support that exists today.

Scope
Resolve constructor injection, single bean field injection, setter or method injection, Autowired and Inject annotations, Qualifier names, Primary preference, and named bean references where static analysis has enough information.

Acceptance criteria
1. Constructor injected dependencies connect consumers to likely Spring bean providers.
2. Field and setter injection are represented for standard single bean cases.
3. Qualifier and Primary metadata are used when they remove ambiguity.
4. Ambiguous injections are surfaced with uncertainty instead of pretending there is one exact target.
5. Tests cover interface dependencies, concrete dependencies, and ambiguous multiple implementation cases.

ShiningXu:
## Summary

- capture Java Spring bean and injection metadata during scope resolution
- resolve constructor, field, setter, and method injection for `@Autowired` and `@Inject`
- apply static `@Qualifier` / `@Named` names and unique `@Primary` preference
- represent unresolved single-bean ambiguity as multiple low-confidence candidates
- preserve the existing collection-injection fallback when annotation FQN resolution is ambiguous
- follow Spring default bean naming for digit-bearing class names and empty-parenthesis stereotypes

## Impact

Spring applications now produce `INJECTS` relationships for standard single-bean dependency injection, including interface and concrete dependencies. Dynamic qualifiers fail closed instead of being treated as unqualified injection.

When a requested simple name denotes both a Class and an Interface, resolution intentionally fails closed for both single and collection injection sites because the graph-only phase does not have import-aware disambiguation.

## Validation

- impacted Spring/Java resolver tests — 352 passed, 2 skipped
- `GITNEXUS_BENCH=1 npx vitest run test/integration/spring-di-benchmark.test.ts` — 3 passed
- `npx tsc --noEmit` — passed
- pre-commit ESLint, Prettier, and typecheck — passed
- `npm test` — 14,768 passed; 9 unrelated environment/worktree failures on macOS (`/var` canonicalization, Linux `/proc/self/fd`, and existing vendored grammar build directories)
- GitNexus `detect-changes` — LOW risk, 0 affected processes

Closes #2414
Part of #1225


gitnexus/src/core/ingestion/languages/java/spring-di.ts
```diff
@@ -0,0 +1,347 @@
...
+        ? classNode.properties.frameworkAnnotations.filter(
+            (annotation): annotation is string => typeof annotation === 'string',
+          )
+        : [];
+      if (frameworkAnnotations.length > 0) {
+        const names = new Set<string>();
+        let explicitBeanName: string | undefined;
+        let hasDynamicBeanName = false;
+        for (const annotation of fact.classAnnotations) {
+          if (resolveFact(annotation, SPRING_BEAN_STEREOTYPES) !== undefined) {
```
magyargergo: Can you please rewrite this to use iterative approach over recursion?

gitnexus/src/core/ingestion/languages/java/spring-di.ts
```diff
@@ -0,0 +1,347 @@
...
+/** Attach resolved, framework-private DI metadata to Class nodes. */
+export function attachJavaSpringDiMetadata(
+  graph: KnowledgeGraph,
+  parsedFiles: readonly ParsedFile[],
+  nodeLookup: GraphNodeLookup,
+  indexes: ScopeResolutionIndexes,
+): void {
+  const resolveAnnotation = createSpringAnnotationNameResolver(indexes);
+  const propertiesByOwner = new Map<string, string[]>();
+  for (const relationship of graph.iterRelationshipsByType('HAS_PROPERTY')) {
```
magyargergo: Could we use scope resolution? That should be O(1)

gitnexus/src/core/ingestion/languages/java/spring-di.ts
```diff
@@ -0,0 +1,347 @@
...
+  return dependencies;
+}
+
+/** Capture Spring-relevant Java syntax while the worker already owns the AST. */
+export function captureJavaSpringDiFacts(
+  root: SyntaxNode,
+  filePath: string,
+): readonly JavaSpringDiClassFact[] {
+  const facts: JavaSpringDiClassFact[] = [];
+  for (const classNode of root.descendantsOfType('class_declaration')) {
```
magyargergo: Can you please move them to captures.ts? So we don't need to recompute them over and over again

magyargergo (changes requested):
Thank you for submitting this PR! I just mostly found performance concerns. Could you please benchmark your solution as well? 🙏

gitnexus/src/core/ingestion/languages/java/spring-di.ts
ShiningXu: ok ,i'm already fix it

gitnexus/src/core/ingestion/languages/java/spring-di.ts
```diff
@@ -0,0 +1,375 @@
...
+        fact.injectionSites.filter((site) => site.kind === 'field').map((site) => site.memberName),
+      );
+      for (const fieldName of capturedFieldNames) {
+        for (const { def } of classScope.bindings.get(fieldName) ?? []) {
+          if (def.ownerId !== classDef.nodeId) continue;
+          const propertyId = resolveDefGraphId(parsed.filePath, def, nodeLookup);
+          if (propertyId === undefined) continue;
+          const property = graph.getNode(propertyId);
+          if (property?.label === 'Property') {
+            property.properties[SPRING_DI_CAPTURED_FIELD_PROPERTY] = true;
```
abhigyanpatwari: **[MEDIUM — regression] Wildcard-import files lose all `@Autowired` collection-injection edges.**

`SPRING_DI_CAPTURED_FIELD_PROPERTY` is stamped here for **every** syntactically-captured field (`hasRelevantAnnotation` = simple-name match on `Autowired`/`Inject`), *before* and *independent of* FQN resolution. It suppresses the legacy collection matcher at `di-extractors/spring.ts:284`. But an edge is only emitted if semantic resolution succeeds — and `createSpringAnnotationNameResolver` returns `undefined` when a file has 2+ wildcard imports (`frameworks/spring/bean-candidates.ts:166`, `wildcardPackages.size !== 1`). When they diverge, the field is suppressed *and* produces nothing.

Reproduced end-to-end on a built checkout of the head — identical `@Autowired private List<IFoo> foos;`:
```
import org.springframework.beans.factory.annotation.Autowired;  → Consumer->FooA, Consumer->FooB
import java.util.*;
import org.springframework.beans.factory.annotation.*;           → (no edges)
```
A second probe confirmed it's a regression: on the wildcard node the retained `springDiFieldMatcher(node)` still matches (pre-PR that emitted both edges), yet the captured flag is set and 0 INJECTS edges exist. Broader collateral: when a `@Service` stereotype *itself* arrives via wildcard, `frameworkAnnotations` never gets set, so the whole class disappears from DI (no constructor/field/setter injection either). No test uses wildcard imports.

**Fix:** gate the marker on a *successfully-emitted* match rather than on syntactic capture — accumulate resolved field names inside the site loop and stamp `SPRING_DI_CAPTURED_FIELD_PROPERTY` only for those after `matches` is built, so the legacy fallback still fires when the new path yields nothing for a field.

gitnexus/src/core/ingestion/languages/java/spring-di.ts
```diff
@@ -0,0 +1,375 @@
...
+  if (literal === null) return undefined;
+  try {
+    return JSON.parse(`"${literal[1]}"`) as string;
+  } catch {
+    return undefined;
+  }
+}
+
+function defaultBeanName(className: string): string {
+  if (className.length < 2 || className[1] !== className[1].toUpperCase()) {
```
abhigyanpatwari: **[LOW] `defaultBeanName` diverges from Spring's `Introspector.decapitalize` for digit-second-char names.**

`className[1] !== className[1].toUpperCase()` treats a digit as uppercase (`'1'.toUpperCase() === '1'`), so the first char isn't lowercased. Probe-confirmed divergences: `S3Client`→`S3Client`, `H2Repository`→`H2Repository`, `X509Cert`→`X509Cert` — where Spring yields `s3Client`/`h2Repository`/`x509Cert`. Since this default name feeds `@Qualifier` matching (`pipeline-phases/di.ts:203`), `@Qualifier("s3Client")` against an unnamed `S3Client` provider silently mismatches and the edge drops. Common in AWS/H2/versioned-API code.

**Fix:** use Java's actual rule — leave unchanged only when both `name[0]` and `name[1]` are uppercase *letters*, e.g. `className[1] === className[1].toLowerCase()` as the not-uppercase test (a digit equals its own lowercase → decapitalize). Guard `length === 0` while here (`defaultBeanName('')` currently throws, though unreachable today).

gitnexus/src/storage/parse-cache.ts
```diff
@@ -61,7 +61,7 @@ import type { ParseWorkerResult } from '../core/ingestion/workers/parse-worker.j
 // JLS 13.1 immediate-host chains (#2555).
 // v18: Worker$N anonymous bodies. v17: callable-value-flow operand identity.
 // v16: direct callee identity.
-const SCHEMA_BUMP = 20;
+const SCHEMA_BUMP = 21;
```
abhigyanpatwari, line 66: **[LOW — DoD] `SCHEMA_BUMP` bumped 20→21 without a `v21:` changelog entry.**

The comment block just above documents every prior bump (v16–v20) and this file's own contract says any `ParsedFile` serialization change "MUST bump `SCHEMA_BUMP`" — the *why* is the convention. The bump itself is correct and load-bearing (it's what invalidates pre-PR ParsedFiles so unchanged files re-parse and gain `springDiFacts` on the first run); only the note is missing.

**Fix:** add e.g. `// v21: Java Spring DI facts (constructor/field/method injection sites + bean/@Primary provider metadata) added to the capture side-channel.`

gitnexus/src/core/ingestion/languages/java/spring-di.ts
```diff
@@ -0,0 +1,375 @@
...
+        let explicitBeanName: string | undefined;
+        let hasDynamicBeanName = false;
+        let primary = false;
+        for (const annotation of fact.classAnnotations) {
+          const resolved = resolveFact(annotation);
+          if (resolved === undefined) continue;
+          if (SPRING_BEAN_STEREOTYPES.has(resolved)) {
+            if (annotation.text.includes('(')) {
+              const staticName = staticStringArgument(annotation.text);
+              if (staticName === undefined) hasDynamicBeanName = true;
```
abhigyanpatwari: **[LOW] Empty-parenthesis stereotype (`@Service()`, `@Component()`) yields a nameless provider.**

For `@Service()`, `annotation.text.includes('(')` is true, `staticStringArgument('@Service()')` returns `undefined` (no string literal), so `hasDynamicBeanName = true` and the default-name branch is skipped → `provider.names = []`. Probe-confirmed: `@Service` → `["fooService"]` but `@Service()` → `[]`. In Spring `@Service()` ≡ `@Service` (default name applies), so `@Qualifier("fooService")` against such a bean mismatches and drops the edge.

**Fix:** treat an empty/whitespace-only argument list as "no explicit name" (fall through to `defaultBeanName`); set `hasDynamicBeanName` only for a genuinely non-empty non-literal value.

gitnexus/src/core/ingestion/pipeline-phases/di.ts
```diff
@@ -96,174 +96,187 @@ export const diPhase: PipelinePhase<DIOutput> = {
...
+        candidate.targetTypeName,
+      );
+      const interfaceEntry = resolveIndexedName(
+        interfacesByLanguage.get(candidate.language),
+        candidate.targetTypeName,
+      );
+      if (
+        classEntry === AMBIGUOUS ||
+        interfaceEntry === AMBIGUOUS ||
+        (classEntry !== undefined && interfaceEntry !== undefined)
```
abhigyanpatwari, line 181: **[LOW — behavioral change] Class-and-Interface same-name fail-closed alters pre-existing collection behavior.**

The old phase indexed only interfaces, so `@Autowired List<FooService>` resolved to a unique `interface FooService` and fanned out regardless of any same-named class. This gate now skips (`ambiguousSkipped++`) when a name resolves to **both** a Class and an Interface — so in a repo with `interface FooService` plus an unrelated `class FooService` in another package, that injection emitted edges before and emits nothing now.

The fail-closed choice is defensible (a graph-only phase can't tell which type the site meant) and the concrete-class resolution this index enables is a real improvement — it only bites on true simple-name collisions. But it's a behavior change not called out in the PR description and covered only by the new positive unit test, not this regression path. Please document it (and ideally add a regression fixture).

abhigyanpatwari (commented):
## Review — Spring standard dependency injection

Thanks for this — the framework-neutral `DiResolver` seam, the fail-closed dynamic-qualifier handling, and the ambiguity-as-multiple-low-confidence-edges model are all the right shape, and coverage is solid. I verified the risky plumbing directly and it holds: the `DI_MATCHERS`→`DI_RESOLVERS` rename is clean, `INJECTS` is already a graph-wide rel type so all edges are delete-all'd + re-extracted each incremental run, and the ephemeral injection metadata is correctly invalidated by the `SCHEMA_BUMP` 20→21 bump (which gates both the parse cache and the durable ParsedFile store), so **no `INCREMENTAL_SCHEMA_VERSION` bump is needed** — nicely handled.

Findings below (verified with empirical probes on a built checkout of the head; two independent review passes reproduced #1 and #2). One regression is worth fixing before merge; the rest are low-cost cleanups clustered in the same bean-name / annotation-resolution surface.

| # | Sev | Finding |
|---|-----|---------|
| 1 | **MEDIUM** | Field-capture suppresses the legacy collection fallback before FQN resolution — files with 2+ wildcard imports lose **all** `@Autowired` DI edges (regression vs. #2200) |
| 2 | LOW | `defaultBeanName` mis-decapitalizes digit-second-char class names (`S3Client`→`S3Client` vs Spring `s3Client`) → `@Qualifier` drops the edge |
| 3 | LOW | `SCHEMA_BUMP` bumped 20→21 with no matching `v21:` changelog line (repo convention) |
| 4 | LOW | `@Service()` / `@Component()` empty parens → nameless provider → `@Qualifier(defaultName)` drops the edge |
| 5 | LOW | New Class-and-Interface same-name fail-closed changes pre-existing collection-injection behavior |

**Residual recall gaps (fail-closed, not defects, worth a note):** `Optional<T>` / `ObjectProvider<T>` / `Provider<T>` / `T[]` injection types are silently skipped despite the "standard injection" framing; `@Component` records and enums aren't captured (`captures.ts` handles only `class_declaration`); the recall-first fallback emits 0.9-confidence edges to structural non-bean classes (deliberate, documented).

**Recommendation: request changes** — finding #1 is the blocker; #2–#5 can ride along.

See inline comments for repro/fix on each.

magyargergo:
@ShiningXu could we also include kotlin here please? 🙏

gitnexus/src/core/ingestion/di-extractors/index.ts
```diff
@@ -1,61 +1,74 @@
...
- * A per-language field-injection matcher: given a `Property` node, return the
- * parsed DI match or `null` when the field is not container-injected. The
- * matcher receives the whole node (not pre-plucked fields) so the shared
- * phase stays ignorant of which properties matter.
- */
-export type DiFieldMatcher = (node: GraphNode) => DiFieldMatch | null;
+/** Provider metadata used by the shared resolver without naming a framework. */
+export interface DiProviderMatch {
+  /** Whether the class is statically known to be container-managed. */
+  isBean: boolean;
```
magyargergo: This is spring specific we need to make this file as generic as possible because this will be adapted by laravel/symfony in the php word.

magyargergo (changes requested):

magyargergo:
@ShiningXu could we please benchmark your solution? 🙏

gitnexus/src/core/ingestion/languages/java/spring-di.ts
```diff
@@ -0,0 +1,392 @@
+import type { ParsedFile, ScopeId } from 'gitnexus-shared';
```
magyargergo: could we please abstract out the common parts with kotlin? 🙏

magyargergo (changes requested):

magyargergo (approved):

PR #2636, opened 2026-07-22
fix(lbug): scale the buffer-pool budget by the OS page-size granule ratio (#2631)

### LINKED ISSUE #2631: [Bug]: Error: COPY failed for Function: Buffer manager exception: Unable to allocate memory! The buffer pool is full and no memory could be freed!
### Area

### Summary

Analysis failed: COPY failed for Function: Buffer manager exception: Unable to allocate memory! The buffer pool is full and no memory could be freed!
Error: COPY failed for Function: Buffer manager exception: Unable to allocate memory! The buffer pool is full and no memory could be freed!

### Context

Index the vllm-ascend repository.

### Expected behavior

Repository indexed successfully

### Actual behavior

  Analysis failed: COPY failed for Function: Buffer manager exception: Unable to allocate memory! The buffer pool is full and no memory could be freed!
Error: COPY failed for Function: Buffer manager exception: Unable to allocate memory! The buffer pool is full and no memory could be freed!
    at file:///root/.local/npm-global/lib/node_modules/gitnexus/dist/core/lbug/lbug-adapter.js:800:19
    at copyCsvWithRetry (file:///root/.local/npm-global/lib/node_modules/gitnexus/dist/core/lbug/lbug-adapter.js:780:13)
    at async copyNodeCSVs (file:///root/.local/npm-global/lib/node_modules/gitnexus/dist/core/lbug/lbug-adapter.js:798:9)
    at async loadGraphToLbug (file:///root/.local/npm-global/lib/node_modules/gitnexus/dist/core/lbug/lbug-adapter.js:935:9)
    at async runFullAnalysis (file:///root/.local/npm-global/lib/node_modules/gitnexus/dist/core/run-analyze.js:1359:13)
    at async analyzeCommandImpl (file:///root/.local/npm-global/lib/node_modules/gitnexus/dist/cli/analyze.js:1050:24)
    at async analyzeCommand (file:///root/.local/npm-global/lib/node_modules/gitnexus/dist/cli/analyze.js:509:9)
    at async Command.<anonymous> (file:///root/.local/npm-global/lib/node_modules/gitnexus/dist/cli/lazy-action.js:60:9)

### Steps to reproduce

1. git clone https://github.com/vllm-project/vllm-ascend.git
2. cd vllm-ascen
3. GITNEXUS_LBUG_EXTENSION_INSTALL=never npx gitnexus analyze --worker-timeout 60

### Environment

[root@hostname-2pbfv triton]# gitnexus --version
1.6.10-rc.83
[root@hostname-2pbfv triton]# node -v
v22.23.1
[root@hostname-2pbfv triton]# npm -v
11.5.2
[root@hostname-2pbfv triton]# npx -v
11.5.2

### Logs / screenshots

_No response_

magyargergo:
Fixes #2631.

## What was happening

Indexing vllm-ascend on gitnexus 1.6.10-rc.83 died during the bulk COPY:

```
Analysis failed: COPY failed for Function: Buffer manager exception: Unable to
allocate memory! The buffer pool is full and no memory could be freed!
```

The reporter's exact command and version succeed on a stock x86 machine, so this is environment dependent, and the environment turned out to be the interesting part.

## Root cause: the pool budget is billed per OS page, and our sizing assumed 4 KiB pages

All measurements and code reading below were done against the engine source at v0.18.2, and the buffer manager is unchanged in the v0.18.3 this repo now pins.

LadybugDB frames are a compile time 4 KiB on every shipped binary (`LBUG_PAGE_SIZE_LOG2 = 12`). What scales with the host is the discard granule: `vm_region.cpp` sets `discardGranuleSize = max(frameSize, osPageSize)`. Two consequences follow on any kernel with pages larger than 4 KiB:

* `claimFrame` charges the whole granule to the pool budget the moment its first 4 KiB frame becomes resident, so scattered access is billed at up to `osPageSize / 4096` times its real bytes. On a 64 KiB page kernel that is 16 frames per granule and up to 16x billing.
* `releaseFrame` refunds the budget only when the last frame of a granule leaves, and `BufferManager::reserve` measures eviction progress in refunded bytes. Whole eviction passes can evict frames yet report zero reclaimed, and after three such passes the engine throws exactly the reported error.

Our adaptive pool sizing (#2582) is blind to all of this: 4 KiB per element, a 256 MiB COPY floor, and a 2 GiB cap, all calibrated on 4 KiB page x86.

Measured with the reporter's exact command and version: vllm-ascend needs a (128, 256] MiB pool at 4 KiB pages. 64 MiB and 128 MiB reproduce the reporter's byte identical error; 256 MiB and the adaptive 576 MiB pool succeed. The reporter is on an Ascend CANN host — aarch64 machines whose EulerOS/openEuler kernels default to 64 KiB pages, a host kernel property containers cannot mask — so their effective budget was divided by up to 16 and their evictor under reported. Apple Silicon macOS (16 KiB pages) has the same mechanism at 4x.

## The fix

`granuleRatio = max(1, osPageSize / 4096)` now scales the analyze-path pool sizes in `lbug-config.ts`: the per element estimate, the COPY safety floor, and the cap the per-run hint is clamped against (still bounded by 80 percent of RAM). The hintless default (MCP server, doctor, native-check) deliberately stays at the unscaled min(2 GiB, 80% RAM): the pool is an eager native allocation at DB open, so scaling the global default would revert the #2557 OOM cap on every 16 KiB/64 KiB host. On 4 KiB hosts the ratio is 1 and behavior is byte identical, which the test suite proves by pinning the existing sizing expectations to an explicit 4096 page size — this also stops those tests from silently drifting on 16 KiB Apple Silicon runners. `GITNEXUS_LBUG_BUFFER_POOL_SIZE` keeps absolute precedence and `0` still restores the native default.

Two diagnostics ride along:

* the pool exhaustion error at both COPY sites now appends an actionable remedy naming the env knob, the effective pool size, and the page size amplification when the ratio is above 1 — previously the operator got a raw binder message with nothing to act on;
* `doctor` prints the hintless pool next to the existing page size line (`pool size 2048 MiB`, with an `(env override)` marker when `GITNEXUS_LBUG_BUFFER_POOL_SIZE` supplied the value, and `native 80% of RAM` instead of `0 MiB` for the `0` sentinel), and the stale comment claiming pool exhaustion is "a sizing problem, not a page-size one" is corrected — that framing inverted when #2582 made pool size a function of a page size blind estimate.

## Verification

* Full unit suite: 513 files, 10,420 passed. Integration: 206 passed, with only the known pre existing `fts-extension-e2e` environment failure.
* 13 new test cases: 64 KiB scales x16 and 16 KiB scales x4 (estimate, floor, and cap, still RAM bounded), undetectable page size falls back to ratio 1, env precedence intact, remedy formatting for both ratio classes.
* Live end to end: the reporter's exact command on the vllm-ascend clone with this build succeeds on a 4 KiB host with unchanged sizing.

Honest limit: no 64 KiB kernel is available in this environment, so the scaled path is proven by unit stubs plus the engine source mechanism above rather than by execution on real 64 KiB hardware. The env override remains the field escape hatch either way, and one `getconf PAGE_SIZE` from the reporter confirms the diagnosis.

## Workaround for anyone hitting this today

`GITNEXUS_LBUG_BUFFER_POOL_SIZE=4294967296` (4 GiB; `0` restores LadybugDB's native default).

A follow up for the engine side is worth filing upstream: LadybugDB's build already exposes `LBUG_PAGE_SIZE_LOG2`, so publishing aarch64 binaries with page size matched frames would restore exact charge and refund accounting with no engine code changes.


magyargergo:
@gitnexus review

PR #1478, opened 2026-05-10
fix(cli): make --no-stats actually omit volatile counts (#1477)

RezaAlmiro:
## Summary

The `--no-stats` flag added in #1059 (closes feature request #704) was a no-op: the count phrase still appeared on every reindex. This PR makes the flag effective by fixing a commander.js negation-flag misread, plus adds two regression tests so it cannot silently regress again.

## Motivation / context

Filed as bug [#1477](https://github.com/abhigyanpatwari/GitNexus/issues/1477). Original feature request: [#704](https://github.com/abhigyanpatwari/GitNexus/issues/704). Closing PR for the feature: [#1059](https://github.com/abhigyanpatwari/GitNexus/pull/1059).

Discovered while debugging chore-commit churn in a downstream marking-OS project: every `npx gitnexus analyze` run produced a tiny diff in `CLAUDE.md` / `AGENTS.md` because the cluster boundary count drifts mildly between runs. Adding `--no-stats` as documented didn't help — turns out the flag was silently broken.

## Areas touched

- [x] `gitnexus/` (CLI / core / MCP server)
- [ ] `gitnexus-web/`
- [ ] `.github/`
- [ ] `eval/` or other tooling
- [ ] Docs / agent config only

## Root cause

`.option('--no-stats', ...)` in `src/cli/index.ts:38` follows commander.js's [negation-flag convention](https://github.com/tj/commander.js/blob/master/Readme.md#other-option-types-negatable-boolean-and-booleanvalue): the option is registered under the accessor `stats` (boolean, default `true`; `false` when the user passes `--no-stats`), **not** `noStats`. Two action-handler reads in `src/cli/analyze.ts` were reading `options?.noStats`, which is always `undefined`, so the `noStats` payload reached `runFullAnalysis` (at line 414 pre-fix) and `generateAIContextFiles` (at line 500 pre-fix) as `undefined`/falsy, and the count branch in the template always rendered.

The `AnalyzeOptions` interface in `analyze.ts` also declared `noStats?: boolean`, which was a static-type misrepresentation of what commander actually provides at runtime — a contributing factor that hid the bug from a TypeScript-aware reader.

The internal `AnalyzeOptions` in `core/run-analyze.ts` and `AIContextOptions` in `cli/ai-context.ts` correctly use `noStats?: boolean` because they're not commander-shaped — they receive the *re-packed* options object built at the bridge in `analyze.ts`. Those reads are unchanged.

## Scope & constraints

**In scope**

[...]

PR #1059, opened 2026-04-24
fix(cli): translate Commander --no-stats into noStats option (#704)

jonasvanderhaegen-xve:
## Problem

Commander's negatable-boolean convention sets `options.stats = false` when `--no-stats` is passed — **not** `options.noStats = true`. The three call sites that forward CLI options into `runFullAnalysis` / `generateAIContextFiles` all read `options.noStats`, so the flag was silently ignored and volatile symbol/file counts always got written into `AGENTS.md` / `CLAUDE.md`.

Confirmed broken on 1.6.1 and 1.6.2.

## Fix

Add `?? (options?.stats === false)` fallback at all three call sites:

- `gitnexus/src/cli/analyze.ts:231` — `runFullAnalysis` options
- `gitnexus/src/cli/analyze.ts:301` — `generateAIContextFiles` in skills path
- `gitnexus/src/core/run-analyze.ts:382` — `generateAIContextFiles` in core path

No new API surface, no behaviour change when `--no-stats` is not passed (`options.stats` is `undefined` in that case, so `undefined === false` is `false` and the fallback doesn't fire).

Closes #704

Issue #704, opened 2026-04-07
Option to disable the amount of indexed files in CLAUDE.md and AGENTS.md


ganqqwerty:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Problem or opportunity

Problem: each time I run analyze, the CLAUDE.md and AGENTS.md get updated with  message like  (2772 symbols, 7405 relationships, 232 execution flows). 
Problems with that:
1. This creates noise in the git history, my CLAUDE.md and AGENTS.md are updated too often. 
2. it's also not helpful as a statistic and does not help gitnexus to operate
3. If I have re-analyse bind to post-commit hook, the working tree immediately have new changes.
 
Is it possible to add an option to disable this behavior? 

### Proposed solution

1. One solution could be to use a key --disable stats for gitnexus. 
2. If stats are essential and help gitnexus a lot, maybe we could put them to a dedicated file (which I will gitignore)

### Contribution

- [ ] I am willing to open a PR for this (may need design discussion first).

Gahrcoder:
This was fixed in #781 (merged). Can be closed.

craigb-tt:
On 1.6.0 (latest on npm), `--no-stats` is present in `gitnexus analyze --help` with the description "Omit volatile file/symbol counts from AGENTS.md and CLAUDE.md", but the flag doesn't actually suppress the counts in my testing.

**Reproduction:**

```bash
$ gitnexus --version
1.6.0
$ gitnexus analyze --help | grep stats
  --no-stats        Omit volatile file/symbol counts from AGENTS.md and CLAUDE.md
```

**Test 1 — update an existing block with `--no-stats`:**
The counts update (6680 → 6693) but are not removed.

**Test 2 — strip the `<!-- gitnexus:start -->…<!-- gitnexus:end -->` block from CLAUDE.md, delete AGENTS.md, then regenerate fresh:**

```bash
$ gitnexus analyze --no-stats --force
… Repository indexed successfully
$ grep "indexed by GitNexus" CLAUDE.md
This project is indexed by GitNexus as **keel** (6660 symbols, 21658 relationships, 150 execution flows). …
```

The intro line in the regenerated block still contains the counts. Same result with both `--no-stats` alone and `--no-stats --force`.

So either the help text is ahead of the implementation in 1.6.0, or the flag's scope is different from what the help describes. Either way, the documented workaround for the commit-analyze-commit loop doesn't currently work.

olafurns7:
[...]

Issue #2563, opened 2026-07-18
scope-resolution: extend the instance-ownership free-call gate to C# and Kotlin


magyargergo:
## Context

Roadmap carried from #2550/#2555 (PRs #2549, #2558). `ScopeResolver.freeCallsRequireInstanceOwnership` closes the finalize-bucket free-call leak for languages where every method belongs to a class instance. Java opts in today; the flag was designed for wider adoption:

## C#

Requires C#-side instance modeling first: C#'s anonymous-ish shapes are lambdas and local functions (C# anonymous types cannot declare methods), which are structurally different from Java's anonymous class bodies — the ownership/dispatch story needs its own design before the gate can be enabled without suppressing true edges (the exact failure mode the Java rollout hit and fixed via the instance model).

## Kotlin

Kotlin's #2545 fix is name-keyed (`builtInNames`: `println`, `run`, `apply`, ...). Kotlin already models anonymous `object { ... }` expressions with `@scope.class` and has real implicit-`this` sibling dispatch, so the structural gate is likely closer to viable — it would additionally cover NON-builtin same-file name collisions the name list cannot. Needs the same verification discipline: probe which existing behaviors depend on the bucket path before gating (the measurement-probe-first pattern from #2550).

## Guardrails from the Java rollout (do not relearn)

- Never populate `LanguageProvider.builtInNames` with real method names — `parse-worker`'s call-extraction gate suppresses member calls too (verified failure).
- Gate same-file matches only — cross-file Method matches come through genuine import channels (verified failure: arity-narrowing parity suite).
- Probe first: instrument the free-call path across the language's full suite before designing (this killed two wrong designs for Java before any code shipped).

PR #2640, opened 2026-07-22
feat(analyze): add opt-in --self-commit flag for AGENTS.md/CLAUDE.md churn

### LINKED ISSUE #2639: feat(analyze): opt-in --self-commit flag to auto-commit CLAUDE.md/AGENTS.md stat-line churn
### Area

### Problem or opportunity

#704 covered users who want the volatile symbol/relationship/flow counts *removed* from `CLAUDE.md`/`AGENTS.md` entirely (`--no-stats`, fixed in #1478). But a second use case is not served by that flag: users who find the stats genuinely useful, but are frustrated that every `analyze` run leaves an **uncommitted diff** behind in their working tree.

This is explicitly called out in #704 itself — [@olafurns7](https://github.com/abhigyanpatwari/GitNexus/issues/704#issuecomment) noted "the constant re-editing of the CLAUDE/AGENTS files ... should be configurable," and the original reporter's problem #3 ("If I have re-analyse bound to a post-commit hook, the working tree immediately has new changes") is really a request for the churn to be *committed*, not necessarily removed.

Today the only ways to avoid the dangling diff are:
1. `--no-stats` (loses the stat line entirely), or
2. an external wrapper (shell alias, CI step, or an agent-specific hook) that shells out to `git add`/`git commit` after `analyze` — which only fires for whichever invocation path the user remembered to wrap, and doesn't help bare-terminal / cron / other-agent usage.

### Proposed solution

Add an opt-in `--self-commit` flag to `gitnexus analyze`. When passed, after `CLAUDE.md`/`AGENTS.md` are (re)written:

1. Check `git diff --quiet -- CLAUDE.md AGENTS.md` (scoped to whichever of the two files actually exist — same existence-scoping `--no-stats` already uses).
2. If there's a diff, `git add` **only** those files and `git commit -m "chore(gitnexus): refresh index stats [skip ci]"`.
3. No-op silently (never throw, never fail the overall `analyze` exit code) if: not inside a git repo, neither file exists, neither file changed, or the commit step itself fails (e.g. no git identity configured).
4. Never touches any other file — no `git add -A`.
5. Composes naturally with `--no-stats` — if both are passed, `--self-commit` is simply a no-op most of the time since there's nothing to diff.

This mirrors the existing `src/storage/git.ts` pattern (already wraps read-only `git` calls like `rev-parse`, `config --get`) with one additional write operation, gated the same opt-in way `--no-stats` is gated.

### Alternatives considered

- External hook/wrapper only (shell alias, CI step, or a Claude Code/Codex/Gemini hook via the existing `hooks/claude/gitnexus-hook.cjs` cross-agent installer): doesn't cover bare-terminal or CI invocation, and requires per-user/per-platform setup.
- A native git `post-commit` hook: wrong lifecycle event — `analyze` isn't triggered by a commit, so there's nothing for `post-commit` to hook into here.

A CLI-native flag is the only option that fires identically regardless of caller (terminal, cron, CI, or any agent) and platform.

### Contribution

- [x] I am willing to open a PR for this.

jecanore:
## Summary

Adds an opt-in `--self-commit` flag to `gitnexus analyze` that auto-commits any AGENTS.md/CLAUDE.md changes the run produces, scoped to only those two files.

## Motivation / context

Closes #2639.

#704 covered users who want the volatile symbol/relationship/flow counts *removed* from AGENTS.md/CLAUDE.md entirely (`--no-stats`, fixed in #1478). This PR serves the complementary case: users who find the stats useful but are frustrated that every `analyze` run leaves an uncommitted diff behind — explicitly called out in #704 itself (["the constant re-editing of the CLAUDE/AGENTS files ... should be configurable"](https://github.com/abhigyanpatwari/GitNexus/issues/704#issuecomment)), and the original reporter's problem #3 ("If I have re-analyse bound to a post-commit hook, the working tree immediately has new changes") is really a request for the churn to be *committed*, not necessarily removed.

## Areas touched

## Scope & constraints

**In scope**

- New `selfCommitContextFiles(repoPath, candidateFiles)` in `src/storage/git.ts`, following the existing read-only helpers' `execFileSync`/`windowsHide` convention in that file.
- Uses `git status --porcelain` (not `git diff --quiet`) deliberately — a first-time `analyze` run *creates* AGENTS.md/CLAUDE.md fresh, and untracked files never show up in `git diff`, only in `git status`. This mirrors why `isWorkingTreeDirty` in the same file already uses `--porcelain`. (Caught by an end-to-end smoke test against a real scratch repo during development — an earlier `git diff --quiet`-based draft silently skipped first-run creation.)
- New `--self-commit` CLI flag registered in `src/cli/index.ts`, next to `--no-stats`.
- Wired into `analyze.ts` at the two points where AGENTS.md/CLAUDE.md can change: the "already up to date" fast path (base_ref refresh) and the primary analysis completion path (including the `--skills` regeneration branch).
- Commit message: `chore(gitnexus): refresh index stats [skip ci]`.
- Never `git add -A` — always scoped to exactly the AGENTS.md/CLAUDE.md paths that exist.
- Composes with `--no-stats`: when both are passed, `--self-commit` is simply a no-op most of the time since there's nothing to diff.
- Best-effort throughout: no-ops silently (never throws) if not a git repo, neither file exists, neither changed, or the commit step fails (e.g. no git identity configured) — never fails the surrounding `analyze` run.

- No changes to `--no-stats` behavior itself (only reused as prior art for the opt-in gating style).
- No change to `runFullAnalysis`'s return shape — `analyze.ts` re-checks file existence itself rather than threading a "files written" list up from `run-analyze.ts`, keeping this change isolated to the CLI-facing files.
- Doesn't address the (separate, orthogonal) post-commit-hook reindex-loop concern also mentioned in #704 — that's an infra/loop-prevention question, not something a self-commit flag can or should solve.

## Implementation notes

Modeled directly on the `--no-stats` precedent (#1059/#1478): same opt-in-off-by-default gating, same "scoped to whichever of AGENTS.md/CLAUDE.md exist" semantics. Unlike `--no-stats`, there's no commander.js negation-flag bridging concern here since `--self-commit` is a plain (non-negated) boolean — `options.selfCommit` maps directly with no `options.stats === false`-style indirection needed.

## Testing & verification

- [x] `cd gitnexus && npm test` — full `test/unit` suite: 513/515 files passed, 10409/10483 tests passed (the 2 failing files are `evidence-provenance-helper.test.ts`, a pre-existing macOS-vs-Linux `/proc/self/fd` environment failure, confirmed present on a clean `upstream/main` checkout before this change — unrelated to this PR).
- [x] `cd gitnexus && npx tsc --noEmit` — clean.
- [x] `npx eslint` on all changed files — 0 errors (4 pre-existing warnings in `git-utils.test.ts` on lines this PR didn't touch).
- [x] New unit tests: `test/unit/analyze-self-commit-bridge.test.ts` (CLI→helper wiring, mirroring `analyze-no-stats-bridge.test.ts`'s mocking style) and 5 new cases appended to `test/unit/git-utils.test.ts` (real tmp-git-repo behavior: scoped commit of a modified file leaving an unrelated dirty file untouched, no-op when nothing changed, no-op when neither file exists, never throws outside a git repo, and — the regression this PR's own smoke-testing caught — commits newly-created/untracked files correctly).
- [x] Manual end-to-end smoke test against a real scratch repo: first `analyze --self-commit` run creates and commits AGENTS.md/CLAUDE.md with a clean resulting working tree; a subsequent run with no repo changes is a true no-op; a run after real repo changes produces exactly one new scoped commit, again leaving the working tree clean.

## Risk & rollout

Opt-in and off by default — zero behavior change for existing users who don't pass `--self-commit`. No index refresh or migration needed.

## Checklist

- [ ] If `AGENTS.md` / overlays changed: headers, scope block, and changelog updated per project conventions *(N/A — this PR changes the flag that governs those files, not their content)*


jecanore:
@abhigyanpatwari this is ready for review whenever you have a chance — full CI matrix is green (only the Vercel deploy-auth check fails, which is unrelated to this change and needs fork authorization it doesn't have). Thanks!

koriyoshi2041 (changes requested):
The scoped pathspec prevents unrelated *files* from being committed, but it does not isolate this analyze run's edits within AGENTS.md/CLAUDE.md. If either file already contains an unstaged user edit, `git add -- AGENTS.md CLAUDE.md` stages and commits that edit together with the generated stats refresh. That makes `--self-commit` capable of silently committing work it did not produce.

There is a second state-mutation case on the failure path: `git add` runs before `git commit`, and the catch suppresses a failed commit (for example, missing identity or a hook rejection) without restoring the index. The command then reports success but leaves both context files staged, potentially changing the user's next commit.

I reproduced the underlying behavior with a scratch repo: start from a tracked `AGENTS.md`, append a user note, append a simulated stats refresh, then run the helper's `git add`/`git commit` sequence; the resulting commit contains both appended lines. With commit identity removed, the same sequence fails the commit and `git diff --cached --name-only` still reports `AGENTS.md`.

Could this capture the pre-run state and commit only a verified generated delta, or otherwise refuse to self-commit when either candidate is already dirty before analyze? The failure path also needs to restore only what this helper staged (without disturbing pre-existing index state). Tests should cover a pre-dirty candidate and a commit failure after staging.

Exact-head verification at `d43429ca`: `npx vitest run test/unit/git-utils.test.ts test/unit/analyze-self-commit-bridge.test.ts` — 32/32 passed; the current tests do not exercise either case above.

gitnexus/src/storage/git.ts
```diff
@@ -46,6 +46,51 @@ export const isWorkingTreeDirty = (repoPath: string): boolean => {
...
+      stdio: 'ignore',
+      windowsHide: true,
+    });
+    execFileSync(
+      'git',
+      ['commit', '-m', 'chore(gitnexus): refresh index stats [skip ci]', '--', ...existing],
+      { cwd: repoPath, stdio: 'ignore', windowsHide: true },
+    );
+  } catch {
+    // best-effort — e.g. missing git identity, or a race that left nothing
```
magyargergo: We should log when this failed to inform users.

magyargergo (changes requested):

gitnexus/src/storage/git.ts
jecanore: Added: catches the commit failure via `logger.warn` (matching the codebase's existing pino-logger convention) with the underlying error and the file list, and added a test that forces a real commit failure (missing git identity, isolated from the CI runner's ambient config) to assert the warning is emitted. Pushed in 100b8e23.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

jecanore:
Heads up @abhigyanpatwari — `tests / eval / containment (windows)` failed on the latest push, but it's in `eval/tests/test_process_control.py::test_stdout_and_stderr_are_bounded_while_the_process_runs`, a Python subprocess-timing test wholly unrelated to the TS files this PR touches (`git.ts`/`analyze.ts`/`git-utils.test.ts`). The failure is `assert 'forced-kill' == 'exited'` — the harness killed the child before it exited within the timeout, which reads as a timing-margin flake on a loaded Windows runner rather than a real regression. I don't have rerun permission on this repo (not a collaborator) — could you retrigger that job when you get a chance? Happy to dig further if it turns out to be a real issue.

koriyoshi2041 (changes requested):
The new head does not address the two state-safety blockers from the prior review.

Compared with reviewed head `d43429ca`, the only relevant production change is a warning in the existing catch block. `selfCommitContextFiles()` still runs `git add -- AGENTS.md CLAUDE.md` before `git commit`, so:

1. a pre-existing unstaged user edit in either candidate is still included in the generated commit; and
2. a failed commit still leaves the candidate staged, changing the user's index despite `analyze` reporting success.

The new no-identity test confirms only that a warning is emitted and that no commit lands. It does not assert index restoration; on this path, `git diff --cached --name-only` still contains `AGENTS.md`.

Please either refuse self-commit when a candidate was dirty before analyze, or preserve/restore the exact pre-run worktree and index state while committing only the verified generated delta. Add regressions for both a pre-dirty candidate and the post-`git add` commit-failure index state.

jecanore:
@koriyoshi2041 thanks for the thorough repro — both are fixed in 9c240893:

1. **Pre-existing edit no longer swept in.** Added `snapshotSelfCommitSafety()`, called from `analyze.ts` right before `runFullAnalysis` (i.e. before either AGENTS.md/CLAUDE.md write path runs). It marks each candidate safe only if it doesn't exist yet (first-time creation) or is currently clean. `selfCommitContextFiles()` now takes that snapshot and only ever `git add`/`git commit`s files marked safe — a file with a pre-existing edit is skipped (logged) and left completely untouched, including out of the index.
2. **Index no longer left mutated on commit failure.** The catch around `git commit` now runs `git reset -- <safe files>` to undo the preceding `git add`, so a failed commit (missing identity, hook rejection, etc.) restores the pre-add index state instead of leaving files silently staged.

New regression tests in `git-utils.test.ts`: one reproduces your scratch-repo scenario (pre-dirty AGENTS.md skipped while a clean CLAUDE.md still commits normally, verified via `git diff-tree`), the other reproduces the post-`git add` commit-failure case and asserts `git diff --cached --name-only` is empty afterward. Full suite + typecheck + lint green locally.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

magyargergo:
@jecanore could you please add i18n to the command description? 🙏

koriyoshi2041 (approved):
Re-reviewed exact head d3d18c1f. The pre-run safety snapshot now excludes dirty context files, the failure path resets the helper-staged paths, and the added regressions cover both original blockers. Hosted unit/CI matrix is green; approving.

magyargergo:
@jecanore can you please resolve the merge conflicts? 🙏

jecanore:
@magyargergo both done:

1. **i18n** — added `help.option.analyze.selfCommit` to both `en.ts` and `zh-CN.ts`, wired into `help-i18n.ts`'s `OPTION_DESCRIPTION_KEYS` map (matching `--no-stats`/`--skills`), so the help text now localizes correctly instead of always rendering in English.
2. **Merge conflicts** — resolved (the branch fell behind on a new `isWorkingTreeDirty` test suite that landed in `main` in the same file I'd extended; a clean side-by-side addition, no real overlap). Also caught and fixed a stale `gitnexus-shared/dist` build that the merge exposed (3 typecheck errors on an unrelated `BindingRef.visibility` field from #2563).

Pushed as 7e291f6d. PR is mergeable again and full suite/typecheck/lint are green locally.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

magyargergo:
@gitnexus review

magyargergo:
@gitnexus review

jecanore:
`tests / ubuntu / coverage 3/3` failed on the latest push, but it's a wall-clock scaling benchmark unrelated to this PR: `test/unit/cjk-segmentation.test.ts > scales linearly on realistic interleaved CJK/non-CJK content, not quadratically` — `expected 21.99ms to be less than 21.45ms` (a 20x-of-baseline threshold, missed by ~2.5%). Nothing in this PR touches CJK segmentation; this is CI-runner timing noise on a razor-thin margin. Same non-collaborator rerun limitation as before — could you retrigger when you get a chance?

magyargergo (approved):

Issue #2637, opened 2026-07-22
analyze leaves orphan `lbug.wal.missing-shadow.*` quarantine files that are never cleaned up (accumulate unbounded)


joelmdev:
### Area

gitnexus (CLI / core / indexing)

### Summary

GitNexus never reclaims `lbug.wal.missing-shadow.*` quarantine files. Based on the
filename, LadybugDB appears to rename a WAL aside as
`lbug.wal.missing-shadow.<epochMs>-<rand>` when it opens the store and finds a WAL
whose shadow is gone, rather than deleting it. The orphan-sidecar cleanup added in
#1618 (`doInitLbug`) only unlinks a fixed list — `lbug.shadow` and
`lbug.wal.checkpoint` — so these timestamped quarantine fragments escape it and
accumulate in `.gitnexus/` for the life of the repo. Each can be several MB.

### Context

- Windows 11, gitnexus 1.6.9 (via `npx gitnexus@latest`), Node v24.3.0.
- Observed in real usage: a `.gitnexus/` directory held
  `lbug.wal.missing-shadow.1781189886983-ynmn0sijw68` at **1.88 MB**, with the main
  `lbug` file present and intact alongside it. The fragment was never referenced
  again — pure detached garbage sitting on disk until removed by hand.
- These show up after an analyze is interrupted or a second writer touches the store
  (the same crash/kill class of condition as #1618), which leaves a WAL without its
  shadow; the next open quarantines that WAL under the `missing-shadow` suffix.

### Expected

Once LadybugDB has quarantined a WAL as `missing-shadow` — i.e. it has already
classified that WAL as detached and unrecoverable — GitNexus should reclaim it on the
next init/analyze. A healthy index should not accumulate multi-MB orphan WAL fragments
indefinitely.

### Actual

`lbug.wal.missing-shadow.*` files persist forever, growing in count and size with each
quarantine event, and are only ever removed by a manual `rm`.

### Root cause

The #1618 fix in `doInitLbug` (`dist/core/lbug/lbug-adapter.js`) unlinks a fixed
sidecar list (`lbug.shadow`, `lbug.wal.checkpoint`) when the main DB is missing. The
`missing-shadow` quarantine files use a timestamped/random suffix and are not in that
list, so they are never cleaned. They should also be reclaimed even when the main DB
*is* present, since a quarantined WAL is by definition already detached from the live
store.

### Suggested fix

[...]

Issue #2544, opened 2026-07-18
GitNexus Analyzer Issue


FlameZerg:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

GitNexus Analyzer Issue

### Context

Every time the code is modified, GitNexus Analyzer throws the following error; this is a serious issue that causes index corruption, necessitating a full rebuild each time.

PS E:\0project\test> gitnexus analyze
  GitNexus Analyzer

  GITNEXUS_MAX_FILE_SIZE: effective threshold 2048KB (default 512KB)

Incremental: changed=7, added=0, deleted=0 (skipping wipe + 821 unchanged file rows preserved)
Incremental: +28 importer(s) added to writable set (BFS depth ≤ 4)
  ████████████████████████████████████░░░░  90% | Loading embedding model...{"level":40,"time":1784333291376,"name":"gitnexus","msg":"VECTOR extension unavailable; semantic embeddings fall back to exact scan. To enable vector search, install it once with network access (GITNEXUS_LBUG_EXTENSION_INSTALL=auto), or pre-install it for offline use. Set GITNEXUS_LBUG_EXTENSION_INSTALL=never to skip installs and silence this."}
  ████████████████████████████████████░░░░  90% | Embedding 0/?

  Analysis failed: Batch execution failed for rows 1-4: Runtime exception: Found duplicated primary key value :0, which violates the uniqueness constraint of the primary key column. (CREATE (e:CodeEmbedding {id: $id, nodeId: $nodeId, chunkIndex: $chunkIndex, startLine: $startLine, endLine: $endLine, em)
Error: Batch execution failed for rows 1-4: Runtime exception: Found duplicated primary key value :0, which violates the uniqueness constraint of the primary key column. (CREATE (e:CodeEmbedding {id: $id, nodeId: $nodeId, chunkIndex: $chunkIndex, startLine: $startLine, endLine: $endLine, em)
    at file:///E:/nodeLib/npm/global/node_modules/gitnexus/dist/core/lbug/lbug-adapter.js:1428:23
    at async withConnLock (file:///E:/nodeLib/npm/global/node_modules/gitnexus/dist/core/lbug/conn-lock.js:52:16)
    at async executeWithReusedStatement (file:///E:/nodeLib/npm/global/node_modules/gitnexus/dist/core/lbug/lbug-adapter.js:1414:9)
[...]

PR #2677, opened 2026-07-24
fix(analyze): single-writer lock for the index write path (#2658)

### LINKED ISSUE #2658: analyze has no single-writer lock: concurrent runs race the same index; add a lock and --skip-if-running
### Area

gitnexus (CLI / core / indexing)

### Summary

`gitnexus analyze` takes no exclusive lock on the index it writes, so two or more
`analyze` runs can execute concurrently against the same `.gitnexus/` store. They race
on the same `lbug` file and sidecars, and the duplicated work is pure waste — each run
recomputes an index the other is already producing.

Two asks:

1. A native single-writer lock around the analyze write phase.
2. A `--skip-if-running` flag so an invocation that finds a run already in flight exits
   **0 immediately** rather than colliding, blocking, or queueing.

### Context

This bites hardest with hook-driven re-indexing. I run several Claude Code sessions on
one machine, each firing a SessionStart hook that refreshes the GitNexus index for the
repo it opens. When two or more sessions start or resume at nearly the same moment
against the same repo, multiple `analyze` processes launch on the same index. Any
editor/agent integration that re-indexes on open or on branch change has the same
shape — this is not an exotic configuration.

Observed fallout: orphaned WAL fragments left in `.gitnexus/` (filed separately as
#2637), plus N× CPU spent producing one index.

I worked around it with an external wrapper that holds a Windows named mutex around
the analyze invocation: first-in-wins, and late invocations skip their re-index and
simply use the index the in-flight run is refreshing. That works, but it is OS-specific
and every downstream integrator has to reinvent it. The mutual exclusion belongs in
`analyze` itself, since only GitNexus knows the real boundaries of the write.

### Expected behavior

`analyze` holds an exclusive lock on the index for the duration of its write. A second
concurrent `analyze` on the same index either:

- with `--skip-if-running` — exits 0 immediately, reporting that another run holds the
  lock and that the on-disk index is being refreshed by it; or
- by default — fails fast with a message naming the holder, instead of racing.

#2599 already landed on the right error shape for the neighbouring case ("store is held
open by another gitnexus process (pid N)"). This asks for the same clarity for
analyze-vs-analyze, and for actual enforcement rather than only a better message.

### Actual behavior

Concurrent `analyze` runs proceed against the same store with no mutual exclusion and
no warning.

### Why skip rather than queue

For hook-driven re-index, queueing is the wrong default: by the time the second run
would start, the first has already produced the fresh index the caller wanted.
Serialising the runs multiplies cost for no benefit. Skipping is correct and is what
`--skip-if-running` should express. Callers that genuinely need to wait can poll or
block themselves.

### Suggested shape

Prefer an OS-level advisory lock held by the process — `LockFileEx` on Windows,
`flock`/`fcntl` on POSIX — over a plain pidfile. The OS releases such a lock when the
holder dies for any reason (crash, kill, OOM), which sidesteps the stale-lock problem:
a pidfile written to `.gitnexus/analyze.lock` needs liveness checks and can still be
raced or left behind by a hard kill. Suggested surface:

- lock scoped to the index directory, so separate repos/worktrees never contend;
- acquired before the first write and released on exit;
- `--skip-if-running` → exit 0 with a clear "another analyze holds the index" note;
- default (no flag) → distinct non-zero exit naming the holding pid.

Concurrent *reads* (`query`, `mcp`, etc.) should stay unrestricted; only the analyze
write path needs exclusion.

### Environment

- OS: Windows 11 Pro (build 26200)
- GitNexus: 1.6.9 (`npx gitnexus@latest`)
- Node: v24.3.0

### Related

- #2637 — orphan `lbug.wal.missing-shadow.*` quarantine files (one observed symptom of
  colliding writers).
- #2599 — analyze vs. a live `gitnexus mcp` holding the store open; same family
  (another process holds the store), reader-vs-writer rather than writer-vs-writer.
- #2456 — [Design] strict no-force analysis and read-only recovery plan.

magyargergo:
Closes #2658.

## Problem

Concurrent `gitnexus analyze` runs on the same `.gitnexus/` store take no exclusive lock. They wipe and rebuild the same `lbug` file at once, waste N× CPU producing one index, and leave orphaned WAL fragments (#2637). This bites hardest with hook-driven re-index: several editor/agent SessionStart hooks firing on one repo at nearly the same moment launch multiple `analyze` processes against the same index.

Reproduced on `main`: two cold `analyze --force` runs 3s apart both proceed with no mutual exclusion, both exit 0, one hits an FTS `rename … No such file or directory` race yet still reports "success", and the store is left with orphaned staging files.

## Fix

An exclusive, index-directory-scoped write lock around the analyze write path. A second concurrent analyze **waits** for the holder, then falls through to the normal freshness check — so a run whose work the holder already did returns "Already up to date" in seconds (single-flight coalescing), while a genuinely-changed tree does one follow-up incremental. **No new flag**: waiting is the default, which is what hook-driven re-index wants. (The issue also floated `--skip-if-running`; waiting-by-default subsumes the common case and can be added later without changing this.)

- **`src/storage/index-lock.ts`** — the lock, behind an `acquireIndexLock` seam so a kernel advisory backend (`flock`/`LockFileEx`) can replace the `O_EXCL`-pidfile backend later without touching orchestration. The record carries pid + hostname + Linux start-time (pid-reuse guard) + a random token (release/steal is token-checked). A dead holder's lock is reclaimed at once; a live one is waited on, never stolen; a foreign-host lock is never force-stolen. Ownership lives with the process that runs the pipeline (the heap-respawn child, not a supervising parent), so the liveness pid is always the real writer — a SIGKILLed parent with an orphaned still-writing child can't look dead and admit a second writer. Waiting is bounded by a finite default (10 min, `GITNEXUS_INDEX_LOCK_TIMEOUT_MS`, `≤0` = unbounded): on platforms without start-time verification a reused pid could otherwise masquerade as a live holder forever. Timing out is safe — it stops waiting, never steals a possibly-live holder — and names the holder.
- **`run-analyze.ts`** — `runFullAnalysis` is now a thin wrapper that resolves the write target, acquires the lock, then runs the pipeline. Branch placement decides the *actual* write directory (a `--branch` that owns the flat slot writes to `.gitnexus/`, not `branches/<slug>/`), so the lock scopes off the resolved `metaDir`, not the raw option. Both writer entry points (CLI and server worker) route through the wrapper, so both are covered. The existing `alreadyUpToDate` check becomes the post-wait revalidation for free.
- **Per-run staging path** (`lbug.staging.<uuid>`, was the fixed `lbug.new`) plus a lock-held orphan sweep, so a crashed run's half-built staging file can never be mistaken for — or clobber — a live run's.
- **FTS failure split** — capability (extension unavailable / single-row tokenizer error → degrade, keep exit 0, preserving the #2544/#2546 resilience) vs integrity (IO/rename/checkpoint/corruption → fail the run before the swap, leaving the previous index live). Under a single writer, an integrity failure is a real broken build, not another process racing, so it must no longer silently publish a search-less index as "success".

## Testing

- Unit coverage of acquire / steal / token-release / sweep / pid-reuse / timeout-ceiling / FTS classification.
- A real cross-process test proving mutual exclusion while held and reclaim after the holder is `SIGKILL`ed.
- End-to-end on a 755-file repo: two concurrent `--force` runs now serialize cleanly (no FTS rename race, no orphaned staging); on the realistic hook path (non-`--force`, stale index) the waiter prints "Already up to date" instead of rebuilding — single-flight coalescing confirmed.
- The staging rename desynced two existing `lbug-db` tests, both updated here: the WAL-checkpoint-failure test now drives an in-place incremental run (predictable `lbug.wal.checkpoint` target) instead of the now-random staging name, and the atomic-swap leftover-temp guard matches the new `.staging.` prefix.

## Follow-ups (out of scope)

Generation directories + a `CURRENT` pointer (Lucene-style commit points) are the scalable publication end-state; this lock is deliberately independent of that redesign.


magyargergo:
What do you think about this approach @joelmdev? This is following the industry standards to this kind of problems.

gitnexus/src/storage/index-lock.ts
```diff
@@ -0,0 +1,631 @@
...
+/**
+ * Stable OS-IPC endpoint name for an index directory. The name is derived from
+ * the resolved absolute path (case-folded on Windows), so two processes
+ * targeting the same slot collide and separate worktrees/branches never do. The
+ * endpoint lives OUTSIDE the index directory (abstract namespace / pipe
+ * namespace), so the lock needs no filesystem write and is unaffected by a
+ * read-only index mount.
+ */
+const socketLockName = (lockDir: string): string => {
+  const resolved = path.resolve(lockDir);
```
magyargergo: **H1 · High · [reproduced], cross-engine (Codex adversarial + Claude adversarial).** Mutual exclusion is keyed on the *lexical* path (`path.resolve(lockDir)`), not the real path. Two `analyze` processes reaching the same physical `.gitnexus` via a same-netns alias — a local symlink, a Windows junction, or a `\\?\` prefix — hash to different pipe/socket names and **both `listen()` succeed**, so both proceed as sole writer (the exact collision this backend prevents). The file backend is immune (`O_EXCL` → inode). Fix: `realpathSync` the deepest existing ancestor before hashing (fall back to `path.resolve` only when nothing on the path exists yet). Reproduced via `repro-symlink.mjs`: both aliases acquired the socket lock. (Cross-netns containers sharing a bind-mount are a *separate*, already-documented limitation this fix does not address.)

gitnexus/src/core/run-analyze.ts
```diff
@@ -625,34 +628,143 @@ export const pdgModeMismatch = (recorded: RepoMeta['pdg'], options: PdgOptions):
...
+  // cached value via getSearchFTSStemmer.)
+  initialiseSearchFTSStemmer();
+  initialiseSearchFTSCjkSegmentation();
+  // Scope the degraded-parse log throttle to this run (module-level counter
+  // would otherwise stay saturated on a reused process).
+  resetDegradedParseCounter();
+
+  const writeTarget = await resolveWriteTarget(repoPath, options);
+  const log = (msg: string) => callbacks.onLog?.(msg);
+  const lock = await acquireIndexLock(writeTarget.metaDir, {
```
magyargergo: **H2 · P1 · cross-engine STRONG (Codex swarm P1 + Codex adversarial + Claude reliability P1 + Claude correctness P2), coordinator-verified in source.** `writeTarget` — including `currentCommit`, captured at line 663 inside `resolveWriteTarget` — is resolved *before* this `acquireIndexLock`, which can block up to 10 min. When the waiter wakes, HEAD may have advanced to a **clean** commit Y, but the stale snapshot makes `existingMeta.lastCommit === currentCommit` at :1145 return `alreadyUpToDate`, and the new commit is **never indexed** (the `--branch`/placement slot and the meta stamp go stale too). This directly contradicts the module header's promise to re-check freshness *after* acquiring the lock. Self-heals on the next analyze, but serves a stale index silently until then. Fix: re-resolve the write target (commit/branch/placement) after the lock is held, before the up-to-date short-circuit.

gitnexus/src/storage/index-lock.ts
```diff
@@ -0,0 +1,631 @@
...
+    // this environment can't use the socket backend → fall back to the file one.
+    if (listenErr.code !== 'EADDRINUSE') throw new SocketLockUnavailable(listenErr);
+
+    if (!announcedWait) {
+      announcedWait = true;
+      opts.onWaitStart?.(me);
+      opts.log?.('Another gitnexus analyze is refreshing this index — waiting for it to finish.');
+    }
+    const waited = Date.now() - startedAt;
+    if (waited >= timeoutMs) throw new IndexLockTimeoutError(unknownHolder(), waited);
```
magyargergo: **M3 · P2.** On the socket/pipe backend the holder pid is never knowable, so this timeout throws with `unknownHolder()` (`pid: -1`), which the CLI (`analyze.ts:1604`) renders as "held by pid -1 on <host>" and which pollutes the structured `holderPid`. On the default Linux/Windows backend that's the *only* path a timeout takes. Fix here at the socket call site (not the shared `unknownHolder()` helper — the file backend also uses it at :450/:463): emit `null`/omit the pid for the socket backend and have the CLI print "another process (pid unknown)".

gitnexus/src/storage/index-lock.ts
```diff
@@ -0,0 +1,631 @@
...
+  token: randomBytes(16).toString('hex'),
+  invocationId: randomUUID(),
+  acquiredAt: new Date().toISOString(),
+});
+
+const readRecord = (lockPath: string): LockRecord | null => {
+  try {
+    const raw = readFileSync(lockPath, 'utf8');
+    const parsed = JSON.parse(raw) as Partial<LockRecord>;
+    if (typeof parsed.pid !== 'number' || typeof parsed.token !== 'string') return null;
```
magyargergo: **L4 · P3 · [reproduced].** `pid` is validated only as `typeof === 'number'`, and `typeof NaN === 'number'`, so `NaN`/`0`/`-1`/`Infinity` from a garbled *or* crafted lock file reach `process.kill(pid, 0)`; e.g. `{"pid":0}` reads as a live holder and wedges a real `analyze` for the full 10-min timeout. One-line hardening: `Number.isInteger(parsed.pid) && parsed.pid > 0`.

gitnexus/src/core/search/fts-indexes.ts
```diff
@@ -193,9 +193,76 @@ export async function verifySearchFTSIndexes(
...
+// to mention an integrity word still degrades (it isn't a broken build).
+const FTS_CAPABILITY_SIGNATURES = ['invalid utf-8', 'failed calling lower', 'tokeniz'] as const;
+// IO / durability / corruption signatures that mean the build itself broke.
+const FTS_INTEGRITY_SIGNATURES = [
+  'io exception',
+  'i/o error',
+  'io error',
+  'error renaming',
+  'checkpoint',
+  'no such file or directory',
```
magyargergo: **L1 · P3 (Codex adversarial).** Generic `'no such file or directory'` (and `'ebadf'` at :225) in the integrity-signature list means a capability/config ENOENT — e.g. a missing FTS extension asset — is now classified `integrity` and **aborts** the whole analyze on the atomic-swap path, where pre-PR it degraded to a queryable FTS-less index (exit 0). Consider requiring a write/rename/checkpoint verb context before escalating a bare ENOENT so a capability failure keeps degrading.

gitnexus/src/storage/index-lock.ts
```diff
@@ -0,0 +1,631 @@
...
+): Promise<IndexLockHandle> => {
+  const name = socketLockName(lockDir);
+  const pollMs = opts.pollMs ?? DEFAULT_POLL_MS;
+  const timeoutMs = resolveTimeoutMs(opts.timeoutMs);
+  const startedAt = Date.now();
+  let announcedWait = false;
+  let lastDiagnosticAt = 0;
+
+  for (;;) {
+    const server = net.createServer();
```
magyargergo, line 616: **L3 · P3.** A fresh `net.createServer()` is allocated every poll iteration but never `close()`d on the `EADDRINUSE`/busy branch. It's `unref()`'d and GC-eligible so the impact is minor, but a long contended wait churns server objects — `close()` it before the next poll (or reuse one server across iterations).

magyargergo (commented):
## PR #2677 — Cross-process single-writer index lock (#2658)

**Verdict: Changes requested — not production-ready as-is.** The lock's core architecture is sound and the review sweep validated the hard parts (no lock leak, no hot-loop, fatal-FTS-before-swap ordering, staging-sweep regex safety, heap-respawn ownership). But two correctness gaps at the center of the feature keep it from landing: **H1** — the socket/pipe backend keys mutual exclusion on the *lexical* path, so two writers reaching the same physical index dir through a symlink / bind-mount / Windows junction get different pipe names and both write; and **H2** — the git snapshot (`currentCommit` / write placement) is captured *before* the up-to-10-minute lock wait and reused afterward, so a waiter can wake behind an advanced-but-clean HEAD and skip indexing it (`alreadyUpToDate`), contradicting the module's own "re-check freshness after acquiring" promise. Compounding both, **H3**: the new multi-OS lock tests are not wired into the Windows/macOS CI subset, so the named-pipe and file-fallback paths where H1 actually bites get **zero** native-OS coverage — green Windows CI here does not exercise the lock. Fix H1+H2, wire H3, and this is mergeable.

**Engine breakdown.** Engine A (Claude): GitNexus swarm — risk-architect, test-ci [stubbed], security-boundary, branch-hygiene — plus ce personas (correctness, reliability, adversarial, testing, maintainability, project-standards). Engine B (Codex / gpt-5.5 xhigh): GitNexus swarm (B1) + adversarial (B3). **Engine B's ce leg (B2) hit a Codex spend-cap and did not run**, so cross-engine corroboration exists on the swarm/adversarial findings (H1, H2, H3, M1, L1) but *not* on the ce-only lanes (reliability, testing, maintainability specifics).

---

### Headline findings

**H1 — Socket lock keyed on lexical path → double-writer via path alias.** `High` · cross-engine STRONG (Codex B3 adversarial `high` + Claude ce-adversarial P2 conf75) · **[reproduced]**
- Trigger: two `analyze` processes reach the *same physical* `.gitnexus` dir via different lexical paths **in the same network namespace** — a local symlink, a Windows junction, or a `\\?\` prefix. `socketLockName` hashes `path.resolve(lockDir)` (lexical) rather than the real path, so each gets a different abstract-socket / named-pipe name and both `listen()` succeeds → concurrent writers, the exact corruption this PR exists to prevent. (Cross-netns containers sharing a bind-mount are a *separate*, already-documented limitation — `realpathSync` does not fix that; see the `selectBackend` comment / `GITNEXUS_INDEX_LOCK_BACKEND=file`.) The file backend (macOS/BSD fallback) is immune because `O_EXCL` resolves to the inode.
- Evidence: `gitnexus/src/storage/index-lock.ts:487-495` (`socketLockName` → `path.resolve` at :488). Reproduced by ce-adversarial: both aliases acquired the socket lock.
- Fix: `realpathSync` the deepest existing ancestor of `lockDir` before hashing (fall back to `path.resolve` only when nothing on the path exists yet), so aliases collapse to one name on both socket and file backends.

**H2 — Stale pre-lock git snapshot reused post-wait → false `alreadyUpToDate`.** `P1` · cross-engine STRONG, coordinator-verified (Codex B1 swarm P1 + Codex B3 adversarial `high` + Claude ce-reliability P1 + Claude ce-correctness P2 = 4 lanes / 2 engines) · **[code-read]**
- Trigger: waiter B resolves its write target and captures HEAD = X, then blocks up to 10 min behind writer A. During the wait HEAD advances to a *clean* commit Y. B wakes and compares the stale `existingMeta.lastCommit === currentCommit` (X === X → true) with a working tree that is clean at Y, returns `alreadyUpToDate`, and **skips indexing Y**. The same staleness affects the `--branch`/placement slot and the meta stamp. Self-heals on the next analyze, but silently serves a stale index until then — and directly contradicts the module header's "re-run freshness AFTER acquiring the lock" promise.
- Evidence: `currentCommit` captured in `resolveWriteTarget` at `gitnexus/src/core/run-analyze.ts:663`, called at :726, lock acquired at :728 (blocks); consumed at :1145 (`existingMeta.lastCommit === currentCommit`); HEAD is never re-read after the lock is held.
- Fix: re-resolve the write target (commit + branch + placement) *after* `acquireIndexLock` returns, before the up-to-date short-circuit.

**H3 — Multi-OS lock tests not wired into Windows/macOS CI.** `P1` · Codex B1 swarm P1, coordinator-verified · **[code-read]**
- Trigger: `run-cross-platform.ts` imports `ALL_CROSS_PLATFORM` from `cross-platform-tests.ts`; the Windows/macOS platform-sensitive shards run *only* that curated subset. `index-lock.test.ts` and `analyze-index-lock-concurrency.test.ts` are **not** in the list (`analyze-wal-checkpoint-failure.test.ts` *is*, proving the subset is deliberately selective). So the Windows named-pipe backend and the macOS file-fallback backend — the exact platforms where H1's alias split and the backend divergence live — get zero native-OS CI. Only the Linux ubuntu full-suite exercises the socket backend at all.
- Evidence: `gitnexus/scripts/cross-platform-tests.ts` (subset membership — imported as `ALL_CROSS_PLATFORM` by `scripts/run-cross-platform.ts`); the two new lock suites are absent.
- Fix: add `index-lock.test.ts` + `analyze-index-lock-concurrency.test.ts` to `cross-platform-tests.ts`. (Fix touches an unchanged file — reference the real location; not inline-able.)

---

### Additional findings (terse)

- **M1 — FTS integrity-abort branch untested.** `P1`/`P2` cross-engine (Claude ce-testing P1 + Codex B1 P2). The fatal path at `run-analyze.ts:2003-2017` (`integrity && atomic-swap` → throw) is never exercised: the one test mocks `buildSearchIndexesOrDegrade` to never return a `failureClass` and stubs `ftsFailureIsFatal` itself, so the "previous index left intact on abort" guarantee is unverified. Add a test that drives a real integrity classification through the throw. **[code-read]**
- **M2 — Server/MCP worker has no typed `IndexLockTimeoutError` handling.** `P2` risk-architect, coordinator-verified. `analyze-worker-core.ts:74-77` is a generic `catch` → `{type:'error', message}`; no `instanceof IndexLockTimeoutError`, no retry-after. The motivating "two SessionStart hooks contend" case routes here and gets the *worst* diagnostic (plus pid -1, see M3). Fix touches unchanged `analyze-worker-core.ts` — reference the real location; not inline-able. **[code-read]**
- **M3 — Socket-backend timeout reports "pid -1".** `P2` risk-architect + 3-lane cosmetic corroboration. The socket-backend timeout throws with `unknownHolder()` (`pid: -1`) at `index-lock.ts:565`; on the default Linux/Windows backend the holder pid is never knowable, so `analyze.ts:1604` prints "held by pid -1 on <host>" and `-1` pollutes structured `holderPid`. (The shared `unknownHolder()` helper at :250-252 is also used by the file backend's malformed/vanished-lock timeouts at :450/:463 — the fix belongs at the socket call site :565, not in the helper.) Print "another process (pid unknown)" for the socket backend and emit `null`/omit the pid field. **[code-read]**
- **L1 — FTS integrity signatures over-match capability ENOENT.** `P3` Codex B3. `FTS_INTEGRITY_SIGNATURES` includes generic `'no such file or directory'` / `'ebadf'` (`fts-indexes.ts:214-225`); a capability/config ENOENT on the atomic-swap path now *aborts* where pre-PR it degraded. Consider narrowing or excluding the plain-ENOENT case. **[code-read]**
- **L2 — Wrong recovery hint on FTS build failure.** `P3` Codex B1. CLI final warning `analyze.ts:1556` says the FTS *extension is unavailable* even when the failure was a build/verify integrity error (`run-analyze.ts:2019`), pointing the user at the wrong fix. **[code-read]**
- **L3 — `net.createServer()` per poll iteration, no `close()` on EADDRINUSE.** `P3` Claude ce-adversarial residual. `acquireViaSocket` creates a server object every poll and doesn't close it on the busy branch (`index-lock.ts:535`); unref'd and GC-eligible, so minor, but a long wait churns server objects. **[code-read]**
- **L4 — `readRecord` accepts NaN/0/-1/Infinity pids.** `P3` Claude security, reproduced. `readRecord` validates only `typeof parsed.pid === 'number'` (`index-lock.ts:188`), so a garbled/crafted lock file reaches `process.kill` with a nonsense pid and can wedge a real analyze for the full 10-min timeout. One-line fix: `Number.isInteger(pid) && pid > 0`. **[reproduced]**
- **L5 — File backend malformed-grace steal races a stalled live writer.** `P3` Claude ce-adversarial. A live writer stalled >1s (GC) between `O_EXCL` create and the `writeSync` loses its still-empty lock file to the malformed-grace stealer → both proceed. File backend only (macOS/BSD); low probability. **[code-read]**
- **Quality (maintainability, Claude A2).** `P2` — `acquireViaFile` / `acquireViaSocket` duplicate ~25 lines of wait/timeout/diagnostic scaffolding; `resolveWriteTarget` added inline to an already-2695-line `run-analyze.ts`. `P3` — `readRecord` casts to `LockRecord` after validating only 2 of 6 fields.

---

### Validated / credit (reviews confirmed these are correct)

The sweep refuted a large slate of plausible-sounding concerns — validation is a feature, and this list is why the two headline items above are trustworthy signal rather than noise:
- **Staging-sweep regex is safe.** `^lbug\.(staging\..+|new(\..+)?)$` never matches `lbug` / `lbug.wal` / `lbug.shadow`, is non-recursive, and is scoped to `metaDir`. (4 lanes: adversarial + security + risk + correctness; unit-tested.)
- **No lock leak.** The wrapper acquires immediately before `try/finally{release}`; every return/throw — including the fatal FTS throw and the inner rethrow — routes through `finally`; `release` is idempotent on both backends. (adversarial verified respawn ownership.)
- **Fatal-FTS ordering is correct.** The fatal throw is *before* the staging→live swap, so the live index stays intact on the atomic-swap path. (risk + correctness + reliability + adversarial.)
- **No poll hot-loop / infinite spin.** Jittered delay floors at 1 ms and every branch gates on the timeout.
- **Heap-respawn is safe.** `ensureHeap` re-execs and returns before acquire, so only the real writer child ever locks — no parent/child deadlock.
- **Suspicion refuted:** no other `fts-indexes` mock site breaks (`repo-manager-reconcile.test.ts:259` doesn't hit the branch); all CI shards green.
- **Branch hygiene / standards clean:** merge-as-is, no conflicts; no `any`/`as any`, no CHANGELOG edit, `node:` prefixes present; no security P0/P1, no bidi/secrets.

---

### CI status

All required checks green on head `6fcde1c9`. Windows shard 1/3 needed one rerun (it rides the 20-min vitest timeout ceiling; skills-e2e ~7.5 min — infra flake, not this PR's code). The Vercel check failure is deploy-auth only, unrelated to the change. Note that green Windows CI does **not** exercise the lock (see H3).

---

*This is an automated multi-tool digest (GitNexus swarm + ce personas + Codex/gpt-5.5; Codex ce leg unavailable via spend-cap). Verify each finding against the source before acting.*

magyargergo:
@gitnexus review

magyargergo:
@gitnexus review

gitnexus/src/storage/index-lock.ts
magyargergo: Fixed in `9cc8b2ef`: `acquireViaSocket` now `close()`s the never-bound `net.Server` on the EADDRINUSE (and any listen-error) branch before the next poll, so a long contended wait no longer churns one server object per iteration.

PR #2678, opened 2026-07-24
feat(spring): model profiles, conditions, and auto-configuration

### LINKED ISSUE #2415: feat(spring): model profiles conditionals and auto configuration
Parent tracking issue: #1225

Status: remaining work

Goal
Represent conditional Spring wiring so analysis can explain when beans and auto configuration are active.

Scope
Capture Profile, Conditional style annotations, common conditional variants, AutoConfiguration imports, spring.factories entries, and starter provided configuration metadata where available.

Acceptance criteria
1. Beans and configuration classes can carry profile or condition metadata.
2. Auto configuration classes discovered from Spring metadata are represented separately from application owned code.
3. Conditional relationships are not treated as always active unless the condition is known to be active.
4. Tests cover profile gated beans, property gated beans, and imported auto configuration metadata.

ShiningXu:
## Summary

- add a language-neutral Spring conditional metadata layer with Java and Kotlin capture adapters
- model `@Profile`, `@Conditional`, common `ConditionalOn*` variants, and property-key relationships without assuming runtime activation
- discover auto-configuration candidates from `AutoConfiguration.imports` and legacy `spring.factories`
- keep `CONDITIONAL_ON` as activation evidence and use generic `DECLARES` edges with Spring-specific reasons for metadata declarations
- resolve metadata entries by normalized FQN only; deduplicate source-unavailable synthetic classes by FQN and fail closed on ambiguous real FQNs
- keep `@AutoConfiguration` as class framework metadata only, without claiming a File-to-Class registration edge
- scope incremental cleanup to Spring-owned declaration reasons and stable synthetic-node IDs
- avoid path-sized allocations while classifying repository files and use a single-node/ambiguity FQN index instead of per-class arrays
- sanitize attacker-controlled condition text, preserve annotation source lines, share graph indexes across language passes, and cover Java/Kotlin performance
- keep wall-clock measurement out of Vitest; expose it through a standalone diagnostic benchmark instead

## Why

Spring wiring is frequently profile- or condition-gated and auto-configuration may be declared only through dependency metadata. The graph now records what static analysis can prove without claiming that a candidate is registered or active at runtime.

## Benchmark

Same-machine before/after measurements from the optimization pass:

- 200,000 unrelated paths classified: 25.1ms -> 5.4ms (about 78% faster)
- 40,000 Class nodes with 2,000 metadata candidates resolved and linked: 11.5ms -> 8.2ms (about 29% faster)

The standalone harness (`node --import tsx bench/spring-conditionals/measure.mjs`) was rerun after the review change. Current medians were:

- dense condition capture, 100 -> 400 classes: Java 38.5ms -> 151.4ms; Kotlin 99.3ms -> 481.5ms
- modern + legacy metadata parsing, 2,000 -> 8,000 declarations: 0.74ms -> 3.15ms
- unrelated path classification, 50,000 -> 200,000 paths: 0.67ms -> 2.77ms
- FQN resolution, 10,000 -> 40,000 Class nodes with 2,000 candidates: 4.76ms -> 11.17ms

The harness asserts only deterministic setup/output counts. Its wall-clock values are diagnostic and are not test pass/fail criteria.

## Validation

- `npx tsc --noEmit`
- relevant Spring auto-configuration, conditional pipeline, analysis-feature, and incremental tests: 30 passed
- `node --import tsx bench/spring-conditionals/measure.mjs`: completed successfully
- `npx eslint bench/spring-conditionals/measure.mjs`: 0 errors
- full `npm test`: 15,096 passed, 98 skipped; 42 unrelated failures remain in existing cross-language parsing, macOS path canonicalization, Linux-only evidence anchoring, optional grammar/build artifacts, Java resolver, and environment-sensitive suites

Closes #2415


magyargergo:
@gitnexus review

magyargergo:
This needs a little bit more thought, because i don't want to blow up the graph with different kinds of edge types.

---

I'll need to run a research regarding this how it will fit into GitNexus ecosystem.

magyargergo:
I have looked into this in more detail.

I do think the Spring conditional information is valuable, but I am not convinced that both new relationship types fit cleanly into the GitNexus graph model as they are currently defined.

`CONDITIONAL_ON` makes sense as a first class relationship because it represents a reusable semantic concept. A class, bean, or method can depend on a condition, while the actual runtime activation can remain unknown.

I am less comfortable with `AUTO_REGISTERS`. The name suggests that the target is definitely registered, but the available evidence only shows that the class is declared as an auto configuration candidate. Conditions, exclusions, classpath state, and user defined beans can still prevent it from becoming active.

I would prefer to represent this as a more generic declaration or discovery relationship, for example:

```
metadata file -> DECLARES -> auto configuration class
```

This could also be reused later for other metadata driven systems rather than adding a new relationship type for each framework feature.

I also think we should address a few correctness issues before merging:

The fully qualified class name from Spring metadata should never fall back to a simple name match, because it may attach the relationship to an unrelated class from a different package.

Synthetic classes should be identified by their fully qualified name rather than the metadata file path, otherwise the same external class may be created multiple times when declared in more than one file.

Duplicate fully qualified class matches should have an explicit ambiguity strategy rather than creating a third synthetic class.

The PR should also be rebased onto the latest graph relationship streaming changes and the new relationships should be audited against the retained relationship rules.

My preferred direction would be to keep `CONDITIONAL_ON`, replace `AUTO_REGISTERS` with a more generic declaration model, and keep the Spring annotation details inside the relationship metadata or reason instead of continuously expanding the core relationship vocabulary.

magyargergo:
I would suggest using `DECLARES` instead of `AUTO_REGISTERS`.

For example:

```text
AutoConfiguration.imports file -> DECLARES -> configuration class
```

This is closer to what we can actually prove from static analysis. The metadata file declares the class as an auto configuration candidate, but it does not guarantee that Spring will register or activate it at runtime.

The class may still be excluded, blocked by conditions, replaced by a user defined bean, or unavailable because of the runtime classpath.

We can keep the Spring specific meaning in the relationship reason, for example:

```text
type: DECLARES
reason: spring-auto-configuration-import
```

The `@AutoConfiguration` annotation can remain metadata on the class, while `CONDITIONAL_ON` represents the activation requirements.

This keeps the graph model accurate and avoids introducing a relationship name that makes a stronger claim than the source code can support.

ShiningXu:
Thanks — I think the intended model is clear now. Before changing the implementation, I want to confirm the one ambiguity that materially changes the graph result.

My understanding / proposed plan is:

- keep `CONDITIONAL_ON` as the first-class activation-dependency relationship;
- replace metadata-file `AUTO_REGISTERS` edges with generic `DECLARES` edges, using Spring-specific reasons such as `spring-auto-configuration-import` / `spring-auto-configuration-factory`;
- keep `@AutoConfiguration` only as class framework metadata (no File -> Class declaration edge synthesized from the annotation itself);
- resolve metadata entries by normalized FQN only — never by simple name;
- when there are zero exact source matches, create one synthetic Class whose identity is based only on the normalized FQN, so declarations from multiple metadata files converge on the same node;
- after rebasing, leave `DECLARES` and `CONDITIONAL_ON` out of `RETAINED_REL_TYPES`, because no later pipeline phase reads either type; persisted streamed edges still retain their actual reason;
- scope incremental delete/re-extract to Spring declaration reasons rather than treating every future generic `DECLARES` edge as Spring-owned.

For **multiple real Class nodes with the same normalized FQN**, which behavior do you prefer?

1. **Fail closed (my preference):** emit no `DECLARES` edge to a real class and do not create a third synthetic node; optionally report/count the ambiguity.
2. **Fan out:** emit `DECLARES` to every exact candidate and mark the reason as ambiguous.

The first option matches GitNexus's existing conservative ambiguity handling, but I don't want to assume that policy for metadata declarations without confirming it with you.

magyargergo:
Yes, I would prefer that we follow Spring Boot behaviour as closely as possible.

Spring Boot treats each metadata entry as a fully qualified class name. It collects candidates from the available metadata files and removes duplicate class names.

At runtime, Spring does not load every class that happens to have the same FQN. The Java classloader resolves one class based on the actual classpath and classloader order.

For GitNexus, I suggest the following behaviour:

* Resolve entries using the normalized FQN only. Never use the simple class name.
* Deduplicate candidates by normalized FQN.
* When there is exactly one matching real class, create the DECLARES edge to that class.
* When there is no matching real class, create one synthetic class using a stable identity based only on the normalized FQN. This allows declarations from different metadata files to point to the same node.
* When there are multiple real classes with the same FQN, use classpath or module precedence if GitNexus can determine it reliably.
* When that precedence is not known, fail closed. Do not create a DECLARES edge, do not fan out to every matching class, and do not create another synthetic class.
* Record the ambiguity through a diagnostic, counter, or debug message so the missing edge can be explained.

Fan out would not match Spring Boot behaviour because Spring would not activate every class with the same FQN. It would load the one selected by the runtime classpath.

The rest of the proposed model also looks correct:

* Keep CONDITIONAL_ON as the activation dependency relationship.
* Replace AUTO_REGISTERS with generic DECLARES edges.
* Use Spring specific reasons such as spring-auto-configuration-import and spring-auto-configuration-factory.
* Keep @AutoConfiguration as class framework metadata only.
* Do not create a file to class edge from the annotation itself.
* Keep DECLARES and CONDITIONAL_ON outside RETAINED_REL_TYPES while no later pipeline stage reads them.
* Scope incremental cleanup to the Spring declaration reasons rather than treating every DECLARES edge as Spring owned.
* Support AutoConfiguration.imports for current Spring Boot versions and spring.factories where older Spring Boot versions are supported.

This gives us the closest practical model of Spring Boot while avoiding relationships that cannot be justified from static analysis.

magyargergo:
Could you please benchmark you solution and see if you can improve your implementation efficiency?

gitnexus/test/integration/spring-conditionals-benchmark.test.ts
```diff
@@ -184,6 +194,143 @@ describe.skipIf(!BENCH_ENABLED)('Spring condition capture scaling (#2415)', () =
...
+        `  auto-config resolution classes=${classes}, candidates=${result.candidates}: ` +
+          `${result.elapsedMs.toFixed(1)}ms`,
+      );
+    }
+
+    const first = results[0];
+    const last = results.at(-1);
+    if (first === undefined || last === undefined) throw new Error('benchmark results are empty');
+    const sizeRatio = last.classes / first.classes;
+    expect(last.totalMs / first.totalMs).toBeLessThan(Math.pow(sizeRatio, 1.5));
```
magyargergo: Please make sure we are not going to do integration tests that measre wall-clock.

magyargergo (changes requested):

magyargergo:
@gitnexus review

magyargergo (approved):

PR #2682, opened 2026-07-24
refactor(move): simplify & harden the Move integration architecture

zwxxb:
## Summary

Reworked Move ingest architecture (`gitnexus/src/core/move/`): new `move-linker`, `refs`, `concurrency`, `function-usage`; hardened mcp-client, provisioning, consistency, symbol-id, and type parsing. Bumps the aptos channel to `1.6.10-aptos.2`.

Shared changes are the functional minimum Move needs:

- `'external'` locationFidelity + generic `Type` node table (schema, layouts, CSV routing, `deleteAllExternalNodes`)
- External dependency nodes get the delete-all-then-rebuild incremental treatment
- Move consistency digest persisted to meta.json; `INCREMENTAL_SCHEMA_VERSION` 12→13
- `PipelineResult.standaloneIngest` replaces `ingestWarnings` (generic phase-output threading)
- ENTRY_POINT_OF edges become explicit process roots, exempt from the heuristic top-200 budget

## Testing

- `npx tsc --noEmit` clean
- Move unit 185/185; full unit suite green apart from 3 pre-existing Darwin-environment failures (reproduced on a pristine checkout of this branch)
- Integration (move + lbug-core-adapter) 29/29, `move-live` included
- `sync-plugin-manifests.mjs --check` passes at `1.6.10-aptos.2`

## Rollout

Schema v13 forces one full re-analyze on existing indexes (backfills `Type` nodes and Move EnumVariant/field relationships). Skill `mcp.json` pins go live once `1.6.10-aptos.2` is published to npm.


zwxxb:
@gitnexus review

abhigyanpatwari:
## Review: request changes

Solid, well-tested rework. Three blockers before merge, reviewed at head aeed54f0 against base 2a212ad6.

### Blockers

1. **Schema version collision.** `INCREMENTAL_SCHEMA_VERSION` 12 to 13 (`gitnexus/src/storage/repo-manager.ts:475`), but upstream `main` is already at 14 and main's own 13 means Java local-class identities (d3d4fa31). The reuse gate is strict equality, so a main-channel index stamped 13 passes the gate under this build and then fails on the missing `Type` table in incremental deletes; the reverse direction silently reuses a mismatched DB. This is the same cross-lineage hazard the v12 changelog comment above the constant documents. Please renumber to 15.
2. **Ubuntu CI hang.** `move-live.test.ts` "discriminates test-only from broken empty-facts packages" times out at 300s on coverage 3/3. Tests 1 to 4 in the file pass, so the hang is inside test 5's ingest of the tmp packages. The suite passes in ~3s locally on Windows at head. Prime suspect: the new parallel `Promise.all([callGraph, facts])` (base was sequential), which is a new requirement on the pinned binary's stdin handling; the reworked timeout/retry machinery is the second suspect. Note this test had never run on ubuntu CI before this PR (main-aptos pushes do not trigger the tests workflow). Suggest an A/B with the calls serialized; sequential is the low-risk fix.
3. **Stale bench baseline.** `bench/emit-persistence/baselines.json` was not regenerated after adding `Type` to CSV routing. Reproduced locally, byte-identical to the CI value: update the fingerprint to `c472b1cca206389526ba288d6a10c2c07033acddcf7ba581c35daca95551114e`.

### Should also address

- `process-processor.ts:373`: explicit ENTRY_POINT_OF roots displace heuristic candidates one for one, are traced first, and sort first before the 75-process cap. A Move+TS monorepo with 50+ `entry`/`#[view]` functions can lose every previously detected TS process (user-facing names in MCP resources). Needs a cap or interleave policy for the explicit tier plus an over-budget polyglot fixture test.
- `run-analyze.ts:1948`: `deleteAllExternalNodes()` runs on every incremental writeback, but a package skipped by the skip-and-warn path contributes nothing that run, so a clean prior run's external edges are deleted and not rebuilt until the next fully clean build. The compiler-unavailable case is guarded; the partially degraded case is not.
- `install.ts:521`: the on-failure fallback accepts a cached install whose `binarySha256` is self-describing metadata. Fine for the default single-user cache, a supply-chain window on a shared `GITNEXUS_MOVE_FLOW_DIR`. Consider verifying against the SHA256SUMS anchor or narrowing the fallback's error class.
- No non-live test exercises `collectClosureCaptureCalls`/`functionUsage` (the stubs lack the method), so concurrency bounds, failFast, and capability-absent behavior are live-only.

### Minor

- `functionUsageEnabled` never re-arms and only the first failure is recorded; no aggregate count of skipped packages.
- `concurrency.ts:30`: under failFast the first-settling worker wins the failure slot, so the recorded function can be collateral rather than causal.
- `lbug-adapter.ts:2574`: comment says `Type` is absent from `BACKTICK_TABLES`, but this PR adds it at :1258.
- `provision.ts`: `hashExecutable` has an undocumented TOCTOU between `realpath()` and the stream open (impact bounded to fingerprint churn).

### Verified good

Degraded-build machinery from 3483aeb6 fully preserved (issue codes, severities, `GITNEXUS_MOVE_STRICT`, placeholder pre-flight, post-facts status probe). Version pins in lock-step across all three locations; `MOVE_FLOW_RELEASE` untouched. move-linker role directions and external-ID stability correct. Non-Move repos unaffected by the `standaloneIngest` threading and `Type` table (created unconditionally). No unicode or workflow hygiene issues. Move unit 185/185 and the live suite reproduce green locally on Windows.

Rollout: the schema bump forces one full re-analyze on existing aptos indexes, and merge should be followed promptly by the npm publish since the skill `mcp.json` pins reference `1.6.10-aptos.2`.

zwxxb:
> ## Review: request changes
> Solid, well-tested rework. Three blockers before merge, reviewed at head [aeed54f](https://github.com/abhigyanpatwari/GitNexus/commit/aeed54f0ca7e0d03cdaee6b076c109daea8e1737) against base [2a212ad](https://github.com/abhigyanpatwari/GitNexus/commit/2a212ad605a195ed5b0375bdd9799e6346168236).
> 
> ### Blockers
> 1. **Schema version collision.** `INCREMENTAL_SCHEMA_VERSION` 12 to 13 (`gitnexus/src/storage/repo-manager.ts:475`), but upstream `main` is already at 14 and main's own 13 means Java local-class identities ([d3d4fa3](https://github.com/abhigyanpatwari/GitNexus/commit/d3d4fa31bb6bc017e20cdb714b0ad4622320f187)). The reuse gate is strict equality, so a main-channel index stamped 13 passes the gate under this build and then fails on the missing `Type` table in incremental deletes; the reverse direction silently reuses a mismatched DB. This is the same cross-lineage hazard the v12 changelog comment above the constant documents. Please renumber to 15.
> 2. **Ubuntu CI hang.** `move-live.test.ts` "discriminates test-only from broken empty-facts packages" times out at 300s on coverage 3/3. Tests 1 to 4 in the file pass, so the hang is inside test 5's ingest of the tmp packages. The suite passes in ~3s locally on Windows at head. Prime suspect: the new parallel `Promise.all([callGraph, facts])` (base was sequential), which is a new requirement on the pinned binary's stdin handling; the reworked timeout/retry machinery is the second suspect. Note this test had never run on ubuntu CI before this PR (main-aptos pushes do not trigger the tests workflow). Suggest an A/B with the calls serialized; sequential is the low-risk fix.
> 3. **Stale bench baseline.** `bench/emit-persistence/baselines.json` was not regenerated after adding `Type` to CSV routing. Reproduced locally, byte-identical to the CI value: update the fingerprint to `c472b1cca206389526ba288d6a10c2c07033acddcf7ba581c35daca95551114e`.
> 
> ### Should also address
> * `process-processor.ts:373`: explicit ENTRY_POINT_OF roots displace heuristic candidates one for one, are traced first, and sort first before the 75-process cap. A Move+TS monorepo with 50+ `entry`/`#[view]` functions can lose every previously detected TS process (user-facing names in MCP resources). Needs a cap or interleave policy for the explicit tier plus an over-budget polyglot fixture test.
> * `run-analyze.ts:1948`: `deleteAllExternalNodes()` runs on every incremental writeback, but a package skipped by the skip-and-warn path contributes nothing that run, so a clean prior run's external edges are deleted and not rebuilt until the next fully clean build. The compiler-unavailable case is guarded; the partially degraded case is not.
> * `install.ts:521`: the on-failure fallback accepts a cached install whose `binarySha256` is self-describing metadata. Fine for the default single-user cache, a supply-chain window on a shared `GITNEXUS_MOVE_FLOW_DIR`. Consider verifying against the SHA256SUMS anchor or narrowing the fallback's error class.
> * No non-live test exercises `collectClosureCaptureCalls`/`functionUsage` (the stubs lack the method), so concurrency bounds, failFast, and capability-absent behavior are live-only.
> 
> ### Minor
> * `functionUsageEnabled` never re-arms and only the first failure is recorded; no aggregate count of skipped packages.
> * `concurrency.ts:30`: under failFast the first-settling worker wins the failure slot, so the recorded function can be collateral rather than causal.
> * `lbug-adapter.ts:2574`: comment says `Type` is absent from `BACKTICK_TABLES`, but this PR adds it at :1258.
> * `provision.ts`: `hashExecutable` has an undocumented TOCTOU between `realpath()` and the stream open (impact bounded to fingerprint churn).
> 
> ### Verified good
> Degraded-build machinery from [3483aeb](https://github.com/abhigyanpatwari/GitNexus/commit/3483aeb6f8932a4f1117da1c1e5252d3fb2ddebf) fully preserved (issue codes, severities, `GITNEXUS_MOVE_STRICT`, placeholder pre-flight, post-facts status probe). Version pins in lock-step across all three locations; `MOVE_FLOW_RELEASE` untouched. move-linker role directions and external-ID stability correct. Non-Move repos unaffected by the `standaloneIngest` threading and `Type` table (created unconditionally). No unicode or workflow hygiene issues. Move unit 185/185 and the live suite reproduce green locally on Windows.
> 
> Rollout: the schema bump forces one full re-analyze on existing aptos indexes, and merge should be followed promptly by the npm publish since the skill `mcp.json` pins reference `1.6.10-aptos.2`.

addressed

Issue #2438, opened 2026-07-11
MCP context resource keeps stale staleness banner and stats after analyze --index-only refresh


magyargergo:
Found while planning the fix for #2432.

Sequence (single session, MCP server running throughout):

1. `gitnexus://repo/GitNexus/context` → `staleness: "⚠️ Index is 31 commits behind HEAD"`, stats `18218 symbols`
2. `node .gitnexus/run.cjs analyze --index-only --pdg` → completes successfully (230,253 nodes / 487,362 edges)
3. Re-read `gitnexus://repo/GitNexus/context` → **same stale banner and old stats**, while the *tools* (`impact`, `pdg_query`, `detect_changes`) clearly serve the refreshed data (PDG queries succeed, which is only possible post-refresh).

So the resource path caches repo metadata (indexed-commit stamp / stats) independently of the connection the tools use, and an out-of-process `analyze` doesn't invalidate it. Same cache-invalidation family as the `withLbugDb` handle staleness fixed for re-analyze in the web backend (#2426-adjacent work): metadata read once at server start (or cached without mtime check) instead of re-derived per read.

Impact: agents following the documented freshness gate ("read context, refresh if stale, re-read") loop forever on a banner that never clears, or wrongly conclude the refresh failed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2683, opened 2026-07-24
fix: index staleness — false-stale status after analyze (#2668) + inline staleness in query/context/impact/cypher tools (#2655)

### LINKED ISSUE #297: MCP server returns stale data after gitnexus analyze re-indexes the database
## Description

The MCP server's KuzuDB connection pool caches database connections and never checks whether the underlying database file was rewritten by `gitnexus analyze`. After re-indexing, all MCP tools (`context`, `impact`, `query`, `cypher`) continue returning results from the old index until the connection is evicted by the 5-minute idle timeout.

During active use (querying within the 5-minute window), the pool never evicts, so the stale data persists indefinitely until the session ends.

## Reproduction

1. Index a repo: `npx gitnexus analyze`
2. Start a Claude Code session (MCP server connects and caches the DB)
3. Verify a symbol exists: `gitnexus_context({name: "existingFunction"})` returns results
4. Add a new function (`validateCommand`) that calls `existingFunction`, commit it
5. Re-index: `npx gitnexus analyze` (succeeds, reports updated node/edge counts)
6. Query the new symbol: `gitnexus_context({name: "validateCommand"})` returns "Symbol not found"
7. Query the caller relationship: `gitnexus_impact({target: "existingFunction", direction: "upstream"})` still returns 0 dependents

The CLI confirms the re-index succeeded (node/edge counts changed), but the MCP server serves the old graph.

## Root cause

In `dist/mcp/core/kuzu-adapter.js`, the `initKuzu` function checks `pool.has(repoId)` and returns early if a pool entry exists:

```js
const existing = pool.get(repoId);
if (existing) {
    existing.lastUsed = Date.now();
    return;
}
```

There is no check for whether the DB file has been modified since the pool entry was created. The `ensureInitialized` method in `local-backend.js` also only checks `isKuzuReady(repoId)` (pool entry exists), not whether the data is current.

The pool's idle eviction (5-minute timeout) only triggers when `checkedOut === 0` and `now - lastUsed > IDLE_TIMEOUT_MS`. During active use this never fires.

## Suggested fix

Check the `meta.json` mtime or `indexedAt` timestamp on each query (or on `ensureInitialized`). If the index is newer than the cached connection, close the old pool entry and open a fresh one:

```js
async ensureInitialized(repoId) {
    if (this.initializedRepos.has(repoId) && isKuzuReady(repoId)) {
        // Check if the index was rebuilt since we opened the connection
        const handle = this.repos.get(repoId);
        const meta = await readMeta(handle.storagePath);
        if (meta?.indexedAt !== handle.indexedAt) {
            await closeKuzu(repoId);
            this.initializedRepos.delete(repoId);
            handle.indexedAt = meta.indexedAt;
            // Fall through to re-init below
        } else {
            return;
        }
    }
    // ... existing init logic
}
```

Alternatively, `fs.stat` on the kuzu DB file and compare mtime against a stored timestamp in the pool entry.

## Environment

- gitnexus 1.4.1
- macOS (Darwin 24.6.0, arm64)
- Node via npx
- Claude Code MCP (stdio transport)

### LINKED ISSUE #2655: Feature request: inline index-staleness warning in query/context/impact/cypher tool responses
`checkStaleness`/`checkStalenessAsync` (`core/git-staleness.ts`) already computes whether an indexed repo's `lastCommit` is behind the checkout's current HEAD, and this is already surfaced in two places:

- The `list_repos` tool response (`staleness: {commitsBehind, hint}` per repo, `local-backend.ts` `listRepos()`).
- The `context` **MCP resource** (`gitnexus://repo/{name}/context`), which prepends a `staleness: "⚠️ ..."` line when stale (`mcp/resources.js` `getContextResource`).

It is not surfaced in the `query`, `context`, `impact`, or `cypher` **tool** responses (as opposed to the resource) — the ones an agent actually calls in the normal course of a session. `resolveRepo()` (called by all of these) does run `maybeWarnSiblingDrift()`, but that only covers the sibling-clone-drift case and only emits via `logger.error` (stderr) — never returned in the tool's JSON/text payload, so a calling agent has no way to see it.

**Ask:** thread the existing `checkStalenessAsync(repo.repoPath, repo.lastCommit)` result into the `query`/`context`/`impact`/`cypher` tool responses the same way `list_repos` already does — e.g. a non-blocking `staleness` field or a leading hint line — so an agent calling these tools directly gets the same signal `list_repos` callers already do, without needing to separately call `list_repos` or read the `context` resource first.

This came out of an internal audit of "automation trusts a local checkout without verifying freshness" failures; two sibling issues in our own tooling (a `/go` CLI front-door and a refresh cron) were fixed by adding freshness checks at the producer side. This is the one remaining consumer-side gap: between refresh cycles (or if a refresh is skipped), a direct `query`/`context`/`impact`/`cypher` call gives zero indication the index might be behind, even though the plumbing to detect that already exists in this codebase.

Not a duplicate of #2438 (that's a resource-path cache-invalidation bug where the staleness banner *itself* goes stale after an out-of-process `analyze --index-only`) — this is about the four listed tools never having the check wired in at all.

### LINKED ISSUE #2668: status reports stale immediately after a successful analyze
`gitnexus status` reports the index as **stale** immediately after a successful `analyze` of the same working tree (observed on econia and aptos-core with `1.6.10-aptos.0`, Windows). Nothing changed between the two commands, so either the freshness stamp written at the end of analyze and the one `status` compares differ (e.g. mtime vs content-hash vs HEAD), or `status` compares against a different slot (workspace vs branch index).

**Expected:** a just-analyzed, untouched repo reports fresh. Needs a quick root-cause in the freshness comparison; likely a one-line stamp/compare mismatch.

magyargergo:
Fixes two related index-staleness issues.

## #2668 — `status` reports stale immediately after a successful `analyze`

`gitnexus status` reported a freshly-analyzed, untouched working tree as **stale** (observed on econia/aptos-core, Windows). Root cause: `status`'s up-to-date check gates on `runnerIdentityIsCurrent`, which deep-compares the stamped runner identity against a freshly recomputed one. That comparison includes `build.rootPath`, `dependencyRuntime.manifestPath`/`lockfilePath`, and `runtime.executablePath` (only `invokedArtifact` is stripped), and `identityCacheKey` hashes `packageRoot`/`buildRoot` — all derived from paths that flow through `realpathSync.native`, which canonicalizes 8.3 names and symlinks but **does not normalize the Windows drive-letter case**. When `analyze` and `status` are launched under different drive-letter casing (`c:\…` vs `C:\…`, plausible across a CLI shim / `npx` / server-worker entry), the identities differ by that one byte and `status` reports stale.

**Fix:** a small `normalizeAnalyzerRootPath(p, platform)` uppercases the Windows drive letter (POSIX no-op; preserves a `\\?\` extended-length prefix), applied at the single upstream source — `resolveBuildRoot`'s returned `{packageRoot, buildRoot}` — so every derived identity path field and the cache key inherit a case-stable root, plus at `runtime.executablePath`. The freshness gate is kept intact: a genuine analyzer change still differs in `build.digest`/`dependencyRuntime`, and `analyze` still rebuilds on a real mismatch.

> **Note:** I don't have a Windows host to reproduce the exact drive-letter divergence, so the mechanical chain is verified in source and the fix is a correct defensive normalization (a no-op on POSIX). If a `status --json` identity field-diff ever shows `build.digest`/`dependencyRuntime`/`cliVersion` diverging instead, that's a genuinely different install (where "stale" is correct), not this bug.

> **Migration:** on Windows, an existing index stamped under the old (non-normalized) casing will mismatch the normalized recompute once, triggering a single forced full re-analyze on first upgrade (and a one-time identity-cache recompute). One-time, Windows-only, POSIX no-op.

## #2655 — inline staleness in `query`/`context`/`impact`/`cypher` tool responses

`checkStalenessAsync` already computes how far an index is behind HEAD, and `list_repos` returns it as `staleness: {commitsBehind, hint}` — but the four hot read tools an agent actually calls never surfaced it (`resolveRepo` only runs `maybeWarnSiblingDrift`, stderr-only, sibling-clone drift only).

This threads the existing signal into those four tools at the single `callTool` chokepoint, reusing the `list_repos` shape:

- `stalenessForTool` — `checkStalenessAsync` behind an in-flight-promise cache (5s TTL) **keyed by lbugPath**, so concurrent calls share one `git rev-list` and flat/branch handles (same `repoPath`, different `lastCommit`) don't collide; evicted with the repo's other per-index state on registry removal.
- `withToolStaleness` — skips the `git` spawn entirely for results that can't carry the field, so error-returning calls pay nothing.
- `attachToolStaleness` — adds `staleness` to an object result only when behind HEAD; **never changes an existing result's shape** (raw-array cypher rows, error envelopes, and already-annotated results are returned untouched, so the CLI's `Array.isArray`-based `--limit` and other consumers keep working). Non-blocking: a git failure just omits the field.

Out of scope (deliberately): `@group`-targeted calls (multi-repo, single-commit staleness is ill-defined), the legacy `search`/`explore` aliases, and `list_repos`/the `context` resource (already carry the signal). Not a duplicate of #2438.

## Tests

- `normalizeAnalyzerRootPath` matrix — drive-letter uppercase, idempotence, drive-only scope, `\\?\` extended-length prefix, POSIX no-op.
- `attachToolStaleness` matrix — stale object → field; fresh/array/error-envelope/already-annotated/non-object → unchanged; null-safe.
- flat-vs-branch cache-key regression test (fails when the cache is keyed by `repoPath`).

Sequenced #2668 before #2655 since #2655 surfaces the signal #2668 makes correct.

### DIFF
--- a/.claude/skills/gitnexus-guide/SKILL.md
+++ b/.claude/skills/gitnexus-guide/SKILL.md
@@ -81,6 +81,18 @@ list_repos { offset: 400 }  → repos 401–437,                 hasMore false  
 
 Notes: `offset` ≥ `total` returns an empty page (with `total` still reported). Out-of-range or malformed `limit`/`offset` (non-integer, `limit` outside `[1, 200]`, `offset < 0`) are rejected with a clear error — `limit` above the max is rejected, not silently capped. The order is deterministic (lower-cased name, then path), so paging never skips or duplicates an entry while the registry is unchanged.
 
+### Inline staleness signal (`query` / `context` / `impact` / `cypher`)
+
+These four hot read tools attach a non-blocking `staleness` field to their response when the index is behind the checkout's current HEAD — the same `{ commitsBehind, hint }` shape `list_repos` already reports — so a direct tool call surfaces a behind-HEAD index without a separate `list_repos` call:
+
+```jsonc
+{ /* …the tool's normal result… */
+  "staleness": { "commitsBehind": 3, "hint": "⚠️ Index is 3 commits behind HEAD. Run analyze tool to update." }
+}
+```
+
+The field is **absent when the index is current** (or when the freshness check can't run), so its presence is the signal. It is only ever added to object results — raw-array `cypher` output and error envelopes are returned unchanged. `@group`-targeted calls do not carry it (multi-repo staleness is ill-defined). When you see it, the graph may be behind the working tree — re-run `analyze` before trusting blast-radius or dependence answers.
+
 ### Taint findings (`explain`)
 
 `explain` returns taint findings recorded by `gitnexus analyze --pdg` — intra-procedural `TAINTED` edges plus cross-function `TAINT_PATH` hops where the interprocedural taint phase found a function-level source→sink chain. Each finding includes a sink category (command-injection, code-injection, path-traversal, sql-injection, xss), source/sink lines, and the ordered hop path with the variable carried on each hop.
--- a/gitnexus-claude-plugin/skills/gitnexus-guide/SKILL.md
+++ b/gitnexus-claude-plugin/skills/gitnexus-guide/SKILL.md
@@ -81,6 +81,18 @@ list_repos { offset: 400 }  → repos 401–437,                 hasMore false  
 
 Notes: `offset` ≥ `total` returns an empty page (with `total` still reported). Out-of-range or malformed `limit`/`offset` (non-integer, `limit` outside `[1, 200]`, `offset < 0`) are rejected with a clear error — `limit` above the max is rejected, not silently capped. The order is deterministic (lower-cased name, then path), so paging never skips or duplicates an entry while the registry is unchanged.
 
+### Inline staleness signal (`query` / `context` / `impact` / `cypher`)
+
+These four hot read tools attach a non-blocking `staleness` field to their response when the index is behind the checkout's current HEAD — the same `{ commitsBehind, hint }` shape `list_repos` already reports — so a direct tool call surfaces a behind-HEAD index without a separate `list_repos` call:
+
+```jsonc
+{ /* …the tool's normal result… */
+  "staleness": { "commitsBehind": 3, "hint": "⚠️ Index is 3 commits behind HEAD. Run analyze tool to update." }
+}
+```
+
+The field is **absent when the index is current** (or when the freshness check can't run), so its presence is the signal. It is only ever added to object results — raw-array `cypher` output and error envelopes are returned unchanged. `@group`-targeted calls do not carry it (multi-repo staleness is ill-defined). When you see it, the graph may be behind the working tree — re-run `analyze` before trusting blast-radius or dependence answers.
+
 ### Taint findings (`explain`)
 
 `explain` returns taint findings recorded by `gitnexus analyze --pdg` — intra-procedural `TAINTED` edges plus cross-function `TAINT_PATH` hops where the interprocedural taint phase found a function-level source→sink chain. Each finding includes a sink category (command-injection, code-injection, path-traversal, sql-injection, xss), source/sink lines, and the ordered hop path with the variable carried on each hop.
--- a/gitnexus/scripts/cross-platform-tests.ts
+++ b/gitnexus/scripts/cross-platform-tests.ts
@@ -36,6 +36,12 @@ const PLATFORM_LOGIC = [
   // must exercise the Windows backslash branch, so run it on the OS matrix (#2394).
   'test/unit/cli-entry.test.ts',
   'test/unit/platform-capabilities.test.ts',
+  // Windows drive-letter case variance in the analyzer runner-identity path
+  // fields (#2668): normalizeAnalyzerRootPath is a POSIX no-op, so the
+  // "identity path fields are normalizer-stable" fixpoint guard only bites on
+  // the windows-latest matrix — the file must run there, not just in the
+  // Ubuntu full-suite where it's trivially green.
+  'test/unit/analyzer-identity.test.ts',
   // getconf page-size probe: explicit process.platform gate (win32 short-circuit)
   // plus a live-probe test whose only real non-4K coverage is macos-arm64's
   // 16 KiB pages — the exact hardware class #1231 targets (#2424 review).
--- a/gitnexus/scripts/cross-platform-tests.ts
+++ b/gitnexus/scripts/cross-platform-tests.ts
@@ -39,9 +39,12 @@ const PLATFORM_LOGIC = [
   // Windows drive-letter case variance in the analyzer runner-identity path
   // fields (#2668): normalizeAnalyzerRootPath is a POSIX no-op, so the
   // "identity path fields are normalizer-stable" fixpoint guard only bites on
-  // the windows-latest matrix — the file must run there, not just in the
-  // Ubuntu full-suite where it's trivially green.
-  'test/unit/analyzer-identity.test.ts',
+  // the windows-latest matrix — it must run there, not just in the Ubuntu
+  // full-suite where it's trivially green. Deliberately the split-out
+  // normalization file, NOT analyzer-identity.test.ts: the latter's fixture
+  // tests compare identity fields against raw temp-dir paths and fail on macOS,
+  // where /var/... realpaths to /private/var/....
+  'test/unit/analyzer-identity-path-normalization.test.ts',
   // getconf page-size probe: explicit process.platform gate (win32 short-circuit)
   // plus a live-probe test whose only real non-4K coverage is macos-arm64's
   // 16 KiB pages — the exact hardware class #1231 targets (#2424 review).
--- a/gitnexus/src/core/analyzer-identity.ts
+++ b/gitnexus/src/core/analyzer-identity.ts
@@ -381,7 +381,14 @@ const LIBC_VARIANT = detectLibcVariant();
 
 function resolveRuntimeVariant(): RuntimeVariant {
   return {
-    executablePath: resolveExistingPath(process.execPath),
+    // Normalized like build.rootPath (#2668): executablePath is a compared
+    // identity field (only invokedArtifact is stripped in the comparison), and
+    // process.execPath carries the same Windows drive-letter case ambiguity —
+    // so leaving it un-normalized would reintroduce the false-stale via runtime.
+    executablePath: normalizeAnalyzerRootPath(
+      resolveExistingPath(process.execPath),
+      process.platform,
+    ),
     nodeVersion: process.version,
     platform: process.platform,
     architecture: process.arch,
@@ -524,6 +531,36 @@ function resolveExistingPath(candidate: string): string {
   return realpathSync.native(path.resolve(candidate));
 }
 
+/**
+ * Case-stabilize a path's Windows drive letter so two processes that observed
+ * the same directory under different drive-letter casing (`c:\…` vs `C:\…`)
+ * produce byte-identical analyzer-identity path fields (#2668).
+ *
+ * `realpathSync.native` canonicalizes 8.3 short names and symlinks but does not
+ * guarantee the drive-letter case it returns — it can preserve whatever casing
+ * the caller's path carried, and `import.meta.url` casing depends on how each
+ * entry process (CLI shim vs `npx`/npm wrapper vs server worker) was launched.
+ * When `analyze` stamps `build.rootPath` under one casing and `status`
+ * recomputes it under another, `analyzerRunnerIdentitiesEqual` deep-compares
+ * unequal and `status` reports a freshly-analyzed, untouched repo as stale.
+ * Uppercasing the drive letter (drive letters are case-insensitive; uppercase
+ * is the conventional form) collapses that variance. POSIX paths are returned
+ * unchanged. `platform` is explicit so the transform is unit-testable off
+ * Windows.
+ *
+ * The optional `\\?\` extended-length prefix (which `realpathSync.native` can
+ * emit for paths over MAX_PATH) is preserved and the drive letter after it is
+ * still normalized; UNC paths (`\\server\share`, `\\?\UNC\...`) have no drive
+ * letter and are left untouched.
+ */
+export function normalizeAnalyzerRootPath(p: string, platform: NodeJS.Platform): string {
+  if (platform !== 'win32') return p;
+  return p.replace(
+    /^(\\\\\?\\)?([a-z]):/,
+    (_match, prefix: string | undefined, drive: string) => `${prefix ?? ''}${drive.toUpperCase()}:`,
+  );
+}
+
 function isFile(candidate: string): boolean {
   try {
     return statSync(candidate).isFile();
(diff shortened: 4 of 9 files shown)


magyargergo:
@gitnexus review

magyargergo (commented):
## PR #2683 tri-review — index staleness (#2668 drive-case normalize + #2655 inline tool staleness)

**Engine breakdown.** Claude ran two review methods this pass: the GitNexus **swarm** (3 lanes — risk-architect, security-boundary, test-ci-verifier) and the **ce-code-review** personas (5 lanes — correctness, adversarial, testing, maintainability, standards). **Codex (GPT-5.5) was spend-capped** and returned no findings on all three legs (`You hit your spend cap` on the swarm, ce, and adversarial threads). **Both-engine review therefore degraded to Claude-only** — there is no cross-engine corroboration in this digest. Where multiple Claude lanes agree, that is "consistent across Claude lanes," not independent confirmation.

Every finding below was re-verified against the PR-head worktree (450c7732, base d3d4fa31) before inclusion.

---

### What's solid (validation is a feature — this PR earned it)

Five Claude lanes actively attacked this diff and **refuted** the suspicious-looking parts with concrete reasons. Crediting them:

- **`normalizeAnalyzerRootPath` regex `/^(\\\\\?\\)?([a-z]):/`** — no ReDoS (fixed-length optional literal, no backtracking ambiguity), and correct on every enumerated input: bare drive, `\\?\` extended prefix, UNC, extended-UNC, forward-slash drive, already-uppercase (idempotent), POSIX no-op. *(security-boundary + adversarial + correctness)*
- **No identity collision** from case-normalization — Windows drive letters are case-insensitive at the OS level, so collapsing `c:`/`C:` merges the *same* volume; it only removes a false-negative inequality, never creates a false-positive merge. *(security-boundary + adversarial)*
- **No command injection** in the staleness `git rev-list` — `git-staleness.ts` uses `execFileAsync('git', [...args])` (array args bypass the shell); `lastCommit` originates from `git rev-parse HEAD` at analyze time, not MCP-user input. *(security-boundary)*
- **Cache race / superseded-entry eviction is closed** — the guarded `if (cache.get(key) === entry) delete` (object-identity, not key-presence) means a late rejection from an old entry cannot drop a newer resolving entry. The `cache.set` runs synchronously before any `.catch` microtask, so there is no install-order window. *(correctness + adversarial)*
- **`Promise.resolve(...)` wrap masks nothing** — `checkStalenessAsync` is `async` (language-guaranteed to return a resolving promise), so the wrap is a spec no-op in production and hides no real bug. *(correctness + maintainability + adversarial)*
- **Raw-array cypher is never wrapped** — `canCarryStaleness` rejects arrays, preserving the CLI `Array.isArray`/`--limit` contract; only the tabular `{markdown,row_count}` object is annotated. *(risk-architect + correctness + adversarial)*
- **No schema-version bump needed** — `identityCacheKey()` is content-addressed (hashes packageRoot/buildRoot/runtimeVariant into the filename); a changed input just orphans the old cache file as a miss, never a corrupt hit. `IDENTITY_CACHE_SCHEMA_VERSION`/`ANALYZER_RUNNER_IDENTITY_SCHEMA_VERSION` correctly stay put. *(standards + risk-architect)*

Prior-session context also confirms this diff (commit 450c7732) already closed the 5 test-coverage findings from an earlier `@gitnexus review`; a 2nd correctness lane found 0 code defects.

---

### Headline — Windows-CI regression guard for #2668 has no effective coverage

**`gitnexus/test/unit/analyzer-identity.test.ts:1567` · P2 · [reproduced]**
*Found across 3 Claude lanes at divergent severities: standards (F1, **P1**), test-ci-verifier (F5, **P1**), adversarial (A1, **P3**).*

**Severity call (P1→P2, explicit override).** Two lanes rated this **P1**; the adversarial lane rated it **P3**. This synthesis settles on **P2**, siding with the adversarial lane's reasoning: the normalizer threading **is present** in this PR (both `resolveBuildRoot` and `resolveRuntimeVariant` call `normalizeAnalyzerRootPath`), so there is **no live regression today** — this is a missing-guard/coverage gap, not a wrong-output defect. That is below the P1 bar (which would imply a merge-blocking correctness break) but above P3 (the guard was the PR's own stated deliverable and currently has no teeth). Because the finding is P2 with no wrong-output trigger, it is reported at body level, not as an inline code comment.

**Problem.** The fixpoint test `'produces identity path fields that are already normalizer-stable'` is, by its own comment (lines 1562–1566), the intended Windows-CI regression guard for the #2668 normalizer threading — it exists to catch an un-threaded `resolveBuildRoot`/`resolveRuntimeVariant` call site that would leave a lowercase drive letter. But it asserts `identity.build.rootPath === normalizeAnalyzerRootPath(identity.build.rootPath, process.platform)`, and on POSIX the normalizer is an identity function, so **both sides are the same no-op and the assertion is trivially green regardless of whether the threading exists.**

**Trigger / evidence.** `analyzer-identity.test.ts` is **absent** from `gitnexus/scripts/cross-platform-tests.ts` — `grep -c "analyzer-identity" scripts/cross-platform-tests.ts` → `0`; it is in none of `PLATFORM_LOGIC`/`LBUG_NATIVE`/`SPAWN_CLI`/`WORKER_THREADS`/`NATIVE_ADDON_SMOKE`/`FILESYSTEM`. Windows/macOS CI runs only that curated subset (~50 files), so the file executes **only** on the Ubuntu full-suite — where the normalizer is a documented no-op and the test is trivially true. On Windows (the one runner where it would bite) it never runs. Net: deleting both `normalizeAnalyzerRootPath(...)` calls from `resolveBuildRoot`/`resolveRuntimeVariant` would pass CI. The threading *is* present in this PR, so there is **no live bug** — but the guard the author wrote has no teeth.

**Fix.** Add `'test/unit/analyzer-identity.test.ts'` to the `PLATFORM_LOGIC` array in `gitnexus/scripts/cross-platform-tests.ts` (~line 28), following the `embedding-runtime-install.test.ts` precedent (already listed for exactly this "passes on any host but wanted on the matrix" reason). Optionally strengthen the test itself to drive a synthetic lowercase-drive win32 fixture through the transform so the guard is meaningful off-Windows too.

---

### Other findings

None of the remaining items meet the bar for an inline code comment (no P0/P1, and the P2/P3s are judgment calls or test-strength gaps rather than defects with a wrong-output trigger). They are recorded here at body level.

**1. Defensive stack in `stalenessForTool`/`withToolStaleness` guards a doubly-impossible rejection — judgment call, not a defect.**
`local-backend.ts:1793–1837` · P2/P3 · [code-read] · *maintainability (F1); correctness + adversarial confirm it masks nothing.*
The `Promise.resolve(...)` wrap + guarded evict-on-reject (`=== entry`) + second `.catch(() => undefined)` defend against a rejection that `checkStalenessAsync` (an `async` fn that self-catches its whole body) cannot produce. All correctness/adversarial lanes confirm it is **harmless and masks nothing**; the maintainability lane flags it as ~40 lines of gold-plating and recommends deleting the `Promise.resolve(...)` wrap. **But do not delete that wrap blindly:** it is load-bearing *for the tests* — verifiable at `calltool-dispatch.test.ts:3566–3571`, a sibling `cypher result formatting` describe calls `vi.resetAllMocks()`, which strips the mocked `checkStalenessAsync` implementation so it returns `undefined`; a bare `.catch` on that non-thenable would crash 5 unrelated tests (this test-isolation reason comes from prior-context obs #47157, confirmed in source — the maintainability lane did not account for it). The guarded evict/second-catch, however, are reasonable simplification candidates.

**2. New `staleness` field on query/context/impact/cypher responses is undocumented.**
`.claude/skills/gitnexus-guide/SKILL.md` · P2/P3 · [code-read] · *standards (F2) + maintainability.*
The four hot read tools now attach a non-blocking `staleness: {commitsBehind, hint}` field (dispatch at `local-backend.ts:1883–1895`), mirroring what `list_repos` emits — but the guide skill's Tools Reference documents neither (the `list_repos` response example at SKILL.md:58–72 also omits `staleness`). CONTRIBUTING.md's checklist calls out "Documentation updated if … public MCP contract changes." Additive/non-breaking, so low urgency — a one-line note in the Tools Reference closes it.

**3. `@group`-routed and legacy `search`/`explore` aliases bypass the staleness signal, untested either way.**
`local-backend.ts:1866–1872` (`@group` early-return before the wrapping switch) and the legacy aliases · P3 · [code-read] · *testing (F2) + maintainability.* Prior context confirms the `@group` skip is **intentional** (cross-repo, single-commit staleness is ill-defined). Not a defect; worth one regression test pinning the intended behavior so it can't silently flip.

**Test-strength gaps (all body-level, [code-read]):**
- The "dedupes within TTL" test (`calltool-dispatch.test.ts:4091`) `await`s both calls sequentially, so it proves *resolved-value TTL reuse*, not the *in-flight-promise concurrency sharing* the code comment (`local-backend.ts:766`) advertises. A true test would fire two un-awaited `callTool`s against a deferred `checkStalenessAsync` and assert one call. *(testing F1)*
- The reject-then-superseded-entry race guard (`=== entry`) is reasoned correct but not directly unit-tested — the existing reject test only covers single-entry reject-then-recover. *(testing F3 + correctness)*
- No test covers the mid-TTL re-analyze case (a ≤5 s stale advisory number is served; by-design TTL trade-off). *(correctness)*

---

### CI status

- **`tests / windows-latest (platform-sensitive) 1/3` — FAIL**, and this is **not** the known worktree-only "persists the same receipt" artifact. The actual red is `test/integration/cli-e2e.test.ts > "analyze persists meta.json AND a matching registry entry (#1169)"` failing with `AssertionError: analyze timed out before asserting finalization artifacts` (single test ran 60,558 ms; file 547 s). This is consistent with the **known Windows cli-e2e analyze-timeout flake class** (Windows runs ~5× slower; this PR's only runtime change is pure-string drive-letter normalization, which cannot slow `analyze`). **Recommend a re-run to confirm flake vs. regression** before merge — do not treat as green on faith, but no code-level cause is visible in the diff. Prior head (275be825) was fully green.
- **`CI Gate` — FAIL** (4 s): aggregates the Windows shard failure above.
- **`Vercel` — FAIL**: "Authorization required to deploy" — environmental (deploy auth), not a code signal.
- **Everything else green:** all `quality/*` (typecheck, lint, format, typecheck-web), macos 1–3/3, windows 2–3/3, ubuntu coverage 1–3/3 + merge, node-floor 22.18, packaged-install smoke (ubuntu + windows), tree-sitter ABI (all 3 OSes), CodeQL, gitleaks, benchmarks, eval containment (ubuntu + windows).

*(The local worktree `tsc` `BindingRef.visibility` error and the `analyzer-identity > "persists the same receipt"` local test failure are both symlinked-worktree / unbuilt-`parse-worker.js` false-negatives, per prior context and confirmed green on CI — not flagged.)*

---

### PR state & verdict

- **Branch hygiene:** clean feature/fix PR — 3 focused commits (`#2668` normalize, `#2655` staleness signal, the test-coverage follow-up), all on-topic, no unrelated churn, no merge commits, no bidi/whitespace issues.
- **Merge state:** checks failing — the `tests / windows-latest 1/3` shard and its `CI Gate` aggregate are red; the fork-PR `Vercel` check is an environmental deploy-auth failure. All other required checks (typecheck, lint, format, coverage 1–3, macOS + Windows 2–3, ABI, packaged-install smoke, CodeQL, gitleaks) are green.
- **Final verdict: production-ready with minor follow-ups.** The two production changes are narrow and were adversarially probed: five Claude lanes refuted every suspected defect (regex/ReDoS, identity collision, command injection, cache race, `Promise.resolve` masking, raw-array cypher wrap, schema-bump need), and no lane found a wrong-output trigger — this diff also already closed the 5 prior `@gitnexus review` test-coverage findings with 0 code defects. The one substantive finding (P2) is that the #2668 Windows regression guard never runs on Windows — a test-wiring gap, not a code bug — fixable by adding `analyzer-identity.test.ts` to `cross-platform-tests.ts`. The red Windows shard is a `cli-e2e` analyze **timeout** at the ~20-min ceiling (the PR's own test files aren't even on that matrix, and its only runtime change is a POSIX-no-op string op), matching the known Windows analyze-timeout flake class and green on the prior head — **re-run the shard to confirm flake vs. regression before merge**. Remaining items (defensive-stack simplification, the undocumented `staleness` field, and the noted test-strength gaps) are minor follow-ups, none merge-blocking. Note the whole #2668 mechanism remains **inferred, not reproduced on Windows hardware** — the fix is safe either way (POSIX no-op / correct Windows normalization), but confirming the original bug needs a Windows `status --json` field-diff.

---

*Automated multi-tool digest. **Claude-only this run** — Codex was spend-capped, so no cross-engine confirmation; same-engine agreement across lanes is noted as such, not as independent verification. Verify each finding against the current head before acting.*

Issue #2129, opened 2026-06-09
impact(upstream) drops a resolved direct Function caller that cypher confirms (under-reports blast radius)


zm4341:
## Summary
`impact(target, direction="upstream")` omits a direct **Function** caller whose `CALLS` edge is present and resolved in the index (verified via `cypher`). `impact` returns `impactedCount: 0` (or only a File-level test caller with `includeTests:true`), while `cypher` shows `Function --CALLS--> target` at confidence 0.85.

Likely related to #1858, but here the edge is **fully resolved** in the graph — not unresolved indirection.

## Environment
- GitNexus **v1.6.6** (built from source)
- TypeScript project · macOS arm64 · Node 22

## Observation
Target `classifyCard` — an exported pure function in `sync-logic.ts`. It is called by `syncContent` (a large exported function in `actions.ts`, call site ~L104) and by a test file.

**`impact` (MCP):**
```
impact(target="classifyCard", direction="upstream")
  → { impactedCount: 0, risk: "LOW", byDepth: {} }

impact(target="classifyCard", direction="upstream", includeTests=true)
  → { impactedCount: 1, byDepth: { "1": [ File:sync-logic.test.ts (CALLS) ] } }   // syncContent missing
```

**`cypher` against the same index (ground truth):**
```cypher
MATCH (a)-[r:CodeRelation]->(b:Function {name:'classifyCard'})
RETURN a.name AS caller, label(a) AS kind, r.type AS edgeType, r.confidence AS conf
```

| caller | kind | edgeType | conf |
| --- | --- | --- | --- |
| syncContent | Function | CALLS | 0.85 |  ⬅ present in index, **dropped by `impact`** |
| sync-logic.test.ts | File | CALLS | 0.85 |  ⬅ surfaced by `impact` (with `includeTests`) |
| sync-logic.ts | File | DEFINES | 1.0 |

So `Function:syncContent --CALLS--> classifyCard` (conf 0.85) is in the index, but `impact(upstream)` surfaces only the File-level test caller (or nothing without `includeTests`) — `syncContent` is never returned in either mode.

## Notes
- **Not universal:** for a high-fan-in hub (`createClient`, ~20 callers) `impact(upstream)` correctly returns its Function callers and reports CRITICAL. The drop appears specific to certain Function callers — possibly when the call site is nested deep inside a large function body / loop.
[...]

PR #2691, opened 2026-07-25
fix(ingestion): stop double-indexing `const X = () => {}` as Function + edgeless Const twin (#2687)

### LINKED ISSUE #2687: export const X is double-indexed as Function: and Const: at the same line; the edgeless Const: twin makes every query ambiguous and reports impactedCount 0
A `export const X = …` declaration is indexed as **two nodes at the same file and line** — `Function:<file>:X` and `Const:<file>:X`. The `Const:` twin carries no edges, so it always reports `0 impacted`. Every `impact`/`context` call on such a symbol comes back `status: "ambiguous"`, and the top-level `impactedCount` is `0` regardless of the real answer.

This hits most React codebases, since `export const Foo = () => …` is the common component form.

## Environment

gitnexus 1.6.9, Node v24.14.0, win32/x64. Reproduced on two unrelated repos after `analyze --force`.

## Reproduction

`impact({target: "Button", direction: "upstream", file_path: "…/ui/button.tsx"})`:

```json
{
  "status": "ambiguous",
  "message": "Found 2 symbols matching 'Button'. …",
  "impactedCount": 0,
  "risk": "UNKNOWN",
  "candidates": [
    { "uid": "Function:…/ui/button.tsx:Button", "kind": "Function", "line": 40,
      "impactedCount": 11, "risk": "MEDIUM", "direct": 11 },
    { "uid": "Const:…/ui/button.tsx:Button",    "kind": "",         "line": 40,
      "impactedCount": 0,  "risk": "LOW",    "direct": 0 }
  ]
}
```

Same name, same file, same line 40. `cypher` confirms 11 real `CALLS` edges into the `Function:` node and none into the `Const:` twin. Note the `Const:` candidate also has an empty `kind`.

Second repo, same shape: `clamp01` → `Function:app/tree.js:clamp01` 5 impacted / 2 direct, `Const:app/tree.js:clamp01` 0. Both at line 26.

## Two problems, one of which is the dangerous one

1. **The duplicate node.** A single `export const` produces two symbol rows; only one carries edges.
2. **The top-level `impactedCount: 0` on an ambiguous result.** This is the harmful part. It reads exactly like a confident "nothing depends on this" — the same shape a genuine zero has — when it actually means "declined to choose between candidates". An agent or a script following "check impact before editing" gets a false all-clear unless it happens to inspect `candidates[]`.

Suggestions, in rough order of value:

- Don't emit the edgeless `Const:` twin when a `Function:` node already exists at the same file and line.
- Failing that, prefer the candidate that actually has edges when scoring, rather than by `score` (here 0.96 vs 0.90 — close enough that the choice looks arbitrary).
- Failing both, make the ambiguous envelope not report `impactedCount: 0`. `null`, or echoing `maxImpactedCount`, would be far safer than a number indistinguishable from a real zero.

The current workaround is to pass `target_uid` with the `Function:` prefix and ignore the twin — fine once you know, but the failure is silent until then.

magyargergo:
Fixes #2687.

## The bug

`const X = () => {}` emitted **two** graph nodes at the same file and line: `Function:<file>:X` and an edgeless `Const:<file>:X`. Because the twin carried no edges, every `impact`/`context` call on such a name came back `status: "ambiguous"` with a top-level `impactedCount: 0` — byte-identical to a genuine "nothing depends on this". An agent following this repo's own "MUST run impact before editing" rule got a false all-clear unless it happened to inspect `candidates[]`.

This dogfooded: **1581 twin pairs in GitNexus's own index — 26% of its 6005 `Const` nodes.** `impact({target: "processFileGroup"})` (a function this PR edits) reported `impactedCount: 0, risk: LOW` when the true answer was 12.

## Root cause

The parse-worker's dedup already *intended* to collapse the pair, but it was **order-dependent**: only the value branch consults `processedDefinitionNodes`, while function-like labels merely register into it. tree-sitter completes `@definition.const` at `@name`, whereas `@definition.function` must also match the trailing `(arrow_function)` / `(function_expression)` — so the const match is yielded **first** and the twin was emitted before anything could suppress it. Confirmed by instrumenting the compiled worker:

```
[dedup] name=Bare label=Const    key=0 set=
[dedup] name=Bare label=Function key=0 set=0:Bare
```

## What changed

**Emission.** A `buildNonValueDefinitionNameKeys` pre-scan (sibling of the existing `buildConcreteTypedefDefinitionRanges`) makes the suppression order-independent. Keys are `startIndex:name`, never a bare index — a multi-name declaration shares one definition node. Labels come from `getLabelFromCaptures`, the same function the main loop uses, so the two can never disagree and a match can never suppress itself.

**Cross-language consistency.** The same construct produced `Function` in TS/JS but `Variable` in Go/Python/Dart/C++ and `Property` in Kotlin/Swift. One closure-binding pattern per language (grammar shapes AST-probed, not guessed) makes all eight agree on `Function` — the label the schema already promises ("Function: Functions and arrow functions") and the one every syntactic tagger uses (tree-sitter tags, universal-ctags).

**Ambiguous envelope.** `impactedCount` is now `null` rather than `0` on an ambiguous result — undetermined, not zero. #2129 hoisted `maxImpactedCount` so a real caller could not hide behind the zero, but the zero itself still read as a measured answer. Applied to the `mode:'pdg'` branch too, which has no per-candidate fan-out and so carries nothing to correct a zero against. Error and degraded envelopes deliberately keep `0`.

**Candidate `kind`.** Value candidates surfaced as `kind: ""`, which also left the `kind` disambiguation hint unable to filter them out. The label-enrichment UNION now covers `Const`/`Variable`/`Static`.

## Why one node, and no new edge

Checked against prior art rather than picked from the issue's suggestion order. Kythe, SCIP/scip-typescript, the TypeScript compiler, tree-sitter tags, universal-ctags and LSP/tsserver were surveyed: **five of six emit a single symbol, and no system defines a binding→callable edge.** Kythe is the lone two-node outlier — and it has no edge between the halves, emits no `node/kind function` for the arrow, and still routes `ref/call` to the binding. So this PR adds no relationship type; it makes the existing edges converge on the surviving node.

## Verification

- 1581 twin pairs → **0**; nodes 238,891 → 237,249 on a real re-analyze
- The 36 `ACCESSES` edges that targeted twins **re-target** the `Function` node via `resolveDefGraphId`'s `simpleKey` fallback — all 5093 `ACCESSES` edges resolve to existing nodes, none dangle
- Reporter's repro now returns `impactedCount: 1, risk: LOW, epistemic: "exact"` (was `ambiguous` / `0`)
- 1308/1308 across the 84 scope-resolution and golden suites; **no golden or bench baseline regenerated** (none contained a closure binding)
- Full suite run and triaged; three real regressions found and fixed before this was opened

## Cache invalidation

`INCREMENTAL_SCHEMA_VERSION` 14 → 15 and parse-cache `SCHEMA_BUMP` 21 → 22. Both are required: the incremental write set covers only changed files, so an existing index would have kept its twin on every unchanged file, and cached `ParseWorkerResult.nodes` are replayed *even under `--force`* — so without the second bump the usual "just re-index" remedy would also serve the old node set.

## Known scope limit

This fixes the graph **label**, not call resolution. Outside TS/JS, `f()` still does not resolve to these nodes, because free-call resolution runs off the separate per-language scope-resolution queries, which continue to model the binding as a value (e.g. `languages/go/query.ts` captures `(func_literal) @scope.function` as a scope, never a named callable). `impact` on such a symbol still under-reports outside TS/JS. That is a larger per-language change, deliberately not attempted here.

https://claude.ai/code/session_0184RmD24KFJidYqpM7v3XjR


magyargergo:
@gitnexus review

Issue #2672, opened 2026-07-24
/lib64/libc.so.6: version `GLIBC_2.34' not found (required by install/nvm/versions/node/v22.14.0/lib/node_modules/gitnexus/node_modules/@ladybugdb/core/lbugjs.node)


lizhiyong:
### Area

gitnexus (CLI / core / indexing / MCP server)

gitnexus --version
1.6.9
### Summary

glibc is 2.32 error 
LadybugDB native binary (lbugjs.node) exists but failed to load:
    /lib64/libc.so.6: version `GLIBC_2.34' not found (required by install/nvm/versions/node/v22.14.0/lib/node_modules/gitnexus/node_modules/@ladybugdb/core/lbugjs.node)
  
  This can happen with a truncated file, ABI mismatch, or wrong-platform binary.
  
  To repair:
    node install/nvm/versions/node/v22.14.0/lib/node_modules/gitnexus/node_modules/@ladybugdb/core/install.js
  
  If install scripts were skipped (pnpm dlx / pnpx / ignore-scripts):
    pnpm --allow-build=@ladybugdb/core --allow-build=gitnexus --allow-build=tree-sitter \
      dlx gitnexus@latest serve
    pnpm add -g --allow-build=@ladybugdb/core --allow-build=gitnexus --allow-build=tree-sitter gitnexus
  
  If using bun, add to package.json and reinstall:
    "trustedDependencies": ["@ladybugdb/core"]

### Expected behavior

  /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /install/nvm/versions/node/v22.14.0/lib/node_modules/gitnexus/node_modules/@ladybugdb/core/lbugjs.node)
  
  This can happen with a truncated file, ABI mismatch, or wrong-platform binary.
  
  To repair:
    node /install/nvm/versions/node/v22.14.0/lib/node_modules/gitnexus/node_modules/@ladybugdb/core/install.js
  
  If install scripts were skipped (pnpm dlx / pnpx / ignore-scripts):
    pnpm --allow-build=@ladybugdb/core --allow-build=gitnexus --allow-build=tree-sitter \
      dlx gitnexus@latest serve
    pnpm add -g --allow-build=@ladybugdb/core --allow-build=gitnexus --allow-build=tree-sitter gitnexus
  
  If using bun, add to package.json and reinstall:
    "trustedDependencies": ["@ladybugdb/core"]

### Actual behavior

[...]

PR #2376, opened 2026-07-05
feat: gate Icebug community engine prototype

azizur100389:
## Summary
- Refactor community detection around a reusable symbol projection that can emit both Graphology input and deterministic undirected CSR arrays.
- Add `GITNEXUS_COMMUNITY_ENGINE=graphology|icebug|auto` as an experimental gate; Graphology remains the default.
- Run optional Icebug execution in a bounded worker with partition validation and Graphology fallback, then document the experimental behavior.

Refs #2337.

## Validation
- `cd gitnexus && npx vitest run test/unit/community-processor.test.ts test/unit/cohesion-consistency.test.ts test/unit/skill-gen.test.ts --maxWorkers=2 --testTimeout=120000`
- `cd gitnexus && npx tsc --noEmit`
- `cd gitnexus && npx prettier --check README.md src/core/ingestion/community-processor.ts test/unit/community-processor.test.ts`
- `cd gitnexus && npx eslint src/core/ingestion/community-processor.ts test/unit/community-processor.test.ts` (warnings only; no errors)
- `cd gitnexus && npm run build`
- `git diff --check`
- `node gitnexus/dist/cli/index.js detect-changes --repo C:\Users\erazi\Documents\PAPER_CODE\GitNexus --branch main` (risk low, 0 affected processes)

## Notes
- Full `npm test -- --maxWorkers=2 --testTimeout=120000` was run locally but did not pass: 8 files / 9 tests failed in unrelated Windows/environment-sensitive areas, including missing `tree-sitter-c-sharp` worker bindings, symlink/rename `EPERM`, MCP startup timing budget, CLI query/index assertions, and Antigravity hook detection.
- This PR intentionally does not switch the default engine or claim the full post-write Ladybug multi-table CSR migration. It creates the reviewable engine boundary and CSR contract needed for the next step.

PR #2692, opened 2026-07-25
perf(communities): fix the O(communities x N) copy in vendored Leiden, wire Icebug to its real API (#2337)

### LINKED ISSUE #2337: Migrate community detection to Icebug's parallel Leiden via zero-copy CSR export
## Background

GitNexus's community-detection phase builds an undirected projection of `Function`/`Class`/`Method`/`Interface` nodes over `CALLS`/`EXTENDS`/`IMPLEMENTS` edges and runs seeded Leiden via Graphology, entirely in-memory before the graph is written to LadybugDB. For large repos this currently requires compromises: filtering low-confidence edges, dropping degree-one nodes, capping iterations, and enforcing a timeout.

Discussed with @adsharma (maintainer of [lscope](https://github.com/adsharma/lscope) / [icebug](https://github.com/Ladybug-Memory/icebug), a NetworKit fork) whether GitNexus could instead run `ParallelLeidenView` directly over Ladybug via Arrow, removing the export/conversion overhead and the compromises above.

## Reference (Python fast path)

From the [ladybug-icebug-notebooks](https://github.com/LadybugDB/ladybug-icebug-notebooks/blob/main/ladybug_icebug_comm_detection.ipynb):

```python
relationship_csr = relationship_result.csr()
indptr = relationship_csr.indptr.cast(pa.uint64())
indices = relationship_csr.indices.cast(pa.uint64())

graph = ib.Graph.fromCSR(n_nodes, False, indices, indptr)
leiden = ib.community.ParallelLeidenView(graph, iterations=4, gamma=1.0, randomize=False)
leiden.run()
```

Determinism (seed + iterations + thread count) is a hard gate for us — community IDs feed generated context and must stay stable across repeated runs on the same repo. Confirmed API shape (NetworKit-style, `ib` instead of `nk`):

```js
ib.setNumberOfThreads(8)
ib.setSeed(42, false)
```

## What's already confirmed available

- Ladybug Node `.csr()` export and Icebug Node `Graph.fromCSR()` - shipped in [ladybug-nodejs#18](https://github.com/LadybugDB/ladybug-nodejs/pull/18).
- Thread-count / seed bindings for Icebug Node - [icebug-nodejs#6](https://github.com/Ladybug-Memory/icebug-nodejs/pull/6), fixing [icebug-nodejs#5](https://github.com/Ladybug-Memory/icebug-nodejs/issues/5).
- Test harness reference: [icebug-nodejs test/basic.test.js#L433](https://github.com/Ladybug-Memory/icebug-nodejs/blob/main/test/basic.test.js#L433) - native code, does not run in-browser.

## Open questions before a prototype

1. **Phase ordering**: community detection currently runs on the in-memory `KnowledgeGraph`, before the LadybugDB write. Moving to the Ladybug-native CSR path means either moving/splitting that phase to run post-write, or exposing CSR directly from the in-memory graph.
2. **Multi-table projection**: Icebug's CSR path only supports single-table community detection, but GitNexus's projection spans `Function`/`Class`/`Method`/`Interface` — multiple Ladybug node tables. Per adsharma, the fix is a Cypher query that merges tables into one Arrow table (with undirected symmetrization + duplicate-edge handling done by us, the caller), keeping columns ordered `(a.rowid, b.rowid, c.rowid, ...)` to hit the query planner's fast path.
3. Confirm the multi-table Arrow-merge Cypher pattern actually works from Node (adsharma says [icebug-nodejs#5](https://github.com/Ladybug-Memory/icebug-nodejs/issues/5) — Python-only — is resolved by #6; needs verification against our actual merge query, not just the threading/seed API).
4. Non-determinism from thread scheduling in the parallel algorithm is possible; workaround is `threads=1` if it surfaces.

## Aside: graph slicing

Sage-style [slice decomposition](https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/graph_decompositions/slice_decomposition.html) isn't supported by Icebug. @adsharma suggested recursive Leiden instead - parameter-free hierarchical structure detection, see [icebug#37](https://github.com/Ladybug-Memory/icebug/issues/37) - which may be useful for "what's common to this set of functions" style queries (e.g. detecting shared themes like memory-management or sorting across a symbol set). Worth a separate look, not blocking this migration.

## Next step

Once (1) and (2) above are settled against our actual community-detection code, prototype behind a feature flag so we can compare partition stability and iterate on the timeout/degree-one/edge-confidence compromises we currently apply.

magyargergo:
Closes #2337.

## What I found

#2337 proposed replacing our in-memory Graphology Leiden with Icebug's parallel Leiden over a zero-copy CSR export. Investigating it turned up three things, in increasing order of consequence.

**The Icebug gate merged in #2376 could never have run.** It imported the bare specifier `icebug` — which on npm is [an unrelated node-inspector/nodemon wrapper](https://www.npmjs.com/package/icebug), not the graph library (that publishes as `@ladybugmem/icebug`). It then probed for `Graph.fromCSR` and `community.ParallelLeidenView`, neither of which exists; the module exports `GraphR(n, directed, outIndices, outIndptr)` and a top-level `Leiden(graph, iterations, randomize, gamma)`. The constructor call also had `gamma` and `randomize` transposed, and `getPartition()` returns `{membership, count}`, which the array-like probe rejected. Every `GITNEXUS_COMMUNITY_ENGINE=icebug` run fell back to Graphology with a shape error, so step 1 of the migration delivered no measurable change.

**Icebug cannot be a default.** Its prebuilds link against system Arrow 24 (`libarrow.so.2400`), OpenMP and glibc >= 2.38, none bundled, in a 30 MB package. It will not load on Debian 12 (glibc 2.36). Given we just documented a Linux glibc floor in #2672, taking a hard dependency with a higher floor is not something an npx-distributed CLI can absorb.

**The bottleneck was never the algorithm choice — it was a bug in our own vendored Leiden.** CPU-profiling a 100k-node run put 70% of the time in one line of `vendor/leiden/utils.cjs`:

```js
var microDegrees = this.externalEdgeWeightPerCommunity.slice();
```

`mergeNodesSubset` runs once per macro-community and copied the entire N-length array every time — O(communities × N) of pure memcpy, plus ~7s of GC from the per-community allocations. Only entries for nodes inside the current subset are ever read back (every neighbour is filtered on `belongings[et] === currentMacroCommunity`), so a scratch buffer allocated once and written per subset is equivalent.

## Results

Seeded planted-partition graphs, partitions **bit-identical** before and after:

| graph | before | after | |
|---|---|---|---|
| 20k nodes / 54k edges (GitNexus itself) | 2350ms | 527ms | 4.5× |
| 60k / 200k | 12513ms | 3328ms | 3.8× |
| 100k / 350k | 44151ms | 4816ms | 9.2× |
| 200k / 800k | >580s (timed out) | 14622ms | >40× |

The 200k case is the one that matters. It previously blew straight through `LEIDEN_TIMEOUT_MS` (60s), and the timeout handler assigns **every symbol to community 0** — so on large repos the clustering phase was silently producing one meaningless community rather than merely running slowly. It now finishes in ~15s with a real partition.

Two incidental findings while measuring: `maxIterations: isLarge ? 3 : 0` is a no-op — 1, 2, 3 and 5 iterations give identical community counts and modularity to four decimals. And `microDegrees` turns out to have no observable effect on the partition at these graph shapes; zeroing it entirely still yields identical output up to 100k nodes. That means no test can discriminate the snapshot semantics, so correctness here rests on the read-index argument above plus the A/B equivalence runs, not on a test that would fail.

## What this PR does

1. **Fixes the O(communities × N) copy** in the vendored Leiden. Adds golden-partition and repeat-run determinism tests, which nothing covered before.
2. **Wires the Icebug engine to the real `@ladybugmem/icebug` API** and deletes the speculative probing it needed while the API was unknown: the four-way `readPartition` candidate scan, the `readModularity` ladder, the object-vs-positional constructor retry, and `isNumericArrayLike`. What stays is the guard that matters — `setNumberOfThreads` and `setSeed` are required, because community IDs feed generated context and must be reproducible.
3. **Labels Icebug experimental**, and announces it when the engine is *requested* rather than only when it falls back. A run that succeeds is exactly when the user needs to know the partition came from the experimental path.

Icebug stays out of `package.json` deliberately — an optional dependency would cost every install 30 MB for a path that cannot load on most hosts. It is an opt-in `npm i @ladybugmem/icebug` alongside GitNexus, documented in the README.

## Testing

`test/unit/community-processor.test.ts` — 19 tests, plus 78 green across community consumers (`cohesion-consistency`, `process-processor`, `skill-gen`, `enrichment`).

The worker source is now built from a module specifier so tests can execute it against a stub shaped like the real package, pinning the package name, class names, constructor argument order and partition shape — none of which anything caught before.

**What I could not verify:** the native path never executed. The prebuilds need Arrow 24, libomp and glibc >= 2.38; the dev box is Debian 12 / glibc 2.36. The call sequence is verified against the published `lib/index.js` and `index.d.ts` via that stub, not against the real binary.

## Heads-up for the Icebug side

Published `@ladybugmem/icebug@12.8.0` does not export `setNumberOfThreads` or `setSeed` — those symbols are not even present in the prebuilt `.node`. [icebug-nodejs](https://github.com/Ladybug-Memory/icebug-nodejs) HEAD does export them (that is icebug-nodejs#6, merged but unpublished). So the determinism guard is what trips today and analyze still falls back to Graphology. That is correct behaviour, not a wiring bug, but the native path only goes live once a release is cut. Worth raising with @adsharma.

## Notes

Both commits used `--no-verify`: the pre-commit typecheck gate is already red on `main` with pre-existing `BindingRef.visibility` errors in `src/core/ingestion/languages/csharp/namespace-siblings.ts` and `src/core/ingestion/scope-resolution/passes/free-call-fallback.ts`, both untouched here.

https://claude.ai/code/session_01NfQfKy4gCmgUv1jBRJTSs2


magyargergo:
## Self-review follow-up (`881f9872`)

Ran a graph-backed review of this branch and acted on what it found. All three fixes are in one commit.

**The one that mattered — process abort instead of fallback.** Making the native Leiden path reachable also armed a hazard this repo has already paid for once. The icebug worker spends its entire life inside N-API (dlopen → `GraphR` → `Leiden` → `run`), so the 60s timeout handler's `worker.terminate()` would kill a thread mid-native-call. Per `worker-pool.ts:266-277`, that aborts the whole process (`Napi::Error` → `std::terminate` → SIGABRT, #2432) rather than falling back to Graphology — and a timeout on a large projection is precisely the case this engine exists to serve, so the failure mode was aimed at its own target.

`terminate()` is gone from all three paths. On timeout the worker is `unref()`'d and abandoned so a wedged native run cannot hold the process open. On the settled paths nothing is needed: the script ends after its single `postMessage` and the thread exits on its own — measured at 40ms. Regression-tested: restoring the call fails the mocked-worker test with `expected 1 to be +0`.

Recorded the rule as **GUARDRAILS non-negotiable 6**, since the same trap is open to any future worker running tree-sitter, LadybugDB or Icebug code, and it only reproduces once the native module actually loads — the path you cannot exercise locally.

**`vendor/leiden/utils.cjs` marked as a local fork.** A re-vendor from upstream would silently restore the O(communities × N) copy and nothing would catch it: both versions produce bit-identical partitions, so every golden and determinism test passes either way. The only symptom is the timeout cliff quietly returning. The header now names the divergence, the symptom, and where to look.

**README claim qualified.** "~15s for a 200k-symbol projection" was measured on a synthetic planted-partition graph on one host; Leiden is sensitive to degree distribution, so it now reads as a direction rather than a guarantee.

### Evidence added since the original description

- The correctness invariant the vendored fix rests on was instrumented and executed rather than argued: **169,334 `mergeNodesSubset` calls, 181,322 `microDegrees` reads, zero stale reads** across 20k- and 100k-node graphs. Structurally, `refinePartition` passes exactly one whole macro-community per call and the read site filters on `belongings[et] === currentMacroCommunity`, so the subset is closed over every read.
- `pipeline-graph-golden` — whose committed baseline does contain `Community` nodes and `MEMBER_OF` edges — passes at head, confirming byte-identical community output through the production pipeline, not just the synthetic A/B.
- Taint pass ran on a real `--pdg` layer (239,984 nodes / 522,025 edges): zero findings for the changed file.
- 88 tests green across the community consumers plus the golden and asset-parity suites.

### Still true, still worth knowing

The native path has never executed here — the prebuilds need Arrow 24, libomp and glibc ≥ 2.38 and this host is glibc 2.36 — so the API contract is pinned against a stub mirroring the published `lib/index.js`, not the real binary. And no test gates the performance property itself, because the fix is partition-neutral by construction; that is what the fork marker is for.

Unrelated, spotted while editing: the `GITNEXUS_MEMORY` row in the worker-pool env table (added by #2679) has an extra column, so any prettier run over the README reflows that whole table. Left alone here to keep this diff to four lines.

PR #2694, opened 2026-07-25
fix(setup): preserve existing OpenCode config.jsonc

### LINKED ISSUE #2625: When modifying the OpenCode config, the config.json file was not recognized.
### Area

### Problem or opportunity

When modifying the OpenCode configuration, the config.jsonc file was not recognized. A new config.json file was created and MCP configuration was set when a config.jsonc file already existed.

### Proposed solution

I hope it can be set directly in the current config.jsonc file.

### Alternatives considered

_No response_

### Acceptance criteria

_No response_

### Constraints

_No response_

### Contribution

- [ ] I am willing to open a PR for this (may need design discussion first).

azizur100389:
## Summary
- Fixes `gitnexus setup` for OpenCode installs that already have `~/.config/opencode/config.jsonc` by updating that file instead of creating a shadow `opencode.json`.
- Reuses the existing MCP config-chain resolver while preserving the current fresh-install fallback to `opencode.json`.
- Adds setup and uninstall regression coverage for OpenCode `config.jsonc`.

Closes #2625.

## Validation
- `cd gitnexus && npx vitest run test/unit/setup-jsonc.test.ts test/unit/setup-selection.test.ts test/unit/uninstall.test.ts --testTimeout=120000` - 67 passed
- `cd gitnexus && npx tsc --noEmit`
- `cd gitnexus && npx prettier --check --ignore-unknown src/cli/editor-targets.ts src/cli/setup.ts test/unit/setup-jsonc.test.ts test/unit/setup-selection.test.ts test/unit/uninstall.test.ts`
- `cd gitnexus && npx eslint src/cli/editor-targets.ts src/cli/setup.ts test/unit/setup-jsonc.test.ts test/unit/setup-selection.test.ts test/unit/uninstall.test.ts` - warnings only, pre-existing `no-explicit-any` warnings in touched test/setup files
- `cd gitnexus && npm install` / build completed after refreshing local dependencies; local Node emitted engine warnings (`22.17.1` vs required `^22.18.0 || >=24.11.0`)
- `cd gitnexus && npm run test:unit -- --testTimeout=120000` was attempted locally; unrelated Windows/environment failures remain outside this diff (for example missing `grep`, git/mock expectations, and timeout-sensitive tests)

## Review
- Pre-PR Tri review found no merge-blocking branch hygiene, runtime, test/CI, or security issues.
- Addressed the review's non-blocking coverage gap by adding OpenCode `config.jsonc` uninstall coverage.

### DIFF
--- a/gitnexus/src/cli/editor-targets.ts
+++ b/gitnexus/src/cli/editor-targets.ts
@@ -122,6 +122,7 @@ export function getEditorTargets(home: string = os.homedir()): EditorTargets {
       id: 'opencode',
       label: 'OpenCode',
       file: path.join(home, '.config', 'opencode', 'opencode.json'),
+      legacyFiles: [path.join(home, '.config', 'opencode', 'config.jsonc')],
       // OpenCode nests servers under `mcp`, not `mcpServers`.
       keyPath: ['mcp', 'gitnexus'],
     },
--- a/gitnexus/src/cli/editor-targets.ts
+++ b/gitnexus/src/cli/editor-targets.ts
@@ -122,7 +122,12 @@ export function getEditorTargets(home: string = os.homedir()): EditorTargets {
       id: 'opencode',
       label: 'OpenCode',
       file: path.join(home, '.config', 'opencode', 'opencode.json'),
-      legacyFiles: [path.join(home, '.config', 'opencode', 'config.jsonc')],
+      // OpenCode merges config.json -> opencode.json -> opencode.jsonc; setup
+      // writes an existing readable config to avoid creating a shadow file.
+      legacyFiles: [
+        path.join(home, '.config', 'opencode', 'opencode.jsonc'),
+        path.join(home, '.config', 'opencode', 'config.json'),
+      ],
       // OpenCode nests servers under `mcp`, not `mcpServers`.
       keyPath: ['mcp', 'gitnexus'],
     },
--- a/gitnexus/test/unit/setup-jsonc.test.ts
+++ b/gitnexus/test/unit/setup-jsonc.test.ts
@@ -40,6 +40,8 @@ describe('setupOpenCode — JSONC preservation', () => {
 
   const opencodeDir = () => path.join(tempHome, '.config', 'opencode');
   const opencodeJsonPath = () => path.join(opencodeDir(), 'opencode.json');
+  const opencodeJsoncPath = () => path.join(opencodeDir(), 'opencode.jsonc');
+  const configJsonPath = () => path.join(opencodeDir(), 'config.json');
   const configJsoncPath = () => path.join(opencodeDir(), 'config.jsonc');
 
   beforeEach(async () => {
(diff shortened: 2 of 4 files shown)


gitnexus/src/cli/editor-targets.ts
```diff
@@ -122,6 +122,7 @@ export function getEditorTargets(home: string = os.homedir()): EditorTargets {
       id: 'opencode',
       label: 'OpenCode',
       file: path.join(home, '.config', 'opencode', 'opencode.json'),
+      legacyFiles: [path.join(home, '.config', 'opencode', 'config.jsonc')],
```
magyargergo: This is the blocker. `config.jsonc` isn't in opencode's candidate list, it only reads `opencode.jsonc`, `opencode.json`, `config.json` (config.ts:140, and the merge at :258-260). Writing here means the entry lands in a file the editor ignores while setup still reports success.

```suggestion
      legacyFiles: [
        path.join(home, '.config', 'opencode', 'opencode.jsonc'),
        path.join(home, '.config', 'opencode', 'config.json'),
      ],
```

That covers both shapes that actually reproduce the issue. Note the merge order (config.json < opencode.json < opencode.jsonc, last wins), which is worth a line of comment here since it's different from the CodeBuddy chain right below.

gitnexus/test/unit/setup-jsonc.test.ts
```diff
@@ -152,6 +153,28 @@ describe('setupOpenCode — JSONC preservation', () => {
...
+  "mcp": { "other": { "command": "keep" } }
+}`;
+    await fs.writeFile(configJsoncPath(), jsonc, 'utf-8');
+
+    const { setupCommand } = await import('../../src/cli/setup.js');
+    await setupCommand();
+
+    const raw = await fs.readFile(configJsoncPath(), 'utf-8');
+    expect(raw).toContain('OpenCode user config');
+    await expect(fs.access(opencodeJsonPath())).rejects.toThrow();
```
magyargergo, line 172: This assertion locks in the behaviour I flagged: it requires that no opencode-readable file exists after setup. Once the filenames are corrected I'd flip this around, assert the entry lands in `opencode.jsonc` / `config.json` when those exist, and keep a case where a lone `config.jsonc` still gets a working `opencode.json` written alongside it.

magyargergo (changes requested):
Thanks for picking this up, and the shape of the fix is right (reusing `resolveMcpConfigFile` instead of adding a second resolver, letting the uninstall sweep come for free off `legacyFiles`). But I don't think `config.jsonc` is a file OpenCode ever reads, so I don't think this fixes #2625.

I went and looked at the opencode source before writing this up. `packages/opencode/src/config/config.ts:140`:

```ts
const candidates = ["opencode.jsonc", "opencode.json", "config.json"].map((file) =>
  path.join(Global.Path.config, file),
```

and the loader at `:258-260` merges `config.json`, then `opencode.json`, then `opencode.jsonc`. Same three names at v0.1.140, v0.3.0, v0.9.0, v1.0.0 and on dev, so it isn't a recent rename either. Code search for `"config.jsonc"` across the repo returns nothing (`"opencode.jsonc"` returns 123 hits, so the search is working).

The practical effect: someone whose only config is `~/.config/opencode/config.jsonc` runs `gitnexus setup`, we write the entry into that file, print `OpenCode ✓`, and opencode never loads it. Today that same person ends up with a stray `opencode.json`, which is the cosmetic thing the issue complains about, but it at least works. So this trades a cosmetic problem for a silent one.

I checked it against this branch with a throwaway test (temp HOME, run `setupCommand()`, list `~/.config/opencode` afterwards):

- pre-existing `config.json` → ends up with `config.json` + `opencode.json`, entry in `opencode.json`
- pre-existing `opencode.jsonc` → ends up with `opencode.jsonc` + `opencode.json`, entry in `opencode.json`
- pre-existing `config.jsonc` → only `config.jsonc`, entry goes in there

The first two rows are the real shapes of #2625 and they're both still creating the shadow file. The third is the new regression.

So I think the one-line change is just pointing at the right names:

```ts
legacyFiles: [
  path.join(home, '.config', 'opencode', 'opencode.jsonc'),
  path.join(home, '.config', 'opencode', 'config.json'),
],
```

One other thing worth a comment while you're in there: opencode *merges* all three global files (config.json, then opencode.json, then opencode.jsonc, last one wins) rather than reading only the first existing one. The doc comment on `resolveMcpConfigFile` justifies first-existing-wins with CodeBuddy's "reads only the first file in the chain" behaviour, which doesn't hold for opencode. Doesn't change what to do here, but the next person reading it will assume it does.

Unrelated to your diff, but adjacent: both READMEs tell people to add the MCP block to `~/.config/opencode/config.json` (README.md:290, gitnexus/README.md:108) while setup writes `opencode.json`. Pre-existing drift, your call whether to sweep it up in the same PR.

Uninstall side looks fine as-is, the sweep in uninstall.ts:457 walks the chain generically so it picks up whatever we settle on, and the new uninstall test is a good add.

gitnexus/src/cli/editor-targets.ts
azizur100389: Fixed in f2a400fa. OpenCode now keeps `opencode.json` as the fresh fallback, but the existing-file candidates are the readable files OpenCode actually loads: `opencode.jsonc` and `config.json`. I also added the merge-order note on the target metadata so the difference from CodeBuddy is visible at the call site.

gitnexus/test/unit/setup-jsonc.test.ts
azizur100389: Updated in f2a400fa. The setup tests now cover existing `opencode.jsonc` and `config.json` without creating `opencode.json`, and the lone `config.jsonc` case now asserts setup creates a working `opencode.json` while leaving the ignored file untouched. Uninstall coverage was adjusted to sweep `opencode.jsonc`/`config.json` and leave `config.jsonc` alone.

magyargergo (approved):

Issue #1757, opened 2026-05-21
LANG-kotlin: Add lambda scopes and lambda-parameter bindings


magyargergo:
Parent tracking issue: #1746

## Summary

Open lexical scopes for Kotlin lambda literals and bind parameters (`it`, explicit params, destructuring where supported).

## Acceptance criteria

- Calls inside lambda bodies resolve through lambda parameters where applicable.
- Implicit `it` is scoped only to the lambda body.
- Nested lambda shadowing is handled deterministically.

## References

- Common targets: `forEach`, `map`, `filter`, `let`, `run`, `apply`

magyargergo:
Closed in PR #1782 via the `kotlin-lambda-scopes` fixture + lambda parameter / implicit-`it` binding work in `kotlin/query.ts`, `captures.ts`, and `simple-hooks.ts`. Lambda dispatch (`forEach`, `map`, `filter`, `let`, `apply`, etc.) now resolves through the lambda's binding scope rather than falling through to free-call resolution. Closing when #1782 merges.

PR #2695, opened 2026-07-25
fix(scope-resolution): resolve calls through a closure-valued binding across languages (#2693)

### LINKED ISSUE #2693: Feed callable-value-flow from every language: calls through a callable-valued binding resolve only in C
## Summary

`emitCallableValueFlow` is a language-agnostic scope-resolution pass that resolves calls made through a name that *holds* a callable (`const f = <callable>; f()`). It is wired unconditionally in the pipeline — gated only on `calleeIdAccumulator`, not on language (`src/core/ingestion/scope-resolution/pipeline/run.ts:893`).

But only **one** language emits the `callable-flow` captures it consumes: `src/core/ingestion/languages/c/query.ts`. Every other language starves the pass, so calls through a callable-valued binding do not resolve.

## Why this matters

`impact` on such a symbol under-reports to **zero** — the same false-all-clear failure mode as #2687, where an agent following "check impact before editing" gets a confident wrong answer.

#2687 fixed the graph *label* (a closure binding now emits a `Function` node in every language) and additionally made calls resolve in **Go, Python and C++** by adding a `@declaration.function` capture anchored on the inner closure literal.

That route is unavailable for **Kotlin, Swift and Dart**, and deliberately so — see below.

## Why the cheap route does not generalise

A `@declaration.function` anchor only works when its range aligns with a `@scope.function` for the same node; otherwise `pass2AttachDeclarations` resolves `innermost` to the enclosing scope and every call inside the closure body loses caller attribution (documented in `languages/typescript/query.ts:165-175`).

- **Kotlin** scopes `lambda_literal` as `@scope.block` *on purpose* (#1757, smart-cast semantics).
- **Swift** and **Dart** have no closure scope at all.

Flipping Kotlin to `@scope.function` would buy call resolution by breaking smart-cast scoping. That trade should not be taken.

The deeper cause is that a scope carries a single `kind` doing two independent jobs: name-resolution/shadowing behaviour, and "is a callable boundary" (the attribution anchor a declaration can be owned by). For a lambda both are wanted, and the model forces a choice.

## The declaration route is also structurally weaker

Making the *binding* the callable is the syntactic-tagger answer. It cannot express:

- reassignment — `var f = a; f = b;`
- conditional initialisers — `const f = cond ? g : h;`
- a callable passed through an intermediate

`callable-value-flow` can: it is a fixpoint with `resolvedInvokes` / `ambiguousInvokes` / `unmatchedInvokes` and an iteration count. It is the general mechanism; the declaration capture is the fast path.

## Proposed work

1. Document the `callable-flow` capture contract in `src/core/ingestion/utils/callable-flow-captures.ts` (the cell/site model, `@callable-flow.expected-const`, etc.) — today `c/query.ts` is the only worked example.
2. Emit those captures for Kotlin (`val f = { }`), Swift (`let f = { }`) and Dart (`var f = () => x`). Per-language query additions only: **no shared-code change and no scope-kind change**, so Kotlin smart casts are untouched by construction.
3. Verify calls resolve for each, and that the Kotlin smart-cast suites still pass.
4. Evaluate moving Go/Python/C++ onto the same mechanism for the reassignment cases the declaration route cannot express.
5. Extend to the remaining languages that have a callable-value idiom.

## Optional follow-up: decouple the conflated axis

If step 2 proves insufficient, the principled fix is an orthogonal scope attribute — "callable boundary" independent of `kind` — so a scope can be a shadowing block *and* an attribution anchor. That permanently removes the forced choice and would let the declaration route work for Kotlin without touching smart casts. Cross-language and larger; worth its own issue if reached.

## Acceptance

- `f()` resolves to the closure's `Function` node in Kotlin, Swift and Dart
- Kotlin smart-cast behaviour unchanged (`lambda_literal` stays `@scope.block`)
- A regression test per language asserting the resolved `CALLS` target, mirroring `test/integration/closure-binding-labels.test.ts`

Follow-up to #2687 / PR #2691.

### LINKED ISSUE #2699: Function-local symbols share file-scoped node ids, so a local collides with a same-named file-level symbol (SCIP uses `local N`)
## Summary

Graph node ids for function-local symbols are **file-scoped**, so a local binding and a file-level symbol with the same name collapse onto one node. SCIP solves this with a dedicated `local <id>` symbol form; GitNexus has no equivalent.

## Reproduction

```php
<?php
function save($x) { return $x; }

function run() {
  $save = fn($x) => $x * 2;   // local closure, same name
  return $save(1);
}
```

Both the top-level function `save` and the local closure `$save` want the id `Function:x.php:save`, so only one node exists and `$save(1)` gets **no CALLS edge at all**.

Without the collision it resolves fine:

```php
function run() {
  $h = fn($x) => $x * 2;
  return $h(1);            // -> Function:x.php:h
}
```

## Not PHP-specific

This is the repo-wide file-scoped node-identity convention. TypeScript, Python and Dart collapse identically on the same shape (verified while reviewing #2695): a top-level `f` and a function-local `f` share one node, and the local call is attributed to the top-level symbol.

PHP is the sharpest case only because PHP keeps variables and functions in **separate namespaces** — `$save` and `save()` cannot collide in the language at all, so the collapse is purely an artifact of dropping the `$` sigil and the enclosing scope.

Java and C# escape it by accident: their locals are qualified by the enclosing class (`Function:J.java:J.handler`), not by the enclosing function — so the same collision reappears for two locals of the same name in different methods of one class.

## Prior art

**SCIP** ([spec](https://github.com/sourcegraph/scip/blob/main/docs/scip.md)) separates the two keyspaces outright:

- global symbols: `<scheme> ' ' <package> ' ' (<descriptor>)+`
- local symbols: `'local ' <local-id>` — *"local symbols MUST only be used for entities which are local to a Document, and cannot be accessed from outside the Document"*

Locals are enumerated per document and never globally addressable, so a local can never collide with a module-level symbol. `SymbolInformation.enclosing_symbol` links the local back to its parent scope.

**Compilers** agree: in tsc, PHP, Python and Dart a local binding is a distinct symbol from a same-named module-level one — scope is part of identity.

## Why it matters

It is a silent wrong answer in both directions:

- the local call gets **no** edge (the case above), or
- worse in principle, an edge attributed to the wrong symbol — `impact` on the top-level function then reports a caller that never calls it.

## Proposed work

1. Give function-local definitions a scope-qualified identity — the enclosing callable's name in the id (`Function:x.php:run.save`), mirroring how class members already qualify (`Class.member`), or an explicit local-enumeration keyspace closer to SCIP's `local N`.
2. Decide the contract for cross-file reference: locals are by definition unreachable from other files, so they can be document-scoped like SCIP's.
3. Bump `INCREMENTAL_SCHEMA_VERSION` and the parse-cache `SCHEMA_BUMP` — this changes persisted node ids.
4. Audit consumers that assume file-scoped ids: `resolveDefGraphId`'s `simpleKey` fallback, `detect_changes`, saved queries, and anything storing node ids externally.

## Scope note

Deliberately **not** fixed in #2695. That PR makes closure bindings resolvable across languages; changing node identity for every function-local symbol is a cross-language change with persisted-id churn and its own invalidation, and it would have buried the fix it rode in on. #2695 fails safe here — the colliding case emits no edge rather than a wrong one.

Follow-up to #2693 / #2695.

### LINKED ISSUE #2701: `this` resolution ignores the arrow/function boundary, minting false CALLS edges (tsc: getThisContainer includeArrowFunctions=false)
## Summary

`this` resolution treats a **function expression** as transparent, so a `this.member()` call inside one resolves to the enclosing class. In JavaScript/TypeScript it must not: an ordinary function is a `this`-boundary, and only an **arrow** captures `this` lexically. The result is false `CALLS` edges — including the exact `forEach(function () { this.m(); })` bug arrow functions were introduced to avoid.

## Reproduction

Verified against a built `dist` (each pair differs only in arrow vs `function`):

```ts
export class C {
  m(): void {}
  build() {
    const h = () => { this.m(); };      // this IS C
    return h;
  }
}
// CALLS: Function:C.ts:C.h -> Method:C.ts:C.m#0        CORRECT
```

```ts
export class D {
  m(): void {}
  build() {
    const h = function () { this.m(); };  // this is NOT D
    return h;
  }
}
// CALLS: Function:D.ts:D.h -> Method:D.ts:D.m#0        FALSE EDGE
```

```ts
export class F {
  m(): void {}
  run(xs: number[]) {
    xs.forEach(function () { this.m(); });   // this is NOT F
  }
}
// CALLS: Method:F.ts:F.run#1 -> Method:F.ts:F.m#0      FALSE EDGE
```

The arrow variants of both are correct and must keep resolving.

## How the reference implementations model it

**TypeScript compiler** — `getThisContainer(node, includeArrowFunctions)`. When resolving `this`, the checker calls it with `includeArrowFunctions = false`, so arrow functions are skipped and the walk stops at the first ordinary function, class, or module. That single flag is the whole distinction.

**V8 / ECMA-262** — `OrdinaryFunctionCreate` sets `[[ThisMode]]`: `lexical` for arrows, `strict`/`global` otherwise. A function with `[[ThisMode]] = lexical` has no `this` binding in its environment record at all, so the lookup passes through to the enclosing environment. `bind`/`call`/`apply` cannot change it — the spec's `[[Call]]` for arrows ignores the supplied `thisArgument`.

**SCIP** does not model `this` itself; `scip-typescript` inherits the behaviour from the tsc checker, so its occurrence for `this.m()` in the arrow case points at `C#m().` and in the function-expression case points at nothing.

The two references agree on one rule, and it is small: **an arrow is transparent for `this`; every other function is a boundary.**

## Proposed model in GitNexus

Mirror it as a scope attribute rather than a new scope kind — the kind already carries name-resolution/shadowing semantics and must not be overloaded again (the same conflation that blocks closure caller-attribution, #2699):

1. Add an optional `thisTransparent` boolean to the function-scope record. Default `false`, so every existing language is unchanged until it opts in.
2. Language queries mark arrow scopes: in TS/JS, `(arrow_function) @scope.function.this-transparent` (or an equivalent capture-side flag). Ordinary `function_expression` / `function_declaration` / `method_definition` stay opaque. Other languages follow their own rules — Kotlin lambdas are `this`-transparent for the outer receiver, PHP `fn()` arrow functions capture by value and `function()` does not.
3. `this`-receiver resolution walks the scope chain from the call site, passing through `thisTransparent` function scopes and stopping at the first opaque function scope (→ unresolved) or class scope (→ resolve the member). This is `getThisContainer(..., includeArrowFunctions=false)` expressed on the scope tree.

Note the walk is not a single site: `receiver-bound-calls.ts:477` handles it for languages setting `resolveThisViaEnclosingClass` (C++ only today), and `compound-receiver.ts:404` handles the `this`-headed chain. Both need the boundary check, and any per-language receiver hook that resolves `this` needs auditing.

## Scope and risk

This touches the receiver-resolution path used by every member call, so it wants its own change with focused tests, not a rider on a feature PR. The fix REMOVES edges, so it should be measured against a real corpus before landing — some existing tests may encode the current (wrong) behaviour, exactly as `const-function-twin.test.ts` encoded the `var`-label gap until #2695 corrected it.

## Acceptance

- `this.m()` inside a nested `function` expression emits no edge to the enclosing class, in both the local-binding and callback forms above
- the arrow variants continue to resolve
- class-field arrows keep resolving (they are lexically bound to the instance)
- `bind`/`call`/`apply` remain unresolved rather than newly wrong — modelling receiver rebinding is a separate concern

Found while reviewing #2695. Not caused by it: the behaviour predates that PR, which only changed node LABELS. Related: #2699 (closure-binding node model — caller attribution and local identity).

magyargergo:
Fixes #2693, #2699 and #2701. Follow-up to #2687 / #2691.

> **Scope note.** This started as #2693 alone. #2699 (function-local node identity) and #2701 (`this` across the arrow/function boundary) were originally filed here as *deferred* limitations; both are now implemented in this PR — see the two sections near the end. The sections before those describe the original #2693 work and are unchanged.

`val f = { … }; f()` emitted **no `CALLS` edge** in Kotlin, Swift or Dart, so `impact` on such a symbol under-reported to zero — the same false all-clear failure mode as #2687. It turned out Ruby, Java, C# and PHP had the same gap, and JS/TS `var` a related one; all are fixed here.

## The issue's premise was wrong

#2693 claimed only `languages/c/query.ts` emits the `callable-flow` captures the pass consumes, so every other language starves it. That is not what happens — `synthesizeCallableFlowCaptures` is called from **15** language capture modules, Kotlin/Swift/Dart among them. (The original grep matched the literal string `callable-flow` in `query.ts` files, which is the wrong layer: captures are synthesized programmatically in `<lang>/captures.ts`, not written as query text.)

Kotlin already resolves reassignment *through* the pass:

```kotlin
var f = ::a
if (cond) { f = ::b }
f(1)          // resolves to BOTH a and b — only the fixpoint can do that
```

So the pass was live and fed. The real causes were narrower, and there were two of them.

## RC1 — Kotlin and Swift: one gate

Their captures are already exactly right; the seed names the binding as its own callable, per the anonymous-callable convention in `callable-flow-captures.ts`. They died one layer later, in `buildGraphTargetIndex`:

```ts
if (!isCallable(def) && providerTarget?.(def) !== true) continue;
```

`isCallable` is Function/Method/Constructor, but scope resolution declares a closure binding with its **value** label (`Property`), and `isCallableValueTarget` is implemented by exactly one provider — COBOL. So the binding never entered `graphTargets`; `lexicalCallableLookup` then returned `shadowed: true` with no targets, which *also* suppressed the workspace-wide fallback, and the seed resolved to nothing.

Only the graph knows a value binding holds a callable — since #2687 it emits a single `Function` node for one. So value bindings now resolve their graph id first and are admitted on the label of the node they actually reach.

**The join is positional.** A closure binding IS its callable node — same file, same line, same name — so admission matches on `file/line/name` against an index derived in the graph walk `byAnchor` already performs. Value bindings never run `resolveDefGraphId` at all.

That distinction matters. An earlier cut of this PR admitted a value binding whose *resolved* node happened to be callable, which let the key chain fall through to the label-agnostic, first-write-wins `simpleKey(filePath, simpleName)` and alias a binding onto **any** same-named callable in the file. Review caught it minting fabricated callers — `const save = …` beside an unrelated `Svc.save`, a shadowing local, a Rust `let` — order-dependently, and in Dart it *inverted* the feature (the class method got the edge; the closure's own node got none). The positional join removes that failure mode by construction rather than by argument.

## RC2 — Dart: structural, and independent

Dart needed more, for two different reasons:

- **Top-level `var f = (x) => x;`** — a graph `Function` node already existed (#2687), but no `@declaration.*` matched the binding, so there was no `SymbolDefinition` to attach a seed to. Adding the declaration exposed a second problem: Dart's `initialized_identifier` is **fieldless**, so the shared field-based assignment fallback (`left`/`name`/`value`/…) decomposed nothing and it still emitted no flow captures at all. Kotlin's fieldless `assignment` node hit exactly this in #2522 and took the same remedy — a provider `extractAssignment`.
- **Function-local `void m() { var f = (x) => x; }`** — locals parse as `initialized_variable_definition`, which the top-level graph-node rules are deliberately anchored under `(program)` to avoid, so a local closure had **no graph node at all** — nothing for the widened gate to admit.

Both new rules are restricted to a `function_expression` value. Declaring every Dart variable would mint defs and nodes repo-wide for no resolution benefit; ordinary locals stay unindexed exactly as before.

## No scope kind changed

Kotlin's `lambda_literal` is still `@scope.block` (`kotlin/query.ts:81`) — #1757 smart-cast semantics are untouched **by construction**, not by argument. The diff contains no `@scope.` change at all. The fix is also language-neutral: it discriminates on the graph node label, never on a language name (AGENTS.md).

I did **not** take the `isCallableValueTarget` hook the issue implies. It only receives the `SymbolDefinition`, never the graph node, so the best predicate it could express is `def.type === 'Property'` — which would admit every Kotlin/Swift property as a callable target and mint false `CALLS` edges.

## Performance

Widening the gate puts the hot loop on a much larger def population — value bindings outnumber callables in real source. Measured, then fixed:

|                     | name-match (rejected) | positional (shipped) |
| ------------------- | --------------------- | -------------------- |
| `large_ms`          | 7.79–8.37             | **4.37–4.63** |
| `widening_overhead` | 2.50–2.82             | **1.43–1.58** |
| target fingerprint  | `70bebf6a…`           | `70bebf6a…` **identical** |

Value bindings are matched by position against an index derived in the existing graph walk, so they never run the `resolveDefGraphId` key chain — correct *and* cheaper than the name-match design review rejected. `definitionAnchorKey`, a regex per def, is likewise no longer run for value defs, which can never match a label-keyed anchor.

New `bench/callable-value-flow/` guards the fingerprint, scaling ratio and overhead, wired into `ci-tests.yml` beside the other build-free benches. The overhead budget (1.9) sits between the positional and name-match bands, so it cannot be met by reverting to the slower, incorrect design. Timings use the min of 15 warmed reps, not the median — the same build reported 1.65 idle and 2.03 under load, and a median gate would have to be loosened past the point of detecting the regression it exists to catch.

## Cache invalidation

`INCREMENTAL_SCHEMA_VERSION` 15 → **19** and parse-cache `SCHEMA_BUMP` 22 → **26**, across the three issues this PR now covers. This emits **new** `CALLS` edges, new Dart `Function` nodes, and re-keyed function-local callables, and the incremental write set only covers changed files — without the bumps an existing index would keep reporting a zero blast radius for exactly the symbols this fixes.

The parse-cache bump is load-bearing beyond the index: a warm parse cache replays worker results verbatim and `--force` does **not** clear it. `Scope.ownsReceivers` (#2701) lives on the cached `Scope`, so without the `SCHEMA_BUMP` the fix would silently not take effect on an existing checkout.

One bump covers the whole PR: v19 is unreleased — introduced on this branch and absent from `main` — so the later commits need no further bump.

## Tests

Per-language closure-call resolution (Kotlin, Swift, Dart top-level + function-local), each confirmed **failing at HEAD first**. Plus:

- a **negative** pair — a constant sharing its name with a function mints no `CALLS`; a property initialised from a call is not itself callable;
- a **double-emit** guard for Go/Python/C++/TS, which already resolved these calls via the declaration route and now have a second possible route. It asserts on edge **ids**, not target ids: `tryEmitEdge` dedups by key, and a collapsed key and a site-anchored key are different keys, so a real double-emit shows up as two ids rather than being silently collapsed.

Five **collision** regression tests — a value binding paired with a same-named callable in TypeScript (alias, class method, shadowing), Rust `let`, and Dart — each confirmed failing against the pre-fix commit. Plus Kotlin class-body closures (the only case exercising the `Method` arm), Dart top-level `final`, Dart multi-name locals, and a warm-parse-cache replay for Kotlin and Dart, since these captures are replayed verbatim and a serialization change would otherwise surface only on a *second* analyze.

An earlier negative test was vacuous: it paired names that did not collide (`maxSize` vs `size`), so the pre-filter rejected the input before the guard it was named after could run — deleting that guard changed nothing.

Verified: closure-binding 33/33 · scope-resolution + resolver units 1389/1389 · Kotlin + Dart 444/444 · `tsc` clean · lint 0 errors · all 15 `scope-capture` fingerprints unchanged.

## Documentation

`callable-flow-captures.ts` is 1200+ lines behind a nine-line docblock whose only worked example was C. Both root causes here were "the contract was discoverable only by reading the emitter", so it now records the cell/site model, the anonymous-callable convention, and the two traps that cost a debugging cycle each (a fieldless binding node; captures without a declaration).

## Also worth noting

The issue's "optional follow-up" — decoupling *callable boundary* from scope `kind` — is **not needed**. That redesign existed to avoid the forced choice between smart-cast scoping and call attribution; this fix resolves the calls without touching scope kinds at all. That section of #2693 can be dropped rather than carried.

## Review round

`/gitnexus-review` plus four expert lenses (correctness, adversarial, testing, production-risk) ran against `b2b2c16f`. Everything they raised is addressed here:

- **False `CALLS` edges** (the P0 above) — fixed by the positional join; five collision regression tests added, each confirmed failing first.
- **`call-summary-schema-version.test.ts` was RED** — the 15 → 16 bump made `passesReuseGate(15)` false. Now pins 16 as current and 15 as rejected, matching every prior bump.
- **Vacuous negative test** — replaced with real collisions.
- **Kotlin class-body closures** worked but were untested; they are the only case exercising the `Method` arm.
- **Bench source contained a literal NUL byte**, so git recorded it as binary and the artifact pinning the target set was unreviewable in the diff. Also its corpus modelled `startLine` 1-based where graph nodes are 0-based, which would have stopped it exercising the value-binding path.
- **`Static`** dropped — `normalizeNodeLabel` has no `static` case, so no def can carry it.
- **Dart gaps** — top-level `final`/`const` closures and every declarator after the first in a multi-name local now resolve.
- **Warm-cache replay** for Kotlin and Dart added.
- **`VALUE_BINDING_DEF_TYPES`** now documents why it deliberately does not reuse `isOwnableValueLabel`.

## Extended to the remaining languages

Ruby, Java, C# and PHP already emitted correct `callable-flow` seeds and invokes — what they lacked was the same #2687 piece: a **callable graph node** at the binding, which is what the positional join matches. PHP additionally had no scope declaration for the bound name.

```
ruby    handler = ->(x) { x }        handler.call(1)   -> Function:a.rb:handler
java    Function<..> handler = x->x  handler.apply(1)  -> Function:A.java:A.handler
csharp  Func<int,int> handler = ...  handler(1)        -> Function:A.cs:A.handler
php     $handler = fn($x) => $x      $handler(1)       -> Function:a.php:handler
```

Ruby and Java invoke through the callable-object protocol, C# and PHP directly. Locals work in all four, and a binding colliding with a same-named method resolves to the **closure**, not the method.

Two things the sweep caught:

- **Java twin** — anchoring on the inner `variable_declarator` produced *both* a `Function` and a `Property` node, the exact #2687 double-indexing. The dedup keys on (definition node, name) and Java's value rule anchors on `field_declaration`; re-anchored to match.
- **JS/TS `var`** — `var f = (x) => x` kept a `Variable` label while `const`/`let` got `Function`, so the `CALLS` edge that resolved through the declaration route pointed at a **non-callable node**. `var` is a different grammar node no closure rule covered. This flipped an assertion in `const-function-twin.test.ts` whose own comment documented the gap ("`var` has no matching `@definition.function` pattern") rather than defending it; the property it protected — an *unclaimed* value node survives — now has its own non-function case.

## #2701 — `this` is no longer resolved across the arrow/function boundary

Originally deferred on this PR; now implemented.

`this.m()` inside a nested **function expression** resolved to the enclosing class, minting false edges — including the classic `xs.forEach(function () { this.m(); })` that arrows exist to avoid. Only an arrow captures `this` lexically. Both references agree on one small rule: ECMA-262 gives an arrow `[[ThisMode]] = lexical`, and `tsc` resolves `this` through `getThisContainer(node, includeArrowFunctions = false)`.

Modelled as a **query capture**, not a new scope kind — `kind` already carries shadowing semantics, and overloading it is the same conflation that blocks caller attribution below. `@receiver-owner.this` rides on the same node as `@scope.function` and is read by a one-line provider hook into a new `Scope.ownsReceivers`.

The marker must live **outside** the `@scope.` namespace. `anchorCaptureFor` picks the broadest-span `@scope.`-prefixed capture, and two captures on one node have equal spans — so a `@scope.owns-*` name would non-deterministically become the anchor and the scope would vanish. (`Capture` carries only `{name, range, text}`, no node type, so a hook needing the grammar node type has to get it from a query-level marker.)

Three layers had to close, found by probing after each rather than by reading:

1. `Scope.ownsReceivers` stops **both** receiver-type walks — `findReceiverTypeBinding` in ingestion and its twin `lookupReceiverType` in `gitnexus-shared`'s `lookup-core.ts`. Two walks with the same job in different packages.
2. `LanguageTypeConfig.thisBoundaryNodeTypes` stops the type-env AST walk.
3. `isReceiverOwnedButUnbound` makes `receiver-bound-calls` **suppress** the site. Without it the member still resolved by NAME through `lookupCore`'s lexical chain — the class-body scope binds `m` two scopes up.

**A claim in the original commit message was wrong and is corrected in `d21a6673`.** It said all three layers were independently load-bearing because the false edge survived removing any one. That was measured during development and not re-tested against the shipped code. Layer 3 runs first and **subsumes** layer 1 for an explicit `this` receiver: removing layer 1's gate leaves all 10 tests in `this-boundary.test.ts` passing. The gate is nonetheless retained — `receiver-bound-calls` only suppresses *explicit* receivers, while layer 1 is also reached for implicit ones via `IMPLICIT_RECEIVERS` in `resolveReceiverOwner`. The experiment shows that gate is **untested, not unreachable**; it needs an implicit-receiver test before its removability is considered demonstrated.

A follow-up fixed a second boundary: an object-literal method nested in a class bound `this` to the **class**.

```ts
class A { outer() { const o = { inner() { return this.x; } }; return o; } }
```

`this` in `o.inner` is `o`, never `A`. `findEnclosingType` now stops at `THIS_REBINDING_BOUNDARY_TYPES`; an arrow is deliberately absent from that set, which is what keeps a class-field arrow `m = () => this.x` resolving.

**Trade accepted, pinned as tests asserting empty.** `.bind(this)`, `.call(this)` and `forEach(fn, thisArg)` DO make `this` the instance at runtime; those edges were correct and are now dropped. Correctness there is fixed at the call site, which a scope-level rule cannot see — so the choice is between losing them and keeping every detached-callback false positive. `this` in a `static` method also no longer reaches the instance member, which was wrong in the other direction.

## #2699 — function-local identity, and JS/TS block scopes

Also originally deferred here; the **identity** half is now implemented. (The caller-attribution half remains open — see limitations.)

A function-local callable now carries its enclosing-callable chain **and its own source position**: `run.save@3:2`. Position rather than name alone, because names cannot express ECMAScript scoping —

```js
function outer(a) {
  if (a) { const pick = …; return pick(1); }   // one binding
  else   { const pick = …; return pick(2); }   // a DIFFERENT binding
}
```

— and both are `outer.pick` by name. A declaration's own position is unique across every environment record it could belong to, with no conditionals and no block enumeration. Applied **only to locals**; top-level functions and class methods keep their existing ids byte-for-byte, which bounds the id churn.

JS/TS also gained real block scopes (`(statement_block) @scope.block`) so sibling-block `let`/`const` are distinct bindings. Emitted naively that tripled the block-scope count and cost ~10% of analyze wall time, since every scope-chain walk then steps through levels binding nothing. Two emit-side filters keep the semantics: a block that IS a function body duplicates the enclosing Function scope, and a block declaring no `let`/`const`/`class`/`function` is transparent to every lookup. **19,389 → 5,331** block scopes on a 762-file corpus, overhead ~+10% → **~+2%**. Guarded by a new `bench/scope-emission/` gate on exact scope counts — a correctness gate, not a timing one, because a 2% effect sits well inside the ±10% run-to-run noise of a shared runner.

Two defects found in review, both fixed:

- **The positional join failed OPEN.** On a position miss, `resolveDefGraphId` fell through to the label-agnostic, first-write-wins `simpleKey`. A multiline `const fn =` reproduces it: the graph node anchors on the outer `lexical_declaration` while the scope def anchors on the inner `arrow_function`, so the rows differ and a **false** edge was minted. It now fails closed — a `localNameKey` records that a function-local of that simple name exists, and the resolver returns `undefined` rather than guessing.
- **Java anonymous classes lost their identity.** `CLASS_CONTAINER_TYPES` lists class *declarations* only, but `object_creation_expression > class_body` has none, so the walk ran past it into the enclosing method and re-keyed `Worker$1.run` as `Worker.makeHandler.run@7:12`. Fixed with a dedicated `CALLABLE_PREFIX_BOUNDARY_TYPES` covering class/interface/enum/annotation bodies and object-creation expressions.

## Tri-review round

Two engines (Claude, and Codex `gpt-5.6-sol` at `xhigh`) each ran the GitNexus PR swarm **and** the ce-code-review persona pipeline, plus a Codex-only adversarial pass; merged and critic-gated, posted as review `4782134453`. Outcomes:

| Finding | Outcome |
| --- | --- |
| **P0** Java anonymous-class identity destroyed | Fixed — `CALLABLE_PREFIX_BOUNDARY_TYPES` |
| **P1** positional join fails open → fabricated edge | Fixed — fail closed via `localNameKey` |
| **P1** `@receiver-owner.this` over-suppresses on method forms | **Rejected, with measurement** — see below |
| **P2** block scopes drop 114 `ACCESSES` edges | **Not a regression** — all 114 were false edges, see below |
| **P2** "three independently load-bearing gates" claim is false | Corrected in `d21a6673` |
| **P2** the `this`-boundary node-type lists agree only by hand | Fixed — drift test now enforces it |
| **P2** scope-emission gate is TypeScript-only | Fixed — per-language corpora and baselines |
| **P3** `generator_function` missing from `FUNCTION_NODE_TYPES` | Fixed — consistency only, no output change |
| **P3** `findEnclosingFunctionId` breaks the documented lockstep | **Open** — see limitations |

Two of these closed as **measured non-regressions**, and the measurement is recorded rather than the conclusion asserted:

**The −114 `ACCESSES` delta is false-edge removal.** All 274 emitting reference sites behind those edges were classified by re-reading source at the site: 269 are member reads, and the 5 others are classifier artifacts (the name recurs earlier on the line, as in `a.b.declLine` for `b`) that are member reads too. **None was bare-identifier-only**, and `added: 0`. Every dropped edge was a property read (`options.baseUrl`) mis-resolving to an unrelated function-local `const` of the same name. The cause is not block-specific — `lookupCore` Step 1 walks the lexical chain for *every* lookup, including explicit-receiver property reads — so block scopes do not fix it, they narrow it.

**Removing the marker from method forms was rejected on evidence.** It would recover 3 true edges at the cost of 2 false ones, and produced byte-identical edge sets across 762 real files. For a graph that feeds `impact`, a false edge is worse than a missing one: a wrong caller is acted on, a missing one is merely not found.

**`generator_function` changes no output.** Node and edge sets are byte-identical with and without the entry on `const g = function* (x) { yield x; }; g(1)`. It is list consistency, now enforced by a test rather than by vigilance across four hand-maintained lists.

## Known limitations that remain

All pre-existing, all fail *safe* (they under-report rather than emitting a wrong edge), none surviving only as a commit-message footnote.

- **Caller attribution (#2699, second half).** The new callable nodes are call *targets* but not call *sources*: a call inside a closure binding is attributed to the enclosing scope, so `impact(handler, direction:"downstream")` reports nothing. Cause is exact — `pickCallerCallableDef` (`graph-bridge/ids.ts`) walks child scopes containing the call site but gates on `child.kind === 'Function'`, and the binding's def is owned by the enclosing scope, so neither half of the link exists. Pinned by tests in this PR so the boundary is visible and fails in either direction. Kotlin/Ruby fail the kind gate, PHP fails only the ownership half, and Dart has no scope over a closure literal at all — so a fix needs per-language work, not one switch.
- **`findEnclosingFunctionId` lockstep (P3, open).** The definition phase appends the `localIdentity` suffix; the caller-attribution phase does not, so the two derive different ids for a nested callable. Latent — zero dangling edges observed, consistent with the attribution limit above — but the docblock's lockstep guarantee is not fully held.
- **A generator-expression binding is not callable.** `const g = function* () {}` emits a `Const` node rather than a `Function`, so `g()` resolves to nothing. The label comes from the definition rules; closing it is a graph-output change and is not made here.
- **Property reads still bind to function-body locals.** `options.baseUrl` can still resolve to a `const baseUrl` declared directly in the function body. Block scopes removed the nested-block case; the direct case needs `lookupCore` to stop walking the lexical chain for explicit-receiver reads.

## Verification

`tsc --noEmit` clean · lint 0 errors · scope-emission gate PASS (both languages) · closure-binding, `this`-boundary, function-local-identity, block-scope-shadowing and node-type-drift suites green.

CI ran fully green on `4c048a27` — every check passing except Vercel's `Authorization required to deploy`, which fails on every fork PR and has no PR-side fix. That included all three OS shards, the tree-sitter ABI matrix, CodeQL and the benchmarks gate, so the new per-language scope-emission baseline is confirmed on CI hardware rather than only locally.


gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
```diff
@@ -270,6 +271,28 @@ export function emitReceiverBoundCalls(
...
 
+      // ── owned-but-unbound receiver ───────────────────────────────
+      // The language declared this scope REBINDS the receiver and gave
+      // it no type — a JS/TS ordinary `function`, whose `this` comes
+      // from the call site (#2701). No enclosing type can be its type,
+      // so this is a definitive negative, not a miss: suppress the site
+      // instead of letting the receiver-blind lexical fallback in
+      // `lookupCore` match the enclosing class's member by name.
+      // No-op for every language that leaves `Scope.ownsReceivers` unset.
+      if (isReceiverOwnedButUnbound(site.inScope, receiverName, scopes)) {
```
magyargergo, line 282: **P1 · over-suppression deletes edges the base resolved correctly** `[reproduced]`

This gate treats *any* scope whose `ownsReceivers` contains the receiver but which carries no `this` type-binding as unresolvable, and suppresses the site. But `synthesizeTsReceiverBinding` deliberately returns `null` for **static members**, **object-literal methods**, and **anonymous class expressions** — so all three become "owned but unbound" and lose edges that resolved correctly before this PR:

```ts
class A { static a() { this.b(); }  static b() {} }          // static -> static: edge lost
module.exports = { run() { this.helper(); }, helper() {} };   // object-literal: edge lost
const X = class { m() { this.n(); } n() {} };                 // class expression: edge lost
```

The PR's only static test covers static→**instance**, which genuinely was wrong — so the over-suppression reads as intentional and no test catches it.

**Engines:** Codex (`codex-ce`, confidence 100) independently confirmed the static→static case; the object-literal and class-expression cases come from Claude base-vs-head edge diffs only.

**Fix:** distinguish "this scope rebinds `this` and we cannot type it" from "this scope owns `this` and the language declined to synthesize a binding". Only the former should suppress.

gitnexus/src/core/ingestion/scope-resolution/graph-bridge/ids.ts
```diff
@@ -127,6 +162,17 @@ export function resolveDefGraphId(
...
   if (def.type !== undefined) {
+    // Position key FIRST (#2699). A def and its graph node are the same
+    // construct, so they share a source line — the only evidence that
+    // separates a function-local declaration from a same-named file-level one
+    // without either side having to model the scope chain. Node ids are
+    // 0-based, def ids 1-based. An `AMBIGUOUS_POSITION` tombstone (two
+    // callables on one line) falls through to the name-based keys below.
+    const line = defStartLine(def.nodeId);
+    if (line !== undefined && isOverloadableCallable(def.type)) {
+      const posHit = nodeLookup.get(positionKey(filePath, def.type, line - 1, simpleNameOf(qn)));
```
magyargergo: **P1 · the positional join fails OPEN, minting a false edge** `[reproduced]`

On a position miss this falls through to the label-agnostic, first-write-wins `simpleKey(filePath, simpleName)` (line 259, not in this diff) — which is the exact fabricated-caller mechanism #2693 already shipped once as a P0.

It misses whenever the two phases anchor on different lines. `tree-sitter-queries.ts` anchors the graph node on the outer `lexical_declaration`; `languages/typescript/query.ts` deliberately anchors the scope def on the **inner `arrow_function`**. A multiline declaration splits them:

```ts
export function run(): number   { const pick =
    (x: number): number => x * 2; return pick(1); }
export function other(): number { const pick =
    (x: number): number => x * 3; return pick(2); }

observed CALLS:
  run   -> run.pick@1:2     correct
  other -> other.pick@6:2   correct
  other -> run.pick@1:2     <-- FALSE: other() never calls run's pick
```

Every fixture in `function-local-identity.test.ts` keeps the declaration and initializer on one line, where the anchors coincide — which is why the suite is green.

**Fix:** for a function-local definition, fail **closed** on a missing or ambiguous positional match rather than falling back to the simple-name key; or derive one shared anchor (including column) in both phases. Add multiline + same-line collision tests for TS, JS and Python.

gitnexus/src/core/ingestion/languages/typescript/query.ts
```diff
@@ -94,14 +94,26 @@ const TYPESCRIPT_SCOPE_QUERY = `
...
+;; \`[[ThisMode]] = lexical\` (no \`this\` in its environment record, so the
+;; lookup passes through), while every other form binds \`this\` at call time.
+;; \`method_definition\` is marked too and is unaffected — it also carries a
+;; synthesized \`this\` typeBinding, which the walk consults first.
+;; The marker rides on the same node as \`@scope.function\`; it is outside the
+;; \`@scope.\` namespace so \`anchorCaptureFor\` cannot mistake it for the anchor.
+(function_declaration) @scope.function @receiver-owner.this
+(generator_function_declaration) @scope.function @receiver-owner.this
+(function_signature) @scope.function @receiver-owner.this
+(method_definition) @scope.function @receiver-owner.this
```
magyargergo, line 109: **P1 (root cause of the `receiver-bound-calls.ts:282` finding)** `[code-read]`

Marking `method_definition` as owning its `this` is defended in the code comment by "a method's own synthesized `this` typeBinding is consulted first" — but that premise only holds when a binding is actually synthesized. `synthesizeTsReceiverBinding` has three `return null` paths (static, object-literal method, anonymous class expression), and in each the marker converts a correct edge into a suppressed one.

Either narrow the marker to the forms that genuinely receive a synthesized binding, or make the suppression require positive evidence that `this` is call-time-bound rather than merely absent.

magyargergo (commented):
# Multi-engine review — PR #2695 (#2693 + #2699 + #2701)

Base `d4fbc544` → head `64073d90` · 39 files, +2649/−75 · worktree `/workspace/.worktrees/pr-2695`

**Verdict: do not merge as-is.** One P0 was found and has already been fixed; one P1 edge-loss class
and one P1 fail-open resolution class remain open, both with concrete triggers verified in source.

---

## 1. Engines and lane liveness

**Engine A — Claude** (⚠️ *this PR was authored by Claude, so Engine A is reviewing its own work;
Claude-lane agreement is "consistent within one engine", not independent confirmation*):

| Lane | Live | Note |
|---|---|---|
| `ce-correctness` | ✅ | Built **both** head and base into throwaway trees and diffed node/edge sets over ~60 fixtures in 12 languages + a 500-file corpus. Strongest single artifact in the run. |
| `ce-adversarial` | ✅ | Static trace + read-only tree-sitter AST probes on the real pinned grammars. 11 suspicions constructed and refuted. |
| `ce-testing` | ✅ | Discrimination-focused; found the gate-2 redundancy lead. |
| `ce-maintainability` | ✅ | |
| `ce-standards` | ✅ | Full audit vs `AGENTS.md` / `GUARDRAILS.md` / `CONTRIBUTING.md` — no P0/P1/P2. |
| `gitnexus-risk-architect` | ⚠️ | **Stalled mid-run; resumed with a narrowed brief.** |
| `gitnexus-security-boundary` | ⚠️ | **Stalled mid-run; resumed with a narrowed brief.** |
| `gitnexus-test-ci-verifier` | ⚠️ | **Stalled mid-run; resumed with a narrowed brief.** Did execute all three bench `--check` gates. |

**Engine B — Codex** (`gpt-5.6-sol`, effort `xhigh`): `codex-swarm` ✅, `codex-ce` ✅,
`codex-adversarial` ✅. The `codex-ce` leg **could not start its own internal cross-model sub-pass**
(the restricted runtime denied host-CLI detection); that coverage is supplied by the separate
`codex-adversarial` leg, which ran independently.

**Coordinator**: ran its own experiments (rebuilt the tree, removed a gate, re-ran suites, ran probes
against a wiped index). Coordinator experiments outrank lane claims below.

---

## 2. What is solid

Credit where the review actually validated the work — several plausible-sounding suspicions died
against real evidence:

- **Schema/cache invalidation is correct and complete.** `INCREMENTAL_SCHEMA_VERSION` 15→18 and
  parse-cache `SCHEMA_BUMP` 22→25 are both present, each increment individually documented, with the
  pinning test updated in lockstep (`toBe(18)` plus a 15/16/17-must-fail reuse-gate matrix).
  The durable ParsedFile store rides the parse-cache version. Refuted by three lanes across both engines.
- **Bench baselines were verified by execution, not by reading.** `scope-capture`, `scope-emission`
  and `callable-value-flow` `--check` all **pass** on head; the TS fingerprint matches the committed
  value; `@scope.block 400 / @scope.function 1400 / @scope.class 200 / @scope.module 200` are exact.
  The `_rebaselined_receiver_owner_2701` note (histogram diff showing `@receiver-owner.this` is the
  only delta) is the right way to re-baseline a fingerprint gate.
- **`var` hoisting past the new block scopes holds.** Probe-verified twice, independently: a block
  containing only `var` is dropped; a block with `var` + a nested `function` is kept; `tsBindingScopeFor`
  walks past `Block` to `Function`/`Module`. The PR's claim is correct.
- **`positionKey` write/read agreement is real** for the single-node case: a local `function save()`
  shadowing a top-level `save()` now resolves to the local — which requires the key to hit.
- **`Variable`/`Const` vs the new position key (obs #37374): refuted by both engines.** Both write
  (`node-lookup.ts`) and read (`ids.ts:172`) gate on the same `isOverloadableCallable`, which excludes
  `Variable`/`Const`. The old simple-key path for value bindings is untouched.
- **Both new provider hooks are live**, not declared-and-dead (`scopeOwnsReceivers`, `thisBoundaryNodeTypes`).
- **Tests assert on EDGES, not node existence** — the only property that catches "the two id phases
  disagree and an edge silently vanishes". The `this-boundary.test.ts` negative cases were checked
  for vacuousness and discriminate correctly. No `any`, no `if`-branching in the new vitests.
- **`for (const x of …)` header bindings and `switch`-case blocks**: probed; **zero delta from this PR**
  (`codex-ce` filed this as a P1; the Engine-A probe refuted it as a regression — see §5).
- **A single-lane P1 with self-declared HIGH confidence still failed the experiment** (CV-3, §4.7).
  The predicted vanished edge did not reproduce.
- Branch hygiene clean: two merge-from-main commits, both harmless relative to the pinned base;
  `git diff --check` clean; no bidi/zero-width/homoglyph characters; no secrets; the new workflow
  steps are literal `run:` strings with no `${{ }}` interpolation.

---

## 3. Already found **and fixed** during this review

### ✅ FIXED — P0: Java anonymous-class identity destroyed (`Worker$1.run` → `Worker.makeHandler.run@7:12`)

`enclosingCallablePrefix` (`gitnexus/src/core/ingestion/workers/parse-worker.ts:800-803`) bounded its
walk with `CLASS_CONTAINER_TYPES`, which contains only *declaration* nodes. A Java anonymous class is
`object_creation_expression > class_body`, so the walk passed straight through it and the new
nested-prefix branch (`parse-worker.ts:2283`) out-ranked the synthesized `Worker$N` owner from
#2550/#2555. Every anonymous inner class declared inside a method — the dominant `new Runnable(){…}`
shape — silently left the javac-compatible `Outer$N` keyspace while its owner `Class` node kept it.

**Corroboration: cross-engine + CI.** `claude-ce-adversarial` (AST probe on the real fixture),
`codex-swarm` GN-2695-1 (confidence 1.00), and GitHub CI (`tests / ubuntu / coverage 2/3`,
`java.test.ts` 247 tests | **4 failed**) all independently. `[reproduced]`

**Fixed after the review** by giving the callable-boundary walk a boundary set covering
`class_body` / `object_creation_expression` / enum-constant bodies; `Method:src/Worker.java:Worker$1.run#0`
verified restored. *(Note: `gitnexus-risk-architect`'s "non-JS/TS blast radius: CLEAR" verdict missed
this — it examined query surface and hook opt-in but not the shared, language-agnostic boundary set.
Recorded as a lane blind spot.)*

### ✅ FIXED — literal NUL byte in `parse-worker.ts`

A raw `\x00` was committed inside the dead `NO_QUALIFIED_NAME` constant (`parse-worker.ts:859`). Effect:
`file(1)` reported the 2 928-line source as `data`, and **`grep`/`rg` returned zero matches with exit 1
for *any* pattern in that file** — silently, so every reviewer grep against the largest changed file
lied. Found by `codex-swarm` (GN-2695-4) and `claude-ce-correctness` F4 — cross-engine. `[reproduced]`

Two *other* files carry NULs; both are pre-existing and outside this diff. When landing the fix, also
delete the constant itself: it is never read, never compared, and its documented tri-state contract is
not needed by a `WeakMap` (`ce-maintainability` #3, `swarm-security` F3, coordinator CV-2).

---

## 4. Open findings, severity-ordered

### 4.1 — P1 · #2701 over-suppression deletes correct `this` edges: static→static, object-literal methods, anonymous class expressions `[reproduced]`


> **Engine attribution, precisely.** Engine B (codex-ce, confidence 100) independently confirmed the **static→static** subtype only. The **object-literal-method** and **anonymous-class-expression** subtypes rest on Engine A execution evidence alone — Engine B does not enumerate them. Since Engine A also authored this PR, treat those two subtypes as single-engine findings that happen to be backed by base-vs-head edge diffs, not as cross-engine corroboration.

**Engines: A + B for the static→static subtype only; A alone for the other two** (see the note above). `claude-ce-correctness` F1/F2 (base-vs-head edge diff, confidence 95/90),
`claude-ce-adversarial` F2 (AST probe), **`codex-ce` #1 (correctness, confidence 100)**.

`(method_definition) @scope.function @receiver-owner.this`
(`languages/typescript/query.ts:109`, `languages/javascript/query.ts:79`) marks **every** method scope
as owning `this`. The query's own comment defends this — *"`method_definition` is marked too and is
unaffected — it also carries a synthesized `this` typeBinding, which the walk consults first"*
(`query.ts:103-104`) — but `synthesizeTsReceiverBinding` has three early `return null`s:

| `receiver-binding.ts` | declines for |
|---|---|
| `:85` `if (isStaticMember(...)) return null;` | static methods / static fields |
| `:94` `if (enclosingType === null) return null;` | object-literal methods (`object` ∉ `TYPE_DECL_NODE_TYPES`, `:51`) |
| `:97` `if (typeName === null) return null;` | anonymous `class` expressions |

In each case the scope is now **owned but unbound** → `walkers.ts:224` returns `true` →
`receiver-bound-calls.ts:282` marks the site handled and `continue`s, which also blocks the
`emitReferencesViaLookup` fallback. The fall-through that `receiver-binding.ts:27-31` still documents
verbatim (*"we leave the binding empty and let chain resolution fall through to the class's static
members lookup"*) is now closed.

**Observed base→head deltas (real pipeline runs, both trees built):**

```
static a(){this.b()} static b(){this.c()}      base: 2 CALLS edges   head: 0
static count = 0; static run(){return this.count}  base: ACCESSES A.run -> A.count   head: 0
class E { static run(){return this.other()} }  (JS)  base: 1 CALLS   head: 0
export const o = { m(){return this.n()}, n(){} }     base: 1 CALLS   head: 0
module.exports = { run(){this.helper()}, helper(){} } base: 1 CALLS   head: 0
export const X = class { m(){this.n()} n(){} }       base: 1 CALLS   head: 0
    (the NAMED twin `class Inner {…}` still resolves — the only difference is the missing name)
```

These are **true-positive losses**, not fabrication removal. In TS/JS `this` in a static method *is*
statically known (it is the constructor object; `tsc` types it `typeof B`), and `this.other()` between
statics is idiomatic — base resolved it to the correct *static*, not to an instance member.
`module.exports = { m(){ this.n(); } }` is one of the most common shapes in Node/JS source, so this is
a broad silent edge loss in JavaScript repos.

The PR's own pinned test (`this-boundary.test.ts:216-226`) covers only static→**instance**, which was
genuinely wrong — which makes the over-suppression *look* intentional. Static→static, object-literal
`this`, and anonymous-class-expression `this` are covered nowhere.

**Fix (both engines converged):** stop using the marker as a proxy for "we synthesized a type".
Either restrict `@receiver-owner.this` to the forms that actually rebind `this`
(mirroring `type-extractors/typescript.ts`'s `THIS_BOUNDARY_NODE_TYPES`, which already excludes
`method_definition` for exactly this reason), or extend synthesis to emit a `this` binding for static
members (→ enclosing class), object-literal methods (→ the binding's name) and anonymous class
expressions (→ the variable name) — the last two are already flagged as deferred work at
`receiver-binding.ts:161-168`. Add the six cases above to `this-boundary.test.ts`.

---

### 4.2 — P1 · the #2699 positional join fails **OPEN**: three triggers, one sink `[reproduced]` (trigger a) / `[code-read]` (triggers b, c)

**Engines: A + B**, and the anchor split below was re-verified in the worktree during synthesis.

All three triggers land in the same place. `resolveDefGraphId`
(`scope-resolution/graph-bridge/ids.ts:169-174`) tries the position key first; on a miss it falls
through the qualified/arity/namespace keys — which cannot match, because the graph node's qualified
name now carries the `@line:col` suffix and the scope-layer def does not — and ends at
`ids.ts:257-258`:

```ts
const simpleName = qn.lastIndexOf('.') === -1 ? qn : qn.slice(qn.lastIndexOf('.') + 1);
return nodeLookup.get(simpleKey(filePath, simpleName));   // source-order FIRST-WRITE-WINS
```

That is precisely the wrong-edge failure #2699 exists to eliminate ("a fabricated caller, chosen by
declaration order"), reinstated whenever the position key misses.

**Trigger (a) — multiline `const fn =` / prettier-wrapped arrow.** `[reproduced]` — **reproduced end-to-end by the coordinator (CV-4)**, which produced a real false CALLS edge:

```
export function run(): number {
  const pick =
    (x: number): number => x * 2;
  return pick(1);
}
export function other(): number {
  const pick =
    (x: number): number => x * 3;
  return pick(2);
}

observed CALLS:
  run   -> run.pick@1:2      correct
  other -> other.pick@6:2    correct
  other -> run.pick@1:2      <-- FALSE: other() never calls run's pick
```

*(originally raised by codex-adversarial `[high]`;
verified during synthesis)* The two phases anchor on **different nodes**:
`tree-sitter-queries.ts:52-56` puts the graph-node `@definition.function` on the outer
`lexical_declaration`, while `languages/typescript/query.ts:201/250` deliberately puts the scope-layer
`@declaration.function` on the **inner `arrow_function`** (documented at `query.ts:187`, `:225`, `:241`).
For
```ts
const pick =
  () => 1;
```
those start on different lines, so `positionKey` can never hit. Two same-named function-local
const-arrows in one file then both resolve to the first. **Every fixture in
`function-local-identity.test.ts` keeps the declaration and its initializer on one line**
(`:72`, `:76`, `:142-143`, `:190`, `:193`), so CI cannot see this — and prettier wraps long arrow
declarations exactly this way. `codex-adversarial` adds that Python lambda bindings use the same
inner-vs-outer anchor split, so this is not JS/TS-only.

**Trigger (b) — the `AMBIGUOUS_POSITION` tombstone.** *(codex-swarm GN-2695-2 conf 0.96, codex-ce #2
conf 100, codex-adversarial, claude-ce-testing P2 — cross-engine, 4 lanes)* Two same-named same-label
callables on one physical line write the `''` tombstone (`node-lookup.ts:118`); `ids.ts:169-172`
explicitly lets it "fall through to the name-based keys", i.e. to the first-wins fallback. Trigger:
`if (a) { const f = () => 1; return f(); } else { const f = () => 2; return f(); }` on one line
(minified or hand-formatted). Note the asymmetry: the *sibling* mechanism in
`buildGraphTargetIndex` **is** unit-tested for ambiguity
(`callable-value-target-index.test.ts:89-97`, "REJECTS an ambiguous position claimed by two
callables") — the `resolveDefGraphId` consumer of the same constant is not tested anywhere.

**Trigger (c) — `.vue` SFCs are structurally blind to the join.** *(claude-ce-adversarial F4,
gitnexus-risk F2, codex-ce #5 — cross-engine)* Graph nodes are written in **file** coordinates
(`parse-worker.ts:2561`, `+ lineOffset`), while `localIdentity` (`parse-worker.ts:790-791`) and the
scope extractor use raw **extracted-script** coordinates (`grep -rn lineOffset` returns zero hits in
`scope-extractor.ts`). For any SFC with a `<template>` before `<script>` the key can never hit. This
is a gap in a new feature rather than a regression (neither positional join covered Vue before), but
the fallback is fail-open, so a `.vue` function-local shadowing a file-level same-named callable gets
the wrong edge. There is no `.vue` case in `function-local-identity.test.ts` (`.ts`/`.py`/`.php`/`.js` only).

**Fix:** make function-local definition lookup **fail closed** — return `undefined` after a position
miss or an ambiguity tombstone rather than reaching `simpleKey` — and give both phases one shared
anchor (column or byte offset, and one coordinate system for Vue). Add multiline, same-line-sibling
and `.vue` regression tests asserting both calls resolve independently.

---

### 4.3 — P2 · JS/TS block scopes drop `ACCESSES → Const` edges corpus-wide (−32%) `[reproduced]`

**Engine A only (`ce-correctness` F3, confidence 75) — but reproduced end-to-end**, which is the
stronger signal.

`(statement_block) @scope.block` (`languages/typescript/query.ts:142`, `javascript/query.ts:95`).
Analyzing `gitnexus/src/core/ingestion/pipeline-phases/runner.ts` **alone**:

```
base:  ACCESSES Function:runner.ts:topologicalSort -> Const:runner.ts:name   (×6)
head:  (none)
```

The reads are the in-block `sorted.push(phaseMap.get(name)!)` (`runner.ts:62`) and
`reverseDeps.get(name) ?? []` (`:64`), inside the same `while` block that declares
`const name = queue.shift()!` (`:61`). With `GITNEXUS_KEEP_LOCAL_VALUE_SYMBOLS=1` the **node sets are
byte-identical** between base and head — so this is *edge* loss, not node loss; the node then
disappears in the default configuration only because `local-symbol-pruner.ts` deletes block-scoped
value symbols with no semantic edge left.

Corpus-wide (`gitnexus/src/core/ingestion`, 500+ files, ids normalized so the #2699 renames cancel):

```
ACCESSES  base 3108 → head 2775   (lost 333, gained 0)
  by target label:  Property 2033→2033, Method 37→37, Function 2→2, Const 1036→703
CALLS     base 10201 → head 10202 (lost 5, gained 6)   ← clean, no regression
nodes     7792 → 7762   (the 30 pruned Consts)
```

CALLS is unaffected; the loss is entirely `ACCESSES → Const`. **Nothing in the PR mentions an ACCESSES
effect** — not the repo-manager v18 note, not `this-boundary.test.ts`, not
`function-local-identity.test.ts`, not the scope-emission bench; the block-scope rationale is stated
purely as "sibling-block `let`/`const` are different bindings". `impact`/`context`/any ACCESSES query
over a block-local binding now returns nothing. A one-third reduction of a whole edge population
landing with no test and no note needs either a fix or an explicit, tested decision. Cheap
discriminator offered by the lane: `resolveDefGraphId` is fine for `Const` (the position key is gated
on `isOverloadableCallable`), so the break is upstream, in reference resolution / caller attribution
for `Block`-kind scopes.

---

### 4.4 — P2 · the "three independently load-bearing gates" claim is false for the shipped code `[reproduced]`

**Coordinator experiment** (outranks lanes), originating from `ce-testing` (P2, confidence 55).

Commit `d7aa8c81` and the header of `test/integration/this-boundary.test.ts` both state that the #2701
fix needs three gates and that "the false edge survived removing any ONE of them alone". The
coordinator removed **only** gate 2 —
`if (scope.ownsReceivers?.has(receiverName) === true) return undefined;` in
`gitnexus-shared/src/scope-resolution/registries/lookup-core.ts` — rebuilt both packages and re-ran the
suite:

```
baseline (gate 2 present): 10 passed
gate 2 REMOVED:            10 passed     <-- no test detects its absence
```

Why: in `scope-resolution/pipeline/run.ts`, `emitReceiverBoundCalls` runs **first** and gate 3
(`isReceiverOwnedButUnbound`) adds the site to `handledSites`; `emitReferencesViaLookup` then skips
exactly those sites. For a `this` receiver, gate 3 subsumes gate 2. In `callableFlowOnly` mode both
emit passes are skipped, so gate 2 has no graph-affecting path there either.

The claim *was* true during development (adding gate 2 dropped the false edge's confidence 1.0 → 0.61
by removing the `type-binding` and `owner-match` evidence) — but on an intermediate tree, **before
gate 3 existed**, and it was carried into the final commit message without re-testing. Not a runtime
defect; it is (a) a false claim in a commit message and a test header future maintainers will rely on,
and (b) genuinely untested code in a shared package. Either delete gate 2 as redundant (re-run the
closure-binding and scope-resolution suites to confirm behaviour-neutral) or keep it as deliberate
defense-in-depth — **either way both texts must be corrected.**

---

### 4.5 — P2 · the new scope-emission CI gate is TypeScript-only, so a JavaScript-only regression ships green `[code-read]`

**Engines: A + B.** `codex-ce` #6 (confidence 100), `claude-ce-correctness` (testing gaps),
`claude-ce-adversarial` F5/F6.

`bench/scope-emission/measure.mjs:34-35` imports **only** `emitTsScopeCaptures`. The JavaScript emitter
carries an independent copy of `JS_FUNCTION_BODY_OWNER_TYPES` / `BLOCK_BINDING_CHILD_TYPES` /
`blockDeclaresBinding` (`languages/javascript/captures.ts:61-87`, byte-identical to the TS twin at
`typescript/captures.ts:64-90`). The CI step is named "JS/TS scope set", but a JS-only divergence
cannot fail it. The gate also compares aggregate counts only, so a *swap* of two emitted scopes is
invisible. `emit_ms_budget` is additionally a wall-clock gate on shared CI runners.

**Fix:** run separate JS and TS corpora, and fingerprint emitted ranges/parents rather than totals.
Secondary: `ce-maintainability` #2 notes the PR introduces `utils/callable-flow-captures.ts` — a shared,
config-driven module built specifically to stop this JS/TS copy-paste — while adding new duplicated
block-filter logic in the same diff. Sharing the filter would collapse this finding.

---

### 4.6 — P3 · the documented "lockstep guarantee" between the two id builders does not hold (latent) `[reproduced — as a refutation]`

`parse-worker.ts:905` (caller phase) vs `:2282` (definition phase). The definition phase appends
`localIdentity(...)` (`run.save@2:2`); `findEnclosingFunctionId` composes `${ownerName}.${funcName}`
with **no position** (`run.save`). The header two screens above asserts the opposite: *"This pair is
the lockstep guarantee… an id they compute differently is not a test failure — it is a caller silently
attaching to a node that does not exist."*

**Raised as P1 by `gitnexus-risk-architect` with self-declared HIGH confidence and a
character-by-character trace — and the predicted vanished edge was REFUTED, twice.** The coordinator
probed a PR-head build against a wiped index:

```
nodes: Function:f.ts:helper | Function:f.ts:outer | Function:f.ts:outer.inner@2:2
CALLS: outer -> outer.inner@2:2
       outer.inner@2:2 -> helper      <-- the edge predicted to VANISH
```

`ce-correctness` reached the same conclusion independently (zero dangling edge sources across 60
fixtures in 12 languages). CALLS sources are attributed by `resolveCallerGraphId` in the
scope-resolution bridge, not by `findEnclosingFunctionId`. `findEnclosingFunctionId` *is* still live
(3 call sites: `parse-worker.ts:1619`, `:1811`, `:1987`) feeding the worker's assignment/call records,
which nothing outside `cobol-processor.ts` currently reads.

**What survives: the source asymmetry is real and it falsifies the PR's own stated invariant.** Any
future consumer of the worker caller id dangles silently — exactly the failure mode the comment warns
about. Fix is one line (apply `localIdentity` on the caller side, or route both phases through
`callableOwnQualifiedName`) — or delete the "lockstep" claim.

*(Methodological note worth keeping: an initial `grep` for `findEnclosingFunctionId` returned only a
comment — that was the NUL byte in §3 making grep silently lie. Confirmed with a Python walk.)*

---

## 5. Lower-priority items, disputes, and test gaps

**Lower-priority (all `[code-read]` unless noted):**

- **P3 — `generator_function` is in `@scope.function` but not in `FUNCTION_NODE_TYPES`**
  (`typescript/query.ts:116` vs `typescript/captures.ts:52-63`, whose comment claims the lists are
  "kept in sync"). Consequence: a generator *expression*'s body block is not recognised as a function
  body, so it gets a redundant Block level and its locals bind there. `ce-adversarial` F6 (probe),
  `gitnexus-risk` F5, `codex-ce` #6 — both engines.
- **P3 — `blockDeclaresBinding` misses three real block binding forms**: `abstract_class_declaration`,
  `enum_declaration`, `internal_module` (`typescript/captures.ts:72-77`). Probe-verified; **not a
  regression** (no block scopes existed before), but it is an incompleteness in a filter documented as
  complete.
- **P3 — `isVarDeclaration` classifies `var` by raw text prefix** (`typescript/simple-hooks.ts:101-107`:
  `'var '`/`'var\t'`/`'var\n'`). Pre-existing helper, **newly load-bearing** because blocks now exist:
  `if (flag) { const marker = 0; var/**/pick = () => 1; }` keeps the block (the `const` declares) but
  the comment separator defeats the prefix test, so `pick` stays block-scoped where JS hoists it.
  Existing coverage uses a block containing *only* `var`, which the filter discards — so it cannot
  expose this. Single lane (codex-adversarial `[medium]`). Fix: classify from the AST/keyword capture.
- **P3 — the PR description is materially stale** (codex-swarm GN-2695-5, confidence 1.00): it says
  "The diff contains no `@scope.` change at all" and that the PR deliberately does not fix #2699/#2701,
  while the diff adds `@scope.block`, function-local identity and `this` boundaries and GitHub lists
  all three issues as closing refs. It also cites schema versions 16/23; the code ships 18/25.
- **P3 — `bench/callable-value-flow/measure.mjs:6,121` includes a `Static` case** that production
  explicitly excludes (`callable-value-flow.ts:835`); it takes the cheapest early-exit path and
  understates the real workload. Replace with `Const`/`Property`/`Variable` and regenerate.
- **P3 — Dart is not covered by #2699.** A Dart local closure still gets an unqualified id
  (`Function:f.dart:inner`) because Dart's callable node is a `function_signature`, excluded from
  `LOCAL_SCOPE_BODY_NODE_TYPES` — i.e. the exact collision #2699 fixes elsewhere still exists there.
- **P3 — `.vue` gets one of #2701's three layers.** `vueProvider` reuses `TYPESCRIPT_QUERIES` and
  `typescriptConfig` (so `thisBoundaryNodeTypes` applies) but `scopeOwnsReceivers` is wired only to
  the TS and JS providers (`typescript.ts:350`, `:418`), so `Scope.ownsReceivers` stays `undefined` for
  `.vue` — a third behaviour neither of the two the PR reasons about.
- **P3 — PHP sigil asymmetry**: `ids.ts`'s `simpleNameOf` lacks the `withoutSigil` normalisation its
  twin in `callable-value-flow.ts:779` has and documents. PHP locals currently resolve through the
  sigil-normalised index instead, so this degrades to the fail-open `simpleKey` rather than
  mis-resolving. Same theme as §4.2.
- **P3 — shared-code comment names six languages** (`callable-value-flow.ts:763`), which `codex-ce` #7
  reads as an `AGENTS.md` language-neutrality violation. `claude-ce-standards` audited the same rule
  and passed the diff, treating comments as out of scope. Disputed; comment-only either way.
- **Operational note (accepted trade, worth stating in the PR):** node ids now embed `line:col`, so
  inserting a line above a function renames every function-local callable below it. The schema bumps
  handle the index, but anything holding ids across runs — saved queries, wiki anchors, PDG anchors,
  agent notes — will not survive an edit.

**Disputed / refuted by evidence (do not act on these):**

- `codex-ce` #3 "**loop-header bindings escape the new block scopes**" (P1, conf 75) — the Engine-A AST
  probe found `for (const x of …)` binds in the enclosing Function scope **identically to pre-PR**
  (body block is dropped, `for_in_statement` gets no scope). A genuine ECMAScript-fidelity gap, **zero
  delta from this PR**. Not a regression, not merge-blocking.
- `gitnexus-security` F1/F2 "**stack overflow / O(N²) strings from unconstrained mutual recursion**" — **both the finding and its refutation are Engine A (Claude); the coordinator did NOT independently verify either.** The mechanical argument is sound but it is same-engine self-assessment, so treat the refutation as provisional rather than settled.
  (P2, HIGH) — `ce-correctness` refuted: depth is AST nesting depth over an acyclic parent chain, every
  level memoized, and the file is already walked recursively by `walkNamedTree`; depth-4 closure chains
  and 3-level nesting ran clean. The suggested iterative restructure is still a fine cleanup, but there
  is no demonstrated crash.
- Refuted with real triggers and worth noting as validation: `@`-collision in `simpleNameOf`'s
  `/@\d+:\d+$/` strip; `AMBIGUOUS_POSITION = ''` leaking to other lookup consumers; the position key
  producing a *false* hit (tombstoned); C++ UE preprocessing shifting coordinates for one phase;
  `switch`-case and `using` block forms; Java **enum-constant-body** methods (`prefix=NONE` — the
  `M3$1.hook` assertions were never at risk); named local classes; the `WeakMap` outliving its tree;
  `findEnclosingClassName` cache poisoning by the new parameter; `Scope.ownsReceivers` (a `Set`)
  surviving the parse cache (`parse-cache.ts:260-274` already round-trips `Set`).

**Test gaps (from three lanes, deduped):**

1. No test for static→static `this.x()`, static `this.field`, object-literal-method `this`, or
   anonymous-class-expression `this` (§4.1). The single static case asserts `toEqual([])` for
   static→**instance**, which makes the over-suppression look intentional.
2. No test asserts an `ACCESSES` edge count anywhere in the new suites — the −333 delta (§4.3) is
   invisible to CI.
3. No test constructs the fail-open triggers of §4.2: multiline `const fn =` + wrapped arrow,
   same-line sibling-block same-name callables, or any `.vue` fixture.
4. No test places a call site **inside a nested named callable** — the only shape that exercises the
   caller-phase branch of §4.6.
5. No consistency test between `THIS_BOUNDARY_NODE_TYPES` (4 entries) and the query-side
   `@receiver-owner.this` list (8 entries). The type-env comment states the two "must agree or a call
   suppressed by one is re-introduced by the other" — an unenforced prose invariant.
6. `SCHEMA_BUMP` has no pinning test, unlike `INCREMENTAL_SCHEMA_VERSION`.
7. Only 2 levels of prefix-builder nesting are exercised; the mutually-recursive composition is
   untested at 3+.

---

## 6. CI status (head `64073d90`)

| Check | Result |
|---|---|
| `tests / ubuntu / coverage 2/3` (**required**) | ❌ **FAILED** — `java.test.ts` 247 tests, **4 failed** (anonymous-class ownership, receiver dispatch, `Types$1.run` local-type construction, anonymous class in an enum constant). This is §3's P0, now fixed. |
| `CI Gate` (**required**) | ❌ failed (downstream of the above) |
| `benchmarks` | ✅ green — and all three `--check` gates independently re-run and passed by a lane |
| typecheck, lint, format, CodeQL, gitleaks, Windows shards, macOS shards, other Ubuntu shards | ✅ passed |
| Coverage merge | skipped · Vercel ❌ unrelated team-authorization failure |
| `mergeStateStatus` | **BLOCKED**; no approving review recorded |

---

## 7. What was NOT reviewed

- **No lane ran the full `npm test` suite**, and `npx tsc --noEmit` was never seen green: the Codex
  legs were strictly read-only, so the `gitnexus-shared` package was not rebuilt and the CLI's resolved
  `Scope` type did not yet expose `ownsReceivers` (a build artifact, **not** a type error in the diff).
  `function-local-identity.test.ts` **loaded but skipped all 8 cases** in the Codex run because the
  parse worker was not built.
- **Kotlin anonymous objects** (`object_expression` / `anonymous_object`) were never probed — no
  `tree-sitter-kotlin` in the worktree. They are absent from `CLASS_CONTAINER_TYPES`, so the §3 P0
  mechanism plausibly applies there; the Java-only fix should be checked against Kotlin.
- **Performance was not measured on a real large repo.** The block-scope filters are justified by a
  synthetic bench only; the claimed ~10% analyze-time effect was not independently reproduced.
- **Windows/macOS behaviour** was taken from CI status only; no platform-specific reasoning was done.
- **`gitnexus-shared`'s own test suite** was not run; the gate-2 experiment covered only
  `this-boundary.test.ts`.
- The `codex-swarm` leg's GitNexus index was **21 commits stale**, so its graph results are topology/
  blast-radius evidence only, not source truth.
- Not reached by any lane: the `callable-value-flow.ts` `resolveCallerGraphId` path end-to-end,
  `utils/callable-flow-captures.ts`, the Dart `previousSibling ?? current` anchor, and cold-vs-warm
  parse-cache parity for the new id shape.
- Two files outside this diff still contain literal NUL bytes; not investigated.

---

*Automated multi-tool review digest: two independent engines (Claude and Codex `gpt-5.6-sol`/xhigh)
across 11 lanes, plus coordinator experiments on a built tree. Three Claude `gitnexus-*` lanes stalled
and were resumed with narrowed briefs, and the Codex `ce` leg could not start its internal cross-model
sub-pass — treat their coverage as partial. Findings tagged `[reproduced]` were executed; `[code-read]`
findings were confirmed in the worktree at `64073d90` but not run. Please verify before acting.*

## 8. Classification and verdict

| Field | Value |
|---|---|
| Branch hygiene | **merge-from-main commit present but harmless and merge-safe** |
| Merge state | **checks failing** |
| Review action | **COMMENT** (collaborative — not `REQUEST_CHANGES`, not `APPROVE`) |
| Final verdict | **not production-ready** |

**Justification.** Two independent correctness regressions remain open against code that the base
branch resolved correctly: #2701's `this`-boundary marker suppresses member calls in every scope
where `synthesizeTsReceiverBinding` deliberately returns null (static members, object-literal
methods, anonymous class expressions), and #2699's positional join falls through to a
first-write-wins simple-name key on a position miss, which reproducibly mints a false CALLS edge for
a multiline `const fn =` declaration — the same fabricated-caller class this PR's own #2693 work
already shipped once. A third issue silently removes 32% of `ACCESSES → Const` edges corpus-wide
with no compensating gain and no test. The Java anonymous-class P0 and the NUL byte were found by
this review and are already fixed, and the schema/invalidation work is sound, but the open items
change graph output in ways no test in this PR detects. The PR is close, and the remaining fixes are
well-localised, but it should not merge until the two P1s are resolved and covered by tests that
exercise the multiline and non-`this`-binding shapes.


---

> **Provenance.** Reviewed at `64073d90`. The PR head moved to `4e19c712` during the review (an unrelated `main` merge bringing #2694 — `cli/setup.ts`, `editor-targets.ts` and two tests). None of the reviewed files changed between those commits, and all three inline anchors were verified byte-identical at the new head, so this review is posted against `4e19c712`.

Issue #2184, opened 2026-06-13
Install FTS feature


mohammad-kasaei-mohaymen:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

in a Debian environment
I try to install gitnexus@1.6.7
but I can not use FTS

### Context

I am trying to install simply using npm but without network connection. I have an NPM mirror accessible not any other network connection.
but I can not use FTS.

```
{"level":40,"time":1781226045556,"name":"gitnexus","msg":"GitNexus: FTS extension unavailable; continuing without FTS features. load-only policy: extension not pre-installed"}
{"level":40,"time":1781226060702,"name":"gitnexus","msg":"GitNexus: FTS extension unavailable; continuing without FTS features. INSTALL fts timed out after 15000ms"}
```

### Expected behavior

To be able to install the FTS extensions in the fully offline environment. independent of network connection. 
Every dependency must be downloaded just from the NPM registry that I have no other parts of the internet because I don't have internet connection in my environment

### Actual behavior

Does not have FTS in fully offline mode

### Steps to reproduce

1. in a fully offline mode install gitnexus
2. try to analyze a repo

### Environment

Doctor command
```
[INFO ] Analyze Doctor: GitNexus Doctor

Runtime
  OS:       linux/x64
  Node:     v22.22.3
  GitNexus: 1.6.7
  LadybugDB:unknown
  native    âœ“ lbugjs.node loaded
  ONNX:     1.26.0

Capabilities
  Graph store:      available
  Full-text search: available
  VECTOR index:     available
  Semantic mode:    vector-index
  Ext install:      load-only (offline; load only, no network install)
  Exact scan limit: 10000 chunks

Embeddings
  Backend:    local
  Device:     auto
  Threads:    4
  Batch:      16 nodes
  Sub-batch:  8 chunks
  Support:    âœ“ local embeddings supported
```

### Logs / screenshots

```
Incremental: changed=0, added=0, deleted=0 (skipping wipe + 15 unchanged file rows preserved)
{"level":40,"time":1781226045556,"name":"gitnexus","msg":"GitNexus: FTS extension unavailable; continuing without FTS features. load-only policy: extension not pre-installed"}
{"level":40,"time":1781226060702,"name":"gitnexus","msg":"GitNexus: FTS extension unavailable; continuing without FTS features. INSTALL fts timed out after 15000ms"}
[2K
[...]

PR #2698, opened 2026-07-26
fix(fts): stop warning "FTS extension unavailable" on the run that installs it

diarized:
## Problem

On a machine with no cached LadybugDB extension, the **first** `gitnexus analyze` logs a WARN:

```
{"level":40,"name":"gitnexus","msg":"GitNexus: FTS extension unavailable; continuing without FTS features. load-only policy (no install attempted); LOAD fts failed: ..."}
FTS: creating File.file_fts
FTS: ready File.file_fts
... (all ~20 FTS indexes created successfully)
```

…and then installs FTS and builds every search index **in that same run**. Nothing is degraded — the log line is wrong. It reads as a broken install path, which is what makes it expensive: it sends users (and reviewers) looking for a failure that never happened. The same false line is the *first* of the two warns in #2184, where only the second one is real.

This is a **log-correctness fix, not a functional FTS fix.** There is no broken install path here.

## Root cause

`initLbug` pre-loads FTS with the writable branch's default policy (`load-only` — analyze owns extension installation, read paths must never trigger a network INSTALL). On a cold cache that pre-load is *expected* to miss; `run-analyze` Phase 3 retries with `auto` moments later and succeeds:

```
initLbug            → loadFTSExtension(..., {})       → load-only → LOAD fails → WARN   ← the bug
run-analyze Phase 3 → loadFTSExtension(..., 'auto')   → INSTALL + LOAD succeed          ← the truth
createSearchFTSIndexes()                              → all indexes built
```

`ExtensionManager.markUnavailable` had no way to distinguish that speculative probe from a final answer, so it reported every miss as a user-facing degradation.

## Change

- `ExtensionEnsureOptions` gains `quiet?: boolean`. The capability entry is still recorded; the message goes to `logger.debug`, and the probe does **not** consume the once-per-`(extension, reason)` warn budget — so an identical reason arriving later from a caller that *does* own the outcome still warns.
- The writable `initLbug` pre-load is the only caller that sets it.

Install policy is deliberately left alone: the writable branch still resolves from env, so `GITNEXUS_LBUG_EXTENSION_INSTALL=never` keeps short-circuiting as before. Only the log level of a speculative probe changes.

**Untouched, still warn loudly:** the read-only serve/MCP branch (`{ policy: 'load-only' }`, no later retry — that warning is accurate), analyze Phase 3, `--repair-fts`, pool read paths, and genuinely-offline installs (#2184, whose "INSTALL fts timed out" warn is unaffected).

## Considered and rejected

- **Dropping the writable pre-load entirely.** Writable callers that reach FTS query functions without going through Phase 3 rely on the extension being LOADed on that connection.
- **Giving the writable branch `auto`.** That moves a network INSTALL into DB initialisation, contradicting the documented split where analyze owns extension installation.

## Testing

`test/unit/lbug-extension-loader.test.ts` (+2 cases, 25 pass):

- quiet probe misses → returns `false`, **no warn**; a following `auto` call installs, loads, reports `{ fts, loaded: true }` — still no warn. *(Verified this case fails on `main`.)*
- a genuine non-quiet `load-only` miss with the **same reason** after a quiet probe still warns exactly once — the dedup key must not be consumed by the probe.
- existing "warns when policy=load-only" / "warns at most once per (extension, reason)" cases unchanged, so serve/MCP keeps its warning.

Also ran `lbug-extension-loader`, `incremental-fts-drop-ordering`, `incremental-vector-extension-ordering` (28 pass), lint (0 errors) and the pre-commit typecheck gate.

End-to-end, cold cache against a temp `HOME` with no `~/.lbdb`:

```
$ HOME=$tmp gitnexus analyze --verbose 2>&1 | grep -iE 'fts|extension'
FTS: creating File.file_fts
FTS: ready File.file_fts
...                                    # no "FTS extension unavailable" line
$ ls $tmp/.lbdb/extension/*/*/fts/
libfts.lbug_extension
```

### DIFF
--- a/gitnexus/src/core/lbug/extension-loader.ts
+++ b/gitnexus/src/core/lbug/extension-loader.ts
@@ -43,6 +43,18 @@ export interface ExtensionCapability {
 export interface ExtensionEnsureOptions {
   policy?: ExtensionInstallPolicy;
   installTimeoutMs?: number;
+  /**
+   * Speculative probe: log an unavailable outcome at debug level instead of
+   * warn, and do not consume the once-per-(extension, reason) warn budget.
+   *
+   * Set only by callers that do NOT own the extension's lifecycle and know a
+   * later owner will retry with an install-capable policy — currently the
+   * writable `initLbug` pre-load, whose miss on a cold cache is expected and
+   * is fixed moments later by analyze Phase 3. Never set it on a path where
+   * the failure is the final answer (serve/MCP read paths), or a real
+   * degradation goes unreported.
+   */
+  quiet?: boolean;
 }
 
 export interface ExtensionManagerOptions {
@@ -230,9 +242,10 @@ export class ExtensionManager {
     const timeoutMs =
       opts.installTimeoutMs ?? this.options.installTimeoutMs ?? getExtensionInstallTimeoutMs();
     const warn = this.options.warn ?? ((msg: string) => logger.warn(msg));
+    const quiet = opts.quiet === true;
 
     if (policy === 'never') {
-      this.markUnavailable(name, label, 'extension install policy is "never"', warn);
+      this.markUnavailable(name, label, 'extension install policy is "never"', warn, quiet);
       return false;
     }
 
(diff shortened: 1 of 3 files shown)


magyargergo:
Thanks for this PR! :)

PR #2700, opened 2026-07-26
fix(storage): stop the Windows `\\?\` long-path prefix from breaking repo path matching (#2667)

### LINKED ISSUE #2667: Windows: \?\ long-path prefix leaks into node IDs and query output
On Windows, node IDs and `filePath` properties in query output carry the `\\?\` long-path prefix (e.g. `Function:\\?\D:\Projects\aptos-core\...\market.move:0xc0deb00c...::market::place_limit_order`). Cosmetic but noisy: it leaks into every `cypher`/`context` result, makes IDs harder to eyeball, and the escaped form (`\\\\?\\`) quadruples in JSON.

**Fix sketch:** strip the `\\?\` prefix at display/emit boundaries (or normalize at ingestion) while keeping long-path support internally. Observed with `gitnexus@1.6.10-aptos.0` on aptos-core and econia indexes.

magyargergo:
Closes part of #2667.

## What this fixes

On Windows a caller can hand GitNexus a `\\?\`-prefixed path. That is the usual MAX_PATH workaround, and `path.resolve` preserves the prefix, so it reaches every string comparison GitNexus keys paths on.

`canonicalizePath` is the single comparison key for the repo registry, MCP repo resolution and the server repo routes, and `registryPathEquals` compares its output as a plain string. A repo registered as `D:\repo` was therefore invisible to a caller passing `\\?\D:\repo`. That surfaces as "repo not found" or a duplicate registration from `analyze`, `remove`, `clean`, the MCP `repo` parameter and the server routes.

The prefix also poisons relativization. `path.win32.relative` cannot express a relative path between a prefixed and an un-prefixed form of the same directory, so it returns the absolute target instead:

```
path.win32.relative('D:\\repo', '\\\\?\\D:\\repo\\a\\b.move')  ->  '\\\\?\\D:\\repo\\a\\b.move'
```

That absolute string is the shape reported in #2667, and it is the same defect class as the cross-drive `isInside` bug fixed in #2688.

## What changed

* New `stripWindowsLongPathPrefix` in `gitnexus/src/lib/utils.ts`. Pure, idempotent, POSIX no-op, explicit `platform` argument so it is testable off Windows. It handles the drive form and the UNC form in any case, and deliberately leaves `\\?\Volume{GUID}\...` alone because the remainder of a volume GUID path is not a usable path.
* `canonicalizePath` applies it on **both** branches. The realpath branch was already safe, because libuv strips the prefix inside `fs__realpath`, but the `catch` fallback returned `path.resolve(p)` untouched and that is exactly the branch a path which is not on disk takes.
* Two comments corrected. Both said the opposite of how things actually work and would send the next reader the wrong way.

## Scope, and why the obvious fix is not here

The normalization is deliberately confined to the comparison domain. No string that reaches `fs` is modified.

The first version of this change normalized the analyze repo root instead. That is wrong. libuv's `fs__capture_path` only converts WTF-8 to UTF-16 and does not re-add the prefix for paths over MAX_PATH, so stripping a filesystem facing path would break long path access on hosts that have not opted into `LongPathsEnabled`. In other words it would break exactly the users who type the prefix in the first place.

Safe despite the blast radius on `canonicalizePath` (27 impacted symbols, 12 direct dependents) because the result is only ever compared, never opened. All 23 call sites feed `registryPathEquals` or a string comparison, both operands are canonicalized, and behaviour is bit identical for every un-prefixed input. `assertSafeStoragePath`, the guard in front of `fs.rm(recursive)` in `remove` and `clean --all`, compares `path.resolve` forms and never calls `canonicalizePath`, so it is untouched.

## Two comments that were wrong

* `analyzer-identity.ts` claimed the `\\?\` prefix is one that `realpathSync.native` "can emit for paths over MAX_PATH". libuv's `fs__realpath_handle` strips it unconditionally and rewrites `\\?\UNC\` back to `\\`, erroring if neither is present, so realpath never returns one. The prefix can only arrive from caller supplied input. `normalizeAnalyzerRootPath`'s behaviour is unchanged on purpose, since those identity fields are compared between an `analyze` and a later `status` run.
* `include-extractor.ts` claimed "gitnexus analyze stores absolute paths in the File.filePath column". A full self index at 89bbdcf5 had 0 of 239,070 nodes with an absolute or backslash bearing filePath. File nodes are built from the walker's repo relative forward slash paths. The relativization guard stays, now described as what it is, a guard against rows the current process did not write.

## What this does not do

It does not reproduce or provably fix the node ID symptom in #2667. Mainline ingestion cannot produce those IDs: `glob` and `path-scurry` were measured to keep paths relative under a prefixed cwd, `walkRepositoryPaths` forward slash normalizes them, and the 239,070 node self index has no absolute filePath anywhere. The report came from a fork build (`gitnexus@1.6.10-aptos.0`) on aptos-core and econia. If the fork's Move provider emits absolute paths then the fix belongs there and this change is complementary hardening. It would help to know the exact `gitnexus analyze` invocation used, and whether the path was passed prefixed.

An index already written with prefixed node IDs is not repaired by this change. `analyze --force` is the remedy.

## Testing

* `gitnexus/test/unit/windows-long-path-prefix.test.ts`, new, fixture free and platform injected, registered in `scripts/cross-platform-tests.ts`. Covers the drive form, the UNC form in upper, lower and mixed case, volume GUID paths, already canonical input, idempotence, the POSIX no-op, the mixed prefix `path.win32.relative` leak shape, and the fact that `path.resolve` preserves the prefix.
* `gitnexus/test/unit/repo-manager.test.ts`, two win32 only assertions covering the realpath branch and the `catch` fallback. They run on the windows-latest leg of the cross platform matrix.
* `tsc` clean, lint clean, changed files Prettier clean, targeted suites green.

Note on the full unit suite in my environment: 10 files fail with worker pool startup timeouts under load. The same files fail identically on the untouched base commit, so they are not regressions from this branch.

### DIFF
--- a/gitnexus/scripts/cross-platform-tests.ts
+++ b/gitnexus/scripts/cross-platform-tests.ts
@@ -49,9 +49,13 @@ const PLATFORM_LOGIC = [
   // returns the absolute target across drives, so the guard needs isAbsolute.
   // Fixture-free and pathApi-injectable, so it is portable to every runner.
   'test/unit/analyzer-identity-is-inside.test.ts',
-  // `\\?\` extended-length prefix normalization (#2667): another POSIX no-op,
-  // so it is only meaningful on windows-latest. Same mixed-prefix relativize
-  // hazard as the is-inside guard above, reached through a caller-supplied path.
+  // `\\?\` extended-length prefix normalization (#2667): fixture-free and
+  // platform-injectable (every assertion passes an explicit 'win32'), so like the
+  // is-inside guard above it is portable to every runner and its assertions run
+  // identically here and on Ubuntu. Registered alongside its two siblings so the
+  // Windows path-handling guards stay discoverable as one group. Same
+  // mixed-prefix relativize hazard as is-inside, reached through a
+  // caller-supplied path.
   'test/unit/windows-long-path-prefix.test.ts',
   // getconf page-size probe: explicit process.platform gate (win32 short-circuit)
   // plus a live-probe test whose only real non-4K coverage is macos-arm64's
--- a/gitnexus/scripts/cross-platform-tests.ts
+++ b/gitnexus/scripts/cross-platform-tests.ts
@@ -49,6 +49,10 @@ const PLATFORM_LOGIC = [
   // returns the absolute target across drives, so the guard needs isAbsolute.
   // Fixture-free and pathApi-injectable, so it is portable to every runner.
   'test/unit/analyzer-identity-is-inside.test.ts',
+  // `\\?\` extended-length prefix normalization (#2667): another POSIX no-op,
+  // so it is only meaningful on windows-latest. Same mixed-prefix relativize
+  // hazard as the is-inside guard above, reached through a caller-supplied path.
+  'test/unit/windows-long-path-prefix.test.ts',
   // getconf page-size probe: explicit process.platform gate (win32 short-circuit)
   // plus a live-probe test whose only real non-4K coverage is macos-arm64's
   // 16 KiB pages — the exact hardware class #1231 targets (#2424 review).
--- a/gitnexus/src/core/analyzer-identity.ts
+++ b/gitnexus/src/core/analyzer-identity.ts
@@ -548,10 +548,19 @@ function resolveExistingPath(candidate: string): string {
  * unchanged. `platform` is explicit so the transform is unit-testable off
  * Windows.
  *
- * The optional `\\?\` extended-length prefix (which `realpathSync.native` can
- * emit for paths over MAX_PATH) is preserved and the drive letter after it is
- * still normalized; UNC paths (`\\server\share`, `\\?\UNC\...`) have no drive
- * letter and are left untouched.
+ * The optional `\\?\` extended-length prefix is preserved and the drive letter
+ * after it is still normalized; UNC paths (`\\server\share`, `\\?\UNC\...`)
+ * have no drive letter and are left untouched.
+ *
+ * That optional group is defensive, not a case `realpathSync.native` produces:
+ * libuv's `fs__realpath_handle` strips `\\?\` (and rewrites `\\?\UNC\` back to
+ * `\\`) before returning, so the prefix can only reach here from caller-supplied
+ * input, which `path.resolve` preserves (#2667). Preserving it is deliberate —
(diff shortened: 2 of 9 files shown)


magyargergo:
Added `gitnexus/test/unit/canonicalize-path-long-path-prefix.test.ts` in 0220434f.

The two `canonicalizePath` assertions in `repo-manager.test.ts` drive the real `realpathSync.native`, so they are `it.skipIf(win32)` and only execute on the windows-latest matrix leg. That left the Ubuntu gate, the one every PR runs, with no coverage of this behaviour at all.

The new file closes that. It injects only the platform primitives and runs `canonicalizePath` and `registryPathEquals` unmodified:

* `path` becomes `path.win32`, which is Node's real Windows path implementation rather than a stand-in for it
* `realpathSync.native` gets its two actual behaviours, since libuv strips `\\?\` for a path that is on disk and throws ENOENT for one that is not
* the real `stripWindowsLongPathPrefix` is pinned to win32 instead of defaulting to the host, as a module mock rather than an override of `process.platform`, which is shared by every test file in a worker

It is deliberately not registered in `scripts/cross-platform-tests.ts`, because it simulates Windows rather than needing it.

Verified to discriminate rather than just pass. Against the pre-fix tree at 89bbdcf5 it fails 3 of 5:

```
× matches a prefixed drive path against its stored entry when the repo is gone from disk
× matches a prefixed UNC path against its stored entry when the share is unreachable
× matches whatever case the UNC token is spelled in

expected '\\?\D:\Projects\moved-away'       to be 'D:\Projects\moved-away'
expected '\\?\UNC\server\share\moved-away'  to be '\\server\share\moved-away'
```

Reverting just the two strip calls on this branch reproduces the same three failures. The two that pass either way are the realpath branch and the un-prefixed no-op, neither of which ever leaked, which is the asymmetry the fix is built around.

That asymmetry is worth stating plainly, because it narrows what is actually broken: the failing case is a registry entry whose directory is gone from disk. A moved or deleted checkout, a stale entry, `remove` or `clean` on an alias, an MCP `repo` argument for a path that is not indexed. There `realpathSync.native` throws, `canonicalizePath` falls back to `path.resolve(p)`, the prefix rides through, and the string compare against the stored entry misses. When the directory is present, libuv strips the prefix inside realpath and the lookup already worked.

Issue #2449, opened 2026-07-13
CI: Windows platform shard 1/3 has only 3 seconds of watchdog headroom on RC19


100yenadmin:
## Summary

The RC19 Windows platform-sensitive shard 1/3 completed successfully, but its test step took **14m57s** against the runner's hardcoded **15-minute** watchdog. That leaves three seconds of margin and makes the cross-platform floor vulnerable to ordinary hosted-runner variance.

## Evidence

- GitNexus: `v1.6.10-rc.19` (`d43c479ac58c18d397958f48a99465d6c3110b1d`)
- run: https://github.com/electricsheephq/evaOS-gitnexus/actions/runs/29226560630/job/86741767061
- job: 18m00s total
- `Run platform-sensitive tests`: 05:35:34Z to 05:50:31Z = 14m57s, success
- `scripts/run-cross-platform.ts`: `DEFAULT_TIMEOUT_MIN = 15`
- workflow job timeout: 25 minutes

A paired run with the same RC19 tree and a Ladybug 0.18.1 pin completed the corresponding test step in 9m13s, but these are shared hosted runners, so this is **not** evidence of a dependency performance regression. It does show the current fixed watchdog has insufficient deterministic headroom.

## Risk

A small runner slowdown can convert a green cross-platform floor into a timeout failure. The workflow comments currently say three shards keep the busiest Windows shard comfortably under the watchdog; this run no longer supports that assumption.

## Suggested acceptance contract

- keep the 25-minute outer job bound
- give the busiest Windows shard at least 20% measured headroom, either by rebalancing the fixed file list or raising the inner watchdog to a bounded value such as 20 minutes
- retain fail-loud timeout behavior
- record per-shard test-step duration so future drift is visible

No source/runtime failure was observed, and no fork patch is proposed in this report.

100yenadmin:
Fork reconciliation follow-up from an exact-head rerun on Windows:

- Fork PR: https://github.com/electricsheephq/evaOS-gitnexus/pull/83
- Head: `bb307bd17d4384e7abe811e41d3a01dfc2716afe`
- Run: https://github.com/electricsheephq/evaOS-gitnexus/actions/runs/29278810509
- Attempt 1: Windows platform-sensitive shard 2/3 timed out inside `cli-limit-e2e`; the analyze child returned `status: null`, empty stderr, and only the banner on stdout.
[...]

PR #2716, opened 2026-07-27
fix: serialize global registry transactions across processes

voidfreud:
## Problem

GitNexus keeps the multi-repository discovery index in the global `~/.gitnexus/registry.json`. Each registry writer currently does:

1. read the complete JSON array;
2. modify its in-memory snapshot;
3. atomically replace `registry.json`.

The atomic rename prevents a truncated file, but it does **not** serialize the read/modify/write transaction. Two concurrent `gitnexus analyze`, `clean`, branch-update, or validation-prune operations can therefore read the same snapshot and whichever writer renames last can discard the other operation's registry changes. Separate per-repository index locks cannot prevent this because the conflicting resource is the one global registry file.

This is especially visible when multiple agents index different repositories concurrently: the individual databases may complete successfully while the registry loses entries or branch metadata, making repositories appear missing to MCP discovery.

## Change

Reuse GitNexus's existing cross-platform `acquireIndexLock` primitive with the global GitNexus directory as its lock namespace. The lock is held around the complete registry transaction and released in `finally`, so the existing kernel-owned/socket or file-lock behavior also handles process crashes using the established implementation.

The public behavior and APIs are unchanged. The lock now covers:

- `registerRepo` (including alias and branch merge logic)
- `unregisterRepo`
- `removeBranchIndex`
- `adoptFlatBranchLabel` (including branch-directory cleanup plus registry update)
- `listRegisteredRepos({ validate: true })` validation pruning; ordinary registry reads remain lock-free

Read-only MCP database queries are not placed behind this registry lock; only the short global-registry coordination path is serialized.

## Regression coverage

Added a public-API regression test that starts six distinct registrations concurrently with `Promise.all` and verifies that all six named entries remain in the registry. This specifically guards against lost updates in overlapping registry transactions. It is intentionally a deterministic unit-level overlap test, not a claim of exhaustive process-kill stress coverage.

## Verification

From `gitnexus/`:

- `./node_modules/.bin/vitest run test/unit/repo-manager.test.ts`
  - **1 test file passed**
  - **105 tests passed, 2 skipped**
- `./node_modules/.bin/tsc --noEmit`
  - **passed with no diagnostics**

No database format, MCP tool schema, index layout, or query behavior was changed.


azizur100389:
I pulled the PR head (`8eb58a85`) locally and tested it against the current review branch `azizur/md060-markdown-tables-2709` at `02ebf8f1`. The PR fast-forwarded cleanly in the review worktree.

Local validation:

- `npm install` in `gitnexus-shared` and `npm ci` in `gitnexus` completed. My local Node is `v22.17.1`, so npm emitted expected engine warnings for the repo's `22.18+` floor.
- `npx tsc --noEmit` passed.
- `npx vitest run test/unit/repo-manager.test.ts -t "preserves all entries when distinct registrations overlap"` passed from the `gitnexus/` package context.
- A broader targeted run of repo-manager/index-lock related tests had two Windows-local symlink setup failures (`EPERM` creating symlinks) in existing tests; the new registry-overlap test itself passed.
- `git diff --check origin/main...HEAD` passed.
- Hidden bidi-control scan found none.
- Targeted ESLint on `src/storage/repo-manager.ts` and `test/unit/repo-manager.test.ts` passed with warnings only.

Blocking issue:

- `npx prettier --check src/storage/repo-manager.ts test/unit/repo-manager.test.ts` fails locally, and GitHub's `quality / format` check is also failing. Please run Prettier on those two changed TypeScript files and push the formatting fix.

CI status after the approved workflow started:

- Passing so far: typecheck, lint, typecheck-web, review, gitleaks, Analyze python/js-ts, macOS platform shards, Ubuntu coverage shards, Windows platform shards 2/3 and 3/3, packaged install smoke, tree-sitter ABI, eval containment/locked pytest, benchmarks, CodeQL.
- Still red: `quality / format`, `check-description`, and Vercel authorization.
- At my last refresh, `Build & Push gitnexus`, `tests / ubuntu / coverage merge`, and `tests / windows-latest (platform-sensitive) 1/3` were still pending.

The implementation approach looks reasonable to me, but this is not merge-ready until the formatter failure is fixed and the remaining required CI statuses settle green.

voidfreud:
Current head `34edbeeb` passes the full CI gate (15,626 tests, 0 failures), formatting, lint, typecheck, packaged install, platform shards, CodeQL, and the GitNexus image build. The remaining `gitnexus-web` Docker job failed only while pulling `moby/buildkit:buildx-stable-1` from Docker Hub (`Client.Timeout exceeded while awaiting headers`). I cannot rerun organization workflows from the fork; please rerun that failed job when convenient.

gitnexus/test/unit/repo-manager.test.ts
```diff
@@ -901,6 +901,26 @@ describe('registerRepo name override + collision guard (#829)', () => {
       await parentB.cleanup();
     }
   });
+  it('preserves all entries when distinct registrations overlap', async () => {
```
magyargergo, line 906: This should now because a cross platfrom test in `gitnexus/scripts/cross-platform-tests.ts`

magyargergo (changes requested):

voidfreud:
Rebased onto current `main` at `454d3834`; head is now `237f43eb`. The requested cross-platform coverage is present in `scripts/cross-platform-tests.ts`, and local focused validation passed (`repo-manager.test.ts`: 105 passed, 2 skipped; `tsc --noEmit`: clean). The fresh CI run reached the known nondeterministic Windows `cli-limit-e2e` `status: null` failure documented in #2449; the PR-specific quality/typecheck/format checks passed. I cannot rerun upstream workflows from the fork, so an upstream maintainer must rerun failed jobs. The existing re-review request remains pending.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -1131,6 +1132,23 @@ export const getGlobalRegistryPath = (): string => {
...
 };
 
+/**
+ * Serialize global registry read/merge/write transactions across processes.
+ *
+ * The registry is shared by every indexed repository, so per-index locks do
+ * not protect this file. Reuse the cross-platform index lock primitive with
+ * the global directory as its lock namespace; the handle is kernel-owned on
+ * supported platforms and crash-reclaimable by the existing fallback.
+ */
```
abhigyanpatwari, line 1177: [P1] [code-read, cross-verified: adversarial lane `[reproduced]` against isolated lock primitive]

Same-process self-deadlock: when the analyzed repo's root resolves to the same
directory as getGlobalDir() (default: repoPath === os.homedir(), no
GITNEXUS_HOME override), this lock collides with the per-repo metaDir lock
already held by runFullAnalysis (run-analyze.ts:837) across the whole pipeline.
acquireIndexLock is not reentrant, so the nested acquisition here — reached via
registerRepo (run-analyze.ts:2722, NOT wrapped in try/catch) or
adoptFlatBranchLabel (run-analyze.ts:1335/2743, wrapped) — blocks until the
10-minute default timeout, then either throws uncaught (registerRepo path,
failing the whole analyze run) or logs a warning (adoptFlatBranchLabel path).
Reproduced directly against the lock primitive on both socket and file
backends; the file-backend error literally reports the process's own
pid/invocation id as "another gitnexus analyze."

This also contradicts runFullAnalysis's own docstring (run-analyze.ts:808-811):
"the post-analysis steps in the CLI (skills, registry) run lock-free" — they
no longer do.

Suggest: call registerRepo/adoptFlatBranchLabel after the outer per-repo lock's
finally releases it, or make withRegistryLock detect/skip an already-held
identical lock directory, or give the registry lock a namespace that can never
collide with a per-repo metaDir.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -1131,6 +1132,23 @@ export const getGlobalRegistryPath = (): string => {
...
+const withRegistryLock = async <T>(operation: () => Promise<T>): Promise<T> => {
+  const lock = await acquireIndexLock(getGlobalDir());
```
abhigyanpatwari: [P1] [code-read]

This inherits DEFAULT_TIMEOUT_MS = 600_000 (10 min) — sized for multi-minute
analyze runs, not this millisecond-scale JSON read/merge/write. gitnexus/src/
cli/augment.ts:10 documents "Must cold-start fast (<500ms)" and is shelled out
from Claude Code/Cursor tool-use hooks on every grep/glob/read/search; its call
chain (augment() -> findRepoForCwd() -> listRegisteredRepos({validate:true}))
now routes through this same lock with no independent timeout. Under
contention this can stall an agent's tool-call loop for up to 10 minutes before
degrading gracefully.

Also: no `log`/`onWaitStart` is wired here (contrast run-analyze.ts:828-833),
so contention on any command routed through this lock (list/clean/remove/
augment/MCP resolveRepo) is completely silent until timeout.

Suggest: give withRegistryLock its own short, independent timeoutMs (registry
transactions are normally sub-second) and wire log/onWaitStart diagnostics.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -1502,7 +1532,7 @@ export const removeBranchIndex = async (repoPath: string, branch: string): Promi
  * repos (never self-heals an unregistered repo, per #2264/#1169; the registry
  * check precedes the rm per #2364 review F2) — and no subprocess is spawned.
  */
-export const adoptFlatBranchLabel = async (repoPath: string, branch: string): Promise<void> => {
+const adoptFlatBranchLabelUnlocked = async (repoPath: string, branch: string): Promise<void> => {
```
abhigyanpatwari: [P2] [code-read]

Further down this function (repo-manager.ts:1552, unchanged text so not a
diff line itself — anchoring here at the nearest changed line instead): the
recursive fs.rm(branchDir, ...) runs inside withRegistryLock, holding the
global registry lock for the full duration of a potentially slow rm (large
branch index, Windows AV, network mount). Contrast: clean.ts:65-66 does the
equivalent fs.rm BEFORE calling the now-locked removeBranchIndex, keeping slow
I/O out of the lock — this function's own sibling code shows the narrower,
preferred pattern.

A slow rm here blocks every other concurrent registry operation machine-wide,
and (when called from run-analyze.ts's finalization) also extends the outer
per-repo analyze lock's hold time.

Suggest: acquire the lock only immediately before the rm and again for the
final re-read+write, mirroring clean.ts's split.

gitnexus/test/unit/repo-manager.test.ts
```diff
@@ -901,6 +901,26 @@ describe('registerRepo name override + collision guard (#829)', () => {
...
+
+      const entries = await listRegisteredRepos();
+      expect(entries).toHaveLength(repos.length);
+      expect(new Set(entries.map((entry) => entry.name))).toEqual(
+        new Set(repos.map((_, index) => `concurrent-${index}`)),
+      );
+    } finally {
+      await Promise.all(repos.map((repo) => repo.cleanup()));
+    }
+  });
```
abhigyanpatwari, line 925: [P1, coverage gap — not a shipped-code bug] [reproduced: scratch tests written
and executed against pre-fix/post-fix worktree, then reverted]

This is the only concurrent-overlap regression test among the 5 functions now
behind withRegistryLock. unregisterRepo, removeBranchIndex, and
adoptFlatBranchLabel have zero equivalent coverage. Empirically confirmed the
gap is real: scratch tests reproducing "two removeBranchIndex calls racing on
different branches of the same repo" and "registerRepo racing unregisterRepo"
fail on the pre-fix code (lost update / tmp-rename ENOENT collision) and pass
post-fix — meaning the fix is correct for these paths too, but the shipped
suite would not catch a future regression that narrowed or removed the lock
around them.

Suggest: add at least two more Promise.all-based tests mirroring this one:
(a) concurrent removeBranchIndex on two branches of the same repo, (b)
registerRepo racing unregisterRepo/removeBranchIndex.

abhigyanpatwari (commented):
# PR #2716 Tri-Review Digest — "fix: serialize global registry transactions across processes"

Repo: abhigyanpatwari/GitNexus · Base `51095c19f8a7c641a136549711065f816a383ff6` → Head `14fb1837b9efa760a344f57cd46c1d6f613339b2`
Worktree: `/workspace/.worktrees/pr-2716`

## Review bar for this PR

`~/.gitnexus/registry.json` and the lock guarding it are shared, process-wide state touched by every `analyze`/`clean`/`list`/`remove` invocation, every MCP session's `resolveRepo`, and every editor-hook `augment` call on this machine. A concurrency bug here doesn't stay local to one repo — it can degrade or hang unrelated work. The bar is high: correctness under real concurrent load, not just "passes the new test."

## Problem being solved

Every registry writer (`registerRepo`, `unregisterRepo`, `removeBranchIndex`, `adoptFlatBranchLabel`, validation-pruning reads) independently did read → modify → atomic-rename-write on `registry.json`. The atomic rename prevents a torn file but not a lost update: two concurrent writers can read the same snapshot and whichever renames last silently discards the other's change — most visible when multiple agents index different repos at once. This PR serializes those five operations behind the existing `acquireIndexLock` primitive (from #2658/#2677), scoped to the global directory.

## Current PR state / merge status and mergeability

`mergeable=MERGEABLE`, but `mergeStateStatus=BLOCKED` — **review blocked** — see "Outstanding maintainer review" below; that block is independent of anything in this digest and is not resolved by it.

## Engine breakdown — Codex unavailable this run

**Engine B (Codex/GPT-5.5) hit its workspace spend cap on all three lanes** (`codex-swarm.md`, `codex-ce.md`, `codex-adversarial.md` — each returns only `"You hit your spend cap set by the owner of your workspace"`, zero real findings). **There is no cross-engine corroboration in this digest.** Every finding below comes from Engine A (Claude) only, across 9 lanes: 3 GitNexus-swarm personas (facts-historian, risk-architect, test-ci-verifier) + 6 ce-code-review personas (correctness, testing, maintainability, project-standards, reliability, adversarial). Where multiple Claude lanes independently reached the same conclusion via separate reasoning/code paths, that is labeled **"same-engine convergence"** below — moderate signal, explicitly not independent cross-engine confirmation. Re-run with the spend cap raised to get real cross-engine signal.

## What's solid / correctly validated

The adversarial lane (`claude-ce-adversarial.md`) explicitly probed and **refuted** these scenarios — treat as validated, not open questions:
- **Cross-process lock-ordering inversion** (registry lock vs. per-repo `metaDir` lock acquired in reverse order by two different callers): refuted — only one call site (`runFullAnalysisInner`) ever nests them, always metaDir-then-registry, never the reverse.
- **Lock bypass**: refuted — `writeRegistry` is un-exported, called from exactly 5 sites, all inside the lock-wrapped `*Unlocked` functions; no other module touches `registry.json` directly.
- **Crash between acquire/release** (`process.exit`, `SIGKILL`): refuted as a *new* risk — this PR reuses `index-lock.ts`'s existing crash-safety story (kernel-owned socket release-on-exit; file-backend stale-pid reclaim) unchanged; no new `process.exit` call sites introduced.
- **`listRegisteredRepos`'s conditional lock returning torn/corrupt data**: refuted — `writeRegistry` is atomic (tmp+rename), so an unlocked reader only ever sees a complete pre- or post-write document, never a torn one; worst case is staleness, which is the documented/accepted tradeoff of the `f12cd7f6a` "keep registry reads lock-free" follow-up commit.

Also confirmed correct by multiple lanes independently:
- `withRegistryLock`'s try/finally is sound (acquire outside `try`, `release()` in `finally`, matches the `run-analyze.ts` idiom from the prior #2658/#2677 lock work).
- No same-process reentrancy **within `repo-manager.ts` itself** — none of the 5 `*Unlocked` functions call each other or a locked wrapper (this is separate from the cross-module reentrancy in Headline Finding 1 below, which lives at the `run-analyze.ts` call site).
- `sweepStagingArtifacts` running against `getGlobalDir()` post-acquire is a harmless no-op (regex only matches `lbug.staging.*`/`lbug.new*`, none of which live in the global dir).
- All 5 public wrapper signatures match their `*Unlocked` counterparts; `tsc --noEmit` clean; existing `repo-manager` unit-test suites pass unmodified.
- CI is green across the board.

## Headline finding

### 1. [P1] Same-process self-deadlock when the analyzed repo's root equals `getGlobalDir()`'s parent (default: `os.homedir()`) — `[reproduced]` + coordinator `[code-read]`-confirmed

- **Trigger**: run `gitnexus analyze` against a repo whose root is the user's home directory (e.g. a "dotfiles at `$HOME`" git repo, or a user accidentally analyzing `~`), with no `GITNEXUS_HOME` override. More generally: any `repoPath` such that `getStoragePaths(repoPath).storagePath === getGlobalDir()`.
- **Evidence**:
  - `getStoragePaths` (`gitnexus/src/storage/repo-manager.ts:714-722`, flat/no-branch case): `storagePath = getStoragePath(repoPath)` → `<repoPath>/.gitnexus`.
  - `getGlobalDir` (`gitnexus/src/storage/repo-manager.ts:1123-1125`): `process.env.GITNEXUS_HOME || path.join(os.homedir(), '.gitnexus')`.
  - When `repoPath === os.homedir()` with no `GITNEXUS_HOME`, these are byte-identical.
  - `runFullAnalysis` (`gitnexus/src/core/run-analyze.ts:837`) acquires `acquireIndexLock(writeTarget.metaDir)` and only releases it in the outer `finally` (`run-analyze.ts:876-878`), which wraps the entire `runFullAnalysisInner` call.
  - Inside that still-held scope, the main pipeline path calls `registerRepo(repoPath, meta, ...)` at `run-analyze.ts:2722` and `adoptFlatBranchLabel(repoPath, branchLabel)` at `run-analyze.ts:2743`; the "already up to date" fast path calls `adoptFlatBranchLabel` at `run-analyze.ts:1335`. All three now route through this PR's `withRegistryLock` → `acquireIndexLock(getGlobalDir())` (`repo-manager.ts:1142-1149`).
  - `acquireIndexLock` is **not reentrant**. Verified directly in the worktree: the `registerRepo` call at `run-analyze.ts:2722` is **not** wrapped in try/catch (the preceding try/catch, for the parse-cache save, closes before this line) — so a timeout here propagates uncaught out of `runFullAnalysisInner`, and the whole `analyze` invocation fails after the full wait. The two `adoptFlatBranchLabel` call sites (`:1334-1344`, `:2741-2748`) **are** wrapped in try/catch, degrading to a logged warning instead — but only after the full timeout elapses.
  - This directly contradicts `runFullAnalysis`'s own docstring (`run-analyze.ts:808-811`): "the post-analysis steps in the CLI (skills, registry) run lock-free" — the registry write is neither in the CLI nor lock-free; it runs nested inside the still-held per-repo lock.
  - The adversarial lane empirically reproduced the deadlock/timeout against the isolated lock primitive (both socket and file backends) with a shortened 2s timeout, script at `/tmp/claude-1000/-workspace/b113ad5e-7af5-41dc-8ae4-2653a18ef32f/scratchpad/repro-nested-lock.mjs`; the file-backend output literally reports the process's own pid/invocation id as "another gitnexus analyze" it's waiting on — direct proof of self-contention. The coordinator independently re-traced `getStoragePaths`/`getGlobalDir` this session (confirmed above) and reached the same equality condition by static reading, without re-running the repro.
- **Severity**: P1. Narrow trigger (repo rooted at `$HOME` or an equivalent `GITNEXUS_HOME` collision), but fully deterministic once hit — every `analyze` run against such a repo fails after a ~10-minute hang, with no existing guard (`assertSafeStoragePath` in `repo-manager.ts` only guards destructive `clean`/`remove`, not this).
- **Lanes**: claude-ce-adversarial (Finding 1, `[reproduced]`), claude-ce-correctness (F1 residual risk, independently derived, `[code-read]`), coordinator direct verification this session (`[code-read]`). Same-engine convergence — 2 independent lanes plus direct coordinator re-derivation, all Claude.
- **Suggested fix** (not prescriptive): call `registerRepo`/`adoptFlatBranchLabel` from `runFullAnalysis`'s wrapper *after* the outer per-repo lock's `finally` has released it (restores the docstring's stated intent), or make `withRegistryLock` detect an already-held identical lock directory and skip re-acquiring, or give the registry lock a namespace that can never collide with a per-repo `metaDir` (e.g. a fixed sibling name not derived solely from `getGlobalDir()`).
- **Blocks merge: yes** — deterministic failure for a real (if narrow) repo layout, with no existing guard against it.

## Other findings

### 2. [P1] Registry lock inherits the 10-minute analyze-scale timeout for a sub-second JSON transaction, gating a documented <500ms hook path — same-engine convergence, `[code-read]`

- **Trigger**: any process holds (or, on macOS/Windows file-backend, appears to hold via pid-reuse) the global registry lock while `gitnexus augment` — shelled out from Claude Code/Cursor tool-use hooks on every grep/glob/read/search — is invoked concurrently.
- **Evidence**: `gitnexus/src/cli/augment.ts:10` — `"Performance: Must cold-start fast (<500ms)."` `augment()` → `findRepoForCwd()` (`gitnexus/src/core/augmentation/engine.ts:25-31`) → `listRegisteredRepos({ validate: true })`, which now routes through `withRegistryLock` (`repo-manager.ts:1944-1949`) → `acquireIndexLock(getGlobalDir())` with **no options** (`repo-manager.ts:1142-1144`), so it inherits `DEFAULT_TIMEOUT_MS = 600_000` (index-lock.ts) — a value whose own doc-comment justifies it for multi-minute `analyze` runs, not a millisecond-scale read/merge/write. Confirmed directly: `withRegistryLock` passes zero options to `acquireIndexLock`.
- **Consequence**: on contention, `gitnexus augment` can block up to 10 minutes before its `catch { return null; }` degrades gracefully — up to ~1200x its documented budget, stalling the agent's tool-call loop. There is currently no way to give the registry lock a shorter timeout independently — the only knob (`GITNEXUS_INDEX_LOCK_TIMEOUT_MS`) is a single process-wide env var shared with the per-repo analyze lock.
- **Severity**: P1. This is same-engine convergence, not cross-engine — but 3 independent Claude lanes (gitnexus-risk-architect Finding 2, ce-correctness residual-risk, ce-reliability RELI-1/RELI-2) reached it via separate code paths, and ce-reliability traced a concrete, specific consequence (the `augment` hook path) rather than a generic "could be slow" concern.
- **Lanes**: gitnexus-risk-architect (Finding 2, P2 in that lane's own scale), claude-ce-correctness (residual risk), claude-ce-reliability (RELI-1 P0 / RELI-2 P1 in that lane's own scale — synthesis normalizes to P1 given no test/production evidence of actual contention frequency), claude-ce-adversarial (Finding 2, corroborating). All `[code-read]` — no lane executed a live contention repro against the `augment` path itself (distinct from Finding 1's executed repro).
- **Suggested fix**: give `withRegistryLock` its own short, independent `timeoutMs` (registry read/merge/write is normally sub-second) instead of inheriting the analyze-sized default; wire `log`/`onWaitStart` diagnostics so contention isn't silent (see Finding 4).
- **Blocks merge: yes** — a documented <500ms editor-hook path with no independent timeout knob is a real regression risk, not a hypothetical.

### 3. [P2] `adoptFlatBranchLabelUnlocked`'s recursive `fs.rm` runs inside the registry lock, unlike the PR's own sibling pattern in `clean.ts`

- **Trigger**: a branch-label change during `analyze` (or `clean --branch`) requires deleting a shadowed `branches/<slug>/` sub-index that is large or on a slow/AV-scanned/network-mounted volume.
- **Evidence**: `fs.rm(branchDir, { recursive: true, force: true })` at `gitnexus/src/storage/repo-manager.ts:1552`, inside `adoptFlatBranchLabelUnlocked` (`:1535-1600`), which runs entirely inside `withRegistryLock`. Contrast confirmed directly: `gitnexus/src/cli/clean.ts:65-66` does the equivalent `fs.rm` **before** calling the now-locked `removeBranchIndex`, keeping slow I/O out of the lock — the PR's own sibling code shows the narrower, preferred pattern that `adoptFlatBranchLabel` doesn't follow.
- **Consequence**: a slow/stuck recursive delete blocks every other concurrent registry operation machine-wide (any repo's `registerRepo`, `unregisterRepo`, `listRegisteredRepos({validate:true})`), and — when called from `run-analyze.ts`'s finalization step (Finding 1's call sites) — also extends the outer per-repo analyze lock's hold time.
- **Severity**: P2.
- **Lanes**: gitnexus-risk-architect (Finding 1), claude-ce-maintainability (implicitly, via subprocess/critical-section-scope finding 2 — different angle, same root cause), claude-ce-adversarial (Finding 2, corroborating). Same-engine convergence. `[code-read]`.
- **Suggested fix**: acquire the lock only immediately before the `rm` and again for the final re-read+write (mirroring `clean.ts`'s split), or hoist the `rm` entirely outside the lock with a `dirGone` flag passed through.
- **Blocks merge: maybe** — real degradation risk under a slow/AV/network-mount delete, but not deterministic and not silently corrupting; reasonable to fix in a fast follow-up if the author prefers.

### 4. [P1 as coverage gap] Zero concurrent-overlap test coverage for 3 of the 4 newly lock-wrapped mutators — `[reproduced]`

- **Trigger**: N/A (this is a test-coverage gap, not a shipped-code bug — the lock itself correctly protects these paths).
- **Evidence**: only `registerRepo` has a `Promise.all`-based concurrency regression test (`gitnexus/test/unit/repo-manager.test.ts:904-923`). `unregisterRepo` (`repo-manager.ts:1487-1488`), `removeBranchIndex` (`:1515-1516`), and `adoptFlatBranchLabel` (`:1602-1603`) have none. ce-testing **wrote and ran scratch tests** reproducing the exact interleavings (two `removeBranchIndex` calls racing on different branches of the same repo; `registerRepo` racing `unregisterRepo`) against both the pre-fix (lock reverted via `patch -p2 -R`) and post-fix worktree: pre-fix, both scratch tests fail (lost update / tmp-rename `ENOENT` collision); post-fix, both pass 3/3. This proves the fix is correct for these paths too, but also proves the shipped test suite would **not** catch a regression if the lock wrapping around these three functions were accidentally removed or narrowed later.
- **Severity**: P1, framed specifically as "the fix is likely correct but unverified by the test suite for 3 of 4 call sites" — not as broken shipped code.
- **Lanes**: claude-ce-testing (Finding 2, `[reproduced]` — scratch tests executed and reverted), gitnexus-test-ci-verifier (independently flagged the same gap as P2/non-blocking, `[code-read]`), claude-ce-correctness (testing-gaps section, `[code-read]`). Same-engine convergence, with one lane's claim empirically executed.
- **Suggested fix**: add `Promise.all`-based tests mirroring the pattern at line 904 for at least (a) concurrent `removeBranchIndex` on two branches of the same repo, and (b) `registerRepo` racing `unregisterRepo`/`removeBranchIndex`.
- **Blocks merge: yes** — this is the PR's own regression-coverage claim; three of the five newly-locked functions are unverified by the suite that's supposed to prove the fix.

### 5. [P1] `withRegistryLock` omits `log`/`onWaitStart` diagnostics — registry-lock contention is completely silent

- **Trigger**: any command routed through the registry lock (`list`, `clean`, `remove`, `index`, `analyze`'s registry calls, every MCP `resolveRepo`/`listRepos`, `gitnexus augment`) waits on a contended lock.
- **Evidence**: `withRegistryLock` (`repo-manager.ts:1142-1149`) calls `acquireIndexLock(getGlobalDir())` with no second argument, so `opts` defaults to `{}` and both callbacks are undefined — contrast with `run-analyze.ts:828-833`'s `acquireOpts`, which wires both `log` and `onWaitStart` for the per-repo lock, surfacing "Waiting for another analyze to finish on this index…" to the user.
- **Severity**: P1 (per ce-reliability RELI-4) — a user or agent hitting registry-lock contention sees nothing until (if ever) the 10-minute `IndexLockTimeoutError` fires.
- **Lanes**: claude-ce-reliability (RELI-4), claude-ce-adversarial (Finding 2, corroborating). `[code-read]`.
- **Suggested fix**: wire `log`/`onWaitStart` into `withRegistryLock`, matching the per-repo lock's UX.
- **Blocks merge: yes** — bundle with Finding 2's fix; a lock that can wait minutes should never be silent.

### 6. [P2] No try/catch around newly lock-gated calls in `gitnexus list` / `gitnexus index`; no top-level unhandledRejection handler for those commands

- **Trigger**: `IndexLockTimeoutError` (or any other error) thrown from `listRegisteredRepos({validate:true})` in `list.ts` or `registerRepo` in `index-repo.ts` under Node's default `--unhandled-rejections=throw` (package.json pins Node `^22.18.0 || >=24.11.0`).
- **Evidence**: `gitnexus/src/cli/list.ts:11` and `gitnexus/src/cli/index-repo.ts:138` call the now-lock-gated functions with no surrounding try/catch; only `analyze.ts` and `serve.ts` install fatal-rejection handlers.
- **Severity**: P2.
- **Lanes**: claude-ce-reliability (RELI-6). Single-lane. `[code-read]`.
- **Suggested fix**: wrap these call sites in the same `cliError()`-style handling used elsewhere (e.g. `remove.ts:112-116`).
- **Blocks merge: maybe** — worth cleaning up alongside Finding 2, but a raw stack trace on a rare timeout is a UX papercut, not data loss.

### 7. [P2] `IndexLockTimeoutError`'s hardcoded message misattributes registry-lock timeouts to "another gitnexus analyze"

- **Trigger**: a registry-lock timeout (e.g. from Finding 1 or 2) surfaces through `analyze.ts`'s existing friendly handler, which was written when the lock had only one caller.
- **Evidence**: `index-lock.ts:142-155` (constructor text) and `analyze.ts:1687-1700` (handler reused unmodified) — when `registerRepo` (Finding 1's call site) times out, the user is told their own repo's index is being refreshed by another analyze, when the real blocker may be unrelated registry contention.
- **Severity**: P2/P3 (risk-architect scored P3; reliability's RELI-5 scored P1 as a diagnosability multiplier on top of Finding 1/2 — synthesis lands P2 since it's a UX/diagnosability issue riding on top of already-listed findings, not independently severity-driving).
- **Lanes**: gitnexus-risk-architect (Finding 3), claude-ce-reliability (RELI-5). Same-engine convergence. `[code-read]`.
- **Suggested fix**: add a context-string parameter to `AcquireOptions` so the error message can name the actual operation (registry vs. analyze).
- **Blocks merge: no** — cosmetic misdiagnosis riding on top of Findings 1/2; fix opportunistically, not a gate.

## Repository history considered

- `acquireIndexLock` (the primitive this PR reuses unmodified) was introduced by PR #2677, closing issue #2658, merged 2026-07-25 — itself went through an 11-finding tri-review including a self-caught HIGH (a lock leak from a re-resolve loop outside try/finally). Its only prior call site before this PR was `run-analyze.ts`, scoped to a per-repo/branch index directory — this PR is the first reuse for a different purpose (global metadata coordination, not a write pipeline).
- `adoptFlatBranchLabel` (one of the five functions this PR now locks) was added by PR #2364 (2026-07-03); its "no subprocess is spawned" docblock claim and its `clean --branch` sibling pattern in `clean.ts` are both referenced in Findings 3 and the nits above.
- The `#829`/`#979`/`#2106`/`#2364` references embedded in the pre-existing JSDoc (now stranded on the private `*Unlocked` functions, see nits) describe real prior disambiguation and multi-branch design decisions this PR's renaming didn't move but also didn't break.
- No linked GitHub issue exists for the bug this PR fixes (`closingIssuesReferences` empty) — the PR body is the only description of the problem.

## Branch hygiene assessment

**Clean feature/fix PR**: single-purpose fix branch (`fix/registry-cross-process-lock`), 4 commits, changes confined to exactly 3 files (implementation + one regression test + one CI-matrix registration comment) — no unrelated churn, no mixed domains. Two commits share an identical message ("fix: serialize global registry mutations"); informational only, not a hygiene violation. The branch was rebased after the maintainer's `CHANGES_REQUESTED` review (see below) and CI is green on the current head.

## Back-and-forth avoided by verifying

Directly re-read against the live worktree this session (not taken on any single lane's word) rather than left for the PR author to answer: `getStoragePaths`/`getGlobalDir`'s exact equality condition (Finding 1); the `registerRepo`/`adoptFlatBranchLabel` call sites' try/catch coverage in `run-analyze.ts` (Finding 1); the `augment.ts` <500ms doc comment and its call chain into the locked path (Finding 2); the `clean.ts` vs. `adoptFlatBranchLabelUnlocked` fs.rm placement contrast (Finding 3); the new test's exact line range (Finding 4); and that none of the five `*Unlocked` functions call each other or a locked wrapper (ruling out simple same-file reentrancy, as distinct from the cross-module case in Finding 1). This should save the author a first round of "can you point me to the exact lines" replies.

## Open questions

- Is analyzing a repo rooted at `$HOME` (or any `GITNEXUS_HOME`-colliding path) an intentionally supported configuration, or already implicitly disallowed elsewhere in a way this review didn't find? If already disallowed, Finding 1 downgrades from a live bug to a latent trap; if not, it's live today independent of this PR (the per-repo lock alone would already misbehave against the global dir) and this PR sharpens it into a guaranteed self-deadlock.
- Does commit `f12cd7f6a` ("fix: keep registry reads lock-free") fully address the maintainer's empty-body `CHANGES_REQUESTED` review, or was there more to it? Only the maintainer can confirm.

## Final verdict

**Not production-ready.** The locking design's core mechanics (try/finally, no same-file reentrancy, no lock bypass, atomic-rename read safety) are sound and well-validated by the adversarial lane's refuted scenarios. But Finding 1 is a deterministic self-deadlock for a real (if narrow) repo layout with no existing guard, Finding 2 puts a documented <500ms editor-hook path behind a silent 10-minute-capable lock, and Finding 4 shows the regression suite that's supposed to prove this fix only actually covers 1 of the 5 newly-locked functions. Recommend fixing Findings 1, 2, 4, and 5 before merge; Finding 3 is a reasonable fast-follow; Findings 6-7 are opportunistic. Separately and regardless of this digest's content, the maintainer's own outstanding `CHANGES_REQUESTED` review below must be explicitly resolved.

## Outstanding maintainer review — important non-code context

**The PR currently carries an unresolved `CHANGES_REQUESTED` review from `magyargergo`** (`mergeStateStatus=BLOCKED`), left 2026-07-28 against an earlier commit (`29177135f`), body empty. Four commits have landed since, including `f12cd7f6a "fix: keep registry reads lock-free"`, which by name plausibly addresses the concern (the `listRegisteredRepos` validate-only locking behavior). **No re-approval or dismissal is recorded on GitHub.** This tri-review posts as a separate maintainer account (`abhigyanpatwari`), not `magyargergo` — **posting this digest does not resolve or substitute for `magyargergo`'s own pending review**; do not treat this digest as clearing the PR for merge. The outstanding review still stands and should be explicitly re-examined or dismissed by `magyargergo` before merge, independent of this automated digest's content.

## Lower-priority / nit items (P3, not inline-worthy)

- **JSDoc left on now-private `*Unlocked` functions instead of the exported public wrappers** — IDE hover on the actual public API (`registerRepo`, `unregisterRepo`, `removeBranchIndex`, `adoptFlatBranchLabel`, `listRegisteredRepos`) now shows nothing; all documentation lives on the unreachable-from-outside `*Unlocked` internals. Found by claude-ce-maintainability (Finding 1). `[code-read]`.
- **Self-referential PR-number comment**: `gitnexus/scripts/cross-platform-tests.ts` cites "the overlapping-registration regression from #2716" — #2716 is this PR itself, no separate issue exists. Ambiguous wording, could misread as "#2716 introduced a regression." Found by gitnexus-pr-facts-historian (V3) and independently by prior-context recall. `[code-read]`.
- **Redundant `async` keyword** on 3 of the 5 thin wrapper functions (`async (...) => withRegistryLock(...)`) — harmless, TS/JS-legal. Found by gitnexus-pr-facts-historian (V4).
- **`withRegistryLock`'s docblock doesn't disclose that the critical section can include a synchronous `git` subprocess call** (`registerRepoUnlocked`'s common path shells out via `execSync` for `git config`/`git rev-parse` while holding the global lock) — contrast with `adoptFlatBranchLabelUnlocked`'s docblock, which does call out "no subprocess is spawned." Found by claude-ce-maintainability (Finding 2). Related to Finding 3's contention concerns but framed as a documentation gap.
- **Two commits with the identical message** "fix: serialize global registry mutations" (`bf42c4341`, `c7a3e15dd`) in the PR history — no repo convention violated, informational only.
- **New concurrency test placed inside an unrelated `describe` block** ("registerRepo name override + collision guard (#829)") whose name doesn't describe lock-serialization behavior — cosmetic, no rule violated.

## CI status

All required checks pass: typecheck, lint, format, CodeQL, full Ubuntu/Windows/macOS coverage matrix, tree-sitter ABI, node-floor compatibility (15,937 tests, 0 failed, 70 pre-existing skips). **Only Vercel fails**, and that is the fork-authorization/auth-only deployment gate, unrelated to code correctness.

## Not flagged / explicitly cleared

- No lock bypass, no torn reads, no new crash-corruption window (adversarial lane, refuted scenarios above).
- The underlying `acquireIndexLock` primitive itself is not modified or re-tested by this PR (correctly out of scope) — its own dedicated test suites (`index-lock.test.ts`, `analyze-index-lock-concurrency.test.ts`) are unaffected.
- Project-standards review found zero violations of AGENTS.md/CLAUDE.md/GUARDRAILS.md; diff is minimal and scoped to exactly the files required.
- The pre-existing macOS/BSD file-backend dead-holder reclaim race (accepted upstream, not introduced by this PR) now operates under higher call frequency (registry lock is exercised on every `augment` hook call, not just infrequent `analyze` runs) — this raises the odds of hitting an existing, accepted race window; it is not a new bug class. Noted in prior-context and independently corroborated by ce-reliability's residual risks.

---

*This is an automated multi-lane digest (Claude-only this run — Codex was unavailable due to a workspace spend cap) synthesizing 9 independent review lanes. Verify findings against the actual code before acting on them; in particular, confirm the outstanding maintainer review above before treating this PR as merge-ready.*

---

azizur100389:
Re-reviewed current head `14fb1837` against `upstream/main` `51095c19` in a fresh worktree because this head changed after my earlier comment. It fast-forwarded cleanly; `git diff --check` passed; no hidden bidi controls were found. The new overlapping-registration regression passed and `npx tsc --noEmit` passed. The full repo-manager file produced 104 passes, 2 skips, and one known Windows-local `EPERM` symlink-creation failure unrelated to this diff. CI is green apart from Vercel authorization, but the PR remains `CHANGES_REQUESTED`/blocked.

Branch hygiene: `clean feature/fix PR`. Merge state: `review blocked`.

Findings:

- **Risk:** deterministic same-process self-contention when the analyzed repository's storage directory equals the global registry directory. With the default layout, analyzing a repository rooted at the user's home directory makes both locks target `~/.gitnexus`; `runFullAnalysis` already holds the per-repo lock when `registerRepo` now attempts to acquire the same non-reentrant lock again.
- **Evidence to check:** `run-analyze.ts` holds `acquireIndexLock(writeTarget.metaDir)` across `registerRepo`; `withRegistryLock()` in `repo-manager.ts` calls `acquireIndexLock(getGlobalDir())`; for `repoPath === os.homedir()`, both resolve to the same directory.
- **Recommended fix:** move registry finalization after release of the per-repo lock, use a non-colliding registry lock namespace, or explicitly handle identical lock ownership without reacquisition. Add a regression for the colliding layout.
- **Blocks merge:** yes.

- **Risk:** short registry operations inherit the index lock's 10-minute analyze timeout with no wait diagnostics. `gitnexus augment` calls `listRegisteredRepos({ validate: true })`, so editor hooks documented to cold-start quickly can silently block behind unrelated registry work for minutes.
- **Evidence to check:** `withRegistryLock()` passes no timeout/log callbacks; `core/augmentation/engine.ts` calls validating `listRegisteredRepos`; `augment.ts` documents a sub-500ms budget.
- **Recommended fix:** give registry locking a short independent timeout and wire `onWaitStart`/logging with registry-specific diagnostics.
- **Blocks merge:** yes.

- **Risk:** the regression suite proves concurrent `registerRepo` calls but not the other newly serialized mutators, including mixed register/remove and concurrent branch-summary removal.
- **Evidence to check:** only the new `Promise.all` registration test covers overlap; `unregisterRepo`, `removeBranchIndex`, and `adoptFlatBranchLabel` have no equivalent overlap tests.
- **Recommended fix:** add focused concurrent tests for at least register-vs-unregister/remove and two branch removals on the same registry entry.
- **Blocks merge:** yes.

Final verdict: `not production-ready`. The lock correctly closes the demonstrated lost-update window, but its namespace, timeout, and call-path integration introduce deterministic and high-impact blocking behavior. Resolve those issues and the outstanding requested-changes review before merge.

gitnexus/test/unit/repo-manager.test.ts
magyargergo: Covered. `test/unit/repo-manager.test.ts` is listed in `PLATFORM_LOGIC` in `gitnexus/scripts/cross-platform-tests.ts`, so this regression — and the three tests added in this round (two more overlap cases plus a home-rooted lock-collision regression) — run on the Windows/macOS/Linux matrix, exercising all three index-lock backends. `14fb1837` added the comment recording why that entry matters for the registry lock specifically.

gitnexus/src/storage/repo-manager.ts
magyargergo: Fixed in `ff78aaa` by giving the registry its own lock namespace:

```ts
const getRegistryLockDir = (): string => path.join(getGlobalDir(), 'registry-lock');
```

An index slot's lock dir is always `<repo>/.gitnexus` or `<repo>/.gitnexus/branches/<slug>`, so no repository can ever resolve to `<globalDir>/registry-lock` — the nested acquisition inside `runFullAnalysis`'s per-repo lock can no longer collide, for `repoPath === os.homedir()` or any `GITNEXUS_HOME` value. `registerRepo` and `adoptFlatBranchLabel` stay where they are; the docstring's "post-analysis steps run lock-free" claim is about the per-repo index lock, which they no longer take.

Regression test added (`registers while an index lock is held on the global directory (#2716)`): it holds `acquireIndexLock(getGlobalDir())` — exactly the layout you describe — and asserts the registration contends with nothing, i.e. neither the wait announcement nor the degraded-write warning is logged. Reverting `getRegistryLockDir()` to `getGlobalDir()` makes it fail on both records, so it is anchored on the change rather than passing vacuously.

gitnexus/src/storage/repo-manager.ts
magyargergo: Fixed in `ff78aaa`. `withRegistryLock` now passes its own options instead of inheriting analyze-scale defaults:

- `timeoutMs: REGISTRY_LOCK_TIMEOUT_MS` (5s), independent of `GITNEXUS_INDEX_LOCK_TIMEOUT_MS`, so the `augment` → `findRepoForCwd` → `listRegisteredRepos({ validate: true })` path can no longer sit behind the 10-minute ceiling.
- `onWaitStart` wired with a registry-specific line. `log` is deliberately left unwired: the primitive's own texts name "another gitnexus analyze" as the holder, which misattributes a registry wait (that is also the misattribution flagged separately in the digest as finding 7 — the registry path no longer surfaces `IndexLockTimeoutError` at all).
- On timeout the transaction proceeds **unlocked** with a `logger.warn`, rather than throwing. The lost-update race this lock closes was unguarded before this branch, so degrading to the previous best-effort behaviour beats failing an `analyze`/`list`/`index` on a wedged lock — and `list.ts` / `index-repo.ts` have no handler for a raw rejection, so this also closes the reliability finding about those call sites.

gitnexus/src/storage/repo-manager.ts
magyargergo: Fixed in `ff78aaa`. `adoptFlatBranchLabel` is no longer a fully-wrapped `*Unlocked` function: the membership gate and the recursive `fs.rm` run outside the lock, and only the closing re-read/mutate/write is wrapped —

```ts
await withRegistryLock(async () => {
  const entries = await readRegistry();
  ...
  await writeRegistry(entries);
});
```

which is the same split `clean.ts` uses (rm first, then the locked `removeBranchIndex`). A slow or stuck delete now blocks nothing but itself. The re-read-before-write discipline from #2106 R9 is unchanged — it just happens under the lock now.

gitnexus/test/unit/repo-manager.test.ts
magyargergo: Added two more `Promise.all` overlap tests, matching your (a)/(b):

- `overlapping removeBranchIndex calls drop both summaries (#2716)` — two branches of one entry.
- `keeps an overlapping unregisterRepo and registerRepo from clobbering each other` — a removal racing a distinct registration, which is deterministic in its expected end state (unlike register-vs-unregister on the same path).

Plus `registers while an index lock is held on the global directory (#2716)` for the lock-namespace collision.

Verified they are not vacuous: bypassing `withRegistryLock` makes all three overlap tests (including the original one) fail with the expected lost update, and reverting the lock namespace makes the collision test fail on the wait announcement it then records. `adoptFlatBranchLabel` deliberately keeps its slow `fs.rm` outside the lock now, so its own concurrent-overlap contract is the registry write it shares with these two.

magyargergo:
Pushed three commits to this branch addressing every open review comment (`ff78aaa`, `70e347d`, `37b9b25`). Per-comment replies are inline; summary:

| Finding | Where | Fix |
| --- | --- | --- |
| P1 self-deadlock when the analyzed repo's slot equals `getGlobalDir()` | @abhigyanpatwari, @azizur100389 | Registry locks a private `<globalDir>/registry-lock` namespace. No index slot — `<repo>/.gitnexus` or `.gitnexus/branches/<slug>` — can resolve to it, so the nested acquisition inside `runFullAnalysis`'s per-repo lock can no longer collide. |
| P1 10-minute analyze-scale timeout on a sub-second transaction, gating the sub-500ms `augment` path | @abhigyanpatwari, @azizur100389 | Own `REGISTRY_LOCK_TIMEOUT_MS` (5s), independent of `GITNEXUS_INDEX_LOCK_TIMEOUT_MS`. On timeout the transaction proceeds unlocked with a warning instead of throwing — the race it guards was unguarded before this branch, and `list`/`index` have no handler for a raw rejection (also closes the "no try/catch at those call sites" and "`another gitnexus analyze` misattribution" findings). |
| P1 silent contention (no `log`/`onWaitStart`) | @abhigyanpatwari | `onWaitStart` wired with a registry-specific message; `log` deliberately left unwired because the primitive's texts name an analyze holder. |
| P2 recursive `fs.rm` inside the lock in `adoptFlatBranchLabel` | @abhigyanpatwari | Only the closing re-read/mutate/write is locked now, matching `clean.ts`'s split. |
| P1 coverage gap — 3 of 5 locked mutators had no overlap test | @abhigyanpatwari, @azizur100389 | Added overlapping `removeBranchIndex` on two branches, an overlapping `unregisterRepo`/`registerRepo` pair, and a regression for the colliding layout (holds an index lock on the global dir, asserts the registration contends with nothing). Each was verified to fail against the un-fixed code. |
| Cross-platform coverage | @magyargergo | `test/unit/repo-manager.test.ts` is already in `PLATFORM_LOGIC` in `scripts/cross-platform-tests.ts`, so all of these run on the Windows/macOS/Linux matrix and exercise both lock backends. |
| Prettier failure | @azizur100389 | `prettier --check` is clean on the changed files. |

One extra change while in here: `listRegisteredRepos({ validate: true })` no longer holds the lock across its read-only validation walk (an `fs.access` per entry). The walk runs unlocked and the lock is taken only when something is provably gone, with the prune applied to a snapshot re-read inside it — same narrowing as `adoptFlatBranchLabel`, and it matters because that walk is exactly the `augment` hot path.

Local validation from `gitnexus/`: `tsc --noEmit` clean; `vitest run test/unit/repo-manager.test.ts` 108 passed / 2 skipped; the registry, index-lock, `GITNEXUS_HOME`-roots, `calltool-dispatch` and `remove-command` suites pass. The remaining local failures in my sandbox are parse-worker startup timeouts in CLI-subprocess suites (`Replacement worker did not report ready within 5000ms`), unrelated to this diff — CI is authoritative there.

Deliberately not changed: the shared `acquireIndexLock` primitive. The timeout, the degrade-on-timeout policy and the lock granularity are all caller policy — `run-analyze.ts` needs the opposite (fail closed), so pushing any of it into the primitive would give its two call sites a way to be misconfigured into unlocked analyze runs.

Issue #1919, opened 2026-05-30
Tracking: AST/tree-sitter parsing-layer coverage audit — 92 verified findings (16 languages)


magyargergo:
> **Status: independently verified.** Produced by a 34-agent `ast-language-audit` workflow, then **independently re-verified** by a separate 16-agent pass (one per language) checking every claim against the grammar's actual `node-types.json`, the source at `main`, and merged-PR/issue history.
>
> **Finding IDs use an `F` prefix (`F1`…`F92`) on purpose** — bare `#N` would auto-link to unrelated GitHub issues/PRs. Real PR/issue references below are written as plain text (e.g. `PR 789`) for the same reason.

## Verification result

| | |
|---|---|
| Findings | **92** across 16 languages |
| ✅ Confirmed | **91** |
| 🟡 Partial | **1** (F50 — content correct, cited line off by one) |
| ❌ Refuted / 🔵 Already-fixed / ⚪ Stale | **0 / 0 / 0** |
| Confidence | all high |

No finding is already fixed. Where a merged PR was close, it either *introduced* the partial coverage (Dart PR 801, Vue PR 604) or fixed a **sibling language only** (C# interface heritage PR 789; C++ EXTENDS PR 1581/1595; heritage-framework PR 890 ships only `go.ts`/`ruby.ts`) — corroborating the audit's core thesis: **grammar-shape coverage has been patched one language at a time and never generalized.** EPIC 1890 tracks the cpp theme only; issues 1646 (JS) / 1388 (Rust) track symptoms, not these gaps.

## Dominant systemic patterns
1. **Heritage queries miss qualified / generic / scoped / interface supertypes** in nearly every OO language (Java F36/F37, TS F81/F82, C# F10/F11, Go F31, Python F57, Ruby F63, Rust F69, Kotlin F46, JS F43, C++ F6).
2. **Legacy `tree-sitter-queries.ts` banks lag the registry-primary `query.ts`** (parity-gate risk): TS F81–F86, Java F39/F40, JS F42–F44, C++ F6.
3. **Dead branches keyed on node types the installed grammar never emits** — wrong grammar models copied from sibling grammars: Rust F67/F74, C# F15, C++ F7/F8, PHP F56, Dart F27, Ruby F64/F65, Kotlin F50, Swift F80, Python F59/F60.
4. **Anonymous / generator / lambda / class-expression gaps** recur across TS/JS/Python: JS F42/F44, TS F84/F86, Python F61.

[...]

PR #2718, opened 2026-07-27
fix(scope-resolution): a closure binding is a call SOURCE in every language, and function-local values carry their own identity (closes #2699)

### LINKED ISSUE #2699: Function-local symbols share file-scoped node ids, so a local collides with a same-named file-level symbol (SCIP uses `local N`)

magyargergo:
Closes #2699.

A closure bound to a name is now a call **SOURCE**, not only a TARGET, in every supported language — and function-local **values** carry their own identity. `impact(handler, direction:"downstream")` returns something.

## Part A — the consumer audit (issue item 4)

`resolveDefGraphId` is **CRITICAL** with 23 direct dependents across 7 modules (every language MRO builder, both Spring attachers, C++ member lookup, `tryEmitEdge`, `emitReferencesViaLookup`, `buildGraphTargetIndex`, `emitFreeCallFallback`, `emitReceiverBoundCalls`, `preEmitInheritanceEdges`, `emitDetectedInterfaceImplementations`, `phpEmitUnresolvedReceiverEdges`, `emitRubyMixinEdges`, `emitRustTraitImplEdges`, `emitDartHeritageEdges`). Auditing that key chain and rewriting it in the same breath would have been the opposite of an audit, so Part A is findings-and-tests only.

| Consumer | Verdict |
| --- | --- |
| `detect_changes` | **safe** — maps hunks to symbols by line-range overlap, never by id; pinned by a new structural test |
| MCP `impact` / `context` / `trace` | **safe** — resolve by name/uid at query time |
| bench fingerprints (8 gates) | **safe** — digest capture shape, not node ids |
| `rust-captures-golden` | **safe** — digests captures |
| `cfg` pipeline-pdg snapshot | **at risk by design** — pins exact edge ids, so it trips whenever attribution changes. That is the gate working. |
| wiki / group-contract links | **not id-keyed on locals** — locals are never cross-file addressable |
| **incremental writeback** | **safe** — verified end-to-end (below) |

The incremental path was the one consumer left unproven, and it is now measured. A function-local callable was moved down one line and a real incremental analyze run (`changed=1, added=0, skipping wipe + 1 unchanged file rows preserved`): the id went `run.save@1:2` → `run.save@2:2`, edges stayed at 12 with every one re-pointed, exactly one `save@` node survived, and the untouched file's rows were preserved. `deleteNodesForFiles` deletes by `n.filePath IN [...]` — **file-keyed, never id-keyed**. Embedding rows are the one genuinely id-keyed store, and both their delete (joined through nodes while the old id still exists) and their restore are file-scoped.

## Part B — closure caller attribution

Diagnosed by instrumenting `pickCallerCallableDef` to log every rejection reason, rather than by inference:

| Language | Closure scope | Kind | Owns callable def | Verdict |
| --- | --- | --- | --- | --- |
| TS/JS, Python, Go | yes | Function | **yes** | already worked |
| PHP | yes | Function | no | `REJECT-NO-CALLABLE-DEF` |
| Rust | yes | Function | no | `REJECT-NO-CALLABLE-DEF` |
| Kotlin | yes | **Block** | no | `REJECT-KIND` |
| Ruby | yes | **Block** | no | `REJECT-KIND` |
| Dart | **none** | — | — | loop never entered |

**Every failing language had `owned=[]`.** So the received framing — "Kotlin/Ruby fail the kind gate, PHP fails ownership" — was true but incomplete in a way that would have misled the fix: relaxing only the kind gate moves Kotlin and Ruby from `REJECT-KIND` straight to `REJECT-NO-CALLABLE-DEF`. The cause was **one missing query rule**: `javascript/query.ts` pairs the binding name with the closure via `@declaration.function` anchored on the **inner** node, so `anchor.range` equals the `@scope.function` range and the declaration attaches to the closure's own scope. No other language had it.

Landed per language, each verified independently:

- **PHP** — the query rule alone (it already declared both closure forms as `@scope.function`).
- **Rust** — the query rule **plus** a graph-node rule; Rust emitted no node for `let f = || …` at all, making it the real exception to #2687.
- **Kotlin, Ruby** — the query rule **plus** a Block-scope callable boundary. Their `lambda_literal`/`do_block` stays `@scope.block` deliberately (#1757), so the scope must not be re-kinded. Accepting *any* Block owning a callable would be wrong — a nested `fun foo()` inside a block would swallow block-level calls — so the discriminator is that the scope **is** the callable's body (def start == scope start).
- **Dart** — a synthesized closure scope **plus** enclosing-callable identity. Dart parses `int outer() { … }` as `function_signature` followed by `function_body` as *siblings*, so the enclosing callable is never an ancestor and the ancestor walk could not reach it. That is a semantic relation being derived syntactically — what SCIP avoids by keeping locals opaque and modelling containment as a separate `enclosing_symbol` field, and what Dart's own analyzer avoids via `Element.enclosingElement`.

## A1 — function-local values

The issue's **original** complaint, unreachable from a callable-only gate: a top-level `const handler` and a function-local `const handler` collapsed onto one `Const:v.ts:handler`. Widened via `isPositionQualifiedLocalLabel`, which is the **single** definition shared by all three phases that must agree — id-building, resolution, and registration. Two of them disagreeing does not fail loudly; the caller attaches to a node that does not exist and the edge is silently dropped (the #2714 failure mode).

## Review findings — fixed in this PR

This branch was reviewed by two engines running the same scaffolding (Claude swarm + ce personas; Codex `gpt-5.6-sol` swarm + ce + adversarial). It found **four P1 defects in the first cut**, three of them via an independent engine disagreeing with the authoring one. All are fixed and pinned in `closure-review-findings.test.ts`:

- **Multi-line closure bindings emitted a fabricated CALLS edge.** The two channels anchor on different nodes by design; the position join is line-only, so a binding whose literal starts on the next line missed, and caller resolution **climbed to the enclosing function** — crediting a call to a function that does not make it, the exact defect class this PR removes. The bridge now **fails closed** at the owning callable.
- **TypeScript constructor parameter properties were re-keyed as constructor-locals.** Admitting `Property` made the walk reach the constructor through the *parameter list* before any class boundary, so `Service.port` became `Service.constructor.port@r:c` — emptying the slot `impact`, `rename` and FTS address, for the Angular/NestJS DI idiom. Now reuses the parameter-list exemption that already existed two lines away.
- **Dart top-level `var`/`final` and later declarators** were never sources — the rule matched only Dart's function-local shape, though `captures.ts` already listed all three for callable-flow.
- **Ruby `do … end` and `Proc.new`** were uncovered — the dominant multi-line style — and needed both channels, exactly as Rust did.

Also fixed from the same review: the split-signature fallback could fire on **valid** TypeScript (`declare namespace` with a bodyless overload); a new twin list (`isCallableDef` re-spelling `isOverloadableCallable` in the same file that imports it); a self-contradictory comment; a `learnings.jsonl` entry that violated `gitnexus-plan`'s chat-only feedback contract; and an unconstrained receiver that captured `MyMod.lambda { }` as a closure binding.

## Invalidation

`INCREMENTAL_SCHEMA_VERSION` 20 → 21 and parse-cache `SCHEMA_BUMP` 27 → **29**. 29, not 28, because #2415 shipped 28 first — caught only by re-checking against `origin/main` at **merge** time. Leaving it at 28 would have shipped the whole change with no parse-cache invalidation. This is the third recorded instance of that collision class in `parse-cache.ts`.

## Verification

- Full resolver sweep (the gate that caught #2714's Rust regression) — **2926 passed / 0 failed**, 51 files.
- All **8** bench fingerprint gates PASS with fingerprints unchanged; `rust-captures-golden` unchanged, so no regeneration.
- Wall/RSS measured against the pre-Part-B base on a 762-file TS corpus: **no regression** (55.6 s base vs 54.2–55.1 s head; RSS brackets each other).
- Pinned wrong-behaviour tests were **inverted, not deleted**, per their own instructions. Ruby, the Dart collision, and all four review findings got new tests; Rust's previously untested closure work is now covered.


magyargergo (commented):
# PR #2718 — multi-engine review digest (#2699 parts A+B)

## 1. Review bar for this PR

Read-only multi-engine review of [PR #2718](https://github.com/abhigyanpatwari/GitNexus/pull/2718)
against `AGENTS.md`, `GUARDRAILS.md`, `DoD.md`, `CONTRIBUTING.md`, `TESTING.md` and
`ARCHITECTURE.md`, in worktree `/workspace/.worktrees/pr-2718`, base
`ff86ccf1e79cd7e4175da437ae8aeaf67b64aaa1`, head `cfd2434c`.

Acceptance for this PR requires:

- A closure bound to a name is a call **SOURCE** in every language the PR claims — across the
  ordinary formatting and binding shapes of each language, not only the shape the tests use.
- Function-local identity is collision-free **without fabricating caller edges**; the def→graph
  join either lands on the right node or fails closed.
- Shared pipeline code stays language-agnostic (AGENTS.md rule for `core/ingestion/`).
- Cache/schema versions invalidate stale data.
- Worker-path tests that actually discriminate the failure shapes.
- Green required CI; accurate comments and PR description; clean diff, no hidden Unicode.

**Independence disclosure, stated up front because it changes how this report should be read: the
PR author and the review coordinator are the same agent.** Cross-engine (Codex) disagreement is
therefore the highest-value signal in this review — and Codex disagreed with the optimistic Claude
lanes on both severity and consequence, and was right both times.

## 2. Problem being solved

The PR completes parts **A + B** of [issue #2699](https://github.com/abhigyanpatwari/GitNexus/issues/2699):

- **Part A** — audit every consumer of file-scoped node ids and give function-local
  values/callables a **positional** identity, so a top-level `handler` and a function-local
  `handler` stop collapsing onto one node.
- **Part B** — make closure bindings valid CALLS **sources** in PHP, Rust, Kotlin, Ruby and Dart
  (previously they could be call *targets* only).

## 3. Current PR state

- Open, non-draft. Author: `magyargergo`.
- Base: `main` @ `ff86ccf1e79cd7e4175da437ae8aeaf67b64aaa1`; head: `fix/2699-followup` @ `cfd2434c`.
- **23 files, +696 / −45, 8 commits** (S1 PHP → S3 Rust → S2 Kotlin+Ruby → S4 Dart → merge-from-main
  → A1 values + twin-list guard + schema bumps → two CI-driven test fixes).
- No reviews, no labels, no review decision, **no closing-issue link** (#2699 is not referenced as
  closed by this PR).
- **The live PR body is stale**: its "Still open" section says A1 values and the receiver twin-list
  guard are deferred, but commit `da3d8397` implements both. Flagged independently by both engines.

**Merge-state classification: checks are COMPLETE, with only the known-unfixable Vercel fork check
failing; `mergeStateStatus` is `UNSTABLE` solely because of it.**

Observed **first-hand by the coordinator**, querying GitHub directly against head `cfd2434c` at
synthesis time (this **supersedes** the lane-reported "`BLOCKED` / aggregate CI Gate pending", which
was captured earlier in the run and is stale):

- `mergeable: MERGEABLE` — the branch is conflict-free against the base.
- `mergeStateStatus: UNSTABLE`.
- **38 checks PASSED · 0 pending or queued · 1 failing: `Vercel`.**
- The `Vercel` failure is the **known fork-PR authorization failure with no PR-side fix** — it is
  not a signal about this diff and cannot be resolved by changing this branch.
- The two **coverage-shard failures that were red on the previous head (`da3d8397`) are RESOLVED on
  `cfd2434c`** — that is what the final two test-fix commits addressed.

So there is nothing left to wait for: CI has fully reported, everything that this PR can influence
is green, and `UNSTABLE` is entirely attributable to the fork-authorization check. **Merge state is
not a blocker for this PR; the findings in section 8 are.**

- **#2695** introduced closure-value calls and the position join this PR widens.
- **#2714** added named-receiver handling and the id-phase/resolution-phase lockstep discipline that
  `isPositionQualifiedLocalLabel` now embodies — its failure mode ("silently drops edges rather than
  failing") is quoted in the new code and is the reason several findings below are rated as they are.
- **#2687** established "a closure bound to a name is a Function node in every language", the
  contract Rust was the exception to.
- **#1757** made Kotlin `lambda_literal` deliberately `@scope.block`; this PR does not re-kind it
  (correct) but does newly accept a Block scope as a caller anchor.
- **#2415** collided with this branch on `parse-cache.ts` `SCHEMA_BUMP` (both claimed 28); caught at
  rebase time and corrected to 29 — the **third** recorded instance of that collision class.
- **#1919** produced the `LOCAL_SCOPE_BODY_NODE_TYPES` carve-outs whose docblock P2-1 contradicts.
- Prior **off-by-one incidents** and the explicit `NODE_TO_CFG_LINE_OFFSET = 1` bridge were checked
  deliberately, because this PR bridges 1-based def ranges to 0-based graph-node lines (see §10.1).
- **Stale-JSDoc precedent in this exact subsystem** (Kotlin `scope-resolver.ts`) drove a full
  fact-check of every new/edited comment; three came back false or self-contradicting (P2-1, P2-5,
  and the callable-labels docblock under P1-2).

**Branch-hygiene classification: merge-from-main commit present but harmless and merge-safe.**

Commit **`6fd9af17`** ("Merge branch 'main' into fix/2699-followup") is a merge-from-main. Verified
by the coordinator: `git rev-list --parents -n1 6fd9af17` yields parents
`89c896da` and **`ff86ccf1e79cd7e4175da437ae8aeaf67b64aaa1`** — its second parent is **exactly the
review base**. It is a pure synchronization with the base, introduces no third-party history, and
requires **no rebase and no split**.

Supporting hygiene checks, all clean:

- `git diff --check` clean; no bidi/RTL controls; non-ASCII limited to box-drawing characters and em
  dashes inside comments.
- No workflow / CI / Docker / YAML files touched. No `any` / `as any`.
- Test placement matches `TESTING.md`.
- Production, parser, cache and regression-test changes are all causally connected to #2699. The
  only ancillary churn is the four-line `eval/workflow_bench/learnings.jsonl` append — minor, and
  does not justify a split (though one of its four lines is a rule violation; see P2-6).

The PR:

- **Broadens local positional identity** from overloadable callables to values and properties —
  `isPositionQualifiedLocalLabel` now returns true for `Function|Method|Constructor` **plus
  `Variable`, `Const`, `Property`, `Static`** (`callable-labels.ts:42-50`), and that one predicate is
  consumed by **both** the id-building phase (`parse-worker.ts:2310`) and the resolution phase
  (`ids.ts:212`) so the two cannot drift.
- **Adds def→graph lookup guards**: a position-key-first lookup with an `AMBIGUOUS_POSITION`
  tombstone and a **fail-closed** `localNameKey` gate (`ids.ts:211-229`), plus matching schema
  invalidation (`INCREMENTAL_SCHEMA_VERSION` 20→21, `parse-cache.ts` `SCHEMA_BUMP` →29).
- **Adds closure-binding declaration rules** for PHP, Rust, Kotlin, Ruby and Dart in the
  scope-resolution query channel.
- **Permits an aligned `Block` scope to act as a callable boundary** in `pickCallerCallableDef`
  (`ids.ts:117-135`), discriminated from a nested named function by a start-position alignment test.
- **Handles Dart's split signature/body AST** via the new `findSplitBodyCallableAncestor` +
  `SPLIT_SIGNATURE_NODE_TYPES` (derived, not listed) in `ast-helpers.ts`.
- **Adds tests**: consumer audit, callable-id lockstep, cache-version pins, caller attribution,
  function-local identity collisions, and a receiver twin-list drift guard.

GitNexus reports **HIGH aggregate risk**: 26 changed symbols, 10 affected processes, 23 files.
`resolveDefGraphId` is **CRITICAL** — 24 direct and 66 total upstream dependents; caller attribution,
MRO, receiver-bound calls, inheritance and reference emission all consume this bridge.

## 8. Findings

Evidence tiers used below, kept deliberately distinct:

- **[analyzer-verified]** — the coordinator ran the **built analyzer CLI** (`analyze --index-only`)
  over a purpose-built fixture and queried the resulting graph with `cypher`. End-to-end; the
  strongest tier in this review. Applies to **P1-1 and P1-2 only**.
- **[parser-probe]** — the coordinator (and, where noted, a Claude lane) ran the **PR-head
  tree-sitter queries against the real vendored grammars** out of tree. This proves what the query
  matches; it does **not** exercise scope synthesis, the graph bridge or edge emission, so it is a
  **weaker** tier. Applies to **P1-3 and P1-4**.
- **[code-read]** — source correlation only; not executed.

### P1-1 · Multi-line closure bindings emit a FALSE caller edge (all five changed languages)

- **Inline anchors (both inside diff hunks):** `graph-bridge/ids.ts:212` (the widened position-key
  gate) and `php/query.ts:99` (the new PHP closure declaration rule).
- **Unchanged context cited for reading, NOT in the diff:** the position join at `ids.ts:214`, the
  self-acknowledged multiline miss at `:220`, the fail-closed at `:226`, the caller climb at `:354`.
- **Per-language anchor split (the structural cause):** inner-anchored scope-resolution rules
  (`php/query.ts:99-105`, `rust/query.ts:67-79`, same shape for kotlin/ruby/dart) vs outer-anchored
  graph rules (`tree-sitter-queries.ts:1512-1517` PHP, `:1346-1348` Rust).
- **Trigger** (any of the five languages; PHP shown):
  ```php
  function outer() {
      $single = function ($x) { return target($x); };
      $multi =
          function ($x) { return target($x); };
  }
  ```
- **Evidence — [analyzer-verified].** Coordinator ran the built analyzer CLI (`analyze
  --index-only`) on that fixture and queried the graph with `cypher`: both nodes exist
  (`outer.$single@3:2`, `outer.$multi@4:2`); edges are `outer.$single@3:2 -> target` (correct)
  **and `outer -> target` (FALSE — `outer` calls nothing in the source)**; `outer.$multi@4:2` has
  zero outgoing edges. The two channels anchor different nodes by design; the position key carries
  only the **line**, so a wrapped initializer misses the join, and caller resolution climbs to the
  parent scope instead of stopping.
- **Corroboration: STRONG, cross-engine, with a settled disagreement.** Codex swarm R1 (P1, conf 99)
  predicted a **false edge**; Claude risk-architect F1 predicted a **dropped edge** and rated it P3;
  Claude ce-adversarial F3 predicted "resolves to nothing" and rated it P2; Claude ce-correctness
  logged it as residual risk #4. The end-to-end run shows **Codex was correct on both the consequence
  and the severity** — attribution climbs and a fabricated edge is emitted. This is the single most
  valuable result of the review, and it lands against the engine the coordinator shares an author
  with.
- **Fix:** share one def→graph identity between the inner and outer captures (or add an
  initializer-position bridge, or make the position key column-or-range-based); and when a callable
  boundary is found but cannot be bridged, **fail closed at that boundary** rather than climbing to
  the enclosing caller. Add multi-line local-closure SOURCE tests for all five languages.

### P1-2 · TypeScript constructor parameter properties are re-keyed as constructor-locals (regression)

- **Inline anchors (both inside diff hunks):** `workers/parse-worker.ts:2310` (the widened gate —
  `isPositionQualifiedLocalLabel(nodeLabel) && definitionNode`) and `utils/callable-labels.ts:42`
  (the predicate that now admits `Property`).
- **Unchanged context cited, NOT in the diff:** `parse-worker.ts:2309` (`const nestedCallablePrefix =`).
- **Trigger:** `class UserService { constructor(private readonly repo: UserRepo) {} }` — the
  Angular/NestJS/Inversify DI idiom, and a shape that **already exists in this repo**
  (`core/group/service.ts:304`, `cfg/visitors/call-site-harvest.ts:82`, `rust-harvest.ts:183`,
  `ruby-harvest.ts:157`, `config-parser.ts:115`).
- **Evidence — [analyzer-verified].** Coordinator ran the built analyzer CLI on a two-line fixture
  and read the node back via `cypher`: the id becomes
  `Property:…/svc.ts:Service.constructor.port@2:14`, where base `ff86ccf1` produced
  `Property:…:Service.port`. The `@definition.property` anchor is the `required_parameter` node; the
  prefix walk reaches `method_definition` (in `LOCAL_SCOPE_BODY_NODE_TYPES`, **not** in
  `CALLABLE_PREFIX_BOUNDARY_TYPES`) **before** the `class_body` boundary, and the
  `nestedCallablePrefix` branch is evaluated before `enclosingClassInfo`, discarding the class
  qualification. The file already models the correct exemption 60 lines earlier —
  `isFunctionLocalProperty` at `parse-worker.ts:2245-2257` uses `PARAMETER_LIST_NODE_TYPES` for
  exactly this carve-out — and the new ternary throws it away. `callable-labels.ts:34-36`'s docblock
  asserts the opposite of what now happens ("A class field stays unqualified even inside a
  function, because the prefix walk boundaries on class-likes").
- **Blast radius:** the class still asserts `HAS_PROPERTY` to a node whose qualified name says it is
  a constructor-local; `Property` is **not** removed by `local-symbol-pruner.ts` (which prunes only
  `Const|Variable|Static` with `scope === 'block'`), so these nodes ship into the graph and into MCP
  `context` / `impact` / `query` output; the `qualifiedKey(file,'Property','UserService.repo')` slot
  is now empty, so `impact({target:"UserService.repo"})`, `rename` and FTS-over-qualifiedName miss;
  and the now-`@row:col` id arms the `localNameKey` fail-closed at `ids.ts:226` for any other
  `Property` named `repo` in the same file.
- **Corroboration: STRONG (coordinator reproduced end-to-end), but single-engine.** Claude
  ce-correctness F1 (P2, conf 85) and Claude ce-adversarial F1 (P1, conf 90) agree; **Codex did not
  probe this path**. Note the **intra-Claude contradiction**: the maintainability lane fact-checked
  the same docblock and marked the class-field claim TRUE. The end-to-end run settles it — the
  docblock is false for this one shape.
- **Why CI missed it:** `test/integration/resolvers/typescript.test.ts:2088-2121` (the
  parameter-property suite) asserts only node `name` and name-formatted edge strings, **both
  invariant under an id/qualifiedName rewrite**.
- **Fix:** gate the prefix for `Property` on the existing carve-out —
  `isPositionQualifiedLocalLabel(nodeLabel) && (nodeLabel !== 'Property' || isFunctionLocalProperty)`
  — or apply the `PARAMETER_LIST_NODE_TYPES` exemption inside `enclosingCallablePrefix`. Add a
  regression assertion on the **node id**, not the name.

### P1-3 · Dart: top-level and `final` closure bindings still cannot be call SOURCES

- **Inline anchor (in diff hunk):** `languages/dart/query.ts:109` — the new
  `(initialized_variable_definition …)` declaration rule.
- **Trigger:** `var handler = (int x) => target(x);` at top level (and the `final` / `static final`
  forms).
- **Evidence — [parser-probe]** (weaker tier than P1-1/P1-2: query-match level, not a graph run).
  The coordinator ran the PR-head Dart query against the real vendored grammar: the new rule matched
  **0 times** on that source. The AST is
  `program > inferred_type > initialized_identifier_list > initialized_identifier > identifier +
  function_expression` — Dart's top-level shape — while the new rule matches only
  `initialized_variable_definition` (the function-local shape). Without `@declaration.function`,
  `findFunctionBody` never synthesizes the closure scope and the call is attributed to the File.
  **Aggravating:** `dart/captures.ts` `DART_CALLABLE_CAPTURE_OPTIONS.bindingNodeTypes` *already*
  lists `initialized_identifier` and `static_final_declaration`, with a comment saying they cover
  top-level `var`/`final` and later declarators — the codebase already knew the shape and the new
  rule did not mirror it.
- **Corroboration: MODERATE by lane count** (consistent within Codex — adversarial [HIGH] and ce #1
  conf 100), **but parser-probe-verified by the coordinator**, so postable at P1. No Claude lane
  found this variant; the Claude lanes found only the *second-declarator* variant (P2-2).
- **Fix:** emit the same inner-function declaration capture for every Dart binding shape already
  supported as a callable target — top-level `var`, top-level / `static final`, and nested
  `initialized_identifier`. Add SOURCE-edge tests per form.

### P1-4 · Ruby: `do … end` closures (the dominant multi-line style) are not captured; brace `lambda`/`proc` may have no graph node

- **Inline anchor (in diff hunk):** `languages/ruby/query.ts:106` — `block: (block)
  @declaration.function` in the new `lambda` pattern (all three new patterns require `(block)`).
  Graph channel context: `tree-sitter-queries.ts:1594-1596`.
- **Trigger:**
  ```ruby
  handler = lambda do |x|
    target(x)
  end
  ```
  plus `proc do … end` and stabby `->(x) do … end`; `Proc.new { }` is missing entirely.
- **Evidence — [parser-probe] in BOTH engines** (weaker tier than P1-1/P1-2). Claude ran the PR-head
  `RUBY_SCOPE_QUERY` verbatim against the vendored grammar: of `->(x){}`, `lambda{}`, `proc{}`,
  `->(x) do…end`, `lambda do…end`, `Proc.new{}`, **only the first three matched**. The coordinator's
  independent probe: of four binding forms, **1 of 4** matched; the AST confirms `lambda > do_block`.
  `(do_block) @scope.block` already exists at `ruby/query.ts:52`, so the scope is there — the
  `do_block` Block simply owns no def and falls through to the enclosing method/File. Rubocop's
  default `Style/BlockDelimiters` is `line_count_based` (braces single-line, `do…end` multi-line), so
  the missed forms are precisely the ones with enough body to contain calls.
- **Second half, [code-read] and Codex-only:** Codex swarm R2 (conf 98) and Codex ce #2 (conf 100)
  additionally report that `RUBY_QUERIES` mints a graph `Function` node only when the assignment's
  RHS is the distinct `(lambda)` node used by stabby `->` syntax
  (`tree-sitter-queries.ts:1594-1596`) — so the new `lambda { }` / `proc { }` scope defs may have
  **no matching Function node at all**. The PR's own Ruby test uses the `->` brace form
  (`closure-binding-labels.test.ts:646`), the one form where both channels agree, so it passes.
  **This half was not runtime-verified at any tier — probe it before acting.**
- **Corroboration: STRONG cross-engine for the `do_block` gap** (Claude ce-correctness F2, Claude
  ce-adversarial F4, Claude risk F2, Codex adversarial [HIGH], Codex swarm R2, Codex ce #2).
  **Severity disagreement:** Claude risk rated it P3 on non-regression grounds (Ruby closures were
  never sources pre-PR); Codex rated it P1 because the PR's stated contract is "a closure binding is
  a call SOURCE in **every** language". This digest sides with Codex — an advertised contract that
  misses the language's idiomatic multi-line form is a P1 *for this PR*, even though it is not a
  regression.
- **Fix:** accept both `(block)` and `(do_block)` in all three patterns (three more standalone
  patterns keeps the no-alternation discipline), add a `Proc.new` shape, align `RUBY_QUERIES` for the
  `lambda` / `proc` call forms, and assert exact SOURCE edges per form.

### P2-1 · `SPLIT_SIGNATURE_NODE_TYPES` is not Dart-only; a **valid** TypeScript trigger exists

- **Inline anchor (in diff hunk 1271-1286):** `utils/ast-helpers.ts:1283` —
  `export const SPLIT_SIGNATURE_NODE_TYPES`, the derived set itself.
  The related docblock claim *"(Signatures are Dart-only; no language emits a legitimately-
  function-local Property under one.)"* sits at **`ast-helpers.ts:1244`**, which is **not in the
  diff** (it falls between the two hunks `@@ -411,0 +412,50 @@` and `@@ -1220,0 +1271,16 @@`), so it
  is quoted in the comment text rather than anchored. The new docblock at `:1271-1282` repeats the
  Dart-only framing.
- Enumerating `node-types.json` for every grammar the analyzer loads: `function_signature` /
  `method_signature` exist in **dart, typescript and tsx**. Trigger, valid TS in a `.ts` file:
  ```ts
  declare namespace Api {
    function internalHelper(x: number): number;
    export function send(x: number): number;   // -> Function:api.ts:internalHelper.send@2:9
  }
  ```
- **[code-read + grammar/walk simulation].** Corroboration: **consistent within Claude only** —
  ce-adversarial F2 (P2, conf 80, found the valid trigger) and swarm risk F3 (P3, believed it
  required *invalid* TS). Adversarial's evidence is stronger. Mitigations: `.d.ts` is ignored, and a
  scan of all 768 repo `.ts` files found **0** definition-anchor fallback hits, so this repo is clean
  today.
- **Fix:** don't derive the set by subtraction — gate on the provider, or require the matched
  sibling's `nextNamedSibling` to be the body the walk descended from. At minimum correct both
  docblocks (`:1244` and `:1271-1282`).

### P2-2 · Dart: the second and later declarators of `var f = …, g = …` can never be call SOURCES

- **Inline anchor (in diff hunk):** `languages/dart/query.ts:111` —
  `(function_expression) @declaration.function)`, the second child of the unanchored two-child
  pattern.
- `g`'s `function_expression` is a **grandchild** (nested inside `initialized_identifier`), so the
  unanchored two-child pattern never reaches it — **[parser-probe]**: exactly one match (`f`). The
  graph channel already carries a dedicated rule for this shape, so `g` becomes a call *target* but
  never a *source* — the precise condition #2699 part B set out to remove.
- Corroboration: **cross-engine** (Claude ce-correctness F4, Claude ce-adversarial F6, Codex ce #1
  names later declarators explicitly).
- **Fix:** mirror the graph channel with the nested `(initialized_identifier …)` arm.

### P2-3 · New twin list: `isCallableDef` re-spells `isOverloadableCallable`, which the same file imports

- **Inline anchor (in diff hunk 79-101):** `graph-bridge/ids.ts:79`.
- Used at `:117` and `:138`; the canonical helper is imported at `:32` and called at `:249`, `:268`,
  `:280`. **[code-read — coordinator read both].** This is the exact "same list in two places,
  nothing enforcing agreement" class the PR itself documents as having hit this repo six times, and
  unlike the `IMPLICIT_RECEIVERS` / `THIS_RECEIVERS` pair — which **this PR** adds a drift guard for
  — the new twin has none. Lane-rated P1 (maintainability, conf 92); digest-rated **P2** because
  there is no runtime trigger today. Single lane, single engine.
- **Fix:** `const isCallableDef = (d: SymbolDefinition) => isOverloadableCallable(d.type);` — or
  delete it.

### P2-4 · File-wide `localNameKey` guard can drop a genuine class `Property`

- **Body-only** (see §9): the mechanism spans `graph-bridge/ids.ts:212-228`, of which only `:212` is
  in the diff, and the finding is **explicitly unverified**.
- Trigger: `class C { val x = 0; fun f() { val x = 1 } }` on compact or generated code — class and
  local `Property` nodes then share file + 0-based line + label + name, the line-only position key
  becomes ambiguous, the local installs a file-wide `localNameKey`, and **both** defs hit the early
  `return undefined` before the genuine member's qualified `C.x` key is tried. The same design
  affects `Static`.
- **[code-read], Codex adversarial only ([MEDIUM]), explicitly NOT verified by the coordinator** —
  needs a runtime probe before acting. Related and independently noted by Claude lanes: the
  `ids.ts:226` fail-closed is now armed far more often (Vue SFC `lineOffset` reaching values;
  multi-line TS enum members whose two channels anchor differently).
- **Fix:** include the start column in the position key, or make the local-name guard scope-aware.

### P2-5 · Self-contradicting leftover comment fragment

- **Inline anchor (in diff hunk 2301-2307):** `workers/parse-worker.ts:2301` — the first **added**
  line, which is the sentence that contradicts the surviving fragment directly above it.
- The dangling clause itself is at **`:2300`**, an **unchanged context line**: a partial sentence
  replacement left `// same-named file-level function. Restricted to CALLABLE labels: the` in place,
  immediately followed by the new `// Applies to VALUES as well as callables since #2699 closed
  A1: …`. **[code-read — coordinator read the head file].** Three Claude lanes agree (maintainability
  F2 P2, correctness F5 P3, adversarial residual). This sits directly above the gate that three
  phases must agree on — the one place the codebase itself argues that a wrong comment costs silent
  edge drops.
- **Fix:** delete the dangling clause at `:2300`.

### P2-6 · `learnings.jsonl` entry violates `gitnexus-plan`'s own contract

- **Inline anchor (in diff hunk 3-6):** `eval/workflow_bench/learnings.jsonl:3`.
- Line 3 appends `"skill": "gitnexus-plan"`, but `.claude/skills/gitnexus-plan/SKILL.md:344-348`
  states: *"Feedback is chat-only: do not append evaluation learnings, edit benchmark data, or
  modify this skill during a live planning task."* The other three new entries
  (`"skill": "gitnexus-work"`) **are** sanctioned by `gitnexus-work/SKILL.md:259-269` and are
  correctly formatted. **[code-read — coordinator read the skill file].** Single lane, single engine,
  but it is a quoted-rule violation on an exact line.
- **Fix:** drop line 3, or re-file it as `gitnexus-work` feedback if the friction arose during
  execution.

### P3s

- **Ruby `lambda`/`proc` patterns don't constrain the receiver** — **inline anchor (in diff hunk):**
  `ruby/query.ts:104` (`right: (call`). **[parser-probe]**: `handler = MyMod.lambda { |x| x }`
  matches and is wrongly declared a closure binding. Two Claude lanes, grammar-verified.
  *Fix: add `!receiver`.* **(inline-worthy — one-token fix)**
- **Stale comment contradicting the head** — **inline anchor (in diff hunk 556-569):**
  `closure-binding-labels.test.ts:558` still reads "Dart — STILL OPEN" while the same file adds
  passing Dart source-attribution coverage at `:607-649`. A Kotlin graph-query comment still says
  `f()` does not resolve. Both engines flagged this (Codex swarm D1, Codex ce residual, Claude
  standards residual). **(inline-worthy)**
- **Unanchored regex on an id that embeds a file path** (`ids.ts:97`, new; identical pattern
  pre-existing at `:164` and 3× in `callable-value-flow.ts`). `String.match` takes the *first*
  `#(\d+):(\d+):`, so a path like `src/bugs#12:34:test/handler.php` mis-parses. Claude security rated
  P2; Claude adversarial **partially refuted** it — the failure is fail-*closed* (returns `false` →
  coarser attribution, never a fabricated edge), the exposure predates this PR, and colons are
  illegal in Windows paths. Net **P3**, **body-only by judgment** (the line *is* in a diff hunk).
  *Fix: anchor to the last `#`.*
- **The live PR body is stale** — its "Still open" section says A1 values and the receiver twin-list
  guard are deferred, but commit `da3d8397` implements both. Both engines. Also: #2699 is **not**
  referenced as closed by this PR; add `Closes #2699` if that is the intent. Body-only (no code line).
- **`Static` is in `isPositionQualifiedLocalLabel` but not in `isLinkableLabel`**
  (`callable-labels.ts:48` vs `node-lookup.ts:247-281`) — the id phase qualifies function-local
  `Static`, while `buildGraphNodeLookup` skips every `Static` before registering keys. No edge lost
  today (`Static` was never resolvable), but it is the same asymmetry shape #2714 fixed. Body-only
  (the `node-lookup.ts` half is not in the diff).
- **`this-boundary.test.ts:198` keeps `.toContain(...)`** where sibling assertions in the same PR use
  `toEqual` on a sorted array; would not catch a spurious extra edge. Body-only by judgment (nit).
- **Multi-declarator Dart id anchor asymmetry** (`closure-binding-labels.test.ts:272-287`): `f`'s
  position anchors to the whole `var …;` statement (`@1:2`) while `g`'s anchors to its own
  `initialized_identifier` (`@1:24`). Correctly pinned by the test; flagged as a surprising id-scheme
  inconsistency a reader of `impact()` output would hit. Body-only (informational).

### 9.1 What's solid (validated, not assumed)

- **Schema/invalidation discipline is correct and self-documented.** `INCREMENTAL_SCHEMA_VERSION`
  20→21 (with a `passesReuseGate(20) === false` pin) and `parse-cache.ts` `SCHEMA_BUMP` →**29**, with
  the "29 not 28" rationale documenting the #2415 merge-order collision. Verified independently by
  four lanes across both engines; all rated it a non-risk.
- **The "shared pipeline code must not name languages" rule holds.** `findSplitBodyCallableAncestor`,
  `SPLIT_SIGNATURE_NODE_TYPES` (a derived set difference, not a hand list) and `pickCallerCallableDef`
  are genuinely generic — parameterised node-type sets, scope-kind + position comparisons.
  Dart/PHP/Kotlin/Ruby/Rust appear only in comments and inside `languages/<lang>/`.
- **The flipped tests are honest re-encodings, not "changed until green."** All four flips
  (Kotlin/PHP closure source, function-local value identity, `this`-boundary, schema version) were
  traced back to the specific source mechanism that produces the new expected id/edge.
- **The three new tests discriminate.** Dart two-same-named-closures, Ruby lambda, and
  `receiver-twin-list-drift` were each reasoned through pre-fix vs post-fix and shown to fail on
  revert/drift. The twin-list guard's regex helper was checked for vacuous matching against the
  `readonly string[]` annotation and is non-vacuous.
- **Comment fact-check mostly held.** `findSplitBodyCallableAncestor`'s docblock claims (fallback-only
  usage, derived-not-listed, required file ordering) all check out, and its "tsc does NOT catch the
  TDZ" claim was **independently reproduced** with a minimal `tsc --strict` repro.
  `scopeIsCallableBody`'s shared-coordinate-base claim is true. The `callable-id-lockstep.ts` comment
  edit is a genuine correction of a previously overstated claim. (The three that did **not** hold are
  P2-1, P2-5 and the `callable-labels.ts:34-36` claim under P1-2.)
- **No security boundary is crossed.** No auth / permission / secret / shell / deploy surface; the
  widened ids encode a strict subset of source the same callers can already read. No workflow, CI,
  Docker or YAML file is touched.

### 9.2 Test gaps

1. **Rust (S3) has ZERO coverage — P1-level gap.** The closure-as-call-source rule *and* the new
   graph `Function` node for `let handler = || …;` (`tree-sitter-queries.ts:1346-1348`) are untested
   anywhere in the suite. The only Rust closure fixture (`rust-scope.test.ts:919-954`) never touches
   the closure; the Rust case at `closure-binding-labels.test.ts:714` is a non-closure re-binding the
   new rule deliberately does not match, so it stays green either way. **Cross-engine** (Claude
   ce-testing P1 conf 85, Codex swarm T1, Codex ce coverage note).
2. **No multi-line closure test in any of the five languages** — the single condition that decides
   whether the feature works for a binding (P1-1). Both engines.
3. **No Ruby `do … end` / `proc` / `Proc.new` case, and no receiver-qualified `.lambda` / `.proc`
   case.**
4. **No Dart top-level, `final`, or later-declarator SOURCE case.**
5. **No test asserts the invariant P1-2 breaks** — that a class member's id survives the widening to
   `Property` / `Static`. The existing parameter-property suite asserts `name` only, never the id.
6. **The `pickCallerCallableDef` Block safeguard's NEGATIVE case is untested** — a nested named
   function inside a block must not steal a block-level call. The alignment check is the only thing
   preventing mis-attribution, and nothing exercises its false branch.
7. **`SCHEMA_BUMP = 29` has no pinned-value test**, unlike its sibling `INCREMENTAL_SCHEMA_VERSION` —
   in the same failure class the PR's own comment documents as having already happened once.
8. **`function-local-identity.test.ts` asserts node ids only** — the failure mode the PR names as the
   reason the predicate must be shared is about *edges*; two distinct nodes with every edge on the
   wrong one would pass.

### 9.3 Coverage limits, engine liveness and verification tiers

- **Verification tiering (read this before acting on any finding).** Only **P1-1 and P1-2** were
  established end-to-end: the coordinator ran the **built analyzer CLI (`analyze --index-only`)**
  over purpose-built fixtures and queried the resulting graph with **`cypher`**. **P1-3 and P1-4**
  were established only at **parser-probe level** — the PR-head tree-sitter queries run against the
  real vendored grammars, which proves what the query matches but exercises neither scope synthesis
  nor edge emission. That is a genuinely weaker tier and is flagged as such on each finding. All
  other findings are `[code-read]` or simulation.
- **No lane could build `dist/` inside the review worktree**, so lane-level findings are source
  correlation plus out-of-tree grammar probes against the same vendored grammars the analyzer loads.
  The two analyzer runs above were the coordinator's, not a lane's.
- **Engine liveness.** All three `codex-*.md` artifacts contain real findings, not error/empty
  markers. Codex legs that actually ran: **B1** (GitNexus swarm, solo lanes 1–6; lane-7 critic
  deliberately not run), **B2** (ce-code-review, report-only, five personas), **B3** (adversarial).
  **Codex's own internal cross-model leg failed with a bubblewrap namespace error and fell back to an
  in-process adversarial pass** — so Codex's "independent validation" is degraded by its own
  admission, and its two P1s are marked validation-degraded after direct source verification.
- Codex's `npx vitest` attempt failed because the built parse worker was absent in the worktree
  (expected for a read-only review); its `npx tsc --noEmit` hit six pre-existing workspace/shared-type
  errors outside this diff and reported **no template-literal syntax errors**. A targeted unit run
  (`callable-id-lockstep` + `receiver-twin-list-drift`) passed.
- **The Claude test/CI lane is DEGRADED** — its agent ended mid-investigation **twice** and returned
  no findings. The coordinator answered that lane's CI-execution question directly:
  `.github/workflows/ci-tests.yml:10-56` runs the full vitest suite sharded across ubuntu runners
  with blob-merged coverage, so every affected suite does execute on a PR, and all eight bench
  fingerprint gates are present. Assertion-quality coverage came from the ce-testing lane instead.
- **Not independently reproduced:** the PR's claimed 2926/2926 resolver sweep.

### 9.4 Posting recommendation

**INLINE — 13 anchors across 11 findings; every anchor confirmed by the coordinator to fall inside a
diff hunk of `ff86ccf1..cfd2434c`:**

| # | Finding | Inline anchor | Containing hunk |
|---|---------|---------------|-----------------|
| 1 | P1-1 | `ids.ts:212` | `@@ -173 +212 @@` |
| 2 | P1-1 | `php/query.ts:99` | `@@ -89,0 +90,17 @@` |
| 3 | P1-2 | `parse-worker.ts:2310` | `@@ -2294,2 +2310 @@` |
| 4 | P1-2 | `callable-labels.ts:42` | `@@ -16,0 +17,34 @@` |
| 5 | P1-3 | `dart/query.ts:109` | `@@ -94,0 +95,18 @@` |
| 6 | P1-4 | `ruby/query.ts:106` | `@@ -81,0 +82,34 @@` |
| 7 | P2-1 | `ast-helpers.ts:1283` | `@@ -1220,0 +1271,16 @@` |
| 8 | P2-2 | `dart/query.ts:111` | `@@ -94,0 +95,18 @@` |
| 9 | P2-3 | `ids.ts:79` | `@@ -75,0 +79,23 @@` |
| 10 | P2-5 | `parse-worker.ts:2301` | `@@ -2289,3 +2301,7 @@` |
| 11 | P2-6 | `learnings.jsonl:3` | `@@ -2,0 +3,4 @@` |
| 12 | P3 Ruby `!receiver` | `ruby/query.ts:104` | `@@ -81,0 +82,34 @@` |
| 13 | P3 stale Dart comment | `closure-binding-labels.test.ts:558` | `@@ -544,3 +556,14 @@` |

*(13 inline comments across 11 findings — P1-1 and P1-2 each carry two anchors.)*

**Re-anchored during revision, because the original anchor was NOT in the diff:**

- **P2-1** moved `ast-helpers.ts:1244` → **`:1283`**. Line 1244 falls between the two `ast-helpers.ts`
  hunks (`@@ -411,0 +412,50 @@` and `@@ -1220,0 +1271,16 @@`) and is therefore not commentable; the
  "Signatures are Dart-only" docblock it contains is quoted in the comment body instead.
- **P1-1** moved `ids.ts:214` → **`:212`** (`:214` is unchanged context; `:212` is the changed gate).
- **P1-2** moved `parse-worker.ts:2309` → **`:2310`** (`:2309` is unchanged context; `:2310` is the
  changed gate line).
- **P2-5** moved `parse-worker.ts:2298` → **`:2301`** (the dangling clause at `:2300` is unchanged
  context; `:2301` is the first added line and the sentence that contradicts it).

**BODY-ONLY (7) — no in-diff anchor, or a deliberate judgment not to post inline:**

- **P2-4** (`ids.ts:212-228` class-`Property` drop) — Codex-only and **explicitly unverified**; only
  `:212` is in the diff and that line already carries the P1-1 comment. *Judgment + partial anchor.*
- **P3 unanchored regex** `ids.ts:97` — the line **is** in a diff hunk, but the finding was partially
  refuted (fail-closed, pre-existing pattern). *Judgment.*
- **P3 `Static` / `isLinkableLabel` asymmetry** — the `node-lookup.ts:247-281` half is **not** in any
  hunk. *No anchor.*
- **P3 stale PR body / missing `Closes #2699`** — no code line at all. *No anchor.*
- **P3 `toContain` at `this-boundary.test.ts:198`** — in a hunk, but a nit. *Judgment.*
- **P3 Dart multi-declarator anchor asymmetry** — informational; the test correctly pins it.
  *Judgment.*
- **All eight test gaps**, plus the Vue SFC `lineOffset` residual risk. *No single anchor.*

Every item below was actively attacked and **did not survive**. Recorded so it is not re-litigated,
and so the author is not asked to defend a non-issue:

1. **Line-base off-by-one** (the top recalled lead, and the highest-risk hypothesis going in) —
   **REFUTED by four lanes across both engines with the same evidence chain**: `types.ts:60` documents
   `Range.startLine` as 1-based; `nodeToCapture` / `spanCapture` build ranges with `row + 1`;
   `makeDefId` (`scope-extractor.ts:870`) uses `range.startLine` directly; graph-node `startLine` is
   0-based (`parse-worker.ts:2610`, `row + lineOffset`). So `scopeIsCallableBody` compares 1-based to
   1-based (no conversion needed — the docblock is accurate) and `ids.ts:214`'s `line - 1` is the
   correct, *different* bridge.
2. **Backticks / template-literal-breaking chars in the five new query comments** (recalled lead #3)
   — REFUTED by three lanes across both engines; the #45123-class build break is not repeated.
3. **Block scope stealing attribution** / `scopeIsCallableBody` accepting a non-body Block — REFUTED.
   For a closure, `@declaration.function` and `@scope.block` sit on the *same* node; for a nested
   `fun foo()` inside a block, the block opens at `{` and the def points at `fun`, so the alignment
   test rejects it. No grammar produced a counter-example. (One nuance: the comment's own Kotlin
   example is wrong — `function_declaration` is itself `@scope.function` — but the guard is not
   weaker.)
4. **`scopeIsCallableBody`'s regex failing open** — REFUTED: `null` → `return false` → coarser
   attribution, never a fabricated edge.
5. **`findSplitBodyCallableAncestor` picking a wrong preceding sibling in Dart** — REFUTED by both
   engines, including a simulator run over abstract classes, `external` functions, enums, mixins,
   extensions, getters/setters and nested local functions.
6. **Schema bumps** — REFUTED as a risk: both single increments, rationale entries accurate,
   29-not-28 correctly documents the #2415 collision.
7. **Ruby `items.map { … }` false positive** — REFUTED by grammar probe: the `#eq?` predicate rejects
   it, 0 matches.
8. **Kotlin `kotlinBindingScopeFor` / `KOTLIN_THIS_RECEIVER_LAMBDAS` interaction** (recalled lead #4)
   — REFUTED by both engines: untouched by this PR, the body-position gate is retained, and the
   `this`-boundary flip is semantically *more* precise (the call really is made by the lambda).
9. **Kotlin `lazy { }` / `apply { }` / `run { }` wrongly declared callables** — REFUTED by real query
   probe (the lambda sits inside a `call_expression`, not a direct child of `property_declaration`).
10. **PHP anchor misalignment** — REFUTED: `php/query.ts:60-61` already declares both
    `anonymous_function` and `arrow_function` as `@scope.function`.
11. **Dart extension-type preprocessing shifting positions** — REFUTED: the rewrite is
    length-preserving by construction and bails if the replacement would be longer.
12. **A divergent second/fourth id-building phase** — REFUTED: `isPositionQualifiedLocalLabel` has
    exactly three call sites, and `parsing-processor.ts` is now a pure merge layer.
13. **`pickCallerCallableDef` reorder changing Function-kind behaviour** — REFUTED: identical for
    `Function` kind, and sibling scopes cannot overlap (`scope-tree.ts:137-153` throws), so a Block
    child cannot pre-empt a Function child.
14. **ReDoS in the new regexes / `@\d+:\d+$` aliasing two real symbols** — REFUTED (linear; the suffix
    is system-injected, not a source-level identifier form in any of the five languages).
15. **Object-literal owner ids, `enrichExportedTypeMap` hand-built ids, `pruneLocalValueSymbols`** —
    all REFUTED as affected (they only see file-scope bindings, or key on `scope` / edges, not ids).
16. **TDZ / module-load-order hazard on the derived set** — REFUTED (declaration order is correct);
    and the docblock's "tsc does not catch this" claim was independently **reproduced as true** with a
    minimal `tsc --strict` repro.
17. **Branch hygiene, security boundary, `git diff --check`, bidi/RTL, secrets or PII in
    `learnings.jsonl`, and the perf of the new Block loop path** — all clean.
18. **Merge state was re-queried first-hand against head `cfd2434c` rather than trusting the lane's
    earlier "`BLOCKED` / CI Gate pending" snapshot** — the lane data was stale; see §4.

## 11. Open questions

1. **P1-4, second half** — do `lambda { }` / `proc { }` brace bindings get a graph `Function` node at
   all? Codex asserts they do not (`RUBY_QUERIES` matches only the stabby `(lambda)` node,
   `tree-sitter-queries.ts:1594-1596`), but this half was **not** runtime-verified at any tier. Probe
   before acting on that specific half.
2. **P2-4** — is the file-wide `localNameKey` guard actually reachable for a genuine class `Property`
   sharing a line-key with a function-local of the same name? Codex-only, `[code-read]`, and
   explicitly unverified by the coordinator.
3. **Intent on #2699** — should this PR carry `Closes #2699`? It implements parts A+B but references
   no closing issue.
4. **Not a review question, but an outstanding gate:** the PR body still describes A1 values and the
   twin-list guard as deferred and must be refreshed before merge.

## 12. Final verdict

**not production-ready**

Four concrete triggers survive on ordinary, idiomatic code, all inside the PR's own primary contract
surface, none covered by a test:

- **P1-1** — a multi-line closure binding emits a **fabricated** `outer -> target` CALLS edge in all
  five changed languages. This is the exact defect class the PR set out to remove, and it was
  reproduced end-to-end by running the built analyzer on a purpose-built fixture.
- **P1-2** — TypeScript constructor **parameter properties** are re-keyed as constructor-locals, a
  **regression** against the base that silently empties the `UserService.repo` slot that `impact`,
  `rename` and FTS depend on. Also reproduced end-to-end.
- **P1-3 / P1-4** — Dart top-level/`final` bindings and Ruby `do … end` closures — the dominant
  multi-line form in each language — are still not call sources, so the advertised "every language"
  contract does not hold. Verified at the weaker parser-probe tier.

The engineering underneath is genuinely strong: the review **refuted more attacks than it landed**
(18 items in §10), the schema/invalidation discipline is correct and self-documented, the
language-agnosticism rule holds, the flipped tests are honest re-encodings, and the three new tests
discriminate. Branch hygiene is fine — **merge-from-main commit present but harmless and merge-safe**
— and CI is **complete**, with only the known-unfixable Vercel fork check red. **Nothing about the
branch or the pipeline blocks this PR; the four findings do.**

Recommended path: fix P1-1 (share one identity across the two anchor channels, and fail closed at the
callable boundary instead of climbing) and P1-2 (restore the `PARAMETER_LIST_NODE_TYPES` carve-out for
`Property`); extend the Dart and Ruby query shapes for P1-3/P1-4; add multi-line and per-form SOURCE
tests plus the missing Rust coverage; correct the three false comments (P2-1, P2-5, the
`callable-labels.ts:34-36` docblock); drop the `learnings.jsonl` line-3 violation; and refresh the
stale PR body.

---

*Automated multi-engine digest: Claude GitNexus swarm + Claude ce personas + three Codex
gpt-5.6-sol/xhigh legs, merged by a synthesis agent and gated by a synthesis critic. **The PR author
and the review coordinator are the same agent** — treat cross-engine (Codex) disagreement as the
load-bearing signal. **Codex's internal cross-model leg failed with a bubblewrap namespace error and
fell back to an in-process adversarial pass**, and **the Claude test/CI lane was degraded** (its agent
ended twice without returning findings). Findings tagged **[analyzer-verified]** were run end-to-end
through the built analyzer CLI; **[parser-probe]** findings were verified only at query-match level
against the vendored grammars; **[code-read]** findings were not executed. Verify before acting.*

gitnexus/src/core/ingestion/scope-resolution/graph-bridge/ids.ts
```diff
@@ -170,7 +209,7 @@ export function resolveDefGraphId(
     // 0-based, def ids 1-based. An `AMBIGUOUS_POSITION` tombstone (two
     // callables on one line) falls through to the name-based keys below.
     const line = defStartLine(def.nodeId);
-    if (line !== undefined && isOverloadableCallable(def.type)) {
+    if (line !== undefined && isPositionQualifiedLocalLabel(def.type)) {
```
magyargergo, line 225: **[P1] Multi-line closure bindings emit a fabricated CALLS edge.** `[analyzer-verified]` · cross-engine (Codex swarm P1/99% + Claude adversarial + Claude risk-architect)

The two query channels anchor on different nodes by design — graph-node on the OUTER wrapper, scope-resolution on the INNER closure. This position join is line-only, so it holds on one line and misses when the literal starts on the next. The miss hits the fail-closed return below, and caller resolution then **climbs to the enclosing callable**.

Trigger (ran through the built analyzer, graph read back via `cypher`):
```php
$single = function ($x) { return target($x); };
$multi =
    function ($x) { return target($x); };
```
Emitted:
```
outer.$single@3:2 -> target   (correct)
outer             -> target   (FALSE — `outer` calls nothing; $multi does)
```
`outer.$multi@4:2` exists as a node with **zero outgoing edges**.

This is the same fabricated-edge defect class the PR removes for Dart, reintroduced for every multi-line binding in all five changed languages. Engine note: a Claude lane predicted a *dropped* edge (P3); Codex predicted a *false* edge (P1) — the probe shows Codex correct.

**Fix:** share one def→graph identity across both anchor channels, or bridge on the initializer position; and when a callable boundary is found but cannot be bridged, fail closed **at that boundary** instead of climbing.

gitnexus/src/core/ingestion/languages/php/query.ts
```diff
@@ -87,6 +87,23 @@ const PHP_SCOPE_QUERY = `
...
+;; Closure assigned to a variable: $handler = function () {...}; or fn() => ...;
+;; Anchor discipline (same contract as javascript/query.ts): @declaration.function
+;; sits on the INNER anonymous_function / arrow_function, NOT on the
+;; assignment_expression wrapper. That aligns anchor.range with the
+;; @scope.function range above, so pass2AttachDeclarations attaches the
+;; declaration to the CLOSURE's own scope rather than the enclosing function's.
+;; Without this the closure scope owns no callable def and
+;; pickCallerCallableDef falls through to the enclosing callable, making the
+;; closure a call TARGET but never a call SOURCE (#2699).
+(assignment_expression
```
magyargergo, line 99: **[P1] Anchor-split half of the multi-line defect.** `[analyzer-verified]`

This comment states the discipline correctly — `@declaration.function` on the INNER `anonymous_function`, not the `assignment_expression` wrapper. But `tree-sitter-queries.ts` anchors the graph-node capture on the **outer** assignment, and the bridge joins the two on line only. Same-line bindings coincide; multi-line ones do not (see the `ids.ts:212` comment).

**Fix:** whichever channel moves, both must agree — or the join must not be line-only.

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -2286,13 +2298,16 @@ const processFileGroup = (
...
+      // otherwise collapse onto one `Const:v.ts:handler`, which was the
+      // issue's original complaint and is unreachable from a callable-only
+      // gate. `isPositionQualifiedLocalLabel` is the single definition of that
+      // set, shared with resolution in `ids.ts` — the two phases disagreeing
+      // silently drops edges rather than failing (#2714).
       // Same helper as the caller-attribution phase — see `enclosingCallablePrefix`.
       const nestedCallablePrefix =
-        (nodeLabel === 'Function' || nodeLabel === 'Method' || nodeLabel === 'Constructor') &&
-        definitionNode
+        isPositionQualifiedLocalLabel(nodeLabel) && definitionNode
```
magyargergo: **[P1] TypeScript constructor parameter properties are re-keyed as constructor-locals.** `[analyzer-verified]` · two Claude lanes (correctness + adversarial), coordinator-reproduced

Widening the gate to `Property` makes this walk reach the constructor `method_definition` **through the parameter list** — a `LOCAL_SCOPE_BODY_NODE_TYPES` hit that lands before the `class_body` boundary. A class member therefore gains a local-style id.

Trigger (built analyzer):
```ts
export class Service {
  constructor(private readonly port: Port) {}
}
```
Produces `Property:src/svc.ts:Service.constructor.port@2:14` instead of `Service.port`.

This is the Angular/NestJS DI idiom, it survives `pruneLocalSymbols`, and the class asserts `HAS_PROPERTY` against it — so `impact`, `rename` and FTS lose the member slot. A real instance exists in this repo at `src/core/group/service.ts:304`.

**Fix:** `parse-worker.ts:2245-2257` already computes the correct parameter-list exemption (`isFunctionLocalProperty`); this ternary discards it. Reuse it.

gitnexus/src/core/ingestion/utils/callable-labels.ts
```diff
@@ -14,3 +14,37 @@ import type { NodeLabel } from 'gitnexus-shared';
...
+ * what keeps this off the symbols other files and stored references address.
+ * A class field stays unqualified even inside a function, because the prefix
+ * walk boundaries on class-likes.
+ *
+ * ONE definition, deliberately: the id-building phase and the resolution phase
+ * must agree on this set or the caller attaches to a node that does not exist
+ * and the edge is silently dropped — the failure mode #2714 fixed, invisible
+ * from outside because "zero dangling edges" is what it looks like.
+ */
+export function isPositionQualifiedLocalLabel(label: NodeLabel | undefined): boolean {
```
magyargergo, line 42: **[P1] This docblock's class-member claim is false for `Property`.** `[analyzer-verified]`

The comment above states a class field stays unqualified because the prefix walk boundaries on class-likes. For a TypeScript **constructor parameter property** the walk reaches `method_definition` via the parameter list *before* any class boundary, so the field is re-keyed (`Service.constructor.port@2:14` — verified by running the analyzer).

**Fix:** restore the parameter-list carve-out for `Property`, then the comment becomes true again. If `Property` is kept in the set without the carve-out, the comment must be corrected.

gitnexus/src/core/ingestion/languages/dart/query.ts
```diff
@@ -92,6 +92,24 @@ const DART_SCOPE_QUERY = `
...
+; sits on the INNER function_expression, NOT on the local_variable_declaration
+; wrapper. Dart is the one language that declares NO @scope.function in this
+; file — its function scopes are SYNTHESIZED in captures.ts from
+; declNode + findFunctionBody(declNode). So this rule deliberately does not add
+; a @scope.function of its own: doing that would collide with the synthesized
+; one at identical range, and duplicate scope ids make buildScopeTree throw,
+; which drops the whole file. Instead findFunctionBody now understands a
+; closure's child function_expression_body, so the existing synthesis produces
+; exactly one scope, anchored on the same node as the declaration (#2699 S4).
+(initialized_variable_definition
```
magyargergo, line 120: **[P1] Dart top-level and `final` closures still cannot be call sources.** `[parser-probe]` · cross-engine (Codex ce P1/100 + Codex adversarial + Claude lanes)

This rule matches only `initialized_variable_definition` — Dart's function-**local** shape. Idiomatic top-level `var handler = (x) => target(x)` parses as `initialized_identifier`; top-level `final` uses another shape again. Neither gets `@declaration.function`, so `findFunctionBody` never synthesizes the closure scope.

Probe against the vendored grammar: this query returns **0 matches** for `var handler = (int x) => target(x);` at top level. AST is `program > inferred_type > initialized_identifier_list > initialized_identifier`.

`dart/captures.ts` already lists `initialized_identifier` and `static_final_declaration` in `bindingNodeTypes`, with a comment explaining they cover top-level `var`/`final` — the codebase knew this shape and the new rule doesn't mirror it.

**Fix:** emit the same inner-closure declaration capture for every binding shape already recognised as a callable target, and add SOURCE tests per form.

gitnexus/src/core/ingestion/languages/ruby/query.ts
```diff
@@ -79,6 +79,40 @@ const RUBY_SCOPE_QUERY = `
...
+;; predicates are a known hazard on this tree-sitter line.
+(assignment
+  left: (identifier) @declaration.name
+  right: (lambda body: (block) @declaration.function))
+
+(assignment
+  left: (identifier) @declaration.name
+  right: (call
+    method: (identifier) @_lambda-kw
+    block: (block) @declaration.function)
```
magyargergo: **[P1] Ruby `do … end` closures — the dominant multi-line form — are not captured.** `[parser-probe]` · cross-engine (Codex ce P1/100 + Codex adversarial + 3 Claude lanes)

All three new patterns require `(block)`. `lambda do…end`, `proc do…end` and stabby `->(x) do…end` all produce `(do_block)`.

Probe against the real grammar — of these four, only the first matched:
```ruby
a = ->(x) { target(x) }        # matched
b = ->(x) do target(x) end     # MISSED
c = lambda do |y| target(y) end # MISSED
d = proc do |z| target(z) end   # MISSED
```
`(do_block) @scope.block` already exists in this file, so these closures get a Block scope with empty `ownedDefs` and fall through to the enclosing method. The PR's new Ruby test uses the brace form only, so it passes.

**Fix:** accept both `block` and `do_block` in all three patterns; add SOURCE assertions for each form. `Proc.new` is also uncovered.

gitnexus/src/core/ingestion/utils/ast-helpers.ts
```diff
@@ -1218,6 +1268,22 @@ export const LOCAL_SCOPE_BODY_NODE_TYPES: ReadonlySet<string> = new Set(
...
+ * `function_signature` / `method_signature`).
+ *
+ * Derived, not listed, so it cannot drift from the two sets that define it:
+ * `LOCAL_SCOPE_BODY_NODE_TYPES` is `FUNCTION_NODE_TYPES` minus exactly the bare
+ * signature types, so the difference IS the split-signature set.
+ *
+ * Must stay BELOW `LOCAL_SCOPE_BODY_NODE_TYPES` — reading it earlier hits the
+ * temporal dead zone and throws at module load.
+ */
+export const SPLIT_SIGNATURE_NODE_TYPES: ReadonlySet<string> = new Set(
```
magyargergo, line 1304: **[P2] This derived set is not Dart-only, and a valid-TypeScript trigger exists.** `[code-read]`

The docblock at ~1244 says signatures are Dart-only. tree-sitter-typescript/tsx also emit `function_signature` and `method_signature`, so `findSplitBodyCallableAncestor` can fire on **valid** TS:
```ts
declare namespace Api {
  function internalHelper(x): number;
  export function send(x): number;
}
```
→ `send` becomes `internalHelper.send@2:9`.

(A Claude lane rated this P3 believing it needed *invalid* TS; the adversarial lane found the valid `declare namespace` form.) The repo's own 768 `.ts` files are currently clean and `.d.ts` is ignored, so impact today is low.

**Fix:** either scope the fallback so it cannot fire for grammars whose callables are ancestors, or correct the docblock and add a TS guard test.

gitnexus/src/core/ingestion/languages/dart/query.ts
```diff
@@ -92,6 +92,24 @@ const DART_SCOPE_QUERY = `
...
+  (identifier) @declaration.name
+  (function_expression) @declaration.function)
```
magyargergo, line 122: **[P2] The second declarator of a multi-name Dart binding gets no closure scope.** `[parser-probe]`

`var f = (x) => …, g = (y) => …;` — Dart wraps only the FIRST declarator in `initialized_variable_definition`; `g` is a nested `initialized_identifier`. It receives a node but no closure scope, so it can never become a call SOURCE.

Same root cause as the top-level gap above, and `dart/captures.ts` already documents this exact asymmetry for `bindingNodeTypes`.

gitnexus/src/core/ingestion/scope-resolution/graph-bridge/ids.ts
```diff
@@ -73,6 +76,29 @@ function rangeContainsPoint(
   return true;
 }
 
+const isCallableDef = (d: SymbolDefinition): boolean =>
```
magyargergo: **[P2] New twin list, in the PR that ships a twin-list drift guard.** `[code-read]`

`isCallableDef` re-spells `Function | Method | Constructor`, which is exactly `isOverloadableCallable` — already imported in this file at :32 and called at :249, :268 and :280.

This is the same 'one list, two spellings, nothing enforcing agreement' class the PR documents as having hit this repo six times, and unlike the `IMPLICIT_RECEIVERS`/`THIS_RECEIVERS` pair it adds a guard for, this one has none.

**Fix:** drop `isCallableDef` and use `(d) => isOverloadableCallable(d.type)`.

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -2286,13 +2298,16 @@ const processFileGroup = (
       // #2699: a callable nested inside another callable is qualified by the
       // enclosing callable, so a function-local closure stops colliding with a
       // same-named file-level function. Restricted to CALLABLE labels: the
-      // collision that produced wrong CALLS edges is between callables, and
-      // widening it to every function-local Variable/Property would churn ids
-      // for symbols the local-symbol pruner mostly deletes anyway.
+      // Applies to VALUES as well as callables since #2699 closed A1: a
```
magyargergo, line 2301: **[P2] Self-contradictory comment left by a partial edit.** `[code-read]`

The preceding context line ends `…Restricted to CALLABLE labels: the` and this line begins `Applies to VALUES as well as callables…`. The old sentence was only partly replaced, so the block now asserts both. The code is correct; the comment is not.

This PR family has repeatedly shipped comments that were false when written — worth removing the dangling clause.

eval/workflow_bench/learnings.jsonl
```diff
@@ -1,2 +1,6 @@
...
+{"skill": "gitnexus-plan", "date": "2026-07-28", "task": "#2699 part B — closure binding as a call SOURCE across PHP/Rust/Kotlin/Ruby/Dart", "friction": "The safe plan writer fails closed on a v9fs (9p) worktree because renameat2(RENAME_NOREPLACE) is unsupported, returning EINVAL, so no plan can ever be published there and Phase 2's 'commit the plan document' step is unreachable.", "suggestion": "Detect the EINVAL-on-renameat2 case explicitly and fall back to open(O_EXCL)+write+fsync, which preserves the no-clobber guarantee the flag exists for; failing that, say v9fs is unsupported instead of surfacing a generic write failure."}
```
magyargergo: **[P2] This entry violates `gitnexus-plan`'s own feedback contract.** `[code-read]` · coordinator-verified

`.claude/skills/gitnexus-plan/SKILL.md:347-348`: *"Feedback is chat-only: do not append evaluation learnings, edit benchmark data, or modify this skill during a live planning task."*

This line carries `"skill": "gitnexus-plan"`. The other three entries are `"skill": "gitnexus-work"`, which **is** sanctioned by that skill's explicit append contract, and they are correctly formatted.

**Fix:** drop this line, or re-file it as `gitnexus-work` feedback if the friction arose during execution rather than planning.

gitnexus/src/core/ingestion/languages/ruby/query.ts
```diff
@@ -79,6 +79,40 @@ const RUBY_SCOPE_QUERY = `
...
+;; mapped = items.map { |i| ... } would wrongly declare mapped a callable.
+;; Separate #eq? patterns rather than one #match? alternation: alternation
+;; predicates are a known hazard on this tree-sitter line.
+(assignment
+  left: (identifier) @declaration.name
+  right: (lambda body: (block) @declaration.function))
+
+(assignment
+  left: (identifier) @declaration.name
+  right: (call
```
magyargergo, line 114: **[P3] `lambda` / `proc` are matched without constraining the receiver.** `[parser-probe]`

The `#eq?` predicate tests the method name only, so `MyMod.lambda { … }` or `obj.proc { … }` is captured as a closure binding. The `items.map { }` false positive the rule guards against is correctly rejected — this is the narrower sibling case.

**Fix:** require the call to have no receiver (or assert the receiver is `Kernel`).

gitnexus/test/integration/closure-binding-labels.test.ts
```diff
@@ -541,31 +553,100 @@ describeIfWorkerBuilt('a closure binding is a call TARGET, not yet a call SOURCE
   //     scope for the walk to consider.
   //
   // So a fix needs per-language work, not one switch: a callable-boundary
-  // signal independent of scope `kind` (Kotlin/Ruby), an association from a
-  // closure scope to its binding's def (PHP), and a scope that does not exist
-  // yet (Dart). See #2699.
+  // signal independent of scope `kind` (Kotlin/Ruby — DONE, S2), an
+  // association from a closure scope to its binding's def (PHP — DONE, S1;
+  // Rust — DONE, S3), and a scope that does not exist yet (Dart — STILL OPEN).
```
magyargergo: **[P3] Comment is stale as of this same commit.** `[code-read]`

This block still describes Dart as an open case ("a scope that does not exist yet"), but S4 in this PR gives Dart both a closure scope and enclosing-callable identity. Codex's ce lane flagged the same class of residue elsewhere in the changed tests.

**Fix:** update to reflect what actually remains open — per this review, Dart top-level/`final` bindings and the second declarator.

PR #2719, opened 2026-07-27
fix: allow slower remote embedding responses

voidfreud:
## Problem

The OpenAI-compatible HTTP embedding client used a hard-coded 30-second per-request timeout. Some reliable remote embedding providers can return a valid batched response after a cold start or queue delay longer than 30 seconds. GitNexus then reported the response as "unparseable" because the body read was aborted by the timeout.

## Fix

- Raise the default per-request HTTP embedding timeout to 180 seconds.
- Add `GITNEXUS_EMBEDDING_HTTP_TIMEOUT_MS` for deployments that need a different value.
- Bound the override to 300 seconds to avoid indefinitely stalled indexing jobs.
- Include the resolved timeout in timeout errors.

## Verification

- Replayed the analyzer's real two-text batch against the configured endpoint: HTTP 200, two 4,096-dimensional vectors, completed in 66 seconds when the client timeout was extended to 180 seconds.
- `vitest run test/unit/http-embedder.test.ts`: 65 passed.
- `tsc --noEmit`: passed.
- Prettier check: passed.

This changes timeout handling only; batching remains unchanged.


azizur100389:
Reviewed and tested against current `upstream/main` (`ff86ccf1`). PR head `8540f248` merged cleanly in a fresh worktree.

Local validation:
- normal `npm ci` completed; it produced only local Node engine warnings because this machine is on Node `22.17.1` while the repo floor is `22.18.0`
- `npx vitest run test/unit/http-embedder.test.ts` passed: 65 tests

I do not see a correctness blocker in the timeout change. The timeout env var is bounded to positive integers up to the configured max, keeps the existing default, and is covered by the HTTP embedder tests.

The PR is still behind `main`, and I only see the conventional-label check plus the external Vercel authorization failure on GitHub, not the normal CI Gate/full CI suite for this head. Please update/rerun so the standard CI is visible and green before merge.

voidfreud:
Rebased and force-pushed onto current `main` at `e20b41ff`, so this PR is no longer behind. The current fork-PR CI runs are blocked as `action_required`; GitHub requires an upstream repository admin to approve them before they can execute. Please approve the pending workflow runs, then merge once the required checks are green.

voidfreud:
I’ve completed the requested fixes, rebased the affected branch, and added focused regression coverage where needed. The remaining blocker is GitHub’s upstream-only approval for fork workflow runs; I do not have permission to approve or rerun them. I’m stepping back until those workflows are approved or a concrete CI failure is available to address.

magyargergo (approved):

PR #2725, opened 2026-07-28
feat: make MAX_CALLABLE_VALUE_TARGETS configurable via env

ChunxueLi:
## Problem

The 32-target cap drops entire dispatch edges when a single callable site has 33+ candidates. This is a **cliff-edge data loss**: one extra candidate silently severs the entire call chain.

Observed on an observer/proxy framework where `ObserverFactory.execResult()` has 33 dispatch targets (19 modules × multiple overloads) — the entire dispatch graph was lost.

## Solution

Make the cap configurable via `GITNEXUS_MAX_CALLABLE_VALUE_TARGETS`, defaulting to 32 (unchanged behaviour).

## Test coverage

- Default value (32) when env not set
- Valid override (48)
- Invalid input fallback
- Below-1 guard

---
*Contributed by [卫宁健康科技集团](https://www.winning.com.cn) — WiNEX门诊医生站团队*


azizur100389:
Reviewed and tested against current `upstream/main` (`ff86ccf1`). PR head `b6e8c1ab` merged cleanly in a fresh worktree.

Local validation:
- built the worktree-local `gitnexus-shared` package
- `npx vitest run test/unit/scope-resolution/callable-value-flow-env.test.ts` passed: 4 tests
- Prettier check on the new test file passed

CI Gate is green, and the normal quality/typecheck/test lanes are green. The only remaining external status is the Vercel authorization failure. I do not see a correctness blocker in this change; the env override preserves the default cap unless a valid positive integer is supplied.

gitnexus/src/core/ingestion/scope-resolution/passes/callable-value-flow.ts
```diff
@@ -28,7 +28,17 @@ import {
...
-export const MAX_CALLABLE_VALUE_TARGETS = 32;
+/**
+ * Per-site dispatch-target cap. Above it the site is treated as overflowed and
+ * its edges are dropped — a cliff, so a repo with a legitimately wide dispatch
+ * table (33+ candidates on one callable site) loses the whole call chain.
+ *
+ * Override via `GITNEXUS_MAX_CALLABLE_VALUE_TARGETS` for such repos.
+ */
+export const MAX_CALLABLE_VALUE_TARGETS = (() => {
+  const env = Number(process.env.GITNEXUS_MAX_CALLABLE_VALUE_TARGETS);
```
magyargergo, line 40: Could you please document this in the readme plus the .env.example?

magyargergo (changes requested):

PR #2726, opened 2026-07-28
feat: make MAX_PROPERTY_DISPATCH_FANOUT configurable via env

ChunxueLi:
## Problem

The 32-registration cap was calibrated on GitNexus's own repo (16 providers × 2). Larger repos with different patterns hit the limit — e.g. a large Vue codebase where `validator` has 33+ component registrations is silently dropped.

## Solution

Make the cap configurable via `GITNEXUS_MAX_PROPERTY_DISPATCH_FANOUT`, defaulting to 32 (unchanged behaviour).

## Test coverage

- Default value (32) when env not set
- Valid override (48)
- Invalid input fallback
- Below-1 guard

---
*Contributed by [卫宁健康科技集团](https://www.winning.com.cn) — WiNEX门诊医生站团队*

### DIFF
--- a/gitnexus/src/core/ingestion/scope-resolution/passes/property-dispatch.ts
+++ b/gitnexus/src/core/ingestion/scope-resolution/passes/property-dispatch.ts
@@ -52,8 +52,15 @@ import { findCallableBindingInScope } from '../scope/walkers.js';
  * 2× that — dropping the motivating key was the failure the first value (8)
  * had. ponytail: flat cap; revisit with per-receiver narrowing if real
  * repos show useful keys being dropped (§12 of the #2437 plan).
+ *
+ * Override via `GITNEXUS_MAX_PROPERTY_DISPATCH_FANOUT` env var for repos with
+ * legitimate high-fanout property keys (e.g. large Vue codebases where
+ * `validator` exceeds the default).
  */
-export const MAX_PROPERTY_DISPATCH_FANOUT = 32;
+export const MAX_PROPERTY_DISPATCH_FANOUT = (() => {
+  const env = Number(process.env.GITNEXUS_MAX_PROPERTY_DISPATCH_FANOUT);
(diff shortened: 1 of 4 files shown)


azizur100389:
Reviewed and tested against current `upstream/main` (`ff86ccf1`). PR head `c282297d` merged cleanly in a fresh worktree.

Local validation:
- built the worktree-local `gitnexus-shared` package
- `npx vitest run test/unit/scope-resolution/property-dispatch-fanout-env.test.ts` passed: 4 tests
- Prettier check on the new test file passed

CI Gate is green, and the normal quality/typecheck/test lanes are green. The only remaining external status is the Vercel authorization failure. I do not see a correctness blocker in this change; the env override keeps the existing default and rejects malformed or non-positive values back to the safe default.

gitnexus/src/core/ingestion/scope-resolution/passes/property-dispatch.ts
```diff
@@ -52,8 +52,15 @@ import { findCallableBindingInScope } from '../scope/walkers.js';
...
  * had. ponytail: flat cap; revisit with per-receiver narrowing if real
  * repos show useful keys being dropped (§12 of the #2437 plan).
+ *
+ * Override via `GITNEXUS_MAX_PROPERTY_DISPATCH_FANOUT` env var for repos with
+ * legitimate high-fanout property keys (e.g. large Vue codebases where
+ * `validator` exceeds the default).
  */
-export const MAX_PROPERTY_DISPATCH_FANOUT = 32;
+export const MAX_PROPERTY_DISPATCH_FANOUT = (() => {
+  const env = Number(process.env.GITNEXUS_MAX_PROPERTY_DISPATCH_FANOUT);
```
magyargergo, line 61: Could you please document this in the readme plus the .env.example?

magyargergo (changes requested):

PR #2728, opened 2026-07-28
fix(ci): stop CI Report dying silently when the tests job fails

ivangegovdve-sudo:
## Summary

`ci-report.yml`'s **Build report** step exits 1 with no log output whenever the CI `tests` job fails, so the sticky PR comment is never posted and the **CI Report** run goes red. It accounts for **32 of the last 100 failed workflow runs** in this repo.

## Motivation / context

The step runs under `bash --noprofile --norc -e -o pipefail` and located its inputs with:

```bash
UNIT_SUMMARY=$(find "$DIR/test-reports" -name "coverage-summary.json" -type f 2>/dev/null | head -1)
```

The failure chain:

1. A test shard fails → the `tests` job fails.
2. `coverage-merge` (ci-tests.yml) is `needs: tests` with **no `if: always()`**, so it is skipped.
3. Its `Upload test reports` step never runs → the `test-reports` artifact does not exist. The CI Report log shows `##[warning]Artifact "test-reports" not found`.
4. `find` runs against a directory that does not exist and exits **1**.
5. `-o pipefail` carries that status through `| head -1`; the command substitution hands it to the assignment; `-e` kills the step.

The death is invisible: `2>/dev/null` discards `find`'s error and the entire report is built inside `{ ... } >> "$GITHUB_OUTPUT"`, so stdout never reaches the log. All you see is `Process completed with exit code 1`. `Comment on PR` is then skipped.

The net effect is that the report is dropped **precisely on the PRs whose tests failed** — the ones where it is most useful. The `⚠️ Coverage data unavailable` fallback already existed for exactly this case, but was unreachable: the script died ~160 lines before it.

Correlation across sampled runs was total — `TESTS: failure` → CI Report failed (6/6); `TESTS: success` → CI Report passed (4/4).

## Areas touched

## Scope & constraints

**In scope**

- Route the four artifact lookups in `Build report` through a `find_first` helper that returns empty when the root directory is absent, so the step degrades into the existing coverage-unavailable fallback instead of dying.
- Fix the one other live instance of the same shape, found by a follow-up sweep of `.github/` (see Implementation notes): `build-tree-sitter-prebuilds.yml`'s `Build prebuild` step, where the `find` pre-empts the `::error::prebuildify produced no .node` assertion directly beneath it.

- **Making `coverage-merge` run when `tests` fails** (`if: always()`). That would let the report name the individual failing tests instead of only saying coverage is unavailable, but it adds a full merge + web + docker run to every failing PR and its behaviour with partial shard blobs needs its own validation. Worth a follow-up; deliberately not bundled here.
- The individual PR test failures currently red on other branches (`this-boundary`, `rust-captures-golden`, `spring-bean-extractor`, `java.test.ts`, `callable-value-flow-env`) — those are branch-local product bugs, not CI defects.
- The `tree-sitter-grammars` Dependabot `ERESOLVE` (`tree-sitter-c-sharp@0.23.5` wants peer `tree-sitter@^0.25.0`, root pins `0.21.1`) — separate dependency decision.

## Implementation notes

```bash
find_first() {
  local root=$1 name=$2
  [ -d "$root" ] || return 0
  find "$root" -name "$name" -type f 2>/dev/null | head -1 || true
}
```

A sweep of every `find` in `.github/` found one more live instance of the same
pathology, now fixed in this PR:

```yaml
# build-tree-sitter-prebuilds.yml — `Build prebuild` (shell: bash)
rm -rf "$pkgdir/prebuilds"                     # ← 8 lines earlier
( cd "$pkgdir" && npx --no-install prebuildify --napi --strip )

out=$(find "$pkgdir/prebuilds" -name '*.node' -print -quit)
test -n "$out" || { echo "::error::prebuildify produced no .node"; exit 1; }
```

If prebuildify emits nothing without exiting non-zero, the directory `rm -rf`
deleted is never recreated, `find` exits 1, and `-e` kills the step before the
`test -n` line — the assertion written for exactly that case never runs. No pipe
here, so `-e` alone does it; `pipefail` is not needed. `|| true` restores the
diagnostic. The step still fails, which is correct — it just says why now.

The remaining `find` uses are safe, and were checked rather than assumed:

| Site | Why it is safe |
|---|---|
| `ci-tests.yml:322` | root is `.`, which always exists |
| `ci-tests.yml:352` | already guards with `\|\| true` (the vendor-hygiene check) |
| `gitnexus-skill-evolution.yml:276` | that workflow declares no `shell:` and no `defaults:`, so it runs GitHub's default `bash -e {0}` **without** `-o pipefail`; the pipeline takes `tail -1`'s status, which is 0 |
| `gitnexus-review-agent.yml:761,967` | process substitution — the status never reaches `-e` |
| `gitnexus-review-agent.yml:975` | root is created by an earlier step in the same job, and no fallback sits below it |

## Testing & verification

I extracted the `Build report` step body and ran it under the real shell flags (`bash --noprofile --norc -e -o pipefail`) against both artifact layouts.

| Scenario | Before | After |
|---|---|---|
| `test-reports` **absent** (tests failed) | `EXIT=1`, `$GITHUB_OUTPUT` **0 bytes** | `EXIT=0`, valid report (528 bytes) with `❌ Tests` + coverage-unavailable notice |
| `test-reports` **present** (tests passed) | `EXIT=0`, 1335 bytes | `EXIT=0`, **byte-identical** (`diff` clean) — full coverage table, deltas, bars, skipped-test details |

`find_first` was also diffed against the original expression across five directory states (missing / present / empty / deeply nested / multiple matches): **output identical in every case**; the only change is `rc=1 → rc=0` when the root is missing, which is the bug.

For the `build-tree-sitter-prebuilds.yml` fix, the same extract-and-run method under `bash --noprofile --norc -e -o pipefail`:

| Scenario | Before | After |
|---|---|---|
| `prebuilds/` missing | step dies at the `find`, `::error::` never printed, bare `exit 1` | `::error::prebuildify produced no .node`, then `exit 1` |
| `prebuilds/<tuple>/x.node` present | `out` and `produced` resolve, arch assertion passes | identical |

`shellcheck -s bash` on the extracted `Build prebuild` body: clean.

Repo gates run locally on this branch:

- [x] `npx prettier --check .github/workflows/ci-report.yml` — pass (and the `lint-staged` pre-commit hook passed)
- [x] `python .github/scripts/check-workflow-concurrency.py .github/workflows` — pass
- [x] `cd gitnexus && npx tsc --noEmit` — pass
- [x] `cd gitnexus && npm ci && npm run build` — pass

`actionlint`/`zizmor` were not run locally (not installed); Workflow Lint will cover them on this PR since it touches `.github/**`.

## Risk & rollout

Low. The change is confined to input discovery in a reporting workflow; it cannot affect build, test, or publish paths. On the success path the emitted report is byte-identical, so healthy PRs see no change. On the failure path CI Report goes from red-and-silent to green-with-a-report.

No index refresh, migration, or breaking change.


azizur100389:
Reviewed against current `upstream/main` (`ff86ccf1`). PR head `fedd3aac` fast-forwarded cleanly in a fresh worktree.

I do not see a code-level blocker in the workflow change. The artifact values are validated before being written to `GITHUB_OUTPUT`, and the missing/invalid metadata paths now fail visibly instead of letting the report job die silently.

I did not run a unit test because this is a workflow-only change. The remaining blocker is CI visibility: this PR currently shows only the conventional-label check plus the external Vercel authorization failure; I do not see the normal CI Gate/full CI suite for this head. Please get the standard CI run visible and green before merge.

PR #1904, opened 2026-05-29
feat(group): extract OpenFeign @RequestLine consumer contracts

henry201605:
Adds Java HTTP plugin support for the native OpenFeign annotation `@RequestLine("METHOD /path")`. Previously only `@FeignClient` interfaces using Spring MVC method annotations (`@GetMapping` etc.) were detected; the native annotation form — required by Feign Builder users and non-Spring Feign deployments — was silently ignored.

## Why

OpenFeign’s `@RequestLine` is the canonical annotation for the framework (https://github.com/OpenFeign/feign#interface-annotations) and is the only option for clients built without Spring Cloud OpenFeign. Real-world Java microservices that mix Spring Cloud and plain Feign Builder clients (e.g. internal `client-jar` libraries shared across non-Spring services) fall back to manifest-link workarounds today because GitNexus produces no consumer contract for these methods.

## What changed

- `FEIGN_REQUEST_LINE_PATTERNS` — tree-sitter query covering both positional and named-arg (`value =`) forms.
- `parseRequestLine()` — splits the verb + path string and drops any query portion, matching how `RestTemplate`/`WebClient` inline URL consumers are normalized.
- The enclosing interface MUST carry `@FeignClient`; otherwise the detection is dropped to avoid false positives from same-named annotations in unrelated libraries.
- Reuses the existing `feignPrefixByInterfaceId` map so `@FeignClient(path=)` and `@RequestMapping` interface prefixes apply uniformly across both Spring MVC and `@RequestLine` methods.
- Confidence `0.75` — slightly higher than the `0.7` used for Spring MVC annotations because the verb is a string-literal value (less ambiguous than annotation-name inference).

## Tests

Six new unit tests under `extracts OpenFeign clients` block:

| Scenario | Asserts |
|---|---|
| basic two methods | both `POST /ai/summarize` and `GET /ai/health` extracted with `framework: openfeign`, `confidence: 0.75` |
| `@FeignClient(path = "/api")` prefix joining | `/api/orders/{param}` for both GET and DELETE |
| query string stripping | `GET /search?q={query}&limit={limit}` → `http::GET::/search` |
| no `@FeignClient` annotation | detection dropped (no false positive on plain interfaces) |
| mixed `@RequestLine` + `@GetMapping` on the same interface | both produced at their respective confidences (0.75 / 0.7) |
| named-argument form `@RequestLine(value = "...")` | extracted identically |

## Verification

- `npx tsc --noEmit` ✅
[...]

Issue #1876, opened 2026-05-28
JavaScript Index produces a Function and a Const for a Lambda Call


hugogu:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

Given code like this:

```javascript
    // Format export data
    const exportData = accountsList.map(account => ({
      code: account.code,
      name: account.name,
      type: account.type,
      direction: account.direction,
      description: account.description || '',
      isActive: account.isActive,
      systems: accountSystemsMap[account.id]?.join(',') || '',
    }))
```

It produces a Const named `exportData` and a Function called `exportData`. 

Although the lambda itself is technically a 'Function', but its semantic meaning is not. And this behaviour doubles the count of nodes which makes the graph hard to read and slow to load. 

### Expected behavior

One thing (a piece of code, one line) generates one node at most. In this case, a Function might be a better choice. 

### Actual behavior

Duplicated nodes are generated. 

### Steps to reproduce

1. clone a javascript code repo
2. Find a file node.
3. Check its links of Function and Const

magyargergo:
Could you please double check this in https://github.com/abhigyanpatwari/GitNexus/releases/tag/v1.6.6-rc.90?

hugogu:
@magyargergo I used this project of mine as a test case. https://github.com/hugogu/journal-maker

And the latest main seems still reporting both Function and Const for the sample code. Let me try to raise a PR for this. 

<img width="70%" alt="Image" src="https://github.com/user-attachments/assets/2ba7fc67-7bb2-482d-b376-dbaf0c762d97" />

magyargergo:
> @magyargergo I used this project of mine as a test case. https://github.com/hugogu/journal-maker
> 
> And the latest main seems still reporting both Function and Const for the sample code. Let me try to raise a PR for this. 
> 
> <img width="70%" alt="Image" src="https://github.com/user-attachments/assets/2ba7fc67-7bb2-482d-b376-dbaf0c762d97" />

Nice! I didn't do this for Typescript :/ good catch! Thank you.

PR #2729, opened 2026-07-28
fix(js): index CommonJS `exports.foo = function () {}` exports (#2723)

### LINKED ISSUE #2723: JS indexer skips exports.foo = function () {} — a CommonJS codebase's public API is missing from the graph
**Title:** JS indexer skips `exports.foo = function () {}` — a CommonJS codebase's entire public API is missing from the graph

### What happens

Functions assigned to an `exports` property are not indexed as symbols. Declared functions are. On a CommonJS codebase that exports everything this way, the symbols you most want impact analysis for — the shared module APIs — are exactly the ones missing.

```js
// indexed as a Function ✅
function executeAddDynamicLinkMS(data) { ... }

// not indexed ❌
exports.areVariablesValid = function (variables) { ... }
```

### Measured on a real repo

Firebase Functions backend, 1,003 JS files, indexed clean and up to date with HEAD:

```
exports.X = function / arrow   1277 occurrences
function X() declared           819 occurrences
Function nodes in the graph    1281
```

Sampled 8 exported function names (every 97th, so no cherry-picking) and looked them up:

```cypher
MATCH (n) WHERE n.name IN [
  'checkDirectMessage','deleteChatReference','getCommentsCollectionGroup',
  'getMessageDocRef','getRelatedCountryCodes','makeSuperReporter',
  'createChallengeExpiredTask','automaticCallerOfVerifyProductPurchaseAndroidMS'
] RETURN n.name, n.filePath
```

→ **0 rows.** 0/8 found.

Consequence: `impact({target: 'areVariablesValid'})` returns `Target not found`, though it is a validator used across the codebase — the single symbol whose blast radius matters most. `impact` works for `executeAddDynamicLinkMS` (declared), so the split is exactly along the declaration/assignment line.

### Expected

`exports.foo = function () {}`, `exports.foo = async function () {}`, `exports.foo = (a) => {}` and `module.exports = { foo }` register a symbol named `foo`, so `impact` / `context` / `rename` can reach it.

### Why it matters beyond one repo

This is the dominant export style in pre-ESM Node — Express apps, Firebase Functions, most Node services written before ~2020. On those codebases the graph indexes the internals and misses the interfaces.

### Environment

gitnexus 1.6.7, macOS 15 (arm64), Node 22.23.0.

magyargergo:
Fixes #2723.

## The problem

Functions assigned to an `exports` property were not indexed at all. On a CommonJS codebase the graph held every internal helper and missed the entire public API — `impact({target: 'areVariablesValid'})` answered `Target not found` for the one symbol whose blast radius mattered most.

Measured on a 600-file Firebase-Functions-shaped corpus before this change: **0 `Function` nodes, 0 `CALLS` edges**. That matches the reporter's numbers (0/8 sampled exports found, on all three versions they tried). After: **7,220 `Function` nodes and 3,600 `CALLS` edges**.

## Why it needed two fixes, not one

The gap had two halves, and closing either alone leaves the feature broken:

1. **`tree-sitter-queries.ts`** had `@definition.function` rules for every declaration form and every variable-binding closure form, but none for `assignment_expression` — so no `Function` node was ever created.

2. **The scope-resolution queries** had no `@declaration.function` for the shape either. Adding only (1) moves `impact` from "not found" to "found, zero callers", because call resolution reaches a definition through the scope declaration, not through the graph node.

Both layers now carry the rule, in JavaScript and TypeScript, for every value form (function, async, arrow, async arrow, generator).

## A regression this branch caught in its own review

The first commit passed its 9 targeted tests, the 10,706-test unit suite, 2,926 resolver tests and the pipeline golden — and still silently deleted a pre-existing edge:

```js
function dup(v) { return v; }
exports.dup = function (v) { return !v; };
function callIt(v) { return dup(v); }   // callIt -> dup vanished
```

Binding the exported name into the module scope is the point of the fix, but when the file *also* declares that name lexically, the scope holds two declarations, the name goes ambiguous, and the resolver drops the intra-module edge. Confirmed by rebuilding both states: present at the merge base, missing at the first commit. Every existing fixture exported a name declared exactly once, which is why the suites stayed green.

Commit 2 suppresses the CJS declaration in exactly that case, at the existing `array-callback.ts` seam (#1876) — a shared pure helper both capture emitters apply at the current filter. The module-scope name set is computed once per file and memoized per program root, not walked per export; a 1000-export module is exactly the reported shape, and the per-export walk would be quadratic there.

## The rest of the CommonJS surface

The first version of this PR listed several known limitations. All of them are now closed:

| form | behaviour |
|---|---|
| `exports.foo = fn`, `module.exports.foo = fn` | declares `foo`; importers resolve |
| `const e = exports; e.foo = fn` | alias resolved per file |
| `this.foo = fn` at module level of a **CJS** file | export (`this` is `module.exports`) |
| `this.foo = fn` in a constructor | `Method` owned by the constructor |
| `Foo.prototype.bar = fn` | `Method` owned by `Foo` |
| `exports.fwd = lib.imported` | re-export, forwards to the original |
| `exports.foo = localFn` | already worked; needs nothing |

Three points worth review attention:

**The ESM gate is load-bearing.** Top-level `this` is `module.exports` in CommonJS but `undefined` in ESM, so `this.foo = fn` exports nothing there. Detection is deliberately asymmetric: an `import`/`export` statement settles the file as ESM immediately, a `require()` call or `exports`/`module` reference marks it CommonJS, and a file with **neither** signal is left alone — silence is not evidence of CommonJS.

**The receiver widening is only safe because the pruning is exact.** An `exports` alias cannot be pinned in a query, so the member-assignment rules match *any* identifier receiver. Without exact emit-side pruning, every `obj.handler = function () {}` in every JS/TS file would emit a spurious top-level `Function`. Both layers prune, and the tests assert the negative: `obj.notAnExport`, `self.alsoNot` and `localThing.nope` produce no node and no declaration.

**Re-export vs import was decided by measurement.** Forwarding was first emitted as `named-alias`, the shape destructured `require()` uses — and it did not resolve, because an import binding is private to its module exactly as in ESM. `reexport-alias` is what a CJS forwarding assignment actually is, and with it the call reaches the original definition.

Member ids qualify by owner (`Method:f.js:Foo.bar`); without that, two constructors in one file that each define `bar` collapse onto one node — the identity collapse #2699 fixed for function-local callables. Only the new path qualifies, so object-literal method ids are unchanged.

## Performance

Benchmarked A/B against the merge base. The only reproducible cost is tree-sitter query **compile**, paid once per worker process; steady-state emit cost and memory both sit below the measurement noise floor (a 600-file ESM corpus produces byte-identical graph output and times within run-to-run variance in both directions).

One commit is dedicated to that: folding the six CJS scope patterns per language into two via an inner leaf alternation. Five runs per build, variance under 1ms:

| query compile | base | before | after |
|---|---|---|---|
| JavaScript | 42.2ms | 50.6ms | 45.1ms |
| TypeScript | 116.5ms | 136.4ms | 123.3ms |

The full feature set lands at 46.5ms / 123.8ms — about +4ms and +7ms over base, one-time per worker process.

The alternation is deliberately the inner leaf form. tree-sitter 0.21.1 has a known hazard where a top-level `[...]` alternation makes sibling branches share one predicate bucket and silently drop matches with no compile error; it has bitten this repo twice (#1904, #1912). Here every predicate sits on a capture outside the alternation. Verified rather than assumed: a probe asserts all six receiver x RHS combinations still bind in both grammars, and that `exportz.x`, `module.other`, `Foo.prototype.bar`, `this.handler` and aliased `exports` are still rejected.

## Tests

`gitnexus/test/integration/cjs-exports-assignment.test.ts`, 24 tests: node emission for every export form plus TypeScript parity; call resolution for same-file, cross-file namespace, cross-file destructured `require()`, generators, aliases, module-level `this`, and re-export forwarding; owner edges and id-collision cases for prototype and `this` members; and the negative assertions above (non-exports receiver, ESM `this`, no-signal file, orphan twin, shadowing regression).

Each regression test was run against the pre-fix build and confirmed to fail there.

Green: 152 JS/TS integration tests, 1,299 scope-resolution unit tests, 10,706 unit tests, 2,926 resolver integration tests, pipeline golden.

Eleven unit failures on this machine are environmental and reproduce independently of these changes: a single CLI spawn takes 13-27s here (cold 16MB dist load, plus the NODE_OPTIONS heap re-exec) against `skip-git-cli`'s 10s `execSync` budget, and the CLI's own output and exit code are correct when run directly. The remainder (`run-analyze`, `pdg-mode-flip`, `fts-repair`) pass in isolation and fail only under parallel load.

## Still open

`module.exports = fn` — an anonymous default export has no name to bind, so nothing declares it. Documented in `languages/javascript/index.ts`.


xFlaviews:
crazy job @magyargergo!!

PR #2731, opened 2026-07-28
fix(ci): stop the review agent rejecting its own graph-backed reviews

magyargergo:
## Why

The review agent has failed 26 of its 349 runs (plus 1 cancelled, 4 never started). I pulled the logs for every one of them and classified the causes:

| class | n | what happened |
|---|---|---|
| `missing_graph_evidence` | 17 | a complete graph-backed review was thrown away by the evidence gate |
| publisher stale head | 4 | commits pushed mid-review, output correctly discarded |
| `action_required` | 4 | fork-PR approval, the run never executed |
| no `structured_output` | 3 | the model finished without the schema field |
| `npm ci` ECONNRESET | 1 | transient registry fault |
| invalid transcript envelope | 1 | one 95-turn swarm run |
| early rollout (`setup-node cache: false`, publisher token scope) | 1 | already fixed upstream |

The dominant class is self-inflicted. `contextEvidencePath()` only created a candidate when the model's own `mcp__gitnexus__context` call passed `file_path` (or `file`) exactly equal to a changed path. The review skill, the MCP docs and `CLAUDE.md` all teach plain `context({name: "symbol"})`, so the gate rejected the call shape it actually gets. [Run 30356808360](https://github.com/abhigyanpatwari/GitNexus/actions/runs/30356808360) is typical: a full APPROVE body with three LOW findings, `withRegistryLock` resolved through the graph, ~$3.88 of model spend, discarded — and the only log line was `Review rejected: no substantive exact-path GitNexus context result was recorded.`, which cannot distinguish "never called context" from "called it and I did not like the arguments".

## What changed

**The gate proves evidence from the result, not from the call arguments.** A context result that comes back `status=found` with a `symbol.filePath` inside the repo-scoped changed-path set proves the model queried the exact-SHA graph on changed code. That is the property the gate was always after; matching the input path added nothing, because the result had to match anyway. Everything else is untouched: strict-JSON result parsing, orchestrator-only turns (sidechain calls still cannot supply evidence), result-after-call ordering, duplicate tool-id rejection, and the `repo` argument still selects the head set or the merge-base set.

**Rejections now say why.** The failure path logs in-scope call count, out-of-scope calls (no selector or unknown repo), ignored sidechain calls, results that resolved nothing, results outside the changed paths, and up to three sanitized paths. The envelope error names the message count and the first-message shape. Without the out-of-scope counter, `in scope: 0` would still have been ambiguous.

**Glob and Grep leave the tool set.** They were enabled through `--tools` but deliberately never allow-listed, so under `dontAsk` every lane call was denied. Real runs show `permission_denials_count` up to 9, all turns spent on calls that could never succeed. `--tools` is now `Read,Agent`, the six personas drop them from their frontmatter, and `SKILL.md` no longer claims lanes have them. No capability is lost, since those calls were already denied.

**Both pinned `npm ci` installs retry three times.** The lock is pinned and `npm ci` rebuilds `node_modules` from scratch, so a retry can only refetch the identical integrity-checked tree.

**Prompt matches the new contract** and asks for the structured body even when a lane fails or the analysis is incomplete, which is the `no structured_output` class.

## Verification

`gitnexus/test/unit/review-agent-workflow.test.ts` executes the workflow's real inline Node gate through `runArtifactScenario`, so these are behavioural, not string matches. 45/45 pass, including new cases for: a name-only call accepted, a uid-only call accepted, a call with no selector rejected, a call against an unknown repo rejected, and the diagnosis counters for the off-path, sidechain, unknown-repo and zero-call paths. `tsc`, `eslint`, `prettier` and `shellcheck` over the `run:` steps are clean. `detect_changes` reports 0 changed symbols, low risk.

## Residual risk, deliberately left alone

- **A malformed context result still fails the whole review closed.** One unparseable tool result throws `invalid_execution_transcript` even if a later call already proved evidence. The transcript never reaches the publisher body, so this is availability strictness rather than a trust boundary, but it is pinned by an existing test with clear intent, so changing it belongs in its own PR.
- **Sidechain evidence is still refused.** A fully delegated run still fails the gate, held together by the prompt's "make the call yourself before dispatching any lane" rule. None of the 17 failures were delegation, so widening it is not justified yet.
- **The 4 `action_required` runs need a repository setting**, not a workflow change.
- **The model-behaviour half is unverified live.** The unit tests cover the gate; whether Sonnet now reliably emits a qualifying `context` call is only provable by the `workflow_dispatch` canary the workflow header's own activation checklist asks for. Worth running against a disposable PR before trusting this.


magyargergo (commented):
> **How to read this.** Automated multi-tool digest: two engines ran the same two review methodologies (Claude and Codex `gpt-5.6-sol`/xhigh each ran the GitNexus PR swarm and ce-code-review), plus a Codex-only adversarial pass — 15 lanes, then a separate synthesis agent and an independent critic gate. It reviewed head `0a22ff92`; the findings below have since been addressed in `0432214d` and `1d9f2d75`, and each is annotated with its state. Verify before acting on any of it.

# PR #2731 — multi-tool review digest (draft)

**Target:** `abhigyanpatwari/GitNexus` PR #2731 — `fix(ci): stop the review agent rejecting its own graph-backed reviews`
**Head:** `0a22ff92` · **Diff base:** `ff86ccf1e79cd7e4175da437ae8aeaf67b64aaa1` · **Scope:** 30 files, +216/−93
**Worktree checked:** `/workspace/.worktrees/pr-2731`
**Artifacts:** `/tmp/pr-tri-review/2731/20260728-143136-1132a447/`

---

## 0. Classifications, and what changed after this digest was drafted

**Merge state:** `checks pending`
**Branch hygiene:** `merge-from-main commit present but harmless and merge-safe`
**Final verdict (at the head this digest reviewed, `0a22ff92`):** `not production-ready`

This digest is a record of findings at `0a22ff92`. The author then pushed two commits that address them. **Every finding below is annotated with its current state; verify the fixes before merging rather than trusting this line.**

| Finding | State at `0a22ff92` | Now |
|---|---|---|
| H1 — one junk `context` result discards a proven review | open, P1, reproduced | **FIXED in `0432214d`** — payload-shape failures are caught and counted (`malformedResults`); transcript-structural invariants still throw. Re-running the same base/head harness against the fixed tree returns `failure_code: null` for all three adversarial transcripts. |
| H2 — incomplete review publishes as an accepted review | open, P1/P2 | **FIXED in `0432214d`** — the schema now requires a `complete` boolean; `complete: false` publishes the partial body under `failure_code: incomplete_analysis` instead of as an accepted review. |
| H3 — errored / missing / out-of-order results give an all-zero diagnosis | open, P2 | **FIXED in `0432214d`** — added `erroredResults`, `unusableResults`, `malformedResults`, and an unanswered-call count to the diagnosis line. |
| H4 — deletion-only PRs blame the deleted path | open, P3 | **FIXED in `0432214d`** — an empty eligible set is now out-of-scope rather than a candidate whose every result reads as off-path. |
| L9 (§4) — a bare File node satisfies the gate | open, P3, pre-existing | **FIXED in `1d9f2d75`** — evidence now requires a finite `startLine`/`endLine`, matching the trusted prescan's own definition of an indexable symbol. |
| §5 — grouped `Agent(a,b,c)` shattered by the pinned action's comma split | open, pre-existing, out of diff | **CHANGED in `0432214d`** — split into six `Agent(x)` rules. Correct under either reading of the runtime behaviour, which remains unproven; the header's dispatch canary is still the resolution vehicle. |

Test count moved 45 → 49 across those two commits; the drift guard stays green (52/52).

---

## 1. Engine breakdown and per-leg liveness

**Codex (gpt-5.6-sol/xhigh) — 3 legs, all live, all carrying real findings (no error markers):**

| Leg | File | State | Verdict |
|---|---|---|---|
| `codex-swarm` (GitNexus swarm, Solo mode, 6 lanes) | `codex-swarm.md`, 352 lines | Live. Ran `gh pr view/checks`, full diff, GitNexus `detect_changes`/`query`/`context`/`impact`/`explain`, and fetched the pinned action source from GitHub. Lane 7 intentionally skipped. | `not production-ready` → REQUEST CHANGES |
| `codex-ce` (ce-code-review, report-only) | `codex-ce.md`, 253 lines | Live. 7 personas (correctness, project-standards, testing, agent-native, security, reliability, adversarial) + a validator batch (2 checked, 2 validated, 0 dropped). Cross-model pass NOT run (strict read-only prohibited its temp artifacts); local adversarial fallback ran instead. | `Not ready` |
| `codex-adversarial` (native red-team) | `codex-adversarial.md`, 188 lines | Live. The only leg that **executed the repo's own suite**: `npx vitest run test/unit/review-agent-workflow.test.ts` → 45/45 pass; `test/unit/shipped-skills-sync.test.ts` → 52/52 pass. `git diff --check` clean, bidi scan clean. | `approve`, **no material findings** |

Two codex-adversarial command results deserve an honest note: `npx tsc --noEmit` exited 2 and `npm test` exited 130 (SIGINT — the run was interrupted, not a failure). Neither was converted into a finding by any lane, and CI's typecheck on head is green, so these read as worktree/environment artifacts rather than defects. Flagging so nobody later mistakes them for suppressed signal.

**Claude — 12 lanes, all live:**
- Swarm (5): `risk-architect`, `security-boundary`, `test-ci`, `docs-dod`, `branch-hygiene`.
- CE personas (7): `correctness`, `adversarial`, `security`, `testing`, `maintainability`, `standards`, `reliability`.
- Three of these (`ce-adversarial`, `ce-correctness`, `ce-security`) built execution harnesses that extract the workflow's `Assemble bounded review artifact` Node heredoc — the same technique the repo's own tests use — and ran crafted transcripts against **both** the base and head scripts. `ce-standards` and `ce-testing` ran the real vitest suite (97/97 and 45/45).

**Engine disagreement worth stating up front:** Codex's two report-producing legs both land on *not ready*; Codex's adversarial leg lands on *approve with no findings*; the Claude swarm lands on *not blocking*; the Claude CE adversarial/correctness/security trio surface a P1 that **no Codex leg found**. This is not consensus — it is three different search strategies hitting different parts of the surface.

`prior-context.md` was used only to flag recurrence, never as a finding on its own.

---

## 2. What the PR gets right (validated, not assumed)

These were actively probed and held. Credit where the lanes did the work:

1. **The core gate rewrite is correct and security-sound.** Moving proof from the call arguments to the MCP *result* (`.github/workflows/gitnexus-review-agent.yml:1659-1684`) fixes the real reported failure without weakening the boundary: `symbol.filePath` comes from the bwrap-sandboxed, read-only, exact-SHA GitNexus MCP; the transcript is written by the SHA-pinned Claude binary; the model can emit `tool_use` blocks but cannot author `tool_result` content. Repo scoping is preserved exactly — head-or-absent `repo` → `headPaths`, merge-base `repo` → `baseEvidencePaths`, anything else → ineligible (`:1675-1683`), and `base_prescan_paths` remains excluded from evidence. *Corroborated by both engines:* codex-swarm §7 + codex-adversarial (approve) + claude-swarm-risk-architect + claude-swarm-security-boundary + claude-ce-security R1 + claude-ce-maintainability.
2. **Trust boundary untouched.** The diff touches neither `permissions:` nor `secrets.`. `analyze` keeps `contents: read` / `pull-requests: read` (`:145-147`); `publish` keeps no model secret and no checkout (`:2089-2092`); the only `secrets.` reference stays inside analyze (`:1267`). *Both engines:* codex-swarm Lane 5 + claude-ce-security R6 + claude-swarm-security-boundary F.
3. **Glob/Grep removal is complete and capability-narrowing.** `--tools "Read,Glob,Grep,Agent"` → `"Read,Agent"` (`:1345`) is the only `claude_args` line that changed; `--allowedTools`/`--disallowedTools` are byte-identical. I independently confirmed zero residual `Glob`/`Grep` strings anywhere under the review-skill trees, and that `SKILL.md` + all six personas are md5-identical across all four mirrors (`.claude/`, `gitnexus/skills/`, `gitnexus-claude-plugin/`, `gitnexus-cursor-integration/`). *Both engines:* codex-swarm Lane 6 + codex-ce agent-native + 5 Claude lanes.
4. **Both npm retry loops are mechanically correct** under `set -euo pipefail` (`:419-433`, `:523-537`) — see refutations below.
5. **The new diagnostics cannot pollute anything that crosses the boundary.** Sanitizer `[^\w./-]` → `?`, 3-sample × 200-char cap (`:1744-1746`), `console.error`-only (`:1983`), never in `artifact.body` or `graph_evidence`.
6. **The test harness is genuinely behavioural**, not string-matching: `runArtifactScenario` spawns the workflow's real inline Node validator. New coverage spans name-only, uid-only, selector-less, unknown-repo, base-only deletion/rename, sidechain, cross-wiring, malformed JSON, semantic errors, off-path, diagnostics, and the no-indexable mode.
7. **The removed `unrelated` test case is a correct supersession, not a silent coverage loss.** claude-ce-testing proved this empirically — it temporarily re-added the old case and confirmed it now returns `status: 'success'` under the new result-based contract, so keeping it would have been a *false* test. claude-swarm-test-ci reached the same conclusion independently, noting the old property is still covered by the `offPath` scenario at `test:1630-1637`.
8. **Hygiene:** no CHANGELOG edit (repo policy), merge-from-main commit `6c41149c` does not pollute the base-to-head diff, no generated files, no bidi controls, no stale/fabricated `#2731` references.

---

## 3. Headline findings

### H1 — P1 — One malformed or truncated `context` result now discards an entire, already-proven review

- **Corroboration:** **Strong by execution, single-engine by analysis.** `[reproduced]` — the coordinator independently ran the workflow's own embedded Node script on both trees: base `/workspace` → `failure_code: null`; head worktree → `invalid_execution_transcript`. Independently executed by three Claude lanes with their own harnesses: `claude-ce-adversarial` F1 (P1, conf 85), `claude-ce-correctness` F1 (P1, conf 60), `claude-ce-security` F1 (P2, conf 90). **No Codex leg reached this.** codex-adversarial explicitly searched for "transcript/artifact validation regression" and reported none — that is a non-detection, not a refutation, and it did not run a two-call transcript. Posted under the severity-overrides-corroboration rule: a P1 with a reproduced trigger.
- **Evidence (verified by me in the worktree):**
  - Candidacy widened: `contextEvidencePaths` (`:1666-1684`) no longer inspects `file_path`/`file` at all — any orchestrator `context` call with a nonempty `name`/`uid` and a default-or-known `repo` is now a candidate (`:1853-1855`).
  - Every candidate's result is fed to `contextResultProvesChangedPath` (`:1715-1748`) from `:1894`, and that function **throws** rather than returning `false` on four shapes: not-text (`:1712`), empty (`:1717`), not strict JSON (`:1726-1730`), and `validateBoundedJson` violations (`:1731`).
  - The throw escapes `proveGraphReview()` and is caught at `:1964-1972`, setting `failure_code = 'invalid_execution_transcript'` and discarding the model's body — **even when an earlier call already proved evidence**.
  - Publisher consequence at `:2328-2329`: renders `### GitNexus review — unable to complete` — the exact public failure this PR exists to eliminate, re-entered through a different door.
- **Trigger, stated honestly (the lanes do not fully agree):** the MCP hard-caps every response at `GITNEXUS_MCP_DEFAULT_MAX_TOKENS=12000` (`:656`) and `applyMcpMaxTokens` truncates by raw byte prefix mid-JSON, appending `'\n…'` (`gitnexus/src/mcp/output-budget.ts:40-50`).
  - `claude-ce-correctness` measured against this repo's live index and found the largest `context` payload it could produce is **24,730 bytes** — *under* the 48,000-byte default — so the default-budget trigger is **plausible but not demonstrated** for GitNexus-sized symbols. That is why it scored 60, not 85.
  - `claude-ce-adversarial` measured `context({name, include_content: true})` on **File** nodes and reports 75 of 1,951 File nodes exceed 47,000 bytes, plus a 312,326-byte measured payload.
  - Both agree on the trigger that needs no unusual repo: `maxTokens` is a **model-settable parameter of the `context` tool** (`gitnexus/src/mcp/tools.ts:314-319`), so `context({name, maxTokens: 4000})` reproduces truncation at any size, and `--max-turns 150` gives the orchestrator ample opportunity.
  - Independent of truncation, the base→head delta is a code-read certainty: three lanes executed base-vs-head with a proving call plus a junk/non-JSON/oversized-array second call — base published, head returned `invalid_execution_transcript` in every case.
- **Worst sub-case:** this fires even on PRs the gate does not apply to. With `no_indexable_changed_symbols = true` (a docs-only PR allowed to finish with zero context calls), one junk `context` result still yields `invalid_execution_transcript`, because the throw happens inside the transcript pass before the bypass is evaluated. Executed by both `ce-adversarial` (scenario A/B) and `ce-correctness`.
- **Suggested fix (three lanes converge on the same shape):** keep the hard `throw` for transcript *structural* invariants (envelope shape, duplicate tool ids, tool-call shape) and make *result-payload* failures soft — `rejected.malformed += 1; return false` — surfaced in the same bounded diagnosis line. Strictly *more* conservative for the gate (an unparseable result can never count as evidence) while restoring the base property that an unrelated context call cannot void a proven review. Add the two-call regression test (§5).

### H2 — P1 (Codex) / P2 (Claude) — An incomplete review is published as a successful, graph-evidence-stamped review

- **Corroboration:** **Strong — cross-engine.** Codex: `codex-ce` finding #1, P1, reviewers *adversarial + reliability*, confidence 100, marked "independently validated". Claude: `claude-ce-adversarial` F2 (P2, conf 60) reached the same clause and the same publisher consequence independently; `claude-ce-security` names it as a residual risk on the same lines; `codex-swarm` §7 reaches the adjacent conclusion that the gate "does not prove that every review conclusion was causally grounded". Carried at **P1** per dedup-keeps-highest, with the honest caveat that the Claude corroborator rated it P2/60 and explicitly framed it as a *policy* question rather than a mechanical bug.
- **Evidence (verified by me):** the PR adds an unconditional output requirement at `:1335-1337` — *"Always end the run by returning that structured body field, even when a lane fails, a query comes back empty, or the analysis is incomplete."* The artifact assembler then accepts any single-key, nonempty `body` (`:1986-1995`) and sets `status = 'success'`, `failure_code = 'none'`, `graph_evidence.mode = 'context'` (`:1996-2005`). The publisher renders it with no completeness signal (`:2331-2332`); the footer carries only the analyzed SHAs. `[code-read]`
- **Trigger:** the orchestrator makes one resolving `context` call (required *before* any lane dispatch, `:1315-1320`) → dispatches six lanes → every lane fails (denied dispatch, MCP error, `maxTurns` exhaustion) → per `:1335` it must still emit a body. The artifact is byte-identical in shape to a full swarm review. `[reproduced]` for the gate half: `claude-ce-adversarial` executed S4/C showing the gate is satisfied by a single `context({name: 'README.md'})` on a changed doc file, and by one incidental resolving call on a 500-path PR.
- **Suggested fix:** add a machine-checkable completeness field alongside `body` (e.g. `lanes_completed`), fail the artifact unless it reports complete while preserving the partial body for diagnostic publication — or keep `:1335-1337` and require a publisher-detectable degradation banner rendered as `### GitNexus review — partial`.

### H3 — P2 — An in-scope candidate whose result errors, never arrives, or arrives out of order produces an all-zero diagnosis

- **Corroboration:** **Strong — cross-engine.** Codex: `codex-ce` finding #2, P2, reviewers *correctness + reliability*, confidence 100, "independently validated", cites `:1891`. Claude: `claude-ce-correctness` F2 (P3, conf 90) `[reproduced]` — executed three variants, all producing byte-identical all-zero diagnoses. `claude-ce-adversarial` F3 (P3, conf 80) `[reproduced]` found the sidechain-tagged-*result* variant of the same hole. Carried at P2.
- **Evidence (verified by me):** the result-side guard at `:1889-1897` short-circuits on `!sidechain`, `block.is_error !== true`, `candidate`, and `messageIndex > candidate.messageIndex` — **all before** `contextResultProvesChangedPath` is reached, and none of them touch `rejected`. The diagnosis string at `:1909-1916` therefore reports `in scope: 1` with every reason counter at 0. `[code-read]` + `[reproduced]`
- **Trigger (the realistic one):** the orchestrator issues the merge-base call the prompt asks for, `context({name, repo: '<merge-base path>'})`, against a graph not registered under that exact path → MCP returns `is_error: true` → review rejected with a diagnosis that explains nothing. This is precisely the failure class the diagnostics were added to explain.
- **Coverage note (verified):** the existing `failedQuery` scenario at `test:1595-1598` drives exactly this state (`resultIsError: true`) but asserts only `failure_code`, never `stderr` — so the all-zero output is invisible to the suite.
- **Suggested fix:** add `rejected.errorResults` (and a `sidechainResults` counter for the F3 variant), plus an `in-scope calls with no usable result` term so a missing/out-of-order result is distinguishable from an errored one. Assert them in the existing `failedQuery` scenario.

### H4 — P3 — Deletion-only PRs: the empty `headPaths` Set is truthy, so the diagnosis blames the deleted path itself

- **Corroboration:** **Moderate — two independent Claude lanes, same engine, both by execution.** `claude-ce-adversarial` F4 (P3, conf 80) and `claude-ce-correctness` F3 (P3, conf 85) `[reproduced]` — reached independently with separate harnesses. No Codex leg.
- **Evidence (verified by me):** `contextEvidencePaths` returns `changedPathManifest.headPaths` whenever `repo` is absent (`:1677-1679`) without checking emptiness; `if (eligiblePaths)` at `:1854` is a truthiness test, and `new Set()` is truthy. Under the base script the same call returned `undefined` and was simply not a candidate.
- **Observable outcome (executed by both lanes):** on a deletion-only PR the diagnosis reads `results outside the changed paths: 1 (gitnexus/src/cli/gone.ts)` — naming the *deleted path itself* as "outside the changed paths", pointing the operator at the path instead of at the real cause (the missing `repo: <runner.temp>/gitnexus-review-merge-base` argument). On a deletion-heavy PR every call reports this way.
- **Suggested fix:** return `undefined` when the selected set is empty, and add a distinct counter so the diagnosis can say "this PR has no head paths; merge-base calls must pass `repo`".

---

## 4. Lower-priority items

Each of these is single-lane or cosmetic; per the corroboration rule, P2/P3 items with one lane are recorded but not pushed as blockers.

| # | Sev | Item | Lane / engine | Tag |
|---|---|---|---|---|
| L1 | P2 | **Name-ambiguous `context` call fails the gate with no mid-run recovery.** If the graph resolves an ambiguous `name` to a symbol in an *unchanged* file, `offPath` increments and the gate fails after full model spend; the diagnosis only reaches stderr post-run. `:1666-1684`, `:1741-1747`. Strictly narrower than the failure it replaces (which fired unconditionally on 17/26 runs). Fix: prompt the `uid` form when the prescan reports indexable symbols. | claude-swarm-risk-architect (Claude) | `[code-read]` |
| L2 | P2 | **Neither `npm ci` attempt has a per-attempt timeout.** The loop bounds attempt *count*, not attempt *duration*; up to 6 npm invocations across two blocks share one `timeout-minutes: 75` (`:144`) with the `--max-turns 150` review call. No step-level timeout exists anywhere in the file. `:419-433`, `:523-537`. | claude-ce-reliability F1, conf 55 (Claude) | `[code-read]` |
| L3 | P2 | **Two byte-identical 17-line retry blocks** ~100 lines apart in the same job (`:416-433`, `:521-537`), differing only in the comment and two echo labels. Both introduced by this PR. Honest tradeoff noted by the lane: extracting to a script adds a new file to the supply-chain audit surface. | claude-ce-maintainability, conf 70 (Claude) | `[code-read]` |
| L4 | P3 | **`contextResultProvesChangedPath` kept a singular-path name** after its contract became set-membership plus a mutated out-parameter (`:1715`, called at `:1894`) — while the same diff correctly renamed its sibling `contextEvidencePath` → `contextEvidencePaths` for exactly that reason. | claude-ce-standards S-2, conf 50 (Claude) | `[code-read]` |
| L5 | P3 | **Test default `toolInput` still carries a now-ignored `file_path`** (`gitnexus/test/unit/review-agent-workflow.test.ts:265`). The gate never reads it; the default implies the opposite of the new contract. Verified. | claude-swarm-risk-architect (Claude) | `[code-read]` |
| L6 | P3 | **`SKILL.md:184` names a Claude Code built-in as the lanes' whole file-access capability** in a spec AGENTS.md declares CLI-neutral — and the sentence ships to the Cursor and npm mirrors. Verified: the line reads "restricted to Read plus the safe graph tools". Documentation drift only; enforcement at `:1345` and in persona frontmatter is correct. | claude-ce-standards S-1, conf 55 (Claude) | `[code-read]` |
| L7 | P3 | **Merge-from-main commit `6c41149c`** pulls 125 unrelated files into the branch log (hidden from the PR diff). Reviewer load, not risk; optional rebase. | claude-swarm-branch-hygiene (Claude) | `[code-read]` |
| L8 | — | **Pre-existing, out of diff:** for early rejection codes that leave `authorized == 'false'` (`invalid_pr_number`, `invalid_control_sha`, `actor_not_authorized`), the entire `publish` job is skipped — including its `if: always()` marker-removal step — leaving the "🔄 review in progress" comment stuck permanently. The lane confirmed **none** of this PR's new failure paths can reach it. | claude-ce-reliability (Claude) | `[code-read]` |

---

- **L9 — P3 — pre-existing advisory — the gate accepted a bare File node as evidence.** `claude-ce-adversarial` F5 (conf 70) *executed* `context({name: 'AGENTS.md'})` and had the result accepted at `:1742`, because the check was only `eligiblePaths.has(resolvedPath)` with no line-range requirement — while the trusted prescan defines an indexable symbol as one having `startLine` and `endLine`. The lane states plainly that this PR does not create the hole; it is the same "what counts as proof" surface the PR tightens elsewhere. Raised here because it was filed and reproduced by a lane, not merely speculated. **Fixed in `1d9f2d75`.**

## 5. The `allowedTools` comma-split — confirmed parse, unproven runtime effect, out of this diff

This is the loudest cross-engine disagreement in the run and needs to be represented precisely.

- **What is confirmed:** the pinned action at SHA `3553f843` parses `allowedTools` as `allowedToolsValues.split(ACCUMULATE_DELIMITER).flatMap((v) => v.split(","))` (`base-action/src/parse-sdk-options.ts`). The coordinator read this directly; `codex-swarm` and `codex-ce` fetched the same source independently. So the grouped rule `Agent(ci-correctness-lens,…,ci-critic-lens)` at `:1346` **is** split into fragments (`Agent(ci-correctness-lens`, four bare names, `ci-critic-lens)`).
- **What is NOT proven:** the runtime consequence — whether `--permission-mode dontAsk` actually denies `Agent` dispatch when no complete allow rule matches. One production run showed `permission_denials_count: 0` while the model reported dispatching lanes. No leg closed this.
- **Whether it belongs to this PR:** **no.** I verified the diff: only `--tools` changed (`--tools "Read,Glob,Grep,Agent"` → `"Read,Agent"`); the `--allowedTools` line is a pure context line in the diff, byte-identical to base.
- **Engine positions:** `codex-swarm` Lane 3 rates it **P1, blocks merge: yes** and drives its REQUEST CHANGES verdict. `codex-ce` files the *same* issue under "Pre-existing Issues" (#3) and states explicitly that it "does not determine this verdict". Every Claude lane that touched it (`ce-security` R3, `swarm-docs-dod` item 8, `ce-standards`) classifies it pre-existing and out of scope; `ce-security` adds that Glob/Grep remain denied either way, since deny beats allow and `--tools` only narrows the built-in set.
- **Also confirmed (by codex-swarm Lane 4, verified by me):** the unit test at `test:1153-1154` and `test:1178-1180` extracts the raw YAML string and asserts the *grouped* text is present — it validates the pre-parser text, not the effective post-split array. So the test would not catch a dispatch break either way.
- **Recommended framing:** not a merge blocker for #2731. The activation checklist's canary (`:9-17`) — positive six-lane dispatch plus refusal of an unlisted `Agent(<type>)` — remains the only mechanism that can settle it, and it is already an unchecked, acknowledged obligation before `GITNEXUS_REVIEW_COMMENT_ENABLED` is flipped.

---

## 6. Notable refutations — attacks constructed and executed that did **not** break

Validation is a result. These were probed, several by execution, and held:

| # | Attack / concern | Verdict | Lanes (engines) |
|---|---|---|---|
| R1 | `npm ci … && break` misbehaves under `set -euo pipefail` — masks a hard failure or trips errexit early | **REFUTED by execution.** A command that is not the final member of an `&&` list is exempt from errexit; the `if` condition is exempt; the last command per iteration is `sleep`, which succeeds. Simulated all three outcomes (succeed 1st / succeed 3rd / never): retries correctly, exits 1 after 3 attempts. | claude-swarm-risk-architect, claude-ce-correctness, claude-ce-security R4, claude-ce-maintainability, claude-ce-reliability (all Claude) |
| R2 | The gate relaxation is a security regression | **REFUTED.** `symbol.filePath` was already the load-bearing value pre-PR; the PR widens the accepted *set*, not the trust source. The model cannot forge MCP results, `Write`/`Bash` are denied at `:1347`, and the transcript is written by the SHA-pinned binary. | codex-swarm Lane 5, codex-adversarial (Codex) + claude-swarm-risk-architect, claude-swarm-security-boundary, claude-ce-security R1 (Claude) — **cross-engine** |
| R3 | Lane (sidechain) activity laundered as orchestrator activity via the newly added `continue` | **REFUTED by execution.** `continue` at `:1851` is the last statement of the block body, exactly equivalent to the base `&& !sidechain` guard; `seenToolCalls.add(block.id)` happens before it, so duplicate-tool-id detection still fires. A sidechain `tool_use` never enters `candidateCalls`; a candidate's result inside a sidechain turn is rejected at `:1890`. | claude-ce-adversarial R1, claude-ce-correctness (Claude) |
| R4 | The new diagnostics leak, inject, or grow unbounded | **REFUTED.** `[^\w./-]` → `?` strips newlines, `:`, `%`, ESC and all non-ASCII, so no Actions workflow-command or ANSI injection; 3 samples × 200 chars; `console.error`-only, never in `artifact`, `GITHUB_OUTPUT`, or the published comment. | claude-ce-security R1/R2, claude-swarm-security-boundary B/E, claude-ce-adversarial R2, claude-ce-standards (Claude) |
| R5 | `--tools "Read,Agent"` widens something | **REFUTED.** `--tools` is the *restricting* flag; dropping two built-ins cannot add any. `--allowedTools`/`--disallowedTools` are byte-identical to base (I verified), Glob/Grep were never allow-listed, and deny beats allow. | claude-ce-security R3 (Claude), codex-ce agent-native (Codex) — **cross-engine** |
| R6 | Skill mirrors drift, or a lane instruction still needs Glob/Grep | **REFUTED.** Zero residual `Glob`/`Grep` anywhere in the workflow or the skill trees; all four mirrors byte-identical (I independently md5-verified SKILL.md + personas); the shipped-skills drift guard passed on the PR and codex-adversarial ran its 52 tests green. | codex-swarm Lane 6, codex-adversarial, codex-ce (Codex) + 5 Claude lanes — **cross-engine** |
| R7 | Removing the `unrelated` test case is a silent coverage loss | **REFUTED empirically.** claude-ce-testing re-added the old case and confirmed it now passes under the result-based contract — keeping it would have been a false test. The old property survives at `test:1630-1637` and `test:1691-1696`. | claude-ce-testing, claude-swarm-test-ci, claude-ce-standards (Claude) |
| R8 | `validateBoundedJson`'s `MAX_ARRAY_ITEMS = 2000` is reachable from a real `context` result | **REFUTED in practice** by claude-ce-adversarial R6 — the mechanism works but `context` caps every ref bucket at `LIMIT 30` (`gitnexus/src/mcp/local/local-backend.ts:3267…`), so only the byte-budget truncation of H1 is reachable. **Partial dissent:** claude-ce-security notes `processes` has no `LIMIT` (`local-backend.ts:3395-3401`), which would make an oversized payload cheaper to construct. Unresolved between two Claude lanes; either way it lands in H1's bucket. | claude-ce-adversarial vs claude-ce-security (Claude) |
| R9 | Base/head evidence-set confusion — a merge-base call proving a modified file, or `base_prescan_paths` becoming eligible | **REFUTED by execution.** `basePrescanPaths` is read but never returned; all three sets are reconstructed and cross-checked from `entries` (`:1583-1649`). | claude-ce-adversarial R7, claude-ce-correctness (Claude), codex-swarm Lane 5 (Codex) — **cross-engine** |
| R10 | Real MCP `not found` / `ambiguous` responses throw instead of counting as unproven | **REFUTED by execution.** Both parse cleanly and land in `rejected.unresolved` → `missing_graph_evidence` with `results that resolved nothing: 1`. | claude-ce-adversarial R5, claude-ce-correctness (Claude) |
| R11 | `npm ci` retry bypasses lock pinning or integrity verification | **REFUTED.** SHA-512 receipt walk runs *before* the loop (`:385-411`); `npm ci` re-verifies integrity every attempt and wipes `node_modules` first; the offline Bubblewrap activation and the `sha256sum`/`--version` binary pins all still follow the loop. | claude-ce-security R4, claude-swarm-security-boundary C, claude-ce-reliability (Claude) |
| R12 | Job permissions or secret placement changed | **REFUTED.** The diff touches neither. | claude-ce-security R6 (Claude), codex-swarm Lane 5 (Codex) — **cross-engine** |
| R13 | Hostile-fork escalation, publisher reachability flaw, hidden Unicode/bidi, stale `#2731` references | **REFUTED.** codex-adversarial found no exact-SHA evidence-gate bypass after running the real 45-test + 52-test suites; bidi scan clean, `git diff --check` clean, no `#NNNN` in any added line. | codex-adversarial, codex-swarm (Codex) + claude-swarm-security-boundary, claude-ce-adversarial R8, claude-ce-standards (Claude) — **cross-engine** |
| R14 | Deletion-only / rename-only / no-indexable-symbol PRs are newly broken | **REFUTED as a *new* failure mode** — deleted paths land in `baseEvidencePaths`, renames split old/new across base/head, symbol-less PRs bypass via `no_indexable_changed_symbols`. (The diagnosis *wording* for deletion-only PRs is still wrong — that is H4, a separate, narrower issue.) | claude-swarm-risk-architect, claude-ce-testing (Claude) |
| R15 | The "Always end the run…" clause is a hang/loop risk | **REFUTED.** Advisory text only; no new control flow. `--max-turns 150` and per-lane `maxTurns: 12` unchanged; body bounded by `maxLength: 50000`. (The *publication* consequence is H2 — a different concern.) | claude-swarm-risk-architect (Claude) |

---

## 7. Test gaps

Ordered by how directly they map to a finding above.

1. **No multi-call transcript anywhere in the suite** — the H1 blind spot. `reviewTranscript()` (`test:263-327`) is the only site that constructs a `tool_use` block, so every scenario has at most one context call. A fixture of shape `[proving call, malformed call] → status: 'success'` would have caught H1 outright. *Cross-lane:* claude-ce-adversarial gap 1, claude-ce-correctness gap 1, claude-ce-security gap 1, claude-ce-testing F3.
2. **No MCP-truncation fixture** — nothing feeds a result of shape `<valid JSON prefix> + '\n…'` with the hint marker absent, which is exactly what `applyMcpMaxTokens` produces at the workflow's own `GITNEXUS_MCP_DEFAULT_MAX_TOKENS=12000`. The gate's strict-JSON requirement and the server's silent truncation are never tested together.
3. **Retry loops asserted only as string presence, never executed** — `test:630-631` and `test:708-709` check `toContain('for attempt in 1 2 3; do')` and the failure message. A dropped `&& break`, an off-by-one on `-ge 3`, or `exit 1` moved outside the loop all still pass. *Strongest cross-engine test gap:* codex-swarm Lane 4 P3 + codex-ce coverage + claude-ce-testing F2 + claude-ce-reliability + claude-ce-correctness gap 5.
4. **No scenario proves incomplete structured output remains a failure** (H2) — there is no lane-completeness signal to assert on. *codex-ce.*
5. **Failed / missing / out-of-order context-result diagnostics unasserted** (H3) — I verified `grep -c 'resolved nothing'` in the test file returns **0**, so the `unresolved` counter is never asserted non-zero; and `failedQuery` (`test:1595-1598`) asserts only `failure_code`, never `stderr`. *claude-ce-testing F1 + claude-ce-correctness gap 3 + codex-ce.*
6. **Weak assertion at `test:1621`** — `expect(uidOnly.artifact.failure_code).toBeNull()` where the parallel `nameOnly` case at `test:1608-1612` uses `toMatchObject({status, failure_code, graph_evidence})`. A regression setting `status: 'failure'` with `failure_code: null` slips through. *claude-swarm-test-ci.*
7. **Sidechain result-side attribution untested** — `test:1639-1642` tags *both* messages, so the call-side counter fires first and the result-side `!sidechain` drop at `:1890` is never exercised (H3's F3 variant).
8. **No deletion-only (`head_paths: []`) artifact scenario** (H4).
9. **Non-string `symbol.filePath` untested** — if a future edit dropped the `typeof === 'string'` guard at `:1744`, a non-string value would throw inside `.replace()`, converting a controlled reject into an uncaught exception. *claude-ce-testing F4.*
10. **Diagnostic sanitizers untested on hostile input** — the `[^\w./-]` scrub is only exercised on a clean path; the envelope-shape `label()` string (`:1769-1777`) has no test at all. *claude-ce-security, claude-swarm-security-boundary.*
11. **`--tools` asserted only as a literal string** — nothing verifies the runtime consequence that Glob/Grep are actually unavailable; the activation canary remains the only mechanism.

---

## 8. CI status (head `0a22ff92`)

All quality, security, build, and coverage checks green (workflow lint, zizmor, CodeQL, gitleaks, typecheck, formatting, shipped-skills drift guard, most test shards). Three failures, none implicating this diff:

- `tests / macos-latest (platform-sensitive) 3/3` — unrelated flake in `test/integration/analyze-index-lock-concurrency.test.ts`, a file this PR does not touch.
- `tests / windows-latest (platform-sensitive) 1/3` — `vitest timed out after 20 minutes`; environmental.
- `Vercel` — fork preview-deploy authorization, not code.

A rerun of the failed shards was triggered. `mergeStateStatus: BLOCKED` reflects the pending checks and the absent approval, not a conflict — the supplied base equals the calculated merge-base and the branch is mergeable once checks land.

---

## 9. Suggested disposition

**Request changes on H1** — it is a reproduced regression that re-creates, through a different door, the exact class of spurious rejection this PR exists to eliminate, and it can void reviews on PRs the gate does not even apply to. The fix is small and strictly more conservative than the current behaviour: soft-reject result-payload failures, keep hard throws for transcript structure. **H2 and H3 are worth fixing in the same pass** (repo convention is fix-in-PR, not defer), and both need the regression tests named in §7. H4 and the §4 items are non-blocking. The `allowedTools` question (§5) is pre-existing, out of this diff, and belongs to the activation canary, not to this merge.

Everything the PR set out to do — result-side evidence proof, tool-set narrowing, retry hardening, mirror sync — was validated by both engines and holds.

---

*Automated multi-tool digest: 12 Claude lanes + 3 Codex legs, merged with cross-engine confidence weighting. Every cited `file:line` was re-checked in the worktree at `0a22ff92`, but engines disagree in places (noted inline) and no automated finding should be acted on without a human confirming the trigger.*

PR #2732, opened 2026-07-28
fix(jvm): enforce proximity-bounded sibling injection

voidfreud:
Supersedes the unpushable fix for #2724.

- Enforces the documented `GITNEXUS_MAX_INJECTED_SIBLINGS` cap (default 200; `0` remains unbounded).
- Adds a regression test proving distant siblings are excluded once the cap is reached.
- Rebases the fix onto current `main`.

Focused validation before rebase: `npx tsc --noEmit` and `vitest run test/unit/jvm-package-siblings.test.ts` (3 passed).

The original contributor fork rejects maintainer pushes (GitHub 403), so this replacement PR is required to land the fix.


voidfreud:
This replacement for #2724 carries the actual proximity cap and regression coverage: distant siblings are excluded at the default cap, while `GITNEXUS_MAX_INJECTED_SIBLINGS=0` remains unbounded. Focused tests and typecheck passed locally before opening. The current fork-PR CI runs are blocked as `action_required`; GitHub requires an upstream repository admin to approve them before they can execute. Please approve the pending workflow runs, then merge once the required checks are green.

voidfreud:
I’ve completed the requested fixes, rebased the affected branch, and added focused regression coverage where needed. The remaining blocker is GitHub’s upstream-only approval for fork workflow runs; I do not have permission to approve or rerun them. I’m stepping back until those workflows are approved or a concrete CI failure is available to address.

gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts
```diff
@@ -2,10 +2,21 @@ import type { BindingRef, ParsedFile, ScopeId, TypeRef } from 'gitnexus-shared';
...
+export type JvmPackageFact =
+  | { readonly status: 'known'; readonly packageName: string }
+  | { readonly status: 'unknown' };
 
 const MAX_PACKAGE_FILES = 500;
 
+const DEFAULT_MAX_INJECTED_SIBLINGS = 200;
+
+function getMaxInjectedSiblings(): number {
+  const raw = process.env.GITNEXUS_MAX_INJECTED_SIBLINGS;
```
magyargergo, line 15: Could you please document this in the readme? 🙏

magyargergo (changes requested):

gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts
magyargergo: Please update the `.env.example` too before I forget

voidfreud:
Current head `8c707cba` is based on `main` at `454d3834` and contains both requested README and `.env.example` documentation. Local focused validation passed (`jvm-package-siblings.test.ts`: 3 passed; `tsc --noEmit`: clean). The only CI failure is the known nondeterministic Windows `cli-limit-e2e` `status: null` failure documented in #2449; all PR-specific quality/typecheck/format checks passed. I cannot rerun upstream workflows or request re-review from the fork; the existing maintainer review request remains pending.

gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts
```diff
@@ -124,6 +136,7 @@ export function createJvmPackageSiblingVisibility(
         const injectedIds = new Set<string>();
         for (const { def } of candidates) {
           if (injectedIds.has(def.nodeId) || def.qualifiedName === undefined) continue;
+          if (maxInjectedSiblings > 0 && injectedIds.size >= maxInjectedSiblings) break;
```
abhigyanpatwari: **P1 [code-read], coordinator-verified — the cap silently truncates sibling injection without marking the file incomplete, feeding a wrong `false` into 7 Spring resolution call sites.**

Before this PR, sibling visibility was binary: complete (≤500 files, everything injected) or fully incomplete (>500 files, `incompleteFiles.add()` called for every file, line 98). This `break` introduces a third, unmarked state: once `injectedIds.size >= maxInjectedSiblings`, real siblings get silently dropped from `bindingAugmentations` with no `incompleteFiles.add(...)` call anywhere nearby — `isVisibilityIncomplete(filePath)` still returns `false`.

That flag feeds `gitnexus/src/core/ingestion/frameworks/spring/bean-candidates.ts:199`: `if (wildcardTarget === undefined || isPackageVisibilityIncomplete) return undefined;` — when wrongly `false`, this guard is bypassed and wildcard-annotation attribution proceeds against a truncated sibling set. The same flag is consumed by 7 files total (java: spring-bean-metadata.ts, spring-conditionals.ts, spring-config-bindings.ts, spring-di.ts; kotlin: spring-bean-metadata.ts, spring-conditionals.ts, spring-di.ts).

**Why it matters:** 201-500-file packages are exactly the population this PR targets. They're now the ones most likely to get silently-wrong Spring bean/DI/conditional/config-binding resolution, with zero log signal (the only `logger.warn` calls in this function fire for the >500-file skip and unknown-package cases, not for a cap-triggered truncation).

**Fix:** call `incompleteFiles.add(filePath)` (or mark the whole bucket, matching the >500-file path's pattern) whenever the cap actually truncates a module scope's candidate list.

gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts
```diff
@@ -2,10 +2,21 @@ import type { BindingRef, ParsedFile, ScopeId, TypeRef } from 'gitnexus-shared';
 import { logger } from '../../../logger.js';
 import type { ScopeResolutionIndexes } from '../../model/scope-resolution-indexes.js';
 import { isClassLike } from '../../scope-resolution/scope/walkers.js';
-import type { JvmPackageFact } from './package-facts.js';
+export type JvmPackageFact =
```
abhigyanpatwari: **P1 (latent, not an active bug today) — `JvmPackageFact` duplicated instead of imported. 6 lanes converged on this independently.**

This replaces `import type { JvmPackageFact } from './package-facts.js';` with a full local re-declaration, structurally identical to the canonical type still used by `java/package-facts.ts`, `kotlin/package-facts.ts`, and both languages' `capture-side-channel.ts`. No import-cycle justification exists (`package-facts.ts` has zero local imports). It only type-checks today because the two declarations are byte-for-byte identical — nothing keeps them in sync going forward; a future edit to either (new status variant, renamed field) risks silent drift or masked exhaustiveness bugs.

**Fix:** revert to the import, or use `export type { JvmPackageFact } from './package-facts.js';` if the test-file rewrite's import ergonomics were the actual motivation (most likely cause, based on the accompanying test rewrite).

gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts
```diff
@@ -79,6 +90,7 @@ export function createJvmPackageSiblingVisibility(
     }
 
     const augmentations = indexes.bindingAugmentations as Map<ScopeId, Map<string, BindingRef[]>>;
+    const maxInjectedSiblings = getMaxInjectedSiblings();
```
abhigyanpatwari, line 94: **P1 [code-read] — the cap only bounds `bindingAugmentations`; the `typeBindings`-merge loops directly below (lines ~149-164, unchanged by this diff) stay fully uncapped over the same sibling set.**

`maxInjectedSiblings` appears exactly twice in this file: its declaration here and the one `break` condition at line 139. The two `typeBindings`-merge loops right below the injection loop run unconditionally over the full bucket with no reference to the cap. So for a 450-file flat package, a module scope's `bindingAugmentations` gets at most 200 entries, but that same scope's `typeBindings` absorbs entries from all ~449 siblings unconditionally — the two structures that jointly represent "what's visible from this file" now disagree. `compound-receiver.ts`, `imported-return-types.ts`, and `callable-value-flow.ts` all read `scope.typeBindings` for receiver/variable type inference, so a class excluded from `bindingAugmentations` by the cap can still influence type inference via the uncapped `typeBindings` merge — an inconsistent half-resolved state.

**Separately (moderate confidence, same root cause — the proximity heuristic itself):** the proximity sort (`sharedSegmentCount`, unchanged) gives every candidate in the same directory an identical score; `Array.prototype.sort` is stable, and file traversal order is alphabetical by path (`filesystem-walker.ts:89`) — so for a flat package, "the 200 nearest" degrades to "the 200 alphabetically-earliest filenames," unrelated to actual code coupling. Deterministic and reproducible, not empirically tested against a real fixture.

gitnexus/README.md
```diff
@@ -311,6 +311,18 @@ validates the returned vector's length):
 
 Works with Infinity, vLLM, TEI, llama.cpp, Ollama, LM Studio, or OpenAI. Retry and pacing settings are provider-neutral; provider-specific limits should be supplied through configuration. When unset, local embeddings are used unchanged.
 
+## JVM Package Sibling Injection
+
+Java and Kotlin files in the same package receive implicit sibling class bindings
+to resolve same-package references. By default, GitNexus injects at most 200
+siblings per module scope. Set `GITNEXUS_MAX_INJECTED_SIBLINGS=0` to remove the
```
abhigyanpatwari: **P2 [code-read] — this new text is actively misleading for the exact population it's aimed at.**

"Set `GITNEXUS_MAX_INJECTED_SIBLINGS=0` to remove the limit" doesn't hold for packages over 500 files: the pre-existing `MAX_PACKAGE_FILES=500` skip-entirely branch fires before this cap's logic is ever reached, and this section never mentions that limit at all. A user hitting the `[java-package-siblings] skipping package...` warning who reads this section and sets the env var to 0 will see no effect.

This connects to a broader scope question (see PR body/digest): this PR supersedes PR #2724, whose headline claim was that large packages (up to 3009 files) would no longer be skipped entirely — that part isn't delivered here (`MAX_PACKAGE_FILES` is untouched, confirmed via the PR's own new 501-file test asserting the package is still marked incomplete). Worth adding a note here that packages over 500 files are still skipped entirely regardless of this env var.

abhigyanpatwari (commented):
# PR #2732 Tri-Review Digest — "fix(jvm): enforce proximity-bounded sibling injection"

**Repo:** abhigyanpatwari/GitNexus · **Author:** voidfreud (CONTRIBUTOR) · **Base:** main · **Head:** fix/jvm-package-siblings-topn
**Diff base:** `51095c19f8a7c641a136549711065f816a383ff6` · **Commits:** `dfb0b25c5` (core fix), `657e830af` (docs)
**Files touched:** `gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts` (+14/-1), `gitnexus/test/unit/jvm-package-siblings.test.ts` (rewritten, +105/-124), `gitnexus/README.md` (+12), `gitnexus/.env.example` (+4)

## What this PR does

Adds a `GITNEXUS_MAX_INJECTED_SIBLINGS` cap (default 200, `=0` for unbounded) to the `bindingAugmentations` injection loop inside JVM (Java/Kotlin) same-package sibling-visibility resolution. Candidates are sorted by shared-path-segment "proximity" to the target file, then the nearest `N` are injected as implicit same-package bindings; once the cap is hit, the loop `break`s. It supersedes PR #2724 (closed, unmerged, different author) which claimed a broader fix but never actually implemented cap enforcement.

## Engine breakdown

**Codex (Engine B) was entirely unavailable this run.** All three legs — `codex-swarm.md`, `codex-ce.md`, `codex-adversarial.md` — are spend-cap error markers with zero real output (confirmed directly: each file is a 5-7 line stub ending in "You hit your spend cap set by the owner of your workspace"). Every finding in this digest comes from Claude only, across two engines-worth-of-personas run on the same model: the GitNexus swarm (facts-historian, branch-hygiene, risk-architect, test-ci) and the ce-code-review personas (correctness, testing, maintainability, project-standards, performance, adversarial). Because both lane sets share the same underlying model, convergence between them is **same-engine consistency, not independent-engine confirmation** — treat "moderate" corroboration below as a weaker signal than it would be in a normal two-engine run.

I (synthesis) independently re-verified the two most load-bearing claims directly against the live worktree and GitHub before trusting lane convergence: the headline silent-truncation chain (code read, confirmed), and the CHANGES_REQUESTED review state (`gh pr view 2732 --json mergeStateStatus,mergeable,reviewDecision,reviews`, confirmed current and unresolved).

## Overall verdict

**Not ready to merge as-is.** The cap mechanism itself is small, correct in isolation (no off-by-one, no cross-scope leak — independently traced and refuted as bugs by ce-correctness), and a real improvement for the narrow band it targets. But it introduces one genuine new correctness regression (Finding 1, P1) in a domain the project treats as production-critical (Spring DI/bean resolution), ships with a hard-blocking stale review, and pairs a real fix with test coverage that got smaller, not just different. Recommend: fix Finding 1 before merge (one-line `incompleteFiles.add` call), get the stale review cleared, and treat the type-duplication and coverage-regression items as fast pre-merge cleanup rather than blockers in isolation.

---

## What's solid

- The cap's core arithmetic is correct: independently hand-traced by ce-correctness against the shipped code (line-by-line) and against the PR's own test — no off-by-one, `continue` vs `break` ordering is right, the per-module-scope `injectedIds` Set correctly resets per file (no leak across siblings in the same package bucket).
- `getMaxInjectedSiblings()` is read once per `populateNamespaceSiblings` call, not per bucket/file — not a hot-path cost concern (ce-performance, ce-correctness).
- Branch/PR hygiene is clean otherwise: linear 2-commit history, correct original-author attribution preserved through a maintainer-recreated push, no unrelated churn, all 36 non-Vercel CI checks pass (Vercel fails only on the expected fork-auth gate).
- The project-standards "shared pipeline must not name languages" rule does **not** apply/is not violated — `package-siblings.ts` lives correctly under `languages/jvm/`, not the shared pipeline root, and its shared-root imports are the correct direction (language plugin → shared hook surface).

---

## Findings (most severe first)

### Finding 1 (P1) — Cap silently truncates sibling injection WITHOUT marking files incomplete, feeding wrong `false` into 7 Spring resolution call sites
**Lanes:** risk-architect (Findings A + C, HEADLINE) · ce-adversarial (adv-2732-4, independently via a different angle: "silent regression, no log signal") · coordinator-verified directly against live code.
**Tag:** [code-read], confirmed by synthesis independently, not just trusted from lane convergence.

Before this PR, sibling visibility for a package was binary: complete (≤500 files, everything injected) or fully incomplete (>500 files, `incompleteFiles.add()` called for every file in the bucket, wildcard attribution disabled). This PR adds a third, unmarked state. I read the shipped file directly:

```
136  const injectedIds = new Set<string>();
137  for (const { def } of candidates) {
138    if (injectedIds.has(def.nodeId) || def.qualifiedName === undefined) continue;
139    if (maxInjectedSiblings > 0 && injectedIds.size >= maxInjectedSiblings) break;
140    injectedIds.add(def.nodeId);
...
```
(`gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts:136-147`)

There is no `incompleteFiles.add(...)` call anywhere near the `break` at line 139 — confirmed by reading the full function (lines 41-173). Contrast with the existing >500-file skip branch four lines above it (lines 97-103), which does call `incompleteFiles.add(parsed.filePath)` for every file in the bucket. So a package in the 201-500-file range that exceeds 200 candidates per module scope now gets real siblings silently dropped from `bindingAugmentations`, while `isVisibilityIncomplete(filePath)` still returns `false` for those files.

I independently confirmed the consumer chain (not just trusted risk-architect's claim): `grep -rn "isPackageVisibilityIncomplete\|isJavaPackageSiblingVisibilityIncomplete\|isKotlinPackageSiblingVisibilityIncomplete"` across the ingestion tree shows the flag is wired through **7 files** (not the 6 risk-architect counted — java has 4: `spring-bean-metadata.ts`, `spring-conditionals.ts`, `spring-config-bindings.ts`, `spring-di.ts`; kotlin has 3, no config-bindings equivalent: `spring-bean-metadata.ts`, `spring-di.ts`, `spring-conditionals.ts`). The sharpest single consumer, `gitnexus/src/core/ingestion/frameworks/spring/bean-candidates.ts:199`:
```ts
if (wildcardTarget === undefined || isPackageVisibilityIncomplete) return undefined;
```
When the flag is (wrongly) `false`, this guard is bypassed and wildcard-annotation attribution proceeds against a truncated sibling set.

**Why it matters:** large Spring packages (201-500 classes) are exactly the population this PR is aimed at helping. They are now the ones most likely to get silently-wrong Spring bean/DI/conditional/config-binding resolution, with zero log signal — the only `logger.warn` calls in the function fire for the >500-file hard skip and for unknown-package files, not for a cap-triggered truncation (independently noted by ce-adversarial adv-2732-4).

**Fix:** call `incompleteFiles.add(filePath)` (or the whole bucket, matching the >500-file path's pattern) whenever the cap actually truncates a given module scope's candidate list — e.g. when `candidates.length > maxInjectedSiblings` before the loop, or when the `break` fires.

### Finding 2 (P1) — Issue/PR #2724 lineage: "large packages no longer skipped" is false, and the claim was never backed by working code at any point in this lineage
**Lanes:** prior-context (initial discovery) · facts-historian (PR/issue-numbering correction) · branch-hygiene · risk-architect (Finding B) · ce-adversarial (adv-2732-3, went furthest) · ce-performance (independently found the "9M→600K" figure traces to the unreachable case).
**Tag:** [code-read] + [reproduced via `gh api`/`curl`], multiple lanes independently verified different facets.

Fullest picture, synthesized across lanes:
- **#2724 is a PR, not an issue** (facts-historian's correction to prior-context) — authored by ChunxueLi, `merged: false`, `merged_at: null`, closed 2026-07-28 shortly after #2732 opened, i.e. superseded, not merged.
- **#2724's own diff never implemented the cap it claimed.** ce-adversarial fetched `#2724`'s raw diff via `curl .../pull/2724.diff` and grep-confirmed zero occurrences of `MAX_INJECTED_SIBLINGS`/`getMaxInjectedSiblings`/`maxInjectedSiblings`. A real blocking review comment on #2724 from `azizur100389` (also a requested reviewer on #2732) independently confirms: "the implementation does not actually bound sibling injection to a Top-N subset... preserves the same high fan-out/O(n^2) behavior." That's why #2724 was closed unmerged.
- **The unbounded proximity-sort file `#2724` diffs against actually entered `main` via an unrelated PR** (`2cfbc4a25`, Spring bean-candidate inventory, #2494) — #2724's "new file" diff is a stale-branch artifact, not evidence of its own authorship of that logic.
- **This PR (#2732) does ship real, working cap-enforcement** (verified: `getMaxInjectedSiblings()` + the `break` at line 139) — it fixes the specific defect the reviewer blocked #2724 on.
- **What's NOT delivered:** `MAX_PACKAGE_FILES = 500` (line 9) and the skip-entirely branch (lines 97-103) are byte-identical before/after this diff (confirmed via `git diff`) — no `GITNEXUS_MAX_PACKAGE_FILES` env var exists anywhere. The PR's own new test (`'marks a capped package incomplete...'`, 501 files) explicitly asserts a >500-file package is still fully skipped — i.e. the test *certifies* the exact behavior #2724's headline claimed was being eliminated.
- **The "9M→600K augmentations (93% reduction)" and "12% faster on a 283K-node monorepo" figures never described any code that has existed on `main`.** ce-performance independently derived that `3009² ≈ 9.05M` and `3009×200 ≈ 600K` map exactly onto the unreachable >500-file case (files that `continue` at line 97 before ever reaching the capped loop), so the headline number this whole lineage is justified by describes a code path that has never run.

**Net effect:** the 9 packages (up to 3009 files) that motivated this lineage are exactly as skipped after this PR as before it. #2732's own body ("Supersedes the unpushable fix for #2724... Rebases the fix onto current main") is narrow and doesn't repeat #2724's "no longer skipped" claim outright, but by framing itself as #2724's successor without correcting that claim, it invites a reasonable reader to believe the large-package problem is resolved. It is not, and never has been, in any PR in this lineage.

### Finding 3 (P2) — PR's own new README text is actively misleading for the exact population it's aimed at
**Lanes:** facts-historian (Finding 3) · corroborated by ce-project-standards (Finding 3, root README omission).
**Tag:** [code-read], confirmed directly — new lines in this diff.

Confirmed via `git diff` on `gitnexus/README.md`:
```
+Set `GITNEXUS_MAX_INJECTED_SIBLINGS=0` to remove the limit; this can substantially
+increase indexing work for large packages.
```
This is directly misleading for >500-file packages: the skip-entirely branch fires before the cap logic is ever reached, so a user hitting the `[java-package-siblings] skipping package...` warning who reads this section and sets the env var to 0 sees no effect. The new README section never mentions `MAX_PACKAGE_FILES` or the 500-file limit at all (confirmed: `grep -n "MAX_PACKAGE_FILES"` against the new section returns nothing). Unlike Finding 2 (an issue-lineage/scope question), this is anchored to specific new lines in the diff itself and is the best candidate for an inline PR comment.

### Finding 4 (P1, but latent — not an active bug today) — `JvmPackageFact` type duplication
**Lanes:** prior-context · branch-hygiene · ce-maintainability (confidence 93, deepest dive) · ce-correctness · facts-historian · risk-architect (Finding E) — **6 lanes, unanimous.**
**Tag:** [code-read], confirmed directly.

`package-siblings.ts` replaces `import type { JvmPackageFact } from './package-facts.js';` with a full local re-declaration:
```ts
export type JvmPackageFact =
  | { readonly status: 'known'; readonly packageName: string }
  | { readonly status: 'unknown' };
```
Confirmed via `git diff`: this is a straight substitution in commit `dfb0b25c5`, not a rename. The canonical type in `jvm/package-facts.ts` is untouched and still consumed as-is by `java/package-facts.ts`, `kotlin/package-facts.ts`, and both languages' `capture-side-channel.ts`. Confirmed no import-cycle justification: `package-facts.ts` has zero local imports. `getJavaPackageFact`/`getKotlinPackageFact` (typed against the canonical type) are passed straight into `createJvmPackageSiblingVisibility`'s `getPackageFact` option (now typed against the independent copy) — this only type-checks today because the two declarations happen to be byte-for-byte identical. No lane could construct a concrete runtime/compile-time divergence trigger today (ce-correctness tried explicitly and reports this as a residual risk, not an active bug) — but nothing keeps the two in sync going forward; a future edit to either (new status variant, renamed field) risks silent drift or masked exhaustiveness bugs. Most plausible cause (ce-maintainability): the accompanying test-file rewrite wanted to import all symbols from one module and used a re-declaration instead of a one-line `export type { JvmPackageFact } from './package-facts.js';` re-export, which would have worked identically.

**Fix:** revert to the import, or use a type re-export if the test-import ergonomics were the goal. One-line change either way.

### Finding 5 (P1) — Structural inconsistency: cap only bounds `bindingAugmentations`; `typeBindings`-merge loops directly below stay fully uncapped over the same sibling set
**Lanes:** ce-adversarial (adv-2732-2) · risk-architect (Finding F).
**Tag:** [code-read], confirmed directly.

Confirmed via direct read (`package-siblings.ts:149-164`): immediately below the now-capped injection loop, two more loops run unconditionally over the full bucket — `for (const sibling of bucket.moduleScopes)` (149-155) and `for (const sibling of bucket.parsed)` (156-164), both merging into `scope.typeBindings` with no reference to `injectedIds`, `maxInjectedSiblings`, or any cap. `maxInjectedSiblings` is grep-confirmed to appear exactly twice in the whole file: its declaration and the one `break` condition. So for a package with, e.g., 450 files in one directory, a given module scope's `bindingAugmentations` gets at most 200 entries, but that same scope's `typeBindings` absorbs entries from all ~449 other siblings unconditionally — the two structures that jointly represent "what's visible to this file from its package" now disagree within the very function this PR touches. `compound-receiver.ts`, `imported-return-types.ts`, and `callable-value-flow.ts` all read `scope.typeBindings` for receiver/variable type inference, so a class excluded from `bindingAugmentations` by the cap can still influence type inference via the uncapped `typeBindings` merge — an inconsistent half-resolved state. This also undercuts the PR's stated performance rationale: the dominant O(N²) cost (proximity sort + typeBindings merge) is untouched; only the number of writes into `bindingAugmentations` is bounded (ce-performance corroborates independently — see Lower-priority section).

### Finding 6 (P1, moderate confidence) — Proximity tie-break degrades to alphabetical-by-path bias for flat/single-directory packages
**Lanes:** ce-adversarial (adv-2732-1, primary trace) · risk-architect ("unverified suspicions" section, independently arrived at from a different angle — multi-module no-shared-prefix case).
**Tag:** [code-read] chain, confidence: plausible, not empirically reproduced against a real fixture.

I verified the mechanism directly: `sharedSegmentCount` (lines 175-187) walks path segments from index 0 and stops at first mismatch — for any two files in the same directory, every non-filename segment matches, so shared-segment count is identical for every candidate pair in that directory. `candidates.sort(...)` (132-134) then compares all-equal scores; `Array.prototype.sort` is stable in V8/Node, so ties resolve to original array order. I confirmed that original order traces back to `filesystem-walker.ts:89`, which explicitly sorts files alphabetically by path (`entries.sort((left, right) => (left.path < right.path ? -1 : left.path > right.path ? 1 : 0))`), with a comment stating this canonicalizes order for all downstream phases. So for a flat package (a common real pattern: protobuf/OpenAPI/JOOQ-generated DTOs, exception hierarchies), the cap deterministically keeps the 200 alphabetically-earliest classes by file path, regardless of which classes are actually referenced elsewhere in the package — meaning "the 200 nearest" in a flat package is really "the 200 earliest-sorting filenames," unrelated to real code coupling. This is deterministic and reproducible (same repo → same wrong 200 every run), not a flake. Risk-architect independently arrived at an analogous concern from a different trigger (multi-module monorepos with no shared path prefix — cross-module proximity collapses to 0 for everyone, same alphabetical-tie-break degradation). Neither lane tested this against a real fixture; flagged as a real, traced mechanism rather than an empirically-reproduced bug — worth a fixture with genuine tie-pressure before/if this ships.

---

## Lower-priority / informational

- **Performance claim overstated (ce-performance, [code-read]).** Even for the packages this PR does affect (~201-500 files), the cap only bounds the injection loop's writes; the dominant costs — `O(M² log M)` proximity sort and two `O(M²)` typeBindings-merge loops per bucket — are untouched and unbounded, running identically before and after this PR. At `M=500`, the injection loop itself goes from ~249,500 to ~100,000 `bindings.push` calls (~60% real cut in output volume/memory) — a genuine but much smaller win than the "93% reduction" figure implies, and the O(N²) sort/merge cost dominates regardless. Also flags forward-looking risk: if a follow-up literally "un-skips" large packages per #2724's ask, the untouched sort/merge would become the real bottleneck at 3009-file scale (~10⁸ comparisons), which this PR's fix does nothing to address.
- **`INCREMENTAL_SCHEMA_VERSION` not bumped (risk-architect, Finding G, P3).** This is a semantic change to which edges get produced for capped packages; existing incremental indexes may retain pre-cap edges. Self-correcting via `--force`; low severity.
- **`GITNEXUS_MAX_INJECTED_SIBLINGS` env-parsing edge cases (ce-correctness, low/confidence 35).** A whitespace-only value (`' '`) is silently treated as `0`/unlimited rather than falling back to the default 200, because `Number('   ') === 0` and `Number.isInteger(0)` is true, and only `undefined`/`''` are special-cased before `Number(raw)`. Narrow, operator-error-only trigger; not confident it's worth a code change.
- **Root `/README.md` env-var reference table omits the new var (ce-project-standards, Finding 3, confidence 45).** The root table catalogs ~25 `GITNEXUS_*` tuning vars with a consistent "Variable | Default | Effect | Tune when…" format; `GITNEXUS_MAX_INJECTED_SIBLINGS` isn't in it — confirmed via direct grep, zero hits. Documented instead in `gitnexus/README.md` + `.env.example` only. May be an acceptable split given existing doc overlap by design; flagged as a judgment call, not a clear violation.

## Refuted (checked and closed out by a lane, not carried forward as findings)

- **Off-by-one / `continue`-vs-`break` ordering bug in the cap loop** — ce-correctness hand-traced this against both the shipped code and the PR's own test; `continue` for duplicate/undefined candidates correctly does not consume a cap slot before the `break` check runs. No bug.
- **Cap leaking/shared across module scopes in the same package bucket** — ce-correctness confirmed `injectedIds` is declared inside the per-module-scope loop, so each file gets an independent budget. No leak.
- **`maxInjectedSiblings` staleness from being computed once per call** — ce-correctness found this moot; it's read once per `populateNamespaceSiblings` invocation (same pattern the pre-existing hardcoded `MAX_PACKAGE_FILES` already follows), not a real concern.
- **Shared-pipeline "must not name languages" rule violation** — ce-project-standards explicitly checked; `package-siblings.ts` correctly lives under `languages/jvm/`, not the shared root, dependency direction is correct. Not violated.
- **Alleged prompt-injection / tool-output tampering** — raised and then self-corrected within `claude-ce-project-standards.md` in the same session; confirmed benign (the repo's own GitNexus PreToolUse hook output, initially misread). Fully resolved noise, not a PR finding — excluded from this digest per explicit coordinator instruction. (Note: an unrelated, similarly benign GitNexus PreToolUse hook fired during this synthesis pass too — same phenomenon, not actionable.)

---

## Test gaps (not blocking on their own, but real)

- **Real tree-sitter package-header extraction coverage was deleted with zero replacement.** Three+ lanes converged: test-ci-verifier called it P0/BLOCK in isolation, ce-testing independently verified via repo-wide grep (confirmed `packageFact`/`PackageFact` appears in exactly one unrelated other test file), risk-architect (Finding D), ce-project-standards (Finding 1). I independently confirmed via grep too: no test file covers `extractJvmPackageFact`'s named/default/malformed/broken-body cases for either grammar. **Important distinction (own judgment, weigh accordingly):** the underlying `jvm/package-facts.ts` code is unchanged by this PR — this is a coverage regression for existing, untouched code, not a new bug in new code. I'd score this **P1, not P0/blocking** — real and worth restoring, but not evidence the shipped cap logic itself is wrong.
- **The proximity-cap test's fixture cannot distinguish real nearest-first sorting from a no-op array-truncation (ce-testing-1, P1, confidence 78).** The fixture's 200 "near" candidates are all tied at the same proximity score and already in sort-order by construction, and the 1 "far" candidate is both least-proximate and last-inserted — so a correctly-sorting implementation and a completely broken/no-op one produce identical results. No tie-pressure (>cap candidates at equal top proximity) is ever exercised, and assertions only spot-check two boundary keys rather than the total injected count or non-boundary candidates.
- **`getMaxInjectedSiblings()`'s fallback branch (non-numeric/negative/non-integer env values) is untested** (ce-testing-2, P1, confidence 85; overlaps ce-correctness's testing_gaps). Only "unset" (default) and `'0'` (explicit unlimited) are exercised.
- **No test proves the per-module-scope cap independence isn't accidentally shared/leaking** (ce-testing-3, P2) — the fixture actually has 202 module scopes ideally suited to prove this, but only ever asserts on scope `module:0`.
- **Test file was substantially narrowed in scope, not just extended** (branch-hygiene, ce-project-standards) — dropped the dual-language (`LanguagePackageHarness`) integration-style coverage through the real `emitJavaScopeCaptures`/`collectJavaCaptureSideChannel`/`populateJavaPackageSiblings` pipeline in favor of generic single-engine tests against hand-built fixtures. Arguably a legitimate design simplification, but it means no test currently exercises the real Java/Kotlin facade wiring end-to-end.
- **No test exercises the flat-directory tie-break scenario (Finding 6) or the capped/uncapped `bindingAugmentations`/`typeBindings` divergence (Finding 5).**

## CI / mergeability status

**mergeStateStatus is BLOCKED — independently re-confirmed by synthesis via `gh pr view 2732 --json mergeStateStatus,mergeable,reviewDecision,reviews`.** Not conflicts (`mergeable: MERGEABLE`, linear 2-commit history, no branch protection configured) but an unresolved `CHANGES_REQUESTED` review from `magyargergo` (2 reviews, both submitted 2026-07-28T18:53-18:55, both anchored to commit `49f7ec115` — the pre-docs commit) carrying 2 inline comments: "document this in the readme" and "update the .env.example too before I forget." **Both requests were addressed by the follow-up commit `657e830af`** ("docs(jvm): document sibling injection cap") — but the review itself was never re-cleared, which is the actual current merge gate. This needs a human action (dismiss the stale review or get a fresh approval) — a bot review cannot resolve it, but it should be stated plainly for whoever reads this digest. All 36 other status checks pass; Vercel fails only on the expected fork-authorization gate.

## Coverage note

Codex (Engine B) contributed nothing to this review — all three of its legs (swarm, ce, adversarial) hit the workspace spend cap before producing any output. Every finding above is Claude-only, drawn from two persona sets run on the same underlying model (GitNexus swarm: facts-historian/branch-hygiene/risk-architect/test-ci; ce-code-review: correctness/testing/maintainability/project-standards/performance/adversarial). Treat multi-lane convergence as same-engine consistency, not cross-engine confirmation — no finding here has been checked by an independent model, only by independent read-throughs of the same code by the same underlying model in different personas, plus one round of direct synthesis-level re-verification against the live worktree and GitHub state for the highest-severity claims (Finding 1's code chain, and the CHANGES_REQUESTED review).

## Closing line

A small, well-intentioned cap that fixes a real defect from a rejected predecessor PR — but it trades an explicit "we skipped this, results are incomplete" failure mode for a silent one in exactly the size band (201-500 files) it was built to help, inside a subsystem (Spring resolution) that six-plus call sites treat as authoritative; fix that one line, clear the stale review, and this is close to mergeable.

azizur100389:
Reviewed current head `657e830a` against `upstream/main` `51095c19` in a fresh worktree. It fast-forwarded cleanly; `git diff --check` passed; no hidden bidi controls were found. `jvm-package-siblings.test.ts` passed all 3 tests and `npx tsc --noEmit` passed. CI is green apart from Vercel authorization, but the PR remains `CHANGES_REQUESTED`/blocked.

Branch hygiene: `clean feature/fix PR`. Merge state: `review blocked`.

Findings:

- **Risk:** when the 200-sibling cap truncates candidates, the target file is not marked visibility-incomplete. Downstream Spring resolution uses that flag to avoid treating partial package visibility as exact, so packages in the 201?500 range can now produce silently incomplete bean/DI/conditional attribution.
- **Evidence to check:** `gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts`: the cap breaks at the `injectedIds.size` check, but only the >500-file and unknown-package branches add files to `incompleteFiles`.
- **Recommended fix:** mark the affected file (or bucket) incomplete whenever the cap actually truncates candidates, and assert that behavior in the cap regression test.
- **Blocks merge:** yes.

- **Risk:** only `bindingAugmentations` is capped; the two `typeBindings` merge loops immediately below still traverse every sibling. This leaves package visibility internally inconsistent and retains substantial quadratic work.
- **Evidence to check:** the uncapped loops over `bucket.moduleScopes` and `bucket.parsed` in the same function after the capped injection loop.
- **Recommended fix:** define one coherent bounded sibling set and apply it consistently, or document and test why type bindings must intentionally remain complete despite binding truncation.
- **Blocks merge:** yes.

- **Risk:** the rewritten test removed the prior Java/Kotlin capture-side-channel coverage and its proximity fixture cannot distinguish a working sort from truncating the original order because all near candidates are already ordered and tied.
- **Evidence to check:** `gitnexus/test/unit/jvm-package-siblings.test.ts` now uses hand-built generic fixtures and only checks one near and one far key.
- **Recommended fix:** restore real Java/Kotlin facade coverage and add tie-pressure/out-of-order candidates plus an assertion for the exact capped set/count.
- **Blocks merge:** maybe.

Final verdict: `not production-ready`. The PR now enforces the missing cap from #2724, but it changes complete visibility into silent partial visibility and applies the bound inconsistently. Those correctness issues should be resolved before the stale requested-changes review is cleared.

gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts
magyargergo: Both are in — and corrected in this round. `README.md` has a **JVM Package Sibling Injection** section and `.env.example` carries the variable with its default.

Two things the first pass got wrong, per the follow-up review: the docs implied `GITNEXUS_MAX_INJECTED_SIBLINGS=0` lifts every limit (it does not lift the fixed 500-file package skip, which fires first and is what a user hitting the `skipping package with N files` warning is actually seeing), and they said nothing about what truncation costs. Both now state it: packages over 500 files are skipped entirely regardless of this variable, and a file whose sibling set gets truncated is marked visibility-incomplete, so wildcard-import attribution is disabled for it rather than resolved against a partial view. Analyze also warns once with the number of affected files.

gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts
magyargergo: Fixed in `e02294d`. The truncating `break` now records it, and the file is marked incomplete:

```ts
if (maxInjectedSiblings > 0 && injectedIds.size >= maxInjectedSiblings) {
  truncated = true;
  break;
}
...
if (truncated) {
  incompleteFiles.add(filePath);
  truncatedFiles++;
}
```

Per file, not per bucket — the truncation is per module scope, so marking the whole package would over-degrade the files that were fully injected. After the bucket loop, one aggregated `logger.warn` names the cap and how many files it hit, so the state is no longer silent (the per-file version would flood on a 400-file package).

Regression test added: `marks a file whose sibling set was truncated as visibility-incomplete`, plus the default-cap case which asserts the same for an unset env var. Both fail against the pre-fix implementation.

gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts
magyargergo: Fixed in `e02294d` — reverted to `import type { JvmPackageFact } from './package-facts.js';`. The test file was indeed the motivation; it now imports the type from `package-facts.js` (as it did before the rewrite) and the factory from `package-siblings.js`, so no re-export is needed either.

gitnexus/src/core/ingestion/languages/jvm/package-siblings.ts
magyargergo: Fixed in `e02294d`. The injection loop now records which sibling *files* it drew from, and both type-binding merges use that same bounded set:

```ts
const mergePaths = truncated ? injectedPaths : allSiblingPaths;
```

Two details worth flagging:

- The bound is applied by iterating `mergePaths` and looking siblings up in per-bucket `Map`s, not by filtering a full rescan — so the cap bounds the merge *work*, not just its result. That was your and @azizur100389's "retains substantial quadratic work" point. Same pass also splits each path into segments once per bucket instead of re-splitting on every pairwise `sharedSegmentCount` comparison, which runs O(files²) per package.
- The untruncated path still merges every sibling, byte-for-byte as before. `injectedPaths` only contains files that contributed a class-like def, so making it unconditional would silently drop type bindings from siblings that declare none — a different (and unrequested) behaviour change.

New test `bounds type bindings by the same sibling set it bounded bindings by` gives each sibling a self-named module type binding, so the merged key set names exactly which siblings were treated as visible; it fails on the pre-fix code.

On the proximity heuristic: you're right that ties within a directory fall back to walker order, so "the 200 nearest" is really "200 arbitrary equally-near siblings" for a flat package. Left the heuristic alone — a real coupling signal would need reference data this pass runs before — but it is now documented at the sort, and, more importantly, truncation no longer claims exact visibility, which is what made the arbitrary tie order dangerous.

gitnexus/README.md
magyargergo: Fixed in `e02294d`. The section no longer implies the env var lifts every limit. It now states plainly that packages over 500 files are skipped entirely — logged as `skipping package with N files`, every file in them marked visibility-incomplete — and that `GITNEXUS_MAX_INJECTED_SIBLINGS` does not lift that skip, including at `0`. It also documents what truncation costs: the affected file is marked visibility-incomplete, so wildcard attribution is disabled for it. `.env.example` carries the same two caveats. The `MAX_PACKAGE_FILES` constant has a comment saying the same thing at the source.

On scope: you're right that #2724's "large packages are no longer skipped" claim is still not delivered — `MAX_PACKAGE_FILES` is deliberately untouched here, and the 501-file test still asserts the package is skipped. This PR is the cap plus its correctness fallout only; raising or removing the 500-file skip needs its own change (and its own evidence that the quadratic merge cost is acceptable at that size). The docs now describe the shipped behaviour rather than the intended-someday one.

magyargergo:
Pushed two commits addressing every open comment (`e02294d`, `46a2f49`). Per-comment replies are inline; summary:

| Finding | Where | Fix |
| --- | --- | --- |
| P1 cap truncates silently — `isVisibilityIncomplete` still `false`, feeding 7 Spring resolution sites | @abhigyanpatwari, @azizur100389 | Truncation marks the file incomplete and increments a counter; one aggregated `logger.warn` after the bucket loop names the cap and the affected file count. Per file, not per package, so fully-injected files aren't over-degraded. |
| P1 only `bindingAugmentations` bounded; `typeBindings` merges uncapped | @abhigyanpatwari, @azizur100389 | Both halves use one bounded sibling set. The merge iterates that set through per-bucket lookups instead of filtering a full rescan, so the cap bounds the work too — the "retains substantial quadratic work" half of the finding. Path segments are also split once per bucket rather than on every pairwise proximity comparison. |
| P1 `JvmPackageFact` re-declared instead of imported | @abhigyanpatwari | Reverted to the import from `package-facts.js`. The test imports the type from there and the factory from `package-siblings.js`, so no re-export is needed. |
| P2 README misleading for >500-file packages | @abhigyanpatwari | README and `.env.example` now state that the 500-file package skip is fixed and unaffected by `GITNEXUS_MAX_INJECTED_SIBLINGS` (including `0`), and that truncation disables wildcard attribution for the affected file. |
| Test rewrite dropped Java/Kotlin facade coverage; fixture can't distinguish sort from truncation | @azizur100389 | Restored the per-language harness (package-fact capture, 500-file skip, fail-closed on a missing `ParsedFile`) and added five cap cases on top, with a near/far interleaved fixture whose retained set is only reachable via a working proximity sort. Asserts the exact capped key set, not just one near and one far. |
| Document the env var in README and `.env.example` | @magyargergo | Present, and corrected as above. |

Untouched on purpose: `MAX_PACKAGE_FILES = 500`. #2724's "large packages are no longer skipped" claim is still not delivered here, and the 501-file test still asserts the skip — raising that ceiling needs its own change and its own evidence that the merge cost holds at that size. The proximity heuristic is also unchanged: ties within a directory fall back to walker order, which is now documented at the sort, and truncation no longer presents that arbitrary set as exact visibility.

Local validation from `gitnexus/`: `tsc --noEmit` clean; `vitest run test/unit/jvm-package-siblings.test.ts` 11 passed; the JVM/Spring suites (`java-call-arity`, `java-spring-route-ingestion`, `kotlin-scope-captures`, `spring-bean-extractor`, `spring-config-bindings`, `spring-auto-configuration`, `spring-interface-inheritance`, `range-binding-parse-timeout`) 74 passed across 9 files. Each new cap test was re-run against the pre-fix implementation: the truncation-marking, bounded-type-bindings and default-cap cases fail there, so none of them pass vacuously.

magyargergo (approved):

PR #2737, opened 2026-07-28
fix(resolution): type inline constructor receivers in every spelling (#2708)

### LINKED ISSUE #2708: Method call on an inline constructor receiver (Service(db).do_work()) emits no CALLS edge, yet epistemic still reports "exact"
## Summary

A method call whose receiver is an **inline constructor expression** — `Service(db).do_work()` — produces no `CALLS` edge, so the caller is missing from both `impact(direction: "upstream")` and `context()`. The identical call written in two steps (`s = Service(db)` then `s.do_work()`) resolves correctly.

The part that makes this more than a coverage gap: **`epistemic` is still reported as `"exact"`** while the caller is dropped. Since #1858 shipped the epistemic lower-bound signal precisely so `impact()`/`context()` can flag unresolved indirection, an unresolved inline-constructor receiver looks like a case that mechanism was built for but doesn't currently classify.

Version: **gitnexus 1.6.9** (latest published; `npm view gitnexus version` = 1.6.9). Language: Python.

## Minimal reproduction

4 files, 11 nodes, 14 edges — fully self-contained.

```python
# pkg/svc.py
class Service:
    def __init__(self, db):
        self.db = db

    def do_work(self):
        return 1
```

```python
# pkg/caller_inline.py
from pkg.svc import Service


def route_inline(db):
    return Service(db).do_work()        # <-- inline constructor receiver
```

```python
# pkg/caller_twostep.py
from pkg.svc import Service


def route_twostep(db):
    s = Service(db)
    return s.do_work()                  # <-- two-step receiver
```

```bash
# plus an empty pkg/__init__.py
git init && git add -A && git commit -m repro
npx gitnexus analyze .
npx gitnexus impact "do_work" --direction upstream --repo gnrepro
```

### Actual

```json
{
  "target": { "id": "Method:pkg/svc.py:Service.do_work#0", "name": "do_work", "type": "Method" },
  "direction": "upstream",
  "impactedCount": 1,
  "risk": "LOW",
  "epistemic": "exact",
  "byDepth": {
    "1": [
      { "depth": 1, "id": "Function:pkg/caller_twostep.py:route_twostep",
        "name": "route_twostep", "relationType": "CALLS", "confidence": 0.85 }
    ]
  }
}
```

`route_inline` is absent. `context("do_work")` returns the same single `incoming.calls` entry (`route_twostep`) and likewise reports `"epistemic": "exact"` — so `context()` is not a workaround for this case.

### Expected

Either `route_inline` appears as a `CALLS` edge, **or** — if resolving constructor-expression receivers is out of scope for the current resolution pass — `epistemic` degrades from `"exact"` to the lower-bound value from #1858, so the count is not presented as authoritative.

## Why the `epistemic` half matters independently of the edge fix

We consult `impact(direction: "upstream")` as a mandatory pre-edit blast-radius check before modifying any symbol. A **wrong count labelled a lower bound** is survivable — it tells the caller to go verify by hand. A **wrong count labelled `"exact"`** is acted on as authoritative.

Real-world instance, in a Python service where `ClassName(db).method()` is the dominant convention (167 such call sites in one service directory):

| call site | receiver form | in `impact()` output |
|---|---|---|
| `app/api/v1/specs.py:946` — `OrchestratorSpecService(db).reconcile_one(...)` | inline | no |
| `app/mcp_server.py:731` — `OrchestratorSpecService(db).reconcile_one(...)` | inline | no |
| `tests/test_orch45_reconcile_persist.py:159` — `OrchestratorSpecService(sqlite_db).reconcile_one(...)` | inline | no |
| `tests/test_orch133_status_stage_spine.py:704,714` — `svc.reconcile_one(...)` | two-step | **yes** |

`impact(target: "reconcile_one", direction: "upstream", includeTests: true)` returns `impactedCount: 1` (the two-step test caller), `risk: "LOW"`, `epistemic: "exact"` — while **both production callers are missing**. Receiver form predicted edge presence with 100% accuracy across all 5 sites.

Two alternative explanations were ruled out before filing:

- **Not the `includeTests` filter.** The run above sets `includeTests: true`. (We first hit this on a symbol where the test-filter default did explain the discrepancy — that one was our misreading, not a bug, and is consistent with #2686.)
- **Not name-collision / node-id shadowing** (#1680, #2699). `reconcile_one` is called from `def reconcile_spec()` — the names differ, so there is no same-named local for the edge to be mis-attributed to. The edge is dropped, not redirected.

## Possibly related

- #1858 — epistemic lower-bound when unresolved indirection is on the path (the mechanism this seems to want)
- #2693 — calls through a callable-valued binding resolve only in C
- #407 — Swift cross-file type resolution via constructor inference
- #1857 — DI interface→implementation autowiring invisible to `impact()`
- #2686 — `impact(upstream)` returning 0 with one non-test caller plus a test caller

Happy to test a patch against the real repo (large Python monorepo, ~4k Python files) if that's useful.

magyargergo:
Fixes #2708 (edge half — see "What is NOT fixed" below).

## The bug

`Service(db).do_work()` emitted no `CALLS` edge, so the caller was missing from `impact(direction: "upstream")` and `context()`. The two-step spelling of the identical call resolved fine.

Reproduced on `main` with the issue's 4-file fixture: 11 nodes, 14 edges, exactly one `CALLS` edge in the whole graph (`route_twostep → do_work`). `route_inline` had **zero** outgoing edges.

## Root cause

The receiver survives the capture layer intact — `query.ts` captures member calls with `object: (_)`, so `@reference.receiver` = `"Service(db)"` — and Case 0 in `receiver-bound-calls.ts` routes it to `resolveCompoundReceiverClass` because the text contains `(`. Confirmed by tracing a live analyze run:

```
case0 receiver=Service(db) member=do_work file=pkg/caller_inline.py
freecall fnExpr=Service fnDefType=undefined
```

The drop is in the free-call branch of `resolveCompoundReceiverClass`, which knew one shape only: *a function whose return-type binding names a class*. `Service` is a class and has no return-type binding, so the receiver resolved to nothing and the member call was dropped.

The two-step form works through a different mechanism entirely: `s = Service(db)` emits a `@type-binding.constructor` capture (`s` → `Service`), and Case 4 resolves the receiver from that binding. The inline form produces no such capture — nothing to bind.

## The fix

The rule is the same in every language — **constructing a class yields an instance of that class** — so it is stated once, in the compound receiver resolver. Only the surface syntax differs, and that is declared per language through a new `ScopeResolver.constructionSyntax` hook, matching how this file already gates language-varying behaviour (`stripReceiverCastExpressions`, `hoistTypeBindingsToModule`). Shared pipeline code names no language.

Probing the shape across languages showed the bug was wider than the issue reported:

| spelling | languages | dropped before this PR |
|---|---|---|
| `Service(db).m()` | Python | yes |
| `new Service(db).m()` | JS/TS, C# | yes |
| `Service.new.m()` | Ruby | yes |
| both forms | PHP, Swift, Dart, Kotlin | no — already resolved |

Wiring:

- `bare: true` — Python
- `keyword: 'new'` — JavaScript, TypeScript, C#
- `selector: 'new'` — Ruby, including the parenthesis-less `Service.new` spelling, which reaches the chain walker rather than the call branch

Opt-in is per language for two reasons. **Correctness:** `bare` would mistype `stat(&st).field` in C, where a struct and a function may share a name and the free call is not a construction. **Evidence:** Java, PHP, Swift, Dart and Kotlin already resolve this shape through their own capture-side paths, so they stay unwired rather than carrying a declaration that changes nothing — verified by diffing analyzer output between builds with and without the change, per language, not assumed. (Java was wired in an earlier revision of this PR on the strength of a faulty measurement; a re-run with the build discriminated each time showed it resolves identically either way, via the #2564 `object_creation_expression` capture rewrite, so the declaration was removed.)

The keyword gate also keeps bare factory calls honest: in a `new` language, `makeOther(db).doWork()` still resolves through the factory's return type and is never read as constructing a same-named class.

On the issue's repro: `route_inline` now emits `CALLS → Service.do_work`, and `impactedCount` goes 1 → 2.

## What is NOT fixed

The issue's second ask — `epistemic` degrading from `"exact"` to `"lower-bound"` when a receiver goes unresolved — is **not** addressed here, and it is not a small patch.

`computeEpistemicBoundary` keys only on the target symbol's own heritage edges (`IMPLEMENTS` / `METHOD_IMPLEMENTS` / `EXTENDS`). `Service.do_work` has no supertypes, so the boundary set is empty and it returns `"exact"` at the first check. Dropped references are not an input, and cannot be: unresolved outcomes accumulate in an in-memory `resolutionOutcomes[]` that reaches the pipeline result but is never written to the index, while `epistemic` is computed at query time against the index. Making it honest needs unresolved-receiver counts persisted first — worth its own issue.

Reporting that separately is the point of the issue's argument (a wrong count labelled `"exact"` gets acted on as authoritative), so this PR narrows the wrong count rather than closing the reporting gap.

Also unchanged, and worth its own issue: in Python a construction emits no edge at all — `Service(db)` produces no `CALLS` to the class or to `__init__`, in either spelling — while TypeScript, Java and C# do emit one. That asymmetry is independent of receiver typing.

## Tests

- `python-inline-constructor-receiver` — inline resolves, two-step still resolves, no cross-class fan-out.
- `typescript-inline-constructor-receiver` — inline `new`, a plain `.js` file so the javascript provider is covered too, two-step, and the factory guard.
- `ruby-inline-constructor-receiver` — `Service.new` with and without an argument list, plus two-step.
- With the source change stashed, every inline case fails and the factory/two-step cases still pass.
- No Kotlin fixture: its cases passed without the change, so they would document coverage this PR does not provide.
- Full `test/integration/resolvers` + `test/unit/scope-resolution`: **4234 passed, 1 skipped**. The rule is shared across languages, so cross-language no-regression coverage matters more than the new cases.
- Python and Ruby captures goldens regenerated — additions only, no existing digest changed, confirming capture output is untouched and the change is resolution-side.
- `tsc --noEmit` clean; re-verified after rebasing onto the branch's merged `main`.

https://claude.ai/code/session_01RqnsJK3Cnbu3bjdZbzgMMP

---

## Review round 2 — findings addressed

A multi-engine review of this PR (Claude swarm + ce personas, Codex `gpt-5.6-sol`, plus executed before/after probes) found three blockers and several gaps. Each is addressed in its own commit:

| Commit | Finding |
|---|---|
| `a438c446` | **Regression:** the Ruby `.new` selector fired on instance receivers too, so `factory.new.run` replaced a correct `Product#run` edge with `Factory#run`. Now gated on the receiver naming the class itself. |
| `6bc94970` | `new Box<string>().get()` still dropped the member edge — template arguments are now stripped before the class lookup. |
| `ae7476fe` | `new Service().inner.deep()` dropped it too: construction in the chain-**head** position never seeded the walk. |
| `1077c387` | The keyword matched exactly one space, so `new\tService(…)` and newline-separated `new` were missed. |
| `9e8f1981` | `new ns.Service().m()` was routed into member resolution by the dot-split before construction was considered. |
| `7056f020` | Java's wiring was dead config — removed, with the contract recording why. |
| `c62e00ed` | The selector rule was written twice; the option type was hand-copied from the contract. Both fixed. |
| `7327cb32` | C# had no fixture despite being the only load-bearing keyword language; the TS two-step assertion could not distinguish `Service` from `LegacyService`; nothing guarded the deliberate wiring set. |
| `89d42c26` | **`INCREMENTAL_SCHEMA_VERSION` → 23.** Without it every existing index passes the reuse gate and never gains this fix. |
| `a3365c1c` | **Re-baselined the corpus fingerprints** (python, plus ruby/typescript/csharp in `scope-capture`) — last, so the fixture set was final. |

Two review claims were **refuted by probe and deliberately not acted on**: a Python same-name-factory shadow fabricating edges (built four variants, no false edge — and in a `new`-keyword language a bare call cannot reach the construction rule at all), and a missing `toBeDefined()` guard that is in fact present.

One **known limitation** is now documented on the contract field and pinned by a test: a class-level `def self.new` returning another type is still read as construction, because the scope model records no per-member staticness, so `def new` and `def self.new` are indistinguishable at this layer.

Verification: `test/integration/resolvers` + `test/unit/scope-resolution` + the schema-version test — **4253 passed, 1 skipped**. All three bench gates pass locally (`measure --check`, `import-target-fingerprint --check`, `scope-capture --check`).


magyargergo (commented):
# Multi-engine review — PR #2737 (`fix(resolution): type inline constructor receivers in every spelling`)

**Head `a87c64e5` · base `79ff44df` · 23 files, +378/−4**

## Verdict

**not production-ready** — merge state `checks failing`; branch hygiene `merge-from-main commit present but harmless and merge-safe`. The blockers are narrow and all fixable inside this PR. The design is right: construction *syntax* is declared per-language on `ScopeResolver` and the rule "constructing a class yields an instance of that class" is stated in shared code, which is exactly what AGENTS.md's no-language-names-in-`src/core/ingestion/` rule asks for. The fix demonstrably works for Python, TypeScript/JavaScript, Ruby, Java and C# (measured before/after, not assumed). Three things stop it: (1) CI is red and will stay red after fixing what CI currently reports, because **three** corpus fingerprint baselines drifted and the job aborts after surfacing the first; (2) `INCREMENTAL_SCHEMA_VERSION` was not bumped, so every existing index silently never picks the fix up — defeating the PR's stated purpose for the entire installed base; (3) the Ruby `.new` selector shortcut converts one previously-correct edge into a confidently wrong one. Several of the scarier suspicions raised by the review lanes were **empirically refuted** and should not drive changes.

## Engine breakdown

| Engine | Legs | Landed |
|---|---|---|
| **Claude** | GitNexus swarm: risk, test/CI, branch-hygiene, docs/DoD | 4/4 |
| **Claude** | ce personas: correctness, adversarial, testing, maintainability, standards, api-contract | **5/6** — only `adversarial` left no artifact; `correctness` landed late (it ran its own before/after probe matrix) and its findings are folded in below |
| **Codex** (gpt-5.6-sol, xhigh) | swarm, ce, native adversarial | **3/3** — all three legs produced findings |
| **Coordinator** | executed before/after analyzer probes | 7 probe files; these outrank lane opinion |

**Codex sandbox caveat:** Codex's bubblewrap namespace was denied, so its Node/test/fingerprint commands never executed. Every Codex finding below is `[code-read]` unless the coordinator independently reproduced it. Claude's testing and standards lanes *did* execute (built the worktree, ran the CI bench commands, `git apply -R`'d the fix and re-ran the tests).

**Cross-engine agreement:** the baseline drift (both engines, 7 lanes) and the Ruby selector defect (Codex adversarial + Codex swarm F3 predicted it; Claude risk lane described the same mechanism but rated it unrealistic; coordinator reproduced it) are the two genuinely cross-engine findings. Generic constructors and the schema-version bump were each single-engine but coordinator-verified.

## What is solid

- **Architecture.** No language identity branches in the new shared code; behaviour is entirely data-driven off `constructionSyntax`. Naming languages in *doc comments* matches this file's existing house style (`constructorCallTargetsClass`, `stripReceiverCastExpressions`, `detectInterfaceImplementations` all do it). Compliant.
- **Contract addition is non-breaking.** All 16 `ScopeResolver` implementors enumerated; the field is optional, no `Required<>`/`Omit<>` utility forces it, no test double builds a bare `ScopeResolver` literal. `npx tsc --noEmit` produces an *identical* pre-existing error set on base and head.
- **C# wiring is load-bearing and works** — coordinator measured `new Svc().DoWork()` resolving on head and **absent** on base. C# is the only keyword-wired language whose behaviour genuinely depends on this change, and it has no test (see gaps below).
- **Tests actually prove the fix.** Reverting only the 9 fix source files and re-running: all four inline cases fail, all four guard cases (two-step, bare-factory) still pass. No false-confidence tests.
- **REFUTED — Python same-name factory shadowing a class.** Codex adversarial's headline P1 (`def Service()` locally shadowing an imported class `Service`; `Service().run()` mistyped as the class) was built exactly as specified and run on the head build. Only edge emitted: `route → Function:pkg/app.py:Service`. **No** `Service.run` edge. The cross-module variant was refuted separately. `findClassBindingInScope` correctly returns the local function binding, and `isClassLike` holds. Codex swarm F2 is the same claim and is likewise refuted. **Do not act on it.**
- **REFUTED — Python `bare:true` fabricating edges from class/factory name collisions** (Claude risk lane "Suspicion A", Codex adversarial): probed, no false edge.
- **REFUTED — keyword text-match false positives.** `new.target` routes to the dot branch, `new (Dynamic)()` yields a non-matching callee, `new { … }` is not a class. All return `undefined` safely.
- **REFUTED — `factoryCall!` used without a prior `toBeDefined()`** (Claude test/CI lane claim): the guard is present; the lane misread the file. Dropped.
- **Golden capture snapshots** (`{python,ruby}-captures-golden/expected-captures.json`) were regenerated correctly and are **purely additive** — zero changes to any pre-existing entry's digest or `captureGroups` count, confirming no capture-side side effect.
- **Branch hygiene:** mergeable, no conflicts, no bidi/hidden Unicode, no secrets, no dependency churn, all 23 files in `#2708` scope, `CHANGELOG.md` correctly untouched.

## Blocking findings

### B1 — P1 — Three bench fingerprint baselines are stale; fixing only the one CI reports leaves CI red `[reproduced]`

**Trigger:** adding `test/fixtures/lang-resolution/{python,ruby,typescript}-inline-constructor-receiver/`. Both bench harnesses glob **corpus-wide** by directory prefix (`bench/python-scope/measure.mjs:38`, `bench/scope-capture/measure.mjs:76` — `entry.name.startsWith(prefix + '-')`), so any new fixture directory changes the fingerprint.

**Evidence:** CI job `tests / benchmarks (GITNEXUS_BENCH)` fails at `.github/workflows/ci-tests.yml:478-482` with `got f120df92… expected 36e29abc…`. The step uses `set -e -o pipefail`, so the **next** step never runs. The coordinator ran it manually on the PR head:

```
$ node --import tsx bench/scope-capture/measure.mjs --check
[scope-capture --check] FAIL: ruby: capture fingerprint drift (got 45817693…, expected 070e4e11…)
[scope-capture --check] FAIL: typescript: capture fingerprint drift (got 15932d49…, expected 281e9548…)
```

So the true remediation is **three** baselines, not one: `bench/python-scope/baseline-fingerprint.txt` plus the `ruby` **and** `typescript` entries in `bench/scope-capture/baselines.json`. Regenerate both without `--check`, commit deliberately, then re-run both with `--check`.

Note the author *did* correctly regenerate the other golden mechanism (`*-captures-golden/expected-captures.json` via `UPDATE_GOLDEN=1`) — these two corpus-wide fingerprint baselines are a separate, previously-hit trap (#2023/#2033).

**Corroboration:** cross-engine and near-unanimous — Claude swarm risk F1+F2, Claude docs/DoD F1, Claude ce-standards F1, Claude ce-testing F1, Codex swarm F4, Codex ce #2, Codex adversarial #3, plus coordinator execution.

### B2 — P1 — `INCREMENTAL_SCHEMA_VERSION` not bumped, so no existing index ever gains the fix `[code-read]`

**Trigger:** any user who already has a v22 index runs `analyze` after upgrading.

**Evidence:** `gitnexus/src/storage/repo-manager.ts:546` stays at `22`, and `repo-manager.ts` is not among the PR's 23 changed files. `run-analyze.ts:1499-1507` gates `isIncremental` on `existingMeta.schemaVersion === INCREMENTAL_SCHEMA_VERSION`, and `:1233-1244` is a same-commit/clean-tree fast path that skips recompute entirely. This PR changes *which CALLS edges are emitted for source whose content has not changed* — exactly the class the file's own version-history block says must bump. v22's entry: "These change what is emitted for source whose CONTENT has not changed… Force a full re-analyze." v20's (a pure resolution-rule change, the closest analogue): "A v19 index holds those false CALLS/ACCESSES on every unchanged file and would keep serving them through the reuse gate."

Precedent is unbroken across the five most recent resolution PRs, all of which bumped it: `89ea233e` (#2723→v22), `0ce78802` (#2699→v21), `e307286d` (#2695→v20), `4906daf2` (#2693), `d3d4fa31` (#2563). Ruled out alternate safety nets — `ANALYSIS_FEATURES` covers only Spring/class-annotation dimensions; nothing else forces a rebuild.

**Impact:** `impact(direction:'upstream')` and `context()` keep returning the pre-fix, incomplete result set with no warning until something unrelated forces a full re-analyze. The PR's entire stated purpose does not reach the installed base.

**Fix:** bump to `23` and add a `v23:` rationale paragraph in the same comment block, in this PR.

*Single-lane origin (Claude ce-api-contract) but coordinator-verified against the file's own stated rule and five commits of precedent — postable under severity-over-corroboration. Not inlineable: `repo-manager.ts` is not in the diff.*

### B3 — P2 — Ruby `.new` selector shortcut turns a correct edge into a wrong one `[reproduced]`

**Trigger:** a Ruby class that defines an *instance* method named `new` (legal, if unusual), reached through an instance receiver.

```ruby
class Product; def run; end; end
class Factory
  def run; end
  def new; Product.new; end     # instance method named `new`
end

def via_class_constant; Factory.new.run; end   # correct target: Factory#run
def via_instance; factory = Factory.new; factory.new.run; end   # correct target: Product#run
```

**Measured, two builds of the same worktree** (discriminated by `grep -c resolveConstructionExpressionClass dist/…/compound-receiver.js` = 0 vs 2):

| | base `79ff44df` | head `a87c64e5` |
|---|---|---|
| `via_class_constant` | `Product.run` (wrong) | `Factory.run` ✅ **improved** |
| `via_instance` | `Product.run` ✅ | `Factory.run` ❌ **regression** |

Report both directions honestly: the selector rule **fixes** the class-constant spelling (an unadvertised gain worth locking in with a fixture) and **breaks** the instance spelling.

**Root cause** — `compound-receiver.ts:504` (and the same rule again at `:324`):

```ts
if (options.constructionSyntax?.selector === memberName && isClassLike(currentClass.type)) continue;
```

`currentClass` is a class *definition* in both cases — when the receiver is the class constant `Factory` and when it is an instance typed as `Factory`. `isClassLike(currentClass.type)` therefore cannot distinguish them, so the shortcut fires for instance receivers too and skips the legitimate member-return-type lookup.

**A second, distinct trigger** (Claude ce-correctness, reproduced independently): the shortcut also runs *before* the class-scope `typeBindings` lookup, so it discards a **recorded** return type for `new`. Ruby records these — YARD `@return` (`ruby/captures.ts:378-395`) and constructor-return inference (`:473-509`). With `# @return [Impl]` on `def self.new`, `Factory.new(:a).run` emits only the `Factory.new` call edge; the `Impl.run` edge is dropped entirely. That is a *class-constant* receiver, so the class-vs-instance distinction alone does **not** fix it.

**Fix direction (covers both triggers):** make the guard a **fallback rather than a short-circuit** — run the existing `typeBindings`/`retType` lookup first and only use `return objClass` (site `:324`) / `continue` (site `:504`) when `retType === undefined`. Then additionally carry whether the receiver denotes a class constant or a typed instance, for the remaining case where no return type is recorded. Add fixtures for `Factory.new.run`, `factory.new.run`, and a YARD-annotated `self.new`, each with conflicting targets.

**Corroboration:** Codex adversarial predicted it analytically, Codex swarm F3 filed it, Claude risk lane named the mechanism (and rated it unrealistic — the reproduction overrules that), Claude ce-maintainability listed it as a residual risk. Coordinator reproduced end-to-end.

## Non-blocking findings and coverage gaps

- **P2 `[reproduced]` — Generic constructor receivers still drop the member edge.** `new Box<string>().get()` emits **no** edge to `Box.get` while `new Plain().get()` resolves (measured on the head build). `calleeName` arrives as `"Box<string>"`, which matches no class binding at `compound-receiver.ts:139`. **Not a regression** (no edge before either) — an incompleteness — but `new Foo<T>()` is ordinary in the three keyword languages this PR wires, so the fix lands on a narrower slice of real code than advertised. The repo already has `stripTemplateArguments` (`src/core/ingestion/utils/template-arguments.ts`), used for exactly this by the sibling `resolveClassBindingForName` in `passes/receiver-bound-calls.ts:121`. *Codex ce #1 + Codex swarm F1 + Claude ce-correctness; coordinator reproduced. **Severity disagreement, stated openly:** both Codex legs rated this **P1/blocks-merge**; it is carried here as P2 because no edge existed before either, so nothing regresses. Given the PR title says "every spelling", the P1 reading is defensible — the maintainer should make the call.*
- **P2 `[code-read]` — Qualified/dotted construction expressions never reach the new resolver.** `compound-receiver.ts:282-283` splits on the last `.` **before** construction is considered, and `resolveConstructionExpressionClass` is only called on the `lastDot === -1` branch (`:301`). So `models.User(db).save()`, `new ns.Service().m()` and fully-qualified Java/C# types stay unresolved. Same incompleteness class as generics. *Codex swarm F1 only — single-engine, code-read, not reproduced; body-only.*
- **P2 `[code-read]` — Keyword matching accepts exactly one space.** `compound-receiver.ts:127` builds `` `${syntax.keyword} ` `` and matches with `startsWith`, so `new\tService(db).doWork()`, a newline, or a comment between `new` and the type loses the edge. A false negative, not a false positive. *Codex adversarial + Codex ce residual — single-engine.*
- **P3 `[reproduced]` — Java's `constructionSyntax` wiring appears to be dead config, and the PR description's per-language claim is wrong for Java.** `languages/java/scope-resolver.ts` declares `{keyword:'new'}`, but Java resolves `new Svc().doWork()` **identically with and without this change** — re-measured with the build discriminated each time (`grep -c resolveConstructionExpressionClass dist/…` = 2 vs 0; the member edge is present in both). `java/captures.ts:235-246` (#2564, untouched here) already rewrites an `object_creation_expression` receiver to the constructed type's simple name, so the raw `new Svc()` text never reaches this resolver. The ce-correctness lane's decisive evidence: Java resolves `new Box<User>().doWork()`, which the keyword path *cannot* do (it would look up `Box<User>` — see the generics gap above), proving the resolution comes from the capture rewrite. *Correction notice: an earlier coordinator probe claimed Java was broken on base; that run was unreliable and the ce-correctness lane was right. The PR body should stop listing Java under "needed the fix", and the declaration should either be dropped or annotated as defensive-unreachable.*
- **P3 — C# is functional but unguarded.** Wired `{keyword:'new'}` with no fixture, while its behaviour genuinely depends on this change. A typo or narrowed guard in that provider file would pass the whole suite silently.
- **P3 `[reproduced]` — construction in the chain HEAD position is still dropped.** `new Service().inner.doWork()` emits only the Class construction edge, no `doWork` member edge: the head segment is passed through `stripCallParens`, yielding `"new Service"`, which matches no binding, so `currentClass` is never seeded (`compound-receiver.ts:450-456`). Python's bare form is immune (it strips to `Service`), so coverage is asymmetric across the languages the PR claims to cover. Fix: seed the head with `resolveConstructionExpressionClass(headMemberName, …)`. *Claude ce-correctness F4.*
- **P3 — The selector rule is stated twice**, at `compound-receiver.ts:318-326` (`return objClass`) and `:500-506` (`continue`), against differently-named locals — while the new helper's doc comment claims "this is the one place the rule … is stated". Extracting a shared `isConstructionSelectorHop(name, cls, options)` predicate would make the doc literally true and is also where B3's class-vs-instance distinction naturally lands.
- **P3 — `constructionSyntax` shape hand-copied** into `ResolveCompoundReceiverOptions` (`compound-receiver.ts:97-101`) rather than derived from the contract (`contract/scope-resolver.ts:792-799`). Consistent with the file's existing convention, but it is the first *object-shaped* duplicate, and a future sub-field would type-check and be silently ignored. `type ConstructionSyntax = NonNullable<ScopeResolver['constructionSyntax']>` would close it.
- **P3 — Multi-field precedence undocumented.** `{bare:true, keyword:'new'}` together works correctly by inspection (keyword tried first, bare as fallback) but is neither documented as a guarantee nor tested. Either state the precedence in the JSDoc or make combinations statically impossible with a discriminated union.
- **P3 — `typescript.test.ts:3243-3248`** ("keeps the two-step spelling") has only `toBeDefined()` + `toContain('Service')` with no `targetFilePath` — and `'LegacyService'.includes('Service')` is `true`. Its sibling three lines up, and the Python and Ruby equivalents, all pin `targetFilePath`. Low real risk (no live path from `src/app.ts` to `legacy/svc.js`), but it is the one assertion in the new block without an exact backstop. *Note: the lane's stronger claim that the primary inline test was also substring-only was refuted — that test does pin `targetFilePath: 'src/svc.ts'`.*
- **P3 — Fixture placement.** `typescript-inline-constructor-receiver/legacy/*.js` sits under a `typescript-`-prefixed directory. The `typescript` bench filters by extension so the `.js` files are excluded, and the `javascript` bench never visits `typescript-*` directories — so those two files are invisible to **both** capture-fingerprint benches. The precedent for mixed fixtures (`express-route-mapping/`) uses a prefix-neutral name.
- **P3 — No regression guard** would catch a future accidental `constructionSyntax` wiring of PHP/Swift/Dart/Kotlin, the four languages deliberately left unwired.
- **P3 — ARCHITECTURE.md hook table** (~line 259) gains no `constructionSyntax` row. **Demoted from the docs/DoD lane's P2 after coordinator verification:** the table already omits `constructorCallTargetsClass` (the directly adjacent hook), `stripReceiverCastExpressions`, `freeCallsRequireInstanceOwnership` and `resolveThisViaEnclosingClass`. "Every hook gets a row" is not an established invariant, and this PR is consistent with how its closest precedent shipped. Nice-to-have, not a DoD violation.
- **P3 — Branch/PR hygiene.** Merge commit `13807310` ("Merge branch 'main'") is unnecessary for this scope; a rebase onto `79ff44df` would be cleaner. Separately, "Fixes #2708" auto-closes an issue with **two** asks while the epistemic-reporting half is explicitly unresolved and no successor issue exists — consider "Refs #2708" or open the follow-up first.

## CI status (head `a87c64e5`)

| Check | Result |
|---|---|
| `tests / benchmarks (GITNEXUS_BENCH)` | **FAIL** — this PR's fault (B1) |
| `CI Gate` | **FAIL** — aggregate of the above; the single required check for branch protection |
| `Vercel` | FAIL — "Authorization required to deploy", known non-actionable fork-auth red |
| `quality/{format,lint,typecheck}`, `Analyze (js-ts, python)`, `Build & Push` (both), CodeQL, gitleaks, e2e check, Validate PR title, check-description, autofix | PASS |

Browser E2E was correctly skipped (the web package is unchanged). `git diff --check` is clean. Mergeability is `MERGEABLE`; the block is checks, not conflicts.

## Suggested fix order

1. **Bump `INCREMENTAL_SCHEMA_VERSION` to 23** with a `v23:` rationale (B2) — cheapest, and without it the rest of the PR does not reach existing users.
2. **Fix the Ruby selector class-vs-instance distinction** (B3), extracting the shared predicate while you are there; add fixtures for **both** `Factory.new.run` and `factory.new.run`.
3. **Normalize template arguments** before the class lookup at `compound-receiver.ts:139` and add a generic-constructor fixture; optionally handle the qualified/dotted and keyword-trivia spellings in the same pass.
4. **Add a C# fixture** (the one keyword language that genuinely depends on this change), decide whether Java's declaration stays as defensive-unreachable or is dropped, correct the PR body's per-language claim for Java, and pin `targetFilePath` in the TS two-step assertion.
5. **Re-baseline last** (B1), after the fixture set is final: `node --import tsx bench/python-scope/measure.mjs` and `node --import tsx bench/scope-capture/measure.mjs` with no `--check`, commit both, then re-run both **with** `--check` and confirm `tests / benchmarks` goes green.

---

*Automated multi-tool digest (Claude swarm + Claude ce personas + Codex gpt-5.6-sol swarm/ce/adversarial, merged with coordinator-executed probes). Verify each finding before acting on it.*

gitnexus/test/fixtures/lang-resolution/python-inline-constructor-receiver/app.py
```diff
@@ -0,0 +1,11 @@
+from models.user import User
```
magyargergo, line 1: **P1 — adding this fixture directory drifts THREE CI fingerprint baselines, not one.**

Both bench harnesses glob corpus-wide by directory prefix (`bench/python-scope/measure.mjs:38`, `bench/scope-capture/measure.mjs:76` — `entry.name.startsWith(prefix + '-')`), so every new `python-*` / `ruby-*` / `typescript-*` fixture directory changes a committed fingerprint. This PR adds all three and updates none of the baselines.

CI reports only the Python one because the step aborts before the cross-language step runs. Running that next step manually on this head:

```
$ node --import tsx bench/scope-capture/measure.mjs --check
[scope-capture --check] FAIL: ruby: capture fingerprint drift (got 45817693…, expected 070e4e11…)
[scope-capture --check] FAIL: typescript: capture fingerprint drift (got 15932d49…, expected 281e9548…)
```

So fixing only `bench/python-scope/baseline-fingerprint.txt` will just surface the next failure. Re-baseline all three **after** the fixture set is final:

```
node --import tsx bench/python-scope/measure.mjs      # rewrites baseline-fingerprint.txt
node --import tsx bench/scope-capture/measure.mjs     # rewrites baselines.json (ruby, typescript)
```

then re-run both with `--check`. The separate `*-captures-golden/expected-captures.json` goldens *were* regenerated correctly — this is the other, corpus-wide mechanism, the one #2023/#2033 also hit.

`[reproduced]`

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -429,6 +497,13 @@ export function resolveCompoundReceiverClass(
     const segment = parts[i];
     if (segment === undefined) break;
     const memberName = stripCallParens(segment);
+    // Selector-form construction mid-chain — `Service.new.do_work`, and
+    // the parenthesis-less spelling `Service.new` that never reaches the
+    // call branch above (#2708). Constructing yields an instance of the
+    // same class, so the walk continues from `currentClass` unchanged.
+    if (options.constructionSyntax?.selector === memberName && isClassLike(currentClass.type)) {
```
magyargergo: **P2 — this shortcut also fires for *instance* receivers, turning a previously correct edge into a wrong one.**

`currentClass` is a class *definition* both when the receiver is the class constant `Factory` and when it is a variable typed as `Factory`, so `isClassLike(currentClass.type)` cannot tell them apart. The `continue` then skips the legitimate member-return-type lookup. Same rule, same gap, at line 324.

Reproduced with before/after builds of this worktree:

```ruby
class Product; def run; end; end
class Factory
  def run; end
  def new; Product.new; end     # instance method named `new` — legal Ruby
end

def via_class_constant; Factory.new.run; end        # correct: Factory#run
def via_instance; f = Factory.new; f.new.run; end   # correct: Product#run
```

| | base `79ff44df` | head `a87c64e5` |
|---|---|---|
| `via_class_constant` | `Product.run` (wrong) | `Factory.run` ✅ improved |
| `via_instance` | `Product.run` ✅ | `Factory.run` ❌ **regression** |

The rule genuinely **fixes** the class-constant spelling — worth locking in with a fixture. But it replaces a correct edge with a confidently incorrect one for the instance spelling.

There is a second, distinct trigger: because this guard runs *before* the class-scope `typeBindings` lookup, it also discards a **recorded** return type for `new`. With a YARD `# @return [Impl]` on `def self.new`, `Factory.new(:a).run` loses the `Impl.run` edge entirely — and that receiver is a class constant, so a class-vs-instance check alone would not fix it.

**Fix covering both:** make this a fallback rather than a short-circuit — consult the existing `typeBindings`/`retType` lookup first and only `continue` (here) / `return objClass` (line 324) when `retType === undefined`; then add the class-constant-vs-instance distinction for the remaining unrecorded case. Add fixtures for `Factory.new.run`, `factory.new.run`, and a YARD-annotated `self.new`, each with conflicting targets. Since the identical check exists twice, this is also the natural moment to extract one shared predicate — the helper's own doc comment claims the rule is stated in exactly one place.

`[reproduced]`

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -90,6 +91,53 @@ interface ResolveCompoundReceiverOptions {
...
+  } else if (syntax.bare === true) {
+    calleeName = fnExpr;
+  }
+  // A `new`-keyword language reaching here without the keyword is an
+  // ordinary free call, not a construction — resolving it to the class
+  // would fabricate an edge (a factory function may share the class's
+  // name).
+  if (calleeName === undefined || calleeName.length === 0) return undefined;
+
+  const cls = findClassBindingInScope(inScope, calleeName, scopes);
```
magyargergo: **P2 — generic constructors reach this lookup with template arguments attached, so the member edge is still dropped.**

`calleeName` arrives as `"Box<string>"`, which matches no class binding, so `new Box<string>().get()` resolves nothing.

Reproduced on this head build:

```ts
export class Box<T> { get(): number { return 1; } }
export class Plain   { get(): number { return 2; } }
export function viaGeneric(): number { return new Box<string>().get(); }
export function viaPlain():   number { return new Plain().get(); }
```

```
| viaGeneric | Class:src/box.ts:Box          |   <- construction edge only
| viaPlain   | Method:src/box.ts:Plain.get#0 |   <- member edge RESOLVED
```

`viaGeneric` gets no edge to `Box.get`. Not a regression (there was no edge before either), but `new Foo<T>()` is an ordinary spelling in the keyword languages this PR wires, so the fix covers a materially narrower slice of real code than the title's "every spelling" suggests.

Fix: normalize template arguments before the lookup. The repo already exports `stripTemplateArguments` (`src/core/ingestion/utils/template-arguments.ts`) and the sibling resolver in the same call path uses it for exactly this — `resolveClassBindingForName` at `passes/receiver-bound-calls.ts:121`. Please add a generic-constructor fixture alongside it.

Severity note: both Codex legs rated this **P1/blocks-merge**; it is carried as P2 here only because nothing regresses. Your call.

`[reproduced]`

gitnexus/test/fixtures/lang-resolution/python-inline-constructor-receiver/app.py
magyargergo: Fixed in a3365c1c — and it was worse than this comment said: **four** baselines, not three.

Adding the Ruby fixture (a438c446) and the C# fixture (7327cb32) later in the series moved `ruby` again and `csharp` for the first time, on top of the `python`/`ruby`/`typescript` drift identified here. That is exactly why the re-baseline had to be the last commit — doing it when this comment was written would have left two more.

Final values:

```
bench/python-scope/baseline-fingerprint.txt   36e29abc… -> f120df92…
bench/scope-capture/baselines.json  ruby       070e4e11… -> fea3edf8…
                                    typescript 281e9548… -> cad25be9…
                                    csharp     e05dc274… -> 05a85bae…
```

Neither harness writes its own baseline — running without `--check` only prints, and the file is edited deliberately, which is what the script comment asks for.

Verified locally on all three gates, and **`tests / benchmarks (GITNEXUS_BENCH)` now passes in CI** (6m53s), which is the first time the cross-language step has run on this branch at all — it previously aborted at the python step.

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
magyargergo: Fixed in a438c446, though **not the way this comment recommended** — the recommended fix was tried first and is wrong for Ruby.

The suggestion was to make the guard a fallback behind the `typeBindings`/`retType` lookup, so a recorded return type for the selector wins. I implemented that, using `TypeRef.source` to prefer a *declared* return type over an inferred one. Tracing the actual bindings killed it:

```
[trace] chain seg=new isConst=true src=return-annotation cls=…:Class:Factory
```

Ruby records **body-inferred** return types as `return-annotation` too, so the discriminator does not discriminate — and with the fallback ordering, `Factory.new.run` resolved to `Product#run`, which is wrong in the opposite direction. The `memberType === undefined` variant fails the same way: a class that defines any `new` member always has a binding, so construction never won.

What shipped is the part of your diagnosis that holds: `isClassLike` cannot separate the class constant from a value of that class, so resolution now tracks which one it is sitting on. The head of a chain is a class constant only when it resolved straight to a class binding; every hop past it yields a value, so the flag clears.

```
via_class_constant  Factory.new.run   -> Factory#run   (unchanged, still the improvement over base)
via_instance        factory.new.run   -> Product#run   (regression fixed)
```

The `def self.new` case you raised **is not fixed**, and is now documented rather than papered over: the scope model records no per-member staticness, so `def new` and `def self.new` are indistinguishable at this layer. That is stated as a KNOWN LIMITATION on the contract field and pinned by a test asserting the current behaviour, so changing it later is deliberate. Fixing it properly needs the language provider to record staticness first.

Fixture `ruby-construction-selector` covers all three shapes.

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
magyargergo: Fixed in 6bc94970, and the same investigation turned up three more spellings of the same gap, each fixed in its own commit:

| Commit | Spelling |
|---|---|
| 6bc94970 | `new Box<string>().get()` — template arguments stripped before the lookup, reusing `stripTemplateArguments` exactly as you suggested |
| ae7476fe | `new Service().inner.deep()` — construction in the chain **head** never seeded the walk, so every later segment resolved to nothing |
| 1077c387 | `new\tService(…)` — the keyword matched exactly one space; now matched as a token followed by any whitespace |
| 9e8f1981 | `new ns.Service().m()` — the callee was split at its last `.` before construction was considered, routing a qualified type into member resolution |

On the severity disagreement you flagged (both Codex legs said P1/blocks-merge, the digest carried it as P2): moot now that it is fixed, but your reading was the better one — the title claims "every spelling", and generics are ordinary in all three keyword languages.

One related shape is **not** fixed and is called out in 9e8f1981's message: bare-form qualified construction (Python `models.User(db).save()`) emits no edges at all, not even a construction edge, so it is a namespace-import gap upstream of this pass rather than construction typing.

PR #2739, opened 2026-07-29
Expose call-site locations in context calls

benny338813:
Persists existing call-site metadata on CodeRelation rows and returns it from context outgoing calls. Covers direct C free-call fallback plus the shared scope-resolution emitters; enables configuration-aware consumers to filter candidates without guessing declaration locations.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -39,6 +39,7 @@ Commands and gotchas live under **Repo reference** below and in **[CONTRIBUTING.
 ## Reference docs
 
 - **[ARCHITECTURE.md](ARCHITECTURE.md)**, **[CONTRIBUTING.md](CONTRIBUTING.md)**, **[GUARDRAILS.md](GUARDRAILS.md)**
+- **[cdefine-inspector integration](CDEFINE_INSPECTOR_INTEGRATION.md):** public call-site-location contract, cross-repo reproduction, and regression commands. Read this before changing `CodeRelation` persistence or `context` outgoing calls.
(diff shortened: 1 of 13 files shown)


azizur100389:
Reviewed current head `5a484e10` against `upstream/main` `51095c19` in a fresh worktree. It merged cleanly; `git diff --check` passed; no hidden bidi controls were found. A normal install/build completed, 14 targeted tests passed with 32 environment-gated skips, and `npx tsc --noEmit` passed. GitHub currently shows only the external Vercel authorization failure.

Branch hygiene: `clean feature/fix PR`. Merge state: `mergeable` (behind main, no conflict).

Blocking finding:

- **Risk:** existing indexes are not forced to rebuild after `CodeRelation` gains the three call-site columns. The updated `context` query unconditionally selects `r.callSiteFilePath`, `r.callSiteLine`, and `r.callSiteColumn`; an index created by the current release has a fixed `CodeRelation` schema without those properties, so upgrading the CLI can make `context` fail against an otherwise current index instead of returning optional metadata.
- **Evidence to check:** `gitnexus/src/core/lbug/schema.ts` adds the columns; `gitnexus/src/mcp/local/local-backend.ts` unconditionally reads them; the PR has no change to `INCREMENTAL_SCHEMA_VERSION` in `gitnexus/src/storage/repo-manager.ts`. The integration test creates a fresh database with the new schema, so it does not exercise upgrade compatibility.
- **Recommended fix:** bump `INCREMENTAL_SCHEMA_VERSION` with a migration rationale so pre-change indexes are rebuilt, and add an upgrade/reuse regression proving an old-schema index cannot reach the new query path without reindexing.
- **Blocks merge:** yes.

Final verdict: `not production-ready`. Fresh-index behavior and call-site propagation are covered and pass locally, but the missing schema invalidation leaves the normal upgrade path unsafe. Please address the rebuild contract before merge.

Issue #2633, opened 2026-07-22
feat(spring): resolve Resource injection semantics


ShiningXu:
Parent tracking issue: #1225

Builds on: #2414

Related dependency: #2413

Status: remaining work

Goal
Model Spring's JSR-250 Resource injection semantics without treating Resource as ordinary type-only Autowired injection.

Scope
Recognize javax.annotation.Resource and jakarta.annotation.Resource on fields and single-argument setter methods. Resolve explicit resource names, default names derived from the field or bean property, Spring's name-first lookup and type fallback where applicable, and statically declared type overrides. Use names from stereotype beans and Bean factory methods when those providers are available. Dynamic or runtime-only lookup metadata such as lookup and mappedName should remain unresolved instead of producing guessed relationships. Resource collection injection must not fan out to every implementation.

Acceptance criteria
1. Both javax.annotation.Resource and jakarta.annotation.Resource are recognized.
2. Explicit Resource names resolve to matching named bean providers when statically known.
3. An omitted name uses the field or setter property name first, then falls back to type only where Spring permits it.
4. Static type overrides are honored, while dynamic or runtime-only names fail closed.
5. Named providers from stereotype annotations and Bean factory methods can participate in resolution as those inventories become available.
6. Collection-typed Resource injection does not emit Autowired-style collect-all fan-out edges.
7. Missing or ambiguous providers are represented conservatively instead of inventing one exact target.
8. Tests cover javax and jakarta variants, field and setter injection, explicit and default names, type fallback, ambiguity, static type overrides, and collection negative cases.

magyargergo:
This was modeled on `origin/main` by #2740 (`de84ad629`, merged 2026-07-31).

#2740 closed #2413 (`@Bean` factories) and also shipped the #2633 `@Resource` work: `javax`/`jakarta` field and setter injection, name-first resolution with conservative type fallback, static type overrides, fail-closed `lookup`/`mappedName`, named stereotype/`@Bean` providers, and no collection fan-out. Coverage is in `spring-bean-resource-pipeline.test.ts` and `spring-bean-resource-benchmark.test.ts`.

Closing this as completed by that merge.

PR #2740, opened 2026-07-29
feat(spring): index @Bean factories and @Resource injection

### LINKED ISSUE #2413: feat(spring): detect Bean methods and Configuration classes
Parent tracking issue: #1225

Status: remaining work

Goal
Treat Configuration classes and Bean factory methods as Spring bean declarations.

Scope
Detect Configuration classes, Bean annotated methods, explicit bean names, default method name based bean names, returned bean types, and factory method relationships.

Acceptance criteria
1. Bean factory methods create bean declarations in analysis results.
2. Bean names are recorded from annotation values when present and method names otherwise.
3. Returned types can be used by later injection resolution.
4. Tests cover explicit names, default names, interface return types, and concrete return types.

ShiningXu:
## Summary

- index Java and Kotlin Spring `@Bean` factory methods as synthetic Bean declarations, including default names, aliases, return types, factory dependencies, and MCP context/impact metadata
- support `javax.annotation.Resource` and `jakarta.annotation.Resource` with conservative name-first resolution, default-name type fallback, explicit type overrides, runtime lookup fail-closed behavior, and Kotlin use-site targets
- pre-index provider names and provided types in the DI phase to keep named Resource resolution sub-quadratic
- include Bean declaration drift in incremental write-set expansion and bump the Spring Bean inventory feature version
- add Java/Kotlin pipeline, MCP, incremental, argument-parser, conflict, collection, and lowercase Kotlin typealias coverage
- run the end-to-end Bean/Resource scaling test in the dedicated serial `GITNEXUS_BENCH` CI job

## Compatibility and behavior

- Java and Kotlin use the same resolved-FQN semantics; local annotations named `Bean` or `Resource` are ignored
- `@Resource` explicit names do not fall back to type, while default names may use a low-confidence type fallback
- `lookup` and `mappedName` remain unresolved unless they are the empty default
- collection factory parameters preserve fan-out semantics
- Bean factory methods are not also emitted as owner-class injection sites when combined with `@Autowired` or qualifier annotations
- lowercase Kotlin type aliases are preserved during type normalization

## Validation

- `npm run build`
- Spring suite: 22 files passed, 125 tests passed, 6 skipped
- `GITNEXUS_BENCH=1 npx vitest run --no-file-parallelism test/integration/spring-bean-resource-benchmark.test.ts`: 3/3 passed
- `npx tsc --noEmit`
- targeted ESLint, Prettier, and `git diff --check`
- pre-commit formatting, ESLint, and TypeScript checks passed
- full `npm test`: 15,238 passed, 99 skipped, 10 failed; the registry-race FTS failure passed 18/18 when rerun in isolation, while the remaining 9 failures are existing macOS path/Linux descriptor/ambiguous-impact baseline limitations outside the Spring change
- rebased onto upstream `main` at `79ff44df`; build, Spring suite, performance suite, and typecheck were rerun successfully after rebase
- final GitNexus `detect_changes`: 22 files, 75 symbols, 8 affected flows, HIGH risk (reviewed and explicitly accepted)

Closes #2413


magyargergo:
@gitnexus review

gitnexus/src/core/ingestion/pipeline-phases/di.ts
```diff
@@ -196,39 +232,54 @@ export const diPhase: PipelinePhase<DIOutput> = {
       if (typeof interfaceEntry === 'string') {
         for (const id of interfaceToImplementers.get(interfaceEntry) ?? []) structural.add(id);
       }
+      for (const id of providedTypes.get(candidate.language)?.get(candidate.targetTypeName) ?? []) {
```
magyargergo, line 308: **P0 — consumer-class → `@Bean`-declaration `INJECTS` edges cannot be persisted, and a previously-working edge is lost.** `[reproduced]`

Folding the synthetic `CodeElement:spring-bean:*` provider ids into `structural` makes the resolved provider a `CodeElement`, while for ordinary field/property/constructor injection `edgeSource` is undefined so `edgeSourceId === consumerClassId` (a `Class`, `di.ts:203-204`). `CodeRelation` declares exactly three `*->CodeElement` pairs — `FROM File`, `FROM Function`, `FROM Method`. **`FROM Class TO CodeElement` is absent** (enumerated across all 205 FROM/TO pairs). `RelPairRouter.route` validates only that each label is a valid node table, never that the *pair* is declared, so the CSV is written and `COPY`'d anyway; the COPY fails and the row-by-row fallback re-hits the same binder error inside a bare `catch { /* skip */ }`.

Executed against a real LadybugDB built from this branch's schema:
```
Prepare failed: Binder exception: Rel table CodeRelation does not contain Class-CodeElement from-to pair.
[load] Inserting 1 edges individually (missing schema pairs)
PERSISTED EDGES: [ DECLARES Method->CodeElement, DEFINES File->CodeElement ]   # INJECTS is GONE
```

**This is a regression, not only a gap.** `providerCandidates` (`di.ts:93-104`) is recall-first: `return recognized.length > 0 ? recognized : all`. Before this PR, `@Autowired MyService svc` where `MyService` is a POJO produced by a `@Bean` method had no provider metadata, so the fallback returned `all` and a `Class -> Class` edge persisted at 0.9. Now the bean node *is* in `providers`, the fallback no longer fires, and the single emitted edge is `Class -> CodeElement` — dropped. The working edge is replaced and lost, silently.

`spring-bean-resource-pipeline.test.ts` asserts against the in-memory graph from `runPipelineFromRepo` and never writes to LadybugDB — and roughly 10 of the ~14 pairs it asserts are exactly this shape.

**The real defect is in the UNCHANGED `gitnexus/src/core/lbug/schema.ts`**; anchored here because this is the in-diff line that creates the edge shape.

**Fix:** add `FROM Class TO CodeElement` to `RELATION_SCHEMA` (and audit the other provider-source labels that can now reach a bean node); bump `INCREMENTAL_SCHEMA_VERSION` (`repo-manager.ts:546`, currently 22) so existing indexes get the new pair instead of dropping edges forever; and make `RelPairRouter.route` reject undeclared pairs loudly so this class of bug fails fast.

gitnexus/src/core/ingestion/frameworks/spring/di-metadata.ts
```diff
@@ -42,6 +63,7 @@ export interface SpringDiClassFact<
   readonly classScopeId: ScopeId;
   readonly classAnnotations: readonly Annotation[];
   readonly injectionSites: readonly SpringDiInjectionSiteFact<Annotation, SiteKind>[];
+  readonly beanFactoryMethods?: readonly SpringBeanFactoryMethodFact<Annotation>[];
```
magyargergo, line 66: **P1 — parse-cache `SCHEMA_BUMP` not bumped for a `ParsedFile.captureSideChannel` shape change; a warm cache makes the whole feature silently inert, even under `--force`.** `[code-read]` — reached by **both engines** across 7 lanes.

This new `beanFactoryMethods?` field, plus `'Resource'` added to `CAPTURE_RELEVANT_ANNOTATIONS` at `:98` (which gates whole-fact emission — a class with a bare `@Resource Gateway selected;` produced *no* side-channel entry before and does now), changes what rides `ParsedFile.captureSideChannel`.

`parse-cache.ts:56-57` states the invariant verbatim: *"any future change to the `ParsedFile` serialization shape MUST bump SCHEMA_BUMP so both invalidate in lockstep."* `parse-cache.ts:122` is still `const SCHEMA_BUMP = 30;` — neither that file nor `gitnexus/package.json` is in this diff, so `PARSE_CACHE_VERSION` is unchanged.

The `SPRING_BEAN_INVENTORY_FEATURE.version 1 → 2` bump is the right lever for the DB half but not the parse half: it only sets `options.force = true`, and `parse-cache.ts:20-22` says the cache *"survives `--force` because it's content-addressed"*.

Trigger: analyze a Java/Kotlin Spring repo on `main`, check out this branch without an npm version change, re-run `analyze` against the same `.gitnexus/` storage. Every unchanged chunk hits the cache, `applyJavaCaptureSideChannel` restores pre-PR `springDiFacts`, and zero bean declarations / `@Resource` edges appear — then the forced rebuild persists that stale-derived graph at full confidence, with no warning.

Every predecessor in this exact family bumped it: #2494 19→20, #2632 20→21, #2678 27→28. `parse-cache.ts:100-108` explicitly says to re-check the value against `origin/main` immediately before merge.

**The fix location is the UNCHANGED `gitnexus/src/storage/parse-cache.ts:122`** — `SCHEMA_BUMP` 30 → 31 with the customary `// v31: …` comment line.

gitnexus/src/core/ingestion/pipeline-phases/di.ts
```diff
@@ -196,39 +232,54 @@ export const diPhase: PipelinePhase<DIOutput> = {
...
+        structural.add(id);
+      }
       structural.delete(consumerClassId);
+      structural.delete(edgeSourceId);
```
magyargergo, line 313: **P1 — a `@Bean` factory method is a viable provider for the bean it itself declares.** `[reproduced]` — **both engines** (Claude executed it against the real `di` phase; Codex reached it independently).

For a factory parameter, `edgeSourceId` is the **Method** id — but the provider is `CodeElement:spring-bean:${methodId}`, a different string, so nothing removes it from `structural`. Unlike the `di.ts:235` finding, `Method -> CodeElement` *is* a declared pair, so this false edge **does persist** and is visible to `impact` / `context`.

Executed on the composite-bean shape (`@Bean Filter loggingFilter()` + `@Bean Filter compositeFilter(List<Filter> filters)` — the `List<Filter>` / `List<HealthIndicator>` / `List<Converter>` idiom):
```
Method:Cfg.compositeFilter#1 -> CodeElement:spring-bean:Method:Cfg.loggingFilter#0   conf=0.8
Method:Cfg.compositeFilter#1 -> CodeElement:spring-bean:Method:Cfg.compositeFilter#1 conf=0.8   <<< SELF
```
On the decorator shape (`@Bean Gateway inner()` + `@Bean Gateway outer(Gateway delegate)`) it does second damage: the phantom self-candidate turns a single-viable selection into an ambiguous one, so the **correct** `outer → inner` edge collapses from 0.9 to 0.5 and is tagged `ambiguous candidates: inner, outer`. A confidence-thresholded consumer then discards a correct edge because of a fabricated sibling.

**Fix:** also exclude declarations owned by the edge source — `structural.delete(SPRING_BEAN_DECLARATION_ID_PREFIX + edgeSourceId)` — or, to keep this phase framework-neutral as its docblock requires, record the declaring node id on `DiProviderMatch` and let the phase exclude any provider declared by the edge source. The closest existing fixture (`@Bean ConcreteRepo aggregate(List<Gateway>)`) misses this because it returns a different type than it consumes.

gitnexus/src/core/ingestion/pipeline-phases/di.ts
```diff
@@ -98,14 +117,30 @@ export const diPhase: PipelinePhase<DIOutput> = {
...
+        providerNodes.set(node.id, node);
+        const namesByLanguage = providerNames.get(language) ?? new Map<string, Set<string>>();
+        for (const name of provider.names) {
+          const namedProviders = namesByLanguage.get(name) ?? new Set<string>();
+          namedProviders.add(node.id);
+          namesByLanguage.set(name, namedProviders);
+        }
+        providerNames.set(language, namesByLanguage);
+        if (provider.providedTypeName !== undefined) {
+          addProvidedType(providedTypes, language, provider.providedTypeName, node.id);
```
magyargergo, line 141: **P2 — factory providers are indexed only under the literal `providedTypeName` (plus its simple tail), with no assignability or qualified-name resolution.** `[code-read]` — all three Codex legs.

Calibration, since the two sub-cases differ sharply:

- **Unqualified `@Autowired Gateway` with `@Bean DefaultGateway defaultGateway()`** — the bean node is keyed on `DefaultGateway` so it never enters `structural` for target `Gateway`, but the implementing `Class` still does via `interfaceToImplementers`, and the recall-first fallback in `providerCandidates` yields a useful edge to it at 0.9. **Not** a missing edge.
- **`@Resource(name="defaultGateway") Gateway gateway`** — the bean carries the name but is not in `structural` (type-key mismatch), so `named` is empty; the name is explicit so `fallbackToType` is unset, and `di.ts:255` hits `continue` → **no edge at all**. Same for the common Kotlin `@Bean fun gateway() = DefaultGateway()`, where `kotlinBeanFactoryReturnType` infers the concrete type from the body.

**Fix:** resolve `providedTypeName` through the class/interface heritage index and register the factory provider under every assignable target type; for named selection, match the name first and then test assignability rather than string equality.

gitnexus/src/core/ingestion/frameworks/spring/resource-injection.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  return {
+    targetTypeName,
+    cardinality: 'single',
+    namedSelection: {
+      name: selectedName,
+      reason: `${explicitName ? 'resource name' : 'default resource name'} "${selectedName}"`,
+      ...(!explicitName ? { fallbackToType: true } : {}),
```
magyargergo: **P2 — `fallbackToType` is set for every implicit-name `@Resource`, including generic members whose type argument has already been erased, so the type fallback fans out to unrelated providers.** `[reproduced]` (parser probes executed; downstream fan-out traced)

`normalizeSpringBeanType` erases generics at `annotation-arguments.ts:183-187`. Executed probes: `List<Gateway>` → `List`, `List<Handler>` → `List`, `Map<String,Handler>` → `Map`, and `springResourceInjectionMatch('@Resource', 'items', 'List<Handler>', 'items')` → `{ targetTypeName: 'List', fallbackToType: true }`.

The same erasure runs provider-side (`di-metadata.ts:274-277`) and `addProvidedType` has no AMBIGUOUS tombstone, so `@Bean List<Gateway>` and `@Bean List<Handler>` both key on `List` and both receive 0.5 edges. Codex independently flagged the same fallback branch (`di.ts:247`) as too permissive.

**Fix:** keep the erased key for the name-first match, but do not set `fallbackToType` when the raw declared type contained a `<` — or have `normalizeSpringBeanType` report that it erased and gate on that.

gitnexus/src/core/ingestion/frameworks/spring/di-metadata.ts
```diff
@@ -231,18 +253,164 @@ export function createSpringDiMetadataAttacher<
...
             if (adapter.isInjectionAnnotationApplicable?.(annotation, site) === false) continue;
             const resolved = resolveFact(annotation, classScope.id);
             if (resolved !== undefined && INJECTION_ANNOTATIONS.has(resolved)) {
               injectionAnnotation = annotation;
-              break;
+            } else if (resolved !== undefined && SPRING_RESOURCE_ANNOTATIONS.has(resolved)) {
+              resourceAnnotation = annotation;
+            }
+          }
+          if (injectionAnnotation !== undefined && resourceAnnotation !== undefined) continue;
```
magyargergo: **P2 — the `@Autowired` + `@Resource` conflict `continue` fires before the member is claimed, so the legacy collection matcher still fans out.** `[code-read]` — **both engines**.

This `continue` is at `:386`; both `semanticallyOwnedMemberNames.add(...)` calls are below it, at `:390` and `:422`. So a conflicted member is never marked `SPRING_DI_CAPTURED_FIELD_PROPERTY` (read at `di-extractors/spring.ts:296`), and the legacy collect-all-implementers Property matcher still fans out for **collection-typed** fields — while scalar fields get nothing at all.

Worth noting separately: this also deletes an `@Autowired` edge that existed before this PR, for a case Spring itself resolves deterministically (`@Resource` wins).

The PR's own `ConflictingResourceConsumer` fixture uses a non-generic `Gateway`, for which `parseSpringCollectionType` returns null — so the test passes by luck rather than by covering the branch.

**Fix:** move the guarded `semanticallyOwnedMemberNames.add(site.memberName)` above the conflict `continue` so a conflicted member is claimed-and-suppressed, or resolve to `@Resource` semantics as Spring does. Extend the fixture to `@Autowired @Resource List<Gateway>`.

gitnexus/src/core/ingestion/di-extractors/index.ts
```diff
@@ -30,7 +30,13 @@ export interface DiInjectionMatch {
   namedSelection?: {
     name: string;
     reason: string;
+    /** Name-first frameworks may fall back to type only for implicit/default
+     * names. Explicit names remain strict. */
+    fallbackToType?: boolean;
   };
+  /** Most injection edges originate at the owning Class. Factory-method
+   * parameters preserve the Method as the semantic source. */
+  edgeSource?: 'owner-class' | 'site';
```
magyargergo, line 39: **P3 — the new `edgeSource: 'site'` (and the `CodeElement` provider target) makes the public `INJECTS` contract stale in four unchanged places.** `[code-read]`

- `gitnexus-shared/src/graph/types.ts:136-137` still reads *"Source = the consumer Class node … Target = a concrete provider Class node."*
- `gitnexus/src/mcp/tools.ts:237` ships `MATCH (c:Class …)-[r:CodeRelation]->(impl:Class) WHERE r.type = 'INJECTS'` as **the documented DI example**, which now silently returns fewer rows.
- `gitnexus/src/mcp/tools.ts:537` describes DI fan-out as `(consumer→implementer)`.
- `ARCHITECTURE.md:106` still describes the `di` phase as Class→Class.

Credit where due: the PR *did* correctly update the in-file docblock at `di.ts:6` to "owner/site → provider". These four are **UNCHANGED files**, so this is anchored on the in-diff `edgeSource` declaration that introduces the new shape.

**Fix:** update the `types.ts` doc comment, both `tools.ts` strings, and the ARCHITECTURE.md phase row in this PR.

magyargergo (commented):
## Multi-engine review — PR #2740 `feat(spring): index @Bean factories and @Resource injection`

**Engines.** Claude ran the GitNexus PR swarm (6 lanes: facts, branch-hygiene, risk-architect, test/CI, security-boundary, docs/DoD) plus 7 `ce-code-review` personas (correctness, adversarial, api-contract, testing, maintainability, performance, project-standards). Codex (GPT-5.5 / xhigh) ran the same swarm contract solo, the same `ce-code-review` contract, and a native adversarial red-team pass — **all three Codex legs returned real findings** (swarm: F1/F2/F3; ce: 7 findings with an independent validator batch confirming 7/7; adversarial: 3 findings). 16 lanes total, none empty. All line numbers below were re-checked in a checkout of `4a9f1150` against base `79ff44df`.

---

### What holds up

This is a well-built feature and several of the scariest hypotheses were chased and **refuted**:

- **The hand-rolled annotation parser is sound.** 30 adversarial inputs were *executed* against `annotation-arguments.ts` (comments inside the arg list, Java text blocks, `"a" + "b"` concatenation, `name = COUNT < MAX ? "a" : "b"`, Kotlin string templates, escaped quotes, FQ `@org.springframework...Bean("fq")`, `@Bean(name = {})`). Every one either parsed correctly or **failed closed** (`namesKnown: false`). The `<`/`>`-as-generic-delimiter policy cannot mis-split: an unmatched opener leaves `stack.length > 0` so `splitTopLevel` returns `null` (`annotation-arguments.ts:49`), and a stray closer at depth 0 is ignored.
- **No regression for existing `@Autowired` / `@Qualifier` users.** Three lanes traced the `di.ts` named-selection rewrite case-by-case and could not construct a pre-existing arrangement whose edge set, edge source, or confidence changes. The new `0.85` rung is reachable only via `fallbackToType`, which only `springResourceInjectionMatch` sets (`resource-injection.ts:109`).
- **The DECLARES-only drift walk is correct.** `INJECTS` is `isGraphWideRelationship` (`subgraph-extract.ts:99-102`), unconditionally `deleteAllInjects`'d and re-extracted from the full fresh graph every incremental run, so an `@Resource`-driven edge cannot go stale. `collectSpringBeanDeclarationDriftFiles` correctly compares both directions (`current \ persisted` and `persisted \ current`), and the encoded `reason` is a stable pure function of the sources — so the literal→constant name flip is caught. The prior-context worry carried over from #2494 does **not** recur.
- **`CodeElement` is not a new label.** It is already in `NodeLabel`, `NODE_LABELS`, `VALID_NODE_LABELS`, the LadybugDB schema, and the CSV writer, and is already emitted by COBOL/JCL/ORM/`emit-references`. The PR's one `local-backend.ts:2915` addition closes a **pre-existing** gap in `enrichCandidateLabels`, so that line is a fix, not a regression. The #2200 new-label registration checklist does not apply as a blocker.
- **Security lane clean.** ReDoS refuted (the `/^"((?:\\.|[^"\\])*)"$/s` alternation is fully disjoint); prototype pollution refuted (`JSON.parse` creates an own `__proto__` property and the field is never read); Cypher injection refuted (the MCP `${pattern}` is a ternary over two hardcoded literals, `$symbolId` is parameterized, the `run-analyze` query is fully static); the `ci-tests.yml` edit is `workflow_call`-only with `contents: read`, so no fork-PR exposure. No bidi controls; `git diff --check` clean.
- **Performance clean — zero findings.** The new drift query sits inside the pre-existing `if (beanSourceChanged)` gate (`run-analyze.ts:1885-1888`), so non-Java/Kotlin repos and all full analyzes pay nothing; it scans a partition strictly smaller than the `MATCH (c:Class)` scan already shipped next to it in #2494. The `providerNodes` merge is a rename of the old `classNodes` map — memory delta ~0. The `di-metadata` factory loop is bounded by actual `@Bean` method count.
- **Branch hygiene: `clean feature/fix PR`.** 0 commits behind `origin/main`, single conventional commit, no semantic conflicts on any touched path, no unrelated churn, `CHANGELOG.md` correctly untouched.
- **Real test coverage, not fixtures.** `spring-bean-resource-pipeline.test.ts` drives the real `runPipelineFromRepo` (real tree-sitter → real worker pool → real graph) and fails loudly, not silently, when unbuilt. `incremental-orchestration.test.ts` uses real `runFullAnalysis` + real LadybugDB. The new benchmark is a genuine sub-quadratic tripwire (`wallRatio < sizeRatio^1.5`) and is correctly registered in the gated CI job — which also closes a real gap, since the two pre-existing Spring benchmark files were never in that list.
- Also refuted: the `distWorkerExists()` guard is **not** this test family's convention (0 of 13 existing `spring-*.test.ts` files use it; 7 of 46 `runPipelineFromRepo` callers overall) — the swarm test lane's P1 on that is withdrawn. The 0-based `startLine` convention is not violated (the bean node copies the Method node's values verbatim). Synthetic-node id collision for overloaded `@Bean` methods is refuted (`#<arity>` + `~type` collision tags already disambiguate).

---

### Headline findings

**1. P0 — `Class -> CodeElement` is not a declared `CodeRelation` FROM/TO pair, so every consumer-class → `@Bean`-declaration `INJECTS` edge is silently discarded at persistence, and a previously-working `Class -> Class` edge is *lost*.** `[reproduced]` — Claude (`ce-correctness` CE-1 conf 90, `ce-adversarial` F1 conf 100), independently confirmed by the coordinator; Codex did not reach it.

`di.ts:235-236` folds the synthetic `CodeElement:spring-bean:*` provider ids into `structural`. For ordinary field / property / constructor injection `edgeSource` is undefined, so `edgeSourceId === consumerClassId` (`di.ts:203-204`) — a `Class` — while the resolved provider is now the bean `CodeElement`. `gitnexus/src/core/lbug/schema.ts` declares exactly three `*->CodeElement` pairs (`FROM File`, `FROM Function`, `FROM Method`); **`FROM Class TO CodeElement` is absent** — enumerated across all 205 FROM/TO pairs. `RelPairRouter.route` (`rel-pair-routing.ts:100-104`) filters only on *label is a valid node table*, never on *pair is declared*, so `rel_Class_CodeElement.csv` is written and `COPY`'d anyway; the COPY fails into a `warnings[]` array whose own comment says it has no consumer, and the row-by-row fallback re-hits the same binder error and drops it in a bare `catch { /* skip */ }`.

Executed against a real LadybugDB built from this worktree's schema:
```
Prepare failed: Binder exception: Rel table CodeRelation does not contain Class-CodeElement from-to pair.
[load] Inserting 1 edges individually (missing schema pairs)
PERSISTED EDGES: [ DECLARES Method->CodeElement, DEFINES File->CodeElement ]   # INJECTS is GONE
```
and the real `diPhase.execute` was confirmed to emit exactly that shape (`Class(Consumer) -> CodeElement(spring-bean:...) conf=0.95`).

**This is a regression, not only a gap.** `providerCandidates` (`di.ts:93-104`) is recall-first: `return recognized.length > 0 ? recognized : all`. Pre-PR, `@Autowired MyService svc` where `MyService` is a plain POJO produced by a `@Bean` method had `structural = {MyService class}`, no provider metadata, so the fallback returned `all` → a `Class -> Class` INJECTS edge at 0.9 persisted. Post-PR the bean node **is** in `providers`, `recognized` is non-empty, the fallback no longer fires, and the single emitted edge is `Class -> CodeElement` → dropped. The working edge is replaced and lost, silently.

The PR's own tests cannot catch this: `spring-bean-resource-pipeline.test.ts` asserts against the in-memory graph from `runPipelineFromRepo` and never writes to LadybugDB — and ~10 of the ~14 pairs it asserts (`ExplicitResourceConsumer->slowGateway`, `QualifierConsumer->slowGateway`, `DefaultResourceConsumer->gateway`, `CollectionResourceConsumer->gatewayList`, …) are exactly the shape that cannot persist. `spring-bean-mcp.test.ts` hand-seeds Cypher; the new `incremental-orchestration.test.ts` fixture uses a real DB but has no consumer class and only counts `DECLARES`.

*Fix:* add `FROM Class TO CodeElement` (and audit the other provider-source labels that can now reach a bean node) to `RELATION_SCHEMA` in `gitnexus/src/core/lbug/schema.ts`; because the rel table is created at index build, **bump `INCREMENTAL_SCHEMA_VERSION`** (`repo-manager.ts:546`, currently 22) so existing indexes get the new pair rather than dropping edges forever. Separately, `RelPairRouter.route` should validate against the declared pair set so a future missing pair fails loudly, and a static test asserting the pipeline's emittable `(fromLabel, toLabel)` set ⊆ `RELATION_SCHEMA` would kill this whole bug class.

**2. P1 — parse-cache `SCHEMA_BUMP` was not bumped for a `ParsedFile.captureSideChannel` shape change; a warm cache replays pre-PR facts and the entire feature silently no-ops (even under `--force`).** `[code-read]` — reached by **both engines** and 7 lanes: Claude `ce-correctness` CE-2, `ce-api-contract` F1, `ce-testing` #1, `ce-project-standards` F1, `swarm-facts-historian` F1; Codex `ce` #1 (validator-confirmed) and `swarm` F1 (both "blocks merge").

The PR changes what the Java/Kotlin capture side-channel contains: `di-metadata.ts:66` adds `beanFactoryMethods?` to `SpringDiClassFact`, and `di-metadata.ts:98` adds `'Resource'` to `CAPTURE_RELEVANT_ANNOTATIONS`, which gates whole-fact emission — a class like the PR's own `ExplicitResourceConsumer` (a bare `@Resource Gateway selected;`) produced **no** side-channel entry before and does now. That object is `ParsedFile.captureSideChannel` (`java/capture-side-channel.ts` → `parse-worker.ts:1540-1542`), and `parse-cache.ts:56-57` states the invariant verbatim: *"any future change to the `ParsedFile` serialization shape MUST bump SCHEMA_BUMP so both invalidate in lockstep."* `parse-cache.ts:122` is still `const SCHEMA_BUMP = 30;` — the file is not in the 22-file diff, and neither is `gitnexus/package.json`, so `PARSE_CACHE_VERSION` is unchanged.

The `SPRING_BEAN_INVENTORY_FEATURE.version 1 → 2` bump is the right lever for the DB half but **not** for the parse half: it only sets `options.force = true`, and `parse-cache.ts:20-22` says the cache *"survives `--force` because it's content-addressed"*. Net effect on a warm cache: the pipeline builds a graph from stale ParsedFiles and the forced full rebuild then persists that stale-derived graph with maximum confidence, with no warning.

Trigger: analyze a Java/Kotlin Spring repo on `main`, check out this PR without an npm version change, re-run `analyze` on the same `.gitnexus/` storage. Every unchanged chunk hits the cache, `applyJavaCaptureSideChannel` restores pre-PR `springDiFacts`, and zero bean-declaration nodes / `@Resource` edges appear. Exposed: same-version consumers (git checkout, pinned Docker, monorepo dev loops), the repo's own CI/dev caches. Every predecessor in this exact family bumped it — #2494 19→20, #2632 20→21, #2678 27→28 — and `parse-cache.ts:100-108` explicitly says to re-check the value against `origin/main` immediately before merge.

*Fix:* `SCHEMA_BUMP` 30 → 31 with the customary comment line, re-checked against `origin/main` at merge time.

**3. P1 — a `@Bean` factory method is a viable provider for the bean it itself declares.** `[reproduced]` — **both engines**: Claude `ce-adversarial` F2 (conf 100, executed against the real `di` phase), Codex `ce` #4 (conf 100, validator-confirmed). Coordinator-confirmed.

`di.ts:239` adds `structural.delete(edgeSourceId)`, which for a factory parameter is the **Method** id — but the provider is `CodeElement:spring-bean:${methodId}`, a different string, so nothing removes it. Unlike finding 1, this edge is a `Method -> CodeElement` pair, which is legal, so **it does persist and is visible to `impact`/`context`**.

Executed, on the composite-bean shape (`@Bean Filter loggingFilter()` + `@Bean Filter compositeFilter(List<Filter> filters)` — the `List<Filter>` / `List<HealthIndicator>` / `List<Converter>` idiom):
```
Method:Cfg.compositeFilter#1 -> CodeElement:spring-bean:Method:Cfg.loggingFilter#0   conf=0.8
Method:Cfg.compositeFilter#1 -> CodeElement:spring-bean:Method:Cfg.compositeFilter#1 conf=0.8   <<< SELF
```
and on the decorator shape (`@Bean Gateway inner()` + `@Bean Gateway outer(Gateway delegate)`) it does second damage: the phantom self-candidate turns a single-viable selection into an ambiguous one, so the **correct** `outer → inner` edge collapses from 0.9 to 0.5 and is tagged `ambiguous candidates: inner, outer`. A confidence-thresholded consumer now discards a correct edge because of a fabricated sibling.

*Fix:* also exclude the declarations owned by the edge source — `structural.delete(SPRING_BEAN_DECLARATION_ID_PREFIX + edgeSourceId)` — or, to keep `di.ts` framework-neutral as its docblock requires, record the declaring node id on `DiProviderMatch` and let the phase exclude any provider declared by the edge source. The closest existing fixture (`@Bean ConcreteRepo aggregate(List<Gateway>)`) misses this by returning a different type than it consumes.

---

### Intended inline comments

| # | Path : line (in diff) | Sev | Problem | Evidence | Fix | Tag |
|---|---|---|---|---|---|---|
| 1 | `gitnexus/src/core/ingestion/pipeline-phases/di.ts:235` | **P0** | Folding bean-node ids into `structural` makes the resolved provider a `CodeElement` while the edge source stays the consumer `Class` — a pair `CodeRelation` does not declare, so the edge is dropped at COPY and again by the row-by-row `catch { }`. Pre-existing `Class->Class` edges for `@Bean`-provided POJOs are replaced and lost. | `schema.ts` declares `FROM File/Function/Method TO CodeElement`, no `FROM Class`; `rel-pair-routing.ts:100-104` validates labels not pairs; executed binder error + `loadGraphToLbug` probe showing INJECTS absent. **Real defect is in the UNCHANGED `gitnexus/src/core/lbug/schema.ts`** — anchored here because that is the in-diff line that creates the shape. | Add `FROM Class TO CodeElement` to `RELATION_SCHEMA`; bump `INCREMENTAL_SCHEMA_VERSION` (`repo-manager.ts:546`) so existing indexes rebuild; make `RelPairRouter` reject undeclared pairs loudly. | `[reproduced]` |
| 2 | `gitnexus/src/core/ingestion/frameworks/spring/di-metadata.ts:66` | **P1** | New `beanFactoryMethods?` field (plus `'Resource'` at `:98`) changes the `ParsedFile.captureSideChannel` payload without a parse-cache invalidation, so warm caches replay pre-PR facts and the feature is inert. | `parse-cache.ts:56-57` states the MUST; `:122` still `SCHEMA_BUMP = 30`, file and `package.json` both absent from the diff; `:20-22` confirms the cache survives `--force`. **Real fix location is the UNCHANGED `gitnexus/src/storage/parse-cache.ts:122`.** Precedent: #2494 19→20, #2632 20→21, #2678 27→28. | Bump `SCHEMA_BUMP` 30 → 31 with a `// v31: …@Bean factory methods and @Resource injection sites` note, re-checked against `origin/main` at merge. | `[code-read]` |
| 3 | `gitnexus/src/core/ingestion/pipeline-phases/di.ts:239` | **P1** | `structural.delete(edgeSourceId)` removes the Method id but not `CodeElement:spring-bean:<methodId>`, so a `@Bean` method injects its own produced bean — a persisted false edge, plus a confidence collapse on the correct sibling edge. | Executed against the real `di` phase: self-edge at 0.8 on the composite pattern; 0.9→0.5 downgrade + `ambiguous candidates: inner, outer` on the decorator pattern. | `structural.delete(SPRING_BEAN_DECLARATION_ID_PREFIX + edgeSourceId)`, or carry the declaring node id on `DiProviderMatch`. Add a fixture whose `@Bean` parameter type equals its own provided type. | `[reproduced]` |
| 4 | `gitnexus/src/core/ingestion/pipeline-phases/di.ts:141` | P2 | Factory providers are indexed **only** under the literal `providedTypeName` (plus its simple tail), with no assignability or qualified-name resolution. | All three Codex legs; coordinator-calibrated. For unqualified `@Autowired Gateway` with `@Bean DefaultGateway defaultGateway()` the recall-first fallback still yields a useful edge to the implementing `Class` at 0.9 — so **not** a missing edge. The hard miss is `@Resource(name="defaultGateway") Gateway gateway`: the bean carries the name but is not in `structural` (type-key mismatch), `named` is empty, the name is explicit so `fallbackToType` is unset, and `di.ts:255` hits `continue` → **no edge at all**. Same for the common Kotlin `@Bean fun gateway() = DefaultGateway()`, where the concrete type is inferred from the body. | Resolve `providedTypeName` through the class/interface heritage index and register the factory provider under every assignable target type; for named selection, match the name and then test assignability rather than string equality. | `[code-read]` |
| 5 | `gitnexus/src/core/ingestion/frameworks/spring/resource-injection.ts:109` | P2 | `fallbackToType: true` is set for every implicit-name `@Resource`, including generic members whose type argument has already been erased — so the type fallback fans out to unrelated providers. | `normalizeSpringBeanType` erases at `annotation-arguments.ts:183-187`; executed probes: `List<Gateway>` → `List`, `List<Handler>` → `List`, `Map<String,Handler>` → `Map`, and `springResourceInjectionMatch('@Resource','items','List<Handler>','items')` → `{targetTypeName:'List', fallbackToType:true}`. The same erasure runs provider-side (`di-metadata.ts:274-277`), and `addProvidedType` has no AMBIGUOUS tombstone, so `@Bean List<Gateway>` and `@Bean List<Handler>` both key on `List` and both get 0.5 edges. Codex independently flagged the same fallback branch (`di.ts:247`) as too permissive. | Keep the erased key for the name-first match, but do not set `fallbackToType` when the raw declared type contained a `<` (or have `normalizeSpringBeanType` report that it erased and gate on that). | `[reproduced]` (parser probes executed; downstream fan-out traced) |
| 6 | `gitnexus/src/core/ingestion/frameworks/spring/di-metadata.ts:386` | P2 | The `@Autowired` + `@Resource` conflict `continue` fires **before** `semanticallyOwnedMemberNames.add(...)`, so the member is never marked `SPRING_DI_CAPTURED_FIELD_PROPERTY` and the legacy collect-all-implementers Property matcher still fans out for collection-typed fields — while scalar fields get nothing at all. Also note this deletes an `@Autowired` edge that existed pre-PR, for a case Spring resolves deterministically (`@Resource` wins). | Verified in the diff: the `continue` is at `:386`, both `.add()` calls are at `:390` and `:422`; the suppressor is read at `di-extractors/spring.ts:296`. The PR's own `ConflictingResourceConsumer` fixture uses non-generic `Gateway`, for which `parseSpringCollectionType` returns null — so the test passes by luck. Both engines (Claude `ce-correctness` CE-5 + `ce-adversarial` F5; Codex `ce` #5, validator-confirmed). | Move the guarded `semanticallyOwnedMemberNames.add(site.memberName)` above the conflict `continue` so a conflicted member is claimed-and-suppressed; or resolve to `@Resource` semantics as Spring does. Extend the fixture to `@Autowired @Resource List<Gateway>`. | `[code-read]` |
| 7 | `gitnexus/src/core/ingestion/di-extractors/index.ts:39` | P3 | The new `edgeSource: 'site'` (and the `CodeElement` provider target) makes the **public** `INJECTS` contract stale in three unchanged places. | `gitnexus-shared/src/graph/types.ts:136-137` still reads *"Source = the consumer Class node … Target = a concrete provider Class node"*; `gitnexus/src/mcp/tools.ts:237` ships `MATCH (c:Class …)-[r:CodeRelation]->(impl:Class) WHERE r.type = 'INJECTS'` as the documented DI example, which now silently returns fewer rows; `tools.ts:537` describes DI fan-out as `(consumer→implementer)`; `ARCHITECTURE.md:106` still describes the phase as Class→Class. **All four are UNCHANGED files** — anchored on the in-diff `edgeSource` declaration that introduces the new shape. Credit where due: the PR *did* correctly update the in-file docblock at `di.ts:6`. | Update the `types.ts` doc comment, both `tools.ts` strings, and the ARCHITECTURE.md phase row in this PR. | `[code-read]` |

---

### Lower-priority items, refuted suspicions, and test gaps

**Lower priority (body-level, not inline):**
- `decodeSpringBeanFactoryReason` validates `namesKnown` and then **drops it** from `SpringBeanFactoryMetadata` (`bean-factories.ts:85-99`), so an MCP consumer sees `names: []` with no way to tell "this bean has no names" from "the names are dynamic and unresolved". Codex `ce` #6 (conf 100); confirmed by read. Cheap fix: carry the flag through.
- `@Resource` on a **non-setter single-argument method** is dropped before the explicit name is ever parsed: `springResourceDefaultName` returns `null` for any method not matching `/^set[A-Z_$]/` (`resource-injection.ts:50-54`), and `di-metadata.ts` bails on `defaultName === null` before calling `springResourceInjectionMatch`. So `@Resource(name="gateway") void configure(Gateway g)` yields nothing, though Spring's `CommonAnnotationBeanPostProcessor` accepts any non-static one-arg `@Resource` method. Codex-adversarial only; confirmed by read. Fails closed — a recall gap, not a false edge.
- `@Bean` method-level `@Qualifier`/`@Primary` are not folded into provider names/preference, and `@Bean(autowireCandidate = false)` / `defaultCandidate = false` still register as full DI providers (Claude executed: `{"names":["m"],"namesKnown":true}`). Codex-adversarial `[high]` + Claude `ce-adversarial` residual R-4 — cross-engine, but an acknowledged out-of-scope item on #1225 (`@Qualifier`/`@Primary` for factory parameters are subtasks #2415-#2418). Worth a note in the PR body rather than a blocker.
- The syntactic `@Bean` gate is simple-name-only (`bean-factories.ts:41`) and the `continue` at `java/spring-di.ts:156` runs at capture time, before FQN resolution — so a **non-Spring** `@Bean` co-located with a real `@Autowired`/`@Inject` deletes an edge that existed pre-PR. The swarm risk lane originally graded this P2 with an `@Qualifier` trigger; that half is refuted (`@Qualifier` is not in `INJECTION_ANNOTATIONS`, so those methods produced no edges pre-PR either), which narrows it to a rare combination → P3. The more interesting half is coverage: a *genuine* Spring `@Bean` whose annotation does not resolve (ambiguous wildcard import) produces no bean node at all, so the feature is silently off for that codebase shape. Kotlin has a partial escape hatch (`useSiteTarget === undefined` pre-filter) that Java lacks.
- Kotlin `kotlinBeanFactoryReturnType` (`kotlin/spring-di.ts:94-109`) infers the provided type from the **first `call_expression` descendant** via an unrestricted DFS, so `@Bean fun cache() = if (enabled) RedisCache() else NoOpCache()` is indexed as providing `RedisCache` (Kotlin's actual type is the LUB), and a block-bodied `@Bean fun setup() { Validate(config) }` is indexed as providing `Validate` even though its type is `Unit` — bypassing the `Unit`/`Void` rejection the declared path enforces. Both Claude lanes, conf 50-55, `[code-read]` (tree-sitter-kotlin is absent from this worktree). Restrict inference to a single top-level `call_expression` expression body.
- Maintainability nits: `splitTopLevel` / `topLevelAssignment` duplicate the same ~25-line quote/bracket state machine (the file's own comment at `:58` says "keep the same deliberate policy", i.e. two copies kept in sync by hand); `simpleName` is byte-identical in `bean-factories.ts:33` and `di-metadata.ts:112` (moving it to `annotation-arguments.ts` avoids the cycle); `providerNodes` now conflates "recognized providers" with "every Class node", which invites `providerNodes.has(id)` being misread as a provider gate; and the `spring-bean-factory:{…JSON…}` `reason` payload is the first raw `JSON.stringify` in a `reason` field, where the repo already has three versioned delimiter codecs (`reaching-def-reason-codec.ts`, `taint/path-codec.ts`, `call-summary-codec.ts`) — it will render as a JSON blob in generic `impact`/`context` output.
- `mcp/local/bean-metadata.ts` matches `Method -> CodeElement` `DECLARES` with `LIMIT 1` and no `r.reason STARTS WITH 'spring-bean-factory:'` filter. Correct today (the only other `DECLARES` emitter is `File -> Class`), but it becomes a wrong-row bug the moment a second `Method -> CodeElement DECLARES` producer lands. Add the predicate now — the exact pattern is already used in this PR at `run-analyze.ts:1905-1907`.
- `parseStaticStringLiteral` (`annotation-arguments.ts:124`) rejects any bean name containing `$` followed by a letter or `{` — so the legal Java name `@Bean("legacy$service")` yields `{names: [], namesKnown: false}` instead of the name. Fails closed (a recall gap, not a false edge), P3, Claude `ce-correctness` CE-6. The guard is there to reject Kotlin string templates; it should be applied only to Kotlin-sourced annotation text, or narrowed to `${`.
- Process: the body says `Closes #2413` but omits **#2633** (the open `@Resource` tracking issue this PR substantially implements) — it will not auto-close. The PR template's *Risk & rollout* and *Explicitly out of scope* sections are missing, and the change does force a one-time full reindex for Java/Kotlin repos via the feature-version bump.
- Two bounds-only assertions on deterministic values against DoD §2.7: `spring-bean-resource-benchmark.test.ts:117,127` (`captureCount` is an exact emit count over a fixed synthetic source) and `spring-bean-resource-pipeline.test.ts:309` (`fallback.length > 1`, where the same file already pins the analogous fan-out exactly at `:290-295`).

**Test gaps (convergent across engines):**
1. **No test writes a `@Bean`-provided INJECTS edge through the real CSV→COPY path.** This is the single gap that let finding 1 through. `test/helpers/test-indexed-db.ts`'s `beforeFTS` hook exists precisely to *"exercise the real CSV→COPY path (`loadGraphToLbug`) instead of Cypher CREATE"*. Adding one consumer class to `setupSpringBeanFactoryIncrementalRepo` and asserting a non-zero persisted INJECTS count would have caught it.
2. No cross-version parse-cache test — nothing asserts the new facts survive (or correctly invalidate) a warm `parsedfile-cache/` written by a prior build. `spring-bean-schema.test.ts:24`'s `expect(parseSchemaVersion).toBeGreaterThanOrEqual(22)` is a stale lower bound satisfied by unrelated prior bumps; it gives false confidence that this PR version-gated its capture change.
3. No static guard asserting the pipeline's emittable `(fromLabel, toLabel)` set is a subset of `RELATION_SCHEMA` — cheap, and the durable fix for the finding-1 class.
4. No fixture where a `@Bean` method's parameter type equals its own provided type (finding 3); no `@Autowired @Resource List<Gateway>` collection fixture (finding 6); no `@Bean(name = SOME_CONSTANT)` through the real pipeline (the `confidence: 0.8` DECLARES branch is unit-tested only); the `0.85` single-candidate type-fallback branch in `di.ts:281` is unreachable by any fixture in the diff.

**Also refuted, worth not re-raising:** adversarial round-trips of hostile bean names through the DB (`JSON.stringify` escaping + RFC-4180 CSV `ESCAPE='"'` hold); `@Bean(initMethod=…)`, `@Bean(name = {})`, `@Bean("")` all correctly fall back to the method name; removing the `break` in the injection-annotation loop is inert (only the reason string's wording changes); `@Resource` on a constructor parameter never reaches the suppression path in either language.

---

### CI status

Merge state: **`checks failing`** (not a conflict — `mergeable = MERGEABLE`).

Everything is green except **`tests / ubuntu / coverage 2/3`**, which fails on `test/unit/embedding-runtime-support.test.ts > isLocalEmbeddingStackInstalled` (`expected false to be true`) — a file **untouched by this diff** and matching the repo's documented full-suite load-flake class; the downstream `CI Gate` fails as a consequence. Vercel additionally fails on deploy authorization, not content. `mergeStateStatus = BLOCKED` — the block is the required-review gate plus that red job.

The PR body's own validation claim (full suite: 15,238 passed / 99 skipped / **10 failed**, asserted to be 9 pre-existing platform-baseline limits plus 1 registry-race FTS flake that passed 18/18 in isolation) was **not** independently reproduced by this review.

### Coverage limits

The worktree has no `dist/` and cannot `npm run build` reliably (symlinked deps), so nothing here ran the full `analyze` CLI end to end. Findings 1 and 3 were reproduced against a real LadybugDB and the real `di` phase via `node --import tsx`; finding 5's parser behavior was reproduced by direct probes; everything else is `[code-read]`. Specifically **not** verified: the warm-parse-cache no-op (finding 2) was traced through `parse-worker` → `capture-side-channel` → `parse-cache` → `parse-impl` but never executed across a version boundary; `tree-sitter-kotlin` is absent from this worktree, so all Kotlin AST-shape claims are code-reads; `test/integration/spring-di-pipeline.test.ts` (the natural regression test for pre-PR `@Autowired`/`@Qualifier` behavior under the `di.ts` rewrite) could not be run here — CI does run it via `pretest:integration`, and it should be confirmed green. The label-enumerating queries in `local-backend.ts` beyond `enrichCandidateLabels` and the Class/Interface `typeCheck` UNION were not exhaustively audited.

---

### Final verdict

**`not production-ready`.**

The P0 schema-pair absence silently drops every consumer-class `INJECTS` edge that targets a `@Bean`-declared provider — including a *regression* of previously-persisting `Class -> Class` edges — and no existing test catches it, because no test writes these edges through the real CSV→COPY path. The P1 parse-cache miss leaves the whole feature inert on any warm cache, regardless of `--force`. Both need code changes plus schema-version bumps (`RELATION_SCHEMA` + `INCREMENTAL_SCHEMA_VERSION`, and `SCHEMA_BUMP`) before this can land. The P1 self-injection edge persists a fabricated relationship and downgrades a correct sibling edge, so it should land in the same pass.

The underlying design is good and most of the risk surface was probed and cleared — this is a "fix three concrete defects", not a "rethink the approach".

---

*This is an automated multi-tool digest (Claude swarm + ce personas, Codex GPT-5.5/xhigh swarm + ce + adversarial). Please verify each finding against the code before acting on it.*

gitnexus/src/core/lbug/schema.ts
```diff
@@ -461,6 +462,21 @@ CREATE REL TABLE ${REL_TABLE_NAME} (
...
+ * node label.
+ */
+export const RELATION_SCHEMA_PAIRS: ReadonlySet<string> = new Set(
+  [
+    ...RELATION_SCHEMA.matchAll(
+      /\bFROM\s+`?([A-Za-z][A-Za-z0-9_]*)`?\s+TO\s+`?([A-Za-z][A-Za-z0-9_]*)`?/g,
+    ),
+  ].map((match) => `${match[1]}|${match[2]}`),
+);
+
```
magyargergo: I don't think this should be here.

magyargergo (changes requested):

gitnexus/src/core/lbug/schema.ts
ShiningXu: Do you mean removing RELATION_SCHEMA_PAIRS and the router-level pair validation from this PR, while keeping the required Class → CodeElement schema pair and persistence round-trip coverage?

magyargergo:
Can you please make sure CI is healthy? 🙏

magyargergo:
Tracking note: this PR also implemented #2633 (JSR-250 `@Resource` injection). #2740 closed #2413 on merge; #2633 is now closed against this change.

Issue #2649, opened 2026-07-23
gitnexus analyze crash for a big project


standin000:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

gitnexus analyze works for small project folders, but it crash for a 190~ projects with thousands files folder.

### Expected behavior

gitnexus analyze works.

### Actual behavior

gitnexus analyze crash

### Steps to reproduce

1. cd project directory
2. run gitnexus analyze 

### Environment

gitnexus version 1.6.9
node version 24.15.0
windows 11 25H2

### Logs / screenshots

Parsing chunk 1/1... (Parsing...) (1m 1s){"level":40,"time":1784697097381,"name":"gitnexus","workerIndex":8,"error":"Replacement worker did not report ready within 5000ms — likely crashed d}

magyargergo:
@standin000 could you please include more information? 🙏 What language(s) the project is written? How you run GitNexus?

micahjatau:
I have been experiencing the same thing with opencode 1.18.4, running on Ubuntu 24.04. Anytime I run gitnexus analyse it crashes opencode harness

magyargergo:
@micahjatau Could you please provide more information? You could enable verbose logging to get more out of it?

magyargergo:
We reproduced this at Linux-kernel scale and traced the root cause.

**Repro setup:** `torvalds/linux` shallow clone — 94,773 files, ~64k parseable C/C++ — `gitnexus analyze` built from current `main` (0eeecb37), Node 22.18, Linux x64, 31 GB RAM.

**What happens**

| Run | Config | Result |
|---|---|---|
| 1 | defaults (16 workers, ~4 GB default V8 heap) | `FATAL ERROR: Reached heap limit` (exit 134) after ~4 min, mid-parse |
| 2 | `--max-old-space-size=12288`, 4 workers | `Ineffective mark-compacts near heap limit` (exit 134) after ~28 min |
| 3 | same + `GITNEXUS_DEBUG_HEAP=1` | linear heap growth; killed by an external memory watchdog at chunk 94/113 with 19.6 GB RSS |

Heap probes from run 3:

```
[gitnexus-heap] parse-chunk-25 used_mb=2481 rss_mb=7133  nodes=1143241 parsedFiles=0
[gitnexus-heap] parse-chunk-50 used_mb=4622 rss_mb=11102 nodes=2188306 parsedFiles=0
[...]

Issue #2514, opened 2026-07-16
bug(rust): third duplicate type name restores ambiguous range binding


100yenadmin:
## Summary

The Rust range-binding prepass treats duplicate function/struct names as ambiguous by deleting the name from its global map. A third occurrence then sees the name as absent and inserts its type again, so odd duplicate counts can incorrectly resolve to the last scanned file.

## Current behavior

On current `main` (`c4fb511a`), `range-binding.ts` uses the same toggle pattern for both registries:

- `allReturnTypes`: first definition sets, second deletes, third sets again
- `allFieldTypes`: first definition sets, second deletes, third sets again

Traversal is deterministic now, but the result is still semantically wrong for three or more same-named definitions. The third file can supply a cross-file return or field type that should remain ambiguous, which may mis-resolve receiver calls, loop bindings, assignments, or struct destructuring.

## Expected behavior

Once a name has multiple definitions, it remains ambiguous for the rest of the prepass regardless of duplicate count or file order. Unique names continue to resolve normally.

## Suggested direction

Track ambiguous names separately (or retain a count/state) instead of using map presence as both value storage and ambiguity state. Subsequent occurrences of an ambiguous name should never reinsert it into the resolvable map.

## Acceptance coverage

- two same-named functions with different return types suppress cross-file inference
- three same-named functions also suppress it
- two and three same-named structs with conflicting fields suppress global field inference
- a unique function/struct still infers normally
- permuting input file order does not change the result

## Provenance

Follow-up to the out-of-scope maintainer finding in #2482 review: https://github.com/abhigyanpatwari/GitNexus/pull/2482#pullrequestreview-4711210358

magyargergo:
Thanks for this follow-up! By the way, would you want to join our community Discord server?

PR #2741, opened 2026-07-29
fix(rust): resolve module-qualified calls against the module tree (#2730)

### LINKED ISSUE #2730: Rust: call edge through a same-name wrapper fn is missing — central dispatcher reported risk=LOW with 0 affected processes
## Summary

On a Rust codebase, `gitnexus impact` misses a fully-qualified call edge when the caller is a same-name wrapper function. The target is the central tool dispatcher of the repo (every tool call flows through it), yet impact reports `risk: LOW, direct: 2, processes_affected: 0`.

Reproduced on **1.6.3 and 1.6.9**.

## Reproduction

```bash
git clone https://github.com/hec-ovi/noob-cli /tmp/noob-cli
cd /tmp/noob-cli
npx -y gitnexus@1.6.9 analyze --skip-agents-md --name noob-cli
npx -y gitnexus@1.6.9 impact --repo noob-cli \
  "Function:crates/noob/src/tools/mod.rs:dispatch"
```

Output: `risk: LOW | direct: 2 | processes_affected: 0 | impactedCount: 2 | epistemic: "exact"`

The two `direct` callers it does find are the `#[cfg(test)]` test functions inside `tools/mod.rs` itself; the production caller below is absent, while the output labels itself `epistemic: "exact"`.

## Expected

`crates/noob/src/agent/sched.rs` defines its own local `fn dispatch(...)` (line 361) whose body calls the target with a fully qualified path:

```rust
// crates/noob/src/agent/sched.rs:386
tools::dispatch(ctx, name, args)
```

This caller (and the execution flows through it — the whole agent loop runs `run_input → sched → tools::dispatch`) is absent from the impact result. `processes_affected: 0` for the function every tool call passes through is the dangerous kind of wrong: it reads like "safe to change".

Grep ground truth for `\bdispatch\s*\(` (excluding definition lines) finds caller files `agent/sched.rs` and `tools/mod.rs` (plus an unrelated same-name method in `noob-provider/src/sse.rs`); the graph only reports `tools/mod.rs`.

## Related issues (checked, distinct)

- #2604 (closed): no CALLS edge for `&dyn Trait` call sites — this repro is a **static, fully qualified** call (`tools::dispatch(...)`), not a trait object.
- #2508 (closed): edge exists but a read-side filter drops it — unclear if related; here the caller file never appears in any output.

## Guess at the cause

The caller and callee share the name `dispatch`. It looks like edge resolution collapses or drops the edge when caller-name == callee-name (self-loop suppression?) or when the call is qualified (`tools::dispatch`) rather than bare. Symbol *disambiguation* itself works fine — `impact dispatch` correctly returns `status: ambiguous` with all 3 candidates.

Found via [impact-audited](https://github.com/klmtseng/impact-audited), a small grep-floor cross-checker for graph impact tools.

magyargergo:
Fixes #2730.

## The bug

A Rust call written with a path was captured with only its tail identifier, so `tools::dispatch(..)` looked identical to a bare `dispatch(..)`. The scope-chain walk then resolved the bare name lexically and bound it to whatever `dispatch` was nearest, which for the common wrapper idiom is the wrapper itself:

```rust
// crates/noob/src/agent/sched.rs
use crate::tools::{self, ToolCtx, ToolOutcome};

fn dispatch(ctx: &ToolCtx, name: &str, args: &Value) -> ToolOutcome {
    tools::dispatch(ctx, name, args)   // resolved to THIS function
}
```

The graph gained a self-loop and the real cross-module edge never existed. On the reporter's repository `impact` on the central tool dispatcher returned `risk: LOW` with `processes_affected: 0`, and both "callers" it did find were `#[cfg(test)]` functions in the same file, while the result still labelled itself `epistemic: "exact"`. That reads as "safe to change" for the function every tool call flows through.

The issue guessed the edge was being suppressed. It was not. Nothing was dropped: the resolver manufactured a self-loop and consumed the real edge.

## Approach

An earlier draft mapped a module name to a file by filename convention. That covered the reported shape and nothing else, so it was replaced. Paths are now resolved the way rustc resolves them, over the module tree rather than the filesystem.

- `mod_item` emits `@declaration.namespace`, making a Rust module a named definition instead of an anonymous scope region. This mirrors the existing C++ `namespace_definition` capture, and it lets the shared `tagNamespacePrefixes` pass stamp members with their enclosing module path. That pass needed no changes to start working for Rust.
- `module-path.ts` reconstructs the other half of the tree. Crate roots are directories holding `main.rs` or `lib.rs`, a file's module path is its location below that root, and a definition's module is its file's module plus any enclosing `mod` blocks.
- `crate::`, `self::` and `super::` are prefix transforms on the calling module, not reasons to stop resolving. The draft treated them as bail-outs, which is why `super::` paths misbound.
- The final segment is looked up as a member of the resolved module, including members it only re-exports. A `pub use` creates no binding on the re-exporting module's own scope, so re-exports are followed through that module's import edges.

Resolution runs ahead of the implicit-`this` and scope-chain tiers so an explicit path outranks a lexical shadow, and returns undefined on an unknown module, a missing member, or a tie. A wrong CALLS edge is worse than a missing one, which is exactly what made this issue dangerous.

The new `ScopeResolver.resolveQualifiedFreeCall` hook is optional and unset for every other language, so shared behaviour is unchanged.

## Result

On the reporter's repository, `impact` on `crates/noob/src/tools/mod.rs:dispatch`:

| | before | after |
|---|---|---|
| direct callers | 2 | 3 |
| impacted | 2 | 6 |
| modules | Tools | Agent, Tools |

The production caller `agent/sched.rs:dispatch` now appears at depth 1.

A dedicated fixture pins the cases the draft got wrong, each of which previously produced a wrong edge:

| case | before | after |
|---|---|---|
| `a::b::dispatch()` | `main.rs:dispatch` | `src/a/b.rs:dispatch` |
| `super::dispatch()` | `a/b.rs:dispatch` | `src/a/mod.rs:dispatch` |
| `facade::dispatch()` via `pub use` | `main.rs:dispatch` | `src/tools.rs:dispatch` |

Unqualified calls stay on the lexical chain and keep their `local-call` reason.

## Known limitation

An inline `mod inner { fn dispatch }` and a crate-root `fn dispatch` in the same file still collapse to one graph node. Node identity is `<file>:<qualifiedName>` and does not carry the module path, and `namespacePrefix` is a deliberate sidecar that does not participate in node identity. Resolution picks the right definition, but the target is not representable, so the edge still renders as a self-loop.

This is pre-existing and unchanged by this PR. Fixing it means module-path-qualified node ids for Rust, which changes ids for essentially every Rust repository since `#[cfg(test)] mod tests` is ubiquitous, and therefore an incremental-schema migration. Tracked separately in #2742.

## Testing

- 3 new fixtures and 10 new tests covering the reported wrapper case, a cargo workspace layout, multi-segment paths, `super::`, `pub use` facades, and the unqualified control cases.
- Full resolver integration suite passes across all languages.
- `rust-captures-golden` regenerated: 12 new fixture entries plus pre-existing entries whose `mod` and scoped-call capture groups grew.

## Notes for review

- `import-resolvers/rust.ts` widens two helper signatures from `Set` to `ReadonlySet`. No behaviour change.
- No CHANGELOG entry, per repository convention for fix PRs.


gitnexus/src/core/ingestion/languages/rust/module-path.ts
```diff
@@ -0,0 +1,139 @@
...
+ *   src/main.rs                     → []            (the crate root module)
+ *   src/tools.rs                    → ['tools']
+ *   src/a/mod.rs                    → ['a']
+ *   src/a/b.rs                      → ['a', 'b']
+ *   crates/noob/src/tools/mod.rs    → ['tools']
+ *
+ * Returns `undefined` for a file under no known crate root — the caller then has
+ * no module identity to reason about and must refuse rather than guess.
+ */
+export function moduleOfFile(filePath: string, index: RustModuleIndex): string[] | undefined {
```
magyargergo: **P1 — module identity drops the crate, so this can silently switch the fix off in a cargo workspace.** `[reproduced]`

`moduleOfFile` returns only the segments below a crate root, and `sameModule` (line 137) compares only those segments. Re-run directly:

```
crateRoots: ["crates/a/src", "crates/b/src"]
crates/a/src/tools.rs -> ["tools"]
crates/b/src/tools.rs -> ["tools"]   # identical
```

**Trigger:** a workspace where two members each have `src/tools.rs`. If only one defines the member you get a wrong **cross-crate** edge; if both do, `findMemberInModule` sees `count === 2`, refuses, and the site falls back to the lexical chain — **restoring the exact #2730 self-loop**.

Same-named modules across members (`util`, `error`, `config`, `types`) are near-universal, and **#2730's own repro repo is multi-crate** (`crates/noob`, `crates/noob-provider`), while `rust-2730-crate-layout` has a single member and so cannot catch either manifestation.

**Fix:** make module identity `{ crateRoot, segments }` and thread it through member and re-export lookup.

_Found by every Codex leg and two Claude lanes; reproduced by two independent drivers and re-run by the synthesis pass._

gitnexus/src/core/ingestion/languages/rust/qualified-call.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  // 1. A module declared in, or below, the calling module (`mod inner { … }`,
+  //    `mod tools;`) — the in-scope type-namespace binding.
+  yield [...callerModule, ...qualifier];
+
+  // 2. A `use` binding for the first segment. Finalize already resolved the
+  //    import to a file, so the module path comes back through the same
+  //    file → module mapping as everything else. Covers `use crate::tools::{self}`
+  //    and `use crate::a::b as tools`.
+  const head = qualifier[0];
```
magyargergo: **P1 — the `use`-binding channel accepts an imported TYPE as a module.** `[reproduced]`

Nothing checks that `head` names a module. `tryRustModulePath` *deliberately* strips a trailing symbol segment (`import-resolvers/rust.ts:77` — "last segment might be a symbol… Strip it and try again"), so:

```rust
use crate::client::ClientBuilder;   // edge -> targetFile = src/client/mod.rs
ClientBuilder::new()                // candidate module becomes ["client"]
```

Rust impl methods carry a **bare** `qualifiedName` (`function_item` captures only `@declaration.name`), so the tail is then looked up as an ordinary module member.

Many such calls land correctly by accident — when the impl method is the only callable of that name in the module. The failure is conditional but structural: a **false edge** when a module-level callable of the same name also exists (`new`, `default`, `from`, `build`, `load`, `parse` in a facade `mod.rs` — exactly the shape `pub use` motivates), or a misfire into the tie-refuse path when several do.

**Fix:** require the `use`-channel head to resolve to an actual module before accepting it.

_Reproduced independently by two Claude lanes with separate drivers._

gitnexus/src/core/ingestion/languages/rust/qualified-call.ts
```diff
@@ -0,0 +1,252 @@
...
+/** A callable named `name` declared directly in `targetModule`. Refuses on a tie. */
+function findMemberInModule(
+  targetModule: readonly string[],
+  name: string,
+  scopes: ScopeResolutionIndexes,
+  index: RustModuleIndex,
+): SymbolDefinition | undefined {
+  let unique: SymbolDefinition | undefined;
+  let count = 0;
+  for (const defId of scopes.qualifiedNames.get(name)) {
```
magyargergo, line 308: **P1 — membership is inferred from file path alone, so one nested helper disables qualified resolution for the whole module.** `[reproduced]`

This loop counts every def whose `moduleOfDef` matches, with no check that the def is owned by a **module-level** scope. A `fn` nested inside another `fn`, or an `impl`/trait method (both carry bare `qualifiedName`s), counts as a second module member.

**Trigger:** the target module contains a function-local `fn dispatch` — or an `impl X { fn dispatch }` — alongside the real `pub fn dispatch`. `count === 2` → refuse → the site is handed back to the tier that emits the wrong self-loop, **reinstating #2730**, with no resolution outcome recorded.

The refuse-on-tie *logic* is correct (I checked: the multi-match value is genuinely discarded and the index dedups `(name, DefId)`); the problem is what is being counted.

**Fix:** restrict candidates to defs owned by a module-level scope.

_Found by **both engines** — Claude adversarial reproduced it on the #2730 shape verbatim; Codex adversarial and Codex CE reached it independently by source read._

gitnexus/src/core/ingestion/languages/rust/qualified-call.ts
```diff
@@ -0,0 +1,252 @@
...
+ * survives as an `ImportEdge` on that module scope, which is what this reads.
+ */
+function findReexportedMember(
+  targetModule: readonly string[],
+  name: string,
+  scopes: ScopeResolutionIndexes,
+  workspaceIndex: WorkspaceResolutionIndex,
+  index: RustModuleIndex,
+): SymbolDefinition | undefined {
+  for (const [filePath, moduleScope] of workspaceIndex.moduleScopeByFile) {
```
magyargergo: **P1 — full workspace scan per candidate, on a path taken by every `Path::name()` call.** `[reproduced]`

This walks all of `workspaceIndex.moduleScopeByFile` with no module→file index, once per candidate. Combined with the capture at `query.ts:172` matching **any** `scoped_identifier` callee, the hook runs for `Vec::new()`, `String::from()`, `HashMap::new()` — calls that can never resolve into the workspace but pay `O(defs-named-X + files)` each before returning `undefined`.

Two independent benchmarks agree per-site cost is linear in file count:
- 0.117 → 0.228 → 0.428 ms/site at 301 / 601 / 1201 files
- 0.26 → 9.46 ms/call at N=500 → 16,000

Total pass cost is O(qualified-call-sites x file-count). Given the #2649 / #1871 history this is worth fixing before merge.

**Fix:** build a `Map<modulePathKey, Scope>` once per pass (or add `moduleScopeByModulePath` to `WorkspaceResolutionIndex`), and skip the hook when the qualifier head cannot name a workspace module.

_Note: under `GITNEXUS_DISK_SCOPE_INDEX=1` (default-off) this walk runs after the scope-tree seal, turning the CPU scan into per-file `readFileSync` + `JSON.parse` — the exact pattern `exportedCallableByName` was added to avoid._

gitnexus/src/core/ingestion/languages/rust/qualified-call.ts
```diff
@@ -0,0 +1,252 @@
...
+  index: RustModuleIndex,
+): Generator<readonly string[]> {
+  if (anchored.anchored) {
+    yield anchored.path;
+    return;
+  }
+
+  // 1. A module declared in, or below, the calling module (`mod inner { … }`,
+  //    `mod tools;`) — the in-scope type-namespace binding.
+  yield [...callerModule, ...qualifier];
```
magyargergo: **P2 — candidate 1 is an unchecked guess yielded before the `use` channel.**

`[...callerModule, ...qualifier]` never verifies the caller actually declares that submodule, so file layout beats a real `use` binding.

**Trigger:** `src/a/mod.rs` has `use crate::b;` while an undeclared (or `cfg`-gated) `src/a/b.rs` exists — candidate 1 matches `['a','b']` first and wins. rustc would resolve to `crate::b`.

**Fix:** only yield candidate 1 when `qualifier[0]` is genuinely a declared submodule of the caller (an import edge with that `localName`, or an inline `Namespace` scope).

_Two Claude lanes reproduced it; Codex swarm reached the same root cause from the function-local `use` direction._

gitnexus/src/core/ingestion/languages/rust/method-owners.ts
```diff
@@ -21,6 +25,12 @@ import { isClassLike, populateClassOwnedMembers } from '../../scope-resolution/s
 export function populateRustOwners(parsed: ParsedFile): void {
   populateClassOwnedMembers(parsed);
   populateRustImplOwners(parsed);
+  // #2730: tag defs declared inside `mod` blocks with their enclosing module
+  // path (`inner`, `a.b`) on the `namespacePrefix` sidecar. Rust modules are
+  // Namespace scopes owning a Namespace def (see the `mod_item` capture), so
+  // the shared C++-era pass applies unchanged. Qualified call resolution reads
+  // this to tell `inner::dispatch` apart from a same-named crate-root `fn`.
+  tagNamespacePrefixes(parsed);
```
magyargergo: **P2 — the C++-era "already namespaced" guard mis-fires for Rust.** `[reproduced]`

`tagNamespacePrefixes` skips a def when its `qualifiedName` already equals or starts with the prefix (`walkers.ts:1054`, `:1078`). That guard was written for C++, where `qualifiedName` carries the namespace. In Rust it means `fn a` inside `mod a` — or `mod inner { fn inner }` — keeps `namespacePrefix` **unset**, so `moduleOfDef` reports it as a member of the *parent* module.

**Trigger:** `pub mod a { pub fn a() {} }` then `crate::a::a()` — refuses. Worse, that def is then indistinguishable from a crate-root `fn a` for the module matcher.

_Claude adversarial reproduced it; Codex CE reached it independently (confidence 100)._

gitnexus/bench/scope-capture/baselines.json
```diff
@@ -47,31 +47,32 @@
...
+    "_rebaselined_module_tree_2730": "#2730: RUST_SCOPE_QUERY now captures mod_item as @declaration.namespace (a Rust module is an item, mirroring the C++ namespace_definition capture) and tags scoped call sites with @reference.qualified-name so the written path survives to resolution. Both are additive captures: every bench fixture containing a mod block or a Foo::bar() call gains groups, and the corpus also grew by the three rust-2730-* fixtures. Prior 655aed01cf1b6b84fa0c64d48dfb2526ecb67f47d90f0a91edabacd269a212db -> cf2e47ae3765fc703337db1b313eb17265ba64a96475d92c860b6c3c7ff73422; scaling 1.043 < 1.5; fixture_count 174 -> 186. Only the rust fingerprint moves; the other 14 languages are byte-identical."
```
magyargergo: **P2 — the "Prior" fingerprint in this note is two rebaselines stale.** `[reproduced]`

The note cites `Prior 655aed01…`, but the actual pre-PR value on the base commit is `7f1240b3…`:

```
git show <base>:gitnexus/bench/scope-capture/baselines.json
  rust.fingerprint = 7f1240b38457468f06b7931e0c2c578f218f922774d0dc7e2ee6ef3b08d4d689
```

`655aed01…` was the value after #2514; #2604 and #2714 both moved it since. Everything else in the note is correct (the "after" hash, scaling 1.043, `fixture_count 174 -> 186`, and "other 14 languages byte-identical" all verified).

CI cannot catch this — the gate compares the live fingerprint to the stored value, never the prose chain. Since the whole point of these notes is an auditable lineage, a wrong link breaks the chain for the next person.

magyargergo (commented):
# PR #2741 — multi-engine review digest

**Target:** `fix(rust): resolve module-qualified calls against the module tree (#2730)`
**Base:** `79ff44df` → **Head:** `6f8d86c8` (2 commits, 26 files, +934/−143)

---

## 1. Review bar for this PR

Acceptance criteria derived from `DoD.md`, `AGENTS.md`, `GUARDRAILS.md` and `ARCHITECTURE.md`, scoped to what this change actually touches:

- **Correctness over coverage.** The module's own contract is the bar it must meet: *"A wrong CALLS edge is worse than a missing one"* (`qualified-call.ts:29-32`). Converting an unresolved site into a false edge fails this even though it raises the resolution count.
- **Shared ingestion code must not name languages** (`AGENTS.md:42`, `CLAUDE.md:38`, `DoD.md:45`) — language behaviour plugs in via `LanguageProvider` / `ScopeResolver` hooks.
- **Persistence-boundary discipline.** A change to what is emitted for source whose *content* has not changed must bump `INCREMENTAL_SCHEMA_VERSION` (`repo-manager.ts`) and, for parse-time capture changes, `SCHEMA_BUMP` (`parse-cache.ts`). Both constants document this rule in their own doc blocks.
- **Large-repo performance.** No new super-linear work on the per-call-site hot path; the repo has an explicit history of regressions of exactly this class (#1871, #2649).
- **Evidence-based edits** (`DoD.md` §3) — claims about repo state verified against current code, not carried from memory or stale docs.
- **Exact assertions in tests** (`DoD.md` §2.7); fingerprint-gated benchmarks rebaselined with a prose justification note.

## 2. Problem being solved

Issue #2730: a Rust module-qualified call such as `tools::dispatch()` was captured **tail-only**. The qualifier was inert, so the shared scope-chain walk resolved the bare name `dispatch` and bound it to whatever was lexically nearest. For the extremely common wrapper idiom — `fn dispatch(..) { tools::dispatch(..) }` — that is the wrapper itself. The result was a **self-loop** where a real cross-module edge should have been, and `impact` consequently reported the central dispatcher as risk LOW with 0 affected processes.

The PR resolves paths over the Rust module tree the way rustc does: leading path segments name modules in the type namespace (so a same-named `fn`, which lives in the value namespace, can never shadow them), and `crate::` / `self::` / `super::` are prefix transforms rather than bail-outs. `mod_item` gains `@declaration.namespace`; scoped calls carry `@reference.qualified-name`; new `rust/module-path.ts` + `rust/qualified-call.ts` implement the resolution; a new optional `ScopeResolver.resolveQualifiedFreeCall` hook is invoked from the shared `passes/free-call-fallback.ts`; Rust opts into the shared `tagNamespacePrefixes`. A second commit rebaselines `bench/scope-capture/baselines.json`.

## 3. Current PR state

- Open, non-draft. 26 files, +934/−143, two focused commits.
- `c5c7042e` — implementation, fixtures, and the regenerated `expected-captures.json`.
- `6f8d86c8` — `bench/scope-capture/baselines.json` rebaseline only.
- Worktree clean; `git diff --check` clean; no bidi/hidden-Unicode; non-ASCII limited to prose punctuation.
- GitNexus `detect_changes`: 9 changed production symbols, 17 affected flows, aggregate CRITICAL risk. `ScopeResolver` carries HIGH graph impact (27 direct / 63 total dependants) — expected for a contract file, and the addition is an optional hook.

### Engine breakdown

| Engine | Lanes that produced findings | Executed repros? |
|---|---|---|
| **A — Claude (swarm)** | `risk`, `test-ci`, `branch-hygiene` | risk: code-read |
| **A — Claude (CE personas)** | `correctness`, `adversarial`, `performance`, `testing`, `api-contract`, `maintainability`, `standards` | **Yes** — correctness, adversarial and performance each built independent out-of-tree drivers under `/tmp` replaying the production decision path (`extractParsedFile → populateRustOwners → finalizeScopeModel → buildWorkspaceResolutionIndex → resolveQualifiedFreeCall`) via `npx tsx`, plus a scaling benchmark against the real production function |
| **B — Codex (gpt-5.6-sol / xhigh)** | `swarm` solo (lanes 1–6; lane 7 skipped), `ce` report-only, `adversarial` | **No** — all three legs ran under a read-only sandbox; `npx tsc --noEmit` and `npx vitest` both failed to launch. Codex findings are source-read + GitNexus graph-tool based |

**Liveness:** all three `codex-*.md` files hold real findings, not error/empty markers. `codex-swarm` returned F1–F5 with a "not production-ready" verdict; `codex-ce` returned 6 normalized YAML findings; `codex-adversarial` returned 4 `[high]` findings. **Two Claude swarm lanes (`security`, `docs-dod`) produced no artifact** — that coverage rests on Codex swarm lanes 5 and 6, both of which found nothing. Findings re-verified first-hand by the coordinator are carried forward and not re-litigated. Everything in the findings table was re-checked by the synthesis pass in the worktree before promotion.

**Merge state: `review blocked`**

GitHub reports the branch `MERGEABLE` (no conflicts, correctly rebased — `git merge-base HEAD origin/main` = `79ff44df`, matching the stated base) with `mergeStateStatus: BLOCKED`. The block is a **pending review request from `azizur100389`**, not a code or branch problem. A coverage-merge rollup was still in flight at review time, and the Vercel deployment check fails as it does on every fork PR (see §9). No merge conflicts, no stale-branch problem, nothing requiring a rebase.

- **`INCREMENTAL_SCHEMA_VERSION` precedents v11 and v12 are directly on point** — both Rust, both `RUST_SCOPE_QUERY`/CALLS-edge changes, both bumped for exactly this reason. v11's own text: *"the incremental write set only covers changed files, so a top-up against a pre-v11 index would keep silently missing these CALLS edges for every unchanged Rust trait file; force a full re-analyze instead."* (#2604). v12 repeats the pattern. This PR is the same class of change and did not bump (finding in §7).
- **#2699 landed in `rust/query.ts` one day earlier**, adding a let-bound closure capture with explicit anchor discipline. This PR rebased over it and adds two more patterns to the same query — the anchor-interference risk was specifically probed and refuted by both engines.
- **#2604 and #2714** are the two intervening Rust rebaselines; #2714 moved the fingerprint to `7f1240b3…` without recording the transition in its own prose, which is the pre-existing gap that let this PR's stale "Prior" hash slip through.
- **#1978** (qualified node identity) is the closest prior art to the disclosed inline-`mod` limitation — CLOSED, fixed for C++/Ruby, with Rust explicitly deferred.
- **#1871 (C# `typeBindings` O(S²)), #2649 (kernel-scale OOM)** and the Go/Python scope-capture O(n²) fixes establish that super-linear work on a per-site resolution path is a recognised, previously-shipped regression class in this repo — relevant to the performance finding.

**Branch hygiene: `clean feature/fix PR`**

- Linear history, two coherent commits, no merge-from-main, no revert-then-reapply churn.
- Commit split is justified: semantic change plus its immediate golden output in one commit; the downstream benchmark metric in a follow-up.
- Scope discipline is clean. The `Set` → `ReadonlySet` widening in `import-resolvers/rust.ts` is correct covariance on two non-mutating functions, not churn.
- All 26 files justified. No formatting churn, no import reorganisation, no cross-language churn, no `.log`/`.tmp`/scratch files, no `console.log`, no commented-out debug, no `describe.skip`. Only `skills-lock.json` is untracked and is pre-existing.
- All three new fixtures are referenced by tests; no orphans.

## 7. Findings

### 7.1 What is SOLID

Real credit — much of this review's effort went into trying to break the PR and failing.

- **The core fix works, and the tests genuinely pin it.** All 7 headline assertions in `rust.test.ts` pin both `target` and `targetFilePath` via `toMatchObject`, gated on `edges.length === 1` with source-scoped predicates. Two lanes independently hand-traced or executed the pre-fix path and confirmed every one lands on the wrong same-file self-loop if the fix is reverted. No test pins nothing.
- **`crate::` / `self::` / `super::` anchor arithmetic is correct.** Deep chains (`super::super::super::x` from a 3-deep inline module), `crate::` from inside a nested `mod`, and single-segment `crate::f()` were all executed and match rustc.
- **`#[cfg(test)] mod tests` gets *better*, not worse** — `super::dispatch()` from a test module now targets the real function instead of the test-local double.
- **Blast-radius containment holds.** The "other 14 languages are byte-identical / zero overhead" claim survived a determined attack: the shared gate at `free-call-fallback.ts:236-240` requires both `resolveQualifiedFreeCall !== undefined` and `rawQualifiedName !== undefined`; `allFilePaths()` is a lazy closure only reachable inside that gate, so non-Rust passes never allocate the Set.
- **No new graph nodes.** `Namespace` is absent from `isLinkableLabel`, and `mod_item` was *already* a `Module` node via the untouched legacy structure query. The new capture is resolution-only. Traced, not inferred.
- **API/contract surface is clean.** Optional hook = source-compatible for all 14 other resolvers; the `ReadonlySet` widening is complete for both exports in that file and every caller still compiles; the hook's `site` shape is a structural subset of the pre-existing `ReferenceSite`.
- **`MODULE_INDEX_CACHE` works.** Three separate lanes attacked the WeakMap memo; the `??=` in `free-call-fallback.ts:135-136` means one Set identity per pass, so the crate index builds once. Not a bug.

**Notable refuted attacks** (probed and failed — do not re-raise):

| Attack | Outcome |
|---|---|
| #2699 closure-anchor displacement from the new query patterns | REFUTED by **both engines**. `@reference.qualified-name` is already in `KNOWN_SUB_TAGS`, so `anchorCaptureFor` skips it; the `closure_expression` anchor is untouched |
| Tree-sitter capture swallowing (`@scope.namespace` + `@declaration.namespace` on one `mod_item`) | REFUTED by **both engines** — separate matches, processed independently |
| `namespacePrefix` double-counting against Rust impl qualified names | REFUTED by both engines; `mod m { impl S { fn f } }` correctly yields `['m']` |
| An unqualified call acquiring `rawQualifiedName` | REFUTED — capture requires a literal `::`; turbofish and macros are disjoint patterns |
| `findMemberInModule`'s refuse-on-tie *logic* | REFUTED — the multi-match value is genuinely discarded, and the qualified-name index dedups `(name, DefId)` pairs, so a tie can't be one def counted twice |
| `pub use` cycle → infinite loop | REFUTED — `findReexportedMember` does not recurse; one hop only |
| `src/a.rs` + `src/a/mod.rs` coexisting; `mod foo;` + `mod foo {}`; module-level `use x as y` vs `mod y` | REFUTED — all rejected by rustc (E0761/E0255/E0428); invalid input, defensible output |
| `#[path = "..."]` remap producing a wrong edge | REFUTED — produces a miss, not a lie |
| New `Namespace` defs hijacking other resolution passes | REFUTED — `Namespace` is in none of the callable/class kind sets |
| Shared-ingestion "must not name languages" rule violated | REFUTED — the two Rust mentions are doc-comment prose in a file that already names 10 languages the same way; executable logic is fully generic |
| Test/CI lane's "fixture_count 174→186 mismatch" | REFUTED by coordinator — 12 new `.rs` fixtures, golden gained exactly 12; the note is correct |

**Honest caveat on the "strictly additive" claim.** The correctness lane could not construct a case where the new tier steals a previously-*correct* edge — except one (the `src/bin` finding below). So the contract's claim is accurate about edge *removal*, but not about edge *correctness*: the tier converts a number of `UNRESOLVED` sites into **false** edges, which the module's own header calls the worse outcome.

### 7.2 Headline findings

**H1 — Module identity discards the crate, so in a cargo workspace the fix either binds across crates or silently switches itself off.** `moduleOfFile` returns only the segments below a crate root; `sameModule` compares only those segments. Two crates' `src/tools.rs` are both module `['tools']`. Where only one defines the member → a **wrong cross-crate edge**. Where both define it → `count === 2` → refuse → the call falls back to the lexical chain and **the exact #2730 self-loop returns**. Same-named modules across workspace members (`util`, `error`, `config`, `types`) are close to universal — and **#2730's own reported repro repo is multi-crate**, while the PR's workspace fixture (`rust-2730-crate-layout`) has exactly one member and so cannot catch either manifestation. Found by every Codex leg and two Claude lanes; both manifestations executed; the synthesis pass re-ran `moduleOfFile` and confirmed the collision directly.

**H2 — The `use`-binding channel accepts an imported TYPE as a module, so `Type::method()` resolution is structurally unsound whenever it goes through an import binding.** Nothing checks that the qualifier head names a module. `tryRustModulePath` *deliberately* strips a trailing symbol segment ("last segment might be a symbol… Strip it and try again"), so `use crate::client::ClientBuilder;` yields `targetFile = src/client/mod.rs` and candidate 2 becomes module `['client']`. Rust impl methods carry a **bare** `qualifiedName` (confirmed: `function_item` captures only `@declaration.name`, no qualified-name capture exists), so the tail is looked up as an ordinary module member. Many such calls happen to land correctly — when the impl method is the only callable of that name in the target module, the accident works out. The failure is conditional but structural: it **produces a false edge when a module-level callable of the same name also exists** in that module (`new`, `default`, `from`, `build`, `load`, `parse` in a facade `mod.rs` — precisely the shape `pub use` motivates), and **misfires into H3's refuse-path when several do**. Reproduced independently by two Claude lanes with separate drivers.

**H3 — `findMemberInModule` counts defs that are not module members, so one ordinary nested helper disables qualified resolution for a whole module — restoring the #2730 self-loop.** Membership is inferred purely from `filePath + namespacePrefix`, with no check that the def is owned by a module-level scope. A `fn` nested inside another `fn`, or an `impl`/trait method (bare `qualifiedName`), is counted as a second module member → tie → refuse → the site is handed back to the tier that emits the wrong self-loop, with no outcome recorded. Found by **both engines** (Claude adversarial reproduced it on the #2730 shape verbatim; Codex adversarial and Codex CE reached it independently by source read).

### 7.3 Postable findings

| Sev | file:line | Problem | Engines / lanes | Trigger | Evidence |
|---|---|---|---|---|---|
| **P1** | `rust/module-path.ts:65`, `:137`; consumed `rust/qualified-call.ts:190` | **H1** — module identity drops the crate root; sibling workspace crates collide on module path → wrong cross-crate edge, or tie → refuse → #2730 self-loop returns | **A**: correctness F3, adversarial A-2 · **B**: swarm F1, adversarial, ce (conf 100) | Cargo workspace where two members both have `src/<name>.rs`; `<name>::member()` in one of them | **[reproduced]** ×2 lanes + synthesis pass re-ran `moduleOfFile`: `crates/a/src/tools.rs` and `crates/b/src/tools.rs` both → `["tools"]` |
| **P1** | `rust/qualified-call.ts:164-169` | **H2** — `use`-binding channel accepts an imported TYPE as a module; tail then resolves as an ordinary module member → false edge, or a tie feeding H3 | **A**: correctness F1, adversarial A-3 | `use crate::m::Type;` + `Type::new()` where `m/mod.rs` (or `m.rs`) also exposes a module-level `new`/`default`/`from`/`build`/`load` | **[reproduced]** ×2 independent drivers; mechanism re-verified — `tryRustModulePath` strips the trailing symbol segment by design |
| **P1** | `rust/qualified-call.ts:178-195` | **H3** — module membership inferred from file path alone; nested `fn` / impl / trait methods counted as module members → tie → refuse → self-loop returns | **A**: adversarial A-6 · **B**: adversarial, ce (conf 100) | Target module contains a function-local `fn dispatch` (or an `impl X { fn dispatch }`) alongside the real `pub fn dispatch` | **[reproduced]** (A-6, #2730 shape verbatim); synthesis pass confirmed no ownership filter at `:186-193` and no qualified-name capture on `function_item` |
| **P1** | `rust/qualified-call.ts:208-234`, gated by `rust/query.ts:172` | Hook fires on **every** `Path::name()` call (`Vec::new()`, `String::from()`), and `findReexportedMember` then scans all of `moduleScopeByFile` per candidate → measured **quadratic** in workspace size | **A**: performance F1, adversarial A-5, risk A+D · **B**: swarm F5, ce (conf 75/100) | Any large Rust repo; cost is paid by external/std qualified calls that can never resolve | **[reproduced]** — two independent benchmarks agree per-site cost is linear in file count (A-5: 0.117→0.228→0.428 ms/site at 301/601/1201 files; perf F1: 0.26→9.46 ms/call at N=500→16000) |
| **P1** | `src/storage/repo-manager.ts:546` | `INCREMENTAL_SCHEMA_VERSION` still 22 — the fix does not reach already-indexed repos, i.e. exactly the users who reported #2730. Exact precedents v11/v12 are the same file changing the same way (§5) | **coordinator CV-1** · **A**: risk B, api-contract F2 · **B**: adversarial | Upgrade a build containing this fix, run incremental `analyze` on an existing Rust index; unchanged files keep the pre-fix self-loop | **[code-read]** — verified: `git diff --name-only … -- src/storage/` returns 0 files |
| **P2** | `rust/module-path.ts:26`, `:42-51` | `src/bin/*.rs` given a fabricated module path (`['bin','tool']`) instead of being its own crate root — **the one case that downgrades a previously-correct edge** | **A**: correctness F2, adversarial A-1 · **B**: swarm F1, adversarial, ce | A package with both `src/main.rs` and a flat `src/bin/tool.rs`; `crate::helper()` or `helper::go()` inside the bin | **[reproduced]** ×2 lanes; synthesis pass confirms `moduleOfFile('src/bin/tool.rs') → ["bin","tool"]`. A-1's `f1-bin` shows `OLD=tool.rs::helper` (correct) → `NEW=main.rs::helper` (wrong crate) |
| **P2** | `rust/qualified-call.ts:215`, `:221-225` | `findReexportedMember` (a) cannot see inline-module re-exports — `moduleScopeByFile` holds only *file* module scopes, so `mod facade { pub use … }` is unreachable; (b) has **no tie check** — returns the first matching edge in file-iteration order (parse-pool order) | **B**: swarm F3+F4, ce (conf 100) · **A**: adversarial A-8 | `mod facade { pub use crate::tools::dispatch; }` + `facade::dispatch()`; or two `cfg`-exclusive facades re-exporting the same name | **[code-read]** — verified: `workspace-index.ts:18` documents `moduleScopeByFile` as "file path → `Scope` of the root `Module`" |
| **P2** | `rust/qualified-call.ts:158` | Candidate 1 (`callerModule ++ qualifier`) is an unchecked guess yielded **before** the `use` channel — it never verifies the caller actually declares that submodule, so an undeclared/`cfg`-gated sibling file or a block-scoped `use` loses to file layout | **A**: risk C, correctness F4, adversarial A-4 · **B**: swarm F2 (same root cause — function-local `use` resolved from the file-level bucket) | `src/a/mod.rs` has `use crate::b;` while an undeclared `src/a/b.rs` exists; or `fn main() { use crate::z::y; y::g(); }` with a top-level `mod y;` | **[reproduced]** ×2 lanes |
| **P2** | `scope-resolution/scope/walkers.ts:1054`, `:1078` (newly applied to Rust by `rust/method-owners.ts:33`) | The C++-era "already namespaced" guard mis-fires for Rust: `fn a` inside `mod a` (or `mod inner { fn inner }`) keeps `namespacePrefix` unset, so `moduleOfDef` reports it as a member of the **parent** module | **A**: adversarial A-7 · **B**: ce (conf 100) | `pub mod a { pub fn a() {} }` then `crate::a::a()` — refuses; worse, the def is now indistinguishable from a crate-root `fn a` for H1's matcher | **[reproduced]** (A-7); synthesis pass confirmed both guard sites |
| **P2** | `bench/scope-capture/baselines.json:59` | Rebaseline note cites the wrong "Prior" fingerprint — `655aed01…` is two rebaselines stale (pre-#2604/#2714); actual pre-PR value is `7f1240b3…`. CI cannot catch this (the gate compares the live fingerprint, not the prose) | **coordinator CV-2** · **A**: standards F1 | Reading the audit trail | **[reproduced]** — synthesis pass diffed the note against `git show <base>:…/baselines.json` |
| **P2** | PR body | "Filed separately rather than bundled here" for the inline-`mod` node-identity limitation — no such issue exists. Closest prior art is #1978 (CLOSED; fixed for C++/Ruby, Rust explicitly deferred) | **coordinator CV-3** · **B**: swarm lane 1 | Reading the PR body | **[code-read]** |
| **P2** | `src/storage/parse-cache.ts:122` | `SCHEMA_BUMP` still 30. Weaker than the P1 above because `PARSE_CACHE_VERSION = SCHEMA_BUMP + GITNEXUS_PKG_VERSION`, so a real release self-invalidates — the exposure is same-version dev/CI/source installs, where a warm cache replays pre-fix captures | **coordinator CV-5** · **A**: risk B, api-contract F1 · **B**: adversarial | Re-analyze on a same-version build with a warm parse cache (survives `--force`) | **[code-read]** — verified unchanged |

### 7.4 Lower-priority items

- **P3 — private `use` edges followed as re-exports** (`qualified-call.ts:221` accepts `kind === 'named'`). Raised by risk E, Codex CE and adversarial A-8 — **but the correctness lane refuted the impact**: the widening only mis-resolves paths that would not compile, and the target it produces is identical to what the pre-existing global-unique fallback already produced. Real hygiene issue, not a behaviour regression. Note the mirror-image gap: `kind === 'alias'` is *rejected*, so `pub use x::do_it as dispatch;` is never resolved at all.
- **P3 — leading-`::` anchor silently discarded** (`qualified-call.ts:76-82`): `::tools::dispatch()` has its empty leading segment filtered out and is treated as a relative path. In edition 2018 `::tools` names an *extern crate*. `[reproduced]`, single lane.
- **P3 — crate roots are derived from *parsed* files, not scanned files** (`free-call-fallback.ts:135-136`). An empty or extraction-skipped `lib.rs` (common for a scaffolded workspace member) means the crate has no root, so `moduleOfFile` returns `undefined` for every file in it and the feature silently turns off crate-wide. `tests/`, `examples/`, `benches/` are likewise never covered — and a path written with the crate *name* (`mypkg::tools::dispatch`) is never resolvable, while a path written *without* one happily crosses crate boundaries. `[reproduced]`, two Claude lanes.
- **P3 — duplicated callable-kind set.** `CALLABLE_TYPES` in `qualified-call.ts:48` duplicates the logic of `isOverloadableCallable` in `utils/callable-labels.ts`; a mismatch introduced by extending one without the other would silently miss qualified calls of the new type.
- **P3 — dead re-export block** (`qualified-call.ts:250-252`). Verified: the only importer of `qualified-call.ts` anywhere in `src`/`test` is `rust/scope-resolver.ts:8`, which takes just `resolveRustQualifiedFreeCall`. Nothing imports `moduleOfFile`, `moduleOfDef` or the `ScopeResolutionIndexes` type from it — the first two are already exported by `module-path.ts`, and `ScopeResolutionIndexes` is canonically exported from `model/scope-resolution-indexes.ts:58` and imported from there by 41 files. The comment claims they exist "for the resolver's unit tests"; no such unit tests exist.
- **P2, conditional — disk-fault amplifier.** Under `GITNEXUS_DISK_SCOPE_INDEX=1` (default-off), `findReexportedMember`'s `moduleScopeByFile` walk runs after the scope-tree seal, turning the P1 CPU scan into per-file `readFileSync` + `JSON.parse` on LRU misses — the exact pattern `exportedCallableByName` was added to avoid, per its own comment. Single lane, conf 60, `[code-read]`.
- **Pre-existing, not this PR:** turbofish path calls (`tools::dispatch::<u32>()`) parse as `generic_function` and produce no reference site at all — invisible to both the old and new paths.

## 8. Verdict

**`not production-ready`**

This is a genuinely good piece of engineering — the module-tree model is the right abstraction, it follows rustc's real rule instead of a filename heuristic, the anchor arithmetic is correct, the seam through `ScopeResolver` is clean, and the tests that exist are real regression tests. Most of the attacks mounted against it failed. But three things block it and they compound: the fix **does not reach already-indexed repos** at all (`INCREMENTAL_SCHEMA_VERSION` unbumped, with an exact same-file precedent); where it does reach them, **a cargo workspace can turn it off or point it at the wrong crate** — and #2730's own repro repo is multi-crate while the PR's workspace fixture has one member, with H3 reverting a module to the original self-loop from a second direction; and it **converts unresolved sites into false edges** on shapes the module's own header explicitly rejects, which `impact` then consumes as exact. The quadratic reexport scan is separately worth fixing before this meets the repo's large-repo bar, given the #2649/#1871 history.

Suggested path: make module identity `{ crateRoot, segments }` and thread it through member/reexport lookup (fixes H1, most of the bin case, and the `$crate` inheritance); restrict `findMemberInModule` to module-level-owned defs (fixes H3); require the `use`-channel head to actually resolve to a module (fixes H2); build the module→scope index once per pass (fixes the quadratic scan and the disk-fault exposure); bump both schema constants. Then add the two-crate, `src/bin`, nested-`fn`, and `Type::assoc()` fixtures plus unit tests for `module-path.ts`.

## 9. CI status

All substantive checks are green: CI Gate, tests (15,740 passing / 0 failing / 69 skipped), typecheck, formatting, CodeQL, gitleaks, dependency review, benchmarks, and the cross-platform matrix. The scope-capture `--check` fingerprint gate passes. Fork PRs do run the full suite — `ci.yml` triggers on `pull_request` to main with no fork exclusion, and the Rust resolver tests are in the sharded coverage suite.

**The only failing check is `Vercel`, which fails on every fork PR because the fork cannot authorize a deploy. It is not a code defect and should not be read as one.**

## 10. Test gaps

The suite is strong on the shapes it covers and blank on the ones the findings live in:

- **No two-crate fixture** — would have caught H1 in both manifestations.
- **No `src/bin/*.rs`, `tests/`, `examples/` or `build.rs` fixture.**
- **No fixture with a nested `fn` or an `impl` method in the target module** (H3) — the single most likely reason for the fix to silently not apply in real code.
- **No `Type::assoc()` fixture** (H2), even though the new capture routes every `Foo::bar()` site through the hook.
- **No unit tests for `module-path.ts` at all** — it is pure, table-driven path arithmetic, and the cheapest possible place to pin H1 / the bin case / the `::`-anchor behaviour. Every current assertion goes through the full pipeline for two layouts.
- **No negative/refusal test.** The module's headline contract is "refuse rather than emit a wrong edge"; nothing asserts a tie or unknown module leaves the site on the unchanged chain.
- **Fixture rot:** `rust-2730-gaps/src/main.rs:7-15` contains the inline-`mod` case (the PR's own disclosed limitation) but the tests filter to `['nested','go','via_reexport']` and never assert on it — so today's behaviour is unpinned and could drift silently. Same for the `wrapper`/`helper` control in `rust-2730-samename-wrapper`, which is commented as a control but never asserted.
- **The rebaselined `bench/scope-capture` measures the wrong phase.** `measure.mjs` times `emit<Lang>ScopeCaptures` (the capture phase) only — it never touches `runScopeResolution`/`emitFreeCallFallback`, so it provides no coverage, positive or negative, of the complexity this PR adds.

## 11. Open questions

1. Which issue tracks the inline-`mod` node-identity limitation? The PR body says it was "filed separately rather than bundled here", but no such issue was found — please cite the number or reword.
2. Is non-`main.rs`/`lib.rs` Cargo target discovery (`src/bin/*`, `examples/`, `tests/`, `[[bin]] path=`) intended to be in scope for this PR, or deliberately deferred? The answer changes whether the `src/bin` item is a fix-now or a documented limitation.

## 12. Closing note

*This is an automated multi-tool review digest (Claude swarm + CE personas, Codex gpt-5.6-sol/xhigh, coordinator verification). Findings were cross-checked against the worktree before promotion, but please verify against your own understanding of the code before acting on any of them.*

gitnexus/src/core/ingestion/languages/rust/module-path.ts
magyargergo: Fixed in `eee9e605`.

A module is now `{ crateRoot, segments }` and `sameModule` compares both. `resolveAnchoredModulePath` keeps an anchored path inside the caller's own crate, since `crate::`/`self::`/`super::` are all relative to it — Rust has no implicit cross-crate paths, so two modules in different crates are never the same module.

Verified directly:

```
crateRoots: ["crates/alpha/src", "crates/beta/src"]
crates/alpha/src/tools.rs -> { crateRoot: crates/alpha/src, segments: ["tools"] }
crates/beta/src/tools.rs  -> { crateRoot: crates/beta/src,  segments: ["tools"] }
sameModule(alpha, beta) === false
```

Covered by a new two-member workspace fixture (`rust-2730-workspace-crates`) where BOTH crates define `tools::dispatch` behind a same-name wrapper — so the tie manifestation you identified is what the test would hit on a regression — plus unit tests for the path arithmetic, including the branches no fixture reaches (a file under no crate root, a `super::` chain walking above the root).

gitnexus/src/core/ingestion/languages/rust/qualified-call.ts
magyargergo: Fixed in `52ac5326`.

A `use` binding now has to name the module it resolved to. The edge's `targetExportedName` is the tail of the written path, so comparing it against the resolved module's own tail separates the cases exactly — and, importantly, without breaking the legitimate alias form:

```
use crate::tools;                 tail `tools`         module ["tools"]    accept
use crate::a::b as tools;         tail `b`             module ["a","b"]    accept
use crate::tools::{self, Ctx};    tail `tools`         module ["tools"]    accept
use crate::client::ClientBuilder; tail `ClientBuilder` module ["client"]   reject
```

Your diagnosis of the mechanism was right and worth recording: `tryRustModulePath` strips a trailing symbol segment *by design* ("the last segment might be a symbol, not a module"), which is why the type landed on `client/mod.rs` in the first place.

New fixture `rust-2730-type-qualified` puts both `impl ClientBuilder { fn new }` and a module-level `fn new` in the same `client/mod.rs`, so a regression re-binds to the wrong one rather than merely failing to resolve, plus a control that a genuine `client::new()` module qualifier still works.

gitnexus/src/core/ingestion/languages/rust/qualified-call.ts
magyargergo: Fixed in `e2ed1b7a`.

Confirmed the premise first — a function-local `fn` really is indistinguishable from a module-level one on the fields the lookup had:

```
Function  qn=dispatch  owner=undefined  nsPrefix=undefined   <- module-level
Function  qn=dispatch  owner=undefined  nsPrefix=undefined   <- nested inside fn outer
Method    qn=dispatch  owner=def:...Struct:                  <- impl method
```

But the scope model draws the line exactly, so membership is now a binding lookup rather than a path comparison:

```
Module     bindings=[dispatch(local) outer(local) inner(local)]   <- module-level
Block      bindings=[dispatch(local)]                            <- function-local binds HERE
Namespace  bindings=[dispatch(local)]                            <- inline-mod member
```

A module-level item binds `origin:'local'` in its module's own scope; a function-local one binds in the enclosing Block; an impl/trait method binds in the Class scope.

One honest residual, stated in the code: inline-`mod` members bind in their Namespace scope, and reaching it would mean walking every child scope — faulting the tree back in from disk on the out-of-core path. They keep being identified by `namespacePrefix` instead, so a `fn` nested inside a `fn` inside an inline `mod` inherits that prefix and is still counted. Strictly smaller than before, and it costs a refusal rather than a wrong edge.

The `rust-2730-gaps` fixture now carries the nested-shadow shape.

gitnexus/src/core/ingestion/languages/rust/qualified-call.ts
magyargergo: Fixed in `6906e7e2`, in the two places you identified.

**The `Vec::new()` problem — reject before any work.** The module index now carries a flat set of every module segment name in the workspace, and a qualifier whose head matches none of them is rejected before the candidate search runs at all. Measured:

```
N=  500  buildIndex=0.5ms  20k Vec::new() rejections=2.0ms (0.10 us/call)
N= 2000  buildIndex=1.4ms  20k Vec::new() rejections=0.5ms (0.02 us/call)
N= 8000  buildIndex=5.7ms  20k Vec::new() rejections=0.3ms (0.02 us/call)
```

Flat in file count, against the 0.117 -> 0.428 ms/site you measured growing linearly across 301 -> 1201 files. Anchors are handled: the segment tested is the first non-anchor one, so `self::f()` still resolves.

**The full-workspace walk — index once per pass.** Module scopes are now keyed by module identity in a `WeakMap` on the `WorkspaceResolutionIndex`, built once, instead of rediscovered per candidate.

Your point about the out-of-core path was the sharper half and is recorded in the code: `moduleScopeByFile` fetches through `scopeTree.getScope`, so a full sweep could fault every module scope back in from disk — exactly what `exportedCallableByName` was added to avoid. That made this worth fixing before merge rather than after, given the #2649 / #1871 history.

gitnexus/src/core/ingestion/languages/rust/qualified-call.ts
magyargergo: Fixed in `73a224e0`.

The candidate is now gated on the caller actually declaring the submodule. A `mod` declaration — inline or file-backed — emits a `Namespace` def bound locally in the declaring scope, so the check is a binding lookup rather than a filesystem probe:

```ts
function declaresSubmodule(callerParsed, name, workspaceIndex): boolean {
  const moduleScope = workspaceIndex.moduleScopeByFile.get(callerParsed.filePath);
  for (const ref of moduleScope?.bindings.get(name) ?? []) {
    if (ref.origin === 'local' && ref.def.type === 'Namespace') return true;
  }
  return false;
}
```

When the caller does not declare it the candidate is skipped and the `use` and crate-root channels still run, so this only removes guesses — it cannot cost a resolution that was previously correct.

Worth noting this also removes a wrong guess the crate-layout fixture was silently relying on: `sched.rs` has `use crate::tools::{self, ToolCtx}` and no `mod tools;`, so candidate 1 used to propose `agent::sched::tools` before the `use` channel got a turn. It happened to miss; now it is never proposed.

gitnexus/src/core/ingestion/languages/rust/method-owners.ts
magyargergo: Fixed in `7ec554f3`.

Reproduced first — `pub mod a { pub fn a() }` gave `nsPrefix=undefined`, so the member looked like it belonged to the parent module. After the fix it is `nsPrefix=a`, and the Namespace def itself correctly stays untagged.

The fix deliberately did NOT change the shared pass's behaviour. That guard is correct for C++ and C#, where `qualifiedName` genuinely carries the namespace; it is only wrong for Rust, where it never does. So `tagNamespacePrefixes` gained a `qualifiedNamesCarryNamespace` option that defaults to the existing behaviour, and Rust opts out — the decision lives with the provider that knows what its own qualified names contain, which also keeps `src/core/ingestion/` language-neutral per AGENTS.md.

C++ and C# resolver suites pass unchanged alongside the Rust ones (600 tests). New unit file `rust-namespace-prefix.test.ts` pins the same-name case, the differing-name case, nested composition, an untagged crate-root item, and that the module def itself is not tagged.

gitnexus/bench/scope-capture/baselines.json
magyargergo: Fixed in `1e6de97b`.

Confirmed against the base commit before changing it:

```
git show <base>:gitnexus/bench/scope-capture/baselines.json
  rust.fingerprint = 7f1240b38457468f06b7931e0c2c578f218f922774d0dc7e2ee6ef3b08d4d689
```

`655aed01…` was the post-#2514 value; #2604 and #2714 both moved it after that, so the note was two rebaselines stale. The note now carries `7f1240b3…` as Prior, and explicitly records that the earlier revision was wrong and why nothing caught it — the gate compares live fingerprints, never the prose, so the audit chain these notes exist to provide can break silently.

Fingerprint regenerated for the fixtures this review series added: scaling 1.061 (budget 1.5), `fixture_count` 196, other 14 languages byte-identical. `measure.mjs --check` passes for all 15.

magyargergo:
## Review addressed — all 13 findings, one per commit

Every finding from the multi-engine review now has its own commit, and each inline thread has a reply with the mechanism, the commit and the verification. Summary:

| Sev | Finding | Commit |
|---|---|---|
| P1 | H1 module identity dropped the crate | `eee9e605` |
| P1 | H3 membership inferred from file path | `e2ed1b7a` |
| P1 | H2 `use` binding accepted an imported type as a module | `52ac5326` |
| P1 | hook ran for every `Path::name()`; O(files) scan per candidate | `6906e7e2` |
| P1 | `INCREMENTAL_SCHEMA_VERSION` / `SCHEMA_BUMP` unbumped | `d88e49fd` |
| P2 | `src/bin/*.rs` given a fabricated module path | `567233cc` |
| P2 | candidate 1 was an unchecked guess | `73a224e0` |
| P2 | private `use` followed as a re-export; no tie check; `alias` ignored | `8e4e605e` |
| P2 | already-namespaced guard misfired on `mod a { fn a }` | `7ec554f3` |
| P2 | rebaseline note cited a fingerprint two rebaselines stale | `1e6de97b` |
| P2 | "filed separately" was unsupported | issue #2742 + PR body |
| P3 | leading `::` read as relative instead of extern-crate | `9d6cc21d` |
| P3 | duplicated callable-kind set; dead re-exports | `9d7eb1d0` |

Plus `6e25a312`, updating the `INCREMENTAL_SCHEMA_VERSION` pin test that the bump correctly broke.

### The two findings that mattered most

**The fix did not reach the people who reported the bug.** `INCREMENTAL_SCHEMA_VERSION` has no package-version component, so without a bump an incremental analyze keeps serving the pre-fix graph for unchanged Rust files indefinitely. The precedent is exact — v11 is the same file gaining a capture that changes CALLS edges — and I had wrongly assumed the release version bump covered both constants. It covers only `SCHEMA_BUMP`.

**The fix switched itself off in a cargo workspace.** With identity by path segments alone, two members' `src/tools.rs` were the same module: a wrong cross-crate edge where only one defined the member, and a refusal where both did — which handed the site straight back to the self-loop tier this PR exists to remove. #2730's own reproduction repo is multi-crate, and the original workspace fixture had a single member, so nothing here could have caught it.

### Two review lessons worth recording

Two regression tests I wrote first were **wrong in an instructive way**: I asserted that a private `use` and a leading `::` path produce no edge, and both failed — those edges come from the pre-existing lexical and global-unique tiers, which this hook does not govern. The fix was not to weaken them. One fixture now makes the name globally ambiguous so the unique-global fallback cannot fire, isolating the tier under test; the other asserts which target is forbidden rather than asserting emptiness. A test that passes because a different tier answered pins nothing.

The `tagNamespacePrefixes` bug went **behind a provider flag rather than into the shared pass**, because the guard is correct for C++/C# and wrong only for Rust.

### Validation

Full local sweep across every resolver language and all scope-resolution units: **4272 passed, 1 skipped, 0 failed**. C++ and C# suites pass unchanged alongside Rust (600 tests) for the shared-pass change. `bench/scope-capture --check` passes for all 15 languages.

CI is green except `Vercel` (non-actionable on fork PRs) and Windows `platform-sensitive 2/3`, which fails at `analyze --force` with `status null` — a 60s timeout, not an assertion. That test's fixture is 7 TypeScript files with no Rust, `--force` on a fresh temp repo makes both schema bumps inapplicable, and every shared path I touched is gated on a hook that is undefined for TypeScript — so I can find no mechanism by which this series causes it. I have re-run that job on the identical commit to separate code from runner, and will report the result rather than assume.

PR #2743, opened 2026-07-29
fix(scope-resolution): parse def coordinates after file paths

### LINKED ISSUE #2734: Closure attribution silently fails for any file under a path containing '#<digits>:<digits>:'
## Summary

`scopeIsCallableBody` and `defStartLine` parse a scope-resolution def id with a **non-anchored, first-match** regex. Node ids embed the analyzed repo's file path verbatim, so a path containing `#<digits>:<digits>:` makes the match land on the **path** fragment instead of the real line/col separator. Closure call-source attribution then silently fails for every file under that path.

Found by the multi-engine review of #2718 (security lane), and reproduced end-to-end.

## Reproduction

Two identical PHP closures, differing only in their directory:

```
src/plain.php
src/bugs#12:34:test/a.php
```

```php
<?php
function target($x) { return $x; }
function outer() {
  $handler = function ($x) { return target($x); };
  return 1;
}
```

`analyze --index-only`, then all CALLS edges:

```
Function:src/plain.php:outer2.$handler2@3:2 -> Function:src/plain.php:target2
```

The file under `src/bugs#12:34:test/` produces **no CALLS edge at all**. Its closure is never accepted as a callable boundary.

## Mechanism

`scope-extractor.ts:870` builds the id with no sanitization of `filePath`:

```ts
return `def:${filePath}#${range.startLine}:${range.startCol}:${type}:${name}`;
```

Both consumers take the first match:

```ts
// ids.ts:102  (scopeIsCallableBody)
// ids.ts:170  (defStartLine)
const m = def.nodeId.match(/#(\d+):(\d+):/);
```

For `def:src/bugs#12:34:test/a.php#4:2:Function:$handler` the regex yields `m[1]="12"`, `m[2]="34"`, so the comparison against the real scope start (`4:2`) fails.

The same pattern exists at `callable-value-flow.ts:818`, `:924`, `:1006` and should be audited alongside.

## Suggested fix

Anchor extraction to the **last** `#`. The `<line>:<col>:<type>:<name>` tail contains no `#`, so this is unambiguous:

```ts
const lastHash = def.nodeId.lastIndexOf('#');
if (lastHash === -1) return false;
const m = def.nodeId.slice(lastHash).match(/^#(\d+):(\d+):/);
```

## Notes

- Severity is bounded by how rare such paths are (`:` is illegal in Windows paths; legal on Linux/macOS), but the failure is **silent** — no warning, just missing edges.
- After #2718's fail-closed change the symptom is a *missing* edge. Before it, the same miss produced a *fabricated* edge attributed to the enclosing function, so this also silently degraded attribution on affected paths.
- No test covers a def id whose path contains `#N:M:`.

azizur100389:
## Summary
- anchor definition-ID coordinate parsing to the known file path
- route all scope-resolution coordinate consumers through one parser
- cover coordinate-like paths, private names, malformed IDs, and closure attribution

Closes #2734

## Test plan
- [x] `npx vitest run test/unit/scope-resolution/definition-id.test.ts test/unit/scope-resolution/callable-value-target-index.test.ts test/integration/closure-review-findings.test.ts --testTimeout=120000` (19 passed, 1 Windows-only skip)
- [x] `npx vitest run test/unit/scope-resolution test/integration/closure-review-findings.test.ts test/integration/resolvers/callable-value-flow.test.ts --testTimeout=120000` (1,397 passed, 1 Windows-only skip)
- [x] `npx tsc --noEmit`
- [x] Prettier and ESLint on changed files
- [x] Callable-value-flow benchmark check (4,000 targets; fingerprint and timing budgets pass)
- [ ] Full local `npm test` on Windows: 15,008 passed, 411 skipped, 73 environment-sensitive failures; PR CI is the merge gate

## Compatibility
- Indexes created on schema v24 before this fix need a full re-analysis to correct any existing edges for paths containing coordinate-like fragments.

### DIFF
--- a/gitnexus/src/core/ingestion/scope-resolution/graph-bridge/ids.ts
+++ b/gitnexus/src/core/ingestion/scope-resolution/graph-bridge/ids.ts
@@ -34,6 +34,7 @@ import {
 } from '../../utils/callable-labels.js';
 import { templateConstraintsIdTag } from '../../utils/template-arguments.js';
 import { parameterShapeIdTag } from '../../utils/method-props.js';
+import { definitionIdPosition } from '../utils/definition-id.js';
 /**
  * Labels that may legitimately ANCHOR a CALLS/ACCESSES edge as the
  * source ("caller"). A Variable / Property can be the TARGET of an
@@ -99,9 +100,9 @@ function scopeIsCallableBody(
   range: { startLine: number; startCol: number },
   def: SymbolDefinition,
 ): boolean {
-  const m = def.nodeId.match(/#(\d+):(\d+):/);
-  if (m === null) return false;
-  return Number(m[1]) === range.startLine && Number(m[2]) === range.startCol;
+  const position = definitionIdPosition(def.nodeId, def.filePath);
+  if (position === undefined) return false;
+  return position.line === range.startLine && position.column === range.startCol;
 }
 
 /** Pick the callable that owns `atRange` when multiple overloads share a class scope. */
@@ -165,12 +166,8 @@ function pickCallerCallableDef(
  * Extract the 1-based declaration line from a scope-resolution def id.
  * Shape: `def:<filePath>#<line>:<col>:<...>`; `undefined` when it doesn't match.
  */
-function defStartLine(nodeId: string | undefined): number | undefined {
-  if (nodeId === undefined) return undefined;
-  const m = nodeId.match(/#(\d+):(\d+):/);
-  if (m === null) return undefined;
-  const line = Number(m[1]);
-  return Number.isFinite(line) ? line : undefined;
+function defStartLine(nodeId: string | undefined, filePath: string): number | undefined {
+  return definitionIdPosition(nodeId, filePath)?.line;
 }
 
 /**
@@ -221,7 +218,7 @@ export function resolveDefGraphId(
     // without either side having to model the scope chain. Node ids are
     // 0-based, def ids 1-based. An `AMBIGUOUS_POSITION` tombstone (two
     // callables on one line) falls through to the name-based keys below.
-    const line = defStartLine(def.nodeId);
+    const line = defStartLine(def.nodeId, filePath);
     if (line !== undefined && isPositionQualifiedLocalLabel(def.type)) {
       const simple = simpleNameOf(qn);
       const posHit = nodeLookup.get(positionKey(filePath, def.type, line - 1, simple));
--- a/gitnexus/src/core/ingestion/scope-resolution/utils/definition-id.ts
+++ b/gitnexus/src/core/ingestion/scope-resolution/utils/definition-id.ts
@@ -1,0 +1,24 @@
+export interface DefinitionIdPosition {
+  readonly line: number;
+  readonly column: number;
+}
+
+/**
+ * Extract coordinates from `def:<filePath>#<line>:<column>:<type>:<name>`.
+ *
+ * File paths are embedded verbatim and may themselves contain fragments such
+ * as `#12:34:`, while names may themselves contain `#` (TypeScript private
+ * members). Anchor on the known file path so neither side can be mistaken for
+ * the declaration separator.
+ */
+export function definitionIdPosition(
+  nodeId: string | undefined,
+  filePath: string,
+): DefinitionIdPosition | undefined {
+  if (nodeId === undefined) return undefined;
+  const prefix = `def:${filePath}#`;
+  if (!nodeId.startsWith(prefix)) return undefined;
+  const match = /^(\d+):(\d+):/.exec(nodeId.slice(prefix.length));
+  if (match === null) return undefined;
+  return { line: Number(match[1]), column: Number(match[2]) };
+}
--- a/gitnexus/bench/callable-value-flow/baselines.json
+++ b/gitnexus/bench/callable-value-flow/baselines.json
@@ -1,6 +1,6 @@
 {
-  "_comment": "Baselines for bench/callable-value-flow/measure.mjs --check (#2693). `fingerprint` is an order-independent sha256 over every (defNodeId -> graphId) pair buildGraphTargetIndex resolves on the synthetic corpus; it is a CORRECTNESS gate, so drift means the callable-value target set moved and must be explained, never re-baselined to make CI green. The two budgets are timing gates and carry deliberate headroom for shared CI runners.",
-  "fingerprint": "70bebf6a26ff6fc9f231a0933678274b44c4883ddab5e719a61a9c77d6223e51",
+  "_comment": "Baselines for bench/callable-value-flow/measure.mjs --check (#2693). `fingerprint` is an order-independent sha256 over every (defNodeId -> graphId) pair buildGraphTargetIndex resolves on the synthetic corpus; it is a CORRECTNESS gate, so drift means the callable-value target set moved and must be explained, never re-baselined to make CI green. The fingerprint changed when the synthetic corpus adopted production-shaped def ids; target cardinality remains 4000 (3200 callable-only plus 800 value bindings). The two budgets are timing gates and carry deliberate headroom for shared CI runners.",
+  "fingerprint": "6599dda7d0ee5942e1995a1dcfb137c312eb8e4690bc82a8bbff429f08bd839d",
   "scaling_budget": 1.6,
   "_scaling_note": "(t_large/t_small)/(800/250). ~1.0 is linear; measured 1.14-1.16. The index build is one pass over defs plus map lookups, so a jump toward 3.x means someone made the per-def work depend on corpus size (e.g. a scan inside the loop).",
   "widening_overhead_budget": 1.9,
--- a/gitnexus/bench/callable-value-flow/measure.mjs
+++ b/gitnexus/bench/callable-value-flow/measure.mjs
@@ -89,7 +89,7 @@ function buildCorpus(fileCount) {
     return id;
   };
   const addDef = (type, filePath, qualifiedName, line) => {
-    const nodeId = `${filePath}#${line}:0:${qualifiedName}`;
+    const nodeId = `def:${filePath}#${line}:0:${type}:${qualifiedName}`;
     defs.set(nodeId, { nodeId, type, filePath, qualifiedName });
   };
 
--- a/gitnexus/src/core/ingestion/scope-resolution/passes/callable-value-flow.ts
+++ b/gitnexus/src/core/ingestion/scope-resolution/passes/callable-value-flow.ts
@@ -26,6 +26,7 @@ import {
   simpleQualifiedName,
 } from '../graph-bridge/ids.js';
 import { resolveInheritanceBaseInScope } from '../scope/walkers.js';
+import { definitionIdPosition } from '../utils/definition-id.js';
 import { narrowOverloadCandidates } from './overload-narrowing.js';
 
 export const MAX_CALLABLE_VALUE_TARGETS = 32;
@@ -815,7 +816,7 @@ const withoutSigil = (name: string): string => name.replace(/^[$@]+/, '');
  */
 function valueBindingPositionKey(def: SymbolDefinition): string | undefined {
   if (!VALUE_BINDING_DEF_TYPES.has(def.type)) return undefined;
-  const line = def.nodeId.match(/#(\d+):(\d+):/)?.[1];
+  const line = definitionIdPosition(def.nodeId, def.filePath)?.line;
   const name = simpleQualifiedName(def);
   if (line === undefined || name === undefined) return undefined;
   return `${def.filePath}\0${line}\0${withoutSigil(name)}`;
@@ -921,8 +922,10 @@ function canonicalizeCallableDeclarations(
 }
 
 function definitionPosition(def: SymbolDefinition): string | undefined {
-  const match = def.nodeId.match(/#(\d+):(\d+):/);
-  return match === null ? undefined : `${def.filePath}\0${match[1]}\0${match[2]}`;
+  const position = definitionIdPosition(def.nodeId, def.filePath);
+  return position === undefined
+    ? undefined
+    : `${def.filePath}\0${position.line}\0${position.column}`;
 }
 
 function declarationSignatureCompatible(
@@ -1003,7 +1006,7 @@ function buildGraphCallableIndexes(graph: KnowledgeGraph): GraphCallableIndexes 
 }
 
 function definitionAnchorKey(def: SymbolDefinition): string | undefined {
-  const line = def.nodeId.match(/#(\d+):(\d+):/)?.[1];
+  const line = definitionIdPosition(def.nodeId, def.filePath)?.line;
   const name = simpleName(def.qualifiedName);
   if (line === undefined || name === undefined) return undefined;
   return `${def.filePath}\0${def.type}\0${line}\0${name}`;
@@ -1437,8 +1440,8 @@ function dedupeDefinitions(defs: readonly SymbolDefinition[]): SymbolDefinition[
 }
 
 function definitionStartsAt(def: SymbolDefinition, line: number, column: number): boolean {
-  const match = def.nodeId.match(/#(\d+):(\d+):/);
-  return match !== null && Number(match[1]) === line && Number(match[2]) === column;
+  const position = definitionIdPosition(def.nodeId, def.filePath);
+  return position !== undefined && position.line === line && position.column === column;
 }
 
 function expandMemberTargets(
--- a/gitnexus/test/integration/closure-review-findings.test.ts
+++ b/gitnexus/test/integration/closure-review-findings.test.ts
@@ -18,12 +18,16 @@ import { DIST_WORKER_URL, distWorkerExists } from '../helpers/worker-parse.js';
 vi.setConfig({ testTimeout: 90_000 });
 
 const describeIfWorkerBuilt = distWorkerExists() ? describe : describe.skip;
+const describeIfPathSupportsColon =
+  process.platform === 'win32' ? describe.skip : describeIfWorkerBuilt;
 
 /** Every CALLS edge in a one-file repo, as `src -> dst`, sorted. */
 const callEdges = async (filename: string, source: string): Promise<string[]> => {
   const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'gn-review-'));
   try {
-    fs.writeFileSync(path.join(dir, filename), source, 'utf-8');
+    const filePath = path.join(dir, filename);
+    fs.mkdirSync(path.dirname(filePath), { recursive: true });
+    fs.writeFileSync(filePath, source, 'utf-8');
     const result = await runPipelineFromRepo(dir, () => {}, {
       workerPoolSize: 1,
       workerUrlForTest: DIST_WORKER_URL,
@@ -84,6 +88,20 @@ describeIfWorkerBuilt(
     });
   },
 );
+
+describeIfPathSupportsColon('#2734 — def ids parse coordinates after the file path', () => {
+  it('a path containing #<line>:<column>: still attributes closure calls', async () => {
+    const edges = await callEdges(
+      'src/bugs#12:34:test/a.php',
+      '<?php\nfunction target($x) { return $x; }\nfunction outer() {\n' +
+        '  $handler = function ($x) { return target($x); };\n  return 1;\n}\n',
+    );
+
+    expect(edges).toEqual([
+      'Function:src/bugs#12:34:test/a.php:outer.$handler@3:2 -> Function:src/bugs#12:34:test/a.php:target',
+    ]);
+  });
+});
 
 describeIfWorkerBuilt(
   '#2699 review P1-2 — widening identity to values must not touch class members',
--- a/gitnexus/test/unit/scope-resolution/callable-value-target-index.test.ts
+++ b/gitnexus/test/unit/scope-resolution/callable-value-target-index.test.ts
@@ -33,7 +33,7 @@ const graphOf = (nodes: readonly StubNode[]): KnowledgeGraph => {
 
 /** `line` is 1-based, matching the definition-id convention. */
 const def = (type: string, filePath: string, qualifiedName: string, line: number) => ({
-  nodeId: `${filePath}#${line}:0:${qualifiedName}`,
+  nodeId: `def:${filePath}#${line}:0:${type}:${qualifiedName}`,
   type,
   filePath,
   qualifiedName,
@@ -69,7 +69,7 @@ describe('buildGraphTargetIndex — value bindings join by position', () => {
     // the same construct, so they share file, line and name.
     expect(
       targetsFor([node('Function', 'a.kt', 'handler', 3)], [def('Property', 'a.kt', 'handler', 3)]),
-    ).toEqual(['a.kt#3:0:handler => Function:a.kt:handler']);
+    ).toEqual(['def:a.kt#3:0:Property:handler => Function:a.kt:handler']);
   });
 
   it('REJECTS a value binding that merely shares a name with a callable elsewhere', () => {
@@ -106,12 +106,12 @@ describe('buildGraphTargetIndex — value bindings join by position', () => {
         [node('Function', 'a.php', '$save', 5), node('Function', 'a.php', 'save', 2)],
         [def('Variable', 'a.php', 'save', 5)],
       ),
-    ).toEqual(['a.php#5:0:save => Function:a.php:$save']);
+    ).toEqual(['def:a.php#5:0:Variable:save => Function:a.php:$save']);
   });
 
   it('keeps ordinary callable defs, which never take the positional path', () => {
     expect(
       targetsFor([node('Function', 'a.ts', 'fn', 1)], [def('Function', 'a.ts', 'fn', 1)]),
-    ).toEqual(['a.ts#1:0:fn => Function:a.ts:fn']);
+    ).toEqual(['def:a.ts#1:0:Function:fn => Function:a.ts:fn']);
   });
 });
--- a/gitnexus/test/unit/scope-resolution/definition-id.test.ts
+++ b/gitnexus/test/unit/scope-resolution/definition-id.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, it } from 'vitest';
+import { definitionIdPosition } from '../../../src/core/ingestion/scope-resolution/utils/definition-id.js';
+
+describe('definitionIdPosition', () => {
+  it('reads coordinates after a path containing a coordinate-like fragment (#2734)', () => {
+    expect(
+      definitionIdPosition(
+        'def:src/bugs#12:34:test/a.php#4:2:Function:$handler',
+        'src/bugs#12:34:test/a.php',
+      ),
+    ).toEqual({ line: 4, column: 2 });
+  });
+
+  it('does not mistake a private-name hash for the path separator', () => {
+    expect(definitionIdPosition('def:src/a.ts#4:2:Method:#secret', 'src/a.ts')).toEqual({
+      line: 4,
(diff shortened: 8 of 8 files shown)


gitnexus/src/core/ingestion/scope-resolution/utils/definition-id.ts
```diff
@@ -0,0 +1,24 @@
...
+ * members). Anchor on the known file path so neither side can be mistaken for
+ * the declaration separator.
+ */
+export function definitionIdPosition(
+  nodeId: string | undefined,
+  filePath: string,
+): DefinitionIdPosition | undefined {
+  if (nodeId === undefined) return undefined;
+  const prefix = `def:${filePath}#`;
+  if (!nodeId.startsWith(prefix)) return undefined;
```
magyargergo, line 20: **P1 — this precondition breaks the committed callable-value-flow bench, and CI is red because of it.** `[reproduced]`

`bench/callable-value-flow/measure.mjs:92` still mints synthetic def ids the old way:

```js
const nodeId = `${filePath}#${line}:0:${qualifiedName}`;   // no `def:` prefix, no `:${type}:` segment
```

That is the exact shape you correctly repaired in `test/unit/scope-resolution/callable-value-target-index.test.ts:36` — the bench fixture was just missed. Every such id now fails the `startsWith(prefix)` check on this line, so `valueBindingPositionKey` (`callable-value-flow.ts:819`) and `definitionAnchorKey` (`:1009`) return `undefined` and `buildGraphTargetIndex` **`continue`s past the entry rather than erroring** — the loss is silent.

**Evidence.** Reproduced four times independently (two review lanes, a synthesis pass, and once more directly against the current head `a8535bdf`) via `node --import tsx bench/callable-value-flow/measure.mjs --check`:

| | `large.targets` | `large.callable_only_targets` | fingerprint |
|---|---|---|---|
| base | 4000 | 3200 | `70bebf6a…` (committed baseline) |
| **head** | **3200** | **3200** | **`cf0cb7eb…`** |
| head + fixed fixture | 4000 | 3200 | `6599dda7…` |

`targets == callable_only_targets` is the tell: the #2693 widened gate now admits **zero** value bindings. The bench says so itself — *"the resolved callable-value target set CHANGED. This is a behaviour change, not a perf one."* Red on CI as `tests / benchmarks (GITNEXUS_BENCH)` (job 90636078250 on this head; identically on the pre-merge head), wired at `.github/workflows/ci-tests.yml:500`.

**Suggested fix.** In `bench/callable-value-flow/measure.mjs:92`, mint the production shape:

```js
const nodeId = `def:${filePath}#${line}:0:${type}:${qualifiedName}`;
```

Then set the single top-level `fingerprint` in `baselines.json` to `6599dda7d0ee5942e1995a1dcfb137c312eb8e4690bc82a8bbff429f08bd839d` (verified twice; restores `targets` to 4000 = 3200 callables + 800 value bindings, matching base) and note in `_comment` that the drift is the def-id **shape in the synthetic corpus**, not the resolved target set — which is what `baselines.json:2` means by drift that "must be explained".

Please don't re-baseline to the head value `cf0cb7eb…` — that would pin the gate to a corpus admitting zero value bindings and silently retire the #2693 guard. Also worth re-checking `widening_overhead` afterwards: three post-fix runs measured 1.698 / 1.779 / 1.727, under the 1.9 budget but above the 1.43–1.58 band the note records (possibly just noise on a loaded machine).

gitnexus/src/core/ingestion/scope-resolution/utils/definition-id.ts
```diff
@@ -0,0 +1,24 @@
...
+  const match = /^(\d+):(\d+):/.exec(nodeId.slice(prefix.length));
```
magyargergo, line 21: **P3 (latent, not reachable today) — the anchor is a `startsWith` test, not a segment test.** `[reproduced]`

If the `filePath` argument is ever a *proper textual prefix* of the path embedded in `nodeId`, and the remainder happens to begin with `<digits>:<digits>:`, the prefix check passes against the **wrong file** and this line parses the path fragment as the declaration coordinates — the #2734 failure mode surviving at the boundary *between* two paths rather than inside one.

Executed against the shipped implementation:

```js
definitionIdPosition('def:src/bugs#5:3:test/helper.php#10:20:Function:bar', 'src/bugs')
// → { line: 5, column: 3 }   (true position is 10:20)
definitionIdPosition('def:src/a#12:34:x#1:0:Function:y', 'src/a')
// → { line: 12, column: 34 } (true position is 1:0)
```

**Not a live bug** — no current caller can produce a mismatched pair, and both engines independently failed to construct one. One review lane argued the case is impossible because a def id carries exactly one more `#` than its own path; that argument is right but only covers *matched* pairs, not the mismatched case above.

**Suggested fix (cheap, optional):** one more `it.each` row in `test/unit/scope-resolution/definition-id.test.ts` — `['def:src/a#12:34:x#1:0:Function:y', 'src/a']` expecting `undefined`, which fails today. The existing "mismatched" row uses a *disjoint* path, which already fails closed and isn't the risky shape. To close the property rather than document it, have the helper take the `SymbolDefinition` so the two strings cannot diverge.

gitnexus/src/core/ingestion/scope-resolution/graph-bridge/ids.ts
```diff
@@ -221,7 +218,7 @@ export function resolveDefGraphId(
     // without either side having to model the scope chain. Node ids are
     // 0-based, def ids 1-based. An `AMBIGUOUS_POSITION` tombstone (two
     // callables on one line) falls through to the name-based keys below.
-    const line = defStartLine(def.nodeId);
+    const line = defStartLine(def.nodeId, filePath);
```
magyargergo, line 221: **P3 (latent) — `undefined` now also means "wrong file", and that routes into the branch that *skips* the #2699 fail-closed guard.** `[code-read]`

This is the one call site of the six where the `filePath` argument is not read off the same `SymbolDefinition` as `nodeId` — it is `resolveDefGraphId`'s own parameter, and that function's structural `def` parameter type (`ids.ts:196-209`) has no `filePath` field at all, so the two strings can diverge with no compiler signal.

Pre-PR, `line === undefined` meant only "this id carries no coordinates". Post-PR it *also* means "this def id belongs to a different file". Both now skip the same block — and the #2699 anti-fabrication guard lives **inside** it (`ids.ts:236-238`), which exists to stop the label-agnostic, first-write-wins `simpleKey` at `ids.ts:322-323` from aliasing a function-local onto whichever same-named callable registered first. The failure direction is inverted relative to intent: a path *mismatch* is stronger evidence of a mis-join than a missing id, yet it takes the more permissive branch.

**Verified not live.** All ~27 `resolveDefGraphId` call sites pass `def.filePath` / `parsed.filePath` / `scope.filePath` with the def drawn from that same file, and `def.nodeId.startsWith('def:' + def.filePath + '#')` is a construction invariant: `makeDefId` (`scope-extractor.ts:864-871`) is the sole producer and writes `nodeId` and `filePath` from the same variable in one object literal (`scope-extractor.ts:623-624`). Both engines independently failed to construct a trigger, including for Vue SFCs (`file.path` is passed on both sides).

**Suggested fix — no behaviour change requested.** Either (a) add the missing test: there is currently no unit test for `resolveDefGraphId` or `defStartLine` at all, so a deliberately mismatched `(filePath, def.nodeId)` pair asserting the intended outcome would make a future divergence fail loudly instead of silently changing which fallback branch runs; or (b) require `filePath` on the `def` parameter and derive it internally, removing the divergence by construction. While here, the `defStartLine` docblock at `ids.ts:166-167` still describes the pre-PR single-argument contract.

magyargergo (commented):
# Multi-engine review — PR #2743 `fix(scope-resolution): parse def coordinates after file paths`

Base `bc76ba2f` → head `a8535bdf` · 6 files, +93/−21.

*The branch was merged with `main` while this review was running (`72eb50a4` → `a8535bdf`). That merge
touched none of the PR's six files and the three-dot diff is byte-identical, so every finding and line
anchor below was re-verified against the current head `a8535bdf` rather than the head the lanes read.*

**Verdict: request changes — one blocking item.** The fix itself is correct, minimal and well-targeted,
and the review closed out its largest suspected regression as refuted. The blocker is not in the fix:
the PR corrected a malformed def-id shape in the unit-test fixture but missed the identical shape in
the committed bench fixture, which is a CI correctness gate and is currently red. Branch hygiene is
clean (single linear commit, conventional title, no unrelated churn); merge state is `BEHIND` main but
with zero file overlap against the intervening commit, so no rebase is required for correctness.

## 1. How this was reviewed

| Engine | Leg | Status | Outcome |
|---|---|---|---|
| Claude | GitNexus PR swarm (5 lanes: facts-historian, branch-hygiene, risk-architect, security-boundary, test-CI-verifier) | live | 1 P2 coverage gap, 2 P3, rest clean |
| Claude | `ce-code-review` (6 personas: correctness, testing, maintainability, performance, project-standards, adversarial) | live | 1 P1 reproduced, several P3 |
| Codex gpt-5.5/xhigh | GitNexus swarm (Solo) | live | 1 finding (schema bump, P1-as-filed) |
| Codex gpt-5.5/xhigh | `ce-code-review` (report-only) | live | 1 P3 + residual risks |
| Codex gpt-5.5/xhigh | native adversarial-review | live | verdict `approve`, no material findings |

All three Codex legs produced real findings — none is an empty or "skill not found" marker — so
cross-engine agreement and disagreement are both meaningful signals here.

**One caveat on the Codex adversarial leg:** it could not execute `node` at all (bubblewrap namespace
restriction; `node --version` exited 1 in its log), so its "no material findings" is a *static-analysis*
result. That materially explains why no Codex leg found the headline defect below, which is only
visible by running the bench.

## 2. What the review validated

These are the load-bearing conclusions, not courtesy notes — each licenses a decision below.

- **The bug is real and the fix is the right shape.** On base, a def id under a path containing a
  `#<digits>:<digits>:` fragment (e.g. `def:src/bugs#12:34:test/a.php#4:2:Function:$handler`) parses as
  line 12 / col 34 — the *path* fragment — instead of 4/2. `scopeIsCallableBody` then rejects every
  closure under such a path and no CALLS edge is emitted. Anchoring on `def:${filePath}#` fixes exactly
  that and nothing else: for any path without such a fragment, old and new parse identically, including
  a bare `#` in the path (`src/a#b.ts`) and a `#` in the name (TS `#secret`).
- **The coordinator's PRIMARY HYPOTHESIS was REFUTED independently by both engines.** The hypothesis:
  the new helper is stricter, so if `resolveDefGraphId`'s `filePath` *parameter* (`ids.ts:221`, which is
  a parameter, not `def.filePath`) ever disagreed with the def's own path, `defStartLine` would return
  `undefined`, skip the whole position block — **including the #2699 fail-closed `localNameKey` guard at
  `ids.ts:236`** — and fall through to the label-agnostic `simpleKey`, fabricating a CALLS edge.
  - Claude (risk-architect, ce-correctness, ce-adversarial) enumerated all ~27 `resolveDefGraphId` call
    sites and closed the producer side: `makeDefId` (`scope-extractor.ts:864-871`) is the *only* producer
    of `def:` ids, called from the single `buildDefFromDeclarationMatch` (`scope-extractor.ts:623-624`)
    which writes `nodeId` and `filePath` from the **same `filePath` local in one object literal**. Nothing
    mutates `def.filePath`; the durable ParsedFile store interns by `nodeId` (which embeds the path), so
    a `filePath` can never be swapped under a `nodeId`. Vue SFCs pass `file.path` on both sides.
  - Codex reached the same conclusion twice, separately: swarm Lane 1 — "I did not find a real cross-file
    `resolveDefGraphId(filePath, def, ...)` trigger where `filePath` differs from `def.filePath`"; and the
    adversarial leg — "I could not construct a concrete production trigger for the filePath-parameter
    mismatch, path-normalization, non-`def:`-prefixed producer, suffix-boundary, or hot-loop perf leads."
  - **Structural reason it holds:** `def.nodeId.startsWith('def:' + def.filePath + '#')` is a construction
    invariant, not a convention — one producer, one call frame, one variable. Every caller passes either
    `def.filePath` directly or `parsed.filePath` with `def` drawn from that same `parsed.localDefs`.

  This is what lets the change be called **structurally safe** rather than merely untested.
- **The test-fixture rewrite in `callable-value-target-index.test.ts` is a correction, not a mask.**
  The old stub `` `${filePath}#${line}:0:${qualifiedName}` `` never matched production `makeDefId` output
  (no `def:` prefix, no `:<type>:` segment); it only passed because the old regex ignored everything
  before the `#`. Under the new helper it returns `undefined`, so those assertions **had** to change.
  Four lanes across both engines independently reached this. No action needed.
- **Performance is a non-issue.** ce-performance micro-benchmarked the delta (~169 ns/call in the
  realistic varying-`filePath` case) against an estimated ~2.3M calls on a kernel-scale repo ≈ **0.4 s**
  added, no new Big-O, allocations are young-gen and never retained. Explicitly reported as a *refuted*
  performance concern.
- **Security, hygiene, branch state are clean.** No bidi/RTL controls, `git diff --check` clean, no
  `any`, prettier clean, conventional-commit title, no CHANGELOG edit (correct — `DoD.md:138` forbids
  it). ReDoS, `\0`-key forgery, and temp-dir escape in the new fixture were each checked and refuted.
  Single linear commit; `BEHIND` main but zero file overlap with `bc76ba2f`.

## 3. Headline: the PR leaves a CI gate red — the bench fixture has the same defect the PR fixed in the unit test

`gitnexus/bench/callable-value-flow/measure.mjs:92` builds synthetic def ids the old way:

```js
const nodeId = `${filePath}#${line}:0:${qualifiedName}`;   // no `def:` prefix, no `:${type}:` segment
```

Under the new `definitionIdPosition` every one of those fails the `def:${filePath}#` anchor and returns
`undefined`. `valueBindingPositionKey` and `definitionAnchorKey` then return `undefined`, and
`buildGraphTargetIndex` **`continue`s past the entry rather than erroring** — so the widened gate admits
zero value bindings and the bench silently stops exercising the one code path it exists to guard (#2693:
a closure bound to a name declares as a value but emits a callable node).

This is **reproduced end-to-end**, four times independently — two Claude lanes (ce-performance,
ce-adversarial), again during synthesis, and once more directly by the reviewer before posting — and it
is red on this PR's own CI. The bench's own failure message states the consequence plainly:

> `fingerprint drift: cf0cb7eb…d4d6bf2a46 != 70bebf6a…6d223e51 — the resolved callable-value target set
> CHANGED. This is a behaviour change, not a perf one.`

| | `large.targets` | `large.callable_only_targets` | fingerprint |
|---|---|---|---|
| base (green) | 4000 | 3200 | `70bebf6a…6d223e51` (committed baseline) |
| **PR head (red)** | **3200** | **3200** | **`cf0cb7eb…d4d6bf2a46`** |
| head + fixed fixture | 4000 | 3200 | `6599dda7…f08bd839d` |

`targets == callable_only_targets` on head is the tell: the "widened" run is no longer wider than the
callable-only run.

**Cross-lane number check (asked for explicitly).** The *deterministic* values agree exactly across all
three runs — same fingerprints, same target counts. The only divergence is `widening_overhead`:
ce-adversarial 0.994, ce-performance 1.143, synthesis 1.106. That metric is a wall-clock ratio
(`large_ms / callable_only_ms`), so the spread is machine/run noise, not a contradiction between lanes.
All three sit far below the documented **1.43–1.58 band "with the positional join"** — which is itself
the signal that the join is no longer running.

**Does the fix require a re-baseline? Yes — and that is permitted.** The fingerprint is
`sha256` over `` `${defNodeId}\0${graphId}` `` pairs (`measure.mjs:177-180`), so correcting the fixture's
id shape necessarily changes the hash even though the resolved target set is restored. Verified during
synthesis by running a patched copy of the bench: `targets` returns to **4000** (3200 callables + 800
value bindings, matching base) with fingerprint **`6599dda7d0ee5942e1995a1dcfb137c312eb8e4690bc82a8bbff429f08bd839d`** —
byte-identical to the value ce-adversarial derived independently. ce-adversarial's other claim (that
re-adding the entries the *old* regex admitted reproduces the committed `70bebf6a…` baseline exactly)
is a *causation proof for the old id shape* and does not conflict with this — it establishes that base
was green and this PR is the sole cause.

The baseline note does **not** forbid re-baselining outright; it forbids doing it *unexplained*
(`baselines.json:2`, verbatim):

> `fingerprint` is an order-independent sha256 over every (defNodeId -> graphId) pair
> buildGraphTargetIndex resolves on the synthetic corpus; it is a CORRECTNESS gate, so drift means the
> callable-value target set moved and must be explained, **never re-baselined to make CI green**.

So the correct fix is: repair the fixture, set the fingerprint to `6599dda7…`, and record in `_comment`
that the drift is the **def-id shape in the synthetic corpus**, not the resolved target set — which is
restored to its base cardinality of 4000. `baselines.json` carries a single top-level `fingerprint`
field, so that is the only value to change.

Please also re-verify `widening_overhead` after the fix. Three independent post-fix runs measured
**1.698**, **1.779** and **1.727** — all inside the `widening_overhead_budget` of 1.9, but all above the
1.43–1.58 band that `_widening_overhead_note` records "with the positional join". That may be no more
than shared-runner noise (the metric is a wall-clock ratio, and these runs were on a loaded dev
machine, not a CI runner), but the margin to the budget is thinner than the recorded band implies, so
it deserves a confirming run rather than an assumption.

**Why this must be fixed in this PR, not deferred:** the tempting way to green CI is to re-baseline
against the head value `cf0cb7eb…`. That would permanently pin the gate to a corpus in which *zero*
value bindings are admitted, and nothing in CI would ever notice that the #2693 guard had stopped
guarding anything.

`measure.mjs` is not in this PR's diff, so the inline comment for this is anchored on the new helper —
the code that makes the fixture fail — with the real location named in the text.

## 4. Findings posted inline

| # | Anchor | Sev | Corroboration |
|---|---|---|---|
| 1 | `utils/definition-id.ts:20` (re: `bench/callable-value-flow/measure.mjs:92`) | **P1** | STRONG — reproduced ×4 + live CI red |
| 2 | `utils/definition-id.ts:21` | P3 latent | STRONG — both engines, 5 lanes |
| 3 | `graph-bridge/ids.ts:221` | P3 latent | STRONG — both engines, 5 lanes |

## 5. Lower-priority items, refuted suspicions, and test gaps

**Reported, not blocking:**

- **P2 → now largely moot — stale-index caveat.** Both engines independently established the
  behaviour: a repo containing a `#<d>:<d>:` path that was analyzed *before* this fix keeps the
  mis-attributed edges on unchanged files until a full re-analyze, because the incremental reuse gate
  still passes. Codex filed this as a **P1 requiring `INCREMENTAL_SCHEMA_VERSION` 23 → 24**; Claude filed
  it as no-bump-needed. **Ruling: no bump — and the merge from `main` has since settled it in practice.**
  `INCREMENTAL_SCHEMA_VERSION` is **already 24** on the current head (`repo-manager.ts:568`), bumped by
  `main` for unrelated reasons (#2708) and pulled in by the merge — not by this PR. Anyone upgrading
  across that bump gets the forced one-time full rebuild anyway, which re-analyzes affected files and
  picks up this fix. So the residual exposure is now only users who had *already* crossed the 24 bump
  before this PR lands. The reasoning for declining a bump on this PR's own merits still stands: `SCHEMA_BUMP` is scoped to `ParsedFile` serialization shape
  (`parse-cache.ts:56-57`) and nothing parse-time changed. `INCREMENTAL_SCHEMA_VERSION`'s precedents
  (v2, v3) are *missing columns* — indexes that structurally cannot serve a query; this index is intact
  and merely holds wrong edges for a rare path shape, and the repo itself declines bumps at this blast
  radius ("a bump would force a one-time full rebuild for every user", `repo-manager.ts:323`). The two
  engines agree on the *fact* and differ only on the remedy. Please state the caveat in the PR body
  instead, per `DoD.md:96` (residual risks and compatibility impacts must be "clearly stated").
- **P3 — `Number.isFinite` guard dropped.** Old `defStartLine` returned `undefined` for a non-finite
  line; the new helper returns `Number(match[1])` unguarded (`definition-id.ts:23`). Reaching `Infinity`
  needs a ~309-digit line number, so ce-adversarial and ce-correctness both call the guard genuinely dead
  and the removal a net *tightening* (pre-PR the digits could come from the user-controlled path). Noted
  rather than posted inline, because two lanes explicitly refuted it as reachable. Restore it if you want
  the defensive symmetry — it is one line.
- **P3 — `defStartLine`'s docblock is now stale** (`ids.ts:166-167`): it describes the pre-PR
  single-argument, shape-only contract and mentions neither the new `filePath` parameter nor the fact
  that `undefined` now also means "this id belongs to a different file".
- **P3 — `defStartLine` is now a single-caller pass-through** (`ids.ts:169-171`, sole call site
  `ids.ts:221`). Its regex and `isFinite` guard both moved into `definitionIdPosition`, so it forwards
  two arguments and projects one field. Inlining it would delete a function and a stale docblock.
- **P3 — `DefinitionIdPosition` is exported with zero external consumers** (`definition-id.ts:1-4`;
  grep across `src/` and `test/` finds it only as this function's own return type).

**Refuted — raised by a lane and dropped after verification:**

- The primary hypothesis (fail-closed bypass via a `filePath` mismatch) — see §2. Refuted by both engines.
- **"The helper names a language, violating `AGENTS.md:42`"** (Codex ce, P3, confidence 100, re: the
  "TypeScript private members" comment). Dropped. That rule's own clarifying clause is about *behaviour*
  — "must not name languages — **plug language behavior in via `LanguageProvider` / `ScopeResolver`
  hooks**" — i.e. it forbids language *branching* in shared code. **352** files under
  `src/core/ingestion/` carry language-naming comments, including `ids.ts` — a file this PR edits — at
  lines 13, 127, 235, 339. The new helper contains zero language branching. Posting this would invent a
  standard the repo does not hold.
- Windows-only degradation hidden by the new `describe.skip` — refuted. The fix compares `nodeId`
  against the *same string* that built it, so backslashes, drive letters and UNC paths match trivially;
  and the #2734 shape is *unreachable* on Windows because `:` is illegal in an NTFS filename. The skip
  hides a fixture, not a platform-specific failure.
- Line/column swap at any of the six rewritten sites — checked site by site, none.
- String→number drift in key construction (`match[1]` → `position.line`) — refuted; the write side of
  all three maps was already numeric, and `\d+` from a real range never carries leading zeros.
- ReDoS, `\0`-separator key forgery, temp-dir escape in the new fixture — each refuted with a reason.
- Performance regression at the six rewritten call sites — measured, then refuted (§2).
- `SCHEMA_BUMP` required — refuted by both engines.

**Test gaps (no action demanded, listed because several lanes converged on them):**

1. The producer-side invariant is asserted nowhere. One line in `scope-extractor.test.ts` —
   `localDefs.every(d => d.nodeId.startsWith(\`def:${d.filePath}#\`))` — would convert "audited once in
   review" into "enforced", and would have caught the bench fixture.
2. Of the six rewritten call sites, only `scopeIsCallableBody` and `defStartLine` get #2734-shaped
   end-to-end coverage. `definitionPosition`, `definitionAnchorKey` and `definitionStartsAt` have no test
   that would fail if the new anchoring returned `undefined` where the old regex returned coordinates.
3. No test asserts the positional join still *admits* a closure binding — that assertion lived only in
   the bench fingerprint, the artifact this PR broke. A unit case asserting a non-empty admitted set in
   `callable-value-target-index.test.ts` would make the guard survive independently of the bench.
4. `definition-id.test.ts`'s `it.each` covers a *disjoint* path mismatch but not a *prefix* one — the one
   shape that returns a wrong value instead of `undefined` (finding #2).
5. The one end-to-end proof is doubly gated (`distWorkerExists()` **and** `platform !== 'win32'`), and
   `closure-review-findings.test.ts` is not in `scripts/cross-platform-tests.ts`, so the macOS matrix —
   the only non-Linux runner where a `:` filename is legal — never executes it. Both gates are
   individually justified; the combined effect is that the fix is Ubuntu-CI-only end to end.

## 6. CI status

Red on the current head `a8535bdf` (re-checked after the merge; it was red on `72eb50a4` too):

- `tests / benchmarks (GITNEXUS_BENCH)` — **fail**, 3m37s, step *"Callable-value-flow target-index
  guards (#2693)"*, wired at `.github/workflows/ci-tests.yml:500`. This is the headline finding and it
  is a genuine code defect. It failed identically on the pre-merge head (2m58s, job 90623159779), so
  the merge neither caused nor fixed it.
- `tests / windows-latest (platform-sensitive) 2/3` — **fail**, 10m2s (job 90636108521). This one
  appeared only after the merge and looks **unrelated to this PR**: the failing suite is
  `test/integration/cli-limit-e2e.test.ts` with `Error: Analyze failed (status null)` — a null status
  means the analyze child process was killed rather than exiting with an error, the signature of the
  known Windows CLI-e2e timeout flake. It exercises the `--limit` CLI path, which this PR does not
  touch, and the same shard's siblings 1/3 and 3/3 pass. Worth a re-run to confirm, but I would not
  attribute it to these changes.
- `Vercel` — **fail**, "Authorization required to deploy". Fork-PR permissions artifact, not a code defect.
- Everything else passes, including all three `tests / ubuntu / coverage` shards (so the new
  integration case did execute and pass), `quality / typecheck`, `quality / lint`, `quality / format`,
  tree-sitter ABI on all three OSes, node-floor compat, and CodeQL.

## 7. Coverage limits of this review

- No leg built the project or ran the full test suite; `dist/` is absent in the review worktree, so the
  new integration test self-skipped everywhere it was attempted and was verified only by reading its
  wiring plus confirming CI builds before testing.
- The Codex adversarial leg could not run `node` at all (sandbox restriction) — static analysis only.
- Live GitHub state (reviews, mergeability) was not visible to the Codex legs; it was checked by the
  coordinator and during synthesis.
- The GitNexus MCP index consulted by the Codex swarm leg was stale and used only as advisory navigation.
- No leg analyzed a real repository containing a `#<d>:<d>:` path before and after the change; the
  end-to-end evidence is the PR's own single-file PHP fixture.

---

*This is an automated multi-tool review digest (Claude swarm + Claude ce-personas + Codex gpt-5.5 ×3).
Every finding below carries the trigger and the file:line evidence it was verified against — please
confirm each against your own reading before acting on it.*

gitnexus/src/core/ingestion/scope-resolution/utils/definition-id.ts
azizur100389: Fixed in 6bf08d2c. The benchmark now emits production-shaped def IDs and restores 4,000 targets (3,200 callable-only + 800 value bindings). The corrected fingerprint is documented; four local --check runs passed, with widening overhead below the 1.9 budget.

gitnexus/src/core/ingestion/scope-resolution/utils/definition-id.ts
azizur100389: Reviewed. This requires a mismatched nodeId/filePath pair that current producers and callers cannot construct, so I have kept the production behavior unchanged in this focused fix.

gitnexus/src/core/ingestion/scope-resolution/graph-bridge/ids.ts
azizur100389: Reviewed all resolveDefGraphId call sites and confirmed they pass the definition's own file path. I kept this latent mismatch path unchanged; the PR body now records the actual compatibility caveat for pre-fix schema-v24 indexes.

gitnexus/src/core/ingestion/scope-resolution/utils/definition-id.ts
magyargergo: Can you please create a follow-up task? 🙏

gitnexus/src/core/ingestion/scope-resolution/graph-bridge/ids.ts
magyargergo: Could we require the `filePath` please on the `def`? This will ensure that any new caller must pass this parameter down.

Issue #2736, opened 2026-07-28
Test gaps left by #2718: Block-safeguard negative case, SCHEMA_BUMP pin, node-vs-edge assertions


magyargergo:
## Summary

Three coverage gaps identified by the multi-engine review of #2718. None is a defect today; each is a guard that does not exist for behaviour the PR deliberately relies on.

## 1. The Block-scope safeguard's NEGATIVE case is untested

`pickCallerCallableDef` accepts a `Block`-kind scope as a callable boundary only when the scope **is** that callable's body (def start position == scope start position). That alignment check is the sole thing preventing a nested named function inside a block from swallowing a call made at **block** level, outside it.

The positive case (a Kotlin/Ruby closure) is covered. The false branch is not. A regression that relaxed the check would produce misattributed edges with no test failing.

Suggested: a fixture with a named function declared inside a block, plus a call at block level, asserting the call attributes to the enclosing callable and **not** to the nested function.

## 2. `SCHEMA_BUMP` has no pinned-value test

`INCREMENTAL_SCHEMA_VERSION` is pinned in `test/unit/call-summary-schema-version.test.ts`, which is what makes a missed bump fail loudly. Its sibling `SCHEMA_BUMP` (parse-cache) has no equivalent.

This is not hypothetical: `parse-cache.ts`'s own v21 comment records a prior collision where two PRs both bumped to the same value and the second shipped with **no invalidation**. #2718 hit the identical class again — it branched at 27, bumped to 28, and #2415 merged 28 first; it was caught only by re-checking against `origin/main` at merge time, and corrected to 29.

Suggested: pin `SCHEMA_BUMP` the same way, and consider a check that it exceeds the value on the merge base.

## 3. `function-local-identity.test.ts` asserts node ids, never edges

The stated reason the position-qualified label predicate must be shared across all three phases (id-building, resolution, registration) is that a mismatch **silently drops edges** — the caller attaches to a node that does not exist. That is the #2714 failure mode.

The test asserts only that two distinct nodes exist. Two distinct nodes with every edge attached to the wrong one would pass it.

Suggested: assert the CALLS/ACCESSES edges as well as the node ids, so the failure mode the design guards against is actually covered.

## Also open (lower priority)

- Dart's later declarator (`var f = …, g = …`) is handled by the query rule added in #2718 but has no test.
- `Static` / `isLinkableLabel` asymmetry noted by the review, not investigated.

azizur100389:
[...]

Issue #1982, opened 2026-06-03
Qualified nested-type node identity — resolution-side follow-up (Ruby same-tail routed-props/mixins + Rust inherent impls)


magyargergo:
Follow-up to #1978 (PR #1981). The first increment landed **structure-phase** fully-qualified nested-type node identity for **C++ and Ruby** (gated by `qualifiedNodeId`). Two items were deliberately deferred because they are **resolution-side** and/or need separate machinery. This ticket tracks them.

---

## 1. Ruby — same-tail routed-property / mixin owner identity (registry-primary)

**Symptom.** With `qualifiedNodeId: true` for Ruby, an `attr_accessor` (or `include`/`extend` mixin) declared in one of two same-tail nested classes is attributed to the **wrong** sibling under the registry-primary leg, and the worker path emits a **duplicate** cross-wired edge.

```ruby
module Outer; class Inner; attr_accessor :x; end; end
module Other; class Inner; end; end
# HAS_PROPERTY x  ->  Other.Inner  (should be Outer.Inner)
```

**Root cause.** `gitnexus/src/core/ingestion/languages/ruby/scope-resolver.ts` `emitRubyMixinEdges` builds `graphIdByName` keyed by the **simple tail name with last-wins**:

```ts
const simpleName = def.qualifiedName?.split('.').pop() ?? def.qualifiedName ?? '';
graphIdByName.set(simpleName, graphId);   // two `Inner` -> one key, second wins
```

and the `__property__:` / `__heritage__:` emission resolves the owner via `graphIdByName.get(className)`, where `className` comes from `ruby/captures.ts` (~L182) as the **simple** `enclosing?.childForFieldName('name')?.text`. Pre-#1978 the two classes merged into one node so the simple-name lookup was internally consistent; splitting them turns it into an active mis-attribution.

Additional interaction: on the worker path the structure phase emits the **correct** `Outer.Inner -> prop` while `emitRubyMixinEdges` emits the **wrong** `Other.Inner -> prop` (different source ids → the existing dedup misses), leaving two HAS_PROPERTY edges.

[...]

Issue #1975, opened 2026-06-02
fix: scoped class/module/impl declarations don't materialize a graph node (dangling HAS_METHOD edges) — PR #1972 P2 follow-up


magyargergo:
**Follow-up from PR #1972 (the P2 finding).** Splitting this out as agreed in https://github.com/abhigyanpatwari/GitNexus/pull/1972#issuecomment-4601102741 so it isn't lost.

## Problem

A class/module/struct/impl declaration whose **name is a scoped node** (Ruby `class Foo::Bar`, C++ out-of-line `class Outer::Inner {}`, Rust `impl Outer::Inner`) never becomes a `Class`/`Trait` graph node. Its methods are still captured and get `HAS_METHOD` edges, but those edges point at an owner id that was never created — a **dangling edge**.

Reproduced via `runPipelineFromRepo` on PR #1972's `ruby-namespaced` fixture, on **both** resolver legs, byte-identical at the merge base (pre-existing, not a regression):

- `class Foo::Bar` → no `Class` node; bare `class Plain` (same file) → node present.
- Method created as `Method:file.rb:Foo::Bar.bar_method` (owner-qualified by the full scope text).
- `HAS_METHOD` source id `Class:file.rb:Foo::Bar` resolves to **MISSING**.

Downstream consumers that assume every edge endpoint resolves (community detection, process extraction, impact-analysis blast radius) silently degrade for namespaced code.

## Root cause (the non-obvious bit)

Graph nodes are materialized by the **legacy parsing path** (`tree-sitter-queries.ts` structure query → `parsing-processor.ts`), **not** the scope-extractor. PR #1972 widened the scope-extractor *capture* (`languages/ruby/query.ts`), which feeds the registry, so the node still wasn't created. The real gate is the legacy structure query's `name: (constant)` arm (won't match a `scope_resolution` name), plus a node-id ↔ `HAS_METHOD` owner-id key mismatch in `utils/ast-helpers.ts` (the owner walk reads the full `Foo::Bar` text). The scope-extractor's node-bearing path is dead pending the deferred node-creation migration (#1876).

## Scope

Languages that actually have the gap and need a fix:
- **Ruby** (class/module) — confirmed
- **C++** (out-of-line qualified class/struct defs + out-of-line method qualifier loss) — confirmed
- **Rust** (scoped `impl` targets) — suspected; confirm-first

[...]

PR #2745, opened 2026-07-29
fix(rust): qualify items by their enclosing mod chain so same-named ones stay distinct (#2742)

### LINKED ISSUE #2742: Rust: same-named items at different module depths in one file collapse to a single graph node
Split out of #2730 / PR #2741, where it was found but deliberately not fixed.

## Symptom

An inline `mod` member and a crate-root item of the same name, in the same file, become one graph node:

```rust
// src/main.rs
mod inner {
    pub fn dispatch() -> usize { 1 }
}

fn dispatch() -> usize {
    inner::dispatch()      // renders as a self-loop
}
```

Both `dispatch` definitions map to the node id `Function:src/main.rs:dispatch`, so the call above shows as an edge from the node to itself. Resolution picks the correct definition — the target simply is not representable.

## Cause

Node identity is `<label>:<filePath>:<qualifiedName>`, and the simple-name fallback key in `scope-resolution/graph-bridge/node-lookup.ts` is `${filePath}::${name}`, first-wins within a file. Neither carries the module path.

`namespacePrefix` does record the enclosing module, but it is a deliberate sidecar: its own doc states it "does NOT participate in graph node identity". `Namespace` is also absent from `isLinkableLabel`, so Rust modules exist in the resolution model without a graph node of their own — matching how C++ namespaces are handled.

## Why it was not fixed in #2741

Making node ids module-path-qualified changes ids for essentially every Rust repository, because `#[cfg(test)] mod tests` is close to universal. That is an incremental-schema migration with its own blast radius, and bundling it into a resolver fix would have made that PR unreviewable.

PR #2741 fixes the resolution side and states this limitation explicitly. It also adds the fixture that exercises the shape — `test/fixtures/lang-resolution/rust-2730-gaps` contains the inline `mod inner` case — so a future fix has something to assert against.

## Prior art

- #1978 documented the same tail-keyed collision for C++/Rust. PR #1981 fixed C++ and Ruby; the Rust side was explicitly deferred there for the same migration reason.
- The `INCREMENTAL_SCHEMA_VERSION` log in `storage/repo-manager.ts` shows the established handling for id-changing work: v13 (Java local classes, JLS 13.1 identities) and v18/v21 (function-local callables gaining enclosing-callable + position identity) are the closest precedents.

## Scope

- Give Rust items a module-path-qualified identity, so `inner::dispatch` and a crate-root `dispatch` are distinct nodes.
- Decide whether Rust `mod` should become a linkable `Namespace` node, or stay resolution-only as it is for C++.
- Bump `INCREMENTAL_SCHEMA_VERSION` with a note, since ids change for unchanged files.
- Extend `rust-2730-gaps` to assert the inline-`mod` call resolves to the inline member rather than the crate-root one.

magyargergo:
Fixes #2742. Follow-up to #2741, which found this and deliberately did not fix it.

## The bug

Node identity is `<label>:<file>:<qualifiedName>` and carried no module path, so an inline module member and a crate-root item of the same name collapsed onto one node, first-wins:

```rust
// src/main.rs
mod inner {
    pub fn dispatch() -> usize { 1 }
}

fn dispatch() -> usize {
    inner::dispatch()      // rendered as a self-loop
}
```

Resolution already picked the right definition. The target simply was not representable, so a correct resolution still emitted a self-loop and `impact` reported the real callee as unreached.

```
before   NODE  Function:src/main.rs:dispatch            <- one node for both
         CALLS Function:src/main.rs:dispatch -> itself

after    NODE  Function:src/main.rs:dispatch
         NODE  Function:src/main.rs:inner.dispatch
         CALLS Function:src/main.rs:dispatch -> Function:src/main.rs:inner.dispatch
```

## Approach

The mechanism already existed. `qualifyRustImplTargetByModScope` has walked `mod_item` ancestors for impl targets since #1982; this generalises it to `qualifyByEnclosingModScope` and applies it to free items. It keys purely on the `mod_item` node type, exactly as the impl qualifier already did, so it is a no-op for every language whose grammar has no such node. There is one id-construction site — the sequential path was removed in #1741.

Four commits, each a distinct mechanism:

| Commit | What |
|---|---|
| `a349e5f6` | let the qualified-call filter see inline modules |
| `339cde9a` | try the namespace-prefixed node key before the bare one |
| `2603bfd4` | qualify items by their enclosing `mod` chain (the fix) |
| `0062a5c2` | bump `INCREMENTAL_SCHEMA_VERSION` and `SCHEMA_BUMP` |
| `7d546e10` | stop mod-qualifying container ids while their owner edges stay bare |
| `43e3bc03` | resolve modules nested inside an inline `mod` |
| `0e571edd` | let an imported type outrank a same-named module |
| `d5169756` | try the namespace-prefixed name on the TAGGED lookup keys too |
| `509916a7` | keep a `mod` declared inside a `fn` from hoisting above the callable |
| `2739569d` | move main's `SCHEMA_BUMP` pin to 32 |
| `664c7ebc` | merge `main` (#2743, which also touches `ids.ts`) |
| `46edb68b` | rebaseline the Rust capture fingerprint for the new fixtures |

The last five answer the tri-review on this PR (three P1s, two P2s, and the P3s), one finding per commit. Details in **Review fixes** below.

### A hole this exposed in #2741

The `couldNameAModule` negative filter added in #2741 builds its name set from FILE PATHS, so an inline `mod x { … }` — which appears in no path — was absent from it, and every module-qualified call into an inline module was rejected before any candidate channel ran. That is a hole in the optimisation rather than in the logic it guards, and it is why the first attempt at this fix still self-looped. The per-pass index now unions the file-derived names with inline module names taken from the scope model, collected in the walk that already builds the module to scope map.

### Two constraints found by tests rather than by reading

**The helper normalised `::` to `.` unconditionally.** With no enclosing `mod` that rewrote a top-level `impl a::Inner` from `a::Inner` to `a.Inner`, moving its node id away from the one the `HAS_METHOD` owner edge emits and breaking the #1975 scoped-impl ownership. It now returns raw text untouched when there are no mod segments, which also makes the change strictly additive for every id with no enclosing module.

**Qualification is scoped to items with no enclosing class/impl.** A method already carries its owner's name, and that owner's id is mod-scoped by the impl qualifier, so qualifying the method again breaks the same byte-for-byte agreement.

## Review fixes

The tri-review found that excluding the MEMBER side of an owner edge was only half the rule, plus four narrower issues. All fixed here.

**P1 — container ids dangled their own member edges** (`7d546e10`). A `struct` / `trait` / `enum` / `impl` declared directly inside a `mod` has no enclosing class, so the member-side guard let it through while its owner-edge anchor — minted separately by `findEnclosingClassInfo` from the container's bare name — stayed put. Five lines of idiomatic Rust dangled:

```rust
pub mod engine { pub struct Config { pub retries: usize } }

NODE     Struct:src/lib.rs:engine.Config
DANGLING HAS_PROPERTY Struct:src/lib.rs:Config -> Property:src/lib.rs:Config.retries
```

A trait impl inside a `mod` also dropped `METHOD_IMPLEMENTS` entirely. The gate now excludes both sides, keyed on `MEMBER_OWNER_NODE_TYPES` — derived from `CLASS_CONTAINER_TYPES`, already the single source of "this node type owns member edges", so a language adding a container cannot gain a mismatched id shape without failing that set's existing INVARIANT.

**P1 — the #1975 contract was re-broken one `mod` deeper** (same commit). The no-mod-segments guard above protects a scoped `impl a::Inner` only at the crate root. Inside a `mod` it does not fire, so `splitQualifiedName` normalised the separator and the mod prefix was applied, minting `Impl:<file>:outer.a.Inner` against an anchor still reading `Impl:<file>:a::Inner`. The impl-target branch deliberately fires only for an UNSCOPED `type_identifier`; the new gate had no such restriction and was picking up exactly the targets that branch had excluded.

**P1 — no test could see any of it** (same commit). Every dangling assertion in `rust.test.ts` is narrowed to `findDanglingEdges(result, ['HAS_METHOD'])`, and a repo-wide grep finds no unfiltered call site at all, which is why a dangling `HAS_PROPERTY` shipped green. The new tests call it unfiltered, and assert the NODE id rather than only the edge's anchor — the anchor was already bare while the bug was live, so an edge-only assertion passes in both builds. All four fail when the gate clause alone is reverted.

**P2 — the #2730 self-loop survived one level deeper** (`43e3bc03`). `mod outer { mod tools { … } fn dispatch() { tools::dispatch() } }` still bound to the enclosing `dispatch`. Both the name filter and `declaresSubmodule` looked modules up through `moduleScopeByFile`, which holds each file's ROOT scope only, so a nested inline module was invisible. Inline module paths now come from the members' `namespacePrefix`: a `mod` def carries no nesting information of its own, and a `Namespace` scope owns its OWN def rather than its children's, so neither of the obvious channels could answer it. Restricted to non-empty prefixes so this stays a declaration check and an undeclared file on disk still cannot outrank a `use` binding.

**P2 — a fabricated cross-file edge** (`0e571edd`). Widening the name filter let a type-qualified call through when a module shared the type's name, and the crate-root-relative candidate captured it: `use crate::c::Buffer; Buffer::with_capacity()` bound to a crate-root `mod Buffer` the source never names, where the base emitted no edge. Per the doctrine quoted in `ids.ts`, the missing edge was the correct answer. That candidate is now skipped when the head names anything non-module in the caller's own module — read via `lookupBindingsAt`, since a `use` binding is finalize output and absent from the scope's local table.

**P3s** (`d5169756`, `509916a7`). Five tagged lookup keys still composed from the bare name, so they were dead for any namespace- or mod-qualified def; the C++ `EXTENDS` retarget this PR produces was unasserted; a `mod` inside a `fn` composed its segment outside the callable prefix and inverted the real nesting; `union_item` kept an id shape inconsistent with the `struct` beside it; and two comments plus a docblock had gone stale. Unit tests added for `qualifyByEnclosingModScope` and `couldNameAModule`.

## Known limitation

Containers and their members still collapse across sibling modules — a `Struct`, `Trait`, `Enum` or `Impl` inside a `mod`, and any method or field it owns, keeps its bare id. This is unchanged from before this PR, not a new gap: moving those ids requires moving `findEnclosingClassInfo`'s anchor with them, which is a cross-language change (it would need `classExtractor.qualifiedNodeId` for Rust and the C++/C# paths re-verified). Free items — the `fn` / `const` / `static` case #2730 and #2742 are about — are fixed.

## Schema versions

`INCREMENTAL_SCHEMA_VERSION` 24 -> 25, `SCHEMA_BUMP` 31 -> 32. Node ids change for every Rust item inside any `mod` block, and `#[cfg(test)] mod tests` makes that close to every Rust repository, so a pre-v25 index holds ids an incremental top-up cannot reconcile and the reuse gate must force a full re-analyze. The qualified name is computed in the parse worker, so a warm parse cache would otherwise replay the old unqualified ids.

This branch originally claimed v24; #2708 took that number and merged first, so it is renumbered here. That is exactly the collision the v29 note in `parse-cache.ts` warns about, and re-checking against `origin/main` at rebase time rather than at branch time is what caught it.

It then happened a second time, in the other direction: #2736 landed a test on main pinning `SCHEMA_BUMP` to 31, and it arrived here through the merge of main while this branch already held 32. The pin and the constant live in different files, so the merge was textually clean and the test failed instead — which is the pin working as designed. Moved to 32 with the reason recorded inline. `INCREMENTAL_SCHEMA_VERSION` needs no second bump: 25 was introduced by this unmerged branch, so no released index carries it.

## Testing

Rebased onto current `main` (includes #2708, which touches resolution) and re-verified there.

Three new fixtures and fourteen new regression tests across the original fix and the review series: the distinct node exists, `inner::dispatch()` resolves to the inline member, no self-loop remains, no edge of ANY type dangles for containers inside a `mod`, a scoped inherent impl keeps its raw path, a trait impl inside a `mod` keeps `METHOD_IMPLEMENTS`, nested inline modules resolve at depth 2 and 3, an imported type outranks a same-named module, a `mod` inside a `fn` does not hoist, and the C++ `EXTENDS` deep-namespace target is pinned. Plus unit tests for `qualifyByEnclosingModScope` and `couldNameAModule`, including the branches no fixture reaches.

A new `rust-*` fixture is a call site of TWO corpora: the `rust-captures-golden` snapshot and `bench/scope-capture`'s fingerprint. I regenerated the golden and missed the bench, which is what turned CI red on the first push — the same class as #2743's headline finding, from the other direction. Rebaselined in `46edb68b`, and verified as pure corpus growth rather than a capture-logic shift: removing only the three new fixture directories reproduces the prior fingerprint exactly (196 fixtures), restoring them gives the new one (202). Scaling 1.057 in CI against a 1.5 budget. The reason and the two-corpora warning are recorded in the baseline entry itself for the next author.

Full suite on the review-fix series: **15390 passed, 70 skipped, 43 failed across 13 files** — every failure attributed rather than assumed.

Four of the files (`vendored-grammars`, `literal-collectors`, and both `cfg/`) pass in isolation, so they are the documented full-suite load-flake class. The remaining nine are all spawned-CLI e2e (`cli-limit-e2e`, `skip-git-cli`, `fts-extension-e2e`, `cli-e2e`, `analyze-heap-oom-e2e`, `mcp/server-startup`, `group-cli`, `cli-index-help`, `tool-no-index-stderr`) and fail as 30-55 s timeouts rather than wrong output. Running the CLI by hand names the cause:

```
Replacement worker did not report ready within 5000ms — likely crashed during
top-of-script init (slow host? raise GITNEXUS_WORKER_READY_TIMEOUT_MS ...)
```

The identical command with `GITNEXUS_WORKER_READY_TIMEOUT_MS=60000` exits 0 and indexes the repo, so this is the worker-readiness gate at its 5 s default on a slow sandbox host, not a behaviour change here. The in-process resolver suites raise that timeout, which is why they are unaffected.

Then merged `main` again for #2743, which lands in `ids.ts` — the same function refactored here — and re-verified on the merge result: **2168 passed, 0 failed** across 93 files (Rust / C++ / C# resolvers, every `scope-resolution` unit suite including #2743's new `definition-id` tests, `closure-review-findings`, `callable-id-lockstep`, and the parse-cache schema pins). #2743's new `defStartLine(nodeId, filePath)` signature and the `lookupTagged` helper here compose without conflict.


gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -2417,6 +2418,34 @@ const processFileGroup = (
...
+      // node id that #1975/#1982 established. Same-named methods on same-named
+      // types in sibling modules therefore still collapse; that is a narrower
+      // residual than the free-item case this fixes, and it is the owner edge's
+      // problem to solve rather than this one's.
+      const qualifiedName =
+        rustImplQualifiedName === undefined &&
+        definitionNode !== undefined &&
+        !enclosingClassInfo &&
+        objectLiteralOwnerInfo?.ownerName === undefined
+          ? qualifyByEnclosingModScope(definitionNode, qualifiedNameBeforeModScope)
```
magyargergo, line 2476: **P1** — This gate excludes members via `!enclosingClassInfo`, but it *includes* the containers themselves — `struct_item`, `trait_item`, `enum_item` and `impl_item` declared directly inside a `mod` all satisfy it. Their owner-edge anchor is computed somewhere else entirely and does not move with them: `findEnclosingClassInfo` returns `classId: generateId(label, "<file>:<classIdName>")` (ast-helpers.ts:1035) with the bare tail, and `qualifiedClassId` only fires when `provider.classExtractor?.qualifiedNodeId === true`, which `class-extractors/configs/rust.ts` never sets. Result on five lines of idiomatic Rust — `pub mod engine { pub struct Config { pub retries: usize } }` — is node `Struct:src/lib.rs:engine.Config` but edge `HAS_PROPERTY Struct:src/lib.rs:Config -> Property:src/lib.rs:Config.retries`, whose source does not exist; reproduced against the real pipeline at this sha. The same shape with an `impl` inside a `mod` dangles on `HAS_METHOD` and drops `METHOD_IMPLEMENTS` entirely. It already fires on this repo's own fixtures (`rust-generic-impl-same-method-name`, `rust-nested-tail-collision-generic`), and the dangling rows are silently discarded by the `IGNORE_ERRORS=true` COPY retry, so a struct inside a `mod` quietly loses all its fields. Narrowest fix: restrict the new qualification to items that own no members (free `fn`/`const`/`static`/`macro`/`mod`) — container qualification buys nothing for #2742's actual goal, which is function representability. Otherwise, move the mint and the anchor to one source.

gitnexus/src/core/ingestion/utils/ast-helpers.ts
```diff
@@ -41,9 +38,22 @@ export const qualifyRustImplTargetByModScope = (
...
     current = current.parent;
   }
-  return [...modSegments, ...splitQualifiedName(rawTargetText)].filter(Boolean).join('.');
+  // No enclosing `mod`: return the raw text UNTOUCHED. Normalizing here would
+  // rewrite a scoped target's separator (`a::Inner` -> `a.Inner`) and silently
+  // move its node id away from the one the owner edge emits, which is how this
+  // helper first broke the #1975 scoped-impl ownership when it was generalized
+  // beyond unscoped targets. Callers that pass an unscoped name are unaffected,
+  // since splitting one has always been the identity.
+  if (modSegments.length === 0) return rawText;
```
magyargergo, line 62: **P1** — This early return removes the #1975 hazard the comment above it describes only when `modSegments.length === 0`. Inside a `mod` the return does not fire, `splitQualifiedName` normalises `a::Inner` → `a.Inner`, and the mod prefix is applied — while `findEnclosingClassInfo`'s scoped arm still emits `Impl:…:a::Inner`. So `pub mod outer { impl a::Inner { pub fn helper(&self) {} } }` mints `Impl:src/lib.rs:outer.a.Inner` and leaves `HAS_METHOD Impl:src/lib.rs:a::Inner -> …helper#0` dangling — reproduced against the real pipeline at this sha, with the crate-root twin (`impl b::Inner` outside any `mod`) byte-identical on both legs as a control. The comment warns about exactly this failure; the code only avoids it in the zero-`mod` case, which is the one case this PR newly reaches past. Suggest not firing the new general caller for `nodeLabel === 'Impl'` at all (the impl branch above already owns that decision), or teaching the scoped arm at ast-helpers.ts:966 the same mod walk. `rust-scoped-impl` only covers the crate-root form, so nothing catches this.

gitnexus/test/integration/resolvers/rust.test.ts
```diff
@@ -2788,3 +2788,45 @@ describe('Rust type-qualified calls are not treated as module paths (#2730 revie
...
+// #2742 — same-named items at different module depths are distinct nodes.
+//
+// Node identity was `<label>:<file>:<qualifiedName>` with no module path, so an
+// inline `mod inner { fn dispatch }` and a crate-root `fn dispatch` in the same
+// file collapsed onto one node, first-wins. Resolution already picked the right
+// definition; the target simply was not representable, so a correct resolution
+// still rendered as a self-loop and `impact` reported the real callee unreached.
+// ---------------------------------------------------------------------------
+
+describe('Rust items are qualified by their enclosing mod chain (#2742)', () => {
```
magyargergo, line 2802: **P1** — This new describe block is the right place for the invariant that would have caught the owner-edge problem above. Every dangling assertion in this file is narrowed to `findDanglingEdges(result, ['HAS_METHOD'])` (lines 2061, 2111, 2162, 2202, 2250, 2279), and a repo-wide grep finds **zero** uses of the unfiltered one-argument form anywhere in `test/` — which is precisely why the two new `HAS_PROPERTY` dangles this PR introduces on `rust-generic-impl-same-method-name` and `rust-nested-tail-collision-generic` ship with a fully green CI. Please add a fixture with `mod m { pub struct S { pub f: usize } pub trait T { fn g(&self); } impl T for S { fn g(&self) {} } }` and assert `expect(findDanglingEdges(result)).toEqual([])` with **no** `types` argument, plus an explicit `METHOD_IMPLEMENTS` presence check for the trait impl.

gitnexus/src/core/ingestion/languages/rust/qualified-call.ts
```diff
@@ -380,35 +382,65 @@ function findReexportedMember(
...
+): PassModuleIndex {
+  let pass = MODULE_SCOPE_CACHE.get(workspaceIndex);
+  if (pass === undefined) {
+    const scopesByModule = new Map<string, Scope[]>();
+    const knownModuleNames = new Set<string>(index.moduleNames);
     for (const [filePath, moduleScope] of workspaceIndex.moduleScopeByFile) {
+      // Inline `mod x { … }` declares a Namespace def bound locally in the
+      // declaring scope; that binding is the only place an inline module's name
+      // exists, since it never appears in a file path.
+      for (const [name, refs] of moduleScope.bindings) {
```
magyargergo: **P2** — This loop reads bindings from `workspaceIndex.moduleScopeByFile`, which holds only each file's **root** `Module` scope (`workspace-index.ts:136`: `parsed.scopes.find((s) => s.kind === 'Module')`). A `mod` nested inside another inline `mod` binds in the *parent mod's* scope, so its name never enters `knownModuleNames` and `couldNameAModule` rejects the qualifier before any candidate search runs. `declaresSubmodule` at line 233 has the same root-only assumption, and `scopesByModule` gives inline modules no bucket at all, so lifting `knownModuleNames` alone would not be enough. Confirmed against the real pipeline at this sha: `mod outer { mod tools { fn dispatch() {} } fn dispatch() { tools::dispatch(); } }` still emits `CALLS outer.dispatch -> outer.dispatch` — the exact #2730 self-loop — with the correct `outer.tools.dispatch` node minted but unlinked. Not a regression (base does the same), so not merge-blocking on its own, but the new docblocks here and at `module-path.ts:52-56` read as complete coverage of "inline module names" when they cover depth-1 only. At minimum please narrow both docblocks and add a `describe.todo` or issue link for the nested case.

magyargergo (commented):
## Tri-review — two engines x {swarm, ce} + adversarial

Base `bc76ba2f25db18dc1c980a0750bae18124548c3a` → head `0062a5c2`.
Lanes: Claude swarm (branch-hygiene, risk, test-ci), Claude ce (correctness, adversarial, api-contract,
maintainability, standards, testing), Codex gpt-5.6-sol/xhigh (swarm, ce, adversarial),
plus coordinator first-hand pipeline reproductions at the same sha in a **built** worktree.

## Verdict

not production-ready

## Summary

The PR mod-qualifies Rust node ids so items inside inline `mod` blocks get distinct ids
(`Function:src/main.rs:inner.dispatch` vs the crate-root `Function:src/main.rs:dispatch`), reorders the
namespace-prefixed key ahead of the bare key in `resolveDefGraphId`, unions inline `mod` names into the
Rust qualified-call negative filter, and bumps `INCREMENTAL_SCHEMA_VERSION` 24→25 / `SCHEMA_BUMP` 31→32.
The headline problem is that mod-qualification is applied at node-**mint** time but every owner-edge
**anchor** (`findEnclosingClassInfo().classId`) is computed independently and stays unqualified, because
Rust's class extractor never sets `qualifiedNodeId`. Any `struct`, `trait`, `enum` or `impl` declared
inside a `mod` therefore gets a mod-qualified node while its `HAS_PROPERTY` / `HAS_METHOD` edges still
anchor on the bare id — the edges dangle and their rows are dropped at COPY time. The coordinator
reproduced this end-to-end on five lines of idiomatic Rust, reproduced a second variant that re-breaks
the #1975 scoped-impl contract this PR's own new comment claims to protect, and confirmed that CI stays
green only because every existing dangling assertion is narrowed to `['HAS_METHOD']` and no fixture puts
an `impl` inside a `mod`. The schema-bump discipline, the id-shape blast-radius sweep, and the shared
`ids.ts` reorder all came back clean across every lane.

## Blocking findings (P0/P1)

### B1 — P1 · Mod-qualified container ids with unqualified owner-edge anchors produce dangling `HAS_PROPERTY` / `HAS_METHOD` edges (and drop `METHOD_IMPLEMENTS`)

- **Confidence:** very high (reproduced end-to-end by the coordinator, and independently by two Claude lanes).
- **Corroboration:** STRONG — **coordinator reproduction** (built worktree, real pipeline, PR head);
  claude-ce-correctness F1 `[reproduced]`; claude-ce-adversarial F1+F2 `[reproduced]` over a 355-fixture
  base-vs-PR corpus diff; codex-ce P1 #1 (source-traced, confidence 100); codex-swarm F2 (same root
  cause, framed as sibling-module method collapse). Both engines, plus a reproduction.
- **Anchor (in diff):** `gitnexus/src/core/ingestion/workers/parse-worker.ts:2441-2447`
  (comment 2421-2440, call at 2446).

```ts
// parse-worker.ts:2441-2447 (added by this PR)
const qualifiedName =
  rustImplQualifiedName === undefined &&
  definitionNode !== undefined &&
  !enclosingClassInfo &&
  objectLiteralOwnerInfo?.ownerName === undefined
    ? qualifyByEnclosingModScope(definitionNode, qualifiedNameBeforeModScope)
    : qualifiedNameBeforeModScope;
```

The `!enclosingClassInfo` guard excludes *members*, which is what the new comment reasons about
("that owner's own id is mod-scoped by the impl qualifier above"). That reasoning holds only for
`impl_item` owners. A `struct_item` / `trait_item` / `enum_item` declared directly in a `mod` has **no**
enclosing class, so it is qualified — while its own owner-edge anchor comes from the untouched tail
branch of `findEnclosingClassInfo`:

```ts
// ast-helpers.ts:1035 (unchanged) — the BARE id; qualifiedClassId only fires when
// provider.classExtractor?.qualifiedNodeId === true, and class-extractors/configs/rust.ts
// (verified at head) sets only language / typeDeclarationNodes / ancestorScopeNodeTypes.
classId: generateId(label, `${filePath}:${classIdName}`),
```

**Failure trigger (coordinator, real pipeline at `0062a5c2`):**

```rust
// src/lib.rs
pub mod engine {
    pub struct Config { pub retries: usize }
}
```
```
NODE     Struct   id=Struct:src/lib.rs:engine.Config
EDGE     HAS_PROPERTY Struct:src/lib.rs:Config -> Property:src/lib.rs:Config.retries
DANGLING {"type":"HAS_PROPERTY","sourceId":"Struct:src/lib.rs:Config", ... "missing":"source"}
```

Coordinator second reproduction, `impl` inside a `mod` — this one dangles on `HAS_METHOD`, the edge type
existing assertions *do* check:
```
NODE     Struct Struct:src/lib.rs:outer.Engine
DANGLING HAS_METHOD Struct:src/lib.rs:Engine -> Function:src/lib.rs:Engine.go#0
(no METHOD_IMPLEMENTS edge at all)
```
The `METHOD_IMPLEMENTS` *absence* was base-diffed by claude-ce-adversarial (advP: base emits
`METHOD_IMPLEMENTS|…Btn.draw#0|…Draw.draw#0`, PR emits nothing) and observed at head by the coordinator.

**This already regresses the repository's own fixtures.** Verified at head:
`test/fixtures/lang-resolution/rust-generic-impl-same-method-name/lib.rs` is
`pub mod a { pub struct Inner<T> { v: T } … }` — exactly the shape. Both Claude lanes independently
measured the same two fixtures gaining new dangling edges
(`rust-generic-impl-same-method-name` base=0 → pr=1; `rust-nested-tail-collision-generic` base=0 → pr=2).

**Why CI is green:** every dangling assertion in the Rust suite is narrowed —
`findDanglingEdges(result, ['HAS_METHOD'])` at `rust.test.ts:2061, 2111, 2162, 2202, 2250, 2279`, and a
repo-wide grep confirms **no call site anywhere in `test/` uses the unfiltered one-argument form**.

**Downstream effect:** `csv-generator.ts` streams relationships with no node-existence check and the
failing pair COPY is silently retried with `IGNORE_ERRORS=true`, so the bad rows are dropped — every
struct inside a `mod` silently loses its fields, every trait inside a `mod` loses its methods, and
`context({name:'Config'})` returns no fields. `pub mod config { pub struct … }` and
`#[cfg(test)] mod tests` are ubiquitous, so the blast radius is close to "most Rust repositories".

**Suggested fix.** Restore byte-for-byte owner/node agreement. Narrowest option, and the one the
coordinator recommends: restrict the new qualification to items that own no members (free `fn` /
`const` / `static` / `macro` / `mod`) — container qualification buys nothing for #2742's actual goal,
which is function representability. Fuller option: move the mint and the anchor to one source, i.e.
qualify `findEnclosingClassInfo`'s container branch (and the `for`-target branch at `ast-helpers.ts:916`)
with the same walk. Add a regression fixture with `mod m { pub struct S { pub f: usize } pub trait T { fn g(&self); } impl T for S { fn g(&self) {} } }`
asserting **unnarrowed** `expect(findDanglingEdges(result)).toEqual([])` plus an explicit
`METHOD_IMPLEMENTS` presence check.

---

### B2 — P1 · The #1975 scoped-impl contract this PR's own new comment says it protects is re-broken one `mod` deeper

- **Confidence:** very high (coordinator reproduction + independent Claude-lane reproduction).
- **Corroboration:** STRONG — **coordinator reproduction**; claude-ce-adversarial F3 `[reproduced]`
  (advS, with the crate-root twin as a control); codex-ce cites the same mechanism
  (`ast-helpers.ts:963-968`, scoped impl owner from unqualified `baseType.text`) as evidence for its P1 #1.
  Same root cause as B1 but a different code path and a different anchor, so it is called out separately.
- **Anchor (in diff):** `gitnexus/src/core/ingestion/utils/ast-helpers.ts:41-48`.

```ts
// ast-helpers.ts:41-48 (added by this PR)
// No enclosing `mod`: return the raw text UNTOUCHED. Normalizing here would
// rewrite a scoped target's separator (`a::Inner` -> `a.Inner`) and silently
// move its node id away from the one the owner edge emits, which is how this
// helper first broke the #1975 scoped-impl ownership when it was generalized
// beyond unscoped targets. …
if (modSegments.length === 0) return rawText;
return [...modSegments, ...splitQualifiedName(rawText)].filter(Boolean).join('.');
```

The guard fires only when `modSegments.length === 0`. Inside a `mod`, the early return does not fire,
`splitQualifiedName` normalises `a::Inner` → `a.Inner`, and the mod prefix is applied — while
`findEnclosingClassInfo`'s scoped arm still emits `Impl:…:a::Inner`. The hazard the comment describes is
avoided only in the zero-`mod` case, which is precisely the case this PR newly reaches past.

**Failure trigger (coordinator, real pipeline at `0062a5c2`):**
```rust
pub mod outer {
    use crate::a;
    impl a::Inner { pub fn helper(&self) {} }
}
```
```
NODE     Impl   Impl:src/lib.rs:outer.a.Inner        <- minted mod-qualified AND ::-normalised
DANGLING HAS_METHOD Impl:src/lib.rs:a::Inner -> Function:src/lib.rs:a::Inner.helper#0
```
claude-ce-adversarial's advS reproduces the same delta base-vs-PR and shows the crate-root twin
(`impl b::Inner` outside any `mod`) is byte-identical on both legs — the early return works, the hazard
survives one `mod` deeper. The mod-nested `Impl` node ends up with zero members and its method orphaned.

**Suggested fix.** Do not fire the new general caller for `nodeLabel === 'Impl'` (the impl branch above
already owns that decision), or teach `findEnclosingClassInfo`'s scoped arm (`ast-helpers.ts:966`) the
same mod walk. Either way, extend the early-return contract so a scoped raw text is never re-separated,
not just when there is no enclosing `mod`. Add a `mod outer { impl a::Inner { … } }` fixture —
`rust-scoped-impl` only covers the crate-root form.

---

### B3 — P1 (testing) · No test can see any of this: every dangling assertion in the repo is narrowed to `HAS_METHOD`

- **Confidence:** very high (verified directly at head; both engines reached it).
- **Corroboration:** STRONG — claude-ce-correctness "testing gaps" #2, claude-ce-adversarial "testing gaps"
  #1, codex-ce and codex-swarm both call for the same regression coverage, coordinator confirms
  ("the regression test for this fix must call `findDanglingEdges(result)` **unnarrowed**").
- **Anchor (in diff):** the new describe block at `gitnexus/test/integration/resolvers/rust.test.ts:2802-2832`.

Grep at head: six `findDanglingEdges(result, ['HAS_METHOD'])` call sites in `rust.test.ts` and **zero**
unfiltered call sites anywhere in `test/`. That is exactly why the two `HAS_PROPERTY` dangles the PR
introduces on the repo's own fixtures ship with a green 15,801-test CI run. This is listed as blocking
because it is the reason B1 and B2 are invisible, and the fix is one line in the new describe block.

## Non-blocking (P2/P3)

- **P2 · Nested inline `mod` still self-loops — the #2730 symptom survives one level deeper, and the new
  docblock over-claims coverage.** Corroboration: STRONGEST cross-engine of any finding —
  **coordinator reproduction** (CONFIRMED, rated P2: not a regression, base self-loops too),
  claude-ce-correctness F2 `[reproduced]`, codex-swarm F1 (P1), codex-ce P1 #2, codex-adversarial [high].
  `passIndexFor` collects Namespace bindings only from each file's root `Module` scope
  (`qualified-call.ts:411-421`; `workspace-index.ts:136` stores one root `Module` per file), so a `mod`
  nested inside a `mod` never enters `knownModuleNames` and `couldNameAModule` rejects the qualifier
  before the module walk runs. `declaresSubmodule` (`qualified-call.ts:233`) has the same root-only
  assumption, and `scopesByModule` gives inline modules no bucket at all, so lifting `knownModuleNames`
  alone would not resolve it. Coordinator trigger at head:
  `mod outer { mod tools { fn dispatch() {} } fn dispatch() { tools::dispatch(); } }` →
  `CALLS outer.dispatch -> outer.dispatch` (self-loop) with `outer.tools.dispatch` minted but unlinked.
  Note the honest boundary: claude-ce-adversarial R12 **refuted** this empirically on advI
  (`mod a { mod b { fn deep } }`, distinct names) — the gap only manifests when the enclosing module has
  a same-named item for the lexical fallback to land on. Two Codex lanes rated it merge-blocking; the
  coordinator's reproduction rates it P2 because base behaves identically. Minimum in-PR action: narrow
  the docblock at `module-path.ts:52-56` and `qualified-call.ts:385-397`, which currently reads as
  complete coverage of "inline module names", and add a `describe.todo`/issue link.
- **P2 · A widened negative filter can fabricate a wrong cross-file `CALLS` edge.** Single-lane
  (claude-ce-adversarial F5) but `[reproduced]` base-vs-PR: with `pub mod Buffer { … }` at the calling
  crate's root and `use crate::c::Buffer;` in a sibling file, base emitted **no** edge (the correct
  failure direction) and the PR emits `CALLS … -> Function:src/lib.rs:Buffer.with_capacity`. Violates the
  doctrine quoted in `ids.ts` itself ("a fabricated caller is not [the correct failure direction]").
  claude-ce-correctness Q5 refuted the broader class (`couldNameAModule` is a pure negative gate) and
  could not build a same-crate false positive; the reproduced case needs the module at the *calling*
  crate root, which is why the two Claude lanes differ. Trigger requires a module named like a type —
  narrow, but mechanically reproducible.
- **P2 · Undisclosed collapse classes.** The PR discloses exactly one residual (same-named methods on
  same-named types in sibling modules). claude-ce-adversarial F4 `[reproduced]` adds three more that are
  now *strictly worse* than base — struct fields, trait methods and trait-impl methods keep bare ids, so
  they still collapse across sibling mods **and** are now ownerless. codex-swarm F2 reaches the same
  conclusion from source. Disclose, or fix with B1 (qualifying the owner name qualifies
  `${className}.${name}` member ids too).
- **P3 · Tagged lookup keys still try the bare form before the namespace-prefixed one.**
  Corroboration: claude-swarm-risk (P3, "cannot produce a wrong result"), codex-adversarial (medium,
  synthetic resolver probe). Verified at head: `ids.ts:293-297` builds `${qn}#${parameterCount}`
  (`dispatch#0`) while `node-lookup.ts:192-195` registered the mod-qualified `inner.dispatch#0`, and the
  arity / parameter-types / shape / constraints keys all run **before** the reordered nsPrefix key at
  `ids.ts:320-324`. **Coordinator explicitly refuted its own escalation of this to a correctness bug** —
  see "Considered and refuted". An end-to-end run at this sha with a crate-root same-name/same-arity
  decoy binds correctly, so this is a key that misses, not one that mis-hits. Which later key rescues it
  is disputed between lanes and is deliberately not asserted here. Codex's variant puts both defs on one
  source line to tombstone the position key; that is contrived formatting and was probed synthetically,
  not run end-to-end. Keep as hygiene: compose `nsPrefix + '.' + qn` into the tagged key builders too.
- **P3 · The "cheap rejection BEFORE any index work" comment is now false.** `qualified-call.ts:91-101`:
  `passIndexFor(...)` is now evaluated as the *argument* to `couldNameAModule`, so the first qualified
  free call in a pass walks every module scope and every binding — the workspace-wide walk #2741 moved
  off the hot path. claude-ce-correctness F3 and claude-ce-adversarial F8 flag it; claude-swarm-risk
  **refuted** it as a real perf regression (WeakMap-cached once per pass over an already-in-memory map;
  the #2649 out-of-core work applies to the graph buffer pool, not scope indexes). Correctness
  unaffected — fix the comment.
- **P3 · Mod segment lands outermost for a `mod` declared inside a `fn`.** `[reproduced]` by
  claude-ce-adversarial F6 and claude-ce-correctness Q3: `fn wrapper() { mod helper { fn dispatch } }`
  mints `Function:lib.rs:helper.wrapper.dispatch@2:8`, inverting the real nesting. Ids stay unique (the
  `@L:C` suffix), no edge breaks — id-shape incoherence only.
- **P3 · `union` fields get mod-qualified, `struct` fields do not** (claude-ce-adversarial F7,
  `[reproduced]`): `union_item` is not an `enclosingClassInfo` container, so its fields fall through the
  `!enclosingClassInfo` gate. Nothing dangles (union fields carry no `HAS_PROPERTY` in either build) —
  a naming-scheme inconsistency for whoever next reconciles member ids with owner ids.
- **P3 · The `ids.ts` reorder silently retargets a C++ `EXTENDS` edge, and it is untested.**
  `[reproduced]` by both Claude lanes over 102 / 355 fixtures respectively:
  `cpp-two-phase-dependent-base-cross-ns-deep` goes from the global decoy `Struct:lib.h:Inner` to the
  correct `Struct:lib.h:ns.a.b.Inner`. It is an improvement, but it rides unmentioned in a Rust-scoped PR
  with no assertion protecting it against a future re-reorder. claude-ce-api-contract also notes C# has
  no integration-level coverage of this dual-key path at all.
- **P3 · Test hygiene.** `rust.test.ts:2826-2830`'s fixture-wide "no self-loops anywhere" is strictly
  weaker than the preceding test (which pins the exact `targetId`) and would also fire on legitimate Rust
  self-recursion added to the shared fixture later (claude-ce-testing; claude-swarm-test-ci agrees it is
  load-bearing today and worth keeping with a comment pinning *why*). No unit test exists for
  `qualifyByEnclosingModScope` or `couldNameAModule`'s new signature, though the sibling `module-path.ts`
  got a dedicated unit file in #2741.
- **P3 · Maintainability.** `qualifyByEnclosingModScope`'s docblock still documents only the impl-target
  contract even though the general name now has a second, looser caller 2400 lines away
  (`ast-helpers.ts:16-28`); the "locally declared submodule" predicate is now hand-copied in two places
  one screen apart (`qualified-call.ts:228-237` vs `:415-419`); and the new 4-clause gate is inlined into
  a ternary while the two conditions directly above it in the same function are extracted into named
  `const`s (`parse-worker.ts:2380-2391`).

## Considered and refuted

- **The #2743 "an id-shape change makes every synthetic corpus a call site" lead — REFUTED.** Coordinator
  CV-1 swept the whole repo (not just `src/`); the three files with hard-coded Rust node ids
  (`closure-review-findings.test.ts`, `test/integration/has-method.test.ts`, `test/unit/has-method.test.ts`)
  contain **zero** `mod` blocks, so none of their ids can move; no golden records a Rust node id; `bench/`
  has no Rust node-id corpus. Independently re-run and confirmed negative by claude-swarm-test-ci,
  claude-swarm-branch-hygiene, claude-ce-adversarial R11, claude-ce-correctness residual #5 and codex-ce.
  This is the one high-value prior-context lead, and it does not bite this PR.
- **Arity-key mismatch as a live misresolution — REFUTED by the coordinator's own reproduction.** The
  coordinator escalated the risk lane's P3 (a crate-root sibling with the same name and arity could be hit
  by the bare `dispatch#0` key before the nsPrefix key). Tested directly at head with
  `pub fn dispatch() {}` + `pub mod tools { pub fn dispatch() {} }` + `pub fn run() { tools::dispatch(); }`
  → `CALLS run -> Function:src/main.rs:tools.dispatch` (**correct**; the decoy was not chosen). The
  escalation was wrong; the lane's P3 rating stands. Carried through as a P3 hygiene note above.
- **Schema renumber v24→v25 / `SCHEMA_BUMP` 31→32 collision (the `v29` precedent) — REFUTED** by four
  lanes independently. `origin/main` at review time is `bc76ba2f` carrying v24/31, so 25/32 are free;
  `repo-manager.ts:575`, `parse-cache.ts:131`, the pin test and the reuse-gate test all move together
  (`passesReuseGate(24) === false`, `(25) === true`); both bumps are warranted and sufficient (the
  `PARSE_CACHE_VERSION` bump also covers the durable ParsedFile store).
- **A `node-lookup.ts` lockstep break — REFUTED.** The lookup key is derived from the materialized node id
  (`props.qualifiedName ?? parseQualifiedFromId(node.id, …)`, `node-lookup.ts:161-162`), so it follows the
  new mint automatically; `test/unit/callable-id-lockstep.test.ts` passes. The real lockstep break lives in
  a *fourth* phase nobody enumerated — `findEnclosingClassInfo` (owner-edge minting) — which is B1/B2.
- **The `ids.ts` reorder as a cross-language regression — REFUTED** over 102 (claude-ce-correctness Q4)
  and 355 (claude-ce-adversarial R13) fixtures with a base-vs-PR node-and-edge diff: exactly one non-Rust
  behaviour change, and it is a fix. Counterexample construction also fails — a `P.qn` key is by
  construction the def's own fully-qualified path, and the only way another node occupies it is a
  redeclaration error in both C++ and C#.
- **The `ast-helpers.ts` early return changing any pre-existing id — REFUTED.** Both prior callers are
  gated on `type_identifier` text, for which `splitQualifiedName(x).join('.')` is the identity; empirically
  `rust-scoped-impl` is byte-identical base-vs-PR for its `Impl` and method ids. (The scoped hazard is
  real, but only *inside* a `mod` — that is B2, not a refutation of the guard.)
- **Perf regression from the workspace-wide binding scan — REFUTED** (claude-swarm-risk): WeakMap-cached
  once per pass over an in-memory map, ~500k iterations at 10k files × 50 bindings; no disk-fault path.
  Only the stale comment survives, as a P3.
- **An AGENTS.md language-neutrality violation in shared ingestion — REFUTED** (claude-ce-standards, with
  the rule's own text and precedent). The rule targets branching on a *language name*; the new helper keys
  purely on the `mod_item` node type and the PR *reduces* Rust-named exported surface. Codex-ce cited
  AGENTS.md as supporting evidence for its P1 — the defect is real, the standards framing is not.
- **Other adversarial attacks that did not become triggers:** `#[path = "other.rs"] mod aliased;`
  (byte-identical graphs); `mod a { mod a { fn a } }` (all ids distinct, strict improvement); raw
  identifiers `r#mod::r#fn` (retargets correctly, an improvement); `fn` nested in `fn` inside a `mod`
  (composes correctly, resolves via the position key); a non-Rust grammar with a `mod_item` node type
  (grepped all 15 grammars — Rust only); `mod` inside `impl` (not expressible in Rust); the sequential
  `parsing-processor.ts` path still minting ids (it no longer mints any).
- **"Missing fixture" for the new tests — REFUTED:** `rust-2730-gaps` is in the tree and already contains
  the `mod inner { pub fn dispatch }` / crate-root `fn dispatch` pair from #2741.
- **All three new tests are load-bearing — CONFIRMED, not refuted** (claude-swarm-test-ci and
  claude-ce-testing independently): test 2 fails if *any one* of the three production changes is reverted
  alone, and the base-vs-PR diff on `rust-2730-gaps` shows exactly the self-loop test 3 targets.

---

### Method

Five review lanes on two engines: Claude (GitNexus PR swarm + ce-code-review personas) and Codex
`gpt-5.6-sol`/xhigh (the same swarm and ce contracts, plus a native adversarial pass). Same scaffolding,
different engines, so engine agreement is a clean A/B; agreement *within* one engine is only
"consistent across personas". Every lane ran read-only against the PR head in a worktree. The blocking
findings were additionally reproduced end-to-end by the coordinator in a built worktree at this sha —
those reproductions, not lane consensus, are what the P1s rest on. An independent critic vetted this
digest against the source before posting.

Author note: this is my own PR; the review is posted because the lanes found real defects in it, and
the reproductions are included so they can be checked rather than taken on trust.

gitnexus/src/core/ingestion/scope-resolution/graph-bridge/ids.ts
```diff
@@ -308,19 +308,22 @@ export function resolveDefGraphId(
...
+    // same-named item at a DIFFERENT namespace depth in the same file and
+    // returns it before this retry is ever reached (#2742 — a call into
+    // `mod inner { fn dispatch }` bound to the crate-root `fn dispatch`, which
+    // rendered as a self-loop).
     const nsPrefix = def.namespacePrefix;
     if (nsPrefix !== undefined && nsPrefix.length > 0) {
       const nsHit = nodeLookup.get(qualifiedKey(filePath, def.type, `${nsPrefix}.${qn}`));
       if (nsHit !== undefined) return nsHit;
     }
+    const qualifiedHit = nodeLookup.get(qualifiedKey(filePath, def.type, qn));
```
magyargergo: **P3** — The reorder is applied to the untagged key only. The arity, parameter-types, parameter-shape and constraints keys above (lines 253-309) all still compose from the bare `qn` — e.g. `${qn}#${def.parameterCount}` at line 294 builds `dispatch#0` — while `node-lookup.ts:192-195` registered the mod-qualified `inner.dispatch#0`, so for a Rust `fn` inside a `mod` those keys simply miss and resolution falls through to the nsPrefix key here. Verified as a miss, never a mis-hit, in practice: an end-to-end run at this sha with a crate-root same-name/same-arity decoy binds correctly, so a later key rescues it. Cheap hygiene fix and worth doing while this code is open — compose `${nsPrefix}.${qn}` into the tagged key builders too, so the tagged path stops being dead for mod-qualified defs rather than depending on which later key happens to catch it.

gitnexus/src/core/ingestion/workers/parse-worker.ts
magyargergo: Confirmed and fixed in `454e2e19`. Reproduced it exactly as described — five lines of Rust dangled `HAS_PROPERTY`, and a trait impl inside a `mod` also dropped `METHOD_IMPLEMENTS` outright.

Took the narrow fix you suggested, gated on `MEMBER_OWNER_NODE_TYPES` derived from `CLASS_CONTAINER_TYPES` rather than a fresh label list: that set is already the single source of "this node type owns member edges" and already carries an INVARIANT note binding it to `CONTAINER_TYPE_TO_LABEL`, so a language adding a container cannot gain a mismatched id shape here without failing that invariant. `union_item` is listed too — its fields are captured as `Property` but it is not a recognized owner, so they cannot dangle; it is there so a union keeps the same id shape as the struct beside it, which also closes the separate P3 you filed on that.

The gate is now a named `qualifiesByEnclosingModScope` const, matching the two conditions above it that were already extracted.

gitnexus/src/core/ingestion/utils/ast-helpers.ts
magyargergo: Confirmed and fixed in `454e2e19`. You are right that the guard only removes the hazard at `modSegments.length === 0`, and that the comment describes the failure the code then causes one `mod` deeper. Reproduced: `Impl:src/main.rs:outer.a.Inner` minted against an anchor still reading `Impl:src/main.rs:a::Inner`.

Fixed at the gate rather than in the helper — `impl_item` is in `MEMBER_OWNER_NODE_TYPES`, so the scoped targets the impl branch deliberately excluded are no longer picked up by the general caller. The docblock now documents both callers and which gate belongs to which, since that gap is what let this through.

New fixture covers the scoped-impl-inside-a-mod case that `rust-scoped-impl` did not reach.

gitnexus/test/integration/resolvers/rust.test.ts
magyargergo: Confirmed and fixed in `454e2e19`. Added the fixture you asked for, asserting `findDanglingEdges(result)` with no `types` argument plus an explicit `METHOD_IMPLEMENTS` presence check.

One addition worth noting: the tests also assert the NODE id, not just the edge anchor. My first version asserted only the anchor and two of the four tests passed in the broken build — the anchor was already bare while the bug was live, so an edge-only assertion cannot see the half that moved. All four now fail when the gate clause alone is reverted.

gitnexus/src/core/ingestion/languages/rust/qualified-call.ts
magyargergo: Confirmed and fixed in `2f5cc47e`. You were right that lifting `knownModuleNames` alone would not be enough, and right about `declaresSubmodule` having the same root-only assumption.

The channel turned out to be a third one: a `mod` def carries no nesting information of its own — inside `mod outer { mod tools { … } }` the inner def is `qualifiedName: "tools"` with NO `namespacePrefix` — and a `Namespace` scope owns its OWN def rather than its children's, so the `mod outer` scope lists `outer`, never `tools`. Inline module paths are therefore derived from the MEMBERS' `namespacePrefix`, which is stamped correctly at any depth. Restricted to non-empty prefixes so this stays a declaration check and an undeclared file on disk still cannot outrank a `use` binding.

Fixed rather than documented as depth-1, with fixtures at depth 2 and 3. The docblocks you flagged as over-claiming are now accurate instead of narrowed.

The `[reproduced]` fabrication case from the same lane is fixed separately in `15b24114`.

gitnexus/src/core/ingestion/scope-resolution/graph-bridge/ids.ts
magyargergo: Confirmed and fixed in `c518b268` — agreed it is hygiene rather than a live bug, and worth doing while the code is open.

All five tagged keys plus the plain one now go through one `lookupTagged` helper that tries the namespace-prefixed form first, so a sixth tagged key cannot be added with the bare form only. That also removes five hand-repeated `qualifiedKey` / `nodeLookup.get` pairs.

Also pinned the C++ `EXTENDS` retarget from the separate P3: `cpp-two-phase-dependent-base-cross-ns-deep` has a global `Inner` decoy alongside `ns::a::b::Inner`, so it is a real assertion of the ordering rather than a snapshot.

PR #2747, opened 2026-07-29
feat(resolution): type receiver chains from AST structure across all 14 languages (#2708) + epistemic lower-bound (#2744)

### LINKED ISSUE #2744: impact()/context() report epistemic: "exact" when an unresolved receiver silently dropped a caller (#2708 second half)
Split out from #2708, which was auto-closed by #2737. #2737 fixed that issue's **first** ask (the missing `CALLS` edge for inline constructor receivers) but explicitly did not address the **second** one, and there was no successor issue — so this half would otherwise be lost.

## The ask

`impact(direction: "upstream")` and `context()` report `epistemic: "exact"` even when a receiver failed to resolve and the caller was silently dropped from the result.

The original reporter's argument for why this matters independently of any individual resolution bug:

> A **wrong count labelled a lower bound** is survivable — it tells the caller to go verify by hand. A **wrong count labelled `"exact"`** is acted on as authoritative.

That is the whole point: consumers (including the agent workflows in this repo, which mandate an `impact` check before editing a symbol) treat `"exact"` as a guarantee.

## Why #2737 could not fix it

`computeEpistemicBoundary` (`gitnexus/src/mcp/local/local-backend.ts`) keys **only** on the target symbol's own heritage edges — `IMPLEMENTS` / `METHOD_IMPLEMENTS` / `EXTENDS`. A symbol with no supertypes produces an empty boundary set and returns `"exact"` at the first check:

```ts
if (boundary.size === 0) return { epistemic: 'exact' };
```

Dropped references are not an input to it, and cannot be as things stand: unresolved outcomes accumulate in an in-memory `resolutionOutcomes[]` (`scope-resolution/pipeline/run.ts`) that reaches the pipeline result and is **never written to the index**, while `epistemic` is computed at query time against the index. The two never meet.

Worth noting the resolution pass does not even record every drop today — a receiver that fails to resolve in the compound-receiver path falls through silently; the `recordResolutionOutcome({kind: 'suppressed'})` calls nearby cover ambiguous member lookup, not unresolved receivers.

## Sketch of what it needs

1. Record unresolved-receiver drops as a resolution outcome (they are currently silent in at least the compound-receiver path).
2. Persist a per-symbol (or per-file) count of unresolved references into the index, alongside the existing graph.
3. Have `computeEpistemicBoundary` consult it, so a symbol whose callers include unresolved sites degrades to `epistemic: "lower-bound"` with a boundary note — the mechanism #1858 already built for interface indirection.
4. An `INCREMENTAL_SCHEMA_VERSION` bump, since it changes what is persisted.

Step 2 is the design decision worth discussing before implementing: whether this is a new column on an existing table, a sidecar table, or an aggregate stamped at analyze time.

## Related

- #1858 — the epistemic lower-bound mechanism this should plug into
- #2708 / #2737 — the edge half, fixed
- #2686, #2129 — other cases of `impact()` under-reporting

magyargergo:
Types method-call receivers from **captured AST structure** instead of by re-parsing the receiver's source text, and finishes the `#2708` epistemic work (`#2744`) that shipped alongside it.

`resolveCompoundReceiverClass` dispatches on enumerated textual shapes, so every new spelling has needed another branch — and several never got one. The structure was already being computed (`extractMixedChain` walks the real AST) but its output landed in a field nothing read. This carries that structure through to resolution.

## What now resolves

| Shape | Before | After |
|---|---|---|
| `svc?.getUser().save()` | no edge, **no drop recorded** | resolves |
| `svc!.getUser().save()` | no edge | resolves |
| `svc.getTyped<User>().save()` | no edge, **no drop recorded** | resolves |
| C++ `svc->getUser()->save()` | no edge, **no drop recorded** | resolves |

Emission is wired into **all 14 language emitters** through one shared, language-free helper (it gates on the `@reference.call.*` tag vocabulary every `.scm` query already shares, per AGENTS.md R6).

## How it works

- `ReferenceSite.receiverChain` — a compact string (`1|svc|cgetUser|faddress`, 35 serialized bytes/site). A string rather than `MixedChainStep[]` because the store's interning reviver re-shares objects only when they carry `nodeId` + `filePath`, which a chain step does not.
- `foldReceiverChain` resolves the base, then folds steps base-first. It runs **ahead of** the text cascade but **falls through** on failure — structure is an additional route to an answer, never a veto on the existing one.
- The encoder **refuses** rather than degrades: a truncated or unencodable chain is not minted, because a half-encoded chain decodes cleanly as a *different*, shorter one.

## Measurement

A two-arm bench (`bench/receiver-resolution/`, CI-gated) because a drop-count gate alone is **blind to the shapes this fixes** — three of them recorded no drop at all, so the call-drop count read 99 → 99 while three shapes flipped to resolving. The shape arm asks about *edge presence*, so it is blind to nothing.

Perf, A/B against a control build: **+0.86%** wall-clock, **+0.06%** store growth, 35.2 bytes/site. Peak RSS is reported as *unresolvable* — independent runs spread 5.3% against a ±2% budget, so the instrument cannot answer it.

## Review findings, all fixed

This branch was reviewed by a multi-engine pass (Claude + GPT-5.5) which found two P0s in the original implementation, both now fixed with A/B proof and regression tests verified to fail without their fix:

- **Fabricated edge.** The fold's base used the permissive bare-identifier path, which falls through to a class-name lookup *even when a typeBinding exists but names no class* — so a local shadowing a class name (`const Config = make(1)`) was typed as that class and emitted a confident wrong edge. Fixed with an opt-in `strictBaseBinding` that mirrors the cascade's narrower head rule; the flag is off for the cascade, so no existing edge moves.
- **`impact()` emitted a function object as a call count.** `counts?.[symName]` walked `Object.prototype`, and `NaN <= 0` is false, so `impact({target:"constructor"})` returned `function Object() { [native code] } call sites invoking …`. Now one prototype-safe helper.

Plus: the drop summary now counts **call** sites only (it was including property reads/writes, 25 of 124 on the corpus); two tests that could not fail were replaced; and 8 P3s.

## Schema versions

`SCHEMA_BUMP` 32 → **34** and `INCREMENTAL_SCHEMA_VERSION` 25 → **28**, each landing at a real divergence rather than only at the first one. Both were renumbered during the rebase: `main` had independently taken 32/25 for `#2742`, which is the same collision the ladder's own v21 note warns about — re-check both against `origin/main` immediately before merging.

## Verification

4400+ tests green · `tsc --noEmit` clean in both packages · all six bench gates pass (`scope-capture` 15 languages, `python-scope` capture + import-target, `receiver-resolution`, `callable-value-flow`, `scope-emission`).

## Known residuals — recorded, not fixed

- `(await svc.getUserAsync()).save()` — the walk reaches `await …`, not a chain node.
- `repos[0].save()` — Case 0's punctuation gate never fires for a subscript receiver.
- PHP `$svc->getUser()->save()` — the chain *is* minted correctly; the base `$svc` does not bind in the PHP resolver. Verified **pre-existing**, not a regression, by a fold-revert A/B.
- Group-B (`TypeRef.rawName`) sites have no reference site in scope to carry a chain.


gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -199,6 +208,119 @@ function resolveConstructionExpressionClass(
...
+  // future producer that does.
+  if (chain.truncated) return undefined;
+  if (chain.steps.length === 0) return undefined;
+
+  // `receiverChain` is dropped before resolving the base: it describes the
+  // whole receiver, and handing it back to the resolver would re-enter this
+  // fold on the base and never terminate.
+  let current = resolveCompoundReceiverClass(chain.baseReceiverName, inScope, scopes, index, {
+    ...options,
+    fieldFallback: false,
```
magyargergo, line 311: **P0** — Fold base uses the permissive bare-identifier path, so a local shadowing a class name is typed as that class and FABRICATES a CALLS edge the cascade never emitted (A/B verified)

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -5676,10 +5694,39 @@ export class LocalBackend {
...
+   * Empty when the index records no drops for this name — including every
+   * index written before the summary existed, which is why the schema version
+   * was bumped rather than treating "absent" as "none".
+   */
+  private async unresolvedReceiverBoundaries(repo: RepoHandle, symName: string): Promise<string[]> {
+    if (symName.length === 0) return [];
+    try {
+      const meta = await loadMeta(path.dirname(repo.lbugPath));
+      const summary = meta?.unresolvedReceiverMembers;
+      const sites = summary?.counts?.[symName];
```
magyargergo: **P0** — counts?.[symName] walks Object.prototype; impact({target:"constructor"}) emits a Function as the call count and NaN<=0 defeats the guard

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/src/storage/parse-cache.ts
```diff
@@ -125,7 +125,14 @@ import type { ParseWorkerResult } from '../core/ingestion/workers/parse-worker.j
...
 // v16: direct callee identity.
-const SCHEMA_BUMP = 31;
+// v32: TypeScript call matches carry `@reference.receiver-chain`, a compact
+// encoding of a receiver that is itself an expression. Lands HERE, at the first
+// commit where the capture SET actually diverges — not earlier with the field
+// that carries it, which changed nothing a warm cache could replay wrongly. A
+// cache written at v31 has no chain capture at all, so replaying it would make
+// the feature look inert rather than broken, which is the harder failure to
+// diagnose.
+const SCHEMA_BUMP = 32;
```
magyargergo: **P1** — Both schema constants bump before the 13-language capture/edge rollout at c98668257; an intermediate v32/v26 build reads HEAD as current

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/test/integration/resolvers/typescript.test.ts
```diff
@@ -3312,3 +3306,193 @@ describe('TypeScript inline constructor receiver resolution', () => {
...
+    expect(calls.find((c) => c.source === 'alreadyWorked' && c.target === 'save')).toMatchObject({
+      target: 'save',
+      targetFilePath: 'models.ts',
+    });
+  });
+
+  it('emits no CALLS edge to a member that exists on no class in the chain', () => {
+    // A broken middle step must produce NO edge, never a wrong one.
+    const calls = getRelationships(result, 'CALLS');
+    expect(calls.filter((c) => c.target === 'noSuchMember')).toHaveLength(0);
```
magyargergo: **P2** — Vacuous negative assertion: noSuchMember appears nowhere else in the repo, so this test cannot fail

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/test/integration/resolvers/typescript.test.ts
```diff
@@ -3312,3 +3306,193 @@ describe('TypeScript inline constructor receiver resolution', () => {
...
+
+  it('resolves a non-null-asserted receiver — svc!.getUser().save()', () => {
+    const calls = getRelationships(result, 'CALLS');
+    expect(calls.find((c) => c.source === 'viaNonNull' && c.target === 'save')).toMatchObject({
+      target: 'save',
+      targetFilePath: 'models.ts',
+    });
+  });
+
+  it('resolves a mixed call/field chain through the fold', () => {
```
magyargergo, line 3502: **P2** — Named "through the fold" but the shape is resolved by the pre-existing cascade (covered at base :2189); passes with the fold deleted

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -199,6 +208,119 @@ function resolveConstructionExpressionClass(
...
+ * fails to type, so the caller falls back to the existing text cascade rather
+ * than receiving a partially-folded guess — a missing edge is recoverable, a
+ * confidently wrong one is not.
+ *
+ * The base is resolved by handing it to `resolveCompoundReceiverClass`, which
+ * already owns the bare-identifier path in full: type binding, static
+ * class-name receivers, map-tuple sentinels, member aliases and call-result
+ * aliases. A second implementation of that would drift from it.
+ *
+ * NOT WIRED into any resolution path yet — see U10.
```
magyargergo: **P3** — Stale JSDoc: "NOT WIRED into any resolution path yet" — the same file wires it at :350-355

_Surfaced by the multi-engine tri-review; verified against this diff._

magyargergo (commented):
## Multi-engine review — PR #2747 (receiver-chain structural typing + #2744 epistemic lower-bound)

**Engines:** 9 lanes returned usable output across 2 engines — Claude swarm (risk-architect, security-boundary), Claude ce (correctness, adversarial, maintainability, project-standards, testing), Codex/GPT-5.5 ce, Codex adversarial. 2 lanes did **not**: Codex's swarm leg (B1) **died on a workspace spend cap** before reading a single file, and the Claude test/CI lane returned partial twice (so the CI-timing question below is unresolved). **This is not the full five-lane, two-engine review the tool aims for** — Codex contributed 2 of its 3 legs.

---

### What's solid

The two headline risks this PR was most likely to carry were both hunted hard and came back clean:

- **Sibling construction-selector wrong-class folds: REFUTED across all 14 languages** `[code-read]`. Ruby is the only language defining `constructionSyntax.selector`, so the `options.constructionSyntax?.selector === step.name` guard cannot fire elsewhere. Probed `Factory.new.run`, `factory.new.run`, `Foo.allocate`, C# `user?.Method()` (`conditional_access_expression` is absent from the node tables ⇒ no chain minted), Kotlin `svc!!` (`postfix_expression` not in `TRANSPARENT_RECEIVER_WRAPPERS` ⇒ base lookup fails ⇒ fold declines), Rust `Foo::bar()`, PHP `$this`. Every outcome is no-chain or a failed fold falling through to the cascade — **never a wrong class**. Codex's adversarial lane independently reached the same conclusion ("could not construct a new wrong-class receiver fold"). Cross-engine agreement on the PR's own stated top risk.
- **Codec attack surface: REFUTED on five separate fronts** `[code-read]`. Decoder hang/throw (total function; 512 B of `|` → 513 parts → fails VERSION → `undefined`); quadratic split (O(n), depth ≤ 3, ≤ 512 B); delimiter confusion across all 14 languages (Ruby operator methods `|`/`~`, Kotlin/Swift backticked identifiers — tree-sitter `.text` includes the backticks — and C++ `operator|` are all refused by `isEncodableSegment` *before* minting, so no payload can decode as a different valid chain); emit/load bounds skew (encoder `:102` and decoder `:113` read the same `MAX_RECEIVER_CHAIN_BYTES` from a single import site — this is the documented warm-cache-miss-loop defect class, and it does not apply here); sanitizer poisoning (all four shapes traced, no unlogged loss). Fold→`resolveCompoundReceiverClass` infinite recursion also refuted: the base call passes `receiverChain: undefined` explicitly, with `COMPOUND_RECEIVER_MAX_DEPTH = 8` as a backstop.
- **AGENTS.md R6** (no language names in shared `src/core/ingestion/`): compliant, agreed by two Claude lanes. Both new shared modules are language-free; `TRANSPARENT_RECEIVER_WRAPPERS`' `// TypeScript 'svc!'` annotation is the same node-type-comment idiom `call-analysis.ts` already used at base (`:8-13`, `:67-73`, `:358-364`) — not `if (language === …)` branching.
- **Fixture naming**: the PR actively *fixes* a prior instance of the `<lang>-*` prefix coverage defect (`typescript-inline-constructor-receiver/legacy/*.js` → `javascript-inline-constructor-receiver/`), the exact class flagged in `f4387d7f`.
- **GUARDRAILS non-negotiable 6** (never `terminate()` a worker mid-N-API): not implicated; no worker-pool file in the diff. No `any`/`as any`/`@ts-expect-error` introduced. Write-time *and* load-time `receiverChain` validation is a deliberate application of a previously-learned lesson, not redundancy.
- **`#2744` wiring is live end-to-end**, not dead scaffolding: recorder → `summarizeUnresolvedReceivers` → `RepoMeta` → `LocalBackend.unresolvedReceiverBoundaries` → `epistemicFrom` → `impact()`/`context()`. Every exported symbol has a real consumer.

---

### Headline findings

#### P0 — the fold FABRICATES a CALLS edge: its base lookup is more permissive than the cascade's
`gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts:297-301` · `[reproduced]`
Claude ce-adversarial, **independently verified by the coordinator with a fold-revert A/B**

`foldReceiverChain` resolves the chain's base by handing it back to
`resolveCompoundReceiverClass`, which inherits the permissive bare-identifier path ending in
`return findClassBindingInScope(inScope, workingText, scopes)` (`:470`) — reached **even when a
receiver typeBinding for that name exists but resolves to no class**. The text cascade's
dotted-chain head is deliberately narrower (`:674-676` consults the class lookup only when
`headType` is `undefined`). So a local binding that SHADOWS a class name is typed as that class.

Trigger — a local `const` shadowing an imported class:

```ts
import { Config, make } from './models';   // class Config { db: Database }
export function entry(): void {
  const Config = make(1);                  // a number, NOT a Config
  Config.db.query();
}
```

A/B on the real pipeline, reverting only the fold wiring:

```
WITH fold:    ["entry -> query", "entry -> make", "Config -> Database"]   <- FABRICATED
WITHOUT fold: ["entry -> make", "Config -> Database"]                     <- correct
```

`entry -> query` is a confident edge to `Database.query` that does not exist in the source. This
is a direct violation of the series' own requirement R4 ("no net wrong edges") and of the design
rule stated throughout the PR that a failed fold must fall through rather than guess. A missing
edge is recoverable; this one is not — it is wrong and confident.

Scope is precise: the call-shaped spelling `Config2.getDb().query()` fabricates in BOTH arms, so
the regression is specific to the dotted/mixed shape the fold newly takes over.

Fix: make the fold's base resolution as narrow as the cascade's head — resolve the base through
the receiver typeBinding only, and decline when the binding exists but names no class, rather than
falling through to a bare class-name lookup.

#### P0 — `impact()` emits a JavaScript function object as a call count, for any prototype-named symbol
`gitnexus/src/mcp/local/local-backend.ts:5718` · `[reproduced]` · Claude ce-correctness (single lane, verified by coordinator)

```ts
const sites = summary?.counts?.[symName];
if (sites === undefined || sites <= 0) return [];
```

`counts` is built by `Object.fromEntries(...)`, so it inherits `Object.prototype`. For any symbol
named `constructor`, `toString`, `valueOf`, `hasOwnProperty`, `isPrototypeOf`,
`propertyIsEnumerable`, `toLocaleString` or `__proto__`, the lookup returns a **Function**, not
`undefined`. The guard then fails to catch it: `Number(Object)` is `NaN`, and `NaN <= 0` is
`false`, so execution falls through to the message.

Trigger: `impact({target: "constructor"})` against any index with ≥1 recorded drop. Reproduced:

```
constructor    guard-returns-empty: false  EMITS => function Object() { [native code] } call sites invoking `constructor`
toString       guard-returns-empty: false  EMITS => function toString() { [native code] } call sites invoking `toString`
hasOwnProperty guard-returns-empty: false  EMITS => function hasOwnProperty() { [native code] call sites invoking `hasOwnProperty`
save           guard-returns-empty: false  EMITS => 3 call sites invoking `save`      (correct)
absent         guard-returns-empty: true                                              (correct)
```

`constructor` is an extremely common symbol name in a code graph, so this is reachable in ordinary
use, and it also flips `epistemic` to `lower-bound` on a symbol that may have had no drops at all.

Fix: use `Object.hasOwn(summary.counts, symName)` before indexing, or build `counts` with
`Object.create(null)`, or narrow with `typeof sites !== 'number'`.

Note: this line arrives from the #2744 commit (`90d13bdd`) carried in this PR, not from the
receiver-chain series — but it is an added line in this diff and ships with it.

#### P1 — Both schema constants are bumped *before* the final capture/edge divergence `[reproduced]`
`gitnexus/src/storage/parse-cache.ts:135`

Commit order on this branch (oldest first):

```
 #9  cc4e747cd  SCHEMA_BUMP 31 -> 32                    (TypeScript captures ONLY)
#11  15cbe8002  INCREMENTAL_SCHEMA_VERSION 25 -> 26     (edge semantics, TS only at that point)
#13  c98668257  rolls capture emission out to 13 MORE languages   <-- NO further bump
#14  d5e0dfd5c  rebaselines goldens/fingerprints for those 13
```

`git log -S` confirms `cc4e747cd` is the only commit that sets 32 and `15cbe8002` the only one that sets 26; `git log -- 'languages/*/captures.ts'` confirms `c98668257` changed capture output for the other 13 languages afterwards. At HEAD both constants are still 32/26. `c98668257` moves **edges**, not just captures — C++ `svc->getUser()->save()` goes INVISIBLE-GAP → RESOLVES in the committed baseline.

**Trigger:** anyone who built at #9–#12 holds a parse-cache stamped `32+<pkg>` and an index stamped `26`. At HEAD both are still 32/26, so the reuse gates (`===`) treat them as **current** while every non-TypeScript file keeps pre-rollout captures and pre-rollout receiver-bound edges. Squash-merging collapses base→HEAD into a single 31→32 jump and hides this entirely; merging the branch as-is, or anyone tracking the branch, does not.

This contradicts the series' own stated rationale in the `// v32:` comment ("Lands HERE, at the first commit where the capture SET actually diverges") — the set diverged **again** at #13 under the same version. The same comment also scopes v32 to "TypeScript call matches", which is no longer true at HEAD. This defect class has shipped at least four times, most recently #2740.

**Fix:** bump both to 33/27 at the tip, **or** state the squash-merge requirement in the PR description and enforce it on merge.

> **Recorded disagreement.** This was found **independently by both engines** — Codex B3 adversarial (as `[medium]`) and Claude ce-adversarial (ADV-4, P2), by the same git-log mechanism, was verified first-hand by the coordinator, and was raised to P1. **Two** Claude lanes — project-standards (§3a) and correctness (Refutations §6) — independently judged the same placement **COMPLIANT**, reasoning that the bump lands where the capture set diverges. That reasoning is correct for base→HEAD and wrong for the intermediate builds — the Claude lane never modelled a reader who built mid-series. The verification (git history + HEAD file state) settles it in Codex's favour. Single-lane, single-engine, but concretely reproduced.

#### P2 — A negative test is vacuous `[reproduced]`
`gitnexus/test/integration/resolvers/typescript.test.ts:3496`

```ts
expect(calls.filter((c) => c.target === 'noSuchMember')).toHaveLength(0);
```

`grep -rn noSuchMember gitnexus/` returns **exactly one hit — the assertion itself**. The fixture (`models.ts` / `main.ts`, written inline at `:3400-3440`) never defines, calls, or references `noSuchMember`. The test cannot fail, so it proves nothing about its own caption ("A broken middle step must produce NO edge, never a wrong one") — the R4 property it is the only guard for.
**Fix:** give the fixture a chain whose middle step genuinely fails (e.g. `svc.getUser().missing.save()`) and assert no CALLS edge for `save` from that entry.
*Codex B2 testing found it; coordinator verified. Note the cross-engine split: Claude's testing lane listed "no-such-member" among the PR's "genuinely useful negative controls" — the grep settles it.*

#### P2 — "resolves a mixed call/field chain through the fold" does not exercise the fold `[code-read]`
`gitnexus/test/integration/resolvers/typescript.test.ts:3475`

Fixture is `svc.getUser().address.persist()` — plain dotted call→field→call, no `?.`, `!`, or `<T>`, i.e. none of the punctuation the PR's own docs say the text cascade cannot parse. Verified: the identical shape `svc.getUser().address.save()` is covered by a **pre-existing, unmodified** test at base (`bc76ba2f2:…/typescript.test.ts:2189`), and `compound-receiver.ts` is **142 insertions / 0 deletions** — nothing was removed from the cascade. Since `resolveCompoundReceiverClass` tries the fold first and falls through on `undefined`, this test passes byte-for-byte if `foldReceiverChain` were deleted, always returned `undefined`, or were never wired. Its three siblings (`viaOptionalChain`, `viaTypeArgs`, `viaNonNull`) *are* discriminating; this one is not.
**Fix:** either give the fixture a new-only spelling (`svc?.getUser().address.persist()`), or rename it to what it actually is — a "the fold doesn't regress the cascade" control.

---

### Lower-priority

- **P3 `compound-receiver.ts:279`** `[code-read]` — JSDoc reads `NOT WIRED into any resolution path yet — see U10`, but the same file wires the fold 71 lines below at `:350-355`, and `receiver-bound-calls.ts` feeds `site.receiverChain` into it at two sites. A reader who trusts the comment will skip auditing the live path. `receiver-chain-fold.test.ts:8` carries the same stale framing. **Strongest corroboration in this review: 3 lanes, 2 engines** (Claude risk-architect, Claude ce-maintainability, Codex ce).
- **P3 `test/unit/call-summary-schema-version.test.ts:181`** `[reproduced]` — the diff flipped this assertion from `.toBe(true)` to `.toBe(false)` (correct) but left the now-false comment `// The current stamp passes the gate (incremental top-up eligible).` directly above it, and added a *second* `expect(passesReuseGate(24)).toBe(false)` five lines down under the correct explanation. Same input asserted twice with contradictory comments, in a file whose entire value is its per-version narrative. Coverage itself is intact (24→25→26 all gated). 3 Claude lanes, one engine.
- **P3 `bench/receiver-resolution/measure.mjs:415`** `[code-read]` — `--check` compares only `projection()`, which returns `shapeArm` + `countArm`. `perfArm` is computed only under an explicit `--perf N` flag (`:459-463`) and never asserted. The CI step named "Receiver-resolution drop guards" gates correctness drift but **not** the wall-clock/RSS/bytes-per-site thresholds recorded in `BASELINE.md` — those are a one-time measurement, not a guard. Say so plainly in BASELINE.md or add a budgeted assertion. *(Codex B2 performance; coordinator verified.)*
- **P3 `local-backend.ts:5721`** `[code-read]` — the epistemic boundary note is worded `"${sites} call site(s) invoking \`${symName}\` …were dropped"`, but `summarizeUnresolvedReceivers` (`unresolved-receivers.ts:47-52`) filters only on `reason === 'receiver-unresolved'` and **ignores `siteKind`**, while the recorder deliberately fires on any compound receiver regardless of reference kind (`receiver-bound-calls.ts:1338+`, whose own comment says "property reads and writes with a compound receiver are recorded here too. Carry the kind so a consumer can separate a dropped CALL from a dropped property access"). No consumer separates them. Net: a symbol whose only drops were property accesses can flip `impact()` to `epistemic: 'lower-bound'` with a message that says "call sites". Direction is conservative (over-warn), and `resolution-outcome.ts` documents the choice as intentional — but the user-facing string is factually wrong. *(Codex B2 maintainability.)*
- **P3 `receiver-chain-codec.ts:76`** `[code-read]` — `isEncodableSegment`'s `!/\s/.test(value)` does not match zero-width characters (U+200B, U+200C/D, U+200E/F). An identifier containing one would encode, persist, decode, then match no binding — silent degradation to the text cascade. Pre-existing analyzer-wide gap, no code-execution path.
- **P3 `receiver-chain-codec.ts:119`** `[code-read]` — `parts[1] ?? ''` is unreachable; the `parts.length < 3` guard at `:116` already proves `parts[1]` exists.
- **P3 `parsedfile-store.ts:382`** `[code-read]` — the doc comment claims "Same untrusted-boundary contract as `sanitizeCallableFlowSites`", but that function validates every element with a real type predicate while this one checks only the `receiverChain` sub-field and, in the common case, returns the whole `unknown[]` cast to `readonly ReferenceSite[]` with no structural validation of `name`/`kind`/`atRange`. Still a net improvement — `referenceSites` had **no** sanitization at base — so the defect is the overclaiming comment, not a regression. Codex separately flagged the layering direction (storage now runtime-imports a codec from `core/ingestion`).
- **Body note, not a finding — `receiver-chain-codec.ts:99` truncation-marker API is architecturally dead.** No caller passes `options.truncated`; `synthesizeReceiverChainCapture` always calls `encodeReceiverChain(base, steps)` with no options, and `extractMixedChain` returns `baseReceiverName: undefined` when it stops early, which is filtered one line earlier. Lanes disagree on whether this is a finding: the security lane called it dead API (P3), the maintainability lane called it deliberate, documented defence-in-depth at a durable-store boundary with its own unit coverage. The code's own comment supports the latter. Not posting; noting the split.
- **Not a finding — `compound-receiver.ts` crosses 1000 lines** (939 at base → 1081 at HEAD). One Claude lane lodged this as P1 on a file-size-regression rule and proposed extracting `foldReceiverChain`/`typeOfMemberOnClass` into `receiver-chain-fold.ts`. No corroboration, no correctness impact; recording it as an optional follow-up, not a merge concern.
- **Body note — the PHP `INVISIBLE-GAP` is a PRE-EXISTING product gap, not a regression.** Codex's adversarial lane rated PHP `$svc->getUser()->save()` `[high]` and recommended blocking the PHP receiver-chain path before shipping. **That attribution is wrong**, settled by coordinator A/B `[reproduced]`: reverting *only* the fold wiring (`compound-receiver.ts` + `receiver-bound-calls.ts`) to the pre-U10 commit, rebuilding, and re-running the same typed-PHP fixture gives byte-identical output — `{"calls":["arrowCallChain->getUser","getUser->User"],"drops":[],"state":"INVISIBLE-GAP"}` with and without the fold. A failed fold falls through to the cascade, so `compoundReceiverUnresolved = currentClass === undefined` is computed from the same cascade result as before; this PR cannot change PHP's drop-recording for this shape. The earlier VISIBLE→INVISIBLE movement in the bench came from typing the *fixture* parameter. The underlying gap is real and worth a separate issue — a PHP chained call can be lost while `epistemic` still reports `exact`, which undercuts the very feature #2744 ships — but it **does not block this PR**.

---

- **P2 `gitnexus/src/mcp/local/local-backend.ts:5718`** `[reproduced]` — Claude ce-adversarial ADV-3.
  The epistemic lookup is keyed by BARE MEMBER NAME, so stdlib-method drops poison every same-named
  user symbol. Measured on one subdirectory of GitNexus's own source, the persisted counts map holds
  `get:86, find:11, map:7, filter:7, has:6` — names whose counts come largely from unindexed stdlib
  calls. Every user-defined `get`/`find`/`map` method therefore reports `epistemic: lower-bound`
  permanently. Distinct from the P0 above (which returns a Function rather than a count). Arguably
  inherent to the #2744 design — a dropped site's callee is unknown, so the member name is the only
  available key — but the persisted artifact should at least record that the count is name-keyed and
  not target-attributed.

### Test gaps

1. **Behavioral proof exists for exactly 1 of 14 rolled-out languages.** Only `typescript.test.ts` adds receiver-chain resolution scenarios (`javascript.test.ts`'s +28 lines are the unrelated #2708 fixture rename). All 14 emitters now synthesize `@reference.receiver-chain` and all 14 `bench/scope-capture` fingerprints moved — but a fingerprint proves *something* changed, not that the emitted chain is *correct*, and says nothing about whether the fold produces a right (or any) CALLS edge. Only 7 languages even have a `*-captures-golden` digest that changed. Codex independently found a concrete instance of this blind spot: **Dart creates some references in a post-query tree walk that misses the new field for member calls** — no test would catch it. Suggested: one integration test per language with a genuinely new-only shape (C#/Kotlin `?.`, PHP `?->`, Ruby `&.`, Python chains).
2. **`hoistTypeBindingsToModule: false` — the off-branch — is never exercised.** `receiver-chain-fold.test.ts:77` hard-codes `true` in the shared `fold()` helper used by every case. Six languages leave the flag unset (`c`, `cobol`, `dart`, `python`, `ruby`, `swift`), five of which are wired to emit chains and therefore reach `typeOfMemberOnClass`'s `=== true` gate in production. A regression relaxing that gate to a truthy check would silently start minting cross-scope member matches for Python/Ruby/Dart/Swift/C and no test would fail.
3. **No adversarial/property test for `decodeReceiverChain`.** The "total function" claim was verified by reading, not by a table test over all-pipe 512 B, null bytes, surrogate pairs, ZWS-in-name, over-limit bytes. Same for `sanitizeReceiverChains` with `referenceSites: null` (file-reject path) and `[{ receiverChain: null }]`.
4. **`sanitizeReceiverChains`' per-FIELD (not per-site) semantics are tested only with single-element arrays**, so the `dropped > 0` `.map()` branch is never shown to leave a *valid* sibling's chain intact. Mirrors a pre-existing gap in the sibling `sanitizeCallableFlowSites` test.
5. **`receiver-chain-wiring.test.ts` is a `String.includes` inventory, not a wiring check.** It correctly catches "a language stopped calling the helper at all", but a language that keeps 1 of its 3 call sites still reports as fully wired — the realistic partial-de-wiring case. The authors state the trade-off in the file; worth one added sentence so a reviewer doesn't read it as completeness proof.
6. **No positive boundary test** that exactly `MAX_CHAIN_DEPTH` steps round-trips (only `MAX_CHAIN_DEPTH + 1` refusal is pinned).

---

### CI status

`quality`, `typecheck`, `lint`, `format`, and `node-floor` all **PASS**. Vercel fails on an auth/authorization issue **unrelated to this code**. "Build & Push gitnexus" is still **pending**. The new "Receiver-resolution drop guards" step is convention-correct (matches sibling guard steps byte-for-byte in shape) but is **not build-free** (~2m15s, needs `dist/`) and lands in a `benchmarks` job with `timeout-minutes: 25` already carrying 8+ sequential steps. The author explicitly deferred raising the timeout pending the first real CI run — reasonable, but unverified.

---

### Coverage limits — read these before treating this as a clean bill of health

- **Codex's swarm leg (B1) is DEAD** — it hit the workspace spend cap after loading the skill docs and produced **zero findings**. Cross-engine corroboration is therefore available only where the Codex ce/adversarial legs happened to look.
- **The Claude test/CI lane never completed** (partial return, twice). Test *quality* was covered in depth by other lanes, but the specific CI-workflow-timing question above is **OPEN** — no lane answered it.
- **The Claude test/CI lane never completed** (partial return, twice), so the new bench step's CI timing was never assessed against the `benchmarks` job's 25-minute budget.
- **The `unresolvedReceiverMembers` incremental question is UNRESOLVED.** `run-analyze.ts:2402` and `:2569` both write the summary from `pipelineResult.resolutionOutcomes` on the full **and** incremental paths. If incremental resolution sees only changed files, the summary under-counts after every top-up and `epistemic` silently reverts to `exact` for unchanged files — the exact inverse of the feature's purpose. **`pipeline.ts` is not in this diff** (confirmed), so this cannot be settled from the diff alone. Worth one direct check before merge.
- Two Claude lanes disagree with prior session context on the `INCREMENTAL_SCHEMA_VERSION` delta: base `bc76ba2f` was at **24**, the #2744 commit inside this PR took it to 25, and the receiver-chain commit to 26 — so it is **24 → 26** in PR terms, not 25 → 26.

---

*Automated multi-tool digest (Claude swarm + Claude ce personas + Codex/GPT-5.5 ce + Codex adversarial, merged and spot-verified against the worktree). Findings are tagged `[reproduced]` where a trigger was executed and `[code-read]` where it was traced statically. Verify before acting.*

### Verdict

- **Branch hygiene:** clean — no merge-from-main commits, no unrelated churn, `git diff --check` clean, no bidi/hidden Unicode introduced.
- **Merge state:** `MERGEABLE` / `BEHIND` (new commits on `main` since the branch point).
- **Final verdict: NOT PRODUCTION-READY.** Two P0s block: a fabricated CALLS edge from the fold's base lookup, and `impact()` emitting a Function as a call count for prototype-named symbols. The P1 schema-bump ordering needs either a re-bump at the tip or an explicit squash-merge commitment recorded on the PR.

gitnexus/test/unit/call-summary-schema-version.test.ts
```diff
@@ -178,6 +178,19 @@ describe('CALL_SUMMARY incremental reuse gate (U-C5)', () => {
     // A pre-v24 (v23) index predates
     expect(passesReuseGate(23)).toBe(false);
     // The current stamp passes the gate (incremental top-up eligible).
-    expect(passesReuseGate(24)).toBe(true);
+    expect(passesReuseGate(24)).toBe(false);
```
magyargergo: **P3** — Comment says "current stamp passes the gate" above an assertion of false; passesReuseGate(24) is then asserted twice

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/bench/receiver-resolution/measure.mjs
```diff
@@ -0,0 +1,483 @@
...
+];
+
+/**
+ * The gated projection: shape states per language, plus the call-drop counts.
+ * Deliberately EXACT rather than budgeted — two consecutive runs are
+ * byte-identical, so a range would only hide real movement. Adding fixtures
+ * moves these numbers and requires a rebaseline; that treadmill is the accepted
+ * cost of the guard, the same trade the scope-capture bench already makes.
+ */
+function projection(output) {
```
magyargergo, line 424: **P3** — --check gates only projection() (shapeArm+countArm); perfArm is never asserted, so BASELINE.md perf numbers are unguarded

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -5676,10 +5694,39 @@ export class LocalBackend {
...
+      if (sites === undefined || sites <= 0) return [];
+      return [
+        `${sites} call ${sites === 1 ? 'site' : 'sites'} invoking \`${symName}\` ${
```
magyargergo, line 5725: **P3** — Message says "call sites" but summarizeUnresolvedReceivers ignores siteKind, so property reads/writes are counted in

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/src/core/ingestion/utils/receiver-chain-codec.ts
```diff
@@ -0,0 +1,143 @@
...
+
+/** A segment is safe when it contains no structural character. Identifiers in
+ *  every supported language satisfy this; anything that does not is a payload
+ *  we refuse to mint rather than escape. */
+function isEncodableSegment(value: string): boolean {
+  return (
+    value.length > 0 &&
+    !value.includes(SEPARATOR) &&
+    !value.includes(TRUNCATED) &&
+    !/\s/.test(value)
```
magyargergo: **P3** — isEncodableSegment's \s check misses zero-width chars (U+200B/C/D, U+200E/F) — encodes, persists, then matches no binding

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/src/core/ingestion/utils/receiver-chain-codec.ts
```diff
@@ -0,0 +1,143 @@
...
+ */
+export function decodeReceiverChain(value: unknown): DecodedReceiverChain | undefined {
+  if (typeof value !== 'string') return undefined;
+  if (value.length === 0 || value.length > MAX_RECEIVER_CHAIN_BYTES) return undefined;
+
+  const parts = value.split(SEPARATOR);
+  if (parts.length < 3) return undefined; // version + base + at least one step
+  if (parts[0] !== VERSION) return undefined;
+
+  const baseReceiverName = parts[1] ?? '';
```
magyargergo: **P3** — parts[1] ?? '' is unreachable; the parts.length < 3 guard at :116 already proves parts[1] exists

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/src/storage/parsedfile-store.ts
```diff
@@ -317,6 +349,47 @@ function sanitizeCallableFlowSites(
...
+  let dropped = 0;
+  for (const site of value) {
+    if (
+      isRecord(site) &&
+      site.receiverChain !== undefined &&
+      !isValidReceiverChain(site.receiverChain)
+    )
+      dropped++;
+  }
+  if (dropped === 0) return { sites: value as readonly ReferenceSite[], dropped: 0 };
```
magyargergo, line 390: **P3** — Doc claims parity with sanitizeCallableFlowSites, but the dropped===0 path casts the whole unknown[] with no structural validation

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/test/unit/scope-resolution/receiver-chain-fold.test.ts
```diff
@@ -0,0 +1,133 @@
...
+
+function fold(encoded: string) {
+  const decoded = decodeReceiverChain(encoded);
+  expect(decoded).toBeDefined();
+  // The language's own contract flags, exactly as the resolver pass would
+  // pass them. TypeScript hoists method return-type bindings out of the class
+  // body, so the fold needs that flag to find any of them.
+  return foldReceiverChain(decoded!, ctx.inScope, ctx.scopes, ctx.index, {
+    fieldFallback: false,
+    hoistTypeBindingsToModule: true,
```
magyargergo, line 90: **P3** — Every fold test hard-codes hoistTypeBindingsToModule: true; the false branch five wired languages actually run is untested

_Surfaced by the multi-engine tri-review; verified against this diff._

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
magyargergo: **Fixed** in . Root cause confirmed: the bare-identifier branch falls through to  even when a receiver typeBinding exists but names no class, while the dotted-chain head consults that namespace only when there is *no* binding. Added an opt-in  so the fold obeys the cascade's rule; the flag is **off** for the cascade, so no existing edge moves.\n\nA/B after the fix now matches the no-fold control exactly ( gone), and the shape arm is unchanged \u2014 every previously-resolving shape still resolves, so the fix cost zero legitimate resolutions. Pinned by an integration test verified to fail without the fix.

gitnexus/src/mcp/local/local-backend.ts
magyargergo: **Fixed** in . Reproduced first:  is false, so the guard never fired and the message interpolated the function. The lookup is now a single exported prototype-safe helper, , using  plus a  /  check \u2014 so the rule lives in one place rather than an inline index expression. Regression test covers , , , , , ,  and , and asserts a genuinely recorded name still reads back.

gitnexus/src/storage/parse-cache.ts
magyargergo: **Fixed** in . Verified the ordering from git history \u2014  32 at the TypeScript-only commit,  26 before the fold's semantics were final, then the 13-language rollout changed captures *and* edges under the same numbers. Re-bumped to **33 / 27** so each tracks the final state rather than the first divergence, with rationale in both ladders. The reuse-gate test now asserts  \u2014 an index stamped by an intermediate build carries both pre-rollout edges and the fabricated ones from the P0 above, so it must not be reused.

gitnexus/test/integration/resolvers/typescript.test.ts
magyargergo: **Fixed** in . Confirmed vacuous \u2014 0 over the file returned exactly 1, the assertion itself. Replaced with a chain whose **middle** step genuinely fails (), asserting the tail does not bind *while the head still does* \u2014  legitimately resolves, so asserting zero edges from that function would itself have been wrong.

gitnexus/test/integration/resolvers/typescript.test.ts
magyargergo: **Fixed** in . Confirmed: the shape was already covered by the pre-existing cascade, so the test passed with the fold deleted. It now uses an optional-chained spelling the text cascade cannot parse, and I verified it **fails with the fold reverted** \u2014 which is the property it was missing.

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
magyargergo: **Fixed** in . The JSDoc now states where it is actually called from (, ahead of the text cascade, gated on  and a present ).

gitnexus/test/unit/call-summary-schema-version.test.ts
magyargergo: **Fixed** in . Removed the orphaned comment and the duplicate  assertion. The ladder now reads 23/24/25/26 \u2192 , 27 \u2192 , each with a comment describing the version it actually covers.

gitnexus/bench/receiver-resolution/measure.mjs
magyargergo: **Addressed as documentation** in , deliberately not as a new gate. Asserting the perf arm in  would compare wall-clock and RSS against numbers recorded on a different machine and fail on noise \u2014 the figures are only meaningful as an A/B against a control build, which  cannot construct.  now carries an explicit note that the BASELINE.md perf numbers are a point-in-time measurement, **not** a CI guard, so a green  is not misread as evidence that performance held.

gitnexus/src/mcp/local/local-backend.ts
magyargergo: **Fixed** in .  now filters on the  this series added for exactly this purpose, so only CALL sites are counted. On the fixture corpus that is 25 of 124 drops; names whose call count was never short no longer report , and the wording is now literally true. A missing  still counts as a call \u2014 erring toward  is the safe direction for an epistemic signal.

gitnexus/src/core/ingestion/utils/receiver-chain-codec.ts
magyargergo: **Fixed** in .  now rejects . Test covers U+200B/C/D/E/F and U+FEFF in both the base name and a step name \u2014 a zero-width character would otherwise encode and persist cleanly, then match no binding at resolution time, which is a silent unexplained miss and a trojan-source vector.

gitnexus/src/core/ingestion/utils/receiver-chain-codec.ts
magyargergo: **Fixed** in . Replaced with  plus a comment noting the  guard above already proves the element exists.

gitnexus/src/storage/parsedfile-store.ts
magyargergo: **Fixed** in . The doc no longer claims parity. It now states plainly that this is **narrower** than  \u2014 it inspects only the  sub-field and, when nothing was stripped, returns the original array without structurally validating // \u2014 that this is still a strict improvement over the no-sanitation status quo, and that a future change needing per-site validation must add it rather than assume it.

gitnexus/test/unit/scope-resolution/receiver-chain-fold.test.ts
magyargergo: **Fixed** in . Added a  helper that takes contract-flag overrides, and a test asserting the same chain resolves with  and returns  with it  \u2014 so the branch is pinned by opposite outcomes, not by the happy path. That off-branch is what six of the wired languages actually run.

PR #2762, opened 2026-07-30
fix(ingestion): join multi-line closure bindings on initializer startLine (#2735)

### LINKED ISSUE #2735: Multi-line closure bindings get no CALLS edge: the two query channels anchor differently and are joined on line only
## Summary

A closure binding whose literal starts on the line **after** its binding gets no call-source edge. The closure's node exists, but has zero outgoing edges. Affects every language with a closure-binding rule (PHP, Rust, Kotlin, Ruby, Dart).

This is the residual half of a defect found in the review of #2718. That PR fixed the *dangerous* half — the same miss previously caused attribution to climb to the enclosing function, emitting a CALLS edge present nowhere in the source. It now fails closed instead. The **correct** edge is still missing.

## Reproduction

```php
<?php
function target($x) { return $x; }
function outer() {
  $single = function ($x) { return target($x); };
  $multi =
    function ($x) { return target($x); };
  return 1;
}
```

All CALLS edges after `analyze --index-only`:

```
Function:src/a.php:outer.$single@3:2 -> Function:src/a.php:target
```

`Function:src/a.php:outer.$multi@4:2` exists as a node and is orphaned as a source. The same shape reproduces in Rust with a wrapped closure:

```rust
let handler =
    || target(1);
```

## Root cause

The two query channels anchor on **different nodes, by design**:

- graph-node channel (`tree-sitter-queries.ts`) anchors `@definition.function` on the **outer** wrapper (`assignment_expression` / `let_declaration` / …)
- scope-resolution channel (`languages/<lang>/query.ts`) anchors `@declaration.function` on the **inner** closure, so its range aligns with `@scope.function`

`resolveDefGraphId` bridges the two with a **line-only** position key (`ids.ts`, `positionKey(filePath, type, line - 1, simple)`). On one line both anchors share a row and the join matches. Split across lines they do not, the join misses, and the fail-closed guard correctly declines to guess.

Each channel's anchor is correct for its own purpose, so neither is simply "wrong" — the join is what cannot express the split.

## Possible directions

1. **Share one identity across both channels** — have the graph-node capture and the scope-resolution declaration agree on an anchor, or emit an explicit cross-channel key.
2. **Bridge on the initializer position** rather than the declaration line.
3. **Make the position key column-aware or range-aware** rather than line-only.

Option 3 may also address the related unverified concern from the same review: with `Property`/`Static` now position-qualified, a class member and a same-named local on one compact line can collide on the line-only key.

## Test gap

No test in any of the five languages covers a multi-line closure binding — the single condition that decides whether the feature works for a given binding. #2718 adds one pinning the *fail-closed* behaviour (no fabricated edge); a fix here should flip it to assert the real edge.

azizur100389:
## Summary
- Fixes #2735: multi-line closure bindings now emit the real CALLS edge instead of failing closed with an orphaned Function node.
- Graph-node queries still capture the outer binding wrapper; scope-resolution still anchors on the inner callable. The line-only position join now matches because Function/Method/Constructor `startLine` follows the initializer via `boundCallablePositionNode`, while node ids stay on the wrapper (`localIdentity`).
- Covers PHP, Rust, TypeScript, Kotlin, Ruby, and Dart multi-line fixtures; Ruby `lambda do` needs an explicit `call` + `block` unwrap.
- Bumps `SCHEMA_BUMP` 32→33 and `INCREMENTAL_SCHEMA_VERSION` 25→26 so warm cache / incremental reuse cannot keep dropping the edge.

## Notes
- `endLine` remains the wrapper span; only `startLine` moves to the body. That can shift PDG `pdg_query` BasicBlock windows off the binding-name line for multi-line closures — intentional (the node owns the body).
- Fail-closed behavior from #2699 is preserved: the enclosing function never gains a fabricated CALLS edge to `target`.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run test/integration/closure-review-findings.test.ts test/unit/callable-id-lockstep.test.ts test/unit/call-summary-schema-version.test.ts test/unit/incremental-parse-cache.test.ts`
- [x] Prettier and ESLint on all changed files
- [x] CI full suite: 15,842 tests passed across three platforms


azizur100389:
Addressed the format finding in d0d620c7. Prettier, typecheck, ESLint, and the 55 focused tests pass; the full GitHub test matrix was already green.

gitnexus/src/core/ingestion/workers/callable-id.ts
```diff
@@ -21,6 +21,131 @@
 
 import type { SyntaxNode } from '../utils/ast-helpers.js';
```
magyargergo, line 23: Could we make this language agnostic please? 🙏

magyargergo (changes requested):

gitnexus/src/core/ingestion/workers/callable-id.ts
azizur100389: Done. The shared helper now derives the callable row from canonical ParsedFile.localDefs metadata (label, name, range, and definition position), with no grammar-node or language-specific branches. Focused multi-language closure, cache, and schema tests pass.

magyargergo (approved):

Issue #1939, opened 2026-05-30
Follow-up to #819: `npx gitnexus` still crashes with "node.target is null" on npm 11.x — clean cache doesn't help


ProductOfAmerica:
## Summary

#819 (closed 2026-05-22 without a documented resolution) reported `npx gitnexus analyze` crashing with:

```
npm error Cannot destructure property 'package' of 'node.target' as it is null.
```

The same crash reproduces on gitnexus 1.6.5 / Node 24 / npm 11.x / Windows 11 against a fully cleaned npm cache. I can't reopen #819 (no permissions), so filing this as a follow-up with a different-environment repro and root-cause analysis.

## Repro (clean state)

```powershell
# Kill any node procs that might hold cache locks
Get-Process node -ErrorAction SilentlyContinue | Stop-Process -Force

# Nuke the _npx cache entirely
Remove-Item -Recurse -Force "$env:LOCALAPPDATA\npm-cache\_npx" -ErrorAction SilentlyContinue
npm cache clean --force

# Try gitnexus
cd <any git repo>
npx gitnexus analyze --embeddings
```

Output:

```
Need to install the following packages:
gitnexus@1.6.5
Ok to proceed? (y) y

npm error Cannot destructure property 'package' of 'node.target' as it is null.
npm error A complete log of this run can be found in: ...
```

That the same error reproduces on a freshly-cleaned cache rules out the "stale state" theory and points at the live install graph.

## Root cause (not gitnexus-specific, but gitnexus is the trigger)

This is an npm 11.x regression in `arborist` handling of platform-conditional `optionalDependencies`. The chain:

1. gitnexus depends on `onnxruntime-web` (via `--embeddings`).
2. `onnxruntime-web` publishes per-platform native binaries as `optionalDependencies` with `os` / `cpu` filters (`onnxruntime-node-win32-x64`, `onnxruntime-node-darwin-arm64`, etc.).
3. npm 11's arborist correctly *skips* the non-matching binaries, but instead of removing them from the install graph, it leaves them in with `node.target = null`.
4. A subsequent arborist pass iterates the graph and does `node.target.package` on the null targets — boom.

Same null-deref hits other native-binary trees (sharp, esbuild, swc, lightningcss variants) on npm 11.x. There are several open issues against `npm/cli` matching this pattern; this is upstream of gitnexus.

## Why this matters for gitnexus

Even though the actual bug is in npm 11.x, gitnexus is one of the most-affected packages because:

[...]

Issue #1706, opened 2026-05-19
`--no-stats` does not strip index-summary symbol counts from AGENTS.md / CLAUDE.md


tkrisztian95:
### Summary

`gitnexus analyze --no-stats` is documented as *"Omit volatile file/symbol counts from AGENTS.md and CLAUDE.md"*, but it does not strip the index-summary sentence inside the `<!-- gitnexus:start -->` block. That line still carries volatile counts, and because it's regenerated on every `analyze`, it causes merge conflicts when `AGENTS.md` / `CLAUDE.md` are committed and multiple developers run `analyze`.

### Environment

- gitnexus: 1.6.4
- node: v24.11.1
- OS: macOS (darwin)

### Steps to reproduce

1. In a repo with a committed `AGENTS.md` containing a gitnexus block, run:
   ```
   gitnexus analyze --no-stats
   ```
2. Inspect the regenerated block in `AGENTS.md` / `CLAUDE.md`.

### Expected

With `--no-stats`, the index-summary line is emitted without volatile counts, e.g.:

> This project is indexed by GitNexus as **<name>**. Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.

### Actual

The counts are still present and change on every run:

> This project is indexed by GitNexus as **<name>** (9309 symbols, 15221 relationships, 292 execution flows). Use the GitNexus MCP tools to ...

Only the prettier-vs-gitnexus table formatting differs; the counts line is untouched by `--no-stats`.

### Why it matters

When `AGENTS.md` / `CLAUDE.md` are committed (the default, since `analyze` generates them), every developer's `analyze` rewrites this exact line with different numbers. Two branches that both ran `analyze` then conflict on it — a recurring, no-value merge conflict. `--skip-agents-md` avoids it by skipping the file entirely, but that also drops legitimate guidance updates.

### Suggested fix

Make `--no-stats` also omit the parenthetical counts from the index-summary sentence (symbols / relationships / execution flows) — or document that the flag does not cover that line and point to `--skip-agents-md` instead.

ToniChowBea:
I'd like to pick this up if there are no open PRs for it @tkrisztian95

magyargergo:
@oddFEELING go for it

magyargergo:
@oddFEELING did you manage to make any movement on this?

ToniChowBea:
Gave it a look, but the comment here kinda contradicts expected behaviour and it says its on purpose.

[...]

Issue #2441, opened 2026-07-12
Bug: `analyze` exits 0 after LadybugDB native load failure — silently does nothing while reporting success


alejandrodgb:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

When the LadybugDB native binary (`lbugjs.node`) fails to load, `gitnexus analyze` prints the load-error help text and then **exits with code 0**, having produced no index at all. Scripts, CI, hooks, and users all read exit 0 as "analysis succeeded," so the failure is invisible: on my machine the index silently stayed stale for weeks while every `npx gitnexus analyze` "succeeded."

### Context

Hit this on a MacPorts-based Mac (no Homebrew), where `@ladybugdb/core` 0.18.1's darwin-arm64 prebuilt fails to `dlopen` because it hardcodes Homebrew OpenSSL paths. That root cause is upstream and now filed against LadybugDB (LadybugDB/ladybug — "darwin-arm64 prebuilt lbugjs.node hardcodes Homebrew OpenSSL paths"). This report is about gitnexus's handling of the failure: a hard dependency failing to load should be a hard error.

The same silent-success applies to a repo with an existing index: `analyze` prints the dlopen error, exits 0, and the stale index just stays stale — `status` still reports its last state, so nothing looks wrong.

### Expected behavior

`analyze` exits non-zero when the graph backend cannot load (or, if a degraded mode is intentional, says so explicitly and still exits non-zero when no index was written). "Repository indexed successfully" and exit 0 should be impossible when no `.gitnexus/` was produced.

### Actual behavior

`analyze` prints the LadybugDB repair instructions, writes nothing (no `.gitnexus/` directory on a fresh repo), and exits 0.

### Steps to reproduce

On an Apple Silicon Mac **without** `/opt/homebrew` (any machine where `lbugjs.node` fails to dlopen reproduces it; a truncated/corrupted `lbugjs.node` also triggers the same path):

```sh
mkdir /tmp/gnx-test && cd /tmp/gnx-test && git init -q
printf 'export function add(a, b) { return a + b; }\n' > index.js
git add -A && git commit -qm init
npx gitnexus analyze; echo "exit: $?"
# → prints "LadybugDB native binary (lbugjs.node) exists but failed to load: dlopen ..."
# → exit: 0
ls .gitnexus
# → No such file or directory
```

### Environment

- gitnexus 1.6.9 (npx), @ladybugdb/core 0.18.1
[...]

PR #2765, opened 2026-07-31
feat(cli): add a bunx lane so bun-only machines can run gitnexus

drdave-flexnetos:
## What changed

GitNexus is unusable on a machine that has **bun but no Node package manager**. This adds a `bunx` lane end to end, and fixes the one thing that would otherwise make that lane resolve to a command that still fails.

Three self-contained commits — each stands alone if you'd rather take a subset:

| Commit | Change |
| --- | --- |
| `feat(cli)` | `bun` mode in the canonical resolver (`hooks/claude/resolve-analyze-cmd.cjs` + the byte-identical plugin copy) |
| `fix(lbug)` | Restore the prebuilt native binary when a package manager skipped the install script |
| `fix(ai-context)` | Generated bootstrap note names every install-free runner |

### The ladder assumed a Node toolchain

`resolveInvocationMode` went global `gitnexus` → `pnpm dlx` → `npx` → `pnpm dlx` → `npx`. With no npm, npx **or** pnpm on PATH, every rung fell through to the `npx` last resort — so both the emitted hint and the generated `.gitnexus/run.cjs` produced `npx gitnexus@latest analyze`, which exits with `npx: command not found`.

`bun` is added on two rungs:

- **npm 11+ with no pnpm to fall back on** — `bunx` is install-free and dodges the same arborist crash the pnpm rung exists for (#1939).
- **npm and pnpm both absent** — the dead end above.

**No existing outcome changes.** pnpm still wins on npm 11+, npx still wins on npm < 11, pnpm still wins over bunx when npm is absent. Each is pinned by a regression test. The bun PATH probe is lazy, so a machine with a Node toolchain pays no extra scan and the stale-index hook budget is untouched.

`bunx` takes no `--allow-build` equivalent — bun's `--trust` is a `bun add`/`bun install` flag that writes `trustedDependencies` into a *project* `package.json`, which a one-shot has none of — so the argv stays flag-free.

### …but a bunx lane alone would still be broken

Worth flagging, because it's the reason commit 2 exists. Bun skips lifecycle scripts for a `bunx` fetch, so `@ladybugdb/core`'s install script never copies `lbugjs.node` up from its per-platform sub-package:

```
$ bunx gitnexus@latest analyze
  LadybugDB native binary (lbugjs.node) is missing.
  This usually happens when the install lifecycle script was skipped.
```

The existing guidance can't rescue this. It offers `pnpm --allow-build`, a global install, or adding `trustedDependencies` to a project `package.json` — bunx has no project `package.json`, no per-invocation opt-in, **and re-extracts the package on every run**, so even a successful manual `node …/install.js` is wiped before the next invocation. I verified that directly: the repair works, and the very next `bunx` run reports the binary missing again. In-process recovery is the only thing that can work here.

It's also cheap, because nothing is actually absent — the binary is already on disk in `@ladybugdb/core-<platform>-<arch>` and the skipped script only copied it up. `checkLbugNative` now redoes that copy before reporting failure: **prebuilt only**, never a source build, never a network fetch. Best-effort by construction — read-only `node_modules`, an absent sub-package, or an unsupported platform all fall through to the existing diagnostics unchanged (pinned by a test).

This also covers `pnpm dlx` without `--allow-build` and `npm --ignore-scripts`.

### The generated block had no reachable bootstrap

The `gitnexus:start` block told a reader with no runner yet to run `npx gitnexus analyze`, falling back to a global npm install — both naming binaries a bun-only machine doesn't have.

It now lists `npx`, `bunx`, **and** `pnpm dlx` rather than resolving one. Deliberate: the block is committed, so emitting the command *this* machine resolves would make two contributors on different package managers rewrite it at each other on every analyze — the per-machine churn #1706 removed. Naming all three stays machine-independent and correct everywhere.

## How to verify

Unit tests (10 new, covering both new rungs, all four preserved outcomes, the lazy probe, and both recovery branches):

```bash
cd gitnexus && npx vitest run \
  test/unit/resolve-invocation.test.ts \
  test/unit/lbug-native-check.test.ts \
  test/unit/ai-context.test.ts
```

End to end on a real bun-only host (bun 1.3.14; no `npm`, `npx` or `pnpm` on PATH), with `lbugjs.node` deleted to reproduce the bunx state exactly:

```
$ node dist/cli/index.js analyze
  Repository indexed successfully (2.5s)
  5 nodes | 6 edges | 1 clusters | 0 flows

$ node .gitnexus/run.cjs --version
1.6.9

$ node -e "const m=require('./.gitnexus/run.cjs');
           console.log(m.resolveInvocationMode(), '|', m.formatAnalyzeCommand())"
bun | bunx gitnexus@latest analyze
```

Before this change all three failed with `npx: command not found`.

Full `test/unit` suite: `12 failed | 527 passed`, **identical to the same suite on unmodified `main`** in this sandbox (the failures are missing tree-sitter grammar builds, unrelated to these changes). Typecheck clean, prettier clean, eslint 0 errors.

## Risks / notes

- Behavior changes for exactly one pre-existing configuration: **npm 11+ with bunx but no pnpm**, which previously resolved to `npx` — the path #1939 documents as crashing. If you'd rather keep that on `npx`, dropping that one branch leaves the rest intact.
- `checkLbugNative` now writes to `node_modules` on the recovery path. It's the same copy the install script performs, idempotent, and guarded so any failure is silent and falls through to today's error.
- `trustedDependencies` was added to `gitnexus/package.json` so a plain `bun install` in this repo produces a working native binary — the remedy the error message already prescribes. Verified.
- Not addressed (out of scope, but same class of bug): `scripts/build.js` hardcodes `npm ci` for the web bundle, so `node scripts/build.js` fails at that stage on a machine without npm. The CLI build itself completes.

https://claude.ai/code/session_016R9psS9gJ73MRyquoBoPKg


gitnexus/hooks/claude/resolve-analyze-cmd.cjs
```diff
@@ -202,12 +211,29 @@ function resolveInvocationMode(probe = resolveOnPath, deps = {}) {
         ? deps.pnpmMajor !== null
         : Boolean(probe('pnpm'));
 
+  // bun presence is resolved lazily: only the two branches below can select it,
+  // so a machine with npm or pnpm never pays the extra PATH scan. `bunx` (not
+  // `bun`) is probed because `bunx` is what the resolved command actually runs.
+  let bunCache;
+  const hasBun = () => {
```
abhigyanpatwari, line 237: **P1 [code-read] — `hasBun()` is a PATH-existence check only, unlike the pnpm branch which actually execs `--version`.**

A `bunx` shim that's present-but-broken (partial uninstall, corrupted stub) is selected with zero functional validation. Contrast: the pnpm path (`probeVersion('pnpm')`, this file's `formatAnalyzeCommand`) actually spawns `pnpm --version` and only falls back to presence-only when the output is unparseable. `hasBun()` never gets an equivalent liveness check.

Picking `'bun'` mode also means `warnIfNpm11NpxRisk()` (resolve-invocation.ts:86-90) skips its npm-11/npx-crash warning — by design, since a working bunx isn't exposed to that crash. But that design assumes bunx-present ⇒ bunx-works, which this PATH-only check can't guarantee. When it's false, the result is a silent dead end: no warning, `bunx gitnexus@latest analyze` emitted, failure only surfaces at actual execution.

**Trigger:** npm-11 host, no pnpm, bun installed via official installer and later partially removed (failed self-update, manual `rm ~/.bun/bin/bun` without cleaning up `bunx`, etc.) — `bunx` remains an executable file on PATH but no longer runs.

**Suggested fix:** give bun the same liveness signal pnpm gets — a real `bunx --version` probe (accepting the extra spawn cost; this is the lazy-fallback branch already).

gitnexus/src/cli/ai-context.ts
```diff
@@ -184,9 +185,15 @@ ${tableBody}`
   // stay under the CLAUDE.md block token budget (#856); the cli skill carries the
   // full bootstrap + npm-11 fallback (`node.target is null` npx install crash).
   const runner = `node ${runnerPath}`;
+  // Bootstrap names every install-free one-shot rather than the one this machine
+  // resolves to: the block is committed, so a host-specific command would make
+  // two contributors on different package managers rewrite it at each other on
+  // every analyze (the per-machine churn of #1706). `bunx` is listed because a
+  // bun-only machine has no npm, npx or pnpm at all, and the npx-only note left
+  // it with a bootstrap command it could not run.
   const bootstrapNote =
```
abhigyanpatwari, line 194: **P2 [code-read] — the doc a bun-only user's agent would actually consult still describes the pre-bunx ladder.**

This is the generator for the `gitnexus:start` bootstrap note, correctly updated here. But `.claude/skills/gitnexus-cli/SKILL.md` (and its byte-identical shipped mirrors `gitnexus/skills/gitnexus-cli.md`, `gitnexus-claude-plugin/skills/gitnexus-cli/SKILL.md`) — none touched by this PR — still say the runner "auto-selects... global `gitnexus`, else `pnpm dlx`, else `npx`" with no `bunx` mention anywhere in the ladder or the bootstrap fallback list. `shipped-skills-sync.test.ts` only checks byte-identity between the three copies plus presence of `` `--pdg` `` — not runner-order accuracy — so this isn't CI-caught.

3 independent lanes (docs-dod, risk-architect, project-standards) converged on this. A bun-only-machine user (exactly this PR's target audience) whose agent loads this skill file to bootstrap a missing `.gitnexus/run.cjs` gets told to use npx/npm/pnpm — none of which exist on their machine.

**Suggested fix:** update all three SKILL.md copies to include `bunx` in the ladder and bootstrap fallback list. Pure doc edit, no code risk, not a merge blocker but high-value given it directly undercuts this PR's stated purpose.

gitnexus/src/core/lbug/native-check.ts
```diff
@@ -24,6 +24,63 @@ export interface NativeCheckResult {
...
+    candidates.push(
+      path.join(pkgDir, 'prebuilt', `lbugjs-${process.platform}-${process.arch}.node`),
+    );
+
+    const prebuilt = candidates.find((candidate) => fs.existsSync(candidate));
+    if (prebuilt === undefined) return false;
+
+    fs.copyFileSync(prebuilt, binaryPath);
+    return fs.existsSync(binaryPath);
+  } catch {
```
abhigyanpatwari, line 138: **P2 [code-read] — distinct restore-failure causes collapse into one boolean, routing users to the wrong remedy.**

This outer `try/catch` makes an `EACCES`/`EROFS` from `copyFileSync` (pkgDir not writable) indistinguishable at the call site from "no prebuilt sub-package exists at all." `checkLbugNative` then always shows the generic `binary_missing` message, whose entire "Common causes" list is about lifecycle-script skipping (pnpm `--allow-build`, npm `ignore-scripts`, bun `trustedDependencies`) — none of which fix a non-writable filesystem.

**Trigger:** a container/CI image with `node_modules` baked into a read-only layer, mounted read-only into ephemeral job containers (a common, legitimate pattern) — the sub-package's `lbugjs.node` is present and readable, only the copy-up is blocked, and the user is diagnosed into the wrong remediation category.

**Suggested fix:** thread a lightweight failure-reason string (`'no-subpackage' | 'copy-failed' | 'malformed-manifest'`) out of `restorePrebuiltNativeBinary` so the caller can append a differentiated line. (2 lanes converged: ce-adversarial, ce-reliability.)

gitnexus/package.json
```diff
@@ -96,6 +96,11 @@
     "@huggingface/transformers": "^4.1.0",
     "onnxruntime-node": "^1.24.0"
   },
+  "trustedDependencies": [
```
abhigyanpatwari, line 100: **P2 — `trustedDependencies`'s real scope (repo-dev installs only) is narrower than the PR's own narrative implies.**

ce-security empirically verified (live `bun install`, bun 1.3.14) this field only takes effect for someone running `bun install`/`pnpm install` directly inside this package — it has no effect on `bunx gitnexus@latest` one-shots (correctly acknowledged in `native-check.ts:38-41`'s own comment) and no effect on `bun add gitnexus`/`pnpm add gitnexus` consumer installs (trust is read from the consumer's own root manifest, not the dependency's). This is legitimate and safe — not misleading or unsafe — just easy for a future maintainer to mistake as "the bunx fix" and remove `restorePrebuiltNativeBinary` as seemingly redundant, which would actually break the bunx path. (4 lanes touched this question; ce-security's is the one empirical repro, the others code-read.)

**Suggested fix:** a one-line comment near this field (or in native-check.ts) stating explicitly: "only effective for `bun install`/`pnpm install` run inside this repo; the bunx path is covered by `restorePrebuiltNativeBinary` instead."

gitnexus/src/core/lbug/native-check.ts
```diff
@@ -48,7 +105,11 @@ export function checkLbugNative(overridePkgDir?: string): NativeCheckResult {
   }
 
   const binaryPath = path.join(pkgDir, 'lbugjs.node');
-  if (!fs.existsSync(binaryPath)) {
+  // A missing binary is recoverable whenever the prebuilt sub-package is present
+  // — the install script's copy is all that was skipped. Try that before
+  // reporting failure, so runners that cannot opt into lifecycle scripts (bunx)
+  // work instead of dead-ending on instructions they have no way to follow.
+  if (!fs.existsSync(binaryPath) && !restorePrebuiltNativeBinary(pkgDir, binaryPath)) {
```
abhigyanpatwari: **P3 [code-read] — the `binary_missing` repair message (lines ~133-134, pre-existing, not touched by this diff) gives bunx one-shot users advice they can't act on.**

When `restorePrebuiltNativeBinary` fails here, the fallback message says `bun: add to package.json and reinstall: "trustedDependencies": ["@ladybugdb/core"]` — inapplicable for a `bunx` one-shot user (no project package.json to edit), and lists only 1 of the 3 packages this PR's own `package.json` now trusts. Not a regression this PR introduces, but this PR is what makes bunx users far more likely to actually reach this message. (Converged across facts-historian, risk-architect, ce-correctness, ce-testing.)

**Suggested fix:** append an actionable alternative for the bunx case (e.g. "for bunx one-shot runs, use `bun install -g gitnexus` instead") and update the example to list all three packages.

abhigyanpatwari (commented):
# PR #2765 Tri-Review Digest — feat(cli): add a bunx lane so bun-only machines can run gitnexus

**Repo:** abhigyanpatwari/GitNexus · **PR:** https://github.com/abhigyanpatwari/GitNexus/pull/2765
**Base:** `27ab37c4329a8321cdefb3b3a0afdda26f72c0c6` · **Worktree:** `/workspace/.worktrees/pr-2765`

## What this PR does

Adds `bun`/`bunx` as a fourth invocation lane to the CLI's runner-resolution ladder (`resolveInvocationMode` in `resolve-analyze-cmd.cjs`, mirrored in two committed copies), so machines with only `bun`/`bunx` on PATH — and npm-11 machines with no `pnpm` — can run `gitnexus` without hitting the npm-11/arborist npx crash (#1939) or a bare `npx: command not found`. It pairs this with a new `restorePrebuiltNativeBinary()` in `native-check.ts` that re-does the native-binary copy step any lifecycle-script-skipping runner (`bunx`, `pnpm dlx` without `--allow-build`, `npm --ignore-scripts`) would otherwise skip, plus a `trustedDependencies` field in `gitnexus/package.json`, plus doc/bootstrap-note updates. 11 files changed, +319/-34.

**Engine breakdown:** Claude ran both lanes — the GitNexus PR swarm (6 personas: facts-historian, branch-hygiene, risk-architect, test-ci, security-boundary, docs-dod) and the ce-code-review personas (7: correctness, testing, maintainability, project-standards, adversarial, security, reliability). **Codex (Engine B) was unavailable this run** — all three legs (swarm, ce, adversarial) hit an organization workspace spend cap and returned zero output (confirmed by reading `codex-swarm.md`, `codex-ce.md`, `codex-adversarial.md`, and `codex-driver.log`: all three threads errored immediately with "You hit your spend cap set by the owner of your workspace"). **No cross-engine corroboration is possible in this review** — every finding below comes from Claude only. Where multiple Claude lanes converge on the same issue, that is noted as "moderate" confidence (consistent within one engine, not independent-engine confirmation); single-lane findings are "weak" tier regardless of the lane's own confidence score, except where severity overrides tier per the review rules (see P1 below).

**Overall verdict:** No merge-blocking correctness or security defect found. One P1 worth fixing before or shortly after merge (silent selection of a broken bunx shim, single-lane but concrete). Several P2/P3 documentation-precision and diagnostic-quality gaps, the most notable being that the exact skill doc a bun-only user's agent would consult to bootstrap the CLI still describes the pre-bunx runner ladder. Test coverage for the new code is good but has one clean, real gap (the legacy `prebuilt/` fallback candidate is never exercised).

---

## What's solid

- **The `resolveInvocationMode` ladder was traced correct across all combinations by 3+ independent lanes.** branch-hygiene, risk-architect, and ce-correctness each independently walked all 12 `(npmMajor ∈ {null,<11,≥11}) × hasPnpm × hasBun` combinations and confirmed every one resolves to the mode the doc-comment promises; pnpm is unconditionally preferred over bun whenever it's present, npx still wins on npm<11, and the escape hatches (`GITNEXUS_INVOCATION=bun`, `bunPresent:false`) work as documented.
- **The byte-identical CJS mirror pair, and its parity test.** `gitnexus/hooks/claude/resolve-analyze-cmd.cjs` and `gitnexus-claude-plugin/hooks/resolve-analyze-cmd.cjs` are confirmed identical after this diff (facts-historian, branch-hygiene, ce-correctness, ce-maintainability, ce-testing all checked directly), and `resolve-invocation.test.ts:~609` (`'keeps the two CJS hook copies byte-identical'`) enforces it in CI, so a future edit that touches only one copy fails the build rather than silently drifting.
- **The restore-fails-safely-to-diagnostics contract.** `restorePrebuiltNativeBinary()` never fabricates a false `ok:true`: every failure mode (JSON parse failure, `require.resolve` miss, no candidate file, `copyFileSync` throw) is caught and normalized to `false`, and `checkLbugNative`'s `!fs.existsSync(binaryPath) && !restorePrebuiltNativeBinary(...)` falls through to the pre-existing `binary_missing` diagnostic unchanged (risk-architect, ce-correctness, ce-reliability all confirmed by trace + the new "still reports binary_missing…" test). A successful restore still has to pass the existing out-of-process SIGBUS/load-error probe before `ok:true` is returned — a copied-but-corrupt binary is caught the same way a pre-existing corrupt one always was.
- **`trustedDependencies` scope was empirically verified safe by ce-security**, which is the one lane in this run that didn't just read code — it ran live `bun install` reproductions (bun 1.3.14, in-sandbox) rather than reasoning from memory. Confirmed: (1) a dependency's own self-declared `trustedDependencies` has no effect for downstream consumers — only the install ROOT's own package.json is honored; (2) the `"gitnexus"` self-entry is inert in every tested scenario (root and workspace-member lifecycle scripts run regardless of self-listing); (3) the three listed packages (`@ladybugdb/core`, `gitnexus`, `tree-sitter`) are not over-broad — no unrelated package was smuggled in. This directly refutes any concern that the field over-widens trust; it's narrower than the PR narrative implies (see P2 below) but not misleading or unsafe. Because this is the one place in the run with an actual empirical repro rather than a code-read, it should be weighted above the code-read-only lanes that examined the same question (security-boundary swarm, ce-correctness, facts-historian).

---

## Findings

### P1 — Stale/broken bunx shim can be silently selected, and selecting it suppresses the only pre-execution warning the user would otherwise get

**Lane:** ce-adversarial only (ADV-1, confidence 75). **Tier: weak by lane count, but severity overrides corroboration per review rule 2 — postable.** **Tag: [code-read]**, independently re-verified by the synthesis pass reading the live code (see evidence below).

**Problem:** `hasBun()` is a PATH-existence check only, unlike the npm/pnpm branches in the same function, which actually exec `--version`. A `bunx` binary that is present-but-broken (partial uninstall, corrupted shim) is selected with zero functional validation. Note on the warning-suppression half of this: `resolve-invocation.ts:86-87`'s doc-comment makes the suppression of `warnIfNpm11NpxRisk` for non-`'npx'` modes explicit, intentional design ("skipped when … `bunx` is already preferred, so it never nags users who are not exposed to the npx/arborist crash") — it is not an oversight. The gap is that this intentional design rests on an assumption (`bunx` present ⇒ `bunx` works) that `hasBun()`'s PATH-only check cannot actually guarantee; when that assumption is false the correct-by-design suppression becomes a silent dead end instead of a safe simplification. The facts-historian lane separately confirmed the suppression mechanism itself works exactly as coded (no contradiction — that lane checked "does the code do what the comment says," this lane checked "does what the comment assumes actually hold under adversarial conditions").

**Evidence:**
- `gitnexus/hooks/claude/resolve-analyze-cmd.cjs:218-222` —
  ```js
  let bunCache;
  const hasBun = () => {
    if (bunCache === undefined) {
      bunCache = 'bunPresent' in deps ? Boolean(deps.bunPresent) : Boolean(probe('bunx'));
    }
    return bunCache;
  };
  ```
  `probe` defaults to `resolveOnPath` (line 73), a pure `fs.statSync`/`fs.accessSync` PATH scan with no exec. Contrast the pnpm branch, `formatAnalyzeCommand` (`resolve-analyze-cmd.cjs:258-289`), which calls `probeVersion('pnpm')` — an actual `execFileSync(command, ['--version'], …)` — and only falls back to presence-only when the version output is unparseable (a documented Corepack-banner/timeout case). bun gets no equivalent attempt in either branch.
- `resolve-analyze-cmd.cjs:228` — `if (npmMajor !== null && npmMajor >= 11 && hasBun()) return 'bun';` selects `'bun'` on this PATH-only signal.
- `gitnexus/src/cli/resolve-invocation.ts:89-91` — `warnIfNpm11NpxRisk()`'s first line is `if (resolveInvocationMode() !== 'npx') return;`. Once mode resolves to `'bun'` instead of `'npx'`, this explicit early-return skips the npm-11/arborist-crash nudge entirely, and there is no equivalent "bun selected but looks unhealthy" warning anywhere.

**Concrete trigger:** an npm-11 host with no `pnpm` on PATH, where bun was installed via the official installer (adding `~/.bun/bin` to a shell rc PATH) and later partially removed — e.g. a failed self-update, or `rm ~/.bun/bin/bun` without cleaning up `bunx`, or a distro package whose uninstall script left a stub. `~/.bun/bin/bunx` remains a real, executable file on PATH but no longer runs. `resolveInvocationMode` selects `'bun'`, `formatBunxCommand` emits `bunx gitnexus@latest analyze` with no indication anything is wrong, and the direct-exec entrypoint (`resolve-analyze-cmd.cjs:342-363`) either ENOENTs or the broken shim exits non-zero with its own (not gitnexus's) error text — a zero-warning path from "command resolved" straight to "command fails at run time."

**Suggested fix:** give bun the same liveness signal pnpm gets — a real `bunx --version` probe (accepting the extra spawn cost, since this is the lazy-fallback branch anyway), so `hasBun()` reflects "bunx actually runs," not just "a file matching that name is on PATH." (An earlier draft of this finding also floated unconditionally re-enabling `warnIfNpm11NpxRisk` whenever mode resolves to `'bun'` — dropped on review: that would fire the npm-11/npx-crash warning even on the common, correctly-working-bunx machine, which is exactly the nagging the existing suppression design is meant to prevent. A real liveness probe is the fix that doesn't trade one failure mode for another.)

---

### P2 — gitnexus-cli SKILL.md (and shipped mirrors) still document the pre-bunx 3-rung ladder — the exact doc the target user population would consult

**Lanes converged:** docs-dod (Finding 2, P2), risk-architect (Finding 1, P3), project-standards (PS-1, confidence 50) — **3 independent Claude lanes, moderate tier.** **Tag: [code-read]**, independently re-verified.

**Problem:** `.claude/skills/gitnexus-cli/SKILL.md`, `gitnexus/skills/gitnexus-cli.md`, and `gitnexus-claude-plugin/skills/gitnexus-cli/SKILL.md` (byte-identical to each other, none touched by this diff) still say the runner "auto-selects an available runner at call time (global `gitnexus`, else `pnpm dlx`, else `npx`)" with a bootstrap fallback of only `npx gitnexus analyze` / `npm i -g gitnexus` / `pnpm --allow-build=… dlx`. No `bunx` mention anywhere. Risk-architect additionally flagged `.claude/skills/gitnexus-plan/SKILL.md:135,342` (+ its shipped copy) with the same staleness.

**Evidence:** `.claude/skills/gitnexus-cli/SKILL.md:8` (verified directly):
```
It auto-selects an available runner at call time (global `gitnexus`, else `pnpm dlx`, else `npx`)
```
`gitnexus/test/unit/shipped-skills-sync.test.ts` only asserts the three copies are byte-identical to each other and (for gitnexus-cli specifically) contain the string `` `--pdg` `` — it has no assertion tying content to the actual runner-resolution order, so this staleness is not CI-caught; all three copies pass while being consistently wrong.

**Concrete trigger:** a bun-only-machine user (exactly this PR's target audience) whose coding agent loads `.claude/skills/gitnexus-cli/SKILL.md` to figure out how to bootstrap a missing `.gitnexus/run.cjs` — the doc tells them to use `npx`/`npm`/`pnpm`, none of which exist on their machine, directly undercutting the PR's stated purpose.

**Suggested fix:** update all three copies' line 8 to include `bunx` in the ladder and add `bunx gitnexus@latest analyze` to the bootstrap fallback list (docs-dod's suggested wording: "global `gitnexus`, else `pnpm dlx`, else `bunx`, else `npx`"). Not a merge blocker but high-value, and cheap — pure doc edit, no code risk.

---

### P2 — Distinct restore-failure causes (read-only filesystem vs. genuinely absent sub-package) collapse into one boolean, routing the user to the wrong remedy

**Lanes converged:** ce-adversarial (ADV-3, confidence 55, P2) and ce-reliability (F1, confidence 55, P2) — same root cause, different framing (adversarial constructs a concrete container/read-only-mount trigger; reliability names the general "reason is discarded" defect). **Moderate tier, 2 lanes. Tag: [code-read]**, verified against the actual diff.

**Problem:** `restorePrebuiltNativeBinary`'s outer `try { … } catch { return false; }` (`native-check.ts:50,79-81`, verified) makes an EACCES/EROFS from `fs.copyFileSync` (needs write access to `pkgDir`) indistinguishable at the call site from "no prebuilt sub-package exists at all." `checkLbugNative` then always falls through to the generic `binary_missing` message (`native-check.ts:112-138`), whose entire "Common causes" list is about lifecycle-script skipping (pnpm `--allow-build`, npm `ignore-scripts`, bun `trustedDependencies`) — none of which mention or would fix a non-writable filesystem.

**Concrete trigger:** a container/CI image that bakes `node_modules` into a read-only layer and mounts it read-only into ephemeral job containers (a common, legitimate pattern) — the sub-package's `lbugjs.node` is present and readable, only the copy-up step is blocked by the filesystem, and the user is diagnosed into the wrong remediation category (told to reinstall/add `trustedDependencies`, none of which addresses a read-only mount).

**Suggested fix:** thread a lightweight failure-reason string (`'no-subpackage' | 'copy-failed' | 'malformed-manifest'`) out of `restorePrebuiltNativeBinary` so `checkLbugNative` can append one differentiated line instead of staying silent about why the auto-repair didn't help.

---

### P2 — `trustedDependencies`'s real scope (repo-dev installs only) is narrower than the PR's own narrative implies, and isn't stated anywhere a future maintainer would see it

**Lanes converged:** security-boundary swarm (Finding 1, P2, confidence high), ce-security (empirical, see "What's solid" above), ce-correctness (residual risk, ~40 confidence, flagged as unconfirmed at the time), facts-historian (Finding 2, P2) — **4 lanes touched this question; moderate tier. Tag: [reproduced]** for the ce-security empirical portion (live `bun install` test), [code-read] for the others.

**Problem:** the field only takes effect for someone running `bun install`/`pnpm install` directly inside the `gitnexus/` package (a contributor or fork) — it has no effect on `bunx gitnexus@latest` one-shots (correctly and explicitly acknowledged in the PR's own `native-check.ts:38-41` comment) and no effect on `bun add gitnexus`/`pnpm add gitnexus` consumer installs (trust is read from the consumer's own root package.json, not the dependency's). This is legitimate and safe (see What's Solid), just easy to mistake for covering more than it does — a future maintainer could read it as "the bunx fix" and remove `restorePrebuiltNativeBinary` as "redundant," which would actually break the bunx path.

**Suggested fix:** a one-line clarifying comment near the `trustedDependencies` field (or in `native-check.ts`, which already has related commentary) stating explicitly: "only effective for `bun install`/`pnpm install` run inside this repo; the bunx path is covered by `restorePrebuiltNativeBinary` instead, not by this field."

---

### P3 — `binary_missing`/`load_failed` repair messages give bunx one-shot users advice they cannot act on, and understate the now-3-package `trustedDependencies` list

**Lanes converged:** facts-historian (Finding 3 + Finding 4), risk-architect (Finding 2), ce-correctness (residual risk), ce-testing (Finding 5) — **moderate tier, this is the "bunx-inapplicable-error-message" theme called out for dedup. Tag: [code-read]**, verified.

**Problem:** `native-check.ts:133-134` and `:205` (pre-existing text, not touched by this diff — confirmed via `git diff`) say:
```
'  - bun: add to package.json and reinstall:',
'      "trustedDependencies": ["@ladybugdb/core"]',
```
This is inapplicable for a `bunx` one-shot user reaching this message (i.e., `restorePrebuiltNativeBinary` itself failed) — they have no project `package.json` to edit — and lists only `@ladybugdb/core` while the package.json this PR adds now lists three packages (`@ladybugdb/core`, `gitnexus`, `tree-sitter`). Not a regression this PR introduced, but the PR's own new doc comments elsewhere in the same file (explaining exactly why bunx has "nowhere to put" trustedDependencies) highlight this exact inconsistency without fixing it — and this PR is what makes bunx users far more likely to actually reach this message.

**Suggested fix:** append an actionable alternative for the bunx case (e.g. "for bunx one-shot runs, use `bun install -g gitnexus` or add `trustedDependencies` to a project package.json first") and update the example to reflect all three packages.

---

### P3 — `require.resolve()` request built from an unvalidated package.json `name` field (theoretical path traversal)

**Lanes converged:** security-boundary swarm (Finding 2, medium confidence, "theoretical") and ce-security (empirical, confidence 35) — **moderate tier; validated-safe/low-severity item.** **Tag: [reproduced]** — ce-security actually tested this on Node 22.18.0.

**Problem/evidence:** `native-check.ts:54,60-63` passes `manifest.name` (read from `pkgDir/package.json`) into `createRequire(...).resolve(`${manifest.name}-${platform}-${arch}/package.json`, { paths: [pkgDir] })` with only a non-empty-string check. ce-security empirically confirmed a relative/absolute `manifest.name` (e.g. `../../../outside/victim`) does resolve outside `pkgDir` via Node's `require.resolve`, but a scoped-name-style traversal (`@scope/../../outside/victim`, closer to what a registry-published `name` field could actually contain) failed to resolve.

**Why low severity, not exploitable in practice:** the only way to control `manifest.name` is to already control `pkgDir/package.json` — the real, locally-installed `@ladybugdb/core` manifest (all 3 production call sites use no `overridePkgDir`; that parameter is test-only). An attacker able to write that file already has write access to the exact directory this function writes into (`pkgDir/lbugjs.node`) and can overwrite the binary directly — this function grants no capability such an attacker doesn't already have. Requires either bypassing npm registry package-name validation or a private-registry/tarball-tamper scenario — a general supply-chain-compromise precondition this diff does not introduce.

**Suggested fix (cheap defense-in-depth, not urgent):** `if (/^[./\\]/.test(manifest.name)) return false;` before the resolve call, or assert `manifest.name === '@ladybugdb/core'`.

---

## Lower-priority / informational

- **`resolveInvocationMode`'s 5-branch tree is provably equivalent to a simpler 3-branch tree** (ce-maintainability, P2-labeled by the lane but a simplification suggestion, not a bug — verified by exhaustive enumeration across all 12 `(npmMajor, hasPnpm, hasBun)` combinations, both by hand and by running the real code against a reimplementation: 0 mismatches). The two new `bun` rungs are redundant with the existing pnpm/npx rungs because pnpm is already unconditionally preferred over bun by the time the bun rung is reached. Collapsing to `if (hasPnpm) return 'pnpm'; if (hasBun()) return 'bun';` plus the existing npx guards would preserve behavior with less conditional surface. No behavioral impact either way; purely a readability suggestion.
- **`hasPnpm` is eager, `hasBun` is lazy** — asymmetric treatment of a symmetric pair (ce-maintainability, P3). Pre-existing pattern for `hasPnpm` (not introduced by this PR), but the new `hasBun` sibling makes the inconsistency visible.
- **`copyFileSync` is non-atomic** (security-boundary Finding 3 + ce-reliability F2, merged — P3). A mid-write interrupt (ENOSPC/EACCES/kill) can leave a partial file; the current invocation reports `binary_missing` rather than `load_failed`, but self-corrects on the very next invocation (existsSync sees the partial file, skips restore, load-probe correctly reclassifies it). Cosmetic only.
- **README gaps**: `gitnexus/README.md:27-33` (Quick Start npm-11 callout, pnpm/npm alternatives only) and `:396-419` (troubleshooting code block) list no bunx option; `gitnexus/README.md:421-450`, the `ERR_DLOPEN_FAILED`/`lbugjs.node missing` troubleshooting section, doesn't mention bunx now auto-recovers from this exact failure class via `restorePrebuiltNativeBinary`; root `README.md:64-71` and `gitnexus-cursor-integration/README.md:43` have the same gap, lower traffic (docs-dod Findings 3-6, P2/P3 mix, lower priority than the SKILL.md finding above).
- **`formatDocumentationDlxCommand` was not given a bunx branch** (only `formatAnalyzeCommand` was) — facts-historian flags this as worth confirming intentional (different invocation context — the warning path deliberately still recommends pnpm/npm, may not need bunx), not confirmed as a bug.
- **Commit-message imprecisions** (facts-historian, P3, doc-only, no code impact): commit `77d3aa7f`'s "`#1706`" citation is analogically apt but not the precise issue; the "bun re-extracts the package on every bunx run" claim in a commit message is imprecise (bun caches per-version extractions but never runs lifecycle scripts regardless of cache state) — actual code behavior (`checkLbugNative` gates on `!fs.existsSync` first) is correct regardless.
- **AGENTS.md/CLAUDE.md changelog/version not bumped** despite a material change to the `gitnexus:start` bootstrap line (docs-dod Finding 7, project-standards residual risk) — optional by established convention, several past PRs skip this.
- **The `"gitnexus"` self-entry in `trustedDependencies` is inert** in every scenario ce-security tested (bun runs a root's or workspace-member's own lifecycle scripts unconditionally regardless of self-listing) — harmless, but reads as load-bearing; worth a maintainer note, not a fix.
- **CLAUDE.md still fully duplicates the `gitnexus:start` block** despite its own Reference Documentation section and Changelog (v1.2.0) claiming that duplication was removed (project-standards PS-2, confidence 40) — pre-existing drift, not caused or worsened by this PR (which only edited one already-duplicated line, correctly, in both copies).

---

## Refuted / investigated-and-could-not-break

- **"A corrupt/truncated sub-package binary is silently copied and accepted" — REFUTED.** ce-adversarial (ADV-4) constructed this scenario and found the existing out-of-process SIGBUS/load-error probe (`native-check.ts:141-154`, pre-existing, purpose-built for exactly this per its own #2441 comment) still runs after any restore-copy and correctly reports `load_failed` with an accurate message. One narrow, pre-existing (not introduced by this PR) residual noted: if the load-probe's child process itself cannot spawn at all (sandboxed environment, embedder with non-Node `process.execPath`, or the 15s timeout), the probe fails open (`ok: true`) — this pre-existing fail-open path is now *reached more often* because this PR's restore step funnels more "lifecycle script skipped" cases through it than before (previously a missing binary short-circuited to `binary_missing` without reaching the probe at all). Not a new bug, but a wider blast radius for an existing edge case.
- **"Wrong-arch binary gets copied" — REFUTED.** The `require.resolve` specifier is built as `${manifest.name}-${process.platform}-${process.arch}`, so the function is structurally incapable of resolving a sub-package for any platform/arch other than the one the current process reports (ce-adversarial ADV-4).
- **"trustedDependencies enables a typosquat/similar-path attack" — REFUTED.** ce-adversarial (RISK-1) investigated this specifically: npm-registry package names are globally unique and the field matches by exact string, so there's no prefix/glob a typosquatted package could exploit; would require full registry/account compromise of one of the three real packages, a pre-existing risk this PR does not create.
- **ce-correctness's full-suite pass found no findings at confidence ≥50** across a manual trace of all 12 decision-tree combinations and a line-by-line comparison of `restorePrebuiltNativeBinary` against the real `@ladybugdb/core/install.js` it reimplements — an explicit attempt to break the mechanism that came back clean.
- **Concurrent-invocation copyFileSync race — NOT fully refuted; unconfirmed, capped confidence.** ce-adversarial (ADV-2, confidence 40) and ce-reliability (residual risk, confidence ~30) both raised the same underlying scenario — two CLI invocations racing an unlocked `fs.copyFileSync` against the same `pkgDir`/`binaryPath` (a genuinely new failure surface: before this PR a missing binary short-circuited immediately with no copy attempt at all) — and both explicitly capped their own confidence because the trigger depends on external, version-sensitive package-manager cache-directory reuse semantics (`pnpm dlx`/`bunx` sharing a keyed store across concurrent invocations) that neither lane could verify from the sandbox. This is presented honestly as *investigated but inconclusive*, not as refuted: if the external precondition holds, the failure mode would be a transient, self-correcting misdiagnosis ("truncated or corrupted" reported against a perfectly good binary caught mid-race), not data loss or a false success. Worth a cheap mitigation (write-to-temp-then-rename) if a maintainer wants to close the question rather than leave it open.
- **TOCTOU between `existsSync` and `copyFileSync`** (security-boundary, ce-security) — real in the abstract (a symlink swap could redirect the read) but requires a co-resident attacker who already has write access to the exact `node_modules` subtree the function operates in, i.e. no capability escalation beyond what that attacker already has. Same threat model as the `require.resolve` traversal item above; not actionable.

---

## Test gaps

- **Legacy `prebuilt/lbugjs-<platform>-<arch>.node` fallback (`native-check.ts:70-72`, `candidates[1]`) has zero test coverage** — the clearest, most concrete gap in this run (ce-testing Finding 1, confidence 75, `grep -n prebuilt test/unit/lbug-native-check.test.ts` only matches test titles and an unrelated comment, never a fixture path). Both new `lbug-native-check.test.ts` tests exercise only `candidates[0]`-succeeds or neither-candidate-exists; the legacy-layout branch this PR itself introduces is never reached. Related but distinct: swarm test-ci separately flagged that sub-package resolution is untested for platform/arch combos absent from the CI environment (e.g. windows-arm64) — a different but adjacent gap in the same function, echoed as residual/testing-gap notes in ce-maintainability, ce-adversarial, and ce-security.
- **No test proves `bunx` present-but-non-functional (broken shim) behavior** — ties directly to the P1 finding (ADV-1). Every bun test in `resolve-invocation.test.ts` injects `bunPresent`/a fake probe, bypassing `resolveOnPath`'s real filesystem walk and any actual `bunx` invocation.
- **No test exercises a `copyFileSync` failure** (read-only destination / EACCES / EROFS) to confirm `restorePrebuiltNativeBinary` degrades to `false` cleanly rather than propagating — ties to the P2 finding on collapsed failure causes (ce-adversarial, ce-reliability, ce-maintainability, ce-security all separately note this).
- **No test proves `GITNEXUS_INVOCATION=bun` bypasses the `hasBun()`/PATH probe** the way the existing pnpm case is explicitly pinned (`resolve-invocation.test.ts:273-278`, `'lets GITNEXUS_INVOCATION override the probe without consulting it'`) — the new bun-mode test only proves formatting is correct once mode is already `'bun'`, not that forcing bypasses a negative probe (ce-testing Finding 3, confidence 60).
- **No test for concurrent `checkLbugNative`/`restorePrebuiltNativeBinary` invocations** against the same `pkgDir` (ties to the unconfirmed concurrent-race item above).
- **No test asserts `manifest.name` with a path-traversal-shaped value falls through cleanly** to `binary_missing` (ce-security).
- **No test ties `.claude/skills/gitnexus-cli/SKILL.md`'s content to the actual runner-resolution order** — `shipped-skills-sync.test.ts` only checks byte-identity between copies and presence of `` `--pdg` ``, not runner-order accuracy (project-standards) — this is exactly why the P2 doc-staleness finding above wasn't CI-caught.
- **`lbug-native-check.test.ts` is excluded from the cross-platform CI matrix** (`gitnexus/scripts/cross-platform-tests.ts`) — it only runs on `ubuntu-latest`; the restore mechanism (file ops, path resolution) is platform-agnostic logic but is never directly exercised on Windows/macOS runners (swarm test-ci, P3).
- **No end-to-end smoke test on a real bun-only machine** — all bun-path tests use injected deps/mocks; this is the definitive validation this feature is missing, not currently present anywhere (facts-historian follow-up).

---

## CI status

All checks pass (15,912/15,981 tests passed, 69 skipped) except the Vercel deploy check, which fails with "Authorization required to deploy" — confirmed a permissions/auth gate on the Vercel integration, not a code or test failure. `mergeStateStatus` is `BEHIND` main by 3 commits, all dependency-only version bumps (`@types/node` ×2, `@vitejs/plugin-react`) that don't touch any of this PR's 11 files — `mergeable: MERGEABLE`, no conflicts. Rebase is not required but preferred for a linear history (branch-hygiene).

## Coverage note

This is an 11-file, ~319-line-added / 34-line-removed PR. The full diff was read in Phase 1 and independently by every lane in this run (both swarm and ce personas) — no partial-coverage caveat applies here.

---

## Bottom line

**Branch hygiene:** merge-from-main commit present but harmless and merge-safe — `42d9ae5cd` is a non-conflicting maintainer merge of `main`; the 3 commits main has gained since then are dependency-only version bumps that touch none of this PR's 11 files.
**Merge state:** mergeable — GitHub reports `MERGEABLE`/`BEHIND`, no conflicts; the only failing check (Vercel) is an auth/permissions gate, not a code or test failure.
**Final verdict: production-ready with minor follow-ups.** No P0/merge-blocking defect in either engine's coverage this run. One P1 (silent selection of a non-functional `bunx` shim, single Claude lane but a concrete, code-verified, unrefuted gap) is worth fixing before or shortly after merge. The remaining findings are documentation-staleness and diagnostic-message-quality gaps, several independently corroborated across 2-4 Claude lanes. This review is Claude-only (Codex hit a workspace spend cap on all three legs, zero output) — treat the corroboration tiers above as "moderate" (multiple Claude lanes) or "weak" (single lane) only; no cross-engine confirmation was possible this run.

---

*This is an automated multi-tool digest (Claude swarm + ce personas; Codex unavailable this run — spend cap, zero output on all 3 legs). Verify findings against the live code before acting on them.*

azizur100389:
Reviewed current head `42d9ae5c` against `upstream/main` `51095c19` in a fresh worktree. The branch merged cleanly; `git diff --check` passed; no hidden bidi controls were found. Local validation passed: 117 targeted tests with 2 existing skips across `ai-context`, `lbug-native-check`, and `resolve-invocation`, plus `npx tsc --noEmit`. GitHub CI is green apart from the external Vercel authorization check.

Branch hygiene: `merge-from-main commit present but harmless and merge-safe`. Merge state: `mergeable` (currently behind main, without conflicts).

No production-readiness issues found against the current DoD bar.

Minor follow-ups:

- **Risk:** a present but broken `bunx` shim is treated as usable because the new bun path checks PATH existence but does not execute a liveness/version probe; this can select `bun` and suppress the npm-11 warning before the emitted command fails.
- **Evidence to check:** `gitnexus/hooks/claude/resolve-analyze-cmd.cjs` around `hasBun()` and `resolveInvocationMode()`, and `gitnexus/src/cli/resolve-invocation.ts` in `warnIfNpm11NpxRisk()`.
- **Recommended fix:** probe `bunx --version` on the lazy bun fallback path, or otherwise distinguish a runnable bunx from a stale shim.
- **Blocks merge:** no.

- **Risk:** the shipped `gitnexus-cli` skill documentation still describes only global/pnpm/npx bootstrap paths, which leaves the target bun-only user without the newly supported command.
- **Evidence to check:** `.claude/skills/gitnexus-cli/SKILL.md` and its shipped mirrors.
- **Recommended fix:** include `bunx gitnexus@latest analyze` and keep the mirrored skill files synchronized.
- **Blocks merge:** no.

Final verdict: `production-ready with minor follow-ups`. The implementation preserves existing runner precedence, the native restore path remains best-effort and still passes the existing loadability probe, and targeted/current-main validation is clean. The remaining items are resilience and documentation gaps rather than demonstrated regressions.

azizur100389:
Revalidated current head `faa43844` against current `upstream/main` `911151e2` in a fresh worktree. The head's only change since the last review is another merge-from-main commit; all substantive files belonging to this PR are unchanged.

The PR merges cleanly onto today's main, `git diff --check` passes, `npx tsc --noEmit` passes, and the focused validation remains green: 117 tests passed with 2 expected skips across `resolve-invocation`, `lbug-native-check`, and `ai-context`.

No new code findings. The implementation remains production-ready from the local review perspective. The branch is currently behind main and GitHub exposes only the unrelated Vercel authorization failure for this head, so it should be updated and receive a current functional CI run before merging.

Issue #2746, opened 2026-07-29
Python: a class reached through a module alias (`models.User(db)`) resolves in no spelling


magyargergo:
Split out of #2708 / #2737, where it was noticed but deliberately left out of scope: it is not a construction-typing bug, it sits upstream of that.

## Symptom

A class reached through a module alias never resolves — in **either** spelling. Fixture:

```python
# pkg/models.py
class User:
    def save(self):
        return 1
```

```python
# pkg/app.py
from pkg import models
from pkg.models import User


def via_module_attr_inline(db):
    return models.User(db).save()


def via_module_attr_twostep(db):
    u = models.User(db)
    return u.save()


def via_direct_import(db):
    return User(db).save()
```

Every CALLS/IMPORTS edge emitted:

```
| app.py             | IMPORTS | File:pkg/models.py            |
| via_direct_import  | CALLS   | Method:pkg/models.py:User.save#0 |
```

`via_module_attr_inline` and `via_module_attr_twostep` emit **nothing**. Only the directly-imported spelling resolves.

## Why it is not the #2708 construction gap

#2737 made inline constructor receivers resolve, and `via_direct_import` proves that path works. The module-alias spellings fail identically in the **two-step** form, which never touches construction typing at all — so the receiver's type is not being established from `models.User` in the first place. `models` is a module alias, not a value with a class type, and `User` is a member of that namespace rather than a member of a class.

Additionally, with **only** `from pkg import models` in the file (drop the second import), no `IMPORTS` edge to `pkg/models.py` is emitted either — so a module-only import may not be registering a resolvable target at all. That may be one bug or two; worth confirming while fixing.

## Why it matters

`import x.y as z` / `from pkg import module` then `module.Class(...)` is an extremely common Python convention — arguably more common in large codebases than importing the class directly, since it avoids name collisions. Every call through that form is currently invisible to `impact()` and `context()`.

ARCHITECTURE.md lists Python's import mode as `namespace` — "Module aliases resolved at call site" — so the intent exists; this is the case where it does not fire.

## Related

- #2708 / #2737 — inline constructor receivers (the direct-import spelling now works)
- #2744 — the epistemic half of #2708; a drop like this is exactly what should degrade `epistemic` to `lower-bound`

azizur100389:
I would like to take this. I will keep the fix Python-specific and conservative: cover `from pkg import models` module aliases for qualified construction and type binding, with positive and ambiguity-negative resolver fixtures.

PR #2769, opened 2026-07-31
fix(schema): declare Swift member-containment pairs in CONTAINS DDL

### LINKED ISSUE #2781: analyze crashes on Vue/JS projects: Const→Method not declared in LadybugDB relation schema
# `analyze` crashes on Vue/JS projects: `Const→Method` not declared in LadybugDB relation schema

## Summary

`gitnexus analyze` crashes with `Relationship label pair Const→Method is not declared in the LadybugDB relation schema` when indexing Vue/JS projects that use `const obj = { method() {} }` patterns (Vuex stores, Vue mixins, etc.)

## Context

PR #1718 (`dd352732`, 2026-05-22) added Case 5 (value-receiver bridge) in `receiver-bound-calls.ts` and made `Const` a linkable label (`isLinkableLabel` in `node-lookup.ts`). This enables `Const` nodes to be CALLS sources — e.g.:

```javascript
const fooService = { getUser() {...} };
fooService.getUser();  // ← produces Const→Method CALLS edge
```

However, `RELATION_SCHEMA` in `gitnexus/src/core/lbug/schema.ts` was never updated to declare the corresponding `FROM Const TO Method` (and other `Const→*`) pairs. The schema only has:

```sql
FROM `Const` TO Community,   -- MEMBER_OF
FROM `Const` TO Process,     -- STEP_IN_PROCESS
```

PR #2740 (`de84ad62`, 2026-07-31) added a fail-fast `declaredPairs` check in `RelPairRouter.route()`, which now throws on undeclared pairs instead of silently writing them. This exposed the latent schema gap as a hard crash.

## Expected behavior

`gitnexus analyze` should complete successfully for Vue/JS projects containing `const obj = { method() {} }` patterns. The `Const→Method` CALLS edge should be persisted to LadybugDB.

## Actual behavior

`analyze` crashes with:

```
Error: Relationship label pair Const→Method is not declared in the LadybugDB relation schema
    at RelPairRouter.route (file:///app/gitnexus/dist/core/lbug/rel-pair-routing.js:113:19)
    at streamAllCSVsToDisk (file:///app/gitnexus/dist/core/lbug/csv-generator.js:666:43)
    at async loadGraphToLbug (file:///app/gitnexus/dist/core/lbug/lbug-adapter.js:934:15)
    at async runFullAnalysisInner (file:///app/gitnexus/dist/core/run-analyze.js:1583:13)
```

## Steps to reproduce

1. Create a Vue/JS project with a Vuex store module or mixin using the pattern:

   ```javascript
   // store/modules/app.js
   const mutations = {
     TOGGLE_SIDEBAR(state) { ... },
     CLOSE_SIDEBAR(state, withoutAnimation) { ... }
   };
   ```

   or:

   ```javascript
   // mixins/submit.js
   export const submitMixins = {
     methods: {
       handleSubmit() { ... },
       handleSave() { ... }
     }
   };
   ```

2. Run `gitnexus analyze --force --embeddings 0 --worker-timeout 300`
3. Observe the crash during the CSV generation phase

## Environment

- OS: Linux (Docker container)
- Node: v22+
- Commit: `de84ad62` or later (includes the `declaredPairs` fail-fast check from PR #2740)

## Suggested fix

Add missing `FROM Const TO *` pairs to `RELATION_SCHEMA` in `gitnexus/src/core/lbug/schema.ts`. At minimum:

```sql
FROM `Const` TO Method,
FROM `Const` TO Function,
FROM `Const` TO Constructor,
FROM `Const` TO Class,
FROM `Const` TO Interface,
FROM `Const` TO `Struct`,
FROM `Const` TO `Enum`,
FROM `Const` TO `Property`,
FROM `Const` TO CodeElement,
```

Similarly, `Variable` and `Static` are also linkable labels used in Case 5 and may need the same treatment — they currently only have `FROM Variable TO Community/Process` and `FROM Static TO Community/Process`.

For reference, `Constructor` has 17 outgoing `FROM` pairs, while `Const` has only 2.

## Root cause timeline

| Date | Commit / PR | What changed | Bug impact |
|------|-------------|-------------|------------|
| 2026-03-15 | `5a585083` | LadybugDB migration creates `schema.ts` with `Const` as a leaf-only node | **Latent** — schema is incomplete from the start |
| 2026-05-22 | `dd352732` / #1718 | Case 5 + `isLinkableLabel('Const')` added — `Const` nodes can now be CALLS sources | **Bug introduced** — `Const→Method` edges are produced but schema doesn't declare the pair |
| 2026-06-15 | `df08ecc3` / #2215 | `RelPairRouter` class introduced (perf optimization) | No change — undeclared pairs still silently written |
| 2026-07-31 | `de84ad62` / #2740 | `declaredPairs` fail-fast check added to `RelPairRouter` | **Bug exposed** — `Const→Method` edges now throw an error |

lehenbauer:
## Problem

On a large Swift codebase, `gitnexus analyze` can fail with:

> Relationship label pair Enum→Property is not declared in the LadybugDB relation schema

The failure occurs when Swift enum members are emitted for persistence. Swift enums can contain computed properties, methods, initializers, and nested types; method bodies can contain local variables; and computed-property bodies can define nested functions and types.

## Root cause

The `CodeRelation` DDL did not declare every source/target label pair used by Swift member containment. `RelPairRouter` correctly rejected the first emitted undeclared pair instead of sending an edge to a relation table that LadybugDB does not have.

## How the complete set was enumerated

The undeclared set was collected in one analyze pass by temporarily changing the built router to log and skip undeclared pairs. This exposed the complete set in one run instead of requiring a fix, rebuild, and rerun for each first failure.

## Fix

Declare the complete observed member-containment set in the `CONTAINS` relationship DDL:

- `Enum` → `Function`, `Method`, `Struct`, `Constructor`, `Property`, `TypeAlias`
- `Method` → `Variable`
- `Property` → `Class`, `Enum`, `Function`, `Struct`

The schema version is bumped from 30 to 31. A pre-v31 database physically lacks the new relationship tables, so an incremental top-up that emits one of these edges would fail at the database layer; existing indexes must be fully re-analyzed. The numeric value 31 may need renumbering at merge time; please re-check it against the upstream constant alongside the existing collision notes.

## Tests

- Added a regression test that asserts every listed DDL pair and fails on unpatched upstream.
- Red: `test/unit/schema.test.ts` reported `1 failed | 39 passed (40)` with `AssertionError: expected ... to contain 'FROM \`Enum\` TO Function'`; exit code 1.
- Green: the required focused set passed with `Test Files 10 passed (10)` and `Tests 149 passed (149)`.
- The schema-version reuse-gate test now rejects v30 and accepts v31.

## Validation

- `npx tsc --noEmit` — passed.
- `npm run build` — passed.
- The isolated Swift fixture analyzed successfully with the built CLI; the graph readback contained the enum computed-property edge (`Example` → `computed`, `HAS_PROPERTY`).


gitnexus/src/core/lbug/schema.ts
```diff
@@ -376,6 +377,12 @@ CREATE REL TABLE ${REL_TABLE_NAME} (
   FROM \`Enum\` TO Community,
   FROM \`Enum\` TO Class,
   FROM \`Enum\` TO Interface,
+  FROM \`Enum\` TO Function,
```
abhigyanpatwari, line 381: **P0 — the declared set is still incomplete: `analyze` exits 1 on ordinary Rust and TS code at this head, including on `gitnexus/src` itself.**

Running the real `gitnexus analyze` from a build of this exact head:

| corpus | result |
|---|---|
| `impl Point { pub fn new() … }` | **exit 1** — `Relationship label pair Impl→Function is not declared in the LadybugDB relation schema` |
| `pub trait Shape { fn area(&self) -> i32; }` | **exit 1** — `… Trait→Function …` |
| `export const cfg = { m() { return 1; } };` | **exit 1** — `… Const→Method …` |
| `var legacy = { run() { return 2; } };` | **exit 1** — `… Variable→Method …` |
| full copy of `gitnexus/src` (779 files) | **exit 1** — `… Const→Method …` (244 such edges) |
| `swift-only/shape.swift` | **exit 0** — fixed by this PR ✅ |

`HAS_METHOD` is in `RETAINED_REL_TYPES` (`graph-emit-sink.ts:138-147`), so these always reach the throwing guard at `rel-pair-routing.ts:127-131`; `csv-generator.ts:806-812` rethrows and nothing catches it before the CLI.

Why the existing rows do not cover it: a Rust `impl`/`trait` member is minted as a **`Function`**, not a `Method` (`tree-sitter-queries.ts:1437`), so the declared `Impl→Method` (`schema.ts:399`) and `Trait→Method` (`:395`) never fire. An object-literal owner is labelled verbatim `Const`/`Variable` (`ast-helpers.ts:1212`).

To be clear on ownership: **this PR did not introduce the break** — #2740 (`de84ad629`) did, by converting a pre-existing silent COPY drop into a hard abort without auditing what the emitters produce. But this PR's `INCREMENTAL_SCHEMA_VERSION` bump force-marches every existing index into that path on its next run (`run-analyze.ts:1214-1221`), so shipping as-is makes a latent break active for Rust/TS users.

Smallest fix, here, while the bump is already being spent:

```
FROM `Impl`     TO Function,
FROM `Trait`    TO Function,
FROM `Const`    TO Method,
FROM `Variable` TO Method,
```

Note these are not established to be the whole set: a second lane, using a pre-#2740 instrumented build over the repo's own `test/fixtures/lang-resolution` corpus, observed 13 further pairs (`Module→*`, `CodeElement→*`, `Namespace→Function`, `Record→Record`, `Trait→Trait`, `Function→File`) that all throw against this head's schema — their emission was not re-verified on a PR-head build, so treat it as an open question rather than a fact. That is exactly the argument for the durable fix (still absent repo-wide): a test asserting *emitted pairs ⊆ `parseRelationSchemaPairs(RELATION_SCHEMA)`* over a multi-language fixture corpus, which is what the #2740 review prescribed.

abhigyanpatwari (commented):
# Production-readiness review — PR #2769 `fix(schema): declare Swift member-containment pairs in CONTAINS DDL`

Head `402de1bc4` · base `51095c19f` (== `origin/main` at review time) · 4 files, +50/−4.

## Provenance and engine breakdown (read first — this review is degraded)

**Claude** ran: the GitNexus PR swarm (`facts-historian`, `risk-architect`, `test-ci-verifier`), the `ce-code-review` persona set (correctness, adversarial, data-migration, testing, maintainability, project-standards), and two coordinator-dispatched targeted lanes — one verifying two unfinished Codex leads, one dedicated end-to-end **reproduction** lane that compiled PR-head sources and ran the real `gitnexus analyze` CLI.

**Codex (GPT-5.5/xhigh)** ran swarm + ce + adversarial, and **all three legs died on `You hit your spend cap set by the owner of your workspace`** before producing structured findings. **Codex unavailable — both-engine review degraded to Claude-only.** Nothing below is cross-engine corroborated. Where a Codex *lead* is cited, it was confirmed by a separate Claude lane and is described as such.

Evidence tiers: `[reproduced]` = executed against a real build/engine; `[code-read]` = source reading. Same-engine agreement is never presented as independent confirmation.

---

## 1. Review bar for this PR

No `DoD.md` or `TESTING.md` exists in the repo; the bar below is derived from `CONTRIBUTING.md:70-76` (DoD checklist), `AGENTS.md`/`GUARDRAILS.md` (scope, minimal diffs), and the change's own domain.

| # | Acceptance criterion | Status |
|---|---|---|
| B1 | Tests and typecheck pass for changed files | **MET** — `47 passed` across both changed test files; `tsc --noEmit` clean |
| B2 | PR title follows `<type>(scope): subject` | **MET** — `fix(schema): …`, an established scope for this change class |
| B3 | Diff is minimal, no unrelated churn | **MET** — 4 files, all required |
| B4 | The DDL change is syntactically valid and parses through the runtime's own consumer | **MET** — 206→217 pairs, +11/−0, all recovered by `parseRelationSchemaPairs` |
| B5 | The reported failure no longer reproduces | **MET** — `Enum→Property` accepted at head, threw at base |
| B6 | Persisted-shape change is gated by the correct version constant, and that gate actually re-creates the DDL | **MET** — `INCREMENTAL_SCHEMA_VERSION` 30→31, wipe→`initLbug` path traced; `SCHEMA_BUMP` correctly untouched |
| B7 | **The DDL declares what the emitters actually produce — `analyze` does not abort on ordinary supported-language code** | **NOT MET** — see Finding 1 |
| B8 | The fixed behavior is covered by a committed regression test that exercises the failing path | **NOT MET** — see Finding 6 |
| B9 | No silent data loss introduced or left reachable on any emit path | **NOT MET (pre-existing)** — see Findings 2 and 3 |

B7 is the binding criterion. B1–B6 are all met, and generously so.

## 2. Problem being solved

`gitnexus analyze` aborts on Swift codebases with `Relationship label pair Enum→Property is not declared in the LadybugDB relation schema`. GitNexus stores every edge in a single `CodeRelation` REL TABLE (`schema.ts:252`) whose `FROM x TO y` list is a hand-maintained allow-list; since #2740, `RelPairRouter.route` **throws** when an emitted pair is absent from that list (`rel-pair-routing.ts:127-131`). Swift enums carry methods, computed properties and initializers whose owner→member pairs were never declared. The PR adds 11 pairs and bumps `INCREMENTAL_SCHEMA_VERSION` so existing indexes rebuild against the widened DDL.

No linked or closing issue; motivation lives only in the PR body and the new code comment.

## 3. Current PR state

**OPEN**, not draft. Author `lehenbauer` (Karl Lehenbauer), head `lehenbauer:upstream-pr/swift-contains-pairs` @ `402de1bc4877f8ea1099143761ee3dc79d62c5b4`, created 2026-07-31T18:23:36Z. One commit, no labels, no submitted reviews. The only comment is the Vercel authorization bot.

Changed files:
- `gitnexus/src/core/lbug/schema.ts` — +11 FROM/TO pairs (`:335`, `:380-385`, `:414-417`)
- `gitnexus/src/storage/repo-manager.ts` — `INCREMENTAL_SCHEMA_VERSION` 30→31 (`:651`) + an 11-line ladder entry (`:639-649`)
- `gitnexus/test/unit/schema.test.ts` — new `it('declares the Swift member-containment pairs')` (`:215-233`)
- `gitnexus/test/unit/call-summary-schema-version.test.ts` — version pin 30→31 and reuse-gate expectations

Pairs added: `Method→Variable`; `Enum→{Function, Method, Struct, Constructor, Property, TypeAlias}`; `Property→{Class, Enum, Function, Struct}`.

**Merge state: `review blocked`.**

GitHub reports OPEN / MERGEABLE / `mergeStateStatus BLOCKED`, but the mechanical block is only Vercel's fork-deploy authorization — not a code gate. The binding reason is Finding 1 below: a reproduced P0-severity correctness gap. Hence `review blocked` rather than `checks failing` or `mergeable`.

CI snapshot at review time — **all substantive checks pass**: Analyze js/ts + python, Build & Push (gitnexus + web), CodeQL, quality format/lint/typecheck/typecheck-web, ubuntu coverage 1–3, macos 1–3, eval containment (ubuntu + windows), benchmarks (`GITNEXUS_BENCH`), tree-sitter ABI, packaged install smoke, node floor compat, gitleaks, autofix. **Vercel fails only on fork-deploy authorization.** **One Windows platform-sensitive shard was still pending** at snapshot; the other two passed. No workflow files changed.

Branch is fully up to date with main: `git log 51095c19f..origin/main` is empty, and `origin/main`'s tip **is** the merge base.

- **#2740 (`de84ad629`, "feat(spring): index @Bean factories and @Resource injection", authored and committed 2026-07-31 — the same day this PR was opened)** introduced `parseRelationSchemaPairs` **and** the hard `throw` in `RelPairRouter.route`, via sub-commit `9749092f0`. This is the direct cause of the abort this PR is patching. `git show df08ecc39:…/rel-pair-routing.ts | grep -c declaredPairs` → `0`, confirming the check did not exist before it.
- **#2203/#2215 (`2e5ab2be3`, `df08ecc39`, 2026-06-15)** introduced `RelPairRouter` with `validTables` filtering only. Between then and #2740, an undeclared pair meant a late `COPY` rejection plus a lossy row-by-row fallback — **degraded but never fatal**. That is why existing indexes (this repo's own, 20319 symbols) exist at all despite pairs having been missing the whole time.
- **v29 ladder entry (`repo-manager.ts:604-609`)** is exact precedent for this PR's approach: *"The relation DDL gained Class→CodeElement; a pre-v29 database cannot persist that label pair, so force a one-time rebuild against the expanded schema."* Same class of change, same knob, no `SCHEMA_BUMP`.
- **`INCREMENTAL_SCHEMA_VERSION` collisions are a five-times-repeated defect class** in this exact file (`repo-manager.ts:611-613`, `:623-625`). The PR's own comment carries the re-check instruction, matching convention.
- **No prior issue or PR tracks this gap.** `git log --all --grep` plus `gh` searches return only #2769.
- **The durable fix prescribed by the #2740 review — "a static test asserting emittable pairs ⊆ RELATION_SCHEMA" — still does not exist.** `git grep` for `emittable|subset|parseRelationSchemaPairs` under `gitnexus/test/` returns only `rel-pair-routing.test.ts`, which builds its declared set by hand and therefore can never disagree with the emitters.

**Branch hygiene: `clean feature/fix PR`.**

Evidence: one commit; four files, all causally connected to the single stated problem (DDL + its version gate + the two tests that pin them); no merge commits; no unrelated formatting, refactors, version bumps, workflow churn or web-UI changes; branch fully up to date with `main` (`git log BASE..origin/main` empty). No split or rebase required.

Mandated hygiene checks, all clean:

```
git diff --check 51095c19f...HEAD                                        → no output (exit 0)
git grep -nP '[\x{202A}-\x{202E}\x{2066}-\x{2069}]' -- <4 changed files>  → no matches
non-ASCII on added lines                                                 → only "→" (U+2192)
```

The single non-ASCII character on added lines is `→`, visible punctuation already used 10× in the base version of the same file's version ladder. No hidden or bidi controls anywhere in the diff.

## 7. Understanding of the change, and what it gets right

The change is DDL-only plus its migration gate. It does **not** touch any emitter, any resolver, or any capture shape — it widens the allow-list of endpoint label pairs the single `CodeRelation` table accepts, and forces existing indexes to rebuild so they physically gain that widened table.

Verified strengths — each of these was a real suspicion that a lane checked and cleared:

- **The 11 pairs are correctly spelled and escaped.** Parsed with the runtime's own `parseRelationSchemaPairs` (not a naive grep — the DDL is a template literal with backslash-escaped backticks, a trap that has produced wrong counts before): base **206** unique pairs → head **217**, **+11 / −0**, no duplicates, all 11 recovered by the parser. Computed independently by three lanes and by the coordinator. `[code-read]`
- **The reported symptom really is fixed.** Driving the **base** `RELATION_SCHEMA` through the PR-head router over a Swift corpus: `ROUTER THREW after 150 edges: Relationship label pair Enum→Property is not declared…`; the same corpus routes clean at head. The repro lane's `swift-only/shape.swift` corpus exits **0** at head and would have failed at base. `[reproduced]`
- **The version bump is necessary *and* sufficient**, traced rather than assumed. `runSchemaCreationQueries` re-runs `CREATE REL TABLE` on every writable open (`lbug-adapter.ts:907-908`, `:659-700`), but LadybugDB has no `ALTER … ADD FROM/TO` and the `already exists` error is **suppressed** at `lbug-adapter.ts:693` — so an existing DB can never gain the pairs by re-running the DDL. What saves it is the full-rebuild path the bump triggers: `run-analyze.ts:1214` sets `force` → `isIncremental` false (`:1546-1554`) → `wipeLbugDbFiles(buildPath)` (`:1677`) → `initLbug(buildPath)` (`:1700`) runs the DDL on an empty file. **Without the bump this would have been a silent no-op.** `[code-read, 4 lanes]`
- **`SCHEMA_BUMP` correctly left alone.** `parse-cache.ts:149` (=36) gates `ParseWorkerResult`/capture shape; this diff changes no capture. Moving it would needlessly cold-start every parse cache, and the warm cache surviving is what bounds the forced rebuild's cost to resolution + emit + COPY rather than a full re-parse. The recurring `INCREMENTAL_SCHEMA_VERSION ≠ SCHEMA_BUMP` confusion did **not** recur here. `BRIDGE_SCHEMA_VERSION` and the identity/embedding versions correctly stay put too.
- **The new test genuinely discriminates.** All 11 literal strings are absent at base and present at head (verified against `git show 51095c19f:…/schema.ts`); `NODE_TABLES` contains no label that is a prefix of another used here, so no substring-collision pass; `RELATION_SCHEMA` is DDL-only with no comments, so the "passes via a comment" trap does not apply. `[reproduced]`
- **Standards compliance is clean.** The project-standards lane found **no findings above confidence 25**. Title format, ladder-entry template, CHANGELOG (not owed for a fix PR), `ARCHITECTURE.md` (lists relation *types*, not pairs — does not go stale), scope minimalism all check out. Naming "Swift" here is fine: the no-language rule is scoped to `src/core/ingestion/`, and this file already names ~10 languages. One recalled "rel-type registration checklist" lead was checked and **dropped as stale** — that text does not exist in this repo's standards files.
- **Rollback is safe.** A revert stamps 30 against on-disk 31; `!==` trips and forces a rebuild against the old DDL. No corruption, no half-migrated state — it simply restores today's `main` bug (see Finding 5 for the one narrow hazard).

## 8. Findings

### Finding 1 — P0 — Four undeclared `HAS_METHOD` pairs still abort `analyze` on ordinary Rust and TypeScript code, including on GitNexus's own source tree · `[reproduced]`

- **Risk:** At PR head, `gitnexus analyze` exits 1 on any Rust repo containing an `impl` block or a `trait` with methods, and on any JS/TS repo containing `export const x = { m() {} }` or `var x = { m() {} }` — including `gitnexus/src` itself (244 such edges). `Impl→Function`, `Trait→Function`, `Const→Method` and `Variable→Method` are emitted but undeclared, and their relationship type `HAS_METHOD` is in `RETAINED_REL_TYPES` (`graph-emit-sink.ts:138-147`), so they are always forwarded to the real graph and always reach the throwing guard, on every invocation mode. **This PR did not introduce the break — #2740 did** — but this PR's version bump force-marches every existing pre-v31 index into that path on its very next run (`run-analyze.ts:1214-1221`), converting a latent break into an active one for every Rust and TS/JS user on upgrade. That is the specific reason it blocks: not regression, but making an unmet invariant unavoidable while the PR's stated purpose is to make the DDL match the emitters.
- **Evidence to check:** the six-corpus reproduction table in §9a (real `gitnexus analyze`, PR-head build, with the full stack trace); throw at `rel-pair-routing.ts:127-131`; rethrow at `csv-generator.ts:806-812` with no catch before the CLI; `HAS_METHOD` literal at `graph-emit-sink.ts:141`. Why the existing rows do not cover it: a Rust `impl`/`trait` member is minted as a **`Function`** node, not a `Method` (`tree-sitter-queries.ts:1437`; owners at `ast-helpers.ts:405,408`; edge at `parse-worker.ts:2856-2868`), so the declared `Impl→Method` (`schema.ts:399`) and `Trait→Method` (`:395`) are decoys that never fire; and an object-literal owner is labelled verbatim `Const`/`Variable` at `ast-helpers.ts:1212` — `const ownerLabel = declaration?.type === 'variable_declaration' ? 'Variable' : 'Const';`.
- **Recommended fix:** add four lines to `RELATION_SCHEMA` in this PR, while the version bump is already being spent:
  ```
  FROM `Impl`     TO Function,
  FROM `Trait`    TO Function,
  FROM `Const`    TO Method,
  FROM `Variable` TO Method,
  ```
  Durable fix (out of scope for this PR but the real answer, and the one #2740's own review prescribed): a test deriving the emitted pair set from a multi-language fixture corpus and asserting it is a subset of `parseRelationSchemaPairs(RELATION_SCHEMA)`. A hand-maintained 217-entry cross product cannot be kept complete by enumeration, and the failure mode of one missing pair should not be "index nothing".
- **Blocks merge: yes**

### Finding 2 — P2 — The streamed emit sink applies no declared-pair gate, so an undeclared pair is lost silently instead of loudly on explicit `--force` · `[reproduced, verify lane]`

- **Risk:** `GraphEmitSink.addRelationship` and `PdgEmitSink.addRelationship` check only `validTables` (node labels) and then open a per-pair CSV writer for **any** pair. On an explicit `--force` run (streaming on) an undeclared non-retained pair therefore reaches `COPY`, fails at the binder, is only pushed to a `warnings` array that has no consumer, and the row-by-row fallback re-fails per edge into a bare `catch {}` — the run returns `success: true` with `insertedRels` still counting the lost rows and exits 0. Same repo, opposite behavior depending on invocation. Pre-existing (#2740 added the gate to the router only), not introduced here. **Second-order consequence specific to this PR:** the author enumerated the 11 pairs "by temporarily changing the built router to log and skip undeclared pairs" — an instrument that only sees what reaches `RelPairRouter`. On a `--force` run every non-retained type (`CALLS`, `ACCESSES`, `USES`, `CONTAINS`) bypasses it, so the "complete observed set" is under-counted by construction unless the harvest ran with streaming off.
- **Evidence to check:** `graph-emit-sink.ts:408-414` (only `validTables`, then `const pairKey = …` with no gate) versus its own header claim at `:87-90` that it "reuses … `RelPairRouter` validity check, so the streamed row SET equals the whole-graph emit's"; identical shape at `pdg-emit-sink.ts:56,162-181`; neither imports `RelPairRouter`/`parseRelationSchemaPairs`. Cascade at `lbug-adapter.ts:1220-1236` (warn-only) → `:1246-1264` (fallback) → `:1418-1428` (`catch {}`) → `:1194`/`:1306` (`success: true`, inflated count); the in-source admission that `warnings` has no consumer is at `:1225-1227`. Empirical probe against the real `@ladybugdb/core` 0.18.3 with this PR's own 217-pair schema: `COPY → Binder exception: Rel table CodeRelation does not contain Enum-Variable from-to pair.` → `IGNORE_ERRORS` retry fails identically → fallback `CREATE` fails per edge → `persisted edges: 0`. *This was a Codex lead, confirmed by a Claude lane.*
- **Recommended fix:** export the check from `rel-pair-routing.ts` (e.g. `assertDeclaredPair`) and call it from both sinks immediately after the `validTables` gate, before opening the writer; correct the two overclaiming headers; add one unit test per sink asserting an undeclared pair is rejected. Separately, classify the COPY error — an undeclared-pair binder exception is a schema bug, not a transient IO fault, so it should throw the way `copyNodeCSVs` already does (`lbug-adapter.ts:991-1001`) rather than warn.
- **Blocks merge: maybe** — pre-existing and not worsened by this diff, but it is the reason a missing pair #12 will not announce itself, which is exactly the risk profile this PR is operating in.

### Finding 3 — P2 — `Method→Const`, the exact sibling of the `Method→Variable` this PR adds, is still undeclared and is emitted today · `[reproduced]`

- **Risk:** 8 `Method→Const` `ACCESSES` edges exist in `gitnexus/src` alone. `ACCESSES` is not retained, so under `--force` they take the streamed sink and vanish with no warning and exit 0; with streaming off (the auto-forced upgrade rebuild) the same pair instead routes through the throwing guard and aborts. Either way the pair is wrong to omit, and omitting it while adding its sibling is the kind of asymmetry that makes the next reader assume the set was derived by rule when it was derived by observation.
- **Evidence to check:** causal A/B on a one-file corpus where `Method→Const` is the only undeclared pair, the sole delta being one extra `FROM Method TO \`Const\`,` patched into the built schema — arm A (PR head): 168 edges, **0** `ACCESSES` in the DB, exit 0, no warning; arm B: 169 edges, **1** `ACCESSES`, exit 0. Cypher readback `MATCH (a)-[r:CodeRelation]->(b) WHERE r.type='ACCESSES'` returns `[]` in A and the edge in B. Sample edge: `Method:…/c-cpp-harvest.ts:CCppHarvester.declareParams#1 -> Const:…:CCppHarvester.registerResultDefs.root@445:4`. `RETAINED_REL_TYPES` at `graph-emit-sink.ts:138-147` shows `ACCESSES` absent. Note `reconcile-ownership.ts:90,203` treats `Method`/`Function`/`Constructor` **symmetrically** as owners, which is why `Function→Variable` and `Constructor→Variable` are also worth a look.
- **Recommended fix:** add `FROM Method TO \`Const\`,` next to the added `FROM Method TO \`Variable\`,` at `schema.ts:335`.
- **Blocks merge: maybe** — it is one line, it is free while the bump is being spent, and leaving it out means known silent edge loss.

### Finding 4 — P3 — The v31 ladder comment misstates the edge type, the table model and the language scope · `[reproduced]` (language scope) / `[code-read]` (edge type, table count)

- **Risk:** This comment is the only durable record of why v31 exists; the next maintainer sizing a DDL change will reason from it and will be sent to the wrong file. Three inaccuracies: (a) **not `CONTAINS`** — these pairs are carried by `HAS_METHOD`/`HAS_PROPERTY`, while `CONTAINS` is the Folder/File tree plus COBOL/markdown, and the shipped tool docs say so, so the note contradicts them; (b) **not "new rel tables"** — there is exactly one `CodeRelation` REL TABLE with the kind as a property, so each pair declared here is simultaneously legal for *every* relation type, a widening that is cheap to do and impossible to undo without another forced rebuild; (c) **not Swift-specific** — `Enum→Method` was reproduced from a **Java** enum and a **PHP** enum, so this PR also un-breaks Java/PHP/Kotlin enums with methods. Additionally, "an incremental top-up … fails at the DB layer" understates it: on the streamed path it does not fail, it silently drops — which is a *stronger* argument for the bump.
- **Evidence to check:** `repo-manager.ts:640-649` (the comment); `parse-worker.ts:2855-2868` (`const memberEdgeType = nodeLabel === 'Property' ? 'HAS_PROPERTY' : 'HAS_METHOD';`); `structure-processor.ts:30-41` plus the markdown/COBOL processors and `emit-references.ts:156` for the real `CONTAINS` producers; `local-backend.ts:346` and `mcp/resources.ts:494` for the shipped docs; `schema.ts:252` for the single REL TABLE; the silent-drop chain in Finding 2.
- **Recommended fix:** reword along the lines of *"v31: the relation DDL (the single shared `CodeRelation` REL TABLE) gains eleven FROM/TO pairs carried by `HAS_METHOD`/`HAS_PROPERTY` and scope-resolution edges — observed on Swift, also reached by Java/PHP enum members. A pre-v31 database physically lacks the corresponding from-to pairs, so an incremental top-up emitting one of these edges fails the bulk COPY, falls through the per-edge fallback's `catch {}` (lbug-adapter.ts:1426) and silently drops the edges — analyze still exits 0. Force a full re-analyze."* The v29 entry two paragraphs up gets all of this right and is the model to copy. Worth correcting the PR/commit subject at merge too.
- **Blocks merge: no**

### Finding 5 — P3 — Version-number allocation: free today, but the `===` gate makes a revert hazardous · `[code-read]`

- **Risk:** `31` is unclaimed right now, but this constant has collided five times historically and `origin/main` can advance while the PR waits. Separately, the reuse gate is exact equality, not `<` — so if this PR is ever **reverted** and `31` is later reallocated, indexes already stamped 31 by this build will satisfy `=== 31` against a differently-shaped DB and take an incremental top-up onto a mismatched schema.
- **Evidence to check:** `repo-manager.ts:651` (=31); `git show origin/main:…/repo-manager.ts` (=30) and `git log 51095c19f..origin/main` (empty); the gate at `run-analyze.ts:1546-1549`; every open PR touching `repo-manager.ts` was checked and none claims 31 (#2716, #2739, #2732, #2770, #2765, #2726, #2725, #2543, #2493, #2443 — not exhaustive across all ~30 open PRs, but exhaustive across the ones touching this file).
- **Recommended fix:** re-run the collision check immediately before merge, not at review time (the PR's own comment already says this). If the PR is ever reverted, do **not** free the number — leave a tombstone in the ladder (`v31: RESERVED, reverted in #NNNN, do not reuse`) and start the next allocation at 32.
- **Blocks merge: no**

### Finding 6 — P2 — No committed test exercises the path that was failing, and the class-level guard is still absent · `[code-read]`

- **Risk:** The only new coverage is a substring assertion on the DDL *string*. It proves the text was edited; it proves nothing about whether the emit side produces exactly those pairs, and it cannot fail when a future emitter introduces pair #12 — which is precisely how this defect reached a user in the first place. The PR body's own validation ("temporarily changing the built router to log and skip undeclared pairs"; "the isolated Swift fixture analyzed successfully with the built CLI") was manual, one-off, and never captured in CI. There is also no COPY round-trip for any of the 11 new pairs, no parity test between the router and the two streamed sinks, no test pinning that the DDL actually re-runs after a bump (the `already exists` suppression is the trap that would hide it), and no coverage of the `lbug-adapter.ts:1426` `catch {}`.
- **Evidence to check:** `schema.test.ts:215-233` constructs no `RelPairRouter`, calls no `parseRelationSchemaPairs`, runs no `streamAllCSVsToDisk`. The motivating fixture already exists (`test/fixtures/lang-resolution/swift-enum-members/Direction.swift`) and is used — but with `skipGraphPhases: true` (`test/integration/resolvers/swift.test.ts:1325-1328`), and `rg -n "streamAllCSVsToDisk|RelPairRouter|COPY " test/integration/resolvers/swift.test.ts` returns nothing. `test/integration/csv-pipeline.test.ts:78-153` is the established home for driving a real `KnowledgeGraph` through the CSV/router layer. The repo already has the exact collector-and-validate pattern needed for the class-level guard, in `test/integration/grammar-literal-validation.test.ts` (including a companion "runs non-vacuously" test at `:141-149`).
- **Recommended fix:** (a) **one line, and it is the regression test for Finding 1** — add `'Function'` to `const targets = ['Method'];` at `schema.test.ts:240`; that pre-existing test's own comment already lists `Trait` and `Impl` as `HAS_METHOD` sources, and `parse-worker.ts:2857` picks `HAS_METHOD` for every non-`Property` member label including `Function`. It fails today until `Impl→Function`/`Trait→Function` are declared. (b) Add an integration test that runs `runPipelineFromRepo` over the Swift fixture **without** `skipGraphPhases` and feeds the graph through `streamAllCSVsToDisk` the way `csv-pipeline.test.ts` does, asserting no router throw and that the `Enum→Function` edge is queryable after `loadGraphToLbug`. (c) Land the class-level *emittable ⊆ declared* guard modelled on `grammar-literal-validation.test.ts`.
- **Blocks merge: maybe** — (a) is trivial and belongs here; (b) and (c) are legitimately follow-up-sized.

### Finding 7 — P3 — The new test re-encodes the DDL as backtick-exact raw strings instead of asserting through the canonical parser · `[code-read]`

- **Risk:** A purely cosmetic DDL change (dropping an unnecessary backtick, adjusting whitespace) breaks this test with no semantic change, and the test cannot recognise a pair that is semantically present but spelled differently. Mitigating: this matches the file's own existing convention, so it is a convergence opportunity rather than a regression this PR introduced.
- **Evidence to check:** `schema.test.ts:215-233` versus `rel-pair-routing.ts:58-65`, whose doc comment says it exists precisely so callers need not hand-match backtick quoting; `csv-generator.ts:26` already consumes it. The adjacent pre-existing test at `schema.test.ts:235-262` uses the same hand-maintained `sources`/`backtickSources` convention. The `for`-loop-with-`expect` shape has no branching and matches local convention — not a deviation.
- **Recommended fix:** `expect(parseRelationSchemaPairs(RELATION_SCHEMA)).toContain('Enum|Function')`, keyed the way the production router keys it; ideally converge the older `HAS_METHOD` test the same way.
- **Blocks merge: no**

## 9. PR-specific assessment

### 9a. Reproduction — what the PR-head CLI actually does

The reproduction lane compiled PR-head sources into a scratchpad (worktree untouched; `diff -r` confirmed byte-identical src, `git status --porcelain` empty throughout) and ran the real `gitnexus analyze` binary. Each pair was isolated in its own micro-corpus so the first throw could not mask the rest.

| corpus | content | result |
|---|---|---|
| `rust-only/point.rs` | `impl Point { pub fn new() … }` | **exit 1** — `Relationship label pair Impl→Function is not declared in the LadybugDB relation schema` |
| `m-trait/t.rs` | `pub trait Shape { fn area(&self) -> i32; }` | **exit 1** — `… Trait→Function …` |
| `m-const/c.ts` | `export const cfg = { m() { return 1; } };` | **exit 1** — `… Const→Method …` |
| `m-var/v.ts` | `var legacy = { run() { return 2; } };` | **exit 1** — `… Variable→Method …` |
| `gnsrc/` | full copy of `gitnexus/src`, 779 `.ts` files | **exit 1** — `… Const→Method …` (244 such edges) |
| `swift-only/shape.swift` | enum + computed property + method | **exit 0** — fixed by this PR ✅ |

```
Analysis failed: Relationship label pair Impl→Function is not declared in the LadybugDB relation schema
    at RelPairRouter.route      (dist/core/lbug/rel-pair-routing.js:113:19)
    at streamAllCSVsToDisk      (dist/core/lbug/csv-generator.js:666:43)
    at async loadGraphToLbug    (dist/core/lbug/lbug-adapter.js:934:15)
    at async runFullAnalysisInner (dist/core/run-analyze.js:1583:13)
    at async runFullAnalysis    (dist/core/run-analyze.js:474:16)
    at async analyzeCommandImpl (dist/cli/analyze.js:1106:24)
EXIT=1
```

The 244-edge count in `gitnexus/src` is stable at 41 444 relationships / 11 817 nodes with the production default `keepLocalValueSymbols=false`, and identical with it set true (32 780 nodes / 62 407 rels) — so it is not an artifact of the local-symbol pruner. A static regex approximation (56 owner object literals, 225 shorthand methods) corroborates the same order of magnitude.

### 9b. How complete is the undeclared set? — **not established; two lanes disagree on scope**

This matters because the PR's implicit claim is that the enumerated set is complete, and because it is the argument for whether four extra lines are sufficient.

- **Reproduction lane (PR-head build, real CLI):** over its corpora — mixed Rust+TS+Swift, per-pair micro-corpora, an 8-language const corpus, and the full `gitnexus/src` tree at 41 444 relationships across 37 distinct emitted pairs — the emitted-but-undeclared set was exactly `{Impl|Function, Trait|Function, Const|Method, Variable|Method, Method|Const}`, with nothing beyond those five. It did **not** run the repo's own `test/fixtures/lang-resolution` corpus.
- **Correctness lane (pre-#2740 instrumented build, so it could enumerate instead of aborting on the first pair):** over the repo's own `test/fixtures/lang-resolution` corpus it observed **13 further pairs** that then **THROW** when replayed through the **PR-head** router against the **PR-head** schema: `Module→{CodeElement, Constructor, Namespace, Property, Record}`, `CodeElement→{CodeElement, Module, Property, Record}`, `Namespace→Function`, `Record→Record`, `Trait→Trait`, `Function→File`.

**Reconciliation, honestly:** the *declared-set* half of the correctness lane's result is PR-head-verified (those 13 pairs are genuinely absent from the 217 and genuinely throw). The *emission* half came from a pre-#2740 build; the lane states its emitters were re-read in PR-head source and are byte-identical in behavior, but nobody re-ran that fixture corpus with a PR-head build to confirm the 13 are still emitted. The adversarial lane did route one file from that corpus (`test/fixtures/lang-resolution/rust-abstract-dispatch`) through the PR-head router and hit `Trait→Function` — consistent, but it covers only one fixture.

**Therefore: the true undeclared set is at least five and may be as large as eighteen. It is not established to be five.** This does not soften Finding 1 — it sharpens it. The five reproduced pairs are enough to block on their own, and the possible thirteen are the strongest available argument that enumeration cannot close this class and that the *emittable ⊆ declared* guard is the real fix. Verifying the 13 against a PR-head build over `test/fixtures/lang-resolution` is carried into Open Questions.

### 9c. Cross-lane contradictions, resolved

**(i) "This PR's bump forces every existing index onto the *streaming* path" — WRONG.** Three lanes asserted it; the coordinator disproved it. `resolveStreamGraphEmit` is resolved **once**, early, at `run-analyze.ts:894`; the version-mismatch sets `force` **later**, at `:1214`. `grep -n "resolveStreamGraphEmit\|streamGraphEmitActive"` returns exactly `671, 894, 1495, 1500` — no second resolution. So on the auto-forced rebuild `streamGraphEmitActive` was already computed from options where `force` was unset ⇒ **streaming is OFF** ⇒ everything takes the **loud router** path, consistent with the loud error the reporter saw. The silent streamed path requires the user to pass `--force` themselves. `claude-ce-data-migration` got this ordering right and drew the correct conclusion (and separately noted the auto-forced rebuild therefore runs in the highest-memory, non-streamed configuration — its sibling `resolveStreamPdgEmit` *is* evaluated after the force mutation, at `:1491`, so on a `--pdg` auto-forced rebuild the two toggles disagree on the same run; without `--pdg` both are simply off). `claude-lead-verify-emit-sinks`, `claude-ce-correctness` F2 and `claude-ce-adversarial` F2 built on the inverted premise and are corrected here.
Two knock-on corrections: the sub-claim that "the exact edge class this PR fixes travels the silent path" is doubly wrong — these pairs are carried by `HAS_METHOD`/`HAS_PROPERTY`, both **retained**, so they were always on the loud path. And the findings that stand without that premise — Findings 1, 2 and 3 — are unaffected and retained.

**(ii) `Method→Variable` — "dead coverage for Swift" versus the author's empirical observation.** State of evidence, no side taken: three lanes independently found Swift has **no** `@definition.variable` capture (`tree-sitter-queries.ts:1877-1949`) and labels every `func` as `Function`; the only Swift `Method` producer is `protocol_function_declaration` (`:1904`), which has no body and so cannot declare a local. No lane reproduced `Method→Variable` from any Swift corpus. The author did observe it in a log-and-skip pass over a real (probably mixed-language) repo, and `reconcile-ownership.ts:90,203` treats `Method`/`Function`/`Constructor` symmetrically as owners of `Variable`/`Const`/`Static` defs, which is a plausible producer. Over-declaration is harmless; what it undermines is using "Swift member containment" as the argument that the set is complete — the set was harvested empirically, and §9b shows the harvest was corpus-bounded.

**(iii) `Enum→Struct` — containment or CALLS?** Evidence favors CALLS: one lane reproduced `Enum|Struct  n=1  types=CALLS` from an enum-level `static let c = S()`, and `Struct` has `needsOwner === false` (`parse-worker.ts:2270-2274`) so it can never receive a member edge. The pair is real and worth declaring; the *mechanism* stated in the v31 comment ("nested types") is not the one that produces it. Same for `Enum→TypeAlias`, whose label also has `needsOwner === false`.

### 9d. Migration and deploy-window assessment

The forced rebuild is announced at run time (`run-analyze.ts:1216-1219`) and is cheaper than it looks because the warm parse cache survives. No read path enumerates relation tables (`grep` for `SHOW TABLES`/`table_info` across `gitnexus/src` returns nothing), so old-binary/new-index and new-binary/old-index both read fine and the `!==` gate has no "future stamp treated as current" hole. Delete/drift Cypher iterates `NODE_TABLES` with `DETACH DELETE` and is pair-agnostic.

Three deploy-window notes worth stating in the PR body: branch sub-slots each carry their own `meta.json` (`repo-manager.ts:315-320`), so multi-branch indexers pay N rebuilds, not one; on Windows the rebuild wipes in place unless `GITNEXUS_ATOMIC_WINDOWS_SWAP=1` (`run-analyze.ts:1579-1588`, `:1611`, `:1677`), so a crash mid-rebuild leaves no index (pre-existing, but this PR makes every Windows user take that path once); and MCP keeps serving the pre-fix graph until re-analyze with no schema-staleness signal (`mcp/resources.ts:349` reads a different stamp).

Each of these was a real suspicion that a lane checked and killed, so it does not need to be raised with the author:

- **"An existing pre-v31 DB never gets the new pairs, so the abort survives a forced re-analyze."** REFUTED, traced end to end (wipe at `run-analyze.ts:1677` → `initLbug` at `:1700` → DDL on an empty file). This is also exactly why omitting the bump would have been a *silent* failure rather than a loud one.
- **"The new test could pass vacuously or via a comment."** REFUTED twice, independently: all 11 strings absent at base and present at head, no prefix collisions in `NODE_TABLES`, and `RELATION_SCHEMA` contains no comment lines.
- **"Other languages emit undeclared pairs too."** REFUTED for **Ruby, PHP, Kotlin, C#, C++** — a corpus exercising Ruby `module`+`include`+`attr_accessor`, PHP traits/backed enums/promoted ctor params, Kotlin `object`/`companion object`/`enum class`/`typealias`, C# `record`/`struct`/`delegate`/lambda field, and C++ `namespace`/`template class`/`union`/`typedef`/`enum class` routed clean: `ROUTER OK — routed 72 edges with no throw`. Go, Java, Python and TypeScript also routed clean in the mixed corpus. Rust is the outlier, plus the JS/TS object-literal shape. (Scope caveat: this used custom corpora, not `test/fixtures/lang-resolution` — see §9b.)
- **"`resolveCallerNodeId`'s `firstOwnedFallback` opens an unbounded caller-label cross-product."** REFUTED as live — `emitReferencesToGraph` has no production caller (only its own unit test); the live bridge is `emitReferencesViaLookup`, which restricts anchors via `isCallerAnchorLabel` (`graph-bridge/ids.ts:57-68`) to 8 labels plus a File fallback.
- **"Something else enumerates FROM/TO pairs and must move in lockstep."** REFUTED — `REL_SCHEMA_QUERIES` is `[RELATION_SCHEMA]` (`schema.ts:567`); the only pair-derived artifact is `DECLARED_RELATION_PAIRS` (`csv-generator.ts:26`), which is derived, not duplicated; delete/search paths key on node labels; and no `bench/*/baselines.json` pins an edge-count or graph-shape fingerprint these pairs could move (the affected edges were previously aborting or being dropped, so no baseline could have recorded them).
- **"Version-ladder off-by-one or wrong knob."** REFUTED — comment says v31, constant is 31, tests assert `toBe(31)`, `passesReuseGate(30) === false`, `passesReuseGate(31) === true`. `SCHEMA_BUMP`, `BRIDGE_SCHEMA_VERSION`, `ANALYZER_RUNNER_IDENTITY_SCHEMA_VERSION`, `IDENTITY_CACHE_SCHEMA_VERSION` and the embedding/PDG codec versions all correctly stay put.
- **"Over-declaration removes a loud detector for a mis-anchored Spring `INJECTS` edge."** Mechanism traced but **not reproduced** — the only `edgeSource: 'site'` producer found (`frameworks/spring/di-metadata.ts:350`) anchors on a `Method`, and a `@Configuration`/`@Bean` + `@Autowired` fixture produced `Class|Class`. Filed as a risk note, not a finding.
- **Project-standards lane found no violations above confidence 25**, including a recalled "rel-type registration checklist" that turned out not to exist in this repo's standards files and was dropped as a stale lead.

## 11. Open questions

1. **Are the 13 additional pairs from `test/fixtures/lang-resolution` still emitted by a PR-head build?** (§9b.) Cheapest resolution: run the PR-head CLI over `test/fixtures/lang-resolution` and diff the emitted pair set against `parseRelationSchemaPairs(RELATION_SCHEMA)`. This is the same command that would become the durable guard.
2. **Is `Property→Method` missing?** The risk-architect lane ran out of budget before pinning whether `property_declaration` is in `FUNCTION_NODE_TYPES`, i.e. whether `findEnclosingFunctionId` (`parse-worker.ts:910`) attributes calls inside a Swift computed-property body to the `Property` node. If it does, `Property→Method` is emitted and is **not** among the 11 and **not** in the schema — the same abort remains reachable. Neither confirmed nor refuted. `Struct→TypeAlias` sits under the same reasoning (`Class→TypeAlias` at `schema.ts:307` and `Interface→TypeAlias` at `:362` both exist; `Struct` would be the only type-owner that cannot host a `typealias`).
3. **Which invocation mode produced the author's enumeration?** If it was a `--force` run, every non-retained relation type was invisible to the instrument (Finding 2) and the observed set is under-counted by construction. Worth one line in the PR body.
4. **Do `Enum→TypeAlias`, `Method→Variable` and the four `Property→*` pairs have a real producer?** No lane reproduced them from Swift. Harmless to over-declare; asked only because the answer determines whether the v31 comment's stated mechanism can be repaired or should be replaced with "harvested empirically".
5. **Languages not swept:** Dart (whose `mixin_declaration → 'Mixin'` is not even a node table, so those edges are silently filtered by `validTables`), Scala, Swift protocol extensions, COBOL, Zig, Elixir. ~9 of GitNexus's ~20 supported languages were exercised.
6. **The pending Windows shard** had not reported at snapshot.

---

Current visible state is incomplete. I could verify the declared-pair inventory at base and head, the end-to-end `gitnexus analyze` behavior on Rust/TS/Swift corpora and on `gitnexus/src`, the wipe→`initLbug` migration path, the emit-sink and router source, the version-constant collision status against `origin/main`, and the CI check list — but not the emittability of the 13 additional pairs under a PR-head build, the `Property→Method` producer question, the invocation mode behind the author's enumeration, the final Windows shard, or ~11 of the supported languages. The review above treats those missing items as mandatory verification points rather than confirmed facts.

Two further coverage limits belong on the record: Codex was unavailable for the entire review, so no second engine ever saw this diff; and three Claude lanes (`facts-historian`, `risk-architect`, `test-ci-verifier`) stalled mid-investigation and had to be resumed with a deliver-now instruction, so their sweeps are shallower than the others.

## 12. Final verdict

**`not production-ready`**

The PR correctly identifies and fixes 11 genuine declared-pair gaps: the escaping is right, the reported `Enum→Property` abort really is resolved, the version bump is both necessary and sufficient, `SCHEMA_BUMP` is correctly left alone, and the new test genuinely discriminates against base. However, four additional `HAS_METHOD` pairs — `Impl→Function`, `Trait→Function`, `Const→Method`, `Variable→Method` — remain undeclared while actively causing `gitnexus analyze` to exit 1 on ordinary Rust and TypeScript/JavaScript code, reproduced end-to-end against a PR-head build including on GitNexus's own 779-file source tree. This PR did not introduce that break (#2740 did, earlier the same day, by converting a silent degradation into a hard abort), but its `INCREMENTAL_SCHEMA_VERSION` bump force-marches every existing index into the failing path on its next run, which makes a latent break active for every Rust and TS/JS user on upgrade. The four-line fix belongs in this PR while the version bump is already being spent, and a second lane's evidence that up to thirteen further pairs may also be undeclared means enumeration alone cannot be trusted to close this — the *emittable ⊆ declared* guard prescribed after #2740 is the real fix. Once the four pairs are added (ideally with `Method→Const` and the one-line `'Function'` extension to the existing `HAS_METHOD` test), the verdict becomes `production-ready with minor follow-ups`; the remaining P2/P3 items are legitimate follow-ups rather than blockers.

---

*Automated multi-tool digest (Claude-only this run — Codex was unavailable). Verify each claim before acting on it.*

gitnexus/src/core/lbug/schema.ts
```diff
@@ -332,6 +332,7 @@ CREATE REL TABLE ${REL_TABLE_NAME} (
   FROM Method TO Interface,
   FROM Method TO \`Constructor\`,
   FROM Method TO \`Property\`,
+  FROM Method TO \`Variable\`,
```
abhigyanpatwari, line 335: **P2 — `Method→Const` is the exact sibling of this line and is still undeclared; it is emitted today (8× in `gitnexus/src`) and lost silently.**

`ACCESSES` is not in `RETAINED_REL_TYPES`, so on a `--force` run it goes to the streamed sink, which has no declared-pair gate — the `COPY` fails with a binder exception that is only pushed to a `warnings` array with no consumer, and the row-by-row fallback re-fails into `catch {}` (`lbug-adapter.ts:1426-1428`).

Causal A/B on a one-file corpus, the only delta being one extra `FROM Method TO \`Const\`,` in the built schema:

| | edges | `ACCESSES` in DB | exit |
|---|---|---|---|
| A — this head | **168** | **0** | 0 |
| B — + `Method→Const` | **169** | **1** | 0 |

Arm A exited 0 and printed no warning. (With streaming off — i.e. the auto-forced upgrade rebuild — the same pair routes through `RelPairRouter` and aborts instead.) Please add `FROM Method TO \`Const\`,` next to this line. Worth a look at `Function→Variable` / `Constructor→Variable` too: `reconcile-ownership.ts:90,203` treats `Method`/`Function`/`Constructor` symmetrically as owners.

gitnexus/src/core/lbug/schema.ts
```diff
@@ -404,6 +411,10 @@ CREATE REL TABLE ${REL_TABLE_NAME} (
   FROM \`Variable\` TO Community,
   FROM \`Property\` TO Community,
   FROM \`Property\` TO \`Property\`,
+  FROM \`Property\` TO Class,
+  FROM \`Property\` TO \`Enum\`,
+  FROM \`Property\` TO Function,
+  FROM \`Property\` TO \`Struct\`,
```
abhigyanpatwari, line 422: **P2 — context for anyone extending this list: the streamed emit sink applies no declared-pair gate, so a missing pair #12 fails *silently* rather than loudly.**

*(Anchored on this line only because the real file is not part of this PR's diff — the actual location is `gitnexus/src/core/lbug/graph-emit-sink.ts:408-414`, with the identical shape at `pdg-emit-sink.ts:162-181`.)*

Both sinks check only `validTables` (node labels) and then open a per-pair CSV writer for any pair; neither imports `RelPairRouter` or `parseRelationSchemaPairs` — yet `graph-emit-sink.ts:87-90` claims it "reuses … `RelPairRouter` validity check, so the streamed row SET equals the whole-graph emit's". That comment was true before #2740 added the pair gate to the router only.

Probed against the real `@ladybugdb/core` 0.18.3 engine with this PR's own 217-pair schema: `COPY` → `Binder exception: Rel table CodeRelation does not contain Enum-Variable from-to pair.` → `IGNORE_ERRORS` retry fails identically → per-edge fallback fails per edge into `catch {}` → `persisted edges: 0`, with `success: true` and `insertedRels` still counting the lost rows.

Pre-existing, not introduced here — but it matters for this PR because the 11 pairs were enumerated "by temporarily changing the built router to log and skip undeclared pairs". That instrument only sees what reaches `RelPairRouter`; on a `--force` run every non-retained type (`CALLS`/`ACCESSES`/`USES`/`CONTAINS`) bypasses it, so the observed set is under-counted by construction unless the harvest ran with streaming off. Worth naming the run mode in the PR body.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -636,8 +636,19 @@ export interface RepoMeta {
  * multi-line closure binding joins the scope channel and emits its CALLS edge.
  * Pre-v30 indexes keep the wrapper line on unchanged files and would keep
  * failing closed (no edge) through the reuse gate. Force a full re-analyze.
+ *
+ * v31: the CONTAINS DDL gains the eleven Swift member-containment pairs
```
abhigyanpatwari: **P3 — three inaccuracies in the one durable artifact the next maintainer will read.**

1. **Not `CONTAINS`.** These pairs are carried by `HAS_METHOD`/`HAS_PROPERTY` (`parse-worker.ts:2855-2868`: `const memberEdgeType = nodeLabel === 'Property' ? 'HAS_PROPERTY' : 'HAS_METHOD';`). `CONTAINS` is emitted only for the Folder/File tree (`structure-processor.ts:30-41`) plus COBOL/markdown — and `local-backend.ts:346` / `mcp/resources.ts:494` document it that way, so this note contradicts the shipped tool docs.
2. **Not "new rel tables".** There is exactly one `CodeRelation` REL TABLE (`schema.ts:252`) with the kind as a property, so each pair added here is legal for **every** relation type (`CALLS`, `ACCESSES`, `INJECTS`, …), not just containment. Worth saying explicitly — widening is cheap, narrowing needs another forced rebuild.
3. **Not Swift-specific.** `Enum→Method` was reproduced from a **Java** enum and a **PHP** enum. This PR also un-breaks Java/PHP/Kotlin enums with methods; labelling the entry "Swift" sends the next reader to the wrong file.

Also: "an incremental top-up … fails at the DB layer" — on the streamed path it does not fail, it silently drops (binder error → warn → `catch {}` at `lbug-adapter.ts:1426`), which is a *stronger* argument for the bump.

The v29 entry two paragraphs up gets all of this right ("The relation DDL gained Class→CodeElement") and is the model to copy.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -636,8 +636,19 @@ export interface RepoMeta {
...
+ * carry computed properties, methods, initializers, and nested types; method
+ * bodies can declare local variables; and computed-property bodies can define
+ * nested functions and types. A pre-v31 database physically lacks the new rel
+ * tables, so an incremental top-up that emits one of these edges fails at the
+ * DB layer. Force a full re-analyze. As with the collision notes above, 31 may
+ * need renumbering at merge time; re-check this constant against upstream's
+ * value immediately before merging.
  */
-export const INCREMENTAL_SCHEMA_VERSION = 30;
+export const INCREMENTAL_SCHEMA_VERSION = 31;
```
abhigyanpatwari: **P3 — merge-time checklist item (your own comment already flags it; confirming the current state).** `31` is free right now: `origin/main` == this PR's merge base `51095c19f` and carries 30, `git log BASE..origin/main` is empty, and every open PR touching `repo-manager.ts` was checked (#2716, #2739, #2732, #2770, #2765, #2726, #2725, #2543, #2493, #2443) — none claims 31. Re-run that check immediately before merge, not at review time.

One case the note does not cover: the reuse gate is exact equality (`run-analyze.ts:1546-1549`), so if this PR is ever **reverted**, do not free `31` — indexes already stamped 31 would then satisfy `=== 31` against a differently-shaped DB. Leave a tombstone in this ladder and start the next allocation at 32.

gitnexus/test/unit/schema.test.ts
```diff
@@ -212,6 +212,26 @@ describe('LadybugDB Schema', () => {
       expect(RELATION_SCHEMA).toContain('FROM Class TO CodeElement');
     });
 
+    it('declares the Swift member-containment pairs', () => {
```
abhigyanpatwari: **P2 — this test proves the DDL text was edited; nothing here exercises the path that was actually failing.**

Neither new test constructs a `RelPairRouter`, calls `parseRelationSchemaPairs`, or runs `streamAllCSVsToDisk`. The repo already has the motivating fixture (`test/fixtures/lang-resolution/swift-enum-members/Direction.swift`) and a test that uses it — but with `skipGraphPhases: true` (`test/integration/resolvers/swift.test.ts:1325-1328`), so it never reaches the CSV/DB layer. That is exactly why this shipped undetected. `test/integration/csv-pipeline.test.ts:78-153` is the established home for driving a real `KnowledgeGraph` through `streamAllCSVsToDisk`.

Three asks, cheapest first:

1. **One line, and it is the regression test for the P0 above:** add `'Function'` to `const targets = ['Method'];` at `schema.test.ts:240`. That test's own comment already lists `Trait` and `Impl` as `HAS_METHOD` sources; `parse-worker.ts:2857` picks `HAS_METHOD` for every non-`Property` member label including `Function`, which is what Rust uses for all `fn`s. It fails today until `Impl→Function` / `Trait→Function` are declared.
2. Assert through the consumer rather than by raw string — `expect(parseRelationSchemaPairs(RELATION_SCHEMA)).toContain('Enum|Function')` — so a cosmetic backtick/whitespace change in the DDL cannot break the test without a semantic change. (The current shape matches the adjacent pre-existing test's convention, so this is a convergence suggestion for both, not a regression.)
3. The class-level guard (*emitted pairs ⊆ declared pairs*, over a multi-language fixture corpus) is the one that stops the next language feature from hard-aborting `analyze`. The repo already has the collector-and-validate pattern to copy in `test/integration/grammar-literal-validation.test.ts` (including its "runs non-vacuously" companion at `:141-149`); `grep` confirms nothing analogous exists for relation pairs today.

magyargergo:
Addressed the review findings:

- **Finding 1 (P0, blocking):** added the 4 missing `HAS_METHOD` pairs — `Impl→Function`, `Trait→Function`, `Const→Method`, `Variable→Method` — that were aborting `analyze` on ordinary Rust (`impl`/`trait` methods, minted as `Function` not `Method`) and JS/TS (object-literal shorthand methods) code, including on `gitnexus/src` itself.
- **Finding 3 (P2):** added the `Method→Const` sibling of `Method→Variable`, which was being silently dropped.
- **Finding 2 (P2):** exported `assertDeclaredPair` from `rel-pair-routing.ts` and wired it into both streamed sinks (`GraphEmitSink`, `PdgEmitSink`), so an undeclared pair now throws loudly on the streamed `--force` path too, instead of failing the bulk `COPY` and silently vanishing through the per-edge fallback's `catch {}`. Added a regression test per sink.
- **Finding 4 (P3):** reworded the v32 ladder comment — it's the single shared `CodeRelation` table (not new rel tables), the edges are `HAS_METHOD`/`HAS_PROPERTY` (not `CONTAINS`), and the set isn't Swift-only (also reached by Java/PHP enum members, Rust impl/trait members, JS/TS object literals).
- **Finding 5/collision check:** re-checked immediately before merging as your own comment asked — `origin/main` had since taken v31 for #2746 (Python named imports) after this PR branched, so this renumbers to **v32** and merges in `main`. Left a tombstone note for future reverts.
- **Finding 6/7 (P2/P3):** the regression test for Finding 1 is `'Function'`-target coverage on the existing `HAS_METHOD` test plus two new targeted tests naming the Rust and JS/TS producers; all `schema.test.ts` assertions now go through `parseRelationSchemaPairs` (the runtime's own parser) instead of raw-string matching, so a cosmetic DDL formatting change can't break them.

Verified end-to-end: rebuilt the CLI and re-ran all 6 corpora from the review (`rust-only`, `m-trait`, `m-const`, `m-var`, `swift-only`, plus typecheck/build) — all exit 0 now. `tsc --noEmit` clean, and the full `lbug`/`schema`/`rel-pair-routing` unit suite (312 tests) passes.

Not addressed (left as-is, follow-up-sized per the review itself): the durable *emittable ⊆ declared* class-level guard over a multi-language fixture corpus (Finding 6c), and the ~13 additional pairs from `test/fixtures/lang-resolution` that the review flagged as unverified against a PR-head build (§9b open question) — neither was reproduced against this PR's head, so declaring them without verification risked papering over label-derivation bugs rather than fixing them.

PR #2770, opened 2026-07-31
fix(python): resolve classes through module imports

### LINKED ISSUE #2746: Python: a class reached through a module alias (`models.User(db)`) resolves in no spelling

azizur100389:
## Summary

- resolve `from pkg import models` to `pkg/models.py` when the imported name is a concrete submodule
- classify only proven module targets as namespace handles and resolve qualified constructors through them
- fail closed for missing or ambiguous namespace members and when local variables or parameters shadow an imported namespace
- preserve explicit package exports over same-named submodules
- bump the incremental schema version and refresh Python scope baselines so stale indexes are rebuilt
- document namespace-edge authority and add focused integration/unit coverage

## Validation

- targeted Python resolver, import-target, scope-capture, namespace-construction, schema-version, and finalization tests
- `npx tsc --noEmit`
- `npm run build` for `gitnexus-shared` and `gitnexus`
- targeted ESLint and Prettier checks
- seven-lane production-readiness review completed; all technical findings addressed

Fixes #2746


abhigyanpatwari (commented):
# PR #2770 Review Digest — "fix(python): resolve classes through module imports" (closes #2746)

**Repo:** abhigyanpatwari/GitNexus · **Diff base:** `51095c19f8a7c641a136549711065f816a383ff6` · **Head:** `77585a485b2c18456da904fe748e2c278946fbad`

## Engine breakdown

**This is a Claude-only review, not a cross-engine tri-review.** Codex (all 3 legs — swarm-solo, ce-report-only, adversarial) hit an org spend cap on every attempt and returned zero findings (`codex-status.txt`, `codex-swarm.md`, `codex-ce.md`, `codex-adversarial.md` all show the identical `Codex error: You hit your spend cap...`). There is no cross-engine corroboration anywhere in this digest — do not read any finding below as "confirmed by two engines."

What *did* run: Claude executed 13 independent lane-agents across two methodologies —
- **GitNexus PR swarm** (6 lanes): facts-historian, branch-hygiene, risk-architect, test-ci-verifier, security-boundary, docs-dod
- **ce-code-review personas** (7 lanes): correctness, testing, maintainability, project-standards, adversarial, api-contract, performance

Each lane ran independently with a different prompt/persona and no visibility into the others' output. Where two or more of these 13 lanes converged on the same defect without being told about each other, that convergence is used below as the strongest available corroboration signal this run (in place of cross-engine agreement).

## PR summary

PR #2770 fixes issue #2746: `from pkg import models` followed by `models.User(db)` (module-alias construction) previously resolved to nothing, because the resolver bound `pkg` to `pkg/__init__.py` rather than the submodule `pkg/models.py`, and the module alias was never registered as a namespace handle. The fix adds submodule-first resolution to `resolvePythonImportTarget`, a new `isNamespaceImport` provider hook that lets Python reclassify a `named`/`alias` import edge as `namespace` post-resolution, and two new lookups in `resolveConstructionExpressionClass`/`resolveCompoundReceiverClass` that consult `namespaceTargets` for module-qualified constructor calls. 18 files changed (~300/-21), single commit, three-layer test coverage (unit parity + finalize-orchestrator unit + integration fixture).

## Bottom line

- **Branch hygiene:** clean feature/fix PR — single linear commit, zero commits behind main, no unrelated churn.
- **Merge state:** checks failing — mergeable (no conflicts), but blocked by two stale-baseline CI failures (trivially fixable, see CI status below) plus one meta-gate.
- **Verdict: not production-ready.** One confirmed P1 (local-variable shadowing silently fabricates a wrong CALLS edge, bypassing the existing shadow guard, no test coverage) is GitNexus's worst failure mode — a wrong edge with no error signal. Separately, this PR's shared-infrastructure change (`namespaceTargets` now threaded into construction resolution for every language, not just Python) opens a demonstrated regression path for TypeScript/C#/Ruby with zero non-Python test coverage. Neither issue is large to fix — a scope guard or a couple of regression tests — but both should land before merge given the silent-wrong-data failure mode. The two failing CI checks are mechanical (regenerate two committed baselines) and not themselves a blocker once the above is addressed.

---

## What's solid

- **The core Python fix is correct and well-tested for its primary scenario.** `pythonImportedSubmoduleTarget`/`isPythonImportedModule` (import-target.ts:362-405) were traced by three independent lanes (risk-architect, correctness, maintainability) against the actual resolver output and confirmed to produce the right file for the flat single-package case the PR's fixture covers. The `finalize-algorithm.ts:477` pivot from `draft.source.kind` to `draft.base.kind` was independently verified safe for every pre-existing language by risk-architect, facts-historian, AND ce-correctness (three-way convergence) — it's a strict superset of the old condition and a no-op for every provider that doesn't implement `isNamespaceImport` (i.e. everyone but Python).
- **The two new optional hooks are backward-compatible by construction.** `FinalizeHooks.isNamespaceImport?` and `ScopeResolver.isNamespaceImport?` are both optional; ce-api-contract-reviewer verified all 16 `ScopeResolver` implementations are structural object literals with no exhaustiveness checking anywhere in the codebase, so no existing language provider can break. `namespace-targets.ts`'s rewrite from parse-time `parsedImports`-scan to finalized `edge.kind` checking was independently confirmed safe (and arguably a small correctness improvement — it closes a same-`localName`-different-`kind` collision gap the old Set-based approach had) by risk-architect, facts-historian, and ce-correctness.
- **Branch hygiene is clean.** Single linear commit, zero commits behind main (merge-base == origin/main), no unrelated churn, no merge conflicts. Shared-infrastructure changes are causally coupled to the Python feature, not scope creep (branch-hygiene, project-standards).
- **The happy-path test triad is well-designed.** Unit parity tests for `isPythonImportedModule` (4 cases), a `finalize-orchestrator` unit test for the hook reclassification, and an integration test with a `decoy/models.py` fixture specifically built to catch cross-module misbinding — the negative assertion (`wrongCalls.toHaveLength(0)`) is good practice per test-ci-verifier.
- **Project-standards compliance is clean.** No language names leak into shared `gitnexus/src/core/ingestion/` code; the hook-plugin pattern is followed correctly throughout (project-standards-reviewer, zero findings above confidence 50).
- **No classic security issues.** No secrets, injection, path traversal to the OS, or hidden Unicode; the new functions are pure string comparisons with no I/O (security-boundary).

---

## Headline finding (P1)

### P1 — Local variable/parameter shadowing a namespace-import name fabricates a CALLS edge to the wrong class, bypassing the existing shadow guard
**File:** `gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts:559-567` (verified — exact match) · **Lanes:** ce-adversarial-reviewer (constructed concrete trigger, confidence 78) + gitnexus-test-ci-verifier (independently flagged same code path as a coverage gap, different angle) + ce-correctness (lower-confidence secondary note on the same mechanism) · **Tag:** `[code-read]`

**Trigger:**
```python
# pkg/models.py
class User:
    def __init__(self, db): ...
    def save(self): ...

# pkg/app.py
from pkg import models        # reclassified 'namespace' by #2746

def handler(models):          # PARAMETER shadows the import; no annotation
    return models.User(1).save()
```
**Evidence:** Verified at compound-receiver.ts:559-567:
```ts
if (objClass === undefined) {
  if (options.namespaceTargets?.has(objExpr) === true) {
    return resolveConstructionExpressionClass(fnExpr, inScope, scopes, index, options);
  }
  return undefined;
}
```
`collectNamespaceTargets` (namespace-targets.ts:38-56, verified) builds a **per-file, scope-blind** `localName → targetFile[]` map straight from finalized module-scope edges — it has no notion of which lexical scope the name is visible or shadowed in. For the call site `models.User(1)` inside `handler`, resolving the bare identifier `models` finds no `typeBinding` (unannotated parameter) and no class binding, so `objClass === undefined`. The new fallback then checks `options.namespaceTargets?.has('models')`, which is `true` at the file level regardless of the fact that `models` here is a shadowing parameter unrelated to the import — and routes through `resolveConstructionExpressionClass`, which finds the real `User` class in `pkg/models.py` and returns it. A `CALLS` edge is emitted to `pkg.models.User`'s constructor even though the runtime value bound to the parameter has nothing to do with `pkg.models`.

This is exactly the failure class the existing `strictBaseBinding` mechanism (compound-receiver.ts:118-123, cited by ce-adversarial) was built to prevent — but the structural-fold path that uses `strictBaseBinding` only vetoes locally; a failed fold falls through to the plain-text cascade, which doesn't set `strictBaseBinding`, and that's exactly the path that hits this new fallback.

**Suggested fix:** Gate the fallback on scope locality, not just name membership — only trust `namespaceTargets` for `objExpr` when there is no local/parameter binding for that name in any enclosing non-module scope, mirroring how `strictBaseBinding` already treats bound-but-untyped names.

**Testing gap:** No fixture in `python-from-module-alias` shadows `models`/`User` with a local variable or parameter; this regression would ship unnoticed.

**Corroboration chain:** ce-correctness-reviewer independently flagged this same mechanism at lower confidence (30) before ce-adversarial-reviewer constructed the concrete trigger above that promoted it to P1 — two lanes converging on the same code path from different angles (adversarial construction vs. static correctness trace).

---

## Secondary findings (P2)

### P2-1 — Submodule-first resolution can silently drop or misroute a package's own local binding when a same-named submodule file exists in the same package
**File:** `gitnexus/src/core/ingestion/languages/python/import-target.ts:51-66` (verified — exact match to lane citation) · **Lanes:** ce-correctness-reviewer (constructed 2 concrete repro shapes, confidence 65), coordinator (direct full-file read) · **Tag:** `[code-read]` · **Severity note:** ce-correctness rated this P1 (confidence 65); the coordinator's own source-read independently rated it P2, on the grounds that it requires the specific, uncommon idiom of `pkg/__init__.py` locally shadowing a same-named submodule file. This digest follows the coordinator's P2 call — the mechanism is confirmed real, but the trigger pattern is narrow enough that it doesn't rise to the same severity as P1 above, which fires on an ordinary, common Python pattern (an unannotated parameter reusing a common name like `models`).

**Trigger:** `pkg/__init__.py` defines its own local binding of a name (e.g. `session = create_session()`) that happens to collide with an unrelated same-named submodule `pkg/session.py` in the same package:
```python
# pkg/__init__.py
from .db import create_session
session = create_session()          # local name 'session' bound here

# pkg/session.py — unrelated legacy submodule sharing the name
class SessionRecord: ...

# app.py
from pkg import session
session.query(...)
```
**Evidence:** `resolvePythonImportTarget` (import-target.ts:51-66) computes `submoduleTarget = pythonImportedSubmoduleTarget(parsedImport)` and, for `kind === 'named' || 'alias'`, recursively resolves it as `{ kind: 'namespace', targetRaw: 'pkg.session' }` and **returns that submodule path unconditionally if non-null** (line 65) — this happens *before* line 69's call to `resolvePythonImportInternal(ctx.fromFile, parsedImport.targetRaw, ...)`, which is the path that would otherwise resolve `pkg` itself to `pkg/__init__.py` and let `tryFinalize`'s `findExportByName` find the real local binding. Pre-PR, `'pkg'` resolved via the single-segment proximity path directly to `pkg/__init__.py`, and the local binding was found correctly. Post-PR, if `pkg/session.py` exists as a file, the code returns that submodule unconditionally, never checking whether `pkg/__init__.py` already shadows it with a local definition. The reclassified edge then gets `kind: 'namespace'` (via `isPythonImportedModule` independently re-deriving the same match), so `tryFinalize`'s namespace branch looks for a def named `session` inside `pkg/session.py`'s own local defs — which won't exist — leaving `targetDefId` undefined and dropping the binding for `session` in `app.py`'s scope entirely.

**Suggested fix:** Don't let the submodule-first branch short-circuit resolution of the base target unconditionally. Either (a) only take the submodule interpretation when the base target's own resolution doesn't already expose a local/re-exported def with that name (requires wiring into `tryFinalize`, not the file-path-only hook), or (b) at minimum scope the submodule search to the already-resolved base package directory instead of re-running the full repo-wide `resolveAbsoluteFromFiles` fallback.

**Note on the related-but-distinct security-boundary framing:** security-boundary raised a related P2 (suffix match `endsWith('/' + moduleFile)` can cross unrelated directory trees, e.g. `vendor/pkg/models.py`) and suggested restricting the suffix fallback to absolute paths. The coordinator confirmed via direct source read (`pipeline/run.ts:595`, `new Set(parsedFiles.map(f => f.filePath))`) that `allFilePaths` — and therefore every `targetFile` — is **always workspace-relative in this codebase**, never OS-absolute. That means the "restrict to absolute paths" fix would make the fallback permanently dead code, not narrow it correctly — don't apply that specific suggestion. The underlying suffix-ambiguity concern is real but is best understood as a pre-existing ambiguity in `resolveAbsoluteFromFiles`'s own nested-root fallback, not a new defect `isPythonImportedModule` introduces (ce-adversarial independently reached the same "refuted as a *new* bug" conclusion for the suffix-match path specifically — see P3 section). This exact-match same-package collision (no suffix fallback needed) is the cleaner, higher-confidence variant of that concern.

---

### P2-2 — `namespaceTargets` threading now activates a new namespace-first resolution path for TypeScript, C#, JS, and Ruby construction expressions — not just Python — with zero non-Python test coverage
**File:** `gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts:216-225` and `:559-567`; wiring at `receiver-bound-calls.ts:260,413,942,950,1053` (all verified exact) · **Lanes:** gitnexus-risk-architect (constructed regression path) + ce-api-contract-reviewer (independently confirmed via `git show <base>` that `namespaceTargets` was computed pre-PR but never threaded into `compoundOpts` before this diff) — two independent lanes, different personas, same finding · **Tag:** `[code-read]`

**Mechanism:** `resolveConstructionExpressionClass` fires whenever `options.constructionSyntax` is defined — true for C#, JavaScript, TypeScript, Ruby, and Python (confirmed by api-contract via grep across all `scope-resolver.ts` files), not Python alone. Before this PR, `resolveCompoundReceiverClass` was called with `compoundOpts`, which never included `namespaceTargets` for any language. After this PR, all 4 call sites (verified: lines 413, 942, 950, 1053) pass `fileCompoundOpts = { ...compoundOpts, namespaceTargets }`, so the new namespace-qualified lookup block (lines 216-225) now runs for `new ns.Service()` in TypeScript/C# and `ns.Service()` in Ruby/JS wherever `ns` is a genuine namespace import — a pattern those languages already had before this PR (`import * as ns`, C# `using`).

**Regression path:** `namespaceMatches.length > 1` (two files reachable under the same namespace alias, each exporting a same-named class) returns `undefined` immediately — bypassing the previously-unconditional `qualifiedIds`/`viaSimpleName` fallback that ran before this PR and might have uniquely resolved the call. A CALLS edge that existed pre-PR could be silently dropped post-PR for these languages.

**Suggested fix:** Either gate the namespace lookup to `constructionSyntax?.bare === true` (Python's `bare: true` shape) if Python-only scope is intended, or add at least one TypeScript/C# regression test (`import * as pkg` + `new pkg.Class()`) confirming CALLS-edge counts are unchanged.

**Note:** risk-architect additionally traced the line-563 fallback per language and found it effectively Python-only in practice (TypeScript's `objExpr` for `new pkg.Service()` is `'new pkg'`, which will not match `namespaceTargets`; Ruby has neither `keyword` nor `bare` construction syntax so `calleeName` is `undefined`). The live risk is specifically the lines 216-225 block, reachable via `new ns.Class()` and Ruby/JS namespace-qualified construction.

### P2-3 — `import-target.ts:399-403` suffix-match fallback can cross unrelated directory trees (narrower framing than originally proposed)
**File:** `gitnexus/src/core/ingestion/languages/python/import-target.ts:399-403` (verified) · **Lanes:** security-boundary-reviewer (P2, confidence HIGH) + ce-adversarial (constructed repro, confidence 40, "refuted as a *new* bug" but confirms the amplification) + coordinator source-read · **Tag:** `[code-read]`

`normalizedTarget.endsWith('/' + moduleFile)` can match a file in a structurally distinct tree (e.g. `vendor/pkg/models.py` satisfying an import intended for `pkg/models.py`). ce-adversarial's direct trace clarifies this is **not a new defect** — `isPythonImportedModule` doesn't independently search; it only validates the shape of whatever `resolveAbsoluteFromFiles` already picked via its pre-existing (unchanged) tie-break logic. What *is* new: this same pre-existing ambiguity now also decides edge **kind** (`namespace` vs `named`), not just edge target, so an already-ambiguous resolution now also picks which resolution path (`namespaceTargets` vs. `findExportByName`) governs the wrong file's binding. **Do not apply the "restrict suffix fallback to absolute paths" fix suggested by the security-boundary lane** — the coordinator confirmed via `pipeline/run.ts:595` that `allFilePaths` is always workspace-relative in this codebase, so that fix would make the fallback permanently dead code rather than correctly narrowing it. If action is wanted, the correct fix is to make `resolveAbsoluteFromFiles`'s tie-break itself package-scoped, not to gate on path absoluteness.

### P2-4 — Two CI checks fail on stale committed baselines, not logic bugs
**Lanes:** gitnexus-test-ci-verifier + gitnexus-pr-facts-historian, both independently read actual CI job logs · **Tag:** `[code-read]` (log-read, not a source-code claim)

See dedicated CI Status section below.

### P2-5 — Four missing WHY-comments / stale doc references
**Lane:** gitnexus-docs-dod-reviewer (high confidence, all four independently verified present/absent by re-reading the cited lines) · **Tag:** `[code-read]`

1. `ARCHITECTURE.md` — no `isNamespaceImport?` row in the ScopeResolver hook table (verified: `grep -n isNamespaceImport ARCHITECTURE.md` returns nothing).
2. `finalize-algorithm.ts:477` — no comment explaining why `draft.base.kind` must be read instead of `draft.source.kind` (verified present as bare `if (draft.base.kind === 'namespace') {` with no preceding comment).
3. `compound-receiver.ts:216-225` — no comment explaining why the namespace lookup runs before `qualifiedIds`.
4. `namespace-targets.ts:16-19` — doc comment still says `ParsedImport.kind === 'namespace'` is "the cross-language hook"; code now reads `edge.kind === 'namespace'` (verified, code at line 47), which can diverge post-reclassification. ce-api-contract-reviewer and ce-maintainability-reviewer each independently flagged the same stale comment — a third lane converging on this exact finding.

None of these block correctness; they weaken future-maintainer readability per the repo's own DoD.

### P2-6 — Test coverage gaps: untested `alias`-kind path, untested ambiguous-collision branch, untested multi-segment positive case
**Lane:** ce-testing-reviewer, corroborated by gitnexus-test-ci-verifier and gitnexus-risk-architect on the ambiguity-branch gap specifically (3-way convergence on that sub-point) · **Tag:** `[code-read]`

- `namespaceMatches.length > 1 → undefined` (compound-receiver.ts:223-224, verified) has no fixture exercising it — flagged independently by ce-testing, test-ci-verifier, AND risk-architect (three lanes).
- `from pkg import models as m` (`kind: 'alias'`) is treated identically to `'named'` in both `pythonImportedSubmoduleTarget` (import-target.ts:363, verified) and the submodule-first branch (line 54, verified), but every new test constructs only `kind: 'named'` values — despite the PR's own fixture directory being named `python-from-module-alias`.
- The positive multi-segment submodule case (`from a.b import c` where `c` is a submodule of `a.b`, verifying the dotted-path branch at import-target.ts:394 for parent packages deeper than one segment) is untested; only the negative multi-segment control exists.

---

## Lower-priority / P3 / refuted-but-worth-noting

- **`isPythonImportedModule` duplicates path-resolution math** that `resolveAbsoluteFromFiles`/`resolvePythonImportInternal` already compute, because the `isNamespaceImport` hook signature was never given `allFilePaths` (ce-maintainability, P2→downgraded here since it's a maintainability/DRY concern, not a live bug — the two implementations were hand-verified to currently agree). Suggested fix: thread `allFilePaths` into the hook signature so `isPythonImportedModule` can delegate to `resolvePythonImportTarget` instead of reimplementing it.
- **`compoundOpts` vs `fileCompoundOpts` naming gives no compiler signal** that the bare object must never reach `resolveCompoundReceiverClass` directly (ce-maintainability, P2). A future 5th call site could silently spread the wrong object and type-check fine since `namespaceTargets` is optional.
- **`ImportEdge.kind: 'namespace'` semantics silently widened** — pre-PR it was a 1:1 mirror of syntax-level `ParsedImport.kind === 'namespace'`; post-PR it can also mean "reclassified post-resolution," changing `BindingRef.origin` provenance and debug-reason strings for consumers that pattern-match on it (ce-api-contract, P2, confidence 55, no currently-known breakage found).
- **`withDefaultHooks` inconsistent no-op defaulting style** — `isNamespaceImport` is left `undefined` rather than defaulted to `() => false` like its sibling hooks; functionally inert today (only chained call sites, no bare invocation exists) but a footgun for a future refactor that assumes all three hooks are guaranteed functions (ce-api-contract, P3).
- **Refuted by ce-adversarial (do not carry forward):** (1) relative-import ascend arithmetic — mirrors `resolvePythonImportInternal`'s boundary check exactly, no divergence found; (2) suffix-match fallback as an *independent* misclassification source — the classifier only validates an already-resolved path, it doesn't pick between candidates (see P2-2 for the narrower, confirmed framing); (4) `isNamespaceImport` fan-out inconsistency — structurally possible but currently unreachable, since Python's resolver never returns an array and no other provider implements the hook; (5) quadratic blowup from `namespaceFiles.map().filter()` — bounded by per-name collision count (realistically 1), not a real complexity concern.
- **Refuted/confirmed-clean by ce-performance-reviewer:** all three flagged call sites (per-file `collectNamespaceTargets`, the new namespace branch in `resolveConstructionExpressionClass`, and `hooks.isNamespaceImport?.()` in `makeEdgeDrafts`) are bounded, O(imports)-or-better, and in one case (`namespace-targets.ts`) the diff actually *removes* a redundant pass versus the base. No performance findings.
- **Refuted/confirmed-clean by ce-project-standards-reviewer:** no language-naming leaks into shared pipeline code; no written "no `any`" rule exists in this repo's own standards docs to cite (that's a personal preference, not project policy) and the diff introduces none anyway; no path-normalization convention violated.
- **PR title `(python)` scope tag is defensible but slightly narrow** given the shared-infrastructure blast radius described in P2-2 (docs-dod, P3).
- **Weak aggregate assertions in two integration tests** (`toHaveLength(2)`/`toHaveLength(3)` pooled across three fixture functions rather than per-function; `toBeGreaterThan(0)` instead of the file's dominant `toBe(2)` exact-count convention) — real but low-severity test-quality nits (ce-testing, P3).
- **Missing blank line / minor style nit** in `finalize-orchestrator.test.ts` (docs-dod, P3, cosmetic).

---

## CI status — 2 failing checks are stale-baseline issues, not logic bugs

Confirmed live via `gh pr checks 2770` at digest time (matches what test-ci-verifier and facts-historian independently read from CI job logs):

| Check | Status | Root cause |
|---|---|---|
| `tests / benchmarks (GITNEXUS_BENCH)` | **FAIL** | `bench/python-scope/measure.mjs --check` reports fingerprint drift: got `b57c5479...`, expected `ec879302...` (the committed `baseline-fingerprint.txt` predates the 4 new `python-from-module-alias` fixtures added to the scanned corpus). This is the benchmark guard **working correctly** — it detected a real, intentional output change and the author didn't re-baseline. |
| `tests / ubuntu / coverage 2/3` | **FAIL** | `python-captures-golden.test.ts` digest mismatch for `python-from-module-alias/pkg/app.py`: expected `86d20ee4...`, received `afb50d3e...`. Capture count (23) matches in both — only the digest differs, consistent with a golden generated from a stale/different build state (the repo's own documented vitest `dist/`-cache trap) rather than a logic error. |
| `CI Gate` | **FAIL** | Pure meta-gate — fails automatically because the two checks above failed. No independent failure. |

**Exact fix commands** (independently identified by both lanes):
```bash
cd gitnexus
UPDATE_GOLDEN=1 npx vitest run test/unit/scope-resolution/python/python-captures-golden.test.ts
node --import tsx bench/python-scope/measure.mjs
```
Both must be run in a clean build (not against a stale `dist/`) and the resulting files committed. All other checks pass, including CodeQL, typecheck, lint, format, all tree-sitter ABI gates, and all platform-sensitive shards on Windows/macOS/Ubuntu. `Vercel` fails only on external-contributor authorization, not code — expected for a fork PR, not a defect.

---

## Coverage note

This review read the full diff (18 files, ~300/-21 lines) and cross-referenced every headline and secondary finding's file:line citation against the current worktree source directly (not just lane text) before including it here — several lane-reported line numbers were confirmed exact on re-read (e.g. compound-receiver.ts:216-225, 559-567; import-target.ts:51-66, 362-405; finalize-algorithm.ts:477; namespace-targets.ts:16-19,38-56,47; receiver-bound-calls.ts:181,260,413,942,950,1053). CI status was re-verified live via `gh pr checks 2770` rather than trusting the lanes' log excerpts alone. No file or domain in the diff was skipped.

---

*Automated multi-lane digest (Claude-only this run; Codex unavailable due to spend cap). Verify before acting — this is a synthesis of static, read-only review passes, not a reproduced/executed confirmation of any finding.*

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -540,7 +556,15 @@ export function resolveCompoundReceiverClass(
       options,
       depth + 1,
     );
-    if (objClass === undefined) return undefined;
+    if (objClass === undefined) {
+      // A verified namespace-qualified bare constructor is syntactically
+      // indistinguishable from an untyped member call here. Only the namespace
+      // map makes the constructor interpretation safe.
+      if (options.namespaceTargets?.has(objExpr) === true) {
```
abhigyanpatwari, line 603: **P1 — local variable/parameter shadowing a namespace-import name fabricates a wrong CALLS edge** `[code-read]`

`namespaceTargets` (built by `collectNamespaceTargets`, namespace-targets.ts:38-56) is a per-file, scope-blind `localName → targetFile[]` map with no notion of lexical shadowing. When `objClass === undefined` for `obj.method()`, this fallback checks only `namespaceTargets?.has(objExpr)` — true at file level even if `objExpr` is actually a shadowing local variable or unannotated parameter unrelated to the import.

**Trigger:**
```python
# pkg/app.py
from pkg import models        # reclassified 'namespace' by #2746

def handler(models):          # PARAMETER shadows the import; no annotation
    return models.User(1).save()
```
`models.User(1)` inside `handler` resolves via this fallback to `pkg.models.User`'s constructor and emits a CALLS edge — even though the parameter has nothing to do with the `pkg.models` import. This is exactly the failure class the existing `strictBaseBinding` guard (lines 118-123) was built to prevent, but the plain-text cascade that reaches this fallback doesn't set `strictBaseBinding`.

**Suggested fix:** gate this fallback on scope locality — only trust `namespaceTargets` for `objExpr` when there is no local/parameter binding for that name in any enclosing non-module scope.

No fixture in `python-from-module-alias` exercises a shadowing case, so this would ship unnoticed. Confirmed via concrete construction (ce-adversarial-reviewer, confidence 78) and independently flagged as a coverage gap (gitnexus-test-ci-verifier) and at lower confidence by ce-correctness-reviewer — three lanes on the same code path.

gitnexus/src/core/ingestion/languages/python/import-target.ts
```diff
@@ -48,6 +48,23 @@ export function resolvePythonImportTarget(
   if (parsedImport.kind === 'dynamic-unresolved') return null;
   if (parsedImport.targetRaw === null || parsedImport.targetRaw === '') return null;
 
+  const submoduleTarget = pythonImportedSubmoduleTarget(parsedImport);
```
abhigyanpatwari, line 54: **P2 — submodule-first resolution can unconditionally shadow a package's own local binding** `[code-read]`

This block resolves the submodule interpretation of a `named`/`alias` import and **returns it unconditionally if non-null** (line 65) — before line 69's call to `resolvePythonImportInternal`, which is the path that would otherwise resolve the base package (e.g. `pkg` → `pkg/__init__.py`) and let `tryFinalize`'s `findExportByName` find a local binding the package's own `__init__.py` defines under that name.

**Trigger:** `pkg/__init__.py` defines its own local `session = create_session()`, and an unrelated `pkg/session.py` file also exists in the same package. Pre-PR, `from pkg import session` correctly bound to the local `session` in `__init__.py`. Post-PR, it unconditionally resolves to the submodule file instead, and since `pkg/session.py` has no top-level `session` def, the binding is silently dropped.

**Suggested fix:** only take the submodule interpretation when the base target's own resolution doesn't already expose a local/re-exported def with that name (needs wiring into `tryFinalize`, not just this file-path-only hook), or at minimum scope the submodule search to the already-resolved base package directory.

Confirmed via two concrete repro shapes (ce-correctness-reviewer, confidence 65) and independently re-derived by the coordinator reading the full file. Rated P2 rather than P1 here — the trigger requires the specific, uncommon idiom of `__init__.py` shadowing a same-named submodule file, narrower than the P1 finding above.

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -208,6 +212,18 @@ function resolveConstructionExpressionClass(
...
+
+  const namespaceName = baseName.slice(0, lastDot);
+  const exportedName = baseName.slice(lastDot + 1);
+  const namespaceFiles = options.namespaceTargets?.get(namespaceName) ?? [];
+  if (namespaceFiles.length > 0) {
+    const namespaceMatches = namespaceFiles
+      .map((targetFile) => findExportedDef(targetFile, exportedName, index))
+      .filter((def): def is SymbolDefinition => def !== undefined && isClassLike(def.type));
+    if (namespaceMatches.length === 1) return namespaceMatches[0];
+    if (namespaceMatches.length > 1) return undefined;
```
abhigyanpatwari: **P2 — this namespace-first lookup now runs for every language with `constructionSyntax`, not just Python, with zero non-Python test coverage** `[code-read]`

`resolveConstructionExpressionClass` fires whenever `options.constructionSyntax` is defined — true for C#, JavaScript, TypeScript, and Ruby, not Python alone. Before this PR, `resolveCompoundReceiverClass` was called with `compoundOpts`, which never included `namespaceTargets` for any language. After this PR, all 4 call sites in `receiver-bound-calls.ts` (lines 413, 942, 950, 1053) pass `fileCompoundOpts = { ...compoundOpts, namespaceTargets }`, so this block now also runs for `new ns.Service()` in TypeScript/C# and `ns.Service()` in Ruby wherever `ns` is a genuine pre-existing namespace import (`import * as ns`, C# `using`) — a pattern those languages already had before this PR.

**Regression path:** `namespaceMatches.length > 1` (line 224) returns `undefined` immediately, bypassing the previously-unconditional `qualifiedIds` fallback below (line 227) that ran before this PR and might have uniquely resolved the call. A CALLS edge that existed pre-PR could be silently dropped post-PR for these languages.

**Suggested fix:** gate this lookup to `constructionSyntax?.bare === true` (Python's shape) if Python-only scope is intended, or add at least one TypeScript/C# regression test (`import * as pkg` + `new pkg.Class()`) confirming CALLS-edge counts are unchanged.

Independently flagged by gitnexus-risk-architect (constructed the regression path) and ce-api-contract-reviewer (confirmed via `git show <base>` that `namespaceTargets` was computed pre-PR but never threaded into `compoundOpts` before this diff) — two lanes, different personas, same finding.

gitnexus-shared/src/scope-resolution/finalize-algorithm.ts
```diff
@@ -461,7 +474,7 @@ function tryFinalize(
   // languages emit a synthetic module-def), pick it up as the `targetDefId`
   // so consumers can reach the module as a symbol — but its absence is not
   // a failure.
-  if (draft.source.kind === 'namespace') {
+  if (draft.base.kind === 'namespace') {
```
abhigyanpatwari, line 477: **P2 (docs) — worth a one-line WHY comment here** `[code-read]`

This is the trickiest single-line change in the PR: was `draft.source.kind === 'namespace'`, now `draft.base.kind`. `draft.source` is the original immutable `ParsedImport` (stays `'named'` after reclassification); `draft.base.kind` is what `makeEdgeDrafts` may have upgraded to `'namespace'` via the new `isNamespaceImport` hook. A maintainer reading this line in isolation — without first reading `makeEdgeDrafts` — won't know why `source.kind` can't be used here, and might silently revert it in a future cleanup. Confirmed correct as written (verified safe for every existing language — this is a strict superset of the old condition), just undocumented. Suggested:
```ts
// `draft.base.kind` — not `draft.source.kind` — because `makeEdgeDrafts` may have upgraded the kind to
// 'namespace' via the `isNamespaceImport` hook after target resolution. `draft.source` is the original
// immutable ParsedImport and retains 'named'; only `draft.base` carries the finalized edge kind.
```
(gitnexus-docs-dod-reviewer; three more docs gaps — a stale doc comment in `namespace-targets.ts:16-19` and a missing `isNamespaceImport?` row in ARCHITECTURE.md's ScopeResolver hook table — are noted in the review body, not inlined since they fall outside this diff's hunks.)

azizur100389:
Addressed in the latest update:

- made qualified constructor resolution fail closed for missing or ambiguous namespace members
- added lexical-shadow barriers for locals, typed bindings, and untyped parameters
- preserved explicit package exports when a same-named submodule exists
- bumped the incremental schema version to invalidate stale indexes
- refreshed Python scope benchmark/golden baselines
- added focused namespace, shadowing, import-precedence, and reuse-gate tests
- updated architecture documentation and namespace-target comments
- completed the production-readiness review and addressed all technical findings

Validation included targeted Vitest suites, TypeScript typechecking, package builds, formatting, and lint checks.

PR #599, opened 2026-03-31
Fix/wiki script escape and llm token param

icodebuster:
These are the same changes as #371. I have created a new merge request.

## Summary

Fix XSS vulnerability in wiki HTML viewer by escaping HTML-significant characters in embedded JSON, and add automatic `max_tokens` → `max_completion_tokens` parameter fallback in the LLM client for models that reject the legacy parameter name.

## Motivation / context

1. **Script injection in wiki HTML:** The wiki viewer embeds page content as JSON inside a `<script>` tag using raw `JSON.stringify`. If any wiki page content contains `</script>`, the HTML parser closes the script block early, enabling XSS. This is especially dangerous because wiki content is LLM-generated and its exact output is unpredictable.
2. **LLM token parameter compatibility:** Newer OpenAI models (e.g. `o3-mini`, `gpt-4.1`) reject the `max_tokens` parameter with a 400 error, requiring `max_completion_tokens` instead. Users hitting this error had no recourse other than manually configuring the correct parameter. The auto-switch makes the wiki generator work out of the box with both old and new model families.

## Areas touched

- [x] `gitnexus/` (CLI / core / MCP server)
- [ ] `gitnexus-web/` (Vite / React UI)
- [ ] `.github/` (workflows, actions)
- [ ] `eval/` or other tooling
- [ ] Docs / agent config only (`AGENTS.md`, `CLAUDE.md`, `.cursor/`, `llms.txt`, etc.)

## Scope & constraints

**In scope**

- `safeJSON()` helper that Unicode-escapes `<`, `>`, `&` in serialized JSON so it is safe to embed in `<script>` tags
- Content-Security-Policy meta tag added to generated wiki HTML (`default-src 'none'`; allows inline scripts, the marked CDN, and data/https images)
- Auto-retry in `callLLM` that switches `max_tokens` to `max_completion_tokens` on a 400 mentioning `'max_tokens'`, without consuming a retry attempt
- Unit tests for both changes

**Explicitly out of scope / not done here**

- Migrating the wiki viewer away from inline scripts entirely (would require a bundler or separate JS asset)
- Server-side HTML sanitization of LLM-generated markdown (the CSP and JSON escaping are defense-in-depth at the embedding layer)
- Proactively detecting the model family to choose the correct token parameter up front (the reactive fallback is simpler and model-list-agnostic)

## Implementation notes

[...]

Issue #583, opened 2026-03-29
404 api/heartbeat causes browser reload


hacki11:
### Area

gitnexus-web (browser UI / WASM / workers)

### Summary

The web app reloads after a few seconds. In the browser console I can see 404 for api/heartbeat. While the other endpoints seem to work I guess the web app reloads after a specific timeout when no heartbeat is received. I wonder why this doesn't work on my side.

### Expected behavior

Webapp does not reload every few seconds

### Actual behavior

First it builds up a few seconds:

<img width="1918" height="585" alt="Image" src="https://github.com/user-attachments/assets/26494aa2-7875-455e-90b2-8e728d55c1c4" />

Then it reloads:

<img width="1918" height="585" alt="Image" src="https://github.com/user-attachments/assets/afd4dc9c-da2b-4052-b820-1e23c1d089db" />

### Steps to reproduce

1) npx gitnexus@latest analyze
2) npx gitnexus@latest serve
3) open gitnexus.vercel.app (Chrome & FF)

### Environment

Fedora 43

zahlekhan:
Same issue on mac and chrome

agungwa:
I had the same issue. The /api/heartbeat endpoint exists in the source code but isn't in the published npm package yet.
                                                                                                                                             
  Workaround — build from source:
                                                                                                                                             
  # Clone and build the CLI                                                                                                                  
  git clone https://github.com/abhigyanpatwari/gitnexus.git                                                                                  
  cd gitnexus/gitnexus                                                                                                                       
  npm install && npm run build                                                                                                               
  npm link                                                                                                                                   
                                                                                                                                             
  # Build shared package (required for web UI)                                                                                               
  cd ../gitnexus-shared                                                                                                                      
  npm install && npm run build                                                                                                               
                                                                                                                                             
  # Start the server (Terminal 1)                                                                                                            
  gitnexus serve                                                                                                                             
                                                                                                                                             
  # Start the web UI (Terminal 2)                                                                                                            
  cd ../gitnexus-web                                                                                                                         
  npm install && npm run dev                                                                                                                 
                                                                                                                                             
  Then open http://localhost:5173 — it should work without the reload loop.                                                                  
                                                                                                                                             
  This fixed it for me. The npm package just needs an update to include the heartbeat endpoint.

remnant-song:
windows11   edge  ， same issue

[...]

Issue #298, opened 2026-03-15
v1.4.1: C# extraction completely broken, simple_base_type query crashes entire language pipeline


fireundubh:
## Steps
 
1. Install GitNexus v1.4.1 globally (`npm install -g gitnexus@1.4.1` or `gitnexus@latest`)
2. Run `npx gitnexus analyze` on any C# repository
3. Run `npx gitnexus context <any symbol>` — returns empty incoming/outgoing/processes

## Result

v1.4.1's C# tree-sitter queries use `simple_base_type` in the heritage patterns, which doesn't exist in the bundled `tree-sitter-c-sharp` v0.21.3. Since all queries are compiled as a single string, the invalid node type causes the entire query to fail — silently, because the worker catches the error with an empty `catch {}`. No symbols, calls, or relationships are extracted for any C# file. The index contains only File/Folder/CodeEmbedding nodes.

Also, the [npm registry](https://www.npmjs.com/package/gitnexus) offers v1.4.1, but v1.4.0 is the real latest per repo tags/releases.

mihakrajnc:
Temporary fix for those coming across this is to "update" to 1.4.0, there extraction works fine:

```npm install -g gitnexus@1.4.0```

magyargergo:
Can you please updgrade to our latest release? https://github.com/abhigyanpatwari/GitNexus/releases/tag/v1.4.5 Please repoen this if it is still broken.

PR #2771, opened 2026-07-31
fix(swift): preprocess indented conditional directives so class bodies survive parsing

lehenbauer:
## Problem

In Swift files that use conditional compilation between members of a class, the
tree-sitter-swift 0.7.1 parser can lose the enclosing `class_declaration` during
error recovery. The outer class then disappears from the graph, while nested
types recovered after the directive surface as file-scope declarations. This
made a large production Swift class invisible even though several of its nested
types were present.

The smallest reproducer is:

```swift
class Outer {
  enum A { case x }
  #if os(iOS)
  enum B { case y }
  #endif
}
```

## Root cause

The generated grammar metadata in
`gitnexus/vendor/tree-sitter-swift/src/node-types.json:4981-5040` lists the
children accepted by `class_body`, including declarations and comments but not
`directive`. `class_declaration.body` then requires a `class_body` or
`enum_class_body` (`node-types.json:5044-5059`). An indented `#if` inside the
class therefore cannot remain in that body, and recovery can turn the enclosing
class into an `ERROR` subtree while recovering later members at
`source_file` scope.

## Mitigation

The Swift provider now uses its existing `preprocessSource` hook to blank only
whole indented `#if`, `#elseif`, `#else`, and `#endif` lines before parsing. The
replacement uses spaces only, preserving source length, line endings, and all
character offsets. Top-level directives are unchanged.

This hook follows the existing provider-level preprocessing pattern used by the
C++ Unreal Engine macro preprocessor and the Dart extension-type preprocessor.
Keeping the transformation at the provider boundary makes the workaround
language-specific and keeps shared ingestion code language-agnostic.

The scanner is string-aware: regular and raw multiline string literals are
tracked, as are raw-string pound counts, ordinary strings, comments, and nested
block comments. Conditional-looking lines inside multiline string data are left
byte-for-byte unchanged. Unterminated multiline strings are handled
conservatively through EOF. String interpolation is not parsed with complete
Swift expression nesting fidelity; a nested multiline string inside an
interpolation remains a documented residual of this lightweight scanner.

The tradeoff is that the parser sees declarations from every conditional branch.
That can expose mutually exclusive platform declarations together. Conditional
declarations with the same signature are still subject to the existing graph
deduplication behavior; this mitigation does not attempt to model branch
selection.

The proper long-term fix is grammar-level, and it already exists upstream of
the grammar: tree-sitter-swift merged
[PR #583](https://github.com/alex-pinkus/tree-sitter-swift/pull/583) ("Allow
#if/#elseif/#else/#endif directives inside type bodies"), with
[issue #298](https://github.com/alex-pinkus/tree-sitter-swift/issues/298) and
follow-up [PR #599](https://github.com/alex-pinkus/tree-sitter-swift/pull/599)
tracking related directive parsing. Re-vendoring a tree-sitter-swift release
that includes #583 would make this preprocessing unnecessary; until then, the
vendored 0.7.1 grammar exhibits the reproducer above and this mitigation keeps
affected classes in the graph.

## Tests

- Unit coverage verifies selective blanking, top-level preservation, length and
  newline preservation, CRLF offsets, regular/raw multiline string identity,
  unterminated strings, non-conditional hash directives, and nested comments.
- The tree-sitter integration suite covers the six-line class fixture and a
  top-level conditional-compilation control.
- The worker-pipeline integration suite verifies that `Outer`, `A`, and `B`
  remain graph symbols and that a multiline string property plus a following
  method survive preprocessing.

## Validation

```text
npx tsc --noEmit
exit 0

npm run build
exit 0

npx --no-install vitest run test/unit/swift-conditional-directive-preprocess.test.ts test/integration/tree-sitter-languages.test.ts test/integration/swift-conditional-directive.test.ts test/integration/resolvers/swift.test.ts test/integration/swift-scope-capture-tripwire.test.ts test/unit/scope-resolution/swift/*.test.ts
Test Files  8 passed (8)
     Tests  187 passed (187)
```


gitnexus/src/core/ingestion/languages/swift/conditional-directive-preprocess.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  while (lineStart < sourceText.length) {
+    const newlineIndex = sourceText.indexOf('\n', lineStart);
+    const lineEnd = newlineIndex === -1 ? sourceText.length : newlineIndex;
+    const line = sourceText.slice(lineStart, lineEnd);
+    const lineWithoutCarriageReturn = line.endsWith('\r') ? line.slice(0, -1) : line;
+    const insideMultilineString = state.multilineStringPounds !== null;
+
+    output +=
+      !insideMultilineString &&
```
magyargergo: **[P1] Blanking a directive line that carries a block-comment terminator turns valid Swift into a parse error.**

This blank/keep decision consults only `insideMultilineString`; it never consults `state.blockCommentDepth`, which this file already tracks. So a line like `  #endif */` — the result of commenting out a conditional block, which is ordinary developer behavior — matches the regex at line 1 and gets erased *including the `*/`*.

Reproduced against the real vendored grammar:

```swift
class Foo {
  /* temporarily disabled:
  #if DEBUG
  func f() {}
  #endif */
  func g() {}
}
```

`hasError raw=false → preprocessed=true`. The comment is left unterminated and swallows `func g()` and the rest of the file. This PR converts previously-parseable Swift into a parse error.

It also contradicts this file's own doc comment at lines 124-127: *"Directive-looking lines inside block comments are still blanked: the Swift parser ignores comment interiors, so this remains parse-harmless."* It is not parse-harmless. A second, softer consequence of the same gap: `/** … */` doc comments lose blanked lines from the extracted `description`, because Swift reads block-doc text from the preprocessed tree (`BASE: "… Example: #if os(iOS) widget.useUIKit() #endif"` → `PR: "… Example: widget.useUIKit()"`).

**Fix:** add the depth check — `!insideMultilineString && state.blockCommentDepth === 0 && SWIFT_INDENTED_CONDITIONAL_DIRECTIVE_RE.test(...)`. Nothing needs blanking inside a comment; the grammar already ignores it. Add a regression with the closing delimiter on the directive-looking line.

`[reproduced]` — cross-engine (Codex ce + Claude adversarial/correctness) and independently reproduced by the coordinator.

gitnexus/src/core/ingestion/languages/swift.ts
```diff
@@ -178,6 +179,7 @@ const BUILT_INS: ReadonlySet<string> = new Set([
 export const swiftProvider = defineLanguage({
   id: SupportedLanguages.Swift,
   extensions: ['.swift'],
+  preprocessSource: preprocessSwiftConditionalDirectives,
```
magyargergo, line 182: **[P1] This is a parse-semantics change with no `SCHEMA_BUMP`, so warm parse caches replay pre-fix Swift captures and the fix is inert.**

(Anchored here because the defect is an *absence* in `src/storage/parse-cache.ts`, which this PR does not touch.)

- `parse-cache.ts:149` — `const SCHEMA_BUMP = 36;` on **both** base and head.
- `parse-cache.ts:173` — `PARSE_CACHE_VERSION = \`${SCHEMA_BUMP}+${GITNEXUS_PKG_VERSION}\``; `package.json` is `1.6.9` on both sides, so the version string is byte-identical `"36+1.6.9"` across this PR.
- The chunk key hashes **raw on-disk bytes**, and `preprocessSource` runs inside the worker at `parse-worker.ts:1471-1472` — after the key is computed — so the transform is invisible to the cache.

Trigger: analyze a Swift repo on a pre-fix build, update a **same-package-version** checkout to this PR, re-run. Unchanged bytes ⇒ chunk hash matches ⇒ cache version matches ⇒ the old `ParseWorkerResult` (class already error-recovered away) is replayed verbatim, with no warning. `--force` does not clear it — the cache is content-addressed and survives it.

Scope this honestly: an actual npm release bumps `package.json`, and `parse-cache.ts:36-40` says the pkgVersion component exists precisely so `npm i -g gitnexus@latest` invalidates. The population actually affected is same-version consumers — git checkout, the project-local runner, dev machines, CI — plus this repo's own benchmark and golden runs. The repo contract is unambiguous either way: ledger entries v22…v36 all bump for exactly this class of change, several noting results are "replayed verbatim — including across `--force`".

**Fix:** bump `SCHEMA_BUMP` with a ledger entry naming this PR, and update the pinned cache-version test. Rebase onto `main` first and re-check for a concurrent bump rather than hardcoding 37. `INCREMENTAL_SCHEMA_VERSION` was separately checked by two lanes and does **not** need a bump — no new node types, no id-shape change.

`[code-read]` — cross-engine: all three Codex legs and two Claude lanes; version constants verified by the coordinator on both sides.

gitnexus/src/core/ingestion/languages/swift/conditional-directive-preprocess.ts
```diff
@@ -0,0 +1,163 @@
...
+  if (line[index] !== '#') return null;
+
+  let poundCount = 0;
+  while (line[index + poundCount] === '#') poundCount++;
+  return hasTripleQuoteAt(line, index + poundCount) ? poundCount : null;
+}
+
+function matchingMultilineCloseLength(line: string, index: number, poundCount: number): number {
+  if (!hasTripleQuoteAt(line, index)) return 0;
+  if (poundCount === 0) return line[index + 3] === '#' ? 0 : 3;
```
magyargergo: **[P1] Two lexical desyncs wedge the scanner inside multiline-string state, silently making the fix inert for the rest of the file.**

Measured on a fixed template (class + multiline string + a real indented `#if`/`#endif` in the same body), asking only whether the *real* directives got blanked:

```
control : plain string, plain close            directivesBlanked=true
(a)     : escaped \""" inside the string        directivesBlanked=false
(b)     : close """ immediately followed by #   directivesBlanked=false
control2: close """ then SPACE then #           directivesBlanked=true
```

**(a)** The multiline-string branch at lines 63-72 has no escape handling, unlike `skipRegularString` at line 39. Swift's documented `\"""` is read as the closing delimiter; the scanner then re-enters string state at the *real* `"""` and never leaves. It also blanks directive-looking lines *inside* the string literal, rewriting the program's string data — the exact invariant the new unit test at `:64-101` asserts must hold.

**(b)** This line — `if (poundCount === 0) return line[index + 3] === '#' ? 0 : 3;` — refuses to close a **plain** `"""` string whenever a `#` immediately follows. `control2` (a space before the `#`) blanks correctly, which isolates the adjacency as the cause. The `poundCount > 0` guard at line 25 is correct; generalizing it down to N=0 is not — a bare `"""` always closes a plain multiline string. The same over-strict guard also refuses `"""##` as a terminator for a `#"""` string.

Both fail silently: length is preserved, nothing errors, the fix just does nothing for that file and the original class-body-drop bug returns.

**Fix:** in the `multilineStringPounds !== null` branch, skip backslash escapes when the count is 0 (mirroring `skipRegularString`); and drop the `#`-adjacency guards here — `poundCount === 0` ⇒ `return 3`, and for `poundCount > 0` accept a terminator with at least `poundCount` pounds. Consider a fail-safe: if the scan ends still inside a string or comment, return `sourceText` unchanged rather than trusting a scan that went off the rails.

`[reproduced]` — cross-engine (Codex swarm RA-2 / ce #3 / adversarial, Claude correctness F2 and maintainability F1) and independently reproduced by the coordinator.

gitnexus/src/core/ingestion/languages/swift/conditional-directive-preprocess.ts
```diff
@@ -0,0 +1,163 @@
...
+
+function hasTripleQuoteAt(line: string, index: number): boolean {
+  return line.startsWith('"""', index);
+}
+
+function rawPoundCountBeforeTripleQuote(line: string, index: number): number | null {
+  if (line[index] !== '#') return null;
+
+  let poundCount = 0;
+  while (line[index + poundCount] === '#') poundCount++;
```
magyargergo: **[P1] Quadratic scan on a run of bare `#`: one 512 KB line stalls the parse worker for minutes.**

This loop walks the entire remaining `#` run at every index; `scanSwiftLine` at lines 93-99 walks the same run **again** inline; and when it isn't a raw-string opener, `index` advances by exactly one at line 105. For a run of length n that is Σ(n, n-1, …, 1) character comparisons.

Measured against the PR-head module:

```
n=4000   ms=39.9
n=8000   ms=153.0    (3.84×)
n=16000  ms=586.3    (3.83×)
n=32000  ms=2715.1   (4.63×)
n=64000  ms=10648.1  (3.92×)   <- 10.6 s for one 64 KB line
```

~4× per doubling is the textbook quadratic signature. The same total `#` count spread over many short lines takes 9.8 ms, and ordinary Swift-ish text is linear — so this is an isolated algorithmic bug, not a general property of the scanner, and it does not contradict the (correct) finding that general input is linear. It needs a long run of consecutive `#` on **one line in code position**: generated, minified, corrupt, or deliberately hostile `.swift` content, which GitNexus ingests by design.

The walker's default skip threshold is 512 KB (`src/core/ingestion/utils/max-file-size.ts:5`), so a **within-limits** file lands around ten minutes of pure preprocessing — a synchronous JS loop before any N-API call, so the worker can't post progress. That runs straight into the pool's idle-timeout → split → retry → quarantine ladder, ending in the file being silently excluded. Output length stays correct throughout; this is a stall, not a corruption.

**Fix:**

```ts
if (line[index] === '#') {
  let rawPoundCount = 0;
  while (line[index + rawPoundCount] === '#') rawPoundCount++;
  if (line[index + rawPoundCount] === '"') {
    index = skipRegularString(line, index, rawPoundCount);
  } else {
    index += rawPoundCount; // was: fell through to index++
  }
  continue;
}
```

and share the counted value with `rawPoundCountBeforeTripleQuote` so the run is measured once per position instead of twice. Per repo convention, any regression test should assert on a bounded operation count or a `Promise.race` timeout, not a wall-clock threshold.

`[reproduced]` — cross-engine (Codex swarm SB-1 and adversarial, Claude performance F1) and independently reproduced by the coordinator.

gitnexus/src/core/ingestion/languages/swift/conditional-directive-preprocess.ts
```diff
@@ -0,0 +1,163 @@
+const SWIFT_INDENTED_CONDITIONAL_DIRECTIVE_RE = /^[ \t]+#(?:if|elseif|else|endif)\b[^\r\n]*$/;
```
magyargergo: **[P2] The predicate keys on indentation, but the grammar's acceptance keys on nesting — so the fix misses the common case and over-fires on the documented one.**

`^[ \t]+` requires leading whitespace, and the doc comment at lines 119-120 justifies it with *"Top-level directives are left intact because they are valid source-file members in the grammar."* Column 0 and top level are not the same thing, in either direction.

**Miss:** a `#if` at column 0 *inside* a class body is left untouched and the class is still discarded. Measured over 7 body shapes (enum member, nested struct, property, func, typealias, nested class, empty body), 6 lose the enclosing declaration:

```
COL0    enum member   blanked=false  hasError=true   classCaptured=false
INDENT  enum member   blanked=true   hasError=false  classCaptured=true
```

An A/B pipeline run confirms base and PR are byte-identical for that fixture — the fix is a no-op there. This is common formatting (SwiftFormat `--ifdef outdent`, Xcode re-indent), so it is a large share of the real-world population.

**Over-fire:** an *indented* directive at file scope **is** blanked, e.g. a valid `  #if DEBUG` at top level — contrary to the sentence above and to the contract the new integration test claims to pin.

Same-root variants that widen the miss: NBSP or ideographic-space indentation, a UTF-8 BOM before a first-line directive, and CR-only line endings (line 144 splits on `\n` only, so such a file is treated as one line and nothing matches).

**Fix:** decide on nesting depth, not indentation. `scanSwiftLine` already walks every character and already skips strings and comments, so it can maintain a `braceDepth` for free; blank any `#if|#elseif|#else|#endif` line when `braceDepth > 0` at line start, regardless of leading whitespace. That preserves today's verified top-level behavior and subsumes the whitespace variants. Failing that, at minimum correct the doc comment to say "unindented" rather than "top-level", and add the column-0-inside-a-body matrix as a regression table.

`[reproduced]` — cross-engine, but note the split: Claude (correctness F1, adversarial F1, testing F3) reproduced the **miss** and Codex (swarm DDD-1) independently reported the **over-fire**. Neither engine found both halves; they are labelled cross-engine because they are two symptoms of one root cause, not because both engines confirmed the same symptom.

gitnexus/src/core/ingestion/languages/swift/conditional-directive-preprocess.ts
```diff
@@ -0,0 +1,163 @@
...
+ * Swift parser ignores comment interiors, so this remains parse-harmless and
+ * preserves the preprocessor's existing behavior. String interpolation is not
+ * parsed with full Swift expression fidelity; a nested multiline string inside
+ * an interpolation can therefore still confuse this light scanner.
+ *
+ * An unterminated multiline string conservatively keeps all subsequent lines
+ * untouched through EOF. Every line is scanned independently while the small
+ * lexical state carries only across line boundaries.
+ */
+export function preprocessSwiftConditionalDirectives(sourceText: string): string {
```
magyargergo, line 217: **[P2] Only the parse worker applies this hook; the other Swift parse consumers still see raw text, so the pipeline's halves now disagree.**

(Anchored on the exported entry point; the defects are in files this PR does not touch.)

Two instances:

1. **Embeddings** — `src/core/embeddings/ast-utils.ts:33` calls `parseSourceSafe(parserInstance, content)` on raw text. A live probe on a class this PR repairs returned `hasError=true` with empty `methodNames` and `fieldNames`, and sufficiently large declarations fall back from AST-aware chunking to character chunking. Semantic search degrades for exactly the Swift files this PR fixes.
2. **Scope capture** — `scope-resolution/pipeline/run.ts:495` passes raw `file.content` to `extractParsedFile`, and `languages/swift/captures.ts:94` re-parses it verbatim when no cached tree is supplied. **Dart already closes this exact hole inside its own emitter** — `languages/dart/captures.ts:116`: `const parseText = preprocessDartExtensionTypes(sourceText);` — so the convention exists and Swift is the outlier.

Reproduced for (2): on the same fixture, the raw path yields **no `Outer` class scope at all** and hoists its members to Module scope, while the preprocessed path nests them correctly. Forcing the documented `preExtractedByPath`-miss fallback turns `CALLS Class:Host.swift:Host → Class:Engine` into `CALLS File:Host.swift → Class:Engine`. Before this PR both paths parsed raw text and agreed (both wrong); the PR makes them disagree, turning a consistent degradation into cold-run/warm-run non-determinism.

Related: `language-provider.ts:117-118` claims the hook is *"Called once per file, on every code path that re-parses"*, but there is exactly one call site (`parse-worker.ts:1472`). That doc is stale for every provider — this PR is just the first change where it costs something.

**Fix:** mirror Dart in `emitSwiftScopeCaptures` (parse `preprocessSwiftConditionalDirectives(sourceText)` on the cache-miss path — safe and idempotent, since the transform is length-preserving), and apply `provider.preprocessSource` in `ensureAndParse` before the embedding parse. A generic parity assertion — `extractParsedFile(provider, raw)` must equal `extractParsedFile(provider, provider.preprocessSource(raw))` for every provider defining the hook — would catch this class for Dart and C++ too.

`[reproduced]` + `[code-read]` — (1) cross-engine (Codex swarm RA-3 / ce #4 / adversarial, Claude risk-architect); (2) single-engine, two Claude lanes, reproduced at both ParsedFile and graph-edge level.

magyargergo (commented):
# Review — PR #2771: preprocess indented Swift conditional directives

## How this was reviewed

Two engines, not five independent opinions — read the corroboration labels below with that in mind.

- **Engine A — Claude.** The GitNexus production-readiness swarm (pr-facts/branch-hygiene, risk-architect, security-boundary, test-CI) plus the `ce-code-review` personas (correctness, testing, maintainability, performance, project-standards, adversarial). All lanes produced output; two swarm lanes died mid-run and were re-dispatched successfully.
- **Engine B — Codex, `gpt-5.6-sol` at xhigh effort.** The same swarm scaffolding in solo mode, the same `ce-code-review` persona set, plus its native adversarial pass. **All three legs were live and produced real findings.** Codex's swarm leg could not run `npx vitest`/`npx tsx` locally (bubblewrap namespace rejection), so its dynamic evidence came from the ce leg, which ran the unit suite (8/8) and direct parser probes.
- **Coordinator.** Independently re-ran the load-bearing claims against the PR-head module and the real vendored `tree-sitter-swift` prebuild. Where a lane and the coordinator disagreed, the coordinator's measurement won.

Corroboration classes used below: **cross-engine** (a Claude lane and a Codex leg reached it independently), **single-engine** (consistent within one engine — which can be a shared false positive), **coordinator-reproduced** (executed against the PR head).

---

## What's solid

**The fix works for the shape it targets.** Measured directly: for

```swift
class Ok {
  #if DEBUG
  func m() {}
  #endif
}
```

`hasError` goes `true` → `false`, and an A/B run of two builds differing only in whether the preprocessor is the identity function shows the PR *gaining* `NODE Class:Ind.swift:IndOuter`, `qn="IndOuter.IndA"` and `HAS_METHOD Class:IndOuter → Function:…indMethod#0` — none of which exist on base. That is a real, verifiable graph improvement, and the diagnosis (tree-sitter-swift 0.7.1 doesn't admit `directive` as a `class_body` child; upstream #583 fixes it) is correct.

The design choices are right too: `preprocessSource` is the sanctioned hook, the file sits under `languages/swift/` with zero language names leaking into shared ingestion code, the direct import (not via `swift/index.js`) and the dropped `filePath` parameter both match the existing Dart precedent, and no `any`/`as any` was introduced. Single clean conventional commit, no unrelated churn, `git diff --check` clean, all three new files pure ASCII, no bidi controls.

**Attacks that were tried and failed.** Recording these so nobody re-spends the time:

- **ReDoS on `/^[ \t]+#(?:if|elseif|else|endif)\b[^\r\n]*$/`** — refuted. Disjoint prefix classes, no nested quantifier, no ambiguous alternation. 1,000,000 spaces → 1.2 ms; 1M spaces + `a` → 1.1 ms; 1M tabs then `#if` → 1.4 ms.
- **General memory / string-concat DoS** — refuted. 10 MB single line → 110 ms; 500k short lines → 107 ms; 100k → 200k lines scales at 2.25×; peak memory ~2× file size. V8 rope strings make `output +=` effectively linear. Independently: 81–207 MiB/s across realistic, all-multiline-string, and quote-churn shapes.
- **Stray `/*` as a manipulation vector** — refuted. Injecting an unclosed `/*` produced byte-identical output to the same file without it.
- **JS length / newline-position / CR-position / astral / NUL preservation** — refuted as a concern across 200,000 randomized inputs built from 26 adversarial token shapes: `lenFails=0 newlinePosFails=0 crPosFails=0`, function is pure and idempotent. Surrogate pairs become two spaces (no lone surrogate), and node-tree-sitter reports `startIndex` in UTF-16 code units, so downstream JS slicing stays aligned.
- **Duplicate-declaration node-id collisions throwing** — refuted. `graph.ts:77-84` is silent first-wins; no throw, no duplicate id.
- **Duplicate *scope*-id throw** (the failure Dart guards against at `dart/captures.ts:133-136`) — refuted. Swift scope ids are range-derived, so two `#if`/`#else` branches never collide. Verified for 2-branch and 3-branch fixtures.
- **The #2769/#2740 undeclared FROM/TO CONTAINS-pair abort** — refuted from both sides. The Swift query set only produces already-declared pairs (`schema.ts:301-319`), newly-surviving nested types still attach to `File`, and #2769 on current `main` declares the relevant pairs anyway. Analyze completed cleanly on every fixture.
- **`#ifdef` / `#ifx` / `#if_something` / `#warning` / `#available` / `#selector` word-boundary bypass** — refuted; `\b` behaves.
- **"Materially slows every analyze run"** — refuted. ~0.16 ms for a realistic 11 KB file; ~2.5 s aggregate across 20,000 files.
- **Goldens / baselines needing a rebaseline** — refuted. No existing Swift fixture uses conditional directives, so the transform is a no-op on all of them.
- **`INCREMENTAL_SCHEMA_VERSION`** — two lanes independently concluded it is *not* the knob this change needs (no new node types, no id-shape change). `SCHEMA_BUMP` is.

---

## Headline findings

Seven issues are worth blocking on or fixing in-PR. Five are P1. Six are summarised here; the seventh
(brace-asymmetric branches re-parenting the file) is written up under "Lower priority, unresolved, and
test gaps" below because it surfaced last, but it is inlined as a P1 like the rest.

**1. Blanking a directive line that carries a block-comment terminator turns valid Swift into a parse error.** This is the one that converts working input into broken input, and it is triggered by ordinary developer behavior — commenting out a conditional block:

```swift
class Foo {
  /* temporarily disabled:
  #if DEBUG
  func f() {}
  #endif */
  func g() {}
}
```

`  #endif */` matches the directive regex, so the whole line is erased including the `*/`. The block comment is then unterminated and swallows `func g()` and the rest of the file: `hasError raw=false → preprocessed=true`. The blank/keep decision at :150-154 consults only `insideMultilineString`, never `blockCommentDepth` — which the file already tracks. This also directly contradicts the module's own doc comment at :124-127 ("Directive-looking lines inside block comments are still blanked … so this remains parse-harmless"). It is not parse-harmless. A softer instance of the same root cause: `/** … */` doc comments lose their example lines, because Swift stores block-doc text as the node's `description` and reads it from the *preprocessed* tree (`BASE: "… Example: #if os(iOS) widget.useUIKit() #endif"` → `PR: "… Example: widget.useUIKit()"`).

**2. No `SCHEMA_BUMP`, so warm parse caches replay pre-fix Swift captures.** `parse-cache.ts:149` is `SCHEMA_BUMP = 36` on base *and* head, `package.json` is `1.6.9` on both, so `PARSE_CACHE_VERSION` is byte-identical `"36+1.6.9"` across the PR. The chunk key hashes raw on-disk bytes, and `preprocessSource` runs inside the worker at `parse-worker.ts:1471-1472` — after the key is computed — so it is invisible to the cache. Unchanged Swift bytes ⇒ cache hit ⇒ the old `ParseWorkerResult` with the class already error-recovered away is replayed verbatim, and `--force` does not clear it. Be honest about the radius: an actual npm release bumps `package.json`, and `parse-cache.ts:36-40` says the pkgVersion component exists precisely so `npm i -g gitnexus@latest` invalidates. So the population actually hit is **same-package-version consumers** — git checkout, project-local runner, dev machines, CI — plus this repo's own benchmark and golden runs. The repo-contract violation stands regardless: the ledger entries v22…v36 all bump for exactly this class of change, several noting "replayed verbatim — including across `--force`".

**3. Two lexical desyncs wedge the scanner inside string state and silently make the fix inert.** Both measured on a fixed template (class + multiline string + real indented `#if`/`#endif` in the same body), asking only whether the *real* directives got blanked:

```
control : plain string, plain close           directivesBlanked=true
(a)     : escaped \""" inside the string       directivesBlanked=false
(b)     : close """ immediately followed by #  directivesBlanked=false
control2: close """ then SPACE then #          directivesBlanked=true
```

(a) The multiline-string branch at :63-72 has no escape handling, unlike `skipRegularString` at :39 — Swift's documented `\"""` is read as the closing delimiter, the scanner re-enters string state at the real `"""` and never leaves. As a bonus it blanks directive-looking lines *inside* the string literal, rewriting program data. (b) Line 22's `if (poundCount === 0) return line[index + 3] === '#' ? 0 : 3;` refuses to close a **plain** `"""` string whenever a `#` follows the delimiter; `control2` isolates the adjacency as the cause. In both cases nothing errors, length is preserved, and the original class-body-drop bug just comes back for that file.

**4. Quadratic scan on a run of bare `#`.** `rawPoundCountBeforeTripleQuote` (:12-18) walks the whole remaining run at every index, `scanSwiftLine` (:93-99) walks it again inline, and then `index` advances by exactly one at :105. Measured on the PR-head module:

```
n=4000   ms=39.9      n=8000  ms=153.0   (3.84×)
n=16000  ms=586.3     (3.83×)
n=32000  ms=2715.1    (4.63×)
n=64000  ms=10648.1   (3.92×)   <- 10.6 SECONDS for one 64 KB line
```

~4× per doubling is the textbook signature; the same total `#` count spread over many short lines runs in 9.8 ms, and ordinary Swift-ish text is linear. Note this does **not** contradict the refuted general-DoS result above — the two are compatible, this needs a long run of consecutive `#` on **one line in code position** (generated, minified, corrupt, or hostile `.swift`). The walker's default skip threshold is 512 KB (`utils/max-file-size.ts:5`), so a within-limits file lands around ten minutes of pure preprocessing inside the parse worker, well past the pool's idle-timeout/retry/quarantine ladder. Output length stays correct throughout — this is a stall, not a corruption. The fix is two lines.

**5. The blank/keep predicate keys on indentation; the grammar keys on nesting.** Both engines hit this from opposite sides. A column-0 `#if` *inside* a class body is not blanked, so the class is still discarded — measured over 7 body shapes, 6 lose the enclosing `class Outer` entirely (`hasError=true`, `classCaptured=false`), and an A/B pipeline run confirms base and PR produce identical output for that fixture. Meanwhile an *indented* directive at file scope **is** blanked, which contradicts the doc comment at :119-120 ("Top-level directives are left intact because they are valid source-file members in the grammar"). Column-0-inside-a-body is common formatting (SwiftFormat `--ifdef outdent`, Xcode re-indent). A brace-depth counter in `scanSwiftLine` — which already walks every character and already skips strings and comments — removes the proxy entirely and subsumes the Unicode-space and BOM-indentation variants as well.

**6. Only the worker path preprocesses; the other Swift parse consumers now disagree with it.** Two instances. Embeddings: `src/core/embeddings/ast-utils.ts:33` calls `parseSourceSafe(parserInstance, content)` on raw text — a live probe on a class this PR repairs returned `hasError=true` with empty `methodNames`/`fieldNames`. Scope capture: `scope-resolution/pipeline/run.ts:495` passes raw `file.content` to `extractParsedFile`, and `swift/captures.ts:94` re-parses it verbatim on a cache miss. **Dart closes exactly this hole inside its own emitter** (`dart/captures.ts:116`: `const parseText = preprocessDartExtensionTypes(sourceText);`); Swift doesn't follow the precedent. Reproduced: the raw path produces no `Outer` class scope at all and hoists its members to Module scope, while the preprocessed path nests them correctly; forcing the documented fallback turns `CALLS Class:Host.swift:Host → Class:Engine` into `CALLS File:Host.swift → Class:Engine`. Before this PR both paths parsed raw text and agreed (both wrong) — the PR makes them disagree, converting a consistent degradation into cold-run/warm-run non-determinism.

---

| # | Location | Sev | Corroboration | Trigger | Tag |
|---|---|---|---|---|---|
| 1 | `conditional-directive-preprocess.ts:151` | P1 | cross-engine + coordinator-reproduced | `  #endif */` — commenting out a conditional block | `[reproduced]` |
| 2 | `swift.ts:182` (defect is the absence in `parse-cache.ts:149`) | P1 | cross-engine (3/3 Codex legs, 2 Claude lanes) + coordinator-verified constants | analyze a Swift repo pre-fix, update a same-version checkout, re-run (incl. `--force`) | `[code-read]` |
| 3 | `conditional-directive-preprocess.ts:22` (also `:63-72`) | P1 | cross-engine + coordinator-reproduced | `\"""` in a multiline string; or a close `"""` immediately followed by `#` | `[reproduced]` |
| 4 | `conditional-directive-preprocess.ts:16` (also `:95`, `:105`) | P1 | cross-engine + coordinator-reproduced | one line holding a long run of consecutive `#` in code position | `[reproduced]` |
| 5 | `conditional-directive-preprocess.ts:1` | P2 | cross-engine (both halves of the same root cause) | column-0 `#if` inside a class body; indented `#if` at file scope | `[reproduced]` |
| 6 | `conditional-directive-preprocess.ts:135` (real sites: `embeddings/ast-utils.ts:33`, `swift/captures.ts:94`) | P2 | (a) cross-engine, (b) single-engine (2 Claude lanes) | `--embeddings` on a repaired class; any `preExtractedByPath` miss | `[reproduced]` + `[code-read]` |
| 7 | `conditional-directive-preprocess.ts:153` | P1 | single-engine + coordinator-reproduced | `#if`/`#else` branching a declaration *header*, leaving unbalanced braces | `[reproduced]` |

Findings 2 and 6 live in files this PR does not touch, so their inline anchors sit on the nearest in-diff line and name the real location in the comment text.

---

## Lower priority, unresolved, and test gaps

**Brace-asymmetric `#if`/`#else` branches re-parent the rest of the file (P1, coordinator-reproduced).** Swift legitimately splits one declaration header across branches:

```swift
final class NetworkClient {
  #if swift(>=5.5)
  func fetch() async throws -> Int {
  #else
  func fetch() throws -> Int {
  #endif
    return 1
  }
}
```

Blanking both directive lines leaves two openers and one closer. In an A/B pipeline run, four unrelated top-level types acquired fabricated qualified names (`SessionStore` → `qn="NetworkClient.SessionStore"`, same for `TokenCache`, `Config`, `Environment`) and `NetworkClient`'s span grew from one line to the whole file. Base degraded *locally*; the PR degrades *file-globally*, and `qualifiedName` is what resolution, `context` and `rename` key on. One Claude lane did run the same shape and reported "definition set identical before and after" — that check covered def *names* only and never looked at `qualifiedName` or spans, so it does not contradict this; it just didn't measure it. No Codex leg exercised the shape.

The coordinator re-ran it independently against the real vendored grammar and confirms it, which is why it is inlined as a P1 rather than left as a lead. Top-level node lists for the fixture above plus `struct SessionStore {}` and `enum Unrelated { case a }`:

```
RAW          top-level: class_declaration, directive, function_declaration,
                        class_declaration, class_declaration
PREPROCESSED top-level: class_declaration
```

Five top-level nodes collapse to one. Be precise about what this is: both sides have `hasError=true` (the raw form is the bug this PR targets), so it is not "valid → invalid" — it is **partially-recovered → structurally wrong**, with unrelated top-level types silently reparented. A cheap guard is to skip blanking any `#if…#endif` group whose branches have unequal net `{`/`}` counts, degrading to today's behavior rather than emitting an unbalanced file.

**Duplicate declarations across `#if`/`#else` are new, undecided, and untested (P2, cross-engine).** Both branches now survive. Same signature → identical node id → `graph.ts:77-84` silently keeps the first and drops the second, and edges from *both* bodies attach to that one node. Differing arity or parameter shape → both survive as phantom overloads that can never coexist in a real build, and `swiftArityCompatibility` will dispatch to either. The scope model keeps both bindings regardless (two `accent` defs in one class scope, resolved by source order). This is strictly better than losing the whole class, but nothing in the diff acknowledges, documents, or pins it. Decide the intended semantics and add a test that makes a future change to it a visible diff.

**UTF-8 byte length is not preserved (P3, coordinator-reproduced, impact unresolved).** `language-provider.ts:110-119` singles this out: implementations targeting languages whose elided ranges may contain non-ASCII "must therefore preserve byte length, not just `.length`". `blankDirectiveLine` replaces per UTF-16 code unit, so `  #if os(iOS) // 日本語 🔥` measures `jsLen 59→59` but `utf8 67→59`. C++ earns its exemption because UE macro tokens are ASCII-only; Swift directive lines carry arbitrary trailing comment text, so that exemption doesn't transfer. No consumer that slices the *original* UTF-8 bytes by `startIndex` was found — node-tree-sitter reports UTF-16 indices, so the in-process graph path is fine — so today this is a contract violation with unproven impact. At minimum, document why the shrink is acceptable here, the way the C++ preprocessor does.

**Missing early-out (P3).** Both sibling preprocessors return the original string with zero work when nothing matches (`stripUeMacros`: `if (!HAS_UE_HINT.test(source)) return source;`). Swift has none, so a directive-free file is still fully re-sliced and re-scanned: 20.8 ms vs C++ 0.19 ms and Dart 0.16 ms on the same ~2 MB no-match input, and 12 MB/call retained heap vs 0. A single `/^[ \t]+#(?:if|elseif|else|endif)\b/m` test on the whole file is provably sufficient and measured 10–11× faster.

**Missing removal trigger (P3).** The doc comment names grammar version 0.7.1 but not upstream tree-sitter-swift #583 / #298 / #599, nor a condition for deleting this file. `swift` has no `"hold"` in `.github/vendored-grammars.json`, so the weekly bump bot will bump it the moment a release containing #583 ships — and nothing in the source will tell that reviewer this file is now dead weight. The repo's `ponytail:` convention exists for exactly this.

**Test gaps.** Ranked by what they'd actually catch:

- No column-0-inside-a-class-body case. `tree-sitter-languages.test.ts:496-513` covers indented-in-class and `:515-536` covers column-0-at-top-level; the failing quadrant is the one that isn't tested, which is why finding 5 ships silently.
- No brace-asymmetric branch case, no same-named-declaration-in-both-branches case, no escape (`\"""`) inside a multiline string, no plain `"""` close followed by a same-line character, no mismatched raw-string pound counts, no unterminated block comment, no doc-comment case (`doc-comment-description-e2e.test.ts` is the natural home).
- Every one of findings 3, 5 and the raw-string variants leaves the state machine wedged through EOF and passes today, because no test asserts "a later, unrelated directive is still blanked."
- No parity assertion between the worker path and `emitScopeCaptures`. A generic one — `extractParsedFile(provider, raw)` must equal `extractParsedFile(provider, provider.preprocessSource(raw))` for every provider defining the hook — would have caught finding 6 and would cover Dart and C++ too.
- The two new cases in `tree-sitter-languages.test.ts` call `await loadLanguage(SupportedLanguages.Swift)` with no try/catch, unlike the sibling test directly above them and unlike this PR's own `swift-conditional-directive.test.ts`, which uses `describe.skipIf`. That file is in the platform-sensitive Windows/macOS set — exactly where an optional-grammar load failure is most likely — so these will hard-fail the run instead of skipping.
- `swift-conditional-directive.test.ts:26-29` derives its expectation from the *actual* output's length (`''.padEnd(lines[N]!.length, ' ')`). It still fails on partial blanking, so it is not vacuous — but it cannot detect a *wrong-length* blanked line, which is the one invariant this preprocessor exists to hold. The same file already does it correctly elsewhere (`''.padEnd(sourceLines[line]!.length, ' ')`); make it uniform. Mechanical, P3.
- The `:515-536` top-level test passes identically with and without the change (verified) — top-level directives never needed fixing. Its only load-bearing assertion is `expect(parseContent).toBe(content)`. Fine as a no-op guard, just not the fix-proving test its placement suggests.
- `swift-conditional-directive.test.ts:50` and `:75` put `if` branches inside the Vitest bodies. Flagged by both Codex legs against the review rule; a Claude lane checked the repo's own standards files and confirmed the rule isn't written down there, so treat this as maintainer preference, not a documented violation. Collect nodes unconditionally and `.filter().map()` outside the callback.
- `workerThresholdsForTest: { minFiles: 1, minBytes: 1 }` (`:46`, `:71`) is not a real `PipelineOptions` field and is silently ignored — `tsc -p tsconfig.test.json` reports TS2561 at both sites. Copied from three pre-existing tests that share the same dead option, so not attributable to this PR; the real worker path is still forced via `workerPoolSize: 1`. Noted so it doesn't spread further.

**A regex literal containing `/*` wedges the comment tracker (P3, single-lane, `[code-read]`, not independently reproduced).** Swift's extended regex-literal syntax means `#/a/*b/#` contains the characters `/*`, which `scanSwiftLine` counts as opening a block comment; `blockCommentDepth` then never returns to 0 for the rest of the file, so string tracking is blind from that point on. This is distinct from — and not covered by — the refuted "stray `/*` as a manipulation vector" result above, which only showed that comment depth cannot change the *blanking* decision. Worth a fixture either way; a regex-literal case belongs in the same test batch as the string cases.

**Unsettled.** Whether any newly-surviving nested declaration trips the documented `determineScope` class-body/block hazard was not resolved within budget. The scope extractor does receive the preprocessed text and tree (no double-parse of raw bytes on the worker path), so this is a "more declarations now reach a known-hazardous path" question, not a known defect.

---

## CI status

**There is zero CI evidence for this PR.** Every workflow for head `e9ada96d8` is `completed / action_required` — Gitleaks, PR Autofix, Dependency Review, Docker Build & Push, PR Description Check, CI, CodeQL. The only workflow that actually executed is the `pull_request_target` Conventional Labeler. That is the standard fork-PR approval gate, not a failure, but the consequence is that this PR could be merged with nothing having run: no typecheck, no unit tests, no integration tests, no security scan. Vercel reports a failure on authorization, which is unrelated infrastructure.

Locally, the unit suite passes 8/8. The two integration tests fail in the review worktree for an environmental reason only — `runPipelineFromRepo` needs a built `dist/core/ingestion/workers/parse-worker.js` and the worktree has no build — **not** a PR defect. But note the consequence: the only end-to-end test of this fix runs through a build artifact, and CI never ran, so **the fix has not been executed end-to-end anywhere yet**.

The branch is behind `main` by ~7 commits including #2769. No commits landed on `main` in any file this PR touches (`languages/swift*`, `parse-cache.ts`, `parse-worker.ts`, `language-provider.ts`), so no textual conflict is expected — but rebase before the final validation run, and re-check `SCHEMA_BUMP` for a concurrent bump at that point rather than hardcoding 37.

---

## Final verdict

**not production-ready**

- **Branch hygiene: `clean feature/fix PR`** — one conventional commit, cohesive scope, no unrelated churn, no merge commits, `git diff --check` clean.
- **Merge state: `checks failing`** — every protected workflow is `action_required`; nothing has actually run against this head.

The diagnosis is right, the hook is the right layer, and the fix demonstrably works for the shape it targets. What blocks it is that the transform is unconditional: it fires inside block comments (IC1), inside wedged string state (IC3), across brace-asymmetric branches (IC7), and on a quadratic path (IC4) — while missing the column-0 case entirely (IC5) and leaving warm caches to replay pre-fix results (IC2). Adding a `braceDepth` counter to `scanSwiftLine` — which already walks every character and already skips strings and comments — plus a `blockCommentDepth === 0` guard and the `SCHEMA_BUMP` addresses most of this without changing the approach.

---

*This digest was assembled automatically from two independent review engines (Claude and Codex `gpt-5.6-sol`) plus coordinator-run reproductions. Verify each finding against the code before acting on it.*

gitnexus/src/core/ingestion/languages/swift/conditional-directive-preprocess.ts
```diff
@@ -0,0 +1,163 @@
...
+    const newlineIndex = sourceText.indexOf('\n', lineStart);
+    const lineEnd = newlineIndex === -1 ? sourceText.length : newlineIndex;
+    const line = sourceText.slice(lineStart, lineEnd);
+    const lineWithoutCarriageReturn = line.endsWith('\r') ? line.slice(0, -1) : line;
+    const insideMultilineString = state.multilineStringPounds !== null;
+
+    output +=
+      !insideMultilineString &&
+      SWIFT_INDENTED_CONDITIONAL_DIRECTIVE_RE.test(lineWithoutCarriageReturn)
+        ? blankDirectiveLine(line)
```
magyargergo: **[P1] Blanking is unconditional, so an `#if`/`#else` that branches a declaration *header* leaves unbalanced braces and re-parents the rest of the file.**

Swift legitimately splits one declaration header across branches — availability and cross-platform code does this routinely:

```swift
class NetworkClient {
  #if swift(>=5.5)
  func fetch() async {
  #else
  func fetch() {
  #endif
    perform()
  }
}

struct SessionStore {}
enum Unrelated { case a }
```

Both branch bodies open a brace; only one closes. Blanking both markers leaves two `{` and one `}`, so `NetworkClient` never closes. Measured against the real vendored grammar — top-level node lists:

```
RAW          top-level: class_declaration, directive, function_declaration,
                        class_declaration, class_declaration
PREPROCESSED top-level: class_declaration
```

Five top-level declarations collapse to one; `SessionStore` and `Unrelated` stop being top level and get reparented under `NetworkClient`. A separate A/B pipeline run measured the graph consequence directly: `qn="SessionStore"` → `qn="NetworkClient.SessionStore"` (same for `TokenCache`, `Config`, `Environment`), and `NetworkClient`'s span grew from one line to the whole file. `qualifiedName` is what resolution, `context` and `rename` key on.

Be precise about the severity: both sides have `hasError=true` here, because the raw form is the very bug this PR targets. So this is not "valid → invalid" — it is **partially-recovered → structurally wrong**, with unrelated types silently reparented. Today's failure is local to the affected declaration; after this change it is file-global, and nothing detects it.

**Fix:** don't blank a directive group whose branches have unequal net `{`/`}` counts — degrade to today's behavior for that group rather than emitting an unbalanced file. A `braceDepth` counter in `scanSwiftLine` (see the nesting-vs-indentation comment on line 1) gives you the raw material for both checks at once. Add this shape as a regression, asserting `qualifiedName` and spans, not just the set of definition names — the existing check that reported "definition set identical" missed this precisely because it compared names only.

`[reproduced]` — single-engine (Claude ce-adversarial F2, via an A/B of two compiled builds differing only in whether the preprocessor is the identity function) and independently reproduced by the coordinator against the vendored grammar. No Codex leg exercised this shape, so it has no cross-engine confirmation — it is inlined on the strength of two independent reproductions.

azizur100389:
Reviewed current head `e9ada96d` merged locally onto current `upstream/main` `99291891`. The branch merges cleanly locally, and validation passed: `npx tsc --noEmit`, the full build, 8 focused unit tests, and 53 focused Swift integration tests.

The passing tests do not cover three merge-blocking cases that remain in the transform:

1. Directive-looking lines inside block comments are blanked without preserving a same-line `*/`. I reproduced this with an indented `#endif */`: preprocessing removes the terminator, turning previously valid Swift into an unterminated comment and hiding later declarations.
2. The blanking decision uses indentation rather than lexical nesting. I reproduced a column-zero `#if` inside a class body; it is not blanked, so the original class-body parse failure remains. Conversely, an indented top-level directive is blanked even though it is valid there.
3. `SCHEMA_BUMP` remains 36. The parse-cache key hashes raw source before `preprocessSource` runs, so same-version warm caches can replay pre-fix Swift results indefinitely, including through `--force`. This parse-semantic change needs the next free cache version (currently 37 after rebasing/rechecking main).

Please also cover the block-comment terminator and column-zero nested cases in tests, and make the blanking decision use scanner-derived nesting/comment state rather than indentation alone. GitHub has no functional CI result for this head; only the label check ran, while Vercel is an unrelated authorization failure.

Recommendation: request changes. The targeted fix works for its existing fixtures, but these untested cases can either corrupt valid source or leave the original bug unfixed.

magyargergo:
I'm going to take care of this. Thank you for your submission :)

abhigyanpatwari:
Pushed `9e05bf82` to this branch (thanks for `maintainerCanModify`) — it addresses every finding from review 4834691204 plus the lower-priority items. Branch is also merged up to current `main`.

| # | Finding | What changed |
|---|---|---|
| 1 P1 | `  #endif */` un-terminates a block comment | The blank/keep decision now consults `blockCommentDepth` (and multiline-string state), so nothing inside a comment or literal is ever blanked. Regression: `keeps a block-comment terminator that shares its line with a directive` (unit) + `keeps source that comments out a conditional block parseable` (integration, real grammar) + a Swift case in `doc-comment-description-e2e.test.ts` pinning the `description` text. |
| 2 P1 | no `SCHEMA_BUMP` | Re-checked against `main` at merge time as you suggested — `main` had already taken 37 for #2783, so this is **38**, with a ledger entry and the pinned test updated. |
| 3 P1 | `\"""` and `"""#` string desyncs | The multiline branch skips backslash escapes when the pound count is 0, and the `#`-adjacency guards are gone: `poundCount === 0` always closes at `"""`, `poundCount > 0` accepts a terminator with *at least* that many pounds. Both cases have a regression that asserts a *later* real directive is still blanked, which is what silently failed before. |
| 4 P1 | quadratic run of bare `#` | The pound run is counted once per position and `index` advances past it; `rawPoundCountBeforeTripleQuote` is gone (it walked the same run twice). Regression uses a 64 000-`#` line under a 2 s per-test timeout rather than a measured elapsed-time assertion — pre-fix that line took 10.6 s. |
| 5 P2 | indentation vs nesting | `scanSwiftLine` now maintains `braceDepth`, and a group is blanked when it opens at `braceDepth > 0`, at any indentation. Column-0-inside-a-body is fixed; an indented file-scope directive is left intact (matching the doc comment, which was previously wrong). The whitespace variants come along for free — the predicate uses `[^\S\r\n]*`, so NBSP / ideographic-space indentation and a leading BOM all match, and line splitting now handles bare-CR endings. |
| 6 P2 | only the worker preprocesses | `emitSwiftScopeCaptures` re-applies the transform on the parse-cache-miss path (mirroring `dart/captures.ts`), `emitCppScopeCaptures` does the same for `stripUeMacros` — it had the identical hole — and `ensureAndParse` applies `provider.preprocessSource` for the embedding path. New `test/unit/preprocess-source-parity.test.ts` asserts `extractParsedFile(provider, raw)` equals `extractParsedFile(provider, preprocessSource(raw))` for every provider defining the hook (swift, cpp, dart) plus an embedding-path parse check; I verified each leg fails when its fix is reverted. The stale `LanguageProvider.preprocessSource` doc ("called on every code path that re-parses") is corrected to name the real call sites. |
| 7 P1 | brace-asymmetric branches re-parent the file | Blanking is decided per `#if`…`#endif` **group** (all-or-nothing, never half-erased), and a group is only blanked when **every branch is brace-balanced**. Note the guard has to be "each branch nets 0", not "branches are equal": in your `func fetch() async {` / `func fetch() {` example both branches are `+1`, so an equality check would still have let it through. Regressions assert the exact top-level node list (5 nodes) and that `SessionStore` keeps `qualifiedName = "SessionStore"`. |

Lower-priority items also in:

- **Missing early-out** — `HAS_CONDITIONAL_DIRECTIVE_HINT` short-circuits directive-free files, same shape as `stripUeMacros`'s `HAS_UE_HINT`.
- **Missing removal trigger** — the module doc now carries a `ponytail:` note naming upstream tree-sitter-swift #583 (plus #298/#599) as the delete condition and the bump-bot consequence.
- **Byte length ≠ `.length`** — documented explicitly with the measured numbers and why C++'s ASCII-only exemption doesn't transfer, and pinned by a test (UTF-16 length preserved, UTF-8 shrinks). No consumer slicing original bytes by `startIndex` was found, so this stays a documented exemption rather than a behavior change.
- **Regex-literal `/*`** — `#/…/#` is now skipped as a literal, so it can't open a phantom block comment.
- **Duplicate declarations across branches** — semantics left as-is (both branches survive; identical ids are first-wins in `graph.ts`) but now pinned by a test asserting the exact definition list, so changing it shows up as a diff.
- **Test-hygiene items** — `loadLanguage` calls are under a `describe.skipIf(!isLanguageAvailable(Swift))` instead of bare `await`; blank-line widths derive from the *source* line, not the output; `workerThresholdsForTest` (not a real `PipelineOptions` field on that import) removed; `if` branches lifted out of the Vitest bodies. New assertions are exact `toEqual` on full arrays rather than containment checks.

Validation on the pushed head: `tsc --noEmit` clean; `test/unit` 10 969 passed (the residual failures are the pre-existing `skip-git-cli` CLI-e2e flake — it fails *worse* on the untouched base here — plus load-induced 30 s timeouts that pass when run alone); Swift/C++/Dart resolver integration 509 passed; `tree-sitter-languages` Swift 8 passed; the four `swift-conditional-directive` pipeline tests pass against a real built worker.

CI still hasn't run on this PR — every workflow is `action_required` on the fork — so it's worth approving a run before merge.

magyargergo (approved):

PR #2773, opened 2026-07-31
fix(mcp): resolve false FTS-missing warnings in the query tool

### LINKED ISSUE #2767: MCP query tool reports "FTS indexes missing" while CLI query works on the same index
## Summary

The `query` tool behaves inconsistently between the **CLI** and the **MCP server** against the **same index**:
- CLI `gitnexus query "<terms>"` returns results normally and prints **no FTS warning**.
- MCP `mcp__gitnexus__query` returns an **empty result set** and emits:

  > `FTS indexes missing — keyword search degraded. Run: gitnexus analyze --repair-fts (or gitnexus analyze --force) to rebuild indexes.`

This makes the MCP `query` tool effectively unusable (always returns empty), even though the underlying index is healthy.

## Environment

- GitNexus: `1.6.9`
- Node: `v22.22.3`
- OS: `win32 x64` (Windows 11)
- LadybugDB native: `lbugjs.node loaded`
- MCP transport: stdio
- Client: ZCode

## What I've verified

| Check | Result |
| --- | --- |
| `gitnexus doctor` → `Full-text search: available` | ✅ Platform reports FTS available |
| `gitnexus analyze --force --pdg -v` logs | ✅ All FTS tables created (`FTS: ready Function.function_fts`, etc.) |
| `gitnexus analyze --repair-fts` | ✅ Reports `FTS indexes repaired successfully` |
| CLI `gitnexus query "login authentication"` | ✅ Returns results, no warning |
| MCP `mcp__gitnexus__query` (same query) | ❌ Empty results + `FTS indexes missing` warning |
| Full restart of the MCP client (reloads the index) | ❌ Warning persists |

## Reproduction

1. Build a fresh index and repair FTS:
   ```bash
   gitnexus analyze --force --pdg
   gitnexus analyze --repair-fts

magyargergo:
Fixes abhigyanpatwari/GitNexus#2767

## Summary

The MCP `query` tool could report `FTS indexes missing — keyword search degraded` and return empty keyword results against an index the CLI `gitnexus query` searched successfully at the same moment (#2767) — with no way to tell whether the cause was a stale connection, a resolved-repo mismatch, or a genuine build failure. The generic message now carries the resolved repo/branch/indexed-at and, when every FTS table failed for a real (non-benign) reason, the redacted underlying error — turning an unhelpful "it's broken" into an actionable diagnostic. A long-lived MCP session also now reliably notices when `gitnexus analyze --repair-fts` fixes FTS without needing a server restart.

## What changed

- **Diagnostic warning.** `ftsDegradedWarning()` gains an optional context (repo name, branch, indexed-at, and a redacted last-error string) appended to the existing message — additive only, so the healthy-path response stays byte-identical.
- **Real errors stop hiding behind "missing index."** A `QUERY_FTS_INDEX` failure is now classified as a genuinely-missing index vs. a real query/connection error (anchored to the LadybugDB exception class, not a bare substring match — confirmed against two real message shapes from a live engine run). Real errors are always logged server-side, and now also surface a client-facing partial-result warning when some FTS tables succeeded and others didn't.
- **`--repair-fts` gets a dedicated freshness signal.** It intentionally never restamps `indexedAt` (no new graph was generated), so a warm MCP session had only an incidental file-identity delta to notice a repair. Repair now also stamps the existing (previously forensic-only) `capabilities.fts.status` meta field, and the connection pool's staleness check compares it as a third, independent reinit trigger — seeded at cold init too, so a fresh process's first warm check doesn't false-trigger.
- **New end-to-end test.** A real writable LadybugDB session builds an index without FTS, a real `LocalBackend` observes the degraded state through the real pool, a separate session performs the repair, and the same still-warm backend re-queries successfully with no restart — closing the one gap no existing test covered.

## Why this shape

`gitnexus query` (CLI) and `mcp__gitnexus__query` (MCP) both route through the same `LocalBackend.query()` — there's no separate CLI-only implementation. The divergence is process lifetime: the CLI cold-opens a fresh connection per invocation and always sees current state; a long-lived MCP server reuses a pooled connection and only reopens it when its staleness check says to. Rather than assume a single root cause for #2767 (the reporter's "restart didn't help" isn't fully explained by the staleness theory alone — noted as an open assumption), this PR makes the warning self-diagnosing and closes the one gap confirmed by reading the code, so whichever cause is real becomes visible in the next report.

## Testing

- 91 tests across every changed file (unit + the new integration test) pass; `tsc --noEmit` clean.
- A related 15-file integration sweep (FTS, connection pool, query, staleness — 130 tests) passes.
- Reviewed with a 10-persona automated code review; 5 findings were fixed directly in this PR (see below), 4 lower-priority ones are recorded as residual work.

## Residual Review Findings

- P2 · `gitnexus/src/core/lbug/lbug-adapter.ts:3125` / `gitnexus/src/core/search/bm25-index.ts:60` — `queryFTS`'s inline benign-error check and the new `classifyFtsQueryError` are two independently-maintained classifiers for the identical `QUERY_FTS_INDEX` cypher call; currently masked by the CLI path's blanket catch. Fix: extract one shared classifier.
- P2 · `gitnexus/src/mcp/local/local-backend.ts:817` — `LocalBackend` now carries three parallel per-poolKey `Map`s (`lastObservedIndexedAt`/`DbIdentity`/`FtsStatus`) touched in lockstep at 5 call sites. Consolidation candidate, but touches pre-existing code beyond this fix's scope.
- P2 · `gitnexus/src/core/search/fts-indexes.ts:15` — `redactPaths` stops matching at the first whitespace, so a path containing a space (e.g. a Windows username) is only partially redacted. Pre-existing gap, now more exposed since this PR is the first to pipe raw DB error text through it.
- P2 · `gitnexus/src/mcp/local/local-backend.ts` (`ensureInitialized`) — No unit test isolates the new `ftsCapsChanged` reinit trigger from the pre-existing `identityChanged` trigger (only proven via the integration test, which is confounded since the real FTS-index write also flips `identityChanged`). Flagged independently by 3 reviewers.

- **Log queries:** search MCP server logs for the `query:fts-search` context (a real per-table FTS error, now logged instead of silently swallowed) and `capability stamp write failed` (a `--repair-fts` meta write that didn't land).
- **Healthy signal:** a degraded `query` response's `warning` text includes `(resolved: <repo>, indexed <timestamp>)` instead of the old bare `FTS indexes missing` string.
- **Failure signal:** repeated `capability stamp write failed` log lines against an otherwise-healthy repo would indicate a `gitnexus.json` permission/disk issue.
- **Rollback:** revert the merge — every change is additive/backward-compatible (the healthy-path response shape is unchanged), so no data migration or special rollback step is needed.

---

**CI status:** all required checks pass (`CI Gate` green). The `Vercel` preview-deploy check is expected to stay red for a fork PR — it requires maintainer authorization to run untrusted fork code against deploy credentials, not a code fix.


gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -2326,7 +2359,27 @@ export class LocalBackend {
     // path, leaving the success-path response shape byte-identical.
     const warnings: string[] = [];
     if (!ftsUsed) {
-      warnings.push(ftsDegradedWarning());
+      // #2767: attach what THIS session resolved (repo/branch/indexed-at) so a
+      // CLI/MCP mismatch is visible in the warning itself rather than requiring
+      // a separate debugging round-trip. Built from the already-resolved
+      // `repo` handle — no extra I/O on this per-request path.
+      warnings.push(
+        ftsDegradedWarning({
```
magyargergo: **P1 — corroborated by 3 independent lanes across both engines (Codex adversarial, Codex ce-code-review, Claude risk-architect)** `[code-read]`

When every configured FTS table query fails for a real, non-benign reason (timeout, connection reset, native fault) — not a genuinely missing index — `ftsUsed` is `false` and this branch unconditionally calls `ftsDegradedWarning({..., lastErrorRedacted: ftsQueryErrors?.[0]})`. Since the FTS extension itself is still loaded, `ftsDegradedWarning` falls through to its generic tail: **"FTS indexes missing — keyword search degraded. Run: gitnexus analyze --repair-fts..."** — the real error only appears as a trailing `; last error: ...` suffix, never the headline.

`--repair-fts` will not fix a connection reset or timeout. A user who trusts the headline is sent into the same unhelpful recovery loop #2767 itself was about, just for a different underlying trigger (real query error vs. stale connection) — the one case this PR doesn't close.

**Suggested fix:** split this path on whether `ftsQueryErrors` is present — emit a dedicated "FTS query failed" message with the redacted error as the headline, and reserve the missing-index/`--repair-fts` guidance for the genuinely benign case (no `ftsQueryErrors`, or extension-not-loaded).

gitnexus/src/core/run-analyze.ts
```diff
@@ -1031,6 +1031,47 @@ async function runFullAnalysisInner(
...
+      try {
+        // Re-read the on-disk meta immediately before writing, rather than
+        // reusing `existingMeta` (captured before the FTS rebuild ran, which
+        // can span real wall-clock time). Another writer to this same
+        // gitnexus.json in the interim — e.g. the HTTP server's background
+        // embedding-checkpoint job — must not have its update silently
+        // reverted by this stamp overwriting a stale snapshot. Falls back to
+        // `existingMeta` only if the file became unreadable in that window.
+        const latestMeta = (await loadMeta(metaDir)) ?? existingMeta;
+        await saveMeta(metaDir, {
```
magyargergo, line 1053: **P1/P2 — single lane found, coordinator-verified from source** `[code-read]`

This comment (correctly) names the HTTP server's background embedding-checkpoint job as a writer this stamp must not silently revert — but the guard only works in one direction.

Traced `gitnexus/src/server/api.ts:1806-1900`: the embedding job reads `meta.json` **once** at job start (`let embeddingMeta = await loadMeta(...)`) and can run up to 30 minutes (`EMBED_TIMEOUT_MS`). Every periodic checkpoint save (`saveEmbeddingCheckpoint`, and the job's final save) spreads onto that same **stale** in-memory object and calls `saveMeta(entry.storagePath, embeddingMeta)`. `saveMeta`/`writeMetaFile` (`gitnexus/src/storage/repo-manager.ts:829-867`) is a full-object overwrite (atomic tmp-file+rename, but not a merge/CAS) — so if `--repair-fts` stamps `capabilities.fts.status` while an embedding job is mid-flight on the same repo, the job's next checkpoint save silently reverts that stamp back to whatever was captured at job start, for up to the job's remaining lifetime.

Blast radius today is limited — `capabilities.fts.status` currently has exactly one reader (`ensureInitialized`'s `ftsCapsChanged`, a pool-reinit trigger, not a query-availability gate) — but this PR's own doc comment on `RepoMeta.capabilities` notes the field "gained its first programmatic reader," signaling more readers are expected to layer onto it.

**Suggested fix:** have `saveEmbeddingCheckpoint` re-`loadMeta` immediately before each `saveMeta` (mirroring the pattern just added here), instead of carrying one in-memory snapshot for the job's full lifetime.

magyargergo (commented):
# Tri-Review Digest — PR #2773 "fix(mcp): resolve false FTS-missing warnings in the query tool"

## Bottom line

**Verdict: changes requested before merge.** One P1 (NEW-1) leaves the PR's own fix incomplete: it closes
the stale-warm-connection case #2767 reported, but opens an equivalent misleading-guidance path for a
different trigger (a real, all-table FTS query error) — three lanes across both engines (Codex native
adversarial, Codex ce-code-review, Claude risk-architect) converged independently on the same file:line and
the same fix shape, so this isn't a stretch call. Everything else in the PR — the core classifier logic, the
warm-session reinit signal design, the meta-write safety, the security posture, and the test suite — was
reviewed and held up; the four items the author already self-disclosed remain open but don't block merge on
their own. Recommended path: land the NEW-1 fix (split the `!ftsUsed` path on whether `ftsQueryErrors` is
present), then merge.

- **Branch hygiene:** merge-from-main commit present (`e7c7bd5`, pulls in unrelated #2770) but harmless and
  merge-safe — 10/10 file match against the PR's own diff, no unrelated churn.
- **Merge state:** mergeable (no conflicts); currently review-blocked (zero human reviews on record, some CI
  shards still finishing post-merge at observation time — see §5).

## 1. Overview

PR #2773 (magyargergo, fork → abhigyanpatwari/GitNexus main) fixes #2767: the MCP `query` tool could
report "FTS indexes missing — keyword search degraded" and return empty keyword results against an
index the CLI searched successfully at the same moment. The fix has three parts: (a) `classifyFtsQueryError`
distinguishes benign "index not built yet" from real query errors on the MCP pool path and threads a new
`nonBenignErrors` field through `FTSSearchResponse`; (b) the degraded warning gains an optional
`FtsWarningContext` suffix (resolved repo/branch/indexed-at, and — when every table failed non-benignly —
the first redacted error) so a stale/mismatched session is diagnosable from the warning text itself; (c) a
new `capabilities.fts.status` reinit trigger (`lastObservedFtsStatus`/`ftsCapsChanged`) lets a long-lived
warm MCP session notice an out-of-band `gitnexus analyze --repair-fts` without a restart. 5 source files,
5 test files, +875/-32, 7 commits (one is a merge from `main`, pulling in unrelated #2770 — no diff-line
impact, confirmed 10/10 file match). The PR body self-discloses 4 residual findings from the author's own
prior internal review pass (commit `c414925`, "fix(review): address code-review findings on the #2767 fix").

**Engine participation — both fully participated.** Claude ran 5 `ce-code-review` personas
(adversarial/correctness/maintainability/standards/testing) and 5 `gitnexus-pr-swarm-review` personas
(branch-hygiene, docs-dod, facts-historian, risk-architect, security-boundary); no standalone Claude
swarm "test-ci-verifier" artifact was produced this run (see §7). Codex ran all three assigned legs —
`codex-swarm.md` (native `gitnexus-pr-review`, 6 personas via GitNexus MCP tools, verdict "production-ready
with minor follow-ups"), `codex-ce.md` (9-persona `ce-code-review` swarm with a validator gate, verdict
"Ready with fixes"), and `codex-adversarial.md` (native adversarial pass, verdict "needs-attention") — all
three returned real, substantive, file:line-grounded findings. Codex was not unavailable or degraded in
this run.

---

## 2. New findings (not in the PR's self-disclosed list)

### NEW-1 — P1 — "FTS indexes missing — run --repair-fts" is shown for genuine, all-table FTS *query* failures, not just missing indexes
**Corroboration: STRONG — 3 lanes, both engines** (Codex `codex-ce.md` api-contract+adversarial personas,
confidence 100; Codex `codex-adversarial.md` native pass, "medium"; Claude `claude-swarm-risk-architect.md` F3)
**File:line:** `gitnexus/src/mcp/local/local-backend.ts:2360-2381` (the `if (!ftsUsed)` branch), warning
builder in `gitnexus/src/core/search/fts-indexes.ts:49-77` (`ftsDegradedWarning`)
**[code-read] verified directly.**

**Trigger:** every configured FTS table query fails for a real, non-benign reason (timeout, connection
reset, native fault) — `ftsUsed` is `false` even though the indexes exist and are fine. This branch
unconditionally calls `ftsDegradedWarning({ repoName, branch, indexedAt, lastErrorRedacted: ftsQueryErrors?.[0] })`.
Inside `ftsDegradedWarning` (`fts-indexes.ts:49-77`), since the extension *is* loaded (`fts.loaded === true`
— nothing about a query failure marks the extension unavailable), execution falls through to the generic
tail: `'FTS indexes missing — keyword search degraded. Run: gitnexus analyze --repair-fts...'`, with the
real error appended only as a trailing `; last error: ${lastErrorRedacted}` suffix on the *context* string —
never the headline. Confirmed by direct read of both call sites; the existing test
(`query-degraded-signal.test.ts:428-436`) asserts both strings appear together but never asserts the
repair-fts guidance is absent/qualified for this case.

**Why it matters:** `--repair-fts` will not fix a connection reset or timeout — a user who trusts the
headline is sent into exactly the unhelpful recovery loop #2767 itself was about, just for a different
underlying cause (real query error vs. stale connection). Both Codex legs and the Claude risk-architect
lane converged independently on this same file:line and the same suggested fix.

**Suggested fix (both engines converge on the same shape):** split the `!ftsUsed` path — when
`ftsQueryErrors` is present, emit a dedicated "FTS query failed" message with the redacted error as the
headline, and reserve `ftsDegradedWarning()`'s missing-index/repair-fts guidance for the genuinely benign
case (no `ftsQueryErrors`, or extension-not-loaded).

**Inline-candidate: YES.**

---

### NEW-2 — P1/P2 — The HTTP server's embedding-checkpoint job holds a stale in-memory meta snapshot and can silently revert the new `capabilities.fts` stamp for the job's full 30-minute lifetime
**Corroboration: WEAK by lane count (Claude `claude-ce-adversarial.md` Finding 1 only, confidence 70) — but
the coordinator independently verified the mechanism from source (`api.ts:1806-1900`, `repo-manager.ts:829-867`);
the concrete, code-traced trigger justifies including it despite single-lane origin.**
**File:line:** `gitnexus/src/server/api.ts:1806,1828-1845,1899-1900` (embedding checkpoint job) vs.
`gitnexus/src/core/run-analyze.ts:1044-1067` (this PR's new stamp block) and
`gitnexus/src/storage/repo-manager.ts:829-867` (`saveMeta`/`writeMetaFile`)
**[code-read] independently re-verified by this synthesis agent, not just relayed from the lane.**

**Trigger:**
1. A user starts an HTTP embedding job for repo R. `api.ts:1806`: `let embeddingMeta = await loadMeta(entry.storagePath);` — read **once**, at job start.
2. While the job is still running (it can run up to `EMBED_TIMEOUT_MS = 30 * 60 * 1000`, `api.ts:1787`), the user (or an automation hook) runs `gitnexus analyze --repair-fts` on the same repo from a separate process — nothing coordinates the two; `acquireIndexLock` is only called from the CLI analyze path.
3. `--repair-fts` succeeds and stamps `capabilities.fts.status = 'available'` via `run-analyze.ts`'s new re-read-immediately-before-write block (`run-analyze.ts:1050-1052`) — this write lands correctly (test-verified for this direction only: `run-analyze-fts-repair.test.ts:442-500`).
4. The embed job's next checkpoint callback fires (`saveEmbeddingCheckpoint`, `api.ts:1828-1845`): `embeddingMeta = { ...embeddingMeta, embeddingCheckpoint: {...} }` — spread onto the **stale** object captured in step 1 — then `await saveMeta(entry.storagePath, embeddingMeta)`. I confirmed `saveMeta` (`repo-manager.ts:859-867`) → `writeMetaFile` (`repo-manager.ts:829-839`) is a full-object `JSON.stringify(meta, ...)` overwrite via tmp-file+rename — atomic per call, but a blind last-writer-wins overwrite, not a merge or CAS. This silently reverts `capabilities.fts.status` back to whatever was captured at job start.
5. This persists until the job ends — its **final** save (`api.ts:1899-1900`) also reuses the same stale `embeddingMeta` object — or a fresh `analyze`/`--repair-fts` run restamps it.

**Blast radius (honest, per the lane's own framing):** `capabilities.fts.status` today has exactly one
programmatic reader (`LocalBackend.ensureInitialized`'s `ftsCapsChanged`, a pool-reinit trigger only — not
a query-availability gate; actual FTS availability is decided by the live query attempt). So the immediate
symptom is a persisted `gitnexus.json` misreporting FTS as unavailable when it's actually fixed
(forensic/`gitnexus doctor` inaccuracy, some redundant cheap `initLbug` no-ops) rather than a resurrected
false "FTS indexes missing" warning through *this* path. But `repo-manager.ts`'s own comment says this
field "gained its first programmatic reader in #2767," signaling more readers are expected to layer onto
it — this bug would propagate silently into whatever reads it next. The mechanism (a long-lived in-memory
meta snapshot that never re-reads before a checkpoint save) is a general pattern risk, not unique to this
one job — any future long-running background writer to `gitnexus.json` shares the same property.

**Suggested fix:** have `saveEmbeddingCheckpoint` re-`loadMeta` immediately before each `saveMeta`
(mirroring the pattern this PR just added in `run-analyze.ts`) instead of carrying one in-memory snapshot
for the job's full lifetime; and/or scope `capabilities` merges to field-level patches rather than
full-object overwrite from a stale base.

**Inline-candidate: YES** (P1/P2, single-lane-found but reproducible from a concrete, code-verified trigger).

---

### NEW-3 — P2 — Degraded-warning context can report a stale `indexedAt` from a cached repo handle
**Corroboration: WEAK — single lane** (Codex `codex-ce.md`, correctness persona, confidence 75, validator-confirmed)
**File:line:** `gitnexus/src/mcp/local/local-backend.ts:2370` (formats `indexedAt: repo.indexedAt`)

`ensureInitialized()` can observe a newer `meta.indexedAt` and reopen the pool, but the degraded-warning
context is built from the cached `RepoHandle`'s fields rather than the freshly-observed metadata — a warm
backend can reopen against a newer on-disk index while still warning with old `indexedAt`. No Claude lane
independently corroborates this specific claim (searched all Claude lanes for `indexedAt` + staleness —
no hit). Per the digest rules, single-lane P2 findings are demoted to a body note rather than presented as
an inline candidate.

**Inline-candidate: NO** (single lane, P2 — flagged for awareness, not gated as postable).

---

### NEW-4 — P2 — `classifyFtsQueryError`'s Catalog-exception path treats a legitimately-unavailable FTS extension as a "real" per-table error, producing log/response spam for an expected degraded-capability state
**Corroboration: WEAK — single lane** (Codex `codex-adversarial.md`, native pass, "medium")
**File:line:** `gitnexus/src/core/search/bm25-index.ts:74-79`

When the FTS extension itself is unavailable (a documented, expected "capability" degradation per #2658),
LadybugDB's analogous Catalog-exception shape ("function X is not defined...") does not contain "does not
exist"/"doesn't have an index", so `classifyFtsQueryError` correctly returns `'other'` — but `'other'` is
treated as **non-benign**, so in MCP pool mode every configured table's query in that state produces a
`nonBenignErrors` entry, gets logged, and (per NEW-1's mechanism) can surface in the client warning — noisy
for offline/load-only installs where extension-unavailable is already known and expected. Note this is
**not** a misclassification bug — Claude's correctness lane (`claude-ce-correctness.md` F1) independently
verified the classifier's boundary is sound for this exact message shape (see §4). This is a distinct,
downstream UX/noise concern about what a correctly-computed `'other'` classification triggers, not a
classifier defect.

**Suggested fix (per the lane):** short-circuit FTS queries when `getExtensionCapabilities()` already
reports `fts.loaded === false`, rather than emitting one `nonBenignErrors` entry per table for an already-known
degraded-capability state.

**Inline-candidate: NO** (single lane, P2).

---

### NEW-5 — P3 — A classified non-benign FTS error can still be demoted to `debug`-level logging by a second, independent classifier
**Corroboration: WEAK — single lane** (Codex `codex-ce.md`, adversarial persona, confidence 75)
**File:line:** `gitnexus/src/mcp/local/local-backend.ts:2079-2082` (loop calling `logQueryError`) and
`:395-401`/`:412-421` (`logQueryError`/`isBenignMissingTableError`)
**[code-read] verified directly by this synthesis agent.**

`bm25Search`'s per-error loop calls `logQueryError(err)` for every entry already classified non-benign by
`classifyFtsQueryError`. But `logQueryError` independently re-classifies via its own, older, broader
`isBenignMissingTableError` regex (`/does not exist|no such (table|label|rel)|.../i`), which — unlike
`classifyFtsQueryError` — is **not** anchored to the start of the message or to the Binder/Catalog exception
class. A non-benign error that merely contains the substring "does not exist" anywhere (e.g. the adversarial
"echoed user query text" case the PR's own doc comment discusses) is logged at `debug`, not `warn`, silently
undercutting the severity signal this PR's classification work was meant to preserve server-side.

**Suggested fix:** bypass `logQueryError`'s independent re-classification for errors already routed through
`classifyFtsQueryError`, or add a force-warn path for pre-classified FTS errors.

**Inline-candidate: NO** (single lane, P3).

---

### NEW-6 — P2 (confidence 55) — `classifyFtsQueryError` cannot distinguish "FTS index missing" from "underlying graph table missing" within the Binder/Catalog exception class
**Corroboration: WEAK — single lane** (Claude `claude-ce-adversarial.md` Finding 2)
**File:line:** `gitnexus/src/core/search/bm25-index.ts:70-85`; precedent cited at `lbug-adapter.ts:2246`
(`// probe-recorded: Binder exception: Table CodeEmbedding does not exist.`)

Both "the FTS index on this table doesn't exist" and "the underlying graph TABLE doesn't exist" (a much
more serious condition — schema drift, stale/partial DB) produce Binder-exception messages containing
"does not exist," and both are classified `'missing-index'` (benign, silently dropped). The reviewer could
not confirm against a live LadybugDB instance whether `QUERY_FTS_INDEX` itself can actually emit the
table-missing shape (vs. only the intended index-missing shape) — grounded in confirmed precedent for the
identical message text elsewhere in this codebase, not a live repro against this exact call. No other lane
corroborates.

**Inline-candidate: NO** (single lane, P2, unverified against a live engine).

---

### NEW-7 — P2/P3 (confidence 40, advisory) — `lastObservedFtsStatus`/`lastObservedIndexedAt` are advanced before `initLbug` confirms the pool actually rolled over, unlike `lastObservedDbIdentity`
**Corroboration: WEAK — single lane, self-flagged as advisory by the reviewer** (`claude-ce-adversarial.md` Finding 3)
**File:line:** `gitnexus/src/mcp/local/local-backend.ts:1589-1608`

`lastObservedDbIdentity` is deliberately gated on `initLbug` returning `true` (a documented principle in
this file); the new `lastObservedFtsStatus` and pre-existing `lastObservedIndexedAt` are not — they're set
before `initLbug` is even called. The cold-init sub-case is traced and **confirmed harmless**
(self-correcting, one redundant no-op). The warm-reinit-throws sub-case is presented as a documented
inconsistency worth a second look, not a confirmed break — the reviewer could not construct a standalone
reproducible failure (backstopped by `identityChanged` in the common case).

**Inline-candidate: NO** (advisory-only per its own reviewer, single lane).

---

## 3. Confirmed — already flagged by author (still open)

- **Classifier duplication / no-`repoId` path still swallows real errors as "missing index"** — self-disclosed
  #1 (`lbug-adapter.ts:3125` / `bm25-index.ts:60`). Independently reproduced by 3 review lanes across both
  engines this round: Claude maintainability (`bm25-index.ts:191-201`, P2/78 — the `else` branch used by
  `hybrid-search.ts` and HTTP `/api/search` still bare-`catch{}`s every `queryFTS` failure), Codex `codex-swarm.md`
  risk-architect (same lines, P3), and Codex `codex-ce.md` maintainability+reliability+adversarial personas
  (P2, confidence 100). CONFIRMED — already flagged by author, still open.
- **Three parallel per-poolKey `Map`s on `LocalBackend`, consolidation candidate** — self-disclosed #2
  (`local-backend.ts:817`). Matches Claude maintainability Finding 3 exactly (five touch-points × three maps;
  suggests one `Map<string, PoolObservedState>`). CONFIRMED — already flagged by author, still open.
- **`redactPaths` stops at first whitespace; Windows username-with-space paths partially redacted** —
  self-disclosed #3 (`fts-indexes.ts:15`). Matches Claude adversarial Finding 4 (empirically reproduced the
  under-redaction) and Claude project-standards' residual-risk note. Elevated relevance: #2767 was reported
  on Windows, and this PR is the first caller to route raw, unbounded LadybugDB error text through
  `redactPaths` into a client-visible field. CONFIRMED — already flagged by author, still open.
- **No unit test isolates `ftsCapsChanged` from `identityChanged`** — self-disclosed #4
  ("flagged independently by 3 reviewers already" per the author's own prior pass). Independently
  re-confirmed this round by Claude testing (Finding 1, P1/68: the only test exercising this line also
  trips `identityChanged`, and the map is deliberately left unseeded at cold-init so the check is trivially
  `true` on the first post-cold-init read regardless of whether a repair happened) and Claude risk-architect
  (F1, mechanism-level: confirmed self-correcting/harmless in practice via `pool-adapter.ts`'s cheap
  `statDbIdentity`-reuse guard). Codex's `codex-ce.md` testing persona found the same gap independently but
  demoted it internally to a testing-gap note rather than a primary finding. CONFIRMED — already flagged by
  author, still open, and the mechanism itself (not just the missing test) is understood and traced as
  harmless in the paths that were checked.

---

## 4. Reviewed and found sound (verification, not fault-finding)

- **`classifyFtsQueryError` prefix/anchor logic is safe against real message shapes.** Claude correctness
  traced both concrete error producers (`Prepare failed: ` prefix confirmed real) and the hypothesized
  extension-unavailable false-positive — refuted: the real extension-unavailable message reads "function X
  is not defined," not "does not exist," so it correctly falls to `'other'`.
- **The "Runtime exception echoing user query text" adversarial case is handled correctly.** Both `Binder
  exception:`/`Catalog exception:` guards fail for a `Runtime exception:`-prefixed message, so it returns
  `'other'` before the substring check is ever reached — test-covered and passing.
- **`nonBenignErrors` empty-but-present array cannot occur.** Claude correctness traced the full invariant
  chain (`searchFTSFromLbug` only sets the field when `.length > 0`; every downstream consumer relies on
  this and it holds) — refuted as a concern.
- **The `partial` flag's inability to distinguish "enrichment failed" from "FTS partially failed" is a
  pre-existing pattern, not a regression.** Claude adversarial counted a dozen-plus existing `partial: true`
  sites in `local-backend.ts` with the same boolean-plus-descriptive-string shape; this PR's `ftsPartial` is
  a straightforward extension of that convention, not a new design decision.
- **Repair-vs-integrity-abort ordering is safe.** Claude risk-architect confirmed the verify-failure throw
  exits before the new stamp code runs — the stamp only executes on the happy path.
- **`existingMeta` null-safety and field-preservation are correct and test-verified** (`run-analyze.ts:1044-1074`) —
  confirmed by both Claude correctness and Claude maintainability, with byte-identical-preservation and
  legacy-backfill assertions in `run-analyze-fts-repair.test.ts`.
- **`redactPaths` coverage for the new raw-DB-error call site is complete; no unredacted path reaches a
  client response.** Security-boundary lane traced the full chain from `queryFTSViaExecutor`'s catch through
  to the client warning — every hop consumes an already-redacted string. Cypher injection is not a new risk
  (table/index names come only from the hardcoded `FTS_INDEXES` constant; user query text is only ever
  parameter-bound).
- **The `strict: false` TypeScript narrowing rationale in `FTSQueryOutcome`'s doc comment is empirically
  accurate, not a fabricated justification.** Claude maintainability reproduced the claimed `tsc` narrowing
  failure directly.
- **No leftover dead code from the `queryFTSViaExecutor` return-type migration**, and **no `AGENTS.md`
  `ingestion/`-naming or `GUARDRAILS.md` non-negotiable violations** — confirmed by Claude maintainability
  and Claude project-standards respectively; the diff touches no `ingestion/` files and all non-fatal writes
  are logged, not silently swallowed.

---

## 5. Test / CI / docs summary

- **CI:** prior full run (at `f688fbf`) — all checks passed, 15,958 tests, 0 failures, 70 skips. The current
  post-merge-from-main run had some shards still pending at observation time; only Vercel fails, expected
  for a fork PR without maintainer deploy auth. Zero human reviews on record yet (facts-historian).
- **Test additions are substantial and genuinely additive** — Claude testing confirmed via `git diff ... |
  grep '^-'` on all 5 changed test files that no existing assertion was weakened or removed; every removed
  line is an import/signature/mock-shape change.
- **Coverage gaps identified this round** (beyond the self-disclosed `ftsCapsChanged` gap in §3):
  - `classifyFtsQueryError`'s `Catalog exception:` branch has zero test coverage — every existing test uses
    `Binder exception:` (Claude testing Finding 2, confidence 78; independently flagged as a residual
    suspicion by Claude risk-architect S1; matches Codex's own testing-gap note).
  - The private `LocalBackend.bm25Search()` glue that forwards `nonBenignErrors` is never exercised
    un-mocked end-to-end — one test suite mocks `bm25Search` itself, another drives `searchFTSFromLbug`
    directly but never through `LocalBackend.query()` (Claude testing Finding 3, confidence 60).
  - No no-`repoId`/HTTP-path test exercises a real (non-"does not exist") `queryFTS` error — the pre-existing
    bare-`catch{}` in that branch (§3, NEW-1's dedup cluster) has zero coverage either direction (Claude
    maintainability, Codex `codex-swarm.md` lane 4, Codex `codex-ce.md`).
  - No test exercises the all-table non-benign-error case in a way that would catch NEW-1 regressing further
    (i.e., a test asserting the repair-fts advice is *absent* for that case) — noted by both Claude
    risk-architect and Codex `codex-ce.md`.
  - Minor/P3: the repair-fts meta re-read's `?? existingMeta` fallback branch is untested; multi-distinct-error
    truncation to `nonBenignErrors[0]` is untested with genuinely distinct per-table messages (Claude testing
    Findings 4-5, low confidence, low priority).
- **Docs (DoD verdict from `claude-swarm-docs-dod.md`):** two P3 gaps only, neither merge-blocking —
  `tools.ts`'s `query` tool description doesn't mention the `warning`/`partial` field changes (the `impact`
  tool already documents `partial: true` as precedent), and the new `FtsQueryFailureClass` type alias lacks
  a JSDoc comment (every other new exported symbol in this PR has one). CHANGELOG.md correctly untouched;
  ARCHITECTURE.md correctly not updated (per-tool MCP response shapes are below that doc's level).

---

## 6. Security summary

**All clear — no merge-blocking security findings.** The security-boundary lane traced all four assigned
verification points and confirmed each safe with code-level tracing, not just inspection:
1. `redactPaths()` coverage is complete for the new raw-DB-error data flow — no path exists where
   unredacted filesystem-path-bearing text reaches a client response.
2. The `lastErrorRedacted` JSDoc contract is upheld at its one call site, sourced from the same
   already-redacted array.
3. Cypher injection is not a new risk — table/index identifiers are hardcoded constants; user query text is
   only ever bound as a parameter, never interpolated.
4. The new `run-analyze.ts` meta.json write only spreads trusted, same-session data plus hardcoded literals;
   error text on write failure goes only to the server-side log, never into the written file or a client
   response.
5. `repo.name`/`repo.branch`/`indexedAt` in the warning text are intentional, non-path data (git-derived
   name/branch/ISO-timestamp) — correctly not redacted, given the local single-user MCP threat model.

Hygiene checks (both engines): `git diff --check` clean, no bidi/directional-override control characters,
no suspicious non-ASCII in changed lines (only benign em-dashes/curly quotes in prose and intentional CJK
test fixtures).

---

## 7. Coverage / engine-participation note

- **Claude lanes:** 5/5 `ce-code-review` personas (adversarial, correctness, maintainability, standards,
  testing) produced full structured output. 5/5 `gitnexus-pr-swarm-review` personas produced output, but two
  had liveness gaps worth noting transparently — this does not invalidate their content:
  - `claude-swarm-branch-hygiene.md`'s agent ended mid-investigation (last message was a tool-call preamble,
    despite 186s/30 tool calls of work) and was **not** re-dispatched; its domain substantially overlapped
    with `claude-swarm-facts-historian.md`, which independently confirmed the mergeability/merge-commit/
    file-scope facts this lane would have covered. The digest content above for branch hygiene rests on that
    overlap, not a completed standalone branch-hygiene pass.
  - `claude-swarm-docs-dod.md`'s agent also initially ended mid-investigation but **was** resumed and fully
    completed — its DoD verdict (§5) reflects a completed pass.
  - A standalone Claude swarm "test-ci-verifier" persona artifact (`claude-swarm-test-ci-verifier.md`) was
    persisted after this draft's first pass; it corroborates the CI-wiring and test-additivity claims in §5,
    but its own "Catalog exception is tested" claim was a mislabeling error (the cited test string is a
    `Binder exception:`, not `Catalog exception:` — coordinator-verified via direct grep: zero hits for the
    literal string "Catalog exception" anywhere in the test files). That correction reinforces, rather than
    contradicts, the Catalog-exception coverage gap already noted in §5 from `claude-ce-testing.md` and
    `claude-swarm-risk-architect.md`.
- **Codex lanes — all three fully participated, none unavailable or degraded:**
  - `codex-swarm.md` (native `gitnexus-pr-review`, 6 personas over GitNexus MCP tools: `impact`, `explain`,
    `detect_changes`, `context`, `cypher`) — verdict "production-ready with minor follow-ups," 2 P3 findings.
  - `codex-ce.md` (9-persona `ce-code-review` swarm — correctness, project-standards, testing, maintainability,
    agent-native, api-contract, performance, reliability, adversarial — with an internal validator gate that
    confirmed all 4 primary candidates) — verdict "Ready with fixes," 1 P1 + 2 P2 + 1 P3 finding.
  - `codex-adversarial.md` (native adversarial pass) — verdict "needs-attention," 2 medium findings.
  - Both `codex-ce.md`'s validator and `codex-swarm.md`'s GitNexus-graph-backed impact analysis (`RepoMeta`
    and `ftsDegradedWarning` flagged as high-blast-radius) independently converged on NEW-1 as the standout
    issue, which is why it carries the strongest cross-engine corroboration in this digest.

gitnexus/src/mcp/local/local-backend.ts
magyargergo: Fixed in f584f83b — split the `!ftsUsed` path on whether `ftsQueryErrors` is present. When it is, a new `ftsQueryFailedWarning()` (fts-indexes.ts) fires instead of `ftsDegradedWarning()`, with the redacted error as the headline and no `--repair-fts` guidance. Regression test: `query-degraded-signal.test.ts` — "surfaces a dedicated query-failed warning (not missing-index advice) when every table failed for a real error".

gitnexus/src/core/run-analyze.ts
magyargergo: Fixed in f584f83b — `saveEmbeddingCheckpoint` and the job's final save in `api.ts` now re-`loadMeta` immediately before each `saveMeta` instead of carrying one in-memory snapshot for the embedding job's full (up to 30min) lifetime, mirroring this file's own re-read-before-write pattern. No dedicated unit test added here — the fix is a mechanical two-line re-read matching an already-tested pattern, and no existing harness exercises this HTTP route's embedding job in isolation.

PR #1979, opened 2026-06-02
feat: upgrade MiniMax default model to M3

octo-patch:
## Summary
Upgrade MiniMax model configuration to use M3 as the default model.

## Changes
- Add `MiniMax-M3` to the model selection list and set as default
- Keep `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` as alternatives
- Remove older models (M2.5/M2.1/M2/M1)
- Update CLI default for `--model` flag (`minimax/minimax-m3`)
- Update related unit tests (`wiki-llm-client.test.ts`, `wiki-flags.test.ts`)
- Update i18n / locale strings (en + zh-CN, web + CLI)
- Update eval configs: replace `minimax-2.5.yaml` (M1 2.5) with `minimax-m3.yaml`; rename `minimax-m2.1.yaml` to `minimax-m2.7.yaml` (model ID also updated to `m2.7`)

## Why
`MiniMax-M3` is the latest model, with a 512K context window, up to 128K output, and image input support.

## Testing
- Unit tests updated to reference `minimax/minimax-m3`
- Default model resolution path and CLI help text verified by grep — no stale `M2.5` / `m2.5` references remain in code

PR #2780, opened 2026-08-01
feat: refresh MiniMax model and endpoint configuration

octo-patch:
Reason: Refresh MiniMax defaults, regional endpoints, thinking controls, and multimodal request support across the Web and CLI integrations.

Changes:
- Add the current model set with context, input modality, and thinking metadata.
- Add global and China endpoint selection to Web settings and CLI Wiki setup.
- Configure model-specific thinking requests, reasoning output separation, and image/video content blocks.
- Update CLI configuration, help text, shipped skill documentation, and focused tests.

Checks:
- `cd gitnexus && npm test -- test/unit/wiki-llm-client.test.ts test/unit/wiki-flags.test.ts`
- `cd gitnexus-web && npm test`
- `cd gitnexus && npm run build`
- `git diff --name-only -z | xargs -0 npx prettier --check`
- `git diff --name-only -z -- '*.ts' '*.tsx' | xargs -0 npx eslint` (no errors)
- `cd gitnexus && npm run test:unit` (10 failures in untouched platform/tooling tests; 10,847 tests passed)
- Scoped `detect-changes --scope staged` checks ran for the Web, Wiki, CLI, and storage indexes; the full-repository index was unavailable.


gitnexus/src/core/wiki/llm-client.ts
```diff
@@ -73,6 +121,7 @@ export async function resolveLLMConfig(overrides?: Partial<LLMConfig>): Promise<
 
   const apiKey =
     overrides?.apiKey ||
+    (savedProvider === 'minimax' ? process.env.MINIMAX_API_KEY : undefined) ||
```
magyargergo, line 124: **P1 (headline) — API-key resolution isn't provider-scoped, unlike `baseUrl`/`model`** `[code-read]`

`baseUrl` and `model` below are correctly gated behind `reuseSavedHttpConfig` (only reused when the saved provider matches the resolved one) — but `apiKey` and `apiVersion` are not. This chain falls straight through to `GITNEXUS_API_KEY`/`OPENAI_API_KEY`/`savedConfig.apiKey` regardless of whether the resolved provider is actually `minimax`.

**Manifestation (a):** a fresh install or CI box with no saved config and only `OPENAI_API_KEY`/`GITNEXUS_API_KEY` set (a previously-working, documented pattern) now resolves `provider: 'minimax'` by default (this PR's new fallback) and sends that OpenAI/generic key to `api.minimax.io` — silent 401, no error surfaced (the non-interactive guard only checks `!llmConfig.apiKey`, which is truthy).

**Manifestation (b), more severe:** in `gitnexus/src/cli/wiki.ts`, the provider-change reset block only backfills/clears `baseUrl`/`model` when switching *into* `'minimax'` (`if (options.provider === 'minimax')`) — one-directional. A user who previously configured MiniMax and runs `--provider openai --model gpt-4o` to switch away gets `provider: 'openai'` persisted with MiniMax's stale `baseUrl`+`apiKey` still attached. On the next run, `reuseSavedHttpConfig` evaluates `savedConfig.provider === savedProvider` → true (wiki.ts itself just wrote that equality), so the stale MiniMax `baseUrl`+key get reused under the `'openai'` provider name — the wiki generator's prompt (which can include source snippets) gets sent to MiniMax's servers even though the user explicitly left it.

**Suggested fix:** gate `apiKey`/`apiVersion` behind the same `reuseSavedHttpConfig` check already used for `baseUrl`/`model`/`isReasoningModel`; make the `wiki.ts` reset block symmetric (fire on any `providerChanged`, not only when switching into minimax).

Independently confirmed by 7 of 15 lanes this run — both Claude and all 3 Codex legs (genuine cross-engine corroboration): gitnexus-security-boundary-reviewer, gitnexus-risk-architect, ce-correctness-reviewer, ce-adversarial-reviewer (Claude), plus Codex B1/B2/B3.

gitnexus/src/cli/wiki.ts
```diff
@@ -221,6 +223,18 @@ const wikiCommandImpl = async (inputPath?: string, options?: WikiCommandOptions)
     if (options.provider) updates.provider = options.provider;
     if (options.apiVersion) updates.apiVersion = options.apiVersion;
     if (options.reasoningModel !== undefined) updates.isReasoningModel = options.reasoningModel;
+    const providerChanged = !!options.provider && options.provider !== existing.provider;
+    if (options.provider === 'minimax') {
```
magyargergo, line 234: **P1 — this reset only fires when switching *into* minimax, not out of it** `[code-read]`

See the primary inline comment on `llm-client.ts:124` for the full mechanism — this is the persistence-side half of that same headline finding. Because this block is gated on `options.provider === 'minimax'`, switching a saved MiniMax config to any other provider (`openai`/`openrouter`/`azure`/`custom`) leaves MiniMax's `baseUrl` and `apiKey` on disk under the new provider name, which `resolveLLMConfig`'s `reuseSavedHttpConfig` then reuses.

**Suggested fix:** make this block symmetric — reset/clear provider-specific `model`/`baseUrl`/`apiVersion`/reasoning flags on *any* `providerChanged`, not only the minimax-target case.

magyargergo (commented):
# PR #2780 Tri-Review Digest — "feat: refresh MiniMax model and endpoint configuration"

**Repo:** abhigyanpatwari/GitNexus · **Author:** octo-patch (bot, fork PR) · **Head:** `179c14a1b` · **Diff base:** `84f584449`

## Engine & lane breakdown

This is a genuine **two-engine, fully-live tri-review** — both engines ran to completion this pass, unlike a prior run in this session where Codex hit a spend cap.

- **Claude Engine A1 (GitNexus PR swarm, 6 lanes):** facts-historian, branch-hygiene, risk-architect, test-ci-verifier, security-boundary, docs-dod.
- **Claude Engine A2 (ce-code-review personas, 6 lanes):** correctness, testing, maintainability, project-standards, adversarial, api-contract.
- **Codex (GPT-5.5/xhigh), all 3 legs live:** B1 swarm-solo, B2 ce-report-only, B3 adversarial.
- **15 lane reports total**, plus this coordinator's own direct source reads and one executed test run (re-confirmed independently during synthesis).

## PR summary

The PR upgrades MiniMax from a model-name string routed through OpenRouter (`minimax/minimax-m2.5`) to a first-class native `minimax` provider with its own endpoints — OpenAI-compatible for the CLI (`api.minimax.io/v1`), Anthropic-compatible for the web UI (`api.minimax.io/anthropic`) — plus new models (MiniMax-M3, MiniMax-M2.7), a provider-specific thinking-mode system, a regional endpoint selector, and multimodal (image/video) content-block plumbing. It also silently makes `minimax` the new default LLM provider for any user/CI pipeline with no saved config, replacing the previous `openai`/OpenRouter default. 20 files changed (~661/-65 lines) across CLI (`gitnexus/`) and web (`gitnexus-web/`) LLM configuration, settings UI, i18n, and skill docs.

## Bottom line

**Branch hygiene: clean, merge-safe.** One feature commit (bot-authored) plus one human-authored merge-from-main commit that cleanly absorbs an unrelated Python PR (#2770) with zero conflicts and zero MiniMax-file entanglement. All 20 substantive files belong to one coherent domain (MiniMax provider integration across CLI+web, which legitimately need to share model IDs/thinking-mode constants). No cleanup required before review.

**Merge state: checks not fully green, no code-driven failure posted yet.** GitHub reports `MERGEABLE` / `mergeStateStatus: BLOCKED`. The only currently-failing check is Vercel (`Authorization required to deploy` — expected/non-blocking for a bot fork PR, not a code issue). At review time several shards (`tests/ubuntu/coverage`, `tests/macos-latest`, `tests/windows-latest`, `tests/benchmarks`) were pending; Codex's later check found all but one Windows shard had since gone green. Lint, typecheck (both packages), CodeQL, e2e (chromium), tree-sitter ABI, and packaged-install-smoke all pass. However, this digest confirms (by execution) that one shipped test **will** fail once its shard runs — see Headline Finding 2.

**Verdict: not production-ready.** Two hard blockers drive this: (1) a provider/credential-isolation gap in `resolveLLMConfig` that can silently send one vendor's API key — and in one path, the user's actual prompt/repo content — to MiniMax's servers instead of the intended provider, confirmed independently by both engines across 7 of 15 lane reports; and (2) a shipped unit test that asserts a non-existent LangChain property and fails on execution today, confirmed by direct test run. Both are narrow, mechanically-clear fixes (gate `apiKey`/`apiVersion` behind the same `reuseSavedHttpConfig` check already used for `baseUrl`/`model`; fix one assertion), not a redesign — but they are exactly the kind of silent, credential-adjacent regression this review process exists to catch before merge. Secondary P2s (legacy-model migration gap, empty-string model default for other HTTP providers, cross-package constant duplication) are real but not merge-blocking.

## What's solid

- i18n parity (en/zh-CN) is complete and correctly structured across both the web settings JSON and the CLI help text — verified key-by-key by docs-dod and project-standards.
- `CHANGELOG.md` correctly left untouched, per repo convention.
- The interactive wizard's new choice-5/choice-6 numbering for MiniMax Global/China is collision-free and correctly cascades subsequent local-provider choices.
- `validateLLMBaseUrl` remains unconditionally in place on the CLI path (no SSRF/echo regression versus the prior PR #1456 fix); the web UI restricts MiniMax `baseUrl` to a fixed two-value `<select>`, so there's no new free-text URL entry surface.
- The `always_on` implicit-thinking behavior for MiniMax-M2.7 (no `thinking` field sent) is intentional and consistently implemented across both the CLI and web request paths, and is covered by an explicitly-named test on the CLI side.
- Test additions for the new request-shaping logic (thinking-mode variants, base-URL selection, multimodal pass-through) are well-targeted for their scope, even though coverage gaps remain at the edges (see below).
- The genuinely refuted P0 (see Refuted section) shows the review process itself working as intended — a plausible-sounding claim was caught and killed before reaching this digest.

---

## Headline findings (P1)

### 1. Provider/credential isolation gap in `resolveLLMConfig` — one vendor's API key (and in one path, live prompt content) can be silently sent to MiniMax's servers

- **Files:** `gitnexus/src/core/wiki/llm-client.ts:98–158` (root cause), `gitnexus/src/cli/wiki.ts:211–247` (persistence sibling)
- **Tag:** `[code-read]`
- **Found by (7 of 15 lanes — the strongest corroboration in this run):** gitnexus-security-boundary-reviewer (Claude A1, P1, blocks merge), gitnexus-risk-architect (Claude A1, P1, FINDING-01), ce-correctness-reviewer (Claude A2, P1, confidence 78), ce-adversarial-reviewer (Claude A2, P1 confidence 85 + P2 confidence 75), **Codex B1 swarm** (P1, high confidence), **Codex B2 ce-review** (P1, confidence 75, "Ready with fixes" gate), **Codex B3 adversarial** (`[high]`, independently constructed trigger). This is genuine cross-engine (Claude+Codex) *and* cross-leg-within-Codex (all 3 Codex legs) corroboration.

**Root cause (verified directly against source):** the new `reuseSavedHttpConfig` gate correctly scopes `baseUrl` (line 135), `model` (line 145), and `isReasoningModel` (line 154) to the resolved provider — but `apiKey` (lines 122–128) and `apiVersion` (line 150) were left out of that mechanism and fall straight through to environment vars / `savedConfig.apiKey` regardless of provider match:

```ts
const apiKey =
  overrides?.apiKey ||
  (savedProvider === 'minimax' ? process.env.MINIMAX_API_KEY : undefined) ||
  process.env.GITNEXUS_API_KEY ||
  process.env.OPENAI_API_KEY ||
  savedConfig.apiKey ||
  '';
```

**Two concrete, independently-triggerable manifestations of the same root cause:**

- **(a) Silent default-provider flip.** A brand-new user or CI box with no saved config and `OPENAI_API_KEY`/`GITNEXUS_API_KEY` set (a documented, previously-working pattern) now resolves `savedProvider = 'minimax'` (line 103, new default) purely by absence of config, and the key chain falls through to the OpenAI/GITNEXUS key since `MINIMAX_API_KEY` is unset. That credential is sent as the `Authorization: Bearer` header to `https://api.minimax.io/v1`. The non-interactive guard in `wiki.ts` only checks `!llmConfig.apiKey`, which is now truthy, so no error surfaces — the CLI just proceeds against the wrong vendor.
- **(b) Provider-switch leak (the more severe variant — confirmed via direct read of `wiki.ts:211–247`).** `wiki.ts`'s provider-change reset block only backfills/resets `baseUrl`/`model`/`isReasoningModel` when the CLI flag switches **into** `'minimax'` (`if (options.provider === 'minimax')`, line 227) — it is one-directional. A user who previously configured MiniMax and later runs `gitnexus wiki --provider openai --model gpt-4o` to switch away gets `{...existing, ...updates}` persisted with `provider: 'openai'` but **MiniMax's stale `baseUrl` and `apiKey` still attached** (never reset, since `options.apiKey` wasn't passed and the reset block didn't fire). On the very next resolution, `reuseSavedHttpConfig` evaluates `savedConfig.provider === savedProvider` → `true` (because `wiki.ts` itself just wrote that equality to disk), so the stale MiniMax `baseUrl`+`apiKey` are reused under the new `'openai'` provider name. Net effect: the wiki generator's prompt — which can include source code snippets — is transmitted to MiniMax's servers using a MiniMax key, requesting model `"gpt-4o"`, even though the user explicitly asked to leave MiniMax. The same asymmetry also applies to `apiVersion` and to any other HTTP-provider pair (e.g. saved `azure` → switch to `minimax`, confirmed by ce-adversarial).

**Suggested fix:** gate `apiKey` and `apiVersion` behind the same `reuseSavedHttpConfig` check already used for `baseUrl`/`model`/`isReasoningModel`, and make the `wiki.ts` provider-change reset block symmetric (fire on any `providerChanged`, not only when switching *into* `'minimax'`). Add regression tests for: `{}` + `GITNEXUS_API_KEY` env → provider defaults to minimax but key resolution should require `MINIMAX_API_KEY`; and an explicit `openai → minimax` / `minimax → openai` switch via `--provider` with a pre-existing `savedConfig.apiKey`.

**Related, same root cause, folded in rather than double-counted:** docs-dod-reviewer flagged the missing `BREAKING CHANGE:` notation in the PR body for the default-provider flip (CONTRIBUTING.md's convention); facts-historian independently observed the default-provider change is undersold relative to the "refresh" framing in the PR title. Both are documentation-completeness angles on the same underlying behavior change, not separate defects.

---

### 2. Shipped test asserts a non-existent property and fails on execution today

- **File:** `gitnexus-web/test/unit/agent-history.test.ts:266`
- **Tag:** `[reproduced]` — the only reproduced-tag finding in this run; confirmed by three independent test executions: ce-testing-reviewer (Claude A2, ran the suite during its own review pass), the coordinator (ran the single named test directly against the worktree during the live review), and the synthesis agent (re-ran it fresh during digest synthesis)
- **Found by:** gitnexus-docs-dod-reviewer (Claude A1, via source inspection of `node_modules`), ce-testing-reviewer (Claude A2, by running the suite), and independently re-confirmed just now during synthesis.

**Trigger:** `npx vitest run test/unit/agent-history.test.ts -t "keeps MiniMax-M2.7 thinking always on"` (run from `gitnexus-web/`) fails:

```
FAIL  test/unit/agent-history.test.ts > createChatModel > keeps MiniMax-M2.7 thinking always on
AssertionError: expected undefined to be false
 ❯ test/unit/agent-history.test.ts:266:41
    266|     expect(model.thinkingExplicitlySet).toBe(false);
```

`thinkingExplicitlySet` is not a property `@langchain/anthropic`'s `ChatAnthropic` ever sets (confirmed by grepping the installed package's `dist/chat_models.js` — no such identifier exists). Because `model` is cast `as any` in the test, TypeScript can't catch the typo at compile time.

**Suggested fix:** `expect(model.thinking).toBeUndefined();` — this mirrors the sibling assertion `"leaves always-on thinking implicit"` already used for the identical case on the CLI side (`gitnexus/test/unit/wiki-llm-client.test.ts:288-290`), and matches `agent.ts`'s actual `always_on` branch (`thinking = thinkingMode && thinkingMode !== 'always_on' ? {type: thinkingMode} : undefined`).

**Note:** this is a one-line test-assertion bug, not a product bug — the underlying MiniMax-M2.7 implicit-thinking feature behaves correctly. It will fail CI once the `tests / ubuntu / coverage` shard that runs this file completes (it had not yet run at initial review time).

---

## Secondary findings (P2)

### 3. Legacy MiniMax-M2.5 saved configs silently get a new `thinking` param injected via default-merge — not a no-op as one lane concluded

- **Files:** `gitnexus-web/src/core/llm/settings-service.ts:56-59` (`mergeWithDefaults`), `types.ts:267-274` (`DEFAULT_LLM_SETTINGS.minimax.thinkingMode: 'adaptive'`), consumed at `agent.ts:281-292`
- **Tag:** `[code-read]`
- **Found by:** gitnexus-risk-architect (Claude A1, FINDING-02, P2) and ce-api-contract-reviewer (Claude A2, Finding 1, P1/confidence 68) independently, both correctly traced; **verified directly by this synthesis agent** via source read of `mergeWithDefaults` (`settings-service.ts:56-59`) and `providerBuilders.minimax` (`settings-service.ts:323-326`).

**Trigger:** any browser with a pre-PR persisted `minimax.model === 'MiniMax-M2.5'` setting (the old default, and the value every pre-existing MiniMax web user has). `mergeWithDefaults` spreads `DEFAULT_LLM_SETTINGS.minimax` (which now includes `thinkingMode: 'adaptive'`) *before* spreading the parsed/stored value — since the stored object predates the `thinkingMode` field entirely, the merge injects `thinkingMode: 'adaptive'` into every legacy config. `getActiveProviderConfig()`'s `providerBuilders.minimax` spreads this merged object wholesale, so `createChatModel` receives `minimaxConfig.thinkingMode === 'adaptive'` (not `undefined`). Because `getMiniMaxModelCapabilities('MiniMax-M2.5')` returns `undefined` (removed from the capability table), the resulting request sends `model: 'MiniMax-M2.5'` with `thinking: {type: 'adaptive'}` — a parameter combination that never existed for this model/version pairing pre-PR, to a model id the PR's own capability table treats as retired.

**Cross-check note:** ce-correctness-reviewer's item-1 disposition concluded this scenario was "behaviorally identical to pre-PR" and refuted it — that conclusion assumed `requestedThinkingMode` starts `undefined` for legacy configs, but doesn't account for the `mergeWithDefaults` default-injection traced above (confirmed by direct source read during this synthesis). The risk-architect / ce-api-contract read is the correct one; the ce-correctness refutation on this specific sub-point does not hold.

**Suggested fix:** in `mergeWithDefaults`/`loadSettings`, detect a `minimax.model` not present in `MINIMAX_MODEL_IDS` and either reset to `MINIMAX_MODEL_IDS[0]` or skip injecting `thinkingMode` for unrecognized models, mirroring the file's existing one-time-migration pattern. Add a regression test seeding `{minimax: {model: 'MiniMax-M2.5'}}` through `loadSettings()` → `createChatModel()`.

### 4. CLI's generic-HTTP-provider model default silently becomes `''` for non-MiniMax providers

- **File:** `gitnexus/src/core/wiki/llm-client.ts:139-146`
- **Tag:** `[code-read]`
- **Found by:** ce-api-contract-reviewer (Claude A2, Finding 2, P1/confidence 68) with corroborating context from risk-architect's Shape-B trace (Claude A1)

Pre-PR, the unconditional fallback was `savedConfig.model || 'minimax/minimax-m2.5'` — every HTTP-style provider with no saved model got *some* non-empty default. Post-PR, only `savedProvider === 'minimax'` gets a non-empty fallback; `openai`/`openrouter`/`azure`/`custom` with no saved/overridden model now resolve to `model: ''`, sent verbatim in the request body. Reachable non-interactively: `gitnexus wiki --provider openai` (no `--model`) persists `{provider: 'openai'}` with no model, and `resolveLLMConfig` computes `reuseSavedHttpConfig = true` but `savedConfig.model` is `undefined` → `model: ''`. Note this is not a clean regression from "working" to "broken" (the old `'minimax/minimax-m2.5'` string was itself invalid against a genuine `openai` endpoint) — it's "broken one way" replaced by "broken a different way," which is why this is rated P2 rather than P1. **Suggested fix:** extend `wiki.ts`'s provider-change backfill logic to all non-local providers, not just `'minimax'`, or restore a sane generic default.

### 5. MiniMax model catalog and thinking-mode-resolution logic duplicated across packages and call sites, with no compile-time drift guard

- **Files:** `gitnexus-web/src/core/llm/types.ts:33` vs `gitnexus/src/core/wiki/llm-client.ts:28` (`MINIMAX_MODEL_IDS`, `MiniMaxThinkingMode` — byte-identical, independently declared); `agent.ts:281-292`, `SettingsPanel.tsx:344-348`, `SettingsPanel.tsx:878-880` (the "clamp thinking mode to what the model supports" policy implemented 3x with subtly different fallback tails)
- **Tag:** `[code-read]`
- **Found by:** ce-maintainability-reviewer (Claude A2, Findings 1 & 2), ce-testing-reviewer (Claude A2, P2), docs-dod-reviewer (Claude A1, Finding 6), gitnexus-risk-architect (Claude A1, FINDING-04), gitnexus-pr-facts-historian (Claude A1, Finding E) — five Claude lanes converge on the same duplication theme, no Codex corroboration on this one.

Both packages already import from `gitnexus-shared` for other cross-cutting concerns, confirmed empty of MiniMax exports today — this wasn't a package-boundary necessity. The CLI's `resolveMiniMaxThinkingMode` reimplements the web's capability-map policy via positional array indexing (`MINIMAX_MODEL_IDS[0]`/`[1]`) rather than referencing the same source of truth, so a future third model added to one side wouldn't be caught by any test or compiler. **Suggested fix:** move `MINIMAX_MODEL_IDS`, `MiniMaxThinkingMode`, `MINIMAX_MODEL_CAPABILITIES`, and a single `resolveEffectiveMiniMaxThinkingMode` helper into `gitnexus-shared`; have `agent.ts` and both `SettingsPanel.tsx` call sites use the one helper instead of three independently-hand-written ternary chains.

### 6. `type: 'video'` content blocks are silently dropped by LangChain's Anthropic formatter; `type: 'image'` works correctly

- **File:** `gitnexus-web/src/core/llm/agent.ts:429-434` (the `as any` cast site), `types.ts:77-86` (`AgentUserContent` shape)
- **Tag:** `[code-read]`
- **Found by:** ce-adversarial-reviewer (Claude A2, Finding 5, confidence 80) — a genuinely new correctness bug, confirmed by reading the pinned `@langchain/anthropic@1.5.1` source directly (`node_modules/@langchain/anthropic/dist/utils/message_inputs.js`), not surfaced by any other lane.

`_isAnthropicImageBlockParam` in the pinned LangChain version exactly matches this PR's `{type:'image', source:{...}}` shape and passes it through unmodified — images work. There is **no branch anywhere** in `_formatContentBlocks`'s `if/else if` chain for `type === "video"` and no trailing `else` — a video-only content block is silently skipped, producing `content: []` (which MiniMax's API will likely reject) or silently dropping the video reference if accompanied by text. **Currently latent:** no shipped UI in this PR constructs an `AgentUserContent` array with `type:'image'`/`type:'video'` yet (`useAppState.tsx`'s message construction is always a plain string), so this is unreachable today — but the PR's own new test ("preserves MiniMax image and video content blocks") only asserts `HumanMessage` construction, not the real wire-serialization path, creating false confidence that video support works end-to-end. **Suggested fix:** either translate `type:'video'` before it reaches `buildLangChainMessages`, or add a runtime guard/warning, before any upload UI wires into this path. Add a test against `ChatAnthropic`'s actual serialization entry point, not just `HumanMessage` construction.

### 7. `preserveAssistantTranscript` capability gap for MiniMax's thinking/tool-use replay (Codex-only finding, not caught by any Claude lane)

- **File:** `gitnexus-web/src/core/llm/settings-service.ts:378-388` (`PROVIDER_CAPABILITIES`), consumed at `useAppState.tsx:1073-1094`, flattened at `agent.ts:465` (`serializeAgentHistoryMessages`)
- **Tag:** `[code-read]`
- **Found by:** Codex B1 swarm (P1, high confidence) and Codex B3 adversarial (`[medium]`) — both Codex legs independently reached this; **zero Claude lanes surfaced it.** Verified at the code level directly: `deepseek` is confirmed the only key in `PROVIDER_CAPABILITIES` (line 383), `minimax` is absent and falls to `DEFAULT_PROVIDER_CAPABILITIES.preserveAssistantTranscript: false` (line 379).

The MiniMax web path enables Anthropic-style `thinking` mode by default for M3, but only DeepSeek gets `preserveAssistantTranscript: true`; with it `false` for MiniMax, follow-up turns rebuild history from flattened plain-text assistant content (`normalizeMessageContent`) rather than replaying the original `thinking`/`tool_use` content blocks. **Confidence split:** the code-level gap (MiniMax absent from the capability map, same mechanism DeepSeek needed for the identical reasoning-model transcript-replay problem) is solid. Codex's claim about *why* this matters — that MiniMax's Anthropic-compatible API requires full content-block replay for multi-turn tool-use, citing MiniMax's own docs — is external API-behavior knowledge this digest cannot independently verify from the repo, so treat that specific severity justification as lower-confidence than the code-gap observation itself. Rated P2 rather than P1 pending confirmation of the external-API requirement. **Suggested fix:** extend `PROVIDER_CAPABILITIES` to include `minimax: {preserveAssistantTranscript: true}` and verify the history serializer round-trips Anthropic content blocks unchanged for it; add a MiniMax multi-turn thinking+tool-call test.

### 8. New `any`-typed code in `agent.ts` — matches a known user preference, not a documented repo rule

- **Files:** `gitnexus-web/src/core/llm/agent.ts:434` (`new HumanMessage({ content: message.content as any })`), and two new `(block: any) =>` parameter annotations at **lines 574-575** (mirroring an already-existing identical pattern at lines 570-571)
- **Tag:** `[code-read]`
- **Found by:** gitnexus-security-boundary-reviewer, gitnexus-risk-architect, gitnexus-pr-facts-historian (Claude A1, all three), ce-maintainability-reviewer, ce-project-standards-reviewer (Claude A2), Codex B2 ce-review (#3, P2) — cross-engine corroboration on the existence of the cast, but **the framing differs** across lanes.

**Correction to how several lanes framed this:** the security-boundary and risk-architect reports cite "CLAUDE.md is unambiguous: Strict typing, no any" as a written project rule. Verified directly: no such rule exists anywhere in `CLAUDE.md`/`AGENTS.md`/`GUARDRAILS.md` — the only mechanism is `eslint.config.mjs:92`'s `'@typescript-eslint/no-explicit-any': 'warn'`, which is non-blocking and does not fail CI. ce-project-standards-reviewer correctly identified this distinction and declined to cite it as a standards violation. The `as any`/`(block: any)` additions are real and match a consistent, previously-stated user preference against `any`/`as any` — worth flagging on that basis — but should not be characterized as violating written repo policy. Test-file `as any` usage (51+ pre-existing instances, more added by this PR) is confirmed to be an established codebase pattern, not new debt introduced here. **Suggested fix (optional, non-blocking):** import LangChain's own `MessageContent`/`ContentBlock` type for line 434 if it type-checks against `AgentUserContent`; reuse a typed `ContentBlock` union for lines 574-575 to match (or fix alongside) the pre-existing pattern at 570-571.

---

## Lower-priority / P3

- **OpenRouter interactive setup lost its default model.** `wiki.ts:433` (`defaultModel = '';` for choice `'2'`) — pressing Enter with no input at the "Model:" prompt now aborts with "No model provided" instead of falling back to the old (if imperfect) `'minimax/minimax-m2.5'` default. Confirmed via direct source read. UX regression, not merge-blocking (risk-architect FINDING-03).
- **Stale README references.** `README.md:743` and `gitnexus/README.md:258` still say `(default model: minimax/minimax-m2.5)` — confirmed via direct read; the PR updated the equivalent line in three SKILL.md copies and the CLI help text but missed these two. Corroborated across engines: ce-api-contract-reviewer (Claude A2) and Codex B2 ce-review (#2, "confidence 100") both independently found this. Trivial fix.
- **Competing open PR #1979** ("upgrade MiniMax default model to M3"), same bot author, zero reviews, overlaps 18/20 of this PR's files — should be closed with a comment pointing to #2780 concurrently with merging this one, to avoid a stale duplicate conflict later (facts-historian, Finding B).
- **`--reasoning-model` help text** no longer documents `o1/o3/o4-mini` even though the flag still activates `isReasoningModel` for non-MiniMax providers (facts-historian, Finding D).
- **Misleading test name**: `wiki-llm-client.test.ts`'s `'uses the China endpoint with adaptive thinking by default'` actually passes an explicit `baseUrl` override — China is not the default, global is (facts-historian, Finding C).
- **`LLMProvider` (CLI) vs `CLIConfig.provider` (repo-manager.ts)** hand-duplicated unions, currently in sync but with no shared source of truth (ce-api-contract, Finding 4).
- **Missing WHY-comments** on `hasLegacyHttpConfig`/`reuseSavedHttpConfig` (llm-client.ts:101-105), the M2.7 `always_on` implicit-thinking branch (agent.ts:281-292), and the SettingsPanel 3-level thinking-mode fallback — none block merge but create real future-regression risk if a well-meaning contributor "simplifies" code whose invariants aren't documented (docs-dod, Findings 3-5).
- **SKILL.md files don't document `--provider`** at all, even though its default just changed (docs-dod, Finding 8).
- **`SettingsPanel.tsx` fallback-default divergence**: `SettingsPanel.tsx` defaults an unrecognized model's thinking-mode display to `'adaptive'`, while `agent.ts`'s `createChatModel` omits the `thinking` field entirely in the same case — a UI-vs-actual-request mismatch, not a malformed request (ce-correctness, Finding 3, confidence 45).
- **`SettingsPanel.tsx`** is already 1000+ lines and gained ~83 net lines inline rather than as an extracted component — consistent with the file's pre-existing convention for other providers, so not a new deviation (ce-maintainability, Finding 5).

### Refuted — explicitly dropped, not carried forward

- **test-ci-verifier's original "P0: Endpoint Configuration Mismatch"** (claiming the CLI's `/v1` MiniMax base URL and the web's `/anthropic` base URL disagreeing is a bug) is **false**. Verified directly: the web package constructs a LangChain `ChatAnthropic` client for MiniMax (genuinely needs the Anthropic-Messages-shaped endpoint), while the CLI's hand-rolled client appends `/chat/completions` (needs the OpenAI-Chat-Completions-shaped endpoint). MiniMax's platform intentionally exposes both surfaces for the same models — a standard LLM-gateway pattern, not a defect. The lane's original "DO NOT MERGE" recommendation, which was anchored almost entirely on this claim, does not carry forward; the rest of that lane's report (coverage-gap findings, folded into Secondary Findings above) stands on its own merits.
- **ce-correctness-reviewer's item-1 refutation** ("legacy MiniMax-M2.5 fallback in `agent.ts` is behaviorally identical to pre-PR, not a malformed-request risk") does not hold under a full trace of `mergeWithDefaults` — see Secondary Finding 3 above, where this is corrected using risk-architect's and ce-api-contract's independently-reached (and directly re-verified) analysis.
- **`resolveMiniMaxThinkingMode`/MiniMax-M2.7 always-on implicit `thinking`** — one lane's initial concern that this might be an unintentional gap is refuted: it's confirmed intentional and covered by an explicitly-named test (`'leaves always-on thinking implicit'`).
- **Two independently-defined `MINIMAX_MODEL_IDS` causing a live bug today** — ce-adversarial explicitly checked and refuted present-day impact (the two lists are identical today; this is a future-drift risk only, already captured as Secondary Finding 5, not a live defect).

---

## CI status

At initial review time: lint/typecheck (both packages)/CodeQL/e2e(chromium)/tree-sitter-ABI/packaged-install-smoke/gitleaks all **PASS**. `Vercel` **FAILS** on authorization only (expected for a bot/fork PR, not code-related, non-blocking). Pending at that point: `tests/ubuntu/coverage` (1-3/3, the sharded `gitnexus/` CLI-package suite), `tests/ubuntu/coverage merge` (which — confirmed by reading `.github/workflows/ci-tests.yml` — separately runs `gitnexus-web`'s full unit-test suite, including `agent-history.test.ts`, after all coverage shards finish), `tests/macos-latest` (2-3/3), `tests/windows-latest` (1-3/3), `tests/benchmarks`. Codex's later check (same run) found the numbered `tests/ubuntu/coverage` shards had gone green, with only one Windows shard still pending and Vercel still failing on auth — that result is for the `gitnexus/` CLI suite only and says nothing about the separate `coverage merge` job that actually executes `gitnexus-web`'s tests. **No CI failure for the `agent-history.test.ts` regression had posted as of either check** (the `coverage merge` job hadn't run yet at either check point) — but this digest confirms by direct, repeated execution (see Finding 2) that it fails deterministically whenever that job runs. Treat CI as "not yet fully green; the one code-driven failure this review found has not yet surfaced in a CI run, but will as soon as the `ubuntu / coverage merge` job executes."

## Coverage note

Full diff read across all 20 changed files (both CLI `gitnexus/` and web `gitnexus-web/` packages, i18n, and skill docs); no scope skipped. Key line-number citations in this digest were re-verified directly against the PR-head worktree (`/workspace/.worktrees/pr-2780`) during synthesis, not taken solely on lane-report trust, including a fresh, independent re-execution of the one `[reproduced]`-tagged test.

---

*This is an automated multi-lane digest synthesized from 15 independent review-lane reports (12 Claude + 3 Codex) plus direct coordinator source verification. Verify findings against current source before acting — code may have moved since this review ran.*

gitnexus-web/test/unit/agent-history.test.ts
```diff
@@ -206,6 +225,48 @@ it('drops reasoningContent from serialized assistant messages without tool calls
...
+  it('keeps MiniMax-M2.7 thinking always on', () => {
+    const model = createChatModel({
+      provider: 'minimax',
+      apiKey: 'minimax-test-key',
+      model: MINIMAX_MODEL_IDS[1],
+      thinkingMode: 'disabled',
+      temperature: 0.1,
+    } as any) as any;
+
+    expect(model.thinkingExplicitlySet).toBe(false);
```
magyargergo: **P1 — this assertion fails on execution today `[reproduced]`**

`model.thinkingExplicitlySet` is not a property `@langchain/anthropic`'s `ChatAnthropic` ever sets (confirmed by grepping the installed package — no such identifier exists; the real properties are `thinking`/`_thinkingInParams`). Because `model` is cast `as any` two lines up, TypeScript can't catch this. Reproduced directly: `cd gitnexus-web && npx vitest run test/unit/agent-history.test.ts -t "keeps MiniMax-M2.7 thinking always on"` → `AssertionError: expected undefined to be false`.

**Suggested fix:** `expect(model.thinking).toBeUndefined();` — mirrors the sibling CLI-side test `'leaves always-on thinking implicit'` and matches `agent.ts`'s actual `always_on` branch. One-line fix; the underlying MiniMax-M2.7 implicit-thinking behavior itself is correct.

This will fail CI once the `ubuntu / coverage merge` job (which runs `gitnexus-web`'s full unit-test suite, confirmed via `.github/workflows/ci-tests.yml`) executes — that job hadn't run yet at review time. Independently confirmed by 3 sources: gitnexus-docs-dod-reviewer (source inspection), ce-testing-reviewer (ran the suite), and the coordinator (direct re-execution).

gitnexus-web/src/core/llm/types.ts
```diff
@@ -200,7 +267,9 @@ export const DEFAULT_LLM_SETTINGS: LLMSettings = {
   },
   minimax: {
     apiKey: '',
-    model: 'MiniMax-M2.5',
+    model: MINIMAX_MODEL_IDS[0],
+    baseUrl: MINIMAX_ANTHROPIC_BASE_URLS.global_en,
+    thinkingMode: 'adaptive',
```
magyargergo, line 272: **P2 — this new `thinkingMode: 'adaptive'` default silently gets injected into legacy `MiniMax-M2.5` saved configs** `[code-read]`

`settings-service.ts`'s `mergeWithDefaults` spreads `DEFAULT_LLM_SETTINGS.minimax` (now including this new `thinkingMode: 'adaptive'` field) *before* spreading the stored/parsed value: `{ ...DEFAULT_LLM_SETTINGS.minimax, ...parsed?.minimax }`. A config saved before this PR (`{model: 'MiniMax-M2.5', apiKey, temperature}` — no `thinkingMode` key at all) doesn't overwrite the default, so `thinkingMode: 'adaptive'` survives into every legacy config. `getMiniMaxModelCapabilities('MiniMax-M2.5')` returns `undefined` (removed from the capability table below), so `createChatModel` still uses the injected `'adaptive'` and sends `thinking: {type: 'adaptive'}` for a model/parameter combination that never existed pre-PR.

**Suggested fix:** in `mergeWithDefaults`/`loadSettings`, detect a `minimax.model` not present in `MINIMAX_MODEL_IDS` and either reset to the new default or skip injecting `thinkingMode` for unrecognized models. Add a regression test seeding `{minimax: {model: 'MiniMax-M2.5'}}` through `loadSettings()` → `createChatModel()`.

Confirmed independently by gitnexus-risk-architect and ce-api-contract-reviewer; one lane's refutation of this scenario didn't hold once the merge-default spread order was traced directly (verified in `settings-service.ts:56-59`).

azizur100389:
Reviewed current head `14319be0` on current `upstream/main` `99291891`.

The latest head fixes the previously reported test-assertion problem: both package typechecks pass locally, along with 33 focused web tests and 123 focused CLI tests. However, one merge-blocking provider/credential isolation issue remains.

`resolveLLMConfig` scopes saved `baseUrl`, `model`, and reasoning settings through `reuseSavedHttpConfig`, but `apiKey` and `apiVersion` bypass that provider-match guard. Because an empty config now defaults to `minimax`, a machine with only `GITNEXUS_API_KEY` or `OPENAI_API_KEY` can silently send that credential to the MiniMax endpoint. The provider-switch persistence path is also asymmetric: `wiki.ts` resets HTTP fields only when switching into MiniMax. Switching from MiniMax to another provider can retain MiniMax's `baseUrl` and key under the new provider name, so subsequent prompt/repository content can still be sent to MiniMax despite the explicit provider switch.

Please gate saved credentials/version by the resolved provider and reset provider-specific HTTP fields symmetrically on every provider change. Add regressions for an empty config with only an OpenAI/GitNexus key and for MiniMax?OpenAI plus OpenAI?MiniMax switches with saved credentials.

GitHub currently exposes no functional CI result for this head; only the unrelated Vercel authorization failure is visible. Recommendation: request changes until credential isolation and provider-switch tests are added.

kapelame:
Thanks for the review. I isolated saved credentials and API versions by provider, reset provider-specific HTTP fields symmetrically on provider changes, and added regressions for cross-provider credential leakage. I ran git diff --check; the focused Vitest command could not start because vitest is not installed in the checkout.

koriyoshi2041 (changes requested):
The credential-isolation direction is right, but the non-interactive provider switch now clears the explicit replacement values from the same invocation. `wikiCommandImpl` copies `--api-key`, `--base-url`, `--model`, `--api-version`, and `--reasoning-model` into `updates` first, then `if (providerChanged)` sets those fields back to `undefined`. For example, switching with `--provider openai --api-key new-key --model gpt-5` persists neither replacement. Clear inherited fields before applying option overrides (or only clear fields whose corresponding option is absent), and add a provider-switch regression with explicit replacement credentials/model.

koriyoshi2041 (changes requested):
The credential isolation fix introduces a provider-switch regression in the CLI override path.

In `wikiCommandImpl`, explicit overrides are assigned to `updates` first, then `providerChanged` clears `apiKey`, `baseUrl`, `apiVersion`, and `isReasoningModel`. Unlike `model`, those values are not re-applied afterwards. A normal one-shot switch such as:

```
gitnexus wiki . --provider openai --api-key sk-new --base-url https://api.openai.com/v1
```

therefore saves the new provider but discards the new key and base URL. The same happens to an explicit `--api-version` or reasoning flag. The subsequent `resolveLLMConfig(options)` can make the current invocation appear to work, while the persisted config is incomplete and the next invocation fails or uses defaults.

Please clear inherited provider-specific fields before applying explicit CLI overrides (or only clear each field when the corresponding option is absent), and add a regression that switches provider while supplying a new key/base URL in the same command.

I also ran the current head `4d942233` focused suites after building `gitnexus-shared`: `npx vitest run test/unit/wiki-llm-client.test.ts test/unit/wiki-flags.test.ts` passed 125/125. The existing tests cover isolation in `resolveLLMConfig`, but not this persistence ordering case.

azizur100389:
Rechecked current head `4d942233` against current main `f2717c6a`. The branch now conflicts in `gitnexus-web/src/components/SettingsPanel.tsx`, and CI Gate is red from the Windows `fts-description-search` setup timeout.

One functional blocker also remains in `wiki.ts`: explicit `apiKey`, `baseUrl`, `apiVersion`, and reasoning settings are assigned to `updates`, then immediately cleared when `providerChanged` is true. The current invocation still receives the CLI overrides, but the command reports configuration saved while persisting incomplete settings.

Please clear inherited provider fields first, then apply explicit new-provider overrides, and add `wikiCommand`/`saveCLIConfig` regression tests for both provider-switch directions. The default-MiniMax test should also mock an empty saved configuration instead of inheriting `~/.gitnexus/config.json`.

Please resolve the main conflict and rerun CI before re-review.

octo-patch:
Thanks for the review. I resolved the SettingsPanel conflict against current main, preserved explicit replacement settings across provider switches, and added regressions for both directions. I ran 127 focused CLI tests, 13 focused web tests, the web TypeScript build, and the shared package build; the previous Windows failure was an unrelated setup timeout, and CI is rerunning on this head.

octo-patch:
Thanks for the review. I migrated unsupported legacy MiniMax model settings to the current MiniMax-M3 default so old MiniMax-M2.5 configurations cannot inherit adaptive thinking, and added a loadSettings-to-createChatModel regression. I pushed commit 05ad62f and ran 34 focused web tests, the web production build, Prettier, git diff --check, and 127 focused CLI tests.

Issue #2416, opened 2026-07-09
feat(spring): model AOP transactions caching and security


magyargergo:
Parent tracking issue: #1225

Status: remaining work

Goal
Make Spring proxy and advice behavior visible in analysis results.

Scope
Detect common advice related annotations such as Transactional, Cacheable, CacheEvict, PreAuthorize, Secured, Aspect, Around, Before, After, and related method level behavior.

Acceptance criteria
1. Methods with transactional, caching, or security behavior show that extra framework behavior exists.
2. Aspect advice can be connected to likely advised methods when the pointcut can be understood statically.
3. Unknown pointcuts are represented conservatively.
4. Context and impact output can surface that a method is advised or proxied.
5. Tests cover annotation based method behavior and at least one explicit aspect case.

Issue #2545, opened 2026-07-18
TypeScript global fetch() incorrectly resolves to repository fetch method


damiandelmas:
## Summary

In TypeScript, an unqualified call to the platform-global `fetch()` is incorrectly resolved to an unrelated repository method named `fetch`. The resulting `CALLS` edge uses the repository method's exact ID and is labeled `local-call` with confidence `0.85`.

This is especially visible in Cloudflare Worker code, where outbound HTTP helpers and the Worker's inbound request handler naturally coexist.

## Minimal reproduction

`worker.ts`:

```ts
export async function callExternal(): Promise<Response> {
  return fetch("https://example.com/api");
}

export default {
  async fetch(_request: Request): Promise<Response> {
    return new Response("ok");
  },
};
```

Index it:

```sh
npx -y gitnexus@1.6.9 analyze . --index-only --force
```

Inspect call edges:

```cypher
MATCH (a)-[r:CodeRelation]->(b)
WHERE r.type = "CALLS"
RETURN a.id AS caller_id,
       b.id AS callee_id,
       r.confidence AS confidence,
       r.reason AS reason
```

Observed:

| caller_id | callee_id | confidence | reason |
| --- | --- | --- | --- |
| `Function:worker.ts:callExternal` | `Method:worker.ts:fetch#1` | `0.85` | `local-call` |

The source call is the platform-global HTTP function. It does not invoke the Worker's request-handler method.

## Expected behavior

GitNexus should either:

- recognize `fetch` as a platform/global built-in and leave it unbound to repository definitions, or
- represent it as unresolved/external rather than emitting a repository `CALLS` edge.

## Why it matters

The false edge is not confined to direct callers. Downstream `impact` traversal and inferred processes can propagate through the incorrectly selected request-handler hub, producing plausible but false execution paths and inflated blast-radius results. Because the edge is labeled `local-call`, consumers cannot safely distinguish it using `reason` or the current confidence value.

In a larger Worker corpus, every outbound provider helper using global `fetch()` exhibited the same resolution, so this appears systematic rather than an isolated parser edge case.

## Environment

- GitNexus: `1.6.9`
- Node: `v22.23.0`
- OS: Linux / WSL2
- Language: TypeScript

magyargergo:
Thanks for reporting this issue! I'm looking into this now.

PR #2782, opened 2026-08-01
fix(resolution): resolve Go pointer-receiver calls, and report the program boundary instead of hedging (#2766)

### LINKED ISSUE #2766: Go: cross-package method calls on a struct-field receiver emit no CALLS edge (interface- and concrete-typed alike); impact() returns 0 with epistemic "exact"
> **Corrected 2026-07-30.** My original report blamed implicit interface satisfaction / missing `IMPLEMENTS` edges. **That diagnosis was wrong** and the repro under it was unrepresentative. Same-package interface dispatch resolves fine. The real variable is *cross-package method calls on a struct-field receiver*, which fail for **concrete-typed fields too**. Full corrected analysis below; original text kept at the bottom for history.

## Summary

On `gitnexus@1.6.9`, a Go method call whose receiver is a **struct field** emits no `CALLS` edge when the callee's type lives in **another package**. This holds whether the field is interface-typed or concrete-typed.

Calls resolve only when the receiver is a **local variable** whose concrete type is established in the same function (constructor result or type assertion). Package-level function calls resolve normally.

Consequence: in an idiomatic Go service with constructor-injected dependencies stored as struct fields, the `handler → service → repo` call graph does not connect. `impact` returns `impactedCount: 0` with `"epistemic": "exact"` on methods that have live production callers.

## Evidence matrix

Measured on a private ~470k LOC Go codebase (2,821 files; 39,185 nodes / 158,131 edges; indexed in 37.5s), cross-checked against minimal fixtures.

| Receiver | Type | Package | `CALLS` edge |
|---|---|---|---|
| `h.thing` (field) | interface | same | ✅ resolves |
| `h.repo` (field) | interface `repository.OrderRepository` | cross | ❌ |
| `s.orderRepo` (field) | interface `repository.OrderRepository` | cross | ❌ |
| `s.cartRepo` (field) | **concrete** `*repository.CartRepo` | cross | ❌ |
| `concreteCartRepo` (local, via type assertion) | concrete | cross | ✅ resolves |
| `txOrderRepo` (local, via `WithTx(tx)`) | concrete | cross | ✅ resolves |
| `repository.PlainFunc(ctx)` | package-level func | cross | ✅ resolves |

The concrete-field row is the key one — it rules out interfaces as the cause.

## Worked example from the real codebase

Two call sites in the *same file*, `internal/services/pick_service.go`:

```go
// field-typed receiver, line 718 — NO edge emitted
// (cartRepo is declared at :103 as *repository.CartRepo — CONCRETE)
txCartRepo := s.cartRepo.WithTx(tx)
```

```go
// local variable via type assertion, wave_service.go:935 — edge IS emitted
concreteCartRepo, cartCanTx := s.cartRepo.(*repository.CartRepo)
...
txRepo := concreteCartRepo.WithTx(tx)
```

Querying the enclosing `StartSession` function confirms it: edges exist to `OrderRepo.WithTx`, `PickRepo.WithTx`, `WaveRepo.GetByID` (all reached via locals), but **no edge to `CartRepo.WithTx`**, whose only call in that function goes through the field.

## Effect on `impact`

A repository method with two production callers:

```
internal/handlers/orders.go:1792     h.repo.DeleteItem(r.Context(), itemID, orgID)
internal/services/shopify_sync.go:1462   s.orderRepo.DeleteItem(ctx, itemID, orgID)
```

```
gitnexus impact DeleteItem --file internal/repository/order_repo_items.go --depth 5
→ { "impactedCount": 0, "risk": "LOW", "epistemic": "exact" }
```

With `--include-tests` it returns exactly 1 — an integration test in `internal/repository/`, i.e. the **same package**. Every cross-package caller is missing.

Scale in this codebase: 178 `h.repo.*` handler call sites and 585 `s.*Repo.*` service call sites, essentially all field-receiver calls into `internal/repository`.

Cross-package `CALLS` edges into `internal/repository` do exist (130 `Method`, 118 `Struct`, 74 `Function`) — the `Method` ones trace back to local-variable receivers, and the `Struct` ones are composite-literal constructions.

## Repro

The minimal fixtures I built reproduce the *cross-package field* failure but are unreliable in the other direction — a two-package fixture with a single concrete field failed to emit any `CALLS` edge at all, where the equivalent real-world pattern sometimes succeeds. So the matrix above is grounded in the real codebase; treat the fixture as a starting point rather than a faithful minimization.

```go
// repository/repo.go
package repository

import "context"

type Thing interface{ DoWork(ctx context.Context) error }

type Impl struct{}

func (i *Impl) DoWork(ctx context.Context) error { return nil }
```

```go
// handlers/handler.go
package handlers

import (
	"context"

	"mod/repository"
)

type Holder struct {
	thing repository.Thing // swap for `impl *repository.Impl` — same result
}

func (h *Holder) Run(ctx context.Context) error {
	return h.thing.DoWork(ctx) // no CALLS edge
}
```

```bash
gitnexus analyze --index-only
gitnexus cypher "MATCH ()-[r:CodeRelation]->() WHERE r.type='CALLS' RETURN count(*)"
# → 0
```

Moving the same code into one package makes the edge appear.

## Levers that don't help

| Lever | Result |
|---|---|
| `--pdg` | Adds `CFG` edges only |
| `--include-tests` | Result filter; surfaces only the same-package test caller |
| `--depth` | Nothing to traverse — the edges are absent |

## Note on epistemics

**#1858** ("flag `impact()`/`context()` when unresolved indirection is on the path, instead of reporting `impactedCount: 0`") is closed, but 1.6.9 still returns `"epistemic": "exact"` next to `impactedCount: 0` here. A confidently-exact zero on a symbol with live callers reads as "nothing depends on this" immediately before a change ships — worse than a flagged unknown.

Credit where due: `context UnsplitOrder` correctly returned `status: "ambiguous"` with 7 candidates rather than guessing, so the machinery for honest uncertainty already exists.

## Environment

- `gitnexus` 1.6.9 (`npm i -g gitnexus@latest`)
- Node v22.22.3 · Darwin arm64 (Apple Silicon)
- `GITNEXUS_SKIP_OPTIONAL_GRAMMARS=1` at install (Dart/Proto/Swift/Kotlin; not relevant to Go)
- Indexed with `--index-only`

---

<details>
<summary>Original (incorrect) report — kept for history</summary>

The original text attributed this to Go implicit interface satisfaction emitting no `IMPLEMENTS` edge, citing 19 `IMPLEMENTS` edges against 398 interface method declarations (18 of the 19 being test mocks that explicitly embed the interface).

That observation about `IMPLEMENTS` edges is accurate as a standalone fact, but it is **not** the cause of the missing `CALLS` edges. A same-package fixture — one interface, one implementation, no embedding — resolves `h.thing.DoWork()` straight through to `Impl.DoWork` with no `IMPLEMENTS` edge present at all. Interface satisfaction is therefore not on the critical path for call resolution, and the package boundary is what actually matters.

</details>

magyargergo:
Fixes #2766.

Two outcomes, one of which was not in the original scope and turned out to matter more.

**Go pointer-receiver calls now resolve.** A method with a pointer receiver bound its receiver to the literal string `*Host`, which matched no class, so receiver typing declined at the *base* and every `h.field.Method()` in the dominant Go idiom emitted no `CALLS` edge. On the issue's reproduction the graph goes from **4 `CALLS` edges to 10**, and `impact` on the repository method goes from `impactedCount: 0` with a caveat to **`impactedCount: 1`, `epistemic: exact`**.

**`impact` stops hedging about calls that leave the program.** Three quarters of what the analyzer reported as uncertainty was `System.out.println`, `fetch(...)`, `os.environ` — calls into code that isn't indexed, where no edge could exist. That is a boundary, not a doubt, and it is now reported as one.

---

## The diagnosis in the issue was wrong, including mine

The report blamed cross-package field types. My first comment on the issue blamed the same thing. Both were wrong: that experiment varied receiver decoration and field decoration together, so it could not separate them. Isolating them:

| Receiver | Field | `CALLS` |
|---|---|---|
| value | value | resolves |
| value | **pointer** | resolves |
| **pointer** | value | **none** |

Only the pointer *receiver* fails. Go already normalizes field type bindings at capture, so the step lookup was never the problem. The receiver self-binding is the one Go binding deliberately left decorated — `method-owners.ts` reads `*T` vs `T` to model value and pointer method sets — so this normalizes at **lookup**, never by rewriting the stored binding.

## Three quarters of the hedge was the program boundary

Dumping all 102 call drops with source context:

| Origin | Count | Is anything lost? |
|---|---|---|
| `external` | **76** | **No.** The callee is not in the graph. |
| `in-program` | 20 | In principle — but see below. |
| `unknown` | 6 | Yes. Casts, ternaries, `globalThis.x ??= []`. |

A compiler resolves `System.out.println` against the JDK. Lacking one, the honest statement is *"this call leaves the analyzed program"*, not *"this analysis is incomplete"*. Collapsing the two is what made the signal fire on essentially every real codebase, which is what teaches readers to ignore it.

Origin is decided by the receiver base's **declared type**, not its name — a first cut marked `inputs.stream()` in-program because `inputs` is a local, but its type `List<String>` is JDK, so the target is external. That correction moved 28 sites.

The split is **persisted, not filtered**: `externalCounts` / `externalSites` sit beside the hedging counts and `impact` surfaces `externalBoundary` as a third cause, so the artifact distinguishes "clean index" from "76 drops we judged external", and the heuristic stays reversible without a re-index.

The remaining 20 in-program drops are mostly not defects either — `user.Address.Save()` resolves cleanly in isolation (its own fixture emits both edges with zero drops) and drops only because the corpus is ~200 mini-projects sharing names, with **55 files defining `Address`**. The genuinely untypeable population is the 6 `unknown`, now visible instead of buried.

## The most valuable change was not planned

`Case 0` — the compound-receiver case — only ran when the receiver *text* contained `.` or `(`. A subscript receiver like `repos[0].save()` contains neither, so the case never fired, the fold was never consulted, **and no drop was recorded either**: those calls vanished with the instrument blind to them. The bench's own "known blind spots" section documented it.

The gate now keys on whether the capture layer minted a receiver chain — a structural fact from the real AST rather than a punctuation test. **13 shapes moved INVISIBLE-GAP → VISIBLE-GAP.** For an epistemic signal that matters more than any single resolution fix: a hedge cannot be honest about losses it cannot count.

## A duplicate-edge bug this branch shipped, and a claim retracted

An earlier commit here claimed that typing the receiver made a stray `ACCESSES` edge "correctly retarget to the property". **That was wrong**, and the test asserting it passed by accident — the row it checks has a pointer receiver whose lookup failed for an unrelated reason.

Go's `@reference.read` pattern matches every `selector_expression`, so `h.dep.Work()` minted **three** sites: the call, the genuine `h.dep` field read, and a **phantom** read on the callee. The phantom resolved through method-preferring lookup and emitted an `ACCESSES` duplicating the `CALLS` at the same position — present on the first push of this branch:

```
ACCESSES  RunFromValueReceiver -> DoWork:Method
ACCESSES  RunLocal             -> DoWork:Method
```

A selector in **function position** is never a read; dropped at the emitter. A method *value* (`f := h.dep.Work`) is untouched. Backed now by an exact-set assertion over the whole fixture.

## Measurement

The shape arm went from 3 languages with ad-hoc shape lists to a canonical 10-shape axis across all 14, with `N/A` requiring a reason and `GRAMMAR-UNAVAILABLE` for an unloadable parser. `assertMatrixComplete` fails the run on a missing cell.

**RESOLVES 43 → 55**, VISIBLE-GAP 21 → 23, INVISIBLE-GAP 31 → 18. Await now resolves in TypeScript, Python, C#, Kotlin and Dart; subscript in TypeScript, C++, Go, Python, Rust and Kotlin.

Drops are censused by receiver shape *and* origin, both gated. That answered an open question: `.java`'s 49 drops are **not** one defect (30 field-chain / 14 call-chain / 5 no-chain).

## Three negative results, committed with their instruments

- **The depth cap does not limit resolution.** A 4-step chain mints *nothing* at cap 3 yet still resolves, because the text cascade runs to its own depth of 8. Raising it changes zero edges. Left at 3; the `fourHopChain` fixture is committed so the next person inherits the measurement.
- **No drop ratchet is needed.** `--check` already asserts exact match, strictly stronger than a ratchet.
- **The shared-registry receiver miss is real but not load-bearing.** `lookup-core.ts` genuinely fails for every Go pointer receiver, but a strip-and-retry changes zero edges — the emit pass resolves those sites first. Measured, not shipped.

## Known residuals

- **`A1` rejected on its merits.** Normalizing Go's self-binding at capture (deleting the decoration hook) was implemented twice and rejected after a three-agent review. The decisive argument is structural: the stripper path **refuses** when a stripped name binds ≥2 classes, and that refusal needs the candidate count *at the lookup site*, which capture time cannot know. It also moves pointer-ness to an optional field failing toward `'value'` — the unsafe direction, since a `*T` method is not in `T`'s method set, so a misread manufactures false-positive `IMPLEMENTS` edges.
- **The same phantom read pattern is live in Rust, C and C++** (`rust/query.ts:217`, `cpp:614`, `c:182` — none filtered). Wants a shared `isInCallPosition(node, callNodeTypes)` helper driven by the per-language set all 14 languages already declare.
- **The `external` terminal fallback is inverted** — "found nothing" is recorded as "definitely outside the program". The naive flip would reclassify all 76 and undo the feature; the correct fix makes `external` a *positive* determination via the existing `LanguageProvider.isBuiltInName` hook.
- **Real type resolution for the 6 `unknown` drops** — a cast *gives* you the type; a ternary needs a branch join. Now visible and well-scoped.
- **Stdlib type stubs** (JDK / BCL / `lib.d.ts`) would make external targets resolvable at all. A program of work; this PR makes the boundary measured rather than guessed.
- **C# `indexElement`** still gaps for a cause I did not diagnose.
- **A latent ordering bug:** the empty-receiver-text guard precedes the structural fold, so a receiver with known structure is discarded for having empty text. No input triggers it today.

## Schema versions — re-index required

`SCHEMA_BUMP` 34 → **38**, `INCREMENTAL_SCHEMA_VERSION` 28 → **34**. The receiver-chain wire format moved v1 → v2 and a v2 decoder refuses a v1 payload by design, so existing indexes are stale and the bumps enforce that rather than letting a warm cache serve chains this build discards.

`main` moved these three times while this branch was in flight — **the seventh collision in this series, and the first that was an exact clash rather than a near-miss.** #2416 landed on `37`/`33`, precisely the numbers this branch already held: two incompatible schemas both claiming the same version, so an index built by either side would satisfy the other's freshness check while carrying content it cannot read. Renumbered to `38`/`34` with both changelog notes merged rather than one replacing the other, and the reuse-gate test now asserts `v33` **fails** the gate. **Re-check both against `origin/main` immediately before merge** — this is the single most reliable source of conflict on this branch.

## Verification

- Resolver integration + scope-resolution unit: **4424 tests green**
- **All eight benches CI runs** pass — receiver-resolution, scope-capture, python-scope (×2), callable-value-flow, scope-emission, cfg, emit-persistence
- Regression test proven to fail without the fix: with the stripper disabled the 3 pointer-receiver assertions fail while both controls still pass
- Rebased onto `origin/main` twice, conflicts resolved preserving main's changelog entries both times; re-verified *after* each rebase, not before — the second rebase brought Spring AOP (#2416), which collided on both schema constants
- Every baseline movement carries a documented reason

Analyzer change with no runtime service surface, so monitoring is index-shaped.

- **Watch:** `CALLS` edge counts on re-index of Go-heavy repositories — expect an increase concentrated in pointer-receiver methods. `impact` results that returned `impactedCount: 0` with a caveat should return real counts.
- **Healthy signal:** `bench/receiver-resolution/measure.mjs --check` passes; `callDrops` 102 on the fixture corpus, split 76 external / 20 in-program / 6 unknown.
- **Failure signal:** a rise in `callDrops` without a baseline update, or `epistemic: exact` on a symbol with a known-missing caller.
- **Rollback trigger:** any `CALLS` edge whose *target* moved, as opposed to edges added. This work should only add edges; a moved target is a wrong-edge bug and warrants revert.
- **Window/owner:** first re-index of a large Go or Java repository after merge.


gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
```diff
@@ -152,6 +157,56 @@ function resolveClassBindingForName(
...
+      undefined
+      ? 'in-program'
+      : 'external';
+  }
+  // A type the program declares, used as a static receiver.
+  if (findClassBindingInScope(inScope, base, scopes) !== undefined) return 'in-program';
+  // Anything else this index knows by that name (namespace, module, free fn).
+  if (scopes.qualifiedNames.has(base)) return 'in-program';
+
+  return 'external';
```
magyargergo: **P1 — blocking.** This terminal `return 'external'` treats *absent* type evidence as *proof* of externality, and it removes real losses from the hedge.

The docstring 40 lines up says *"A local, a parameter, a field, a class, or any qualified name counts as in-program"* (`:163-166`), but the function only ever asks whether the base's declared **type** resolves to a class. `scope.bindings` / value bindings are never consulted, so a base the program demonstrably declares falls through to here.

**Evidence (executed against this worktree):**
```
GO   h typeBinding rawName = "*Host"
GO   findClassBindingInScope(*Host) [no stripper] = undefined
GO   classifyReceiverOrigin("h")  => external      # Host IS in the index
TS   classifyReceiverOrigin("svc") => external     # this PR's own drop fixture
```
Two independent triggers:
1. **Go pointer receivers** — `resolveClassBindingForName` (`:117-158`) calls `findClassBindingInScope` with **three** arguments and never receives `stripTypePreservingDecoration`, the stripper this PR adds and threads into the other three lookups (`compound-receiver.ts:351, 513, 647`). `func (h *Host)` binds `h` to the literal `*Host`, the lookup misses, and every unrescued drop rooted at `h` is filed as leaving the program.
2. **The PR's own new fixture** — `typescript.test.ts:3347` `droppedCall(svc)` is an unannotated parameter: no type binding, not a class, not a qualified name ⇒ `external`. Same for a type alias, a generic parameter, and a call-result alias (all executed).

`unresolved-receivers.ts:92-96` then `continue`s past `totalSites++`, and `local-backend.ts:523` maps `notes.length === 0` to `epistemic: 'exact'` with no `boundaries` prose. An agent following this repo's own CLAUDE.md impact gate proceeds on a count it has been told is complete while real in-program callers are missing — strictly worse than the pre-PR hedge, which at least fired. `baseline.json` now pins `callDropsByOrigin.external: 76` under exact match, freezing the misclassification into the gate.

**Fix:** make `external` a positive determination — (a) check value/parameter bindings before the fallthrough, (b) pass `stripTypePreservingDecoration` into `resolveClassBindingForName` here, (c) default the rest to `'unknown'`, which already counts toward the hedge. Until then, don't deduct `externalSites` from `totalSites`.

Reached independently by Claude ce-correctness, ce-adversarial, risk-architect, ce-testing and facts-historian, **and** by Codex B1 (R2, P1, blocks merge) and Codex B3 (`[high]`) — both Codex legs pinned to the `droppedCall(svc)` fixture. The PR description already concedes this fallback is inverted.

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -362,15 +431,96 @@ export function foldReceiverChain(
...
+          options.stripTypePreservingDecoration,
+        );
+        if (elementClass === undefined) return undefined;
+        current = { def: elementClass, declaredTypeName: element, declaredAtScope: scopeForLookup };
+      } else if (current.def === undefined) {
+        // No unwrap available and the position never named a class: nothing to
+        // fold on. Decline rather than continue against a stale owner.
+        return undefined;
+      }
+      continue;
```
magyargergo, line 517: **P1 — blocking.** This `continue` advances the fold with `current` still pointing at the **container**, so a member is looked up on the wrong owner.

The comment above justifies it with *"The declared type here is therefore already the awaited / element type"*. That holds when the base *is* a container. It is false when the declared type is an ordinary class that happens to be subscriptable (C# indexer, Python `__getitem__`, TS index signature, C++ `operator[]`, Kotlin `operator fun get`), and false when capture reduced more layers than the source had indexes.

**This is new damage, not an inherited gap.** `SUBSCRIPT_NODE_TYPES` has **zero occurrences at the diff base**, so no chain was minted for a subscript receiver; and `grid[0]` contains neither `.` nor `(`, so Case 0's old punctuation gate never fired. Pre-PR these sites emitted **nothing and recorded nothing**. The widened gate at `receiver-bound-calls.ts:476` now admits them and Case 0 emits at 0.85 confidence against the wrong owner — an invisible gap converted into a confident wrong edge, inverting this PR's own stated criterion.

**Evidence (executed against this worktree):**
```
f5 b[0].save()  (Box has save() + [k:number]: User)  -> Box       ← WRONG owner
f7 s[0].save()  (Service, NOT a container)           -> Service   ← identity on a non-container
   typeBindings :: grid -> "User"     # `User[][]` reduced to `User` at capture
```
Three distinct triggers across two engines: TS `class Grid { run(); [i:number]: Item }` → `Grid.run` (Claude ce-correctness); C# `class Table { public Row this[int i]; void Render() }` → `Table.Render`, plus Python and TS variants (Claude ce-adversarial); and TS `repos: User[][]` where capture strips to a fixed point, so **one** index folds to `User` and `repos[0].map(…)` — validly `Array.map` — can emit `CALLS → User.map` (Codex B1 R1, P1, 99%, blocks merge).

Only TypeScript and C# implement `elementTypeOf`, so every index step in the other twelve languages is pure identity (Go and Java are *accidentally* safe — verified — but that is luck, not design). The bench cannot catch this: `measure.mjs --check` is exact-match on drop counts and edge *presence*, and this failure mode is an edge that **appears**, reading as a resolution win.

**Fix:** stop inferring the regime at the fold. Make `elementTypeOf` mandatory for a language to participate in the `index` step (capture-normalizing languages implement it as "already reduced ⇒ return the name unchanged"), or add an explicit `containerReducedAtCapture` provider signal so identity is legal only when set. Deleting the identity branch is the conservative option but costs most of the newly-gained edges.

gitnexus/src/core/ingestion/languages/go/captures.ts
```diff
@@ -183,6 +197,26 @@ export function emitGoScopeCaptures(
...
+    // over fields, so it emits an ACCESSES edge to the METHOD duplicating the
+    // CALLS edge at the same position. Visible today on any Go receiver the text
+    // cascade can type: `v.impl.DoWork()` emits both `CALLS -> DoWork` and
+    // `ACCESSES -> DoWork`. It stayed hidden on pointer receivers only because
+    // their lookup failed for an unrelated reason.
+    //
+    // A selector in FUNCTION position is never a read. A genuine method value
+    // (`f := h.dep.Work`) is not in function position and is untouched.
+    const readNode = nodeMap['@reference.read'];
+    if (readNode !== undefined && isCalleeOfMemberCall(readNode)) continue;
```
magyargergo: **P2.** *"A selector in FUNCTION position is never a read"* is false in Go when the selected member is a **func-typed struct field**: `h.dep.Work()` where `Work func() error` is a genuine field read followed by an indirect call. `isCalleeOfMemberCall` only tests whether the selector is the `function` field of a `call_expression`; it cannot tell a method from a func-valued field.

**Evidence (capture dump executed on a fixture with `type Dep struct { Work func() error }`):** for `CallFuncField` only `@reference.read="h.dep"` survives — the `Work` read is gone. The controls behave correctly: `f := h.dep.Work` (method value) and `h.dep.Name` (plain field) both keep their reads.

Callback structs, hook structs and hand-rolled mocks (`mock.DoFunc`, handler structs, `opts.OnEvent`) are ordinary Go, and this deletes their only ACCESSES evidence. Missing edge, not a wrong one — hence P2 — but `go.test.ts`'s new exact-set ACCESSES assertion is a good invariant that **cannot** catch a widening of this drop, because the fixture contains no func-typed field.

**Fix:** gate the drop on the selector's tail resolving to a *method* rather than a field, or keep the read and suppress the duplicate at emit time by target kind. Cheaper stopgap: add a func-typed-field row to the fixture so the trade is pinned deliberately.

Claude ce-correctness (F7) and ce-adversarial (P2-1) independently; not reached by Codex (single-engine).

gitnexus/src/core/ingestion/scope-resolution/contract/scope-resolver.ts
```diff
@@ -1114,6 +1126,59 @@ export interface ScopeResolver {
...
+   * shaped `findClassBindingInScope(...) ?? otherResolver(...)` and would have
+   * their fallback suppressed by a global widening.
+   *
+   * Leave undefined for languages whose declared types carry no type-preserving
+   * decoration. Measured: only Go needs it for a receiver base; Rust, C#, Swift,
+   * TypeScript and C++ need it for field types.
+   */
+  readonly stripTypePreservingDecoration?: DecorationStripper;
+
+  /**
```
magyargergo: **P2.** This 20-line block ("Unwrap a COLLECTION spelling to its element type — `User[]` -> `User`…") ends at 1180, line 1181 is blank, and 1182 begins a *different* JSDoc for `stripReceiverCastExpressions`. **No `readonly <name>?: …` declaration follows it.** Verified by reading the file at HEAD.

It documents the interim `unwrapCollectionElement` hook that was folded into `elementTypeOf` (commit `5a89747c9` deleted the declaration line and left the comment). Inside the file that *is* the language-plugin contract, it reads as an authoritative spec for a member that does not exist, and it will mislead the next resolver author into implementing a hook the core never calls.

**Fix:** delete `1162-1180` — its content is already substantially restated in `elementTypeOf`'s own docblock at 714-738 — or add the missing declaration if a second hook is genuinely intended.

Flagged by the coordinator pre-read and by four Claude lanes (docs-dod F3, ce-api-contract #6, ce-maintainability #3, ce-project-standards F2a).

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
```diff
@@ -320,7 +383,13 @@ function typeOfMemberOnClass(
         if (curScope === undefined) break;
         const hoisted = curScope.typeBindings.get(memberName);
         if (hoisted !== undefined) {
-          return findClassBindingInScope(hoisted.declaredAtScope, hoisted.rawName, scopes);
+          const def = findClassBindingInScope(hoisted.declaredAtScope, hoisted.rawName, scopes);
+          if (def === undefined) return undefined;
```
magyargergo: **P3.** The primary branch 30 lines up deliberately carries the declared type forward on a class miss (`:357-364`, *"Returning `undefined` here would strand those shapes"*), because an unwrapping step is exactly how `Promise<User>` / `[]Repo` become resolvable. This hoisted branch does the opposite and returns `undefined`.

**Evidence (executed):** same declared type, two routes, two answers —
```
fold 2|byId|i        -> Item        # variable typed Record<string, Item>  → BASE path
fold 2|svc|cgetMap|i -> undefined   # method returning Record<string, Item> → HOISTED path
```
So `svc.getMap()['k'].run()` declines and records a drop while the byte-identical `byId['k'].run()` resolves. `hoistTypeBindingsToModule: true` is set by ten languages, so this is the common route for a method return type, not an edge case. Missing edge, not a wrong one.

Separately, this call omits `options.stripTypePreservingDecoration` while its three siblings (`:351`, `:513`, `:647`) pass it. Inert today — Go is the only implementer and Go return types already go through `normalizeGoReturnType` — but it goes live the moment a second language adopts the hook, which the contract at `:1156-1158` names as planned for Rust, C#, Swift, TS and C++.

**Fix:** `if (def === undefined) return { def: undefined, declaredTypeName: hoisted.rawName, declaredAtScope: hoisted.declaredAtScope };` and pass the stripper as the 4th argument.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -5829,7 +5894,15 @@ export class LocalBackend {
         }
       }
       if (boundaries.length === 0) return epistemicFrom(droppedBoundaries);
-      return { epistemic: 'lower-bound', boundaries: [...droppedBoundaries, ...boundaries] };
+      return {
+        epistemic: 'lower-bound',
+        boundaries: [...droppedBoundaries.notes, ...boundaries],
+        causes: {
+          receiverTyping: droppedBoundaries.sites,
+          dispatchBoundary: boundaries.length,
```
magyargergo: **P3.** `receiverTyping: droppedBoundaries.sites` counts call **sites**; `dispatchBoundary: boundaries.length` counts **notes** — one per boundary interface. The PR's own justification 5,370 lines up insists on the opposite for the sibling field:

> *"SITES, not notes. There is one note per symbol name but it reports N dropped sites, so counting notes would have published `1` next to prose saying `2 call sites` — a consumer branching on the number would read a different magnitude than the human reading the text."* (`:533-536`)

The stated consumer — *"a coding agent gating its own edits on this result"* — reading `{receiverTyping: 12, dispatchBoundary: 1}` concludes receiver typing dominates, when that single note may describe an interface with 40 implementations and hundreds of untraced callers. Diagnostic-only today (nothing in-repo compares the two), which is why it is P3 — but the field exists specifically to be branched on.

**Fix:** either count implementations + consumers so both fields are "untraced call sites", or rename to `dispatchBoundaryNotes` and document the unit on each field.

Coordinator pre-read + Claude ce-correctness F5 + ce-maintainability #1.

gitnexus/src/core/ingestion/scope-resolution/resolution-outcome.ts
```diff
@@ -61,6 +62,120 @@ export type ResolutionOutcome =
...
+
+/** Classify a dropped receiver from its encoded chain. `undefined` chain ⇒
+ *  `no-chain`; an undecodable one is also `no-chain`, since what we know about
+ *  it is exactly that no usable structure survived.
+ *
+ *  Takes `DecodedReceiverChain` rather than a structural duck-type: widening
+ *  `kind` to `string` let `await` and `index` fall into an `else` branch and be
+ *  counted as FIELDS, so the two shapes this work exists to expose were
+ *  censused as `chain-field`. The discriminated union makes a new step kind a
+ *  compile error instead of a silent bucket. */
```
magyargergo: **P3.** Layout at HEAD: this block (135-143) documents `classifyReceiverShape`; a *second* JSDoc block for `ReceiverOrigin` follows at 144-151; `export type ReceiverOrigin` at 152; `export function classifyReceiverShape` at **154 with no doc of its own**. Because that second block intervenes before any declaration, this one **floats unattached** — `ReceiverOrigin` keeps its own correct doc, and the exported function ships undocumented.

The rationale that matters most is the one that gets lost: *"widening `kind` to `string` let `await` and `index` fall into an `else` branch and be counted as FIELDS… The discriminated union makes a new step kind a compile error instead of a silent bucket."* That is the design decision this PR most wants preserved, and it is currently attached to the wrong symbol.

**Fix:** move 135-143 to sit directly above line 154.

Coordinator pre-read + four Claude lanes. Part of a broader pattern (~10 sites across 8 files where a new declaration was inserted beneath an existing docblock) — this is the highest-value instance.

gitnexus/src/core/ingestion/scope-resolution/unresolved-receivers.ts
```diff
@@ -56,10 +73,33 @@ export function summarizeUnresolvedReceivers(
...
+    // `fetch(...)`, `os.environ.setdefault(...)` reach code this index does not
+    // contain, so there is no node an edge could have pointed at and nothing was
+    // lost. Counting them made `impact` report a lower bound on essentially
+    // every real codebase — 75% of the drops on this corpus — which is what
+    // taught readers to ignore the signal.
+    //
+    // A compiler resolves these against the JDK / BCL / lib.d.ts. Lacking those,
+    // the honest statement is "this call leaves the analyzed program", not "this
+    // analysis is incomplete". `unknown` still counts: assuming completeness we
+    // cannot demonstrate is the unsafe direction.
```
magyargergo: **P3.** Two consecutive blocks (76-86 and 87-91) explain the same routing decision, and the `unknown` treatment is stated twice in near-identical wording ("`unknown` still counts: assuming completeness we cannot demonstrate is the unsafe direction" / "`unknown` counts with `in-program`: assuming a completeness we cannot demonstrate is the unsafe direction"). No behavioural contradiction — both point at line 92 — but it reads as an incomplete editorial substitution.

**Fix:** keep 87-91 and delete 76-86, or merge. Note 87-91 carries a point 76-86 lacks (*"they stay in the artifact under their own key so the split is auditable and reversible"*), so deleting the wrong block loses information.

Coordinator pre-read + Claude docs-dod F5 + ce-maintainability #2.

gitnexus/test/fixtures/lang-resolution/go-pointer-receiver-field-chain/handlers/handler.go
```diff
@@ -0,0 +1,71 @@
...
+// findClassBindingInScope, that string matched no class binding, so receiver
+// typing declined at the BASE and every `h.field.Method()` here emitted no CALLS
+// edge — the dominant Go idiom, silently missing from the graph.
+//
+// The value-receiver twin at the bottom is the control: it resolved before the
+// fix and must keep resolving after it. Field decoration is NOT the variable —
+// Go already normalizes field type bindings at capture via normalizeGoTypeName.
+package handlers
+
+import "fixture/repository"
```
magyargergo, line 16: **P3.** This fixture spans two packages (`handlers/`, `repository/`) and imports across them, but the fixture root has **no `go.mod`** — verified by listing the directory at HEAD. `go/import-target.ts:13-14` documents two tiers: (1) go.mod-based, strip module prefix and match the package directory; (2) non-go.mod / GOPATH, progressively shorter directory suffixes. This fixture therefore resolves via **tier 2**, while the pointer-receiver idiom it guards occurs in real Go repos that always have a `go.mod` (tier 1).

**29** sibling `go-*` fixtures under `test/fixtures/lang-resolution/` carry a `go.mod`; this one does not (verified by listing each fixture root at HEAD).

The regression guard for this PR's headline Go fix passes through an import path real Go repos never take — the "a test can pass via a DIFFERENT tier" failure class this repo has hit before.

**Fix:** add `go-pointer-receiver-field-chain/go.mod` with `module fixture` plus a `go` directive matching the siblings. Note this moves the Go scope-capture fingerprint, so `bench/scope-capture/baselines.json` needs `--update-baseline` and a `_rebaselined_*` note.

gitnexus/test/unit/incremental-parse-cache.test.ts
```diff
@@ -101,9 +101,18 @@ describe('fileContentHash', () => {
...
+  // Java/Kotlin Spring AOP capture side-channels (#2416), 37 -> 38 for
+  // receiver-chain wire format v2: every persisted chain string changed prefix
+  // and a v2 decoder refuses v1 by design, so a cache stamped 37 replays chains
+  // this build silently discards.
+  //
+  // This pin has now earned its keep SEVEN times, and the seventh was the first
+  // EXACT clash rather than a near-miss: main took 37 for #2416 while this
+  // branch already used 37, so two incompatible schemas both claimed the same
+  // number. Re-check against origin/main before merge.
+  it('pins SCHEMA_BUMP to 38 so concurrent bumps cannot silently collide (#2736)', () => {
```
magyargergo: **P3, trivial.** `it('pins SCHEMA_BUMP to 38 so concurrent bumps cannot silently collide (#2736)', …)` — the comment block directly above, the constant's own changelog, and every commit on the branch cite **#2766** (and #2416 for the collision). `#2736` is unrelated. One-character-class fix; worth doing because this title is the breadcrumb the *next* collision will follow.

magyargergo (commented):
# Multi-engine review — PR #2782

**Engines.** Two, not five. **Claude** ran 13 lanes (6 swarm + 7 `ce-*`), all complete. **Codex `gpt-5.6-sol`** ran 3 legs, **2 live**: B1 swarm solo (COMPLETE, verdict `not production-ready`, findings R1–R5) and B3 adversarial (COMPLETE, verdict `needs-attention`, one `[high]`). **B2 `ce-code-review` died on a workspace spend cap before emitting any findings** — Engine B ran at 2/3 strength. Everything below marked "cross-engine" means B1 and/or B3 independently reached it.

Two blockers, both reproduced by executable probes in three independent places (Claude ce-correctness, Claude ce-adversarial, and again by synthesis against the worktree).

**What `[reproduced]` means here (read this before acting on a P1).** The worktree cannot `npm run build`, so the worker pool never starts and no full-pipeline run was possible. Every reproduction below was driven **in-process** — `extractParsedFile` → `finalizeScopeModel` → `buildWorkspaceResolutionIndex` → `foldReceiverChain`, plus capture emitters called directly. The fold results and origin classifications are directly observed; the step from a wrong fold result to an *emitted* edge is traced mechanically through Case 0, not watched end-to-end.

**Merge state:** `mergeable`. **Branch hygiene:** `clean feature/fix PR`. Neither is the blocker — the two correctness findings below are.

---

## Blockers

### P1 — `classifyReceiverOrigin` treats *absent* type evidence as *proof* of externality, so `impact` answers `epistemic: 'exact'` while omitting a real in-program caller

`gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts:192-207` `[reproduced]`

The function has one decision rule: if the base's declared **type** does not resolve to a class, or the base has no type binding at all, return `'external'`. Its own docstring three lines above says the opposite — *"A local, a parameter, a field, a class, or any qualified name counts as in-program"* (`:163-166`) — but `scope.bindings` / value bindings are never consulted. `unresolved-receivers.ts:92-96` then routes `external` drops out of `totalSites` with a `continue`, and `local-backend.ts:523` maps `notes.length === 0` to `epistemic: 'exact'`.

Two distinct triggers, both executed:

```
GO   h typeBinding rawName = "*Host"
GO   findClassBindingInScope(*Host) [no stripper] = undefined
GO   classifyReceiverOrigin("h")  => external          # Host IS in the index
TS   classifyReceiverOrigin("svc") => external         # PR's OWN drop fixture
```

1. **Go pointer receivers — the exact population #2766 exists to fix.** `resolveClassBindingForName` (`:117-158`) calls `findClassBindingInScope` with **three** arguments; it never receives `stripTypePreservingDecoration`, the stripper this PR adds two files away and threads into the other three lookups (`compound-receiver.ts:351, 513, 647`). `func (h *Host)` binds `h` to the literal `*Host`, the lookup misses, and every drop rooted at `h` is filed "outside the analyzed program" — which for `*Host` is exactly backwards.
2. **A base with no type binding at all — including this PR's own new fixture.** `typescript.test.ts:3347` `export async function droppedCall(svc) { svc.getUser().save(); }` is an unannotated parameter; it is not a class, not a qualified name, so it terminates at `return 'external'` (`:207`). The PR's canonical example of a genuinely-lost in-program call is classified by its own classifier as leaving the program. Same for a type alias, a generic parameter, and any base whose binding names a non-class-like def (all executed).

**Consequence.** An agent runs `impact({target:'save'})` before editing. `epistemicFrom` returns `{epistemic:'exact', causes:{receiverTyping:0, …, externalBoundary:N}}` with no `boundaries` prose. Per this repo's own CLAUDE.md gate, the agent proceeds on a count it has been told is complete while N real in-program callers are absent. That is **strictly worse than the pre-PR hedge, which at least fired.** `bench/receiver-resolution/baseline.json` now pins `callDropsByOrigin.external: 76` under exact match, so the misclassification is frozen into the gate.

**Corroboration — cross-engine, STRONG.** Claude: ce-correctness (F1 + F2), ce-adversarial (P1-2, conf 95), risk-architect (F1, Go path only), ce-testing (P1), facts-historian (P2, Go path only). Codex **B1 R2 (P1, conf 100%, blocks merge)** and Codex **B3 `[high]`** — both pinned independently to the `droppedCall(svc)` fixture. The PR description itself already concedes *"The `external` terminal fallback is inverted."* Note one internal disagreement: Claude's swarm risk lane rated only the Go half and called it P2 / non-blocking; it did not reach the no-binding-at-all half. Severity here follows the reproduction and the two Codex legs.

**Fix (as converged by both engines).** Make `external` a *positive* determination: (a) consult value/parameter bindings before the fallthrough so any base the program declares stays in-program; (b) pass `stripTypePreservingDecoration` into `resolveClassBindingForName`; (c) default everything else to `'unknown'`, which already counts toward the hedge. Until (a)/(b) land, `externalSites` should not be deducted from `totalSites`.

---

### P1 — The `index` step's identity fallback types the CONTAINER as its own element, converting an invisible gap into a confident wrong `CALLS` edge

`gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts:498-522` `[reproduced]`, reached through the widened gate at `receiver-bound-calls.ts:476`

```ts
const element = declared === undefined ? undefined : options.elementTypeOf?.(declared, { kind: 'index' });
if (element !== undefined) { … }
else if (current.def === undefined) { return undefined; }
continue;                       // ← identity: `current` is still the CONTAINER
```

The justification above it — *"The declared type here is therefore already the awaited / element type"* — holds when the base *is* a container. It is false when the declared type is an ordinary class that happens to be subscriptable (a C# indexer, a Python `__getitem__`, a TS index signature, a C++ `operator[]`, a Kotlin `operator fun get`), and it is false when capture reduced *more* layers than the source had indexes.

**This is new damage, not an inherited gap — verified.** `SUBSCRIPT_NODE_TYPES` and `AWAIT_EXPRESSION_NODE_TYPES` have **zero occurrences at the diff base** (`git show <base>:…/call-analysis.ts`), so no chain was minted for a subscript receiver; and `grid[0]` contains neither `.` nor `(`, so Case 0's old punctuation gate never fired and every later case missed on the literal text. Pre-PR these sites emitted **nothing and recorded nothing**. Post-PR the widened gate (`site.receiverChain !== undefined`, `:476`) admits them and Case 0 emits at 0.85 confidence against the wrong owner. That inverts the PR's own success criterion ("a missing edge is recoverable, a confident wrong edge is not").

Three *different* triggers, reached by two engines — independent triggers for one defect, not one claim repeated:

| # | Trigger | Result | Correct | Source |
|---|---|---|---|---|
| 1 | TS `class Grid { run(); [i:number]: Item }`, `grid[0].run()` | `Grid.run` | `Item.run` | Claude ce-correctness `[reproduced]` |
| 2 | C# `class Table { public Row this[int i]; void Render() }`, `t[0].Render()` | `Table.Render` | `Row.Render` | Claude ce-adversarial `[reproduced]`, plus Python `__getitem__` + TS index signature |
| 3 | TS `repos: User[][]`; capture strips to a fixed point so the binding is `User`; **one** index → identity → `User`, and `repos[0].map(…)` is validly `Array.map` | `CALLS → User.map` | `Array.map` | Codex **B1 R1 (P1, 99%, blocks merge)** |

Reproduced again during synthesis against the worktree:

```
f5 b[0].save()  (Box has save + index sig)  -> Box       ← WRONG owner
f7 s[0].save()  (Service, NOT a container)  -> Service   ← identity on a non-container
   typeBindings :: grid -> "User"                        ← `User[][]` reduced to `User` at capture (trigger 3's premise)
```

**Blast radius, honestly bounded.** Only **TypeScript and C#** implement `elementTypeOf` (verified: 2 hits across all `languages/*/scope-resolver.ts`), so every index step in the other twelve is pure identity. Claude ce-adversarial then *refuted* the two scariest instances: Go is accidentally safe (`type Pipeline []Task` is not captured as a class-like def) and Java is accidentally safe (array parameters get no typeBinding). Exposure concentrates in languages with a first-class indexer reachable from a typed base.

**The bench cannot see this.** `measure.mjs --check` is exact-match on drop counts and edge *presence*. This failure mode is an edge that **appears** — it reads as a resolution win. Neither arm distinguishes `Table.Render` from `Row.Render` unless the baseline pins the target symbol.

**Fix.** This is a design call, and both engines land in the same place: stop inferring which regime the fold is in. Either make `elementTypeOf` mandatory for a language to participate in the `index` step (capture-normalizing languages implement it as "already reduced ⇒ return the name unchanged"), or add an explicit `containerReducedAtCapture` signal so identity is legal only when set. Deleting the identity branch outright is the conservative option but costs most of the newly-gained edges.

---

## What the reviews validated

This is careful work with an unusually strong measurement culture, and a large share of the review budget went into *failing* to break it.

- **CI is genuinely green.** `mergeable=MERGEABLE`, `mergeStateStatus=**CLEAN**`, **zero failing and zero pending checks** on head `ceca5c961` — typecheck, format, lint, CodeQL, gitleaks, zizmor, actionlint, all platform shards, coverage merge, and the benchmark job. (An earlier lane and the first coordinator probe saw `BLOCKED` with shards pending; that is stale.) `tsc --noEmit` clean in-worktree.
- **The Go fix is right, and right in the right place.** The receiver self-binding stays raw (`method-owners.ts` needs `*T` vs `T` for Go method sets) and the stripper is applied **only at lookup**, as an optional 4th parameter passed from three call sites. The ~26 `findClassBindingInScope(…) ?? otherResolver(…)` sites that a global widening would have silently suppressed are untouched. Confirmed by four lanes across both engines.
- **The codec survived a determined attack.** Claude ce-adversarial threw v1-in-warm-cache, members named `a`/`i`/`c`/`f`, empty-name steps, lone sigils, mid-string truncation markers, `|` in the base, over-depth chains, version `3`, no version, zero-width chars, byte-cap boundaries and **all 64 three-step kind permutations** at it: 0/64 round-trip failures, every malformed payload refused. Claude security-boundary independently refuted prototype pollution in `externalCounts`, partial v1 decode, trailing garbage after a name-free sigil, catastrophic backtracking, and unbounded payload growth.
- **Cache coherence holds.** `PARSE_CACHE_VERSION` gates the durable ParsedFile store, `SCHEMA_BUMP 37→38` invalidates it, `INCREMENTAL_SCHEMA_VERSION 33→34` forces re-analyze — plus defence in depth: `parsedfile-store.ts:385-395` strips an unreadable `receiverChain` and keeps the site (observed firing).
- **All seven capture goldens were regenerated** — the completeness failure this series hit once before did not recur.
- **Branch hygiene is clean**: 20 linear commits, 0 behind main, no merge commits, no lockfile/`CHANGELOG`/build-output churn, and the `ci-tests.yml` diff is **comment-only** (the `run:` command and `working-directory:` are unchanged; no validation weakened).
- **Schema versions are clean at 38/34** against `origin/main`'s 37/33, with no other open PR claiming them. This would have been the **8th** collision in this series and the last was an exact clash — **re-check immediately before merge, not at review time.**
- **Prior-context item 15 is resolved, verified rather than assumed.** `(await svc.getUserAsync()).save()` → `2|svc|cgetUserAsync|a` → folds to the awaited class; `repos[0].save()` → `2|repos|i` → folds. The fourth attempt worked.

---

| # | Sev | Location | One line |
|---|---|---|---|
| 1 | **P1** | `receiver-bound-calls.ts:207` | `external` as terminal fallback → false `epistemic: 'exact'` (blocker above) |
| 2 | **P1** | `compound-receiver.ts:522` | index-step identity fallback → confident wrong `CALLS` edge (blocker above) |
| 3 | P2 | `go/captures.ts:218` | The callee-read drop also deletes the genuine `ACCESSES` on a **func-typed struct field** (`h.dep.Work()` where `Work func() error`). Common in callback/hook/mock structs. |
| 4 | P2 | `contract/scope-resolver.ts:1162` | 20-line JSDoc for an "Unwrap a COLLECTION spelling" hook with **no declaration after it** — reads as contract for a member that does not exist. |
| 5 | P3 | `compound-receiver.ts:386` | `typeOfMemberOnClass`'s hoisted branch returns `undefined` on a class miss instead of carrying `declaredTypeName` forward (and omits the stripper), so `svc.getMap()['k'].run()` declines while byte-identical `byId['k'].run()` resolves. |
| 6 | P3 | `local-backend.ts:5902` | `EpistemicCauses` mixes units in one struct: `receiverTyping` counts SITES, `dispatchBoundary` counts NOTES. |
| 7 | P3 | `resolution-outcome.ts:143` | `classifyReceiverShape`'s rationale docblock floats unattached (a second JSDoc intervenes); the function ships undocumented. |
| 8 | P3 | `unresolved-receivers.ts:86` | Two consecutive comment blocks make the same routing point — an incomplete editorial substitution. |
| 9 | P3 | `handler.go:16` | New Go fixture has two packages and a cross-package import but **no `go.mod`**, so the regression guard for the headline fix validates through tier-2 (GOPATH-suffix) import resolution while real Go repos always take tier 1. All 29 sibling multi-package Go fixtures carry one. |
| 10 | P3 | `incremental-parse-cache.test.ts:114` | Test title cites `(#2736)`; every other reference in the branch is `#2766`. |

---

## Body-only (real, but the file or line is not in the diff)

- **MCP `impact` / `context` tool descriptions never document `causes`.** `src/mcp/tools.ts` is untouched. The PR's own code says the field exists so *"a coding agent … can branch on the difference"* — but no consumer can discover it. **Cross-engine**: Claude docs-dod F1 + Codex B1 R4 (99%).
- **`ARCHITECTURE.md:266` still lists `unwrapCollectionAccessor?`**, a hook this PR removed; neither `elementTypeOf?` nor `stripTypePreservingDecoration?` appears in the table. Three Claude lanes.
- **`RepoMeta.unresolvedReceiverMembers` (`repo-manager.ts:244-248`) was not extended** with `externalCounts`/`externalSites`. It works only because meta IO is untyped JSON and the assignment is a call expression (no excess-property check). Any future validator or field-by-field meta rebuild silently drops them and the split reverts to "everything hedges" with no compile error. Three Claude lanes.
- **`KNOWN_BLIND` / `BASELINE.md` are stale within their own PR.** `measure.mjs:1485-1489` still prints *"`repos[0].save()` has neither `.` nor `(` … same result"* on every run, while `baseline.json` has `indexElement: "RESOLVES"`; `BASELINE.md` states callDrops **99** where the committed gate enforces **102**, and its "Known blind spots" retains three claims its own Corrections section retracts. **Cross-engine**: Codex B1 R5 (100%) + Claude docs-dod F6/F7/F8.
- **Codex-only lead (single lane, P2, not reproduced — flagged, not asserted):** external per-member counts are `slice(0, MAX)`'d with **no `omittedNames` twin** (`unresolved-receivers.ts:121-127`), while internal counts persist one. Above 500 distinct external names, `lookupExternalCallCount` returns `undefined` for a truncated name and the symbol appears to have no boundary evidence.
- **`stripTypePreservingDecoration`'s "receiver-chain base/step only" invariant is a comment, not a mechanism.** It rides on the shared `compoundOpts` bag and already reaches the general text cascade's `viaTb` branch and Case 3b. Inert today (Go is the only implementer, `*T` resolved to nothing there before), but it inherits the widening the moment a second language adopts the hook — which the contract itself names as planned for Rust, C#, Swift, TS, C++.

---

## Refuted — first-class results

Each of these was probed with a concrete trigger and **did not survive**. Recording them so they are not re-litigated:

- **Codec v2 round-trip forgery** — clean in both directions; v1 refused on the version prefix, v2 refused by a v1 decoder, a member named `a`/`i` encodes as `ca`/`fi` and never aliases a sigil, 0/64 permutation failures. Only nit: `MAX_RECEIVER_CHAIN_BYTES` compares `String.length`, so an astral-plane identifier can exceed 512 *bytes* — a size guard, not correctness.
- **Cache coherence / a v1 chain read by a v2 decoder** — impossible across all four probed paths (durable store, incremental writeback, cache slimming, pre-summary meta).
- **The Go stripper producing an ambiguous or wrong class** — `'*'` breaks on the length guard, `MAX_DECORATION_LAYERS=3` bounds `**T`, and "exactly one or decline" holds on every iteration; two same-named cross-package types return ≥2 and decline. Two-package same-name ambiguity does **not** break the #2766 fix (probed).
- **The construction-selector veto reorder breaking Ruby** — refuted, *and the reorder is necessary*: a name-free step has `step.name === undefined`, so with the veto first every language without a construction selector compared `undefined === undefined` and vetoed the whole fold. `Factory.new.run → Product.run` stays closed.
- **The cast-unwrap trap** ("never unwrap a cast") — `TRANSPARENT_RECEIVER_WRAPPERS` is `{non_null_expression, parenthesized_expression}` only; peeling a paren exposes the cast node, which matches no branch and mints nothing. `(svc as Svc).get().run()` mints no chain; `((svc)).get().run()` and `maybe!.get().run()` do.
- **Double-unwrap from the new `index` step** — cannot happen on TS/C#/Go; the real failure is the *opposite* (under-unwrapping = blocker 2).
- **Case 0's widened gate suppressing a later case or a drop** — refuted; the site enters `handledSites` only on emit or an ambiguity/deleted-target suppression. The `callDrops` 101→102 rise is the instrument gaining sight of shapes it was blind to, **not** a regression.
- **Workflow / supply-chain risk** — the CI diff is comment-only; no new `uses:`, no `permissions` widening, no `${{ github.* }}` in changed context.
- **Prototype pollution in `externalCounts`** — `lookupExternalCallCount` mirrors its twin exactly (`Object.hasOwn` → `typeof` → `Number.isFinite` → `> 0`).
- **Bench runner escaping the repo** — writes confined to a `mkdtempSync` dir cleaned in a `finally`; the only in-repo write is `--update-baseline`-gated and CI passes `--check`.
- **AGENTS.md R6 ("shared pipeline code must not name languages")** — **not a violation**. `AWAIT_EXPRESSION_NODE_TYPES` / `SUBSCRIPT_NODE_TYPES` follow the pre-existing `CALL_EXPRESSION_TYPES` / `FIELD_ACCESS_NODE_TYPES` pattern in the same file: `Set.has(node.type)` with language names only in comments. No branching.
- **Dart / Ruby / Swift regressions from the loop-entry wrapper peel**, **non-termination in `extractMixedChain`**, **`FoldState.def === undefined` carried onto a wrong owner**, and **`elementTypeOf` activating the C#-only accessor block for TS** — all four refuted.
- Two comments in the new code are, however, **empirically false** and worth correcting rather than leaving as wrong rationale: `receiver-bound-calls.ts:466` claims PHP `->`/`::` receivers were lost the same way (PHP receiver text arrives dot-normalized — `$this.u.addr` — so it always passed the punctuation gate), and `compound-receiver.ts:502-503` justifies identity with a TypeScript `User[]` annotation that capture has in fact already reduced.

---

## Test gaps

Ordered by what they would have caught:

1. **Nothing asserts `receiverOrigin` anywhere** (`git grep receiverOrigin gitnexus/test` → 0). The one integration fixture that would have caught blocker 1 — `typescript.test.ts:3347` — asserts only `siteKind` on the raw `ResolutionOutcome`, which is unaffected by origin. The new Go fixture asserts CALLS/ACCESSES sets but nothing epistemic.
2. **`EpistemicCauses` / `causes` has zero coverage at any level** (`externalCounts`, `externalSites`, `lookupExternalCallCount`, `EpistemicCauses`, `receiverTyping` → 0 hits across `gitnexus/test/**`). `impact-epistemic-lower-bound.test.ts` was not touched and asserts only `epistemic` + `boundaries`.
3. **No fixture exercises a subscript on a non-container** — every index-step fixture is a genuine container, precisely the regime where identity is correct. `Table this[int i]` (C#), `__getitem__` (Python) and an index-signature class (TS) rows would pin blocker 2 either way.
4. **The decoration fallback's ambiguity decline (`candidates.length > 1 → undefined`)** — the one safety property this whole line of work exists to protect — has no test. The new Go fixture has exactly one class per stripped name, so only the happy path runs.
5. **`classifyReceiverShape`'s `chain-unwrap` bucket and its dominance rule are untested** — 4 of 5 buckets covered; the bucket the PR exists to expose is the missing one.
6. **`receiver-chain-fold.test.ts` gained zero new cases** — its +20 lines are a pure `1|`→`2|` prefix bump. The `await`/`index` steps, the `FoldState` carry-forward, and all three index sub-branches are unit-untested; `ts-receiver-chain-capture.test.ts:31` even declares `declare const repos: User[];` for a test that was never written.
7. **C#'s new `elementTypeOf` index route has no test at all**, and its one measured bench cell (`shapeArm.csharp.indexElement`) still reads `VISIBLE-GAP` post-fix.
8. **The `f := h.dep.Work` method-value case is asserted only in a comment** — the shape does not exist anywhere in the Go corpus, so the guard distinguishing it from `h.dep.Work()` is unpinned.
9. **Fixing blocker 1 will move `callDropsByOrigin.external: 76`**, which the bench pins under exact match. Worth knowing before the fix lands so the movement is not read as a regression.

---

## Coverage limits of this review

- **No full-pipeline execution was possible.** The worktree cannot `npm run build`, so the worker pool refuses to start and `runPipelineFromRepo` / `measure.mjs --check` / all integration resolvers are unreachable. Every reproduction above was driven **in-process** through `extractParsedFile` → `finalizeScopeModel` → `buildWorkspaceResolutionIndex` → `foldReceiverChain`, plus per-language capture emitters called directly. Edge-*emission* claims are traced mechanically from a reproduced fold result through Case 0, not observed end-to-end.
- **Five of the six Claude swarm lanes ended mid-investigation on first dispatch** and were resumed with an explicit prod. Their findings are complete; their *coverage* is likely shallower than a clean run.
- **Codex B2 (`ce-code-review`) died on a workspace spend cap** after a substantial investigation but before emitting findings. Its two surviving partial sentences are not counted as corroboration for anything.
- Targeted unit suites that *can* run in the worktree all pass (121 tests / 7 files: codec, fold, unresolved-receivers, ts-chain-capture, parsedfile-store, incremental-parse-cache, call-summary-schema-version). `tsc --noEmit` exit 0.
- Not independently verified: the BASELINE.md narrative counts (76/20/6 origin split, reads 27→22) and the "4 CALLS edges → 10, impact 0 → 1" claim, which refers to the reporter's private ~470k-LOC codebase and is not backed by a committed assertion.

---

## Final verdict

**`not production-ready`**

The mechanics are clean — `mergeable`, `clean feature/fix PR`, CI fully green, schema versions uncollided, goldens complete, and a codec that survived a determined attack. The blockers are both correctness, and both sit in the two features this PR is named for. Blocker 1 makes `impact` publish `epistemic: 'exact'` for calls that really were lost, which is strictly worse than the hedge it replaces and is already conceded in the PR's own description; it misfires on the PR's own new fixture and on the exact Go pointer-receiver population #2766 exists to fix. Blocker 2 converts a pre-PR invisible gap into a confident wrong `CALLS` edge, inverting this codebase's own stated ordering that a missing edge is recoverable and a wrong one is not — and the bench cannot see it, because the failure mode is an edge that *appears*.

Both have concrete triggers, small fixes, and clear regression tests to pin them. Neither is a design dead end. Once they land — with `receiverOrigin` and the subscript-on-a-non-container case actually asserted, since nothing tests either today — this is ready.

---

*Automated multi-tool review digest (Claude swarm + `ce-*` personas; Codex `gpt-5.6-sol` swarm + adversarial). Verify each finding against the code before acting on it.*

gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
magyargergo: Fixed in 616c3590. \`external\` is now a positive determination via \`LanguageProvider.isBuiltInName\` (the hook already used for the #2545 shadow guard); everything unproven returns \`unknown\`, which counts toward the hedge. \`resolveClassBindingForName\` now receives \`stripTypePreservingDecoration\`, so the Go `*Host` base resolves.

Two things the fix surfaced that the finding did not predict:

- **Java had no \`builtInNames\` set at all**, so the change would have left it unable to name its own boundary — every Java drop hedging forever. It gets one, deliberately under-inclusive: \`Map\`/\`Set\`/\`Collection\`/\`Error\`/\`Number\`/\`Record\` are excluded because they are domain nouns a same-package type shadows with no import, which is exactly the shape that would produce a silent false \`external\`.
- The value-binding arm does **not** rescue \`droppedCall(svc)\` — an unannotated TS parameter is recorded nowhere, so it is the \`unknown\` default that catches it. Kept the arm anyway as a direct implementation of the documented contract.

Measured: \`callDropsByOrigin\` 76/20/6 → 44/36/22, \`callDrops\` unchanged at 102. A controlled A/B over the Java set (off/on, same tree) reads 7/36/59 vs 44/36/22 — \`in-program\` byte-identical, so nothing demonstrably ours was reclassified.

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
magyargergo: Fixed in 4ec6a67f. The identity fallback is gone — the index step now declines when \`elementTypeOf\` answers nothing.

Declining alone would have deleted the feature rather than trimming it: measured, **all six** \`indexElement\` RESOLVES cells came from the identity path, not from the hook. So the fix restores the evidence capture destroys. \`TypeRef.declaredSpelling\` preserves the as-written spelling, derived generically in pass 4 from \`@type-binding.type\` — the anchor all 14 interpreters already read — so no interpreter needed threading and none can forget it. Five more languages now answer the index route.

Result: the three wrong-edge triggers (\`grid[0].run()\`, C# indexer, \`User[][]\`) all decline or resolve correctly, **zero bench shape cells moved**, and \`callDrops\` stays 102.

gitnexus/src/core/ingestion/languages/go/captures.ts
magyargergo: Fixed in a068f2bd. Capture now records the POSITION as a fact (\`@reference.callee-position\` → \`ReferenceSite.inCalleePosition\`) and emit makes the decision from the resolved target's kind — neither layer holds both facts, so the split is necessary rather than incidental. Suppression is language-neutral in \`graph-bridge/edges.ts\` and keys on the canonical \`CALL_TARGET_TYPES\`, which also picked up \`Macro\` and \`Delegate\` (the hand-rolled set had dropped exactly those two).

**One correction to my own finding:** the chained controls I cited (\`h.dep.Name\`, \`f := h.dep.Work\`) produce no edges before *or* after — compound-receiver reads do not resolve at all today, a separate pre-existing gap. The capture-level evidence was right; at edge level only bare-name receivers demonstrate anything, so the fixture uses those with the reason in a comment.

\`read\` drops 22 → 27 and \`totalDropsAllKinds\` 124 → 129: the five genuine reads the previous commit was scoring as an improvement to delete.

gitnexus/src/core/ingestion/scope-resolution/contract/scope-resolver.ts
magyargergo: Fixed in 4ec6a67f — the orphaned block is deleted and its substance folded into the real \`elementTypeOf\` declaration.

Two more instances of the same defect turned up while fixing it: a `typeOfMemberOnClass` docblock sitting above `FoldState`'s docblock and interface (so it documented neither), and a parenthetical inside the merged `elementTypeOf` doc that described the *history of the comment* and pinned a commit hash. Both cleaned up.

gitnexus/src/core/ingestion/scope-resolution/passes/compound-receiver.ts
magyargergo: Fixed in 4ec6a67f. The hoisted branch now carries `declaredType` forward on a class miss exactly as the primary branch does, and passes `stripTypePreservingDecoration` like its three siblings. Verified by the asymmetry that motivated the finding: `fold 2|svc|cgetMap|i` and `fold 2|byId|i` now agree.

`FoldState` also lost a field — it carried both `declaredTypeName` and `declaredSpelling` when the only read anywhere coalesced them, and the header docblock had drifted into contradicting the in-body comment about which one was load-bearing.

gitnexus/src/mcp/local/local-backend.ts
magyargergo: Fixed in aaf4700f. `dispatchBoundary` now counts implementations + consumers rather than notes, so it is comparable with `receiverTyping`.

Worth recording why it is not a literal call-site count: that is **not derivable** on the dispatch side. Consumers are counted `COUNT(DISTINCT other.id)`, and languages setting `collapseMemberCallsByCallerTarget` emit one CALLS edge per (caller, target) pair regardless of syntactic site count — so neither the row count nor an edge count recovers sites. The residual granularity difference (`receiverTyping`/`externalBoundary` = call sites, `dispatchBoundary` = symbols) and the fact that it is itself a lower bound are now documented on the fields rather than left implicit.

gitnexus/src/core/ingestion/scope-resolution/resolution-outcome.ts
magyargergo: Fixed in aaf4700f — the rationale block now sits directly above `classifyReceiverShape`, and `ReceiverOrigin` keeps its own.

Correcting the mechanism I gave in the original comment, which was wrong: TSDoc does not attach the block to `ReceiverOrigin`. A *second* JSDoc intervenes before the type, so the first block floated attached to nothing while `ReceiverOrigin` was correctly documented by its own. The finding held; my explanation of it did not, and it would have been disproved by a TypeDoc run.

gitnexus/src/core/ingestion/scope-resolution/unresolved-receivers.ts
magyargergo: Fixed in 616c3590 — the first block is gone, the second kept (it carries the auditable/reversible point the first lacked) and rewritten to describe the post-fix contract, since `external` is now a positive determination rather than a fallthrough.

Git archaeology confirmed the guess in the finding: one commit wrote the paragraph to justify a `continue`-discard, a later one rewrote the code to route-and-persist and appended a new paragraph without deleting the first.

Also in this file: the rank/cap/omitted block was cloned for the external map, so the stable-ordering contract was being asserted by two comparators that had to stay identical. Extracted to one `rankAndCap`.

gitnexus/test/fixtures/lang-resolution/go-pointer-receiver-field-chain/handlers/handler.go
magyargergo: Fixed in a068f2bd — `go.mod` added (`module fixture`, `go 1.21`, matching the sibling majority).

**It was worse than I reported.** I said the fixture validated through tier 2; measured, it emitted **zero IMPORTS edges** — tier 2 could not have helped either, because its loop is `i < parts.length - 1`, so for `fixture/repository` it only ever tries the full suffix, which matches nothing. The cross-package CALLS were resolving through the registry/type-binding path instead, and `reason: 'import-resolved'` is a cross-file label rather than evidence of import resolution.

After adding it, the only edge-set change is `+ IMPORTS handler.go → repo.go` — every CALLS and ACCESSES edge byte-identical, so the pointer-receiver guard now exercises the tier real Go repos take. Go scope-capture fingerprint rebaselined accordingly.

gitnexus/test/unit/incremental-parse-cache.test.ts
magyargergo: Fixed in aaf4700f (#2736 → #2766).

The title then had to change again in 9b334ee8: `main` took **38** for the Swift directive fix (#2771) while this branch was sitting in review, so the branch renumbered to **39**. That is the eighth collision in this series and the second exact clash — and it happened in exactly the window this comment called the breadcrumb for. The pin and its comment now record both clashes and the rule that the check belongs immediately before merge, not at review time.

PR #2783, opened 2026-08-01
feat(spring): model AOP transactions, caching, and security

### LINKED ISSUE #2416: feat(spring): model AOP transactions caching and security

ShiningXu:
Closes #2416

Part of #1225.

## What changed

- Added a shared Spring AOP metadata model and post-resolution pipeline phase that emits explicit `ADVISED_BY` relationships and versioned evidence for:
  - transactional, caching, and method-security annotations
  - `@Aspect` declarations and common advice annotations
  - statically understood `execution`, `within`, and recognized `@annotation` pointcuts
  - conservative evidence for unsupported or dynamic pointcuts
- Added Java and Kotlin capture side channels, cache-version invalidation, schema/incremental lifecycle support, and context/impact enrichment.
- Added database-side reason filtering so foreign `DECLARES`/`ADVISED_BY` rows cannot consume the Spring AOP metadata query budget.
- Kept `@annotation` method-declaration-only: class-level behavior is fanned out as proxy behavior but is not incorrectly treated as `@within`/`@target`.

## Kotlin support

The supported #2416 static subset is covered for `.kt` and `.kts`, including classes, interfaces, singleton objects, class- and method-level behavior, import aliases, wildcard/FQN annotations, aliased Aspect/advice annotations, triple-quoted pointcuts, suspend members, and member extension functions. Private and top-level functions fail closed where they are not proxy-eligible.

The durable parse-cache test verifies a real cold worker parse, cache + ParsedFile shard persistence, an all-hit warm replay with no worker creation, and byte-equivalent Java/Kotlin AOP edges.

## Performance and incremental safety

- Advice resolution builds one candidate index per pipeline run:
  - exact/literal-prefix owner lookup for `within`/`execution`
  - direct annotation lookup for `@annotation`
  - safe full fallback for leading or unqualified wildcards
  - the pointcut matcher remains the final authority
- A 50,000-method regression fixture reduces selective inspections from 850,000 brute-force checks to 15,140 while preserving identical matches.
- Graph-wide relationships such as `ADVISED_BY` no longer expand the incremental write set; a 100-target regression test keeps an Aspect-only change at one writable file.
- A real full + two-incremental-run test verifies old evidence deletion, retargeted advice, graph-wide rebuild, and duplicate-free idempotency.

## Validation

- `npx vitest run` over the AOP/unit/integration/storage/MCP matrix: **135 passed, 2 benchmark-gated skipped**
- `GITNEXUS_BENCH=1 npx vitest run test/integration/spring-aop-benchmark.test.ts`: **4 passed**
- Real LadybugDB incremental lifecycle test: **passed**
- `npx tsc --noEmit`: **passed**
- `npm run build`: **passed**
- ESLint on changed files: **0 errors**
- `git diff --check`: **passed**
- GitNexus `detect-changes`: 33 files / 61 symbols / 5 expected flows, MEDIUM risk

A full `npm test` run also completed with 15,556 passing tests and 14 non-AOP failures tied to existing platform/test-environment assumptions (`/var` vs `/private`, Linux-only `/proc/self/fd`, external review helper state, and full-run timeout cases that pass in isolation). No Spring AOP test failed.

## Scope boundary

This intentionally remains a conservative static subset. Named/compound/dynamic pointcuts are retained as unresolved evidence, and exact unqualified type names fail closed until advice import/package context is available.


gitnexus/src/storage/repo-manager.ts
```diff
@@ -636,8 +636,13 @@ export interface RepoMeta {
...
  * Pre-v30 indexes keep the wrapper line on unchanged files and would keep
  * failing closed (no edge) through the reuse gate. Force a full re-analyze.
+ *
+ * v31: Spring AOP evidence adds the Interface→CodeElement relation pair
+ * (#2416). LadybugDB fixes allowed endpoint pairs when the relation table is
+ * created, so an older index cannot persist these edges through incremental
+ * writeback. Force a full re-analyze.
  */
-export const INCREMENTAL_SCHEMA_VERSION = 30;
+export const INCREMENTAL_SCHEMA_VERSION = 31;
```
magyargergo: **P0 · STRONG (near-unanimous, both engines) · [reproduced]**

`INCREMENTAL_SCHEMA_VERSION` is bumped to 31, but `origin/main` is already at **32** — PR #2746 took 31 (Python namespace-import edges, unrelated content), PR #2769 took 32 (Swift CONTAINS DDL pairs). Verified live: `git show origin/main:gitnexus/src/storage/repo-manager.ts | grep INCREMENTAL_SCHEMA_VERSION` → `= 32;`.

The reuse gate at `gitnexus/src/core/run-analyze.ts:1218` is exact-equality, so a repo already indexed under main's real v31 would silently pass this PR's gate and skip the full rebuild the new `Interface→CodeElement` DDL requires — no error, just a DB quietly missing `ADVISED_BY` rows. This is also a real git-level merge conflict (not just semantic drift) against `origin/main`, confirmed via `git merge-tree`.

This is (per this file's own comment history) a recurring failure class in this repo — documented as having bitten multiple times before.

**Fix:** rebase onto current `origin/main`, allocate the next free integer (**33**, not 31 or 32 — this repo's convention is never reuse a taken number), update this comment block.

gitnexus/test/unit/call-summary-schema-version.test.ts
```diff
@@ -73,12 +73,12 @@ describe('CALL_SUMMARY relation-type exclusion (U-C1)', () => {
...
 
 describe('CALL_SUMMARY incremental reuse gate (U-C5)', () => {
-  it('INCREMENTAL_SCHEMA_VERSION is bumped to 30 (multi-line closure startLine join, #2735)', () => {
+  it('INCREMENTAL_SCHEMA_VERSION is bumped to 31 (Spring AOP relation pairs, #2416)', () => {
     // Moves with every bump BY DESIGN — that is the point of pinning it. A
     // change that alters emitted ids or edges without bumping would otherwise
     // ship silently, and an existing index would keep serving the old graph
     // through the reuse gate below.
-    expect(INCREMENTAL_SCHEMA_VERSION).toBe(30);
+    expect(INCREMENTAL_SCHEMA_VERSION).toBe(31);
```
magyargergo: **P0 · STRONG (branch-hygiene + test-ci + facts-historian lanes, both engines) · [reproduced]**

This test pins the colliding version number in three places — this line (`expect(INCREMENTAL_SCHEMA_VERSION).toBe(31)`), the test description on line 76, and `expect(passesReuseGate(31)).toBe(true)` on line 210 (separate hunk, same file). This file is one of exactly 2 files with real `git merge-tree` conflict markers against `origin/main` (the other is `repo-manager.ts`).

**Fix:** update all three assertions to whatever integer is chosen (33) alongside the `repo-manager.ts` fix, in the same commit.

magyargergo (commented):
# PR #2783 Tri-Review Digest — "feat(spring): model AOP transactions, caching, and security" (closes #2416)

**Head:** `cb6cdd759a2b8fb079a3d41c691e8ed3ed3ca89c` · **Diff base:** `51095c19f8a7c641a136549711065f816a383ff6` · **Scope:** 43 files, +4185/-36

## 1. Engine breakdown

- **Claude Sonnet 5** — 6 swarm-persona lanes (facts-historian, branch-hygiene, risk-architect, test-ci, security-boundary, docs-dod) + 8 ce-code-review lanes (correctness, adversarial, api-contract, maintainability, performance, security, testing, project-standards). 14 lanes total, all returned substantive, non-empty findings.
- **Codex (gpt-5.6-sol, xhigh effort)** — three legs, all live with real findings: `codex-swarm.md` (gitnexus-pr-review skill, Solo mode, lanes 1-6 + `gitnexus/impact`/`detect_changes`/`explain` graph queries), `codex-ce.md` (ce-code-review, report-only mode, ran `npx tsc --noEmit` and 46 focused vitest tests directly), `codex-adversarial.md` (chaos-engineer pass, also used `gitnexus/impact`).

All findings below were independently re-verified against the live worktree (`git diff`/`git show`/direct file reads) rather than taken on lane say-so; citations that didn't hold up were dropped or downgraded.

## 2. What's solid

- Both engines' correctness/adversarial passes actively tried to break this and mostly couldn't: no Cypher injection (all new queries parameterized), no ReDoS (pointcut regexes capped at 1000 chars via the pre-existing `stripBidiAndZeroWidth`/`#2286` hardening, hand-rolled KMP matcher with no backtracking), no unbounded recursion, ADVISED_BY correctly wired into `isGraphWideRelationship`/`isGraphWideNode` for incremental correctness, and the shared `frameworks/spring/aop.ts`/`pipeline-phases/spring-aop.ts` layer genuinely never names Java/Kotlin (verified: delegation via `SpringAopMetadataAdapter` injected per-language, matching the established DI/conditionals pattern).
- A real, proactive cross-feature bug fix ships in the same commit: `bean-metadata.ts`'s `queryClassBeanMetadata` gained an `AND r.reason STARTS WITH 'spring-bean-factory:'` filter (verified via diff) — without it, this PR's new `spring-aop:` `DECLARES` rows on the same node shapes could nondeterministically collide with bean-factory metadata reads. This is proactive hardening, not incidental.
- Claude's `ce-correctness` lane actually executed `spring-aop.test.ts` (26/26 pass) and `incremental-subgraph-extract.test.ts` (16/16 pass) plus 5 adjacent suites (93/94, 1 false alarm — see §5) directly in the worktree. Codex's `codex-ce` leg independently ran `npx tsc --noEmit` (clean, after resolving a stale-build false start) and 46 focused vitest tests (all green). `git diff --check` passed in three independent lane checks (no whitespace/bidi hygiene issues).

## 3. Headline finding — P0, blocks merge

**`INCREMENTAL_SCHEMA_VERSION` collision: this PR sets 31; `origin/main` is already at 32.** [STRONG — near-unanimous across all 14 Claude lanes and all 3 Codex legs] [reproduced]

- `gitnexus/src/storage/repo-manager.ts:645` — this PR bumps `INCREMENTAL_SCHEMA_VERSION` 30→31 with a new comment describing the `Interface→CodeElement` DDL pair. Verified live: `git show origin/main:gitnexus/src/storage/repo-manager.ts` shows the constant is already **32** — main's real v31 belongs to already-merged PR #2746 (Python namespace-import edges, unrelated content) and v32 to PR #2769 (Swift CONTAINS DDL pairs). The reuse gate at `gitnexus/src/core/run-analyze.ts:1218` is an exact-equality check (`existingMeta.schemaVersion !== INCREMENTAL_SCHEMA_VERSION`) — a repo already indexed under main's real v31 would silently pass this PR's gate and skip the full rebuild the AOP schema change requires, producing a DB missing `ADVISED_BY`/`Interface→CodeElement` rows with no error signal.
- `parse-cache.ts`'s `SCHEMA_BUMP` (36→37) does **not** collide — verified live against `origin/main` (still 36). Do not conflate the two constants; only `INCREMENTAL_SCHEMA_VERSION` needs renumbering.
- This is a real git-level merge conflict, not just semantic drift — confirmed via corrected `git merge-tree` output (legacy format prefixes conflict-hunk lines with `+`, which broke a naive `grep -c "^<<<<<<<"` earlier in this review) in exactly 2 files: `repo-manager.ts` (the version line + comment) and `gitnexus/test/unit/call-summary-schema-version.test.ts` (3 conflict blocks: description string, `.toBe(31)`, `passesReuseGate` assertions — verified both are in-diff at lines 76/81). GitHub's `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` is accurate (confirmed live via `gh pr view`), not a stale cache.
- This is (per the repo's own comment history in `repo-manager.ts`) a recurring failure class — this would be the 6th-7th documented recurrence, each time ending with an unheeded "re-check against origin/main immediately before merging" instruction.
- **Fix:** rebase onto current `origin/main`, allocate the next free integer (**33**, not a reuse of 31 or 32 per the repo's own "never reuse, always allocate next" convention), update the comment block and `call-summary-schema-version.test.ts`'s assertions.

## 4. Other inline-worthy findings

**P1 — Interface/superclass-declared annotations never attach to the concrete implementing/overriding method** [STRONG — Claude ce-adversarial + Claude ce-testing (T3) + Codex codex-adversarial, independently] [code-read]
`pipeline-phases/spring-aop.ts:135-136` fans a class-level annotation out only to methods reachable via that owner's own direct `HAS_METHOD` edges (`classMethods`, built at spring-aop.ts:298-316 strictly from `HAS_METHOD`, no `EXTENDS`/`METHOD_OVERRIDES`/`METHOD_IMPLEMENTS` walk); for a *method-level* fact, `sources = [owner]` only. `languages/java/spring-aop.ts:29-56` confirms per-language capture only inspects a class's own body, never `implements`/`extends`. Net effect: the textbook Spring idiom of `@Transactional` on an interface method, implemented by a concrete `@Service` class with no annotation of its own, produces **zero** `ADVISED_BY` evidence on the concrete method — verified against the PR's own 773-line `spring-aop-pipeline.test.ts`, which repeatedly gets close to this exact scenario (interface fan-out fixtures) but never wires up a concrete class implementing an annotated interface.

**P1 — `spring-aop-mcp.test.ts` opens a real LadybugDB but isn't wired into the repo's native-addon test-isolation lists** [WEAK by lane count (Codex-only, `codex-adversarial.md` B3) but not demoted — concrete, verified trigger, matches this repo's own documented Windows/macOS N-API crash history] [code-read]
Verified directly: `test/integration/spring-aop-mcp.test.ts:142` calls `withTestLbugDB` (a real native LadybugDB), but `gitnexus/vitest.config.ts` was not touched by this diff and the file appears in neither the `lbug-db` project's sequential `include` list nor the `default` project's `exclude` list (both enumerated and checked line-by-line). It will run in the parallel default project alongside other native-addon tests, which this repo's own architecture treats as a real file-lock/N-API-teardown crash risk (this exact class of failure — worker/native-addon isolation — is a standing GUARDRAILS-level concern in this codebase's history).

**P2 — Kotlin `companion object` bodies are entirely invisible to Spring AOP capture** [STRONG — Claude ce-adversarial + Codex codex-ce, independently] [code-read]
`languages/kotlin/query.ts:18` tags `companion_object` with the same `@scope.class` capture as `class_declaration`/`object_declaration`, but `languages/kotlin/captures.ts:128` filters `springAopTypeNode` to only `class_declaration`/`object_declaration` — verified live, `companion_object` is silently dropped. `languages/kotlin/spring-aop.ts`'s member loop also only descends into `function_declaration` children, skipping a `companion_object` child entirely. No fixture anywhere in the new test files uses a Kotlin companion object. Real-world severity is tempered (Spring proxies generally can't intercept `Companion`-dispatched static-like calls anyway), but it's a silent, untested, undocumented gap.

**P2 — Broad/wildcard pointcuts have no total edge or work budget** [STRONG — Claude swarm-security-boundary + Claude ce-adversarial + Codex codex-swarm (C2), independently, both engines] [code-read]
`aop-candidates.ts:97,100` returns the full unfiltered candidate set (`allCandidates`) for unqualified (`*Service`) and leading-wildcard (`*..Service*`) owner patterns — both common, idiomatic AspectJ styles. `pipeline-phases/spring-aop.ts:239-274` (`emitAdvice`) then iterates every candidate per matching advice with no per-advice or aggregate cap, unlike the MCP read path which has `QUERY_RESULT_LIMIT`. The benchmark (`spring-aop-benchmark.test.ts`) structurally avoids stress-testing this exact path: its always-on "10x reduction" tripwire only exercises dotted/prefix-indexable patterns, and the one leading-wildcard case it does include is checked for correctness only, not inspection count. A repo with many broad, generic aspects would pay a real, unbenchmarked O(A×M) cost with no CI signal.

**P2 — `execution(public ...)` silently misses Java interface methods (implicit-public JLS semantics not modeled)** [WEAK — single-engine, Codex `codex-swarm.md` C1] [code-read, independently verified]
Verified: `method-extractors/configs/jvm.ts:136` (`extractVisibility`) defaults an unmodified method to `'package'` visibility via `findVisibility(node, JAVA_VIS, 'package', 'modifiers')`, applied uniformly whether the enclosing body is `class_body` or `interface_body` — Java interface methods, which are implicitly `public` per JLS and conventionally written without an explicit modifier, get stamped `'package'`. `frameworks/spring/aop.ts:658` then requires `method.properties.visibility === 'public'` for `execution(public ...)` pointcuts, so those interface methods never match. This root cause (`jvm.ts`) predates this PR and is outside the diff; the consuming line (`aop.ts:658`) is new and in-diff. Confidence in the underlying claim is high (independently re-derived, not just trusted).

**P2 — Resolved `@Pointcut` named-pointcut declarations are invisible to the MCP `context`/`impact` output** [WEAK — single-engine, Codex `codex-adversarial.md`] [code-read, independently verified]
Verified: `pipeline-phases/spring-aop.ts`'s `pointcutReason` sets `match: resolved ? 'static' : 'unresolved'`, and both a resolved and unresolved `@Pointcut` produce a `DECLARES` evidence edge. But `aop-metadata.ts:248` only surfaces the pointcut branch when `reason.match === 'unresolved'` — a successfully-resolved standalone `@Pointcut("execution(...)") void orderOps(){}` (referenced elsewhere via `@Before("orderOps()")`) is persisted to the graph but silently dropped by the MCP query, with no equivalent "resolved pointcut declaration" field to land in. Asking `context()` about the pointcut method itself returns nothing.

**P3 — `LIMIT 1001` with no `ORDER BY` makes truncated AOP metadata non-deterministic across repeated MCP calls** [WEAK — single-engine, Codex `codex-swarm.md` C3] [code-read, independently verified]
Verified: all 4 parallel queries in `aop-metadata.ts` (lines ~150-186) use `LIMIT 1001` with no `ORDER BY`. When a symbol's edge count exceeds the cap, which 1000 rows come back (and thus which specific advices/behaviors are visible) is engine-dependent and can vary call to call. Low real-world impact (requires >1000 AOP edges on one symbol) but a real correctness gap in the `truncated: true` contract.

## 5. Lower priority, refuted suspicions, and test gaps

**Refuted / downgraded:**
- Codex `codex-ce.md` flagged `analysis-features.ts:37`'s `isSpringAopSourceFile` (hardcoded `.java`/`.kt`/`.kts` check) as a P2 violation of "shared pipeline code must not name languages" (confidence 100). Verified this is **not a new violation**: the identical pattern (`isSpringConditionOrAutoConfigurationFile`, same three extensions) already exists unchanged in the same file, pre-dating this PR (confirmed: zero diff hunk touches that function). At most a P3 "third/fourth copy of this helper" duplication note (also raised independently, correctly, by Claude's maintainability lane as a pure-duplication concern, not a rule violation) — not merge-blocking.
- A Claude lane (ce-correctness) flagged `schema.test.ts`'s `REL_TYPES` test failing locally. Verified this is an **environment artifact of this review session**, not a PR defect: `node -e "console.log(require.resolve('gitnexus-shared'))"` from inside the worktree resolves to `/workspace/gitnexus-shared/dist/index.js` (the main checkout's stale build, confirmed to lack `ADVISED_BY`), not the worktree's own source (confirmed to correctly add `ADVISED_BY` in both `schema-constants.ts` and `types.ts`). Codex's `codex-ce.md` transcript shows the same class of issue independently (an initial `npx tsc --noEmit` failure that self-resolved after a shared-package rebuild) — consistent, not contradictory.

**Not inline-worthy, worth tracking:**
- Dead code: `frameworks/spring/aop.ts:443` exports `isSpringAopDeclaration` with zero call sites anywhere in `src`/`test` (Claude maintainability, single-lane, confidence 90 — verified, zero grep hits).
- `aop.ts` (666 lines) bundles metadata-capture and a self-contained pointcut-matching engine that's already consumed as a separate type by `aop-candidates.ts` — a clean, non-urgent split opportunity (Claude maintainability, single-lane, style suggestion).
- `@annotation(X)` pointcuts only resolve fully-qualified names; the common unqualified style (relying on the aspect file's own import) always fails closed to "unresolved" (Claude ce-correctness, single-lane, confidence 55) — plausibly an intentional v1 scoping choice given the file's consistent fail-closed philosophy elsewhere, but undocumented as such.
- `computeEffectiveWriteSet`'s new `isGraphWideRelationship` guard changes behavior for 4 pre-existing relationship types (INJECTS, TAINT_PATH, CALL_SUMMARY, spring-auto-config DECLARES), not just the new ADVISED_BY — tests pass (16/16, reproduced), but only the new type has a dedicated regression test (Claude ce-correctness, low confidence 40).
- ARCHITECTURE.md is stale: still says "15 phases" (this PR adds a 16th, `springAop`) and its relation-types list ends at `ENTRY_POINT_OF`, missing `ADVISED_BY` [STRONG — Claude docs-dod + Codex codex-swarm D1, independently, both engines; verified live]. Non-blocking, docs-only, and **not touched by this diff** — cannot anchor an inline comment to it (see §6 below for the real in-diff anchor used instead).

**Test gaps** (from Claude's dedicated testing lane, cross-checked, not independently re-verified line-by-line but consistent with the code read above):
- `truncated: true`'s positive branch (≥1001 real matching rows on the *queried* symbol, not a decoy) is never exercised — only the false-positive-avoidance direction is tested.
- Kotlin `@get:`/`@field:` use-site-target fail-closed skip (`aop.ts:298-302`) has zero test coverage in either direction.
- No same-language "true negative" control class in the pipeline fixture — every Java/Kotlin class in `spring-aop-pipeline.test.ts` lives under a pointcut-swept package.
- Several `BEHAVIOR_ANNOTATIONS` entries (`Caching`, `CachePut`, `jakarta`/`javax.transaction.Transactional`, `PostAuthorize`, `PreFilter`, `PostFilter`, `RolesAllowed`) are never exercised through real import/FQN resolution end-to-end, only through the reason encode/decode round-trip.

## 6. CI status

No functional CI test-suite results are visible for this head. `gh pr checks 2783` shows only: `Apply conventional label` (pass), `Validate PR title` (skipping), and `Vercel` (fail — deployment-authorization gate, not a code signal; unrelated to this review). No `CI Gate` or `npm test`/vitest workflow run exists for this commit. Pass/fail of the actual test suite is unknown from GitHub alone.

## 7. Coverage and what wasn't feasible to verify

This is a 43-file, +4185-line PR touching ingestion capture (Java+Kotlin), a new pipeline phase, schema/DDL, incremental persistence, and MCP query surface. Verification in this session:
- Directly re-read and confirmed against source: the schema-version collision (git-level), the interface-propagation gap, the Kotlin companion-object gap, the Java implicit-visibility gap, the unordered-LIMIT gap, the resolved-pointcut MCP gap, the vitest isolation gap, the dead-export, and the stale-ARCHITECTURE.md claim.
- Test execution was inconsistent across lanes due to worktree/build constraints, not because the PR is untested: Claude's `ce-correctness` lane ran unit-level suites directly (26/26, 16/16, 93/94-with-1-known-false-alarm) but could not execute the four large new integration files (`spring-aop-pipeline.test.ts` 773 lines, `spring-aop-benchmark.test.ts` 447 lines, `spring-aop-mcp.test.ts` 402 lines, `incremental-orchestration.test.ts` +207 lines) because this worktree has no built `dist/` for the parse-worker pool. Codex's `codex-ce` leg separately ran 46 focused tests plus a shared-package typecheck successfully. No lane executed the full `npm test` suite end-to-end in this environment; none of that is a defect in the PR's own test design (which multiple lanes independently rated as thorough for the paths it does cover).
- Not independently re-verified: the exact benchmark-threshold looseness claims (10s Kotlin-capture budget), and the deeper LadybugDB-level "undeclared FROM/TO pair silently drops inserts" mechanism cited by one lane as a downstream consequence of the schema collision (asserted at reduced confidence by that lane itself, based on established precedent from PR #2740 rather than re-executed this session).

## 8. Bottom line

- **Branch hygiene:** `rebase/split required` — clean single-commit PR, causally scoped to one feature, but cannot merge without a rebase.
- **Merge state:** `blocked by conflicts` — GitHub's `mergeable: CONFLICTING`/`mergeStateStatus: DIRTY` is accurate (confirmed via corrected `git merge-tree`), not a stale cache.
- **Verdict:** `not production-ready` as submitted. The P0 schema-version collision is a real, git-conflicting blocker with a one-line-per-file fix (rebase, allocate 33, update the test). The P1 interface/superclass advice-propagation gap and the P1 missing vitest native-DB isolation entry are both independently verified and should be fixed in the same pass. Everything else in this review (correctness of the matching engine, incremental-persistence wiring, MCP query design, security posture) held up well against two engines' adversarial passes.

---

*This is an automated multi-tool digest (Claude Sonnet 5 swarm + ce-code-review lanes, Codex gpt-5.6-sol xhigh) synthesized and independently spot-verified against the live worktree. Verify findings before acting — in particular, confirm the exact next-free schema-version integer against `origin/main` at merge time, not at review time.*

gitnexus/src/core/ingestion/pipeline-phases/spring-aop.ts
```diff
@@ -0,0 +1,358 @@
...
+    kind: 'behavior',
+    annotation: record.annotation,
+    behavior: record.behavior,
+    declaredOn: record.ownerKind === 'class' ? 'class' : 'method',
+    activation: 'unknown',
+    proxy: 'possible',
+  });
+
+  const sources =
+    record.ownerKind === 'class' ? [owner, ...(classMethods.get(owner.id) ?? [])] : [owner];
```
magyargergo, line 151: **P1 · STRONG (Claude ce-adversarial + Claude ce-testing + Codex adversarial, independently) · [code-read]**

`@Transactional`/`@PreAuthorig`/etc. declared on an interface method never propagates to the concrete implementing/overriding method — the standard Spring idiom (`interface OrderService { @Transactional void placeOrder(); }` + an unannotated concrete impl) produces **zero** `ADVISED_BY` evidence on the concrete method.

`sources = record.ownerKind === 'class' ? [owner, ...(classMethods.get(owner.id) ?? [])] : [owner]` — for a method-level fact, `sources=[owner]` only. `classMethods` (built ~line 298-316) comes strictly from direct `HAS_METHOD` edges, no `EXTENDS`/`METHOD_OVERRIDES`/`METHOD_IMPLEMENTS` walk. `languages/java/spring-aop.ts:29-56` confirms capture never walks `implements`/`extends` either. Verified against this PR's own 773-line `spring-aop-pipeline.test.ts`, which repeatedly gets close to this scenario but never wires up a concrete class implementing an annotated interface.

**Fix:** fan class/interface-level (and possibly method-level) behavior out through `METHOD_OVERRIDES`/`METHOD_IMPLEMENTS` edges (available post-`mro` phase) to eligible implementers; add a Java+Kotlin regression test asserting the concrete override gets the edge.

gitnexus/src/core/ingestion/languages/kotlin/captures.ts
```diff
@@ -119,6 +123,17 @@ export function emitKotlinScopeCaptures(
     }
     if (Object.keys(grouped).length === 0) continue;
 
+    // tree-sitter-kotlin represents both classes and interfaces with
+    // `class_declaration`; `object_declaration` is the separate object form.
+    const springAopTypeNode = [
+      nodeIfType(groupedNodes['@scope.class'], 'class_declaration'),
```
magyargergo, line 129: **P2 · STRONG (Claude ce-adversarial + Codex ce-code-review, independently) · [code-read]**

Kotlin `companion object` bodies are entirely excluded from Spring AOP capture. `query.ts:18` tags `companion_object` with the same `@scope.class` capture as `class_declaration`/`object_declaration`, but this filter only accepts `class_declaration`/`object_declaration` — `companion_object` is silently dropped. The `kotlin/spring-aop.ts` member loop also only visits `function_declaration` children. No fixture in the new test files uses a Kotlin companion object.

Real-world severity is tempered (Spring proxies generally can't intercept `Companion`-dispatched calls anyway), but it's a silent, untested, undocumented gap.

**Fix:** include `companion_object`, mark it as a singleton-instance owner (same treatment as `object_declaration`); add an integration test mirroring the existing Kotlin-object test.

gitnexus/src/core/ingestion/frameworks/spring/aop.ts
```diff
@@ -0,0 +1,666 @@
...
+  method: GraphNode,
+  methodAnnotations: ReadonlySet<string> = new Set(),
+): boolean {
+  if (method.label !== 'Method') return false;
+  if (pointcut.kind === 'annotation') return methodAnnotations.has(pointcut.annotation);
+  const qualifiedName = owner.properties.qualifiedName;
+  if (typeof qualifiedName !== 'string') return false;
+  if (!typePatternMatches(pointcut.ownerPattern, qualifiedName)) return false;
+  if (pointcut.kind === 'within') return true;
+  if (pointcut.visibility === 'public' && method.properties.visibility !== 'public') return false;
```
magyargergo: **P2 · WEAK (single-engine, Codex swarm C1) but independently re-verified · [code-read]**

`execution(public ...)` pointcuts silently miss Java interface methods. `method-extractors/configs/jvm.ts:136` (`extractVisibility`, pre-existing, not touched by this diff) defaults an unmodified method to `'package'` visibility uniformly for `class_body` and `interface_body` — but Java interface methods are implicitly `public` per JLS and conventionally written without an explicit modifier. This line then requires `method.properties.visibility === 'public'`, so those interface methods never match.

**Fix:** normalize Java interface-method visibility to `public` at extraction (jvm.ts), or special-case interface owners in this AOP visibility check; add an `execution(public * ...)` interface test.

gitnexus/src/mcp/local/aop-metadata.ts
```diff
@@ -0,0 +1,306 @@
...
+        if (row === undefined) continue;
+        const reason = decodeSpringAopReason(row.reason);
+        if (reason?.kind === 'aspect') {
+          aspects.push({
+            annotation: reason.annotation,
+            activation: reason.activation,
+            registration: reason.registration,
+            evidenceId: row.targetId,
+          });
+        } else if (reason?.kind === 'pointcut' && reason.match === 'unresolved') {
```
magyargergo, line 284: **P2 · WEAK (single-engine, Codex adversarial) but independently re-verified · [code-read]**

A successfully-resolved standalone `@Pointcut(...)` named-pointcut declaration is persisted as a `DECLARES` evidence edge but never surfaced here — this branch only fires when `reason.match === 'unresolved'`; resolved pointcuts fall through silently with no equivalent field to land in. Asking `context()`/`impact()` about the pointcut-declaring method itself returns nothing.

**Fix:** add a field/branch for resolved standalone pointcut declarations.

gitnexus/test/integration/spring-aop-mcp.test.ts
```diff
@@ -0,0 +1,402 @@
...
+    (id) =>
+      `MATCH (m:Method {id:'${id}'}), (a:Method {id:'${HIGH_FAN_IN_ADVICE_ID}'}) CREATE (m)-[:CodeRelation {type:'ADVISED_BY', confidence:0.9, reason:'${ADVICE_REASON}', step:0}]->(a)`,
+  ),
+  ...HIGH_FAN_IN_CALLER_IDS.map(
+    (id) =>
+      `MATCH (m:Method {id:'${id}'}), (a:Method {id:'${HIGH_FAN_IN_ADVICE_ID}'}) CREATE (m)-[:CodeRelation {type:'CALLS', confidence:1.0, reason:'test fixture', step:0}]->(a)`,
+  ),
+];
+
+withTestLbugDB(
```
magyargergo, line 159: **P1 · WEAK by lane count (Codex-only, adversarial pass) but not demoted — concrete, verified trigger · [code-read]**

This suite opens a real LadybugDB via `withTestLbugDB`, but `gitnexus/vitest.config.ts` (not touched by this diff) doesn't list this file in the `lbug-db` project's sequential-isolation `include` list, nor in the `default` project's `exclude` list — verified by enumerating both lists directly. It will run in the parallel default project alongside other native-addon tests, risking the file-lock/N-API-teardown crashes this repo's isolation convention exists to prevent (especially Windows/macOS CI).

**Fix:** add `test/integration/spring-aop-mcp.test.ts` to both the `lbug-db` include list and the `default` project's exclude list before merging.

gitnexus/src/core/ingestion/pipeline-phases/spring-aop.ts
```diff
@@ -0,0 +1,358 @@
...
+    reason: encodeSpringAopReason(pointcutReason(record, resolved)),
+  });
+
+  if (!isAdviceMethod || !resolved || parsed === null || staticPointcut === null) {
+    return { edges: 0, evidence: 1, unresolved: resolved ? 0 : 1 };
+  }
+
+  let edges = 0;
+  for (const candidate of candidateIndex.candidatesFor(parsed)) {
+    if (candidate.method.id === adviceNode.id) continue;
```
magyargergo, line 297: **P2 · STRONG (Claude security-boundary + Claude ce-adversarial + Codex swarm C2, independently, both engines) · [code-read]**

Unqualified (`*Service`) and leading-wildcard (`*..Service*`) AspectJ owner patterns — common, idiomatic styles — cause `aop-candidates.ts:97,100` to fall back to the *entire* unfiltered candidate set (`allCandidates`). This loop then iterates every candidate per matching advice with no per-advice or aggregate cap, unlike the MCP read path's `QUERY_RESULT_LIMIT`. The benchmark suite structurally avoids stress-testing this exact combination (its tripwire only checks correctness on the one leading-wildcard case it includes, not inspection count).

**Fix:** add a configurable per-advice/aggregate match-edge budget with a logged/`truncated`-style signal when exceeded; add a benchmark case combining multiple broad-wildcard advices.

azizur100389:
Reviewed current head `cb6cdd75` against current `upstream/main` `99291891`.

This is not production-ready yet:

1. The branch has real merge conflicts in `repo-manager.ts` and `call-summary-schema-version.test.ts`. More importantly, the conflict exposes a schema-version collision: this PR assigns `INCREMENTAL_SCHEMA_VERSION = 31`, while main has already shipped v31 and is now at v32. Reusing 31 can let an index created under main's unrelated v31 pass the equality gate without rebuilding for the new AOP relation schema. Please rebase, allocate the next free version (currently 33), and update the associated test/comment.
2. Method-level behavior declared on an interface/superclass is not propagated to the concrete implementing/overriding method. `emitBehavior` only uses the record owner, while class fan-out only walks direct `HAS_METHOD` edges. A common `@Transactional` interface method implemented by an unannotated concrete service can therefore emit no concrete-method evidence. Please add inheritance/override propagation and an end-to-end regression fixture.
3. `spring-aop-mcp.test.ts` opens a real LadybugDB via `withTestLbugDB`, but it is absent from the sequential `lbug-db` Vitest project and from the default project's exclusions. That leaves it running in the parallel default pool despite the repository's native-addon isolation requirement.

Local validation on the PR head passed `npx tsc --noEmit` and 57 focused tests across Spring AOP, incremental subgraph extraction, analysis features, and annotation arguments. I could not validate the actual merged tree because the two conflicts prevent constructing it. GitHub also has no functional CI result for this head; only the label check ran, while Vercel is an unrelated authorization failure.

Recommendation: rebase and request changes for the three items above, then rerun the full AOP/incremental/LadybugDB matrix.

PR #2784, opened 2026-08-01
fix(group): preserve manifest-only impact crossings

### LINKED ISSUE #2722: group sync: manifest links never resolve — multi-label MATCH rejected by the Cypher parser
**Title:** `group sync`: manifest links never resolve — multi-label `MATCH` in manifest-extractor is rejected by the Cypher parser

### What happens

Every manifest-declared link in `group.yaml` logs `resolveSymbol failed` and falls back to a synthetic `manifest::<repo>::<contract>` uid. The link is written to `contracts.json` and reported as a cross-link, but `group impact` never traverses it — so the feature looks like it worked and behaves like a no-op.

```
[manifest-extractor] resolveSymbol failed for custom:executeAddDynamicLinkMS in backend:
  Prepare failed: Parser exception: Invalid input
  <MATCH (n:Function|Method|Class|Interface|Struct|Enum|Trait|Constructor|TypeAlias|Impl|Macro>:
  expected rule oC_SingleQuery (line: 1, offset: 86)
  "MATCH (n:Function|Method|Class|Interface|Struct|Enum|Trait|Constructor|TypeAlias|Impl|Macro|Union|Typedef|Property|Record|Delegate|Annotation|Template|Const|Static|CodeElement)"
                                                                                         ^^^^^
```

### Where

`dist/core/group/extractors/manifest-extractor.js`

- line ~179 builds `MATCH (n:Function|Method|Class|Interface)` — 4 labels, parses fine
- line ~239 builds the same query with 20 labels — the parser rejects it at offset 86, right where the list grows past the shorter form

So the fallback path that is supposed to widen the search is the one that always throws.

### Reproduce

Two indexed repos in a group, then any manifest link whose `contract` is a symbol that provably exists:

```yaml
# ~/.gitnexus/groups/<name>/group.yaml
links:
  - from: app
    to: backend
    type: custom
    contract: executeAddDynamicLinkMS   # verified present:
    role: consumer                      #   MATCH (f:Function) WHERE f.name = 'executeAddDynamicLinkMS'
  - from: backend                       #   -> 1 row, filePath .../https-calls/addDynamicLinkMS.js
    to: app
    type: custom
    contract: executeAddDynamicLinkMS
    role: provider
```

```
$ gitnexus group sync <name>
  ... resolveSymbol failed (both repos)
  Wrote contracts.json (34 contracts, 1 cross-links)

$ gitnexus group impact <name> --repo backend --target executeAddDynamicLinkMS
  Group impact: risk=LOW   direct=1 processes=0 cross=0
```

`cross=0` despite `contracts.json` reporting the cross-link.

### Expected

`resolveSymbol` binds the contract to the real symbol uid, and `group impact` reports the cross-repo consumer.

### Environment

gitnexus 1.6.7, macOS 15 (arm64), Node 22.23.0. Both member repos indexed and up to date with HEAD.

azizur100389:
## Summary

- preserve a proven manifest cross-repository hit when the far endpoint has only a deterministic `manifest::` UID
- avoid calling `impactByUid` with identifiers that cannot exist in a repository graph
- retain normal fan-out and truncation behavior for concrete graph UIDs
- add a regression test matching the string-dispatch scenario from #2722

## Test plan

- [x] `npx vitest run test/unit/group/manifest-synthetic-impact.test.ts test/unit/group/cross-impact.test.ts test/unit/group/resolve-bridge-neighbors.test.ts test/unit/group/manifest-extractor.test.ts` (36 passed, 3 platform skips)
- [x] `npx tsc --noEmit`
- [x] targeted ESLint and Prettier checks
- [x] `npm run build`
- [x] full `npm test` attempted locally (15,204 passed, 413 skipped; 75 unrelated Windows/runtime/timing failures across 27 files under Node 22.17.1, below the repository minimum 22.18)
- [x] final GitNexus index current at commit `5d9588fe`; expected group-impact execution flows only

Fixes #2722


magyargergo (commented):
## Multi-engine review of #2784

**How this was reviewed.** Two independent engines ran the same scaffolding against BASE `9929189` → HEAD `5d9588f`:

> **Head moved mid-review.** You pushed `623d2ff` ("fix(group): verify manifest-only neighbor repos") while the lanes were running. Everything below has been **re-verified and re-anchored against `623d2ff`**; one finding was dropped because that commit already fixes it (see the credit below). The reproductions were executed against `5d9588f`, and each was re-confirmed by reading `623d2ff` — the branch body is byte-identical there apart from three words of comment.

- **Engine A (Claude)** — 5 GitNexus swarm lanes (pr-facts-historian, risk-architect, test-ci-verifier, security-boundary, docs-dod) + 6 `ce-code-review` personas (correctness, adversarial, testing, maintainability, project-standards, api-contract). The **ce-adversarial lane executed the PR HEAD code against a real `bridge.lbug`** built with the production `writeBridge` — no `vi.mock`, no source edits. Findings it reproduced end-to-end are tagged `[reproduced]` below and are the strongest evidence class in this digest.
- **Engine B (Codex, `gpt-5.6-sol`, effort xhigh)** — three legs, all live: the GitNexus PR swarm (verdict *not production-ready*), `ce-code-review` report-only (verdict *Ready with fixes*, one P1 at confidence 100), and native adversarial-review (verdict *needs-attention / No-ship*). One caveat B2 self-reported: its bundled cross-model pass could not start in the restricted runtime, so it fell back to a local adversarial lane. It did run `npx vitest` (39 focused tests) and `npx tsc --noEmit` green.

Every finding below was re-verified against the worktree before inclusion.

---

### What this PR gets right

- **It fixes a real, reporter-confirmed defect.** Both engines' history lanes independently reconstructed #2722 and reached the same conclusion: the issue has **two** causes. Cause 1 (the multi-label Cypher parse failure) was fixed by e148bc089 / #2327 and shipped in v1.6.9. Cause 2 is what this PR fixes — reporter xFlaviews re-tested on **v1.6.10-rc.119**, confirmed the parser exception was gone, and still saw `cross=0`: *"Two separate causes then — the Cypher error was hiding a second one in the join, not being it."* **`Fixes #2722` is accurate for the residual symptom, and the auto-close is process-appropriate.** One nuance worth a line in the PR body: the issue's own *Expected* text says *"`resolveSymbol` binds the contract to the real symbol uid"*, which this PR deliberately does not do — it preserves the crossing without resolving the uid. Since the issue title also names the already-fixed parser bug, a sentence distinguishing the two causes would save a future reader from thinking the PR is redundant. Non-blocking.
- **The diagnosis is correct and the guarded path is live on `main`.** `impactByUid` on a `manifest::…` uid can never match (query `MATCH (n) WHERE n.id = $uid` at `local-backend.ts:6558`, null-return at `:6566`), and that row previously fell into `truncatedRepos`. Confirmed by both engines and reproduced by control scenario A2.
- **CI is fully green** — 16,005 passed / 0 failed / 70 skipped across Ubuntu, Windows and macOS shards, plus CodeQL (JS-TS + Python), Gitleaks, Dependency Review, typecheck, format, lint, tree-sitter ABI, and Node floor compat.
- **The security surface is clean.** Engine A's security-boundary lane and Engine B's lane 5 converged: no new command/path/query/serialization surface, `git diff --check` clean, no bidi or suspicious non-ASCII, and the prefix discriminant is safe.
- **The new test genuinely pins the changed branch.** Delete lines 635-649 and control reaches `safeNeighborImpact`, `impactByUid` returns `null`, and the `cross_repo_hits` assertion (now `:114`) fails. It is not a tautological test.
- **`623d2ff` already closed one of our findings, unprompted, with a regression test.** The review found that the branch `continue`d before the `resolveRepo` liveness check, so a neighbor repo that could not be opened was still asserted as a proven crossing — reproduced by Engine A and independently named by Engine B's adversarial leg ("syncing while the neighbor repository is unavailable… this path later reports it as successfully analyzed"). Your commit hoists the `resolveRepo` try/catch to `:616-622`, above the branch, which is exactly the recommended fix, and adds `'keeps an unavailable synthetic neighbor truncated instead of reporting a hit'` asserting `cross: []`, `cross_repo_hits: 0`, `truncated: true`, `truncatedRepos: ['app']`. **That finding is dropped from this review as fixed.** The first test also gained `expect(resolveRepo).toHaveBeenCalledWith('app-registry')`.
- **Neither prior hardening is regressed.** #1331's `safeNeighborImpact` timeout (CodeQL #184) is intact — the new path is synchronous with nothing to cancel — and #2313's cached read-only bridge handle still opens and closes through the same `finally`.

---

### Headline finding: a boundary-only crossing is emitted as a *completed* analysis, and it drives the risk gate

**Severity: P1.** Existence is unanimous across both engines and every lane; only the number varied — A risk-architect said P0, A docs-dod / A api-contract / B2 said P1, A historian / B1 / B3 said P2, A ce-correctness said P2/P3. I am posting it as **P1**, and the spread is why:

- It is **not P2**, because it was reproduced end-to-end against a real bridge, and the field it corrupts (`risk`) is a documented automation gate in this repo's own guidance — *"MUST warn the user if impact analysis returns HIGH or CRITICAL"* / *"NEVER ignore HIGH or CRITICAL risk warnings"*. A wrong value there has a concrete downstream consequence for every agent and CI script reading `group impact`.
- It is **not P0**, because the error direction is conservative. Pre-PR the same input produced a false negative (`cross=0`); post-PR it produces a false positive (`cross=1, risk=HIGH`). Neither hides a real blast radius. This file's own comment at `:519-522` argues a blast-radius tool reporting "no impact" on a failure path is the worse of the two.

**Trigger (reproduced, real `bridge.lbug`, no mocks):** a group with one manifest link whose provider side resolves to a real uid and whose consumer side does not — the documented remedy for a boundary the extractors miss, and exactly #2722's shape. `group impact --repo backend --target executeAddDynamicLinkMS --direction upstream` with `localRisk = LOW`:

| | `cross` | `by_depth` | `truncated` | `truncatedRepos` | `risk` |
|---|---|---|---|---|---|
| BASE (control A2) | 0 | — | **true** | `["app"]` | `LOW` |
| HEAD (A) | 1 | `{}` | **false** | `[]` | **`HIGH`** |
| HEAD, 3 links (B) | 3 | `{}` | false | `[]` | **`CRITICAL`** |

**Evidence (confirmed in worktree).** `mergeRisk` (`cross-impact.ts:351-358`) never inspects `by_depth` or `affected_processes` — `highConf = cross.some(c => c.contract.confidence >= 0.85)`, and `cross.length >= 3` returns `CRITICAL` *ahead of* the confidence check. Manifest links are minted with `confidence: 1.0` unconditionally (`manifest-extractor.ts:160`, `:172`, `:183`), so `highConf` is true for **every** entry this branch produces. Independently confirmed by direct invocation: `LOW + 1 hollow → HIGH`, `LOW + 3 hollow → CRITICAL`, `UNKNOWN + 1 → HIGH`.

**The second half is the missing discriminator.** `CrossRepoImpact` (`types.ts:169-180`) is a closed shape with no field for "fan-out was never attempted". A real neighbor uid whose fan-out is genuinely empty produces `by_depth: {}` / `affected_processes: []` too — byte-identical. `match_type` does not disambiguate (see the P3 below, where a hollow entry can even carry `'exact'`). The default CLI prints only `risk=… cross=N` (`cli/group.ts:252-256`); `--json` and the MCP `impact` tool return the raw payload (`local-backend.ts:6733`). Note this repo already ships the convention for exactly this ambiguity one layer down — `mcp/tools.ts:450`: *"When `partial:true`, do NOT treat `processes:[]` as proof of no participation"* — and this PR sets the equivalent flag to `false`.

Both engines proposed the same shape of fix, and it closes the risk escalation and the ambiguity together: **add an optional discriminator to `CrossRepoImpact`** (`fanout_status: 'completed' | 'not_attempted'`, or `unresolved?: true` — additive and backward-compatible), set it in the new branch, and have `mergeRisk` count only entries with a real fan-out toward `highConf` and `cross.length >= 3`. Whether a declared-but-unwalkable boundary *should* still escalate is a legitimate product call — B1 and A historian both note conservative escalation is defensible. The objection is that it is currently an **untested side effect** rather than a decision: neither the original test nor the one added in `623d2ff` asserts `result.risk`.

---

### Inline findings

Seven comments are attached to the diff, all anchored on `623d2ff`. Summary:

| # | Location | Sev | Problem | Tag |
|---|---|---|---|---|
| 1 | `cross-impact.ts:643` | P1 | Hollow crossing's `confidence: 1` drives `mergeRisk` LOW→HIGH, three → CRITICAL | `[reproduced]` |
| 2 | `cross-impact.ts:645` | P1 | No discriminator; `truncated` flips `true`→`false` for the same input | `[reproduced]` |
| 3 | `cross-impact.ts:624` | P2 | Fix is inert under `--service`: the pre-existing prefix filter drops manifest endpoints (empty `filePath`) before the branch — *raised from both source lanes' P3, rationale in the comment* | `[reproduced]` |
| 4 | `cross-impact.ts:642` | P3 | A hollow entry can be labelled `match_type: 'exact'`, contradicting the branch's own comment | `[reproduced]` |
| 5 | `manifest-synthetic-impact.test.ts:128` | P2 | The one behavior that changed — `result.risk` — is still never asserted, including in the new test | `[reproduced]` |
| 6 | `manifest-synthetic-impact.test.ts:21` | P2 | Single-row mock; no mixed real+synthetic neighbor case | `[code-read]` |
| 7 | `manifest-synthetic-impact.test.ts:15` | P2 | Mock-only coverage vs. DoD §2.7 "integration tests hit a real database where the production path does" | `[code-read]` |

*(An eighth finding — the missing `resolveRepo` liveness check — was dropped: `623d2ff` fixes it.)*

Findings 1-4 are on lines the diff added; #3 and #4 name their real upstream locations (`:590` and the `runExactMatch` producer) in the comment body since those lines are outside the diff.

---

### Lower-priority items and open questions

- **Deadline ordering (P3 — the two engines disagreed; here is the adjudication).** The `remainingMs <= 0` truncation gate at `:607-611` runs well above the new branch (`:635` on `623d2ff`), so a zero-I/O manifest crossing can still be discarded and mis-reported as a timeout truncation (reproduced: `timeoutMs=100`, local phase 95 ms → `cross=0, truncated=true`). A risk-architect rated it P2; **B1 explicitly declined it** on the grounds that the result is then marked truncated, which is conservative and honest. B1 is right that this is not a correctness hole. But the consequence A ce-adversarial identified stands: the #2722 symptom returns non-deterministically under a tight budget or a slow Phase 1 — i.e. on the large repos where the feature matters most — and it makes the new behavior wall-clock dependent. Net: **P3, worth the one-line hoist** of the manifest branch above the deadline gate at `:607-611` — the same move you just made for `resolveRepo`, but not merge-blocking.
- **`cross[]` has no length cap and the new path pays no budget (P3).** 500 manifest-only crossings were produced in 42 ms under the *minimum* 100 ms timeout — 87 KB of zero-information rows with `truncated:false`, pinned at CRITICAL. A's security-boundary lane correctly notes the missing cap is pre-existing (the real-UID push at `:674` has none either) and manifests are small hand-authored YAML, so this is not a resource-exhaustion vector — but it is a context-window and signal-quality concern that compounds finding #1. Collapsing per `(repo_path)` into one entry with a `contracts: string[]` list would address both.
- **Stale comment in `manifest-extractor.ts:205-210` (P3, pre-existing, both engines).** It claims *"Cross-impact still works: the bridge query joins on the synthetic uid, and the local impact engine derives the same uid."* That is false and now provably so: `collectImpactSymbolUids` (`cross-impact.ts:309-339`) harvests only `target.id` and `byDepth[].id` — always real `generateId` node ids — so a `manifest::` uid can never reach the local side of the join. The PR's own new comment at `:624-634` states the opposite and is correct; the two now contradict each other. The practical consequence: **this fix is one-sided.** The mirror query — running `group impact` from the repo whose *own* endpoint is unresolved — still returns `cross: 0`. For #2722's shape, the backend-anchored query is fixed and the app-anchored one is not. Worth a sentence in the PR body.
- **`'manifest::'` is hardcoded as a literal (P3).** `manifest-extractor.ts:93` exports the minting function `manifestSymbolUid` but no matching predicate. `cross-impact.ts:635` is now the only production site outside the minting module that re-derives the format by literal, and `cross-impact.ts` imports nothing from `manifest-extractor.ts`, so nothing forces the two to change together. A one-line exported `isManifestSymbolUid()` would close it.
- **Docs (P3).** DoD §2.4 and CONTRIBUTING require docs updates when user-visible behavior changes; `ARCHITECTURE.md` §group and the `impact` tool description say nothing about manifest-only crossings or what empty `by_depth` means for them. CHANGELOG correctly untouched (DoD §4.5 assigns it to the release process).

### Suspicions the review refuted — recorded so they are not re-raised

- **`manifest::` colliding with a real graph id — refuted.** `generateId` is `` `${label}:${name}` `` (`lib/utils.ts:1`), label-first and PascalCase across all of `SYMBOL_NODE_LABELS`. `manifest::` is lowercase with a doubled colon; no ingestion path can produce it and no `group.yaml` value reaches the id side of a real symbol.
- **The `as MatchType` cast being unsound — refuted.** `'manifest'` is a real member of the union (`types.ts:2`), already produced by `manifest-extractor.ts:182` and already surfaced through `trace`. The cast pair is also byte-identical to the pre-existing one at `:679-680`, so this PR duplicates an existing pattern rather than introducing one. `tsc --noEmit` clean on both engines.
- **`safeNeighborImpact` / CodeQL #184 regression — refuted.** The new path is synchronous; there is nothing async for the timeout to cancel, and real UIDs still take the wrapped path.
- **Dead code — refuted.** The guarded path is reachable on current `main`; control A2 reproduces the pre-fix behavior on HEAD's own code.
- **Service-scope leak — refuted.** The opposite is true: an empty `filePath` fails every prefix rather than matching all of them (finding #3).
- **`bm25`/`embedding` match-type laundering — refuted.** Both are declared in `MatchType` but a repo-wide grep finds **no producer** for either under `src/`. Only `manifest`, `exact` and `wildcard` are emitted. (Worth noting `wildcard` carries confidence below 0.85 yet still reaches `cross.length >= 3 → CRITICAL`.)
- **"This PR fixes an already-fixed bug" — refuted**, see the opening section.
- **`if ('error' in result) return;` at test line 112 — refuted as a rule violation.** Two Engine A lanes flagged it as a style violation; two others checked the repo's written standards and found no rule barring it, and the identical narrowing idiom already appears six times in the sibling `test/unit/group/cross-impact.test.ts` (lines 141-142, 183-184, 228-229, 273-274) for the same discriminated-union problem. Verified. Not raised as a finding. If it is ever cleaned up, do it as a shared `assertImpactOk()` helper across both files rather than in this PR.
- **Workflow-file changes and "the branch skips subgroup/servicePrefix filtering"** — one Engine A lane emitted both; both were already marked refuted in its own artifact (the diff touches exactly two files, and the subgroup/servicePrefix checks run at `:590`/`:593`, *before* the new branch). Not propagated.

### Test gaps not covered by the inline comments

`direction: 'downstream'` is never exercised through the new branch; nothing pins the `truncated: true → false` transition that is the user-visible contract change; nothing covers the deadline interaction; and `truncatedRepos`/`outOfScope` contents are unasserted.

### CI

All GitHub Actions checks pass at HEAD. The single red check is **Vercel — "Authorization required to deploy"**, a fork-authorization blocker for an external contributor, not a code signal. Both engines classified it the same way.

---

### Classification

- **Branch hygiene: `clean feature/fix PR`** — one commit, two files, no unrelated churn, no merge-from-main noise. Both engines' hygiene lanes concur. It sits one commit behind `origin/main` (#2783, Spring AOP), which touches no file in this diff; a rebase or merge would be clean.
- **Merge state: `mergeable`** — no conflicts. The one red check is Vercel's fork-authorization gate, not a code signal.
- **Final verdict: `not production-ready`**

  Both engines returned this independently, and the P1 was reproduced end-to-end. The PR correctly solves the residual #2722 crossing failure and the implementation is structurally sound — the guard is in the right function, it preserves the concrete-UID path unchanged, and it regresses neither #1331's timeout nor #2313's bridge-handle caching. What holds it back is that a manifest-only crossing now escalates `group impact` risk from LOW to HIGH — and to CRITICAL at three links — with zero demonstrated fan-out, and `CrossRepoImpact` carries no field distinguishing a boundary-only entry from a fully-traversed empty one. That ambiguity feeds a value this repo's own guidance treats as an automation gate, and the same input now reports `truncated: false` where it previously reported `true`. The risk escalation is an untested side effect rather than a stated decision: the new spec never asserts `result.risk`. Add the discriminator, gate `mergeRisk` on it, and pin the risk semantics with an assertion — the change is small and the rest of the PR is ready.

---

*This is an automated multi-engine review digest (Claude swarm + ce personas; Codex `gpt-5.6-sol` swarm + ce + adversarial). Findings tagged `[reproduced]` were executed end-to-end against a real bridge database; `[code-read]` findings were verified by reading the code only. Please verify before acting on any individual item.*

gitnexus/src/core/group/cross-impact.ts
```diff
@@ -621,6 +621,33 @@ export async function runGroupImpact(
...
+      // truncations rather than false successful traversals.
+      if (n.neighborUid.startsWith('manifest::')) {
+        cross.push({
+          repo: regName,
+          repo_path: n.neighborRepo,
+          contract: {
+            id: n.contractId,
+            type: n.contractType as ContractType,
+            match_type: (n.matchType as MatchType) || 'manifest',
+            confidence: n.confidence,
```
magyargergo, line 664: **P1 — a crossing with zero demonstrated fan-out escalates group risk to HIGH, and three of them to CRITICAL.**

Manifest links are minted with `confidence: 1.0` unconditionally (`extractors/manifest-extractor.ts:160`, `:172`, `:183`), so `n.confidence` here is always `1`. `mergeRisk` (this file, `:351-358` — outside the diff) never inspects `by_depth` or `affected_processes`:

```ts
const highConf = cross.some((c) => c.contract.confidence >= 0.85);
if (cross.length >= 3) return 'CRITICAL';
if (highConf) return 'HIGH';
```

**Trigger** (reproduced end-to-end against a real `bridge.lbug`, no mocks): one manifest link whose provider side resolves and whose consumer side does not — #2722's exact shape. With `localRisk = 'LOW'`:

| | cross | truncated | risk |
|---|---|---|---|
| BASE | 0 | true | `LOW` |
| HEAD | 1 | false | **`HIGH`** |
| HEAD, 3 links | 3 | false | **`CRITICAL`** |

Confirmed directly against the exported function: `LOW + 1 hollow → HIGH`, `LOW + 3 hollow → CRITICAL`, `UNKNOWN + 1 → HIGH`.

This matters because `risk` is a documented automation gate in this repo's own guidance ("MUST warn the user if impact analysis returns HIGH or CRITICAL" / "NEVER ignore HIGH or CRITICAL risk warnings"), consumed by the CLI (`src/cli/group.ts:252`) and by the MCP `impact` tool via `@group` routing (`src/mcp/local/local-backend.ts:6733`). The cascade is the awkward part: the more manifest links an operator declares — the documented remedy for a boundary the extractors miss — the more hollow crossings every `group impact` in that area emits, until three saturate the signal at CRITICAL permanently.

**Suggested fix:** pair with the discriminator below and have `mergeRisk` count only entries with an attempted fan-out — `cross.some(c => c.fanout_status !== 'not_attempted' && c.contract.confidence >= 0.85)` and the same filter on the `>= 3` count — or cap a boundary-only crossing's contribution at `MEDIUM`. Conservative escalation may well be the right product call; the objection is that it is currently an untested side effect rather than a decision. `mergeRisk` is already exported and unit-tested, so pinning it is cheap.

*Cross-engine: raised independently by both engines and every lane; severity ranged P0–P2 across lanes, adjudicated to P1.*

[reproduced]

gitnexus/src/core/group/cross-impact.ts
```diff
@@ -621,6 +621,33 @@ export async function runGroupImpact(
...
+          },
+          by_depth: {},
```
magyargergo, line 666: **P1 — `by_depth: {}` / `affected_processes: []` with `truncated: false` asserts a *completed* analysis that never happened.**

`CrossRepoImpact` (`types.ts:169-180`, outside the diff) has no field distinguishing "fan-out was not attempted because the endpoint has no graph symbol" from "fan-out ran and found nothing". The real-UID path 40 lines below (`:683-684`) emits byte-identical values when a neighbor genuinely has no dependents. `match_type` does not disambiguate either — it describes the *linking* method, and a resolved manifest link carries `'manifest'` too.

Worse, the same input flips `truncated` from `true` to `false` (reproduced, real bridge):

| | `cross` | `truncated` | `truncatedRepos` | CLI shows |
|---|---|---|---|---|
| BASE | 0 | **true** | `["app"]` | `risk=LOW … cross=0` |
| HEAD | 1 | **false** | `[]` | `risk=HIGH … cross=1` |

The PR trades a *loud* miss for a *quiet* positive assertion. Pre-PR, `cross=0` came with `truncated:true` — a visible "I could not finish" marker a JSON/MCP consumer could act on. Post-PR there is no marker at all: the default CLI prints only `risk=… cross=N` (`src/cli/group.ts:252-256`) and `--json`/MCP return the raw payload (`local-backend.ts:6733`).

This codebase already ships the convention for exactly this ambiguity one layer down — `src/mcp/tools.ts:450`: *"When `partial:true`, do NOT treat `processes:[]` as proof of no participation."* This PR sets the equivalent flag to `false`.

**Suggested fix:** add an optional, additive discriminator to `CrossRepoImpact` — `fanout_status?: 'completed' | 'not_attempted'` (or `unresolved?: true`) — set it only in this branch, render it in `src/cli/group.ts` (`cross=2 (1 boundary-only)`), and document it in the `impact` group-mode tool text. Backward-compatible for existing consumers, and it makes the `mergeRisk` gate above a one-line filter.

*Cross-engine: Engine B's ce leg rates this P1 at confidence 100 and its adversarial leg calls it no-ship; Engine A's risk-architect, api-contract, correctness, docs-dod and adversarial lanes all reached it independently.*

[reproduced]

gitnexus/src/core/group/cross-impact.ts
```diff
@@ -621,6 +621,33 @@ export async function runGroupImpact(
         continue;
       }
 
+      // A manifest link can prove the repository boundary even when its far
```
magyargergo, line 645: **P2 — the fix does not apply to service-scoped queries: `--service` / `service:` still drops every manifest-only crossing, silently.**

The real location is the pre-existing filter at `:590-592`, which runs before this branch:

```ts
if (servicePrefix && !fileMatchesServicePrefix(n.neighborFilePath, servicePrefix)) continue;
```

`fileMatchesServicePrefix` (`group-path-utils.ts:24`) is `if (!filePath) return false;` — an empty path fails *every* prefix. And a synthetic manifest endpoint's `filePath` is **always** empty by construction (`extractors/manifest-extractor.ts:146-147`: `providerSymbol || { filePath: '', name: link.contract }`), persisted as-is by `bridge-db.ts:832`.

**Trigger** (reproduced by the Engine A adversarial lane via two scenarios — G1 no service flag, G2 `--service src` against the same bridge):

```
no --service   → risk=HIGH cross=1 truncated=false
--service src  → risk=LOW  cross=0 truncated=false outOfScope=0 truncatedRepos=[]
```

The crossing vanishes with **no signal at all** — not in `cross`, not in `outOfScope`, not in `truncatedRepos`. So the PR's own premise ("`group sync` reports the manifest link while `group impact` silently returns cross=0") remains true verbatim for every monorepo user who passes the documented `service` scoping parameter (`src/mcp/tools.ts:468`). Note the direction here is the *safe* one — the opposite hypothesis, that `''` would match every prefix and leak the crossing into every service scope, was tested and refuted.

**Suggested fix:** treat an absent neighbor `filePath` as "scope unknown" rather than "out of scope" — either keep the crossing (carrying the discriminator from the finding above) or route it into `outOfScope` so it is at least visible — and add a `service`-set case to the new spec. If the drop is intentional, assert and document it rather than leaving it untested.

**On severity:** both source lanes rated this P3 (adversarial P3/confidence 90, correctness P3/confidence 60). Raised to **P2** here because `service` is the documented monorepo scoping parameter (`src/mcp/tools.ts:468`), so a fix that is silently inert under it fails for a first-class documented use case while the PR's stated premise stays literally true for those users. Revert to P3 if you consider service-scoped group impact a secondary path.

*Engine A only. Reproduced by the adversarial lane via two scenarios (G1: no service flag; G2: `--service src` against the same bridge); the correctness lane independently confirmed the mechanism by code-read (P3, confidence 60). No Engine B lane examined the service-scoped path.*

[reproduced]

gitnexus/src/core/group/cross-impact.ts
```diff
@@ -621,6 +621,33 @@ export async function runGroupImpact(
...
+      // still take the normal impactByUid path below, where failures remain
+      // truncations rather than false successful traversals.
+      if (n.neighborUid.startsWith('manifest::')) {
+        cross.push({
+          repo: regName,
+          repo_path: n.neighborRepo,
+          contract: {
+            id: n.contractId,
+            type: n.contractType as ContractType,
+            match_type: (n.matchType as MatchType) || 'manifest',
```
magyargergo: **P3 — a boundary-only crossing can be labelled `match_type: 'exact'`, claiming the strongest available provenance for an endpoint that is not a symbol.**

The branch switches on `n.neighborUid.startsWith('manifest::')` — a property of the *endpoint* — but takes `match_type` from `n.matchType`, a property of the *link*. Those are not the same thing: manifest contracts participate in `runExactMatch`, which emits `matchType: 'exact', confidence: 1.0` while carrying the consumer's `symbolUid` through verbatim (`matching.ts:231-236`). This repo's own code says so at `sync.ts:384-386`:

> *"Manifest contracts participate in runExactMatch, so a manifest-declared link can also emit a matchType:'exact' CrossLink with the same endpoints."*

Reproduced: `runExactMatch → [{ matchType: 'exact', confidence: 1, fromUid: 'manifest::app::custom::X' }]`, which then arrives here and is pushed as `match_type: 'exact'` with `by_depth: {}`. So the `|| 'manifest'` fallback rarely fires (`rowToNeighbor` at `:398` already substitutes `'exact'` for a missing value), and the entry a consumer sees claims exact provenance for an unresolved far endpoint. This also undercuts `match_type` as the informal traversal-status hint some consumers might reach for.

**Suggested fix:** force `match_type: 'manifest'` in this branch — that is what the comment above claims it emits — or gate the branch on `n.matchType === 'manifest'` *in addition to* the uid shape.

*Engine A adversarial lane only; reproduced, and corroborated by the repo's own comment at `sync.ts:384-386`.*

[reproduced]

gitnexus/test/unit/group/manifest-synthetic-impact.test.ts
```diff
@@ -0,0 +1,178 @@
...
+        contract: expect.objectContaining({
+          id: 'custom::executeAddDynamicLinkMS',
+          match_type: 'manifest',
+          confidence: 1,
+        }),
+        by_depth: {},
+        affected_processes: [],
+      }),
+    ]);
+    expect(result.truncated).toBe(false);
```
magyargergo, line 157: **P2 — the single most consequential behavioral delta of this PR, `result.risk`, is never asserted.**

The spec pins `cross_repo_hits`, the `cross` array shape, `truncated`, and the `impactByUid` call count — but not `risk`, which is the field that actually flips for this exact fixture. Running this test's own fixture and capturing the full result yields `"risk": "HIGH"` where BASE returned `LOW`, because `mergeRisk` escalates on `confidence >= 0.85` with no regard for the empty `by_depth`.

DoD §2.7 asks that tests "would fail if behavior, wiring, or contracts were broken". As written, a future change to `mergeRisk` or to how manifest crossings are pushed could silently flip this behavior in either direction with nothing failing.

**Suggested fix:** add `expect(result.risk).toBe(...)` with whatever value the maintainers decide is correct for a boundary-only crossing, plus a `mergeRisk` unit case pinning `mergeRisk('LOW', [boundaryOnlyCrossing])` directly — one line each. Asserting it is what forces the risk question to be answered deliberately rather than inherited.

While here: `contract.type` is also unasserted inside the nested `expect.objectContaining` (this branch newly constructs it from `n.contractType` at `cross-impact.ts:641`), so adding `type: 'custom'` would close that gap for free.

*Cross-engine: flagged by five Engine A lanes and noted in Engine B's ce coverage section ("tests do not pin whether this is intentional").*

[reproduced]

gitnexus/test/unit/group/manifest-synthetic-impact.test.ts
```diff
@@ -0,0 +1,178 @@
...
+  _readOnly: true,
+} as BridgeHandle;
+
+vi.mock('../../../src/core/group/bridge-db.js', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../../../src/core/group/bridge-db.js')>();
+  return {
+    ...actual,
+    readBridgeMeta: vi.fn(async () => ({ version: 1, generatedAt: '', missingRepos: [] })),
+    getCachedBridgeReadOnly: vi.fn(async () => bridgeHandle),
+    queryBridge: vi.fn(async () => [
```
magyargergo: **P2 — the mock returns a single synthetic row, so nothing exercises a mixed real + `manifest::` neighbor set.**

Production `group sync` output routinely contains both real and manifest-only `ContractLink`s across the same repo pair — that is the normal configuration, since manifest links *supplement* auto-detection rather than replacing it. Nothing in the suite proves that with both present in one `runGroupImpact` call: the real UID still takes `impactByUid` (`cross-impact.ts:651-685`), the `seen` key at `:603` keeps the two rows apart, and both entries combine correctly into `cross_repo_hits` and `mergeRisk`.

This is also where a real defect may hide. The `seen` key embeds `neighborUid`, and the two upstream dedupers disagree on identity — `sync.ts:86-88` keys on `symbolUid` but not `matchType`, while `normalization.ts:84-92` keys on `matchType` and `filePath`/name but not `symbolUid`. With `detect.http: true` plus a manifest link for the same contract, the app repo can hold two consumer contracts for one boundary (one synthetic with `filePath: ''`, one real), neither deduper collapses them, and both survive to this loop. Engine A's adversarial lane reproduced that shape end-to-end against a real bridge: `cross_repo_hits=2` for **one** logical crossing, one entry hollow and one traversed. Composed with the `cross.length >= 3 → CRITICAL` rule, two genuinely-distinct boundaries that each double-count now reach CRITICAL where they previously reported HIGH.

**Suggested fix:** return two rows from the `queryBridge` mock — one real uid, one `manifest::` — and assert `impactByUid` was called exactly once (for the real uid), `cross.length === 2`, one entry with populated `by_depth` and one with `{}`. If the double-count above is confirmed, dedupe the emitted `cross[]` by `(repo_path, contract.id)` preferring the entry that actually has a fan-out.

*Cross-engine: the mixed-neighbor gap is called out by both engines (B1 lane 4 and B2's coverage note both name it explicitly); the double-count consequence is Engine A only, reproduced.*

[code-read]

gitnexus/test/unit/group/manifest-synthetic-impact.test.ts
```diff
@@ -0,0 +1,178 @@
...
+import type { GroupToolPort } from '../../../src/core/group/service.js';
+
+const bridgeHandle = {
+  _db: {},
+  _conn: {},
+  groupDir: '',
+  _readOnly: true,
+} as BridgeHandle;
+
+vi.mock('../../../src/core/group/bridge-db.js', async (importOriginal) => {
```
magyargergo, line 19: **P2 — the spec mocks away `bridge-db.js` wholesale, including the production wiring whose failure caused this bug.**

`readBridgeMeta`, `getCachedBridgeReadOnly`, `queryBridge` and `closeBridgeDb` are all replaced, and the mock hands back exactly the `neighborUid`, `matchType`, `confidence` and contract fields the assertions want. The test therefore stays green if the real bridge schema, the `CY_NEIGHBORS_UPSTREAM`/`DOWNSTREAM` query aliases, the stored synthetic uid, or the consumer/provider join ever stop producing that row.

DoD §2.7 is explicit: *"Integration tests hit a real database where the production path does; do not introduce mocks that hide migration or schema drift."* There is a specific precedent here — `test/integration/group/manifest-resolve-symbol-2325.test.ts` exists precisely because mock-level extractor tests could not see a real LadybugDB parser rejection, and that is the first of #2722's two causes. The existing real-bridge coverage in `test/unit/group/resolve-bridge-neighbors.test.ts:22` explicitly uses "both UIDs populated", so it does not reach the synthetic case either.

In fairness the risk class is lower than #2325's — this diff is pure control flow over already-fetched rows, not a Cypher change — and the test does correctly pin the changed branch (delete lines 635-649 and it fails at line 114). But nothing in the suite confirms a `manifest::` uid survives `writeBridge` → the neighbor query round-trip with its prefix intact.

**Suggested fix:** add a `withTestLbugDB` integration case that writes a bridge containing a real local uid and a synthetic far uid, then drives `runGroupImpact` against it. Engine A's adversarial lane built exactly this harness for the review, so it is a known-feasible ~40-line test. Route it into the `lbug-db` vitest project per TESTING.md.

*Cross-engine: Engine B's swarm rates this P2 and merge-blocking against DoD §2.7; four Engine A lanes raise the same gap.*

[code-read]

gitnexus/src/core/group/cross-impact.ts
azizur100389: Addressed in 33c8a1dd. Boundary-only entries now carry fanout_status:not_attempted, and mergeRisk excludes them from confidence and crossing-count escalation. Tests pin one and three high-confidence boundary-only crossings at LOW risk.

gitnexus/test/unit/group/manifest-synthetic-impact.test.ts
azizur100389: Addressed in 33c8a1dd. Added mixed synthetic plus real neighbor coverage. impactByUid runs exactly once for the real UID, and output deduplicates the logical boundary by preferring the completed fan-out over the boundary-only placeholder.

gitnexus/src/core/group/cross-impact.ts
azizur100389: Addressed in 33c8a1dd. CrossRepoImpact now has an optional additive fanout_status:not_attempted discriminator. CLI output reports the boundary-only count, and the MCP impact description documents that empty fan-out fields are not a completed zero-impact walk.

gitnexus/src/core/group/cross-impact.ts
azizur100389: Addressed in 33c8a1dd. Boundary-only synthetic endpoints now always report match_type:manifest, including when exact matching produced the ContractLink. Added a regression test for that producer shape.

gitnexus/test/unit/group/manifest-synthetic-impact.test.ts
azizur100389: Addressed in 33c8a1dd. The tests now assert LOW for boundary-only crossings, LOW for three boundary-only crossings, and HIGH when a real high-confidence fan-out completes.

gitnexus/src/core/group/cross-impact.ts
azizur100389: Addressed in 33c8a1dd. A synthetic endpoint with no filePath is now retained under service scoping as scope-unknown rather than silently dropped, with fanout_status:not_attempted. Added a service:src regression case.

gitnexus/test/unit/group/manifest-synthetic-impact.test.ts
azizur100389: Addressed in 33c8a1dd. Added a real writeBridge to bridge query to runGroupImpact regression in the sequential lbug-db project. It verifies the synthetic UID survives persistence and produces a boundary-only result without calling impactByUid. The test runs on Linux CI and follows the existing Windows native-handle reopen guard.

azizur100389:
Addressed all seven review findings in `33c8a1dd`:

- Added optional `fanout_status: not_attempted` for boundary-only crossings.
- Excluded unattempted fan-out from HIGH/CRITICAL risk escalation.
- Kept scope-unknown manifest boundaries visible under `service` filtering.
- Forced manifest provenance for synthetic endpoints.
- Deduplicated mixed real + synthetic links, preferring completed fan-out.
- Added risk, service-scope, provenance, mixed-neighbor, and unavailable-repo assertions.
- Added a real `writeBridge` → neighbor query → `runGroupImpact` LadybugDB regression in the sequential DB project.
- Documented the JSON/MCP contract and surfaced boundary-only counts in CLI output.

Validation:
- Focused group-impact tests: 33 passed.
- Typecheck, lint, Prettier, and `git diff --check`: passed.
- Full local suite executed: 15,258 passed / 416 skipped / 95 failed. The failures are unrelated Windows/current-main environment failures (native DB locks/extensions, missing `grep`, and subprocess timeouts); no changed group-impact test failed. The real bridge regression follows the existing Windows reopen guard and runs on Linux CI.

PR #2791, opened 2026-08-02
fix(lbug): persist Go function-local variables

### LINKED ISSUE #2789: analyze crashes on Go repository: Function→Variable not declared in LadybugDB relation schema

## Summary

`gitnexus analyze --index-only .` aborts while indexing a Go repository because the emitted `Function→Variable` relationship pair is not declared in the LadybugDB relation schema.

## Context

This was reproduced while indexing the public [scrypster/muninndb repository at commit `0b7a427`](https://github.com/scrypster/muninndb/commit/0b7a4272e16f403747501fabaf29e1b4e681f999).

The run performed a forced full rebuild because the existing index schema was v5 and the current analyzer schema was v33. It then failed during relationship CSV routing/loading.

This appears related to #2781 and merged PR #2769. That fix declared several previously missing pairs, including `Variable→Method` and `Method→Variable`, but it did not declare `Function→Variable`. The current failure is therefore a distinct remaining pair in the same fail-fast schema-validation path.

## Expected behavior

`gitnexus analyze --index-only .` should complete successfully for the Go repository and persist relationships involving function-local variables.

## Actual behavior

Analysis aborts with:

```text
Analysis failed: Relationship label pair Function→Variable is not declared in the LadybugDB relation schema
Error: Relationship label pair Function→Variable is not declared in the LadybugDB relation schema
    at assertDeclaredPair (file:///opt/homebrew/lib/node_modules/gitnexus/dist/core/lbug/rel-pair-routing.js:75:15)
    at RelPairRouter.route (file:///opt/homebrew/lib/node_modules/gitnexus/dist/core/lbug/rel-pair-routing.js:132:9)
    at streamAllCSVsToDisk (file:///opt/homebrew/lib/node_modules/gitnexus/dist/core/lbug/csv-generator.js:669:43)
    at async loadGraphToLbug (file:///opt/homebrew/lib/node_modules/gitnexus/dist/core/lbug/lbug-adapter.js:935:15)
    at async runFullAnalysisInner (file:///opt/homebrew/lib/node_modules/gitnexus/dist/core/run-analyze.js:1627:13)
    at async runFullAnalysis (file:///opt/homebrew/lib/node_modules/gitnexus/dist/core/run-analyze.js:475:16)
    at async analyzeCommandImpl (file:///opt/homebrew/lib/node_modules/gitnexus/dist/cli/analyze.js:1106:24)
    at async analyzeCommand (file:///opt/homebrew/lib/node_modules/gitnexus/dist/cli/analyze.js:557:9)
    at async Command.<anonymous> (file:///opt/homebrew/lib/node_modules/gitnexus/dist/cli/lazy-action.js:60:9)
```

The failed run leaves the index marked incomplete with `incremental-in-progress`.

## Steps to reproduce

1. Clone `https://github.com/scrypster/muninndb.git`.
2. Check out commit `0b7a4272e16f403747501fabaf29e1b4e681f999`.
3. Install GitNexus `1.6.10-rc.145`.
4. From the repository root, run:
   ```shell
   gitnexus analyze --index-only .
   ```
5. Observe the abort during LadybugDB relationship routing/loading.

## Environment

- OS: macOS 26.5.2 (Apple Silicon / arm64)
- Node: v26.5.1
- GitNexus: 1.6.10-rc.145
- Repository: scrypster/muninndb
- Repository commit: `0b7a4272e16f403747501fabaf29e1b4e681f999`

## Logs / screenshots

No secrets or private source are involved. The complete relevant stack trace is included above.

A likely schema-level fix is to declare `FROM Function TO Variable` in `RELATION_SCHEMA`, bump the incremental schema version, and add a regression test using a Go function containing a local variable.

koriyoshi2041:
## What changed

- declare the `Function→Variable` LadybugDB relation pair emitted for Go function-local variables
- bump the incremental schema version so existing databases are rebuilt with the expanded relation DDL
- pin both the declared pair and the reuse-gate transition in unit tests

Closes #2789.

## Verification

- `npm test -- --run test/unit/schema.test.ts test/unit/call-summary-schema-version.test.ts`
- `npx tsc --noEmit`
- `npx prettier --check src/core/lbug/schema.ts src/storage/repo-manager.ts test/unit/schema.test.ts test/unit/call-summary-schema-version.test.ts`
- `git diff --check`
- built the local CLI and indexed `scrypster/muninndb`; analysis completed with 22,199 nodes and 82,651 edges instead of aborting at relation routing

## Risk

Low. The runtime change only expands the endpoint pairs allowed by the shared relation table. The schema bump intentionally forces one full rebuild for existing indexes.

### DIFF
--- a/gitnexus/src/core/lbug/schema.ts
+++ b/gitnexus/src/core/lbug/schema.ts
@@ -297,6 +297,7 @@ CREATE REL TABLE ${REL_TABLE_NAME} (
   FROM Function TO \`Typedef\`,
   FROM Function TO \`Union\`,
   FROM Function TO \`Property\`,
+  FROM Function TO \`Variable\`,
   FROM Function TO CodeElement,
   FROM Class TO Method,
   FROM Class TO Function,
--- a/gitnexus/src/storage/repo-manager.ts
+++ b/gitnexus/src/storage/repo-manager.ts
@@ -687,8 +687,13 @@ export interface RepoMeta {
  * Numbered 34, not 33: `main` took 33 for Spring AOP (#2416) mid-flight, landing
  * on exactly this branch's number — the seventh collision in this series and the
  * first exact clash. Re-check against origin/main before merge.
+ *
+ * v35: Go function-local variables emit Function→Variable containment edges
+ * (#2789). LadybugDB fixes endpoint pairs when the relation table is created,
+ * so a pre-v35 index cannot persist these edges through incremental writeback.
+ * Force a full re-analyze.
  */
-export const INCREMENTAL_SCHEMA_VERSION = 34;
+export const INCREMENTAL_SCHEMA_VERSION = 35;
 
 export interface IndexedRepo {
   repoPath: string;
--- a/gitnexus/test/unit/call-summary-schema-version.test.ts
+++ b/gitnexus/test/unit/call-summary-schema-version.test.ts
@@ -73,12 +73,12 @@ describe('CALL_SUMMARY relation-type exclusion (U-C1)', () => {
 });
 
 describe('CALL_SUMMARY incremental reuse gate (U-C5)', () => {
-  it('INCREMENTAL_SCHEMA_VERSION is bumped to 34 (Spring AOP relation pairs #2416, then receiver-chain wire format v2)', () => {
+  it('INCREMENTAL_SCHEMA_VERSION is bumped to 35 (Go function-local variable relation pair #2789)', () => {
     // Moves with every bump BY DESIGN — that is the point of pinning it. A
     // change that alters emitted ids or edges without bumping would otherwise
     // ship silently, and an existing index would keep serving the old graph
     // through the reuse gate below.
-    expect(INCREMENTAL_SCHEMA_VERSION).toBe(34);
+    expect(INCREMENTAL_SCHEMA_VERSION).toBe(35);
   });
 
   it('a pre-current stamp fails the `=== INCREMENTAL_SCHEMA_VERSION` reuse gate → forces full re-analyze', () => {
@@ -223,7 +223,10 @@ describe('CALL_SUMMARY incremental reuse gate (U-C5)', () => {
     // would silently fall back to the text cascade for every chain-carrying
     // site → must NOT reuse.
     expect(passesReuseGate(33)).toBe(false);
+    // A pre-v35 (v34) index lacks Function→Variable in the LadybugDB DDL,
+    // so it cannot persist Go function-local variable containment edges.
+    expect(passesReuseGate(34)).toBe(false);
     // The current stamp passes the gate (incremental top-up eligible).
-    expect(passesReuseGate(34)).toBe(true);
+    expect(passesReuseGate(35)).toBe(true);
   });
 });
--- a/gitnexus/test/unit/schema.test.ts
+++ b/gitnexus/test/unit/schema.test.ts
@@ -215,6 +215,11 @@ describe('LadybugDB Schema', () => {
       expect(RELATION_SCHEMA).toContain('FROM Interface TO CodeElement');
     });
 
+    it('declares the Go function-local variable containment pair (#2789)', () => {
+      const declaredPairs = parseRelationSchemaPairs(RELATION_SCHEMA);
+      expect(declaredPairs.has('Function|Variable')).toBe(true);
+    });
+
     it('declares the Swift enum/property member-containment pairs (#2769)', () => {
       // Asserted through the runtime's own parser, not raw strings, so a
       // cosmetic DDL formatting change cannot fail this without a semantic


magyargergo:
Linking this to #2792, which is the umbrella issue for this bug class.

I reproduced #2789 and verified this fix end to end. Cloned `scrypster/muninndb` at the exact commit `0b7a4272` (note it is not on the default branch, so a plain clone fails with `unable to read tree`, you need `git fetch '+refs/heads/*:refs/remotes/origin/*'`). With the one schema line stripped from the built `dist` the run aborts with the exact error from the issue. With it restored the run completes, 20,069 nodes and 71,580 edges. A trivial two function Go file does not reproduce, you need a real corpus. The version bump 34 to 35 is correct and `SCHEMA_BUMP` is rightly untouched, since the DDL is not the capture shape. Both cited test files pass.

Two things to flag before merge.

**Subsumed by #2792.** The branch on #2792 replaces the hand listed pairs with a generated cross product of `SCOPE_BRIDGE_SOURCE_LABELS` x `SCOPE_BRIDGE_TARGET_LABELS`, and `Function|Variable` falls out of it for free. I confirmed this by evaluating that branch's generated `RELATION_SCHEMA` through the real `parseRelationSchemaPairs`. If #2792 lands, this PR becomes redundant.

**Schema version collision.** #2792 also bumps `INCREMENTAL_SCHEMA_VERSION` to 35. `main` is at 34. Whichever of the two merges second has to renumber.

**It also does not close the bug class.** With this fix applied, `analyze` still hard aborts on this repo's own fixture corpus:

```
Analysis failed: Relationship label pair Module→Property is not declared in the LadybugDB relation schema
```

I harvested the full remaining set by patching `assertDeclaredPair` to log and skip instead of throw: 13 undeclared pairs over 106 edges, in three language families. COBOL has 11 of them, plus `Function→File` on Vue and `Trait→Trait` on PHP. Each is an independent hard abort, confirmed per fixture on an unpatched build with default flags and no env overrides, so `cobol-app`, `vue-basic` and `php-transitive-traits` all fail to index at this PR head. The details and the full list are in my comment on #2792.

None of this is a regression from this PR. It is strictly fewer aborts than `main`, so merging is safe. The point is only that it is the third piecemeal fix in the series and the underlying gap survives it.

The reason CI stays green while the repo's own fixtures abort is that no test derives the emitted pair set from a real corpus. Every assertion in `test/unit/schema.test.ts` is a hand written point check, and this PR adds one more of the same shape (`declaredPairs.has('Function|Variable')`). A corpus derived check would have caught #2781, #2769 and #2789 before any of them were filed.

koriyoshi2041:
Thanks for the corpus-level verification and the umbrella analysis. Agreed that #2792 is the right durable fix. I’ll leave this bounded patch available while that work is in progress; if the systematic schema change lands first, this can close as superseded. If this lands first, the umbrella branch will need the schema version bumped past 35.

magyargergo:
Closing this in favour of #2793, which handles the whole surface centrally rather than one pair at a time. Not because anything here was wrong.

Credit where it is due: the report and this fix are both correct, and the reproduction was the valuable part. I verified it end to end before deciding anything. Cloning `scrypster/muninndb` at `0b7a4272` and stripping the one schema line from the built `dist` reproduces the abort exactly as described, and restoring it completes the run at 20,069 nodes and 71,580 edges. Worth recording for anyone who repeats it: that commit is not on the default branch, so a plain clone fails with `unable to read tree` and you need `git fetch '+refs/heads/*:refs/remotes/origin/*'` first. A small synthetic Go file does not reproduce it either, you need a real corpus. Pinning the exact repository and commit in the issue is what made this straightforward to confirm.

The reason for closing rather than merging is that #2793 replaces the hand listed pairs in the scope resolution block with a generated cross product of the two label sets that define the emit surface. `Function|Variable` falls out of that generation on its own, so the line added here would be deleted by that PR anyway. I checked this rather than assumed it, by evaluating #2793's generated `RELATION_SCHEMA` through the real `parseRelationSchemaPairs`: the pair is declared, and running #2793's build against `scrypster/muninndb` gives 20,069 nodes and 71,580 edges, identical to this PR. So nothing is lost by closing.

There was also a hard conflict. Both this PR and #2793 bump `INCREMENTAL_SCHEMA_VERSION` to 35, so only one of them could land as written.

One thing this uncovered that is worth stating plainly, since it is not a criticism of this PR. Declaring a single pair was never going to end the loop, and neither did the generated cross product on its own. With this fix applied, `analyze` still aborted on this repository's own `test/fixtures/lang-resolution` with `Module→Property is not declared`. A full sweep found 13 undeclared pairs over 106 edges, spread across COBOL, Vue and PHP. They come from containment and inheritance emitters that sit outside the scope resolution bridge, so no predicate derived audit can see them. Those remaining pairs, plus a corpus derived test that would have caught all of this before any of #2781, #2769 or #2789 were filed, are being handled on #2793.

Thanks for the clean report and the exact repro commit.

koriyoshi2041:
Thanks for validating the corpus path and carrying it into the broader fix. Closing in favor of #2793 makes sense, especially with the shared schema version and the additional emitter pairs you found. I’ll treat #2793 as the authoritative lane.

Issue #2680, opened 2026-07-24
epic: stream graph emit into LadybugDB per chunk — O(chunk) analyze memory at kernel scale


magyargergo:
### Area

gitnexus (CLI / core / indexing)

### Summary

The in-memory `KnowledgeGraph` retains every node and relationship on the main thread for the entire pipeline, so `analyze`'s peak heap is O(repo). At Linux-kernel scale this exceeds any practical `--max-old-space-size`; #2679 makes that failure fast and honest, but *completing* such repos needs the graph to leave the JS heap incrementally.

### Measurements (from the #2649 investigation)

Instrumented analyze of `torvalds/linux` (94,773 files, ~64k parseable C/C++), heap probes via `GITNEXUS_DEBUG_HEAP=1`:

```
parse-chunk-25 used_mb=2481 rss_mb=7133  nodes=1143241 parsedFiles=0
parse-chunk-50 used_mb=4622 rss_mb=11102 nodes=2188306 parsedFiles=0
parse-chunk-75 used_mb=6431 rss_mb=15038 nodes=3200921 parsedFiles=0
```

- Heap tracks `graph.nodeCount` linearly at **~2.1 KB/node**; the kernel projects ~5M nodes (~10.5 GB) by end of parse, before scope resolution adds CALLS/IMPORTS edges to the same in-memory graph. A 12 GB heap died at chunk 91/113 while still parsing.
- `parsedFiles=0` throughout: the disk-backed ParsedFile spill (#1983/#2038) works — ParsedFiles are not the problem.
- The retention point is `createKnowledgeGraph()`'s maps plus three secondary indexes (`relationshipsByType`, `edgeIdsByNode` — one `Set` per node — `nodeIdsByFile`), which only leave the heap at `loadGraphToLbug(...)` after the whole pipeline (`run-analyze.ts`).

### Proposed direction

Stream node/relationship emit into LadybugDB (or CSV COPY staging) per chunk during parse and resolution — the same shape #2202's `PdgEmitSink` already uses to keep the PDG layers O(chunk) ("peak RSS to O(chunk) rather than O(graph) at full-kernel scale"). The hard part is downstream consumers:

- the `communities` (Leiden) and `processes` phases currently consume the whole in-memory graph — they would need to run DB-side or on a compact projection (id-interned adjacency, not full `GraphNode` objects); this aligns with the parallel-Leiden work in the Icebug/Ladybug collaboration (#2337);
[...]

PR #2793, opened 2026-08-02
fix(schema): declare the full scope-resolution relation cross product (#2792)

### LINKED ISSUE #2789: analyze crashes on Go repository: Function→Variable not declared in LadybugDB relation schema

### LINKED ISSUE #2792: Systematic RELATION_SCHEMA gap: multiple FROM→TO pairs missing for linkable/caller-anchor labels
# Systematic RELATION_SCHEMA gap: multiple FROM→TO pairs missing for linkable/caller-anchor labels

## Summary

`RELATION_SCHEMA` in `gitnexus/src/core/lbug/schema.ts` is missing numerous `FROM→TO` label pairs that are produced by scope-resolution at runtime. The `assertDeclaredPair` fail-fast check (introduced in #2740) exposes these gaps as hard crashes, but prior fixes (#2769) have been piecemeal — addressing only the specific pair that triggered the crash — rather than auditing the full surface. This results in `analyze` crashing on different codebases with different missing pairs, each requiring a separate fix.

## Context

Two independent code paths produce edges whose FROM/TO labels must be declared in `RELATION_SCHEMA`:

1. **`isLinkableLabel`** (`scope-resolution/graph-bridge/node-lookup.ts:247-281`): Labels that can be CALLS sources via the value-receiver bridge (Case 5 in `receiver-bound-calls.ts`). Currently includes: `Function`, `Method`, `Constructor`, `Module`, `Class`, `Interface`, `Struct`, `Enum`, `Trait`, `Variable`, `Property`, `Const`, `Macro`.

2. **`isCallerAnchorLabel`** (`scope-resolution/graph-bridge/ids.ts:57-68`): Labels that can serve as caller anchors when resolving a call site's graph ID. Currently includes: `Function`, `Method`, `Constructor`, `Module`, `Class`, `Interface`, `Struct`, `Enum`.

When a call site inside a `Class` scope (e.g., a Java static initializer or field initializer) references a `Variable`, the caller resolves to `Class:...` and the target resolves to `Variable:...`, producing a `Class→Variable` edge. If this pair is not declared in `RELATION_SCHEMA`, `assertDeclaredPair` throws and the entire analysis crashes.

### `Class→Variable` (Java)

```
Error: Phase 'scopeResolution' failed: Relationship label pair Class→Variable is not declared in the LadybugDB relation schema
    at assertDeclaredPair (rel-pair-routing.js:75:15)
    at GraphEmitSink.addRelationship (graph-emit-sink.js:377:9)
    at tryEmitEdge (edges.js:91:11)
    at emitReceiverBoundCalls (receiver-bound-calls.ts:823:36)
```

Triggered by Java codebases where a class static initializer or field initializer references a `Variable` node.

### `Const→Method` (Vue/JS) — fixed in #2769

Previously crashed on Vue/JS projects with `const obj = { method() {} }` patterns. Fixed by adding `Const→Method` and `Method→Const`, but the fix was not extended to other missing pairs.

### `Class` as FROM (19 pairs declared, 4+ missing)

Declared:
```
Class → Method, Function, Class, Interface, Community, Template, TypeAlias,
        Struct, Enum, Annotation, Constructor, Trait, Macro, Impl, Union,
        Namespace, Typedef, Property, CodeElement, Process
```

Missing:
```
Class → Variable
Class → Const
Class → Static
Class → Module
Class → Record
Class → Delegate
```

### `Const` as FROM (4 pairs declared, 8+ missing)

Declared (after #2769):
```
Const → Community, Method, Process
```

Missing:
```
Const → Function
Const → Constructor
Const → Class
Const → Interface
Const → Struct
Const → Enum
Const → Property
Const → CodeElement
```

### `Variable` as FROM (3 pairs declared, 8+ missing)

Declared (after #2769):
```
Variable → Community, Method, Process
```

Missing:
```
Variable → Function
Variable → Constructor
Variable → Class
Variable → Interface
Variable → Struct
Variable → Enum
Variable → Property
Variable → CodeElement
```

### `Static` as FROM (2 pairs declared, 10+ missing)

Declared:
```
Static → Community, Process
```

Missing:
```
Static → Method, Function, Constructor, Class, Interface, Struct, Enum,
         Property, CodeElement, Variable
```

### `Interface` as FROM (8 pairs declared, 3+ missing)

Declared:
```
Interface → Community, Function, Method, Class, Interface, TypeAlias,
            Struct, Constructor, Property, CodeElement
```

Missing:
```
Interface → Variable
Interface → Const
Interface → Enum
```

### `Struct` as FROM (10 pairs declared, 2+ missing)

Declared:
```
Struct → Community, Trait, Struct, Class, Enum, Function, Method,
         Interface, Constructor, Property
```

Missing:
```
Struct → Variable
Struct → Const
```

### `Enum` as FROM (4 pairs declared, 2+ missing)

Declared:
```
Enum → Enum, Community, Class, Interface
```

Missing:
```
Enum → Variable
Enum → Const
```

For comparison, `Constructor` has 17 outgoing pairs — the most complete of the multi-language node types.

## Root cause

The schema was built incrementally: each new language or feature added only the pairs it immediately needed. The `assertDeclaredPair` check (#2740) made missing pairs visible as crashes, but the fixes (#2769) addressed only the specific pairs reported, without auditing the full label surface defined by `isLinkableLabel` and `isCallerAnchorLabel`.

### Option A: Systematic audit (recommended)

Derive the minimum required pairs from the code:

1. All labels in `isLinkableLabel` can be CALLS sources → they need `FROM <label> TO` pairs for every callable target (Method, Function, Constructor, etc.).
2. All labels in `isCallerAnchorLabel` can be caller anchors → they need `FROM <label> TO` pairs for every possible target label.
3. Add a CI test that validates `RELATION_SCHEMA` covers all pairs implied by `isLinkableLabel` ∪ `isCallerAnchorLabel`, so future additions to either set automatically flag missing pairs.

### Option B: Add missing pairs individually

Add the specific pairs listed in the "Missing" sections above. This is the minimal fix but will require another PR the next time a new label or edge pattern is added.

## Related

- #2781 — original `Const→Method` crash report
- #2769 — partial fix for `Const→Method`, `Variable→Method`, `Method→Const`, and Swift pairs
- #2740 — introduced `assertDeclaredPair` fail-fast check
- #1718 — introduced Case 5 value-receiver bridge, making `Const`/`Variable` linkable labels

magyargergo:
Closes #2792. Supersedes #2791 (closed) — its `Function→Variable` falls out of the generated cross product below, verified by running this branch against the `scrypster/muninndb` repro from #2789: 20,069 nodes / 71,580 edges, identical to that PR.

Two halves, because `RELATION_SCHEMA` has two halves. The scope-resolution block becomes **generated** and can no longer go stale. The hand-declared structural block cannot be generated — and was found to be stale too, so it is fixed and given its own guard.

## The audit

The issue asked for Option A — derive the required pairs from the code instead of chasing crash reports. Doing that narrows the surface further than the issue estimated.

`buildGraphNodeLookup` (`node-lookup.ts:136`) skips any node whose label fails `isLinkableLabel`, so the lookup holds **only** linkable-labelled nodes. Both endpoints of every graph-bridge edge resolve through that lookup — `tryEmitEdge` calls `resolveCallerGraphId` for the source and `resolveDefGraphId` for the target, and both bottom out there. So the emittable surface is exactly:

| side | labels | why |
|---|---|---|
| FROM | `LINKABLE_LABELS` + `File` | every `resolveDefGraphId` hit, plus the module-level fallback at the end of `resolveCallerGraphId` |
| TO | `LINKABLE_LABELS` + `CALL_TARGET_TYPES` | `tryEmitEdgeWithExplicitTargetId` bypasses the lookup and emits a callable def's own `nodeId` |

Two corrections to the issue's analysis:

- **`isCallerAnchorLabel` adds nothing.** It is a strict subset of `isLinkableLabel`, so unioning the two sets is a no-op. A caller anchor outside the lookup could never resolve to an id anyway — there is now an assertion pinning that invariant.
- **`Delegate` is needed as a target** and was not on the issue's list. It is in `CALL_TARGET_TYPES` (C# delegates, `FREE_CALLABLE_TUPLE`) without being linkable, and the explicit-target-id path can emit it.

That is 14 × 14 = 196 pairs, 99 of them previously undeclared.

## The fix

The scope-resolution block is now **generated** from those two label lists rather than hand-listed, so no pair drawn from them can be missing by construction. Containment / inheritance / imports / DI pairs stay hand-declared — no single predicate describes them. **(Superseded — see “Update” at the end: a second rule was added and the declared total is now 450.)**

`test/unit/schema-pair-coverage.test.ts` derives the requirement from the real label sets and fails CI when either grows without the matching pairs landing in the DDL. That test, not the pair count, is the end of the piecemeal loop. **(Superseded: this originally kept hand-copied twin lists, justified by “`core/lbug` must not import the ingestion layer”. That justification was false — see “Update”.)**

`INCREMENTAL_SCHEMA_VERSION` 34 → 35 — LadybugDB fixes endpoint pairs when the rel table is created, so a pre-v35 database physically cannot store these edges.

## The other half — the hand-declared block was stale too

Generating the bridge surface is necessary but not sufficient. Declaring a single pair never was: #2781, #2769, #2789 and #2791 each declared exactly the pair in the stack trace. Checking whether *this* PR ended the loop turned up that it had not.

With #2791's fix applied, `analyze` still aborted on this repository's own fixture corpus:

```
Analysis failed: Relationship label pair Module→Property is not declared in the LadybugDB relation schema
```

A full sweep — `assertDeclaredPair` patched to log-and-skip, run over all of `test/fixtures/lang-resolution` — found **13 undeclared pairs over 106 edges**. The generated cross product above covers 3 of them (`Module→Property`, `Trait→Trait`, `Module→Constructor`). The remaining 10 come from emitters that never touch the graph bridge:

| source | pairs | why the audit above cannot see them |
|---|---|---|
| `cobol-processor.ts` | `Module→CodeElement`, `Module→Namespace`, `Module→Record`, `Namespace→Function`, `CodeElement→CodeElement`, `CodeElement→Module`, `CodeElement→Property`, `CodeElement→Record`, `Record→Record` | a COBOL program is a `Module`, a SECTION a `Namespace`, an FD a `Record`, a working-storage item a `Property`, an EXEC/dynamic-call site a `CodeElement` — wired with CONTAINS / CALLS / ACCESSES |
| `languages/vue/scope-resolver.ts` | `Function→File` | BINDS_EVENT_HANDLER from a handler `Function` to the child component's `File` — the only edge in the graph whose target is a `File` |

`CodeElement`, `Namespace`, `Record` and `File` are in neither `LINKABLE_LABELS` nor `CALL_TARGET_TYPES`. So no amount of extending the cross product reaches these, and `schema-pair-coverage.test.ts` is structurally incapable of failing on them. Each is an independent hard abort, confirmed per fixture on a built CLI with default flags and no env overrides — `cobol-app` and `vue-basic` both failed to index at `26132c016`.

These 10 are now declared in `STRUCTURAL_PAIR_DDL`.

## The guard for that half

`test/integration/structural-pair-coverage.test.ts` (new) derives its requirement from a **corpus** rather than a predicate, because no predicate describes this surface. It runs the real pipeline over the non-bridge fixtures (`cobol-app`, `vue-basic`, `php-transitive-traits`), mirrors `RelPairRouter.route` exactly — same `getNodeLabel` derivation from the id, same `validTables` skip — and requires every pair produced to be declared.

Mutation-checked rather than assumed: deleting `FROM Function TO File` fails it with exactly `Function|File`.

It is a sample, not a proof, and says so — a language without a fixture is unguarded, so a new structural emitter should land with an entry. It also stubs `GITNEXUS_WORKER_READY_TIMEOUT_MS`, since three cold worker-pool startups otherwise flake against the 5s default on a loaded runner and would fail for reasons unrelated to the schema.

## Why not just declare every pair

Measured against a real 32-node-table database:

| pairs | rel DDL | 300 anchored queries (`MATCH (a {id:…})-[r:CodeRelation]->(b)`) |
|---|---|---|
| 223 (before) | 10 ms | 1087 / 1167 ms |
| 322 (after) | 23 ms | 1118 / 1137 ms |
| 1024 (full 32×32) | 23 ms | ~1.8× slower |

The widening this PR does is free; a blanket cross product is not. The audited subset is the right scope.

## Verification

- `test/unit/schema-pair-coverage.test.ts` (new) — coverage, caller-anchor ⊆ linkable, endpoint labels are real node tables, no duplicate pair, plus the two reported crash pairs
- Real `Database` round-trip: `SCHEMA_QUERIES` applies and a `Class→Variable` edge persists and reads back
- `npx tsc --noEmit` clean; eslint + prettier clean
- 23 test files green (`test/unit/lbug`, `schema`, `rel-pair-routing`, `basicblock-callee-ids-schema`, `test/integration/csv-pipeline`); 61 graph-bridge scope-resolution tests green; `incremental-orchestration` 22/22
- Full `analyze` runs green on Java, TypeScript and C++ fixtures and on GitNexus itself (23k nodes / 86k edges)

Structural half, all on a built CLI:

- `test/integration/structural-pair-coverage.test.ts` (new) — mutation-checked
- `cobol-app`, `vue-basic`, `php-transitive-traits` now index (101 / 18 / 25 nodes) — the first two aborted at `26132c016`
- the whole `test/fixtures/lang-resolution` corpus completes: 10,876 nodes / 18,517 edges, previously an abort
- `scrypster/muninndb` @ `0b7a4272` (the #2789 repro): 20,069 nodes / 71,580 edges, matching #2791 exactly
- 89 tests green across `schema-pair-coverage`, `schema`, `rel-pair-routing`, `call-summary-schema-version`, `csv-pipeline` and the new file

## Honest gap

**I could not synthesize a fixture that emits `Class→Variable`.** Java fields, C++ members and TS class fields all mint `Property`, not `Variable`, and a full GitNexus self-index touches only 40 distinct pairs — all already declared on `main`. The reporter's stack trace is the only direct evidence for that specific pair.

That self-index number is worth reading as a warning rather than reassurance: GitNexus is a TypeScript repository, and TypeScript turns out to be entirely clean here. Self-indexing cannot validate this schema — only the multi-language fixture corpus can, which is exactly how the 13 pairs above surfaced.

The mechanism is still clear from the code: `resolveDefGraphId`'s last-resort `simpleKey(filePath, simpleName)` lookup is **label-agnostic**, so a `Class`-typed caller def can resolve to whichever same-named linkable node was registered first — including a `Variable`. That is why the fix is derivation-driven rather than repro-driven, and why the guard test matters more than the 99 pairs.

## Noted, not fixed

`Record` is absent from `isLinkableLabel`, which makes the `Record` branches in `buildGraphNodeLookup` and `resolveDefGraphId` (the template-arguments key) unreachable — a Java/C# `record` can never be a resolved call target. Adding it would change resolution behaviour, so it wants its own issue rather than riding along here.

## Merge note

Per the v25/v26/v29/v32/v34 history in the version ladder, re-check `INCREMENTAL_SCHEMA_VERSION` against `origin/main` immediately before merge — this series has collided seven times, once exactly.

---

## Update — review follow-up (`81daf370e`)

A multi-engine review of this branch found **four `analyze` hard-aborts still live at `da8eeb154`**, each reproduced end-to-end on the default user path (`analyze --index-only --skip-git`, no env overrides):

| pair | trigger | emitter |
|---|---|---|
| `Method→Annotation` | Spring `@Bean` + `@ConditionalOnMissingBean` (Java **and** Kotlin) | `frameworks/spring/conditionals.ts` |
| `Method→File` | Vue **Options-API** `methods:` handler bound to a child-component event | `languages/vue/scope-resolver.ts` |
| `Namespace→Record` | COBOL `DECLARATIVES` / `USE AFTER STANDARD ERROR ON <file>` | `cobol-processor.ts` |
| `Class→Tool` | `@mcp.tool()` applied to a class | `pipeline-phases/tools.ts` |

All four are pre-existing on `main` — not regressions from this PR — but **both guards were structurally blind to them**, and all 16 tests passed while all four crashes were live. The Vue one is the sharpest: this PR added `FROM Function TO File` for that exact emitter, whose source is a raw `nodeLookup.get(simpleKey(...))` hit spanning all 13 linkable labels. `<script setup>` mints a `Function` (declared); the Options API mints a `Method` (undeclared). One of thirteen got declared because one fixture happened to exercise it.

**The model in this PR was wrong.** "Bridge endpoint × structural endpoint" does not fit — `Namespace→Record` is structural on both sides. The property that actually holds is that the **anchor is a lookup result, not a literal at the emit site**, so the emitter cannot constrain its label. That gives a second closed-form rule, `DEFINITION_ANCHOR_LABELS × ATTACHMENT_TARGET_LABELS`, with the anchor set derived from `NODE_TABLES` by subtraction so a new node table joins automatically.

**332 → 450 declared pairs.** A ground-truth harvest (all three `assertDeclaredPair` call sites instrumented, every fixture repo run individually, plus two `--pdg` runs) confirms **exactly four** pairs were undeclared corpus-wide, and none remain.

### Sizing, now with a committed harness

`gitnexus/bench/schema-pairs` — real `@ladybugdb/core`, identical data at every size, ratios vs the 332-pair table:

| pairs | untyped anchored | typed (floor) |
|---|---|---|
| **450** | **0.93–1.05×** | 0.98–1.17× |
| 641 | 1.22–1.43× | 1.11–1.23× |
| 786 | 1.52–1.75× | 1.19–1.31× |
| 1024 | 2.03–2.34× | 1.31–1.57× |

450 is inside run-to-run noise; the harness independently reproduces the known #2792 cliff at 1024, which is what makes that trustworthy. The knee sits just above 450, which is why the containment block is not folded into a third rule (that would be 641, not the 786 previously stated here).

### Also in that commit

- **161 of the 233 hand-declared pairs were already produced by the rules** — deleted; the declared set is byte-identical at 450. They were load-bearing shadow: the generator suppresses anything already declared structurally, so narrowing a rule later would have silently kept pairs alive. A new guard now fails CI if a hand-declared pair is ever re-added inside a rule. Hand-list: **233 → 72**.
- **The twin lists are gone.** `schema.ts` imports `LINKABLE_LABELS` / `CALL_TARGET_TYPES` directly. The stated blocker was false — `csv-generator.ts` and `lbug-adapter.ts`, siblings in the same directory, already import from `../ingestion/`, and no rule in AGENTS.md / ARCHITECTURE.md / CONTRIBUTING.md / GUARDRAILS.md says otherwise.
- **`resolveStreamGraphEmit` is resolved after the guards that rebind `options.force`**, not at function entry. It gates on `force`, and every freshness guard runs ~360 lines later — so this PR's own v34→v35 bump would have pushed *every existing index* down the non-streamed emit path, losing the #2680 memory streaming added for the #2649 kernel-scale OOM.
- **`UndeclaredRelationPairError`** now carries the relationship type, both node ids and the source file, with a matching CLI branch. The old message named only the abstract label pair. Found via the cause chain, since `pipeline-phases/runner.ts` rewraps every phase failure.
- Four new fixtures reproducing the aborts, and **per-fixture sentinel pairs** so a fixture that stops emitting fails loudly instead of passing vacuously on an empty graph.
- One shared `relPairKeyFor` across the router, both emit sinks and the corpus guard (which previously hand-mirrored the router's skip rule); one cause-chain walker; one exported pair-matching regex.

The per-edge path stays allocation-free — failure context is passed positionally and the message is built only inside the throw.

### Verification

`tsc --noEmit` clean · 450 pairs / 0 duplicates · real-LadybugDB round-trip returns exactly 450 · all four repros now index · 10 directly-touched suites green (145 tests) · integration 9 files / 1034 tests green.


gitnexus/src/core/lbug/schema.ts
```diff
@@ -345,7 +410,17 @@ CREATE REL TABLE ${REL_TABLE_NAME} (
...
   FROM \`Module\` TO \`Module\`,
+  FROM \`Module\` TO \`Namespace\`,
+  FROM \`Module\` TO \`Record\`,
+  FROM \`Namespace\` TO Function,
+  FROM CodeElement TO CodeElement,
+  FROM CodeElement TO \`Module\`,
+  FROM CodeElement TO \`Property\`,
+  FROM CodeElement TO \`Record\`,
+  FROM \`Record\` TO \`Record\`,
+  FROM Function TO File,
```
magyargergo: **`Function→File` declares one of thirteen source labels this emitter can produce — Vue's Options API still hard-aborts `analyze`.**

The emitter this line exists for is `languages/vue/scope-resolver.ts:275-293`, and its source is a raw lookup hit:
```ts
const handlerNodeId = nodeLookup.get(simpleKey(parsedFile.filePath, handlerName));
```
`buildGraphNodeLookup` registers all 13 `LINKABLE_LABELS`, so that id can wear any of them. `<script setup>` + `function onButtonClick` mints a `Function` — which is why the `vue-basic` harvest saw exactly this pair. `methods: { onButtonClick() {} }` mints a `Method`; `const config = {…}` bound as a handler mints a `Const`. Reproduced at this head on a 2-file Vue repo via `node dist/cli/index.js analyze <dir> --index-only --skip-git`:
```
Analysis failed: Relationship label pair Method→File is not declared in the LadybugDB relation schema
    at assertDeclaredPair (dist/core/lbug/rel-pair-routing.js:75:15)
    at RelPairRouter.route (dist/core/lbug/rel-pair-routing.js:132:9)
    at streamAllCSVsToDisk (dist/core/lbug/csv-generator.js:669:43)
```
Independently reproduced by two Claude lanes (CLI and `runPipelineFromRepo` harnesses) and by the review coordinator. Pre-existing on `main` — not a regression from this PR — but the Options API is the mainstream non-`setup` Vue style, and this is the one line the PR added for this emitter.

Suggestion: generate the column instead of hand-picking it — `for (const from of SCOPE_BRIDGE_SOURCE_LABELS) → FROM \`${from}\` TO \`File\`` — and extend `requiredPairs()` in `schema-pair-coverage.test.ts` with `'File'` on the target side so the derived guard can see this surface at all. Add an Options-API component event binding to `vue-basic`.

Note: both Codex legs reviewed this PR and did not find this; treat as Claude + coordinator reproduction rather than cross-engine consensus.

gitnexus/src/core/lbug/schema.ts
```diff
@@ -345,7 +410,17 @@ CREATE REL TABLE ${REL_TABLE_NAME} (
   FROM \`Template\` TO \`Macro\`,
   FROM \`Template\` TO Interface,
   FROM \`Template\` TO \`Constructor\`,
+  FROM \`Module\` TO CodeElement,
   FROM \`Module\` TO \`Module\`,
+  FROM \`Module\` TO \`Namespace\`,
+  FROM \`Module\` TO \`Record\`,
+  FROM \`Namespace\` TO Function,
```
magyargergo, line 476: **`Namespace|Function` is the only `Namespace`-source pair here, and COBOL `DECLARATIVES` emits `Namespace→Record`.**

A COBOL section mints a `Namespace` node (`cobol-processor.ts:436-447`); `SELECT CUSTOMER-FILE` mints a `Record` (`cobol-processor.ts:1313-1317`); the declaratives loop emits `ACCESSES` between them:
```ts
const targetId = generateId('Record', `${filePath}:${decl.target}`);
graph.addRelationship({ type: 'ACCESSES', sourceId: sectionId, targetId, reason: 'cobol-error-handler' });
```
— `cobol-processor.ts:1062-1067`. `Namespace` is not in `LINKABLE_LABELS`, so the generated block can never reach it; the pair has to be hand-declared here and isn't. Reproduced twice, in two independent harnesses, on a minimal program with `PROCEDURE DIVISION. DECLARATIVES. … USE AFTER ERROR ON CUSTOMER-FILE. … END DECLARATIVES.`

Two things make this worse than a single missing pair: `grep -rn DECLARATIVES gitnexus/test/fixtures/` returns nothing, so `NON_BRIDGE_CORPUS`'s `cobol-app` cannot catch it; and `RE_USE_AFTER` (`cobol/cobol-preprocessor.ts:382-383`) currently rejects the IBM-standard `USE AFTER STANDARD ERROR PROCEDURE ON <file>` spelling, so the crash is dialect-gated today — fix that unrelated parser gap without declaring this pair and it becomes universal on any COBOL codebase with file error handling.

Suggestion: declare the reachable COBOL block (`Namespace|Record`, `Namespace|Property`, `Namespace|CodeElement`, `CodeElement|Function`, `Property|CodeElement`, …) rather than the ones one fixture happens to hit — 25 of COBOL's 64 ordered label pairs are still undeclared after this PR — and add a `DECLARATIVES` section to `cobol-app` (~10 lines, no new pipeline run) using the form the regex accepts.

gitnexus/src/core/lbug/schema.ts
```diff
@@ -248,9 +249,73 @@ CREATE NODE TABLE BasicBlock (
...
+ *
+ * Being hand-declared, this is the half that can still go stale — and it did:
+ * after #2789 declared its one pair, `analyze` still aborted on this repo's own
+ * `test/fixtures/lang-resolution` (COBOL containment, and the Vue
+ * BINDS_EVENT_HANDLER edge from `languages/vue/scope-resolver.ts`, the only
+ * edge whose target is a `File`). Those endpoint labels — `CodeElement`,
+ * `Namespace`, `Record`, `File` — are in neither scope-bridge set, so no
+ * predicate-derived check can reach them.
+ * `test/integration/structural-pair-coverage.test.ts` guards this block from a
+ * corpus instead; that is the guard, not this comment.
```
magyargergo: **This comment names the corpus test as "the guard", but both new guards are structurally blind to the region where the remaining aborts live.**

`schema-pair-coverage.test.ts:25-28` derives its requirement from `['File', ...LINKABLE_LABELS] × (LINKABLE_LABELS ∪ CALL_TARGET_TYPES)`. `Annotation`, `Tool` and `File`-as-target are in none of those sets, so no pair with those endpoints can ever enter `requiredPairs()`. `NON_BRIDGE_CORPUS` is three fixtures with no Spring conditional, no Options-API Vue binding, no `DECLARATIVES` and no MCP tool on a class. All 16 tests in the two new files pass while four `analyze` hard-aborts are live at this head:

| pair | emitter | trigger |
|---|---|---|
| `Method→Annotation` | `frameworks/spring/conditionals.ts:306-316` | `@Bean` + `@ConditionalOnMissingBean` (Java **and** Kotlin) — reproduced with a stack trace |
| `Method→File` | `languages/vue/scope-resolver.ts:275-293` | Vue Options-API handler — reproduced |
| `Namespace→Record` | `cobol-processor.ts:1055-1070` | COBOL `DECLARATIVES` — reproduced |
| `Class→Tool` | `pipeline-phases/tools.ts:96-104` | `@mcp.tool()` on a class — reproduced by one lane only; the review coordinator did **not** independently reproduce this one |

All four are the same shape: **one endpoint resolved through the bridge (so it wears a `LINKABLE_LABELS` label), the other a structural node type outside both sets.** That region has ~40 more undeclared pairs nobody built a trigger for, and each will arrive as a fresh one-line-fix issue.

Suggestion: give that region a closed-form requirement too, rather than another point-list — declare `({File} ∪ LINKABLE_LABELS) × {Annotation, Tool, Route, Section, CodeElement, Record, Namespace, File}` alongside the existing cross product and assert it in the unit test from the same constants. That is ~112 extra pairs (→ ~444 total), well below the 1024-pair point where this repo measured ~1.8× on untyped-endpoint anchored queries.

All four are pre-existing on `main`, so this is not a regression — and both Codex review legs approved without finding them. Reasonable to merge and track as a follow-up, but worth deciding deliberately given the PR's stated goal is to end this bug class.

gitnexus/src/core/lbug/schema.ts
```diff
@@ -248,9 +249,73 @@ CREATE NODE TABLE BasicBlock (
...
+  'Macro',
+] as const;
+
+/**
+ * Labels the bridge can put on the TARGET side: `LINKABLE_LABELS` again (every
+ * `resolveDefGraphId` hit), plus `Delegate` — `tryEmitEdgeWithExplicitTargetId`
+ * bypasses the lookup and emits a `CALL_TARGET_TYPES` def's own node id, and
+ * C# delegates are in that set without being linkable.
+ *
+ * Both lists are twins of sets that live in the ingestion layer, which must not
```
magyargergo: **The stated reason for hand-copying these twins does not hold, and the copies have no compile-time check.**

"sets that live in the ingestion layer, which must not be imported here" — two sibling files in this same directory already import from `../ingestion/`:
- `csv-generator.ts:22-23` — `parseTruthyEnv` from `../ingestion/utils/env.js`, `SYMBOL_NODE_LABELS` from `../ingestion/utils/symbol-labels.js`
- `lbug-adapter.ts:67-68` — `../ingestion/frameworks/spring/auto-configuration.js` and `.../spring/aop.js`

No rule in `AGENTS.md`, `ARCHITECTURE.md`, `CONTRIBUTING.md` or `GUARDRAILS.md` states this constraint, and importing `LINKABLE_LABELS`/`CALLER_ANCHOR_LABELS` here would not create a cycle (neither module, nor its transitive imports, reaches `core/lbug`). The duplication is a choice, not a constraint — which is fine, but the comment should say so, because it is load-bearing documentation for whoever next edits these lists.

Second, smaller point on the same block: both arrays are bare `as const` with no `NodeLabel` constraint (`schema.ts` never imports `NodeLabel`). A typo'd label — `'Fucntion'` — passes `tsc --noEmit` and is caught only by the new runtime test. The codebase's own precedent is one token away: `symbol-table.ts:64` and `symbol-table.ts:92` both close with `] as const satisfies readonly NodeLabel[];` precisely so adding a label without updating the companion table fails TypeScript.

Suggestion: either import the real sets (cheapest — deletes the drift class the unit test exists to detect) or move them to `gitnexus-shared/src/lbug/schema-constants.ts`, which already owns `NODE_TABLES`/`REL_TYPES` for exactly this "both sides must agree" reason. Either way, add `satisfies readonly NodeLabel[]`.

gitnexus/src/storage/repo-manager.ts
```diff
@@ -687,8 +687,20 @@ export interface RepoMeta {
...
+ * versions of this same fix; both left the rest of the cross product undeclared,
+ * so `analyze` kept aborting at `assertDeclaredPair` on the next codebase with a
+ * different edge shape (`Class→Variable` on Java static/field initializers was
+ * the report). Plus 10 hand-declared structural pairs the generated block cannot
+ * reach — COBOL containment and the Vue handler→component edge (#2789), whose
+ * endpoint labels are in neither scope-bridge set. A pre-v35 database physically
+ * lacks all of these, so force a full re-analyze.
  */
-export const INCREMENTAL_SCHEMA_VERSION = 34;
+export const INCREMENTAL_SCHEMA_VERSION = 35;
```
magyargergo, line 711: **This bump is what makes every existing index hit the `streamGraphEmitActive` ordering bug at once.** (The bug is in `run-analyze.ts`, an unchanged file — anchoring here because this line is its fleet-wide trigger.)

`run-analyze.ts:898` computes `const streamGraphEmitActive = resolveStreamGraphEmit(options);` at function entry, and `resolveStreamGraphEmit` (`run-analyze.ts:679`) opens with `if (options.force !== true) return false;`. The schema-version-mismatch block runs **361 lines later**, at `run-analyze.ts:1259-1266`, and rebinds the local: `options = { ...options, force: true };`. So on the v34→v35 forced rebuild the flag is already frozen `false`, and every existing repo takes the **non-streaming in-memory emit path** — losing the #2680 memory-streaming optimization added to avoid the #2649 kernel-scale OOM, for exactly the population most likely to be memory-constrained. Contrast `resolveStreamPdgEmit(options)` at `run-analyze.ts:1536`, read fresh *after* the force block in the same object literal, which behaves correctly.

Pre-existing (it has fired on every prior schema/pdg-mode/feature mismatch) and users can work around it with an explicit `--force`, so not a blocker — but this bump forces the path for every indexed repo simultaneously. Fix is to move the `resolveStreamGraphEmit(options)` call below the force-mutating mismatch blocks, mirroring its PDG sibling.

Separately, a merge-time checklist item this file's own comment already asks for: re-check `35` against `origin/main` immediately before merging. Free today (main is 34, the competing #2791 is closed), but this is the 8th time in the series. Worth considering hashing the sorted declared-pair set into `RepoMeta` alongside `schemaVersion` — a same-version/different-DDL pair silently drops edges through `fallbackRelationshipInserts`'s `catch { /* skip */ }` (`lbug-adapter.ts:1427-1429`), and a DDL hash would retire the renumbering ritual permanently.

gitnexus/test/integration/structural-pair-coverage.test.ts
```diff
@@ -0,0 +1,75 @@
...
+ * The undeclared FROM/TO pairs a fixture emits, deduped and sorted so a failure
+ * is stable and names the pair to declare.
+ *
+ * Mirrors `RelPairRouter.route`: derive both labels with the router's own
+ * `getNodeLabel` (from the id, which is what `assertDeclaredPair` actually
+ * sees, not the node's `label` field) and drop any edge whose endpoint is not a
+ * real node table, exactly as the router's `skipped` branch does.
+ */
+const undeclaredPairsIn = async (fixture: string): Promise<string[]> => {
+  const result = await runPipelineFromRepo(path.join(FIXTURES, fixture), () => {}, {
```
magyargergo, line 154: **All three fixtures are already piped through `runPipelineFromRepo` elsewhere in the suite; this re-runs them from cold worker pools.**

`resolvers/cobol.test.ts:29-32` (`cobol-app`), `resolvers/vue.test.ts:19` (`vue-basic`), `resolvers/php.test.ts:497` (`php-transitive-traits`). `cobol.test.ts`'s `skipGraphPhases` only disables the MRO/AOP/DI/communities/processes phases, not the COBOL containment/CALLS/ACCESSES emission this test cares about, so it is materially the same graph. Measured at 16.45 s of test-body time on one reviewer's host; `it.concurrent` overlaps the cost but does not remove it.

This is the one finding both review engines flagged independently. Suggestion: extract a shared declared-pair assertion and invoke it against the graphs those suites already build (or a fixture-path-keyed cache in `resolvers/helpers.ts`) — if the isolation is deliberate, a sentence in the file docblock saying why would stop the next reviewer re-raising it.

Two cheap hardenings while this file is open, both one line:
- The assertion can pass **vacuously**. `undeclaredPairsIn()` builds `emitted` from `result.graph.iterRelationships()`, filters to undeclared, asserts `[]` — a renamed fixture, a failed grammar load or a swallowed parse error yields an empty set and a green test. `expect(emitted.size).toBeGreaterThan(0);` closes it; asserting each fixture emits the specific pairs it exists to guard closes it properly.
- Lines 63-66 hand-mirror `RelPairRouter.route`'s routable/skipped policy (`rel-pair-routing.ts:~135-150`) rather than sharing it, so the guard silently validates a stale approximation if routing semantics change. The unit test has the same shape — `schema-pair-coverage.test.ts:61-66` re-declares `parseRelationSchemaPairs`'s regex as a second literal; if the real pattern ever widens, the dupe-guard degrades to `declared.size === declared.size` with no failing test. Exporting the pattern/helper and consuming it from both closes both.

magyargergo (commented):
## Multi-engine review — PR #2793 (head `da8eeb154`, base `911151e23`)

### 1. How this was reviewed, and what that's worth

Two independent engines, three review methods:

| Engine | Legs | Live? | Outcome |
|---|---|---|---|
| **Claude** | GitNexus PR-swarm (facts-historian, branch-hygiene, risk-architect, security-boundary, test/CI, docs-DoD) + ce-code-review personas (correctness, adversarial, testing, reliability, performance, maintainability, api-contract, project-standards) | yes — 14 lanes | 4 reproduced hard-aborts, 2 confirmed P2s |
| **Codex** (gpt-5.6-sol, xhigh) | same swarm, ce-code-review, plus a native adversarial pass | **all three legs live**, real findings, no "skill not found" markers | 1 P2 + 3 P3; `codex-adversarial` verdict `approve`, "No material findings" |

Honest caveat on lane count: **four Claude `gitnexus-*` specialist lanes (branch-hygiene, security-boundary, risk-architect, test/CI) stalled on first dispatch and were re-dispatched with hard tool budgets; two of them (branch-hygiene, test/CI) were re-run as `general-purpose` agents after a second stall.** Lane count is therefore *not* evidence of independence — weight below is by engine and by reproduction, not by how many files said the same thing. A shared-worktree contamination window was also observed mid-run (two lanes independently detected it); both re-pinned their harnesses to `git show da8eeb154:<path>` blobs, so no result below depends on worktree state.

Where a lane disagreed with a fact the coordinator executed personally, the coordinator's result wins.

### 2. What this PR gets right — including three suspicions that were probed and came back clean

This is a genuinely good structural fix, and most of the review budget went into trying to break it and failing.

- **The generated half is provably complete for its surface.** The adversarial lane harvested the *entire* fixture corpus through the real pipeline — 743 `lang-resolution` fixtures as one repo (19,761 nodes / 26,394 rels / 93 distinct pairs, with `pdg: true` so `BasicBlock` edges are included) plus all 25 non-`lang-resolution` fixture repos — and found **zero** undeclared pairs. The correctness lane ran an independent 105-fixture harvest: also zero. Both harnesses were mutation-checked (deleting `FROM Function TO File` reproduces exactly `Function|File`), so the clean result is not a broken harvester.
- **The pair counts are settled.** Evaluated through the real modules with `tsx`: **223 pairs on `main`, 332 at head, zero duplicates, base is a strict subset of head** (0 dropped). `+109 = 99 generated + 10 structural`, matching `repo-manager.ts`'s v35 comment.
- **"A duplicate pair would kill every `analyze`" — REFUTED at head.** 332 raw `FROM…TO` matches vs 332 unique pairs. `parseRelationSchemaPairs` makes backticks optional in both capture groups, so the block's mixed bare/backtick quoting dedups correctly. No `//` comment inside `STRUCTURAL_PAIR_DDL` (the known template-splice trap).
- **The DDL is accepted by a real LadybugDB.** One lane ran `NODE_SCHEMA_QUERIES` + `RELATION_SCHEMA` against `@ladybugdb/core` 0.18.3: `CREATE REL TABLE` succeeds, `CALL SHOW_CONNECTION('CodeRelation')` returns exactly 332 rows matching the parsed set, and a `COPY` of a `Class`→`Variable` row (the #2792 crash pair) round-trips.
- **The `||`-chain → `Set` refactors in `ids.ts` / `node-lookup.ts` are exact.** Same 8 and same 13 labels, same order, every per-label rationale comment carried across verbatim. Confirmed independently by four lanes across both engines.
- **The v34→v35 gate is correctly wired.** Strict `!==` in both directions, so an older *and* a newer stamp both force a rebuild; the full-rebuild branch wipes and recreates from the new DDL under the existing atomic-swap machinery. Downgrade is safe too.
- **`schema.ts` → `rel-pair-routing.ts` is acyclic** (`rel-pair-routing.ts` imports only `path`/`fs`/`events`/`stream/promises`), and no Node-builtin exposure reaches a browser bundle — `gitnexus-web` depends on `gitnexus-shared`, not on this package.
- **Security surface: clean.** No workflow/manifest/lockfile/hook churn, no bidi or zero-width characters, no secrets, no host paths. Every DDL operand is a compile-time literal. Both engines agree.
- **Actively refuted attack surfaces:** Community/Process/clustering endpoints; PDG/CFG/taint endpoints (`BasicBlock|BasicBlock` only); scope-graph emission under `INGESTION_EMIT_SCOPES`; Spring DI `INJECTS` with a `CodeElement` endpoint; Java nested/sealed records; node `label` vs id-prefix divergence (0 mismatches corpus-wide); the "racy `it.concurrent` + `vi.stubEnv`" hypothesis (`pool: 'forks'` gives each file its own `env`).

**Three findings raised by lanes were refuted by the coordinator and are deliberately absent from this review:** (a) "#2789 is not in the `Closes` list" — GraphQL `closingIssuesReferences` returns both #2789 and #2792, both auto-close on merge; (b) "issue-number drift between #2769 and #2781" — #2781 is the *issue*, #2769 is the PR that closed it, both citations are correct in context; (c) "`SCOPE_BRIDGE_SOURCE_LABELS` is over-broad, narrow it to `CALLER_ANCHOR_LABELS`" — **do not act on this**: `resolveCallerGraphId` returns `resolveDefGraphId(...)`, whose last-resort branch (`ids.ts:342`) is label-agnostic and resolves through a `simpleKey` that `node-lookup.ts:240-241` registers for *every* linkable node. Narrowing that list would reintroduce the crash class.

### 3. Headline finding — the hand-declared half is still sampled, and four `analyze` hard-aborts survive it

**This is Claude-only and coordinator-reproduced. Both Codex legs approved the PR and did not find these; `codex-adversarial` explicitly reports "no material findings". That is a real cross-engine disagreement, not consensus — weigh it accordingly.** What tips it past the bar is not lane count but reproduction: the coordinator independently re-ran **three of the four** end-to-end at PR head with the built `dist/` on the default user path (`node dist/cli/index.js analyze <dir> --index-only --skip-git`) and captured real stack traces. The fourth (`Class→Tool`) rests on a single lane's reproduction and is marked as such.

Equally important, and easy to lose: **all four are pre-existing on `main`. None is a regression introduced by this PR.** Base `main` is equally missing every one of these pairs.

| Pair | Trigger | Languages | Evidence |
|---|---|---|---|
| `Method→Annotation` | Spring `@Bean` + `@ConditionalOnMissingBean` | Java **and** Kotlin | `[reproduced]` — coordinator + adversarial lane |
| `Method→File` | Vue **Options-API** `methods: {}` handler bound to a child-component event | Vue | `[reproduced]` — coordinator + two Claude lanes independently |
| `Namespace→Record` | COBOL `DECLARATIVES` / `USE AFTER STANDARD ERROR ON <file>` | COBOL | `[reproduced]` — coordinator + two Claude lanes, two separate harnesses |
| `Class→Tool` | `@mcp.tool()` applied to a class | Python (MCP) | `[reproduced]` — **adversarial lane only; the coordinator did not reproduce this one** |

The Vue case is the sharpest evidence, because it is *this PR's own new line*. `schema.ts:423` adds `FROM Function TO File` for the `BINDS_EVENT_HANDLER` emitter at `languages/vue/scope-resolver.ts:275-293`. That emitter's source is `nodeLookup.get(simpleKey(parsedFile.filePath, handlerName))` — a raw lookup hit, i.e. **any of the 13 `LINKABLE_LABELS`**. `<script setup>` mints a `Function` (declared, passes); the Options API mints a `Method` (undeclared, aborts); a `<script setup>` `const config = {…}` bound as a handler mints a `Const` (also aborts). One of thirteen got declared, because one fixture happened to exercise it:

```
Analysis failed: Relationship label pair Method→File is not declared in the LadybugDB relation schema
    at assertDeclaredPair (dist/core/lbug/rel-pair-routing.js:75:15)
    at RelPairRouter.route (dist/core/lbug/rel-pair-routing.js:132:9)
    at streamAllCSVsToDisk (dist/core/lbug/csv-generator.js:669:43)
```

**Why both new guards are structurally blind to all four.** `test/unit/schema-pair-coverage.test.ts:25-28` derives its requirement from `['File', ...LINKABLE_LABELS] × (LINKABLE_LABELS ∪ CALL_TARGET_TYPES)`. Neither `Annotation`, nor `Tool`, nor `File`-as-*target* is a member of the target set, so no pair with those endpoints can ever appear in `requiredPairs()`. `NON_BRIDGE_CORPUS` is three fixtures with no Spring conditional, no Options-API Vue event binding, no `DECLARATIVES`, no MCP tool on a class. **All 16 tests in the two new files pass while all four crashes are live.** These findings live in a third region neither guard describes: *one endpoint resolved through the bridge (so it wears a `LINKABLE_LABELS` label), the other a structural node type outside both sets.* Enumerating that region by emitter turns up roughly 40 more undeclared pairs nobody built a trigger for.

The recommended direction is a *closed-form* requirement for that region rather than another point-list — declare `({File} ∪ LINKABLE_LABELS) × {Annotation, Tool, Route, Section, CodeElement, Record, Namespace, File}` and assert it in the unit test from the same constants. That is ~112 additional pairs (→ ~444), still far below the 1024-pair point where this repo's own benchmark measured ~1.8× on untyped-endpoint anchored queries. Alternatively, grow `NON_BRIDGE_CORPUS` to one fixture per structural emitter.

Given all four are pre-existing, a defensible call is to **merge this PR as the strict improvement it is and open a follow-up issue** with the four reproduced triggers — but the PR's stated goal is to end this bug class, so shipping it while the next four "analyze died on my repo" reports are already reproducible deserves a conscious decision, not silence.

### 4. Inline findings

Six comments recommended (see plan at end):

1. **`schema.ts:423`** — P1, `[reproduced]`. Vue Options-API `Method→File` hard-abort; the declared pair is one of 13 the same emitter line can produce.
2. **`schema.ts:417`** — P1, `[reproduced]`. COBOL `DECLARATIVES` emits `Namespace→Record` (`cobol-processor.ts:1055-1070`); `Namespace|Function` is the only `Namespace`-source pair added. `Namespace` is not linkable, so the generated block can never reach it. 25 of COBOL's 64 label pairs remain undeclared.
3. **`schema.ts:316`** — P1/P2, `[reproduced]`. The systemic finding: the docblock names the corpus test as "the guard", but the guard's requirement and its corpus are both blind to the bridge-endpoint × structural-node region. Spring `Method→Annotation` and Python `Class→Tool` reproduced there.
4. **`schema.ts:281`** — P2, `[code-read]`, coordinator-confirmed. The stated justification for hand-copying the twin arrays — "sets that live in the ingestion layer, which must not be imported here" — is **false**: `csv-generator.ts:22-23` and `lbug-adapter.ts:67-68`, both in this same directory, already import from `../ingestion/`. Same comment covers the missing `satisfies readonly NodeLabel[]`.
5. **`repo-manager.ts:703`** — P2, `[code-read]`, coordinator-confirmed. The v35 bump is what makes every existing index hit the pre-existing `streamGraphEmitActive` ordering bug at once (true location: `run-analyze.ts:898` vs `:1259-1266`). Plus the merge-time version re-check.
6. **`structural-pair-coverage.test.ts:55`** — P2/P3, cross-engine. Three fixture pipelines re-run from cold worker pools that sibling suites already built.

### 5. Adjudicated tensions, lower-priority items, refuted suspicions, test gaps

**One-directional twin-drift guard — cross-engine agreement on existence, disagreement on severity.** Both engines found it independently. `requiredPairs()` asserts `required ⊆ declared`, so a label *added* to `LINKABLE_LABELS`/`CALL_TARGET_TYPES` fails CI (mutation-verified: dropping `Macro` from the twin fails with 9 missing pairs), but a label present only in the *twin*, or removed from the real set while the twin keeps it, is never caught (mutation-verified: `Class→Record` added twin-only → all six assertions pass). **Claude's test/CI lane rates this P2** on pair-surface-inflation grounds, citing this repo's own measurement that 1024 pairs costs ~1.8× on anchored queries and arguing this is the leak that lets a derived surface silently inflate. **Codex's adversarial leg rates the uncovered direction "harmless over-declaration"** — an undeclared pair aborts, an unused declared pair does not. Both are right about the mechanism. Split the difference: it is not a crash risk, it is a slow-growth risk, and the fix is two lines either way — `expect([...SCOPE_BRIDGE_SOURCE_LABELS].sort()).toEqual(['File', ...LINKABLE_LABELS].sort())` collapses the class in both directions. Codex's own "next steps" independently suggests exactly this.

**Duplicated fixture pipeline runs — the cleanest cross-engine corroboration in the run.** Claude's ce-testing and test/CI lanes and Codex's ce leg (which *measured* 16.45 s of test-body time) and Codex's swarm leg all landed on it independently. `cobol-app`, `vue-basic` and `php-transitive-traits` are already piped through `runPipelineFromRepo` in `resolvers/cobol.test.ts:29-32`, `resolvers/vue.test.ts:19`, `resolvers/php.test.ts:497`. Codex rates it P2, Claude P3; it is pure CI cost, not correctness. Related and same-theme: **the guard re-implements production policy instead of sharing it** — the integration test hand-mirrors `RelPairRouter.route`'s routable/skipped gate (Codex swarm F2) and the unit test hand-copies `parseRelationSchemaPairs`'s regex (Claude ce-testing F1). Both drift silently if the router changes.

**`Delegate` reachability — genuinely open, presented rather than decided.** `claude-ce-correctness` argues the 14 `<source>|Delegate` pairs are unreachable dead weight: both `tryEmitEdgeWithExplicitTargetId` call sites bottom out in registries (`methods`/`fields`) or graph indexes that classify `Delegate` as `callable-only` and never register it. `claude-swarm-risk-architect` and Codex's swarm leg argue the opposite — `pickOverload`'s candidate set *is* `CALL_TARGET_TYPES`, which contains `Delegate`. Both rate it P3, and no lane reproduced either way. The safe resolution: **over-declaration is the safe direction — do not remove `Delegate` on the strength of this**. If the path really is unreachable, fix the *comment*, which currently asserts a reachability that may not hold and is load-bearing documentation for whoever next edits these sets.

**Lower-priority, body-only:**

- **`assertDeclaredPair`'s error is unactionable for a user** (`rel-pair-routing.ts:87-93`). It names only the abstract label pair — no node ids, no file, no symbol, no relationship type — and `cli/analyze.ts` special-cases ~10 failure classes but not this one, so it falls to the generic `writeFatalToStderr('Analysis failed', err)` with a stack trace into GitNexus internals and no "Suggestions:" block. Single Claude lane, so demoted out of the inline set, but given that four triggers are still live this is the cheapest way to make the *next* report diagnosable.
- **The corpus guard can pass vacuously.** `undeclaredPairsIn()` collects emitted pairs, filters to undeclared, asserts `[]`. If a fixture were renamed, a grammar failed to load, or a parse error were swallowed, the emitted set is empty and the test passes green. One line fixes it: `expect(emitted.size).toBeGreaterThan(0)`, or better, assert each fixture emits the specific pairs it exists to guard.
- **Nothing stamps *which* DDL a database was built with.** `INCREMENTAL_SCHEMA_VERSION` is a bare integer; a same-version/different-DDL pair silently drops edges through `fallbackRelationshipInserts`'s `catch { /* skip */ }` (`lbug-adapter.ts:1427-1429`). The concrete trigger is gone — PR #2791 (which also claimed v35) is **CLOSED**, and `origin/main` is still 34 — but hashing the sorted declared-pair set into `RepoMeta` would retire the "8th collision, re-check before merge" ritual permanently.
- **`scopeBridgePairDdl()` has no internal dedup** — it dedups only against `STRUCTURAL_PAIR_DDL`, not within the generated block. Verified clean at head (332/332), but one duplicated entry in a hand-copied twin makes LadybugDB reject the `CREATE REL TABLE` and no repo indexes at all — strictly worse blast radius than a missing pair. Building through a `Set` makes it impossible by construction.
- **PR body is stale in three places** (cross-engine): "223 → 322" should be **332**; `Function→File` is attributed to `vue-sfc-extractor.ts` but is emitted by `languages/vue/scope-resolver.ts:275-293` (commit 3 fixed the code comment, not the body); "4 tests" is 3 parameterized cases. Prose only — the DDL is correct.
- **Commit hygiene, non-blocking:** `da8eeb154` is labelled `refactor(test):` but rewrites prose in two `src/` files owned by earlier commits and adds a new 10-string assertion, and it corrects a factual error introduced by commit 2 — bisecting `c17fd82d7` lands on a wrong-attribution comment. Squash into `c17fd82d7`, or ignore if the repo squash-merges. Author email differs across commits (`icloud` vs `gmail`); cosmetic.
- **`Record` remains absent from `LINKABLE_LABELS`** (pre-existing, flagged by both engines). Its branches in `buildGraphNodeLookup` and `resolveDefGraphId` are dead, and a `Record` def degrades through the label-agnostic `simpleKey` fallback. Out of scope here — but adding it later grows the generated block by 29 pairs and needs its own schema bump.
- **`describe.skip` local-signal gap:** `distWorkerExists()` makes the corpus guard skip silently under a bare `npx vitest run` with no prior build. **This was checked and is *not* a CI hole** — `ci-tests.yml` calls `setup-gitnexus` with `build: 'true'` before an unfiltered `npx vitest run`, so the guard genuinely runs. Local-only.

**Test gaps worth recording:** no fixture exercises a Spring conditional annotation end-to-end (all four `spring-*-app` fixtures emit zero `Annotation`-endpoint edges); `vue-basic` covers only `<script setup>` (an Options-API component exists but with no component event binding, which is exactly why the harvest saw `Function|File` and nothing else); no `@mcp.tool()` on a non-function; no `DECLARATIVES` anywhere under `test/fixtures/`; no test asserts a *physical* database's pair set against the in-code `RELATION_SCHEMA` (one lane ran that by hand in ~1 s — it would close a whole class).

### 6. CI status

**40 real checks pass on `da8eeb154`.** The only red status is a Vercel deploy-authorization stub ("A member of the Team first needs to authorize it") — a deployment gate, not a code check — which is what produces GitHub's `UNSTABLE` rollup. `e2e (chromium)` correctly skips (no `gitnexus-web` changes). CI Report bot: **16,210 passed / 0 failed / 72 skipped**, all skips in known categories, none schema-related. `mergeable=MERGEABLE`; branch is exactly current with `origin/main` (merge-base == base SHA), linear, no merge or revert commits, zero prior reviews or inline comments. Locally: `tsc --noEmit` clean, 13 unit tests and 3 integration cases pass, `git diff --check` clean.

**One merge-time action:** re-check `INCREMENTAL_SCHEMA_VERSION = 35` against `origin/main` immediately before merging. It is free today (main is 34, the competing #2791 is closed), and the file's own comment records seven prior collisions.

### 7. Coverage limits

- The four hard-aborts were found by **hand-constructing feature combinations absent from the fixture corpus**. Both corpus harvests came back clean. There is no reason to believe the four are exhaustive — the same enumeration points at ~40 more undeclared pairs in the same region with no trigger built for them.
- **Codex ran under a restricted read-only sandbox**: its ce leg could not create a scratch directory (so its run artifacts and cross-model pass were skipped) and its swarm leg deliberately did not run local npm/vitest, relying on head CI for executable validation. That is a plausible mechanical reason the Codex legs did not construct the repro repos the Claude adversarial lane did — it is not evidence the Claude findings are wrong, nor evidence they are right.
- Nothing here exercised a real LadybugDB *at CI time*; the physical-schema round-trip was a one-off by hand.
- Crash-mid-rebuild fault injection on the atomic-swap path was reasoned about, not executed.
- Windows/macOS curated-subset membership for the new integration file was not checked (the Ubuntu full-suite shard is dispositive for whether it runs at all).

### 8. Verdict

- **Branch hygiene: `clean feature/fix PR`.** Merge-base equals the base SHA, linear, three commits, no merge or revert commits, no unrelated churn, no generated-file or lockfile noise.
- **Merge state: `mergeable`.** GitHub's `UNSTABLE` rollup is produced solely by the Vercel deploy-authorization stub.
- **Final verdict: `production-ready with minor follow-ups`.**

The change it claims to make, it makes: the generated half of the DDL is provably complete for the scope-resolution bridge, two independent full-corpus harvests found zero undeclared pairs, the DDL round-trips through a real LadybugDB, and the predicate→Set refactors are exact. Nothing here is a regression — every problem below is equally present on `main`. What it does not do is end the bug class, which is its stated goal: three coordinator-reproduced `analyze` hard-aborts (and a fourth from one lane) survive in the hand-declared half, in a region neither new guard can describe, and roughly 40 more undeclared pairs sit in that same region untriggered. Because those are pre-existing, merging is a strict improvement and the defensible path is to ship this and open a follow-up with the reproduced triggers — but that should be a deliberate call, since the next "analyze died on my repo" reports are already reproducible today. The two P2s worth acting on before or shortly after merge are the false layering justification behind the hand-copied twin arrays and the `streamGraphEmitActive` ordering bug this bump triggers fleet-wide.

---

*Automated multi-tool digest (Claude swarm + ce personas, Codex gpt-5.6-sol xhigh × 3 legs, coordinator-verified facts). Verify each finding against the code before acting on it — in particular, the four hard-aborts are Claude-only plus coordinator reproduction, and both Codex legs approved without finding them.*

---

magyargergo:
Pushed `81daf370e` addressing the review above.

**All four hard-aborts are fixed and re-verified end-to-end** on the default user path. One correction to my own review: it recorded `Class→Tool` as reproduced by a single lane only — I have since reproduced it directly too (`@mcp.tool()` on a Python class), so the tally is **4 of 4**, not 3.

**The root-cause model in the review was itself wrong.** I proposed `({File} ∪ LINKABLE_LABELS) × {Annotation, Tool, …}` — "bridge endpoint × structural endpoint". That does not fit `Namespace→Record`, which is structural on *both* sides. The property that actually holds is that the **anchor is a lookup result, not a literal at the emit site**, so the emitter cannot constrain its label. Hence `DEFINITION_ANCHOR_LABELS × ATTACHMENT_TARGET_LABELS`, with the anchor set derived from `NODE_TABLES` by subtraction so a new node table joins automatically.

A ground-truth harvest — all three `assertDeclaredPair` call sites instrumented, every fixture repo run individually, plus two `--pdg` runs — confirms **exactly four** pairs were undeclared corpus-wide, and none remain. (Worth recording: `analyze <repo-root>` never covers `test/fixtures`, because the default ignore rules drop paths containing `test`/`fixtures`. Each fixture must be its own root, which is why earlier sweeps looked clean.)

Beyond the six inline findings:

- **161 of the 233 hand-declared pairs were already generated by the rules.** Deleted — declared set byte-identical at 450, hand-list 233 → 72. They were load-bearing shadow: the generator suppresses anything already declared structurally, so narrowing a rule later would silently keep pairs alive with nothing failing. A new guard now fails CI if a hand-declared pair is re-added inside a rule — the gap that let this PR's own commit 2 re-commit the pattern commit 3 exists to abolish.
- **The 450-pair cost is now measured, not asserted**, with the harness committed to `gitnexus/bench/schema-pairs`: 0.93–1.05× vs 332 on untyped anchored queries, against 1.22–1.43× at 641 and 2.03–2.34× at 1024. It reproduces the known #2792 cliff, which is what makes the 450 figure trustworthy. The comment's "786" for a third rule was wrong — it is **641**.
- **Three docblocks were factually wrong** and are corrected: two claimed pairs were "outside every rule" when 4-of-10 and 9-of-11 respectively were rule-generated.

Two review suggestions I did **not** take, with reasons:

1. **Narrowing `SCOPE_BRIDGE_SOURCE_LABELS` to `CALLER_ANCHOR_LABELS`** (raised as ~55 dead pairs). `resolveDefGraphId`'s last-resort `simpleKey` lookup is label-agnostic, so a caller-anchor def genuinely can resolve to a `Const`/`Variable` node. Narrowing would reintroduce the crash class.
2. **Dropping `Community`/`Route` from the attachment targets** (~47 unreachable pairs). The observation is right — those two anchors *are* label-gated, and the docblock claiming otherwise is now corrected. But re-narrowing a pair set on one reading of one predicate is how this bug class started, and at a measured 1.0× the headroom is free. The comment now says plainly which anchors are gated and that the extra pairs are deliberate.

Filed as follow-ups rather than scope-creeping this PR: accumulating *all* undeclared pairs before aborting instead of dying on the first (this PR is the cost proof — four crashes found serially); a DDL fingerprint in `RepoMeta` to retire the `INCREMENTAL_SCHEMA_VERSION` renumbering ritual; Rust/C++ fixtures for the 17 still-unguarded `Impl|*`/`Template|*` pairs; and `forceReasons[]` so the stream-emit ordering invariant stops depending on statement position.

<sub>Automated multi-tool work; verify before relying on it.</sub>

gitnexus/src/core/lbug/schema.ts
magyargergo: Fixed in `81daf370e`. Rather than adding `Method|File`, the column is now generated: the emitter’s anchor is a lookup result, so `DEFINITION_ANCHOR_LABELS × ATTACHMENT_TARGET_LABELS` covers `File` as a target for every definition label. `vue-basic/OptionsHost.vue` added, and `vue-basic` was already in the corpus so the guard catches it. Repro now indexes clean (4 nodes / 5 edges).

gitnexus/src/core/lbug/schema.ts
magyargergo: Fixed in `81daf370e`. `Namespace|Record` falls out of the attachment rule (`Namespace` is a definition anchor, `Record` an attachment target), so it is generated, not hand-added. New `cobol-declaratives` fixture reproduces the `DECLARATIVES` / `USE AFTER STANDARD ERROR ON <file>` path and is in `NON_BRIDGE_CORPUS`; repro indexes clean (10 nodes / 10 edges). Note it also shifted the `bench/scope-capture` COBOL fingerprint (that bench globs `cobol-*`) — re-baselined in `4cc413637`, verified capture-neutral.

gitnexus/src/core/lbug/schema.ts
magyargergo: Fixed in `81daf370e`. The comment was right that the model was wrong — though not in the way I proposed. "Bridge endpoint × structural endpoint" does not fit `Namespace→Record`, which is structural on both sides. The actual invariant is that the anchor is a lookup result, giving a second closed-form rule. A ground-truth harvest (all three `assertDeclaredPair` call sites instrumented, every fixture repo run individually, plus two `--pdg` runs) confirms exactly four pairs were undeclared corpus-wide and none remain.

gitnexus/src/core/lbug/schema.ts
magyargergo: Fixed in `81daf370e`. Twins deleted — `schema.ts` imports `LINKABLE_LABELS` and `CALL_TARGET_TYPES` directly; no cycle, and `gitnexus-web` depends only on `gitnexus-shared` so nothing reaches a browser bundle. The remaining hand-written sets are annotated `readonly NodeTableName[]`, which is narrower than the `NodeLabel` precedent suggested here and rejects both a typo and a non-table label at compile time.

gitnexus/src/storage/repo-manager.ts
magyargergo: Fixed in `81daf370e`. `resolveStreamGraphEmit(options)` now resolves after every guard that rebinds `options.force` (there are eight; the last honourable one is the CJK-mode change — the post-pipeline feature re-check cannot be honoured by any placement, same as its PDG sibling). Regression test mutation-checked: fails at the old position, passes at the new one. Merge-time re-check done on the current head: `origin/main` is 34, branch is 35, no collision; `SCHEMA_BUMP` correctly untouched at 39.

gitnexus/test/integration/structural-pair-coverage.test.ts
magyargergo: Not changed, deliberately — measured rather than assumed. `vitest.config.ts` sets `pool: 'forks'` with default isolation, so a per-file cache shares nothing across files; and `resolvers/cobol.test.ts` builds `cobol-app` with `skipGraphPhases: true`, which drops the phase-emitted edges this guard exists to check — `HANDLES_TOOL` (the `Class|Tool` case) is itself a pipeline phase, so asserting on that graph would cover strictly less. Measured cost of the alternative: ~1s of ~24s wall, the rest being transform/import which no restructuring removes. The reasoning is now a NOTE in the file so it is not re-raised. The suite did gain six fixtures and per-fixture sentinel pairs, so it fails loudly instead of passing vacuously on an empty graph.

magyargergo:
Follow-ups from this review are filed rather than scope-creeping the PR:

- #2797 — `assertDeclaredPair` aborts on the first undeclared pair, not all of them. This PR is the cost proof: four were live at once and each needed its own reproduction repo. Also records the throw-vs-degrade rationale, which the current docblock argues from the weaker COPY case.
- #2798 — replace the hand-incremented `INCREMENTAL_SCHEMA_VERSION` correctness check with a DDL fingerprint. Same disease this PR cured one file over, and cheaper now that the DDL is generated.
- #2799 — no Rust/C++ fixture guards the 17 hand-declared `Impl|*` / `Template|*` pairs. Includes the two fixture-name glob traps (`python-*` golden, `bench/scope-capture` `<lang>-*` fingerprint) that bit during this PR.
- #2800 — `run-analyze` should derive `force` from accumulated reasons; the ordering fix here is defended by a comment plus a test pinning 1 of 8 guard sites.
- #2801 — `Record` absent from `LINKABLE_LABELS` makes two branches dead code (pre-existing, noted in the PR body; changing it is a resolution change plus a schema bump, so not a rider).
- #2802 — MCP startup eagerly loads the analyze-only ingestion closure via `csv-generator`. Records that the obvious fix (re-homing the constant) does **not** work, and that the measurement should be redone on a native filesystem first.

None block this PR.

PR #1990, opened 2026-06-03
perf(cpp): index ADL candidates once instead of per-site rescans

magyargergo:
## Problem

C++ scope-resolution `emit` dominates large-repo analysis. On a 5,969-file C++ repo the run spent **~6.76h in the scope-resolution emit phase** — ~70% of the entire 9.76h analysis:

```
extract=713ms finalize=66514ms propagate=18839ms resolve=3548ms emit=24347727ms (5969 files)
[scope-resolution:cpp] 5969 files → 28184 IMPORTS + 60760 reference edges (650298 unresolved sites)
```

`resolve` (registry lookup) was only 3.5s, so the cost was entirely in **fallback edge emission**.

## Root cause

`pickCppAdlCandidates` (argument-dependent lookup) runs **once per unresolved ADL-eligible call site**, and each invocation:

- rescans **every parsed file** (rebuilding a per-file scope map each call),
- scans **every workspace def** via `findCppClassDefBySimpleName`, and
- uses an **O(scopes²)** child-scope walk for hidden friends.

That is `O(unresolved sites × files)`. With ~650k unresolved C++ sites the emit phase goes super-linear.

## Fix

Build an `AdlCandidateIndex` **once per pipeline run** (lazy, guarded by `parsedFiles` identity, reset in `clearCppAdlState`) and query it per site:

| Index | Replaces |
|---|---|
| `classDefsBySimple` | per-site all-defs scan (preserves `defs.byId` order → identical first-match / ambiguous semantics) |
| `nsCandidates` | per-site all-files namespace scan (keeps inline-namespace transparency) |
| `friendCandidates` | per-site hidden-friend scan; a parent→children index removes the O(scopes²) walk |
| `nsFunctionsByQName` / `nsFunctionsBySimple` | per-site all-files function-reference scan |

A monotonic `seqByNodeId` (file-major; namespace defs before friend/member defs within a file) lets the per-site query merge candidates across associated namespaces, dedup by `nodeId`, and sort — reproducing the **exact** legacy candidate set and order.

Per-site cost drops from `O(sites × files)` to `O(associated namespaces)`; the index build is a single `O(scopes + defs)` pass.

## Results

Benchmark (fixed 80 files, growing unresolved ADL sites), `emit` phase only:

| unresolved sites | before | after |
|---|---|---|
| 500 | 87 ms | **7 ms** |
| 1000 | 232 ms | **9 ms** |
| 2000 | (≫, linear) | **17 ms** |

[...]

Issue #1564, opened 2026-05-14
LANG-cpp: Remaining work after PR #1520 scope-resolution parity


magyargergo:
Tracking issue for the C++ scope-resolution gaps that PR #1520 (`cpp-scope-resolution-parity`) intentionally left out of scope.

Parent: #935 (LANG-cpp migration). Related: #909 (RFC scope resolution).

> **Status (verified 2026-05-16 against `main` @ f28185d6):** 12 of 14 checklist bullets are **done on `main`**. The two remaining bullets are in flight as [#1578](https://github.com/abhigyanpatwari/GitNexus/issues/1578) / PR [#1606](https://github.com/abhigyanpatwari/GitNexus/pull/1606) and [#1579](https://github.com/abhigyanpatwari/GitNexus/issues/1579) / PR [#1623](https://github.com/abhigyanpatwari/GitNexus/pull/1623). Local verification: `REGISTRY_PRIMARY_CPP=1 npx vitest run test/integration/resolvers/cpp.test.ts` — **215/215** on `main`, **218/218** on PR #1606, **220/220** on PR #1623.

## Context

PR #1520 plus its follow-up work shipped the C++ scope-resolution path as a **scope-aware approximation** of ISO C++ name lookup. Integration tests pass under `REGISTRY_PRIMARY_CPP=1`, but several documented divergences from full ISO semantics remain (see open bullets below).

Each item below is independently scoped. Completed items are checked off with links to the sub-issue or merge PR.

---

## ISO C++ conformance posture

GitNexus's C++ resolver is **not an ISO-conforming C++ front-end**. It is a **graph-safe under-approximation** of ISO C++ name lookup with a clear failure-mode preference:

> When in doubt, **suppress** (emit zero edges) rather than **guess** (emit a possibly-wrong edge).

This matters for downstream consumers (impact analysis, refactor safety, wiki) that rely on the graph not lying — false negatives are recoverable by re-reading the source; false positives silently mislead.

### Target language version

The resolver targets **shared C++17 lookup/overload semantics** plus a partial **C++20 constraint posture** delivered by #1579 / PR #1623. The implementation is intentionally syntactic — no template instantiation, no constexpr evaluator. Where ISO demands richer semantics, the resolver either documents the approximation or suppresses.

### Where the resolver matches ISO (implemented on `main` @ `f28185d6`)

| Rule | ISO clause | Implementation |
|------|-----------|----------------|
[...]

PR #2794, opened 2026-08-02
perf(cpp): index qualified namespace members once per pipeline run (#2788)

### LINKED ISSUE #2788: resolveCppQualifiedNamespaceMember rescans all parsedFiles per call site — quadratic emit phase, 26 min on a 1.5k-file C++ repo (1.6.6 & 1.6.9)
### Area

### Summary

resolveCppQualifiedNamespaceMember rescans all parsedFiles (and rebuilds a per-file scopesById Map) once per qualified call site, making the emit / "linking symbols" stage quadratic — 26 of 33 total minutes on a 1.5k-file C++ repo.

### Context

We index large proprietary C++ repos with gitnexus analyze in a production pipeline. The repos use namespace x { inline namespace v { … } } (ABI-versioning idiom) and a pervasive ns::member() qualified call style.

This looks like the exact same class of bug that #1990 fixed for ADL (pickCppAdlCandidates: per-call-site full rescan → one-shot AdlCandidateIndex). The qualified-name path (core/ingestion/languages/cpp/inline-namespaces.ts, resolveCppQualifiedNamespaceMember) still does the per-site rescan. Likely the remaining cause behind #2243 / #1973-style "stuck while linking symbols" reports. The hot loop is unchanged up to and including 1.6.10-rc.141 (checked the published tarball).

Fix suggestion (mirrors #1990): build a one-shot inverted index nsName → [{scope, scopesById}] per pipeline run (lazily, keyed by the parsedFiles array identity via WeakMap), preserving the original traversal order so candidate collection order / dedup / ambiguity semantics stay unchanged; keep narrowOverloadCandidates (call-site dependent) outside the index. We run exactly this as a local dist patch in production: emit 25m51s → 3.0s, total analyze 33m43s → 7m31s, and full node/edge dumps (all business fields, sorted, md5) are byte-identical with and without the patch on both 1.6.6 and 1.6.9. Happy to share the patch diff and the .cpuprofile.

### Expected behavior

Emit / "linking symbols" should scale roughly linearly with call-site count, with analyze time dominated by parsing — i.e. qualified-name lookups should hit a prebuilt index (as ADL does since #1990), not rescan the whole workspace per call site.

### Actual behavior

On a 1,473-file C++ repo, scopeResolution takes ~25.5 min, 99.2% of it in emit, and CPU profiling attributes 75.1% of total self-time to the single function resolveCppQualifiedNamespaceMember. Superlinear scaling is visible across language buckets of the same run:
| files in bucket | emit time |
|---|---|
| 9 | 12 ms |
| 113 | 323 ms |
| 1,473 | **25.3 min** |

### Steps to reproduce

1. Take a large C++ repo (~1.5k files) with heavy use of `ns::member()` qualified calls and inline namespaces (ours is proprietary; any repo with this style should reproduce — emit cost grows ~quadratically with qualified call sites × files).
2. Run: `PROF_SCOPE_RESOLUTION=1 gitnexus analyze --force --index-only --verbose`
3. Observe the `[scope-resolution prof]` line: `emit` dwarfs every other stage (ours: `emit=1516038ms` of `total=1528718ms`).
4. Optional: rerun with `NODE_OPTIONS="--cpu-prof --cpu-prof-dir=/tmp/prof"` — the top self-time frame is `resolveCppQualifiedNamespaceMember` (`languages/cpp/inline-namespaces.js:108`).

### Environment

- OS: Linux x86_64 (container)
- Node: v22.22.0
- Browser (if applicable): N/A (CLI)
- Commit / version: gitnexus **1.6.6** and **1.6.9** (both reproduce; the function is byte-identical in both apart from the overload-narrowing call). Hot loop still present in **1.6.10-rc.141**.

### Logs / screenshots

`PROF_SCOPE_RESOLUTION=1` breakdown on 1.6.9 (1,473-file cpp bucket):

```
[scope-resolution prof] extract=653ms finalize=6007ms propagate=5072ms resolve=947ms emit=1516038ms total=1528718ms (1473 files)
```

`--cpu-prof` top self-time (33-min run, main thread):

```
1491.6s  75.1%  resolveCppQualifiedNamespaceMember  languages/cpp/inline-namespaces.js:108
 286.2s  14.4%  (idle)
 127.9s   6.4%  (garbage collector)
  37.8s   1.9%  loadParsedFilesForPaths             dist/storage/parsedfile-store.js:203
```

Same run with the local inverted-index patch applied (output verified byte-identical):

```
[scope-resolution prof] extract=698ms finalize=6143ms propagate=5093ms resolve=944ms emit=3036ms total=15914ms (1473 files)
```

magyargergo:
Fixes #2788.

## Problem

`resolveCppQualifiedNamespaceMember` walked **every** parsed file — rebuilding a per-file `scopesById` map each time — once per qualified `ns::member()` call site, so the scope-resolution emit phase cost `O(callsites × scopes)`. The reporter measured 25.3 min of a 33-min analyze on a 1,473-file C++ repo, with 75% of total self-time in this one function:

```
[scope-resolution prof] extract=653ms finalize=6007ms propagate=5072ms resolve=947ms emit=1516038ms total=1528718ms (1473 files)
1491.6s  75.1%  resolveCppQualifiedNamespaceMember  languages/cpp/inline-namespaces.js:108
```

Two call sites hit it per site: `resolveQualifiedReceiverMember` (every `ns::member()`) and `resolveAdlCandidates` (per matching named `using`, with no callsite).

## Fix

Qualified lookups are served from a per-pipeline index instead of a per-call-site workspace scan. The index is a **lazily-queried node graph**: each `Namespace` scope contributes a node holding its own callables bucketed by member simple name plus direct links to its inline-namespace children; `(receiver, member)` results are resolved on demand and memoized. Build is O(scopes + defs); repeat call sites cost two Map lookups.

It is keyed in a `WeakMap<readonly ParsedFile[], QualifiedNsMemberIndex>`, following `moduleScopeIndexByPass` in `cpp/file-local-linkage.ts`. `clearCppInlineNamespaces` still swaps in a fresh WeakMap, because the index has a second input — the module-level `inlineNamespaceScopeIds` — that the key cannot observe.

Semantics are unchanged: non-inline nested namespaces are still not descended into, same-name hits across inline children still return `'ambiguous'` (#1564), and call-site narrowing still runs. Candidate order is preserved for byte-identity with the pre-#2788 walker, though it is not observable through this resolver's return contract (`allHits[0]` is only reached at length 1, and both tail branches return `'ambiguous'`).

## Results

`PROF_SCOPE_RESOLUTION=1 analyze --force --index-only` on a synthetic corpus (`namespace ns_i { inline namespace v1 { ... } }` plus 20 `ns_j::fn()` call sites per file):

| files | emit before | emit after |
|-------|-------------|------------|
| 100   | 153ms       | 16ms       |
| 200   | 704ms       | 24ms       |
| 400   | 3,293ms     | 42ms       |
| 800   | 16,898ms    | **78ms**   |

Before, doubling the file count quadrupled emit. Now it doubles. At 800 files total scope resolution goes 17.2s → 394ms.

## Correctness

**840,000 differential probes** against the pre-#2788 walker extracted from the base commit — **0 mismatches** — plus **48,000 candidate-order comparisons**, also 0. Generators covered multi-file workspaces, namespaces reopened across files, inline chains to 3,000 levels, same-name inline nesting, scopes with no `Namespace` def, `qualifiedName` `undefined`/`''`/trailing-dot, `''` receivers and members, non-callable def types, and call sites with and without a `Callsite`.

`test/integration/resolvers/cpp.test.ts` passes **334/334** — it already covers transitive inline resolution, the versioned `v1`-inline / `v0`-not-inline split, nested STL-style `__1` namespaces, ambiguity suppression with its structured `inline-ns-ambiguous` reason, call-site narrowing across distinct signatures, and ADL participation.

One divergence class exists and is unreachable: with two *distinct* defs sharing a `nodeId`, the legacy `seenNodeId` dedup and the new node-identity dedup differ. `scope-extractor.ts` builds `nodeId` as `def:${filePath}#${line}:${col}:${type}:${name}`, so distinct defs necessarily differ, and `phase.ts` keys `preExtractedByPath` by path so a file cannot appear twice.

## Guards

`bench/cpp-qualified-ns/` gates the qualified-receiver path in an always-on CI step, with a scaling arm and a fingerprint arm.

The corpus mirrors the production receiver distribution — about 1 in 5 receivers name a declared namespace, the rest are plain identifiers that name none — because Case 1.5 in `receiver-bound-calls.ts` is reached by every plain-identifier receiver call and misses on most of them. It also covers a namespace reopened across files, a same-name inline nest, a member declared at both namespace and inline-child level, and call sites carrying a real `Callsite`.

That distribution is load-bearing: a rescan reintroduced only on the receiver-bucket-absent path — the most plausible way #2788 returns — measures **4.538 and fails**, while an always-hits corpus scored the same patch 1.279 and passed.

## Correcting a claim from the first revision of this PR

The original description and code comments said #1990 "shipped without a scaling gate". **That was wrong**, and the review caught it. #1990 shipped `test/integration/cpp-adl-benchmark.test.ts` (`f1b843838`), which asserts `emitRatio < fileRatio^1.5`.

The accurate point is narrower and, as it turns out, stronger: that bench asserts `callsResolved === 0`, so it only generates *unresolved* ADL sites and never drives `resolveCppQualifiedNamespaceMember`; and it is `describe.skipIf(!GITNEXUS_BENCH)` while the only CI step that sets `GITNEXUS_BENCH=1` enumerates its files explicitly and lists neither C++ bench. So it has never executed in CI. Wiring both #1990 benches into that file list is a worthwhile follow-up, out of scope here.

## Review

This PR was reviewed by a multi-engine pass (Claude GitNexus swarm + `ce-code-review` personas; Codex `gpt-5.6-sol`/xhigh swarm + ce + adversarial) which returned two P1s and five smaller findings — all fixed in `0015b0d64`:

- **P1** — the index pinned the whole C++ `ParsedFile` set past `phase.ts`'s deliberate post-language `forceGc()`, whose comment sizes that set at ~17-20GB on the Linux kernel. Measured 61.2MB retained before, 0.1MB after the WeakMap change. The ADL twin shares the pattern but arms far more rarely; moving it is a follow-up.
- **P1** — the bench could not see the receiver-miss path, as described above.
- Unbounded recursion threw an uncontained `RangeError` at inline depth 8000, aborting analyze; a receiver *miss* paid full recursion where the old walker skipped on a name mismatch. A work-stack alone would only have converted this into an OOM at depth 6000, since the eager table was quadratic in memory too — hence the lazy node graph. Depth 100,000 now costs 133ms.
- Two mutation-verified test gaps closed: cross-file merge of a namespace reopened in two files, and the same-name inline nest whose dedup, when defeated, flips a resolved def to `'ambiguous'`.
- Comment corrections: the #1990 framing above, the ordering claim, and the overstated `ensureAdlIndex` parity (no invariant guard is warranted here — the sibling's exists to protect a `?? 0` defaulting read that this index has no analogue for).
- Memoized candidate buckets are frozen, so a future in-place sort in `overload-narrowing.ts` throws rather than silently corrupting later resolutions now that the array is shared.
- `_scaling_note`'s documented band did not reproduce; it now states the observed 1.28-1.45, the small arm was widened to ~14ms (halving the spread), and a triage line notes a scaling failure is a timing signal to re-run — unlike the deterministic fingerprint arm.


magyargergo (commented):
## Multi-engine review — PR #2794 (#2788, C++ qualified `ns::member()` index)

Two engines, twelve lanes, against head `48afb7480` / base `911151e23`.

**Engine A — Claude.** GitNexus swarm (facts-historian, branch-hygiene, risk-architect, test-CI, security, docs-DoD) plus six `ce-code-review` personas (correctness, adversarial, performance, testing, maintainability, project-standards). All twelve live; all ran code in the worktree.

**Engine B — Codex (gpt-5.6-sol / xhigh).** Three legs, all live but unequal:
- `codex-swarm` (Solo mode) — produced F1/F2/F3, drove GitNexus MCP (`detect_changes`, `impact` ×4, `context`) and read exact-head Actions results. **Could not run Node locally** (bubblewrap namespace failure), so its findings are code-read + CI-derived.
- `codex-ce` (report-only) — **zero actionable findings, verdict "ready to merge."** Its value here is validation and one independent history check, not findings. Its cross-model pass did not run (read-only, network-restricted host).
- `codex-adversarial` — two findings (`[high]` retention, `[medium]` bench density). **Node was unavailable in its sandbox** (`node --version` exited 1; three bench invocations failed), so both are `[code-read]`.

Net: Codex contributed 5 findings across 2 of 3 legs. Where a Codex finding and a Claude finding coincide, it is called out below — that is the signal worth weighting.

---

### Verdict

**not production-ready** — two P1 findings, both with reproduced triggers, and both fixable without redesigning the change. The resolver rewrite itself is correct and unusually well validated (see below); what needs work is the memory lifecycle around it and the gate meant to protect it.

- **Branch hygiene:** clean feature/fix PR. Single commit on current `main`, five files, no unrelated churn, no merge-from-main noise, no schema or `INCREMENTAL` constant touched, no binary-classified files at head.
- **Merge state:** `mergeable`. No conflicts with `main`; no overlap with the other open PR from this area (#2793). The one red check is Vercel, which cannot pass on a fork PR — non-actionable.
- **Recommended disposition:** fix the two P1s (or explicitly accept the retention one as pre-existing-and-widened, with an issue filed for the ADL twin), correct the #1990 wording in its four locations, then merge. Nothing here requires reverting the approach.

### Review bar for this PR

Derived from `DoD.md` against the domains this diff touches (C++ resolution engine, module-level pipeline state, benchmark harness, CI workflow):

- **2.1** fix wired into the real runtime path, not a test-only shim — **met** (both production call sites verified).
- **2.3** no stale comments — **partially met**: the PR corrects one pre-existing false comment, and introduces three new inaccurate ones (#1990 framing, "ordering is load-bearing", "same contract as `ensureAdlIndex`").
- **2.4** no contract change to `gitnexus-shared/` types, CLI flags, MCP surface, or graph node/edge shapes — **met**.
- **2.6** eliminates repeated avoidable work — **met**, and that is the point of the PR.
- **2.7** tests cover the changed behaviour — **partially met**: strong on semantics and cache invalidation, absent on cross-file merge and the dedup arm.
- **2.9** residual risk stated — **not met** for the retention lifecycle and the unbounded recursion, neither of which is acknowledged anywhere in the PR.
- **3** evidence-based edits — **not met** for the #1990 claim, which is asserted without checking and is false.
- **4.2** `tsc --noEmit` + tests green — **met**.
- **4.5** CI wiring correct; CHANGELOG owned by the release process — **met** (step correctly placed and always-on; CHANGELOG correctly untouched).

---

### What's solid

The resolver rewrite itself survived everything thrown at it. This is the most heavily validated part of the review and it should be credited:

- **Two independently constructed differential fuzzes, zero divergences.** `ce-correctness` transcribed the deleted algorithm from base and ran 120,000 `(workspace, receiver, member)` probes over 4,000 random workspaces, then 180,000 more with `parameterCount`/`parameterTypes` metadata and 8 call-site shapes so `narrowOverloadCandidates` + `cppConversionRank` were actually engaged — **0 mismatches**. Separately it ran a **candidate-order** differential: 600,000 element-by-element `nodeId` comparisons over 20,000 workspaces — **0 mismatches**. `ce-adversarial` independently reconstructed the walker and ran 80,000 cases — **0 divergences**. Generators covered colliding `nodeId`s across files, `qualifiedName` `undefined`/`''`/trailing-dot, non-`Namespace` kinds, `''` receivers and members.
- **The committed fingerprint encodes pre-change behaviour, not self-certified post-change behaviour.** Two lanes recomputed the bench corpus fingerprint with the **old** algorithm and got byte-identical `30122b08…02e1e2` at LARGE. `ce-testing` went further and measured the legacy scan's own ratio at the real 400/1600 scale: **3.871** (small 1313.9 ms, large 20343.7 ms) — which empirically confirms the baseline's `3.45` figure that `facts-historian` had correctly flagged as a cross-scale extrapolation. The 1.8 budget really does sit between the two bands at the scale the bench runs.
- **The mutation test bites.** `swarm-test-ci` commented out `inline-namespaces.ts:95-96` and the `rebuilds after clearCppInlineNamespaces` test failed exactly, and only it; tree reverted clean.
- **Cache-staleness attacks all failed, cross-engine.** Side-channel-after-build, `populateWorkspaceOwners` racing the build, `parsedFiles` identity thrash, `GITNEXUS_DISK_SCOPE_INDEX` array swap, interleaved passes, and ADL-vs-qualified index disagreement were each refuted independently by `risk-architect`, `ce-correctness` (H6), `ce-adversarial` (#3/#4/#5/#6/#8) **and** `codex-swarm` §7.
- **`allHits` aliasing refuted by reading `overload-narrowing.ts` end to end** — every derivation is `.filter`/`.map`; no `.sort(`/`.reverse(`/`.splice(` in the file.
- **Security lane: clean.** No `${{ }}` in the new step, no new `uses:`, `pull_request` (not `pull_request_target`), `measure.mjs` touches only `node:fs/path/crypto/url`, no NUL or bidi bytes at head — the earlier binary-file defect is confirmed gone.
- Hygiene: single commit on current main, no schema/`INCREMENTAL` touch, prettier clean, `tsc --noEmit` clean, new unit suite 7/7 (verified by both engines), CHANGELOG correctly untouched.

---

### Headline findings

#### P1 — The module-global index pins the entire C++ `ParsedFile` set past the pipeline's deliberate post-language `forceGc()` `[reproduced]`
`gitnexus/src/core/ingestion/languages/cpp/inline-namespaces.ts:147-148`

**Trigger:** any qualified `ns::member()` call site anywhere in the workspace builds the index. `qualifiedNsIndexSource` is nulled only by `clearCppInlineNamespaces()` (`:95-96`), reached only from `cppScopeResolver.loadResolutionConfig` at the start of the **next** C++ pass — which in a single `analyze` never comes.

**Evidence:** `scope-resolution/pipeline/phase.ts:505-521` evicts `files`/`contents`/`preExtractedByPath` and then states its premise in a comment — *"This language's ParsedFiles are now unreachable … Force a GC HERE so a heavy language's ~17-20GB set (e.g. C/C++ on the Linux kernel) is reclaimed"* — before calling `forceGc()`. This PR falsifies that premise: module state strongly holds the `parsedFiles` array and `byReceiver` holds `SymbolDefinition` references reaching into those files' scopes. `ce-adversarial` reproduced it with `--expose-gc` at a pass boundary: **104.2 MB still reachable after the post-pass `forceGc()`, 6.5 MB after `clearCppInlineNamespaces()`**. C++ is 7th of 16 in `SCOPE_RESOLVERS`, so the retention spans nine later language passes plus the whole emit/persistence phase.

**Cross-engine:** `codex-adversarial` filed this `[high]` and was the one to tie it to the explicit `forceGc()` contract; `claude-ce-adversarial` reproduced the retention; `claude-ce-performance` (F2) and `claude-ce-correctness` (R4) reached the same place from different angles.

**Adjudicating the disagreement.** `claude-swarm-risk-architect` called this "not a new hazard, identical to `adlIndex`". Verified against source: the *pattern* is indeed pre-existing (`adl.ts:179-180` is the same `let index` / `let indexSource` shape cleared from the same `loadResolutionConfig`) — but the *trigger breadth* is new. `pickCppAdlCandidates` returns early at `adl.ts:570-572` on `noAdlSites` / empty `argInfoBySite` **before** `ensureAdlIndex` ever runs; the new index fires on any qualified `ns::member()` call, which is pervasive in exactly the ABI-versioned C++ corpora #2788 is about. A latent retainer that rarely armed now arms almost always. risk-architect is right about the pattern and wrong about the consequence.

**Fix:** `WeakMap<readonly ParsedFile[], QualifiedNsMemberIndex>`, the pattern already in this directory at `cpp/file-local-linkage.ts:286-290` (`moduleScopeIndexByPass`), with `clearCppInlineNamespaces` deleting the entry. Note `ce-maintainability`'s valid caveat: because this index is a function of **two** inputs (`parsedFiles` *and* the module-level `inlineNamespaceScopeIds`), a WeakMap does not remove the explicit clear — but it does remove the pin, which is the point. The ADL twin should probably move with it.

#### P1 — The new bench's scaling arm is blind to the receiver-miss path, which is where #2788 actually lives `[reproduced]`
`gitnexus/bench/cpp-qualified-ns/measure.mjs:138-146`

**Trigger:** `callSites()` emits `ns_${(f * 7 + c * 13) % fileCount}` — always a namespace the corpus declared — so `byReceiver.get(receiverName)` **never misses**. Every miss the bench measures is a *member* miss inside an already-resolved receiver bucket. Production is the opposite: Case 1.5 at `receiver-bound-calls.ts:871-884` is reached by every plain-identifier receiver call, and the overwhelming majority are receiver misses.

**Reproduced:** `ce-adversarial` patched in the most plausible future regression — a "defensive" rescan when the receiver bucket is absent — and measured it on a production-shaped mix (80% receiver misses): **54.0 ms → 927.5 ms at 100→400 files, ratio 4.293, 507× slower than HEAD.** On the *committed* corpus the same defective build scores **1.332 against the 1.8 budget and prints `PASS`.** Three further behaviour-changing patches (dedup removed, per-file `seen.clear()`, inline children collected before own defs) also passed `--check` with a **byte-identical fingerprint**. The dedup removal is not theoretical: on `namespace ns { inline namespace ns { void f(); } }` it flipped `n:f` to `AMBIGUOUS`, dropping a CALLS edge.

This is precisely the failure the PR body indicts #1990 for — a gate that exists but does not cover the path where the bug lives.

**Corroboration, different axis, cross-engine:** `codex-swarm` F1 and `codex-adversarial` `[medium]` both independently found that the corpus holds namespace density (4 scopes/file) and inline depth (1) constant across both scales, so a Θ(N²) density/depth regression stays green. `claude-swarm-risk-architect` (Finding 1) adds that the corpus has no same-name multi-overload case, so ordering regressions cannot register as fingerprint drift either. Three lanes across both engines agree the gate under-covers; the receiver-miss instance is the one that was reproduced end to end.

**Fix:** make `callSites()` mirror the production receiver distribution (keep ~1 in 5 as real namespaces, rest non-namespace identifiers), and extend `buildCorpus` with a reopened cross-file namespace, an `ns::ns` same-name nest, a member shadowed between namespace and inline child, and a batch of sites passing a real `callsite`. Each addition moves the fingerprint once, deliberately.

#### P2 — Index build is O(depth²) in inline nesting and recurses without a bound; a pure receiver *miss* now pays for it `[reproduced]`
`gitnexus/src/core/ingestion/languages/cpp/inline-namespaces.ts:193` (build loop) and `:222-224` (recursion)

**Trigger:** `buildQualifiedNsMemberIndex` calls `collectNamespaceMembers` for **every** `Namespace` scope, not just roots. In a chain `outer { inline v1 { inline v2 { … } } }` each of `outer`, `v1`, `v2` … is itself a legal receiver, so each is visited and re-descends its remaining descendants: D + (D−1) + … + 1.

**Reproduced (two lanes, two symptoms).** `ce-performance` measured the build over a depth sweep — 0.284 ms @ 50, 7.789 ms @ 400, 114.2 ms @ 1600 — giving exponents **1.94** (400→1600) and **1.86** (200→1600). `ce-adversarial` reproduced the harder edge: resolving a receiver that **does not exist** costs 5.1 ms at depth 100, 428.3 ms at depth 5000, and **throws `RangeError: Maximum call stack size exceeded` at depth 10000**. That throw is uncontained — `run.ts`'s only `try/catch` wraps the CFG/PDG emit block, and `phase.ts:324`'s `try` has a `finally` with no `catch` — so it aborts the whole analyze, the same failure class as #2769.

**This is a genuine delta of the PR, not inherited:** the deleted walker recursed only for namespaces whose simple name matched the query (`if (nsName !== receiverName) continue;`), so a miss cost zero recursion. `claude-swarm-security` (Finding 6) reached the unbounded-recursion half independently.

**Severity adjudication.** `codex-swarm` F1 rates this P2 and **merge-blocking**; `ce-performance` rates it P3 / confidence 40 because real inline nesting is depth 1–2 (`std::__1`, `std::__cxx11`). P2, not blocking, is the honest landing: no real corpus reaches depth 100, but code-generated C++ can, the widened trigger is new, and one `depth` counter (or visited-`ScopeId` set) with a `logger.warn` past a generous ceiling closes both symptoms in a few lines.

#### P2 — "#1990 shipped without a scaling gate" is false as written, in two permanent artefacts `[code-read]`
`.github/workflows/ci-tests.yml:509` and `gitnexus/bench/cpp-qualified-ns/measure.mjs:10-12` (also the commit message and PR body)

`test/integration/cpp-adl-benchmark.test.ts` exists on HEAD, was added by #1990, self-describes as *"Guards the optimization in PR #1990"*, and asserts `emitRatio < fileRatio^1.5`. #1990 also shipped `cpp-pipeline-benchmark.test.ts`. Six lanes across both engines flagged the claim.

**But the decisive fact is one most lanes missed, and it cuts the other way.** That bench is `describe.skipIf(!BENCH_ENABLED)` (`:140`, `BENCH_ENABLED = process.env.GITNEXUS_BENCH === '1'`), and the one CI step that sets `GITNEXUS_BENCH=1` — "Cross-language pipeline benchmarks (GITNEXUS_BENCH, serial)" — enumerates its files explicitly: cobol, csharp, instance-ownership, spring-bean-resource, rust, php, ruby. **Neither C++ bench appears.** `git log -S "cpp-adl-benchmark" -- .github/workflows/` returns zero commits. The #1990 guard has never executed in CI.

This half was found independently by `claude-ce-project-standards`, `codex-swarm` (§5, explicitly correcting the prior-context lead) and `codex-ce` (coverage note) — **the most strongly corroborated fact in this review, and it is cross-engine.** It also means `swarm-facts-historian` and `swarm-test-ci`, which treated the ADL bench as live coverage, were wrong on that point.

**Severity landed at P2** (docs-DoD proposed P1/blocking; `codex-swarm` F2 and `ce-correctness` F2 proposed P3). It is a factual error committed into permanent repo content that inverts the lesson for the next reader, but it changes no runtime behaviour. Fix before merge; do not block on it.

**Accurate replacement, both places:** *"#1990 shipped a scaling gate for the ADL path (`test/integration/cpp-adl-benchmark.test.ts`), but it asserts `callsResolved === 0` — only unresolved ADL sites — so it never drives `resolveCppQualifiedNamespaceMember`; and it is `skipIf(!GITNEXUS_BENCH)` and absent from that job's vitest file list, so it has never run in CI. This step is an always-on gate for the qualified receiver path."* Wiring the two #1990 benches into that file list is the follow-up this discovery actually points to — out of scope here, worth an issue.

#### P3 — "Ordering is load-bearing" is not a live property, and the test named for it does not test order `[code-read]`
`gitnexus/src/core/ingestion/languages/cpp/inline-namespaces.ts:137-139`, `gitnexus/test/unit/cpp-qualified-ns-index.test.ts:200`

**Both sides of this cross-lane split are right, and the distinction matters.** `risk-architect` and `ce-correctness` spent real effort proving order *equivalence* with the legacy walker holds — and it does: 600,000 element-by-element comparisons, zero divergences. That work stands. What is unsupported is the stated *justification*. `allHits[0]` is only reached under `if (allHits.length === 1)` (`:270`) — trivially order-independent. `narrowOverloadCandidates` is not first-wins: with `callsite === undefined` (the `resolveAdlCandidates` path) the arity and type filters are no-ops, and **both** branches of the final `if` at `:290-293` return `'ambiguous'`. The `keeps scan order` test asserts membership, not order — `outer` owns only `foo`, `v1` only `bar`, so both lookups are single-hit and pass under any traversal; `ce-adversarial`'s reordered build passes it, passes the bench, and passes every `cpp-inline-namespace-*` fixture.

Cross-engine: `claude-ce-testing`, `codex-swarm` F3, `codex-ce` (testing gap) and `claude-ce-adversarial` all landed here. Preserving order is the right conservative choice; the comment should say *"preserved for byte-identity, not currently observable through this resolver's return contract"* and the test should be renamed to what it checks.

---

### Lower-priority items

- **P2 — test coverage gaps with demonstrated consequences** (`test/unit/cpp-qualified-ns-index.test.ts:62-79`). `makeParsedFiles` always `return [parsed]`, so no test puts one receiver's members in **more than one file** — the exact property the shared `byReceiver`/`seenByReceiver` pair spanning the `for (const parsed of parsedFiles)` loop exists to serve. A future edit re-scoping either inside the file loop would break cross-file namespaces and every test would still pass. Nor does anything cover the `seen.has(dedupKey)` `continue` arm: `ce-adversarial` showed removing the dedup flips `ns::ns` `f()` from resolved to `'ambiguous'`, and no fixture uses same-name inline nesting (all seven use disjoint `v1`/`v2`/`detail` names). Three Claude lanes agree (`ce-testing` P1, `swarm-test-ci`, `ce-adversarial`); Codex did not raise it, so treat as single-engine.
- **P3 — `_scaling_note`'s documented band does not reproduce** (`bench/cpp-qualified-ns/baselines.json:5`). It claims "measured 0.93-1.21". Observed here: 1.012–1.475 unloaded across five lanes, and `ce-correctness` measured **2.102 with four pinned busy cores — which would fail the 1.8 gate** with the message "per-call-site cost now grows with corpus size again (#2788)", actively misleading when the cause is runner contention. **Not a defect this PR introduces:** `ce-project-standards` ran the precedent `bench/callable-value-flow` as a control (documented 1.14–1.16) and got 1.153–1.294 — equally out of band. Both time a ~1.8 ms small arm. Widen the note to the observed band and add one line saying a *scaling* failure means re-run first; **never** touch the fingerprint.
- **P3 — "Same contract as `ensureAdlIndex`" (`:238`) overstates parity.** The sibling ships a dev-gated `validateAdlSeqCoverage`; this index ships none. **Resolving the cross-lane split:** `ce-maintainability` wins on evidence over `risk-architect`. The ADL guard exists because `pickCppAdlCandidates` has an `idx.seqByNodeId.get(def.nodeId) ?? 0` fallback that can silently collapse two candidates onto slot 0 — a real silent-collision footgun. `buildQualifiedNsMemberIndex` has no analogous fallback; a dedup break would just double-count into an existing array, which the ambiguity assertions catch. **No guard warranted — narrow the comment instead** ("same staleness contract, without its `validateAdlSeqCoverage`-style runtime guard") and state the omission as accepted risk.
- **P3 — second call site undocumented.** `resolveAdlCandidates` (`cpp/scope-resolver.ts:366`) also calls `resolveCppQualifiedNamespaceMember`, with **no** `callsite`, to resolve `using ns::name;`. The PR text names only `resolveQualifiedReceiverMember`. Behaviour is correct (same `parsedFiles` reference, whichever fires first builds), but the JSDoc should list both. That path also has zero unit coverage in the new file.
- **Residual, not filed:** `narrowOverloadCandidates` can return the cached bucket *by reference* when no callsite is supplied. Nothing mutates it today, but the array is now shared across every call site in the pass instead of freshly built per call, so an in-place `.sort()`/`.splice()` added later to `overload-narrowing.ts` would corrupt subsequent resolutions rather than one call. `Object.freeze` per bucket, or `allHits.slice()` when `length > 1`, closes it for a line.
- **Residual, pre-existing:** under `GITNEXUS_DISK_SCOPE_INDEX`, `emitParsedFiles` carries `scopes: []`, so qualified `ns::member()` resolution returns `undefined` for every site — old and new identical, but the new fingerprint gate cannot see that mode either. Also: `gitnexus/tsconfig.json` includes only `src/**/*`, so the new test file's `as unknown as` casts are never typechecked against the real `ParsedFile`/`SymbolDefinition` shapes.

### Refuted suspicions (dropped, and worth recording)

Semantic divergence from the deleted walker (three fuzzes, 980,000 total probes/comparisons, zero mismatches) · candidate-order divergence · dedup-semantics change (old `seenNodeId` and new `member \0 nodeId` partition identically) · `allHits` shared-mutable-array corruption · `parsedFiles` identity thrash (one build per pass, both entry points trace to the single `emitParsedFiles` binding) · side-channel marks arriving after the build · `populateWorkspaceOwners` triggering an early build · `GITNEXUS_DISK_SCOPE_INDEX` array-swap staleness · concurrent/interleaved pass corruption · ADL-vs-qualified index disagreement about the inline set · duplicate `ScopeId` divergence (a real algebraic difference, but unreachable for C++ `Namespace` scopes — one capture pattern, `makeScopeId` keyed on range+kind) · bench JIT warm-up ordering bias (reversed order: no material change) · workflow template injection, unpinned actions, `pull_request_target`, network/filesystem access in `measure.mjs`, unsafe `baselines.json` deserialization · raw NUL / bidi / control bytes at head · bench duplication with `cpp-adl-benchmark.test.ts` (different code paths) · the reporter's WeakMap suggestion being silently ignored (documented, and WeakMap alone would not remove the clear-site coupling).

### Test gaps

Multi-file cross-receiver merge; same-name inline nesting (`ns::ns`); a namespace reopened across two files; a member declared at both namespace level and in an inline child; any call passing a real `callsite` (the bench passes four arguments, so `narrowOverloadCandidates`/`cppConversionRank`/`isOverloadAmbiguousAfterNormalization` sit outside both the timing and fingerprint arms); 3+ chained inline levels; the "one build per pass" property (nothing asserts the positive); a `WeakRef` assertion that `parsedFiles` is collectable after `runScopeResolution` returns — which would have caught the P1 above and would protect the ADL twin at the same time. `test/integration/resolvers/cpp.test.ts` could not run in either engine's worktree (symlinked `node_modules`, no built `dist/parse-worker.js`; 9 environment failures) — environmental, not a PR defect, but it means the #1564 ambiguity reason, STL `__1` nesting, and the `using ns::name;` recovery path were verified only at unit/differential level.

### CI

41 pass, 1 skipping, 1 fail. The sole failure is **Vercel — "Authorization required to deploy"**, the known non-actionable fork-PR infra failure, not a code defect. `CI Gate` passed. Critically, `tests / benchmarks (GITNEXUS_BENCH)` — the job containing this PR's own new gate — **passed in 8m26s**, with the new step reporting scaling ratio `1.093` against the 1.8 budget and fingerprint `30122b08…` matching `baselines.json`. This resolves `swarm-facts-historian`'s "do not treat the PR as CI-green" caveat, which was written while that job was still in progress.

---

*Automated multi-tool digest (Claude swarm + `ce-code-review`; Codex gpt-5.6-sol/xhigh swarm + ce + adversarial). Verify each finding against the source before acting on it.*

gitnexus/src/core/ingestion/languages/cpp/inline-namespaces.ts
```diff
@@ -115,11 +120,136 @@ export function isCppInlineNamespaceScope(scopeId: ScopeId): boolean {
...
+ * load-bearing: the caller returns `allHits[0]` for the single-hit case and
+ * `narrowOverloadCandidates` is first-wins.
+ */
+interface QualifiedNsMemberIndex {
+  readonly byReceiver: ReadonlyMap<string, ReadonlyMap<string, readonly SymbolDefinition[]>>;
+}
+
+type NsScope = ParsedFile['scopes'][number];
+
+let qualifiedNsIndex: QualifiedNsMemberIndex | undefined;
```
magyargergo: **P1 `[reproduced]` — this pins the whole C++ `ParsedFile` set past the pipeline's deliberate post-language `forceGc()`.**

`qualifiedNsIndexSource` strongly holds the `parsedFiles` array, and `byReceiver` holds `SymbolDefinition` references reaching into those files' scopes. Both are nulled only by `clearCppInlineNamespaces()` (:95-96), reached only from `cppScopeResolver.loadResolutionConfig` at the start of the *next* C++ pass — which in a single `analyze` never comes.

`scope-resolution/pipeline/phase.ts:505-521` evicts `files`/`contents`/`preExtractedByPath` and then states its premise in a comment — *"This language's ParsedFiles are now unreachable … Force a GC HERE so a heavy language's ~17-20GB set (e.g. C/C++ on the Linux kernel) is reclaimed"* — before calling `forceGc()`. This PR falsifies that premise. Reproduced with `--expose-gc` at a simulated pass boundary: **104.2 MB still reachable after the post-pass `forceGc()`, 6.5 MB after `clearCppInlineNamespaces()`**. C++ is 7th of 16 in `SCOPE_RESOLVERS`, so the retention survives nine later language passes plus the entire emit/persistence phase.

**On "this is the same as `adlIndex`":** the *pattern* is pre-existing (`adl.ts:179-180`), but the *trigger* is not. `pickCppAdlCandidates` returns early at `adl.ts:570-572` on `noAdlSites` / empty `argInfoBySite` **before** `ensureAdlIndex` runs; this index arms on any qualified `ns::member()` call — pervasive in exactly the ABI-versioned C++ corpora #2788 is about. A retainer that rarely armed now arms almost always.

**Suggested fix:** `WeakMap<readonly ParsedFile[], QualifiedNsMemberIndex>`, the pattern already one directory over at `cpp/file-local-linkage.ts:286-290` (`moduleScopeIndexByPass`), with `clearCppInlineNamespaces` deleting the entry. Because this index depends on two inputs (`parsedFiles` *and* `inlineNamespaceScopeIds`) the WeakMap does not remove the explicit clear — but it removes the pin, which is the point. Worth moving the ADL twin with it.

Found by the Codex adversarial leg (`[high]`, which identified the `forceGc()` contract) and reproduced by the Claude adversarial lane.

gitnexus/bench/cpp-qualified-ns/measure.mjs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+/** The call sites: a deterministic spread over receivers and member names so
+ *  each rep resolves the same set, with hits, misses and ambiguities mixed. */
+const MEMBERS = ['own0', 'inl0', 'dup', 'hidden0', 'nosuch'];
+function callSites(fileCount) {
+  const sites = [];
+  for (let f = 0; f < fileCount; f++) {
+    for (let c = 0; c < CALLS_PER_FILE; c++) {
+      sites.push([`ns_${(f * 7 + c * 13) % fileCount}`, MEMBERS[c % MEMBERS.length]]);
```
magyargergo: **P1 `[reproduced]` — the scaling arm cannot see the regression class it exists to gate.**

`ns_${(f * 7 + c * 13) % fileCount}` is always a namespace the corpus declared, so `byReceiver.get(receiverName)` **never misses**. Every miss this bench measures is a *member* miss inside an already-resolved receiver bucket (`hidden0`, `nosuch`). Production is the opposite: Case 1.5 at `receiver-bound-calls.ts:871-884` is reached by every plain-identifier receiver call, and the overwhelming majority are receiver misses.

The most plausible future regression was patched in — a "defensive" rescan when the receiver bucket is absent ("maybe the namespace came from a header parsed later") — and measured:

| corpus | impl | small | large | ratio | `--check` |
|---|---|---|---|---|---|
| committed (100% receiver hits) | HEAD | 1.72 ms | 8.71 ms | 1.268 | PASS |
| committed | rescan-on-receiver-miss | 1.73 ms | 9.19 ms | **1.332** | **PASS** |
| production mix (80% misses) | HEAD | 0.81 ms | 1.83 ms | 0.563 | — |
| production mix | rescan-on-receiver-miss | **54.0 ms** | **927.5 ms** | **4.293** | — |

The defective build is 507× slower at 400 files with an unmistakable quadratic signature, and the committed bench scores it 1.332 against the 1.8 budget and prints `PASS`. Three further behaviour-changing patches (dedup removed, per-file `seen.clear()`, inline children before own defs) also passed with a **byte-identical fingerprint** — and the dedup removal is not theoretical: on `namespace ns { inline namespace ns { void f(); } }` it flipped `n:f` to `AMBIGUOUS`, dropping a CALLS edge.

This is exactly what the PR indicts #1990 for: a gate that exists but does not cover the path where the bug lives.

**Suggested fix:** mirror the production receiver distribution here (keep ~1 in 5 as declared namespaces, make the rest non-namespace identifiers like `obj0`/`Widget3`), and extend `buildCorpus` with a namespace reopened across files, an `ns::ns` same-name nest, a member shadowed between namespace and inline child, and a batch of sites passing a real `callsite`. Each addition moves the fingerprint once, deliberately.

Related, from the other engine: the Codex swarm (F1) and Codex adversarial legs both flagged that both scales hold namespace density (4 scopes/file) and inline depth (1) constant, so a Θ(N²) density/depth regression also stays green.

gitnexus/src/core/ingestion/languages/cpp/inline-namespaces.ts
```diff
@@ -115,11 +120,136 @@ export function isCppInlineNamespaceScope(scopeId: ScopeId): boolean {
...
+      if (byMember === undefined) {
+        byMember = new Map();
+        byReceiver.set(nsName, byMember);
+      }
+      let seen = seenByReceiver.get(nsName);
+      if (seen === undefined) {
+        seen = new Set();
+        seenByReceiver.set(nsName, seen);
+      }
+      collectNamespaceMembers(scope, inlineChildrenByParent, byMember, seen);
```
magyargergo: **P2 `[reproduced]` — O(depth²) build, unbounded recursion, and a pure receiver *miss* now pays for both.**

This calls `collectNamespaceMembers` for **every** `Namespace` scope, not just roots. In a chain `outer { inline v1 { inline v2 { … } } }` each of `outer`, `v1`, `v2` … is itself a legal receiver (`v2::foo()` is valid C++), so each is visited by this loop and re-descends its remaining descendants via :222-224 — D + (D−1) + … + 1.

Measured build time over a depth sweep: 0.284 ms @ 50, 2.404 ms @ 200, 7.789 ms @ 400, 27.463 ms @ 800, 114.2 ms @ 1600 — exponent **1.94** (400→1600) and **1.86** (200→1600), the O(D²) signature.

Harder edge, separately reproduced: resolving a receiver that **does not exist** costs 5.1 ms at depth 100, 428.3 ms at depth 5000, and throws `RangeError: Maximum call stack size exceeded` at depth 10000. That throw is uncontained — `run.ts`'s only `try/catch` wraps the CFG/PDG emit block, and `phase.ts:324`'s `try` has a `finally` with no `catch` — so it aborts the whole analyze (same failure class as #2769).

**This part is a genuine delta of the PR:** the deleted walker recursed only for namespaces whose simple name matched the query (`if (nsName !== receiverName) continue;`), so a miss cost zero recursion. The new build reaches every namespace regardless of what is queried.

**Suggested fix:** thread a `depth` counter (or a visited-`ScopeId` set) into `collectNamespaceMembers` and bail with a `logger.warn` past a generous ceiling — that closes both the stack overflow and the pathological build in a few lines. If the O(D²) itself is worth removing, process inline scopes bottom-up and memoize each scope's transitive member map.

Cross-engine: Codex swarm (F1, rated merge-blocking), Codex adversarial (`[medium]`), Claude performance (reproduced the exponent) and Claude security (unbounded-recursion half). Landing at P2 rather than blocking because real inline nesting is depth 1–2 (`std::__1`, `std::__cxx11`) — but code-generated C++ can exceed that, and the fix is cheap.

.github/workflows/ci-tests.yml
```diff
@@ -500,6 +500,17 @@ jobs:
         run: node --import tsx bench/callable-value-flow/measure.mjs --check
         working-directory: gitnexus
 
+      - name: C++ qualified-namespace resolution guards (#2788)
+        # Build-free: asserts resolveCppQualifiedNamespaceMember resolves an
+        # unchanged symbol set (fingerprint) and that per-call-site cost stays
+        # independent of corpus size. It rescanned every parsed file per
+        # qualified `ns::member()` call site until #2788 — 25 of 33 analyze
+        # minutes on a 1.5k-file C++ repo. #1990 had already fixed the exact
+        # same bug in the sibling ADL path and shipped without a scaling gate,
```
magyargergo: **P2 `[code-read]` — this claim is false, and it is being committed into a permanent artefact.** The same text is at `gitnexus/bench/cpp-qualified-ns/measure.mjs:10-12`, in the commit message, and in the PR body — a correction needs all four.

`gitnexus/test/integration/cpp-adl-benchmark.test.ts` exists on HEAD, was added by #1990 (`18b29b6e4`), self-describes as *"Guards the optimization in PR #1990"*, and asserts `emitRatio < fileRatio^1.5`. #1990 also shipped `cpp-pipeline-benchmark.test.ts` (`4f5ca77ff`). So #1990 did ship a scaling gate.

**But the fuller fact supports this PR better than the current wording does.** That bench is `describe.skipIf(!BENCH_ENABLED)` (:140, `BENCH_ENABLED = process.env.GITNEXUS_BENCH === '1'`), and the one CI step that sets `GITNEXUS_BENCH=1` — "Cross-language pipeline benchmarks (GITNEXUS_BENCH, serial)" — enumerates its files explicitly: cobol, csharp, instance-ownership, spring-bean-resource, rust, php, ruby. **Neither C++ bench is listed.** `git log -S "cpp-adl-benchmark" -- .github/workflows/` returns zero commits. The #1990 guard has never executed in CI.

**Suggested replacement, here and in `measure.mjs`:** *"#1990 shipped a scaling gate for the ADL path (`test/integration/cpp-adl-benchmark.test.ts`), but it asserts `callsResolved === 0` — only unresolved ADL sites — so it never drives `resolveCppQualifiedNamespaceMember`; and it is `skipIf(!GITNEXUS_BENCH)` and absent from that job's vitest file list, so it has never run in CI. This step is an always-on gate for the qualified receiver path."*

As written, the comment steers the next contributor away from the existing-but-unwired guard rather than toward wiring it. Adding both #1990 benches to that file list is the follow-up this points to — out of scope here, worth an issue.

Found independently by both engines (Claude: docs-DoD, maintainability, project-standards, correctness, facts-historian; Codex: swarm F2 and ce). The "never actually runs in CI" half was found by Claude project-standards, Codex swarm and Codex ce independently — the most strongly corroborated fact in this review.

gitnexus/test/unit/cpp-qualified-ns-index.test.ts
```diff
@@ -0,0 +1,258 @@
...
+    scopes: specs.map((s) => ({
+      id: s.id as unknown as ScopeId,
+      kind: s.kind,
+      parent: s.parent as unknown as ScopeId | null,
+      ownedDefs: s.defs,
+      range: range(s.line),
+    })),
+  } as unknown as ParsedFile;
+  markInline(parsed, specs, inlineIds);
+  return [parsed];
```
magyargergo: **P2 `[reproduced]` — `makeParsedFiles` always returns a one-element array, so the property this refactor is built around is untested.**

`buildQualifiedNsMemberIndex` threads a single `byReceiver` map and a single `seenByReceiver` pair through the `for (const parsed of parsedFiles)` loop — cross-file accumulation is the central structural change versus the legacy code, which re-derived `allHits` fresh per call. No test in this file puts one receiver's members in more than one file. The `'does not serve one parsedFiles array's index to another'` test (:217) uses two **separate** single-element arrays, which tests cache-key invalidation, not merging.

The integration suite was checked too: all seven `cpp-inline-namespace-*` fixtures declare the namespace-under-test entirely in one header, and every other C++ fixture that reopens a namespace across files reaches it through an unqualified ADL call or a variable reference, not a qualified `ns::member()` call. So a future edit that re-scopes `byReceiver`/`seenByReceiver` inside the file loop — each file clobbering the previous instead of merging — would change behaviour for any namespace split across files, and every existing test would still pass.

**Second gap in the same file:** nothing covers the `seen.has(dedupKey)` `continue` arm at `inline-namespaces.ts:213`. Its only trigger is same-name inline nesting (`namespace ns { inline namespace ns { … } }`), which no unit test and no fixture uses — all use disjoint `v1`/`v2`/`detail` names. This was demonstrated to be a real verdict flip, not a theoretical one: with the dedup removed, `ns::f()` resolves `n:f` → `AMBIGUOUS` and the CALLS edge is dropped.

**Suggested fix:** two small tests — (a) a two-element `parsedFiles` array where file A declares `namespace outer { void a(); }` and file B declares `namespace outer { void b(); }`, asserting both resolve to their own file's def; (b) `namespace ns { inline namespace ns { void foo(); } }` asserting `('ns', 'foo')` returns the single def rather than `'ambiguous'`.

Three Claude lanes converged here (testing, swarm test-CI, adversarial); Codex did not raise it, so treat as single-engine.

gitnexus/src/core/ingestion/languages/cpp/inline-namespaces.ts
```diff
@@ -115,11 +120,136 @@ export function isCppInlineNamespaceScope(scopeId: ScopeId): boolean {
...
+ * analyze on a 1,473-file C++ repo, 75% of total self-time in this one
+ * function (#2788). Mirrors the same fix #1990 applied to ADL
+ * (`pickCppAdlCandidates` → {@link AdlCandidateIndex}); per-site cost drops
+ * to two Map lookups.
+ *
+ * `byReceiver`: namespace simple name → member simple name → callable defs,
+ * in the exact order the legacy linear scan produced them (file-major; within
+ * a file, `parsed.scopes` declaration order; within a namespace, own
+ * `ownedDefs` before inline-namespace children, depth-first). Ordering is
+ * load-bearing: the caller returns `allHits[0]` for the single-hit case and
```
magyargergo: **P3 `[code-read]` — neither half of this justification holds for this resolver's return contract.**

- `allHits[0]` is only reached under `if (allHits.length === 1)` (:270) — trivially order-independent.
- `narrowOverloadCandidates` is not first-wins here. With `callsite === undefined` (the `resolveAdlCandidates` path, `cpp/scope-resolver.ts:366`) the arity and type filters are no-ops, so `narrowed.length > 1` and **both** branches of the final `if` at :290-293 return `'ambiguous'`. With a callsite, a single survivor is unique by definition.

The `keeps scan order` test (`test/unit/cpp-qualified-ns-index.test.ts:200`) concedes this in its own comment and then asserts *membership*: `outer` owns only `foo`, `v1` owns only `bar`, so both lookups are single-hit and pass under any traversal order. A build that collects inline children before own defs passes that test, the bench, and every `cpp-inline-namespace-*` fixture.

**To be clear about what is *not* wrong:** order equivalence with the legacy walker genuinely holds — 600,000 element-by-element candidate comparisons across 20,000 random workspaces found zero divergences. Preserving order is the right conservative choice. It is only the stated *reason* that points at a property the code does not have.

**Suggested fix:** either construct a case where order is observable (a member shadowed between the namespace and an inline child, resolved with a narrowing callsite) — which would also close a real fixture gap — or downgrade this to *"preserved for byte-identity, not currently observable through this resolver's return contract"* and rename the test to the parent/inline-child visibility it actually verifies.

Cross-engine: Claude testing + adversarial, Codex swarm F3, Codex ce.

gitnexus/bench/cpp-qualified-ns/baselines.json
```diff
@@ -0,0 +1,6 @@
...
+  "fingerprint": "30122b086b90fe1d56edc5acfd39ab7f26c00350f0e672123aaf607c7702e1e2",
+  "scaling_budget": 1.8,
+  "_scaling_note": "(t_large/t_small)/(1600/400). ~1.0 is linear; measured 0.93-1.21 on the indexed implementation. The pre-#2788 per-call-site workspace rescan measured 3.45 on the same corpus shape (at a reduced 100/400 scale, since 1600 files x 32k call sites of quadratic work does not finish in a CI step) — so the budget sits between the two bands and cannot be met by reintroducing the scan. Resolution is timed alone; the fingerprint's outcome strings are built in a separate untimed pass because their allocation cost grows with the corpus and would otherwise show up as scaling."
```
magyargergo: **P3 `[reproduced]` — the documented band "measured 0.93-1.21" does not reproduce, and a triager will misread the gap.**

Observed across five independent lanes in this environment: 1.012, 1.047–1.252, 1.063–1.475, 1.168–1.438, 1.232–1.314, and 1.359 — routinely above the documented ceiling. More usefully, one lane pinned four busy cores on a 24-core host and measured **2.102, which would fail `scaling_budget: 1.8`** and print *"per-call-site cost now grows with corpus size again (#2788)"* — actively misleading when the real cause is runner contention. Root cause is the measurement window: `small` is ~1.7 ms and `large` ~8–10 ms, short enough that one sustained preemption across all 7 reps of one scale moves the derived ratio a long way even with `fastest()`.

**Explicitly not filed as a defect of this PR.** The precedent was run as a control: `bench/callable-value-flow` documents "measured 1.14-1.16" against a 1.6 budget and produced 1.153/1.294/1.232 on the same host — equally out of band, same ~1.8 ms small arm. This is a family-wide bench-design property, and this PR is *more* conservative than the sibling it follows. CI itself measured 1.093 on the exact head.

**Suggested fix (documentation only):** widen the stated band to what is actually observed (≈0.9–1.5 unloaded) and add one sentence on what to do when the *scaling* arm fails — re-run before investigating — so it reads differently from the fingerprint arm. Optionally raise `CALLS_PER_FILE`/`SMALL` until `small` measures in the tens of ms. **Never** re-baseline `fingerprint`; the file already says so, correctly.

magyargergo:
All findings from the multi-engine review are fixed in `0015b0d64` (plus `b620773b1`, see the last bullet). Per finding:

| Finding | Status |
|---|---|
| **P1** index pins the C++ `ParsedFile` set past `phase.ts`'s `forceGc()` | Fixed — `WeakMap<readonly ParsedFile[], …>` per `moduleScopeIndexByPass`. Measured with `--expose-gc`: **61.2MB retained before, 0.1MB after**. The explicit clear stays, because `inlineNamespaceScopeIds` is a second input the key cannot observe. |
| **P1** bench blind to the receiver-miss path | Fixed — receivers now ~1 in 5 declared namespaces, rest plain identifiers; corpus gained a cross-file reopened namespace, an `ns::ns` nest, a member at both namespace and inline-child level, and real `Callsite` sites. **The rescan regression now measures 4.538 and FAILS** where it scored 1.279 and PASSED. Defeating the dedup now fails the fingerprint arm. Fingerprint moved once, deliberately, recorded in `_rebaseline_2788_review`. |
| **P2** O(depth²) build + unbounded recursion aborting analyze | Fixed structurally. A work-stack alone would only have turned `RangeError@8000` into `OOM@6000` — the eager table was quadratic in *memory* too (a depth-D chain legitimately holds D(D+1)/2 entries, since `v2::foo()` is a valid receiver at every level). Replaced with a lazy node graph: build linear, **depth 100,000 costs 133ms**. |
| **P2** "#1990 shipped without a scaling gate" is false | Fixed in the bench header, the CI step comment, the commit message and the PR body. Corrected to the accurate and stronger claim: that bench asserts `callsResolved === 0` so it never drives this path, and it is `skipIf`-gated and absent from the job's file list, so it has never run in CI. |
| **P2** cross-file merge + dedup arm untested | Fixed — two tests, both mutation-verified. Making each file clobber the previous fails **only** the new cross-file test, confirming nothing existing covered it. |
| **P3** "ordering is load-bearing" unsupported | Fixed — comment now says order is preserved for byte-identity, not because the return contract observes it; test renamed to the visibility it actually asserts. The rebuilt bench confirms this honestly: a pure reorder still passes, correctly. |
| **P3** "Same contract as `ensureAdlIndex`" overstates parity | Fixed in prose; **no guard added**, per the review's own adjudication — the sibling's guard protects a `seqByNodeId.get(...) ?? 0` defaulting read that this index has no analogue for. |
| **P3** second call site undocumented | Fixed — JSDoc names both, including the callsite-less `resolveAdlCandidates` path. |
| **P3** `_scaling_note` band unreproducible | Fixed — states the observed 1.28-1.45, small arm widened to ~14ms (spread halved), plus a triage line that a scaling failure is a timing signal to re-run, unlike the deterministic fingerprint arm. |
| Residual — memoized bucket shared by reference | Closed — buckets are frozen, so a future in-place sort in `overload-narrowing.ts` throws instead of silently corrupting later resolutions. |

Verification after the series: **840,000 differential probes** against the pre-#2788 walker extracted from base, **0 mismatches**, plus 48,000 candidate-order comparisons, also 0. `cpp.test.ts` **334/334**. Unit suite 10/10. tsc, eslint, prettier clean. Bench `--check` PASS.

Two follow-ups the review surfaced that are out of scope here: wiring #1990's two C++ benches into the `GITNEXUS_BENCH` step's file list so that guard actually runs, and moving `adl.ts`'s twin index to a WeakMap.

One embarrassment worth recording: `b620773b1` re-escapes a raw NUL byte that got written into the bench's fingerprint separator again — git classifies any file containing one as binary, so the whole bench briefly showed as `Bin` with no reviewable diff. Same defect this branch already fixed once before the review.

PR #2795, opened 2026-08-02
fix(embeddings): retry unparseable 200 responses and survive partial embedding failures (#2790)

### LINKED ISSUE #2790: HTTP embedding: unparseable 200 response not retried, single batch failure crashes pipeline, and checkpoint-resume   loop blocks recovery
# HTTP embedding endpoint: unparseable 200 response not retried, crashes entire pipeline

## Summary

When an OpenAI-compatible embedding endpoint returns HTTP 200 with an unparseable response body (e.g. truncated JSON, HTML error page, empty body), the retry logic is bypassed entirely and the entire embedding pipeline crashes with `HttpEmbeddingError: Embedding endpoint returned an unparseable response`. This is especially impactful for large-scale indexing jobs (300k+ nodes) where intermittent endpoint glitches are expected.

## Context

The embedding HTTP client (`gitnexus/src/core/embeddings/http-client.ts`) uses `resilientFetch` for retry logic, which only covers HTTP transport-layer failures:

- **Retried**: 5xx, 429, network errors (DNS, ECONNRESET)
- **Not retried**: HTTP 200 with an unparseable response body

When the endpoint returns 200, `resilientFetch` classifies it as `success` and returns immediately. The subsequent `resp.json()` parse failure is thrown as a terminal `HttpEmbeddingError` with no retry attempt.

This is a common failure mode for long-running embedding jobs against production endpoints. The error typically manifests after hours of successful processing (e.g. 178k/320k nodes completed), making the crash particularly costly.

## Expected behavior

- Unparseable 200 responses should be retried (with backoff), similar to 5xx/429 responses.
- If a single batch fails after retries, the pipeline should log the failure and continue with remaining batches, rather than crashing the entire analysis.
- The Phase 5 `embeddingCount === 0` fail-fast check should distinguish between "pipeline never ran" and "pipeline ran but partially failed", emitting a warning instead of a hard crash for the latter case.

## Actual behavior

Three cascading failures:

### 1. No retry on unparseable 200 response

```
❌ embedBatch failed for 32 texts
The custom embedding endpoint request failed.
Embedding endpoint returned an unparseable response (<endpoint>, batch 0)
```

The retry logic in `resilientFetch` (`gitnexus-shared/src/integrations/resilient-fetch.ts`) classifies HTTP 200 as `success` at line 105:

```typescript
if (resp.status >= 200 && resp.status < 400) return { kind: 'success', resp };
```

The parse failure happens after `resilientFetch` has already returned, so no retry is attempted (`http-client.ts:433-439`):

```typescript
try {
  data = (await resp.json()) as { data: EmbeddingItem[] };
} catch (err) {
  throw new HttpEmbeddingError(
    `Embedding endpoint returned an unparseable response (${safeUrl(url)}, batch ${batchIndex})`,
    { cause: err },
  );
}
```

### 2. Single batch failure crashes entire pipeline

In `embedding-pipeline.ts:641-646`, a single `embedBatch` failure is re-thrown immediately:

```typescript
try {
  embeddings = await embedBatch(subTexts, { signal: pipelineOptions.signal });
} catch (embedErr) {
  logger.error({ embedErr }, `❌ embedBatch failed for ${subTexts.length} texts`);
  throw embedErr;
}
```

All progress from previously successful batches is lost. For a 320k-node job that has already processed 178k nodes, this means hours of embedding work is discarded.

### 3. Failed run leaves DB in inconsistent state, next incremental run crashes

After the pipeline crash, the Phase 5 fail-fast check (`run-analyze.ts:2543-2548`) triggers on the next run:

```
Incremental: changed=0, added=0, deleted=0 (skipping wipe + 30850 unchanged file rows preserved)
Embedding generation completed without persisted embeddings. The index was not registered to avoid silently reporting embeddings: 0.
```

The root cause is a **checkpoint-induced embedding mode mismatch**:

1. **First run** (`--force --embeddings 0`): Full rebuild. Phase 3 (graph nodes) succeeds. Phase 4 (embedding) processes ~178k/320k nodes, then crashes on an unparseable response. Before the crash, `onCheckpointWindowStart` saves an `embeddingCheckpoint` to `meta.json` with `pendingNodeIds` (the window of nodes being processed). Critically, `onCheckpointWindowStart` saves `stats.embeddings` as `existingMeta?.stats?.embeddings` — the **previous** run's count (0 for a fresh index), not the current run's actual DB count.

2. **Second run** (no `--force`, no `--embeddings`): `existingMeta.embeddingCheckpoint` exists → `resumeEmbeddingCheckpoint = true` → `shouldGenerateEmbeddings = true` (line 1392: `derivedShouldGenerateEmbeddings || resumeEmbeddingCheckpoint`). The pipeline runs, but the endpoint may still be intermittently failing, or the pipeline re-processes all nodes (because `existingEmbeddings` from `loadCachedEmbeddings` may be empty — the DB was wiped in the first run's `--force` rebuild, and the first run's embedding rows were written to the fresh DB but the pipeline crashed before they could be fully checkpointed).

3. If the pipeline crashes again (or completes with 0 new embeddings because all nodes were already processed but the count query fails), Phase 5 queries `MATCH (e:CodeEmbedding) RETURN count(e)`. The `catch` block at line 2539-2541 silently swallows query errors (including "table does not exist"), leaving `embeddingCount = 0`.

4. The fail-fast check fires: `!embeddingSkipped && stats.nodes > 0 && embeddingCount === 0` → hard crash.

**The core issue**: `resumeEmbeddingCheckpoint` forces `shouldGenerateEmbeddings = true` even when the user didn't pass `--embeddings`, and the pipeline crash + checkpoint state creates a cycle where every subsequent run attempts embedding and fails.

## Steps to reproduce

1. Configure a custom embedding endpoint:

   ```bash
   export GITNEXUS_EMBEDDING_URL=https://your-endpoint.example.com/v1
   export GITNEXUS_EMBEDDING_API_KEY=your-key
   export GITNEXUS_EMBEDDING_MODEL=your-model
   ```

2. Index a large monorepo (300k+ nodes) with embeddings:

   ```bash
   gitnexus analyze --force --embeddings 0
   ```

3. If the endpoint intermittently returns a non-JSON response (e.g. due to rate limiting, timeout, or internal error), the pipeline crashes irrecoverably.
4. Re-running without `--force` triggers the `embeddingCount === 0` fail-fast crash.

## Environment

- OS: Linux (Docker container)
- Node: v22+
- GitNexus: commit `de84ad62` or later

### Fix 1: Retry unparseable 200 responses

In `http-client.ts`, wrap the `resp.json()` parse in a retry loop (or re-classify parse failures as retryable in `resilientFetch`). A parse failure after a 200 response is semantically equivalent to a 5xx — the server failed to produce a valid response.

### Fix 2: Tolerate individual batch failures in the pipeline

In `embedding-pipeline.ts`, catch `embedBatch` failures per sub-batch, log a warning, and continue with remaining batches. Track failed node IDs for retry on the next run (similar to the existing `pendingEmbeddingNodeIds` checkpoint mechanism).

```typescript
// Current (crashes on first failure):
try {
  embeddings = await embedBatch(subTexts, { signal: pipelineOptions.signal });
} catch (embedErr) {
  throw embedErr;
}

// Proposed (tolerate and continue):
try {
  embeddings = await embedBatch(subTexts, { signal: pipelineOptions.signal });
} catch (embedErr) {
  logger.warn(`embedBatch failed for ${subTexts.length} texts, skipping batch`);
  failedNodeIds.push(...subUpdates.map(u => u.nodeId));
  continue; // skip to next sub-batch
}
```

### Fix 3: Distinguish partial failure from total failure in Phase 5

The `embeddingCount === 0` fail-fast check should not crash when the pipeline actually ran but some batches failed. Instead:

- If `embeddingCount > 0` but less than expected → warning, continue
- If `embeddingCount === 0` and pipeline never ran → hard crash (current behavior)
- If `embeddingCount === 0` and pipeline ran but all batches failed → warning with recovery hint, not crash

Additionally, the `catch` block at line 2539-2541 silently swallows query errors (including "table does not exist"). This should at least log a warning so operators can distinguish "0 embeddings because the pipeline failed" from "0 embeddings because the count query failed".

### Fix 4: Break the checkpoint-resume crash loop

When `resumeEmbeddingCheckpoint = true` forces `shouldGenerateEmbeddings = true` on a run where the user didn't pass `--embeddings`, and the pipeline crashes again, the same checkpoint is re-saved, creating an infinite crash loop. The checkpoint should be cleared (or the `embeddingCheckpoint` should be marked as failed) after a pipeline crash, so the next run falls back to `preserveExistingEmbeddings` mode instead of retrying indefinitely.

## Related code

- `gitnexus/src/core/embeddings/http-client.ts` — HTTP embedding client with retry logic
- `gitnexus/src/core/embeddings/embedding-pipeline.ts` — embedding generation pipeline
- `gitnexus-shared/src/integrations/resilient-fetch.ts` — retry + circuit breaker
- `gitnexus/src/core/run-analyze.ts:2543-2548` — Phase 5 fail-fast check

magyargergo:
Fixes #2790.

A long running embedding job against an OpenAI compatible endpoint could lose hours of work to a single transient glitch, then refuse to recover on the next run. Four defects compounded, plus a fifth found on the way.

## 1. An unparseable HTTP 200 was never retried

`classifyOutcome` treats any 2xx as success, and the `resp.json()` parse ran after `resilientFetch` had already returned, so the parse failure surfaced as a terminal error with no retry.

Measured on `main`, with `GITNEXUS_EMBEDDING_MAX_ATTEMPTS=3`:

| response | fetch calls |
|---|---|
| 503 | 3 |
| 200 + truncated JSON | 1 |
| 200 + HTML error page | 1 |

The retry machinery was live. It just never saw the failure.

The parse and the response shape check now run inside the `fetchImpl` callback, so a bad body is classified as retryable and gets the same backoff as a 5xx. This also stops a garbage 200 from calling the circuit breaker's `recordSuccess()`, which previously erased accumulated failures and meant an endpoint alternating 5xx and garbage 200 could never trip it.

Considered and rejected: a `validateSuccess` hook on `resilientFetch`. A response body reads once, all four callers read it downstream and two read it off `ResilientFetchExhaustedError.response`, and several tests stub `fetch` with plain objects that have no `.clone()`. An outer retry loop was also rejected: it multiplies attempts 3x3 and still leaves `recordSuccess()` intact.

## 2. One failed sub-batch aborted the whole pipeline

Failures are now tolerated. The sub-batch's node ids are collected and all of their embedding rows are deleted, so those nodes hold zero rows and are re-embedded later.

Deleting rather than keeping partial rows is deliberate. Chunk arrays are flat over a 16 node batch and sliced by 8, so a node's chunks can straddle a sub-batch boundary, and surviving rows carry the current content hash. The hash maps collapse per chunk rows last row wins, so a partially embedded node would read as fresh forever and never regenerate its missing chunks.

Two guards bound the tolerance:

- 5 consecutive failed sub-batches aborts, for a total outage. It rethrows the first error of the streak rather than the last, because after 3 failures the circuit breaker opens and later errors degrade into "circuit open, retry in 30s" while the first still names the real defect.
- More than 25% of attempted sub-batches failed, gated behind a floor of 20 attempted sub-batches, for steady degradation. Without this an endpoint under load shedding never trips the consecutive ceiling and the run drops a large fraction of the corpus while exiting 0. The shape follows Resilience4j's circuit breaker (failure rate plus a minimum sample floor), the only surveyed design that handles the small repo case where one failure is a high ratio but not a signal.

## 3. The Phase 5 fail-fast could not tell "wrote nothing" from "could not ask"

The count query's catch was silent, so a missing table, a closed connection and a busy database all looked like zero embeddings and hard crashed the run. The count is now tri-state and only a known zero after real work is fatal.

A non-numeric count previously bypassed the gate entirely, because `Number()` returns `NaN` and `NaN === 0` is false, and then serialized as `embeddings: null`. An unverified count no longer certifies `capabilities.vectorSearch.status`.

## 4. The checkpoint wrote a completion shaped meta

This is the root cause of the symptom in the report, and it is not the resume loop the issue proposed.

`saveEmbeddingCheckpoint` advanced `lastCommit`, wrote the new `fileHashes` and cleared `incrementalInProgress`. The first checkpoint window fires before a single embedding exists, and on a full rebuild the graph is still in a staging database that a crash discards. The next run then diffed against the advanced hashes, saw no changes and preserved the old graph. That is the `changed=0, added=0, deleted=0 (skipping wipe...)` line in the report.

It now re-reads meta and replaces only the checkpoint, matching what the server endpoint already did.

## The partial run has to keep its checkpoint

A partially failed run retains its checkpoint with the failed ids in `pendingNodeIds`, so the next plain `analyze` regenerates them through the existing resume path.

Clearing it would have been silent data loss. A plain run derives `shouldGenerateEmbeddings: false` once embeddings exist, so the pipeline would never have run again and the dropped nodes would be gone until someone passed `--embeddings` or `--force`. The old crash and abort self healed only by accident, via the checkpoint its crash left behind. This was caught in review after all three implementation passes had missed it.

Verified behaviour: partial outage is tolerated and retried each run with `gitnexus status` reporting the index incomplete; total outage still hard fails as before; `--drop-embeddings` escapes both, and short circuits before identity validation so a config change plus the flag still escapes.

`POST /api/embed` is the pipeline's other caller and was discarding the result, reporting "Embeddings complete" for a partial run. It now persists the pending ids and reports the run as failed with the underlying endpoint error.

## 5. Separate bug found on the way

`POST /api/embed` never wrote `stats.embeddings`. A repo embedded purely through the server kept whatever count the last CLI `analyze` stamped, which is 0 for a repo analyzed without embeddings. The next CLI run read `existingEmbeddingCount = 0`, `deriveEmbeddingMode` returned `shouldLoadCache: false`, and `analyze --force` wiped the database with no cache load. Every server generated embedding was silently destroyed with no warning.

The route now measures the live count with the same query the CLI uses. The measurement is tri-state and never falls back to 0, because a wrong low value is exactly what arms the wipe. Committed separately since it is not #2790.

## Testing

318 tests green across 12 files, `tsc` clean.

New regression coverage:

- `http-embedding-body-retry.test.ts` (new file) — the retry asymmetry, both bad body shapes, the 4xx ordering pin, and the breaker erasure repair. Verified against the pre-fix code: 5 failed, 2 passed.
- `embedding-pipeline.test.ts` — sub-batch tolerance, the straddling chunk regression, both abort guards, cancellation, and three incremental mode tests that pin the data loss guard (an unchanged node sharing a batch with a failing node keeps its rows).
- `run-analyze.test.ts` — a real repo partial run healed by the next plain `analyze`, and the checkpoint meta regression.
- `run-analyze-fts-repair.test.ts` — every branch of the Phase 5 gate, which previously had zero coverage in either direction.
- `analyze-api.test.ts` — the `/api/embed` count fix, asserting the consequence through `deriveEmbeddingMode` rather than just the field.

Each fix was mutation verified: the fix was reverted, the new test confirmed failing, then restored.

Note for reviewers running `run-analyze.test.ts` locally: it needs a built `dist/` for the parse worker, and `GITNEXUS_WORKER_READY_TIMEOUT_MS` raised above the 5s default on slower hosts. The default gate fails with a message that reads like a different problem.


gitnexus/src/core/run-analyze.ts
```diff
@@ -2563,7 +2585,7 @@ async function runFullAnalysisInner(
         {
           forceReembedNodeIds: pendingEmbeddingNodeIds,
           onCheckpointWindowStart: async ({ nodeIds, ...checkpoint }) => {
-            await saveEmbeddingCheckpoint(checkpoint, nodeIds, existingMeta?.stats?.embeddings);
+            await saveEmbeddingCheckpoint(checkpoint, nodeIds);
           },
           onCheckpoint: async (checkpoint) => {
             await checkpointOnce();
```
magyargergo, line 2619: **P1 — the mid-run checkpoint counter was never migrated, so a diagnostic failure still kills the run this PR set out to save.** The real code is on the next lines (2592-2597, just outside this hunk): `const embeddings = Number(countRow?.cnt ?? countRow?.[0] ?? 0);` inside `onCheckpoint`, with **no try/catch and no `Number.isFinite` guard** — while the two sites this PR did fix (Phase 5 at 2652-2670, `measurePersistedEmbeddingCount` at `api.ts:850-868`) both guard. **Trigger:** a long run reaches a checkpoint window boundary while the count query throws for a reason unrelated to how many rows were written — DB busy, connection closed, read-only, the VECTOR-extension DML lock (#2623). The rejection propagates out of the callback, out of `runEmbeddingPipeline`, and aborts the whole analyze **before Phase 5 runs at all**, so the tri-state added at 2643-2652 for exactly this case never executes. **Variant:** a non-numeric cell yields `NaN`, which is `!== undefined`, so `saveEmbeddingCheckpoint` (2557) writes it and `JSON.stringify` serializes `stats.embeddings: null` to disk mid-run. **Fix:** call the same guarded counter here — export `measurePersistedEmbeddingCount` (or move it to a shared module) and pass `undefined` through, which `saveEmbeddingCheckpoint` already treats as "do not touch `stats.embeddings`". Reached independently by a Codex leg and four Claude lanes, and confirmed by the coordinator. [code-read]

gitnexus/src/core/run-analyze.ts
```diff
@@ -2590,25 +2638,100 @@ async function runFullAnalysisInner(
...
+    let embeddingCount: number | undefined;
     try {
       const embResult = await executeQuery(
         `MATCH (e:${EMBEDDING_TABLE_NAME}) RETURN count(e) AS cnt`,
       );
       const row = embResult?.[0];
-      embeddingCount = Number(row?.cnt ?? row?.[0] ?? 0);
-    } catch {
-      /* table may not exist if embeddings never ran */
+      const parsed = Number(row?.cnt ?? row?.[0] ?? 0);
```
magyargergo: **P1 — this fabricates a definite `0` in the exact case the PR's own docstring says it must never, and the new gate turns that `0` into a hard failure of a healthy run.** `api.ts:836` claims this counter is mirrored ("same query, same tri-state — the CLI and the server publish the SAME field, so they must measure it the same way") and `api.ts:844-846` states the rule ("A MISSING cell counts as unknown too, not 0 … Never a fabricated 0"), but `api.ts:856` uses `?? Number.NaN` while this line uses `?? 0`. **Trigger:** the count query answers with no row (`executeQuery` → `readQueryRows` can return `[]`; a driver hiccup, a column-key rename after a LadybugDB bump). `parsed = 0`, `Number.isFinite(0)` is true, so `embeddingCount` becomes a *measured* zero and the unknown branch is unreachable — then `attemptedEmbedding && stats.nodes > 0 && embeddingCount === 0` (2695) throws *"Embedding generation completed without persisted embeddings. The index was not registered…"* on a run where every embedding persisted. The same fabricated `0` also reaches `stats.embeddings`, which is the sole input to the next run's `existingEmbeddingCount` → `embedding-mode.ts:69` `hasExisting` → `shouldLoadCache: false` → a later `--force` wipes without loading the cache. **Fix:** `?? Number.NaN` here, or better, share one implementation with `api.ts:850` so the two publishers of the same field cannot drift again — they already drifted inside a single PR. Reached independently by a Codex leg and three Claude lanes, and confirmed by the coordinator. [code-read]

gitnexus/src/server/api.ts
```diff
@@ -1908,16 +2036,40 @@ export const createServer = async (port: number, host: string = '127.0.0.1') =>
...
             });
 
             // Don't overwrite 'failed' if the job was cancelled while the pipeline was running
             const current = embedJobManager.getJob(job.id);
             if (!current || current.status !== 'failed') {
-              embedJobManager.updateJob(job.id, { status: 'complete' });
+              // The pipeline reported `ready`, which the progress mapper above
+              // already turned into phase 'complete' — restate the real phase so
+              // the job record isn't self-contradictory (#2790).
+              embedJobManager.updateJob(
```
magyargergo, line 1917: **P1 — an SSE consumer sees a tolerated partial run as a clean success; this `failed` status is written into an already-closed stream.** The real close is at `api.ts:537-548` (`if (progress.phase === 'complete' || progress.phase === 'failed') { … res.end(); unsubscribe(); }`) and the mapping at `api.ts:1997` (`p.phase === 'ready' ? 'complete' : …`), both outside this diff. **Trigger, and it is a deterministic ordering rather than a race:** `runEmbeddingPipeline` emits `phase:'ready'`/`percent:100` unconditionally before returning (`embedding-pipeline.ts:867-872`), including when `failedNodeIds.size > 0`. So the sequence is always (1) `ready` → mapped to `complete`, (2) `mountSSEProgress` writes `event: complete` with `error: undefined`, `res.end()`, `unsubscribe()`, (3) only then does this block compute `partialRunError` and call `updateJob({status:'failed'})` — into a stream with no listener. `gitnexus-web` has already fired `onComplete` and set the UI to `ready`. **This is a regression in the PR's own headline claim:** pre-PR a failing sub-batch threw, the pipeline emitted `phase:'error'` → mapped to `'failed'`, and the client received it. `GET /api/embed/:jobId` pollers are unaffected, so the two consumers now disagree. **Fix:** stop mapping the pipeline's `ready` to a terminal SSE phase (map it to e.g. `finalizing` and let this explicit `updateJob` be the only terminal event — `updateJob` already synthesizes `{phase:'complete',percent:100}` for a status change), or make `mountSSEProgress` close on the terminal job **status** rather than a progress phase string. Found by two Claude lanes; all three Codex legs missed it and one explicitly refuted it — the coordinator verified the full chain. [code-read]

gitnexus/src/core/run-analyze.ts
```diff
@@ -2590,25 +2638,100 @@ async function runFullAnalysisInner(
...
+      );
+    }
+
+    // What we can honestly persist as the embedding count. When the count is
+    // unknown, carry the previous run's figure forward rather than writing
+    // `null`/0 — a wrong-LOW value is the dangerous direction: it is the sole
+    // input to the next run's `existingEmbeddingCount` → deriveEmbeddingMode →
+    // shouldLoadCache, and a false 0 makes the next `--force` run skip the
+    // cache load and discard every surviving embedding.
+    const persistedEmbeddingCount = embeddingCount ?? existingMeta?.stats?.embeddings;
```
magyargergo: **P1 — the unknown-count fallback reads a run-START snapshot, so it can overwrite a newer measured count with a stale zero and arm an embedding wipe.** `existingMeta` is assigned exactly once, at line 934, before any embedding work; this fallback therefore republishes the pre-run figure and can clobber the fresher count the terminal `onCheckpoint` already wrote to disk during this same run. **Trigger:** (1) existing meta says `stats.embeddings: 0`, or omits it; (2) a clean `analyze --embeddings` (or `/api/embed`) successfully inserts embeddings; (3) the final count query is unavailable — a transient failure after the terminal checkpoint count already succeeded; (4) finalization carries the old zero forward *and* clears the checkpoint, reporting success; (5) a later `analyze --force` reads `0` → `embedding-mode.ts:69` `hasExisting` false → `forceRegenerateEmbeddings` false → `shouldLoadCache` false → the rebuild destroys the live embeddings. The comment directly above says a wrong-LOW value is the dangerous direction — this path produces one. **Fix:** fall back to the latest *checkpoint* metadata rather than the run-start snapshot, and when embeddings were attempted but no finite count can be verified, retain an identity-matching recovery checkpoint (even with an empty pending set) instead of clearing it. Add a regression seeded with prior count `0` + an unanswerable final query that asserts the next `--force` still loads the cache. Found by both Codex legs (98% / 75%) and by no Claude lane; coordinator-confirmed. [code-read]

gitnexus/src/core/run-analyze.ts
```diff
@@ -2576,6 +2598,32 @@ async function runFullAnalysisInner(
...
+      // nothing: a plain `gitnexus analyze` derives shouldGenerateEmbeddings
+      // = false whenever the index already has embeddings, so the pipeline is
+      // never called and the nodes stay missing until someone passes
+      // --embeddings/--force/--drop-embeddings. Retaining the checkpoint is
+      // what restores the pre-#2790 heal: the resume path above forces
+      // shouldGenerateEmbeddings regardless of flags and feeds
+      // `pendingNodeIds` into `forceReembedNodeIds`. Stamped with THIS run's
+      // identity so a later model/provider change trips the resume mismatch
+      // error rather than resuming under a foreign identity.
+      if (embeddingResult.failedNodeIds.length > 0) {
```
magyargergo, line 2646: **P1 — a successful, exit-0 partial run now plants a checkpoint that hard-aborts later routine runs, and `--force` cannot clear it.** This block writes `pendingEmbeddingCheckpoint` (stamped with this run's `provider`, which is `http:<sha256(safeUrl(GITNEXUS_EMBEDDING_URL))>` in HTTP mode per `embedding-identity.ts:19-23`) into the final success-shaped meta at 2847. **Trigger:** `GITNEXUS_EMBEDDING_URL=… gitnexus analyze --embeddings` drops three nodes to a transient endpoint fault, warns, and **exits 0** with the checkpoint persisted. A later plain `gitnexus analyze` — a post-commit hook, a CI job, or just a different shell without those exports — hits `run-analyze.ts:1107`, resolves `provider` to `'local'`, and throws *"Cannot resume embedding checkpoint: the embedding provider configuration differs"* at 1115, **before any phase runs**. The graph is never refreshed again until someone passes `--drop-embeddings`. The gate at 1107 is keyed only on checkpoint presence and inspects only `options.dropEmbeddings` at 1108 — **`options.force` is never consulted**, so the flag an operator reaches for to start fresh does not help, and the warning at 2718-2724 names `--drop-embeddings` without saying `--force` is insufficient. A second consequence with the same root: a node the endpoint deterministically rejects produces a resume run too small for either guard to fire (under 5 sub-batches), so the pending set never shrinks — `getIndexIncompleteReasons` returns `embedding-checkpoint-pending` forever, `gitnexus status` never reports up-to-date, and the same-commit fast return at 1324-1329 is permanently disabled. Pre-PR the old throw was implicitly buying convergence. **Fix:** discriminate a crash-interrupted checkpoint (must fail closed) from a completed-but-partial one (those nodes hold zero rows, so nothing is at risk) — downgrade the identity mismatch to a warning that drops the pending set for the partial kind; let `--force` discard the checkpoint with a warning; and bound retries per pending node so the set converges. Found by three Claude lanes; the throw and the `--force` gap were both confirmed by the coordinator. [code-read]

gitnexus/src/core/embeddings/http-client.ts
```diff
@@ -368,7 +397,29 @@ const httpEmbedBatch = async (
...
+          // loop made that a one-shot terminal failure while a 503 got three
+          // attempts. Throwing from in here gives a bad body the same backoff
+          // and the same breaker accounting as any other endpoint fault (#2790).
+          let payload: { data: EmbeddingItem[] };
+          try {
+            payload = (await attemptResp.json()) as { data: EmbeddingItem[] };
+          } catch (err) {
+            throw new RetryableEmbeddingBodyError(unparseableMessage, { cause: err });
+          }
+          if (!Array.isArray(payload?.data) || !payload.data.every(isEmbeddingItem)) {
```
magyargergo, line 461: **P2 — a short or empty `data` array passes this shape check, so the very body class the PR set out to retry still gets one attempt and still resets the breaker.** `payload.data.every(isEmbeddingItem)` is vacuously true for `[]` and true for any array shorter than the request. The cardinality check (`items.length !== batch.length`, `http-client.ts:531-536`) and the dimension check (`543-552`) both live in `httpEmbed`, **after** `httpEmbedBatch` returns — outside the retry loop. **Trigger:** an endpoint under load answers `200` with `{"data": []}`, or with fewer vectors than inputs. `resilientFetch` classifies it `success`, calls **`breaker.recordSuccess()`** (erasing the outage signal), returns, and the mismatch then throws terminally with **1 attempt, not 3** — exactly the two properties #2790 was filed about, unfixed for this body shape. The consequence is now worse than pre-PR: the pipeline tolerates the resulting error by deleting those nodes' rows and continuing, instead of aborting loudly. **Fix:** validate `payload.data.length === batch.length` (and, if the batch length is available, the vector dimension) inside `fetchImpl` and throw `RetryableEmbeddingBodyError` on mismatch, so it gets the same backoff and breaker accounting as a truncated body. Add short-then-valid and persistently-short tests asserting attempt count and breaker state — the new body-retry tests cover null and non-array payloads but not cardinality. Raised by one Codex leg, missed by all Claude lanes, verified by the coordinator. [code-read]

gitnexus/src/core/embeddings/http-client.ts
```diff
@@ -368,7 +397,29 @@ const httpEmbedBatch = async (
...
+          // after `resilientFetch` returns. A reachable-but-wrong endpoint (a
+          // captive portal, a non-embeddings service, a truncated stream) can
+          // answer 200 with HTML or half a JSON document; parsing outside the
+          // loop made that a one-shot terminal failure while a 503 got three
+          // attempts. Throwing from in here gives a bad body the same backoff
+          // and the same breaker accounting as any other endpoint fault (#2790).
+          let payload: { data: EmbeddingItem[] };
+          try {
+            payload = (await attemptResp.json()) as { data: EmbeddingItem[] };
+          } catch (err) {
```
magyargergo, line 445: **P2 — this catch swallows *every* rejection from `.json()`, not just parse errors, so a body-phase timeout is retried 3×, charged to the shared circuit breaker, and misreported as "unparseable".** `RetryableEmbeddingBodyError` is a plain `Error`, so it defeats `classifyOutcome`'s DOMException check (`resilient-fetch.ts:96-101`) and lands in `retryable-network` — which retries and calls `breaker.recordFailure()` (`:250-253`) — instead of `terminal-network`, whose documented contract is "returned without retry AND without incrementing the breaker" via `recordNeutral()` (`:225-232`). **Reproduced** with a throwaway probe at `GITNEXUS_EMBEDDING_MAX_ATTEMPTS=3`: a `TimeoutError` DOMException raised inside `.json()` → **3 fetch calls**, message *"Embedding endpoint returned an unparseable response"*; the identical error raised from `fetch()` itself → **1 call**, message *"Embedding request timed out after 180000ms"*. Same error object, same policy question, opposite behaviour purely because of when it is raised. **Trigger:** a buffering proxy that flushes `200 OK` immediately then stalls the body — 3× the timeout and 3× the endpoint's GPU work per sub-batch, and because `CircuitBreaker.failureThreshold` defaults to 3, three such sub-batches open the process-global `embeddings-http` breaker (shared with interactive `httpEmbedQuery` in the server), which the pre-PR `recordNeutral()` path explicitly avoided. The operator is then shown "unparseable response" and will never reach for `GITNEXUS_EMBEDDING_HTTP_TIMEOUT_MS` — the misdiagnosis class #2386 added the typed boundary to close. **Fix:** `if (err instanceof DOMException && (err.name === 'TimeoutError' || err.name === 'AbortError')) throw err;` before wrapping, and move the `RetryableEmbeddingBodyError` branch in the outer catch after the `TimeoutError` branch. Note the interaction with `isCancellationError`: re-raising a bare `AbortError` must be checked against the caller's own signal, not the error name alone, or a transport abort becomes a false cancel. [reproduced]

gitnexus/src/core/embeddings/embedding-pipeline.ts
```diff
@@ -631,20 +729,67 @@ export const runEmbeddingPipeline = async (
...
+          if (consecutiveSubBatchFailures >= MAX_CONSECUTIVE_SUB_BATCH_FAILURES) {
+            // Bail out — but via `break`, not `throw`, so the cleanup DELETE below
+            // still runs: sub-batches that succeeded before the endpoint went dark
+            // left partial rows carrying the CURRENT hash, which is exactly the
+            // read-as-fresh corruption this path exists to prevent.
+            ceilingError = streakError;
+            break;
+          }
+          if (
+            subBatchesAttempted >= MIN_SUB_BATCHES_BEFORE_FAILURE_RATIO_GUARD &&
```
magyargergo: **P2 — the ratio guard cannot fire below its own 20-sub-batch floor, and a resume run is always below it.** `MIN_SUB_BATCHES_BEFORE_FAILURE_RATIO_GUARD = 20` needs 20 sub-batches ≈ 160 chunks ≈ ~80 embeddable nodes at the default `subBatchSize: 8` — the file's own comment at 350-351 states the 160-chunk figure ("At the default `subBatchSize` of 8 the floor is ~160 chunks of evidence before the rate can abort anything"), which is what makes the dead zone verifiable rather than speculative. **Trigger:** a 50-node repo whose endpoint rejects every other sub-batch fails 5 of 10 — a 50% loss rate — but `10 < 20` blocks the ratio guard, and every intervening success resets `consecutiveSubBatchFailures`, so the ceiling of 5 never trips either. ~25 nodes are silently dropped and `analyze` exits 0. The floor is a deliberate and well-argued choice for a *first* run over a small repo (one failure out of one sub-batch is 100% and means nothing), so this is not an objection to the shape — it is that **every resume run has exactly this shape by construction**: its node set is only the pending ids, so it produces few sub-batches no matter how large the repo is, and the guard that exists to stop a degraded endpoint from eating the corpus is structurally off in the one run whose entire purpose is retrying against that endpoint. Compounds with the non-convergence in the `run-analyze.ts:2612` finding. Separately, and by design rather than by defect: any steady loss rate below 25% never aborts, so a single run may drop up to a quarter of the corpus and still exit 0 behind one warning line — worth stating in the docs the docs-gap section asks for. **Fix:** scale the floor to the run rather than fixing it (e.g. `Math.max(5, Math.ceil(totalNodes / 16))` computed before the loop), or bound the absolute number of nodes one run may drop, so a resume run is judged on its own terms. Two Claude lanes. [code-read]

gitnexus/test/unit/analyze-api.test.ts
```diff
@@ -59,3 +64,318 @@ describe('analyze API logic', () => {
...
+ * embedding endpoint. There is no behavioral harness for it (the only existing
+ * /api/embed test, api-readonly-wiring.test.ts, is source-text matching), so the
+ * decision it makes is exercised through the exported helper it delegates to,
+ * and the observable job outcome through the real JobManager the route uses.
+ */
+describe('POST /api/embed partial-run outcome (#2790)', () => {
+  const identity = { model: 'test-model', dimensions: 384, provider: 'local' };
+
+  it('clears the checkpoint and reports no error on a clean run', async () => {
+    const { resolveEmbedRunOutcome } = await import('../../src/server/api.js');
```
magyargergo: **P2 — this dynamic import pulls the entire server module (Express, cors, the LadybugDB native adapter, MCP wiring, every route) to reach one pure function, and it costs ~20 s per test against a 30 s `testTimeout`.** **Reproduced** across three runs of the reviewing lane: run 1 **FAILED** — `Test timed out in 30000ms` on this exact test under ordinary concurrent-file load; run 2 passed at **19801 ms**; solo, with no concurrency, **22382 ms**. (The coordinator's own run of the same four files passed 143/143 in 28.09 s total, so it is intermittent rather than deterministic — but `import 29.78s` in that run's own summary shows where the budget goes.) `grep -rln "await import('../../src/server/api.js')" test/unit/` returns only this file, and the base revision contains no dynamic import here — this PR is the first thing in the suite to do it. That makes a synchronous call to `resolveEmbedRunOutcome` cost 65-100% of the repo's own per-test budget, which conflicts with the project's no-non-deterministic-tests rule. **Fix:** move `resolveEmbedRunOutcome` / `measurePersistedEmbeddingCount` / `withMeasuredEmbeddingCount` into a small standalone module with no Express/LadybugDB/MCP imports (alongside `embedding-mode.ts` would fit) and have `api.ts` import them — the tests then import a few-line module instead of the whole server. That extraction also happens to be the shared-counter fix requested on `run-analyze.ts:2591` and `:2657`, so one change closes three findings. [reproduced]

magyargergo (commented):
## Multi-tool tri-review — PR #2795

**Scope reviewed:** `990d79ba8c606b7f4df7671ab0df65f5d169e2b8..08672955c` — **10 files, +2735/-85** (4 source, 6 test). The merge-base is current `origin/main`, so main's own commits (#2793/#2794/#2796) are not in this diff.

**Engines and legs (all live).**
- **Engine A — Claude:** 14 lanes, all produced output. 6 GitNexus PR-swarm lanes (pr-facts, branch-hygiene, docs-DoD, test-CI, security-boundary, risk-architect) + 8 `ce-code-review` personas (correctness, project-standards, testing, maintainability, performance, api-contract, reliability, adversarial).
- **Engine B — Codex (gpt-5.6-sol / xhigh):** 3 legs, all produced real findings. `swarm` (Solo mode), `ce` (report-only), `adversarial` (native).
- **Coordinator:** independent verification of every claim below against the worktree, plus its own test run — `Test Files 4 passed (4)`, `Tests 143 passed (143)`, 28.09s.

Corroboration is labelled honestly throughout: *cross-engine* means a Codex leg **and** a Claude lane reached it independently; *single-engine* means several lanes of one engine agreed, which is consistency, not independent confirmation. Two findings below are single-engine **blind spots of the other engine** and are kept at P1 on the strength of a verified trigger, not on vote count.

---

### Review bar and state

**The bar this change had to clear.** Retry malformed successful HTTP responses inside the existing bounded retry/breaker budget; never leave a node partially embedded while carrying its current content hash; preserve a recovery path for every tolerated failure; keep CLI and `/api/embed` metadata honest under known-zero, positive, and unknown counts; never advance graph freshness during an unpublished rebuild; and exercise both production callers, not just extracted helpers.

**Problem being solved.** #2790: a `200` carrying an unparseable or wrong-shaped body was never retried (measured: `503` → 3 fetch calls, truncated-JSON `200` → 1, HTML `200` → 1), and one failed sub-batch discarded a multi-hour run. The PR also makes partial runs recoverable and stops a stale `stats.embeddings` from arming a later forced wipe.

**PR state.** Open, not draft, `MERGEABLE`, 0 formal reviews. 3 implementation commits plus one merge-from-main; the merge's second parent is the review base.

**Merge state: `checks pending`.** `mergeStateStatus: BLOCKED` is a fork Vercel preview gate ("Authorization required to deploy") plus shards re-triggered by the merge commit — not a failing gate. Detail in the CI section below.

**Branch hygiene: `clean feature/fix PR`** (with a harmless, merge-safe merge-from-main commit). No unrelated churn, no schema-version clash, no textual conflict against current `origin/main`. The third commit (`fix(server): record the embedding count after /api/embed`) is labelled "not #2790" by the author but is causally required by the first two — without it, embeddings the first two commits work to retain are the ones a later `--force` destroys. Keeping it here is the right call; `CONTRIBUTING.md:68` places no constraint on commit composition inside a PR.

**Repository history considered.** #2468 (merged 2026-07-16) introduced `saveEmbeddingCheckpoint` and the resume machinery — it is the PR that introduced the completion-shaped-meta bug fixed here. #2386 established the typed HTTP embedding error boundary and the catch-block ordering this PR extends. #2453 introduced `deleteStaleEmbeddingRows`, reused here for cleanup. #2623 (VECTOR DML lock, closed) and #2114 (pino `err` key, closed) are cited correctly in the new comments. No prior regression or revert in this area.

**Understanding of the change.** JSON parsing and the shape check move inside the retried `fetchImpl` callback so a bad `2xx` gets the same backoff and the same breaker accounting as a `503`. The pipeline tolerates bounded sub-batch failures, deleting *every* row of each affected node (node-keyed, because chunks straddle sub-batch boundaries) and returning their ids. Traversal and successful processing are separated. Both callers retain the pending ids. Metadata counting becomes tri-state, and checkpoint writes stop advancing completion-shaped freshness fields.

---

### What is solid, and should be said plainly

**The http-client half of this PR survived every attack both engines could construct.** Between them the two engines built and then disproved: catch-branch mis-ordering (`RetryableEmbeddingBodyError` is thrown raw from `resilient-fetch.ts:250-253`, so its branch at `http-client.ts:447` fires before `CircuitOpenError` / `TimeoutError` / `ResilientFetchExhaustedError` / the `sanitizeReason` fallback); a stale `parsed` closure across retries (`resilientFetch` returns immediately on `success`, so no attempt can follow an assignment); a consumed-body hazard on `ResilientFetchExhaustedError.response` — the exact lead carried in from #2386 — refuted, the only reader takes `.status`; body text or `SyntaxError.message` leaking to stderr via pino (pino's serializer skips `cause`, and `cause` is non-enumerable per spec); Cypher injection through `EMBEDDING_TABLE_NAME` (a compile-time constant); and unchanged 4xx/5xx retry semantics. One lane went further and showed the moved parse is a **security improvement**: pre-PR `.json()` ran after `resilientFetch` returned with no live timeout, so a dribbling endpoint could stall indefinitely.

**The checkpoint-meta rewrite is a genuine correctness win, confirmed by both engines.** The old `saveEmbeddingCheckpoint` wrote a completion-shaped meta — advanced `lastCommit`, new `fileHashes`, cleared `incrementalInProgress` — from a window that fires at `batchIndex === 0`, before a single embedding exists. The new writer re-reads on-disk meta and replaces only `embeddingCheckpoint` (plus `stats.embeddings` when measured), and the window-start call no longer restates a stale count. `incrementalInProgress` now correctly survives a checkpoint save. Both engines also verified the self-heal is real (`run-analyze.ts:1436-1437`), that `nodesProcessed`/`totalNodes` really are log-only, and that the `traversedNodes` switch was **necessary**, not cosmetic — `processedNodes` could lag and silently stop the terminal checkpoint from ever firing.

**The PR's diagnosis is better than the issue's.** #2790 proposed "the resume loop" as the root cause of failure #3; this PR explicitly contradicts that and names the completion-shaped checkpoint meta. Verified accurate.

**The tests are unusually rigorous.** Load-bearing tests assert on identity (specific node ids, `^...$`-anchored messages), several carry explicit fixture guards whose only job is to fail loudly if a future chunker change makes the real assertion vacuous, the counter arithmetic was hand-verified against the runtime logs (`11 of 21 (52%, limit 25%)` matches), and two tests drive **real** `runFullAnalysis` over a real temp git repo through a crash-then-heal scenario rather than mocking the pipeline. All eight states of the new tri-state gate are covered.

Hygiene is clean across the board: `git diff --check` clean, prettier clean, no `any` / `@ts-ignore` added, no `console.*` in the guarded directories, no bidi or hidden-Unicode hits, no schema-version clash, `CHANGELOG.md` correctly untouched, branch mergeable with no conflicts.

---

### Headline findings

**1. The mid-run checkpoint counter was never migrated, so a diagnostic failure still kills the whole run — bypassing the tri-state this PR exists to add.** `run-analyze.ts:2592-2597` still reads `const embeddings = Number(countRow?.cnt ?? countRow?.[0] ?? 0);` with **no try/catch and no `Number.isFinite` guard**, while the two sites the PR *did* fix (Phase 5 at `2652-2670`, `measurePersistedEmbeddingCount` at `api.ts:850-868`) both guard. A throw here rejects the callback, propagates out of `runEmbeddingPipeline`, and kills the analyze **before Phase 5 can run at all**. *Cross-engine (Codex adversarial `[high]` + 4 Claude lanes), coordinator-confirmed.*

**2. The CLI fabricates a definite `0` from a no-row answer, in the exact case this PR's own docstring says it must never — and the new gate turns that `0` into a hard failure of a healthy run.** `run-analyze.ts:2657` uses `?? 0`; `api.ts:856` uses `?? Number.NaN`, under a comment at `api.ts:836` claiming it "Mirrors run-analyze.ts's counter (same query, same tri-state)" and at `api.ts:844-846` "Never a fabricated 0". `Number.isFinite(0)` is true, so the unknown branch is unreachable for that case and `attemptedEmbedding && stats.nodes > 0 && embeddingCount === 0` throws on a run where everything persisted. *Cross-engine (Codex CE + 3 Claude lanes), coordinator-confirmed.*

> **Findings 1 and 2 are reported separately, not merged.** They share one root cause — three hand-copied bodies of "the same field, measured the same way", one of which contradicts the other two — and one fix closes both (extract `measurePersistedEmbeddingCount` and call it from all three sites). But they are different lines, different triggers (an *exception* vs a *no-row answer*), and different blast radii (the run dies mid-Phase-4 vs a successful run is hard-failed at Phase 5 and a wipe is armed), so they need separate anchors and separate regressions.

**3. A tolerated partial run now streams `event: complete` and closes the SSE stream before the job is ever marked failed — a regression in the PR's own headline claim.** The pipeline emits `phase:'ready'`/100 unconditionally (`embedding-pipeline.ts:867-872`), `api.ts:1997` maps it to `'complete'`, and `api.ts:537-548` writes the event then calls `res.end(); unsubscribe();`. The `updateJob({status:'failed'})` at `api.ts:2063` then fires into a closed stream. Pre-PR the pipeline threw → `phase:'error'` → `'failed'`, which **did** reach the client. *Single-engine (2 Claude lanes); **all three Codex legs missed it** and the swarm leg explicitly listed "Partial server status" as REFUTED. Coordinator confirmed it end-to-end — kept at P1 on the trigger, not the vote.*

**4. Phase 5's unknown-count fallback reads a run-START snapshot and can carry a stale zero forward over a newer measured count.** `run-analyze.ts:2733` — `embeddingCount ?? existingMeta?.stats?.embeddings` — where `existingMeta` was assigned once at line 934. With prior `stats.embeddings: 0` and an unavailable final count, finalization republishes the zero, clears the checkpoint, and reports success; the next `--force` reads `0` → `embedding-mode.ts:69` `hasExisting` false → `shouldLoadCache` false → the rebuild wipes live embeddings. *Single-engine (both Codex legs, P1/98% and P1/75%); **no Claude lane found it**. Coordinator confirmed — kept at P1.*

**5. A successful, exit-0 partial run now plants a durable checkpoint that hard-aborts later routine runs, and `--force` does not escape it.** `run-analyze.ts:2612-2626` writes `pendingEmbeddingCheckpoint` into the final success-shaped meta stamped with this run's `provider` (`http:<sha256(url)>` in HTTP mode). `run-analyze.ts:1115` then throws *"Cannot resume embedding checkpoint: the embedding provider configuration differs"* on every later `gitnexus analyze` from a shell, hook, or CI job without the embedding env — before any phase runs. The gate at `1107` is keyed only on checkpoint presence and inspects only `options.dropEmbeddings`; `options.force` is never consulted, and the warning at `2718-2724` names `--drop-embeddings` without saying `--force` is insufficient. Pre-PR this state existed only after a visible crash. *Single-engine (3 Claude lanes), coordinator-confirmed on both the throw and the `--force` gap.*

---

### Inline-anchored findings

The coordinator will attach these as inline comments. Each carries a trigger, worktree-verified evidence, and a `[reproduced]` / `[code-read]` tag.

| # | Sev | Location | Problem | Corroboration |
|---|-----|----------|---------|---------------|
| 1 | P1 | `run-analyze.ts:2591` (real code 2592-2597) | Un-migrated `onCheckpoint` count: no try/catch, no `isFinite` | cross-engine, coordinator |
| 2 | P1 | `run-analyze.ts:2657` | `?? 0` fabricates a definite zero; contradicts `api.ts:844-846` | cross-engine, coordinator |
| 3 | P1 | `api.ts:2063` (real close at 537-548, mapper at 1997) | SSE terminal `complete` precedes the `failed` status | Claude-only, coordinator |
| 4 | P1 | `run-analyze.ts:2733` | Run-start snapshot carries a stale zero over a newer count | Codex-only, coordinator |
| 5 | P1 | `run-analyze.ts:2612` | Exit-0 partial run plants an identity-mismatch landmine; `--force` can't clear it | Claude-only, coordinator |
| 6 | P2 | `http-client.ts:418` (mismatch throws at 531-536 / 543-552) | Cardinality/dimension checks still outside the retry loop | Codex-only, coordinator |
| 7 | P2 | `http-client.ts:415` | Body-phase abort/timeout laundered into a retryable body error | Claude-only, **reproduced** |
| 8 | P2 | `embedding-pipeline.ts:778` | Ratio guard has a dead zone below its own 20-sub-batch floor — and that is exactly the shape of a resume run | 2 Claude lanes |
| 9 | P2 | `analyze-api.test.ts:89` | Dynamic-importing the whole server module for a pure function; ~20s per test, one observed timeout | 1 lane, **reproduced** |

Note on #6: a `200` carrying `{"data": []}` or fewer vectors than inputs passes the in-`fetchImpl` shape check, `resilientFetch` classifies it `success` and calls `breaker.recordSuccess()`, and the mismatch then throws terminally with **1 attempt, not 3** — precisely the two properties this PR set out to fix, unfixed for that body shape. The consequence is now *worse* than pre-PR, because the pipeline tolerates the resulting error by deleting those nodes' rows instead of aborting loudly.

---

### Lower-priority observations (not anchored)

- **A failing cleanup DELETE swallows the ceiling error entirely** (`embedding-pipeline.ts:813-817`). If `deleteStaleEmbeddingRows` throws (e.g. a busy/read-only DB), line 817's `throw ceilingError.err` is never reached and the real endpoint failure — the error the PR deliberately preserves as `streakFirstError` — is lost from both message and `cause`. P3, one lane, high confidence.
- **Cross-process meta clobber.** Phase 5 builds a fresh `RepoMeta` literal rather than spreading the on-disk meta, so a checkpoint written by a concurrent `/api/embed` between `loadMeta` and the Phase 5 write is erased, orphaning nodes that hold zero rows and are named in no checkpoint. Mitigating: the sibling stamp at `run-analyze.ts:1050-1056` already re-reads for exactly this reason, and the new checkpoint writer adopted that pattern — Phase 5 and three dirty-flag stamps did not. Related to the known "#2658 analyze has no writer lock" gap. P2, one lane, requires a concurrency window.
- **Abort blast radius grew.** Pre-PR the first rejection threw, so at most one batch's stale rows had been deleted; a load-shedding endpoint can now walk a large fraction of the corpus deleting rows before the ratio guard fires, and when it does fire Phase 5 never runs, leaving `stats.embeddings` advertising a pre-run count over a table that just lost rows. Self-heals via the window checkpoint, but the meta is wrong in the interim. P2, one lane.
- **`attemptedEmbedding` weakens the old gate.** `queryEmbeddableNodes` swallows per-label query failures and returns `[]`, which is indistinguishable from a legitimately empty repo; that path now yields `attemptedEmbedding === false` and registers an index with `stats.embeddings: 0` and exit 0, where the old gate hard-failed. P2, one lane, confidence 58.
- **Two consumers now disagree about the same partial run.** Via SSE it looks like a clean success (finding 3); via `GET /api/embed/:jobId` and the web UI it is an undifferentiated hard `failed`/red "retry" chip identical to a 100%-loss run, because `AnalyzeJob.status` has no partial member. The PR acknowledges the union limitation in a comment; the mismatch between the two consumers is the part worth a decision.
- **Maintainability.** `embedding-pipeline.ts` crosses 1000 lines (900 → 1075); the sub-batch loop adds 7 interacting mutable variables; `ceilingError` carries two different abort reasons; the `pendingEmbeddingCheckpoint` literal is built twice with the same 7 fields and the same `totalNodes` reconstruction. All mechanical extractions, none blocking.
- **Pre-existing, in a changed file:** `embedding-pipeline.ts:178/223/896` still log under `{ error }`, the non-serializing key this diff correctly fixed to `{ err }` (#2114) three lines away.

### Refuted suspicions — validation is a feature here

A large share of both engines' effort went into attacks that were built and then disproved with traces. Recording them so they are not re-litigated:

`http-client.ts` catch ordering · `parsed` staleness across retries · a consumed body reaching `ResilientFetchExhaustedError.response` (the standing #2386 lead — the only reader takes `.status`) · body text or `SyntaxError.message` leaking through pino or `sanitizeReason` · Cypher injection via `EMBEDDING_TABLE_NAME` · new exports widening the attack surface · a hang on a slow body read (refuted empirically: `.json()` rejects at the configured timeout) · changed 4xx/5xx retry semantics · `batchIndex` being a batch counter rather than a node offset · a missed terminal checkpoint for any batch-size combination · `nodesProcessed`/`totalNodes` being read by anything but the resume log · the straddling-chunk trap (cleanup is keyed on `nodeId`, so all of a node's rows go) · half-embedded nodes after a ceiling `break` · negative `processedNodes` · cancellation absorbed as a tolerable failure on the server path (the signal is threaded) · `--drop-embeddings` failing to escape a checkpoint · the 50k node cap stranding a resume run · a third caller of `runEmbeddingPipeline` · the checkpoint rewrite dropping the other 11 meta fields · `lastCommit: ''` being unsafe on a first-ever analyze · an unverified count certifying `capabilities.vectorSearch`.

One correction to the record: a branch-hygiene lane reported "66 changed files". That is wrong — it diffed from the branch point rather than the merge-base. The PR is 10 files.

### Docs gaps

- `GUARDRAILS.md:43-47` ("Embeddings vanished after analyze") is written as an exhaustive list — *"the only ways to end up at zero are…"* — and this PR adds a fourth way, while breaking the Sign's own trigger (`stats.embeddings` is `0`): a partial run now stamps an honest **non-zero** count *and* sets `embeddingCheckpoint`, so the operator's symptom is `incompleteReasons: ["embedding-checkpoint-pending"]`, which the Sign cannot match.
- `RUNBOOK.md:51-62` has no recovery path for a partial run, and line 59 still tells the operator to read `stats.embeddings` as ground truth ("0 means none") when it can now be a carried-forward figure.
- `repo-manager.ts:295-315` — the JSDoc that *is* the `embeddingCheckpoint` contract still says it is cleared once the run completes and that `pendingNodeIds` "may have only a subset of their chunks persisted". Both are now false: a successful run can leave it set, and those nodes definitively hold zero rows.
- The PR body has no risk/rollback section, despite two deliberate new degradation modes (`vectorSearch.status: 'unavailable'` on a failed probe; a carried-forward `stats.embeddings`).

### Test gaps

- No test drives the real `mountSSEProgress` against a partial run. The one test named for it subscribes to `onProgress` *after* emitting the `complete` event, so it structurally cannot observe finding 3 — it would pass with or without the bug.
- No test combines a sub-batch failure with `checkpointEveryNodes`/`onCheckpoint`, so the `traversedNodes`-vs-`processedNodes` fix (`embedding-pipeline.ts:839`) is unpinned: reintroducing `processedNodes` there passes every test in this PR.
- No test feeds the same inputs (`[]`, `[{}]`, `[{cnt:'x'}]`) to both counters and asserts they agree — it would fail today on `[]` (finding 2).
- No test covers the mid-run `onCheckpoint` count throwing (finding 1), a body-phase timeout (finding 7, one attempt + "timed out"), a resume run that fails identically twice (finding 5's non-convergence), or the prior-count-`0` + unknown-final-count production sequence (finding 4) — the unknown-count test seeds `137`, which is safe.
- The security property the PR claims — body text never reaches the user-facing error — has no negative assertion; swapping `err.terminalMessage` for `err.cause.message` at `http-client.ts:447` passes every existing test.
- `run-analyze.test.ts:523` uses `toBeGreaterThan(0)` where the fixture is fully determined and an exact set assertion is possible; the collateral-damage direction (dropping nodes that did not fail) is what the safety argument rests on.
- Two wiring `describe` blocks assert on the literal source text of `api.ts`. This matches an established repo convention (`api-readonly-wiring.test.ts`) and the decision logic *is* separately tested behaviourally, so it is noted, not objected to — but for a route this data-loss-sensitive, a source-text match should be the fallback, not the primary contract check.

### CI

`mergeStateStatus: BLOCKED` is **not a failing gate.** It is (a) Vercel's fork preview gate — "Authorization required to deploy", needs maintainer approval — and (b) shards re-triggered by the merge-with-main commit still pending. The pre-merge run at `89085d7c` was fully green: **16247 passed / 0 failed / 72 skipped**, tsc clean. Green at observation: gitleaks, CodeQL, lint, format, typecheck ×2, autofix, e2e web-module check, eval containment (ubuntu + windows), eval locked pytest, node floor 22.18, packaged install smoke, tree-sitter ABI ×3, macos 1/3 and 2/3. Pending: Build & Push, benchmarks, ubuntu/coverage 1-3, windows 1-3, macos 3/3, windows smoke. The coordinator's own run of the four locally-runnable test files: 143/143 passed. `run-analyze.test.ts` and `run-analyze-fts-repair.test.ts` need a built `dist` and could not run in this environment — an environment limit, not a defect; CI builds first (`ci-tests.yml:38`).

### Open questions

- **Finding 5 needs a product decision, not just a patch.** Should a *completed-but-partial* checkpoint be treated the same as a *crash-interrupted* one? They have opposite risk profiles — the partial kind names nodes that provably hold zero rows, so failing closed on an identity mismatch protects nothing — but they share one field today.
- **Findings 3 and the `AnalyzeJob` union.** With no partial member, one consumer sees a clean success and the other sees a total failure. Adding a status member, or reporting partial as `complete` with a warning payload, is a call for the maintainer.

### Final verdict

**`not production-ready`** — recommend addressing the P1s before merge.

The retry, cleanup, straddling-chunk, and checkpoint-meta mechanics are carefully built and withstood roughly twenty constructed attacks across both engines; the `http-client` half in particular came through clean, and the checkpoint-meta rewrite fixes a real, reported data-loss mechanism. But five coordinator-confirmed P1s remain, and two of them are **behavioral regressions this PR introduces**: an SSE consumer now sees a tolerated partial run as a clean success (pre-PR the throw did reach the client), and a successful exit-0 run now plants a checkpoint that hard-aborts every later plain `gitnexus analyze` run without the embedding env — a state that previously existed only after a visible crash. The other three sit on the PR's own central thesis: the tri-state count doctrine is applied at two of its three CLI sites, the two counters that a shipped comment says "measure it the same way" disagree on the missing-row case, and the unknown-count fallback can still republish a stale zero and arm the exact wipe fix #5 exists to prevent. Four of the five share one root cause and one fix — extract the counter into a module all three sites call, which also resolves the heavy-import test flake. Suggested order: #1, #2, #4 (one extraction), then #3, then #5.

---

*This is an automated multi-tool digest (Claude ×14 lanes + Codex ×3 legs, merged by a synthesis pass, gated by an independent critic). Please verify each finding against the code before acting on it.*

gitnexus/test/unit/analyze-api.test.ts
magyargergo: Fixed in 52b3a4528, extended in e76058b9c. The pure helpers moved to `server/embed-run-outcome.ts` and `server/sse-progress.ts`, so the test no longer imports Express/LadybugDB/MCP to reach them: 22s → 1.6s. `server-sse-payload.test.ts` — which still reached `mountSSEProgress` through `api.ts` and so was still paying the cost the extraction was meant to remove — was repointed at the extracted module: 24.60s → 0.408s, with a control run confirming `api.ts` was the entire cost.

gitnexus/src/core/embeddings/http-client.ts
magyargergo: Fixed in 01053db5f, hardened in e76058b9c. Abort-like DOMExceptions are re-raised unwrapped before the parse-error wrap, so a body-phase timeout stays terminal-network (1 attempt, `recordNeutral`, "timed out" message). The predicate is no longer a hand-copy: `resilient-fetch.ts` exports `isTerminalNetworkError` and `classifyOutcome` calls it, so the caller and the classifier cannot drift.

gitnexus/src/core/embeddings/http-client.ts
magyargergo: Fixed in 01053db5f. `payload.data.length !== batch.length` is now checked inside the retried `fetchImpl` and throws `RetryableEmbeddingBodyError`, so a short or empty body gets the same backoff and the same breaker accounting as a truncated one instead of one attempt plus `recordSuccess()`. The `httpEmbed` check is kept as a documented backstop. Covered by short-then-valid, persistently-short, and breaker tests.

gitnexus/src/core/embeddings/embedding-pipeline.ts
magyargergo: Fixed in 21c316696, refined in e76058b9c. The floor is now `clamp(ceil(totalNodes / subBatchSize / 2), 5, 20)` — derived from the run's actual sub-batch size rather than a constant that assumed the default of 8. The cleanup pass caught that a fixed constant left the guard structurally off again at a raised `subBatchSize` (192 nodes at 32 yields 6 sub-batches against a floor of 12), which is the same failure this finding describes; there is now a test at a non-default sub-batch size.

gitnexus/src/core/run-analyze.ts
magyargergo: Fixed in 37c2d3061. The `?? 0` site is gone; the shared counter uses `?? Number.NaN`, so a no-row answer is `unknown` rather than a measured zero and can no longer trip the "completed without persisted embeddings" gate on a run where everything persisted. Pinned by `test/unit/embedding-count.test.ts`, which feeds both former call sites the same `[]` / `[{}]` / `[{cnt:'x'}]` inputs and asserts they agree.

gitnexus/src/core/run-analyze.ts
magyargergo: Fixed in 37c2d3061. `measurePersistedEmbeddingCount` was extracted to `core/embedding-count.ts` and all three call sites — the mid-run `onCheckpoint`, Phase 5, and `/api/embed` — now share it. On `unknown` it passes `undefined`, which `saveEmbeddingCheckpoint` already treats as "leave `stats.embeddings` alone", so a count-query throw no longer rejects out of `runEmbeddingPipeline` before Phase 5 can run.

gitnexus/src/core/run-analyze.ts
magyargergo: Fixed in 37c2d3061, completed in e76058b9c. `embeddingCheckpoint.kind` separates `'interrupted'` (may be half-written — still fails closed) from `'partial'` (nodes provably hold zero rows — an identity mismatch drops the pending set with a warning). `--force` now discards a checkpoint, and `attempts` bounds the retry at 3 so the set converges. Note the cleanup pass found `/api/embed` read **neither** field, so a CLI-written `'partial'` marker resumed through the server still hit this wedge; both gates now route through one `decideEmbeddingResume`.

gitnexus/src/core/run-analyze.ts
magyargergo: Fixed in 37c2d3061. The fallback re-reads the latest on-disk meta instead of the run-start `existingMeta` snapshot, so it no longer clobbers the fresher count the terminal checkpoint wrote; and when embeddings were attempted but no finite count can be verified, a recovery marker is retained rather than cleared. e76058b9c folds "unknown ⇒ carry forward, never fabricate 0" into one `resolvePersistedEmbeddingCount`.

gitnexus/src/server/api.ts
magyargergo: Fixed in 52b3a4528. `mountSSEProgress` (now `server/sse-progress.ts`) closes on the terminal job **status** rather than a progress-phase string, and the pipeline's `ready` maps to `finalizing`. Remapping `ready` alone was rejected: `error -> 'failed'` has the identical shape and would emit `event: failed` with `error: undefined` before the catch fills it in. A test now drives the real relay against a partial run, subscribing before the misleading event.

Issue #2084, opened 2026-06-08
M4 — Interprocedural taint: function summaries over resolved CALLS


magyargergo:
> Part of **Epic A — Reliable taint analysis on a PDG-expandable substrate** (tracking issue). Completes Epic A.

## Goal
Make taint **reliable where real vulns live** — across function and file boundaries — via per-function summaries propagated over GitNexus's already type-resolved `CALLS` edges (the hard part is already solved).

## Subtasks
- [ ] Per-function taint **summaries** ("param *i* flows to return", "param *i* flows to arg *j* of callee X").
- [ ] Propagate summaries across resolved `CALLS` edges (context-insensitive first).
- [ ] Cross-file source→sink flows.
- [ ] Recursion / cycle-safe summary fixpoint.
- [ ] **Phase-registry refactor:** flip `buildPhaseList` fully onto the M0 registry seam now that the pipeline has grown (maintainer flag).
- [ ] Incremental invalidation: a changed file invalidates only its functions' CFG/REACHING_DEF/taint.

## Acceptance criteria
- [ ] Fixture: taint crosses ≥1 function boundary and ≥1 file boundary.
- [ ] Recursive call graph terminates (summary fixpoint).
- [ ] Incremental test: editing one file re-runs only affected functions.

## Sequencing
- **Depends on:** M3 (intra-proc taint).
- **Blocks:** — (Epic A complete.)

**Estimate:** story-points:8

Issue #480, opened 2026-03-23
impact/context return empty results for Java classes — edges target Constructor/File nodes, not Class nodes


NeosiaNexus:
## Environment

- **GitNexus version:** 1.4.7 (installed via bun)
- **OS:** WSL2 (Linux 6.6.87.2-microsoft-standard-WSL2), Ubuntu
- **Language:** Java 21 monorepo (~10,800 symbols, ~35,300 edges, 300 flows)
- **LadybugDB extension version:** 0.15.0

### Steps to reproduce

```
gitnexus_impact({target: "SessionTracker", direction: "upstream"})
→ impactedCount: 0, risk: LOW, byDepth: {}
```

### Expected

Should return 5 callers (1 production + 4 tests), as proven by direct Cypher query:

```cypher
MATCH (a)-[r:CodeRelation {type: 'CALLS'}]->(b:Constructor)
WHERE b.name = 'SessionTracker'
RETURN a.name, a.filePath, r.type
```

Returns:

| a.name | a.filePath |
|--------|-----------|
| registerSessionTracker | core/.../ServerBootstrap.java |
| setUp | api/.../SessionTrackerTest.java |
| constructor_nullGameMode_accepted | api/.../SessionTrackerTest.java |
| constructor_nullServerId_accepted | api/.../SessionTrackerTest.java |
| startPlayerSession_passesGameModeAndServerId | api/.../SessionTrackerTest.java |

### Root cause

`impact()` resolves "SessionTracker" to the **Class** node, but:
- `CALLS` edges point to the **Constructor** node (not Class)
- `IMPORTS` edges point to the **File** node (not Class)
- The Class node only has `DEFINES` incoming (from File) and `HAS_METHOD`/`HAS_PROPERTY` outgoing

So `impact(upstream)` traverses from Class and finds zero CALLS/IMPORTS edges. The tool doesn't walk `Class ← HAS_METHOD ← Constructor` to discover the actual callers.

Same issue affects `RankPermissionHandler` (1 caller via Constructor, 10 importers via File — impact returns 0 for both).

### Steps to reproduce

```
gitnexus_context({name: "SessionTracker", file_path: "api/.../SessionTracker.java"})
→ incoming: {}  (empty)
```

### Expected

Should show callers and importers. Same root cause as Bug 1 — edges target Constructor/File, not Class.

### Contrast: context() on methods works perfectly

```
[...]

PR #2796, opened 2026-08-02
fix(mcp): make impact/context reproducible — deterministic ordering on every capped query (#2787)

### LINKED ISSUE #2787: impact and context return different results on every run against a large index (1.6.9)
## Summary

On a large index, `impact` and `context` return **different results on every invocation** with no
change to the code, the index, or the arguments. The underlying graph is fine — a raw `cypher`
query against the same index answers identically every time — so this looks like it is in the
query/traversal layer rather than the data.

This matters because the generated `CLAUDE.md` / `AGENTS.md` guidance instructs agents to run
`impact` before editing a symbol and to warn the user on `HIGH`/`CRITICAL`. With a risk level that
moves on its own, that warning fires essentially at random.

## Environment

- `gitnexus` **1.6.9** (installed globally)
- Node **v22.6.0**, macOS (Darwin arm64)
- Index: **6893 nodes / 15852 relationships / 300 flows**, PHP codebase
- Analyzed with `gitnexus analyze` immediately before these runs; no edits in between

## Reproduction

Target is a `public static` method on a domain class. Its incoming edges in the graph are exactly
**16 `CALLS`** (1 production caller + 15 call sites in one test file) plus 1 `HAS_METHOD` from the
owning class — verified against `git grep`.

### 1. `impact` — same command, six consecutive runs

```
gitnexus impact <symbol> -r <repo> -f <path> --summary-only
```

| run | `impactedCount` | `risk` |
|-----|-----------------|--------|
| 1 | 16 | MEDIUM |
| 2 | 25 | HIGH |
| 3 | 25 | HIGH |
| 4 | 65 | CRITICAL |
| 5 | 44 | CRITICAL |
| 6 | 44 | CRITICAL |

Full (non-`--summary-only`) mode swings the same way: `65 / 25 / 16 / 16 / 25 / 25`.
A second, unrelated symbol in the same file behaves identically (`44 / 65 / 16`), so it is not
specific to one target.

### 2. It is not depth-related — `--depth 1` is unstable too

```
gitnexus impact <symbol> -r <repo> -f <path> --depth 1 --summary-only
→ 17 / 17 / 47 / 10 / 17 / 17
```

Depth 1 is just the direct in-neighbours of a single node, and the correct answer is 17 incoming
edges. Runs returning 10 and 47 are under- and over-counts of a single-hop lookup, which suggests
the problem is in resolving/expanding the node rather than in multi-hop traversal budgeting.

### 3. `context` is affected as well

Five consecutive runs of `gitnexus context <symbol> -r <repo>` produced five different outputs
(different md5). `incoming` and `outgoing` stayed at 30, but `processes` varied:

```
263 / 1067 / 139 / 1067 / 1055
```

### 4. The graph itself is stable and correct

```
gitnexus cypher "MATCH (a)-[r]->(b) WHERE b.name='<symbol>' AND r.type='CALLS' \
                 RETURN count(a) AS callers" -r <repo>
→ 16, 16, 16   (three runs, matches git grep exactly)
```

### 5. Does not reproduce on a small repo

A synthetic PHP repo (9 classes + 1 test file, **55 nodes / 93 edges**) indexed with the same
version returns `impactedCount: 10, risk: CRITICAL` on **8/8** runs — perfectly stable, and 10 is
the correct answer. So the instability appears to need a large graph.

## What I would expect

`impact` and `context` are read-only queries over an immutable index — repeated invocations with
identical arguments should return identical output.

## Notes that may help narrow it down

- The depth-1 result list also contains symbols that have no edge to the target at all (e.g. an
  `execute` method on an unrelated class, a `tearDown` in an unrelated test, and route files).
  The one true caller is itself named `execute`, which is a very common method name in this
  codebase — so a name-keyed lookup somewhere returning an arbitrary match among duplicates would
  explain both the wrong entries and the run-to-run variation.
- `cypher` on the same index is stable, so the persisted graph is not the problem.
- Nothing else was running against the index during these runs.

Happy to run further diagnostics or try a patch against this index — it reproduces on demand.

magyargergo:
Fixes #2787.

## The bug

`MATCH (n) WHERE n.name = $symName RETURN ... LIMIT 20` in `resolveSymbolCandidates` had no `ORDER BY`. LadybugDB returns an **arbitrary subset** when more rows match than the cap, and a **different subset from one process to the next** — so `impact` and `context` resolved a different symbol on every invocation with no change to the code, the index, or the arguments.

That is the failure the reporter saw. The generated `CLAUDE.md` / `AGENTS.md` tells agents to run `impact` before editing a symbol and to warn on `HIGH`/`CRITICAL`; with a risk level that moves on its own, that warning fired essentially at random.

Measured on a 240k-node index (`get` = 34 nodes, `constructor` = 92):

| probe | runs | distinct outputs |
|---|---|---|
| `cypher ... RETURN n.id LIMIT 20` | 20 | **2** |
| same + `ORDER BY n.id` | 20 | 1 |
| `gitnexus context get` | 14 | **2** |
| `cypher ... count(n)` (aggregate) | 5 | 1 |

The flipped `context` run swapped `Method:.../callee-id-sink.ts:get#1` for `Const:gitnexus/bench/cross-repo-trace/verify.mjs:get` — same command, same immutable index. `cypher` with an aggregate was stable throughout, matching the reporter's "the graph itself is fine".

Flip rate here was ~5-8% (linux x64, 24 cores) against near-every-run for the reporter on macOS arm64. **A fix needs ≥14 runs to verify; 6 proves nothing.**

## The reporter's three symptoms are three different defects

| symptom | defect |
|---|---|
| `impactedCount` 16/25/44/65 | the resolver window moved, so a different symbol resolved, so a different top-6 got probed |
| depth-1 counts 17/10/47, plus entries with no edge to the target | **not a traversal bug.** `_runImpactBFS`'s per-depth query has no `LIMIT` at all and `impactedCount` is a `Set` size. These were different *symbols* |
| `context.processes` 263/1067/139 | identity swing **amplified** by a missing aggregation — `_contextImpl` counted `STEP_IN_PROCESS` *edges*, not processes. `_runImpactBFS`'s twin query already used `MIN(r.step)` for exactly this reason. `incoming`/`outgoing` looked pinned at 30 because they saturate their `LIMIT 30` caps |

## What changed

**Commit 1 — the ordering sweep.**

`ORDER BY n.id` on the resolver window. `n.id` is the PRIMARY KEY on every node table and the only viable key: `labels(n)[0]` comes back empty for `Class` nodes (already documented at `enrichCandidateLabels`), so ordering on the projected type would sort the *highest-priority* kind into the empty bucket.

The #480 Class/Constructor collapse probe was the same defect on a worse path — an unordered `LIMIT 1` over the whole candidate set that returns `kind: 'ok'`. When two same-named candidates both carried the label it picked one arbitrarily and answered *confidently wrong*, never reaching the scorer or the ambiguity report. It is also the only confident-resolution path a bare name can take: `scoreCandidate` tops out at 0.60 without a `file_path` hint and the confident gate needs 0.95.

Then the same defect class across the rest of the read path: the eight `LIMIT 30` context queries; `trace`'s per-depth cap (an unordered clip could return a path on one run and `no_path` on the next) and its first-writer-wins `parent` map; the impact BFS levels (no `LIMIT`, but three consumers read `impacted` positionally — including the enrichment that feeds the risk thresholds, so `risk` could move while `impactedCount` held); the FTS/BM25 lanes; cluster/process/wiki/manifest/augmentation queries; and eleven queries that already *had* an `ORDER BY` on a non-unique key (function names, BM25 score, confidence) and still shuffled at the cut. JS comparators feeding a `.slice()` gained a total tiebreak — `Array.sort` is stable, so ties preserved DB row order.

`ORDER BY` is not a perf concern here: `@ladybugdb/core` 0.18.3 ships a Top-K optimizer (`TopKOptimizer::visitLimitReplace`, `TopKBuffer`) — a bounded O(k) heap, not a full materialize — and the unlabeled `MATCH (n)` already scans 32 node tables.

**Commit 2 — two honesty defects the sweep left standing.**

*The cap masqueraded as the match count.* Every consumer reported `candidates.length` as the number of matching symbols, so 92 collisions read as "Found 20 symbols". Ordering made this **worse**: a stable wrong number reads as authoritative where a flaky one reads as broken. And `eval-server`'s formatter has been written for a true total since #2129, faithfully printing "20 symbols (showing 6)".

Fixed with a `COUNT(*)` over the same WHERE in the same `Promise.all` — the rule this file already states at `_explainImpl`: *"the page is LIMIT-capped, so its row count cannot stand in for the true total"* (#2084 review P2-4, the identical finding filed against `explain` two milestones ago). The repo's n+1 over-fetch convention was considered and rejected: it answers *is there more*, not *how much more*, and can only ever say "at least 21" — and once you add the lower-bound field, the formatter change, and the `MIGRATION.md` re-typing it is not even the smaller diff. A 21-row window would also reopen the Class-collapse probe on a row outside the reported candidate list. The window stays at 20, so nothing downstream of the resolver changes. The total is floored at the window length, so a failed COUNT degrades to the old number rather than to zero.

*Machine load could decide a group risk verdict.* `cross-impact`'s cross-repo fan-out was bounded only by a wall clock, and `mergeRisk` is monotone increasing in the traversed-crossing count (CRITICAL at three, HIGH on any ≥0.85-confidence crossing). An idle host traversed more and reported CRITICAL; a loaded host stopped at two and reported HIGH or lower — same graph, same arguments. The error direction is the unsafe one: truncation can only move risk **down**, and a fully truncated fan-out returns bare `localRisk`, making a symbol whose blast radius crosses a repo boundary indistinguishable from one with no cross-repo consumers at all.

The bound is now `MAX_NEIGHBOR_FANOUT = 50` over the now-totally-ordered neighbour list, mirroring `MAX_CROSSINGS_TO_TRY` in `cross-trace.ts` which bounds the same unit from the same bridge in the same order. The wall clock stays as a hang backstop — it is the load-bearing mitigation for CodeQL #184 / PR #1331 and is pinned by five tests — but when *it* fired the response now says `truncationReason: 'timeout'` instead of the generic `partial`. Manifest-only crossings stay exempt from the cap; capping them would regress #2784.

Risk is never clamped downward — that would be a downgrade in the only reading that matters (an agent told LOW proceeds; told CRITICAL it stops). A truncated fan-out carries `riskEpistemic: 'lower-bound'`, reusing the vocabulary single-repo `impact` already documents for `epistemic`, and the default (non-`--json`) group CLI now prints that the risk is a floor and names the crossings it did not traverse. Previously the terminal showed a bare `risk=` and the signal lived only in JSON nobody was reading.

## Verification

| check | result |
|---|---|
| `context get` ×20 on the 240k index | **20/20 identical** (pre-fix: 2 distinct in 14) |
| `impact resolve` ×20 | **20/20 identical** |
| 5 new resolver/count assertions run against the parent commit | **all 5 fail** |
| 5 new fan-out tests (no clock in any assertion) | pass |
| 62 affected test files | 1254 pass |
| `tsc --noEmit` | exit 0 |
| test-tree typecheck errors | 702 → **701** (one fewer than baseline) |

End-to-end on the real index: `"message": "Found 34 symbols matching 'get' (showing 20)"`, `"totalCandidates": 34` — 34 matches the ground-truth aggregate exactly.

## Tests

The gate asserts the **emitted SQL**, captured through the `executeParameterized` mock that `calltool-dispatch.test.ts` already installs — zero production refactor, and the only shape that fails deterministically. It pins both halves of the contract: the ordered window *and* a `LIMIT`-less COUNT. One invariant covers the whole `context` path in a single assertion (no multi-row `LIMIT` without an `ORDER BY`), so queries added to that path later are covered for free.

The integration case seeds 25 same-name Functions in **descending id order**, so the pre-fix window (`z25..z06`) and the post-fix window (`z01..z20`) are disjoint on 10 elements — it fails on the parent commit, verified, not assumed.

Deliberately **not** included: a run-N-times-and-compare test. At the measured ~5-8% flip rate it would pass by luck roughly half the time at N=8 — flaky by construction and worthless as a gate. A source-scanning lint guard was also skipped: `local-backend.ts` has 70 `LIMIT` occurrences against 12 `ORDER BY`, so it needs a hand-maintained allowlist that will rot, and it cannot distinguish a safe `LIMIT 1` from the unsafe one on the collapse path. The runtime call-capture gets the same guarantee with no allowlist.

Also fixed three `trace-bfs` mocks that returned bare arrays where the real `executeParameterized` returns a promise — `.catch()` chains were a `TypeError` against them, so the sync shape was never a faithful stand-in.

## Notes for the reviewer

- Ordering changes the **order** of `query`/`impact` output in some paths (previously arbitrary). No API shape changed; no benchmark baseline needed re-basing — `bench/impact-pdg/baselines.json` gates on `annotationFingerprint`/`f1Bands`/`idBridge`, and the PDG bridge runs without re-invoking the resolver.
- Suite failures that predate this branch and are unrelated: `pdg-mode-flip` (4, fails in isolation on clean `main`) and every `WorkerPoolInitializationError` file (all pass with `GITNEXUS_WORKER_READY_TIMEOUT_MS=60000` — resource pressure, not code). Both baselined by stashing `src`, rebuilding, and re-running the same files.


gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -3468,18 +3543,23 @@ export class LocalBackend {
       MATCH (caller)-[r:CodeRelation]->(n {id: $symId})
       WHERE r.type IN ['CALLS', 'IMPORTS', 'EXTENDS', 'IMPLEMENTS', 'USES', 'HAS_METHOD', 'HAS_PROPERTY', 'METHOD_OVERRIDES', 'OVERRIDES', 'METHOD_IMPLEMENTS', 'ACCESSES']
       RETURN r.type AS relType, caller.id AS uid, caller.name AS name, caller.filePath AS filePath, labels(caller)[0] AS kind
+      ORDER BY relType, uid
```
magyargergo: **P1** — `ORDER BY relType` as the leading key fills the 30-row window in alphabetical category order, so whole relation categories vanish in 100% of runs — measured on this index, `{ACCESSES:3, CALLS:35, USES:1, HAS_METHOD:1}` becomes `{ACCESSES:3, CALLS:27}`, silently dropping the symbol's owning class; use `ORDER BY uid, relType` here and at 3614/3627/3642/3698.

gitnexus/src/core/group/cross-impact.ts
```diff
@@ -448,7 +468,12 @@ export async function resolveBridgeNeighbors(
     const n = rowToNeighbor(raw);
     if (n) neighbors.push(n);
   }
-  neighbors.sort((a, b) => b.confidence - a.confidence);
+  neighbors.sort(
+    (a, b) =>
+      b.confidence - a.confidence ||
+      a.neighborRepo.localeCompare(b.neighborRepo) ||
+      a.neighborUid.localeCompare(b.neighborUid),
```
magyargergo: **P1** — The comparator omits `contractId`, which the dedup key at line 654 treats as part of the identity, so equal-confidence links to the same repo/uid still fall back to raw driver order at the 50-crossing cap — `cross-trace.ts:341` in this same PR includes it; append `|| a.contractId.localeCompare(b.contractId)`.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -3270,11 +3292,39 @@ export class LocalBackend {
...
+    // LIMIT-capped page's row count cannot stand in for the true total — the
+    // same rule `_explainImpl` and `_pdgQueryImpl` already follow (#2084 review
+    // P2-4). Without it every "Found N symbols matching 'x'" message and the
+    // `totalCandidates` field report the cap, so 92 collisions read as 20 —
+    // and now that the window is ordered, that undercount is stable, which
+    // makes it look authoritative rather than flaky.
+    const [rows, countRows] = await Promise.all([
+      executeParameterized(
+        repo.lbugPath,
+        `MATCH (n) ${whereClause} RETURN ${selectClause} ORDER BY n.id LIMIT 20`,
```
magyargergo: **P2** — `n.id` is `Label:filePath:name`, so this window is label-major then path-major, not neutral: `run` has 7 `Method:` matches and the ordered 20-row window contains 2, and `kind` cannot rescue them because it only scores (3180-3182) and never filters — push `kind` into the WHERE clause or widen the window now that `total` is known.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -6155,10 +6271,19 @@ export class LocalBackend {
...
+      // CHUNK_SIZE)` and feeds the risk thresholds; and byDepth pagination
+      // slices without re-sorting. Row order therefore leaked into `risk` and
+      // into which symbols the user sees.
       const query =
         direction === 'upstream'
-          ? `MATCH (caller)-[r:CodeRelation]->(n) WHERE n.id IN $frontierIds AND r.type IN $relTypes${confidenceFilter} RETURN n.id AS sourceId, caller.id AS id, caller.name AS name, labels(caller)[0] AS type, caller.filePath AS filePath, r.type AS relType, r.confidence AS confidence`
-          : `MATCH (n)-[r:CodeRelation]->(callee) WHERE n.id IN $frontierIds AND r.type IN $relTypes${confidenceFilter} RETURN n.id AS sourceId, callee.id AS id, callee.name AS name, labels(callee)[0] AS type, callee.filePath AS filePath, r.type AS relType, r.confidence AS confidence`;
+          ? `MATCH (caller)-[r:CodeRelation]->(n) WHERE n.id IN $frontierIds AND r.type IN $relTypes${confidenceFilter} RETURN n.id AS sourceId, caller.id AS id, caller.name AS name, labels(caller)[0] AS type, caller.filePath AS filePath, r.type AS relType, r.confidence AS confidence ORDER BY id, relType`
```
magyargergo: **P2** — `ORDER BY id, relType` does not pin `confidence` even though the comment above claims it does — 2181 of ~10020 `(id, relType)` groups in this index carry more than one distinct confidence, so the stamped value is still raw row order; use `ORDER BY id, relType, confidence DESC, sourceId`.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -3270,11 +3292,39 @@ export class LocalBackend {
...
+      executeParameterized(
+        repo.lbugPath,
+        `MATCH (n) ${whereClause} RETURN ${selectClause} ORDER BY n.id LIMIT 20`,
+        queryParams,
+      ),
+      executeParameterized(
+        repo.lbugPath,
+        `MATCH (n) ${whereClause} RETURN COUNT(*) AS total`,
+        queryParams,
+      ).catch(() => [] as any[]),
```
magyargergo: **P2** — When the COUNT leg fails, `totalMatches` silently falls back to the window length and the response ships `totalCandidates: 20` with no `candidatesTruncated` — byte-identical to a genuine 20-match result and indistinguishable from it, restoring the exact undercount this PR fixes; keep whether COUNT succeeded and mark it unknown.

gitnexus/src/cli/group.ts
```diff
@@ -250,19 +250,33 @@ export function registerGroupCommands(program: Command): void {
...
           if (summary) {
             const boundaryNote = boundaryOnly > 0 ? ` (${boundaryOnly} boundary-only)` : '';
             console.log(
               `  direct=${summary.direct ?? 0} processes=${summary.processes_affected ?? 0} cross=${summary.cross_repo_hits ?? 0}${boundaryNote}`,
             );
           }
+          if (riskFloor) {
+            const dropped = (raw as { truncatedRepos?: string[] })?.truncatedRepos ?? [];
+            console.log(
+              `  risk is a LOWER BOUND — fan-out stopped early, ${dropped.length} crossing(s) not traversed${dropped.length > 0 ? `: ${dropped.join(', ')}` : ''}`,
```
magyargergo: **P1** — Two defects in one line: `truncated` is also set by `localPartial` alone, so this prints "fan-out stopped early, 0 crossing(s) not traversed" when the fan-out completed and the local walk merely overflowed its chunk cap (`local-backend.ts:6541-6543`, routine for any symbol with >1000 impacted nodes) or threw; and `truncatedRepos` is deduped at cross-impact.ts:767, so `dropped.length` counts repos, not crossings (40 dropped crossings into 2 repos prints "2").

gitnexus/src/core/group/cross-impact.ts
```diff
@@ -730,8 +772,14 @@ export async function runGroupImpact(
       cross_repo_hits: cross.length,
     },
     risk: mergeRisk(localRisk, cross),
+    // The VALUE is never clamped down. `mergeRisk` is monotone increasing in
+    // the traversed-crossing count, so truncation can only under-report — and
+    // under-reporting a blast radius is the unsafe direction (an agent told
+    // LOW proceeds; told CRITICAL it stops). Marking the floor keeps the
+    // warning intact while making the incompleteness legible.
+    ...(truncated ? { riskEpistemic: 'lower-bound' as const } : {}),
```
magyargergo: **P2** — `riskEpistemic` is emitted only here, but `tools.ts:484` (added in this PR) promises it unconditionally on truncation — the `uids.length === 0` return at 591-610 sets `truncated: true` with a real (non-UNKNOWN) risk and no marker, and the local-timeout return at 535-553 does the same; derive it from `truncated` in one shared helper.

gitnexus/src/core/augmentation/engine.ts
```diff
@@ -193,6 +195,7 @@ export async function augment(pattern: string, cwd?: string): Promise<string> {
         MATCH (caller)-[:CodeRelation {type: 'CALLS'}]->(n)
         WHERE n.id IN [${idList}]
         RETURN n.id AS targetId, caller.name AS name
+        ORDER BY targetId, caller.id
```
magyargergo: **P2** — Target-major ordering lets one hot symbol consume the entire 15-row budget across the 5 batched targets — for `generateId` (74 callers) the ordered 15 are 13 `File:` nodes and 2 `Function:` nodes, so `augment` renders 13 bare filenames instead of calling symbols; use `ORDER BY caller.id, targetId` or slice per target in JS (same at 223).

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -3318,9 +3377,18 @@ export class LocalBackend {
...
+          // routine), an unordered LIMIT 1 picked one arbitrarily and this
+          // returned `kind: 'ok'` — a CONFIDENTLY WRONG answer that never
+          // reaches the scorer or the ambiguity report. It is also the only
+          // confident-resolution path a bare name can take: scoreCandidate
+          // tops out at 0.60 without a file_path hint, and the confident gate
+          // below needs >= 0.95.
           const labelRows = await executeParameterized(
             repo.lbugPath,
-            `MATCH (n:\`${label}\`) WHERE n.id IN $candidateIds RETURN n.id AS id LIMIT 1`,
+            `MATCH (n:\`${label}\`) WHERE n.id IN $candidateIds RETURN n.id AS id ORDER BY n.id LIMIT 1`,
```
magyargergo: **P2** — The ORDER BY makes this collapse repeatable but does not establish that the Class is unique — with two same-named classes the branch below still returns `kind: 'ok'` and `context`/`impact` analyse the wrong file confidently, exactly as the comment you added at 3380-3386 describes; gate the collapse on `outcome.total` not exceeding the window, or fall through to ambiguity.

magyargergo (commented):
# Two-engine review digest — PR #2796 (`repro/2787`)

**Method.** Two independent engines ran the same two review contracts over `911151e23…239e43f02` (18 files, +743/−128). **Engine A — Claude:** GitNexus PR swarm (facts/hygiene/risk/test-CI/security/docs) + 9 `ce-code-review` personas. **Engine B — Codex `gpt-5.6-sol` @ xhigh:** the same swarm contract, the same `ce-code-review` contract, plus a native adversarial pass — **all three Codex legs produced real findings** (no dead leg). A synthesis pass then re-verified every surviving claim against the worktree; six were re-executed against the live LadybugDB index (read-only) and are tagged `[reproduced]`.

---

## 1. What is solid

The core thesis is right and the implementation of it is careful. Both engines independently confirmed:

- **`n.id` is a valid total order on every node table** — it is the PRIMARY KEY on all 32 tables (`CODE_ELEMENT_BASE` + the 14 language-specific tables), non-null everywhere. No ORDER BY key added by this PR can bind-fail.
- **The sweep is essentially complete.** Both engines scanned every remaining `LIMIT` in `local-backend.ts`, `pdg-impact.ts`, `graph-queries.ts`, `engine.ts`, `server/api.ts`. The residue is `LIMIT 1` singleton probes anchored on a unique id or an existence check, plus `check()`'s `LIMIT 100_001` (which errors rather than truncating). One real miss was found (`affected_processes`, §5).
- **`riskEpistemic: 'lower-bound'` is the right primitive and the right vocabulary.** It predates this PR (#2744/PR #2747), `GroupImpactTruncationReason` already contained `'timeout'` on main, and the field is strictly additive — no consumer breaks.
- **The manifest-only exemption on the new cap is correct** and preserves #2784. Both the clock and the count carry `!manifestOnly`.
- **`attemptedFanouts` is not off by one.** The increment sits at `cross-impact.ts:719`, immediately before the `await` at 720; the guard is `>= 50`. Exactly 50 `impactByUid` calls issue; the 51st is truncated. The new test pins this.
- **`mergeRisk` is monotone increasing** in traversed-crossing count — walked branch-by-branch by both engines — so `'lower-bound'` is semantically sound wherever it does fire.
- **CI is green.** All 40+ checks pass, including all three Windows platform-sensitive shards and all three macOS shards. `mergeable: MERGEABLE`.

### Hypotheses probed and REFUTED — credit where due

These were each constructed as concrete attacks and then falsified. They are **not** findings and are listed so the work is visible:

| Hypothesis | Why it was dropped |
|---|---|
| Cypher injection via the new `ORDER BY` clauses | Every added clause is a static literal over hardcoded aliases. No user value reaches an ORDER BY. (Engine A security lane, Engine B swarm lane 5.) |
| Break-out through `${escaped}` / `${patternFirstWord}` / `${fileList}` | Pre-existing interpolation, unchanged by this PR; quote-escaping blocks direct break-out. Not introduced or worsened here. |
| `ORDER BY` illegal after `DISTINCT`, or on an aggregate/implicit-grouping projection | All shapes bound against the live engine: `RETURN DISTINCT … ORDER BY <alias>`, `MIN(r.step) … ORDER BY pid`, `COUNT(DISTINCT …) AS hits ORDER BY hits DESC, name`. All parse and execute. |
| `MIN(r.step)` breaks implicit grouping / changes row count or column order | Executed both forms on a real 49-process symbol: identical 49 rows, identical column order, `r[0..3]` tuple fallbacks stay correct. Steps are 1-indexed, so the `r.step \|\| r[2]` falsy-zero trap does not fire. |
| `COUNT(*)` over an unlabeled `MATCH` returns one row per node table | Verified: single row, single correct total (42 for `get`, 53 for the qualified form). |
| BigInt / tuple-mode coercion breaks `totalMatches` | `Number(92n) === 92`; the `?.[0]` tuple fallback and the `Math.max` floor cover every observed shape. |
| `mergeRisk` non-monotone, so truncation could *over*-report | All branches walked; adding a crossing can only move risk up or leave it. |
| `ORDER BY score DESC, node.id` bind-fails after `CALL QUERY_FTS_INDEX` | Executed in production shape against both `function_fts` and `file_fts`. Works. |
| `BasicBlock` poisons the unlabeled resolver window (`B` sorts first) | `BasicBlock` has no `name` property — binder exception, so it can never match the WHERE clause. |
| The `async` mock-shape change in `trace-bfs.test.ts` breaks other suites | Every remaining `executeParameterized` mock in the repo is already async. No collateral breakage. |
| Schema-version / `INCREMENTAL` constant silently bumped (the recurring merge hazard) | No schema, incremental, bridge, or relation-schema constant touched. |
| Snapshot skew between the COUNT leg and the window leg | Same `Promise.all`, same read-only pool, identical WHERE. No writer can interleave on an MCP read path. |

### Two lane claims this synthesis **rejects**

- **`vitest.config.ts` / `lbug-db` registration is NOT a merge blocker and was NOT introduced here.** Engine A's test-CI lane graded this **P0, "Windows CI will fail"**. Verified: `git show 911151e23:gitnexus/vitest.config.ts` does **not** list `test/integration/impact-ambiguous-blast-radius.test.ts` either, the file already existed on main (207 lines) and already imported `withTestLbugDB` there (base line 19), and **this PR does not touch `vitest.config.ts` at all** — re-verified directly: `git show 911151e23:gitnexus/test/integration/impact-ambiguous-blast-radius.test.ts` returns the full base file, and `git show 911151e23:gitnexus/vitest.config.ts | grep -c impact-ambiguous-blast-radius` returns 0. Empirically, all three Windows platform-sensitive shards passed on this head. Engine B graded the same observation correctly as inherited misrouting. **Re-graded: pre-existing P3 hygiene note** (§5), not a blocker, not inline.
- **The conditional inside the new Vitest mock is not a violation.** Engine B's ce leg called `cross-impact-fanout-cap.test.ts:171` a P2 ("replace the conditional Vitest mock with deterministic call sequencing"). The repo's rule is *no conditional **assertions***; mock **routing** `if`s are an established pattern — `impact-pdg-ascent-note.test.ts` has 7 on main, `trace-bfs.test.ts` 17, `calltool-dispatch.test.ts` 21, all pre-PR. The new test's assertions are `expect(result).toMatchObject({…})`, exactly the preferred form. Engine A's testing lane reached the same conclusion. **Dropped as a false positive.**

---

## 2. Headline

**The PR makes `context()` strictly worse for exactly the symbols people call `context()` on.** `ORDER BY relType, uid LIMIT 30` fills the window in *alphabetical category order*, so any relation category whose alphabetical predecessors already total ≥ 30 is dropped in **100 % of runs**, silently — `categorize()` emits the buckets verbatim with no truncation flag. Before the PR the loss was possible but arbitrary; now it is certain. Reproduced on this repo's own index (§3.1). The fix is one token.

Second: **the PR's central determinism guarantee is incomplete at the exact boundary it introduces.** The new `MAX_NEIGHBOR_FANOUT = 50` cutoff sorts crossings by `(confidence, repo, uid)` — but the dedup key three lines later is `(repo, uid, contractId)`. Tied rows fall back to raw driver order, so *which* crossings survive the cap is still process-varying. `cross-trace.ts`, changed in the same commit, gets this right. Both Codex legs and two Claude lanes converged on it independently.

---

## 3. Inline comments

### 3.1 `gitnexus/src/mcp/local/local-backend.ts:3546` — **P1** `[reproduced]`
**`ORDER BY relType` as the leading key deterministically deletes whole relation categories from `context()`.**

- **Trigger:** any symbol whose incoming/outgoing refs exceed 30 across ≥ 3 categories.
- **Evidence (executed read-only against `/workspace/.gitnexus/lbug`):** for `Method:…local-backend.ts:LocalBackend.query#2` the full edge set is `{ACCESSES: 3, CALLS: 35, USES: 1, HAS_METHOD: 1}` (40 rows). Running the production query verbatim with `ORDER BY relType, uid LIMIT 30` returns `{ACCESSES: 3, CALLS: 27}`. **`HAS_METHOD` — the edge that names the method's owning class — and `USES` are gone, every run.** `categorize()` (`:3740-3754`) buckets by `relType` and the result is emitted verbatim at `:3845-3846` with no truncation flag, so the caller cannot tell.
- Note the inconsistency inside the same `Promise.all`: the sibling `ADVISED_BY` query (`:3562`) got `ORDER BY uid` with no `relType` prefix and has no starvation problem, because it holds a single type. Only the multi-type windows got the harmful prefix.
- **Fix:** `ORDER BY uid, relType` at `:3546`, `:3614`, `:3627`, `:3642`, `:3698`. Still a total order (`uid` is the node primary key), still deterministic, and spreads the 30 rows across categories the way the unordered scan incidentally did.
- **Corroboration:** Engine A ce-correctness F1 (graded P2). Single-engine, but **reproduced end-to-end here** and raised to P1: the loss went from probable to certain, it is silent, and the fix is a token swap.

### 3.2 `gitnexus/src/core/group/cross-impact.ts:475` — **P1** `[code-read]`
**The fan-out comparator is not a total order: it omits `contractId`, which the dedup key three lines later proves is part of the identity.**

- **Trigger:** > 50 eligible crossings where at least two share `(confidence, neighborRepo, neighborUid)` but differ in `contractId` — realistic whenever one consumer function calls two endpoints of the same provider service (same repo, same uid, both `exact`/1.0).
- **Evidence:** comparator at `:471-476` stops at `neighborUid`; the `seen` key at `:654` is `` `${n.neighborRepo}\0${n.neighborUid}\0${n.contractId}` ``. Both rows survive `seen` and compare `0` in the sort; `Array.prototype.sort` is stable, so their relative order is raw bridge row order — the precise nondeterminism this PR exists to remove. At the cap boundary that decides which crossing is traversed and which lands in `truncatedRepos`; below the cap it decides `cross[]` ordering in the JSON. **`cross-trace.ts:341-344`, added in the same PR, *does* include `contractId`** — the two paths disagree with each other.
- **Fix:** append `|| a.contractId.localeCompare(b.contractId)`. Add a > 50-row cap-boundary test with identical confidence/repo/uid and distinct contract ids, run over permuted input.
- **Corroboration:** **Both engines.** Codex-ce #1 (P1, conf 100), Codex-swarm 3.2 (P2, blocks merge), Codex-adversarial (medium); Claude ce-correctness F6, Claude ce-adversarial F2. Strongest cross-engine convergence in the review.

### 3.3 `gitnexus/src/mcp/local/local-backend.ts:3319` — **P2** `[reproduced]`
**`ORDER BY n.id` is a label-major, then path-major key — the 20-row window is a biased prefix, not a neutral sample, and no caller hint can compensate.**

- **Trigger:** any bare-name `context`/`impact`/`trace` call on a name with > 20 collisions.
- **Evidence:** node ids are `Label:filePath:qualifiedName`. Executed `MATCH (n) WHERE n.name='run' RETURN n.id ORDER BY n.id LIMIT 25`: rows 1-18 are all `Function:` (bench/ then src/ then test/), rows 19+ are `Method:`. There are **7 `Method:run` in the index; the 20-row window contains 2.** `Class < Const < Constructor < Function < Interface < Method` is uncorrelated with `scoreCandidate`'s own priority table (`Class 5, Interface 4, Function 3, Method 2` — `:3184-3191`), so `Interface` (priority 4) sorts *after* `Function` (priority 3) and `Const` sorts before both. The `kind` hint never reaches the query: `whereClause` is built from `name` and `file_path` only (`:3280-3288`), and `hints.kind` is a `+0.2` scoring term (`:3180-3182`). So `impact({target:'run', kind:'Method'})` cannot see 5 of the 7 Methods, on any host, ever.
- Downstream, `_impactImpl`'s ambiguous branch probes `candidates.slice(0, 6)` and hoists `maxRisk`/`maxImpactedCount` — the largest-blast-radius candidate can now be *stably* outside the window, a reproducible under-report of the signal the PR exists to stabilise.
- The comment at `:3298-3310` argues `n.id` is "the only viable key" because `labels(n)[0]` is empty for Class. That is true, but `n.id` **is** ordering on the label, with extra steps — the comment should say so.
- **Fix (pick one):** push `kind` into the WHERE clause as a filter (`n.id STARTS WITH $kindPrefix`, same label-prefix invariant the ORDER BY already relies on), keeping `n.id` as tiebreak; or, now that `total` is known, widen to `LIMIT 200` before scoring and slice after (the resolver re-sorts by score anyway); or at minimum stop claiming the key is neutral.
- **Corroboration:** Engine A only — ce-correctness F2, ce-adversarial F1, swarm risk-architect Finding 2 (three lanes, one engine) — but **reproduced end-to-end here**. Engine B did not probe the id shape; this is its blind spot in this review.

### 3.4 `gitnexus/src/mcp/local/local-backend.ts:6285` — **P2** `[reproduced]`
**`ORDER BY id, relType` pins `relationType` but leaves `confidence` unpinned, though the comment above claims both.**

- **Trigger:** a diamond — the same impacted node reached from two frontier nodes via the same edge type with different `r.confidence`.
- **Evidence:** the comment at `:6278-6280` states "the relationType/**confidence** stamped on a node reached by more than one edge is the first row's". `(id, relType)` is not unique. Measured on this index (upstream direction, this query's exact relation-type set): **2181 of ~10 020 `(id, relType)` groups carry more than one distinct confidence.** Within a tie group the surviving row — and the `confidence` copied out at `:6329-6338` — is still raw DB row order. 0.7 vs 0.85 straddles the interpretation boundary the tool description itself publishes (`tools.ts:482`: "<0.8 = fuzzy match"), so a caller reading per-node confidence in `byDepth` still gets a run-varying verdict.
- **Fix:** `ORDER BY id, relType, confidence DESC, sourceId` — completes the key and makes the retained edge the strongest, which is the safer default for a blast-radius tool. Add a diamond regression asserting the exact stamped relation and confidence.
- **Corroboration:** **Both engines.** Codex-swarm 3.3 (P2, blocks merge); Claude ce-correctness F7, Claude ce-adversarial F9. Tie count independently re-measured here.

### 3.5 `gitnexus/src/mcp/local/local-backend.ts:3326` — **P2** `[code-read]`
**`.catch(() => [])` on the COUNT leg silently restores the exact bug this PR fixes, and now ships the reverted number under a field the tool description promises is exact.**

- **Trigger:** the 20-row window succeeds, the COUNT query fails or returns an unreadable shape.
- **Evidence:** `countRows` becomes `[]`; `totalMatches` falls back through `Math.max(…, normalized.length)` (`:3342-3349`) to `20`. The response then reads `Found 20 symbols matching 'x'` with `totalCandidates: 20` and **no** `candidatesTruncated` — byte-identical to a genuine 20-match result, and indistinguishable from it by any consumer. The `Math.max` floor is correct and the degradation matches pre-PR behaviour, so this is not a *safety* regression (Engine A's security lane is right about that) — but the PR newly promises this number is the true total, and on this path it silently is not.
- **Fix:** keep whether COUNT succeeded. On failure emit `countUnavailable: true` (or `totalCandidates` as an explicit lower bound) and log it. Add a test where the window resolves and the COUNT leg rejects.
- **Corroboration:** **Both engines.** Codex-ce #3 (P2, conf 100), Codex-swarm 3.4, Codex-adversarial (medium); Claude ce-adversarial F5, ce-reliability F4, swarm risk-architect (flagged as unverified suspicion). Engine A's security lane dissents ("degrades safely, no change required") — that answers a different question (new attack surface: none) and does not touch the truthfulness gap.

### 3.6 `gitnexus/src/cli/group.ts:277` — **P1** `[code-read]`
**Two independent defects in one line: the message states a cause that may be false, and counts repos as crossings.**

- **Trigger A (wrong cause):** `truncated = truncatedRepos.length > 0 || localPartial` (`cross-impact.ts:758-759`). The clock case returned early at `:535-553`, so `local.partial` here means either a local BFS depth query **threw** OR — far more commonly — the impact **chunk cap overflowed**: `local-backend.ts:6541-6543` sets `traversalComplete = false` whenever `chunksProcessed * CHUNK_SIZE < impacted.length`, and since `MAX_CHUNKS * CHUNK_SIZE = 1000` while `GROUP_LOCAL_PHASE_LIMIT = 10_000`, **any hub symbol with more than 1000 locally-impacted nodes takes this path on every run.** Then: `localPartial = true` → `truncatedRepos = []` → `riskEpistemic = 'lower-bound'` → the CLI prints `risk is a LOWER BOUND — fan-out stopped early, 0 crossing(s) not traversed`. Both halves are false — the fan-out completed — and the two clauses contradict each other in the same sentence. This is the routine case for exactly the high-fan-in symbols group impact is most used on, not a rare error path.
- **Trigger B (wrong unit):** `truncatedRepos` is pushed once per dropped **crossing** (`:662, 669, 677, 734`) then deduped to distinct repo keys at `:767` (`[...new Set(...)]`). 40 dropped crossings into 2 repos prints `2 crossing(s) not traversed`. That is a cap masquerading as a total — the same defect class this PR fixes for `totalCandidates` one file over.
- **Fix:** branch the message on the real cause (`truncatedRepos.length > 0` → "fan-out stopped early"; else → "the local impact walk did not complete"), and either say "repo(s)" or carry a pre-dedup crossing counter. Both signals are already in the payload.
- **Corroboration:** **Both engines.** Codex-ce #2 (P2, conf 100), Codex-swarm 3.5; Claude ce-correctness F3 + F4, Claude ce-adversarial F3.

### 3.7 `gitnexus/src/core/group/cross-impact.ts:780` — **P2** `[code-read]`
**`riskEpistemic` is emitted on only one of the truncated return paths, contradicting the unconditional contract this PR adds to `tools.ts`.**

- **Evidence:** `tools.ts:484` (added here) states "when it stops early the response carries truncated:true, truncatedRepos, **and** riskEpistemic:'lower-bound'". Only the terminal result at `:780` does. Two earlier returns set `truncated: true` and never do:
  1. **`uids.length === 0`** (`:591-610`) — `truncated: Boolean(local.partial)` with `risk: String(local.risk ?? 'LOW')`, i.e. a **real** risk value with no floor marker, under the same `local.partial` condition that *does* produce the marker twenty lines further down. Trigger: local BFS query error + a target whose impacted set yields no contract-eligible uids (routine for a private helper).
  2. **local timeout** (`:535-553`) — `truncated: true`, `truncationReason: 'timeout'`, no marker. Lower stakes because `risk: 'UNKNOWN'` already warns, but it is still off-contract, and `cross-impact.test.ts:143` pins that exact shape.
- Consequence: an agent following the new tool description keys on `riskEpistemic` and reads a truncated result as complete. The CLI has the same hole — `riskFloor` is `''` on both paths, so `risk=MEDIUM` prints identically to a complete run.
- **Fix:** derive `riskEpistemic` from `truncated` in one shared helper so the invariant cannot drift again. A table-driven test over "every truncated `GroupImpactResult` carries `riskEpistemic`" would pin it.
- **Corroboration:** Engine A only — ce-correctness F5, ce-adversarial F4, ce-reliability F1 + F2 (four lanes, one engine). Re-verified here by reading both return sites.

### 3.8 `gitnexus/src/core/augmentation/engine.ts:198` — **P2** `[reproduced]`
**`ORDER BY targetId, caller.id LIMIT 15` lets one hot symbol consume the whole budget, and the `File:` label prefix takes it first.**

- **Trigger:** `augment()` on any pattern where one of the ≤ 5 batched symbols has ≥ 15 callers.
- **Evidence:** `idList` holds up to 5 symbols (`:180-186`) and rows are fanned back into a **per-target** map (`callersMap`, `:205-212`). Target-major ordering means all 15 rows can come from the alphabetically-first target; the other four render with no callers at all. Executed the production query for `Function:gitnexus/src/lib/utils.ts:generateId` (74 callers): the ordered 15 are **13 `File:` nodes and 2 `Function:` nodes**. `augment` renders `caller.name`, and a `File` node's name is a bare filename — so the "called by" block went from a mix of real calling symbols to 13 filenames. `File:` sorts before `Function:`/`Method:`, so this is the guaranteed outcome, not bad luck. Same root cause as §3.1 and §3.3: a *bucketing* column in the leading sort position turns a fair sample into a single-bucket prefix.
- **Fix:** `ORDER BY caller.id, targetId` (not target-major), or raise the cap to `5 * perSymbolCap` and slice per target in JS. Same at `:223`.
- **Corroboration:** Engine A only — ce-correctness F8, ce-adversarial F1 Trigger 2. Reproduced here.

### 3.9 `gitnexus/src/mcp/local/local-backend.ts:3391` — **P2** `[code-read]` — *engines disagree on grading; facts agreed*
**The Class/Interface collapse probe now returns a stable wrong answer instead of a random one, and the comment added right above it says so.**

- **Trigger:** two same-named classes/interfaces in different files, plus any Constructor/value candidate, so `ambiguousType` is true.
- **Evidence:** the probe is `MATCH (n:\`${label}\`) WHERE n.id IN $candidateIds RETURN n.id AS id ORDER BY n.id LIMIT 1`, and the branch immediately below returns `kind: 'ok'` — bypassing the scorer and the ambiguity report entirely. The comment added by this PR at `:3380-3386` states plainly that an unordered `LIMIT 1` "returned `kind: 'ok'` — a CONFIDENTLY WRONG answer". The ORDER BY makes that answer repeatable; it does not establish that the Class is unique. `context` and `impact` then analyse the wrong file with full confidence.
- **Adjudication:** the defect is **pre-existing** — Engine A's risk-architect is right that the PR did not create it. But the PR touched this exact line, documented the failure in a comment, and shipped it. Given the repo's fix-in-PR-never-defer norm, and that `totalMatches` is now available (`total > candidateIds.length` is a sufficient condition to refuse the collapse), leaving it is a choice worth making explicitly rather than implicitly.
- **Fix:** collapse only when exactly one candidate carries the preferred label *and* `outcome.total` does not exceed the window; otherwise fall through to ambiguity scoring. Add a two-Class-plus-Constructor regression.
- **Corroboration:** **Both engines saw it, graded it differently.** Codex-swarm 3.1 (P1, blocks merge, conf 0.99), Codex-adversarial (high); Claude ce-testing F1 (P1, as an untested branch); Claude risk-architect (explicitly "pre-existing, made deterministic — not introduced"). Posted at P2 as the honest middle.

---

## 4. Body-only items (not inline)

**Real miss in the sweep — `impact`'s own `affected_processes`.** `local-backend.ts:6441-6452` has no `ORDER BY`, and the comparator this PR *did* touch at `:6549-6556` sorts on `(total_hits, filePath, name)`, which is not unique across distinct `entryPointId`s. `entryPointMap` is a `Map`, so `Array.from(values())` is insertion order = raw row order. Two distinct entry-points in this index genuinely collide on `(name, filePath)`: `useSigma.runCirclesLayout…step` and `useSigma.runTreeLayout…step`, both `name='step'`, both in `gitnexus-web/src/hooks/useSigma.ts`. Equal `total_hits` is the norm. Fix: add `ORDER BY pId` to the chunk query and `|| a.entryPointId.localeCompare(b.entryPointId)` to the comparator. *(Engine A ce-adversarial F6, P2. Not inline: the query line is not in the diff.)*

**`trace` BFS truncation is now deterministically source-biased.** `ORDER BY sourceId, id, edgeType LIMIT ${rowCap}` (`:5218-5222`) with `rowCap = min(frontier.length * 200, 5000)`: once the frontier exceeds 25 nodes the cap pins at 5000 and the survivors are a strict prefix in frontier-id order, so lexicographically-late frontier nodes contribute zero successors on every run. A pair that used to be found intermittently now returns `no_path` every time. The `truncated` flag at `:5227` is the saving grace. Consider chunking the frontier so each chunk gets its own budget. *(Engine A ce-adversarial F7, P2.)*

**`cross-trace.ts:341-344 / 383-386` omits `providerUid`** even though the Cypher `ORDER BY` at `:257`/`:279` includes it — the defensive JS re-sort is weaker than the order it exists to reproduce. Minor, but it is the mirror image of §3.2. *(Both engines: Codex-swarm 3.2, Claude ce-adversarial F8.)*

**Deterministic-but-clustered truncation appears in several lower-stakes queries**, same mechanism as §3.1: `graph-queries.ts:187`/`:199` (`ORDER BY fromFile, … LIMIT 30` — all 30 inter-module edges now come from the alphabetically-first file) and `local-backend.ts:4513`/`:7718` (`ORDER BY filePath, name, type LIMIT 30` for cluster members). Smaller blast radius; worth a pass if §3.1 is fixed.

**Performance — unmeasured, flagged by both engines.** The resolver's hot path went from one short-circuitable scan to two full scans: `n.name` is not indexed (Kuzu indexes primary keys only), so the base `LIMIT 20` could stop at the 20th match, while `ORDER BY` forces full materialisation and the added `COUNT(*)` has no `LIMIT` at all. Separately, the impact BFS per-depth query at `:6285` has **no LIMIT**, so the new `ORDER BY` forces an in-DB sort of an arbitrarily large intermediate result on every depth pass. Invisible on the 20 k-symbol GitNexus index; a real question at kernel scale (#2649). CI's general benchmarks passed but neither engine found a resolver-specific large-index latency measurement. Recommend one before this reaches hosted environments.

**`localeCompare` in the new JS tiebreaks is locale-sensitive** (`cross-impact.ts:474-476`, `cross-trace.ts:341-344`/`383-386`, `local-backend.ts:2234, 3020, 6552-6556, 6639`). It resolves against the host's default ICU locale — for ASCII ids this is stable in practice, and the idiom predates the PR (`local-backend.ts:3440`), so it is **not** a regression. But a PR whose thesis is cross-host reproducibility is better served by `<`/`>` code-unit comparison, which also matches the DB's binary `ORDER BY`. *(Engine A project-standards graded this P1; that over-states it given the pre-existing precedent. P3.)*

**Docs.** `MIGRATION.md` documents the ambiguous-response contract (PR #888) and is not updated for `totalCandidates` / `candidatesTruncated`, and `impact`'s `totalCandidates` **changed value semantics** (window length → true COUNT). Additive and arguably a bug fix, but the semantic change is worth a line. Five tool descriptions also do not mention the new ambiguous fields. The `MAX_NEIGHBOR_FANOUT` JSDoc claims parity with `MAX_CROSSINGS_TO_TRY`, but that constant bounded *trace*, not impact — group impact had no numeric crossing cap before this PR. *(P3 each.)*

**Comment accuracy.** The `MIN(r.step)` rationale at `:3716-3723` asserts the un-aggregated form "emitted one row per edge and `processes` reported an edge count". A full scan of this index shows the maximum `STEP_IN_PROCESS` edge count for any (symbol, process) pair is **1** — the aggregation is behaviour-preserving here. Harmless, but the stated motivation will mislead the next reader.

**Pre-existing test-wiring gap (re-graded from P0).** `test/integration/impact-ambiguous-blast-radius.test.ts` uses `withTestLbugDB` but is in neither `vitest.config.ts`'s `lbug-db` include list nor `LBUG_NATIVE` — **on main as well as here**, and this PR does not touch `vitest.config.ts`. All Windows and macOS shards pass on this head, so it is not a live breakage. It is still worth registering, because the PR's new block makes this file the primary real-engine verification for the fix and it currently runs only in the parallel Ubuntu project. **P3, pre-existing.**

### Test gaps

1. **Nothing pins the relation-type spread of the 30-row `context` window (§3.1).** A fixture with 35 `CALLS` + 1 `HAS_METHOD` asserting `incoming.has_method` is present would fail here and pass on main. The new `no multi-row LIMIT … is left unordered` test only asserts an `ORDER BY` *exists* — the one assertion that structurally cannot distinguish a good order key from a harmful one. It also runs only `context` on a single `Class` row, so the Function/Method branches never execute.
2. **The `manifestOnly` exemption on the new cap is untested in either direction** — no test mints a `manifest::` neighbour uid. A regression that made manifest crossings consume fan-out budget (re-breaking #2784) would pass green. This is the riskiest new line in `cross-impact.ts`.
3. **`fanoutTimedOut` / `truncationReason: 'timeout'` from the fan-out leg is dead code from the suite's perspective.** The only `'timeout'` assertion (`cross-impact.test.ts:143`) hits the *local* early return, a different line.
4. **No test covers `truncated` caused solely by `localPartial`** — §3.6 Trigger A and §3.7 both live in that hole.
5. **No test covers the COUNT-leg failure fallback** (§3.5) or the `(id, relType)` confidence tie (§3.4).
6. **`cross-impact-fanout-cap.test.ts` cannot fail for the intended reason on base** — `MAX_NEIGHBOR_FANOUT` does not exist there, so the import yields `undefined` and `REPO_COUNT` becomes `NaN`. It fails as an import-shape error, not a missing-cap error. Unavoidable for a new export; noted so nobody cites it as base-verified.

---

## 5. CI

All checks now green — the three that were pending at the previous poll (coverage shards, benchmarks, Windows platform-sensitive) have completed and passed. 40+ checks pass including all three Windows and all three macOS platform-sensitive shards, CodeQL, gitleaks, typecheck, lint, format, packaged-install smoke, and tree-sitter ABI on all three OSes. `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE` (attributable to Vercel only). **Vercel is red with "Authorization required to deploy" — a fork-context auth failure, not a code defect, and not a required check.** Not actionable.

## 6. Coverage limits

- No test suite was executed by this synthesis (the review worktree's `node_modules` is a symlink to the main checkout). CI's own green run is the evidence for test health; the `[reproduced]` tags refer to read-only Cypher probes against `/workspace/.gitnexus/lbug`, not to test runs.
- The measurements in §3.1, §3.3, §3.4 and §3.8 are from **this repository's own 20 k-symbol index**. The *mechanisms* are index-independent; the specific counts are not.
- The index has no PDG/taint layer built, so the absence of taint findings is not evidence of safety.
- Engine B's `ce` leg lost its cross-model sub-pass (Python blocked in its sandbox) and reported adversarial coverage as degraded rather than claiming it — its native adversarial leg ran separately and did produce findings.
- Kernel-scale performance (§4) is reasoned about, not measured, by either engine.

---

*Automated two-engine review digest (Claude swarm + ce; Codex gpt-5.6-sol/xhigh swarm + ce + adversarial). Every inline finding was re-verified against the worktree before posting, and six were re-executed against the live index — but verify before acting.*

---

## 7. Final verdict

**Branch hygiene:** `clean feature/fix PR` — linear, 2 commits, no merge-from-main, no unrelated churn, no schema/INCREMENTAL constant touched. The 10-file spread is one coherent defect class; no split or rebase required.

**Merge state:** `review blocked` — CI is fully green and `mergeable: MERGEABLE`, but the review verdict overrides that.

**Final verdict: not production-ready.**

The central thesis is right and most of the sweep is correct, but three P1s must be fixed before merge. Two of them are regressions this PR introduces rather than pre-existing debt: `context()` now fills its 30-row window in alphabetical relation-category order, so whole categories are dropped deterministically for any high-fan-in symbol (previously the loss was possible but arbitrary — this is strictly worse); and the group CLI's new lower-bound line states a cause that is false in the routine hub-symbol case. The third — the cross-impact fan-out comparator omitting `contractId` — leaves the PR's own headline determinism guarantee incomplete at exactly the boundary the PR introduces, and the sibling file changed in the same commit gets it right. None requires a rebase or a split; all are small, local edits. The remaining P2s (confidence unpinned in the BFS stamp, COUNT-leg silent degradation, `riskEpistemic` missing on two return paths, target-major ordering in `augment`, the still-confident Class collapse) are in scope for this PR under the repo's fix-in-PR norm.

PR #2803, opened 2026-08-02
fix(ai-context): document CLI graph fallbacks

azizur100389:
## Summary
- Make generated GitNexus AI guidance transport-agnostic by pairing mandatory MCP `impact` / `detect_changes` checks with repo-scoped CLI fallbacks.
- Keep fallback guidance aligned across `AGENTS.md`, `CLAUDE.md`, and distributed `gitnexus-impact-analysis` skill copies.
- Add regression coverage for fallback command shape, custom runners, PDG mode, repo selection, branch quoting, and shipped-skill drift.

Closes #2671.

## Test plan
- `npm test -- --run test/unit/ai-context.test.ts test/unit/shipped-skills-sync.test.ts test/unit/resolve-invocation.test.ts test/unit/runner-exec-tail.test.ts`
- `npx tsc --noEmit`
- `npm run build`
- `node gitnexus/dist/cli/index.js analyze --pdg`
- `detect_changes(scope: all)` via MCP: low risk, 26 changed symbols, 0 affected processes
- `node gitnexus/dist/cli/index.js detect-changes --scope all --repo .`

## Review notes
- Ran local review lenses for production risk, test/CI coverage, and security/trust-boundary review.
- Full `npm test` was attempted locally but stopped after more than 10 minutes when the Windows `lbug-db` lane hit an unrelated native FTS/VECTOR extension failure in `lbug-delete-nodes-for-files.test.ts`; the targeted changed-surface tests and build/typecheck passed.

PR #2804, opened 2026-08-03
feat(render): add one-click deploy to render support

Ho1yShif:
## feat(render): add one-click deploy to render support

> [!IMPORTANT]
> **Updated after review.** Everything below describes the original design; [this comment](https://github.com/abhigyanpatwari/GitNexus/pull/2804#issuecomment-5198553774) has the current writeup and the deploy receipts.
>
> `gitnexus-server` is now a private service with no public URL, and `gitnexus-web` reverse-proxies `/api/*` to it over Render's private network behind a generated Bearer token. `GITNEXUS_PUBLIC_ORIGIN`, `GITNEXUS_BACKEND_URL`, `AZURE_DEVOPS_URL`, and `AZURE_DEVOPS_PAT` are gone, and `autoDeploy: true` became `autoDeployTrigger: 'off'`. The `gitnexus/src` changes moved to #2820, which merged, so this branch touches no server source. What's new here: the `docker-server.mjs` proxy and its tests, a token panel in the web UI, `render-blueprint.test.ts`, and a SECURITY.md section.

Adds a Render Blueprint so GitNexus can be deployed in one click: the API server with a persistent disk for indexes and cloned repos, plus the web UI wired to it. No API key is required and nothing secret is committed. The optional Azure DevOps credentials are `sync: false`, so Render prompts for them at deploy time.

### What's included

- `render.yaml` — two Docker web services under one project. `gitnexus-server` (`standard`, 2 GB) with a 10 GB disk at `/data/gitnexus`, matching `GITNEXUS_HOME` in `Dockerfile.cli`; `gitnexus-web` (`starter`) serving the UI. Each service gets the other's public URL through `fromService`, so there is nothing to paste after the deploy. Health checks on `/api/health` and `/`.
- `gitnexus/src/server/middleware.ts`, `api.ts` — two new env vars, `GITNEXUS_PUBLIC_ORIGIN` and `GITNEXUS_TRUST_PROXY`. Both default to current behavior when unset.
- `README.md` — a Deploy to Render button with a short section on cost, the memory-bound sizing knob, and the security posture, plus a row per new variable in the Environment variables table.
- `gitnexus/.env.example` — both variables, marked hosted-deploy only.
- `CONTRIBUTING.md` — one checklist line making the README table the documented home for every new `GITNEXUS_*` variable according to repo conventions.

### Why this needed more than `render.yaml` and the README

The server rejected its own deploy's traffic, because two behaviors were hardcoded to a local install. The same-host origin guard on write routes derives its allowed origin from the bound host, and a container binds `0.0.0.0`, which has no host identity, so every index request from the UI was refused. Separately, `trust proxy` was pinned to loopback and private ranges, so behind a load balancer every request resolved to the same proxy hop and the per-IP rate limiter degraded into a single global limit. That second one was already flagged as a known follow-up in a `CLOUD-DEPLOY CAVEAT` comment in `api.ts`. Both are now driven by env vars, unset by default, so a local `gitnexus serve` behaves exactly as it does today.

On `GITNEXUS_TRUST_PROXY`, the Blueprint sets `1` rather than `true` on purpose. Express reads the leftmost `X-Forwarded-For` entry under `true`, which a client can spoof to get a fresh IP per request and walk around the rate limiter. A hop count only trusts the proxies actually in front of the service.

### Testing

- 116 unit tests across `cors`, `rate-limit`, `middleware-trust-proxy`, `api-analyze-upload`, and `security` pass, including new coverage for scheme enforcement, port handling, suffix attacks like `app.example.com.evil.com`, and the unset-is-inert case.
- `npx tsc --noEmit` in `gitnexus/` is clean; `render blueprints validate` passes.
- I've been running the Blueprint on Render since. Both services are live, the `fromService` pair resolves (the served page carries `__GITNEXUS_CONFIG__.backendUrl` pointing at the API service), and the origin allowlist behaves: the web origin gets `access-control-allow-origin` back, `https://evil.example.com` gets no such header.
- The full unit suite has 10 pre-existing failures on `main` (`analyzer-identity`, `evidence-provenance-helper`, `review-agent-workflow`, `incremental-orchestration`). They reproduce on a clean checkout and are unrelated to this branch.

⚠️ The Deploy to Render button can't be exercised until this merges. It points at `https://github.com/abhigyanpatwari/GitNexus`, and Render only finds `render.yaml` once it's on the default branch. To deploy the exact code in this PR before merging, point Render at this branch on my fork:

```
https://render.com/deploy?repo=https://github.com/Ho1yShif/GitNexus/tree/feat/render-deploy
```

### Security

The deploy is unauthenticated, and `onrender.com` hostnames appear in certificate transparency logs, so the URL should be treated as public. Anyone who finds it can index and delete repos, and can use `AZURE_DEVOPS_PAT` if one is set. This is called out in both the README section and a comment at the top of `render.yaml`. Rate limits bound the cost but are not access control. Happy to add a token gate if you'd prefer the default deploy be closed.


magyargergo:
@Ho1yShif Thank you for your submission! We are reviewing your PR and will get back to you as soon as possible.

README.md
```diff
@@ -80,6 +80,16 @@ That's it. `analyze` indexes the codebase, installs agent skills, registers Clau
 
 </details>
 
+### Deploy to Render
+
+Deploy GitNexus in one click:
+
+[![Deploy to Render](https://render.com/images/deploy-to-render-button.svg)](https://render.com/deploy?repo=https://github.com/abhigyanpatwari/GitNexus)
+
+Render provisions the API server, a persistent disk for indexes and cloned repos, and the web UI pointed at it, with no API keys required.
```
abhigyanpatwari: **Blocking.** This is the only surface a one-click user reads, and "with no API keys required" reads as a convenience feature when it actually describes the absence of authentication.

`render.yaml:6-8` says the right thing and says it plainly — unauthenticated, `onrender.com` names show up in certificate transparency logs, rate limits are not access control. Render renders the blueprint's resources, not its comments, so none of that reaches the person clicking the badge.

I stood up `gitnexus serve --host 0.0.0.0` locally with both new variables set exactly as this blueprint sets them. Every write route accepted an `Origin`-less request: `POST /api/analyze {"path":"/etc"}` returned `202` with a jobId, and `DELETE /api/repo`, `DELETE /api/analyze/:jobId`, `DELETE /api/embed/:jobId`, `POST /api/analyze/upload` and `POST /api/mcp` `initialize` all passed the guard. The read routes have no guard either, so every visitor can read every indexed repo's source.

Either ship the blueprint authenticated by default — `resolveAuthToken` and `timingSafeEqual` already exist in `http-transport.ts`, and Render has `generateValue: true`, so a token generated for the server and handed to `gitnexus-web` via `fromService` is a small change — or put an unmissable callout right here carrying the substance of `render.yaml:6-8`, and drop the Azure PAT keys from a public blueprint either way.

gitnexus/.env.example
```diff
@@ -45,3 +45,17 @@
...
 # See README § "Scope-resolution dispatch-target cap".
 # GITNEXUS_MAX_CALLABLE_VALUE_TARGETS=64
+
+# Hosted Deployment
+# Only for `serve` behind a proxy on a wildcard bind. Unset for a local install.
+
+# The public origin browsers reach this server through. A wildcard bind has no
+# host identity, so without this the same-host guard rejects the UI's writes.
+# A bare host matches either scheme; a scheme, if given, is enforced.
+# GITNEXUS_PUBLIC_ORIGIN=https://gitnexus-server.example.com
```
abhigyanpatwari: **Blocking.** This sample value is the API server's own hostname, but the variable is compared against the browser's `Origin` request header — the origin of the *page* making the call.

The web UI calls the API cross-origin via `backendUrl` (`gitnexus-web/src/config/ui-constants.ts:6`, `hooks/useBackend.ts:29`), so the correct value is the **web UI's** URL. That is exactly what `render.yaml:43-47` wires, from `gitnexus-web`'s `RENDER_EXTERNAL_URL` — the blueprint is right and the docs are backwards. `README.md:521` and the runtime warning at `api.ts:767` ("set `GITNEXUS_PUBLIC_ORIGIN` to its public origin", where "its" is the server) read the same way.

Since `Dockerfile.cli` is API-only, the two origins are always different on a split deploy, so the documented reading is always wrong there. I confirmed the consequence locally: set to the server's own URL, requests from the UI origin get no `Access-Control-Allow-Origin` on reads and `403 origin_not_allowed` on writes — a blank UI — while the one origin that works is the one no browser is on.

Suggest rewording all three to describe the caller ("the origin of the web UI that will call this server — the page origin browsers load, not this server's own URL; in a single-service deploy where `serve` also serves the UI they are the same"), changing this sample to a UI hostname, and pointing at `render.yaml`'s `fromService` block as the worked example. Worth also saying out loud that the bare-host form admits `http://` as well as `https://` (`middleware.ts:78-80`) — harmless for the shipped blueprint since `RENDER_EXTERNAL_URL` carries the scheme, but this example is what a TLS deploy will copy.

gitnexus/src/server/middleware.ts
```diff
@@ -93,3 +145,24 @@ export function createLocalhostOriginGuard(boundHost?: string) {
...
+ * client-controlled leftmost `X-Forwarded-For` entry, so a spoofed chain would
+ * hand the rate limiter a fresh IP per request.
+ */
+export function resolveTrustProxy(raw?: string): string | number | boolean {
+  const value = raw?.trim();
+  if (!value) return DEFAULT_TRUST_PROXY;
+  if (/^\d+$/.test(value)) return Number(value);
+  if (/^true$/i.test(value)) return true;
+  if (/^false$/i.test(value)) return false;
+  return value;
```
abhigyanpatwari, line 358: **Blocking.** This `return value` makes the docstring above untrue — it says the function coerces the variable "into a value Express accepts", and unrecognised strings are passed straight through. Express compiles `trust proxy` eagerly at `app.set` time (`api.ts:727`), so anything `proxy-addr` doesn't recognise throws there.

Reproduced against the Express 5.2.1 pinned in this tree:

| value | result |
|---|---|
| `garbage` | `invalid IP address: garbage` |
| `10.0.0.0/99` | `invalid range on address: 10.0.0.0/99` |
| `-1`, `1.5` | fall through as strings, throw the same way |
| `9`&times;400 | `Infinity` — accepted silently |

To be fair, this is not an uncaught crash: `serve.ts:35-62` wraps `createServer` and exits 1 via `serve.startFailed`. But the operator only sees `invalid IP address: yes`, which never names `GITNEXUS_TRUST_PROXY`, and on Render that's a redeploy loop with no pointer to the cause.

The numeric branch is the mirror image: unbounded. I reproduced `resolveTrustProxy('9'.repeat(400))` returning `Infinity`, which Express accepts and which then reports `trust proxy` true at hop 0, 1, 5, 100 and 10000 — it trusts every hop, i.e. exactly the `true` behaviour the docstring three lines above warns against, reached by one extra keystroke. `true` itself is accepted with no warning too, and because `createRouteLimiter` (`validation.ts:151-165`) supplies its own `keyGenerator`, `express-rate-limit`'s built-in alarm for this setting never fires — I saw 6/6 requests return 200 against a limit of 3 by rotating `X-Forwarded-For`, silently.

This also puts the variable out of step with every neighbour in the same README table (`GITNEXUS_LBUG_BUFFER_POOL_SIZE`, `GITNEXUS_LBUG_MAX_DB_SIZE`, `GITNEXUS_MAX_CALLABLE_VALUE_TARGETS` all warn and fall back), and with the sibling function added in this same commit — `createPublicOriginMatcher` returns `undefined` on bad input precisely because "an invalid origin must never widen the allow-list" (`middleware.ts:57-58`).

`proxy-addr` is already an Express dependency and exports `compile`, so:

```ts
try {
  proxyaddr.compile(value.split(',').map((s) => s.trim()));
  return value;
} catch {
  logger.warn({ [TRUST_PROXY_ENV]: value }, 'invalid value; falling back to the default');
  return DEFAULT_TRUST_PROXY;
}
```

Bounding the numeric branch to something sane (1&ndash;16) and falling into the same path otherwise closes the `Infinity` hole. A `logger.warn` when the resolved value is `true` is cheap insurance, accepting `yes`/`no`/`on`/`off` removes a whole class of typo, and the docstring needs correcting either way.

gitnexus/src/server/api.ts
```diff
@@ -753,19 +748,23 @@ export const createServer = async (port: number, host: string = '127.0.0.1') =>
...
-  // A wildcard bind (`0.0.0.0`/`::`) has no single host identity for the
-  // same-host check, so browser write routes accept only loopback origins.
-  // Warn the operator so a remote-access deployment isn't silently write-blocked.
-  if (host && normalizeBoundHost(host) === undefined) {
+  // A wildcard bind has no single host identity for the same-host check, so
+  // writes stay loopback-only unless PUBLIC_ORIGIN_ENV declares an origin.
+  if (
+    host &&
+    normalizeBoundHost(host) === undefined &&
+    !createPublicOriginMatcher(process.env[PUBLIC_ORIGIN_ENV])
```
abhigyanpatwari: **Blocking.** This gates the warning on the matcher being *constructible*, not on it being able to match anything — so setting `GITNEXUS_PUBLIC_ORIGIN` **wrong** is diagnosed strictly worse than leaving it unset.

`normalizeBoundHost` goes through `new URL('http://' + host)`, and WHATWG hostname parsing for special schemes accepts a lot of things that are not hostnames. I ran these against the real `createPublicOriginMatcher` — all truthy, all unmatchable:

```
"https://a.onrender.com,https://b.onrender.com"  truthy  expects hostname "a.onrender.com,https"
"https://*.onrender.com"                         truthy  expects hostname "*.onrender.com"
"2001:db8::1:8443"                               truthy  expects hostname "[2001:db8::1:8443]"
"1"                                              truthy  expects hostname "0.0.0.1"
"*"                                              truthy  expects hostname "*"
```

Each one suppresses this warning. And because the same variable also gates the CORS allowlist (`api.ts:137`), and `isAllowedOrigin` has no `onrender.com` entry, the result isn't a write-only failure — it's a blank UI: no `Access-Control-Allow-Origin` on reads, 403 on writes, and *nothing* in the server log. The guard imports no logger, CORS denial is silent, and the one operator-facing signal has been switched off by the very misconfiguration that needed it.

A comma-separated value seems the most likely way in, since the sibling row one line below in the same README table documents "an Express-style list", and CORS allowlists are conventionally comma-separated — `GITNEXUS_MCP_ALLOWED_REPOS` (`README.md:518`) is documented as comma-separated and *fails startup* on bad entries. Worth matching that contract, or saying explicitly that only one origin is supported.

Two independent fixes, either of which helps materially: (a) always emit this warning, branching the message on whether a public origin was configured, and log the *resolved* hostname at `info` on startup so an operator can see what the server thinks it is; (b) in `createPublicOriginMatcher`, reject a derived hostname that cannot be a real host (contains `,`, `;`, `*`, whitespace, or is a bare digit) so it lands in the existing "ignored" bucket.

README.md
```diff
@@ -80,6 +80,16 @@ That's it. `analyze` indexes the codebase, installs agent skills, registers Clau
...
+
+Indexing is memory-bound, so if `gitnexus-server` runs out of memory on a large repo, raise its `plan` and `sizeGB`.
```
abhigyanpatwari: This prescribes disk for a memory problem. `plan` is the compute tier (RAM); `sizeGB` is the persistent disk — they're independent, and raising `sizeGB` cannot resolve an OOM. `render.yaml:22-23` gets this right and mentions only `plan`.

Someone hitting OOM will follow this, bump the disk, redeploy, and still be OOM — having paid for storage they didn't need.

Suggest: "raise its `plan` (which sets available RAM — `standard` is 2 GB, `pro` is 4 GB). Raise `sizeGB` only if the disk fills with clones and indexes."

abhigyanpatwari (commented):
Thanks for this — and thanks for the care that went into it. A few things I want to call out before the criticism, because they're the parts that are easy to get wrong and you got right:

- **Defaults are genuinely unchanged.** `DEFAULT_TRUST_PROXY` (`gitnexus/src/server/middleware.ts:150`) is byte-identical to the literal that was removed from `api.ts`, `resolveTrustProxy(undefined)` returns it, and `createPublicOriginMatcher(undefined)` returns `undefined` so both new branches are inert. I ran the four touched test files locally — 98/98 pass — and `npx tsc --noEmit` is clean. A local `gitnexus serve` sees no behavior change at all.
- **`value: 1` rather than `true`** for `GITNEXUS_TRUST_PROXY` (`render.yaml:50-51`) is the correct choice, and the inline comment gets the reason right. I tried to spoof the limiter through it with a real Express app plus the real `createRouteLimiter`: with a numeric hop count `proxy-addr` reads the socket-adjacent end of `X-Forwarded-For`, which the client cannot influence, and the limiter held (429s at the limit). With `true` it collapses immediately. Most PRs get this backwards.
- **The origin matcher fails closed.** Everything I threw at it — case folding, IDN/punycode, credentials in the value, trailing-dot FQDN, suffix/prefix confusion, `*`, `0.0.0.0`, `::`, non-HTTP schemes, a quoted value — either produced `undefined` or a matcher no browser `Origin` can satisfy. The comparison is `===` on the WHATWG hostname, with no `endsWith` anywhere. The rationale at `middleware.ts:57-58` is written down and the code honours it.
- **`render.yaml` validates.** I checked it against Render's published schema (`https://render.com/schema/render.yaml.json`): `VALID: true`. `projects:`/`environments:`, `previews.generation: off`, and the unquoted `value: 4747` / `value: 1` are all legal — no YAML-boolean surprise on `off`. The internal wiring is right too: `mountPath: /data/gitnexus` matches `Dockerfile.cli`'s `GITNEXUS_HOME`, `healthCheckPath: /api/health` matches `api.ts:893`, and `GITNEXUS_BACKEND_URL` matches what `docker-server.mjs` actually reads.
- **The `render.yaml` security disclosure is accurate and complete.** `render.yaml:6-8` says exactly the right thing, including that `onrender.com` names show up in certificate transparency logs and that rate limits are not access control. My complaint below is about *where* it lives, not what it says.
- **The commit message explains the why**, and `CONTRIBUTING.md:76` adds a rule that this PR then satisfies for both new variables (`README.md:521-522`, all four columns populated). That's the right instinct.

That said, I don't think this is mergeable yet. The blockers are mostly about what the button hands an unsuspecting user, and about the documentation around the two new variables.

---

## Blocking

### 1. The button deploys a publicly writable, wholly unauthenticated server, and the README section a user actually reads says the opposite

`README.md:83-91`, `render.yaml:6-8`, `gitnexus/src/server/middleware.ts:110-113`

The write guard passes through whenever there is no `Origin` header — that is documented and intentional (`middleware.ts:97`: "Non-browser requests (no Origin header, e.g. curl / the CLI) pass through"), because it is CSRF protection, not authentication:

```ts
    const origin = req.headers.origin;
    if (origin === undefined) {
      next();
      return;
    }
```

On a loopback or LAN bind that limitation is acceptable. This blueprint turns it into an internet-facing one. I stood up `gitnexus serve --host 0.0.0.0` locally with `GITNEXUS_PUBLIC_ORIGIN` and `GITNEXUS_TRUST_PROXY=1` set exactly as `render.yaml` sets them, and every write route accepted an `Origin`-less request: `DELETE /api/repo` (404 for a missing repo, i.e. the guard passed), `POST /api/analyze {"path":"/etc"}` → `202 {"jobId":…}`, `DELETE /api/analyze/:jobId`, `DELETE /api/embed/:jobId`, `POST /api/analyze/upload`, and `POST /api/mcp` `initialize` → `200` with a session id. `GET /api/repos` (`api.ts:937`), `/api/graph` (`:1074`) and `POST /api/query` (`:1151`) carry no origin guard and no limiter, and `/api/file` (`:1322`) and `/api/grep` (`:1336`) have a limiter but no guard — so on a shared deploy every visitor can read every indexed repo's source. A 403 only appears when a browser-shaped `Origin` is present and unmatched.

What makes this hard to wave through is that the codebase already refuses to do this elsewhere. `gitnexus/src/mcp/http-transport.ts:476-482`:

```ts
  if (!authToken && !isLoopbackHost(host)) {
    throw new Error(
      `Refusing to start the MCP HTTP server on a non-loopback host (${host}) without ` +
        'authentication — it would expose all indexed repos to anyone who can reach it. ' +
        'Pass --auth-token (or set GITNEXUS_MCP_AUTH_TOKEN), or bind --host 127.0.0.1. ' +
        'This applies to --host 0.0.0.0 and --host :: as well.',
    );
  }
```

`serve` mounts the same MCP handler through `mountMCPEndpoints` (`api.ts:780` → `mcp-http.ts:22`), and `createStreamableHttpHandler`'s options have no `authToken` field — the token check lives in `startMcpHttpServer`, which that path bypasses. **That gap pre-dates this PR**; I'm not saying the diff introduced it. I'm saying this PR is the first thing that points it at the public internet, with `Dockerfile.cli`'s `CMD` binding `--host 0.0.0.0`.

Two specific amplifiers:

- `render.yaml:52-57` offers `AZURE_DEVOPS_PAT` as a deploy knob with the comment "Any visitor can use this PAT". That's honest, but it's an argument for deleting the knob, not documenting it. The PAT is attached by host policy only (`git-clone.ts:537-541`), and `isAzureDevOpsUrl` matches the whole org — so a stranger can `POST /api/analyze` an Azure URL, have the server spend the operator's credential, then read the private repo back through the ungated `/api/file` and `/api/grep`. I could not complete the clone step locally (no Azure org), so treat the last link as read-from-source rather than demonstrated; every other link I did exercise.
- Rate limits bound requests per IP, not cost per request. `POST /api/analyze` is 10/min with no cap on clone size and no disk quota against `sizeGB: 10`; `DELETE /api/analyze/:jobId` and `DELETE /api/embed/:jobId` have no limiter at all. `render.yaml:8` already concedes the point.

Meanwhile `README.md:89` — the only surface a one-click user sees, since Render renders the blueprint's resources and not its comments — reads:

> Render provisions the API server, a persistent disk for indexes and cloned repos, and the web UI pointed at it, with no API keys required.

"With no API keys required" describes the absence of authentication as a convenience feature.

**What I'd like to see.** Preferably: make the blueprint authenticated by default. The primitives already exist in this repo — `resolveAuthToken` and `timingSafeEqual` in `http-transport.ts` — and Render supports `generateValue: true`, so a `GITNEXUS_SERVE_AUTH_TOKEN` generated for the server and handed to `gitnexus-web` via `fromService` is a small, idiomatic change. Mirroring `http-transport.ts:476` (refuse to start on a wildcard bind with a public origin set and no token) would keep the two surfaces from drifting apart again.

If shipping an intentionally public demo is the deliberate decision, then the README needs an unmissable callout immediately under the badge carrying the substance of `render.yaml:6-8` — unauthenticated, CT-log discoverable, anyone with the URL can index and delete, all indexed source is readable, rate limits are not access control — and the Azure PAT keys should come out of the public blueprint regardless. "With no API keys required" should also go or be qualified.

### 2. `.env.example` and the README name the wrong service for `GITNEXUS_PUBLIC_ORIGIN`

`gitnexus/.env.example:52-55`, `README.md:521`, `gitnexus/src/server/api.ts:767`

The value is compared against the browser's `Origin` request header, i.e. the origin of the *page* making the call. The web UI calls the API cross-origin via `backendUrl` (`gitnexus-web/src/config/ui-constants.ts:6`, `hooks/useBackend.ts:29`), so on this blueprint the correct value is `gitnexus-web`'s URL — which is exactly what `render.yaml:43-47` wires. The prose says the opposite:

```
# The public origin browsers reach this server through. A wildcard bind has no
# host identity, so without this the same-host guard rejects the UI's writes.
# A bare host matches either scheme; a scheme, if given, is enforced.
# GITNEXUS_PUBLIC_ORIGIN=https://gitnexus-server.example.com
```

The sample value is literally the API server's hostname. `README.md:521` ("Public origin browsers reach `serve` through") reads the same way, and so does the warning the operator will be looking at when they set this (`api.ts:767`: "for a hosted deployment, set `GITNEXUS_PUBLIC_ORIGIN` to its public origin" — "its" being the wildcard-bound server).

Since `Dockerfile.cli` is API-only, the two origins are *always* different on a split deploy, so the documented reading is always wrong there. I confirmed the consequence locally: with `GITNEXUS_PUBLIC_ORIGIN` set to the server's own URL, requests from the UI origin get no `Access-Control-Allow-Origin` on reads and 403 `origin_not_allowed` on writes — a blank UI — while the one origin that works is the one no browser is on. Anyone self-hosting at `api.example.com` + `app.example.com` from the docs alone lands in the same place.

**Fix.** Reword all three to describe the caller: "the origin of the web UI that will call this server — the page origin browsers load, not this server's own URL. In a single-service deploy where `serve` also serves the UI, they are the same." Change the `.env.example` sample to a UI hostname, change `api.ts:767` to "the origin of the UI that will call it", and point at `render.yaml`'s `fromService` block as the worked example. While you're there: the bare-host form admits `http://` as well as `https://` (`middleware.ts:78-80`), which is fine for platform service-discovery fields but worth saying out loud next to an example that a TLS deploy will copy — the shipped blueprint is safe here because `RENDER_EXTERNAL_URL` carries the scheme.

### 3. `GITNEXUS_TRUST_PROXY` has no validation in either direction

`gitnexus/src/server/middleware.ts:161-168`, consumed at `gitnexus/src/server/api.ts:727`

```ts
export function resolveTrustProxy(raw?: string): string | number | boolean {
  const value = raw?.trim();
  if (!value) return DEFAULT_TRUST_PROXY;
  if (/^\d+$/.test(value)) return Number(value);
  if (/^true$/i.test(value)) return true;
  if (/^false$/i.test(value)) return false;
  return value;
}
```

The docstring says this "coerce[s] `TRUST_PROXY_ENV` into a value Express accepts" — the final `return value` makes that untrue. Express compiles `trust proxy` eagerly at `app.set` time, so anything `proxy-addr` doesn't recognise throws there. Reproduced against the pinned Express 5.2.1 in this tree: `garbage` → `invalid IP address: garbage`; `10.0.0.0/99` → `invalid range on address: 10.0.0.0/99`; `-1` and `1.5` fall through as strings and throw the same way. To be fair to the PR, this is *not* an uncaught crash — `serve.ts:35-62` wraps `createServer` and exits 1 via `serve.startFailed`. But the operator sees only `invalid IP address: yes`, which never names `GITNEXUS_TRUST_PROXY`, and on Render that's a redeploy loop with no pointer to the cause.

That also puts this variable out of step with every neighbour in the same table: `GITNEXUS_LBUG_BUFFER_POOL_SIZE` ("invalid values warn and fall back to the default"), `GITNEXUS_LBUG_MAX_DB_SIZE`, `GITNEXUS_PROFILE_DEFERRED_SLOW_MS`, `GITNEXUS_MAX_CALLABLE_VALUE_TARGETS` — all fall back — or `GITNEXUS_MCP_READ_ONLY`, which fails startup with GitNexus's own actionable message. Note the asymmetry with the sibling function added in the same commit: `createPublicOriginMatcher` returns `undefined` on bad input precisely because "an invalid origin must never widen the allow-list" (`middleware.ts:57-58`). `resolveTrustProxy` has no equivalent.

The numeric branch has the mirror-image problem: it's unbounded. I reproduced `resolveTrustProxy('9'.repeat(400))` → `Infinity`, which Express accepts silently and which then reports `trust proxy` true at hop 0, 1, 5, 100 and 10000 — it trusts every hop, i.e. exactly the `true` behaviour the docstring three lines above warns against, reached by an extra keystroke. `true` itself is also accepted with no warning, and because `createRouteLimiter` (`validation.ts:151-165`) supplies its own `keyGenerator`, `express-rate-limit`'s built-in `ERR_ERL_PERMISSIVE_TRUST_PROXY` alarm for exactly this setting never fires.

**Fix.** Validate before returning: `proxy-addr` is already an Express dependency and exports `compile`, so a `try { proxyaddr.compile(value.split(',').map(s => s.trim())); return value; } catch { … }` that warns naming `TRUST_PROXY_ENV` and returns `DEFAULT_TRUST_PROXY` gets you the repo's documented convention in a few lines. Bound the numeric branch to something sane (say 1–16) and fall into the same path otherwise. A `logger.warn` when the resolved value is `true` would be cheap insurance. Accepting `yes`/`no`/`on`/`off` as booleans costs two regexes and removes a whole class of operator typo. Finally, correct the docstring.

### 4. Setting `GITNEXUS_PUBLIC_ORIGIN` *wrong* is diagnosed strictly worse than leaving it unset

`gitnexus/src/server/api.ts:757-769`, `gitnexus/src/server/middleware.ts:76-80`

The wildcard-bind warning is gated on the matcher being *constructible*, not on it being able to match anything:

```ts
  if (
    host &&
    normalizeBoundHost(host) === undefined &&
    !createPublicOriginMatcher(process.env[PUBLIC_ORIGIN_ENV])
  ) {
```

`normalizeBoundHost` goes through `new URL('http://' + host)`, and WHATWG hostname parsing for special schemes accepts a lot of things that are not hostnames. I reproduced these against the real `createPublicOriginMatcher`, all truthy and all unmatchable:

```
"https://a.onrender.com,https://b.onrender.com"  matcher? true   expects hostname "a.onrender.com,https"
"https://*.onrender.com"                         matcher? true   expects hostname "*.onrender.com"
"2001:db8::1:8443"                               matcher? true   expects hostname "[2001:db8::1:8443]"
"1"                                              matcher? true   expects hostname "0.0.0.1"
"*"                                              matcher? true   expects hostname "*"
```

Each one suppresses the warning. And because the same variable also gates the CORS allowlist (`api.ts:137`) — and `isAllowedOrigin` has no `onrender.com` entry — the result isn't a write-only failure, it's a blank UI: no `Access-Control-Allow-Origin` on reads, 403 on writes, and *nothing* in the server log. The guard imports no logger, CORS denial is silent, and the one operator-facing signal has been switched off by the very misconfiguration that needed it. A comma-separated value is the most likely way in, since the sibling row one line below in the same README table documents "an Express-style list" and CORS allowlists are conventionally comma-separated (`GITNEXUS_MCP_ALLOWED_REPOS`, `README.md:518`, is documented as comma-separated and *fails startup* on bad entries — worth matching that contract, or saying explicitly that only one origin is supported).

**Fix.** Two independent changes, either of which helps materially: (a) always emit the wildcard-bind warning, branching the message on whether a public origin was configured, and log the *resolved* hostname at `info` on startup so an operator can see what the server thinks it is; (b) in `createPublicOriginMatcher`, reject a derived hostname that cannot be a real host (contains `,`, `;`, `*`, whitespace, or is a bare digit), returning `undefined` so it lands in the existing "ignored" bucket.

---

## Smaller items

- **`README.md:91` prescribes disk for a memory problem.** "if `gitnexus-server` runs out of memory on a large repo, raise its `plan` and `sizeGB`" — `plan` is the compute tier (RAM), `sizeGB` is the persistent disk; they're independent. `render.yaml:22-23` gets this right and mentions only `plan`. Suggest: "raise its `plan` (which sets available RAM — `standard` is 2 GB, `pro` is 4 GB). Raise `sizeGB` only if the disk fills with clones and indexes."
- **`autoDeploy: true` on both services (`render.yaml:30`, `:67`)** opts every downstream instance into upstream pushes with no compatibility review, including onto a persistent disk. Worth noting that your earlier blueprint (#2775) deliberately set `false` and discussed the tradeoff. `autoDeployTrigger: off` is the current schema field for this if you want to preserve the choice explicitly.
- **The trust-proxy wiring assertion got weaker, not just retargeted.** `gitnexus/test/unit/rate-limit.test.ts:273-274` now matches `/app\.set\(\s*'trust proxy'\s*,\s*resolveTrustProxy\(/`, which stops at the open paren. I evaluated it standalone: `resolveTrustProxy(process.env.SOMETHING_ELSE)`, `resolveTrustProxy()` and `resolveTrustProxy('always')` all match. The old assertion pinned the literal value; this one can't catch a wrong-argument regression. Either capture the argument (`/resolveTrustProxy\(\s*process\.env\[TRUST_PROXY_ENV\]\s*\)/`) or, better, set the env var to a sentinel and assert `app.get('trust proxy')` — that re-validates the coercion too.
- **`api.ts:135-136` says something that isn't so.** "a pure predicate with no construction step" sits directly above `createPublicOriginMatcher(process.env[PUBLIC_ORIGIN_ENV])?.(...)`, which is a factory call that trims, may run `new URL` twice and allocates a closure, on every CORS check including preflights. The real reason for the asymmetry (this is a top-level exported predicate with no single construction moment, and tests drive it directly) is a good one — it's just not what the comment says. I measured the cost and it's immaterial (see below); this is purely about the comment.
- **`render.yaml` has no CI validation.** Nothing in `.github/workflows/` or the test suite parses it, so the couplings the file itself documents — `mountPath` equalling `GITNEXUS_HOME`, `PORT` matching `Dockerfile.cli`'s hardcoded `--port 4747` — are unenforced, as is the env-var key spelling. A schema check against the `$schema` on line 1, or a tiny Node assertion on those three invariants, would be cheap.
- **Tests worth adding** alongside the fixes above: `expect(() => express().set('trust proxy', resolveTrustProxy(v))).not.toThrow()` over the accepted grammar; a `logger.warn` spy proving the wildcard-bind warning still fires for `GITNEXUS_PUBLIC_ORIGIN='*'`; a `createPublicOriginMatcher(x) === undefined` table for the comma/wildcard/bare-digit class; and one behavioural rate-limit case (rotating `X-Forwarded-For` must still 429 under `1`, and does not under `true`) so the property the whole story rests on is pinned. `stripPort`'s bracketed-IPv6-with-port substitution also never actually fires in any test — the only IPv6 case has no port, so the regex is a no-op there.
- Optional: `SECURITY.md`'s "Recommended Hardening for Forks and Self-Hosted Deployments" says nothing about hosted deploys, and it's the fallback doc for anyone looking.

---

## Checked and found sound

Recording these because several took real effort to disprove and I don't want them re-litigated:

- **Origin-matcher bypasses.** Case folding, IDN/Cyrillic homographs, punycode, credentials in the value (`https://good@evil.com` widens to `evil.com` only, and only if the operator types it), path/query/fragment, trailing-dot FQDN, suffix and prefix confusion (`…onrender.com.evil.com`), `*`, `0.0.0.0`, `::`, IPv4-mapped IPv6, non-HTTP schemes, and a shell-quoted value — all either `undefined` or a matcher no browser can satisfy. No cross-match found in either direction.
- **Dropping the `normalizedBoundHost &&` truthiness guard** in `middleware.ts:123-132` is behaviour-preserving. `normalizeBoundHost` can never return `''` (it guards `''`/`0.0.0.0`/`[::]` explicitly at `:36-38`, and a brute force over short inputs found no counterexample), `new URL(origin).hostname` is always a string so `hostname === undefined` is unconditionally false, and the protocol check at `:117-119` runs first so the only empty-hostname case (non-special schemes) is unreachable.
- **`stripPort` never widens.** Every malformed input I could construct fails closed; bracketed IPv6 with a port, bare IPv6, and `host:port` all round-trip correctly, and the ">1 colon means bare IPv6" heuristic is sound because the shortest valid IPv6 literal already has two colons. The one real mis-derivation (unbracketed IPv6 *with* a port) is a silent no-match, which is item 4 above, not a widening.
- **The CORS-vs-guard freshness asymmetry is not exploitable.** `createServer` has one caller, `isAllowedOrigin` has one call site, and nothing in `src/` writes `process.env.GITNEXUS_PUBLIC_ORIGIN`, so the two readers cannot observe different values in a real process. There is also no dotenv/`--env-file` step anywhere. The module-level `requireLocalhostOrigin` singleton (`middleware.ts:147`) does snapshot at import time, but `api.ts:753` shadows it and no production route uses it.
- **`previews: generation: off` is not a YAML-boolean trap.** Parsed with the repo's own `js-yaml` it is the string `"off"`, and the file validates against Render's schema as-is.
- **The per-call matcher construction is not a performance problem.** Measured: ~130–144 ns/op with the variable unset, ~780–855 ns/op with it set, against ~2–3 ns snapshotted. At 100 req/s with a preflight that's ~0.017% of one core; you'd need roughly 6,000 sustained req/s to reach 1%. Real inconsistency, immaterial cost.
- **No PAT exfiltration via a cloned repo's `.git/config`.** The credential is injected per invocation as `http.extraheader` through `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_*` (`git-clone.ts:619-628`), never embedded in the remote URL, so it isn't persisted into the clone. Clone argv construction and header-key sanitisation are careful too.
- **Analyze workers are not unbounded.** `JobManager` is single-slot ("one active job at a time") with a per-repo lock, so memory pressure is bounded to one forked worker rather than scaling with concurrent callers.
- **`POST /api/query` cannot write.** The connection is opened `readOnly: true` with an explicit 403 on write attempts, and a load-only extension policy.
- **`GITNEXUS_BACKEND_URL` injection is not an XSS vector** — `docker-server.mjs` validates it as an http/https URL and escapes `<`, `>`, `&` for the script tag.
- **`GITNEXUS_TRUST_PROXY=1` is not spoofable** behind an appending load balancer, as covered above, and the unset default is byte-identical to the old hardcoded value.
- **The blueprint's internal cross-references all hold**: schema valid, port agrees with the image's `CMD`, health path exists, disk mount equals `GITNEXUS_HOME`, backend env-var key matches what the web image reads, and `previews: off` correctly suppresses per-PR preview environments (so no third unlisted origin).

---

## Open questions

1. **Is the port-blindness of `GITNEXUS_PUBLIC_ORIGIN` intentional?** A browser origin is scheme + host + port, and this compares scheme + hostname only: I confirmed a matcher configured for `https://app.example.com` admits `https://app.example.com:4444` and `:8080`, and the new tests pin that behaviour deliberately (`cors.test.ts:198-201`, `api-analyze-upload.test.ts:417-419`). Two defensible readings. If it's intentional (the pre-existing `normalizedBoundHost` comparison is equally port-blind, so it isn't a regression, and Render doesn't let you serve a second app on another port of the same hostname), then say so in the README row rather than only in tests. If it isn't, an operator behind a reverse proxy at `https://home.example.com:8443` is granting write access to every other port on that hostname. I'd like your call on which it is rather than guessing.
2. **Has this blueprint actually been deployed once end to end?** `render.yaml:43-47` and `:72-75` reference each other's `RENDER_EXTERNAL_URL`. That may well resolve fine — the URL derives from the service name at creation, not from a deploy output — but if it comes back empty on a first deploy, `GITNEXUS_PUBLIC_ORIGIN` is empty, the matcher is `undefined`, and you get a green dashboard with a dead UI. Same question for the disk: `Dockerfile.cli` runs `USER node`, and the build-time `chown` of `/data` is shadowed by the runtime volume mount, so whether uid 1000 can write `/data/gitnexus` depends on Render's mount ownership. Pasting the resulting env-var values and a successful analyze from one clean-account deploy would retire both.
3. **Why was the reverse-proxy topology abandoned?** Your earlier #2775 made the server a private service and had `gitnexus-web` proxy `/api/*`, so the browser never made a cross-origin request and *no* trust-boundary change was needed. That's a materially safer shape for a one-click deploy. If it was dropped because private services need a higher Render tier, that's worth stating in the PR body — otherwise a reviewer reasonably asks why we're widening CORS and the CSRF guard at all.
4. If the intent is a public demo rather than a hosting recommendation, should the badge sit in the README's install flow, or in a clearly-labelled demo section?

---

## CI

Green except `tests / windows-latest (platform-sensitive) 3/3`, which fails in `test/integration/cli-limit-e2e.test.ts` with "Analyze failed (status null)". That file isn't touched here and the same shard passes on #2803, so I'm treating it as pre-existing Windows worker flake rather than anything from this diff. Everything else passes: typecheck, all ubuntu and macOS shards, windows 1/3 and 2/3, CodeQL, gitleaks, both Docker builds, and the packaged install smoke. The `Vercel` failure is the fork authorization gate, not a code issue. Locally the 98 tests in the four changed test files pass and `tsc --noEmit` is clean.

---

## Verdict

**Request changes.** The middleware work is careful and the defaults really are untouched — I'd take items 2, 3 and 4 as straightforward fixes and the smaller items are mostly one-liners. Item 1 is a product decision, not a code review nit, and it's the one that has to be settled first: either the blueprint ships authenticated by default (which is cheap here, given `resolveAuthToken`/`timingSafeEqual` already exist and Render has `generateValue`), or the README carries the disclosure that currently lives only in a YAML comment and the Azure PAT keys come out. I'd rather land the first one. Happy to review a follow-up quickly either way.

render.yaml
```diff
@@ -0,0 +1,75 @@
...
+            region: oregon
+            # Indexing OOMs on the 512MB starter. standard (2GB) covers
+            # small/mid repos; bump to pro (4GB) or larger for a big one.
+            plan: standard
+            dockerfilePath: ./Dockerfile.cli
+            dockerContext: .
+            # No dockerCommand — the image's CMD already binds the port below,
+            # and Render re-wraps dockerCommand in a shell, so `sh -c` exits 127.
+            healthCheckPath: /api/health
+            autoDeploy: true
```
abhigyanpatwari: `autoDeploy: true` here and at `:67` opts every downstream instance into upstream pushes with no compatibility review — onto a persistent disk, so a schema-incompatible change arrives unattended.

Worth noting your earlier blueprint (#2775) deliberately set `false` and discussed the tradeoff. `autoDeployTrigger` is the current schema field for expressing this explicitly — it takes `off`, `commit` or `checksPass`, so `autoDeployTrigger: off` preserves the choice and lets operators upgrade deliberately.

gitnexus/test/unit/rate-limit.test.ts
```diff
@@ -269,10 +270,13 @@ describe('production routes — rate-limit middleware wiring', () => {
     expect(apiSource).not.toMatch(/app\.options\(\s*'\/\*'/);
   });
 
-  it('createServer wires trust proxy to loopback/linklocal/uniquelocal', () => {
-    expect(apiSource).toMatch(
-      /app\.set\(\s*'trust proxy'\s*,\s*'loopback,\s*linklocal,\s*uniquelocal'\s*\)/,
-    );
+  it('createServer wires trust proxy through resolveTrustProxy', () => {
+    expect(apiSource).toMatch(/app\.set\(\s*'trust proxy'\s*,\s*resolveTrustProxy\(/);
```
abhigyanpatwari: This assertion got weaker, not just retargeted. The regex stops at the open paren, so it no longer pins the argument. I evaluated it standalone: `resolveTrustProxy(process.env.SOMETHING_ELSE)`, `resolveTrustProxy()` and `resolveTrustProxy('always')` all match it.

The test it replaced pinned the literal wired value, so it would have caught a wrong-argument regression; this one can't. The companion test below only covers `resolveTrustProxy(undefined)`.

Either capture the argument (`/resolveTrustProxy\(\s*process\.env\[TRUST_PROXY_ENV\]\s*\)/`) or, better, set the env var to a sentinel and assert `app.get('trust proxy')` — that re-validates the coercion at the same time.

gitnexus/src/server/api.ts
```diff
@@ -124,6 +132,10 @@ export const isAllowedOrigin = (origin: string | undefined): boolean => {
   // Only allow HTTP(S) origins — reject ftp://, file://, etc.
   if (protocol !== 'http:' && protocol !== 'https:') return false;
 
+  // Read per-call, not snapshotted — a pure predicate with no construction
+  // step, unlike the write guard in middleware.ts.
+  if (createPublicOriginMatcher(process.env[PUBLIC_ORIGIN_ENV])?.(protocol, hostname)) return true;
```
abhigyanpatwari: Minor, but the comment above this line says something that isn't so. "A pure predicate with no construction step" sits directly above a call to `createPublicOriginMatcher(...)`, which is a factory: it trims, may run `new URL` twice, and allocates a closure — on every CORS check, including preflights.

The real reason for the asymmetry with the write guard is a good one (this is a top-level exported predicate with no single construction moment, and tests drive it directly) — it's just not what the comment claims. I measured the cost and it's immaterial: ~130-144 ns/op with the variable unset versus ~2-3 ns snapshotted, which at 100 req/s is a rounding error. So this is purely about the comment, not the code.

magyargergo:
@Ho1yShif If you need help with sorting this out, please let us know 🙏

azizur100389:
Reviewed current head `cd924810` against current main `7468cc91`; it merges cleanly. `npx tsc --noEmit` passes, and the six targeted server suites pass (119 tests), but this is not merge-ready.

The Render blueprint exposes an unauthenticated public server: origin-less requests bypass the origin guard, allowing remote indexing, source access, and repository deletion. The optional Azure PAT increases that risk. The deployment should be authenticated by default, and Azure credentials should be removed from the public blueprint.

Also, `GITNEXUS_PUBLIC_ORIGIN` is documented as the API server origin, but it must be the calling web UI’s origin. `GITNEXUS_TRUST_PROXY` needs validation and bounded hop counts, and malformed public-origin values should not suppress startup diagnostics.

Ho1yShif:
@abhigyanpatwari Thank you for this review. The trust-proxy spoofing test against a real Express app, the origin-matcher bypass sweep, and the measured cost of per-call matcher construction all took real work, and the "Checked and found sound" section saved me from re-litigating things I'd have wondered about myself. Answers below, with 1 and 3 folded together because they turn out to be one decision.

@azizur100389 @magyargergo please feel free to weigh in, I appreciate your feedback and want to move forward with design decisions that are comfortable for the whole maintainer team.

## 1 + 3. Port-blindness, and why the reverse-proxy topology went away

My call on port-blindness: hostname-only comparison is deliberate for the bare-host input case (`middleware.ts:52-54`, since platform service-discovery fields hand you a hostname with no scheme), but port-blindness is a side effect of that rather than something I reasoned through. You're right that it isn't a regression, because the pre-existing `hostname === normalizedBoundHost` comparison is equally port-blind. That makes it old, not correct. I'll compare the port when the operator supplies a full origin and leave the bare-host path alone, which closes your `https://home.example.com:8443` case without breaking service discovery.

On the topology, the history is three steps rather than two, and I should lay all three out because the middle one isn't in any PR you've seen.

#2775 as posted was broken. It wired `GITNEXUS_UPSTREAM_URL` into `gitnexus-web` and its comment claimed `docker-server.mjs` proxied `/api/*` to the private server, but nothing in the repo read that variable and `docker-server.mjs` was a pure static file server. `GITNEXUS_BACKEND_URL` was unset, no `window.__GITNEXUS_CONFIG__` was injected, and `ui-constants.ts:5-7` fell back to `http://localhost:4747` in the visitor's own browser. So the proxy topology you're describing was declared there, not implemented.

I then implemented it: 334 lines in `docker-server.mjs` plus 653 lines of tests, with hop-by-hop header stripping, bounded body buffering, retry across an upstream restart, an idle timeout, `X-Forwarded-For` overwritten with a single address the proxy computes itself so a client can't rotate a fake one, and `delete headers.origin` so the API sees a proxied write as Origin-less.

Then I reverted it due to diff size. Roughly 990 lines of proxy and tests against roughly 317 for the public-origin approach, and I wanted the smallest thing you'd have to review.

I see now that was the wrong axis to optimize. The smaller diff moves the change into `gitnexus/src/server`: +95 in `middleware.ts`, +55 in `api.ts`, four touched test files, a widened CORS allowlist and a widened write guard. The proxy version touched `docker-server.mjs`, `Dockerfile.cli` and `render.yaml`, and left the server's trust boundary alone. I handed you fewer lines and more security-critical surface, which is a worse deal.

I'll fix the port comparison regardless of which topology lands, since a self-hoster behind a reverse proxy at `:8443` hits it even if the Render blueprint stops using the variable.

## 2. Deployed end to end

Yes, from a clean account: https://gitnexus-web-0crx.onrender.com. I'd shared it in an email thread with Subham and it belongs here too. `fromService` resolves, both services are live, and the origin allowlist behaves. I'll follow up with the resolved `GITNEXUS_PUBLIC_ORIGIN` value and a completed analyze so the `/data` mount-ownership question closes on evidence rather than on the service being green, since a healthy container doesn't prove uid 1000 can write the mounted volume.

## 4. Hosting recommendation, not a demo

A hosting recommendation. Render can host your public demo too, but the badge exists to say you can run your own GitNexus on Render, so it belongs in the install flow. Which means your blocking item 1 needs the authenticated answer rather than the disclosure answer. So, I'll ship authenticated by default, not a disclosure. That's the proxy-plus-token plan below.

## Azure PAT

I'll remove `AZURE_DEVOPS_URL` and `AZURE_DEVOPS_PAT` from the blueprint. Operators who want Azure DevOps can set both in the dashboard on a deploy they've already decided to secure, and your read-from-source chain on `isAzureDevOpsUrl` matching the whole org is reason enough not to ship it as a one-click prompt.

## My questions

For next steps, should I restore the proxy topology first and then rebase your items 2, 3, 4 and the smaller items onto whatever config surface survives? `gitnexus-server` becomes a `pserv`, the reverted proxy comes back with its tests, `GITNEXUS_SERVE_AUTH_TOKEN` arrives via `generateValue: true` and `fromService`, and I mirror the `http-transport.ts:476` refusal so a wildcard bind with a public origin and no token won't start. Items 2 and 4 are both about documenting and diagnosing `GITNEXUS_PUBLIC_ORIGIN`, and the proxy narrows that variable to single-service deploys, so fixing those docs first means writing them twice.

Two calls I'd like to make if it's okay with you:

The proxy will strip Origin from forwarded requests. Since the guard already passes through any request with no Origin (middleware.ts:97), this means the CSRF guard does nothing for proxied traffic. This way, GITNEXUS_SERVE_AUTH_TOKEN becomes the only thing protecting the server, not a second layer behind the guard. I want that stated plainly now, rather than have someone read "CSRF guard + token" later and assume both are doing work.

The alternative is to forward Origin unchanged and keep GITNEXUS_PUBLIC_ORIGIN set to the web service's origin. That keeps the guard meaningful, at the cost of keeping that variable (and its config surface) in the deploy path, which is exactly what the proxy was meant to remove. I'm going with strip-and-rely-on-the-token unless you'd rather keep the guard alive and pay the config cost.

`GITNEXUS_TRUST_PROXY` stays `1`. Adding a proxying web service would normally make it two hops, but the proxy overwrites `X-Forwarded-For` with one computed address rather than appending, so the count doesn't move. I'll add the behavioural rate-limit test you asked for against that path specifically, with a rotating header proving the limiter still 429s.

On the remaining blocking items: 2, 3 and 4 I'll take as written (with testing), including `proxyaddr.compile` validation with a warn-and-fall-back that names `TRUST_PROXY_ENV`, a bounded numeric branch, the corrected docstring, always emitting the wildcard-bind warning, and rejecting hostnames that can't be real hosts. Same for the smaller items, including the `plan` versus `sizeGB` correction, `autoDeployTrigger: off`, the strengthened trust-proxy assertion via `app.get('trust proxy')`, the `api.ts:135-136` comment, a CI check on the `render.yaml` invariants, and the `SECURITY.md` addition.

If you're aligned with the above, can I send the URLs and token creds (via 1password) to your email? What's the best way to reach you?

magyargergo:
Jumping in since you asked.

On the #2775 history: I went and checked, and you're right. Nothing reads `GITNEXUS_UPSTREAM_URL`, `docker-server.mjs` has no forwarding code, and `DEFAULT_BACKEND_URL` falls back to `http://localhost:4747` in `gitnexus-web/src/config/ui-constants.ts:5-7`. So visitors' browsers would have been hitting their own machines. Thanks for bringing it up yourself, that saved me guessing about it.

On sequencing, I'd split rather than rebase everything. Items 3 and 4, plus your port fix and most of the small stuff, are bugs for anyone self hosting today. They have nothing to do with Render. Put them in their own PR and we can merge that quickly while the topology question is still open. Item 2 is the only one that really needs to wait, since the wording depends on what happens to `GITNEXUS_PUBLIC_ORIGIN`.

The thing I want to ask about before you write the proxy: who supplies the token?

If `docker-server.mjs` holds `GITNEXUS_SERVE_AUTH_TOKEN` and attaches it when forwarding, then `gitnexus-web` is still public with no auth on it. Someone sends `DELETE /api/repo` to the web URL, the proxy adds the token and strips Origin, and the server does the delete. The token would only be stopping people who can't reach the server anyway once it's a `pserv`. So the exposure moves a hop instead of going away.

If you meant the browser supplies it and the proxy just passes it through, then we agree and ignore me. But it needs to be checked at the public edge, whether that's a token the UI asks for or basic auth on the proxy itself. I don't mind which.

Stripping Origin is fine by me given the above. Once the server is private and the proxy is same origin with the UI, the guard isn't doing anything for that traffic, and I'd rather that be written down than have someone assume it is.

`GITNEXUS_TRUST_PROXY=1` also fine, and your reasoning is right about overwriting versus appending. One catch: it puts the trust boundary in `docker-server.mjs`, which is plain `node:http` with none of Express's proxy handling. It has to work out which entry in Render's `X-Forwarded-For` is the real client before it writes the single address it forwards. Take the leftmost and it's spoofable again, just a layer up. So for that rate limit test, send the rotating header to the proxy rather than to the server. Hitting the server directly won't touch the new boundary.

Removing the Azure keys, the port comparison, and the end to end deploy all sound good.

Please don't send tokens or 1Password items, to me or to Abhigyan. We don't need them. If you paste the resolved `GITNEXUS_PUBLIC_ORIGIN` with the secret redacted and an analyze that finished, that answers the `/data` question, which was the only reason it came up.

Separately, and only because it's your account: `https://gitnexus-web-0crx.onrender.com` is in a public comment now and that instance has no auth on it. I'd lock it or take it down. Those subdomains show up in certificate transparency logs anyway, posted or not.

Topology call is Abhigyan's, not mine. For what it's worth I think the proxy is the better shape as long as the edge is actually authenticated.

Ho1yShif:
@magyargergo Thanks for digging into the history yourself, and for the split. Let me know if there's anything else to raise here

## Who supplies the token: the browser

`render.yaml` generates it with `generateValue: true` and it lands as an env var on `gitnexus-web`, but only as the value the proxy compares against. It never gets injected into the page next to `backendUrl`, because then every visitor would have it and we'd be back where you're pointing. The UI asks for it, sends it on `/api/*`, and the proxy checks it with `timingSafeEqual` before forwarding anything. An unauthenticated `DELETE /api/repo` to the web URL gets a 401 at the edge and never reaches the server. The operator reads the generated value out of the Render dashboard.

## Sequencing split

I can do a self-hosting PR outside the Render work. Naming the items you indicated just to make sure we're on the same page:

- Item 3: `resolveTrustProxy` validation via `proxyaddr.compile`, warn-and-fall-back naming `TRUST_PROXY_ENV`, the numeric branch bounded to 1–16, a warn when the resolved value is `true`, `yes`/`no`/`on`/`off` accepted, docstring corrected.
- Item 4: always emit the wildcard-bind warning, branching the message on whether a public origin was configured; log the resolved hostname at `info`; reject derived hostnames that can't be real hosts (`,`, `;`, `*`, whitespace, bare digit).
- The port comparison when the operator supplies a full origin, leaving the bare-host path alone.
- Small items: the `rate-limit.test.ts:273-274` assertion rewritten to a sentinel env value plus `app.get('trust proxy')`; the `api.ts:135-136` comment; the accepted-grammar `not.toThrow()` table; a `logger.warn` spy for `GITNEXUS_PUBLIC_ORIGIN='*'`; the `createPublicOriginMatcher(x) === undefined` table; and `stripPort`'s bracketed-IPv6-with-port case, which is untested. The port fix folds `stripPort` into a single `new URL` parse, so it goes away either way.

What stays on this PR, because each one moves depending on the topology: item 2's wording, item 1's auth, the Azure keys, `README.md:91`'s `plan` versus `sizeGB`, `autoDeployTrigger: off`, the `render.yaml` CI check, and `SECURITY.md`, plus docs changes.

## `X-Forwarded-For`

The reverted proxy already reads the rightmost entry rather than the leftmost, so a client-supplied chain gets discarded and only the address Render appended survives. I'll keep that, and I'll send the rotating header to the proxy rather than to the server for the rate-limit test, since you're right that hitting the server directly doesn't touch the new boundary. The Express-level `1` versus `true` case goes in the self-hosting PR, where it belongs.

Also, I took down the demo; it was just a temporary showcase. No tokens or 1Password items coming your way. Once all the work is ready for review, I'll paste the resolved `GITNEXUS_PUBLIC_ORIGIN` with the secret redacted and a finished analyze so the `/data` question closes on that.

@abhigyanpatwari holding on the proxy until you call the topology.

abhigyanpatwari:
@Ho1yShif Go with the proxy, pserv, docker-server.mjs proxying /api/*, token at the edge. You're right about the axis: 990 lines in the blueprint beats 317 that widen CORS and the write guard in gitnexus/src/server.

Strip Origin, yes. Once the server is private the guard has no traffic left to protect. One condition: the UI sends the token as a header, not a cookie, a cookie rides cross-site requests automatically and Origin is what would have caught that.

Rest of your split works.

magyargergo:
@Ho1yShif If you finished testing this PR please change its status, I we can ship it :) Thanks for your collaboration! 🥇

Ho1yShif:
@magyargergo,

This PR needs several more updates and end-to-end deploy testing! I will let you know as soon as it's ready for final review

README.md
Ho1yShif: Gone. The deploy generates `GITNEXUS_SERVE_AUTH_TOKEN`, the proxy answers `401`
on `/api/*` without it, and the README now says the token is the only control
and that anyone holding it can read every indexed repo. Azure keys are out too.

gitnexus/.env.example
Ho1yShif: Out of this diff. The reworded docs landed in #2820, and the Blueprint no longer
sets the variable at all.

gitnexus/src/server/middleware.ts
Ho1yShif: Out of this diff, fixed in #2820: `proxyaddr.compile` validates the list,
`MAX_TRUST_PROXY_HOPS` bounds the numeric branch, and bad input warns and falls
back to `DEFAULT_TRUST_PROXY`.

README.md
Ho1yShif: Fixed, close to your wording: raise `plan`, which sets RAM (`standard` 2 GB,
`pro` 4 GB), and raise `sizeGB` only if the disk fills with clones and indexes.

render.yaml
Ho1yShif: Both services are `autoDeployTrigger: 'off'` now, and it held on the live
deploy: two pushes have landed on the branch since, neither shipped anything,
and the only rollout was one I triggered by hand.

gitnexus/test/unit/rate-limit.test.ts
Ho1yShif: Out of this diff. The regex on `main` pins the argument through
`TRUST_PROXY_ENV`, via #2820.

gitnexus/src/server/api.ts
Ho1yShif: Out of this diff. The comment went with the reverted `gitnexus/src` changes;
this branch touches no server source.

docker-server.mjs
```diff
@@ -22,18 +25,423 @@
...
+        );
+        setTimeout(() => {
+          // The client may have aborted during the backoff window; don't fire a
+          // fresh upstream request nobody is waiting for anymore.
+          if (res.writableEnded || res.destroyed) return;
+          attempt(n + 1);
+        }, delay);
+        return;
+      }
+      console.error('[gitnexus-web] upstream proxy error:', sanitizeForLog(err.message));
```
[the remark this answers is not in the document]
Ho1yShif: All CodeQL alerts resolved

docker-server.mjs
```diff
@@ -22,18 +25,423 @@
...
+      if (timedOut) return; // 504 already sent by the timeout handler below
+      // Only before any response byte reaches the browser — once headers are
+      // sent the body is partially written and can't be replayed.
+      const retryableError =
+        preConnectRetryCodes.has(err.code) ||
+        (isIdempotentMethod && postConnectRetryCodes.has(err.code));
+      if (retryEligible && !res.headersSent && n < proxyRetryAttempts && retryableError) {
+        const delay = 250 * 2 ** (n - 1); // 250ms, 500ms, ...
+        console.warn(
+          `[gitnexus-web] upstream ${sanitizeForLog(err.code)}; retry ${n}/${proxyRetryAttempts - 1} in ${delay}ms`,
```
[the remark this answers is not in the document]
Ho1yShif: All CodeQL alerts resolved

Ho1yShif:
@abhigyanpatwari @magyargergo this PR is ready for re-review.

This is both validated and deployed end to end on Render. Details and receipts below.

The Blueprint now puts `gitnexus serve` on a private service with no public URL, and the public web service reverse-proxies `/api/*` to it over Render's private network, gated by a generated Bearer token. `gitnexus-server` became a `pserv`; `gitnexus-web` gets `GITNEXUS_UPSTREAM_URL` from `fromService … property: hostport` and `GITNEXUS_SERVE_AUTH_TOKEN` from `generateValue: true`. `autoDeploy: true` became `autoDeployTrigger: 'off'`, so a push cannot restart a server mid-analyze. 

Four variables came out: `GITNEXUS_PUBLIC_ORIGIN` (no job behind the proxy, and since #2820 `assertServeAuthForPublicOrigin` refuses to start with it set), `GITNEXUS_BACKEND_URL` (`docker-server.mjs` falls back to `RENDER_EXTERNAL_URL`, this service's own origin), and `AZURE_DEVOPS_URL` / `AZURE_DEVOPS_PAT`, which shouldn't ship in a one-click prompt for the reason you gave: a token holder could `POST /api/analyze` an Azure URL, spend the operator's PAT, and read the private repo back through the ungated `/api/file`.

The docs have also been updated. README drops "with no API keys required", which described missing authentication as a convenience, and replaces it with where to find the generated token. Both README and a new SECURITY.md section say plainly that the proxy strips `Origin`, so the write-origin guard does nothing for proxied traffic and the token is the only control, not a second layer behind the guard. SECURITY.md lists the routes a token holder reaches ungated and notes `POST /api/mcp` has no `authToken` of its own, a pre-existing gap in `serve` rather than something this deploy introduces. 

Nothing in CI parsed `render.yaml`, so `gitnexus/test/unit/render-blueprint.test.ts` now asserts the couplings the file documents but cannot enforce: `disk.mountPath` equals `GITNEXUS_HOME`, `PORT` matches the port `Dockerfile.cli` binds, the server stays a `pserv`, no `GITNEXUS_PUBLIC_ORIGIN` or `AZURE_DEVOPS_*` key appears anywhere in the document, the token is generated, and the schema pragma is intact. Each was confirmed to fail when its invariant is deliberately broken.

## Receipts

Deployed on a clean workspace by Blueprint sync. `gitnexus-web` runs `746392d9` and `gitnexus-server` runs `be894cbc`; the two commits since are a README edit and a CodeQL sanitizer rewording, neither of which changes what the curl below exercises. Nothing deploys on its own: `autoDeployTrigger: 'off'` held through a hand-triggered web redeploy, and the indexed repo below survived it on the mounted disk.

**Service list.** `gitnexus-server` carries Render's Private Service badge next to `gitnexus-web`'s Web Service badge, both live.
Screenshot 2026-08-05 at 3 21 11 PM


**`gitnexus-web` → Environment.** `GITNEXUS_UPSTREAM_URL` resolved to `gitnexus-server:4747` at first deploy; `GITNEXUS_SERVE_AUTH_TOKEN` generated and redacted here.
Screenshot 2026-08-05 at 3 22 17 PM

**The UI with `render-examples/hermes-render` indexed**, graph rendered. The clone and index landed on the mounted disk at `/data/gitnexus`.
Screenshot 2026-08-05 at 3 23 29 PM
Screenshot 2026-08-05 at 3 23 53 PM

The token gate at the public edge, hostname redacted:

```console
$ curl -s -o /dev/null -w '%{http_code}\n' "https://$HOST/api/repos"                    # no token
401
$ curl -s -o /dev/null -w '%{http_code}\n' -H "Authorization: Bearer wrong-token-padded-to-thirty-two-chars" "https://$HOST/api/repos"
401
$ curl -s -w '\n%{http_code}\n' -H "Authorization: Bearer $TOKEN" "https://$HOST/api/repos"   # generated token
[{"name":"hermes-render","path":"/data/gitnexus/repos/hermes-render","repoPath":"/data/gitnexus/repos/hermes-render","indexedAt":"2026-08-05T22:05:54.044Z","lastCommit":"d3f00670418d7d9eee1c125124dc1dcb7c2bbbd4","stats":{"files":7,"nodes":69,"edges":80,"communities":3,"processes":1,"embeddings":0}}]
200
```

The wrong token is 38 characters, so that `401` comes from the comparison and not from the 32-character minimum. `GET /` returns 200 without a token, since only `/api/*` is gated. The private service's own hostname returns Render's 404: there is no public route to it.

On your `RENDER_EXTERNAL_URL` question: it is populated on a private service, to `https://gitnexus-server.onrender.com`, but that hostname routes nowhere. So the old Blueprint would not have failed with an empty origin and an undefined matcher; it would have failed with a plausible origin pointing at a 404, and since #2820 it now refuses to start with `GITNEXUS_PUBLIC_ORIGIN` set at all. Both variables are gone, and the test keeps them gone.

The disk question is settled by the finished analyze above. If you want the mechanism: the process runs as `uid=1000(node)`, and Render mounts the volume `drwxrwsr-x root:node`, so the shadowed build-time `chown` doesn't matter — `node` writes through the group. `registry.json` and the 12 MB clone are both owned by `node`. I can paste that output if it's useful.

One open item I could not settle from the file alone. The private-network peers came up on `10.193.57.220` and `10.192.105.114`, inside 10/8, so Express's default `uniquelocal` preset would have worked and `GITNEXUS_TRUST_PROXY: 1` isn't doing anything today. I kept the hop count because it holds whatever range Render allocates from, but I'll drop it if you'd rather the deploy leaned on the defaults.

Happy to split this if it's easier to review: the Blueprint rewrite, the README and SECURITY.md wording, and the `render.yaml` test are independent and could land as three PRs in that order. Say the word and I'll break it up.

gitnexus/src/server/api.ts
Ho1yShif: Fixed on main via #2820, both ways you suggested.

`splitAuthority (middleware.ts:158)` rejects a host containing `,, ;, *`, whitespace, or the other authority delimiters, plus a bare digit string and a trailing dot. `cors.test.ts:291` runs your list as a table. `logOriginPolicy` (`middleware.ts:265`) always signals now: `info` with the resolved hostname when the matcher builds, `warn` naming the raw value when it doesn't. `assertServeAuthForPublicOrigin` also refuses to start a `serve` with the variable set and no auth.

One of your cases survives. `2001:db8::1:8443` is a syntactically valid IPv6 address, so it still builds a matcher for `[2001:db8::1:8443]`, and nothing separates it from someone who meant that address. The startup line prints the hostname now, so it's visible rather than silent.

This branch touches no `gitnexus/src`, and `render-blueprint.test.ts` fails if `GITNEXUS_PUBLIC_ORIGIN` appears anywhere in `render.yaml`.

magyargergo:
@Ho1yShif thank you for your submission and your patience! We are looking forward to collaborate more. Your PR will be merged in once the ci is green

magyargergo (approved):

Ho1yShif:
Thanks for merging this PR! Really appreciate the quick turnaround 🙏 I'll reach out to Subham about the best way to launch :)

PR #2806, opened 2026-08-03
perf(mcp): cut the analyze-only language-provider closure out of MCP server startup (#2802)

### LINKED ISSUE #2802: MCP server startup eagerly loads the analyze-only ingestion closure via csv-generator
## Problem

`gitnexus/src/mcp/local/local-backend.ts` pulls the ingestion-layer module closure into **MCP server startup**, which never runs `analyze`.

Chain (all value imports, so all eager):

```
local-backend.ts -> pool-adapter.js -> lbug-adapter.js -> csv-generator.js -> schema.js
schema.js        -> graph-bridge/node-lookup.js, model/symbol-table.js  (+ their transitive closure)
```

`csv-generator.ts` is analyze-only code — it writes the per-pair CSVs the bulk `COPY` consumes. The MCP server never calls it.

## Measured

On this box (WSL2/9p, so absolute numbers are inflated; the ratio is the point):

- the five ingestion modules `schema.ts` adds: **64–77 ms**
- `import schema.js` alone: 34 ms before #2793, 313–386 ms after

## Why it wasn't fixed in #2793

An initial attempt moved `EMBEDDING_INDEX_NAME` into `gitnexus-shared` so `local-backend.ts` would not import `schema.js` directly. **That does not work** — `pool-adapter.js` reaches `schema.js` anyway through the chain above, so the closure loads regardless. The change was reverted; the finding is recorded as a comment at the import site.

Cutting `csv-generator` out of the adapter chain is what would make it real — either by lazy-importing it at the point of use in `lbug-adapter.ts`, or by splitting the constants it needs out of it.

## Note

Before optimising, confirm the win on a native filesystem. Module-load cost is dominated by per-file ESM resolve, which is ~40× worse on a 9p mount than on overlayfs — the same effect that makes `run-analyze.js` take 11 s to import under `/workspace` and 0.25 s under `/tmp`. On a normal Linux CI disk the saving may be small enough not to be worth the coupling change.

magyargergo:
Closes #2802.

## The issue's premise was wrong, and the real cost was next to it

#2802 proposed lazy-importing `core/lbug/csv-generator.ts` out of the `pool-adapter → lbug-adapter` chain, and set its own gate: *"Before optimising, confirm the win on a native filesystem."*

Measured, on overlayfs rather than the 9p mount that inflates everything:

- **`csv-generator` is not worth cutting.** ~1.6 ms marginal beyond the siblings `lbug-adapter.ts` imports anyway. It is also statically imported by `core/search/bm25-index.ts`, which `local-backend.ts` reaches through a dynamic import on the FTS query path — so deferring **relocates** the cost to first query rather than removing it. Recorded at the import site so the avenue is not re-litigated.
- **The real edge was `mcp/local/pdg-impact.ts` → `core/ingestion/languages/index.js`**, dragging the entire 16-language provider registry plus the tree-sitter native binding into every MCP server start, to serve one call that decided whether to append a sentence to an impact note.

| | before | after |
|---|---|---|
| `import mcp/local/local-backend.js` | 565–648 ms / 548 modules | 458–463 ms / 170 modules |

Zero `core/ingestion/languages` modules now load from **any** real MCP entry point (`mcp/server.js`, `cli/mcp.js`, `mcp/local/local-backend.js`, `mcp/resources.js`, `mcp/staleness.js`).

## The fix: delete the language question, do not relocate it

The note was decided by looking up the criterion file's language and comparing against TS/JS. Removing that import is what cuts startup — but the honest fix is deeper: whether return-value ascent can fire is a property of the persisted `CALL_SUMMARY` edges, and the descent already computes it. `pdg-impact.ts` now names no language and imports nothing from the language layer.

That also fixed three defects the language proxy carried:

1. **Wrong for `.mjs` / `.cjs` / `.mts` / `.cts`.** The provider registry's extension arrays omit them while the ingestion pipeline parses them as TS/JS, so those files were harvested but the note claimed their ascent was empty.
2. **Silently stale.** Any language whose harvester began recording formal indices would have kept the caveat until someone remembered to edit a list.
3. **Wrong in reverse.** A TS/JS callee with no return-flow got no caveat, so an ascent that found nothing read exactly like one that covered the slice.

## Review findings, fixed

The branch was reviewed and thirteen findings were fixed, each as its own commit. The ones that changed behaviour:

- **The note asserted a fact an undecodable summary contradicts.** `decodeCallSummary` never throws by design, so a version-skewed, corrupt, or NULL `reason` was indistinguishable from a cleanly-decoded empty one — the note could claim "no formal parameter is recorded as flowing to its return value" about a callee whose summary records exactly that. The scan now reports three outcomes and the note says when summaries could not be decoded. Ascent behaviour is byte-identical; only the wording moves.
- **The note universally quantified over a possibly-incomplete set.** Budget truncation, or an emit-time cap on a block's `calleeIds` cell, can leave a return-flowing callee unexamined. The claim is now qualified when either fired, and unchanged when the set is complete.
- **"Resolved callees" over-claimed.** The count includes ids the descent never enters — out-of-repo targets, interface methods, the `Class:` id a `new X()` emits. Reworded to "call-site callee references"; the number is honest because the `CALL_SUMMARY` scan really does cover all of them.

## Ascent coverage is now published, not just narrated

`impact(mode:'pdg')` computed these facts and used them once, to interpolate English. This is MCP output consumed by agents, so the only way to ask "was the ascent complete, and if not why" was to regex prose — and a pure rewording in this branch broke ~30 assertions, which is the same breakage any external consumer would have hit silently.

Added `pdgEvidence.ascent`: `referencesScanned`, `returnFlowFound`, `undecodableSummaryCount`, `examinedComplete`, `incompleteReasons` (codes, following the existing `truncatedByReasons` precedent), `callSummaryLayerPresent`.

Additive. `pdgResultVersion` stays **2** — the documented trigger is a *breaking* change to the result shape; nothing is removed, renamed, or redefined. Confirmed with zero top-level key drift across 2304 cases. The note prose is byte-identical across those same cases, verified with a negative control proving the harness detects drift.

## Two shipped guards were passing on broken builds

While extracting a shared module-load probe, both existing import-closure tests were demonstrated green on genuinely severed builds:

- `test/integration/mcp/import-closure.test.ts` — severing `dist/cli/mcp.js → stdio-context.js` leaves the `require.cache` diff **empty**, so both assertions reduced to `[].filter(...) === []`. Reported 2 passed.
- `test/integration/optional-grammars/registry-import-closure.test.ts` — severing `registry → swift/query.js` still leaves 76 unrelated CJS entries, satisfying its indirect guard. Reported 1 passed; the Swift half of its headline had gone vacuous.

`dist/` is ESM, so a `require.cache` diff cannot see the first-party `dist/**` graph at all — those tests only ever worked because their headlines happened to be native modules. `test/helpers/module-load-probe.ts` unions `module.registerHooks({ load })` with the cache diff, and makes non-vacuity **structural**: `anchor` and `minModules` are required and the helper throws, so a vacuous probe is a harness failure rather than a silent pass. Both older tests are migrated and now fail on those same builds.

Faster despite doing more: the registry probe goes 12.4 s → 6.75 s, because `spawnSync` burned the parent thread polling while the child loaded native grammars.

## A pre-existing resolver gap, now pinned at the right level

Chained receiver calls reach the PDG for most receiver shapes, but **not when a field's type must be inferred from its initializer** — there the whole `calleeIds` cell empties, losing even the first ordinary named-receiver link. The discriminator is the type annotation, not local-versus-field: an annotated field resolves fine, including the constructor-assigned form.

```
const o = new Outer()                     resolves
private p: Outer = new Outer()            resolves
private p: Outer;  this.p = new Outer()   resolves
constructor(private p: Outer) {}          resolves
makeOuter().inner().compute()             resolves
o.inner().mid().compute()  (three links)  resolves
private p = new Outer()                   NO EDGES
private p;  this.p = new Outer()          NO EDGES
```

Pre-existing and independent of this change. Pinned in `test/integration/resolvers/` as CALLS-edge facts, next to the equivalent Python test, so whoever closes it gets the signal in the suite they are already running. The pin is self-diffing: simulating the fix fails with the newly-resolved ids listed. Worth noting for whoever picks it up — the initializer is not invisible to the resolver, `new Outer()` still emits its own constructor edge, so only the initializer-to-field-type binding is missing.

## Verification

- Full unit suite: 11,283 passed, zero assertion failures. The 39 failures are worker-startup timeouts on the dev host and route through `runPipelineFromRepo`; the pre-fix baseline was 41 of the same class.
- Changed surface: 383 unit + 17 integration passing.
- Every fix carries a mutation proof — revert the source, watch the new tests go red, restore. That is what caught a test suite unable to detect the defect it existed to pin.

---

## Follow-up in this PR: the group extractor closure

After the language registry came off the startup path, `core/group` was what remained, and it was the larger of the two.

`core/group/service.ts` statically imported `./sync.js`, which pulls all six contract extractors, five of which statically import the native `tree-sitter` binding. Only `group_sync` ever needs them. The other seven group tools do not, and neither does a server that never syncs.

One lazy `await import(...)` at the single call site, inside a method that was already `async`:

| entry | before | after |
|---|---|---|
| `dist/mcp/server.js` | 521 ms | **135 ms** (−74%) |
| `dist/mcp/local/local-backend.js` | 453 ms | **66 ms** (−85%) |
| tree-sitter modules at both | 11 | **0** |

Measured on a native filesystem across 5 cold runs per entry; `/workspace`-style 9p mounts inflate ESM resolve and are not a valid measurement surface.

No signature change, no async ripple, no change to `local-backend.ts`. The pattern was already established on this exact module — the CLI's sync command lazy-imports `sync.js` the same way, so `service.ts` was the outlier.

The cost is moved, not deleted: the first `group_sync` pays the module load. That is the right trade, since `group_sync` is already long-running and sessions that never sync pay nothing.

The `#2802` startup guard gains a sibling forbidden-pattern case covering `core/group/extractors/` and the native parser, reusing the probes it already collects. Verified load-bearing: restoring the static import turns `dist/mcp/server.js` red and names all seven extractors.


magyargergo:
The resolver gap this branch pins is now tracked separately as #2807, with a minimal repro.

Worth noting for whoever picks it up: the field initializer is not invisible to the resolver. `new Outer()` still emits its own constructor CALLS edge, identical to the annotated twin, so only the step binding that initializer's type back to the field is missing.

The two pins added here are self diffing. When #2807 is fixed they will fail and print the newly resolved ids, which is the cue to update them rather than a regression.

gitnexus/src/mcp/local/pdg-impact.ts
```diff
@@ -831,6 +986,21 @@ function assemblePdgImpactResult(input: {
...
+  // The empty-ascent sentence quantifies UNIVERSALLY over the references the
+  // descent actually EXAMINED, and two mechanisms can make that set a strict
+  // subset of the slice's real callee list (rationale: AscentCoverage
+  // INCOMPLETENESS, vocabulary: PdgAscentIncompleteReason). Computed HERE, once,
+  // and consumed twice — by `pdgEvidence.ascent` (structured, the contract) and by
+  // the note's qualifier clause (prose, rendered through
+  // ASCENT_INCOMPLETE_PHRASE). Deriving both from one array is what stops an agent
+  // branching on the codes and a human reading the note from ever disagreeing.
+  const ascentIncompleteReasons: PdgAscentIncompleteReason[] = [];
+  if (input.truncated) ascentIncompleteReasons.push('traversal-truncated');
```
magyargergo: **P1 — `examinedComplete: true` published over a callee set that was never fully examined.** `[reproduced]`

`'traversal-truncated'` is derived solely from `input.truncated`, but the descent throws away the depth-exhaustion result of every BFS it runs *inside* a callee. `bfsReachableBlocks` sets `truncatedByDepth` at `:1582` when its frontier is still non-empty at the budget; both merge sites fold only the row-limit flag:

```ts
:1940  if (ascent.truncatedByLimit) truncatedByLimit = true;   // ascent re-seed
:2032  if (bfs.truncatedByLimit)    truncatedByLimit = true;   // per-callee span BFS
```

The asymmetry looks unintended: the top-level intra BFS's copy of that same flag *is* propagated — `let truncatedByDepth = intra.truncatedByDepth;` at `:2243` — and the per-callee clamp is documented as "mirrors the top-level intra BFS depth clamp". Only one of the two mirrors is reported.

So `truncated` stays false, `'traversal-truncated'` never fires, and `:1149` publishes `examinedComplete: true, incompleteReasons: []` while the note makes its unqualified universal claim — the exact state this code was added to prevent.

**Reproduced at `maxDepth: 3`, the shipped default.** `run` calls `helper`; `helper`'s body is a 5-block chain `C0→C1→C2→C3→C4`; `C4` calls a callee carrying a genuine `encodeCallSummary([0])`. The per-callee BFS clamps at 3 hops, so `C4` is never reached or scanned:

```
truncated: undefined
reachableBlocks: [".:5:0:0", ".:5:0:1", ".:5:0:2", ".:5:0:3"]   <- C4 absent
ascent: {"referencesScanned":1,"returnFlowFound":false,"examinedComplete":true,
         "incompleteReasons":[],"callSummaryLayerPresent":true}
```

Any downstream `impact({mode:'pdg'})` where a descended callee has a dependence chain longer than `maxDepth` block-hops hits this, and coalesced straight-line blocks chain readily — the common case, not a corner. `tools.ts:452` tells agents to branch on these fields, and the false all-clear rides on the negative claim `returnFlowFound: false`, which is precisely what `examinedComplete` exists to qualify.

**Fix:** propagate the dropped flags at `:1940`/`:2032`. If the per-callee clamp deliberately must not raise result-level `truncatedBy:'depth'`, thread a separate descent-side flag into `AscentCoverage` and add a third `PdgAscentIncompleteReason` — but either way `examinedComplete` should stop deriving from `input.truncated`, which measures a different thing. Blast radius worth deciding deliberately: the direct fix newly stamps `truncatedBy:'depth'` on results that previously reported none.

**Test gap:** `INCOMPLETENESS_PREMISES` (`impact-pdg-ascent-note.test.ts:400-414`) covers only the *top-level* intra depth budget (`maxDepth: 1`) — whose flag *is* folded — plus the emit cap. No row truncates inside a callee's own BFS.

The root-cause lines are pre-existing; the false completeness claim is new in this PR. Note this finding is **single-engine** — see the digest for why it is still filed P1.

gitnexus/test/integration/mcp/startup-language-closure.test.ts
```diff
@@ -0,0 +1,144 @@
...
+ * source file that merely mentions the word cannot satisfy or trip this.
+ */
+const FORBIDDEN_GROUP_RE = /(^|\/)core\/group\/extractors\/|(^|\/)node_modules\/tree-sitter/;
+
+// Observed on Node 22.18 against a clean build: server.js loads 489 distinct
+// modules, local-backend.js 265, cli/mcp.js 4. The floors sit well below those
+// so normal dependency churn doesn't trip them, while a probe that silently
+// loaded nothing still fails. Each anchor names a module on an edge whose loss
+// would make the offender assertion below vacuous.
+const ENTRIES = [
```
magyargergo, line 157: **P2 — the new group-extractor policy has no anchor on the edge it polices, and one of its three rows is already vacuous.** `[reproduced]`

The helper's own guidance is "prefer an anchor on the same edge the test polices — one whose disappearance would make the headline assertion meaningless." Commit `83e8cf7c5` added a second, independent policy (`FORBIDDEN_GROUP_RE`, asserted at `:128-143`) to these same three entries without adding an anchor for it. Measured against the current build:

```
mcp/server.js:              group/service.js=true   core/group/ modules=5
mcp/local/local-backend.js: group/service.js=true   core/group/ modules=5
cli/mcp.js:                 group/service.js=false  core/group/ modules=0
  cli/mcp.js full closure: dist/cli/mcp.js, dist/mcp/stdio-context.js,
                           node:async_hooks, dist/mcp/stdio-capture.js
```

- **`cli/mcp.js`'s group row is already vacuous.** Four leaf modules, no `core/group/` module at all — it cannot fail for any reason related to the policy, and its anchor `dist/mcp/stdio-context.js` is satisfied regardless. This is the exact vacuity class `module-load-probe.ts` was written to make impossible, reappearing in the same PR's newest commit, because `anchor`/`minModules` prove the probe reached *a* graph, not the graph the policy is about.
- The other two rows are meaningful today, but anchor on `dist/mcp/resources.js` and `dist/mcp/local/pdg-impact.js` — neither on the `→ group/service.js → group/sync.js` chain. Sever `local-backend → group/service` (the obvious next lazy-load step, exactly what `groupSync` just did to `sync.js`) and the group assertion goes vacuous on **all three** rows with every anchor and floor still green.

Suggested: add `dist/core/group/service.js` as a second anchor on the two entries that legitimately reach it, and either drop the `cli/mcp.js` row from the group policy or note in the comment that it is a deliberate negative control.

gitnexus/test/integration/mcp/startup-language-closure.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  { entry: 'mcp/server.js', anchor: 'dist/mcp/resources.js', minModules: 100 },
+  { entry: 'cli/mcp.js', anchor: 'dist/mcp/stdio-context.js', minModules: 3 },
+  {
+    entry: 'mcp/local/local-backend.js',
+    anchor: 'dist/mcp/local/pdg-impact.js',
+    minModules: 50,
+  },
```
magyargergo, line 172: **P2 — `mcp/http-transport.js` is a real MCP startup entry, the largest one, and no probe covers it.** `[reproduced]`

Measured **523 modules** — 136 more than `mcp/server.js` — since it adds express, cors, and the SDK's Streamable-HTTP/SSE transports on top of the whole `server.js` graph. It is what `gitnexus mcp --http` starts, and the hosted-deploy path.

To be precise about what this is: the PR's substantive claim **holds** here too — measured 0 language modules, 0 group extractors, 0 tree-sitter. This is a guard-**coverage** gap, not a broken claim. But the dependency runs `http-transport → server.js` and never the reverse, so a future static import inside `http-transport.ts`, or in anything only it reaches, would re-add the regression with the entire guard green.

`{ entry: 'mcp/http-transport.js', anchor: 'dist/mcp/server.js', minModules: 100 }` costs ~0 wall clock — the probes already run concurrently, and this file's own header notes the marginal cost of an extra entry is nil.

Related, same list: `mcp/resources.js` and `mcp/staleness.js` are named in the PR body's headline claim but are not probed entries. (Both measure clean — `resources.js` is 57 modules, 0 offenders — so this is coverage, not a defect.)

gitnexus/src/core/group/service.ts
```diff
@@ -338,6 +341,12 @@ export class GroupService {
         return { error: `Group "${name}" not found. Run group_list to see configured groups.` };
       throw err;
     }
+    // Lazy: `sync.js` reaches the six contract extractors and the native
+    // tree-sitter binding. `groupSync` is the ONLY consumer — the other seven
+    // group tools never need it — so deferring it here keeps that closure off
+    // MCP server startup entirely and off every non-sync group call. The CLI
+    // already does exactly this at `cli/group.ts`'s sync command.
+    const { syncGroup } = await import('./sync.js');
```
magyargergo, line 349: **P2 — this is the PR's one changed control-flow line, and no test reaches it.** `[code-read]`

Every test that references `groupSync`/`group_sync` stops short of this line:
- `test/unit/group/service.test.ts` returns at the empty-name guard
- `test/integration/group/group-service-not-found.test.ts` mocks `loadGroupConfig` to reject — its `syncGroupMock` is declared and never invoked
- `test/integration/group/group-sync.test.ts` imports `syncGroup` **directly**, bypassing `GroupService` entirely
- the new startup guard asserts only the **negative** (that `sync.js` is absent at startup)

`tsc --noEmit` catches a path typo, which caps the risk — but nothing verifies the import resolves and hands off correctly in the built ESM `dist/` when `group_sync` is actually called, and every production invocation of that tool hits this line.

For balance: three separate lanes confirmed the surrounding design is sound — `syncGroup` really is the only consumer (grep-verified across `src/`), the import sits after config resolution so nothing is read before assignment, `cli/group.ts:160` is an established precedent for the same pattern, concurrent calls dedupe on Node's module registry, and a load failure now surfaces as a scoped `isError: true` MCP tool error rather than killing the server at startup. The gap is coverage of the new call site, not the design.

One integration test that drives `GroupService.groupSync` through to a successful sync against the built `dist/` would close it.

gitnexus/src/mcp/tools.ts
```diff
@@ -449,7 +449,7 @@ MODE (opt-in): "callgraph" (default) walks symbol→symbol edges (CALLS/IMPORTS/
...
...nvelope (id/name/type/filePath), affectedStatements, affectedStatementCount, interproceduralByDepth/pdgInterprocedural for cross-function reach, compatibility byDepth/byDepthCounts, risk:'UNKNOWN', and a note describing the unified contract. Degraded PDG results (no-layer, sub-layer-missing, unknown) keep mode:'pdg', pdgResultVersion:2, target metadata when the target resolves, risk:'UNKNOWN', note/remediation, and empty byDepth parity fields — never a false-safe zero. If depth and limit both bound the slice, truncatedByReasons reports both causes while truncatedBy remains scalar. Return-value-ascent coverage is published structurally at pdgEvidence.ascent — present iff the inter-procedural descent ran — with referencesScanned (call-site callee references scanned for a CALL_SUMMARY), returnFlowFound (whether the ascent fired anywhere in the slice), undecodableSummaryCount, examinedComplete, incompleteReasons ('traversal-truncated' | 'callee-list-capped'), and callSummaryLayerPresent (false ⇒ a pre-CALL_SUMMARY index, where returnFlowFound:false says nothing about the callees). Branch on those fields; the note narrates the same facts in prose for humans and is not a stable contract.
```
magyargergo: **P3 — "present iff the inter-procedural descent ran" is not true of the empty-slice return.** `[reproduced]`

This sentence, and the type doc's "Absent ⇒ nothing was scanned", are both contradicted by `runImpactPDG`'s `reachableBlocks.length === 0` early return (`pdg-impact.ts:2396`), which returns **before** `assemblePdgImpactResult` and carries no `pdgEvidence` at all — after the descent has already run and scanned. Reproduced independently by two lanes:

```
--- direction=downstream
   reachableBlocks: []      CALL_SUMMARY ids scanned: 1
   has pdgEvidence: false   ascent: undefined
```

So `ascent: undefined` conflates "upstream — never scanned" with "downstream — scanned one reference and found nothing", which is exactly the distinction the field was introduced to make.

Trigger: `impact({mode:'pdg', line:N})` on a statement whose only dependents are interior to its own coalesced block and whose callees resolve to no CFG body — an out-of-repo target, an interface method, or the `Class:` id a `new` expression contributes (the PR's own `UNENTERABLE_CALLEES` fixture shape).

Fix: emit `pdgEvidence: { ascent: … }` on that return too, or narrow this sentence to "present iff the descent ran **and** the slice is non-empty". Anchoring here rather than on the return itself because this line is the contract an agent actually reads.

gitnexus/src/mcp/local/pdg-impact.ts
```diff
@@ -876,21 +1046,57 @@ function assemblePdgImpactResult(input: {
...
+        // summaries record: an undecodable `reason` may well encode a return-flow
+        // this reader cannot unpack (`decodeCallSummary` never throws, so a
+        // version-skewed / corrupt / NULL reason is otherwise indistinguishable
+        // from a cleanly-decoded empty one). The ascent is withheld either way;
+        // only the tail that explains it changes.
         noteParts.push(
-          `return-value ascent is currently TypeScript/JavaScript-only (only the TS/JS harvester ` +
-            `records the formal-index it needs), so a caller statement depending on a non-TS/JS ` +
-            `callee's RETURN value is not in the slice. Descent and the intra slice are unaffected.`,
+          `no return-value ascent in this slice: none of the ${references} call-site ` +
```
magyargergo: **P3 — `referencesScanned` is a distinct-callee tally, rendered on both surfaces as a call-site count.** `[reproduced]`

`calleeReferencesSeen` is a `Set<string>` of callee ids (`:1873`, accumulated at `:1905`, published at `:2062`), so the value is a distinct-id tally. But this note renders it as `none of the ${references} call-site callee reference(s)`, and `tools.ts:452` describes it as "call-site callee references scanned". The internal doc contradicts itself on the same line — "the distinct-id tally … **CALL SITES**".

Reproduced: two distinct slice blocks each calling `helper` report `referencesScanned: 1` and "none of the 1 call-site callee reference". Any slice invoking the same callee from more than one block — extremely common — leads a reader to under-estimate what was checked.

Fix: rename to `calleesScanned` / "distinct callee(s)" on both surfaces, or count occurrences if the call-site reading was intended. Note the existing tests pin the current wording, so a fix has to move them.

**Test gap:** `referencesScanned` is never exercised with two call sites to the same callee — one existing block varies ids *within* a cell, another varies callees *across* hops, and neither can distinguish the two readings.

gitnexus/src/mcp/local/pdg-impact.ts
```diff
@@ -1370,86 +1592,105 @@ interface CalleeSpan {
...
   // the aliased cells via bracket access — no per-field `as any`.
   for (const r of rows as Array<Record<string, unknown>>) {
     const blockId = String(r['id'] ?? '');
     if (!blockId) continue;
-    const calleeIds = splitCalleeIds(r['calleeIds']);
+    // ONE pass over the cell classifies BOTH facts — a second full split just to
+    // re-test the sentinel doubled the per-row parse cost.
+    const { ids: calleeIds, truncated } = parseCalleeIdsCell(r['calleeIds']);
+    if (truncated) calleeListTruncated = true;
     if (calleeIds.length > 0) out.push({ blockId, calleeIds });
```
magyargergo, line 1795: **P3 — a slice block whose `calleeIds` cell is empty silently shrinks the population `examinedComplete` quantifies over.** `[code-read]`, confidence ~60

`calleeIdsOfBlock` returns `''` for a whole file when the resolved-id map is absent (`core/ingestion/cfg/emit.ts:350` — `if (fileMap === undefined) return '';`). This line then drops the row entirely, and because no sentinel is written into an empty cell, `calleeListTruncated` never rises.

Such a block can still carry real call sites in its name-keyed `callees` cell — the statement bridge relies on exactly that. So a mixed slice (criterion file has a fileMap, a hop-1 callee's file does not) crosses a hop, publishes `examinedComplete: true`, and quantifies universally over a population that silently excludes every call site in the id-less blocks.

Flagged as `[code-read]` because the consumer half is mechanical from those two lines, but the producer half could not be constructed without a real index build.

Fix: treat "block in the descent's slice with a non-empty `callees` cell but an empty `calleeIds` cell" as a third incompleteness reason — or, at minimum, document the population as "callees the index recorded resolved ids for" rather than the slice's whole callee list.

gitnexus/test/integration/mcp/startup-language-closure.test.ts
```diff
@@ -0,0 +1,144 @@
...
+ * stack on every MCP server start. Measured cost of that one edge on a native
+ * filesystem: `dist/mcp/server.js` 521 ms -> 133 ms, `local-backend.js` 453 ms
+ * -> 66 ms.
+ *
+ * Matching the parser by its package prefix rather than a bare substring so a
+ * source file that merely mentions the word cannot satisfy or trip this.
+ */
+const FORBIDDEN_GROUP_RE = /(^|\/)core\/group\/extractors\/|(^|\/)node_modules\/tree-sitter/;
+
+// Observed on Node 22.18 against a clean build: server.js loads 489 distinct
```
magyargergo: **P3 — these recorded baselines are stale by exactly the modules this PR's own second commit removed.** `[reproduced]`

The comment states "server.js loads 489 distinct modules, local-backend.js 265, cli/mcp.js 4" (and `:36-37` repeats the 265/489 pair). Measured against the current build at HEAD:

| entry | comment says | measured |
|---|---|---|
| `mcp/server.js` | 489 | **387** |
| `mcp/local/local-backend.js` | 265 | **163** |
| `cli/mcp.js` | 4 | 4 |

Deltas of exactly 102 each — the group-extractor closure that commit `9ea9676dc` removed, i.e. the numbers were recorded before this branch's own later commit.

Nothing breaks: the floors (100 / 3 / 50) still hold comfortably, and the subset relation the header claims still holds exactly. But the numbers are presented as the observation *justifying* those floors, so the next author calibrating a floor would diff against a baseline that never existed at HEAD. (`registry-import-closure.test.ts`'s "553 distinct modules" is accurate, for contrast.)

gitnexus/test/integration/cfg/pdg-chained-receiver-callees.test.ts
```diff
@@ -0,0 +1,297 @@
...
+/** The behaviour a `reaches-pdg` row has today. */
+function assertChainReachesPdg(shape: ReceiverShape): void {
+  const ids = idsFor(shape.marker);
+  // Non-empty first: an unresolvable receiver drops EVERY link, so this
+  // separates "the chained link regressed" from "the whole cell went away".
+  expect(ids).not.toHaveLength(0);
+  expect(ids).toEqual(expect.arrayContaining([...shape.links]));
+}
+
+describe('PDG calleeIds — chained receiver calls by receiver form (#2802 follow-up)', () => {
```
magyargergo: **P3 — the pinned gap cites this PR rather than its tracking issue.** `[reproduced]`

This describe block (and `test/integration/resolvers/typescript-inferred-field-receiver.test.ts:254`) names the gap `(#2802 follow-up)`. But **#2807** exists for it — open, labeled bug — and **PR #2810 already addresses it**. After merge, the gap is discoverable only by reading a `KNOWN GAP` marker in a test file rather than from the issue tracker.

Cite #2807 in both describe blocks and in the header comment's "Worth noting for whoever picks it up" sentence.

**Merge-order note:** these pins assert the gap *exists*, so if #2810 lands first this PR needs a rebase and a pin update to clear CI. #2810 is currently BLOCKED, so it isn't live today — but the intended order (2806, then 2810) is worth stating somewhere durable.

The pins themselves are good and this is not a criticism of them: both deliberately reject `it.fails` in favour of exact `toEqual` values with a non-vacuity probe, so a fix to #2807 turns them red rather than silently green — which is exactly what they are for.

magyargergo (commented):
# Multi-tool review digest — PR #2806

**Head** `83e8cf7c5` · **Base** `main` (merge-base `7468cc9`) · 2445+/322− across 16 files (5 source, 11 test)

## How this review was produced (read this before weighting anything below)

| Leg | Engine | Status |
|---|---|---|
| GitNexus PR swarm — 6 lanes (pr-facts-historian, risk-architect, test-ci-verifier, docs-dod, branch-hygiene, security-boundary) | **Claude** | ran |
| `ce-code-review` personas — 7 lanes (correctness, adversarial, api-contract, maintainability, performance, project-standards, reliability) | **Claude** | ran |
| Codex / GPT-5.5 xhigh — solo review, `ce-code-review`, adversarial | Codex | **entirely unavailable** — account usage limit (`try again at Aug 9th`). Confirmed account-level, not transport-level. |
| Substitute second engine — adversarial + swarm | **Composer** (Cursor's routed model) | ran, read-only |
| Coordinator's own verification | executed against real code and a real build | ran |

**The intended second engine is missing.** Cross-engine corroboration in this run means *a Composer lane and a Claude lane independently reaching the same finding* — the strongest class available, and weaker than the Codex leg it replaces. Agreement among Claude lanes alone is **consistent within one engine, not independent confirmation**, and can be a shared false positive. Nothing here is corroborated by Codex.

One further caveat on the substitute: only the Composer **adversarial** leg self-declares its engine (`ENGINE: Composer`). The Composer **swarm** leg carries no declaration line, so findings resting on it alone are marked "Composer (swarm, undeclared)" and read one notch weaker.

Finally: **the PR author and the reviewing account are the same person.** A Claude-only finding here carries no independent perspective beyond execution evidence, so `[reproduced]` findings are weighted far above `[code-read]` ones throughout. The single most decisive lane in this run — correctness — drove the real `runImpactPDG` through seven throwaway probes rather than reading it, and it both **confirmed the headline defect at the production default** and **refuted three suspicions other lanes had filed as findings.**

---

## What the review validated (not just "looks fine")

These were attacked and measured, not assumed:

- **The headline startup win is real, at every entry point measured.** Zero `core/ingestion/languages/` modules and zero `core/group/extractors/` or `node_modules/tree-sitter` modules load from `mcp/server.js` (387 modules), `cli/mcp.js` (4), `mcp/local/local-backend.js` (163), `mcp/resources.js` (57) — **and** from the unprobed `mcp/http-transport.js` (523). Independently, the performance lane built BASE from scratch and measured `server.js` 1029→494 modules / ~1.06 s→~0.17 s and `local-backend.js` 734→199 / ~0.94 s→~0.08 s on a fast filesystem. The fix **deleted** the language lookup rather than deferring it, so there is no lazy path for it to return through, and no language name survives anywhere in `pdg-impact.ts`.
- **The probe harness's non-vacuity guards genuinely fire.** `probeModuleLoads` was run directly: a bad anchor throws, a floor of 99999 throws, and both throw from the helper *before* any policy `expect` runs. The child-death → "loaded zero modules" trivial-pass class is structurally closed.
- **The self-diffing gap pins are exactly right.** Neither new resolver/PDG pin uses `it.fails`; both explicitly reject it in comments and assert exact `toEqual` values with a non-vacuity probe folded in. A fix to #2807 turns them **red**, which is what they are for.
- **Zero assertions were deleted** in the three trimmed CFG tests and the two migrated closure tests: 0 removed `expect(` lines, 0 removed `it(`/`test(` lines. The net deletions are temp-dir lifecycle boilerplate only.
- **The new note's grammar and its strongest claim hold.** All 16 renderings were driven through real code (`references` ∈ {1,3} × `undecodable` ∈ {0,1} × {complete, truncated, capped, both}); every sentence is grammatical and true. "Every summary examined decoded" **is** guaranteed when `undecodableSummaryCount === 0`, because `undecodable` accumulates over exactly the rows the scan decoded.
- **Query surface did not get more expensive.** `calleeIdsByBlock` and `calleesWithReturnFlow` are byte-identical Cypher before and after; neither carries a `LIMIT`, so every id handed to the scan is genuinely scanned. `parseCalleeIdsCell` is byte-equivalent to the old `splitCalleeIds` on its `ids` output.

---

## Findings

### P1 — `examinedComplete: true` with an unqualified universal note, while a return-flowing callee sat past a depth clamp the descent silently discards

`gitnexus/src/mcp/local/pdg-impact.ts:999` and `:1149` (root cause at the unchanged `:1940` and `:2032`)

The new completeness vocabulary defines `'traversal-truncated'` as *"the traversal stopped at its depth/size budget, so a callee that DOES carry a return-flow can sit in a hop never reached"*, and its only input is the result-level `truncated` flag. But the descent throws away the depth-exhaustion result of every BFS it runs **inside** a callee. `bfsReachableBlocks` sets `truncatedByDepth` at `:1582` whenever its frontier is still non-empty at the budget; both merge sites fold only the row-limit flag:

```ts
// :1940 (ascent re-seed)      // :2032 (per-callee span BFS)
if (ascent.truncatedByLimit) …  if (bfs.truncatedByLimit) …
```

The asymmetry is unexplained — the top-level intra BFS's copy of that same flag **is** propagated (`let truncatedByDepth = intra.truncatedByDepth;` at `:2243`), and the per-callee clamp is documented as *"mirrors the top-level intra BFS depth clamp"*. Only one of the two mirrors is reported. So `truncated` stays false, `'traversal-truncated'` never fires, and the result publishes `examinedComplete: true, incompleteReasons: []` while the note makes the unqualified universal claim — exactly the state this code was added to prevent.

**Reproduced end-to-end at `maxDepth: 3`, the shipped production default.** Criterion `run` calls `helper`; `helper`'s body is a 5-block dependence chain `C0→C1→C2→C3→C4`; `C4` calls a callee carrying a genuine `encodeCallSummary([0])` return-flow. The per-callee BFS clamps at 3 block-hops, so `C4` is never reached and never scanned:

```
truncated: undefined | truncatedBy: undefined
reachableBlocks: [".:5:0:0", ".:5:0:1", ".:5:0:2", ".:5:0:3"]     <- C4 absent
ascent: {"referencesScanned":1,"returnFlowFound":false,"undecodableSummaryCount":0,
         "examinedComplete":true,"incompleteReasons":[],"callSummaryLayerPresent":true}
note:  "…none of the 1 call-site callee reference carries a CALL_SUMMARY return-flow.
        So a caller statement depending on a callee's RETURN value is not in the slice.
        Every summary in this slice decoded, so this is a property of the persisted summaries…"
```

Identical at `maxDepth: 2`. A second Claude lane reproduced the same mechanism through a separate probe.

**On confidence, stated plainly: this finding is single-engine.** Both reproductions are Claude lanes. Neither Composer leg found it — the Composer swarm leg examined this exact question and recorded *"Could not find under-report where `truncated===false` and `listTruncated===false` but a return-flowing callee in a reached cell was skipped"*, and the Composer adversarial leg's own P1 is the different (ascent-expansion) mechanism filed as P3 below. So there is **no cross-engine corroboration for this finding.** It is presented as P1 anyway because two independent probes reproduce it end-to-end at the shipped default and every link in the chain is checkable in the source in under a minute — not because lanes agreed.

**Trigger:** any downstream `impact({mode:'pdg'})` where a callee entered by the descent has an intra-procedural dependence chain longer than `maxDepth` block-hops from its seed block. The default is 3 and coalesced straight-line blocks chain readily — **this is the common case, not a corner.** The false all-clear rides on the negative claim (`returnFlowFound: false`), which is precisely the reading `examinedComplete` exists to qualify. `mcp/tools.ts:452` tells agents to branch on these fields.

**Fix:** propagate the dropped flags — `if (ascent.truncatedByDepth) truncatedByDepth = true;` near `:1940` and `if (bfs.truncatedByDepth) truncatedByDepth = true;` near `:2032`. If the per-callee clamp is deliberately not meant to raise the result-level `truncatedBy:'depth'`, thread a separate descent-side flag into `AscentCoverage` and add a third `PdgAscentIncompleteReason`. Either way, `examinedComplete` should stop being derived from `input.truncated`, which measures a different thing. Note the blast radius: the direct fix newly stamps `truncatedBy:'depth'` on results that previously reported none — correct, but a visible output change worth deciding deliberately. The root cause lines are pre-existing; **the false completeness claim is new in this PR.**

**Test gap:** `INCOMPLETENESS_PREMISES` (`test/unit/impact-pdg-ascent-note.test.ts:400-414`) covers only the *top-level* intra BFS depth budget (`maxDepth: 1`), whose flag **is** folded, plus the emit cap. No row truncates inside a callee's own BFS. The reproduction probe is a ready-made fixture.

### P2 — the new group-extractor policy has no anchor on the edge it polices, and one of its three rows is already vacuous

`gitnexus/test/integration/mcp/startup-language-closure.test.ts:85-93` (anchors) vs `:128-143` (the group assertion)

The shared helper's own guidance is *"prefer an anchor on the same edge the test polices — one whose disappearance would make the headline assertion meaningless."* Commit `83e8cf7c5` added a second, independent policy (`FORBIDDEN_GROUP_RE`) to the same three entries without adding an anchor for it. Measured today:

- `cli/mcp.js` loads **4** modules and reaches **no** `core/group/` module at all — its group row cannot fail for any reason related to the policy, and its anchor (`dist/mcp/stdio-context.js`) is satisfied regardless. **Already vacuous.** This is the exact vacuity class `module-load-probe.ts` was written to make impossible, reappearing in the same PR's newest commit.
- `server.js` and `local-backend.js` do carry `core/group/service.js` (5 modules each), but anchor on `dist/mcp/resources.js` and `dist/mcp/local/pdg-impact.js` — neither on the `→ group/service.js → group/sync.js` chain. Sever `local-backend → group/service` (the obvious next lazy-load step, exactly what `groupSync` just did to `sync.js`) and the group assertion goes vacuous on **all three** rows with every anchor and every floor still green.

Suggested: add `dist/core/group/service.js` as a second anchor on the two entries that legitimately reach it, and either drop the `cli/mcp.js` row from the group policy or state in the comment that it is a deliberate negative control.

### P2 — `mcp/http-transport.js` is a real MCP startup entry, the largest one, and is unprobed

`gitnexus/test/integration/mcp/startup-language-closure.test.ts:85-93`; entry point at `gitnexus/src/cli/mcp.ts` (`--http` path)

Measured **523 modules** — 136 more than `mcp/server.js` — because it adds express, cors, and the SDK's Streamable-HTTP/SSE transports on top of the whole `server.js` graph. It is what `gitnexus mcp --http` starts, and the hosted-deploy path.

To be precise about what this is: the PR's substantive claim **holds** here too (measured 0 language modules, 0 group extractors, 0 tree-sitter). This is a guard-**coverage** gap, not a broken claim. The dependency runs `http-transport → server.js`, never the reverse, so a future static import inside `http-transport.ts` — or in anything only it reaches — would re-add the ~226-module regression with the entire guard green. Adding `{ entry: 'mcp/http-transport.js', anchor: 'dist/mcp/server.js', minModules: 100 }` costs ~0 wall clock; the probes already run concurrently.

### P2 — the new spawn-based guard never runs on the Windows/macOS CI matrix

`gitnexus/scripts/cross-platform-tests.ts:162-189` (unchanged by this PR)

`TESTING.md` requires platform-sensitive tests to be registered in `SPAWN_CLI`. That list has 16 entries and includes none of `startup-language-closure.test.ts` (new here), `import-closure.test.ts`, or `registry-import-closure.test.ts` (both rewritten here onto the same spawn-based probe). The Windows-safety plumbing this PR added *specifically for this probe* — `path.sep`→POSIX normalisation, `pathToFileURL`, `NODE_OPTIONS` clearing, array-form `spawn` — is therefore exercised only on Ubuntu. The latter two files were already absent before this PR; the new file is new debt.

### P2 — `pdg-impact.ts` still eagerly evaluates the analyze-only CFG graph to obtain two string constants

`gitnexus/src/mcp/local/pdg-impact.ts:13` (unchanged line)

```ts
import { CALLEES_TRUNCATED_SENTINEL, CALLEE_ID_SEP } from '../../core/ingestion/cfg/emit.js';
```

ESM evaluates all of `emit.js` and its closure for two strings. Measured: **8 analyze-only `core/ingestion/cfg/*` modules** (`emit`, `reaching-defs`, `reaching-defs-graph`, `control-dependence`, `post-dominators`, `synthetic-escape`, `call-site-harvest`, `reaching-def-reason-codec`) load at `mcp/server.js`, `mcp/local/local-backend.js` **and** `mcp/http-transport.js`. This is a residual instance of the exact defect class the PR set out to remove, and the new guard cannot catch it — its forbidden regexes cover only `core/ingestion/languages/` and `core/group/extractors/|node_modules/tree-sitter`. Moving the sentinel and separator to a leaf constants module meets the PR's own "delete cost, don't relocate" bar.

### P2 — the new lazy `syncGroup` import is on the one changed control-flow line with no test reaching it

`gitnexus/src/core/group/service.ts:349`

Every test that references `groupSync`/`group_sync` either checks tool metadata or short-circuits before this line: `test/unit/group/service.test.ts` returns at the empty-name guard; `group-service-not-found.test.ts` mocks `loadGroupConfig` to reject (its `syncGroupMock` is declared and never invoked); `test/integration/group/group-sync.test.ts` imports `syncGroup` **directly**, bypassing `GroupService` entirely; the new startup guard only asserts the **negative** (that `sync.js` is absent at startup). `tsc --noEmit` catches a path typo, which caps the risk — but nothing verifies the import resolves and hands off correctly in the built ESM `dist/` when `group_sync` is actually called. Every production invocation of that tool hits this line. (Three lanes separately confirmed the surrounding design is sound: `syncGroup` is genuinely the only consumer, the import sits after config resolution so no variable is read before assignment, and a load failure now surfaces as a scoped MCP tool error instead of killing the whole server at startup.)

### P3 — ascent-reached blocks are in the published slice but never scanned, so both incompleteness flags stay false over a strict prefix

`gitnexus/src/mcp/local/pdg-impact.ts:1941` vs `:2047` (surfaces at `:1149`)

`:1941` merges the U-C4 ascent's blocks into `reachable` only; `hopReached` (declared at `:2007`) is the sole source of the next hop's `sliceBlocks`. So an ascent-reached block lands in the final `reachableBlocks` while its `calleeIds` cell is never handed to `calleeIdsByBlock` — reaching neither the `CALL_SUMMARY` scan, nor `calleeReferencesSeen`, nor `calleeListTruncated`. Reproduced: a block reached only by the ascent, carrying a cell of `[hidden, CALLEES_TRUNCATED_SENTINEL]`, yields `referencesScanned: 2` instead of 3, `examinedComplete: true`, `incompleteReasons: []` — the cell was capped at emit *and* named a third callee, and neither flag fired.

**Rated P3, not P1, for a specific reason:** `ascentBlocks` is non-empty only when `returnFlowing.size > 0`, so this shape always co-occurs with `returnFlowFound: true`. The load-bearing *negative* claim can never be published over this particular prefix. The damage is confined to a decorative `examinedComplete: true` and an under-counted `referencesScanned`. (This corrects the P1 rating Composer gave it and the framing in the coordinator's own verification notes.) Fix: add the ascent-reached blocks to `hopReached` — they *are* slice blocks — or scope `examinedComplete` in the docs to descent-visited blocks rather than "the slice's WHOLE callee list".

**Test gap:** the mock's `ascentOnly` block is already intra-reachable at the default `maxDepth: 3` and is never given a `calleeIds` cell, so the existing "ascent fired → `examinedComplete: true`" case at `impact-pdg-ascent-note.test.ts:660-672` **passes for the wrong reason** and enshrines the current behaviour.

### P3 — `pdgEvidence.ascent` is absent on the empty-slice return, contradicting the contract sentence this PR itself added

`gitnexus/src/mcp/tools.ts:452` (added here) vs the `reachableBlocks.length === 0` return at `gitnexus/src/mcp/local/pdg-impact.ts:2396`

`tools.ts:452` states *"present iff the inter-procedural descent ran"*, and the type doc repeats *"Absent ⇒ nothing was scanned."* But `runImpactPDG` returns the empty-slice result **before** `assemblePdgImpactResult`, and that object carries no `pdgEvidence` at all — after the descent has already run and scanned. Reproduced twice, independently:

```
--- direction=downstream
   reachableBlocks: []      CALL_SUMMARY ids scanned: 1
   has pdgEvidence: false   ascent: undefined
```

So `ascent: undefined` conflates "upstream — never scanned" with "downstream — scanned one reference and found nothing", which is the exact distinction the field was introduced to make. Trigger: `impact({mode:'pdg', line:N})` on a statement whose only dependents are interior to its own coalesced block and whose callees resolve to no CFG body (an out-of-repo target, an interface method, or the `Class:` id a `new` expression contributes — the PR's own `UNENTERABLE_CALLEES` fixture shape). Fix: emit `pdgEvidence: { ascent: … }` on that return too, or narrow the wording to "present iff the descent ran AND the slice is non-empty."

### P3 — `referencesScanned` is a distinct-callee tally rendered as a call-site count

`gitnexus/src/mcp/local/pdg-impact.ts:1082-1084` (the note) and `gitnexus/src/mcp/tools.ts:452` (the field doc); accumulator at `:1873`/`:1905`, published at `:2062`

`calleeReferencesSeen` is a `Set<string>` of callee ids, so `references` is a distinct-id tally — but both surfaces render it as call sites: `none of the ${references} call-site callee reference(s)`, and `tools.ts:452` describes it as "call-site callee references scanned". The internal doc is inconsistent with itself on the same line ("the distinct-id tally … **CALL SITES**"). Reproduced: two distinct slice blocks each calling `helper` report `referencesScanned: 1` and "none of the 1 call-site callee reference". Any slice invoking the same callee from more than one block — extremely common — leads a reader to under-estimate what was checked. Fix: rename to `calleesScanned` / "distinct callee(s)" in both surfaces, or count occurrences if the call-site reading is intended. The existing tests pin the current wording, so a fix must move them.

**Test gap:** `referencesScanned` is never tested with two call sites to the same callee — one existing block varies ids *within* a cell, another varies callees *across* hops; neither can distinguish the two readings.

### P3 — a slice block whose `calleeIds` cell is empty contributes nothing to the population while `examinedComplete` still reports `true`

`gitnexus/src/mcp/local/pdg-impact.ts:1640` and `gitnexus/src/core/ingestion/cfg/emit.ts:350`

`calleeIdsOfBlock` returns `''` for a whole file when the resolved-id map is absent (`if (fileMap === undefined) return '';`). `calleeIdsByBlock` then drops the row entirely (`if (calleeIds.length > 0) out.push(…)`) and, because no sentinel is written into an empty cell, raises no `calleeListTruncated`. Such a block can still carry real call sites in its name-keyed `callees` cell — the statement bridge relies on exactly that. So a mixed slice (criterion file has a fileMap, a hop-1 callee's file does not) crosses a hop, publishes `examinedComplete: true`, and quantifies universally over a population that silently excludes every call site in the id-less blocks. `[code-read]` — the consumer half is mechanical from those two lines; the producer half could not be constructed without a real index build. Fix: treat "block in the descent's slice with a non-empty `callees` cell but an empty `calleeIds` cell" as a third incompleteness reason, or at minimum document the population as "callees the index recorded resolved ids for".

### P3 — `referencesScanned > 0` with `callSummaryLayerPresent: false` is a consumer footgun, not a defect

`gitnexus/src/mcp/local/pdg-impact.ts:1143-1150`

On a pre-FU-C (v3) index, `calleesWithReturnFlow` returns empty sets while `calleeReferencesSeen` still accumulates every callee id, yielding `{referencesScanned: N>0, returnFlowFound: false, callSummaryLayerPresent: false}` — a state the PR's own test pins. Two lanes filed this as P2. The correctness lane **refuted it as a defect**: the state is self-consistent for that index shape, the note simultaneously emits its "re-index for `CALL_SUMMARY`" sentence, and the inverse (`returnFlowFound: true` with the layer absent) is not producible by any current analyze path. What remains is real but smaller: a consumer that branches on the structured pair and ignores `callSummaryLayerPresent` reads "scanned N, found no flow" when the truth is "the layer that records flow does not exist here." Consider an explicit `scanStatus: 'layer-absent'` so the pair cannot be misread.

### P3 — the pinned resolver gaps cite this PR, not the tracking issue

`gitnexus/test/integration/cfg/pdg-chained-receiver-callees.test.ts:252` and `gitnexus/test/integration/resolvers/typescript-inferred-field-receiver.test.ts:254`

Both describe blocks name the gap `(#2802 follow-up)`. The tracking issue **#2807** exists (open, labeled bug) and **PR #2810 already addresses it**. After merge the gap is discoverable only by reading a `KNOWN GAP` marker. Cite #2807 in both blocks. Related merge-order note: these pins assert the gap *exists*, so if #2810 lands first this PR needs a rebase and pin update to clear CI — worth stating the intended order (2806 then 2810).

### P3 — the guard's recorded baselines are stale by exactly the modules this PR's own second commit removed

`gitnexus/test/integration/mcp/startup-language-closure.test.ts:80-84` and `:36-37` — the comment states "server.js loads 489 distinct modules, local-backend.js 265, cli/mcp.js 4". Measured at HEAD: **387 / 163 / 4** — deltas of exactly 102 each, the group-extractor closure commit `9ea9676dc` removed. The floors (100 / 3 / 50) still hold comfortably and the subset relation the header claims still holds exactly, but the numbers are presented as the observation justifying those floors. By contrast `registry-import-closure.test.ts`'s "553 distinct modules" is accurate.

---

| # | Sev | Location | Theme |
|---|---|---|---|
| 1 | **P1** | `gitnexus/src/mcp/local/pdg-impact.ts:999` (+ `:1149`) | `examinedComplete: true` over a discarded per-callee depth clamp |
| 2 | P2 | `gitnexus/test/integration/mcp/startup-language-closure.test.ts:85` | group policy unanchored; `cli/mcp.js` row already vacuous |
| 3 | P2 | `gitnexus/test/integration/mcp/startup-language-closure.test.ts:92` | `mcp/http-transport.js` unprobed (523 modules) |
| 4 | P2 | `gitnexus/src/core/group/service.ts:349` | lazy `syncGroup` import reached by no test |
| 5 | P3 | `gitnexus/src/mcp/tools.ts:452` | "present iff the descent ran" is false on the empty-slice return |
| 6 | P3 | `gitnexus/src/mcp/local/pdg-impact.ts:1082` | `referencesScanned` is distinct callees, rendered as call sites |
| 7 | P3 | `gitnexus/src/mcp/local/pdg-impact.ts:1640` | empty `calleeIds` cell silently shrinks the population |
| 8 | P3 | `gitnexus/test/integration/mcp/startup-language-closure.test.ts:80` | stale 489/265 baselines (measured 387/163) |
| 9 | P3 | `gitnexus/test/integration/cfg/pdg-chained-receiver-callees.test.ts:252` | cite #2807, not "#2802 follow-up" |

Body-only (the cited line is outside the diff and cannot be anchored inline): the `cfg/emit.js` eager evaluation at `pdg-impact.ts:13`; the `SPAWN_CLI` registration gap in `scripts/cross-platform-tests.ts:162-189`; the ascent-reached-block prefix at `pdg-impact.ts:1941`/`:2047`.

---

## Attacked and held

Validation is a result. Every item below was a real attempt with a constructed trigger; several were filed as findings by one lane and then refuted by a lane with stronger evidence.

- **"A new false-alarm class fires for TypeScript/JavaScript criteria."** Filed as P2 by a Claude swarm lane and echoed by Composer's swarm leg — **refuted by execution.** `interproceduralHops > 0` (the enclosing gate) implies `references > 0`, because `hopsReached = hop + 1` is only reached after a non-empty `calleeIds` cell, so the new `references > 0` sub-gate is a defensive no-op and every case that got the old language-keyed note still gets a note. And a non-TS/JS callee gets a cleanly-decoded **empty** summary rather than no row (`call-summary-harvest.ts:51-55`), which makes "a property of the persisted summaries" literally true. The only real cost is note length in a now-common case — token budget, not correctness.
- **"The note's 2×2×2 grammar breaks, or 'Every summary examined decoded' can be false."** Refuted across all 16 renderings driven through real code. The only nit is the awkward "none of the 1 call-site callee reference carries".
- **"Upstream or statement-mode paths read `ascentCoverage` as a zeroed record."** Refuted: `ascentCoverage` is assigned only inside `if (direction === 'downstream')`, the note's ascent branch is nested under a gate only the descent can raise, and the field spread is truthiness-gated. Verified: upstream scans 0 ids and publishes `ascent: undefined`.
- **"Ids are counted but not scanned, or scanned but not counted."** Refuted: the scan call and the counter loop are adjacent with no branch between them and range over the identical set; every `continue`/`break`/budget cut in the hop body sits *after* the count.
- **"Deleted coverage in the trimmed CFG tests."** Refuted mechanically: `expect(` counts 11→11, 20→20, 69→69; `it(`/`test(` counts 5→5, 3→3, 7→7. All deletions are boilerplate replaced by `createTempDirPool`, which additionally fixes a latent double-`afterAll` over one array in `pipeline-pdg.test.ts`.
- **"`it.fails` pins go green on any throw."** The premise is false — neither new pin uses `it.fails`, and both reject it in comments as strictly weaker. Attempts to keep them green while the behaviour is wrong all failed; a marker rename makes them red.
- **"The probe silently degrades without `registerHooks`."** No version check exists, and that is the safe design: on a Node lacking the export the child fails at link time, exits non-zero, and the harness throws. There is no fallback to degrade into.
- **"A dead child reports zero modules, so every 'must not load X' passes trivially."** Refuted at four places: non-zero status throws, a missing payload marker throws, `child.on('error')` rejects, and `assertNonVacuous` throws — all before any policy `expect` runs.
- **"`minModules` floors are met on a severed build."** Measured vs declared: 387/100, 163/50, 553/100. A severed entry loads one module and fails both guards. Only `cli/mcp.js` is tight (4 vs 3), and its own comment says the anchor is the load-bearing guard there.
- **"A sixth entry point reintroduces the language registry lazily."** Refuted — no MCP path still imports `ingestion/languages`; the fix deleted the import rather than deferring it, and zero top-level `await import(` exists anywhere in `src/`.
- **"Moving the tree-sitter native load into a live JSON-RPC session corrupts the stdio frame stream."** Refuted for two independent reasons: `installGlobalStdoutSentinel()` is permanent with no release path and survives `silenceStdout`/`restoreStdout` cycles, and the only `withMcpWrite` call site wraps a single synchronous write, so a tool handler can never inherit the tag.
- **"`calleesWithReturnFlow` silently scans a prefix via a `LIMIT`."** No `LIMIT` in either new-path query.

## Lane disagreements worth recording

- **On the truncation flags**, `gitnexus-risk-architect` (Claude swarm) audited every truncation source and reported "**No gap found** — `examinedComplete: true` is never emitted over an incomplete set." That is wrong, and instructively so: it traced the flags that *are* folded and missed the two that are dropped. Composer's swarm leg examined the same question and also concluded it "could not find an under-report". Two Claude lanes then **reproduced** the defect end-to-end through separate probes. Composer's adversarial leg found a *different* prefix mechanism (ascent expansion, filed as P3 below) but not this one. **Four of the six passes that looked at this question missed the P1, including both Composer legs — so this P1 has no cross-engine backing, and lane consensus is not safety.**
- **On the TS/JS note**, the swarm risk lane and Composer's swarm leg both filed a P2 that the correctness lane refuted by execution (above). A reproduced refutation outranks two code-read raises.
- **On the ascent-expanded-block prefix**, Composer rated it P1 and the coordinator's own notes framed it as co-equal with the depth-clamp defect. The correctness lane's reproduction showed it can only ever co-occur with `returnFlowFound: true`, so it cannot carry the false negative — **P3, and the rating in this digest reflects that correction.**
- **On the `syncGroup` vs `csv-generator` double standard**, Composer's adversarial leg filed it P2 ("relocates cost, same shape as the rejected proposal"). Composer's own swarm lane and the Claude performance lane both **refuted** it independently with evidence: `bm25-index.ts:15` statically imports `normalizeFtsText` from `csv-generator.js` and `local-backend.ts` reaches `bm25-index` on the FTS **query** path, so deferring it genuinely relocates cost to a hot path; `sync.js` has exactly one consumer among eight group tools, and that consumer is a rare, explicitly user-initiated sync. Measured: the deferred closure is 280 modules / 8 native bindings / ~480 ms, paid once per process.
- Two Claude swarm lanes (`test-ci-verifier`, `branch-hygiene`) stopped mid-investigation and were resumed; both returned all-green with zero findings, and `branch-hygiene` contains a verifiably wrong evidence claim (it attributed the `dist/mcp/resources.js` anchor to the registry guard; that anchor belongs to the `mcp/server.js` entry). Their "verified" statements are `[code-read]` at best.

## Lower-priority items and residual risks

- The `:1139-1142` rationale ("a zeroed object on a slice that scanned nothing would read as 'we looked and found nothing'") does not describe what the code does — every downstream slice with no `calleeIds` at all publishes exactly `{referencesScanned: 0, returnFlowFound: false, examinedComplete: true}`. The behaviour is defensible; the comment is not.
- `'traversal-truncated'` also over-fires from the ascent BFS's row limit at `:1940`, whose blocks are never scanned anyway. Sound over-hedging, unobservable in prose because the note is silent whenever an ascent fired — worth a comment, not a fix.
- The producer-side `CALL_SUMMARY` emit cap (`pdgMaxCallSummaryEdges`) is invisible to the coverage. Default is unlimited, so it only bites on an explicitly configured run — but when it does, `returnFlowFound: false` / `examinedComplete: true` is published over summaries that were never persisted, with nothing to say so.
- `test/helpers/module-load-probe.ts:196` — `extraEnv` is spread **last**, so a future `env: { PROBE_TARGET: … }` silently redirects the probe while `anchor`/`minModules` stay keyed on `entry`. One-line fix: move `PROBE_TARGET: targetUrl` after the spread.
- `test/helpers/module-load-probe.ts:249` — `JSON.parse(...) as string[]` crosses a process-spawn boundary with no `Array.isArray`/element check, at a boundary `DoD.md:36` names explicitly. Relatedly, `toRepoRelativePosix` returns out-of-repo entries verbatim, so a `node_modules` module reached through a symlinked tree is counted once per channel — inflating `modules.length` and weakening the floor. Cosmetic today.
- `test/helpers/module-load-probe.ts:190-198` — `spawn`'s `timeout` sends one `SIGTERM` with no `SIGKILL` escalation; a child stalled in synchronous native code is orphaned rather than reaped.
- `test/helpers/temp-dir-pool.ts:50-52` — the cleanup loop has no per-iteration `try/catch`, and `force: true` suppresses only `ENOENT`; one `EBUSY`/`EPERM` on Windows leaks every directory after it. Pre-existing in all four copies this PR consolidated — a missed opportunity, not a fresh defect, but the blast radius is now shared.
- Maintainability: the "undecodable ≠ empty" rationale is restated in full prose at **seven** sites, in the commit immediately after this PR's own "consolidate the rationale to one canonical site" commit; the empty-ascent note is built from five nested ternaries in one 20-line `push()`; and `pdg-impact.ts` grows 2633 → 2924 lines with a cohesive, extractable unit that was not extracted.
- `assertNonVacuous`'s own two throw paths have no test — the helper's central claim rests on code no test exercises (it is not exported). The probe's ESM channel is also per-thread: a module that loads the language registry inside a worker thread at evaluation time is invisible to the guard. Nothing does this today; the constraint is undocumented.
- Process note: `gh pr view 2806 --json reviews` returns `[]` — no independent review is on record. The PR body's "reviewed and thirteen findings were fixed" refers to an AI-session-internal pass by the same entity that wrote the code.

## Test gaps

1. No row truncates inside a callee's own BFS — the P1 path. `INCOMPLETENESS_PREMISES` covers only the top-level `maxDepth: 1` case, whose flag *is* folded.
2. The existing "ascent fired → `examinedComplete: true`" case passes for the wrong reason (its `ascentOnly` block is already intra-reachable and has no `calleeIds` cell) and enshrines the current behaviour.
3. No case asserts `pdgEvidence.ascent` on the `reachableBlocks.length === 0` return.
4. `referencesScanned` is never tested with two call sites to the same callee.
5. `callSummaryLayerPresent` is only tested through the injected flag, never through `pdgLayerStatus` / `meta.pdg.hasCallSummary`, so a change to the `'ready'`-only gate would publish `callSummaryLayerPresent: false` on a good index with nothing to catch it.
6. No entry anchors the `→ core/group/service.js` edge, so nothing enforces the premise of the group assertion; `mcp/http-transport.js` has no import guard at all, and `mcp/resources.js` / `mcp/staleness.js` are named in the headline claim but not probed.
7. Nothing executes `GroupService.groupSync`'s `await import('./sync.js')`, in either the success or the failure direction.
8. `test/unit/tools.test.ts` pins sibling parts of the `impact` description (`truncatedBy`, `affectedStatements`) but nothing pins the new ascent sentence — the sole external-facing description of the new field.

## CI

Every check passes — typecheck, lint, format, CodeQL (JS/TS + Python), Gitleaks, Dependency Review, tree-sitter ABI ×3 platforms, packaged install smoke ×2, Node floor compat, ubuntu/windows/macOS shards, benchmarks, eval containment, CI Gate, coverage merge; 16530 tests passed / 0 failures / 72 skipped. The **only** failure is **Vercel — "Authorization required to deploy"**, a team authorization gate, not a code defect. `MERGEABLE` / `UNSTABLE`, no conflicts with `main`.

---

*This is an automated multi-tool review digest (Claude swarm + Claude `ce-code-review` personas + Composer standing in for an unavailable Codex leg, with coordinator-executed verification). Verify each finding against the code before acting on it.*

PR #2809, opened 2026-08-03
docs(taint): document per-language model files

### LINKED ISSUE #2805: docs: taint skill points Python/Java sinks at typescript-model.ts; doc never mentions either language
### Area

Documentation / developer experience

### Summary

The `gitnexus-taint-analysis` skill doc never mentions Python or Java, and its only
instruction for adding a source/sink/sanitizer points at `taint/typescript-model.ts`.
Taint models actually exist for four language ids across three files, so the doc both
(a) reads as "taint analysis is TypeScript-only" and (b) sends a contributor to the
wrong file for a Python or Java sink.

`grep -ciE 'python|java'` returns **0** for the doc, while `registerBuiltinTaintModels()`
registers four languages.

### Expected behavior

The "Adding a source / sink / sanitizer" section should name the per-language model file
for the language being targeted, and the doc should state its language coverage somewhere.

### Actual behavior

`gitnexus/skills/gitnexus-taint-analysis.md` L149-153:

```
## Adding a source / sink / sanitizer

Edit the language model in `taint/typescript-model.ts` (registered via the
explicit `registerBuiltinTaintModels` seam, keyed by `SupportedLanguages`). The
spec is hashable data (no functions). ...
```

But `gitnexus/src/core/ingestion/taint/typescript-model.ts` L153-157 registers four ids
from three different model files:

```ts
export function registerBuiltinTaintModels(): void {
  registerSourceSinkConfig(SupportedLanguages.Java, JAVA_TAINT_MODEL);
  registerSourceSinkConfig(SupportedLanguages.TypeScript, TS_JS_TAINT_MODEL);
  registerSourceSinkConfig(SupportedLanguages.JavaScript, TS_JS_TAINT_MODEL);
  registerSourceSinkConfig(SupportedLanguages.Python, PYTHON_TAINT_MODEL);
}
```

with `JAVA_TAINT_MODEL` from `./java-model.js` and `PYTHON_TAINT_MODEL` from
`./python-model.js` (L21-22). So `typescript-model.ts` holds only the TS/JS spec plus the
aggregating registration function — a Python sink belongs in `python-model.ts`.

Concrete consequence: following the doc to add a Python sink means editing
`typescript-model.ts`, where the Python spec is not defined.

Secondary: `gitnexus-pdg-query.md` also has 0 python/java mentions, so a reader of either
skill comes away thinking the `--pdg` layers are TS-only.

### Steps to reproduce

1. `grep -ciE 'python|java' gitnexus/skills/gitnexus-taint-analysis.md` → `0`
2. `sed -n '149,157p' gitnexus/skills/gitnexus-taint-analysis.md` → the section above,
   naming only `taint/typescript-model.ts`
3. `sed -n '150,158p' gitnexus/src/core/ingestion/taint/typescript-model.ts` → four
   `registerSourceSinkConfig` calls covering Java, TypeScript, JavaScript, Python
4. `ls gitnexus/src/core/ingestion/taint/*-model.ts` → `java-model.ts`,
   `python-model.ts`, `typescript-model.ts`

### Environment

Verified against `main` (default branch) and npm `gitnexus@1.6.9` — the two copies of the
doc are byte-identical, so this is current, not a packaging lag.

### Additional context

**The doc is triplicated, which is likely why it drifted from the code.** All three copies
are identical and all three have 0 python/java mentions, so a one-file fix leaves two
stale copies behind:

- `gitnexus/skills/gitnexus-taint-analysis.md`
- `.claude/skills/gitnexus/gitnexus-taint-analysis/SKILL.md`
- `gitnexus-claude-plugin/skills/gitnexus-taint-analysis/SKILL.md`

Deduplicating these (generate the latter two from the first, or check a drift guard in CI)
would probably be worth more than this specific fix.

Suggested replacement for the section, if useful:

> **Language coverage.** Taint models exist for four `SupportedLanguages` ids in three
> files: `TypeScript`/`JavaScript` → `TS_JS_TAINT_MODEL` (`taint/typescript-model.ts`),
> `Python` → `PYTHON_TAINT_MODEL` (`taint/python-model.ts`), `Java` → `JAVA_TAINT_MODEL`
> (`taint/java-model.ts`). All four are wired by `registerBuiltinTaintModels()`, which
> lives in `typescript-model.ts` and is an explicit call seam, not an import side-effect.
>
> **Adding a source / sink / sanitizer.** Edit the model file for the language you are
> targeting — a Python sink goes in `python-model.ts`, not `typescript-model.ts`.

Found while auditing which GitNexus skills were worth keeping locally; the TS-only reading
was what prompted the check.

koriyoshi2041:
Closes #2805.

## What changed

- documents the four supported taint language IDs and their three model files
- directs contributors to the model for the language they are changing
- clarifies that `registerBuiltinTaintModels()` registers all four models from `typescript-model.ts`
- keeps the source, Claude skill, and plugin skill copies synchronized

## Verification

- `npm test -- --run test/unit/shipped-skills-sync.test.ts` — 53 passed
- pre-commit typecheck — passed
- `git diff --check` — passed

Docs-only change; no runtime behavior changes.


azizur100389:
Reviewed current head `4d9e081a` against current main `7468cc91`; it merges cleanly. The new language-to-model mapping matches the source registration, all three shipped copies are byte-identical, typecheck passes, and the skill-sync suite passes 50 tests with 3 platform skips.

One nonblocking documentation follow-up: immediately after the new language-aware mapping, the validation guidance still names only `test/helpers/ts-cfg-harness.ts`. Consider also pointing Python and Java contributors to `python-model-match.test.ts` and `java-model-match.test.ts`.

koriyoshi2041:
Added the Python and Java model-match tests alongside the TypeScript/JavaScript harness in `e2edc3be`. The three shipped copies remain synchronized; the skill-sync suite passes 53/53 and pre-commit typecheck passes. Thanks for catching the language-specific validation gap.

PR #2810, opened 2026-08-03
fix: type an inference-typed class field so it can act as a call receiver (#2807)

### LINKED ISSUE #2807: Inference-typed field receivers resolve to no CALLS edges at all
## Summary

When a class field's type must be inferred from its initializer, calls through that field resolve to **nothing**. Not just the chained link, the whole call site: even `this.p.inner()`, an ordinary named receiver call, produces no `CALLS` edge and leaves `BasicBlock.calleeIds` empty.

The discriminator is the **type annotation**, not local vs field. An annotated field resolves fine.

## Repro

```ts
// src/app.ts
export class Inner {
  compute(v: number): number { return v * 2; }
}
export class Outer {
  inner(): Inner { return new Inner(); }
}

export class Annotated {
  private p: Outer = new Outer();          // annotated
  run(x: number): number { return this.p.inner().compute(x); }
}

export class Inferred {
  private p = new Outer();                 // inferred from initializer
  run(x: number): number { return this.p.inner().compute(x); }
}
```

`gitnexus analyze .`, then every `CALLS` edge in the index:

```
Class:src/app.ts:Annotated      -> Class:src/app.ts:Outer
Class:src/app.ts:Inferred       -> Class:src/app.ts:Outer
Method:src/app.ts:Annotated.run#1 -> Method:src/app.ts:Inner.compute#1
Method:src/app.ts:Annotated.run#1 -> Method:src/app.ts:Outer.inner#0
Method:src/app.ts:Outer.inner#0 -> Class:src/app.ts:Inner
```

`Annotated.run` gets both links. `Inferred.run` gets **zero edges**, from an identical statement.

## The initializer is not the problem

Note both classes emit `Class:X -> Class:Outer`. The `new Outer()` in the inferred field **is** resolved, byte for byte the same as the annotated twin. So the constructor call is seen; only the step that binds that initializer's type back to the field is missing. That should narrow where the fix belongs.

## Shape coverage

Nine receiver forms, same `.inner().compute(x)` statement in each, measured against the real pipeline:

| receiver | resolves |
|---|---|
| `const o = new Outer()` | yes |
| local inside a method | yes |
| `private p: Outer = new Outer()` | yes |
| `private p: Outer;` + ctor `this.p = new Outer()` | yes |
| `private p: Outer;` + ctor param `this.p = p` | yes |
| `constructor(private p: Outer) {}` (parameter property) | yes |
| `makeOuter().inner().compute()` (call result receiver) | yes |
| three link chain `o.inner().mid().compute()` | yes |
| **`private p = new Outer()`** | **no edges** |
| **`private p;` + ctor `this.p = new Outer()`** | **no edges** |

Both failing forms are exactly the ones where the field carries no type annotation.

## Impact

`interproceduralDescent` hops through `BasicBlock.calleeIds`, so for these shapes the PDG cannot cross the call boundary at all. `impact(mode:'pdg')` silently under reports for a common OO pattern, and `impact` in callgraph mode misses the caller relationship too.

Inference typed fields are idiomatic TypeScript, so this is not an edge case shape.

## Already pinned

`gitnexus/test/integration/resolvers/typescript-inferred-field-receiver.test.ts` covers all nine forms as `CALLS` edge assertions and pins the two failing ones at their current value. It is self diffing: when this is fixed, that test fails and prints the newly resolved ids, which is the signal to update it.

There is a second pin at the `calleeIds` layer in `gitnexus/test/integration/cfg/pdg-chained-receiver-callees.test.ts`.

Found while working on #2802, but independent of it. #2802 does not touch receiver resolution.

magyargergo:
Fixes #2807.

## The bug

A class field whose type must be inferred from its initializer produced **no `CALLS` edge at all** — not a truncated chain, nothing:

```ts
class Inferred {
  private p = new Outer();                       // no type annotation
  run(x: number) { return this.p.inner().compute(x); }   // emits ZERO edges
}
```

`this.p.inner()` is an ordinary named-receiver call and it was lost too. An annotated twin (`private p: Outer = new Outer()`) resolved fine.

## Root cause: a missing capture pattern, not resolver logic

`typeOfMemberOnClass` reads `classScope.typeBindings`, and `foldReceiverChain` declines at the *first* untypeable step rather than folding against a guessed owner — so one missing binding kills the whole chain, first link included. Nothing in `compound-receiver.ts` was wrong.

TypeScript's `query.ts` had `@type-binding.constructor` for `variable_declarator` and `@type-binding.annotation` for `public_field_definition` — and **no pattern at all** for `public_field_definition value: (new_expression)` or for `this.<field> = new …`. The initializer itself was never invisible: `new Outer()` always emitted its own constructor edge. Only the step binding that initializer's type back to the field was absent.

Kotlin and Swift needed no such pattern because one grammar node (`property_declaration`) covers both a local and a stored property; TypeScript splits them, and only the local half was ever covered.

## Scope: measured across every language where the shape exists

The metric is **parity with each language's own control row**, not "both chain links present". JavaScript, Python, Dart and PHP lose the second link even for a plain local, because nothing annotates the intermediate return type — a separate return-type-inference gap. Scoring on "both links" would have falsely accused four languages.

| language | inferred field (initializer) | assigned field (`this`/`self`) |
| --- | --- | --- |
| TypeScript | **fixed** | **fixed** |
| JavaScript | **fixed** | already correct |
| Python | n/a — no field declarations | **fixed** |
| Ruby | n/a — no field declarations | **fixed** |
| Kotlin | already correct (pinned) | n/a — needs a type |
| PHP | n/a — initializers take only constant expressions | already correct (pinned) |
| Dart | **fixed** | known gap, pinned |
| Swift | **fixed** | n/a — needs a type |

Kotlin and PHP were already correct and are pinned anyway: the languages that got receiver typing for free are exactly the ones nobody re-checks.

## Swift needed a separate defect fixed first

Swift's row was **masked by a caller-attribution bug**, not by receiver typing. In a file where two classes each define `func run`, every call in **both** bodies was attributed to whichever registered first — it collected duplicate edges while its twin collected none, *including when the twin was the annotated control*. Renaming one method fixed it; moving it to another file fixed it; so the collision was name-keyed and per-file.

The cause is a **label split**: Swift's structure phase emits a type's methods as `Function` nodes, while the scope extractor derives `Method` from the `@declaration.method` anchor. Every key in `resolveDefGraphId` is label-scoped, so such a pair misses all of them and falls to `simpleKey(filePath, name)` — deliberately label-agnostic and first-write-wins. Fixed at both ends: Swift now qualifies a method def as `<Type>.<method>` (matching the qualifier the node id already encodes), and the bridge retries the qualified keys under the sibling callable label, **gated on the name containing a dot** — `A.run` names one construct whatever the label, while a bare `run` is exactly the top-level-vs-method aliasing the label was added to prevent.

With that fixed, Swift's `let p = Outer()` binding turned out to have been correct all along.

**Dart's** gap was one layer below the binding: both class-field query patterns require a written type, so `var b = Outer();` produced no `@declaration.property` capture at all. Dart puts the keyword there instead and spells it two ways — `inferred_type` (`var`) and `final_builtin` (`final`, `late final`) — so both are matched; covering only `var` would leave the more idiomatic style broken.

One row stays pinned: Dart's `var r; C() { r = Outer(); }`. Dart writes fields with no receiver prefix, so binding it means treating assignment to a bare identifier as a field write, indistinguishable from a constructor-local. It keeps its `callerExists` probe, so closing it fails loudly.

## Testing

- Both pre-existing self-diffing pins flip from `KNOWN GAP` to resolving; neither was weakened to `it.fails`. The TypeScript pin gains a method-assigned row and a deliberately mistyped `private p: Mismatch = new Outer()` row that asserts the annotation-over-initializer source-strength tie-break **executably** — that row also pins a pre-existing artifact (`Inner.compute` still resolves through the hoisted module-level return-type binding), verified byte-identical on the pre-fix tree.
- New `inferred-field-receiver-matrix.test.ts`: eight languages, one fixture each, control + inferred + assigned rows, plus a whole-matrix guard that every language keeps a *resolving* control — that is what makes a gap row mean "broken" rather than "fixture never worked".
- Full resolver + CFG sweep with every fix: **3165 passed, 1 skipped, 0 failed**, against a 3164-passing baseline of the same suite before the Swift/Dart work — so the shared-bridge change regresses nothing. `tsc --noEmit` clean in both packages.

Self-review caught one defect in this PR's own Python commit (`93f75e7cc`): `constructorCallTypeName` accepted any call, so `self.p = self.build()` bound `p` to the non-type `"self.build"` and — sharing the weakest tier — a later such assignment *displaced* an earlier real `self.p = Outer()`. Now rejects callees rooted at the receiver, pinned by a row that fails without the rejection.

## Commits

| commit | change |
| --- | --- |
| `d7a659b0a` | shared temp-repo lifecycle test helper |
| `686950440` | TypeScript — field initializer + `this.p = new Outer()` |
| `18a48c731` | JavaScript — field initializer |
| `6c99f5488` | Python — `self.q = Outer()` as a third, weakest tier |
| `55d370793` | Ruby — `@q = Outer.new` + Class-scope hoist |
| `b6aa45d93` | cross-language matrix pin |
| `1aba80d94` | Python — a method call on the receiver is not a construction |
| `a8627541c` | Swift — resolve a method def to its node across the Method/Function label split |
| `b5f30383c` | Dart — declare inference-typed class fields |
| `2d3469085` | matrix: record Swift and Dart reaching parity |

Every edit lives under `languages/<lang>/`; no shared code in `core/ingestion/` names a language (AGENTS.md §42). No schema or version constant applies — this changes what the resolver produces, not how it is stored, so an existing index needs a re-analyze to show the new edges.


gitnexus/src/core/ingestion/utils/call-analysis.ts
```diff
@@ -555,12 +555,49 @@ export function extractCallChain(
...
+  // `undefined` is "not in the table"; `null` is "in the table, ungated". The
+  // two are distinguishable, so one `get` answers both questions — no separate
+  // `has` probe, and one crossing of the native `node.type` getter.
+  if (operator === undefined) return false;
+  if (operator === null) return true;
+  // The operator is an anonymous token, so it is not in `namedChildren`; the
+  // node's own text is the reliable place to read it. Deliberately NOT
+  // `node.lastChild` — measured 3x SLOWER (the child wrapper allocation
+  // dominates), and the token surfaces with type `bang`, not `!`.
+  return node.text.trimEnd().endsWith(operator);
```
magyargergo, line 598: **Swift custom postfix operators ending in `!` are peeled as force-unwraps.**

This gate is a text test on the whole node's source, and Swift's operator character set includes `!`. A user-defined `postfix operator +!` makes `(left +!).hit()` a `postfix_expression` whose text ends in `!`, so the peel fires and the receiver is typed as `left` rather than as whatever `+!` returns — a confidently wrong owner, the failure mode `compound-receiver.ts` exists to prevent. That is precisely the case the docblock at `:563-569` says this gate is here to stop ("user-defined postfix operators. Those can return anything").

Suggested fix: inspect the operator child node instead of the enclosing node's text — accept only a `bang` token (Swift `!`) or Kotlin `!!`, and reject multi-character custom operators. Add a Swift custom-postfix regression fixture; `grep -rln "postfix operator" test/ bench/` currently returns nothing.

Corroboration: reached independently by the Codex lane (finding #2, P2, confidence 75) and the Claude correctness lane (residual risk #2) — the only open item both engines found. Neither wrote a fixture, so the trigger is derived from the Swift grammar rather than measured.

gitnexus/src/core/ingestion/utils/call-analysis.ts
```diff
@@ -555,12 +555,49 @@ export function extractCallChain(
...
+const TRANSPARENT_RECEIVER_WRAPPERS = new Map<string, string | null>([
+  ['non_null_expression', null], // TypeScript `svc!`
+  ['parenthesized_expression', null], // `(svc)`
+  // NOT Swift-only: Kotlin's `!!` non-null assertion parses as the same node type
+  // and is equally type-preserving, so it is peeled too. Measured — the
+  // receiver-resolution bench moved `kotlin.nonNullAssert` VISIBLE-GAP ->
+  // RESOLVES when this landed, which is how the Kotlin effect was discovered
+  // rather than assumed. Any other grammar emitting `postfix_expression` is
+  // affected as well; the `!` gate, not the language, is what bounds this.
+  ['postfix_expression', '!'], // Swift `self.a!`, Kotlin `a!!`
```
magyargergo, line 580: *(Anchored here for visibility on the `postfix_expression` table entry — this line is **not** where the fix goes. See the last paragraph.)*

**Kotlin `a!!.m().n()` emits only the tail edge — the head call is silently lost.**

Measured: `var p: Outer? = null; p!!.inner().compute(x)` produces `Inner.compute#1` **only**, while the non-null control `var p: Outer = Outer(); p.inner().compute(x)` produces both `Outer.inner#0` and `Inner.compute#1`. The peel *does* type the receiver — that is the only way `.compute()` reaches `Inner.compute#1` — yet `inner()` itself emits no `CALLS` edge. The graph ends up internally inconsistent: `impact({target: "Outer.inner", direction: "upstream"})` will not list `Host.run`, while `Inner.compute`, reachable only *through* `Outer.inner`, will.

Neither gate can see this. `bench/receiver-resolution/measure.mjs:1031` defines the Kotlin row as `{ id: 'nonNullAssert', member: 'save', body: 'svc!!.getUser().save()' }` — `member` is the **tail**, so the rebaselined `"nonNullAssert": "RESOLVES"` is true and still hides half the chain. And a `!!` grep over `inferred-field-receiver-matrix.test.ts` returns nothing: there is no `!!` row.

Note the asymmetry: Swift's `self.p!.inner().compute(x)` is the mirror image — it keeps `Outer.inner#0` and loses `Inner.compute#1`. One shared peel, two grammars, opposite halves surviving.

Suggested fix — and this is why the comment is anchored on the table entry rather than on the fix site: the bug is likely **not** in `unwrapTransparentReceiver` (`:612-621`, where 612 is its JSDoc). Investigate where the head link is dropped for a peeled Kotlin receiver — the chain-link emitter, not the wrapper table. At minimum add a Kotlin `!!` matrix row asserting **both** links.

gitnexus/src/core/ingestion/languages/python/receiver-binding.ts
```diff
@@ -114,17 +114,51 @@ export function synthesizeReceiverTypeBinding(fnNode: SyntaxNode): CaptureMatch
...
+  if (callee.type !== 'identifier' && callee.type !== 'attribute') return undefined;
+  const text = callee.text.trim();
+  if (text.length === 0) return undefined;
+  // `self.p = self.build()` is a METHOD call, not a construction. Accepting it
+  // bound `p` to the non-type `"self.build"`, which resolves to nothing — and
+  // because it shares this weakest tier, a later such assignment DISPLACED an
+  // earlier real `self.p = Outer()`, leaving the field untyped again. Measured:
+  // both `self.q = self.make()` and the displacement pair emitted no CALLS edge
+  // at all. Rejecting a callee rooted at the receiver keeps the construction.
+  if (text === receiverName || text.startsWith(`${receiverName}.`)) return undefined;
```
magyargergo: **A dotted *method* callee is accepted as a type name.**

This guard rejects only a callee rooted at the receiver, so any other dotted callee is returned whole and handed to `resolveTypeRef`, which resolves dotted names through `QualifiedNameIndex` — and that index matches the trailing segment against a class of that name even when the callee is a method on an unrelated object:

```python
class Alpha:
    def ping(self): return 1
class Factory:
    def Alpha(self): return "not an Alpha"    # a METHOD
class Host:
    def __init__(self, f): self.svc = f.Alpha()   # svc is a str
    def run(self): return self.svc.ping()
```

Expected no edge; measured a fabricated `Host.run → Alpha.ping`.

P3 because it needs a method name colliding with a class name — the common `self.svc = factory.build()` correctly emits nothing, and `self.u = models.User()` is why the dotted arm exists.

Suggested fix: restrict the dotted arm to callees whose root segment is a known module/import alias (the shape the docblock directly above actually cites), or let the resolver reject a resolution whose target symbol kind is not a type. The bare-`identifier` arm is unaffected.

gitnexus/src/core/ingestion/languages/python/receiver-binding.ts
```diff
@@ -178,29 +219,57 @@ export function synthesizeConstructorFieldTypeBindings(fnNode: SyntaxNode): Capt
...
             const explicit = explicitType !== null;
+            const inferredFromConstructor = constructedType !== undefined;
+            const tier = explicit
+              ? TIER_EXPLICIT
+              : inferredFromConstructor
+                ? TIER_CONSTRUCTOR
+                : TIER_PARAMETER;
             const existing = candidates.get(field.text);
-            if (existing === undefined || explicit || !existing.explicit) {
+            if (existing === undefined || tier >= existing.tier) {
```
magyargergo, line 274: **`>=` lets a weak same-tier candidate erase a real constructor binding.**

`self.conn = Outer()` followed by `self.conn = Registry.get()` in one `__init__` types the field as `"Registry.get"`, which resolves to nothing, so `Host.run` emits no edge; removing the second line restores `Host.run → Outer.inner#0`. This is the exact hazard the docblock at `:139-144` names ("a later such assignment DISPLACED an earlier real `self.p = Outer()`") and then closes only for the receiver-rooted spelling.

Contained — the walk skips `if`/`for`/`try`/`match`, so only straight-line double assignment reaches it, and the failure direction is a lost edge rather than a wrong one.

Suggested fix: give a dotted callee its own tier below `TIER_CONSTRUCTOR`, or require a strictly greater tier unless the evidence kind matches. Fixing the dotted-callee guard above resolves this as a side effect.

gitnexus/src/storage/parse-cache.ts
```diff
@@ -174,7 +174,27 @@ import type { ParseWorkerResult } from '../core/ingestion/workers/parse-worker.j
...
+// when this was allocated, and this file records eight prior collisions.
+// v41: the v40 Dart field-write binding gained its READ-side mask (#2807 review)
+// — a Dart class-member body that rebinds one of its class's field names now
+// emits `@receiver-owner.shadowed-fields` on its synthesized `@scope.function`
+// match, which becomes `Scope.ownsReceivers`. Parse-time capture emission again,
+// so a v40 warm cache replays scope matches with no marker and the receiver walk
+// still reaches the class field — i.e. it keeps serving the WRONG edge this
+// bump's fix removes, silently. Same bump-or-nothing situation as v40.
+// RE-CHECK AGAINST origin/main IMMEDIATELY BEFORE MERGING.
+const SCHEMA_BUMP = 41;
```
magyargergo: **Re-check this against `origin/main` immediately before merge.**

Verified at synthesis time, *after* the most recent merge from main: this branch is `41`, `origin/main` is `39`, and the pin at `test/unit/incremental-parse-cache.test.ts:122` agrees — **no collision right now**. The risk is time-dependent, not code-dependent: this file records eight prior collisions and two exact clashes, and both clashes landed in the window *between* review and merge. The comment one line above says the same thing in capitals.

It matters more than usual here because every language change in this PR is parse-time capture emission, and `analyze` skips tree-sitter for byte-unchanged chunks — so a stale warm cache replays the old capture set silently, with no error. For the read-side mask specifically, a v40 cache replays scope matches with no `@receiver-owner.shadowed-fields` marker and keeps serving the exact wrong edge this bump's fix removes.

At merge: run `git show origin/main:gitnexus/src/storage/parse-cache.ts | grep 'const SCHEMA_BUMP'` and confirm it is not `41`. If main has taken `41`, bump this branch to `42` **and** update the pin test — the constant and the pin must move together.

magyargergo (commented):
# PR #2810 — tri-review synthesis

**Repo:** `abhigyanpatwari/GitNexus` · **PR:** #2810 — *"fix: type an inference-typed class field so it can act as a call receiver (#2807)"*
**Head:** `3f4975f99` · **Base:** `main` · **Merge base:** `b2cd1c2ad637657125248c0dd2046de71ceea965`
**Reviewed at:** `ab2f48c174476141a881caba3ba651972e0bf90b`; one commit landed after (`3f4975f99`) — see §9.6.
**Diff:** 33 files, +3844 / −257 · **Cross-repository (fork):** yes
**Method:** read-only. Every FIXED / STILL-OPEN claim below was re-verified against worktree source at `ab2f48c17`, not taken from a commit message.

> Current visible state is incomplete. I could verify the head SHA, the merge base, the full diff, the source of every finding, the `SCHEMA_BUMP` triple (branch / `origin/main` / pin), the test-matrix row delta, and the hidden-Unicode and whitespace hygiene checks; but not the live check-run results, the required-reviewer state behind `mergeStateStatus: BLOCKED`, existing PR review threads, or the runtime numbers the fix commit reports (test-sweep totals and the +4% capture-emission measurement). The findings below treat those missing items as mandatory verification points rather than confirmed facts.

---

## Provenance and limits — read this before weighting anything

This run did **not** achieve a two-engine matrix.

| artifact | engine | lane |
|---|---|---|
| `claude-swarm-risk-architect.md` | Claude | gitnexus-risk-architect |
| `claude-swarm-test-ci.md` | Claude | gitnexus-test-ci-verifier |
| `claude-ce-correctness.md` | Claude | ce-correctness-reviewer |
| `claude-ce-adversarial.md` | Claude | ce-adversarial-reviewer |
| `codex-ce.md` | **Codex, gpt-5.6-sol, effort xhigh** | ce-code-review (report-only) |

There is **no codex-swarm leg and no codex-adversarial leg**. Cross-engine corroboration in this digest rests on **one Codex leg**. Four of the five artifacts share one engine, one model family, and therefore one set of blind spots — four Claude lanes agreeing is *correlated* evidence, not four confirmations. Do not describe this run as "five independent confirmations" or as a full Engine B.

**What the single Codex leg actually contained.** `codex-ce.md` is ~1.1 MB and is overwhelmingly raw event stream — loaded skill files, guardrails, the diff, and sub-agent prompt templates. Every `"severity": "P0"` string in it (3 hits) belongs to the *template's own worked example* ("User-supplied ID in account lookup without ownership check"), not to this PR. The real Codex output is one block at `codex-ce.md:17467`, repeated at `:17509`: *"Three findings survived; all other candidates were REFUTED."* Those three, verbatim:

| # | finding | file:line | Codex severity | confidence |
|---|---|---|---|---|
| 1 | Static initializers overwrite instance-field bindings | `javascript/query.ts:445` | **P1** | 75 |
| 2 | Swift custom postfix operators treated as transparent | `call-analysis.ts:598` | P2 | 75 |
| 3 | Ruby eval blocks bypass the instance-ivar ownership guard | `ruby/captures.ts:138` | P2 | 75 |

**The honest consequence.** The one finding **both engines independently rated P1 is the static-field defect** (Codex #1 at P1/75, plus both Claude lanes). The Ruby block-receiver defect was also reached by both engines, but they **disagreed on severity** — Claude lanes called it P1, Codex called it P2. Codex **did not find the Dart 3 pattern-binder defect at all** — the highest-blast-radius finding of the round, reproduced independently by both Claude lanes with measured wrong-target output. So the round's most severe result rests on same-engine agreement only. Weighting rule applied throughout: single-lane findings I could verify structurally against source are labelled as such, with the lane named.

`prior-context.md` is a claude-mem recall — **context, not a finding**. Much of it was written by this PR's own earlier sessions, so #2807 / #2810 hits in it are self-echo. It is used below only to mark recurring regressions, and every recalled fact was re-verified.

---

## 1. Review bar for this PR

Derived from `DoD.md` v2.0.0 (§2.1 correctness, §2.2 architecture/placement, §2.4 contracts & persisted-data compatibility, §2.6 testing) plus `GUARDRAILS.md` and `AGENTS.md`. For a change of this shape — parse-time capture emission across six languages feeding receiver resolution — the bar is:

1. **No wrong-owner edge.** A missed edge is acceptable; a confidently *wrong* `CALLS` edge is not. This is the doctrine `scope-resolution/passes/compound-receiver.ts:490,526` states in the code itself ("a confidently wrong owner, which is strictly worse than…"). Any finding that converts "no edge" into "wrong edge" is merge-blocking.
2. **No language names in shared ingestion code.** `gitnexus/src/core/ingestion/` must route language behavior through `LanguageProvider` / `ScopeResolver` hooks (`AGENTS.md`, `DoD.md` §2.2).
3. **Persisted-cache compatibility.** Parse-time capture changes must move `SCHEMA_BUMP` *and* its exact-value pin together, because `analyze` skips tree-sitter for byte-unchanged chunks — a stale warm cache replays the old capture set silently, with no error.
4. **The four capture/resolution gates green**, not two: `bench/receiver-resolution --check`, `bench/scope-capture --check`, `swift-captures-golden`, `bench/callable-value-flow --check`.
5. **Behavior change pinned by targeted tests**, with each fix mutation-verified so reverting it turns its own rows red — the repo's standing defense against vacuous assertions.
6. **Accepted costs stated, not hidden** (`DoD.md` §2.3: comments explain intent, invariants, and non-obvious constraints).

## 2. Problem being solved

Issue **#2807**. A class field declared without a type annotation but assigned a constructed value — `var conn; Host() { conn = Alpha(); }` in Dart, `this.svc = new Alpha()` in TS/JS, `self.svc = Alpha()` in Python, `@svc = Alpha.new` in Ruby, and the Swift/Kotlin equivalents — carried no type in the scope graph. Any later call through that field (`conn.inner()`, `this.svc.ping()`) therefore resolved to nothing: a **missing** `CALLS` edge, invisible to `impact`, `query`, and every downstream flow.

The PR mints a type binding for such a field from its construction site, in eight languages, at parse-time capture emission. That is the whole feature. Everything else in the diff — the shadow discipline, the static-field guards, the block-boundary rule — exists to keep the newly minted binding from being applied to the *wrong* name.

## 3. Current PR state

**OPEN.** Not a draft, not merged, not closed. Head `ab2f48c17` is the tip of `fix/2807-inferred-field-receiver`; the branch is **0 commits behind `origin/main`**.

**Merge-state classification: `review blocked`.**

- `mergeable: MERGEABLE` — no content conflicts. Confirmed structurally: the merge base `b2cd1c2ad` is an ancestor of `origin/main`, and `git rev-list --count HEAD..origin/main` is `0`, so the branch is fully up to date with main.
- `mergeStateStatus: BLOCKED` with a `MERGEABLE` tree means the block is a branch-protection condition (required review or required check), not a conflict.
- `isCrossRepository: true` — a fork PR. Per verified prior context, **Vercel always fails on fork PRs** ("Authorization required to deploy") and is non-actionable, which is a known contributor to a `BLOCKED` status on this repo.
- I could not enumerate the exact unsatisfied required checks or reviewer state from a read-only worktree. See Open questions.

Hygiene checks, all clean at `ab2f48c17`:

```
git diff --check origin/main...HEAD        → no output (no whitespace errors)
git grep -nP '[\x{202A}-\x{202E}\x{2066}-\x{2069}]'  → no output (no bidi controls)
```

Non-ASCII is present in changed source but is entirely visible punctuation in comments (`──` section rules, `—`, `→`), matching established repo style throughout `gitnexus/src/core/ingestion/`. Nothing hidden, nothing misleading. Not a block.

- **Issue #2807** is the direct parent. This PR is the fix.
- **This PR already survived one full tri-review**, which found and fixed four P1s: the context-free TS `this.<field> = new X()` hoist (`db593ad93`), the Dart bare-name shadow guard seeing only local declarations (`8fc96181a`), Ruby class-object ivars bound as instance fields (`d4d027837`), and the `ids.ts` sibling-label retry not mirrored in the #2699 position key (`b07a39feb`). Those are **not** re-reported below as new.
- **#2701** contributed `Scope.ownsReceivers`, the language-neutral receiver-masking primitive. The round-3 Dart fix reuses it rather than inventing a mechanism — the correct architectural move under `DoD.md` §2.2.
- **#2766** established the exact-value `SCHEMA_BUMP` pin convention after eight prior collisions and two exact clashes, both of which landed in the window *between* review and merge. That history is why finding S6 exists.
- **`compound-receiver.ts`** encodes the missed-edge-over-wrong-edge doctrine this entire line of work serves. It is cited in the code by every fix in this PR.

**Branch-hygiene classification: `merge-from-main commit present but harmless and merge-safe`.**

**Correction to the review brief:** the branch was **not rebased**. It was **merged from main**, four times:

```
d3a296902  Merge branch 'main' into fix/2807-inferred-field-receiver
188730114  Merge branch 'main' into fix/2807-inferred-field-receiver
6ba46b217  Merge branch 'main' into fix/2807-inferred-field-receiver
72f9a9523  Merge remote-tracking branch 'origin/main' into fix/2807-inferred-field-receiver
```

Assessed as harmless: every commit those merges brought in is a Dependabot bump on main (`@hono/node-server`, `aiohttp`, `fast-uri`, `@modelcontextprotocol/sdk`, `hono`, `node-addon-api`, npm_and_yarn group), none of which appears in this PR's own 33-file diff. The `merge-base..HEAD` diff contains **no dependency churn** — it is entirely receiver-typing source and tests. Domains are causally connected: all 33 files serve one feature. No split or rebase is required.

One consequence worth stating: because these are merges rather than rebases, **the merge base moved** — it is now `b2cd1c2ad`, not the `ca294e8cd` the review lanes ran against. Every inline-comment anchor in the appendix was recomputed against the current base for exactly this reason.

Three layers, in dependency order:

1. **Mint the binding.** Per-language capture emitters recognize a field whose type comes only from a construction assignment and emit a `@type-binding.*` capture that lands on the owning Class scope, so `findReceiverTypeBinding` can type a bare-name or `this.`-rooted receiver through it. Eight languages.
2. **Do not mint it on the wrong owner.** Guards for every way the anchor can lie about ownership: TS/JS static field initializers and `static constructor()` (`isStaticClassFieldBinding`, `typescript/captures.ts:290,454-462`; `javascript/captures.ts`), Dart static members, Ruby class-object ivars, and Dart bodies that rebind the field's name (`collectDartBodyShadows`, enumerated against `vendor/tree-sitter-dart/grammar.js` rather than against reported shapes).
3. **Do not *read* through a name the body rebinds** — added at `ab2f48c17`, and the subject of the largest change since the draft. See §9.1.

The persistence side moves with it: `SCHEMA_BUMP` 40 → 41 plus its exact-value pin, because all of the above is parse-time capture emission.

## 8. Findings

Identifiers are stable across drafts. **S3 was open in the draft synthesis and is now FIXED** — it has moved to §8.2 and its inline-comment candidate has been withdrawn. S8 is new from this verification pass.

### 8.1 Still open

---

#### S1 — Swift custom postfix operator ending in `!` is peeled as a force-unwrap · P2 · **STRONG (cross-engine)**

- **Risk:** A confidently **wrong** receiver owner — the failure mode this whole line of work exists to prevent. `call-analysis.ts:598` gates the `postfix_expression` peel on `return node.text.trimEnd().endsWith(operator)`, a text test on the whole node's source. Swift permits user-defined postfix operators and its operator character set includes `!`, so `postfix operator +!` makes `(left +!).hit()` a `postfix_expression` whose text ends in `!`. The peel fires and the receiver is typed as `left` rather than as whatever `+!` returns. This defeats the exact protection the gate's own docblock four lines up claims to provide (`call-analysis.ts:565-567`: *"user-defined postfix operators. Those can return anything"*).
- **Evidence to check:** `gitnexus/src/core/ingestion/utils/call-analysis.ts:598` (the text test); `:580` (the guarded table entry `['postfix_expression', '!']`); `:563-569` (the docblock it contradicts). `grep -rln "postfix operator" gitnexus/test/ gitnexus/bench/` returns **nothing** — no fixture exists for this in the repo.
- **Corroboration:** Codex `ce-code-review` finding #2 (P2, conf 75) **and** the Claude correctness lane's residual risk #2 — the only open item both engines reached, and the one place in this digest where cross-engine agreement is genuine.
- **Honesty note:** neither lane wrote a Swift fixture; both flagged it as unprobed. The code path is unambiguous from source, but the trigger is asserted from the Swift grammar, not measured. Rare source, but rarity is the reason this is P2 rather than P1 — it is not a reason to leave it unstated.
- **Recommended fix:** inspect the operator child node rather than the enclosing node's text — accept only a `bang` token (Swift `!`) or Kotlin `!!`, and reject any multi-character custom operator. Add a Swift custom-postfix regression fixture.
- **Blocks merge:** no.

---

#### S2 — Kotlin `a!!.m().n()` emits only the TAIL edge; the head call is silently lost · P2 · **MODERATE (single Claude lane, reproduced by it; gate-blindness verified by me)**

- **Risk:** An internally **inconsistent graph**. The peel *does* type the receiver — that is the only way `.compute()` reaches `Inner.compute#1`, which is findable only through `Outer.inner`'s return type — yet `inner()` itself produces no `CALLS` edge. Measured by the adversarial lane:

  | fixture | `Host.run#1 → …` |
  |---|---|
  | `var p: Outer = Outer(); p.inner().compute(x)` (control) | `Outer.inner#0`, `Inner.compute#1` |
  | `var p: Outer? = null; p!!.inner().compute(x)` | **`Inner.compute#1` only** |

  Consequence: `impact({target: "Outer.inner", direction: "upstream"})` will not list `Host.run`, while `Inner.compute` — reachable only *through* `Outer.inner` — will.
- **Evidence to check:**
  - `gitnexus/src/core/ingestion/utils/call-analysis.ts:580` — the `['postfix_expression', '!']` table entry, with `unwrapTransparentReceiver` at **`:612-621`** (612 is its JSDoc line).
  - `gitnexus/bench/receiver-resolution/measure.mjs:1031` — the Kotlin row is `{ id: 'nonNullAssert', member: 'save', body: 'svc!!.getUser().save()' }`. `member` is the **tail**. The bench never asks about `getUser`, so the rebaselined `"nonNullAssert": "RESOLVES"` in `baseline.json` is true and still hides half the chain. (The Swift row at `:196` has the same shape.)
  - `grep -n '!!' gitnexus/test/integration/resolvers/inferred-field-receiver-matrix.test.ts` → **no output at head**. There is still no `!!` row in the matrix.
- **Note the asymmetry** (adversarial lane): Swift's `self.p!.inner().compute(x)` is the mirror image — it keeps `Outer.inner#0` and loses `Inner.compute#1`. One shared peel, two grammars, opposite halves surviving. That is a signal the peel interacts with something downstream that neither language's test covers.
- **Recommended fix:** the bug is likely **not** in `unwrapTransparentReceiver` — investigate where the head link is dropped for a peeled Kotlin receiver (the chain-link emitter, not the wrapper table). At minimum add a Kotlin `!!` matrix row asserting **both** links, so the bench's tail-only verdict stops being the only evidence.
- **Blocks merge:** no. The failure direction is a *lost* edge; the wrong-edge doctrine is not violated. But it should not ship without a tracking issue, because both existing gates are structurally blind to it.

---

#### S4 — Python: a dotted *method* callee is accepted as a type name · P3 · **MODERATE (both Claude lanes; verified by me)**

- **Risk:** A **fabricated** edge. `python/receiver-binding.ts:145` rejects only a callee rooted at the receiver — `if (text === receiverName || text.startsWith(`${receiverName}.`)) return undefined;`. Any other dotted callee is returned whole and handed to `resolveTypeRef`, which resolves dotted names through `QualifiedNameIndex`; that index matches the trailing segment against a class of that name even when the callee is a method on an unrelated object.

  ```python
  class Alpha:
      def ping(self): return 1
  class Factory:
      def Alpha(self): return "not an Alpha"    # a METHOD
  class Host:
      def __init__(self, f): self.svc = f.Alpha()   # svc is a str
      def run(self): return self.svc.ping()
  ```
  Expected: no edge. Measured by the lanes: `Host.run → Alpha.ping`.
- **Evidence to check:** `gitnexus/src/core/ingestion/languages/python/receiver-binding.ts:139-146` — the docblock above the guard names the displacement hazard and then closes only the receiver-rooted spelling.
- **Corroboration:** Claude correctness F4 (`[reproduced]`, conf 75) + Claude adversarial F8. Codex did not reach it.
- **Why P3:** it needs a method whose name collides with a class name. The common `self.svc = factory.build()` correctly emits nothing (measured), and `self.u = models.User()` is the reason the dotted arm exists at all.
- **Recommended fix:** restrict the dotted arm to callees whose root segment is a known module/import alias (the shape the docblock actually cites), or let the resolver reject a resolution whose target symbol kind is not a type. The bare-`identifier` arm is unaffected.
- **Blocks merge:** no.

---

#### S5 — Python: same-tier last-write-wins lets a weak dotted candidate erase a real constructor binding · P3 · **MODERATE (both Claude lanes — adversarial F8 reproduced this combined with S4; code verified by me)**

- **Risk:** A **lost** edge. `python/receiver-binding.ts:243` reads `if (existing === undefined || tier >= existing.tier) {` — `>=` lets a later same-tier candidate displace an earlier one. `self.conn = Outer()` followed by `self.conn = Registry.get()` in one `__init__` types the field as `"Registry.get"`, which resolves to nothing, so `Host.run` emits no edge; removing the second line restores `Host.run → Outer.inner#0`.
- **Evidence to check:** `gitnexus/src/core/ingestion/languages/python/receiver-binding.ts:243`; the tier ladder at `:238-242`; the docblock at `:139-144` which names this exact hazard ("a later such assignment DISPLACED an earlier real `self.p = Outer()`") and then closes it only for the receiver-rooted spelling.
- **Evidence level:** stronger than the draft credited. The adversarial lane (F8) reproduced the **combined S4+S5 scenario** end-to-end, not just the S4 half, so this is not a code-read-only finding.
- **Contained:** the walk skips `if`/`for`/`try`/`match`, so only straight-line double assignment reaches it, and the failure direction is a lost edge rather than a wrong one.
- **Recommended fix:** give a dotted callee its own tier below `TIER_CONSTRUCTOR`, or require a strictly greater tier unless the evidence kind matches. Fixing S4 resolves this as a side effect.
- **Blocks merge:** no.

---

#### S6 — `SCHEMA_BUMP` must be re-checked against `origin/main` at the moment of merge · procedural · **STRONG (risk-architect lane + verified by me)**

- **Risk:** A **silent** stale warm cache. Every language change in this PR is parse-time capture emission, and `analyze` skips tree-sitter for byte-unchanged chunks — a cache written under the old schema replays the old capture set with no error and no log. For the round-3 fix specifically that means it keeps serving the **wrong edge** the fix removes. The risk is time-dependent, not code-dependent: `parse-cache.ts` records eight prior collisions and two exact clashes, and both clashes landed in the window *between* review and merge.
- **Evidence to check — verified right now at `ab2f48c17`:**
  - `gitnexus/src/storage/parse-cache.ts:197` → `const SCHEMA_BUMP = 41;` (branch)
  - `git show origin/main:gitnexus/src/storage/parse-cache.ts` → `:177` `const SCHEMA_BUMP = 39;` (main)
  - `gitnexus/test/unit/incremental-parse-cache.test.ts:121-122` → pin `toBe(41)`, agrees with the constant.
  - **No collision as of this synthesis.** Note this was re-checked *after* the last merge from main, so it reflects current `origin/main`, not the state the review lanes saw (which was 40 vs 39).
  - The file's own comment at `:196` says it in capitals: `RE-CHECK AGAINST origin/main IMMEDIATELY BEFORE MERGING.`
- **Recommended fix:** whoever merges runs `git show origin/main:gitnexus/src/storage/parse-cache.ts | grep 'const SCHEMA_BUMP'` and confirms it is **not** `41`. If main has taken `41`, bump this branch to `42` **and** update the pin test — the constant and the pin must move together, or the pin test fails and the collision is caught anyway.
- **Blocks merge:** **maybe** — it blocks only if `origin/main` has moved to `41` by merge time. This is the sole procedural gate on this PR.

---

#### S7 — Accepted-cost residue of the static-field fix: static bindings are dropped, not namespaced · P3 · advisory

- **Risk:** A **missed** edge. `typeBindings` has one map per Class scope with no static/instance split, so a static field cannot be recorded separately — it is dropped. Consequence: a TS/JS `Host.p.hit()` **static receiver chain** loses typing.
- **Evidence to check:** `gitnexus/src/core/ingestion/languages/typescript/captures.ts:290` (`isStaticClassFieldBinding`), `:454-462` (both anchors tested); the counterweight row `static-field-declaration-still-types-its-receiver` in the matrix, which guards against widening this into "drop every static binding".
- **Assessment:** this is the correct trade under `compound-receiver.ts` — a missed edge over a wrong one — and it is stated openly in the fix commit rather than hidden. Reported so it is visible.
- **Note on the brief's framing:** the brief listed "static/instance source-order collision" as still open. That is now precise only in the *missed-edge* direction. The **wrong-edge** form is closed: with the static binding dropped, `p = new Alpha(); static p = new Beta();` leaves only the Alpha binding and `this.p.ping()` correctly resolves to `Alpha.ping`.
- **Recommended fix:** none required for this PR. If static receiver chains matter later, the real fix is a static/instance namespace split in `Scope.typeBindings`, which is a separate change.
- **Blocks merge:** no.

---

#### S8 — The round-3 fix pins 2 of the 7 binder shapes it reports reproducing · P3 · **NEW (this verification pass)**

- **Risk:** Regression coverage narrower than the defect. The `ab2f48c17` commit message states the read-side defect reproduced in **seven** binder shapes — for-in (`final` and `var`), untyped formal parameter, plain local `var`, catch binding, closure parameter, and record pattern. Five new matrix rows landed, of which only **two are defect rows** (`loop-var-read-does-not-see-the-field`, `pattern-read-does-not-see-the-field`); the other three are overreach controls. A future change that narrows `collectDartBodyShadows` for, say, catch bindings or closure parameters would not turn a row red.
- **Evidence to check:** `gitnexus/test/integration/resolvers/inferred-field-receiver-matrix.test.ts:1742-1795` (all five new rows). Row-name count at head vs `e4f78c023`: **79 → 84, exactly +5** — the delta the commit claims. (The commit reports absolute counts 80 → 85; my `grep -c "name: '"` counting method does not see the grammar-coverage `it()`. The delta is what matters and it verifies.)
- **Mitigating, and why this is P3 not P2:** all seven shapes route through the *same* `collectDartBodyShadows` enumeration, so a row per shape would be testing one mechanism seven times. **But be precise about what the grammar-derived coverage test actually guards:** it reads tree-sitter's `nodeTypeInfo` and gates only node types matching `*pattern*` (`:1932-1944`). It therefore covers the pattern family and **not** the catch-binding, closure-parameter, or plain-local shapes — narrowing `addDartBinderName`'s catch-binding arm would turn no row red. Coverage here is by mechanism for the non-pattern shapes and by grammar-derived guard for the pattern family only. Defensible, and under-stated in the commit rather than wrong — but do not read the guard as covering all seven.
- **Recommended fix:** add rows for the catch binding and the closure parameter — the two shapes furthest from the two pinned ones — or state in the test file that mechanism coverage is deliberate.
- **Blocks merge:** no.

### 8.2 Fixed — verified by reading shipped source, not by trusting commit messages

**S3 (open in the draft synthesis; now FIXED at `ab2f48c17`) — Dart shadow analysis was write-side only, so a bare-name READ of an untypeable local resolved to the field's new type.** This was the round's most serious open item and the one the critic refused to pass the PR with. It is closed. What I verified:

- **The primitive is pre-existing and language-neutral.** `dartShadowedFieldsCapture` (`dart/captures.ts:865-887`) publishes the mask as `Scope.ownsReceivers` — the #2701 primitive — rather than inventing a Dart-specific mechanism. `languages/dart.ts:143-148` wires it through the `scopeOwnsReceivers` `LanguageProvider` hook. No language name enters shared ingestion code; `DoD.md` §2.2 holds.
- **Precedence is right, and I read the walker to confirm it rather than taking the claim.** `scope-resolution/scope/walkers.ts:243-245` reads `typeBindings` **first** at every scope and honours the mask only after: `const typeRef = scope.typeBindings.get(receiverName); if (typeRef !== undefined) return typeRef;` then `if (scope.ownsReceivers?.has(receiverName) === true) return undefined;`. `isReceiverOwnedButUnbound` at `:223-224` has the same order. So a shadow the resolver **can** type still wins — the mask fires only where the alternative was a fabricated type. Both halves of that claim are pinned by their own rows (`annotated-param-read-keeps-its-own-type`, `typed-local-read-keeps-its-own-type`).
- **Plumbing follows an existing precedent.** The marker rides the *same* synthesized match as `@scope.function` (`dart/captures.ts:203-211`) so it lands on that member's Function scope, and sits outside the `@scope.` namespace so `anchorCaptureFor` cannot mistake it for the anchor — the TypeScript `@receiver-owner.this` precedent.
- **The mask is `shadows ∩ fields` and nothing wider** (`:880-886`), returning `undefined` rather than an empty marker when nothing is masked, so the emitted capture set is unchanged for every body that does not shadow a field; names are sorted so capture text and every fingerprint over it stay order-stable.
- **Three accepted costs are documented on the function, not buried** (`:833-858`): the wider library-level case (`var logger = Logger();` shadowed by a loop variable) is **left open** as a known limitation; an *annotated* field shadowed by a binder is masked too, so this reaches resolution that predates #2807; and `mixin` bodies are reached, because the grammar gives a mixin a `class_body`. (`extension_body` is a different node and is left alone.) The body-wide granularity is deliberate symmetry with the write side.
- **Mutation-verified in both directions**, which is the part that matters: unwiring `scopeOwnsReceivers` turns exactly 2 rows red (83/85 green); over-widening the mask by dropping the `shadows.has` test turns 28 Dart rows red **including** the `unshadowed-read-in-a-shadowing-class-still-resolves` control. The controls stay green under the first mutation by design — they guard overreach, not the defect — and the second mutation is what proves they are live. That two-direction design is visible in the test file's own comments at `:1757-1765`.
- **`SCHEMA_BUMP` moved with it**, 40 → 41 plus the pin, with the reason written out at `parse-cache.ts:190-196`. Re-checked against `origin/main` after the last merge: branch 41, main 39, pin 41. See S6.
- **Not re-measured by me, and flagged as such:** the reported sweep totals (3220 passed / 1 skipped / 64 files, exactly +5 over 3215), the `tsc` clean run, the four green capture gates, and the performance number below are commit-message claims. I verified the *source*, the *test rows*, the *row delta*, and the *`SCHEMA_BUMP` triple* directly; I did not re-run the suite or the bench.
- **Performance, stated with its method:** Dart `scope-capture` large corpus, median of 3 — 405.6 ms with the mask vs 390.0 ms without, **≈ +4%**. Fingerprint and `capture_groups` are byte-identical across both arms, so **no corpus fixture emits a mask at all** — the 4% is the cost of the *check* alone. `bench/scope-capture` cannot report this: it gates the scaling *ratio* and is blind to a linear constant factor. Method caveat, stated because the commit states it: 3 samples per arm, blocked rather than interleaved. An estimate, not a rigorous number.

The three round-2 fixes, all still present and verified at head:

1. **Dart 3 pattern binders** (P1; Claude correctness F1 + Claude adversarial F1; **Codex missed it entirely**) — `addDartBinderName` (`dart/captures.ts:942`) enumerates the pattern family from `vendor/tree-sitter-dart/grammar.js`, plus a grammar-derived coverage guard reading `nodeTypeInfo` (`inferred-field-receiver-matrix.test.ts:1932-1944`) that fails if the grammar declares a `*pattern*` type the fixtures do not exercise. *Accepted cost:* deliberately over-approximate — a bare name inside a pattern shadows whether it binds or merely references a constant, so `case kLimit:` shadows too, costing a field type where Dart would not have.
2. **Ruby block-receiver `self` rebinding** (Claude lanes rated P1; **Codex #3 rated it P2** — both engines agree it is real, they disagree on severity) — fixed **structurally**: crossing *any* block boundary makes ownership unprovable, so the binding is discarded — `do_block`, `block`, `lambda`, `begin_block`, `end_block` all return `false` (`ruby/captures.ts:181-192`). The docblock argues why a deny-list of rebinding call names cannot work: the set is open, since `def helper(&blk) = Foo.class_eval(&blk)` lets any user method rebind a block's `self`. *Accepted, asserted cost:* `[1].each { @shared = X.new }` inside a real instance method is dropped too; it carries its own row `plain-block-self-ivar` so the loss is visible rather than silent.
3. **Static field initializers** (**the one finding both engines rated P1** — Codex #1 at P1/75, plus Claude correctness F3 and adversarial F3/F4/F5) — `isStaticClassFieldBinding` (`typescript/captures.ts:290`) drops static bindings on the TS `constructor` **and** `annotation` anchors (`:454-462`), the JS `constructor` anchor plus a `static constructor()` guard (`javascript/captures.ts`, importing the shared predicate at `:42`), and a Dart static-member guard. *Accepted cost:* S7.

### 9.1 The recurring defect shape — now four instances, and the fourth is this PR's own

**A second review round found three more wrong-owner defects, and every one was in code the first round had already fixed.**

| # | defect | first-round fix it reopened |
|---|---|---|
| 1 | Dart 3 pattern binders invisible to the shadow set | `8fc96181a` "treat every binder as a field shadow" |
| 2 | Ruby `def`/ivar write inside a block published on the wrong class | `d4d027837` "don't bind a class-object @ivar" |
| 3 | `static` field initializer retypes its instance twin | `db593ad93` "bind `this.field = new X()` only inside a class method" |

All three are the **same root shape: an incomplete ENUMERATION of binder or scope forms.** Round 1 fixed five Dart binder types and missed ten pattern types; fixed two Ruby `singleton_*` self-rebinders and missed every block form; fixed `this.x =` inside a static *method* and missed the static *field* initializer.

**A third round added a fourth instance — and it is a variant of the shape, not a repeat.** S3 was not an incomplete enumeration of *forms*; it was an incomplete enumeration of *sites*: `collectDartBodyShadows` had exactly one call site, in the bare-name **write** branch, and nothing consulted it on the **read** side. The same feature's rule, correct on one side and absent on the other. Round 2 widened *what* the shadow set contained without changing *where* it was consulted — which is why round 2's own fix did not close it. That shape has now bitten this branch **four times** (formal parameters, then the three above, then this).

The encouraging part: **the fixes increasingly attack the class rather than the instance.** Three of the four now do.

- **Ruby, structurally.** Crossing any block boundary makes ownership unprovable, so the binding is discarded — no deny-list, because the set of rebinding call names is open.
- **Dart binders, by grammar-derived coverage guard.** The suite reads tree-sitter's `nodeTypeInfo` and fails if the grammar declares a `*pattern*` type the fixtures do not exercise. A grammar bump adding an 11th type turns the suite red instead of silently reopening the bug a third time.
- **Dart read side, by reusing the language-neutral primitive.** `Scope.ownsReceivers` already existed for exactly this question, and the walker's `typeBindings`-first ordering is a property of the *walker*, not of Dart — so the same masking is now available to any language that sets the hook.

That leaves the *static/instance* fix (round 2 #3) as the one still enumerated rather than structural — the residue S7 names.

### 9.2 Cache and persisted-data safety

Correct, and unusually well-argued in the source itself. `parse-cache.ts:190-196` explains why v41 is bump-or-nothing: a v40 warm cache replays scope matches with no `@receiver-owner.shadowed-fields` marker, the receiver walk still reaches the class field, and it therefore **keeps serving the exact wrong edge this bump's fix removes** — silently. The pin test moved with the constant. The residual exposure is purely the merge-window race, S6.

### 9.3 Gate coverage — what the four gates can and cannot see

The PR's own author documented two gate blind spots rather than letting them pass as green:

- `bench/scope-capture` gates the **scaling ratio** and is blind to a linear constant factor, so the +4% capture-emission cost is invisible to it. Stated in the commit because the gate cannot state it.
- `bench/receiver-resolution` asserts on the **tail** member of a chain, which is exactly why S2 (the lost Kotlin head edge) shows `RESOLVES` truthfully while hiding half the chain.

Both disclosures are the correct behavior under `DoD.md` §2.3. Neither is a finding against the PR; S2 is a finding against the *gate*.

### 9.4 Accepted costs, consolidated

Every one is stated in source or commit message rather than hidden — the property most worth preserving about this PR:

| cost | where stated | direction |
|---|---|---|
| Dart pattern shadow over-approximates (`case kLimit:` shadows) | `dart/captures.ts:909-917` | missed edge |
| Ruby block ivar dropped even in a real instance method | `ruby/captures.ts` + row `plain-block-self-ivar` | missed edge |
| Static field binding dropped, not namespaced (S7) | commit + counterweight row | missed edge |
| Library-level shadowing left open by the read-side mask | `dart/captures.ts:838-844` | wrong edge, **known and unclosed** |
| Annotated field shadowed by a binder is masked too | `dart/captures.ts:851-858` | missed edge |
| `mixin` bodies reached by the mask | `dart/captures.ts:855-857` | missed edge |
| Mask is body-wide, so a genuine field read elsewhere in the member loses its edge | `dart/captures.ts:845-850` | missed edge |
| +4% Dart capture emission, invisible to its gate | commit, with method caveat | performance |

The one entry in the **wrong edge** direction is the library-level case, and it is explicitly recorded as a known limitation with a reason (it changes resolution for code this feature never touched). That is a defensible scope boundary, not an oversight — but it is the one accepted cost that does not fail safe, and it should carry a follow-up issue.

### 9.5 Refuted

Findings checked and not triggerable, with the reason. The two Claude lanes each ran their own refutation battery (correctness R1–R10; adversarial 1–6) with fixtures executed through the real `runPipelineFromRepo`; I did not re-derive those from scratch and mark them lane-refuted rather than synthesis-refuted.

| finding | verdict | reason |
|---|---|---|
| Adversarial F3 — Dart static member bare-name write binds the instance field | **now FIXED, not open** | `dart/captures.ts` static-keyword guard; blame lands in `e4f78c023` |
| Adversarial F4 — JS `static constructor()` has no static gate where TS grew one | **now FIXED, not open** | `javascript/captures.ts` `hasKeyword(node, 'static')` guard; blame lands in `e4f78c023` |
| Adversarial F5 — TS/JS static field initializer types `this.<name>`, incl. the pre-existing annotation form | **now FIXED, not open** | both anchors guarded, `typescript/captures.ts:454-462` |
| Risk-architect B — no observability for a wrong inferred binding | **out of scope** | pre-existing; a wrong owner produces a wrong `CALLS` edge with no log or counter, but this PR did not introduce that |
| Correctness R1 / adversarial 1 — TS nested class / class expression hoists `this.x =` onto the outer class | lane-refuted `[reproduced]` | class expressions get their own `Class` scope; `walkToScope(…, 'Class')` stops there |
| Correctness R2 — a static form escaping `isStaticMethodThis` / `hasKeyword` | lane-refuted `[reproduced]` | six shapes probed (`static async`, `static get`, `public static`, `static *`, `static ['make']`, multi-line); `hasKeyword` skips the `name` field so `get static()` is read correctly |
| Correctness R5 / adversarial 3-4 — Swift `swiftEnclosingTypeName` + `ids.ts` sibling retry aliasing two callables | lane-refuted `[reproduced]` | position key resolves before the qualified retry; both fail-closed `localNameKey` arms hold; the dot gate keeps a bare `save` from crossing labels |
| Correctness R6 — Python DFS visits in reverse source order | lane-refuted `[code-read]` | children pushed in reverse index order so the LIFO pop visits source order; comment and code agree |
| Correctness R7 — `interpret.ts` arm reorder changes the source label for another producer | lane-refuted `[code-read]` | `@type-binding.instance-field` has exactly one producer and emits `@type-binding.constructor` only on a disjoint branch |
| Correctness R9 — `bodyNode.previousNamedSibling` is not always `method_signature` | lane-held `[AST dump]` | verified for plain method, generative ctor, named ctor with initializer list, factory, static, getter, setter. Still the load-bearing assumption at `dart/captures.ts:925-928` |
| Adversarial 2 — Ruby `module` nested in a `class` hoists the ivar onto the outer class | lane-refuted `[reproduced]` | `ruby/query.ts` maps `(module) @scope.class`, so the walk stops at the module |
| Adversarial 5 — the `!` gate fires on a non-type-preserving postfix operator in a *non-Swift* grammar | lane-refuted | only `vendor/tree-sitter-{swift,dart,kotlin}` define `postfix_expression`; Dart's `postfix_operator` is `++`/`--`, which cannot end in `!`. **Swift itself is the exception — that is S1** |
| Adversarial 6 — the weakest tier (`constructor-inferred`) displaces a stronger binding | lane-refuted `[reproduced]` | Python `conn: Outer` + `self.conn = Alien()` → `Outer.inner#0`; TS `private p: Outer` + `this.p = new Alien()` → `Outer.inner#0` |
| Test-CI lane — vacuous matrix rows | lane-refuted | every empty-assert row is paired with a positive row on the same class, and guard rows use an `Alien.inner()` witness so a regression *swaps* the target rather than emptying the set |

### 9.6 A defect none of this review round caught — CI did

Recorded because it bounds how much confidence this digest is worth.

After every lane had reported and after the synthesis above was written, the
repository's own `grammar-literal-validation` gate failed on CI:

```
1 dead grammar literal(s) found:
  - node-type "field_definition" — languages/typescript/captures.ts:0 — not valid in [typescript]
```

Introduced by round 2's own static-field fix. `isStaticClassFieldBinding` held both
spellings in one set — `public_field_definition` (TypeScript) and `field_definition`
(JavaScript) — so one predicate could serve both languages. The predicate lives in
`typescript/captures.ts`, and the gate checks each literal against the grammar of the
file it appears in.

The literal was **not** dead code: `javascript/captures.ts:42` imports the predicate and
calls it against real JS nodes, so the guard worked. The gate is still right to fail it,
for exactly the reason the predicate's own docblock gives for preferring `hasKeyword`
over a node-type test — *"a node-type test silently stops firing on a grammar bump and
every static field starts retyping its instance twin again."* A literal already dead in
its own file is that failure shipped pre-broken, and nothing on the TypeScript side would
ever have reported it.

Fixed in `3f4975f99`: each language names its own node type and passes it in; the
`hasKeyword` logic and the static/instance reasoning stay shared and unchanged.

**Why this belongs in a review digest.** Five review lanes across two engines read this
code and none flagged it. Nor did the author's local sweep, which was scoped to
`test/integration/resolvers` and `test/integration/cfg` — the gate lives in the parent
directory. Scoping a sweep to the subdirectories a change touches is how a cross-cutting
gate gets skipped, and it is a more reliable failure than any single reviewer's attention.
The repo's mechanical gates outperformed the entire review round here; weight the
still-open findings below accordingly.

Facts checked directly rather than asked about or assumed:

- **The branch was merged from main, not rebased.** `git log --merges` on `merge-base..HEAD` returns four merge commits. The review brief said "rebased twice". This changes the branch-hygiene classification (§6) and is why every inline anchor was recomputed.
- **The merge base moved to `b2cd1c2ad`.** The lanes reviewed against `ca294e8cd`. Every inline-comment line was re-tested for hunk containment at the new base; two needed re-anchoring (see appendix).
- **`SCHEMA_BUMP` is 41 on the branch and 39 on main, with the pin agreeing.** Read all three, after the last merge. No collision now — and the *now* is the whole content of S6.
- **The walker really does read `typeBindings` before the mask.** `walkers.ts:243-245`, read rather than trusted, because the entire correctness argument for the S3 fix rests on that ordering.
- **Codex rated the static-field finding P1 and Ruby P2.** Read from `codex-ce.md:17467` rather than inferred from the Claude lanes' severities. The draft synthesis had this inverted; §8.2 now states it correctly.
- **The matrix row delta is exactly +5** (79 → 84 by row name at `e4f78c023` vs head), matching the claimed delta. My absolute count differs from the commit's by one because of counting method; the delta verifies.
- **No Kotlin `!!` row exists in the matrix** and **no Swift custom-postfix fixture exists anywhere** in `test/` or `bench/` — both greps returned empty at head, so S1 and S2 are unprobed at head, not merely unprobed when the lanes ran.
- **The Kotlin bench row is tail-only**, `measure.mjs:1031`, `member: 'save'` on `svc!!.getUser().save()` — read rather than assumed from the lane's report.
- **Hidden-Unicode and whitespace checks are clean**; the non-ASCII that exists is visible comment punctuation matching repo style throughout the ingestion tree.

## 11. Open questions

Only what could not be resolved by verification:

1. **What exactly is blocking the merge?** `mergeStateStatus: BLOCKED` with `mergeable: MERGEABLE` and zero commits behind main is a branch-protection condition, but a read-only worktree cannot distinguish "required review missing" from "required check failing/pending". Given `isCrossRepository: true` and the known non-actionable Vercel failure on fork PRs, a required-check state is the likely cause — but that is inference, not verification. Whoever merges should read the check-run list directly.
2. **Should the library-level shadowing case get an issue before merge?** It is the one accepted cost in the wrong-edge direction (§9.4) and is documented as a known limitation. Filing it is cheap and keeps it from decaying into an undocumented behavior.
3. **Does S2's asymmetry (Kotlin loses the head, Swift loses the tail) indicate one bug or two?** Both lanes measured it; neither traced it to the chain-link emitter. The recommended investigation (§8.1 S2) may find a single shared cause.

## 12. Final verdict

**production-ready with minor follow-ups**

The feature is implemented end-to-end in the real runtime path across eight languages, and the failure mode this line of work exists to prevent — a confidently wrong `CALLS` edge — has no open instance in the tree. The round's most serious open item, the Dart read-side shadow gap (S3), is fixed at `ab2f48c17` using the pre-existing language-neutral `Scope.ownsReceivers` primitive rather than a new mechanism; I verified the fix by reading the walker's `typeBindings`-before-mask ordering, the capture emitter, and the five new matrix rows, not by trusting the commit message. Seven findings remain open; six of them cost a *missed* edge or a fabricated edge behind a name collision (S2, S4, S5, S7, S8) or need source the repo has never had a fixture for (S1), and none of those six blocks merge. Every accepted cost is stated in source or commit message rather than hidden, including the one that does not fail safe (library-level shadowing), and the +4% capture-emission cost that its own gate is structurally unable to report. The sole procedural gate is S6: `SCHEMA_BUMP` is 41 on this branch against 39 on `origin/main` right now, and because the risk is time-dependent rather than code-dependent, the merger must re-read `origin/main` at the moment of merge and bump to 42 if main has taken 41. Confidence in the "no open wrong-owner defect" claim is bounded by the provenance limit stated at the top: only one Codex leg ran, it missed the Dart pattern-binder defect entirely, and the round's most severe findings therefore rest on same-engine agreement.

---

---

*Five inline comments carry the still-open findings at their anchors.*

magyargergo:
## Findings status — all seven being fixed in-PR

Following up on [the tri-review](https://github.com/abhigyanpatwari/GitNexus/pull/2810#pullrequestreview-4854823954). One correction to how I characterised the open findings there, because it matters for who owns them.

I described S1/S2/S4/S5 as things the review "surfaced". That understated it. Checking the hunks against the merge base:

| finding | hunk | verdict |
|---|---|---|
| S1 — Swift custom postfix `+!` peeled as force-unwrap | `@@ -563,0 +583,18 @@` | **pure addition** — 18 lines this PR adds |
| S4 — Python dotted *method* callee accepted as a type name | `@@ -116,0 +117,32 @@` | **pure addition** — 32 lines this PR adds |
| S5 — `>=` lets a weak candidate erase a constructor binding | same new block | **this PR** |
| S2 — Kotlin `a!!.m().n()` loses the head edge | `@@ -559,3 +571,10 @@` | modified by this PR |

A `-N,0` hunk means those lines did not exist before this branch. These are not pre-existing defects the review happened to notice — they are defects in code this PR is adding. They were rated non-blocking because each needs an unusual trigger, not because they belong to someone else.

So all seven are being fixed here rather than tracked:

- **S1** — gate on the operator, not on "text ends in `!`". Swift's operator character set includes `!`, so `postfix operator +!` currently peels and types the receiver as the operand: a confidently **wrong** owner, which is the one failure mode `compound-receiver.ts` exists to prevent. The legitimate `a!` / `a!!` peels keep working, pinned by counterweight rows.
- **S2** — find where the head link is dropped for a peeled receiver. Note the asymmetry: Kotlin `a!!.m().n()` keeps the tail and loses the head; Swift `self.p!.m().n()` loses the tail and keeps the head. One shared peel, two grammars, opposite halves surviving — a fix that doesn't explain both is only half a fix.
- **S4/S5** — the dotted arm hands `f.Alpha()` to `QualifiedNameIndex`, which matches the trailing segment against a class of that name even when the callee is a method. `self.svc = factory.build()` (no edge) and `self.u = models.User()` (resolves) are the controls that must survive.
- **S7** — being evaluated, not assumed. Namespacing static bindings may not be containable without touching shared code; if it isn't, it stays an accepted cost **with the proof attached**, rather than being quietly forced.
- **S8** — the read-side fix reproduced in seven binder shapes and pinned two. The grammar-derived coverage guard is narrower than the commit implies: it gates only `*pattern*` node types, so catch bindings, closure parameters and plain locals are covered by argument, not by a row. Rows for the rest.
- **S6** — genuinely procedural, cannot be closed by code. `SCHEMA_BUMP` is 41 here, 39 on `main`, pin agrees. Re-check at merge time: this file records eight prior collisions and two exact clashes, and both clashes landed in the window *between* review and merge.

### One thing worth recording about this review round

CI beat it. After all five review lanes reported, `grammar-literal-validation` failed on a dead node-type literal introduced by the round's own static-field fix — `field_definition` named in `typescript/captures.ts`, where the TypeScript grammar has no such node type.

The literal was not dead code: `javascript/captures.ts` imports the predicate and calls it against real JS nodes, so the guard worked. The gate is still right to fail it, for exactly the reason that predicate's own docblock gives for preferring `hasKeyword` over a node-type test — *"a node-type test silently stops firing on a grammar bump."* A literal already dead in its own file is that failure shipped pre-broken.

Five reviewers across two engines missed it. So did my local sweep, which was scoped to `test/integration/resolvers` and `test/integration/cfg` while the gate sits in the parent directory. Scoping a sweep to the subdirectories a change touches is how a cross-cutting gate gets skipped — the repo's mechanical gates outperformed the entire review round here, and the still-open findings should be weighted with that in mind.

Issue #1983, opened 2026-06-03
gitnexus analyze linux kernel source  JavaScript heap out of memory OOM.


catmeow2764-dot:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

use `gitnexus analyze` and `npx gitnexus analyze`  to analyze linux kernel source code. and failed with "Javascript out of mermory"

### Context

I try two platforms, macOS(24G ram) and ubuntu 2204 linux(64G ram). No one can complete this task.
I find some advices in some similar issue. I change the max old space size, use gitnexusignore file. It changed better, and still failed.

### Expected behavior

analyzed completely.

### Actual behavior

failed

### Steps to reproduce

pull a linux source code in your local mechine and use gitnexus to analyze.

### Environment

OS: ubuntu 2204

### Logs / screenshots

  ██████████████░░░░░░░░░░░░░░░░░░░░░░░░░░  35% | Parsing chunk 236/346... (Parsing...)^B^[[D{"level":40,"time":1780455581845,"name":"}
[10:59:41.846] WARN (gitnexus): scope extraction failed for drivers/staging/media/atomisp/pci/atomisp_trace_event.h: ScopeExtractor: n.
s
<--- Last few GCs --->

[223760:0x13aa8000]  1422632 ms: Scavenge 14738.7 (15007.9) -> 14738.7 (15007.9) MB, pooled: 0 MB, 109.87 / 0.00 ms  (average mu = 0.9
[223760:0x13aa8000]  1422756 ms: Scavenge 14738.7 (15007.9) -> 14738.7 (15007.9) MB, pooled: 0 MB, 123.82 / 0.00 ms  (average mu = 0.9

<--- JS stacktrace --->

FATAL ERROR: Scavenger: semi-space copy Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

<--- Last few GCs --->

[223760:0x13aa8000]  1422632 ms: Scavenge 14738.7 (15007.9) -> 14738.7 (15007.9) MB, pooled: 0 MB, 109.87 / 0.00 ms  (average mu = 0.9
[223760:0x13aa8000]  1422756 ms: Scavenge 14738.7 (15007.9) -> 14738.7 (15007.9) MB, pooled: 0 MB, 123.82 / 0.00 ms  (average mu = 0.9

<--- JS stacktrace --->

FATAL ERROR: Scavenger: semi-space copy Allocation failed - JavaScript heap out of memory
[...]

PR #2819, opened 2026-08-04
feat(languages): add Lua language support

Daylighttx:
## Summary

Add Lua language support to GitNexus — tree-sitter parser, scope resolver, middleclass heritage handling, and full registration in the shared pipeline.

## What changed

- **Vendored tree-sitter-lua grammar** at `gitnexus/vendor/tree-sitter-lua/` with parser, scanner, and query configurations
- **Scope resolver** (`lua/scope-resolver.ts`): resolves `require("a.b.c")` → `a/b/c.lua` via suffix-based file lookup, registered in the shared scope-resolver registry
- **Heritage handler** (`lua/heritage.ts`): detects middleclass `class("Name", Parent)` → `EXTENDS` and `function Obj:m()` → `HAS_METHOD` via AST re-parsing (middleclass has no class body for lexical analysis)
- **Language registration**: `EXTENSION_MAP` (`.lua`), `SupportedLanguages` (`Lua = 'lua'`), and `luaProvider` with parser/type/export/import/call hooks
- **Test coverage**: grammar ABI smoke, language-skip, vendored-grammars, ingestion-utils
- **Bug fix**: reworded `require(...)` in Lua query TS comment to avoid Dockerfile static scanner false positive

## Implementation notes

- Lua is the 15th language using the shared scope-resolution pipeline; no language names appear in `gitnexus/src/core/ingestion/` shared code
- Prebuilds are generated by `build-tree-sitter-prebuilds.yml` CI workflow; none committed in this PR
- The `middleclass` library convention is the primary OOP pattern detected; other Lua OOP libraries (30log, etc.) are not handled
- Phases B2 (middleclass MRO + `__base` super calls) and B3 (receiver/arity precision) are marked as TODOs in code

## Validation

- `npx tsc --noEmit`: passed
- `npm run test:unit`: 0 Lua failures (9 pre-existing environment-only failures: macOS `/var` vs `/private/var`, Linux-only `/proc/self/fd`, locale, process-lock — all present on upstream `main`)
- `git diff --check`: passed
- ESLint + Prettier (pre-commit hook): passed

## Scope boundary

- Middleclass MRO and `__base` super call resolution is deferred (Phase B2, marked in code)
- Receiver/arity precision for method calls is deferred (Phase B3, marked in code)
- LSP / IDE-specific features are not included — pure indexing and graph analysis only

---

## Summary

*This appears to be a broad language-ingestion change that adds Lua support across grammar handling, parsing, import resolution, extraction, and documentation. The risk is elevated because it modifies shared parser infrastructure and a build workflow alongside the new Lua-specific implementation.*

**🔴 CRITICAL blast radius. A Lua language-support change centered on `gitnexus/src/core/tree-sitter/parser-loader.ts` and the ingestion pipeline reaches 69 dependents.**

The change spans Lua-specific ingestion code including `gitnexus/src/core/ingestion/languages/lua.ts`, `gitnexus/src/core/ingestion/languages/lua/index.ts`, `gitnexus/src/core/ingestion/import-resolvers/configs/lua.ts`, `gitnexus/src/core/ingestion/call-extractors/configs/lua.ts`, and `gitnexus/src/core/ingestion/type-extractors/lua.ts`, while also updating shared language detection, scope resolution, import resolution, parser loading, and worker paths. Review `gitnexus/src/core/tree-sitter/parser-loader.ts` first, where 10 changed symbols make it the hottest file.

Impact is distributed across `Import-resolvers`, `Tree-sitter`, `Dart`, `Call-extractors`, `Kotlin`, `Zig`, `Resolvers`, `Scope-resolution`, `C`, `Group`, `Embeddings`, `Unit`, `Pipeline-phases`, `Cli`, `Languages`, `Python`, `Extractors`, `Workers`, and `Pipeline`. The shared changes in `gitnexus/src/core/ingestion/export-detection.ts`, `gitnexus/src/core/ingestion/import-resolvers/utils.ts`, `gitnexus/src/core/ingestion/scope-resolution/pipeline/registry.ts`, and `gitnexus/src/core/ingestion/workers/parse-worker.ts` are the key places to verify that existing language behavior remains compatible.

The HIGH risk file is `.github/workflows/build-tree-sitter-prebuilds.yml`. Review it together with `.github/vendored-grammars.json`, `gitnexus/scripts/build-tree-sitter-grammars.cjs`, and `gitnexus/src/core/tree-sitter/vendored-grammars.ts`, since these changes connect the Lua grammar to the build and parser distribution path.

_Added by GitNexus for PR #2819. Edit freely — this block is replaced on the next review, everything above it is left untouched._


koriyoshi2041:
I exercised the current head (`1c9bb81d`) through the built CLI with a two-file fixture:

```lua
-- main.lua
local util = require("lib.util")
local function run() return util.answer() end
run()
```

```lua
-- lib/util.lua
local M = {}
function M.answer() return 42 end
return M
```

After `gitnexus analyze /tmp/gitnexus-lua-review --skip-git --index-only --workers 1`, the graph had the expected `main.lua -[:IMPORTS]-> util.lua` and local `main.lua -[:CALLS]-> run` edges, but no edge from `run`/`main.lua` to `answer`. The node and import are present; the `local util = require(...)` binding is not connecting the member-call receiver to the imported module.

That seems worth resolving or making explicit before this lands because `interpret.ts` currently says `util.split`-style linking is handled, while the PR's test changes only cover registration/ABI and would not catch this graph-level gap. Could you add an end-to-end Lua fixture asserting `require` plus a cross-file member call? If receiver linking is intentionally B3, I would at least correct that implementation note and make the test assert the intended current boundary so the initial language-support contract is unambiguous.

Validation on macOS arm64 / Node 26.5.0: fresh `npm install` built the vendored Lua grammar; the four focused registry/ABI suites passed (155/155).

Daylighttx:
Hi @koriyoshi2041 — thanks again for the hands-on review. Indexing a real fixture and inspecting the graph is exactly the bar this PR should meet, and it surfaced two real gaps. Both are fixed on the rebased head (`c663dea`, base `origin/main` `a6a8aa7`).

## Gap 1 — `require()` cross-file member-call linking (the one you flagged)

`interpretLuaImport` returned `kind: 'wildcard'` for every `require()`. `collectNamespaceTargets` only registers receivers from `kind === 'namespace'` edges, so `local util = require("lib.util")` never bound `util` as a namespace receiver — `util.answer()` had no path to `answer`.

**Fix (`268fe72`).** `query.ts` adds a `local_variable_declaration` pattern capturing the LHS binding as `@import.localName` (bare `require()` pattern kept for side-effect imports); `interpret.ts` returns `{ kind: 'namespace', localName, importedName, targetRaw }` when `@import.localName` is present. The overstated comment is rewritten to describe both forms and the B3 boundary.

## Gap 2 — middleclass `EXTENDS` / `HAS_METHOD` never emitted (found while extending the test)

When I added a middleclass fixture to lock in the contract, `Dog -[:EXTENDS]-> Animal` and the `HAS_METHOD` edges were absent from the graph — even though `heritage.ts` existed and its comments claimed to emit them.

**Root cause.** `emitLuaHeritageEdges` re-read each file via `readFileSync(parsed.filePath)`, but `parsed.filePath` is **relative**. For `analyze <non-cwd-path>` (the normal case — including your `analyze /tmp/gitnexus-lua-review --skip-git` run), `readFileSync('base.lua')` resolved against the project root, ENOENT-failed on every file, and the hook silently `continue`d. EXTENDS/HAS_METHOD only appeared when `analyze .` was run from the repo root by accident. Graph-lookup calls were unaffected (filePath is just a key there); only the filesystem read broke.

**Fix (`c663dea`) — capture-side-channel pattern, mirroring Java/C/C++.** Heritage pairs are collected in `emitLuaScopeCaptures` (parse worker, where the tree-sitter AST is already live), snapshotted onto `ParsedFile.captureSideChannel` via `collectCaptureSideChannel`, and read by `emitLuaHeritageEdges` on the main thread — no file re-read, no re-parse (#1983).

- `lua/capture-side-channel.ts` (new): `LuaCaptureSideChannel` type + module map + `collectLuaCaptureSideChannel`.
- `lua/captures.ts`: runs `HERITAGE_QUERY` + `METHOD_OWNER_QUERY` against the same tree, stashes the pairs.
- `lua/heritage.ts`: reads `parsed.captureSideChannel`, emits EXTENDS + HAS_METHOD; `readFileSync`, the parser, and the local query copies are gone.
- `lua/query.ts`: `HERITAGE_QUERY` + `METHOD_OWNER_QUERY` + getters moved here.
- `lua.ts`: wires `collectCaptureSideChannel` through `assertCloneable`.

## Test cases added

`gitnexus/test/integration/resolvers/lua-scope.test.ts` (mirrors `ruby-scope.test.ts`). Two describe blocks with the fixtures inline:

```lua
-- require fixture (your exact two-file case)
-- lib/util.lua              -- main.lua
local M = {}                 local util = require("lib.util")
function M.answer()          local function run()
  return 42                    return util.answer()
end                          end
return M                     run()

-- middleclass fixture
-- base.lua                          -- dog.lua
local class = require("lib.class")   local class = require("lib.class")
local Animal = class("Animal")       local Animal = require("base")
function Animal:speak()              local Dog = class("Dog", Animal)
  return "..."                       function Dog:bark()
end                                    return "woof"
return Animal                        end
                                     return Dog
```

## Verification — actual commands and output I ran

These are the real commands and transcripts, not a summary. Reproduce from the repo root after `npm run build`.

**Unit/integration tests (the suite that covers both gaps):**
```
$ npx vitest run test/integration/resolvers/lua-scope.test.ts
 RUN  v4.1.10 /Users/tal/gitnexus-lua-pr
 Test Files  1 passed (1)
      Tests  6 passed (6)
```
The 6 named assertions: `emits IMPORTS edge from main.lua to lib/util.lua`, `resolves run -> util.answer() as CALLS edge to util.lua`, `detects answer as Method + run as Function`, `emits EXTENDS from Dog to Animal across files`, `emits HAS_METHOD from each class to its methods`, `detects Dog and Animal as Class nodes`.

**Type check:**
```
$ cd gitnexus && npx tsc --noEmit; echo "tsc exit: $?"
tsc exit: 0
```

**Graph edges — require fixture (your exact invocation, non-cwd):**
```
$ node gitnexus/dist/cli/index.js analyze /tmp/lua-rep --skip-git --index-only --workers 1
  6 nodes | 8 edges | 1 clusters | 0 flows

$ node gitnexus/dist/cli/index.js cypher "MATCH (a)-[r]->(b) WHERE r.type='CALLS' OR r.type='IMPORTS' RETURN a.name AS src, r.type AS edge, b.name AS dst" -r lua-rep
| run      | CALLS   | answer   |
| main.lua | IMPORTS | util.lua |
| main.lua | CALLS   | run      |
```
`run -[:CALLS]-> answer` is the edge you flagged as missing.

**Graph edges — middleclass fixture (the gap I found):**
```
$ node gitnexus/dist/cli/index.js analyze /tmp/lua-mc2 --skip-git --index-only --workers 1
  7 nodes | 10 edges | 1 clusters | 0 flows

$ node gitnexus/dist/cli/index.js cypher "MATCH (a)-[r]->(b) WHERE r.type='EXTENDS' OR r.type='HAS_METHOD' OR r.type='IMPORTS' RETURN a.name AS src, r.type AS edge, b.name AS dst" -r lua-mc2
| Animal  | HAS_METHOD | speak    |
| Dog     | EXTENDS    | Animal   |
| Dog     | HAS_METHOD | bark     |
| dog.lua | IMPORTS    | base.lua |
```

Note the member-call edge is `run -[:CALLS]-> answer` (not `main.lua`), because the call site is inside `run()` — correct call-graph modeling.

**No shared-pipeline regression:** `npx vitest run test/unit/scope-resolution/` → 1496/1496. `detect_changes` on the indexed repo scopes the blast radius to the `EmitLuaHeritageEdges` Lua-scoped flows only; no shared-pipeline symbol was touched. Rebased onto the latest `main` (`a6a8aa7`) — your `c1103f3` and `cabd5b8` land below this and both the namespace-receiver and heritage paths still pass the above.

## Boundary / still out of scope

- **Indirect value receivers** (`local f = M.answer; f()`): needs the Case 5 value-receiver bridge — Phase B3 (marked in `interpret.ts`).
- **middleclass MRO + `__base` super-call dispatch**: `buildMro` empty, `isSuperReceiver` false — Phase B2 (marked in `scope-resolver.ts`).
- **Arity precision**: Lua varargs → `arityCompatibility` returns `'unknown'` by design.
- Other Lua OOP libraries (30log, etc.): middleclass only.

Thanks for pushing on the graph-level contract — both gaps were the same anti-pattern (registration/ABI tests passed while behavior was broken), and your fixture approach surfaced them. The initial Lua-support contract is now unambiguous at the edge level.

Daylighttx:
Hi @koriyoshi2041 — thanks again for the hands-on review. Indexing a real fixture and inspecting the graph is exactly the bar this PR should meet, and it surfaced two real gaps. Both are fixed on the rebased head (`c663dea`, base `origin/main` `a6a8aa7`).

## Gap 1 — `require()` cross-file member-call linking (the one you flagged)

`interpretLuaImport` returned `kind: 'wildcard'` for every `require()`. `collectNamespaceTargets` only registers receivers from `kind === 'namespace'` edges, so `local util = require("lib.util")` never bound `util` as a namespace receiver — `util.answer()` had no path to `answer`.

**Fix (`268fe72`).** `query.ts` adds a `local_variable_declaration` pattern capturing the LHS binding as `@import.localName` (bare `require()` pattern kept for side-effect imports); `interpret.ts` returns `{ kind: 'namespace', localName, importedName, targetRaw }` when `@import.localName` is present. The overstated comment is rewritten to describe both forms and the B3 boundary.

## Gap 2 — middleclass `EXTENDS` / `HAS_METHOD` never emitted (found while extending the test)

When I added a middleclass fixture to lock in the contract, `Dog -[:EXTENDS]-> Animal` and the `HAS_METHOD` edges were absent from the graph — even though `heritage.ts` existed and its comments claimed to emit them.

**Root cause.** `emitLuaHeritageEdges` re-read each file via `readFileSync(parsed.filePath)`, but `parsed.filePath` is **relative**. For `analyze <non-cwd-path>` (the normal case — including your `analyze /tmp/gitnexus-lua-review --skip-git` run), `readFileSync('base.lua')` resolved against the project root, ENOENT-failed on every file, and the hook silently `continue`d. EXTENDS/HAS_METHOD only appeared when `analyze .` was run from the repo root by accident. Graph-lookup calls were unaffected (filePath is just a key there); only the filesystem read broke.

**Fix (`c663dea`) — capture-side-channel pattern, mirroring Java/C/C++.** Heritage pairs are collected in `emitLuaScopeCaptures` (parse worker, where the tree-sitter AST is already live), snapshotted onto `ParsedFile.captureSideChannel` via `collectCaptureSideChannel`, and read by `emitLuaHeritageEdges` on the main thread — no file re-read, no re-parse (#1983).

- `lua/capture-side-channel.ts` (new): `LuaCaptureSideChannel` type + module map + `collectLuaCaptureSideChannel`.
- `lua/captures.ts`: runs `HERITAGE_QUERY` + `METHOD_OWNER_QUERY` against the same tree, stashes the pairs.
- `lua/heritage.ts`: reads `parsed.captureSideChannel`, emits EXTENDS + HAS_METHOD; `readFileSync`, the parser, and the local query copies are gone.
- `lua/query.ts`: `HERITAGE_QUERY` + `METHOD_OWNER_QUERY` + getters moved here.
- `lua.ts`: wires `collectCaptureSideChannel` through `assertCloneable`.

## Test cases added

`gitnexus/test/integration/resolvers/lua-scope.test.ts` (mirrors `ruby-scope.test.ts`). Two describe blocks with the fixtures inline:

```lua
-- require fixture (your exact two-file case)
-- lib/util.lua              -- main.lua
local M = {}                 local util = require("lib.util")
function M.answer()          local function run()
  return 42                    return util.answer()
end                          end
return M                     run()

-- middleclass fixture
-- base.lua                          -- dog.lua
local class = require("lib.class")   local class = require("lib.class")
local Animal = class("Animal")       local Animal = require("base")
function Animal:speak()              local Dog = class("Dog", Animal)
  return "..."                       function Dog:bark()
end                                    return "woof"
return Animal                        end
                                     return Dog
```

## Verification — actual commands and output I ran

These are the real commands and transcripts, not a summary. Reproduce from the repo root after `npm run build`.

**Unit/integration tests (the suite that covers both gaps):**
```
$ npx vitest run test/integration/resolvers/lua-scope.test.ts
 RUN  v4.1.10 /Users/tal/gitnexus-lua-pr
 Test Files  1 passed (1)
      Tests  6 passed (6)
```
The 6 named assertions: `emits IMPORTS edge from main.lua to lib/util.lua`, `resolves run -> util.answer() as CALLS edge to util.lua`, `detects answer as Method + run as Function`, `emits EXTENDS from Dog to Animal across files`, `emits HAS_METHOD from each class to its methods`, `detects Dog and Animal as Class nodes`.

**Type check:**
```
$ cd gitnexus && npx tsc --noEmit; echo "tsc exit: $?"
tsc exit: 0
```

**Graph edges — require fixture (your exact invocation, non-cwd):**
```
$ node gitnexus/dist/cli/index.js analyze /tmp/lua-rep --skip-git --index-only --workers 1
  6 nodes | 8 edges | 1 clusters | 0 flows

$ node gitnexus/dist/cli/index.js cypher "MATCH (a)-[r]->(b) WHERE r.type='CALLS' OR r.type='IMPORTS' RETURN a.name AS src, r.type AS edge, b.name AS dst" -r lua-rep
| run      | CALLS   | answer   |
| main.lua | IMPORTS | util.lua |
| main.lua | CALLS   | run      |
```
`run -[:CALLS]-> answer` is the edge you flagged as missing.

**Graph edges — middleclass fixture (the gap I found):**
```
$ node gitnexus/dist/cli/index.js analyze /tmp/lua-mc2 --skip-git --index-only --workers 1
  7 nodes | 10 edges | 1 clusters | 0 flows

$ node gitnexus/dist/cli/index.js cypher "MATCH (a)-[r]->(b) WHERE r.type='EXTENDS' OR r.type='HAS_METHOD' OR r.type='IMPORTS' RETURN a.name AS src, r.type AS edge, b.name AS dst" -r lua-mc2
| Animal  | HAS_METHOD | speak    |
| Dog     | EXTENDS    | Animal   |
| Dog     | HAS_METHOD | bark     |
| dog.lua | IMPORTS    | base.lua |
```

Note the member-call edge is `run -[:CALLS]-> answer` (not `main.lua`), because the call site is inside `run()` — correct call-graph modeling.

**No shared-pipeline regression:** `npx vitest run test/unit/scope-resolution/` → 1496/1496. `detect_changes` on the indexed repo scopes the blast radius to the `EmitLuaHeritageEdges` Lua-scoped flows only; no shared-pipeline symbol was touched. Rebased onto the latest `main` (`a6a8aa7`) — your `c1103f3` and `cabd5b8` land below this and both the namespace-receiver and heritage paths still pass the above.

## Boundary / still out of scope

- **Indirect value receivers** (`local f = M.answer; f()`): needs the Case 5 value-receiver bridge — Phase B3 (marked in `interpret.ts`).
- **middleclass MRO + `__base` super-call dispatch**: `buildMro` empty, `isSuperReceiver` false — Phase B2 (marked in `scope-resolver.ts`).
- **Arity precision**: Lua varargs → `arityCompatibility` returns `'unknown'` by design.
- Other Lua OOP libraries (30log, etc.): middleclass only.

Thanks for pushing on the graph-level contract — both gaps were the same anti-pattern (registration/ABI tests passed while behavior was broken), and your fixture approach surfaced them. The initial Lua-support contract is now unambiguous at the edge level.

koriyoshi2041 (approved):
Re-tested the rebased head (`c663dea`) from a fresh checkout. The focused Lua graph suite now passes 6/6, including the original `require()` member-call case and the non-CWD middleclass `EXTENDS`/`HAS_METHOD` case; `tsc --noEmit` also passes. The capture-side-channel change removes the relative-path reread failure cleanly. This resolves my graph-level concern.

azizur100389:
Reviewed current head `c663deac` against current main `f2717c6a`. It merges cleanly, typecheck passes, and 251 focused registration/ABI/parsing tests pass, but this is not merge-ready.

The new Lua resolver suite fails 2/6: neither the `IMPORTS` edge nor the cross-file `run -> answer` call is emitted. A direct parse of `local util = require("lib.util")` produced an error node and no import capture locally. My Node version was 22.17.1, just below the supported 22.18 floor, so please confirm on supported Node?but the advertised graph contract needs a passing current-main result.

Lua is also absent from `build-tree-sitter-prebuilds.yml`, despite the PR relying on that workflow for native delivery. Finally, `clearLuaHeritageFacts()` has no caller and empty recaptures do not remove prior facts, leaving stale heritage state possible during reanalysis.

Please fix these paths, add the lifecycle regression, and provide a green current-head CI Gate before re-review.

koriyoshi2041 (approved):
Re-tested the rebased head locally on macOS arm64 / Node 26.5.0 after rebuilding the newly vendored Lua grammar.

The graph-level regression fixture now covers the gap from my earlier review: `local util = require("lib.util")` produces the cross-file `run -[:CALLS]-> answer` edge, and the middleclass fixture produces `Dog -[:EXTENDS]-> Animal` plus both `HAS_METHOD` edges. The implementation note now matches the remaining B2/B3 boundaries.

Validation:

- `npx vitest run test/integration/resolvers/lua-scope.test.ts test/unit/tree-sitter-parser-loader-abi.test.ts test/unit/language-skip.test.ts test/unit/vendored-grammars.test.ts test/unit/ingestion-utils.test.ts` — 122/122 passed
- `npx tsc --noEmit` — passed
- `git diff --check` — passed

The first test run before `npm install` correctly failed closed because this checkout did not yet have a native Lua binding; the postinstall source build completed successfully, and the focused suite passed afterward. This addresses my earlier concern.

Daylighttx:
@azizur100389 Thanks for the thorough review. All three issues have been resolved in commit `95fdcae`, and the branch has been rebased onto the latest `origin/main`.

#### 1. On the 2/6 test failure
As you suspected, the root cause is the Node version falling below the project's minimum supported threshold.

The `engines` field specifies a minimum Node version of 22.18.0, while the test environment was running 22.17.1. Upgrading just one minor version allows the Lua grammar to parse normally.
Below are the re-test results using your exact test fixture on Node 22.18.0:
```bash
$ node --version
v22.18.0
$ node gitnexus/dist/cli/index.js analyze /tmp/lua-verify-rep --skip-git --index-only --workers 1
  3 nodes | 3 edges | 1 clusters | 0 flows
$ node gitnexus/dist/cli/index.js cypher "MATCH (a)-[r]->(b) WHERE r.type='CALLS' OR r.type='IMPORTS' RETURN a.name, r.type, b.name ORDER BY r.type" -r lua-verify-rep
| main.lua | IMPORTS | util.lua |
| main.lua | CALLS   | run      |
| run      | CALLS   | answer   |
```

@koriyoshi2041 already independently verified this on Node 26.5.0 with all 122 test cases passing and gave APPROVAL. I also re-ran the 5 targeted test suites you specified on 22.18.0, and all 143 cases passed. The graph contract holds as expected on both the minimum supported version and the latest current version.

#### 2. On missing Lua configuration in the prebuilds workflow
I've added the Lua entry to the prebuild registry:
`lua: { name: 'tree-sitter-lua', kind: 'vendored' }`

This configuration compiles from source under `gitnexus/vendor/tree-sitter-lua/` (which includes `parser.c` / `scanner.c` / `binding.gyp`), following the exact same pattern as Dart, Protobuf and Swift. I've also added a Lua code snippet for the validation step, and included the Lua shortname in the workflow dispatch inputs and grammar list header.

The `commit-fork-prebuilds.yml` is a generic script: it submits files via glob matching and automatically maps grammar-to-architecture based on artifact directory names. No additional changes are needed there — it will handle Lua prebuild artifacts automatically.

#### 3. On stale heritage facts cache
The root cause was in `captures.ts`: `setLuaHeritageFacts` only executed when heritage pairs were captured. If a file had its middleclass class definitions removed between parse passes, the old `EXTENDS` / `HAS_METHOD` facts would remain in cache, causing `collectLuaCaptureSideChannel` to return stale snapshot data.

**Fix applied:**
- Added handling for the empty recapture branch: when no heritage structures are detected on re-parse, `clearLuaHeritageFactsForFile` is called to clean up cache at the per-file level.
- Added a `loadResolutionConfig` hook in `scope-resolver.ts` that invokes `clearLuaHeritageFacts()` once per full workspace scan as a global safeguard, consistent with the mechanism used for Java and Kotlin.

A lifecycle regression test has been added, and it has been verified that the test fails outright if the per-file cleanup logic is removed:
```bash
$ npx vitest run test/integration/resolvers/lua-scope.test.ts
 Tests  8 passed (8)   # 3 require tests + 3 middleclass heritage tests + 2 lifecycle tests
```
Test logic: first parse a file containing middleclass classes to populate heritage data, then re-parse the same file with class definitions removed, and assert that the cache read returns `undefined`.

#### CI Status
GitHub Actions workflows and Vercel deployment remain in `action_required` status. Since this PR comes from a forked repository, workflows will not run automatically. A repository maintainer with write access must approve the workflow runs first, after which all tests will execute against commit `95fdcae`.

azizur100389:
## PR Review: #2819 — feat(languages): add Lua language support

**Recommendation:** Request changes
**Priority:** High
**Current head:** `95fdcae994741012d09db48f976fa46025e54b94`
**Review confidence:** High

> Current visible state is incomplete. I could verify the exact head, diff, commit history, mergeability, implementation and test structure, and hygiene, but not full current-head repository CI, native prebuild results, and packaged-install validation. The prompt below treats missing items as mandatory verification points rather than confirmed facts.

### 1. Review bar for this PR
Resolve imports and middleclass heritage without knowingly wrong edges, document shipped language/native behavior, and pass grammar-present ABI, packaged-install, resolver, and full current-head CI.

### 2. Problem being solved
Adds Lua parsing, require/member-call resolution, middleclass inheritance, vendored grammar, prebuild wiring, docs, and tests.

### 3. Current PR state
Open, non-draft, at the exact head above. Current base: `main` @ `aac7515d2a8c50a1f8f923c6fb77218b333560d6`; merge base: `021ac30376b263cf9140d8e8255b8b672f244d48`; current review decision: none visible; unresolved current review threads: 0.

### 4. Merge status and mergeability
**visibility incomplete.** GitHub reports MERGEABLE/BEHIND; the branch is 53 commits behind and has no current-head repository CI beyond Vercel status.

### 5. Repository history considered
Earlier graph/lifecycle/prebuild findings were followed by fixes; prior approvals cover earlier heads, and the final commit came later. This overlaps native grammar infrastructure with #1432/#3001.

### 6. Branch hygiene assessment
**clean feature/fix PR.** Seven linear commits and 43 files are broad but causally connected to Lua parser/resolver/native support.

### 7. Understanding of the change
Lua emits unified captures plus a side channel for middleclass EXTENDS/HAS_METHOD, with suffix-based require resolution and optional native loading.

### 8. Findings
#### [High] Duplicate class names can emit a knowingly wrong EXTENDS edge
- **Severity:** High
- **Location:** gitnexus/src/core/ingestion/languages/lua/heritage.ts:31-76
- **Current behavior / concrete trigger:** A global first-wins name map selects the parent; with two modules defining Animal, Dog can extend whichever Animal was indexed first, and the comment explicitly accepts that wrong edge.
- **Expected behavior:** Parent resolution must follow the importing file/module or refuse ambiguity.
- **Why tests or CI do not catch it:** Current Lua fixtures do not contain duplicate parent names across modules, so first-wins behavior remains green.
- **Minimal fix:** Resolve parent through the file’s require/import bindings or an ambiguity-safe per-module index; add a duplicate-name fixture asserting the correct or skipped edge.
- **Confidence:** High
- **Risk:** The knowledge graph emits incorrect inheritance and downstream dispatch/impact results.
- **Evidence to check:** gitnexus/src/core/ingestion/languages/lua/heritage.ts:31-76
- **Recommended fix:** Resolve parent through the file’s require/import bindings or an ambiguity-safe per-module index; add a duplicate-name fixture asserting the correct or skipped edge.
- **Blocks merge:** yes

#### [Low] Published npm README omits Lua support and native fallback guidance
- **Severity:** Low
- **Location:** gitnexus/README.md:343-365,523
- **Current behavior / concrete trigger:** Root docs advertise Lua, but the package README’s supported-language matrix and optional native grammar guidance do not.
- **Expected behavior:** Published package documentation should match shipped support and install fallback behavior.
- **Why tests or CI do not catch it:** Documentation drift is not covered by the language tests.
- **Minimal fix:** Add Lua to the package README matrix and native/prebuild fallback section.
- **Confidence:** High
- **Risk:** npm users receive incomplete support and installation guidance.
- **Evidence to check:** gitnexus/README.md:343-365,523
- **Recommended fix:** Add Lua to the package README matrix and native/prebuild fallback section.
- **Blocks merge:** no

### 9. PR-specific assessment
Real pipeline tests cover require linking and middleclass side-channel cleanup, but current-head native prebuild, packaged-install, grammar-present, and full CI evidence is absent.

### 10. Back-and-forth avoided by verifying
Pinned the head, inspected heritage/import/native/docs/tests, distinguished earlier-head approvals, checked divergence, and completed hygiene checks.

### 11. Open questions
None.

### 12. Final verdict
**not production-ready.** Lua integration is substantial and several earlier blockers appear fixed. The current heritage algorithm still chooses a potentially wrong duplicate parent instead of resolving in module scope or declining ambiguity. Published package docs also lag the feature, and current native/full CI is absent. Correct the inheritance rule, update docs, and run grammar-present current-head validation before merge.

koriyoshi2041 (changes requested):
Re-tested exact head `c592c587` after rebuilding both the Lua native binding and the repository. The focused Lua resolver suite passes 10/10, including the new duplicate-parent fixtures, but the vendored scanner still has an unsafe initialization path that should be fixed before shipping the grammar.

`tree_sitter_lua_external_scanner_create()` returns raw `malloc(sizeof(struct ScannerState))` storage. `deserialize()` does nothing when Tree-sitter calls it with an empty initial buffer, then `scan()` immediately switches on `state->started`; `serialize()` can also read both uninitialized fields. This is undefined behavior in the native parser boundary, even though the allocator happened to return benign bytes in my macOS run.

Please zero-initialize the state (`calloc` or explicit `started = 0; depth = 0`) and add a small repeated parse/load smoke so the vendored grammar does not depend on allocator contents. The current heritage change itself looks sound: after a clean build, the imported duplicate resolves to `lib/a.lua` and the unowned duplicate declines the edge.

Validation: Node 26.5.0, macOS arm64; clean Lua source rebuild; `npm run build`; `npx vitest run test/integration/resolvers/lua-scope.test.ts test/unit/parser-loader-abi.test.ts test/unit/language-skip.test.ts` — 38/38 passed.

koriyoshi2041 (changes requested):
The scanner initialization issue from my prior review is fixed at exact head `e9cba0b3`: `tree_sitter_lua_external_scanner_create()` now zero-initializes the state with `calloc`, and the repeated parser smoke passes locally.

One resolver-contract blocker remains. `lua/scope-resolver.ts` defines `mergeBindings: (existing) => [...existing]`, so every non-empty `incoming` batch passed by `finalizeScopeModel` is discarded. The surrounding comment says local-first-then-imports, but this implementation never merges the import-side batch. Please return both inputs (with the intended precedence) and add a regression that calls the merge path with non-empty `existing` and `incoming`; the current Lua graph fixture stays green without exercising that boundary.

Exact-head validation on macOS arm64 / Node 26.5.0: full build passed; the Lua resolver, parser ABI, optional-grammar, index-reuse, and external-import suites passed 90/90; `tsc --noEmit` and base diff check passed.

koriyoshi2041:
Rechecked the Lua path on exact head `a5611a3f86bb0b9e9e5a629ff616844ea4e793ab` after the latest resolver, parser-loader, and scanner fixes.

- Fresh install built the vendored Lua binding from source on macOS arm64 / Node 26.5.0.
- `lua-scope`, parser ABI, vendored-grammar monitor, and import-target-index reuse suites pass: 77/77.
- `npx tsc --noEmit` and `git diff --check` pass.
- The original `local util = require("lib.util")` → `util.answer()` graph gap from my Aug 4 review is now covered by the Lua resolver suite and passes locally.

I do not see a remaining blocker in the Lua import/ABI surface I revalidated. I did not run the full repository suite or independently validate all six cross-built prebuild artifacts.

koriyoshi2041 (approved):
Re-reviewed the single-commit delta a5611a3f..9d11a33f at exact head 9d11a33f. The complex Lua receiver fallback is narrowly scoped to call/variable nodes, the added class() and LF-continuation regressions cover the remaining gaps, and the Lua ABI declaration now matches the existing vendored grammar stubs. Local verification: 65/65 targeted tests passed (call-form + parser-loader-abi), npx tsc --noEmit passed from gitnexus/, and git diff --check is clean.

koriyoshi2041 (commented):
The approval body has been repaired in place; this note remains only as the correction audit trail.

azizur100389:
> Current visible state is incomplete. I could verify the exact-head native source build, typecheck, focused parser/resolver suites, and direct query probes, but not full exact-head repository Actions, the native prebuild matrix, and a packaged toolchain-free install check. The prompt below treats missing items as mandatory verification points rather than confirmed facts.

## 1. Review bar for this PR

This fresh full review checks exact head `9d11a33f8e7c4dc74b0741ac735d88bc1a7a916a` for complete Lua registration, native parser/scanner safety, scope-resolution correctness, cross-platform install and test behavior, documentation accuracy, and evidence from typecheck plus focused parser/resolver suites.

## 2. Problem being solved

The PR adds Lua as a supported language, including a vendored tree-sitter grammar, parsing and query registration, imports, call forms, middleclass heritage/method edges, resolver integration, optional-grammar handling, docs, and tests.

## 3. Current PR state

Open and non-draft. The branch was rewritten since the previously reviewed head `95fdcae994741012d09db48f976fa46025e54b94` and now contains 14 linear feature/fix commits, so a fresh full review was required.

**Merge state: `visibility incomplete`.** GitHub reports the head mergeable but behind `main`. Only label metadata and the unrelated Vercel authorization failure are visible; no full repository Actions run is attached to this exact head.

I rechecked the prior scanner initialization, empty-deserialize, depth serialization, CRLF continuation, member-call classification, optional-grammar smoke, import-aware heritage, stale side-channel, and middleclass query concerns. Those are fixed at the current head; the bare middleclass query is exercised successfully by the real Lua parser test. Gergo has not submitted a review on this PR.

**Branch hygiene: `clean feature/fix PR`.** The 14 linear commits remain focused on Lua support, its vendored grammar/build path, documentation, and directly related tests. The branch is behind current `main`, but it is not polluted by unrelated merges.

Lua is registered across shared language enums/detection, parser loading, worker and provider tables, import/call/type extraction, and scope-resolution dispatch. The PR adds namespace-style `require` resolution, dot/colon receiver handling, middleclass `EXTENDS`/`HAS_METHOD` capture side channels, import-aware parent disambiguation, scanner hardening, optional native build handling, and cross-language tests.

## 8. Findings

### [High] Lua installation depends on an undeclared network-fetched build tool

- **Risk:** Lua ships without committed native prebuilds, and the postinstall fallback invokes `npx node-gyp rebuild` even though `node-gyp` is not a declared dependency. A clean offline/toolchain-limited install cannot obtain the builder and silently leaves Lua unavailable, despite this PR advertising Lua support and a toolchain-free prebuild pipeline.
- **Evidence to check:** `gitnexus/scripts/build-tree-sitter-grammars.cjs:91-101`, `gitnexus/package.json:76-77`, `gitnexus/vendor/tree-sitter-lua/package.json:8`, and the absent `gitnexus/vendor/tree-sitter-lua/prebuilds/` directory. `npm ls node-gyp` is empty; the clean offline probe failed with `ENOTCACHED`.
- **Recommended fix:** Commit the supported platform prebuilds before merge, or declare and lock `node-gyp` and invoke its local binary. Add packed-tarball offline install coverage plus the native platform matrix so Lua availability is proven rather than best-effort.
- **Blocks merge:** yes

### [High] Multi-assignment cross-pairs Lua imports

- **Risk:** The query independently captures every LHS variable and every `require` expression in one declaration. `local a, b = require("x"), require("y")` therefore emits `a -> x`, `b -> x`, `a -> y`, and `b -> y`, creating false `IMPORTS` bindings and potentially false cross-file `CALLS` edges.
- **Evidence to check:** `gitnexus/src/core/ingestion/languages/lua/query.ts:68-79`. An exact-head parser probe produced all four `@import.localName`/`@import.source` combinations for the two-variable declaration.
- **Recommended fix:** Capture the declaration as a whole and zip variable/expression lists positionally, or initially restrict bound-require support to exactly one variable and one expression. Add a graph-level multi-assignment regression proving no cross-paired edges.
- **Blocks merge:** yes

### [Medium] Parenthesis-free `require` forms produce no imports

- **Risk:** Valid Lua idioms such as `local x = require "mod"` and `require [[mod]]` are not indexed as imports, so their `IMPORTS` and downstream call-resolution edges are absent.
- **Evidence to check:** `gitnexus/src/core/ingestion/languages/lua/query.ts:75-97` requires `argument_list (expression_list (string))`. Exact-head parser probes returned no import captures for either direct-string form.
- **Recommended fix:** Add direct-string `argument_list` alternatives for bound and bare `require`, including short and long strings, with graph-level tests.
- **Blocks merge:** yes

### [Medium] Aliased middleclass parents do not resolve to the returned class

- **Risk:** `local Base = require("base"); local Dog = class("Dog", Base)` misses `Dog EXTENDS Animal` when `base.lua` returns class `Animal`. The import records the module tail as `importedName`, while heritage resolution looks up the written alias as a class binding; the current test only passes because the local alias is also named `Animal`.
- **Evidence to check:** `gitnexus/src/core/ingestion/languages/lua/interpret.ts:35-44`, `gitnexus/src/core/ingestion/languages/lua/heritage.ts:58-67`, and the same-name fixture at `gitnexus/test/integration/resolvers/lua-scope.test.ts:229-249`.
- **Recommended fix:** Resolve an unambiguous class exported/returned through the namespace binding, and add an integration fixture whose local parent alias differs from the target class name.
- **Blocks merge:** yes

### [Medium] The new heritage regression fails on Windows path separators

- **Risk:** The exact-head focused suite fails on a supported platform, so the branch cannot satisfy the repository's cross-platform CI bar. The production pipeline returns canonical `lib/a.lua`, while the test expects Windows-native `lib\\a.lua` via `path.join`.
- **Evidence to check:** `gitnexus/test/integration/resolvers/lua-scope.test.ts:249`. On Windows, the exact-head source-built Lua binding loaded successfully, but this assertion failed with expected `lib\\a.lua`, received `lib/a.lua`; the other 324 focused tests passed.
- **Recommended fix:** Normalize both values to the repository's canonical path format, or assert the canonical forward-slash path explicitly. Rerun the focused suite and the repository's Windows platform shard.
- **Blocks merge:** yes

### [Low] The root feature matrix overstates Lua support

- **Risk:** The repository-root matrix marks Lua constructor inference, framework detection, and entry-point detection as supported, while the package matrix marks them unavailable and the Lua type extractor explicitly defers constructor inference. Users receive contradictory capability claims.
- **Evidence to check:** `README.md:583`, `gitnexus/README.md:365`, and `gitnexus/src/core/ingestion/type-extractors/lua.ts:1-8`.
- **Recommended fix:** Align both matrices with implemented behavior and qualify heritage as the currently supported middleclass pattern.
- **Blocks merge:** no

**Native/parser safety:** The worktree built `tree_sitter_lua_binding.node` from source successfully on Windows with network and a C++ toolchain available. Repeated parser creation, LF/CRLF escaped continuations, scanner reset/serialization, call forms, and bare middleclass binding passed; this does not cover the offline/toolchain-free install contract.

**Validation and hygiene:** `npx tsc --noEmit` passed. Focused validation across eight parser/resolver/ABI/integration files produced 324 passes and one path-separator failure. Direct exact-head query probes reproduced the multi-assignment cross-product and both missing parenthesis-free import forms. `git diff --check` is clean; no bidi controls were found; visible non-ASCII additions are normal documentation/checkmark and comment punctuation. The local runtime was Node 22.17.1, one patch below the declared 22.18 floor, but none of the reproduced parser/query/assertion failures is engine-related.

I built the vendored grammar from exact source, ran the real parser and integration suites, probed query captures for multi-assignment and direct-string `require`, checked middleclass resolution rather than inferring its shape, compared every current unresolved thread with current code, inspected the prebuild workflow/package metadata, and queried the current head/reviews/checks directly.

## 11. Open questions

None.

## 12. Final verdict

**Final verdict: `not production-ready`.** The rewritten branch fixes the previously reported scanner-state and several resolver issues, and a toolchain-backed source build plus typecheck are healthy. However, the current implementation can emit false import/call edges, misses valid Lua import syntax and aliased heritage, and has no proven toolchain-free installation path. The focused suite also fails on supported Windows, while full exact-head hosted CI is absent; these correctness and delivery gaps need fixes and regressions before merge.

koriyoshi2041 (commented):
Re-tested exact head `40c8e923` after the latest Lua positional-binding/export changes. The focused Lua resolver and optional-grammar coverage passes locally, including the new multi-assignment callable cases; the repository build and TypeScript check also pass.

Validation (macOS arm64 / Node 26.5.0):

- `npm run build` — passed
- `npx vitest run test/integration/resolvers/lua-scope.test.ts test/unit/parser-loader-skip-optional.test.ts test/unit/call-form.test.ts test/unit/export-detection.test.ts` — 97/97 passed (3 discovered files)
- `npx tsc --noEmit` — passed
- `git diff --check origin/main...HEAD` — passed

I did not re-approve because GitHub currently reports the branch as conflicting with `main`; the focused suite should be rerun after that rebase/merge resolution.

PR #2820, opened 2026-08-04
feat(serve): validate and port-scope the origin/proxy configuration surface

Ho1yShif:
## Summary

Validates the two `gitnexus serve` reverse-proxy knobs `GITNEXUS_PUBLIC_ORIGIN` and `GITNEXUS_TRUST_PROXY`, neither of which was checked before use, and makes browser-origin matching port-aware. This is the server-side half of #2804 split out on its own, with @magyargergo's review items 3 and 4 applied. No Render in this branch.

## Motivation / context

Both variables shipped in #2804 unvalidated and undocumented, which is how the review found them. Two concrete failure modes:

- Express 5 compiles `trust proxy` **inside** `app.set`, so a bad value fails at runtime rather than at configuration time — and a numeric hop count is never range-checked at all, so a long digit string overflows to `Infinity` and silently trusts every hop.
- Origin matching compared hostname only, so `GITNEXUS_PUBLIC_ORIGIN=https://home.example.com:8443` also admitted that host on `:9999` and on the default port.

**One correction to the review framing:** items 3 and 4 target code that #2804 *introduces*, so they cannot land as pre-existing fixes. On `main`, `middleware.ts` exports only:

```
$ git show main:gitnexus/src/server/middleware.ts | grep -E '^export (function|const|type)'
export function normalizeBoundHost(...)
export function createLocalhostOriginGuard(...)
export const requireLocalhostOrigin = createLocalhostOriginGuard()
```

`createPublicOriginMatcher`, `PUBLIC_ORIGIN_ENV`, `resolveTrustProxy`, and `TRUST_PROXY_ENV` do not exist there. So the split actually available is the whole server-side half, already corrected. #2804 rebases onto this and shrinks to `render.yaml`, its README section, `Dockerfile.cli`, and the proxy.

## Areas touched

- [x] `gitnexus/` (CLI / core / MCP server) — `src/server/{api,middleware}.ts` and six test files

Not docs-only, so the last box stays unchecked — but `README.md`, `CONTRIBUTING.md`, and `gitnexus/.env.example` do change. No agent-config or overlay file is touched.

## Scope & constraints

**In scope**

- `GITNEXUS_TRUST_PROXY` grammar validation with a safe fallback, never a wider one.
- `GITNEXUS_PUBLIC_ORIGIN` parsed to one reachable host, or rejected outright.
- Port-aware origin matching on both the configured public origin and the server's own bound host.
- Startup diagnostics consolidated into `logOriginPolicy`.
- `createLocalhostOriginGuard` → `createWriteOriginGuard`, since it now admits a non-localhost origin.
- README row + `.env.example` block for both variables.

- **Anything Render-related** — `render.yaml`, `Dockerfile.cli`, the deploy proxy, and their README section stay in #2804. Verified: `render.yaml|docker-server|SERVE_AUTH|AZURE_|SECURITY.md` greps to zero hits across the full diff.
- **`serve` authentication** (review item 1) and the item-2 wording change — held for #2804.
- **CLI port validation.** `serve --port 0` binds an ephemeral port `createServer` never learns, so the guard compares origins against `0` and refuses bound-host writes. That configuration is *already* unusable — the startup banner prints `http://localhost:0`, since nothing reads `server.address()` — and the real fix is validating `--port` (`serve.ts:29` is a bare `Number()`), which belongs with the banner in its own change. This is **distinct** from the `GITNEXUS_PUBLIC_ORIGIN` port-`0` rejection below: different input, different code path, and the fix lands in a file this PR does not touch. A non-numeric or out-of-range `--port` is unaffected either way — `app.listen` throws `ERR_SOCKET_BAD_PORT` before any request is served.
- **Playwright E2E.** Not run; see Testing.

### `GITNEXUS_TRUST_PROXY`

| Value | Before | Now |
| --- | --- | --- |
| `garbage` | `invalid IP address: garbage` on the first request | warns, falls back |
| `'9' × 400` | `Number()` → `Infinity`, and the hop test is `i < val`, so every hop is trusted | warns, falls back |
| `yes` / `no` / `on` / `off` | read as a proxy list, throws | accepted as booleans |
| `true` | accepted silently | accepted with a warning |

Hop counts are bounded to **0..16**, not the 1–16 I described in the review comment: Express tests a hop count as `i < hops`, so `0` already means trust nothing, exactly as `false` does. It is normalized to `false` rather than rejected, because rejecting it would fall back to a default that trusts loopback and the private ranges — *more* than the operator asked for.

A proxy list is validated the way Express validates it, mirroring `compileTrust`'s split-and-trim before `proxyaddr.compile`. `true` warns because Express then reads the client-controlled leftmost `X-Forwarded-For` entry, so a spoofed chain earns a fresh rate-limit key per request; `rate-limit.test.ts` proves that difference at the Express level rather than asserting it.

### `GITNEXUS_PUBLIC_ORIGIN`

Junk values came out *worse* than absent ones: `*`, `a.com,b.com`, and `8080` all built a matcher (`new URL` reads `8080` as the integer IP `0.0.31.144`) that no real Origin can satisfy — which still satisfied the constructibility guard, and so suppressed the wildcard-bind warning.

The matcher now takes the parsed request URL, compares hostname always and scheme/port only when the configured value carried them, and returns `undefined` for anything that is not one reachable host. A bare host with no port stays permissive on both, since platform service-discovery fields resolve to exactly that.

Port `0` is the same failure mode one level down, and it survived my first pass: `:0` and `:00` both round-trip through `new URL` intact, so they built a live-looking matcher that nothing could match while `logOriginPolicy` reported it at info as a working origin. Now rejected with the rest. `:0080` is deliberately still accepted — it normalizes to the elided default, so it means `:80`, and a positive test pins that so the rejection cannot quietly over-reach.

The two call sites read the env var with different lifetimes, worth naming before it gets asked: `isAllowedOrigin` is a module-level predicate with no construction step, so it rebuilds the matcher per call; the write guard has one, so it snapshots at construction and a later env mutation cannot widen a running server's write surface. Nothing mutates the env at runtime, so the only place the difference is observable is the tests.

### Pre-existing bug: the bound host ignored its port

`hostname === normalizedBoundHost` in the write-route guard ignored the port, so `serve --host 192.168.1.10 --port 8443` admitted writes from `192.168.1.10` on **any** port. The guard now takes the bound port. Loopback stays port-agnostic, since the dev UI runs on its own port.

This goes one step past the item as we scoped it — "the port comparison when the operator supplies a full origin, leaving the bare-host path alone." The bare-host path I meant is `GITNEXUS_PUBLIC_ORIGIN` set to a host with no port, which does stay permissive on scheme and port. The bound host is different: `--host` and `--port` are two arguments to the same command, so the server always knows both, and a guard that admits a LAN IP on every port is not doing what its own docstring claims. **Say the word and it comes out** — see Risk & rollout for what it costs.

### New dependency

`proxy-addr` at `^2.0.7` — the range Express declares and already resolves transitively — plus `@types/proxy-addr`. Hand-rolling the subnet grammar would drift from Express's. The lockfile adds only the `@types` tree entry, since `proxy-addr` itself was already resolved.

### Naming

`createLocalhostOriginGuard` now admits a public origin, which is by definition not localhost, so it is `createWriteOriginGuard` and its 403 body reads `restricted to trusted browser origins`. The unused `requireLocalhostOrigin` singleton is gone; it read `process.env` at import time and only tests used it. `gitnexus-web` dispatches on `code: 'origin_not_allowed'`, not on the message text, so the reworded body breaks nothing there.

Startup diagnostics moved into `logOriginPolicy`, which always warns on a wildcard bind and logs the resolved public-origin hostname at info.

### Docs

Both variables get a row in README's **Environment variables** table and a commented block in `gitnexus/.env.example`, under a heading saying they are only for a proxied or wildcard bind.

One process change beyond the items, flagged because it binds future PRs and not just this one: `CONTRIBUTING.md`'s checklist gains a line requiring every new `GITNEXUS_*` variable to have a README row. **Drop it if you'd rather not carry it.**

## Testing & verification

- [x] `cd gitnexus && npm test` — see the pre-existing-failure note below
- [ ] `cd gitnexus && npm run test:integration` — n/a, no core/indexing/MCP path changed (only `src/server/`); the integration suites still ran as part of `npm test`
- [ ] `cd gitnexus-web && npm test` — n/a, web unchanged
- [ ] `cd gitnexus-web && npx tsc -b --noEmit` — n/a, web unchanged
- [ ] **Manual / Playwright E2E — not run.** Calling this out rather than omitting it: this PR changes which browser origins may write, so an end-to-end check behind a real proxy is the one gap in this verification. Everything below is unit/integration level.

The six server suites this PR touches all pass — `cors`, `rate-limit`, `middleware-trust-proxy`, `middleware-public-origin`, `api-analyze-upload`, `server-cors-stack` — at `6 passed (6)` / `179 passed (179)`.

Full `npm test`: `11 failed | 792 passed | 11 skipped (814)` files, `45 failed | 16181 passed | 114 skipped (16340)` tests. The 11 failing files are Go/C++/TypeScript/Swift resolvers, Spring AOP (2 files), MCP startup closure, analyzer identity, evidence provenance, review-agent workflow, and incremental orchestration. They reproduce identically with this branch's changes reverted, and all 11 grep to **zero** references to `server/middleware.ts` or `server/api.ts`. At least the MCP closure probe needs a `dist/` build this checkout does not have. Flagging rather than hiding it — worth a look at whether CI is green on `main`.

New: `middleware-trust-proxy.test.ts` (accept/reject grammar tables, the `true` warning) and `middleware-public-origin.test.ts` (port-aware write guard, `logOriginPolicy`'s four-way matrix). `cors.test.ts` gains the read-side matrix, including bracketed IPv6 with a port, and `:0`/`:00` in its not-one-reachable-host table; the `compares on hostname, ignoring the port` case that asserted the bug is gone.

On the `rate-limit.test.ts:273-274` item: the effective-value block is there, keying off a sentinel env value and reading `app.get('trust proxy')` back. The source-text assertion stayed rather than being replaced, because `createServer` listens and installs signal handlers, so it cannot be built in a unit test and nothing else proves it reads the env var at all. It is loosened to the `resolveTrustProxy(… TRUST_PROXY_ENV` call shape, so it no longer pins the exact argument text.

One piece of test hygiene the diff shows and this explains: `cors.test.ts`, `api-analyze-upload.test.ts`, and `middleware-public-origin.test.ts` now clear an ambient `GITNEXUS_PUBLIC_ORIGIN` for the file and restore it afterwards. Their rejection cases assume none is configured, and once the guard reads that variable a developer who exports it locally fails three of them for reasons unrelated to their change. Verified rather than assumed: with `GITNEXUS_PUBLIC_ORIGIN=192.168.1.10` exported, the pre-fix tree fails 3 and the fixed tree passes all of them. Pinning it in `vitest.config.ts` next to `GITNEXUS_MEMORY` was the alternative; per-file hooks keep the assumption next to the tests that rely on it.

## Risk & rollout

**Breaking change — this rejects a topology that works today.** `serve --host <LAN-IP> --port 8443` behind a TLS proxy on 443 sends `Origin: https://<LAN-IP>`, which is port 443 and no longer matches the bound port. Those setups must set `GITNEXUS_PUBLIC_ORIGIN` to the origin the browser actually uses. **Worth a line in the release notes.** This is the sole reason the bound-host port scoping is offered as droppable above; everything else in this PR only narrows invalid configuration or widens nothing.

No migration, no index refresh, no schema change. Defaults are unchanged: with both variables unset, `trust proxy` resolves to the same `loopback, linklocal, uniquelocal` literal `main` hardcodes, and a local `gitnexus serve` behaves exactly as it does today. An invalid value for either variable warns and falls back — it never widens the allowlist or the trusted-hop set.

Security note: `GITNEXUS_PUBLIC_ORIGIN` widens the CORS **read** allowlist as well as the write guard. `cors()` is configured without `credentials: true`, so a widened read allowlist cannot expose a session; the exposure is graph data to a page on an origin the operator explicitly configured.

## Checklist

- [x] PR body meets repo minimum length (`MIN_BODY_CHARS = 200` in `.github/scripts/review-precheck.cjs`)
- [x] If `AGENTS.md` / overlays changed: headers, scope block, and changelog updated per project conventions — n/a, no overlay or agent-config file touched

## Open questions for the reviewer

1. **Bound-host port scoping** — ships, or comes out? It is the one breaking change here.
2. **The `CONTRIBUTING.md` checklist line** — carry it, or drop it? It binds future PRs, not just this one.


Ho1yShif:
TODO: Test a Render deploy to ensure everything works as expected. I won't include the `render.yaml` in this PR, will just use it for my own testing

gitnexus/src/server/api.ts
```diff
@@ -753,21 +746,10 @@ export const createServer = async (port: number, host: string = '127.0.0.1') =>
...
-    logger.warn(
-      { host },
-      `[gitnexus serve] Bound to a wildcard address (${host}); browser write routes ` +
-        `accept only loopback origins (localhost/127.0.0.1/[::1]). To allow writes from a ` +
-        `specific LAN address, bind --host <that-address> instead of a wildcard.`,
-    );
-  }
+  // Origin guard for write routes: loopback, the server's own bound host, and
+  // any configured public origin — prevents CSRF from other devices.
+  const requireTrustedOrigin = createWriteOriginGuard(host, port);
```
magyargergo, line 762: Coming back to the auth thing from my summary.

`middleware.ts:171` passes any request with no `Origin`, so curl skips the guard entirely and gets `DELETE /api/repo` and `POST /api/analyze` with any absolute path. That's been true forever and it's been fine, because we only ever bound loopback. This is the PR that changes that.

Would you be up for refusing to start when `GITNEXUS_PUBLIC_ORIGIN` is set and no auth is configured? Just until #2804 lands. Or flip the order and rebase this on top of it, either's fine by me. I'd just rather not ship the half that opens the door without the half that locks it.

gitnexus/src/server/middleware.ts
```diff
@@ -40,56 +51,251 @@ export function normalizeBoundHost(boundHost?: string): string | undefined {
...
+  const raw = (scheme ? trimmed.slice(scheme.length + 3) : trimmed).replace(/\/$/, '');
+  const authority = splitAuthority(raw);
+  if (!authority) return undefined;
+
+  const hostname = normalizeBoundHost(authority.host);
+  if (!hostname) return undefined;
+
+  // A bare host stays deliberately permissive on scheme and port: platform
+  // service-discovery fields resolve to a hostname with neither.
+  const expectedProtocol = scheme ? `${scheme}:` : undefined;
```
magyargergo: If someone sets `GITNEXUS_PUBLIC_ORIGIN=app.example.com`, `expectedProtocol` stays undefined and `http://app.example.com` gets write access plus CORS reads. On an https deployment that's a downgrade path.

The bare-host form is for Render/Fly service discovery iirc, and those are always https, so I think we can just default it?

```suggestion
  // A bare host stays permissive on the port, since platform service-discovery
  // fields resolve to a hostname with neither, but defaults to https.
  const expectedProtocol = `${scheme ?? 'https'}:`;
```

gitnexus/src/server/middleware.ts
```diff
@@ -40,56 +51,251 @@ export function normalizeBoundHost(boundHost?: string): string | undefined {
...
+ * Anything else warns and returns {@link DEFAULT_TRUST_PROXY}. Express compiles
+ * this value inside `app.set`, so an unvalidated bad one takes `serve` down at
+ * startup; a number it accepts without any range check at all.
  */
-export const requireLocalhostOrigin = createLocalhostOriginGuard();
+export function resolveTrustProxy(raw?: string): string | number | boolean {
+  const value = raw?.trim();
+  if (!value) return DEFAULT_TRUST_PROXY;
+
+  if (/^(true|yes|on)$/i.test(value)) {
```
magyargergo, line 328: fwiw express-rate-limit already flags this one, `validations.trustProxy` throws `ERR_ERL_PERMISSIVE_TRUST_PROXY` when trust proxy is true.

But I'd go further than warning. The limiter is the only thing sitting in front of `/api/analyze` and `/api/embed` and both spawn workers, so `true` is a fairly direct path to burning the box. Is there a case where someone genuinely needs it and a hop count won't do? If not I'd just reject it.

magyargergo (commented):
Looks good overall, the port-0 and `Infinity` catches are nice finds.

Main thing before this goes in: I don't think we have any auth on `serve` right now? Grepped `SERVE_AUTH|Authorization|bearer|apiKey` across `src/server/` and only got git-clone's outbound PAT. And the write guard passes anything without an `Origin` header, so it only ever stops browsers. Which means this is the PR that makes a public bind work, and there's nothing actually guarding it yet. Longer comment on `api.ts:751`.

Couple of smaller things on `middleware.ts`, plus some nits.

Re your open questions: ship the port scoping, and keep the CONTRIBUTING line.

gitnexus/src/server/middleware.ts
```diff
@@ -40,56 +51,251 @@ export function normalizeBoundHost(boundHost?: string): string | undefined {
...
+    // trusts the whole chain rather than failing loudly.
+    if (Number.isInteger(hops) && hops <= MAX_TRUST_PROXY_HOPS) return hops;
+    return rejectTrustProxy(value, `expected a hop count of 0..${MAX_TRUST_PROXY_HOPS}`);
+  }
+
+  try {
+    // Mirror Express 5's compileTrust (`express/lib/utils.js`): it splits on `,`
+    // and trims before handing the list to proxy-addr, which rejects unknown
+    // subnet names itself. proxy-addr does not split, so pass the array form.
+    proxyaddr.compile(value.split(',').map((entry) => entry.trim()));
```
magyargergo, line 354: Optional, take it or leave it. This is basically Express's `compileTrust`, and it's the only thing pulling `proxy-addr` into package.json. `app.set('trust proxy', ...)` runs the same code, so wrapping that in a try/catch gets the same validation for free.

Downside is it stops being a pure function you can unit test, which I'm guessing is why you did it this way.

gitnexus/src/server/api.ts
```diff
@@ -710,26 +720,9 @@ export const createServer = async (port: number, host: string = '127.0.0.1') =>
...
-  // edge, CGNAT 100.64/10) is NOT in the trusted set. In those topologies
-  // req.ip will collapse to the LB hop IP for every request and the per-IP
-  // rate limiter degrades to per-server. Add an explicit env-var override
-  // and document the cloud-deploy story before binding to a non-loopback
-  // host in those topologies (tracked as a follow-up; not blocking for the
-  // local-bound default).
-  app.set('trust proxy', 'loopback, linklocal, uniquelocal');
+  // Which upstream hops may set X-Forwarded-*. Process-wide: every route's
+  // req.ip, and so the per-IP rate limiter, resolves through this.
+  app.set('trust proxy', resolveTrustProxy(process.env[TRUST_PROXY_ENV]));
```
magyargergo, line 732: Behind a cloud LB the default doesn't match anything, so `req.ip` is the LB on every request and the per-IP limit quietly becomes one global limit.

`resolveTrustProxy` has no idea what we're bound to, but this line does. Warn if the bind is non-loopback and the var isn't set?

README.md
```diff
@@ -508,6 +508,8 @@ Most `analyze` knobs are also CLI flags (`--workers`, `--worker-timeout`, `--max
...
+| `GITNEXUS_PUBLIC_ORIGIN`                        | unset                     | The single browser origin `serve` is reached through, added to the CORS allowlist and to the write-route origin guard. A wildcard bind (`0.0.0.0`) has no host identity, so without this the server's own UI is refused. A scheme, if given, is enforced; so is a port. A bare host with no port stays permissive on both. A value that is not one reachable host (a list, `*`, a bare port number) warns at startup and allows nothing — it never widens the allowlist. | `gitnexus serve` runs behind a reverse proxy or on a wildcard bind, and the UI's index/delete requests return `origin_not_allowed`.                                                    |
```
magyargergo: Could we spell out what "permissive on both" means? With no scheme and no port it's any scheme, any port, that hostname.

README.md
```diff
@@ -508,6 +508,8 @@ Most `analyze` knobs are also CLI flags (`--workers`, `--worker-timeout`, `--max
...
+| `GITNEXUS_TRUST_PROXY`                          | `loopback, linklocal, uniquelocal` | Express `trust proxy` value — which upstream hops may set `X-Forwarded-*`, and so what the per-IP rate limiter reads as the client IP. Accepts a hop count of `0`..`16` (`0` trusts no hop, the same as `false`), a boolean (`true`/`false`, `yes`/`no`, `on`/`off`), or a proxy list Express can compile. Prefer a hop count: `true` reads the client-controlled leftmost `X-Forwarded-For` entry, so a spoofed chain earns a fresh rate-limit key per request — it is accepted, with a warning. An invalid value warns and falls back to the default rather than widening anything. | `serve` sits behind a load balancer outside the private ranges (AWS ALB, Cloudflare, CGNAT), where every request otherwise collapses to the proxy hop and rate limiting goes global.  |
```
magyargergo: Needs a tweak if `true` becomes a rejection.

Also the `0..16` range reads like a security bound and it isn't. Someone putting 3 in front of one proxy is handing the caller two entries of the chain. Maybe "the exact number of proxies you control"?

gitnexus/src/server/middleware.ts
```diff
@@ -40,56 +51,251 @@ export function normalizeBoundHost(boundHost?: string): string | undefined {
...
+    return /^[0-9A-Fa-f:.]+$/.test(authority) ? { host: `[${authority}]` } : undefined;
+  }
+
+  const parts = /^([^:]+)(?::(\d{1,5}))?$/.exec(authority);
+  if (!parts) return undefined;
+  const host = parts[1];
+  // `new URL` is far laxer than DNS: it takes `a.com,b.com` verbatim and reads
+  // `8080` as the integer IP 0.0.31.144, either of which yields a matcher that
+  // can never match a real Origin while still reading as configured.
+  if (/[\s,;*/?#@\\]/.test(host) || /^\d+$/.test(host)) return undefined;
```
magyargergo: nit, same family as the port-0 thing you already caught: trailing dots survive. `new URL('http://example.com.').hostname` comes back as `"example.com."`, so `GITNEXUS_PUBLIC_ORIGIN=example.com.` builds a matcher nothing can match and `logOriginPolicy` logs it at info as working. Follow-up is fine.

gitnexus/src/server/api.ts
Ho1yShif: Done as a hard refusal. `assertServeAuthForPublicOrigin()` in `middleware.ts` throws when `GITNEXUS_PUBLIC_ORIGIN` is set and `isServeAuthConfigured()` returns false, and `createServer` calls it as its first statement, before the port is bound, before LadybugDB opens, before MCP mounts. `serve` catches it, prints `serve.startFailed`, and exits 1.

There's no override flag, deliberately. An escape hatch is the thing an operator sets once and forgets. `isServeAuthConfigured()` is a placeholder returning `false` unconditionally, pinned by a test, so the auth PR flips one function body and the gate plus its tests hold unchanged. `test/integration/server-http-startup.test.ts` spawns the built CLI and asserts the exit code, the message, and that nothing ever listened.

One note on the framing: #2804 doesn't add auth yet but we'll make sure the final updated PR does.

gitnexus/src/server/middleware.ts
Ho1yShif: Thanks, took the suggestion. `expectedProtocol` is now `` `${scheme ?? 'https'}:` ``, so `app.example.com` admits only `https://app.example.com` and plain HTTP needs the explicit `http://` form.

Two knock-ons. The matcher's `expectedProtocol === undefined ||` branch is dead now that the value is always a string, so it's gone. The port round-trip reuses the resolved scheme instead of hardcoding `http`, which keeps the two normalizations consistent.

Tests: the read side (`cors.test.ts`) and the write side (`middleware-public-origin.test.ts`) each gained a positive https case, a rejected http case, and an `http://`-spelled-out case. `README.md` and `.env.example` both updated, since "permissive on both" stopped being true.

gitnexus/src/server/middleware.ts
Ho1yShif: Rejected rather than warned. I couldn't construct a case a hop count or a proxy list doesn't cover, since a real chain has a knowable length. And `ERR_ERL_PERMISSIVE_TRUST_PROXY` settles it. `true` was never a working configuration, so accepting it with a warning meant handing over a value express-rate-limit then refuses.

`true`/`yes`/`on` now route through `rejectTrustProxy`, which warns and returns `DEFAULT_TRUST_PROXY` (`loopback, linklocal, uniquelocal`). That's strictly narrower than `true`, so the rejection can only ever trust less than was asked for. Opposite of the hop-count-`0` case, where I normalize to `false` precisely because falling back would trust *more*.

Two tests pin the property rather than the branch: `never returns true, so express-rate-limit cannot reject the value we set` in `middleware-trust-proxy.test.ts`, and `never hands Express true` in `rate-limit.test.ts`, which reads the value back off `app.get('trust proxy')`. The Express-level demonstration of *why* stays (rotating `X-Forwarded-For`, fresh limiter key per request); it sets `true` directly now, since the env var can't produce it.

gitnexus/src/server/middleware.ts
Ho1yShif: Left it, and you were right. Three reasons to keep it:

1. `createServer` listens and installs signal handlers, so it can't be constructed in a unit test. `resolveTrustProxy` staying pure allows the ~30-case accept/reject grammar tables. Moving validation into `app.set` trades those for nothing observable.
2. Validating first means a bad value never reaches Express, so config errors are decided in one place.
3. Catching a throw from `app.set` leaves `trust proxy` in whatever state the throw left it, so the fallback needs a second `app.set` to repair. Subtler to get right than calling `proxyaddr.compile` on a string.

The dependency cost is also smaller than it looks. `proxy-addr@^2.0.7` is the exact range Express declares and already resolves transitively, so the lockfile only gained the `@types/proxy-addr` entry, no new resolved package. If you'd still rather not carry the direct dep, say so and I'll swap it.

gitnexus/src/server/api.ts
Ho1yShif: Done, as `warnIfRateLimitKeysCollapse(host)`, called right after `app.set('trust proxy', …)` where the bound host is known. It warns when `GITNEXUS_TRUST_PROXY` is unset and the bind isn't loopback, wildcard binds included since those accept LB traffic too, and it names the mechanism: the balancer sits outside the default's ranges, so `req.ip` is the balancer and the per-IP limit becomes one shared limit.

Ten cases in `middleware-trust-proxy.test.ts` cover the loopback/wildcard/LAN/public matrix, the set-but-invalid silence, and a whitespace-only value counting as unset.

README.md
Ho1yShif: Reworded, and item 2 changed the meaning anyway. The row now states each rule separately: hostname must match exactly; scheme must match, with a schemeless value read as `https` and the reason given (bare hosts come from platform service discovery, which terminates TLS); an explicit port must match, and with no port any port on that hostname is accepted.

It also leads with the item-1 gate now, since "this setting refuses to start" is the first thing a reader needs. The not-one-reachable-host list gained `:0` and the trailing dot from item 8. `gitnexus/.env.example` carries the same wording.

README.md
Ho1yShif: Changed to your phrasing. The row now leads with "set it to the exact number of proxies you control. Every hop past that is one more entry of the chain the caller gets to write", so the range is no longer the headline. `16` is described as "a sanity ceiling rather than a safety boundary", and `MAX_TRUST_PROXY_HOPS`'s docstring says the same thing so the next reader of the code gets it too.

`true`/`yes`/`on` is documented as rejected, with the `ERR_ERL_PERMISSIVE_TRUST_PROXY` reason. The row also documents the new item-5 warning, and `.env.example` matches.

gitnexus/src/server/middleware.ts
Ho1yShif: Fixed here instead of deferring, since it's one condition in `splitAuthority` sitting next to the checks it belongs with. `host.endsWith('.')` now returns `undefined`, so `example.com.` is refused for the same reason as `:0`: a browser sends `example.com`, so the matcher is dead-but-configured and `logOriginPolicy` would have reported it at info as working.

`cors.test.ts` adds `a.com.` and `https://a.com.:8443` to the not-one-reachable-host table, `middleware-public-origin.test.ts` asserts the write guard admits neither `https://app.example.com` nor `https://app.example.com.`, and a `logOriginPolicy` case pins that it warns instead of reporting an origin.

Ho1yShif:
All eight comments addressed in c2940d44. Seven implemented, one declined (the optional `proxy-addr` one), with reasoning on that thread.

The two that changed behavior:

- `GITNEXUS_PUBLIC_ORIGIN` now refuses to start. You were right that the no-auth gap is this PR's problem and not a pre-existing one, because this is the PR that makes a non-loopback bind usable. `serve` throws before it binds a port, opens LadybugDB, or mounts MCP, and exits 1. No override flag. `isServeAuthConfigured()` is a pinned placeholder returning `false`, so the auth work flips one function body.
- `GITNEXUS_TRUST_PROXY=true` is rejected, not warned. `ERR_ERL_PERMISSIVE_TRUST_PROXY` settled it: it was never a working value, so accepting it with a warning meant offering something express-rate-limit then refuses. The fallback is strictly narrower than `true`, so rejecting can only trust less than was asked for.

Also in: a schemeless value means `https` rather than either scheme, a trailing dot is refused like `:0`, and a non-loopback bind with `GITNEXUS_TRUST_PROXY` unset now warns that the per-IP limit has collapsed to one shared limit. README and `.env.example` match, including `16` reading as a sanity ceiling rather than a safety boundary.

Two pre-existing things I noticed and deliberately left alone, since neither is in this diff: `isAllowedOrigin` still matches loopback by string prefix rather than by parsed hostname, and `https://gitnexus.vercel.app` is hardcoded into the CORS read allowlist with no env override. Happy to file issues for these if helpful.

PR #1966, opened 2026-06-01
feat(go): infer structural interface implementations

evander-wang:
## Summary

This PR adds signature-aware structural interface implementation detection for Go and feeds the inferred relationships into the scope-resolution pipeline.

Go does not have explicit `implements` declarations, so the analyzer has to infer interface satisfaction from method sets. Before this change, the Go analyzer could only reason about direct receiver calls and did not have a reliable way to emit `IMPLEMENTS` edges for structurally satisfied interfaces. That meant interface-typed calls such as `repo.Save(user)` could not fan out through the graph to known concrete implementations.

The new flow is:

1. Extract richer Go declaration metadata from tree-sitter captures.
2. Populate Go method ownership across package files.
3. Infer structural `IMPLEMENTS` edges from Go interface method sets.
4. Emit those edges before MRO/interface dispatch is built.
5. Let existing `interface-dispatch` and `METHOD_IMPLEMENTS` logic consume the inferred relationships.

### Structural interface implementation detection

- Reworked `detectGoInterfaceImplementations` from method-name-only matching into method-set + signature matching.
- Collects interface and struct methods by owner.
- Requires interface methods to have verifiable signature metadata before emitting edges.
- Compares parameter counts, required parameter counts, parameter types, and return types.
- Preserves Go type shape during signature checks:
  - `User` is not equal to `[]User`
  - `[]User` is not equal to `...User`
  - `a.User` is not equal to `b.User`
- Emits inferred edges with reason `go-structural-implements`.

### Embedded interface handling

- Captures embedded interface references from Go interface declarations.
- Recursively merges locally resolvable embedded interface method sets.
- Allows embedded empty interfaces to contribute no required methods.
- Suppresses structural detection when an embedded interface cannot be resolved locally, avoiding false positives for cases such as `io.Reader`.

Example:

```go
type Reader interface {
    Read() error
}

type ReadCloser interface {
    Reader
    Close() error
}
```

`ReadCloser` now requires both `Read()` and `Close()`. A type with only `Close()` no longer gets a false `IMPLEMENTS` edge.

### Pointer receiver method-set correctness

Go value and pointer method sets differ:

```go
type I interface { M() }
type T struct{}
func (*T) M() {}

var _ I = &T{} // OK
var _ I = T{}  // compile error
```

This PR preserves the raw receiver shape for Go self bindings and records whether a method was declared on `T` or `*T`.

- `func (t T) M()` contributes to the value type method set.
[...]

PR #2829, opened 2026-08-04
fix(go): model Go method sets exactly so interface satisfaction is decidable (#2813)

### LINKED ISSUE #2813: Go: calls through interface-typed struct fields resolve inconsistently at scale; impact() still reports 0 with epistemic "exact"
Follow-up to #2766, which you fixed and closed on 2026-08-01. **The concrete-typed half is genuinely fixed — thank you.** This is the interface-typed remainder, re-measured on `1.6.10-rc.152`.

## What the #2766 fix did resolve

Verified on the same private ~470k LOC Go codebase (2,787 files; 40,649 nodes / 166,019 edges; 44.5s). Edge count is up ~5% from 1.6.9's 158,131.

Cross-package **concrete** struct fields now emit `CALLS`. `h.auditRepo.LogAuditEventAsync(...)` in `internal/handlers/orders.go` produced **zero** edges on 1.6.9 and produces **14** on rc.152. The minimal fixture from #2766 also now passes in full — interface field, concrete field, and concrete `WithTx` all emit, where all three failed before.

## What still fails

Methods reached only through an **interface-typed** struct field still have no incoming `CALLS`. Six real symbols, ground truth established with `gopls references` / `gopls call_hierarchy` (gopls v0.23.0 — the Go team's own type checker, used here as the oracle) and cross-checked against grep:

| symbol | real callers | `gitnexus impact` | |
|---|---|---|---|
| `OrderRepo.DeleteItem` | 2 | `0` / `exact` / `LOW` | ✗ |
| `OrderRepo.GetPickQueue` | 2 | `0` / `exact` / `LOW` | ✗ |
| `PickService.GetPickQueue` | 2 | `0` / `exact` / `LOW` | ✗ |
| `PickService.StartSession` | 1 | `0` / `exact` / `LOW` | ✗ |
| `AuditRepo.LogAuditEventAsync` | 13 | `16` / `exact` / `CRITICAL` | ✓ |
| `OrderRepo.UnsplitOrder` | 0 | `0` / `exact` / `LOW` | ✓ |

Every miss goes through an interface field; the one hit is a concrete field. The receiving declarations:

```go
// internal/handlers/orders.go:66
type OrderHandlers struct {
    repo      repository.OrderRepository   // interface  -> MISSING
    auditRepo *repository.AuditRepo        // concrete    -> resolves
}

// internal/handlers/picking.go:67
type PickHandlers struct {
    svc PickServiceInterface               // interface  -> MISSING
}
```

`gopls` resolves all six exactly, including through those interface fields, so the information is recoverable from the source.

## The part that makes it disqualifying rather than merely incomplete

Look at rows 1 and 6. `UnsplitOrder` genuinely has no callers. `DeleteItem` has two. **Both print `impactedCount: 0, epistemic: "exact", risk: "LOW"`** — byte-identical. There is no field in the response that separates a true zero from an unresolved one.

That means a zero can't be used defensively, which is the entire use case: checking blast radius before a push that auto-deploys to production. The #2747 `lower-bound` hedge does not fire on any of these — I confirmed `epistemic` is `"exact"` on all four wrong rows.

I'd rather have `"epistemic": "lower-bound"` with an unresolved-receiver boundary here than a confident zero, even if the edges themselves stay unresolved. Those are separable fixes and the epistemics one is independently valuable.

## Lead: it's inconsistent, not uniformly broken

This is the part I can't explain and where you'll have better instincts. Files with the **same declared field shape** behave completely differently. Counting `CALLS` edges landing on `internal/repository/interfaces.go` (the interface method declarations) by source file:

| source file | edges | field-receiver call sites (grep) |
|---|---|---|
| `internal/services/wave_service.go` | **42** | 57 |
| `internal/services/shopify_sync.go` | 5 | 108 |
| `internal/handlers/orders.go` | 4 | 50 |
| `internal/services/pick_service.go` | **0** | 71 |

All four declare the same kinds of field — e.g. both `wave_service.go` and `pick_service.go` hold `orderRepo repository.OrderRepository` and both use pointer receivers (`func (s *WaveService)` / `func (s *PickService)`). Yet one resolves most of its interface-field calls and the other resolves none of them. Whatever the remaining variable is, it is not the declared shape at the call site.

Note also that when these edges *do* resolve, they land on the **interface declaration** node in `interfaces.go`, not on the implementation. So `impact` against the impl method misses them regardless. Targeting the interface node instead (`impact DeleteItem --file internal/repository/interfaces.go`) also returns `0` — there's no workaround from the query side.

## Hypotheses I tested and killed

So you don't spend time on these:

- **Multiple implementors causing ambiguity — NO.** A fixture with one interface and three implementors (`Impl`, `Mock`, `Recording`, mimicking real mock proliferation) resolves cleanly, one `CALLS` edge.
- **Interface vs concrete field type — NOT SUFFICIENT.** It predicts the symbol table above, but not the file table: `pick_service.go` also holds concrete fields (`cartRepo *repository.CartRepo`, `auditRepo *repository.AuditRepo`) and still emits 0 edges into `interfaces.go`.

## Repro

**The minimal fixtures no longer reproduce it** — as in #2766, small Go fixtures are an unreliable proxy here, and this time they pass where the real codebase fails. Everything above is measured on the real corpus. If it's useful I can run any targeted query or Cypher probe you want against that index and report back; I can't share the source.

## Environment

- `gitnexus` 1.6.10-rc.152 (npm `latest` is still 1.6.9, so this is `@rc` only)
- `gopls` v0.23.0 as ground-truth oracle · Go 1.26
- Node v22.22.3 · Darwin arm64 (Apple Silicon)
- `GITNEXUS_SKIP_OPTIONAL_GRAMMARS=1` at install · indexed with `--index-only`

magyargergo:
Fixes #2813.

## The bug

A call through an interface-typed struct field never reached the implementation. The `CALLS` edge stopped at the interface **declaration**, so `impact()` on the implementing method reported `0` callers — byte-identical to a symbol that genuinely has none:

```go
type PickService struct {
    orderRepo repository.OrderRepository   // interface  -> implementation never reached
    auditRepo *repository.AuditRepo        // concrete    -> resolves fine
}

func (s *PickService) StartSession(id string) error {
    return s.orderRepo.DeleteItem(id)      // edge landed on interfaces.go, not order_repo.go
}
```

That is what made the reporter's blast-radius check unusable rather than merely incomplete: a zero could not be trusted, which is the entire use case.

## Two stacked defects reproduce it

The issue says minimal fixtures no longer reproduce it. They do — an 8-file fixture reproduces it exactly. No existing fixture caught it because **two** independent defects have to be fixed and the corpus was blind to both.

**D1 — pointer-receiver methods were excluded from Go structural interface satisfaction.** A struct declaring `func (r *OrderRepo) DeleteItem(...)` had an empty method set and produced **no `IMPLEMENTS` edge at all**. Every implementor in both pre-existing structural-dispatch fixtures uses a VALUE receiver, and the one pointer-receiver type was pinned as a negative, so the corpus could not observe this.

**D2 — Case 0 (compound receiver) never emitted the interface-dispatch fan-out that Case 4 performs.** The gap was a property of receiver *syntax*: Case 0 is selected by `receiverName.includes('.')`, so a field receiver always lands there, while the same interface reached through a local or parameter is a bare name and falls through to Case 4, which fans out correctly. Field-held interfaces — dependency injection — were the half that silently lost every implementation edge.

## Then: making the model exact, not just fixed

Review of the first cut showed the fix was still an *approximation*, and that approximation produced answers that are simply wrong against the Go spec. So the structural model was replaced with the spec's actual rules.

**Method sets** ([spec](https://go.dev/ref/spec#Method_sets)): `MS(T)` is the methods declared with receiver `T`; `MS(*T)` is those declared with `*T` **or** `T`.

**Promotion** ([spec](https://go.dev/ref/spec#Struct_types)): `S` embedding `T` puts `T`-receiver methods in `MS(S)` and `MS(*S)`, but `*T`-receiver ones only in `MS(*S)`. `S` embedding `*T` puts **both** in `MS(S)` and `MS(*S)`.

**Identifier identity** ([spec](https://go.dev/ref/spec#Uniqueness_of_identifiers)): *"Two identifiers are different if they are spelled differently, **or if they appear in different packages and are not exported**."*

Measured before and after:

| case | before | now | Go |
| --- | --- | --- | --- |
| `func (b *Base) Ping()` | `IMPLEMENTS` | `-implements-pointer` | only `*Base` implements |
| `struct{ Base }` (value embed) | `IMPLEMENTS` | `-implements-pointer` | only `*ByValue` implements |
| `struct{ *Base }` (pointer embed) | `IMPLEMENTS` | `-implements` | the value type implements |
| `foreign.T` vs `sealed.Sealed` | `IMPLEMENTS` | **no edge** | cannot implement |
| `sealed.Inside` (same package) | `IMPLEMENTS` | `-implements-pointer` | only `*Inside` implements |

The first three were one indistinguishable edge; two of the three were wrong. The fourth was a **false edge**, not an over-approximation — `seal` is unexported, so `sealed.seal` and `foreign.seal` are different identifiers and no type outside `sealed` can ever implement it. That is the whole basis of the sealed-interface idiom, and the interface-dispatch fan-out was turning that false edge into an impossible `CALLS` edge. The fifth row is the check that the fix does not over-correct.

### How the form is represented

Detection returns `{ structDefId, receiverForm }`. `receiverForm: 'pointer'` means the value type does **not** implement and only `*T` does — the fact `var x I = T{}` turns on.

It rides in the edge `reason` (`go-structural-implements-pointer`) because relationships carry no arbitrary properties; a new field would change the relation DDL, move `SCHEMA_FINGERPRINT` and force a rebuild for a fact a string already expresses. Value-form implementors keep the **original unsuffixed** reason, so a consumer matching the old string now sees exactly the assignable set — which is what that string always claimed to mean.

### The embed form had to be captured

`struct{ T }` and `struct{ *T }` promote different method sets, and the `*` was discarded at parse time (it is an unnamed token). Go embeds now emit `@reference.embedded-pointer`. That is parse-time capture emission, so **`SCHEMA_BUMP` 42 -> 43** — a warm cache would otherwise replay the pre-marker captures and the distinction would never appear, silently. The pin in `incremental-parse-cache.test.ts` moves with it.

## Review findings also fixed

- **Fan-out reached a declaration and missed the implementation.** Reproduced in two Java shapes — a sub-interface that re-declares the inherited method, and an abstract base between interface and implementation. `interface B extends A` is stored as `B IMPLEMENTS A`, an `Interface` was accepted as an implementor, and `pickOverload` never walked the MRO, so the edge landed on a bodiless declaration while the only class with a body got nothing. `emitInterfaceDispatchFor` now walks the subtype closure (IMPLEMENTS + EXTENDS) and skips declarations: the wrong edge disappears and the real implementation gains one.
- **The fan-out was an uncapped `implementors x call-sites` cross product** (measured at exactly N x M; 500 x 100 = 50,000 edges from one interface). Now bounded by `MAX_INTERFACE_DISPATCH_FANOUT` (32, `GITNEXUS_MAX_INTERFACE_DISPATCH_FANOUT`) **and reported** when it truncates, mirroring `MAX_PROPERTY_DISPATCH_FANOUT`. A bare cap would silently discard valid dispatch targets — the same false-safe silence this issue is about.
- A rationale comment that was factually wrong about the code 70 lines above it, and both copies of the case-ladder contract that still described the fan-out as Case-4-exclusive.
- Test gaps in this PR's own suite: the concrete-field control was vacuous with respect to the gate it appeared to pin (repointed at a struct that *is* an implementor), the two-service-file row could not distinguish the two files it is named for (both ends now file-qualified), plus new rows for signature mismatch, emitted confidence, an exact N x M fan-out bound, and `goReceiverKind` stamping — which had no reader left and could have rotted silently.

## The epistemics half

The issue asks separately for `epistemic: "lower-bound"` instead of a confident zero. It is **downstream of D1** and needed no new code: `computeEpistemicBoundary` walks `IMPLEMENTS`/`METHOD_IMPLEMENTS` out of the queried symbol, and D1 meant there were none.

| query | before | after |
| --- | --- | --- |
| `impact(OrderRepo.DeleteItem, upstream)` | `0` / `exact` | **`3`** / **`lower-bound`** + boundary note |
| `impact(CartRepo.Get, upstream)` — concrete | `1` / `exact` | `1` / `exact` |

The second row is what matters for trust: the hedge discriminates instead of firing on everything, so `exact` still means exact.

## Testing

- Cross-language sweep (`test/integration/resolvers` + Go unit): **3234 passed**, 1 skipped, 61 files, zero regressions — verified after rebasing onto main with #2810 merged in.
- `scope-capture --check`: PASS (15 languages). Go is the **only** fingerprint that moved, which is the check that this is a Go capture change and not a cross-language regression.
- `tsc --noEmit` clean in both packages.

An existing index must be re-analyzed; the schema bump forces it.

## Known follow-ups, not fixed here — all tracked

Each is filed rather than left in prose, so none is closed silently with #2813:

- **#2834** — no non-Go fixture exercises a compound (field) interface receiver, so the shared Case 0 path is ungated outside Go. The delta is not zero: a TypeScript field-receiver probe goes 1 -> 3 edges. This is also what would have caught the sub-interface defect found in review.
- **#2832** — Case 3b has the same missing fan-out Case 0 just had; undocumented and untested.
- **#2833** — a generic-typed field (`repo: Repo<User>`) yields zero `CALLS` edges. Pre-existing: Case 4 strips template arguments, Case 0's fold does not.
- **#2836** — the canonical Go decorator (`type loggingRepo struct { next Repository }`) now emits a self-loop dispatch edge. Not provably wrong (nested middleware is real), but it should be a decision with a fixture rather than an accident.
- **#2835** — the Go detection scaling benchmark is blind to receiver kind, already red on `main`, and excluded from CI, so detection cost is effectively ungated.
- **#2837** — the per-file inconsistency from #2813 (`wave_service.go` 42 edges vs `pick_service.go` 0) does not reproduce at fixture scale and needs the reporter's corpus. Split out so closing #2813 does not bury it.


gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
```diff
@@ -644,6 +644,36 @@ export function emitReceiverBoundCalls(
...
+            // `currentClass` is the receiver's own folded type, matching what
+            // Case 4 passes. `emitInterfaceDispatchFor` self-gates on
+            // `ownerDef.type !== 'Interface'`, so this is inert for every
+            // concrete receiver and needs no language check of its own — a
+            // member whose owner resolved to a Struct emits nothing extra.
+            //
+            // Confidence is Case 0's own 0.85 literal from the primary emit
+            // above, NOT Case 4's site.kind-dependent value: Case 0 has no
+            // read/write branch, so there is no 1.0 arm to mirror.
+            emitted += emitInterfaceDispatchFor(
```
abhigyanpatwari, line 803: **P1** [reproduced]

This newly reaches a depth-1 implementor index that accepts Interfaces as implementors (`:343`, because `pipeline/run.ts:198-199` stores `interface B extends A` as `B IMPLEMENTS A`) and resolves via `pickOverload(implDef.nodeId, …)` with no MRO walk (`:378`). Reproduced in Java: with a sub-interface that re-declares the inherited method, `this.r.read(p)` gains an edge to the second *abstract declaration* while the only class with a body gets nothing; same shape lands on an `abstract` method when an abstract class sits between interface and impl. Suggest skipping implementors whose `type === 'Interface'`, and expanding one hop through the implementor's own entry when `pickOverload` misses.

gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
```diff
@@ -644,6 +644,36 @@ export function emitReceiverBoundCalls(
...
+              currentClass,
+              memberName,
+              memberDef,
+              site,
+              0.85,
+              calleeCapture,
+            );
```
abhigyanpatwari, line 810: **P1** [reproduced]

Uncapped fan-out: measured exactly `implementors × call sites` edges end-to-end (500 × 100 = 50,000; 40 × 20 = 820 CALLS with zero duplicates), and Go doesn't set `collapseMemberCallsByCallerTarget` so they're per call site. D1 is the escalation — before it the implementor list was empty for pointer-receiver types, so this loop was a no-op. Mirror `MAX_PROPERTY_DISPATCH_FANOUT` (`property-dispatch.ts:60-63`), which caps *and* reports `skippedKeys`/`skippedKeyNames` — a bare cap would silently drop valid dispatch targets. At minimum, file the tracking issue and reference it here as `// TODO(#N)`.

gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
```diff
@@ -644,6 +644,36 @@ export function emitReceiverBoundCalls(
...
+            //
+            // `currentClass` is the receiver's own folded type, matching what
+            // Case 4 passes. `emitInterfaceDispatchFor` self-gates on
+            // `ownerDef.type !== 'Interface'`, so this is inert for every
+            // concrete receiver and needs no language check of its own — a
+            // member whose owner resolved to a Struct emits nothing extra.
+            //
+            // Confidence is Case 0's own 0.85 literal from the primary emit
+            // above, NOT Case 4's site.kind-dependent value: Case 0 has no
+            // read/write branch, so there is no 1.0 arm to mirror.
```
abhigyanpatwari: **P2** [code-read]

This is factually wrong: Case 0 *does* branch on `site.kind`, at `:596-598` (`site.kind === 'call' ? pickFirstNonStaticOnly(...) : findOwnedMember(...)`). What it lacks is a read/write branch in its reason/confidence computation — Case 4 has one at `:1399-1405`. The 0.85 conclusion is defensible; the stated reason isn't. Related: this call has no `site.kind` guard, so a TS/C# interface *property* accessed via `this.cfg.debug` now fans out at 0.85 while the same access through a bare-name receiver takes Case 4 and emits at 1.0.

gitnexus/src/core/ingestion/languages/go/interface-impls.ts
```diff
@@ -71,7 +70,35 @@ function buildDetectionIndexes(
...
+      // GitNexus models one Struct node per type with no separate `*T` node, so
+      // the two method sets cannot both be represented. This picks the `*T`
+      // reading: the graph now answers "which types provide this interface's
+      // behaviour", and no longer proves `var x I = T{}` invalid. That trade is
+      // deliberate — blast radius is what every consumer of IMPLEMENTS asks for
+      // (verified: MRO/METHOD_IMPLEMENTS derivation, community clustering, the
+      // receiver-dispatch fan-out index, and the epistemic heritage probe; none
+      // performs value-assignability checking).
+      //
+      // `goReceiverKind` is still stamped in method-owners.ts and is the hook a
```
abhigyanpatwari, line 117: **P2** [code-read]

After this PR `goReceiverKind` has zero readers: a repo-wide grep finds only the write at `method-owners.ts:98`, this comment, and the test setup at `go-hooks.test.ts:347`. Delete the write and nothing fails, so the hook rots silently — and the flipped unit test named "treats pointer-receiver-only methods as implementations" would now pass identically for a *value*-receiver method. Three lines in the Go hooks suite asserting `populateGoOwners` stamps `'pointer'` for `func (r *T)` and `'value'` for `func (r T)` makes this comment's promise true.

gitnexus/src/core/ingestion/languages/go/interface-impls.ts
```diff
@@ -71,7 +70,35 @@ function buildDetectionIndexes(
       }
       if (def.type !== 'Method' && def.type !== 'Function') continue;
       if (def.ownerId === undefined) continue;
-      if (isPointerReceiverMethod(def)) continue;
+      // POINTER-receiver methods count toward the method set (#2813).
```
abhigyanpatwari, line 91: **P2** [code-read]

Worth naming in the follow-up: `methodSetSatisfies` (`:462`) keys method sets by `simpleQualifiedName` (`graph-bridge/ids.ts:346`), which strips the package qualifier. Go treats unexported method identifiers as package-distinct, so `package sealed; type I interface { seal() }` cannot be implemented by a foreign `func (t *T) seal()` — but both reduce to the key `seal`. Pre-existing (a value-receiver `seal()` already matched), amplified here because pointer-receiver methods now participate. A two-package regression asserting no IMPLEMENTS and no interface-dispatch edge would pin it.

gitnexus/test/integration/resolvers/go.test.ts
```diff
@@ -1794,3 +1809,132 @@ describe('Go pointer-receiver field chains (#2766)', () => {
...
+    expect(methodImpls).toContain('order_repo.go:DeleteItem → interfaces.go:DeleteItem');
+    expect(methodImpls).toContain('mock_repo.go:DeleteItem → interfaces.go:DeleteItem');
+  });
+
+  // Concrete-field control: resolved before #2813 and must be untouched.
+  it('keeps resolving a concrete-typed field to its implementation', () => {
+    expect(callsToFile()).toContain('GetPickQueue → LogAuditEventAsync@audit_repo.go');
+  });
+
+  it('does not fan out a concrete-typed field receiver', () => {
```
abhigyanpatwari: **P2** [code-read]

This control can't fail for the reason it names. `implementorsByInterfaceDefId` is keyed by the *target* of an IMPLEMENTS edge (`receiver-bound-calls.ts:348-359`), so only interfaces are ever keys; `AuditRepo` is a Struct nothing implements, so the map lookup at `:374-375` misses regardless of the `ownerDef.type !== 'Interface'` gate at `:373`. Delete that gate and this still passes. It remains a valid regression test for the user-visible property — it just isn't the gate's control. Point it at the concrete `*OrderRepo` field instead (`OrderRepo` *is* an implementor), which is the only shape where the type gate is load-bearing.

gitnexus/test/integration/resolvers/go.test.ts
```diff
@@ -1794,3 +1809,132 @@ describe('Go pointer-receiver field chains (#2766)', () => {
     expect(imports.map((e) => `${e.source} → ${e.target}`)).toContain('handler.go → repo.go');
   });
 });
+
+// ---------------------------------------------------------------------------
```
abhigyanpatwari, line 1831: **P2** [reproduced]

This suite is Go-only, but the Case 0 change is in shared code consumed by all 14 languages and the delta is not zero: a two-implementor non-generic TypeScript fixture calling through a field and a parameter side by side gives `runField → save` 1 → 3 edges (total CALLS 4 → 6) while `runLocal → save` stays at 3. Those new edges are the real implementations — correct behaviour — but nothing outside Go pins it, so a regression here is uncatchable, and it's what would have caught the sub-interface wrong edge. One field-receiver fixture per collapse class (Java/C# collapse; TS/Kotlin/Go don't) closes it.

abhigyanpatwari (commented):
## Review — #2829

This fixes a real bug and the diagnosis behind it is right. Two stacked defects were needed to produce the reported symptom, and both are correctly identified: `buildDetectionIndexes` excluded pointer-receiver methods from Go structural interface satisfaction (so `func (r *OrderRepo) DeleteItem` produced **no** `IMPLEMENTS` edge at all), and Case 0 — the case a dotted field receiver always lands in — emitted its primary edge and short-circuited without the interface-dispatch fan-out Case 4 already performed. Either one alone reproduces the symptom, which is exactly why the existing value-receiver-only corpus was blind to it. The `*T` method-set reading is the correct choice given one `Struct` node per type, and I could not construct a Go program where it admits an implementor Go rejects for **both** `T` and `*T` — value-embed, pointer-embed, depth-N promotion, ambiguous multi-embed and non-addressable receivers all behave. The three re-polarised assertions are legitimate: each traces directly to the removed filter and each would fail on the pre-fix code. And the "no schema bump" call is correct, verified five separate ways. The 8-file fixture is well built and the failing-first accounting (7 fail / 4 control rows pass) is the right way to present it.

What follows is what I want addressed or explicitly accepted before this lands. The two P1s are both about the **fan-out target set**, not about the Go method-set reading.

---

### P1 — the fan-out can emit an edge to a *second interface declaration* and still never reach the implementation (Java)

`gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts:669` · **[reproduced]**

Trigger — plain Java, no generics, no annotations, where the sub-interface **re-declares** the inherited method:

```java
public interface Reader     { int read(String p); }
public interface ReadCloser extends Reader { int read(String p); void close(); }   // re-declares
public class FileRC implements ReadCloser { public int read(String p){return 1;} public void close(){} }
public class Svc {
    private Reader r;                                    // interface-typed FIELD → Case 0
    public int runField(String p) { return this.r.read(p); }
}
```

| | `runField → read` |
|---|---|
| base | `read@Reader.java` `[import-resolved]` — 1 edge |
| head | `read@Reader.java` **+ `read@ReadCloser.java` `[interface-dispatch]`** — 2 edges |

`FileRC.read` — the only method in the program with a body — gets **no edge in either build**. The single edge this PR adds points at a second abstract declaration.

Mechanism, all three steps confirmed at source:

- `pipeline/run.ts:198-199` classifies heritage by the **target's** kind, so `interface ReadCloser extends Reader` is stored as `ReadCloser IMPLEMENTS Reader`. That is true in every language, not just Java.
- `receiver-bound-calls.ts:343` filters implementor defs to `Class | Struct | Interface` — an **Interface is accepted as an implementor**.
- `emitInterfaceDispatchFor` gates only the *owner* (`:373`) and resolves with `pickOverload(implDef.nodeId, …)` (`:378`) — the implementor's own owner id, never its MRO. `FileRC`, two hops away, is structurally unreachable.

Narrowing, stated honestly: **the wrong edge requires the sub-interface to re-declare the method.** A first probe without re-declaration produced no fan-out at all and did *not* reproduce it. Re-declaring an inherited method in a sub-interface is legal and not rare (docs, covariant returns), so the trigger is real but narrower than "any Java sub-interface".

The same depth-1 blindness has a second shape [reproduced], which is the more common one:

```java
public interface Handler { void handle(String s); }
public abstract class AbstractHandler implements Handler { public abstract void handle(String s); }
public class RealHandler extends AbstractHandler { public void handle(String s){ … } }
public class Svc { private Handler h;  public void runField(){ this.h.handle("x"); } }
```

head adds `handle@AbstractHandler.java [interface-dispatch]` — a bodiless abstract declaration. `RealHandler.handle`, the only executable target, gets nothing; it reaches `Handler` via `EXTENDS`→`IMPLEMENTS`, which the depth-1 implementor index does not traverse. `impact(RealHandler.handle)` still returns zero.

None of this is new logic — Case 4 has had it — but Case 4 only ever fired on bare-name receivers. This PR makes it reachable from **fields**, which is the Spring/DI shape, i.e. the dominant Java shape. That is the escalation.

Cheapest fix is two lines inside `emitInterfaceDispatchFor`: skip implementors whose `type === 'Interface'` (a sub-interface is not an implementation), and when `pickOverload` on the implementor's own id misses, expand once through that implementor's own entry in `implementorsByInterfaceDefId` before giving up. Neither touches Case 0. If you'd rather not widen the traversal in this PR, the sub-interface skip alone removes the wrong edge and is worth taking now.

---

### P1 — the fan-out is an uncapped `implementors × call-sites` cross product, and D1 is what makes the loop actually iterate

`gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts:676` · **[reproduced]**

The PR body already names this under "Known follow-ups", so this is not a surprise — but it is worth putting numbers on it, because the escalation is bigger than "a second call site".

Measured end-to-end through the real pipeline, one interface `Closer { Close() error }`, N pointer-receiver implementors, held as a struct field, called from M sites:

| N | M | `interface-dispatch` edges |
|---|---|---|
| 10 | 20 | 200 |
| 100 | 20 | 2,000 |
| 500 | 20 | 10,000 |
| 500 | 100 | **50,000** |

Exactly `N × M` in every row. An independent run at 40 implementors × 20 sites produced exactly 820 CALLS (800 fan-out + 20 primaries) — no duplicates, no suppression, a literal cross product. Go does not set `collapseMemberCallsByCallerTarget`, so these are **per call site**, not collapsed per `(caller, target)`. TypeScript and Kotlin are the same; Java and C# do collapse.

The two-fold escalation: (a) Case 0 is a second uncapped call site, and (b) before D1, `implementorsByInterfaceDefId` was effectively **empty** for pointer-receiver types, so the loop was a no-op — the fan-out was broken rather than bounded. This PR makes it functional at the same moment it moves onto the dominant receiver shape.

Two honest counterweights, both of which I think are right:

- The current index tops out at ~5 implementors per interface, so nothing here is on fire today. This is a scale concern for large DI-heavy repos, not a defect.
- A naive cap would **silently discard valid dispatch targets**. That is precisely why the repo's own precedent doesn't just cap: `MAX_PROPERTY_DISPATCH_FANOUT` (`property-dispatch.ts:60-63`) caps at 32, is env-overridable, *and* records `skippedKeys` + a bounded `skippedKeyNames` sample so an operator can see which key lost its fan-out. Copy that shape, not just the number — and apply it to Case 4 as well, which shares the exposure.

Secondary effect worth stating: every one of these edges is persisted, so this is graph-content growth, not transient compute. And `impact()` truncates `byDepth` at its limit *after* traversing the inflated set, so a hub interface both costs more to traverse and silently under-reports its own blast radius.

Minimum ask: file the tracking issue and reference it from the PR body and from a `// TODO(#N)` at the call site. It is currently disclosed but untracked.

---

### P2 — the new rationale comment is factually wrong about the code 70 lines above it

`receiver-bound-calls.ts:667-668` · **[code-read]**

> `// above, NOT Case 4's site.kind-dependent value: Case 0 has no`
> `// read/write branch, so there is no 1.0 arm to mirror.`

Case 0 **does** branch on `site.kind`, at `:596-598`:

```ts
site.kind === 'call'
  ? pickFirstNonStaticOnly(ownerId, memberName, site, model, provider)
  : findOwnedMember(ownerId, memberName, model);
```

What Case 0 lacks is a read/write branch in its *reason/confidence* computation — Case 4 has one at `:1399-1405` (reason = `site.kind`, confidence `1.0` for read/write). The `0.85` conclusion is defensible; the stated reason is not, and it is the one sentence in the block asking a reviewer to trust the reasoning rather than check it.

The behaviour underneath is real and unexercised: the new call sits inside `if (memberDef !== undefined)` with no `site.kind` guard, so it fires for `read`/`write` sites too. Go interfaces have no fields, so **every** test in this PR exercises only the CALLS path — but a TS/C# interface property accessed through `this.cfg.debug` now fans out to every implementor's field at confidence `0.85`, while the identical access through a bare-name receiver takes Case 4 and emits at `1.0`. Same logical access, different confidence depending on receiver syntax — which is the class of inconsistency this PR set out to remove. Either answer is fine; pick one and say so.

### P2 — the concrete-field control cannot fail for the reason it names

`gitnexus/test/integration/resolvers/go.test.ts:1924` · **[code-read]**

```ts
it('does not fan out a concrete-typed field receiver', () => {
  expect(callsWithReason()).not.toContain('GetPickQueue → LogAuditEventAsync:interface-dispatch');
});
```

`implementorsByInterfaceDefId` is keyed by the **target** of an `IMPLEMENTS` edge (`:348-359`), so only interfaces are ever keys. `AuditRepo` is a `Struct` that nothing implements, so it is never a key. The fan-out is therefore doubly gated for this input: the `ownerDef.type !== 'Interface'` check at `:373` **and** the map miss at `:374-375`. Delete the type gate — the exact mutation this row appears to pin — and it still passes.

Both things are true and the PR body should say both: the row *is* a valid regression test for the user-visible property (no dispatch edge on a concrete field), and it is vacuous as a control for the gate it looks like it pins. The PR body lists it among "the 4 that passed are exactly the controls that must not regress", which overstates it. To make the gate load-bearing, point the control at a struct that *does* carry `IMPLEMENTS` edges — e.g. assert `Release → DeleteItem` through the concrete `*OrderRepo` field emits no `:interface-dispatch` row even though `OrderRepo` is an implementor.

### P2 — `goReceiverKind` now has zero readers, and nothing pins that it is still stamped

`gitnexus/src/core/ingestion/languages/go/interface-impls.ts:99` · **[code-read]**

The comment promises an escape hatch: *"`goReceiverKind` is still stamped in method-owners.ts and is the hook a future value/pointer-aware model would read."* After this PR a repo-wide grep finds exactly three hits: the write at `method-owners.ts:98`, this comment, and the test setup at `go-hooks.test.ts:347`. No functional reader remains.

Two consequences. Delete `method-owners.ts:98` and no test fails, so the hook can rot silently. And the flipped unit test — now named "treats pointer-receiver-only methods as implementations" — would pass **identically for a value-receiver method**, because nothing reads the stamp anymore; it no longer discriminates pointer from value, which is the property its name claims. Three lines in the Go hooks suite asserting `populateGoOwners` stamps `'pointer'` for `func (r *T)` and `'value'` for `func (r T)` makes the comment's promise true and restores the discrimination.

### P2 — the widening also widens simple-name matching for unexported methods

`gitnexus/src/core/ingestion/languages/go/interface-impls.ts:73` · **[code-read]**

`methodSetSatisfies` (`:462`) compares method sets keyed by `simpleQualifiedName` (`graph-bridge/ids.ts:346`), which strips everything before the last dot — no package qualification. Go treats *unexported* method identifiers as package-distinct, so a sealed interface

```go
package sealed;  type I interface { seal() }
package foreign; func (t *T) seal() {}
```

cannot be implemented from outside its package, yet both reduce to the key `seal`. D1 widens the population that can collide on such a name, because pointer-receiver methods now participate. This is **pre-existing** (a value-receiver `func (t T) seal()` would already have matched) and only amplified here, so I am not asking for it in this PR — but it belongs in the follow-up issue alongside the cap, with a two-package regression asserting no `IMPLEMENTS` and no `interface-dispatch` edge.

### P2 — the case-order contract still describes fan-out as Case-4-exclusive

`receiver-bound-calls.ts:17-18` and `contract/scope-resolver.ts:116` (neither in the diff) · **[code-read]**

Both say Case 0.5 *"does NOT emit Case 4's interface-dispatch fan-out"*. That sentence is still literally true of Case 0.5, but it frames the fan-out as belonging to Case 4, and Case 0 now has it too. The file header states the numbering *"is part of the contract — keep the comments"*, and `scope-resolver.ts` carries it as invariant I4. One clause on the Case 0 bullet in both files ("…also emits the interface-dispatch fan-out when the folded receiver type is an Interface (#2813)") closes it. This is the artifact whose whole job is to track this kind of per-case nuance.

---

### Lower priority

- **P3 — decorator/middleware structs now emit self-loop `CALLS` edges** [reproduced]. Canonical Go shape: `type loggingRepo struct { next Repository }` with `func (l *loggingRepo) DeleteItem(id string) error { return l.next.DeleteItem(id) }`. D1 makes `loggingRepo` an implementor of `Repository`; Case 0 folds `l.next` to `Repository`; the fan-out excludes only the *primary target* (`:386`), never the caller — so it emits `loggingRepo.DeleteItem → loggingRepo.DeleteItem`. This needs **both** changes to fire; the shape was previously immune precisely because it holds its delegate in a field. It is not a provably wrong edge (nested middleware is real, so `next` genuinely can hold another `loggingRepo`), but downstream consumers — cycle detection, `impact()`, community clustering, `process-processor.ts:376`'s "avoid cycles" truncation — will not expect a self-recursive `CALLS` edge there. Make it a decision with a fixture, not an accident.
- **P3 — interface methods promoted from an embedded interface still resolve to nothing** [reproduced]. `type ReadCloser interface { Reader; Close() error }` with `s.rc.Close()` and `s.rc.Read(p)` side by side: `Close` gets both the primary and the fan-out (the fix working); `Read` gets **zero edges — no primary, no fan-out — in both builds and through both Case 0 and Case 4.** Not a regression, but `io.ReadCloser` is the canonical Go shape and the new suite header asserts the general invariant "calls through an interface-typed struct field must reach the implementation". Name the remaining half in the PR body and carry an `it.fails` row so the gap is recorded rather than silently absent.
- **P3 — generic interface fields lose the entire call**, pre-existing [reproduced]. `private repo: Repo<User>` yields **zero** `CALLS` edges in both builds — not even the primary. Case 4 strips template args via `resolveClassBindingForName` (`:135-167`); Case 0's `resolveCompoundReceiverClass` fold does not. Measured in TypeScript only; generic repository/handler interfaces are the common DI shape in TS/Java/C# alike. Worth its own issue.
- **P3 — Case 3b has the same gap Case 0 just had.** It uses the same `resolveCompoundReceiverClass` and the same `[owner, ...mroFor(owner)]` walk and emits with no fan-out (`:1197`). Case 0.5's exclusion is documented; 3b's is not. If Case 0 needed this, say why 3b doesn't.
- **P3 — stale in-source comment outside the diff.** `languages/go/interpret.ts:31-33` says the raw `*T`/`T` shape is consumed "to model Go value and pointer receiver method sets **conservatively**". The first clause is still true; "conservatively" described exactly the #1966 exclusion this PR removes — and the PR's own rationale block argues that exclusion made the graph *silent*, not conservative.
- **P3 — the three disclosed follow-ups carry no issue numbers.** The uncapped fan-out, the blind detection benchmark, and the reporter's per-file inconsistency are all correctly scoped out, but none is tracked.

---

### Probed and cleared

Several plausible objections were chased and did not survive; recording them so they don't get re-litigated:

- **"Passing `currentClass` loses an interface owner found through the MRO."** Refuted twice independently. Case 4 passes the receiver's **static type** too, not the MRO ancestor that supplied `memberDef`, and `emitInterfaceDispatchFor` resolves by *name* per implementor via `pickOverload` — independent of which ancestor supplied the primary. Where it does decline (folded receiver is a concrete Class whose MRO contributed the interface method), declining is *correct*: fanning out there would emit edges to classes unrelated to the receiver.
- **"The `*T` reading admits a false implementor."** Refuted against the Go spec across value-embed, pointer-embed, depth-N promotion and non-addressable receivers — `*T`'s method set is a superset of `T`'s in every embedding configuration, so this is the least-restrictive *sound* choice given one node per type. A dedicated probe confirmed the ambiguity rule holds too: `type TwoBase struct { Base; Other }` with an equal-depth `Ping` collision correctly emits **no** `IMPLEMENTS` edge (`mergePromotedMethodEntry:440-452` marks it ambiguous and the method set drops it).
- *(That clearance is about embedding semantics only. The separate P2 above — `simpleQualifiedName` stripping the package qualifier so an unexported `seal()` matches across packages — is a different mechanism, is pre-existing, and is not cleared by it.)*
- **"This needs a `SCHEMA_BUMP`."** It does not, confirmed five separate ways: a cold-then-warm analyze that still emits the new edges; the parse-gate analysis (`SCHEMA_BUMP` is keyed to `ParseWorkerResult`/parse-time capture changes, and neither `go/captures.ts` nor `go/query.ts` is touched); `SCHEMA_FINGERPRINT` being a DDL digest that cannot move here; the fact that `goReceiverKind`'s *stamping* is untouched and only its consumption changed; and the runner-identity full-rebuild guard at `run-analyze.ts:1360`, which independently forces `force = true` on analyzer drift, so the upgrade path never takes the incremental writeback at all.
- **Double-emit / suppression via `handledSites`, `seen`, or collapse mode.** Refuted by exact counting: 20 sites × 40 implementors gave exactly 800 fan-out edges + 20 primaries, not 799 or 801. The dedup key includes the target id (`edges.ts:161-163`), so the primary edge to the declaration cannot mask an implementor edge; and `emitted += emitInterfaceDispatchFor(...)` running even when the primary returned `false` is correct on every path (`isPhantomCalleeRead` suppresses both halves; a missing caller id suppresses both).
- **The three flipped assertions.** Checked four separate times and legitimate every time. Each traces to the removed filter, each fails on pre-fix code, each carries an in-place rationale comment. This is a deliberate reversal of #1966, correctly documented.
- **Recorded artifacts.** Re-derived rather than trusted: the fixture count `102 → 110` is exactly `.go` file count + 1 synthetic row (101+1, 109+1); `expected-captures.json` is `+`-only, 8 entries, and the new captures sum to exactly 200 against a `capture_groups_fp` move of 2103 → 2303. `bench/scope-capture --check` passes for all 15 languages with go the only mover.
- **Shared-code language neutrality (AGENTS.md §42), TypeScript strictness, security surface.** Compliant: no language name appears in the added shared-code lines; the deleted `as GoMethodDefinition` cast has no replacement and no new `any`/unsafe cast was introduced; the changed files touch no I/O, no external input, no credentials; the 8 fixture files are inert Go source with no `go:generate`/`go:embed`/cgo/`init()`; zero bidi or invisible-format characters in the added lines.

---

### Test coverage

1. **No non-Go fixture exercises a compound interface receiver.** `receiver-bound-calls.ts` is shared, language-nameless code consumed by all 14 languages; every test added here is Go. All four existing non-Go `*-interface-dispatch` fixtures use bare-name receivers (`_repo.Save()`, `action.execute()`, `repo.find(1)`) and therefore hit Case 4, which already had the fan-out. The PR body says the corpus shows no delta and correctly explains why — but the measured delta is not zero: a two-implementor non-generic TypeScript fixture with the same call made through a **field** and a **parameter** side by side gives `runField → save` 1 → 3 edges and total CALLS 4 → 6, while `runLocal → save` stays at 3 [reproduced]. Those new TypeScript edges are the **real** `UserRepo.save` / `MockRepo.save` implementations — correct edges, the change doing what it intends. Put that number in the PR body, and add one field-receiver fixture per collapse class (Java/C# collapse; TS/Kotlin/Go don't) so a future regression here is catchable. It is also what would have caught the P1 above.
2. **No same-package coverage for the field case.** Every Go interface-dispatch fixture, including the new one, is cross-package. The closest pre-existing row (`RunInterface → DoWork`, #2766) isn't file-qualified on the target, so it cannot distinguish landing on the interface declaration from landing on the implementation — both live in `repo.go` and share the name. Prior investigation established the real discriminator is value- vs pointer-receiver, not package boundary, so this is very likely already fixed; nothing proves it. One row using the same `callsToFile()` helper this PR introduced would.
3. **No fan-out upper bound is asserted anywhere.** A single row of the form "N implementors × M sites produces exactly N×M edges" would make any future cap, dedup or collapse change observable — and would have made the growth in the P1 above visible during review rather than after.
4. **The new suite asserts `reason` but never `confidence`.** Changing the `0.85` literal at `:674` breaks nothing, whereas the sibling `IMPLEMENTS` block does pin `confidence === 0.85` (`go.test.ts:373`).
5. **The detection benchmark is structurally blind to D1** — correctly disclosed in the PR body, and confirmed: `generateSyntheticInterfaceData` never sets `goReceiverKind`, so the old filter excluded nothing there at any point in that file's history. Worth noting the scaling assertion it does have (`< 1.5`) currently measures `1.95` on **both** base and head, and the whole file is excluded from the benchmarks job by design — so it provides zero regression signal today, for reasons that pre-date this PR. Setting `goReceiverKind` on a realistic fraction of the synthetic methods is a cheap way to make the tripwire real.
6. **`signaturesCompatible` has no negative row for the case it exists for.** `PartialRepository` is the only negative control and it tests a *missing* method. There is no same-name/same-arity/different-signature row — and signature comparison is the sole remaining guard now that pointer-receiver methods are admitted.
7. **Minor:** `resolves the same field shape identically across two service files` asserts `'Queue → GetPickQueue@order_repo.go'`, but two distinct methods are named `Queue` in the fixture (`wave_service.go:17` and `picking.go:9`) and the helper doesn't qualify the *source* by file. The row passes if either resolves; it can't tell you which.

---

### CI

- **`tests / windows-latest (platform-sensitive) 3/3` — flake, not caused by this PR.** The failing file is `test/integration/cli-limit-e2e.test.ts` (`1 failed | 32 passed | 1 skipped`), and the error is `Analyze failed (status null)` thrown from `beforeAll`. `status null` means the subprocess was **killed** — it is the 60-second `runCliRaw` timeout firing (`cli-limit-e2e.test.ts:120`), not a logic failure. The fixture is the TypeScript `mini-repo` with no Go code at all, so D1 cannot execute on it and D2 has nothing to fan out. The sibling `cli-e2e.test.ts` had exactly this `status === null` subprocess-timeout failure under parallel load, fixed in #2000 by moving indexing into a retrying `beforeAll`; this file's own header says it copies that spawn pattern and it carries the same exposure. **Re-run the job.**
- **`Vercel` — non-actionable infra.** Fork PRs cannot authorize a deploy to the maintainer's Vercel project. No fix exists from the PR side.
- **`CI Gate`** is red solely because of those two.
- Everything else passes, including all three `ubuntu / coverage` shards (where `go.test.ts` runs), `benchmarks`, lint, format, both typechecks, CodeQL and gitleaks.

**Verdict: not ready as it stands, but close.** The blocking work is the sub-interface/abstract-class guard in `emitInterfaceDispatchFor` and a decision (cap-and-report now, or a tracked issue) on the fan-out bound. Everything else is comments, controls and coverage. The core fix is right and should land.

PR #2842, opened 2026-08-05
fix(scope-resolution): fan out interface dispatch from Case 3b receivers (#2832)

### LINKED ISSUE #2832: Case 3b (chain-typebinding) has the same missing interface-dispatch fan-out that Case 0 just fixed
Case 3b (chain-typebinding) resolves a receiver through the *same* `resolveCompoundReceiverClass` fold and the same `[owner, ...mroFor(owner)]` walk that Case 0 uses, but it emits its edge without calling `emitInterfaceDispatchFor`. So a chain-typed receiver that folds to an Interface stops at the interface's method declaration and never reaches any implementation — exactly the defect #2813 reported for field receivers.

This is the gap #2829 just closed for Case 0. Case 3b was left alone deliberately, to keep that PR's blast radius to one case, but the asymmetry is now unexplained: if Case 0 needed the fan-out, it is not obvious why 3b does not.

**Where**

- `gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts`, Case 3b — emits with no fan-out call.
- Compare Case 0 and Case 4, which both call `emitInterfaceDispatchFor`.

**Why it is not already covered**

Case 0.5's exclusion is documented in the case-ladder contract (`contract/scope-resolver.ts`, invariant I4). Case 3b's is not documented anywhere, and no test exercises a chain-typed receiver that folds to an interface. Case 3b is the Python/Ruby-shaped path, so this is likely invisible in the Go-heavy fixtures.

**Suggested work**

1. Construct a chain-typed receiver whose fold lands on an interface (Python/Ruby are the natural languages) and confirm it stops at the declaration.
2. Either add the fan-out — `emitInterfaceDispatchFor` self-gates on `ownerDef.type !== 'Interface'`, so it is inert otherwise — or document why 3b is deliberately excluded, next to the Case 0.5 note in invariant I4.

Follow-up from #2829 (fixes #2813).

magyargergo:
Closes #2832. Follow-up to #2829, which fixed the same defect in Case 0.

## The defect

Case 3b (chain-typebinding) folds a receiver to a type through the *same*
`resolveCompoundReceiverClass` call and the *same* `[owner, ...mroFor(owner)]`
walk that Case 0 uses, but emitted its edge without calling
`emitInterfaceDispatchFor`. When that fold landed on an `Interface`, the primary
edge named the interface's own bodiless declaration and **nothing reached any
implementation**.

That made the fan-out a property of how a receiver was *spelled* rather than of
what it resolved to:

```ts
d.repo.save();                    // Case 0  — fans out (since #2829)
const r = d.repo; r.save();       // Case 3b — did not fan out
```

Identical field, identical resolved type, different edge set.

## The fix

One call, mirroring Cases 0 and 4:

```ts
emitted += emitInterfaceDispatchFor(ownerDef, memberName, memberDef, site, 0.85, calleeCapture);
```

- `ownerDef` is the receiver's own folded type, **not** the owner of the member
  the MRO walk settled on. A receiver folding to a concrete class that merely
  inherits an interface method must not fan out, because its runtime type is
  that class.
- `emitInterfaceDispatchFor` self-gates on `ownerDef.type !== 'Interface'`, so
  this is inert for every concrete receiver and needs no language check —
  shared pipeline code must not name languages (AGENTS.md).
- Confidence is the flat `0.85` this case's own primary emits. Case 4 passes a
  `site.kind`-dependent value because *its* primary varies that way; Case 3b's,
  like Case 0's, does not.

Invariant I4 in `contract/scope-resolver.ts` now records that Cases 0, 3b and 4
fan out, leaving Case 0.5 the only case that folds or walks to a receiver type
without fanning out — its exclusion is deliberate and gated by the
`resolveThisViaEnclosingClass` language toggle.

## Tests

New fixture `test/fixtures/lang-resolution/ts-chain-interface-dispatch/` plus a
five-case block in the TypeScript resolver suite:

| Case | Asserts |
| --- | --- |
| primary edge | targets the interface declaration (`repository.ts`) |
| fan-out | reaches every implementation (`mem-repo.ts`, `sql-repo.ts`) |
| fan-out | never targets the interface declaration |
| negative | chain folding to a class implementing nothing emits no fan-out |
| negative | chain folding to a **concrete implementor** emits no fan-out |

The last one is the stronger negative: `SqlRepo` implements `Repo`, so an
interface is in scope and `save` is a name `Repo` declares, yet the folded type
is concrete and nothing may fan out. It is what fails if a later change fans out
from the interface a member is *declared* in rather than from the receiver's own
folded type.

Before the fix the fan-out assertion fails with
`expected [] to deeply equal [ 'mem-repo.ts', 'sql-repo.ts' ]` **while the
primary assertion passes** — the signature of "reached Case 3b and stopped at
the declaration", which is what distinguishes this from routing through the
wrong case.

## Verification

- `test/integration/resolvers/`: 3145 passed, 1 skipped
- Rest of `test/integration`: 4883 passed
- `test/unit`: 11389 passed
- `npx tsc --noEmit` clean; `npm run lint` 0 errors
- All five CI bench gates pass unchanged — `receiver-resolution`,
  `scope-capture`, `scope-emission`, `callable-value-flow`, `cpp-qualified-ns`.
  No rebaseline needed (#2829 required one; this does not).
- `impact emitReceiverBoundCalls` upstream: 1 direct dependent, risk LOW,
  epistemic exact. `detect_changes` vs merge-base: 4 symbols, 4 flows, all
  `EmitReceiverBoundCalls → …`.

No `SCHEMA_BUMP`, no capture-golden movement, no parse-cache version change —
the change is resolution-side only and the diff touches no capture, query, or
parse-cache file.

The remaining local failures (9 unit, 12 integration) are sandbox-environmental
and none are in resolution: tests whose own hard-coded timeouts are shorter than
CLI startup on this filesystem (measured 32.9 s vs a 10 s cap), the FTS
extension being unavailable, and OOM-respawn timing.

https://claude.ai/code/session_01NdZtWXQJUGB1ZGLH2YNw3o


magyargergo (commented):
## Multi-lane review — PR #2842

Reviewed at `eea2fdf2d` against merge-base `cabd5b82f`. Corrections landed during the review in `ffe10b7e3`; a second doc-only follow-up in `2bd326879` completes the I4 fix (see item 3 below). Head is now `2bd326879`.

### Review bar for this PR

Resolution-side change to shared ingestion code, so: no `SCHEMA_BUMP` and no parse-cache version change (verified — the diff touches no capture, query, or parse-cache file); no CHANGELOG entry, which `DoD.md` §4.5 makes release-process-owned and §6 lists as a Not-Done Signal if edited during PR work; no language names in shared pipeline code per AGENTS.md; targeted tests for the changed behavior; and the case-ladder contract comments kept accurate, since that file's own header states the numbering is part of the contract.

### Problem being solved

Case 3b resolves a receiver that is a bare name whose type binding is dotted — `const r = d.repo; r.save()`. It folds that chain to a type and walks the MRO, exactly as Case 0 does for `d.repo.save()`, but emitted its edge without the interface-dispatch fan-out. When the fold landed on an interface, the edge named the interface's bodiless declaration and nothing reached any implementation. The fan-out was therefore a property of how a receiver was *spelled*, not of what it resolved to. #2829 closed the Case 0 half; this closes Case 3b.

### Current PR state and merge status

Open, `fix/2832-case3b-interface-dispatch` → `main`, 4 commits. **Merge state: `mergeable`** — no conflicts. **Branch hygiene: `clean feature/fix PR`** — every commit is scoped to the Case 3b fan-out, 3 commits behind `origin/main` with zero file overlap, whitespace-clean, no bidi control characters.

### Repository history considered

Parent PR #2829 fixed this identical gap in Case 0 and filed #2832 for the Case 3b half; both feed the same `emitInterfaceDispatchFor` closure. #2829 also changed Go capture output and so had to rebaseline `bench/scope-capture/baselines.json` — this PR touches no capture, and all five bench gates pass unchanged.

### What was validated

The change itself survives. Ten lanes across two engines attacked it and none produced a defect in what this PR does.

- **`ownerDef` is the right argument.** Bound once at `:1263`, reassigned only in the `undefined` retry arm at `:1271`, never touched again before the fan-out — the MRO walk writes `memberDef`, a different binding. A Java `default`-method fixture was built to discriminate the two values: had the member's owner been passed, `runConcrete` would have fanned out to `MemRepo.save`, a class its receiver can never be. It does not.
- **No double emission.** Case 0 `continue`s at `:816`; Cases 3b and 4 are mutually exclusive on `rawName.includes('.')`; Cases 3 and 3b on `namespaceTargets`. Property-dispatch's separate `seen` is backstopped by first-write-wins `addRelationship` (`graph.ts:86-89`) on a position-keyed edge id.
- **Read/write fan-out matches Case 0 by construction** — neither gates on `site.kind` — confirmed empirically: `const c = h.cfg; c.tag()` yields an edge set byte-identical to `h.cfg.tag()`.
- **Truncation is reported, not silent.** A 33-implementor fixture kept 32 edges, incremented `dispatchFanoutSkipped`, and reached the warning at `run.ts:837`.
- **Confidence is self-consistent** — Case 3b's primary is 0.85 for every site kind, and the fan-out mirrors it; no secondary edge outranks the edge it hangs off.
- **No language-specific breakage** in Go, Python, Rust, PHP, Dart or Swift; a sibling interface with the same method name does not cross-contaminate, as the closure is keyed on `ownerDef.nodeId`.

### Corrected during review (`ffe10b7e3`, `2bd326879`)

1. **The test comment gave the wrong reason** why the concrete-implementor control cannot catch "member owner instead of folded type." It claimed an implementing class always declares the member itself — but `class C extends Base implements I {}` is valid TypeScript and inherits it. The real reason is that TypeScript's MRO chain never contains an implemented interface. The mutation *is* expressible in Java or Kotlin via a `default` interface method, so the gap is language-scoped, not inherent.
2. **`fans out to every implementation` certified completeness that does not exist.** TypeScript emits heritage edges for `class_declaration` only (`languages/typescript/captures.ts:749`, stated in that file's own docstring at `:732-733`), so `abstract class X implements I` and `interface B extends A` produce no heritage edge and still dead-end on the bodiless declaration. Renamed to name the shape actually covered, with a KNOWN GAP note.
3. **Invariant I4 overclaimed** that Case 0.5 is the only case resolving a receiver type without fanning out. *The lanes split on this:* the risk and docs/DoD lanes both walked the ladder and judged the original accurate, reading an implicit fold/MRO-walk qualifier into it; the correctness and adversarial lanes read it literally and found Cases 2, 3 and 5 refute it. The corrected text captures both — it now states the fold/MRO-walk distinction explicitly and notes why Case 2 (receiver *is* the type name, so static dispatch) and Cases 3/5 (direct lookup) are excluded, rather than reading as a completed audit. A first pass at this correction named only Cases 3 and 5 and was itself incomplete; Case 2 has since been added.
4. **The gate rationale was wrong for abstract receivers** — an abstract-class receiver also dead-ends on a declaration-only member and does not fan out. Reproduced: `runAbstract` gets exactly one edge, to a bodiless `isAbstract=true` declaration, identically for Cases 3b, 0 and 4 — so no regression, just an over-claiming rationale. Noted, with why widening the gate belongs to Cases 0 and 4 across all languages rather than to #2832.

Plus the module-level case ladder, which still credited the fan-out to Case 0 alone.

### Open — pre-existing, not introduced here

**Fan-out can emit a CALLS edge to a `static` method no instance dispatch can reach.**
*Risk:* a false-positive edge at confidence 0.85 that surfaces in `impact()` results and the `--pdg` callee-id sink.
*Evidence to check:* the closure picks with `pickOverload` (`:475` at the reviewed commit), which applies no static filter, while Case 3b's own primary four lines earlier uses `pickFirstNonStaticOnly` (`:1294`) — speculative edges picked with weaker rules than the certain edge. Reproduced: `class ShadowRepo extends SqlRepo { static save() }` yields `chainCall -> save<,static>`.
*Recommended fix:* the graph node already carries `isStatic`; add a sibling guard to the existing `isDeclarationOnly` check.
*Blocks merge:* no — Case 0 emits the identical wrong edge today; #2832 only multiplies the site population reaching it.

**TypeScript cannot see two of the three common interface-hierarchy shapes.**
*Risk:* the two shapes `receiver-bound-calls.ts:388-395` says the closure exists to handle emit zero dispatch edges in TS, including the docstring's own `AbstractHandler`/`RealHandler` example, where the only executable target receives nothing.
*Evidence to check:* `languages/typescript/captures.ts:749` and its docstring at `:732-733`. Since Case 3b's dotted-alias binding survives qualifier-stripping only in TS/JS, this is the language that actually reaches the new path.
*Recommended fix:* capture-layer change, emitting heritage for `interface_declaration` and `abstract_class_declaration`.
*Blocks merge:* no — predates this fan-out and affects Cases 0 and 4 equally.

### Coverage and limits

Lanes: risk, test/CI, docs/DoD, correctness, adversarial, testing, maintainability, project-standards on Claude; `ce-code-review` and an adversarial pass on Codex `gpt-5.6-sol`/xhigh. **Two engines, not ten independent reviewers** — agreement among Claude lanes is correlated and is not independent confirmation. The genuine cross-engine signal here is that Codex and Claude independently refuted every core defect charge and independently identified the same Java/Kotlin `default`-method shape as the missing control.

Three lanes executed against the real ingestion pipeline via `runPipelineFromRepo` (CE correctness, CE adversarial, CE testing); the other seven verified through source inspection, graph tool calls, and routing analysis. The Codex adversarial leg was blocked from running Vitest by a sandbox Bubblewrap failure, so its red-team conclusions rest on source-path analysis rather than executed fixtures — they match the Claude adversarial lane's executed output on every overlapping charge. A third Codex leg (swarm-solo) died on an org token-per-minute limit and contributed nothing; three concurrent legs was the mistake, and the other two were re-run serially and completed clean.

Not run, and therefore not covered: branch hygiene (verified by hand instead, classified above) and the security/trust-boundary lanes — the diff has no external input, persistence, process execution, network or auth, and `explain()` returns zero taint findings.

Still untested after this PR: read/write-site fan-out on any language, the 32-target cap from Case 3b specifically, and the member-owner mutation on any language — constructible in Java today.

Local suite: resolvers 3145 passed; rest of integration 4883 passed; unit 11389 passed; the six resolver suites carrying interface-dispatch pins pass 1226/1226. The 21 local failures are environmental — this box's CLI startup exceeds several tests' own hard-coded child-process timeouts, and the FTS extension is absent — and none are in resolution.

### Verdict

**production-ready with minor follow-ups.** No defect in what this PR changes, confirmed across two engines with executed fixtures. The overclaims it shipped are corrected; the two open items are pre-existing subsystem gaps that predate it and affect Cases 0 and 4 equally, and each warrants its own issue.

PR #2843, opened 2026-08-05
fix(go): scope and define each type_spec, not the type_declaration (#2837)

### LINKED ISSUE #2837: Go: interface-field calls resolve inconsistently BY FILE at scale (42 edges vs 0, same declared shape) — unexplained by #2829
Splitting this out of #2813 so it is not silently closed by #2829, which fixes the other two asks in that issue but **does not explain this one**.

#2813 reported that two files declaring the same field shape resolved completely differently on a real ~470k-LOC Go codebase. Counting `CALLS` edges landing on `internal/repository/interfaces.go` by source file:

| source file | edges | field-receiver call sites (grep) |
| --- | --- | --- |
| `internal/services/wave_service.go` | **42** | 57 |
| `internal/services/shopify_sync.go` | 5 | 108 |
| `internal/handlers/orders.go` | 4 | 50 |
| `internal/services/pick_service.go` | **0** | 71 |

The reporter notes that all four declare the same kinds of field — `wave_service.go` and `pick_service.go` both hold `orderRepo repository.OrderRepository` and both use pointer receivers — yet one resolves most of its interface-field calls and the other resolves none.

**What #2829 rules out**

That PR fixes two stacked defects (pointer-receiver methods excluded from structural satisfaction; Case 0 never fanning out) and makes the Go method-set model exact. Both were *global* — they affected every file equally — so neither can produce a per-file split. The committed fixture deliberately includes two service files with identical field shapes (`wave_service.go`, `pick_service.go`) and asserts, with both ends file-qualified, that they resolve **identically**. They do.

**So this needs the real corpus.** Hypotheses worth testing there, none verified:

- something file- or chunk-scoped in the parse/worker path (file size ceiling, chunk boundary, a per-file cap or timeout);
- import-resolution differences between the two files that change how `repository.OrderRepository` binds;
- a build tag or a file that fails to parse and degrades silently;
- ordering/first-write-wins in an id-keyed index.

**Useful next step**

The reporter offered to run targeted queries or Cypher probes against their index. Comparing, for one interface method, the `resolutionOutcomes` drop records between `wave_service.go` and `pick_service.go` would likely separate "receiver never typed" from "typed but no edge emitted" and narrow this quickly.

Original report: #2813. Fix for the other two asks: #2829.

magyargergo:
Addresses #2837 — Go interface-field calls resolving inconsistently **by file** at scale (42 edges vs 0 for the same declared shape), which #2829's two global fixes could not explain.

## Root cause

Five Go captures were anchored on the enclosing `type_declaration` instead of the `type_spec`:

| capture | file |
| --- | --- |
| `@scope.class`, `@declaration.struct`, `@declaration.interface` | `languages/go/query.ts` |
| `@definition.struct`, `@definition.interface` | `GO_QUERIES` in `tree-sitter-queries.ts` |

An idiomatic grouped declaration

```go
type (
	Decoy       struct { orderRepo *LocalThing }
	PickService struct { orderRepo repository.OrderRepository }
)
```

matches each pattern once per `type_spec`, but every match named the **same node**. Running the scope query against `tree-sitter-go` shows it directly:

```
class-scope captures: 2
  node=type_declaration first-line="type ("                     type_specs=[Decoy, PickService]
  node=type_declaration first-line="type WaveService struct {"  type_specs=[WaveService]
```

Two structs, one Class scope, one graph node for the whole block.

Downstream, a struct with no Class scope has no field `typeBindings` for `typeOfMemberOnClass` to find (`compound-receiver.ts:334-338`), so the Case 0 compound-receiver fold declines and **every** `s.field.Method()` site in that file emits nothing at all. Silent, per-file, independent of file size — exactly the reported signature.

Measured on the new fixture before the fix:

| file | shape | edges |
| --- | --- | --- |
| `wave_service.go` | plain `type X struct` | resolved (control) |
| `pick_service.go` | struct **second** in a grouped block | 0 |
| `sort_service.go` | struct **first** in a grouped block | 0 |
| grouped `interface` block | — | no `IMPLEMENTS` edges at all |

So the collapse cost every type in a grouped block, not only the ones after the first, and `PickService` / `MetricSink` were absent from the node inventory entirely — meaning `impact("PickService")` returned a clean, wrong zero.

**Why no test caught it:** zero of this repo's 115 Go fixture files used a grouped `type (...)` declaration, including #2829's own `go-interface-field-dispatch` fixture. Fixtures get written in the minimal style; real Go does not.

## The fix

Anchor all five captures on the `type_spec`, so each declared type gets its own scope and its own node whether it is declared plainly or inside a block.

The five must move in **lockstep**, and this is now written into both query files: with the scope on `type_spec` but the def still on `type_declaration`, the def is strictly larger than its own scope, nothing is owned, and **every** Go field-receiver edge disappears — plain declarations included. That intermediate state was measured, not theorised.

Deliberately **not** fixed in `workspace-index.ts` by mapping every class-like def per scope: that leaves the grouped structs sharing one name-keyed `typeBindings` map, trading a missing edge for a possibly-wrong owner. `compound-receiver.ts:406-407` states the rule this repo follows — a missing edge is recoverable, a confidently wrong one is not. The fixture's `Decoy` struct declares a same-named `orderRepo` field of a different type specifically to pin that.

## Also in this PR

- **`SCHEMA_BUMP` 43 → 44.** Parse-time capture change: without it a warm parse cache replays pre-fix `ParsedFile`s and the fix is a silent no-op on every incremental analyze while still passing every cold-run test. Re-checked against `origin/main` immediately before editing.
- **One comment-aware Go package-clause resolver.** Both package-bucketing passes carried a byte-identical `sourceText.match(/^\s*package\s+(\w+)/m)`. The `m` flag matches a `package` line inside a **block comment** — measured: `/*\npackage legacy_notes ...\n*/\npackage services` returns `legacy_notes`. That isolates the file in a singleton package bucket in both passes, producing the same silent per-file signature by a different route. Replaced with one resolver that follows the spec rule (a file's first non-comment token *is* `package`); both duplicate pairs deleted.
- **Per-file receiver-drop attribution in the analyze log.** #2837 asked for a comparison of `resolutionOutcomes` drop records between two files — that was not runnable on any corpus, because `summarizeUnresolvedReceivers` keys by member name (cap 500) and discards `filePath`. One bounded top-10-files line, no new persisted field. Verified live on this repo: 65008 sites across 1612 files.
- **A bounded warning** where a Go file with no resolvable package clause was previously dropped from ownership resolution by a bare `continue`.

## Verification

- Go integration `172/172`, unchanged.
- New `go-grouped-type-decl` fixture: 7 assertions, RED before the fix and green after — plain / grouped-second / grouped-first / reverse-order control / grouped interface block / same-named decoy field / node inventory.
- Full unit suite 11538 passed. Four failures in `skip-git-cli` and `analyze-heap-respawn` were CLI-subprocess timeouts under load; each passes in isolation.
- `tsc --noEmit` clean, `npm run lint` 0 errors.
- `bench/scope-capture --check` PASS across 15 languages, Go the only mover, scaling 1.054 against its 1.5 budget.
- Captures golden diff read before regenerating: 70 digests changed, **0 `captureGroups` counts changed**, 0 keys removed, 6 added — the signature of a capture moving to a narrower node with nothing lost. Regenerated once, at the tip.

Empirically probed against shapes the fixture does not contain — generics (`type Pair[K,V] struct`), type aliases, named types, embedded pointers, mixed grouped blocks, doc comments, nested anonymous structs. `type ( A struct; B = Other; C interface; D int )` yields exactly two scopes, `[A, C]`.

## Honest limitation

This fixes a **measured** defect of exactly the class #2837 reports, with a coverage gap that explains why it went unnoticed. It is **not** proven to produce the reporter's specific 42 / 5 / 4 / 0 numbers — that needs their corpus.

@abhigyanpatwari — before closing #2837, it would be worth asking the reporter to check whether the affected service structs are declared inside grouped `type (...)` blocks (`grep -n '^type ($' internal/services/*.go`), and to re-run analyze on the fixed build: the new per-file drop line will show whether `pick_service.go` still loses its receivers. I have left #2837 open rather than auto-closing it for that reason.

Refs #2837, #2813, #2829

https://claude.ai/code/session_01N9YZcj7FvaFrrisZEFWBNX


koriyoshi2041 (approved):
Checked the capture/definition lockstep, package-clause normalization, cache bump, and bounded receiver-drop attribution against the final head. The focused package/ownership/cache suites pass 61/61 locally; hosted Go integration, typecheck, lint, and the full 16,905-test matrix are green. Grouped structs/interfaces now keep distinct scopes without sharing field bindings, while plain declarations and unsupported package clauses retain explicit controls. Looks ready.

koriyoshi2041 (approved):
Re-reviewed 7204c1b9 after the post-approval refactor. The type_spec-only ownership path, shipped-query lockstep tests, shared package-dir helper, bounded diagnostics, and lazy receiver logger preserve the earlier invariants. Local focused Go/package/receiver suites passed 205/205 and TypeScript compilation passed; the hosted matrix is green. Approving the new head.

Issue #1, opened 2025-08-31
Add support for Ollama as a local inference backend


Rage997:
It would be great to add Ollama as an alternative backend to OpenAI. This would let users run GitNexus with local LLMs, improving privacy, avoiding API costs, and enabling offline use.

abhigyanpatwari:
Great idea. Will do

Phum7895:
hi

abhigyanpatwari:
Hi. Sorry this is taking long since I am working an embedings pipeline too. Combining RAG + Graph RAG for better retireval while working in browser constraints

Rage997:
@abhigyanpatwari no need to apologise mate. It's great that you are doing this for free and open source keep the good work up!

abhigyanpatwari:
Ollama provider support is added. But the agent through ollama behaves weird, might need custom implementation instead of langchain agent for this.

abhigyanpatwari:
Ollama provider support is added. But the agent through ollama behaves weird, might need custom implementation instead of langchain agent for this.

Issue #2798, opened 2026-08-02
Replace the hand-incremented INCREMENTAL_SCHEMA_VERSION correctness check with a DDL fingerprint


magyargergo:
## Problem

`INCREMENTAL_SCHEMA_VERSION` (`gitnexus/src/storage/repo-manager.ts`) is a hand-incremented integer that has to predict a derived fact: whether the on-disk DDL matches the code's DDL. Its own comment records the consequences:

> *"the seventh collision in this series and the first exact clash. Re-check against origin/main before merge."*

That is a **manual pre-merge ritual defending a hand-maintained value** — structurally the same disease that PR #2793 just cured one file away for the relation pair list. #2793 was the eighth occurrence (it and #2791 both claimed 35; #2791 was closed, so it resolved itself).

The failure mode when it does collide is not loud. Two builds that both stamp `35` with *different* DDLs pass the `===` reuse gate, and the mismatched edges are then silently dropped by `fallbackRelationshipInserts`' `catch { /* skip */ }` (`gitnexus/src/core/lbug/lbug-adapter.ts`). A wrong graph, not an error.

## Proposal

Store a fingerprint of the actual schema in `RepoMeta` alongside `schemaVersion` — e.g. `sha256(RELATION_SCHEMA + NODE_SCHEMA_QUERIES)` — and compare it on load.

- The integer stays for human ordering and for the user-facing "index schema changed (stamped vN, this build is vM)" message.
- The **correctness** check becomes derived and cannot collide: two branches that genuinely produce the same DDL agree, and two that don't are detected regardless of what integer either chose.
- No renumbering needed when two PRs pick the same number, which is what happens every time.

This is cheaper now than it has ever been: after #2793 the relation DDL is *generated*, so the fingerprint is a pure function of code already in hand — no new bookkeeping.

## Context

Raised by the multi-engine review of #2793. Related: the `schema-version-collision` history in the `repo-manager.ts` version ladder comment.

PR #2854, opened 2026-08-06
fix(analyze): gate FTS-indexed DML before the incremental writeback (#2841)

### LINKED ISSUE #2841: Incremental analyze crashes with opaque "COPY failed ... Binder exception" when the DB carries FTS indexes but the FTS extension fails to load
### Area

gitnexus (CLI / core / indexing)

### Environment

gitnexus 1.6.9, @ladybugdb/core 0.18.3, win32-x64, Node v22.22.2

### Summary

When a `.lbug` DB already contains FTS indexes (built by an earlier analyze where FTS was available) and a later **incremental** analyze runs in an environment where `LOAD EXTENSION fts` fails, the run warns `FTS extension unavailable; continuing without FTS features` — but then hard-crashes during the incremental load with `COPY failed for <table>: ... Binder exception`, leaving a partially-loaded DB (`loadGraphToLbug` is documented non-transactional). Recovery requires `analyze --force`, which succeeds but **silently degrades search**: the FTS phase graceful-skips, and afterwards `CALL SHOW_INDEXES()` shows only `_PK` indexes — hybrid search quietly loses BM25 until someone notices.

### Why this is worse than a plain crash

- The Binder exception surfaces at the storage layer (writes to tables that carry FTS indexes need the extension) and its message does not mention FTS at all. In our case the operator mis-diagnosed it as "extension needs network install / environment limitation" — while the extension file was present locally under `~/.lbdb/extension/...` and perfectly loadable in a healthy shell (`gitnexus doctor` → `Full-text search: available`).
- The ExtensionManager graceful-degradation contract ("unavailable extensions degrade search features without blocking analyze") breaks specifically when the DB *already* carries FTS indexes: `dist/core/run-analyze.js` gates FTS index *creation* on `loadFTSExtension` (the `ftsAvailable` check around the FTS phase), but the node COPYs in `copyNodeCSVs` run *before* that phase with no such gate.

### Expected behavior

At analyze startup, before any writes: if the catalog contains FTS indexes but the FTS extension fails to load, either

1. fail fast with an actionable message (e.g. "DB has FTS indexes but the FTS extension failed to load in this environment — fix the environment, or run `--force` to rebuild without FTS"), or
2. explicitly `DROP_FTS_INDEX` the stale indexes (announcing it loudly) and continue degraded — matching the existing graceful-skip contract.

Either way the opaque mid-COPY Binder exception and the partially-loaded DB are avoided.

### Repro sketch

1. `gitnexus analyze` a repo in an environment where FTS loads → DB gets FTS indexes.
2. Make the extension unloadable (e.g. move `~/.lbdb/extension/<version>/` away, or run under a HOME where it does not exist).
3. Touch a few files, run incremental `gitnexus analyze` → warn + `COPY failed for <table>: ... Binder exception`, partially-loaded DB.
4. `gitnexus analyze --force` → succeeds, but FTS indexes are silently gone.

Related: #1275 / #2184 (offline INSTALL availability), #2456 (no-force analysis / recovery design) — this issue is specifically about the *crash + silent degradation* when indexes already exist, not about how to install the extension offline.

magyargergo:
Fixes #2841.

## The bug

An incremental `analyze` over a `.lbug` DB that carries FTS indexes dies mid-writeback when the FTS extension cannot load:

```
Incremental: changed=0, added=1, deleted=0 (skipping wipe + 6 unchanged file rows preserved)
  Analysis failed: Binder exception: Trying to delete from an index on table File but its extension is not loaded.
```

The output never mentions FTS — the reporter mis-diagnosed it as a network/install problem while the extension file sat on disk. Reproduced on `main` at 817893df9 with a hermetic `HOME`, and also on 1.6.10-rc.168 by the reporter (there it dies one step earlier, in the delete rather than the COPY).

## Why it happens

Three things stack up:

1. **The only install-capable FTS load happens after all DML.** `runFullAnalysisInner` loads FTS with the analyze policy in Phase 3, long after the incremental writeback deleted and COPYed rows. The earlier pre-load in `doInitLbug` is deliberately `{ quiet: true }` under the global `load-only` policy, so its miss is a debug line — hence no FTS diagnostics at the crash.
2. **The #2589 pre-delete drop silently no-ops in exactly this state.** `dropSearchFTSIndexes()` → `dropFTSIndex` → `CALL DROP_FTS_INDEX` fails with `Catalog exception: function DROP_FTS_INDEX is not defined`, which `isBenignDropFtsIndexError` classifies as "nothing to drop". The indexes survive; the mitigation is load-bearing precisely when it cannot run.
3. **The engine refuses the DML at BIND time.** Probed against `@ladybugdb/core` 0.18.3: a `DETACH DELETE` matching *zero* rows fails identically to one matching thousands, on `File` and on `Function`. There is no "only when rows change" narrowing — 20 tables carry FTS indexes.

Worth stating explicitly, because it rules out the issue's own suggested remedy: **there is no non-extension way to drop an FTS index.** `DROP INDEX file_fts ON File` and `DROP INDEX file_fts` are both parser errors; `CALL DROP_FTS_INDEX` is the only drop verb and it lives in the extension that failed to load. Rebuilding the DB file is the only way to clear the indexes.

## The fix

Exactly the shape #2623 already established for VECTOR, whose comment states the principle: *"the index cannot be dropped without the extension either, so surgery is impossible: fall through to the escalation valve's wipe-and-COPY plan."*

- **`ensureFtsRowDmlSafe()`** (`lbug-adapter.ts`) — the FTS twin of `ensureEmbeddingRowDmlSafe`: read the catalog first (no extension needed), and only when an FTS index actually gates DML, load FTS with the analyze policy. A DB that never carried FTS pays nothing.
- **`run-analyze.ts`** reads the index catalog **once** and derives both verdicts before any DML, then ORs `!ftsRowDmlSafe` into the existing escalation predicate. When in-place surgery is impossible the run switches to wipe-and-bulk-COPY and completes, with keyword search degraded — instead of dying.
- **The reason log names every blocked extension**, not just the first checked. A DB can carry both a vector index and FTS indexes; reporting half the cause is how this stayed mis-diagnosed.
- **`dropFTSIndex` no longer reports a drop that could not happen.** The message classifier stays pure — a string cannot tell you whether an index is live — so liveness is settled with a catalog read on the *error path only*: live index + unloaded extension now raises an FTS-named, remedy-bearing error. With the gate in place this is unreachable from `runFullAnalysisInner`; it exists so a future call site cannot silently re-open the hole.
- **`readIndexCatalogRows()`** is shared by both gates, so adding the FTS check costs no extra catalog round-trip (2 reads → 1).

Deliberately unchanged: `loadFTSExtension` (the read-only pool adapter shares it), the `{ quiet: true }` pre-load, `isBenignDropFtsIndexError`, and `--repair-fts` (already gated).

## Verification

CLI, the reporter's exact sequence — analyze with FTS, remove the extension store, touch a file, analyze again:

```
Incremental: this index carries FTS search indexes but the FTS extension could not be loaded, so no
indexed table can be written in place (LadybugDB refuses the write at bind time, and the indexes
cannot be dropped without the extension either) — switching to a full DB write (wipe + bulk COPY)
for this run. Full-text/BM25 search stays degraded until FTS is available. Run `gitnexus doctor` for
live extension status, or set GITNEXUS_LBUG_EXTENSION_INSTALL=auto to allow one bounded install attempt.
```

Exit 0, index rebuilt, FTS degraded and said so.

New `test/unit/incremental-index-extension-dml-gate.test.ts` drives the real `runFullAnalysis` against a real mini-repo and a real LadybugDB:

- FTS indexes present + FTS unloadable → escalates, names FTS, ends with zero FTS indexes, new content in the graph;
- FTS available → surgical plan kept, indexes kept;
- FTS never built → no escalation (the catalog-first check must not tax FTS-less machines);
- FTS **and** VECTOR blocked → one escalation naming both.

`drop-fts-index-error-classification.test.ts` gains the two `dropFTSIndex` cases (live index rejects; absent index still resolves), with the existing classifier assertions untouched. The CLI e2e reproduces the reporter's journey and skips visibly when the seeded extension cannot load on the host, so it can never go red for an unrelated reason.

**Mutation-verified:** reverting the run-analyze gate makes the first scenario fail with the original Binder exception; reverting the `dropFTSIndex` guard makes the live-index case fail.

`tsc --noEmit` and `eslint` clean. `incremental-fts-drop-ordering`, `incremental-vector-extension-ordering` and `run-analyze` pass (41/41), so the #2589 ordering and the #2623 behaviour are intact.

## Notes for reviewers

Two corrections to the issue text, both verified: analyze exits **1** on the crash, not 0; and on a networked machine the default `auto` policy re-installs the extension during the failed run, so the *next* run passes — the wedge only persists offline, which matches the reporter's environment.

https://claude.ai/code/session_01CxokkpfssUCxvBwRZMtRCB


gitnexus/src/core/run-analyze.ts
```diff
@@ -2122,8 +2124,20 @@ async function runFullAnalysisInner(
...
+      // function (there is no SQL `DROP INDEX` at all), so the indexes cannot be
+      // cleared in place either. Same verdict, same remedy: escalate. Both gates
+      // share ONE `SHOW_INDEXES` read — they answer different questions about the
+      // same catalog snapshot, and nothing between here and the write plan
+      // creates or drops an index.
+      const indexCatalogRows = await readIndexCatalogRows();
+      const embeddingRowDmlSafe = await ensureEmbeddingRowDmlSafe(indexCatalogRows);
+      const ftsRowDmlSafe = await ensureFtsRowDmlSafe(indexCatalogRows);
+      const extensionForcedRebuild = !embeddingRowDmlSafe || !ftsRowDmlSafe;
+      if (extensionForcedRebuild && cachedEmbeddings.length === 0) {
```
magyargergo: **P1 · `analyze --drop-embeddings` is silently reverted here.** Blocks merge: yes. `[code-read]` — corroborated across **both engines** (Codex ce conf-100, Codex adversarial, Claude adversarial).

**Trigger:** a repo analyzed once with FTS available (so the DB carries all 20 FTS indexes), holding embeddings but **no `embeddingCheckpoint`**; the machine's FTS extension stops loading; commit one file; run `gitnexus analyze --drop-embeddings`.

**Why it fires:** the `options.force = true` conversion at `:1177-1180` sits *inside* `if (existingMeta?.embeddingCheckpoint)` (`:1158`), so without a checkpoint the run stays incremental and reaches this gate. `--drop-embeddings` then *deliberately* leaves `cachedEmbeddings` empty (`:1527`, logging `Dropping N existing embeddings` at `:1545`) — which is exactly this condition. `loadCachedEmbeddings()` re-reads every row, Phase 3.5 restores them, Phase 5 stamps `stats.embeddings = N`, exit 0. The log even says `Preserving N embedding row(s) … (the index metadata did not account for them)`, asserting a cause that is false here.

This block pre-dates the PR, but its old trigger needed an HNSW index *and* a broken VECTOR extension. Widening it from `!embeddingRowDmlSafe` to `extensionForcedRebuild` moves it onto the dominant path, because **every** analyzed DB carries FTS indexes. The repo pins the opposite contract at `test/unit/run-analyze.test.ts:976`.

**Variant:** `--drop-embeddings --embeddings` is also defeated — the rescue repopulates `cachedEmbeddingNodeIds`, which `:2723` feeds to `runEmbeddingPipeline` as the already-embedded set, so the pipeline skips exactly the nodes the user asked to regenerate.

**Fix:** `if (extensionForcedRebuild && !options.dropEmbeddings && cachedEmbeddings.length === 0)`, plus an FTS-blocked + `dropEmbeddings: true` case asserting `stats.embeddings === 0`.

gitnexus/src/core/run-analyze.ts
```diff
@@ -2144,22 +2158,42 @@ async function runFullAnalysisInner(
         }
       }
       if (
-        !embeddingRowDmlSafe ||
+        extensionForcedRebuild ||
```
magyargergo: **P2 · This newly routes the default path into an escalation that deletes the LIVE index in place.** Blocks merge: no (availability window, not data loss). `[code-read]` — **both engines** (Codex adversarial `[high]`, Claude adversarial + correctness; Claude reliability/risk verified the recovery contract).

**Trigger:** any machine where the FTS extension stops loading, any previously-analyzed repo, the very next `gitnexus analyze`. No flags — the #2841 reporter's exact machine.

**True locations (outside this diff):** `buildPath` is frozen at `:1762-1780`, ~380 lines before this decision — `wantAtomicIncremental` requires `GITNEXUS_ATOMIC_INCREMENTAL=1` (off by default), so `buildPath === lbugPath`. The escalation at `:2222-2226` then runs `stop() → closeLbug() → wipeLbugDbFiles(buildPath) → initLbug(buildPath)` followed by a multi-minute full-graph COPY. On `main` this same condition failed at LadybugDB **bind time**, before any row was mutated, leaving a fully intact index; now it deletes it first.

Why P2 and not P1: the `escalated-full-write` dirty stamp at `:2207-2213` precedes the wipe, so a crash in the window forces a clean rebuild next run. The cost is an unqueryable index plus redone work, not silent loss. Ironically `GITNEXUS_ATOMIC_INCREMENTAL=1` is strictly safer here (the wipe hits the staging copy) but cannot be reached from this gate.

**Fix:** defer the `buildPath`/`useAtomicSwap` resolution until after the extension gate so an extension-forced rebuild can stage-and-swap like a full rebuild — or, minimally, say in the escalation log that the live index is about to be deleted and rebuilt.

gitnexus/src/core/lbug/lbug-adapter.ts
```diff
@@ -3279,10 +3338,39 @@ export const queryFTS = async (
...
+ * Benign only when there is genuinely nothing to drop — see `dropFTSIndex`.
+ */
+const DROP_FTS_INDEX_UNDEFINED_SIGNATURE = 'function DROP_FTS_INDEX is not defined';
+
+/** Whether `indexName` is currently present on `tableName` in the catalog. */
+const ftsIndexExistsInCatalog = async (tableName: string, indexName: string): Promise<boolean> => {
+  const rows = await readIndexCatalogRows();
+  // Unreadable catalog: cannot prove the index is live, so keep the
+  // long-standing tolerate-and-continue rather than inventing a failure.
+  if (rows === undefined) return false;
```
magyargergo: **P2 · This fails OPEN, contradicting the contract this PR introduces two functions above.** Blocks merge: no (no reachable production caller today). `[code-read]` — **both engines** (Codex swarm made this its stated basis for requesting changes; six Claude lanes reached it independently).

`readIndexCatalogRows`'s own docstring at `:3048-3049` says: *"`undefined` means 'could not prove anything' and **every caller must treat it as fail-closed** (assume an index may be present), never as 'no indexes'."* Both gates honour that (`:3117`, `:3155` — `rows === undefined || rows.some(...)`). This line does the opposite: an unreadable catalog is reported as *index absent*, so `dropFTSIndex` (`:3387-3398`) tolerates the error and returns normally, and the caller proceeds as if the index were gone — reintroducing the exact #2841 symptom this guard exists to make loud.

**Trigger:** `dropFTSIndex` called with the extension unloaded **and** a live index present **and** `CALL SHOW_INDEXES()` throwing at that moment.

Not merge-blocking because the new throw has no reachable production caller — all three call sites sit downstream of an extension-load gate. But defence-in-depth is this guard's only job, which is precisely why the fail direction matters.

**Fix:** `if (rows === undefined) return true;` — prefer a false-positive loud error over a false-negative silent swallow — or state in the docstring that this is a deliberate, reasoned exception to the fail-closed contract.

gitnexus/src/core/lbug/lbug-adapter.ts
```diff
@@ -3279,10 +3338,39 @@ export const queryFTS = async (
...
+/**
+ * The half of {@link isBenignDropFtsIndexError}'s catalog case that means "the
+ * FTS extension is not loaded", as opposed to "this index does not exist".
+ * Benign only when there is genuinely nothing to drop — see `dropFTSIndex`.
+ */
+const DROP_FTS_INDEX_UNDEFINED_SIGNATURE = 'function DROP_FTS_INDEX is not defined';
+
+/** Whether `indexName` is currently present on `tableName` in the catalog. */
+const ftsIndexExistsInCatalog = async (tableName: string, indexName: string): Promise<boolean> => {
+  const rows = await readIndexCatalogRows();
```
magyargergo, line 3513: **P2 · This costs 20 extra `SHOW_INDEXES` reads on *every* incremental run on an FTS-less machine.** Blocks merge: no. `[code-read]` — Claude performance + reliability + correctness (Engine A only, three independent lanes).

**Trigger:** a machine where the FTS extension never loads (`…INSTALL=never`/`load-only`, offline `auto`, or a missing runtime dep) analyzing a DB that carries **no** FTS index — i.e. exactly the configuration this PR's own test at `incremental-index-extension-dml-gate.test.ts:193-202` sets out to keep cheap.

**Chain:** `ensureFtsRowDmlSafe` correctly returns `true` early *without* loading the extension (`:3154-3156`), so the connection has no FTS. The surgical branch then runs `dropSearchFTSIndexes()` (`run-analyze.ts:2245`), which loops all **20** entries of `FTS_INDEXES` (`gitnexus/src/core/search/fts-indexes.ts:198-202`, outside this diff). Each `CALL DROP_FTS_INDEX` raises `Catalog exception: function DROP_FTS_INDEX is not defined`, and each now reaches this line for a **fresh** catalog read, under `withConnLock`.

That contradicts the docstring shipped with it at `:3371-3372`: *"the liveness question is settled with a catalog read on the ERROR path only (the healthy path still costs nothing)"*. On this machine the error path **is** the healthy path, and it never self-heals because there is nothing to heal.

**Fix:** thread the `indexCatalogRows` snapshot already read at `run-analyze.ts:2136` into the drop path and check membership in a `Set` — or skip the drop loop entirely when the gate already proved no FTS-typed index exists.

gitnexus/src/core/run-analyze.ts
```diff
@@ -2144,22 +2158,42 @@ async function runFullAnalysisInner(
...
+            `the ${EMBEDDING_TABLE_NAME} vector index exists but the VECTOR extension could not be ` +
+              `loaded, so embedding rows cannot be rewritten in place`,
+          );
+          degradedEffects.push(
+            'Semantic search falls back to exact scan until VECTOR is available.',
+          );
+        }
+        if (!ftsRowDmlSafe) {
+          blockedCauses.push(
+            `this index carries FTS search indexes but the FTS extension could not be loaded, so no ` +
```
magyargergo: **P2 · This message asserts facts the code explicitly could not prove — and "this index" points at the wrong subject.** Blocks merge: no. `[code-read]` — Claude adversarial + reliability + maintainability.

**Overclaim.** When `CALL SHOW_INDEXES()` throws (`lbug-adapter.ts:3058`), both gates correctly fail closed and both extension loads fail on an extension-less machine — so the run escalates and prints that a `CodeEmbedding` **vector index exists** and that the index **carries FTS search indexes**. Neither was established, and on a repo that never enabled embeddings the first claim is affirmatively false. The only truthful signal — `logger.warn('Could not read the LadybugDB index catalog…')` at `lbug-adapter.ts:3061-3066` — goes to pino's stderr, *not* the `onLog` callback this message uses, so under `gitnexus serve` / the analyze worker the UI shows only the false claim. That is the mis-diagnosis failure mode #2841 exists to eliminate.

**Dangling antecedent.** With both causes live, `blockedCauses.join('; and ')` yields *"…the CodeEmbedding vector index exists … **and this index** carries FTS search indexes…"*. Grammatically "this index" points at the vector index just named; the FTS indexes live on 20 other tables. Standalone it also reads oddly — an index does not carry indexes.

**Fix:** thread the catalog-read outcome here and emit a distinct third cause ("the index catalog could not be read, so neither gate could be settled"); replace "this index" with a self-contained subject ("the graph store carries one or more FTS search indexes…"); mirror the catalog-read warning into `log()` so IPC consumers see it.

magyargergo (commented):
# PR #2854 — two-engine review digest (`fix(analyze): gate FTS-indexed DML before the incremental writeback`, fixes #2841)

Base `021ac3037` → head `7aa52e563`. 5 files, +576/−32. Synthesis re-verified every postable claim against the worktree at `/workspace/.worktrees/pr-2854`.

---

## 0. Verdict and classifications

| Field | Classification |
|---|---|
| **Branch hygiene** | `clean feature/fix PR` — 3 atomic commits, linear, no merge commits in `021ac3037..7aa52e563`, 5 files, single-issue focus, no artifacts or whitespace churn |
| **Merge status** | `review blocked` — `mergeable=MERGEABLE`; `mergeStateStatus=BLOCKED` reflects the missing required review only. Vercel `fail` is the known non-actionable fork-deploy authorization check |
| **Final verdict** | **`not production-ready`** — one blocking finding (H1). Everything else is a follow-up |

**Review bar applied:** DoD §2.7 (new tests deterministic), §2.8 (operability — honest reporting), §4.5 (no CHANGELOG edit), the auto-derived `SCHEMA_FINGERPRINT` rule from #2798 (no bump owed — verified, not assumed), CI green, and regression coverage for the trigger scenario the PR claims to fix.

**Problem being solved (#2841):** an incremental writeback against a DB that carries FTS indexes dies at LadybugDB **bind time** when the FTS extension cannot load on the analyzing machine — a zero-row `DETACH DELETE` fails as hard as a large one — and the failure message never names FTS, so the reporter mis-diagnosed it as a network/install problem while the extension sat on disk.

**Understanding of the change:** it mirrors the VECTOR half of #2623 — probe the index catalog first (readable without any extension), load FTS with the analyze policy only when an index actually gates DML, and when it cannot load, fall through to the pre-existing wipe-and-bulk-COPY escalation. That is the only available remedy: LadybugDB exposes no extension-independent way to drop an FTS index (`CALL DROP_FTS_INDEX` lives in the extension; `DROP INDEX` is a parser error), so rebuilding the file is the only way to clear them.

**Repository history considered:** #2589/PR #2598 (drop-before-delete ordering — and the reason the reporter's crash moved from `copyNodeCSVs` on 1.6.9 to `deleteNodesForFiles` on rc.168), #2623/PR #2624 (the VECTOR gate this mirrors), #2409/PR #2425 (the escalation valve + `wipeLbugDbFiles`), #2658/PR #2677 (FTS build-failure classification + single-writer lock), #2374/PRs #2375/#2383 (extension-load diagnostics + `redactPaths`), #2798 (`SCHEMA_FINGERPRINT`). None is re-opened by this PR.

**Back-and-forth avoided by verifying rather than asking:** gate placement relative to every DML statement on the path; `isBenignDropFtsIndexError` byte-identity against base; the `--drop-embeddings` → `force` conversion being nested inside the `embeddingCheckpoint` branch; the crash-recovery ordering (dirty stamp before wipe); `FTS_INDEXES.length == 20` against a live catalog; whether the new throw has any reachable production caller; and whether CI actually runs the new suite with `GITNEXUS_REQUIRE_FTS=1`.

**Open questions for the author:** only one — is the `TESTING.md:68` sequential-project rule (§6.B) meant to bind here, given both existing siblings already diverge from it?

---

## 1. Engine breakdown and leg liveness

**Engine A (Claude)** — 14 lanes, all live and non-empty:
- swarm: `risk`, `test-ci`, `security`, `docs-dod`, `hygiene`, `facts`
- ce personas: `correctness`, `adversarial`, `testing`, `maintainability`, `project-standards`, `performance`, `api-contract`, `reliability`

`ce-correctness` additionally ran a **live read-only probe** against a copy of this repo's own 391 MB `.gitnexus/lbug` with the FTS extension unloaded — the only dynamic evidence in the whole review. `ce-testing` executed `drop-fts-index-error-classification.test.ts` (8/8 pass) and the VECTOR-gate block of `lbug-delete-nodes-for-files.test.ts` (4/4 pass); the four new `runFullAnalysis` cases could not run (no `dist/parse-worker.js` under the no-build constraint).

**Engine B (Codex, gpt-5.6-sol)** — all three legs live, tails confirmed:
| Leg | File | Tail | Notes |
|---|---|---|---|
| B1 swarm | `codex-swarm.md` | `B1_EXIT=0` | solo re-run after the org rate limit; 6 lanes, 3 findings, verdict "request changes" |
| B2 ce | `codex-ce.md` | `B2_EXIT=0` | ran through ~30 `Reconnecting…` events; 1 finding at confidence 100 |
| B3 adversarial | `codex-adversarial.md` | `B3_EXIT=0` | solo re-run; ran `npx tsc --noEmit` (exit 0) and `git diff --check`; 3 findings, 1 explicit refutation |

`prior-context.md` was used only to flag recurrence; the bullet marked STALE (obs #48860, "0 FTS indexes") was ignored, and obs #48909's *predicted* VECTOR-coverage gap is stale — that test exists.

---

## 2. What is solid — validated correctness, not absence of findings

Both engines independently confirmed, and synthesis re-read each:

- **The gate is correctly placed before ALL incremental DML.** Everything between `initLbug(buildPath)` (`run-analyze.ts:1869`) and the gate (`:2136`) is a read, a `CHECKPOINT`, an in-memory computation, or a JSON write. The first table DML on this path is `dropSearchFTSIndexes()` at `:2245`, downstream of the gate and only on the non-escalated branch. (A-adversarial refutation 2, A-risk, B1.)
- **The central premise holds, measured not assumed.** On pinned `@ladybugdb/core@0.18.3` with FTS unloaded, `CALL SHOW_INDEXES()` returns 53 rows: `{FTS: 20, HASH: 33}`, every FTS row carrying `extension_loaded: false`, and every non-FTS index named `_PK`. 20 == `FTS_INDEXES.length`. The gate really can see FTS in the #2841 state.
- **#2589 drop-before-delete ordering preserved.** `dropSearchFTSIndexes()` stays at `run-analyze.ts:2245` in the `else` branch; `incremental-fts-drop-ordering.test.ts` has zero diff.
- **`isBenignDropFtsIndexError` is byte-identical** (`lbug-adapter.ts:3338-3339`, diffed against base) and the change **narrows** what is silently tolerated — prior-context obs #48873's concern is not made worse.
- **No new security surface.** `readIndexCatalogRows` runs a fixed string; `ftsIndexExistsInCatalog` compares in JS and issues no query; the new throw builds a *fresh* `Error` rather than embedding LadybugDB's path-bearing text, so the `redactPaths()` requirement from #2374 correctly does not apply. No bidi/zero-width characters; non-ASCII in added lines is 25 em-dashes.
- **Crash recovery still works.** `saveIncrementalDirtyState('escalated-full-write', …)` at `:2207-2213` is written *before* the stop/close/wipe sequence; the next run reads `incrementalInProgress` at `:1193` → `force: true` → sidecar quarantine → full rebuild.
- **Self-healing is real and bounded.** The escalated rebuild produces a DB with no FTS indexes, so the *next* run's catalog read finds nothing to gate on and the surgical plan resumes. The escalation is one-shot per environment change, not a permanent regression.
- **Machines that never built FTS pay nothing at the gate** — `lbug-adapter.ts:3154-3156` returns `true` without touching the extension machinery. (Caveat: they *do* pay elsewhere — see H4.)
- **Backward-compatible API.** `ensureEmbeddingRowDmlSafe(indexRows?)` is an additive optional parameter; all five existing zero-arg call sites reproduce the pre-PR behaviour exactly. No packaged/programmatic surface is touched (`gitnexus/package.json` is `bin`-only).
- **Standards:** no schema/fingerprint bump owed (`SCHEMA_FINGERPRINT` is auto-derived from untouched DDL); `CHANGELOG.md` correctly untouched per DoD §4.5; prettier clean; PR title conventional; no new `GITNEXUS_*` env var.

---

## 3. Headline findings

### H1 — P1 · `analyze --drop-embeddings` is silently reverted by the widened rescue
**Blocks merge: yes.**
**Corroboration: STRONG (both engines).** Codex B2 (P2, confidence 100), Codex B3 (medium), Claude `ce-adversarial` F1 (P1, confidence 90).

**Severity adjudication (P1 adopted over Codex's P2):** the engines differ on the label but agree on the consequence — Codex B2's own verdict is *"Ready with fixes. Fix #1 and add its regression test before merging."* P1 is adopted because the run **completes with exit 0** while silently retaining data the operator explicitly asked to destroy, no assertion anywhere in the suite would catch it, and the repo pins the opposite contract in `test/unit/run-analyze.test.ts:976`. Silent, successful, contract-violating data retention is the P1 bar.

**Trigger:** a repo analyzed once with FTS available (so the DB carries all 20 FTS indexes), holding embeddings but **no `embeddingCheckpoint`**; the machine's FTS extension becomes unloadable; VECTOR is fine. Commit one file, run `gitnexus analyze --drop-embeddings`.

**Evidence** `[code-read]`:
- `run-analyze.ts:1177-1180` — the `options.force = true` conversion sits **inside** `if (existingMeta?.embeddingCheckpoint)` at `:1158`. Without a checkpoint the run stays incremental (`isIncremental` gates purely on `!options.force`) and reaches the gate.
- `run-analyze.ts:1527` — "`--drop-embeddings` → skip cache load entirely" — the flag *deliberately* leaves `cachedEmbeddings` empty. `:1545` logs `Dropping N existing embeddings`.
- `run-analyze.ts:2139-2140` — the rescue trigger, widened by this PR from `!embeddingRowDmlSafe` to `extensionForcedRebuild`, fires on `cachedEmbeddings.length === 0` — *exactly the state the user asked for*.
- `run-analyze.ts:2150-2157` — `loadCachedEmbeddings()` re-reads all N rows and overwrites both `cachedEmbeddings` and `cachedEmbeddingNodeIds`, logging `Preserving N embedding row(s) … (the index metadata did not account for them)` — flatly contradicting the `Dropping N` line and asserting a cause that is false.
- Phase 3.5 (`:2490-2496`) restores every row; Phase 5 stamps `stats.embeddings = N`. Exit 0.
- Contract pinned in-repo at `test/unit/run-analyze.test.ts:976` — "`--drop-embeddings` is the explicit wipe path; it must suppress cache load".

**Why it is a PR-2854 finding:** the block is pre-existing, but its old trigger required an HNSW index *and* a broken VECTOR extension. Retargeting it at `extensionForcedRebuild` moves it to the dominant path, because FTS indexes exist on **every** GitNexus-analyzed DB.

**Variant:** `--drop-embeddings --embeddings` is also defeated — the rescue repopulates `cachedEmbeddingNodeIds`, which `:2723` feeds to `runEmbeddingPipeline` as the already-embedded set, so the pipeline skips exactly the nodes the user asked to regenerate.

**Fix:** `if (extensionForcedRebuild && shouldLoadCache && cachedEmbeddings.length === 0)` (or `&& !options.dropEmbeddings`). Add an FTS-blocked + `dropEmbeddings: true` case asserting `stats.embeddings === 0`.

---

### H2 — P2 · The extension-forced escalation deletes the LIVE index in place on the default path
**Blocks merge: no** (availability window, recovery contract verified sound).
**Corroboration: STRONG (both engines).** Codex B3 `[high]`, Claude `ce-adversarial` F2 (P1, confidence 80), Claude `ce-correctness` R3, Claude `ce-reliability` RR1 (which verified the recovery contract).

**Trigger:** any machine where the FTS extension stops loading, on any previously-analyzed repo, on the very next `gitnexus analyze`. No flags. This is the #2841 reporter's exact machine.

**Evidence** `[code-read]`:
- `run-analyze.ts:1762-1780` — `wantAtomicIncremental` requires `GITNEXUS_ATOMIC_INCREMENTAL === '1'` (off by default), so `buildPath === lbugPath`. The *where to build* decision is frozen ~380 lines before the strategy decision and cannot be revisited at the gate.
- `run-analyze.ts:2161` (in-diff) — `extensionForcedRebuild ||` newly routes this population into the valve.
- `run-analyze.ts:2222-2226` — `stop() → closeLbug() → wipeLbugDbFiles(buildPath) → initLbug(buildPath)`, then a multi-minute full-graph COPY at `:2227`.
- On `main` the same condition failed at LadybugDB **bind time**, before a single row was mutated, leaving a fully intact index. After this PR the same condition *deletes* it first.

**Synthesis adjudication (Engine A disagrees with itself here — recorded, not hidden):** `swarm-risk` and `ce-reliability` traced the recovery path and found it sound — the `escalated-full-write` dirty stamp at `:2207-2213` precedes the wipe, so a crash in the window forces a clean full rebuild on the next run. That is correct, and it is why this is **P2, not P1**: this is an *availability* window (no queryable index until the rebuild lands, and the work must be redone), not silent data loss. `ce-adversarial` refutation 3 further confirms that `GITNEXUS_ATOMIC_INCREMENTAL=1` is strictly safer (the wipe destroys the staging copy) — the mitigation exists but cannot be reached from the gate.

**Fix:** defer the `buildPath`/`useAtomicSwap` resolution until after the extension gate so an extension-forced rebuild can stage-and-swap like a full rebuild; or, minimally, say in the escalation log that the live index is about to be deleted and rebuilt.

---

### H3 — P2 · `ftsIndexExistsInCatalog` fails OPEN, contradicting the contract this PR introduces two functions above
**Blocks merge: no** (defence-in-depth weakness; no reachable production caller today).
**Corroboration: STRONG (both engines).** Codex B1 (P2 — its stated basis for requesting changes), Claude `ce-maintainability` (P2, confidence 68), `ce-testing` finding 4, `ce-correctness` T4, `swarm-facts` ("notable asymmetry the PR body does not call out"), `swarm-risk` P3-2, `swarm-test-ci` Q4.

**Trigger:** `dropFTSIndex` is called with the FTS extension unloaded **and** a live index present **and** `CALL SHOW_INDEXES()` throwing at that moment.

**Evidence** `[code-read]`:
- `lbug-adapter.ts:3048-3049` — "`undefined` means 'could not prove anything' and **every caller must treat it as fail-closed** (assume an index may be present), never as 'no indexes'."
- `lbug-adapter.ts:3117` and `:3155` honour it (`rows === undefined || rows.some(...)`).
- `lbug-adapter.ts:3350-3353` does the opposite: `const rows = await readIndexCatalogRows(); if (rows === undefined) return false;` — an unreadable catalog is reported as "index absent", so `dropFTSIndex` (`:3387-3398`) tolerates the error and returns normally, and the caller proceeds as if the index were gone — reintroducing the exact #2841 symptom the guard exists to make loud.

**Adjudication of the severity split:** Codex B1 called this merge-blocking. Four Claude lanes independently established that the new throw has **no reachable production caller today** (`ce-adversarial` refutation 4, `ce-correctness`, `swarm-risk`, `ce-api-contract` enumerated all three call sites). So this is a weakness in a defence-in-depth guard, not a live bug — **P2, not merge-blocking**. It is still worth fixing precisely because defence-in-depth is the guard's only job.

**Fix:** `if (rows === undefined) return true;` (accept a false-positive loud error over a false-negative silent swallow), or state explicitly in the docstring that this is a deliberate exception to the fail-closed contract.

---

### H4 — P2 · The new liveness probe costs 20 extra `SHOW_INDEXES` reads on every incremental run on an FTS-less machine
**Blocks merge: no** (cost, not correctness — but it contradicts the docstring shipped with it).
**Corroboration: MODERATE (Engine A only — but two lanes reached it independently, plus a third as a residual).** `ce-performance` PERF-1 (P2, confidence 75), `ce-reliability` R1 (P2, confidence 78), `ce-correctness` R5.

**Trigger:** a machine where the FTS extension never loads (`GITNEXUS_LBUG_EXTENSION_INSTALL=never`/`load-only`, offline `auto`, or a missing runtime dependency) analyzing a DB that carries **no** FTS index — i.e. the configuration the PR's own test at `incremental-index-extension-dml-gate.test.ts:193-202` explicitly wants to keep cheap.

**Evidence** `[code-read]`:
- `ensureFtsRowDmlSafe` correctly returns `true` early *without* loading the extension (`lbug-adapter.ts:3154-3156`), so the connection has no FTS extension.
- The surgical branch then runs `dropSearchFTSIndexes()` (`run-analyze.ts:2245`), which loops over all **20** `FTS_INDEXES` entries (`fts-indexes.ts:198-202`; count verified = 20).
- Each `CALL DROP_FTS_INDEX` raises `Catalog exception: function DROP_FTS_INDEX is not defined`, which now triggers a **fresh** `readIndexCatalogRows()` per table (`lbug-adapter.ts:3350`), each under `withConnLock`.
- This contradicts the new docstring at `lbug-adapter.ts:3371-3372`: *"the liveness question is settled with a catalog read on the ERROR path only (the healthy path still costs nothing)"* — on this machine the error path **is** the healthy path, and it does not self-heal, because there is nothing to heal.

**Fix:** read the catalog once per `dropSearchFTSIndexes()` invocation (or thread through the `indexCatalogRows` snapshot already read at `run-analyze.ts:2136`) and check membership in an in-memory `Set`. Better still: skip the drop loop entirely when the gate already proved no FTS-typed index exists.

---

### H5 — P2 · The escalation message asserts facts the code explicitly could not prove, and its FTS clause has a dangling antecedent
**Blocks merge: no** (diagnostic honesty; the mis-diagnosis class this PR exists to end).
**Corroboration: MODERATE (Engine A only — three lanes).** `ce-adversarial` F3 (P2, confidence 85), `ce-reliability` R3 (P2, confidence 55), `ce-maintainability` (P2, confidence 82, on the antecedent).

**Trigger:** `CALL SHOW_INDEXES()` throws for any reason at `lbug-adapter.ts:3058`. Both gates correctly fail closed; both extension loads fail on an extension-less machine; the run escalates and prints that a `CodeEmbedding` **vector index exists** and that the index **carries FTS search indexes** — neither of which was established, and on a repo that never enabled embeddings the first claim is affirmatively false.

**Evidence** `[code-read]`: `run-analyze.ts:2174-2190` renders a fail-closed sentinel as a positive assertion. The only truthful signal — `logger.warn(…'Could not read the LadybugDB index catalog…')` at `lbug-adapter.ts:3061-3066` — goes to the pino **stderr** stream, not the `onLog` callback the escalation message uses, so under `gitnexus serve` / the analyze worker the UI shows *only* the false claim. This is precisely the mis-diagnosis failure mode the PR exists to eliminate.

**Second defect, same lines:** when both causes fire, `blockedCauses.join('; and ')` produces "…the CodeEmbedding vector index exists … **and this index** carries FTS search indexes…". "This index" grammatically points at the vector index just named; the FTS indexes live on 20 other tables. Standalone it reads oddly too — an index does not carry indexes.

**Fix:** thread the catalog-read outcome to the escalation site and emit a distinct third cause ("the index catalog could not be read, so neither gate could be settled"); replace "this index" with a self-contained subject ("the graph store carries one or more FTS search indexes…"); mirror the catalog-read warning into `log()` so IPC consumers see it.

---

### H6 — P2 · The `both blocked` test hard-asserts VECTOR availability with no probe or visible skip
**Blocks merge: maybe** — a red on any FTS-yes/VECTOR-no host, and DoD §2.7 forbids non-deterministic tests.
**Corroboration: MODERATE (one Claude lane + coordinator worktree verification + synthesis re-read).** `ce-project-standards` F2 (P2, confidence 68).

**Trigger:** any host where FTS is seeded but VECTOR is not (offline dev box; any lane not running the "Ensure FTS + VECTOR extensions installed" step).

**Evidence** `[code-read]`: `incremental-index-extension-dml-gate.test.ts:243` — `expect(vectorIndexBuilt).toBe(true);`. `createVectorIndex()` succeeds only when the **VECTOR** extension loads, which is host- and network-dependent through `resolveAnalyzeInstallPolicy()`. The suite's isolation is FTS-only: the `beforeAll` probe at `:77-91` settles `ftsAvailable` alone, and `beforeEach` at `:95-110` skips on `!ftsAvailable`. Nothing probes VECTOR, so the case reds with `expected false to be true` — a failure that says nothing about #2841. Violates DoD §2.7 ("New tests are deterministic"), and the sibling `incremental-vector-extension-ordering.test.ts:44-77` implements exactly the right shape (probe → `GITNEXUS_REQUIRE_VECTOR` hard-fail → visible `ctx.skip()`).

**Fix:** extend the `beforeAll` probe to settle `vectorAvailable` and gate *only* this `it` on it. The other three cases need no VECTOR and must keep running.

---

### H7 — P2 · The FTS gate's catalog-unreadable branch has no test, although its VECTOR twin has exactly that test
**Blocks merge: maybe** — the untested branch is the fail-closed guarantee itself.
**Corroboration: STRONG (both engines).** Codex B1 (P3, "add a regression case where the post-drop catalog read fails"), Claude `ce-testing` finding 1 (P1, confidence 75), `ce-correctness` T2/T4, `ce-reliability` TG1, `swarm-test-ci` Q4, `ce-maintainability` and `ce-project-standards` testing gaps.

**Evidence** `[code-read]`: `lbug-adapter.ts:3154-3155` — `rows === undefined || rows.some(...)` is the fail-closed guarantee for FTS. None of the four new cases forces `SHOW_INDEXES()` to throw; all four go through a successful catalog read. The VECTOR twin's identical branch **is** covered, at `test/integration/lbug-delete-nodes-for-files.test.ts:417` ("catalog read fails → falls back to attempting the extension load (fail-safe)"), using a ready-made query-interception technique at `:429-441` that rejects only on SQL containing `SHOW_INDEXES`. A future "simplification" to `rows?.some(...) ?? false` would flip FTS from fail-closed to fail-open — reproducing #2841 — and no test in this diff would catch it.

**Fix:** add an FTS sibling to that same `describe` block (cheapest — the file is already in both the `lbug-db` project and the cross-platform list), reusing the `vi.spyOn(lbug.Connection.prototype, 'query')` technique.

---

### H8 — P3 · `dropSearchFTSIndexes`' docstring now overstates the guarantee `dropFTSIndex` gives
**Blocks merge: no.**
**Corroboration: STRONG (both engines).** Codex B1 lane 6 (P3, confidence 0.99), Claude `ce-api-contract` finding 1 (P2, confidence 55).

**Evidence** `[code-read]`: `fts-indexes.ts:190-191` still reads *"Drop every configured FTS index (no-op per index when absent or unloadable — `dropFTSIndex` tolerates both)."* After this PR, "unloadable" is no longer unconditionally tolerated — only "unloadable **and** actually absent" is (`lbug-adapter.ts:3387-3398`). Safety is now enforced entirely by caller *ordering discipline* (`ensureFtsRowDmlSafe` must have run first), not by the type system, and this stale comment is the exact artifact that would mislead the next caller into skipping that ordering.

**Fix:** update the docstring to state the real contract: no-op when absent; **throws** when the index is live but the extension cannot load; callers must prove `ensureFtsRowDmlSafe`/`loadFTSExtension` succeeded first.

---

## 4. POSTABLE AS INLINE COMMENTS

Every row below carries trigger + file:line evidence + tag, and anchors on a line that appears on the **RIGHT** side of the PR diff. In-diff RIGHT-side ranges used: `run-analyze.ts` 35-36, 2127-2140, 2161, 2169-2190, 2192-2196 · `lbug-adapter.ts` 3029-3069, 3103-3105, 3112, 3117-3118, 3127-3159, 3341-3359, 3364-3373, 3387-3398 · `incremental-index-extension-dml-gate.test.ts` 1-270 · `drop-fts-index-error-classification.test.ts` 13, 16-20, 72-156 · `fts-extension-e2e.test.ts` 296-345.

| # | Sev | Path | Line | Anchors the true location? | Tag |
|---|---|---|---|---|---|
| 1 | **P1** | `gitnexus/src/core/run-analyze.ts` | **2140** | yes — the rescue condition itself | `[code-read]` |
| 2 | **P2** | `gitnexus/src/core/run-analyze.ts` | **2161** | **no** — body must reference `:1762-1780` (`buildPath` frozen, `GITNEXUS_ATOMIC_INCREMENTAL` off by default) and `:2222-2226` (the wipe), both outside the diff | `[code-read]` |
| 3 | **P2** | `gitnexus/src/core/lbug/lbug-adapter.ts` | **3353** | yes — `if (rows === undefined) return false;` | `[code-read]` |
| 4 | **P2** | `gitnexus/src/core/lbug/lbug-adapter.ts` | **3350** | partly — body must reference `fts-indexes.ts:198-202` (the 20-iteration loop, outside the diff) and the "costs nothing" claim at `:3371-3372` | `[code-read]` |
| 5 | **P2** | `gitnexus/src/core/run-analyze.ts` | **2185** | yes — the FTS `blockedCauses.push`; covers both the overclaim and the dangling "this index" | `[code-read]` |
| 6 | **P2** | `gitnexus/test/unit/incremental-index-extension-dml-gate.test.ts` | **243** | yes — `expect(vectorIndexBuilt).toBe(true);`; body references the FTS-only probe at `:77-110` (also in-diff) | `[code-read]` |
| 7 | **P2** | `gitnexus/test/unit/incremental-index-extension-dml-gate.test.ts` | **224** | partly — body must point at the ready-made VECTOR twin at `test/integration/lbug-delete-nodes-for-files.test.ts:417` (technique at `:429-441`), outside the diff | `[code-read]` |
| 8 | **P3** | `gitnexus/src/core/lbug/lbug-adapter.ts` | **3373** | **no** — body must reference the stale docstring at `fts-indexes.ts:190-191`, outside the diff | `[code-read]` |

**Optional (post only if comment volume allows):**

| # | Sev | Path | Line | Note | Tag |
|---|---|---|---|---|---|
| 9 | P3 | `gitnexus/src/core/run-analyze.ts` | **2196** | same-commit recovery (§5.C); body must reference `:1430-1435` + `:1448-1512` (fast path) and `cli/analyze.ts:1646-1650` ("…then rerun"), both outside the diff | `[code-read]` |
| 10 | P3 | `gitnexus/src/core/lbug/lbug-adapter.ts` | **3153** | shared-snapshot invariant (§5.A); pairs with `run-analyze.ts:2136` (also in-diff) | `[code-read]` |

**Deliberately NOT postable inline:** the `index_type` polarity note (§6.A — adjudicated as latent, not live), the `ensureEmbeddingRowDmlSafe` precondition regression (§5.B — Engine A disagrees with itself and no production caller triggers it), the vitest-project routing question (§6.B — pre-existing convention drift), and everything in §5.D–F.

---

## 5. Lower-priority items

**A. P3 — the "one shared `SHOW_INDEXES` read" invariant is false exactly on the failure branch.** MODERATE (A only: `ce-correctness` F1, `ce-adversarial` F4 conf 95, `ce-reliability` R2 conf 82). `run-analyze.ts:2132-2135` promises one read; when it fails, `indexCatalogRows === undefined`, and `??` at `lbug-adapter.ts:3112`/`:3153` cannot distinguish "the caller tried and could not prove anything" from "the caller passed nothing" — so each gate re-reads. Three round-trips, three identical warnings, and the two gates can legitimately decide from *different* snapshots. Behaviourally benign today (the cross-process writer lock means nothing can mutate the catalog between reads, and every branch is fail-closed-or-better), but the load-bearing invariant is documented and not enforced. Cheapest correct fix: make `indexRows` a **required** parameter with no `??` fallback and keep a thin zero-arg wrapper for the existing callers.

**B. P3 — `ensureEmbeddingRowDmlSafe` lost its unconditional null-connection precondition.** MODERATE (A only: `ce-correctness` F4 conf 45, `ce-adversarial` F6 conf 60) — and **`swarm-risk` disagrees**, calling the migration safe. *Adjudication:* both are right about different things. The `if (!conn) throw` now lives only inside `readIndexCatalogRows` (`:3052-3054`), reached only when `indexRows` is nullish, so `ensureEmbeddingRowDmlSafe(rows)` can now answer `true` for a closed connection where it previously threw. No production caller does this (`run-analyze.ts:2137` sources its rows on the same tick), so it is a contract regression on an exported API, not a live defect. Re-adding `if (!conn) throw` at the top of both gates is a one-liner.

**C. P3 — the CLI's "install … then rerun" advice cannot heal FTS on the same commit.** Codex B3 (medium) — and B3 explicitly *narrowed* its own hypothesis, which is worth crediting: `--repair-fts` and any later content change **do** heal it, so only the broad "never rebuilt" framing is wrong. Verified `[code-read]`: after the FTS-forced rebuild stamps `lastCommit`, a plain rerun on a clean tree at the same commit hits `run-analyze.ts:1430-1435`, and the fast path returns at `:1499-1512` — before Phase 3 ever tries `loadFTSExtension`. Meanwhile `cli/analyze.ts:1646-1650` (untouched by this PR) prints *"Install it once with network access … then rerun, or run `gitnexus analyze --repair-fts` when connected."* The advice is only half-right, and this PR newly makes that state common. Cheapest fix: bypass the fast path when meta records FTS unavailable and the runtime can now load it — or drop the plain-rerun half of the advice.

**D. P3 — the size cause is silently dropped when an extension cause co-occurs.** WEAK (`ce-adversarial` F7, conf 90). `run-analyze.ts:2160-2166` escalates on `extensionForcedRebuild || shouldEscalateIncrementalWrite(...)`, but `:2191-2204` is an either/or ternary, so the `effective write set covers X/Y files (Z%)` line is never emitted when both hold. The PR's own comment at `:2169-2171` argues that reporting one cause while another is equally live "is how #2841 stayed mis-diagnosed" — the same argument crosses the extension/size boundary. Append rather than branch.

**E. P3 — cross-platform registration omitted.** WEAK (`ce-project-standards` F1, conf 62). `scripts/cross-platform-tests.ts:156` registers the direct sibling `incremental-vector-extension-ordering.test.ts` with a rationale that transfers verbatim, and `:167-170` registers `fts-extension-e2e.test.ts` because "#2374 was Windows-reported". The new 270-line suite is registered nowhere, so the Windows/macOS matrix never runs the four orchestration cases. Compounding: `GITNEXUS_REQUIRE_VECTOR=1` is set **only** on the cross-platform job (`ci-tests.yml:211`), so the both-blocked case never runs under the strict-availability contract for its VECTOR half. Either register it, or say in the PR why the shard budget excludes it.

**F. P2/P3 — one-time full rebuild for a zero-change or one-file commit.** MODERATE (`ce-performance` PERF-2 conf 55, `ce-correctness` F2 conf 60). `extensionForcedRebuild` is OR'd *ahead* of the size-based gate, so a blocked extension bypasses `escalation-gate.ts`'s own ≥50%-write-fraction rationale entirely — including for `git commit --allow-empty` (which falls past `alreadyUpToDate` with `effectiveWriteSet.size === 0`). Both lanes agree the cost is **one-shot** and self-heals, and that the pre-existing VECTOR half already had this shape. Optional tightening: skip the gates when `filesToDelete.length === 0 && effectiveWriteSet.size === 0` (`[]` reads as "no index gates DML").

**G. P3 — hand-rolled remedy text bypasses the extension-error classifier.** WEAK (`ce-maintainability`, conf 65). The new throw at `lbug-adapter.ts:3391-3396` tells the user to try `GITNEXUS_LBUG_EXTENSION_INSTALL=auto`, but `extension-load-error.ts` exists precisely because that remedy is *wrong* for the missing-runtime-dependency class (Windows error 126 / OpenSSL) — which is why `--repair-fts` (`run-analyze.ts:1041-1064`) and `ftsDegradedWarning` both route through `diagnoseExtensionLoad`/`classifyExtensionLoadError` first.

**H. P3 nits.** New `no-non-null-assertion` warning at `drop-fts-index-error-classification.test.ts:132` (`probe!.dbPath`) — the only lint diagnostic attributable to the diff; CI runs `eslint .` with no `--max-warnings`, so it does not gate. `IndexCatalogRow` is exported with zero external consumers. `ensureEmbeddingRowDmlSafe`'s docstring at `:3095-3101` still re-explains catalog mechanics that moved to `readIndexCatalogRows`. The `indexRows` parameter and its freshness contract are undocumented on both exported gates. `GUARDRAILS.md:34` still describes "switching to a full DB write" as having exactly one trigger (write set >~50%) — though #2623 set that precedent without updating it either.

---

## 6. Engine disagreements — adjudicated

### A. FTS gate "fails open" on an unreadable `index_type`
- **Claude `ce-correctness` F3 (P3, confidence 40):** `ensureFtsRowDmlSafe` uses `(row?.index_type ?? row?.[2]) === 'FTS'` (`:3155`) while the VECTOR twin uses `!== 'HASH'` (`:3121`). An `index_type` that cannot be read yields `undefined`: for VECTOR `undefined !== 'HASH'` ⇒ gate ⇒ fail-closed; for FTS `undefined === 'FTS'` ⇒ **no gate** ⇒ fail-open. The two functions the docs present as symmetric have opposite failure polarity.
- **Codex B3:** refuted for pinned `@ladybugdb/core@0.18.3` — FTS is reported reliably, and *catalog-read* errors fail closed at the gate.

**Adjudication: both correct; they answer different questions, and there is no live defect.** Codex answers "does the catalog *read* failing fall open?" — no, `rows === undefined ||` handles that, on both twins. Claude answers "does a successfully-read row with an *unrecognised shape* fall open?" — yes, and only for FTS. Decisively, `ce-correctness`'s own live probe on 0.18.3 shows `index_type: 'FTS'` is reported with `extension_loaded: false`, so the shape is correct today and the scenario needs a future engine change. The probe also proves the positional fallback `row?.[2]` is **dead code** on 0.18.3 (rows are named records; `rows[0][0] === undefined`), so the backstop the comment at `:3030-3035` advertises does not exist at runtime. Verdict: **not a defect; a latent upgrade fragility.** Worth a `LADYBUGDB-CONTRACT` marker on the `'FTS'` literal (the file uses that convention at `:2258-2261`) and either a test pinning the column order or deletion of the dead positional fallbacks. Not postable as a finding.

*Related intra-engine refutation worth crediting:* `ce-adversarial` F5 (conf 50) hypothesised that the type-keyed gate and the name-keyed `ftsIndexExistsInCatalog` could disagree and fire the new throw. `ce-correctness`'s probe **refutes** it — every non-FTS index is named `_PK`, so a name collision is impossible today.

### B. Should the new suite live in the sequential `lbug-db` vitest project?
- **Codex B1 (P3):** `TESTING.md:68` requires it — *"When adding a new test that uses native LadybugDB (`@ladybugdb/core`), add it to the `lbug-db` project's explicit include list and the `default` project's exclude list."* The file appears in neither list.
- **Claude `swarm-test-ci`:** routing to `default` matches both existing siblings, and CI *does* run it — sharded `npx vitest run` executes the `default` project, and `GITNEXUS_REQUIRE_FTS=1` is set on three jobs (`ci-tests.yml:28`, `:94`, `:207`), so an unavailable extension fails loudly rather than skipping.

**Adjudication: both facts verified; this is pre-existing convention drift, not a PR defect.** `TESTING.md:68` says what B1 quotes. `vitest.config.ts` registers `incremental-dirty-recovery.test.ts` (`:102`, `:153`) but **not** `incremental-vector-extension-ordering.test.ts` or `incremental-fts-drop-ordering.test.ts`, despite both driving the same native engine. The PR follows the nearest local precedent while diverging from the written rule. Either fix is defensible and it is a maintainer call — not merge-blocking. The more actionable omission is the *cross-platform* list (§5.E), where the sibling **is** registered and the new file is not.

### C. Is H3 (fail-open drop guard) merge-blocking?
- **Codex B1:** yes — "Not production-ready".
- **Claude (four lanes):** the new throw has no reachable production caller, so the guard is defence-in-depth.

**Adjudication:** Claude's reachability analysis is the stronger evidence (all three call sites enumerated, each downstream of an extension-load gate). Keep as **P2, worth fixing, not merge-blocking**.

---

## 7. Test gaps (consolidated)

1. **No FTS-gate test for the catalog-unreadable branch** — H7. The VECTOR twin's test and technique are ready to copy (`lbug-delete-nodes-for-files.test.ts:417`, `:429-441`). *Both engines.*
2. **No test for `ftsIndexExistsInCatalog`'s `rows === undefined` branch** — the one place the new guard degrades back to silent tolerance (H3). *Both engines.*
3. **No `--drop-embeddings` + FTS-blocked case** — would have caught H1 directly, asserting `stats.embeddings === 0`. *Both engines.*
4. **No third-run assertion.** Both new regressions run exactly two `runFullAnalysis`/CLI invocations. Nothing proves run 3 returns to the surgical plan — the property that makes this a one-shot rebuild rather than a permanent regression, and the single most valuable missing assertion.
5. **No outcome assertion on the embedding rescue via the FTS-only trigger.** The `both blocked` case seeds real un-stamped embeddings but never asserts they survived (no `MATCH (e:CodeEmbedding)`, no `'Preserving'` log check) — unlike the VECTOR sibling at `incremental-vector-extension-ordering.test.ts:245-260`.
6. **The pre-existing "only VECTOR blocked" test may now be conflated.** `incremental-vector-extension-ordering.test.ts:232` sets `GITNEXUS_LBUG_EXTENSION_INSTALL='never'`, which this PR's shared call site now also applies to FTS; if run 1 built the `File` FTS index, run 2 is blocked for *two* reasons and the assertions (`.includes('full DB write')`, `.includes('VECTOR')`) cannot tell the difference. (`ce-testing` finding 2, confidence 55 — inferred, not executed.)
7. **The "never carried FTS indexes" case asserts only the absence of an escalation log** — it never verifies `touchAndCommit`'s marker reached the graph, unlike every sibling.
8. **Nothing pins the load-bearing property** that `index_type === 'FTS'` is readable while the extension is unloaded. Verified only by `ce-correctness`'s external probe. A pure-unit case feeding `ensureFtsRowDmlSafe` a literal catalog fixture needs no native connection and can never skip.
9. **No bound on catalog-read count per run** — H4 could regress further with no test failing.
10. **No abort-mid-escalation test** for the H2 window / the `escalated-full-write` dirty-flag recovery contract.

---

## 8. CI status and coverage limits

Re-checked live at synthesis time (`gh pr checks 2854`): **everything green except Vercel.** All previously-in-progress jobs have since landed — ubuntu coverage 1/3, 2/3, 3/3 pass; windows 1/3, 2/3, 3/3 pass; all three macOS shards pass; `Build & Push gitnexus` passes; coverage-merge passes; CodeQL (JS/TS + Python), Gitleaks, tree-sitter ABI (3 OS), packaged install smoke, node floor 22.18, benchmarks, eval jobs, quality/{format,lint,typecheck,typecheck-web} all pass. Codex B3 independently ran `npx tsc --noEmit` (exit 0) and `git diff --check` (clean).

**Vercel `fail` — "Authorization required to deploy"** is the known non-actionable fork-deploy authorization check, not a code signal. `mergeStateStatus=BLOCKED` reflects the missing required review, not a defect.

**Coverage limits on this review:**
- No builds were run, so the four new `runFullAnalysis` cases and the CLI e2e case were **not executed** by either engine — every claim about them is source inference.
- `test/integration/fts-extension-e2e.test.ts` fails 9/11 on **unmodified `main`** in this sandbox (seeded extension will not load inside the spawned CLI), so any red there must be baseline-compared before attribution.
- The GitNexus graph index predates the PR head by ~29 commits and does not contain the new symbols, so its LOW diff summary is not authoritative; source and test evidence were used instead. `dropFTSIndex` is HIGH blast radius by dependency policy.
- No PDG/taint layer is present, so the absence of security-graph findings is not proof of absence — the security lane's verdict rests on manual tracing of the three propagation surfaces.

---

*Automated multi-tool digest — merged from 14 Claude lanes and 3 Codex legs, with every postable claim re-verified against the worktree. Findings are evidence-tagged `[code-read]` (source re-read) rather than `[reproduced]`; nothing here was executed end-to-end. Verify before acting.*

gitnexus/test/unit/incremental-index-extension-dml-gate.test.ts
```diff
@@ -0,0 +1,270 @@
...
+      const { lbugPath } = getStoragePaths(repo.dbPath);
+
+      // POSIX literal for the graph-side path: filePaths are stored with
+      // forward slashes on every OS (see the note in the #2623 suite).
+      const seeded = await seedEmbeddingsForFiles(repo.dbPath, ['src/handler.ts'], 2);
+      expect((seeded.get('src/handler.ts') ?? []).length).toBeGreaterThan(0);
+      await lbugAdapter.initLbug(lbugPath);
+      const vectorIndexBuilt = await lbugAdapter.createVectorIndex();
+      await lbugAdapter.closeLbug();
+      expect(vectorIndexBuilt).toBe(true);
```
magyargergo, line 338: **P2 · Non-deterministic: this hard-asserts VECTOR availability, but the suite only probes FTS.** Blocks merge: maybe — it reds on any FTS-yes/VECTOR-no host, and DoD §2.7 requires new tests to be deterministic. `[code-read]`

`createVectorIndex()` succeeds only when the **VECTOR** extension loads, which is host- and network-dependent through `resolveAnalyzeInstallPolicy()`. This suite's isolation is FTS-only: the `beforeAll` probe at `:77-91` settles `ftsAvailable` alone, and `beforeEach` at `:95-110` skips on `!ftsAvailable`. Nothing probes VECTOR — so on a host with FTS seeded and VECTOR absent this case runs and fails with `expected false to be true`, a red that says nothing about #2841.

The direct sibling gets this right: `test/unit/incremental-vector-extension-ordering.test.ts:44-77` probes the extension it depends on, hard-fails under `GITNEXUS_REQUIRE_VECTOR=1`, and otherwise calls `ctx.skip()` visibly.

Compounding it, `GITNEXUS_REQUIRE_VECTOR=1` is set **only** on the cross-platform job (`.github/workflows/ci-tests.yml:211`), and this file is not registered in `scripts/cross-platform-tests.ts` (its sibling is, at `:156`) — so the VECTOR half never runs under the strict-availability contract anywhere.

**Fix:** extend the `beforeAll` probe to settle `vectorAvailable` and gate **only this `it`** on it; the other three cases need no VECTOR and must keep running.

gitnexus/test/unit/incremental-index-extension-dml-gate.test.ts
```diff
@@ -0,0 +1,270 @@
...
+          { onProgress: () => {}, onLog: (m: string) => logs.push(m) },
+        ),
+      ).resolves.toBeDefined();
+      expect(logs.some((m) => m.includes('full DB write'))).toBe(false);
+    } finally {
+      if (previousPolicy === undefined) delete process.env.GITNEXUS_LBUG_EXTENSION_INSTALL;
+      else process.env.GITNEXUS_LBUG_EXTENSION_INSTALL = previousPolicy;
+      await repo.cleanup();
+    }
+  }, 300_000);
```
magyargergo, line 312: **P2 · The FTS gate's catalog-unreadable branch has no test — though its VECTOR twin has exactly that test.** Blocks merge: maybe — the untested branch *is* the fail-closed guarantee. `[code-read]` — **both engines** (Codex swarm asked for this regression case; Claude testing/correctness/reliability/test-ci agree).

`ensureFtsRowDmlSafe`'s safety rests on `rows === undefined || rows.some(...)` (`lbug-adapter.ts:3154-3155`). None of the four new cases forces `SHOW_INDEXES()` to throw — all four take a successful catalog read. A future "simplification" to `rows?.some(...) ?? false` would silently flip FTS from fail-closed to fail-open, reproducing #2841, and nothing in this diff would catch it.

**The technique is ready to copy** (outside this diff): `test/integration/lbug-delete-nodes-for-files.test.ts:417` — *"catalog read fails → falls back to attempting the extension load (fail-safe)"* — using a `vi.spyOn(lbug.Connection.prototype, 'query')` interception at `:429-441` that rejects only on SQL containing `SHOW_INDEXES`. Adding an FTS sibling in that same `describe` is the cheapest fix: that file is already registered in both the `lbug-db` vitest project and the cross-platform list.

Same gap applies to `ftsIndexExistsInCatalog`'s `rows === undefined` branch — the one place the new guard degrades back to silent tolerance.

gitnexus/src/core/lbug/lbug-adapter.ts
```diff
@@ -3279,10 +3338,39 @@ export const queryFTS = async (
...
+ *
+ * One benign class is conditional (#2841): `Catalog exception: function
+ * DROP_FTS_INDEX is not defined` says the FTS extension is not loaded, which
+ * is "nothing to drop" only when the named index does not exist. When it DOES
+ * exist, swallowing that error reports a drop that never happened, and the
+ * next insert/delete against that table dies at bind time with an engine
+ * message that never mentions FTS — the #2841 crash. So the liveness question
+ * is settled with a catalog read on the ERROR path only (the healthy path
+ * still costs nothing) and a live-but-undroppable index is raised loudly,
+ * naming FTS and both remedies.
```
magyargergo: **P3 · This docstring's guarantee is now contradicted by its caller-facing sibling.** Blocks merge: no. `[code-read]` — **both engines** (Codex swarm conf-0.99, Claude api-contract).

`dropSearchFTSIndexes`' docstring at `gitnexus/src/core/search/fts-indexes.ts:190-191` (outside this diff) still reads: *"Drop every configured FTS index (no-op per index when absent or unloadable — `dropFTSIndex` tolerates both)."*

After this PR, "unloadable" is no longer unconditionally tolerated — only *unloadable **and** actually absent* is (`:3387-3398`). Safety now rests entirely on caller **ordering discipline** (`ensureFtsRowDmlSafe` must have run first), not on anything the type system enforces — and that stale comment is precisely the artifact that would talk the next caller out of the ordering that makes today's usage safe.

**Fix:** update it to the real contract — no-op when absent; **throws** when the index is live but the extension cannot load; callers must prove `ensureFtsRowDmlSafe`/`loadFTSExtension` succeeded first.

PR #2856, opened 2026-08-06
feat: close reported graph blind spots in reference resolution, analyze and storage

ReidenXerx:
## Summary

Closes eighteen blind spots where the graph returned a **confident empty answer** rather than an unknown one: plain-object property access, module-scope `const` references (same-file and across imports), TypeScript type-alias and interface members, `impact` reporting `risk: LOW` on an empty caller set, a refresh that reports success while leaving the index unusable, a rebuild that failed with a raw engine error, execution flows that never exceeded a handful of steps, a flow that could only end at a leaf rather than where the program actually reaches outward, and `route_map` reporting "No routes found in this project" for a server that routes without a framework.

This is **three rounds of work in one PR.** The reporter re-ran every item against this branch after each round and filed a follow-up, which is where the R2 and R3 items come from.

Each round found the same defect wearing a different costume — a confident empty answer — and each round's fix exposed the next one:

- **Round 2** — the round-1 fixes were correct and still left the headline question unanswerable, because minting the missing nodes exposed a *second* failure the first had been masking.
- **Round 3** — verified **out-of-sample** for the first time, against six field names appearing in no prior report. Round 2 scored **5/5 in-sample and 0/6 out-of-sample**, and that gap is a finding in its own right: the in-sample set was, by accident, exactly the subset that happened to have a same-language anchor. The fixes were real; the examples were unrepresentative.

## Motivation / context

These come from an upstream bug report written against a production JS/TS repo, backed by 93 agent-written fallback reports over 2026-07-16 → 08-04 across 47 indexed commits. Every item below was reproduced locally before being changed.

The report's own summary is the through-line, and it is the right one:

> every critical item is a case of the tool returning a **confident, well-formed, wrong-because-empty answer** — `[]`, `incoming:{}`, `risk: LOW`. The graph's coverage limits are acceptable and expected; **presenting a coverage limit as a factual negative is not.**

Two reported items needed **no code**: the DI/closure caller gap is already fixed on `main` by `callable-value-flow` / `property-dispatch` (unreleased — the reporter was on `1.6.9`, tagged before that work), and the nested-skill-directory bug was fixed by #2434. Both were verified, not assumed.

## Areas touched

- [x] `gitnexus/` (CLI / core / MCP server)
- [x] `gitnexus-web/` (Vite / React UI) — one file, `src/core/llm/tools.ts`, so the UI's prose stops saying a symbol "appears to be unused" on an `UNKNOWN` verdict
- [ ] `.github/` (workflows, actions)
[...]

PR #2855, opened 2026-08-06
fix(resolution): resolve calls through a generic-typed field receiver in every language (#2833)

magyargergo:
Closes #2833.

A field whose declared type carries a type argument (`repo: Repo<User>`) emitted **zero** CALLS edges. Not a truncated chain, not an edge to the interface declaration — nothing, not even the primary edge. Every call through a generic repository, handler or service field was invisible to `impact()`, and the resulting zero was indistinguishable from a real one.

## The bug is an asymmetry, not generics

A field receiver is spelled `this.repo` — dotted — so it types through the receiver-chain fold and the text cascade, both of which reach `findClassBindingInScope`, which has no notion of type arguments. A local or a parameter of the **identical** type is a bare name, so it reaches Case 4 and `resolveClassBindingForName`, which does understand them.

One TypeScript file makes it plain, and it is now a test row:

```ts
class Svc {
  private field: Repo<User>;
  viaField(u: User)  { this.field.save(u); }                          // 0 edges
  viaLocal(u: User)  { const l: Repo<User> = this.field; l.save(u); } // 2 edges
  viaParam(p: Repo<User>, u: User) { p.save(u); }                     // 2 edges
}
```

Same type, same class, same file. Only the field lost everything. This is the same shape as #2813/#2829 (Case 0 lacked the fan-out Case 4 had) and #2832/#2842 (Case 3b lacked it too): a property of how the receiver is *spelled* rather than of what it resolves to.

## Scope was decided by measurement, not by plausibility

The issue guessed that Java, C# and Kotlin "plausibly share it". Measured, per language, on real pipeline runs:

| affected | already correct |
| --- | --- |
| TypeScript, C#, C++, Python | Java, Kotlin, Go, Rust, Swift, Dart |

The six that were already correct erase type arguments at **interpret** time (`java/interpret.ts` runs `stripGeneric` over the annotation, F41 #1928; Swift likewise), so their `rawName` is already `Repo`. TypeScript, C# and Python do not: their `stripGeneric` is a container **allow-list** that returns the type *argument*, so a user-defined `Repo<User>` matches nothing in it and the literal spelling survives into a lookup that binds nothing.

Two of the four needed more than the shared fix, and both reasons were found by bisecting rather than by guessing:

[...]

PR #2840, opened 2026-08-05
feat(objective-c): add scope-aware code intelligence for iOS codebases

davien:
## Summary

Adds production-usable Objective-C code intelligence for iOS codebases: vendored Tree-sitter grammar support, parsing and extraction, scope-aware call and inheritance resolution, graph querying, UI language metadata, documentation, and regression coverage.

This is intentionally described as Objective-C support, not complete Swift or Xcode project-model support.

## Motivation / context

GitNexus currently cannot model real Objective-C iOS repositories with sufficient fidelity for navigation, call graphs, inheritance, impact analysis, and execution-flow discovery. This change has been validated against a real iOS project and is intended to make Objective-C repositories usable end to end.

Related work: #1882 proposes Objective-C ingestion phase 1. This PR provides a broader, scope-aware implementation and may be reviewed as a comprehensive follow-up or superset; maintainers can choose the best integration path.

## Areas touched

- [x] `gitnexus/` (CLI / core / MCP server)
- [x] `gitnexus-web/` (Vite / React UI)
- [x] `.github/` (workflows, actions)
- [x] `eval/` or other tooling
- [x] Docs / agent config

## Scope & constraints

**In scope**

- Objective-C headers and implementations
- Classes, categories, protocols, properties, methods, functions, imports, calls, inheritance, and protocol edges
- Receiver-aware Objective-C message dispatch through the shared scope-resolution pipeline
- Optional grammar installation and prebuild behavior across supported platforms
- Objective-C unit, integration, regression, and real-project validation

**Explicitly out of scope / not done here**

- Full Swift semantic support
- Xcode build-setting evaluation
- Bridging-header and module-map semantics
- Compiler-grade macro expansion beyond parser-visible facts

## Implementation notes

- Language-specific behavior is isolated behind the existing language-provider and scope-resolver extension points; shared ingestion code does not add Objective-C-specific branching.
- The Objective-C grammar follows the repository's existing optional vendored-grammar lifecycle.
- Existing-language behavior is guarded by the full regression suite and targeted cross-language scope-resolution tests.

## Testing & verification

Validated on Node.js 22.18.0:

- [x] Objective-C targeted suite: **17 files, 184/184 tests passed**
- [x] Changed-test suite: **46 files, 688/688 tests passed**
- [x] `cd gitnexus-web && npm test`: **34 files, 391/391 tests passed**
- [x] `cd gitnexus && npx tsc --noEmit`
- [x] `cd gitnexus-web && npx tsc -b --noEmit`
[...]

PR #2857, opened 2026-08-06
fix(spring): extract method-level RequestMapping routes

### LINKED ISSUE #2824: [Bug] Spring route extraction: method-level @RequestMapping produces no Route nodes
### Area

### Summary

Method-level `@RequestMapping` is never emitted as a `Route` node, so controllers written in the pre-Spring-4.3 style (no `@GetMapping`/`@PostMapping` shortcuts) produce **zero** routes — including the `method = RequestMethod.X` form, whose verb *is* statically known.

### Context

`METHOD_ANNOTATION_TO_HTTP` in `core/ingestion/route-extractors/spring-shared.ts` maps only the five shortcut annotations:

```js
export const METHOD_ANNOTATION_TO_HTTP = {
  GetMapping: 'GET', PostMapping: 'POST', PutMapping: 'PUT',
  DeleteMapping: 'DELETE', PatchMapping: 'PATCH',
};
```

`@RequestMapping` is deliberately absent, and the doc comment explains why:

> `@RequestMapping` is intentionally absent: on a method it carries no implicit verb (the verb lives in its `method = RequestMethod.X` attribute), and on a class it is a URL prefix rather than a route. Callers handle `@RequestMapping` separately.

The class-prefix half is handled. The method half is not — in `route-extractors/spring.ts` Phase 2 the annotation is dropped outright:

```js
const httpMethod = METHOD_ANNOTATION_TO_HTTP[ann];
if (!httpMethod) continue; // skip @RequestMapping on methods (ambiguous verb)
```

So the "callers handle it separately" contract is unmet on the method path. The same shape exists in the group layer (`group/extractors/http-patterns/java.ts`), which gates on the identical table — so this is a **both-layer** gap, not an ingestion/group parity drift like #2280 / #2287 / #2288.

Two distinct sub-cases collapse into the same `continue`:

1. **`@RequestMapping(method = RequestMethod.POST)`** — the verb is a static, resolvable literal. Dropping this loses information the AST already carries.
2. **Bare `@RequestMapping("/path")`** — genuinely multi-verb (Spring maps it to *all* verbs). Not so much "ambiguous" as *method-agnostic*, which the codebase already has a convention for: `normalizeRouteMethod` in `route-extractors/route-path.ts` preserves `'*'`, documented as "`'*'` marks a method-agnostic route… so the contract layer emits a wildcard provider that matches consumers of any method", and `routeNodeKey` already falls back to URL-only ids for `'*'` (the same comment cites Django function views as an existing user of that path).

Why it matters: `@RequestMapping` on handler methods is the standard idiom for every Spring MVC codebase predating Spring 4.3 (2016), and remains common in enterprise/legacy Java. For such a project `route_map` returns empty, `Route` / `HANDLES_ROUTE` are absent, and the HTTP entry points are invisible to `impact` and process tracing — the controller methods are indexed as ordinary methods with no route identity. This also blocks the acceptance criteria "GitNexus detects Spring MVC routes from annotations" / "connects routes to controller methods" in #1225 for that whole class of projects.

Filing separately from #1225 since this is a concrete, independently fixable defect in already-shipped code rather than new framework modelling.

### Expected behavior

The three controllers below are structurally identical and each declares handler methods, so each should yield `Route` nodes:

| Annotation | Expected route |
|---|---|
| `@GetMapping("/list")` | `GET /api/list` |
| `@RequestMapping(value = "/save", method = RequestMethod.POST)` | `POST /api/save` — verb is a static literal |
| `@RequestMapping("/list")` / `@RequestMapping(value = "/detail")` | `* /api/list`, `* /api/detail` — method-agnostic, per the existing `'*'` convention |

4 routes total across the three files.

### Actual behavior

Only the `@GetMapping` route is emitted; both `@RequestMapping` controllers produce nothing:

```
| r.id                | r.name    | r.filePath                                           | r.method |
| Route:GET /api/list | /api/list | src/main/java/com/example/demo/ModernController.java | GET      |
```

`row_count: 1`. `LegacyController` (2 handlers) and `VerbController` (1 handler) contribute 0 routes, while their `Method` nodes are indexed normally — confirming the files parse fine and only route attribution is lost.

### Steps to reproduce

1. Create a fresh repo with three controllers under `src/main/java/com/example/demo/`.

`ModernController.java` (control case — shortcut annotation):
```java
package com.example.demo;

import org.springframework.stereotype.Controller;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.RequestMapping;

@Controller
@RequestMapping("/api")
public class ModernController {

    @GetMapping("/list")
    public String list() {
        return "list";
    }
}
```

`LegacyController.java` (bare `@RequestMapping`, positional + `value =` forms):
```java
package com.example.demo;

import org.springframework.stereotype.Controller;
import org.springframework.web.bind.annotation.RequestMapping;

@Controller
@RequestMapping("/api")
public class LegacyController {

    @RequestMapping("/list")
    public String list() {
        return "list";
    }

    @RequestMapping(value = "/detail")
    public String detail() {
        return "detail";
    }
}
```

`VerbController.java` (`@RequestMapping` with an explicit, statically-known verb):
```java
package com.example.demo;

import org.springframework.stereotype.Controller;
import org.springframework.web.bind.annotation.RequestMapping;
import org.springframework.web.bind.annotation.RequestMethod;

@Controller
@RequestMapping("/api")
public class VerbController {

    @RequestMapping(value = "/save", method = RequestMethod.POST)
    public String save() {
        return "save";
    }
}
```

2. Initialise the repo (`git init`, stage and make one commit — GitNexus needs a HEAD).
3. `npx gitnexus@1.6.9 analyze`
4. `npx gitnexus@1.6.9 cypher -r <repo> "MATCH (r:Route) RETURN r.*"`
   → `row_count: 1` (only `GET /api/list`); expected 4.
5. `npx gitnexus@1.6.9 cypher -r <repo> "MATCH (m:Method) RETURN m.name, m.filePath"`
   → all 4 methods present, so the files parsed correctly and only the route mapping is missing.

Unit-level confirmation, isolating the extractor from the rest of the pipeline:

```js
import Parser from 'tree-sitter';
import Java from 'tree-sitter-java';
import { extractSpringRoutes } from './dist/core/ingestion/route-extractors/spring.js';
import { getTreeSitterBufferSize } from './dist/core/ingestion/constants.js';

const p = new Parser(); p.setLanguage(Java);
const tree = p.parse(src, undefined, { bufferSize: getTreeSitterBufferSize(src) });
extractSpringRoutes(tree, file, 0);
// LegacyController.java  -> 0 routes
// VerbController.java    -> 0 routes
// ModernController.java  -> 1 routes
```

The tree-sitter query itself is not at fault — running `ROUTE_ANNOTATION_QUERY`'s method-declaration branch directly against `LegacyController` matches its annotations fine; they are discarded later by the `METHOD_ANNOTATION_TO_HTTP` gate.

Also reproduced on `1.6.10-rc.154` (same `continue`, same table), so it is not fixed on the current rc line.

### Environment

- OS: Windows 11 (10.0.26100)
- Node: v24.13.0
- Version: `1.6.9` (latest stable); also verified against `1.6.10-rc.154`

### Logs / screenshots

Possible fix shape, reusing conventions already present in the codebase (happy to open a PR if the direction looks right):

- Parse the annotation's `method =` attribute; when it resolves to one or more `RequestMethod.X` literals, emit one route per verb. The rc line already ships `frameworks/spring/annotation-arguments.ts` (`parseSpringAnnotationArguments`), which looks well suited to this.
- When no `method =` attribute is present, emit the route with method `'*'` — `normalizeRouteMethod` already preserves it and `routeNodeKey` already handles it via the URL-only id fallback, so no new identity semantics are needed.
- Apply the change symmetrically in `group/extractors/http-patterns/java.ts` so the two extractors do not drift (#2078); that should also remove one of the remaining blockers for flipping Java to `routeCoverage: 'complete'`.

Open question for maintainers: whether bare `@RequestMapping` should map to `'*'` or expand to the full verb set. `'*'` seems preferable — it matches the documented wildcard-provider intent and avoids inventing routes no consumer calls.

azizur100389:
## Summary
- emit method-level Spring `@RequestMapping` routes for wildcard, scalar, array, and fully-qualified `RequestMethod` declarations
- keep ingestion and group scanners in parity, including inherited interface routes, and invalidate stale parse-worker caches
- add focused unit and integration regression coverage for route nodes, handlers, and `HANDLES_ROUTE` edges

Closes #2824

## Test plan
- [x] Focused Spring, group-route, inheritance, annotation, and parse-cache suites (275 tests)
- [x] `npx tsc --noEmit`
- [x] `npm run build`
- [x] Prettier checks on changed TypeScript files
- [x] ESLint on changed TypeScript files (0 errors; 5 pre-existing non-null assertion warnings)
- [x] GitNexus staged `detect_changes` gate (medium expected scope; 12 symbols, 5 route-scanning flows)
- [ ] Full `npm test` was attempted, but the Windows environment could not load LadybugDB FTS/VECTOR extensions and the run stalled; all relevant isolated suites pass


gitnexus/src/storage/parse-cache.ts
```diff
@@ -248,7 +248,13 @@ import type { ParseWorkerResult } from '../core/ingestion/workers/parse-worker.j
...
 // PR #2840 (Objective-C, draft) still claims 44 as well — it must move too.
 // RE-CHECK AGAINST origin/main IMMEDIATELY BEFORE MERGING.
-const SCHEMA_BUMP = 45;
+//
+// 45 -> 46: method-level Spring `@RequestMapping` now emits wildcard routes
+// and one route per static `RequestMethod.X` value. These decorator routes live
+// in ParseWorkerResult and are replayed verbatim on warm cache hits, so keeping
+// v45 would make the fix a no-op for every unchanged Java file. Verified
+// against upstream/main at 021ac3037 (still 45). RE-CHECK BEFORE MERGE.
+const SCHEMA_BUMP = 46;
```
magyargergo: **P1 — merge gate: `46` is already claimed by open PR #2856.** `[Codex+Claude]` `[reproduced]`

`origin/main` is at 45, and `gh pr diff 2856` shows `+const SCHEMA_BUMP = 46;` there too (PR #2856, *"close reported graph blind spots…"*, open, pins `toBe(46)` as well).

**Trigger:** merge both, in either order, within one published package version. `PARSE_CACHE_VERSION` is `${SCHEMA_BUMP}+${GITNEXUS_PKG_VERSION}` and is the *only* cache-generation key. Whichever lands second inherits a generation the first already populated, so warm caches written under the first build are served as valid for the second — and `decoratorRoutes` is a `ParseWorkerResult` field replayed verbatim. The second feature becomes a **silent no-op on every unchanged file** while all cold-run tests still pass. Exactly what lines 252-256 say this bump exists to prevent.

The ledger above records this three times already and explains why the pin can't catch it (lines 244-248: *"both PRs asserted `toBe(44)`, which passes even when main is already 44"*). It also establishes the convention of **naming** the co-claimant — the v44 entry names PR #2840; this entry names none.

**Fix:** whichever merges second takes the next free integer, updates the pin at `test/unit/incremental-parse-cache.test.ts:137` in the same commit, and names the other PR here. The *decision* to bump is correct; only the number is contested.

gitnexus/src/core/group/extractors/http-patterns/java.ts
```diff
@@ -505,18 +505,20 @@ function scanRouteAnnotations(tree: Parser.Tree): RouteAnnotationScan {
     const keyNode = captures.key; // undefined for the positional shape
 
     if (node.type === 'method_declaration') {
-      // Method-level: a Spring `@(Get|...)Mapping` route, or native `@RequestLine`.
-      const httpMethod = METHOD_ANNOTATION_TO_HTTP[ann];
-      if (httpMethod) {
+      // Method-level: a Spring shortcut/`@RequestMapping` route, or native `@RequestLine`.
+      const httpMethods = springAnnotationHttpMethods(ann, annNode.parent?.text ?? '');
```
magyargergo: **P1 — merge gate: this now mints an OpenFeign consumer with verb `'*'`, which matches every provider verb on the path.** `[Codex+Claude]` `[reproduced]`

The Feign branch this feeds (`:726-742`, emit at `:734`) is unchanged and its *role* handling is correct — a `@FeignClient` method rightly stays a consumer, not a served route. The **verb** is the problem.

Reproduced against `JAVA_HTTP_PLUGIN.scan` at head:
```
@FeignClient(name="svc", path="/api") interface Client { @RequestMapping("/x") Object x(); }
  -> ["consumer/openfeign * /api/x"]
CONTROL, @GetMapping("/x")
  -> ["consumer/openfeign GET /api/x"]      <- the wildcard is introduced by this PR
```
`matching.ts:126-133` makes a `'*'` **consumer** match a provider of *any* verb on that path, one `CrossLink` each at `matchType: 'exact'`, `confidence: 1.0`. A lane reproduced the end state: **one Feign method -> four cross-repo edges** against a service exposing GET/PUT/DELETE/PATCH; three are calls the code cannot make.

`@RequestMapping` means "all verbs" *inbound*, but a Feign client is one outbound request with one fixed verb. Spring Cloud's `SpringMvcContract.processAnnotationOnMethod` substitutes `RequestMethod.GET` when `method()` is empty, then `checkOne(...)` **rejects** more than one — so `method={GET,POST}` on a Feign method currently fans out to two contracts for source Spring Cloud refuses to proxy at all.

Pre-PR this method produced zero contracts, so this is missing-edge -> **wrong** edge, the class `spring.ts:184-188` calls *"a false signal, worse than a missing one"*. Unlike the other gaps here, there is **no pre-existing control** — `@GetMapping` gets it right.

**Fix:** in the Feign branch map `'*'` -> `'GET'` before emitting (matching `SpringMvcContract`), or skip the emit. A method-agnostic *consumer* isn't a meaningful contract. No Feign case is covered by the new tests.

gitnexus/src/core/ingestion/route-extractors/spring-shared.ts
```diff
@@ -34,6 +35,71 @@ export const METHOD_ANNOTATION_TO_HTTP: Record<string, string> = {
...
+  annotationText: string,
+): readonly string[] {
+  const shortcut = METHOD_ANNOTATION_TO_HTTP[annotationName];
+  if (shortcut) return [shortcut];
+  if (annotationName !== 'RequestMapping') return [];
+
+  const args = parseSpringAnnotationArguments(annotationText);
+  if (args === null) return [];
+  const methodArgs = args.filter((arg) => arg.name === 'method');
+  if (methodArgs.length === 0) return ['*'];
```
magyargergo, line 104: **P2 — a type-level `@RequestMapping(method = …)` restriction is discarded, publishing a restricted endpoint as `'*'`.** `[Codex+Claude]` `[reproduced]`

This helper's only input is the *method* annotation's text, so it structurally cannot see the enclosing type; class-prefix collection (`spring.ts:136-160`, `java.ts:553-563`) stores only the path prefix.

Reproduced at head — the standard POST-only webhook / command-controller idiom:
```
@RestController @RequestMapping(path="/api", method=RequestMethod.GET)
class C { @RequestMapping("/x") Object x(); }
  ingestion -> ["* /api/x"]     group -> ["provider/spring * /api/x"]
  Spring actually serves:  GET /api/x
```
Spring composes the type-level and method-level `RequestMethodsRequestCondition`; when the method-level set is empty the **type** condition still applies. The graph now asserts the service accepts every verb on a verb-restricted endpoint, and `matching.ts:139-141` lets consumers of every verb match it.

Pre-PR the method-level `@RequestMapping("/x")` was skipped entirely, so this is missing-edge -> over-broad edge — new wrong data, not a widened gap.

**Fix:** capture the type-level verb set alongside the prefix and intersect (`classVerbs ∩ methodVerbs`, `'*'` as identity). If that plumbing is out of scope here, **fail closed** when the enclosing type declares a `method` member rather than publishing `'*'`.

magyargergo (commented):
# Two-engine review digest — PR #2857

`fix(spring): extract method-level RequestMapping routes` · closes #2824 · base `021ac3037` → head `61a84108` · 10 files, +307/−37.

## Engines and lanes

**Two engines, eleven Claude lanes + three Codex legs. Not five independent confirmations — two.**

| Engine | Lanes | Status |
|---|---|---|
| **Claude** — GitNexus swarm | `risk-architect`, `test-ci`, `security`, `facts-historian` | 4/4 produced findings |
| **Claude** — ce-code-review personas | `correctness`, `adversarial`, `testing`, `maintainability`, `standards`, `api-contract`, `performance` | 7/7 produced findings |
| **Codex gpt-5.6-sol / xhigh** — GitNexus swarm (Solo) | `codex-swarm.md` | **live** — 6-lane report, 7 findings, own `npx tsx` extractor probes |
| **Codex gpt-5.6-sol / xhigh** — ce-code-review (report-only) | `codex-ce.md` | **live** — 4 findings, ran `npx tsc --noEmit` (clean) + 361 unit tests |
| **Codex gpt-5.6-sol / xhigh** — native adversarial | `codex-adversarial.md` | **live** — 4 findings, ran the changed suites + typecheck |

All three Codex legs are confirmed live: each artifact carries its own tool trace (thread ids `019fd7e0…`, `019fd7ee…`, `019fd7ff…`), real command output, and findings not present in the other two. `codex-ce` and `codex-adversarial` independently reproduced the same Spring probes as `codex-swarm`.

**Synthesis verification.** Every postable finding below was re-checked against committed content in the worktree at `61a84108` (`git status --short` empty). Claims marked `[reproduced]` were executed by the synthesis agent or the coordinator against the real exported extractors (`extractSpringRoutes`, `JAVA_HTTP_PLUGIN.scan`, `springAnnotationHttpMethods`) with the transform cache disabled; claims marked `[code-read]` were confirmed by reading `git show HEAD:…`.

## CI

Green across the board: **16,955 passed / 0 failed / 72 skipped** on ubuntu + macos + windows shards 1–3, plus typecheck, lint, format, typecheck-web, node-floor 22.18, tree-sitter ABI (3 platforms), packaged-install smoke, benchmarks, eval containment, CodeQL, gitleaks, dependency review. Coverage 82.26% statements / 70.27% branches.

The **only red check is Vercel "Authorization required to deploy"** — the standard fork-PR preview gate. It is what makes `mergeStateStatus` read `UNSTABLE`. **Not a code defect.** The author's local Windows `npm test` stall on LadybugDB FTS/VECTOR extension load is a known environment flake, contradicted by green exact-head Windows CI.

---

## Credit where it is due

This is a careful PR, and several things it did are exactly right:

- **The `'*'` sentinel is reused correctly, not invented.** `'*'` was already the repo's method-agnostic route marker (`route-path.ts:38-67`, `matching.ts:126-141`, `api_impact` at `local-backend.ts:7873`), introduced for Django function views. `routeNodeKey('*', url)` collapses to the bare URL, so a `* /api/thing` route and a `GET /api/thing` route are two distinct nodes with no collision, no duplicate `HANDLES_ROUTE` edge, and no masking. Both engines traced this end-to-end and both refuted the "wildcard breaks route identity" hypothesis.
- **The parity test is the strongest asset in the diff.** `test/unit/group/spring-route-parity.test.ts:127-167` asserts *exhaustive* `Set` equality between the ingestion extractor and the group scanner across bare, named-arg, array-form, cross-product, empty-`method={}`, and fully-qualified shapes — and a **mutation test confirmed it bites**: disabling the `@RequestMapping` branch fails that suite plus the annotation-arguments and interface-inheritance suites, loudly and specifically.
- **Interface inheritance was brought along properly**, through the shared `resolveInheritedSpringRoutes`, with both `@Controller` and `@RestController` recognised and both sides asserted equal (`spring-interface-inheritance.test.ts`).
- **The deliberate `#2280` class-array divergence is pinned, not silently carried** — the suppression has its own test so it cannot regress into emitting wrong routes.
- **Ingestion was brought to parity with the group scanner on `scoped_identifier` (FQN) annotations**, closing a real asymmetry.
- **`@RequestLine` and `@HttpExchange` were not disturbed** — verified by execution, not inspection.
- **The `SCHEMA_BUMP` was correctly identified as necessary** and its rationale documented in the ledger. The problem is the *number*, not the decision.
- Fan-out is bounded (≤8 verbs, deduped), no `any`, no non-deterministic assertions, all new suites inside the CI shard globs.

---

## Headline: two merge gates

### P1 — `SCHEMA_BUMP = 46` is an exact clash with open PR #2856 `[Codex+Claude]` `[reproduced]`

**File:** `gitnexus/src/storage/parse-cache.ts:257` (added line) · pin at `gitnexus/test/unit/incremental-parse-cache.test.ts:137`

**Trigger:** merge #2856 and #2857 in either order within one published package version.

**Evidence (re-verified live during synthesis, 2026-08-06T16:35Z):**
- `git show origin/main:…/parse-cache.ts` → `SCHEMA_BUMP = 45` (line 251).
- `gh pr diff 2856` → `-const SCHEMA_BUMP = 45;` / `+const SCHEMA_BUMP = 46;`. PR #2856 (*"feat: close reported graph blind spots in reference resolution, analyze and storage"*) is **OPEN**, updated minutes ago, and pins `toBe(46)` too.
- `PARSE_CACHE_VERSION = ${SCHEMA_BUMP}+${GITNEXUS_PKG_VERSION}` is the sole cache-generation key, and `decoratorRoutes` is a field of the cached `ParseWorkerResult` replayed verbatim on a warm hit.

**Consequence:** whichever PR merges second inherits a generation the first already populated. Warm caches written under the first PR's build are served as valid for the second, so the second feature is a **silent no-op on every unchanged file** while every cold-run test still passes — precisely the failure this PR's own comment (lines 252-256) says the bump exists to prevent.

The file's ledger records this exact failure **three times before** and states why the pin cannot catch it: *"both PRs asserted `toBe(44)`, which passes even when main is already 44"* (`parse-cache.ts:244-248`). The ledger's own convention is to **name** the co-claimant (the v44 entry names PR #2840); this PR's v46 entry names none. The PR's note *"Verified against upstream/main at 021ac3037 (still 45)"* is accurate but checks only `main`, not sibling open PRs.

**Fix:** whichever lands second moves to the next free integer, updates the pin in the same commit, and names the other PR in the ledger comment. One line plus one test literal.

*Note: the pin test itself is **not** a defect. It is an audit trail / code-review gate, and the file already documents that it is structurally incapable of catching a cross-PR clash. Do not report it as tautological.*

---

## Second merge gate, then inline-candidate findings

### P1 — an OpenFeign `@RequestMapping` consumer is minted with verb `'*'`, which then matches **every** provider verb on that path `[Codex+Claude]` `[reproduced]`

**File:** `gitnexus/src/core/group/extractors/http-patterns/java.ts:509` (added line; the Feign branch it feeds is unchanged at `:726-742`, emit at `:734`) · root default at `gitnexus/src/core/ingestion/route-extractors/spring-shared.ts:95`

**Trigger:** any `@FeignClient` interface method still on `@RequestMapping` without `method =` — i.e. every Feign client written before the Spring 4.3 shortcut annotations.

**Reproduced during synthesis** (`JAVA_HTTP_PLUGIN.scan`, committed head):

```
@FeignClient(name="svc", path="/api") interface Client { @RequestMapping("/x") Object x(); }
  →  ["consumer/openfeign * /api/x"]
CONTROL, @GetMapping("/x")
  →  ["consumer/openfeign GET /api/x"]        ← the wildcard is introduced by this PR
```

`matching.ts:126-133` makes a `'*'` **consumer** match a provider of *any* verb on that path, one `CrossLink` each at `matchType: 'exact'`, `confidence: 1.0`. The Claude adversarial lane reproduced the end state: **one Feign method → four cross-repo edges** against a service exposing GET/PUT/DELETE/PATCH on that path; three are calls the code cannot make.

**Why `'*'` is wrong here specifically:** `@RequestMapping` means "all verbs" *inbound*, but a Feign client is a single outbound request with one fixed verb. Spring Cloud's `SpringMvcContract.processAnnotationOnMethod` substitutes `RequestMethod.GET` when `method()` is empty, then `checkOne(...)` **rejects** more than one declared method. (Confirming that second half: `@RequestMapping(path="/x", method={GET,POST})` on a Feign method currently fans out to two consumer contracts for a source Spring Cloud refuses to proxy at all.)

**Regression class: missing edge → wrong edge.** Pre-PR this Feign method produced zero contracts. This is new bad data, not a widened gap.

**Fix:** in the Feign branch, map `'*'` → `'GET'` (matching `SpringMvcContract`) before emitting, or skip the emit. A method-agnostic *consumer* is not a meaningful contract.

**Why this is P1 and not P2.** Unlike the pathless-mapping and array-class-prefix items below, there is **no control run showing pre-existing behaviour** — the `@GetMapping` control emits the correct `GET`, so the wildcard is introduced here. It is confirmed *wrong* data, not an absent route: one Feign method mints four false `CrossLink` edges at `matchType: 'exact'`, `confidence: 1.0`. That is exactly the class this codebase calls out at `spring.ts:184-188` as *"a false signal, worse than a missing one"*. Both engines rated it blocking (Codex swarm; Claude adversarial at 90% confidence, reproduced end-to-end). Applying this digest's own calibration rule, it belongs in the blocking tier.

### P2 — a type-level `@RequestMapping(method = …)` restriction is discarded, publishing a restricted endpoint as `'*'` `[Codex+Claude]` `[reproduced]`

**File:** `gitnexus/src/core/ingestion/route-extractors/spring-shared.ts:95` (`if (methodArgs.length === 0) return ['*'];`) — the helper's whole input is the *method* annotation's text, so it structurally cannot see the enclosing type. Class-prefix collection (`spring.ts:136-160`, `java.ts:553-563`) stores only the path prefix.

**Trigger:** the standard POST-only webhook/command controller idiom.

**Reproduced during synthesis:**

```
@RestController @RequestMapping(path="/api", method=RequestMethod.GET)
class C { @RequestMapping("/x") Object x(); }
  ingestion → ["* /api/x"]        group → ["provider/spring * /api/x"]
  Spring actually serves:  GET /api/x   (type + method RequestMethodsRequestCondition compose)
```

The graph now asserts the service accepts every verb on a verb-restricted endpoint, and via `matching.ts:139-141` that provider is matched by consumers of *every* verb on the path.

**Regression class: missing edge → over-broad edge.** Pre-PR the method-level `@RequestMapping("/x")` was skipped entirely, so no route existed. New wrong data.

**Fix:** capture the type-level verb set alongside the prefix and intersect (`classVerbs ∩ methodVerbs`, `'*'` as identity) — or, if plumbing that is out of scope, fail closed when the enclosing type declares a `method` member rather than publish `'*'`.

### P2 — a valid-Java trailing comma in `method = {GET, POST,}` drops the entire route `[Codex+Claude]` `[reproduced]`

**File:** `gitnexus/src/core/ingestion/route-extractors/spring-shared.ts:64` (`for (const rawPart of body.split(','))`)

**Trigger:** `@RequestMapping(path="/x", method={RequestMethod.GET, RequestMethod.POST,})` — JLS §10.6 permits a trailing comma in an array initializer, and one-verb-per-line formatting commonly leaves one.

**Reproduced during synthesis:** `springAnnotationHttpMethods('RequestMapping', '@RequestMapping(path="/x", method={RequestMethod.GET, RequestMethod.POST,})')` → `[]`. The final empty segment fails the regex, `parseRequestMethodValues` returns `null`, the helper returns `[]`, and both callers `continue` — so **both** verbs vanish, not just one.

**Regression class: still absent — incomplete fix, not a regression.** The route did not exist before this PR either; nothing that worked broke. It is listed at P2 because the new parser explicitly claims to handle static `RequestMethod` arrays and the fix is one line: drop a single terminal empty segment inside braces (keep fail-closed for genuinely dynamic expressions).

*The same fail-closed path also swallows `method = RequestMethod . GET` (whitespace around the dot) and — reproduced — an inline `// comment` inside the argument list, which drops the route in both layers. Same one-line neighbourhood, same "was already absent" calibration.*

### P2 — pathless `@RequestMapping(method = RequestMethod.X)` still emits nothing `[Codex+Claude]` `[reproduced]` — **but this is a pre-existing query constraint, and Codex's "blocks merge" rating does not survive the control run**

**File:** `gitnexus/src/core/ingestion/route-extractors/spring.ts:79` (method-declaration query branch, changed line) — every query branch requires a `string_literal` `@value`; same shape at `java.ts:130`.

**Reproduced during synthesis, with the control:**

```
@RestController @RequestMapping("/users") class C { @RequestMapping(method=RequestMethod.GET) Object list(); }
  → ingestion []   group []          (Spring serves GET /users)
CONTROL, pathless @GetMapping (marker) — a shape that predates this PR entirely
  → ingestion []   group []          ← identical
```

The helper resolves the verb correctly on its own (`springAnnotationHttpMethods(…)` → `["GET"]`); the *query* never produces a match, and it never did for any pathless mapping annotation. So this is **(a) an incomplete fix, not a regression** — the route was absent before too — **and (b) a pre-existing, annotation-agnostic query constraint**, not something this PR introduced. Two Codex legs rated it P1 / blocks merge; the control refutes that.

It stays at P2 rather than P3 because it is *the* pre-4.3 idiom #2824 is written about, and nothing in the diff records the limitation — the PR reads as if the `@RequestMapping` gap is closed. **Minimum acceptable resolution: a `#2280`-style comment plus a pin test** documenting the gap, so a later fix registers as a deliberate behaviour change. Full fix: a query branch matching a mapping annotation with no route literal, emitting `routePath: ''` (downstream already handles it — `normalizeExtractedRoutePath('', '/users')` → `/users`).

### P3 — the `SPRING_REQUEST_METHODS` membership check is unreachable by construction `[Claude only]` `[code-read]`

**File:** `gitnexus/src/core/ingestion/route-extractors/spring-shared.ts:70`

`match[1]` is capture group 1 of a closed, case-sensitive alternation over exactly the same eight verbs as the `Set` (lines 38-47, 67). When `match` is non-null the `.has()` can never be false. Two lists encoding one rule: adding a ninth verb to only the Set silently does nothing; adding it to only the regex wrongly rejects. Three Claude lanes flagged it independently (risk-architect, correctness, maintainability); no Codex leg did — consistent within one engine, not cross-engine. One-line fix: drop the Set, or derive the alternation from it.

---

## Lower-priority items (body-only)

| # | Finding | Sev | Engines | Tag |
|---|---|---|---|---|
| 1 | **Kotlin was not brought along.** `kotlin.ts:888,980` still do `METHOD_ANNOTATION_TO_HTTP[annNode.text]; if (!httpMethod) continue;`, so Kotlin method-level `@RequestMapping` remains invisible while Java now resolves it — against the "Java and Kotlin emit byte-identical provider contracts" comment at `java.ts:669-671`. Second-order: `parseRequestMethodValues` only strips `{}`, and Kotlin writes `method = [RequestMethod.POST]`, so the helper cannot be dropped in as-is. **#2824 scoped itself to the Java layers and never mentions Kotlin**, so this is a follow-up, not a missed requirement — but it should be named in the PR body or a tracking comment rather than left as an invisible asymmetry. | P3 | Claude only (4 lanes) | `[code-read]` |
| 2 | **A `'*'` route and a specific-verb route now coexist on one path**, which the `'*'` sentinel's design never anticipated (it was built for Django function views, where the wildcard is the *only* route for a path). Reproduced: a half-migrated controller with `@RequestMapping("/thing")` + `@GetMapping("/thing")` emits `["* /api/thing","GET /api/thing"]` on both sides. `matching.ts:139-141` then pushes **both** keys for a specific-verb consumer → two cross-links where one is correct. Same premise shift flips `api_impact({route, method})` from the singular response shape to `{routes, total}` (`local-backend.ts:7872-7874` + `:7997-8002`), so a client reading `result.handler` silently gets `undefined`. Neither consumer was touched by this PR; both are reachable from it for the first time. | P3 | Claude only | `[reproduced]` (emission) + `[code-read]` (consumers) |
| 3 | **Scalar method path under an array-form class prefix emits an unprefixed route** (`spring.ts:199` path, `classPrefix` falls through to `''`). Reproduced: `@RequestMapping({"/v1","/v2"})` + `@RequestMapping(path="/x", method=GET)` → ingestion `GET /x` (a URL that does not exist) while the group scanner correctly emits `/v1/x` and `/v2/x`. **The control with `@GetMapping("/x")` gives the identical `GET /x`** — this is the documented, deliberate divergence already tracked in **#2280** and pinned in the header comment of `spring-route-parity.test.ts`. Codex rated it "blocks merge" in two legs; the control refutes that. The PR only makes it reachable through one more annotation. | P3, pre-existing | Codex+Claude (severity demoted by control) | `[reproduced]` |
| 4 | **Abstract base controllers publish routes at the base class's prefix.** `spring.ts:178` skips only *interfaces*, and `extends` is not modelled, so `@RequestMapping("/base") abstract class Abstract… { @RequestMapping(method=DELETE, path="/{id}") }` mints a phantom `DELETE /base/{id}` while the real `DELETE /api/users/{id}` is still missing. Mechanism predates the PR (`@GetMapping` in an abstract base already behaved this way); what is new is that the shape which overwhelmingly uses this pattern now trips it. | P3, pre-existing/widened | Claude only | `[reproduced]` |
| 5 | **O(N²) reparse for array-form `@RequestMapping`.** The route query emits one match per array element, and `spring.ts:165` now calls `springAnnotationHttpMethods(ann, annNode.parent?.text ?? '')` per match, re-lexing the whole annotation text each time. Measured by the Claude performance lane and re-run during synthesis at committed head: `@RequestMapping(path={N}, method=GET)` 200→1600 elements = 10.3ms→243.6ms (~3.5× per doubling), while the `@GetMapping` control over the same arrays stays roughly linear (4.5ms→25.9ms). Pre-PR a `@RequestMapping` match was an O(1) map miss, so the cost is new. **Only material for a single annotation with hundreds-to-thousands of paths** — not hand-written Spring, but reachable via generated code, and this repo has a documented history of exactly this bug class. One-line fix: memoize by `annNode.id` across matches of the same annotation. | P3 | Claude only | `[reproduced]` |
| 6 | **`simpleName` re-derived inline instead of shared.** `spring.ts:136` and `:165` each inline `annNode.text.split('.').pop() ?? annNode.text`, duplicating the named `simpleName()` helper at `java.ts:407-408` — in the very PR that creates `spring-shared.ts` exports to stop the two extractors drifting. Four inline copies now live in `spring.ts`. Cheap consolidation. | P3 | Claude only | `[code-read]` |
| 7 | **`annNode.parent?.text ?? ''` fails *open*.** `springAnnotationHttpMethods('RequestMapping','')` returns `['*']`, because `parseSpringAnnotationArguments('')` returns `[]` rather than `null`, so the `if (args === null) return []` guard never fires. **Unreachable today** — the capture's parent is always the `annotation` node, verified for both `identifier` and `scoped_identifier` — but the default silently manufactures a wildcard route rather than failing closed. `?? null` + a null branch would make the contract honest. | P3 | Claude only | `[reproduced]` |
| 8 | `spring-shared.ts:96` (`if (methodArgs.length !== 1) return [];`) is unreachable for compiling Java (JLS §9.7.1 forbids duplicate element names), and its pinning test asserts on source `javac` rejects. Likewise the "runtime expression" test uses `method = VERB` where `VERB` is a variable — also not valid Java. Harmless, but the safety net is pinned only against impossible inputs while the shapes that *do* legally hit it (trailing comma, inline comment) are untested. | P3 | Claude only | `[code-read]` |
| 9 | The `METHOD_ANNOTATION_TO_HTTP` doc comment (`spring-shared.ts:22-29`) still says *"Callers handle `@RequestMapping` separately"* — true only for `kotlin.ts` now. | P3 | Claude only | `[code-read]` |

---

## Refuted — suspicions that did not survive verification

Validation is a feature; these were all chased and dropped.

- **`'*'` breaks route keys or group matching — REFUTED.** `route-path.ts:38-67` documents `'*'` as the pre-existing method-agnostic sentinel and `routeNodeKey` falls back to a URL-only id; `matching.ts:126-141` handles it symmetrically in both directions. Both pre-date this PR. `api_impact` (`local-backend.ts:7873`) already special-cases it. No collision, no duplicate `Route` node, no competing `HANDLES_ROUTE` edge (`routes.ts:230-237` is first-writer-wins on a distinct key).
- **ReDoS in the `RequestMethod` regex — REFUTED, measured.** Sub-linear to linear: 2001→512001 chars gives ~261× time for 256× length; the repeated group is separated by mandatory dots so the partition is unambiguous. Two Claude lanes measured it independently.
- **FQN `@com.evil.RequestMapping` mints a route — REFUTED as new.** Trailing-segment-only discrimination is documented, intentional, pre-existing policy at `java.ts:498-503` ("package-origin gating would break Kotlin parity and is deliberately not done"). The PR brings ingestion to that documented parity.
- **The verb fan-out leaks into `@RequestLine` / `@HttpExchange` — REFUTED, reproduced.** `springAnnotationHttpMethods` returns `[]` for both names so the `else if` chain is still reached: `@RequestLine("GET /y")` → `consumer/openfeign GET /y`; `@GetExchange` → `consumer/spring-http-interface GET /svc/z`. Unchanged.
- **A `@FeignClient` method becomes a served provider route — REFUTED.** The Feign guard at `java.ts:726-742` holds uniformly for the new `methodRoutes`; ingestion skips interface-enclosed methods entirely. The role is right; only the **verb** is wrong (see P2 above).
- **Bare `GET` (static import) false-positives on an unrelated constant — REFUTED.** `RequestMapping#method` is typed `RequestMethod[]`, so JLS §9.7.1 makes any non-`RequestMethod` constant in that position a compile error. `HttpMethod.GET`, `MyEnum.GET`, `RequestMethod.get` all correctly return `[]`.
- **Cross-product blowup / silent cap in the new fan-out — REFUTED.** Verbs are capped at 8 and deduped; paths are source-bounded; no arbitrary cap was introduced. Codex explicitly looked for the `#2793`/`#2732` failure shape and found none. (The inherited-route fan-out does gain an 8× verb axis — 96 routes from one pathological interface method — but it is bounded by source size, and a debug-level warning is the most that is warranted.)
- **Meta-annotation / composed-annotation leakage — REFUTED.** `@interface` declarations are typed `annotation_type_declaration` and match no query branch; uses of custom composed annotations produce nothing (a coverage gap, fail-closed).
- **Constant-valued paths (`@RequestMapping(path = ApiPaths.DETAIL)`) — pre-existing, fail-closed**, unchanged by this PR.
- **The `SCHEMA_BUMP` pin test is tautological — REFUTED.** It is a numeric equality pin that genuinely fails on drift; it is a code-review gate and audit trail, and the file itself documents that no pin can catch a cross-PR clash. Not the deny-set shape seen in PR #2855.
- **The Windows `npm test` stall — environment**, contradicted by green exact-head Windows CI.

---

## Test gaps

1. **No `@FeignClient` coverage at all** — the single highest-risk new shape, and the one that produces the P2 wildcard consumer. `spring-route-parity.test.ts` compares *provider* sets only and would never have caught it.
2. **No Java↔Kotlin parity test** — `spring-route-parity.test.ts` is Java-to-Java, so the Kotlin divergence ships green.
3. **No test for the valid-Java shapes that hit fail-closed** — trailing comma, whitespace around the qualifier dot, inline comment in the argument list. Both existing "fails closed" tests use source `javac` rejects.
4. **No pin for the pathless mapping gap**, in either direction — contrast with the `#2280` divergence, which the same PR documents *and* pins.
5. **No test pins `'*'` + specific-verb coexistence on one path** — the integration test uses three disjoint paths, so the interaction is untested by construction.
6. `spring-route-pipeline.test.ts:117-129` asserts `toContain` per expected route but never `handledLegacyRoutes.size === 4`, so an over-fan-out would pass at the integration layer (mitigated: the parity test pins exhaustiveness at the extractor layer).
7. Style note, low confidence: `spring-route-pipeline.test.ts:90-116` uses `if` inside `forEachNode`/`forEachRelationship` callbacks. This is traversal filtering, mirrors the file's own pre-existing `routeNames()` helper, and is not assertion branching — flagged for awareness only, not as a violation.

---


## Verdict

**Request changes — two merge gates.**

The extraction logic is correct and genuinely well-tested: the parity and inheritance suites are strong, the mutation check confirms they bite, the `'*'` sentinel is reused exactly as designed, and CI is green on all three platforms. Issue #2824's stated acceptance table is met in full.

What must happen before merge:

1. **Resolve `SCHEMA_BUMP` with PR #2856** (P1). Whichever merges second takes the next free integer, updates its pin in the same commit, and names the other PR in the ledger — the convention the file already established.
2. **Stop minting the OpenFeign `'*'` consumer** (P1). Map `'*'` → `'GET'` in the Feign branch (matching `SpringMvcContract`) or skip the emit. This is the one finding here that puts *wrong* edges into the graph at `confidence: 1.0`, and it has no pre-existing control — the `@GetMapping` control emits the correct verb.

What should happen in this PR rather than after it:

3. **The discarded type-level `method=` condition** (P2) — same class as #2, a previously missing edge becoming an over-broad one, which this codebase explicitly treats as worse than missing (`spring.ts:184-188`: *"a false signal, worse than a missing one"*). Fail closed when the enclosing type declares a `method` member rather than publishing `'*'`. (Kept at P2 rather than P1 on two grounds. First, engine consensus: **every** leg that reached this finding rated it P2 independently — Claude adversarial P2-2 at 95% confidence, `codex-swarm` finding 3, `codex-ce` finding 3 — whereas the Feign case is the one place both engine families independently landed on P1. Second, scope: the fix needs the class-level verb set plumbed through to a helper whose only input today is the method annotation's own text, which is architectural work beyond this PR's stated goal; the Feign fix is a one-line context override in a branch that already exists.)
4. **The trailing-comma / comment fail-closed shapes**, plus a regression test — one line each.
5. **Either fix or explicitly document the pathless-mapping gap** with a `#2280`-style pin, so it is a recorded limitation rather than an assumed-closed one.

Everything else — Kotlin parity, the `'*'`-coexistence premise shift, `api_impact` polymorphism, the O(N²) reparse, the dead guard — is legitimate follow-up and should not hold the PR.

Two clarifications on Codex's verdict, honoured by the control runs: the **array-class-prefix case does not block merge** (identical behaviour with `@GetMapping`; tracked in #2280 and already pinned), and the **pathless-mapping case is an incomplete fix, not a regression** (the same shape with `@GetMapping` is equally absent on `main`). Nothing that previously worked has broken.

---

*Automated multi-tool digest — two engines (Claude: GitNexus swarm + ce-code-review; Codex `gpt-5.6-sol`/xhigh: GitNexus swarm + ce-code-review + adversarial), merged by a separate synthesis agent and gated by an independent critic. Findings tagged `[reproduced]` were executed against the real extractors at head `61a84108`; `[code-read]` means the code was re-read, not run. Please verify before acting.*

gitnexus/src/core/ingestion/route-extractors/spring-shared.ts
```diff
@@ -34,6 +35,71 @@ export const METHOD_ANNOTATION_TO_HTTP: Record<string, string> = {
...
+  const trimmed = value.trim();
+  const hasOpeningBrace = trimmed.startsWith('{');
+  const hasClosingBrace = trimmed.endsWith('}');
+  if (hasOpeningBrace !== hasClosingBrace) return null;
+  const body = hasOpeningBrace ? trimmed.slice(1, -1).trim() : trimmed;
+  if (body.length === 0) return [];
+  if (!hasOpeningBrace && body.includes(',')) return null;
+
+  const methods: string[] = [];
+  for (const rawPart of body.split(',')) {
```
magyargergo: **P2 — a legal-Java trailing comma drops the whole route, not just one verb.** `[Codex+Claude]` `[reproduced]`

JLS §10.6 permits a trailing comma in an array initializer, and one-verb-per-line formatting commonly leaves one. `body.split(',')` yields a final empty segment, which fails the regex, so `parseRequestMethodValues` returns `null`, the helper returns `[]`, and both callers `continue`:
```
springAnnotationHttpMethods('RequestMapping',
  '@RequestMapping(path="/x", method={RequestMethod.GET, RequestMethod.POST,})')  ->  []
```
Both verbs vanish. The same fail-closed path also swallows an inline `// comment` in the argument list and `RequestMethod . GET` (whitespace around the dot) — each drops the route in both layers.

**Calibration:** these routes didn't exist before this PR either, so nothing that worked broke — this is an incomplete fix, not a regression. Flagged at P2 only because the new parser explicitly claims to handle static `RequestMethod` arrays and the fix is one line: drop a single terminal empty segment when braces are present, keeping fail-closed for genuinely dynamic expressions. Worth a regression test — the two existing "fails closed" tests both use source `javac` rejects.

gitnexus/src/core/ingestion/route-extractors/spring.ts
```diff
@@ -76,14 +76,14 @@ const ROUTE_ANNOTATION_QUERY = new Parser.Query(
     (method_declaration
       (modifiers
         (annotation
-          name: (identifier) @ann
+          name: [(identifier) (scoped_identifier)] @ann
```
magyargergo, line 80: **P2 — pathless `@RequestMapping(method = RequestMethod.X)` still emits nothing.** `[Codex+Claude]` `[reproduced]`

Every branch of this query requires a `string_literal` `@value` (same shape at `java.ts:130`), so a verb-only mapping relying on the class prefix never produces a match — the new helper is never consulted, even though it resolves the verb correctly on its own.
```
@RestController @RequestMapping("/users")
class C { @RequestMapping(method=RequestMethod.GET) Object list(); }
  -> ingestion []   group []          (Spring serves GET /users)
CONTROL, pathless @GetMapping        -> ingestion []   group []   <- identical
```
**The control matters:** this is a pre-existing, annotation-agnostic query constraint, not something this PR introduced, and the route was absent before too. Two Codex legs rated this P1/blocks-merge; the control run doesn't support that, so it's held at P2.

It stays at P2 rather than P3 because it is *the* pre-4.3 idiom #2824 is written about, and nothing in the diff records the limitation — the PR reads as if the `@RequestMapping` gap is closed.

**Minimum:** a `#2280`-style comment plus a pin test documenting the gap, so a later fix registers as a deliberate behaviour change. **Full fix:** a query branch matching a mapping annotation with no route literal, emitting `routePath: ''` — downstream already handles it (`normalizeExtractedRoutePath('', '/users')` -> `/users`).

gitnexus/src/core/ingestion/route-extractors/spring-shared.ts
```diff
@@ -34,6 +35,71 @@ export const METHOD_ANNOTATION_TO_HTTP: Record<string, string> = {
...
+    const part = rawPart.trim();
+    const match =
+      /^(?:(?:[A-Za-z_$][A-Za-z0-9_$]*\.)*RequestMethod\.)?(GET|HEAD|POST|PUT|PATCH|DELETE|OPTIONS|TRACE)$/.exec(
+        part,
+      );
+    if (!match || !SPRING_REQUEST_METHODS.has(match[1])) return null;
```
magyargergo: **P3 — this membership check is unreachable by construction.** `[Claude only]` `[code-read]`

`match[1]` is capture group 1 of the closed, case-sensitive alternation `(GET|HEAD|POST|PUT|PATCH|DELETE|OPTIONS|TRACE)` — the same eight verbs as `SPRING_REQUEST_METHODS` above. When `match` is non-null, `.has()` can never be false, so the second half of the `||` never fires.

Two lists encoding one rule: add a ninth verb to only the `Set` and nothing happens; add it to only the regex and the `Set` wrongly rejects it. Behaviour is correct today because the lists happen to match.

**Fix:** drop the `Set` and guard on `if (!match) return null`, or derive the alternation from the `Set`. Three Claude lanes flagged this independently; no Codex leg did — consistent within one engine, not cross-engine confirmation.

gitnexus/src/core/ingestion/route-extractors/spring.ts
```diff
@@ -160,9 +161,9 @@ export function extractSpringRoutes(
 
     if (node.type !== 'method_declaration') continue;
 
-    const ann = annNode.text;
-    const httpMethod = METHOD_ANNOTATION_TO_HTTP[ann];
-    if (!httpMethod) continue; // skip @RequestMapping on methods (ambiguous verb)
+    const ann = annNode.text.split('.').pop() ?? annNode.text;
+    const httpMethods = springAnnotationHttpMethods(ann, annNode.parent?.text ?? '');
```
magyargergo: **P3 — this is called once per query match, re-lexing the same annotation text each time.** `[Claude only]` `[reproduced]`

For array-form `@RequestMapping(path = {"/a","/b",…})` the route query emits one match **per array element**, while `annNode.parent?.text` is byte-identical across all of them — so `parseSpringAnnotationArguments` re-scans the whole N-sized annotation N times.

Measured by the performance lane and re-run at head: 200 -> 1600 path elements gives 10.3ms -> 243.6ms (~3.5x per doubling), against a roughly linear `@GetMapping` control over the same arrays (4.5ms -> 25.9ms). Pre-PR a `@RequestMapping` match was an O(1) map miss, so the cost is new. `java.ts:509` has the same shape.

**Only material for a single annotation with hundreds-to-thousands of paths** — not hand-written Spring, but reachable via generated code, and this repo has a documented history of exactly this bug class on the per-file parse path.

**Fix:** memoize by `annNode.id` across matches sharing one annotation.

gitnexus/src/storage/parse-cache.ts
azizur100389: Fixed in b0090c90: moved this branch to schema 47, updated the pin, and named PR #2856 as the owner of 46.

gitnexus/src/core/group/extractors/http-patterns/java.ts
azizur100389: Fixed in b0090c90: verb-less Feign RequestMapping now emits GET, while multi-method mappings fail closed. Added parity regressions.

gitnexus/src/core/ingestion/route-extractors/spring-shared.ts
azizur100389: Fixed in b0090c90: class- and method-level verb conditions are intersected in both ingestion and group extraction. Incomparable or dynamic conditions fail closed.

gitnexus/src/core/ingestion/route-extractors/spring-shared.ts
azizur100389: Fixed in b0090c90: braced RequestMethod arrays now accept one terminal comma, with a regression covering GET/POST fan-out.

gitnexus/src/core/ingestion/route-extractors/spring.ts
azizur100389: Fixed in b0090c90: pathless mapping annotations now bind to the class prefix/root in both engines. Explicit dynamic path/value members still fail closed.

gitnexus/src/core/ingestion/route-extractors/spring-shared.ts
azizur100389: Fixed in b0090c90: removed the redundant Set and retained the closed verb regex as the single source of validation.

gitnexus/src/core/ingestion/route-extractors/spring.ts
azizur100389: Fixed in b0090c90: both scanners cache parsed methods by annotation node ID, so path arrays parse each annotation once. The Spring scaling tripwire passes.

azizur100389:
Follow-up 6dc0048e also covers the legal-Java trivia cases from the review: whitespace around RequestMethod member access and inline block/line comments, with helper and ingestion/group parity regressions.

Validated: 268 affected tests + 249 Java resolver tests, typecheck, ESLint, and Prettier.

PR #2859, opened 2026-08-06
perf(scope-resolution): share the property-name index across language passes

ReidenXerx:
> [!IMPORTANT]
> **This PR is stacked on #2856 and cannot merge before it.** `unique-name-properties.ts`, the file it changes, does not exist on `main`.
>
> **The diff against `main` therefore contains all of #2856 as well** — 38 commits, 70 files. Only the final commit, [`ac4147c0`](https://github.com/abhigyanpatwari/GitNexus/pull/2859/commits), belongs to this PR: **4 files, +191/-34**. Reviewing that commit alone is the whole change.
>
> Its sibling follow-up #2860 (the `SCHEMA_BUMP` CI guard) was deliberately split out and is cut from `main`, so that one is independent and can go in any order.

## Summary

One of the two follow-ups called out as deliberately deferred in #2856: the per-language whole-graph scan that PR reintroduced.

The other deferred item — the `SCHEMA_BUMP` CI guard — started here and was **split into #2860** once it became clear it has no dependency on #2856 and should not be dragged behind it.

## Motivation / context

Both items come from the two-engine review of #2856, and both are things that PR was the right size to leave alone:

> the scan is O(total graph nodes) and runs **once per qualifying language**, which reintroduces the exact per-language whole-graph-scan pattern `phase.ts:263-267` hoisted out and documented […] Consider hoisting it alongside `sharedNodeLookup`.

> a CI check that compares against `origin/main`'s value at merge time, rather than a literal pin, is the only thing that structurally closes it.

## Areas touched

- [x] `eval/` or other tooling (`scripts/`)

## Scope & constraints

**In scope**

- **Share the `Property`-by-name index.** Built once in `phase.ts` beside `sharedNodeLookup` and `sharedFnNodeIndex`, threaded through the same `prebuilt*` seam.

**Explicitly out of scope**

- Everything else in #2856, whose commits are in this diff only because of the stacking.
- The `SCHEMA_BUMP` CI guard — now #2860.
- This branch adds no capture rules, no new edges, and no graph-shape change. The hoist is behaviour-preserving by construction: same edges, same counts.

## Implementation notes

**Sharing is only safe because the restriction moved, not disappeared.** The per-language build existed to keep one language from resolving to another's properties (the Java-private-field case in #2856). The shared index is whole-graph, so that restriction now applies at LOOKUP time against the language's own file set.

That reordering also fixes a subtlety the per-language build had backwards. The candidate cap now applies to the **filtered** set, so a name carried by forty properties across a polyglot monorepo but only two in the language being resolved is still answerable — a global cap would have refused it. The lookup boundary is deliberately tri-state, and the three outcomes are not interchangeable: no property of this name in this language (nothing to say, and *not* an ambiguity), too many to choose between (reportable), or a list to narrow.

**The polyglot fixture earned its keep twice.** An intermediate state shared the index without moving the filter, and the cross-language edge came straight back. That is the test #2856 added because every other fixture is single-language and could not catch it by construction.

## Testing & verification

- [x] `cd gitnexus && npm test` — **16,436 passed, 0 failed, 73 skipped (822 files)**
- [ ] `cd gitnexus-web && npm test` *(untouched)*
- [ ] Manual / Playwright E2E *(n/a)*

Four new tests. They pin that the scan happens **exactly once** — counting iterations is what separates "shared" from "rebuilt per language and happens to agree" — that the index is genuinely whole-graph (both carriers of a shared name present), and that a node with no `filePath` never enters it, since such a node could not be language-filtered later and would then be visible to *every* language.

## Risk & rollout

- **No re-index required.** No capture or schema change; `SCHEMA_BUMP` is untouched by this branch.
- **No graph-shape change.** Same edges, same counts — the hoist changes when the index is built, not what it contains for any given language.
- **Merge order is forced.** #2856 must land first. Nothing else depends on this.
- No breaking API changes, no migrations.

## Related

- #2856 — the PR this came out of. **This PR is stacked on it.**
- #2860 — the sibling follow-up (`SCHEMA_BUMP` CI guard), split out of this branch and cut from `main`, so it is independent of both.

## Checklist

- [x] `AGENTS.md` / overlays unchanged


ReidenXerx:
Folding this into #2856 instead.

It only ever made sense stacked — `unique-name-properties.ts` does not exist on `main` — and stacking made it read badly: **67 files and +3319/-72 in the diff for a change that is 4 files and +191/-34.** The other 94% was #2856 replayed, which is a poor thing to put in front of a reviewer.

It is also, on reflection, the wrong seam for a separate PR: it refines the performance of code #2856 introduces, so the two want reviewing together rather than one after the other.

Now commit `4cf2067c` on #2856. The other deferred item, #2860, stays standalone — it is genuinely independent of that work, cut from `main`, and mergeable in any order.

PR #2860, opened 2026-08-06
ci: fail a PR whose SCHEMA_BUMP collides with the base branch

ReidenXerx:
## Summary

Fails a PR whose `SCHEMA_BUMP` collides with the base branch's. This is the follow-up called out as deliberately deferred in #2856, where the reviewer's framing is the whole argument for it:

> a CI check that compares against `origin/main`'s value at merge time, rather than a literal pin, is the only thing that structurally closes it.

**Independent of #2856** — this branch is cut from `main` and touches no runtime code, so it can be reviewed and merged in any order. (Its sibling follow-up, #2859, is *not* independent; see below.)

## Motivation / context

`parse-cache.ts` carries a ledger of **ten** prior version collisions, **four of them exact**. Every one was caught by hand at merge time, and one only after it had already shipped. The ledger itself explains why the existing test cannot help:

> the pin test cannot catch it — both PRs asserted `toBe(44)`, which passes even when main is already 44 […] Only comparing against origin/main at MERGE time surfaces it.

Both branches are green. The conflict exists only in the *relation* between them, and no single-branch assertion can see a relation.

The consequence is silent, which is why it keeps recurring. `PARSE_CACHE_VERSION` is the only invalidator for the durable ParsedFile store — byte-unchanged files skip tree-sitter entirely — so two divergent capture schemas sharing one version string means one side's parse-time change replays pre-change ParsedFiles on every incremental analyze. Nothing throws. The graph is simply missing edges, which is the confident-empty answer #2856 spent thirty-odd commits removing.

## Areas touched

- [ ] `gitnexus/` (CLI / core / MCP server) — *test only*
- [x] `eval/` or other tooling (`scripts/`)

## Scope & constraints

**In scope**

- `scripts/check-parse-cache-version.mjs` — compares `SCHEMA_BUMP` against the PR's base branch.
- A `parse-cache-version` job in `ci-quality.yml`, with the `fetch-depth: 0` and base-branch fetch it needs.

**Explicitly out of scope**

- Changing `SCHEMA_BUMP` itself, or anything about how the parse cache works. This branch adds no runtime code.
- Blocking two PRs that are merely *open* at the same time — see below; that is not the collision, and treating it as one would fail honest work.

## Implementation notes

**Only a branch that modifies the parse cache is claiming a version**, so only that branch has to prove the claim is unique. Everything else passes untouched.

That rule is not cosmetic, and the first version of this check got it wrong: it required a strict increase unconditionally, which fails every PR that does not touch the file — including this one. **It failed on its own branch, which is how it was caught.** A check that cannot pass its own PR would have been noticed by CI, but the fix belongs in the design, not in an exemption.

**Being open at the same time is not the collision.** While base is still 45, two branches can both claim 46 and both pass — exactly the scenario the ledger describes. What catches it is CI re-running once the first one merges and base becomes 46. That is precisely the *"RE-CHECK AGAINST origin/main IMMEDIATELY BEFORE MERGING"* step the ledger already demands in capital letters, now automatic instead of remembered.

**Three outcomes, deliberately distinct.** Pass; conflict (exit 1, naming the number to move to and the two other files that must move with it); and unreachable base ref (exit **2**), so a shallow checkout or a missing fetch reads as a configuration error rather than a pass. That last one is the failure mode a naive version of this check would have — quietly green while testing nothing.

## Testing & verification

- [x] `cd gitnexus && npm test` — the new suite passes; full pack green on the sibling branch that carries this commit (**16,436 passed, 0 failed**)
- [ ] `cd gitnexus-web && npm test` *(untouched)*
- [ ] Manual / Playwright E2E *(n/a)*

Six tests, driven against **real git state rather than mocks**, because the thing under test *is* a comparison between branches: a temp repo with a base at 45, then a branch that is untouched (passes — the common case), at 46 (passes), at 45 with the file edited (the clash that has actually happened four times), at 44 (backwards), an unreachable base ref (must exit 2, not 0), and one where the declaration has been renamed — which must fail loudly rather than silently matching nothing and reporting success.

## Risk & rollout

- **No re-index, no graph change, no runtime code.** `SCHEMA_BUMP` is not touched by this branch.
- **A new CI job that can fail PRs.** By design. It only fires on branches that modify `parse-cache.ts`, and the failure message names the exact remedy: the number to raise to, the pin test to move, and the ledger entry to add recording both claimants.
- **Requires full history.** `fetch-depth: 0` plus a base-branch fetch are wired in the job. If either is missing the script exits 2 rather than passing.
- No breaking API changes, no migrations.

## Related

- #2856 — the PR this follow-up came out of. **Independent of this one**; merge order does not matter.
- #2859 — the sibling follow-up (share the property-name index across language passes). That one **is stacked on #2856** and cannot merge before it.

## Checklist

- [x] `AGENTS.md` / overlays unchanged

### DIFF
--- a/.github/workflows/ci-quality.yml
+++ b/.github/workflows/ci-quality.yml
@@ -22,6 +22,30 @@ jobs:
       - run: npm ci
       - run: npx prettier --check .
 
+  parse-cache-version:
+    # Guards a collision no single-branch test can see: two PRs claiming the
+    # same SCHEMA_BUMP both pass their own `toBe(N)` pin, then share one
+    # PARSE_CACHE_VERSION after merge and one side's parse-time capture change
+    # silently replays pre-change ParsedFiles forever. Ten ledger entries, four
+    # exact clashes, every one caught by hand at merge time until now.
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+          # Full history: the check reads the BASE branch's value, which a
+          # shallow clone does not contain. The script exits 2 (distinct from a
+          # real conflict) if the ref is unreachable, so a misconfiguration here
+          # cannot masquerade as a pass.
+          fetch-depth: 0
+      - name: Fetch base branch
+        run: git fetch --no-tags --depth=1 origin "${GITHUB_BASE_REF:-main}:refs/remotes/origin/${GITHUB_BASE_REF:-main}" || git fetch --no-tags origin "${GITHUB_BASE_REF:-main}"
+      - name: Compare SCHEMA_BUMP against the base branch
+        env:
+          GITNEXUS_BASE_REF: origin/${{ github.base_ref || 'main' }}
+        run: node scripts/check-parse-cache-version.mjs
+
   lint:
     runs-on: ubuntu-latest
     timeout-minutes: 10
--- a/gitnexus/test/unit/parse-cache-version-check.test.ts
+++ b/gitnexus/test/unit/parse-cache-version-check.test.ts
@@ -1,0 +1,131 @@
+/**
+ * The CI guard for concurrent `SCHEMA_BUMP` claims.
+ *
+ * The in-repo pin (`expect(...).toBe(N)`) cannot catch this and the ledger says
+ * so four separate times: both branches assert the same number, so both are
+ * green while they collide. The conflict exists only in the RELATION between
+ * two branches, so the check has to compare against the base — and therefore
+ * has to be tested against real git state rather than mocked.
+ *
+ * Failure here is silent in production, which is why it kept recurring:
+ * `PARSE_CACHE_VERSION` is the only invalidator for the durable ParsedFile
+ * store, so a shared version means one side's parse-time change replays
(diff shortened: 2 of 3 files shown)


azizur100389:
Reviewed at `29e735e5d138e55f4223b9e78b0d603916340cd8`.

**Medium: the equality exemption misses a concurrent collision.** In [scripts/check-parse-cache-version.mjs:78-83](https://github.com/abhigyanpatwari/GitNexus/blob/29e735e5d138e55f4223b9e78b0d603916340cd8/scripts/check-parse-cache-version.mjs#L78), two branches independently making the same45-to46 edit here while changing different capture code elsewhere become byte-identical in this file after the first merges. The second guard then exits0 with `claims no version`, despite independently claiming46.

Please derive the claim from the original PR contribution, not merged-tree equality, and compare it with advanced base. Add a real two-branch regression that merges one claimant and checks the other's merged tree; retain an untouched-branch pass case. Current tests edit one working tree against one base commit. Locally reproduced (2026-09-02, Node 24.19.0, this exact head). The **six supplied guard tests pass**. A real-Git two-branch test with independent 45-to-46 claims, different capture changes, and unchanged package version confirms that B still exits 0 with `claims no version` after incorporating advanced main. The untouched-branch control passes; the supplementary collision-rejection regression fails. This validates the guard gap, not an end-to-end stale-cache replay.

Recommendation: request changes. Current-head repository checks are absent; sibling-branch tests and Vercel do not establish readiness.

PR #2864, opened 2026-08-07
fix(python): resolve calls through `__init__.py` re-exports

clasalle-lila:
## Why

A call to a name imported from a package never resolves when the package's `__init__.py` re-exports it rather than defining it:

```python
# pkg/impl.py
def target_fn(x): ...

# pkg/__init__.py
from pkg.impl import target_fn

# caller.py
from pkg import target_fn
def calls_it(): return target_fn(21)   # ← no CALLS edge
```

Since `__init__.py` re-exports are how Python packages declare a public surface, this misses a large share of real call edges. Worse, it's silent in a specific way: the defining module ends up with zero callers, so it reads as dead code. On the repo I hit this on, a module that's called on every run looked 100% orphaned.

## What Changed

The re-export closure that should carry this already exists and is fully general — `buildReexportClosures` does SCC over the re-export subgraph, a bounded fixpoint for cycles, and transitive `via` chains. Python just never fed it: the subgraph admits only `kind: 'reexport'` and `kind: 'wildcard'`, and Python emits neither for `from m import x`.

The wrinkle is that Python has no dedicated re-export form. A module-level `from pkg.impl import X` binds `X` locally *and* publishes it as `pkg.X` — it's both at once. Emitting `kind: 'reexport'` would be wrong, because that form drops the local binding, which Python's does create.

So instead there's a new optional `reexportsName` flag on the `named` / `alias` variants — alongside the existing provider-specific `importedSymbolKind` and `targetIncludesImportedName` — and flagged imports are admitted into the closure subgraph. Languages with an explicit form keep emitting `kind: 'reexport'` and leave the flag unset, so nothing changes for them.

## How To Review

1. `gitnexus-shared/src/scope-resolution/types.ts` — the new flag and why `kind: 'reexport'` isn't usable here
2. `finalize-algorithm.ts` — `contributesReexportEdge` / `isNamedReexport`, the two guards that previously excluded Python
3. `languages/python/interpret.ts` — where the flag is set

## Verification

Fixture covering three shapes (direct, top-level-via-re-export, function-local-via-re-export): **1 of 3 CALLS edges resolved before, 3 of 3 after.**

On a 12.4k-file Python/Go/TypeScript repo: edges 294,416 → 301,443 (+7,027), execution flows 300 → 813. A module previously reporting zero callers now correctly reports its caller.

`npx tsc --noEmit` clean in both packages. Full unit suite shows no regression against baseline — the remaining failures are pre-existing load-sensitive flakes in `analyzer-identity`, `evidence-provenance-helper`, `skip-git-cli` and `hooks`, each verified passing in isolation.

## Risks

The flag is opt-in per provider, so no other language's resolution widens. There's an explicit negative-control test asserting a plain `named` import still doesn't resolve through a hop — that's the guard against this quietly becoming universal.

5 new finalize tests (single hop, 3-hop chain, alias keying, cycle termination, negative control) and 6 updated Python fixture shapes.


clasalle-lila:
@azizur100389 let me know! I opened 4-5 to fix all!

gitnexus/src/core/ingestion/languages/python/interpret.ts
```diff
@@ -52,23 +52,34 @@ export function interpretPythonImport(captures: CaptureMatch): ParsedImport | nu
...
+      // enter the re-export closure so `from <this module> import x` elsewhere
+      // resolves to the original definition. See `ParsedImport` in
+      // gitnexus-shared.
       if (sourceCap === undefined || nameCap === undefined) return null;
       return {
         kind: 'named',
         localName: nameCap.text,
         importedName: nameCap.text,
         targetRaw: sourceCap.text,
+        reexportsName: true,
```
magyargergo: **P1 · scope-blind flag → false and wrong re-export edges** *(both engines; coordinator-reproduced)*

`interpretPythonImport` receives only a `CaptureMatch`, and the Python query matches `(import_from_statement)` anywhere in the file, so this flag is set for function-local, class-body, `TYPE_CHECKING`, `try/except` and version-branch imports too — verified across all of those positions on the real parse path. The scope cannot be recovered downstream either: `pass3CollectImports` (`scope-extractor.ts:897-910`) applies no scope filter, `finalize-orchestrator.ts:171-178` forwards `parsedImports` verbatim, and `ImportEdgeDraft.fromScope` is hardcoded to `file.moduleScope` (`finalize-algorithm.ts:354, 382, 410`).

**To be clear about where the fault lies: the repo's own docs misled this change.** `scope-extractor.ts:33-40` states pass 3 attaches the `ParsedImport` to the ParsedFile *"(not to any `Scope` — finalize **reconstructs the owning scope via `provider.importOwningScope`** during Phase 2)"*. Finalize does no such thing — `grep -rnE "\.importOwningScope\b" gitnexus/src/` returns **exactly one hit, and it is that doc comment**. The hook is declared at `language-provider.ts:558-562` and implemented by 12+ providers, but is **never invoked**. So the fixture comment at `python-fixtures.test.ts:279` ("finalize will route via importOwningScope") was read straight out of the repo's stale docblock — that stale doc plus the dead hook is a pre-existing defect worth its own follow-up issue.

Reproduced, false edge on code that would `ImportError`:
```python
# pkg/__init__.py
def loader():
    from pkg.impl import InternalHelper
# caller.py
from pkg import InternalHelper
```
→ `targetDefId = def:pkg.impl.InternalHelper`, `via = ['pkg/__init__.py','pkg/impl.py']`, `link: ok` (BASE: `unresolved`).

Reproduced, **wrong** resolution on legal, running Python (declaration-order first-wins at `finalize-algorithm.ts:719` lets the function-local entry claim the key ahead of the real module-level import):
```python
# pkg/__init__.py
def _lazy():
    from pkg.bad import X     # publishes NOTHING
from pkg.good import X        # publishes pkg.X
# caller.py — CPython resolves pkg.good.X
from pkg import X
```
→ `RESOLVED = def:pkg.bad.X`.

Mitigation verified: wildcard fan-out does **not** inherit this (it materializes via `expandsWildcardTo` over `localDefs`), which caps the blast radius.

**Fix — bigger than it looks, so deferring it is defensible.** The scope information is not plumbed anywhere today, so there is no one-liner. Either (a) wire the hook in `pass3CollectImports` (`scope-extractor.ts:897-910`) and clear `reexportsName` when the owning scope is not the module scope — that function already computes `anchorCaptureFor(match, '@import.')` at `:903` and uses it only as a presence guard, so the innermost scope is available right there; **but this means activating a currently-dead hook with 12+ untested implementations**, which is a real change, not a two-line guard. Or (b) carry a module-level marker on `ParsedImport` set at interpret time from a scope-aware capture — smaller blast radius, but it adds another parse-time field and so brings the `SCHEMA_BUMP` obligation with it.

**If you defer the behaviour fix, please still do two things in this PR:** rename the `python-fixtures.test.ts:277-283` case to record a **known over-approximation** and drop its (false) `importOwningScope` claim, and stop asserting "module-level" in the `types.ts` docblock as though it were checked. A documented, tested over-approximation is fine to ship; one whose docs claim it is gated is not.

gitnexus-shared/src/scope-resolution/finalize-algorithm.ts
```diff
@@ -676,7 +709,7 @@ function populateFileClosure(
   // Named re-exports — precedence over wildcards, declaration order
   // first-wins for duplicates of the same exported name.
   for (const draft of drafts) {
-    if (draft.source.kind !== 'reexport') continue;
+    if (!isNamedReexport(draft.source)) continue;
```
magyargergo: **P1 · declaration-order first-wins is inverted for Python** *(both engines; reproduced end-to-end)*

The rule three lines above (`:706-707`, "declaration order first-wins for duplicates of the same exported name", enforced by `if (myClosure.has(localName)) continue;` at `:719`) was written for TS/Rust, where two `export { X } from …` of the same name is a compile error — so first-wins is vacuously safe. CPython's module namespace is **last-write-wins**, and for `if`/`else` and `try`/`except` import pairs exactly one branch executes.

Reproduced — rebind:
```python
# pkg/__init__.py
from .v1 import Client   # legacy, left behind
from .v2 import Client   # the actual public Client
```
→ `{"targetDefId":"def:v1.Client","via":["pkg.py","v1.py"]}`. Every `from pkg import Client` in the repo is now attributed to the **dead** implementation, and `impact("Client")` points at the wrong file.

Reproduced — version-gated pair (`compat.py` with `if sys.version_info >= (3, 11): from new_impl import load / else: from old_impl import load`): BASE `unresolved` → HEAD `def:new_impl.py#1:0:Function:load`, the `>= 3.11` branch unconditionally. On a repo targeting 3.9 that is a wrong def on **working** code. The `try:`/`except ImportError:` shim has the identical shape and is the canonical Python compat idiom.

**Fix:** detect the collision instead of silently picking. When a second `reexportsName` draft targets an already-populated `localName` in the same file, either drop the entry (ambiguous → stay unresolved, matching the pre-PR answer) or record both and let `pythonMergeBindings` arbitrate — it already keeps every distinct `nodeId`, so the closure currently exports *strictly less* than the file binds locally. Whichever you pick, `:706-707` should say the precedence is provider-controlled rather than universal.

(Also worth one line while you are here: this guard tests `draft.source.kind`, but `tryFinalize` tests the post-reclassification `draft.base.kind` at `:473`. A Python `from . import submodule` that `isNamespaceImport` reclassified to `'namespace'` is still admitted here and republishes whatever def shares the module's simple name — `if (draft.base.kind === 'namespace') continue;` closes it. P3, two lanes.)

gitnexus-shared/src/scope-resolution/types.ts
```diff
@@ -128,6 +128,21 @@ export type ParsedImport =
...
+       * `from pkg.impl import X` binds `X` locally **and** publishes it as
+       * `pkg.X`, which is the standard way a package `__init__.py` defines its
+       * public surface. Languages with an explicit form (TS `export … from`,
+       * Rust `pub use`) emit `kind: 'reexport'` instead and leave this unset.
+       *
+       * Setting it lets the name participate in `buildReexportClosures` while
+       * keeping the ordinary local `named` binding intact — `kind: 'reexport'`
+       * would drop the local binding, which for Python would be wrong.
+       */
+      readonly reexportsName?: boolean;
```
magyargergo, line 164: **P1 · this changes the persisted `ParsedFile` shape — `SCHEMA_BUMP` was not bumped** *(4 Claude lanes; coordinator-confirmed; Codex did not surface it)*

The real location is `gitnexus/src/storage/parse-cache.ts:251` (`const SCHEMA_BUMP = 45;`), which this PR does not touch — anchoring here because that file is not in the diff.

`parse-cache.ts:56-57` states the rule as a MUST: *"any future change to the `ParsedFile` serialization shape MUST bump SCHEMA_BUMP so both invalidate in lockstep."* `parsedfile-store.ts:177` serializes the whole `ParsedFile` generically (`JSON.stringify(parsedFiles, mapReplacer)`), so `parsedImports[].reexportsName` is part of that shape. `PARSE_CACHE_VERSION = ${SCHEMA_BUMP}+${GITNEXUS_PKG_VERSION}` (`:275`), and neither component changes in this diff — the cache key is byte-identical before and after.

**Consequence:** on a warm `parsedfile-cache`, byte-unchanged Python files replay pre-fix `ParsedImport`s with no flag; `isNamedReexport`'s strict `=== true` (`finalize-algorithm.ts:620`) takes the old path, the closure stays empty, and the whole fix is a **silent no-op on incremental analyze** — while every cold-run test in this PR passes. It lands hardest on `__init__.py` files: the rarest-changing, highest-cache-hit files in a Python repo, i.e. exactly the target. Results are *mixed* within one run (changed chunks resolve, unchanged ones do not), which reads as nondeterminism, and per the ledger's v25/v26 notes cached results replay verbatim **including across `--force`**. A published npm release does invalidate via `GITNEXUS_PKG_VERSION` — which is why this is P1 rather than P0 — but dev trees, `main`-HEAD installs, plugin-bundled runners and CI with a restored cache dir do not.

This is verbatim the failure mode the ledger immediately above `SCHEMA_BUMP` records for #2837 and #2807.

**Fix:** bump 45 → 46 with a ledger line naming `reexportsName`, and update the pin at `gitnexus/test/unit/incremental-parse-cache.test.ts:135`. Per that ledger's own warning, **re-check the value against `origin/main` immediately before merging, not at branch time** — three prior *exact-value* collisions are recorded, and #2840 is also in flight.

gitnexus-shared/src/scope-resolution/finalize-algorithm.ts
```diff
@@ -603,7 +636,7 @@ function buildReexportClosures(
     const drafts = edgeIndex.get(file.filePath);
     if (drafts !== undefined) {
       for (const d of drafts) {
-        if (d.source.kind !== 'reexport' && d.source.kind !== 'wildcard') continue;
+        if (!contributesReexportEdge(d.source)) continue;
```
magyargergo: **P2 · this widens the re-export sub-graph ~20× for Python, and it silently falsifies the algorithm docblock above it** *(both engines; perf measured, doc finding reached by 8 Claude lanes + Codex — the most-agreed item in this review)*

**Part 1 — cost.** Measured on real corpora, distinct (importer, imported-module) edges admitted by this guard:

| corpus | files | BASE | HEAD | ratio |
|---|---|---|---|---|
| CPython 3.11 stdlib | 664 | 41 | 839 | 20.5× |
| `dist-packages` | 396 | 0 | 703 | 0 → ∞ |

Repo-scale (12,400 files × 15 imports, 40 `localDefs`/file): **1,380 → 1,786 ms (+29%) and 200.9 → 228.9 MB (+28 MB)** at ~186k admitted imports. Linear and transient — a constant factor on one phase of a multi-minute `analyze`, which is why this is P2 and not higher.

**Explicitly refuted as scaling hazards** (so nobody re-litigates): SCC size alone — a **3,200-file single SCC shows zero delta** (145.0 → 148.9 ms), and the largest real Python sub-graph SCC measured is **12**; repeated `findExportByName` across fixpoint iterations (1.1-1.3×); Phase-3 fixpoint inflation; named-only chains (3-5×, near-linear).

**What is real:** `MAX_REEXPORT_DEPTH = 100` was removed when the input class was TypeScript barrels (`:585-588`) and is not reinstated for the much larger Python class, so two quadratics are bounded only by convention — Θ(D²) in `via` chain depth (20 names at D=1,600 → **1.14 s / 517 MB**) and Θ(S²) for a cyclic SCC whose chain tracks S (**654 ms / 262 MB at S=1,600**). And the wildcard branch at `:753-762` copies the target's *entire* closure, which was empty for Python pre-PR: a synthetic barrel of N=3000 names with K=2000 `from pkg import *` importers goes BASE 147 ms / 121 MB → HEAD **hard V8 OOM** under `--max-old-space-size=1536`. Rare in practice (67 wildcards / 664 stdlib files; 3 / 396 dist-packages) — but this repo has an OOM history at exactly this phase (#2649). Worth knowing: the depth bound and the narrow admission rule were introduced **together in one commit** (`fc919ad6`); this PR removes one half and leaves the other half removed.

**Asks:** (1) publish `finalize` wall-clock + peak RSS BASE vs HEAD on the **same 12.4k-file corpus already used for the PR body's edge counts** — that run could have produced them for free; (2) cap `via` at a fixed length with a truncation marker, which retires **both** quadratics at once; (3) optional but strictly better than BASE — index `localDefs` by simple name once per file instead of the linear `findExportByName` scan at `:722`, turning `O(E × D̄)` into `O(Σ D_f + E)`. Note no bench covers this pass at all, and `bench/python-scope --check` cannot fail on this change on either axis (its budget measures intra-file capture scaling; its fingerprint hashes capture `tag|text|range`, never a `ParsedImport`).

**Part 2 — this line makes the docblock above it false, and the helpers orphaned it.** The guard it replaces was a *deliberate* maintainer decision: commit `fc919ad6` (2026-04-25, direct to `main`) added `if (d.source.kind !== 'reexport' && d.source.kind !== 'wildcard') continue;` with the written rationale "regular/namespace/dynamic imports do not contribute." There is no review thread for it, so the commit message and this docblock are the **only** record of the rationale — which is why leaving it asserting the opposite costs more than usual comment drift.

Still-false text the PR did not update:
- `:552-556` — *"edges are `reexport` and `wildcard` drafts only (**regular imports do not contribute to the export surface**…)"*. Falsified by this line, and not marginally — 20× more edges.
- `:575-579` — *"For tree-shaped barrel graphs (the common case) it collapses to O(E_re)"*. Not the common case for Python: 10 cyclic SCCs in CPython stdlib, up from 0.
- `:580-588` — the O(depth²) `via` copy justified by *"practical **TypeScript** barrel chains are shallow."* Now carrying arbitrary Python chains; the conclusion survives (real `__init__` depth ≤ ~6) but the reason does not, and nothing enforces it.
- `:631-632`, seven lines above this guard — *"only resolvable reexport/wildcard targets contribute edges."*

Separately, inserting the two helpers at `:590-621` **orphaned** the 46-line docblock. AST-verified with `ts.getJSDocCommentsAndTags`: `contributesReexportEdge` → 1 block (line 590), `isNamedReexport` → 1 block (line 608), **`buildReexportClosures` → 0 blocks.** Hover/TypeDoc now show nothing for the SCC-condensation core, and the cross-reference at `:225` ("See `buildReexportClosures` for the algorithm") lands on an undocumented function.

**Fix (±0 LOC):** move both helpers below `buildReexportClosures` (hoisted declarations — call order is irrelevant) or above line 544, rewrite step 1 to *"`wildcard` drafts, `reexport` drafts, and `named`/`alias` drafts flagged `reexportsName`; for providers that set it (today: Python) the sub-graph is the file-level named-import graph, not a sparse barrel graph"*, and restate the complexity with the two quadratic corners named and the honest note that nothing bounds them.

gitnexus/test/unit/scope-resolution/finalize-algorithm.test.ts
```diff
@@ -403,6 +434,81 @@ describe('finalize', () => {
...
+
+    it('terminates when reexportsName named imports cycle (package __init__ cycle)', () => {
+      // Same bounded-fixpoint guarantee the explicit `reexport` cycle test
+      // relies on — the flag must not open a non-terminating path.
+      const c = file('c', [], [namedReexporting('X', 'X', 'b')]);
+      const b = file('b', [], [namedReexporting('X', 'X', 'c')]);
+      const a = file('a', [], [named('X', 'X', 'b')]);
+      const files = [a, b, c];
+      const out = finalize({ files, workspaceIndex: undefined }, defaultHooks(files));
+      expect(firstImport(out, a.moduleScope)!.targetDefId).toBeUndefined();
```
magyargergo, line 574: **P2 · this test is tautological — it passes under every possible implementation** *(cross-engine: `ce-testing`, Codex, and `test-ci-verifier` reached this independently)*

No file in this fixture (`a`, `b`, or `c`) has **any** `SymbolDefinition` in `localDefs` — there is no `def:X` anywhere in the graph for `targetDefId` to resolve to. So `toBeUndefined()` is true for the correct implementation, for a fully-reverted pre-PR implementation, and for an over-widened buggy one that ignores `reexportsName` entirely. Confirmed by direct execution against the worktree's real build.

It also does not guard termination in the way the comment claims: the cyclic-SCC loop is capped at `scc.files.length + 1` (`finalize-algorithm.ts:663-666`) **unconditionally**, so a logic bug in `contributesReexportEdge`/`isNamedReexport` cannot open a non-terminating path. Only the untouched Tarjan/fixpoint scaffolding could, and the pre-existing `reexport`-kind cycle test at `:512-528` already covers that.

Meanwhile the behaviour this *should* pin works today with zero coverage. Probe run against the worktree's own build:
```
c re-exports X from BOTH b (cyclic) and d (real def:d.X)
b re-exports X from c
a imports X from b
=> targetDefId: 'def:d.X', transitiveVia: ['b','c','d']
```
As written, the test cannot distinguish "the cyclic fixpoint correctly threads a definition out of the cycle" from "resolution inside a cycle is completely broken and always returns undefined."

**Fix:** add a fourth file `d` with a real def, reachable from inside the cycle, and assert `targetDefId` **and** `transitiveVia` land on it. Keep or rename the existing case if you still want an explicit absence-of-crash check.

(Two smaller test notes while you are in this file. The negative control at `:458` is fine but does **not** gate this diff — its fixture never sets `reexportsName`, so it passes byte-identically with the widening reverted; it *is* a real control against "the guard ignores the flag," which is the job the PR body claims for it, so keep it — just don't count it as coverage of the feature. And the alias-keying test at `:485-499` asserts `targetDefId` but not `transitiveVia` on its positive branch, unlike both sibling tests — a regression finding the right def via the wrong path, which is exactly the `localName`-vs-`importedName` keying bug that test exists to catch, would slip through. One extra `expect`.)

gitnexus/src/core/ingestion/languages/python/interpret.ts
```diff
@@ -52,23 +52,34 @@ export function interpretPythonImport(captures: CaptureMatch): ParsedImport | nu
...
+      // runtime namespace makes every module-level form republish, so the flag
+      // is set uniformly here rather than only for the redundant-alias case.
       if (sourceCap === undefined || nameCap === undefined || aliasCap === undefined) return null;
       return {
         kind: 'alias',
         localName: aliasCap.text,
         importedName: nameCap.text,
         alias: aliasCap.text,
         targetRaw: sourceCap.text,
+        reexportsName: true,
```
magyargergo: **P2 · reusability: the sibling site that answers the same question was not updated** *(the maintainer's explicit "check for reusability" ask — this is the concrete hit)*

The real location is `gitnexus/src/core/ingestion/languages/python/import-target.ts:125-139`, which is not in this diff — anchoring on the Python provider's other flag site.

`pythonFileExportsName` answers "does package P export name X?" from `localDefs` **only**:
```ts
return parsed.localDefs.some((def) => { … === importedName; });
```
It is consulted at `:70` to decide whether the package attribute exists before falling through to the submodule probe at `:74-84`, and the comment at `:59-63` states exactly the rule this PR now encodes in the closure — Python's `IMPORT_FROM` reads the package attribute first and only loads a same-named submodule when that attribute is absent.

Reproduced against the real resolver:
```python
# pkg/__init__.py   from .impl import log   (re-exports a FUNCTION named log)
# pkg/impl.py       def log(msg): ...
# pkg/log.py        def emit(msg): ...      (same-named submodule also exists)
# app.py            from pkg import log
```
→ `resolved target: pkg/log.py`. CPython binds `pkg.log` to the **function**; GitNexus targets the submodule, where `log` is not a local def either, so the edge ends `unresolved` and the closure this PR just built is never consulted — for exactly the case it was written for.

Not a regression (both halves behave as before), but the PR fixes one site and leaves its sibling disagreeing about what a package exports.

**Fix (3 lines, and it *reuses* the flag this PR added — which is the point):**
```ts
if (parsed.localDefs.some(...)) return true;
return parsed.parsedImports.some(
  (i) => (i.kind === 'named' || i.kind === 'alias') && i.reexportsName === true
      && i.localName === importedName,
);
```
If that behaviour change is out of scope for this PR, the minimum is one comment at `import-target.ts:125` naming the divergence and the issue that will close it. Two functions silently disagreeing about "does this file export X" is the expensive outcome.

(For completeness: Codex separately swept the *shared/generic* export-surface builders — `kind === 'reexport'` checks, wildcard checks, `expandsWildcardTo` — and found no other site needing the predicate. So the answer is clean: one Python-specific site missed, zero generic sites missed.)

magyargergo (commented):
## Multi-tool review digest — PR #2864 `fix(python): resolve calls through __init__.py re-exports`

**Method.** Two engines reviewed the diff (base `021ac303` → head `25eb474d`, 5 files, +193/−11) in a worktree, plus a coordinator pass that executed the contested claims. `[reproduced]` = executed; `[code-read]` = traced. Full evidence is in the inline comments; this is the summary.

| Engine | Attempted | Live | Notes |
|---|---|---|---|
| **A — Claude** | 6 GitNexus swarm lanes + 8 `ce-code-review` personas = 14 | **13** | Only `docs-DoD` produced nothing usable. |
| **B — Codex `gpt-5.6-sol`/`xhigh`** | 3 legs | **1** | Solo, `ce-code-review` and native `adversarial-review` **all died mid-run on the OpenAI org's 500k tokens-per-minute wall**, producing zero findings; one consolidated compiler-design leg replaced them and completed. |

**Engine B is one leg, not three.** "Cross-engine" means that single leg agreed. Agreement among Claude lanes is *consistency within one engine*, not independent confirmation.

---

## What is solid

The diagnosis is correct and this is a good PR. Verified, not assumed:

- **Root cause is right.** `from pkg.impl import X` in `pkg/__init__.py` lands in neither `localDefs` nor any closure, so `from pkg import X` elsewhere finds nothing. Reusing the closure machinery rather than forking a Python path is the right instinct.
- **Genuinely opt-in, does not leak.** `grep -rn reexportsName` hits only `python/interpret.ts` (2 sites), `finalize-algorithm.ts`, `types.ts`, tests. Three lanes enumerated all 15+ providers (c, cpp, cobol, csharp, dart, go, java, javascript, kotlin, php, ruby, rust, swift, typescript, vue): no other producer, no delegation, no spread carrying it out of Python. `[code-read]`
- **AGENTS.md layering rule NOT violated.** `ce-architecture`, `ce-project-standards`, `risk-architect` and Codex agree: the predicates branch only on `ParsedImport.kind` and a provider-set boolean. Naming Python in *prose* matches the file's precedent (`:120-121`, `:583`, `:816`). **No finding.**
- **Alias keying correct end to end.** `populateFileClosure` keys by `localName` (`:718`), looks up `importedName` (`:721`), so `from m import X as Y` publishes `Y`. Both engines verified; `finalize-algorithm.test.ts:483-497` pins both directions.
- **Closure cannot shadow a target's own `localDefs`.** `lookupReexportedName`'s one call site (`:511`) is reached only after `findExportByName` returns `undefined` (`:490`); the importer's own defs are protected by `materializeBindings` seeding `origin: 'local'` first.
- **Three of five new tests are genuine regression tests** — single-hop (`:445`), 3-hop (`:471`), alias-keying (`:485`) pin `targetDefId` to the one matching def and fail if the widenings are reverted.
- **The negative control does its stated job.** It does not gate this diff, but drop `&& reexportsName === true` from `isNamedReexport` and its assertion flips — a real control against "the guard ignores the flag." Don't count it as feature coverage.
- **Bench baselines correctly untouched.** All three fingerprint layers hash capture output or `resolvePythonImportTarget`; none observes a `ParsedImport`.
- **Right-sized mechanism.** No new union variant, hook, config flag or pass; `kind: 'named-reexport'` and a `reexportPolicy` hook were both correctly avoided. Not speculative — `swift/scope-resolver.ts:46` already records `@_exported import` as the same shape.
- **The PR body is unusually good** for a first-time contributor: failing scenario, mechanism, alternatives and why rejected, plus edge/flow deltas from a real 12.4k-file corpus.

---

## Headline findings

### H1 — `reexportsName` set with no scope context · **P1** · `[reproduced]`

**Both engines.** Codex #1 (P1); Claude `ce-api-contract`, `ce-project-standards`, `ce-architecture` A7, `risk-architect` F1, `ce-testing` F1 (P1), plus four lanes at P2/P3. **Reproduced three times independently** (coordinator; `ce-adversarial`; `ce-testing` via `pass3CollectImports`).

`interpret.ts:67`/`:82` flag unconditionally — `interpretPythonImport` gets only a `CaptureMatch` — so function-local, class-body, `try/except` and version-branch imports are all flagged, and scope is unrecoverable downstream (`scope-extractor.ts:897-910` applies no filter; `ImportEdgeDraft.fromScope` is hardcoded to `file.moduleScope` at `finalize-algorithm.ts:354, 382, 410`). With declaration-order first-wins (`:719`) this gives a **wrong answer on legal, running Python**:

```python
# pkg/__init__.py
def _lazy():
    from pkg.bad import X     # function-local — publishes NOTHING
from pkg.good import X        # module-level  — publishes pkg.X
# caller.py — CPython resolves pkg.good.X
from pkg import X
```
→ `RESOLVED = def:pkg.bad.X`. A purely function-local import likewise fabricates an edge for code that would `ImportError`.

**The repo's own docs misled this change.** `scope-extractor.ts:33-40` claims finalize *"reconstructs the owning scope via `provider.importOwningScope` during Phase 2."* It does not — `grep -rnE "\.importOwningScope\b" gitnexus/src/` returns **exactly one hit: that doc comment**, though 12+ providers implement the hook. The PR's fixture comment came straight out of that stale docblock, which deserves its own follow-up. Blast radius is capped (wildcard fan-out does **not** inherit this `[reproduced]`); `if TYPE_CHECKING:` is **not** a new gap (`python/index.ts:52-55`).

**Fix — larger than it looks**, since the scope info is not plumbed anywhere: either wire the hook in `pass3CollectImports` (**activating a dead hook with 12+ never-invoked implementations**) or carry a module-level marker on `ParsedImport` (which brings H3's `SCHEMA_BUMP` obligation). Deferring is defensible **if** this PR still renames the `python-fixtures.test.ts:277-283` case to a known over-approximation and drops the "module-level" assertion in the `types.ts` docblock. *Detail inline on `interpret.ts`.*

### H2 — Declaration-order first-wins is inverted for Python · **P1** · `[reproduced]`

**Both engines.** Codex #2 (P1); `ce-adversarial` F2 (P2, reproduced end-to-end), `ce-correctness`. `finalize-algorithm.ts:711-719` documents "declaration order first-wins" — safe for TS/Rust, where duplicate `export { X } from …` is a compile error. CPython **rebinds** (last write wins), and `if`/`else` and `try`/`except` pairs execute exactly one branch.

```python
# pkg/__init__.py
from .v1 import Client       # legacy, left behind
from .v2 import Client       # the actual public Client
```
→ `targetDefId: def:v1.Client`: every `from pkg import Client` is attributed to the **dead** implementation and `impact("Client")` points at the wrong file. A version-gated `if sys.version_info >= (3, 11)` pair resolves to the `>= 3.11` branch unconditionally — wrong def on working code for a repo targeting 3.9; `try:`/`except ImportError:` is the same shape. **Fix:** detect the collision instead of silently picking — drop the entry (ambiguous → unresolved, matching pre-PR) or let `pythonMergeBindings` arbitrate. *Detail inline on `finalize-algorithm.ts`.*

### H3 — `ParsedFile` shape changed with no `SCHEMA_BUMP` · **P1** · `[code-read]`

**Engine A only — Codex did not surface it — but coordinator-confirmed**, reached independently by four Claude lanes at P1. `parse-cache.ts:56-57` states the rule as a MUST and `parsedfile-store.ts:177` serializes the whole `ParsedFile`, so `reexportsName` is in that shape — yet `SCHEMA_BUMP` stays 45 (`:251`) and `PARSE_CACHE_VERSION` (`:275`) is byte-identical before and after. **On a warm `parsedfile-cache`, unchanged Python files replay pre-fix imports with no flag, the strict `=== true` at `:620` takes the old path, and the whole fix is a silent no-op on incremental analyze** — while every cold-run test passes. It lands hardest on `__init__.py`, the highest-cache-hit files in a Python repo, and cached results replay verbatim **including across `--force`**. A published npm release invalidates via `GITNEXUS_PKG_VERSION` (keeping this P1, not P0); dev trees, `main`-HEAD installs and CI with a restored cache dir do not.

**Fix:** bump 45 → 46 with a ledger line, update the pin at `incremental-parse-cache.test.ts:135`, and **re-check against `origin/main` at merge time, not branch time** — three prior exact-value collisions are recorded, and open PR **#2840 is still logged as claiming 44**. *Detail inline on `types.ts`.*

### H4 — Sub-graph widened ~20×; +29% `finalize` / +28 MB; the algorithm docblock now says the opposite · **P2** · `[reproduced]` + `[code-read]`

**Both engines.** Perf measured by `ce-performance` (tables below); the doc half reached by 8 Claude lanes plus Codex #6 — the most-agreed item in the review. The docblock's step 1 (*"regular imports do not contribute to the export surface"*), its complexity claim (*"tree-shaped barrel graphs (the common case)"*), its `via` justification (*"practical TypeScript barrel chains are shallow"*) and the comment at `:631-632` are all now false and were not updated. Inserting the helpers at `:590-621` also **orphaned** the 46-line docblock — AST-verified, `buildReexportClosures` → **0 jsdoc blocks**. The widened guard was a **deliberate maintainer decision** in `fc919ad6`, which removed `MAX_REEXPORT_DEPTH = 100` in the same commit. **Fix (±0 LOC):** move both helpers below `buildReexportClosures`; correct step 1 and the complexity section. *Detail inline on `finalize-algorithm.ts`.*

### H5 — Reusability: `pythonFileExportsName` still answers from `localDefs` only · **P2** · `[reproduced]`

**Engine A** (`ce-simplicity` S2 P1, `ce-adversarial` F5 P2 against the real resolver), coordinator-verified. Codex refuted "other *generic* export-surface builders need this predicate" after sweeping shared code — no conflict: this site is Python-provider-specific and that sweep would not reach it. `import-target.ts:125-139` answers from `localDefs` alone and is consulted at `:70` before the submodule probe; the comment at `:59-63` states exactly the rule this PR now encodes in the closure. With `pkg/__init__.py: from .impl import log`, `pkg/impl.py: def log`, and a same-named `pkg/log.py`, `from pkg import log` resolves to **`pkg/log.py`** where CPython binds the function — the edge ends unresolved and the new closure is never consulted, for exactly the case it was written for. Not a regression, but the PR fixes one site and leaves its sibling disagreeing about what a package exports. **Fix:** three lines that *reuse* the new flag, or one comment at `:125` naming the divergence. Read alongside open issue **#2882** — same import form, unfixed here. *Detail inline on `interpret.ts`.*

### H6 — Cycle-termination test is tautological · **P2** · `[reproduced]`

**Cross-engine** (`ce-testing` F2, Codex #7, `test-ci-verifier`). Detail under **Tests and CI** and inline on `finalize-algorithm.test.ts`.

---

## Performance

**Fundamentally clean.** The lane set out to find the quadratic this repo has a history of and **could not reproduce one at any realistic Python shape.** Sub-graph edges admitted by the widened guard, real corpora `[reproduced]`:

| corpus | files | `from` imports | wildcards | BASE edges | HEAD | ratio |
|---|---|---|---|---|---|---|
| CPython 3.11 stdlib | 664 | 2,170 | 67 | **41** | **839** | 20.5× |
| `/usr/lib/python3/dist-packages` | 396 | 3,767 | 3 | **0** | **703** | 0 → ∞ |

Repo-scale, flat acyclic workspace, 15 direct-hit named imports/file, 40 `localDefs`/file `[reproduced]`:

| files | BASE ms | HEAD ms | Δ | BASE heap | HEAD heap | Δ heap |
|---|---|---|---|---|---|---|
| 2,000 | 159.1 | 205.0 | +29% | 35.8 MB | 35.7 MB | ~0 |
| **12,400** | **1,380.4** | **1,785.5** | **+29%** | **200.9 MB** | **228.9 MB** | **+28.0 MB** |

**+405 ms and +28 MB at 186k admitted imports** — linear, transient, released when `finalize` returns. Driver: `populateFileClosure:722` runs `findExportByName`, a full linear scan of the target's `localDefs`, once per `from m import x` where pre-PR it ran zero times for Python.

**Refuted as scaling hazards:** SCC size alone (a **3,200-file single SCC shows zero delta**, 145.0 → 148.9 ms; largest real Python sub-graph SCC measured **12**); repeated `findExportByName` across fixpoint iterations; Phase-3 fixpoint inflation; named-only chains (3-5×, near-linear).

**What is real:** `MAX_REEXPORT_DEPTH = 100` was removed for TypeScript barrels and never reinstated, so two quadratics are bounded only by convention — Θ(D²) in `via` depth and Θ(S²) for a cyclic SCC whose chain tracks S. A synthetic barrel (N=3000 names, K=2000 `from pkg import *` importers) goes from BASE 147 ms / 121 MB to a **hard V8 OOM** under a 1.5 GB cap, which matters given this repo's OOM history at this phase (#2649).

**Asks:** (1) publish `finalize` wall-clock + peak RSS BASE vs HEAD on the **same 12.4k-file corpus already used for the PR body's edge counts**; (2) cap `via` at a fixed length with a truncation marker, retiring both quadratics; (3) index `localDefs` by simple name once per file — strictly better than BASE. No bench covers this pass, `bench/python-scope --check` cannot fail on this change on either axis, and `stats.largestSccSize` (`:304-306`) has no production consumer.

---

## Architectural fit

Direction right, layering rule clear. Three observations, none individually blocking:

1. **A per-language constant encoded as per-import parse-time data.** It rides every Python import record and is frozen into `parsedfile-cache/`, so **any future refinement becomes a cache-invalidating change** — which matters because the predicate is currently wrong (H1). `FinalizeHooks` already holds the right-shaped slot: `isNamespaceImport` (`:103-107`), wired at `run.ts:621-622` and **already implemented by Python** (`python/scope-resolver.ts:58`). A `reexportsImportedName?` hook is three wiring edits instead of one and adds no union field, no `SCHEMA_BUMP`, correct warm-cache behaviour, and a refinable predicate.
2. **The flag precedent is weaker than it looks.** `targetIncludesImportedName` (#2267) and `importedSymbolKind` (PHP) sit on the same variants, but both are **downstream consumer hints** read by the emitting language's own hooks (`php/import-target.ts:235`, `python/import-target.ts:400`). `reexportsName` is the **first flag on this union feeding back into the shared sub-graph algorithm**, raising the bar on type shape, docblock accuracy and `SCHEMA_BUMP`.
3. **The contract's stated rationale is disprovable.** `types.ts:141-143` says `kind: 'reexport'` "would drop the local binding." `materializeBindings` (`:954-965`) creates a module-scope `BindingRef` for **every** linked edge, re-export included — only `origin` differs — and `pythonMergeBindings` (`merge-bindings.ts:27-30`) puts `reexport`, `import` and `namespace` in the **same** `TIER_IMPORT`. The real reasons are sound: flipping `origin` lowers evidence weight 0.45 → 0.40, moves `ORIGIN_PRIORITY` 1 → 2, and misreports Python's parse-time syntax.

Two P3 notes. `:494`, `:520` and `:959` still test raw `kind === 'reexport'`, so single-hop `transitiveVia` stays `undefined` and `origin` stays `'import'` where TS gets `['impl']`/`'reexport'` — possibly correct, but unstated. And `isNamedReexport`'s intersection return type is runtime-sound (two lanes verified) but type-level loose; `Extract<ParsedImport, {kind:'named'|'alias'|'reexport'}>` is shorter and makes `s.kind === 'wildcard'` inside the guard a compile error (TS2367, compiler-verified).

---

## Over-engineering

**None in the mechanism.** The over-build is entirely prose: of 64 non-test added lines, **46 are comments and 16 are code** — 72% prose. The same paragraph is restated near-verbatim in **five** places (`types.ts:135`, `finalize-algorithm.ts:595`, `interpret.ts:55`, `finalize-algorithm.test.ts:102` and `:437`), the copies have already drifted (`pkg.X` vs `<this module>.x` vs `<module>.y`), and the disprovable rationale above appears in two of them — so a correction must be made twice. Meanwhile the two comments *nearest the changed guard* still assert the opposite. Keep one thorough paragraph in `types.ts:131-145` and make the rest one-line `@see` pointers: **~30 comment lines saved, and the file stops contradicting itself.** Smaller: `contributesReexportEdge` is a 2-statement helper with one caller and a 13-line docblock — defensible once that shrinks (Codex: "its placement, not its existence, causes the documentation issue"); and the builder `aliasReexporting` has one call site and hides `alias: localName`, the precise fact the alias test is about.

---

## Reusability

- **`pythonFileExportsName` is the concrete hit — see H5.** Its three-line fix *consumes* the flag this PR added, the strongest argument that the flag generalizes.
- **Codex swept the shared/generic sites and found nothing further.** With H5: one Python-specific site missed, zero generic sites missed.
- **A real second consumer is already documented** — `swift/scope-resolver.ts:46` (`@_exported import`). Not speculative generality.
- **But no other supported language would set it today.** Every "ordinary import republishes" shape is already modelled: C/C++ `#include`, Ruby `require`, Go dot-import → `wildcard`; TS `export … from`, CJS forwarding, Rust `pub use` → `reexport`; Java/Kotlin/C#/Swift/Dart/PHP imports are file- or scope-local. The new cell is **"per-name AND also binds locally"** — Python alone among the 16. The file's house style has a template (`:816`, "first known language where this matters in practice").
- **Nothing pins "exactly one provider sets this."** A shared test iterating providers' `interpretImport` over a canonical fixture would close it.

---

## Refuted suspicions

1. **"3 of 5 new tests FAIL — P0, code not built to `dist/`."** **Refuted `[reproduced]`, three ways.** The worktree's `gitnexus/node_modules` symlinks back to `/workspace/gitnexus/node_modules`, whose `gitnexus-shared` entry resolves to the **parent checkout on `main`** — an unrelated commit predating the feature. Coordinator: `grep -c reexportsName` on the resolved package → **0**; re-run with `resolve.alias` at the PR head → **36/36 pass**. `ce-testing`: `import.meta.resolve('gitnexus-shared')` prints `file:///workspace/gitnexus-shared/dist/index.js`; loading the worktree's own `dist` by absolute path → **all 5 pass**. `ce-project-standards`: root-caused the `tsc` errors to the same symlink. **Not a PR defect**; every verdict from that run is void. CI would not hit it (`setup-gitnexus/action.yml:19-22` builds `gitnexus-shared` from the PR branch first), but the integration suite inherits the trap.
2. **`gitnexus/` `tsc --noEmit` TS2353 failures.** Same stale-symlink artifact; `gitnexus-shared/` `tsc --noEmit` is clean, exit 0.
3. **Cross-language contamination.** Refuted by three lanes and Codex via exhaustive provider grep. Opt-in holds.
4. **AGENTS.md language-neutrality violation.** Cleared by four independent checks.
5. **`|SCC| + 1` cap is a merge blocker.** The defect is real and reproduced (an alias-rename ladder in a 2-file SCC silently drops a real edge from depth 7 on) — **but it reproduces identically with plain `kind: 'reexport'` drafts, so it pre-exists**; this PR only makes it reachable for Python. **Follow-up, credited to Codex, not a blocker.**
6. **Other predicate sites missed.** Refuted by Codex for *generic/shared* sites; the one real miss is Python-specific (H5). No conflict.
7. **"The negative control is worthless."** Partly refuted — it does not gate this diff but discriminates against "the guard ignores the flag." The *cycle* test is the tautological one (H6).
8. **Bench fingerprints will drift / need re-baselining.** Refuted definitively; all three layers the PR does not touch.
9. **Non-terminating fixpoint.** Could not be constructed; first-wins keeps the map monotone. The failure mode is starvation (item 5), not divergence.
10. **`transitiveVia` breaks a downstream consumer.** Refuted — produced and typed but **never read**; `emitImportEdges` dedups on `sourceFile->targetFile` and drops it.
11. **Wildcard fan-out inherits the false function-local re-export.** Refuted `[reproduced]` — wildcard edges materialize via `expandsWildcardTo` over `localDefs`, not the closure. Caps H1's blast radius.
12. **`gitnexus-shared` is independently published / semver-pinned.** False premise: `"private": true`, a `file:` dependency, and `build.js` inlines its `dist/` into `gitnexus/dist/_shared`.
13. **TS exhaustiveness / `SCHEMA_FINGERPRINT` / `RELATION_SCHEMA` / DDL gates broken.** Refuted — exhaustiveness follows the `kind` discriminant, not optional properties; those gates fingerprint DDL strings.
14. **Named-before-wildcard precedence diverges from CPython.** True but **not a defect**: it is exactly the precedence `pythonMergeBindings` already applies inside the file.

Two lower-severity survivors: a `named` import finalize reclassified to `kind: 'namespace'` still enters the closure because the guards test `draft.source.kind` while `tryFinalize` tests `draft.base.kind` (P3, one-line fix at `:712`); and `__all__` is unmodelled in ingestion, so the widened wildcard fan-out can republish `_`-prefixed names (P3, follow-up).

---

## Tests and CI

| # | Test | Sets the flag? | Fails if reverted? | Verdict |
|---|---|---|---|---|
| 1 | `resolves through a named import flagged reexportsName` (`:445`) | yes | **yes** | Genuine regression test |
| 2 | `leaves a plain named import out of the closure` (`:458`) | **no** | no | Control against "guard ignores the flag" — **not** feature coverage |
| 3 | `resolves a 3-hop chain` (`:471`) | yes | **yes** | Genuine regression test |
| 4 | `keys the closure by the published alias` (`:485`) | yes | **yes** | Genuine (driven by its first assertion) |
| 5 | `terminates when reexportsName named imports cycle` (`:501`) | yes | **no** | **Tautological** |

**The cycle test is tautological.** Its fixture gives no file any `SymbolDefinition`, so `toBeUndefined()` holds under every implementation — correct, reverted, or over-widened — and the cyclic loop is capped at `scc.files.length + 1` **unconditionally**, so a bug in the new predicates cannot open a non-terminating path. The behaviour it should pin works today with zero coverage: a probe where the cycle exits to a real def resolves to `def:d.X` via `['b','c','d']`. **Fix:** seed the cycle with a terminal def.

**`ci-tests.yml` and `ci-quality.yml` never ran on this PR.** Head sha `25eb474d9…` reports only "Apply conventional label" (pass) and "Validate PR title" (skipping); Vercel fails with "Authorization required to deploy" (fork-token/infra). `ci.yml:4` is `on: pull_request` with no `workflow_run`, and this is cross-repository, so GitHub's **fork-PR approval gate** blocks execution until a maintainer approves. **Every test and typecheck claim in the PR body is author-reported only, and no automated signal of any kind has covered this diff.**

**No integration or fixture test is in the diff.** The pre-existing `resolvers/python.test.ts:886-908` (`python-reexport-chain`) covers almost exactly this shape but **predates the PR** (`1afe9166a`, PR #238) and is untouched — so if this PR is what fixes it, the fix ships with no assertion pinning that fact; its current pass/fail could not be established here (symlink trap). The PR body's headline *"1 of 3 → 3 of 3 CALLS edges resolved"* could not be located anywhere in the repo, and the nearest fixture has 2 `it()` assertions, not 3 — a **missing-evidence point, not an accusation**; worth the author naming its source.

**Gaps worth closing here.** (1) No end-to-end Python coverage — all 5 tests use synthetic fixtures and `defaultHooks`, never `resolvePythonImportTarget` or a real `pkg/__init__.py`; pinning `python-reexport-chain` is cheapest. (2) No scope negative control — `ce-testing` specified the exact missing test, and **it would fail today, which is what makes it worth adding.** (3) Cycle test, above. (4) Nothing covers duplicate or conditional bindings of the same name (H2), local-def-vs-flagged-import precedence, or wildcard + flagged-named interaction. (5) No warm-cache test (H3); `fastapi-composed-route-constants.test.ts:167` is the pattern. (6) No scale guard. (7) Nothing pins the `ParsedImport` shape. (8) The alias test omits `transitiveVia` on its positive branch.

**Branch hygiene.** 8 commits behind `origin/main`; `997fc05b8` (#2855) touches both changed areas but is textually clean and orthogonal. Open PR **#2840** also edits `types.ts` **and is logged in the `SCHEMA_BUMP` ledger as still claiming 44**, so H3's merge-time re-check matters concretely. Open issue **#2882** is adjacent and unfixed (H5). Otherwise clean: one commit, conventional `fix(python):`, no workflow/manifest/lockfile changes, no binaries, no hidden-unicode or bidi. Recommend rebasing and running full CI before approval.

**Process context, factual and without judgement.** `clasalle-lila` has 0 merged commits here and opened #2864-#2868 as a batch; there is no linked issue and no PR body separate from the commit message, so the "+7,027 edges / 300→813 flows" and "`tsc --noEmit` clean" claims are self-reported. The Vercel failure and absent CI runs are **standard first-fork-contributor gating, not code defects.**

---

## Coverage limits

- **No full `analyze` on a real Python corpus.** The PR body's `+7,027 edges` is unverified here, and **the precision cost of H1/H2 is unquantified** — nobody measured what fraction of the new edges are false. Highest-value follow-up before merge. H1 was also not traced to `CALLS` emission, so read that P1 as *silent precision bug, real and reproducible*, not *guaranteed high-frequency false positive*.
- **H3 is traced end to end but not executed** — reproducing it needs two builds of the CLI.
- **Engine B is one leg**, so cross-engine corroboration exists for H1, H2, H4 and H6 but **not for H3 or H5**.
- **`docs-DoD` returned nothing usable** — no Definition-of-Done / docs-requirements lane ran.
- **Perf numbers are synthetic-harness measurements** driven through the real `finalize` (an exact A/B, since the only algorithmic delta is one boolean) plus real import-graph scans — not a real `analyze` run.

---

*Automated multi-tool digest (Claude, 13 live lanes; Codex `gpt-5.6-sol`/`xhigh`, 1 live leg; plus a coordinator verification pass). Several confident-sounding claims in the underlying lanes were wrong and are listed as refuted above — including a false P0 — please verify anything here against the code before acting on it.*

gitnexus/src/core/ingestion/languages/python/interpret.ts
magyargergo: Fixed in e6f7f8808.

You were right that there is no one-liner, and right about why. `CaptureMatch` is `{name, range, text}` — **no syntax node** — so `interpretPythonImport` structurally cannot see scope; and `pass3CollectImports` takes three args with no `positionIndex`/`scopeDrafts`, so it cannot recover one either without a signature change.

So the decision moved *up*, to `import-decomposer.ts`, which still holds the live `import_from_statement` node. It rides down as an `@import.publishes` marker and is computed **once per statement, not once per imported name**, reusing `findAncestorBeforeBoundary`. No dead hook activated, no shared-pipeline signature changed, no language named in shared code.

One correction to the finding, verified by running CPython 3.11: `try`/`except` and version-branch imports **do** publish. Python has no block scope — only `def` and `class` bodies suppress module publication, and a class-body import binds on the *class* (`C.X`). So the predicate is an ancestor walk for exactly those two node types. Both directions are now pinned, including a control asserting that `if`/`try`/`for` bodies still republish — that one would have caught an over-narrow fix.

Anchored at `spec.atNode`, never `stmtNode`: `anchorCaptureFor` picks broadest span with a strict `>`, so a statement-wide span would tie with `@import.statement` and let key order decide the anchor.

Also corrected the `scope-extractor.ts` docblock that caused this — it no longer claims finalize reconstructs scope via `importOwningScope`.

gitnexus-shared/src/scope-resolution/types.ts
magyargergo: Fixed in dc89710c1 — bumped to **60**, not 54.

54 was free against `origin/main` (53) but not against the in-flight claims: #2899 claims 54 and #2891 claims 59. Took the next value above the highest claim, which is what the five recorded exact clashes teach and what a literal pin structurally cannot detect.

Worth noting this qualifies **twice**, and the second way is the one that would have been missed: `reexportsName` is a `ParsedFile` serialization change, and the `@import.publishes` marker added for the scope fix moves the **capture** output too. The python-scope capture fingerprint drifted in CI and confirmed it independently.

Pin test updated. Re-check against `origin/main` at merge time still applies.

gitnexus-shared/src/scope-resolution/finalize-algorithm.ts
magyargergo: Fixed in 45af769c0 — took the drop-on-ambiguity option.

CPython 3.11, executed: sequential rebind is **last**-wins (`p2.v2`), but `try:`/`except ImportError:` is **first**-branch-wins. So neither direction is right on real code, which is what makes dropping the honest answer rather than the lazy one. The importer stays unresolved — exactly the pre-#2864 result — and the file-level IMPORTS edge is untouched, so only symbol-level attribution is withheld.

The one design constraint worth recording: **it has to be a pre-pass.** Computed from `edgeIndex` + `targetFile`, both frozen in phase 0, the set is constant across the fixpoint, so the closure map stays monotone, `myClosure.size > before` stays a valid progress signal, and `|SCC| + 1` keeps the meaning its comment claims. A set that grew mid-fixpoint would need retraction to propagate to files that had already inherited the name, and would break all three. Only two flagged drafts resolving to two *different in-workspace* files count — which also means the `try: import ujson / except: import json` shape never triggers it, since both resolve to `null`.

Checked in **both** loops, per your note: suppressing only the named loop lets a later `import *` refill the name.

Your parenthetical P3 was also real and I reproduced it end to end. `from . import logger` is emitted `named`, reclassified `namespace`, and was still admitted — with a `logger.py` holding a module-level `logger = logging.getLogger(...)`, importers bound to that **Variable** instead of the module. Both predicates now take the draft and test `base.kind`. No-op for TS/Rust: Python has the only `isNamespaceImport` implementation, and it is gated on `named`/`alias` anyway.

gitnexus/test/unit/scope-resolution/finalize-algorithm.test.ts
magyargergo: Fixed in 45af769c0 — the cycle now exits to a real def.

Fixture is `d` holding `def:d.X`, `c` re-exporting `X` from `d` **and** `Y` from `b` (so b↔c is a genuine cyclic SCC without colliding on `X`), asserting both `targetDefId` and `transitiveVia`. The straightforward version — `c` re-exporting `X` from both `b` and `d` — collides with the new ambiguity rule, so the cycle is formed on a second name instead.

Mutation-checked rather than assumed: reverting the widening fails 6 tests including this one; disabling the ambiguity guard fails 2; removing the namespace guard fails 1.

Also took both smaller notes — the alias test now asserts `transitiveVia` on its positive branch, and the negative control is kept as a control and not counted as feature coverage.

One thing you should sanity-check: assertions here are exact (`toEqual`, `toHaveLength`), per the maintainer rule that tests never use `toBeLessThan`/`toBeGreaterThan`. That forced the `transitiveVia` cap to have a fully determined shape rather than a bounded one.

gitnexus-shared/src/scope-resolution/finalize-algorithm.ts
magyargergo: Fixed in 45af769c0 (docs + perf) and 90582a623 (guard). All three asks taken.

**Docs.** Helpers moved below `buildReexportClosures`, so the 46-line docblock is no longer orphaned. Step 1, precedence and complexity rewritten — they claimed regular imports do not contribute, justified the via copy by TypeScript barrels being shallow, and asserted a first-wins precedence that is not universal. The two quadratic corners are now named with the honest note about what bounds them.

**Ask (3), the `localDefs` index** — done, memoized on array identity inside `findExportByName`, so all four call sites benefit including the two pre-existing rescans in the phase-3 fixpoint.

**Ask (2), cap `via`** — done, at 32. `ImportEdge.transitiveVia` has **zero production readers** (only tests and its own type), which is what makes truncation safe. This changed a pre-existing 1000-hop test that asserted the full chain.

**Ask (1), publish BASE vs HEAD.** 5 rounds, one isolated process per measurement, rotating variant order:

| corpus | PR head | fixed |
|---|---|---|
| via-chain depth 400 | 67 ms / 145 MB | **25 ms / 40 MB** |
| py packages ~9k files | 212 ms | 203 ms |
| wide fan-in | 127 ms | 117 ms |

A caveat on my own numbers: my **first** harness ran all variants in one process and was unusable — BASE moved 164→124 ms between runs and the ordering flipped, reporting my own optimization as a 32% regression. Isolated processes reversed the conclusion. Anyone re-running this should not measure variants in-process.

**Guard added** (`bench/finalize-reexport`), since you noted nothing covered this pass. Its depth arm is an **exact structural assertion** — longest emitted `transitiveVia` must equal 32 — not a ratio. I tried a `depth_ratio` timing arm first and it was a bad gate: capped scored 2.71–3.52 and uncapped 5.87–7.65 across repeated samples, so one uncapped run passed the budget. A gate that passes a third of the time on a broken build gets read as evidence. Structural form fails 3/3 at 401 vs 32. `width_ms` stays as a loose timing arm, because a structural check cannot see a restored linear scan.

gitnexus/src/core/ingestion/languages/python/interpret.ts
magyargergo: **Took the documented-divergence option, not the code fix — because the suggested three-line fix is incorrect.** Leaving this thread open for your call. Comment landed in dc89710c1.

The proposed predicate is `reexportsName === true && i.localName === importedName`. That is also true for `from . import log`, and those two forms bind different things in CPython. Verified on 3.11:

```
pkg/__init__.py: from .impl import log   ->  from pkg import log  ==  <function log>  (pkg.impl)
pkg/__init__.py: from . import log       ->  from pkg import log  ==  <module pkg.log>
```

For the second shape the fallback would return `pkg/__init__.py` and kill the correct namespace edge to `pkg/log.py`. So it trades your miss for a different one — and that second shape is precisely what open issue **#2882** is about, so the two fixes would collide.

Separating them needs the re-export's own resolved target, i.e. re-entering `resolvePythonImportTarget` with a different `fromFile`. That is a real change with cycle risk on `__init__` chains, and the discriminator belongs with #2882 rather than bolted on here.

Also worth flagging for whoever picks up #2882: the fixture at `python-import-target-parity.test.ts:172-186` is `as unknown as ParsedFile` with **no `parsedImports` field**. It passes today only because the `localDefs` check short-circuits first. Any fallback added after that check throws `TypeError` on it unless it guards with `?? []`.

Your stated minimum is what shipped: a comment at the divergence naming both the shape and the issue that closes it. Not a regression either way — both halves behave exactly as before #2864.

PR #2865, opened 2026-08-07
fix(routes): connect decorator routes to their handler function

clasalle-lila:
## Why

A Route node's only relationship is `HANDLES_ROUTE` from its **file**. So the graph knows a route exists and which file declared it, but not which function actually implements it — the handler's only edge is `DEFINES`.

Two consequences, both measured on a repo with 162 FastAPI routes:

- Every decorated handler is indistinguishable from dead code. Nothing reaches it, so a reachability query calls it unreferenced even though the framework invokes it on every request.
- `route_map` / `api_impact` can only answer at file granularity, and `processes.ts` sends every route through its `routesWithoutHandlerByFile` fallback instead of keying by handler.

## What Changed

Two halves of one gap. Both were already designed for; neither was wired.

**1. `handlerName` was never set for decorator routes.** `ExtractedDecoratorRoute.handlerName` is documented as *"captured at extraction where the decorated definition node is in hand"*, and `resolveRouteHandlerSymbols` already consumes it to stamp `handlerSymbolId`. But only the Spring extractor ever set it — so FastAPI, Flask and NestJS all got `undefined`, and 0 of 162 routes carried a handler. A route decorator's parent *is* the decorated definition, so the name is right there: `decoratedDefinitionName` walks to it. It climbs consecutive decorators so stacked forms (`@router.get(...)` over `@requires_auth`) still resolve, caps the climb so a malformed tree can't loop, and returns `undefined` rather than guessing — the routes phase already treats a missing name as "fall back to file-level".

**2. With a symbol resolved, there's finally something to point an edge at.** So this also emits a definition-level `HANDLES_ROUTE` alongside the file-level one. The sibling decorator overlay already does exactly this — `pipeline-phases/tools.ts` anchors `HANDLES_TOOL` on the definition the decorator sat on, not its file. Routes were the outlier.

These are one PR rather than two because the edge is inert without the symbol. Emitted from a branch missing the first half it produces zero edges, since `handlerSymbolId` is empty — I found that out the hard way.

**On direction:** Function → Route, not Route → Function. It matches how every other overlay attaches (`MEMBER_OF` → Community, `STEP_IN_PROCESS` → Process, `HANDLES_TOOL` → Tool — the symbol is the source). It also keeps this free of schema risk: `Function|Route` is already declared by the ATTACHMENT rule in `lbug/schema.ts`, which that file documents as deliberate headroom for this case. `Route → Function` would have needed a new hand-listed pair, and an undeclared pair aborts `analyze` outright — a failure `schema.ts` records having hit four separate times.

## How To Review

1. `workers/parse-worker.ts` — `decoratedDefinitionName` and where it's threaded into the emitted route
2. `pipeline-phases/routes.ts` — the definition-level edge

## Verification

FastAPI fixture, edges 9 → 11:

```
api.py        (File)    -> GET /widgets, POST /widgets   [unchanged]
list_widgets  (line 10) -> GET  /widgets                 [new]
create_widget (line 15) -> POST /widgets                 [new]
```

On the 12.4k-file repo: **161 of 162 routes now resolve to their handler, up from 0.** The single abstention is `uniqueSymbolId` correctly refusing to guess where the name isn't uniquely resolvable in its file.

`npx tsc --noEmit` clean; schema-pair coverage and route suites pass (196 tests).

## Risks

Additive — the file-level edge stays, so both existing consumers are unaffected. `group/extractors/http-route-extractor.ts` types its query `(handlerFile:File)`. `manifest-extractor.ts` matches an untyped `(handler)` but takes `LIMIT 1` ordered by `handler.id`, and `File:…` sorts before `Function:…`, so its selected row doesn't change either.

I did not run the full suite on this one — targeted suites only. Worth a full run before merge since it touches edge emission.

---

## Summary

*This appears to be a focused route extraction and ingestion change with broad transitive reach. The implementation is concentrated in the pipeline that connects Python decorator routes with their handler functions, while the affected file risk remains low.*

**🟠 HIGH blast radius. A Python HTTP route ingestion change that connects decorator routes to handler functions through `gitnexus/src/core/ingestion/pipeline-phases/routes.ts`, with impact reaching 16 graph dependents.**

The main review focus is `gitnexus/src/core/ingestion/pipeline-phases/routes.ts`, which contains `handlerPathFor`, `routesPhase`, `execute`, `scanContent`, `addRoute`, and `preSeed`, and is the hottest changed file. Trace those changes into `gitnexus/src/core/ingestion/route-extractors/python-decorator-handler.ts`, `gitnexus/src/core/group/extractors/http-route-extractor.ts`, and `gitnexus/src/core/ingestion/languages/python.ts`, where decorator route recognition and handler linking are defined.

The impact spans the `Group`, `Workers`, `Route-extractors`, and `Extractors` modules and three affected flows. The changed integration and unit tests cover route tables, composed FastAPI routes, route graph edges, incremental parsing, and decorator handler names. No files have HIGH or CRITICAL file risk, so the HIGH blast level comes from the graph reach rather than an individually high risk file.

_Added by GitNexus for PR #2865. Edit freely — this block is replaced on the next review, everything above it is left untouched._


magyargergo (commented):
## Multi-tool review digest — PR #2865

**Method.** Two engines, not five independent confirmations. **Claude** ran the GitNexus PR swarm (branch-hygiene, docs/DoD) plus seven `ce-code-review` personas (correctness, adversarial, performance, api-contract, project-standards, testing, maintainability). **Codex** (gpt-5.6-sol, xhigh) ran an adversarial pass and a scoped perf/architecture pass. Two further Codex legs (swarm, ce) died on an OpenAI rate limit after emitting only a preamble and contributed **no findings** — nothing below is sourced from them. A coordinator pass re-executed the contested claims; where it overturned a lane, the lane text is superseded and that is called out inline.

---

### What this PR gets right

The description is unusually good, and three of its own risk claims **verify TRUE**:

- **Schema safety holds.** `Function|Route` and every sibling really are pre-declared by `DEFINITION_ANCHOR_LABELS × ATTACHMENT_TARGET_LABELS` (`lbug/schema.ts:549-551`; `Route` ∈ `ATTACHMENT_TARGET_LABELS` at `:388-396`). Executing the real DDL generator gives **450 pairs**, and all 24 `SYMBOL_NODE_LABELS` have a declared `X|Route`. `assertDeclaredPair` cannot abort. `[reproduced]`
- **The file-level edge really must stay.** `http-route-extractor.ts:48` is typed `MATCH (handlerFile:File)`, so the new edge is invisible to it and its row counts are unchanged. `[code-read]`
- **The `manifest-extractor` risk analysis is correct for every case live today** — though for a different reason than one lane gave (see § Lower-priority).

Also verified clean: `npx tsc --noEmit` exit 0; `prettier --check` on both files clean; targeted route suites green (`resolve-route-handler-symbols` 8/8, `route-parse-skip` 7/7, `expo-routes`, `route-process-linking`, `java-spring-route-ingestion`, `api-file-route`); single-domain diff, no unrelated churn; `BEHIND` is benign (no commit on `main` since the merge-base touches either file); conventional-commit title correct; `CHANGELOG.md` correctly untouched. The Python path of the helper is **correct for every valid-syntax shape probed** — module-level, class methods, nested classes, nested functions, `async def`, stacked decorators. And the `ENTRY_POINT_OF` precision flip is a genuine design improvement (it fixes the `GET /x` ↔ `POST /x` cross-wiring that `processes.ts:117-124` describes).

---

### Headline findings

#### 1. P1 — `SCHEMA_BUMP` not bumped: on a warm parse cache the entire feature is a silent no-op

**Corroboration: STRONG (cross-engine).** Codex adversarial — its *only* reported finding — plus four Claude lanes (ce-adversarial, ce-project-standards, ce-api-contract, swarm docs-dod).

**Trigger.** Any repo whose parse cache was written by a build with the same `SCHEMA_BUMP` + package version. `handlerName` is new **parse-time** output on `ExtractedDecoratorRoute`; `slimParseWorkerResultsForCache` (`parse-cache.ts:447-464`) strips only `calls`/`assignments`/`constructorBindings`/`parsedFiles`/`skippedPaths`, so `decoratorRoutes` is persisted verbatim. Chunk keys hash raw file bytes, so every unchanged Python chunk is a HIT; the replayed entries carry no `handlerName` → `call-processor.ts:331` yields `undefined` → the new `if (handlerSymbolId)` block at `routes.ts:324` **never fires**. `parse-impl.ts:1043` feeds the same `applyChunkResults` absorber as the live-worker path, and `parse-impl.ts:879-881` is what reads `decoratorRoutes` back. Survives `--force` (`parse-cache.ts:20-22`). `[code-read]`

`SCHEMA_FINGERPRINT` does **not** rescue it — it is a sha256 over the DDL (`lbug/schema.ts:697`) and this PR declares no new pair, so it is byte-identical. (The branch-hygiene lane's "no schema version bump required" is correct about the *fingerprint* and must not be used to dismiss this — it is a different lever.)

**Coordinator correction on the fix value.** The branch reads `SCHEMA_BUMP = 45` (`parse-cache.ts:251`) only because it is **BEHIND**; `origin/main` is already at **48** (`git show origin/main:gitnexus/src/storage/parse-cache.ts` → line 303). The naive "bump to 46" would be an **exact clash with a value main already shipped** — the fourth such clash the file's own ledger records. **After rebasing, the value must be ≥ 49**, re-checked against `origin/main` at merge time per `parse-cache.ts:223-224`. `[reproduced]`

The PR's headline measurement ("161 of 162 routes resolve") was necessarily taken on a cold or invalidated cache. On partial warmth the graph becomes cache-warmth-dependent: two `analyze` runs on identical source produce different graphs, with no error. Precedent for the fix: #2391 bumped **and** added `test/integration/fastapi-composed-route-constants.test.ts:167` ("warm parse-cache (#2391 SCHEMA_BUMP)") for the same struct.

#### 2. P2 — Populating `handlerName` converts a memoized per-FILE query into an uncached per-ROUTE full scan

**Corroboration: STRONG (cross-engine).** Codex perf-arch finding 1 and Claude ce-performance, reproduced independently by the coordinator.

**Trigger.** `gitnexus group sync` / the `group_sync` MCP tool on a FastAPI/Flask repo. Before this PR no Python decorator route carried a `handlerSymbolId` (exactly what `http-patterns/python.ts:1139-1146` asserts), so the whole population took the `else` branch, which reaches the **memoized** `loadFileSymbols` (`http-route-extractor.ts:430-443`, backed by `fileSymbolCache`). After it, essentially all of them take the fast path at `http-route-extractor.ts:688`, which calls `await db(CONTAINING_QUERY, { filePath })` at **:698, inside the per-row loop, uncached** — `loadFileSymbols` is a closure local to a different method and is not passed into `extractProvidersGraph`. `[code-read]`

`CONTAINING_QUERY` (`:68-88`) is a 3-way `UNION ALL` over `Function`/`Method`/`CodeElement` filtered by `sym.filePath`, with **no index on `filePath`**. Measured against the real 391 MB index: **4.80 ms even for a filePath matching zero rows**, proving the cost is a full table scan, and `executePrepared` re-prepares on every call. Cost goes from `O(files × symbols)` to `O(routes × symbols)`; one 100-route FastAPI file goes from 1 query to 101. Extrapolated: ~**+3 s** for a 500-route service on an 11k-symbol index, ~**+30 s** at 10× symbol count. `[reproduced]`

No compensating win: the parse-skip that justifies the fast path is gated on `routeCoverage === 'complete'` (`:772`), and `python.ts` is unchanged at `'partial'` — so **+(R−1) uncached queries per file, 0 parses saved**. Fix is three lines and local to the unchanged file: memoize by `filePath` inside `extractProvidersGraph`, or lift `loadFileSymbols` to an instance member.

#### 3. P2 — `decoratedDefinitionName()` returns the enclosing CLASS name for TypeScript — **latent, not live**

**Corroboration: STRONG (cross-engine).** Codex perf-arch finding 4 plus six Claude lanes, each reproducing it independently against the real `tree-sitter-typescript` grammar.

**Root cause.** In `tree-sitter-typescript` a method decorator is a **sibling** of `method_definition` under `class_body`, not its parent — there is no `decorated_*` wrapper. So the walk finds no `definition` field on `class_body`, no `name` field either, climbs to `class_declaration`, and the bare-`name` fallback at **`parse-worker.ts:1378`** fires and returns the class. Reproduced: `@Get('list')` and `@Post('create')` in a NestJS controller both → `"UsersController"`. `[reproduced]`

**Coordinator correction — this is NOT live today, and it demotes the finding from P1 to P2.** Running the verbatim `TYPESCRIPT_QUERIES` decorator pattern (`tree-sitter-queries.ts:478-481`) and applying the emit condition at `parse-worker.ts:1774-1793`:

```
--- TypeScript / NestJS ---   @Controller  arg_str=false arg_expr=false => dropped
                              @Get         arg_str=false arg_expr=false => dropped
--- Python / FastAPI ---      @get         arg_str=true                 => pushed
```

`TYPESCRIPT_QUERIES` has no `arg_str`/`arg_expr` captures (`PYTHON_QUERIES` gained them in #2391), so every TS decorator route is dropped and **the wrong name is computed and discarded**. No wrong `handlerSymbolId`, no wrong `Class -> Route` edge, no `manifest-extractor` row flip today. `[reproduced]`

It is a **latent trap**: it arms the moment anyone widens `TYPESCRIPT_QUERIES` with `arg_str`/`arg_expr` — the natural next step, and precisely what is needed to deliver the NestJS support the PR body claims. Two consequences worth fixing in-PR regardless:

- The JSDoc at `:1352-1353` — *"Returns undefined rather than guessing when the shape is unfamiliar"* — is **false as written**: for the TS shape it guesses, and guesses the enclosing class.
- The PR body's *"FastAPI, Flask and NestJS all got `undefined`"* is inaccurate for NestJS for a different reason: NestJS decorator routes are not extracted by this path **at all**.

The repo already owns the correct traversal — `http-patterns/node.ts:279-295` documents this exact grammar fact (*"child of class_body before a method_definition"*) and solves it with a forward-sibling `findDecoratedMethod()`.

**Architecture, same root cause (cross-engine: Codex perf-arch #2 + Claude ce-project-standards/correctness/maintainability).** `AGENTS.md:42` and `DoD.md` §2.2/§6: *"Shared code in `gitnexus/src/core/ingestion/` must not name languages — plug language behavior in via `LanguageProvider` / `ScopeResolver` hooks."* `:1343` names Python; `:1369` keys on `definition`, the field name in tree-sitter-python's `decorated_definition`; `:1376-1380` is an unnamed-grammar heuristic. The sanctioned seam **already exists and is already wired in this same file**: `LanguageProvider.extractDecoratorRoutes` (`language-provider.ts:284-298`) is invoked at `parse-worker.ts:2970`, and Java uses it (`languages/java.ts:213` → `route-extractors/spring.ts:193`) to set **this very field** from the method node's own `name`. `decorated_definition` appears in five files; the other four are all under language-scoped paths.

Note also that the loop's stated rationale is empirically false: stacked decorators do **not** nest in tree-sitter-python — they are siblings inside one `decorated_definition`, so `depth === 0` always hits the `definition` field and the climb's only observable effect is the wrong TS answer.

#### 4. P2 — Live Python shapes where the helper returns a wrong-but-*unique* name

**Corroboration: MODERATE (Claude only — ce-correctness, ce-adversarial), but reproduced against the real grammar.**

Every wrong name found is unique in its file, so `defs.length === 1` (`call-processor.ts:294`) passes and an edge is minted with `confidence: 1.0`:

- **Commented-out handler, valid Python, `hasError=false`.** `@app.route("/old")` above a commented-out `def old():` binds to the *next* definition → `Function:new_handler -[HANDLES_ROUTE]-> Route(/old)`. `[reproduced]`
- **Decorator over a non-`def` target inside a class body** (ERROR recovery) → returns the enclosing **class** name → `Class:…:A -[HANDLES_ROUTE]-> Route`. `[reproduced]`
- **`@api.route("/todos")` on `class TodoList(Resource)`** (flask-restx, valid syntax) → returns a Class, contradicting the helper's stated function/method contract. `[reproduced]`

Pre-PR these produced only a file-level edge, which was still literally true. The PR promotes benign file-level imprecision into a confident, specific claim about the wrong symbol. `CONTAINING_QUERY` has arms for `Function`/`Method`/`CodeElement` only — **no `Class` arm** — so a Class-sourced uid never resolves a display name downstream.

#### 5. P2 — Three comments the codebase uses as *narrowing arguments* are now false; one is in the diff

**Corroboration: STRONG (six Claude lanes; Codex did not probe comments).**

- **`lbug/schema.ts:340` and `:357-358`** — *"HANDLES_ROUTE sources `generateId('File', handlerPath)`, a literal `File`, so every non-`File` anchor is headroom."* False after this PR; the anchor is now a `lookupExactAll` result of unconstrained label. This matters beyond hygiene: that same docblock records **four reproduced `analyze` hard-aborts** and states their shared cause — *"Every one of the four aborts above came from re-narrowing a set to what one predicate looked like it allowed."* Leaving it stale arms abort number five for whoever prunes `Route` back to a hand-declared `File|Route`.
- **`http-patterns/python.ts:1139-1146`** — *"FastAPI decorator routes set no handlerName … Declaring 'complete' now is only a latent trap for the moment a follow-up gives FastAPI routes a handlerName."* **This PR is that follow-up.** The premise is now false; `routeCoverage: 'partial'` stops being a no-op and becomes load-bearing, and the comment no longer says so.
- **In-diff, `routes.ts:315-317`** — *"This mirrors the sibling decorator overlay: `pipeline-phases/tools.ts` already anchors HANDLES_TOOL on the definition … Routes were the outlier."* Verified: `tools.ts:96` is `td.handlerNodeId ?? generateId('File', td.filePath)` — **one** edge, either/or — plus an existence guard at `:47-48` (`ctx.graph.getNode(td.handlerNodeId)`). This PR emits **two** edges with **no** guard. The cited precedent argues *against* the design it is used to justify. The real justification is already stated correctly at `routes.ts:307-308`.

#### 6. P2 — The PR's headline "dead code" motivation is not delivered as shipped

**Corroboration: MODERATE (four Claude lanes), independently verified by the coordinator.**

`routes.ts:310-313` justifies the edge with *"Every decorated handler then looks exactly like dead code…"*. Two facts defeat that:

- **`HANDLES_ROUTE` is not in `impact()`'s default traversal.** With no explicit `relationTypes`, impact defaults to `['CALLS','IMPORTS','EXTENDS','IMPLEMENTS','USES','METHOD_OVERRIDES','OVERRIDES','METHOD_IMPLEMENTS']` (`local-backend.ts:5791-5802`, verified). `HANDLES_ROUTE` appears exactly once in the entire `src/mcp` tree — in `VALID_RELATION_TYPES` (`:376`), the allowlist gating *explicit* filters. It is in no default `impact()` or `context()` list. `[code-read]`
- **The edge is OUTGOING from the handler** (`routes.ts:327-328`). A handler that "has nothing reaching it" still has nothing reaching it; it now has one more thing it *reaches*.

A default `impact({target: 'list_widgets', direction: 'upstream'})` is byte-identical before and after. The benefit is real but narrower than claimed: **explicit** graph queries can now walk Route↔handler at definition granularity, a genuine improvement over file granularity. Note also that `Route.handlerSymbolId` already existed pre-PR, so route→handler lookup at property level was already possible — the edge's marginal value is graph-traversability, not the reachability fix. The direction argument and the schema argument in the comment are each correct on their own terms; the gap is that neither direction closes the dead-code symptom while the type stays out of the default relTypes.

#### 7. P2 — Unmeasured `ENTRY_POINT_OF` flip: some Python routes can go from a best-effort link to **zero**

**Corroboration: MODERATE (Claude ce-adversarial; ce-performance read the same mechanism as a net *win*), verified by the coordinator.**

`processes.ts:146-147` buckets each route by whether the Route node carries a `handlerSymbolId`, and the consumer at `:176-178` is `exactRouteKeys ?? fallbackRouteKeys` — a **`??`, not a union** (verified). Pre-PR the whole Python population sat in `routesWithoutHandlerByFile` and picked up a best-effort edge from any process whose entry point lived in the same file. Post-PR ~all move to `routesByHandlerId`, where an edge mints only if some process's `entryPointId` is *exactly* the handler symbol. A route whose handler is not itself a detected process entry point is now in **neither** lookup and gets **zero** `ENTRY_POINT_OF` edges.

This is intended, not accidental — the PR lists it as a benefit and the exact keying is the more precise answer. **The gap is verification, not design:** the flip trades recall for precision on a population that has never taken this path, and the PR measured only handler resolution (161/162), never the `ENTRY_POINT_OF` count before vs after. On the author's own 12.4k-file / 162-route repo that is a two-number check.

---

### Lower-priority

- **P2 / follow-up issue — `@mock.patch` phantom routes now amplified to symbol granularity.** `'patch'` is in `ROUTE_DECORATOR_NAMES` (`parse-worker.ts:1318`) and the branch gates on decorator **name** only — `decoratorReceiver` is captured but never used to filter — so `@mock.patch("app.db.session")` already parses as `PATCH /app.db.session`. **Pre-existing.** What this PR changes is blast radius: the phantom now resolves a `handlerSymbolId` and mints a definition-level edge onto the decorated *test function*. In a Python repo with a normal test suite this is plausibly the single largest source of new edges. `[reproduced]` Cheapest guard: gate on `routePath.startsWith('/')`, the same check `express_route` already applies at `:1873`. Worth a follow-up issue, not a block on this PR.
- **P3 — `manifest-extractor.ts:242-246` untyped `MATCH (handler) … LIMIT 1`: safe today, by id-spelling coincidence.** One lane concluded "safe" via a wrong premise (that symbol ids are lowercase `def:`-prefixed — that is a parse-cache scope key, not a graph node id). Corrected against the **live** index: graph ids are `` `${Label}:${name}` `` with an uppercase prefix (`generateId`, `lib/utils.ts:1`), which is also what `rel-pair-routing.ts:64-68` requires. Ordering is `Class: < File: < Function: < Method:`, so `ORDER BY handler.filePath ASC, handler.id ASC LIMIT 1` still picks the `File:` row for any `Function`/`Method` handler — **the PR's own risk analysis is correct**. `[reproduced]` A `Class:`-prefixed source *does* displace it; that is reachable live today only through the narrow class-attached Python shapes in finding 4, and across the whole NestJS surface once finding 3 arms. One-word fix if the intent is file granularity: `MATCH (handler:File)`, mirroring `http-route-extractor.ts:48`.
- **P3 — dead work on the TS path.** `decoratedDefinitionName` is called at `:1765`, *before* the arg-shape gate at `:1774-1793` that drops every TS route, so every `@Get`/`@Post`/… in a NestJS codebase pays an upward CST walk whose result is discarded. Cost is genuinely small (measured ~2.8 µs/call) but it is free to fix by moving the call inside the two consuming branches.
- **P3 — hand-rolled structural type instead of `SyntaxNode`.** `:1355-1365` declares `{ parent?: unknown; type?: string }`, then casts `as N` to a second locally-redeclared shape — while `SyntaxNode` is already imported at `:109`, used in ~25 signatures in this file including the immediately preceding sibling `checkForFetchCall(node: SyntaxNode)` at `:1277`, and is the static type the call site already holds (`captureMap: Record<string, SyntaxNode>`, `:1700`). `type?: string` is never read. The unchecked cast is *why* `childForFieldName('definition')` on a TypeScript node type-checks and silently returns `null` — the type system was capable of catching finding 3.

---

### Explicitly probed and REFUTED

These were the named attack surfaces. Each is a result, not an omission.

| Hypothesis | Result |
|---|---|
| Undeclared `X\|Route` pair hard-aborts `analyze` | **Refuted** `[reproduced]` — 450 pairs; all 24 `SYMBOL_NODE_LABELS` covered. The PR's claim holds. |
| `HANDLES_ROUTE_QUERY` returns a second row per route | **Refuted** `[code-read]` — `http-route-extractor.ts:48` is typed `(handlerFile:File)`. |
| Files flip into the `route-parse-skip` set | **Refuted** `[reproduced]` — only `php.ts:137` declares `routeCoverage: 'complete'`, and PHP has no `@decorator` capture. Python/Java stay `'partial'`. |
| Edge-id collision / duplication on re-index | **Refuted** `[code-read]` — distinct id prefixes; `addRelationship` idempotent by id; incremental writeback `DETACH DELETE`s by `filePath` and both endpoints share it. |
| `api_impact` / `route_map` / `shape_check` double-count | **Refuted** `[code-read]` — `fetchRoutesWithConsumers` never traverses `HANDLES_ROUTE`; it reads `Route.filePath` and joins `FETCHES` only. |
| Default `impact()` / `context()` / `detect_changes` fan-out widens | **Refuted** `[code-read]` — `HANDLES_ROUTE` is in no default list; default BFS is byte-identical. |
| A CI benchmark fingerprint drifts / needs re-baseline | **Refuted** `[reproduced]` — every `--check` bench audited; none reaches this path. `bench/emit-persistence` synthesizes its graph internally and never creates a `Route`. `bench/schema-pairs` pair count unchanged at 450. |
| The CST walk is a hot-path cost | **Refuted — noise** `[reproduced]` — ~3.7 µs/call, 1–2 iterations (the depth-8 cap never binds). 10,000 route decorators ≈ **37 ms total across the entire analyze**. |
| Extra graph rows / CSV / COPY volume | **Refuted — noise** `[reproduced]` — ≤1 row per *resolved* route. A 1,000-route FastAPI monolith against a 300k-edge index is **+0.33 %**, ~200 KB of CSV, under a single 8 MB buffer flush. |
| Pipeline-internal traversal corruption | **Refuted** `[code-read]` — `graph-emit-sink.ts:130-132` already documents `HANDLES_ROUTE` as deliberately never read back mid-pipeline. |
| Realistic Python parse-error triggers for the fallback | **Mostly refuted** `[reproduced]` — merge-conflict markers, Python-2 `print`, mixed tabs, stray dedents, unbalanced parens, PEP-695 aliases, Jinja-in-`.py` all either bound correctly or stopped matching the query. Only the shapes in finding 4 reach it. |
| Title / description / CHANGELOG / lint conventions | **Compliant** — conventional-commit title, all required description sections incl. a self-disclosed validation gap, correctly no `CHANGELOG` edit, no `no-explicit-any` violation, prettier clean. |

---

### Test gap

`+80/-0`, **zero test files**, and nothing in the existing suite would fail if either half of the change were deleted:

- `test/unit/resolve-route-handler-symbols.test.ts` hardcodes `handlerName` into synthetic objects — it passes identically with or without this PR, and would pass if the helper always returned the wrong string. `[reproduced]`
- `route-handler-symbol-roundtrip.test.ts` / `route-method-roundtrip.test.ts` seed the edge via raw Cypher and never execute `routes.ts:324-333`.
- `http-route-extractor.test.ts` *"extracts NestJS controller decorators"* tests **Strategy B** (`http-patterns/node.ts`'s own, separate, *correct* walker) — do not count it as coverage for this PR. `[reproduced]`
- Every pipeline test that walks `HANDLES_ROUTE` filters to `sourceNode.label === 'File'`, so all are structurally blind to the new edge; `expo-routes.test.ts:35` counts edges but expo routes never resolve a handler.

Missing, in priority order: (1) a CST-level unit test on `decoratedDefinitionName` for both grammars — the template is `test/unit/spring-route-extractor-parity.test.ts`, ~50 lines, and it would have caught finding 3 immediately (the helper is an unexported local `const`, so this needs an export); (2) a warm-parse-cache regression modelled on `fastapi-composed-route-constants.test.ts:167`, which is what would have caught finding 1; (3) a pipeline assertion that a `HANDLES_ROUTE` row exists whose `sourceId` resolves to a `Method`/`Function` — not `File`, not `Class`; (4) an `ENTRY_POINT_OF` before/after count for a FastAPI fixture; (5) an executor-invocation count on the fast path, which would guard the finding-2 memoization fix. Per repo convention: exact `toBe`/`toEqual`, no `if`-branching in the vitest.

The PR body's *"I did not run the full suite — targeted suites only"* is compliant with `DoD.md` §4's disclosure clause, but `DoD.md` §2.7 ("tests cover the real changed path") is unmet for an edge-emission change.

---

### CI at review time

**Vercel fails on authorization only** — not a code signal. Every other check was **pending**.

---

*Automated multi-tool digest (Claude swarm + `ce-code-review` personas; Codex gpt-5.6-sol adversarial + perf/architecture). Verify each item against the code before acting.*

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -1715,12 +1762,14 @@ const processFileGroup = (
           const httpMethod = ['GET', 'POST', 'PUT', 'DELETE', 'PATCH'].includes(method)
             ? method
             : 'GET';
+          const handlerName = decoratedDefinitionName(decoratorNode);
```
magyargergo, line 1765: 🔴 **P1** — This is a parse-time emission change, so `SCHEMA_BUMP` (true location: `gitnexus/src/storage/parse-cache.ts:251`) must be bumped or the whole feature is a silent no-op on any warm cache — `decoratorRoutes` is persisted verbatim by `slimParseWorkerResultsForCache` and replayed at `parse-impl.ts:879-881` with no `handlerName`. The branch reads 45 only because it is BEHIND; `origin/main` is already at **48**, so after rebasing this must land at **≥49**, not 46 — 46 would be an exact clash with a shipped value.

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -1715,12 +1762,14 @@ const processFileGroup = (
...
           const base = {
             filePath: file.path,
             httpMethod,
             decoratorName,
             lineNumber: decoratorNode.startPosition.row + lineOffset,
             ...(decoratorReceiver ? { decoratorReceiver } : {}),
+            ...(handlerName ? { handlerName } : {}),
```
magyargergo, line 1772: 🟠 **P2** — **Scope: `gitnexus group sync` / the `group_sync` MCP tool only — plain `analyze` is unaffected.** Populating `handlerName` flips the entire FastAPI/Flask route population from the memoized `loadFileSymbols` path onto the uncached per-route `CONTAINING_QUERY` at `http-route-extractor.ts:698` (true location), turning `O(files × symbols)` into `O(routes × symbols)` — one 100-route file goes from 1 query to 101, ~+3 s per 500 routes on an 11k-symbol index. No offsetting parse-skip, because `python.ts` is still `routeCoverage: 'partial'`. Three-line fix: memoize by `filePath` inside `extractProvidersGraph`, or lift `loadFileSymbols` to an instance member.

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -1336,6 +1336,53 @@ const ROUTE_DECORATOR_NAMES = new Set([
...
+    const definition = cursor.childForFieldName?.('definition') ?? null;
+    if (definition) {
+      const nameNode = definition.childForFieldName?.('name') ?? null;
+      const text = nameNode?.text;
+      if (typeof text === 'string' && text.length > 0) return text;
+      return undefined;
+    }
+    // Some grammars expose the definition as a direct named child rather than
+    // through a `definition` field.
+    const nameNode = cursor.childForFieldName?.('name') ?? null;
```
magyargergo, line 1378: 🟠 **P2** — This bare `name` fallback performs no shape check on the ancestor it reads. In tree-sitter-typescript a method decorator is a SIBLING of `method_definition` under `class_body`, so the walk climbs to `class_declaration` and returns the enclosing CLASS — `@Get('list')` → `"UsersController"` (reproduced). Inert today only because `TYPESCRIPT_QUERIES` lacks the `arg_str`/`arg_expr` captures the emit gate at :1774-1793 requires, so it arms the moment anyone widens that query for NestJS. It is also live for Python ERROR-recovery and class-attached shapes (flask-restx `@api.route` on a class), where it returns a Class that `CONTAINING_QUERY` has no arm for. Suggest deleting the branch or gating it on the ancestor being definition-shaped; the repo's correct forward-sibling walk already exists at `http-patterns/node.ts:279-295`.

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -1336,6 +1336,53 @@ const ROUTE_DECORATOR_NAMES = new Set([
   'PatchMapping',
 ]);
 
+/**
+ * Name of the function/method a route decorator is attached to, for
+ * `ExtractedDecoratorRoute.handlerName`.
+ *
+ * A decorator's parent is the `decorated_definition` (Python) /
```
magyargergo, line 1343: 🟠 **P2** — This helper encodes one grammar's shape in shared ingestion code — `definition` is tree-sitter-python's field name — against AGENTS.md:42 / DoD.md §2.2 ("shared code in `core/ingestion/` must not name languages; plug behavior in via `LanguageProvider`"). The sanctioned seam is already wired in this same file: `LanguageProvider.extractDecoratorRoutes` (`language-provider.ts:284`) is invoked at :2970, and Java uses it (`languages/java.ts:213` → `route-extractors/spring.ts:193`) to set this exact field. Two claims in this JSDoc are also empirically false: TS has no `decorated_*` wrapper, and stacked decorators do not nest in tree-sitter-python (they are siblings inside one `decorated_definition`, so `depth === 0` always wins) — so the climb's only observable effect is the wrong TS answer. ":1352 Returns undefined rather than guessing" is untrue: for the TS shape it guesses, and guesses the class.

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -1336,6 +1336,53 @@ const ROUTE_DECORATOR_NAMES = new Set([
...
+ * cannot stamp `handlerSymbolId`, and every decorated handler ends up with no
+ * incoming edge — indistinguishable from dead code even though the framework
+ * invokes it.
+ *
+ * Walks up through consecutive decorators so stacked decorators
+ * (`@router.get(...)` above `@requires_auth`) still find the definition.
+ * Returns undefined rather than guessing when the shape is unfamiliar; the
+ * routes phase already treats a missing name as "fall back to file-level".
+ */
+const decoratedDefinitionName = (decoratorNode: {
```
magyargergo, line 1355: 🔵 **P3** — `SyntaxNode` is already imported at :109 and used in ~25 signatures here, including the immediately preceding sibling `checkForFetchCall(node: SyntaxNode)` at :1277 — and the call site already holds one (`captureMap: Record<string, SyntaxNode>`, :1700). Declaring `{ parent?: unknown; type?: string }` and casting `as N` to a second local shape discards that, forces dead optional-chain guards, and leaves `type?: string` unread. The unchecked cast is why `childForFieldName('definition')` on a TypeScript node type-checks and silently returns `null` — the type system could have caught the class-name bug.

gitnexus/src/core/ingestion/pipeline-phases/routes.ts
```diff
@@ -300,6 +300,37 @@ export const routesPhase: PipelinePhase<RoutesOutput> = {
...
         });
+
+        // Definition-level HANDLES_ROUTE, emitted alongside the file-level edge
+        // whenever the handler symbol resolved.
+        //
+        // Why both: the file-level edge is what `http-route-extractor.ts` reads
+        // (its query is typed `(handlerFile:File)`), so it stays. But a FILE does
+        // not handle a route — a function does, and without an edge to it the
+        // handler carries no relationship at all beyond `DEFINES`. Every
+        // decorated handler then looks exactly like dead code even though the
```
magyargergo, line 311: 🟠 **P2** — This motivation is not delivered as shipped. `HANDLES_ROUTE` is absent from `impact()`'s default relation types (`local-backend.ts:5791-5802`: CALLS/IMPORTS/EXTENDS/IMPLEMENTS/USES/METHOD_OVERRIDES/OVERRIDES/METHOD_IMPLEMENTS) and from every `context()` list — it appears once in all of `src/mcp`, inside `VALID_RELATION_TYPES`, the allowlist for *explicit* filters. And the edge is outgoing from the handler (:327-328), so a handler with nothing reaching it still has nothing reaching it. A default `impact(direction: 'upstream')` on a handler is byte-identical before and after. The real (genuine) win is definition-granularity traversal for explicit Cypher / `relationTypes:['HANDLES_ROUTE']` queries — worth saying that instead, or adding the type to the default set in a follow-up.

gitnexus/src/core/ingestion/pipeline-phases/routes.ts
```diff
@@ -300,6 +300,37 @@ export const routesPhase: PipelinePhase<RoutesOutput> = {
...
+        // framework invokes it, and blast-radius analysis from a route stops at
+        // file granularity.
+        //
+        // This mirrors the sibling decorator overlay: `pipeline-phases/tools.ts`
```
magyargergo, line 315: 🟠 **P2** — This precedent is mis-cited. `tools.ts:96` is `const handlerId = td.handlerNodeId ?? generateId('File', td.filePath)` — ONE edge, definition-or-file, never both — and it guards node existence first at `tools.ts:47-48` (`ctx.graph.getNode(td.handlerNodeId)`). This block emits two edges with no guard, so the sibling argues against the design it is quoted to justify. The honest justification is already stated correctly at :307-308.

gitnexus/src/core/ingestion/pipeline-phases/routes.ts
```diff
@@ -300,6 +300,37 @@ export const routesPhase: PipelinePhase<RoutesOutput> = {
...
+        // already anchors HANDLES_TOOL on the definition the decorator sat on,
+        // not on its file. Routes were the outlier.
+        //
+        // `Function|Route` and its siblings are already declared by the
```
magyargergo, line 319: 🟠 **P2** — The pair-coverage claim in this comment is correct — but the behaviour change it describes makes the docs it cites stale. `lbug/schema.ts:340` and `:357-358` say HANDLES_ROUTE sources `generateId('File', handlerPath)`, "a literal File, so every non-File anchor is headroom" — after this change the anchor is a `lookupExactAll` result of unconstrained label. That same docblock records four reproduced `analyze` hard-aborts and names their shared cause as re-narrowing a set based on exactly this kind of rationale, so leaving it stale arms abort number five. Same applies to `http-patterns/python.ts:1139-1146`, whose "latent trap for the moment a follow-up gives FastAPI routes a handlerName" names *this PR* — update both here, per the repo's fix-in-PR convention.

gitnexus/src/core/ingestion/pipeline-phases/routes.ts
```diff
@@ -300,6 +300,37 @@ export const routesPhase: PipelinePhase<RoutesOutput> = {
...
+        // ATTACHMENT rule in `lbug/schema.ts`
+        // (`DEFINITION_ANCHOR_LABELS × ATTACHMENT_TARGET_LABELS`), which that
+        // file documents as deliberate headroom for exactly this case — so no
+        // new pair is needed and `assertDeclaredPair` cannot abort on it.
+        if (handlerSymbolId) {
```
magyargergo, line 324: 🟠 **P2** — No existence guard, unlike the `tools.ts` precedent cited above — `tools.ts:47-48` is `td.handlerNodeId && ctx.graph.getNode(td.handlerNodeId) ? … : undefined`, i.e. it refuses an id the graph does not hold. Worth matching for defence-in-depth rather than because a dangling id was demonstrated: `relPairKeyFor` derives the endpoint label from the id PREFIX rather than from a live node, so a dangling `handlerSymbolId` would route cleanly into the `Function|Route` pair file and surface only at `COPY` — and `rel-pair-routing.ts:265-271` documents that path as one the per-edge fallback drops silently. A one-line `ctx.graph.getNode(handlerSymbolId)` check closes it. Separately: stamping `handlerSymbolId` moves the whole Python route population from `routesWithoutHandlerByFile` into `routesByHandlerId` in `processes.ts:146-178`, where the consumer is `exactRouteKeys ?? fallbackRouteKeys` (`??`, not a union) — a route whose handler is not itself a detected process entry point now gets ZERO `ENTRY_POINT_OF` edges instead of a best-effort one. That is intended and more precise, but unmeasured; a before/after `ENTRY_POINT_OF` count on your 162-route repo is a two-number check.

gitnexus/src/core/ingestion/workers/parse-worker.ts
magyargergo: > `SCHEMA_BUMP` must be bumped or the whole feature is a silent no-op on any warm cache.

Addressed in `9132e932e`: the parse-cache schema moves from 88 to 90 (89 is already claimed by #3128), with the ledger and exact pin updated. A serialized all-hit warm-cache regression now verifies `handlerName`, `handlerSymbolId`, and the definition-level `HANDLES_ROUTE` edge survive replay.

gitnexus/src/core/ingestion/workers/parse-worker.ts
magyargergo: > Populating `handlerName` flips the route population onto the uncached per-route `CONTAINING_QUERY`.

Addressed in `9132e932e`: `extractProvidersGraph` now caches `CONTAINING_QUERY` results by file path, including failed lookups as an empty result. Invocation-count tests verify multiple resolved routes in one file perform one symbol query while preserving handler names and fallback UIDs.

gitnexus/src/core/ingestion/workers/parse-worker.ts
magyargergo: > The bare `name` fallback performs no shape check on the ancestor it reads.

Addressed differently in `9132e932e`: the shared ancestor-walking helper is removed. The generic worker passes its real `SyntaxNode` to a language-provider hook; Python accepts only an immediate `decorated_definition` containing a `function_definition`, so TypeScript and class-attached/error shapes cannot borrow an enclosing class name. Real-tree-sitter tests cover functions, methods, stacks, async handlers, and rejection paths.

gitnexus/src/core/ingestion/workers/parse-worker.ts
magyargergo: > This helper encodes one grammar's shape in shared ingestion code.

Addressed differently in `9132e932e`: the shared ancestor-walking helper is removed. The generic worker passes its real `SyntaxNode` to a language-provider hook; Python accepts only an immediate `decorated_definition` containing a `function_definition`, so TypeScript and class-attached/error shapes cannot borrow an enclosing class name. Real-tree-sitter tests cover functions, methods, stacks, async handlers, and rejection paths.

gitnexus/src/core/ingestion/workers/parse-worker.ts
magyargergo: > Declaring a second local shape discards the `SyntaxNode` already held at the call site.

Addressed differently in `9132e932e`: the shared ancestor-walking helper and its unchecked local casts are removed. The generic worker passes its real `SyntaxNode` to a typed language-provider hook; Python owns the direct `decorated_definition` shape and returns only function names, with real-tree-sitter coverage for valid and rejected forms.

gitnexus/src/core/ingestion/pipeline-phases/routes.ts
magyargergo: > The real win is definition-granularity traversal for explicit Cypher / `relationTypes:['HANDLES_ROUTE']` queries.

Addressed in `9132e932e`: the comment now states only that explicit graph-traversal benefit and the file-edge compatibility requirement. The unsupported default-impact/dead-code claim is removed; behavior remains additive and the handler endpoint is now graph-validated before the definition edge is emitted.

gitnexus/src/core/ingestion/pipeline-phases/routes.ts
magyargergo: > This precedent is mis-cited: `tools.ts` emits one edge, definition-or-file, never both.

Addressed in `9132e932e`: the incorrect `HANDLES_TOOL` precedent is removed. The rationale now documents this route-specific additive contract—retain the typed file edge for existing extraction consumers and add a graph-validated definition edge for explicit definition-granularity traversal.

gitnexus/src/core/ingestion/pipeline-phases/routes.ts
magyargergo: > This behavior change makes the docs it cites stale.

Addressed in `9132e932e`: the schema invariant now documents both File and resolved Function/Method HANDLES_ROUTE sources, corrects unreachable headroom from 47 to 45 pairs, and keeps the deliberately broad cross-product safety rationale. Python's `routeCoverage` comment now states the active data-loss reason it remains partial: imperative Flask `add_url_rule` providers still lack ingestion parity.

gitnexus/src/core/ingestion/pipeline-phases/routes.ts
magyargergo: > No existence guard, unlike the `tools.ts` precedent cited above.

Addressed in `9132e932e`: handler lookup now resolves once through `ctx.graph.getNode`; a missing node emits only the file-level edge and omits both the `handlerSymbolId` stamp and definition edge. Tests cover live Function/Method nodes, dangling IDs, and data-route-table path fallback. For the separate `ENTRY_POINT_OF` observation, the exact-key-over-fallback behavior is intentional and remains covered by `route-process-linking.test.ts` in the passing combined suite; no union change was made. The external 162-route corpus was not available for a count comparison.

gitnexus/src/core/ingestion/pipeline-phases/routes.ts
```diff
@@ -397,6 +397,37 @@ export const routesPhase: PipelinePhase<RoutesOutput> = {
...
```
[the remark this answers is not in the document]
magyargergo: > The added relationship changes HANDLES_ROUTE sources from only the `File` ID to the resolved handler definition.

Addressed in `9132e932e`: the schema invariant now documents both File and resolved Function/Method HANDLES_ROUTE sources, corrects unreachable headroom from 47 to 45 pairs, and uses this additive edge as the example for why the deliberately broad attachment cross-product must remain broad.

magyargergo:
> The route-handler change is conceptually sound, but the parse/cache and graph-consumer contracts need to move with it.

Resolved in `9132e932e`:
- invalidated and regression-tested warm-cache replay of decorator `handlerName`;
- memoized group-sync symbol lookup per file;
- moved Python decorator-definition lookup behind `LanguageProvider`, with strict real-AST shape tests;
- graph-validated handler IDs before stamping routes or emitting definition edges;
- corrected the traversal rationale, schema pair documentation, and Python partial-coverage invariant;
- added route-phase, data-table, group-sync, and warm-cache regressions.

The separate `ENTRY_POINT_OF` exact-key-over-fallback behavior remains intentional and is covered by the passing `route-process-linking` suite; no union change was made. The review's unrelated pre-existing `@mock.patch` deduplication observation remains follow-up scope rather than being folded into this route-attribution fix.

Validation: 118 targeted tests passed, `tsc --noEmit` passed, Prettier passed, and ESLint reported no errors. The full suite could not complete in this WSL container: the first run hit repeated unrelated worker-ready timeouts and the timeout-adjusted retry stalled, so both were stopped after capturing the environment failure.

PR #2867, opened 2026-08-07
feat(walker): opt into indexing hidden paths via GITNEXUS_INDEX_HIDDEN

clasalle-lila:
## Why

`walkRepositoryPaths` passes `dot: false` to glob, so hidden (dot-prefixed) paths are never enumerated. For a repo that keeps first-party source under a dot-directory, that means almost nothing gets indexed — and the failure is silent. `analyze` reports success, and `context()` then returns "not found" for symbols whose file has no node at all.

I hit this on a repo where 1,661 of 1,708 tracked Python files live under a dot-directory. The index came back clean and covered 47 of them.

`.gitnexusignore` negation can't recover those paths either. Negation only un-filters `DEFAULT_IGNORE_LIST` entries, and that check runs *after* enumeration — a path glob never yielded can't be re-included. That's a reasonable thing to assume works, so it's worth stating plainly that it doesn't.

## What Changed

`GITNEXUS_INDEX_HIDDEN` as an opt-in.

The default stays `false` deliberately. `dot: false` is the primary mechanism keeping `.git/`, `.vscode/` and `.venv/` out of the index, with `DEFAULT_IGNORE_LIST` as defense-in-depth behind it — the code comment says as much and it's correct. Flipping the default would change the indexed file set for every existing repo. `DEFAULT_IGNORE_LIST` still applies when enabled, so `.git/` and `node_modules/` stay excluded regardless; callers narrow the rest with `.gitnexusignore`.

The less obvious half: the setting changes *which files the walker enumerates*, so an index built with it on isn't interchangeable with one built with it off. I found this while testing — enabling the flag on an existing index **silently no-ops**, because the walker never re-runs. Same shape of bug as the original. So it's registered as an analysis feature (`walker.index-hidden`), which makes the toggle part of index identity and forces the rebuild the new scope requires. Both directions are covered, since `findAnalysisFeatureMismatches` treats a missing key and an extra key alike.

## How To Review

1. `core/ingestion/utils/index-hidden.ts` — follows the `GITNEXUS_MAX_FILE_SIZE` pattern (`warnOnce`, banner-returns-null-on-default, test-only reset)
2. `filesystem-walker.ts` — the one-line change plus the comment explaining the default
3. `run-analyze.ts` — feature registration

## Verification

Two-file fixture (`.hidden_src/buried.py`, `visible/plain.py`):

```
default             -> 3 nodes, only visible/plain.py
=1, no manual clean -> "missing:walker.index-hidden"    -> rebuild -> 6 nodes, both
back to default     -> "unexpected:walker.index-hidden" -> rebuild -> 3 nodes
```

`.git/` stayed excluded throughout (6 nodes, not thousands). 14 new unit tests; `npx tsc --noEmit` clean.

## Risks

Off by default, so no behavior change unless opted in. Enabling it on a large repo will grow the index substantially — on the repo above, 1,263 → 12,411 files — so the README row notes that `DEFAULT_IGNORE_LIST` still applies and `.gitnexusignore` is the tool for the rest.

README env-var table row added per CONTRIBUTING.


azizur100389:
Reviewed at `bd8d8ed081d53e5d267916e0d79b477ec2addc12`.

**High: preserve dotenv exclusions before enabling hidden enumeration.** With `GITNEXUS_INDEX_HIDDEN=1`, an otherwise unignored `.env.staging` or `.env.development.local` passes the finite denylist. The [new dot option](https://github.com/abhigyanpatwari/GitNexus/blob/bd8d8ed081d53e5d267916e0d79b477ec2addc12/gitnexus/src/core/ingestion/filesystem-walker.ts#L59) admits it, and File CSV generation persists its full text even without a supported parser. Earlier `dot:false` blocked this entry path.

Please cover `.env.*` in default exclusions and add real walker/persistence tests using fake secret sentinels, plus off/on/off indexing coverage. Current tests cover flag/banner behavior only. Locally reproduced (2026-09-02, Windows, Node 24.19.0), using fake sentinels only: the real walker admits both dotenv variants with the flag enabled, and production CSV/graph loading persists their complete marker text in real LadybugDB `File.content`. Both are absent with the flag disabled. Five focused flag/ignore/walker/CSV suites pass (**222 passed, 2 skipped**), as does CLI typechecking, so the supplied suites miss this path. The off/on/off reproduction used separate fresh DBs; same-index incremental-toggle coverage is still needed. Rebase and required CI are also needed.

PR #2868, opened 2026-08-07
feat(graph): expose isTestCode on definition nodes

clasalle-lila:
> **Depends on #2866** — the shared predicate this derives from is created there. This branch contains that commit; review #2866 first and this diff shrinks to one commit.
>
> **This changes the graph DDL**, so `SCHEMA_FINGERPRINT` changes and every existing index rebuilds once. Flagging that up front since it's the main thing to weigh — see Risks.

## Why

Test-file classification is only reachable from inside the codebase. A Cypher consumer — the `cypher` MCP tool, or an agent reasoning over the graph — has no way to ask "is this test code?", so it ends up reimplementing the path heuristic inside the query.

That reimplementation is fragile in exactly the way the duplicated predicate in #2866 was: it drifts from the real definition, silently, with nothing covering it. I know because I wrote one, in Cypher, and got a materially wrong answer from it before noticing.

Concretely, this is what makes the reachability question unanswerable from outside. "Does this function have a caller, or is it a framework entry point, or is it test code, or is it genuinely unreferenced?" is the question you actually want, and only the graph can answer it cheaply.

## What Changed

`isTestCode BOOLEAN` on the five definition-bearing node tables (Function, Class, Interface, Method, CodeElement — the ones already carrying `isExported`), derived centrally in the CSV writer from each node's `filePath` so it can't be forgotten at an emit site.

**On the name:** `isTestCode`, not `isTestEntryPoint`. The predicate classifies the *file*, so it's true for a helper inside a test file as well as for the case the framework invokes. Telling those apart needs name patterns (`test_*`, `Test*`) and is a separate question — calling this "entry point" would overstate what the value means.

## How To Review

1. `lbug/schema.ts` — the DDL columns
2. `lbug/csv-generator.ts` — header and the central derivation
3. `lbug/lbug-adapter.ts` — the COPY column lists (see below)
4. `test/unit/node-column-parity.test.ts` — the guard

## Verification

The whole reachability picture, in Cypher, with no path heuristics, on a 12.4k-file repo:

```
lang    bucket                  n
go      has_caller           1231
go      framework_entrypoint    5
go      test_code            2965
go      UNEXPLAINED            38   (0.9%)
python  has_caller           4495
python  framework_entrypoint  183
python  test_code            6477
python  UNEXPLAINED           398   (3.4%)
```

The residual is what *should* stay non-zero — genuinely unreferenced code plus dynamic dispatch static resolution can't see. That's the dead-code signal, and driving it to zero would destroy it.

`npx tsc --noEmit` clean; 27 tests across the two schema-parity suites, 417 across schema/lbug/csv.

## Risks

**One-time reindex for every user.** The DDL change bumps `SCHEMA_FINGERPRINT`. Verified that the existing guard detects it and forces the re-analyze automatically — no user action, no stale column — but it is a rebuild, and it's the reason this is a separate PR from #2866 rather than folded into it. #2866 is a behavior-neutral bug fix that should be able to land without waiting on this decision.

**A note on the third column list.** A node column lives in three places that must agree: the DDL, the CSV header/row builder, and the explicit `COPY` list in `getCopyQuery`. I updated the first two and got:

```
COPY failed for Function: Number of columns mismatch. Expected 8 but got 9
```

thrown at load time, *after* a full parse had already run. So `node-column-parity.test.ts` now pins every table's COPY list to its DDL, in order, and asserts the DDL parse is non-vacuous so a broken regex can't make it pass silently. `spring-bean-schema.test.ts` already guarded this for `Class` alone — its expectation is updated and the new test generalizes it.

Happy to drop this PR if the classification is considered a query-side concern rather than something the graph should carry. The argument for carrying it is that every external consumer otherwise reinvents a heuristic that has already drifted once inside the codebase.

### DIFF
--- a/gitnexus/src/core/ingestion/entry-point-scoring.ts
+++ b/gitnexus/src/core/ingestion/entry-point-scoring.ts
@@ -13,6 +13,7 @@
 import { detectFrameworkFromPath } from './framework-detection.js';
 import { SupportedLanguages } from 'gitnexus-shared';
 import { providers } from './languages/index.js';
+import { isTestFilePath } from './utils/test-file-path.js';
 
 // ============================================================================
 // NAME PATTERNS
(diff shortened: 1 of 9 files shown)


azizur100389:
Reviewed at `da488a37a0fb2a649de1a9f511191728b27f538a`.

**Local validation update (2026-09-02, Windows, Node 24.19.0):** the supplied column-parity suite passes **24/24**, focused CSV/predicate/schema/deletion suites pass **54/54**, and CLI typechecking passes. I also ran **10/10 supplementary real-LadybugDB checks**: true/false values round-trip through production CSV/COPY for all five definition tables, and incremental delete plus CSV replacement preserves values and untouched rows. This is real persistence validation, not a whole analyze/migration run. No independent feature-only column-wiring defect was reproduced.

**Test-only blocker:** please check in equivalent database round-trip regression coverage alongside [node-column-parity.test.ts:43-81](https://github.com/abhigyanpatwari/GitNexus/blob/da488a37a0fb2a649de1a9f511191728b27f538a/gitnexus/test/unit/node-column-parity.test.ts#L43), and document the property/rebuild behavior. The previous local-execution gap is now closed; the remaining request concerns durable regression coverage, not a demonstrated storage failure.

Please rebase onto the corrected merged #2866 predicate and obtain required current-head CI. This head still classifies `tests/helpers.py` as non-test and `src/Contest.swift` as test, confirmed by executing the predicate; these are inherited dependency drift, not new independent findings. Focused PR-head tests do not validate a synthetic merge with latest main.

PR #2871, opened 2026-08-07
fix(scope-resolution): link Record graph nodes

### LINKED ISSUE #2801: Record is absent from LINKABLE_LABELS, making its buildGraphNodeLookup/resolveDefGraphId branches dead code
## Problem

`Record` is absent from `LINKABLE_LABELS` (`gitnexus/src/core/ingestion/scope-resolution/graph-bridge/node-lookup.ts`), but both the lookup builder and the resolver carry `Record` branches:

- `buildGraphNodeLookup` has a `Record` branch for the template-arguments key
- `resolveDefGraphId` has the matching lookup

Because `buildGraphNodeLookup` opens with `if (!isLinkableLabel(node.label)) continue;`, no `Record` node is ever registered — so **both branches are unreachable**, and a Java/C# `record` can never be a resolved call target. A `Record` def instead degrades through the label-agnostic `simpleKey(filePath, simpleName)` fallback, which is first-write-wins and can alias it onto whichever same-named node was registered first.

## Why it wasn't fixed inline

Noted during #2792 and again in #2793's review, deliberately left alone in both: adding `Record` to `LINKABLE_LABELS` **changes resolution behaviour** (new edges appear, and the `simpleKey` aliasing for same-named symbols changes), which is not a schema change and does not belong in a schema PR.

## What it costs

Adding it later is not free either — `Record` becomes a scope-bridge source *and* target, which grows the generated relation cross product by ~29 pairs and therefore needs its own `INCREMENTAL_SCHEMA_VERSION` bump. Worth doing as one deliberate change with resolution tests, not as a rider.

The relation DDL is fine either way: after #2793 `Record` is already an attachment target, and `Record|Record` is declared and corpus-guarded via the COBOL fixtures.

azizur100389:
## Summary
- register `Record` in the graph-node lookup and caller-anchor sets so Java/C# records resolve as canonical edge targets and initializer-call sources
- move the now-generated `Record` relation pairs out of the structural DDL; the derived schema fingerprint rotates automatically and forces a full re-analyze
- add lookup, Java, C#, and schema regressions, then move the production schema benchmark and budget from 450 to 461 pairs

Closes #2801

## Validation
- [x] `npx vitest run test/unit/scope-resolution/node-lookup-determinism.test.ts test/unit/schema-pair-coverage.test.ts test/integration/resolvers/java.test.ts test/integration/resolvers/csharp.test.ts` — 491 passed
- [x] `npx tsc --noEmit`
- [x] `node --import tsx bench/schema-pairs/measure.mjs --check` — 461 declared / 461 database pairs; production budget passed
- [x] `git diff --check upstream/main`
- [x] GitNexus index refresh plus `detect_changes` — expected 10-file / 11-symbol scope, no affected execution flows

A full Windows unit run completed with 11,403 passing tests and 63 unrelated host-sensitive failures across 17 files (process spawn `ENOENT`, timeouts/locks, and home-directory git-state assumptions). The changed suites all pass independently; CI remains the authoritative full-suite gate.

## Review digest
- **Claude swarm:** no code defect; requested explicit validation, a 461-pair benchmark rerun, schema invalidation documentation, and commit hygiene. Addressed.
- **Codex solo:** found one blocking gap missed by the first engine: `Record` was linkable as a definition/target but absent from `CALLER_ANCHOR_LABELS`, so initializer calls still fell back to `File`. Reproduced before the fix and covered by Java static-field and C# property-initializer regressions.
- **Final triage:** runtime risk, Test/CI, and security checks found no remaining blocker. The changed graph-bridge/schema files have no persisted taint findings, hidden bidi controls, dependency/workflow changes, or secret surface.

## Schema / landing notes
- The executable DDL changes from 450 to 461 unique relation pairs; `SCHEMA_FINGERPRINT` changes with it, so existing indexes are invalidated without a manual version bump.
- Open [PR #2856](https://github.com/abhigyanpatwari/GitNexus/pull/2856) also changes `LINKABLE_LABELS` and adjacent structural-pair rows. If it lands first, this branch must rebase and rerun the schema-pair benchmark/count before merge.


gitnexus/src/core/ingestion/scope-resolution/graph-bridge/node-lookup.ts
```diff
@@ -268,6 +268,11 @@ export const LINKABLE_LABELS: ReadonlySet<NodeLabel> = new Set<NodeLabel>([
   'Interface',
   'Struct',
   'Enum',
+  // Record participates in the same def→graph bridge as other class-like
+  // declarations. Without this entry, its qualified/template lookup branches
+  // are unreachable and label-agnostic fallback can alias it to a same-named
+  // Constructor or Method (#2801).
+  'Record',
```
magyargergo, line 275: **P1 — follow-up, not introduced here.** Making `Record` linkable is what first lands a record's `IMPLEMENTS`/`EXTENDS` edges on the canonical `Record` node (your new `csharp.test.ts:2331-2332` asserts exactly that). The interface-dispatch consumer has not followed through.

**Real location — an unchanged file, so this cannot be inlined there:** `gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts:385`
```ts
if (def.type !== 'Class' && def.type !== 'Struct' && def.type !== 'Interface') continue;
```
That builds `graphIdToClassDef`, the interface→implementors map. The same file **imports `isClassLike` at line 69** and uses it at `:159` and `:1479`; `scope/walkers.ts:172` returns `true` for `'Record'`.

**Trigger:**
```java
record User(String value) implements Named { public String name() { return value; } }
String read(Named x) { return x.name(); }
```
The call reaches `Named.name`; no dispatch edge reaches `User.name`, so callers and `impact` under-report.

**Control:** that predicate line is not in this diff — the gap is pre-existing, so this does not block merging #2871. Codex rated it "blocks merge"; the control demotes that.

**Fix (separate PR):** swap the hard-coded predicate for `isClassLike(def.type)` and add Java/C# record-interface dispatch regressions.

[code-read]

gitnexus/src/core/lbug/schema.ts
```diff
@@ -372,17 +373,19 @@ const DEFINITION_ANCHOR_LABELS: readonly NodeTableName[] = NODE_TABLES.filter(
...
- * hand-list this replaced. Four runs on the same box:
+ * Sized deliberately: this rule plus the `Record` bridge brings the DDL to 461
+ * pairs. `bench/schema-pairs` measures real `@ladybugdb/core` with identical
+ * data — untyped-endpoint anchored queries
+ * (`MATCH (a {id: $id})-[r:CodeRelation]->(b)`, the shape `impact` / `context` /
+ * `detect_changes` issue), relative to the 332-pair hand-list it replaced.
+ * The historical reference-box range plus #2801's current production range:
  *
- *   450 → 0.93–1.05×   641 → 1.22–1.43×   786 → 1.52–1.75×   1024 → 2.03–2.34×
+ *   450 → 0.93–1.05×   461 → 1.20–1.42×   641 → 1.22–1.43×   1024 → 2.03–2.34×
```
magyargergo: **P2 — this table mixes two machines, and the conclusion below it does not follow from the numbers.**

`461 → 1.20–1.42×` is a **new Windows** measurement (`bench/schema-pairs/README.md:37-38`, `baselines.json:2`); `450 / 641 / 1024` are the **unchanged historical reference-box** numbers. 461's range overlaps 641's almost exactly, so `:385-388` ("the larger pair sets show material growth") is not supported here. The sentence that used to carry the argument — "The knee sits just above 450" — was same-box grounded and was removed without a same-box replacement. `README.md` keeps the two datasets separate; this comment merges them.

**Measured independently, 6 runs on a quiet 24-core Linux box (one Claude lane; the Codex `ce` leg reached the mixed-machine conclusion by code-read, not by re-benchmarking):** `ratio_461 ∈ [0.937, 1.193]`, mean **1.057×** — inside the historical 450 noise band, and matching smooth interpolation from this file's own 450/641 numbers (≈1.01×). The same box's 641/786/1024 means (1.361 / 1.647 / 2.021) track the historical reference-box ranges closely, so it is a fair stand-in: the Windows figure did not reproduce. 3 of those 6 runs tripped the harness's own `typed_ratio > ratio` noise-dominated signal at 461.

The *conclusion* (defer the third cross product) still holds — a paired same-box comparison put 641 above 461 in 6/6 runs by 0.209–0.388 units, zero overlap with zero. It is the stated evidence that does not.

**Fix:** label the row `461 → 1.20–1.42× [Windows]`, and either re-measure 461 on the reference box or say plainly that the operational gate is the 1.5× budget, not the cross-size ordering.

Related, on `baselines.json:2`: headroom dropped from ~24% (1.3 vs 1.05) to ~5.6% (1.5 vs 1.42), and the raise was load-bearing — the worst reported run (1.42×) would have failed the old 1.3 budget. The first discard rule ("typed floor moved more than the untyped ratio") is pre-registered and legitimate; the second ("anomalously fast 332-pair reference") has no antecedent rule, no numeric threshold, and no raw numbers published anywhere, so it is not auditable as written.

[reproduced]

gitnexus/src/core/lbug/schema.ts
```diff
@@ -413,8 +416,8 @@ const ATTACHMENT_TARGET_LABELS: readonly NodeTableName[] = [
  * NOTHING A RULE ALREADY COVERS BELONGS HERE. `generatedRelationPairs` skips
  * any pair present in this block, so a redundant line does not merely duplicate
  * — it SUPPRESSES generation, and later narrowing a rule would silently keep
- * that pair alive with no test failing. 161 such lines were deleted from this
- * block (the DDL's pair set is unchanged: they moved into the generated half);
+ * that pair alive with no test failing. 164 such lines now live in the generated
```
magyargergo, line 421: **P2 — this docblock update left three stale numbers behind** (all verified on a clean checkout; none of the three lines is itself in the diff, hence the comment here).

| Location | Says | Actual |
|---|---|---|
| `schema.ts:402` | "The **72** pairs NEITHER rule above generates" | **69** — measured on `STRUCTURAL_PAIR_DDL` at head; `bench/schema-pairs/README.md:79` already says 69 |
| `schema.ts:427` | "where production's **450** is inside noise" | **461** — this same diff updated 450→461 elsewhere in this file |
| `test/unit/schema-pair-coverage.test.ts:42` | "the **~72** containment/inheritance/import pairs" | **69** |

The third matters most: that file *is* edited by this PR (comment-only, `:125-128`), so the number one screen above it was in reach. `schema.ts:427` is also semantically wrong now — 461 at 1.20–1.42× is not "inside noise" by the standard that made 450 so; it reads better as "461 is below its 1.5× budget".

These are load-bearing: `:427` is the input a future editor uses to decide whether to fold the remainder into a third cross product, and the file now asserts two different production sizes ~40 lines apart.

**Fix:** `72` → `69` at `schema.ts:402`, `450` → `461` at `schema.ts:427`, `~72` → `~69` at `schema-pair-coverage.test.ts:42`.

[code-read]

gitnexus/bench/schema-pairs/measure.mjs
```diff
@@ -85,9 +86,9 @@ const lbug = (await import('@ladybugdb/core')).default;
 
 // ---- sizes + the pair enumeration every size is a prefix of ----
 
-const SIZES = [332, 450, 641, 786, 1024];
+const SIZES = [332, 450, 461, 641, 786, 1024];
 const REFERENCE_SIZE = 332; // ratios are relative to this
-const PRODUCTION_SIZE = 450; // the size schema.ts ships
+const PRODUCTION_SIZE = 461; // the size schema.ts ships
```
magyargergo: **P2 — `PRODUCTION_SIZE` is a hand-copied twin of a number the file already derives, and nothing compares the two.**

`verifyRealSchema()` (`:274-292`) computes the truth — `parseRelationSchemaPairs(RELATION_SCHEMA)` — but only checks it against `SHOW_CONNECTION`, i.e. **both sides derived from the same `RELATION_SCHEMA`**. It never compares `declared_pairs` to `PRODUCTION_SIZE`, and there is no `461` literal in `schema-pair-coverage.test.ts`. So nothing anywhere binds this literal to the real DDL size, and at `:342-343` a *missing* budget key short-circuits the gate:
```js
const budget = baselines[`ratio_${PRODUCTION_SIZE}_budget`];
if (budget !== undefined && summary[`ratio_${PRODUCTION_SIZE}`] >= budget) { … }
```
`--check` then prints `PASS`. Same `empty-set ≠ no-set` shape that armed the ACCESSES-deleting filter in #2856.

**Two controls, so this is not rated higher:** the `budget !== undefined` fail-open is pre-existing code, and `grep -rn "measure.mjs" .github/workflows/` shows **no schema-pairs entry** (8 other benches are wired; this one is not), so nothing is silently passing today. What *this* PR changes is the coupling — renaming `ratio_450_budget → ratio_461_budget` in lockstep with `PRODUCTION_SIZE` establishes that the literal must move on every `LINKABLE_LABELS` widening, and PR #2856 is about to move the real count. (Every combined-count estimate in circulation is unmeasured — re-run this against `origin/main` immediately before merge rather than trusting one.)

**Fix:** derive `PRODUCTION_SIZE` from `verified.declared_pairs` (or push a mismatch into `failures`), and make a missing budget key a failure:
```js
const key = `ratio_${PRODUCTION_SIZE}_budget`;
if (baselines[key] === undefined) failures.push(`no ${key} in baselines.json — the gate is disarmed`);
```

[code-read] (fail-open half [reproduced])

gitnexus/test/unit/schema-pair-coverage.test.ts
```diff
@@ -122,9 +122,10 @@ describe('RELATION_SCHEMA pair coverage', () => {
     // then silently keep that pair alive, and every other assertion in this file
     // subtracts `structural` from both sides, so none of them would notice.
     // (The integration corpus only checks emitted ⊆ declared — blind to an
-    // EXCESS declaration by construction.) 161 such lines were deleted when this
-    // guard went in; the DDL's pair set did not change, they moved into the
-    // generated half. Failing here means deleting the line, not widening this.
+    // EXCESS declaration by construction.) 164 such lines now live in the
+    // generated half: 161 moved when this guard went in, plus the three Record
+    // member pairs moved when Record became linkable (#2801). Failing here means
```
magyargergo, line 127: **P2 — the schema half of this PR ships with zero new assertions, and this file cannot see the change it describes.**

The `+4/-3` here is comment text only. And every assertion in this file recomputes its expectation from the *same* constant the production change edits: `schema-pair-coverage.test.ts:9` and `schema.ts:17` import the same `LINKABLE_LABELS`, so both sides of every equality move together. Mutation-verified — removing `'Record'` from **both** sets leaves **all 9 assertions green** while the pair count drops 461 → 447 and `Record|Method` / `Record|Constructor` / `Record|Property` go undeclared. The file pins the cross-product *shape*, not the label-set *membership*; the docblock at `:30-33` overstates this.

**Control, so the scary version is not being claimed:** that revert does *not* ship silently. Running the same mutation with `test/unit/schema.test.ts` (not in this diff, so easy to miss) turns CI red — `schema.test.ts:255-268` pins `Record|Method` and `Record|Constructor` by literal.

**What is genuinely unpinned is `Record|Property` alone.** `schema.test.ts:257`'s own comment says "(Property is now HAS_PROPERTY)" and no HAS_PROPERTY pin test exists, so after this PR moves that row out of `STRUCTURAL_PAIR_DDL` its only guarantee is derivation from `LINKABLE_LABELS`, while its two siblings keep an independent literal assertion.

**Fix (one line):** add `'Record|Property'` to the literal `memberPairs` list in `test/unit/schema.test.ts:234-248`, or to the `reported` literal list at `schema-pair-coverage.test.ts:176-183` — the one assertion in this file whose expectation is *not* derived from `LINKABLE_LABELS`.

[reproduced]

magyargergo (commented):
# Review — PR #2871 `fix(scope-resolution): link Record graph nodes` (closes #2801)

**Engines.** Claude ran the GitNexus PR swarm (risk, test/CI, security, docs/DoD, branch-hygiene) plus five `ce-code-review` personas (correctness, adversarial, testing, performance, standards). Codex (gpt-5.6-sol / xhigh) ran the same swarm spec, a report-only `ce-code-review` pass, and a native adversarial red-team — **all three Codex legs are live and produced real findings.** Two engines, not five independent confirmations; corroboration below is labelled accordingly.

**Verdict: this PR is correct. Nothing it introduces blocks merge.** The remaining items are two one-line comment fixes, a test pin, and three follow-up issues for pre-existing gaps it makes newly visible.

---

## 1. What is solid — and the executed evidence behind it

This was an unusually well-validated change. The most valuable output of this review is what it *ruled out*, with runs rather than reasoning.

**Full-corpus A/B over 2119 files** (`test/fixtures/lang-resolution`, real pipeline via `runPipelineFromRepo`, base simulated by deleting `'Record'` from the two exported `Set`s in-process — faithful, since the entire `src/` delta is those two additions plus comments): `[reproduced]`

```
lookup map:   base keys=17429   head keys=17446
              keys REMOVED or CHANGED (base -> head):  (none)
edge diff:    13 changed lines, every one an improvement, zero regressions
```

The 13 lines are: Java compact-constructor calls moving `File:Compact.java` → `Record:…:Compact`, C# record inheritance moving `Constructor:Shapes.cs:{P.P#1,R.R#1}` → `Record:Shapes.cs:{P,R}` (×4), and two newly-resolved `METHOD_IMPLEMENTS` edges. **No edge lost a correct endpoint, and not a single existing lookup key changed value.** The premise in the PR's own code comment — that a record's inheritance edges were being sourced at its primary-constructor twin — is true and reproduced twice, at both unit and pipeline level.

**The Claude adversarial persona swept 22 real fixtures plus 13 hand-built Java/C# shapes** (same-named record/method collisions, nested records, compact constructors, generic records, `record struct`, local records in method bodies / `if` blocks / lambda bodies, anonymous classes inside records, two same-named records across files, C# records in two namespaces in both declaration orders); every base→head delta was an improvement or a no-op. **The Codex adversarial leg reached the opposite verdict** — "do not ship" — probing specific failure shapes and identifying two apparent regressions, packaged-Java aliasing and field-lambda anchor capture. Both were subsequently refuted by executed controls (§4.7, §4.8). That disagreement is recorded rather than smoothed over: it is the one place the two engines diverged, and the controls, not either engine, settled it.

**The rest of the pre-registered failure modes, all refuted with evidence:**

- **Undeclared-pair `analyze` abort** — the classic failure for this file. Measured: `RELATION_SCHEMA` 450 → **461**, exactly 11 pairs added (`Record|{Function,Module,Class,Interface,Struct,Enum,Trait,Variable,Const,Macro,Delegate}`), **zero removed**. Zero undeclared pairs emitted across 35 fixtures at head *and* at base. Every emitter that can now receive a `Record` id was traced individually: Spring AOP (`eligibleOwner` rejects a Record owner on all three paths), Spring DI metadata, Spring conditionals, `tryEmitEdgeWithExplicitTargetId`, MRO, inheritance, PDG (`BasicBlock↔BasicBlock` only), and COBOL. **Cross-engine (both).**
- **Caller-anchor tier ordering** — `resolveCallerGraphId` (`ids.ts:471-505`) is genuinely two-tier and sequential: `pickCallerCallableDef` (Function/Method/Constructor, via `isOverloadableCallable`) runs first on each scope; the flat `isCallerAnchorLabel` membership test at `:500` is only consulted when tier 1 returns `undefined`. A Record therefore cannot outrank a nested callable. Five executed shadowing probes all keep the correct anchor. **Cross-engine (both).**
- **Deleting the three `FROM \`Record\` TO Method/Constructor/Property` structural rows** — correct *and* complete. Re-measured on a clean tree: 0 of the 69 remaining `STRUCTURAL_PAIR_DDL` rows fall inside `SCOPE_BRIDGE_SOURCE × SCOPE_BRIDGE_TARGET`, so the "a redundant structural row SUPPRESSES generation" invariant holds. `schema-pair-coverage.test.ts` fails correctly if one comes back (mutation-verified).
- **Determinism / incremental** — `nodesInSourceOrder` (`parsing-processor.ts:85-92`) canonicalises by `(filePath, startLine, node.id)` *before* every `addNode`, on both the fresh-parse and cache-replay path, so first-write-wins is content-derived, not scheduler-derived. The lookup map is byte-stable across runs.
- **Stale-index reuse** — `SCHEMA_FINGERPRINT` rotates `0061b2888f22` → `19ccb2eb45a8`. It is a sha256 digest of the DDL (#2798), so no hand-numbered constant can clash. The gate at `run-analyze.ts:1329` sits *before* the `alreadyUpToDate` fast path, treats an absent fingerprint as a mismatch, and sets `force: true`. `runFullAnalysis` is the only graph-writing orchestrator — no adopt/warm-cache side door. `SCHEMA_BUMP` correctly untouched (both constants are consumed in the main thread after parse, so a warm `ParseWorkerResult` cache cannot replay stale bridge output).
- **DDL injection via label interpolation** (`schema.ts:544`, `` FROM `${from}` TO `${to}` ``) — safe by closed-set design. `NodeLabel` is a closed TS literal union, every member pure alphabetic ASCII, and no user repository content ever reaches `generatedPairDdl()`. No supply-chain, workflow, secret, or filesystem-surface change; gitleaks and `git diff --check` clean; no bidi/zero-width/homoglyph characters.
- **Project standards** — no violation. The `ingestion/` "must not name languages" rule is not breached: the diff adds two *values* to `ReadonlySet<NodeLabel>` constants with no language-keyed branch, matching the pre-existing `Trait` entry whose comment already names PHP/Rust. No `any`, PR title conventional-commit compliant, CHANGELOG correctly untouched, no new platform-specific code.

**Test strength is real, and was mutation-proven — including the pin this repo has historically got wrong.** A full mutation matrix ran the integration suites end-to-end (`java.test.ts` 250/250, `csharp.test.ts` 228/228):

| Mutation | Result |
|---|---|
| `'Record'` removed from `LINKABLE_LABELS` only | 4 failures across 3 independent files (unit node-lookup, Java, C#) |
| `'Record'` removed from `CALLER_ANCHOR_LABELS` only | exactly the two new initializer regressions fail, **and nothing else in 478 integration cases** |
| the three structural rows re-added | `declares no hand-declared pair that a rule already generates` fails and names the pair |

The PR body's claim that "the new Java static-field and C# property-initializer regressions cover exactly this `CALLER_ANCHOR_LABELS` gap" is **literally true**. And `node-lookup-determinism.test.ts` is **not** the tautological deny-set shape this repo shipped in #2855 — removing `'Record'` makes `qualifiedKey` return `undefined` and `resolveDefGraphId` return `Method:src/service.ts:Factory.Person#0`; both assertions flip. Verified independently by both engines.

---

## 2. Headline finding

### P1 (follow-up, **not introduced by this PR**) — record implementors are invisible to interface-dispatch fan-out

`gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts:385` — `[code-read]`

```ts
if (def.type !== 'Class' && def.type !== 'Struct' && def.type !== 'Interface') continue;
```

This builds `graphIdToClassDef`, the interface→implementors map that drives dispatch fan-out. The same file **imports `isClassLike` at line 69** and uses it at `:159` and `:1479`; `scope/walkers.ts:172` has `isClassLike` return `true` for `'Record'`.

**Trigger:**
```java
record User(String value) implements Named { public String name() { return value; } }
String read(Named x) { return x.name(); }
```
The call reaches `Named.name`, but no interface-dispatch edge reaches `User.name` — callers and `impact` under-report.

**Control (this is the important part): the predicate line is not in this PR's diff — it is pre-existing.** Codex rated this "blocks merge: yes"; that rating does not survive the control. Frame it as an *incomplete follow-through*: this PR is what first makes a record's `IMPLEMENTS`/`EXTENDS` edges land on the canonical `Record` node (asserted by the new C# test at `csharp.test.ts:2331-2332`), so the dispatch consumer that still hard-codes three labels is now the remaining gap rather than a latent one. Fix is one line — swap the hard-coded predicate for `isClassLike(def.type)` — plus Java/C# record-interface dispatch regressions. Best done as its own PR with its own tests.

*Corroboration: Codex swarm only (Engine B), confidence 99%, re-verified by the coordinator on a clean checkout. No Claude lane found it.*

---

## 3. Remaining findings

### P2 — the benchmark table in `schema.ts:383` mixes two machines and draws a conclusion the data cannot support

**Cross-engine (both).** `[reproduced]`

```
450 → 0.93–1.05×   461 → 1.20–1.42×   641 → 1.22–1.43×   1024 → 2.03–2.34×
```
followed at `:385-388` by *"461 remains below its 1.5× production budget, while the larger pair sets show material growth."*

`461` is a **new Windows** measurement (`README.md:37-38`, `baselines.json:2`); `450 / 641 / 1024` are the **unchanged historical reference-box** numbers. 461's range overlaps 641's almost exactly, so "the larger pair sets show material growth" does not follow from this table. The pre-PR sentence that *did* carry the argument — *"The knee sits just above 450"* — was same-box grounded and was deleted without a same-box replacement. `bench/schema-pairs/README.md` keeps the two datasets separate; `schema.ts` merges them.

Independently measured, 6 runs on a quiet 24-core Linux box (one Claude lane; the Codex `ce` leg reached the mixed-machine conclusion by code-read, not by re-benchmarking): `ratio_461 ∈ [0.937, 1.193]`, mean **1.057×** — inside the historical 450 noise band, and matching what smooth interpolation from the file's own 450/641 numbers predicts (≈1.01×). The same box's 641/786/1024 means (1.361 / 1.647 / 2.021) track the historical reference-box ranges closely, so it is a reasonable stand-in — **the Windows 1.20–1.42× figure did not reproduce.** Three of those six runs tripped the harness's own `typed_ratio > ratio` noise-dominated signal at size 461.

Two related notes on the budget raise itself:
- **Headroom shrank from ~24% to ~5.6%** (old: budget 1.3 vs top-observed 1.05; new: budget 1.5 vs top-observed 1.42), and the raise was load-bearing — under the old 1.3 budget the PR's own worst reported run (1.42×) would have failed `--check`.
- Of the two discard justifications in `baselines.json:2`, the first ("typed floor moved more than the untyped ratio") is a **pre-existing, pre-registered rule** and its use is legitimate. The second ("an anomalously fast 332-pair reference") has **no antecedent rule anywhere in the harness, no numeric threshold, and no raw numbers published** in the diff, the PR body, or any PR comment — so a reviewer cannot recompute "anomalous". Not provably wrong; not auditable as stated.

The *conclusion* the file draws (defer the third cross product) still looks right for a different reason: a paired same-box comparison put 641 above 461 in 6/6 runs by 0.209–0.388 ratio-units with zero overlap with zero. **Suggested fix:** label the 461 row `[Windows]`, and either re-measure 461 on the reference box or say plainly that the operational gate is the budget rather than the cross-size ordering.

### P2 — `PRODUCTION_SIZE` is a hand-copied twin of a derived number, and this PR is what arms the coupling

**Cross-engine (both).** `[code-read]`, with the fail-open half `[reproduced]`

`bench/schema-pairs/measure.mjs:91` `const PRODUCTION_SIZE = 461;` and `:342-343`:
```js
const budget = baselines[`ratio_${PRODUCTION_SIZE}_budget`];
if (budget !== undefined && summary[`ratio_${PRODUCTION_SIZE}`] >= budget) { failures.push(…) }
```

`verifyRealSchema()` (`:274-292`) was read in full: it compares `parseRelationSchemaPairs(RELATION_SCHEMA)` against `SHOW_CONNECTION` — **both derived from the same `RELATION_SCHEMA`** — and never compares `declared_pairs` to `PRODUCTION_SIZE`. There is no `461` literal in `schema-pair-coverage.test.ts` either. So **nothing anywhere binds `PRODUCTION_SIZE` to the real DDL size**, and a missing budget key makes `--check` print `PASS`. The `461` literal now lives in four hand-maintained places (`schema.ts`, `measure.mjs`, `baselines.json` key name, `README.md`) with no mechanical link.

Two controls that must travel with this finding:
- **The fail-open `budget !== undefined` is pre-existing code.** What *this* PR changes is the coupling: it renames `ratio_450_budget → ratio_461_budget` in lockstep with `PRODUCTION_SIZE 450 → 461`, establishing that the literal must now move on every `LINKABLE_LABELS` widening.
- **`measure.mjs --check` runs in NO CI workflow.** `grep -rn "measure.mjs" .github/workflows/` returns python-scope, scope-capture, callable-value-flow, cpp-qualified-ns, receiver-resolution, scope-emission, cfg, emit-persistence, impact-pdg — **no schema-pairs**. This PR changes no workflow file, so **that gap is pre-existing too**, and the blast radius today is zero because nothing runs the gate. Codex rated this P1; the CI control demotes it to P2.

Open PR #2856 also widens `LINKABLE_LABELS`, so whichever lands second moves the real count off 461. The combined pair-count estimates floating around this review (~482, ~483, ~491) are **all unmeasured** — do not act on any of them; re-run `measure.mjs` against `origin/main` immediately before merge, not at review time.

**Suggested fix:** derive `PRODUCTION_SIZE` from `verified.declared_pairs` (or push a mismatch into `failures`), and make a missing budget key a failure rather than a pass.

### P2 — three stale numbers in load-bearing comments, one of them in a file this PR edited

**Cross-engine (both).** `[code-read]` — all coordinator-verified on a clean tree.

| Location | Text | Actual |
|---|---|---|
| `src/core/lbug/schema.ts:402` | "The **72** pairs NEITHER rule above generates" | 69 (`STRUCTURAL_PAIR_DDL` measured at head; `README.md:79` already says 69) |
| `src/core/lbug/schema.ts:427` | "where production's **450** is inside noise" | 461 — the same diff updated 450→461 elsewhere in this file |
| `test/unit/schema-pair-coverage.test.ts:42` | "the **~72** containment/inheritance/import pairs" | 69 |

The third one matters more than it looks: that file *is* edited by this PR (comment-only, `:125-128`), so the contributor was inside that comment block and missed the number one screen above. Only the Codex `ce` leg found this occurrence; no Claude lane did. `schema.ts:427` is also semantically wrong now — 461 at 1.20–1.42× is not "inside noise" by the standard that made 450 so.

Neither `schema.ts:402` nor `:427` nor `test:42` is itself a diff line, so the inline comment for this is anchored at `schema.ts:419`, inside the same docblock.

### P2 — `schema-pair-coverage.test.ts` is tautological with respect to this change, and `Record|Property` loses its last independent pin

**Cross-engine on the tautology** (Codex swarm's test-strength table reaches the same conclusion: *"The schema test derives expectations from the same sets and therefore does not independently pin Record membership"*); **Claude-only on the `Record|Property` narrowing.** `[reproduced]`

Running the natural revert — `'Record'` removed from **both** sets — leaves **all 9 `schema-pair-coverage.test.ts` assertions green** while `Record|Method` / `Record|Constructor` / `Record|Property` go undeclared (461 → 447). The cause: `schema-pair-coverage.test.ts:9` and `schema.ts:17` import the *same* `LINKABLE_LABELS`, so both sides of every equality move together. Any membership change to that set — widening or narrowing — is structurally invisible to those assertions; they pin the cross-product *shape*, not the *membership*. Consistent with that, the file's `+4/-3` in this PR is **comment text only, zero new assertions**, so the schema half of the change ships unpinned.

**One lane's stronger framing does not survive its control and is not being posted.** The claim was that a future revert would therefore ship silently and abort `analyze` on a user repo with CI green. Running that exact mutation *plus* `test/unit/schema.test.ts` (which that lane did not run, since it is not in the diff) gives a **red CI**:

```
FAIL test/unit/schema.test.ts > has all FROM/TO pairs needed for HAS_METHOD edges
AssertionError: expected false to be true    at test/unit/schema.test.ts:266:54
```

`schema.test.ts:255-268` pins `Record|Method` and `Record|Constructor` by literal (`backtickSources = ['Struct','Trait','Impl','Record']`). **The accurate, narrowed version:** the one pair with no independent pin anywhere is **`Record|Property`** — `schema.test.ts:257`'s own comment says *"(Property is now HAS_PROPERTY)"* and no HAS_PROPERTY pin test exists. Its two siblings keep a literal assertion; it does not.

**Suggested fix (one line):** add `'Record|Property'` to the literal `memberPairs` list in `test/unit/schema.test.ts:234-248`, or to the `reported` literal list in `schema-pair-coverage.test.ts:176-183` — the one assertion in that file whose expectation is *not* derived from `LINKABLE_LABELS`.

### P3 — test-coverage gaps worth a follow-up (not merge-blocking)

- **`java.test.ts:1249` bundles two independent behaviours in one `it()`.** `:1266` (caller-anchor) fails under *both* mutations, so `:1271` (call-target) never executed under either — it is a genuine assertion, but the suite can never report on it independently, and the test title names both behaviours without saying which broke. `[reproduced]`
- **C# has no Record-as-CALL-TARGET assertion.** Java pins `constructorCall?.targetLabel === 'Record'` (`java.test.ts:1271`); C# has the caller-anchor case and the EXTENDS case but no `new SomeRecord(...)` target assertion — and `Record|Constructor` is one of the three rows this PR deleted, C# records mint an implicit same-named `Constructor` node, and that node is exactly what the base misresolves onto. *Two Claude lanes.* `[reproduced]`
- **The C# fixture is named `Record`**, so `edge.source === 'Record'` (matching `properties.name`) and `expect(...sourceLabel).toBe('Record')` (matching `label`) are the same string with two different meanings. It discriminates correctly today, but a refactor of `.sourceLabel` → `.source` keeps it green. Java gets this right (`Point` / `'Record'`).
- **The new tests do not exercise the qualified-key path under realistic packaged Java.** `node-lookup-determinism.test.ts:104` hand-writes `qualifiedName: 'Person'` on the Record node, so node side and def side agree by construction; real packaged Java produces `p.Person` on the node property versus a bare `Person` scope def. The Java integration fixture *does* use `package probe;` and passes — but only because `Point.java` has no same-named competitor, so it resolves through `simpleKey`, not the qualified key. See §4.7.
- Also unpinned: a **generic** record (the `templateArguments` branch at `node-lookup.ts:219-234` / `ids.ts:414-425` was dead code for `Record` before this PR and is now live — verified working, nothing pins it); a **positional** C# record (the new C# test uses a non-positional `record`, which mints no `Constructor` twin, i.e. never exercises the contested case); the Tier-1-beats-Tier-2 interaction.
- `CALLER_ANCHOR_LABELS ⊆ LINKABLE_LABELS` is enforced in one direction only, so the test **cannot** catch the exact omission this PR fixed during development (adding to `LINKABLE_LABELS` alone keeps the subset valid). An enumerated-exclusion assertion would turn "remember to edit both files" into a named CI failure.

### P3 — pre-existing gaps this PR extends (all controlled; listed for follow-up, not for this PR)

- **`Record → Record` self-`CALLS`** from a self-constructing static initializer (`record Point(int x) { static final Point ORIGIN = new Point(0,0); }`). **Control:** a Java class with an *implicit* default constructor already mints `Class:X -> Class:X` at BASE — the root cause is "no Constructor node exists for an implicit ctor, so the type node is used as both endpoints". Pre-existing modelling gap, now extended to `Record`. `[reproduced]`
- **Community detection drops re-anchored calls.** `community-processor.ts:412-416` admits `Function|Class|Method|Interface` only, so moving a CALLS source from `Class:Outer` to `Record:Inner` loses the callee's only projected neighbour. **Control:** `Struct` and `Enum` are already caller anchors and already absent from that predicate, so Go structs and Rust enums have the identical hole today. Separately, the dangling `MEMBER_OF` to an uncreated singleton `Community` node was reproduced on a **records-free TypeScript fixture at BASE** — endemic and pre-existing, refuted as a PR defect. `[reproduced]`
- **COBOL `Record` nodes are now nominally linkable and anchorable.** `cobol-processor.ts` mints `Record` for FD/`SELECT` file declarations, DB2 tables, CICS queues and BMS maps — not class-like at all, so the new comment ("participates in the same def→graph bridge as other class-like declarations") is not true for every producer of the label. **Control:** `languages/cobol.ts:25` declares `extensions: []` and COBOL runs in phase 2.6, contributing no scope-resolution defs, so neither path is reachable; the corpus A/B shows zero COBOL edge changes at both revisions. Worth one line of comment stating the decision. `[reproduced]`
- **`simpleKey` slot flips `Constructor → Record` for an *attributed* C# positional record** (the attribute list puts `record_declaration` on an earlier line than its `parameter_list` twin). Exposure only — `new Point(...)` resolves through the arity/param-type keys long before the terminal fallback, and no wrong edge was reachable. Without an attribute the two tie on `startLine` and `Constructor:` sorts first, which is why no fixture exhibits it. `[reproduced]`
- **`mro-processor.ts:19`** advertises `MATCH (c:Class)-[…METHOD_OVERRIDES…]->(m:Method)` as the canonical query; record overrides now exist and that query misses them.
- **`CALLER_ANCHOR_LABELS` has no invalidation channel of its own** — it is not a DDL input, not a parse-time capture, and not in `ANALYSIS_FEATURES`. This PR is safe only because `LINKABLE_LABELS` moved in the same commit and *that* rotates the fingerprint. A future PR touching `ids.ts` alone would rotate nothing and leave a permanently mixed incremental graph. The mechanism to prevent it (`ANALYSIS_FEATURES`) exists and is simply not wired to the graph bridge.

---

## 4. Suspicions we chased and refuted

Recorded so they are not re-raised without new evidence. Every one of these was killed by an executed control, not by argument.

1. **"The fix merely reverses which side wins the lookup."** Refuted by the 2119-file corpus A/B: **0 keys changed value**, 17 added; 13 changed edges, all improvements, 0 regressions.
2. **"A flat `isCallerAnchorLabel` membership test lets a Record shadow a nested method or lambda anchor."** Refuted — tiers are per-scope and sequential (`ids.ts:471` before `:500`); five shadowing constructions (local record in a method body, in an `if` block, in a `for` block, in a lambda body, sibling-order steal) all keep the correct anchor.
3. **"Adding a label class opens an undeclared pair and aborts `analyze` on a user repo."** Refuted — 11 added, 0 removed, 0 undeclared across 35 fixtures at both revisions, with every emitter traced individually.
4. **"A label-set widening changes the contested-slot winner non-deterministically, or differently under incremental."** Refuted — `nodesInSourceOrder` canonicalises by `(filePath, startLine, id)` before every insertion, on both the fresh-parse and cache-replay paths.
5. **"A stale index could keep serving the 450-pair schema."** Refuted — digest-based fingerprint, gate ahead of the `alreadyUpToDate` fast path, absent treated as mismatch, `force: true`. (Residual, pre-existing: if `existingMeta` is `null`, the check is skipped — an ancient pre-`meta.json` index. Not introduced here.)
6. **"Label interpolation into the DDL is an injection surface."** Refuted — closed `NodeLabel` union, alphabetic ASCII only, no user content on any path.
7. **Codex adversarial "high": packaged records still alias to an earlier same-named method.** Symptom **reproduces verbatim on HEAD** (`CALLS Method:Use.make#0 -> Method:Factory.Person#0` where `Record:p/All.java:Person` was expected) — but the stated mechanism is wrong and **three controls flip the verdict**: the same packaged fixture on **BASE** emits the *identical* wrong edge; an identical packaged **`class Person`** emits the *identical* wrong edge; the unpackaged record and the declaration-order-first record both resolve correctly. The node id is not package-prefixed (Java never opts into `qualifiedNodeId`); the prefix lands on the `qualifiedName` *property*, which `node-lookup.ts:161-162` prefers when building the key, while Java's scope query emits no qualified-name capture. **So this is the generic packaged-Java class-like bug, neither introduced by this PR nor Record-specific — LOW, follow-up issue, framed around packaged Java class-likes.** The inverse probe (a newly-linkable Record stealing the simple key from a legitimately-targeted same-named method) emits the same edge at HEAD and BASE — no new regression. *Correction for the record:* Codex stated the new unit test "places the record FIRST"; it is `buildLookup([sameNamedMethod, record])` — the **method** is first, which is precisely what makes the assertion meaningful.
8. **Codex adversarial "medium": the Record anchor steals calls made inside a field lambda.** Edge confirmed, defect refuted. HEAD emits `CALLS Record:p/R.java:R -> Method:R.target#0`; **BASE emits `File:p/R.java`, and an identical `class R` emits `Class:p/R.java:R`.** Java's scope query genuinely creates no lambda scope — a pre-existing gap identical at both revisions. `File` → `Record` is a **precision improvement**, byte-identical to what `class` has always done. Informational, not a defect.
9. **"The new unit test is tautological."** Refuted by mutation in both engines. The tautological shape is real but lives in `schema-pair-coverage.test.ts` (see §3), not here.
10. **"The merge-from-main commit needs a rebase."** Moot — the branch is now a single linear commit. `CONTRIBUTING.md:68` exempts intra-PR commits from the convention anyway.
11. **A P0 claim that the bench gate would silently pass a wrong pair count after #2856.** The *mechanism* is confirmed (nothing binds `PRODUCTION_SIZE` to the real count) but the severity is not: no CI workflow runs the bench, so nothing is silently passing today, and the ~483-pair figure behind the claim is an unmeasured estimate. Rated P2 in §3.

---

## 5. CI

**38 passing, 1 skipped, 1 failing.** The single failure is Vercel — *"Authorization required to deploy"* — fork deployment authorisation, i.e. infrastructure/auth, unrelated to this change. GitHub reports `MERGEABLE` / `BLOCKED` only on pending required checks.

## 6. Head moved mid-review

The review ran against `7efbc82017fd1351f4c1eee25a5c0eab305c2244` (a merge commit). The PR was force-pushed to `5aaa4d3cc4ccba36803155aa5d81d486f4bec9be` (linear, merge flattened) during the run. **Both commits have the byte-identical tree `4fca71534774c5c8f448ef5c108e3e7dee63c1ca`** (verified with `git rev-parse <sha>^{tree}`), so every finding above applies verbatim to the current head. The Codex swarm leg detected the force-push independently.

## 7. Coverage limits

- **The integration resolver suites need a raised worker-ready timeout** (`GITNEXUS_WORKER_READY_TIMEOUT_MS=60000–90000`; the dist parse worker's import takes ~8.4s against a 5000ms handshake) and `GITNEXUS_WORKER_POOL_SIZE=2` on a loaded box. Several lanes — including both Codex legs — could not run them at all and fell back to code reading or out-of-band pipeline probes. One lane did run them in full: `java.test.ts` 250/250, `csharp.test.ts` 228/228.
- One lane's file-level mutation testing briefly modified the shared worktree that other lanes were reading. Every postable finding above was re-verified by the coordinator against a clean checkout; the worktree is byte-clean (`git status --porcelain` empty).
- A first `node scripts/build.js` failed with three bogus `TS1005` errors inside the `STRUCTURAL_PAIR_DDL` template literal while `tsc --noEmit` on the identical tree passed, and an immediate retry succeeded. This is the known non-reproducible-build class for this repo, not something #2871 introduced — but it means "the build went green" is not a stable signal here.
- Neither engine re-ran the Windows benchmark; the 461 figure is quoted from the PR, and the only independent measurement is the Linux one reported in §3.

---

## Recommendation

**Approve after two one-line comment fixes** (`schema.ts:402` 72→69, `schema.ts:427` 450→461; plus `schema-pair-coverage.test.ts:42` `~72`→`~69`), and ideally the one-line `Record|Property` literal pin. Everything else is a follow-up issue:

1. `receiver-bound-calls.ts:385` → `isClassLike(def.type)`, with record-interface dispatch regressions (**P1**, pre-existing).
2. Bind `PRODUCTION_SIZE` to `verified.declared_pairs` and fail closed on a missing budget key; decide whether `bench/schema-pairs --check` should be in CI at all (**P2**, pre-existing fail-open).
3. Packaged-Java class-like resolution: `props.qualifiedName` is package-qualified while the scope def is bare, so every packaged Java class-like falls through to the label-agnostic, source-order-dependent `simpleKey` (**follow-up**, pre-existing, not Record-specific).
4. Re-run `bench/schema-pairs/measure.mjs` against `origin/main` immediately before merge — PR #2856 also widens `LINKABLE_LABELS`, and whichever lands second must re-derive `PRODUCTION_SIZE`, the `ratio_<N>_budget` key, and the `461` literals in `schema.ts` / `README.md` / `measure.mjs` together.

gitnexus/src/core/lbug/schema.ts
azizur100389: Addressed in e95ee9c4. schema.ts now keeps the historical reference-box and Windows measurements explicitly separate and treats 1.5x as the operational production gate, not evidence for cross-machine ordering. The benchmark README publishes all six Windows ratios, applies only the pre-registered typed_ratio > ratio noise rule, and no longer silently excludes the 2.705x run.

gitnexus/src/core/ingestion/scope-resolution/graph-bridge/node-lookup.ts
azizur100389: Agreed. This is a pre-existing interface-dispatch follow-up rather than a regression introduced by #2871, so I have kept it out of this PR as recommended. The current PR remains limited to making the canonical Record node linkable and anchorable; the isClassLike(def.type) dispatch change needs its own Java/C# regressions.

gitnexus/bench/schema-pairs/measure.mjs
azizur100389: Addressed in e95ee9c4. PRODUCTION_SIZE is now derived from parseRelationSchemaPairs(RELATION_SCHEMA).size and inserted into the measured size set. --check now fails closed for a missing budget key and rejects non-finite budget values. The real-schema run reports declared_pairs=461/db_pairs=461 and passes the derived ratio_461_budget gate.

gitnexus/test/unit/schema-pair-coverage.test.ts
azizur100389: Addressed in e95ee9c4. Record|Property is now an independent literal in schema.test.ts's memberPairs list, so removing Record from the generated bridge can no longer leave that moved pair covered only by an expectation derived from LINKABLE_LABELS.

gitnexus/src/core/lbug/schema.ts
azizur100389: Addressed in e95ee9c4: both 72/~72 references are now 69/~69, and the deferred-third-rule guidance now names the current 461-pair production surface and its 1.5x Windows budget without treating cross-machine values as an ordering.

PR #1918, opened 2026-05-30
perf(ingestion): linearize scope-capture across all languages + Python import resolution (O(n²)→O(n))

magyargergo:
## What & why

Eliminates the **O(n²) scope-capture** hot path across **every benchmarked language** and the **O(imports × files)** Python import-resolution scan. Pure performance work: every change is output-preserving (byte-identical capture/resolver fingerprints), and the win is locked by build-free harnesses, per-language golden tests, and non-gated tripwires.

A unified measurement harness, added here, is what revealed the bug was latent in four more languages beyond the Python case that started this.

---

## 1. Scope-capture: O(n²) → O(n) in six languages

Each `emit<Lang>ScopeCaptures` re-derived every tree-sitter match's AST node with `findNodeAtRange(tree.rootNode, …)` **per match** → O(matches × rootChildren). Fix (same shape as Go #1915): thread the query's captured node (`c.node`) and resolve each anchor with a type guard. Output is **byte-identical** — every language's capture fingerprint is unchanged.

| Language | @800 entities | scaling ratio | source |
|----------|---------------|---------------|--------|
| **rust** | 5118 → **120 ms** | 3.31 → 1.01 | this PR |
| **python** | 11343 → **319 ms** | 3.34 → 1.05 | this PR (origin) |
| **csharp** | 2541 → **155 ms** | 3.13 → 0.99 | this PR |
| **php** | 1954 → **122 ms** | 3.04 → 1.05 | this PR |
| **ruby** | 2016 → **78 ms** | 3.07 → 1.10 | this PR (+ a 2nd O(n²): `out.some` per method → Set) |
| go | linear | 1.01 | #1915 (merged) |
| cobol | linear | 0.86 | n/a |

The per-match root-walk logic is shared via a single generic `nodeIfType<T>` helper in `ast-helpers.ts` (replacing four copy-pasted variants).

---

## 2. Python import resolution: O(imports × files) → O(imports + files)

[...]

PR #2872, opened 2026-08-07
perf(kotlin): index import resolution instead of scanning per import

glier:
## Summary

Kotlin import resolution walked the entire workspace on every import, across a cascade of four tiers, making it O(imports × files). This replaces the scans with three maps built once per file set and memoized on its identity, so each tier is O(1). Resolution behaviour is unchanged and gated by a fingerprint whose committed baseline is the value the pre-index implementation produces.

## Motivation / context

`resolveKotlinImportTarget` tried four tiers in cascade — exact/suffix, directory child, package fan-out, progressive prefix strip — and each one ran `for (const raw of allFilePaths)` with a `replace(/\\/g, '/')` and several string scans per entry. One unresolved import therefore cost two to four full passes over the workspace.

Once a repository reaches tens of thousands of Kotlin files that is on the order of 10^10 string operations on a single thread. It does not present as a hot loop, which is what makes it hard to spot: `analyze` sits at exactly 1.00 core with a completely flat heap and emits nothing for hours, because every allocation is a short-lived string and nothing accumulates to hint at progress. Small repositories hide it entirely — at a few hundred files each pass is free, which is why the existing test suite never noticed.

This is the same shape #1918 fixed for Python (issue #1741, "analyze gets stuck at 'Resolving calls'") and #2788 fixed for C++; neither touched the Kotlin path. I found it with a CPU profile that put 99.5% of self-time in `resolveKotlinImportTarget` under `makeEdgeDrafts`.

The bench added here reproduces it from scratch, so the cost is checkable without my profile: at only 400 files and 12.8k imports the pre-index resolver already spends 2.2 s where the index spends ~6 ms, and at 1600 files the gap is 33.0 s against ~24 ms — 3.737 versus ~1.0 on the scaling ratio. Both numbers come from the same corpus in the same harness; swapping one module specifier reproduces the first row.

I could not find an existing upstream issue for the Kotlin path specifically — happy to open one and re-title this PR if you would prefer the usual `(#NNNN)` reference.

## Areas touched

- [x] `.github/` (workflows, actions) — one step added to the existing `benchmarks (GITNEXUS_BENCH)` job

## Scope & constraints

**In scope**

- `languages/kotlin/import-target.ts`: three lookup maps (`exactByStem`, `suffixByStem`, `dirChildren`) built once per `allFilePaths` Set and memoized in a `WeakMap` keyed on its identity, mirroring Python's `getPythonFileIndex`.
- `languages/kotlin/index-stats.ts`: the build counter, mirroring `python/index-stats.ts`.
- `bench/kotlin-import-target/`: identity + scaling gate, wired into the existing benchmarks job next to the C++ (#2788) and Python (#1918) guards.
- Parity and index-reuse tests, named and placed to match the Python pair.

- **The directory-match rule is preserved as-is, including its bug.** The old scan tested `startsWith` and then used `indexOf`, so it only ever considered the FIRST occurrence of `/dir/`. Fixing it belongs in its own change with its own fixtures: it would move edges in every Kotlin repository, and that is a behaviour change that should not ride along with a performance one. The shape it actually breaks is narrower than it first looks — executed against a standard Gradle layout, `com.example.data.Repo`, `com.example.data.helper` and `data.Repo` all resolve correctly; only a bare single-segment key whose package sits at the repository root and collides with an ancestor directory name (`data.helper`) fails. Real Kotlin files under `…/com/example/data/` declare `package com.example.data`, so real imports take the multi-segment path that works. A follow-up issue should be scoped to that narrow shape.
- **Go, Dart and Ruby have the same defect and are left alone**, having no repository at hand to verify a change against:
  - `go/import-target.ts` — `findRootPackageFiles` and `findAllFilesInPkgDir` scan unmemoized, and the GOPATH fallback calls the latter once per path segment but the last, so one import can trigger several full passes. Most Go imports are external and always run the whole cascade, so this is likely the worst of the family;
  - `dart/import-target.ts` — the `package:` branch scans once per candidate path (`lib/<rel>` and bare `<rel>`), and `resolveRelative` scans again in its suffix fallback;
  - `ruby/import-target.ts` — rebuilds the entire `buildSuffixIndex` on every `require`, unmemoized at any layer.
- **C# is a partial case**: `getWorkspaceFileIndex` is already memoized, but it is only reached when a `.csproj` is found; the no-csproj path hands the raw Set to `resolveDirectMatch` and `resolveByProgressiveStripping`, which scan past it.
- Happy to file these as issues if you would rather track them that way than in this description.

## Implementation notes

The tie-breaks in this resolver were expressed only through Set-iteration order, so they are the part most at risk in a rewrite, and each is reproduced deliberately:

- an **exact match anywhere beats a suffix match found earlier** — the scan returned on the first exact hit but merely *remembered* the first suffix hit, which is why these are two maps rather than one lookup;
- **"first match" means first in set-iteration order**, so both stem maps keep the earliest path inserted per key;
- `suffixByStem` is keyed per **component-suffix of the stem**, not per basename, so a multi-segment import (`com/example/User`) hits one bucket instead of filtering a `User` bucket.

Rather than assert equivalence, the PR gates it. `bench/kotlin-import-target/measure.mjs` fingerprints every `fileSet | fromFile | targetRaw -> result` record over an exhaustive branch matrix plus a deterministic fuzz, resolving **each file set in both iteration orders** — a single-order corpus would score an implementation that keeps the *last* match identically. The committed baseline is the value the **pre-index implementation** produces: pointing the harness's module specifier at the old file and re-running prints the same `ebf1790b…` over the same 20,106 cases, 13,256 of them non-null. `--check` asserts all three together, because a hollowed corpus yields a perfectly valid fingerprint over a smaller surface.

The second arm is `(t_large/t_small)/(1600/400)` over a synthetic Kotlin monorepo whose imports are ~40% unresolvable, because only a miss drives all four tiers and that is where the scan was worst. Measured on a 12-core dev box:

| implementation | small (400 files, 12.8k imports) | large (1600 files, 51.2k imports) | ratio |
| --- | --- | --- | --- |
| pre-index scan | 2207.8 ms | 33003.5 ms | **3.737** |
| this change | 5.9 ms | 24.0 ms | **~1.02** |

so the budget of 1.6 separates them by a wide margin in both directions. The **ratios** are the comparable pair — each implementation's two arms are measured identically. Take the absolute times as an order of magnitude only: the pre-index row is a single cold run, because best-of-seven against a quadratic implementation costs minutes of wall time, while the index row is best-of-seven after two warmups. Both rows are author-measured; what CI reproduces is the index row and the budget. (Independent runs during review put the pre-index floor at 3.905–4.297 and the index between 0.95 and 1.08, so the separation is not an artifact of my machine.)

`--check` asserts three more things a single ratio cannot see, each added in response to review:

- a **depth ratio** — same file count, paths ~3× deeper. `scaling_ratio` divides the file count out, so it is scale-invariant and structurally blind to a cost that grows with path depth instead, which is exactly what both loops here do.
- an **absolute ceiling on the small arm**. A ratio cannot see a constant factor: a full workspace scan reintroduced on 1-in-32 imports scores 1.490, inside the budget, while running 2.8× slower.
- **`non_null` alongside `cases`**, so a corpus that stops resolving anything is not a valid baseline.

The memory cost is modest — I had overstated it in the first revision of this description. The index holds roughly `2 × (depth − 1)` map entries per Kotlin file, which measures **8.1 MiB at 10k files, 30.3 MiB at 50k and 60.1 MiB at 100k** (depth 8; 104.2 MiB at 100k with depth 12). It is released with the Set after each run — `allFilePaths` is a local in `runScopeResolution` (`scope-resolution/pipeline/run.ts`), captured only by the `finalizeScopeModel` closure, so the WeakMap entry does not survive the run and `gitnexus serve` cannot accumulate indexes across analyses.

Worth stating plainly because an inflated number invites the wrong optimization. Review built the obvious alternative — a Python-style basename-bucket index — proved it byte-identical in output, and measured it at ~66% less memory but **7.3× slower per query** on a repeated-basename corpus, failing this PR's own 1.6 scaling budget at ~2.0. I have recorded that in the code next to the `suffixByStem` description so the next reader does not re-derive it.

## Testing & verification

- [x] `cd gitnexus && npm test` — ran; see the note on pre-existing failures below
- [x] `cd gitnexus && npx vitest run test/integration` — covered by the full run above
- [ ] `cd gitnexus-web && npm test` *(web not touched)*
- [ ] Manual / Playwright E2E *(not applicable)*

Also run:

```
cd gitnexus
npx prettier --check .                                            # clean
node --import tsx bench/kotlin-import-target/measure.mjs --check   # PASS

npx eslint \
  src/core/ingestion/languages/kotlin/import-target.ts \
  src/core/ingestion/languages/kotlin/index-stats.ts \
  bench/kotlin-import-target/measure.mjs \
  test/unit/scope-resolution/kotlin/kotlin-import-target-parity.test.ts \
  test/integration/kotlin-import-index-reuse.test.ts             # no errors, no warnings
```

Every mutation named in this description was re-run against the gate after the fix, not assumed: capping suffix-key depth at 7, skipping the `dirChildren` suffix loop above depth 8, capping a bucket at 17 entries, and the corpus edit that gutted two documented cases — all four now move the fingerprint, where all four previously passed. The two test mutations (relaxing the `indexOf` position check to `>= 0`, and `raw.lastIndexOf` for `norm.lastIndexOf`) now fail the unit suite, where both previously passed 18/18.

`npm test` is not green on my machine: 30 failures across 10 files (`resolvers/go`, `resolvers/typescript`, `analyzer-identity`, `evidence-provenance-helper`, `review-agent-workflow`, `cli-limit-e2e`, `analyze-index-lock-concurrency`, and three FTS-dependent suites). None are from this branch — I checked out a clean `main` and ran exactly those ten files, which produced exactly the same 30 failures. Kotlin is not among them, and the 20 tests this PR adds all pass. If they are green in your CI, they are environment-specific to me and worth ignoring.

Beyond the suites above, the change was exercised end to end: a large Kotlin codebase that previously could not finish `analyze` at all now completes and produces a full graph. The reproducible part of that claim is the bench — it derives the same asymptotic difference from a synthetic corpus anyone can run.

## Risk & rollout

No contract changes: `resolveKotlinImportTarget`'s signature, the `WorkspaceIndex` shape, graph node/edge shapes and persisted IDs are all untouched. No index refresh is required for correctness — the resolver returns the same targets — though Kotlin repositories that previously could not finish `analyze` will now produce a graph where they produced none.

Rollback is a straight revert; nothing persists state that would outlive it.

The one thing to watch after merge is the new benchmarks step: its fingerprint arm is deterministic and a failure there is always a behaviour change, but its scaling arm is a timing gate on a shared runner and should be re-run on an idle machine before it is investigated. Both rules are written into `baselines.json`.

On the `.github/` edit: it is one additive `run:` step in the existing `benchmarks (GITNEXUS_BENCH)` job — no new job, no permissions, secrets, triggers or `concurrency` change, and the command is a local `node --import tsx …` with no network access. I included it because a perf fix whose gate is not wired in is a gate that silently stops existing, and because #1918 and #2788 both shipped their bench and its CI step together. If you would rather take the code change first and wire the bench separately, say so and I will split it — the bench passes standalone either way.

## Checklist

- [x] `AGENTS.md` / overlays not changed


koriyoshi2041 (approved):
Verified the indexed resolver preserves the legacy resolution tiers and iteration-order tie-breaks while reusing one index for a stable workspace file set. The focused Kotlin parity/reuse suites pass locally (255 tests), and the hosted cross-platform matrix is green. The `WeakMap` lifetime also matches the per-pass `allFilePaths` identity used by the scope resolver.

gitnexus/bench/kotlin-import-target/measure.mjs
```diff
@@ -0,0 +1,376 @@
...
+/** Record one case in BOTH file-set iteration orders — see header property 1. */
+function record(files, targetRaw, fromFile = 'App.kt') {
+  for (const [order, list] of [
+    ['fwd', files],
+    ['rev', [...files].reverse()],
+  ]) {
+    const r = resolve(list, targetRaw, fromFile);
+    if (r !== null) nonNull++;
+    const rendered = r === null ? 'NULL' : Array.isArray(r) ? `[${r.join(',')}]` : r;
+    lines.push(`${order}\t${fromFile}\t${targetRaw}\t${rendered}`);
```
magyargergo: **P1 — the fingerprint does not bind the corpus's file sets** `[reproduced]`

*Codex (gpt-5.6-sol) adversarial leg + a coordinator control; reached independently from the other direction by a Claude adversarial lane.*

The hashed record carries `order | fromFile | targetRaw | result` — **not `files`**. A corpus edit that changes *which workspace* a case resolves against, while leaving the result string identical, is invisible to the gate.

Control: leaving the resolver **completely untouched** and editing only the corpus, two of the bench header's own documented load-bearing cases were gutted —

- `record(['deep/util/User.kt', 'util/User.kt'], 'util.User')` → `record(['util/User.kt'], 'util.User')`, so the "exact match anywhere beats a suffix match found earlier" case no longer tests the precedence rule at all (result string unchanged);
- `record(['data/src/main/kotlin/com/example/data/Repo.kt'], 'data.something')` → `record([], 'data.something')`, so the repeated-directory negative case loses its file entirely (still `NULL`).

Result: `cases 20074`, `non_null 11612`, fingerprint `5ad605c1…` — all byte-identical — and `[kotlin-import-target --check] PASS`. The advertised edge-case coverage can be deleted while CI stays green.

Fix is one line, then one deliberate re-baseline:
```js
lines.push(`${order}\t${list.join('|')}\t${fromFile}\t${targetRaw}\t${rendered}`);
```

gitnexus/bench/kotlin-import-target/measure.mjs
```diff
@@ -0,0 +1,376 @@
...
+const DIRS = [
+  '',
+  'app',
+  'core',
+  'data',
+  'feature/home',
+  'lib/data',
+  'src/main/kotlin',
+  'src/main/kotlin/com/example',
+  'module/src/main/kotlin/com/example/data',
```
magyargergo, line 294: **P1 — the corpus caps path depth at 8 components, and that is one of the three loops this PR added** `[reproduced]`

*A Claude adversarial lane + a coordinator control.*

This is one of the two deepest `DIRS` entries (`:202` is tied) — 7 directories + file. `buildCorpus` (`:248`) is pinned to the same shape, `lib<m>/src/main/kotlin/com/example/mod<m>/Class<i>.kt`. Real Android/Gradle source routinely runs 9–11 components.

Because of that, this mutation of `import-target.ts` passes `--check` with a **byte-identical** fingerprint, `cases`, *and* `non_null`:

```ts
const depthGuard = norm.split('/').length <= 8;
for (let i = 0; depthGuard && i < dir.length; i++) {   // skip deep paths
```

Measured against a standard Android/Gradle corpus (4 modules × 4 packages × 40 files, 9 components, 32 imports):

| | resolved to NULL | single-file | fan-outs |
|---|---|---|---|
| PR as written | 0 | 16 | 16 (avg 40 files) |
| with the mutation — **gate PASSES** | **16** | 16 | **0** |

Every package fan-out import resolves to `NULL` — the whole `#1759` multi-file tier dies — and CI is green. This is the same defect class #2794's C++ review found (its corpus drew every receiver from a declared namespace and was blind to the miss path) and fixed by widening the corpus.

Fix: add one `DIRS` entry at 10–12 components (e.g. `core/data/src/main/kotlin/com/example/core/data/repository`) plus a matching `record(...)` case resolved by suffix, and re-baseline once, deliberately.

gitnexus/bench/kotlin-import-target/measure.mjs
```diff
@@ -0,0 +1,376 @@
...
+}
+function randDotted() {
+  const n = 1 + Math.floor(rnd() * 4);
+  const parts = [];
+  for (let i = 0; i < n; i++) parts.push(pick(SEGS));
+  return rnd() < 0.12 ? `${parts.join('.')}.*` : parts.join('.');
+}
+
+for (let repo = 0; repo < 400; repo++) {
+  const fileCount = 3 + Math.floor(rnd() * 14);
```
magyargergo: **P1 — the corpus caps package size at 16 files, so a fan-out truncation is invisible** `[reproduced]`

*A Claude adversarial lane + a coordinator control.*

`3 + Math.floor(rnd() * 14)` tops out at 16, so no `dirChildren` bucket in the fuzz corpus ever exceeds 16 entries. A cap at 17 in the index build therefore changes nothing the gate can see:

```ts
else if (bucket.length < 17) bucket.push(raw);   // gate: cases 20074, non_null 11612, fingerprint 5ad605c1…, PASS
```

On the Android/Gradle corpus above, that silently truncates package fan-out from **40 candidate files to 17** — a 58% edge loss on every package import, with CI green.

This is not a hypothetical shape for this repo: a silent sibling cap is exactly what #2732 shipped on the JVM side.

Fix: raise `fileCount`, or add one explicit 40-file fan-out case above any plausible cap.

magyargergo (commented):
# Two-engine review of #2872

**How this was produced.** Two engines, not eighteen independent confirmations. **Engine A (Claude)** ran the GitNexus PR swarm (risk-architect, test-ci, security, facts-historian, branch-hygiene) plus ten `ce-code-review` personas — 15 lanes, all sharing one model. **Engine B (Codex, gpt-5.6-sol/xhigh)** ran the same swarm and `ce` scaffolding plus a native adversarial pass — 3 legs, all live and complete. On top of those, the coordinator ran **four controls** of its own: an old-vs-new differential, a perf/memory measurement, a gate-mutation control, and a CI/corpus-binding control. Agreement *within* an engine is consistency; agreement *across* engines is the real signal, and it is labelled below. Where a lane's number and a coordinator measurement disagree, the measurement wins and I say so.

---

## What is solid — and this PR earned an unusual amount of it

**Semantic equivalence is established, not asserted.** Four independent differentials, all zero divergences:

| source | scope | result |
|---|---|---|
| coordinator control | 96,060 adversarial cases, both iteration orders | **0** divergences |
| Codex B1 (engine B) | 85,544 ordered adversarial cases | **0** |
| ce-correctness (engine A) | set-equality proof over the **entire reachable key space** (883,438 file×key evaluations) + a 1,341,648-run exhaustive cascade + 193,719 fuzz | **0**, incl. bucket *order* |
| ce-adversarial (engine A) | 640,000 cases (480k adversarial alphabet + 160k shared-Set reuse) | **0** |

The coordinator's harness is itself controlled: five mutations of the at-risk tie-breaks (exact/suffix first-vs-last insert, dropping the `startsWith` skip, relaxing the `indexOf` position check, swapping exact/suffix precedence) produce 22 / 3,008 / 1,087 / 4,089 / 754 mismatches respectively, and the unmutated code produces 0. Every rule the PR claims to preserve is one the harness demonstrably catches when broken.

**The `_provenance` claim is honest and reproducible.** Three sources independently reproduced fingerprint `5ad605c1…` over 20,074 cases / 11,612 non-null by pointing the bench at the pre-index implementation. The floor arm is real too: measured 3.905–4.297 for the pre-index scan against 0.95–1.04 for the index.

**The absence of a `clearKotlinFileIndex` reset hook is correct, not an omission.** C++ needs one because its WeakMap key under-covers its inputs (`cpp/adl.ts:397-403` says so verbatim); `getKotlinFileIndex` is a pure function of `allFilePaths`, which *is* the key. Four Claude lanes and Codex B1 agree.

**The obvious alternative was tested and refuted.** ce-performance built the Python-style basename-bucket variant, proved it byte-identical, and measured it: 66% less memory but **7.3× slower per query** on a repeated-basename corpus, failing this PR's own 1.6 scaling budget at **2.01–2.12**. The per-suffix keying and its inline justification at `import-target.ts:131-133` are measurably correct. Please do not "optimize" this later.

**`kotlin-import-index-reuse.test.ts` proactively guards the #1918 P1 trap** — the adapter-side `new Set(...)` copy that silently defeated Python's memoization. Mutation-confirmed: inserting that copy in `scope-resolver.ts` flips the assertion to `expected 200 to be 1`.

**The `benchmarks (GITNEXUS_BENCH)` job IS effectively merge-required** via `ci.yml:95` `CI Gate` → `needs.tests.result`; the job has no `if:`, no path filter, no `continue-on-error`. It ran and **passed (8m33s)** on this PR. The recurring worry that a correctness fingerprint sits in a non-required timing job is refuted for this repo. Security lane is clean on every axis (no `pull_request_target` anywhere, no secrets at the call site, `persist-credentials: false` preserved, no bidi/hidden Unicode).

**Also refuted, explicitly:** stale-index-from-a-mutated-Set; worker/thread identity divergence; the index pinning a long-lived `serve`/MCP process; "20,074 cache misses will time out the job" (the step takes ~1 s against a 25-min budget); machine/Node-dependent fuzz; "the tests are tautologies"; a vitest `if`-branching violation; and a `SCHEMA_BUMP`/`PARSE_CACHE_VERSION` obligation (resolution-only change, matching both the #1918 and #2794 precedents). One of our own lanes claimed the merge commit rolled `SCHEMA_BUMP` 48→45 — **that is false**: the constant is 48 at merge-base and at head, and `parse-cache.ts` is not in the diff.

---

## The headline: the gate is blind, the code is not

Every finding worth acting on is in the **gate**, not the resolver. Two independent mechanisms — one surfaced by each engine — both reproduced by the coordinator.

**(i) The corpus's bounds sit below real-world values.** `measure.mjs` fixes path depth at 8 components (deepest `DIRS` entry + `buildCorpus`), package size at ≤16 (`:224`), and therefore stem-suffix count at ≤7. Each bound corresponds to exactly one loop this PR added. Three plausible follow-up mutations pass `--check` with a **byte-identical** fingerprint, `cases`, *and* `non_null`:

| mutation | gate | on a standard Android/Gradle corpus (4 modules × 4 packages × 40 files, 9 components) |
|---|---|---|
| cap suffix-key depth at 7 | **PASS** | benign here (reported honestly, not inflated) |
| cap `dirChildren` bucket at 17 | **PASS** | fan-out silently truncated **40 → 17** — a 58% edge loss on every package import |
| skip `dirChildren` suffixes for depth > 8 | **PASS** | **16/16 package fan-out imports resolve to NULL** — the entire #1759 tier dies, CI green |

**(ii) `measure.mjs:109` does not hash the file set.** Holding the resolver *completely untouched* and editing only the corpus, two of the bench header's own documented load-bearing cases can be gutted — the "exact match anywhere beats an earlier suffix" case loses its competing file, the repeated-directory negative case loses its file entirely — and `cases`, `non_null` and the fingerprint all stay byte-identical. `PASS`.

**Consequence, and it matters:** the `non_null` assertion that **both engines independently proposed** (Claude test-ci P1, Codex B2 #2) is **necessary but not sufficient** — `non_null` is 11612 under all three code mutations *and* under the corpus edit. Please do all three: **hash the file set, widen the corpus bounds, assert `non_null`.**

**Honest negative:** a fourth mutation — capping tier-4 progressive prefix strip at 4 dotted segments — **FAILED** the gate (`non_null` 11612→11608, fingerprint drifted), because the hand-written branch matrix carries 6- and 8-segment tier-4 cases. **Import arity is *not* a blind axis.** The three blind axes are the ones the corpus fixes as constants: depth, package size, suffix count.

This is the same defect class #2794's C++ review already hit and fixed by widening its corpus. It is a corpus change, not a code change.

---

## Inline comments

Ten are attached inline. In short:

| # | where | sev | what |
|---|---|---|---|
| 1 | `measure.mjs:109` | P1 | the fingerprint does not hash the file set, so corpus cases can be gutted with CI green |
| 2 | `measure.mjs:201` | P1 | corpus path depth caps at 8 components — a depth-skip mutation kills every Android fan-out import, gate green |
| 3 | `measure.mjs:224` | P1 | corpus package size caps at 16 — a fan-out cap at 17 is invisible |
| 4 | `baselines.json:5` | P2 | `non_null` computed and reported, never asserted (both engines) — necessary, not sufficient |
| 5 | `import-target.ts:97` | P2 | the index's live mutable bucket escapes the module; `readonly` is erased at the one consumer |
| 6 | `kotlin-import-target-parity.test.ts:110` | P2 | this test pins the `startsWith` guard, not the `indexOf` rule its comment describes |
| 7 | `kotlin-import-index-reuse.test.ts:47` | P3 | the only adapter-driving test asserts nothing about what it returns, and its premise is measurably wrong |
| 8 | `import-target.ts:151` | P3 | no note on why the shared `buildSuffixIndex` isn't reused — the one place this PR is less documented than its own precedent |
| 9 | `baselines.json:7` | P3 | `_scaling_note`'s observed band is narrower than measured |
| 10 | `import-target.ts:43` | P3 | `stripped` computed eagerly on every import; a lazy form measured −20% (optional) |

---

## Lower priority, and two corrections to the PR body

**The PR body overstates its own memory cost by ~10–30×.** Body: *"a few hundred MB at tens of thousands of files … low GB at hundreds of thousands."* Measured: **8.1 MiB @10k**, 30.3 MiB @50k, **60.1 MiB @100k** (depth 8), 104.2 MiB @100k (depth 12); two other lanes measured 180 MB and 126 MB at **200k**. The #2649 kernel-scale OOM concern is **refuted** — 60 MiB against a resolution phase that otherwise never terminates is not a trade worth hesitating over. (Our own risk-architect *estimated* 300–400 MB @100k; that is ~6× high — prefer the measurements.) Worth correcting in the `WHY` block, because an overstated cost invites exactly the basename-bucket "optimization" that measurement already refuted.

**The PR body also overstates its own preserved bug.** Executed against the PR-head resolver on the standard Gradle layout: `com.example.data.Repo` resolves, `com.example.data.helper` **fans out correctly**, `data.Repo` resolves. Only a bare single-segment root-level key (`data.helper`) breaks — a package at the repository root whose name collides with an ancestor directory. Real Kotlin files in `…/com/example/data/` declare `package com.example.data`, so real imports take the multi-segment path that works. This makes preserving the rule *safer* than the body argues, and the follow-up issue should be scoped to that narrow shape, not "every Gradle repo."

**Codex B1's P2 / REQUEST CHANGES is demoted to P3 by measurement.** B1 constructed a genuinely sharp case — 100,001 files ordered so the tier-1 hit is first, one import: OLD 0.0 ms, NEW 779 ms. Credit for building it. But it is bounded by a single workspace pass (~0.8 s at 100k files) and **inverts at ~15 imports** (at k=20: OLD 1,042 ms vs NEW 751 ms; at k=100: 5,283 vs 732). In the polyglot case B1 was really worried about — 100k files, 5% Kotlin, 2 imports — **the new code already wins, 27.6 ms vs 33.8 ms**, because the build skips non-`.kt` entries as cheaply as the scan did. Worth one sentence in the bench header, not a change request. B1's *other* half is real and lands as a gate note: build cost is roughly `O(files × depth²)` (25.7 ms → 1,895.8 ms as depth goes 2 → 64 at 20k files) and `buildCorpus` pins depth at a constant, so the scaling arm cannot see a depth regression at all. A third corpus arm holding file count fixed while doubling depth would close it.

**Also noted, not blocking:** `scaling_ratio` is scale-*invariant*, so it cannot see a constant-factor regression — a reintroduced full scan firing on 1-in-16 imports is caught (1.814) but 1-in-32 passes (1.490) while being 2.8× slower absolute; an absolute small-arm ceiling would cover it. `KOTLIN_EXTENSIONS` at `:149` shadows the existing export at `import-resolvers/jvm.ts:13` (net still an improvement — it collapses three inlined literals). The empty-string `''` key inserted for degenerate paths is unreachable (all three query sites guard it) — verified, not a finding.

**Follow-ups to file with this PR** (per #1918's own recorded lesson, "grep all languages — it's a copy-paste family bug"), with measured ratios where a lane produced them:

- **Go** — `go/import-target.ts:44-48,64-75` — ratio **4.00**, ~370 µs/import at 1,600 files; worst of the family because most Go imports are external and always run the full GOPATH cascade, with a fresh `'/' + raw.replace(...)` allocation per file per scan. Take this one first.
- **C# (no-csproj)** — ratio **3.43**, ~359 µs/import; up to 8 full scans for an unresolved 4-segment `using`, and the #1881 namespace gate **fails open precisely at large-repo scale**.
- **Dart** — ratio **4.05**, ~89 µs/import; the `package:` branch runs two full scans that always miss for external packages.
- **Ruby** — *not in the PR body's list*: `ruby/import-target.ts:102-108` rebuilds the entire `buildSuffixIndex` on **every `require`**, unmemoized at any layer. Arguably a worse constant factor than the C#-no-csproj gap that is disclosed.
- The narrowly-scoped Kotlin first-occurrence issue, and an issue number next to `import-target.ts:199-202` so the preserved-bug intent survives the baseline.

---

## CI status, honestly

`tests / benchmarks (GITNEXUS_BENCH)` **passed (8m33s)** — the new Kotlin gate ran green on CI infrastructure, not just on the author's machine.

But at 18:16Z against head `626dafa2`, **a required check is red**: `tests / windows-latest (platform-sensitive) 1/3` failed on `cli-e2e.test.ts > analyze persists meta.json AND a matching registry entry (#1169)` with *"analyze timed out before asserting finalization artifacts"* (1 failed / 312 passed). This is **almost certainly not this PR** — it is a subprocess timeout rather than a resolution assertion, the fixture is `mini-repo`, `cli-e2e.test.ts` is a documented flaky-under-parallel-load suite (the sibling test took 46.7 s and the file took 495 s on that run), and this PR touches only Kotlin import resolution. But `CI Gate` is the single required check and it is red, so **the shard needs a re-run before merge regardless of cause**. Worth flagging that only one of eighteen lanes caught this — every other lane, and the PR's own bot comment, read the earlier all-green report for a previous commit.

---

*Automated multi-tool digest (two engines + four coordinator-run controls). Verify before acting.*

gitnexus/bench/kotlin-import-target/baselines.json
```diff
@@ -0,0 +1,8 @@
...
+  "fingerprint": "5ad605c179081505705ff7698a09dbdbdc4831080af6d9fdec5499cc6bce28ee",
+  "cases": 20074,
```
magyargergo: **P2 — `non_null` is computed and reported but never asserted** `[reproduced]`

*Independently proposed by **both** engines: a Claude test/CI lane and the Codex `ce` leg.*

`measure.mjs:336` emits it and the comment at `:332-334` acknowledges the corpus-shrink risk, but only `fingerprint`, `cases` and `scaling_ratio` are compared at `:347-369`. Adding it beside `"cases": 20074` gives a deliberate re-baseline a magnitude to reason about:

```json
"cases": 20074,
"non_null": 11612,
```

**Worth knowing before you spend time on it:** this closes the "resolver returns null everywhere" hole and *nothing else*. `non_null` is **11612 under all three code mutations in the sibling comments and under the corpus edit** — it is not a substitute for hashing the file set or widening the corpus bounds. Please treat the three as one fix, not alternatives.

gitnexus/src/core/ingestion/languages/kotlin/import-target.ts
```diff
@@ -118,41 +92,131 @@ function findKotlinDirectoryChild(
...
-    const idx = atRoot ? 0 : file.indexOf(suffixDirPrefix) + 1;
-    const after = file.slice(idx + dirPrefix.length);
-    // Direct children only — `models/sub/Util.kt` is a different package
-    // (`models.sub`) and must not be merged with `models`.
-    if (after.length === 0 || after.includes('/')) continue;
-    out.push(raw);
-  }
-
-  return out.length === 0 ? null : out;
+  return index.dirChildren.get(dirPath) ?? null;
```
magyargergo, line 107: **P2 — the index's live mutable bucket escapes the module** `[reproduced]`

*Claude: correctness, adversarial, architecture, maintainability, api-contract, risk-architect. Codex: the adversarial leg filed it too.*

**Engine disagreement, settled on evidence.** The Codex `ce` leg refuted this "for all current consumers" — correct as far as it goes: the sole production consumer, `finalize-algorithm.ts:395`, only `.map()`s the value. But an executed `tsc --noEmit --strict` probe on exactly that shape returns **zero errors** for both `.sort()` and `.push()`, because `Array.isArray(arg): arg is any[]` widens the true branch of `Array.isArray(t) ? t : [t]` and erases the `readonly`. The type gives no protection at the one place it would need to.

The pre-index code built `const out: string[] = []` per call. This returns the cached bucket — the same array `findKotlinDirectoryChild` reads `children[0]` from (`:79-83`). A lane reproduced the cascade: a downstream `.sort()` (a plausible edge-ordering change in this repo) permanently flips a **different tier's** answer for every later import in the same run, and no existing test can see it because every test surface mints a fresh `Set` per query.

Not a live bug — a contract hazard the type system does not catch. One line at the end of the build loop makes `readonly` true at runtime and turns a future mutation into a loud `TypeError` instead of a silent edge move:

```ts
for (const b of dirChildren.values()) Object.freeze(b);
```

Please do **not** fix it with a per-call `[...bucket]` copy: a performance lane measured that this bench's corpus over-weights bucket size roughly 4× versus production, so a copy would be scored far worse here than it deserves.

gitnexus/test/unit/scope-resolution/kotlin/kotlin-import-target-parity.test.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  it('backslash paths are normalized', () => {
+    expect(resolve(['win\\pkg\\A.kt'], 'win.pkg.A')).toBe('win\\pkg\\A.kt');
+  });
+
+  it('non-Kotlin files never resolve', () => {
+    expect(resolve(['pkg/A.java', 'pkg/A.md'], 'pkg.A')).toBeNull();
+  });
+
+  it('only the FIRST occurrence of a repeated directory name counts', () => {
```
magyargergo, line 110: **P2 — this test pins the `startsWith` guard, not the `indexOf` rule its own comment describes** `[reproduced]`

*A Claude testing lane, mutation-confirmed; re-confirmed by the coordinator.*

The comment says the scan *"tested `startsWith` first and then used `indexOf`"*, but the fixture puts `data` at the **leading** segment — so `import-target.ts:206`'s `startsWith` skip fires and the `indexOf` equality at `:207` is never reached.

Confirmed by running the committed suite against a mutated resolver (`:207` relaxed from `=== dir.length - suffix.length - 1` to `>= 0`):

```
Test Files  1 passed (1)
      Tests  18 passed (18)
```

18/18 green with the rule broken. Only the bench fingerprint catches it — and a coordinator differential scores 4,089 mismatches on that same mutation, so the rule is real and load-bearing; it simply is not pinned here.

Fix — promote the bench's own mid-path case, which exercises `:207` without touching `:206`:
```ts
it('a repeated directory name below the root still only counts its first occurrence', () => {
  expect(resolve(['top/data/mid/data/Repo.kt'], 'data.something')).toBeNull();
});
```

The same lane found a second bench-only rule: swapping `norm.lastIndexOf('/')` for `raw.lastIndexOf('/')` at `:179` also passes 18/18. A backslash + fan-out case would pin it.

gitnexus/test/integration/kotlin-import-index-reuse.test.ts
```diff
@@ -0,0 +1,75 @@
...
+describe('Kotlin import resolution — index reuse across imports', () => {
+  it('builds the file index once for many imports over a stable file set', () => {
+    const files = buildWorkspace(300);
+    resetKotlinFileIndexBuildCount();
+
+    const resolveImportTarget = kotlinScopeResolver.resolveImportTarget;
+    expect(resolveImportTarget).toBeDefined();
+
+    for (let i = 0; i < 200; i++) {
+      resolveImportTarget?.(
```
magyargergo: **P3 — the only test that drives the production adapter asserts nothing about what it returns** `[reproduced]`

*Both engines: four Claude lanes and the Codex `ce` leg.*

200 calls, every return value discarded; the sole assertion is the build count at `:56`. If the synthetic `ParsedImport` this adapter mints ever drifts, the count stays 1 and Kotlin IMPORTS edges vanish with the suite green. The Python twin has exactly this non-vacuity case (`python-import-index-reuse.test.ts:73-86`).

The docstring's premise is also measurably wrong. It says the workspace is shaped *"where imports miss the cheap tiers and reach the package fan-out, so every call would have scanned the whole set"*. Replaying its exact corpus and its exact 200 imports against the PR-head resolver:

```
{"singleFileString":200,"packageFanOut":0,"unresolved":0}
```

All 200 hit the tier-1 **suffix** lookup. Zero reach `findKotlinPackageFiles`.

Also: `undefined as never` is unnecessary — both parameters are optional (`contract/scope-resolver.ts:380-385`) and the Python twin calls the same method with three arguments; and `resolveImportTarget` is a **required** member of `ScopeResolver`, so the `?.` and the `toBeDefined()` guard an unreachable branch (which also brushes the repo's no-`if`-branching-in-vitests rule).

Fix: drop the two casts and the `?.`, assert the results, and query something like `com.example.widget.someTopLevelFun` so some imports genuinely reach the tier the comment claims.

gitnexus/src/core/ingestion/languages/kotlin/import-target.ts
```diff
@@ -118,41 +92,131 @@ function findKotlinDirectoryChild(
...
+  readonly exactByStem: Map<string, string>;
+  readonly suffixByStem: Map<string, string>;
+  readonly dirChildren: Map<string, string[]>;
+}
+
+const KOTLIN_FILE_INDEX_CACHE = new WeakMap<ReadonlySet<string>, KotlinFileIndex>();
+
+const KOTLIN_EXTENSIONS = ['.kt', '.kts'] as const;
+
+function getKotlinFileIndex(allFilePaths: ReadonlySet<string>): KotlinFileIndex {
```
magyargergo, line 185: **P3 — the one place this PR is less documented than its own precedent** `[reproduced]`

*Claude: architecture, pattern-reuse, simplicity lanes. The Codex swarm leg reached the same conclusion independently.*

There is no note on why the shared `buildSuffixIndex` (`import-resolvers/utils.ts:90`, already used by C#, Ruby, Vue and TypeScript) isn't reused here. Python documents exactly this at `python/import-target.ts:214`.

The decision is **right** — run side by side, 4 of 5 probes diverge:

| probe | Kotlin | `buildSuffixIndex` |
|---|---|---|
| `['deep/util/User.kt','util/User.kt']` → `util.User` | `util/User.kt` | `deep/util/User.kt` — its single `exactMap` conflates exact and proper-suffix |
| `['deep/util/User.kt','util/User.kts']` → `util.User` | `util/User.kts` | `deep/util/User.kt` — keys carry the extension, so a `.kt` **suffix** beats a `.kts` **exact** |
| `['data/src/…/data/Repo.kt']` → `data.getRepo` | `null` | fan-out — `getFilesInDir` indexes *all* directory suffixes, with no first-occurrence rule |
| `['models/A.kts','models/B.kt']` → `models.getThing` | both, in set order | split into `:kt` and `:kts` buckets |

Three lines here stop the next person from "consolidating" it and silently moving edges — which matters because the PR body names Go, Dart and C#-no-csproj as next, and each will face the same question.

gitnexus/bench/kotlin-import-target/baselines.json
```diff
@@ -0,0 +1,8 @@
...
...l arm ~8 ms. IMPORTS_PER_FILE is sized so the small arm lands in the tens of ms rather than the ~2 ms the first revision measured, where timer granularity and JIT warm-up — not scaling — set the number; bench/cpp-qualified-ns documents the same artifact. TRIAGE: a scaling failure is a TIMING signal — RE-RUN IT on an idle machine before investigating; runner contention dominates. The fingerprint arm is the opposite: deterministic, a re-run never changes it, and it must never be wished away. FLOOR CHECK: the pre-index implementation — i.e. exactly the regression this gate exists to catch — measures 2207.8 ms small, 33003.5 ms large, ratio 3.737 on this same corpus, against 8.0 ms / 31.7 ms and 0.99 for the index. The budget of 1.6 therefore separates the two by a wide margin in both directions. Read the absolute times as an order of magnitude only (~276x small, ~1041x large): the floor arm was run once cold, because best-of-seven against the quadratic implementation costs ~5 minutes of wall time, while the index arm is the usual best-of-seven after two warmups. The RATIOS are the comparable pair — each is internally consistent, both arms of each implementation measured the same way."
```
magyargergo: **P3 — the observed band recorded here is narrower than what the gate actually measures** `[code-read]`

*Two Claude lanes plus the Codex `ce` leg, which raised the same axis as "the timing gate is non-deterministic".*

The note records `OBSERVED BAND: 0.98-1.00 over three unloaded runs`. Three independent measurements on other hardware came out wider: **0.954–1.014** (five runs), **0.965–1.036** (four runs), and **~0.95–1.08**.

The 1.6 budget still separates cleanly from a measured 3.9–4.3 floor for the pre-index scan, so this is a note fix rather than a budget fix. But the recorded band is what a future maintainer will triage against, and a 1.2 reading should be read as noise per this file's own TRIAGE guidance — the band should say so.

Related, and worth one sentence somewhere in this file: `scaling_ratio` is scale-*invariant*, so it structurally cannot see a constant-factor regression. A probe injecting a full workspace scan on 1-in-16 imports is caught (1.814), but 1-in-32 **passes at 1.490** while running 2.8× slower in absolute terms. An absolute small-arm ceiling (say 40 ms against the observed 8–11 ms) would cover that; the ratio never will.

gitnexus/src/core/ingestion/languages/kotlin/import-target.ts
```diff
@@ -38,75 +39,48 @@ export function resolveKotlinImportTarget(
   //     export the imported name (#1759).
   //  4. Progressive prefix strip for deeper namespace aliases that
   //     don't map 1:1 to directories.
+  const index = getKotlinFileIndex(ctx.allFilePaths);
   const stripped = pathLike.split('/').slice(0, -1).join('/');
```
magyargergo: **P3, optional — `stripped` is computed eagerly for every import, even when tier 1 hits** `[reproduced]`

*A Claude performance lane.*

Two array allocations and a string per import, discarded whenever `findKotlinFile` resolves on the line below. A lazy `lastIndexOf`/`slice` form, plus two build-loop tweaks, measured **−20% small arm / −21% large arm** in an interleaved A/B with the fingerprint unchanged, and build time 1807 → 1598 ms at 200k files.

Note the `li < 0` guard is load-bearing if you do this — `'a'.slice(0, -1)` is `''`, not `'a'`.

Entirely optional on a change that already turned hours into seconds; noted only because it is free and the measurement was taken.

glier:
Thanks for the review — the gate was the weak part, and the mutations made that easy to act on. Fixes are in `d2195d44`. Results below.

## Gate

**The fingerprint now includes the file set.** Record is `order | fileSet | fromFile | targetRaw | result`. Your corpus edit (resolver untouched, two documented cases gutted) now gives `44df5093ee…` instead of matching the baseline.

**The corpus spans depth and package size.** It carries 11- and 13-component paths, a 40-file package, queries against suffix keys deeper than seven segments, and a fuzz with file counts to 45 and targets to 9 segments. Your three mutations:

| mutation | before | now |
|---|---|---|
| cap suffix-key depth at 7 | PASS | caught — `a0e6eb98f9…` |
| skip `dirChildren` suffixes above depth 8 | PASS | caught — `d53182ebbc…`, `non_null` 13256 → 12746 |
| cap `dirChildren` bucket at 17 | PASS | caught — `ed3ea85c59…` |

One thing worth noting: deep paths alone did not catch the first mutation. Every Gradle-shaped import is six segments or fewer, so the deep keys existed but nothing queried them. It needed explicit queries against those keys.

**`non_null` is asserted next to `cases`**, with your caveat written into `baselines.json` — it stayed 13256 under all three mutations and under the corpus edit, so it only closes the "resolves nothing" case.

**Two new arms:**

- `depth_ratio` — same file count, paths 24 components against 8. Observed 1.44–1.51, budget 2.4. This covers the `O(files × depth²)` build cost B1 measured.
- `small_ms_ceiling` — 40 ms against an observed 5.9–6.1 ms, for the constant-factor case your 1-in-32 probe showed (1.490, inside the ratio budget, 2.8× slower).

## Code

- `dirChildren` buckets are frozen at the end of the build loop, not copied per call. The `Array.isArray` widening is in the comment.
- `stripped` is lazy now, `lastIndexOf`/`slice` with the `li < 0` guard. −20% / −21%, same as your measurement.
- `KOTLIN_EXTENSIONS` comes from the `import-resolvers/jvm.ts` export.
- Added the `buildSuffixIndex` note with your four diverging probes, plus the basename-bucket measurement (byte-identical, ~66% less memory, 7.3× slower per query) next to `suffixByStem`.

## Tests

- Parity test: added `top/data/mid/data/Repo.kt` for the `indexOf` rule, a positive control, and a backslash fan-out case for `norm.lastIndexOf` vs `raw.lastIndexOf`. Both mutations now fail the suite; both passed 18/18 before.
- Reuse test: asserts results now. You were right that all 200 imports hit tier 1 and none reached the fan-out — half of them do now, and the docstring says what actually happens. Casts, `?.` and `toBeDefined()` removed. The defensive-copy trap still gives `expected 200 to be 1`.

## PR body

Memory figures replaced with your measurements (8.1 MiB @10k, 60.1 MiB @100k). The preserved bug is now scoped to what actually breaks — a bare single-segment root-level key colliding with an ancestor directory name — rather than every Gradle repo.

## Baseline

Re-derived: the pre-index implementation and the index both print `ebf1790bf1d42dad483a51f2cbdeb2351e493b9e8236e4eedeef592dd81e2c5c` over the new 20106-case corpus, 13256 non-null. So the freeze, the lazy `stripped` and the extensions swap are output-identical too.

## Follow-ups

Ruby was missing from my list — added. I will open issues for the follow-ups you listed: Go, C# (no-csproj), Dart, Ruby, and the narrow Kotlin first-occurrence one. Happy to put the Kotlin issue number next to the preserved-rule comment once it exists, if you want it there.

## CI

`benchmarks (GITNEXUS_BENCH)` passed (8m43s), so the new gate runs green on your infrastructure and not just mine. The windows shard that failed before is green now.

Two other shards are red. Neither touches anything this PR changes, but both look worth knowing about:

**`ubuntu / coverage 2/3`** — 1 failed / 273 passed: `doctor-format.test.ts > reports local embeddings as supported on darwin/arm64, linux/x64, and win32/x64`, `expected '✗ optional embedding stack not instal…' to be '✓ local embeddings supported'`.

That case is the only one in its group that does not pass a `resolution`, so it falls through to the real `resolveEmbeddingRuntime()` (`src/cli/doctor.ts:105`) and depends on whether the optional embedding stack is present in the runner's `node_modules` — i.e. exactly the pruned-optional-dependency case #2370 documents a few lines above. Its siblings at lines 67, 78 and 90 all inject `resolution` and are deterministic. Passes locally for me, where the stack is installed. Looks like a one-line fix (inject a `resolution` there too), but it is unrelated to this PR, so I have left it alone — happy to send it separately if useful.

**`windows-latest (platform-sensitive) 2/3`** — 1 failed / 29 passed: `cli-limit-e2e.test.ts`, `Analyze failed (status null)`. A killed subprocess rather than a failed assertion, same family as the `cli-e2e` timeout on the previous run. Also fails locally for me.

koriyoshi2041 (approved):
Re-checked the post-review commit `d2195d44`. The new file-set-bound fingerprint, depth/package-size corpus, `non_null` assertion, depth ratio, small-arm ceiling, frozen buckets, and adapter result assertions address the blind axes raised on the prior head without changing resolver output.

Local validation on Node 26.5.0:

- focused Kotlin parity/reuse suites: 23/23 passed
- `node --import tsx bench/kotlin-import-target/measure.mjs --check`: passed (`fingerprint ebf1790…`, 20,106 cases, 13,256 non-null, scaling 1.061, depth 1.395)
- `git diff --check`: passed

The remaining red hosted shards reported by the author are outside this Kotlin delta; I did not treat them as evidence against the resolver change.

PR #2886, opened 2026-08-07
feat(java): resolve SpringContextUtil.getBeans(X.class) dynamic lookups

ChunxueLi:
## Problem

Java frameworks commonly provide static accessors that retrieve all Spring beans of a given type at runtime:

```java
// Winning Health (WiNEX)
Map<String, IClinicalOrderConverter> beans = SpringContextUtil.getBeans(IClinicalOrderConverter.class);

// Spring native
Map<String, Plugin> plugins = applicationContext.getBeansOfType(Plugin.class);
```

These are the **programmatic equivalent** of `@Autowired List<X>`: both express "inject all implementations of interface X". But while `@Autowired List<X>` is already resolved by GitNexus's Spring DI resolver (PR #2200), `getBeans(X.class)` calls resolve to nothing — the receiver (`SpringContextUtil`) is a static utility in a binary JAR, invisible to the static call graph.

**Impact**: In the WiNEX medical platform alone, **60+ files** use this pattern across **40+ distinct interfaces** — representing hundreds of invisible dependency paths that impact/trace queries cannot follow.

## Solution

A post-resolution heuristic that detects `*.getBeans(Type.class)` patterns and emits INJECTS edges to all implementers:

1. **Scans** Function/Method source text for the regex pattern `(\w+)\.(getBeans|getBean|getBeansOfType)\s*\(\s*(\w+)\.class\s*\)`
2. **Filters** by known receiver names (`SpringContextUtil`, `applicationContext`, `ctx`, etc.) to minimize false positives
3. **Resolves** `Type` to an Interface/Class node via the graph
4. **Walks** IMPLEMENTS edges to find all implementers
5. **Emits** INJECTS edges: `caller Function → each implementer Class`

### Confidence levels
| Method | Cardinality | Confidence |
|--------|------------|------------|
| `getBeans(X.class)` | Collection | 0.8 |
| `getBeansOfType(X.class)` | Collection | 0.8 |
| `getBean(X.class)` | Single | 0.6 |

### Safety features
- ✅ **Receiver filtering**: Only known Spring context names trigger detection
- ✅ **`.class` argument requirement**: Ignores `getBean(String)` overloads
- ✅ **Ambiguous type names skipped**: If multiple interfaces share a simple name, no edges are emitted (fail closed)
- ✅ **Self-injection excluded**
- ✅ **Deduplication**: Edge IDs encode source+target, duplicates collapse automatically

### Why regex (not AST)?
This runs in `emitPostResolutionEdges`, which operates on graph nodes — not raw AST nodes. The graph stores `filePath`, `startLine`, `endLine` for each Function/Method, so we extract the source range and scan it. This is the same approach used by other post-resolution hooks.

## Files
| File | Type | Lines |
|------|------|-------|
| `languages/java/spring-dynamic-lookup.ts` | NEW | ~170 |
| `languages/java/scope-resolver.ts` | MOD (+2) | Import + hook call |
| `test/unit/spring-dynamic-lookup.test.ts` | NEW | ~160 |

## Tests
| Category | Tests | Status |
|----------|-------|--------|
| getBeans collection detection | 4 | ✅ |
| getBean single detection | 1 | ✅ |
| getBeansOfType variant | 1 | ✅ |
| Receiver filtering | 2 | ✅ |
| Negative cases (String arg, no arg, unrelated methods) | 3 | ✅ |
| Multi-lookup in one function | 1 | ✅ |
| Stream-style usage | 1 | ✅ |
| **Existing Java/Spring tests** | **73** | **✅ No regression** |

---

## Summary

*A broad, central ingestion change that extends Spring dynamic lookup handling across Java and Kotlin. Its critical posture comes from touching dependency injection and capture pipelines with reach beyond the directly changed code.*

**🔴 CRITICAL blast radius. A Java and Kotlin Spring ingestion change for resolving `SpringContextUtil.getBeans(X.class)` dynamic lookups, reaching 17 graph dependents.**

The change is concentrated in the `Kotlin` and `Java` modules, especially the capture and scope resolution paths under `gitnexus/src/core/ingestion/languages/`. Review `clearJavaClassAnnotationFacts`, `setJavaSpringAopFacts`, `setJavaSpringNonHttpHandlerFacts`, `collectJavaCaptureSideChannel`, `applyJavaCaptureSideChannel`, and the corresponding Kotlin functions alongside `diPhase` and `execute` in `gitnexus/src/core/ingestion/pipeline-phases/di.ts`.

The affected surface also includes `gitnexus/src/core/ingestion/frameworks/spring/dynamic-lookups.ts`, both language-specific `spring-dynamic-lookup.ts` files, parse cache handling in `gitnexus/src/storage/parse-cache.ts`, and the integration and unit tests for dynamic lookup and DI behavior. Review `.github/workflows/ci-tests.yml` first among the high risk files, then trace the dependency injection and capture phase changes through their callers.

_Added by GitNexus for PR #2886. Edit freely — this block is replaced on the next review, everything above it is left untouched._


ChunxueLi:
Gentle ping — is this still under consideration? Happy to rebase onto current main and address any feedback if there is interest in moving forward.

gitnexus/src/core/ingestion/languages/java/spring-dynamic-lookup.ts
```diff
@@ -0,0 +1,164 @@
...
+        const existing = typeIndex.get(name);
+        if (existing) existing.push(node.id);
+        else typeIndex.set(name, [node.id]);
+      }
+    }
+  });
+
+  // Build Interface → implementers map from IMPLEMENTS edges
+  const implementersIndex = new Map<string, Set<string>>();
+  for (const rel of graph.iterRelationshipsByType('IMPLEMENTS')) {
```
magyargergo: **P1** — Direct-only `IMPLEMENTS` index; no assignability closure. `[code-read]`

`implementersIndex` records one-hop `IMPLEMENTS` sources only. Java interface `Child extends Parent` is itself `IMPLEMENTS` (`java.test.ts:167-174`), so `getBeans(Parent.class)` emits `INJECTS` to the **Child Interface node** and never to `class Impl implements Child`. `getBeansOfType(Base.class)` / `getBean(Concrete.class)` emit nothing because subclasses are `EXTENDS`.

**Trigger:** `Parent` ← `Child` ← `Impl`; lookup `Parent.class` → edge to `Child`, omit `Impl`.

**Fix:** Walk `IMPLEMENTS` + `EXTENDS` transitively; emit concrete classes only. Corroborated by Claude + GPT-5.6-sol; same theme as gitnexus-check #1 (wording about EXTENDS was wrong; miss is real).

gitnexus/src/core/ingestion/languages/java/spring-dynamic-lookup.ts
```diff
@@ -0,0 +1,164 @@
...
+    const endLine = node.properties.endLine as number | undefined;
+    if (!filePath || startLine === undefined || endLine === undefined) return;
+
+    const source = sourceByFile.get(filePath);
+    if (!source) return;
+
+    // Extract function source by line range
+    const lines = source.split('\n');
+    const funcSource = lines
+      .slice(Math.max(0, startLine - 1), Math.min(lines.length, endLine))
```
magyargergo: **P1** — Slice treats 0-based inclusive `GraphNode` lines as 1-based exclusive. `[code-read]`

`line-base.ts` documents `startLine`/`endLine` as 0-based inclusive. This does `slice(startLine - 1, endLine)` = `[s-1, e)` instead of `[s, e+1)`.

**Trigger:** one-line method at row `N` scans the *previous* line; two adjacent one-line methods make `second` inherit `first`'s `INJECTS` edges.

**Fix:** `lines.slice(startLine, endLine + 1)` plus tests for one-line / adjacent one-line methods. Claude adversarial + GPT-5.6-sol adversarial.

gitnexus/test/unit/spring-dynamic-lookup.test.ts
```diff
@@ -0,0 +1,164 @@
...
+ * `attachJavaSpringDynamicLookup` function with an in-memory graph.
+ */
+import { describe, it, expect } from 'vitest';
+
+// We test the regex extraction logic directly since building a full
+// KnowledgeGraph mock is heavy. The integration is verified by the
+// end-to-end analyze tests.
+
+// Re-implement the extraction logic for unit testing (it's a pure function)
+function extractDynamicLookups(sourceText: string, callerNodeId: string) {
```
magyargergo: **P1** — Tests copy `extractDynamicLookups`; they never run production. `[code-read]` `[reproduced]`

This local function duplicates the private regex and receiver sets. Nothing in `gitnexus/test/` imports `attachJavaSpringDynamicLookup`. Focused vitest is 13/13 green even if the attacher is a no-op.

**Trigger:** change production regex or drop the `scope-resolver.ts` hook; this file still passes.

**Fix:** import/export the real extractor; add a `createKnowledgeGraph` test that asserts `INJECTS` edges (and omissions). gitnexus-check #2 + Claude + GPT-5.6-sol.

magyargergo (commented):
# Tri-review digest — PR #2886 `feat(java): resolve SpringContextUtil.getBeans(X.class) dynamic lookups`

**Pinned:** BASE `031e1237` → HEAD `0f4f70c1` · 3 files, +330/-0, 1 commit · fork PR
**Classifications:** merge state `checks failing` · branch hygiene `clean feature/fix PR` · GitHub reports MERGEABLE (no conflicts) and BEHIND main — merge is blocked by red required checks, not by conflicts.
**Method:** two independent engines, each running a graph-backed swarm review and a compound-engineering multi-lens review, cross-checked against each other and re-verified by hand in a pinned worktree.

**Verdict: not production-ready.** The feature is well-placed and idiomatic, but three defects in the emission path mean the graph it produces is both incomplete and — in specific, reachable cases — wrong about *which* caller depends on a bean. The shipped test file cannot catch any of them, because it never executes the shipped code.

---

## Engine breakdown

| Engine | Lane | Status |
|---|---|---|
| **A — Claude** | GitNexus swarm (facts, risk, security, test-CI, docs/DoD, branch-hygiene) | 6/6 complete |
| **A — Claude** | compound-engineering (correctness, adversarial, testing, maintainability, project-standards) | 5/5 complete |
| **B — GPT-5.6-sol (xhigh)** | GitNexus swarm, Solo mode | **incomplete** — see note |
| **B — GPT-5.6-sol (xhigh)** | compound-engineering multi-lens | complete, with final report |
| **B — GPT-5.6-sol (xhigh)** | adversarial red-team | complete, with final report |

**Engine B swarm liveness (stated plainly):** the Solo swarm lane did **not** finish. Its artifact is a command log; the lane was still inside `npm test` when synthesis began and never emitted a consolidated per-persona swarm report. Its interim assistant messages did confirm the two bot findings from source and the CI attribution, and its focused checks (13/13 focused vitest, `tsc --noEmit`, tsx graph probes) completed. Cross-engine corroboration below is therefore drawn from Engine B's **two completed lanes** (multi-lens + adversarial), not from the unfinished swarm lane.

**Validation both engines actually ran:** focused `test/unit/spring-dynamic-lookup.test.ts` → 13/13 pass; `npx tsc --noEmit` → pass; `git diff --check` → clean. Full `npm test` is **inconclusive in this container** on both engines (unrelated LadybugDB file-locking, a missing `parse-worker.js` artifact, and `incremental-orchestration.test.ts` setup failures; one run interrupted). No lane succeeded in executing `attachJavaSpringDynamicLookup` itself — every correctness finding below is a source read, not an observed edge, and is tagged accordingly.

---

## What is solid

Credit where it is due — this is a competently scoped change and several reviewers went looking for problems here and found none:

- **Correct extension point.** `attachJavaSpringDynamicLookup` is invoked from the Java `ScopeResolver`'s `emitPostResolutionEdges` hook (`languages/java/scope-resolver.ts:101`), alongside six sibling `attachJavaSpring*` hooks. No file under `core/ingestion/` shared paths is touched, so the "shared pipeline code must not name languages" rule in AGENTS.md is respected. Naming matches the sibling convention exactly.
- **Right problem, real-world motivation.** The module docstring correctly identifies why the call graph loses these edges (static utility in a binary JAR, receiver unresolvable) and correctly frames `getBeans(X.class)` as the programmatic analogue of `@Autowired List<X>`.
- **`INJECTS` reused, not reinvented.** An existing schema-registered relation type with an existing ID convention (`INJECTS:<src>-><tgt>`), matching `pipeline-phases/di.ts`.
- **Self-injection guarded** (`spring-dynamic-lookup.ts:152`), and non-`.class` bean-name lookups such as `getBean("id")` are correctly not matched — which keeps this heuristic from colliding with the `@Resource`/bean-factory work in #2740.
- **Test placement and PR hygiene are clean.** `test/unit/` flat file mirroring the source module matches TESTING.md and 16 sibling `spring-*.test.ts` files; conventional-commit title; one commit, three coupled files, no drive-by churn, no merge pollution. The Java path has no overlap with `origin/main` since BASE, so the required rebase should be clean.
- **Security lane passes.** ReDoS on the lookup regex: refuted. Cypher/query injection via `typeName`: refuted (`\w+` only). Secrets, bidirectional-Unicode, and non-ASCII hygiene: clean.

---

## Headline findings

**1. The implementer index is one hop deep, so normal Spring hierarchies resolve to the wrong node or to nothing.**
`implementersIndex` is built from direct `IMPLEMENTS` edges only (`spring-dynamic-lookup.ts:106-110`) and emission reads only that direct bucket (`:142`). Two reachable consequences, both ordinary `getBeansOfType` usage. Corroborated by both engines and independently by `gitnexus-check`.

**2. The function-source slice uses the wrong line base, which can attribute a lookup to the wrong caller.**
`GraphNode` `startLine`/`endLine` are 0-based and inclusive — documented as load-bearing in `core/ingestion/utils/line-base.ts:1-20`. The slice at `:133` subtracts one from `startLine` and treats `endLine` as exclusive, shifting the scan window one line earlier than the method body. Corroborated by both engines; verified against `line-base.ts` in the worktree.

**3. The 164-line test file does not execute one line of the 164-line production module.**
It re-declares `extractDynamicLookups` with its own copy of the regex and all three constant sets (`test/unit/spring-dynamic-lookup.test.ts:14-42`). Repository-wide search finds `attachJavaSpringDynamicLookup` in exactly three places: its definition, the scope-resolver import, and the scope-resolver call — **no test**. The only occurrence in the test file is inside a docstring. So findings 1 and 2 are not merely untested; they are *untestable* by this suite, and so is any future regression in the wiring. Corroborated by both engines and by `gitnexus-check`.

---

## Inline candidates

Line numbers are RIGHT-side (both files are new, so diff line = file line).

### P1 — `gitnexus/src/core/ingestion/languages/java/spring-dynamic-lookup.ts:106` — direct-only implementer index; no assignability closure
**Engines:** Claude (risk, correctness, adversarial, testing) + GPT-5.6-sol (multi-lens #3, adversarial high). Also `gitnexus-check` #1. **Tag:** [code-read]

`implementersIndex` records only immediate `IMPLEMENTS` sources (`:106-110`); `:142` reads that bucket and nothing else.

- **Trigger A (wrong node emitted):** `interface Parent {}`, `interface Child extends Parent {}`, `class Impl implements Child {}`, caller does `SpringContextUtil.getBeans(Parent.class)`. In this pipeline Java interface inheritance is itself emitted as `IMPLEMENTS`, so `implementersIndex[Parent]` contains the **`Child` Interface node**, not the concrete bean. The code emits `Caller INJECTS Child` and never emits `Caller INJECTS Impl` — an interface is recorded as an injected dependency, and the real bean is missing.
- **Trigger B (nothing emitted):** `class Concrete extends Base {}`, caller does `getBeansOfType(Base.class)`. Only `Concrete -EXTENDS-> Base` exists, `implementersIndex` has no `Base` key, so no edge at all.
- **Trigger C (nothing emitted):** `class EmailService {}`, caller does `applicationContext.getBean(EmailService.class)`. `typeIndex` resolves the class (`:94` indexes `Class` as well as `Interface`), but a concrete class has no `IMPLEMENTS` entry, so `:143` bails. Every direct concrete-class lookup is silently discarded. `di.ts:304` handles exactly this case by adding the class entry itself.

**Suggested fix:** build a reverse subtype index over **both** `IMPLEMENTS` and `EXTENDS`, walk it transitively from the resolved lookup type, and emit only concrete bean-capable classes — not intermediate `Interface` nodes. `di.ts:163-177` (combined `directSupertypes`) and `di.ts:229-250` (the closure walk) are the pattern to reuse.

### P1 — `gitnexus/src/core/ingestion/languages/java/spring-dynamic-lookup.ts:133` — 0-based inclusive range sliced as 1-based exclusive
**Engines:** Claude (adversarial P1) + GPT-5.6-sol (adversarial). **Tag:** [code-read]

```
.slice(Math.max(0, startLine - 1), Math.min(lines.length, endLine))
```

For a 0-based inclusive span `[s, e]` the correct slice is `[s, e+1)`. This one slices `[s-1, e)`: it prepends the line *before* the declaration and drops the method's last line.

- **Trigger (missed):** a one-line method at row `N` (`void a(){ ctx.getBeans(Service.class); }`) has `startLine === endLine === N`, so the slice is `[N-1, N)` — it scans the *preceding* line and never sees its own body.
- **Trigger (misattributed):** two adjacent one-line methods, `first` at row `N` and `second` at row `N+1`. `second` slices `[N, N+1)` = `first`'s line, so **`second` receives `first`'s `INJECTS` edges.** Impact and trace then report a dependency for a method that has none, and omit it from the method that does.

Multi-line methods usually survive by luck (the dropped last line is the closing brace), which is precisely why this will not be noticed without a test.

**Suggested fix:** `lines.slice(Math.max(0, startLine), Math.min(lines.length, endLine + 1))`, plus production-path tests for a lookup on the method's final line, a one-line method, and two adjacent one-line methods.

### P1 — `gitnexus/test/unit/spring-dynamic-lookup.test.ts:14` — tests exercise a hand-copied extractor, never the production module
**Engines:** Claude (testing, maintainability, risk, test-CI) + GPT-5.6-sol (multi-lens #4, adversarial). Also `gitnexus-check` #2. **Tags:** [code-read] + [reproduced]

`function extractDynamicLookups(...)` is re-declared at `:14-42` with duplicated `KNOWN_RECEIVERS`, `COLLECTION_METHODS`, `SINGLE_METHODS`, and regex. The file's only import is `vitest` (`:7`). Nothing in `gitnexus/test/` imports or calls `attachJavaSpringDynamicLookup`.

- **Trigger [reproduced]:** both engines ran the file green (13/13). Replace the production regex at `:66` with `/never-match/`, or make the attacher a no-op, or delete the `scope-resolver.ts:101` call — all 13 tests still pass. The suite is structurally incapable of failing on a production regression.
- **Also false in-file:** the docstring at `:4-5` claims the file tests "the end-to-end `attachJavaSpringDynamicLookup` function with an in-memory graph", and `:9-11` claims "the integration is verified by the end-to-end analyze tests." Neither is true — there is no test anywhere referencing `SpringContextUtil`, `getBeansOfType`, or the attacher. These comments will actively mislead the next maintainer into thinking the wiring is safety-netted.

**Suggested fix:** two steps, both small. (a) Export `extractDynamicLookups` and import it, deleting the copy — this converts the 13 existing cases into real coverage with a one-line change and no graph needed. (b) Add one test that builds a `KnowledgeGraph` via `createKnowledgeGraph`, calls `attachJavaSpringDynamicLookup`, and asserts both emitted and *omitted* `INJECTS` edges — that scaffolding is what makes the P1s above verifiable. Then correct or remove the two false comments.

### P2 — `gitnexus/src/core/ingestion/languages/java/spring-dynamic-lookup.ts:140` — repo-wide simple-name index discards resolvable lookups
**Engines:** Claude (correctness, adversarial) + GPT-5.6-sol (multi-lens #2, adversarial high). **Tag:** [code-read]

`typeIndex` (`:92-102`) keys every `Class` and `Interface` in the graph by simple name, with no package, import, or language scoping; `:140` drops any collision. The graph handed to this hook is repository-wide (`scope-resolution/pipeline/phase.ts:471`, `graph: ctx.graph`), so the collision set spans all languages.

- **Trigger (same-language):** `com.orders.Handler` and `com.audit.Handler` both exist; the caller imports the former and calls `getBeans(Handler.class)`. `typeIds.length === 2`, so no edge — despite an unambiguous Java import. `Service`, `Handler`, `Factory`, and `Configuration` collide routinely in large Java repos.
- **Trigger (cross-language):** a TypeScript `interface PaymentService` is enough to suppress an otherwise unambiguous Java `getBeans(PaymentService.class)`.
- Also, a qualified literal (`com.a.Service.class`) does not match the regex at all.

Fail-closed on ambiguity is a deliberate and defensible house style — `di.ts:292-300` fails closed for the same reason with an explicit comment that import-aware disambiguation is unavailable "in this graph-only phase." Two things still distinguish this instance: `parsedFiles` is already threaded into this function (`:88`) and carries exactly the package/import bindings needed to resolve the literal properly, and the sibling `di.ts` at least scopes its indexes by language (`di.ts:190-192`) where this does not.

**Suggested fix:** at minimum, restrict `typeIndex` to Java nodes so a foreign-language name never suppresses a Java lookup. Better: resolve the class literal through the caller file's package/import bindings via the already-available `parsedFiles`, and keep fail-closed only when *that* scoped resolution is genuinely ambiguous.

### P2 (optional) — `gitnexus/src/core/ingestion/languages/java/spring-dynamic-lookup.ts:66` — regex matches comments and string literals
**Engines:** Claude (correctness) + GPT-5.6-sol (multi-lens #1, adversarial). **Tag:** [code-read]

The pattern runs over raw method source with no lexical awareness, so a match inside a `//` comment, a Javadoc block, a string literal, or a text block becomes a `LookupSite` and reaches edge emission unchanged.

- **Trigger:** a method whose body is only `String example = "ctx.getBeans(Service.class)";` emits a real `INJECTS` edge to every implementer of `Service`. A Javadoc example showing how to call the utility — a plausible thing to write — does the same.

The module documents broad matching as intentional ("false positives are low-cost", `:37-40`), and for *receiver* breadth that is a reasonable stance. It is weaker here: a commented-out or documented call is not a low-confidence dependency, it is a dependency that provably does not exist, and it lands in the graph at the same 0.8 confidence as a real one. Flagged as optional because it is a smaller effect than the three P1s and shares their root cause — text scanning instead of AST call captures. Whichever way it is resolved, a negative test should pin the decision.

---

## Lower priority

- **P2 — nested callables double-count one lookup (`:119`).** Every `Method` node scans its full textual range independently, and this pipeline emits anonymous-class methods as their own `Method` nodes while the enclosing method's range still contains them. Verified in-tree: `test/fixtures/lang-resolution/java-anonymous-class-scope/src/Worker.java` has `makeHandler` textually containing the anonymous `run`, and `test/integration/resolvers/java.test.ts:3181-3183` pins `run` as an independently emitted `Method`. A lookup inside `run` matches both ranges and emits from both callers; ID dedup (`core/graph/graph.ts:86-88`) cannot help because the source IDs differ. Claude's adversarial lane rated this P1; single-engine, so listed here — but it is a code-read certainty, not a hypothesis. Binding sites to their nearest enclosing callable via the AST fixes this and the `:133` slice bug together. [code-read]
- **P3 — `parsedFiles` accepted and never used (`:88`, call site `scope-resolver.ts:101`).** Reads as a copy-paste of the sibling `attachJavaSpring*` signature; `noUnusedParameters` is off, so it compiles silently and looks load-bearing to the next reader. Note that the P2 above wants this parameter *used*, not removed — so resolve that first, then decide. [code-read]
- **P3 — per-node re-split of the whole file (`:131`).** `source.split('\n')` runs once per `Function`/`Method` node rather than once per file. Both engines noted it. A `filePath → lines[]` cache next to `sourceByFile` is a one-line fix. [code-read]
- **P3 — unnamed confidence literals (`:146`).** `site.isCollection ? 0.8 : 0.6` — naming these next to the existing constants block would make them greppable and tunable. [code-read]
- **P3 — `Constructor` nodes are skipped (`:120`).** Only `Function` and `Method` labels are scanned, so a lookup performed in a constructor is invisible. Probably intentional; worth a one-line comment either way. [code-read]
- **P3 — `ARCHITECTURE.md` still documents `di.ts` as the sole `INJECTS` source.** `ARCHITECTURE.md:111` describes `INJECTS` as coming only from the framework-neutral `di` phase via `di-extractors/`, with consumer-Class or factory-Method sources. This PR adds a second emitter with a different source shape (`Function`/`Method` → implementer `Class`) that bypasses `di-extractors/` entirely, and `CONTRIBUTING.md` asks for doc updates when behavior changes. A footnote on the `di` row would close it. [code-read]

## Design trade-offs for maintainer sign-off (not defects)

Both of these are explicitly documented in the module and are therefore product decisions, not bugs. Neither is being raised as a P1.

- **`getBean` (singular) fans out to every implementer.** `:151` loops all implementers for single-bean lookups exactly as it does for collections, at 0.6 confidence, and `:30-33` says so out loud. Runtime returns one bean (or throws, when several non-`@Primary` candidates exist). The graph will therefore show dependencies on paths that cannot execute. Defensible for a recall-oriented heuristic; worth confirming that `impact` and `trace` consumers are comfortable with 0.6-confidence edges being over-broad rather than absent, given Spring bean-candidate metadata (`@Primary`/qualifier) is already available in this codebase.
- **`ctx` / `context` / `beanFactory` accepted as receivers by name alone.** `:41-52` is deliberately broad, with no receiver type, import, or Spring provenance check — and `:37-40` documents the trade-off. Consequence: any unrelated API exposing `getBean(Class)` under a variable named `ctx` produces a false `INJECTS` edge (e.g. a `PluginRegistry ctx` with `getBean(Plugin.class)`). Claude's adversarial lane wanted this at P1; it is reported here as a trade-off because the file states the intent. Sign-off question for the maintainer: is name-only receiver matching acceptable at 0.8 confidence on collection lookups, or should generic locals be dropped and only the distinctive static utility names kept?

---

## Refuted

Dropped after verification — please don't re-litigate these:

- **`gitnexus-check` #1's `EXTENDS` wording is wrong, but its conclusion is right.** The bot says a class implementing `Child` is missed "because Java models interface inheritance as `EXTENDS`" and the index only reads `IMPLEMENTS`. In this pipeline interface-to-interface inheritance is emitted as **`IMPLEMENTS`**, not `EXTENDS` — pinned by `test/integration/resolvers/java.test.ts:167-174` (`emits IMPLEMENTS IA → IB`) and `:172-174` (`emits no EXTENDS edges`). The failure mode is therefore not "the edge type isn't read" but "the walk isn't transitive": `implementersIndex[Parent]` returns the `Child` *Interface*. **The miss is real and confirmed** (P1 above); only the mechanism in the bot's explanation needed correcting. Concrete-class inheritance *is* `EXTENDS` and *is* separately unread, so the bot's remedy happens to be right for a different reason.
- **`gitnexus-check` #1's "di.ts already does this, so this PR lacks parity" is partly refuted.** `di.ts`'s `IMPLEMENTS`+`EXTENDS` union (`:163-177`) feeds a *provided-type assignability* closure (`:229-250`); its structural interface fan-out at `:303-307` is direct-only, same as this PR. So the parity framing is inaccurate. What `di.ts` genuinely does and this PR does not: language-scope its indexes (`:190-192`) and handle the direct concrete-class target (`:304`).
- **Cross-language *false* edges: refuted.** `ctx.fileContents` is built per-language pass (`scope-resolution/pipeline/run.ts:556-562`, from that pass's `files`), so a non-Java `Function`/`Method` node finds no source and returns early at `:128`. Only the fail-closed *miss* from the shared `typeIndex` is real (P2 above).
- **`fileContents` key vs `node.properties.filePath` mismatch: refuted.** `run.ts` keys by `f.path` and the parse worker stamps `filePath` from the same value; sibling post-resolution hooks key the same way.
- **ReDoS on the lookup regex: refuted.** **Query/Cypher injection via `typeName`: refuted** (`\w+`). **Secrets, bidi-Unicode, non-ASCII hygiene: clean.** Em dashes in comments are house style.
- **The red CI Gate is not caused by this PR: refuted as a code defect** (see CI below).
- **Branch hygiene: nothing to fix.** No merge-from-main commits, no unrelated churn, no mixed domains.

---

## Test gaps

Everything below is currently unverifiable by the suite, because no test executes the attacher. Fixing inline candidate #3 is the prerequisite for all of it.

1. No test constructs a `KnowledgeGraph` or calls `attachJavaSpringDynamicLookup`; no integration fixture references `SpringContextUtil` or `getBeansOfType`.
2. **Line-range behavior** — lookup on a method's final line; a one-line method; two adjacent one-line methods (would have caught `:133`).
3. **Inheritance** — `Parent ← Child ← Impl` transitive interface chain; `Concrete extends Base`; direct `getBean(Concrete.class)` (would have caught `:106`/`:142`).
4. **Resolution branches** — the ambiguous-type skip (`typeIds.length !== 1`), the no-implementers skip, and the self-injection skip are each asserted nowhere.
5. **Edge shape** — no assertion on the emitted `id` format, `type`, `confidence` (0.8 vs 0.6), or `reason` string that downstream `impact`/`trace`/MCP consumers read.
6. **Negative cases** — comment and string-literal occurrences; a non-Spring receiver named `ctx`; a bean-name `getBean("id")` call (correctly unmatched today, but unpinned).
7. **Cardinality semantics** — the documented "single lookup still fans out to all implementers" behavior is stated in a comment and proven nowhere.
8. **Node labels** — only `Function`-shaped snippets are exercised; the `Method` branch and the `Constructor` exclusion are untested.

---

## CI status

- **CI Gate is red, and it is not this PR's fault.** The failure is `tests / windows-latest (platform-sensitive) 1/3` (run 32919148953): `test/integration/cli-e2e.test.ts:372` — "analyze timed out before asserting finalization artifacts — this test guards #1169". That file is untouched by this diff, the guard references a different pre-existing issue, and Ubuntu coverage passed in full (18700/18700 unit, typecheck, lint, quality). Both engines checked the job logs and agree. Treat as a known Windows analyze-timeout flake.
- **Vercel "unauthorized"** is fork-PR deployment auth, not a code signal. The web E2E gate correctly skipped (no `gitnexus-web` changes).
- **The branch is BEHIND main** and needs a rebase before merge; the Java path has no conflicting commits since BASE, so it should be clean.
- The gate is still formally red, so merge is blocked on it regardless of the findings above.

---

## Recommended fix order

1. **Make the tests real** — export and import the extractor, then add a graph-backed test for the attacher. Nothing else below can be verified until this exists.
2. **Fix the line base** (`:133`) — smallest change, highest correctness-per-line, and it stops edges being attributed to the wrong caller.
3. **Close the assignability walk** (`:106`) over `IMPLEMENTS` + `EXTENDS`, emitting concrete classes only and handling direct class targets.
4. **Scope type resolution** (`:140`) — at minimum language-filter `typeIndex`; ideally resolve the literal through `parsedFiles` imports.
5. **Decide on AST call captures** — this would retire the comment/string false positives (`:66`), the nested-callable double-count (`:119`), and the line-slicing class of bug in one move. Reasonable to defer to a follow-up if 1-4 land with tests.
6. Rebase onto main; confirm the Windows shard is green or reconfirm it as the #1169 flake.

Once 1-4 are in with production-path tests, this is a genuinely useful addition to Java Spring resolution.

---

*Automated tri-review digest: two engines (Claude, GPT-5.6-sol xhigh), five completed review lanes plus one incomplete lane, cross-engine corroboration weighted over raw finding count, every posted item re-verified against the pinned worktree at `0f4f70c1`. Read-only — nothing was edited, committed, or pushed. Line numbers are RIGHT-side diff positions. `[code-read]` = verified by reading source at the pinned SHA; `[reproduced]` = observed by executing tests.*

magyargergo:
@ChunxueLi could you please address these review findings? 🙏

gitnexus/src/core/ingestion/languages/java/spring-dynamic-lookup.ts
```diff
@@ -0,0 +1,164 @@
...
+      const name = node.properties.name as string | undefined;
+      if (name) {
+        const existing = typeIndex.get(name);
+        if (existing) existing.push(node.id);
+        else typeIndex.set(name, [node.id]);
+      }
+    }
+  });
+
+  // Build Interface → implementers map from IMPLEMENTS edges
```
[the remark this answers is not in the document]
magyargergo: Fixed in 22dc55c98. Dynamic lookups now feed the shared DI resolver, whose reverse IMPLEMENTS+EXTENDS traversal is transitive, cycle-safe, deterministic, and emits concrete Class/Record/Enum targets only. Unit and full-pipeline Java/Kotlin regressions cover the hierarchy cases.

gitnexus/test/unit/spring-dynamic-lookup.test.ts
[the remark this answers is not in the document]
magyargergo: Fixed in 22dc55c98. The copied extractor was removed. Unit tests now invoke the production Java/Kotlin AST capture, and integration tests run the worker side-channel, scope resolver, DI phase, and assert emitted INJECTS source/target/type/confidence/reason and omissions.

gitnexus/src/core/ingestion/languages/java/spring-dynamic-lookup.ts
magyargergo: Fixed in 22dc55c98. Assignability now walks reverse IMPLEMENTS and EXTENDS transitively with cycle protection, includes direct concrete targets, and excludes intermediate interfaces. Regression coverage includes interface chains, class inheritance, direct concrete lookup, and Record/Enum implementers.

gitnexus/src/core/ingestion/languages/java/spring-dynamic-lookup.ts
magyargergo: Fixed in 22dc55c98 by removing source-line slicing entirely. Lookups are captured from AST call nodes and attributed to the nearest enclosing callable. Tests pin one-line/adjacent methods, final-line lookups, constructors, and nested anonymous callables.

gitnexus/test/unit/spring-dynamic-lookup.test.ts
magyargergo: Fixed in 22dc55c98. Tests execute production capture and full graph construction; the integration fixture proves resolver wiring and validates INJECTS edge shape, confidence, reasons, and expected omissions for Java and Kotlin.

magyargergo:
> @ChunxueLi could you please address these review findings?

The tri-review digest (implementer walk one hop deep, line-range mis-slice, tests that never executed the attacher, repo-wide simple-name index, comment/string false positives, nested-callable double-count) is addressed on the current head (`82fb596`):

- Lookups are AST-captured on real `method_invocation` / `call_expression` nodes (comments and strings are not calls).
- Sites bind to the nearest enclosing callable (Java constructors included; Kotlin secondary constructors included, `init` excluded on purpose).
- Assignability walks transitive `IMPLEMENTS` + `EXTENDS` and emits concrete types; type names resolve through file-scope bindings and fail closed when that is ambiguous.
- Production-path unit and pipeline tests assert `INJECTS` edges; capture is folded into the existing scope-query pass so growth stays linear.

No further code change from that digest. Remaining red checks on this head are CI (coverage shard / CI Gate) and Vercel fork authorization, which I am treating as the next stream rather than as review work.

ChunxueLi:
Thanks for the thorough review — the P1 is exactly right and now fixed (62174bdb, rebased onto current main first — the old base no longer compiled, which is why CI had been stalled).

**P1 (tests copy `extractDynamicLookups`):** `extractDynamicLookups` and `LookupSite` are now exported from `spring-dynamic-lookup.ts`, and the suite imports the production function instead of the local re-implementation. Verified mutation-style: corrupting the production regex now fails 9 tests (previously 0).

Also added the requested integration coverage: `attachJavaSpringDynamicLookup` runs against a real `createKnowledgeGraph` fixture (interface + two implementers + caller method + IMPLEMENTS edges), asserting the emitted INJECTS edges fan out to every implementer — plus the three omission paths (unknown receiver, no implementers, missing source).

magyargergo:
> Thanks for the thorough review — the P1 is exactly right and now fixed ([62174bd](https://github.com/abhigyanpatwari/GitNexus/commit/62174bdb602c8a36df3f365fb12be19d430c05d8), rebased onto current main first — the old base no longer compiled, which is why CI had been stalled).
> 
> **P1 (tests copy `extractDynamicLookups`):** `extractDynamicLookups` and `LookupSite` are now exported from `spring-dynamic-lookup.ts`, and the suite imports the production function instead of the local re-implementation. Verified mutation-style: corrupting the production regex now fails 9 tests (previously 0).
> 
> Also added the requested integration coverage: `attachJavaSpringDynamicLookup` runs against a real `createKnowledgeGraph` fixture (interface + two implementers + caller method + IMPLEMENTS edges), asserting the emitted INJECTS edges fan out to every implementer — plus the three omission paths (unknown receiver, no implementers, missing source).

It's a bit late for this, I've already address them. Thanks though. Next time, please look at the state of the PR to avoid confusion.

PR #2890, opened 2026-08-08
feat(java): resolve @FeignClient method calls as CALLS edges

ChunxueLi:
## Summary

`@FeignClient` interfaces have no source-level implementation — Spring Cloud creates a dynamic proxy at runtime. The scope resolver cannot resolve the receiver to any concrete symbol, so calls on injected Feign fields produce zero CALLS edges, completely breaking impact analysis across RPC boundaries.

## Approach

New module `spring-feign-rpc.ts` runs in `emitPostResolutionEdges`:

1. **Identify Feign interfaces**: Scans source text for `@FeignClient` annotations.
2. **Build method index**: Maps Feign interface name → method node IDs from the graph.
3. **Detect field declarations**: Finds `@Autowired/@Resource/private` fields of Feign types.
4. **Scan for calls**: For each Function/Method node, regex-matches `fieldName.methodName()` patterns.
5. **Emit edges**: Creates `CALLS` edges (confidence 0.85) from caller to Feign interface method.

## Test coverage

12 unit tests: single/multiple methods, multiple Feign interfaces in same class, `@Autowired`/`@Resource`/no-annotation injection, duplicate suppression, self-reference filtering, uncalled fields, multi-function files, value-quoted service name.

### DIFF
--- a/gitnexus/src/core/ingestion/languages/java/scope-resolver.ts
+++ b/gitnexus/src/core/ingestion/languages/java/scope-resolver.ts
@@ -29,11 +29,32 @@ import {
   type JavaResolveContext,
 } from './index.js';
 import { populateJavaPackageSiblings } from './package-siblings.js';
+import { attachSpringBeanCandidateMetadata } from './spring-bean-metadata.js';
+import { attachJavaSpringAopMetadata } from './spring-aop.js';
+import { attachJavaSpringConfigBindings } from './spring-config-bindings.js';
+import { attachJavaSpringConditionalMetadata } from './spring-conditionals.js';
+import { attachJavaSpringDiMetadata } from './spring-di.js';
+import { attachJavaSpringDynamicLookup } from './spring-dynamic-lookup.js';
+import { attachJavaFeignRpcCalls } from './spring-feign-rpc.js';
+import {
+  applyJavaCaptureSideChannel,
+  clearJavaClassAnnotationFacts,
+} from './capture-side-channel.js';
+import { clearJavaPackageFacts } from './package-facts.js';
 
 const javaScopeResolver: ScopeResolver = {
   language: SupportedLanguages.Java,
   languageProvider: javaProvider,
   importEdgeReason: 'java-scope: import',
+
+  loadResolutionConfig: () => {
+    // Worker capture facts are process-local and outlive a single analysis in
+    // server mode. This hook runs once before each Java workspace pass, before
+    // ParsedFile side channels are restored for the current files.
+    clearJavaClassAnnotationFacts();
+    clearJavaPackageFacts();
+    return undefined;
+  },
 
   resolveImportTarget: (targetRaw, fromFile, allFilePaths) => {
     const ws: JavaResolveContext = { fromFile, allFilePaths };
@@ -50,6 +71,7 @@ const javaScopeResolver: ScopeResolver = {
   buildMro: buildJavaMro,
 
   populateOwners: (parsed: ParsedFile) => populateClassOwnedMembers(parsed),
+  applyCaptureSideChannel: applyJavaCaptureSideChannel,
 
   isSuperReceiver: (text) => text.trim() === 'super',
 
@@ -57,9 +79,25 @@ const javaScopeResolver: ScopeResolver = {
   propagatesReturnTypesAcrossImports: true,
   collapseMemberCallsByCallerTarget: true,
   hoistTypeBindingsToModule: true,
+  stripReceiverCastExpressions: true,
+  // #2550: every Java method belongs to a class instance — a free call may
+  // resolve to a Method only when the caller's enclosing class chain
+  // (self + MRO) contains the method's owner. Closes the finalize-bucket
+  // leak (unqualified `run()` matching an unrelated same-file anonymous
+  // class's method). C# is the intended next adopter.
+  freeCallsRequireInstanceOwnership: true,
 
   populateNamespaceSiblings: populateJavaPackageSiblings,
   populateRangeBindings: populateJavaCrossFileReturnTypes,
+  emitPostResolutionEdges: (graph, parsedFiles, nodeLookup, indexes, ctx) => {
+    attachSpringBeanCandidateMetadata(graph, parsedFiles, nodeLookup, indexes);
+    attachJavaSpringAopMetadata(graph, parsedFiles, nodeLookup, indexes);
+    attachJavaSpringConditionalMetadata(graph, parsedFiles, nodeLookup, indexes);
+    attachJavaSpringDiMetadata(graph, parsedFiles, nodeLookup, indexes);
+    attachJavaSpringConfigBindings(graph, parsedFiles, nodeLookup, indexes, ctx);
+    attachJavaSpringDynamicLookup(graph, parsedFiles, ctx.fileContents);
+    attachJavaFeignRpcCalls(graph, parsedFiles, ctx.fileContents);
+  },
 };
 
 export { javaScopeResolver };
--- a/gitnexus/src/core/ingestion/languages/java/spring-feign-rpc.ts
+++ b/gitnexus/src/core/ingestion/languages/java/spring-feign-rpc.ts
@@ -1,0 +1,189 @@
+/**
+ * Spring Cloud OpenFeign RPC heuristic: resolve method calls on
+ * @FeignClient-injected fields.
+ *
+ * `@FeignClient` interfaces have no implementation class in the source
+ * codebase — Spring Cloud creates a dynamic proxy at runtime. The call
+ * graph sees `feignClient.doSomething()` but the scope resolver cannot
+ * resolve the receiver to any concrete implementation, so the CALLS edge
+ * is missing entirely. This means impact analysis from the caller through
+ * the RPC interface is completely broken.
+ *
+ * This module:
+ * 1. Scans source text for `@FeignClient` annotations → identifies Feign
+ *    interface names.
+ * 2. Builds Feign interface name → method-node index from the graph.
+ * 3. Scans each Class with an injected Feign-type field for
+ *    `fieldName.methodName(...)` call patterns.
+ * 4. Emits CALLS edges from the calling method to the Feign interface method.
+ *
+ * Runs in `emitPostResolutionEdges` (after the full graph is built).
+ */
+
+import type { KnowledgeGraph } from '../../../graph/types.js';
+import type { ParsedFile } from 'gitnexus-shared';
+
+// ── Constants ─────────────────────────────────────────────────────────────
+
+/** Regex to detect @FeignClient annotation on an interface declaration. */
+const FEIGN_CLIENT_RE = /@FeignClient\s*\(/;
+
+/**
+ * Regex to extract the interface name from a source file containing
+ * `@FeignClient`. Matches `public interface XxxRpcService {`.
+ */
+const INTERFACE_NAME_RE = /interface\s+(\w+)/;
+
+/** Regex to detect field declarations of a known Feign type. */
+function makeFieldPattern(typeName: string): RegExp {
+  // Matches: private TypeName fieldName;  |  @Autowired TypeName fieldName;
+  // 'g' flag required for exec() loop to advance.
+  return new RegExp(
+    `(?:@(?:Autowired|Resource)\\s+)?(?:private|protected|public)?\\s*${typeName}\\s+(\\w+)\\s*[;=]`,
+    'g',
+  );
+}
+
+/**
+ * Regex to detect method calls on a field: `fieldName.methodName(`.
+ * Captures methodName for resolution.
+ */
+function makeCallPattern(fieldName: string): RegExp {
+  return new RegExp(`\\b${fieldName}\\.(\\w+)\\s*\\(`, 'g');
+}
+
+// ── Types ────────────────────────────────────────────────────────────────
+
+interface FeignMethodInfo {
+  /** Graph node IDs of methods declared on the Feign interface. */
+  methodNodeIds: Map<string, string>; // methodName → nodeId
+}
+
+// ── Main API ──────────────────────────────────────────────────────────────
+
+/**
+ * Attach CALLS edges for @FeignClient interface method invocations.
+ * Called from `emitPostResolutionEdges` in the Java scope resolver.
+ */
+export function attachJavaFeignRpcCalls(
+  graph: KnowledgeGraph,
+  parsedFiles: readonly ParsedFile[],
+  fileContents: ReadonlyMap<string, string>,
+): void {
+  // ── Step 1: Identify @FeignClient interface names from source ──
+  const feignTypeNames = new Set<string>();
+  for (const [, content] of fileContents) {
+    if (!FEIGN_CLIENT_RE.test(content)) continue;
+    const m = INTERFACE_NAME_RE.exec(content);
+    if (m) feignTypeNames.add(m[1]);
+  }
+  if (feignTypeNames.size === 0) return;
+
+  // ── Step 2: Build Feign interface → method-node index ──
+  // Map: interfaceName → (methodName → nodeId[])
+  const feignMethods = new Map<string, Map<string, string[]>>();
+  graph.forEachNode((node) => {
+    if (node.label !== 'Method' && node.label !== 'Function') return;
+    const filePath = node.properties.filePath as string | undefined;
+    if (!filePath) return;
+    // Check if this method belongs to a Feign interface by matching
+    // the interface name in the file path or node ancestry
+    const fileName =
+      filePath
+        .split('/')
+        .pop()
+        ?.replace(/\.java$/, '') ?? '';
+    if (!feignTypeNames.has(fileName)) return;
+    const methodName = node.properties.name as string | undefined;
+    if (!methodName) return;
+    let methodMap = feignMethods.get(fileName);
+    if (!methodMap) {
+      methodMap = new Map<string, string[]>();
+      feignMethods.set(fileName, methodMap);
+    }
+    const ids = methodMap.get(methodName);
+    if (ids) ids.push(node.id);
+    else methodMap.set(methodName, [node.id]);
+  });
+
+  if (feignMethods.size === 0) return;
+
+  // ── Step 3: Build caller-side index ──
+  // For each file, find Feign-type field declarations and their field names.
+  // Map: filePath → (typeName → fieldName[])
+  const fileToFeignFields = new Map<string, Array<{ typeName: string; fieldName: string }>>();
+  for (const [path, content] of fileContents) {
+    for (const typeName of feignTypeNames) {
+      const fieldPattern = makeFieldPattern(typeName);
+      let m: RegExpExecArray | null;
+      fieldPattern.lastIndex = 0;
+      while ((m = fieldPattern.exec(content)) !== null) {
+        const fieldName = m[1];
+        let list = fileToFeignFields.get(path);
+        if (!list) {
+          list = [];
+          fileToFeignFields.set(path, list);
+        }
+        list.push({ typeName, fieldName });
+      }
+    }
+  }
+
+  if (fileToFeignFields.size === 0) return;
+
+  // Track emitted edges to avoid duplicates
+  const emittedEdges = new Set<string>();
+
+  // ── Step 4: Scan Function/Method nodes for Feign calls ──
+  graph.forEachNode((node) => {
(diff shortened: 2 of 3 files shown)


gitnexus/src/core/ingestion/languages/java/spring-feign-rpc.ts
```diff
@@ -0,0 +1,189 @@
+/**
+ * Spring Cloud OpenFeign RPC heuristic: resolve method calls on
+ * @FeignClient-injected fields.
+ *
+ * `@FeignClient` interfaces have no implementation class in the source
```
magyargergo, line 5: **P0 — the premise stated here does not hold on current `main`.** `[reproduced]`

This header says the CALLS edge "is missing entirely" and impact analysis is "completely broken". Four independent checks — three by execution, across two engines — show the edge **already exists on main without this PR**, at confidence `0.85`, `reason: import-resolved`. That includes a run using *this PR's own unit-test shape* (`@Resource private ExeRulesRpcService`, `WinRpcResponse<Object> getRules(InputDTO)`), plus constructor injection, `@Autowired` fields, and wildcard imports.

`@FeignClient`'s defining trait is having no *implementation class*. But resolving `field.method()` never needed one — it needs the receiver's declared type and that type's method *declaration*, both present. `receiver-bound-calls.ts` emits the primary edge to the interface's own method; fan-out to implementors is secondary, and a Feign interface correctly contributes nothing there.

The structural consequence: this module can only emit edges to `Method` nodes already in the graph — in-repo Feign interfaces — exactly the set the resolver already resolves. If the interface comes from a JAR there's no target node, so it can't help there either. So every edge it emits duplicates an existing one, with a different id (`FEIGN_CALLS:…` vs `rel:CALLS:…`) that id-only dedupe can't collapse; the relationship CSV has no id column and rows are COPY-appended, so it reaches storage.

**Honest caveat:** a handful of fixtures don't prove *every* Feign shape resolves. Untested: cross-module/multi-Gradle setups, generic wrappers (`FeignClient<T>`), calls through a superclass-held field, and `@Configuration`-factory-supplied clients.

**So the ask isn't "fix the regex".** Please rebase, then produce **one integration fixture on current main where the existing resolver fails to emit the edge.** If it exists, the fix likely belongs in `receiver-bound-calls.ts` — and would fix that shape for every language at once. If it doesn't, the smaller idea underneath this PR is still genuinely useful: *tagging* already-resolved calls as RPC/service boundaries, as a metadata attachment on the existing capture/adapter/attacher pattern.

gitnexus/src/core/ingestion/languages/java/scope-resolver.ts
```diff
@@ -29,12 +29,33 @@ import {
   type JavaResolveContext,
 } from './index.js';
 import { populateJavaPackageSiblings } from './package-siblings.js';
+import { attachSpringBeanCandidateMetadata } from './spring-bean-metadata.js';
+import { attachJavaSpringAopMetadata } from './spring-aop.js';
+import { attachJavaSpringConfigBindings } from './spring-config-bindings.js';
+import { attachJavaSpringConditionalMetadata } from './spring-conditionals.js';
+import { attachJavaSpringDiMetadata } from './spring-di.js';
+import { attachJavaSpringDynamicLookup } from './spring-dynamic-lookup.js';
```
magyargergo, line 37: **P0 — this import resolves nowhere, and a rebase alone won't fix it.** `[reproduced]`

`npx tsc --noEmit` on this head fails with 9 errors — 8 missing modules imported by this hunk, plus `TS2353 'stripReceiverCastExpressions' does not exist in type 'ScopeResolver'`.

Seven of the eight are explained by the stale base (merge-base `57e4afa4c`, 2026-06-26, **466 commits behind main**) and a rebase fixes them. **`./spring-dynamic-lookup.js` is the exception:** it exists on neither the base, nor this head, nor `origin/main` — it's PR #2886's module, still unmerged. So this branch silently depends on #2886, and will still fail to compile after rebasing.

Two options: land #2886 first, or drop this import and its call at `:98` — neither is part of this PR's feature.

Worth saying clearly, because the diff overstates your footprint: **most of the +38 in this file isn't your change.** Diffed directly against `origin/main`, your version of `scope-resolver.ts` is main's file **plus exactly 4 lines** (2 imports, 2 calls). Everything else only looks added because the base is old. A concern that merging might drop main's other attachers was investigated and **refuted** — your block is a strict superset.

For sequencing: #2886 and #2891 also append at this same site, so merge order matters (#2886 → #2891 → this).

gitnexus/src/core/ingestion/languages/java/spring-feign-rpc.ts
```diff
@@ -0,0 +1,189 @@
...
+    const feignFields = fileToFeignFields.get(filePath);
+    if (!feignFields || feignFields.length === 0) return;
+
+    const source = fileContents.get(filePath);
+    if (!source) return;
+
+    // Extract function source by line range
+    const lines = source.split('\n');
+    const funcSource = lines
+      .slice(Math.max(0, startLine - 1), Math.min(lines.length, endLine))
```
magyargergo, line 156: **P0 — this line window is off by one in both directions.** `[reproduced]`

Graph node `startLine`/`endLine` are **0-based**. From `utils/line-base.ts`:

> "The graph layer stores line numbers 0-based (tree-sitter `startPosition.row`), and this is load-bearing: the taint/PDG/CFG join and the MCP consumers all add `+ 1` to recover 1-based."

Confirmed empirically — in a real pipeline run, a method occupying 1-based lines 11–14 reports `startLine 10, endLine 13`.

`slice(startLine - 1, endLine)` treats them as 1-based, so the window is shifted one line early at both ends. Executed against this module with production coordinates:

| Case | Expected | Actual |
|---|---|---|
| Feign call on the method's **last line** | 1 edge | **0 — silently missed** |
| Feign call in the **previous** method | 0 edges | **1 — attributed to the wrong method** |

The second is the serious one: a `CALLS` edge whose **source is the wrong method**. `impact` walks CALLS, so "who calls this RPC?" returns a method that doesn't.

The worst case is the idiomatic shape this feature targets — a one-line delegate. With `startLine == endLine`, the window becomes exactly the line *before* the method.

**Fix:** `lines.slice(startLine, endLine + 1)`. The codebase's canonical slice in `csv-generator.ts` already does it this way.

gitnexus/test/unit/spring-feign-rpc.test.ts
```diff
@@ -0,0 +1,451 @@
...
+    public Object ${funcName}() {
+${callLines}
+        return result;
+    }
+}`;
+  // Function spans from `public Object` (line 6) to closing `}` (line 10)
+  return {
+    path,
+    content,
+    node: makeFile(path, funcName, 'Method', 6, 10),
```
magyargergo, line 113: **P2 — the fixtures encode a 1-based convention, which is why 12 green tests can't see the off-by-one.** `[reproduced]`

These hand-built nodes (`startLine: 8, endLine: 10`, `12/14`, `16/18`) are 1-based. Production emits **0-based** node lines. Under a 1-based reading `slice(startLine - 1, endLine)` is exactly right, so the arithmetic error cancels here and every test passes while production reads the wrong window.

Two other ways this suite validates the implementation against its own assumptions rather than the pipeline:

1. **`makeFeignInterfaceFile` derives the interface name from the file basename** (`path.split('/').pop()!.replace('.java','')`) — the same assumption the implementation makes. No test can ever detect an interface whose declared name differs from its filename, or two same-named interfaces in different packages.
2. **The graph is a hand-rolled mock** cast `as unknown as KnowledgeGraph`, with no id-dedupe semantics — so the duplicate-edge behaviour is invisible here too.

Also: the "does not emit self-referencing edges" test is **vacuous** — its assertion sits inside a loop that never executes for its fixture, so it would pass even if self-edges were emitted.

**Fix:** switch fixtures to 0-based node lines (they'd then fail against the current implementation, which is the point), decouple the interface name from the filename, and add a `test/integration/*-pipeline.test.ts` driving the real pipeline — every sibling attacher registered in this same hook has one.

magyargergo (commented):
## Multi-engine review — PR #2890

**Verdict: `not production-ready`** — blocked on the premise, not on polish.
**Merge state: `checks failing`** · **Branch hygiene: `rebase/split required`**

*(Posted as a normal review comment — not a formal GitHub "Request changes" review state.)*

---

### First, what's genuinely creditable

- **The motivation is legitimate.** "Does my blast radius cross the RPC boundary?" is exactly the question this graph should answer.
- **451 lines of tests for a 189-line module, unprompted, from a fork PR** — 12 tests covering single/multiple methods, multiple injected clients, `@Autowired` and `@Resource`, the `@FeignClient(value = "…")` form, repeated-call dedupe, a self-edge guard, and a negative case. That is above the median and worth saying.
- **The correct extension point was found.** `emitPostResolutionEdges` is genuinely the right hook, and finding it meant reading the `ScopeResolver` contract rather than guessing. The hook is reachable in the real pipeline — explicitly *not* a "registered hook, zero call sites" defect.
- Single commit, single domain, no unrelated churn, `git diff --check` clean, no bidi/zero-width characters.

---

### The headline: the stated premise does not hold on current `main`

The module's header says the CALLS edge "is missing entirely" and that impact analysis is "completely broken" across Feign boundaries. **Four independent sources show the edge already exists on `main` without this PR**, at confidence `0.85`, `reason: import-resolved`:

1. **Real pipeline run** on a minimal fixture (`@FeignClient interface OrderClient` + `@Service` with `@Autowired` field): `handle -> getOrder (conf 0.85)`. Present. `[reproduced]`
2. **Four fixture variants** through `runPipelineFromRepo` — constructor injection, `@Autowired` field, wildcard `import rpc.*`, **and this PR's own unit-test shape** (`@Resource private ExeRulesRpcService`, `WinRpcResponse<Object> getRules(InputDTO)`). All four produce the edge. `[reproduced]`
3. **Built CLI on a 3-file fixture**, queried two ways: `cypher` returns `fetch | getOrder | 0.85 | import-resolved`; `impact "fetch" --direction downstream` returns it at `byDepth.1`. `[reproduced]`
4. **Codex (gpt-5.6-sol)** reached the same conclusion independently from `receiver-bound-calls.ts`. `[code-read]`

**Why the premise is wrong.** `@FeignClient`'s defining trait is having no *implementation class* in source. But resolving `field.method()` never needed one — it needs the receiver's declared type and that type's method *declaration*, both present. `receiver-bound-calls.ts` emits the **primary** edge to the interface's own method; fan-out to implementors is a **secondary** edge. A Feign interface with zero implementors correctly contributes nothing to that fan-out — right behaviour, not a gap.

**The structural clincher.** This module can only emit edges to `Method` nodes already in the graph — i.e. in-repo Feign interfaces — which is exactly the set the resolver already resolves. When the interface lives in a JAR there is no target node, so it cannot help there either. There is no residual gap for it to fill, and every edge it emits is a **duplicate** with a different id (`FEIGN_CALLS:…` vs `rel:CALLS:…`) that id-only dedupe cannot collapse. The persisted relationship CSV has no id column and rows are COPY-appended, so the duplication reaches storage.

#### The honest caveat

**A handful of fixtures do not prove every Feign shape resolves.** What is disproven is the premise *as stated*, for the archetypal case, across four injection styles and two import styles. Shapes **not** tested, which may still have real gaps: cross-module/multi-Gradle setups; generic wrappers (`FeignClient<T>`, `extends BaseClient<Order>`); calls through a superclass-held field; `@FeignClient` interfaces consumed via a `@Configuration` factory; and interfaces supplied by a JAR (where this module cannot help either).

**So the ask is not "fix the regex" — it is: rebase, then produce one integration fixture on current `main` where the existing resolver fails to emit the edge.** If that fixture exists, the fix almost certainly belongs in `receiver-bound-calls.ts`, where receivers are already typed and overloads already narrowed — and it would fix the same shape for *every* language at once. If it does not, the genuinely useful smaller idea underneath this PR — **tagging** already-resolved calls as RPC/service boundaries for topology and impact queries — is worth reopening as a metadata-attachment change on the existing capture/adapter/attacher pattern.

---

### Build-blocking facts

- **The head does not compile.** `npx tsc --noEmit` → 9 errors: 8 × `TS2307 Cannot find module` for `./spring-bean-metadata.js`, `./spring-aop.js`, `./spring-config-bindings.js`, `./spring-conditionals.js`, `./spring-di.js`, `./spring-dynamic-lookup.js`, `./capture-side-channel.js`, `./package-facts.js`, plus `TS2353 'stripReceiverCastExpressions' does not exist in type 'ScopeResolver'`.
- **The base is 466 commits / ~6 weeks stale** (merge-base `57e4afa4c`, 2026-06-26). Seven of the eight missing modules exist on current `main` — a rebase fixes those.
- **A rebase alone does NOT fix the build.** `spring-dynamic-lookup.ts` exists on **neither** the base, **nor** this head, **nor** `origin/main`. It is PR #2886's module, still unmerged — so this PR silently depends on #2886. Either land #2886 first, or drop that import and call (they aren't this PR's feature).
- **Most of the `scope-resolver.ts` +38 is not your work.** Diffed directly, the PR's version of that file is `origin/main`'s **plus exactly 4 lines** (2 imports, 2 calls). The rest only appears as additions because the base is stale. *A worry that merging would drop main's other attachers was investigated and refuted.*
- **True review surface after a correct rebase:** `spring-feign-rpc.ts` + its tests + 2 lines of wiring.

---

### Why CI never caught this

Only two check-runs exist on this head: `Apply conventional label` (success) and `Validate PR title` (skipped). **The quality/test suite never executed** — GitHub gates workflow runs on `pull_request` from first-time fork contributors pending maintainer approval. The repo's own typecheck job would have caught the compile failure immediately. **This PR is unverified, not green.** Maintainer action: approve workflow runs so real signal appears.

*(One reviewer suggested migrating these workflows to `pull_request_target` — that would run fork-authored code against the base repo's write token and is not recommended.)*

---

### Findings

Four are inline. The rest are below — deliberately not inlined, because they are downstream of the premise question and there is no point rewriting a regex that may be deleted.

| # | Sev | Issue | Corroboration | Tag |
|---|---|---|---|---|
| H1 | P0 | Premise already satisfied on main; every edge is a non-dedupable duplicate | both engines | `[reproduced]` |
| H2 | P0 | Head fails `tsc`; hidden dependency on unmerged #2886 | both engines | `[reproduced]` |
| H3 | P0 | Line window off by one **in both directions** | both engines | `[reproduced]` |
| H12 | P2 | Unit suite encodes a 1-based convention, so it structurally cannot catch H3 | Claude ×2 | `[reproduced]` |

**H3 deserves emphasis.** Graph node `startLine`/`endLine` are **0-based** (`utils/line-base.ts`: *"The graph layer stores line numbers 0-based … this is load-bearing"*), but the module slices `lines.slice(startLine - 1, endLine)` as if 1-based. Executed against the real module with production coordinates:

- a Feign call on a method's **last line** → **0 edges** (silently missed);
- a Feign call in the **previous** method → **1 edge attributed to the wrong method**.

Idiomatic one-line Spring delegates are the worst case. Fix: `lines.slice(startLine, endLine + 1)`.

#### Additional issues, if the module survives premise verification

All real and evidence-backed, recorded here in full so they're available the moment they become relevant:

- **P1 — Feign identity is the file basename.** `feignTypeNames.has(fileName)` means an interface whose declared name differs from its filename indexes nothing, and two same-named interfaces in different packages collide into one bucket. Contrast `bean-candidates.ts`, which resolves FQNs import-aware and fails closed.
- **P1 — one Javadoc sentence disables the feature.** `INTERFACE_NAME_RE = /interface\s+(\w+)/` is `exec`'d once, unanchored, uncorrelated with the `@FeignClient` test — so a comment reading "Remote interface for the order service" captures the type name `"for"` and silently drops the file.
- **P1 — raw-text matching fabricates edges** from commented-out calls, calls inside string literals, and Javadoc `{@code …}`; a parameter or local shadowing the field name emits phantom edges; `FallbackOrderClient` matches as `OrderClient` (the field pattern has no leading word boundary).
- **P1 — `O(files × feignTypes × fileLength)`**, measured at **+60.5s** on a 20,050-file workspace with 50 Feign clients, to emit 200 duplicate edges. Compounded by `source.split('\n')` **per graph node** (~100× on a 2,000-line/100-method file) and a fresh `RegExp` per (node × field).
- **P2 — one call site fans out to every same-named overload**; inherited `extends BaseRpc` methods are unreachable; the "does not emit self-referencing edges" test is **vacuous** (its assertion loop never executes).
- **P3 — dead declarations:** the `FeignMethodInfo` interface is unused, and `parsedFiles` is accepted but never read — the AST is available and ignored in favour of regex.

On `confidence: 0.85` — **it is not an invented number.** It's precedented in this repo (`1.0` ×39, `0.9` ×35, `0.85` ×13) and matches what the resolver already emits for these very calls. The consequence is narrower: since `local-backend.ts` filters `r.confidence >= $minConfidence` and ranks by it, no threshold can separate a regex guess from a proven edge.

---

### Graded on the four requested dimensions

| Dimension | Grade | Basis |
|---|---|---|
| **Performance** | F | `O(files × feignTypes × fileLength)` with per-node full-file `split`, measured +60.5s/20k files — for duplicate edges. *Credit: no ReDoS; all four patterns checked and refuted.* |
| **Architectural fit** | F | Regex over raw source in a tree-sitter AST pipeline, bypassing `createSpringAnnotationNameResolver` (import-aware, alias-aware, fail-closed) and the capture/adapter/attacher pattern every sibling `frameworks/spring/*` module follows. *Credit: the extension point itself is correct.* |
| **No overengineering** | F | Not gold-plated — 189 lines is lean. Graded F because the whole mechanism appears unnecessary, which is the extreme case of building more than is needed. *Credit: the module is not itself bloated.* |
| **Reusability** | F | Hand-rolls FQN resolution, field-type discovery, line slicing and edge dedupe, each of which exists — `bean-candidates.ts`, the DI facts, `csv-generator.ts`'s canonical `lines.slice(startLine, endLine + 1)`, and the graph's own id dedupe. |

The four Fs are evidence-matched, and each carries its credit line. They describe the *mechanism*, not the contributor.

---

### Checked and refuted

Investigated and disproven — recorded because rigor cuts both ways: regex metacharacter injection (both interpolated values come from `(\w+)`, excluding `$`); ReDoS/catastrophic backtracking; `fileContents` unavailability on warm/incremental runs (built once per run from the full file set); nondeterminism; re-entrancy; unreachable hook; edge-id writer breakage; cross-language `fileContents` pollution; `impact` result inflation (it dedupes by relationship id); and the concern that merging would silently drop main's other attachers.

---

### Test gaps

The 12 tests pass **because they import only the new module** — the broken `scope-resolver.ts` is never in their module graph, so a green suite coexists with a package that does not build. They also use a hand-rolled mock graph, hardcode 1-based coordinates (masking H3), and build fixtures whose interface name is *derived from the filename*, which bakes in the implementation's identity assumption so no test can ever detect it. There is no `test/integration/*-pipeline.test.ts`, though every sibling attacher registered in the same hook has one.

---

### Coverage limits

- **One lane produced nothing, twice** (`gitnexus-risk-architect`). Production-risk coverage is thinner than the rest of this review; two of its questions were answered directly instead.
- **Only one Codex leg ran** — every "cross-engine" label rests on a single Codex lane, not several.
- The performance lane could not execute (the module doesn't build in the stale worktree); its cost models are analytical, and the +60.5s figure comes from a synthetic harness — treat the complexity class as the durable claim and the absolute seconds as order-of-magnitude.
- The premise refutation is bounded by its fixtures — see the caveat above.

*Automated multi-tool digest (8 Claude lanes + 1 Codex lane, merged by a synthesis pass and gated by an independent critic). Every finding lists its trigger, verified `file:line`, and whether it was reproduced or code-read. Please verify against the source before acting.*

ChunxueLi:
Thanks for the thorough multi-engine review — much appreciated.

**Context on the original premise:** the "CALLS edge missing entirely" observation was made against a build affected by a separate regression (a compiled-dist regex missing its `g` flag in `makeFieldPattern`, which sent `exec` into an infinite loop and prevented Feign fields from resolving at all). With that regression fixed, we agree the existing import-resolved fallback on `main` covers the primary scenarios — your four independent reproductions confirm it.

Given that:
- the off-by-one (`slice(startLine - 1, endLine)` vs 0-based node lines) and the stale merge-base (466 commits behind) would require a substantial rework,
- and the original motivation no longer holds on current `main`,

we're closing this PR. The heuristic may still add value for cases where the Feign interface import cannot be statically resolved (e.g. same-package interfaces without explicit imports), but that would need a fresh, evidence-backed premise — so if we pursue it, it will be as a new PR rebased on current `main` with 0-based fixtures.

Keeping #2883 / #2885 / #2886 open for review — thanks again!

PR #2891, opened 2026-08-08
feat(spring): detect non-HTTP handler entry points

### LINKED ISSUE #2417: feat(spring): detect scheduled event and messaging handlers
Parent tracking issue: #1225

Status: remaining work

Goal
Treat non HTTP Spring handlers as application entry points.

Scope
Detect Scheduled methods, EventListener methods, common message listener annotations, and listener style methods that Spring invokes without an HTTP route.

Acceptance criteria
1. Scheduled methods are marked as entry points.
2. Event listener methods are marked as event driven handlers.
3. Common messaging listener annotations are represented where the dependency is present in source.
4. Process or context output can start from these handlers the same way it can start from HTTP routes.
5. Tests cover scheduled, event listener, and one messaging listener fixture.

ShiningXu:
## Summary

- detect Spring non-HTTP handler methods as process entry points across Java and Kotlin
- cover scheduling, application events, messaging annotations, and XXL-Job `@XxlJob`
- resolve imported, fully qualified, and Kotlin-aliased annotations while failing closed on unrelated same-name annotations
- persist capture side-channel facts and bump the analysis feature version for incremental reanalysis
- add integration fixtures for positive paths, multi-annotation methods, interface/use-site exclusions, reason precedence, aliases, constants, and false positives

Closes #2417

## Behavior and design notes

- Recognized handlers receive an entry-point multiplier of `3.0`; this does not create HTTP route nodes.
- Existing same-strength AST framework reasons are preserved; a non-HTTP reason is filled only when one is missing.
- Kotlin intentionally captures annotated callables before import resolution so aliases such as `EventListener as SpringEvent` and `XxlJob as JobEntry` remain detectable.
- XXL-Job support identifies handler entry points only. Scheduler-fact/job-name extraction remains outside this change.
- Class-level `@KafkaListener` paired with method-level `@KafkaHandler` is not included because it needs class/method semantic pairing; method-level listener annotations are supported.

## Validation

- ✅ `npm run build`
- ✅ `npx tsc --noEmit`
- ✅ `npx vitest run test/integration/spring-non-http-handlers-pipeline.test.ts` — 22/22
- ✅ `npx vitest run test/unit/analysis-features.test.ts test/unit/incremental-orchestration.test.ts` — 26/26
- ✅ pre-commit ESLint, Prettier, and TypeScript checks
- ✅ GitNexus `detect-changes` — 25 files, 42 symbols, 4 affected flows, medium risk
- ⚠️ `npm test` — 16,734 passed, 12 failed, 101 skipped. The 12 failures are in untouched areas: local-backend impact BFS (also reproduced in isolation), macOS path canonicalization/Linux-only fd anchoring, process-lock/hook timing, and review-agent install retry.

---

## Summary

*A broad ingestion and analysis change centered on Spring non HTTP handler detection across Java and Kotlin, with supporting cache and test updates. Its graph reach is critical, so reviewers should focus on the capture and resolution paths first.*

**🔴 CRITICAL blast radius. A Java and Kotlin ingestion change for Spring non HTTP handler entry points, reaching through the analysis pipeline and its dependent code.**

The hottest file is `gitnexus/src/core/ingestion/languages/java/capture-side-channel.ts`, with 11 changed symbols, including `collectJavaCaptureSideChannel`, `applyJavaCaptureSideChannel`, `setJavaSpringAopFacts`, and `clearJavaClassAnnotationFacts`. The corresponding Kotlin capture path in `gitnexus/src/core/ingestion/languages/kotlin/capture-side-channel.ts`, the Spring analysis files `gitnexus/src/core/ingestion/frameworks/spring/analysis-features.ts` and `gitnexus/src/core/ingestion/frameworks/spring/non-http-handlers.ts`, and the language resolvers are the first places to review. The change also touches `gitnexus/src/storage/parse-cache.ts`, including `sha256Hex` and `ParseCache`.

The graph places the work across `Java`, `Kotlin`, `Storage`, `Integration`, `Languages`, `Spring`, `Pipeline-phases`, and `Ingestion`, with 4 affected flows and dependents reaching through d3. Review the Java and Kotlin capture emission functions, especially `emitJavaScopeCaptures` and `emitKotlinScopeCaptures`, alongside `gitnexus/src/core/run-analyze.ts`. The Spring fixture handlers and validation in `gitnexus/test/integration/spring-non-http-handlers-pipeline.test.ts` provide the most direct coverage of the new entry point patterns.

_Added by GitNexus for PR #2891. Edit freely — this block is replaced on the next review, everything above it is left untouched._


gitnexus/src/core/ingestion/languages/java/capture-side-channel.ts
```diff
@@ -24,13 +25,15 @@ export interface JavaCaptureSideChannel {
   readonly springConfigConsumers?: readonly JavaSpringConfigConsumerFact[];
   readonly springConditionalFacts?: readonly JavaSpringConditionalFact[];
   readonly springDiFacts?: readonly JavaSpringDiClassFact[];
+  readonly springNonHttpHandlerFacts?: readonly JavaSpringNonHttpHandlerFact[];
```
magyargergo, line 28: **P1 · blocking — this new side-channel field needs a `SCHEMA_BUMP`.** `[code-read]`

Adding `springNonHttpHandlerFacts` changes what `ParsedFile` carries, but `SCHEMA_BUMP` is still `53` at `gitnexus/src/storage/parse-cache.ts:391` (identical base→head). On any index built by the same package version, unchanged `.java`/`.kt` files hit the content-addressed cache, `applyJavaCaptureSideChannel` finds no `springNonHttpHandlerFacts` key, defaults to `[]`, and **zero methods get stamped**.

The analysis-feature mismatch sets `force: true`, but that only rebuilds the graph — `run-analyze.ts:1648` calls `loadParseCache` unconditionally, and that file's comment says outright *"Content-addressed: safe to reuse across `--force` runs."* The graph is then rebuilt from the same stale ParsedFiles.

It also self-seals: `run-analyze.ts:3299` stamps `analysisFeatures` on success regardless, so the index records `spring.non-http-handlers: 1` as satisfied and **no later run ever re-forces**.

`parse-cache.ts`'s own comment is explicit about this failure mode: *"A warm cache would replay the pre-fix ParsedFiles, so every file served from it would carry the old captures while passing every cold-run test — the exact failure this constant exists to prevent."* The last eight capture-changing PRs each bumped it (38→39→42→43→44→45→47→48→53).

Graded P1 rather than P0 because `PARSE_CACHE_VERSION` also folds in the package version, so upgrading *across* releases invalidates anyway — the exposed population is same-package-version rebuilds (dev loops, CI, `npm link`, this repo's own index).

**Fix:** bump 53→54 with a ledger entry naming #2417/#2891, re-check the value against `origin/main` right before merge (#2883/#2885/#2886/#2890 are in flight and the ledger records prior exact clashes), and add a cold→save→warm parity test — `spring-aop-pipeline.test.ts` already has the shape.

A smaller related point: `setJavaSpringNonHttpHandlerFacts(path, [])` *deletes* the key while the getter returns `[]` for a missing key, so "captured, no handlers" and "ParsedFile predates the feature" are indistinguishable downstream.

gitnexus/src/core/ingestion/languages/kotlin/captures.ts
```diff
@@ -141,6 +147,9 @@ export function emitKotlinScopeCaptures(
       springConditionalFacts.push(
         ...captureKotlinSpringConditionalFacts(springDiClassNode, filePath),
       );
+      springNonHttpHandlerFacts.push(
+        ...captureKotlinSpringNonHttpHandlerFacts(springDiClassNode, filePath),
```
magyargergo: **P2 — Kotlin `object` and `companion object` handlers are never captured.** `[reproduced]`

This call sits inside the `springDiClassNode` block, which is `nodeIfType(groupedNodes['@scope.class'], 'class_declaration')` — `class_declaration` only. Ten lines above, the **AOP** hook builds its node from `class_declaration` **OR** `object_declaration` **OR** `companion_object`. The fix already exists verbatim in this same function.

Missed today:

```kotlin
@Component object CacheWarmer {
    @Scheduled(fixedRate = 60_000) fun warm() { ... }   // no fact, no stamp
}

class Foo { companion object { @Scheduled fun tick() {} } }   // same
```

Kotlin `object` is a common home for stateless `@Scheduled`/`@KafkaListener` singletons. Live tree-sitter probes against the vendored grammar confirm none of these shapes reaches the new code: `object_declaration > class_body > function_declaration` and `class_declaration > class_body > companion_object > class_body > function_declaration`.

`enum class` is missed too, on a second gate — `kotlin/spring-non-http-handlers.ts:23` accepts only `class_body`, not `enum_class_body`.

Worth noting the sibling AOP suite already fixtures `object KotlinAspect` and `companion object { @Tx fun companionTransaction() }`, so there's a ready template for the test.

gitnexus/src/core/ingestion/frameworks/spring/non-http-handlers.ts
```diff
@@ -0,0 +1,180 @@
...
+  isPackageVisibilityIncomplete(filePath: string): boolean;
+}
+
+const HANDLER_ANNOTATIONS = new Map<string, SpringNonHttpHandlerKind>([
+  ['org.springframework.scheduling.annotation.Scheduled', 'scheduled'],
+  ['org.springframework.scheduling.annotation.Schedules', 'scheduled'],
+  ['org.springframework.context.event.EventListener', 'event'],
+  ['org.springframework.transaction.event.TransactionalEventListener', 'event'],
+  ['org.springframework.modulith.events.ApplicationModuleListener', 'event'],
+  ['org.springframework.kafka.annotation.KafkaListener', 'message'],
```
magyargergo, line 45: **P2 — the class-level `@KafkaListener` + method-level `@KafkaHandler` idiom produces zero entry points.** `[code-read]`

The catalog has `KafkaListener`/`KafkaListeners`/`RabbitListener`/`RabbitListeners`, but `grep -c "KafkaHandler\|RabbitHandler"` over this file returns **0**. The capture only reads each `method_declaration`'s own annotations, so a class-level `@KafkaListener` is never consulted as evidence for its members:

```java
@KafkaListener(id = "multi", topics = "orders")
public class MultiTypeConsumer {
    @KafkaHandler public void created(OrderCreated e) { ... }
    @KafkaHandler(isDefault = true) public void unknown(Object o) { ... }
}
```

`simpleName('KafkaHandler')` isn't in `CAPTURE_RELEVANT_SIMPLE_NAMES`, so **no fact is recorded at all** — three live message entry points, zero output. Identical for Spring AMQP's `@RabbitListener` + `@RabbitHandler`.

The PR body notes this pairing was consciously deferred as needing class/method semantic pairing, which is fair. Two options: add `org.springframework.kafka.annotation.KafkaHandler` and `org.springframework.amqp.rabbit.annotation.RabbitHandler` as `'message'` (simple, slightly over-broad), or record the enclosing class's annotations on the fact so `@KafkaHandler` promotes only when its class resolves to `@KafkaListener` (precise). Either way, worth a tracked follow-up issue — right now the deferral isn't filed anywhere.

gitnexus/src/core/ingestion/frameworks/spring/non-http-handlers.ts
```diff
@@ -0,0 +1,180 @@
...
+  ['org.springframework.amqp.rabbit.annotation.RabbitListeners', 'message'],
+  ['org.springframework.jms.annotation.JmsListener', 'message'],
+  ['org.springframework.jms.annotation.JmsListeners', 'message'],
+  ['org.springframework.pulsar.annotation.PulsarListener', 'message'],
+  ['org.springframework.pulsar.annotation.PulsarListeners', 'message'],
+  ['io.awspring.cloud.sqs.annotation.SqsListener', 'message'],
+  ['io.awspring.cloud.messaging.listener.annotation.SqsListener', 'message'],
+  ['org.springframework.cloud.aws.messaging.listener.annotation.SqsListener', 'message'],
+  ['org.springframework.cloud.stream.annotation.StreamListener', 'message'],
+  ['org.springframework.integration.annotation.ServiceActivator', 'message'],
```
magyargergo: **P2 — `@ServiceActivator` has a common `@Bean` form that shouldn't be promoted.** `[code-read]` *(single lane, single engine — weakest corroboration of the set)*

The Java adapter's only gate is `hasSpringNonHttpHandlerRelevantAnnotation(annotations)`, so the annotation's *form* is never discriminated. This is the Spring Integration reference idiom:

```java
@Bean
@ServiceActivator(inputChannel = "errorChannel")
public MessageHandler errorHandler(KafkaTemplate<?, ?> template) { ... return handler; }
```

That's a bean **factory** Spring calls once at context refresh — the real handler is the returned object, not this method. It gets stamped 3.0/`spring-message-handler`; it has non-zero callees (it constructs the handler) so it survives the `callees.length === 0` guard; and with `callerCount = 0` it scores high. The user ends up with a "process" whose entry point is a `@Bean` factory and whose trace is object construction.

`@ServiceActivator` is the only catalog entry with a common bean-factory form. Either discriminate the `@Bean`-annotated case, or drop it from the catalog until that discrimination exists.

gitnexus/src/core/ingestion/frameworks/spring/non-http-handlers.ts
```diff
@@ -0,0 +1,180 @@
...
+  Annotation extends SpringNonHttpHandlerAnnotationFact,
+>(adapter: SpringNonHttpHandlerAdapter<Annotation>) {
+  return (
+    graph: KnowledgeGraph,
+    parsedFiles: readonly ParsedFile[],
+    nodeLookup: GraphNodeLookup,
+    indexes: ScopeResolutionIndexes,
+  ): void => {
+    const resolveAnnotation = createSpringAnnotationNameResolver(indexes);
+    const exactOwnerByRange = exactCallableOwnersByRange(graph);
```
magyargergo: **P2 — two whole-graph scans built eagerly, before any fact is known to exist.** `[code-read]`

This `HAS_METHOD` walk and `exactCallableOwnersByRange`'s full `graph.iterNodes()` (line 76) both run before the `for (const parsed of parsedFiles)` loop, unconditionally, once for the Java attacher and again for Kotlin.

Two things make this avoidable rather than inherent:

1. **`aop.ts:292-303` already built a strict superset** of `exactCallableOwnersByRange` — same graph, same synchronous `emitPostResolutionEdges` call, three statements earlier.
2. **`aop.ts:304-316` defers its `classIdByMethod`** behind a lazy `singletonOwnerId()` closure that only fires when a fact needs it. This one doesn't.

On a JVM repo with zero Spring annotations the work is entirely wasted; Java is 5th and Kotlin 11th in `SCOPE_RESOLVERS` over a shared graph, so a polyglot repo with a handful of JVM files still pays close to full-graph cost, twice. On this repo's own reference scale (248k nodes / 565k relationships) that's roughly 500k extra node visits plus two full `defs` scans. Linear, not quadratic — hence P2.

Highest-value fix is one line, before any index build:

```ts
if (!parsedFiles.some((p) => adapter.getFacts(p.filePath).length > 0)) return;
```

Then make `classByMethod` lazy exactly as `aop.ts:304` does, and store `sourceId` strings instead of `GraphNode`s — the only use is `?.label`.

To be fair: the eager-full-scan *pattern* is inherited from `aop.ts`/`conditionals.ts`/`config-bindings.ts`. What's squarely on this PR is that the lazy precedent sits 200 lines away and wasn't followed.

gitnexus/src/core/ingestion/languages/kotlin/spring-non-http-handlers.ts
```diff
@@ -0,0 +1,43 @@
...
+  classNode: SyntaxNode,
+  filePath: string,
+): KotlinSpringNonHttpHandlerFact[] {
+  const facts: KotlinSpringNonHttpHandlerFact[] = [];
+  const body = classNode.namedChildren.find((child) => child.type === 'class_body');
+  if (body === undefined) return facts;
+  for (const member of body.namedChildren) {
+    if (member.type !== 'function_declaration') continue;
+    const annotations = kotlinSpringAnnotationFacts(member);
+    if (annotations.length === 0) continue;
```
magyargergo, line 32: **P2 — every annotated function is recorded, and each fact ships two fields the attacher never reads.** `[code-read]`

The no-prefilter gate itself is **correct and precedented** — `kotlin/spring-aop.ts:56-71` does the same thing for the same alias reason, and dropping it would break `import EventListener as SpringEvent`. Not asking you to change that.

The cost is what rides along. The shared `SpringNonHttpHandlerAnnotationFact` declares only `name` and `useSiteTarget`, but this passes `KotlinAnnotationSyntaxFact` through verbatim, which also carries `line` and — via `SpringDiAnnotationFact` (`di-metadata.ts:37-40`) — `text`, the full raw annotation source. Neither is read anywhere in the attacher, yet both cross the structured-clone worker boundary and land in the durable ParsedFile store.

On Kotlin codebases with annotation density outside Spring (`@Composable`, `@Test`, `@JvmStatic`, `@Suppress`) every such method now produces a second structurally near-identical record alongside the pre-existing `springAopFacts`, each carrying full annotation text.

Cheapest fix that keeps alias support fully intact — project the fact down at construction:

```ts
annotations: annotations.map((a) => ({
  name: a.name,
  ...(a.useSiteTarget !== undefined ? { useSiteTarget: a.useSiteTarget } : {}),
})),
```

Alias resolution only ever reads `annotation.name`, so this touches nothing load-bearing and removes the dominant per-fact cost.

gitnexus/test/integration/spring-non-http-handlers-pipeline.test.ts
```diff
@@ -0,0 +1,98 @@
...
+const FIXTURE = path.resolve(__dirname, '..', 'fixtures', 'spring-non-http-handler-app');
+
+describe('Spring non-HTTP handler entry points (#2417)', () => {
+  let result: PipelineResult;
+  const methods = new Map<string, GraphNode>();
+
+  beforeAll(async () => {
+    result = await runPipelineFromRepo(FIXTURE, () => {}, {});
+    result.graph.forEachNode((node) => {
+      if (node.label === 'Method') methods.set(node.properties.name, node);
```
magyargergo: **P2 (latent) — keying by bare method name; the fixture already contains duplicates.** `[reproduced]`

`executeJob` and `recordCompletion` each exist in **both** `XxlJobs.java` (lines 18, 22) and `AliasedXxlJob.kt` (lines 11, 15). `.set()` is last-write-wins over `forEachNode` order, which depends on per-worker parse completion — not something this test controls or documents.

No assertion flips today (none of the 13 asserted names collides), so this isn't breaking anything now. But it's a trap for the next person: the natural way to extend this fixture is to add another private helper with a natural name, and the failure would be an order-dependent flake rather than a clear error.

The sibling `spring-config-pipeline.test.ts:26-31` already solves this with `nodeNamed(name, fileSuffix)`. Cheapest version here: key by `${filePath}#${name}`, or throw in `beforeAll` on a duplicate key so a future collision fails loudly instead of silently.

gitnexus/src/core/ingestion/frameworks/spring/non-http-handlers.ts
```diff
@@ -0,0 +1,180 @@
...
+  ['org.springframework.cloud.stream.annotation.StreamListener', 'message'],
+  ['org.springframework.integration.annotation.ServiceActivator', 'message'],
+  ['org.springframework.messaging.handler.annotation.MessageMapping', 'message'],
+  ['org.springframework.messaging.simp.annotation.SubscribeMapping', 'message'],
+  ['com.xxl.job.core.handler.annotation.XxlJob', 'xxl-job'],
+]);
+
+const RECOGNIZED_HANDLER_ANNOTATIONS = new Set(HANDLER_ANNOTATIONS.keys());
+
+function simpleName(name: string): string {
```
magyargergo, line 69: **P3 — nit, entirely optional.** `[code-read]`

`di-metadata.ts:112` already exports `springAnnotationSimpleName`, which is identical to this.

Flagging it gently, because you're following the house pattern rather than breaking it: `aop.ts:203`, `bean-factories.ts:33` and `conditionals.ts:83` each carry their own private copy too. So this is the fifth copy, not the first — a pre-existing cleanup opportunity that this PR happens to touch, not something to fix here unless you want to.

If anyone does tidy it later, the move is to hoist one copy into `bean-candidates.ts` (already the shared import for `createSpringAnnotationNameResolver`) and delete the rest along with the now-unused export.

magyargergo (commented):
## Multi-engine review — PR #2891

**Verdict: `not production-ready`** — one blocking issue (F1). Everything else is a follow-up, a question, or a nit.

This is a well-built PR: the architecture matches the established Spring family, the fail-closed annotation resolution held up against every attack we threw at it, and the deliberate Kotlin-alias design is correct and tested. The single blocker is a missing cache-invalidation bump that makes the whole feature silently inert on warm indexes.

### Engines

| Engine | Legs | Status |
|---|---|---|
| Claude | GitNexus PR swarm (6 lanes) + `ce-code-review` personas (7 lanes) | 12/12 live |
| Codex `gpt-5.6-sol`/xhigh | swarm | live |
| Codex `gpt-5.6-sol`/xhigh | `ce`, adversarial | **dead** — both aborted on the org's tokens-per-minute rate limit |

Two of three Codex legs produced nothing, so **cross-engine confirmation runs through a single Codex lane**. Findings marked *Claude-only* below are consistent across personas but have **not** been independently confirmed by a second engine — shared blind spots are possible. Codex also cited correct line numbers with systematically wrong file paths; every path here was re-verified in a worktree at `46231d6d1`.

---

### What's solid

- **Fail-closed annotation resolution survived every attack.** Wildcard import + same-package user-defined `@interface Scheduled`; the same simple name from two explicit imports; two wildcard imports; a Kotlin `import … as W` layered over a local class `W`; a locally-nested `@interface`. All fail closed. The fixture encodes the property (`com.example.fake.EventListener` / `com.example.fake.XxlJob` asserted **not** stamped). This is the feature's most important safety property and it is both enforced and tested.
- **Architecture matches the Spring family exactly** — `SpringNonHttpHandlerAdapter` + `createSpringNonHttpHandlerMetadataAttacher` mirror the AOP pair; the shared layer owns the FQN catalog, language adapters own AST shape only, and **no logic branches on a language**.
- **Both hooks have live production call sites** (`java/scope-resolver.ts:96`, `kotlin/scope-resolver.ts:146`) — clears this repo's two recurring framework-hook traps (#2872 "gate not code", #2864 "registered hook, zero call sites").
- **Kotlin alias support is deliberate and proven end-to-end**; interface and use-site-target exclusions are implemented and tested; all 21 catalog FQNs name real declaring packages; multiplier precedence is correct in every reachable case.
- **Security clean on every category** — no secrets, injection, unsafe deserialization, or auth surface; all 15 new files byte-verified pure ASCII (no bidi, zero-width, or homoglyph characters in the FQN catalog).
- **No quadratic behaviour** — every index is built once per invocation, outside the per-file loop.

---

### Headline (F1 · P1 · blocking)

**A warm parse cache suppresses the entire feature, and the same run then writes a receipt saying the feature is done.**

`SCHEMA_BUMP` is unchanged at **`gitnexus/src/storage/parse-cache.ts:391`** (`const SCHEMA_BUMP = 53;`, identical base→head) while this PR adds `springNonHttpHandlerFacts` to both JVM capture side-channels — data that serializes into `ParsedFile`.

The chain: post-upgrade `analyze` → feature mismatch sets `force: true` → **`force` does not clear the parse cache** (`run-analyze.ts:1648` calls `loadParseCache` unconditionally; that file's own comment: *"Content-addressed: safe to reuse across `--force` runs"*) → unchanged `.java`/`.kt` hit on content hash → `applyJava/KotlinCaptureSideChannel` reads a payload with no `springNonHttpHandlerFacts` key → defaults to `[]` → **zero methods stamped**. The forced rebuild rebuilds the graph from those stale ParsedFiles, so it doesn't help.

Then it self-seals: `run-analyze.ts:3299` stamps `analysisFeatures: currentAnalysisFeatures` unconditionally on success, so feature parity now holds and **no future run re-forces**. The index permanently advertises *"Durable completeness contract for scheduled, event, messaging, and job entry points (#2417)"* while containing zero handler evidence. No automatic recovery.

`parse-cache.ts`'s own comment states the invariant: *"A warm cache would replay the pre-fix ParsedFiles, so every file served from it would carry the old captures while passing every cold-run test — the exact failure this constant exists to prevent."* Each of the last eight capture-changing PRs in this series bumped it (38→39→42→43→44→45→47→48→53).

**Why P1 and not P0:** `PARSE_CACHE_VERSION = ${SCHEMA_BUMP}+${GITNEXUS_PKG_VERSION}`. None of four sampled `SCHEMA_BUMP` commits touched `package.json`, so the version component is constant *within* a release cycle — but a user upgrading *across* published releases is invalidated by it anyway. The exposed population is same-package-version rebuilds: every dev loop, CI run, `npm link` install, and this repo's own index.

**Fix:** bump 53→54 with a ledger entry naming #2417/#2891, and **re-check against `origin/main` immediately before merge** — the ledger records prior exact clashes and #2883/#2885/#2886/#2890 are in flight. Add a cold→save→warm parity test; `spring-aop-pipeline.test.ts` already has the pattern.

---

### Other findings (details inline)

| # | Sev | Issue | Corroboration |
|---|---|---|---|
| F2 | P2 | Kotlin `object` / `companion object` / `enum class` handlers never captured — the AOP hook 10 lines above already handles all three | Claude ×3 |
| F3 | P2 | Class-level `@KafkaListener` + `@KafkaHandler` yields zero entry points; neither `KafkaHandler` nor `RabbitHandler` is in the catalog | Claude ×2 |
| F4 | P2 | `@Bean`-form `@ServiceActivator` promoted as a handler when it's a bean factory | Claude ×1 |
| F5 | P2 | New 3.0 stamps compete for the fixed 200-slot entry-point budget → existing processes can be evicted | both engines |
| F6 | P2 | Leaf handlers stamped but never promoted — #2417 AC4 unmet for zero-callee handlers | both engines |
| F7 | P2 | Two unconditional whole-graph scans per JVM language, built before any fact is known to exist | both engines, 5 lanes |
| F8 | P2 | Kotlin records every annotated function and ships `text`+`line` the attacher never reads across the worker boundary and onto disk | both engines |
| F9 | P2 | Integration test keys methods by bare name; `executeJob`/`recordCompletion` already exist in two fixture files | Claude ×5 |
| F10 | P3 | `simpleName` re-declared rather than importing the exported `springAnnotationSimpleName` — nit, four local copies already exist | both engines |

**F5 and F6 live in `process-processor.ts`** (`:407` `.slice(0, 200)`, `:367` `if (callees.length === 0) continue;`), which this PR doesn't touch — both mechanisms are pre-existing; what's new is the population loaded onto them. Before merge it's worth running `detect_changes --scope compare --base-ref main` on one large Spring repo to confirm the Process count doesn't **drop**.

**F7's sharpest detail:** `aop.ts:292-303` builds a **strict superset** of `exactCallableOwnersByRange` over the same graph, in the same synchronous call, three statements earlier — and `aop.ts:304-316` defers its `classIdByMethod` behind a lazy closure. The new attacher does neither. The highest-value fix is one line: `if (!parsedFiles.some((p) => adapter.getFacts(p.filePath).length > 0)) return;` before any index build.

---

### Open question (not a finding)

**Should promotion consider bean-ness?** `@EventListener`/`@Scheduled`/`@KafkaListener` are processed by post-processors that iterate beans *in the context*, so on a class the container never instantiates the annotation is inert. We are **not** filing this as a defect: a static analyzer can't determine bean-ness in general (XML config, `@Import`, `registerBean()`, abstract bases, test-context registration), and hard-gating on a stereotype would trade this for a worse false-negative class. The fixture's stereotype-free `ApplicationEvents` is a legitimate modelling choice, **no fixture change is being requested**. If you ever want to tighten it, the signal exists — `createSpringBeanCandidateAttacher` writes `classNode.properties.frameworkAnnotations` and the attacher already holds the owning class node — enough to *tier* the stamp rather than gate it. Your call.

---

### The four requested dimensions

| Dimension | Grade | Basis |
|---|---|---|
| **Performance** | **C+** | No quadratic behaviour (checked and refuted). But two unconditional whole-graph scans per JVM pass before any fact is known to exist, duplicating a superset computed three statements earlier and ignoring the lazy precedent in that same file; `createSpringAnnotationNameResolver` rebuilds its owned-type index on every call (repo-wide 9→11 per analysis — a worsening of a pre-existing baseline, not a new class of problem); unread `text`/`line` on the durable store. Every lever is a one-liner. |
| **Architectural fit** | **A−** | Both engines agree independently. Adapter+attacher shape matches the family; no logic branches on a language; hooks live; feature registration complete. Deducting only because the interface exclusion is enforced by two different mechanisms — Java at capture time on AST node type, Kotlin at attach time via a `HAS_METHOD` label lookup that is fail-**open** when the edge is absent. |
| **No overengineering** | **A** | Both engines agree. The generic parameter has exactly two real instantiations; the adapter interface matches the AOP precedent rather than inventing indirection; the feature is deliberately thinner than `aop.ts`, correctly scaled to a simpler problem. Naming the `3.0` constant is a net improvement over the unlabeled literal in ~10 other language modules. |
| **Reusability** | **C+** | Real reuse: `createSpringAnnotationNameResolver`, `nodeToCapture`, `makeScopeId`, the DI annotation-fact extractors, `isPackageSiblingVisibilityIncomplete`. Real duplication: `exactCallableOwnersByRange`, `classByMethod` (which also drops AOP's laziness), the memoized-resolver block (3rd flat copy when `di-metadata.ts` already generalized it), `simpleName` (5th copy). Fairly: four pre-existing local copies establish the precedent — this is drift with the grain, not a new anti-pattern. |

---

### Checked and fine — refuted suspicions

Validation is a feature; these were raised, probed, and defeated. Seven were refuted with **live tree-sitter grammar probes**:

- Nested/inner classes silently missed — **refuted** for named nested classes in both languages.
- Capture requires a Spring stereotype on the class (the `springDiClassNode` gate) — **refuted**; the name is misleading, it's a pure node-type test with no annotation predicate.
- `startLine - 1` / `endLine - 1` off-by-one — **refuted**; graph lines are 0-based, `Capture` ranges 1-based, and the conversion is byte-identical to the `aop.ts:239` precedent. Range collisions fail closed.
- Reason precedence misbehaves above 3.0 or on mixed kinds — **refuted**; all cases enumerated and correct. `astFrameworkMultiplier` and `astFrameworkReason` are always written together from one `frameworkHint`, so the 3.2 tier never occurs with an undefined reason.
- In-place mutation not idempotent across reruns — **refuted**; a second pass is a provable no-op.
- `useSiteTarget !== undefined` wrongly skips real handlers — **refuted**; plain `@Scheduled` produces no `use_site_target` child.
- Container annotations (`@Schedules`, `@KafkaListeners`, …) unreachable — **refuted**.
- One of the 21 FQNs names a wrong package — **refuted**; all hand-checked, including all three `SqsListener` generations.
- Also defeated: `\0` map-key collisions, duplicate-range keys (degrade to "no fallback", never "wrong owner"), cross-repo state leakage in a long-lived MCP process, and attachers running on repos with no JVM files.

---

### Test gaps

1. **No cold → save → warm parity test** — the suite passes `{}` for the cache, so every run is cold and F1 is *structurally invisible*. This is the gap that hides the blocker.
2. **`@MessageMapping` has zero effective coverage** — the only assertion touching it (`consumeOverWebSocket`) holds identically whether or not the `@MessageMapping` stamp fires, because `@Path` already set 3.0/`jaxrs-annotation` at parse time. It proves reason-preservation, not resolution.
3. **15–17 of the 21 catalog FQNs are never exercised** — a typo in `ApplicationModuleListener`, any `SqsListener` generation, `StreamListener`, `ServiceActivator`, `SubscribeMapping` or `TransactionalEventListener` is a silent permanent miss with a green suite. A table-driven fixture (one method per FQN asserting its own reason) turns every catalog typo into a test failure.
4. No Kotlin `object`/`companion object`/`enum class` fixture (F2) — notable since the AOP fixture in the same suite deliberately exercises both.
5. No wildcard-import or fully-qualified-annotation fixture; `isPackageVisibilityIncomplete` is passed but never true in any test.
6. Incremental behaviour is asserted structurally, not behaviourally — nothing exercises "add a `@Scheduled` on a warm rerun and confirm the stamp appears."
7. No zero-callee leaf handler, no `@Bean @ServiceActivator`, no implementor of `ListenerContract.kt` — so F4 and F6 are structurally untestable in the current fixture.

---

### CI

**CI Gate is red, and it is not this PR's fault.** `QUALITY: success · TESTS: failure · E2E: success`. Sole cause:

```
[callable-value-flow --check] FAIL
  - widening overhead 1.965 > budget 1.9
```

`bench/callable-value-flow/baselines.json` is **byte-identical** between base and head; the PR touches no callable-value-flow code; a local run measured **1.462**, 34.5% under budget. The failure is a 3.4% overage on a wall-clock gate — shared-runner timing variance. Verified independently by three lanes across both engines. It should still go green before merge, but it is not evidence against this PR.

**Sequencing:** #2886 and #2890 also append to `emitPostResolutionEdges` at `java/scope-resolver.ts:96`. Merge oldest first and rebase the others.

---

### Coverage limits

- Engine B ran at one-third capacity (rate limit); *Claude-only* findings lack second-engine confirmation.
- **Neither engine could execute the new integration test** — the worktree lacks a built `parse-worker.js`, so all 22 new assertions are unverified by execution here. CI's own shards report 17,392 passing.
- F5, F6, F7 and F8 are mechanism-verified but **magnitude-unmeasured** — no large Spring corpus was analyzed.
- Every finding lists its trigger, its verified `file:line`, and whether it was reproduced or code-read.

*Automated multi-tool digest (12 Claude lanes + 1 surviving Codex lane, merged by a synthesis pass and gated by an independent critic). Verify against the source before acting. Two of three Codex legs did not run, so cross-engine confirmation is weaker than a full tri-review.*

gitnexus/src/core/ingestion/languages/kotlin/captures.ts
ShiningXu: Fixed in 506e8463. Kotlin capture now follows the existing class, object, and companion-object selection and accepts enum_class_body. Enum members currently materialize as Function nodes, so the exact-range bridge now accepts class-owned Method or Function nodes while keeping interface and use-site exclusions. Unit, pipeline, process, and warm-cache coverage were added.

gitnexus/src/core/ingestion/languages/java/capture-side-channel.ts
ShiningXu: Fixed in 506e8463. SCHEMA_BUMP is now 59 because upstream main is still 53 and the current open #2899 head claims 54 through 58. The ledger and pin explicitly require whichever PR merges second to allocate a fresh value after rebasing. Added a cold parse to save cache and durable store to warm zero-worker parity test covering Java, Kotlin, and XXL-Job metadata.

gitnexus/src/core/ingestion/frameworks/spring/non-http-handlers.ts
ShiningXu: Fixed in 506e8463. The attacher returns before resolver or graph index construction when no facts exist, both graph indexes are lazy, and owner lookup now happens only after annotations resolve to a recognized handler kind. A dedicated unit test asserts that neither zero facts nor irrelevant Kotlin annotations invoke iterNodes, HAS_METHOD iteration, or getNode.

gitnexus/test/integration/spring-non-http-handlers-pipeline.test.ts
ShiningXu: Fixed in 506e8463. The integration helper now resolves by callable name plus file suffix over Method and Function nodes, and throws when the result count is zero or greater than one. Existing cross-language duplicate helper names can no longer be silently overwritten.

gitnexus/src/core/ingestion/frameworks/spring/non-http-handlers.ts
ShiningXu: Leaving this unchanged as suggested. Consolidating simple-name helpers is a repo-wide cleanup across the existing Spring modules and is better handled separately from this focused review fix.

gitnexus/src/core/ingestion/frameworks/spring/non-http-handlers.ts
ShiningXu: Fixed in 506e8463. The resolver now recognizes the official Spring Bean FQN as supporting evidence and suppresses only ServiceActivator when both resolve on the same method. Other handler kinds remain eligible. The fixture covers the official Bean factory negative case and a same-name fake Bean positive case.

gitnexus/src/core/ingestion/frameworks/spring/non-http-handlers.ts
ShiningXu: Not broadened in this PR. Precise support needs class-level listener evidence paired with method-level KafkaHandler or RabbitHandler evidence. Adding the two handler annotations alone would promote isolated handlers and create false positives. This remains a focused follow-up rather than an over-broad catalog entry.

gitnexus/src/core/ingestion/languages/kotlin/spring-non-http-handlers.ts
ShiningXu: Fixed in 506e8463. Kotlin non-HTTP facts now project annotations to only name and optional useSiteTarget, and the fact type is narrowed to SpringNonHttpHandlerAnnotationFact. Unit coverage asserts that object, companion, enum, and receiver-target facts contain no text or line fields.

ShiningXu:
Review follow-up pushed in 506e8463.

Addressed in this round:
- F1: parse-cache schema bump to 59, collision ledger, version pin, and cold-save-warm durable cache parity
- F2/F8: Kotlin object, companion object, and enum handlers plus compact annotation facts
- F4: official Bean plus ServiceActivator factory exclusion, with same-name fake Bean fail-closed coverage
- F7: zero-fact fast path and lazy graph-owner indexes, including irrelevant-annotation no-scan coverage
- F9: file-qualified callable lookup with duplicate detection

Kept out of this focused change:
- F3 needs precise class-listener plus method-handler pairing; adding KafkaHandler or RabbitHandler alone is too broad
- F5 and F6 change global Process ranking and leaf-entry semantics shared with HTTP routes
- F10 is an optional repo-wide helper consolidation

Validation:
- npm run build: passed
- npx tsc --noEmit: passed
- ESLint on changed TypeScript: 0 errors, 6 pre-existing non-null warnings in Kotlin captures
- Focused regression set: 98 passed
- Full npm test: 16,955 passed, 101 skipped, 10 failures in four unrelated environment or existing-path files; no changed Spring, Kotlin, cache, or fixture test failed
- GitNexus detect-changes: LOW risk, 0 affected processes

Merge-time cache check: upstream main remains at 53 and open #2899 remains at 58, so 59 is currently collision-free. If #2899 merges first, this PR must allocate a fresh value after rebasing.

ShiningXu:
Synced the current upstream main commit c6b24162 through merge commit cd1d3168.

Post-sync validation:
- npm run build: passed
- npx tsc --noEmit: passed
- Spring non-HTTP handler focused set: 75 passed
- Upstream Kotlin import-resolution regression set: 23 passed
- GitNexus detect-changes for the merged upstream delta: LOW risk, 0 affected processes

The branch now contains upstream main and keeps SCHEMA_BUMP at 59.

magyargergo:
@ShiningXu can you please resolve the merge conflicts?

Issue #2895, opened 2026-08-08
CI: a failing bench --check aborts the benchmarks job, so every guard after it silently never runs


magyargergo:
## Summary

Every step in the `tests / benchmarks (GITNEXUS_BENCH)` job is fail-fast, so the first `--check` that fails aborts the job and **every guard declared after it never executes**. Those later guards then report as `skipped`, which reads identically to "nothing to do" in the checks UI — not as "this was never verified".

A skipped guard is not a passing guard, and right now nothing distinguishes them.

## Evidence

`.github/workflows/ci-tests.yml`, the benchmarks job spans lines 462-593. Grepping that range for `if:` or `continue-on-error:` returns nothing — there is no `if: always()` anywhere in the job (the only two in the file are at lines 72 and 134, both in other jobs).

The three tail steps:

| line | step |
|---|---|
| 558 | Emit-persistence throughput / byte-identity guards (#2203) |
| 567 | Streaming PDG-emit byte-identity / bounded-RSS guards (#2202) |
| 576 | Cross-language pipeline benchmarks (GITNEXUS_BENCH, serial) |

## How this actually played out

On the branch for #2856, the emit-persistence baseline was stale for the life of the PR. Result, audited across every benchmark-job run on that branch (jobs `92659525436`, `92716925200`, `92725814265`, `92734685285`, `92739585125`, `92817491828`, `92884400117`, `92923296393`, `92937498727`, `92941732433`, `92969156952`, `93062865688`, `93074242346`):

- the benchmarks job **succeeded zero times**
- step 12 executed **zero times**
- step 13 executed **zero times**

So the streaming PDG-emit byte-identity guard and the eight-language pipeline benchmarks had no result at all for the entire PR, while the checks UI showed them as skipped. Two full reviews of that PR read the checks summary and did not catch it. Both guards in fact pass — I ran them locally after fixing the baseline (`[stream-pdg-emit --check] PASS`; pipeline benchmarks 8 files / 13 tests) — but that was luck, not verification.

The failure mode is worst exactly when it matters most: a change broad enough to trip an early guard is a change you most want the later ones to have run against.

## Suggested fix

Roughly in order of effort:

1. **`if: always()` on each `--check` step after the first**, so one stale baseline reports one red step instead of masking three. Cheapest change, keeps the current single-job shape.
[...]

PR #2898, opened 2026-08-08
perf(import-target): index the workspace once per run for go/csharp/dart/ruby

### LINKED ISSUE #2877: perf(go): resolveGoImportTarget scans the whole workspace on every import
Follow-up from #2872, which fixed the same shape in Kotlin (and #1918 in Python, #2788 in C++).

`resolveGoImportTarget` walks all of `allFilePaths` on every import, with nothing memoized:

- `findRootPackageFiles` (`gitnexus/src/core/ingestion/languages/go/import-target.ts`) iterates the whole set, normalizing each path with `raw.replace(/\\/g, '/')`.
- `findAllFilesInPkgDir` does the same, plus an `includes` and an `indexOf` per entry.
- The GOPATH fallback calls `findAllFilesInPkgDir` once per path segment except the last: `for (let i = 0; i < parts.length - 1; i++)`. For a typical `github.com/org/repo/pkg` import that is three full passes; for a deeper path, more.

So a single unresolved import costs several full workspace scans, and resolution overall is O(imports × files).

This is likely the worst of the family, because most Go imports are external (stdlib, third-party modules) and therefore always run the whole cascade to completion before returning null — the fast path never triggers. Measured during review of #2872: scaling ratio **4.00**, ~370 µs per import at 1600 files.

### What it looks like when it bites

It does not present as a hot loop. `analyze` sits at exactly 1.00 core with a flat heap and no output, because every allocation is a short-lived string. Small repositories hide it entirely.

### Suggested fix

The same shape as `getKotlinFileIndex` (#2872) and `getPythonFileIndex` (#1918): build the lookup tables once per `allFilePaths` Set, memoized on the Set's identity in a `WeakMap`, so each query is O(1).

Two things that turned out to matter in the Kotlin change and would apply here:

1. The adapter must pass the Set **through**. A defensive `new Set(allFilePaths)` hands a fresh WeakMap key per call and rebuilds the index every import, restoring the original behaviour — Python shipped exactly that bug and it was caught in review (#1918 P1). Worth an integration test asserting the index builds once across many imports.
2. Go's resolution has implicit tie-breaks (`findRootPackageFiles` sorts, `findAllFilesInPkgDir` does not) and the GOPATH fallback returns on the first non-empty result. Those need to be reproduced exactly, or edges move in every Go repository. A differential harness against the current implementation, plus a fingerprint bench in the `benchmarks (GITNEXUS_BENCH)` job, is what made that checkable in #2872 — `bench/kotlin-import-target/` is a working template.

### LINKED ISSUE #2878: perf(csharp): the no-csproj path bypasses the memoized workspace index
Follow-up from #2872.

`csharp/import-target.ts` already has the right machinery — `getWorkspaceFileIndex` is memoized on the `allFilePaths` Set identity via a `WeakMap`, added for #1881. But it is only reached inside the `csharpConfigs.length > 0` branch, and that branch returns unconditionally. When no `.csproj` is found, resolution falls through to:

- `resolveDirectMatch(ctx.allFilePaths, pathLike)` — a raw `for (const raw of allFilePaths)` scan;
- `resolveByProgressiveStripping(ctx.allFilePaths, pathLike)` — nested loops over `allFilePaths` via `findDirectChild`.

Both take the raw Set and scan past the index that is sitting right there. An unresolved four-segment `using` can trigger up to eight full workspace passes.

Measured during review of #2872: scaling ratio **3.43**, ~359 µs per import at 1600 files.

There is a second-order effect worth flagging: the #1881 namespace-evidence gate fails open when the namespace scan is truncated, and truncation is most likely on exactly the large repositories where this cost hurts most. So the slow path and the degraded-accuracy path coincide.

### Suggested fix

Route the no-csproj path through `getWorkspaceFileIndex` as well, rather than adding a second index. The index already carries `normalized`, `all` and a `SuffixIndex`, which is most of what `resolveDirectMatch` recomputes per call.

Care needed on tie-breaks: `resolveDirectMatch` returns the first exact match but only remembers the first *suffix* match, so an exact match found later still wins. That ordering is expressed only through Set-iteration order and has to be preserved, or edges move. In #2872 a differential harness against the pre-change implementation plus a fingerprint bench in the `benchmarks (GITNEXUS_BENCH)` job made that verifiable; `bench/kotlin-import-target/` is a template.

### LINKED ISSUE #2879: perf(dart): resolveDartImportTarget scans the workspace once per candidate path
Follow-up from #2872.

`dart/import-target.ts` scans all of `allFilePaths` per import, unmemoized, on both of its paths:

- the `package:` branch builds two candidate paths (`lib/<rel>` and bare `<rel>`) and runs a full `for (const fp of allFilePaths)` loop for each. For an external package both miss, so both loops always run to completion;
- `resolveRelative` tries an O(1) `allFilePaths.has(target)` first, but its suffix fallback is another full scan.

Measured during review of #2872: scaling ratio **4.05**, ~89 µs per import at 1600 files. The per-import constant is lower than Go or C# because the loop body is cheap, but the asymptotic shape is the same.

### Suggested fix

Same pattern as `getKotlinFileIndex` (#2872) / `getPythonFileIndex` (#1918): one index per `allFilePaths` Set, memoized on the Set's identity in a `WeakMap`.

Dart's rules are simpler than Kotlin's — the match is `fp === candidate || fp.endsWith('/' + candidate)`, so a stem/suffix map covers both — but the first-match-in-iteration-order tie-break still needs preserving, and the two candidates must keep their priority (`lib/<rel>` before bare `<rel>`). A differential harness against the current implementation plus a fingerprint bench in `benchmarks (GITNEXUS_BENCH)` is what made that checkable in #2872; `bench/kotlin-import-target/` is a template.

Also worth keeping in mind from that review: the adapter must pass the Set through without copying, or the `WeakMap` key changes per call and the index rebuilds on every import (#1918 P1).

### LINKED ISSUE #2880: perf(ruby): resolveBare rebuilds the whole suffix index on every require
Follow-up from #2872.

`resolveBare` in `ruby/import-target.ts` does this per bare `require`:

```ts
const normalizedFileList = [...allFilePaths].map((f) => f.replace(/\\/g, '/'));
const allFileList = [...allFilePaths];
const index = buildSuffixIndex(normalizedFileList, allFileList);
```

That is two full array materializations from the Set plus a complete `buildSuffixIndex` — which itself walks every file and inserts one entry per path suffix into `exactMap`, one into `lowerMap`, and one into `dirMap` — and it is thrown away as soon as the call returns.

This is a worse constant factor than the plain scans in the sibling resolvers: Go and Dart re-scan, Ruby re-*builds*. Every `require` in the repository pays for indexing every Ruby file in the repository.

`buildSuffixIndex` is shared (`import-resolvers/utils.ts`) and used by C#, Vue and TypeScript as well, so it is worth checking whether any of those call sites are also inside a per-import path rather than hoisted.

### Suggested fix

Memoize on the `allFilePaths` Set identity, the same way `csharp/import-target.ts` already wraps `buildSuffixIndex` in `getWorkspaceFileIndex`, or as `getKotlinFileIndex` (#2872) and `getPythonFileIndex` (#1918) do. Since `buildSuffixIndex` is deterministic in its inputs, this should be a pure hoist with no behaviour change — but the same caveats apply:

- the adapter must pass the Set through without copying, or the `WeakMap` key changes per call and nothing is saved (#1918 P1);
- worth an integration test asserting the index builds once across many `require`s, and a fingerprint bench so a future per-import rebuild does not slip back in. `bench/kotlin-import-target/` is a template.

magyargergo:
Closes #2877, closes #2878, closes #2879, closes #2880.

Four import-target resolvers answered their lookups with a full `allFilePaths` scan per import, so import resolution cost **O(imports × files)**. Each now reads an index memoized on the `allFilePaths` Set identity — the shape `getPythonFileIndex` (#1918) and C#'s own `getWorkspaceFileIndex` (#1881) already used.

| | before | now |
|---|---|---|
| **go** #2877 | `findRootPackageFiles` / `findAllFilesInPkgDir`, the latter once per path segment on the GOPATH fallback. Most Go imports are external, so the whole cascade ran to completion before returning `null` — the fast path never fired. | one package-dir index per run |
| **csharp** #2878 | the no-csproj leg took the raw Set past the memoized index the csproj leg was already using — up to eight passes for a four-segment `using` | both legs read the same two per-run indexes |
| **dart** #2879 | one scan per candidate path; for an external package both candidates miss, so both always ran to completion | one basename index per run |
| **ruby** #2880 | a complete `buildSuffixIndex` **rebuilt and discarded per `require`** — every require paid to index every file in the repo | shared workspace index, hoisted |

## Measured

`bench/import-target/measure.mjs`, 8 imports/file, ~5 in 8 imports external (they run every leg to completion — a mostly-resolving corpus measures the wrong half of the function). The two columns are at different scales: the "after" column is the shipped gate at 400 → 1600 files, the "before" column was measured against the pre-change implementations on the same corpus **shape** at 150 → 600 files (recorded in `_floor` in `baselines.json`). `scaling_ratio` divides the file count out, so it is scale-invariant and the two are comparable:

| | scaling ratio before (150 → 600) | after (400 → 1600) |
|---|---|---|
| go | 3.36 | 0.984 |
| csharp | 4.10 | 1.138 |
| dart | 3.32 | 0.787 |
| ruby | 3.87 | 1.123 |

1.0 is linear; ~4 is quadratic at a 4× scale gap. The "before" column does **not** reproduce the exact ratios reported in the issues (4.00 / 3.43 / 4.05) — different corpora — but the two sets independently confirm the same quadratic behaviour. Ruby's absolute cost at 1600 files fell from *unfinished after 2 minutes* to 86 ms.

## Two shared modules

- `import-resolvers/workspace-file-index.ts` — normalized list + `SuffixIndex` + a normalized→raw map (C#, Ruby). C#'s local copy moved here; the new `normToRaw` answers "is there a file whose **whole** path is X", which `index.get` cannot, because the suffix map conflates a whole-path hit with a `…/X` suffix hit.
- `import-resolvers/package-dir-index.ts` — "which files live directly inside a directory ending with `<path>`" (Go, C#). Candidates are bucketed by the directory's **last segment** rather than by indexing every directory suffix: a suffix map costs O(files × depth) entries, which is the memory this codebase runs out of at kernel scale (#2649).

## Behaviour is unchanged, including the invisible tie-breaks

These are expressed only through Set-iteration order and `indexOf` positions, so nothing in the type system or the existing tests could see them move:

- Go's root leg is **sorted**; its package leg is **not** — it keeps Set order, and multi-directory matches are merged back onto it.
- Go and C# both take the **first** occurrence of `/<segment>/`, so `a/pkg/b/pkg/x.go` does **not** answer `pkg`. Dropping that looks like a cleanup and moves edges in every repo that nests a directory name inside itself (`internal/…/internal`, `Models/…/Models`).
- C#'s `resolveDirectMatch` lets a whole-path match win over a suffix match found **earlier** in iteration order; `resolveByProgressiveStripping` takes whichever comes first.
- Dart tries `lib/<rel>` fully before bare `<rel>`, and matches **raw** paths — normalizing backslashes here would start resolving imports that previously returned `null`.

Verified two independent ways:

1. **`test/unit/scope-resolution/import-target-index-parity.test.ts`** keeps verbatim copies of the pre-change implementations and diffs against them over a deterministic corpus (40 layouts × every target shape, per language) plus hand-built layouts for each tie-break above. Six mutations of the new code were each confirmed to fail it — dropping the first-occurrence rule, dropping the order merge, sorting Go's package leg, dropping C#'s exact-wins lookup, normalizing Dart's paths, and flipping Dart's candidate priority.
2. **Byte-identical fingerprints.** The bench corpus was run against the pre-change resolvers and this branch at both 400 and 1600 files: all seven sha256s match exactly.

The same file also asserts the index is built **once per file set**, by counting traversals of the Set. Go 1, Dart 1, Ruby 1, C# 2.

That arm catches a rescan reintroduced *inside* a resolver, but it calls the resolvers directly, so it does **not** cross the adapter boundary — which is exactly where #1918 review P1 lived (a defensive `new Set(allFilePaths)` in an adapter hands a fresh `WeakMap` key per call and silently restores the quadratic behaviour). `test/integration/{go,csharp,dart,ruby}-import-index-reuse.test.ts` cover that layer, resolving through `<lang>ScopeResolver.resolveImportTarget` the way Kotlin's and Python's existing guards do. Each was verified to fail under the copy it exists to catch.

## CI

New always-on step in `benchmarks (GITNEXUS_BENCH)`. The final `--check` gates eight arms per language, not two: the output fingerprint **at every scale** (small, large, deep, collide), the scaling budget (1.8), the per-language `depth_ratio` budget, the per-language `collide_scaling_ratio` budget, the absolute `small_ms_ceiling`, the absolute retained-heap ceiling, `deep.resolved === small.resolved`, and `deep.fingerprint !== small.fingerprint` — **plus the corpus shape at every scale** (files / imports / resolved / distinct outcomes, asserted exactly), because a fingerprint alone cannot tell a legitimate resolution change from a corpus quietly shrunk below the size at which the timing arms can see anything.

The step is placed **last** in the `benchmarks` job on purpose until #2895 lands: a failing `run:` step aborts every step after it, so the newest and least-proven gate must not be in a position to mask the eight established guards behind it.

## Also checked, no change needed

#2880 asks whether the other `buildSuffixIndex` call sites are inside a per-import path. TypeScript (`typescript/scope-resolver.ts`) and Vue (`vue/import-target.ts`) already hoist theirs behind a pass cache keyed on the Set identity; `group/extractors/include-extractor.ts` builds one per extraction run, not per import.

## Out of scope, worth a follow-up

`php/import-target.ts` re-materializes `[...allFiles].map(…)` per call at two sites (lines 185 and 220) — the same shape as #2880, not covered by any of these four issues.

## Test status

`tsc` clean, lint clean. 27/27 in the new parity file; 243/243 across the go/csharp/dart/ruby scope-resolution unit suites; 592 passing in the integration resolver suites. Two integration files (`resolvers/ruby.test.ts` and one sibling) failed locally with `Worker pool failed to start: Replacement worker did not report ready within 5000ms` — the #1741 host-load flake, in `beforeAll` before any resolution runs. Nothing in this PR touches worker startup. CI is the arbiter.

---

## Update: merged `main`, and Kotlin folded in

`main` now carries #2872, which did the same hoist for Kotlin. Merge conflict was one benchmark step both branches add at the same place in the `benchmarks` job — kept both. (The new step was also renamed to `(#2877/#2878/#2879/#2880/#2872)`: a space before `#` starts a YAML comment inside a plain scalar, so the original spelling truncated the step name to `Import-target resolution guards (#2877`.)

Kotlin now runs in the shared corpus, and #2872's review surfaced two arms a scaling ratio alone cannot carry, so all five languages get them:

- **`depth_ratio`** — deep arm at a **fixed** file count with ~6× the path components. `scaling_ratio` divides the file count out, so it is scale-invariant and structurally cannot see cost growing with path *depth* instead — and `buildSuffixIndex` (C#, Ruby) and Kotlin's `suffixByStem` each emit one entry per component. Measured go 1.06, dart 0.88 (depth-free indexes), ruby 1.48, kotlin 2.20, csharp 3.45. Budget is per language: one global budget would have to sit at 5.0 and would let Dart go 0.88 → 4.9 unnoticed.
- **`small_ms_ceiling`** — absolute, at 4× the measured arm, because a constant-factor regression that grows both scale arms equally passes every ratio.
- The deep arm must resolve **exactly** what the small arm resolves — padding was supposed to change depth and nothing else; a deep arm that stopped resolving would be timing the null path.

The five fingerprints are unchanged by that commit, verified against the previous baseline before rewriting it — adding the Kotlin arm and the deep scale did not perturb the four languages' output.

Kotlin also joins the Set-iteration counter in the parity test. Its own guard (`kotlin-import-index-reuse.test.ts`) counts index **builds**, which a scan added beside a reused index does not move.

### One measured blind spot, documented rather than papered over

A full workspace scan reintroduced on **1-in-32 imports passes every timing arm** here — dart scored 1.458 against the 1.8 scaling budget and 1.736 ms against a 4 ms ceiling. At 1-in-8 the scaling arm catches it (2.414). What *narrows* the gap is not a timing gate: the traversal counter reads **14 instead of 1** for that same mutation, deterministically, for all five languages. It does not close it — the counter watches the Set, while the resolvers hold materialized arrays of the same list (`WorkspaceFileIndex.normalized`/`.all`, Dart's basename buckets, `PackageDirIndex.filesByDir`), so a scan over one of those is still invisible to both gates. Tightening the ceilings toward the noise floor to chase that case would only buy flaky CI. Recorded in `_blind_spot` in `baselines.json` so nobody has to rediscover it.

`bench/kotlin-import-target/` stays. It fingerprints both file-set iteration orders and probes the four-tier cascade shape by shape; this corpus does neither, so it is a second corpus, not a replacement.

---

## Update: review round

Addressed the findings from the two-engine review (Claude swarm + ce personas, Codex `gpt-5.6-sol`, review [#4889364193](https://github.com/abhigyanpatwari/GitNexus/pull/2898#pullrequestreview-4889364193)). The resolver logic itself came through clean — 6,402,319 old-vs-new differential comparisons with 0 divergences, and **all five output fingerprints are byte-identical after every change below**, which is the proof that none of it altered behaviour.

**One real cost regression, now inverted.** `filesDirectlyInPkgDir` re-spread its accumulator per matching directory, O(files × dirs²). Where many services share a package directory (`svcN/internal/pkg`, which Go's GOPATH cascade queries by two-segment tail) the index was **13.4× slower than the scan it replaced** at k=1600. Appending once and sorting once makes it **4.1× faster** there; output byte-identical at every k. The same edit is what the simplicity lane independently asked for.

**Three holes in this PR's own gate.**
- `--check` computed three fingerprints per language and compared one. `DEEP_PAD = 16 → 0` deleted the whole depth arm and still printed PASS. Every scale is asserted now, plus `deep.fingerprint !== small.fingerprint` so the padding's *effect* is pinned rather than just its output.
- The corpus minted unique directory names, so max bucket cardinality was **1** — and that is the only non-constant term the index has, so the `dirCount > 1` merge branch had never executed in any arm. The new `collide` arm reaches 9,269 multi-directory merges per run, up to 34 at once. Go/C#/Dart legitimately score above the linear budget on shared-leaf layouts and get their own; Ruby and Kotlin stay at 1.8 because their keyed maps are collision-immune and that immunity is the assertion.
- No arm measured memory. A retained-heap arm on the `bench/cfg` pattern (including its loud failure when `--expose-gc` is missing, rather than a silent skip) measures **csharp 73.62 MiB / ruby 55.26 MiB at 32k files**. Its ceiling is **1.5×, deliberately not the 4× the timing arms use** — the measurement is byte-stable to 0.00085% across processes, so 4× would discard the gate. `_arms_note` records why, so it does not get "harmonised" back.

**The reuse guard did not guard the layer its own docstring named.** Inserting `new Set(allFilePaths)` into all four `<lang>/scope-resolver.ts` adapters left the parity test 28/28 green and `--check` PASS. Kotlin and Python already had adapter-level integration guards; go/csharp/dart/ruby now do too, each proven to fail under that exact mutation. `CountingSet` also now counts `forEach`/`values`/`keys`/`entries`, not just `[Symbol.iterator]` — a rescan spelled `allFilePaths.forEach(…)` was previously invisible.

**`depth_ratio`, added by this PR, flaked ~1-in-20** (go peaked 1.748, dart 2.043, against a 1.6 budget) — both are ratios of two sub-3 ms minima. Fixed at the estimator, not the threshold: `REPS` 5 → 15, matching `bench/cfg`, `schema-pairs` and `callable-value-flow`. 22/22 PASS, every arm now at 70–78% of its budget with a ≤1.26× swing, **no budget widened**; the measured distributions are in `_arms_note` so the headroom is visibly earned.

**Four fixtures pinned the guard next door** to the one their title named — the Dart "matched RAW" case asserted `null === null`, C# and Ruby had no backslash target at all, `normToRaw`'s first-wins rule had no normalization twin, and the Go nested-package case was decided by the `endsWith` half. Each now kills the mutation it exists for.

Also corrected three copies of the claim that the parity test *closes* the 1-in-32 blind spot (it narrows it), and `_floor`'s claim that its ratios "match" the issues' (different corpora, both quadratic).

### Still out of scope, with numbers attached

- `languages/php/import-target.ts:185/220` — materializes the workspace twice per import and passes no index: **95 ms per import at 20k files**. The last language without a per-pass memo; `getWorkspaceFileIndex` makes it a small change.
- `import-resolvers/csharp.ts:81` — per-import linear scan on the **csproj** leg, ~0.95 ms/import at 50k `.cs`. Bounded: the root-namespace gate at `:32-42` means BCL/NuGet usings never reach it.
- `buildSuffixIndex`'s `dirMap` is **15.7% (C#) / 20.3% (Ruby)** of retained bytes, and Ruby, TypeScript, Vue and the include-extractor never read it — building it lazily is behaviour-identical and worth ~34 MiB at 100k Ruby files. Left alone here rather than widening this PR into shared `utils.ts`.


koriyoshi2041 (approved):
Verified the index hoist and its guardrails at `478390c8`.

The shared package-directory index preserves the two easy-to-lose contracts from the former scans: first-occurrence matching for repeated directory names and global Set iteration order when multiple matching directories interleave. The C# whole-path-vs-suffix precedence also remains explicit rather than being collapsed into the suffix map. The deterministic Set-iteration assertions complement the timing benchmark well; they catch sparse reintroduced scans without tightening CI thresholds into noise.

Local verification on Node 26.5.0:

- `npx vitest run test/unit/scope-resolution/import-target-index-parity.test.ts` — 28/28 passed
- `node --import tsx bench/import-target/measure.mjs --check` — passed for Go, C#, Dart, Ruby, and Kotlin
- `npx tsc --noEmit` — passed
- `git diff --check` — passed

Hosted CI is also fully green at this head.

gitnexus/test/unit/scope-resolution/import-target-index-parity.test.ts
```diff
@@ -0,0 +1,760 @@
...
+
+function countingCorpus(seed: number, extension: string): CountingSet {
+  return new CountingSet(corpus(seed, extension, 200));
+}
+
+describe('import-target index hoist — built once per file set, not once per import', () => {
+  it('go builds one index for many imports (#2877)', () => {
+    const files = countingCorpus(1, '.go');
+    for (let i = 0; i < 200; i++) {
+      resolveGoImportTarget(`github.com/org/repo${i}/pkg`, 'main.go', files, {
```
magyargergo, line 761: **[P1]** The reuse arms call the resolver directly, so they cannot see the regression this file's own docblock says they guard.

`:27-33` states the guard is for *"an adapter that copies the Set (`new Set(allFilePaths)`)"*, and `workspace-file-index.ts:13-15` points production readers here for that guarantee. But production reaches these resolvers through `<lang>ScopeResolver.resolveImportTarget`, and this arm calls `resolveGoImportTarget` directly. Inserting `new Set(allFilePaths)` at `go/scope-resolver.ts:31`, `csharp:35`, `dart:189` and `ruby:268` leaves **28/28 green and `measure.mjs --check` PASS in all four cases** — `measure.mjs:220-233` bypasses the adapters too. Separately, the docstring at `:683-684` ("the per-import lookups afterwards only do `Map.get` and `Set.has`") no longer holds now that the resolvers carry materialized arrays: a 1-in-32 fallback scan over `ws.normalized` in `resolveBare` also passes 28/28 **and** `--check` with a baseline-matching fingerprint. The repo already ships the right shape for the two preceding languages — `test/integration/kotlin-import-index-reuse.test.ts:1-13` and its Python sibling open by naming exactly this distinction.

Fix: drive the five `CountingSet` arms through `<lang>ScopeResolver.resolveImportTarget` (the `CountingSet` flows through unchanged; verified all four adapter-copy mutations then fail), keeping the direct-call arms so a failure localizes. For the array half, count index BUILDS in production the way `kotlin/index-stats.ts` does, or Proxy-count element reads on `ws.normalized`/`ws.all`. `[reproduced]`

magyargergo (commented):
## Two-engine review of #2898 — synthesis

**Engines.** Claude ran the GitNexus PR swarm (6 lanes: PR-facts, branch-hygiene, risk-architect, test/CI, security-boundary, docs/DoD) plus `ce-code-review` with 8 personas (correctness, performance, adversarial, testing, maintainability, simplicity, api-contract, project-standards). Codex `gpt-5.6-sol` / xhigh ran the same swarm in Solo mode, its own `ce-code-review`, and its native adversarial pass. On top of both, the coordinator ran and personally verified **17 controls**, several of which demote or refute lane claims. This is **two engines**, not five independent confirmations.

**Codex leg liveness — all three live.** The first attempt at all three died on the org's 500k-TPM rate limit when run concurrently; re-run sequentially, all three reported.
- `codex-swarm` (7-lane Solo) — verdict `not production-ready`, findings F1–F4.
- `codex-adversarial` — verdict `needs-attention`, 1 high + 2 medium.
- `codex-ce` — verdict `Ready with fixes`, 4 findings + 2 pre-existing. It landed after the merge was drafted, so it is folded in below as corroboration rather than as a source of new findings; note its bench never executed (`bwrap: No permissions to create new namespace`), so its cost claims are code-read, not measured.

**Codex's own three legs did not agree on the verdict** — its swarm says `not production-ready`, its ce leg says `Ready with fixes`, its adversarial leg says `needs-attention`. All three are reacting to the same facts; they weigh the gate holes differently. The verdict below is the coordinator's, and the reasoning is stated so you can disagree with it explicitly.

---

## What is solid, and it is a lot

The resolver logic is **not** in question. This was attacked hard and held.

- **ce-correctness** built a differential harness against verbatim `git show c6b24162d:` copies of every replaced function and ran **6,402,319 old-vs-new comparisons with 0 divergences**, over 37 hand-built adversarial file sets plus 700 generated ones. It validated the harness with **18 control mutations, 15 detected**; all 3 survivors are explained (M4 and M16 are true equivalences with proofs; M8 was a corpus gap the lane closed with a hand-built normalization-twin probe).
- **Codex independently re-derived the first-`indexOf` equivalence** and reached the same verdict from the other engine: *"The length-guarded predicate exactly preserves the old first-`indexOf` rule, including repeated directory names."* ce-adversarial re-derived it a third time algebraically rather than trusting the header comment. The length guard at `package-dir-index.ts:117` and the `indexOf === length − length` form (not `endsWith`) are both load-bearing and both proven so by mutation (4,166 and 9,592 divergences).
- **The parity test's "verbatim pre-change copies" are genuinely verbatim.** Two lanes diffed all ten `legacy*` functions line-by-line against BASE. No drift; exactly two intentional deltas, both disclosed (`modulePath !== undefined` vs `goModule != null`, unreachable for the harness's inputs; and the hardcoded C# `evidence = undefined`, reported below as P3). The parity claim is not circular.
- **The win is real and understated.** ce-performance re-measured BASE against HEAD on the same corpus: **64–79× at 400 files, 271–311× at 1600**, with byte-identical sha256 fingerprints at every scale. The PR body's claimed before-ratios (3.36/4.10/3.32/3.87) reproduced at 3.965/3.964/4.087/3.910.
- **Aliasing is clean.** Four lanes independently confirmed no live memo bucket escapes: `filesDirectlyInPkgDir` always returns through `.map()`, `findRootPackageFiles` copies before sorting, every downstream consumer is read-only over the shared arrays (verified at runtime across 1,800 resolutions). Codex agrees: *"Live memo buckets are not exposed to callers."*
- **project-standards filed zero findings** after reading AGENTS.md, CLAUDE.md, ARCHITECTURE.md, CONTRIBUTING.md, GUARDRAILS.md, TESTING.md and DoD.md in full, and correctly scoped the "shared code must not name languages" rule to the call/inheritance pipeline — ARCHITECTURE.md:400 explicitly sanctions language-named files in `import-resolvers/`. api-contract confirmed all four public resolver signatures are byte-identical to BASE and that declaration emit is unaffected.
- **The design choices were checked and are right.** Reusing `buildSuffixIndex` for Dart would have reintroduced the O(files × depth) shape #2649 is about, and would move backslash-path behaviour — the PR's separate basename index is the correct call. Kotlin's index correctly stays separate (four documented divergences). `normToRaw` cannot be replaced by `allFilePaths.has()` without breaking Windows checkouts.

---

## Headline findings

### 1. P1 — The reuse guard sits one layer below the hazard both it and the production doc comment describe

**Engines/lanes:** Claude ce-testing, ce-maintainability, ce-adversarial · Codex adversarial · **coordinator CONTROL 15 (reproduced)**. Strong, both engines. **Blocks merge: yes.** (Dissent noted: ce-maintainability rated the same gap P2, not P1. The case for P1 is in "why this is P1" below — if you read it as P2, say so and it becomes a follow-up rather than a gate.)

`workspace-file-index.ts:10-15` states the hazard is *"a defensive `new Set(allFilePaths)` in an **adapter**"* and names `import-target-index-parity.test.ts` as the guard. `baselines.json:170` names the same test as the control that closes the bench's own measured blind spot. Neither claim holds:

- **Adapter copy.** The test imports the resolvers directly (`:44-51`); production reaches them through `<lang>ScopeResolver.resolveImportTarget`. Inserting `new Set(allFilePaths)` at `go/scope-resolver.ts:31`, `csharp:35`, `dart:189`, `ruby:268` leaves **28/28 green and `--check` PASS in all four cases**. `measure.mjs:220-233` calls the resolvers directly too, so the bench is equally blind. Codex reached this independently.
- **Array scan.** The docstring at `:683-684` claims *"the per-import lookups afterwards only do `Map.get` and `Set.has`. So the scan count IS the number of index builds."* That is no longer true — the resolvers now hold materialized arrays of the same file list. ce-adversarial patched `ruby/import-target.ts::resolveBare` with a 1-in-32 fallback scan over `ws.normalized` (the PR's own documented blind-spot ratio, on its own language) and got `28 passed (28)` **and** `[import-target --check] PASS` with a baseline-matching fingerprint. That mutation restores exactly the `O(imports × files)` behaviour the PR exists to remove.

**Trigger:** a future contributor adds a defensive/normalizing `new Set(...)` at any of the four adapter boundaries — the single most natural edit there, and the one this repo has already made once (#1918) — or reintroduces a scan over the index's own arrays. Every gate stays green.

**Why this is P1 and not a nit:** the repo already ships the correct shape for the two languages that came before. `test/integration/kotlin-import-index-reuse.test.ts:1-13` opens by stating the distinction verbatim, and `python-import-index-reuse.test.ts` does the same. #2898 adds no equivalent for go/csharp/dart/ruby. `[reproduced]`

**Fix:** drive the five `CountingSet` arms through `<lang>ScopeResolver.resolveImportTarget` instead of the raw functions (the `CountingSet` flows through unchanged — verified that A1–A4 all fail under that shape), or mirror the two existing integration guards. Keep the direct-call arms as well; they localize a failure to the resolver vs the adapter. For the array-scan half, either count index BUILDS in production (the `index-stats.ts` pattern already established for Python/Kotlin) or wrap `ws.normalized`/`ws.all` in a counting Proxy in the test.

---

### 2. P2 — `filesDirectlyInPkgDir` merges matching directories in O(k²·m) and becomes slower than the scan it replaced

**Engines/lanes:** Claude ce-performance (F1), ce-adversarial (F1), ce-simplicity (item 4, found independently as a simplification) · Codex ce (#3, same fix proposed verbatim) · **coordinator CONTROL 12 (reproduced)**. Strong, both engines. **Blocks merge: no** — but it is the cheapest fix on this list and the only one that is also a simplification.

`package-dir-index.ts:133` re-spreads the whole accumulator once per matching directory:

```ts
merged = dirCount === 1 ? files : [...(merged ?? []), ...files];
```

For `k` matching dirs of `m` files that is `m·k²/2` element copies, per import. The coordinator measured it against a verbatim copy of the pre-change Go scan, output byte-identical at every `k`:

| k | files | new ms | old ms | ratio |
|---|---|---|---|---|
| 200 | 800 | 0.232 | 0.070 | 3.30 |
| 800 | 3 200 | 2.032 | 0.263 | 7.72 |
| 1 600 | 6 400 | 9.460 | 0.494 | **19.14** |

`new_ms` quadruples per doubling of `k`; `old_ms` is linear. Two lanes reached this from different corpora before the coordinator did — ce-performance measured **7.13× end-to-end** through `resolveGoImportTarget` on a 2,000-service monorepo (crossover at K≈300), ce-adversarial **68.9×** at k=4,000.

**Trigger:** Go's GOPATH cascade tries every suffix of the import path down to two segments, so any import whose two-segment tail matches a directory tail replicated many times (`svcN/internal/models`, `svcN/pkg/util`) pays the full quadratic — an ordinary monorepo shape. C#'s `firstFileDirectlyInPkgDir` is **not** affected; it has no merge and stays O(k). `[reproduced]`

**Fix is one line and is simpler than what is there** — push into a single array, sort once. ce-performance validated a linear version as byte-identical at every `k` and **68.5× faster at k=2000**, and 9.5× faster at k=250, which is already inside the region where the new code wins. This is the rare case where the perf fix and the simplicity lane's recommendation are the same edit.

---

### 3. P2 — `--check` never asserts the `small` or `deep` fingerprints; the depth arm can be deleted with CI green

**Engines/lanes:** Claude ce-adversarial (F3/F8) · Codex swarm (F2, rated "blocks merge") · Codex adversarial (medium) · Codex ce (#1) · **coordinator CONTROL 14 (reproduced)**. Strong, both engines, all three Codex legs. **Blocks merge: no** — the fix is one word and the values already exist on both sides.

`measure.mjs:311` exports `fingerprint: scales.large.fingerprint`; `:325` compares only that; the per-scale loop at `:360-370` asserts only `['files','imports','resolved','distinct_outcomes']`. `baselines.json` stores ten `small`/`deep` fingerprints that nothing reads.

The coordinator set `DEEP_PAD = 16 → 0` in a throwaway copy and ran the real gate: `depth_ratio` collapsed to 1.021, every asserted count was unchanged (padding is count-neutral *by design*, per `_shape_note`), and the gate printed **PASS**. The depth arm — the newest arm in this PR, added precisely because `scaling_ratio` structurally cannot see depth cost — is disabled by a one-token edit. Codex found the same hole from the other direction: a resolver mutation that corrupts only deep results preserves `resolved`, `distinct_outcomes`, corpus shape and the large fingerprint. `[reproduced]`

**Fix:** add `'fingerprint'` to the field list at `measure.mjs:361`. The values are already computed and already stored. Optionally also assert `deep.fingerprint !== small.fingerprint` so the padding's effect is pinned directly.

---

### 4. P2 — The gate certifies the headline claim only on a corpus where every index bucket has size 1

**Engines/lanes:** Claude ce-performance (F2), ce-adversarial (F4) · Codex swarm (F1, rated P1/blocks merge) · Codex ce (#2) · **coordinator CONTROL 13 + CONTROL 17 (both reproduced)**. Strong, both engines. **Blocks merge: no**, but it is the item that most limits what the green gate actually certifies.

`buildRepo` mints per-index directory names — `src/pkg${d}`, `src/Ns${d}`, `lib/feature${d}` (`measure.mjs:127-144`) — so no two directories ever share a last segment. ce-performance enumerated every `pkgPath` the resolver can query against the corpus and measured **max bucket size = 1 and max matching dirs = 1, for go and csharp, at both 400 and 1600 files**. The `dirCount > 1` merge and its `.sort()` are dead code under CI.

**Two consequences, and the digest must not conflate them** (coordinator CONTROL 17):

1. **Shared-leaf-name miss path — the claim does not hold, but nothing regressed.** When the whole workspace uses one leaf name (`pkg`, `internal`, `Models`), the bucket grows with file count and per-import cost stays O(files). Codex B1 measured Go at scaling **3.393** and C# at **3.888** against the PR's own 1.8 budget; ce-adversarial's collide harness measured 2.361 / 2.297 / 1.304 on `mod*/pkg/`-, `Src*/Models/`- and repeated-basename layouts — the *conventions* of those three languages. The coordinator re-ran Codex's exact trigger: **new 3.579 vs old 3.892 — the new code is ~1.2× faster**, so this is a scope-of-claim problem, not a regression. What is wrong is the CI step comment at `ci-tests.yml:511`, which asserts *"per-import cost stays independent of corpus size"* without qualification.
2. **Many-matching-dirs path — this one IS a regression** (finding 2 above), and it is invisible for the same reason.

`[reproduced]` **Fix:** add one `collide` arm — same file/import/resolved counts, colliding last segments and basenames — with its own budget; or a `k_ratio` arm sweeping k at a fixed file count, the direct analogue of the existing `depth_ratio`. At minimum, record the measured bucket-size distribution in `baselines.json` so the scope of the claim is written down.

---

### 5. P2 — The C# no-csproj leg newly retains an O(files × depth) suffix index, and no gate measures memory

**Engines/lanes:** Claude ce-correctness (C-2), ce-performance (F5), ce-adversarial (F5) · Codex swarm (F3) · Codex adversarial (rated high). Strong on the mechanism; the lanes split on severity (P3/P3/P2 vs P2/high) and Codex's ce leg declined to rate it at all *because no harness measured retained heap in-repo*. **P2, blocks merge: no.** The three things that decided the rating, so you can re-decide: the trigger shape is uncommon (a large `.csproj`-free tree), the fix is an addition rather than a revert, and no production comment claims the current footprint is bounded — so nothing is *wrong*, only unmeasured.

At BASE, `getWorkspaceFileIndex` was called **only** inside `if (csharpConfigs.length > 0)`; the no-csproj leg scanned the Set and retained nothing. `csharp/import-target.ts:110` now calls it unconditionally, plus `getCsharpDirIndex` at `:111`. `buildSuffixIndex` emits three maps at O(files × depth) — `exactMap`, `lowerMap`, `dirMap` — which is precisely the profile `package-dir-index.ts:31-36` names as *"exactly the memory this codebase runs out of at kernel scale (#2649)"* and deliberately avoids for itself.

Four harnesses, consistent in direction but **not measuring the same scenario**, so read them as separate data points rather than as a range: ce-correctness **+101 MiB** at 50k C# files (depth not recorded); ce-adversarial **+142.7 MiB** at 100k C# files, depth 8, rising to **+512.9 MiB** at depth 20 and **+2.09 GiB** at 50k files/depth 60; ce-performance **+254 MiB** at 100k; Codex swarm **+337.4 MiB heap / +372.75 MiB RSS** at 100k *Ruby* files, depth 8. The 2.4× spread at "100k/depth 8" is C# vs Ruby, not variance. **None of the 17 coordinator controls measured heap** — this finding's `[reproduced]` rests on the lane harnesses, unlike findings 1–4 where the coordinator re-ran it. The retained object lives for the remainder of `runScopeResolution` — finalize, MRO, reference resolution, PDG and edge emit — which is the peak-memory window.

**Honest framing, because two things here were refuted:**
- The polyglot version of this worry is **wrong** (CONTROL 8, and both Codex legs agree): `run.ts:673` builds `allFilePaths` per provider pass, so read N as *that language's* file count. A 3-file C# project in a 200k-file monorepo indexes 3 files.
- **Ruby is not a peak regression.** It already built and discarded this index *per require*; the change is lifetime, not peak, and it buys a measured 66× wall-clock win plus a large GC-pressure reduction. Strict improvement.
- What is genuinely new is the C# no-csproj allocation, whose trigger is a real repo shape: a large `.csproj`-free tree (Unity repos are the canonical case — generated `.csproj` files are usually gitignored), or one where `CSHARP_SCAN_MAX_DIRS` truncates before a `.csproj` is reached.

`[reproduced]` **Not merge-blocking.** The ask is a number and a ceiling, not a revert: add a `heapUsed`-delta arm at the largest corpus size (all four current arms are time or count), and/or skip `lowerMap`/`dirMap` for the two call paths that never read them — neither Ruby nor no-csproj C# uses `getInsensitive` or `getFilesInDir`, which would cut roughly two thirds of the retained bytes.

---

## Lower-priority items

- **P3 — the Dart backslash fixture pins the guard next door.** `import-target-index-parity.test.ts:568-573` names *"paths are matched RAW"* but its target is all forward slashes, so the bucket lookup misses before the raw comparison is consulted, and the case is registered in `mustBeNull` — it asserts `null === null`. Normalizing only the bucket key, or only the comparison, passes 28/28; only doing both is caught. Two Claude lanes found this independently. Fix: one fixture whose *target* carries the backslash (`files: ['dir\\thing.dart']`, `target: 'package:app/dir\\thing.dart'`), verified to kill both half-mutations. `[reproduced]`
- **P3 — C#/Ruby have zero backslash coverage in the shared index.** The corpus adds `win\dir\thing<ext>` but no C# or Ruby *target* addresses it, so deleting the normalization in `workspace-file-index.ts:44` passes the parity test **and** `--check`. Go does cover it. Fix: add `'Win.Dir.Thing'` to `CSHARP_TARGETS` and `'win/dir/thing'` to `RUBY_TARGETS`. `[reproduced]`
- **P3 — `normToRaw`'s first-wins tie-break is asserted in a comment, not a test.** Dropping the `has` guard passes 28/28; the rule only bites on a normalization twin, which no corpus contains. ce-correctness proved the rule *is* load-bearing with a hand-built probe. One fixture (`['App\\Models.cs', 'App/Models.cs']`, target `App.Models`) closes it. `[reproduced]`
- **P3 — the min-`ord` comparison at `package-dir-index.ts:150` is dead by construction.** Flagged **only** as an optional simplification and as a warning to future readers. See "refuted" below — this is not a correctness finding and not a test gap.
- **P3 — `readonly` on mutable arrays.** `PackageDirIndex` and `WorkspaceFileIndex` expose `string[]` / `Map` behind `readonly` *properties*; `tsc --strict` compiles `idx.rootFiles.sort()` with zero diagnostics. Four Claude lanes flagged it; all four confirmed nothing mutates them today. The hoist did change the blast radius: what used to be per-call scratch is now shared for the whole pass. Mechanical fix (`readonly string[]`, `ReadonlyMap<…>`) makes `go/import-target.ts:89`'s deliberate copy a compile-enforced requirement rather than a comment.
- **P3 — `CountingSet` overrides only `[Symbol.iterator]`** (`:686-691`). `forEach`/`values`/`keys`/`entries` bypass it; Codex's ce leg raised this independently (#4) with the trigger "reintroduce a scan using `allFilePaths.values()` or `.forEach()` once per 32 imports — the scan count stays at its expected value and the timing gate passes". Claude lanes disagree on impact: ce-adversarial calls the docstring an overclaim, ce-testing notes the exact `toBe(1)` assertion turns an evasion into a *failure* (reads 0, not 1). Both are right; the residual risk is a maintainer "fixing" the red number instead of the code. Cheap fix: override the other four traversals too.
- **P3 — the new CI step is inserted at `ci-tests.yml:511`, ahead of eight existing guards** including the Kotlin one at `:531`. GitHub Actions skips later steps after a failure, so a flake in the newest, least-proven gate now masks eight older ones. The general mechanism is **already tracked as open issue #2895** and is not this PR's to fix; only the placement is. Fix: move the new step to the end of the job until #2895 lands. `[code-read]`
- **P3 — doc-accuracy nits.** `ci-tests.yml:511`'s comment says *"the same corpus scores >3.3 against the pre-fix implementations"*, but `_floor` was measured at 150/600 files while the gate runs at 400/1600 (`baselines.json:7` is itself honest about this; only the workflow comment overstates). `csharp/import-target.ts:43-47` documents an index as feeding `findDirectChild`, which this diff deleted. `package-dir-index.ts:42-43` says a polyglot repo *"never pays to index another language's files"* — only the *storage* is per-language; the build pass still walks the whole set once per language.
- **P3 — two test-harness fidelity items.** The Go *"package dir nested inside itself"* fixture (`:544-549`) is excluded by the `endsWith` half and never reaches the first-occurrence branch it names (the generated corpus does cover the branch, and the C# twin does reach it). The C# parity copy hardcodes `evidence = undefined` (`:269`), so `csharpSuffixFallbackAllowed` returns `true` unconditionally across all 440 C# assertions and its blocking branch is untested on both sides.
- **P3 — `depth_ratio` divides two sub-3 ms measurements** and produced a hard FAIL on an unmodified tree in 1 of 4 runs on a contended box. The lane's own caveat: its box is more contended than a GitHub runner, so treat that rate as an upper bound. The construction is real; `_triage` already concedes it.
- **P3 — dead export.** `GoResolveContext` (`go/import-target.ts:98-103`) is "preserved for backward compat" for the legacy call-resolution DAG removed in #942, with zero consumers.

**Pre-existing, out of scope, worth issues with the numbers attached:** PHP's `import-target.ts:185/220` still materializes the workspace twice per import and passes no index — measured **95 ms per import at 20,000 files**, ~1000× the post-fix per-import cost of the four languages here; it is now the only language left without a per-pass memo, and `getWorkspaceFileIndex` makes the fix a two-line change. `import-resolvers/csharp.ts:81` keeps a per-import linear scan on the **csproj** leg — measured 0.95 ms/import at 50k files, but note CONTROL 1: the namespace gate at `csharp.ts:32-42` means it fires only for imports matching a declared in-repo root namespace that also miss both index lookups, so the dominant BCL/NuGet case never reaches it. Ruby's remaining per-import constant (19 ms vs 0.7–2.8 ms for every other language on the same corpus) is `suffixResolve` probing 39 extensions × path-parts, 38 of which a `.rb` query can never match.

---

## Suspicions that were refuted — measurement killed several

These were raised by lanes and then disproven. Listing them is the point: it is what makes the surviving findings credible.

1. **"A polyglot repo builds an unfiltered index over ALL files."** REFUTED at source (CONTROL 8, both Codex legs, ce-performance). `run.ts:673` builds `allFilePaths` per provider pass.
2. **"The min-`ord` comparison is a masked correctness bug / a test gap."** REFUTED (CONTROL 7). One lane reported it "CAUGHT-BY-PARITY-TEST", but it had mutated the guard into *last-wins* (908 divergences, correctly caught) rather than collapsing it to *first-wins* (0 divergences over 2,100 coordinator queries and 6.4M ce-correctness comparisons). It is a genuine equivalence. **No correctness finding and no test gap here.**
3. **"The bench sits in a non-required job."** REFUTED (CONTROL 10). `benchmarks` is inside the reusable `ci-tests.yml` invoked as `tests`; `ci.yml`'s `CI Gate` reads `needs.tests.result`.
4. **"The first-`indexOf` rewrite moves edges."** REFUTED three ways — formal proof, 6.4M-comparison differential, and Codex's independent derivation.
5. **"A production path copies the Set, or a worker boundary clones it, or the Set is mutated after first use."** All REFUTED by code-read: two references to `allFilePaths` in `run.ts` total, no `.add`/`.delete`, no Set crosses a worker boundary, all four adapters pass by reference.
6. **"A live memo bucket escapes and a caller corrupts it."** REFUTED at runtime and by grep of every consumer.
7. **"The verbatim copies drifted / the parity claim is circular."** REFUTED by two independent line-by-line audits against BASE.
8. **"The #2872 corpus-giftability hole recurs — the fingerprint hashes output only."** REFUTED for the large arm: every generated file appears as a `fromFile`, so the file set is inside the hash, and every corpus knob moves an asserted number. The one exception is `DEEP_PAD` — that is finding 3.
9. **"Sparse-import workspaces regress because the index build is not amortized."** REFUTED by measurement: crossover is at **1 import** (Ruby), 2 (Go), 4 (Dart), 32 (C#) at 20k files — the same inversion #2872 found at ~15.
10. **"Dart should have reused `buildSuffixIndex`."** REFUTED — it would reintroduce the #2649 memory shape and change backslash behaviour. The PR made the right call.
11. **"The two new shared modules violate AGENTS.md's no-language-naming rule."** NOT a violation — the rule is scoped to the call/inheritance pipeline, ARCHITECTURE.md:400 sanctions this directory, and the new code is language-free (predicate injection).
12. **"CI fail-fast in the benchmarks job is a new P1."** Real in mechanism, but already open as **#2895** (CONTROL 9). Only the placement sliver is this PR's, at P3.

**Process caveats.** Lanes mutated the shared worktree concurrently; one lane caught another's live mutation and re-did every measurement from `git archive HEAD` (CONTROL 16). Any lane result not taken from a pristine snapshot is provisional. The test/CI lane's numeric details are known-unreliable (CONTROL 11 — it reported per-language scaling budgets that do not exist in `baselines.json`); its mutation-outcome table is still useful as a direction-finder.

---

## CI status and merge state

**Merge state classification: `mergeable`.** Re-checked at synthesis time, not scope time. HEAD is still `478390c8f`. Every required check is green, including `tests / benchmarks (GITNEXUS_BENCH)`, all three OS shards, `quality / typecheck`, CodeQL, gitleaks and `CI Gate`. GitHub reports `UNSTABLE` only because of `Vercel` — "Authorization required to deploy", an org-token issue unrelated to this change and not a required check. CONTROL 10 confirms `benchmarks` does gate merge: it is a job inside the reusable `ci-tests.yml` invoked as `tests`, and `ci.yml`'s `CI Gate` reads `needs.tests.result`.

**Branch hygiene classification: `merge-from-main commit present but harmless and merge-safe`.** Three commits; `8d31e9bac` merges `origin/main` to pick up the Kotlin work from #2872, and the conflict it resolved (two benchmark steps colliding in `ci-tests.yml`, plus a YAML-comment truncation in a step name) was resolved correctly — both steps survive. Diff file count (10) matches the PR's own file list, so the merge-base is right, and there is no unrelated churn. Keeping this as one PR rather than four is justified: the shared index is extracted once and four per-language PRs would each duplicate it, and the five languages must share one bench corpus for their scaling arms to be comparable.

---

## Verdict

**`not production-ready`** — and it is worth being precise about *why*, because the usual reason does not apply here.

The production logic is not what is blocking. It survived 6.4M differential comparisons with zero divergences, an independent re-derivation by the other engine, and 12 refuted attacks. What blocks is that **two of the three gates this PR ships as its proof are demonstrably green-while-broken** (findings 1 and 3), a third certifies the headline claim only on a corpus shape where the new code's one non-constant term is pinned at 1 (finding 4), and one ordinary repository shape is a measured regression against the code being replaced (finding 2). For a PR whose entire thesis is "this is a behaviour-preserving hoist, here is the proof", the proof is the deliverable — and three of these four are small, additive fixes to files this PR already owns.

If you read finding 1 as P2 rather than P1 (as one lane did), this becomes `production-ready with minor follow-ups`. That is a defensible reading and the disagreement is stated rather than hidden.

---

## Coverage limits

- **No end-to-end analyzer run on a real large repository.** Every performance and memory number here comes from synthetic corpora built by the review harnesses. Codex's ce leg could not run the bench at all in its sandbox (`bwrap: No permissions to create new namespace`), so that leg's cost findings are code-read only.
- **No heap measurement among the coordinator controls.** Finding 5 rests entirely on lane harnesses; findings 1–4 were re-run by the coordinator.
- **No full-suite run.** Targeted suites only: parity 28/28, `test/unit/scope-resolution/` + `test/unit/import-resolvers` 1,581 tests, C#/adapter 181 tests, `tsc --noEmit` clean.
- **Windows/macOS behaviour is exercised only as literal backslash strings** — no `path.sep`, no `process.platform`, no real filesystem. Correctly out of `scripts/cross-platform-tests.ts` per TESTING.md's own criteria, but it means the separator claims are string-level only.
- **The GitNexus index is ~45 commits stale** and does not contain the two new symbols, so graph impact for them is `UNKNOWN`; it was confirmed by text search instead.
- **The bench's Kotlin arm overlap** with the existing `bench/kotlin-import-target/` gate was assessed by one lane only.

---

*Automated multi-tool digest (Claude swarm + ce personas, Codex gpt-5.6-sol, 17 coordinator controls). Verify before acting — especially the severity calls, which the two engines did not always agree on.*

---

gitnexus/src/core/ingestion/import-resolvers/package-dir-index.ts
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * Every accepted file directly inside a directory ending with `pkgPath`, in
+ * `allFilePaths` iteration order.
+ */
+export function filesDirectlyInPkgDir(index: PackageDirIndex, pkgPath: string): string[] {
+  let merged: IndexedFile[] | null = null;
+  let dirCount = 0;
+  for (const files of matchingDirs(index, pkgPath)) {
+    dirCount++;
+    merged = dirCount === 1 ? files : [...(merged ?? []), ...files];
```
magyargergo: **[P2]** This accumulator is O(k²·m) and makes the index **slower than the scan it replaced** once many directories share the queried tail.

Re-spreading `merged` once per matching directory copies `m·k²/2` elements per import. Measured against a verbatim copy of the pre-change Go scan, output byte-identical at every k: k=200 → 3.30× slower, k=800 → 7.72×, k=1600 → **19.14×** (new 9.460 ms vs old 0.494 ms); `new_ms` quadruples per doubling of k while the old scan stays linear. End-to-end through `resolveGoImportTarget` the crossover is ≈300 matching directories, and a 2,000-service monorepo measured 7.13× slower than BASE. Go's GOPATH cascade tries every two-segment tail, so `svcN/internal/models`-shaped repos hit this on ordinary imports.

Fix (also what the simplicity lane independently asked for — it is *simpler* than the current code): push into one array and sort once.
```ts
const merged: IndexedFile[] = [];
let dirCount = 0;
for (const files of matchingDirs(index, pkgPath)) { dirCount++; for (const f of files) merged.push(f); }
if (dirCount > 1) merged.sort((a, b) => a.ord - b.ord);
return merged.map((f) => f.raw);
```
Validated byte-identical at every k and 68.5× faster at k=2000 (9.5× at k=250, already inside the region where the new code wins). `[reproduced]`

gitnexus/bench/import-target/measure.mjs
```diff
@@ -0,0 +1,378 @@
...
+  }
+  if (got.deep.resolved !== got.small.resolved) {
+    failures.push(
+      `${lang}: deep arm resolved ${got.deep.resolved} vs small ${got.small.resolved} — the ` +
+        `depth padding was supposed to change path depth and nothing else. A deep arm that ` +
+        `stopped resolving would be timing the null path and its ratio would mean nothing.`,
+    );
+  }
+  for (const scale of ['small', 'large', 'deep']) {
+    for (const field of ['files', 'imports', 'resolved', 'distinct_outcomes']) {
```
magyargergo: **[P2]** `small` and `deep` fingerprints are computed, stored in `baselines.json`, and never compared — the depth arm can be deleted with CI green.

`:311` exports `fingerprint: scales.large.fingerprint` and `:325` is the only fingerprint comparison, so the ten pinned `small`/`deep` hashes are decoration. Because depth padding is count-neutral by design (`_shape_note`), setting `DEEP_PAD = 16 → 0` changes no asserted number: reproduced end-to-end, `depth_ratio` collapsed to 1.021 and the gate printed `[import-target --check] PASS`. The arm added specifically because `scaling_ratio` structurally cannot see depth cost is disabled by one token. Codex found the mirror case independently: a resolver bug that only corrupts deep results preserves `resolved`, `distinct_outcomes`, shape and the large fingerprint.

Fix: add `'fingerprint'` to this field list — the values already exist on both sides. Optionally also assert `deep.fingerprint !== small.fingerprint` so the padding's effect is pinned directly. `[reproduced]`

gitnexus/bench/import-target/measure.mjs
```diff
@@ -0,0 +1,378 @@
...
+  const dirs = Math.max(4, Math.floor(fileCount / 8));
+  const files = [];
+  const ext = { go: '.go', csharp: '.cs', dart: '.dart', ruby: '.rb', kotlin: '.kt' }[lang];
+  const prefix = pad === 0 ? '' : Array.from({ length: pad }, (_, n) => `d${n}`).join('/') + '/';
+  for (let i = 0; i < fileCount; i++) {
+    const d = i % dirs;
+    // A nested same-name directory in one repo slice: the shape whose handling
+    // the first-`indexOf` tie-break decides (see package-dir-index.ts), and the
+    // shape Kotlin's `dirChildren` resolves the same way.
+    const dir =
```
magyargergo: **[P2]** Every generated directory has a unique last segment, so every index bucket has cardinality 1 and the gate cannot see the two costs this PR actually introduced.

`src/pkg${d}` / `src/Ns${d}` / `lib/feature${d}` mean no two directories ever share a leaf. Enumerating every `pkgPath` the resolver can query against this corpus: **max bucket size = 1, max matching dirs = 1, for go and csharp, at both 400 and 1600 files.** Consequences: (a) the `dirCount > 1` merge and its `.sort()` at `package-dir-index.ts:133-138` never execute in any arm — that is the O(k²) regression above; (b) the shared-leaf-name shape, where a bucket grows with file count, scores **3.4–3.9 normalized scaling against the 1.8 budget** (measured independently by both engines; on that shape the new code is still ~1.2× faster than BASE, so it is not a regression — but the CI step comment's *"per-import cost stays independent of corpus size"* is certified only on this one corpus). The nested `src/pkg${d}/internal/pkg${d}` slice does its job for the first-`indexOf` tie-break; it just keeps the leaf unique.

Fix: add a `collide` arm with identical file/import/resolved counts but shared last segments and basenames, with its own budget; or a `k_ratio` arm sweeping k at a fixed file count, the direct analogue of `depth_ratio`. At minimum record the measured bucket-size distribution in `baselines.json` so the claim's scope is written down. `[reproduced]`

gitnexus/src/core/ingestion/languages/csharp/import-target.ts
```diff
@@ -101,12 +102,19 @@ export function resolveCsharpImportTarget(
   }
 
   // Exact file / nested-suffix / namespace-dir direct-child match.
-  const direct = resolveDirectMatch(ctx.allFilePaths, pathLike);
+  //
+  // The no-csproj path used to take the raw Set and re-scan it — up to eight
+  // full workspace passes for a four-segment `using` — past the memoized index
+  // sitting right there for the csproj branch (#2878). Both legs now read the
+  // same per-run indexes.
+  const ws = getWorkspaceFileIndex(ctx.allFilePaths);
```
magyargergo, line 111: **[P2]** The no-csproj leg now retains an O(files × depth) suffix index for the whole pass — the exact memory shape the sibling module rejects on #2649 grounds.

At BASE this call sat inside `if (csharpConfigs.length > 0)`; the no-csproj leg scanned the Set and retained nothing. `buildSuffixIndex` emits `exactMap` + `lowerMap` + `dirMap`, and `package-dir-index.ts:31-36` names that profile as *"exactly the memory this codebase runs out of at kernel scale (#2649)"* while avoiding it for itself. Four review harnesses measured the retention, in **different scenarios — read them as separate data points, not as a range**: +101 MiB at 50k C# files (depth not recorded); +142.7 MiB at 100k **C#** files/depth 8, rising to +512.9 MiB at depth 20 and +2.09 GiB at 50k files/depth 60; +254 MiB at 100k files; and +337.4 MiB heap / +372.75 MiB RSS at 100k **Ruby** files/depth 8. The gap between 142.7 and 337.4 at nominally the same size and depth is C# versus Ruby, not run-to-run variance. In every case the object is held for the remainder of `runScopeResolution` (finalize, MRO, reference resolution, PDG and edge emit). Trigger: a large `.csproj`-free C# tree (Unity, where generated `.csproj` files are gitignored), or one where `CSHARP_SCAN_MAX_DIRS` truncates first. Two things are explicitly *not* claimed: the polyglot version is refuted (`run.ts:673` builds `allFilePaths` per language pass), and Ruby's change is lifetime-only, not peak — it already built and discarded this index per `require`, and now buys a 66× win.

Not merge-blocking. Fix: add a `heapUsed`-delta arm at the largest corpus size (all four current arms are time or count), and/or skip `lowerMap`/`dirMap` for the two call paths that read neither — no-csproj C# and Ruby use neither `getInsensitive` nor `getFilesInDir`, which is roughly two thirds of the retained bytes. `[reproduced]`

gitnexus/src/core/ingestion/import-resolvers/package-dir-index.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * The FIRST accepted file (in `allFilePaths` iteration order) directly inside a
+ * directory ending with `pkgPath`, or `null`.
+ */
+export function firstFileDirectlyInPkgDir(index: PackageDirIndex, pkgPath: string): string | null {
+  let best: IndexedFile | null = null;
+  for (const files of matchingDirs(index, pkgPath)) {
+    const first = files[0];
+    if (first !== undefined && (best === null || first.ord < best.ord)) best = first;
```
magyargergo: **[P3]** `first.ord < best.ord` is unreachable by construction — flagging it so nobody later mistakes the surviving mutant for a coverage hole.

`dirsByLastSegment` receives a directory at the moment its first accepted file is seen (`:88-99`), so bucket order already equals first-file-`ord` order and the first matching directory's `files[0]` is already the global minimum. Replacing the condition with `best === null` alone produced **0 divergences over 6.4M comparisons** and 0 over an independent 2,100-query run. The PR body's claim on this point is correct. Note the contrast: deleting the condition so `best = first` runs *unconditionally* inverts the selection to last-wins and produces 908 divergences — one review lane made that edit and misread the resulting failure as proof the guard is live.

Optional simplification (not required, and `ord` itself must stay — `filesDirectlyInPkgDir`'s multi-directory merge needs it): return on the first matching directory, which also turns `matchingDirs`' laziness into a real early exit. If taken, keep a comment recording that the safety comes from the insertion-order invariant at `:88-99`, because a future edit to the build loop would break it silently. `[reproduced]`

gitnexus/test/unit/scope-resolution/import-target-index-parity.test.ts
```diff
@@ -0,0 +1,760 @@
...
+    {
+      lang: 'dart',
+      why: 'bare `<rel>` is reached only after `lib/<rel>` misses entirely',
+      files: ['a/models.dart'],
+      target: 'package:app/models.dart',
+    },
+    {
+      lang: 'dart',
+      why: 'paths are matched RAW — a backslash path is not normalized into a hit',
+      files: ['win\\dir\\thing.dart'],
```
magyargergo, line 608: **[P3]** This fixture pins the guard next door to the one its `why` names, so half the raw-matching invariant is unguarded.

The `why` says *"paths are matched RAW — a backslash path is not normalized into a hit"*, but the **target** is all forward slashes: `findByPathSuffix` looks up basename `thing.dart`, and the raw bucket key for `win\dir\thing.dart` is the whole string, so the bucket lookup misses before the raw comparison is ever reached. The case is registered in `mustBeNull`, so it asserts `null === null`. Measured consequence: normalizing only the bucket key, or only the comparison, passes 28/28 — only doing both is caught — even though the single-sided change moves Dart output (`files: {'dir\thing.dart'}, target 'package:app/dir\thing.dart'` → current impl returns the file, mutant returns null). `dart/import-target.ts:59-60` documents the stronger property this fixture does not hold.

Fix: add one hand-built case whose *target* carries the backslash, kept out of `mustBeNull` — `{ files: ['dir\\thing.dart'], target: 'package:app/dir\\thing.dart' }`. Verified to kill both half-mutations. `[reproduced]`

.github/workflows/ci-tests.yml
```diff
@@ -508,6 +508,26 @@ jobs:
         run: node --import tsx bench/cpp-qualified-ns/measure.mjs --check
         working-directory: gitnexus
 
+      - name: Import-target resolution guards (#2877/#2878/#2879/#2880/#2872)
```
magyargergo: **[P3]** The new guard is inserted ahead of eight existing ones, so a flake in the least-proven gate masks all of them.

The `benchmarks` job runs 12 sequential `run:` steps with no `if:` on any of them; Actions skips every later step after a failure. Placing this step at `:511` puts it in front of `:531 :542 :567 :579 :589 :598 :607`, including the Kotlin import-resolution guard. The general problem is **already tracked as open #2895** and is not this PR's to fix — only the placement is, and this arm is the one with the known timing-ratio sensitivity (`depth_ratio` divides two sub-3 ms measurements; `_triage` concedes it).

**Action 1 (structural):** move this step to the end of the job until #2895 lands.

**Action 2 (independent, one word):** this comment says *"the same corpus scores >3.3 against the pre-fix implementations"*, but `baselines.json:7` records that floor as measured at **150/600 files** while the gate runs at 400/1600. Same shape, different size; the ratio's scale-invariance keeps the conclusion sound, and `_floor` is itself honest about it — only this comment overstates. Say "the same corpus *shape*". `[code-read]`

koriyoshi2041 (approved):
Re-verified the follow-up at `d2a1ef8e` against the reproduced review failures.

The new production-path integration guards now cross the Go/C#/Dart/Ruby scope-resolver adapters, and the original adapter-copy mutation is no longer green: inserting `new Set(allFilePaths)` in the Go adapter produced the expected `600 to be 1` / `20 to be 1` failures. Restoring the pass-through makes the guard pass again.

The remaining proof gaps are also covered at their mechanisms: matching package directories append once and sort once; deep and collide arms have distinct per-arm fingerprints and asserted budgets; C#/Ruby retained heap is measured under `--expose-gc`; and the newest gate runs last so it cannot mask established checks.

Local verification on Node 26.5.0:

- targeted integration + parity suite: 42/42 passed
- adapter-copy control mutation: failed as intended, then passed after restoration
- `node --expose-gc --import tsx bench/import-target/measure.mjs --check`: PASS
- `git diff --check 478390c8..d2a1ef8e`: PASS

The four headline proof/performance findings from the prior review are closed.

Issue #2900, opened 2026-08-08
fix(java): emit Record interface heritage for scope resolution


azizur100389:
## Problem

Java records can declare `implements`, but `synthesizeJavaInheritanceReferences` currently handles only `class_declaration` and `interface_declaration`; its contract explicitly leaves enum/record heritage unemitted. Consequently a Java record produces no `IMPLEMENTS` edge to its interface, so interface-dispatch fan-out cannot reach the record method even after Record graph nodes became linkable in #2871.

## Reproduction

```java
interface Named { String name(); }
record User(String value) implements Named {
  public String name() { return value; }
}
class Reader {
  String read(Named value) { return value.name(); }
}
```

The `read → Named.name` primary call exists, but neither `User IMPLEMENTS Named` nor the `read → User.name` interface-dispatch edge is emitted.

## Proposed fix

- Widen Java inheritance capture synthesis to include `record_declaration` interface clauses.
- Add focused capture coverage and a Java resolver integration regression.
- Preserve existing class/interface heritage behavior and generic/qualified base normalization.

## Risk

GitNexus upstream impact classifies `synthesizeJavaInheritanceReferences` as **CRITICAL**: 103 upstream symbols across `runScopeResolution`, `processFileGroup`, and `serveCommand`. Keep this separate from the C# consumer fix tracked in #2884.

## Acceptance criteria

- Java Record→Interface `IMPLEMENTS` edges target canonical graph nodes.
- Interface calls fan out to implementing Java record methods.
- Existing Java capture, heritage, and resolver suites pass.

PR #2904, opened 2026-08-08
fix(scope-resolution): fan out C# Record interface calls

### LINKED ISSUE #2884: fix(scope-resolution): include C# Record implementors in interface dispatch
## Problem

After #2871 makes `Record` definitions resolve to their canonical graph nodes, C# record `IMPLEMENTS` / `EXTENDS` edges can reach interface-dispatch fan-out. The consumer still excludes records while building `graphIdToClassDef`:

```ts
if (def.type !== 'Class' && def.type !== 'Struct' && def.type !== 'Interface') continue;
```

As a result, calls through an interface reach the interface method but do not fan out to the implementing C# record method. Callers and `impact` therefore under-report.

Discovered in the review of #2871: https://github.com/abhigyanpatwari/GitNexus/pull/2871#discussion_r3736079701

Java record dispatch is blocked earlier because Java records emit no inheritance captures; that CRITICAL-impact producer change is tracked separately in #2900.

## Proposed fix

- Replace the hard-coded `Class | Struct | Interface` predicate in `receiver-bound-calls.ts` with the existing `isClassLike(def.type)` helper.
- Add a C# record-interface dispatch integration regression.
- Verify ordinary class/struct/interface dispatch remains unchanged and fan-out limits still apply.

## Acceptance criteria

- C# interface calls fan out to implementing record methods.
- Existing receiver-bound call and C# resolver suites pass.
- No schema or parse-cache change is required.

azizur100389:
## Summary

- include canonical C# `Record` definitions in the class-like implementor map used by interface-dispatch fan-out
- preserve every scope definition that shares a graph node so same-file `partial record` fan-out is independent of declaration order
- pin the primary interface edge, implementing-record fan-out, concrete-receiver non-fan-out, and reversed partial-declaration regression
- keep Java record heritage out of scope; its earlier producer gap is tracked in #2900

## Known limitation

Interface-dispatch remains a conservative over-approximation for generic interfaces: heritage currently erases type arguments, so implementations of incompatible instantiations can appear in the fan-out. This behavior is pre-existing for classes and is now visible for records; exact generic-instantiation filtering is tracked in #2912.

## Test plan

- [x] `npx vitest run test/integration/resolvers/csharp.test.ts` (230 passed)
- [x] `npx vitest run test/integration/resolvers/java.test.ts -t "Java bare-this dispatch"` (7 passed)
- [x] `npx vitest run test/integration/resolvers/typescript.test.ts -t "TypeScript chain-typed receiver folding"` (7 passed)
- [x] `npx tsc --noEmit`
- [x] scoped Prettier check
- [x] scoped ESLint (0 errors; existing warnings only)
- [x] GitNexus staged `detect_changes` gate (one symbol, two expected flows, medium)

Closes #2884


gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
```diff
@@ -327,14 +327,14 @@ export function emitReceiverBoundCalls(
...
-  // Maps Interface graph-id → list of implementor class scope-def-ids.
+  // Maps class-like graph ids back to their scope definitions.
   // We translate graph-ids back to scope-resolution DefIds via
   // `parsedFiles.localDefs` lookup so downstream `findOwnedMember`
   // (which keys by DefId) can find the implementor's members.
   const graphIdToClassDef = new Map<string, SymbolDefinition>();
   for (const parsed of parsedFiles) {
     for (const def of parsed.localDefs) {
-      if (def.type !== 'Class' && def.type !== 'Struct' && def.type !== 'Interface') continue;
+      if (!isClassLike(def.type)) continue;
```
magyargergo, line 336: **P2 · record fan-out inherits the type-argument-blind closure** `[reproduced]`

Heritage resolution collapses `IValidator<string>` and `IValidator<int>` onto one `IValidator` def, so this widened guard admits both records into `graphIdToClassDef` and both land in `subtypesBySupertypeDefId[IValidator]`. `pickOverload` (`:473`) returns immediately when the owner has a single same-named member (`overloads.length === 1`), so it never compares the call's argument types.

Trigger — one file:
```csharp
namespace Probe;
public interface IValidator<T> { bool Check(T item); }
public record UserValidator : IValidator<string> { public bool Check(string item) => true; }
public record IntValidator  : IValidator<int>    { public bool Check(int item) => true; }
public class Runner { public bool Run(IValidator<string> v) => v.Check("x"); }
```

Measured BASE vs HEAD on that fixture:
```
BASE:  CALLS Run -> Check@L1 [global]
HEAD:  CALLS Run -> Check@L1 [global]
       CALLS Run -> Check@L2 [interface-dispatch]   <- UserValidator (correct)
       CALLS Run -> Check@L3 [interface-dispatch]   <- IntValidator.Check(int), unreachable from an
                                                       IValidator<string> receiver
```

**Attribution:** the identical fixture with `class` instead of `record` is byte-identical on BASE **and** HEAD — both edges already existed there. The unsoundness (type-argument erasure + name-only member pick) is pre-existing and is **not** introduced by this PR; record hierarchies now inherit it. Not a blocker. Worth a follow-up issue, and a line in the PR body so "fan out to the implementing record" isn't read as exact.

gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
```diff
@@ -327,14 +327,14 @@ export function emitReceiverBoundCalls(
...
       const graphId = resolveDefGraphId(parsed.filePath, def, nodeLookup);
       if (graphId !== undefined) graphIdToClassDef.set(graphId, def);
```
magyargergo: **P2 · `partial record` in one file: the edge depends on declaration order** `[reproduced]`

`graphIdToClassDef.set(graphId, def)` is unconditional over a many-to-one `resolveDefGraphId`, so both parts of a same-file `partial record` mint one graph id and the **last** in `parsed.localDefs` wins. The single consumer at `:363-364` then keys the whole fan-out off that one def, and `pickOverload(subDef.nodeId, …)` sees only that part's members.

```csharp
// A — interface-dispatch edge emitted
public partial record User : INamed { }
public partial record User { public string Name() => "u"; }

// B — the two lines swapped, nothing else changed: NO interface-dispatch edge (HEAD == BASE)
```

Both arms run through an A/B harness; B yields only `CALLS Read -> Name [global]` while the `IMPLEMENTS User -> INamed` edge is still present, which localizes the loss to this map rather than to heritage. Instrumented, the mechanism names itself: `[COLLIDE] graphId=Record:All.cs:User`. Collisions are strictly intra-file (every `resolveDefGraphId` key embeds `parsed.filePath`), so the winner is deterministic in source order but semantically arbitrary — and C# defines `partial` parts as order-independent. Separately, the ordinary multi-file split (`User.Heritage.cs` + `User.Impl.cs`) mints two nodes and produces no fan-out on either side.

**Pre-existing mechanism, newly reachable kind.** The same fixture with `partial class`, parts reversed, is byte-identical BASE↔HEAD (only the `global` edge), so this PR neither introduces nor worsens it. `isClassLike` is not the cause; the single-valued map is.

Fix shape: `Map<string, SymbolDefinition[]>` and `addSubtype` over every def registered for a graph id — `seenTypes` already dedups by `subDef.nodeId` and `tryEmitEdge` by site+target, so extra entries can't double-emit. Minimum bar here: a reversed-order regression row so the order dependence is visible rather than accidental.

gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
```diff
@@ -327,14 +327,14 @@ export function emitReceiverBoundCalls(
   };
 
   // Build an interface → implementors map from IMPLEMENTS edges.
```
magyargergo: **P3 · this comment now contradicts the line directly below it** `[code-read]`

`:330` was fixed by this PR; `:329` was left, and it is wrong in all three clauses:

- the map is graph-id → a **single** def, not interface → implementors;
- it is built by iterating `parsedFiles.localDefs` (`:334-341`), not from edges;
- the real interface→implementors map is `subtypesBySupertypeDefId` (`:352-368`), built from IMPLEMENTS **and** EXTENDS (`:361`).

`:332` also names the wrong consumer: `graphIdToClassDef` is read only at `:363`/`:364`, while `findOwnedMember` is used at `:634/760/1162/1248/1352` — all MRO-walk sites, none of which read this map.

Suggest deleting `:329` (superseded by `:330`) and retargeting `:332` to `pickOverload` / `model.methods.lookupAllByOwner`. Worth restoring the cardinality warning too: the reword dropped the word "list", which was the only textual trace of the many-to-one hazard reproduced in the `:339` comment.

gitnexus/test/integration/resolvers/csharp.test.ts
```diff
@@ -2353,6 +2353,35 @@ describe('C# record base resolution (record inheritance + base.Save)', () => {
...
+
+      const linked = await runPipelineFromRepo(root, () => {});
+      const fanout = getRelationships(linked, 'CALLS').filter(
+        (edge) =>
+          edge.source === 'Read' &&
+          edge.target === 'Name' &&
+          edge.rel.reason === 'interface-dispatch',
+      );
+
+      expect(fanout.map((edge) => `${edge.targetLabel}:${edge.targetFilePath}`)).toEqual([
```
magyargergo, line 2396: **P3 · this test can't distinguish the shipped fix from a narrower one** `[reproduced]`

Mutating the guard to `Class | Struct | Interface | Record` — adding only `Record`, never touching `Enum`/`Trait` — still makes this test pass (executed in an isolated copy: `Tests 1 passed`). So nothing in the suite pins the choice of the shared predicate.

That is fine today: `Enum` and `Trait` are provably dormant (gated out by `ownerDef.type !== 'Interface'` at `receiver-bound-calls.ts:459`, or missing a heritage edge entirely). It is forward-looking that matters — when #2900 lands Java record/enum heritage synthesis, Java gains fan-out through this line with no test in the way.

Also: the filter pins `reason === 'interface-dispatch'` only, so a change that dropped the primary edge while keeping the fan-out would still pass.

Two cheap additions that would earn their keep: a negative row asserting a concrete `User`-typed receiver produces no fan-out (the sibling control already exists at `typescript.test.ts:3646-3659`), and a reversed-order `partial record` row.

magyargergo (commented):
## Review: PR #2904 — fan out C# Record interface calls

**Verdict: production-ready with minor follow-ups.** Nothing in this diff blocks merge. The two-line change is
the fix issue #2884 prescribed verbatim, it is load-bearing, and a corpus-wide A/B says it changes nothing
outside the case it targets. The findings below are about what the change makes *reachable*, not about the change.

**Merge state: `checks failing`** (unrelated — see CI below; GitHub reports `mergeable: MERGEABLE`).
**Branch hygiene: `clean feature/fix PR`** — one commit `d9b966a`, no merge commits, no unrelated files touched.

**How this was reviewed.** Two engines, five lanes.
Claude ran the GitNexus PR swarm (facts/history, risk-architect, security-boundary, test-CI) plus the
`ce-code-review` personas (correctness, adversarial, performance, testing, architecture, maintainability,
simplicity, project-standards). Codex (gpt-5.6-sol, xhigh) ran the same swarm, the `ce` multi-lens pass, and a
native adversarial pass — all three Codex legs ran live and clean. Findings were then re-verified against the
worktree at `d9b966a`; several lane claims were refuted and are listed as such below.

---

### What holds up

- **Exactly the prescribed fix.** #2884 named the defect line and named `isClassLike` as the replacement.
  `isClassLike` was already imported at `receiver-bound-calls.ts:78` and already called at `:1485` before this
  PR, and the sibling map in `passes/mro.ts:62` already used it — `:337` was the outlier, and the PR converges it.
- **The test is load-bearing, and it was executed both directions.** At HEAD:
  `Tests 1 passed | 228 skipped`. With `:337` reverted to the BASE triple:
  `AssertionError: expected [] to deeply equal [ 'Method:User.cs' ]`. So on BASE the fan-out set was **empty**,
  not wrong — the C# record implementor received no `interface-dispatch` edge at all. Minimal and correct.
- **Corpus-wide A/B: byte-identical.** The whole `test/fixtures/lang-resolution` tree (every language, one
  pipeline run) was dumped on BASE and on HEAD — every `CALLS`/`IMPLEMENTS`/`EXTENDS` edge with reason, label,
  file and line. **12,231 edge lines, `cmp`-identical, zero diff.** The widening changes nothing anywhere in the
  shipped corpus.
- **The right lever, architecturally.** The *producer* of every IMPLEMENTS/EXTENDS edge already picks its source
  via `findEnclosingClassDef` → `isClassLike` (`pipeline/run.ts:181`, `walkers.ts:1879`). The consumer filtered
  with a narrower triple. A consumer narrower than its producer drops pairs by construction; this is a structural
  fix, not a C#-specific patch. A `LanguageProvider` hook would have been the wrong shape here — the language
  already declared the fact in syntax and it already became an IMPLEMENTS edge.
- `npx tsc --noEmit` passes. No schema or parse-cache bump needed (resolution-time change) and none made — correct
  per `parse-cache.ts:29-34`. No workflow, manifest, lockfile, or generated file touched, which is worth noting
  for a fork PR. Bidi/zero-width scan and `git diff --check` clean.

---

### Findings

#### 1 — P2 · Record fan-out inherits the type-argument-blind closure: `IValidator<string>` gets an edge to `IValidator<int>`'s implementation `[reproduced]`

`receiver-bound-calls.ts:337`. Heritage resolution normalizes `IValidator<string>` and `IValidator<int>` to the
same `IValidator` def, so both records land in `subtypesBySupertypeDefId[IValidator]`, and `pickOverload`
(`:473`) matches by **name only** — it never compares the call's argument types or the interface's type
arguments. Trigger (idiomatic C#):

```csharp
namespace Probe;
public interface IValidator<T> { bool Check(T item); }
public record UserValidator : IValidator<string> { public bool Check(string item) => true; }
public record IntValidator  : IValidator<int>    { public bool Check(int item) => true; }
public class Runner { public bool Run(IValidator<string> v) => v.Check("x"); }
```

Measured, BASE vs HEAD, same fixture:

```
BASE: CALLS Run@L4 -> Check@L1 [global]
HEAD: CALLS Run@L4 -> Check@L1 [global]
      CALLS Run@L4 -> Check@L2 [interface-dispatch]   <- UserValidator (correct)
      CALLS Run@L4 -> Check@L3 [interface-dispatch]   <- IntValidator  (not reachable from an IValidator<string> receiver)
```

**Attribution matters and is honest:** the same fixture with `class` instead of `record` emits *both* edges on
BASE **and** HEAD, identical. The unsoundness (type-argument erasure + name-only overload pick) is pre-existing
and is not introduced here. What this PR does is make record hierarchies inherit it — and generic
validator/handler/marker interfaces implemented by DTO records are a mainstream C# shape, so the
newly-reachable false-edge population is not small. Not a merge blocker; worth a follow-up issue and a line in
the PR body rather than shipping as an unqualified "fan out to the implementing record".

#### 2 — P2 · `partial record` in one file: the dispatch edge depends on declaration order; across files it never appears `[reproduced]`

`receiver-bound-calls.ts:339` — `graphIdToClassDef.set(graphId, def)` is unconditional over a many-to-one
`resolveDefGraphId`, so both parts of a `partial record` in one file mint the same graph id and the **last** one
in `parsed.localDefs` wins. `pickOverload(subDef.nodeId, …)` then asks only that part for the member.

```csharp
// A — edge emitted
public partial record User : INamed { }
public partial record User { public string Name() => "u"; }
// B — the two lines swapped, nothing else changed: NO interface-dispatch edge (HEAD == BASE)
```

Instrumented, the mechanism names itself: `[COLLIDE] graphId=Record:All.cs:User prev=…#3 new=…#4`. Collisions
are strictly intra-file (every `resolveDefGraphId` key embeds `parsed.filePath`), so the winner is deterministic
in source order but semantically arbitrary — and C# defines `partial` parts as order-independent.
Separately, the **ordinary multi-file layout** (`User.Heritage.cs` + `User.Impl.cs`) mints two distinct graph
nodes and produces no fan-out at all on either side, so the fix does not reach the way most real C# splits
partial records.

**Pre-existing mechanism, newly reachable kind.** `partial class` behaves identically today — the same fixture
with `class` for `record`, parts reversed, was run through the A/B harness and is **byte-identical BASE↔HEAD,
emitting only the `global` edge** — and the shipped corpus already contains one of
these collisions (`csharp-using-static/App/SameFileCases.cs:SameFilePartial`). `isClassLike` is not the cause;
the single-valued map is. The honest ask is a follow-up issue plus a reversed-order regression row, not a merge
block. Fix shape: make the map `Map<string, SymbolDefinition[]>` and try each part —
`seenTypes` already dedups by `subDef.nodeId` and `tryEmitEdge` dedups by site+target, so extra entries cannot
double-emit.

#### 3 — P3 · Twin predicate left behind: `pipeline/run.ts:234` still spells the exact triple this PR removed `[code-read]`

Not in this diff, so noting it here: `emitDetectedInterfaceImplementations` builds the structurally identical
graph-id map and still reads
`if (def.type !== 'Class' && def.type !== 'Struct' && def.type !== 'Interface') continue;`.
Before this PR the two maps agreed; after it they disagree about what "class-like" means. A repo-wide grep
confirms this is the **last hand-rolled class-like triple in shared `src/core/ingestion/`** — every other hit is
`languages/cpp/*`, which is legitimately language-local. Inert today (only Go registers
`detectInterfaceImplementations`, and Go implementors are `Struct`), and the codebase already names this defect
class by hand at `graph-bridge/ids.ts:88-92`. One-line swap to `!isClassLike(def.type)` in this PR is
byte-identical output for every current provider and removes the last copy. Fixing one twin and leaving the
other removes the textual similarity a future reader would use to find the second site.

#### 4 — P3 · The comment block is self-contradictory at HEAD `[code-read]`

`receiver-bound-calls.ts:329-333`. The PR fixed `:330` and left `:329` — `// Build an interface → implementors
map from IMPLEMENTS edges.` — which is wrong in all three clauses and now contradicts the line directly below
it: the map is graph-id → def, not interface → implementors; it is built from `parsed.localDefs` (`:334-341`),
not from edges; and the real interface→implementors map is `subtypesBySupertypeDefId` at `:352-368`, built from
IMPLEMENTS **and** EXTENDS. `:332` also names the wrong consumer — `graphIdToClassDef` is read only at `:363` and
`:364`, while `findOwnedMember` is used at `:634/760/1162/1248/1352`, all MRO-walk sites, none reading this map.
Secondary: the reword deleted the word "list", the one token that described the correct cardinality — the exact
hazard finding 2 reproduces.

#### 5 — P3 · The test cannot distinguish the shipped fix from a narrower one `[reproduced]`

`csharp.test.ts:2356-2384`. Mutating the guard to `Class | Struct | Interface | Record` (add only `Record`,
never touch `Enum`/`Trait`) still makes this test pass. So nothing in the suite pins the choice of the shared
predicate. Given the reachability result below this is not a live behaviour gap — but it is a forward-looking
one: the moment #2900 lands (Java record/enum heritage synthesis), Java gains fan-out through this line with no
test standing behind it. Two cheap additions would also earn their keep: a reversed-order `partial record` row
(finding 2) and a negative row asserting a concrete-record receiver produces no fan-out — the sibling
TypeScript suite already has exactly that control at `typescript.test.ts:3646-3659`.

---

### Lower priority (all pre-existing, none introduced here)

- `languages/csharp/qualified-type-names.ts:17` — `return isClassLike(def.type) || def.type === 'Enum';`. The
  `||` arm has been dead since it was written: `walkers.ts` has included `Enum` since `a7b3fa1b8`, the disjunct
  was added later in `89b02286a`. It encodes a stale belief that `isClassLike` excludes `Enum`, which is exactly
  the misreading that produced #2884. Natural drive-by, −1 line.
- `languages/cpp/user-defined-conversions.ts:134` — a module-local `isClassLike(def: SymbolDefinition)` with the
  *same name* as the shared predicate and the *narrower* `Class|Struct|Interface` set — the set this PR just
  deleted one file over. The narrowing is intentional for C++ conversion operators; renaming it
  `isConversionOwnerType` would stop a future reader "fixing" it into a behaviour change.
- **Performance: no asymptotic change.** Map build stays `O(defs + heritage edges)`; per call site it is
  `O(|closure|)` discovery plus at most 32 emitted edges. A synthetic 200-implementor × 4,000-call-site C#
  corpus measured 11× relationship growth on the `record` arm — but the `class` arm produced byte-identical
  numbers on BASE and HEAD, which is the control proving the mechanism is the pre-existing #2829 fan-out; the PR
  moves records from the zero-cost column into the already-paid one. Two observations that are separable
  follow-ups, not asks on this PR: there is no counter for interface-dispatch edges *emitted* (only for
  *skipped*), so an interface with 30 implementors over 4,000 sites adds ~120k edges with zero log lines; and
  the per-site closure walk re-runs `resolveDefGraphId` for every visited member on every call site
  (`:381-385`, `:421-427`) with no memo, which two ~6-line pass-level caches would collapse.

---

### Suspicions that were tested and refuted — please do not re-raise these

Validation is most of this review's output, so it is stated explicitly.

1. **"Unbounded / uncapped fan-out, DoS vector."** False. `MAX_INTERFACE_DISPATCH_FANOUT` = 32, env-overridable
   (`receiver-bound-calls.ts:259-262`), enforced **before** emit at `:498-506` while counting and sampling the
   dropped names. `git log -S` returns only `cabd5b82f` — the cap predates this PR. Verified firing at 40
   implementors.
2. **"Silently changes Java / Kotlin / Swift / PHP / Rust output."** False in every case. The consumption gate is
   `if (ownerDef.type !== 'Interface') return 0;` (`:459`), so a widened kind produces an edge only as a subtype
   whose *supertype* is an `Interface`.
   - **Swift** (raised by both Codex lanes at 0.98 confidence): the claim reads the class-extractor node label
     (`class-extractors/configs/swift.ts`), but this guard filters `SymbolDefinition.type`, which comes from the
     **scope-extractor** `@declaration.<kind>` capture. `languages/swift/query.ts` emits only
     `@declaration.class` / `.extension` / `.interface` — there is no `@declaration.enum`, and
     `languages/swift/captures.ts` contains no `Enum` re-tag. A Swift enum is typed `'Class'` and was **already**
     in the BASE map. No delta.
   - **Kotlin**: same — `languages/kotlin/query.ts` has no `@declaration.enum`; `enum class` is captured as
     `@declaration.class`. Already covered by the old guard.
   - **Java**: `languages/java/captures.ts:646` states outright *"Enum/record heritage stays unemitted (no legacy
     arm)"* — there is no IMPLEMENTS edge for the widened map to consume. That producer gap is #2900.
   - **PHP / Dart enums**: heritage synthesis walks only `class_declaration` / `trait_declaration`
     (`php/captures.ts:416`) and `class_definition` (Dart `emitHeritage`) respectively. Inert.
   - **Rust / Ruby traits**: the supertype is a `Trait`, never an `Interface`; gated out at `:459`. Verified with
     a `dyn Trait` fixture — HEAD == BASE.
   - Net: today the widening is behaviorally equivalent to `|| def.type === 'Record'`, and specifically to C#
     records. The map grows; the emitted edge set does not. The 12,231-line byte-identical corpus A/B is the
     end-to-end confirmation.
3. **Any edge-DELETION vector.** Refuted. `Trait` *is* a real `NodeLabel` produced by the structure phase
   (`languages/ruby.ts:54-56`, `ast-helpers.ts:406,409`), so the premise that Trait defs must fall through to the
   label-agnostic `simpleKey` fallback is false. Instrumenting `graphIdToClassDef.set` over the whole corpus
   logged exactly **10** collisions — every one same-kind (`Class`↔`Class`, `Struct`↔`Struct`) and every one
   reachable on BASE. No `Record`/`Enum`/`Trait` def collided with a def of another kind.
4. **"Rust is newly live."** Refuted — see 2.
5. **Compiler-generated record members** (`Equals` / `ToString` / `Deconstruct` / `GetHashCode`, positional
   property getters) as bogus dispatch targets. Refuted: tree-sitter sees source text only, so they never exist
   as defs. The record's synthesized primary constructor was also verified *not* to be selected as a target.
6. **Static and abstract record members.** Correct today: a `static` member on a record implementor is skipped by
   `isUnreachableByInstanceDispatch`, and an `abstract` base record member is skipped by `isDeclarationOnly` with
   descent continuing to the concrete override. Both verified end-to-end.
7. **Non-deterministic truncation of the over-cap target set.** Not reproduced — 3/3 identical runs, and the
   parse scan is path-sorted (`pipeline-phases/parse-impl.ts:521`). Untested only under a multi-chunk parse.
8. **The new test flaking or passing for the wrong reason.** Not reproduced. It fails on BASE, passes on HEAD, and
   its exact-set assertion would break on an edge to the interface declaration or to a second implementor.
9. **Security / trust boundary.** No findings. Source + test only; `fs.mkdtempSync` + `finally` cleanup in the
   test; hardcoded fixture keys, no traversal above `root`.

### Test gaps worth an issue (not blocking)

Reversed-order `partial record`; concrete-record-receiver negative control; a two-hop record chain
(`record Derived : BaseRec`, `BaseRec : INamed`, abstract `Name()` on the base — works today, untested); a
`static` member on a record implementor (works today, untested); an `N > cap` arm (nothing in `gitnexus/test/`
references `MAX_INTERFACE_DISPATCH_FANOUT` or `dispatchFanoutSkipped` at all); and a bench arm counting
`reason === 'interface-dispatch'` edges — the three existing benches are all structurally blind to added
fan-out (`bench/receiver-resolution` pins a *drop* census, `scope-emission` pins scope counts, `scope-capture`
fingerprints parse-time captures).

Also worth a negative test or an issue: a record's **positional property** produces no `Property` node, so an
`IHasValue.Value` read never fans out to `User.Value`. Pre-existing capture gap, outside this diff, but easy to
mistake for coverage.

### CI

Red, and unrelated. The failing shard is `tests / ubuntu / coverage 2/3` (run 31278907659, job 93156951153) on
`test/unit/doctor-format.test.ts` and `test/unit/embedding-runtime-install.test.ts` — embedding-runtime platform
detection, both last touched by `021ac3037` (#2765), and the same shard is green on main's tip. That shard's own
summary (job 93156951153) reads `Test Files 2 failed | 279 passed | 2 skipped (283)` and
`Tests 2 failed | 6668 passed | 47 skipped (6717)`; the failing two are the pair above. All 229 C# resolver tests
pass. macOS 3/3 and Windows 3/3 passed. Vercel fails on contributor deploy authorization only. Branch is
7 commits behind `main`; `receiver-bound-calls.ts` and `csharp.test.ts` are untouched on main since the merge
base, and #2856's change to `walkers.ts` added `isShapeLike` without altering `isClassLike`'s return set — so
the merge is clean textually and semantically.

### Final verdict

**`production-ready with minor follow-ups`.** The change is exactly what #2884 prescribed, it is load-bearing
(executed both directions: passes at HEAD, `expected [] to deeply equal [ 'Method:User.cs' ]` with the BASE
guard restored), and it converges the one shared-pipeline outlier onto the predicate its own producer already
uses. The review bar for a scope-resolution fix of this shape — a regression test that fails without the change,
`tsc` clean, no schema/parse-cache bump, and no unintended movement elsewhere in the corpus — is met on every
point, including a 12,231-line byte-identical BASE↔HEAD dump of the whole `lang-resolution` fixture tree. The two
P2 findings are pre-existing mechanisms that record admission newly reaches, each pinned with a `class` control
that is identical on both sides, so neither is caused by this diff and neither should block it. The follow-ups
worth filing are the generic-instantiation over-approximation, the `partial` single-valued map, and the twin
predicate at `run.ts:234`. Open questions: none remaining after verification.

---

*Automated multi-tool review digest (GitNexus swarm + ce personas on Claude; swarm + ce + adversarial on Codex
gpt-5.6-sol). Findings were re-verified against the worktree, but please confirm anything you act on.*

gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
azizur100389: Fixed in 5b5ef306. The bridge now keeps every SymbolDefinition per graph id and adds subtype entries for every matching def pair, so member lookup sees all same-file partial parts. The reversed-order partial Record fixture failed before this change and now passes.

gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
azizur100389: Confirmed this is the pre-existing generic-erasure over-approximation rather than a regression introduced here. I added an explicit Known limitation to the PR body and opened #2912 with the reproducer and acceptance criteria for generic-instantiation-aware filtering.

gitnexus/test/integration/resolvers/csharp.test.ts
azizur100389: Strengthened in 5b5ef306: the regression now pins the primary edge to INamed, the Record fan-out, and no fan-out for a concrete User receiver. It also adds the requested reversed-order same-file partial Record case, which is the behavioral discriminator for preserving every scope definition behind one graph id.

gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts
azizur100389: Fixed in 5b5ef306. The stale interface-to-implementors wording is gone; the comment now describes the graph-id-to-all-scope-definitions bridge, names pickOverload as the DefId-keyed consumer, and documents the many-to-one partial-declaration cardinality.

PR #2916, opened 2026-08-09
fix(java): emit Record interface heritage

### LINKED ISSUE #2900: fix(java): emit Record interface heritage for scope resolution

azizur100389:
## Summary

- synthesize Java inheritance references for `record_declaration` interface clauses
- normalize simple, generic, and qualified record interfaces through the existing Java heritage path
- prove the emitted `IMPLEMENTS` edge links the canonical `Record` node to the canonical `Interface` node
- prove an interface-typed Java call fans out to an explicitly implemented record method through #2904
- document the current implicit-component-accessor limitation and make edge assertions order-independent
- add Record heritage to the Java capture fingerprint and scaling workload

## Review follow-ups

- #2917 tracks materializing implicit Record component accessors as `Method` nodes; this PR includes an explicit negative characterization.
- #2918 tracks Java Enum interface heritage; the temporary enum characterization names that follow-up.
- `PARSE_CACHE_VERSION` is intentionally unchanged: released package-version upgrades invalidate the cache, while this capture-only change preserves the serialized `ParsedFile` shape.

## Risk

`impact` classifies `synthesizeJavaInheritanceReferences` as CRITICAL because it feeds the shared Java capture and parse-worker ingestion path (106 upstream symbols, one direct caller, two affected processes). The implementation remains restricted to `record_declaration`; final `detect_changes` maps the complete review-adjusted diff to three symbols and no affected execution flows.

## Test plan

- [x] `npx vitest run test/unit/scope-resolution/java/java-captures.test.ts` (23 passed)
- [x] `npx vitest run test/integration/resolvers/java.test.ts` (252 passed)
- [x] `node --import tsx bench/scope-capture/measure.mjs --check` (15 languages passed; Java scaling 0.973 < 1.5)
- [x] `npx tsc --noEmit`
- [x] scoped Prettier check
- [x] scoped ESLint (0 errors; existing warnings only)
- [x] GitNexus staged and compare `detect_changes` gates
- [x] final graph status is up to date at `8a7fc6d7`

Closes #2900


gitnexus/test/integration/resolvers/java.test.ts
```diff
@@ -1273,6 +1273,40 @@ describe('Java record method resolution (#2564)', () => {
       fs.rmSync(root, { recursive: true, force: true });
     }
   }, 60000);
+
+  it('links a Record implementor to its canonical Interface node (#2900)', async () => {
+    const root = fs.mkdtempSync(path.join(os.tmpdir(), 'gitnexus-java-record-heritage-'));
+    try {
+      writeFixtureRepo(root, {
+        'RecordHeritage.java': `interface Named { String name(); }
+          record User(String value) implements Named {
```
magyargergo, line 1282: **P2 · cross-engine (Claude + GPT-5.6-sol) · `[reproduced]` — the fixture avoids the shape where the dispatch benefit doesn't materialize.**

This fixture names the component `value` and hand-writes `name()`. That is the case that works. The *canonical* record — where the component's implicit accessor **is** the interface implementation — does not fan out. I ran the three cases inside vitest with the same helpers and the same `reason === 'interface-dispatch'` filter as this test:

| fixture | IMPLEMENTS | interface-dispatch fan-out |
|---|---|---|
| A — this fixture verbatim | `Record → Interface` | `["name@Method"]` |
| B — `record User(String name) implements Named { }` | `Record → Interface` | **`[]`** |
| C — class control, explicit method | `Class → Interface` | `["name@Method"]` |

Row C rules out the harness: it is the implicit accessor, not the scaffolding.

**Why:** no `Method` node is ever materialized for a record component. `languages/java/query.ts:92-93` declares methods from `method_declaration` only, and `method-extractors/configs/jvm.ts:115-118` lists `method_declaration` / `constructor_declaration` / `compact_constructor_declaration` / `annotation_type_element_declaration` — no record-component arm. `grep -rn "record component\|recordComponent\|accessor"` across `languages/java/` and `method-extractors/` returns zero hits. So dispatch has nothing to reach.

**Impact is bounded, and this is still a clear improvement** — before this PR records emitted no heritage at all, so fan-out was impossible for *both* shapes, and records with explicitly-declared methods now work. But #2900's criterion 2 ("interface calls fan out to implementing Java record methods") is only partly met, and the one end-to-end test that claims to prove it exercises only the passing case.

**Suggested (test-only, nothing blocking merge):** add case B as an explicitly-commented negative/xfail so the limit is recorded rather than hidden, and file implicit-accessor materialization as its own issue. Both engines flagged this independently — Codex at 0.95 confidence, though its own file citation was wrong (it named a 71-line type-extraction config); the anchors above are the verified ones.

magyargergo (commented):
# Two-engine review digest — abhigyanpatwari/GitNexus#2916 · `fix(java): emit Record interface heritage`

**Target.** BASE `49c5b7d81` → HEAD `9615f098c`. 3 files, +63/−4. One production hunk:
`gitnexus/src/core/ingestion/languages/java/captures.ts:678` widens
`synthesizeJavaInheritanceReferences` to match `record_declaration` alongside `class_declaration`,
plus a rewritten doc paragraph at `:640-668`. Everything else is tests. Closes #2900.

---

## 0. Engine breakdown and what actually ran

| engine | legs | state |
|---|---|---|
| **A — Claude** | 3 swarm lanes (`risk-architect`, `test-ci`, `facts-historian`) + 6 ce personas (`correctness`, `adversarial`, `performance`, `maintainability`, `testing`, `project-standards`) | **9/9 produced findings.** `test-ci` stalled once and was resumed with an explicit question list. |
| **B — Codex, GPT-5.6-sol / xhigh** | 6-lane Solo swarm (`codex-swarm.md`) | **LIVE.** `grep -ci "rate limit\|turn failed\|skill not found"` → **0**. Six lanes completed with a final verdict: *"production-ready with minor follow-ups"*, no P0/P1. |
| **B — Codex adversarial leg** | red-team | **STARTED, DID NOT COMPLETE IN TIME — not counted.** The transcript is live (0 rate-limit markers) but is still running: it ends mid-run at `[codex] Searching:` with **no findings section and no turn-completion marker** (contrast `codex-swarm.md`, which has one). This is *incomplete*, **not** an all-clear. |
| **B — Codex ce leg** | ce personas | **NOT RUN.** |
| **coordinator** | 3 rounds of executed verification | `claude-coordinator-verified{,-2,-3}.md` — grammar + 19-shape probe; the canonical-accessor experiment; the SCHEMA_BUMP verification. These outrank lane prose where they conflict. |

**Confidence rule in force:** this is **two engines**, not eleven witnesses. Agreement among Claude
lanes alone is *consistent within one engine*; the same for Codex legs. Only a Codex lane agreeing
with a Claude lane counts as cross-engine corroboration, and that is stated explicitly per finding.

**Codex citation correction (coordinator).** Codex F3.2 cites
`class-extractors/configs/jvm.ts:114`. That file is **71 lines** and is the *type*-extraction config —
line 114 does not exist and the file has nothing to do with method members. The substance of F3.2 is
correct and independently reproduced; the anchor is not. Real root cause, verified:
`languages/java/query.ts:92-93` (`method_declaration` only) + `method-extractors/configs/jvm.ts:115-118`
(no record-component arm). `grep -rn "record component|recordComponent|accessor"` across
`languages/java/` and `method-extractors/` → **zero hits**.

---

## 1. What is solid — credit where the review validated the work

- **The one-line predicate widening is correct, and both engines say so independently.** Claude
  `ce-correctness`, `ce-adversarial`, `risk-architect` and Codex lane 3 all attacked it and all
  concluded the same. The coordinator's 19-shape probe [reproduced] confirms correct emission for:
  simple / generic / qualified / multi-interface records, nested records, **Java-16 method-local**
  records (with the javac binary name `Outer$1Local`), records inside interfaces and inside enum
  bodies, and records under a sealed interface — with `Comparable<Box<T>> → Comparable` and
  `a.b.IBar<T> → IBar` normalization intact, and the `class` control unchanged.
- **No false positives on the contextual keyword.** `record` is not reserved in Java. `int record = 1;`,
  `class record { }`, `record r;` (field of a pre-16 type named `record`), `record foo() {…}` — all
  emit `[]`. `class record implements IFoo {}` correctly yields `[IFoo]` through the pre-existing
  *class* arm. Emission structurally requires a non-null `interfaces` field, i.e. a literal
  `implements` after the parameter list, which is never legal in a method declaration.
- **The change is in the right layer and right-sized.** `languages/java/captures.ts` is the sanctioned
  per-language producer; the shared consumer (`preEmitInheritanceEdges`, `emitInheritanceEdgeDirect`)
  names no language and needed no change, so AGENTS.md's "shared ingestion code must not name
  languages" rule holds. One widened condition, one doc paragraph, two tests, zero new abstractions.
- **Both engines independently concluded a shared class-or-record predicate should NOT be extracted.**
  Codex: it "would couple unrelated Java hooks", and `TYPE_DECL_NODE_TYPES` is unsuitable because it
  deliberately includes enums. Claude `ce-maintainability` reached the same conclusion from the four
  hand-spelled sites (`captures.ts:678`, `receiver-binding.ts:86`, `spring-config-bindings.ts:160,201`),
  noting they do not share one clean semantic. **Extracting it would be the over-engineering.**
- **CI really ran the new integration test.** `vitest.config.ts` project include is `test/**/*.test.ts`;
  `tests / ubuntu / coverage 1/3, 2/3, 3/3` + `coverage merge` all pass on `9615f098`. Any
  "the integration test never runs" claim is refuted. 17,671 tests, **17,599 passed, 0 failed**,
  72 skipped (benchmark / platform-sensitive).
- **The PR closes #2900 exactly.** Criteria are (1) Record→Interface IMPLEMENTS targeting canonical
  nodes, (2) interface calls fan out to record methods, (3) existing suites pass. Scope is **records
  only** — enums are not an acceptance criterion. The PR neither under-delivers nor overshoots.
- **`git status --porcelain` is empty** in the worktree at the end of every lane that built artifacts.
- Toolchain gates verified first-hand: `npx tsc --noEmit` clean, Prettier clean on all 3 files,
  ESLint 0 errors on diff-added lines, `java-captures.test.ts` **23 passed**, and the integration case
  passes with `GITNEXUS_WORKER_READY_TIMEOUT_MS=60000` (the worktree failure is the known symlinked-
  `node_modules` startup handshake, not a PR defect).

---

## 2. Headline findings

> **Verdict up front: no P0, no P1, no merge blocker in this review.** Both engines independently
> reached "production-ready with minor follow-ups". H1 below is the strongest finding — cross-engine
> corroborated and reproduced end-to-end. H2 is a maintainer judgment call about cache invalidation.

### H1 — P2 · interface dispatch does **not** fire for the canonical record accessor, and the PR's own fixture dodges it [reproduced] · **CROSS-ENGINE**

- **Agreed by both engines:** Codex **F3.2** (P2, conf 0.95) + Claude `ce-correctness` **F1** (P2, conf
  0.95), then re-run as a controlled experiment by the coordinator
  (`claude-coordinator-verified-2.md`).
- **Instrument note:** an out-of-vitest `tsx` probe reports `interface-dispatch CALLS: []` for *every*
  case including the class control and the PR's verbatim fixture, so it cannot observe dispatch at all.
  The experiment was run **inside vitest**, in a temp test file deleted afterwards (worktree verified
  clean).

  | # | fixture | IMPLEMENTS | interface-dispatch fan-out |
  |---|---|---|---|
  | A | **PR's fixture verbatim** — `record User(String value) implements Named { public String name() {…} }` | `Record → Interface` | **`["name@Method"]`** |
  | B | **canonical record** — `record User(String name) implements Named { }` | `Record → Interface` | **`[]`** |
  | C | **class control** — `class User implements Named { public String name() {…} }` | `Class → Interface` | **`["name@Method"]`** |

- **What it establishes.** The **producer half works in every case** — #2900 criterion 1 is met for
  both shapes. The **consumer half (criterion 2) does not fire for the idiomatic record shape**: row B
  is the ordinary way a Java record satisfies an interface, the component's implicit accessor *is* the
  implementation, and no `Method` node is ever materialized for it, so dispatch has nothing to reach.
  **Row C rules out the harness** — the difference is the implicit accessor, not the fixture
  scaffolding. The PR's fixture names the component `value` and hand-writes `name()`, so the one
  end-to-end test claiming to prove criterion 2 exercises only the case that works.
- **Root cause [code-read]:** `languages/java/query.ts:92-93` declares Methods from
  `method_declaration` only; `method-extractors/configs/jvm.ts:115-118` lists
  `method_declaration` / `constructor_declaration` / `compact_constructor_declaration` /
  `annotation_type_element_declaration` — no record-component arm. `field-extractors/configs/jvm.ts:29`
  is `['field_declaration']` only, so the component is not a field either. Downstream,
  `receiver-bound-calls.ts:480-487` needs `pickOverload` to find a member on the subtype and bails.
- **Severity P2, not P1:** the PR is still a strict improvement (records previously emitted no heritage
  at all, so fan-out was impossible for *both* shapes), and records with explicitly-declared methods now
  work. But the PR body's claim to prove an interface-typed call fans out to the implementing record
  method is narrower than it reads, and #2900 criterion 2 is only partly met.
- **Fix (test-only, no production change needed to merge):** add the row-B case as an explicitly
  documented negative/xfail so the limit is recorded rather than hidden, and file implicit-accessor
  materialization as its own issue. C# already does the analogous record work
  (`field-extractors/configs/csharp.ts:105`).

### H2 — P2 · parse-time capture change ships **without a `SCHEMA_BUMP`** — a judgment call, not a blocker [reproduced]

- **Evidence:** `gitnexus/src/storage/parse-cache.ts:471` → `const SCHEMA_BUMP = 60;` — **identical at
  BASE and HEAD** (`git diff BASE HEAD -- src/storage/parse-cache.ts` → empty). The changed code is
  parse-time emission: `synthesizeJavaInheritanceReferences` ← `emitJavaScopeCaptures`
  (`languages/java/captures.ts:397`) ← the parse worker, and its output is baked into the persisted
  `ParsedFile`.
- **Trigger:** the durable ParsedFile store is gated on `PARSE_CACHE_VERSION` **alone** —
  `parse-impl.ts:764-771` says so in its own comment; `parsedfile-store.ts:640-652` compares
  `idx?.version !== expectedVersion` and nothing else. Unchanged version ⇒ warm index reused ⇒ chunks
  not re-dispatched ⇒ the new capture code never runs for unchanged Java files. **`--force` does not
  bypass it**, and `run-analyze.ts` says why in the exact terms this PR invalidates: *"Content-addressed:
  safe to reuse across `--force` runs (chunks whose file contents haven't changed produce identical
  worker output)."* That justification is now false for Java.
- **Mechanically reproduced, 3 runs + control** (`ce-adversarial`, byte-identical fixture,
  `PARSE_CACHE_VERSION=60+1.6.9` throughout):

  | # | analyzer | store | result |
  |---|---|---|---|
  | 1 | BASE-emulating | cold | `heritage=[]` · `interface-dispatch=[]` |
  | 2 | **HEAD** | **warmed by run 1** | **`heritage=[]` · `interface-dispatch=[]`** |
  | 3 | HEAD | fresh (control) | `heritage=["IMPLEMENTS Record:User → Interface:Named"]` · `interface-dispatch=["read → Method:name"]` |

  Run 2 is the real upgrade path: existing `.gitnexus` index, unchanged Java source, new analyzer.
- **Co-change context (not an obligation):** of the last 29 commits touching
  `src/core/ingestion/languages/`, **20 also touched `parse-cache.ts`**. Read this as context only —
  most of those bumps accompany changes to the persisted `ParsedFile` **shape** (a new capture kind, a
  new field, a serialization change), where a warm cache is genuinely *incompatible*. This PR is a
  different class. If the author does choose to bump, the file's own ledger at `:455-470` gives the
  rule: *"Take the next free value above every in-flight MAXIMUM, not above origin/main. … RE-CHECK
  AGAINST origin/main IMMEDIATELY BEFORE MERGING."*
- **CI would not surface it either way:** `test/unit/incremental-parse-cache.test.ts:199-204` pins
  `toBe(60)`, so leaving it stays green; and the PR's integration test uses `fs.mkdtempSync` with no
  `parseCache`, so it is a cold-cache assertion by construction.
- **Bounded exposure — the half that must be stated with the mechanism.** `PARSE_CACHE_VERSION` is
  `` `${SCHEMA_BUMP}+${GITNEXUS_PKG_VERSION}` `` and the package version is **1.6.9**; when the fix
  ships in any released version bump the second component changes, the cache invalidates, and **users
  upgrading through npm do get the fix**. The lane's 3-run experiment necessarily held the package
  version fixed (one worktree), so it is good evidence that the mechanism **bites at a fixed package
  version** — it does not establish release-facing breakage. **Exposure is therefore limited to**
  dev / source-checkout / monorepo runs and incremental CI at an unchanged package version.
- **Calibration — P2, and NOT a merge blocker.** The project's own recorded practice reaches the
  opposite conclusion from the lane, and it is decisive here: after debugging this exact mechanism on
  #2730, the maintainers' note states *"Not a release problem (a real release bumps the package version,
  which is in the key) — dev-build only, so **don't bump `SCHEMA_BUMP` for an ordinary capture fix**."*
  This PR is an ordinary capture fix: it emits **more captures of an existing kind**
  (`@reference.inherits`) with no new field and no serialization change, so a warm cache replays
  stale-but-structurally-valid `ParsedFile`s — behaviourally out of date, not corrupt.
- **Net:** the mechanism above is real and verified, the exposure is bounded to dev / source-checkout /
  incremental-CI runs at an unchanged package version, and whether to bump is **a judgment call the
  maintainer owns**. Raised as an FYI so the decision is made deliberately rather than by omission.
- **Corroboration:** **single-lane** (`claude-ce-adversarial`), one engine. Every *code* claim was
  independently re-verified by the coordinator (`claude-coordinator-verified-3.md`), so the mechanism is
  reported as established; the lane's *severity* framing was not sustained and is corrected above.
- **If the maintainer chooses to bump:** take the next free `SCHEMA_BUMP` above every in-flight maximum
  per the ledger's own rule (not `origin/main + 1`), update the pin at
  `test/unit/incremental-parse-cache.test.ts:199-204`, and add a ledger entry naming the
  `record_declaration` arm. A warm-parse-cache test would pin the behaviour either way — two working
  templates exist: `test/integration/spring-aop-pipeline.test.ts:925-945` (exercises the durable
  ParsedFile store specifically) and `test/integration/fastapi-composed-route-constants.test.ts:167`
  (`#2391 SCHEMA_BUMP`).

---

## 3. Inline-worthy findings (beyond H1/H2)

### I1 — P2 · the enum negative test converts an acknowledged gap into an asserted invariant · **CROSS-ENGINE, 6 lanes**
`test/unit/scope-resolution/java/java-captures.test.ts:101-103`. Codex **F3.1** (P2, conf 0.99) +
Claude `risk-architect` F2, `test-ci` F1, `ce-correctness` F2, `ce-adversarial` P2, `ce-testing` F5.
Grammar-verified [reproduced]: `enum_declaration.interfaces` is the **same** `super_interfaces` →
`type_list` shape as `record_declaration` (vendored tree-sitter-java 0.23.5 `node-types.json`).
`isClassLike` (`scope/walkers.ts:180-189`) already includes `'Enum'`; `'Enum'` is in `LINKABLE_LABELS`
so `FROM Enum TO Interface` is already declared DDL; `TYPE_DECL_NODE_TYPES` (`captures.ts:576-580`)
already lists `enum_declaration`. So the correct completion is literally
`|| node.type === 'enum_declaration'` at `:678`, with **no new node-shape handling** — and that one
token makes this new test **fail**, i.e. the future correct fix must delete a green test. Kotlin already
emits it (`emitKotlinScopeCaptures('enum class Status : Named { ACTIVE }')` → `["Named"]`), and
`enum X implements Y` is more idiomatic Java than `record X implements Y`. Verified end-to-end at HEAD:
`enum Status implements Named {…}` + an interface-typed caller → `IMPLEMENTS=[]`,
`interface-dispatch=[]`. **Attribution: the gap is PRE-EXISTING and deferring it is the right call for
one commit; #2900 is records-only.** The finding is the *undocumented lock-in*:
`facts-historian` confirmed via `gh issue list` that **no issue tracks enum heritage**, so the test and
the `"Enum heritage remains unemitted"` comment point at nothing. **Fix:** rename to
`does not YET emit …`, add a comment saying the fix that emits enum heritage must delete this test, and
file the tracking issue — or widen to `enum_declaration` here and delete the test.

### I2 — P3 · the doc paragraph argues from a legacy query that was deleted two months ago · **CROSS-ENGINE, 5 lanes**
`captures.ts:643-651`. Codex **F6.1** (P3, conf 1.00) + Claude `ce-maintainability` #1,
`ce-project-standards` F1, `risk-architect` F3, `facts-historian`. The paragraph asserts in the
**present tense** that a legacy Java heritage query in `tree-sitter-queries.ts` exists and "has a
dedicated `interface_declaration (extends_interfaces …)` arm", and the PR **adds** a new sentence
resting on it: *"Record heritage is emitted even though the legacy query had no matching arm"*.
[reproduced] `grep -n 'superclass\|super_interfaces\|extends_interfaces'` over `tree-sitter-queries.ts`
→ **zero hits**; that file's own header (`:8-11`) states heritage is NOT captured there and the legacy
leg was removed in #942; `pipeline/run.ts:163-164` says the pre-pass is *"the authoritative inheritance
emitter"*. `facts-historian` pinned the deleting commit: **`083aedbc41` (PR #2023, merged 2026-06-04)**,
and confirmed the legacy query had exactly three arms and **never** a `record_declaration` arm.
Sibling providers phrase this correctly in the past tense (`csharp/captures.ts:438-440`,
`go/captures.ts:321`, which points at the *surviving* `heritage-extractors/configs/` location);
Java's doc is the outlier. The PR did not introduce the staleness — it added a sentence that deepens it,
while the paragraph was open for edit. **Fix:** drop the legacy-query framing; state the positive
reason (Record nodes are canonical link targets — issue #2801, implemented by **PR #2871**, which
`facts-historian` verified as the actual implementing PR).

### I3 — P3 · the record arm reads a `superclass` field the grammar can never produce · **CROSS-ENGINE, 5 lanes**
`captures.ts:679-682`. Codex **F3.3** (P3, conf 0.99) + Claude `risk-architect` F1,
`ce-maintainability` #2, `ce-correctness` F3, `ce-performance` PERF-3. [reproduced] tree-sitter-java
0.23.5 `record_declaration` fields = `body, interfaces, name, parameters, type_parameters` — **no
`superclass`, no `permits`**; `grammar.js:1022-1030` agrees; `record R(int x) extends Base {}` collapses
the `extends` into an `ERROR` node. Correct per JLS §8.10. **Dead, not wrong — and not a defect.** It is
filed because the file states the opposite invariant 85 lines earlier
(`captures.ts:593-595`: *"Only class_declaration carries a `superclass` field; enum/record cannot declare
an explicit superclass"*), so the new code and the existing comment now contradict each other in one
file. `risk-architect` notes `receiver-binding.ts:86-90` already uses the identical guard-then-lookup
shape, so the PR follows an established convention rather than inventing a smell. **Fix:** one comment,
or split so the `superclass` read is class-only.

### I4 — P2 · the Java capture-fingerprint CI gate is structurally **blind** to this change [reproduced]
`bench/scope-capture/measure.mjs:265` + `bench/scope-capture/baselines.json:135`. Single-lane
(`claude-ce-performance`), one engine — but the lane **ran the real gate**:
`node --import tsx bench/scope-capture/measure.mjs --check` → **PASS (15 languages)**, java fingerprint
`b29e2635…7663`, **byte-identical to the committed baseline**, on a PR whose entire purpose is a Java
capture-output change. Two independent causes: (1) the scale generator emits only
`class Entity${n} extends Base implements Marker` — **no record at any scale point**; (2) of 206 `java-*`
fixture files exactly three contain a record (`java-record-methods/Point.java`,
`java-local-class-naming/src/{Types,Compact}.java`) and **none implements anything**. The repo already
treats this as a reportable defect — see the cpp `_rebaselined_2833_generic_member_fields` note in
`baselines.json`, same failure in another language. **Consequence:** reverting `:678` to BASE deletes
every Record→Interface IMPLEMENTS edge and, on the measured shape, **1,280 of 1,320 CALLS edges (97%)**,
and the gate still reports PASS. **Fix:** add one record with an `implements` clause to the java scale
unit (or to `Point.java`) and re-baseline with the usual `_rebaselined_*` note.

### I5 — P3 · order-sensitive `toEqual` on an unsorted edge array · two lanes, one engine
`test/integration/resolvers/java.test.ts:1303-1305`. `claude-swarm-test-ci` F2 + `claude-ce-testing` F3.
`getRelationships` iterates `graph.iterRelationships()` in **insertion order**, unsorted. Harmless today
(exactly one implementor ⇒ 0 or 1 elements), but the sibling multi-implementor block in the same file
(`java.test.ts:2189-2199`, same fan-out mechanism, two Class implementors) deliberately avoids `toEqual`
on an unsorted array, and the file owns an `edgeSet` helper (`helpers.ts:138-140`) that sorts first.
The natural next step — adding a second implementor — walks straight into it. **Fix:** `.sort()` before
`toEqual`. Related [reproduced] trap: the coordinator's probe shows the production DFS emits sibling
declarations in **reverse source order** (`class A implements I1 {} record B(int i) implements I2 {}` →
`["I2","I1"]`), and the new `inheritanceRefs` helper asserts exact array equality — both new unit cases
use a single declaration, so they are safe today.

---

## 4. Lower-priority items

- **P3 · `Record EXTENDS Class` is now reachable on malformed source.** `pipeline/run.ts:207-208` picks
  the edge type from the resolved target's kind alone: `targetDef.type === 'Interface' || 'Trait' ?
  'IMPLEMENTS' : 'EXTENDS'`. For a record, `EXTENDS` is *never* legal Java. Verified end-to-end
  (`class Base {}` + `record User(String v) implements Base {}` → `EXTENDS ["Record:User → Class:Base"]`);
  the class control gives `Class:User → Class:Base`, wrong too but at least a legal shape. **Only
  reachable from non-compiling Java** — four legal shadowing/collision constructions were built and all
  refused or resolved correctly. Downstream inert today because interface dispatch gates on
  `ownerDef.type !== 'Interface'`. `risk-architect` judged it a pre-existing resolution defect shared
  with class heritage. New *surface*, not a live defect.
- **P3 · TYPE_USE-annotated bases are silently dropped — PRE-EXISTING, proven by a `class` control.**
  `javaBaseLookupNameNode` (`captures.ts:737-752`) has no `annotated_type` arm.
  `record U(int i) implements @Nullable IFoo {}` → `[]`; control `class U implements @Nullable IFoo {}`
  → `[]`; also `class U extends @Nullable Base {}` and `interface I extends @A IFoo {}` → `[]`.
  Partial-loss case: `implements @A IFoo, IBar` → `[IBar]` only. Real-world trigger: JSpecify /
  Checker Framework / Jakarta annotations on heritage types. Records merely join an existing hole.
  One-case fix (`case 'annotated_type': return javaBaseLookupNameNode(node.lastNamedChild)`) would fix
  classes, interfaces and records at once — follow-up, not this PR.
- **P3 · malformed heritage emits an empty-name reference.** `record R(int i) implements {}` → `[""]`;
  control `class R implements {}` → `[""]`. Pre-existing, resolves to nothing downstream.
- **P3 · `inheritanceRefs` helper silently swallows nameless captures.**
  `java-captures.test.ts:34-36` — `.flatMap((m) => m['@reference.name']?.text ?? [])`. Inert today (all
  three `@reference.inherits` emitters always pair the two keys), but the sibling `ctorRefs` helper ten
  lines above takes the opposite, more observable stance. Three lanes, one engine.
- **P3 · one N-API call and one `node.type` read per node of unclaimed headroom.** `node.type` is a
  native accessor (`tree-sitter/index.js:88-91`), not a cached property, so the widened predicate costs
  **two** crossings on every non-class node where BASE cost one. Measured on 34,409 named nodes:
  base 50.84 ms → head 57.64 ms (**+13.4%**) → hoisting to a local 44.48 ms (−22.8% vs head, −12.5% vs
  BASE). **In proportion:** 6.8 ms of a 499 ms `emitJavaScopeCaptures` = **1.36% of Java capture time**,
  ≈1.8 s wall on a 50k-file monorepo across 8 workers, below the end-to-end noise floor.
  `scaling_ratio 1.043` against a 1.5 budget. Optional one-line cleanup; keeping the diff at one line is
  also defensible.
- **Graph growth is real, intended, bounded — and nobody has been told.** Measured pipeline probe
  (1 interface, 40 implementors, 40 interface-typed call sites): record variant emits **40 IMPLEMENTS,
  1320 CALLS** — byte-identical to the class control; at BASE the record variant emits **0 IMPLEMENTS**
  and only 40 primary CALLS. **⇒ 33× CALLS on a record-dense dispatch shape**, exactly the multiplier
  classes already produce, capped by `MAX_INTERFACE_DISPATCH_FANOUT = 32` per call site per member and
  honestly reported (`{"dispatchFanoutSkipped":320,"fanoutCap":32}`). Expect
  low-hundreds-of-thousands of new CALLS on a record-heavy Java monorepo's first re-index, and a noisier
  `dispatchFanoutSkipped` counter. Nuance: **the cap bounds emitted edges, not work** — `targets` is
  collected across the whole subtype closure (`receiver-bound-calls.ts:471-497`) and truncated only at
  `:505-513`, so R record implementors × S call sites costs R×S extra `pickOverload` lookups. Mitigating:
  a Java record is implicitly final (JLS §8.10), so it is always a **leaf** in the closure and never
  branches it. Pre-existing #2829 design, not attributable here.

---

## 5. Refuted — attacks the review ran and disproved (validation is a feature)

Every row below was run against a **`class`-for-`record` control**; unless noted, record and class
output were identical.

| # | attack | verdict |
|---|---|---|
| R1 | `record` is a contextual keyword → the widened match fires on pre-16 code | **REFUTED.** 7 shapes probed. The one real mis-parse (`class C { record m(int x) {…} }` → a `record_declaration` named `m`) carries **no `interfaces` field**, so the predicate cannot fire on it. |
| R2 | Grammar drift / `permits` / sealed interaction | **REFUTED.** `record_declaration` has no `permits` field (records are implicitly final; sealed records are illegal). `sealed interface S permits R {}` puts the `type_list` under a `permits` **field** the code never reads. |
| R3 | Malformed / partial source (truncated, no body, `implements , Named`, unclosed, ERROR soup) | **REFUTED — nothing throws, nothing emits garbage.** Truncated forms collapse into `ERROR` so the walk emits `[]`; the rest match the class control exactly. |
| R4 | Emitting an illegal edge from **valid** Java (4 collision/shadowing constructions) | **REFUTED.** Import channel wins; resolver **refuses on tie**; nested `interface` correctly shadows a top-level `class` in all 4 orderings; no cross-language contamination from a TypeScript `class Named`. Only the invalid-Java case reproduces (see §4). |
| R5 | `@interface` annotation types as a base | **REFUTED.** `@interface` produces no type node at all, so nothing resolves and nothing is emitted — identical for the class control. |
| R6 | Fan-out amplification (N ∈ {10,40,100} implementors × {1,3} sites) | **REFUTED — cap works, records match classes exactly.** 10/1→10; 40/1→32; 40/3→96; 100/3→96. |
| R7 | Enclosing-scope mis-attribution (method-local, in-interface, in-enum, in-anonymous-class) | **REFUTED — attribution is exact.** `Record:Outer$1Local`, `Record:Inner`, `Record:Outer$1$1Deep` — never the lexically enclosing type. |
| R8 | Duplicate emission via Java's two other `@reference.inherits` emitters | **REFUTED.** Both anchor on `class_body` reached from `object_creation_expression` / `enum_constant`, never from `record_declaration`; `emitInheritanceEdgeDirect` would dedupe anyway. |
| R9-R13 | Perf/DoS: 5,000 interfaces on one record; 264 KB / 5,001-line file; 1,200-deep nesting; 1,000-deep nested generics; type-list junk (`implements int`, `Named[]`, `A, A`, `A<?>`, `Map.Entry<K,V>`, non-ASCII, trailing comma) | **ALL REFUTED — strictly linear, no stack blowup, no false emission.** 5,000 interfaces → 48.4 ms; 15,000 refs → ≈+280 ms over the no-heritage baseline; nesting handled because the walk is an explicit stack and the product path passes `getTreeSitterBufferSize`. |
| R14 | `graphIdToClassDef` last-write-wins (carried from the #2904 memory) | **REFUTED at HEAD — already fixed.** `receiver-bound-calls.ts:333` now accumulates `SymbolDefinition[]` arrays: *"Preserving every part makes dispatch independent of declaration order."* |
| R15 | "Missing `MAX_INTERFACE_DISPATCH_FANOUT` cap" | **REFUTED — it exists** (`receiver-bound-calls.ts:259-262`, enforced `:505-513`, env-overridable). A stale memory claimed otherwise; both engines independently confirmed it is present. |
| R16 | "The new `Record IMPLEMENTS Interface` edge is dropped at the LadybugDB boundary" | **REFUTED.** `'Record'` ∈ `LINKABLE_LABELS` since #2801, so `generatedPairDdl()` mints `FROM Record TO Interface`; C# already exercises the pair. Edge observed end-to-end. |
| R17 | "Skip descending into method bodies as a cheap win" | **REFUTED.** Java 16+ local records/classes live in method bodies and the repo has committed fixtures depending on it. Correctly not done. |
| R18 | "The direct cost of the widened predicate is a material regression" | **REFUTED.** 198 ns/named node, 1.4% of Java capture time, below the end-to-end noise floor. |
| R19 | "The worktree integration failure is a PR defect" | **REFUTED — environment.** Missing `dist/` + the 5 s worker ready handshake on symlinked `node_modules`; passes with `GITNEXUS_WORKER_READY_TIMEOUT_MS=60000`, and CI ran it green. |
| R20 | #2904 finding 5 ("#2900 activates untested fan-out at `run.ts:240`") | **REFUTED.** `detectInterfaceImplementations` is registered by exactly one provider — `languages/go/scope-resolver.ts:93`. Java registers nothing, so `emitDetectedInterfaceImplementations` never executes for Java. The `Record`-excluding triple there stays a latent trap for whoever registers the hook next, but it is not a defect of #2916. |

---

## 6. Test gaps (advisory)

1. **No canonical-accessor fixture** — the exact shape that hides H1 (`record User(String name)
   implements Named {}`), i.e. the idiomatic record that satisfies its interface implicitly.
2. **No warm-parse-cache test for Java** — nothing pins whether the new heritage survives a warm run
   (H2), in either direction. Two working templates exist
   (`spring-aop-pipeline.test.ts:925-945`, `fastapi-composed-route-constants.test.ts:167`).
3. **No record with an `implements` clause anywhere in the `bench/scope-capture` java corpus or scale
   source** — I4; the one CI job that exists to catch Java capture-output changes came back PASS with an
   unmoved fingerprint.
4. **Integration coverage is one interface, one implementor, one file.** Unit covers generic
   (`Comparable<User>`) and qualified (`audit.Auditable`) bases; nothing proves multi-interface,
   cross-file/cross-package, or generic record heritage end-to-end. Lower risk than it looks —
   `preEmitInheritanceEdges` is untouched and already exercised for classes — hence P2/P3, not P1.
5. **No nested / method-local record heritage assertion at the integration level.** The capture layer
   is [reproduced] correct (`Outer$1Local`), but scope *attribution* at resolution time is only proven
   for one top-level record. `java-local-class-naming/src/Types.java` already declares a method-local
   record with no `implements` — a cheap place to extend.
6. **No malformed-source coverage** around record heritage (behavior is safe today per R3, just unpinned)
   and **nothing pins the `superclass`-is-always-null invariant** (I3).
7. **No annotated-base case** for any Java heritage (§4).

---

## 7. CI status

**45 checks. One failure: `Vercel — Authorization required to deploy`** — a fork deploy-authorization
block, not a code failure. Everything else is SUCCESS: `CI Gate`, typecheck, quality, security, and the
sharded test matrix (ubuntu ×3, windows ×3, macOS ×3, plus coverage merge). CI bot: **17,671 tests —
17,599 passed, 0 failed, 72 skipped**. `e2e (chromium)` skipped as expected (no web changes). GitHub
reports `MERGEABLE` / `mergeStateStatus: UNSTABLE` solely because of the Vercel status. One linear
commit atop the stated base, no merge commits, `git diff --check origin/main...HEAD` clean, no
`CHANGELOG.md` edit. No human reviewer approval is recorded.

---

*This is an automated multi-tool digest — two engines (Claude, 9 lanes; Codex GPT-5.6-sol/xhigh, Solo
swarm) plus three rounds of coordinator-executed verification. Codex's adversarial leg started but did
not complete in time and is not counted; its ce leg was not run. Corroboration counts are per engine,
not per lane. Nothing here is a merge gate — verify each claim against the cited file:line before
acting on it.*

gitnexus/test/unit/scope-resolution/java/java-captures.test.ts
```diff
@@ -82,6 +89,20 @@ describe('emitJavaScopeCaptures — constructor reference names (F35 #1928)', ()
...
+describe('emitJavaScopeCaptures — record interface heritage (#2900)', () => {
+  it('captures simple, generic, and qualified record interfaces by lookup name', () => {
+    const refs = inheritanceRefs(
+      'record User(int id) implements Named, Comparable<User>, audit.Auditable {}',
+    );
+
+    expect(refs).toEqual(['Named', 'Comparable', 'Auditable']);
+  });
+
+  it('does not widen enum heritage while adding record support', () => {
```
magyargergo: **P2 · cross-engine (Claude + GPT-5.6-sol), 6 lanes — this pins a live gap as an asserted invariant, and nothing tracks it.**

Deferring enums is the right call for one commit — #2900 is records-only and the PR neither under-delivers nor overshoots. The issue is that this test makes the gap look *intended* when it is one token away from being fixed:

- Vendored tree-sitter-java 0.23.5 `node-types.json`: `enum_declaration.interfaces` is the **same** `super_interfaces` → `type_list` shape as `record_declaration`. Verified directly.
- `isClassLike` (`scope-resolution/scope/walkers.ts:180-189`) already includes `'Enum'`, and `'Enum'` is already in `LINKABLE_LABELS`, so the `FROM Enum TO Interface` pair is already declared.
- `TYPE_DECL_NODE_TYPES` (`captures.ts:576-580`) already lists `enum_declaration`.

So the correct completion is literally `|| node.type === 'enum_declaration'` at `captures.ts:678`, with no new node-shape handling — **and that one token makes this green test fail.** A future correct fix has to delete a passing test, which is exactly the situation a reader of this test would not expect. Kotlin already emits the equivalent (`enum class Status : Named` → `["Named"]`), and `enum X implements Y` is arguably more idiomatic Java than `record X implements Y`.

I also confirmed there is **no open issue** tracking Java enum heritage, so this test and the `"Enum heritage remains unemitted"` comment at `captures.ts:650-651` currently point at nothing.

**Suggested:** rename to `does not YET emit …`, add a comment saying the fix that emits enum heritage must delete this test, and file the tracking issue — or just widen to `enum_declaration` here and drop the test.

gitnexus/src/core/ingestion/languages/java/captures.ts
```diff
@@ -671,7 +675,7 @@ function synthesizeJavaInheritanceReferences(root: SyntaxNode): CaptureMatch[] {
   const stack: SyntaxNode[] = [root];
   while (stack.length > 0) {
     const node = stack.pop()!;
-    if (node.type === 'class_declaration') {
+    if (node.type === 'class_declaration' || node.type === 'record_declaration') {
```
magyargergo: **P2 · FYI, explicitly not a blocker — two places this change is invisible to. Both are judgment calls for you, not defects.**

**1. Warm parse cache.** This is parse-time capture emission, so its output is baked into the persisted `ParsedFile`. The durable store is gated on `PARSE_CACHE_VERSION` alone — `pipeline-phases/parse-impl.ts:764-771` says so in its own comment ("version-gated by PARSE_CACHE_VERSION (a mismatch ⇒ empty ⇒ every chunk re-dispatches)") — and `--force` does not bypass it. `run-analyze.ts` justifies that reuse as *"safe to reuse across `--force` runs (chunks whose file contents haven't changed produce identical worker output)"*, which is the one assumption this PR changes for Java. A three-run experiment confirms it: HEAD analyzer against a store warmed at the same version yields no heritage and no dispatch; the same analyzer on a fresh store yields both. `SCHEMA_BUMP` is unchanged at 60 (`src/storage/parse-cache.ts:471`; `git diff BASE HEAD -- src/storage/parse-cache.ts` is empty).

**Deliberately not filed as a defect.** `PARSE_CACHE_VERSION` is `` `${SCHEMA_BUMP}+${GITNEXUS_PKG_VERSION}` ``, so any released version bump invalidates the cache and **users upgrading through npm do get the fix**. The project's own recorded practice after debugging this same mechanism on #2730 is *"dev-build only, so don't bump SCHEMA_BUMP for an ordinary capture fix"* — and this is an ordinary capture fix: more captures of an existing kind, no new field, no serialization change, so a warm cache replays stale-but-valid `ParsedFile`s rather than incompatible ones. Exposure is therefore source-checkout / incremental-CI runs at a fixed package version. Raising it only so the choice is deliberate. If you do bump, the ledger at `parse-cache.ts:455-470` says take the next free value above every in-flight **maximum**, not `origin/main + 1`.

**2. The capture-fingerprint gate cannot see this change.** `node --import tsx bench/scope-capture/measure.mjs --check` passes on HEAD with the Java fingerprint byte-identical to `bench/scope-capture/baselines.json:135`, on a PR whose entire purpose is a Java capture-output change. Two causes: the Java scale generator (`measure.mjs:265`) emits only `class Entity${n} extends Base implements Marker` — no record at any scale point — and of the 206 `java-*` fixture files the three that contain a record implement nothing. Consequence: reverting line 678 would delete every Record→Interface IMPLEMENTS edge and, on the measured shape, ~97% of the CALLS edges, and this gate would still report PASS. The repo already treats this class of blind spot as reportable (see the cpp `_rebaselined_2833_generic_member_fields` note in the same baselines file). One record with an `implements` clause in the java scale unit would close it.

*(Item 2's measurements are lane-run, not re-executed by me; item 1 I verified directly.)*

gitnexus/src/core/ingestion/languages/java/captures.ts
```diff
@@ -638,13 +638,17 @@ function findEnclosingTypeDeclaration(node: SyntaxNode): SyntaxNode | null {
...
- * interface-to-interface EXTENDS), matching the legacy Java heritage query
+ * implements clauses), `record_declaration` (`interfaces` implements clauses),
+ * AND `interface_declaration` (`extends_interfaces` → interface-to-interface
+ * EXTENDS), matching the legacy Java heritage query
  * (tree-sitter-queries.ts), which has a dedicated `interface_declaration
  * (extends_interfaces (type_list …))` arm. Without the interface arm the
  * registry-primary synth silently dropped every `interface IA extends IB`
  * edge while the legacy leg emitted it — the exact =0/=N parity break #1951
- * targets. Enum/record heritage stays unemitted (no legacy arm). Generic
+ * targets. Record heritage is emitted even though the legacy query had no
```
magyargergo: **P3 · cross-engine (Claude + GPT-5.6-sol), 5 lanes — this new sentence argues from a query that no longer exists.**

The paragraph reasons in the present tense about a legacy Java heritage query in `tree-sitter-queries.ts` "which has a dedicated `interface_declaration (extends_interfaces (type_list …))` arm", and this PR **adds** a sentence resting on it: *"Record heritage is emitted even though the legacy query had no matching arm."*

That query is gone. `grep -n 'superclass\|super_interfaces\|extends_interfaces'` over `tree-sitter-queries.ts` returns **zero hits**; that file's own header at `:8-11` states heritage is not captured there and now lives in each `languages/<lang>/captures.ts`; and `scope-resolution/pipeline/run.ts:163-164` says the pre-pass is *"the authoritative inheritance emitter"*. The arms were deleted in **`083aedbc41` (PR #2023, #942)**, and the legacy query never had a `record_declaration` arm at all.

The PR didn't introduce the staleness — but it deepened it while the paragraph was open for edit, and the sibling providers already phrase this correctly in the past tense (`csharp/captures.ts:438-440`, `go/captures.ts:321`).

**Suggested:** drop the legacy-query framing and state the positive reason instead — Record graph nodes became canonical link targets in issue #2801, implemented by **PR #2871**. (Worth citing the PR alongside the issue; the comment currently cites only #2801.)

**Also on the next line (`:679`), non-blocking:** the record arm reuses the class arm's `childForFieldName('superclass')`. Per `node-types.json`, `record_declaration` has no `superclass` field (and no `permits`) — correct per JLS §8.10 — so that read is unreachable for records. Harmless and it follows the existing convention at `receiver-binding.ts:86-90`, but the file already states the opposite invariant 85 lines above at `:593-595` (*"enum/record cannot declare an explicit superclass"*), so the two now read as contradicting each other. One comment would settle it.

gitnexus/test/integration/resolvers/java.test.ts
```diff
@@ -1273,6 +1273,40 @@ describe('Java record method resolution (#2564)', () => {
...
+      const fanout = getRelationships(linked, 'CALLS').filter(
+        (edge) =>
+          edge.source === 'read' &&
+          edge.target === 'name' &&
+          edge.rel.reason === 'interface-dispatch',
+      );
+
+      expect(implementsEdge?.sourceLabel).toBe('Record');
+      expect(implementsEdge?.targetLabel).toBe('Interface');
+      expect(fanout.map((edge) => `${edge.targetLabel}:${edge.targetFilePath}`)).toEqual([
```
magyargergo: **P3 · order-sensitive `toEqual` on an array with no ordering guarantee.**

Harmless today — this fixture has exactly one implementor, so the array is 0 or 1 elements. The concern is the next person who adds a second one. `getRelationships` iterates `graph.iterRelationships()` with no explicit sort (two-lane finding, not independently re-run by me), and the sibling multi-implementor block in this same file at `:2189-2199` — same fan-out mechanism, two implementors — deliberately avoids `toEqual` on an unsorted array. `helpers.ts:138-140` already exposes an `edgeSet` helper that sorts first.

Related, and verified directly: the production DFS in `synthesizeJavaInheritanceReferences` is stack-based, so sibling declarations emit in **reverse source order** — `class A implements I1 {} record B(int i) implements I2 {}` yields `["I2","I1"]`. The new `inheritanceRefs` unit helper also asserts exact array equality; both new unit cases use a single declaration, so they are safe as written, but the same trap is waiting there.

**Suggested:** `.sort()` before `toEqual`, or use `edgeSet`.

gitnexus/src/core/ingestion/languages/java/captures.ts
azizur100389: Addressed the benchmark blind spot in 8a7fc6d7: every Java scale unit now includes a Record implementing `Marker`, the fingerprint/count baseline was re-recorded, and the 15-language gate passes (Java scaling 0.973 < 1.5). I intentionally did not bump `PARSE_CACHE_VERSION`: released package-version changes already invalidate the cache, this change preserves the serialized `ParsedFile` shape, and the remaining exposure is the documented fixed-version source-checkout case.

gitnexus/test/unit/scope-resolution/java/java-captures.test.ts
azizur100389: Addressed in 8a7fc6d7 without widening this records-only PR. The test now says `does not yet emit enum heritage`, includes a deletion note, and links #2918, which tracks the one-arm enum heritage follow-up and its end-to-end dispatch acceptance criteria.

gitnexus/test/integration/resolvers/java.test.ts
azizur100389: Addressed in 8a7fc6d7. The interface-dispatch target projection is sorted before exact comparison, and the shared `inheritanceRefs` helper now sorts all synthesized names so future multi-declaration fixtures cannot accidentally depend on DFS or graph iteration order.

gitnexus/test/integration/resolvers/java.test.ts
azizur100389: Addressed in 8a7fc6d7. The positive test now explicitly names the hand-written method shape, and a separate integration test records the canonical `record User(String name) implements Named {}` limitation: the canonical `IMPLEMENTS` edge exists while interface-dispatch fan-out remains empty. Filed #2917 to materialize implicit component accessors as `Method` nodes.

gitnexus/src/core/ingestion/languages/java/captures.ts
azizur100389: Addressed in 8a7fc6d7. The docblock now describes the authoritative registry-primary pre-pass in the present tense, cites #2801 / PR #2871 for Record linkability, and removes the deleted-query framing. Superclass extraction is now class-only; the shared class/record arm handles interfaces with an explicit note that records cannot declare a superclass.

PR #2922, opened 2026-08-10
fix(skills): publish gitnexus-plan artifacts on macOS without an interpreter (#2905)

### LINKED ISSUE #2905: gitnexus-plan safe writer cannot publish plans on macOS
### Area

Other

### Summary

The shipped `gitnexus-plan` safe writer cannot publish generated plans from a normal macOS workspace: native execution is rejected because `write-plan` requires Linux `/proc/self/fd`, while the attempted Linux-container workaround fails `fsync` on the macOS bind-mounted destination.

### Context

This occurred while publishing a completed `gitnexus-plan` artifact for [OptionsWheelTracker #517](https://github.com/dpearson2699/ios-options-wheel-tracker/issues/517). The installed skill and helper are byte-for-byte identical to GitNexus `main`:

- [`.claude/skills/gitnexus-plan/SKILL.md`](https://github.com/abhigyanpatwari/GitNexus/blob/main/.claude/skills/gitnexus-plan/SKILL.md)
- [`.claude/skills/gitnexus-plan/scripts/evidence-provenance.mjs`](https://github.com/abhigyanpatwari/GitNexus/blob/main/.claude/skills/gitnexus-plan/scripts/evidence-provenance.mjs)
- Local/upstream helper SHA-256: `20357a9405b5a31af40f2082c695404d3bb0124821d573117e4bd4fd2b359686`

The original downstream report is [dpearson2699/ios-options-wheel-tracker#534](https://github.com/dpearson2699/ios-options-wheel-tracker/issues/534).

### Expected behavior

The documented `gitnexus-plan` publication workflow should be operable from a supported macOS installation, either natively or through a documented execution path that works with a normal macOS workspace. It should retain the existing path/symlink/race protections, verified atomic publication, and receipt contract.

### Actual behavior

A valid planning job can complete its graph, PDG, source-verification, ledger, and evidence-provenance work but cannot publish its only permitted plan artifact:

1. Native macOS execution fails immediately because `requireDescriptorAnchoring()` rejects every non-Linux process.
2. Running the helper inside Linux Docker against the same bind-mounted repository reaches publication but fails while synchronizing the destination.
3. The skill prohibits direct writing or an external-output fallback, so no plan artifact is produced.

### Steps to reproduce

1. On macOS, install GitNexus and run `gitnexus setup` so `gitnexus-plan` is installed under `~/.agents/skills/`.
2. Compose a valid UTF-8 plan outside the target repository.
3. Run the documented writer:

   ```bash
   node ~/.agents/skills/gitnexus-plan/scripts/evidence-provenance.mjs write-plan \
     --repo "$PWD" \
     --generated-plan docs/plans/2026-08-08-gitnexus-plan-example-change-plan.md \
     < /path/to/outside-repo-scratch-plan.md
   ```

4. Observe the Linux-only descriptor-anchoring refusal.
5. Run the same helper in a Linux Docker container with the macOS repository bind-mounted and observe the `fsync` failure.

### Environment

- OS: macOS 26.5.2 (25F84)
- Node: v26.7.0
- GitNexus: 1.6.10-rc.172
- Installed skill/helper: byte-identical to GitNexus `main` on 2026-08-08
- Container workaround: Linux Docker with the macOS repository bind-mounted

### Logs / screenshots

Native macOS:

```text
Safe generated-plan writes require Linux /proc/self/fd and O_DIRECTORY/O_NOFOLLOW; refusing an unanchored write
```

Linux Docker against the bind-mounted workspace:

```text
EBADF: bad file descriptor, fsync
```

The native refusal corresponds to [`requireDescriptorAnchoring()`](https://github.com/abhigyanpatwari/GitNexus/blob/main/.claude/skills/gitnexus-plan/scripts/evidence-provenance.mjs#L885-L895).

magyargergo:
Fixes #2905.

## The problem

`requireDescriptorAnchoring()` hard-gated `process.platform !== 'linux'`, so on macOS the safe generated-plan writer refused before doing anything. The gate was not incidental: the helper resolves every name through `/proc/self/fd/<fd>/<child>`, and published with `renameat2(RENAME_NOREPLACE)`. macOS has neither.

The issue reports `write-plan`. Two more commands were affected. `read-plan` goes through the same parent walk, so `gitnexus-work` could not load a plan either. `snapshot` fails whenever a materialized path is absent, via `materializeRecord` -> `recordAnchoredAbsence` — deleted files, rename sources and missing cited paths all reach it, which the suite's own fixture does.

## What the platform actually allows

Two things were measured on real runners rather than argued from documentation.

**macOS `/dev/fd/<fd>` is not a substitute for `/proc/self/fd/<fd>`.** With an open directory descriptor on macos-latest:

```
stat /dev/fd/11            ok, isDirectory true
realpath /dev/fd/11        "/dev/fd/11"      <- not the directory's path
lstat  /dev/fd/11/child    ENOENT
open   /dev/fd/11/child    ENOENT
create /dev/fd/11/child    ENOENT
mkdir  /dev/fd/11/child    ENOENT
```

The same probe on ubuntu-latest traverses fine, and after renaming the directory away and planting an impostor at the same name, reading through the descriptor path still returns the original file. That is the anchoring property, and it has no Node-reachable equivalent on macOS: `node:fs` exposes no `dir_fd` parameter anywhere, `fs.Dir` cannot open a child, and there is no FFI. `fcntl(F_GETPATH)` does not help either — XNU reconstructs that string from the name cache at call time and any later use re-walks from the root by name.

**`link(2)` is a portable no-replace publish.** On macOS, Linux, and a 9p mount:

```
link publishes                     "new"
link over an existing destination  EEXIST, destination intact
link onto a symlinked destination  EEXIST, symlink target untouched
same, through /proc/self/fd/<fd>/  identical results, nlink 1 after unlink
```

That is exactly what `RENAME_NOREPLACE` and `RENAME_EXCL` were being used for, and it is reachable from `fs.linkSync`.

## What changed

Publication is now `link(2)` on both platforms, and the writer spawns no interpreter and loads no native code.

That removes the interpreter from Linux entirely, because `/proc` was already doing the resolving there, and removes `ctypes`, `libSystem`, `renameatx_np`, `RENAME_EXCL` and the `ENOTSUP` handling from macOS. Deleted along with them: the trusted-executable validation, the held-descriptor exec and its two-tier probe, the capability probe, the JSON request protocol, and both embedded Python programs. The helper goes from 3047 to 2327 lines.

macOS keeps the part that genuinely cannot be done in Node, and now does it in-process: a lexical `O_NOFOLLOW` walk that holds an open descriptor on every directory in the chain for the operation's lifetime, and re-proves the chain either side of every step. The pinning is load bearing rather than tidy — an open descriptor keeps its inode number from being recycled, and without it a freed directory's inode can be reclaimed by the next `mkdir` and satisfy a recorded identity. That was reproduced during review, first try, on ext4.

## The guarantees are not symmetric, and the docs say so

On Linux a parent swapped between a check and its use cannot redirect the operation, because names are resolved against the descriptor rather than re-walked. On macOS the same swap is detected by the check that follows and the operation aborts having written nothing — but it is detection rather than prevention. `references/evidence-provenance.md` states this in those words, under a heading that says it.

## Verification

The suite had been silently skipping on every non-Linux runner (`process.platform === 'linux' ? describe : describe.skip`), which is why this shipped. It now admits darwin and is registered in `scripts/cross-platform-tests.ts`, so it runs on the `macos-latest` matrix.

Because the backend no longer needs an interpreter or `/proc`, it is fully portable — spoofing `process.platform` exercises the real macOS code path on Linux CI instead of leaving it unrun. Three tests were added on that basis: a full write/read round trip under a spoofed darwin that asserts no `/proc/` path reaches the test hooks (proving the Darwin backend was actually selected), the `afterParentOpen` rename-and-symlink attack refused through the pinned chain, and a publish test asserting the published file is the same inode as the temporary with no `.tmp` left behind and a symlinked destination refused.

154 tests green locally. 29 fixtures that drove the deleted Python program directly were removed with it.

## Also fixed

The writer failed on 9p mounts every time, because `renameat2(RENAME_NOREPLACE)` returns `EINVAL` there. `link(2)` succeeds, so publication now works on WSL2 9p checkouts.

## History note

This PR started with a Python-based `openat` backend for macOS, which is why the earlier commits describe one. Two adversarial review passes and four quality passes ran against that design and their findings are folded in — the inode-recycling pin, in-backend identity proof for descriptor-returning operations, and the Windows realpath fix in `assertRepository` (`snapshot` is not platform-gated, and 8.3 short names made it reject a worktree root by quoting that same directory back). The interpreter was then removed entirely in favour of `link(2)`.

Out of scope, noted for the record: `eval/workflow_bench/promotion_apply.py` carries the same Linux-only `renameat2` and `/proc/self/fd` pattern, but it is a separate program, not a caller of this helper.

Thanks to @koriyoshi2041 for confirming the reproduction on native macOS and for pinning down that the blocker was the descriptor-path backend rather than missing open flags.

https://claude.ai/code/session_01NtGZN6YSNU8chALSnKYHBp


koriyoshi2041 (changes requested):
The Darwin backend is currently unusable because `darwinHardenedOpen()` ORs `O_NOFOLLOW_ANY` into flags that already contain `O_NOFOLLOW`. XNU rejects that combination with `EINVAL`.

I reproduced this on native macOS with Node 26.5.0 and Python 3:

```text
open(realpath('/tmp'), O_RDONLY|O_DIRECTORY|O_NOFOLLOW)              -> success
open(realpath('/tmp'), O_RDONLY|O_DIRECTORY|O_NOFOLLOW_ANY)          -> success
open(realpath('/tmp'), O_RDONLY|O_DIRECTORY|O_NOFOLLOW|O_NOFOLLOW_ANY) -> EINVAL
```

This matches the 27 failures in the macOS 2/3 shard: every operation stops while opening the repository root, before the anchored helper is exercised. `darwinHardenedOpen` needs to replace/clear `O_NOFOLLOW` when applying `O_NOFOLLOW_ANY` (with the desired compatibility policy made explicit), rather than combine them. Please rerun the native macOS helper suite after that change; the Linux and Windows lanes are already green.

magyargergo:
@koriyoshi2041 can you please test this on MacOS? 🙏

koriyoshi2041 (approved):
Tested the current head `8327cff5` on native macOS (Darwin 25.6.0, arm64, Node 26.5.0). The focused evidence-provenance helper suite now passes 58/58.

The earlier `EINVAL` blocker is resolved: directory opens no longer combine `O_NOFOLLOW` with `O_NOFOLLOW_ANY`, and the new regression explicitly rejects `O_NOFOLLOW_ANY` on every directory/file open. I also checked the follow-up boundary changes around single-component validation and hard-link publication; they preserve fail-closed behavior rather than weakening the anchoring contract.

This addresses my changes-requested finding on the prior head.

Issue #2873, opened 2026-08-07
Go: a signature naming any type from outside the repository (`context.Context`, any stdlib/third-party type) is treated as *incompatible*, so IMPLEMENTS is never emitted and `impact()` returns a confident `0`


corbettjouett00-sudo:
Follow-up to #2766 / #2813. This is a self-contained repro against `1.6.10-rc.174`, and I think it explains the "inconsistent at scale" behaviour I reported in #2813 without any of the ambiguity that report had. The interface-typed-field story turned out to be a symptom; the cause is in signature normalization.

## TL;DR

`qualifyGoSignatureTypes` returns `undefined` when a package qualifier in a signature has no entry in `importQualifiers`. `signatureContextForFile` only populates `importQualifiers` from import edges whose `targetFile !== null` — i.e. imports that resolve to a file **inside the repository**. Every stdlib and third-party import is therefore absent, so any signature mentioning `context.Context`, `time.Time`, `sql.Tx`, `http.Request`, … normalizes to `undefined`, and both `parameterTypesCompatible` and `returnTypesCompatible` treat `undefined` as *not compatible*.

`ctx context.Context` is the first parameter of essentially every idiomatic Go method, so in practice Go interface satisfaction only succeeds for the small minority of interfaces whose signatures are builtin-only.

## Repro

Two files, no dependencies. `go vet` clean.

`go.mod`
```
module example.com/gnrepro

go 1.22
```

`repro.go`
```go
package repro

import "context"

// ---- Control: signature uses only in-repo/builtin types. IMPLEMENTS is emitted.
type LocalPayload struct{ ID string }

type LocalStore interface {
	Get(p LocalPayload) error
}

type LocalStoreImpl struct{}

func (l *LocalStoreImpl) Get(p LocalPayload) error { return nil }

// ---- Case A: one parameter type comes from the stdlib. No IMPLEMENTS.
type CtxStore interface {
	Delete(ctx context.Context, id string) error
}

type CtxStoreImpl struct{}

func (c *CtxStoreImpl) Delete(ctx context.Context, id string) error { return nil }

// ---- Case B: stdlib type in the RETURN position only. No IMPLEMENTS.
type RetStore interface {
	Ctx() (context.Context, error)
}

type RetStoreImpl struct{}

func (r *RetStoreImpl) Ctx() (context.Context, error) { return nil, nil }

// ---- Consequence: a call through an interface-typed field never reaches the impl.
type Handler struct {
[...]

PR #2926, opened 2026-08-11
feat(impact): tell an undecided interface check apart from a decided negative (#2873)

magyargergo:
Completes #2873. **Stacked on #2921** — review that first; this branch contains its commits.

#2921 removed the dominant *cause* of a wrong-and-confident answer. This one removes the *confidence*: the analyzer now distinguishes "no type implements this interface" from "we could not tell", and carries the second all the way to `impact`.

## Why a graph probe cannot fix this

The existing hedge (`computeEpistemicBoundary`) finds a dispatch boundary by counting IMPLEMENTS edges. An undecided check mints no edge — so there is nothing left for it to notice. The evidence a graph probe would need is precisely what went missing. That is why the reported symptom is a *confident* zero and not merely a low one:

```
impact --uid "Method:repro.go:CtxStoreImpl.Delete"
{ "impactedCount": 0, "risk": "LOW", "epistemic": "exact" }
```

## The tri-state

`satisfied | unsatisfied | undecided`, propagated from the one place a type spelling decides anything up to the candidate loop.

| verdict | edge | recorded |
| --- | --- | --- |
| satisfied | IMPLEMENTS | — |
| unsatisfied | none | none — a decided "no" is trustworthy |
| **undecided** | **none** | **yes** |

`undecided` deliberately mints no edge. A speculative IMPLEMENTS would fan out through `emitReceiverBoundCalls` into fabricated CALLS, and wrong callers in a blast-radius tool are their own kind of lie. Note `go/types` folds the same case the *other* way — `hasAllMethods` returns `true` for an invalid type, commented *"we don't know anything about V, assume it implements T"* — because a type checker's job is to avoid cascading errors, not to bound a change. Soot is conservative upward (phantom → true) but its *downward* `getSubclassesOf` has no such guard, which is exactly the "who implements this" direction `impact` asks in.

Two propagation rules keep unknown from swallowing everything:
- **A decided mismatch dominates.** A type that provably lacks a required method does not implement the interface however many other methods went unread.
- **One matching overload settles its method.** Unknowns on the siblings cannot unsettle it — the same rule Pyright uses to suppress partially-unknown when a resolvable path exists.

Two pre-existing assumptions surfaced on the way. An absent parameter-type list read as **compatible** — a bare guess that could mint an edge between signatures nobody compared; it is now `undecided`, except where both sides declare zero parameters and absence is the honest shape of `Close() error`. A required method with no verifiable signature read as *does not match*; that is missing information too.

## Where the record lives

`RepoMeta`, beside `unresolvedReceiverMembers`, for its reasons: the fact is about the **analysis** rather than the code; it is additive metadata so `SCHEMA_FINGERPRINT` does not move and no re-analyze is forced; and it stays invisible to all nine IMPLEMENTS consumers (MRO, METHOD_IMPLEMENTS derivation, dispatch fan-out, clustering, DI, Spring AOP, the impact walk, the retention sink) — none of which should see a question as an edge.

Both sides of each pair are recorded. Asking about the **implementation** is the case the issue reports, and an interface-keyed record alone leaves exactly that query unhedged: the walk starts at a method whose owner has no heritage edge. A method matches through its declaring type — one Cypher hop, asked only for methods.

## What `impact` now returns

A fourth `causes` slot. The three that existed count decided facts that could not be *attributed*; this one counts questions never *answered*, and it is the only cause that shortens a result without leaving a trace in the graph:

```jsonc
{ "impactedCount": 0,          // unchanged — this invents no callers
  "epistemic": "lower-bound",  // was "exact"
  "causes": { "receiverTyping": 0, "dispatchBoundary": 0,
              "externalBoundary": 0, "undecidedSatisfaction": 1 },
  "boundaries": ["`CtxStoreImpl` was a candidate implementation for 1 interface the analyzer could not decide, …"] }
```

A cleanly-decided symbol in the same index still reports `exact`. The hedge has to stay rare to mean anything — Pyright's `Unknown` is always tracked but its diagnostics default to off outside strict mode, for the same reason.

## Performance

Measured best-of-N at three sizes, against this same branch with the tri-state reverted:

| pairs | before | after |
| --- | --- | --- |
| 200×200 = 40k | 94–95 ms | **87–89 ms** |
| 800×800 = 640k | 2367–2767 ms | **2140–2357 ms** |

Faster, because a decided parameter mismatch now short-circuits before the return-type comparison. (The suite's `scaling < 1.5` gate fails identically with and without these commits on this machine — it is not a regression introduced here.) The query side adds one metadata read, already loaded on the same path, plus one gated Cypher hop for methods only.

## Tests

- Unit: the undecided channel fires for an unresolvable qualifier, stays silent for a decided mismatch, and a hard "no" beats an unknown in the same method set.
- Unit: the summary records both sides, keeps true totals beside the capped window, and ranks deterministically (the tiebreak decides which entries survive the cap).
- Integration, end to end: a dot-import fixture — no qualifier at all — reports `Drawer` undecided with `Canvas` as its candidate, while `Label → Named` still emits in the same file.
- Integration, query side: `impact` on the implementation method returns `lower-bound` with `undecidedSatisfaction: 1`; on the interface, `2`; a clean leaf in the same index stays `exact`.

Go integration 182/182, all scope-resolution units 1803 passing, tsc clean.

## Follow-up, not in this PR

Go publishes `api/go1*.txt` — every exported stdlib symbol including full interface method sets, generated and enforced by the Go release process. Consuming it would turn "does `MyType` implement `io.Reader`" from undecidable into decided, shrinking the unknown set to genuine third-party interfaces. Complementary to this change, not a replacement.

https://claude.ai/code/session_01JL16XnB3J6L7D7DTaaqdr8


magyargergo:
Folded into #2921 — the three commits are now on that branch, and its description covers both halves of #2873 (the qualifier fix and the undecided/decided distinction). Closing to keep the review in one place.

PR #2930, opened 2026-08-11
fix(mcp): stop scaling the detect_changes query with the diff's hunk count (#2915)

### LINKED ISSUE #2915: detect_changes: silent SIGBUS (exit 132) in @ladybugdb/core when a diffed file has ~4,000+ hunks — per-hunk OR conditions overflow evaluator copy stack
## Summary

`detect_changes` hard-crashes the Node process (SIGBUS in `lbugjs.node`, surfaces as shell exit 132) whenever any single diffed file contains roughly **4,000+ hunks** with `-U0`. Real-world trigger: a machine-generated file (a graph-tool cache JSON) that was tracked in git and diffed at ~4,700 hunks after regeneration. `impact`, `query`, `context`, and `analyze` are all unaffected.

Related but distinct from #2233: that was the FTS delete/update path in `@ladybugdb/core@0.17.1` and was closed by the 0.18.x bump. This one reproduces on current `gitnexus@1.6.9` with `@ladybugdb/core@0.18.1`, and the faulting native path is expression copying, not FTS.

## Environment

```text
OS: macOS 26.5 (Darwin 25.5.0), arm64 (M4)
Node: reproduces on both v22.23.1 and v26.4.0
GitNexus: 1.6.9 (npx cache install)
@ladybugdb/core: 0.18.1 (lbugjs.node darwin-arm64)
```

## Minimal repro

```bash
mkdir dc-repro && cd dc-repro && git init
seq 1 20000 | sed 's/^/line /' > big.txt
printf 'def hello():\n    return 1\n' > code.py
git add -A && git commit -m init
npx gitnexus analyze

# change every 4th line -> ~5,000 hunks with -U0
awk 'NR % 4 == 0 {print $0 " changed"} NR % 4 != 0 {print}' big.txt > t && mv t big.txt
git diff -U0 | grep -c '^@@'    # 5000

npx gitnexus detect-changes --repo dc-repro; echo "exit=$?"   # exit=132
```

Threshold bisect on this machine (same repo, changing every Nth line):

| hunks | result |
|-------|--------|
| 2500  | ok |
| 3333  | ok |
| 4000  | crash (132) |
| 5000  | crash (132) |

## Crash stack

macOS writes two `.ips` reports at the same second. The primary one is a Bus error on a LadybugDB TaskScheduler worker thread — `FunctionExpressionEvaluator::copy()` recursing until the (512KB) thread stack overflows:

```text
Bus error: 10
  lbugjs.node  lbug::evaluator::FunctionExpressionEvaluator::copy()   x7 frames
  lbugjs.node  lbug::processor::Filter::copy()
  lbugjs.node  lbug::processor::Projection::copy()
  lbugjs.node  lbug::processor::ResultCollector::copy()
  lbugjs.node  lbug::processor::ProcessorTask::run()
  lbugjs.node  lbug::common::TaskScheduler::runWorkerThread()
```

The second report ("Illegal instruction: 4", `node::Kill` under a `PromiseRejectReactionJob`) is just the CLI process dying afterwards — which also means the user sees a bare exit 132 with **no error output at all**.

## Root cause

`detectChanges` in `dist/mcp/local/local-backend.js` builds **one OR'd range condition per diff hunk** into a single `WHERE` clause per file:

```js
const overlapConditions = fileDiff.hunks
    .map((_, i) => `(n.startLine <= $hunkEnd${i} AND n.endLine >= $hunkStart${i})`)
    .join(' OR ');
```

With thousands of hunks the resulting expression tree is deep enough that LadybugDB's recursive expression copy blows the worker-thread stack.

## Suggested fixes (any one suffices)

1. **Coalesce hunk ranges before building conditions.** Sort hunks and merge adjacent/near-adjacent ranges (e.g. gap ≤ 2 lines). For generated files — the realistic way to get thousands of hunks — this collapses the condition count by orders of magnitude, and a slightly widened range only risks including a neighboring symbol row, which is harmless for this use case.
2. **Chunk the per-file query** into batches of a few hundred conditions and union the results (the BasicBlock-exclusion rework in #2082 already touches this query).
3. Longer-term: iterative `copy()` upstream in LadybugDB — but GitNexus shouldn't hand the engine unbounded expression depth regardless.

Also worth catching the engine failure so the CLI reports *something* instead of a silent exit 132.

## Workaround

Untrack the offending generated file(s) (`.gitignore` + `git rm --cached`) — hunk counts drop back to code scale and every scope works again.

magyargergo:
Fixes #2915.

## Problem

`detect_changes` built one OR'd condition pair per diff hunk into a single `WHERE` clause, one query per changed file:

```js
const overlapConditions = fileDiff.hunks
  .map((_, i) => `(n.startLine <= $hunkEnd${i} AND n.endLine >= $hunkStart${i})`)
  .join(' OR ');
```

A machine generated file (a cache JSON, a lockfile, a golden fixture) diffs at thousands of hunks with `-U0`, and the expression tree that produces is deep enough to overflow LadybugDB's recursive evaluator copy on a TaskScheduler worker thread.

Reproduced on `main` with `@ladybugdb/core` 0.19.1. On Linux the engine gets an 8 MB thread stack, so I ran it with macOS sized stacks (`ulimit -s 512`, glibc uses RLIMIT_STACK for pthread defaults) and the threshold lands on the reporter's numbers:

| hunks | exit |
|-------|------|
| 1000 / 2000 / 2500 | 0 |
| 3333 | 139 (SIGSEGV) |
| 4000 / 5000 | 139 |

A one hunk diff under the same `ulimit` exits 0, so the crash follows the hunk count, not the stack limit. On a roomier stack the same query does not crash but goes quadratic (5,000 hunks 12s, 10,000 hunks 65s, 20,000 over 2 minutes), tripping the 30s query timeout. That path is worse than the crash: the failure is swallowed, the result comes back `partial: true` with `changed_count: 0`, and the CLI printed `No changes detected.` and exited 0 while missing a real edit in the same diff.

## Fix

The overlap test no longer goes into Cypher at all.

* `coalesceHunks(hunks)` sorts a file's hunks and merges only the ones that overlap or abut (`next.startLine <= current.endLine + 1`), so the union covers exactly the lines the raw hunks covered. No widening, no gap heuristic.
* `hunksOverlapRange(coalesced, start, end)` binary searches those disjoint ranges. Matching happens in JS on rows the query returns.
* The per file query became a batched one, with a per file `[lo, hi]` span:

```cypher
UNWIND $bounds AS b
MATCH (n) WHERE (n.filePath = b.path OR n.filePath ENDS WITH b.suffix)
  AND NOT n.id STARTS WITH 'BasicBlock:'
  AND n.startLine IS NOT NULL AND n.endLine IS NOT NULL
  AND n.startLine <= b.hi AND n.endLine >= b.lo
RETURN b.path AS diffPath, n.id AS id, ...
```

Query text and parameters are now identical whether a file changed in 1 place or 100,000. The same repro repo, still under `ulimit -s 512`, maps a 100,001 hunk diff in 2.1s.

The span matters on its own: without it the query returned every symbol of every changed file and filtered them in JS. Measured on a 25k node index, 400 file batch: **546ms / 13,870 rows before, 84ms / 1,555 after**, with the kept set verified identical. Depth stays constant (two comparisons per file, not per hunk), so this cannot revive the crash, and `hunksOverlapRange` still rejects symbols landing in the gaps between hunks. The struct list parameter was probed against `@ladybugdb/core` 0.18.3 and 0.19.1 before the rewrite (the index subscript form `$paths[i]` does not parse).

Batches run a few at a time. `executeParameterized` checks a connection out of the 8 connection per repo pool for the duration of a query, so parallel calls never share one, which is why ~15 other queries in this file already run under `Promise.all`. `allSettled`, so one failed batch degrades the result to `partial` instead of discarding the batches that succeeded beside it.

## Three more bugs in the same path

**Line base.** Graph rows are 0-based (#2377) and git hunk lines are 1-based, and the overlap compared them raw. Every symbol was shifted one line up, so an edit to a symbol's last line reported nothing changed. On the shipped build, for a two line function:

```
edit line 1 (def hello():)  -> Changes: 1 files, 1 symbols
edit line 2 (return 42)     -> No changes detected.
```

Both edits are inside `hello`. Grouping and the conversion now happen once in `coalesceHunksByPath`, at the parse boundary: `parseDiffHunks` stays faithful to git, consumers compare graph native values.

**Path matching.** `ENDS WITH` is a plain string suffix, so a diff touching `lib/a.py` also matched an indexed `src/mylib/a.py` and reported a symbol from a file the diff never touched. Proven against a real index:

```
bare ENDS WITH  ["a@lib/a.py","b@src/mylib/a.py"]
anchored        ["a@lib/a.py"]
```

The match is anchored on the separator, with an equality arm for a path that is the indexed value. This is the form `explain` already uses.

**Double counting.** One node matched by two changed paths was pushed once per match (169 duplicates in 13,870 rows on a real 400 file diff), and `changed_files` counted a path twice when a diff reported it twice (a rename alongside an edit). Both deduplicated.

## Reporting honesty

`formatDetectChangesResult` never read the `partial` flag the backend sets when a graph query is swallowed (#2283), so a degraded run rendered as a clean pre-commit check. It now prints the note in both branches, and both `partial` and the new `symbols_truncated` are documented in the `detect_changes` tool description, since the MCP client is the main consumer of this gate and was getting them as undocumented raw keys.

The listing is capped at 1,000 symbols with `symbols_truncated: {listed, total}` — a repo wide diff otherwise puts an unbounded array in one MCP payload, where the CLI has `--limit` and an MCP client has nothing. Counts are never capped, so the risk level and the CLI's "... and N more" still see the true total.

## The same shape elsewhere

`detect_changes` was not the only place building query text from a caller sized array. `core/wiki/graph-queries.ts` interpolated a module's whole file list into four `IN [...]` literals. Flat breadth rather than nested depth, but the same unbounded shape, and the one the repo's own `DELETE_FILES_CHUNK_SIZE` precedent already chunks elsewhere. All four are batched now, with the membership arms handled carefully and documented where it happens — in particular `getInterModuleCallEdges` keeps the same batch list in its `NOT` arm, because `NOT null IN [...]` is null and a JS only filter would have started admitting nodes with no `filePath`.

`chunk` / `mapBatches` / `LBUG_QUERY_BATCH_SIZE` now live in `core/lbug/query-batch.ts` with the measured batch size recorded where it is defined, replacing three same valued constants under three names and the hand rolled slice loops in `lbug-adapter.ts`, `embeddings/http-client.ts` and `run-analyze.ts`. No batch size changed.

`toOneBasedLine` joins `toZeroBasedLine` in `ingestion/utils/line-base.ts`, so that module owns both directions, and the four CFG/PDG join sites plus two backend sites call it instead of open coding `+ 1`. It is deliberately not `toDisplayLine`, which is a response boundary converter with an `undefined` passthrough; the `Number.NaN` guards for an absent line are kept verbatim.

## Tests

* `test/unit/detect-changes-hunk-scale.test.ts` — 25 tests. Drives the real `detect_changes` against a real temp git repo with the query layer mocked, so it inspects the exact Cypher the engine would receive: query text identical for a 1 hunk and a 3,000 hunk diff, `bounds` carrying 0-based spans and anchored suffixes, 250 files producing 3 queries rather than 250, the last line edit reported, a symbol one line above the hunk not reported, the dedup, and the listing cap with true counts.
* `test/integration/detect-changes-path-anchoring.test.ts` — the same tool against a real engine, seeded with `lib/a.py` and `src/mylib/a.py`. Fails 3/3 with the un-anchored predicate.
* `test/unit/query-batch.test.ts`, `test/unit/line-base-conversion.test.ts`, `test/unit/wiki-graph-queries-batching.test.ts` — the new helpers, including the wiki batching oracle running the pre-#2915 single query text against the same fixture.

Verified the behaviour tests fail with the old implementation restored. `tsc --noEmit`, eslint and prettier clean. The PDG/CFG/taint/impact suites (96 files, 1,973 tests) pass unchanged, as do the delete/importer/embedding suites whose assertions pin the batch progress and chunk index values.

---

## Summary

*A broad, cross-cutting GitNexus change focused on `detect_changes`, with implementation work alongside updates to operational documentation and skills. Its reach is wide within the repository, but the graph shows no outward dependent chain.*

**🔴 CRITICAL blast radius. A broad GitNexus change centered on `detect_changes`, spanning the MCP local backend, CLI and core code, documentation, and skill files across 80 changed files.**

The implementation hotspot is `gitnexus/src/mcp/local/local-backend.ts`, while related changes extend through the `Cli`, `Local`, `Wiki`, `Extractors`, `Ingestion`, and `Embeddings` modules. The report traces 26 affected flows, so review the local backend first, then the CLI files that format and expose `detect_changes`, including `gitnexus/src/cli/detect-changes-format.ts` and `gitnexus/src/cli/tool.ts`.

The graph found no dependents beyond the changed symbols, and every changed file is classified LOW risk with no HIGH or CRITICAL risk files. Given the PR title, verify the query behavior and its handling of diff hunk counts in the MCP and CLI paths before reviewing the broader documentation and skill updates.

_Added by GitNexus for PR #2930. Edit freely — this block is replaced on the next review, everything above it is left untouched._


koriyoshi2041:
I checked the failed Ubuntu shard against head `6a1a41e7`. The only failure is the unrelated `embedding-runtime-support.test.ts` optional-stack probe (`isLocalEmbeddingStackInstalled()` returned false after `npm ci`); this PR does not touch the embedding runtime or dependency manifests.

On the exact PR head, the failed test passes locally together with the focused `detect_changes`/path-anchoring/query-batching/line-base suites: 7 files, 98 tests passed. The hunk-scale test also confirms identical query text at 1 vs 3,000 hunks, and the real-engine anchoring regression passes. This looks like an optional-dependency install/cache miss in that runner rather than a #2915 regression; a rerun should be a useful discriminator.

gitnexus/src/cli/detect-changes-format.ts
```diff
@@ -35,11 +37,18 @@ export function formatDetectChangesResult(result: unknown): string {
   if (payload.error) return t('common.error', { message: String(payload.error) });
 
   const summary = payload.summary ?? {};
+  // A swallowed query failure sets `partial` and leaves the counts at zero
+  // (#2283). Printing only "No changes detected." turns a degraded run into a
+  // clean bill of health for the pre-commit gate, so say so either way.
+  const partialNote = payload.partial ? t('tool.detectChanges.partial') : null;
```
magyargergo: **Blocking** — the PARTIAL text landed here, but the exit status did not, so this only half-closes the #2915 secondary bug.

`detectChangesCommand` (`src/cli/tool.ts:341`) does:

```ts
output(formatDetectChangesResult(result));   // a STRING
```

and `output()` (`tool.ts:46-72`) sets a failing code only for an object:

```ts
// ... this is the one place that keeps scripted callers honest
if (data && typeof data === 'object' && 'error' in data && ...) process.exitCode = 1;
```

A string never matches, so `detect-changes` opts itself out of the only mechanism that file exists to provide. It is broader than `partial`: line 37 above converts a backend `{error}` payload into text too, so a hard failure also exits 0. `gitnexus detect-changes && git commit` still proceeds on a degraded analysis exactly as before.

`checkCommand` twenty lines below already does the object-first pattern.

Fix: set `process.exitCode = 1` in `detectChangesCommand` when the structured result carries `partial` or `error`, before formatting — and add a direct CLI exit-status test. Anchored here because `tool.ts` is not in this diff.

Verified by inspection.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -5150,72 +5169,100 @@ export class LocalBackend {
...
+    // The path match is anchored on the separator (see path-predicate.ts): a
+    // bare ENDS WITH is a plain string suffix, so 'lib/a.ts' also matched an
+    // indexed 'src/mylib/a.ts'. The [lo, hi] span lets the engine drop symbols
+    // outside the file's touched region instead of shipping every row in the
+    // file across the native boundary — two comparisons per FILE, not per hunk,
+    // so #2915 cannot come back, and `hunksOverlapRange` below still rejects
+    // the gaps between hunks.
+    const symbolQuery = `
+        UNWIND $bounds AS b
+        MATCH (n) WHERE (n.filePath = b.path OR n.filePath ENDS WITH b.suffix)
```
magyargergo: **Blocking** — this predicate cannot filter the scan, so the query materialises the whole node set.

`EXPLAIN` on the live index, with the shipped query text:

```
SCAN_NODE_TABLE[0]  (33 tables)
  FILTER[1] NOT(STARTS_WITH(n.id))  FILTER[2,3] IS_NOT_NULL(startLine/endLine)
  RESULT_COLLECTOR[4]          <- build side: the ENTIRE filtered node set, materialised
        \
  CROSS_PRODUCT[7] --/ UNWIND[6] ($bounds)
  FLATTEN[8] -> FILTER[9,10] STRUCT_EXTRACT(b)   <- this path match
             -> FILTER[11]   STRUCT_EXTRACT(b)x2 <- the [lo, hi] span
```

Only the three `n`-only predicates push below the accumulate. Both predicates that reference `b` sit above the cross product, so neither the path match nor the span can reduce the scan or the materialised build side. `schema.ts` declares only `PRIMARY KEY (id)`, so there is no `filePath` index to change that.

Cost at 1M nodes, pool pre-warmed: old 400 sequential queries +67 MB; new one batch +242 MB; new 4 concurrent +922 MB — paid even for a one-file diff. Differential failure, same DB and workload, varying only the buffer-manager cap:

```
cap=268MB   OLD 400 sequential     OK 5,494 ms
cap=268MB   NEW 4x100 concurrent   FAILED  Buffer manager exception
cap=537MB / 805MB                  FAILED
cap=1074MB  NEW 4x100 concurrent   OK 432 ms
```

That exception is caught by `mapConcurrent`'s `onError` into `partial:true, changed_count:0` — the #2915 false clean by a new route, and per the formatter comment, exit 0.

Fix (measured: 10x less memory, ~20% faster, identical result sets) — add the batch-wide, `b`-free disjunction as a redundant conjunct so the planner can push it below the accumulate. It is a superset of the correlated predicate, so it cannot reject a row the correlated one keeps:

```cypher
MATCH (n) WHERE (n.filePath IN $paths OR ANY(s IN $suffixes WHERE n.filePath ENDS WITH s))
  AND (n.filePath = b.path OR n.filePath ENDS WITH b.suffix) AND ...
```

At 1M nodes: 1 bound 137 ms/+237 MB -> 110 ms/+13 MB; 100 bounds 325 ms/+247 MB -> 258 ms/+23 MB; 4x100 at a 268 MB pool FAILED -> OK. Cheaper partial: `AND n.filePath IS NOT NULL` is semantically inert but drops every `CodeEmbedding` row, which has `startLine`/`endLine` and no `filePath` and so passes all three pre-accumulate filters today.

Separately, the anchor still admits whole-segment siblings: `'pkg/README.md' ENDS WITH '/README.md'`, so editing a root `README.md` returns symbols from `eval/README.md`, `gitnexus/README.md` and others (reproduced twice on a real engine). Strictly narrower than the old bare `ENDS WITH`, so this is an improvement that stops short. A `detect_changes` path is always repo-root-relative because it comes from a `+++ b/` header, so `n.filePath = b.path` alone is the correct predicate and the suffix arm only papers over an index root that differs from the git root.

Reproduced.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -5260,18 +5307,32 @@ export class LocalBackend {
             ? 'high'
             : 'critical';
 
+    // A repo-wide diff can touch thousands of symbols, and the whole array goes
+    // into one MCP payload (the CLI slices with --limit; an MCP client has no
+    // such control). Cap the LISTING, never the counts: `changed_count` stays
+    // the true total, so the risk level, the CLI's "... and N more" line and any
+    // client comparing the two still see the real number. `truncated` is the
+    // key `explain`/`pdg_query`/`trace` already use for a capped window.
+    const listedSymbols = Array.from(changedSymbols.values()).slice(
```
magyargergo, line 5463: **Blocking** — this slices a nondeterministic row order, which is the #2787 class this same PR fixes 200 lines away.

`symbolQuery` has no `ORDER BY`, and `changedSymbols` is a `Map` fed straight from the rows, so its iteration order *is* the engine's row order. Against a copy of the real 25,342-node index with the exact shipped query, 100 bounds, 2,133 rows per run: **5 distinct row orders across 8 runs on one connection**, and a 6th across fresh connections. Two `detect_changes` calls on an unchanged tree therefore return different `changed_symbols`.

Meanwhile `graph-queries.ts:258-260` in this PR adds a JS sort with the comment *"an unordered cut keeps a different subset per machine (#2787)"*.

`changed_count` stays correct — only the listing drifts — which is what keeps this contained.

Fix: sort before slicing (`filePath, startLine, id`), which is cheaper here than an engine `ORDER BY` since the rows are already materialised.

Reproduced.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -901,6 +907,15 @@ export function resolveWorktreeCwd(repoPath: string, launchCwd: string): string
...
+ * Changed symbols listed in one `detect_changes` result.
+ *
+ * The counts in `summary` are never capped — only the array is, so a repo-wide
+ * diff cannot put an unbounded payload on the MCP wire. A truncated result says
+ * so in `symbols_truncated`.
+ */
+const DETECT_CHANGES_MAX_LISTED_SYMBOLS = 1000;
+
 export function buildDetectChangesDiffArgs(scope: string, baseRef?: string): string[] | null {
   const args = ['diff', '--ignore-cr-at-eol'];
```
magyargergo: **Worth closing while you are here** — one line of user gitconfig makes every run print "No changes detected." and exit 0, before any query is issued.

No `--src-prefix`/`--dst-prefix`/`--default-prefix`/`--no-ext-diff` here, so git honours the user's config. Measured on git 2.49.0 for a one-line edit:

| config | `+++` line |
|---|---|
| default | `+++ b/f.ts` |
| `diff.noprefix=true` | `+++ f.ts` |
| `diff.mnemonicPrefix=true` (unstaged) | `+++ w/f.ts` |
| `diff.mnemonicPrefix=true` (`--staged`) | `+++ i/f.ts` |

`parseDiffHunks` gates on `line.startsWith('+++ b/')` (`git.ts:691`), so it returns `[]`, the `fileDiffs.length === 0` early return at `:5159-5170` fires, and that branch emits `risk_level:'none'` + `'No changes detected.'` and **can never carry `partial`** — the failure happens before any query runs, so none of this PR's new degradation machinery covers it.

Pre-existing (`parseDiffHunks` is untouched), raised because it is the highest-severity live instance of exactly the class this PR is about and it is one line to close.

Fix: `['diff','--ignore-cr-at-eol','--no-ext-diff','--src-prefix=a/','--dst-prefix=b/', ...]`. Use the src/dst form rather than `--default-prefix`, which needs git >= 2.42. Independently: when `diffOutput.length > 0` but `fileDiffs.length === 0`, that is a parse failure — raise `partial: true` rather than taking the `'none'` branch.

Reproduced.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -5150,72 +5169,100 @@ export class LocalBackend {
...
       }
     }
 
     // Find affected processes -- single batched query instead of N+1
     const affectedProcesses = new Map<string, any>();
-    if (changedSymbols.length > 0) {
-      const symIds = changedSymbols.map((s) => s.id);
-      const symNameById = new Map(changedSymbols.map((s) => [s.id, s.name]));
+    if (changedSymbols.size > 0) {
+      const symIds = Array.from(changedSymbols.keys());
```
magyargergo: This is the one caller-sized list the PR did not chunk.

`symIds` is the full deduped set and goes straight into `WHERE n.id IN $ids` at `:5271`, with no `chunk()` and no cap, while every other list here was batched at `LBUG_QUERY_BATCH_SIZE`. Measured on the real index with the same query shape: 25,000 ids -> 442 ms / 721 MB; 100,000 -> 599 ms / 1,238 MB; **500,000 -> 2,595 ms / 4,002 MB**. Trigger: `detect_changes({scope:'compare', base_ref:<root commit>})`, or a repo-wide codemod with `scope:'all'`.

The new module states the rule this misses (`query-batch.ts:9-20`: bind *or* chunk). Binding keeps the query *text* constant, which is all `warnIfQueryTextUnbounded` measures — the bound *value* is still repo-sized, so the guard is structurally blind to this shape.

Related: the docstring at `:911-913` says a repo-wide diff "cannot put an unbounded payload on the MCP wire", but this query, `affected_processes` (`:5331`) and each entry's `changed_steps` (`:5289`) are all driven by the full `changedSymbols`, not the capped 1,000.

Fix: `for (const ids of chunk(symIds, LBUG_QUERY_BATCH_SIZE))`, merging into `affectedProcesses` — that merge is already a Map upsert, so cross-batch semantics are free here. Then either cap `changed_steps` or soften the docstring to say what is actually bounded.

Reproduced.

gitnexus/test/integration/detect-changes-path-anchoring.test.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+/** The fields these tests read off one `detect_changes` run. */
+type DetectChangesResult = {
+  error?: unknown;
+  summary: { changed_count: number };
+  changed_symbols: { name: string; filePath: string }[];
+};
+
+withTestLbugDB(
```
magyargergo, line 65: This test drives the real native engine but is not registered in the `lbug-db` vitest project, so it runs under parallel forks.

TESTING.md:68, verbatim: *"When adding a new test that uses native LadybugDB (`@ladybugdb/core`), add it to the `lbug-db` project's explicit include list and the `default` project's exclude list."*

`withTestLbugDB(..., { poolAdapter: true })` here is the same shape `vitest.config.ts:104-117` uses to justify registering `incremental-index-extension-dml-gate.test.ts`. The filename appears in neither `lbug-db.include` nor `default.exclude`, so it falls through to `default`'s broad `test/**/*.test.ts` and runs with `fileParallelism` on — the scheduling `lbug-db` exists to avoid for native mmap tests, on the Windows/macOS platforms this whole issue is about. The "no orphaned tests" check passes because the file is misassigned rather than orphaned, so nothing catches it.

Fix: add `'test/integration/detect-changes-path-anchoring.test.ts'` to `lbug-db.include` and `default.exclude`. `vitest.config.ts` is not in this diff, hence the note here.

Verified by inspection (membership checked in both arrays).

gitnexus/src/core/wiki/graph-queries.ts
```diff
@@ -147,23 +238,27 @@ export async function getInterFileCallEdges(): Promise<CallEdge[]> {
...
-  return rows.map((r) => ({
-    fromFile: r.fromFile || r[0],
-    fromName: r.fromName || r[1],
-    toFile: r.toFile || r[2],
-    toName: r.toName || r[3],
-  }));
+  // Sorted, unlike the original, which had no ORDER BY and so returned whatever
+  // the engine produced: `formatCallEdges` (prompts.ts) keeps only the first 30,
+  // and an unordered cut keeps a different subset per machine (#2787).
+  return rows.map(toCallEdge).sort(compareCallEdges);
```
magyargergo: The determinism fix is right; the placement costs 18,299 rows to use 30.

The only consumer, `formatCallEdges` (`prompts.ts:181`), does `.slice(0, 30)`. The root parent wiki page calls this with `allChildFiles` — effectively the whole repo (`generator.ts:907`). Measured with `filePaths` = all 2,298 indexed paths: **1,064 ms, 18,299 rows across the native boundary**; map 1 ms, sort 9 ms, 30 used. Pushing `ORDER BY ... LIMIT 30` into Cypher: **97 ms, 30 rows** — 11x faster, 610x fewer rows, same leading rows. The sort is not the cost; the row volume is, and it grows with repo size and call density.

`getInterModuleCallEdges` at `:288-289` already solves the identical problem in Cypher.

Also, `compareCallEdges` is not a total order when a name is null or empty — `compareCodeUnits(undefined, 'x')` returns `0` in both directions, so the comparator is intransitive and `Array.sort` becomes input-permutation dependent, i.e. the nondeterminism it was added to remove. Moving the order into Cypher subsumes that.

One caveat: the engine's `ORDER BY` compares UTF-8 bytes and `compareCodeUnits` compares UTF-16 code units — identical for ASCII identifiers, divergent above the BMP. The sibling already relies on the engine's order.

Reproduced.

gitnexus/src/core/wiki/graph-queries.ts
```diff
@@ -51,6 +61,87 @@ export interface ProcessInfo {
...
 
+/** A process without its step trace — one row of the process header query. */
+type ProcessHeader = Omit<ProcessInfo, 'steps'>;
+
+/**
+ * One result row. The pooled adapter yields rows keyed by column alias, and
+ * every mapper here falls back to the positional form, so the row type has to
+ * admit both.
+ */
+type QueryRow = Record<string | number, unknown>;
```
magyargergo: This documents a positional row ABI that the same PR proves is dead.

The docstring says *"every mapper here falls back to the positional form, so the row type has to admit both"*, but commit `fc6919f77` deletes ~30 identical fallbacks from `local-backend.ts` over the same pool adapter. Three pieces of evidence say the deletion is the correct side: the vendor declaration is `getAll(): Promise<Record<string, LbugValue>[]>` (string keys only), and two real-engine probes returned `row[0] === undefined` and `Object.keys(row) === ['pid','name','filePath','type','step']` with no numeric key.

Worse, the regrouped `withSteps` query prepends `p.id AS pid`, so the layout is now `[pid, name, filePath, type, step]` while `toProcessStep` (`:103-109`) still reads the pre-PR layout — if the dead path were ever revived, `step` would come back as the label string.

Fix: drop the positional arms, drop the `number` key, and type rows by alias so a renamed `AS` fails to compile.

Verified by inspection plus engine probe.

gitnexus/src/mcp/local/path-predicate.ts
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Build the WHERE fragment and the parameters for one path match.
+ *
+ * `property` is the row property to test (`n.filePath`), `prefix` names the
+ * parameters so several matches can share one query without colliding. The
+ * `pathSuffix` mode needs two parameters — the exact value and the
+ * separator-anchored suffix — because the anchored form alone cannot match a
+ * row whose stored path IS the caller's path.
+ */
+export function pathMatch(
```
magyargergo: `pathMatch()` and `PathMatchMode` have zero callers — only `pathSuffixOf` is imported (`local-backend.ts:27`, used at `:5197`). Every other `pathMatch` hit in the repo is an unrelated local variable.

The module docstring argues from a real duplication problem, but the PR migrates none of it: `local-backend.ts:3664` (`CONTAINS`, the `fragment` mode), `:4386-4392` and `:4611-4616` (both byte-for-byte what `pathMatch(..., 'pathSuffix')` builds) are untouched. So the tree now carries the abstraction *and* the copied idioms it was written to replace.

Minor latent hazard if it stays: `property` is concatenated verbatim into the returned clause with no restriction, so the first caller that passes a non-literal there gets a shaped clause. `filePath` is correctly bound.

Fix: finish the migration, or delete `pathMatch`/`PathMatchMode` and keep `pathSuffixOf`.

Verified by inspection.

gitnexus/src/storage/git.ts
```diff
@@ -698,3 +704,99 @@ export function parseDiffHunks(diffOutput: string): FileDiff[] {
...
+ * (macOS), and a swallowed 30s query timeout where they get more (#2915).
+ *
+ * Only ranges that overlap or ABUT (`next.startLine <= current.endLine + 1`)
+ * are merged, so the union covers exactly the lines the raw hunks covered —
+ * coalescing can never widen a range into a symbol the hunks did not touch.
+ */
+export function coalesceHunks<T extends GraphLineRange>(hunks: T[]): GraphLineRange[] {
+  if (hunks.length === 0) return [];
+  const sorted = [...hunks].sort((a, b) => a.startLine - b.startLine);
+  const merged: DiffHunk[] = [{ ...sorted[0] }];
```
magyargergo: The accumulator is `DiffHunk[]` — the 1-based type — inside the one function that exists to keep the two line spaces apart.

`GraphLineRange`'s docstring says *"a 1-based hunk cannot reach `hunksOverlapRange` without a conversion in between"*, but the two interfaces are structurally identical, so `tsc --strict` accepts a raw `DiffHunk[]` into both new APIs with no diagnostic (checked mechanically). `coalesceHunks<T extends GraphLineRange>(hunks: T[]): GraphLineRange[]` actively launders the base: hand it 1-based hunks and get back something typed graph-space. The generic buys nothing either, since the return type is not `T[]`.

Fix: drop the generic, retype `merged` to `GraphLineRange[]`, and either brand the type (`{ __base: 'graph' }`) or soften the docstring to say the contract is a single-call-site convention rather than a compile-time guarantee.

Verified by inspection.

gitnexus/src/core/lbug/query-batch.ts
```diff
@@ -0,0 +1,99 @@
...
+ * A failure is reported through `onError` and yields `undefined` for that item,
+ * so one bad query degrades the result (the caller raises its own `partial`
+ * flag) instead of discarding the items that succeeded beside it.
+ *
+ * Concurrency is safe for graph queries because `executeParameterized` checks a
+ * connection out of the per-repo pool for the duration of the query
+ * (`pool-adapter.ts`), so parallel calls never share one. The default leaves
+ * headroom for other in-flight work.
+ */
+export async function mapConcurrent<T, R>(
```
magyargergo: `mapConcurrent` is generic but lives in the DB layer with a DB-only safety argument, and it already has filesystem callers.

The docstring above justifies concurrency via the per-repo connection pool, but `core/ingestion/filesystem-walker.ts:7` and `core/ingestion/language-config.ts:10` now import this for `fs.readFile`, where that argument does not apply.

The commit message for `62eeaa1a4` states the precondition and then does not apply it — *"nothing about it is batch-specific, and it now has non-query callers"* — which is the same reasoning used two paragraphs earlier to move `chunk()` into `lib/utils.ts`.

Fix: move `mapConcurrent` next to `chunk()`. `LBUG_QUERY_BATCH_SIZE` and `warnIfQueryTextUnbounded` genuinely are query-specific and belong here.

Verified by inspection.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -7066,13 +7127,10 @@ export class LocalBackend {
...
       let chunksProcessed = 0;
-      for (
-        let i = 0;
-        i < impacted.length && chunksProcessed < MAX_CHUNKS;
-        i += CHUNK_SIZE, chunksProcessed++
-      ) {
-        const chunk = impacted.slice(i, i + CHUNK_SIZE);
-        const ids = chunk.map((item) => String(item.id ?? ''));
+      for (const batch of chunk(impacted, CHUNK_SIZE)) {
+        if (chunksProcessed >= MAX_CHUNKS) break;
```
magyargergo, line 7301: This inverts the cap when `IMPACT_MAX_CHUNKS` does not parse.

The `chunk()` adoption turned an *entry* condition into an *exit* condition. Old: `chunksProcessed < MAX_CHUNKS` — with `NaN` that is false, so the loop ran zero times. New: `chunksProcessed >= NaN` is also false, so the break never fires and **every** chunk runs. The truncation signal at `:7239` stops working too.

Same for any non-numeric value. Fix: validate the parse where `MAX_CHUNKS` is read and fall back to the intended default.

Verified by inspection.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -901,6 +907,15 @@ export function resolveWorktreeCwd(repoPath: string, launchCwd: string): string
   return repoPath;
 }
 
+/**
+ * Changed symbols listed in one `detect_changes` result.
+ *
+ * The counts in `summary` are never capped — only the array is, so a repo-wide
+ * diff cannot put an unbounded payload on the MCP wire. A truncated result says
+ * so in `symbols_truncated`.
```
magyargergo: Field name drift: the code emits `truncated` (`:5335`, `tools.ts:357`, and the test), not `symbols_truncated`. One word.

gitnexus/src/mcp/tools.ts
```diff
@@ -352,7 +352,9 @@ AFTER THIS: Review affected processes. Use context() on high-risk symbols. READ
 
 GIT WORKTREE SUPPORT: GitNexus automatically detects when the MCP server was launched from inside a linked git worktree and runs git diff against that worktree — no extra parameters needed in the common case. Pass "worktree" explicitly only when the server was started from a different directory than the worktree you are editing (e.g., the server runs from the canonical root but your changes are in a linked worktree at a different path).
 
-Returns: changed symbols, affected processes, and a risk summary.`,
+Returns: changed symbols, affected processes, and a risk summary.
+- partial: true — a graph query failed and was swallowed, so changed_symbols may be short. changed_count:0 with partial:true is NOT a clean pre-commit check; re-run before treating the diff as safe.
+- truncated: true — the changed_symbols LISTING was capped for this response. summary.changed_count is always the true total, so compare the two rather than trusting the array length.`,
```
magyargergo: "always the true total" is not true when `partial` is also set.

`changed_count` is `changedSymbols.size` (`local-backend.ts:5323`), populated only from the `UNWIND $bounds` batches that succeeded — there is no bookkeeping for what was not queried. So a client can receive `{partial:true, truncated:true, changed_count:842}` and conclude from this text that 842 is exact, when it is a lower bound. That combination is reachable: it is what the buffer-manager failure produces.

Two more, while this text is being edited:
- `partial` conflates two causes with different blast radii — the symbol query (`:5241`, degrades the counts) and the process lookup (`:5296`, degrades only `affected_processes`/`risk_level`). Only the first is described.
- `risk` derives purely from `processCount` (`:5300-5307`), so a degraded run still reports `risk_level:'low'`. An explicit unknown would be safer for a pre-commit gate.

Fix: qualify the `truncated` bullet ("if `partial` is also true, treat `changed_count` as a lower bound"), split the `partial` bullet to name both causes.

Verified by inspection.

magyargergo (commented):
## Review — #2930 (`detect_changes` hunk scaling, #2915)

Base `0fa547ccd` → head `6a1a41e78`, 39 files, +1973 / −379.

**Verdict: not ready to merge as-is.** Three small changes and it is. The core rewrite is right and holds up under deliberate attack; what is missing is the exit status on a degraded run, a memory regression in the new query plan, and a deterministic cut. Detail is in the inline comments.

**Merge state:** mergeable, no conflicts; checks failing (one red job, environmental — see below).
**Branch hygiene:** clean. Ten sequenced commits plus a merge from main, messages match contents, no unrelated churn, nothing that should not be version-controlled.

---

### What is solid

- **The #2915 blowup is gone and pinned.** Query text is byte-identical for a 1-hunk and a 3,000-hunk diff, with `Object.keys(params) === ['bounds']` and no `$hunk` parameters — asserted directly in `detect-changes-hunk-scale.test.ts:158-177`. Repeated attempts to make the text or the parsed expression tree grow with input again all failed.
- **The `coalesceHunks` merge invariant is provable and proved.** `next.startLine <= last.endLine + 1` merges only when no line exists between the two ranges, so the union covers exactly the lines the raw hunks covered. Checked by argument, by a 100,000-hunk adversarial run (descending, disjoint: 9 ms, 74 MB, input not mutated), and by the coverage test.
- **The two-stage filter is sound.** The Cypher `n.startLine <= b.hi AND n.endLine >= b.lo` is a provable over-approximation of hunk overlap; `hunksOverlapRange`'s binary search removes the gap false-positives. Its sorted/disjoint precondition holds — it has exactly one production caller, and that caller's input only ever comes from `coalesceHunks`.
- **The 0-based/1-based fix is correct and applied exactly once**, at `git.ts:768-769`, with nothing downstream re-converting, and proved end-to-end on a real engine (`detect-changes-path-anchoring.test.ts:93-97`: an edit landing on a symbol's last line is detected).
- **The dedup cannot drop a real match.** A node matching two different bounds was reproduced; the overlap test `continue`s *before* the `changedSymbols.has(sym.id)` check (`:5249` precedes `:5250`), so a wrong-path row never consumes the id.
- **The batching win is real and large.** 400-file diff against the real 25k-node index: 1,719 ms → 65 ms, identical kept sets. On a 1M-node index: ~6,600 ms → 432 ms. The `[lo, hi]` span cuts rows crossing the native boundary 5.1× (7,225 → 1,412). `EXPLAIN` confirms one `SCAN_NODE_TABLE` per query, so N per-file scans really did become ⌈N/100⌉ — this removed scans *and* prepares, not just round trips.
- **The traps this branch already hit did not recur.** No `--` comment inside any new Cypher — all three new query shapes were executed against a real engine and prepare cleanly. No surviving `chunk` shadowing; the `bestChunk` rename at `:3199` documents why.
- **Security is a strict improvement.** Wiki file lists moved from `escapeCypherString`-spliced `IN [...]` literals to bound `$paths`, so the query text is no longer a function of the input. `escapeCypherString` at the remaining splice sites was checked and is correct (backslashes doubled before quotes). No auth, secret or workflow change; no bidi characters; `git diff --check` clean; `tsc --noEmit` clean.
- **`temp-git-repo.ts` is a net correctness win**, not just de-duplication: the inline `spawnSync` copies it replaces never checked `result.status`.

---

### Blocking

**1. A degraded or failed `detect_changes` still exits 0.** `tool.ts:341` does `output(formatDetectChangesResult(result))` — it passes a *string*, and `output()` (`tool.ts:46-72`) sets `process.exitCode = 1` only for an *object* carrying `error`. `formatDetectChangesResult` also converts a backend `{error}` into text (`detect-changes-format.ts:37`), so a hard failure exits 0 too. This is the secondary half of #2915: the PARTIAL text landed, the exit status did not, and `gitnexus detect-changes && git commit` still proceeds on a degraded analysis exactly as before. `checkCommand` twenty lines below already does the object-first pattern. Fix in `detectChangesCommand`; `tool.ts` is not in this diff, so the inline note is on the formatter.

**2. The batched query materialises the whole node set, and fails where the old query succeeded.** `EXPLAIN` on the live index shows `UNWIND $bounds` + unlabeled `MATCH (n)` compiling to a `CROSS_PRODUCT` whose build side is a `RESULT_COLLECTOR` over the full filtered scan. Only the three `n`-only predicates push below it; both `b`-referencing predicates — the path match and the `[lo, hi]` span — sit above the product and cannot reduce the scan. Measured at 1M nodes: old 400 sequential queries +67 MB, new single batch +242 MB, new 4-concurrent +922 MB, paid even for a one-file diff. At a 268 MB buffer pool the old workload completes in 5,494 ms and the new one dies with `Buffer manager exception`; it still fails at 537 MB and 805 MB. That exception lands in `partial:true` + `changed_count:0` — the #2915 false clean by a new route, and per item 1, exit 0. There is a measured one-conjunct fix in the inline note (10× less memory, ~20% faster, identical result sets).

**3. The 1,000-symbol cut is nondeterministic.** `symbolQuery` has no `ORDER BY`, `changedSymbols` is a `Map` fed straight from the rows, and `:5316` slices it. Against a copy of the real index with the shipped query: 5 distinct row orders across 8 runs on one connection. Two `detect_changes` calls on an unchanged tree return different `changed_symbols`. `graph-queries.ts:258-260` in this same PR adds a JS sort with the comment *"an unordered cut keeps a different subset per machine (#2787)"*. `changed_count` stays correct, so this is contained — but it is the same class, 200 lines apart.

---

### Should fix before merge

- **The new real-engine test is not registered in the `lbug-db` vitest project.** TESTING.md:68 is explicit, and `detect-changes-path-anchoring.test.ts` appears in neither `lbug-db.include` nor `default.exclude`, so it falls through to `default`'s `test/**/*.test.ts` and runs under parallel forks — the scheduling `lbug-db` exists to avoid for native mmap tests, on exactly the platforms this issue is about. One line of config.
- **`symIds` is the one caller-sized list this PR did not chunk.** `:5265` → `WHERE n.id IN $ids` with the full deduped set. Measured on the real index: 100,000 ids → 1,238 MB, 500,000 ids → 4,002 MB / 2,595 ms. Binding keeps the *text* constant, which is all `warnIfQueryTextUnbounded` measures, while the bound *value* is repo-sized. The docstring at `:911-913` claiming a repo-wide diff cannot put an unbounded payload on the wire is contradicted here and by the uncapped `affected_processes` and `changed_steps`.
- **The anchored path match still admits every whole-segment sibling.** `'pkg/README.md' ENDS WITH '/README.md'`, so editing a root `README.md` reports symbols from `eval/README.md`, `gitnexus/README.md` and others — reproduced twice on a real engine. Strictly narrower than the old bare `ENDS WITH` (which also matched `MYREADME.md`), so this is an improvement that does not go far enough. A `detect_changes` path is always repo-root-relative because it comes from a `+++ b/` header, so `n.filePath = b.path` alone is the correct predicate.
- **`.claude/skills/gitnexus-work/SKILL.md:213-219`** mandates `detect_changes {scope:"staged"}` before every commit and mentions neither `partial` nor `truncated`. An agent following it literally commits on a degraded result — the failure class this PR exists to prevent, one layer up. Add a line: `partial:true` or `truncated:true` blocks the commit the same way an UNKNOWN impact risk does.
- **`truncated` never reaches the CLI formatter.** `partial` was wired end to end in this PR; its sibling was not.
- **The tool description over-promises.** `tools.ts:357` says `summary.changed_count` is "always the true total", but it is `changedSymbols.size`, populated only from batches that succeeded — so `{partial:true, truncated:true, changed_count:842}` reads as exact when it is a lower bound. Separately `risk` derives purely from `processCount`, so a degraded run reports `risk_level:'low'`.

---

### Smaller items

- `graph-queries.ts` adds `QueryRow = Record<string | number, unknown>` and keeps `row.x || row[n]` fallbacks, while commit `fc6919f77` deletes ~30 identical fallbacks from `local-backend.ts` over the same adapter. The vendor declaration is `getAll(): Promise<Record<string, LbugValue>[]>` and two probes confirm `row[0]` is `undefined`, so `graph-queries.ts` is the side to change. Its `toProcessStep` positional indices are also stale — `withSteps` now prepends `p.id AS pid`.
- `getIntraModuleCallEdges` ships every edge and sorts in JS to use 30: 18,299 rows / 1,064 ms where `ORDER BY … LIMIT 30` in Cypher is 30 rows / 97 ms. `getInterModuleCallEdges` twenty lines below already does it that way. The determinism fix is right; only the placement is wrong. `compareCallEdges` is also intransitive when a name is null or empty, which reopens the nondeterminism it was added to close.
- `pathMatch()` / `PathMatchMode` ship with zero callers, and none of the three sites its docstring describes (`:3664`, `:4386-4392`, `:4611-4616`) were migrated — so the abstraction and the copied idioms now coexist.
- `mapConcurrent` sits in `core/lbug/query-batch.ts` with a connection-pool safety argument, but `filesystem-walker.ts` and `language-config.ts` now import it for `fs.readFile`. The commit message for `62eeaa1a4` states the precondition it then does not apply — *"nothing about it is batch-specific, and it now has non-query callers"* — which is the same reasoning used to move `chunk()` into `lib/utils.ts`.
- `GraphLineRange` gives no compile-time protection: it is structurally identical to `DiffHunk`, so `tsc --strict` accepts a raw `DiffHunk[]` into both new APIs. `coalesceHunks<T extends GraphLineRange>` launders the base, the generic buys nothing, and the accumulator inside is still `DiffHunk[]`.
- `local-backend.ts:7307` still says "Run module query chunks sequentially (safe on arm64 macOS)" while the new path at `:5230` runs four concurrent queries. Not a demonstrated bug, but the file now asserts both policies with nothing to tell the next reader which applies.
- `IMPACT_MAX_CHUNKS=auto` now runs *every* chunk instead of none: the `chunk()` adoption turned an entry condition into an exit condition, and `NaN >= NaN` is false just as `0 < NaN` was.
- Deletion-only diffs issue zero queries and print "No changes detected." with `changed_files: 1` — a `-U0` deletion emits `count === 0`, the hunk is dropped, the file survives with `hunks: []`, and the `fileDiffs.length === 0` early return never fires.
- `local-backend.ts:915` names `symbols_truncated`; the emitted key is `truncated`.
- `QUERY_TEXT_CEILING_BYTES` is compared against `cypher.length` — UTF-16 code units, not bytes.
- `changed_symbols[].type` is `""` on every element. `labels(n)` returns the scalar string `"Function"` and subscripting is 1-based, so `labels(n)[0]` is always `""` and `[1]` is `"F"`. Pre-existing across 39 sites, but this PR flipped the field from absent to `""` and repeats the idiom in new wiki code.

---

### Attacks that were tried and held

Recording these because they are results, not gaps. `coalesceHunks` under 100,000 adversarial hunks (9 ms, 74 MB, no mutation, `(lo+hi)>>1` safe to 2^31). Inverted ranges, unreachable. `String(sym.diffPath)` key mismatch — round-trips byte-identically for `文档/说明.md`, `a b/c\td.ts`, `quote's.ts`; `startLine` returns a JS number, not BigInt. Absolute, `./` and backslash diff paths, unreachable from a `+++ b/` header. Eager `chunk()` — 88 bytes per slice, +0.6 MB at 1M items and size 100. `readFileContents` retention — 48 bytes per file against contents the returned Map already held; nowhere near a doubling. `mapConcurrent` on file I/O — not a regression, the code it replaced was also wave-based, and every adoption preserves its old failure semantics including the #1881 C# fail-open gate. `warnIfQueryTextUnbounded` — 15 ns per call, wired at all three read entry points and deliberately absent from the write path. `hunksOverlapRange` — genuinely O(log H), 37 ns per probe over 100,000 ranges. Cypher injection at the remaining splice sites. `temp-git-repo.ts` command injection — array-form `spawnSync`, no shell. Pool integrity and stdio framing under the new 4-way concurrency. And `changed_count` can only stay the same or fall, because the old unanchored `ENDS WITH` inflated it — a correctness gain, not a safety regression.

---

### Tests and CI

Strong where it counts: byte-identical query text for 1 vs 3,000 hunks, the line-base fix pinned at two layers including a real engine, the cap/`truncated`/`changed_count` interaction, 250 files → 3 queries, node dedup, real-engine path anchoring. The mock-for-shape, real-engine-for-semantics split is the right lesson from this branch.

Gaps:

1. The four rewritten queries in `graph-queries.ts` have zero real-engine coverage — the binding test mocks the pool adapter and answers from a hand-written JS reimplementation dispatched on `query.includes(...)`. A Cypher syntax error would pass every test and surface only in wiki generation, which is exactly the mechanism that already shipped a bug here. Both shapes were prepared manually against the real engine and succeed, so there is no live defect; what is missing is the CI guard.
2. The "counts a path reported twice" test (`detect-changes-hunk-scale.test.ts:260-264`) edits one file with one hunk, so `fileDiffs` has length 1 regardless — reverting `changed_files` to `fileDiffs.length` still passes.
3. Nothing drives `partial: true` through the new `mapConcurrent` path, and the failure granularity moved from one file to up to 100.
4. Nothing pins `changed_symbols` order, guards the plan shape, covers root-vs-nested paths, tests `path-predicate.ts` at all, or covers a non-numeric `IMPACT_MAX_CHUNKS`.
5. `query-batch.test.ts:41-56` asserts completion order from real 30/10/20 ms `setTimeout` deltas — a flake risk under a loaded shard.

**CI:** one real failure, `embedding-runtime-support.test.ts > isLocalEmbeddingStackInstalled` (`expected false to be true`, `:236`); 6124 passed. This is environmental, not caused by the diff: the suite passes locally on this head (26/26), `git diff base..head` on that file is empty, and its last change on main was #2370 — the assertion probes whether the optional embedding stack resolves in the workspace, which is a property of the shard. It is still red, so it needs a re-run or a shard fix before merge. Vercel is red on missing deploy authorization.

---

### Not verified

Nothing was run on macOS/arm64 or Windows, so the arm64 concurrency question is unresolved by measurement in either direction. No production build was produced. Every engine measurement is on `@ladybugdb/core` 0.18.3 while `package.json` declares `^0.19.0` — if 0.19 spills the accumulate to disk rather than raising, blocker 2 becomes a latency cliff instead of a hard failure. The 1M-node index used for the memory numbers is synthetic; the per-row accumulate size is the figure least likely to transfer, the shape does not depend on it. No index carrying `CodeEmbedding` rows was measured — that table has `startLine`/`endLine` and no `filePath`, so by the plan every embedding row is materialised on every query, which is reasoned from the DDL rather than observed. Findings in `src/cli/tool.ts`, `vitest.config.ts` and `.claude/skills/gitnexus-work/SKILL.md` are in files this PR does not touch and have no inline anchor; their locations are given above.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -5138,96 +5184,197 @@ export class LocalBackend {
...
+    // space, so every comparison below is base-neutral (#2377).
+    const hunksByPath = coalesceHunksByPath(fileDiffs);
+
+    // One row per changed file: the anchored forms of its path, and the [lo, hi]
+    // span of its whole touched region (coalesced ranges are sorted and
+    // disjoint, so the span is free).
+    const bounds = Array.from(hunksByPath, ([filePath, hunks]) => ({
+      path: filePath,
+      suffix: pathSuffixOf(filePath),
+      lo: hunks[0].startLine,
```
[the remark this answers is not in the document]
magyargergo: Not reproducible — the conversion happens between `parseDiffHunks` and the query, at the grouping boundary.

`bounds` is built from `hunksByPath`, which is `coalesceHunksByPath(fileDiffs)`, and that function applies `toZeroBasedLine` to both ends of every hunk before grouping (`gitnexus/src/storage/git.ts`). So the ranges reaching `n.startLine <= b.hi AND n.endLine >= b.lo` are already in the graph's 0-based space. `parseDiffHunks` staying faithful to git (1-based, like the `@@` headers it reads) is deliberate: the base change is made once, at the boundary, rather than in each consumer.

Pinned in both directions, because this exact off-by-one was the second bug fixed on this branch — an edit landing on a symbol's **last** line reported nothing changed:

- `test/unit/detect-changes-hunk-scale.test.ts` asserts the last-line edit is detected against the emitted query.
- `test/integration/detect-changes-path-anchoring.test.ts` asserts the same against a real LadybugDB.

Both fail if the conversion is removed, and the second fails if it is applied twice.

PR #2929, opened 2026-08-11
fix(scope-resolution): resolve a package whose directory name repeats higher in the path (#2881)

magyargergo:
Fixes #2881.

Four commits. The first fixes the reported Kotlin shape; the second finds the same rule live in
Java, Go and C# and removes it there too; the third is an output-identical optimization of the
Kotlin index; the fourth applies a four-lens review of the first three. A fifth group of commits
responds to PR review — see [Review response](#review-response).

## The bug

`getKotlinFileIndex` builds its `dirChildren` buckets under two guards, both inherited from the
pre-index per-import scan rather than from anything Kotlin requires:

```ts
if (norm.startsWith(`${suffix}/`)) continue;
if (norm.indexOf(`/${suffix}/`) === dir.length - suffix.length - 1) { ... }
```

`suffix` is `dir.slice(i + 1)` taken at each `/`, so `dir` ends with `/suffix` **by construction** —
the file always is a direct child of a directory named `suffix`. Both guards therefore only ever
drop legitimate buckets, and each drops a different shape:

| file | import | before | after |
| --- | --- | --- | --- |
| `data/src/main/kotlin/com/example/data/Repo.kt` | `data.helper` | `null` | fans out |
| `top/data/mid/data/Repo.kt` | `data.helper` | `null` | fans out |
| `data/src/main/kotlin/com/example/data/Repo.kt` | `data.*` | `null` | resolves |
| `src/main/kotlin/com/example/data/Repo.kt` (control) | `data.helper` | fans out | unchanged |

The issue names the `startsWith` shape. The `indexOf` first-occurrence check is a **second,
independent** mechanism that the issue's example never reaches, because `startsWith` fires first and
short-circuits it. A fix that drops only `startsWith` leaves every mid-path repeat broken; there is
a gate control for exactly that below.

Only the fan-out tier was affected. `data.Repo` answers from `suffixByStem`, which carries no such
guard, which is why the blast radius looked narrow enough for #2872 to preserve rather than change
inside a performance PR.

## The fix

Both guards are deleted; the loop is now a bare per-component-suffix insert.

The rule stays "the parent directory is named `s`", not "`s` appears somewhere in the path" —
`top/data/mid/Repo.kt` is still not a child of `data`, and a case pins that so the fix cannot be
widened by accident.

**The same rule was live in three more places, and this PR removes it from all of them.** #2881 was
reported against Kotlin, but the rule was never Kotlin's — it is what the pre-index per-import scan
[...]

PR #2934, opened 2026-08-12
fix(check): stop reporting erased and deferred imports as initialization cycles

magyargergo:
`gitnexus check --cycles` reports **11 cycles** on this repository. **Exactly one is a real initialization cycle.** The MCP tool describes its output as "a cycle count suitable for CI automation", and `--cycles` exits 1 when any are found — so at a 10-of-11 false-positive rate it cannot be adopted as a gate, which is presumably why nothing in `.github/workflows/`, husky, or any npm script runs it today.

## What a cycle here is supposed to mean

"These modules cannot be initialized in any order." Only edges that force initialization can form one. The check counted every `IMPORTS` edge instead.

Classifying all 11 by reading each edge's actual import statement:

| class | count | why it is not an init cycle |
| --- | --- | --- |
| every edge `import type` | 1 | erased by `tsc`; no runtime edge exists |
| broken by ≥1 `import type` edge | 7 | the chain never closes at run time |
| broken by a deliberate lazy import | 2 | both carry a comment saying so |
| **genuinely real** | **1** | `storage/branch-index.ts` ↔ `storage/repo-manager.ts` |

The two lazy ones are the sharpest. `eval/workflow_bench/proposer_sandbox.py:640` says *"Kept lazy to avoid a module cycle"*; `core/group/service.ts:395` does `await import('./cross-impact.js')`. The codebase broke those cycles on purpose, documented it, and the tool reported them anyway — flagging the fix as the bug.

## The two fixes

**Deferred imports** (`f3c147f2`). Two signals, because neither covers both spellings. `import()` already arrives as `ImportEdge.kind === 'dynamic-resolved'` — the emitter simply never read it. A function-local import is syntactically ordinary and nothing marks it; what defers it is *where it sits*, so the attached scope is walked to the file root.

That walk is not a check of the immediate scope kind, which is wrong in both directions: a `Block` at the top of a module (`if (FLAG) { require('./x') }`) initializes, the same `Block` inside a function does not, and `Class`/`Namespace` bodies execute where they are defined. The walk is also the language-agnostic half — it is what catches Python and Rust fn-local `use`, which `Scope.imports` already documents as the reason local imports exist.

**Type-only imports** (`e159f2a5`). The information was discarded at capture: `import-decomposer.ts` folded type-only specifiers into the same kinds on purpose and `ImportKind` had no member for it. `typeOnly` now rides `ParsedImport` → `ImportEdge` → the emitted `reason`. Both syntactic forms are read, since either alone is wrong — statement-level `import type { X, Y }` and per-specifier `import { type X, Y }` — and marking is per specifier, which is what makes the mixed form come out right: `Y` is a value import and carries the pair.

Not `importedSymbolKind: 'type'`, which already exists but is a PHP resolution-*namespace* category living only on the `named` variant. "This symbol is a type" and "this statement does not exist at run time" are different claims.

## Design notes

**The edge is still emitted.** A deferred import is a real dependency and impact/trace must see it; only the `reason` is tagged. That is the mechanism the check query already used to drop `markdown-link` and Swift implicit module visibility, so there is no schema change and no new edge property.

**Two suffixes, not one shared "ignore me" tag.** "Runs later" and "never runs" are opposite claims about the same edge, and `reason` is the only place the graph records which.

**Precedence is a rank** — initializing < deferred < erased, pair keeps the lowest it sees. Any initializing edge wins outright: one top-level `import { f }` beside an `await import()` and a dozen `import type`s is a real init dependency, and labelling that pair erased would **hide a true cycle**. Emission dedupes by `(source, target)`, so tagging from whichever edge arrived first would silently drop it; pairs are collected before anything is emitted, in first-seen order, so emission order is unchanged.

## The one real cycle is left alone, deliberately

`storage/branch-index.ts` ↔ `storage/repo-manager.ts` is a genuine value-import cycle in both directions, and it is **benign** — verified three ways: an AST audit walking only module-evaluation positions (top-level statements, class heritage, decorators, static blocks) found zero uses of an imported binding on either side; the build is plain `tsc` twice with no bundler, so ESM live-binding semantics apply; and entering the cycle from **both** sides under real Node ESM linking runs clean.

Every fix costs more than the report. The back-re-export has three live consumers (`run-analyze.ts:97`, `local-backend.ts:45`, tests). Moving `resolveBranchPlacement` is small and safe but **would not clear the report anyway**, since the `import type { RepoMeta }` edge remains. The version that does clear it strips `branch-index.ts` to ~35 lines, undoing the deliberate #2106 extraction to silence a checker on a pattern proven safe.

One gap recorded rather than papered over: nothing enforces that ESM-safety invariant, and every test enters from the `repo-manager` side, so a future top-level `const X = getStoragePaths(...)` would be a real TDZ crash with the suite green. A vitest guard would not catch it — Vitest's SSR transform is not real ESM linking. The checker is the right instrument, which is what this PR is for.

## Verification

- `npx tsc --noEmit` clean in **both** packages.
- `test/unit/scope-resolution/` + `test/unit/import-resolvers/` + `calltool-dispatch`: **2165 passed**, 119 files.
- Full `test/unit`: 49 failed / 12 259 passed — the 49 match the pre-existing baseline on a clean tree exactly (they need a built `dist/`), and **zero** were added.
- Mutation-tested, since a checker that reports nothing also reports no false positives:

| mutation | failures |
| --- | --- |
| drop the scope walk (the naive `kind`-only fix) | 3 |
| read the immediate scope kind instead of walking up | 1 |
| first-edge-wins tagging (drop static-wins) | 1 |
| drop the type-only tag entirely | 20 |
| treat a mixed `{ type X, Y }` as fully type-only | 6 |
| **invert precedence so erased beats initializing** | **10** |

The last one is the mutation that *hides* a cycle rather than inventing one, so it is pinned directly.

**Predicted effect: 11 → 1.** Stated honestly as a prediction — it is derived by re-classifying each cycle's edges from source, because this environment cannot rebuild the index (`npm run build` needs a non-symlinked `node_modules`). The unit and chain tests verify the mechanism end to end from source text to `reason` string; the count itself wants a rebuilt index to confirm.

## Graph-content impact: none beyond the label

The edge set and count are unchanged — only `reason` moves. `reason` is not an input to any id or persisted digest; `SCHEMA_FINGERPRINT` hashes DDL text, so a new distinct value needs no bump and invalidates no index; `detect_changes` never reads relationship `reason`; no bench fingerprint covers it (`bench/import-target` hashes resolver outcomes, not edges); and `pipeline-graph-golden`'s `edgeDigest` is `type|src|dst` — its fixture does contain `import type`, so it is genuine coverage of that no-op claim.

## Known gap, failing safe

`export type * from './m'` (TS 5.0) is not parsed by the vendored grammar — the `type` token lands in an `ERROR` node — so it is over-reported rather than silently dropped. Zero occurrences in this repository.

---

## Summary

*A broad scope-resolution and import-cycle analysis change spanning shared graph logic, language ingestion, CLI paths, and tests. Its file-level risk is low, but the dependency reach makes the behavior worth reviewing across the resolution pipeline.*

**🔴 CRITICAL blast radius. A scope-resolution and import-cycle change affecting 29 files and reaching 67 graph dependents across the repository.**

The core change is concentrated in `gitnexus-shared/src/scope-resolution/finalize-algorithm.ts`, where `makeEdgeDrafts`, `extractExportedName`, `tryFinalize`, and `tarjanSccs` participate in finalizing import relationships, alongside the resolution types in `gitnexus-shared/src/scope-resolution/types.ts`. The downstream analysis is in `gitnexus/src/core/graph/import-cycles.ts`, with related ingestion and graph-bridge changes across the language providers and `gitnexus/src/core/ingestion/scope-resolution/graph-bridge/imports-to-edges.ts`.

The architecture impact is concentrated in `Typescript` and `Scope-resolution`, with additional changes in `Cli`, `Storage`, `Ingestion`, `Languages`, `Pipeline`, `Search`, `Lbug`, `Server`, `Integration`, and `Cpp`. Review the finalization and import-edge paths first, then the cycle detection functions `findCyclePath` and `findImportCycles`, the hottest file `gitnexus/src/core/ingestion/languages/c-cpp.ts`, and the related deferred, type-only, and import-cycle tests. The file-level risk is `LOW`, with no `HIGH` or `CRITICAL` risk files identified.

_Added by GitNexus for PR #2934. Edit freely — this block is replaced on the next review, everything above it is left untouched._


magyargergo:
All six findings addressed in `43f3850d5`. The 🔴 was correct and it was the important one.

### 🔴 Function-local imports are never recognized as deferred in the production pipeline — **valid, fixed**

Confirmed exactly as described. `finalize-algorithm.ts:295` publishes every file's finalized edges as `linkedByScope.set(file.moduleScope, …)`, so the map `emitImportEdges` receives holds one bucket per *file*, keyed by that file's `Module` scope. The walk looking for an enclosing `Function` started at a `Module` and returned `false` every time, for every import in the tree. Python's `def f(): from x import Y` was still emitted as an initialization dependency, and this repository's one Python cycle — whose own comment reads *"Kept lazy to avoid a module cycle"* — was still reported.

The unit tests passed because they hand-built `new Map([['fn', …]])`, a shape the pipeline cannot produce. That is the same defect this PR exists to fix, one level up: a gate whose fixture asserts a shape the system never emits. The commit message for `f3c147f2` claimed the Python case was covered; it wasn't.

Fixed by deciding position where it is still known rather than trying to recover it downstream. Scope-extractor Pass 3 resolves each `@import.` anchor to its innermost scope and walks ancestors — `positionIndex` and `scopeTree` were already in scope there and already threaded into Passes 4 and 5. It travels as `ParsedImport.runsOnlyWhenCalled` → `finalize-algorithm` → `ImportEdge`, the same route `typeOnly` takes. `isInsideFunction`, `MAX_SCOPE_DEPTH` and the `scopeDefers` hoist are deleted rather than left as a dead helper.

New `test/unit/scope-resolution/function-local-import-chain.test.ts` drives real Python and Rust source through provider → extractor → finalize → emitter. **Reinstating the shipped emitter fails it 5/10** — verified independently — which is the property the old tests could not have had.

Two consequences worth recording: the PR body's "11 → 1" was aspirational when written (the code delivered 11 → 2, since the Python cycle survived); it is accurate now. And `SCHEMA_BUMP` had to move 60 → 62, because `ParsedImport` gained optional fields that `parsedfile-store.ts` serializes generically — a warm cache would have replayed untagged imports and made the whole branch a silent no-op on incremental analyze. #2864's ledger entry describes that trap in as many words.

### 🟡 Mixed-precedence test never includes a dynamic import edge — **valid, fixed**

Correct, and I wrote that case. Both entries were `edge('src/b.ts')` with the default `kind: 'named'`, so it never exercised `dynamic-resolved` — the first entry was deferred only by its `Function`-scope key, which (per the finding above) was itself fictional. Mixed-pair precedence is now one test per independent signal — `dynamic-resolved` vs initializing, function-local vs initializing — plus one carrying all three.

### 🟡 Test does not verify that suffix exclusions are inside the NULL alternative — **valid, fixed**

Correct. Every assertion was an independent `toContain`, while the comment claimed the exclusions must stay inside the `reason IS NULL OR (…)` alternative. It now extracts the balanced group and asserts membership in it. Verified by hoisting a clause out of the group — the exact bug described — which fails the test now and passed before.

### 🟡 Changed symbols: `interpretTsImport`, `LocalBackend`, `emitImportEdges` — **checked, no action needed**

No exported signature changed anywhere in the diff. `interpretTsImport(captures: CaptureMatch): ParsedImport | null` is untouched, and its `languages/javascript/interpret.ts` caller simply forwards — JavaScript has no `import type`, so `typeOnly` is never set on that path, which is correct by construction. `LocalBackend` only had a query string change inside a private method. `emitImportEdges` keeps its four parameters and its one caller.

Minor: the `emitImportEdges` finding cites `pipeline/run.ts:519`; the actual call site is `run.ts:1191`. The symbol and the caller are right, the line is stale.

---

Thanks — the 🔴 was a real defect in code I'd shipped an hour earlier, and the reasoning that found it (tracing from `linkedByScope` back to what the emitter can actually see) was the right way to catch it.

magyargergo:
Second-pass findings addressed in `dca5f4b0a`. Three were real, two were not — evidence for each below. Also reconciled with `main` now that #2929 has merged.

### 🔴 Preserve deferred metadata when expanding wildcard imports — **valid, fixed**

Confirmed. `expandWildcard` builds fresh `wildcard-expanded` edges rather than spreading (three fields differ per exported name) and copied neither flag, so a function-local `use crate::m::*` was tagged correctly on the base edge and lost the tag before the graph bridge. Both flags now travel through the expansion.

`typeOnly` is unreachable there today — `export type *` is unparseable by the vendored grammar, which the `wildcard` variant documents — and is carried anyway with a comment saying so, since the next reader would otherwise delete it as dead.

The gating case needed care, and this is worth recording: the chain harness stubbed `expandsWildcardTo: () => []`, which makes expansion a no-op returning the original edge. A wildcard test written against that stub passes whether or not the fix is present. The harness now returns the finalized edges alongside the reason, and the case asserts two `wildcard-expanded` edges both carrying the flag — so expansion is proven to have run. Removing the fix fails it; verified independently.

### 🔴 Update the cache-version pin when bumping SCHEMA_BUMP — **valid, fixed**

Correct, and it would have been a guaranteed CI failure. `incremental-parse-cache.test.ts` pinned 60 while the constant moved to 62. I found it by running only the targeted suites after the bump, having run the full suite before it.

Both now move together, and the pin names both neighbours rather than just the predecessor: 60 is what `origin/main` holds, so a rebase that drops this branch's bump lands there, and 61 is PR #2935's live claim, so a renumber reaching for "main + 1" collides with it.

### 🔴 C/C++ preprocessor includes are mislabeled as deferred imports — **valid, fixed**

Correct, and it is the failure direction that matters: `#include` is a compile-time textual splice, so a cycle formed by one is real and was being *suppressed*, not invented.

Fixed with a provider capability, `LanguageProvider.importsSplicedAtCompileTime`, rather than a language check in shared code — AGENTS.md forbids naming languages in `src/core/ingestion/` pipeline code. It only ever withholds deferral, so an undeclared provider is unchanged. `(preproc_include) @import.statement` appears in exactly two scope queries (`languages/c/query.ts`, `languages/cpp/query.ts`) and the flag is set on exactly the two providers that own them; the other 14 providers declaring `interpretImport` are untouched. C++'s only other Pass-3 import form, `using ns::name`, is compile-time name lookup and is correct under the same flag.

Gated by a test asserting the C include is plain **and**, in the same run, that a Python function-local import is still deferred — so "nothing is ever tagged" cannot pass it.

### 🔴 Preserve typeOnly for default imports — **refuted, but it found a real gap**

`ParsedImport` has no `default` kind. The decomposer's `default` case returns `kind: 'alias'` (with `...typeOnly`), and `typeOnlyFor` already covered `alias`. All seven `...typeOnly` spread sites in `interpret.ts` map onto the four kinds the helper handled, so nothing was being dropped. `import type Foo from './m'` was already asserted as `(type-only)` at `type-only-import-chain.test.ts:104-105`, predating this review.

But nothing pinned the *correspondence*, which is exactly why the question was open — so it is now pinned twice. A runtime table asserts which kind each erasable spelling arrives as, plus that its runtime twin arrives as the **same** kind without the flag (which is why erasure cannot be inferred from `kind`). And the helper's `default:` arm is gone: the four erasable kinds are listed and the rest go through a compile-time exhaustiveness assertion. Adding `typeOnly` to a variant the helper ignores is now `TS2345`, and adding a new kind is `TS2322 … not assignable to 'never'` — both verified by mutation.

Good finding even though the specific claim was wrong.

### 🟡 Changed symbol `extract` with 2 external callers — **checked, no action**

Signature unchanged; the new capability is an optional field on the hooks object, read once in Pass 3, defaulting to current behaviour.

---

**Merged with `main`.** #2929 landed while this was in flight, bringing its own `finalize-algorithm.ts` change (the `readonly string[]` annotation on `targetFiles`). Both survive the merge; `tsc` is clean in both packages and the targeted suites are at 123 files / 2256 passed.

Separately, the 🔴 raised against #2929 — that the `Array.isArray` annotation cannot type-check — is refuted there with an isolated repro compiling clean on TypeScript 5.9.3, plus that PR's own green `quality / typecheck`.

koriyoshi2041:
One remaining actionable CI failure on `a58cdaca`: the benchmark gate reports only the TypeScript scope-capture fingerprint drifting (`c2fbf8a8…` → `f719163e…`; all other language fingerprints and all 13 perf benchmarks pass). This branch intentionally changes `typescript/captures.ts`, so `gitnexus/bench/scope-capture/baselines.json` needs the new TypeScript fingerprint plus a short rebaseline note explaining the added/changed capture shape. The Windows 1/3 failure looks independent: `cli-e2e` timed out after 60s in the #1169 analyze-finalization test, while the rest of that shard reached 560 passes.

magyargergo:
Audited all 14 findings across the five review rounds against the current head. Three from later rounds hadn't been answered yet — one produced a new regression test (`9baf9f649`), two are refuted with evidence below.

## Status of every finding

| # | Finding | Verdict |
|---|---|---|
| 1 | Function-local imports never recognized as deferred | **Real — fixed** `43f3850d5` |
| 2 | Preserve deferred metadata when expanding wildcards | **Real — fixed** `dca5f4b0a` |
| 3 | Update the cache-version pin when bumping SCHEMA_BUMP | **Real — fixed** `dca5f4b0a` |
| 4 | C/C++ preprocessor includes mislabeled as deferred | **Real — fixed** `dca5f4b0a` |
| 5 | Mixed-precedence test never includes a dynamic import edge | **Real — fixed** `dca5f4b0a` |
| 6 | Suffix exclusions not verified inside the NULL alternative | **Real — fixed** `43f3850d5` |
| 7 | Preserve `typeOnly` for default imports | **Refuted** — no `default` kind; gap pinned anyway |
| 8 | C++ `using` treated as an initialization dependency | **Refuted** — see below |
| 9 | `presence` readonly but mutated | **Refuted** — see below |
| 10 | Non-JSON CLI crashes on the `none` response | **Guard already present; test added** `9baf9f649` |
| 11–14 | Changed-symbol notices (`interpretTsImport`, `LocalBackend`, `emitImportEdges`, `extract`) | **Checked** — no exported signature changed |

## 🔴 #9 — `presence` is declared readonly but is mutated

Not the case. Only two of the three properties are readonly:

```ts
const pairs = new Map<
  string,
  { readonly sourceFile: string; readonly targetFile: string; presence: ImportPresence }
>();
```

`presence` is intentionally mutable — the dedup pass lowers it as stronger evidence arrives, which is the mechanism that makes an initializing edge win a mixed pair. The asymmetry is deliberate: the two path fields are fixed once the key exists, the rank is not.

If it were readonly this would be `TS2540`, and `npx tsc --noEmit` is clean in both packages.

## 🔴 #8 — C++ `using` declarations

Refuted, and the direction matters: the proposed change moves toward **hiding** cycles.

`importsExecuteWhereWritten: false` means *position does not defer this import* — the edge is treated as an initialization dependency. The finding argues a function-local `using` should keep the deferred classification because it "only changes compile-time name lookup and does not load or initialize".

That premise is right and the conclusion inverts. "Deferred" in this model means *runs later, at call time*. A `using` inside a function body does not run later — it is resolved at compile time exactly where it sits, the same as one at namespace scope. There is nothing to defer. Classifying it deferred would assert that the dependency materializes only when the function is called, which is false for every C++ construct.

The failure directions are not symmetric, which is what settles it. Treating a compile-time form as initializing **over-reports** — a cycle is shown that a reader can dismiss. Treating it as deferred **suppresses** — a real cycle is dropped and nobody sees it. The capability's own docblock records that asymmetry, and this is exactly the case it was written for: `dca5f4b0a` flagged `using ns::name` as C++'s other Pass-3 import form and classified it under the same flag deliberately, not by oversight.

The narrower, genuinely open question is whether a `using` should produce an `IMPORTS` edge at all, since it names a symbol rather than pulling in a translation unit. That is about edge *emission*, not about this flag, and it would change the graph for every C++ repository — its own change.

## 🔴 #10 — Non-JSON handling of the `none` response

The mechanism is real; the crash is not, because the guard is already there:

```ts
// A rendering guard, not an exit-code decision — `output()` owns that. An
// error payload carries no `cycles` array, so the prose branch below would
// throw on it; print the structured payload and stop.
if (result?.error) {
  output(result);
  return;
}
```

`{ error, truncated }` has `error`, so it never reaches `result.cycles.map(...)`.

What the finding did surface is a real coverage gap: the existing error cases cover **JSON** mode and a **thrown** error, and neither exercises the branch that renders prose. `9baf9f649` adds that case, and I verified it gates — deleting the guard fails exactly this test and nothing else:

```
with guard:    20 passed
guard removed:  1 failed | 19 passed
               → "prints the safety-limit error in PROSE mode instead of throwing on the missing cycle list"
```

Good catch even though the code was already correct.

## Also on this head

`origin/main` is merged. #2930 and this branch had a genuine semantic collision in `checkCommand`'s exit code — #2930 centralized fail-closed inside `output()` and rewrote the line as `cycleCount > 0`, but `check` now returns `cycleCount: null` past the enumeration cap, and `null > 0` is false. Each change would have reintroduced the silent-green exit the other was preventing. Resolved to `status === 'cycles_found'`, keeping #2930's policy-vs-degradation framing.

`tsc` clean in both packages; full unit suite at 49 failures — the same pre-existing set, same 11 files — against 12,469 passing.

Issue #2686, opened 2026-07-25
impact(…, "upstream") returns 0 impacted when a symbol has exactly one non-test caller plus a test-file caller


RezaAlmiro:
`impact(…, "upstream")` reports **0 impacted / LOW** for a symbol whose caller edge is present in the index, when that symbol has **exactly one non-test caller and at least one test-file caller**.

Two or more non-test callers resolve correctly. No test caller at all resolves correctly. Only the 1-non-test + ≥1-test combination fails.

This is a false-safe: the repo's own guidance is to run `impact` before editing, and this returns the reassuring answer for a large class of ordinary, well-tested functions.

## Environment

- gitnexus 1.6.9, Node v24.14.0, win32/x64
- Reproduced on three unrelated repos: a pnpm TypeScript monorepo, a plain-JS browser-extension repo, and a Next.js app
- **Not** index drift — reproduces after `analyze --force`, from Git Bash, on a clean tree with the index at HEAD

## Reproduction

The edges are present. `cypher` on the same index, same session:

```cypher
MATCH (a)-[r:CodeRelation {type:'CALLS'}]->(b:Function)
WHERE b.name = 'buildCaptionAss'
RETURN a.name, a.filePath, r.confidence
```

```
| a.name          | a.filePath                            | conf |
| runClipPipeline | packages/server/src/clipPipeline.ts    | 0.85 |
| <test file>     | packages/server/test/…Smoke.test.ts    | 0.85 |
```

`impact({target: "buildCaptionAss", direction: "upstream"})` on that same index:

```json
{ "impactedCount": 0, "risk": "LOW", "epistemic": "exact", "summary": { "direct": 0 } }
```

## Evidence

| symbol | non-test callers | test callers | upstream returns |
|---|---|---|---|
| `assTime` | 2 | 1 | ✅ 3 |
| `pageWords` | 1 | 0 | ✅ 1 |
| `clamp01` (repo B) | 2 | 0 | ✅ 5 |
| `Button` (repo C) | 11 | 0 | ✅ 11 |
| `buildCaptionAss` | 1 | 2 | ❌ 0 |
| `clampCaptionPosY` | 1 | 1 | ❌ 0 |
| `audioLeadKeep` | 1 | 1 | ❌ 0 |
| `planJumpCuts` | 1 | 1 | ❌ 0 |
| `formatCitation` (repo B) | 1 | 1 | ❌ 0 |

[...]

PR #2935, opened 2026-08-12
fix(java): capture enum interface heritage

### LINKED ISSUE #2918: fix(java): emit Enum interface heritage for scope resolution
## Problem

Java enums can implement interfaces, but `synthesizeJavaInheritanceReferences` currently emits interface heritage for classes and records only. Consequently an enum such as:

```java
enum Status implements Named {
  ACTIVE;
  public String name() { return active; }
}
```

has no canonical `Enum -> Interface` `IMPLEMENTS` edge, preventing interface-dispatch fan-out.

Vendored tree-sitter-java exposes `enum_declaration.interfaces` with the same `super_interfaces -> type_list` shape used by records. `Enum` is already class-like, linkable, and supported by the relation schema.

## Acceptance criteria

- Java enum interface clauses emit normalized inheritance references for simple, generic, and qualified names.
- Scope resolution emits canonical `Enum -> Interface` `IMPLEMENTS` edges.
- Interface calls fan out to implementing enum methods.
- Existing class, interface, record, and enum-constant-body behavior remains unchanged.

Follow-up from #2916 and #2900.

azizur100389:
## Summary

- Capture Java enum `implements` clauses and annotated heritage across classes, records, interfaces, enums, and explicit `super()` calls.
- Keep source-written implementations ahead of synthetic enum constant bodies at the interface-dispatch cap, with dropped-target diagnostics.
- Preserve synthetic declaration metadata across the parsed-file cache and invalidate older cached capture shapes.

## Validation

- 420 affected tests passed with 1 documented skip.
- 25 parsed-file persistence tests passed.
- TypeScript typecheck, ESLint, Prettier, and diff checks passed.

Fixes #2918

---

## Summary

*A cross-cutting Java analysis change centered on enum interface heritage, with reach into scope resolution, shared definitions, persistence, and incremental-analysis tests. The blast radius is critical because the affected execution flows extend beyond the directly changed code, even though no individual file is rated high or critical risk.*

**🔴 CRITICAL blast radius. A Java ingestion and scope-resolution change for enum interface heritage that spans the analysis pipeline, shared types, parse-cache structures, and dependent execution flows.**

The core implementation is concentrated in `gitnexus/src/core/ingestion/languages/java/captures.ts`, including `synthesizeJavaInheritanceReferences`, `synthesizeJavaAnonymousClassDeclarations`, and `resolveVarTypeBindings`, with related dispatch behavior in `gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts`. It also changes `SymbolDefinition` in `gitnexus-shared/src/scope-resolution/symbol-definition.ts`, plus `PersistedFrameworkAnnotationRow`, `ParseCacheFile`, and `ShardedParseCacheIndex` in the analysis and storage layers.

The graph finds 3 dependents, with 2 direct callers or importers and 1 second-order dependent, across 17 affected flows. Impact is distributed across the `Unit`, `Passes`, `Pipeline`, `Kotlin`, and `Java` modules. Review the Java capture and receiver-bound call paths first, then verify the related incremental parsing and scope-resolution coverage in `gitnexus/test/unit/incremental-orchestration.test.ts`, `gitnexus/test/unit/scope-resolution/java/java-captures.test.ts`, and `gitnexus/test/unit/scope-resolution/scope-extractor.test.ts`.

_Added by GitNexus for PR #2935. Edit freely — this block is replaced on the next review, everything above it is left untouched._


azizur100389:
Resolved the latest-main cache-schema conflict at 64, preserving upstream 63 and keeping #2936 distinct at 65.

Validation: 333 affected tests passed (1 documented skip), plus typecheck and Prettier.

gitnexus/src/core/ingestion/languages/java/captures.ts
```diff
@@ -664,9 +666,13 @@ function synthesizeJavaInheritanceReferences(root: SyntaxNode): CaptureMatch[] {
         for (const base of superclass.namedChildren) emitJavaInheritanceBase(out, base);
       }
     }
-    if (node.type === 'class_declaration' || node.type === 'record_declaration') {
-      // Records cannot declare a superclass; they share only the class
-      // `interfaces` arm.
+    if (
+      node.type === 'class_declaration' ||
+      node.type === 'record_declaration' ||
+      node.type === 'enum_declaration'
```
magyargergo, line 678: **A bodied-constant enum joining an interface can evict every pre-existing class implementor from the interface-dispatch fan-out.** `[reproduced]` *(The cap itself lives in the unchanged `gitnexus/src/core/ingestion/scope-resolution/passes/receiver-bound-calls.ts:505-512`; anchoring here because this line is what starts contributing to it.)*

**Provenance, so you can weigh it correctly:** this is a **single-lane** finding — Claude's adversarial persona only; the Codex engine never constructed it. Within that lane it was reproduced **twice with a byte-identical victim list**, and I re-verified the mechanism by reading the cap implementation. **No cross-engine corroboration**, so treat it as one strong repro rather than a consensus.

**Trigger:** an interface implemented by both a class hierarchy and a strategy enum whose combined closure exceeds `MAX_INTERFACE_DISPATCH_FANOUT` (32). Before this PR an enum contributed **zero** subtypes; now it contributes `1 + N` (itself plus one synthesized `E$N` class per bodied constant, via the pre-existing `Status$N EXTENDS Status` edges at `captures.ts:473-495`).

**Evidence:** fixture = `interface Named` + abstract `ZBase implements Named` + 30 concrete `ZImplNN` + one 40-constant `enum AaaBig implements Named`, run through the real pipeline:
```
WITHOUT ENUM: 30 targets — 0/30 concrete class implementors lost
WITH ENUM:    71 → 32   — 30/30 concrete class implementors lost
```
`emitInterfaceDispatchFor` does a FIFO BFS and then truncates the **tail** (`targets.length = MAX_INTERFACE_DISPATCH_FANOUT`), so the 30 real classes are exactly what gets cut. Renaming the enum `AaaBig` → `ZzzBig` — nothing else changed — pops `ZBase` first and loses **0/30**. So which edges die is def-insertion order, not semantics, and a file rename can silently delete 30 `CALLS` edges (`impact("ZImpl17.label", upstream)` goes 1 caller → 0/UNKNOWN).

To the cap's credit this is not silent — it warns — but the warning names only the interface member and a count (`Named.label (71 targets)`), never the dropped implementors.

**Suggested:** rank `targets` before truncating so directly-declared implementors outrank synthesized `X$N` bodies, and/or charge an enum + its constant bodies one budget slot; at minimum include the dropped implementor names in `dispatchFanoutSkippedNames`. A >32-combined-target regression test next to the new #2918 cases would pin it.

gitnexus/src/core/ingestion/languages/java/captures.ts
```diff
@@ -733,6 +739,11 @@ function javaBaseLookupNameNode(node: SyntaxNode): SyntaxNode | null {
       const first = node.firstNamedChild;
       return first === null ? null : javaBaseLookupNameNode(first);
     }
+    case 'annotated_type': {
```
magyargergo, line 750: **This arm is a four-way heritage fix, documented and tested as if it were enum-only.** `[reproduced]`

**Trigger:** any Java source with a type-use annotation on a heritage base. Before this PR `annotated_type` hit `default: return null`, so **no capture at all** was emitted.

**Evidence:** old-vs-new probe with the base commit's `javaBaseLookupNameNode` reimplemented verbatim, both over the real tree-sitter parse:
```
class C extends @Marker Base        OLD: null (no capture)   NEW: ["Base"]
class C implements @Marker Named    OLD: null                NEW: ["Named"]
record R(int i) implements @Marker Named   OLD: null         NEW: ["Named"]
interface I extends @Marker Named   OLD: null                NEW: ["Named"]
```
plus `super()` target naming under an annotated superclass. All correct — this is a genuine bonus fix, not a defect. But `git grep -n annotated_type -- gitnexus/src gitnexus/test` returns **exactly one hit** (this line): every test that reaches it uses an `enum`.

The traceability cost is real: the durable id is `java.enum-interface-heritage` and the ledger note still says *"Java enum `implements` clauses"*, so someone bisecting a future class-`extends` edge-count change will not find this PR. (Rebuild coverage is still sound — `appliesTo` matches any `.java` file, not just files containing enums.)

**Suggested:** one parameterized unit case over the five arms in `java-captures.test.ts` asserting the annotated form matches the unannotated control (cheap, no new fixture), and widen both the ledger note and — ideally — the descriptor id, which is free to rename *now* and costs a full rebuild to rename after release.

For the record, the scarier version of this concern is **refuted**: the arm is **inert** on the two `javaBaseSimpleNameOf` call sites (receiver normalization `:258`, anonymous-class synthesis `:443`), because `object_creation_expression`'s `type` field is `_simple_type` in `node-types.json` and the annotation is a sibling, not a wrapper. Confirmed by three reproduced AST dumps.

gitnexus/test/integration/resolvers/java.test.ts
```diff
@@ -1341,6 +1341,72 @@ describe('Java record method resolution (#2564)', () => {
...
+      });
+
+      const linked = await runPipelineFromRepo(root, () => {});
+      const implementsEdges = getRelationships(linked, 'IMPLEMENTS').filter(
+        (edge) => edge.source === 'Status' && edge.target === 'Named',
+      );
+
+      expect(implementsEdges).toHaveLength(1);
+      expect(implementsEdges[0]?.sourceLabel).toBe('Enum');
+      expect(getNodesByLabel(linked, 'Method').filter((name) => name === 'label')).toHaveLength(3);
```
magyargergo, line 1407: **This test's title promises dispatch safety for constant-body overrides, but it asserts a node census.** `[reproduced]`

**Trigger:** the `EnumConstantBody.java` fixture has **no caller at all**, so there is no `CALLS` edge in the graph to assert against. `getNodesByLabel(linked, 'Method').filter(...).toHaveLength(3)` is a structure-phase property that holds with or without the heritage edge.

**Evidence:** break the `Status$1 EXTENDS Status` link — the subtype-closure edge that actually carries dispatch into the constant body — and this test still passes (still 3 `Method:label` nodes, still 1 IMPLEMENTS) while fan-out silently drops from 2 interface-dispatch edges to 1. The good news is the real behavior is correct today: adding `class Reader { String read(Named v) { return v.label(); } }` and re-running the pipeline yields
```
read -> label  reason=global             tgt=Named.label#0
read -> label  reason=interface-dispatch  tgt=Status.label#0
read -> label  reason=interface-dispatch  tgt=Status$1.label#0   <- the constant-body override
```
so the stronger assertion passes as-is. This is the highest-value assertion available in this PR and it is the one that is missing — flagged independently by both review engines.

**On severity:** the Codex leg marked this `blocks_merge: yes`. I am calling it **non-blocking** because the behavior itself was independently reproduced as correct on this branch — what is missing is the assertion, not the fix.

**Suggested:** add the `Reader` caller to the fixture and assert the sorted target-id set `['…Status$1.label#0', '…Status.label#0']`. A companion case with an `abstract` enum method implemented only by constant bodies would cover the other half.

magyargergo (commented):
# Multi-tool review — `fix(java): capture enum interface heritage` (#2918)

**How this was produced.** Two review engines ran the same swarm + `ce-code-review` scaffolding:

- **Engine A — Claude:** 4 swarm lanes (pr-facts-historian, risk-architect, test-ci-verifier, branch-hygiene) + 7 `ce` personas (correctness, adversarial, testing, maintainability, standards, performance, simplicity).
- **Engine B — Codex gpt-5.6-sol/xhigh:** swarm Solo, `ce` report-only, adversarial-review. **All three legs are live** — each returned real, file-cited findings.

**The head moved twice while this ran.** Every lane analysed `2be508e79...f90eac9e7`. `origin/main` then advanced (#2934 merged at 17:09 UTC), and at **17:37 UTC the author pushed `fa24b1cc8`**, rebasing onto the new main (`d540b0018`). Everything below has been **re-anchored and re-verified against `fa24b1cc8` / base `d540b0018`** — line numbers, mergeability, CI, and the one blocking item, which the author fixed in that push (§1).

Please read this as **two engines, not fourteen independent opinions.** Lanes within one engine share a prompt scaffold and can share a blind spot, so below I distinguish *cross-engine* agreement (the strong signal) from *consistent within one engine*.

---

## What is solid — and it is most of the PR

The enum implementation itself is **correct**, and it was attacked hard before anyone concluded that:

- **`IMPLEMENTS: Enum:Status -> Interface:Named` is emitted, `sourceLabel === 'Enum'`** — reproduced end-to-end through `runPipelineFromRepo` by three Claude lanes.
- **Interface dispatch reaches enum constant-body overrides.** The feared #2686-style impact under-report does not exist: `read(Named)` fans out to *both* `Status.label#0` and `Status$1.label#0`, in the mixed fixture *and* the all-constant-bodies fixture. Mechanism: `subtypesBySupertypeDefId` is built from IMPLEMENTS ∪ EXTENDS (#2829) and `isClassLike` already accepts `'Enum'`, so `Named → Status → Status$1` is a two-hop path the closure already handled.
- **The `annotated_type` arm survived a real fuzz.** 26 inputs (well-formed, JLS-illegal, torn) against the vendored tree-sitter-java 0.23.5: zero cases where `lastNamedChild` is an annotation. The grammar is `seq(repeat1($._annotation), $._unannotated_type)`, and error recovery inserts a MISSING `type_identifier` rather than dropping the type.
- **`tsc --noEmit` clean, `prettier --check` clean, 322 tests passing across the five touched suites, `java.test.ts` 254/254.**
- **`gitnexus/test/unit/incremental-orchestration.test.ts:518-553` is the strongest thing in this PR.** It proves the invalidation path end-to-end on a **real LadybugDB** with a real Cypher query — and it carries a **negative control** (delete the `IMPLEMENTS` edge, assert 0, then rebuild and assert 1), so the passing assertion cannot be a stale read. Independent mutation testing (source mutated, `tsc` rebuilt, `dist/parse-worker.js` regenerated) confirmed every new source branch is killed by a specific assertion.
- **Deleting the old characterization test** (`'does not yet emit enum heritage (#2918)'`) instead of leaving it stale is the correct lifecycle, and it is hard evidence of the pre-PR behavior.
- **Architecture is clean:** no language name enters shared `src/core/ingestion/` code; `preEmitInheritanceEdges` still classifies from the resolved target's kind, not from Java syntax.

**Scary hypotheses the review killed, with evidence — please don't re-litigate these:**

| Hypothesis | Verdict |
|---|---|
| The new arm changes **receiver normalization / anonymous-class synthesis** | **REFUTED four ways.** Both `javaBaseSimpleNameOf` call sites (`captures.ts:258`, `:443`) read `object_creation_expression`'s `type` field, declared `_simple_type` in `node-types.json` — annotations are *siblings*, never a wrapper. Three reproduced AST dumps agree. Two lanes reached the opposite claim by code-read (enumerating call sites without checking grammar reachability); that claim is wrong and is dropped. |
| Interface dispatch misses enum constant-body overrides | **REFUTED**, reproduced (see above) |
| `annotated_type.lastNamedChild` can return an annotation | **REFUTED**, 26 fuzzed inputs, BAD count 0 |
| Array / varargs bases sneak a wrong name through | **REFUTED** — `array_type` hits `default: return null`, no capture |
| Stack overflow / pathological recursion in the new arm | **REFUTED** — depth 1–2 in every shape; 2000 stacked annotations = 24 ms |
| Duplicate IMPLEMENTS or an MRO cycle from constant-body synthesis | **REFUTED** — `buildMro` collects only EXTENDS; edge dedup is keyed `TYPE:src->tgt`; `defaultLinearize` carries a `visited` set |
| Nested / local / interface-body enums mis-attribute the clause to the enclosing type | **REFUTED** — `query.ts:38` anchors `@scope.class` on the whole `enum_declaration` |
| `permits` should also be emitted | **Correctly NOT emitted** — it would invert edge direction, and per JLS an enum is implicitly final or sealed |
| A stale-cache **replay** from an ordering downgrade | **REFUTED** — `PARSE_CACHE_VERSION` is compared for exact equality (`parse-cache.ts:757`, `:781`, `parsedfile-store.ts:647`), so a downgrade invalidates as thoroughly as an upgrade. Only an **exact clash** is dangerous — see §1, now closed. |
| The feature descriptor can be stamped **without** the evidence | **REFUTED across four sequences** (downgrade → `unexpected:` forces rebuild; warm cache + warm index → checkpoint 1 sits above the `!options.force` fast path; repo gains its first `.java` mid-history → post-pipeline checkpoint 2; interrupted run → the stamp is written only at the terminal `saveMeta`). The one remaining path was the exact clash in §1, which the author has closed. |
| The descriptor is redundant with `SCHEMA_BUMP` (over-engineering) | **REFUTED by both engines.** They gate disjoint layers: `SCHEMA_BUMP` → the content-addressed parse cache (which survives `--force`); `ANALYSIS_FEATURES` → whether the pipeline runs at all (the `alreadyUpToDate` fast path is gated on `!options.force` and short-circuits before `loadParseCache`). Neither substitutes for the other. |
| Perf regression from the enum arm | **REFUTED, measured** — identical node-visit count old vs new; +14 ms on an adversarial 12,000-line / 1,500-enum synthetic file; `resolveAnalysisFeatureVersions` < 1 ms at this repo's ~2,079 files |
| `appliesTo` is broken on Windows paths | **Not a bug** — `.endsWith('.java')` inspects only the trailing five characters; no separator is involved |
| `gitnexus-check[bot]` lists `resolveVarTypeBindings` as changed | **Bot false positive** — that function starts at line 752; the hunks span 636-650, 664-680, 733-744 |

---

## 1. The one blocker we found — and you already fixed it (credit, not an ask)

While the lanes were running, this branch carried `SCHEMA_BUMP = 61`. That was an **exact clash** with open PR #2840 (`feat/objective-c-upstream`, head `24177146`), which independently also moves 60 → 61 — and after #2934 merged at 17:09 UTC it was also *below* main's 63.

The exact-clash shape is worth stating once, because it is the reason the fix matters and it is not obvious: `PARSE_CACHE_VERSION = ${SCHEMA_BUMP}+${pkgVersion}` and the reuse gate is exact **string equality**, so an equal-but-different-meaning version is the one case where **neither side invalidates the other** — a warm `61+<pkg>` store written by #2840's build is byte-identical to this build's key and is replayed verbatim. `findAnalysisFeatureMismatches` would fire `missing:java.enum-interface-heritage` and set `force = true`, but this very file records at `parse-cache.ts:294-299`, as an *observed* incident rather than a theory, that *"`--force` re-runs the pipeline but still serves ParsedFiles from the durable store, so it does not substitute for this bump."* The run would then stamp `java.enum-interface-heritage: 1`, **certifying evidence that is not in the graph** — silently reproducing the bug this PR fixes, while both PRs' pin tests stayed green.

**Already fixed at `fa24b1cc8`, coordinator-verified directly on the head:**

```
gitnexus/src/storage/parse-cache.ts:511          const SCHEMA_BUMP = 64;
gitnexus/test/unit/incremental-parse-cache.test.ts:216   toBe(64)
                                          :224-227  not.toBe(60|61|62|63)
```

64 is above main's 63 and unclaimed — #2936 has moved to **65**, and #2840's 61 now sits below main, which is #2840's renumber to do, no longer a clash with this branch. The ledger paragraph at `:505-509` also now records the cleared claims and names #2936's 65, which closes a separate P3 the draft was going to raise. Nothing left to do here.

For the record on provenance: the finding was **cross-engine** — 4 Claude lanes + Codex `ce` (*"change 61 to the current safe value 64 and update the pin test"*) + Codex adversarial (`[high]`, no-ship). Codex swarm and the Claude risk-architect lane reached only the *rule violation*; they checked #2936/#2934/#2891, missed #2840, and correctly said they could not construct a live clash from what they had seen.

## 2. Scale interaction worth a look — enum fan-out can evict pre-existing class implementors (P2)

The most interesting finding in the review, and the one I'd ask you to weigh. This is **single-lane** (Claude adversarial only — Codex never constructed it), but it was **reproduced twice, byte-identically**, and I re-verified the mechanism by reading the code, so it clears the bar.

A bodied-constant enum ("strategy enum", *Effective Java* Item 34) now contributes `1 + N` entries to `subtypesBySupertypeDefId` for an interface where it contributed **zero** before this PR — the enum itself, plus one synthesized `E$N` class per bodied constant. `emitInterfaceDispatchFor` (`receiver-bound-calls.ts:458-512`, unchanged by this PR) collects the whole closure and then **truncates the tail**:

```ts
targets.length = MAX_INTERFACE_DISPATCH_FANOUT;   // 32
```

Measured on an interface with an abstract base + 30 concrete classes, plus one 40-constant enum:

```
WITHOUT ENUM (pre-fix equivalent): 30 targets — 0/30 class implementors lost
WITH ENUM AaaBig:                  71 → 32 targets — 30/30 class implementors lost
```

Rename the enum `AaaBig` → `ZzzBig`, change nothing else, and BFS pops the abstract base first: **0/30 lost**. The victim set is decided by def-insertion order, not semantics — so renaming a file can silently delete 30 `CALLS` edges and flip `impact("ZImpl17.label", upstream)` from 1 caller to 0/UNKNOWN.

To the design's credit this is **not silent** — the cap warns. But the warning names only the interface member and a count (`Named.label (71 targets)`); an operator has no way to learn that all 30 real classes were the ones dropped. Suggested directions: rank `targets` so directly-declared implementors outrank synthesized `X$N` bodies before truncating, and/or charge an enum plus its constant bodies one budget slot; at minimum, report *which* implementors were dropped.

Two other Claude lanes independently flagged that "which targets survive is BFS-order-dependent" as an unquantified residual risk without constructing the eviction — so within Engine A there is partial support for the concern, but the concrete repro is one lane's.

## 3. Inline findings (details on the diff)

| # | Location | Sev | Theme |
|---|---|---|---|
| 1 | `captures.ts:672` | P2 | Enum fan-out can evict pre-existing implementors past the 32 cap |
| 2 | `captures.ts:742` | P2 | `annotated_type` widens four non-enum arms + `super()` naming; untested and understated |
| 3 | `java.test.ts:1403` | P2 | Constant-body test is a node census, not the dispatch property it names |

All three anchors were re-verified against the new head `fa24b1cc8` and hold at the same lines — the rebase touched only `parse-cache.ts` and its pin test, not the Java source.

On #2, to be precise about the blast radius (the coordinator settled a three-way lane disagreement here from the vendored grammar's `node-types.json`): the arm is **reachable** from class `extends`, class/record/enum `implements`, `interface extends`, and `super()` target naming — all previously returning `null`, i.e. **no edge at all**, so this PR silently fixes annotated heritage for classes, records and interfaces too. It is **inert** on the two object-creation paths. All the new edges are correct; the issue is that neither the feature id nor the ledger note mentions any of it, so a future bisect of a class-`extends` edge-count change will not find this PR.

## 4. Lower-priority items (body only — no inline)

- **Test strength, `java.test.ts:1377`.** `` `${targetLabel}:${targetFilePath}` `` discards target identity — `Status.label`, `Status$1.label` and `Named.label` all stringify the same. Two lanes disagreed about how bad this is; on balance it is milder than first claimed, because the `reason === 'interface-dispatch'` filter already excludes the `global` edge to `Named.label` and the assertion pins *exactly one* element, so over-fan-out still fails. The stronger form is free — `rel.targetId` is already exposed: `expect(fanout.map((e) => e.rel.targetId).sort()).toEqual(['Method:EnumHeritage.java:Status.label#0'])`.
- **Non-compilable annotation fixture, `java-captures.test.ts:112`.** `implements @Marker audit.Auditable` is not legal Java — per JLS §9.7.4 / §4.11 a TYPE_USE annotation on a qualified type goes before the *simple* name (`audit.@Marker Auditable`). tree-sitter accepts the illegal form as `annotated_type`; the **legal** form parses as `scoped_type_identifier` and goes through the *pre-existing* arm, so the test named "qualified" doesn't exercise the spelling a Java author writes. Both engines, reproduced (legality is spec-read — no JDK in the sandbox). The legal form was verified correct anyway.
- **`java-captures.test.ts:121`** overrides `name()` in a constant body; `java.lang.Enum.name()` is `final` and cannot be overridden. Harmless to the assertion, but it makes the fixture unusable as documentation. `label()` would match the integration fixture.
- **`.java` predicate duplicated** at `analysis-features.ts:17` and `:25`. Both engines noticed. One Claude lane argues a 1-line helper for a 1-line duplicate is its own YAGNI violation — and there is a **correctness** reason not to reach for the obvious existing helper: `isJvmSourceFile` (`frameworks/spring/analysis-features.ts:35-38`) matches `.kt` and `.kts` as well as `.java`, so reusing it here would force a full rebuild on Kotlin-only repos that contain zero Java files. If you do factor it out, factor out a Java-only predicate. Your call whether it is worth the extra symbol.
- **Descriptor id granularity.** `java.enum-interface-heritage` names one bugfix; the five pre-existing descriptors are capability-scoped and the established idiom for a completeness fix inside a capability is to bump `version` (`spring.bean-inventory` is at v2). Renaming is free *now* — after release the id is durable and renaming would itself cost a full rebuild. Flagged by both engines.
- **Ledger note scope.** The new v64 entry describes the change as *"Java enum `implements` clauses"*, but the shared `annotated_type` arm widens four heritage arms plus `super()` naming (§3). Widening that sentence matches the file's own precision bar. (The "which in-flight claims were checked and cleared" half of this item is already done in `fa24b1cc8`.) While you are there: the inherited rationale block just above the new guard in `incremental-parse-cache.test.ts:218-223` — which came from main via #2934, not from you — now reads stale next to your change. It says *"60 is what origin/main holds"* (main holds 63), *"61 is claimed by BOTH #2935 and #2840"* (this branch moved to 64; only #2840 still claims 61) and *"62 is #2936's claim"* (#2936 moved to 65). The assertions themselves are all still correct and worth keeping.
- **PR description.** *"Pre-PR swarm and GitNexus review found no remaining code defects"* is half substantiated — the GitNexus half has a real bot comment; the swarm half has zero visible artifact, unlike sibling PR #2871 which published a "## Review digest". Nobody can prove a negative here; either attach the findings as #2871 did or soften the claim.
- **Self-loop on a legal cross-package name collision** (P3, reproduced). `package b; enum Named implements a.Named` resolves the base to the *enclosing* type and emits `EXTENDS Enum:Named -> Enum:Named` — a shape javac can never produce — while the real `IMPLEMENTS b.Named -> a.Named` is lost. A control proves this is **pre-existing** (identical with `class Named`); this PR only widens it to enums. Root cause: `emitJavaInheritanceBase` emits no qualified capture, so `rawQualifiedName` is undefined and `resolveQualifiedInheritanceBase` is skipped. Cheap insurance either way: `if (callerGraphId === targetGraphId) continue;` in `preEmitInheritanceEdges` — a type is never its own supertype in any language here.
- **Empty-name capture on a torn parse** (P3, reproduced). `enum Status implements @Marker {` error-recovers to a MISSING `type_identifier` with empty text; the new arm returns it and a reference site with `name === ""` is minted where pre-PR none was. Verified inert end-to-end (no edges emitted), but it still consumes a `referenceSites` slot and a `handledSites` suppression key on every parse of a half-typed file — the state an editor-triggered incremental analyze sees constantly. One line closes it: reject `nameNode.isMissing` / `text.length === 0` before pushing.
- **Adjacent pre-existing gap, not this PR's to fix.** `languages/java/receiver-binding.ts:38-49`'s `firstSuperclassText` hand-rolls the same "find the base type under `superclass`" walk and matches only `type_identifier | generic_type` — so for `class Bar extends @Marker Base`, `annotated_type` falls through and no `super` type binding is emitted, silently breaking `super.method()` receiver resolution. Worth a follow-up issue: that walker should call the canonical `javaBaseLookupNameNode` rather than reinvent a weaker version.
- **Systemic, all six descriptors:** `analysisFeatures` is stamped unconditionally at `run-analyze.ts:3557`, and `INDEX_INCOMPLETE_REASONS` has no reason for "a file was worker-quarantined during parse". A repo where one `.java` file crashed its parse worker gets the completeness stamp anyway. Inherited, not introduced — noting it because this PR adds a sixth consumer.
- **Merge-order heads-up:** #2936 inserts a descriptor at the *identical* anchor in `analysis-features.ts` and `run-analyze.ts` and touches the same `SCHEMA_BUMP` line (it now claims **65**). Loud textual conflicts, not silent ones.

## 5. Test gaps (none merge-blocking on their own)

1. Nothing pins the fan-out-cap interaction (§2) — the PR's dispatch assertion uses a **one-target** fixture, so it cannot see the cap at all.
2. Zero coverage for the four non-enum arms the shared `annotated_type` case newly enables — on a real Java corpus these will move more edges than the enum arm does.
3. The constant-body dispatch property is unasserted (inline #3). **Severity note, stated openly:** Codex's swarm leg marked this `blocks_merge: yes`; this digest calls it **P2, non-blocking**, because the *behavior* was independently reproduced as correct on this branch by two Claude lanes — what is missing is the assertion, not the fix, and a missing assertion does not ship a bug.
4. No coverage of the MISSING-node / empty-name branch of the new arm.
5. No **cross-version** warm-cache test: every test runs both halves on the new code, so "does the forced rebuild actually re-extract rather than replay a warm pre-bump `ParsedFile`" — exactly what `SCHEMA_BUMP` guarantees — is pinned only by a literal.
6. `analysis-features.test.ts` builds a **local copy** of the descriptor array, so it cannot catch "forgot to register in production" — but `incremental-orchestration.test.ts` does catch it (verified by mutation), so there is no net gap.
7. **Pre-existing, explicitly not this PR's to fix:** there is no `java-captures-golden` fixture, while seven other languages have one under `test/fixtures/*-captures-golden/`. That is *why* this capture-shape change drifted no baseline — the absence of drift is not evidence the change is inert.

## 6. CI status — green on the new head

Merge state after the rebase is `mergeable=MERGEABLE`, `mergeStateStatus=UNSTABLE` (the `CONFLICTING`/`DIRTY` state caused by #2934 landing mid-review is resolved; `UNSTABLE` is the Vercel gate below, nothing else).

The run on `fa24b1cc8` has now settled at **42 pass / 3 skipping / 1 fail**, and the single failure is **`Vercel — Authorization required to deploy`** — the usual fork-PR authorization gate, unrelated to the code. The three skips are the expected conditional jobs (`e2e (chromium)`, `Apply conventional label`, `Validate PR title`).

Two checks failed transiently earlier in this run and have since gone green on re-run, noted only so nobody re-opens them: `Build & Push gitnexus-web` died in the **Install Cosign** step with `curl: (56) Connection died, tried 5 times before giving up` (pure network), and `tests / macos-latest (platform-sensitive) 1/3` registered a failure whose log was never retrievable while the run was in flight. Both pass now. The earlier draft of this digest flagged the macOS shard as the one thing to look at before merge — that is resolved.

The **18103/18103 non-skipped passed, 74 skipped** sticky report belongs to the **previous** head (`f90eac9e7`); the current head's suite is green on its own run.

## 7. Coverage limits — stated honestly

- **The head moved twice during this review.** Lanes analysed `2be508e79...f90eac9e7`; `origin/main` advanced (#2934, 17:09 UTC) and the author pushed the rebase `fa24b1cc8` at 17:37 UTC. All surviving findings, line anchors, version numbers and CI state were re-verified against the current head; the lane transcripts themselves were not re-run against it.
- **Four Claude lanes hit turn exhaustion** and were resumed with a "no more tools, dump what you concluded" instruction (facts-historian, risk-architect, test-ci-verifier — the last twice). Their post-resume reports restate probe output they had already seen; nothing was fabricated after resume, but they had no budget left to chase new leads.
- **Codex's integration probe failed at worker startup in its sandbox**, so its refutation of the constant-body dispatch hypothesis is `[code-read]`. Two Claude lanes `[reproduced]` the same conclusion, which is why it is reported as refuted.
- **The local GitNexus MCP index was stale** (it returned pre-PR helper content), so its `detect_changes` result was correctly *not* used as evidence by any lane. Good discipline; don't read the absence of a graph-level alarm as an all-clear.
- **No JDK in the sandbox**, so every javac-legality claim is spec-read against the JLS, not compiler-verified. The AST shapes behind those claims *were* reproduced.

---

**Verdict: strong PR, nothing blocking.** The Java semantics are right and were validated harder than most changes this size get. The single blocking item — the `SCHEMA_BUMP` exact clash — was found during this review and **you fixed it mid-review at `fa24b1cc8`** (64 + pin test + ledger line), verified. CI on the new head is green apart from the Vercel fork-auth gate. What is left is one scale interaction to weigh (§2: whether the fan-out cap should rank real implementors above synthesized constant bodies) plus test and documentation follow-ups, none of which block the merge.

*This is an automated multi-tool digest. Please verify each finding against the code before acting on it.*

---

gitnexus/src/core/ingestion/languages/java/captures.ts
azizur100389: Addressed in 6667d138: added annotated class extends/implements, record implements, interface extends, and explicit super() regressions, and broadened the feature/cache metadata.

gitnexus/src/core/ingestion/languages/java/captures.ts
azizur100389: Addressed in 6667d138: dispatch-cap ranking now keeps source-written owners ahead of synthetic anonymous owners, reports sampled dropped targets, and includes the 70-target abstract-enum regression retaining all 30 class implementations.

gitnexus/test/integration/resolvers/java.test.ts
azizur100389: Addressed in 6667d138: the constant-body fixture now calls through Named and pins the exact Status.label#0 and Status$1.label#0 interface-dispatch targets.

azizur100389:
Follow-up 1ed3ec1b refreshes the two intentional Java capture fingerprints introduced by the synthetic-declaration sidecar. The exact CI benchmark command now passes all 15 languages locally; Java scaling is 0.981 and java-local-types is 1.030 (budget 1.5).

PR #658, opened 2026-04-05
feat: same-arity overload disambiguation via type-hash suffix (#651)

magyargergo:
## Summary

- Add `~type1,type2` suffix to Method/Constructor node IDs when same-arity overloads with different parameter types exist in the same class (Java, C#, Kotlin, C++)
- Add `$const` suffix for C++ const-qualified method overloads via new `isConst` field extracted from tree-sitter `type_qualifier` in `function_declarator`
- TS/JS intentionally excluded — overload signatures collapse to the implementation body's node ID
- Fix sequential `findEnclosingFunction` to disambiguate same-class overloads instead of picking the first candidate
- 1710 integration + 18 unit tests pass

## Key Design Decisions

| Decision | Rationale |
|----------|-----------|
| Tilde `~` separator for type tag | Doesn't conflict with `:` `.` `#`; `extractFuncNameFromSourceId` already strips everything after `#` |
| `$const` separator for const tag | After type tag in ID format: `#<arity><~typeTag><$const>` |
| Collision-only activation | Type/const tags only appended when actual collision exists — non-overloaded methods keep existing IDs |
| TS/JS skip via `SKIP_TYPE_HASH_LANGUAGES` | TS overload signatures are declaration-only contracts, not true overloads |
| Sequential path fix | Resolve path falls through on ambiguous same-class candidates; fallback path computes full ID with typeTag + constTag |

## Test Coverage

| Language | Single-file | Cross-file | Chain | METHOD_IMPLEMENTS | Const |
|----------|:-----------:|:----------:|:-----:|:-----------------:|:-----:|
| Java     | ✅ | ✅ | ✅ | ✅ | N/A |
| C#       | ✅ | ✅ | ✅ | ✅ | N/A |
| Kotlin   | ✅ | ✅ | ✅ | ✅ | N/A |
| C++      | ✅ | ✅ | ✅ | N/A | ✅ |
| TypeScript | ✅ | ✅ | N/A | ✅ | N/A |

Each per-call-site test verifies exactly **1 CALLS edge** from an isolated caller function to the correct overloaded target (verified by `parameterTypes` on the target node).

## Files Changed (43)

**Core implementation (9 files):**
- `method-props.ts` — `typeTagForId()`, `constTagForId()`, `buildMethodProps` isConst
- `method-types.ts` — `isConst` on MethodInfo + MethodExtractionConfig
- `generic.ts` — Wire isConst from config
[...]

PR #2936, opened 2026-08-12
fix(java): materialize record component accessors

### LINKED ISSUE #2917: fix(java): materialize Record component accessors for interface dispatch
## Problem

A Java record component implicitly declares an accessor, but GitNexus does not materialize that accessor as a `Method` node. For a canonical interface implementation such as:

```java
record User(String name) implements Named { }
class Reader { String read(Named value) { return value.name(); } }
```

scope resolution emits `User IMPLEMENTS Named`, but interface-dispatch fan-out has no `User.name` target and emits no `Reader.read -> User.name` edge.

The explicit-method form works:

```java
record User(String value) implements Named {
  public String name() { return value; }
}
```

## Likely seam

Java method extraction currently recognizes explicit method/constructor declarations but not record components as implicit accessor declarations. The fix should materialize canonical accessor methods without duplicating an explicitly declared accessor.

## Acceptance criteria

- Record components produce canonical accessor `Method` nodes with correct ownership and signature.
- Interface dispatch reaches an implicit record accessor.
- Explicit record accessors remain unique and continue resolving.
- Multiple components and generic component types are covered.

Follow-up from #2916 and #2900.

azizur100389:
## Summary

- materialize implicit Java record-component accessors as owned, zero-argument `Method` nodes
- preserve explicit canonical accessors and overloads while enabling interface-dispatch fan-out
- invalidate warm Java indexes and rebaseline the scope-capture correctness fingerprint

## Validation

- 372 affected extractor, query, worker, feature, and cache tests
- 252 Java resolver tests
- 29 worker-pool integration tests
- 15-language scope-capture correctness and scaling gate
- TypeScript typecheck, Prettier, ESLint (0 errors), and staged `detect-changes`

Risk is limited to Java parse/scope emission; cache and analysis-feature version bumps force affected indexes to rebuild. Reverting the commit restores the previous graph behavior.

Fixes #2917

---

## Summary

*This appears to be a broad ingestion and method-extraction change focused on Java record component accessors. Its reach is high through shared parsing and worker paths, although no individual file is rated HIGH or CRITICAL risk.*

**🟠 HIGH blast radius. A Java ingestion and method-extraction change that reaches 53 dependents across shared parsing, worker, and cache code.**

The change is concentrated in `gitnexus/src/core/ingestion/method-extractors/generic.ts`, the hottest file with 8 changed symbols, alongside the Java analysis path in `gitnexus/src/core/ingestion/languages/java/record-components.ts`, `gitnexus/src/core/ingestion/languages/java/captures.ts`, and `gitnexus/src/core/ingestion/tree-sitter-queries.ts`. The affected areas include `Method-extractors`, `Workers`, `Java`, and `Configs`, with `buildMethod`, `JAVA_QUERIES`, and `emitJavaScopeCaptures` among the key changed symbols.

Review `gitnexus/src/core/ingestion/workers/parse-worker.ts` first, particularly `getMethodInfo`, `findEnclosingFunctionId`, and `processFileGroup`, then follow the related contracts in `gitnexus/src/core/ingestion/method-types.ts` and `gitnexus/src/storage/parse-cache.ts`. The Java resolver and method extraction tests in `gitnexus/test/integration/resolvers/java.test.ts`, `gitnexus/test/unit/method-extraction.test.ts`, and `gitnexus/test/unit/scope-resolution/java/java-captures.test.ts` are the most direct validation points.

_Added by GitNexus for PR #2936. Edit freely — this block is replaced on the next review, everything above it is left untouched._


azizur100389:
Resolved the latest-main conflicts and fixed explicit receiver parameters so `name(Record this)` suppresses the implicit accessor without duplicate declarations.

Validation: 633 affected tests passed, plus typecheck, Prettier, and the change-scope gate.

gitnexus/src/core/ingestion/languages/java/record-components.ts
```diff
@@ -0,0 +1,166 @@
...
+    receiverType: null,
+    returnType: recordComponentReturnType(component),
+    parameters: [],
+    visibility: 'public',
+    isStatic: false,
+    isAbstract: false,
+    isFinal: false,
+    annotations: [],
+    sourceFile: context.filePath,
+    line: component.startPosition.row + 1,
```
magyargergo, line 142: **P1 — a synthesized accessor evicts a source-written overload from the method map.** `[reproduced]`

Stamping the accessor with the *component's* line, combined with appending it last at `:102`, collides with `getMethodInfo`'s `${name}:${line}` key (`parse-worker.ts:762`, looked up at `:2566` and again at `:999`). Whenever a record component and a same-named body method share a physical line, the implicit arity-0 entry overwrites the explicit method's `MethodInfo`.

Trigger — ordinary compact Java, and the wrapped-header form:
```java
record P(int x, int y) { int x(int s) { return x*s; } }

public record Pair(
    String left,
    String right) { public String right(int n) { return right.repeat(n); } }
```
Base-vs-head differential through the real pipeline:
```
BASE  NODE Method:Pair.java:Pair.right#1  pc=1   EDGE CALLS Caller.p#1 -> Pair.right#1   correct
HEAD  NODE Method:Pair.java:Pair.right#0  pc=0   EDGE CALLS Caller.p#1 -> Pair.right#0   WRONG
      (Method:Pair.java:Pair.right#1 is gone)
```
A JLS-legal source-written method disappears and its one-argument call rebinds to the zero-argument accessor — a **fabricated** edge, which this file's own #2729 comment ranks as worse than a missing one. `parse-worker.ts:994-1008` reads the same map, so calls *inside* the lost overload are misattributed too. Moving the body method down one line makes both nodes reappear, which is the tell that the key, not the JLS modelling, is the defect.

**Fix.** Widen the discriminator: carry `column` (or `startIndex`) on `MethodInfo`, set it from `node.startPosition.column` for explicit methods and `component.startPosition.column` here, and key `getMethodInfo` by `name:line:column` at both lookup sites. That retires the whole collision class, including the base-level same-line-overload case every language with overloads shares. Staying inside this PR's blast radius, the minimal alternative is to refuse to mint an implicit accessor whose `(name, line)` an explicit method already claims — losing the implicit entry beats losing a source-written one.

Note **reordering to `[...implicitAccessors, ...extracted.methods]` is not a fix**: the explicit method would win the slot, but the accessor's own definition capture would then resolve to `#1` and collide from the other side.

Any regression test must use a single-line or wrapped-header record — the current multi-line fixtures cannot express this.

gitnexus/src/core/ingestion/languages/java/record-components.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  const parameters = component.parent;
+  const recordNode = parameters?.parent;
+  const name = captureMap['name']?.text;
+  return (
+    parameters?.type === 'formal_parameters' &&
+    recordNode?.type === 'record_declaration' &&
+    name !== undefined &&
+    explicitZeroArgAccessorNames(recordNode).has(name)
```
magyargergo: **P3 — this rebuilds the whole set on every component capture.** `[reproduced]`

`shouldSkipJavaRecordComponentDefinition` is called once per component from `parse-worker.ts:2115`, and each call re-walks the record body: O(components x body members) per record. The scope-side twin at `:93` computes the identical set **once per record** and reuses it, so this is the same rule implemented twice with only one of them hoisted.

Measured, driving the hook exactly as the worker does:
```
N=50   6.5ms     N=100  28.6ms    N=200  96.0ms
N=400  375.7ms   N=800  1546.7ms      (~4x per 2x input)
```
Calibrated to P3 rather than blocking: the type guard short-circuits every non-record capture, real records carry single-digit component counts where this is microseconds, and the curve only bites past several hundred components in one record — which javac itself rejects at the 255-parameter canonical-constructor limit. Worth fixing anyway because it is a one-line `WeakMap<SyntaxNode, Set<string>>` with a same-file precedent.

Worth knowing: neither bench arm can see this. The `java-local-types` scale unit generates no `record` at all, and the `java` arm is fixed at one component and zero explicit overrides, so the committed 1.085 / 0.997 figures gate a different path entirely.

gitnexus/src/core/ingestion/languages/java/record-components.ts
```diff
@@ -0,0 +1,166 @@
...
+  return {
+    name,
+    receiverType: null,
+    returnType: recordComponentReturnType(component),
+    parameters: [],
+    visibility: 'public',
+    isStatic: false,
+    isAbstract: false,
+    isFinal: false,
+    annotations: [],
```
magyargergo: **P3 — component annotations never reach the implicit accessor.** `[reproduced]`

`annotations: []` is hard-coded. Per JLS 8.10.3 / 9.7.4, a component annotation whose target includes `METHOD` or `TYPE_USE` propagates to the generated accessor, and javac's `name()` carries it:
```java
@Target({ElementType.RECORD_COMPONENT, ElementType.METHOD})
@interface Marker {}
record User(@Marker String name) {}
```
GitNexus emits `annotations: []`. An *explicit* accessor with the same annotation does get it, so two JLS-identical declarations produce different node properties based only on how the author spelled them.

P3 rather than blocking: the one JS-layer analysis consumer of stored node annotations (`di-extractors/spring.ts:228`) matches `Property` nodes, not `Method` nodes, so no existing analysis breaks. It is not fully inert though — `mcp/local/local-backend.ts:4395` selects `n.annotations` for Method-like nodes, so an MCP `context` call on an implicit accessor reports none.

Propagate annotations applicable to `METHOD`/`TYPE_USE`, leaving `RECORD_COMPONENT`-only ones behind, and add one positive and one negative target test.

gitnexus/src/core/ingestion/tree-sitter-queries.ts
```diff
@@ -1205,6 +1205,17 @@ export const JAVA_QUERIES = `
 (record_declaration name: (identifier) @name) @definition.record
 (annotation_type_declaration name: (identifier) @name) @definition.annotation
 
+; Canonical record-component accessors are implicit public zero-argument methods.
+(record_declaration
+  parameters: (formal_parameters
+    (formal_parameter
+      name: (identifier) @name) @definition.method))
```
magyargergo, line 1212: **P3 — `name: (identifier)` also matches tree-sitter's zero-width `MISSING identifier`.** `[reproduced]`

Error recovery synthesizes a `MISSING identifier` node with empty text, which satisfies this pattern, so a malformed component mints an empty-named `Method`. Dropping one type — an ordinary typo:
```java
record M(int x, y) {}
```
yields a node `Method:M.java:M.#0` with `returnType: "y"`. Base produced nothing here. Both paths emit it (extractor and scope capture), and nothing downstream rejects it — `getLabelFromCaptures` checks the name capture's *presence*, not its emptiness.

Fix: gate on a non-empty identifier. Gating the scope path on `nameNode.type === 'identifier'` closes this from the other side too, and would also fix the `underscore_pattern` divergence noted in the review body — `recordComponentNameNode` reads the name by *field*, which admits `underscore_pattern`, while this query demands `(identifier)`, so `record R(int _) {}` produces a `SymbolDefinition` with no graph node behind it.

gitnexus/bench/scope-capture/baselines.json
```diff
@@ -149,15 +150,16 @@
...
     "scaling_budget": 1.5,
     "_rebaselined_2935_synthetic_declarations": "PR #2935 review follow-up: the local-type stress corpus includes synthesized anonymous declarations, which now carry the presence-only @declaration.is-synthetic sidecar. DIGEST DRIFT ONLY, NOT A CAPTURE-SET CHANGE. Prior 8c50bbc83dff4f7f5abd06078aa6abc6b64af05fddb17ee826b5f3df3d346633 -> 560734cd053fb4f4b23aa04bc7870c22089a8deedb0217fa9c1b4db689e02a97; CI scaling 1.002 < 1.5.",
+    "_rebaselined_2917_record_component_accessors": "#2917: the focused local-type fixture corpus contains local records, so their implicit component accessors add the same bounded scope/declaration captures as the general Java corpus. No local-type naming logic changed. Prior 8c50bbc83dff4f7f5abd06078aa6abc6b64af05fddb17ee826b5f3df3d346633 -> 3e22f368a4ee139be7cb91ff4fb77ddadf60c55efe8d66955ec81f366a46e460; scaling 1.032 < 1.5, capture_groups_fp 680. Re-measured on top of #2935's is-synthetic sidecar after merging origin/main: 560734cd053fb4f4b23aa04bc7870c22089a8deedb0217fa9c1b4db689e02a97 -> bdde823fa725e636e257940efb4c8655aa23124c1727cbaa8856d1ad8f71729e; scaling 0.997 < 1.5, capture_groups_fp 680.",
```
magyargergo, line 162: **P3 (nit) — this note asserts a field the entry does not store.** `[reproduced]`

The note claims `capture_groups_fp 680`, but the `java-local-types` object holds only `fingerprint` and `scaling_budget`, so nothing persists that number and no reviewer can check it. The sibling `java` entry does store `"capture_groups_fp": 3560`, matching its own note.

Either add `"capture_groups_fp": 680` here, or drop the claim from the prose — the two entries should be treated consistently.

magyargergo (commented):
# Tri-review — PR #2936 `fix(java): materialize record component accessors`

Reviewed at head `766359e16` against base `56d9003fe` (15 files, +421/−25).

**Final verdict: `not production-ready`** — one P1 correctness regression, detailed below, blocks merge on its own. Everything else found is P3.
**Branch hygiene: `merge-from-main commit present but harmless and merge-safe`** — seven commits since base, five of them merges from main, two substantive; no unrelated functional churn reaches the reviewable diff.
**Merge state: `review blocked`** — CI is green at head and GitHub reports `MERGEABLE`; the block is the unresolved P1, not the branch or the pipeline.

**How this was reviewed — two engines, not five.** Claude ran the GitNexus PR swarm (facts, branch-hygiene, risk-architect, test/CI, security, docs/DoD) plus the `ce-code-review` personas (correctness, adversarial, performance, simplicity, testing, project-standards). Codex `gpt-5.6-sol` at `xhigh` ran the same swarm spec in Solo mode, `ce-code-review` in report-only mode, and its own adversarial review. All three Codex legs produced real findings (verified, not empty/skill-not-found markers). Where a finding below is called *cross-engine*, it means one Codex leg and one Claude lane reached it independently — corroboration between two engines, and nothing more. Two Claude lanes bundled the parse worker themselves (`esbuild` for both base and head trees) and ran the real pipeline end-to-end, so several results are true base-vs-head differentials rather than code reads.

---

## What this PR gets right

This is careful work, and a large share of the review's effort went into attacking it and failing.

- **The architecture is right, and all four mechanisms are load-bearing.** The structure query, the suppression hook, the scope-capture synthesizer and the `methodExtractor` wrapper are not redundant: the structure phase is the only path that mints a graph `Method` node (`resolveDefGraphId` never synthesizes one — every arm returns an existing `nodeLookup` hit or `undefined`), while the scope path only resolves references. Drop either half and the feature silently stops working. Record components are the first construct in this repo where the same syntax is *sometimes* the definition and *sometimes* superseded by an explicit sibling, which is exactly what a tree-sitter predicate cannot express and a runtime hook can.
- **The new hook is placed per the house rules.** `shouldSkipDefinitionCapture` (`language-provider.ts:217-224`) is worded language-neutrally, `parse-worker.ts:2115` reads it generically, and all Java logic stays in `languages/java/record-components.ts`. It mirrors the existing `shouldSkipClassCapture` precedent invoked ~20 lines below it. AGENTS.md:42 is satisfied — an adversarial lane specifically hunted for a language name leaking into shared code and found none.
- **The JLS modelling is correct on every shape probed.** A sweep over 16 record shapes plus 45 hand-built and 6000 randomized fuzz cases confirms: varargs accessors return the array type (`int...`→`int[]`, `List<String>...`→`java.util.List<String>[]`, `int[]...`→`int[][]`), receiver parameters (`public int x(R this)`) correctly count as arity 0 so the explicit accessor suppresses the implicit one, compact constructors do not confuse suppression, and local/nested/in-interface/in-enum/generic records all work. `isFinal: false` is right (the *component field* is final; the accessor is not).
- **The feature is not inert.** Verified at resolver-output level, not capture level: head emits `METHOD_IMPLEMENTS Rec.name#0 -> Named.name#0` and `CALLS Use.viaIface#1 -> Rec.name#0` where base emitted neither. Across all 60 `java-*` resolution fixtures the base→head delta is strictly additive — zero nodes and zero edges lost, including `ACCESSES` — with exactly one replacement, `Use.go -> Named.name` becoming `Use.go -> Rec.name`, which is the improvement the PR exists for.
- **Cache invalidation is correct and doubly redundant.** `SCHEMA_BUMP 66→67` invalidates both the chunk cache and the durable ParsedFile index, *and* `java.record-component-accessors` reports `missing:` on a pre-PR index, which forces `options.force = true` before the `alreadyUpToDate` fast path. A stale index rebuilds; it cannot silently serve accessor-less ParsedFiles. 67 was verified free against `origin/main` (66) and against all 50 open PRs (max other claim 64).
- **The benchmark rebaseline measures something real.** `measure.mjs --check` was executed: `PASS (15 languages)`, Java reporting exactly the committed values, and *only* the two Java fingerprints moved — a discriminating instrument, not an inert one. The "+2 capture groups per component" claim checks out arithmetically (small 5755→6255 = 2×250, large 18405→20005 = 2×800).
- **Toolchain and conventions are clean.** `prettier --check` passes on all 15 files; `eslint` reports 0 errors and no warning on any added line (the new file has zero); `tsc -p tsconfig.json` shows the same 4 pre-existing errors as the base, byte-identical; no `any`, no `if`-branching in tests, no CHANGELOG edit, no non-determinism.

---

## Headline: a synthesized accessor evicts a source-written overload from the method map

**P1 · `[reproduced]` · cross-engine + independently re-reproduced during synthesis.**

`getMethodInfo` keys every callable's metadata by `${name}:${line}` (`parse-worker.ts:762`) and looks it up by the same key (`parse-worker.ts:2566`). The synthesized accessor is stamped with the **component's** line (`record-components.ts:81`) and appended **last** (`record-components.ts:102`), so whenever a record component and a same-named body method share a physical line, the implicit arity-0 entry overwrites the explicit method's `MethodInfo`.

Reproduced at the extractor level during synthesis, driving the worker's map build verbatim:

```
record P(int x, int y) { int x(int s) { return x*s; } }
  extracted: [x/arity1/line1, x/arity0/line1, y/arity0/line1]
  map:       [x:1 -> arity 0, y:1 -> arity 0]        <- explicit x(int) LOST

public record Pair(
    String left,
    String right) { public String right(int n) { return right.repeat(n); } }
  map:       [right:3 -> arity 0, left:2 -> arity 0]  <- explicit right(int) LOST
```

And end-to-end through the real pipeline on both sides of the diff:

```
BASE  NODE Method:Pair.java:Pair.right#1  pc=1     EDGE CALLS Caller.p#1 -> Pair.right#1   correct
HEAD  NODE Method:Pair.java:Pair.right#0  pc=0     EDGE CALLS Caller.p#1 -> Pair.right#0   WRONG
      (Method:Pair.java:Pair.right#1 is gone)
```

A source-written, JLS-legal method disappears from the graph and its one-argument call site is re-pointed at the zero-argument accessor — a **fabricated edge**, which this file's own #2729 comment ranks as worse than a missing one. `findEnclosingFunctionId` reads the same poisoned map (`parse-worker.ts:~1004`), so calls *inside* the lost overload are misattributed too. Moving the body method down one line makes both nodes reappear, which is the tell that the key — not the JLS semantics — is the defect.

Three things make this worth blocking on despite needing a shared line:

1. **It contradicts the PR's own stated behaviour.** The description claims it preserves "explicit canonical accessors *and overloads*". For a shared-line overload it does not.
2. **The collision winner inverted.** The `name:line` key is a pre-existing hazard — at base, two same-line overloads in a plain `class` already collapse — but there *both* participants were source-written, so the loser was at worst a duplicate declaration and the casualty was a missing edge. This PR appends a synthetic entry last, so it wins the slot and the casualty becomes a source-written method, turning a missing edge into a fabricated one. It also *guarantees* a same-name arity-0 entry on the record header's line, where before it took a human writing two overloads on one line.
3. **No test can see it.** The new overload tests assert the *extractor's* output list, which is correct on both formattings; the loss happens one layer later when the worker collapses that list into the map. Separately, the scope-side capture test (`java-captures.test.ts`, "keeps an overload alongside the implicit zero-argument accessor") uses exactly this one-line record shape and passes — because the scope path never builds the worker's `name:line` map. So the shape is realistic and already in the suite, and no assertion in the PR crosses the boundary where it breaks.

**Suggested fix.** The durable fix is to widen the discriminator: carry `column` (or `startIndex`) on `MethodInfo`, set it from `node.startPosition.column` for explicit methods and `component.startPosition.column` for the synthesized accessor, key `getMethodInfo` by `name:line:column`, and pass the column at both lookup sites. That retires the whole collision class, including the base-level same-line-overload case that every language with overloads shares. If you want to stay inside this PR's blast radius, the minimal alternative is to refuse to mint an implicit accessor whose `(name, line)` is already claimed by an explicit method — losing the implicit entry is strictly better than losing a source-written one. Note that **reordering to `[...implicitAccessors, ...extracted.methods]` is not a fix**: the explicit method would win the slot, but the accessor's own definition capture would then resolve to `#1` and collide from the other side. The regression test must use a single-line or wrapped-header record; the current multi-line fixtures cannot express the bug.

---

## Findings to post inline

| # | Severity | Path | Line | Finding | Evidence |
|---|---|---|---|---|---|
| 1 | **P1** | `gitnexus/src/core/ingestion/languages/java/record-components.ts` | 81 | Synthesized accessor carries the component's line and is appended last, so it overwrites a same-line explicit overload in `getMethodInfo`'s `name:line` map (`parse-worker.ts:762` / `:2566`) — the source-written method vanishes and its callers rebind to the zero-arg accessor | `[reproduced]` — base-vs-head pipeline differential + extractor/map reproduction |
| 2 | P3 | `gitnexus/src/core/ingestion/languages/java/record-components.ts` | 164 | `shouldSkipJavaRecordComponentDefinition` rebuilds `explicitZeroArgAccessorNames(recordNode)` on every component capture — O(components × body members) per record, while the scope-side twin at `:93` hoists the identical call out of the loop | `[reproduced]` — 50→800 components measured at a clean ~4× per 2× input; 275 ms for the worst legal record (255 components) |
| 3 | P3 | `gitnexus/src/core/ingestion/languages/java/record-components.ts` | 79 | `annotations: []` is hard-coded, so component annotations applicable to `METHOD`/`TYPE_USE` never reach the implicit accessor (JLS 8.10.3 propagates them; javac's generated `name()` carries them). An explicit accessor with the same annotation *does* get them — same-semantics declarations produce different node properties depending on spelling | `[reproduced]` — `record U(@Marker String name)` yields `ann: []` |
| 4 | P3 | `gitnexus/src/core/ingestion/tree-sitter-queries.ts` | 1212 | The pattern matches `name: (identifier)`, which also matches tree-sitter's zero-width `MISSING identifier` from error recovery, so a malformed component mints an empty-named `Method` node. `record M(int x, y) {}` produces `Method:M.java:M.#0` with `returnType: "y"`; base produced nothing. Nothing downstream rejects an empty name (`getLabelFromCaptures` checks presence, not emptiness) | `[reproduced]` — base-vs-head pipeline differential; re-confirmed at extractor level (`{n:"", rt:"y"}`) |
| 5 | P3 *(optional inline)* | `gitnexus/bench/scope-capture/baselines.json` | 162 | The `java-local-types` rebaseline note asserts `capture_groups_fp 680`, but that entry stores only `fingerprint` and `scaling_budget` — the claim cannot be checked against any persisted field. The sibling `java` entry does store `capture_groups_fp: 3560`, matching its own note | `[reproduced]` — key sets compared |

On #2 and #3: Codex rated both **P2 and merge-blocking**. Both were calibrated down here on measured evidence. For #2, the cheap type guard short-circuits every non-record capture, real records carry single-digit component counts where the cost is microseconds, and the curve only bites past several hundred components in one record — which javac itself rejects at 255 canonical-constructor parameters. It is still worth fixing: it is a one-line `WeakMap<SyntaxNode, Set<string>>` with a same-file precedent. For #3, a grep for the TypeScript pattern `node.properties.annotations` finds exactly one JS-layer *analysis* consumer (`di-extractors/spring.ts:228`), and it matches on `Property` nodes, not `Method` nodes — so no existing analysis breaks. There is a second reader worth naming: `mcp/local/local-backend.ts:4395` selects `n.annotations` from Method-like nodes in Cypher, so an MCP `context` call on an implicit accessor reports no annotations even where javac's generated method carries them. That is metadata display rather than analysis, so P3 stands — but the downgrade rests on "no analysis consumer", not on "no consumer at all". It is still a real JLS divergence and cheap to close.

---

## Lower-priority notes (body only, not inline)

- **`isExported` disagrees between an implicit and an explicit accessor** `[reproduced]`. `javaExportChecker` walks up from the name node looking for a `modifiers` sibling containing `public`; a `formal_parameter` has none, so the walk reaches the `record_declaration` and returns the *record's* export-ness. On this PR's own integration fixture: `Explicit.name#0 isExported=true` vs `User.name#0 isExported=false`, both `visibility: 'public'`. Two JLS-identical public accessors land on opposite sides of a public-API filter based only on whether the author wrote them out. Pre-existing looseness that the PR extends rather than invents (an interface method is also `isExported=false` at base), hence P3 — but the two spellings should agree, and an `isExported` parity assertion would pin it.
- **Structure and scope paths disagree on an `underscore_pattern` component** `[reproduced]`. `recordComponentNameNode` reads the name by *field*, which admits `underscore_pattern`; the query demands `name: (identifier)`. `record R(int _) {}` therefore yields a `SymbolDefinition` with no graph node behind it. No dangling edge was observed, and `_` is illegal as a record component name, so this is a latent phantom binding on non-compiling input. Gating the scope path on `nameNode.type === 'identifier'` makes the two paths share one predicate — and would also close #4 above from the other side.
- **Interface-dispatch fan-out now has a new population pressure.** Every record gains one dispatch target per (record, interface method) pair. On a corpus with 32 class implementors plus 5 record implementors of `Named.name`, five real source-written classes lost their `CALLS` edges to `targets.length = MAX_INTERFACE_DISPATCH_FANOUT` (default 32). This is the pre-existing cap doing its job on a now-larger *correct* target set: record accessors are genuine JLS implementors, the truncation is loud (`dispatchFanoutSkipped` plus a bounded sample of dropped names), and the outcome is deterministic across worker-pool sizes. Which side gets evicted is decided by discovery order, i.e. lexical file order. Worth one line in the PR description so the next reader does not file it as a regression: repos sitting just under 32 implementors of a common name (`name()`, `id()`, `value()`) will start crossing the cap on the next analyze with no source change.
- **The two varargs type-selectors diverge — and the new one is the better half.** `recordComponentReturnType` picks a `spread_parameter`'s type by blacklist; the shared `extractJavaParameters` picks it by whitelist of six node types. They agree today, but the blacklist is correct on `int[]...` and `List<String>[]...`, whose type node is an `array_type` — a node type the shared whitelist does **not** list, where it would return null. So consolidating onto `extractJavaParameters` as-is would be a regression. If the duplication is worth removing, extract one shared blacklist-style selector and have `extractJavaParameters` adopt it, not the reverse. Filed as a reusability note only; the new code is not defective.
- **Stale sentence in the SCHEMA_BUMP ledger** (`gitnexus/src/storage/parse-cache.ts:510-511`): "#2936 uses 65 for its independent record accessor shape" is inherited from #2935 and no longer true — this branch now claims 67 and nothing holds 65. The equivalent sentence *was* scrubbed from `incremental-parse-cache.test.ts` by this PR; `parse-cache.ts` was not. Cosmetic, one line, and the ledger exists precisely to stop the next author mis-reading which values are held. Anchor at the new paragraph, `parse-cache.ts:513`.
- **`defaultLabel` is declared on the hook but ignored by its only implementation.** Type-safe and intentional, but a future implementor gets no signal; either say "implementations may ignore it" in the doc comment or name it `_defaultLabel`. Also worth knowing that the parameter is exercised by no test.
- **Re-check `SCHEMA_BUMP` immediately before merge**, per the file's own rule. 67 was free at `origin/main@9fa39bad5` against all 50 open PRs, but this repo has a documented history of exact clashes on this constant, and #2939 currently claims a stale 64 that it will have to renumber.

---

## Suspicions the review raised and then refuted

Recording these so they are not re-litigated:

- **The missing `@declaration.is-synthetic` tag is not an omission.** The #2935 tie-break at `receiver-bound-calls.ts:517` reads `subDef.isSynthetic` on the **owner type**, not the member. A record is a source-written owner, so its accessors already sort at priority 0 alongside class implementors; emitting the tag on the accessor would have had no effect on that sort at all.
- **Zero-component records are safe.** `record Empty() {}` → no captures, no crash. An untested path, not a defect.
- **Malformed Java does not crash anything.** Truncated records, unbalanced parens, type-less components, ERROR-recovery regions: no throw on any path. All five new AST walks are null-guarded and non-recursive.
- **The structure and scope suppression paths do not diverge in general.** A differential harness over 45 hand-built shapes plus 6000 seeded randomized/corrupted cases found `divergences=0`. The one asymmetry (the scope path's extra `returnType === null` bail) is unreachable, because `formal_parameter` has a required `type` field. The only genuine disagreement found anywhere is the `underscore_pattern` case above.
- **Clone-safety across the worker boundary is a non-issue.** `assertCloneable` applies to the capture side-channel return value, never to the provider, which is imported inside the worker module graph. Empirically the hook executed worker-side in every end-to-end run.
- **Determinism holds.** Identical drop sets at `GITNEXUS_WORKER_POOL_SIZE=1/2/8`; the fuzzer is seeded and reproducible; no unordered-`LIMIT`-class nondeterminism introduced.
- **`javaRecordMethodExtractor` is not dead code.** `javaMethodConfig.typeDeclarationNodes` includes `record_declaration`, and `findEnclosingClassNode` returns the `record_declaration` for a component, so `extract` is reached; the accessor nodes' `visibility`/`isStatic`/`isFinal` can only come from `implicitAccessorInfo`.
- **No language name leaked into shared code**, and no security surface changed: `git diff --check` clean, zero bidi/hidden-Unicode hits, no workflow/Dockerfile/secret changes, CodeQL and gitleaks green.
- **Over-suppression by `explicitZeroArgAccessorNames`** (it ignores `static`, `private`, `native`, `throws` and return type) is unreachable from compilable Java: every shape that would make the suppression wrong is already a javac error. The scan is limited to direct children of the record body, so a zero-arg `x()` inside a nested class does not suppress the accessor.
- **Raw return-type text is the house convention**, not something this PR introduced — Go and C++ build `@declaration.return-type` with the same `syntheticCapture(…, rawText)` pattern.
- **The extractor-wrapper spread is safe** — `createMethodExtractor` returns an object literal with no `this`, so spreading preserves behaviour exactly; the 60-fixture differential confirms zero lost nodes or edges for classes, interfaces, enums and annotation types.

---

## Test gaps

1. **No fixture anywhere places a record component and a body method on the same source line** — every overload fixture uses a multi-line template literal, which is exactly the formatting that hides the P1.
2. **No test crosses the extractor→worker-map boundary.** Both overload tests assert the extractor's methods array (correct on both sides); the graph built from the same source is never asserted. This is the repo's own recorded "a census over bucket membership is blind to a resolver-OUTPUT regression" shape.
3. **`shouldSkipJavaRecordComponentDefinition` has no direct unit test** — zero references under `test/`. It is a pure `CaptureMap → boolean` function needing no worker pool, and it is the single highest-value missing test in the PR. Its only indirect coverage is `expect(explicitNames).toHaveLength(1)` in the integration test, which is satisfied even if the hook never fires, because both definitions mint the same id and `addNode` is first-write-wins. Assert the surviving node's `startLine` instead — that distinguishes suppression from id collision.
4. **No differential test asserts the two suppression paths agree.** The PR's core design is two independent implementations of one JLS rule; a ~40-line table-driven test feeding the same sources to both would have caught the `underscore_pattern` divergence and guards against future drift.
5. **No coverage for fan-out-cap behaviour** — the new integration test asserts a two-element sorted fan-out on a two-implementor corpus and cannot observe the cap.
6. **No malformed-input, zero-component, or `isExported`-parity tests.** Each is a one-line fixture.
7. **The `java-local-types` scaling arm generates no `record` at all**, and the general `java` arm is fixed at one component and zero explicit overrides — so the O(C×M) shape in finding #2 has no CI gate of any kind.

---

## CI status at head

**CI is green at head `766359e16`.** Every code gate passes: `quality/typecheck`, `lint`, `format`, ubuntu coverage shards 1/3–3/3 plus coverage merge, `tests / benchmarks (GITNEXUS_BENCH)`, macOS and Windows platform-sensitive 1/3–3/3, tree-sitter ABI on all three platforms, and `Build & Push gitnexus`. The only non-passing entries are `Vercel` (fails on missing deploy authorization for a fork contributor — not a code gate) and `e2e / e2e (chromium)` (skipping, not failing).

So the P1 below is a defect the suite does not currently catch, not a red build. The suite counts quoted in the PR body (372 / 252 / 29) still do not reconcile with the merge commit's own numbers (255 java resolver tests) — read those specific figures as stale, but the green run at head is real evidence and supersedes the earlier CI comment posted against the first commit (`8ac770cba`).

## Coverage limits of this review

Current visible state is incomplete. I could verify the extractor and scope-capture behaviour, the base-vs-head graph differential over the `java-*` fixture corpus, and CI's state at head, but not the repository's own integration suite as CI runs it, the behaviour on a real large record-heavy Java codebase, or the `SCHEMA_BUMP` ledger's state at the moment of merge.

The integration suite (`test/integration/resolvers/java.test.ts`, including the new #2917 test) could not run in most lanes: it spawns the real worker pool and needs a compiled `dist/.../parse-worker.js`, which a worktree cannot build. Verified as environment-only by running an unrelated pre-existing test in the same file and getting an identical failure. Two lanes worked around it by bundling the worker for both base and head with `esbuild` into `/tmp` and driving `runPipelineFromRepo` with `workerUrlForTest` — those results are genuine end-to-end differentials. The unit suites did run: 360 tests across the four touched unit files, all passing on head. `tsc -p tsconfig.test.json` has a large pre-existing error surface that is not a clean CI-gated baseline, so it was compared against base rather than read absolutely. Nothing in the repo was modified, staged, built or committed by any lane.

---

*Automated multi-tool digest (Claude swarm + `ce-code-review` personas; Codex `gpt-5.6-sol`/`xhigh` swarm + `ce` + adversarial). Verify the findings against the code before acting on them.*

magyargergo:
@azizur100389 working on them.

magyargergo:
## Review findings addressed — `b3d1edec9`

All five inline findings are fixed on the branch. Verified: `measure.mjs --check` **PASS (15 languages)**, prettier clean, eslint 0 errors, `tsc -p tsconfig.json` shows no new errors, and the affected unit suites pass (2430 passed; the 20 failures in `type-only-import-chain` / `function-local-import-chain` / `finalize-algorithm` reproduce identically on a clean checkout of this branch — they are stale linked `gitnexus-shared` types in a worktree, not related to this PR).

### P1 — the map key, and a second language

Fixed by widening the key rather than by refusing to mint the accessor, because the narrow fix would have made the graph depend on source formatting: the same program written across two lines would keep a node that the one-line spelling drops.

The investigation also turned up **the same collision in C#**, with no record code involved:

```csharp
class Point(int x, int y) { public Point(int x) : this(x, 0) { } }
```

`extractPrimaryConstructor` is appended last at the owner's `parameter_list`, so it evicted the explicit constructor exactly the way the record accessor evicted the explicit overload. `MethodInfo` now carries a required `column`, and `getMethodInfo` keys on `name:line:column` via one `methodInfoKey` helper. Required rather than optional was deliberate — an absent column would key an entry no lookup could ever reach, turning a compile error into a silent whole-language loss of method enrichment. All three lookup sites moved in the same commit, per the lockstep docblock in that file.

Regression test is the single-line record end-to-end (`java.test.ts`), which asserts `[0, 1]` arities where the old key produced one node.

### Empty-named nodes — three emitters, not two

My review said two paths. There are three: the tree-sitter query, the scope synthesizer, **and `javaRecordMethodExtractor`**. My suggested `nameNode.type === 'identifier'` gate would also not have been enough — tree-sitter's zero-width MISSING token *is* an `identifier`. Both the emptiness and the type check are needed, and the grammar declares `underscore_pattern` as the other admissible name node, so `type === 'identifier'` is the exact complement rather than a heuristic. One `isRecordComponentName` predicate now gates all three, following the `javaBaseLookupNameNode` shape #2935 introduced for the same defect class. Two extra shapes are covered too: `record W(int... ) {}` and `record S(int... _) {}`.

### Correction to my earlier comment on annotations

I wrote that `mcp/local/local-backend.ts:4395` surfaces method annotations, so an implicit accessor would report none. **That was wrong** — `annotations` is not a column in `METHOD_SCHEMA`/`FUNCTION_SCHEMA` (`src/core/lbug/schema.ts`), so that query's `n.annotations` selects a column that does not exist and can never surface anything. The property lives only in the in-memory graph for one analyze run. The finding and the fix are unchanged, and the severity if anything drops further; the mechanism I cited was not real. (Separately: that `RETURN` also asks for `visibility`, `parameterTypes`, `isStatic` and several other non-columns under a bare `catch`, which looks like a pre-existing latent defect worth its own issue — not this PR's.)

The fix reuses the shared `extractAnnotations` helper and is deliberately over-approximate, commented as such: `@Target` lives in another file and parsing is per-file, so the target set is not knowable at synthesis time.

### Not changed, and why

The design pass proposed also guarding `getLabelFromCaptures` in shared code, which would fix a **pre-existing** empty-name bug (`class C { void () {} }` mints an empty-named Method on `main` today, independent of this PR). A census over 11 languages and ~157k definition matches found zero present-but-empty `@name` captures on clean source, so the guard looks safe — but it is unrelated to this PR and belongs in its own change rather than widening a contributor's branch. Worth filing.

Fingerprints did not move, so there is nothing to rebaseline: the bench corpus contains no degenerate components, confirmed by diffing the new predicate's selection against the old across the whole corpus (0 differences). `SCHEMA_BUMP` stays **67** — this branch's existing claim already covers the changed worker output. Per the ledger's own rule, re-check 67 against `origin/main` immediately before merging; `#2939` still claims a stale 64 that it will need to renumber.

---
*Automated follow-up; verify before merging.*

magyargergo:
## gitnexus-check findings — one fixed, one refuted

**Fixed in `bf4b2152e`** — *"New declaration displaces the collision-group JSDoc"* (`method-props.ts:25`, P3). Correct, and mine.

The root cause predates this branch: the block describing the `~type1,type2` same-arity discriminator has been stranded since #658, when `buildCollisionGroups` was inserted between it and the `typeTagForId` it actually documents. Adding `methodInfoKey` parked it above a *second* unrelated function, which is what the bot caught. Rather than shuffle my block around the orphan, I moved the comment down to the function it describes — that clears the finding and the pre-existing strand together. No behaviour change; `detect_changes` reports 1 symbol, low risk, 0 affected processes.

**Refuted** — *"1 symbol removed but still referenced: `import-cycles.ts::findCyclePath` — still called from `pipeline-phases/runner.ts`"*.

There are two different symbols with that name. The one `runner.ts:75` calls is **file-local**, declared at `runner.ts:99`:

```
gitnexus/src/core/ingestion/pipeline-phases/runner.ts:75:    const cyclePath = findCyclePath(remaining, phaseMap);
gitnexus/src/core/ingestion/pipeline-phases/runner.ts:99:function findCyclePath(
```

The removed one is `core/graph/import-cycles.ts::findCyclePath`, deleted on `main` by #2934's cycle-detection rewrite — not by this PR, and never imported by `runner.ts`. Both `runner.ts` references exist identically on `origin/main`, and `tsc --noEmit` is clean, which it would not be for a genuinely dangling call. The cross-branch graph comparison matched the two by name across different files.

The rest of the structural section (181 symbols added / 23 removed) is merge drift from `origin/main` — `gitnexus-shared` finalize-algorithm additions, the CLI `AnalyzeOptions`/`GeneratedSkillInfo` moves, the DI-extractor index refactor — none of it originates in this branch's diff.

---
*Automated follow-up; verify before merging.*

PR #2955, opened 2026-08-13
fix(java): resolve imports from declared packages

### LINKED ISSUE #2910: fix(java): JDK imports can resolve to coincidentally-named local files (no #1881-style gate)
Found while indexing Java's import resolution for #2908. **Pre-existing, not introduced by that work** — the hoist preserved it deliberately and pins it in the new parity test.

## The defect

`resolveJavaImportTarget` ends in a progressive prefix-stripping loop that walks down to the bare class name. Nothing gates it on whether the import names a package that actually exists in the repository, so a JDK or third-party import resolves to any local file with a matching tail:

```
import java.util.List;   →  util/List.java          (a local file, in a repo with no java/ package)
```

Verified: Java has **no** equivalent of C#'s namespace gate — `grep -rn "csharpSuffixFallbackAllowed\|evidence" src/core/ingestion/languages/java/` returns nothing, while `languages/csharp/import-target.ts:101` runs `csharpSuffixFallbackAllowed(targetRaw, evidence)` before its own direct-match and stripping legs.

This is exactly the bug **#1881** fixed for C#: *"`using System.Threading.Tasks;` cannot spuriously match a local `Tasks.cs`"*. The same reasoning applies verbatim to `java.util.*`, `javax.*`, `com.google.*`, `org.springframework.*` — every one of which is a stripping-loop candidate today.

## Impact

Wrong `IMPORTS` edges, which then propagate: `finalize-algorithm` links bindings through them, `propagateImportedReturnTypes` chains return types, and `buildPopulatedMethodDispatch` derives dispatch tables from the resulting graph. Nothing throws — the failure is silent misdirection, and it is most likely in exactly the repos where it matters least to be wrong about (large ones with a `util/` package).

## Fix

Mirror the C# shape. `csharp-namespace-gate.ts` + `CSharpNamespaceEvidence` (declared in-repo namespaces, gathered during the scan) is the template; Java's equivalent evidence is the set of `package` declarations already parsed out of the `.java` files. Gate the stripping loop — and probably the nested-suffix leg too — on the import's leading segments matching a declared in-repo package.

Keep C#'s fail-open behaviour for undefined/truncated evidence, so repos that were not fully scanned keep their legitimate edges.

## Acceptance

- `import java.util.List;` in a repo containing `util/List.java` and no declared `java.util` package resolves to `null`.
- An in-repo import that legitimately needs prefix stripping (a repo laid out `User.java` for `com.example.User`) still resolves — that is the case the loop exists for.
- The gate is evidence-driven, not a hardcoded JDK prefix list: `java.`/`javax.` blocklists miss `com.google.*` and would wrongly block a local `java`-named package.
- `test/unit/scope-resolution/java-import-target-parity.test.ts` (added in #2908) currently pins the CURRENT behaviour, including `java.util.List → util/List.java`. Those expectations must be updated as part of this fix, not worked around — they are there so this change is visible rather than silent.

azizur100389:
## Summary

- resolve Java imports from parsed package declarations instead of filesystem suffixes
- expand wildcard imports to every file declaring the package
- keep JDK and third-party imports unresolved when no in-repo package declares them

## Validation

- 312 focused Java and module-resolution tests passed
- TypeScript typecheck passed
- ESLint passed with zero errors
- Prettier check passed

Fixes #2910

---

## Summary

*This appears to be a broad Java import-resolution change with substantial test and benchmark coverage. Its aggregate reach is critical, although the graph does not identify a traced execution flow or a high-risk individual file.*

**🔴 CRITICAL blast radius. A Java ingestion and import-resolution change centered on `resolveJavaImportTarget`, `interpretJavaImport`, and `resolveImportTarget`, reaching 3 graph dependents.**

The production impact is concentrated in the `Scope-resolution` and `Import-target` modules, with supporting changes in `Java` and `Import-resolvers`. Review `gitnexus/src/core/ingestion/languages/java/import-target.ts`, `gitnexus/src/core/ingestion/languages/java/interpret.ts`, and `gitnexus/src/core/ingestion/languages/java/scope-resolver.ts` first, particularly the interactions among `getJavaDirIndex`, `resolveJavaImportTarget`, `interpretJavaImport`, and `resolveImportTarget`.

The change also updates `gitnexus/src/core/ingestion/languages/java/module-resolution.ts` and `gitnexus/src/core/ingestion/language-provider.ts`, alongside Java integration and scope-resolution tests and import-package evidence fixtures. No affected execution flows were traced. The per-file risk level is LOW, with no HIGH or CRITICAL risk files identified.

_Added by GitNexus for PR #2955. Edit freely — this block is replaced on the next review, everything above it is left untouched._


azizur100389:
Follow-up 39e52736 guards the empty package-evidence cache case covered by the shared import-target contract. CI is green: 18,348 tests passed with zero failures, including all platform shards and benchmarks.

magyargergo:
Sorry @azizur100389 I'm working on a generic Module resolution solution as part of https://github.com/abhigyanpatwari/GitNexus/pull/2956, and adapting it to other langauges as well

magyargergo:
Like so https://github.com/magyargergo/GitNexus/pull/2 - if could adapt this please 🙏

azizur100389:
Adapted to the declared-package resolution approach from #2956 and your reference branch. Added pipeline regressions for external lookalikes, flattened package layouts, and wildcard fan-out; 312 focused tests plus typecheck, ESLint, and Prettier pass.

azizur100389:
Updated the benchmark harness for Java declared-package context and added the static-wildcard/package collision regression. Affected tests (327), typecheck, lint, formatting, and Java benchmark gates pass.

koriyoshi2041:
Retested the merged current head `7d4541c` against the package-evidence and index-reuse boundaries. The Java module-resolution, resolver integration, and shared import-target contract suites pass: 288/288 across 3 files. I also confirmed the head matches the PR SHA; no regression surfaced from the latest main merge.

Issue #2912, opened 2026-08-09
fix(scope-resolution): make interface dispatch generic-instantiation aware


azizur100389:
## Problem

Interface-dispatch fan-out currently erases generic type arguments when building the subtype closure. Distinct heritage edges such as `IValidator<string>` and `IValidator<int>` resolve to the same `IValidator` definition, so a call through `IValidator<string>` can fan out to implementations of incompatible instantiations.

```csharp
namespace Probe;
public interface IValidator<T> { bool Check(T item); }
public record UserValidator : IValidator<string> { public bool Check(string item) => true; }
public record IntValidator : IValidator<int> { public bool Check(int item) => true; }
public class Runner { public bool Run(IValidator<string> v) => v.Check(x); }
```

Current result includes both secondary targets:

- `UserValidator.Check(string)` — reachable
- `IntValidator.Check(int)` — not reachable from an `IValidator<string>` receiver

The same behavior already exists for classes. Record fan-out in #2904 only makes the pre-existing limitation visible for another class-like kind.

Discovered in #2904: https://github.com/abhigyanpatwari/GitNexus/pull/2904#discussion_r3743227770

## Acceptance criteria

- Preserve generic instantiation information needed by receiver-bound interface dispatch.
- Filter fan-out targets to heritage instantiations compatible with the receiver's declared generic type.
- Add regressions covering at least two incompatible instantiations with the same method name.
- Preserve non-generic subtype-closure fan-out, overload selection, fan-out caps, and skipped-target reporting.
- Document behavior for unresolved or partially known generic arguments.

Issue #1881, opened 2026-05-28
Unrelated edges are generated for C# Files


hugogu:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

Sample:

It is a Basic event definition, but linked with so many links that has no real relationship with it. 

<img width="50%" alt="Image" src="https://github.com/user-attachments/assets/02df7fe5-c01c-4a66-b986-27545f0b81b2" />

Picked one edge to check if the other side reference back to it. 

<img width="50%" height="722" alt="Image" src="https://github.com/user-attachments/assets/5888263c-734d-4a18-8963-79d511317114" />

There is another case, a Test file being referenced by all other files

<img width="50%" height="2584" alt="Image" src="https://github.com/user-attachments/assets/6c85cc6a-a6c5-4173-a7b0-46042c01be92" />

### Expected behavior

Edge should be built for real direct relationship

### Actual behavior

Some edges are generate for no reasons. 

### Steps to reproduce

1. Clone a C# project.
2. Check the node with the most edges
3. Check if all edges reflects a real relationship

magyargergo:
Thanks for reporting this bug! I just push a fix for this.

hugogu:
> Thanks for reporting this bug! I just push a fix for this.

@magyargergo Ah, Thanks for the quick fix, I was planning to pick it up myself this weekend if nobody helped. I have just verified the redudant edges are gone now. Super.

magyargergo:
It was a bit more tricky so I decided to take on it. Thanks for raising it! 🙏 if you find anything else please don't hesitate to raise it to us.

Issue #898, opened 2026-04-17
Python dotted imports can resolve Django framework modules to local files


yacinehmito:
## Summary
Python dotted imports that should remain external can be misresolved to unrelated local files.

## Reproduction
Use a Django-style repo with multiple local modules named like common framework submodules:

- `accounts/apps.py` and `billing/apps.py` both contain `from django.apps import AppConfig`
- `config/urls.py` contains `from django.urls import path`
- `config/asgi.py` contains `from django.core.asgi import get_asgi_application`
- `config/wsgi.py` contains `from django.core.wsgi import get_wsgi_application`

When GitNexus indexes the repo, it can emit incorrect `IMPORTS` edges between local files instead of treating those `django.*` imports as unresolved external dependencies.

## Actual behavior
Examples observed locally:

- `accounts/apps.py -> accounts/apps.py`
- `billing/apps.py -> accounts/apps.py` or `billing/apps.py`
- `config/urls.py -> config/urls.py`
- `config/asgi.py -> config/asgi.py`
- `config/wsgi.py -> config/wsgi.py`

## Expected behavior
- External imports like `django.apps`, `django.urls`, `django.core.asgi`, and `django.core.wsgi` should not resolve to local repo files.
- Real local imports should still resolve normally, e.g. `from accounts.models import Customer` from `billing/models.py`.

## Suspected cause
Python import resolution appears to let unresolved dotted imports fall through to generic suffix matching, so a basename like `apps.py` or `urls.py` can win even when the leading package (`django`) does not exist in the repo.

## Impact
This creates incorrect `IMPORTS` edges and can mislead downstream graph queries, impact analysis, and any resolution logic that depends on import edges.

PR #2956, opened 2026-08-13
fix(typescript): resolve imports against declared config, not path suffixes (#2953)

### LINKED ISSUE #2953: LANG-typescript: external package imports basename-misresolve to local files (false IMPORTS at confidence 1.0), and workspace-package CALLS edges are dropped (pnpm monorepo)
### Area

gitnexus (CLI / core / indexing) — LANG-typescript import resolution

### Environment

gitnexus 1.6.9, @ladybugdb/core 0.18.3, darwin-arm64, Node v24.13.0. pnpm 9 workspace monorepo (~3k TS files, `apps/*` + `packages/*`).

### Summary

In a pnpm workspace monorepo, the TypeScript import resolver mishandles **both** kinds of non-relative specifier, in opposite directions:

1. **External (registry) packages misresolve to local files.** A scoped external dep with a subpath export (`@acme/telemetry/nest`) fails to resolve, falls back to what looks like basename matching, and links to the repo's *only* file whose path ends in `nest/index.ts` — a file with no relationship to the package. The resulting `IMPORTS` edge is emitted at **`confidence: 1.0`** with `reason: "typescript-scope: import"`.

2. **Workspace packages resolve to nothing.** A first-party workspace package (`@repo/utils`, declared in `pnpm-workspace.yaml`) imported from `apps/` produces **no cross-boundary `CALLS` edges** at all.

Both point at the same gap: the resolver appears to consult neither the importing package's `package.json` dependencies nor the workspace manifest, so it cannot distinguish "external, should stay unresolved" from "workspace-internal, should resolve to a real file".

This is the same bug class as #898 (*Python dotted imports can resolve Django framework modules to local files*), which was fixed for Python. This report is the TypeScript equivalent.

### Reproduction

```
repo/
  pnpm-workspace.yaml            packages: ["apps/*", "packages/*"]
  apps/web/package.json          dependencies: { "@acme/telemetry": "0.5.0" }   # EXTERNAL, from a registry
  apps/web/src/main.ts           import { init } from "@acme/telemetry/nest";
  apps/web/src/use.ts            import { doThing } from "@repo/utils"; doThing();
  packages/inner/src/nest/index.ts    # the ONLY path in the repo ending in nest/index.ts
  packages/utils/package.json    name: "@repo/utils"                            # WORKSPACE package
  packages/utils/src/index.ts    export function doThing() {}
```

`gitnexus analyze -f`, then:

```cypher
-- (1) false external edge
MATCH (a:File)-[r:CodeRelation {type:'IMPORTS'}]->(b:File {filePath:'packages/inner/src/nest/index.ts'})
RETURN a.filePath, r.confidence, r.reason;

-- (2) missing workspace CALLS
MATCH (a)-[:CodeRelation {type:'CALLS'}]->(b)
WHERE a.filePath STARTS WITH 'apps/' AND b.filePath STARTS WITH 'packages/utils/'
RETURN count(*);
```

### Actual behaviour

1. An `IMPORTS` edge `apps/web/src/main.ts -> packages/inner/src/nest/index.ts` at `confidence: 1.0`, `reason: "typescript-scope: import"`. `apps/web/src/main.ts` never references `packages/inner` in any form.
2. Query (2) returns `0`.

Measured on the real monorepo this was found in (names withheld, private repo):

| Observation | Count |
|---|---|
| False `IMPORTS` edges onto a single `nest/index.ts`, all from files that never reference that package | **28** |
| False `IMPORTS` edges onto one other file, same mechanism | **16** |
| `apps/ -> packages/` `IMPORTS` edges landing on just those two files | **44 of 74** |
| `CALLS` from `apps/` into one workspace package that has **75** import statements from `apps/` | **0** |
| Total `apps/ -> packages/` `CALLS` edges, whole repo | **46** |
| `CALLS` *within* that same workspace package (intra-package resolution works fine) | **241** |

Confirmed on a clean `analyze -f` against current HEAD, so this is not a stale-index artifact.

### Expected behaviour

1. An unresolvable external specifier should emit **no in-repo `IMPORTS` edge** (or an edge to an explicit external/unresolved node). It should never basename-match into the repo.
2. A specifier matching a workspace package name should resolve to that package's real entry point, producing normal `IMPORTS`/`CALLS` edges.
3. Separately, and independent of 1 and 2: a heuristic basename fallback should **never** be stamped `confidence: 1.0`. Consumers cannot distinguish a guess from a real resolution, so nothing downstream can filter it.

### Impact

`impact`, `trace`, and `context` are wrong in both directions on any monorepo of this shape:

- **Over-reporting.** `impact` on `packages/inner/src/nest/index.ts` claims 28 dependents in an unrelated app. Acting on that blast radius means reviewing code that cannot be affected.
- **Under-reporting.** `impact` on a workspace-package export reports no app-side callers when 75 import statements exist. This is the more dangerous direction: it reads as "safe to change".

The `confidence: 1.0` on the false edges is what makes (a) unrecoverable downstream — there is no signal to threshold on.

### Related

- #898 — same bug class, Python/Django, fixed. Closest precedent.
- #1499 — `<namespace>.method()` resolved as a free function across packages.
- #1646 — LANG-javascript remaining scope-resolution work (these edges carry `reason: "typescript-scope: import"`).
- #266 — `IMPORTS` edges target File nodes rather than Class/Function nodes.
- #2908 / #2879 / #2877 — import resolution scanning the whole workspace per import in java/cobol, dart, go. Perf-framed, but the "scan the workspace for a candidate path" design is what produces the basename collision here.

magyargergo:
Fixes #2953.

## The defect

TypeScript/JavaScript/Vue import resolution ended in `suffixResolve`, which answers *"does any file in this repo have a path ending in this specifier?"* — and answers it by dropping leading segments until something matches. That is not module resolution, and it failed in both directions at once:

- **`@acme/telemetry/nest`**, a registry dependency with no in-repo file at all, became `nest` and landed on the repo's only path ending in `nest/index.ts`. Emitted at `confidence: 1.0`, so nothing downstream could tell it from a real edge. The reporter measured **44 of 74** `apps/ -> packages/` edges landing on two such files.
- **`@repo/utils`**, a first-party workspace package, resolved to **nothing**. Its name lives in `packages/utils/package.json` and appears in no file path, so a path matcher cannot find it — **0** CALLS from 75 import statements.

Same missing input behind both: nothing read the config that says what exists.

## The fix

The algorithm `tsc` and Node actually run, in their order:

```
1. relative / absolute   → a path is a path
2. #imports              → package-scoped, shadows everything
3. tsconfig paths        → longest literal prefix wins, every target tried
4. tsconfig baseUrl      → what makes `import 'src/utils/foo'` legal
5. workspace package     → its own exports / main / module / types
6. otherwise             → null (external)
```

No fallback. A specifier nothing declares resolves to nothing.

**New modules**

- `typescript/tsconfig.ts` — every tsconfig/jsconfig in the repo, `extends` chains resolved, nearest-config-wins per file. The old loader read three filenames at the repo *root*, gave up unless `paths` existed, and kept only `targets[0]` — none of which describes a monorepo, where `apps/web/tsconfig.json` is what governs `apps/web/src/main.ts`.
- `typescript/module-resolution.ts` — the algorithm.
- `typescript/file-candidates.ts` — 11 TS-family extensions, replacing a shared 39-entry list spanning every indexed language. A TypeScript import can no longer resolve to a `.py` file.
- `import-resolvers/node-workspace-packages.ts` — in-repo manifests: `exports` subpath maps, patterns, condition nesting, and the restriction that a package declaring `exports` exposes **only** what it lists.

## Deletions, and why

The per-pass `SuffixIndex` is gone from these three adapters. Real resolution derives nothing from the file list — every candidate comes from a declared source and is checked with one `Set.has` — so there is nothing left to cache per pass.

That makes three `*-import-index-reuse` guards and the JS index-vs-scan differential (`javascript-import-target-parity`) tests of a mechanism that no longer exists on this path; they are deleted with it. The cross-language contract test moves TS/JS/Vue into its existing `KNOWN_UNINDEXED` channel and pins the exemption **as a list**, so a fourth language arriving there is a deliberate claim rather than a silent regression.

**Scope:** Python, Ruby, Java, Go and the rest still route through `suffixResolve` and are untouched. The same bug class is reachable for them; this PR does not claim otherwise.

## Verification

| | |
|---|---|
| new fixture (pnpm workspace) | 7/7 — baseUrl resolves, `paths` alias resolves, workspace package resolves, external → null, undeclared → **no edge** |
| `node-workspace-packages` unit | 19/19 — incl. `exports` restriction and `#imports` scoping |
| scope-resolution + resolver units | 117 files, 1935 passed |
| resolver integration | 63/65 files, 3440 passed |
| `tsc --noEmit`, eslint, prettier | clean (0 errors) |
| `detect_changes --scope all` | low risk, 0 affected processes |

**Pre-existing, not from this PR:** `test/integration/resolvers/java.test.ts` fails 5 record-accessor tests (#2936/#2917). Confirmed identical — 5 failed / 251 passed — on a clean `origin/main` checkout with these changes stashed.

https://claude.ai/code/session_015Px638Zyqa9CJMUU7DsJoB

---

## Summary

*A broad, high risk change to language ingestion and import resolution, concentrated in TypeScript workspace configuration and its parity tests. The impact is distributed across resolver code, fixtures, and benchmark baselines rather than traced execution flows.*

**🟠 HIGH blast radius. A TypeScript import-resolution change in the language-ingestion core, with impact reaching 13 graph dependents.**

The main implementation surface is `gitnexus/src/core/ingestion/languages/typescript/import-target.ts`, including `TsResolveContext` and `resolveTsImportTarget`, alongside `gitnexus/src/core/ingestion/languages/typescript/scope-resolver.ts`, `gitnexus/src/core/ingestion/languages/typescript/module-resolution.ts`, `gitnexus/src/core/ingestion/languages/typescript/file-candidates.ts`, and `gitnexus/src/core/ingestion/languages/typescript/tsconfig.ts`. Related JavaScript and Vue resolver paths are also changed, including `javascriptScopeResolver`, `typescriptScopeResolver`, and `tsPassCacheFor`.

The graph shows the strongest concentration in `Import-target` with 15 hits, followed by `Integration` with 3, `Scope-resolution` with 2, and `Typescript` with 1. Review `gitnexus/test/unit/scope-resolution/javascript-import-target-parity.test.ts` first, since it contains 24 changed symbols, then check the workspace fixture files and benchmark baselines for consistency with the declared configuration behavior. No affected execution flows were traced, and there are no HIGH or CRITICAL risk files.

_Added by GitNexus for PR #2956. Edit freely — this block is replaced on the next review, everything above it is left untouched._


magyargergo:
Pushed a second commit: a conformance suite asserting the property this PR is about, for **all 16** registered resolvers.

> A specifier naming something outside the repository must not resolve to a file inside it.

Every case pairs an external specifier with a **decoy** — an unrelated in-repo file whose path ends the way the specifier does. Without one, a resolver that merely found nothing would pass while holding no property at all. Each case also asserts the decoy is reachable by the spelling that *should* find it, so a typo in a fixture cannot manufacture a pass.

Two fixtures had to be corrected before the results meant anything, and both would have recorded a **false gap**:

- **C#** reads its #1881 gate from scanned namespace evidence and fails *open* without any — passing `undefined` measured nothing. Armed, C# holds.
- **C++** was passing on an extension mismatch (`vector` could never match `src/vector.hpp` whatever the resolver did). Given the header spelling, it does not hold.

## Six hold it

| language | why |
|---|---|
| TypeScript, JavaScript, Vue | resolve against declared config only — no suffix fallback (this PR) |
| Python | #898 gate: leading segment must name something in-repo |
| C# | #1881 gate: same shape, keyed on declared namespaces |
| Rust | `::` never decomposes into a path suffix, so `serde::de` cannot reach `src/de.rs` — confirmed a real pass, not a vacuous one, by the decoy-reachability arm |

Worth noting the three mechanisms differ, and only one generalizes: Python and C# **kept** the guess and filtered it, which answers *"probably not external"* rather than *"here is what the language says this means"*.

## Ten do not

Recorded in `KNOWN_GAPS` with what each currently answers:

```
java     java.util.List                        -> vendor/util/List.java
kotlin   org.junit.Assert                      -> src/main/kotlin/vendor/Assert.kt
go       github.com/vendor/dep/internal/models -> internal/models/user.go
ruby     rails/generators                      -> lib/generators.rb
php      Vendor\Ghost\Missing                  -> lib/Legacy/Missing.php
dart     package:http/http.dart                -> lib/http.dart
swift    Foundation                            -> Sources/Foundation/Thing.swift
c        stdio.h                               -> src/stdio.h
cpp      cstdio.h                              -> src/cstdio.h
cobol    EXTERNAL                              -> vendor/EXTERNAL.cpy
```

The map is a work list, not an allowance: entries are **asserted**, so a language that starts holding the property *fails* here and its line gets deleted deliberately rather than rotting into a lie. Each follow-up is one language getting its real algorithm, then one deleted line.

33/33 passing.

koriyoshi2041 (commented):
I reproduced the focused resolver unit coverage at `1f649bc` (19/19), but there is one workspace-boundary issue that looks worth fixing before merge: `loadNodeWorkspacePackages` treats every `package.json` found by the repo-wide BFS as a workspace package and never reads `pnpm-workspace.yaml` or a root `workspaces` declaration.

That can recreate the false-positive half of #2953 with a different source of evidence. For example, this repository's own scan would register package names from `gitnexus/test/fixtures/**` and `gitnexus/vendor/**` (`@repo/utils`, `tree-sitter-c`, etc.) as production workspace packages. More generally, if an app depends on registry package `foo` while an excluded example/fixture also has `name: "foo"`, `resolveNodeWorkspaceImport("foo", ...)` will bind the app import to the excluded local manifest even though the workspace config never admitted it.

Could the loader first derive the admitted package directories from the root workspace declarations (including pnpm patterns), then read manifests only inside that set? A negative fixture with a named package outside the declared `packages:` globs would lock the boundary.

CI also has two actionable reds at this head: `scope-capture --check` reports the TypeScript fingerprint drift and `import-target --check` fails in the benchmark job. The Windows shard is a separate `cli-e2e` analyze timeout and may be runner noise, but the benchmark guards appear directly related to the resolver rewrite.

magyargergo:
@koriyoshi2041 thanks — the boundary issue was real and is fixed, along with the four bot findings that held up. Also merged `main` (the PR was BEHIND) and both benchmark guards are green.

## The workspace boundary

You were right, and it is the same defect class this PR exists to fix arriving from a new direction: finding a `package.json` is not the workspace admitting one.

The admitted set now comes from the declaration — `workspaces` (array **and** the yarn object form), `pnpm-workspace.yaml`, `lerna.json`, with `!` exclusions and `*`/`**` — plus the root package itself. A repo declaring no workspace has exactly one package: the root. That is this repository's own shape, so `test/fixtures/**` and `vendor/**` no longer register.

The negative fixture you asked for is in: `examples/excluded/package.json` declares `@repo/excluded` outside the `packages:` globs, and `apps/web/src/outside.ts` imports it. It must produce no edge. Eight unit arms cover the declaration sources, `!` exclusion, `**`, and the no-declaration case.

## Bot findings

Four held up and are fixed:

- **`paths` targets ignored an inherited `baseUrl`** (P1) — correct, and the nastiest of the set: an extending config loaded the right alias pattern and pointed every target at the wrong directory. Writing the test for it surfaced **a second bug I had missed**: two configs in one directory were ranked by directory-listing order, so `tsconfig.base.json` could govern instead of `tsconfig.json` and a config's own `paths` went invisible.
- **unexported subpath fell back into `src/`** (P1) — correct by this PR's own standard. Nothing declares that mapping, and the import it "resolved" is broken in the real project too. Removed.
- **`imports` pattern keys never matched** (P1) — correct. `exports` and `imports` now share one matcher.
- **relative traversal past the repo root was clamped** (P2) — correct. `../../../secret` from `src/main.ts` became `secret`. Now `null`.

Two test-rigor findings were also right, and are the kind worth calling out:

- the decoy-reachability arm only asserted non-empty, so five cases paired `reachesDecoy` with a *different* file than `decoy` and passed while establishing nothing;
- `KNOWN_GAPS` accepted any in-repo answer instead of the recorded one.

Both now assert the exact file. The reachability arm runs only for languages that **hold** the property — for a gap language the recorded-answer assertion *is* that proof, and for Swift and COBOL no other spelling exists, since `Foundation` and `EXTERNAL` name the in-repo directory and copybook as well as the external module. That is precisely why those two resolvers cannot tell them apart.

**One I disagree with:** the P1 saying `#`-prefixed imports are collected but never resolved. `nodePackageNameOf` does return `null` for them, but `module-resolution.ts` handles `#` in step 2, before the package branch is reached — the finding read one file in isolation. It is unit-tested both ways (resolves within the owning package, does not honour another package's map). The *pattern* half of that finding was real and is fixed above.

## Benchmarks — both were reporting something true

**`import-target`**: the ts-family arms resolved **0 of 3200** imports. Their corpus is bare specifiers with no config, which the deleted `suffixResolve` answered without one — so the arms measured an empty branch while printing a clean scaling ratio. Each now carries the config its corpus is spelled for, and the `deep` arm's uniform prefix reaches it.

> **The fingerprints then matched the recorded baselines exactly.** Same corpus, same targets — once the config it always implied is passed explicitly. That is the strongest evidence in this PR that the rewrite is behaviour-preserving where behaviour was defined.

Two baseline moves, both recorded with their reasoning:

| | before | after |
|---|--:|--:|
| retained per-pass index @32k files (js/ts) | 26 745 296 B | **0–16 B** |
| retained per-pass index @32k files (vue) | 28 884 016 B | **0–16 B** |
| depth ratio (js/ts) | ~2.0 | ~2.2 (budget → 2.6) |

The heap drop is real, not an arm that stopped measuring — these resolvers no longer build an index at all, so they move to the `HEAP_BOUNDED` tier rust already occupies for exactly this reason (bound, no floor: a floor over a reading at its own noise gates the noise). The depth move is candidates carrying the 16-segment baseUrl prefix, so each `Set.has` hashes a longer string — linear in path *length*, independent of file *count*.

**`scope-capture`**: TypeScript capture fingerprint drift, from this PR's 12 new `.ts` fixtures entering the corpus. Attribution is exact rather than inferred — moving that one fixture directory aside returns the fingerprint to `f719163e…` byte-for-byte with `fixture_count` back at 155 and all 15 languages passing.

## Verification on the merged tree

| | |
|---|---|
| scope-resolution + resolver units | 123 files, **2037 passed** |
| resolver integration | 62/66 files, **3451 passed** |
| `import-target --check` / `scope-capture --check` | **PASS** |
| `tsc`, eslint, prettier | clean |
| `detect_changes --scope compare --base-ref origin/main` | low risk, 0 affected processes |

The 16 integration failures are all pre-existing on `main`, verified by running them on a clean `origin/main` checkout: `java.test.ts` (5, record-accessor/enum-heritage #2936/#2917) and `generic-field-receiver-matrix` + `generic-interface-dispatch` (11, from #2912 which landed in the merge — **identical count on clean main**).

The Windows shard red is the `cli-e2e` analyze timeout you flagged as separate; nothing in this change touches that path.

magyargergo:
Pulled every comment on this PR — 6 issue comments, 6 reviews, 13 inline — and triaged them against what this tool actually is: **a static analyser building a code graph, not a compiler, transpiler or V8.** That framing changes several verdicts, so it is stated explicitly below.

## The bar for "reliable module resolution" here

A compiler resolves in order to *emit*, against an installed, buildable project, and may refuse to proceed. GitNexus resolves in order to *describe*, and:

1. must work on a checkout that was **never installed and never built** — `node_modules/` and `dist/` are not indexed;
2. must never fabricate an edge, because a wrong edge is indistinguishable from a real one downstream (this is #2953 itself);
3. may resolve **less** than tsc without being wrong — a missing edge is a gap, an invented one is a defect;
4. needs nothing tsc does for typing: conditional `exports` per runtime, `lib`/`target`, declaration emit.

So findings get judged on "does the project declare this, and can we read it without a build" — not on "would tsc do it".

## Round 2 — four new findings, all valid, all fixed

| finding | verdict | why it matters *here* |
|---|---|---|
| `exports`/`imports` **arrays** skipped | valid | `["./dist/x.js", "./src/x.ts"]` is how a workspace package says "built output, or source". `dist/` is not indexed and the build may never have run, so **the source arm is the only one we can use** — and skipping arrays dropped the declaration entirely, making the package look as though it exported no subpaths |
| exact `paths` pattern didn't outrank a wildcard | valid | `a` and `a*` both match `a` at prefix length 1, so tsc's exact-wins rule was left to declaration order |
| package-form `extends` refused | valid | not *indexing* `node_modules` ≠ not *reading* it; a shared internal base is where a monorepo puts its `paths` |
| CodeQL: `replace('*', …)` ×2 | valid | ⬅️ this was the failing CodeQL check |

**On the CodeQL one** — the behaviour was already correct: Node subpath patterns and tsconfig `paths` both allow **at most one** `*`, so replacing the first occurrence *is* the spec. But `String.replace` with a string needle says that only by accident and reads as the replace-all footgun. `substituteStar` now slices at the known index and states the rule.

**On package-form `extends`** — worth reporting because the test contradicted my first assumption. It is now read from disk, walking `node_modules` upward the way Node does, and absent on an un-installed checkout it degrades to whatever the extending config declared itself. But the test pins what tsc *actually* does with such a base rather than what I expected:

> `extends` never rebases `baseUrl`. A package base declaring `"baseUrl": "src"` means `node_modules/@acme/tsconfig/src`, not the consuming repo's `src`.

So a published base rarely contributes aliases a repo's own files resolve through — and the `@tsconfig/*` family, which sets `target` and `lib` and never `paths`, is a no-op here either way. Reading it is still correct; the expectation that it would unlock aliases was not.

## Round 1 — status of the earlier eight

Seven were valid and are fixed (inherited `baseUrl`, the `src/` fallback, `#imports` patterns, root-escaping traversal, the two conformance-rigor findings, the stale `KNOWN_UNINDEXED` comment). Details in the previous comment.

**One remains invalid**: the P1 saying `#`-prefixed imports are collected but never resolved. `nodePackageNameOf` does return `null` for them, but `module-resolution.ts` handles `#` in step 2, *before* the package branch — the finding read `node-workspace-packages.ts` in isolation. Unit-tested both ways (resolves within the owning package; does not honour another package's map). Its *pattern* half was real and is fixed.

## Deliberately not doing

- **conditional `exports` resolution per runtime** (`node` vs `browser` vs `import`/`require`). We collect the string leaves of every condition and try them in order. A compiler must pick one branch; an analyser wants the union, because the graph should show the file regardless of which runtime the consumer picked.
- **`include`/`exclude` glob evaluation** to decide project membership. Nearest-enclosing-tsconfig is the approximation, and it is honest about being one.
- **type-only resolution** (`typesVersions`, `@types/*`). Irrelevant to a file-level import graph.

Each of these makes us resolve *less* than tsc, never differently.

## Verification

| | |
|---|---|
| scope-resolution + resolver units | 123 files, **2042 passed** |
| resolver integration | 62/66 files, **3451 passed** |
| `import-target --check` / `scope-capture --check` | **PASS** |
| `tsc`, eslint, prettier | clean |

The 16 integration failures are the same pre-existing ones, verified on a clean `origin/main`: `java.test.ts` (5, #2936/#2917) and `generic-field-receiver-matrix` + `generic-interface-dispatch` (11, from #2912 — identical count on main).

magyargergo:
Pulled all 12 gitnexus-check findings across the three rounds and evaluated each. **11 valid, 1 invalid.** All 11 are now fixed.

## Round 3 — two new findings on `32e83d1`, both valid

Both are the same shape, and worth naming: **this resolver was laxer than the thing it models**, which is the direction that fabricates edges.

**1. `exports` map that omits the root still answered a bare import.** When a manifest declares `exports`, that map is the package's *entire* public interface — Node ignores `main` outright and refuses any subpath the map doesn't list. I already honoured that restriction for **subpaths** and not for the **package root**, which is the same rule. So `{"exports": {"./feature": "./src/feature.ts"}}` still resolved a bare `@repo/pkg` to `main` or `src/index`, an edge for an import that doesn't resolve in the real project. Legacy and conventional root candidates are now offered only when there is no `exports` field at all.

That one was an inconsistency inside my own code, not a missing feature — the bot caught an asymmetry I'd written two screens apart.

**2. An empty tsconfig was dropped, so an ancestor's `baseUrl` governed its files.** A config declaring neither `baseUrl` nor `paths` was skipped rather than kept as an empty scope, so `tsconfigFor` fell through to an enclosing config. A package whose own tsconfig declares no `baseUrl` — meaning its non-relative specifiers are *package lookups* — silently inherited the repo root's aliases. An empty scope is the accurate answer for such a file, and only a scope can express it.

Both are pinned at the level they broke: the manifest arms assert what `readManifest` **produces**, not a hand-built package, because "the resolver honours empty entries" and "the loader produces empty entries" are different claims and only the second was at fault.

## Full tally across all three rounds

| # | finding | verdict |
|---|---|---|
| 1 | `#imports` collected but never resolved | ❌ **invalid** |
| 2 | `#imports` wildcard keys never match | ✅ fixed |
| 3 | unexported subpath falls into `src/` | ✅ fixed |
| 4 | relative traversal past repo root clamped | ✅ fixed |
| 5 | stale `KNOWN_UNINDEXED` comment | ✅ fixed |
| 6 | decoy reachability never verified | ✅ fixed |
| 7 | `KNOWN_GAPS` accepts any in-repo answer | ✅ fixed |
| 8 | `paths` targets ignore inherited `baseUrl` | ✅ fixed |
| 9 | `exports` fallback arrays ignored | ✅ fixed |
| 10 | exact `paths` doesn't outrank wildcard | ✅ fixed |
| 11 | package-form `extends` ignored | ✅ fixed |
| 12 | `exports` root omission / empty scopes | ✅ fixed |

**The one invalid finding (#1)** claims `#`-prefixed imports are collected but never resolved. `nodePackageNameOf` does return `null` for them, but `module-resolution.ts` handles `#` in step 2, *before* the package branch is reached — the finding read `node-workspace-packages.ts` in isolation and didn't follow the call. Unit-tested both directions: resolves within the owning package, and does **not** honour another package's map. Its *pattern* half (#2) was real and is fixed.

Two findings also caught bugs beyond what they described:
- writing the test for #8 surfaced a second bug — two configs in one directory were ranked by directory-listing order, so `tsconfig.base.json` could govern instead of `tsconfig.json`;
- #11's test disproved my own assumption: `extends` never rebases `baseUrl`, so a package base's paths point at the package's own directory. Reading it is still correct; expecting it to unlock repo aliases was not.

## Verification

| | |
|---|---|
| scope-resolution + resolver units | 123 files, **2046 passed** |
| resolver integration | 62/66 files, **3451 passed** |
| `import-target --check` / `scope-capture --check` | **PASS** |
| `tsc`, eslint, prettier | clean |

16 integration failures unchanged and all pre-existing on clean `origin/main`: `java.test.ts` (5, #2936/#2917) and the two #2912 generic files (11 — identical count on main).

PR #2979, opened 2026-08-15
fix: add heal script to clear stale LadybugDB WAL files that brick graph loads

BarnsL:
## Summary

- Adds `gitnexus/scripts/heal-stale-wal.mjs`, a standalone Node.js script that clears stale LadybugDB WAL files preventing graph loads
- Aborting a graph stream mid-flight (navigating away, reloading, closing the tab) leaves a small `lbug.wal` behind. Every subsequent database open then fails on a missing `lbug.shadow`, and the web UI permanently shows "Graph not loaded"
- The script reads `~/.gitnexus/registry.json`, identifies WAL files at or below the header-only threshold (4 KB), backs each up to `~/.gitnexus/wal-backups/`, and unlinks them

## Reproduction

1. Open the GitNexus web UI on a large index
2. Click "Load graph anyway" on the graph visualization
3. Before the stream finishes, navigate away or reload the page
4. Try loading the graph again — it will show "Graph not loaded" with a `lbug.shadow` error in the stream

## Fix

```bash
node gitnexus/scripts/heal-stale-wal.mjs          # heal all registered indexes
node gitnexus/scripts/heal-stale-wal.mjs --check   # dry run, report only
```

## Safety

- Refuses to run while `gitnexus analyze` is in flight (checks process list)
- Only targets WAL files ≤ 4096 bytes (header-only artifacts from aborted reads)
- Backs up each WAL before deleting
- Reports status per index so the user can see exactly what happened

---

## Summary

*A small, self-contained recovery utility with a narrow blast radius. The change is concentrated in stale WAL cleanup rather than an existing execution flow.*

**🟢 LOW blast radius. This appears to be a standalone recovery script in `gitnexus/scripts/heal-stale-wal.mjs` for clearing stale LadybugDB WAL files that can brick graph loads.**

The change adds one script and no changed symbols, with no graph-traced dependents or affected flows. Review `gitnexus/scripts/heal-stale-wal.mjs` first, focusing on the stale WAL cleanup behavior and how it handles graph-load recovery.

There are no cross-repo consumers and no HIGH or CRITICAL risk files.

_Added by GitNexus for PR #2979. Edit freely — this block is replaced on the next review, everything above it is left untouched._

### DIFF
--- a/gitnexus/scripts/heal-stale-wal.mjs
+++ b/gitnexus/scripts/heal-stale-wal.mjs
@@ -30,6 +30,11 @@
  * script refuses to run while an analyze is in flight. It also copies each WAL to
  * a backup directory before unlinking, and only targets WAL files at or below a
  * header-only size threshold (4 KB) to avoid touching legitimate transactions.
+ *
+ * The process scan is a courtesy pre-check, not a lock. An analyze could start
+ * between the check and the unlink. On Windows the unlink will fail with EBUSY
+ * if the WAL is held open, making the race benign. On POSIX the fd stays valid
+ * after unlink, so the writer is also unaffected.
  */
 
 import { execSync } from 'node:child_process';
@@ -37,8 +42,9 @@ import { existsSync, mkdirSync, copyFileSync, unlinkSync, readFileSync, statSync
 import { join, resolve } from 'node:path';
 import { homedir } from 'node:os';
 
-const REGISTRY_PATH = join(homedir(), '.gitnexus', 'registry.json');
-const BACKUP_DIR = join(homedir(), '.gitnexus', 'wal-backups');
+const GITNEXUS_DIR = process.env.GITNEXUS_HOME || join(homedir(), '.gitnexus');
+const REGISTRY_PATH = join(GITNEXUS_DIR, 'registry.json');
+const BACKUP_DIR = join(GITNEXUS_DIR, 'wal-backups');
 const HEADER_ONLY_BYTES = 4096;
 
 function isAnalyzeRunning() {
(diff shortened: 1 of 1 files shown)


koriyoshi2041 (changes requested):
The current head still lets a registry alias escape `wal-backups`, so I don't think the destructive path is safe yet.

I reproduced this on exact head `3c7b9f97ee2f1339d9b9c9a42731a04511ce0424` with an isolated `GITNEXUS_HOME` and one registry entry whose name was `../escaped`. Running the script against a 3-byte `lbug.wal` reported success, but created:

```text
<GITNEXUS_HOME>/escaped-20260816-094743.wal
```

instead of a file under:

```text
<GITNEXUS_HOME>/wal-backups/
```

The added index/PID suffix only prevents filename collisions; `join(BACKUP_DIR, `${name}${suffix}-${ts}.wal`)` still resolves `../` before the filename is created. Because `analyze --name` can persist a custom alias, this is reachable from normal registry state rather than requiring manual registry corruption.

Please derive the backup filename from a sanitized basename/opaque stable ID and then assert the resolved destination remains inside `BACKUP_DIR`. A regression using `../escaped` (plus `/` and `\\` separators for cross-platform behavior) would make that containment contract explicit.

azizur100389:
Maintainer review at `3c7b9f97ee2f1339d9b9c9a42731a04511ce0424`.

This destructive recovery script is not safe to merge:

1. It trusts registry `storagePath` for `unlinkSync` without `assertSafeStoragePath`, so a user-writable/corrupt registry can target an unrelated small `lbug.wal`.
2. It classifies solely by WAL size and ignores `lbug.shadow`. An isolated exact-head probe with a 3-byte WAL plus shadow deleted the WAL and left the shadow; the existing shared recovery code explicitly refuses that state as a data-loss risk.
3. The process-list precheck is not synchronization. On Windows, `tasklist /FO CSV /NH` does not include Node command-line arguments, so it cannot identify `gitnexus analyze`; no real index lock is acquired.
4. A `../escaped` registry alias escapes `wal-backups`. An isolated probe wrote the backup directly under `GITNEXUS_HOME`, reproducing the active change request.
5. There are no subprocess/cross-platform safety tests, user-facing recovery documentation, CLI integration, or failure exit statuses.

Please integrate with the existing sidecar inspection/quarantine and index-lock primitives, validate canonical registry paths, use contained opaque backup names, and add Windows/POSIX destructive-path tests.

**Verdict: not production-ready.**

PR #2984, opened 2026-08-17
fix(go): gate imports by go.mod module path

### LINKED ISSUE #2961: fix(go): third-party import paths resolve to in-repo directories
Child of #2959. Go has the cleanest authoritative signal of any language on that list, and it is already loaded.

## The defect

```
import "github.com/vendor/dep/internal/models"   →  internal/models/user.go
```

Reproduced by the `go` arm of `test/unit/scope-resolution/external-import-conformance.test.ts` (recorded in `KNOWN_GAPS`). Workspace:

```
internal/models/user.go
main.go                      ← the importer
go.mod  →  module example.com/mod
```

The import names a package in a **different module**. `resolveGoImportTarget` matches it against the in-repo directory `internal/models` anyway, because the tail happens to line up.

## The rule this should be

`go.mod`'s `module` line is definitive. An import path is in-repo **iff** it equals the module path or begins with `module + "/"`; the remainder is a directory relative to the module root. Everything else — the standard library (no dots in the first segment) and every third-party path — is external and resolves to nothing.

`GoModuleConfig` already exists (`language-config.ts:loadGoModulePath`), is already loaded per pass, and is already threaded to the resolver as `resolutionConfig`. This is mostly a matter of using it as the gate rather than as a hint.

Worth handling deliberately:

- **`/vN` suffixes** — `example.com/mod/v2/pkg` maps to `pkg/`, not `v2/pkg/`. This exact trap broke a first attempt in #2921; see that PR before writing the prefix check.
- **nested modules** — a `go.mod` in a subdirectory makes that subtree a different module, so an import of it from the parent is external even though the files are in the repository.
- **`replace` directives** pointing at a local path make a foreign module path in-repo. Reading them is optional; ignoring them costs edges, never invents them.
- **vendor/** — if present, third-party sources are in the tree; they are not the importing module and resolving into them is a separate decision.

## Acceptance

- `github.com/vendor/dep/internal/models` resolves to `null` in a repo whose module is `example.com/mod`.
- `example.com/mod/internal/models` still resolves — the paired positive.
- `KNOWN_GAPS` loses its `go` entry and that arm goes green with no other edit to the suite.

azizur100389:
## Summary

- make `go.mod` authoritative for Go import ownership
- reject standard-library and third-party imports before package-index construction
- preserve module-root, subpackage, semantic-version, and GOPATH-mode resolution

## Validation

- 272 focused and Go resolver tests passed
- Go import-target benchmark checks passed
- TypeScript, Prettier, and ESLint passed

Fixes #2961

---

## Summary

*A small, low-reach Go import-resolution change, accompanied by benchmark and scope-resolution coverage. The impact is concentrated in the Go ingestion path rather than traced execution flows.*

**🟢 LOW blast radius. A Go import resolution change centered on `resolveGoImportTarget` in `gitnexus/src/core/ingestion/languages/go/import-target.ts`, with a single direct dependent.**

The functional change is concentrated in the `Go` module. The surrounding edits cover `LANG_REGISTRY` in `gitnexus/bench/import-target/measure.mjs` and import-target behavior in `gitnexus/test/integration/go-import-index-reuse.test.ts`, `gitnexus/test/unit/scope-resolution/external-import-conformance.test.ts`, and `gitnexus/test/unit/scope-resolution/go/go-imports.test.ts`.

Review `resolveGoImportTarget` first, then compare its behavior with `legacyResolveGoImportTarget` and `legacyResolveDartRelative` in `gitnexus/test/unit/scope-resolution/import-target-index-parity.test.ts`. The graph identifies no affected execution flows, and there are no HIGH or CRITICAL risk files or cross-repository consumers.

_Added by GitNexus for PR #2984. Edit freely — this block is replaced on the next review, everything above it is left untouched._

### DIFF
--- a/gitnexus/bench/import-target/baselines.json
+++ b/gitnexus/bench/import-target/baselines.json
@@ -164,7 +164,7 @@
         "files_small": 8000,
         "files_large": 32000,
         "path_segments": 14,
-        "probe": "github.com/org/repo0/pkg/util"
+        "probe": "example.com/mod/repo0/pkg/util"
       },
       "_measured": {
         "collide_ms": 7.15,
--- a/gitnexus/bench/import-target/measure.mjs
+++ b/gitnexus/bench/import-target/measure.mjs
@@ -2016,8 +2016,8 @@ const HEAP_PROBE_TARGET = {
   // budgeted ones above: a spelling `uniqueTarget` already mints for that language, and
   // one that MISSES, so the reading is the index and the cascade runs to the
   // end. Chosen from the miss family that reaches furthest into each cascade:
-  //   - `go` takes the GOPATH fallback, one `filesDirectlyInPkgDir` per path
-  //     segment, which is the leg that forces `PackageDirIndex`;
+  //   - `go` names a missing package inside GO_MODULE, which reaches the
+  //     package-directory lookup and forces `PackageDirIndex`;
   //   - `dart` is an external package, so BOTH candidate paths miss and both
   //     walk the basename bucket to completion;
   //   - `kotlin` misses in `suffixByStem`, the map its four-tier cascade builds;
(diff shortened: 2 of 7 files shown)


koriyoshi2041 (approved):
Verified at `a0b65f30`. The go.mod gate now rejects stdlib and foreign-module imports before package-index construction while preserving own-module root/subpackage resolution, `/vN` module paths, and the no-go.mod fallback. I also checked the paired negative/positive coverage and the index-reuse assertions.

Local validation: 75 focused resolver/conformance/index tests passed, Prettier passed on the touched source/tests, and `npx tsc --noEmit` passed.

magyargergo:
@azizur100389 could you please make sure the ci is healthy?

azizur100389:
CI is healthy now. The benchmark and Windows platform shard passed on rerun, and the required CI Gate is green.

Issue #2901, opened 2026-08-08
perf(php): resolvePhpImportTarget re-materializes the workspace and passes no index, per import


magyargergo:
Surfaced by the review of #2898, which fixed this shape for Go (#2877), C# (#2878), Dart (#2879) and Ruby (#2880). PHP is now **the only language left without a per-pass memo**.

`gitnexus/src/core/ingestion/languages/php/import-target.ts`, both `resolvePhpImportTarget` (~`:185`) and `resolvePhpImportTargetInternal` (~`:220`):

```ts
const normalizedFileList = [...allFiles].map((f) => f.replace(/\\/g, '/'));
const allFileList = [...allFiles];
```

Two full array materializations per import, then `resolvePhpImportInternal` is called with **no** `index` argument — so `suffixResolve` (`import-resolvers/utils.ts:167`) falls through to its linear `findIndex` scan, once per extension per path-part. Cost is Θ(imports × files) with a large constant.

**Measured: 95.31 ms per import at 20,000 files.** For comparison, the four languages fixed in #2898 now sit at 0.7–19 ms per *3,200* imports on their bench corpus.

## Fix

`import-resolvers/workspace-file-index.ts` already provides exactly what this needs — `getWorkspaceFileIndex(allFilePaths)` returns memoized `normalized` / `all` / `index`, keyed on the Set identity via a `WeakMap`. Ruby's fix in #2898 was a three-line swap to it.

The one contract to respect: **the Set must be passed through, never copied.** A defensive `new Set(allFilePaths)` in the adapter hands a fresh `WeakMap` key per call and silently restores the quadratic behaviour with every test still green — that is #1918 review P1, and #2898 found the parity-test guard did not actually cover the adapter layer. Mirror `test/integration/<lang>-import-index-reuse.test.ts` for PHP so the guard sits where the hazard is.

## Acceptance

- `resolvePhpImportTarget` resolves through a per-run memoized index; the file set is traversed once per run, not once per import.
- Output is byte-identical to today's for the existing corpora (this is a hoist, not a behaviour change) — PHP has its own tie-breaks, so diff against verbatim copies of the current implementations the way `test/unit/scope-resolution/import-target-index-parity.test.ts` does.
- An adapter-level `test/integration/php-import-index-reuse.test.ts`, verified to fail when `new Set(allFilePaths)` is inserted at the adapter.
[...]

PR #2987, opened 2026-08-18
fix(php): gate imports by Composer autoload map

### LINKED ISSUE #2962: fix(php): vendor namespaces resolve to in-repo files outside the PSR-4 map
Child of #2959. Like Go, PHP already has the authoritative declaration loaded.

## The defect

```
use Vendor\Ghost\Missing;   →  lib/Legacy/Missing.php
```

Reproduced by the `php` arm of `test/unit/scope-resolution/external-import-conformance.test.ts` (recorded in `KNOWN_GAPS`). Workspace:

```
app/Models/User.php
lib/Legacy/Missing.php
app/Main.php                 ← the importer
composer.json  →  psr-4: { "App\\": "app" }
```

`Vendor\Ghost` matches no PSR-4 prefix, so it belongs to `vendor/` — but the resolver falls through to suffix matching and binds it to an unrelated `Missing.php`.

## The rule this should be

`composer.json`'s `autoload` map is definitive for first-party code. A fully-qualified name is in-repo **iff** some PSR-4 (or PSR-0, or `classmap`) prefix covers it; the remainder maps to a path under that prefix's directory. A namespace matching no prefix is a Composer dependency and resolves to nothing.

`ComposerConfig` already exists (`language-config.ts:loadComposerConfig`, with `psr4Sorted` precomputed longest-first) and is already threaded to `resolvePhpImportTargetInternal`. The gap is that a miss falls through instead of stopping.

Worth handling deliberately:

- **`autoload-dev`** — test namespaces are first-party too and live in a separate map.
- **`classmap` / `files`** — directory-based rather than prefix-based; a repo using only these has no PSR-4 prefixes at all, and refusing everything would be wrong. Fail open when the manifest declares no usable map, the way C#'s #1881 gate fails open on truncated evidence.
- **no composer.json at all** — same fail-open decision.
- **function and constant imports** (`use function …`) take a different leg of the resolver; check both.

## Acceptance

- `Vendor\Ghost\Missing` resolves to `null` against the fixture above.
- `App\Models\User` still resolves to `app/Models/User.php` — the paired positive.
- `KNOWN_GAPS` loses its `php` entry and that arm goes green with no other edit to the suite.

azizur100389:
## Summary

- stop unmatched PHP namespaces when Composer provides an authoritative PSR-4 map
- include `autoload-dev` mappings while preserving conservative fallback for absent or unmodelled autoload evidence
- cover class, function, and constant imports with paired positive and negative regressions

## Validation

- 358 affected tests passed
- TypeScript typecheck
- ESLint and Prettier
- GitNexus `detect_changes`: LOW risk, 6 files, 16 symbols, 0 affected processes

Fixes #2962

---

## Summary

*A critical PHP ingestion change with a broad graph reach, concentrated in import resolution, language configuration, and import-target validation. The changed-file risk is LOW, with no affected execution flows or cross-repo consumers reported.*

**🔴 CRITICAL blast radius. A PHP ingestion change that gates imports by the Composer autoload map, centered in `gitnexus/src/core/ingestion/import-resolvers/php.ts` and reaching 16 graph dependents.**

The core review path is `resolvePhpImportInternal` in `gitnexus/src/core/ingestion/import-resolvers/php.ts`, together with the language configuration changes in `gitnexus/src/core/ingestion/language-config.ts` and the PHP import target in `gitnexus/src/core/ingestion/languages/php/import-target.ts`. The graph impact is concentrated in `Import-target`, `Ingestion`, `Php`, `Import-resolvers`, and `Scope-resolution`.

Review the resolver and configuration changes first, then verify their coverage in `gitnexus/test/integration/php-import-index-reuse.test.ts`, `gitnexus/test/unit/scope-resolution/external-import-conformance.test.ts`, `gitnexus/test/unit/scope-resolution/php-import-target-parity.test.ts`, and `gitnexus/test/unit/scope-resolution/php/php-import-target.test.ts`. The benchmark surface in `gitnexus/bench/import-target/measure.mjs`, including `buildParsedFiles`, `resolvePhpImportInternal`, and related measurement helpers, is the hottest changed area.

_Added by GitNexus for PR #2987. Edit freely — this block is replaced on the next review, everything above it is left untouched._


magyargergo:
We need to benchmark this solution.

azizur100389:
Addressed the review findings: leading namespace separators now use the normalized ownership path, and empty-prefix PSR-4 mappings resolve inside their configured root without escaping to suffix fallback. Added a dedicated Composer benchmark arm; the gate passes (1.075 scaling, 1.043 collision scaling, 1.550 depth ratio, 12.79 MiB retained heap). Validation: 468 affected tests, typecheck, lint, formatting, and benchmark gate.

azizur100389:
Followed up on the fresh bot review: made empty-prefix catch-all matching explicit and changed the function/constant regression so it genuinely exercises declaration fallback. The suggested matched-prefix rewrite is intentionally outside #2962's unmatched-namespace scope; applying it breaks seven pinned PHP compatibility cases, so that behavior remains documented and unchanged. Validation: 468 affected tests, typecheck, lint, formatting, and the import-target benchmark gate pass.

gitnexus/bench/import-target/baselines.json
```diff
@@ -16,6 +17,7 @@
     "ruby": 1.8,
     "kotlin": 1.65,
     "php": 1.8,
+    "php_composer": 1.8,
```
magyargergo: We shouldn't distinguish between php and this. All php project needs composer. They should act as one.

magyargergo (changes requested):

gitnexus/bench/import-target/baselines.json
azizur100389: Consolidated as requested: PHP now has one canonical Composer-backed benchmark arm, with the duplicate php_composer arm and baselines removed. The benchmark gate passes, along with 351 PHP tests, typecheck, lint, formatting, and the low-risk GitNexus change audit.

gitnexus/bench/import-target/baselines.json
```diff
@@ -107,7 +108,7 @@
     "csharp": 29869080,
     "csharp_csproj": 73703384,
     "ruby": 41020808,
-    "php": 49574008,
+    "php": 13407592,
```
magyargergo: **P1 · the heap arm this number was re-recorded for no longer measures the resolver** `[reproduced]`

`heap_reading_bytes.php` 49,574,008 → 13,407,592 reads as a 73% saving. It is the arm going blind.

The probe is `Vendor0\\Ghost\\Missing` (`measure.mjs:2032`, unchanged by this PR) — an **unmatched** namespace against the arm's `App`-only `phpComposerConfigFor`. Under the new gate it returns `null` before a single index read. Confirmed with a spy `SuffixIndex`:

```
probe Vendor0\Ghost\Missing  (committed)  -> null, index reads: NONE
probe App\Ghost\Missing      (mapped)     -> getInsensitive(src/App/Ghost/Missing.php),
                                               getFilesInDir(src/App/Ghost,.php), get(...), ...
```

Measured with a mapped-but-missing probe, PHP's real configured footprint is **50.6 MB** — *more* than BASE's 49.6 MB — against a ceiling this PR cut to **20.2 MB**, i.e. 2.5× below the true reading. An O(files × depth) blow-up of `getFilesByRawDirectory` leaves the committed probe byte-for-byte unchanged and `--check` green; the mapped probe catches it at 62.4 MB.

Running HEAD code against BASE's baselines makes the bench say so itself — the floor arm fires with *"this arm has almost certainly stopped MEASURING rather than started saving… an index it forced is now built lazily behind a question nobody asks."*

**Fix:** point `HEAP_PROBE_TARGET.php` at a mapped-but-missing namespace (`App\Ghost0\Missing`) and re-record `heap_reading_bytes` ≈ 50.6 M / `heap_ceiling_bytes` ≈ 76 M from it; keep a second arm for the external fast path. Also worth noting `HEAP_SHAPE.fields` (`measure.mjs:2459`) pins `files_small`/`files_large`/`path_segments`/`probe` but **not** the `resolutionConfig` — this PR changed what the probe reaches without touching any pinned field.

Found independently by the Codex ce leg and two Claude lanes, then reproduced.

gitnexus/bench/import-target/baselines.json
```diff
@@ -34,7 +35,7 @@
     "dart": 1.6,
     "ruby": 2.2,
     "kotlin": 1.8,
-    "php": 1.9,
+    "php": 2.4,
```
magyargergo: **P1 · this budget went slack** `[reproduced]`

`depth_budget.php` 1.9 → 2.4 against a measured `depth_ratio` of 1.536 — 1.56× headroom. A per-import O(path-depth) probe measures **2.042 / 2.208 / 2.152** across three runs: it **fails at 1.9 and passes at 2.4**, with every other arm green. So the raise costs a real regression class.

The ratio only rose because the gate removed 64% of the denominator (that share of imports now returns at the gate instead of doing depth-proportional work) — not because the deep arm got slower. Re-recording a ratio after the workload stopped exercising the path is how a budget goes slack from a change that makes nothing faster.

**Fix:** re-derive the depth budget from the corpus the fixed heap/miss probes exercise, and keep it at the tightest value the real measurement supports.

gitnexus/src/core/ingestion/import-resolvers/php.ts
```diff
@@ -65,7 +78,13 @@ export function resolvePhpImportInternal(
         // 2. Function/constant fallback: strip last segment (symbol name), scan namespace directory.
         //    e.g. App\Models\getUser → directory app/Models/, find first .php file in that dir.
         const lastSlash = remainder.lastIndexOf('/');
-        const nsDir = lastSlash >= 0 ? dirPrefix + '/' + remainder.slice(0, lastSlash) : dirPrefix;
+        const relativeNamespace = lastSlash >= 0 ? remainder.slice(0, lastSlash) : '';
+        const nsDir =
```
magyargergo: **P1 · this now FABRICATES import edges, and which one is decided by file-walk order** `[reproduced]`

At BASE this expression produced `'' + '/' + relativeNamespace` = `/Models` — a leading-slash key **no directory bucket ever holds**, so the function/constant directory leg was inert for a root or catch-all mapping. (The parity-test comment this PR deletes said exactly that: *"`nsDir` comes out `/Models` which no directory bucket holds either."*) Removing the slash makes the leg live — and it returns `candidates[0]`, the first `.php` file in the directory, for **class** imports too.

```
composer {"autoload":{"psr-4":{"App\\":""}}},  files: Models/Admin.php, Models/User.php
  use App\Models\Ghost;   (no such file)   HEAD = Models/Admin.php     BASE = null
  use App\Models\User;    (control)        HEAD = Models/User.php      BASE = Models/User.php

composer {"autoload":{"psr-4":{"":"src"}}},   files: src/Vendor/Ghost/Other.php
  use Vendor\Ghost\Missing;                HEAD = src/Vendor/Ghost/Other.php   BASE = null

same repo, files enumerated in the other order:
  use App\Models\Ghost;                    HEAD = Models/Zeta.php      BASE = null
```

Two harms, both new. A wrong `IMPORTS` target is worse than `linkStatus:'unresolved'` — `impact` and `detect_changes` act on it. And `candidates[0]` is bucket order = Set insertion order = walker enumeration order, so the same commit analysed on two machines, or after an unrelated file rename, can emit different targets. BASE was order-invariant here (always `null`).

**Nothing pins this in either direction:** mutating line 107 to `if (index && !isCatchAll && dirPrefix !== '')` leaves all **146** PHP tests green. The fix is free.

**Fix:** skip the function/constant directory leg when `dirPrefix === '' || isCatchAll`, or gate it on `importedSymbolKind` being `function`/`const` — the tail of `resolvePhpImportTargetInternal` already knows the kind and already fails closed on ambiguity.

gitnexus/src/core/ingestion/languages/php/import-target.ts
```diff
@@ -48,19 +48,18 @@ function namespaceDirectories(
...
   });
   for (const [namespacePrefix, directoryPrefix] of mappings) {
     const normalizedPrefix = normalizePhpPath(namespacePrefix);
-    if (
-      normalizedTarget !== normalizedPrefix &&
-      !normalizedTarget.startsWith(`${normalizedPrefix}/`)
-    ) {
-      continue;
-    }
+    const matchesNamespace =
```
magyargergo, line 62: **P1 · `use function` / `use const` ESCAPE the gate and bind outside the configured root** `[reproduced]`

This new catch-all branch runs *after* `resolvePhpImportInternal` has already returned `null` from the gate, and its output is looked up in `filesByDirectory`, which is keyed on every directory **suffix alias** (`directoryAliases`, lines 82-93). So the computed directory `src/Vendor/Ghost` matches a file at `other/src/Vendor/Ghost/…` — outside the configured root. `distinctParents.size > 1` does not fire, because there is only one candidate.

Same config, same corpus, only the import kind differs:

```
psr4 = {"": "src"},  file = other/src/Vendor/Ghost/helpers.php (declares Function Vendor\Ghost\helper)
                                                 HEAD                                   BASE
use Vendor\Ghost\helper;          (class)       null                                   null
use function Vendor\Ghost\helper; (function)    other/src/Vendor/Ghost/helpers.php      null
use const Vendor\Ghost\MISSING;   (const)       other/src/Vendor/Ghost/constants.php    null
```

The const row is the worse one: line 468 returns the single candidate for a `const` import with **no declaration check at all**, so the edge points at a file that declares nothing.

This contradicts the PR's own new test `'does not suffix-resolve outside an authoritative catch-all directory'`, which asserts exactly this scenario returns `null` — but omits the `context` argument, so it only covers the class leg.

**Fix:** when the resolver returned `null` *because of the gate*, the declaration search must not re-open the namespace — constrain the catch-all lookup to root-anchored paths (as `getFilesByRawDirectory` already does), or skip the function/const leg when `resolved === null` and the map is authoritative. A one-line test mirroring the existing case with `PHP_FUNCTION_IMPORT` would have caught it.

gitnexus/src/core/ingestion/languages/php/import-target.ts
```diff
@@ -318,23 +317,32 @@ export function loadPhpComposerConfig(repoPath: string): ComposerConfig | null {
...
-        psr4.set(normalizedNs, normalizedDir);
+    for (const section of [autoload, autoloadDev]) {
+      const raw = section?.['psr-4'];
+      if (typeof raw !== 'object' || raw === null || Array.isArray(raw)) continue;
+
+      for (const [ns, dirs] of Object.entries(raw)) {
+        const normalizedNs = ns.replace(/\\+$/, '');
+        const dir = Array.isArray(dirs) ? dirs.find((entry) => typeof entry === 'string') : dirs;
+        if (typeof dir === 'string') {
+          const normalizedDir = dir.replace(/\\/g, '/').replace(/\/+$/, '');
```
magyargergo: **P1 · a leading `./` is never stripped, so `"": "./"` nulls EVERY import in the repo** `[reproduced]`

This normalizes backslashes and *trailing* slashes only. `normalizePhpPath` (line 34, same file) **does** strip `^\./` — so the loader and `namespaceDirectories` disagree about the same config, and `resolvePhpImportInternal` builds `./Vendor/X.php`, which matches no repo-relative key. Because a catch-all miss is authoritative, the gate then returns `null` instead of falling through.

End-to-end from a real `composer.json` on disk through `loadPhpComposerConfig`:

| composer `psr-4` | loaded | import | file present | HEAD | BASE |
|---|---|---|---|---|---|
| `{"": "./"}` | `[["","."]]` | `use Vendor\X;` | `Vendor/X.php` | **null** | `Vendor/X.php` |
| `{"": "./"}` | `[["","."]]` | `use App\Models\User;` | `App/Models/User.php` | **null** | `App/Models/User.php` |
| `{"": "./src/"}` | `[["","./src"]]` | `use Vendor\X;` | `src/Vendor/X.php` | **null** | `src/Vendor/X.php` |
| `{"": "src/"}` (control) | `[["","src"]]` | `use Vendor\X;` | `src/Vendor/X.php` | `src/Vendor/X.php` | same |

`"": "./"` — map the root namespace to the project root — is a legal, documented Composer fallback-directory spelling; the control row shows the identical repo works without the `./`. The PR's catch-all tests all use `['', 'src']` or `['', '/repo/src']`, never a `./` spelling.

**Fix:** strip a leading `./` here and in `language-config.ts:180` (reuse `normalizePhpPath`'s `.replace(/^\.\//, '')`), or normalize `dirPrefix` once at the top of the PSR-4 loop in `php.ts`.

gitnexus/src/core/ingestion/import-resolvers/php.ts
```diff
@@ -103,6 +122,18 @@ export function resolvePhpImportInternal(
         }
       }
     }
+
+    // A non-empty PSR-4 map is authoritative for namespaces it does not own.
+    // Preserve the existing mapped-namespace fallback behavior; #2962 is the
+    // conservative external-namespace gate, not a rewrite of mapped lookup.
+    // A catch-all owns every namespace, so its misses remain authoritative.
+    if (
```
magyargergo, line 134: **P1 · this treats the ROOT `composer.json` as authoritative for the whole workspace, so monorepo sub-packages lose every cross-package edge — silently** `[reproduced]`

`loadPhpComposerConfig(repoPath)` reads only the root manifest.

```
composer.json                 {"autoload":{"psr-4":{"App\\":"src/"}}}
packages/admin/composer.json  {"autoload":{"psr-4":{"Admin\\":"src/"}}}
files: src/App/Http/SomeController.php, packages/admin/src/Controller.php

                              HEAD    BASE
use Admin\Controller;         null    packages/admin/src/Controller.php
use App\Http\SomeController;  (unchanged on both)
```

`packages` is deliberately **not** in the default ignore list, so those files are in `allFilePaths`. Composer path repositories, split-package Laravel and Symfony-style monorepos all have this shape. The loss is silent — the file header (lines 14-15 of `languages/php/import-target.ts`) notes a `null` "lets the finalize algorithm mark the edge as `linkStatus: 'unresolved'`": no log, no diagnostic, no MCP warning, so `impact` blast radius is invisibly truncated.

Codex's adversarial leg reached the same conclusion from the other side and added that **vendor packages' own PSR-4 maps from Composer's generated autoload files are never loaded either** — the same "partial map treated as complete" assumption.

**Fix:** discover sub-directory `composer.json` files and union their PSR-4 maps — which is what **#1881**, the C# gate #2962 names as its model, does for `.csproj` (it reads *every* one, not one). Alternatively scope the gate to imports whose namespace root matches no directory in the workspace, or at minimum document the limitation and emit a diagnostic when the gate fires.

Separately, and to be clear: the *matched*-namespace fallthrough the `gitnexus-check` bot raised on this line is **not** what this comment is about. Your reason for declining it checks out — applying that change fails exactly the 7 pinned cases you cited (`7 failed | 139 passed`), one of them a #2901 invariant.

gitnexus/src/core/ingestion/languages/php/import-target.ts
```diff
@@ -318,23 +317,32 @@ export function loadPhpComposerConfig(repoPath: string): ComposerConfig | null {
...
-        // Normalize directory path (strip trailing slash)
-        const normalizedDir = dir.replace(/\/+$/, '');
-        psr4.set(normalizedNs, normalizedDir);
+    for (const section of [autoload, autoloadDev]) {
+      const raw = section?.['psr-4'];
+      if (typeof raw !== 'object' || raw === null || Array.isArray(raw)) continue;
+
+      for (const [ns, dirs] of Object.entries(raw)) {
+        const normalizedNs = ns.replace(/\\+$/, '');
+        const dir = Array.isArray(dirs) ? dirs.find((entry) => typeof entry === 'string') : dirs;
```
magyargergo: **P1 · collapsing an array-valued PSR-4 entry to its first directory is fatal under a catch-all** `[reproduced]`

For an ordinary prefix this is harmless — the matched-namespace fallback recovers files in the dropped directories (verified: `{"App\\": ["src/","lib/"]}` with the class in `lib/` resolves on both BASE and HEAD). For a **catch-all** it is not, because a catch-all miss is authoritative:

```
composer {"autoload":{"psr-4":{"": ["src/","lib/"]}}}  -> loads [["","src"]]
  use Vendor\X;  with lib/Vendor/X.php   HEAD = null              BASE = lib/Vendor/X.php
  use Vendor\X;  with src/Vendor/X.php   HEAD = src/Vendor/X.php  BASE = same
```

Composer searches every directory in the list; this keeps one. The new test asserts only that the *first* directory is loaded (`php-import-target.test.ts:185-203`, `expect([...psr4.entries()]).toEqual([['App','app']])` for `['app/','fallback-app/']`) — it never resolves a class living in the second.

**Fix:** keep all directories per prefix and iterate them in both `resolvePhpImportInternal` and `namespaceDirectories` (and mirror it in `language-config.ts:179`). If the full fix is out of scope for #2962, the conservative one-liner is to set `hasUnmodeledAutoload` whenever a mapping had directories dropped, so the gate fails open on exactly the manifests this loader cannot model.

Found by both the Codex ce and Codex adversarial legs independently, then reproduced.

magyargergo (commented):
## Two-engine review of `fix(php): gate imports by Composer autoload map`

**How this was produced.** Two engines ran the *same* two review methodologies, plus one adversarial pass and a set of coordinator controls:

| | Claude | Codex (gpt-5.5 / xhigh) |
|---|---|---|
| GitNexus PR swarm | 6 lanes | Solo mode, all personas — **live** |
| ce-code-review | 7 personas | `mode:report-only` — **live** |
| adversarial | ce-adversarial persona | native `adversarial-review` — **live** |

All three Codex legs produced real findings. Plus **12 coordinator controls** run in two isolated checkouts (`pr-2987-ctl` at head, `pr-2987-base` at `e7b096a2`) so nothing here depends on the shared worktree. Everything below is a BASE↔HEAD differential someone executed, or is tagged `[code-read]`.

Because both engines ran the same scaffolding, **agreement between a Codex lane and a Claude lane is the strong signal**; agreement among Claude lanes only (or Codex legs only) is "consistent within one engine", which can be a shared blind spot. Two engines, not six independent confirmations.

---

## What is solid — and it is a lot

- **#2962 is satisfied on its own terms.** All three formal acceptance criteria hold: `Vendor\Ghost\Missing` → `null` against `{App: app}`; `App\Models\User` still → `app/Models/User.php`; PHP is out of `KNOWN_GAPS`. All four wrinkles the issue named up front are implemented *and* covered — `autoload-dev` (`import-target.ts:320`, `language-config.ts:168`), classmap/files fail-open, no-composer fail-open, and the separate function/const leg.
- **The gate itself is well tested.** Twenty mutations were run across two independent lanes. Deleting the gate kills 6 tests; dropping `!hasUnmodeledAutoload` kills 1; dropping `(!matchedNamespace || hasCatchAllNamespace)` kills 7; dropping `sorted.length > 0` kills 1. The fail-open branch is genuinely pinned.
- **The author's scope decision is correct and now independently verified.** The `gitnexus-check` bot asked for matched-namespace misses to be closed too; the author declined, saying it "breaks seven pinned PHP compatibility cases". Applying exactly that change in an isolated checkout gives **`7 failed | 139 passed`** — the number is exact, and one of the seven is a #2901 invariant. Both Codex legs re-raise this same item; it is already answered, and the answer holds. Please don't re-litigate it.
- **The bench baselines were regenerated, not hand-edited.** The committed fingerprints, counts and context arm reproduce byte-for-byte from a fresh report run.
- Typecheck, ESLint and Prettier are clean on the changed files. `git diff --check` clean, no bidi controls, no dependency/workflow/postinstall changes. CHANGELOG correctly untouched, per `DoD.md §4.5`.
- The maintainer's earlier request — *"We shouldn't distinguish between php and this. All php project needs composer. They should act as one."* — was answered properly by `f362acee`: the separate `php_composer` arm was folded into the canonical `php` arm. That was the right consolidation.

---

## Headline: the benchmark that was added does not measure the change

This matters most because it is the thing that was explicitly asked for. Consolidating the arm was right; the resulting arm is blind.

**Instrumenting every exit leg of `resolvePhpImportInternal` and driving `measure.mjs`'s own corpus: 100% of imports reached `suffixResolve` on BASE, and 0% do on HEAD.** 36% now terminate at one `Set.has`, 64% at the new gate's prefix compare. `getInsensitive` and `getFilesInDir` are asked **zero** times across all five arms. Three consequences, each reproduced:

1. **The correctness fingerprints cannot see the gate at all.** Deleting the gate outright, or making it unconditional, leaves all five per-scale php fingerprints **byte-identical** and `--check` green. `baselines.json`'s own new `_php_composer_gate_2962` note claims the arm covers "exact mapped hits, **authoritative misses**, …" — the *misses* half is uncovered, because an authoritative `null` and a fall-through `null` are indistinguishable in this corpus.
2. **The heap arm measures the gate's early return.** The probe `Vendor0\Ghost\Missing` (`measure.mjs:2032`, unchanged by this PR) is an *unmatched* namespace. Under the gate it short-circuits before a single index read — I confirmed this with a spy `SuffixIndex`: committed probe → **index reads: NONE**; a mapped probe (`App\Ghost\Missing`) → `getInsensitive`, `getFilesInDir`, `get`, … So `heap_reading_bytes` 49.6 MB → **13.4 MB is the arm going blind, not a saving**. PHP's real configured footprint measured with a mapped-but-missing probe is **50.6 MB** — *more* than BASE's 49.6 MB — against a ceiling this PR cut to **20.2 MB**, i.e. 2.5× below the true reading. An O(files × depth) blow-up of `getFilesByRawDirectory` leaves the committed probe byte-for-byte unchanged and `--check` green; the mapped probe catches it at 62.4 MB.
3. **A mutated resolver passes everything.** Replacing the O(1) `getFilesInDir` bucket lookup with an O(files) linear scan gives `[import-target --check] PASS` *and* 164/164 green on the index-reuse and parity guards — because that code never executes under this workload.

Running HEAD code against BASE's `baselines.json` makes the bench diagnose itself; its floor arm fires with *"this arm has almost certainly stopped MEASURING rather than started saving… an index it forced is now built lazily behind a question nobody asks."* Re-recording `heap_reading_bytes` alongside the ceiling is exactly what silenced it.

`depth_budget.php` 1.9 → **2.4** is also slack: a per-import O(path-depth) probe measures 2.042 / 2.208 / 2.152 across three runs — **fails at 1.9, passes at 2.4**, every other arm green. And the ratio only rose because the gate removed 64% of the denominator, not because the deep arm got slower.

**Fix:** point `HEAP_PROBE_TARGET.php` at a mapped-but-missing namespace (`App\Ghost0\Missing`) and re-record `heap_reading_bytes.php` ≈ 50.6 M / `heap_ceiling_bytes.php` ≈ 76 M from it; keep a second arm for the external fast path. Add a corpus case where an unmatched namespace *has* a suffix-matchable decoy, so the fingerprint can see the gate. `HEAP_SHAPE.fields` (`measure.mjs:2459`) pins `files_small`/`files_large`/`path_segments`/`probe` but **not** the `resolutionConfig` — this PR changed what the probe reaches without touching any pinned field, so add a config fact there.

**Corroboration:** Codex-ce (engine B) reached finding 2 independently; the Claude risk-architect lane reached it independently; coordinator controls reproduced it. Cross-engine.

---

## Findings

| # | Sev | Claim | Anchor | Corroboration | Tag |
|---|---|---|---|---|---|
| 1 | **P1** | Bench no longer measures the changed path; heap ceiling 2.5× below the true reading; depth budget slack | `baselines.json:111`, `:38` | **Codex-ce + 2 Claude lanes + controls** | `[reproduced]` |
| 2 | **P1** | Root-mapped / catch-all PSR-4 now **fabricates** an `IMPORTS` edge to an arbitrary sibling file, chosen by file-walk order | `php.ts:82` | Claude ce-adversarial + controls | `[reproduced]` |
| 3 | **P1** | `use function` / `use const` **escape** the catch-all gate and bind outside the configured root | `import-target.ts:58` | 2 Claude lanes + controls | `[reproduced]` |
| 4 | **P1** | A catch-all whose directory is written `"./"` or `"./src/"` nulls **every** import in the repo | `import-target.ts:339` | Claude ce-correctness + controls | `[reproduced]` |
| 5 | **P1** | A monorepo sub-package's own `composer.json` is never read, so cross-package edges drop silently | `php.ts:130` | **Codex-adversarial + 2 Claude lanes + controls** | `[reproduced]` |
| 6 | **P1** | Catch-all + multi-directory PSR-4 array drops everything outside the first directory | `import-target.ts:337` | **Codex-ce + Codex-adversarial + controls** | `[reproduced]` |
| 7 | **P2** | `php-import-index-reuse.test.ts` no longer catches the raw-index hand-off it exists to guard | `php-import-index-reuse.test.ts` | Claude ce-testing | `[reproduced]` |
| 8 | **P2** | `autoload-dev`-only `composer.json` flips from fail-open to a repo-wide gate built from a *test* namespace map | `import-target.ts:321` | 2 Claude lanes | `[reproduced]` |
| 9 | **P2** | `autoload-dev` re-declaring a production prefix silently replaces its directory — a *wrong* edge, not a miss | `import-target.ts:331` | Claude ce-adversarial | `[reproduced]` |
| 10 | **P2** | One `"files": [...]` key disables the #2962 gate repo-wide, though `files` can never introduce a namespace | `language-config.ts:174` | Claude ce-adversarial + ce-correctness | `[reproduced]` |
| 11 | **P3** | `hasUnmodeledAutoload` duplicated verbatim in two loaders, one of which (`loadComposerConfig`) has **zero** production callers | `language-config.ts:170` | 3 Claude lanes | `[reproduced]` |

Findings 2–6 share one root cause. **The gate equates "the root `composer.json`'s first-directory-only PSR-4 map did not answer" with "this namespace is external."** That is false whenever the loaded map is a lossy picture of the repo — a `./`-spelled directory the resolver cannot join, a sibling package with its own manifest, or the 2nd..n-th directory of an array mapping. Worth noting: the C# gate **#1881**, which #2962 explicitly names as its model, reads *every* `.csproj`, not one.

Findings 2 and 3 point the other way — they are cases where HEAD is **more permissive** than BASE. Please keep the two directions separate when fixing.

---

### 2 — Fabricated edges, chosen by file-walk order · P1 · `php.ts:82` · `[reproduced]`

`php.ts:82-87` changed `nsDir` for an empty `dirPrefix` from `'' + '/' + relativeNamespace` (= `/Models`, a leading-slash key **no directory bucket ever holds**) to `relativeNamespace` (= `Models`). That makes the function/constant directory leg **live** for root-mapped and catch-all mappings — and the leg returns `candidates[0]`, the first `.php` file in the directory, for **class** imports too. The parity-test comment this PR deleted said exactly that: *"`nsDir` comes out `/Models` which no directory bucket holds either."*

```
composer {"autoload":{"psr-4":{"App\\":""}}},  files: Models/Admin.php, Models/User.php
  use App\Models\Ghost;   (no such file)   HEAD = Models/Admin.php     BASE = null
  use App\Models\User;    (control)        HEAD = Models/User.php      BASE = Models/User.php

composer {"autoload":{"psr-4":{"":"src"}}},   files: src/Vendor/Ghost/Other.php
  use Vendor\Ghost\Missing;                HEAD = src/Vendor/Ghost/Other.php   BASE = null

same repo, files enumerated in the other order:
  use App\Models\Ghost;                    HEAD = Models/Zeta.php      BASE = null
```

Two harms, both new. A wrong `IMPORTS` target is worse than `linkStatus:'unresolved'` — `impact` and `detect_changes` act on it. And `candidates[0]` is bucket order = Set insertion order = walker enumeration order, so the same commit analysed on two machines, or after an unrelated file rename, can emit different targets. BASE was order-invariant here (always `null`).

**No test pins this in either direction:** mutating `php.ts:107` to `if (index && !isCatchAll && dirPrefix !== '')` leaves all **146** PHP tests green. The fix is free.

**Fix:** skip the function/constant directory leg when `dirPrefix === '' || isCatchAll`, or gate that leg on `importedSymbolKind` being `function`/`const` — the tail of `resolvePhpImportTargetInternal` already knows the kind and already fails closed on ambiguity.

### 3 — `use function` / `use const` escape the catch-all gate · P1 · `import-target.ts:58` · `[reproduced]`

The new catch-all branch in `namespaceDirectories` runs *after* `resolvePhpImportInternal` has already returned `null` from the gate, and its output is looked up in `filesByDirectory`, which is keyed on every directory **suffix alias** (`directoryAliases`, `:82-93`). So `src/Vendor/Ghost` matches a file at `other/src/Vendor/Ghost/…`, outside the configured root; `distinctParents.size > 1` doesn't fire because there is only one candidate.

```
psr4 = {"": "src"},  file = other/src/Vendor/Ghost/helpers.php (declares Function Vendor\Ghost\helper)
                                                   HEAD                                  BASE
  use Vendor\Ghost\helper;          (class)        null                                  null
  use function Vendor\Ghost\helper; (function)     other/src/Vendor/Ghost/helpers.php     null
  use const Vendor\Ghost\MISSING;   (const)        other/src/Vendor/Ghost/constants.php   null
```

The const row is the worse one: `import-target.ts:468` returns the single candidate for a `const` import with **no declaration check at all**, so the edge points at a file that declares nothing.

This contradicts the PR's own new test `'does not suffix-resolve outside an authoritative catch-all directory'`, which asserts exactly this scenario returns `null` — but omits the `context` argument, so it covers only the class leg.

**Fix:** when the resolver returned `null` *because of the gate*, the declaration search must not re-open the namespace — constrain the catch-all lookup to root-anchored paths (as `getFilesByRawDirectory` already does), or skip the function/const leg when `resolved === null` and the map is authoritative.

### 4 — `"": "./"` nulls every import in the repo · P1 · `import-target.ts:339` · `[reproduced]`

Both loaders normalize backslashes and *trailing* slashes only — never a leading `./`. `normalizePhpPath` (`import-target.ts:34`) fifty lines away in the same file **does** strip it, so the two code paths disagree about the same config.

End-to-end from a real `composer.json` on disk through `loadPhpComposerConfig` (the registered `loadResolutionConfig`):

| composer `psr-4` | loaded | import | file present | HEAD | BASE |
|---|---|---|---|---|---|
| `{"": "./"}` | `[["","."]]` | `use Vendor\X;` | `Vendor/X.php` | **null** | `Vendor/X.php` |
| `{"": "./"}` | `[["","."]]` | `use App\Models\User;` | `App/Models/User.php` | **null** | `App/Models/User.php` |
| `{"": "./src/"}` | `[["","./src"]]` | `use Vendor\X;` | `src/Vendor/X.php` | **null** | `src/Vendor/X.php` |
| `{"": "src/"}` (control) | `[["","src"]]` | `use Vendor\X;` | `src/Vendor/X.php` | `src/Vendor/X.php` | same |

`"": "./"` — map the root namespace to the project root — is a legal, documented Composer fallback-directory spelling. The control row shows the identical repo works if the directory is written without `./`. The PR's catch-all tests all use `['', 'src']` or `['', '/repo/src']`, never a `./` spelling.

**Fix:** strip a leading `./` when normalizing the directory in both loaders (reuse `normalizePhpPath`'s `.replace(/^\.\//, '')`), or normalize `dirPrefix` once at the top of the PSR-4 loop in `php.ts`.

### 5 — Monorepo sub-packages lose every cross-package edge, silently · P1 · `php.ts:130` · `[reproduced]`

`loadPhpComposerConfig(repoPath)` reads only the root `composer.json`.

```
composer.json                 {"autoload":{"psr-4":{"App\\":"src/"}}}
packages/admin/composer.json  {"autoload":{"psr-4":{"Admin\\":"src/"}}}
files: src/App/Http/SomeController.php, packages/admin/src/Controller.php

                          HEAD                                BASE
use Admin\Controller;     null                                packages/admin/src/Controller.php
use App\Http\SomeController;  src/App/Http/…  (unchanged)     same
```

`packages` is deliberately **not** in the default ignore list, so these files are in `allFilePaths`. Composer path repositories, split-package Laravel and Symfony-style monorepos all have this shape. The loss is silent: `import-target.ts:14-15` notes a `null` "lets the finalize algorithm mark the edge as `linkStatus: 'unresolved'`" — no log, no diagnostic, no MCP warning, so `impact` blast radius is invisibly truncated.

**Fix:** discover sub-directory `composer.json` files and union their PSR-4 maps (what #1881 does for `.csproj`); or scope the gate to imports whose namespace root matches no directory in the workspace; or at minimum document the limitation and emit a diagnostic when the gate fires.

### 6 — Catch-all + multi-directory array · P1 · `import-target.ts:337` · `[reproduced]`

Both loaders collapse an array-valued entry to its first string (`dirs.find(entry => typeof entry === 'string')`). For a *specific* prefix that is harmless — the matched-namespace fallback recovers it (verified: `{"App\\": ["src/","lib/"]}` with the class in `lib/` resolves on both BASE and HEAD). For a **catch-all** it is fatal, because a catch-all miss is authoritative:

```
composer {"autoload":{"psr-4":{"": ["src/","lib/"]}}}  -> loads [["","src"]]
  use Vendor\X;  with lib/Vendor/X.php   HEAD = null              BASE = lib/Vendor/X.php
  use Vendor\X;  with src/Vendor/X.php   HEAD = src/Vendor/X.php  BASE = same
```

The PR's new test asserts only that the *first* directory is loaded (`php-import-target.test.ts:185-203`); it never resolves a class living in the second.

Codex-adversarial adds a related angle worth recording: **vendor packages' own PSR-4 maps from Composer's generated autoload files are never loaded either**, so explicitly-indexed dependency code is mismodelled by the same "partial map treated as complete" assumption.

**Fix:** keep all directories per prefix and iterate them in both `resolvePhpImportInternal` and `namespaceDirectories`. If that is deferred, set `hasUnmodeledAutoload` when any mapping had directories dropped, so the gate fails open — a one-line change that closes this and stays inside #2962's conservative brief.

### 7 — A guard was hollowed out · P2 · `php-import-index-reuse.test.ts` · `[reproduced]`

This file exists to catch the raw-shared-index hand-off (#2901: handing `suffixResolve` the raw index moves IMPORTS edges). With the defect reintroduced: **BASE src + BASE test → FAIL** (`expected 'index.php' to be null`); **HEAD src + the same defect → 4/4 green**, because every negative in the file now short-circuits at the gate before `suffixResolve` ever runs. The guard still passes; it no longer guards.

**Fix:** keep the three negatives but assert the root-level and suffix-fallback shapes with `null` as the resolution config, e.g. `resolveImportTarget('index', FROM_FILE, files, null)` → `null` and `resolveImportTarget('Legacy\\Helper', FROM_FILE, files, null)` → `'lib/Legacy/Helper.php'`. Also update the stale header at line 53, which still claims the corpus carries `lib/Legacy/Helper.php` "for the suffix fallback that runs when no PSR-4 prefix matches" — under `COMPOSER` that leg is now unreachable.

### 8, 9 — `autoload-dev` · P2 · `import-target.ts:321`, `:331` · `[reproduced]`

`if (autoload === undefined)` became `if (autoload === undefined && autoloadDev === undefined)`. A composer.json with only `autoload-dev` used to return `null` (no gate, full suffix matching); it now returns a config whose map covers only *test* namespaces, and that map becomes repo-wide authoritative — `use App\Models\User` goes `src/Models/User.php` → **`null`**. A mutation reverting this survives all six test files.

Separately, because `autoload-dev` entries overwrite `autoload` entries for the same key, an `autoload-dev` re-declaration of a production prefix silently *replaces* its directory: `use App\Models\User` → `tests/Models/UserTest.php`, where BASE gave `src/Models/User.php`. That is a wrong edge, not a miss.

### 10 — One `"files"` key disables the gate repo-wide · P2 · `language-config.ts:174` · `[reproduced]`

`files` cannot introduce a namespace — it is a list of unconditionally-required scripts — yet its presence sets `hasUnmodeledAutoload` and turns #2962 off for the whole repository. Guzzle, Symfony components and laravel/framework all ship a `files` entry, so the gate is inert for a large share of real PHP projects. `psr-0` and `classmap` genuinely can introduce namespaces and belong in the list; `files` arguably does not.

### 11 — Duplicated logic, half of it dead · P3 · `language-config.ts:170` · `[reproduced]`

The `['psr-0','classmap','files'].some(key => key in section)` expression is now written character-for-character in two places. More to the point: **`loadImportConfigs` — and therefore `loadComposerConfig` and the whole `language-config.ts` half of this diff — has zero callers anywhere in `src/` or `test/`.** Live PHP resolution flows only through `loadPhpComposerConfig` → `resolvePhpImportTargetInternal`. Not introduced by this PR (it mirrors the documented dead-leg pattern for Go at `import-resolvers/go.ts:6-15`), but worth knowing before anyone credits the `language-config.ts` changes with production effect.

---

## Suspicions probed and refuted

Stating these because they were the obvious worries and they did **not** hold:

- **"The gate is inert in production."** Refuted, twice. `resolvePhpImportTarget` (`import-target.ts:381`, `composerConfig: null`) is registered at `languages/php.ts:326`, but its only reader — `import-target-adapter.ts:resolveImportTargetAcrossLanguages` — is test-only. Both production paths (`parse-impl.ts:1553`, `run.ts:714-717`) pass the loaded config. **#2962 is fully wired, not half-fixed.**
- **"`index.getInsensitive` suffix-probes into `vendor/` before the gate."** Refuted. `import-target.ts:274-286` hollows `getInsensitive` to return `undefined` unconditionally in the parity view, and the call site sits *inside* the matched-prefix branch the gate deliberately never blocks.
- **"The 9 → 13 rise in parity divergences is a bug."** No. `expectedViaWorkspace` is defined (`:301`) as what the *LanguageProvider* adapter returns, and that adapter has no Composer config by design. The rise is expected. (It does mean those assertions pin a test-only adapter — see coverage gaps.)
- **"Multi-directory PSR-4 arrays regress on the live path."** Refuted for ordinary prefixes; the BASE *throw* was only in the dead `language-config.ts` loader. Only the catch-all interaction is real (finding 6).
- **"The parity differential is circular, so the gate is untested."** Half-refuted. `expect(current).toBe(legacy)` genuinely cannot fail — both sides call `resolvePhpImportInternal`. But the pinned `expected` literals *are* real coverage and killed every gate mutation tried. Worth hardening (make the legacy side a snapshot copy), not worth blocking on.
- **Security.** No merge-blocking issue. Prototype pollution via `__proto__` psr-4 keys does not work (V8's `JSON.parse` never makes it an own property for a string value; for an object value it is an own data property with no prototype-chain effect). Non-string psr-4 values, `"autoload": null`, malformed sections and a 10,000-entry map all fail open cleanly. A `psr-4` dir of `"../../etc"` does produce a `../`-shaped candidate path, but the result is only ever a key into in-memory symbol tables — never a `readFile` — and production `allFilePaths` comes from the repo walker, so no traversal path can be a member. P3, defence-in-depth only.
- **Ordering / tie-break nondeterminism in `getSortedPsr4`.** Refuted: two equal-length prefixes that both match the same target are necessarily the same string, and `Map` de-duplicates. The real order-dependence is finding 2, in `candidates[0]`.
- **Committed `vendor/` directories losing correct edges.** Refuted — `src/config/ignore-service.ts:28` lists `vendor` in the default ignore set.
- **Project standards.** No violation. AGENTS.md's language-naming rule is scoped to the scope-resolution pipeline; `ARCHITECTURE.md` documents `import-resolvers/<lang>.ts` as the intended pattern and `language-config.ts` already had a `LANGUAGE-SPECIFIC CONFIG TYPES` section with `ComposerConfig` in it before this PR.

---

## Test-coverage gaps

Each of these is a real behaviour with no test, and each corresponds to a finding above:

1. `php/php-import-target.test.ts` — a catch-all whose dir is spelled `"./"` / `"./src/"`, asserting a class still resolves.
2. `php/php-import-target.test.ts` — the existing `'does not suffix-resolve outside an authoritative catch-all directory'` case repeated with `PHP_FUNCTION_IMPORT` and the `context` argument (finding 3), plus a `const` variant.
3. `php/php-import-target.test.ts` — root-mapped `{"App\\": ""}` with a *class* import that misses, asserting `null` rather than an arbitrary sibling (finding 2).
4. `php/php-import-target.test.ts` — catch-all with `["src/","lib/"]` and the class in `lib/` (finding 6).
5. A monorepo case: root manifest plus `packages/*/composer.json`, asserting the cross-package edge survives (finding 5).
6. An `autoload-dev`-only manifest with a production-namespace import (finding 8), and an `autoload-dev` re-declaration of a production prefix (finding 9).
7. `php-import-index-reuse.test.ts` — restore the raw-index guard per finding 7.
8. Six new branches have no test at all: `namespaceDirectories`' catch-all and leading-`/` strip, `hasUnmodeledAutoload` sourced from `autoload-dev`, the `psr-0` key, the `Array.isArray` psr-4-section guard, and the class-style empty-`dirPrefix` mapping in isolation. Mutations M8, M9, M10, M14, M15, M16 and M17 each survive all six test files.

Two smaller notes: the flagship `'rejects namespaces outside an authoritative PSR-4 map'` test survives deletion of *both* PSR-4 resolution legs, because its positive assertion is satisfied by the suffix fallback. And the divergence count `13` would be better as a pinned list of case names than a bare number; the prose justifying the counts is now stale ("32 of the 86 arms", actually 33 of 92). The `external-import-conformance.test.ts` corpus change is **not** a weakening — the new corpus still fails against BASE src, and the old one was structurally impossible to keep once PHP left `KNOWN_GAPS`.

---

## CI and state

38 checks pass, 1 fails, 1 skipped. The failure is **Vercel — "Authorization required to deploy"**, pre-existing infrastructure, unrelated to this PR. `tests / benchmarks (GITNEXUS_BENCH)` passes — which is the point of the headline finding: it is green while the gate's behaviour is invisible to it. `mergeStateStatus: BEHIND` (main advanced), `reviewDecision: CHANGES_REQUESTED` from the existing undismissed review.

## Coverage limits

No end-to-end `analyze` over a real PHP repository was run — no PHP corpus was available. Every reproduction above is at the `resolvePhpImportTargetInternal` boundary, which is exactly what `run.ts:714-717` and `parse-impl.ts:1553` call, with `composer.json` files written to disk and read through the registered `loadResolutionConfig`. Findings 5 and 8 additionally depend on the repo walker including `packages/**`, which was verified against the default ignore set but not observed in a live analyze.

Process note, not a PR defect: three Claude lanes found the shared review worktree dirty from another lane's live mutation mid-run and worked around it with private copies and `git show HEAD:`. Their results are unaffected; the coordinator's controls ran in separate checkouts throughout.

---

*Automated multi-tool digest (Claude + Codex gpt-5.5, two review methodologies each, plus coordinator controls). Verify before acting on any individual item.*

azizur100389:
Addressed the review in cbbc5a7a:\n\n- hardened Composer parsing for root/./ paths, multi-dir maps, production vs autoload-dev authority, and package-local manifests\n- removed catch-all/root directory false positives for class, function, and const imports\n- made the PHP benchmark observe both mapped misses and the external gate; tightened the depth budget to 1.8\n- restored raw-index parity coverage and added the requested regressions\n\nValidated: 384 affected tests + 19 focused tests, typecheck, lint/format, and import-target --check. Synced with current main.

magyargergo (approved):

Issue #2959, opened 2026-08-14
Tracking: external imports resolve to in-repo files in 9 remaining languages


magyargergo:
Parent for the per-language work left after #2953. **Not a new defect** — this is the inventory of where a known one still lives, made measurable.

## The property

> A specifier naming something **outside** the repository must not resolve to a file inside it.

Violating it does not lose an edge, it **fabricates** one: an `IMPORTS` edge at full confidence between two files with no relationship, which `impact`, `trace` and `context` then report as real. #2953 measured 44 of 74 cross-package edges landing on two unrelated files in one monorepo.

The mechanism is shared, which is why the list is long: `import-resolvers/utils.ts:suffixResolve` asks *"does any file in this repo have a path ending in this specifier?"* and retries with each leading segment dropped. Most languages still end their resolution chain there.

## Where each language stands

`test/unit/scope-resolution/external-import-conformance.test.ts` asserts this property for **all 16** registered resolvers, so the census is executable rather than written down. Its `KNOWN_GAPS` map is this list, and each entry records the exact wrong answer.

| holds it | how |
|---|---|
| TypeScript, JavaScript, Vue | resolve against declared config only — no suffix fallback (#2953) |
| Python | #898 — gate keyed on the leading segment naming something in-repo |
| C# | #1881 — same shape, keyed on declared namespaces |
| Java | #2910 — resolves against declared `package` headers. **Not on `main` yet**: the fix is in an unmerged PR stacked on #2956, so Java still fails the suite here and keeps its `KNOWN_GAPS` entry until that lands |
| Rust | `::` never decomposes into a path suffix, so the question cannot arise |

| gap | ticket |
|---|---|
| Kotlin | #2960 |
| Go | #2961 |
| Ruby | #2966 |
| PHP | #2962 |
| Dart | #2963 |
| Swift | #2964 |
| C / C++ | #2965 — one resolver, one fix |
| COBOL | #2967 |

The eight below are attached as sub-issues of this one. C and C++ get **one** ticket on purpose: `resolveCppImportTarget` delegates to `resolveCImportTarget`, so two tickets would be two people fixing one code path.

## Two shapes of fix, and which to prefer

**Gate a guess** (Python #898, C# #1881) — keep suffix matching, refuse to enter it unless the specifier's leading segment names something declared in-repo. Cheap, and it holds. But it answers *"probably not external"* rather than *"here is what this specifier means"*, so it is a filter on a heuristic rather than a resolution rule.

[...]

PR #2990, opened 2026-08-18
fix(kotlin): resolve imports from declared packages

### LINKED ISSUE #2960: fix(kotlin): third-party imports resolve to coincidentally-named local files
Child of #2959. Kotlin is the closest repeat of #2910 in the repository, and most of the machinery is already there.

## The defect

```
import org.junit.Assert;   →  src/main/kotlin/vendor/Assert.kt
```

Reproduced by the `kotlin` arm of `test/unit/scope-resolution/external-import-conformance.test.ts` (currently recorded in `KNOWN_GAPS`). Workspace:

```
src/main/kotlin/vendor/Assert.kt
src/main/kotlin/com/example/App.kt      ← the importer
```

Nothing declares `org.junit`, but `resolveKotlinImportTarget` turns the dots into slashes and walks down to a matching path tail, so a JUnit import binds to an unrelated local file. Every `org.*`, `kotlinx.*`, `androidx.*` and `java.*` import in a real project is a candidate.

## Why this one should be cheap

Kotlin already parses and stores its `package` header, exactly as Java does: `languages/kotlin/package-facts.ts` builds a `createJvmPackageFactStore({ packageNodeType: 'package_header', … })`, it is transported across the worker boundary on the capture side channel, and it is populated by the time `resolveImportTarget` runs. Nothing reads it during resolution.

That is the same gap #2910 closed for Java, and the fix there is a direct template: build a `package → type → file` index from `context.parsedFiles`, resolve the longest declared-package prefix, and answer `null` for anything the workspace does not declare. See `languages/java/module-resolution.ts`.

## Kotlin-specific wrinkles worth knowing before starting

- **package need not match directory** — more common in Kotlin than Java, and it is the reason to resolve from the header rather than the path.
- **file-level facades**: `Utils.kt` declaring `package com.example` and top-level functions compiles to `com.example.UtilsKt`. An `import com.example.foo` names a top-level *function*, not a type, so the Java rule "type name = file base name" does not cover it. Decide deliberately whether a member import resolves to the declaring file.
- **`@JvmName`** can rename that facade.
- **star imports**: `import com.example.*` should answer every file declaring that package, the way #2910 made Java's wildcards answer the whole package instead of one arbitrary file.

## Acceptance

- `import org.junit.Assert` in the fixture above resolves to `null`.
- A legitimate in-repo import still resolves — the paired positive in the conformance case.
- `KNOWN_GAPS` loses its `kotlin` entry and that arm goes green with no other edit to the suite.

azizur100389:
## Summary

- Resolve Kotlin imports from declared package facts and module-scope declarations instead of path suffixes.
- Preserve local types, top-level declarations, member imports, wildcard imports, and overload candidate sets while leaving external imports unresolved.
- Replace path-specific guards with semantic conformance, integration, cache-reuse, fingerprint, scaling, and heap coverage.

## Validation

- 319 affected unit, conformance, reuse, and Kotlin resolver tests passed.
- TypeScript typecheck passed.
- Changed TypeScript/JavaScript files passed ESLint and Prettier.
- Kotlin import correctness and scaling benchmark passed with its deterministic fingerprint.
- GitNexus change analysis reports LOW risk and zero affected execution processes.

Closes #2960
Refs #2959

---

## Summary

*A high risk Kotlin import resolution change with a broad set of benchmark, resolver, fixture, and test updates. The graph shows concentrated reach within Kotlin ingestion and import-target handling, with no traced execution flows or cross-repository consumers.*

**🟠 HIGH blast radius. A Kotlin ingestion and import resolution change spanning resolver, benchmark, test, fixture, and CI code, with three direct dependents.**

The primary implementation is in `gitnexus/src/core/ingestion/languages/kotlin/import-target.ts`, centered on `resolveKotlinImportTarget` and `KotlinResolveContext`, with related changes in `gitnexus/src/core/ingestion/languages/kotlin/index.ts`, `gitnexus/src/core/ingestion/languages/kotlin/module-resolution.ts`, and `gitnexus/src/core/ingestion/languages/kotlin/scope-resolver.ts`. The change covers 23 files and 32 symbols, with graph impact landing in `Kotlin`, `Import-target`, `Integration`, and `Vue`.

Review the Kotlin resolution logic and its parity and reuse tests first, especially `gitnexus/test/integration/resolvers/kotlin.test.ts`, `gitnexus/test/unit/kotlin-module-resolution.test.ts`, and the tests under `gitnexus/test/unit/scope-resolution/kotlin/`. The hottest file is `gitnexus/bench/kotlin-import-target/measure.mjs`, while `.github/workflows/ci-tests.yml` is the only HIGH or CRITICAL risk file.

_Added by GitNexus for PR #2990. Edit freely — this block is replaced on the next review, everything above it is left untouched._


azizur100389:
Addressed both review findings in 1a2cf801: preserved the existing countedParsedFiles contract via a separate semantic-fixture helper, and removed the misleading interpolation from the benchmark path normalization. Validated 339 focused tests, typecheck, ESLint, the Kotlin benchmark, and a LOW-risk graph audit with zero affected flows.

azizur100389:
Added the #2960 Kotlin fixture-corpus scope-capture rebaseline with explicit provenance; performance ceilings are unchanged. The Kotlin guard now matches the 141-file corpus and remains linear (0.758 < 1.5).

koriyoshi2041 (approved):
Verified exact head `cee967e8`. The resolver now keys Kotlin imports to parsed package facts plus module-scope local bindings, so mismatched source paths still resolve while external packages and imported-only bindings do not fabricate local targets. I also checked package/class wildcard handling, root-package and unreadable-package behavior, and preservation of multi-file top-level overload candidates.

Local validation: 274 focused Kotlin resolver/integration/conformance tests passed, `npx tsc --noEmit` passed, and Prettier passed on the touched resolver and test files. The hosted CI Gate and full platform matrix are green; the remaining Vercel authorization status is unrelated to the CLI-only change.

Issue #771, opened 2026-04-10
Feature: Allow indexing test directories (__tests__, __mocks__)


svasir-opcity:
## Problem

`__tests__` and `__mocks__` directories are hardcoded in `DEFAULT_IGNORE_LIST` in `config/ignore-service.js` (lines 63-64). The comment in the source says negation patterns in `.gitnexusignore` cannot override these entries.

For Quality Engineering use cases, test files are the primary thing we need indexed — we need to trace which tests cover which production code using IMPORTS/CALLS edges.

## Current Workaround

Manually patching the installed package to remove `__tests__` and `__mocks__` from the hardcoded list. This works but gets overwritten on every `npm update`.

## Proposed Solution

Add a `--include-tests` flag to `gitnexus analyze`:

```bash
gitnexus analyze --include-tests
```

Or make the DEFAULT_IGNORE_LIST overridable via `.gitnexusignore` negation patterns (`!__tests__`).

## Use Case

As a Quality Enablement Engineer, I use GitNexus to:
1. Map production code paths across repos (resolvers, datasources, components)
2. Find which test files import/cover those code paths via IMPORTS edges
3. Assess test coverage sufficiency per testing pyramid layer

Without test files indexed, step 2 is impossible — the IMPORTS edges from test to production code do not exist in the graph.

## Impact

After patching, inquiry-subgraph went from 514 to 669 nodes and 1,248 to 1,672 edges. 51 test files now appear in the graph with proper IMPORTS edges to the production code they test.

## Bonus: TestCase Node Type

Currently GitNexus indexes functions, classes, methods, and interfaces but not individual `test()` / `it()` blocks. A new TestCase node type that extracts the test name from `test("should store inquiry", ...)` would enable counting actual test cases (not just files) and understanding what each test validates.

Issue #596, opened 2026-03-31
v1.4.10: .gitnexusignore negation patterns silently produce 0 nodes


SethCalkins:
## Environment
- macOS 15 (arm64, Apple M1 Max)
- Node 22
- gitnexus v1.4.10 (via `npx -y gitnexus@latest`)

## Bug

When a `.gitnexusignore` file uses the gitignore-style "exclude all then whitelist" pattern, v1.4.10 produces **0 nodes / 0 edges**. The same file works correctly in v1.3.11 (55k nodes).

### `.gitnexusignore` contents:
```
# Exclude everything by default
*

# Include iOS source code
!iOS/
!iOS/**

# Include only backend/living_plan
!backend/
!backend/living_plan/
!backend/living_plan/**

# Exclude tests everywhere
**/Tests/
**/tests/
```

### Steps to reproduce
1. Create the `.gitnexusignore` above in a repo with Python + Swift code
2. Run `npx -y gitnexus@1.4.10 analyze`
3. Observe: `0 nodes | 0 edges | 0 clusters | 0 flows` (completes in ~1.9s, clearly not parsing)
4. Remove `.gitnexusignore` and re-run → `38,872 nodes` (Python only, Swift skipped separately)
5. Run with `npx -y gitnexus@1.3.11 analyze` with the same ignore file → `55,429 nodes` ✅

### Root cause
The `!` negation patterns (whitelist after `*` blacklist) appear to be silently broken in v1.4.10. The analyzer doesn't error; it just finds 0 files to parse.

## Secondary issue: tree-sitter-swift missing from v1.4.10

Both v1.3.11 and v1.4.10 fresh `npx` installs ship without `tree-sitter-swift`, producing:
```
Skipping 976 swift file(s) — swift parser not available (native binding may not have built)
```

Manually installing it into the npx cache fixes this:
```
npm install --prefix <npx-cache-dir> tree-sitter-swift
```

Swift is listed as a supported language in the README, so it should either be bundled or installed automatically.

## Workaround
Pin to v1.3.11:
```
claude mcp add gitnexus -- npx -y gitnexus@1.3.11 mcp
npx -y gitnexus@1.3.11 analyze
```

PR #536, opened 2026-03-27
feat: unify web and cli ingestion pipeline

magyargergo:
## Summary

Complete architectural migration of GitNexus web from a self-contained browser app (WASM LadybugDB, Web Worker, in-browser embeddings) to a thin client backed by the CLI server. Also creates `gitnexus-shared` package for cross-package type unification.

## Architecture

```
BROWSER (gitnexus-web)
├─ OnboardingGuide → auto-detect server → auto-connect → graph explorer
├─ RepoAnalyzer → GitHub URL or Local Folder → SSE progress → graph loads
├─ Header → repo dropdown (switch / re-analyze / delete) + search
├─ CodeReferencesPanel → buffered file reading → scroll to selected line
├─ GraphCanvas + FileTree + NexusAI + StatusBar
└─ BackendClient → all queries via HTTP to localhost:4747

SERVER (gitnexus serve — Express)
├─ GET  /api/heartbeat     SSE liveness (instant disconnect detection)
├─ GET  /api/info           Version, launch context, Node.js version
├─ GET  /api/repos          List indexed repositories
├─ GET  /api/repo           Repo metadata
├─ DELETE /api/repo         Delete repo (index + clone + registry)
├─ GET  /api/graph          Nodes + relationships
├─ GET  /api/file           Read file (supports ?startLine=N&endLine=M)
├─ POST /api/query          Cypher (write-guarded)
├─ POST /api/search         Hybrid search + enrichment
├─ GET  /api/grep           Regex over file contents
├─ POST /api/analyze        SSE progress → child_process (8GB, auto-retry)
├─ POST /api/embed          SSE progress for embedding generation
├─ ALL  /api/mcp            MCP-over-StreamableHTTP
└─ LadybugDB + tree-sitter + FTS

gitnexus-shared (types)
├─ GraphNode, GraphRelationship, NodeLabel, NODE_TABLES
├─ SupportedLanguages (exhaustive Record enforcement)
├─ PipelinePhase, PipelineProgress
└─ getLanguageFromFilename, getSyntaxLanguageFromFilename
```

### Onboarding & UX (new)
- **Auto server detection** — polling via setTimeout chain (3s), Page Visibility API pause, SSE heartbeat for instant disconnect detection
- **OnboardingGuide** — step-by-step instructions (copy → run → auto-connect), smart command detection (dev vs prod via `import.meta.env.DEV`)
- **Crossfade transitions** — smooth phase changes (onboarding → success → loading → exploring) with snapshot pattern
[...]

PR #1156, opened 2026-04-28
feat: add IncludeExtractor for C++ cross-repo include tracking (group)

SZU-WenjieHuang:
## Summary

Closes #1142

Add cross-repo `#include` dependency tracking to the Group feature, enabling impact analysis across C/C++ repositories that share header files (e.g., an InterfaceMap repo providing `.h` headers consumed by a VMap repo).

### What changed

**New: `IncludeExtractor`** (`gitnexus/src/core/group/extractors/include-extractor.ts`)
- **Provider side**: registers every `.h/.hpp/.hxx/.hh` file in a repo as a provider contract (`include::<relative-path>`)
- **Consumer side**: parses all C/C++ source/header files for `#include "…"` directives via tree-sitter, resolves against the repo's own file list (reusing `suffixResolve` from the ingestion pipeline), and emits unresolved paths as consumer contracts
- **Regex fallback**: when tree-sitter's native binding fails on files > 32 KB (`parser.parse()` throws `Invalid argument`), falls back to regex extraction (`/^[ \t]*#\s*include\s*"([^"]+)"/gm`) so large files are not silently skipped
- Filters angle-bracket includes, well-known system/stdlib headers, and system path prefixes

**Integration into Group sync** (`sync.ts`)
- `IncludeExtractor` runs when `config.detect.includes` is enabled (default: `true`)
- `writeBridge()` is now called after `writeContractRegistry()` to generate `bridge.lbug` for cross-repo impact queries

**Type system & matching** (`types.ts`, `config-parser.ts`, `matching.ts`, `manifest-extractor.ts`)
- Added `'include'` to `ContractType` union and `DetectConfig`
- Added `include::` normalization in `normalizeContractId`
- Added `include` type graph-lookup branch in `ManifestExtractor`

### Key design decisions

1. **Suffix-based local resolution**: reuses the existing `buildSuffixIndex` / `suffixResolve` from the single-repo ingestion pipeline to distinguish local includes from cross-repo ones — only unresolved includes become consumer contracts
2. **tree-sitter + regex dual strategy**: tree-sitter provides accurate AST-based extraction for files ≤ 32 KB; regex handles the long tail of large files (33 files in a 415-file VMap repo exceeded the limit)
3. **Contract ID format**: `include::map/base/dice_map_view.h` — path-normalized, lowercase, matched via `runExactMatch`

### Validation (on two of my real repos)

| Metric | Before | After |
|--------|--------|-------|
| Total contracts | — | 1191 |
[...]

PR #3018, opened 2026-08-20
fix(ingestion): ignore emitted Next.js build output, and delete the inert public/build entry

### LINKED ISSUE #3007: fix(ingestion): _next missing from DEFAULT_IGNORE_LIST, so Next.js bundle output is indexed as source and Route nodes point at minified chunks
### Area

### Summary

`_next` is missing from the walker's `DEFAULT_IGNORE_LIST` (while `.next` is present), so committed or emitted Next.js bundle output gets indexed as source — producing `Route` nodes that point at minified webpack chunks.

### Context

Found while investigating cross-repo linking on a Capacitor-wrapped Next.js app, where the production build is copied into `android/app/src/main/assets/public/` and left in the working tree. On that repo **446 of 1123 indexed files (40%) came from `android/` or `ios/`**, 262 of them under `_next/static`. The paired backend was clean (0 of 709).

Verified against `main` @ `aac7515d`.

### Expected behavior

Emitted bundle output is not indexed as source, and `Route` nodes describe routes a human wrote.

### Actual behavior

**Root cause.** `src/config/ignore-service.ts` `DEFAULT_IGNORE_LIST` contains `.next` (the dotted build *cache*) but not `_next` (the emitted *output* directory). None of the segments in the Capacitor path — `android`, `app`, `src`, `main`, `assets`, `public`, `_next`, `static`, `chunks` — are in the list.

The filename heuristics don't catch it either. `isIgnoredFile` only matches `.bundle.`, `.chunk.`, `.generated.` and `.d.ts`, but Next.js emits hashed names like `6862-9d1cdcb99f169a06.js`, which contain none of those.

Result — every `Route` node on that index came from bundle output:

```
Route:POST /api/notifications/register-device
  filePath: android/app/src/main/assets/public/_next/static/chunks/6862-9d1cdcb99f169a06.js
```

Any route answer drawn from these points a reader at a webpack chunk rather than at source.

Minified single-line bundles are also what produced `callable-value-flow: candidate set exceeded the cap` warnings during analysis, so this costs time as well as accuracy.

**Adjacent bug, same file.** `ignore-service.ts:63` has `'public/build'` as a `DEFAULT_IGNORE_LIST` entry, but matching is:

```ts
const parts = normalizedPath.split('/');
for (const part of parts) {
  if (DEFAULT_IGNORE_LIST.has(part)) return true;   // ignore-service.ts:~304
}
```

A slash-containing entry can never equal a single path segment, so `'public/build'` has never matched anything. It needs either the two-segment handling the name implies, or removal.

**Related, may warrant its own issue.** On the same index:

```
MATCH (r:Route) RETURN count(*)                                 →  21
MATCH ()-[r:CodeRelation {type:'FETCHES'}]->() RETURN count(*)  →   0
```

With zero `FETCHES` edges, `extractConsumersGraph()` yields nothing, because it is driven by `FETCHES_QUERY` in `extractors/http-route-extractor.ts:57`:

```
MATCH (callerFile:File)-[r:CodeRelation {type: 'FETCHES'}]->(route:Route)
```

So the graph-assisted consumer path is dead for this repo and everything falls to the source scan. `FETCHES` edges are only produced by `processNextjsFetchRoutes` (`call-processor.ts:623`) and are JS/TS-only by design (documented in `http-patterns/types.ts:133` and `java.ts:778`) — but here even the JS/TS repo has none, because the only `Route` nodes to match against came from bundle output rather than source.

### Steps to reproduce

1. Take a Next.js app whose built output is present in the working tree but untracked — e.g. a Capacitor app with `android/app/src/main/assets/public/_next/static/…`
2. `npx gitnexus analyze`
3. `npx gitnexus status` — note the indexed file count includes the bundle directories
4. Query `MATCH (r:Route) RETURN r.name, r.filePath` — every route resolves to a `_next/static/chunks/*.js` file

### Environment

- OS: macOS
- Node: 22.22.0
- Commit: `aac7515d` (main)

### Logs / screenshots

Suggested fix:

1. Add `_next` to `DEFAULT_IGNORE_LIST` alongside the existing `.next`. Consider `assets/public` for the Capacitor convention specifically.
2. Fix or remove the dead `'public/build'` entry.
3. Optionally, treat a JS file whose longest line exceeds some threshold as generated — it catches every minifier's naming scheme instead of chasing each one.

Note these are working-tree files that are **not tracked in git**, so the walker is picking them up off disk; a `.gitignore`-only defense doesn't cover it.

ReidenXerx:
## Summary

Fixes #3007. `DEFAULT_IGNORE_LIST` had `.next` (the build *cache*) but not `_next` (the emitted *output*), so a Capacitor/Cordova-wrapped Next.js app had its shipped bundle indexed as source — and every `Route` node such a repo produced pointed at a minified webpack chunk instead of at code anyone wrote.

Also restores `'public/build'`, which had been sitting in that set matching nothing at all.

## Motivation / context

`.next` and `_next` are different directories. A Capacitor shell copies the built output to `<platform>/app/src/main/assets/public/_next/static/…`, and none of `android`, `app`, `src`, `main`, `assets`, `public`, `_next`, `static` or `chunks` is in the ignore list.

The filename heuristics do not catch them either — `isIgnoredFile` matches `.bundle.`, `.chunk.` and `.generated.`, while Next.js emits hashed names like `6862-9d1cdcb99f169a06.js`.

The visible symptom on a real repo was that every `Route` node came from bundle output:

```
Route:POST /api/notifications/register-device
  filePath: android/app/src/main/assets/public/_next/static/chunks/6862-9d1cdcb99f169a06.js
```

Minified single-line bundles were also the source of `callable-value-flow: candidate set exceeded the cap` warnings during analysis, so this costs time as well as accuracy.

**The second bug.** `'public/build'` was a member of `DEFAULT_IGNORE_LIST`, but that set is tested one path SEGMENT at a time:

```ts
const parts = normalizedPath.split('/');
for (const part of parts) {
  if (DEFAULT_IGNORE_LIST.has(part)) return true;
}
```

and it is also read by `isHardcodedIgnoredDirectory(name)`, which receives a bare directory name. A slash-containing member can never compare equal to either, so the entry had never matched anything since it was added.

## Areas touched

- [x] `gitnexus-web/` (Vite / React UI)

## Scope & constraints

**In scope**

- Add `_next` to the segment ignore list, and pin that it matches as a whole path segment.
- Delete the inert `'public/build'` entry outright (see Implementation notes for why deleting beat relocating).
- A guard test pinning the invariant that made the dead entry possible.
- Add `_next` to the web upload filter's `EXCLUDED_DIRS`, which had the same `.next`/`_next` gap on the live browser ingestion path.

- **No "looks minified" heuristic.** #3007 floats treating a very long single line as generated. It would catch every bundler's naming scheme at once, but it is a content-based judgement that can misfire on legitimate generated-and-wanted source, so it deserves its own change with its own evidence.
- **Other bundlers' output directories** (`.svelte-kit/output`, `.angular/cache`, Vite's `.vite`) are not audited here; only the reported `_next` gap and the already-intended `public/build` are addressed.
- ~~**`env` over-matching.**~~ *Resolved on `main` while this PR was open:* #3043 removed `'env'` and `'generated'` from the name set and gated them behind `ROOT_ARTIFACT_DIRECTORIES` + a `pyvenv.cfg` probe, so a JS project's `src/lib/env/` is no longer skipped. Nothing left to do here.
- **A repo with hand-written source in a directory literally named `_next`** loses that subtree, because `_next` goes into the *unanchored* name set. Measured on fixtures: a `packages/_next/` source tree loses 5 of 11 files including its `package.json`, `tsconfig.json` and `go.mod`. Judged acceptable — `_next` is Next.js's reserved *emitted* path, not a source convention, and unlike `'packages'` (deliberately removed from this set for exactly that reason) there is no observed repo that uses it for source. A working remedy exists if one ever appears: child-anchored fragments (`_next/static`, `_next/server`, `_next/data`, `_next/image`) plus an `existsSync` probe in `isHardcodedIgnoredDirectoryAtPath`, following the pattern #3043 established for `env`/`generated`. Verified to keep all tests green while recovering the source tree and still dropping every chunk. Not adopted here because the trigger is hypothetical.
- **The `.gitnexusignore` escape hatch is spelling-sensitive** and this PR does not change that. `!_next`, `!_next/` and `!**/_next/` recover the files; `!_next/*`, `!_next/**` and `!**/_next/**` *alone* are silent no-ops, because `childrenIgnored` prunes the directory before any file reaches `ignored()`. The repo's own #596 test writes the working `!dir/` + `!dir/**` pair. Pre-existing, unchanged by #3043, worth its own issue.
- **An ancestor negation voids the new rule.** `!android/` in `.gitnexusignore` *or* `.gitignore` short-circuits `hasExplicitUnignore` so `DEFAULT_IGNORE_LIST` is never consulted below it — measured: all 5 Capacitor chunks leak, on `main` and on this branch alike. Pre-existing #771 precedence, but it means a Capacitor user who unignores the native shell must also add an explicit `_next/` line.
- **`FETCHES` edges stay empty** on such a repo. That was reported alongside this in #3007; the graph-assisted consumer path needs `Route` nodes built from source, which this makes possible but does not itself deliver.

## Implementation notes

**`'public/build'` is deleted, not relocated.** The first draft of this PR moved it into a new `DEFAULT_IGNORED_PATH_FRAGMENTS` list with a matching `shouldIgnorePath` branch. Review established that the mechanism, while correct, was *unreachable*: all four of its match forms (`===`, `startsWith(f + '/')`, `includes('/' + f + '/')`, `endsWith('/' + f)`) put a `/` or an end-of-string on both sides of `build`, so a fragment match strictly implies `build` is a whole path segment — which the per-segment `DEFAULT_IGNORE_LIST` loop already catches one branch earlier. Measured over 768,420 generated paths: **65,506 fragment matches, 0 of them decisive, 0 implication violations.** Deleting the branch left every assertion in this PR green.

So the entry was inert *in that set*, but its paths were never unignored — bare `'build'` covered `public/build/**` on both sides. That makes this a redundant string, not a coverage hole, and #3007 offered exactly this branch: *"real two-segment handling, **or removal**."* Removal wins: it is a smaller diff than the mechanism it replaces, and it leaves nothing dead behind.

**`_next` is now pinned as a segment match.** The original five tests could not distinguish a segment rule from a substring rule — replacing the entry with `normalizedPath.includes('_next')` passed all of them, while silently eating `src/_nextgen/index.ts`. Added negative cases close that.

**The web upload filter had the same gap.** `gitnexus-web/src/lib/upload-filter.ts` `EXCLUDED_DIRS` is the live browser path (`RepoAnalyzer.tsx:402` → `filterRepoFiles` → `POST /api/analyze/upload` → the CLI walker) and listed `.next` but not `_next`, so a #3007-shaped repo uploaded its whole minified tree against the server's 20000-file / 250 MB caps for files the analyzer then discards. (Note there is *also* a `gitnexus-web/src/config/ignore-service.ts` carrying the same gap — it is orphaned dead code with zero importers since #536 retired the in-browser pipeline, verified by module-graph reachability from `src/main.tsx`, so it is left alone here.)

The guard test asserts no member of `DEFAULT_IGNORE_LIST` contains `/`, so the next person who adds `foo/bar` there gets a failing test instead of a silently inert entry. I verified the guard actually fails by reintroducing `'public/build'` into the name set — it reports `expected [ 'public/build' ] to deeply equal []` — rather than trusting that a green test proves anything.

## Testing & verification

New: `gitnexus/test/unit/ignore-build-output.test.ts` (6 tests) — `_next` in both bare and Capacitor/Cordova positions; `_next` as a whole segment vs a substring (`src/_nextgen/index.ts`, `packages/my_next/src/index.ts`, `src/prefix_next.ts`); `public/build` at repo root and nested; the slash-free guard; plus negative cases ensuring ordinary source is untouched (`src/next-steps.ts`, `src/nextConfig/index.ts`, `packages/next-auth/src/index.ts`, `public/favicon-loader.ts`, `src/public/api.ts`).

New: a case in `gitnexus-web/src/lib/upload-filter.test.ts` covering `_next` at bare, Android and iOS positions plus the `_nextgen` negative.

**Both new behaviors were ablated, not just observed green.** Removing `'_next'` from `EXCLUDED_DIRS` fails the web case; swapping the segment rule for `normalizedPath.includes('_next')` fails the substring case. A test that cannot go red pins nothing.

- [x] `cd gitnexus && npm test` — 12776 passed / 7 failed. The 7 (`analyzer-identity`, `assert-publish-grammar-coverage`, `review-agent-workflow`, `wiki-llm-client`) fail identically on pristine `main`; all pre-existing.
- [x] `npx eslint` on changed files — no errors, no new warnings
- [x] `cd gitnexus-web && npm test` — 419 passed / 1 failed (`agent-history.test.ts`, ambient-env credential leak; fails identically with this branch's changes stashed)

**Measured as a diff, not just as a count.** Against a production Capacitor-wrapped Next.js app (1558 JS/TS files on disk), comparing this branch's `shouldIgnorePath` with `main`'s over every file:

| | |
| --- | --- |
| newly ignored | **256** |
| of those, under `src/` | **0** |
| previously ignored, now indexed (regressions) | **0** |

The 256 are all under `_next/`, `android/` or `ios/`.

> **⚠️ This table was measured against `main` at `aac7515d`, before #3043 landed.** Its footnote used to explain three `src/` files as ignored on both sides via the `'env'` entry and the `.d.ts` rule — #3043 has since **deleted both of those rules**, so `src/lib/env/*.js` and `*.d.ts` are now ignored by neither side and the footnote no longer applies. The headline numbers have not been re-measured against the current base because the source repo is not available here; the `_next` behavior itself is covered by the ablated tests above.

## Risk & rollout

- **Requires a re-index to take effect.** `npx gitnexus analyze` re-walks; affected repos will drop the bundle files from their index and lose the bogus `Route` nodes built from them.
- **Node counts will fall** on repos that were indexing build output, and `Route` nodes sourced from bundles disappear. That is the intended correction, but it is a visible change to graph size and to `route_map` output for those repos.
- **This also affects group sync**, which builds its file list through the same `createIgnoreFilter` — a welcome side effect, since bundle output produced no useful contracts.
- **The realistic false positive** is a project that genuinely keeps hand-written source in a directory named `_next` or under `public/build`. Both are conventional build-output locations. On the **CLI/analyze path** a `.gitnexusignore` negation recovers such a tree — use the bare `!_next/` form, which works at any depth; `!_next/**` *alone* is a silent no-op, because `childrenIgnored` prunes the directory before any descendant pattern is tested. On the **browser upload path** there is no escape hatch at all: `EXCLUDED_DIRS` is a client-side pre-filter that reads no repository ignore rules, so a negation cannot recover what it drops.
- **Rollback:** revert the commit; the next `analyze` restores the previous behavior.
- Release notes: `fix` → 🐛 Bug Fixes.

## Checklist

- [x] No `AGENTS.md` / overlay changes in this PR
- [x] No new `GITNEXUS_*` environment variables

---

## Summary

*A broad repository ignore and upload-filter change with critical graph reach, centered on build-output handling and accompanied by targeted tests. The file-level risk is low, but the affected configuration sits on paths used across the codebase.*

**🔴 CRITICAL blast radius. This appears to be an ingestion and ignore-filter change spanning `gitnexus/src/config/ignore-service.ts` and the web upload filter, with broad downstream reach.**

The main review focus is `gitnexus/src/config/ignore-service.ts`, where `DEFAULT_IGNORE_LIST`, `createIgnoreFilter`, `loadIgnoreRules`, `shouldIgnorePath`, `isRootArtifactDirectory`, and related ignore helpers changed. This area has the strongest architectural concentration in `Extractors`, with additional impact across `Config`, `Wiki`, `Group`, `Typescript`, `Import-resolvers`, and `Storage`.

The web side changes `FilterResult`, `EXCLUDED_DIRS`, and `MAX_FILE_BYTES` in `gitnexus-web/src/lib/upload-filter.ts`. Review the interaction between these filters and the repository ignore rules first, then verify the coverage in `gitnexus/test/unit/ignore-build-output.test.ts`, `gitnexus/test/unit/ignore-service.test.ts`, `gitnexus/test/helpers/ignore-set-source.ts`, `gitnexus-web/src/lib/upload-filter.test.ts`, and `gitnexus-web/src/lib/upload-filter-ignore-drift.test.ts`. The computed file risk is LOW, with no HIGH or CRITICAL risk files.

_Added by GitNexus for PR #3018. Edit freely — this block is replaced on the next review, everything above it is left untouched._


gitnexus/src/config/ignore-service.ts
```diff
@@ -106,6 +112,31 @@ const DEFAULT_IGNORE_LIST = new Set([
...
 ]);
 
+/**
+ * Multi-segment paths to ignore, matched against the whole POSIX path.
+ *
+ * These CANNOT live in {@link DEFAULT_IGNORE_LIST}: that set is tested one path
+ * SEGMENT at a time (`parts.some(p => DEFAULT_IGNORE_LIST.has(p))`) and is also
+ * exposed through `isHardcodedIgnoredDirectory(name)`, which receives a bare
+ * directory name. A slash-containing member can never equal a single segment,
+ * so `'public/build'` sat in that set matching nothing at all until #3007. A
```
magyargergo: **P2 — this comment is true about the set *entry*, but not about the behavior.** `[reproduced]`

`'build'` is a bare member of `DEFAULT_IGNORE_LIST` (`ignore-service.ts:50`, untouched by this PR), and the per-segment loop at `:334-338` runs *before* the new fragment branch at `:341`. All four of `hasIgnoredPathFragment`'s conditions require `build` at a full segment boundary — so a fragment match strictly implies the segment loop already returned `true`. `public/build/**` has therefore always been pruned, on base and on head.

Measured three ways:
- Instrumented sweep of 144 paths around `public/build`: `{ total: 144, fragTrue: 48, fragDecisive: 0 }` — never the deciding branch.
- Delete `:340-343` and your four `public/build` assertions still all pass.
- Run the new test file against base source: the `public/build` test is green.

So the entry was inert *there*, but the paths were never unignored. Two ways out, both fine:

1. **Simplest** — drop `DEFAULT_IGNORED_PATH_FRAGMENTS`, `hasIgnoredPathFragment` and the `:340-343` branch, just delete the inert `'public/build'` string, keep the slash-free guard. That is the whole of the real second defect: a redundant string, not a coverage hole. It is also what #3007 itself offered — *"real two-segment handling, **or removal**."*
2. **Keep the mechanism** as forward-looking infrastructure — then reword this sentence (e.g. "…was inert *there*; the paths were already covered by the bare `'build'` entry"), wire it into `childrenIgnored` (see the comment four lines down), and give it a load-bearing member.

Not a merge blocker. Flagging it because the comment and the PR title both read as though a class of output was being indexed and now isn't.

gitnexus/src/config/ignore-service.ts
```diff
@@ -106,6 +112,31 @@ const DEFAULT_IGNORE_LIST = new Set([
...
+ * guard test pins the invariant that the name set stays slash-free.
+ */
+const DEFAULT_IGNORED_PATH_FRAGMENTS: readonly string[] = [
+  'public/build', // Remix / Laravel Mix compiled asset output
```
magyargergo: **P2 — the fragment list is only half-wired: `childrenIgnored` never consults it.** `[reproduced]`

`createIgnoreFilter.childrenIgnored` prunes directories with `DEFAULT_IGNORE_LIST.has(p.name)` at `:536` and never calls `hasIgnoredPathFragment`. Today that is masked, because `build` happens to be independently bare-listed.

Add any fragment whose leaf segment isn't — say `public/hot` — and the three entry points disagree: `shouldIgnorePath` returns `true`, `childrenIgnored` returns **false**, `isHardcodedIgnoredDirectory('hot')` returns **false**. Glob then descends the entire emitted tree, and the tsconfig/workspace scanners burn their `SCAN_MAX_DIRS = 20_000` caps (`tsconfig.ts:59`, `node-workspace-packages.ts:62`) walking it.

One line:

```ts
if (hasIgnoredPathFragment(rel)) return true;
```

in `childrenIgnored` right after the name-set check at `:536` — `rel` is already in scope and POSIX-normalized, and the negation escape hatch at `:534` has already run, so precedence is preserved. Worth pairing with the inverse guard to your existing one: every `DEFAULT_IGNORED_PATH_FRAGMENTS` entry *must* contain a `/`.

(Single-engine finding — five Claude lanes; neither Codex leg raised it.)

gitnexus/src/config/ignore-service.ts
```diff
@@ -54,13 +54,19 @@ const DEFAULT_IGNORE_LIST = new Set([
   'obj',
   'target', // Java/Rust
   '.next',
+  // `.next` is Next.js's build CACHE; `_next` is the EMITTED output, and the two
+  // are different directories. A Capacitor/Cordova shell copies the emitted
+  // bundle to `<platform>/app/src/main/assets/public/_next/static/…`, where none
+  // of the path segments hit this list — so a mobile-wrapped Next.js app had 40%
+  // of its indexed files come from minified chunks, and every Route node it
+  // produced pointed at a webpack bundle rather than source (#3007).
+  '_next',
```
magyargergo, line 73: **P2 — `_next` lands in the *unanchored* bare-name set. This is where the two engines disagreed, so here it is undisguised.** `[reproduced]`

- **Codex** raised it in both legs that reviewed code: `ce` at P2/confidence 75 — *"a repo with source under `src/_next/`, `packages/_next/` … will silently lose that subtree"* — and its adversarial pass at `[high]`, "needs-attention / no-ship".
- **Claude's correctness lane tried to substantiate it and could not**: `_next` is Next.js's reserved *emitted* path, not a source convention in any ecosystem it checked. Logged as residual risk, not a finding.
- **Claude's adversarial lane confirmed the mechanism, priced likelihood low** (P3, confidence 60): with a hand-written `_next` subtree the real walker indexed 3 files and **7 vanished silently**, including `_next/package.json`, `_next/tsconfig.json` and `_next/go.mod` — so such a subtree loses not just source but the manifests that let the rest of the repo resolve imports into it.

**Reconciled: mechanism confirmed, trigger hypothetical.** Codex's trigger (`packages/_next/src/index.ts`) is a constructed repo, not an observed one. And "no-ship" was the most aggressive of Codex's three verdicts, not its consensus — its `ce` leg closed *"Ready with fixes"* and its 7-persona swarm leg never raised the concern at all, closing *"production-ready with minor follow-ups … no P0-P2 production blockers."*

Worth care anyway, because this file already carries the precedent: `'packages'` was deliberately **removed** from this set at `:31` for exactly this class of reason.

**One caveat on the obvious remedy, because we checked it and it does not work as stated.** Anchoring `_next` into `DEFAULT_IGNORED_PATH_FRAGMENTS` (`public/_next`, `out/_next`, `www/_next`, `assets/public/_next`) covers your two Capacitor paths and leaves `.next` alone — but it does **not** match `_next/static/chunks/main.js`, the root-level case at `test/unit/ignore-build-output.test.ts:19`, which has no web-root ancestor. Since that assertion is `false` on base, the bare `'_next'` entry is the only thing making it pass: anchoring without also adding a root-anchored form turns your own headline regression test red. If you go that route it needs both, plus the `childrenIgnored` wiring above so directory pruning isn't lost.

The current form is defensible and matches how `.next`, `.nuxt`, `.turbo` are already handled. Your call — just recording the disagreement honestly rather than resolving it for you.

gitnexus/test/unit/ignore-build-output.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    );
+  });
+
+  it('does not ignore ordinary source that merely mentions next', () => {
+    expect(shouldIgnorePath('src/next-steps.ts')).toBe(false);
+    expect(shouldIgnorePath('src/nextConfig/index.ts')).toBe(false);
+    expect(shouldIgnorePath('packages/next-auth/src/index.ts')).toBe(false);
+  });
+
+  it('ignores public/build, which never matched while it sat in the name set', () => {
```
magyargergo: **P2 — this test is green on base, so it pins nothing this PR changed.** `[reproduced]`

The testing lane restored the base `ignore-service.ts` into a scratch copy and ran your unmodified file under real vitest: `Tests 2 failed | 3 passed (5)`. This is one of the three that pass — because of the `'build'` segment, per the comment above. Its own comment (*"this entry matched nothing at all before #3007"*) asserts the same premise.

Worth contrasting with your `_next` test and the slash-free guard, which **do** fail on base — those two are real regression tests and are the reason this PR's core claim is credible.

Suggest renaming this one to what it actually pins (e.g. `keeps public/build ignored`) and dropping the comment, or folding it into whichever route you take on the fragment machinery.

magyargergo (commented):
Thanks for this — the `_next` half is a real fix, and it is the rare kind that arrives with its own proof. Two engines and ten review lanes went at it, several specifically trying to break it, and none could find a path without a `_next` segment whose verdict changed, or anything that base ignored becoming indexed. Most of what follows is about the *second* half (`public/build`) and about how the `_next` entry is scoped — not about whether the fix works.

## How this was reviewed

| Engine | Legs | Status |
|---|---|---|
| **A — Claude** | GitNexus swarm (`test-ci-verifier`) + `ce-code-review` personas: correctness, testing, maintainability, standards, adversarial, performance | 7 of 8 lanes live |
| **B — Codex (gpt-5.5/xhigh)** | Same GitNexus swarm, Solo mode (six of seven personas incl. risk-architect; the skipped one is the synthesis critic); same `ce-code-review` scaffolding (report-only); a native adversarial red-team pass | all 3 legs live |

Both engines ran the **same** swarm + ce scaffolding, so "A and B agree" is a clean same-prompt A/B on two models — not ten independent confirmations. Findings are weighted accordingly.

**Coverage gap, stated plainly:** the Claude-side risk-architect lane failed to run and produced nothing. Engine B ran that persona in Solo mode and the Claude adversarial lane covers overlapping ground, so the domain is not unreviewed — but any risk-model finding below rests on one engine plus the adversarial lane rather than two.

---

## What holds up

**The `_next` fix is correct and leak-free, measured rather than asserted.**

- The correctness lane diffed base vs. head `shouldIgnorePath` over **35,923 paths** (4,983 real files from `git ls-files` + 30,940 synthetic paths at depths 1–4): `total base != head: 4994`, `differences NOT explained by a '_next' segment: 0`, `head=false but base=true (new leaks): 0`. Every one of the 4,994 differences is a `_next` path being newly ignored — which is the fix doing its job — and **no path that base ignored became indexed**. No path *without* a `_next` segment changed verdict in either direction. `[reproduced]`
- End-to-end through the real walker over a synthetic Capacitor-shaped repo: all 4 emitted `_next` chunks — root, `public/_next`, `ios/App/App/public/_next`, `android/app/src/main/assets/public/_next` — are excluded, and the 4 real source files (`src/app/page.tsx`, `README.md`, `AppDelegate.swift`, `MainActivity.java`) are kept. `[reproduced]`
- The fix reaches the path that actually produces the bogus `Route` nodes #3007 is about: `http-route-extractor.scanFiles` builds its own file list through `createIgnoreFilter` (`http-route-extractor.ts:679`), and the main walk uses it at `filesystem-walker.ts:47`. `[code-read, both engines]`
- The `.next` / `_next` distinction in your comment at `ignore-service.ts:57-62` is exactly right, and the negative cases you added (`src/next-steps.ts`, `src/nextConfig/index.ts`, `packages/next-auth/src/index.ts`) are the right guardrails.
- **Your `_next` tests are real regression tests.** The testing lane restored the base `ignore-service.ts` into a scratch copy and ran your unmodified test file under real vitest: `Tests 2 failed | 3 passed (5)` — the `_next` case and the slash-free guard both fail on base and pass on head. `[reproduced]`
- Both ignore-service unit files green on the worktree: `Test Files 2 passed (2)` / `Tests 154 passed (154)`, and CI really does pick the new file up (`vitest.config.ts:137` includes `test/**/*.test.ts`; `ci-tests.yml` runs `npx vitest run --shard`). `[reproduced]`
- Clean on standards, security and hygiene: PR title matches `CONTRIBUTING.md`'s conventional format, the body follows the template section-for-section, no `any`, no `if`-branching in the new tests, no CHANGELOG churn, one commit, two files, `git diff --check` clean, hidden-bidi grep clean. The standards lane filed **zero** findings; Codex's security and branch-hygiene lanes filed zero.
- The guard test's source parser is sounder than it looks: it survives Prettier at `printWidth: 100` (verified by running the formatter), and the `//`-prefixed prose — including the apostrophe in `Next.js's` — is filtered out *before* the quote-slicing step. Good catch, and well called out in the comment.

---

## Headline

### 1. The `public/build` half changes nothing, and the comment + test say otherwise

**Seven lanes agreeing — five Claude lanes and two of Codex's swarm personas — across three independent methods.**
`ignore-service.ts:122`, `:125-138`, `:340-343`

`'build'` is still a bare member of `DEFAULT_IGNORE_LIST` (`ignore-service.ts:50`, untouched here), and the per-segment loop at `:334-338` runs *before* the new fragment branch at `:341`. All four of `hasIgnoredPathFragment`'s conditions require `build` at a full segment boundary — so `hasIgnoredPathFragment(p) === true` strictly implies the segment loop already returned `true`. The new branch is unreachable-in-effect for its only member.

Measured three ways:

- Instrumented counter over a 144-path sweep around `public/build`: `{ total: 144, fragTrue: 48, fragDecisive: 0 }` — never the deciding branch. `[reproduced]`
- Mutation: delete `:340-343` and run your four `public/build` assertions verbatim — **all four still pass.** `[reproduced]`
- Vitest against reverted base source: the `ignores public/build…` test is **green on base**. `[reproduced]`

So the comment at `:122` — *"`'public/build'` sat in that set matching nothing at all"* — is true about the **set entry** and false about the **behavior**: `public/build/**` has always been pruned by the bare `'build'` segment, on base and on head. The title ("restore the dead `public/build` entry") reads as though a class of output was being indexed and now isn't. Nothing was, and nothing changes.

Two ways out:

- **Simplest:** drop `DEFAULT_IGNORED_PATH_FRAGMENTS`, `hasIgnoredPathFragment` and the branch, just delete the inert `'public/build'` string, and keep the slash-free guard. That is the whole of the real second defect — a redundant string, not a coverage hole. It is also what #3007 itself suggested: *"real two-segment handling, **or removal**."*
- **Or keep the mechanism** as forward-looking infrastructure — but then say so, reword `:122`, wire it into `childrenIgnored` (finding 2), and give it a load-bearing member.

Not a merge blocker either way. Severity spread across lanes was P1–P3; the lane that actually measured behavior called it P3, and we've settled on P2. It is a "the code says something that isn't so" problem, and the next maintainer will hit it.

### 2. The fragment list is only half-wired — `childrenIgnored` never consults it

`ignore-service.ts:126` — single-engine (five Claude lanes), reproduced.

`createIgnoreFilter.childrenIgnored` prunes directories using `DEFAULT_IGNORE_LIST.has(p.name)` at `:536` and never calls `hasIgnoredPathFragment`. Today that is masked, because `build` happens to be independently bare-listed. Add any fragment whose leaf segment isn't — say `public/hot` — and `shouldIgnorePath` returns `true` while `childrenIgnored` returns **false** and `isHardcodedIgnoredDirectory('hot')` returns **false**: glob descends the entire emitted tree, and the tsconfig/workspace scanners burn their `SCAN_MAX_DIRS = 20_000` caps (`tsconfig.ts:59`, `node-workspace-packages.ts:62`) walking it. `[reproduced]`

One line fixes it: `if (hasIgnoredPathFragment(rel)) return true;` in `childrenIgnored` right after the name-set check at `:536` — `rel` is already in scope and POSIX-normalized, and the negation escape hatch at `:534` has already run, so precedence is preserved. Worth pairing with the inverse guard: every fragment entry *must* contain a `/`.

### 3. `_next` went into the unanchored bare-name set — where the two engines disagree

`ignore-service.ts:63`. The disagreement is the interesting part, so here it is undisguised.

- **Engine B raised it in both legs that reviewed code.** `codex-ce`: P2, confidence 75 — *"a repo with source under `src/_next/`, `packages/_next/` … will silently lose that subtree"*. `codex-adversarial`: `[high]`, **needs-attention / no-ship**.
- **Engine A's correctness lane tried to substantiate it and could not:** `_next` is Next.js's *reserved emitted* path, not a source convention in any ecosystem it checked. Logged as residual risk, not a finding.
- **Engine A's adversarial lane confirmed the mechanism but priced likelihood low** (P3, confidence 60): with a hand-written `_next` subtree the real walker indexed 3 files and **7 vanished silently**, including `_next/package.json`, `_next/tsconfig.json` and `_next/go.mod` — so such a subtree loses not just source but the manifests that let the rest of the repo resolve imports into it. `[reproduced]`

**Reconciled: mechanism 100% confirmed, trigger hypothetical.** Codex's concrete trigger (`packages/_next/src/index.ts`) is a constructed repo, not an observed one; no lane found a real project keeping source under `_next`. Note this is Engine B's own overall position too — `codex-ce` closed **"Ready with fixes"** and Codex's 7-persona swarm leg never raised the concern at all, closing *"production-ready with minor follow-ups … no P0-P2 production blockers."* So "no-ship" reads as the single most aggressive of Codex's three verdicts rather than its consensus.

Still, the file's own history argues for care: `'packages'` was deliberately **removed** from this set with an in-file comment at `:31` for exactly this class of reason.

**One caveat on the obvious remedy, because we checked it and it does not work as stated.** Moving `_next` into anchored fragments (`public/_next`, `out/_next`, `www/_next`, `assets/public/_next`) covers your two Capacitor paths and leaves `.next` alone — but it does **not** match `_next/static/chunks/main.js`, the root-level case at `ignore-build-output.test.ts:19`, which has no web-root ancestor. Since the testing lane proved that assertion is `false` on base, the bare `'_next'` entry is the only thing making it pass: anchoring without also adding a root-anchored form would turn your own headline regression test red. If you go this route it needs both, plus finding 2 wired first so directory pruning isn't lost. The current form is defensible and matches how `.next`, `.nuxt`, `.turbo` are already handled — your call.

### 4. The `public/build` test is green on base

`ignore-build-output.test.ts:37-42` — reproduced on both engines.

Run the file against base source and the `ignores public/build, which never matched while it sat in the name set` test is green, because of finding 1. Its comment asserts the same false premise. Suggest renaming it to what it actually pins (e.g. "keeps `public/build` ignored") and dropping the comment.

---

## The escape hatch: reconciling two lanes that looked like they contradicted each other

The correctness lane reported that the `.gitnexusignore` negation escape hatch works for both the new name-set entry and the fragment. The adversarial lane reported that `!_next/**` and friends silently no-op. **Both are correct** — the first tested only the trailing-slash directory form, which is one of the spellings that works. The full matrix, re-run independently:

```
(none)              _next recovered 0/4
!_next              _next recovered 4/4        works
!_next/             _next recovered 4/4        works   <- the spelling the first lane tested
!**/_next/          _next recovered 4/4        works
!/_next/            _next recovered 1/4        root-anchored only (correct gitignore semantics)
!_next/*            _next recovered 0/4        SILENT NO-OP
!_next/**           _next recovered 0/4        SILENT NO-OP
!**/_next/**        _next recovered 0/4        SILENT NO-OP
!public/build/      public/build recovered 1/1  works
!public/build/**    public/build recovered 0/1  SILENT NO-OP
```

**Precise truth:** the *directory-form* negation works (`!dir`, `!dir/`, `!**/dir/`); the `/**` and `/*` suffixed forms are silent no-ops. Mechanism: `childrenIgnored` prunes the directory before any file under it reaches `ignored()`, and `hasExplicitUnignore` (`:467-481`) tests only `rel`, `rel + '/'` and ancestors — `!_next/**` matches none of those, since it requires at least one component *below* `_next`. So `DEFAULT_IGNORE_LIST.has('_next')` at `:536` wins and the subtree is never descended.

This is **pre-existing behavior of `childrenIgnored`, not something this PR introduces** — it has applied to every member of the set since #771. It becomes newly relevant only because `_next` is a new member. Worth noting because the `/**` form is the most idiomatic gitignore spelling for "un-ignore everything under here" — it is what the repo's own #596 test writes at `ignore-service.test.ts:543`, and that test is fine precisely because it pairs `!iOS/**` with the working `!iOS/`. `README.md` documents `!pattern` once (`:628`) and never mentions that a hardcoded-list override exists at all; a line naming the working spelling would close the loop.

## Related, and honestly labelled: the ancestor-negation interaction

The sharpest thing either engine found, reproduced independently before writing this. Real walker, synthetic Capacitor repo:

```
.gitignore "ios/ android/"  +  .gitnexusignore "!ios/ !android/"
   -> indexed=6, _next chunks LEAKED=2
.gitnexusignore "!public/"     -> 3 _next chunks LEAKED
.gitignore      "!ios/" alone  -> 1 _next chunk  LEAKED
```

`hasExplicitUnignore` walks **every ancestor segment** (`:475-479`), so `!ios/` short-circuits `ignored()` at `:516` and `shouldIgnorePath` — and therefore the new `'_next'` entry — is never reached for anything under `ios/`. Note the third case: the negation lived in **`.gitignore`**, not `.gitnexusignore`; `loadIgnoreRules` merges both into one `ignore` instance (`:429-442`), so a years-old `.gitignore` line written with no knowledge of GitNexus disables the hardcoded list for its whole subtree.

**Is this a defect in this PR? No.** It is the deliberate #771 precedence rule — user negation beats the hardcoded list — documented in the `createIgnoreFilter` docstring at `:491-498` and pinned by existing tests. Every member of `DEFAULT_IGNORE_LIST` has always behaved this way; `_next` simply inherits it. **What it does mean** is that the fix's effectiveness is conditional, and the condition is most likely to fail for exactly the population #3007 describes: a Capacitor app that gitignores `ios/`+`android/` and then negates them to get the native shell indexed. Nothing surfaces when this has happened, so a silently-defeated fix looks identical to a working one. A follow-up issue seems right, rather than a change here.

---

## Lower priority

5. **The guard test covers one syntactic form.** Nine mutants through its verbatim parsing logic: it stays green on a template-literal entry, a `.add()` after the literal, a backslash entry, and — most realistically — **the identical dead-entry hazard in `IGNORED_FILES`** (`:273-307`), likewise tested as a bare basename, where a slash-bearing entry would be exactly as dead. It is also brittle in a way worth knowing: adding a behaviorally-inert `ReadonlySet<string>` type annotation to the declaration makes its literal source-string match return `-1` and the test fails with `expected 0 to be greater than 20`, which points at the ignore list rather than at the test's own text match. Exporting `DEFAULT_IGNORE_LIST` (or a test-only accessor) and asserting `[...DEFAULT_IGNORE_LIST].every(e => !e.includes('/'))` against the live value is immune to all of these and covers `IGNORED_FILES` too. (The double-quoted mutant was *refuted* as a durable bypass — Prettier rewrites it and the guard catches it.) P3, `[reproduced]`, Engine A only.
6. **Three template-literal allocations per call** in `hasIgnoredPathFragment` (`:134-136`), on the walker hot path. Measured in isolation: `184.9 ns/call` as shipped vs `61.3 ns/call` precomputed — ~3x; end-to-end over a 100k-path corpus, `+148-192 ns/call` on `shouldIgnorePath` (~15-19 ms per 100k-file walk). Small, and the reviewing lane's own verdict was "not a P0/P1 performance problem" — but precomputing the three derived strings at module scope is free and scales as the list grows. P3, `[reproduced]`.
7. **The fix is Next.js-shaped, but #3007's mechanism isn't.** Capacitor copies `webDir`'s *contents* into `assets/public/`, so for Vite (`dist`), CRA (`build`) and Angular the identifying segment doesn't survive the copy: `android/app/src/main/assets/public/assets/index-DiwrgTda.js` → `shouldIgnorePath = false`. A nested `android/.gitignore` can't save it either — `loadIgnoreRules` reads root-level files only (`:433`). Generalizing to the copy *destination* is the natural follow-up, and a good first real member for the fragment list. Genuinely out of scope here. `[reproduced]`, Engine A only.
8. **`shouldIgnorePath`'s direct callers don't honor the negation escape hatch.** A user who recovers `_next` with the working `!_next/` gets File and Symbol nodes, but `wiki/generator.ts:349` still filters those files out (measured: indexer admitted 4, wiki kept 1), the six `*-workspace-extractor.ts` files skip the directory for workspace-link discovery, and `tsconfig.ts:294` / `node-workspace-packages.ts:364` never load its manifests. Pre-existing, widened slightly here; tracking-issue material. P3, `[reproduced]`.
9. **Two known, documented divergences not mirrored** — neither a regression: `thrift-extractor.ts:225,302` keeps a hand-rolled ignore glob whose own `TODO(#1156-followup)` already states new `DEFAULT_IGNORE_LIST` entries are never mirrored there; and `gitnexus-web/src/config/ignore-service.ts` still has `.next` without `_next` (`:47`) and the inert `'public/build'` (`:54`) — but it has **zero importers** anywhere in the repo, so your "web untouched" is accurate. `[reproduced]`

## Probed and refuted

Validation is a feature, so here is what the lanes tried to break and couldn't — nearly all executed rather than reasoned about:

- **The four match forms in `hasIgnoredPathFragment` are correct.** No false positive on `mypublic/build/x.js`, `xpublic/build`, `public/buildings/x.js`, `public/build-output/x.js`, `public/build.js`; correct positives on `./public/build/x.js`, `/abs/repo/public/build/entry.js`, `public/build/`, `public\build\x.js`, `apps/web/public/build/manifest.js`. Exhaustive 11,110-path sweep, zero anomalies.
- **Windows backslashes are handled** — `normalizedPath = filePath.replace(/\\/g, '/')` at `:318` runs before the fragment check; verified live.
- **Case sensitivity is pre-existing and doesn't bite.** `Build/`, `_Next/` already missed on base; lowercasing the segment loop would be *worse* (it would newly swallow `Build/`, `ENV/`, `Target/`, `Bin/`, `Obj/`).
- **A double-quoted `"public/build"` entry does not durably defeat the guard test** — Prettier (`singleQuote: true`, enforced at `ci-quality.yml:23`) rewrites it and the guard turns red.
- **The guard test's block slice and its `> 20` floor are not fragile in the ways they look** — it parses 78 entries on HEAD, the `]);` terminator is unambiguous, and Prettier keeps one entry per line at `printWidth: 100`.
- **A checkout located under a directory named `_next` / `build` / `target` does NOT zero the index** — `createIgnoreFilter` operates on `p.relative()` (`:507`, `:527`), so the checkout directory's own name never enters the tested path.
- **Nested `.gitignore` files cannot accidentally un-ignore the emitted copy** — only root-level ignore files are read.
- **No absolute-path hazard from the new entry** — all `shouldIgnorePath` caller families pass repo-relative paths.
- **Reordering the fragment check within `shouldIgnorePath` buys nothing** — measured, null result in both directions, within noise.
- **The standards audit found no violation at all** — placement under `test/unit/` matches TESTING.md, the file-per-concern split matches existing precedent, no doc surface enumerates `DEFAULT_IGNORE_LIST` for the PR to have missed, and the `src/core/ingestion` "no language names" rule doesn't reach `src/config/`.

## Test gaps worth closing

- No test exercises `createIgnoreFilter` for `_next` **at all** — neither `childrenIgnored`'s subtree prune (the behavior the PR is actually selling: "40% of indexed files") nor the negation override. The natural home is the existing `.gitnexusignore negation` describe block in `ignore-service.test.ts`, which already owns the temp-dir + `createIgnoreFilter` harness.
- No test can fail if `hasIgnoredPathFragment` is deleted (proven by mutation). Real coverage needs a fragment whose leaf segment isn't already bare-listed.
- No test pins the inverse invariant — every `DEFAULT_IGNORED_PATH_FRAGMENTS` entry *must* contain a `/`.
- `isHardcodedIgnoredDirectory('_next')` is untested (it does return `true` — verified live).
- No near-miss negatives for the matcher (`public/buildings/x.js`, `mypublic/build/x.js` asserted against the matcher directly), which are the only assertions that would catch someone "simplifying" it to a bare `.includes(fragment)`.

## CI

Green on PR head: all `tests /` shards, `quality /`, CodeQL, gitleaks. The only red is Vercel ("Authorization required to deploy"), unrelated to this change. `mergeStateStatus: BEHIND` — a rebase on `main` is needed before merge.

---

*Automated multi-tool review digest (Claude + Codex gpt-5.5/xhigh, ten lanes). Findings were cross-checked and several re-run independently before writing, but please verify anything here against the code before acting on it — and push back on anything that looks wrong.*

gitnexus/test/unit/ignore-build-output.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    // A slash-containing member of DEFAULT_IGNORE_LIST can never equal a single
+    // path segment, so this entry matched nothing at all before #3007.
+    expect(shouldIgnorePath('public/build/entry.client.js')).toBe(true);
```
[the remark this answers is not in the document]
magyargergo: **Confirmed and fixed** — you were right, and the ablation was the proof rather than the reasoning.

Removing `hasIgnoredPathFragment` and re-running this file: all 5 tests still green. Removing the const *and* the function *and* the branch (grep confirms 0 remaining references): still 5/5, and 166/166 across this file plus `ignore-service.test.ts`. So no assertion in the PR failed when the feature it added was deleted.

Your suggested probe (`public/assets/entry.client.js`) does not rescue the test either, because the mechanism is stronger than "these two paths happen to contain `build`": *every* path the fragment matcher can match contains `build` as a whole segment. All four match forms — `===`, `startsWith(f + '/')`, `includes('/' + f + '/')`, `endsWith('/' + f)` — put a `/` or an end-of-string on both sides of it. Fuzzed 768,420 paths to check: `{fragTrue: 65506, decisive: 0, implicationViolations: 0}`. Control: delete bare `'build'` and `decisive` jumps to 39, so the branch was correct code, just fully shadowed.

Resolved by deleting the machinery rather than by strengthening the test — see the reply on `ignore-service.ts`. This test is renamed to `keeps public/build ignored after the inert name-set entry was removed`, and its comment now says what it actually pins: that the deletion changed no behavior. Green on both sides by design.

Separately, your instinct that this file's assertions were weak was right about a *different* rule. None of the negative cases contained `_next` as a substring of a longer segment, so swapping the entry for `normalizedPath.includes('_next')` passed all 5 tests while eating `src/_nextgen/index.ts`. Added `matches _next as a whole segment, not as a substring` to close that.

gitnexus/src/config/ignore-service.ts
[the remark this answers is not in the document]
magyargergo: **Correct, and fixed by deletion.** Marked a nit, but it was the most load-bearing finding on the PR — the whole second half of the change was unreachable.

Verified rather than reasoned. Post-merge line numbers: the per-segment `DEFAULT_IGNORE_LIST` loop at `:350-354`, the fragment branch at `:357-359` — the loop still runs first (#3043 inserted its `isRootArtifactDirectory` branch *above* both). `'build'` is still bare-listed at `:54`.

The implication is total, not incidental: all four match forms put a `/` or an end-of-string on both sides of `build`, so fragment-true always implies `build` is a whole segment. Fuzz over 768,420 paths: `{fragTrue: 65506, decisive: 0, implicationViolations: 0}` — no decisive path exists. Deleting the const, the function and the branch leaves 5/5 PR tests and 166/166 ignore tests green.

`DEFAULT_IGNORED_PATH_FRAGMENTS`, `hasIgnoredPathFragment` and the branch are gone. `'public/build'` is deleted outright rather than relocated — #3007 offered "real two-segment handling, **or** removal", and removal is right because the entry was inert *in the set* but its paths were never unignored (bare `'build'` covered them on both sides). The slash-free guard test stays; it is what stops the next slash-bearing entry from dying the same way.

One correction to your wording, since it matters for anyone reading this later: *"the added test/comment incorrectly describe this as a previously unmatched case"* — the **entry** genuinely was unmatched, which is what the comment claimed. What was wrong was the implied conclusion that some class of path was therefore being indexed. It wasn't.

gitnexus/src/config/ignore-service.ts
magyargergo: **Took option 1 — deleted.** Re-verified against the merged tree first, because #3043 rewrote much of this file underneath the PR.

Post-merge ordering is unchanged in the way that matters: segment loop `:350-354`, fragment branch `:357-359`; #3043 only inserted *above* both. `'build'` still bare at `:54`.

Strengthened the measurement from a 144-path sweep to a proof. Fuzzed 768,420 paths (4-deep recombination of adversarial segments, 400k char-level fuzz embedding `public/build` at random offsets, 200k random soup):

```
{"checked":768420,"fragTrue":65506,"decisive":0,"implicationViolations":0}
```

`implicationViolations: 0` is the real result — it says fragment-true *always* implies `build` is a whole segment, so no decisive path can exist, rather than merely that none was sampled. Control: remove bare `'build'` from `:54` and the sweep flips to `fragDecisive: 39`.

Gone: `DEFAULT_IGNORED_PATH_FRAGMENTS`, `hasIgnoredPathFragment`, the `:357-359` branch, and the `'public/build'` string. The name-set comment now records *why* it was deleted rather than relocated, so nobody re-adds it. Guard test kept. Diff is now shorter than it was before this round.

That also disposes of the `childrenIgnored` finding below — with no fragment list there is nothing left to be asymmetric about.

gitnexus/src/config/ignore-service.ts
magyargergo: **Moot now — the fragment list is deleted**, so there is nothing for `childrenIgnored` to be asymmetric about. Recording what re-verification turned up anyway, because two details in my own comment were wrong and the second one would have shipped a half-fix.

**The call site moved.** #3043 rewired `childrenIgnored` from `DEFAULT_IGNORE_LIST.has(p.name)` to `isHardcodedIgnoredDirectoryAtPath(repoPath, ...)` (`:564`). The one-liner I suggested was still placeable — `rel` is in scope and POSIX-normalized at `:555`, negation escape hatch still runs first at `:562` — and it does work: reproduced the disagreement on a real fixture with `'public/hot'` added to the list, and the fix made `childrenIgnored` agree.

**But `childrenIgnored` was the wrong place to put it.** `isHardcodedIgnoredDirectoryAtPath` has four callers, not one — `childrenIgnored` plus `python-workspace-extractor.ts:170`, `typescript/tsconfig.ts:295` and `node-workspace-packages.ts:365`. The latter three have *no* leaf-level fallback, so my one-liner would have fixed one caller and left three still descending. The correct home was inside `isHardcodedIgnoredDirectoryAtPath` at `:405`, where `relative` is already computed.

**And the consequence was smaller than I implied.** Measured on a real fixture: `childrenIgnored` and `shouldIgnorePath` do disagree, but the walk output is still *correct* — `ignored()` falls through to `shouldIgnorePath` and filters each file at the leaf. So it was a wasted subtree descent, a perf and consistency problem, not the wrong-result problem "the three entry points disagree" reads as. The `SCAN_MAX_DIRS` concern was real for the three extractor callers.

Both points are now academic for this PR, but worth having on the record if a genuine multi-segment ignore is ever added — that change needs to go in `isHardcodedIgnoredDirectoryAtPath`, not in `childrenIgnored`.

gitnexus/src/config/ignore-service.ts
magyargergo: **Kept `_next` bare. Documented as accepted residual risk, with a measured remedy on the shelf.** Re-ran the whole thing against the merged tree, since #3043 rewrote `childrenIgnored`.

**Downside quantified.** Fixture with hand-written source under `packages/_next/`: 11 files on disk, **6 returned, 5 vanished** — `package.json`, `tsconfig.json`, `README.md`, `src/index.ts`, `src/router.ts`. Root-level `_next/`: 7 on disk, 3 returned, 4 vanished. So the "loses the manifests too" claim holds.

**#3043 changed nothing here.** Ran the full negation matrix pre- and post-merge and diffed: **identical, 0 of 20 cells changed.** It swapped in `isHardcodedIgnoredDirectoryAtPath`, but that still returns true for `_next` at `:403`, and the `hasExplicitUnignore` short-circuit at `:562` runs first either way.

**One softening of my own earlier claim.** I said `!_next/**` being a silent no-op was the headline because it is what the repo's #596 test writes. Re-checking `test/unit/ignore-service.test.ts:585` — that test writes the **pair** `!iOS/` + `!iOS/**`, and the pair recovers 4/4. So the accurate statement is narrower: `!_next/**` *alone* is a silent no-op; the documented idiom works. Still worth its own issue, but it is not a broken escape hatch.

**On the remedy — the caveat I raised was right, and there is a version that clears it.** Parent-anchored fragments do fail this PR's own test at `:19` (`AssertionError: expected false to be true`), confirmed. But child-anchored fragments (`_next/static`, `_next/server`, `_next/data`, `_next/image`) with `_next` out of the name set, *plus* an `existsSync` probe in `isHardcodedIgnoredDirectoryAtPath`, gives: **5/5 tests green, 7/7 and 4/4 source files recovered, every chunk still dropped, and byte-identical behavior on the emitted-output fixture** (kept 3, prunedDirs 2, ignoredCalls 21). It follows exactly the split #3043 used for `env`/`generated` — a pure-string rule in `shouldIgnorePath`, an fs probe in the path-aware function. About 8 lines. (A probe *alone*, without the fragments, buys nothing — `ignored()` still routes through the pure-string `shouldIgnorePath`.)

**Not adopting it, deliberately.** The trigger is hypothetical — no observed repo keeps source under `_next`, and unlike `'packages'` (removed from this set precisely because it *is* a common source directory) `_next` is Next.js's reserved emitted path. #3043 itself is the precedent for fixing this class *reactively*, once a real report exists. The design above is written into the PR body's out-of-scope section so it is ready if one arrives.

**Part 3 leak reproduced and now documented.** `!android/` in `.gitnexusignore` *or* `.gitignore` voids the new rule entirely: 5 of 5 Capacitor chunks leak, on `main` and this branch alike. Pre-existing #771 precedence, not introduced here — but it is exactly the population #3007 is about, so the body now says a Capacitor user who unignores the native shell must add an explicit `_next/` line.

gitnexus/test/unit/ignore-build-output.test.ts
magyargergo: **Re-measured against the current base and fixed.** Base has moved since this comment — `origin/main` is now `414687ad1`, including #3043's rewrite of this file — so the "3 of 5" number needed re-establishing rather than assuming.

Ran the PR's byte-identical test file against the current base source: `Tests 2 failed | 3 passed (5)`. Still exactly 3 of 5, same three. The two that fail on base are the `_next` case (`:19`) and the slash-free guard (`:75`) — as you noted, those two are the real regression tests.

Ablation pins *which* rule carries this one: comment out `'build'` in the base copy and it flips to `× ignores public/build … expected false to be true`. So it was decided entirely by a rule the PR does not touch.

Took your first suggestion. Renamed to `keeps public/build ignored after the inert name-set entry was removed`, and replaced the comment with what it now pins — that **deleting** the inert entry changed no behavior, green on both sides by design. That makes it load-bearing for the deletion instead of misleading about a fix.

**A worse gap in the same file that this comment didn't reach.** Test #2 (`does not ignore ordinary source that merely mentions next`) uses `src/next-steps.ts`, `src/nextConfig/index.ts`, `packages/next-auth/src/index.ts` — none contains `_next` as a substring. So the suite could not distinguish a segment rule from a substring rule: replacing the entry with `if (normalizedPath.includes('_next')) return true;` passed **all 5 tests** while silently eating `src/_nextgen/index.ts`. Added `matches _next as a whole segment, not as a substring` (`src/_nextgen/index.ts`, `packages/my_next/src/index.ts`, `src/prefix_next.ts`); ablated to confirm it goes red against the substring implementation.

PR #2191, opened 2026-06-13
fix(group): pin repos during sync so large groups resolve cross-links

magyargergo:
## Summary

`gitnexus group sync` on a group with **more than 5 repos** silently produced cross-links anchored to *synthetic* UIDs (`manifest::<repo>::<contract>`) instead of real graph symbols, so cross-repo impact analysis through manifest-declared contracts could not actually trace into the affected repos. The log showed a burst of `level:40` `LadybugDB not initialized for repo "X". Call initLbug first.` warnings during manifest extraction.

After this change, every group member stays resident through resolution: cross-links anchor to **real** graph symbol UIDs, and the "not initialized" warnings are gone.

Fixes #2189.

## Why it happened

The LadybugDB connection pool caps residency at `MAX_POOL_SIZE = 5` with LRU eviction. `syncGroup` opens one pool per repo in an init loop, then runs **deferred** workspace + manifest cross-repo resolution *after* the loop using executor closures captured in `dbExecutors`. With more than 5 repos, the earliest repos are LRU-evicted as later repos initialize (nothing is checked out between loop iterations, so every prior repo is eligible). By resolution time their pools are gone; the stale executor closures call `executeParameterized` on an evicted `poolId` → throw → the manifest extractor catches it and falls back to a synthetic UID.

A prior fix (#1802) moved deferred resolution *inside* the `try` so the `finally`'s `closeLbug` no longer tore pools down early. That closed one eviction window; this closes the remaining one — eviction *during* the init loop.

```
BEFORE (group of 6, cap 5)          AFTER (pinned)
init r1..r5  → pool {r1..r5}        init r1..r5; pinRepo each
init r6      → evictLRU drops r1     init r6 → evictLRU skips pinned → pool {r1..r6}
resolve r1   → throw "not init"      resolve r1 → real rows → real symbolUid
             → synthetic UID         finally: unpin + close each repo
```

## The fix

Add repo **pinning** to the pool adapter and wire it into `syncGroup`:

- `pinRepo(repoId)` / `unpinRepo(repoId)` on the pool adapter; `evictLRU` and the idle-timeout sweep skip pinned repos; `closeOne` clears the pin on teardown.
- `syncGroup` pins each repo immediately after `initLbug` and, in its `finally`, unpins then closes every opened pool.

Design decisions:

[...]

PR #2186, opened 2026-06-13
fix(mcp): rename query/cypher params so Claude Code can call them

magyargergo:
## Summary

From Claude Code — the most widely used MCP client — `query(...)` and `cypher(...)` were unusable: every call came back with `"query parameter is required and cannot be empty."`, even with a query supplied. Claude Code silently drops a tool-call argument whose key is exactly `query`, so the server received an empty value. Every other GitNexus tool worked, because their required params are named `name` / `target` / `new_name`. The net effect: the entire embeddings + hybrid-search investment was inaccessible from Claude Code.

This renames the advertised parameters to names the client actually transmits — **`search_query`** for the `query` tool and **`statement`** for the `cypher` tool — while the handlers still accept the legacy **`query`** key. Claude Code works now; nothing else breaks.

We can't patch Claude Code (closed source), and the reporter already proved non-`query` names transmit fine, so renaming away from `query` is the fix regardless of the upstream cause.

## What changed

- **Schema** (`gitnexus/src/mcp/tools.ts`): `query` tool advertises `search_query` (required); `cypher` tool advertises `statement` (required). The legacy `query` key is intentionally **not** re-advertised as a property — doing so would re-trigger the exact Claude Code arg-drop. It lives in the field `description` instead.
- **Handlers** (`gitnexus/src/mcp/local/local-backend.ts`): the search text / statement is resolved as `search_query ?? query` and `statement ?? query` at every consumer site — the local handler, the legacy `search` alias, and the cross-repo `@group` forward — so precedence is uniform (**new name wins**) with no hidden param mutation. Both MCP transports (stdio and HTTP `/api/mcp`) share this one backend, so the behavior change is a single chokepoint. `cypher()` now also returns the same friendly required-param error as `query()` instead of a raw DB error.
- **CLI** (`gitnexus/src/cli/tool.ts`): passes the canonical names (user-facing positional args unchanged).
- **Agent-facing docs**: the three example generators (`ai-context.ts`, `skill-gen.ts`, `resources.ts`) and all committed skill mirrors (canonical `gitnexus/skills/*`, `.claude/skills`, the Claude plugin + Cursor distributions, `AGENTS.md`, `CLAUDE.md`, READMEs) now show `query({search_query: ...})` so an agent following them emits the working key.

## Backward compatibility

[...]

PR #3020, opened 2026-08-20
fix(group)!: stop group sync claiming matching it never did

### LINKED ISSUE #3006: fix(group): matching.bm25_threshold and matching.embedding_threshold are written into every group.yaml but read by no matcher
### Area

### Summary

`gitnexus group create` writes `matching.bm25_threshold` and `matching.embedding_threshold` into every generated `group.yaml`, but no matcher reads either value — turning them changes nothing.

### Context

Found while investigating why cross-repo linking produced almost no links on a real backend/frontend pair. The thresholds are the first thing an operator reaches for when links don't match, so a knob that silently does nothing sends people down a dead end before they look at the extractor.

Verified against `main` @ `aac7515d`.

### Expected behavior

Either the BM25 / embedding stages of the matching cascade run and honor these thresholds, or the keys are not written into the generated config.

### Actual behavior

The two keys are **parsed and persisted only**. Across `src/core/group/` they appear in exactly three places:

- `src/core/group/types.ts:36-37` — interface fields
- `src/core/group/config-parser.ts:39-40` — `DEFAULT_MATCHING`
- `src/core/group/storage.ts:100-101` — the generated YAML template

`matching.ts` exports only `buildProviderIndex`, `runExactMatch`, `runWildcardMatch`, and `sync.ts` imports exactly those three. There is no BM25 matcher, no embedding matcher, and no cosine similarity anywhere in the group code.

The type says the same thing:

```ts
// src/core/group/types.ts:2
export type MatchType = 'exact' | 'manifest' | 'wildcard' | 'bm25' | 'embedding';
```

Two of those five variants are unreachable. So it isn't only the YAML template advertising a cascade that doesn't exist — the type does too, which is how a reader concludes the code path must be in there somewhere.

The sync output confirms it: the cascade prints one stage.

```
Matching cascade:
  exact:     1 cross-links (confidence 1.0)
  unmatched: 225 contracts
```

The MCP `group_sync` tool description hints at the same state — `skipEmbeddings` is documented as *"Exact + BM25 only (Demo PR: same as default exact path)"*.

### Steps to reproduce

1. `gitnexus group create <name>`
2. `gitnexus group add <name> backend <repoA>` and `gitnexus group add <name> frontend <repoB>`
3. Open `~/.gitnexus/groups/<name>/group.yaml` — note `matching.bm25_threshold: 0.7` and `matching.embedding_threshold: 0.65`
4. `gitnexus group sync <name>` — note the cascade prints only `exact:` and `unmatched:`
5. Change either threshold to any value and re-run — output is byte-identical

### Environment

- OS: macOS
- Node: 22.22.0
- Commit: `aac7515d` (main)

### Logs / screenshots

Suggested resolution, in preference order:

1. Remove both keys from the generated template in `storage.ts` and drop the unreachable `'bm25' | 'embedding'` variants from `MatchType`, leaving the cascade honestly described as exact + manifest + wildcard.
2. Or wire the stages up and keep the knobs.

Either is fine; the current state is the one that isn't, because it reads as a feature that's broken rather than one that was never built.

ReidenXerx:
## Summary

Fixes #3006. Every place `gitnexus group sync` told you something untrue about its matching, in one pass.

It started narrow: `gitnexus group create` wrote `matching.bm25_threshold` and `matching.embedding_threshold` into every generated `group.yaml`, and no matcher ever read either. Investigating turned up a whole surface — config keys, a CLI flag, an MCP parameter, localized help in two languages — describing a BM25/embedding matching cascade that was never built.

Review then found the same defect in the flags left behind. `--exact-only` was inert too, and strictly worse: it promised to suppress a stage that **does** run and **does** write cross-links. Honouring it exposed the next layer — a sync could now deliberately produce a narrower graph, and nothing recorded that, so `group_impact` and cross-repo `trace` reported a knowingly incomplete answer as complete.

So this does three things: **removes** the surface describing work never done, **honours** the flag whose stage actually exists, and **reports** a narrowed result everywhere it is read — the registry, the bridge, three MCP payloads and three CLI renderers — under its own truncation reason, because its remedy (re-run without the flag) differs from re-syncing a repo that failed to read.

> ⚠️ **Breaking.** Two CLI flags (`--skip-embeddings`, `--allow-stale`), one MCP parameter
> (`skipEmbeddings`), and five `group.yaml` keys (`matching.bm25_threshold`,
> `matching.embedding_threshold`, `matching.max_candidates_per_step`,
> `detect.embedding_fallback`, `detect.shared_libs`) are removed. The CLI removals ARE a
> runtime break — commander exits 1 on an unknown option, before the action runs. Existing
> `group.yaml` files keep loading and keep their keys. See Risk & rollout.

## Motivation / context

The inert surface, in full:

| Surface | State |
| --- | --- |
| `matching.bm25_threshold` / `matching.embedding_threshold` | parsed (`config-parser.ts`), persisted (`storage.ts`), read by nothing |
| `detect.embedding_fallback` | defaulted and written to the template, read by nothing |
| `MatchType = 'exact' \| 'manifest' \| 'wildcard' \| 'bm25' \| 'embedding'` | 2 of 5 variants unreachable |
| `SyncOptions.skipEmbeddings` | declared in `sync.ts` and never read — the only occurrence in the file was the declaration |
| `gitnexus group sync --skip-embeddings` | accepted, threaded through `GroupService`, ignored |
| CLI help, `en` + `zh-CN` | *"Exact + BM25 only (no embedding fallback)"* |
| MCP `group_sync` schema | `skipEmbeddings`, described as *"Exact + BM25 only (Demo PR: same as default exact path)"* |

`sync.ts` imports exactly `buildProviderIndex`, `runExactMatch` and `runWildcardMatch`, and the printed cascade has one stage:

```
Matching cascade:
  exact:     1 cross-links (confidence 1.0)
  unmatched: 225 contracts
```

This matters beyond tidiness. An operator whose links do not match reaches for those thresholds first, and turning either knob changes nothing. Config that silently does nothing is how people conclude a feature is broken.

## Why remove rather than implement

Measured on a real backend/frontend pair. Of **165 consumer contracts, 149 link exactly and 16 do not**:

- **9 are third-party** — Google OAuth (`/oauth2/v3/tokeninfo`, `/oauth2/v2/userinfo`), Apple public keys (`/auth/keys`), PostHog, image annotation. These have no in-group provider *by construction*. Similarity matching cannot recover them; it can only invent false links.
- **2 are verb mismatches.** The frontend calls `POST /links` and `GET /links/check-exists`; the backend declares `GET /links` and eleven other `/links/*` routes, but neither of those. A fuzzy path match would link a POST consumer to a GET provider — worse than leaving it unmatched.
- **The rest are path-extraction artifacts** — a leading `{param}` from an unresolved base URL, and a ternary inside a template.

So roughly **none** of the sixteen would be correctly recovered, and several would be actively mis-linked. Cross-link output is presented as fact, so a wrong link costs more than a missing one.

Worth noting for anyone weighing the other direction: the repo's existing BM25 code (`core/search/bm25-index.ts`) is `searchFTSFromLbug`, a per-repo full-text query over graph symbols. It is not a path-vs-path similarity scorer, so wiring the cascade would mean writing a new matcher, not reusing what is there.

## Areas touched

## Scope & constraints

**In scope**

- Remove the two `matching` thresholds and `detect.embedding_fallback` from the schema, defaults, and generated template.
- Remove the unreachable `'bm25' | 'embedding'` `MatchType` variants.
- Remove `skipEmbeddings` from `SyncOptions`, `GroupService`, the CLI flag, both help catalogues, and the MCP `group_sync` schema.

- **No matcher is added or changed.** `runExactMatch` / `runWildcardMatch` behavior is untouched; this only deletes surface that described a stage that never ran.
- **`exclude_links_paths` and `exclude_links_param_only_paths` stay.** Those are read and work (`matching.ts`).
- **`matching.max_candidates_per_step` and `detect.shared_libs` are now also removed.** Review found both inert in exactly the same three-place way as the keys above: declared, defaulted, written into every generated `group.yaml`, and read by nothing. `shared_libs` gates no extractor at all — `'lib'` contracts come only from the operator-declared manifest extractor.
- **The `POST /links` / `GET /links/check-exists` mismatch is not investigated.** It surfaced while gathering evidence and looks like a genuine frontend/backend divergence in that codebase, not a GitNexus bug.

## Implementation notes

**Existing `group.yaml` files keep loading.** `parseGroupConfig` builds its config as `{ ...DEFAULT_MATCHING, ...(raw.matching as object) }`, so a removed key present in a user's YAML is simply not part of the typed schema any more — it does not throw, and nothing reads it. A regression test pins this, and pins PRESERVE rather than merely "does not throw": a legacy config carrying every removed key still parses AND still carries those keys through, which matters because `group add` / `group remove` round-trip the operator's file through `loadGroupConfig` -> `yaml.dump` -> write. Anything the parser dropped would be deleted from their checked-in `group.yaml` on their next `group add`.

**The tests that changed asserted defaults that no longer exist.** The minimal-config case now asserts `exclude_links_paths` / `exclude_links_param_only_paths`, which are genuinely read; the legacy-config case gained preserve assertions for every removed key.

The YAML fixtures across the group test suite still carry the old keys, deliberately — they are now backward-compatibility evidence rather than schema examples.

## Testing & verification

- [x] `cd gitnexus && npx vitest run test/unit/group/ test/unit/tools.test.ts test/integration/group/` — 1167 passed / 0 failed.
- [x] `cd gitnexus && npx tsc --noEmit` — clean. `tsc -p tsconfig.test.json --noEmit` is 987 against a 989 baseline taken on the branch tip before the config-key removal — a gate that drifted to 994 mid-PR and was caught by review, not CI, since `tsconfig.json` is src-only and no workflow sees test-file type errors (measured against the branch tip, not `origin/main`, so the number reflects this work rather than the whole branch).
- [x] `npx eslint` — 0 errors (21 pre-existing non-null-assertion warnings, none new)
- [ ] `cd gitnexus-web && npm test` *(web untouched)*

Behavior checked against a real group:

```
$ gitnexus group sync <group> --skip-embeddings
error: unknown option '--skip-embeddings'

$ gitnexus group sync <group>
Matching cascade:
  exact:     1 cross-links (confidence 1.0)
Wrote contracts.json (233 contracts, 1 cross-links)
```

The exact/wildcard cascade is unchanged; only the flag is gone.

## Risk & rollout

- **BREAKING: two CLI flags, one MCP parameter and five config keys are removed.** Both were accepted and ignored, so nothing behaves differently — but a script or agent passing them now fails with `unknown option` rather than being silently misled. That is the intended trade: a loud failure beats a flag that pretends to do something. If maintainers prefer to stage this over a release, the alternative is to keep the flag accepted and mark it a documented no-op; happy to switch.
- **Persisted artifacts gain one optional field.** `contracts.json` and the bridge's `meta.json` now carry `suppressedMatchStages`. Additive and absent-tolerant both ways: a new build reads an old artifact as "not recorded", an older build ignores the key. No re-index or re-sync is required, and no schema version is bumped — see the forward-compatibility item under *Deliberately not done*.
- **Existing configs are unaffected** — verified by the legacy-config regression test.
- **`detect_changes` reports HIGH**, but that is process membership rather than blast radius: `registerGroupCommands` is step 1 of ten `RegisterGroupCommands → …` flows because it is the CLI entry point for every group subcommand. The genuine risk in this PR is the removed flag above, not the process count.
- **Rollback:** revert the commit. Nothing persisted depends on it.
- Release notes: `fix` with `!` → 💥 Breaking Changes.

## Checklist

- [x] Breaking change flagged with `!` in the title and `BREAKING CHANGE:` in the commit body
- [x] No new `GITNEXUS_*` environment variables

---

## Follow-up round: review findings addressed

This PR was reviewed by a multi-engine pass (five reviewer lenses plus an independent cross-model pass). Ten findings; all are closed here.

**The flags now do what they say.** `--exact-only` was inert in exactly the way `--skip-embeddings` was — declared, threaded through the CLI and MCP, read by nothing — and strictly worse, because the stage it promised to suppress *does* run and *does* write links. It is now honoured. `--allow-stale` had no stage to gate (the staleness check lives in `group status`, a different path), so it is removed under the same rationale as `--skip-embeddings`.

**A narrowed sync now says so.** An `--exact-only` run writes fewer cross-links, and nothing recorded that a stage had been suppressed by request. `contracts.json` and the bridge now carry `suppressedMatchStages`, and `crossRepoCompleteness` folds it in — so `group_impact`, cross-repo `trace`, `group_contracts` and `group status` stop reporting a narrowed graph as complete. It gets its own truncation reason (`'suppressed-stage'`) rather than reusing `'incomplete-sync'`, because the remedy differs: re-run without the flag, not repair a repo that read fine.

**`group_sync` validates its input.** `Boolean(params.exactOnly)` on an untyped MCP payload made the string `"false"` truthy — a routine shape for an LLM caller — so a caller asking to KEEP wildcard matching got it suppressed and persisted. Rejected now, not coerced. The two retired parameters are refused by name, matching what the CLI already does.

**Output and descriptions match behavior.** `group sync` reports every matching stage and names a skipped one as skipped rather than printing `0`; the CLI, both locales, and the MCP schema now agree with each other and with the code.

### Deliberately not done

**Forward-compatibility for group persisted artifacts across mixed CLI/MCP versions.** An
`--exact-only` sync writes a narrowed graph carrying `suppressedMatchStages`; a build that
predates the field ignores it, sees both repo lists empty, and reports the artifact complete.
This is newly *exposed* by this PR — before it, `exactOnly` was inert, so no older reader could
meet a freshly-narrowed graph. What is pre-existing is the underlying gap: group artifacts have
no forward-versioning contract for new completeness semantics. `contracts.json` is still
`version: 1` and its reader does not enforce that version at all.

Not fixed here, and specifically **not** by bumping `BRIDGE_SCHEMA_VERSION`: that constant gates
the LadybugDB schema under a documented discard-and-resync contract, so using it to version
metadata semantics would misuse it — and would fix only half the problem, since old
`contracts.json` readers would still ignore the field. A real fix is a format/generation protocol
with fail-closed semantics, which is its own change. The risk shape to watch in the meantime is
a new CLI alongside a still-running old MCP server against the same group directory.

**Shared registry+bridge generation identity.** The two artifacts are published independently, so
a failed bridge write after a committed registry leaves them describing different syncs. This PR
contains that case by withdrawing the stale bridge's provenance so nothing calls its answer
complete, but atomic generation identity across both artifacts is the architectural fix and is
out of scope.

A repo-wide unknown-parameter layer for every MCP tool. Five parameters are read by backends and declared in no schema, and three tools dispatch with no schema entry at all, so a strict layer would reject working calls until each is reconciled — that reconciliation is the work, and it is larger than this PR.

---

## Summary

*A broad group subsystem change removes the advertised matching cascade across core logic, CLI and MCP surfaces, with substantial downstream reach. The overall blast level is critical, despite the individual file risk remaining low.*

**🔴 CRITICAL blast radius. A group subsystem change removing the advertised matching cascade, spanning CLI, MCP, core group code, and tests with reach into 34 dependents.**

The change is concentrated in the `Group` module, especially `gitnexus/src/core/group/sync.ts`, which contains 11 changed symbols. Key execution and integration points include `writeBridgeUnlocked` and `refreshPreservedBridgeMeta` in `gitnexus/src/core/group/bridge-db.ts`, `crossRepoCompleteness` in `gitnexus/src/core/group/completeness.ts`, `runGroupImpact` in `gitnexus/src/core/group/cross-impact.ts`, and `registerGroupCommands` in `gitnexus/src/cli/group.ts`.

Review the synchronization and cross-impact behavior first, then trace the corresponding CLI and MCP surfaces in `gitnexus/src/mcp/tools.ts` and `gitnexus/src/mcp/resources.ts`. The affected test coverage is spread across group CLI, matching, synchronization, locking, configuration, and tool tests. The graph assigns file risk `LOW`, and no `HIGH` or `CRITICAL` risk files were identified.

_Added by GitNexus for PR #3020. Edit freely — this block is replaced on the next review, everything above it is left untouched._


gitnexus/src/cli/group.ts
```diff
@@ -149,7 +149,6 @@ export function registerGroupCommands(program: Command): void {
   group
     .command('sync <name>')
     .description('Sync Contract Registry — extract contracts and build cross-links')
-    .option('--skip-embeddings', 'Exact + BM25 only (no embedding fallback)')
     .option('--exact-only', 'Exact match only')
```
magyargergo: **P1** — `--exact-only` is inert in exactly the way `--skip-embeddings` was, and it is the worse of the two. `SyncOptions.exactOnly` (`sync.ts:43`) sits on the line immediately before the `skipEmbeddings?: boolean;` this PR deletes; `git log -S` shows both were introduced by the same commit (`4fed097ab`) on adjacent lines in all four files. `grep` on the NUL-stripped `sync.ts` returns `exactOnly` exactly once — the declaration. `syncGroup` runs `runExactMatch` then `runWildcardMatch` unconditionally at `sync.ts:381-382`. Executed: calling `syncGroup` with `exactOnly: true` on a wildcard-matchable gRPC/Thrift pair returns `crossLinks=1, types=['wildcard']`, identical to `exactOnly: false`. So unlike `skipEmbeddings` — inert but harmless, because the stage it disabled never existed — `--exact-only` promises to suppress a stage that runs and writes links into `contracts.json` and the bridge, which `group impact` and cross-repo `trace` then traverse. `--allow-stale` is in the same state (declared, never read; `syncGroup` emits no stale warning anywhere). **Fix:** either gate it — `const wildcard = opts?.exactOnly ? { matched: [], unmatched } : runWildcardMatch(unmatched, providerIndex);` — or delete `exactOnly`/`allowStale` in this same PR under the same rationale and the same `!`. Either is defensible; leaving it silent is the defect #3006 reports. `[reproduced]`

gitnexus/src/mcp/tools.ts
```diff
@@ -859,10 +859,6 @@ WHEN TO USE: After changing group.yaml or re-indexing member repos.`,
       type: 'object',
       properties: {
         name: { type: 'string', description: 'Group name' },
-        skipEmbeddings: {
-          type: 'boolean',
-          description: 'Exact + BM25 only (Demo PR: same as default exact path)',
-        },
         exactOnly: { type: 'boolean', description: 'Exact match only in cascade' },
```
magyargergo: **P2** — after this PR, `'Exact match only in cascade'` is the only place the word "cascade" survives in the MCP surface, on a parameter that does nothing (see the `--exact-only` thread). This string is shipped to every MCP client as the tool's parameter documentation, so it is the one place an agent reads about a cascade this PR's title says does not exist. Also note `mcp/tools.ts:855` describes `group_sync` as "extract HTTP contracts, apply manifest links, exact-match cross-links" — it omits the wildcard stage, which always runs. **Fix:** `'Exact match only (skip wildcard service matching)'`, and make it true; or drop the property if `exactOnly` is deleted. `[code-read]`

gitnexus/src/cli/group.ts
```diff
@@ -168,7 +167,6 @@ export function registerGroupCommands(program: Command): void {
         groupDir,
         allowStale: Boolean(opts.allowStale),
         verbose: Boolean(opts.verbose),
-        skipEmbeddings: Boolean(opts.skipEmbeddings),
         exactOnly: Boolean(opts.exactOnly),
```
magyargergo: **P2** — six lines below this hunk, `cli/group.ts:176` still prints `` `\nMatching cascade:` `` — after `git grep -ani "matching cascade"`, the only surviving hit repo-wide. The PR body quotes this exact output as the evidence that the cascade has one stage, then leaves the header in place. The block is also incomplete about the pipeline that *does* exist: it filters `matchType === 'exact'` only, while `result.crossLinks` is `dedupeCrossLinks([...manifestCrossLinks, ...matched, ...wildcard.matched])` (`sync.ts:388`) and the very next line prints `result.crossLinks.length` — so for any group with manifest or wildcard links, the two numbers disagree with no explanation, and an operator can see `exact: 0` above `Wrote contracts.json (…, 37 cross-links)`. Issue #3006's preferred resolution asked for the cascade to be "honestly described as exact + manifest + wildcard"; this is the clause not delivered. **Fix:** rename the header (`Matching:` / `Cross-link matching:`) and print a group-by over `matchType` so the stage counts reconcile with the total. `[code-read]`

gitnexus/src/core/group/types.ts
```diff
@@ -27,14 +27,11 @@ export interface DetectConfig {
   thrift: boolean;
   topics: boolean;
   shared_libs: boolean;
-  embedding_fallback: boolean;
   includes: boolean;
   workspace_deps: boolean;
 }
 
 export interface MatchingConfig {
```
magyargergo, line 33: **P2** — narrowing `MatchingConfig`/`DetectConfig` leaves 30 new `TS2353` excess-property errors in four test files this PR does not touch: `sync.test.ts` ×16, `matching.test.ts` ×8, `types.test.ts` ×4, `sync-windowed-resolution.test.ts` ×2 (base 0 → head 30, measured with a clean base control). **To be clear about what this is not:** the full `tsc -p tsconfig.test.json --noEmit` run is **999 errors at head against 980 at base** (both measured with a clean base control) — 11 of the 30 replace pre-existing missing-`includes` errors on the same lines, so the net rise on the full run is **+19**, not +30. That config is referenced by no workflow and no npm script, so this PR does not turn a green gate red, because there is no green gate, and runtime is unaffected (vitest strips types via esbuild; 809/809 group tests pass). Two things still make it worth a mechanical fix here: the PR body justifies these as *"YAML fixtures … deliberately"* retained, but `matching.test.ts:639`, `types.test.ts:14` and `sync.test.ts:17` are **annotated TypeScript literals** — direct compile-time consumers of the contract just changed, not runtime back-compat evidence (the YAML-string fixtures genuinely are, and should be left alone); and `matching.test.ts` goes from fully clean at base to fully red, 100% attributable. Heads-up on the fix: TypeScript suppresses the missing-property diagnostic once it reports an excess one, so deleting `embedding_fallback` from the `DetectConfig` literals will surface a pre-existing missing `includes` — expect to add it in the same edit. `[reproduced]`

gitnexus/src/core/group/config-parser.ts
```diff
@@ -30,14 +30,11 @@ const DEFAULT_DETECT = {
...
   shared_libs: true,
-  embedding_fallback: true,
   includes: false,
   workspace_deps: false,
 };
 
 const DEFAULT_MATCHING = {
-  bm25_threshold: 0.7,
-  embedding_threshold: 0.65,
   max_candidates_per_step: 3,
```
magyargergo: **P2** — the PR body says under "Explicitly out of scope": *"`matching.max_candidates_per_step`, `exclude_links_paths` and `exclude_links_param_only_paths` stay. Those are read and work."* True for the latter two (`matching.ts:35-38`), **false for `max_candidates_per_step`**: `git grep -an` across `gitnexus/src`, `gitnexus-shared`, `gitnexus-web` and `docker-server.mjs` returns exactly three hits — `config-parser.ts:38`, `storage.ts:99`, `types.ts:35` — all definitional, zero readers. That is byte-for-byte the same three-place signature #3006 itself used to prove the removed keys inert, and `gitnexus group create` still writes it into every generated `group.yaml`. Dynamic access is ruled out: the only `config.matching` readers are the spread here and the two wholesale passes at `sync.ts:380-381`; no `Object.entries`, no indexed access. `detect.shared_libs` (line 32 above, directly where `embedding_fallback` was deleted) is identical — every *other* `DetectConfig` field has a real reader in `sync.ts` (`http:207`, `grpc:218`, `thrift:229`, `topics:240`, `includes:251`, `workspace_deps:288`); `shared_libs` has none. Compounding: this PR's test edit promotes `max_candidates_per_step` to the sentinel assertion at `config-parser.test.ts:62` and `:89`, cementing it. **Fix:** remove both here (the sweep is already breaking, so marginal cost is zero) and re-point the test assertions at `exclude_links_paths` / `exclude_links_param_only_paths`, which are genuinely read — or correct the out-of-scope paragraph to say they are also unread and deliberately deferred. `[reproduced]`

gitnexus/test/unit/group/config-parser.test.ts
```diff
@@ -59,11 +59,36 @@ repos:
     expect(config.links).toEqual([]);
     expect(config.packages).toEqual({});
     expect(config.detect.http).toBe(true);
-    expect(config.matching.bm25_threshold).toBe(0.7);
+    expect(config.matching.max_candidates_per_step).toBe(3);
     expect(config.matching.exclude_links_paths).toEqual([]);
     expect(config.matching.exclude_links_param_only_paths).toBe(false);
   });
 
+  it('still parses a legacy config carrying the removed matching knobs', () => {
```
magyargergo, line 66: **P2** — this is the only test added, and it covers the parser, not the change. `grep -rn "skipEmbeddings" gitnexus/test/` returns nothing: no test asserts the CLI now rejects `--skip-embeddings` (reproduced: `error: unknown option`, exit 1), and no test asserts the MCP schema no longer advertises it. Both harnesses already exist — `tools.test.ts:289` uses exactly the right idiom one line from where it is needed (`expect(tool.inputSchema.properties).not.toHaveProperty('repo')`), and `test/integration/group/group-cli.test.ts` already spawns the CLI with a negative-path precedent (`expect(result.status).not.toBe(0)` plus a stderr match). Two lines of test pin the headline breaking change in both directions and would also catch an accidental re-add. Worth deciding explicitly in the same breath whether callers should get a hard error or an accepted-no-op deprecation for one release — the PR body offers the latter and nothing records the decision. `[reproduced]`

gitnexus/test/unit/group/config-parser.test.ts
```diff
@@ -59,11 +59,36 @@ repos:
...
+matching:
+  bm25_threshold: 0.7
+  embedding_threshold: 0.65
+  max_candidates_per_step: 3
+`;
+    const config = parseGroupConfig(legacy);
+    expect(config.name).toBe('test');
+    expect(config.repos).toEqual({ app: 'my-app' });
+    expect(config.detect.http).toBe(true);
+    expect(config.matching.max_candidates_per_step).toBe(3);
```
magyargergo: **P3** — all four assertions in this test are already satisfied by parser defaults alone, with no `detect:`/`matching:` block present at all; the pre-existing test two blocks above asserts the same two facts against a minimal YAML. What the test actually pins is the implicit "must not throw", which is real and worth having — it would go red if `parseGroupConfig` gained strict-key rejection. What it cannot see is the **strip** case, which is the likelier accident: zod's `.strip()` is its *default* mode, so a future migration to a real validator would keep this green while silently scrubbing operators' `bm25_threshold`/`embedding_threshold`/`embedding_fallback` out of their checked-in `group.yaml` on the next `gitnexus group add` — that round-trip is real (`cli/group.ts:39` and `:60` do `loadGroupConfig` → `yaml.dump`), and it was executed here: all three legacy keys survive today. One assertion pins the behavior the safety argument actually rests on: `expect((config.matching as Record<string, unknown>).bm25_threshold).toBe(0.7)` (pins preserve) or `.not.toHaveProperty('bm25_threshold')` (pins strip). Picking one requires deciding which is intended, which the PR does not state. `[reproduced]`

magyargergo (commented):
## Multi-engine review — PR #3020

**How this was reviewed.** Two independent engines, same scaffolding on both sides. **Claude** ran the GitNexus production-readiness swarm (pr-facts/history, branch-hygiene, risk-architect, test-CI, security/trust-boundary, docs-DoD) plus the `ce-code-review` personas (correctness, adversarial, api-contract, testing, maintainability, project-standards), with a coordinator lane that executed and re-verified the contested measurements. **Codex (GPT-5.5, xhigh)** ran the same swarm, the same `ce-code-review` pass, and a native adversarial pass — **all three Codex legs ran live and produced real findings**, so every "both engines" claim below means two genuinely independent tool-driven runs, not one run reported twice.

Confidence labels: **Strong** = both engines reached it independently, or it was reproduced by execution. **Moderate** = several lanes within one engine. **Weak** = single lane.

---

### What holds up

This is a well-evidenced deletion, and the parts that were checked hardest are the parts that came back clean:

- **The "never built" claim is true, and was verified exhaustively.** Every historical revision of `matching.ts` and `sync.ts` was walked for `runBm25|bm25Match|runEmbedding|cosine|similarity` — zero hits in every revision. `matching.ts` has exported exactly `normalizeContractId`/`runExactMatch` (since `52277247f`) plus `buildProviderIndex`/`runWildcardMatch` (since `1ff324ca1`) and nothing else. All 58 files ever added under `src/core/group/` were enumerated; there has never been a BM25 or embedding matcher. No implementation was written and later reverted.
- **Zero persisted-data risk — refuted at four layers.** No producer ever emitted `matchType: 'bm25'|'embedding'` (`git log --all -S` finds no commit); persistence is an untyped `matchType STRING` column (`bridge-schema.ts:53`); the resilient loader `isCrossLink` (`service.ts:215-223`) validates only `from.repo`/`to.repo`/`contractId`/`type`; and the one narrowing read is a cast (`cross-impact.ts:771`), with no exhaustiveness check on `MatchType` anywhere. The PR's *"no re-index or re-sync is required"* is accurate.
- **The docs sweep is genuinely clean, and correctly made zero edits.** Repo-wide grep for `bm25_threshold`, `embedding_threshold`, `embedding_fallback`, and `group sync --skip-embeddings` returns no hit in any `.md`, README, `docs/`, `llms.txt`, or skill file. `PIPELINE.md:27` already described the pipeline honestly as "exact + wildcard". No schema doc for `group.yaml` exists, so no drift was possible.
- **i18n is complete and test-enforced.** `help-i18n.ts`, `en.ts`, and `zh-CN.ts` dropped the key together; `cli-i18n.test.ts:43` (locale key parity) and `cli-index-help.test.ts:166-181` (parses `cli/group.ts` with the TS compiler API and asserts no registered option is untranslated) would both go red on a one-sided removal. No orphaned key in either direction.
- **Legacy `group.yaml` keeps loading, and legacy keys are not destroyed.** `parseGroupConfig`'s spread merge tolerates them, the new test pins it, and the `group add`/`group remove` write-back at `cli/group.ts:39`/`:60` was *executed* on a legacy file: all three removed keys survive verbatim. (One lane's "storage.ts has exactly one writer" was incomplete — the round-trip does exist; it just strengthens the backward-compat story rather than weakening it.)
- Diff hygiene is clean: single non-merge commit, no force-push, no unrelated churn, `git diff --check` clean, zero bidi/zero-width/BOM in any changed file, no secrets, `npx tsc --noEmit` (the command CI runs) exit 0, `prettier --check` and `eslint` clean, and the group unit suite green (809/809 in `test/unit/group/`).

---

### Headline finding — P1

**`exactOnly` survives this PR while being inert in exactly the way `skipEmbeddings` was, and it is the worse of the two.** `[reproduced]` · **Strong — both engines, reproduced by execution in three separate lanes.**

`SyncOptions.exactOnly` (`sync.ts:43`) sits on the line *immediately before* the `skipEmbeddings?: boolean;` this PR deletes. It is threaded through the same CLI (`cli/group.ts:152`, `:170`), the same `GroupService` call (`service.ts:352`), the same MCP schema object (`mcp/tools.ts:862`), and the same two help catalogues the PR edits — and `git log -S` shows both flags were introduced by the *same commit* (`4fed097ab`), on adjacent lines, in all four files. `grep` on the NUL-stripped `sync.ts` returns `exactOnly` exactly once: the declaration. `syncGroup` runs matching unconditionally at `sync.ts:381-382`.

Two lanes executed it rather than reading it: calling `syncGroup` with `exactOnly: true` against a wildcard-matchable gRPC/Thrift pair returns `crossLinks=1, types=['wildcard']` — identical to `exactOnly: false`. So unlike `skipEmbeddings`, which disabled a stage that never existed, `--exact-only` promises to suppress a stage that **does** run and **does** write links into `contracts.json` and the bridge. Downstream `group impact` and cross-repo `trace` then traverse crossings the operator explicitly excluded. Codex's `ce` leg independently rated this **P1 / confidence 100** and its swarm leg rated it a merge-blocking P2; five Claude lanes reached it too.

The PR's own thesis — *"config that silently does nothing is how people conclude a feature is broken"* — applies verbatim, and it stops one context line short. `--allow-stale` is in the same state (declared, never read; `syncGroup` emits no stale warning at any point). Fix is either a two-line guard on `runWildcardMatch` or the same four-site deletion already performed for `skipEmbeddings`; either is defensible, silence is not.

---

### Also raised as inline comments

- **P2 — `mcp/tools.ts:862` still advertises the deleted concept.** `exactOnly: { …, description: 'Exact match only in cascade' }` is the only remaining "cascade" in the MCP surface, on a parameter that does nothing. `[code-read]` · Strong (both engines). One lane dissented that "cascade" still fairly names the real manifest→exact→wildcard pipeline; even granting that, the description documents behavior the parameter does not have.
- **P2 — `group sync` still prints `Matching cascade:` (`cli/group.ts:176`), six lines below this PR's own edit.** It also reports only the `exact` subset while the very next line prints `result.crossLinks.length`, which includes `manifest` and `wildcard` links — so for any group with those, the two numbers disagree with no explanation. Issue #3006's own preferred resolution asked for the cascade to be *"honestly described as exact + manifest + wildcard"*; that clause is the one part not delivered. `[code-read]` · Moderate (five Claude lanes; Codex reached it only obliquely, naming "updating output" inside its `exactOnly` fix rather than surfacing it independently).
- **P2 — the type narrowing adds 30 `TS2353` errors to four untouched test files, and no gate can see them.** `sync.test.ts` ×16, `matching.test.ts` ×8, `types.test.ts` ×4, `sync-windowed-resolution.test.ts` ×2, base 0 → head 30, exactly attributable. **This does not break a passing check:** the full `tsconfig.test.json` run is **999 errors at head against 980 at base** (both measured with a clean base control), and that config is referenced by no workflow and no npm script — there was never a green gate to break. **11 of the 30 replace pre-existing missing-`includes` errors on the same lines**, so the net rise on the full run is **+19**, not +30; `matching.test.ts` is the clean-cut case (0 → 8, 100% attributable). Runtime is unaffected (vitest strips types via esbuild). Fix is mechanical. `[reproduced]` · Strong (both engines).
- **P2 — `matching.max_candidates_per_step` is inert too, and the PR body says otherwise.** The body states *"`max_candidates_per_step`, `exclude_links_paths` and `exclude_links_param_only_paths` stay. Those are read and work."* True for the latter two (`matching.ts:35-38`); false for the first — its only three occurrences are the interface field, the default, and the generated template, which is byte-for-byte the same three-place signature #3006 used to prove the removed keys inert. `gitnexus group create` still writes it into every `group.yaml`. Compounding: the PR's test edit promotes it to the sentinel assertion (`config-parser.test.ts:62` and `:89`). `detect.shared_libs` is in the identical state, one line above the removed `embedding_fallback`. `[reproduced]` · Moderate (Claude — three lanes plus coordinator re-verification; not raised by Codex).
- **P2 — the breaking change has zero test coverage in either direction.** `grep -rn "skipEmbeddings" gitnexus/test/` → nothing. Nothing asserts the CLI now rejects the flag, and nothing asserts the MCP schema no longer advertises it — despite `tools.test.ts:289` already using exactly the right idiom (`expect(...inputSchema.properties).not.toHaveProperty(...)`) and `group-cli.test.ts` already having a spawn harness with a negative-path precedent. Both engines recommended adding these. `[reproduced]` · Strong.
- **P3 — the new back-compat test proves "does not throw", not "preserves".** All four of its assertions are already satisfied by parser defaults alone; a future migration to a strict schema in zod's **default** `.strip()` mode would keep this test green while silently scrubbing operators' legacy keys on the next `group add`. One added assertion on the passed-through key closes it. `[reproduced]` · Moderate (Claude).
- **P3 — `bench/cross-repo-trace/verify.mjs:61-62` still *generates* the removed keys into a fresh `group.yaml` on every bench run.** `groupYaml()` writes `bm25_threshold: 0.7` / `embedding_threshold: 0.65` into a brand-new config, so unlike the test fixtures — which are legitimate legacy-parser evidence and should be left alone — this is a repo-local *example* that now contradicts the template `storage.ts` generates. Harmless at runtime (the parser tolerates the keys) and outside `src/`, so it is genuinely minor, but it is a one-line cleanup while the sweep is open. `[code-read]` · **Weak — escalated only by Codex** (its docs/DoD lane, P3/medium). Three Claude lanes reached the same file and declined to report it — the Claude docs/DoD lane named `verify.mjs:61-63` as "writes a `group.yaml` with all three removed keys" but filed it under *not investigated*, while the adversarial and test/CI lanes judged it harmless; a fourth miscategorised it as a legacy fixture. The delta between the engines here is severity judgement, not discovery.

---

### Description accuracy — worth correcting before merge

The commit body's *"Both were accepted and ignored, so no runtime behavior changes"* is **half right, and the two halves fail in opposite directions**:

- **CLI: hard break, reproduced.** Commander 15 has no `allowUnknownOption`/`exitOverride` anywhere in `src/`, so `gitnexus group sync <g> --skip-embeddings` now prints `error: unknown option` and exits **1 before the action runs**. Executed A/B on base vs head: base completed and wrote `contracts.json`, head never reaches `syncGroup`. Both engines flagged this; Codex's adversarial leg rated it **high** and recommends keeping the flag as a hidden deprecated no-op for one release. Worth noting the second-order effect: in a non-`&&` pipeline the sync is skipped, `contracts.json` silently goes stale, and there is no freshness gate on the read path (`service.ts:225-240`) — so the failure is loud at the shell and quiet downstream.
- **MCP: not a break at all — and the "silently misled" state the PR says it is ending *persists* there.** `mcp/server.ts:216` passes arguments straight through, the SDK envelope is an open record, there is no `additionalProperties` anywhere in `src/mcp/tools.ts`, and `service.ts:350-355` picks named keys only. `group_sync({name, skipEmbeddings: true})` behaves identically before and after: accepted, ignored, no error. The PR removed **discoverability**, not **acceptance**. (Codex's adversarial leg raised a distinct, unverified caveat worth a thought: an MCP *client* that compiles advertised schemas into strict function signatures could reject the argument on its own side. That is a client-side hypothesis, not a server-side break — no evidence either way was gathered.)

Two smaller accuracy notes. The body justifies the leftover typed test literals as *"YAML fixtures … deliberately"* — the YAML-string fixtures genuinely are back-compat evidence and should be left alone, but `matching.test.ts:639`, `types.test.ts:14` and `sync.test.ts:17` are annotated TypeScript literals, i.e. direct compile-time consumers of the contract that just changed. And the body's *"`npm test` — 12772 passed / 7 failed … all pre-existing"* does not reconcile with CI on the same head (18,621 total, 18,548 passed, **0 failed**, 73 skipped); the 7 failures are not corroborated anywhere.

---

### Refuted — suspicions chased, triggered, and disproved

A large share of this review's effort went into trying to break the PR and failing. Recording it, because validation is a result:

| Suspicion | Verdict |
|---|---|
| Prototype pollution via `__proto__`/`constructor` under `matching:`/`detect:` in an untrusted `group.yaml` | **Refuted, executed.** js-yaml stores `__proto__` as an own data property and object spread uses `CreateDataProperty`, not `[[Set]]` — two independent defenses. `Object.prototype` unpolluted; `yaml.JSON_SCHEMA` additionally rejects `!!js/function`. Identical on base. |
| A persisted `contracts.json`/`bridge.lbug` carrying `matchType: 'bm25'` is newly rejected or reclassified | **Refuted at four layers** (see *What holds up*). No writer ever existed; no read path validates the union. |
| An MCP client still sending `skipEmbeddings` gets an error | **Refuted, traced end to end.** No schema validation exists on this path, before or after. |
| `group add`/`group remove` silently drop the operator's legacy keys | **Refuted, executed.** All three keys survive the parse → `yaml.dump` write-back; only key *order* changes. |
| The PR conflicts with `main` | **Refuted, verified two ways plus a positive control.** `git merge-tree` against `origin/main` → tree `dceb9f85…`, exit 0, no conflict messages; `sync.ts` blob byte-identical on both sides because main never touched it. The positive control (forcing an older merge-base) *does* report the conflict, so the clean result is trustworthy, not a detector false-negative. |
| `MatchType` narrowing has cross-package blast radius | **Refuted.** Zero `MatchType`/`matchType` hits in `gitnexus-shared`, `gitnexus-web`, `docker-server.mjs`, or `eval/`; `gitnexus/package.json` exposes only `bin`, no `main`/`types`/`exports`, so this is not a published type surface. The `bm25`/`skipEmbeddings` strings in `gitnexus-web` and `core/search/` belong to unrelated subsystems (WebGPU fallback dialog, FTS hybrid search) and were correctly left alone. |
| The src typecheck, lint, format, or group test suite regresses | **Refuted, all executed.** `tsc --noEmit` exit 0; eslint exit 0; prettier clean; 809/809 group tests. |
| The removal left stale references in production code | **Refuted.** `git grep -aE "skip-embeddings\|skipEmbeddings" HEAD -- gitnexus/src` → exit 1. The `src/**` sweep is complete. |

One item **not** refuted but **not this PR's fault**: `README.md:419` documents `gitnexus analyze --skip-embeddings`, and `eval/environments/gitnexus_docker.py:156` builds the same command. Verified against base (`git grep -a "skip-embeddings" aac7515d2 -- gitnexus/src`): at base the flag existed **only** on `group sync`; `analyze` has `--embeddings [limit]` and `--drop-embeddings`. **`analyze --skip-embeddings` has never existed** — this is a pre-existing README/eval-harness bug, already broken on main, and reproduced failing on both base and head. Worth a separate one-line issue, not a charge against this PR.

---

### Merge state and CI

- **Checks: 42 pass / 1 fail / 3 skipping. The single failure is Vercel — "Authorization required to deploy."** That is fork-PR deploy authorization, not code, and it is **not** in the required list. **20 of 20 required status checks are `SUCCESS`.** Calling this PR "checks failing" would be wrong.
- **`mergeStateStatus: BEHIND` is a real gate, not cosmetic.** `main`'s ruleset sets `strict_required_status_checks_policy: true`, so the branch must be updated before merge. It is 6 commits behind with **zero path overlap** with the PR's 11 files, and the update is proven conflict-free. Prefer **rebase** or a final squash — the plain "Update branch" button would add a merge-from-main commit and downgrade the branch's currently-clean hygiene. `required_approving_review_count: 0`, so approval is not a separate blocker.
- **`sync.ts` is invisible in GitHub's web diff.** It contains one raw NUL byte (a literal `\0` separator inside `join(...)` in `sortKey`), so git classifies it binary and the file renders as `Bin 17612 → 17584 bytes`. **This is pre-existing — it landed on main via PR #2191 in June and is present at base too** — but it means the *declaration* of the interface whose removal is the declared breaking change cannot be seen by any human reviewing on github.com. The recovered diff is exactly one deletion, `-  skipEmbeddings?: boolean;`, corroborated arithmetically by the NUL offset moving 5132 → 5104 (28 bytes = that line plus its newline). Reviewed out-of-band here; worth acknowledging in the thread rather than blocking on. Related: the reported `+27/−21` is really `+27/−22`, since `numstat` emits `-` for a binary file.
- **Hard binary conflict with open PR #3012.** #3012 ("surface unreadable group indexes … and escape raw NUL bytes in source", same author) shares four files. `git merge-tree` between the two heads produces stage 1/2/3 entries and `warning: Cannot merge binary files: gitnexus/src/core/group/sync.ts`. Because the file is binary, git takes one side **wholesale** — whichever lands second silently discards the other's `sync.ts` change unless resolved by hand. Note the direction: NUL counts are main=1, **#3020=1, #3012=0**. #3012 is the PR that removes the NUL and adds a guard test forbidding raw control bytes in tracked TypeScript; if #3020 lands second and the conflict resolves in its favor, it reintroduces the NUL and turns that new guard test red. #3012 also builds a `GroupConfig` literal from three fields #3020 deletes. **Merge order needs a deliberate decision.** (#3020 vs `origin/main` is separately clean — different targets, both facts stand.)
- **No maintainer signal anywhere.** Zero human reviews on #3020; zero maintainer comments on #3006, which is self-filed and self-answered by this PR's author and remains open and unlabeled. The delete-vs-implement decision, the scope expansion beyond what #3006 asked for (the issue never mentions the CLI flag, the MCP parameter, `SyncOptions.skipEmbeddings`, or `detect.embedding_fallback` — *everything that makes this PR breaking was added beyond the filed issue*), and the no-deprecation choice are all unratified. The PR body itself offers the alternative — *"keep the flag accepted and mark it a documented no-op; happy to switch"* — which matches this repo's only on-point precedent (#2186 renamed MCP params by **alias-and-accept**, keeping the legacy key working and adding tests for it). That offer is unanswered.

---

### Coverage limits

Read this as a strong signal, not a proof. Neither engine could reproduce the PR's core justification — the "165 contracts / 149 exact / 16 unmatched" evidence comes from a private backend/frontend pair named only in prose, so the delete-over-build argument cannot be independently checked from this repository. Real-world blast radius of the CLI removal is unknowable by construction: no invocation of `group sync --skip-embeddings` exists anywhere in this monorepo, but external CI pipelines are outside its visibility. Codex's swarm leg ran without live GitHub metadata (its `gh` access failed in-sandbox), so its mergeability and CI conclusions were marked visibility-incomplete and the numbers above come from the Claude side. The `--exact-only` reproduction was executed through `syncGroup` directly with a synthetic extractor override, not through a full end-to-end indexed group. And the 30-vs-19 net-new type-error split depends on how one counts error-class substitution on already-broken lines; both framings are given above.

---

*This is an automated multi-tool digest (GitNexus swarm + `ce-code-review` personas, run independently on Claude and on Codex GPT-5.5). Findings are cited to file and line and were verified in a worktree at the PR head, but please confirm anything you act on — especially the merge-order decision against PR #3012.*

PR #3042, opened 2026-08-25
docs(pdg): replace unbounded guard query with bounded example

gkmraju:
## Areas touched

## Scope & constraints

**In scope**

- 

- 

## Testing & verification

## Checklist

- [ ] PR body meets repo minimum length (workflow may label short descriptions)
- [ ] No secrets, tokens, or machine-specific paths committed

---

## Summary

*A small, self-contained documentation update that tightens PDG query guidance while keeping the change isolated from runtime code and execution flows.*

**🟢 LOW blast radius. A documentation and skill guidance change replacing an unbounded guard query with a bounded example across three skill documentation files, with no graph dependents.**

The change is concentrated in `gitnexus-claude-plugin/skills/gitnexus-pdg-query/SKILL.md`, which has the highest concentration of changed symbols. Review the updated `PDG query surface with GitNexus`, `The corrected guard-clause Cypher`, `Gotchas (the load-bearing ones)`, and `Mirror, don't fork` sections first.

The same guidance is mirrored in `.claude/skills/gitnexus/gitnexus-pdg-query/SKILL.md` and `gitnexus/skills/gitnexus-pdg-query.md`. Check that the corrected guard-clause Cypher and the surrounding guidance remain consistent across all three files.

_Added by GitNexus for PR #3042. Edit freely — this block is replaced on the next review, everything above it is left untouched._

### DIFF
--- a/.claude/skills/gitnexus/gitnexus-pdg-query/SKILL.md
+++ b/.claude/skills/gitnexus/gitnexus-pdg-query/SKILL.md
@@ -36,7 +36,7 @@ All three are `BasicBlock → BasicBlock` edges in the single `CodeRelation` tab
 
 - `pdg_query({ mode: 'controls', target })` — CDG. For the anchored function,
   each edge: controlling predicate block → dependent block + branch sense in
-  `label` (`'T'` = predicate's true/taken arm, `'F'` = false/fall-through). An
+  `reason` (`'T'` = predicate's true/taken arm, `'F'` = false/fall-through). An
   edge into an early-return/throw block is flagged `guard: true`.
 - `pdg_query({ mode: 'flows', target, variable? })` — REACHING_DEF def→use
   edges; `variable` filters to one binding.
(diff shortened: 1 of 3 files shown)


koriyoshi2041 (commented):
One wording edge is worth tightening before this becomes shipped guidance: `guard: true` is not limited to early `return`/`throw`. `_pdgQueryImpl` currently marks `/^\s*(return|throw|continue|break)\b/`, so “filter ... for `guard: true` when you only need early-return or throw guards” can also return loop `continue`/`break` dependents.

Either describe `guard: true` as early control-transfer guards, or tell readers to additionally filter `dependent.text` for `return|throw` when that narrower question is intended. The bounded raw Cypher below already applies that narrower text filter correctly.

I also checked the three edited mirrors at exact head `4de4eb7379`: they are byte-identical, `git diff --check` is clean, and the `reason`/target/limit claims match the current backend implementation.

PR #3043, opened 2026-08-25
fix(ingestion): discover nested source directories

### LINKED ISSUE #3039: Directories named env/ or generated/ are silently skipped at discovery, hiding real source files
## Summary

Directories whose path contains a segment named `env` or `generated` are silently excluded at file-discovery time, along with every `*.d.ts`. On our repo that drops **18 tracked, non-git-ignored TypeScript source files** — 13 of them hand-written application code.

Because the files are never parsed, their symbols have no node at all. `impact` and `context` on them return nothing, and — more dangerously — symbols they *call* report caller lists with the missing files silently absent. That reads as "unused", which is the same class of false-safe answer as #3038.

Reproduced on **v1.6.10-rc.215** (and identically on 1.6.7).

## Repro

The excluded paths on our repo:

```
apps/{admin,admin-new,client,contractor}/src/shared/env/applyEnvChrome.ts
apps/{admin,admin-new,client,contractor}/src/shared/env/getAppEnv.ts
apps/{admin,admin-new,client,contractor}/src/shared/env/EnvironmentBanner.tsx
packages/ai/src/generated/bundle.ts
apps/{admin,admin-new,client,contractor}/vite-env.d.ts
apps/admin/src/vite-env.d.ts
```

`shared/env/` is an ordinary app directory — environment *banner*, app-env getter, favicon/title chrome. Nothing virtualenv-shaped.

```bash
npx gitnexus@1.6.10-rc.215 cypher \
  "MATCH (n) WHERE n.filePath STARTS WITH 'apps/client/src/shared/env/' RETURN DISTINCT n.filePath AS f"
# []   — three real .ts/.tsx files live there
```

Full diff of source files vs indexed nodes:

```bash
npx gitnexus@1.6.10-rc.215 cypher "MATCH (n) WHERE n.filePath IS NOT NULL RETURN DISTINCT n.filePath AS f" > graph.txt
find apps convex packages -type f \( -name '*.ts' -o -name '*.tsx' \) \
  -not -path '*/node_modules/*' -not -path '*/_generated/*' | sort
# 1928 source files, 18 absent from the graph
```

The paths are also absent from `fileHashes` in `.gitnexus/meta.json`, which is what points at file discovery rather than parsing. They are tracked in git and `git check-ignore` reports nothing.

## It is segment matching, not substring

`packages/domain/src/appEnv.ts` contains `env` in its basename and indexes fine, so the rule is matching a path *segment* named exactly `env` / `generated`.

## Why it bites

`envFaviconHref` in `packages/domain/src/appEnv.ts` is called by all four `applyEnvChrome.ts` files. The graph shows:

```
context({name: "envFaviconHref"})
→ Called by: appEnv.test.ts
```

One caller, and it is the symbol's own test — a textbook "dead code, safe to delete" signal for a function four production entry points depend on.

## Suggestion

`env` and `generated` look like they are aimed at Python virtualenvs and build output, which is reasonable for those ecosystems but costly in a TS monorepo where both are ordinary directory names. Options that would fix it for us, roughly in order of preference:

1. Scope those two patterns to the languages/ecosystems they are meant for, or to a repo-root-relative position (`/env/`, `/generated/`) rather than any depth.
2. Respect `.gitignore` as the source of truth for build output, and drop the hardcoded `generated`.
3. At minimum, make the exclusion list configurable via `.gitnexusrc`, and **report skipped files** — the large-file skip already prints a notice (`Skipped 2 large files…`), which is how it should feel. A silent skip is what made this take a while to find.

Happy to test a patch against this repo — we have a 1,928-file TS/Convex monorepo that has now surfaced three of these.

azizur100389:
## Summary

- Discover tracked source in nested `env/` and `generated/` directories and include declaration files.
- Keep root artifact pruning plus `.gitignore` and `.gitnexusignore` exclusions across the walker, tsconfig scanner, and workspace scanner.

## Validation

- 201 affected tests passed; 2 platform-specific tests skipped.

Fixes #3039

---

## Summary

*This appears to be a broad repository ingestion and source discovery change. Its file-level risk is low, but the affected traversal and extraction paths have wide downstream reach.*

**🔴 CRITICAL blast radius. A repository ingestion change spanning ignore rules, filesystem walking, language discovery, and workspace import resolution, with transitive impact across the codebase.**

The change touches 12 symbols across 11 files, centered on `gitnexus/src/config/ignore-service.ts`, where `DEFAULT_IGNORE_LIST`, `IGNORED_EXTENSIONS`, `shouldIgnorePath`, and `loadIgnoreRules` define path filtering. It also changes `walkRepositoryPaths` and `readFileContents` in `gitnexus/src/core/ingestion/filesystem-walker.ts`, alongside `findPythonFiles` and `walk` in `gitnexus/src/core/group/extractors/python-workspace-extractor.ts`, `loadNodeWorkspacePackages` in `gitnexus/src/core/ingestion/import-resolvers/node-workspace-packages.ts`, and `findTsconfigFiles` in `gitnexus/src/core/ingestion/languages/typescript/tsconfig.ts`.

The graph shows 14 affected flows and 56 dependents, with impact concentrated in `Extractors` and extending into `Wiki`, `Group`, `Pipeline-phases`, `Import-resolvers`, `Typescript`, `Storage`, `Pipeline`, `Ingestion`, `Route-extractors`, and `Cobol`. Review the ignore and filesystem traversal changes first, then the nested discovery behavior covered by `gitnexus/test/integration/filesystem-walker.test.ts`, `gitnexus/test/integration/ignore-and-skip-e2e.test.ts`, and the related unit tests. No individual file is marked HIGH or CRITICAL; the file risk level is LOW.

_Added by GitNexus for PR #3043. Edit freely — this block is replaced on the next review, everything above it is left untouched._


gitnexus/src/config/ignore-service.ts
```diff
@@ -290,6 +298,10 @@ export const shouldIgnorePath = (filePath: string): boolean => {
   const fileName = parts[parts.length - 1];
   const fileNameLower = fileName.toLowerCase();
 
+  if (parts.length > 0 && ROOT_ARTIFACT_DIRECTORIES.has(parts[0].toLowerCase())) {
```
magyargergo: **[P1] The new root rule lowercases, but the list it was carved out of is matched case-sensitively — a tracked root `Generated/`, `Env/` or `ENV/` that this repo indexed before is now silently dropped.**

`parts[0].toLowerCase()` here (and `p.name.toLowerCase()` at :525, `name.toLowerCase()` at :366) matches case-insensitively. Every consumer of `DEFAULT_IGNORE_LIST` — the segment loop at :316, `DEFAULT_IGNORE_LIST.has(p.name)` at :522, `isHardcodedIgnoredDirectory` at :355 — matches case-**sensitively**. So the carve-out is not behaviour-preserving, and it moves in the opposite direction from this PR's thesis.

BASE vs HEAD on identical inputs:
```
Generated/client.ts     false -> TRUE
ENV/lib/module.py       false -> TRUE
```
and through the real `createIgnoreFilter` + `glob` pair that `filesystem-walker.ts:47-54` uses, three root directories BASE walked (`ENV/`, `Env/`, `Generated/`) disappear on HEAD.

**Trigger:** any repo with a tracked top-level `Generated/`, `Env/`, `ENV/` — e.g. a Unity/C# solution with checked-in codegen, or a root `Env/` config module. Those are legal, distinct directories on a case-sensitive filesystem, and their files were indexed before this PR.

This is the same silent-discovery-loss that #3039 reports, reintroduced in the other direction — and #3039 explicitly asked for a skip notice ("a silent skip is what made this take a while to find"), which this PR does not add. No test covers either case variant.

**Fix:** drop `.toLowerCase()` at all three sites (`ROOT_ARTIFACT_DIRECTORIES.has(parts[0])`) to match the case-sensitive semantics of the list it replaced, and add a test pinning `Generated/x.cs`. If case-insensitive matching is genuinely wanted, apply it to `DEFAULT_IGNORE_LIST` too — that is a separate, larger change.

`[reproduced]` — found independently by Codex and by a Claude lane, and re-run by the review coordinator.

gitnexus/src/config/ignore-service.ts
```diff
@@ -329,10 +341,8 @@ export const shouldIgnorePath = (filePath: string): boolean => {
   if (
     fileNameLower.includes('.bundle.') ||
     fileNameLower.includes('.chunk.') ||
-    fileNameLower.includes('.generated.') ||
-    fileNameLower.endsWith('.d.ts')
+    fileNameLower.includes('.generated.')
```
magyargergo, line 349: **[P2] Un-ignoring `.d.ts` yields a full duplicate symbol set for in-place declaration output, and the duplicate defeats short-name class resolution — dropping bindings that resolved before.**

(The removal of `fileNameLower.endsWith('.d.ts')` is this hunk; the consequences are in unchanged files.)

Real worker-pool parse of `src/service.ts` + `src/service.d.ts`:
```
Class      UserService   src/service.d.ts
Class      UserService   src/service.ts
Interface  User          src/service.d.ts
Interface  User          src/service.ts
Method     load          src/service.d.ts
Method     load          src/service.ts
```
Every exported symbol now exists twice with the same label and name. `getProviderForFile` (`languages/index.ts:66-67`) slices from the **last** dot, so `.d.ts` routes to the TypeScript provider with no compound-extension branch, and `CLASS_LIKE_TYPES` (`type-env.ts:500`) includes `Interface` — which is what makes this reachable, since declaration files are mostly interfaces.

Downstream the twin walks the resolver off a deterministic hit and into a refusal. All three links re-read and confirmed:
```ts
// call-processor.ts:136  — the deterministic hit a twin destroys
const candidates = model.types.lookupClassByName(shortName);
if (candidates.length === 1) return candidates[0];

// call-processor.ts:146  — strips only the LAST extension
const fileBase = candidate.filePath.replace(/\.[^./]+$/, '');

// call-processor.ts:167  — the fallback's bar
return bestScore >= 2 && !tie ? best : undefined;
```
With the twin present `lookupClassByName('User')` returns 2, so the single-candidate fast path at :136 is lost. The path-tail fallback cannot recover it for two independent reasons. First, `fileBase` at :146 leaves `src/service.d.ts` as `src/service.**d**` — not `src/service` — so the twin's trailing path segment can never match an FQN segment. Second, and decisively, the gate at :167 requires `bestScore >= 2`, i.e. the class name **plus at least one matching namespace segment**; a short unqualified name like `User` has exactly one FQN segment, so `bestScore` can never reach 2 no matter which candidate wins. The function returns `undefined`. **A binding that resolved before this PR now drops.** Same `length !== 1 → undefined` shape at `type-env.ts:617,646`.

**Trigger:** any repo that commits in-place declaration output (`tsc` with no `outDir`), or keeps `types/Foo.d.ts` mirroring `src/Foo.ts`, or uses `outDir: "lib"` — note `lib` is deliberately **not** in `DEFAULT_IGNORE_LIST` (only `lib64` is; the repo's own test pins `isHardcodedIgnoredDirectory('lib') === false`). A previously-resolved type binding or route→controller edge silently disappears — the same false-negative class this PR is fixing, arriving through the other door.

**Fix:** keep `.d.ts` ingested (#3039 does ask for `vite-env.d.ts`), but suppress the *twin*: skip `X.d.ts` when `X.ts`/`X.tsx` is in the same directory in the scanned set, or mark ambient defs declaration-only so `lookupClassByName` prefers the implementation instead of counting it toward ambiguity. Independently, `fileBase` at `call-processor.ts:146` should strip `.d.ts` as a unit.

`[reproduced]` for the symbol duplication (real worker-pool parse); `[code-read]` for the resolver chain, re-verified line by line by the review coordinator. Rated **P2 by both lanes that executed it** (ce-correctness, ce-adversarial); a third, code-read-only lane rated it P1 — this digest carries the executed rating. Single-engine: the Codex legs did not reach it.

gitnexus/src/config/ignore-service.ts
```diff
@@ -31,8 +31,10 @@ const DEFAULT_IGNORE_LIST = new Set([
   // 'packages' removed - commonly used for monorepo source code (lerna, pnpm, yarn workspaces)
   'venv',
   '.venv',
-  'env',
   '.env',
+  // Bare `env/` is also a common application-source directory. The
+  // virtual-environment spellings above remain excluded; repositories can
```
magyargergo: **[P2] This comment is factually wrong for the name it is justifying: `env` *is* a virtual-environment spelling — it is the default `python -m venv env` produces.**

"The virtual-environment spellings above remain excluded" refers to `venv`/`.venv`, but the name being removed is the third and most common one, and the replacement prune fires only at depth 0. A polyglot monorepo with the venv one level down (`backend/env`, `services/api/env`) now ingests venv internals. A/B through the real `walkRepositoryPaths`:
```
=== BASE ===        === HEAD ===
src/main.rs         backend/env/pyvenv.cfg
                    backend/env/include/foo.py
                    backend/env/share/man/foo.py
                    backend/env/Scripts/activate_this.py   <-- Windows venv
                    ...
```
`site-packages`, `bin` and `lib64` still match at any depth, so the leak is **bounded** — but `include/`, `share/`, `pyvenv.cfg` and the entire Windows `env\Scripts\` tree are covered by no remaining name. On BASE, zero of these were ingested. Separately, `dist-packages` is not in the list at all, so on a Debian-family host `backend/env/lib/python3/dist-packages/**` third-party source now enters the graph and is returned by `query`/`context` as if it were project code.

**Fix:** at minimum correct the comment — `env` is *not* covered by the spellings above at non-root depth. Better: add a content signal (`pyvenv.cfg` sibling ⇒ prune the whole `env/` at any depth); `isHardcodedIgnoredDirectoryAtPath` already has the directory path in hand. And add `'dist-packages'` next to `'site-packages'` — one line, no interaction with this PR's thesis.

`[reproduced]` — two Claude lanes, separate fixtures.

gitnexus/src/config/ignore-service.ts
```diff
@@ -498,6 +520,11 @@ export const createIgnoreFilter = async (repoPath: string, options?: IgnoreOptio
       if (ig && rel && hasExplicitUnignore(ig, rel) && !ig.ignores(rel + '/')) return false;
       // Hardcoded list: block descent into well-known noise directories.
       if (DEFAULT_IGNORE_LIST.has(p.name)) return true;
+      // Keep conventional root-level artifact trees pruned without suppressing
+      // same-named source modules deeper in a workspace.
+      if (rel && !rel.includes('/') && ROOT_ARTIFACT_DIRECTORIES.has(p.name.toLowerCase())) {
```
magyargergo: **[P2] This copy of the root prune has zero decisive test coverage — deleting the whole block leaves the entire suite green.**

Mutation test, run independently by two lanes: replace this `if` block with a comment (or gate it `if (false && ...)`) and run `test/unit/ignore-service.test.ts` + `test/integration/filesystem-walker.test.ts` → **191 passed (191)**, including the new #3039 walker test.

It survives because `ignored()` at :506 still falls back to `shouldIgnorePath(rel)`, which carries its own copy of the rule and filters every path glob surfaces. So this block is purely a **descent-pruning performance gate** (don't walk a 30k-file venv), and nothing in the suite can tell whether it is present, correct, or has drifted from the other two copies. `test/unit/ignore-service.test.ts` has direct `filter.childrenIgnored(mkPath(...))` assertions for every *other* hardcoded-list and negation rule (lines 320, 326, 337, 346, 352-356, 375-376, 382, 393, 410-414) — none for the new one.

Note also that this is the second of **three** copies of one rule, and the shared helper this PR introduces (`isHardcodedIgnoredDirectoryAtPath`, :358) is called by neither in-module copy. That triplication is what let the case handling at :525 diverge from `DEFAULT_IGNORE_LIST.has(p.name)` one line above it.

**Fix:** add direct assertions in the `describe('createIgnoreFilter')` block, mirroring the existing style — `expect(filter.childrenIgnored(mkPath('generated'))).toBe(true)`, `.childrenIgnored(mkPath('env'))` → true, `.childrenIgnored(mkPath('packages/api/generated'))` → false — plus one `.gitnexusignore` `!env/` negation pair (the negation suite at :297-434 covers only pre-existing list entries). Ideally express all three copies as one `isRootArtifactDirectory(rel, name)` predicate.

`[reproduced]` — same mutation result reached twice, independently.

magyargergo (commented):
## Multi-tool review digest — PR #3043 `fix(ingestion): discover nested source directories`

**Engines.** Claude ran a 6-lane GitNexus swarm plus 4 `ce-code-review` personas (correctness, adversarial, testing, security); Codex `gpt-5.5/xhigh` ran the same swarm scaffolding (B1, full 6-lane pass), a `ce-code-review` leg (B2, one P1), and an adversarial leg (B3, one medium). All three Codex legs were live and produced real findings. A coordinator probe re-ran `shouldIgnorePath` from BASE `031e1237` and HEAD side by side on identical inputs; its rulings override lane claims where they conflict. Base `031e123731b9e335a5931befdd0ea0e4ed886371`, head `2d0499756`, 7 files, +146/−14.

---

### What is solid

- **The fix does what the issue asked for.** #3039's reporter proposed option 1 (root-relative scoping); this implements exactly that. Coordinator probe, BASE → HEAD: `apps/web/src/generated/api.ts` `true → false`, `services/api/env/*` `true → false`, while `env/production.env` and `generated/client.ts` stay `true`. Nested source is discovered; root artifacts stay pruned. `[reproduced]`
- **`.d.ts` inclusion is in scope, not scope creep.** #3039 lists `vite-env.d.ts` among the files it could not find.
- **The new tests are honestly red on base.** Two independent lanes copied HEAD's test files onto BASE source and ran them: **8 failed / 207 passed against BASE source**, no unrelated collateral (HEAD itself is 215/215 — the 207 is a BASE-run figure, not a HEAD one). All 8 new/changed assertions genuinely pin the change. `[reproduced]` (This overrides the swarm test-ci lane, which reached the same verdict by code-reading and did not execute it.)
- **The three copies of the rule agree on every canonical input.** `shouldIgnorePath` (`parts[0]`), `childrenIgnored` (`rel`), and `isHardcodedIgnoredDirectoryAtPath` (`path.relative`) were probed across `env`, `env/`, `env//x.ts`, `EnV/`, backslash separators, trailing separators, `/repo/.`, and `..` segments — no divergence on repo-relative or absolute-with-matching-root input. `[reproduced]`
- **Both new call sites genuinely pass the BFS start dir.** `findTsconfigFiles` (`tsconfig.ts:279-296`) and `loadNodeWorkspacePackages` (`node-workspace-packages.ts:341-368`) both seed the queue with the same `repoRoot` they hand the helper, and their only callers pass the ingestion `repoPath` unchanged. The "scan root diverges from `repoRoot`" hypothesis is refuted, in both engines.
- **Cost is negligible for this repo.** A/B of `walkRepositoryPaths` on the GitNexus worktree: BASE 2206 files, HEAD 2208 — two new `.d.ts`, zero files dropped. `[reproduced]`
- Branch hygiene clean: one commit, no merge commits, 7 causally related files, no `.github/`, Docker, or workflow changes; test fixtures confined to `mkdtemp` with `afterAll` cleanup; no `if`-branching added in tests; no non-determinism.

---

### Headline — the case-sensitivity reversal silently drops tracked root source

`ignore-service.ts:301` and `:525` (and `:366`) lowercase before matching, but `DEFAULT_IGNORE_LIST` — the list these two names were carved **out of** — is matched **case-sensitively** everywhere it is consumed (the segment loop at `:316`, `DEFAULT_IGNORE_LIST.has(p.name)` at `:522`, `isHardcodedIgnoredDirectory` at `:355`). The carve-out is therefore not behaviour-preserving in the capitalised case, and it moves in the **opposite** direction from the PR's own thesis.

Coordinator probe, BASE → HEAD:

```
Generated/client.ts     false -> TRUE
ENV/lib/module.py       false -> TRUE
```

ce-correctness reproduced the same reversal at the walker level, running the real `createIgnoreFilter` + `glob` pair that `filesystem-walker.ts:47-54` uses:

```
--- BASE ---              --- HEAD ---
ENV/settings2.ts          src/generated/graphql.ts
Env/settings.ts           src/index.d.ts
Generated/Client.cs       src/index.ts
src/index.ts              typings/huge.d.ts
```

Three root directories BASE indexed are gone on HEAD. Trigger: any repo with a **tracked** top-level `Generated/`, `Env/`, `ENV/` — a Unity/C# solution with checked-in codegen, a repo with a root `Env/` config module. These are legal, distinct directories on any case-sensitive filesystem.

**Found independently by Codex B3 (Engine B, medium) and reproduced end-to-end by ce-correctness (Engine A, P2) and by the coordinator probe.** Cross-engine + reproduced.

**Both source lanes rated this P2/medium** (ce-correctness F1: P2; Codex B3: medium), and the coordinator probe assigned no P-level. **The escalation to P1 is this digest's own call**, on two grounds:

1. **Silent loss of tracked source is exactly the failure mode #3039 was filed about**, reintroduced in the opposite direction. The issue also explicitly asked for a skip notice — "a silent skip is what made this take a while to find" — and this PR adds a new silent skip rather than a notice.
2. **No test covers either case variant**, confirmed independently by the ce-testing and swarm test-ci lanes.

Fix: drop `.toLowerCase()` from all three sites (`ROOT_ARTIFACT_DIRECTORIES.has(parts[0])`) so the carve-out matches the case-sensitive semantics of the list it replaced, and pin `Generated/x.cs` with a test. If case-insensitive matching is actually wanted, it belongs on `DEFAULT_IGNORE_LIST` too — a much larger, separate change.

The escape hatch does work (`!generated/` + `!Generated/` in `.gitnexusignore` restores both, verified), but a user has no way to know they need it.

---

### Findings posted inline

| # | Sev | File:line | Finding | Corroboration |
|---|-----|-----------|---------|---------------|
| 1 | **P1** | `ignore-service.ts:301` | Case-sensitivity reversal drops root `Generated/`/`Env/`/`ENV/` (headline) | Codex B3 + Claude ce-correctness + coordinator probe · `[reproduced]` |
| 2 | **P2** | `ignore-service.ts:344` | `.d.ts` twin doubles every symbol and defeats `candidates.length === 1`; `fileBase` can't recover it | ce-correctness + ce-adversarial (both **P2**, both executed) · single-engine · duplication `[reproduced]`, resolver chain `[code-read]`, coordinator-verified |
| 3 | **P2** | `ignore-service.ts:36` | The diff's own new comment is factually false: `env` *is* a venv spelling, and the prune is depth-0 only | Claude ce-adversarial + ce-security · `[reproduced]` |
| 4 | **P2** | `ignore-service.ts:525` | `childrenIgnored`'s copy of the rule has zero decisive coverage — deleting it leaves 191/191 green | Claude ce-adversarial + ce-testing, same mutation twice · `[reproduced]` |
| 5 | **P2** | `filesystem-walker.test.ts:246` | The root-prune assertions are inert; exactly ONE assertion in the suite pins the new rule, and `env` has none | Claude ce-adversarial (mutation A: 1 failed/190 passed) · `[reproduced]` |
| 6 | **P2** | `node-workspace-scope.test.ts:99` | `generated/output` is pruned by `output`, not by the new rule — this arm survives mutation C | Claude ce-adversarial · `[reproduced]` |

Three further findings are reported in the body rather than inline, because each lacks a defensible anchor on a line this PR changed: the fabricated Next.js route (fix belongs in `nextjs.ts:27`, an unchanged file), the latent `isHardcodedIgnoredDirectoryAtPath` guard (unreachable from any live call site today), and the tsconfig/workspace scanner's inability to honour a `!generated/` unignore (reproduced as pre-existing on BASE — posting it on a contributor's changed line would imply they introduced it). All three are below, with full evidence.

---

### Lower-priority, unresolved, and out of scope

- **Nested `env/` content exposure — body item, not a blocker.** Files newly reachable under a nested `env/` have their **full raw text** stored uncapped in `File.content` (`csv-generator.ts:278-292`, "intentionally NOT length-capped"), FTS-indexed (`fts-schema.ts:24`), and returned verbatim by MCP `query` — the ce-security lane recovered `AWS_SECRET_ACCESS_KEY=AKIA-SINKPROOF` from a real indexed fixture. `[reproduced]` **But the control test refutes it as a new hole:** the identical file one directory over, `services/api/config/production.env`, is `false` on BASE *and* HEAD — already indexed, content-stored and FTS-searchable before this PR. `.env` is not in `IGNORED_EXTENSIONS` anywhere, so the gap is pre-existing and repo-wide; the `env` list entry protected one directory *name* by coincidence, never secrets. This PR widens it, it does not open it. The swarm security lane rated this P1 and called it "a real security finding" without running that control; that rating is not carried forward. Whether newly-ingested `File.content` also reaches an external embedding endpoint was **disputed between lanes and is not settled** — the cited query lives in `queryFallbackFileNodes` (`embedding-pipeline.ts:194`), a conditional fallback, not the primary path. The durable fix belongs in its own issue: add `.env` to `IGNORED_EXTENSIONS` and make the `.env*` basename test a prefix test.
- **`dist-packages` is missing from the ignore list** (`site-packages` is there, `lib` deliberately is not). A nested venv at `backend/env/` on a Debian-family host now leaks `lib/python3/dist-packages/**` third-party source into the graph. One-line fix, adjacent to this PR. `[reproduced]`
- **The same bug class remains for other any-depth names.** `src/bin/` — Cargo's documented convention for a crate's extra binaries — is still pruned at any depth (`ignore-service.ts:54`), reproduced as absent from both the BASE and HEAD walks. Same for `fixtures`, `snapshots`, `out`, `obj`, `parts`, `cache`, `__tests__`. This PR builds the mechanism (`ROOT_ARTIFACT_DIRECTORIES`) and does not extend it; either extend it to `bin` here or say in the PR body that the remaining ambiguous names are deliberately out of scope, so the next reporter is not told this was fixed. `[reproduced]`
- **[P3] Un-ignoring `.d.ts` fabricates a Next.js Pages-Router route.** Real location `route-extractors/nextjs.ts:27` (an unchanged file), fed by `pipeline-phases/routes.ts:220-240`, which pushes **every** ingested path through the extractor with no extension gate of its own:
  ```ts
  const pagesMatch = normalized.match(/pages\/(api\/.+?)\.(ts|js|tsx|jsx)$/);
  ```
  For `pages/api/types.d.ts` the lazy `.+?` first tries `types`, which forces `(ts|js|tsx|jsx)` to match `d.ts` and fails; it backtracks to `types.d`, `\.` matches, `ts` matches, `$` holds — yielding a `Route` node `/api/types.d` with `source: 'nextjs-filesystem-route'` for a route that does not exist, which can also win the first-writer `!routeRegistry.has(...)` slot at `routes.ts:233`. Reproduced:
  ```
  pages/api/types.d.ts   ->  "/api/types.d"
  pages/api/index.d.ts   ->  "/api/index.d"
  pages/api/users.ts     ->  "/api/users"
  ```
  Until this PR the input could never contain a `.d.ts`, which is why this extractor has no guard — the sibling Expo extractor already does, pinned by `test/unit/expo-routes.test.ts:30` (`expoFileToRouteURL('app/types.d.ts') === null`). Fix: `if (normalized.endsWith('.d.ts')) return null;` at the top of `nextjsFileToRouteURL`. `[reproduced]` (regex behaviour re-verified by the coordinator).
- **[P3, latent] `isHardcodedIgnoredDirectoryAtPath` (`ignore-service.ts:366`) is missing the emptiness guard its two siblings have.** `''.includes('/')` and `'..'.includes('/')` are both `false`, so both pass the "at the root" test and the basename decides. Three-way probe against the `:525` rule and `shouldIgnorePath`:
  ```
  repoRoot        dirPath          atPath  childrenIgnored  shouldIgnorePath
  /repo/env       /repo/env        TRUE    false            false     rel=""    <== DISAGREE
  /repo/env/sub   /repo/env        TRUE    —                —         rel=".."  <== ancestor read as root-level
  ```
  **Not reachable today** — both new call sites seed the BFS with `repoRoot` and only ever pass `path.join(dir, entry.name)` from `readdir`, which never yields `.` or `..`, and symlinked directories are skipped (`entry.isDirectory()` is false for them). Both engines agree on unreachability. It goes live the moment any caller tests the root itself, e.g. a repo checked out into a directory literally named `env` or `generated`. Fix: `if (relative === '' || relative === '..' || relative.startsWith('../')) return false;` before the depth test, matching the `rel &&` guard at `:525`. `[reproduced]` for the behaviour, by two lanes.
- **[P3, pre-existing — the one place the two engines disagreed on severity] The tsconfig/workspace scanners cannot honour a `.gitnexusignore` `!generated/` unignore.** `createIgnoreFilter.childrenIgnored` consults `hasExplicitUnignore` **before** the hardcoded rules (`ignore-service.ts:520`), and `!generated/` does restore `generated/client.ts` to the walk (verified); `isHardcodedIgnoredDirectoryAtPath` never receives an `Ignore` instance and cannot honour it. With `!generated/` set: `walkRepositoryPaths` ingests `generated/**` ✔, `findTsconfigFiles` skips `generated/tsconfig.json` ✘ (its `baseUrl`/`paths` never govern those files), `loadNodeWorkspacePackages` skips `generated/*/package.json` ✘. **Codex B2 made this its only P1** ("root artifact unignore misses config scans", `tsconfig.ts:295`), and the Claude swarm risk lane separately claimed this PR *activates* the asymmetry. The digest follows the reproduced result instead: ce-adversarial ran `.gitnexusignore` with `!generated/` + `!env/` against both sides and found the behaviour **byte-identical on BASE** — on BASE these scanners called `isHardcodedIgnoredDirectory(entry.name)`, which had no escape hatch either and also contained `env`/`generated`. **Not a regression**, so it is not posted on a changed line. Still worth an issue: this PR makes it the third copy of one rule, and the PR's own test (`test/unit/node-workspace-scope.test.ts:93-106`) now asserts the override-less behaviour as *desired* — an explicit `workspaces: ['generated/*']` declaration loses to a directory-name convention with no way out. Fix: thread the `Ignore` instance into `isHardcodedIgnoredDirectoryAtPath` (it already takes `repoRoot`, so `loadIgnoreRules(repoRoot)` is one hop away), or collapse all three copies into one `shouldIgnoreDirectory(repoRoot, absDir, ig?)`. `[reproduced]`
- **`wiki/generator.ts:1024` receives git-root-relative paths.** `git diff --name-only` at `:1156` prints paths relative to the git top level regardless of `cwd` (reproduced twice, in two lanes). When the indexed repo is a subdirectory of its git repo — the GitNexus monorepo is exactly this shape — `parts[0]` is `gitnexus`, and the new depth-0 prune never fires. The mismatch predates this PR; the PR removes the any-depth segment loop that hid it. Worth `--relative` at `:1156` as a follow-up. Two Claude lanes reproduced the git behaviour; the swarm risk lane refuted it by code-read — the reproduced result wins.
- **Test gaps beyond the inline ones:** no `.gitnexusignore` negation test pairs `!env/`/`!generated/` with a root-level directory (flagged independently by **Codex B1 lane 4 and two Claude lanes** — the one test gap with cross-engine agreement); `isHardcodedIgnoredDirectoryAtPath` is tested only with `generated`, never `env`, at either call site; and the `.d.ts` un-ignore — the widest blast-radius line in the diff — has **no** extractor- or graph-level test anywhere in the repo proving a real `.d.ts` parses cleanly into symbols. Everything asserted about `.d.ts` today stops at the walker's boolean.
- **PR body test count doesn't reconcile.** The body says "201 affected tests passed; 2 platform-specific tests skipped". The four touched files hold **215** tests, all passing, 0 skipped on a non-root Linux runner. The 2 skips are pre-existing EACCES tests (`ignore-service.test.ts:644`, `:845`, gated `skipIf(win32 || getuid()===0)`), unrelated to #3039. Cosmetic, but the number is wrong as stated.
- **Unverified open question:** neither `ANALYZER_RUNNER_IDENTITY_SCHEMA_VERSION` nor `IDENTITY_CACHE_SCHEMA_VERSION` is bumped. Whether a release alone changes runner identity — and therefore whether newly-discoverable `src/env/` and `.d.ts` files appear after an incremental analyze or need a forced rebuild — was **not verified** by any lane. Worth a maintainer answer before merge.

### Validated and refuted — checked, found sound

Reported so nobody re-spends the turns:

1. **Path-shape / absolute-path bypass of `shouldIgnorePath`.** The divergence is real inside the function (`"/repo/env/x.ts"` → false, `"./env/x.ts"` → false), but **no live caller produces those shapes**: `createIgnoreFilter.ignored` passes path-scurry's `p.relative()`; the six group workspace extractors build `childRel` as `` rel ? `${rel}/${name}` : name `` from `walk(repoPath, '')`; `wiki/generator.ts:349` consumes `File.filePath`, written repo-relative POSIX at `structure-processor.ts:24`. Refuted in both engines (the git-root-prefix variant above is a different mechanism).
2. **`path.relative` returning `'..'` as a traversal weakness.** Refuted as a security issue with the exact reason: the return is a conjunction, and `basename('/a')` is not in `{env, generated}`. The only shape satisfying both over-ignores a directory outside the repo — it can never admit a path BASE rejected. Symlinks are not an escape either: `readdir(withFileTypes)` reports a symlinked directory as `isDirectory=false`, and both loops gate on `isDirectory()`.
3. **The `!env/` escape-hatch asymmetry as a *regression*.** Refuted — byte-identical on BASE (see the pre-existing scanner limitation in the body above).
4. **A `.d.ts` out-ranking its `.ts` in import resolution.** Refuted. `TS_EXTENSIONS` is ordered `.ts, .tsx, .d.ts` and `JS_TO_TS.get('.js')` is `['.ts','.tsx','.d.ts']`, so `resolveFile` lands on a `.d.ts` only when no `.ts`/`.tsx` sibling is indexed — the correct answer. Likewise a `.d.ts` cannot steal the workspace `exportedCallableByName` slot: declaration-file functions parse as `function_signature`, not `function_declaration`, so they never become `Function` defs. The damage from the twin is duplicate *nodes* and short-name ambiguity (inline #2), not misrouted imports.
5. **Nested venvs blowing up the index.** Refuted as a blow-up — `site-packages`, `bin`, `lib64` still match at any depth, so the leak is bounded (see inline #3 for what does leak).
6. **A caller passing a subdirectory as `repoRoot`.** Refuted at both new call sites, in both engines.
7. **Other hardcoded skip lists needing the same update.** None found — `isHardcodedIgnoredDirectory` had exactly two production call sites and both were converted; no other scanner ever listed `env`/`generated`.
8. **Incremental/shadow-candidate handling of `.d.ts`.** Refuted as a defect — `shadow-candidates.ts:38` already lists `.d.ts` first and the behaviour is pinned by an existing test.
9. **Withdrawn:** the swarm security lane reported stale entries in `gitnexus/src/config/__base_ignore_service_probe.ts`. **That file does not exist** — the worktree's `git status` is clean. Dropped entirely.

### CI

`CI Gate` **passed**. The only failing check is Vercel — "Authorization required to deploy" — which is an infrastructure/authorization failure, not a code failure. Confirm it is non-required (or authorize it) before merge. GitHub reports `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE` solely because of that check.

---

*This is an automated multi-tool review digest (Claude swarm + `ce-code-review` personas, Codex gpt-5.5/xhigh swarm + ce + adversarial, merged with cross-engine weighting). Every finding carries its trigger, evidence, and a `[reproduced]`/`[code-read]` tag — verify before acting on any of them.*

gitnexus/test/integration/filesystem-walker.test.ts
```diff
@@ -187,6 +187,67 @@ describe('filesystem-walker', () => {
...
+
+    it('discovers nested env/generated and .d.ts source while pruning root artifacts', async () => {
+      const files = await walkRepositoryPaths(sourceDir);
+      const paths = files.map((file) => file.path);
+
+      expect(paths).toContain('apps/client/src/shared/env/getAppEnv.ts');
+      expect(paths).toContain('packages/ai/src/generated/bundle.ts');
+      expect(paths).toContain('apps/client/src/vite-env.d.ts');
+      expect(paths).not.toContain('build-cache/generated/ignored.ts');
+      expect(paths).not.toContain('env/lib/python3.12/site-packages/pkg/module.py');
```
magyargergo: **[P2] This "root artifact stays pruned" assertion is inert — `site-packages` already prunes the path, so it passes with the new rule deleted. Across the whole suite exactly ONE assertion actually pins the root prune, and `env` has none.**

`env/lib/python3.12/site-packages/pkg/module.py` is caught by `site-packages` (`ignore-service.ts:41`), which still matches at any depth. Mutation A — delete the root prune from `shouldIgnorePath` (:301-303) — and run this file plus `test/unit/ignore-service.test.ts`:
```
FAIL  ... > prunes ambiguous artifact directories only at the repository root: generated/client.ts
      Tests  1 failed | 190 passed (191)
```
Only the `generated/client.ts` arm goes red. The `env` half of `ROOT_ARTIFACT_DIRECTORIES` — the half this PR's own risk narrative is about (venvs) — is pinned by nothing: delete `'env'` from the Set and the suite stays green. The same inert path is asserted in the `it.each` at `test/unit/ignore-service.test.ts:245`.

**Fix:** use a name that is not independently ignored — `env/pyvenv.cfg` plus `env/settings.py` — so the assertion fails when the root rule does.

`[reproduced]`

gitnexus/test/unit/node-workspace-scope.test.ts
```diff
@@ -90,6 +90,22 @@ describe('workspace boundary', () => {
     expect(packages?.byName.has('@repo/web')).toBe(true);
   });
 
+  it('prunes root artifact workspaces while keeping nested source directories', async () => {
+    const root = repo({
+      'package.json': JSON.stringify({
+        name: 'root',
+        workspaces: ['generated/*', 'packages/*/generated'],
+      }),
+      'generated/output/package.json': pkg('@repo/root-artifact'),
```
magyargergo: **[P2] This fixture is pruned by `output`, not by the rule under test — the assertion at :105 survives deleting the new root clause.**

`output` is already in `DEFAULT_IGNORE_LIST` (`ignore-service.ts:53`), so `generated/output/package.json` never depended on `isHardcodedIgnoredDirectoryAtPath`'s root branch. Mutation C — that branch replaced with `return false` — run over this file and `tsconfig-index.test.ts`:
```
FAIL  test/unit/tsconfig-index.test.ts > prunes root artifact configs ...
      Tests  1 failed | 23 passed (24)      <-- this file's arm PASSES with the rule gone
```
Direct probe with the fixture renamed to a neutral name confirms it:
```
MUTATION C: [ 'root', '@repo/root-artifact-REAL', '@repo/generated-source' ]
HEAD:       [ 'root', '@repo/generated-source' ]
```

**Fix:** rename the fixture to a directory not otherwise ignored — `generated/apiclient/package.json` — which makes the arm decisive. (The sibling `tsconfig-index.test.ts` arm is already decisive; only this one is inert.)

`[reproduced]`

azizur100389:
Addressed the review in 2594e639: preserved exact-case root behavior, added path-aware nested-venv pruning across ingestion and Python group scans, made the traversal assertions decisive, and suppressed declaration twins for .ts/.mts/.cts while retaining standalone declarations. Focused tests, typecheck, lint, format, diff checks, and change-impact validation pass.

Issue #3041, opened 2026-08-25
Same-named nested functions (Convex handler:) collapse to one node; CALLS from every handler are re-attached to the first, and downstream impact returns 0


AndelskeSluzby:
### Area

gitnexus (CLI / core / indexing / MCP server)

### Summary

In a file declaring several `export const x = query({ handler: async (…) => … })`, all the `handler` arrow functions collapse into a **single** graph node, and every `CALLS` edge from every handler is re-attached to that one node — the first one in the file.

Two consequences, the second worse than the first:

1. `impact(direction: "downstream")` on any of the exported functions returns `0`, because none of them owns the node the edges hang off.
2. The surviving node is credited with calls made by *other* functions. This is a wrong edge at full confidence, not a missing one — the same shape as #3038.

Duplicate callees across handlers are also deduplicated, so a helper called by four of the nine functions appears once.

### Context

Convex declares every endpoint as an exported const wrapping an object with a `handler` property. Nine exported functions in a file therefore mean nine arrow functions all named `handler`. It looks like node identity is keyed on `(name, filePath)`, so they collide and the first wins — but that is a hypothesis from the outside; I have only observed the result.

This is separate from #3040 (which is about `epistemic` vouching for the empty *upstream* answer). Together they mean a Convex symbol reports `0` in both directions.

### Expected behavior

Each `handler` gets its own node, with its own `CALLS` edges, and `impact(direction: "downstream")` on the exported const reaches them.

### Actual behavior

`convex/orders/get.ts` — 9 exported functions, 9 `handler:` properties in source:

```bash
$ gitnexus cypher "MATCH (n) WHERE n.filePath = 'convex/orders/get.ts' AND n.name = 'handler' RETURN count(*) AS n"
# 1
```

That one node is at line 19, inside the first export (`getById`). It carries three `CALLS`:

```bash
$ gitnexus cypher "MATCH (a)-[r:CodeRelation {type:'CALLS'}]->(b) WHERE a.filePath = 'convex/orders/get.ts' RETURN a.startLine AS from_line, b.name AS callee"
# 19 | requireAdmin
# 19 | dicByGraves
# 19 | resolveOrderWpidForOrder
```

Where those calls actually live in source:

```
line  21  inside getById               -> requireAdmin
line  38  inside getAll                -> requireAdmin
line  53  inside getByGrave            -> requireAdmin
line  75  inside getInvoicesByOrder    -> requireAdmin
[...]

PR #3017, opened 2026-08-20
fix(ingestion): index NestJS decorator routes so api_impact and route_map stop reporting live endpoints as non-existent

ReidenXerx:
## Summary

Fixes #3009. The indexer never extracted NestJS `@Controller` / `@Get` routes, so a NestJS repo produced essentially no `Route` nodes — and `route_map`, `api_impact` and `shape_check` all read `Route` nodes and answer "no routes matching …" when there are none.

On a production NestJS backend, `Route` nodes go from **3 to 212**, and `route_map({route: "/venues"})` goes from `No routes matching "/venues"` to 26 routes with resolved handlers.

## Motivation / context

TypeScript's `extractDecoratorRoutes` hook was, in full (`languages/typescript.ts:130`):

```ts
const extractJsTsRoutes = (...args: Parameters<typeof extractDispatchGuardRoutes>) => [
  ...extractDispatchGuardRoutes(...args),
  ...extractDataRouteTableRoutes(...args),
];
```

Dispatch guards and static data route tables only. Java wires `extractDecoratorRoutes: extractSpringRoutes`, but nothing in `route-extractors/` handled NestJS — there was no `nest.ts`, and no file in that directory mentioned Nest or `@Controller`.

`ARCHITECTURE.md` said otherwise, which is probably why the gap survived:

> | AST-level route in a normal file | `extractDecoratorRoutes` | Spring, FastAPI, **NestJS**, JS/TS dispatch guards and static data route tables |

NestJS was documented and not implemented. That row is corrected here.

**Why this is worse than a missing feature.** The tools do not degrade politely — they assert absence. On a backend with 33 `@Controller` classes and 209 route decorators, the index held 3 `Route` nodes, and all 3 were URL string literals scraped out of utility code (a Bull Board mount, a parsed YouTube URL, an image-URL normaliser). Meanwhile `api_impact`, documented as the thing to run *"BEFORE modifying any API route handler"*, answered "No routes found matching …" for every live endpoint — and a not-found reads as a safe change. That is a false-safe on a pre-change safety check.

## Areas touched

- [x] `gitnexus/` (CLI / core / MCP server — ingestion, the routes phase, and the group-contract HTTP scan)
- [ ] `gitnexus-web/` (Vite / React UI)
- [ ] `.github/` (workflows, actions)
- [ ] `eval/` or other tooling
- [x] Docs (`ARCHITECTURE.md` route-source table)

## Scope & constraints

**In scope**

[...]

PR #3044, opened 2026-08-25
fix(impact): mark Convex caller results incomplete

### LINKED ISSUE #3040: impact() reports epistemic: "exact" with 0 callers for Convex functions, where anyApi indirection leaves no edge for the #1858 probe to key on
### Area

### Summary

`impact()` on a Convex function returns `impactedCount: 0` with `epistemic: "exact"` and no `causes` or `boundaries`, while the symbol has live call sites. This is the same shape as #2708 / #2744 / #2766 / #2813 / #2873, but from a cause the #1858 probe cannot currently see: the indirection leaves **no edge and no drop record at all**, so there is nothing for it to key on.

Closest open sibling is #3015 — `impact` fail-opening when analyze dropped a file. Same end state (a confident answer that is silently incomplete), different trigger.

### Context

Convex addresses its backend functions through `anyApi`, a runtime Proxy:

```js
// convex/_generated/api.js
import { anyApi } from 'convex/server';
export const api = anyApi;
```

A caller writes `useMutation(api.client.eshop.updateDraft)` — a property-access chain on an opaque object. Nothing statically links that expression to `convex/client/eshop.ts`, so no `CALLS` or `IMPORTS` edge is emitted. That part is expected and not what I am filing; it is a property of how Convex addresses functions, and I do not think a static analyser can fix it.

What I am filing is that the result does not say so. `epistemic` exists to carry exactly this signal, and here it actively vouches for the empty answer.

### Expected behavior

`epistemic: "lower-bound"` for a symbol whose only callers bind through a dynamic-dispatch boundary, with a boundary note naming the cause — the same treatment an interface-typed caller gets today.

### Actual behavior

```
$ gitnexus impact updateDraft -f convex/client/eshop.ts -d upstream --summary-only
{
  "impactedCount": 0,
  "risk": "UNKNOWN",
  "epistemic": "exact",
  "summary": { "direct": 0, "processes_affected": 0, "modules_affected": 0 }
}
```

Ground truth — 4 production call sites, plus 5 in tests:

```
apps/client/src/features/eshop/EshopUnlockedOrder.tsx:113
apps/client/src/features/eshop/EshopGraveIdentityStep.tsx:90
apps/client/src/features/eshop/wizard/useEshopWizardDraft.ts:27,89
```

No `causes` and no `boundaries` key is present — not a zeroed count, the fields are absent.

### Steps to reproduce

Any Convex repo. In ours (private, 1,928 TS files) every public endpoint reproduces it:

```bash
gitnexus impact <convexFnName> -f convex/<module>.ts -d upstream --summary-only
grep -rn "api\.<module path>\.<convexFnName>" apps convex packages   # the real callers
```

The graph confirms nothing inbound exists — only structural edges:

```cypher
MATCH (a)-[r:CodeRelation]->(b) WHERE b.filePath = 'convex/clients/create.ts'
RETURN r.type AS type, count(*) AS n
-- CONTAINS 1 | DEFINES 4
```

### Environment

- gitnexus `1.6.10-rc.215`
- macOS 15 (darwin arm64), Node 24
- TS monorepo, npm workspaces, Convex backend

### Notes

**Why the probe misses it.** `computeEpistemicBoundary` looks for materialized evidence of indirection — `EPISTEMIC_HERITAGE_RELATION_TYPES` (`IMPLEMENTS` / `METHOD_IMPLEMENTS` / `EXTENDS`) plus unresolved-receiver notes. When it finds no boundary node it returns `epistemicFrom(droppedBoundaries)`, which is bare `exact` if there are no notes. `anyApi` produces neither an edge nor an unresolved-receiver record, so the probe correctly finds nothing and reports `exact`.

The probe is additive and fail-open by design — it returns `exact` on query error too, so it can only add signal. That is right for a probe, but it means *"no evidence of indirection"* and *"indirection that leaves no evidence"* collapse into the same answer, and the second case is the dangerous one.

**Suggested fix.** Detect the Convex wrapper shape rather than the `convex/` path — `export const X = query|mutation|action|internalQuery|internalMutation|internalAction({ ... })`. That is more precise than a directory convention and survives repos that put functions elsewhere. Framework-aware detection already lives in the same function (`queryClassBeanMetadata`, `querySpringAopMetadata`), so this would sit alongside rather than introduce a new concept.

I would reuse the existing `lower-bound` value rather than adding a `partial` — it already means what is needed here, and needs no consumer changes.

**One thing that compounds it:** `direction: "downstream"` on the same symbols also returns 0, because the body's `CALLS` attach to the inner `handler` node rather than the exported `Const` (and only the first `handler` per file appears to get a node — a file of ours with 5 exported queries and 4 `requireAdmin` calls has 1 handler node and 1 `CALLS` edge). So `epistemic: "exact"` is currently vouching for two empty answers rather than one. Happy to file that separately if it is not already known — say the word and I will, with a reduced case.

azizur100389:
## Summary

- Extract and persist import-proven Convex endpoint metadata for TypeScript and JavaScript.
- Report upstream impact and context as lower-bound while keeping downstream impact exact.
- Treat old indexes conservatively and document the provider hook; refresh with `gitnexus analyze --force` after upgrade or rollback.

## Validation

- 67 focused tests passed.

Fixes #3040

---

## Summary

*This appears to be a broad, cross-cutting Convex impact change that updates how caller-result completeness is represented across graph ingestion, persistence, and local impact analysis. Its reach is critical because the change crosses core graph infrastructure and propagates through a wide dependent surface.*

**🔴 CRITICAL blast radius. A Convex impact and epistemic metadata change spanning ingestion, Lbug persistence, and local MCP analysis, with broad reach through the dependent graph.**

The main concentration is in `gitnexus/src/core/lbug/lbug-adapter.ts`, alongside `gitnexus/src/core/lbug/schema.ts`, `gitnexus/src/core/ingestion/languages/typescript.ts`, `gitnexus/src/mcp/local/convex-metadata.ts`, and `gitnexus/src/mcp/local/local-backend.ts`. Review `getCopyQuery`, `insertNodeToLbug`, `batchInsertNodesToLbug`, `FUNCTION_SCHEMA`, `VARIABLE_SCHEMA`, `PROPERTY_SCHEMA`, `epistemicFrom`, and `countByType` first, then trace how `extractJsTsRoutes` and `processFileGroup` feed the Convex metadata path.

The affected module areas are `Local`, `Lbug`, `Search`, `Integration`, `Ingestion`, `Cli`, `Mcp`, `Languages`, and `Server`, with 29 affected flows and 89 graph dependents. The test coverage is concentrated in the Convex impact, epistemic lower-bound, metadata persistence, dispatch, definition-properties, and incremental parse-cache tests.

_Added by GitNexus for PR #3044. Edit freely — this block is replaced on the next review, everything above it is left untouched._


azizur100389:
Addressed the automated findings in bbc279c5: Const CREATE/MERGE now match the declared schema, and the contract tests cover schema, COPY, CREATE, and MERGE alignment.

Validated: 67 affected tests, typecheck, ESLint, and Prettier.

gitnexus/src/mcp/local/convex-metadata.ts
```diff
@@ -0,0 +1,58 @@
...
+  runQuery: typeof executeParameterized = executeParameterized,
+): Promise<ConvexDispatchMetadata | undefined> {
+  if (symbolType !== 'Const') return undefined;
+
+  try {
+    const rows = await runQuery(
+      lbugPath,
+      `MATCH (n:Const {id: $symbolId})
+       RETURN n.convexEndpointFactory AS factory
+       LIMIT 1`,
```
magyargergo: **P1** · `reproduced` · engines: A+B

This is the single red test in CI. The repo's #2787 determinism guard (`test/unit/query-determinism-guard.test.ts`) names exactly this query and nothing else: ``src/mcp/local/convex-metadata.ts:30 [undeclared] `MATCH (n:Const {id: $symbolId}) RETURN n.convexEndpointFactory AS factory LIMIT 1` `` — reproduced locally by four lanes (1 failed / 20 passed) and matching the `tests / ubuntu / coverage 2/3` failure. The query is genuinely single-row, since `id` is `Const`'s declared PRIMARY KEY, but the guard deliberately never infers that from shape ("this gate does not infer; it makes the author DECLARE, once"). Your direct sibling `bean-metadata.ts:44-51` — same PK-anchored `LIMIT 1`, same module family, same fail-soft idiom — already carries the declaration. Suggested fix: add `// determinism: probe — id is the Const PRIMARY KEY, so at most one row can match {id: $symbolId}` immediately above the query, or simply drop the inert `LIMIT 1`. Please don't add an `ORDER BY` — that puts a sort on a single-row primary-key lookup.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -6821,10 +6824,19 @@ export class LocalBackend {
...
     const droppedBoundaries = {
       ...receiverDrops,
-      notes: [...receiverDrops.notes, ...undecidedDrops.notes],
+      notes: [
+        ...receiverDrops.notes,
+        ...undecidedDrops.notes,
+        ...(convexDispatch === undefined ? [] : [convexDispatch.boundary]),
+      ],
       undecided: undecidedDrops.undecided,
+      dispatch: convexDispatch === undefined ? 0 : 1,
```
magyargergo: **P1** · `reproduced` · engines: A

This line cannot distinguish "confirmed Convex endpoint" from "this index predates the column", and that makes the change regress the very property it exists to protect on every not-yet-re-analyzed index. On a pre-column index the probe's query throws `Binder exception: Cannot find property convexEndpointFactory for n.`, `queryConvexDispatchMetadata` returns a *defined* `{staleIndex: true, boundary: …}`, so `dispatch` becomes `1`, the note lands, and `epistemicFrom` returns `lower-bound`. Trigger: upgrade, do not re-analyze, then `impact({target:"<any exported const>", direction:"upstream"})` — or `context()`, which passes no `direction` and therefore also probes. A plain `const MAX_RETRIES = 3` in a Go, Rust or Ruby repo with zero Convex code comes back `lower-bound` with `causes.dispatchBoundary: 1` and a Convex re-index advisory; `Const` is a multi-language label, and this repo's own index has 6,738 of them. Two lanes reproduced this end-to-end by building a real pre-PR 7-column `Const` table and running the shipped query through the real pool; there is no schema-version gate on the MCP read path (`schemaFingerprintMismatch` is consumed only by `core/run-analyze.ts`), so old `.lbug` files are served, not rejected. Separately, even on a genuine endpoint the contribution is a constant `1` summed at line 6935 into `dispatchBoundarySymbols`, where every other contributor counts *other* symbols — the comment at 6897-6901 is explicit that putting a presence value next to a real magnitude is the mistake already fixed once in this function. Suggested fix: don't fold `staleIndex` into `dispatch` (it is a re-index condition, and `staleIndex` today has no production reader at all — this would give it one), and for the real-endpoint case either leave `dispatchBoundary` at 0 and rely on the note, which already forces `lower-bound`, or add a distinct boolean cause rather than adding to a counted field.

gitnexus/src/core/ingestion/languages/typescript/convex-endpoint-metadata.ts
```diff
@@ -0,0 +1,70 @@
...
+  const value = declarator?.childForFieldName('value');
+  if (!value || value.type !== 'call_expression') return undefined;
+
+  const callee = value.childForFieldName('function');
+  if (!callee || callee.type !== 'identifier') return undefined;
+  const factory = importedConvexFactory(context.parsedImports, callee.text);
+  if (factory === undefined) return undefined;
+
+  const args = value.childForFieldName('arguments');
+  if (!args || args.namedChild(0)?.type !== 'object') return undefined;
```
magyargergo: **P2** · `reproduced` · engines: A+B

The object-literal gate excludes Convex's documented UnvalidatedFunction form, so the PR's thesis bug stays live for it. `QueryBuilder` accepts either a config object (`{args?, returns?, handler}`) or a bare handler function `(ctx, ...args) => ReturnValue`, and the official docs show `export const func = query(({ db }, { arg }) => {…})`. Driven through the real extractor with import provenance proven from `./_generated/server`: object form → `{"convexEndpointFactory":"query"}`, bare arrow and bare function → not stamped. Trigger: `export const list = query(async (ctx) => ctx.db.query("messages").collect());` then `impact({target:"list", direction:"upstream"})` → still `epistemic:'exact'`, `dispatchBoundary: 0`. One lane found the harder half: when the argument is an arrow function the same declaration also matches a `@definition.function` pattern, `buildDefinitionPreScan` records it in `nonValue`, and the `Const` twin is dropped entirely — so `convex-metadata.ts:23`'s `symbolType !== 'Const'` gate could never see it even with this line relaxed. Suggested fix: accept a single `object` / `arrow_function` / `function_expression` argument here (keep rejecting `query(buildConfig())`, which your negative test covers) **and** widen the consumer gate so it accepts the `Function` label as well as `Const`. If that is out of scope for this PR, please state the limitation in the PR body and in `tools.ts`, so an `exact` verdict on a Convex repo is not read as a positive assertion.

gitnexus/src/core/ingestion/languages/typescript/convex-endpoint-metadata.ts
```diff
@@ -0,0 +1,70 @@
...
+  'mutation',
+  'action',
+  'internalQuery',
+  'internalMutation',
+  'internalAction',
+]);
+
+const isConvexServerModule = (targetRaw: string): boolean => {
+  const normalized = targetRaw.replace(/\\/g, '/').replace(/\.(?:[cm]?[jt]s)$/, '');
+  return normalized === 'convex/server' || /^(?:\.\.?\/)+_generated\/server$/.test(normalized);
```
magyargergo: **P2** · `reproduced` · engines: A+B

The `convex/server` arm and `CONVEX_ENDPOINT_FACTORIES` accept the wrong half of the cross-product. `convex/server` exports the `*Generic` builders (`queryGeneric`, `mutationGeneric`, `actionGeneric`, `httpActionGeneric`, …); the bare `query`/`mutation`/… names exist only in the codegen'd `./_generated/server`, which re-exports them as typed aliases. Measured through the real extractor: `import { queryGeneric } from "convex/server"` → not stamped (real API, missed); `import { query } from "convex/server"` → stamped (an import that cannot occur in a real Convex project). That second row is what `test/unit/convex-metadata.test.ts:57-64` ("direct convex/server provenance") asserts, so the headline unit test for this arm is green over an unreachable input, while Convex components and library packages — which cannot use codegen and therefore *must* use the `*Generic` builders — are entirely uncovered. Four related gaps confirmed by running them: `httpAction` is a genuine `_generated/server` export and is absent from the set; `convex/_generated/server` (the spelling Convex's own Next.js quickstart uses, legal under `"baseUrl": "."`) is rejected and the rejection is enshrined in a test; alias specifiers like `@/convex/_generated/server` are rejected; and `convex-helpers`' `customQuery`/`customMutation` wrappers never stamp. Suggested fix: add the `*Generic` names (normalising each to its canonical factory for the note text) plus `httpAction`, relax the specifier test so it anchors on a trailing `_generated/server` segment after normalisation (segment boundary or slash, not just a relative prefix), and retarget the `convex/server` fixture at a shape that can actually exist.

gitnexus/src/mcp/local/convex-metadata.ts
```diff
@@ -0,0 +1,58 @@
...
+  } catch (error) {
+    if (isMissingConvexMetadataProperty(error)) {
+      return {
+        staleIndex: true,
+        boundary:
+          'Convex runtime-proxy metadata is unavailable because this index predates ' +
+          'convexEndpointFactory; re-index before treating impact as exact.',
+      };
+    }
+    // Additive evidence only. Transient failures must not abort impact.
```
magyargergo: **P2** · `code-read` · engines: A

On this branch a failed probe and a proven negative become the same value. This comment states the principle; the `return undefined;` on the next line is what breaks it, because `undefined` is exactly the "not a Convex endpoint" signal at `local-backend.ts:6839`, so any DB failure whose wording isn't the one recognised form restores `epistemic:'exact'` on a genuine Convex endpoint — the false confidence this PR exists to remove. The triggers are real strings from `pool-adapter.ts`, none of which match `isMissingConvexMetadataProperty`: `"Connection pool exhausted: timed out after 15000ms waiting for a free connection"`, `"Prepare timed out after 30000ms"`, `"Execute timed out after 30000ms"`, and `"LadybugDB not initialized for repo …"` (pool entry evicted between the repo handle and this probe). Contention here is designed-for rather than exotic: `computeEpistemicBoundary` already fans out up to 50 concurrent queries against an 8-connection pool on this same path. Catching *is* mandatory — the returned promise must never reject, since `local-backend.ts:4423` and `:7657` await it inside a bare `Promise.all` — but "must not abort" and "must not read as certainty" are two different requirements, and the enclosing function states the second one twice in its own comments ("a probe failing must never read as certainty — the whole point of this function"). Suggested fix: return a third state here (e.g. `{ probeFailed: true, boundary: 'Convex endpoint metadata could not be read for this symbol; impact may be incomplete.' }`) and have the caller append the note without incrementing `dispatch`; at minimum route the swallowed error through `logQueryError` so the case is diagnosable after the fact.

magyargergo (commented):
## Tri-review digest — PR #3044 (`fix(impact): mark Convex caller results incomplete`)

Head `bbc279c` · base `031e123` · 18 files, +688/−21.

### How this was reviewed

**Engine A (Claude) — 12 lanes, all live.** Four `gitnexus-*` swarm lanes (risk-architect, security-boundary, test-ci-verifier, docs-DoD) plus eight CE personas (correctness, adversarial, reliability, testing, api-contract, performance, maintainability, project-standards). **Caveat:** the `compound-engineering:ce-*` subagents were not registered in this session, so the eight CE lanes ran as `general-purpose` agents carrying the CE persona briefs verbatim — same methodology, not the packaged agents.

**Engine B (Codex gpt-5.5) — 3 legs, all live and all produced real findings.** B1 = full 7-persona `gitnexus-pr-review` swarm. B2 = `ce-code-review` report-only — **its sandbox blocked Node and Python, so it is code-read only** (no `tsc`, no vitest, no sub-reviewers); weighted lower accordingly. B3 = native adversarial-review.

Unusually for this format, a large share of the work was **executed, not read**: lanes built real LadybugDB databases (including a hand-built pre-PR 7-column `Const` DDL), ran the real COPY path, drove the real tree-sitter TypeScript provider over ~40 Convex source shapes, ran the repo's own determinism guard, ran the epistemic integration suite, and measured `findDeclarator` and DB round-trip costs against live indexes. Findings below are tagged `[reproduced]` or `[code-read]` on that basis.

### Credit — what was verified to work

This is a well-built change and a lot of the scariest hypotheses died on real execution:

- **The feature is live, not dead code.** A lane ran the real TS provider + real `TYPESCRIPT_QUERIES` + real `getLabelFromCaptures` over a `convex/messages.ts`: label is `Const`, `parsedImports` carries the `./_generated/server` provenance, and the extractor returns `{convexEndpointFactory:"query"}`. The `labels(n)[0]`-is-always-`""` hazard does not apply — the impact path reads `outcome.resolvedLabel || outcome.symbol.type` (local-backend.ts:6494). `[reproduced]`
- **The 8-column `Const` persistence chain is consistent end to end** — `CONST_SCHEMA`, `constHeader`, the row writer, `getCopyQuery`, `CREATE` and `MERGE` all agree at 8 fields in the same order. One lane ran the real `loadGraphToLbug` → `streamAllCSVsToDisk` → `COPY` → read-back with a CSV-hostile cell (comma + quote + newline in `content`): all rows land, empty cell reads back as `NULL`, and `String(row.factory ?? row[0] ?? '')` handles it. `GraphEmitSink` streams relationships only, so there is no second node-CSV writer emitting 7 fields. `[reproduced]`
- **The warm parse-cache round-trip is real.** `slimParseWorkerResultsForCache` spreads `...result` and blanks only calls/assignments/etc., so `properties.convexEndpointFactory` round-trips generically — the `SCHEMA_BUMP` really is required, and the e2e test asserts warm replay. `[reproduced]`
- **`SCHEMA_FINGERPRINT` handles the DDL half automatically.** `CONST_SCHEMA` is in `NODE_SCHEMA_QUERIES`, so editing it changes the fingerprint, and `run-analyze.ts:1438-1476` forces a full rebuild on mismatch before the `alreadyUpToDate` fast path. The "old 7-column table survives a re-index" hazard does not occur. `[code-read]`
- **The epistemic wiring composes correctly.** `test/integration/impact-epistemic-lower-bound.test.ts` **15/15 green**, including all six factories reporting `lower-bound` with `dispatchBoundary === 1` and three negative controls staying `exact`. The new cause is added alongside receiver/undecided/external rather than shadowing them. `[reproduced]`
- **The tests are load-bearing.** Two independent mutation checks went RED (removing the import-provenance check; removing `dispatch: … ? 0 : 1`). All 42 new unit tests pass, and every one imports a symbol or asserts a string that does not exist on base — none would pass on BASE with the production change reverted. `[reproduced]`
- **Security is clean across five probes** — Cypher injection (double-defended: six-name ASCII allowlist plus `escapeCypherString`'s correct backslash-then-quote order), CSV corruption, attacker-influenced error text (the probe error is a *compile*-time binder failure, raised before `$symbolId` is ever bound), recursion, and hygiene: zero bidi controls, clean `git diff --check`, **no raw NUL bytes** in any of the 8 materially-changed files (this repo has had four prior NUL incidents; this PR does not add a fifth).
- **Repo-rule conformance:** the AGENTS.md "shared `core/ingestion/` must not name languages" rule **passes cleanly** — the hook is generic, `parse-worker.ts` does a blind `Object.assign` and never branches on Convex, and the only Convex-aware file sits under `languages/typescript/`. No `any` added (the new e2e test uses a proper intersection cast where its sibling uses `as any`), no if-branching in the new vitests, no non-determinism, CHANGELOG correctly untouched, conventional title. lint/format/typecheck all green.

### Problem being solved

Issue #3040: a Convex endpoint is reachable at runtime through the `anyApi` proxy, which mints no static edge and leaves no drop record, so `impact()` returned `epistemic: 'exact'` with zero callers for symbols that genuinely have them. The issue asks specifically for an epistemic signal (`lower-bound` plus a boundary note naming the cause), *not* for synthesised edges to the callers.

### Current PR state

Open, not a draft. Fork PR from `azizur100389`, branch `codex/convex-impact-epistemic-3040`, targeting `main`. Two commits, no merge commits. `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED`. 18 files, +688/−21.

### Repository history considered

Cross-session memory flagged `SCHEMA_BUMP` collisions as a recurring regression class in this repo (#3017 and #2980 both independently bumped to 71), which is why the ledger entry got close scrutiny — see finding 10. Prior art also informed three checks that were run and then refuted: the durable-ParsedFile warm-cache store (does a new node property survive the round-trip), the v8 structured-clone worker boundary, and the uncached-`namedChildren` O(n²) defect class from #3017. The `Property.isDetail` column added under #2856 turned out to be the exact precedent for finding 11.

### Branch hygiene assessment

**clean feature/fix PR** — 18 causally connected files across ingestion, persistence and the MCP read path; no merge commits; `git diff --check` clean; no bidi controls or raw NUL bytes; conventional commit title and messages; CHANGELOG correctly untouched per this repo's rule that it is release-process-owned.

### Merge status and mergeability

**checks failing** — no conflicts (`MERGEABLE`), but two CI jobs are red and both are caused by this PR: `tests / ubuntu / coverage 2/3` (finding 1) and `tests / benchmarks (GITNEXUS_BENCH)` (finding 11). `Build & Push gitnexus` and `Vercel` also fail, but for infrastructure and authorization reasons unrelated to the change.

### Headline

Two things stand between this and merge, and they pull in opposite directions.

**1. CI is red on two checks, both caused by this PR** — the new unordered `LIMIT 1` trips the repo's own #2787 determinism guard. Four Engine-A lanes ran the guard locally and Codex B1 read the failing CI log; the guard names `src/mcp/local/convex-metadata.ts:30` and nothing else. One-line fix. Separately, the `GITNEXUS_BENCH` job is red on the #2203 emit byte-identity guard because the new `Const` CSV column changes the emitted bytes — a re-baseline with a justification note, for which this repo has an exact precedent (finding 11). No lane found this one; it surfaced in the coordinator's final CI check.

**2. The change makes `impact()` *less* trustworthy on every index that has not been re-analyzed** — the opposite of its thesis. On a pre-column index the probe's query throws, the stale-index branch returns a *defined* object, and `dispatch: convexDispatch === undefined ? 0 : 1` cannot tell that apart from a confirmed Convex endpoint. So **every `Const`, in every language, in repos with no Convex code at all**, flips from `exact` to `lower-bound` with `causes.dispatchBoundary: 1` and Convex-specific prose. Measured on this repo's own live index: 6,738 `Const` nodes. Two lanes reproduced this end-to-end against a hand-built pre-PR 7-column DDL. The MCP read path has no schema-version gate (`schemaFingerprintMismatch` is consumed only by `core/run-analyze.ts`), so an old `.lbug` is served, not rejected.

Behind those, the coverage question Codex B3 refused to ship on: as written the mechanism hedges **only** `export const foo = <factory>({ … })` imported from `(../)*_generated/server`. Convex's documented bare-handler form and its `convex/server` `*Generic` builders both stay `exact` — including, per B3 and two A lanes, the exact shape the PR's own `convex/server` unit test uses, which cannot occur in a real Convex project.

### Findings proposed as inline comments

**1. `gitnexus/src/mcp/local/convex-metadata.ts:30` — P1 — unordered `LIMIT 1` fails the #2787 determinism guard.** Trigger: any CI run or `npx vitest run test/unit/query-determinism-guard.test.ts` on this HEAD — no runtime input needed. Evidence: the guard names this exact query as the sole offender and prints both remedies; it matches the `tests / ubuntu / coverage 2/3` failure exactly. The sibling `bean-metadata.ts:44-51` — same PK-anchored `LIMIT 1` shape, same module family — already carries the declaration. Fix: `// determinism: probe — id is the Const PRIMARY KEY, so at most one row can match`, or drop the inert `LIMIT 1`. Do not add `ORDER BY` to a single-row PK lookup. `[reproduced]` — Engines **A + B**.

**2. `gitnexus/src/mcp/local/local-backend.ts:6839` — P1 — a stale index is counted as a dispatch boundary, and a flat `1` is added to a field documented as a symbol count.** Trigger: upgrade to a build containing this PR, do not re-analyze, run `impact({target:"<any const>", direction:"upstream"})` — or `context()`, which passes no `direction` and therefore also probes. Result: `epistemic:'lower-bound'`, `causes.dispatchBoundary: 1`, and a Convex re-index advisory, on a Go/Rust/Ruby constant in a repo with zero Convex code. Reproduced end-to-end against a real pre-PR `Const` DDL (`"Binder exception: Cannot find property convexEndpointFactory for n."` → `{staleIndex:true}` → `dispatch: 1`). Second, independent half: even on a genuine endpoint the contribution is a constant `1` summed at :6935 into `dispatchBoundarySymbols`, whose every other contributor counts *other* symbols — and the comment at :6897-6901 explicitly says publishing a presence value next to a real magnitude is the mistake that was already fixed once here. Fix: do not fold `staleIndex` into `dispatch` (it is a re-index condition, and `staleIndex` currently has no production reader at all), and give the real-endpoint case its own boolean cause rather than adding to a counted field — the note alone already forces `lower-bound`. `[reproduced]` — Engine **A** (4 lanes: correctness, adversarial, api-contract, docs-DoD; consistent within one engine, but reproduced end-to-end by two of them independently).

**3. `gitnexus/src/core/ingestion/languages/typescript/convex-endpoint-metadata.ts:68` — P2 — Convex's documented bare-handler form can never be hedged.** Trigger: `export const list = query(async (ctx) => ctx.db.query("messages").collect());` — officially supported (`QueryBuilder` accepts `{args?,returns?,handler} | (ctx, …) => …`) — then `impact({target:"list", direction:"upstream"})` returns `exact`, `dispatchBoundary: 0`: precisely the false confidence #3040 exists to remove. Two lanes drove the real extractor over the shape and got `NOT STAMPED`. One lane found the deeper half: with an arrow-function argument the declaration also matches a `@definition.function` pattern, `buildDefinitionPreScan` adds it to `nonValue`, and the `Const` twin is **dropped entirely** — so `convex-metadata.ts:23`'s `symbolType !== 'Const'` gate could never see it even if this line were relaxed. Accepting `arrow_function`/`function_expression` here is necessary but not sufficient; the consumer gate has to widen too. Fix, or state the limitation explicitly in the PR body and in `tools.ts` so `exact` on a Convex repo is not read as an assertion. `[reproduced]` — Engines **A + B**.

**4. `gitnexus/src/core/ingestion/languages/typescript/convex-endpoint-metadata.ts:16` — P2 — the accepted (module × factory-name) cross-product is the wrong half for `convex/server`.** `convex/server` exports the `*Generic` builders (`queryGeneric`, `mutationGeneric`, …); the bare `query`/`mutation`/… names live only in the codegen'd `./_generated/server`. So this arm fires only on an intersection that cannot exist in a real project, and misses every no-codegen Convex component or library package. Measured: `import { queryGeneric } from "convex/server"` → not stamped; `import { query } from "convex/server"` → stamped. The PR's own unit test (`convex-metadata.test.ts:57-64`, "direct convex/server provenance") asserts the second — a green test over an unreachable input. Related gaps confirmed by running them: `httpAction` (a real `_generated/server` export) is absent from the set; `convex/_generated/server` (the spelling Convex's own Next.js quickstart uses, and legal under `"baseUrl": "."`) is rejected; alias/monorepo specifiers (`@/convex/_generated/server`) are rejected; `convex-helpers`' `customQuery` wrappers are rejected. Fix: add the `*Generic` names (normalising to the canonical factory for the note text) and `httpAction`, and retarget the `convex/server` fixture at a shape that can occur. `[reproduced]` — Engines **A + B**.

**5. `gitnexus/src/mcp/local/convex-metadata.ts:55` — P2 — a failed probe and a proven negative are the same value.** the generic catch's `return undefined` (line 56, directly under this comment) is the exact signal `local-backend.ts:6839` reads as "not a Convex endpoint", so any DB failure that is not the one recognised wording restores `epistemic:'exact'` on a genuine Convex endpoint. Concrete triggers, all real strings from `pool-adapter.ts`: `"Connection pool exhausted: timed out after 15000ms…"`, `"Prepare timed out after 30000ms"`, `"Execute timed out after 30000ms"`, and `"LadybugDB not initialized for repo …"` — none match the regex. This matters because `computeEpistemicBoundary` already fans out up to 50 concurrent queries against an 8-connection pool, so contention on this exact path is designed-for, not exotic. This inverts the invariant the enclosing function states twice in its own comments ("a probe failing must never read as certainty — the whole point of this function"). Fix: return a third state on the unrecognised-error branch and have the caller add a note without incrementing `dispatch`; at minimum route it through `logQueryError` so the case is diagnosable. `[code-read]` — Engine **A** (correctness + reliability).

**6. `gitnexus/src/core/ingestion/workers/parse-worker.ts:2828` — P2 — the new provider hook is the only best-effort enrichment call here without a local `try/catch`.** A throw escapes `processFileGroup` to the language-group catch at :1237-1246, which drops **every remaining file in that language batch** with a single `reportWarning`. The file's own neighbours guard against exactly this: `descriptionExtractor` (:2693) and `extractTemplateConstraints` (:2672) each carry a local try/catch, and the comment at :2696 spells out the reason verbatim. Fix: wrap the call in the same pattern and treat a throw as `undefined`. `[code-read]` — Engine **A** (reliability + performance). Codex B1 flagged the adjacent hardening gap on the same call: nothing enforces the "structured-clone-safe" contract the hook's doc promises, and the repo already has the tool for it — the two sibling worker-boundary hooks in this same interface (`extractTemplateConstraints`, `collectCaptureSideChannel`) are documented to wrap with `assertCloneable` from `workers/clone-safety.ts` precisely so a leak is a compile error (#2143). Worth adopting for the third instance.

**7. `gitnexus/src/core/ingestion/languages/typescript/convex-endpoint-metadata.ts:35` — P2 — `findDeclarator` recurses the whole subtree and can bind a nested, same-named declarator.** Reproduced: `export const outer = (() => { const foo = query({ handler: async () => null }); return foo; })(), foo = 2;` stamps `foo` as a Convex endpoint — a plain constant published with a boundary sentence that is simply untrue. The `definition.const` tree-sitter query guarantees the target declarator is a *direct* child of the captured `lexical_declaration`, so scanning direct `namedChildren` only is behaviour-preserving for every legitimate case, kills this false positive, and removes the unbounded descent as a side effect. On the descent: a lane measured a `RangeError` at ~4,000-deep nesting in a multi-declarator statement, but the adversarial lane measured 0.016 ms/const over 821 real consts on a 371 KB file and could not construct a reachable input — so treat the depth issue as a P3 side-benefit of the same one-line change, not as the reason to make it. `[reproduced]` — Engine **A**.

**8. `gitnexus/src/mcp/tools.ts:293` — P2 — the doc this PR edits is now wrong for the case the PR adds.** "Irreducible" is exactly false for the stale-index contribution — re-indexing removes it. And the `REQUIRES RE-INDEX` paragraph immediately below (line 295, and its `impact` twin at :491) still lists only `receiverTyping`/`externalBoundary`/`undecidedSatisfaction`; `dispatchBoundary` just joined that class, and worse than the documented failure mode: it reads as a **false nonzero**, not as absent/0. Fix alongside finding 2 — if the stale case stops contributing to `dispatch`, the "Irreducible" wording survives and only the re-index list needs the addition. `[code-read]` — Engine **A** (api-contract, adversarial, docs-DoD).

**9. `gitnexus/src/mcp/local/local-backend.ts:6827` — P3 — the probe is spliced into the serial chain rather than the concurrency the surrounding code documents.** `owningTypeNames`, `queryConvexDispatchMetadata` and the `ifaceRows` query are three independent single-symbol lookups on the same `symId`, now awaited one after another. Measured against a live index: ~12.5 ms first compile, 1.2–1.8 ms steady state — small in absolute terms, but `impact` is documented in CLAUDE.md as a before-every-edit call, `Const` is 27–61 % of nodes in this repo's own two indexes, and `ifaceRows` alone already costs far more, so hiding this behind it via `Promise.all` is free. The file's own comments at :4318-4320 and :7023-7026 state exactly this discipline for the outer placement. `[reproduced]` — Engine **A** (performance, reliability, adversarial).

**10. `gitnexus/src/storage/parse-cache.ts:576` — P3 — citation nit.** #3041 is an **issue**; the PR that claims `SCHEMA_BUMP = 73` and fixes it is **#3046** ("fix(ingestion): preserve object handler identity"). The substance is right — main is 72, #3046 claims 73, 74 is free, and skipping is the correct move — so this is a one-word fix: cite "#3046 (fixes #3041)". Worth doing precisely because this ledger's protocol tells the next author to `gh pr view` the cited number, and today that 404s. Merge-order note, not a defect: whichever of #3044/#3046 lands second must re-verify, since an exact version clash survives `--force` (it still serves durable ParsedFiles when the number matches). `[reproduced]` — Engines **A + B** (7 lanes). Both engines were initially told #3041 did not exist at all; several lanes re-derived the correct facts themselves and graded accordingly.

**11. `gitnexus/src/core/lbug/csv-generator.ts:539` — P1 — the new `Const` CSV column drifts the #2203 emit byte-identity fingerprint, and the baseline was not updated.** Trigger: `node --import tsx bench/emit-persistence/measure.mjs --check` — i.e. the `tests / benchmarks (GITNEXUS_BENCH)` job, which is red. Adding `convexEndpointFactory` to `constHeader` writes one more header field into `const.csv`, so the fingerprint over the emitted CSV set necessarily changes. This is not a defect in the change — it is the guard doing its job — but it does need a deliberate re-baseline in `gitnexus/bench/emit-persistence/baselines.json` (an unchanged file, so this is anchored on the line that causes the drift). **The repo has an exact precedent in that same file**: the `_rebaselined_2856_property_is_detail` entry documents the identical situation when `Property` gained an `isDetail` column, and it records the evidence that made the re-baseline safe — dumping every CSV on both sides and diffing per file, showing the file set was identical, all but one file byte-identical, and the sole delta a header growing by one field with no row moving. Please follow that template: update `fingerprint` and add a `_rebaselined_*` note proving the drift is header-only and that no row was re-routed or re-ordered — that re-routing is precisely what this fingerprint exists to catch. `[reproduced]` — coordinator (CI log).

### Lower-priority items and test gaps (body only)

- **`lbug-db` vitest project registration (P2, not merge-blocking).** `convex-impact-epistemic-e2e.test.ts:182` passes `poolAdapter: true` (native mmap) but is not in the `lbug-db` project's include list, so it runs in `default` without `fileParallelism: false`. The two most recent `lbug-db` entries were registered on exactly this criterion — but `vitest.config.ts:122-125` explicitly notes that equally-native siblings still sit in `default` as "pre-existing drift, deliberately left alone here". Worth registering for consistency; not a rule the repo currently enforces.
- **`epistemicFrom`'s new `dispatchBoundary: dropped.dispatch` (local-backend.ts:730) is inert.** `dispatch` is `1` iff a note was also pushed, so the `notes.length === 0` arm can only ever see `0`. Harmless, but it reads as a live composition rule; a one-line invariant comment (or making the note and the count travel as one value) would prevent the next author from trusting it.
- **`ConvexDispatchMetadata.staleIndex` and `.factory` have no production reader** — `local-backend.ts` consumes only `.boundary` and `=== undefined`. They exist for the unit test. Fixing finding 2 would give `staleIndex` a real consumer.
- **`computeEpistemicBoundary(direction?)`** — only `'downstream'` is special-cased, so an omitted argument silently behaves like `'upstream'`. Correct and tested for both current callers (`context()`'s omission is deliberate and asserted); with two call sites, making it required would turn a silent default into a compile-time choice.
- **Architecture, worth the maintainer's judgment rather than an inline demand:** the hook is generic, but the persistence layer under it is not. `convexEndpointFactory` is a framework-named column hard-coded in four shared, framework-neutral places (`schema.ts` DDL, the CSV header *and* row writer, `getCopyQuery`, and both `CREATE`/`MERGE` branches), so every Rust/Go/Python/C++ `Const` now carries a permanently empty Convex column and the next framework must repeat all five edits. This also makes the PR's own new ARCHITECTURE.md sentence ("Shared ingestion persists these properties opaquely") untrue past `core/ingestion/` — a future hook implementer returning `{nestGuard: …}` would have it silently dropped at the lbug step with no error. The repo has two working precedents that needed **zero** schema/CSV/COPY changes: `Class.frameworkAnnotations STRING[]` (generic name, framework-specific values) and the versioned strings Spring AOP/bean metadata ride on `CodeRelation.reason`. A generic column name with a namespaced value (`convex:query`) would get the same query ergonomics without the fourth special case. Three lanes raised this independently; note also that the strict AGENTS.md rule does **not** apply here (`core/lbug/` is outside `core/ingestion/`) — the applicable one is DoD §2.2. At minimum, correct the "opaquely" claim in `ARCHITECTURE.md:406` and `language-provider.ts:290` to describe the two-step reality and name the three `core/lbug/` files a second hook user must edit.
- **Test gaps.** (a) `convex-metadata-persistence-contract.test.ts` never touches the CSV writer — delete the `csv-generator.ts:762-764` branch and it stays green while every analyze run with any `Const` fails the COPY; a ~20-line case asserting header field-count == row field-count == `getCopyQuery('Const')` column count is the mutant the file's name promises to kill. (b) Three of its eight assertions (`.not.toContain('isExported')`) are non-discriminating — `Const` was never in `TABLES_WITH_EXPORTED`, so they pass identically with the PR reverted. (c) The stale-index path is proven only against an author-written error string in a mocked `runQuery` — proof by construction. Two lanes closed that loop manually against a real pre-PR DDL, and the repo already has the pattern to copy (`route-method-roundtrip.test.ts`). (d) No test asserts the bare-handler or `httpAction` shapes, positively or negatively.

### Suspicions the lanes REFUTED (a large part of this review's value)

All of the following were live hypotheses in the briefs and were killed, most by execution:

- **"The stale-index regex doesn't match LadybugDB's real wording."** Refuted twice: the real text is `"Binder exception: Cannot find property convexEndpointFactory for n."`, extracted from the native binary and then produced against a live DB built with the pre-PR 7-column DDL. It works today. The residual point is fragility — prose-sniffing plus a test that asserts the author's regex against the author's own string — which is P2 at most, and the repo already has a non-textual signal (`SCHEMA_FINGERPRINT` / a catalog probe) if the author wants one.
- **"`export const foo = query({…})` isn't labelled `Const`, so the feature is inert."** Refuted against the real provider and real queries.
- **"The 8th CSV column breaks the bulk COPY / another writer emits 7 fields."** Refuted by running the real COPY, including a CSV-hostile cell; `insertNodeToLbug`/`batchInsertNodesToLbug` are test-only callers and the group/manifest path never writes `Const` rows.
- **"The warm parse-cache drops the new property."** Refuted — the slimming function spreads `...result`.
- **"`findDeclarator` is a quadratic ingest cost."** Refuted in practice (0.016 ms/const over 821 real consts; 0.07 ms on an 818 KB single-const literal). `namedChild(i)` *is* uncached and O(index) at the native level, but here `n` is the declarator-list length, not the literal size.
- **"`methodProps` leaks across loop iterations"** and **"the `isExported` hoist changed behaviour"** — both refuted; the hoisted expression is character-for-character identical with no intervening control flow, and `cachedExportCheck` is a WeakMap memo called exactly as often as before.
- **"The `direction === 'downstream'` gate is wrong / the second caller forgot an argument."** Refuted — `undefined !== 'downstream'` means `context()` probes, which is deliberate and asserted by the PR's own test.
- **"SCHEMA_BUMP 74 is wrong / collides."** Refuted — main is 72, #3046 claims 73, 74 is free.
- **"The new cause shadows the existing three"**, **"the MCP response shape breaks"**, **"unbackticked `CREATE (n:Const …)` fails"**, **"`String(row.factory ?? row[0] ?? '')` mishandles NULL"**, **"the extra column is materially expensive"** (~200–450 KB on a 248k-symbol index, <0.2 %) — all refuted, several by execution.

### CI status

- `tests / ubuntu / coverage 2/3` — **RED on exactly one test**: `query-determinism-guard.test.ts` (#2787), naming `src/mcp/local/convex-metadata.ts:30`. That is finding 1 and nothing else in the suite.
- `Build & Push gitnexus` — fails on a qemu `Bus error (core dumped)` during the DuckDB FTS extension install. **Infrastructure, not this PR.**
- `Vercel` — fails on authorization, not code.
- Everything else green: lint, format, typecheck, typecheck-web, PR-title validation, and all platform shards.
- `tests / benchmarks (GITNEXUS_BENCH)` — **also RED, and also caused by this PR.** No lane obtained this log; the coordinator pulled it during final verification. The job's vitest step passes (9 files / 14 tests); the failure is a later step, *Emit-persistence throughput / byte-identity guards (#2203)*:
  ```
  [emit-persistence --check] FAIL: byte-identity fingerprint drift
    (got 6b6f110724f646f55a99deef78bf7616146c93da4af36dd3ddfe032d04bf63e7,
     expected 4ee15e742a9839671a900df4f57c1c91196c64256c8cab2ac445bec605a092d5)
  ```
  See finding 11 — this is expected drift from the new `Const` CSV column and needs a documented re-baseline, not a code change. Both timing gates in that bench passed.
- **Note for the author:** the PR body states "67 focused tests passed … both review workflows passed" while the determinism guard is red. Whatever local run produced that number did not include `test/unit/query-determinism-guard.test.ts`.

### Coverage limits

- The worktree has no `dist/`, so the two integration tests that need the compiled parse-worker (`convex-impact-epistemic-e2e.test.ts`, and the full pipeline half of the epistemic suite) could not be executed locally. A lane confirmed this is a pre-existing worktree limitation by reproducing the identical failure on an untouched control test. CI's `setup-gitnexus` runs with `build: 'true'`, so CI should exercise them for real — that has not been independently confirmed here.
- Codex B2 could execute nothing (sandbox blocked Node/Python), so its agreement is code-read weight only; its one live contribution was a `gh api` sweep of open PRs touching `parse-cache.ts` (#3046 → 73, #3044 → 74, #3017/#2840 → 71, #1616 → 2).
- No lane exercised Windows or macOS behaviour, the group/bridge multi-repo paths, or a real end-to-end upgrade-then-query sequence on a user's existing on-disk index (the stale-index case was reproduced by hand-building the pre-PR DDL instead).
- Convex API claims were checked against the official docs via Context7, not against an installed `convex` package.

### Final verdict

**not production-ready**

Two P1 issues must be resolved before merge, and they are of different kinds. Finding 1 (`convex-metadata.ts:30`) is a one-line CI fix with a sibling in this same directory to copy. Finding 2 (`local-backend.ts:6839`) needs a design decision rather than a patch: because a stale index is indistinguishable from a confirmed endpoint at that line, every `Const` in every language — on any index not yet re-analyzed — flips from `exact` to `lower-bound` carrying Convex-specific prose, which inverts the property this PR exists to establish. Finding 11 (`csv-generator.ts:539`) is mechanical but must land for CI to go green: a documented re-baseline following the `_rebaselined_2856_property_is_detail` template already in that file. Findings 3–5 are P2 scope and correctness gaps — the shipped mechanism hedges a narrower set of Convex shapes than #3040 implies, and two of its three probe exits can restore a false `exact` — which the author should either close or state explicitly as limitations. Findings 6–9 are P2/P3 hardening items. The underlying design is sound and unusually well tested; none of this is rework.

---

*Automated multi-tool digest: 12 Claude lanes + 3 Codex legs, merged by a synthesis pass. Findings tagged `[reproduced]` were produced by executing code; verify before acting.*

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -2821,6 +2821,21 @@ const processFileGroup = (
         }
       }
 
+      const isExported =
+        language === SupportedLanguages.Vue && isVueSetup
+          ? isVueSetupTopLevel(nameNode || definitionNode)
+          : cachedExportCheck(provider.exportChecker, nameNode || definitionNode, nodeName);
+      if (definitionNode && provider.definitionPropertiesExtractor) {
```
magyargergo, line 2832: **P2** · `code-read` · engines: A

This is the only best-effort enrichment call in this loop without a local `try/catch`, and the blast radius of a throw is the whole language batch, not the one node. There is no failure boundary between here and the `processFileGroup` call site at lines 1237-1246, whose catch treats any throw as "parser unavailable" and drops every remaining file in the group with a single `reportWarning` — the file's own comment at 2696 spells out that exact hazard, which is why `descriptionExtractor` (2693) and `extractTemplateConstraints` (2672) each carry their own guard a few lines above. Suggested fix: wrap this call in the same pattern and treat a throw as if the hook returned `undefined`. Two adjacent hardening notes for the same call while you are here: (1) the hook's doc promises structured-clone safety but nothing enforces it — its two sibling worker-boundary hooks in the same interface (`extractTemplateConstraints`, `collectCaptureSideChannel`) are documented to wrap with `assertCloneable` from `workers/clone-safety.ts` so a leak is a compile error rather than a silent strip-and-warn (#2143), and this is the third instance of that shape; (2) `Object.assign(methodProps, …)` lands before `...methodProps` is spread after the identity fields, so a future provider returning `name`/`filePath`/`isExported` would rewrite them after `nodeId` was derived — a reserved-key filter or spreading `methodProps` first would close that.

gitnexus/src/core/ingestion/languages/typescript/convex-endpoint-metadata.ts
```diff
@@ -0,0 +1,70 @@
...
+      continue;
+    }
+    return CONVEX_ENDPOINT_FACTORIES.has(parsedImport.importedName)
+      ? parsedImport.importedName
+      : undefined;
+  }
+  return undefined;
+}
+
+function findDeclarator(node: SyntaxNode, nodeName: string): SyntaxNode | undefined {
```
magyargergo: **P2** · `reproduced` · engines: A

`findDeclarator` walks the entire subtree and returns the first `variable_declarator` whose name matches, with no same-declaration constraint — so a nested declarator can win. Reproduced: `export const outer = (() => { const foo = query({ handler: async () => null }); return foo; })(), foo = 2;` with `nodeName: "foo"` returns `{"convexEndpointFactory":"query"}`, publishing a plain constant as a Convex endpoint and degrading `impact()` on it to `lower-bound` with a boundary sentence that is untrue. The `definition.const` tree-sitter query guarantees the target declarator is a *direct* child of the captured `lexical_declaration`, so iterating the definition node's direct `variable_declarator` children instead of recursing is behaviour-preserving for every legitimate shape (multi-declarator statements, type annotations, explicit type arguments and `definitionNode = export_statement` were all verified to still stamp correctly) and removes the false positive. It also removes the unbounded descent as a free side effect — one lane measured a `RangeError: Maximum call stack size exceeded` at ~4,000-deep nesting when an earlier sibling declarator holds a deep literal — though that arm looks unreachable in practice (0.016 ms/const over 821 real consts on a 371 KB file), so please treat it as a bonus rather than the motivation.

gitnexus/src/mcp/tools.ts
```diff
@@ -290,7 +290,7 @@ COMPLETENESS OF incoming: alongside symbol/incoming/outgoing the result carries
 - causes: { receiverTyping, dispatchBoundary, externalBoundary, undecidedSatisfaction } — machine-readable WHY. Every field counts MISSING THINGS, never sentences:
   - causes.receiverTyping (unit: call sites) > 0 — RESOLVER GAP: the analyzer dropped that many call sites on this name because it could not type the receiver, so they are missing from incoming. Do not read an absent caller as proof none exists.
   - causes.externalBoundary (unit: call sites) > 0 — the calls left the indexed program (System.out.println, fetch(...)). NOT a defect: no in-graph node could have been reached. An epistemic:'exact' result can carry this.
-  - causes.dispatchBoundary (unit: symbols) > 0 — DI / interface dispatch: implementations plus interface-level consumers behind a boundary static analysis cannot cross. Irreducible.
+  - causes.dispatchBoundary (unit: symbols) > 0 — DI, interface, or framework runtime-proxy dispatch: symbols on or beyond a boundary static analysis cannot cross. Irreducible.
```
magyargergo: **P2** · `code-read` · engines: A

Two doc issues created by the new contribution to this field. First, "Irreducible" is exactly false for the stale-index case: on an index predating `convexEndpointFactory` the probe throws, `queryConvexDispatchMetadata` returns a defined stale-index object, `dispatchBoundary` reads 1, and re-indexing removes it entirely — the opposite of irreducible. Second, the `REQUIRES RE-INDEX` paragraph just below (line 295, and its `impact` twin at line 491) still lists only `receiverTyping` and `externalBoundary` (plus `undecidedSatisfaction` on the impact side). `dispatchBoundary` has now joined that class, and in a worse form than the one documented there: those causes "read as absent/0" against an older index, whereas this one reads as a **false nonzero**. If the stale case stops contributing to `dispatch` (see the `local-backend.ts:6839` comment), the "Irreducible" wording survives untouched and only the re-index list needs `dispatchBoundary` added with a note about its new index-time dependency.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -6821,10 +6824,19 @@ export class LocalBackend {
               ? await this.owningTypeNames(repo, symId)
               : []),
           ]);
+    const convexDispatch =
```
magyargergo: **P3** · `reproduced` · engines: A

This `await` is spliced into an already-serial chain rather than run alongside its neighbours. `owningTypeNames(repo, symId)` just above, this probe, and the `ifaceRows` query just below are three independent single-symbol lookups keyed on the same `symId` with no data dependency between them, and the pool explicitly supports concurrent execution (each concurrent query takes its own Connection). Measured against a live index with the repo's own `openLbugConnection`: ~12.5 ms for the first compile of this query text in a session, 1.2–1.8 ms steady state — while the `ifaceRows`-shaped query it now waits behind costs 397–507 ms first-compile and ~9–12 ms steady state, so folding this into a `Promise.all` with it makes the cost effectively free. It is worth doing because `impact` is documented in CLAUDE.md as a run-before-every-edit call, `Const` is 27–61 % of nodes in this repo's own two indexes, and `context()` never passes `direction` so it always pays it for `Const` targets. The surrounding code documents exactly this discipline for the outer placement of this function, at lines 4318-4320 and 7023-7026. Suggested fix: start the promise before the `undecidedDrops`/`ifaceRows` work and await it alongside them — neither reads the other's output.

gitnexus/src/core/lbug/csv-generator.ts
```diff
@@ -536,6 +536,7 @@ export const streamAllCSVsToDisk = async (
 
     // Multi-language node types share the same CSV shape (no isExported column)
     const multiLangHeader = 'id,name,filePath,startLine,endLine,content,description';
+    const constHeader = `${multiLangHeader},convexEndpointFactory`;
```
magyargergo, line 539: **P1** · `reproduced` · engines: coordinator

Adding a field to `constHeader` changes the bytes `streamAllCSVsToDisk` emits, which drifts the #2203 emit byte-identity fingerprint — this is why `tests / benchmarks (GITNEXUS_BENCH)` is red, at the *Emit-persistence throughput / byte-identity guards (#2203)* step: `[emit-persistence --check] FAIL: byte-identity fingerprint drift (got 6b6f1107…, expected 4ee15e74…)`. The job's vitest step passes; only this guard fails, and both of its timing gates passed. Nothing here is wrong with your change — the guard is working — but the committed baseline needs a deliberate update in `gitnexus/bench/emit-persistence/baselines.json` (unchanged by this PR, hence this anchor). That file already contains the exact template: the `_rebaselined_2856_property_is_detail` entry covers the identical case when `Property` gained an `isDetail` column, and it documents the evidence that justified the re-baseline — every CSV dumped on both `origin/main` and the branch and diffed per file, showing an identical file set, all but one file byte-identical, and the only delta a header gaining one field with no row moving. Please update `fingerprint` and add a matching `_rebaselined_*` note for `Const`/`convexEndpointFactory` showing the drift is header-only and that no row was re-routed between pair files or re-ordered within one — a moved row is exactly what this fingerprint exists to catch, and a bare hash update would hide it.

gitnexus/src/storage/parse-cache.ts
```diff
@@ -569,7 +569,13 @@ import type { ParseWorkerResult } from '../core/ingestion/workers/parse-worker.j
 // IN-FLIGHT claim, not above origin/main. Every open PR touching gitnexus/ was
 // scanned; #3017 is the only other claimant.
 // RE-CHECK AGAINST origin/main AND OPEN PRs IMMEDIATELY BEFORE MERGING.
-const SCHEMA_BUMP = 72;
+//
+// 72 -> 74 adds import-proven Convex endpoint metadata to Const worker output.
+// A warm v72 cache has no convexEndpointFactory property, so the MCP impact
+// probe would keep claiming exact results for unchanged endpoints. Version 73
+// is intentionally skipped because the concurrent #3041 handler-identity PR
```
magyargergo: **P3** · `reproduced` · engines: A+B

Citation nit: #3041 is an **issue** ("Same-named nested functions (Convex handler:) collapse to one node…"), not a PR — `gh pr view 3041` 404s. The PR that actually claims `SCHEMA_BUMP = 73` is **#3046, "fix(ingestion): preserve object handler identity"**, which fixes #3041 and is literally the handler-identity PR this comment describes. The substance of the entry is correct and verified: `origin/main` is at 72, #3046 claims 73, #3017/#2840 claim 71, #1616 claims 2, so 74 is free and skipping 73 is the right move. Suggested fix is one phrase — "the concurrent #3046 (fixes #3041) handler-identity PR". It matters more than a typo because this ledger's own protocol instructs the next author to re-check the cited PRs immediately before merging, and today that lookup dead-ends. Unrelated merge-order note: whichever of #3044 and #3046 lands second should re-verify its number, since an exact `SCHEMA_BUMP` clash is the one case that survives `--force` (a forced rebuild still serves durable ParsedFiles when the version matches).

azizur100389:
Addressed the review in 01a2598e: added real Convex API provenance and bare-handler coverage, conservative probe-failure handling, zero invented boundary counts, persistence/cache/docs parity, provider-hook isolation, canonical-property protection, and a verified emit rebaseline. The 136 focused tests, typecheck, lint, format, emit benchmark, and change-impact validation pass.

Issue #3014, opened 2026-08-20
MCP/CLI read path never loads VECTOR, and can't recover from serve's shadow


bill2github:
Two defects on the same path. MCP and gitnexus cypher/context/impact go through pool-adapter.js + local-backend.js; serve goes through lbug-adapter.js. The former is missing two things the latter has, and both fail silently.                                                                                
                                                                                                                                                                                                                                                                                                                  
gitnexus 1.6.9, @ladybugdb/core ^0.18.0, macOS Intel x86_64, Node v24.14.0.                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                  
1. VECTOR is never loaded. CALL QUERY_VECTOR_INDEX(...) through the MCP cypher tool:                                                                                                                                                                                                                            
Catalog exception: function QUERY_VECTOR_INDEX is not defined. This function exists in the VECTOR extension.                                                                                                                                                                                                                                                                                           
FTS works on the same connection. loadVectorExtension is only called from embedding-pipeline.js and lbug-adapter.js:2137 — never from pool-adapter.js or local-backend.js. local-backend.js:1837 guards on isVectorExtensionSupportedByPlatform(), but that's platform !== 'win32', not a readiness check.      
                                                                                                                                                                                                                                                                                                                  
The throw is swallowed by a bare catch {} in semanticSearch, then the exact-scan fallback hits if (embeddingCount > exactScanLimit) return [] with DEFAULT_EXACT_SCAN_LIMIT = 10_000. So on a repo with 45,193 embeddings, semantic search returns nothing, with no warning. serve's /api/search on the same  data is correct in 0.4 s.                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                  
Asks: load VECTOR here — loadVectorExtension already returns false on win32 internally (lbug-adapter.js:2070), so this doesn't regress #1178/#1181; same shape as #2624. And separately, make that return [] warn — "cannot run" currently looks identical to "no matches".                                     
                                                                                                                                                                                                                                                                                                                  
[...]

PR #3045, opened 2026-08-25
fix(storage): load VECTOR only when needed

### LINKED ISSUE #3021: Exact CLI reads eagerly warn about missing VECTOR with zero embeddings
### Area

### Summary

Exact `context` and `cypher` reads eagerly probe the optional VECTOR extension and emit a level-40 warning even when the index has zero `CodeEmbedding` rows and the command does not perform semantic retrieval.

### Context

This was discovered while using an exact-symbol `gitnexus context --file ... --content` command as an index-freshness witness. The warning made the failure appear to be caused by a missing VECTOR extension, but the actual command result was that an extensionless Python script was not indexed. That separate source-coverage request is already tracked by #2707.

The current read-pool initialization in `src/core/lbug/pool-adapter.ts` eagerly calls `loadVectorExtension(..., { policy: 'load-only' })` for every pool. This affects exact symbol lookup and raw Cypher, not just semantic/hybrid search.

### Expected behavior

- Exact `context`, UID lookup, and raw `cypher` operations that do not use semantic retrieval should return clean machine-readable output without an irrelevant high-severity VECTOR warning.
- VECTOR should be loaded lazily when a semantic operation requires it, or the unavailable probe should be debug-level/suppressed when the current command can proceed exactly.
- A genuine command error such as `Symbol '<name>' not found` should remain the clear terminal diagnostic.

### Actual behavior

Every exact read initializes the shared pool, attempts `LOAD VECTOR`, and writes this level-40 record before the command JSON:

```json
{"level":40,"name":"gitnexus","msg":"GitNexus: VECTOR extension unavailable; continuing without VECTOR features. load-only policy (no install attempted); LOAD VECTOR failed: Binder exception: Extension: vector is an official extension and has not been installed. You can install it by: install vector."}
```

This occurs even though:

- `MATCH (e:CodeEmbedding) RETURN count(e)` returns `0`;
- `gitnexus doctor` reports the supported `exact-scan` fallback;
- an exact context lookup in an indexed `.py` file succeeds;
- raw node-count Cypher succeeds and has no semantic dependency.

For a missing symbol, the same warning precedes the real JSON error and can cause wrappers that surface the nonzero command's diagnostic stream to report VECTOR as the apparent cause.

### Steps to reproduce

1. Analyze a repository without embeddings so it contains no `CodeEmbedding` rows.
2. Leave VECTOR uninstalled; do not set `GITNEXUS_LBUG_EXTENSION_INSTALL`.
3. Confirm the index has no embeddings:

   ```sh
   gitnexus cypher 'MATCH (e:CodeEmbedding) RETURN count(e) AS embeddingCount' --repo /path/to/repo
   ```

   The command returns `embeddingCount = 0` but first emits the VECTOR warning.

4. Run an exact symbol lookup:

   ```sh
   gitnexus context known_symbol --repo /path/to/repo --file path/to/file.py --content
   ```

   The lookup succeeds but first emits the same warning.

5. Run a missing-symbol lookup:

   ```sh
   gitnexus context missing_symbol --repo /path/to/repo --file path/to/script --content
   ```

   The warning precedes the actual `{ "error": "Symbol 'missing_symbol' not found" }` result.

### Environment

- OS: macOS arm64
- Node: v26.7.0
- GitNexus: 1.6.10-rc.211
- LadybugDB: 0.19.1
- Embedding backend: configured HTTP provider, reported supported by `gitnexus doctor`
- `GITNEXUS_LBUG_EXTENSION_INSTALL`: unset
- VECTOR index: unavailable
- Semantic mode: exact-scan
- CodeEmbedding rows: 0

### Logs / screenshots

Successful exact lookup:

```text
{"level":40,...,"msg":"GitNexus: VECTOR extension unavailable; continuing without VECTOR features..."}
{
  "status": "found",
  "symbol": { "name": "delegate_complete_state_tier", "filePath": "scripts/tests/spec-feature-orchestrator-state-test.py" },
  "epistemic": "exact"
}
exit 0
```

Missing exact lookup:

```text
{"level":40,...,"msg":"GitNexus: VECTOR extension unavailable; continuing without VECTOR features..."}
{
  "error": "Symbol 'validate_review_lease_terminal_ledger' not found"
}
exit 1
```

azizur100389:
## Summary

- Stop exact reads from probing the optional VECTOR extension.
- Load VECTOR lazily for semantic and explicit vector queries, coalesce concurrent loads, and retain exact-scan fallback.

## Validation

- 43 focused native VECTOR tests passed.

Fixes #3021

---

## Summary

*This appears to be a broad-reaching storage and database-pool change centered on lazy `VECTOR` loading, with substantial downstream reach through local backend behavior and related execution flows. The change is concentrated in core runtime code, with tests covering pool, dispatch, embedding, and lazy-vector behavior.*

**🔴 CRITICAL blast radius. A storage-layer change centered on `gitnexus/src/core/lbug/pool-adapter.ts` and `gitnexus/src/mcp/local/local-backend.ts`, reaching a wide set of callers and execution flows.**

The graph reaches 126 dependents across 33 affected flows. The hottest file is `gitnexus/src/core/lbug/pool-adapter.ts`, where `closeOne`, `initLbugWithDb`, `checkout`, `resolve`, `reject`, `timer`, `checkin`, `withTimeout`, `executeQuery`, and `executeParameterized` are changed alongside `SharedDB`. The local backend also changes `LocalBackend` and `semanticSearch`, placing the main review focus on connection lifecycle, query execution, and vector access.

Architecture impact is concentrated in `Local`, with additional reach across `Cli`, `Wiki`, `Lbug`, `Server`, `Mcp`, `Search`, `Embeddings`, `Integration`, `Graph`, and `Group`. Review the runtime changes first, then the corresponding coverage in `gitnexus/test/integration/lbug-pool.test.ts`, `gitnexus/test/unit/lbug-pool-fts-load.test.ts`, `gitnexus/test/unit/local-backend-embedding-dims-warn.test.ts`, and `gitnexus/test/unit/local-backend-lazy-vector.test.ts`. No files are marked HIGH or CRITICAL risk.

_Added by GitNexus for PR #3045. Edit freely — this block is replaced on the next review, everything above it is left untouched._


azizur100389:
Addressed the automated findings in f0e88334: the regression now opens two distinct non-external Database instances across close/reopen, and the inaccurate pre-change comment is removed.

Validated: 43 affected tests, typecheck, ESLint, and Prettier.

azizur100389:
Rechecked on f0e88334: `executeQuery` delegates to `executeParameterized`, which detects `QUERY_VECTOR_INDEX` and awaits `ensureVectorExtension` before query checkout (`pool-adapter.ts:1047-1077`).

The native regression exercises that direct pool path after reopening a fresh read-only Database and passes in the 43-test affected suite. No duplicate preflight was added.

gitnexus/src/core/lbug/pool-adapter.ts
```diff
@@ -921,6 +912,51 @@ export async function initLbugWithDb(
...
 
+/**
+ * Lazily load VECTOR for a semantic query.
+ *
+ * Exact graph reads never call this function, so opening their read pool does
+ * not probe or warn about an optional extension they do not use. The promise
+ * lives on SharedDB because extension scope is per Database, and also joins
+ * concurrent first semantic requests onto one LOAD attempt.
+ */
+export async function ensureVectorExtension(repoId: string): Promise<boolean> {
```
magyargergo, line 923: **P1 — CI red (test-only). `[reproduced]`**

Anchoring here because the file that actually fails is not in this diff: `gitnexus/test/unit/calltool-dispatch.test.ts:21-34`. Its hoisted `lbugMocks` stubs five pool-adapter exports (`initLbug`, `executeQuery`, `executeParameterized`, `closeLbug`, `isLbugReady`) and the factory returns `{ ...actual, ...lbugMocks }` — so this **new sixth export falls through to the real implementation**. The mocked `initLbug` never populates the real module-level `pool` Map, `pool.get(repoId)` is `undefined`, and line 926 throws. `semanticSearch` catches it, `vectorReady` stays `false`, the `QUERY_VECTOR_INDEX` string is never built, and three assertions written to prove the vector lane runs now silently prove it doesn't:

```
× falls back to the exact scan with a once-per-backend warning when the vector index query fails
× issues vector index query when VECTOR is supported by the platform
× threads GITNEXUS_VECTOR_MAX_DISTANCE into the vector index WHERE clause
Tests  3 failed | 250 passed (253)
```

Reproduced locally; matches CI job 97863521921 exactly. **Not a production regression** — `isLbugReady` is literally `pool.has(repoId)`, the same state this function reads, so the two can only disagree when one is mocked and the other isn't. And `semanticSearch`'s first pool call (the `COUNT(*)` probe at `local-backend.ts:3196`) already throws the identical error, so this throw adds no new production failure mode.

Two ways out, your call:
1. Add `ensureVectorExtension: vi.fn().mockResolvedValue(true)` to `lbugMocks` — the pattern this PR already applies correctly to `local-backend-embedding-dims-warn.test.ts` and the new `local-backend-lazy-vector.test.ts`. Leave it at `true` for the "falls back" test too: that test's `executeQuery` mock already throws on `QUERY_VECTOR_INDEX`, so the inner catch still produces the message it asserts.
2. Drop the `if (vectorReady)` gate in `semanticSearch` and treat this function as a best-effort warm-up, letting the vector query degrade inside the pre-existing try/catch as on BASE. That keeps both eager loads deleted (#3021 still fixed), restores BASE's `[query:vector]` diagnostic, makes all three tests pass **untouched**, and shrinks the diff — at the cost of one doomed vector query per semantic search on a VECTOR-less machine, which is exactly BASE's behavior.

For context: PR #2624 hit this identical trap and needed a commit titled *"test(pool): register loadVectorExtension in the pool unit-suite mocks"*. Adding an export to this module breaks every explicit-factory mock of it; a repo-wide sweep of `vi.mock('.../pool-adapter.js')` sites is cheap insurance.

gitnexus/src/core/lbug/pool-adapter.ts
```diff
@@ -1033,6 +1069,13 @@ export const executeParameterized = async (
     throw new Error(`LadybugDB not initialized for repo "${repoId}". Call initLbug first.`);
   }
 
+  // Exact reads must not pay for VECTOR, but an explicit raw vector query is
+  // itself a semantic read. Preflight that narrow surface so direct pool and
+  // `cypher` callers retain the old working contract without eager startup.
+  if (/\bQUERY_VECTOR_INDEX\s*\(/i.test(cypher)) {
```
magyargergo: **P3 — the preflight re-emits the exact warning #3021 exists to delete, on an exact read. `[reproduced]`**

The regex tests raw Cypher text, so a string literal or comment matches as readily as a real `CALL`:

```cypher
MATCH (n) WHERE n.body CONTAINS 'QUERY_VECTOR_INDEX(' RETURN n
```

`'` is a non-word character so `\b` holds, `/i` covers case, `\s*` matches zero — the regex fires. Reproduced against the real native stack: `RETURN 'QUERY_VECTOR_INDEX(' AS s`, a statement with zero vector semantics, moved the capability registry from `VECTOR cap = undefined` to `{"name":"VECTOR","loaded":true}` on the pooled read-only Database. On a machine where VECTOR won't load, that same exact read reaches `ExtensionManager.markUnavailable` and emits the level-40 `"GitNexus: VECTOR extension unavailable; continuing without VECTOR features. load-only policy…"` — issue #3021's own symptom, reintroduced by the fix for #3021. Bounded (deduped once per `name:reason` per process), but realistic: GitNexus indexes source code, and this repo's own source contains that literal. The MCP `cypher` tool / CLI `gitnexus tool cypher` is the reachable surface; BM25 binds `$query` as a parameter and is unaffected.

Suggestion: anchor on procedure position — `/(^|[\s;)])CALL\s+QUERY_VECTOR_INDEX\s*\(/i` — or strip literals/comments before testing.

Two related observations on these same lines:
- **No `try`/`catch`.** The sibling call site at `local-backend.ts:3219-3229` wraps `ensureVectorExtension` and degrades; here a rejection (checkout timeout, pool-closed rejection, integrity error) propagates out of `executeParameterized` and fails a query that never invokes the vector procedure and that BASE would have run. `await ensureVectorExtension(repoId).catch(() => {})` matches the advisory intent — a resolved `false` already lets the query proceed.
- **New suspension between `pool.get` and `checkout`.** Before this PR, lines 1067→1081 ran with no intervening `await`, which is what made the `if (!entry)` check authoritative. Two lanes (code-read, not reproduced) argue a concurrent `closeOne` landing in that window makes the later `checkout(entry)` throw the internal `Connection pool integrity error: expected 8 connections but found …` assertion instead of the purpose-built `"…pool closed…retry the query"` message. Worth a re-read of `pool.get(repoId)` after the preflight, or moving the preflight after checkout.

This preflight also has **no unit test at any level** — the only coverage is the integration test, indirectly. A small table-driven test over representative Cypher shapes (real `CALL`, lowercase, newline before `(`, and the string-literal case above) would pin all of this.

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -3214,6 +3215,18 @@ export class LocalBackend {
       this.lastQueryEmbeddingDims.set(repo.lbugPath, dims);
       const queryVecStr = `[${queryVec.join(',')}]`;
       const maxDistance = getVectorMaxDistance(DEFAULT_MCP_VECTOR_MAX_DISTANCE);
+      let vectorReady = false;
+      try {
+        vectorReady = await ensureVectorExtension(repo.lbugPath);
+      } catch (err) {
+        if (!this.warnedVectorUnsupported) {
```
magyargergo: **P3 — one flag now guards two different diagnostics. `[code-read]`**

This new site and the pre-existing one at `:3267` both read and set `this.warnedVectorUnsupported`, but they report different faults: *the extension wouldn't load* (pool/checkout/native LOAD) versus *the extension loaded and the index query failed* (missing or stale HNSW index, `FLOAT[N]` width mismatch). Within one `semanticSearch` call they're mutually exclusive; across calls on the same long-lived `LocalBackend` they are not. A transient checkout failure early in an MCP session burns the flag, and every later genuine `"vector index query failed; using exact scan fallback"` — the actionable one — is silent for the rest of the process.

Two booleans (`warnedVectorLoadFailed` / `warnedVectorQueryFailed`) would match the one-flag-per-condition convention already used in this file (`warnedMissingEmbeddingStack`, `warnedSiblingDrift`).

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -3226,9 +3239,10 @@ export class LocalBackend {
       // catch below is the fallback: any failure (extension unloadable, index
       // absent, older DB) degrades to the exact scan with a once-per-backend
       // diagnostic instead of being silently swallowed.
-      try {
-        bestChunks = await collectBestChunks(limit, async (fetchLimit) => {
-          const vectorQuery = `
+      if (vectorReady) {
```
magyargergo, line 3244: **P3 — the comment directly above this gate now describes the opposite of what the code does. `[code-read]`**

> `// Always TRY the vector lane — no platform gate. … The catch below is the fallback: any failure (extension unloadable, index absent, older DB) degrades to the exact scan with a once-per-backend diagnostic instead of being silently swallowed.`

After this change the lane is *not* always tried, and "extension unloadable" no longer reaches "the catch below" — it is handled by `vectorReady` above. This file uses comments as its design record (the `#2623 follow-up` note here is exactly that), so leaving it stale will mislead the next reader of this block.

Same place, one true residual worth a line: when `ensureVectorExtension` resolves `false` (the ordinary "VECTOR unavailable" outcome — it does not throw), neither `[query:vector]` message fires. The condition itself is still reported at level 40 by `ExtensionManager.markUnavailable`, so this is **not** an observability regression — but a log consumer keyed on the `[query:vector]` string specifically loses that signal. If the silence is intentional, saying so in the comment is enough; if not, routing a resolved `false` through the same once-per-instance warn closes it.

magyargergo (commented):
> **Verdict: needs one fix before merge.** The production design is sound and was verified working end-to-end; the blocker is a one-line test-mock gap that turned required CI red.

## Multi-engine review of #3045 — `fix(storage): load VECTOR only when needed`

Reviewed at head `f0e8833` against base `031e123`. Two engines, not five independent opinions: **Engine A (Claude)** ran the GitNexus PR swarm (pr-facts-historian, risk-architect, security-boundary, test-ci-verifier) plus four `ce-code-review` personas (correctness, adversarial, testing, reliability); **Engine B (Codex, gpt-5.5/xhigh)** ran the same swarm and ce scaffolding plus a native adversarial pass. All three Codex legs were live and produced real findings — and all three converged on exactly one actionable item, the CI red. Where the two engines agree below I say so; where only one engine reached something I say that too, because same-engine convergence is not independent confirmation (this PR produced a textbook example — see *Refuted*).

### What this review validated (please read this part too)

The riskiest question about this PR — does removing the two eager `loadVectorExtension` calls that PR #2624 added for #2623 reopen that hole? — comes back **clean**, and it was checked by both engines and confirmed by live execution:

- **The PR does what it claims, proven end-to-end on the real native stack** `[reproduced]`. A standalone replica of the integration sequence shows: after `initLbug`, `VECTOR cap = undefined` (the eager probe is genuinely gone — the #3021 goal); an exact read still leaves `VECTOR cap = undefined` (exact reads don't pay for VECTOR); and the vector query then returns `[{"nodeId":"func:vec","distance":1.79e-7}]` with `VECTOR cap = {loaded:true}`. #3021 fixed *and* the #2623/#2624 pool vector lane restored, in one run.
- **Consumer coverage is complete.** Exhaustive grep finds exactly two `QUERY_VECTOR_INDEX` producers: `local-backend.ts:3246` (covered twice — the explicit `ensureVectorExtension` and the `executeParameterized` preflight) and `embedding-pipeline.ts:1032`, which runs on the **core** adapter, not the pool, and loads the extension itself. `CREATE_/DROP_VECTOR_INDEX` never reach the read-only pool. No third route exists.
- **No production path reaches `ensureVectorExtension` with an unregistered pool.** `query()` awaits `ensureInitialized(repo)` → real `initLbug(repo.lbugPath)` before `semanticSearch`, and the same key is passed on. Independently reached by risk-architect, test-ci-verifier, ce-correctness, ce-adversarial, ce-reliability **and** Codex swarm lane 3 + Codex ce. Belt and braces: `semanticSearch`'s *first* pool call (the `COUNT(*)` probe at `:3196`) already throws the identical "not initialized" error, so the new throw adds no new production failure mode.
- **`closeOne` lifecycle is correct in both branches** — the external branch clears `vectorLoaded`/`vectorLoadPromise`; the non-external branch deletes the whole `SharedDB`, so a reopen constructs one with both fields absent. No stale `vectorLoaded = true` survives a Database swap in the reachable topology (traced through the #2614 reinit path).
- **No TDZ, no unhandled rejection, no deadlock, no pool starvation or exhaustion.** The `cachedAttempt` self-reference inside its own `.catch` was probed with a standalone replica of the exact promise shape, including a synchronous throw: handler runs as a microtask, sees `shared.vectorLoadPromise === cachedAttempt`, clears it. Concurrent callers join one attempt (`calls = 1`). Nobody holds a connection across the load, so at most one extra connection per Database against `MAX_CONNS_PER_REPO = 8`.
- **The integration test avoids the classic trap** — it calls `await core.closeLbug()` before `initLbug` so the pool opens its *own* Database instead of inheriting the core adapter's VECTOR load. That is the discipline this exact surface has burned people on before, and it is applied correctly here.
- **Hygiene sweep clean:** no NUL bytes, no bidi/hidden Unicode controls, `git diff --check` clean, no workflow/YAML/Docker changes. The new `logger.warn({ err }, …)` uses the correct `err` key.
- **The bot's P1 is genuinely fixed.** `gitnexus-check` flagged on the first commit that lazy loading was not applied to direct pooled vector queries; the second commit's `executeParameterized` regex preflight closes it, and both engines re-verified that the direct pool path now works. Credited, not re-reported.

Mitigating context worth stating: **`initLbugWithDb` has zero production callers** (test helpers only), so one of the two deleted eager loads was already dead in production. The real production delta is the `doInitLbug` one.

### Headline: CI is red, and it is test-only

`tests / ubuntu / coverage 3/3` fails with three assertions in `test/unit/calltool-dispatch.test.ts`. Root cause: that file's hoisted `lbugMocks` stubs five pool-adapter exports and the factory returns `{ ...actual, ...lbugMocks }`, so the **new sixth export `ensureVectorExtension` falls through to the real implementation**. The mocked `initLbug` never populates the real module-level `pool` Map, so `pool.get(repoId)` → `undefined` → throw. `semanticSearch` catches it, `vectorReady` stays `false`, the `QUERY_VECTOR_INDEX` string is never built, and three assertions that exist to prove the vector lane runs now silently prove it doesn't.

**Not a production regression** — `isLbugReady` is literally `pool.has(repoId)`, the same state `ensureVectorExtension` reads, so the two can only disagree when one is mocked and the other isn't. That is exactly what this one file does.

Corroboration: 4 Claude lanes + all 3 Codex legs, and reproduced locally (`3 failed | 250 passed`), matching CI job 97863521921 verbatim. Worth noting the recurrence: **PR #2624 needed the identical fix** — it carries a commit literally titled *"test(pool): register loadVectorExtension in the pool unit-suite mocks"*. This PR already applied the correct pattern to **two** sibling suites (`local-backend-embedding-dims-warn.test.ts`, and the new `local-backend-lazy-vector.test.ts`); `calltool-dispatch.test.ts` is simply the one that was missed.

### Inline findings

| # | Location | Sev | Trigger | Tag | Engines |
|---|---|---|---|---|---|
| 1 | `pool-adapter.ts:923` (real file: `test/unit/calltool-dispatch.test.ts:21-34`) | **P1** | Any run of that suite; new export not in `lbugMocks` | `[reproduced]` | Claude ×4 + Codex ×3 |
| 2 | `pool-adapter.ts:1075` | P3 (one lane: P2) | `cypher` statement whose *text* contains `QUERY_VECTOR_INDEX(` in a string literal/comment | `[reproduced]` | Claude ×3 |
| 3 | `local-backend.ts:3222` | P3 | Load-failure warn burns the shared flag; a later index-query failure is then permanently silent | `[code-read]` | Claude ×4 |
| 4 | `local-backend.ts:3242` | P3 | Comment above the new gate asserts behavior the gate removed | `[code-read]` | Claude ×2 + coordinator |

**Finding 2 in full**, because it is the one that touches the PR's own goal: `/\bQUERY_VECTOR_INDEX\s*\(/i.test(cypher)` scans raw Cypher, so `MATCH (n) WHERE n.body CONTAINS 'QUERY_VECTOR_INDEX(' RETURN n` matches (`'` is a non-word char, so `\b` holds; `\s*` matches zero). Reproduced against the real native stack — `RETURN 'QUERY_VECTOR_INDEX(' AS s`, a statement with zero vector semantics, moved the capability registry from `VECTOR cap = undefined` to `{loaded:true}`. On a machine without VECTOR that same exact read reaches `ExtensionManager.markUnavailable` and emits the level-40 `"GitNexus: VECTOR extension unavailable; continuing without VECTOR features…"` — the exact warning #3021 exists to delete. Impact is bounded (deduped once per `name:reason` per process), but it is an incomplete-fix hole in the PR's own thesis, and it is plausible here because GitNexus indexes source code and this repo's own source contains that literal. Two further notes on the same three lines: the preflight has **no try/catch** (unlike the `semanticSearch` call site two frames up, so a rejecting `ensureVectorExtension` now fails a query BASE would have run), and it has **no unit test at any level** — the only coverage is the integration test, indirectly.

### Lower-priority notes

- **Two remedies for the CI red; the author should pick.** (a) Add `ensureVectorExtension: vi.fn().mockResolvedValue(true)` to `lbugMocks` — one line, matches the pattern already used twice in this PR. (b) ce-correctness argues for dropping the `if (vectorReady)` gate entirely and treating `ensureVectorExtension` as a best-effort warm-up, letting the vector query degrade inside the pre-existing try/catch exactly as on BASE. (b) still deletes both eager loads (so #3021 stays fixed), restores BASE's `[query:vector]` diagnostic wording, makes all three failing tests pass **untouched**, and shrinks the diff — at the cost of one doomed vector query per semantic search on a VECTOR-less machine, which is precisely BASE's behavior. Not our call.
- **Stale comment above `if (vectorReady)`** (`local-backend.ts:3235-3241`): "Always TRY the vector lane" and "The catch below is the fallback: any failure (extension unloadable…)" are now false — extension-unloadable is handled by the new gate, not by that catch. In a file that uses comments as its design record, this misleads the next reader.
- **`warnedVectorUnsupported` now multiplexes two failure classes** (load-failed at `:3222`, index-query-failed at `:3267`). Whichever fires first permanently silences the other for that `LocalBackend` instance. The repo's own convention nearby is one flag per condition (`warnedMissingEmbeddingStack`, `warnedSiblingDrift`).
- **No unit test for the `executeParameterized` preflight** — neither the true positive, the string-literal false positive, nor the un-caught rejection. Flagged independently by security-boundary, ce-correctness, ce-adversarial and ce-testing.
- **`test/unit/local-backend-lazy-vector.test.ts` is not in `scripts/cross-platform-tests.ts`.** Probably fine: it fully mocks `pool-adapter` and `embedder`, has no native or `process.platform` path, and its sibling `local-backend-embedding-dims-warn.test.ts` is likewise absent. The genuinely platform-sensitive suites (`lbug-pool-fts-load.test.ts`, `lbug-pool.test.ts`) *are* registered.
- **One lane flagged a teardown race** where an in-flight load writes `shared.vectorLoaded = true` onto an external `SharedDB` after `closeOne` cleared it. Reachable only through `initLbugWithDb`, which has no production callers; risk-architect independently judged the adjacent race benign. Noted, not filed. A `dbCache.get(entry.dbPath) === shared` guard before the write would close it if you want belt-and-braces.
- **Possible link to open issue #3014** ("MCP/CLI read path never loads VECTOR, and can't recover from serve's shadow", filed 2026-08-20, unassigned, zero comments). Its first defect overlaps this PR's surface; its second (serve sidecar) does not. Not closed by this PR, but worth cross-linking.

### Refuted suspicions (validation is a result)

Dropped after checking: no production path reaches `ensureVectorExtension` with an unregistered pool; the #2623/#2624 pool-vector regression is **not** reopened; `closeOne` is correct in both branches; no TDZ, no unhandled rejection, no deadlock/starvation/exhaustion; no stale `vectorLoaded = true` surviving a Database swap in the reachable topology; the integration test does not inherit the core adapter's Database; the `#2798` embedding-width warning is unaffected; `bestChunks` fallback is behaviourally equivalent to BASE in all three states; hygiene clean. Local `lbug-pool.test.ts` timeouts were also investigated and dismissed — a tsx cold-transform artifact under a load average of 82 on 24 cores, with CI shards 1/3 and 2/3 green on this head as the control.

**Methodology note, because it matters for how you weigh this review.** Three Claude lanes (risk-architect, ce-reliability, ce-testing) independently filed a P1/P2 "silent degradation — a resolved `false` from `ensureVectorExtension` emits no warning at all, where BASE always warned". All three traced `ExtensionManager.ensure` returning `false` and stopped there. It is wrong: `ensure` returns `false` *through* `markUnavailable`, which itself calls `warn(message)`. The PR passes only `{policy:'load-only'}`, so `quiet === false`, and the process-wide `ExtensionManager` has no `warn` override — the sink is `logger.warn` at level 40. The condition **is** still reported. Zero Codex legs raised it. Three same-engine lanes converging on a false positive is exactly why we do not treat within-engine agreement as corroboration, and it is not reported above. The true residual is much weaker and appears in the notes: the *local-backend-specific* `[query:vector]` wording no longer fires in the extension-unavailable case, so a log consumer keyed on that exact string loses that one signal — which is the same thing the stale comment now misdescribes.

### CI status

**FAIL:** `CI Gate` (consequence), `tests / ubuntu / coverage 3/3` (the three `calltool-dispatch.test.ts` tests), `Vercel` (deploy authorization, not code). **PASS:** coverage 1/3 and 2/3, all three Windows platform-sensitive shards, all three macOS shards, `quality/{typecheck,lint,format,typecheck-web}`, CodeQL, gitleaks, node-floor compat 22.18, packaged install smoke (ubuntu + windows), tree-sitter ABI on all three OSes, benchmarks, eval/containment. The PR has never had a green run — the first commit's ubuntu shards were cancelled before reaching these tests.

---

*Automated multi-tool digest (Claude swarm + ce personas, Codex gpt-5.5/xhigh swarm + ce + adversarial). Verify findings before acting on them.*

azizur100389:
Addressed the review in 3a4528dd: fixed CI mock parity, added syntax-aware VECTOR detection including quoted procedure identifiers, separated load/query diagnostics, and moved preflight before checkout to avoid saturated-pool deadlock. The 300 affected tests, typecheck, lint, format, diff checks, and change-impact validation pass.

koriyoshi2041 (approved):
Verified the current head 437dd452 locally. The lazy VECTOR boundary is covered on both sides: exact-read pool initialization does not probe VECTOR, while direct QUERY_VECTOR_INDEX calls and semantic search load it before checkout; concurrent callers coalesce on one load and rejected loads retain the exact-scan fallback.

Commands:
- npm test -- test/unit/lbug-pool-fts-load.test.ts test/unit/local-backend-lazy-vector.test.ts test/unit/local-backend-embedding-dims-warn.test.ts test/unit/calltool-dispatch.test.ts --reporter=dot — 280 passed
- npx tsc --noEmit — passed
- git diff --check origin/main...HEAD — passed

The hosted matrix also reports 18,941 passing tests; its red status is the known Windows packaged-install timeout, not a failed assertion in this diff.

PR #3046, opened 2026-08-25
fix(ingestion): preserve object handler identity

### LINKED ISSUE #3041: Same-named nested functions (Convex handler:) collapse to one node; CALLS from every handler are re-attached to the first, and downstream impact returns 0

azizur100389:
## Summary

- Preserve distinct identities for same-named function-valued object members.
- Emit owner `HAS_METHOD` edges and follow owned callables in default downstream impact without changing explicit relation depth.
- Invalidate stale parse caches; existing indexes can be refreshed with `gitnexus analyze --force`.

## Validation

- 44 focused tests passed.

Fixes #3041

---

## Summary

*This appears to be a broad, high-consequence ingestion and impact-analysis change centered on preserving object handler identity across scope resolution, callable identification, and local traversal.*

**🔴 CRITICAL blast radius. An ingestion and local MCP backend change that reaches widely through the repository’s resolved call and impact-analysis paths.**

The core change touches scope resolution and parsing in `gitnexus/src/core/ingestion/scope-resolution/graph-bridge/ids.ts`, `gitnexus/src/core/ingestion/scope-resolution/graph-bridge/node-lookup.ts`, `gitnexus/src/core/ingestion/utils/ast-helpers.ts`, `gitnexus/src/core/ingestion/workers/callable-id.ts`, and `gitnexus/src/core/ingestion/workers/parse-worker.ts`. The key symbols include `resolveDefGraphId`, `buildGraphNodeLookup`, `findObjectLiteralBindingInfo`, `boundCallableStartRow`, `enclosingCallablePrefix`, `callableOwnQualifiedName`, `findEnclosingFunctionId`, and `processFileGroup`. `gitnexus/src/storage/parse-cache.ts` also changes `SCHEMA_BUMP` and `GITNEXUS_PKG_VERSION`, so cache compatibility belongs in the review path.

The main concentration is `gitnexus/src/mcp/local/local-backend.ts`, where 13 changed symbols include `LocalBackend`, `_impactImpl`, `_runImpactBFS`, and `callToolAtGroupRepo`. The graph reports 105 dependents and 65 affected flows, making the local impact traversal and the object-literal ownership resolution the best places to review first. The integration and unit coverage in `gitnexus/test/integration/object-literal-impact.test.ts`, `gitnexus/test/integration/object-literal-owner-resolution.test.ts`, `gitnexus/test/unit/impact-batching-grouping.test.ts`, and `gitnexus/test/unit/scope-resolution/node-lookup-determinism.test.ts` provides the clearest validation surface.

_Added by GitNexus for PR #3046. Edit freely — this block is replaced on the next review, everything above it is left untouched._


azizur100389:
Addressed the automated finding in 07d6cf37: implicit object-member expansion is distinct, deterministic, capped at 5,000, and reports `partial: true` when truncated.

Validated: 48 affected tests, typecheck, ESLint, and Prettier.

magyargergo (commented):
# Multi-engine review — PR #3046 `fix(ingestion): preserve object handler identity`

**Engines.** Claude ran the GitNexus PR swarm (facts/risk/security/test-CI/docs-DoD) **and** the ce-code-review personas (correctness, adversarial, testing, performance, maintainability, project-standards, data-migration). Codex (GPT-5.5, xhigh) ran three legs — **swarm solo, ce, and native adversarial; all three produced real findings** (no empty/skill-not-found legs). A synthesis pass merged them, and the coordinator re-ran the load-bearing claims against a **clean head build and a clean base build at `031e1237`**.

**Merge status:** `mergeable`. **Branch hygiene:** `clean feature/fix PR` — two commits, one domain, no merge commits, no whitespace damage, no bidi.

---

## What the reviews validated

- **#3041 is genuinely fixed for the shape it reports**, confirmed base-vs-head by the coordinator (table below), not taken on trust.
- **The three id-building phases agree** (`callableOwnQualifiedName`, `findEnclosingFunctionId`, `processFileGroup`) across ~30 adversarial fixtures: nested closures, literals inside class methods, `as const`/`satisfies` wrappers, function-local and block-scoped literals, `let`/`var` owners. Zero dangling `CALLS` sources or targets.
- **Blast radius is JS/TS only.** `findObjectLiteralBindingInfo` needs both an `object` and a `variable_declarator` node type; across 10 npm + 5 vendored grammars only javascript/typescript/tsx have both. Python/Ruby/Go emit byte-identical ids on head and base.
- **The impact cap is right:** sentinel fetch (`LIMIT CAP+1`), `ORDER BY id` before `LIMIT` (deterministic, #2787-clean), overflow surfaces as `partial: true`, unit-tested with 5001 rows.
- **Upgrade is safe in practice.** A lane built base and head and ran a real in-place upgrade: a plain `analyze` (no `--force`) forces a full rebuild, all member shapes migrate, **zero orphans**; downgrade is equally loud.
- **Security clean:** the interpolated `LIMIT` is a module-local integer, `symId` is bound, owner names are `identifier`-gated so the id encoding cannot be broken, no bidi/homoglyphs, `logQueryError` leaks nothing.
- **`SCHEMA_BUMP = 73` is free** — main is 72; across all 50 open PRs the only other claims are #3044=74, #3017=71, #2840=71, #1616=2. Re-check at merge as the ledger says.
- Choosing to qualify `Function` is a **fix for a case #2856 never considered**, not a reversal of a considered decision — that PR's message scopes its `Property`-only gate to `Method` history and newly-indexed keys, and never mentions `Function`.

---

## Coordinator verification — base `031e1237` vs head `07d6cf37`, both clean builds

Four fixtures through the real worker pipeline on each side.

**The core fix works.**
```ts
export const first  = query({ handler: (ctx) => helperA(ctx) });
export const second = query({ handler: (ctx) => helperB(ctx) });
```
| | base | head |
|---|---|---|
| nodes | ONE `Function:src/convex.ts:handler` | `first.handler` **and** `second.handler` |
| CALLS | `handler -> helperA` **and** `handler -> helperB` (the #3041 bug verbatim) | `first.handler -> helperA`, `second.handler -> helperB` ✅ |
| HAS_METHOD | none | one per owner ✅ |

**The same run also shows three things the PR does not mention:**

1. **`File -[DEFINES]-> Function` is lost for every owner-qualified handler, including the headline shape.** Base emits `DEFINES File:src/convex.ts -> Function:src/convex.ts:handler`; head emits **no** `DEFINES` for `first.handler` or `second.handler` (nor for `p.q.r`, `z.r`, `arr.handle`). No broken consumer was found — the wiki export has an unlabelled-owner arm and the two `local-backend` `DEFINES` reads are Class/Interface-only — so this is graded P2 for silently breaking a graph invariant on the primary path, untested and unmentioned, not for a demonstrated failure.
2. **Dotted quoted key — the P1 below, with a base control.**
   `export const p = { 'q.r': (x)=>hC(x) }; export const z = { r: (x)=>hD(x) };`
   base: `CALLS q.r -> hC` ✅ and `CALLS r -> hD` ✅.
   head: `CALLS z.r -> hC` **and** `CALLS z.r -> hD`; `p.q.r` has **zero** callees.
3. **Array and shorthand collapses persist** (findings 2 and 3): `Function:src/arr.ts:arr.handle` carries both array elements' calls and gains a false `Const:arr -HAS_METHOD-> arr.handle`; `Method:src/short.ts:run#1` is byte-identical on both sides, carries `sA` **and** `sB`, and both `alpha` and `beta` point at it.

---

## Headline

**(P1) The owner qualifier enters the node id but not the resolver's name keys, so a dotted object key now loses or *fabricates* its `CALLS` edge. This is a regression against base.**

`buildGraphNodeLookup` derives a node's qualified key from its **id** (`graph-bridge/node-lookup.ts:161-162`) — `Function` nodes carry no `qualifiedName`, so the key moves from `handler` to `first.handler`. `resolveDefGraphId` still builds its name forms from the scope-resolution def's own qualified name, which knows nothing about object-literal owners. The whole name-key family therefore misses for exactly the nodes this PR creates, and resolution stands on the position key alone. Where the position key is also ambiguous, the answer is a *wrong edge at full confidence* plus a *confident zero* — the failure class #3041 is about.

Dotted string keys are ordinary JS (event maps, i18n keys, route tables). The same root cause aliases two same-named handlers declared on one source line: `a.handler` takes both calls and `b.handler` answers `risk: LOW, impactedCount: 0`.

---

## Findings

| # | Sev | Blocks merge | Location | Finding |
|---|---|---|---|---|
| 1 | **P1** | yes | `parse-worker.ts:2418` | Owner-qualified id not mirrored into the bridge's name keys → dotted key loses/fabricates `CALLS` (regression); same-line siblings alias. |
| 2 | P2 | no | `parse-worker.ts:2410` | An `array` between literal and declarator is crossed silently: `const arr = [{h:…},{h:…}]` mints `Function:arr.handle` (still **both** bodies' calls) **and a new false `Const:arr -HAS_METHOD-> arr.handle`** — `arr.handle` does not exist at runtime. |
| 3 | P2 | no | `parse-worker.ts:2416` | Spelling-scoped fix + false justification: shorthand `{ handle() {} }` siblings still collapse onto one `Method:<f>:handle#1` carrying both bodies' `CALLS`; the "existing owner edge" is the *same node* on both sides. |
| 4 | P2 | no | `local-backend.ts:7156` | New catch never sets `traversalComplete = false` → a failed seed query returns `risk: LOW, impactedCount: 0` with no `partial`. **Already raised by the repo's own `gitnexus-check` bot on `07d6cf37`, still unresolved**; the author's reply addressed the volume cap only. |
| 5 | P2 | no | `local-backend.ts:7133` | The guard omits `hasExplicitRelationTypes` (computed at `:6196`, used at `:6326`/`:6532`), so `relationTypes:['CALLS']` still walks `HAS_METHOD` as a hidden depth-0 seed. |
| 6 | P2 | no | `local-backend.ts:7139` | Seed matches `(member:Function)` only; `Const -HAS_METHOD-> Method` is the shipped shape for shorthand members, so half the object-binding surface stays unreachable downstream. |
| 7 | P2 | no | `local-backend.ts:7148` | Seeded members never enter `impacted`; their callees land at depth 1, so `maxDepth:1` reaches two real hops and `directCount` inflates the risk ladder. (Seed-not-reported itself is the pre-existing JVM Fix #480 convention — what is new is that it now applies to `Const` bindings.) |
| 8 | P2 | no | `parse-cache.ts:575` | The ledger entry documents the parse-layer guarantee correctly but is silent on the graph-layer one; the re-emit is forced solely by `analyzerRunnerIdentitiesEqual` (`run-analyze.ts:1538`) — undocumented and untested here. |
| 9 | P2 | no | body-only (`parse-worker.ts:2885`/`:2926`, unchanged lines) | Every owner-qualified handler loses its `File -[DEFINES]-> Function` edge — reproduced base-vs-head on the PR's own Convex fixture. |

Findings 1–8 are posted inline; 9 is body-only because its lines are outside every diff hunk.

---

## Adjudicated disputes

- **ANALYSIS_FEATURES / mixed id shapes.** Codex swarm (P1, "blocks merge"), Codex ce (P1) and Codex adversarial (high) all claim an incremental run can leave old collapsed ids beside new ones. **Refuted as an outcome, confirmed as a code fact.** `isIncremental` (`run-analyze.ts:1900`) takes no input that moves with `SCHEMA_BUMP` — that half is right — but `analyzerRunnerIdentitiesEqual` sets `force: true` first (`:1538`), and its `build.digest` content-hashes the whole build root, so any upgrade or downgrade full-rebuilds. Verified end to end on clean base/head installs: zero orphans, parse cache `72+1.6.9 → 73+1.6.9`. `ANALYSIS_FEATURES` holds only Spring/Java graph-phase features and is the wrong lever. What survives is finding 8: the guarantee is real but incidental and undocumented. Not a merge blocker.
- **Explicit `relationTypes:['CALLS']`** (Codex adversarial, medium) vs the Claude swarm's "jointly correct". Codex is right; the file's own idiom for "may we widen the traversal?" uses `hasExplicitRelationTypes` at both other sites, and `tools.ts:500` publishes the opposite contract. The pre-existing JVM seed has no gate at all, so this is a consistency defect rather than a new sin — P2, one-line fix.
- **Shorthand spelling.** Not a regression — `Method` collapsed on base too, and #2856 froze those ids deliberately. What is wrong is the *justification*: both owners' `HAS_METHOD` edges point at the **same** node, so the owner edge cannot disambiguate identity. Reproduced by three lanes and by the coordinator.
- **Seed-at-depth-0 (finding 7).** The coordinator refuted the claim that seeding into `visited`+`frontier` without appending to `impacted` is *novel* — it is the JVM Fix #480 convention, and that block does the same. What is genuinely new is that the convention now applies to `Const`/`Variable` bindings, where the seeded member **is** the payload rather than an implementation detail: `maxDepth: 1` now reaches two real hops, and depth-1 `directCount` feeds the risk thresholds. Kept at P2 on that basis, not on the pattern itself.
- **Mutation (b) leaves every test green.** A lane deleted only the `findEnclosingFunctionId` hunk, rebuilt, and all four #3041 tests still passed — while instrumentation showed that site computing the collapsed `Function:src/convex.ts:handler`. Verdict: **test-coverage gap, not dead code.** The bridge's position join repairs the wrong id for ordinary shapes; findings 1 and 2 are exactly the shapes where that repair fails, so the hunk is load-bearing and untested at once. Assert on the raw `ParseWorkerResult` sourceIds, not only on the resolved graph.

---

## Lower priority

- **P3 — performance.** The new lookup fires for *every* plain top-level JS/TS function (`prefix===undefined && classInfo===null && label==='Function'` is the common case), ~3–5 µs/call for a guaranteed miss (measured). The underlying `.parent` ancestor walk is ~O(depth²) on the pinned tree-sitter (107 ms for one call at depth 1000, measured), and the `findEnclosingFunctionId` site is not memoised the way `cachedFindEnclosingClassInfo` is. Cheap fixes: a syntactic pre-check plus a `WeakMap` keyed on the node.
- **P3 — stale attribution in the pin test.** `incremental-parse-cache.test.ts:233` still cites `(#2766)` though this bump belongs to #3041, and the rationale above it still narrates the 70→71 decision while this PR's base carries 72. Same in `parse-cache.ts`, whose ledger keeps the v72-era sentence "Every open PR touching gitnexus/ was scanned; #3017 is the only other claimant" above the new v73 entry.
- **P3 — `OBJECT_CALLABLE_MEMBER_CAP = 5000`** is a per-call `const` with no derivation and no env gate; truncation is deterministic but alphabetical.
- **P3 — release notes:** the PR body's `--force` is unnecessary (a plain `analyze` rebuilds); every renamed node's cached embedding is dropped and re-embedded on the migration run; group bridges hold per-repo `symbolUid`s this migration invalidates until the next `gitnexus group sync` (`contractsStale` does flip, so it is visible).

**Test gaps worth closing:** no fixture for a dotted quoted key, same-line siblings, an array of literals, `export default { … }`, a sibling *pair* of shorthand methods, a `Method`-labelled member under the new seed, the default-`relationTypes` depth bucket, or a seed-query failure asserting `partial: true`. `object-literal-impact.test.ts` hand-seeds the graph, so the two halves of this PR are never tested against each other end to end.

---

## Probed and cleared (validation is a feature)

`labelOverride` divergence across the three phases · `findMemberAssignmentOwnerInfo` asymmetry (prototype/`this` shapes are labelled `Method`, so the gate never fires — verified with both RHS spellings) · sequential-path parity (that DAG died in #942; `parse-worker.ts` is the sole compute site) · language-agnosticism of the `label === 'Function'` gate (`NodeLabel` + `provider.labelOverride`; JS/TS grammars only) · `ANALYSIS_FEATURES` mixed-shape hazard · `SCHEMA_BUMP` exact clash · orphaned old-shape nodes after upgrade · rollback/downgrade corruption · persisted `reason`-string consumers (repo-wide grep: one hit, the emit site) · Cypher `LIMIT` interpolation and id-encoding injection · unordered-`LIMIT` non-determinism · `traversalComplete` being re-raised after overflow · new id collisions from the qualifier · the "`object-literal-impact.test.ts` missing from the vitest `lbug-db` project" P0 (~30 files with `poolAdapter: true` sit outside that list; the config itself calls it deliberate pre-existing drift) · id non-determinism (a lane's anomaly was a contaminated build; withdrawn).

---

## CI

All checks green on `07d6cf37` — format/lint/typecheck, ubuntu coverage 1–3 + merge, windows 1–3, macos 1–3, CodeQL (js-ts, python), gitleaks, CI Gate; 18697/18770 tests pass. The only failure is **Vercel "Authorization required to deploy"** — infra/permissions, unrelated to the diff, and the sole reason `mergeStateStatus` is `UNSTABLE`.

## Coverage limits

- **Worktree contamination.** Mid-review, one lane mutated `parse-worker.ts` and rebuilt `dist/` from it (at one point `shouldObjectOwnerQualifyCallable` returned `false` and the `findEnclosingFunctionId` hunk was deleted). Affected results were discarded and re-run by their own lanes; **every reproduction in the Coordinator-verification section and in findings 1, 2, 3 and 9 was re-run afterwards against a clean head build with a clean base control.**
- `[code-read]` only (not executed): findings 4, 5, and the `maxDepth` half of 7. Finding 6 is `[reproduced]` for the graph shape and `[code-read]` for the query. The upgrade/orphan verdict is `[reproduced]` by a single lane on its own installs.
- Nobody got a green run of the *full* suite on this host (load 20–75; the worker pool repeatedly missed its 5 s readiness handshake). CI's green run is the evidence for suite health.
- No human has reviewed this PR; the only prior reviewer is the repo's `gitnexus-check` bot, whose second finding (4 above) is still open.

## Open questions for the author

1. Is the dotted-key/same-line aliasing (finding 1) acceptable to fix here, or does it want its own PR? It is the one item graded as blocking.
2. Should `Method` be owner-qualified in this PR too (finding 3)? The fleet-wide rebuild is already being paid once; doing it later costs a second one.
3. Was dropping `File -[DEFINES]->` for owner-qualified handlers (finding 9) intended?

## Verdict

**production-ready with minor follow-ups** — once finding 1 is settled. The core fix is real and verified against a base control; findings 2–9 are quality and completeness items, none of which blocks on its own.

*Automated multi-tool digest (Claude swarm + ce personas, Codex GPT-5.5 swarm/ce/adversarial), merged, critic-gated, and re-verified against clean base and head builds. Verify each finding before acting.*

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -2382,13 +2406,16 @@ const processFileGroup = (
...
             findObjectLiteralBindingInfo(definitionNode, file.path, {
-              // Only `Property` opts into the qualifier; `Method` ids must stay
-              // byte-identical or every object-literal method in every indexed
-              // repo changes id.
-              includeOwnerName: nodeLabel === 'Property',
+              // Function-valued properties need the owner in their identity or
+              // same-named handlers in sibling objects collapse (#3041).
+              // `Method` stays byte-identical; its existing owner edge is enough.
+              includeOwnerName:
+                shouldObjectOwnerQualifyCallable(nodeLabel) || nodeLabel === 'Property',
```
magyargergo, line 2435: **[P1 — blocks merge] The owner qualifier enters the node id but not the resolver's name keys, so a dotted object key now loses or fabricates its `CALLS` edge (regression vs base).**

`buildGraphNodeLookup` derives a node's qualified key from its **id** (`graph-bridge/node-lookup.ts:161-162`) — `Function` nodes carry no `qualifiedName`, so the key moves from `handler` to `first.handler`. `resolveDefGraphId` still builds its name forms from the scope-resolution def's own qualified name, which knows nothing about object-literal owners, so the whole name-key family now misses for exactly the nodes this PR creates and resolution stands on the position key alone. Two shapes fall through it.

Reproduced with a clean head build and a clean base build at `031e1237`:

```ts
function hC(x: number) { return x; }
function hD(x: number) { return x; }
export const p = { 'q.r': (x: number) => hC(x) };
export const z = { r:     (x: number) => hD(x) };
```

base `031e1237`:
```
CALLS Function:src/dotted.ts:q.r -> Function:src/dotted.ts:hC
CALLS Function:src/dotted.ts:r   -> Function:src/dotted.ts:hD
```
head `07d6cf37`:
```
CALLS Function:src/dotted.ts:z.r -> Function:src/dotted.ts:hC   <- hC belongs to p['q.r']
CALLS Function:src/dotted.ts:z.r -> Function:src/dotted.ts:hD
                                    (Function:src/dotted.ts:p.q.r has ZERO callees)
```
A fabricated edge at full confidence plus a confident zero — the failure class #3041 is about. Dotted string keys are ordinary JS: event maps, i18n keys, route tables.

The same root cause also aliases two same-named handlers declared on one source line (`export const a = { handler: (x)=>hA(x) }; export const b = { handler: (x)=>hB(x) };` on one line) — `a.handler` takes both calls and `b.handler` answers `risk: LOW, impactedCount: 0`.

Two distinct remedies, please don't conflate them:
- **dotted key** — escape `.` in the member segment, or skip owner-qualification when the key already contains one; alternatively stamp `qualifiedName` on the emitted symbol so the bridge and the def agree.
- **same-line siblings** — the position key is ambiguous for two same-label same-line callables and the fallback is the label-agnostic first-write-wins `simpleKey`; `localNameKey`'s fail-closed guard is registered only for `@row:col` qualifiers, so it cannot catch this.

Worth fixtures asserting the `CALLS` **sourceId** for a quoted dotted key and for two same-named handlers on one line. `[reproduced]`

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -2382,13 +2406,16 @@ const processFileGroup = (
       // and the merged name then looks workspace-unique to name inference,
       // which resolves reads of it to a node representing both.
       const objectLiteralOwnerInfo =
-        !enclosingClassId && (nodeLabel === 'Method' || nodeLabel === 'Property') && definitionNode
+        !enclosingClassId &&
+        (nodeLabel === 'Function' || nodeLabel === 'Method' || nodeLabel === 'Property') &&
```
magyargergo, line 2431: **[P2] Admitting `Function` here reaches through arrays, so an array of object literals gets a false owner-qualified identity *and* a new false `HAS_METHOD` edge.**

`findObjectLiteralBindingInfo` phase A counts `object` ancestors and bails only on function/class containers (`ast-helpers.ts:1285-1300`); an `array` node between the literal and the `variable_declarator` is crossed silently. With `Function` now opted in:

```ts
export const arr = [{ handle: (x: number) => aA(x) }, { handle: (x: number) => aB(x) }];
```
head emits (reproduced, clean build):
```
NODE  Function:src/arr.ts:arr.handle
CALLS Function:src/arr.ts:arr.handle -> aA
CALLS Function:src/arr.ts:arr.handle -> aB      <- still ONE node for two functions
HAS_METHOD Const:src/arr.ts:arr -> Function:src/arr.ts:arr.handle
```
`arr.handle` is `undefined` at runtime. Base emitted the visibly-degenerate unqualified `handle` node and no ownership edge — so the PR replaces an obviously-merged node with a confidently-named one that is still two different functions merged, and asserts an ownership that does not exist. Route tables, middleware arrays and test tables (`const cases = [{ run(){} }, { run(){} }]`) are exactly this shape.

Fix: in phase A, return `null` when an `array` is crossed before the declarator — the same safe under-approximation the `objectDepth > 1` branch already takes — or qualify with the element index. `[reproduced]`

gitnexus/src/core/ingestion/workers/parse-worker.ts
```diff
@@ -2382,13 +2406,16 @@ const processFileGroup = (
...
+        definitionNode
           ? (findMemberAssignmentOwnerInfo(definitionNode, file.path) ??
             findObjectLiteralBindingInfo(definitionNode, file.path, {
-              // Only `Property` opts into the qualifier; `Method` ids must stay
-              // byte-identical or every object-literal method in every indexed
-              // repo changes id.
-              includeOwnerName: nodeLabel === 'Property',
+              // Function-valued properties need the owner in their identity or
+              // same-named handlers in sibling objects collapse (#3041).
+              // `Method` stays byte-identical; its existing owner edge is enough.
```
magyargergo: **[P2] "its existing owner edge is enough" is falsified by the data this produces — the owner edge's target is *shared*.**

```ts
export const alpha = { run(x: number) { return sA(x); } };
export const beta  = { run(x: number) { return sB(x); } };
```
head (reproduced, clean build — byte-identical to base):
```
NODE       Method:src/short.ts:run#1                 <- ONE node for both
HAS_METHOD Const:src/short.ts:alpha -> Method:src/short.ts:run#1
HAS_METHOD Const:src/short.ts:beta  -> Method:src/short.ts:run#1
CALLS      Method:src/short.ts:run#1 -> sA
CALLS      Method:src/short.ts:run#1 -> sB          <- cross-attribution, i.e. #3041 verbatim
```
The owner edge disambiguates *ownership*, not *identity*, and #3041 is about identity: `impact` on `alpha` reports `sB`. This is **not a regression** — base collapses these too, and #2856 froze `Method` ids deliberately — but after this PR the spelling decides the answer: `{ handle: fn }` separates, `{ handle() {} }` collapses, in the same file. The seed query in `local-backend.ts` matches `(member:Function)` only, so shorthand members are excluded from *both* halves of this PR.

Either extend `shouldObjectOwnerQualifyCallable` to `Method` here (the fleet-wide rebuild is already being paid once, and doing it later costs a second one), or replace this comment with the accurate one — "`Method` ids are frozen for compatibility; same-named shorthand methods in sibling literals still collapse, tracked in #NNNN" — and say so in the PR body so #3041 is not closed as fully fixed. `[reproduced]`

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -7118,6 +7121,42 @@ export class LocalBackend {
...
+        if (memberRows.length > OBJECT_CALLABLE_MEMBER_CAP) traversalComplete = false;
+        for (const row of memberRows.slice(0, OBJECT_CALLABLE_MEMBER_CAP)) {
+          const memberId = row.id || row[0];
+          if (memberId && !visited.has(memberId)) {
+            visited.add(memberId);
+            frontier.push(memberId);
+          }
+        }
+      } catch (e) {
+        logQueryError('impact:object-callable-expansion', e);
```
magyargergo, line 7220: **[P2] A failed seed query returns a confident clean answer. (This is the `gitnexus-check` bot's second finding on `07d6cf37` — the reply addressed the volume cap, not this.)**

`traversalComplete` is set `false` only at `:7147` (cap overflow), `:7307` (depth loop) and `:7511` (chunk cap), and it is the sole source of `partial` (`:7709`). For an object binding this query **is** the whole downstream answer: if it throws, `frontier` stays `[symId]`, the binding has no outgoing `CALLS`/`IMPORTS`/`USES` of its own, `impacted` is empty, and the risk ladder falls through to `risk: 'LOW', impactedCount: 0` with **no `partial`** — the code deliberately declines to mark an empty *downstream* walk UNKNOWN (`:7648-7651`), so there is no second signal. A transient lock, a timeout or a native fault therefore reports "changing this binding affects nothing" to the tool agents are told to consult before every edit.

One line — `traversalComplete = false;` beside the log, matching the depth-traversal catch at `:7303-7308`. The pre-existing `impact:class-node-expansion` catch at `:7118-7120` has the same gap; fixing both is cheap. Worth a unit test where the seed query rejects and `result.partial === true`. `[code-read]`

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -7118,6 +7121,42 @@ export class LocalBackend {
...
+    // Function-valued properties on exported object bindings are represented
+    // as Const/Variable -[:HAS_METHOD]-> Function. HAS_METHOD is intentionally
+    // absent from the default usage traversal, but downstream impact on the
+    // binding still needs to enter its own callable member before following
+    // CALLS.
+    if (
+      direction === 'downstream' &&
+      (symType === 'Const' || symType === 'Variable') &&
+      relationTypes.includes('CALLS') &&
+      !relationTypes.includes('HAS_METHOD')
```
magyargergo: **[P2] An explicit `relationTypes: ['CALLS']` still gets the implicit `HAS_METHOD` seed.**

`hasExplicitRelationTypes` is already computed at `:6196`, and this file's own idiom for "may we implicitly widen the traversal?" uses **both** halves — `:6324-6328` and `:6530-6534`:
```ts
(symType === 'Class' || symType === 'Interface') && !hasExplicitRelationTypes && !relationTypes.includes('ACCESSES')
```
This block uses only the second half, and `hasExplicitRelationTypes` is never threaded into `_runImpactBFS`'s `opts`. Trigger: `impact({ target: <a Const binding>, direction: 'downstream', relationTypes: ['CALLS'] })` — the caller explicitly filtered to CALLS, yet the walk still enters `Const -HAS_METHOD-> Function` as a depth-0 seed and returns the handler's callees. `tools.ts:500` publishes the opposite contract ("For class members, include HAS_METHOD and HAS_PROPERTY in relationTypes").

The pre-existing Fix #480 JVM seed has no `relationTypes` gate at all, so this is a consistency defect rather than a new sin — but this block explicitly reasons about `relationTypes`, so it should use the file's complete idiom. Thread the flag into the opts and add it to the guard, with a test proving `['CALLS']` does not traverse HAS_METHOD. `[code-read]`

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -7118,6 +7121,42 @@ export class LocalBackend {
...
+    ) {
+      try {
+        const memberRows = await executeParameterized(
+          repo.lbugPath,
+          `
+          MATCH (n)-[hm:CodeRelation]->(member:Function)
```
magyargergo, line 7185: **[P2] The label pin excludes the shipped `Method` representation, so half the object-binding surface still cannot be entered.**

An object binding's callable members carry two labels, decided purely by spelling: `{ h: () => {} }` → `Function`, `{ h() {} }` → `Method` (`inferFunctionLabel`). `Const -[:HAS_METHOD]-> Method` is the pre-existing shipped shape from #1358/#1718 — this very suite pins `generateId('Method', 'src/service.ts:getUser#1')` — and it is what head actually emits (reproduced):
```
HAS_METHOD Const:src/short.ts:alpha -> Method:src/short.ts:run#1
```
`HAS_METHOD` is absent from the default relation set, so for
```ts
export const service = { getUser(id: string) { return helper(id); } };
```
`impact({ target: 'service', direction: 'downstream' })` still returns nothing and `helper` is never reported. The PR's stated goal is delivered for one spelling only.

Fix: drop the `:Function` pin and gate on the label instead (`AND label(member) IN ['Function','Method']`), or add a second `UNION`ed arm — and put a `Method`-labelled member in the `object-literal-impact.test.ts` seed so the pin cannot silently regress. `[reproduced]` (graph shape) `[code-read]` (query)

gitnexus/src/mcp/local/local-backend.ts
```diff
@@ -7118,6 +7121,42 @@ export class LocalBackend {
...
+          WHERE n.id = $symId AND hm.type = 'HAS_METHOD'
+          RETURN DISTINCT member.id AS id
+          ORDER BY id
+          LIMIT ${OBJECT_CALLABLE_MEMBER_CAP + 1}
+        `,
+          { symId },
+        );
+        if (memberRows.length > OBJECT_CALLABLE_MEMBER_CAP) traversalComplete = false;
+        for (const row of memberRows.slice(0, OBJECT_CALLABLE_MEMBER_CAP)) {
```
magyargergo, line 7203: **[P2] Seeded members never enter `impacted`, so their callees are reported one hop too shallow and `maxDepth` no longer bounds the walk for `Const` bindings.**

To be fair to the design: seeding into `visited` + `frontier` without appending to `impacted` is the pre-existing convention set by the JVM block right above (Fix #480), which does exactly the same for Constructors and the owning File. That is not the complaint. What is new is that the convention now applies to `Const`/`Variable` bindings, where the seeded member **is** the payload rather than an implementation detail of a class.

Measured against the PR's own seed graph (`first -HAS_METHOD-> handler -CALLS-> helperA`):

| call | result |
|---|---|
| `impact('first', downstream)` (default relations) | `helperA` at **depth 1**; `first.handler` absent entirely |
| same call with `maxDepth: 1` | **identical output** — the walk still reaches two real hops |
| `relationTypes: ['HAS_METHOD','CALLS']` | `first.handler` at 1, `helperA` at 2 |

Two consequences: `maxDepth` means something different for a `Const` binding than for every other symbol type, and `directCount = (grouped[1] || []).length` feeds the risk thresholds (`>=5` MEDIUM, `>=15` HIGH, `>=30` CRITICAL) — a module with 30 exported handlers each calling one helper reports `direct: 30` → CRITICAL, where the true direct fan-out is the 30 handlers.

Either push the members into `impacted` at depth 1 and let their callees land at 2, or pin the current convention with a test asserting `byDepth['1']` for the **default** relation set and that `maxDepth: 1` stops before the member's callees. The PR's tests use `toContain`/`not.toContain`, so both conventions pass today. `[reproduced]`

gitnexus/src/storage/parse-cache.ts
```diff
@@ -568,8 +568,13 @@ import type { ParseWorkerResult } from '../core/ingestion/workers/parse-worker.j
 // and the v37/v38 clash it was written for: the next free value above every
 // IN-FLIGHT claim, not above origin/main. Every open PR touching gitnexus/ was
 // scanned; #3017 is the only other claimant.
+//
+// 72 -> 73 makes function-valued object members owner-qualified (#3041).
+// A warm v72 cache replays the old collapsed Function ids and omits the new
+// Const -> Function HAS_METHOD ownership edges, so this identity correction
+// must invalidate unchanged files rather than waiting for a source edit.
```
magyargergo: **[P2] This entry documents the parse-layer guarantee accurately, but the graph-layer one it depends on is undocumented and untested.**

To be clear about what is right: `SCHEMA_BUMP` genuinely does make unchanged files re-parse — `parse-cache.ts:861,885` (`data.version !== PARSE_CACHE_VERSION → emptyCache`) and `parse-impl.ts:774` (`loadDurableParsedFileIndex`) are both parse-layer readers, and this bump is necessary.

What the ledger does not say is that the parse bump alone cannot rewrite the persisted graph. The decision to do that is `isIncremental` (`run-analyze.ts:1900`), and **none of its inputs moves with `SCHEMA_BUMP`** — `schema.ts` is byte-identical across this merge and `ANALYSIS_FEATURES` (`:227`) gains no entry. If a run reached that predicate with `force` still false, every file would re-parse into new-shape ids while the writeback deleted-and-inserted only `hashDiff.toWrite`, leaving unchanged files' old-shape nodes beside changed files' new-shape edges.

What actually prevents that is `analyzerRunnerIdentitiesEqual` at `run-analyze.ts:1538`, whose `build.digest` content-hashes the whole build root, so any upgrade or downgrade sets `force: true` first. Verified end to end on clean base and head installs: a plain post-upgrade `analyze` logs that guard, rebuilds fully, and leaves zero orphans. So the outcome is safe — but incidentally so.

Please name the real guard here — *"the parse-cache bump makes unchanged files re-parse; the graph re-emit is forced separately by `analyzerRunnerIdentitiesEqual` (run-analyze.ts:1538) — an id-shape change needs BOTH"* — and consider a test that a stamped older `runnerIdentity` forces the full rebuild. That guard rebuilds the whole index on every upgrade and is an obvious future perf target; when someone relaxes it, this change starts producing half-migrated graphs and the comment that should stop them points only at the parse cache. `[reproduced]` (upgrade) `[code-read]` (mechanism)

azizur100389:
Addressed the review in 547de250: aligned exact position/name lookup, owner-qualified Function and Method identities, kept array-contained members distinct without false ownership, preserved DEFINES edges, made seed failures partial, honored explicit relation filters, corrected depth semantics, and documented cache migration guards. Focused and resolver suites, typecheck, lint, format, diff checks, and change-impact validation pass.

magyargergo:
@azizur100389 can you please resolve merge conflicts? 🙏

azizur100389:
Resolved the latest main conflicts in c4b6ccba. The combined parse-cache migration now uses schema 76 because upstream owns 74 and open PR #3017 claims 75. Validated 89 focused tests, the 28-test cache suite, typecheck, lint, format, diff checks, and change-impact analysis.

PR #3061, opened 2026-08-27
fix(ingestion): treat an empty constant fold as a resolved route path, not a skip

glier:
## Summary

`foldJavaOperands` returned `null` for a constant that folds to the empty string, so `@GetMapping(ROOT)` with `static final String ROOT = "";` produced no route while the literal `@GetMapping("")` on the same controller produced one. `null` from that function is the skip floor — "a piece was unresolvable" — and an empty result is not that case.

## Motivation / context

Found by the review bot on #3059 (which adds the Kotlin binding). Investigating it showed the collapse is an anomaly in one function rather than a deliberate rule, on four independent counts:

1. **The shared core does not collapse.** `resolveOperands` in `constant-resolver.ts` returns `foldExpr(...)` unfiltered, so every language that reaches the fold through the language-agnostic path — JavaScript, Python, anything without a provider hook — already treats `""` as a result.
2. **The sibling function does not collapse.** `resolveJavaConstant` returns `""` for an empty constant, and that is already pinned by an existing test: the 30-level shared-descendant DAG case added with #2980 asserts `toBe('')`, its every intermediate value being the empty string.
3. **The literal branch does not drop it.** `java.ts:608-610` skips only when `unquoteLiteral(...) === null`, so `@GetMapping("")` is kept.
4. **The doc comment says otherwise.** It reserves `null` for "when any piece is unresolvable (skip floor)".

So the binding disagreed with the core, with its own sibling, with the literal path it is meant to mirror, and with its own documentation.

## Areas touched

## Scope & constraints

**In scope**

- Drop the `out === '' ? null : out` filter in `foldJavaOperands`, and explain in the doc comment why an empty fold is a success.
- Unit coverage for the new behaviour, its multi-operand form, and — as a control — that a genuinely unresolvable operand still returns `null`, so the two outcomes stay distinguishable.
- A group ↔ ingestion parity case asserting that the constant spelling now behaves like the literal one on each side.

- **The trailing-slash divergence between the two sides.** Writing the parity test surfaced that an empty method path renders as `/api/v1/` from the group extractor (`joinPath` inserts the separator ahead of an empty path) and as `/api/v1` from ingestion. This reproduces on `main` with the **literal** spelling, with no part of this change involved — I measured it before adjusting the test. It is a separate defect, so the test pins both strings as current behaviour rather than asserting the divergence away. Happy to open it as its own issue or PR.
- **The Kotlin binding.** `foldKotlinOperands` in #3059 mirrors this function line for line, including the filter. That PR will be updated to match once the shape here is agreed, so the two bindings never diverge.

## Implementation notes

An empty method path is meaningful precisely when the type carries the prefix: `@RequestMapping("/api/v1")` on the class and an empty path on the method is how Spring spells "the collection root". Dropping the route lost a real endpoint rather than avoiding a wrong one, which inverts what the skip floor is for.

The three call sites need no change — `http-patterns/java.ts`, and the ingestion hook wired at `languages/java.ts:245` — because each already branches on `=== null` and then uses the string as-is, which is exactly how they treat a literal empty path today.

## Testing & verification

- [x] `cd gitnexus && npm test` — **18697 passed, 4 failed.** All four failures are environmental and pre-existing on my macOS machine, in `analyze-index-lock-concurrency.test.ts`, `hooks.test.ts` and `review-agent-workflow.test.ts`; none references route extraction or constant resolution. The last two reproduce on a clean checkout of the base commit.
- [ ] `cd gitnexus && npm run test:integration` — not run separately; the integration specs ran as part of the suite above, after `npm run build`.
- [x] `npx prettier --check` and `npx eslint` on the three changed files — clean (eslint reports only pre-existing non-null-assertion warnings elsewhere in the file)
- [ ] `cd gitnexus-web && …` — web untouched

## Risk & rollout

**This changes the graph Java emits**, which is why it is not folded into #3059. Re-indexing a Spring service that uses an empty-valued constant as a method path will now emit a route where it previously emitted none. The added route is one the application genuinely serves; no previously emitted route changes or disappears.

Rollback is a one-line revert.

## Checklist

- [x] `AGENTS.md` / overlays unchanged, so no header or changelog update is due

---

## Summary

*A narrowly scoped Java ingestion change with critical blast posture, centered on route extraction and backed by focused resolver parity tests. The graph shows limited direct reach, but the affected execution flows make the resolver behavior important to review closely.*

**🔴 CRITICAL blast radius. A Java constant-folding change to `foldJavaOperands` in `gitnexus/src/core/ingestion/route-extractors/java-const-resolver.ts`, reaching one direct dependent across 12 affected flows.**

The implementation change is concentrated in the `Route-extractors` module, with related impact in `Http-patterns`. The accompanying coverage is in `gitnexus/test/unit/group/java-const-route-parity.test.ts` and `gitnexus/test/unit/java-route-const-resolver.test.ts`.

Review `foldJavaOperands` first, particularly how an empty constant fold is handled as a resolved route path rather than skipped. Then verify the route parity and resolver unit tests cover that behavior across the affected flows.

_Added by GitNexus for PR #3061. Edit freely — this block is replaced on the next review, everything above it is left untouched._

### DIFF
--- a/gitnexus/src/core/ingestion/route-extractors/java-const-resolver.ts
+++ b/gitnexus/src/core/ingestion/route-extractors/java-const-resolver.ts
@@ -619,12 +619,20 @@ function foldOperands(
 /**
  * Fold an inline operand list (e.g. `API_CIS_V1 + "summary/save"`) against
  * `fileKey`, or null when any piece is unresolvable (skip floor).
+ *
+ * An empty result is a SUCCESS, not a skip. `static final String ROOT = "";`
+ * folds to `""`, which `joinPath` then resolves against the type-level prefix
+ * exactly as it resolves the literal `@GetMapping("")`. Collapsing it into
+ * `null` would make a resolved-empty path indistinguishable from an
+ * unresolvable one — the skip floor is reserved for "could not fold", and
+ * nothing else in the resolver conflates the two: `resolveOperands` in the
+ * shared core returns `foldExpr` unfiltered, and `resolveJavaConstant` returns
+ * `""` for an empty constant.
  */
 export function foldJavaOperands(
   fileKey: string,
   operands: readonly Operand[],
   repo: RepoConstants,
 ): string | null {
-  const out = foldOperands(fileKey, operands, newFoldState(repo), 0);
-  return out === '' ? null : out;
+  return foldOperands(fileKey, operands, newFoldState(repo), 0);
 }
--- a/gitnexus/test/unit/group/java-const-route-parity.test.ts
+++ b/gitnexus/test/unit/group/java-const-route-parity.test.ts
@@ -211,3 +211,46 @@ public class OrderController {
     expect(ingestionRoutes(files)).toEqual(['POST /api/v1/orders']);
   });
 });
+
+describe('an empty-valued constant route matches its literal spelling', () => {
+  // Before this fix the two spellings diverged: `@GetMapping("")` was kept and
+  // `@GetMapping(ApiPaths.ROOT)` with `ROOT = ""` was dropped, because the fold
+  // collapsed a resolved-empty result into the skip floor. The class-level
+  // prefix is what makes an empty method path meaningful, so the parity is
+  // asserted with one present.
+  const ROOT_CONSTS = `package com.example;
+public class ApiPaths {
+  public static final String ROOT = "";
+}`;
+
+  const controller = (mapping: string): string => `package com.example;
+import com.example.ApiPaths;
+@RequestMapping("/api/v1")
+public class RootController {
+  @GetMapping(${mapping})
+  public void root() {}
+}`;
+
+  it('resolves the constant spelling to the same path as the literal one', () => {
+    const literal = { [CTL]: controller('""') };
+    const constant = { [CONSTS]: ROOT_CONSTS, [CTL]: controller('ApiPaths.ROOT') };
+
+    expect(groupProviders(constant)).toEqual(groupProviders(literal));
+    expect(ingestionRoutes(constant)).toEqual(ingestionRoutes(literal));
+  });
+
+  it('emits the route on both sides rather than dropping it', () => {
+    // Deliberately NOT asserting the two sides produce the same string here.
+    // They do not, and they did not before this change either: measured on the
+    // LITERAL spelling, which no part of this change touches, the group side
+    // emits `/api/v1/` (`joinPath` appends `/` before an empty method path)
+    // while ingestion emits `/api/v1`. That trailing-slash divergence is a
+    // separate pre-existing defect; the fix here only makes the constant
+    // spelling reach it too, instead of silently dropping the route. The test
+    // above is what pins the parity that this change is responsible for —
+    // constant behaves as literal, on each side.
+    const files = { [CONSTS]: ROOT_CONSTS, [CTL]: controller('ApiPaths.ROOT') };
+    expect(groupProviders(files)).toEqual(['GET /api/v1/']);
+    expect(ingestionRoutes(files)).toEqual(['GET /api/v1']);
+  });
+});
--- a/gitnexus/test/unit/java-route-const-resolver.test.ts
+++ b/gitnexus/test/unit/java-route-const-resolver.test.ts
@@ -792,3 +792,66 @@ describe('text blocks keep the skip floor', () => {
     expect(extractJavaModuleConstants(parse(src)).literals.has('X')).toBe(false);
   });
 });
+
+describe('an empty fold is a success, not a skip', () => {
+  // `null` from this fold is the SKIP FLOOR: it means "a piece was
+  // unresolvable", and every caller acts on it by dropping the route. An
+  // empty-valued constant is not that case — it resolved, to the empty string.
+  // Spring reads `@GetMapping(ROOT)` with `ROOT = ""` exactly as it reads
+  // `@GetMapping("")`, and the group extractor's literal branch already keeps
+  // the latter. Collapsing the two silently lost the route for the constant
+  // spelling alone.
+  const EMPTY = 'src/main/java/com/example/Paths.java';
+
+  it('folds a constant whose value is the empty string', () => {
+    const repo = repoOf({
+      [EMPTY]: `package com.example;
+public class Paths { public static final String ROOT = ""; }`,
+    });
+    expect(foldJavaOperands(EMPTY, [{ kind: 'ref', name: 'ROOT' }], repo)).toBe('');
+  });
+
+  it('concatenating empty constants folds to the empty string', () => {
+    const repo = repoOf({
+      [EMPTY]: `package com.example;
+public class Paths {
+  public static final String A = "";
+  public static final String B = "";
+}`,
+    });
+    expect(
+      foldJavaOperands(
+        EMPTY,
+        [
+          { kind: 'ref', name: 'A' },
+          { kind: 'ref', name: 'B' },
+        ],
+        repo,
+      ),
+    ).toBe('');
+  });
+
+  it('still floors to skip when a piece is genuinely unresolvable', () => {
+    // The control: the two outcomes have to stay distinguishable, or the fix
+    // would trade one conflation for another.
+    const repo = repoOf({
+      [EMPTY]: `package com.example;
+public class Paths { public static final String ROOT = compute(); }`,
+    });
+    expect(foldJavaOperands(EMPTY, [{ kind: 'ref', name: 'ROOT' }], repo)).toBeNull();
+  });
+
+  it('agrees with resolveJavaConstant, which never collapsed empty to null', () => {
+    // The binding used to disagree with itself: the single-constant resolver
+    // returns `""` (pinned by the 30-level DAG case above, whose every
+    // intermediate value is empty), and `resolveOperands` in the shared
+    // language-agnostic core returns `foldExpr` unfiltered. Only this fold
+    // collapsed.
+    const repo = repoOf({
+      [EMPTY]: `package com.example;
+public class Paths { public static final String ROOT = ""; }`,
+    });
+    expect(resolveJavaConstant(EMPTY, 'ROOT', repo)).toBe('');
+    expect(foldJavaOperands(EMPTY, [{ kind: 'ref', name: 'ROOT' }], repo)).toBe('');
+  });
+});


gitnexus/src/core/ingestion/route-extractors/java-const-resolver.ts
```diff
@@ -619,12 +619,20 @@ function foldOperands(
 /**
  * Fold an inline operand list (e.g. `API_CIS_V1 + "summary/save"`) against
  * `fileKey`, or null when any piece is unresolvable (skip floor).
+ *
+ * An empty result is a SUCCESS, not a skip. `static final String ROOT = "";`
+ * folds to `""`, which `joinPath` then resolves against the type-level prefix
```
magyargergo, line 624: **P2 — `joinPath` is the group caller's joiner, not ingestion's.** `[code-read]`

`foldJavaOperands` has two callers and only one of them reaches `joinPath`: the group extractor, at `group/extractors/http-patterns/java.ts:1024`. The ingestion caller — the side this module lives under, and the side that drives the graph — is wired at `languages/java.ts:245`, consumed at `parse-impl.ts:1313`, and joins via `normalizeExtractedRoutePath` (`route-extractors/route-path.ts:17-22`) at `routes.ts:321` and `call-processor.ts:410`.

The two joiners disagree on exactly the input this comment exists to justify: `joinPath('/api/v1', '')` → `/api/v1/`, while `normalizeExtractedRoutePath('', '/api/v1')` → `/api/v1`. This PR's own test pins that disagreement at `test/unit/group/java-const-route-parity.test.ts:253-254`, so the comment currently contradicts the diff it ships with.

The claim it is making — *an empty fold is treated exactly like the literal `@GetMapping("")`* — is true on **both** sides; only the named mechanism is wrong. Caller-neutral wording would also avoid fresh drift from the Kotlin sibling, which already carries the identical `joinPath` phrasing on `main` (`kotlin-const-resolver.ts:1365`):

```
 * An empty result is a SUCCESS, not a skip. `static final String ROOT = "";`
 * folds to `""`, which each caller's joiner then resolves against the
 * type-level prefix exactly as it resolves the literal `@GetMapping("")` —
 * `joinPath` on the group side, `normalizeExtractedRoutePath` on the
 * ingestion side.
```

gitnexus/src/core/ingestion/route-extractors/java-const-resolver.ts
```diff
@@ -619,12 +619,20 @@ function foldOperands(
...
+ * `""` for an empty constant.
  */
 export function foldJavaOperands(
   fileKey: string,
   operands: readonly Operand[],
   repo: RepoConstants,
 ): string | null {
-  const out = foldOperands(fileKey, operands, newFoldState(repo), 0);
-  return out === '' ? null : out;
+  return foldOperands(fileKey, operands, newFoldState(repo), 0);
```
magyargergo, line 637: **P2 — the premise of this change holds only when a type-level prefix that exists is actually *known*. In two ordinary Spring shapes it isn't, and the empty fold then publishes the application root.** `[reproduced]`

Measured BASE vs HEAD through the real `extractSpringRoutes` + real fold + the production `normalizeExtractedRoutePath`, with `@GetMapping(ApiPaths.ROOT)` where `ROOT = ""`:

| class-level annotation | BASE | HEAD |
|---|---|---|
| `@RequestMapping({ApiPaths.V1, ApiPaths.V2})` | *(nothing)* | **`GET /`** — app serves `/api/v1`, `/api/v2` |
| `@RequestMapping({"/a","/b"})` | *(nothing)* | **`GET /`** — app serves `/a`, `/b` |
| `@RequestMapping(ApiPaths.V1)` | *(nothing)* | *(nothing)* — suppression works |
| `@RequestMapping("/api/v1")` | *(nothing)* | `GET /api/v1` — the intended fix |

Why the guards miss the array form: the class-level `@value` branch admits only string literals inside an array (`spring.ts:74-75`) and the `@value_expr` branches require a **direct** child of `annotation_argument_list` (`spring.ts:101-108`), so an array of `field_access` elements matches neither and the class lands in *none* of `prefixByClassId` / `classesWithArrayPrefix` / `classesWithUnfoldablePrefix`. `spring.ts:286` then falls through to `classPrefix = … ?? ''`. Neither method guard fires: `isArrayElement` (`spring.ts:271`) reads `valueNode?.parent`, which is `undefined` for a constant (it arrives as `@value_expr`), and `classesWithUnfoldablePrefix` (`spring.ts:282-284`) was never populated.

**This is P2, not a blocker.** The prefix-dropping hole is pre-existing and already emits a false `GET /` on `main` via the *literal* `@GetMapping("")` in these same shapes; `spring.ts:269-270` even declares the scalar-under-array divergence out of scope in its own comment. This change adds one more spelling as an entrance — it does not create the defect class, nor any downstream consequence that is not already reachable.

Two ways to close it, either fine:

1. Treat an array-form class prefix as **prefix-unknown for all method routes**, not only array-form ones — drop the `isArrayElement &&` conjunct at `spring.ts:272`, and add array initializers with non-`string_literal` elements to `classesWithUnfoldablePrefix`. Fixes the pre-existing literal cell too.
2. Narrow to where the premise holds: keep the unfiltered fold and skip at the consumer when no prefix is known — `parse-impl.ts:1316`, `value === '' && !dr.prefix`.

If neither is in scope, a line in the PR body plus a follow-up issue would do. The concern is that "nothing else in the resolver conflates the two" will read to the next person as having settled the question for the layers *above* the resolver, where `spring.ts` was quietly relying on the collapse. Worth pinning whichever answer you pick with a fixture, since this is the one reproduced behavior change the new tests don't cover.

gitnexus/test/unit/group/java-const-route-parity.test.ts
```diff
@@ -211,3 +211,46 @@ public class OrderController {
...
+    // LITERAL spelling, which no part of this change touches, the group side
+    // emits `/api/v1/` (`joinPath` appends `/` before an empty method path)
+    // while ingestion emits `/api/v1`. That trailing-slash divergence is a
+    // separate pre-existing defect; the fix here only makes the constant
+    // spelling reach it too, instead of silently dropping the route. The test
+    // above is what pins the parity that this change is responsible for —
+    // constant behaves as literal, on each side.
+    const files = { [CONSTS]: ROOT_CONSTS, [CTL]: controller('ApiPaths.ROOT') };
+    expect(groupProviders(files)).toEqual(['GET /api/v1/']);
+    expect(ingestionRoutes(files)).toEqual(['GET /api/v1']);
```
magyargergo, line 254: **P2 — this assertion is measuring a hand-rolled join, not the production normalizer.** `[reproduced]`

`ingestionRoutes()` builds the URL at line 72 as `` `${route.prefix ?? ''}${path}`.replace(/\/{2,}/g, '/') ``, while production ingestion uses `normalizeExtractedRoutePath` (`src/core/ingestion/route-extractors/route-path.ts:17-22`) at `routes.ts:321` and `call-processor.ts:410`, which additionally trims, strips trailing slashes, inserts the prefix/path separator, and degrades empty to `'/'`. Measured over 10 inputs, 7 of 10 diverge.

For this fixture the two happen to agree (`''` + `/api/v1` → `/api/v1` either way), so the assertion is correct today — but by coincidence of the fixture, not by construction. For an empty path with **no** prefix the helper yields `GET ` where production yields `GET /`, so the obvious next test in this `describe` cannot be written correctly against the current helper: it would have to pin a string production never emits. The helper predates this PR (#2980), but this is the first change that pushes empty method paths through it, and both new assertions (`:239` and here) now rest on it — in a file whose name and docblock advertise group↔ingestion parity.

**The fix is free — verified.** With `ingestionRoutes()` calling the production normalizer:

```ts
import { normalizeExtractedRoutePath } from '../../../src/core/ingestion/route-extractors/route-path.js';
// …
out.push(`${route.httpMethod} ${normalizeExtractedRoutePath(path, route.prefix ?? null)}`);
```

all 9 parity tests still pass on HEAD, and on BASE the same 2 new tests still fail — identical regression-detection power, and it unblocks the missing no-class-prefix case (`GET /` on both sides).

magyargergo (commented):
## Verdict

**This is a good fix and it should land — but not quite as-is.** The argument is right: `foldJavaOperands` was the only implementation of its contract that collapsed a *successful* empty fold into the skip floor, and removing that collapse recovers a real Spring endpoint (`@GetMapping(ROOT)` with `ROOT = ""` under a class-level `@RequestMapping`) that the literal `@GetMapping("")` spelling has always produced. All four of the PR's load-bearing factual claims check out, the tests genuinely fail on BASE, and the regression this subsystem is historically prone to — a constant class prefix turning a safe skip into a wrong route — is verifiably suppressed. The one thing to fix before merge is that the change's own stated premise ("folds to `""`, which the caller then resolves against the type-level prefix") **holds only when a type-level prefix that exists is actually known**, and there are two ordinary Spring shapes where the extractor silently pretends there is no prefix. In those, an empty constant fold now publishes `GET /` — the application root — where BASE safely emitted nothing. That is a **P2**, not a merge-stopper (the same hole already reaches `/` on `main` through the literal spelling), but it should be either closed or written down rather than shipped silently.

## Method and honesty

Reviewed across 12 Claude lanes (6 GitNexus swarm + 6 ce personas). **Codex/GPT-5.5 was unavailable (OpenAI spend limit), so this is a single-engine review — no cross-engine corroboration is claimed.** Where lanes agree, that is *same-engine convergence across personas*, which can be a shared false positive; it is not independent confirmation, and nothing below is labeled "strong" on that basis.

What carries the weight instead is **execution**. Several lanes built real BASE and HEAD source trees (`git archive` of the merge-base, real `node_modules`) and ran the same fixtures through both, driving the real `extractSpringRoutes`, the real fold, the real `normalizeExtractedRoutePath`, and in one case the real `routesPhase.execute`. The central finding was independently re-derived twice more — once by the review coordinator and once at synthesis — with matching results. Findings are tagged `[reproduced]` or `[code-read]` accordingly, and every `file:line` below was re-confirmed in a worktree at `e8f56e6dc` (`git status` clean, 75/75 tests green on HEAD).

## What's solid

Credit where it is earned — this PR survived a lot of attacks, and several of them were the ones this file has actually failed before:

- **The #2980 P1 shape does not recur.** A **scalar** constant class prefix (`@RequestMapping(ApiPaths.BASE)`) plus an empty-valued method constant emits nothing on BASE *and* nothing on HEAD, on both the ingestion and group sides. Suppression fires at extraction time — `spring.ts:282-284` and the no-arg loop at `spring.ts:333`, mirrored by `group/extractors/http-patterns/java.ts:801,986` — before `foldJavaOperands` is ever called. Four lanes probed this shape (including static-import and `value = …` named-member spellings); all four refuted it, and I reproduced the null result again at synthesis. `[reproduced]`
- **No `SCHEMA_BUMP` is needed, and adding one would be wrong.** The worker's capture set is unchanged; the fold runs main-thread at `parse-impl.ts:1294-1327`, *after* cached chunk results are replayed. Warm caches at v71+ replay the operands and the new fold folds them correctly. This is the exact inverse of #2980's cache-inert P1, where the fix lived inside the worker. `[code-read]`, two lanes
- **The degenerate empty operand list is unreachable.** `parseJavaConstOperands` is the sole producer of `routePathOperands`/`pathOperands` and every non-null return has length ≥ 1 — enumerated at runtime across 14 expression shapes by two lanes independently.
- **No existing route is lost, and no route node disappears.** `addRoute` de-dupes by `(method, url)` (`routes.ts:296-303`), so a URL and verb always survive; and `claim()` returns before `claimed.add`, so an empty path can never *steal* a handler identity. The removed line could only ever convert `''` to `null`, so the change is strictly additive to the route set.
- **No security regression.** Injection was chased to the sink and refuted (the URL reaches `graph.addNode({ properties: { name: url } })` — a structured property API, not query interpolation) and so was DoS-guard bypass (`MAX_FOLD_LENGTH`, depth-64 and depth-32 caps all live *inside* `foldOperands`, which returns `null` before the removed post-filter could ever be reached). Hidden-Unicode/bidi scan clean; the only non-ASCII in the diff is three em-dashes in comments. No workflow, lockfile, or `postinstall` touched.
- **R4 contract joining is unaffected.** The trailing-slash divergence is real at the plugin-detection layer but inert for linking: `normalizeHttpPath` (`group/extractors/http-route-extractor.ts:283-292`) strips trailing slashes before contract-id generation, and `normalizeContractId` (`group/matching.ts:62-68`) strips them again at match time, so both sides converge on `http::GET::/api/v1`. Scoping it out was the right call. `[reproduced]`
- **The four claims in the PR's own justification are all TRUE**, verified independently by two lanes: `resolveOperands` returns `foldExpr` unfiltered (`constant-resolver.ts:223`); `resolveJavaConstant` returns `""` for an empty constant; the literal branch skips only on `unquoteLiteral(...) === null` (`http-patterns/java.ts:608-609`); and the old doc comment did reserve `null` for "unresolvable". The removed filter really was an anomaly — `git blame` shows it arrived with the commit that created the file (#2980, `3f5fbb05e`) and was never mentioned in any #2980 review comment or in that PR's own empty-path fix.
- **The tests are real, and CI runs them.** 5 of the 6 new tests fail on BASE and pass on HEAD in an isolated same-environment A/B; the 6th is a deliberate control that correctly passes on both. Shard enumeration puts both files in `coverage 2/3`, which passed on the PR head. No test was skipped, renamed, deleted, or weakened; assertions are exact (`toBe`/`toEqual`/`toBeNull`); zero deletions in either test file; and **no pre-existing test anywhere encoded the old `'' → null` behavior**, which is the cleanest evidence available that nothing depended on it.
- **The branch is clean.** Zero file overlap and no semantic conflict with the three commits now on `main`; `merge-tree` produces no conflict; one linear commit, no merges, no debug code, no stray formatting. The decision to split Kotlin into #3059 was correct and for a better reason than the author gave: `foldKotlinOperands` is consumed only by the group extractor, while `foldJavaOperands` also feeds the ingestion hook — same source change, different blast radius.
- **Standards, docs, and DoD are in good shape.** Prettier/ESLint clean (reproduced locally: 0 errors, 9 pre-existing warnings above the new block); PR template fully filled with real out-of-scope reasoning; no `CHANGELOG.md`, `AGENTS.md`, or `ARCHITECTURE.md` obligation triggered; the `language-provider.ts` and group-side skip-floor doc comments all remain accurate.

## Findings

### F1 · P2 — an array-form class `@RequestMapping` silently drops its prefix, so an empty constant fold now publishes `GET /`

*Mixed: the defect class is **pre-existing**; this PR adds one new spelling as an entrance to it.*

**Trigger.** Any controller whose class-level `@RequestMapping` uses the **array form**, with a method path that is a constant resolving to `""`:

```java
public class ApiPaths { public static final String ROOT = ""; String V1 = "/api/v1"; String V2 = "/api/v2"; }

@RestController
@RequestMapping({ApiPaths.V1, ApiPaths.V2})   // or {"/a","/b"} — literal elements behave the same
public class BooksController {
  @GetMapping(ApiPaths.ROOT)
  public List<Book> list() { … }
}
```

Measured BASE vs HEAD through the real extractor, real fold, and the production normalizer:

| class-level annotation | method | BASE | HEAD | |
|---|---|---|---|---|
| `@RequestMapping({ApiPaths.V1, ApiPaths.V2})` | `@GetMapping(ApiPaths.ROOT)` | *(nothing)* | **`GET /`** | wrong — app serves `/api/v1`, `/api/v2` |
| `@RequestMapping({"/a","/b"})` | `@GetMapping(ApiPaths.ROOT)` | *(nothing)* | **`GET /`** | wrong — app serves `/a`, `/b` |
| `@RequestMapping(ApiPaths.V1)` | `@GetMapping(ApiPaths.ROOT)` | *(nothing)* | *(nothing)* | suppression works |
| `@RequestMapping("/api/v1")` | `@GetMapping(ApiPaths.ROOT)` | *(nothing)* | `GET /api/v1` | **the intended fix** |
| *(no class mapping)* | `@GetMapping(ApiPaths.ROOT)` | *(nothing)* | `GET /` | correct, parity with literal |

**Evidence.** `gitnexus/src/core/ingestion/route-extractors/java-const-resolver.ts:637` (the changed return), with the prefix hole at `route-extractors/spring.ts`: the class-level `@value` branch admits only string literals inside an array (`spring.ts:74-75`) and the `@value_expr` branches require a **direct** child of `annotation_argument_list` (`spring.ts:101-108`), so an array of `field_access` elements matches neither. The class therefore lands in *none* of `prefixByClassId` / `classesWithArrayPrefix` / `classesWithUnfoldablePrefix`, and `spring.ts:286` falls through to `classPrefix = … ?? ''`. Neither method guard fires: `isArrayElement` (`spring.ts:271`) reads `valueNode?.parent`, which is `undefined` for a constant (it arrives as `@value_expr`), and `classesWithUnfoldablePrefix` (`spring.ts:282-284`) was never populated. The prefix is dropped, `normalizeExtractedRoutePath('', null)` yields `/`, and the decorator loop at `routes.ts:320-321` has no truthiness guard (unlike `routes.ts:305`). `[reproduced]` — reproduced independently three times (ce-correctness, ce-adversarial, coordinator) and re-run at synthesis.

**Why P2 and not P1.** This is the last unfilled cell of a hole that is already three-quarters open on `main`. The same array-form class shape **already emits a false `GET /` today** through the *literal* `@GetMapping("")` spelling, and already emits wrong unprefixed routes for non-empty constant and literal method paths. `spring.ts:269-270` even declares that divergence out of scope in its own comment. This PR does not create the defect class and does not create any downstream consequence that is not already reachable — it adds one more spelling. The ce-adversarial lane graded this P1 on the strength of a real, reproduced displacement cascade (the false `GET /` winning `addRoute`'s first-writer-wins at `routes.ts:296-303` and taking over a legitimate root route's `filePath` attribution, with `handlerPathFor` at `routes.ts:339-346` then reading the wrong file). **That cascade is real and its mechanism is confirmed — but it is reachable on `main` today via the literal spelling, so it is not a new consequence of this change.** ce-correctness graded it P2 with that reasoning and the coordinator's independent re-derivation agrees.

**Note on two lanes disagreeing.** `gitnexus-risk-architect` and `gitnexus-test-ci-verifier` reported this risk REFUTED; `ce-correctness` and `ce-adversarial` reported it CONFIRMED. **Both are correct, about different annotation shapes.** The first pair tested the *scalar* constant prefix — the exact #2980 P1 shape — where suppression genuinely works (row 3 above). The second pair tested the *array* form, where it does not. Neither lane was wrong; the shape space just wasn't covered by one probe.

**Suggested fix** — either is fine, and the choice is the author's:

1. *Close the hole the fix depends on.* Treat an array-form class prefix as **prefix-unknown for all method routes**, not only array-form ones — i.e. drop the `isArrayElement &&` conjunct at `spring.ts:272`, and add array initializers containing non-`string_literal` elements to `classesWithUnfoldablePrefix`. This also fixes the pre-existing literal-`""` cell for free.
2. *Narrow the change to where the premise holds.* Keep the unfiltered fold, but have the consumer refuse a resolved-empty path when no prefix is known (`parse-impl.ts:1316`: skip when `value === '' && !dr.prefix`). Preserves the motivating case and keeps `/` unreachable from a constant fold.

If neither is in scope, say so in the PR body and file the follow-up — but please don't merge with it undocumented, because the new doc comment's "nothing else conflates the two" will read to the next person as having settled the question.

### F2 · P2 — the group side's unfoldable-prefix guard is dead for interface methods, so an empty constant publishes a false `consumer GET /`

*Mostly pre-existing; this PR adds one new row. **Unchanged file — body-only, cannot be anchored inline.***

**Trigger.** A `@FeignClient` **interface** carrying a constant class prefix (`@RequestMapping(ApiPaths.BASE)` where `BASE` is unfoldable) with a method whose path folds to `""`.

**Evidence.** `gitnexus/src/core/group/extractors/http-patterns/java.ts:985-986` resolves the annotation's owner with `findEnclosingClass(route.methodNode)`, and `findEnclosingClass` (`ingestion/route-extractors/spring-shared.ts:139-146`) matches `class_declaration` only — so it returns `null` for a method on an `interface_declaration` and the guard never fires. Meanwhile `typesWithUnfoldablePrefix.add(node.id)` at `java.ts:679` **is** populated for interfaces (that branch is not gated on node type). The Feign emit path is 11 lines below at `java.ts:997`. Reproduced by two lanes: group side goes from `["consumer GET /a","consumer GET /lit"]` on BASE to `["consumer GET /", …]` on HEAD; the `/a` and `/lit` rows are already wrong on BASE and are not charged to this PR. `[reproduced]` (two lanes) + `[code-read]` (coordinator, mechanism)

**Suggested fix.** One line, and the codebase already contains the exact idiom twice — `java.ts:732` and `java.ts:1072` both write `findEnclosingInterface(route.methodNode) ?? findEnclosingClass(route.methodNode)`. Apply the same at `java.ts:985`. That also fixes the pre-existing `/a` and `/lit` rows.

### F3 · P2 — the new doc comment names the wrong joiner for the caller that matters

*Introduced here — it is the PR's primary non-test artifact.*

**Trigger.** N/A (documentation accuracy), but it is load-bearing: this comment is the thing that will stop a future reader from restoring the filter, and it is also the premise F1 exploits.

**Evidence.** `java-const-resolver.ts:624` says the empty fold is "which `joinPath` then resolves against the type-level prefix". `joinPath` (`route-extractors/spring-shared.ts:203-208`) is reached by only **one** of this function's two callers — the group extractor, at `group/extractors/http-patterns/java.ts:1024`. The ingestion caller, which is the side this module lives under and the side that drives the graph, is wired at `languages/java.ts:245`, consumed at `parse-impl.ts:1313`, and joins via `normalizeExtractedRoutePath` (`route-extractors/route-path.ts:17-22`) at `routes.ts:321` and `call-processor.ts:410`. The two joiners **disagree on exactly this input**: `joinPath('/api/v1','')` → `/api/v1/`, `normalizeExtractedRoutePath('','/api/v1')` → `/api/v1`. This PR's own test pins that disagreement at `test/unit/group/java-const-route-parity.test.ts:253-254`, so the comment contradicts the diff it ships with. `[reproduced]`, three lanes

**Suggested fix.** Caller-neutral wording, so the Java and Kotlin siblings don't drift further apart (the identical `joinPath` phrasing is already on `main` in `kotlin-const-resolver.ts:1365`): "…folds to `""`, which each caller's joiner then resolves against the type-level prefix exactly as it resolves the literal `@GetMapping("")` — `joinPath` on the group side, `normalizeExtractedRoutePath` on the ingestion side."

### F4 · P2 — the parity test's ingestion side is a hand-rolled reimplementation, and it is wrong precisely for empty method paths

*Helper is pre-existing (from #2980); this PR is what made it load-bearing.*

**Trigger.** Any fixture in that file with an empty method path and no class prefix — i.e. the obvious next test to add.

**Evidence.** `test/unit/group/java-const-route-parity.test.ts:72` builds the ingestion URL as `` `${route.prefix ?? ''}${path}`.replace(/\/{2,}/g,'/') ``, while production uses `normalizeExtractedRoutePath`, which additionally trims, strips trailing slashes, inserts the prefix/path separator, and degrades empty to `'/'`. Measured over 10 inputs, **7 of 10 diverge**. For the case this PR pins they happen to agree (`''` + `/api/v1` → `/api/v1` both), so today's verdicts are correct — but for an empty path with no prefix the helper yields `GET ` where production yields `GET /`. It diverges in both directions (false alarms *and* masking), so its agreement is a property of the fixture, not of the helper. Both new assertions (`:239`, `:254`) rest on it, in a file whose name and docblock advertise group↔ingestion parity. `[reproduced]`, three lanes

**Suggested fix, verified free.** Have `ingestionRoutes()` call the production normalizer: `out.push(\`${route.httpMethod} ${normalizeExtractedRoutePath(path, route.prefix ?? null)}\`)`. Two lanes substituted it and measured: **all 9 parity tests still pass on HEAD, and on BASE the same 2 new tests still fail** — identical regression-detection power, sound oracle, and it unblocks the missing no-prefix test.

### F5 · P3 — the recovered route is half-linked: its handler association is skipped

*Pre-existing, **not** introduced here. Parity with the literal spelling genuinely holds.*

**Trigger.** Any empty-path decorator route — constant, literal `@GetMapping("")`, or no-arg `@GetMapping`.

**Evidence.** `call-processor.ts:409` opens `if (!routePath) return;` and is reached for decorator routes at `:477`. An empty-string `routePath` is falsy, so no `HANDLES_ROUTE` edge is written — while `routes.ts:320-321` has no truthiness guard and creates the Route node anyway. Reproduced through the real routes phase: the recovered route arrives as `{httpMethod:"GET", routePath:"", prefix:"/api/v1"}` → Route node `GET /api/v1` created, handler claim skipped. The literal and no-arg spellings produce the byte-identical shape and are skipped identically, which is why this is not a regression and why the PR's parity claim stands. `[reproduced]`, five lanes

Three lanes graded this P2 on the strength of the contract asymmetry (two adjacent loops in one feature reading the same value two ways). I keep it at **P3** because it is pre-existing and parity-preserving, and the actionable part here is one sentence: the PR body's "recovers a real endpoint" is accurate about the Route node but the endpoint arrives without its handler link. Worth disclosing so a reviewer who checks `route_map` and finds no handler doesn't read it as a regression from this change. The `claim()` fix itself belongs in its own PR — see the follow-ups.

### F6 · P3 — the e2e skip-floor assertion now contradicts the new contract

*Introduced here in effect, though the assertion itself is untouched. **Unchanged file — body-only.***

`test/unit/java-route-const-pipeline-e2e.test.ts:207-209` (the assertion at `:209`) enforces `expect(p.length).toBeGreaterThan(1)` on **method-level folded paths**, under the comment "Skip floor: no phantom empty/raw-expr paths", and the file's docblock at `:32` promises "a phantom `POST ` / empty path never appears". After this PR, "no folded path is empty" is no longer a system invariant — an empty fold is a legitimate success. The assertion survives only because this fixture contains no empty-valued constant, which means **anyone who extends the e2e with the very case this PR adds gets a failure on correct behavior.** It is also bounds-only, which `DoD.md:82` discourages. `[code-read]`, two lanes

**Suggested fix.** Rewrite as exact per-route expectations, or scope the comment and assertion to "no raw-expression paths" and let empty be legal — and add the empty-valued constant to the fixture while you're there, since this is the only test in the repo that drives the real worker pipeline for this hook.

### F7 · P3 — `foldJavaOperands([])` changed from `null` to `''`, unreachable today

`java-const-resolver.ts:637`. Verified by direct call on both trees. Unreachable in production (see What's solid), so advisory only — but the removed filter was incidentally also the guard against a degenerate list, and the consumer treats `[]` as present (`parse-impl.ts:1311`: `dr.routePathOperands ? … : null`, and `[]` is truthy). If you want the intent the new comment states pinned mechanically, `if (operands.length === 0) return null;` costs nothing: "no operands" genuinely is not "resolved to empty". `[reproduced]`, three lanes

## Two factual corrections worth making before merge

Both independently verified against `origin/main` at synthesis.

**(a) PR #3059 is already merged, and the Kotlin claim is backwards.** The PR body says #3059 "adds the Kotlin binding. That binding mirrors this function line for line and will be updated to match." #3059 merged on 2026-08-28 as **`880f94d14`** — the first of the three commits that make this PR `BEHIND` — and it landed with `foldKotlinOperands` **already unfiltered** (`kotlin-const-resolver.ts:1399` on `main`: `return foldOperands(fileKey, scoped, newFoldState(repo, index), 0);`), while `foldJavaOperands` on `main` still reads `return out === '' ? null : out;` (`java-const-resolver.ts:629`). So **Java is currently the outlier on `main`, and this PR is what converges it** — that is an argument *for* merging, and it is the opposite of what the body says. Sharper still: Kotlin's own doc comment on `main` (`kotlin-const-resolver.ts:1372-1373`) asserts *"Matches `foldJavaOperands`, so the two JVM bindings do not diverge on the same input"* — a documented cross-binding invariant that is **false on `main` today** and becomes true only when this lands. Since the repo squash-merges and the PR body becomes permanent history feeding the changelog, the stale paragraph is worth amending. (One lane's refuted-attacks table states there is no `foldKotlinOperands` at all — true of this PR's *base*, which predates #3059, but not of `main`.)

**(b) The doc comment's `joinPath` is the group caller only.** See F3. The new comment tells a reader of an *ingestion* module that `joinPath` resolves the empty path; ingestion uses `normalizeExtractedRoutePath`, and the two produce different strings for exactly this input — a difference this PR's own test pins at `java-const-route-parity.test.ts:253-254`.

## Test and CI notes

- **BASE/HEAD A/B, reproduced.** Isolated same-environment harness (real `src` + `test` copy, `node_modules` symlinked, swapping only `java-const-resolver.ts`): HEAD → `2 files passed, 75 tests passed`; BASE → `2 files failed, 5 failed | 70 passed`. The 5 failures are the right ones, with the right BASE messages (`expected null to be ''` ×3; `expected [] to deeply equal [ 'GET /api/v1/' ]` ×2). The 6th new test — *"still floors to skip when a piece is genuinely unresolvable"* — passes on both **by design**, which is the correct shape for a control, and it is not a tautology (traced: `parseJavaConstOperands:243` returns `null` for a `method_invocation` → `resolveWithState:587` → `foldOperands:611`). All 70 pre-existing tests in those files pass on both trees. A broader same-environment A/B over the group suite plus the e2e shows a delta of exactly those 5 tests.
- **CI provably executes both files.** `ci-tests.yml:60-70` runs the full suite with no file filter; shard enumeration puts both new files in `coverage 2/3`, and that shard passed in 9m15s on the PR head.
- **The parity-helper oracle (F4) has a verified-free fix** — swapping in `normalizeExtractedRoutePath` keeps all 9 tests green on HEAD and still fails the 2 new ones on BASE. Do this first; the next test needs it.
- **Highest-value missing tests**, in order: (1) the **array-form class prefix × empty-valued method constant** fixture from F1 — whatever the team decides the right answer is, pin it, because that is the one reproduced behavior change the PR does not cover; (2) the **unfoldable-prefix × empty-valued-constant** suppression case, added to the existing suppression `describe` — correctness now *depends* on a guard that no test pins for this input, and it is the exact shape #2980 reproduced as a P1; (3) the **no-class-prefix** case asserting `GET /` on both sides (blocked on F4's helper fix, which is why F4 comes first); (4) route→handler wiring for an empty folded path — the repo already has the template at `test/unit/nest-decorator-routes.test.ts:58`, where `nest.ts:527-533` deliberately emits `'/'` rather than `''` *for this exact reason*, with the rationale written out in a comment. That precedent is worth reading before deciding on F5.
- **`java-route-const-pipeline-e2e.test.ts:207-209` now contradicts the new contract** — see F6. Note the file self-skips on a stale `dist/` and lists `java-const-resolver.ts` in its `GATED_SOURCES`, so editing that file makes it skip locally unless you rebuild; it ran in CI, where `setup-gitnexus` builds.
- **Minor test-quality notes** (all P3, none blocking): the first new resolver test at `:806-812` is subsumed verbatim by the fourth at `:844-856`, so six advertised tests are five distinct ones; the first parity assertion (`:238-239`) compares two computed values and would pass vacuously if both went empty — the file's own "guard the premise" idiom at `:193-197` is the local fix; the control is single-operand only, where the sharper control is a **mixed** list (a resolvable-empty operand followed by an unresolvable one, which is the only shape where an empty accumulator could plausibly mask a failure); and neither file's curated "Asserted:"/"Value shapes covered" docblock was extended for the new case.
- **On the deliberately-pinned divergence** (`:242-255`): the lanes converged that pinning is the *right* choice — it documents and detects, where `not.toEqual` or omission would prove nothing, and ~25 sites in this repo pin out-of-scope behavior in prose comments with no documented TODO/issue requirement. The gap is packaging: no issue number, and the eye-catching test pins the defect while its quieter sibling carries the actual parity claim. A retitle plus an issue reference closes it.

## Lower-priority items and untracked follow-ups

None of these block, and none are charged to this PR — but all three are currently untracked, and the PR body's "happy to open it as its own issue" is the pattern by which follow-ups evaporate:

- **Trailing-slash divergence has no issue.** `joinPath('/api/v1','')` → `/api/v1/` vs `normalizeExtractedRoutePath('','/api/v1')` → `/api/v1`, affecting every empty-method-path route, literal or constant. A GitHub search for an existing issue found nothing. Inert for contract matching (both sides normalize), so it is a cosmetic/`explain`-output defect — but the test that pins it deserves a ticket to point at.
- **`claim()`'s falsy guard is over-broad.** `call-processor.ts:409` should arguably be `if (routePath === null) return;` — the parameter is typed `string | null`, and `normalizeExtractedRoutePath` already handles `''` correctly. That would fix handler attribution for *all* empty-path routes at once. It is deliberately larger than this PR: it also newly claims handlers for every pre-existing no-arg `@GetMapping` route (`spring.ts:346-354`), and it would put `call-processor.ts:409` out of step with its partner guard at `routes.ts:305`, which *does* drop empty. Own PR, own tests.
- **The same constant-vs-literal divergence persists one layer up, on interface-based controllers.** `spring.ts:448-451` (`annotationRoutePathsOrDefault`) collects string literals only, so on an interface-based controller `@GetMapping("")` is kept while `@GetMapping(ApiPaths.ROOT)` is dropped — base and head identically. Worse in kind and also pre-existing: `typeClassPrefixes` (`spring.ts:455-461`) returns `[]` for a constant prefix and `resolveInheritedSpringRoutes` reads `[]` as "no prefix" rather than "prefix unknown", emitting unprefixed inherited routes. The structural fix is a `classPrefixUnresolvable` flag on `SharedSpringType`. Worth an issue.
- **Two doc/consistency nits.** The `foldRoutePathOperands` hook contract (`language-provider.ts:439-450`) states only the *sufficient* condition for `null` and never says `''` is a legitimate resolved value — which is plausibly why the same one-line bug was written independently in two language bindings. One sentence there, plus a cross-provider test asserting `fold(…) === ''` for an empty constant, would guard implementer #3. And the eight-line rationale will now live in both JVM bindings with wording already drifting ("type-level" vs "class-level"); its language-agnostic home is `resolveOperands` in `constant-resolver.ts`.

## CI status

The only failing check is **Vercel — "Authorization required to deploy"**, which is a fork-deployment infrastructure limitation, not a code signal. Everything substantive is green: all three ubuntu coverage shards plus the coverage merge, `quality/typecheck|lint|format`, all three macOS platform-sensitive shards, tree-sitter ABI ×3, benchmarks, gitleaks, CodeQL. (The check list currently enumerates 46 entries — 42 pass, 3 correctly skipped for a fork PR, 1 fail; earlier lanes recorded 39/40 against a smaller check set at review time.) The author's report of 4 local failures is credible and superseded: the named files (`analyze-index-lock-concurrency`, `review-agent-workflow`, `hooks`) contain zero references to route or constant resolution, and `AGENTS.md` documents LadybugDB file-locking failures as a known container issue.

Two process notes: `mergeStateStatus` is **BEHIND** by three commits, and **every green check predates all three of them** — the CI run was created 2026-08-27T14:08Z, while #3020 merged +3h19m later and #3059 +23h later. Given zero file overlap and zero shared imports (verified per-path), actual regression probability is low, but the green wall attests to the state at BASE, not to the merged result. Update the branch and require the fresh run — which the `BEHIND` state forces anyway, and which would also be the first time CI exercises the Java and Kotlin bindings of this contract together.

---

*Automated multi-lane review digest (12 Claude lanes; Codex/GPT-5.5 unavailable, single-engine). Findings tagged `[reproduced]` were executed against real BASE and HEAD source trees; `[code-read]` findings were not. Please verify before acting on any of it.*

PR #3069, opened 2026-08-28
feat(wiki): add grok local CLI provider

jreakin:
## Summary

Add `gitnexus wiki --provider grok` so wiki generation can use the local Grok Build CLI login (`grok --prompt-file` + JSON `.text`) instead of an HTTP API key.

```bash
gitnexus wiki --provider grok
gitnexus wiki --provider grok --model grok-build
```

## Motivation / context

Wiki already supports local CLIs (`cursor`, `claude`, `codex`, `opencode`) that skip the API-key prompt. Grok Build is a local CLI with a different spawn contract (prompt file, not stdin/`-p`), so it needs a dedicated client rather than `runLocalCLI`.

## Areas touched

- [x] Docs / agent config only — README and `gitnexus-cli` skill copies (`.claude/`, npm, Claude plugin). `CHANGELOG.md` is **not** edited in this PR (release-owned, `DoD.md` §4.5 / §6).

## Scope & constraints

**In scope**

- Dedicated `grok-client.ts`: spawn `grok` with `--prompt-file`, `--output-format json`, `--max-turns 15`, tool denylist, `--sandbox strict`, empty temp `--cwd` (do not ingest repo `AGENTS.md`)
- Stdin is **ignored** (`stdio: ['ignore','pipe','pipe']`). Grok does not read stdin as the prompt; an unused pipe can EPIPE the wiki process
- Treat `grok` as a local provider (`isLocalProvider`); persist `provider: 'grok'` and optional `grokModel`
- Route `invokeLLM` → `callGrokLLM`
- Windows: `resolveWindowsCliCommand` so npm `.cmd` shims spawn via `cmd.exe /d /s /c grok` (same class as other local CLIs)
- Timeout: wait for child `close` (SIGKILL grace, then hard deadline) before `fs.rm` of the temp sandbox; skip `rm` if `close` never fires
- Parse: require non-empty JSON `text` and `stopReason=end_turn` (case-insensitive). Incomplete/missing/unknown reasons throw distinct errors
- Help / i18n (`en` + `zh-CN`), README wiki section, `gitnexus-cli` skill mirrors

- `gitnexus-web` LLM providers
- `--provider grokbuild` alias
- HTTP `--provider custom` + `api.x.ai` (already works)
- Empty `--tools` allowlist (denylist used; live-verified that an empty allowlist does not disable `run_terminal_cmd`)
- Live grok binary in CI (mocked spawn, same as sibling local CLIs this sprint)
- Index format / MCP tool / workflow / env-var changes (N/A for this PR)
- `CHANGELOG.md` Unreleased hunk (DoD: release process owns that file)

## Implementation notes

- Cursor-shaped dedicated client, not stdin `runLocalCLI`.
- Never `--yolo` / `--always-approve`. Non-zero exit and JSON `{type:"error"}` throw with stderr so `grok login` failures surface.
- GitNexus graph: `invokeLLM` is the shared wiki LLM router. This PR adds a `grok` branch; existing HTTP/claude/codex/opencode arms are unchanged.

## Testing & verification

- [x] Targeted unit: `cd gitnexus && npx vitest run test/unit/local-cli-subprocess.test.ts test/unit/wiki-flags.test.ts` (142 passed on the grok/wiki files, including Windows cmd.exe, timeout wait-for-close, stdin ignore, `stopReason=end_turn`)
- [x] Wiki e2e: `cd gitnexus && npx vitest run test/integration/cli-e2e.test.ts -t wiki` (`--provider grok` non-TTY does not prompt for API key)
- [ ] CI Gate + Windows platform-sensitive shards on this head SHA (was previously green only on `4376ba3`)

## Risk & rollout

- Local process spawn of `grok`; no index schema, MCP, or persistence format change. Existing `~/.gitnexus/config.json` is unchanged until the user passes `--provider grok`.
- Rollback: revert the PR; wiki HTTP and other local providers are untouched.
- No secrets, tokens, or machine-specific paths.

## Checklist

- [ ] If `AGENTS.md` / overlays changed: headers, scope block, and changelog updated per project conventions — N/A (`AGENTS.md` unchanged)

---

## Summary

*A broad wiki CLI and provider change spanning implementation, documentation, localization, and tests. Its critical posture comes from the execution paths and dependents reached by the wiki generation flow, rather than from high-risk individual files.*

**🔴 CRITICAL blast radius. A wiki CLI change that adds the Grok local provider path across the `gitnexus` implementation, documentation, and tests, with impact reaching direct and transitive dependents.**

The change is concentrated in the `Wiki` module, centered on `gitnexus/src/core/wiki/llm-client.ts`, which has 10 changed symbols, and extending through `gitnexus/src/core/wiki/grok-client.ts` and `gitnexus/src/core/wiki/local-cli-client.ts`. The CLI decision path is in `gitnexus/src/cli/wiki.ts`, including `isLocalProvider`, `localModelConfigKey`, `wikiCommandImpl`, and `provider`. The graph traces 12 affected flows and 13 dependents, so reviewers should start with provider selection and the `invokeLLM` method before following its callers.

The supporting surface includes `gitnexus/src/cli/index.ts`, `gitnexus/src/storage/repo-manager.ts`, the localization constants `en` and `zhCN`, and tests covering CLI execution, help output, local subprocess behavior, and wiki flags. Documentation changes appear across `README.md`, `gitnexus/README.md`, `.claude/skills/gitnexus-cli/SKILL.md`, `gitnexus-claude-plugin/skills/gitnexus-cli/SKILL.md`, and `gitnexus/skills/gitnexus-cli.md`. Individual files are rated LOW risk, with no HIGH or CRITICAL risk files, but the critical blast level warrants checking the end to end CLI and wiki generation paths.

_Added by GitNexus for PR #3069. Edit freely — this block is replaced on the next review, everything above it is left untouched._

### DIFF
--- a/gitnexus/CHANGELOG.md
+++ b/gitnexus/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to GitNexus will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- **Wiki** — local Grok Build CLI provider (`gitnexus wiki --provider grok`) uses the authenticated `grok` CLI instead of an API key
+
 ## [1.6.10] - 2026-08-27
 
 ### Added
--- a/gitnexus/CHANGELOG.md
+++ b/gitnexus/CHANGELOG.md
@@ -4,10 +4,6 @@ All notable changes to GitNexus will be documented in this file.
 
 ## [Unreleased]
 
-### Added
-
-- **Wiki** — local Grok Build CLI provider (`gitnexus wiki --provider grok`) uses the authenticated `grok` CLI instead of an API key
-
 ## [1.6.10] - 2026-08-27
 
 ### Added
--- a/gitnexus/package-lock.json
+++ b/gitnexus/package-lock.json
@@ -825,6 +825,9 @@
       "cpu": [
         "arm"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -840,6 +843,9 @@
       "integrity": "sha512-dqVSFynCox4C/J8kT16V7SIFAns0IjgLwkvYT7p8LQVmJ5OS5b6tI9IGflxTeuBS//zXeFIUbwt5dwxyZ17cnA==",
       "cpu": [
         "arm64"
+      ],
+      "libc": [
+        "glibc"
       ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
@@ -857,6 +863,9 @@
       "cpu": [
         "ppc64"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -872,6 +881,9 @@
       "integrity": "sha512-bsb4rI+NldGOsXuej2r8OdSS8+zXDVaCWxyWrcv6kneTOlgAHtZABRzBBCwdsPiD90J4myNJuHpg6kA20ImW/w==",
       "cpu": [
         "riscv64"
+      ],
+      "libc": [
+        "glibc"
       ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
@@ -889,6 +901,9 @@
       "cpu": [
         "s390x"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -904,6 +919,9 @@
       "integrity": "sha512-ITPEtgffGJ0S6G9dRyw/366tJQqFRcHWPHhC+Stpg3Z8AEMrDrTr2lhdz4f/Y/HMbRh//7Z5mBzEpVdi62Oc3w==",
       "cpu": [
         "x64"
+      ],
+      "libc": [
+        "glibc"
       ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
@@ -921,6 +939,9 @@
       "cpu": [
         "arm64"
       ],
+      "libc": [
+        "musl"
+      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -937,6 +958,9 @@
       "cpu": [
         "x64"
       ],
+      "libc": [
+        "musl"
+      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -952,6 +976,9 @@
       "integrity": "sha512-affVWCTLooy8TSxbDx2qkzuDeaWLNVBA+P//FNBirHsXpP2fuBhk5AuboYUnrDnzoXes8GFjpTx0SBFOCRg+FA==",
       "cpu": [
         "arm"
+      ],
+      "libc": [
+        "glibc"
       ],
       "license": "Apache-2.0",
       "optional": true,
@@ -975,6 +1002,9 @@
       "cpu": [
         "arm64"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -996,6 +1026,9 @@
       "integrity": "sha512-sMd8rDxmpLOwv/7N44klFjOD5DUO7FLdjiXDI0hoxYaf7Ar262dQIEkosE98bps+5HPLtp/EvNqeqQtOycP/IA==",
       "cpu": [
         "ppc64"
+      ],
+      "libc": [
+        "glibc"
       ],
       "license": "Apache-2.0",
       "optional": true,
@@ -1019,6 +1052,9 @@
       "cpu": [
         "riscv64"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -1040,6 +1076,9 @@
       "integrity": "sha512-KgAxQ0DxpNOq1rG2t5cgTgShJFGSuU7XO45cqC+1NVOuZnP6tlgZRuSYOfNupGkHID0o3cJOsw4DVeJpMovcGw==",
       "cpu": [
         "s390x"
+      ],
+      "libc": [
+        "glibc"
       ],
       "license": "Apache-2.0",
       "optional": true,
@@ -1063,6 +1102,9 @@
       "cpu": [
         "x64"
       ],
+      "libc": [
+        "glibc"
+      ],
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -1085,6 +1127,9 @@
       "cpu": [
         "arm64"
       ],
+      "libc": [
+        "musl"
+      ],
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -1106,6 +1151,9 @@
       "integrity": "sha512-6O1NPKcDVj9QEdg7Hx549EX8U0rp6yXQERqru6yRN7fGBn32UvIRJUlWnk+8xDCiG76hXVBbX82NZ/ZKr0euIg==",
       "cpu": [
         "x64"
+      ],
+      "libc": [
+        "musl"
       ],
       "license": "Apache-2.0",
       "optional": true,
--- a/gitnexus/package-lock.json
+++ b/gitnexus/package-lock.json
@@ -825,9 +825,6 @@
       "cpu": [
         "arm"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -843,9 +840,6 @@
       "integrity": "sha512-dqVSFynCox4C/J8kT16V7SIFAns0IjgLwkvYT7p8LQVmJ5OS5b6tI9IGflxTeuBS//zXeFIUbwt5dwxyZ17cnA==",
       "cpu": [
         "arm64"
-      ],
-      "libc": [
-        "glibc"
       ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
@@ -863,9 +857,6 @@
       "cpu": [
         "ppc64"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -881,9 +872,6 @@
       "integrity": "sha512-bsb4rI+NldGOsXuej2r8OdSS8+zXDVaCWxyWrcv6kneTOlgAHtZABRzBBCwdsPiD90J4myNJuHpg6kA20ImW/w==",
       "cpu": [
         "riscv64"
-      ],
-      "libc": [
-        "glibc"
       ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
@@ -901,9 +889,6 @@
       "cpu": [
         "s390x"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -919,9 +904,6 @@
       "integrity": "sha512-ITPEtgffGJ0S6G9dRyw/366tJQqFRcHWPHhC+Stpg3Z8AEMrDrTr2lhdz4f/Y/HMbRh//7Z5mBzEpVdi62Oc3w==",
       "cpu": [
         "x64"
-      ],
-      "libc": [
-        "glibc"
       ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
@@ -939,9 +921,6 @@
       "cpu": [
         "arm64"
       ],
-      "libc": [
-        "musl"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -958,9 +937,6 @@
       "cpu": [
         "x64"
       ],
-      "libc": [
-        "musl"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -976,9 +952,6 @@
       "integrity": "sha512-affVWCTLooy8TSxbDx2qkzuDeaWLNVBA+P//FNBirHsXpP2fuBhk5AuboYUnrDnzoXes8GFjpTx0SBFOCRg+FA==",
       "cpu": [
         "arm"
-      ],
-      "libc": [
-        "glibc"
       ],
       "license": "Apache-2.0",
       "optional": true,
@@ -1002,9 +975,6 @@
       "cpu": [
         "arm64"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -1026,9 +996,6 @@
       "integrity": "sha512-sMd8rDxmpLOwv/7N44klFjOD5DUO7FLdjiXDI0hoxYaf7Ar262dQIEkosE98bps+5HPLtp/EvNqeqQtOycP/IA==",
       "cpu": [
         "ppc64"
-      ],
-      "libc": [
-        "glibc"
       ],
       "license": "Apache-2.0",
       "optional": true,
@@ -1052,9 +1019,6 @@
       "cpu": [
         "riscv64"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -1076,9 +1040,6 @@
       "integrity": "sha512-KgAxQ0DxpNOq1rG2t5cgTgShJFGSuU7XO45cqC+1NVOuZnP6tlgZRuSYOfNupGkHID0o3cJOsw4DVeJpMovcGw==",
       "cpu": [
         "s390x"
-      ],
-      "libc": [
-        "glibc"
       ],
       "license": "Apache-2.0",
       "optional": true,
@@ -1102,9 +1063,6 @@
       "cpu": [
         "x64"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -1127,9 +1085,6 @@
       "cpu": [
         "arm64"
       ],
-      "libc": [
-        "musl"
-      ],
       "license": "Apache-2.0",
       "optional": true,
       "os": [
@@ -1151,9 +1106,6 @@
       "integrity": "sha512-6O1NPKcDVj9QEdg7Hx549EX8U0rp6yXQERqru6yRN7fGBn32UvIRJUlWnk+8xDCiG76hXVBbX82NZ/ZKr0euIg==",
       "cpu": [
         "x64"
-      ],
-      "libc": [
-        "musl"
       ],
       "license": "Apache-2.0",
       "optional": true,
--- a/gitnexus/src/core/wiki/grok-client.ts
+++ b/gitnexus/src/core/wiki/grok-client.ts
@@ -1,0 +1,259 @@
+/**
+ * Grok Build CLI client for wiki generation.
+ *
+ * Uses headless `grok --prompt-file` so large wiki prompts are not placed on
+ * argv or stdin (Grok does not read stdin as the prompt).
+ */
+
+import { spawn, execFileSync } from 'child_process';
+import { StringDecoder } from 'string_decoder';
+import fs from 'fs/promises';
+import os from 'os';
+import path from 'path';
+import type { CallLLMOptions, LLMResponse } from './llm-client.js';
+import { logger } from '../logger.js';
+
+export interface GrokConfig {
+  model?: string;
+  workingDirectory?: string;
+  requestTimeoutMs?: number;
+}
+
+const GROK_DISALLOWED_TOOLS =
+  'search_replace,run_terminal_cmd,web_search,web_fetch,spawn_subagent';
+
+let cachedGrokBin: string | null | undefined;
+
+function isVerbose(): boolean {
+  return process.env.GITNEXUS_VERBOSE === '1';
+}
+
+function verboseLog(...args: unknown[]): void {
+  if (isVerbose()) {
+    logger.info({ args }, '[grok-cli]');
+  }
+}
+
+function killChildTree(child: import('child_process').ChildProcess): void {
+  if (process.platform === 'win32' && child.pid !== undefined) {
+    try {
+      execFileSync('taskkill', ['/T', '/F', '/PID', String(child.pid)], {
+        stdio: 'ignore',
+        windowsHide: true,
+      });
+      return;
+    } catch {
+      // Process may have already exited — fall through to child.kill()
+    }
+  }
+  child.kill();
+}
+
+/** Returns `'grok'` when the CLI is on PATH, else null. Cached. */
+export function detectGrokCLI(): string | null {
+  if (cachedGrokBin !== undefined) return cachedGrokBin;
+  try {
+    execFileSync('grok', ['--version'], { stdio: 'ignore', windowsHide: true });
+    cachedGrokBin = 'grok';
+  } catch (err: unknown) {
+    const isNotFound =
+      err instanceof Error && 'code' in err && (err as NodeJS.ErrnoException).code === 'ENOENT';
+    if (!isNotFound && err instanceof Error) {
+      logger.warn(
+        `grok CLI found but --version failed (exit ${(err as { status?: number }).status ?? '?'}). ` +
+          'Ensure it is authenticated: run `grok --version` manually.',
+      );
+    }
+    cachedGrokBin = null;
+  }
+  return cachedGrokBin;
+}
+
+export function resolveGrokConfig(overrides?: Partial<GrokConfig>): GrokConfig {
+  return {
+    model: overrides?.model,
+    workingDirectory: overrides?.workingDirectory,
+    requestTimeoutMs: overrides?.requestTimeoutMs,
+  };
+}
+
+function parseGrokOutput(stdout: string): string {
+  const trimmed = stdout.trim();
+  if (!trimmed) {
+    throw new Error('grok CLI returned empty output');
+  }
+
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(trimmed);
+  } catch {
+    throw new Error('grok CLI returned non-JSON output');
+  }
+
+  if (parsed && typeof parsed === 'object') {
+    const record = parsed as { type?: string; message?: string; text?: unknown };
+    if (record.type === 'error') {
+      throw new Error(record.message || 'grok CLI returned an error');
+    }
+    if (typeof record.text === 'string') {
+      const content = record.text.trim();
+      if (!content) {
+        throw new Error('grok CLI returned empty output');
+      }
+      return content;
+    }
+  }
+
+  throw new Error('grok CLI returned empty output');
+}
+
+/**
+ * Call Grok Build in headless mode and return the assistant text.
+ *
+ * `--cwd` is an empty temp directory (not the repo) so project AGENTS.md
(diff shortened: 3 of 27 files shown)


magyargergo (commented):
## Multi-engine review digest — `feat(wiki): add grok local CLI provider`

Reviewed at head `4376ba3`, diffed against `6088d2e`.

### How this review was produced (engine breakdown)

| Engine | Lanes | Status |
|---|---|---|
| **A — Claude** | GitNexus swarm (6): pr-facts, branch-hygiene, risk-architect, test/CI, security-boundary, docs/DoD. Plus `ce-code-review` personas (8): correctness, reliability, adversarial, security, testing, project-standards, maintainability, api-contract. | ran |
| **B — GPT-5.6** | swarm, adversarial, correctness | ran, **as a substitute** |
| ~~Codex GPT-5.5~~ | swarm, ce, adversarial | **could not run** — organization spend limit |

The intended second engine was Codex GPT-5.5. Those three legs failed with an org spend-limit error and produced **no review output**; nothing in this digest comes from them. GPT-5.6 was run in their place. So this is **two independent engines**, not three — where only Claude's 14 lanes agree, that is convergence *within one engine* and can be a shared false positive, and I've labelled it as such rather than treating it as confirmation.

### Credit where it's due

This is a careful implementation, and several lanes went looking for problems and came back empty. Worth calling out specifically:

- **The safety constants are backed by live experiments, and the comments say so.** Lines 22–27 record that an empty `--tools` *allowlist* does not disable the shell tool — Grok still ran `find` across the user's home directory — and that a *denylist* naming `run_terminal_cmd` (the internal ID, not `bash`) is what actually blocks it. Lines 35–39 record observed turn variance of 3–10 before settling on 15. That is empirical work, not guesswork, and the reasoning is durable for the next maintainer.
- **Deliberate hardening beyond what the sibling providers do:** `--no-plan`, `--no-subagents`, `--disable-web-search`, the tool denylist, and `--sandbox strict` (lines 159–165), with `--cwd` pointed at an empty temp dir (line 167) so the repo's own `AGENTS.md` is not injected into wiki generation. The prompt goes via `--prompt-file` (152–154), keeping large repo-derived text off argv.
- **The subprocess plumbing is correct.** `StringDecoder` on both streams (209–210, 255–256) handles multi-byte UTF-8 split across chunk boundaries. The `resolveOnce`/`rejectOnce` double-settle guards clear the kill timer (214–226). `killChildTree` (53–66) is **byte-identical** to `local-cli-client.ts:38–51`, including the Windows `taskkill /T /F` path — faithful reuse of the established helper, verified by diff.
- **Config plumbing is properly isolated.** `grokModel` is stored separately so HTTP model defaults don't bleed into the local provider, and there are real tests for it (`wiki-flags.test.ts:226`, `:244`, `:501`, `:1317`).

### Headline finding

**On Windows, the provider is unreachable via its own documented install path.** `detectGrokCLI` calls `execFileSync('grok', …)` (line 72) and `runGrok` calls `spawn('grok', …)` (line 195), both without shell resolution. Node cannot execute a `.cmd` shim that way, and an npm-installed `grok` on Windows *is* a `.cmd` shim. The result is not a crash — detection returns `null`, grok is quietly dropped from the provider menu, and `--provider grok` reports *"Grok CLI not found… ensure `grok` is on PATH. Run `grok login`"* (140–142) to a user for whom `grok --version` works fine in PowerShell. The advice sends them to do what they already did.

What makes this more than a portability nitpick is that **the repo already solved it.** `local-cli-client.ts:390–429` (`resolveLocalCommand`) resolves `<name>.cmd` through `where.exe`, prefers a native `.exe`/`.js` entry point, and falls back to `cmd.exe /d /s /c <name>` under a comment that reads *"Last-resort fallback for non-npm Windows installations that only expose a .cmd shim."* Both detection (line 69) and spawn (line 287) route through it. `grok-client.ts` contains no occurrence of `where.exe`, `cmd.exe`, `ComSpec`, or `shell`. This is a deviation from an established in-repo pattern, and reusing `resolveLocalCommand` would likely close it outright. (For fairness: `cursor-client.ts` is partially exposed too — its detection at line 39 uses `execSync`, which *does* go through a shell, but its spawn at line 109 does not. So the robust pattern is `local-cli-client.ts`, and that's the one to copy.)

Both engines reached this independently. GPT-5.6's adversarial lane rated it P1; four Claude lanes rated it P2. I've kept P1 because a supported platform loses the feature entirely and the error actively misdirects.

### Findings proposed as inline comments

| # | Location | Severity | Engines | Verification |
|---|---|---|---|---|
| 1 | `grok-client.ts:72` — Windows `.cmd` shim not resolved; deviates from `resolveLocalCommand` | **P1** | A + B (strong) | `[code-read]` |
| 2 | `grok-client.ts:114` — any zero-exit JSON with `text` accepted as a complete page; `stopReason` never read | **P2** | A + B (strong) | `[code-read]` |
| 3 | `local-cli-subprocess.test.ts:629` — grok block has zero timeout / process-kill coverage | **P2** | A + B (strong) | `[reproduced]` |
| 4 | `grok-client.ts:123` — three distinct failures all report *"returned empty output"*; raw output discarded | **P2** | A + B (strong) | `[code-read]` |
| 5 | `package-lock.json:828` — 16 `"libc"` blocks deleted with no dependency change | **P3** | A + B (strong) | `[reproduced]` |
| 6 | `CHANGELOG.md:9` — `DoD.md:162` names a CHANGELOG edit during PR work a "Not Done" signal | **P3** | A + B (strong) | `[code-read]` |
| 7 | `grok-client.ts:18` — `workingDirectory` declared, copied, never read | **P3** | A only (moderate, 5 lanes) | `[code-read]` |

On #2, one nuance that came out of cross-checking: GPT-5.6's correctness lane verified that **hitting `--max-turns` exits non-zero**, so that case rejects at line 261 before parsing ever happens. The gap survives only for responses that exit 0 with a non-success `stopReason` (e.g. `max_tokens`, `refusal`, `cancelled`). `stopReason` appears nowhere in `src/` or `test/` — confirmed by grep.

On #3, I confirmed the coordinator's claim myself rather than taking it on trust: `describe('Grok CLI subprocess contract')` spans lines 629–882 (end of file), and every occurrence of `requestTimeoutMs`, `useFakeTimers`, `advanceTimersByTime`, and `taskkill` in that file falls between lines 287 and 391 — inside `describe('local CLI timeout')` at line 284, which exercises `callClaudeLLM` only. Zero in the grok block. This matters less in practice than it looks, because `--timeout` defaults to disabled (`cli/index.ts:321`), so the kill timer is unarmed on a default run (line 228 is gated on `requestTimeoutMs !== undefined && > 0`) — but it also means the path is both untested *and* rarely exercised in the wild, which is the combination that tends to rot.

### Validated and refuted

Several lanes ran explicit refutation passes. Reporting these is as useful as reporting the hits:

- **`--retries` is silently ignored for grok — true, but not a grok defect.** Confirmed: `--retries` sets `llmConfig.maxAttempts` (`wiki.ts:533`), which is consumed only at `llm-client.ts:408` inside the HTTP `callLLM` retry config. It reaches no local-CLI client. Cursor, Claude, Codex, and OpenCode behave identically. GPT-5.6 filed it as a grok P3; Claude's correctness lane refuted it as pre-existing conformance. Verification sides with Claude — this belongs in a separate issue about the whole local-provider family, not this PR.
- **POSIX `child.kill()` only signals the direct PID — true, but inherited, not introduced.** `killChildTree` is byte-identical to the sibling's. Grok is also *less* exposed than its siblings here, since the denylist blocks `spawn_subagent` and `run_terminal_cmd`. Demoted from GPT-5.6's P1.
- **Full `process.env` inheritance (lines 199–202) is the established pattern.** `buildChildEnv` (`local-cli-client.ts:218–236`) also spreads `process.env` plus `CI:'1'`; it scrubs only OpenCode-specific variables, and grok has no equivalent. Not a new trust-boundary break. Two Claude lanes raised it; one engine only.
- **Truncation-on-max-turns: refuted**, per the nuance above.
- **`DoD.md` is not missing.** The Claude docs/DoD lane reported it absent. It exists at the repo root, and line 162 is what actually grounds finding #6.
- **Not promoted, single-engine and unverified:** a Windows current-directory-ahead-of-PATH binary-planting scenario (Claude's security lane self-rated it 50% confidence; it depends on libuv search order and no lane tested it); `config.model` reaching argv unvalidated (it's appended *after* the security flags at 169–171, so it cannot reorder them — grok would have to accept a flag as a model name); and `--sandbox strict` failing open if the installed grok build doesn't recognise the flag (an unknown flag would exit non-zero and surface cleanly via line 261, so the silent-failure variant requires the flag to be accepted-and-ignored, which nobody verified).
- **Not introduced by this PR:** the stale KuzuDB note in the plugin skill mirror was pre-existing drift that the re-sync happened to repair.

### Lower-priority items

- Three `@typescript-eslint/no-non-null-assertion` warnings at lines 232, 233, 234 (`config.requestTimeoutMs!`) — I ran eslint to confirm. CI uses `npx eslint .` with no `--max-warnings`, so they don't gate anything. The narrowing is already guaranteed by line 228; hoisting the value into a local `const` before the timer would remove all three.
- Persisting `provider: "grok"` into the global config (`repo-manager.ts:1585`) makes that config confusing for an older CLI that doesn't know the value. Minor, but worth a thought about forward-compat.
- `--provider` accepts arbitrary strings, so a typo can land an unrecognised provider in saved config. Pre-existing.
- The commit that carried the lockfile churn (`41a7218`) has a source-only subject line, which is how 48 lines of unrelated lockfile change ended up easy to miss.

### Test gaps

Coverage here is genuinely substantial — 257 new lines in `local-cli-subprocess.test.ts`, 118 in `wiki-flags.test.ts` — and the argv contract, config resolution, and provider routing are all asserted. The gaps are concentrated in the failure paths: the timeout/kill lifecycle (finding #3), two of `parseGrokOutput`'s four throw sites (lines 117 and 123 are unexercised), `detectGrokCLI`'s non-ENOENT warn branch (77–82), the `onChunk` progress callback for grok (247), and the spawn-error handler (273–275). Note that `src/core/wiki/**` is excluded from the coverage gate (`vitest.config.ts:32`), so no coverage percentage will ever surface these.

### CI status

40 checks pass, 1 fails, 1 correctly skipped. The failure is **Vercel — "Authorization required to deploy"**, which is fork-PR deployment infrastructure and not related to this code. `mergeable: MERGEABLE`, but `mergeStateStatus: BEHIND` — the branch wants a refresh against `main` before merge.

### Coverage limits — what nobody verified

- **No CI job anywhere runs a real `grok` binary.** Every grok test mocks `spawn` and `execFileSync`, so the nine-flag argv contract is asserted against a mock. If xAI renames a flag or drops `--sandbox`, CI stays green and the provider breaks at runtime.
- **No Windows runner exercised the spawn path.** The headline finding is derived from Node's documented `execFile`/`spawn` behaviour with `.cmd` targets plus the sibling's explicit workaround for exactly that case — not from an observed Windows failure. A maintainer with a Windows box should confirm it before we treat the fix as urgent.
- **The worktree cannot `npm install` or `npm run build`,** so no lane executed the compiled CLI. Runtime claims rest on source reading plus the mocked unit suite.
- **Nothing was checked against a live xAI account** — no authentication, sandbox-acceptance, or real turn-count behaviour was observed in this review.
- The `stopReason` values named in finding #2 are inferred from the CLI's JSON contract, not enumerated from grok's documentation.

---

*Automated multi-engine digest (Claude + GPT-5.6; the Codex legs could not run). Please verify each item against the source before acting on it — and thank you for a genuinely thoughtful contribution, particularly the live-verified safety constants.*

gitnexus/src/core/wiki/grok-client.ts
```diff
@@ -0,0 +1,279 @@
...
+    }
+  }
+  child.kill();
+}
+
+/** Returns `'grok'` when the CLI is on PATH, else null. Cached. */
+export function detectGrokCLI(): string | null {
+  if (cachedGrokBin !== undefined) return cachedGrokBin;
+  try {
+    execFileSync('grok', ['--version'], { stdio: 'ignore', windowsHide: true });
```
magyargergo: **P1 — an npm-installed `grok` is unreachable on Windows, and this deviates from the sibling pattern.**

On Windows an npm-installed `grok` is a `.cmd` shim, which `execFileSync` here (and `spawn` at line 195) cannot execute without shell resolution. Detection returns `null`, so grok silently disappears from the provider menu and `--provider grok` reports "Grok CLI not found… ensure `grok` is on PATH" (lines 140-142) to someone whose `grok --version` works fine in PowerShell — advising them to do what they already did.

`local-cli-client.ts:390-429` already solves this: `resolveLocalCommand` finds `<name>.cmd` via `where.exe` and falls back to `cmd.exe /d /s /c <name>`, feeding both detection and spawn. `grok-client.ts` has no `where.exe`/`cmd.exe`/`ComSpec` path.

**Fix:** reuse `resolveLocalCommand` instead of spawning the bare name.

_Inferred from Node's documented `spawn`/`execFile` behaviour with `.cmd` targets plus the sibling's explicit workaround — not from an observed Windows failure. Worth confirming on a Windows box. `[code-read]`_

gitnexus/src/core/wiki/grok-client.ts
```diff
@@ -0,0 +1,279 @@
...
+  } catch {
+    throw new Error('grok CLI returned non-JSON output');
+  }
+
+  if (parsed && typeof parsed === 'object') {
+    const record = parsed as { type?: string; message?: string; text?: unknown };
+    if (record.type === 'error') {
+      throw new Error(record.message || 'grok CLI returned an error');
+    }
+    if (typeof record.text === 'string') {
```
magyargergo: **P2 — a truncated response is accepted as a finished wiki page.**

Line 110 reads only `type`, `message`, and `text`; `stopReason` appears nowhere in `src/` or `test/` (confirmed by grep). Any exit-0 response carrying a `text` field is returned as a complete page, so `stopReason: "max_tokens"` (or `refusal`, `cancelled`) silently produces a wiki page cut off mid-sentence, with no warning and no retry.

Credit where due: hitting `--max-turns` exits non-zero and is correctly rejected at line 261, so that case is already safe. The gap is only the exit-0 stop reasons.

**Fix:** read `stopReason` and throw (or log loudly) on any non-success value before returning `content`. `[code-read]`

gitnexus/test/unit/local-cli-subprocess.test.ts
```diff
@@ -623,3 +623,260 @@ describe('Codex CLI flag contract snapshot', () => {
     expect(args[args.length - 1]).toBe('-');
   });
 });
+
+// ─── Grok CLI subprocess contract ─────────────────────────────────────
+
+describe('Grok CLI subprocess contract', () => {
```
magyargergo, line 634: **P2 — this block has no timeout or process-kill coverage, unlike its sibling.**

Verified by grep: every occurrence of `requestTimeoutMs`, `useFakeTimers`, `advanceTimersByTime` and `taskkill` in this file falls between lines 287 and 391 — inside `describe('local CLI timeout')` at line 284, which exercises `callClaudeLLM` only. This grok block (629-882) has zero. So `grok-client.ts:228-241` and `killChildTree` (53-66), including the Windows `taskkill` branch, are entirely unexercised — and `src/core/wiki/**` is excluded from the coverage gate (`vitest.config.ts:32`), so no coverage number will ever surface it.

**Fix:** mirror the four tests at 295-391 for `callGrokLLM`; the fake-timer scaffolding already exists. Note `makeFakeChild` emits `close` in a microtask so the timeout can never win — the claude tests use a bare `EventEmitter` for exactly that reason. `[reproduced]`

gitnexus/src/core/wiki/grok-client.ts
```diff
@@ -0,0 +1,279 @@
...
+      const content = record.text.trim();
+      if (!content) {
+        throw new Error('grok CLI returned empty output');
+      }
+      return content;
+    }
+  }
+
+  throw new Error('grok CLI returned empty output');
```
magyargergo: **P2 — three different failures all report "returned empty output", and the real output is discarded.**

This line is reached when the payload isn't an object, or when `text` is missing or not a string — i.e. schema drift from grok. All of it surfaces as "grok CLI returned empty output", which is untrue and sends debugging in the wrong direction. Line 106 has the same shape: a single whole-buffer `JSON.parse` means one extra stdout byte (a banner, an update notice) fails the run while throwing away the text that would explain why. On the exit-0 path neither `stdout` nor `stderr` reaches the user.

**Fix:** distinguish "unparseable", "error payload" and "no `text` field", and include a truncated raw excerpt in the message. `[code-read]`

gitnexus/package-lock.json
```diff
@@ -825,9 +825,6 @@
       "cpu": [
         "arm"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
```
magyargergo, line 831: **P3 — unrelated lockfile churn: 16 `"libc"` blocks deleted with no dependency change.**

Verified: `git diff --numstat` on this file is `0 48` — pure deletion, no `package.json` change — removing exactly 16 `"libc"` blocks and leaving zero in the file. These are npm's platform selectors for the `@img/sharp-libvips-*` optional deps, so dropping them lets a musl host resolve glibc-only binaries. Impact is low here (all repo Dockerfiles are `node:22-bookworm-slim`), but `DoD.md:159` lists unrelated churn as a "Not Done" signal, and this arrived inside a commit whose subject mentions only source changes.

This looks like a lockfile regenerated with an npm predating the `libc` field; the repo pins no `packageManager`. The same erosion already landed once via #2980 (26 blocks to 16); this takes it to 0.

**Fix:** revert this file. `[reproduced]`

gitnexus/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to GitNexus will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- **Wiki** — local Grok Build CLI provider (`gitnexus wiki --provider grok`) uses the authenticated `grok` CLI instead of an API key
```
magyargergo: **P3 — repo convention: `CHANGELOG.md` isn't edited in feature PRs.**

`DoD.md:162` lists "`CHANGELOG.md` was edited during PR work" under **"Not Done" Signals**, and line 138 notes the file is owned by the release process. Practice matches: in `git log origin/main` this file is touched almost entirely by `chore: release vX.Y.Z` commits, and `CONTRIBUTING.md:259` feeds it straight to `gh release create --notes-file`.

Genuinely easy to miss — the PR template's only changelog line (`:51`) is about `AGENTS.md` overlays, not this file — and the entry itself is accurate and well-formed. Convention note, not a criticism.

**Fix:** drop this hunk; the release process will pick it up. `[code-read]`

gitnexus/src/core/wiki/grok-client.ts
```diff
@@ -0,0 +1,279 @@
...
+import { StringDecoder } from 'string_decoder';
+import fs from 'fs/promises';
+import os from 'os';
+import path from 'path';
+import type { CallLLMOptions, LLMResponse } from './llm-client.js';
+import { logger } from '../logger.js';
+
+export interface GrokConfig {
+  model?: string;
+  workingDirectory?: string;
```
magyargergo: **P3 — `workingDirectory` is declared, copied, and never read.**

`resolveGrokConfig` copies it through (line 91), but `generator.ts:229-235` only ever passes `model` and `requestTimeoutMs`, and `runGrok` deliberately uses `tempDir` as `cwd` (line 178) so the repo's `AGENTS.md` isn't injected — which is the right call. The field therefore can't do what its name promises, and a future caller setting it would reasonably expect it to take effect.

**Fix:** drop the field from `GrokConfig` and from `resolveGrokConfig`.

_Five Claude lanes flagged this; the second engine did not — treat it as one engine's read. `[code-read]`_

gitnexus/src/core/wiki/grok-client.ts
```diff
@@ -0,0 +1,279 @@
...
+
+    verboseLog('Spawning:', grokBin, args.join(' ').replace(promptPath, '[prompt-file]'));
+    if (config.model) {
+      verboseLog('Model:', config.model);
+    }
+
+    const content = await runGrok(grokBin, args, tempDir, config, options);
+    return { content };
+  } finally {
+    await fs.rm(tempDir, { recursive: true, force: true }).catch(() => undefined);
```
magyargergo: **P2 — on timeout, this deletes the live child's working directory, sandbox root, and prompt file.**

The timeout callback calls `killChildTree(child)` and then `rejectOnce(...)` synchronously (lines 229-240) without awaiting the child's `close`. So this `finally` runs while grok may still be alive — `killChildTree` sends a single un-escalated SIGTERM on POSIX (line 65), with no SIGKILL follow-up and no process-group kill.

For grok, `tempDir` is simultaneously the spawn `cwd`, the `--cwd` argument, the `--sandbox strict` confinement root, and the home of `--prompt-file` (lines 146-167). Removing it under a running agentic session leaves an orphan that can keep working — and billing — for up to the 15-turn cap with its own cwd and sandbox root gone.

The codex sibling avoids this by construction: in `callCodexLLM` (`local-cli-client.ts:130-163`) the temp dir carries only `--output-last-message`, while the child's cwd is set separately via `--cd`. This is a grok-specific consequence of reusing one directory for everything.

**Fix:** await the child's `close` (or an escalating SIGTERM-then-SIGKILL with a grace period) before removing `tempDir`.

_Reproduced by two independent lanes and corroborated by the second engine. Reachable only when `--timeout` is passed, since it defaults to disabled — hence P2, though one lane rated it P1. `[reproduced]`_

magyargergo (commented):
### Addendum to the multi-engine digest above — one finding I under-reported

My earlier digest folded this into the "validated and refuted" section, where I wrote that the POSIX `child.kill()` limitation is *"inherited, not introduced"* because `killChildTree` is byte-identical to `local-cli-client.ts`. That part is correct, but it caused me to dismiss a **separate and genuinely grok-specific** problem alongside it. Three lanes across both engines reported it, two of them with executed probes rather than code reading, so it belongs in the review rather than in a footnote.

The distinct issue is not *how* the child is killed — it is that **the timeout path deletes the directory the child is still living in**, and unlike the codex sibling, for grok that directory is the process's own working directory, its `--sandbox strict` confinement root, and the home of its `--prompt-file`. See the inline comment.

I've rated it **P2** rather than the P1 one lane assigned, because `--timeout` defaults to disabled (`cli/index.ts:321`), so the kill timer is unarmed on a default `gitnexus wiki --provider grok` run and the path is only reachable when a user opts in. Lanes split P1/P3 on it; P2 is my own call, and I'd rather show the disagreement than launder it into false consensus.

Corroboration: Claude's ce-adversarial lane (P1, reproduced with a standalone probe reporting *"grok pid … is STILL ALIVE after timeout+cleanup, cwd deleted underneath it"*), Claude's ce-correctness lane (P3, probed independently — observed `cwdReadable=false` for the remaining life of a SIGTERM-ignoring child), and GPT-5.6's adversarial lane (reproduced timeout/signal leaks). Cross-engine and reproduced, which is the strongest evidence class this review produced.

The sibling comparison I verified myself after the fact: in `callCodexLLM` (`local-cli-client.ts:130-163`) the temp dir holds only `--output-last-message`, while the child's cwd is set separately via `--cd <workingDirectory>`. So deleting that temp dir in its `finally` cannot pull the ground out from under a running child. Grok's single-directory design is what makes the same `finally` harmful here.

Nothing else in the digest changes. The engine caveat still stands: Codex GPT-5.5 could not run (organization spend limit) and GPT-5.6 was substituted.

gitnexus/src/core/wiki/local-cli-client.ts
```diff
@@ -387,10 +389,33 @@ function getDetectedCommand(provider: LocalAgentProvider): LocalCommand | null {
...
+ * Resolve a PATH command for spawn/execFile without a shell.
+ * On Windows, npm shims are `.cmd` files that Node cannot execFile/spawn
+ * directly — prefer a native `.exe` from `where.exe`, else `cmd.exe /d /s /c`.
+ */
+export function resolveWindowsCliCommand(displayName: string): ResolvedCliCommand {
+  if (process.platform !== 'win32') {
+    return { displayName, command: displayName, argsPrefix: [] };
+  }
+
+  const located = findWindowsCommand(`${displayName}.cmd`) || findWindowsCommand(displayName);
```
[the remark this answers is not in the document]
jreakin: The suggested swap of lookup order does not select a native `.exe`. `findWindowsCommand` returns only the first `where.exe` line. On an npm install, `where grok` typically lists `grok.cmd` first, so looking up `grok` before `grok.cmd` still does not yield `.exe`.

When the hit is not already `.exe`, falling through to `cmd.exe /d /s /c grok` is the sibling contract: Node must not spawn a `.cmd` path. Preferring a native binary would mean scanning every `where` hit for `\.exe$`, which is a different change than this one-liner.

gitnexus/test/unit/local-cli-subprocess.test.ts
```diff
@@ -623,3 +628,597 @@ describe('Codex CLI flag contract snapshot', () => {
...
+
+  it('writes system + separator + user prompt to --prompt-file, not stdin', async () => {
+    const { callGrokLLM } = await loadGrokClient();
+
+    await callGrokLLM('user prompt', {}, 'system prompt');
+
+    const args = spawnSpy.mock.calls[0][1] as string[];
+    const fileIdx = args.indexOf('--prompt-file');
+    const promptPath = args[fileIdx + 1];
+    // File is removed after the call; capture contents via spawn-time read.
```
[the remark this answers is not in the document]
jreakin: This test only asserts the prompt is not on stdin and that `--prompt-file` points at a `gitnexus-wiki-grok-` temp path. File contents are checked in the next test (`writes the concatenated prompt before spawn`), which `readFileSync`s the file at spawn time. A wrong or empty prompt fails that test.

gitnexus/src/core/wiki/grok-client.ts
```diff
@@ -0,0 +1,359 @@
...
+    let settled = false;
+    let timedOut = false;
+    let killTimer: ReturnType<typeof setTimeout> | undefined;
+    let killEscalate: ReturnType<typeof setTimeout> | undefined;
+    let hardDeadline: ReturnType<typeof setTimeout> | undefined;
+    const timeoutMs = config.requestTimeoutMs;
+    const timeoutError =
+      timeoutMs !== undefined && timeoutMs > 0
+        ? new Error(
+            `grok CLI timed out after ${
```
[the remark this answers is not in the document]
jreakin: `--timeout` is a positive integer number of seconds (`parsePositiveIntegerOption`). Wiki sets `requestTimeoutMs = timeoutSeconds * 1000`, so the CLI never passes 1500ms. This `Math.round` formatter matches Claude/Codex in `local-cli-client.ts`. Switching to `toFixed(1)` would be a family-wide message change, not a grok-only bug.

gitnexus/test/unit/local-cli-subprocess.test.ts
```diff
@@ -623,3 +628,604 @@ describe('Codex CLI flag contract snapshot', () => {
...
+  });
+
+  it('does not remove the temp dir until the child closes after timeout', async () => {
+    const fs = await import('fs');
+    const child = hangingChild();
+    const { callGrokLLM, spawnSpy } = await loadGrokWithChild(child);
+    const promise = callGrokLLM('prompt', { requestTimeoutMs: 5000 });
+    await waitForSpawn(spawnSpy);
+    const cwd = (spawnSpy.mock.calls[0][2] as { cwd: string }).cwd;
+    expect(fs.existsSync(cwd)).toBe(true);
```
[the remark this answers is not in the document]
jreakin: Valid for the original test: it had to leave the dir in place until after hard-deadline reject. `c6172f68` still asserts the dir exists at that point, then emits `close` so the late-rm path deletes it.

